PR #3681, opened 2021-05-26
shell: add plugin to detect first task exit

garlick:
I thought this was going to be a throw-away demo of a plugin that detects the first task exit and logs an exec eventlog entry, but after adding a timeout that throws a fatal job exception 30s after that, it seems like it might actually be sufficient to check off that feature for now.

So posting this as a WIP to get design feedback.

Possibly the throw-away name should be changed, as `doom` was chosen becuase it had morbid connotations and was short :-)

Is it OK for the shell to log the event directly to the eventlog or should it be notifying the `job-exec` service?

I sure am liking the ability we have to implement stuff like this as plugins.

### DIFF
--- a/src/shell/doom.c
+++ b/src/shell/doom.c
@@ -1,0 +1,245 @@
+/************************************************************\
+ * Copyright 2021 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* doom.c - log first task exit
+ *
+ * Each shell sends a message to shell-0 when its first task exits.
+ * Shell-0 posts an event to the exec eventlog for the first one received.
+ *
+ * Shell-0 sets a timer and posts a fatal exception when the timer fires.
+ */
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+#include <flux/core.h>
+#include <jansson.h>
+#include <assert.h>
+
+#include "src/common/libeventlog/eventlog.h"
+#include "src/common/libutil/fsd.h"
+
+#include "builtins.h"
+#include "internal.h"
+#include "task.h"
+
+#define TIMEOUT_NONE (-1.)
+
+static const double default_timeout = 30.;
+
+struct shell_doom {
+    flux_shell_t *shell;
+    bool done; // event already posted (shell rank 0) or message sent (> 0)
+    flux_watcher_t *timer;
+    double timeout;
+};
+
+static void doom_post (struct shell_doom *doom, json_t *task_info)
+{
+    flux_kvs_txn_t *txn;
+    json_t *entry = NULL;
+    char *entrystr = NULL;
+    flux_future_t *f = NULL;
+
+    assert (doom->shell->info->shell_rank == 0);
+
+    if (!(txn = flux_kvs_txn_create ())
+        || !(entry = eventlog_entry_pack (0.,
+                                          "shell.task-exit",
+                                          "O",
+                                          task_info))
+        || !(entrystr = eventlog_entry_encode (entry))
+        || flux_kvs_txn_put (txn,
+                             FLUX_KVS_APPEND,
+                             "exec.eventlog",
+                             entrystr) < 0
+        || !(f = flux_kvs_commit (doom->shell->h, NULL, 0, txn)))
+        shell_log_errno ("error posting task-exit eventlog entry");
+
+    if (f && doom->timeout != TIMEOUT_NONE)
+        flux_watcher_start (doom->timer);
+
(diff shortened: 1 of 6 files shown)


garlick:
Force pushed a bit of cleanup here.

grondo:
I tested this up to 256 shells on fluke with no detected fallout, so this seems good to go in!

I was initially wondering if the eventlog event should be named `first-task-exit` vs `task-exit`, but I think since it is the first `task-exit` event in the eventlog, the "first" is implied!

grondo:
Slurm also has an option to kill a job immediately on the first nonzero task exit. It seems like this plugin could be extended to support that feature if necessary.

garlick:
> Seems fine as the internal name for the plugin, but probably the shell option to set the timeout should be renamed to something that better indicates what it does.

>  Slurm also has an option to kill a job immediately on the first nonzero task exit.

OK I'll throw out some probably horrible ideas and defer to others like you and @ryanday36 for better suggestions?
```
$ flux mini run -ojobicide.exit-timeout=10s  ...
$ flux mini run -ojobicide.exit-on-error ...
```

grondo:
I like just `-o exit-timeout=30s`.

Then add documentation to `JOBSPEC OPTIONS` section of `flux-shell(1)` and call it good for now?
This fixes a real problem so it would be cool to get it in our next tag.

garlick:
Just hitting the road so I'd you'd like to force push something here feel
free. O/w I'll make changes tonight or tomorrow.  Thanks!

src/shell/doom.c
```diff
@@ -0,0 +1,243 @@
...
+    flux_shell_task_t *task;
+
+    if (!(shell = flux_plugin_get_shell (p))
+        || !(doom = flux_plugin_aux_get (p, "doom"))
+        || !(task = flux_shell_current_task (shell)))
+        return -1;
+    if (!doom->done) {
+        json_t *entry;
+
+        if (!(entry = doom_eventlog_entry_create (task)))
```
grondo, line 146: Hm, maybe it would be better to create the eventlog entry on the rank 0 shell? This avoids strange timestamps when clocks have drifted:

```ƒ(s=392,d=0,builddir) grondo@fluke2:~/git/f.git$ flux mini run -N 392 -n 392 -vvv sh -c 'sleep .$((392 - $FLUX_TASK_RANK))'
jobid: ƒ3s3MjkBZ
0.000s: job.submit {"userid":6885,"urgency":16,"flags":0}
0.014s: job.depend
0.014s: job.priority {"priority":16}
0.021s: job.alloc {"annotations":{"sched":{"resource_summary":"rank[0-391]/core0"}}}
0.382s: job.start
0.022s: exec.init
0.024s: exec.starting
0.327s: exec.shell.init {"leader-rank":0,"size":392,"service":"6885-shell-6320950345728"}
0.379s: exec.shell.start {"task-count":392}
-10.770s: exec.shell.task-exit {"localid":0,"rank":391,"state":"Exited","pid":1327786,"wait_status":0,"signaled":0,"exitcode":0}
1.335s: exec.complete {"status":0}
1.335s: exec.done
1.335s: job.finish {"status":0}
1.339s: job.release {"ranks":"all","final":true}
1.342s: job.free
1.342s: job.clean
```

garlick:
OK, force pushed with:
* Eventlog "envelope" is created on rank 0 for uniform timestamps
* Option is `exit-timeout`
* Added shell(1) blurb
* Rebased on current master

src/shell/doom.c
```diff
@@ -0,0 +1,231 @@
...
+        errno = saved_errno;
+    }
+}
+
+static int parse_args (flux_shell_t *shell, double *timeoutp)
+{
+    const char *fsd = NULL;
+    double timeout;
+
+    if (flux_shell_getopt_unpack (shell, "exit-timeout", "s", &fsd) < 0)
```
grondo, line 164: Should an integer or real work here as well? (implied seconds), maybe unpack as a json_t object and use json_string_value()?

garlick: Good idea - that would avoid the somewhat generic error you get now with an integer argument.

garlick: Oh, just realized it already does since I was using the FSD parser

https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_23.html

I'll push a fix to the value error message though.

garlick: Oh wait, no it doesn't get past jansson does it.  Fixing.

garlick: OK, force pushed a change and some more tests.

garlick:
Oops, I was using `json_real_value()` where I should have been using `json_number_value()` and had inverted the logic in the test for it.  Pushed again.

grondo:
That json_t to FSD snippet might eventually be useful in libutil as this is a common use case. Too bad we can't extend `unpack()` with a special symbol to unpack a number or string as FSD into a double...

grondo (approved):
LGTM, nice work!

grondo:
Ok to set MWP?

garlick:
Thanks, I'll go ahead and set it.

PR #3685, opened 2021-05-27
libflux: plugin: make FLUX_PLUGIN_ARG_UPDATE the default for set/pack

### LINKED ISSUE #3683: libflux: make FLUX_PLUGIN_ARG_UPDATE default for flux_plugin_arg_set/pack
The behavior of `FLUX_PLUGIN_ARG_UPDATE` is probably the desired behavior in the common case for `flux_plugin_arg_set(3)` and `flux_plugin_arg_pack(3)`. Especially now that we have a few plugin "stacks" in which the plugin args (especially the "out" arguments) are accumulated across multiple plugin callbacks.

The default should be changed to update plugin args, with a new flag `FLUX_PLUGIN_ARG_REPLACE` or similar used to get the old behavior. Not sure the best way to do this without breaking existing users. Maybe we go ahead and break the API, removing `FLUX_PLUGIN_ARG_UPDATE` so that existing callers get a compile error. For the ABI (though it is very unlikely existing plugins are out there that won't be recompiled) we could abandon `2` and make FLUX_PLUGIN_ARG_REPLACE `4`.

Edit: or a simpler approach, keep `FLUX_PLUGIN_ARG_UPDATE` but document it as the default behavior and ignore it.

grondo:
Fixes #3683

This was pretty straightforward, but @SteVwonder and @cmoussa1 should sign off since I think they are developing plugins external to flux-core.

This PR removes `FLUX_PLUGIN_ARG_UPDATE` and makes its behavior the default. The new flag `FLUX_PLUGIN_ARG_REPLACE` can be used to get the old default behavior. Therefore existing out-of-tree plugins may get compile-time errors if they use `FLUX_PLUGIN_ARG_UPDATE`

### DIFF
--- a/doc/man7/flux-jobtap-plugins.rst
+++ b/doc/man7/flux-jobtap-plugins.rst
@@ -70,7 +70,7 @@ entry      o    posted eventlog entry, including context
 ========== ==== ==========================================
 
 Return arguments can be packed using the ``FLUX_PLUGIN_ARG_OUT`` and
-optionally ``FLUX_PLUGIN_ARG_UPDATE`` flags. For example to return
+optionally ``FLUX_PLUGIN_ARG_REPLACE`` flags. For example to return
 a priority::
 
    rc = flux_plugin_arg_pack (args, FLUX_PLUGIN_ARG_OUT,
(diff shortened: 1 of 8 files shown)


cmoussa1 (approved):
Thanks @grondo! I think these changes make sense to me, so LGTM.

garlick (approved):
FWIW, makes sense to me too!

grondo:
Ok, setting MWP since likelihood of critical impact is low.

PR #3687, opened 2021-05-27
job-manager: jobtap enhancements

grondo:
This PR is a set of jobtap plugin enhancements which were developed while working on a simple "begin time" dependency plugin. The features include:

 * Adopt an idiom where builtin plugins are named with a leading `.` (like hidden filesystem files). This makes it so that builtin plugins are _not_ removed by `*` or `all`, and are hidden by default in `flux jobtap list`.
 * Add jobtap interface for `job_aux_*` functions so that a plugin can associate state with a jobid, and the state can be destructed automatically when the job goes away
 * Add flux_jobtap_raise_exception() so that jobtap plugins can synchronously raise a job exception on a job
 * Recheck job dependencies (i.e. call `job.dependency.*`) on job-manager restart and plugin load
 * Remove the special-case code to ensure jobs move past PRIORITY state when no plugin is loaded. Instead move the default priority plugin back to a builtin. If the `.priority-default` is unloaded, jobs will hang in PRIORITY state.
 * Update documentation
 * Update tests


src/modules/job-manager/jobtap.c
```diff
@@ -254,7 +254,7 @@ static int jobtap_remove (struct jobtap *jobtap,
     while (p) {
         const char *name = jobtap_plugin_name (p);
         if (all
-            || (isglob && fnmatch (arg, name, 0) == 0)
+            || (isglob && fnmatch (arg, name, FNM_PERIOD) == 0)
```
garlick, line 327: Oh ha hah, FNM_PERIOD worked out nicely there.  Points for co-opting a file system convention.

garlick (approved):
This all LGTM.  It feels like this design is maturing nicely!  Thanks for taking the time to write the additional documentation.

grondo:
Well I have some other enhancements I was going to push here, but seems like maybe this PR is growing a little too much (and it will be nice to at least have the `.priority-default` builtin plugin in), so I'll set MWP and save other stuff for a post-0.27 PR.

PR #3611, opened 2021-04-23
job-list: support inactive-cache-size option

chu11:
Per issue #3607, support an option to limit the number of inactive jobs cached by the `job-list` module.

Primary benefit: improve performance over time as the `job-list` module would use up a large amount of memory.

Consequence: job listing tools will no longer see inactive jobs "evicted" by the `job-list` module.  Because of this, the default inactive cache size is set to `-1`, which is "unlimited" cache size and the same as current behavior.  This is to maintain backwards compatibility until a more long term solution involving the `job-archive` can be designed and developed.  The option is there for any power users that may be seeing throughput limitations.

Medium term benefit: #3610 is the long term solution and a "cache size" config will be needed anyway.  So implement this easy-ish solution for the medium term.

Performance example:

w/o the option turned on

```
Running throughput.py 4096 jobs per iteration, 25 iterations
throughput:     26.0 job/s (script:  25.9 job/s)
throughput:     24.4 job/s (script:  24.4 job/s)
throughput:     23.7 job/s (script:  23.7 job/s)
throughput:     23.2 job/s (script:  23.2 job/s)
throughput:     22.9 job/s (script:  22.9 job/s)
throughput:     22.4 job/s (script:  22.3 job/s)
throughput:     21.9 job/s (script:  21.8 job/s)
throughput:     21.6 job/s (script:  21.5 job/s)
throughput:     21.4 job/s (script:  21.3 job/s)
throughput:     20.7 job/s (script:  20.6 job/s)
throughput:     20.6 job/s (script:  20.6 job/s)
throughput:     20.2 job/s (script:  20.1 job/s)
throughput:     19.5 job/s (script:  19.4 job/s)
throughput:     18.8 job/s (script:  18.8 job/s)
throughput:     18.7 job/s (script:  18.6 job/s)
throughput:     18.5 job/s (script:  18.4 job/s)
throughput:     18.4 job/s (script:  18.4 job/s)
throughput:     18.0 job/s (script:  18.0 job/s)
throughput:     17.6 job/s (script:  17.6 job/s)
throughput:     17.2 job/s (script:  17.2 job/s)
throughput:     17.0 job/s (script:  17.0 job/s)
throughput:     16.7 job/s (script:  16.7 job/s)
throughput:     16.3 job/s (script:  16.3 job/s)
throughput:     15.8 job/s (script:  15.8 job/s)
throughput:     15.8 job/s (script:  15.8 job/s)
```

w/ the option turned on with the following config

```
[job-list]
[...]

Issue #3607, opened 2021-04-21
job-list: support cache size configuration


chu11:
As discussed in #3583 the job-list module can impact job throughput and general performance as it stores the history of all jobs without ever freeing them, thus using more and more memory.  In one test, the `job-list` module alone was responsible for the broker to use twice as much memory.

There has been long term discussion of how to deal with this, such as archiving jobs, and perhaps having `job-list` access archived jobs for the user.

However short term, we should support a configuration that allows the user to limit the amount of job history the `job-list` module stores, so users can control this and its potential effect on performance.  Such a configuration shouldn't matter much, b/c access of archived jobs should be independent of how much job history we're willing to cache.

In #3583 I showed an experimental patch I used to show performance impact.

```
diff --git a/src/modules/job-list/job_state.c b/src/modules/job-list/job_state.c
index fb51207..0ec1040 100644
--- a/src/modules/job-list/job_state.c
+++ b/src/modules/job-list/job_state.c
@@ -1093,6 +1093,12 @@ static int depthfirst_map_one (struct list_ctx *ctx, const char *key,
     }
     job_insert_list (ctx->jsctx, job, job->state);
 
+    if (zhashx_size (ctx->jsctx->index) > 10000) {
+        struct job *ptr = NULL;
+        zlistx_last (ctx->jsctx->inactive);
+        ptr = zlistx_detach_cur (ctx->jsctx->inactive);
+        zhashx_delete (ctx->jsctx->index, &ptr->id);
+    }
     rc = 1;
 done:
     if (rc < 0)
@@ -1310,6 +1316,12 @@ static int journal_submit_event (struct job_state_ctx *jsctx,
             errno = ENOMEM;
             return -1;
         }
+        if (zhashx_size (jsctx->index) > 10000) {
+            struct job *ptr = NULL;
+            zlistx_last (jsctx->inactive);
+            ptr = zlistx_detach_cur (jsctx->inactive);
+            zhashx_delete (jsctx->index, &ptr->id);
+        }
     }
 
     if (job_update_eventlog_seq (jsctx, job, eventlog_seq) == 1)
```

I believe something not so dissimilar to this patch should work (obviously should be cleaned up, errors handled, and the 10000 would be configurable).

garlick:
Any other fallout besides reducing how much history `flux job list -a`  can display?

Since it looks like `flux job list -a` defaults to `--count=1000` anyway.  Would it make sense to make the default that low?

chu11:
[...]

PR #3688, opened 2021-05-27
job-list: support RPC to purge cache

chu11:
Per comment by @grondo on #3611, here is an alternative take on a solution for #3607.

Instead of supporting a configurable cache size, instead support an RPC to purge the cache of some number (or all) cached inactive jobs.

This helps in two ways.  1) Reduce/eliminate cached data, which can help w/ performance.  2) "reset" the job-list output.  This can be useful for testing purposes or just general use when you wish for really old inactive jobs to no longer be listed.

Throughput tests should similar results to what was in #3611, so won't post them.  Basically regularly calling `flux job list-purge-inactive` ensures job throughput doesn't decrease that much over time.

### DIFF
--- a/t/t2262-job-list-purge.t
+++ b/t/t2262-job-list-purge.t
@@ -1,0 +1,179 @@
+#!/bin/sh
+
+test_description='Test flux job list purge inactive'
+
+. $(dirname $0)/job-list/job-list-helper.sh
+
+. $(dirname $0)/sharness.sh
+
+export FLUX_CONF_DIR=$(pwd)
+test_under_flux 4 job
+
+RPC=${FLUX_BUILD_DIR}/t/request/rpc
+
+get_completed_count() {
+    count=`flux jobs --stats-only 2>&1 | awk '{print \$3}'`
+    echo ${count}
+}
+
+get_failed_count() {
+    count=`flux jobs --stats-only 2>&1 | awk '{print \$5}'`
+    echo ${count}
+}
+
+# submit a whole bunch of jobs for job list testing
+#
+# - the first loop of job submissions are intended to have some jobs run
+#   quickly and complete
+# - the second loop of job submissions are intended to eat up all resources
+# - job ids are stored in files in the order we expect them to be listed
+#   - pending jobs - by priority (highest first), job id (smaller first)
+#   - running jobs - by start time (most recent first)
+#   - inactive jobs - by completion time (most recent first)
+#
+# TODO
+# - alternate userid job listing
+
+test_expect_success 'submit jobs for testing' '
+        #  Create `hostname` and `sleep` jobspec
+        #  N.B. Used w/ `flux job submit` for serial job submission
+        #  for efficiency (vs serial `flux mini submit`.
+        #
+        flux mini submit --dry-run hostname >hostname.json &&
+        flux mini submit --dry-run --time-limit=5m sleep 600 > sleeplong.json &&
+        #
+        # submit jobs that will complete
+        #
+        for i in $(seq 0 3); do
+                flux job submit hostname.json >> inactiveids
+                fj_wait_event `tail -n 1 inactiveids` clean
+        done &&
+        #
+        #  Currently all inactive ids are "completed"
+        #
+        tac inactiveids | flux job id > completed.ids &&
+        #
+        #  Run jobs that will fail, copy its JOBID to both inactive and
+        #   failed lists.
+        #
+        ! jobid=`flux mini submit --wait nosuchcommand` &&
+        echo $jobid >> inactiveids &&
+        flux job id $jobid >> failedids &&
+        jobid=`flux mini submit --time-limit=0.25 sleep 30` &&
+        fj_wait_event $jobid clean &&
+        echo $jobid >> inactiveids &&
+        flux job id $jobid >> failedids &&
+        #
+        #  Submit 8 sleep jobs to fill up resources
+        #
+        for i in $(seq 0 7); do
+                flux job submit sleeplong.json >> runningids
+                fj_wait_event $jobid alloc
+        done &&
+        tac runningids | flux job id > running.ids &&
+        #
+        #  Submit a job and cancel it, copy its JOBID to both inactive
+        #    and failed lists
+        #
+        jobid=`flux mini submit --job-name=canceledjob sleep 30` &&
+        flux job wait-event $jobid depend &&
+        flux job cancel $jobid &&
+        flux job wait-event $jobid clean &&
+        flux job id $jobid >> inactiveids &&
+        flux job id $jobid >> failedids &&
+        tac inactiveids | flux job id > inactive.ids &&
+        tac failedids | flux job id > failed.ids &&
+        # no pending jobs, create empty file of pending job ids
+        touch pending.ids &&
+        #
+        #  Synchronize all expected states
+        #
+        wait_states
+'
+
+test_expect_success HAVE_JQ 'flux job list all inactive jobs' '
+        flux job list -s inactive | jq .id > list_inactive1.out &&
+        test_cmp list_inactive1.out inactive.ids
+'
+
+test_expect_success HAVE_JQ 'flux job list-ids lists all jobs' '
+        ids=`cat inactive.ids | xargs` &&
+        flux job list-ids ${ids} | jq .id > list_inactive2.out &&
+        test_cmp list_inactive2.out inactive.ids
+'
+
+test_expect_success HAVE_JQ 'job-list stats has correct counts' '
+        count=`flux module stats --parse "jobs.inactive" job-list` &&
+        test ${count} -eq $(state_count inactive) &&
+        count=`get_completed_count` &&
+        test ${count} -eq $(state_count completed) &&
+        count=`get_failed_count` &&
+        test ${count} -eq $(state_count failed)
+'
+
+test_expect_success HAVE_JQ 'flux job list-purge 2 jobs' '
+        flux job list-purge --count=2
+'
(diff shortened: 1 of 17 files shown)


grondo:
Sorry for taking so long to do a review:

A couple early comments before I look at any code:

Should the pure update the job stats as well?
```console
$ flux jobs --stats-only
0 running, 256 completed, 0 failed, 0 pending
$ flux job list-purge-inactive
$ flux jobs --stats-only
0 running, 256 completed, 0 failed, 0 pending
```

After the purge, when trying `flux jobs JOBID` where JOBID has been purged, I'm seeing a strange non-error error:
```console
$ flux mini submit hostname
ƒ2LrxhKkj
$ flux job list-purge-inactive
$ flux jobs ƒ2LrxhKkj
rpc: 
```

I have no idea where that single `rpc:` line is coming from :shrug:

chu11:
> Should the purge update the job stats as well?

Yeah, it should.  I'm not sure what's going on.  I guess I had missed testing this, so it'll be good to add a unit test too.

chu11:
Just re-pushed.  I completely forgot/missed that there were two stats outputs in `job-list` instead of 1.  Added additional tests to cover them both.

The `rpc:` was b/c in the bindings I don't check for the correct errno to output an appropriate message for a jobid being purged.  It then goes down into a general `rpc: {err.strerror}` output, but I didn't set an error message.  So I just check for the errno now and output that the job Id was purged.

I should note, I couldn't think of a good errno to use when encountering a purged job id, so I used EFAULT.  Which may not be the best one.  I could alternately set an error message to be returned instead.

chu11:
re-pushed, i just renamed `list-purge-inactive` to `list-purge` and decided to rename the RPC and various other things from "purge inactive" to just "purge" .. just because :P

I was going to go with the suggested `--purge`, but I forgot about the `--count` option to limit the number of purges.  We could go with `--purge=N|all`.  But ended up going this path instead.

Also, figured with:

```
   list            List jobs
   list-inactive   List Inactive jobs
   list-ids        List job(s) by id
   list-purge      Purge inactive jobs from job-list cache
```

we could consider create a set of sub-commands, like `flux job list all | inactive | ids | purge` someday.

grondo:
> but I forgot about the --count option to limit the number of purges.

Ah, I had noticed `list-inactive` already has a `--count` option:

Usage: flux-job list-inactive [OPTIONS]
List Inactive jobs
  -c, --count=N          Limit output to N jobs
  -s, --since=T          Limit output to jobs that entered the inactive state
                         since timestamp T

I realize now it would be confusing since `list --count` returns the most recent COUNT jobs, and `list-purge --count` would presumably purge the oldest COUNT jobs, so I think  you were right to do it this way for now.

chu11:
re-pushed with re-done tests to try and get coverage higher.  not purging failed/canceled jobs missed a big block.

grondo:
> Edit: I'm a little perplexed, code coverage doesn't think I'm covering a certain block of code. But the sharness tests that pass should basically guarantee that its being covered (job fail count going down).

I ran the latest version of your branch and those lines do show as covered, so perhaps this is a codecov issue? Not sure since it appears that the latest coverage report uploaded successfully. :shrug:

chu11:
re-pushed adding a timeout job to the list of jobs tested, so hopefully can capture that 1 extra line :P

chu11:
I get the error "Unmerged Base Commits: The base reference contains commits not present in the head reference. Please merge the base reference into the head reference to see the most up to date coverage information", which is why the codecov isn't updating.  However, the current head of master seems to have passed in codecov.  So I have no idea what's going wrong.  Perhaps just a bug in codecov.   I tried a few stupid things (cherry-pick all my commits on top of master again), but to no avail.

grondo:
That has been happening on all PR codecov runs lately. I haven't tried to run it down, perhaps due to the way github checkout action runs. It isn't specific to your PR though.

Still confused why the coverage is different when run locally vs codecov -- maybe something else going on CI or as you say a codecov bug.

grondo:
Ah, I think I found the/a problem. When a PR is merged to master the fix for annotated tags in commit e24afca845d24ad6f429b6bed5bcf8e8298efa8f causes the CI to fail. This means that coverage reports haven't been uploaded for master since that point. :facepalm: 

I'll have to figure out a workaround. Not sure this will fix the erroneous missing coverage lines in this PR though. That one still mystifies me.

chu11:
#3716 and #3717 didn't seem to help w/ the coverage issue here. Perhaps just a bug in codecov :-(

t/t2262-job-list-purge.t
```diff
@@ -0,0 +1,181 @@
...
+#
+state_count() {
+    state_ids "$@" | wc -l
+}
+
+test_expect_success 'job-list: submit 6 jobs' '
+        #  Create `hostname` and `sleep` jobspec
+        #
+        flux jobspec --format json srun -N1 hostname > hostname.json &&
+        flux jobspec --format json srun -N1 sleep 300 > sleeplong.json &&
```
grondo, line 45: We probably should avoid using `flux jobspec` in any new tests, I think we were considering deprecating it?

t/t2262-job-list-purge.t
```diff
@@ -0,0 +1,181 @@
...
+        #
+        # submit jobs that will complete
+        #
+        for i in $(seq 1 4); do \
+                flux job submit hostname.json >> completedids; \
+                fj_wait_event `tail -n 1 completedids` clean ; \
+        done &&
```
grondo, line 52: This could be done a lot more efficiently and succinctly with 
```
flux mini submit --cc=1-4 --wait hostname
```

t/t2262-job-list-purge.t
```diff
@@ -0,0 +1,181 @@
...
+        #
+        jobid=`flux mini submit nosuchcommand` &&
+        fj_wait_event $jobid clean &&
+        echo $jobid >> inactiveids &&
+        flux job id $jobid > failedids &&
+        #
+        #  Run a job that will fail, copy its JOBID to both inactive and
+        #   failed lists.
+        #
+        jobid=`flux mini submit --time-limit=0.25 sleeplong.json` &&
```
grondo, line 70: This isn't going to work :stuck_out_tongue_winking_eye: 

Why not just `flux mini submit --time-limit=0.25 sleep 30`?

t/t2262-job-list-purge.t
```diff
@@ -0,0 +1,181 @@
...
+        #  Currently all completed ids are also inactive
+        #
+        tac completedids | flux job id > completed.ids &&
+        cp completedids inactiveids &&
+        #
+        #  Run a job that will fail, copy its JOBID to both inactive and
+        #   failed lists.
+        #
+        jobid=`flux mini submit nosuchcommand` &&
+        fj_wait_event $jobid clean &&
```
grondo, line 63: again, could be `flux mini submit --wait nosuchcommand`

t/t2262-job-list-purge.t
```diff
@@ -0,0 +1,181 @@
...
+        echo $jobid >> inactiveids &&
+        flux job id $jobid >> failedids &&
+        #
+        #  Submit 8 sleep jobs to fill up resources
+        #
+        for i in $(seq 0 7); do \
+                flux job submit sleeplong.json >> runningids; \
+                fj_wait_event $jobid alloc; \
+        done &&
```
grondo, line 80: Is there a bug here? jobid is not set in this loop but each iteration waits for `$jobid`'s alloc event?

Since we know we have a FIFO scheduler, I'd suggest something like: (untested)
```
   flux mini submit --cc=0-7 sleep 300 >runningids &&
   fj_wait_event $(tail -1 runningids) alloc
```

t/t2262-job-list-purge.t
```diff
@@ -0,0 +1,181 @@
...
+        tac runningids | flux job id > running.ids &&
+        #
+        #  Submit a job and cancel it, copy its JOBID to both inactive
+        #    and failed lists
+        #
+        jobid=`flux mini submit canceledjob` &&
```
grondo, line 86: I was at first confused because `canceledjob` is not a command. Why not submit a real command here (maybe give it `--job-name=canceledjob` to aid in debugging)?

grondo (commented):
I ran this locally and found some issues with the added tests that _could_ be contributing to the strange codecov results (not really sure though)

I noted on possible bug and other possible races in the tests. Also, just for completeness I added some suggestion on modernizing the tests.

grondo:
> #3716 and #3717 didn't seem to help w/ the coverage issue here. Perhaps just a bug in codecov :-(

The coverage run in the last build of this PR timed out, so coverage wasn't uploaded. I restarted the build.

t/t2262-job-list-purge.t
chu11: ahhh, this style exists in a number of other test files already, which is why I cut & pasted it over.  Perhaps we should clean up those other test files too as well.

chu11:
re-pushed, rebasing on #3733 work

chu11:
rebased & re-pushed, added commits to do some minor cleanup and to have a `job-list-helper.sh` file with common helper functions.

garlick:
Since we're now tracking the job manager inactive job purge that was added in #4286, should this PR be closed and any ancillary changes that appear here resubmitted separately?  See: 028b71f84b8cf40eece5a22d2b3f60695028d6d8 and 9b8989c633bc364ecbde88ca50ec914386cea75c .  The job manager publishes a `job-purge-inactive` event, and job-list responds to it by cleaning up.

chu11:
@garlick, yup, we should close

Issue #3691, opened 2021-05-29
shell: add option to raise exception on first nonzero task exit


garlick:
Slurm also has an option to kill a job immediately on the first nonzero task exit. It seems like this plugin could be extended to support that feature if necessary.

_Originally posted by @grondo in https://github.com/flux-framework/flux-core/issues/3681#issuecomment-850478774_

PR #3692, opened 2021-05-29
shell: add exit-on-error option

garlick:
This PR adds a shell option `exit-on-error` (to the `doom` plugin), which implements the behavior suggested in #3691.

It also allows the MPI abort tests to work on all the MPIs we are testing in CI, so that test is changed to eliminate the skip logic and set this new shell option.


grondo (approved):
LGTM. Tested and working as advertised!

chu11 (approved):
LGTM too

garlick:
OK then I'll set MWP.  Thanks!

PR #3695, opened 2021-06-02
libflux: test: fix segfault of test_plugin.t under rpmbuild

grondo:
This PR fixes a unit test failure of `src/common/libflux/test_plugin.t` under `rpmbuild` on RHEL8.

The problem was only discovered under `rpmbuild` since rpm adds `'-Wl,-z,relro -Wl,-z,now'` to `LDFLAGS`, causing a test plugin to fail to `dlopen` due to missing symbols that were ignored during normal circumstances.

Addition of `libccan.la` to `test_ldflags` fixes the underlying problem. The test was also enhanced to better report this particular failure (rather than going on to segfault). Finally, the `LDFLAGS` mentioned above are added to the `centos8` build to catch problems like this a little earlier.


chu11 (approved):
LGTM!

grondo:
Thanks @chu11! Setting MWP

PR #3696, opened 2021-06-03
job-manager: support simple "after*" job dependencies

grondo:
This is  WIP PR which adds support for simple `after{,any,ok,notok}` dependency schemes.

The implementation roughly mimics that of Slurm's `--dependency=after*:JOBID` feature. I'm not sure Slurm's interface is the greatest, but it didn't seem like a good idea to arbitrarily change it, at least at first.

This is posted as WIP before it is fully complete (mostly need testing) to ensure that the approach and design of the feature are on the right track (before spending a bunch of time testing). Especially @SteVwonder and @ryanday36 may want to give their approval of the interface.

One difference between this and Slurm's `--dependency=after*` feature is that, in the current implementation, jobs for which a dependency could never be satisfied, e.g. `afterok` is used and the requisite job fails, have a fatal exception raised. In Slurm these jobs stay in the queue until they are canceled or updated. (since we can't update jobs yet, there seemed no reason to keep the job queued)

Part of the reason this dependency plugin was implemented was to ensure that the jobtap interface had enough support for something like this. Indeed, several bugs and required features were found, fixed and added here, including:

 * Addition of `flux_jobtap_job_lookup()` to lookup any currently active job in the job-manager. The result is returned as a `flux_plugin_arg_t *` to reduce any code duplication, however that choice could be controversial...
 *  Addition of `flux_jobtap_get_job_result()` to return the `flux_job_result_t` of a job in the `job.state.{CLEANUP,INACTIVE}` callback contexts. Providing this function required that the job-manager store the `end_event` for jobs whether they are waitable or not. Since the event adds only a small amount of extra memory to non-waitable jobs, and only in the final two job states, this will probably have zero impact. However @garlick should probably comment.
 *  Replace `jobtap->current_job` with a stack of "current jobs" to make the jobtap interface reentrant. This fixes a bug when jobtap callbacks end up calling callbacks of other jobs (e.g. when releasing a dependency which causes a state transition)
 * Ensure that all cached dependency-remove events are emitted _after_ all dependency-add events, otherwise a job can be prematurely released from the DEPEND state.

Finally, some examples:

Job runs after another job successfully completes:
```
$ flux mini run -vvv --dependency=afterok:$(flux mini submit sleep 2) hostname
jobid: ƒBpaaPscj
0.000s: job.submit {"userid":1000,"urgency":16,"flags":0}
0.013s: job.dependency-add {"description":"after-success=ƒBpUKAuaF"}
1.844s: job.dependency-remove {"description":"after-success=ƒBpUKAuaF"}
1.844s: job.depend
1.844s: job.priority {"priority":16}
1.845s: job.alloc {"annotations":{"sched":{"resource_summary":"rank0/core0"}}}
1.850s: job.start
[snip]
```

Job exception raised if dependency can't be satisfied
```
$ flux mini run -vvv --dependency=afternotok:$(flux mini submit sleep 2) hostname
jobid: ƒC9C3kwd9
0.000s: job.submit {"userid":1000,"urgency":16,"flags":0}
0.013s: job.dependency-add {"description":"after-failure=ƒC961ss7m"}
1.853s: job.exception type=dependency severity=0 dependency after-failure=ƒC961ss7m can never be satisfied
1.853s: job.clean
```

job with multiple dependencies
```
$ flux mini run -vvv --dependency=after:$(flux mini submit sleep 1) --dependency=afterany:$(flux mini submit sleep 3) --dependency=afterok:$(flux mini submit sleep 1) --dependency=afternotok:$(flux mini submit -t 5s sleep 10) hostname
jobid: ƒCPtQi2mm
0.000s: job.submit {"userid":1000,"urgency":16,"flags":0}
0.013s: job.dependency-add {"description":"after-start=ƒCPTe82hd"}
0.013s: job.dependency-add {"description":"after-finish=ƒCPZg17D1"}
0.013s: job.dependency-add {"description":"after-success=ƒCPfkrAH5"}
0.013s: job.dependency-add {"description":"after-failure=ƒCPn6X6Ab"}
0.013s: job.dependency-remove {"description":"after-start=ƒCPTe82hd"}
0.585s: job.dependency-remove {"description":"after-success=ƒCPfkrAH5"}
2.349s: job.dependency-remove {"description":"after-finish=ƒCPZg17D1"}
4.779s: job.dependency-remove {"description":"after-failure=ƒCPn6X6Ab"}
4.779s: job.depend
4.779s: job.priority {"priority":16}
4.784s: job.alloc {"annotations":{"sched":{"resource_summary":"rank0/core0"}}}
4.789s: job.start
[snip]
```

### DIFF
--- a/src/cmd/flux-mini.py
+++ b/src/cmd/flux-mini.py
@@ -225,6 +225,7 @@ class Xcmd:
         "bcc": "--bcc=",
         "log": "--log=",
         "log_stderr": "--log-stderr=",
+        "dependency": "--dependency=",
     }
 
     class Xinput:
--- a/src/modules/job-manager/plugins/dependency-after.c
+++ b/src/modules/job-manager/plugins/dependency-after.c
@@ -1,0 +1,564 @@
+/************************************************************\
+ * Copyright 2020 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* dependency-simple.c - don't start a job until after another starts,
+ *   completes, or fails.
+ */
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+#include <jansson.h>
+#include <czmq.h>
+#include <flux/core.h>
+#include <flux/jobtap.h>
+
+#include "src/common/libutil/iterators.h"
+
+/* Types of "after*" dependencies:
+ */
+enum after_type {
+    AFTER_START =   0x1,
+    AFTER_FINISH =  0x2,
+    AFTER_SUCCESS = 0x4,
+    AFTER_FAILURE = 0x8
+};
+
+struct after_info {
+    enum after_type type;
+    flux_jobid_t depid;
+    char *description;
+};
+
+/*  Reference to an after_info object on another job's dependency list
+ */
+struct after_ref {
+    flux_jobid_t id;
+    zlistx_t *list;
+    struct after_info *info;
+};
+
+static const char * after_typestr (enum after_type type)
+{
+    switch (type) {
+        case AFTER_START:
+            return "after-start";
+        case AFTER_FINISH:
+            return "after-finish";
+        case AFTER_SUCCESS:
+            return "after-success";
+        case AFTER_FAILURE:
+            return "after-failure";
+    }
+    return "";
+}
+
+static int after_type_parse (const char *s, enum after_type *tp)
+{
+    if (strcmp (s, "after") == 0)
+        *tp = AFTER_START;
+    else if (strcmp (s, "afterany") == 0)
+        *tp = AFTER_FINISH;
+    else if (strcmp (s, "afterok") == 0)
+        *tp = AFTER_SUCCESS;
+    else if (strcmp (s, "afternotok") == 0)
+        *tp = AFTER_FAILURE;
+    else
+        return -1;
+    return 0;
+}
+
+static void after_info_destroy (struct after_info *after)
+{
+    if (after) {
+        free (after->description);
+        free (after);
+    }
+}
+
+/*  zlistx_destructor_fn for after_info objects
+ */
+static void after_info_destructor (void **item)
+{
+    if (*item) {
+        after_info_destroy (*item);
+        *item = NULL;
+    }
+}
+
+static struct after_info *after_info_create (flux_jobid_t id,
+                                             enum after_type type,
+                                             const char *desc)
+{
+    struct after_info *after = calloc (1, sizeof (*after));
+    if (!after
+        || asprintf (&after->description,
+                     "%s=%s",
+                     after_typestr (type),
+                     desc) < 0)
+        goto error;
+    after->type = type;
+    after->depid = id;
+    return after;
+error:
+    after_info_destroy (after);
+    return NULL;
+}
+
+static void after_ref_destroy (struct after_ref *ref)
+{
+    free (ref);
+}
+
+/*  zlistx_destructor_fn for after_ref objects
+ */
+static void after_ref_destructor (void **item)
+{
+    if (*item) {
+        after_ref_destroy (*item);
+        *item = NULL;
+    }
+}
+
+static struct after_ref * after_ref_create (flux_jobid_t id,
+                                            zlistx_t *l,
+                                            struct after_info *after)
+{
+    struct after_ref *ref = calloc (1, sizeof (*ref));
+    if (!ref)
+        return NULL;
+    ref->id = id;
+    ref->list = l;
+    ref->info = after;
+    return ref;
+}
+
+/*  flux_free_f destructor for zlistx_t
+ */
+static void list_destructor (void *arg)
+{
+    zlistx_t *l = arg;
+    zlistx_destroy (&l);
+}
+
+/*  Get or create a list embedded in jobid id
+ */
+static zlistx_t * embedded_list_get (flux_plugin_t *p,
+                                     flux_jobid_t id,
+                                     const char *name,
+                                     zlistx_destructor_fn destructor)
+{
+    zlistx_t *l = flux_jobtap_job_aux_get (p, id, name);
+    if (!l) {
+        if (!(l = zlistx_new ())) {
+            errno = ENOMEM;
+            return NULL;
+        }
+        if (flux_jobtap_job_aux_set (p, id, name, l, list_destructor) < 0) {
+            zlistx_destroy (&l);
+            errno = ENOENT;
+            return NULL;
+        }
+        if (destructor)
+            zlistx_set_destructor (l, destructor);
+    }
+    return l;
+}
+
+static zlistx_t * after_list_get (flux_plugin_t *p, flux_jobid_t id)
+{
+    return embedded_list_get (p, id,
+                              "flux::after_list",
+                              (zlistx_destructor_fn *) after_info_destructor);
+}
+
+static zlistx_t * after_refs_get (flux_plugin_t *p, flux_jobid_t id)
+{
+    return embedded_list_get (p, id,
+                              "flux::after_refs",
+                              (zlistx_destructor_fn *) after_ref_destructor);
+}
+
+static zlistx_t *after_list_check (flux_plugin_t *p, flux_jobid_t id)
+{
+    return flux_jobtap_job_aux_get (p,
+                                    id > 0 ? id : FLUX_JOBTAP_CURRENT_JOB,
+                                    "flux::after_list");
+}
+
+static zlistx_t *after_refs_check (flux_plugin_t *p)
+{
+    return flux_jobtap_job_aux_get (p,
+                                    FLUX_JOBTAP_CURRENT_JOB,
+                                    "flux::after_refs");
+}
+
+/*  Lookup a job and return its userid and state information.
+ *
+ *   `id` may  be FLUX_JOBTAP_CURRENT_JOB to return information for
+ *   the current jobtap jobid.
+ */
+static int lookup_job_uid_state (flux_plugin_t *p,
+                                 flux_jobid_t id,
+                                 uint32_t *puid,
+                                 flux_job_state_t *pstate)
+{
+    int rc = 0;
+    flux_plugin_arg_t *args = flux_jobtap_job_lookup (p, id);
+    if (!args
+        || flux_plugin_arg_unpack (args, FLUX_PLUGIN_ARG_IN,
+                                   "{s:i s:i}",
+                                   "userid", puid,
+                                   "state", pstate) < 0)
+        rc = -1;
+    flux_plugin_arg_destroy (args);
+    return rc;
+}
+
+/*  Handler for job.dependency.after*
+ *
+ */
+static int dependency_after_cb (flux_plugin_t *p,
+                                const char *topic,
+                                flux_plugin_arg_t *args,
+                                void *data)
+{
+    const char *scheme = NULL;
+    const char *jobid = NULL;
+    enum after_type type;
+    flux_jobid_t afterid;
+    flux_jobid_t id;
+    uint32_t uid;
+    uint32_t target_uid;
+    flux_job_state_t target_state;
+    struct after_info *after;
+    struct after_ref *ref;
+    zlistx_t *l;
+
+    if (flux_plugin_arg_unpack (args,
+                                FLUX_PLUGIN_ARG_IN,
+                                "{s:I s:i s:{s:s s:s}}",
+                                "id", &id,
+                                "userid", &uid,
+                                "dependency",
+                                "scheme", &scheme,
+                                "value", &jobid) < 0)
+        return flux_jobtap_reject_job (p, args,
+                                       "dependency: after: %s",
+                                       flux_plugin_arg_strerror (args));
+
+    /*  Parse the type of dependency being requested from the scheme:
+     */
+    if (after_type_parse (scheme, &type) < 0)
+        return flux_jobtap_reject_job (p,
+                                       args,
+                                       "invalid dependency scheme: %s",
+                                       scheme);
+
+    /*  Parse the value argument, which must be a valid jobid
+     *  Do not allow FLUX_JOBID_ANY/FLUX_JOBTAP_CURRENT_JOBID to be specified
+     */
+    if (flux_job_id_parse (jobid, &afterid) < 0
+        || afterid == FLUX_JOBTAP_CURRENT_JOB)
+        return flux_jobtap_reject_job (p, args,
+                                       "%s: %s: \"%s\" is not a valid jobid",
+                                       "dependency",
+                                       scheme,
+                                       jobid);
+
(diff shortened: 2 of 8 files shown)


garlick:
> Since the event adds only a small amount of extra memory to non-waitable jobs, and only in the final two job states, this will probably have zero impact. However @garlick should probably comment.

That sounds completely OK to me. 

The full eventlog looks like it is typically under 1K (serialized) so maybe its worth keeping in back of mind that caching the whole thing for active jobs would be low overhead, should a need arise.  A "need" might just be to provide general access to jobtap plugins...

ryanday36:
> 
> One difference between this and Slurm's `--dependency=after*` feature is that, in the current implementation, jobs for which a dependency could never be satisfied, e.g. `afterok` is used and the requisite job fails, have a fatal exception raised. In Slurm these jobs stay in the queue until they are canceled or updated. (since we can't update jobs yet, there seemed no reason to keep the job queued)

We have 'kill_invalid_depend' set on our Slurm clusters, so having jobs with unsatisfiable dependencies die is expected behavior for LC users at least.

Does the plugin support something like `--dependency=JOBID`, and if so, what is the default condition? (I'd argue for `afterany`. The default in LSF is their equivalent of `afterok`, and that has caused confusion for users since jobs that finish due to hitting the time limit are not considered 'ok'.)

grondo:
> Does the plugin support something like `--dependency=JOBID`

No, we can't support that directly with `--dependency` since it takes an [RFC 26](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_26.html) dependency URI of the form `scheme:value[?options]`. So users will have to be explicit.

ryanday36:
> So users will have to be explicit.

That's good too. It never hurts for people to know what they're doing.

grondo:
FYI - I've split the underlying jobtap and `flux-mini` changes into a separate PR #3698.

grondo:
Rebased this on current master, added a simple sharness test to verify operation, and updated the `flux-mini(1)` manpage.

SteVwonder (commented):
The interface LGTM!  Thanks @grondo!

src/cmd/flux-mini.py
```diff
@@ -225,6 +225,7 @@ class Xcmd:
         "bcc": "--bcc=",
         "log": "--log=",
         "log_stderr": "--log-stderr=",
+        "dependency": "--dependency=",
```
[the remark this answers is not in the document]
SteVwonder: I just realized that this could also be handled with the OpenMP-style schema without a special variable by using the `--define` argument of bulksubmit:
```
seq 1 5 | flux mini bulksubmit --define=prev="int(x)-1" --dependency=string:{}?type=out --dependency=string:{.prev}?type=in hostname
```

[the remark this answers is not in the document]
grondo: Unfortunately, I think the OpenMP-style would have to be used for this case since jobids are not known at the time of submission in bulksubmit/submit `--cc`.

grondo:
It should be noted that this implementation is not a drop-in replacement for the Slurm dependencies of the same name. Slurm features missing from this simpler implementation include:

 * Slurm's `after` syntax supports a `+time` specifier to release the dependency `time` minutes after the job starts.
 * Slurm allows a colon-separated list of jobs to be specified after  `after*` dependencies, e.g. `after:id1:id2:id3`. With the implementation in this PR, multiple jobids currently would have to be specified via separate `--dependency=after*` options.
 * In Slurm, if JOBIDs are separated by `?` instead of `:` then _any_ job can satisfy the dependency instead of all jobs

Any or all of these features _could_ be implemented by the `dependency-after` plugin. However, it might be better to push some of the more advanced features off to the flux-depend module (especially the last bullet above).

I'm not sure if any of these features would be considered critical for an initial delivery system instance. Perhaps @ryanday36 could comment.

Oh, also as noted offline by @garlick, Slurm supports a `--dpendency=singleton` option, which prevents multiple jobs with the same user and job name to run at the same time. It would be trivial to implement a `dependency-singleton` plugin (though I think `--dependency=singleton:NAME` is a better user interface, as it allows multiple single jobs to have different job names). However, the OpenMP style dependencies already handle this case, so perhaps we do not need to go there.

ryanday36:
> I'm not sure if any of these features would be considered critical for an initial delivery system instance. Perhaps @ryanday36 could comment.

I think that what you have is definitely good for an initial delivery system instance. I believe that the 'after*' style dependencies are what the vast majority of users who use dependencies are looking for.

grondo:
Thanks @ryanday36!

garlick:
Just to double check, this scenario is intentional, correct?
```
$ flux mini submit hostname
ƒ4ktvxnCf
$ flux mini submit --dependency afterok:ƒ4ktvxnCf hostname
[Errno 22] dependency: afterok: id ƒ4ktvxnCf: job not found
```
The first job completes before the second job that depends on it completing can be submitted.

grondo:
> The first job completes before the second job that depends on it completing can be submitted.

Yes, this only works on _active_ jobs, which keeps the implementation simple. That should probably be explicitly noted in the documentation.

grondo:
> Yes, this only works on active jobs

This is another overlooked feature discrepancy with Slurm, which caches completed jobs and their status for `MinJobAge` seconds (300 by default)

The dependency plugin _could_ be extended to keep a similar cache of inactive jobids and their "result" if the current behavior is a showstopper.

ryanday36:
> This is another overlooked feature discrepancy with Slurm, which caches completed jobs and their status for MinJobAge seconds (300 by default)
> 
> The dependency plugin could be extended to keep a similar cache of inactive jobids and their "result" if the current behavior is a showstopper.

I don't think it's a showstopper. It would be good to get in eventually, but it's not necessarily needed for the initial delivery system instance.

grondo:
hm. I just tested and Slurm `--dependency=afterok:JOBID`  runs the job immediately even if JOBID was purged from the slurmctld's internal list of jobs, and in fact, even if JOBID refers to a JOBID that doesn't refer to job that was ever run.

e.g.:
```console
$ squeue -t completed
             JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
                84    pbatch hostname   grondo CD       0:01      1 fluke62
                85    pbatch hostname   grondo CD       0:00      1 fluke62
                87    pbatch hostname   grondo CD       0:01      1 fluke62
                88    pbatch hostname   grondo CD       0:01      1 fluke62
$ srun --dependency=afterok:1234567890 hostname
fluke62

```

ryanday36:
Well, that's... interesting. It looks like Slurm just silently drops dependencies on jobs that it doesn't know about. That doesn't seem like great behavior, but this is the first time I've heard about it, so maybe it just doesn't come up much. I might poke around in schedmd's bugzilla and see if anyone has complained about it to them.

grondo:
Thanks @ryanday36!

I'm going to update the documentation in this PR to indicate that, currently at least, Flux's `after*` dependency scheme does not support dependencies on inactive jobs. We'll get an issue open on supporting that eventually (within some potential time limit perhaps).

src/test/generate-matrix.py
```diff
@@ -169,7 +169,12 @@ def __str__(self):
 )
 
 # Ubuntu: coverage
-matrix.add_build(name="coverage", coverage=True, jobs=2, args="--enable-pmix-bootstrap")
+matrix.add_build(
+    name="coverage",
+    coverage=True,
+    jobs=2,
+    args="--with-flux-security --enable-pmix-bootstrap",
+)
```
garlick, line 177: Oh duh, sorry I didn't realize I was doing that!

src/modules/job-manager/plugins/dependency-after.c
```diff
@@ -0,0 +1,563 @@
...
+{
+    flux_plugin_arg_t *args = flux_jobtap_job_lookup (p, id);
+    if (!args
+        || flux_plugin_arg_unpack (args, FLUX_PLUGIN_ARG_IN,
+                                   "{s:i s:i}",
+                                   "userid", puid,
+                                   "state", pstate) < 0)
+        return -1;
+    flux_plugin_arg_destroy (args);
+    return 0;
```
garlick, line 221: args leaked if unpack fails

src/modules/job-manager/plugins/dependency-after.c
```diff
@@ -0,0 +1,563 @@
...
+            flux_log_error (flux_jobtap_get_flux (p),
+                            "flux_jobtap_dependency_remove");
+        after_info_destroy (after);
+        return 0;
+    }
+
+    /*  Append this dependency to the deplist in the target jobid:
+     */
+    if (!(l = after_list_get (p, afterid))
+        || !(handle = zlistx_add_end (l, after))) {
```
garlick, line 327: `handle` assigned but not used

src/modules/job-manager/plugins/dependency-after.c
```diff
@@ -0,0 +1,563 @@
...
+        if (flux_jobtap_raise_exception (p,
+                                         after->depid,
+                                         "dependency",
+                                         0,
+                                         "Failed to remove dependency %s",
+                                         after->description) == 0) {
+            flux_log_error (flux_jobtap_get_flux (p),
+                            "flux_jobtap_raise_exception: id=%ju",
+                            (uintmax_t) after->depid);
+        }
```
garlick, line 367: Use of `flux_log_error()` when errno not defined?

src/modules/job-manager/plugins/dependency-after.c
```diff
@@ -0,0 +1,563 @@
...
+            if (after->type & typemask) {
+                /*  Remove dependency (possibly moving dependent job
+                 *   out of the DEPEND state.
+                 */
+                remove_jobid_dependency (p, after);
+
+                /*  Delete this entry since it has been resolved.
+                 */
+                if (zlistx_delete (l, zlistx_cursor (l)) < 0)
+                    flux_log_error (flux_jobtap_get_flux (p),
```
garlick, line 389: errno is not set by `zlistx_delete()` so use `flux_log (h, LOG_ERR, ...)`?

src/modules/job-manager/plugins/dependency-after.c
```diff
@@ -0,0 +1,563 @@
...
+
+    if (l == NULL)
+        return 0;
+
+    if (flux_jobtap_get_job_result (p,
+                                    FLUX_JOBTAP_CURRENT_JOB,
+                                    &result) < 0) {
+        flux_log_error (h, "dependency-after: flux_jobtap_get_result");
+        return -1;
+    }
```
garlick, line 470: Not a comment on this code, but in `flux_jobtap_get_job_result()`, errno is not set on `json_unpack_ex()` error, but it appears from this usage and from other error paths that it should be.

src/modules/job-manager/plugins/dependency-after.c
```diff
@@ -0,0 +1,563 @@
...
+                       void *data)
+{
+    zlistx_t *l = after_refs_check (p);
+    if (l) {
+        /*  Job has proceeded out of DEPEND state.
+         *  Delete any dependency reference list since it will no longer
+         *  be used, and we don't need to attempt deref of dependencies
+         *  (see bottom of inactive_cb()).
+         */
+        if (flux_jobtap_job_aux_delete (p, FLUX_JOBTAP_CURRENT_JOB, l) < 0) {
```
garlick, line 505: Would it be slightly more clear here to delete using the key `flux::after_refs`? (I guess that would be aux_set with value of NULL).

I just got slightly confused wondering why this way, when the value was stored under a key.  Definitely not a big deal.

garlick (approved):
LGTM!  Nice addition.

I just had a few nitpicky comments that aren't all that important.  Feel free to set MWP when you're ready.

src/test/generate-matrix.py
grondo: No, it is a problem with the way the script works (took me a second to figure out)

src/modules/job-manager/plugins/dependency-after.c
grondo: Ah, good catch. In previous versions handle was used in the reference but then I remembered use of zlistx handles is not safe since they are re-assigned on a sort.

src/modules/job-manager/plugins/dependency-after.c
grondo: Yes, should be. thanks.

src/modules/job-manager/plugins/dependency-after.c
grondo: Hm, I guess `delete` is clearer to me than `aux_set(NULL)`, however I can make this change.

src/modules/job-manager/plugins/dependency-after.c
grondo: errno should be set by `flux_jobtap_raise_exception()`, is there some path where you see it is not?

src/modules/job-manager/plugins/dependency-after.c
garlick: Yeah, either way is fine by me, actually.  I probably shouldn't have commented.

src/modules/job-manager/plugins/dependency-after.c
garlick: Oh maybe the test for error is wrong then. `flux_jobtap_raise_exception()` returns -1 on error.

grondo: Oh, strange, I didn't even see that! I don't know why that `== 0` is there. Thanks for catching it.

src/modules/job-manager/plugins/dependency-after.c
grondo: the `aux_delete` call also has less useless parameters than `aux_set` with NULL. Maybe I could add a convenience macro in this case if deletion by key is more efficient, e.g.:
```
#define flux_jobtap_job_aux_destroy(p, id, key) flux_jobtap_job_aux_set (p, id, key, NULL, NULL);
```

garlick: I think it's no more efficient - I just thought if we're naming things, it's clearer to use the names, but now I think it was fine as is and we should leave it.

grondo:
Ok, I've addressed @garlick's comments, squashed and pushed the result.

src/test/generate-matrix.py
```diff
@@ -169,7 +169,12 @@ def __str__(self):
 )
 
 # Ubuntu: coverage
-matrix.add_build(name="coverage", coverage=True, jobs=2, args="--enable-pmix-bootstrap")
+matrix.add_build(
```
garlick, line 172: > ci: add --with-flux-seucrity to coverage build 

just noticed - security is misspelled in the commit title

grondo:
The codecov uploader seems to be having problems in GitHub actions today. I may try switching to the official github action instead of the bash uploader (though I think they should be equivalent).

Should the ci updates be placed in a separate PR?

garlick:
> Should the ci updates be placed in a separate PR?

Yeah that sounds right to me.  Then we'll have that in the release notes.

grondo:
Ok, the CI specific commits have been moved over to #3703

garlick:
Thoughts on merging this?  I think it ought to go in - full speed ahead!

grondo:
I had to restart the build because the test-install builder timed out.

grondo:
Ok, all looks good now, I'll set MWP.

Issue #3617, opened 2021-04-27
libflux-core.so depends on libzmq.so


garlick:
Problem:  As noted in #3611, the `flux-shell` takes a long time to start up on a FIPS compliant system due to crypto self-tests in `libcrypto.so` (openssl).   This seems to be due to this chain of DSO dependencies: `libflux-core.so` -> `libzmq.so` -> `libgssapi_krb5.so` -> `libcrypto.so`.

Since  0MQ sockets are used only by the broker and the shmem connector, it seems like we should be able to eliminate the 0MQ dependency in `libflux-core.so` by migrating some code to static internal libraries.  Doing so would also benefit other users of our APIs.

Some obvious areas iin `libflux` that would need attention:
* reactor: zmq watcher
* message use of zframe, zsock
* log use of `zmq_strerror()`
* use of czmq containers

grondo:
> Doing so would also benefit other users of our APIs.

It is probably obvious, but this includes users of the Python APIs since `import flux` will presumably also be impacted negatively by the FIPS selfcheck on library initialization.

garlick:
On the use of czmq containers - we could pull in the other container types (zlistx, zlist, zhash) to `libczmqcontainers` and then try to make it work without referencing the externally packaged czmq.

I don't think this is too difficult if we comment out some of the functions that we don't need that require other czmq classes.  In zhashx, that would include `zhashx_test()`, `zhashx_load()`, `zhashx_save()`, `zhashx_refresh()`, `zhashx_pack()`, and `zhashx_comment()`, for example.

A small number of support functions would need to be added to `czmq_internal.h`, and I think we'll need a `czmq_internal.c` for at least one function - `zstr_free()` since its used a default destructor via function pointer in `zhashx_new()`

garlick:
On the message functions: here is a small brain dump.

I think the general approach could be to switch the internal implementation from zframe/zmsg to a native format, expose that format in an internal header, then re-implement the 0MQ send/recv functions in librouter using methods similar to usock/sendfd.

Specifically the functions that would need to be re-implemented/moved are these
```c
/* Send message to zeromq socket.
 * Returns 0 on success, -1 on failure with errno set.
 */
int flux_msg_sendzsock (void *dest, const flux_msg_t *msg);
int flux_msg_sendzsock_ex (void *dest, const flux_msg_t *msg, bool nonblock);

/* Receive a message from zeromq socket.
 * Returns message on success, NULL on failure with errno set.
 */
[...]

PR #3697, opened 2021-06-04
libflux: msg API cleanup

chu11:
Various cleanup while I was working on #3617

### DIFF
--- a/src/common/libflux/message.c
+++ b/src/common/libflux/message.c
@@ -668,6 +668,10 @@ int flux_msg_get_errnum (const flux_msg_t *msg, int *e)
     int type;
     uint32_t xe;
 
+    if (!e) {
+        errno = EINVAL;
+        return -1;
+    }
     if (!zf || proto_get_type (zframe_data (zf), zframe_size (zf), &type) < 0
             || (type != FLUX_MSGTYPE_RESPONSE && type != FLUX_MSGTYPE_KEEPALIVE)
             || proto_get_u32 (zframe_data (zf), zframe_size (zf),
@@ -699,6 +703,10 @@ int flux_msg_get_seq (const flux_msg_t *msg, uint32_t *seq)
     zframe_t *zf = zmsg_last (msg->zmsg);
     int type;
 
+    if (!seq) {
+        errno = EINVAL;
+        return -1;
+    }
     if (!zf || proto_get_type (zframe_data (zf), zframe_size (zf), &type) < 0
             || type != FLUX_MSGTYPE_EVENT
             || proto_get_u32 (zframe_data (zf), zframe_size (zf),
@@ -729,6 +737,10 @@ int flux_msg_get_matchtag (const flux_msg_t *msg, uint32_t *t)
     zframe_t *zf = zmsg_last (msg->zmsg);
     int type;
 
+    if (!t) {
+        errno = EINVAL;
+        return -1;
+    }
     if (!zf || proto_get_type (zframe_data (zf), zframe_size (zf), &type) < 0
             || (type != FLUX_MSGTYPE_REQUEST && type != FLUX_MSGTYPE_RESPONSE)
             || proto_get_u32 (zframe_data (zf), zframe_size (zf),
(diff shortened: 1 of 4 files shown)


src/common/libflux/message.c
```diff
@@ -277,11 +277,15 @@ void *flux_msg_aux_get (const flux_msg_t *msg, const char *name)
     return aux_get (msg->aux, name);
 }
 
-size_t flux_msg_encode_size (const flux_msg_t *msg)
+ssize_t flux_msg_encode_size (const flux_msg_t *msg)
```
garlick, line 280: Probably should update the one user in librouter/sendfd.c to handle the error case that's now possible.

src/common/libflux/message.c
```diff
@@ -1683,11 +1683,6 @@ flux_msg_t *flux_msg_recvzsock (void *sock)
     return msg;
 }
 
-int flux_msg_frames (const flux_msg_t *msg)
-{
-    return zmsg_size (msg->zmsg);
-}
-
```
garlick, line 1639: I think this is here mainly for testing.  By removing the frame count checks from the test, are we diminishing our unit test?  Since we will still need to have a variable length stack of message frames in the rework, does it make sense to either have a function that returns a count of those, or just keep this function?  There will still be the same message frames on the wire in the rework unless we are planning to redo the protocol and RFC 3.

See zmq_send(3) - the terminology is slightly different without czmq.  zmq refers to multi-part messages, i.e. "message parts" instead of "message frames".  So maybe a rename to `flux_msg_part_count()` or similar if we decide to keep this function.

garlick (commented):
It would be good to give each of them a "problem style" commit message.

src/common/libflux/message.c
chu11: We could just keep the function.  I was mostly thinking that counting frames suggests how data is stored internally and since we're planning on changing that, the function would have less purpose going forward.

chu11:
rebased & re-pushed.  updated more comments to have problem/solution format.  checked return code in `sendfd` per comment above.  Removed commit that removed `flux_msg_frames()`.

src/common/libflux/message.c
garlick: I think you make a good case for not exposing this in the public API.  If you plan to have a `message_private.h` in the next PR to privately share the message struct with librouter, maybe this function belongs there under a new name like `msg_multipart_count()` or similar.

src/common/libflux/message.c
```diff
@@ -419,7 +419,7 @@ int flux_msg_get_flags (const flux_msg_t *msg, uint8_t *fl)
 
 int flux_msg_set_private (flux_msg_t *msg)
 {
-    uint8_t flags;
```
garlick, line 420: Just a tiny bit more info in the commit message explaining why this is a problem now, as opposed to before this PR(?) would be helpful.

garlick (approved):
Looks good!  Just one nit on a commit message.  Feel free to fix or not, and force push and set MWP when done.

Thanks for cleaning this up!

src/common/libflux/message.c
chu11: this fix was "preemptive" as it really appeared after some refactoring for #3617.  But I'll add a note of how it is "sensitive" to changes.

garlick: When I'm perusing C code, I usually take a local variable initialization as a subconscious hint to watch out for a code path in which its value is used before set, so when there isn't one, it tends to make my brain do extra work.  For that reason, sometimes I'll clean up a function by _removing_ unnecessary initialization when it's obvious it's not needed.  I'm not sure if that's reasonable, or if I should reprogram myself :-)

garlick:
Did you mean to close this and delete the branch without merging?

chu11:
oh crap!  I was cleaning up my remote branches, guess i deleted the wrong one.

src/common/libflux/message.c
chu11: i'm actually the same way and am not entirely sure why this started popping up recently.  It's this style of code that was hit:

```
int flags;
if (get_flags (ctx, &flags) < 0)
   goto error;
```

i don't know why changes somehow triggered the warning :-(   Perhaps some other code changes led to an extra pass in the compiler and that's all that happened.

chu11:
crap, i guess i can't re-open a PR, I'll have to re-open this.  my bad

PR #3698, opened 2021-06-04
jobtap: fixes and api enhancements to support dependency plugins

grondo:
This PR is pulled from #3696, it felt better to split these more generic jobtap fixes and enhancements from the addition of the `dependency-after` jobtap plugin itself.

This PR includes the following:
  * change `--dependency=scheme:value` in `flux-mini` to always encode `value` as a string per RFC 26
  * fix reentrancy of jobtap->current_job
  * capture job->end_event for all jobs, not just waitable
  * add `flux_jobtap_get_job_result()`
  * add `flux_jobtap_lookup_job()`
 
Along with corresponding testsuite updates.

### DIFF
--- a/src/cmd/flux-mini.py
+++ b/src/cmd/flux-mini.py
@@ -65,7 +65,7 @@ class Dependency:
     def entry(self):
         entry = {
             "scheme": self.scheme,
-            "value": self._try_number(self.path),
+            "value": self.path,
         }
         if self.uri.query:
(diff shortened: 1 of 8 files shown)


garlick (approved):
All looks good to me!

grondo:
Thanks! I found a stray `fprintf()` debugging line, removed it, and force-pushed. I'll set MWP now.

PR #3704, opened 2021-06-08
job-manager: add builtin begin-time dependency plugin

grondo:
This small WIP PR adds a minimum begin-time (or start-after time) via a new `begin-time` dependency scheme.
For this scheme, the `value` is interpreted as floating-point seconds since epoch after which the dependency shall be removed. The plugin uses the convenient flux "periodic" watcher, which watches system time instead of relative time, so it should "just work". If the time for which the dependency is set has already lapsed then the dependency is satisfied immediately.

The next step is to add a `flux-mini` option equivalent to Slurm's `--begin` option, (perhaps named `--begin-time` or `--start-after` to be more explicit), with support for parsing a slew of user friendly datetimes and datetime offsets, e.g.

 * `--begin-time=+FSD`
 * `--begin-time=Tuesday`
 * `--begin-time=2021-06-15`

etc. There are some promising python packages in this area (e.g. [dateutil](https://dateutil.readthedocs.io/en/stable/), [dateparser](https://dateparser.readthedocs.io/en/latest/)) but each has its drawbacks and I'd hate to add a new dependency to support just one option.

For now maybe the best approach would be to support a series of `strptime` formats, and additionally if the argument starts with `+` it would be considered `now+FSD`.

To try out the support presently, you can still use `--dependency=begin-time:TIME`, e.g.:
```
$ flux mini run -vvv --dependency=begin-time:$(($(date +%s)+2)) hostname
jobid: ƒ28x5DMfv7
0.000s: job.submit {"userid":1000,"urgency":16,"flags":0}
0.014s: job.dependency-add {"description":"begin-time=1623194817.000"}
1.188s: job.dependency-remove {"description":"begin-time=1623194817.000"}
1.188s: job.depend
1.188s: job.priority {"priority":16}
1.190s: job.alloc {"annotations":{"sched":{"resource_summary":"rank0/core0"}}}
1.250s: job.start
```

### DIFF
--- a/doc/man1/flux-mini.rst
+++ b/doc/man1/flux-mini.rst
@@ -105,7 +105,6 @@ following additional job parameters:
    ("s", "m", "h", or "d"). If unspecified, the job is subject to the
    system default time limit.
 
-
 STANDARD I/O
 ============
 
@@ -180,9 +179,16 @@ afternotok:JOBID
    This dependency is satisfied after JOBID enters the INACTIVE state
    with an unsuccessful result.
 
-In any of the above cases, if it is determined that the dependency cannot
-be satisfied (e.g. a job fails due to an exception with afterok), then
-a fatal exception of type=dependency is raised on the current job.
+begin-time:TIMESTAMP
+   This dependency is satisfied after TIMESTAMP, which is specified in
+   floating point seconds since the UNIX epoch. See the ``flux-mini``
+   ``--begin-time`` option below for a more user-friendly interface
+   to the ``begin-time`` dependency.
+
+In any of the above ``after*`` cases, if it is determined that the
+dependency cannot be satisfied (e.g. a job fails due to an exception
+with afterok), then a fatal exception of type=dependency is raised
+on the current job.
 
 ENVIRONMENT
 ===========
@@ -341,6 +347,16 @@ OTHER OPTIONS
    is interpreted as a string. If KEY starts with a ``^`` character, then
    VAL is interpreted as a file, which must be valid JSON, to use as the
    attribute value.
+
+**--begin-time=DATETIME**
+   Convenience option for setting a ``begin-time`` dependency for a job.
+   The job is guaranteed to start after the specified date and time.
+   If *DATETIME* begins with a ``+`` character, then the remainder is
+   considered to be an offset in Flux standard duration (RFC 23), otherwise,
+   any datetime expression accepted by the Python 
+   `parsedatetime <https://github.com/bear/parsedatetime>`_ module
+   is accepted, e.g. ``2021-06-21 8am``, ``in an hour``,
+   ``tomorrow morning``, etc.
 
 **--dry-run**
    Don't actually submit job. Just emit jobspec on stdout and exit for
--- a/doc/test/spell.en.pws
+++ b/doc/test/spell.en.pws
@@ -532,3 +532,4 @@ reprioritization
 afterany
 afterok
 afternotok
+parsedatetime
--- a/src/cmd/flux-mini.py
+++ b/src/cmd/flux-mini.py
@@ -85,6 +85,21 @@ def dependency_array_create(uris):
     return dependencies
 
 
+class BeginTimeAction(argparse.Action):
+    """Convenience class to handle --begin-time file option
+
+    Append --begin-time options to the "dependency" list in namespace
+    """
+
+    def __call__(self, parser, namespace, values, option_string=None):
+        uri = "begin-time:" + str(util.parse_datetime(values).timestamp())
+        items = getattr(namespace, "dependency", [])
+        if items is None:
+            items = []
+        items.append(uri)
+        setattr(namespace, "dependency", items)
+
+
 def filter_dict(env, pattern, reverseMatch=True):
     """
     Filter out all keys that match "pattern" from dict 'env'
@@ -450,6 +465,12 @@ class MiniCmd:
             action="append",
             help="Set an RFC 26 dependency URI for this job",
             metavar="URI",
+        )
+        parser.add_argument(
+            "--begin-time",
+            action=BeginTimeAction,
+            metavar="TIME",
+            help="Set minimum begin time for job",
         )
         parser.add_argument(
             "--env",
--- a/src/modules/job-manager/Makefile.am
+++ b/src/modules/job-manager/Makefile.am
@@ -61,7 +61,8 @@ libjob_manager_la_SOURCES = \
 	jobtap.h \
 	jobtap.c \
 	plugins/priority-default.c \
-	plugins/dependency-after.c
+	plugins/dependency-after.c \
+	plugins/begin-time.c
 
 fluxinclude_HEADERS = \
 	jobtap.h
--- a/src/modules/job-manager/jobtap.c
+++ b/src/modules/job-manager/jobtap.c
@@ -39,6 +39,7 @@
 
 extern int priority_default_plugin_init (flux_plugin_t *p);
 extern int after_plugin_init (flux_plugin_t *p);
+extern int begin_time_plugin_init (flux_plugin_t *p);
 
 struct jobtap_builtin {
     const char *name;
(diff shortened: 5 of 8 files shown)


grondo:
Ok, I've pushed an _experimental_ `flux.util.parse_datetime()` function to the Python bindings, along with a new  `--begin-time` option to `flux-mini` which uses it here to see what people think.

I was playing with natural language date parsers and found that [parsedatetime](https://github.com/bear/parsedatetime) is by far the best. It supports various date and time formats like `Jun 11, 8am`, `tomorrow`, `eod`, `in 1.5 hours`, etc. as well as the more formal formats like `2021-06-11 08:00`.

Of course it is not a standard package (though an ubuntu 18.04 package exists at least), and I'd hate to require it just for this one option (so far), so what I've done in this version of `parse_datetime()` is to try to import the module and fall back to other methods if the import fails. Currently, the function does the following when trying to convert the input string to a datetime:

 * If the string begins with `+` or `-` then the  string is considered to be an offset and the remainder is parsed as FSD
 * Else, try to import `parsedatetime` and use its parser
 * Else, try to import `dateutil` and use its parser
 * Finally, try `datetime.fromisoformat()`

As I said, this is just an experiment. I'm not sure it is good Python (or otherwise) practice to have an option behave differently depending on what modules are installed. However, I also don't want to go down the slippery slope of having a project that requires umpteen third party Python packages just to install it, even when you'll never use even 1% of them.

However, in this case we may have other uses for a natural language datetime parser. For instance, for `flux jobs` it might be nice to be able to make datetime queries like `flux jobs --since='last wednesday'`, etc... This is why `parse_datetime()` went into the bindings.

If you want to play with the input strings, I found this helper script useful:
```python
#!/usr/bin/env python3

import sys
from datetime import datetime
from flux.util import parse_datetime

dt1 = datetime.now().astimezone()
print(dt1)

dt2 = parse_datetime(sys.argv[1], now=dt1)
print(dt2)

print("difference", dt2.timestamp() - dt1.timestamp())
```

```console
$ ./dutil.py 'in 3 hours'
2021-06-10 11:17:31.050679-07:00
2021-06-10 14:17:31-07:00
difference 10799.94932103157
$ ./dutil.py 'in 1.25h'
2021-06-10 11:17:39.196635-07:00
2021-06-10 12:32:39-07:00
difference 4499.803364992142
$ ./dutil.py 'Friday morning'
2021-06-10 11:17:55.101016-07:00
2021-06-11 06:00:00-07:00
difference 67324.89898395538
$ ./dutil.py 'Friday eod'
2021-06-10 11:17:59.555792-07:00
2021-06-11 17:00:00-07:00
difference 106920.44420790672
$ ./dutil.py 'next thu'
2021-06-10 11:18:06.668633-07:00
2021-06-17 11:18:06-07:00
difference 604799.3313670158

```

SteVwonder:
> Of course it is not a standard package (though an ubuntu 18.04 package exists at least), and I'd hate to require it just for this one option (so far), so what I've done in this version of parse_datetime() is to try to import the module and fall back to other methods if the import fails
> As I said, this is just an experiment. I'm not sure it is good Python (or otherwise) practice to have an option behave differently depending on what modules are installed. However, I also don't want to go down the slippery slope of having a project that requires umpteen third party Python packages just to install it, even when you'll never use even 1% of them.

Totally valid to not want to add another python dependency. My main concern with changing behavior based on installed modules is documentation and inconsistency across systems.  If we put `flux min run '--dependency=begin-time:after 2hrs' sleep 0` in the docs and someone copies-pastes into their system, it might not work.  It could also hurt the portability of scripts across systems.

Thoughts on vendoring `parsedatetime` in flux-core? It is Apache-2.0, which I believe is compatible with LGPLv3 [[1]](https://en.wikipedia.org/wiki/License_compatibility#/media/File:Floss-license-slide-image.svg).  We could install it under `flux.utils.parsedatetime` (for example).  It also doesn't seem to have a high churn, with the core code not changing in the past 5 years, and only minor updates to various locales more than a year ago.

garlick:
Nice! 

Thoughts on creating a Flux Absolute and Relative Time Specification?

grondo:
> Thoughts on vendoring parsedatetime in flux-core? It is Apache-2.0, which I believe is compatible with LGPLv3 [1]. We could install it under flux.utils.parsedatetime (for example).

I was going to also ask if that would be acceptable. I'm fine with vendoring it, it seems so useful even though we only have the one use case now... One other really nice thing about parsedatetime is that it seems to have good support for localization (though I think another dependency, pyicu, is required for that to work).

garlick:
> One other really nice thing about parsedatetime is that it seems to have good support for localization

But no support for teatime or [fika](https://github.com/SchedMD/slurm/blob/1d9c85b770e0ce1e3a7a5bb0848ca217f84b99a6/NEWS#L6941)?

grondo:
> But no support for teatime or fika?

If we vendor `parsedatetime` then we could add these special times to the `locale/base.py` `re_sources` list:
```python
# special day and/or times, i.e. lunch, noon, evening
# each element in the dictionary is a dictionary that is used
# to fill in any value to be replace - the current date/time will
# already have been populated by the method buildSources
re_sources = {
    'noon': {'hr': 12, 'mn': 0, 'sec': 0},
    'afternoon': {'hr': 13, 'mn': 0, 'sec': 0},
    'lunch': {'hr': 12, 'mn': 0, 'sec': 0},
    'morning': {'hr': 6, 'mn': 0, 'sec': 0},
    'breakfast': {'hr': 8, 'mn': 0, 'sec': 0},
    'dinner': {'hr': 19, 'mn': 0, 'sec': 0},
    'evening': {'hr': 18, 'mn': 0, 'sec': 0},
    'midnight': {'hr': 0, 'mn': 0, 'sec': 0},
    'night': {'hr': 21, 'mn': 0, 'sec': 0},
    'tonight': {'hr': 21, 'mn': 0, 'sec': 0},
    'eod': {'hr': 17, 'mn': 0, 'sec': 0},
}

```

SteVwonder:
> I was going to also ask if that would be acceptable. I'm fine with vendoring it, it seems so useful even though we only have the one use case now... One other really nice thing about parsedatetime is that it seems to have good support for localization (though I think another dependency, pyicu, is required for that to work).

:+1: I'd vote to vendor it.  I really like the `flux jobs --since='last thursday'` example that you gave.  I'm sure once it is easily accessible, more use cases will pop up.

grondo:
> I'd vote to vendor it. I really like the flux jobs --since='last thursday' example that you gave. I'm sure once it is easily accessible, more use cases will pop up.

Great! I went ahead and did that. Now just have to figure out how to get format to ignore files under `src/bindings/python/flux/utils`.

grondo:
I've pushed an update to rebase on current master to pick up our vendored `parsedatetime` dependency, added tests and simple docs, and finally removed WIP since this is ready for a more serious review.

grondo:
It will be difficult to bump coverage since there is a frustrating amount of error handling in the otherwise short begin-time plugin.

doc/man1/flux-mini.rst
```diff
@@ -342,6 +348,16 @@ OTHER OPTIONS
    VAL is interpreted as a file, which must be valid JSON, to use as the
    attribute value.
 
+**--begin-time=DATETIME**
+   Convenience option for setting a ``begin-time`` dependency for a job.
+   The job is guaranteed to start after the specified date and time.
+   if *DATETIME* begins with a ``+`` character, then the remainder is
```
garlick, line 354: "if" should be capitalized.

garlick (approved):
LGTM!  Just one minor man page nit.

This should check off an item on https://github.com/orgs/flux-framework/projects/27

Hmm, what's referenced there is discussion #3243, so I guess that should be capped off somehow.

grondo:
I fixed that one manpage issue and rebased on current master. Thanks @garlick!

grondo:
> Hmm, what's referenced there is discussion #3243, so I guess that should be capped off somehow.

I'll reference this PR in that discussion thread and mark it as an "answer". Then move that card to the "done" column.

garlick:
Great - feel free to set MWP when you're ready!

grondo:
Thanks!

PR #3706, opened 2021-06-09
optparse: make optional_argument apply to long options only

grondo:
As discussed in #3700, it is a bummer that `.has_arg = 2` (optional_argument) doesn't work well for `--verbose`. Then you can't use the idiom of `-vvv` for verbosity level 3.  But if you don't allow an argument, users that want to use long options have to use `--verbose --verbose --verbose`.

It is rare that optional arguments are useful with short options, so this PR makes optional args apply to long options only by default. The result is that `-vvv` works as does `--verbose=3`. A flag is added to restore the old behavior.

As part of this work it seemed useful to have `optparse_get_int()` return the number of times an option was used on the command line instead of a fatal error if it is called on an option with no argument, so that change was made here as well.

### DIFF
--- a/src/common/liboptparse/optparse.c
+++ b/src/common/liboptparse/optparse.c
@@ -1216,6 +1216,8 @@ static char * optstring_append (char *optstring, struct optparse_option *o)
      *   an option with no argument (has_arg = 0), 2 characters
      *   for a required argument (has_arg = 1), and 3 chars for
      *   an optional argument "o::" (has_arg = 2)
+     *  N.B.: optional argument is applied to short options only if
+     *   OPTPARSE_OPT_SHORTOPT_OPTIONAL_ARG is set on flags
      */
     len = strlen (optstring);
     n = len + o->has_arg + 1;
@@ -1224,7 +1226,7 @@ static char * optstring_append (char *optstring, struct optparse_option *o)
 
     if (o->has_arg == 1)
         colons = ":";
-    else if (o->has_arg == 2)
+    else if (o->has_arg == 2 && o->flags & OPTPARSE_OPT_SHORTOPT_OPTIONAL_ARG)
         colons = "::";
 
     sprintf (optstring+len, "%c%s", o->key, colons);
--- a/src/common/liboptparse/optparse.c
+++ b/src/common/liboptparse/optparse.c
@@ -880,7 +880,9 @@ int optparse_get_int (optparse_t *p, const char *name, int default_value)
     }
     if (n == 0)
         return default_value;
-    if (s == NULL || strlen (s) == 0)
+    if (s == NULL)
+        return n;
+    if (strlen (s) == 0)
         goto badarg;
     errno = 0;
     l = strtol (s, &endptr, 10);
--- a/src/common/liboptparse/optparse.h
+++ b/src/common/liboptparse/optparse.h
@@ -130,6 +130,12 @@ struct optparse_subcommand {
  */
 #define OPTPARSE_OPT_HIDDEN       0x2
 
+/*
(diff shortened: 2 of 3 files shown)


src/common/liboptparse/optparse.h
```diff
@@ -358,8 +358,10 @@ bool optparse_hasopt (optparse_t *p, const char *name);
 
 /*
  *   Return the option argument as an integer if 'name' was used,
- *    'default_value' if not.  If the option is unknown, or the argument
- *    could not be converted to an integer, call the fatal error function.
+ *    'default_value' if not.  If the option does not take an argument,
+ *    then retuns the number of times the option was used.
```
garlick, line 368: misspelled returns

src/common/liboptparse/optparse.c
```diff
@@ -1216,6 +1216,8 @@ static char * optstring_append (char *optstring, struct optparse_option *o)
      *   an option with no argument (has_arg = 0), 2 characters
      *   for a required argument (has_arg = 1), and 3 chars for
      *   an optional argument "o::" (has_arg = 2)
+     *  N.B.: optional argument on short option only with
+     *   OPTPARSE_OPT_OPTIONAL_ARG_SHORTOPT
```
garlick, line 1220: s/only/also/ ?

src/common/liboptparse/optparse.h
```diff
@@ -130,6 +130,12 @@ struct optparse_subcommand {
  */
 #define OPTPARSE_OPT_HIDDEN       0x2
 
+/*
+ *  Apply optional argument to short options as well as long. The default
+ *   is that optional arguments only apply to longopts.
+ */
+#define OPTPARSE_OPT_SHORTOPT_OPTIONAL_ARG 0x4
```
garlick, line 137: Nice flag name :-)

Making fun of names is easy, thinking of a better one...more difficult.  I've got nothing.

garlick (approved):
Nice!  Thank you for working this up so quickly, as it lets me minimize changes to #3700.

src/common/liboptparse/optparse.c
grondo: yeah, that is ambiguous, the comment is meant to say the optional argument will only be applied if the flag is set, I'll reword. (Also the name of the flag is wrong :facepalm: )

src/common/liboptparse/optparse.h
grondo: Yeah, the optional short option opt is too long :slightly_frowning_face:

grondo:
Ok, I addressed the comments (I hope), squashed the result and forced a push. Sorry about the lengthy flag name. :frowning:

src/common/liboptparse/optparse.c
```diff
@@ -1216,6 +1216,8 @@ static char * optstring_append (char *optstring, struct optparse_option *o)
      *   an option with no argument (has_arg = 0), 2 characters
      *   for a required argument (has_arg = 1), and 3 chars for
      *   an optional argument "o::" (has_arg = 2)
+     *  N.B.: optional argument is applied to short options only iff
```
garlick, line 1219: Sorry, did you mean "only iff" or just "iff"?

grondo: blah! typo. "iff" would work, but "only if" is probably more clear. Wil fix.

grondo: Thanks for catching that, force pushed a fix.

garlick: Thanks!  Feel free to set MWP!

grondo:
Thanks!

PR #3711, opened 2021-06-11
python: add flux.util.parse_datetime

grondo:
This was split off #3704.

This PR vendors the Python [parsedatetime](https://github.com/bear/parsedatetime) package into `flux.utils.parsedatetime` and adds a `flux.util.parse_datetime()` function which converts any supported datetime-like string to a timestamp.

`parse_datetime()` will treat the string as an offset in FSD if it begins with `+` or `-` as a special case. This is because `parsedatetime` ignores microseconds by default and it may be useful to have a way to add less than a second to a timestamp.

If not `+/-FSD`, the string is passed to `parsedatetime`.


src/bindings/python/flux/util.py
```diff
@@ -230,6 +231,49 @@ def parse_fsd(fsd_string):
...
 
 
+def parse_datetime(string, now=None):
+    """Parse a possibly human readable datetime string or offset
+
+    If string starts with `+` or `-`, then the remainder of the string
+    is assumed to be a duration to add or subtract to `now` in Flux Standard
+    Duration.
+
+    Otherwise, the parsedatetime package willbe used to parse the input
```
SteVwonder, line 241:
```suggestion
    Otherwise, the parsedatetime package will be used to parse the input
```

src/bindings/python/flux/util.py
```diff
@@ -230,6 +231,49 @@ def parse_fsd(fsd_string):
...
+    string. If parsedatetime cannot be found, then the dateutil package
+    will be used. If both these packages cannot be found, then `string`
+    will be parsed with datetime.fromisoformat().
```
SteVwonder, line 244:
```suggestion
    string.
```

src/test/checks_run.sh
```diff
@@ -109,6 +109,8 @@ if test "$COVERAGE" = "t"; then
 	include = $(pwd)/src/*
 	parallel = True
 	relative_files = True
+        [report]
+        omit = */flux/utils/*
```
SteVwonder, line 113: Nit: Looks like maybe the previous lines are tabs and these are spaces.  Hard to tell for sure in the Github UI.

.lgtm.yml
```diff
@@ -6,3 +6,7 @@ extraction:
         - python3-six
         - python3-yaml
         - python3-jsonschema
+
+path_classifiers:
+  vendor:
+    - src/bindings/python/utils
```
SteVwonder, line 12:
```suggestion
    - src/bindings/python/flux/utils
```

SteVwonder (commented):
Thanks @grondo! Just a few comments but generally LGTM!

grondo:
Ok, _finally_ figured out how to correctly suppress code coverage under `src/bindings/python/flux/utils`, but I can't seem to suppress the LGTM report, even though I've followed the docs to classify all files under that path.

grondo:
> Ok, finally figured out how to correctly suppress code coverage under src/bindings/python/flux/utils,

Argh, nevermind, all I succeeded in doing was breaking the python code coverage.

src/test/checks_run.sh
grondo: Good catch, that was the reason coverage broke for python. Already addressed.

.lgtm.yml
grondo: Yep, another thing I caught in debugging. :+1: nice catch!

src/bindings/python/flux/util.py
grondo: Oh, heh, good catch!

garlick (approved):
This looks good to me.  Feel free to set MWP once @SteVwonder 's comments are addressed and you're happy.  This seems like the right move to me.

grondo:
Ok, looks like I solved the coverage.py/codecov issues. LGTM is still reporting errors for the vendored files, but maybe it is using `.lgtm.yml` from the main branch?

I'll address @SteVwonder's comments (thanks!) in the next push (I'll just squash the result)

grondo:
> Feel free to set MWP once @SteVwonder 's comments are addressed

Great, I believe I addressed all of those so setting MWP.

grondo:
> Feel free to set MWP once @SteVwonder 's comments are addressed

Great, I believe I addressed all of those so setting MWP.

SteVwonder:
> LGTM is still reporting errors for the vendored files

I wonder if it is the `vendor` tag. I think `library` might do what you want. From [their docs](https://help.semmle.com/lgtm-enterprise/user/help/file-classification.html#built-in-tags):

> As you can see, file classification is really a question of identifying the source or purpose of a file, and giving it a meaningful tag. Currently, the built-in tags are: docs, generated, library, template and test.
> Library code, imported code or third-party

grondo:
> I wonder if it is the vendor tag. I think library might do what you want. From their docs:

I probably should have tried that, but also from the docs:

> You can use LGTM's built-in tags or you can use your own tags. It depends on whether you want to overwrite the default behavior, or just mark specific files in your project so that their results are not shown or used in statistics in LGTM.

Seemed like that implied that any classified file should be ignored in reports.. :shrug:

PR #3709, opened 2021-06-10
testsuite: add tests for late joining broker

garlick:
This PR adds a `flux start --test-start-mode` option that can start the instance with only the leader broker running.  The `system` sharness test personality is modified so that this can be configured from a test script.  Finally, a test script is added that checks a few basics with a late joining broker such as:
* `flux exec` to down broker doesn't hang
* down broker is not drained
* down broker is still part of the resource set used to evaluate jobspec satisfiability
* once the broker is brought up, jobs run that were stuck before

A few minor nits were fixed while working on the test.

PR #3712, opened 2021-06-11
broker: send offline responses while broker is initializing

garlick:
Problem: if a user submits a job on a login node, but the broker upstream of that node is not yet online, the request may hang leaving the user wondering what's going on.

Send an EAGAIN response to user requests (route count >= 2) received before the broker enters INIT state
```
$ flux mini run hostname
flux-mini: ERROR: [Errno 11] Upstream Flux broker is not yet online. Try again later.
```
A sharness test is added that re-creates this situation.

This is based on top of #3709 (this PR is just the last 3 commits).

### DIFF
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -1540,13 +1540,13 @@ static void module_cb (module_t *p, void *arg)
             }
             /* Requests sent by the module on behalf of _its_ peers, e.g.
              * connector-local module with connected clients, will have a
-             * route count greater than one here.  If this broker is not yet
+             * route count greater than one here.  If this broker is not
              * "online" (entered INIT state), politely rebuff these requests.
              * Possible scenario for this message: user submitting a job on
              * a login node before cluster reboot is complete.
              */
             else if (count > 1 && !ctx->online) {
-                const char *errmsg = "Upstream Flux broker is not yet online."
+                const char *errmsg = "Upstream Flux broker is offline."
                                      " Try again later.";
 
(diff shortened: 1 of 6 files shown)


garlick:
I think the coverage target is not really going to be achievable here.  It's a small code change, that unfortunately includes a high ratio of difficult to cover error paths.

src/broker/broker.c
```diff
@@ -1528,13 +1529,31 @@ static void module_cb (module_t *p, void *arg)
...
+            /* Requests sent by the module on behalf of _its_ peers, e.g.
+             * connector-local module with connected clients, will have a
+             * route count greater than one here.  If this broker is not yet
+             * "online" (entered INIT state), politely rebuff these requests.
+             * Possible scenario for this message: user submitting a job on
+             * a login node before cluster reboot is complete.
+             */
+            else if (count > 1 && !ctx->online) {
+                const char *errmsg = "Upstream Flux broker is not yet online."
+                                     " Try again later.";
```
grondo, line 1550: Would it make the error more general to just say `Upstream broker is offline`? Or does this really only apply at startup when brokers have "not yet" been online?

garlick: Offline is probably better, even though we say we don't upstream to go down without shutting down downstream.  Soon that will have to change.

garlick:
The error message looks like this now (commit messages updated to match):
```
flux-mini: ERROR: [Errno 11] Upstream Flux broker is offline. Try again later.
```

grondo (approved):
LGTM!

garlick:
Thanks!  Setting MWP.

PR #3713, opened 2021-06-11
flux-python: fix use of virtualenv python

SteVwonder:
Problem: when a virtualenv is created, it symlinks the system python
into the virtualenv directory.  The full path to this symlink is then
compiled into the `flux-python` subcommand as the first argument to
`execve`, but the first value in the argument list (`execve`'s second
arg), is just `python`.  This leads to an "escape" of the virtualenv
and a return to the system python environment:

```
PATH=/usr/bin/:$PATH strace -e process -f flux python -c 'import sys; print(sys.executable)'
execve("/usr/bin/flux", ["flux", "python", "-c", "import sys; print(sys.executable"...], 0x7fffd6b27020 /* 18 vars */) = 0
arch_prctl(ARCH_SET_FS, 0x7f849efccb00) = 0
execve("/ve_exaworks/bin/python3", ["python", "-c", "import sys; print(sys.executable"...], 0x20e73f0 /* 26 vars */) = 0
arch_prctl(ARCH_SET_FS, 0x7f7545579740) = 0
/usr/bin/python
exit_group(0)                           = ?
+++ exited with 0 +++
```

Solution: set the first value in the argument list to the full python
interpreter path.  This results in the argv of the python process being
the full interpreter path as opposed to just `python`, ensuring that the
virtualenv environment is maintained:

```
PATH=/usr/bin/:$PATH flux python -c 'import sys; print(sys.executable)'
/ve_exaworks/bin/python
```

### DIFF
--- a/src/cmd/builtin/python.c
+++ b/src/cmd/builtin/python.c
@@ -16,6 +16,13 @@
 
 static int cmd_python (optparse_t *p, int ac, char *av[])
 {
+    /*
+     *  Ensure av[0] matches the full path of the interpreter we're executing.
+     *  Other than just being good practice, this ensures that sys.executable
+     *  is correct (since python uses sys.argv[0] for sys.executable) and so
+     *  that symlink'd binaries in virtualenvs are respected.
+     */
+    av[0] = PYTHON_INTERPRETER;
     execv (PYTHON_INTERPRETER, av); /* no return if sucessful */
     log_err_exit ("execvp (%s)", PYTHON_INTERPRETER);
     return (0);
--- a/t/t0001-basic.t
+++ b/t/t0001-basic.t
@@ -59,6 +59,13 @@ test_expect_success 'flux-python command runs a python that finds flux' '
 	flux python -c "import flux"
 '
 
+test_expect_success 'flux-python command runs the configured python' '
+	define_line=$(grep "^#define PYTHON_INTERPRETER" ${FLUX_BUILD_DIR}/config/config.h) &&
+	expected=$(echo ${define_line} | sed -E '\''s~.*define PYTHON_INTERPRETER "(.*)"~\1~'\'') &&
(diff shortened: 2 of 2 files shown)


src/cmd/builtin/python.c
```diff
@@ -16,6 +16,7 @@
 
 static int cmd_python (optparse_t *p, int ac, char *av[])
 {
+    av[0] = PYTHON_INTERPRETER; // Respect symlink'd binaries in virtualenvs
```
grondo, line 19: It should be noted that this affects more than just virtualenv. Before this change, even a non-virtualenv python interpreter will have `av[0]` set to simply `python` (since `av[0]` in a subcommand is always the name of the subcommand). Since python uses `av[0]` to set `sys.executable`, on my system where `PYTHON_INTERPRETER` is `/usr/bin/python3`, `sys.executable` isn't `/usr/bin/python3` or even `python3`, it is just `python`. I guess we've gotten away with it until now.

Maybe we can make the comment here more general, e.g.

```c
/*
 *  Ensure av[0] matches the full path of the interpreter we're executing.
 *  Other than just being good practice, this ensures that sys.executable
 *  is correct (since python uses sys.argv[0] for sys.executable) and so
 *  that symlink'd binaries in virtualenvs are respected.
 */
```

grondo (approved):
This LGTM.

I added a note about maybe generalizing the comment, since the fix will definitely apply outside of virtualenv. It is optional if you take the suggestion though

SteVwonder:
Thanks @grondo.  Just force pushed a change that includes your suggest comment and a test that checks the `sys.executable` against the compiled-in interpreter value.  I confirmed that this test fails in the flux-core test-env docker image without the fix and passes after the fix is applied.  Let me know what you think.

grondo:
Looks good to me! Nice job on the test, and thanks for discovering this bug!

Feel free to set MWP (I had already approved) when ready :+1:

PR #3714, opened 2021-06-12
Python/FluxExecutor bug fixes

jameshcorbett:
Two bug fixes for FluxExecutor:

First, in the event that a FluxExecutor thread crashes, the executor becomes completely unusable and your Python program will hang. That's fine as long as threads just don't crash, but I found that they do crash when the executor is connected to a remote Flux instance which has gone away. The first commit makes the threads a little more robust. When a thread raises an exception, all outstanding futures have that exception set on them. That is, ``fut.result()`` might return something like ``RuntimeError("can't connect to remote Flux instance at ...")``. Moreover, all future calls to ``submit()`` will raise a ``RuntimeError`` (this aspect is not necessary and I'm not sure it's even desirable---feedback welcome).

Second, when a job cannot be submitted (e.g. because of an invalid jobspec) the jobid will never be set. But then calls to ``fut.jobid()`` hang. Fix that by detecting when a job could not be submitted and raise a ``RuntimeError`` when accessing the jobid.

### DIFF
--- a/src/bindings/python/flux/job/executor.py
+++ b/src/bindings/python/flux/job/executor.py
@@ -40,6 +40,7 @@ class _FluxExecutorThread(threading.Thread):
     # pylint: disable=too-many-arguments
     def __init__(
         self,
+        broken_event,
         exit_event,
         jobspecs_to_submit,
         poll_interval,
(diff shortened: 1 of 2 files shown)


grondo (approved):
LGTM, and thanks for including tests!

I'm not capable of fully reviewing the Python code, but I checked out the branch, verified the tests pass and ran some of my scripts that use the executor with no issues. 

Thanks!

jameshcorbett:
Thanks @grondo. I pushed some extra tests and rebased with master. I didn't make any changes to the executor though.

src/bindings/python/flux/job/executor.py
```diff
@@ -40,6 +40,7 @@ class _FluxExecutorThread(threading.Thread):
     # pylint: disable=too-many-arguments
     def __init__(
         self,
+        broken_event,
```
SteVwonder, line 43: Thoughts on making this an optional argument, and if it is not provided, just automatically instantiating a new `threading.Event()`?  Seems like that might clean up some of the new test code.

SteVwonder (approved):
LGTM! Thanks @jameshcorbett. One optional comment below:

src/bindings/python/flux/job/executor.py
jameshcorbett: It definitely would. If you ignore the tests, though, I think the intention could be unclear, since ``FluxExecutor`` always passes all of the arguments to its threads---although it could always be explained with a comment. So on balance I'm ambivalent.

SteVwonder: Sounds good. Happy to leave as is.

SteVwonder:
Ready for this to go in @jameshcorbett ?

jameshcorbett:
Yup! Thanks.

grondo:
Ok, I've set MWP.

grondo:
Thanks again, @jameshcorbett!

PR #3717, opened 2021-06-13
github: fix the fix for ci builds on master

grondo:
Problem: The fix for builds on master didn't work, since GITHUB_BRANCH
was not set in the _environment_ of the annotated tags step (it was set
as an output).

Fall back to using github.ref, and skip the annotated tags checkout
when `github.ref == 'refs/heads/master'`.

Add a line of debugging in case this step starts to fail again in the
future.

This time I actually tested by pushing to the main branch of my own repo.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -122,13 +122,14 @@ jobs:
     - name: fetch annotated tag
       if: >
         (matrix.create_release || matrix.docker_tag) &&
-        env.GITHUB_BRANCH != 'master'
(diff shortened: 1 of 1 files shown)


grondo:
This will need a manual merge button press (assuming it passes ci).

Sorry for the noise!

garlick (approved):
Ready to merge?

garlick:
I'm just going to merge it - seems like that was your intention.

grondo:
Yes, seems to have worked this time. Thanks!

PR #3720, opened 2021-06-15
libflux: make flux_plugin_handler topic member const

### LINKED ISSUE #3719: libflux: `struct flux_plugin_handler` should use `const char *` for topic string
Currently `struct flux_plugin_handler` uses `char *topic` to store the topic string, but this causes a compilation failure on >= C++11 when placing the topic string in the initializer because string literals are now `const char []` 

https://en.cppreference.com/w/cpp/language/string_literal
> String literals are not convertible or assignable to non-const CharT*. An explicit cast (e.g. const_cast) must be used if such conversion is wanted. (since C++11)

It also just seems like expected behavior to make this member of `struct flux_plugin_handler` `const char *`.

grondo:
Problem: The topic string in `struct flux_plugin_handler` is not
const, which causes compilation errors when using static initializers
to define a table of handlers on compilers where static strings
are const (e.g. C++11).

Make the topic string const. The only reason it wasn't const was as a
convenience when destroying copied flux_plugin_handler structs. Just
cast the topic back to non-const when freeing the copy.

Fixes #3719

### DIFF
--- a/src/common/libflux/plugin.c
+++ b/src/common/libflux/plugin.c
@@ -52,7 +52,7 @@ typedef const struct flux_plugin_handler *
 static void flux_plugin_handler_destroy (struct flux_plugin_handler *h)
 {
     if (h) {
-        free (h->topic);
+        free ((char *) h->topic);
         free (h);
     }
 }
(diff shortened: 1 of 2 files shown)


garlick (approved):
LGTM!

grondo:
Thanks!

Issue #3723, opened 2021-06-15
t3470-multithread-reactor-run.py issue test intermittent failures


garlick:
This test is failing intermittently on my system (Ubuntu focal, `python-3.8.2-0ubuntu2`), more often when running tests in parallel but occasionally standalone.
 
```
$ ./t4000-issues-test-driver.t -v
[snip]
expecting success: run_timeout 120 /home/garlick/proj/flux-core/t/issues/t3470-multithread-reactor-run.py
starting 2 threads
got response from <Thread(Thread-1, started 140374035314432)>
got response from <Thread(Thread-2, started 140373959636736)>
2 threads started
published test-event
1: hello
Fatal Python error: ffi.from_handle() detected that the address passed points to garbage. If it is really the result of ffi.new_handle(), then the Python object has already been garbage collected
Python runtime state: initialized

Current thread 0x00007fab6082b700 (most recent call first):
  File "/home/garlick/proj/flux-core/src/bindings/python/flux/message.py", line 122 in message_handler_wrapper
  File "/home/garlick/proj/flux-core/src/bindings/python/flux/wrapper.py", line 194 in __call__
  File "/home/garlick/proj/flux-core/src/bindings/python/flux/core/handle.py", line 274 in reactor_run
  File "/home/garlick/proj/flux-core/t/issues/t3470-multithread-reactor-run.py", line 23 in get_events
  File "/usr/lib/python3.8/threading.py", line 870 in run
  File "/usr/lib/python3.8/threading.py", line 932 in _bootstrap_inner
  File "/usr/lib/python3.8/threading.py", line 890 in _bootstrap

Thread 0x00007fab63f42740 (most recent call first):
  File "/usr/lib/python3.8/threading.py", line 1027 in _wait_for_tstate_lock
  File "/usr/lib/python3.8/threading.py", line 1011 in join
  File "/home/garlick/proj/flux-core/t/issues/t3470-multithread-reactor-run.py", line 55 in main
  File "/home/garlick/proj/flux-core/t/issues/t3470-multithread-reactor-run.py", line 61 in <module>
not ok 15 - t3470-multithread-reactor-run.py
#	run_timeout 120 /home/garlick/proj/flux-core/t/issues/t3470-multithread-reactor-run.py
[snip]
```

garlick:
Is it possible that the watcher is eligible for garbage collection immediately upon being started, since it is not (directly)? referenced from the `get_events()` thread after that?

[...]

Issue #3546, opened 2021-03-04
testsuite: possible hang in t2607-job-shell-input.t


grondo:
A couple times I've seen the ci timeout with `t2607-job-shell-input.t` implicated:

```
Error: Detected 1 missing missing tests:
t/t2607-job-shell-input.t
```

The apparently hanging test seems to be this one:
```
 expecting success: 
          id1=$(flux mini submit -n1 \
                ${TEST_SUBPROCESS_DIR}/test_echo -O -n)
          id2=$(flux mini submit -n1 \
                ${TEST_SUBPROCESS_DIR}/test_echo -O -n)
          id3=$(flux mini submit -n1 \
                ${TEST_SUBPROCESS_DIR}/test_echo -O -n)
          id4=$(flux mini submit -n1 \
                ${TEST_SUBPROCESS_DIR}/test_echo -O -n)
          flux job attach $id1 < input_stdin_file > pipe5_1.out 2> pipe5_1.err &
          pid1=$!
          flux job attach $id2 < input_stdin_file > pipe5_2.out 2> pipe5_2.err &
          pid2=$!
          flux job attach $id3 < input_stdin_file > pipe5_3.out 2> pipe5_3.err &
          pid3=$!
          flux job attach $id4 < input_stdin_file > pipe5_4.out 2> pipe5_4.err &
          pid4=$!
          wait $pid1 &&
          wait $pid2 &&
          wait $pid3 &&
          wait $pid4 &&
          test_cmp input_stdin_file pipe5_1.out &&
          test_cmp input_stdin_file pipe5_2.out &&
          test_cmp input_stdin_file pipe5_3.out &&
          test_cmp input_stdin_file pipe5_4.out
```

PR #3725, opened 2021-06-15
testsuite: fix intermittent test, speed up others

garlick:
Here are a couple of minor test improvements.   One commit fixes a test that occasionally fails for me (#3723) and the other speeds up two tests by loading the heartbeat module with a very fast heartrate, and not loading unnecessary modules.


grondo (approved):
LGTM, thanks!

garlick:
Thanks!

garlick:
Restarted tests - looks like one builder was canceled during the shell input test (appeared to have been canceled though - not an outright failure like #3546).

PR #3728, opened 2021-06-16
cron: fix race in task timeouts

grondo:
For whatever reason I was getting reproducible failures in  `t0015-cron.t`:  `not ok 2 - flux-cron can set timeout on tasks` (a failure I've seen intermittently, but never reproducible). It turns out this was not due to a race in the test, but a race in the cron module itself. The timeout timer was being started when the cron task was started, not when it was actually running. If the timer fired before the task was running, then the attempt to kill it will fail.

There are other problems in the cron module, but we'll not tackle those now. This PR just fixes the race by starting the timer when the task is Running instead of Starting.

### DIFF
--- a/src/modules/cron/cron.c
+++ b/src/modules/cron/cron.c
@@ -82,14 +82,6 @@ double get_timestamp (void)
     return ((double) tm.tv_sec + (tm.tv_nsec/1.0e9));
 }
 
-static void timeout_cb (flux_t *h, cron_task_t *t, void *arg)
-{
-    cron_entry_t *e = arg;
-    flux_log (h, LOG_INFO, "cron-%ju: task timeout at %.2fs. Killing",
-              e->id, e->timeout);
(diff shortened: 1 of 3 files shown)


src/modules/cron/cron.c
```diff
@@ -82,14 +82,6 @@ double get_timestamp (void)
     return ((double) tm.tv_sec + (tm.tv_nsec/1.0e9));
```
garlick, line 82: Commit message: "remove" is misspelled.

garlick (approved):
Thanks!

src/modules/cron/cron.c
grondo: Thanks, I fixed that and forced a push.

grondo:
Didn't realize this was already approved. Thanks! Setting MWP.

grondo:
Restarted all tests because the `test-install` build timed out.

PR #3730, opened 2021-06-17
testsuite: fix several tests on slower systems

grondo:
I was getting several reproducible testsuite failures with a build directory configured with `--enable-code-coverage` while running `make -j 16 check` on a system with 4 cores.

This PR fixes all of the frequently failing tests in this environment.

### DIFF
--- a/t/sharness.d/flux-sharness.sh
+++ b/t/sharness.d/flux-sharness.sh
@@ -34,6 +34,29 @@ test_size_large() {
 
 
 #
+#  Like test_must_fail(), but additionally allow process to be
+#   terminated by SIGKILL or SIGTERM
+#
+test_must_fail_or_be_terminated() {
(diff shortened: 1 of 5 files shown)


t/sharness.d/flux-sharness.sh
```diff
@@ -33,6 +33,29 @@ test_size_large() {
 }
 
 
+#
```
chu11, line 36: commit message english error, "does we want" -> "does what we want"

chu11 (approved):
LGTM, just one typo in commit message.

t/sharness.d/flux-sharness.sh
grondo: Thanks! Fixed and pushed

grondo:
Setting MWP on this one

Issue #3736, opened 2021-06-22
 ./t3300-system-basic.t fails on a TOSS3 machine


dongahn:
```console
rzhasgpu18{dahn}50: pwd
/usr/global/tools/flux/toss_3_x86_64_ib/build/2021-06-21-c0.28.0.pre-s0.17.0.pre/flux-core/t

rzhasgpu18{dahn}49: ./t3300-system-basic.t
lt-flux-broker: overlay_connect: Cannot allocate memory
lt-flux-broker: overlay_connect: Cannot allocate memory
lt-flux-broker: overlay_bind: File name too long
lt-flux-broker: bootstrap failed
flux-start: 0 (pid 27065) exited with rc=1
flux-start: 1 (pid 27066) Killed
flux-start: 2 (pid 27068) Killed
```

Maybe there was a change in security authentication which hit a short file path limit for my /usr/global configuration?

dongahn:
Same issue with two more tests:

```console
rzhasgpu18{dahn}71: ./t3301-system-latestart.t
lt-flux-broker: overlay_bind: File name too long
lt-flux-broker: bootstrap failed
flux-start: 0 (pid 30591) exited with rc=1

rzhasgpu18{dahn}73: ./t3302-system-offline.t
lt-flux-broker: overlay_bind: File name too long
lt-flux-broker: bootstrap failed
flux-start: 0 (pid 30788) exited with rc=1
```

garlick:
That test does use the `system` personality which  is new, and would select the bind address to be in the sharness trash directory instead of in /tmp, e.g.
```
/usr/global/tools/flux/toss_3_x86_64_ib/build/2021-06-21-c0.28.0.pre-s0.17.0.pre/flux-core/t/trash-directory.t3300-system-basic/tbon-0
```
Oh!  Ot sounds like the max path for UNIX domain sockets is on the order of 100 bytes!   (I did not know or had forgotten this)
https://unix.stackexchange.com/questions/367008/why-is-socket-path-length-limited-to-a-hundred-chars

We will need to fix that.

dongahn:
Yes I agree with your assessment. I had a similar issue with one of my tests and wasn't immediately clear how to fix it, though.

garlick:
I think what we need to do is place those sockets in the `rundir` like the FLUX_URI, since normally that is a short path relative to /tmp.  The config file referencing the socket paths is created before `flux start` runs.

So I think either
[...]

PR #3739, opened 2021-06-23
testsuite: avoid long ipc:// paths in system test personality

garlick:
This is a fix for the problem @dongahn encountered when building flux-core in a long directory, reported in #3736.

The fix is to use a directory in /tmp for the broker "rundir" in the system test personality, and put the sockets there instead of using the test trash directory, which is relative to the build tree.  To get the directory cleaned up properly, I added a new option to `flux-start`.  That seemed a bit easier than getting sharness to do it (at least, I was having a hard time getting that to work).  This approach is fine IMHO.

### DIFF
--- a/doc/man1/flux-start.rst
+++ b/doc/man1/flux-start.rst
@@ -81,9 +81,13 @@ OPTIONS
    the other brokers are online.  Default: ``all``.
 
 **--test-rundir**\ =\ *PATH*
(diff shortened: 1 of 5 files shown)


grondo (approved):
LGTM.

garlick:
Thanks!  I'll set MWP.

PR #3740, opened 2021-06-23
allow jobs to bypass scheduler with alloc-bypass jobtap plugin

grondo:
This is a proof-of-concept implementation of the idea discussed in #3737. This will probably need a bit more work before it could be a merge candidate, but I thought it might be useful to push here for now in order to continue the discussion around a concrete implementation.

The current approach is pretty straightforward:

 * Add a new `alloc_bypass` flag to the job manager's `struct job`. If this flag is set then alloc/free requests are not sent to the loaded scheduler
 * Allow a jobtap plugin to return an `R` in the `job.state.sched` callback, which, if the job was submitted by the instance owner, enables "alloc bypass" mode, writes the `R` to the appropriate KVS path, and finally generates an `alloc` event directly
 * Add an `alloc-bypass.so` loadable jobtap plugin which returns an `R` in `job.state.sched` if one is set in `attributes.system.R`
 * Add a new convenience `--with-R` option to `flux-mini.py` which writes its argument to `attributes.system.R` and also sets the `per-resource.type` shell option to `node` (running one process per node by default)

Actually, while working on this, I wondered if instead of returning an `R` directly in the plugin "out" arguments, a plugin should be allowed to simply set the alloc_bypass flag on a job. This would allow the plugin requesting alloc bypass to asynchronously commit `R` itself to the KVS and generate the `alloc` event when ready. We'd have to add some helper functions to the interface, but this seems a bit more powerful than the current approach, since it allows for cases where `R` is not immediately available.

In kind, the generation of the `free` event should be left to the plugin as well. Currently the `free` event is simply posted immediately upon entry to the CLEANUP state if `job->alloc_bypass` is set.

In fact, I wonder if that kind of interface would be useful for the simulator?

Open to any other ideas of course, just thought it might be useful to have a somewhat working implementation. Examples:

```console
$ flux resource list
      free      4        8        0 asp,asp,asp,asp
 allocated      4        8        0 asp,asp,asp,asp
      down      0        0        0 
# load required jobtap plugin:
$ flux jobtap load alloc-bypass.so

# run across arbitrary resources using `flux R encode`
$ flux mini run --with-R="$(flux R encode -r 2-3 -c 3)" --label-io sh -c 'echo rank is $(flux getattr rank); taskset -c -p $$'
1: rank is 3
0: rank is 2
1: pid 28180's current affinity list: 3
0: pid 28181's current affinity list: 3

# Run across job resources using `flux job info` and `flux R decode`
$ flux jobs
    ƒAufacbR grondo   sleep       R      4      2   6.479m asp,asp
    ƒASKFs4T grondo   sleep       R      4      2   6.497m asp,asp
$ flux mini run --with-R="$((flux job info ƒASKFs4T R && flux job info ƒAufacbR R) | flux R decode)"  --label-io sh -c 'echo rank is $(flux getattr rank); taskset -c -p $$'
1: rank is 1
3: rank is 3
2: rank is 2
1: pid 30856's current affinity list: 0,1
0: rank is 0
0: pid 30855's current affinity list: 0,1
2: pid 30853's current affinity list: 0,1
3: pid 30854's current affinity list: 0,1

# Run multiple tasks per core instead of 1 per node:
$ flux mini run --with-R="$(flux R encode -r 2 -c 0)" -o per-resource.type=core -o per-resource.count=4 --label-io sh -c 'echo rank is $(flux getattr rank); taskset -c -p $$'
0: rank is 2
1: rank is 2
0: pid 32098's current affinity list: 0
1: pid 32099's current affinity list: 0
2: rank is 2
3: rank is 2
2: pid 32101's current affinity list: 0
3: pid 32105's current affinity list: 0

```

### DIFF
--- a/src/modules/job-manager/plugins/alloc-bypass.c
+++ b/src/modules/job-manager/plugins/alloc-bypass.c
@@ -1,0 +1,229 @@
+/************************************************************\
+ * Copyright 2021 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* alloc-bypass.c - If attributes.system.R exists in jobspec, then
+ *  bypass scheduler alloc protocol and use R directly (for instance
+ *  owner use only)
+ */
+
+
+#include <unistd.h>
+#include <sys/types.h>
+
+#include <jansson.h>
+#include <flux/core.h>
+#include <flux/jobtap.h>
+
+#include "src/common/librlist/rlist.h"
+
+static void alloc_continuation (flux_future_t *f, void *arg)
+{
+    flux_plugin_t *p = arg;
+    flux_jobid_t *idptr = flux_future_aux_get (f, "jobid");
+
+    if (flux_future_get (f, NULL) < 0) {
+        flux_jobtap_raise_exception (p,
+                                     *idptr,
+                                     "alloc", 0,
+                                     "failed to commit R to kvs: %s",
+                                      strerror (errno));
+        goto done;
+    }
+    if (flux_jobtap_event_post_pack (p,
+                                     *idptr,
+                                     "alloc",
+                                     "{s:b}",
+                                     "bypass", true) < 0)
+        flux_jobtap_raise_exception (p,
+                                     *idptr,
+                                     "alloc", 0,
+                                     "failed to post alloc event: %s",
+                                     strerror (errno));
+
+    /*  Set "needs-free" so that alloc-bypass knows that a "free"
+     *   event needs to be emitted for this node.
+     */
+    if (flux_jobtap_job_aux_set (p, *idptr, "alloc-bypass::free", p, NULL) < 0)
+        flux_log_error (flux_jobtap_get_flux (p),
+                        "id=%ju: Failed to set alloc-bypass::free",
+                        *idptr);
+
+done:
+    flux_future_destroy (f);
+}
+
+static int alloc_start (flux_plugin_t *p,
+                        flux_jobid_t id,
+                        const char *R)
+{
+    flux_t *h;
+    char key[64];
+    flux_future_t *f = NULL;
+    flux_kvs_txn_t *txn = NULL;
+    flux_jobid_t *idptr = NULL;
+
+    if (!(h = flux_jobtap_get_flux (p))
+        || flux_job_kvs_key (key, sizeof (key), id, "R") < 0
+        || !(txn = flux_kvs_txn_create ())
+        || flux_kvs_txn_put (txn, 0, key, R) < 0
+        || !(f = flux_kvs_commit (h, NULL, 0, txn))
+        || flux_future_then (f, -1, alloc_continuation, p)
+        || !(idptr = calloc (1, sizeof (*idptr)))
+        || flux_future_aux_set (f, "jobid", idptr, free) < 0) {
+        flux_kvs_txn_destroy (txn);
+        flux_future_destroy (f);
+        free (idptr);
+        return -1;
+    }
+    *idptr = id;
+    flux_kvs_txn_destroy (txn);
+    return 0;
+}
+
+
+static int sched_cb (flux_plugin_t *p,
+                     const char *topic,
+                     flux_plugin_arg_t *args,
+                     void *arg)
+{
+    const char *R = NULL;
+    flux_jobid_t id;
+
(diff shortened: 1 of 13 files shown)


grondo:
Oh, I also did sanity check that at least a basic MPI job will run in this configuration (not too surprising, but good to check since `per-resource` isn't well tested outside of `flux mini batch/alloc`)
```console
$ flux mini run --with-R="$(flux R encode -r 0-3 -c 0-3)" ./hello
ƒAnR2JKqZ: completed MPI_Init in 0.017s.  There are 4 tasks
ƒAnR2JKqZ: completed first barrier in 0.000s
ƒAnR2JKqZ: completed MPI_Finalize in 0.001s
$ flux mini run --with-R="$(flux R encode -r 0-3 -c 0-3)" -o per-resource.type=core ./hello
ƒA12E9jgK: completed MPI_Init in 0.222s.  There are 16 tasks
ƒA12E9jgK: completed first barrier in 0.044s
ƒA12E9jgK: completed MPI_Finalize in 0.037s
```

grondo:
I just pushed an update here that adds a couple new functions to the jobtap interface `flux_jobtap_job_set_flag()` and `flux_jobtap_event_post_pack()`, which the `alloc-bypass.so` plugin can then use to set `alloc-bypass` mode on a job if requested, and issue the `alloc` and `free` events internally. The feature to return an _R_ in the plugin out args in SCHED state (and associated code) is then removed from the job-manager internals.

I left these as additional commits for now. They will need to be squashed if the general approach is acceptable.

src/modules/job-manager/plugins/alloc-bypass.c
```diff
@@ -21,33 +21,116 @@
...
+    if (flux_job_kvs_key (key, sizeof (key), id, "R") < 0
+        || !(txn = flux_kvs_txn_create ())
+        || flux_kvs_txn_put (txn, 0, key, R) < 0
+        || !(f = flux_kvs_commit (h, NULL, 0, txn))
+        || flux_future_aux_set (f, "jobid", idptr, free) < 0
+        || flux_future_then (f, -1, alloc_continuation, p)) {
+        flux_kvs_txn_destroy (txn);
+        flux_future_destroy (f);
+        return -1;
+    }
```
garlick, line 77: idptr is leaked if any of the chained checks fail up to and including the aux_set.

garlick (approved):
This looks good to me! Just one memory leak on an error path I noticed.

src/modules/job-manager/plugins/alloc-bypass.c
grondo: Good catch. Passing a jobid to a callback (future or message) may be a common case. I wonder if we could make this easier somehow.. kind of annoying to have to allocate memory for this.

garlick: Maybe it would be useful to have a handle for a job that is opaque to the plugin, but actually references a struct job?

grondo: Uh yeah, I was wishing we'd gone that way before.. it would be a change to the general approach of the API, but perhaps a lot more user friendly given the evolution of the api..

grondo:
I've pushed a rework of this PR with the following changes:

 * move responsibility of emitting `alloc` and `free` events in `alloc-bypass` mode to the plugin requesting this flag. This will allow plugins to generate/fetch R asynchronously if necessary and emit the `alloc` event when R is committed to the KVS.
 * To allow the above, add `flux_jobtap_event_post_pack()`
 * add a new job eventlog event `set-flags` to allow setting of named flags on jobs. This ensures those flags are reestablished on eventlog replay. the only supported flags are currently `debug` and `alloc-bypass`
 * do not allow any events to be emitted before the DEPEND state in `event_job_post_pack()`. These events would appear in the event journal before the `submit` event (though they appear after the `submit` event in the actual eventlog)
 * add basic tests
 * remove the addition of `--with-R` from `flux-mini`. Probably we want to think more carefully about the `flux-mini` interface that would be used here, but the alloc-bypass functionality could be used now using `--setattr` or hand-constructed jobspec.

grondo:
Latest round of updates (sorry I already rebased and squashed the result)

 - Add some state to the job aux item list when a `free` needs to be emitted by the plugin to avoid posting a free event when the job is canceled before an `alloc` event, add test to ensure extraneous `free` event is not emitted
 - change the jobspec location of the alloc-bypass R to `attributes.system.alloc-bypass.R`
 - parse the incoming R in `alloc-bypass.so` with librlist to detect completely invalid R, add tests with invalid R

I'll remove WIP since this should get a more extensive review now.

garlick (approved):
Great - no problems found in another spin through all the commits.

grondo:
Ok, setting MWP. Once there are some use cases we can see what kind of front end makes sense here, but we can have this plugin in our back pocket in case anyone needs it.

PR #3567, opened 2021-03-22
WIP: add flux job exec

garlick:
As discussed in #3556, this PR adds an experimental run command with less overhead, and fewer features, than `flux mini run`:
```
Usage: flux-job exec [OPTIONS]
Run job with one task per core to completion
  -h, --help             Display this message.
  -n, --ntasks=COUNT     Run N tasks (default 1)
  -l, --label-io         Label output by rank
```
It is available only to the instance owner, and has very limited options as shown above.  A specialized service in the job-manager accepts the job submission and commits jobspec to the KVS.  Once the KVS commit completes, it posts the `submit` event and the state machine takes it from there.   To obtain the result, the FLUX_JOB_WAITABLE flag is set, and the original request effectively gets a wait response once the job becomes inactive.  That allowed the service to be implemented with a fairly small amount of code.

Other limitations include
- no shell options
- no submit options
- no duration or other options
- resources expressed only in cores
- task to resource mapping is one task per core
- output is not produced while the job runs; it is emitted all at once at the end
- signals are not forwarded (e.g. aborting before job is done does not abort job)
- no environment options

There are probably more I'm not thinking of.

Anyway, I thought I would post this half formed thing as a demo, for @jedbrown to poke at and see if it helps his use case.

PR #3741, opened 2021-06-23
WIP: add flux job exec and job manager fastpath

garlick:
Zombie alert - #3567 is not quite dead!

This is a rebased version of the original `flux job exec` hack.  The C jobspec support in the original was since improved and merged.  There is less and less code needed to make this work, so I thought I might as well bring it forth again.

It's still pretty fast, e.g.
```
$ time flux job exec hostname
system76-pc

real	0m0.075s
user	0m0.025s
sys	0m0.003s

$ time flux mini run hostname
system76-pc

real	0m0.245s
user	0m0.071s
sys	0m0.020s

# one more time after validator started:

$ time flux mini run hostname
system76-pc

real	0m0.146s
user	0m0.072s
sys	0m0.020s
```

I didn't try to change the user interface - I'm sure that could be improved.

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -29,6 +29,7 @@ MAN1_FILES_PRIMARY = \
 	man1/flux-cron.1 \
 	man1/flux-event.1 \
 	man1/flux-mini.1 \
+	man1/flux-job.1 \
 	man1/flux-version.1 \
 	man1/flux-jobs.1 \
 	man1/flux-shell.1 \
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -131,6 +131,7 @@ man_pages = [
     ('man1/flux-kvs', 'flux-kvs', 'Flux key-value store utility', [author], 1),
     ('man1/flux-logger', 'flux-logger', 'create a Flux log entry', [author], 1),
     ('man1/flux-mini', 'flux-mini', 'Minimal Job Submission Tool', [author], 1),
+    ('man1/flux-job', 'flux-job', 'Job Housekeeping Tool', [author], 1),
     ('man1/flux-module', 'flux-module', 'manage Flux extension modules', [author], 1),
     ('man1/flux-ping', 'flux-ping', 'measure round-trip latency to Flux services', [author], 1),
     ('man1/flux-proxy', 'flux-proxy', 'create proxy environment for Flux instance', [author], 1),
--- a/doc/man1/flux-job.rst
+++ b/doc/man1/flux-job.rst
@@ -1,0 +1,151 @@
+.. flux-help-include: true
+
+===========
+flux-job(1)
+===========
+
+
+SYNOPSIS
+========
+
+**flux** **job** **cancel** *id* [*message...*]
+
+**flux** **job** **cancelall** [*OPTIONS*] [*message...*]
+
+**flux** **job** **kill** [*--signal=SIG*] *id*
+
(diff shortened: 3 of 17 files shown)


garlick:
Just pushed a change in which `flux_jobspec1_t` is used now to create the jobspec in its entirety on the client end, rather than creating most of it and putting the pieces together on the server end.  Migrating towards that API should let the original `specutil` class be absorbed into the `jobspec1` one, though I haven't done that yet.

To do this efficiently, I added accessors for the `json_t` stored inside the `flux_jobspec_t`.

Then I noticed that `job-ingest` and `job-manager` could get rid of some one-off code for redacting the environment if they used `flux_jobpec1_t` and converted those.

(restarted unit tests - ASAN failed on a libsubprocess unit test... didn't seem related)

garlick:
Huh, same failure in ASAN
 ```
not ok 22 - flux_open on loop works
  FAIL: test_subprocess.t 22 - flux_open on loop works
  #   Failed test 'flux_open on loop works'
  #   at test/subprocess.c line 150.
```

grondo:
> Then I noticed that job-ingest and job-manager could get rid of some one-off code for redacting the environment if they used flux_jobpec1_t, so I went ahead and converted those too.

Does that mean flux-core can only support Jobspec V1, and does that break any existing external use cases, e.g. that submit an extended Jobspec for use with Fluxion?

grondo:
Also, not sure what you think of this, but the `runjob` module in the job-manager could easily be a builtin jobtap plugin. I'm not sure what that would buy us except slightly better abstraction, and the ability for someone to unload it (probably unwise) or replace it (unlikely) at runtime. Just thought I'd throw the idea out there.

garlick:
> Does that mean flux-core can only support Jobspec V1, and does that break any existing external use cases, e.g. that submit an extended Jobspec for use with Fluxion?

Oh, that is an excellent point.  Will revert.  I had it in my head that we were exclusively V1 at this point.

The motivation for using `flux_jobspec1_t` in job-ingest and job-manager was just to avoid some duplicated code.   Perhaps I can figure out another way to factor that out into an internal library.  In particular, there are some functions (now in jobspec1.c) for manipulating json_t objects using period-delimited hierarchical paths that would cover the duplicated code, and be nice to abstract in a libutil library with tests.  I'll ponder if that's a good move.

> The runjob module in the job-manager could easily be a builtin jobtap plugin

I like that idea.  In particular, it would be nice to not glom onto the wait logic as I have done here - I think a jobtap plugin now has sufficient capability that it could push a request into the aux list for a job and respond to it upon job completion.  One challenge might be creating a new job struct, initializing it, and inserting it into the active job hash.

I'm thinking the right next step for this PR is to pull out the jobspec1 changes to a separate PR, and in that PR create the libutil module mentioned above and replace the duplicated code with it.

Then the actual fastrun code could continue to be developed here, potentially as a jobtap plugin.

grondo:
>  One challenge might be creating a new job struct, initializing it, and inserting it into the active job hash.

Oh, oops! I had totally glossed over the fact that this facility needs to be able to instantiate jobs in the job manager, sorry! I wouldn't waste any time at this point trying to add that capability to the jobtap interface. We could consider it later if the use case comes up again though.

garlick:
I went ahead and rebased this on current master, after jobspec1 changes were merged.

garlick:
This may get changed yet again so adding back the WIP.

Issue #3640, opened 2021-05-08
Invalid MIT-MAGIC-COOKIE-1 key


garlick:
Problem: when starting flux on a new Ubuntu 20.04.2 LTS system, a "Invalid MIT-MAGIC-COOKIE-1 key" error is emitted to stderr:
```
$ src/cmd/flux start /bin/true
Invalid MIT-MAGIC-COOKIE-1 key
```
suppressing the `gl` hwloc component gets rid of the error.  The nvidia GPU is still detected:
```
$ HWLOC_COMPONENTS=-gl src/cmd/flux start flux resource list
     STATE NNODES   NCORES    NGPUS NODELIST
      free      1        6        1 system76-pc
 allocated      0        0        0 
      down      0        0        0 
```
This is the version of hwloc packaged with the OS (`2.1.0+dfsg-4`).
The `hwloc-gl` component is part of the `libhwloc-plugins` subpackage.

Flux can start an `xclock` (for example) as the initial program and it has no trouble authenticating to the Xorg X server.

Edit: `libhwloc-plugins` can be de-installed, but then the GPU is not detected as a resource.

garlick:
This error also causes `t3200-instance-restart.t` to fail:
```
expecting success: 
	grep $(cat id1.out) list.out &&
	grep $(cat id2.out) list.out &&
	grep $(cat id3.out) list.out

grep: MIT-MAGIC-COOKIE-1: No such file or directory
grep: keyf9SEuzo: No such file or directory
```
In the trash I find:
```
$ cat id1.out
Invalid MIT-MAGIC-COOKIE-1 keyf9SEuzo
```
Looks like the code that captures job IDs to files is including stderr.  Probably shouldn't do that.

garlick:
If it's acceptable to just prevent this plugin from being loaded, this does it programmatically
```diff
diff --git a/src/common/librlist/rhwloc.c b/src/common/librlist/rhwloc.c
index da5c92278..4d9c44d27 100644
--- a/src/common/librlist/rhwloc.c
+++ b/src/common/librlist/rhwloc.c
@@ -76,6 +76,13 @@ hwloc_topology_t rhwloc_local_topology_load (void)
 
     if (topo_init_common (&topo) < 0)
         goto err;
+    /* gl probes the NV-CONTROL X server extension, and requires X auth
+     * to be properly set up or errors are emitted to stderr.
+     * Nvidia GPUs can still be discovered via opencl.
+     */
[...]

PR #3743, opened 2021-06-24
testsuite: generalize valgrind suppressions

garlick:
I found the valgrind suppressions added for #3640 no longer worked after a system update.

This PR generalizes library versions and paths using wildcards, and consolidates two suppressions that were identical except for one line.

Sorry, should have done better on the first attempt.

### DIFF
--- a/t/valgrind/valgrind.supp
+++ b/t/valgrind/valgrind.supp
@@ -92,20 +92,7 @@
    <issue_3640>
    Memcheck:Leak
    match-leak-kinds: possible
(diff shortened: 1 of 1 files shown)


grondo (approved):
LGTM, Thanks!

I ran `t5000-valgrind.t` on TOSS 3 and TOSS 4 as well with no issues. (Just thought I'd double check since sometimes we use off-brand versions of libhwloc on these distros)

garlick:
Thanks!

PR #3748, opened 2021-06-28
libjob: refactor flux_jobspec1 class

garlick:
As discussed in #3741, here is a refactor of the `flux_jobspec1` class which
1) eliminates specutil (no longer needed)
2) factors out `libutil/jpath`
3) replaced redaction code duplicated in job-ingest and job-manager with `jpath_del()`
4) exposes internal json_t object from `flux_jobspec1` for in tree use via inclusion of `jobspec1_private.h`


src/common/libjob/jobspec1.c
```diff
@@ -342,35 +509,67 @@ int flux_jobspec1_check (flux_jobspec1_t *jobspec, flux_jobspec1_error_t *error)
...
+    char *value;
+    int rc = -1;
+
+    if (!(name = strdup (entry)))
         return -1;
+    if ((value = strchr (name, '=')))
+        *value++ = '\0';
+    if (!value || value == name + 1) {
+        errno = EINVAL;
+        goto done;
```
chu11, line 567: perhaps this is obvious and I'm just not seeing it, but what is the purpose of the `value == name + 1` corner case check?  Is for when the user doesn't specify the environment variable name? i.e. `=value`?  I think `name[0] == '\0'` might be more clear?

Also, should a user be able to specify an environment variable without a value?

chu11 (commented):
I like `jpath`.  Another potential cleanup is in `flux-module` w/ the `parse_json()` function.  There might be more but that's the one that I immediately could think of.

src/common/libjob/jobspec1.c
garlick: That was a weird way to write it, agreed :-/  Will fix.

> Also, should a user be able to specify an environment variable without a value?

I'll add a test that a user can set an environment var whose value is an empty string since `putenv ("name=")` or `setenv ("name", "")` seem to be allowed.

chu11 (approved):
your two fixes LGTM.  I'll let you decide if you wanna squash or not.  Or if you want to use `jpath` in `flux-module`.

grondo (approved):
LGTM, too.

garlick:
Just rebased on current master, squashed the two incremental changes with the `make jobspec1 standalone` commit, and converted `flux-module`.  OK to set MWP?

Thanks!

PR #3758, opened 2021-06-30
libflux/message: reset routes length on clear

chu11:
Problem: flux_msg_clear_route() did not reset routes_len back to
zero after all routes were destroyed.

Solution: Reset the length back to zero before returning to the user.


garlick (approved):
LGTM! Thanks.

Issue #3760, opened 2021-07-01
t3302-system-offline.t fails in the rpm buildfarm


garlick:
```
flux-mini: ERROR: [Errno 22] sign-munge-sign: Failed to access "/run/munge/munge.socket.2": No such file or directory
flux-mini: DEBUG: Traceback (most recent call last):
  File "/builddir/build/BUILD/flux-core-0.28.0/src/bindings/python/flux/util.py", line 201, in __call__
    main_func()
  File "/builddir/build/BUILD/flux-core-0.28.0/src/cmd/flux-mini.py", line 1615, in main
    args.func(args)
  File "/builddir/build/BUILD/flux-core-0.28.0/src/cmd/flux-mini.py", line 1321, in main
    jobid = self.submit(args)
  File "/builddir/build/BUILD/flux-core-0.28.0/src/cmd/flux-mini.py", line 673, in submit
    return JobID(self.submit_async(args, jobspec).get_id())
  File "/builddir/build/BUILD/flux-core-0.28.0/src/bindings/python/flux/job/submit.py", line 22, in get_id
    return submit_get_id(self)
  File "/builddir/build/BUILD/flux-core-0.28.0/src/bindings/python/flux/util.py", line 53, in func_wrapper
    raise EnvironmentError(error.errno, errmsg.decode("utf-8")) from None
OSError: [Errno 22] sign-munge-sign: Failed to access "/run/munge/munge.socket.2
```

garlick:
Normally munge signing of job requests is bypassed when the submitting user is the same as the instance owner.  It determines this by making a getattr request on `security.owner`.  In this test, the broker is not connected to its upstream peer and rejects the getattr request, and the submit code falls back to signing the request.  However, in the buildfarm munge is not running so this fails.

One idea is to use SO_PEERCRED to determine the instance owner on the other end of the socket.  This would probably be a bit less overhead too, since getting the attribute incurs a message round-trip delay.

grondo:
To reproduce, stop the `munge` service and rerun the test by hand (flux-core must be configured `--with-flux-security`):
```
$ ./t3302-system-offline.t -d -v
sharness: loading extensions from /home/grondo/git/f.git/t/sharness.d/01-setup.sh
sharness: loading extensions from /home/grondo/git/f.git/t/sharness.d/flux-sharness.sh
sharness: loading extensions from /home/grondo/git/f.git/t/sharness.d/01-setup.sh
[...]

PR #3761, opened 2021-07-01
libjob: improve method for determining instance owner

garlick:
This is proposed solution for #3760, in which the `t3302-system-offline.t` test fails when flux is built `--with-flux-security` and the munge daemon is not running.  Use of the `security.owner` broker attribute is replaced with a connector method based on SO_PEERCRED.    The existing `flux_opt_get()` interface is leveraged.

I tested this using @grondo's method from the issue to reproduce the failure and it works.

As a bonus, the first job submitted no longer has to make the extra synchronous RPC to the broker.

### DIFF
--- a/src/common/libflux/handle.h
+++ b/src/common/libflux/handle.h
@@ -61,7 +61,7 @@ enum {
     FLUX_POLLERR = 4,
 };
 
-/* Options for flux_setopt().
+/* Options for flux_opt_set().
  * (Connectors may define custom option names)
  */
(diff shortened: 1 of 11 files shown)


grondo (approved):
LGTM, thanks!

grondo:
Restarted CI checks because one builder exceeded the 30m timeout

garlick:
Let me see if I can bump the coverage just a bit.

grondo:
Nice work! Ready for MWP?

garlick:
Yep thanks!  I'll go ahead and set it.

PR #3763, opened 2021-07-02
add flux-job(1) manual page, plus minor fixes

garlick:
This was peeled off of pr #3741.

### DIFF
--- a/doc/man1/flux-job.rst
+++ b/doc/man1/flux-job.rst
@@ -1,0 +1,97 @@
+.. flux-help-include: true
+
+===========
+flux-job(1)
+===========
+
+
+SYNOPSIS
+========
+
+**flux** **job** **cancel** *id* [*message...*]
+
+**flux** **job** **cancelall** [*OPTIONS*] [*message...*]
+
+**flux** **job** **kill** [*--signal=SIG*] *id*
+
+**flux** **job** **killall** [*OPTIONS*]
+
+**flux** **job** **raise** [*OPTIONS*] *id* [*message...*]
+
+**flux** **job** **raiseall** [*OPTIONS*] *type* [*message...*]
+
+DESCRIPTION
+===========
+
+flux-job(1) performs various job related housekeeping functions.
+
+CANCEL
+======
+
+A single job may be canceled with ``flux job cancel``.
+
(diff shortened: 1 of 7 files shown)


doc/man3/index.rst
```diff
@@ -67,6 +67,7 @@ man3
    flux_shell_task_subprocess
```
chu11, line 67: commit message typo "add it t othe index"

doc/man1/flux-job.rst
```diff
@@ -0,0 +1,97 @@
...
+
+**-s, --severity=N**
+   Set exception severity.  The severity may range from 0=fatal to
+   7=least severe (default: 0).
+
+**-t, --type=TYPE**
+   Set exception type (default: cancel).
+
+Exceptions may be raised in bulk with ``flux job raiseall``, which requires a
+type (positional argument) and accepts the following options:
```
chu11, line 78: not for this issue, but the fact `raiseall` takes a positional `type` but `raise` takes it as an option seems sort of weird. perhaps an issue to be fixed?

chu11 (approved):
LGTM, one typo below and a comment.  One additional comment is that the subcommands you list in the manpage are probably for users, not system administrators?  The latter might be interested in a few other sub-commands like `urgency` .  Perhaps a `flux-job(8)` manpage as well?  Just an idea, perhaps just need an issue to remind us to do this.

doc/man1/flux-job.rst
garlick: I actually did change that, but while working on tests realized that the default severity is 7, so `cancel` would not be a reasonable default type.  What would be a reasonable default to go with 7?  We don't define any standard, non-fatal exception types.  If type is required, then the tiny @grondo in my brain says it should not be an option but a required positional argument.  We could change the default severity to 0 and go with a default type of `cancel`, but that's what `cancelall` is for.  So meh, I reverted the change and went with the status quo.

garlick:
I'll go ahead and set mwp.  Hope that's OK.

doc/man1/flux-job.rst
chu11: ahh ok, makes sense.  just seemed weird at a glance.

garlick: Yeah, to me too.

Issue #3754, opened 2021-06-29
libflux: [cleanup] consider breaking out flux_msg_fprint() to new source module


garlick:
Just an idea while reviewing `libflux/message.c`, which is getting quite large.

As a first step toward moving zsock functions over to librouter, would it make sense to define the message struct in `message_private.h` and then break `flux_msg_fprint()` and `flux_msg_typestr()` and their ilk out to a `message_fprint.c` or similar?

Edit: additional thought: are these functions needed in the public API?  Perhaps they only need to be internal to libflux.

chu11:
i was going to go full steam ahead on the big change, but I had missed that `flux_msg_fprint()` is only used by `libflux`.  So I think putting it into a "private" would be a good small step.

`flux_msg_typestr()` is used externally in a number of places though, although they could use `message_private` as well.  Its worth considering.

garlick:
Suggestion
* strip off `flux_` prefix => `msg_fprint()`
* define prototype in `message_private.h` (along with the message struct)
* move function def and helpers to `message_fprint.c`

Main point is I think it may be preferable to segment C code by functionality rather than by symbol scope, so avoid `message_private.c`.

chu11:
per discussion in #3765, we no longer want to make `flux_msg_fprint()` private, but we may still want to consider breaking `message.c` into separate files since its getting large.  We can consider that for later cleanup independent on this issue.  I'll close this issue.

garlick:
Yeah, the original idea in this PR was just to move that source to a different .c file, so maybe #3764 could just be retooled for that?  Sorry for being all over the map on this.

chu11:
i think a general cleanup where we create `message_private.h` and split up things into `message_fprint` and `message_route` is probably a cleanup we could do.

PR #3764, opened 2021-07-04
[WIP] libflux: make flux_msg_fprint() private

chu11:
Per #3754, begins minor step by creating `message_private.h`. 

Built on top of #3746.


chu11:
per discussion in #3765, we no longer want to make `flux_msg_fprint()` private

PR #3765, opened 2021-07-07
libflux: add time stamp to message trace

garlick:
Problem: message traces obtained by setting `FLUX_HANDLE_TRACE=1` do not contain timestamps, thus are not as helpful as they could be for performance investigations.

This adds timestamps that look like this (they are in seconds):
```
$ FLUX_HANDLE_TRACE=1 flux getattr size
> 0.00000
> attr.get
> {"name":"size"}
< 0.00024
< attr.get
< flags=topic,payload,route userid=5588 rolemask=owner errnum=0 matchtag=1
< {"value":"1","flags":1}
1
```
So this RPC took 240 microseconds.

### DIFF
--- a/src/common/libflux/message.c
+++ b/src/common/libflux/message.c
@@ -1582,7 +1582,7 @@ static void nodeid2str (uint32_t nodeid, char *buf, int buflen)
     assert (n < buflen);
 }
 
-void flux_msg_fprint (FILE *f, const flux_msg_t *msg)
+void flux_msg_fprint_ts (FILE *f, const flux_msg_t *msg, double timestamp)
 {
     int hops;
     const char *prefix;
@@ -1597,6 +1597,10 @@ void flux_msg_fprint (FILE *f, const flux_msg_t *msg)
         return;
     }
     prefix = msgtype_shortstr (msg->type);
+    /* Timestamp
+     */
+    if (timestamp >= 0.)
+        fprintf (f, "%s %.5f\n", prefix, timestamp);
     /* Topic (keepalive has none)
      */
     if (msg->topic)
@@ -1676,6 +1680,11 @@ void flux_msg_fprint (FILE *f, const flux_msg_t *msg)
         else
             fprintf (f, "malformed payload\n");
     }
+}
+
+void flux_msg_fprint (FILE *f, const flux_msg_t *msg)
+{
+    flux_msg_fprint_ts (f, msg, -1);
 }
 
 static zmsg_t *msg_to_zmsg (const flux_msg_t *msg)
--- a/src/common/libflux/handle.c
+++ b/src/common/libflux/handle.c
@@ -30,6 +30,7 @@
 #include "src/common/libutil/dirwalk.h"
 #include "src/common/libutil/aux.h"
 #include "src/common/libutil/errno_safe.h"
+#include "src/common/libutil/monotime.h"
 
 #include "handle.h"
 #include "reactor.h"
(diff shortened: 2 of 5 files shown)


chu11 (commented):
my only thought, is there going to be a circumstance that we'd want to "reset" the trace start?  B/c normally we can set/unset FLUX_HANDLE_TRACE.

Since `flux_msg_fprint()` doesn't have any public users, it's not a big deal to change the API here.  But could consider adding on top of/after #3764?

src/common/libflux/message.c
```diff
@@ -1597,6 +1597,9 @@ void flux_msg_fprint (FILE *f, const flux_msg_t *msg)
         return;
     }
     prefix = msgtype_shortstr (msg->type);
+    /* Timestamp
+     */
+    fprintf (f, "%s %.5f\n", prefix, timestamp);
```
grondo, line 1602: What about keeping the existing `flux_msg_fprint()` (since it is in the public API, and there isn't much use case for the timestamp when using the function publicly, at least that I can think of), and adding a private `flux_msg_fprint_ts()` which takes a timestamp, but doesn't print it if `timestamp < 0.`. Then `flux_msg_fprint()` could just be:
```c
void flux_msg_fprint (FILE *f, const flux_msg_t *msg)
{
    flux_msg_fprint_ts (f, msg, -1.);
}
```

`handle_trace()` could be the sole user of `flux_msg_fprint_ts()`?

grondo (approved):
A great improvement! I had one suggestion for keeping changes out of the libflux API if you are amenable.

Feel free to ignore if you think the timestamp will be useful in the public `flux_msg_fprint()`

garlick:
NP adding after #3764 @chu11.  I can rebase this after that goes in.

src/common/libflux/message.c
garlick: There don't seem to be any other users, but I'd be fine with this proposed change.

grondo:
> Since flux_msg_fprint() doesn't have any public users, it's not a big deal to change the API here.

`flux_msg_fprint()` is in the public API, so we can't know if there are any public users. I've used it in external test programs and such before. Frankly, I'd find it a pain to have to provide a timestamp when using it in testing circumstances.

src/common/libflux/message.c
grondo: How can you know there are no other users?

garlick: I withdraw my prior statement. :-)

garlick:
Re-pushed with @grondo's suggested API change.   Amended a sharness test to enables handle tracing for one RPC, since coverage seems very low (maybe that'll help).

chu11:
> flux_msg_fprint() is in the public API, so we can't know if there are any public users. I've used it in external test programs and such before. Frankly, I'd find it a pain to have to provide a timestamp when using it in testing circumstances.

My comment was in the context of a #3754 / #3764.  Perhaps we should not consider dropping `flux_msg_fprint()` publicly then.

garlick:
oops somehow I lost track of that, despite the title of the PR!   Yeah, let's keep it public, since clearly there's at least one user we want to stay on the good side of :-)

grondo:
> My comment was in the context of a #3754 / #3764. Perhaps we should consider not dropping flux_msg_fprint() publicly then.

Ah, sorry @chu11. I find it is useful for debugging to be able to dump a message, so I'd argue for keeping that particular function in the API.. however, I could probably be talked out of it if you guys have differing opinions.

chu11:
> ￼Edit: Yeah, I apologize I had lost track of that too. TBH, removing unnecessary functions from the public API is probably a good thing, and my use cases have always been internal to flux-core, so I'm on the fence if you guys feel strongly.

No strong feelings.  @garlick simply felt it was a good "starting point" to some of the bigger refactoring, and I didn't see any other users so figured, "why not" :-)

I'll close those issues / PRs.

garlick:
@chu11 did you want me to hold off merging this until you do other work in message.c or should we get this in?  (Again sorry for my confusion!)

chu11:
> did you want me to hold off merging this until you do other work in message.c or should we get this in? (Again sorry for my confusion!)

go ahead and push it in, it's no biggie :-)

PR #3766, opened 2021-07-07
add flag to bypass jobspec validation

garlick:
This adds a FLUX_JOB_NOVALIDATE flag that can be passed by the instance owner to the `job-ingest` module to bypass jobspec validation on a per-job basis, as opposed to reloading the `job-ingest` module with the `disable-validator` module option and disabling the validator for all jobs.

I originally added this as part of the `flux job exec` work to improve turnaround time for short lived jobs.  I broke it out for separate consideration here, since it's not really all _that_ helpful for the case it was written for, as the cost of validation is low once the python co-process is up and running.

Can we imagine this being useful for other things?  I'm fine if we prefer to dump it.

### DIFF
--- a/src/bindings/python/flux/job/submit.py
+++ b/src/bindings/python/flux/job/submit.py
@@ -29,6 +29,7 @@ def submit_async(
     waitable=False,
     debug=False,
     pre_signed=False,
+    novalidate=False,
 ):
     """Ask Flux to run a job, without waiting for a response
 
@@ -53,6 +54,10 @@ def submit_async(
     :param pre_signed: jobspec argument is already signed
         (default is False)
     :type pre_signed: bool
+    :param novalidate: jobspec does not need to be validated.
+        (default is False) novalidate=True is restricted to the
+        instance owner.
+    :type novalidate: bool
     :returns: a Flux Future object for obtaining the assigned jobid
     :rtype: Future
     """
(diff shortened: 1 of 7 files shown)


grondo:
Seems like this could be useful when submitting a job that purposefully violates jobspec requirements, e.g. when testing Fluxion or another new scheduler, etc. Maybe even in the testsuite to ensure the job shell or other jobspec consumer can tolerate a bad jobspec (I think we already have some tests that "load a permissive validator")

However, you can also just reload the job-ingest module with `disable-validator`, so you'd have to have a situation where you want most jobspecs validated but not all of them...

garlick:
Oh yeah, `disable-validator` (corrected in description above).

Yeah, I'm not sure this is worth it.  Nix it?

grondo:
I don't know, it seems like it is pretty minimal and could be useful.. I say we merge this. Doesn't seem like it would get in the way at all :thinking:

garlick:
OK, forced a push with python format fix, and also added a tiny bit of test coverage.

garlick:
Added a bit more testing and rebased on current master.

garlick:
Coverage revealed that my test to have a guest try to use the flag wasn't resulting in coverage of the error handling.  Upon closer look, the test was failing at the command instead of in job-ingest.  Fixed that, then found a bug in the job-ingest error handling and fixed.

Go coverage tool!

src/bindings/python/flux/job/submit.py
```diff
@@ -53,6 +54,10 @@ def submit_async(
     :param pre_signed: jobspec argument is already signed
         (default is False)
     :type pre_signed: bool
+    :param novalidate: jobspec does not need to be validated.
+        (default is False) Novalidate=True is restricted to the
```
grondo, line 58:
```suggestion
        (default is False) novalidate=True is restricted to the
```

grondo (approved):
LGTM! One minor documentation typo that was hardly worth mentioning.

src/bindings/python/flux/job/submit.py
garlick: I'll fix and set MWP if that's OK.

PR #3767, opened 2021-07-07
WIP: make flux job wait return highest shell exit code

garlick:
Peeled off of  #3741 (or rather a work in progress based on that).

This updates the job manager `wait` RPC to return the terminating event rather than the interpreted (success, error string) tuple.  Existing API interfaces are unchanged -  the interpretation moves from the job manager to libjob.  A new function, `flux_job_wait_get_exit_code()` interprets the terminating event into a reasonable proxy exit code for the job.

Then this gets used in the `flux job wait` command, which now returns the highest exit code of the tasks, which I think came up as a deficiency when this fast exec stuff was originally discussed.  The `flux_job_wait()` interface will be used by `flux job exec` if that goes in, but that's a different PR.

### DIFF
--- a/src/cmd/flux-job.c
+++ b/src/cmd/flux-job.c
@@ -2897,9 +2897,11 @@ int cmd_wait (optparse_t *p, int argc, char **argv)
     flux_t *h;
     int optindex = optparse_option_index (p);
     flux_future_t *f;
-    flux_jobid_t id = FLUX_JOBID_ANY;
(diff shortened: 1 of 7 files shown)


garlick:
rebased after merge of #3768

chu11 (approved):
nothing stood out to me

garlick:
Switched to WIP  - I was going to have "wait" stash a "flux::result" so the `flux_job_result()` accessors would work with it as well, but didn't quite get there before vacation.

PR #3773, opened 2021-07-13
libflux: use iovec-like array over zmsg

chu11:
Per discussion in #3617, this PR converts `libflux/message` to use `struct iovec` like arrays for sending/receiving messages instead of using `zmsg_t` and `zframe_t`.  This serves several purposes:

A) it continues the #3617 path to reduce dependence on `czmq`.  With this change, czmq dependence is now isolated to just `flux_msg_sendzsock()` and `flux_msg_recvzsock()`.  This will make it easy to migrate those functions over to `librouter`.

I have some experimental branches that in fact moved those functions over to `librouter` and the `message_private.h` file that is eventually created is quite small.  I have elected to not include those changes here and do that in another PR in the future.

B) Using `zmsg_t` and `zframe_t` made sense when `flux_msg_t` internally stored data via `zmsg_t`.  However, now that we store decoded information in `flux_msg_t`, using `zmsg_t` and `zframe_t` makes little sense.  We're building lists (need to malloc a list and malloc each node of a list) and we're often copying data in & out of the list.  By using the `iovec` array we can sometimes do a single alloc of an array and just point to the data already decoded in `flux_msg_t`.

Some notes on development (some of this is copied from #3617 notes):

- Most data is already decoded in `flux_msg_t`, so the `iovec` array is great to avoid copying of data.  The exception is the `proto` frame buffer.  I toyed with putting the proto frame buffer in `flux_msg_t`, but that broke some const-ness.  So instead I have the caller of `msg_to_iovec()` pass in a buffer for the proto-frame.  Perhaps not best.

- For functions that don't know the number of frames to expect (e.g. `flux_msg_recvzsock()`) i use `realloc` to extend the array.  non-optimal, but probably better on average than the regular mallocing of list nodes.  We could do games to limit the number of realloc calls, like alloc N frames and only realloc when you run out and add another N.  But I didn't investigate if that was that much better performance wise.

- zeromq recv data is stored inside a `zmq_msg_t` data structure. So we need to have a `zmq_msg_t` in memory when we set each iovec pointer. I ended up adding an aux parameter to the internal iovec struct like so.

```
+struct msg_iovec {
+    const void *data;
+    size_t size;
+    void *aux;
+};
```

I `malloc` space for a `zmq_msg_t` for each iovec entry. It introduces an extra malloc that I wish I could have avoided but in terms of total mallocs, we're even (assuming no `realloc` internal copying).  Note that I toyed with the idea of doing this trick to avoid the extra malloc:

```
struct msg_iovec {
    const void *data;
    size_t size;
    uint8_t aux[0];
};
```

but calling `realloc()` on a `zmq_msg_t` was dangerous b/c how they store their pointers.

Performance:

Over many experiments there was a clear performance benefit on throughput tests.  A sample:

Current master

```
Running throughput.py 2048 jobs per iteration, 5 iterations
throughput:     32.2 job/s (script:  32.0 job/s)
1: 64 seconds
throughput:     29.4 job/s (script:  29.3 job/s)
2: 134 seconds
throughput:     29.5 job/s (script:  29.3 job/s)
3: 204 seconds
throughput:     27.4 job/s (script:  27.2 job/s)
4: 280 seconds
throughput:     28.9 job/s (script:  28.7 job/s)
5: 352 seconds
Total of 352 seconds for everything
```

After the iovec work

```
Running throughput.py 2048 jobs per iteration, 5 iterations
throughput:     35.4 job/s (script:  35.2 job/s)
1: 59 seconds
throughput:     32.3 job/s (script:  32.1 job/s)
2: 123 seconds
throughput:     32.4 job/s (script:  32.2 job/s)
3: 187 seconds
throughput:     30.0 job/s (script:  29.8 job/s)
4: 255 seconds
throughput:     29.2 job/s (script:  29.0 job/s)
5: 326 seconds
Total of 326 seconds for everything
```

I've consistently seem 8-ish percent improvement.

### DIFF
--- a/src/common/libflux/message.c
+++ b/src/common/libflux/message.c
@@ -1685,7 +1685,7 @@ error:
 int flux_msg_sendzsock_ex (void *sock, const flux_msg_t *msg, bool nonblock)
 {
     void *handle;
-    int flags = ZFRAME_MORE;
+    int flags = ZMQ_SNDMORE;
     zmsg_t *zmsg = NULL;
     zframe_t *zf;
     size_t count = 0;
@@ -1701,19 +1701,20 @@ int flux_msg_sendzsock_ex (void *sock, const flux_msg_t *msg, bool nonblock)
         return -1;
 
     if (nonblock)
-        flags |= ZFRAME_DONTWAIT;
+        flags |= ZMQ_DONTWAIT;
 
     handle = zsock_resolve (sock);
     frames = zmsg_size (zmsg);
-    zf = zmsg_pop (zmsg);
+    zf = zmsg_first (zmsg);
     while (zf) {
         if (++count == frames)
-            flags &= ~ZFRAME_MORE;
-        if (zframe_send (&zf, handle, flags) < 0) {
-            zframe_destroy (&zf);
+            flags &= ~ZMQ_SNDMORE;
+        if (zmq_send (handle,
+                      zframe_data (zf),
+                      zframe_size (zf),
+                      flags) < 0)
             goto error;
-        }
-        zf = zmsg_pop (zmsg);
+        zf = zmsg_next (zmsg);
     }
     rc = 0;
 error:
@@ -1726,30 +1727,61 @@ int flux_msg_sendzsock (void *sock, const flux_msg_t *msg)
     return flux_msg_sendzsock_ex (sock, msg, false);
 }
 
+static int recvzsock_frame (void *handle, zmsg_t *zmsg)
+{
+    zframe_t *zf = NULL;
+    zmq_msg_t zmqmsg;
+    int rv = -1;
+
+    zmq_msg_init (&zmqmsg);
+    if (zmq_recvmsg (handle, &zmqmsg, 0) < 0)
+        goto error;
+    if (!(zf = zframe_new (zmq_msg_data (&zmqmsg), zmq_msg_size (&zmqmsg))))
+        goto error;
+    zframe_set_more (zf, zsock_rcvmore (handle));
+    if (zmsg_append (zmsg, &zf))
+        goto error;
+    rv = 0;
+error:
+    if (rv < 0)
+        zframe_destroy (&zf);
+    zmq_msg_close (&zmqmsg);
+    return rv;
+}
+
 flux_msg_t *flux_msg_recvzsock (void *sock)
 {
-    zmsg_t *zmsg;
+    void *handle;
+    zmsg_t *zmsg = NULL;
     flux_msg_t *msg;
+    flux_msg_t *rv = NULL;
 
     if (!sock) {
         errno = EINVAL;
         return NULL;
     }
-    if (!(zmsg = zmsg_recv (sock)))
-        return NULL;
-    if (!(msg = flux_msg_create_common ())) {
-        zmsg_destroy (&zmsg);
+    if (!(zmsg = zmsg_new ())) {
         errno = ENOMEM;
         return NULL;
     }
-    if (zmsg_to_msg (msg, zmsg) < 0) {
-        int save_errno = errno;
-        zmsg_destroy (&zmsg);
-        errno = save_errno;
-        return NULL;
+    handle = zsock_resolve (sock);
+    while (true) {
+        if (recvzsock_frame (handle, zmsg) < 0)
+            goto error;
+        if (!zsock_rcvmore (handle))
+            break;
     }
+
+    if (!(msg = flux_msg_create_common ())) {
+        errno = ENOMEM;
+        goto error;
+    }
+    if (zmsg_to_msg (msg, zmsg) < 0)
+        goto error;
+    rv = msg;
+error:
     zmsg_destroy (&zmsg);
-    return msg;
+    return rv;
 }
 
 int flux_msg_frames (const flux_msg_t *msg)
--- a/src/common/libflux/message.c
+++ b/src/common/libflux/message.c
@@ -478,10 +478,8 @@ static int zmsg_to_msg (flux_msg_t *msg, zmsg_t *zmsg)
             return -1;
         }
         msg->payload_size = zframe_size (zf);
-        if (!(msg->payload = malloc (msg->payload_size))) {
-            errno = ENOMEM;
+        if (!(msg->payload = malloc (msg->payload_size)))
             return -1;
-        }
         memcpy (msg->payload, zframe_data (zf), msg->payload_size);
         if (zf)
             zf = zmsg_next (zmsg);
@@ -1114,10 +1112,8 @@ int flux_msg_set_payload (flux_msg_t *msg, const void *buf, int size)
      */
     } else if (!(flags & FLUX_MSGFLAG_PAYLOAD) && (buf != NULL && size > 0)) {
         assert (!msg->payload);
-        if (!(msg->payload = malloc (size))) {
-            errno = ENOMEM;
+        if (!(msg->payload = malloc (size)))
             return -1;
-        }
         msg->payload_size = size;
         memcpy (msg->payload, buf, size);
         flags |= FLUX_MSGFLAG_PAYLOAD;
@@ -1410,7 +1406,7 @@ flux_msg_t *flux_msg_copy (const flux_msg_t *msg, bool payload)
         if (payload) {
             cpy->payload_size = msg->payload_size;
             if (!(cpy->payload = malloc (cpy->payload_size)))
-                goto nomem;
+                goto error;
             memcpy (cpy->payload, msg->payload, msg->payload_size);
         }
         else
--- a/src/common/libflux/message.c
+++ b/src/common/libflux/message.c
@@ -47,6 +47,8 @@
 
 #include "message.h"
 
+#define IOVECINCR 4
+
 struct flux_msg {
     // optional route list, if FLUX_MSGFLAG_ROUTE
     struct list_head routes;
@@ -85,6 +87,16 @@ struct flux_msg {
 struct route_id {
     struct list_node route_id_node;
     char id[0];                 /* variable length id stored at end of struct */
+};
+
+/* 'transport_data' is for any auxiliary transport data user may wish
+ * to associate with iovec, user is responsible to free/destroy the
+ * field
+ */
+struct msg_iovec {
+    const void *data;
+    size_t size;
+    void *transport_data;
 };
 
 /* PROTO consists of 4 byte prelude followed by a fixed length
@@ -393,7 +405,7 @@ int flux_msg_encode (const flux_msg_t *msg, void *buf, size_t size)
     return 0;
 }
 
-static void proto_get_u32 (uint8_t *data, int index, uint32_t *val)
+static void proto_get_u32 (const uint8_t *data, int index, uint32_t *val)
 {
     uint32_t x;
     int offset = PROTO_OFF_U32_ARRAY + index * 4;
@@ -416,18 +428,25 @@ static int msg_append_route (flux_msg_t *msg,
     return 0;
 }
 
-static int zmsg_to_msg (flux_msg_t *msg, zmsg_t *zmsg)
+static int iovec_to_msg (flux_msg_t *msg,
+                         struct msg_iovec *iov,
+                         int iovcnt)
 {
-    uint8_t *proto_data;
+    unsigned int index = 0;
+    const uint8_t *proto_data;
     size_t proto_size;
-    zframe_t *zf;
 
-    if (!(zf = zmsg_last (zmsg))) {
+    assert (msg);
+    assert (iov);
+
+    if (!iovcnt) {
         errno = EPROTO;
         return -1;
     }
-    proto_data = zframe_data (zf);
-    proto_size = zframe_size (zf);
+
+    /* proto frame is last frame */
+    proto_data = iov[iovcnt - 1].data;
+    proto_size = iov[iovcnt - 1].size;
     if (proto_size < PROTO_SIZE
         || proto_data[PROTO_OFF_MAGIC] != PROTO_MAGIC
         || proto_data[PROTO_OFF_VERSION] != PROTO_VERSION) {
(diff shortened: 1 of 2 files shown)


chu11:
hmm there was one LGTM warning.  I decided to comment out code with a note.

```
+        /* On first access index == 0 && iovcnt > 0 guaranteed
+         * Re-add check on if code changes. */
+        /* if (index == iovcnt) { */
+        /*     errno = EPROTO; */
+        /*     return -1; */
+        /* } */
```

garlick:
Poking at this...

I'm seeing a throughput bump when I realloc 3 iovecs at a time (see bbc998c6e13cc805467974254486f0172aac86c5).  I'm not sure I`m testing the same way as you (and running 3 times may not be sufficient):
```
test:
  for i in $(seq 1 3) ; do src/test/throughput.py -n 2000 -x /bin/true; done

Commit 252ea49 (this PR):
  number of jobs: 2000
  submit time:    0.515 s (3883.3 job/s)
  script runtime: 42.535s
  job runtime:    42.429s
  throughput:     47.1 job/s (script:  47.0 job/s)
  number of jobs: 2000
  submit time:    0.531 s (3764.8 job/s)
  script runtime: 48.775s
  job runtime:    48.608s
  throughput:     41.1 job/s (script:  41.0 job/s)
  number of jobs: 2000
  submit time:    0.573 s (3490.4 job/s)
  script runtime: 52.081s
  job runtime:    51.942s
  throughput:     38.5 job/s (script:  38.4 job/s)

WIth batch iov in bbc998c:
  number of jobs: 2000
  submit time:    0.523 s (3827.0 job/s)
  script runtime: 38.429s
  job runtime:    38.313s
  throughput:     52.2 job/s (script:  52.0 job/s)
  number of jobs: 2000
  submit time:    0.549 s (3642.9 job/s)
  script runtime: 48.772s
  job runtime:    48.625s
  throughput:     41.1 job/s (script:  41.0 job/s)
  number of jobs: 2000
  submit time:    0.556 s (3598.8 job/s)
  script runtime: 48.032s
  job runtime:    47.906s
  throughput:     41.7 job/s (script:  41.6 job/s)
```
My thought was that wasting a view iov slots wouldn't be very costly, but calling realloc potentially dozens of times when receiving a long-path message might be.

src/common/libflux/message.c
```diff
@@ -87,6 +87,14 @@ struct route_id {
     char id[0];                 /* variable length id stored at end of struct */
 };
 
+/* 'aux' for any auxiliary data user may wish to associate with iovec,
+ * user is responsible to free/destroy */
+struct msg_iovec {
+    const void *data;
+    size_t size;
+    void *aux;
```
garlick, line 95: The name `aux` is slightly overloaded here.  We already have a `struct aux_item` in the message struct, as well as the aux `uint32_t` fields in the PROTO block.  Man I am having a hard time thinking of an alternative that is not "data" though.  Maybe `transport_data`?

src/common/libflux/message.c
```diff
@@ -416,18 +424,26 @@ static int msg_append_route (flux_msg_t *msg,
...
 
-    if (!(zf = zmsg_last (zmsg))) {
+    assert (msg);
+    assert (iov);
+    assert (iovcnt >= 0);
+
+    if (!iovcnt) {
         errno = EPROTO;
         return -1;
     }
```
garlick, line 442: Isn't this redundant with the assertion above?

src/common/libflux/message.c
```diff
@@ -416,18 +424,26 @@ static int msg_append_route (flux_msg_t *msg,
...
-    uint8_t *proto_data;
+    unsigned int index = 0;
+    const uint8_t *proto_data;
     size_t proto_size;
-    zframe_t *zf;
 
-    if (!(zf = zmsg_last (zmsg))) {
+    assert (msg);
+    assert (iov);
+    assert (iovcnt >= 0);
```
garlick, line 437: Should this assertion on iovcnt be `>` not `>=` ?

src/common/libflux/message.c
```diff
@@ -444,50 +460,50 @@ static int zmsg_to_msg (flux_msg_t *msg, zmsg_t *zmsg)
...
     }
     if ((msg->flags & FLUX_MSGFLAG_TOPIC)) {
-        if (!zf) {
+        if (index == iovcnt) {
             errno = EPROTO;
             return -1;
         }
-        if (!(msg->topic = zframe_strdup (zf))) {
+        if (!(msg->topic = strndup ((char *)iov[index].data,
+                                    iov[index].size))) {
```
garlick, line 486: `strndup()` already sets errno.

src/common/libflux/message.c
```diff
@@ -444,50 +460,50 @@ static int zmsg_to_msg (flux_msg_t *msg, zmsg_t *zmsg)
...
+        if (index == iovcnt) {
             errno = EPROTO;
             return -1;
         }
-        msg->payload_size = zframe_size (zf);
+        msg->payload_size = iov[index].size;
         if (!(msg->payload = malloc (msg->payload_size))) {
             errno = ENOMEM;
             return -1;
         }
```
garlick, line 502: `malloc()` already sets errno (I realize you didn't touch this line :-)

It might be good to throw in EPROTO if the payload size is zero, otherwise `malloc()` would return NULL without errno set.   The existing code would at least set errno to _something_ but ENOMEM is the wrong one.

src/common/libflux/message.c
```diff
@@ -523,20 +537,22 @@ flux_msg_t *flux_msg_decode (const void *buf, size_t size)
             errno = EINVAL;
             goto error;
         }
-        if (!(zf = zframe_new (p, n)))
-            goto nomem;
-        if (zmsg_append (zmsg, &zf) < 0)
+
+        if (!(iov = realloc (iov, (iovcnt + 1) * sizeof (struct msg_iovec))))
             goto nomem;
```
garlick, line 544: `realloc()` already sets errno

Also this code leaks iov if the `realloc()` fails.  Assign result to tmp, check for NULL, then assign, like the other realloc in this commit.

src/common/libflux/message.c
```diff
@@ -523,20 +537,22 @@ flux_msg_t *flux_msg_decode (const void *buf, size_t size)
...
+    if (iovec_to_msg (msg, iov, iovcnt) < 0)
         goto error;
-    zmsg_destroy (&zmsg);
+    free (iov);
     return msg;
 nomem:
     errno = ENOMEM;
 error:
-    zmsg_destroy (&zmsg);
+    free (iov);
```
garlick, line 557: wrap with `ERRNO_SAFE_WRAP` or equivalent.

src/common/libflux/message.c
```diff
@@ -1641,88 +1657,98 @@ void flux_msg_fprint (FILE *f, const flux_msg_t *msg)
...
+    if ((frame_count = flux_msg_frames (msg)) < 0)
+        return -1;
+
+    assert (frame_count);
+
+    if (!(iov = malloc (frame_count * sizeof (*iov)))) {
         errno = ENOMEM;
-        goto error;
+        return -1;
     }
```
garlick, line 1680: `malloc()` already sets errno

src/common/libflux/message.c
```diff
@@ -1641,88 +1657,98 @@ void flux_msg_fprint (FILE *f, const flux_msg_t *msg)
...
+                      iov[count].size,
                       flags) < 0)
             goto error;
-        zf = zmsg_next (zmsg);
+        count++;
     }
     rc = 0;
 error:
-    zmsg_destroy (&zmsg);
+    free (iov);
```
garlick, line 1753: Prob need to wrap the free in `ERRNO_SAFE_WRAP` or equiv.

src/common/libflux/message.c
```diff
@@ -1731,46 +1757,38 @@ int flux_msg_sendzsock (void *sock, const flux_msg_t *msg)
...
+        if (!(tmp = realloc (iov, (iovcnt + 1) * sizeof (struct msg_iovec))))
+            goto error;
+        iov = tmp;
+        if (!(iov[iovcnt].aux = malloc (sizeof (zmq_msg_t))))
             goto error;
+        zmq_msg_init ((zmq_msg_t *)iov[iovcnt].aux);
+        if (zmq_recvmsg (handle, (zmq_msg_t *)iov[iovcnt].aux, 0) < 0)
+            goto error;
+        iov[iovcnt].data = zmq_msg_data ((zmq_msg_t *)iov[iovcnt].aux);
+        iov[iovcnt].size = zmq_msg_size ((zmq_msg_t *)iov[iovcnt].aux);
```
garlick: Just a suggestion, but it might be a bit cleaner to use a temporary variable for the zmq_msg_t that is eventually assigned to aux rather than all the casts.

src/common/libflux/message.c
```diff
@@ -1779,11 +1797,18 @@ flux_msg_t *flux_msg_recvzsock (void *sock)
...
 error:
-    zmsg_destroy (&zmsg);
+    if (iov) {
+        int i;
+        for (i = 0; i < iovcnt; i++) {
+            zmq_msg_close ((zmq_msg_t *)iov[i].aux);
+            free (iov[i].aux);
+        }
+        free (iov);
+    }
```
garlick: This block should preserve errno.

garlick (changes requested):
Really like where this is going!  It's nice to see the endpoint start to materialize.

Minor comments, mostly errno stuff.

src/common/libflux/message.c
chu11: yeah, this assert should be removed.  The `if (!iovcnt)` you mention above should just capture the 0 case.

src/common/libflux/message.c
chu11: doh!  that's bad of me.  I remember wanting to fix this line, but i guess i forgot.

src/common/libflux/message.c
chu11: perhaps we should just check for payload size > 0 in `flux_msg_set_payload()`, so we know its > 0 in all code later.

chu11: oh actually, `flux_msg_set_payload()` already ensures payload size > 0, so I think we're safe here.

chu11:
Re-pushed with fixes per above review.

I also no longer call `realloc` on every loop iteration.  I yoinked and then tweaked the @garlick patch mentioned above.

Not knowing the optimal number of `N` chunks to alloc each time you call `realloc`, I added a printf to `msg_to_iovec()` to output the number of frames in a message being sent.  I gathered that output and ran some stats.  The mean number of frames was 4.68, the mode 4, and the median 5.  The median of 5 was "close" to 4 though.

Granted this is a benchmark and not a real world case, I ended up picking N = 4 for the `realloc` chunks.  It performed the best in my throughput tests below, although it wasn't a huge / obvious win.  I'm not sure choosing N = 3, 4 or 5 matter much as just not calling `realloc` on every iteration is the real win.

realloc 1

```
Running throughput.py 2048 jobs per iteration, 5 iterations
throughput:     20.6 job/s (script:  20.6 job/s)
1: 100 seconds
throughput:     19.4 job/s (script:  19.3 job/s)
2: 206 seconds
throughput:     18.7 job/s (script:  18.6 job/s)
3: 316 seconds
throughput:     18.4 job/s (script:  18.4 job/s)
4: 428 seconds
throughput:     19.3 job/s (script:  19.2 job/s)
5: 535 seconds
Total of 535 seconds for everything
```

realloc 3

```
Running throughput.py 2048 jobs per iteration, 5 iterations
throughput:     20.6 job/s (script:  20.5 job/s)
1: 100 seconds
throughput:     19.4 job/s (script:  19.3 job/s)
2: 206 seconds
throughput:     19.3 job/s (script:  19.2 job/s)
3: 313 seconds
throughput:     20.1 job/s (script:  20.0 job/s)
4: 416 seconds
throughput:     20.2 job/s (script:  20.2 job/s)
5: 518 seconds
Total of 518 seconds for everything
```

realloc 4

```
Running throughput.py 2048 jobs per iteration, 5 iterations
throughput:     21.5 job/s (script:  21.4 job/s)
1: 96 seconds
throughput:     20.1 job/s (script:  20.0 job/s)
2: 198 seconds
throughput:     19.9 job/s (script:  19.9 job/s)
3: 302 seconds
throughput:     19.3 job/s (script:  19.2 job/s)
4: 408 seconds
throughput:     19.1 job/s (script:  19.0 job/s)
5: 516 seconds
Total of 516 seconds for everything
```

realloc 5

```
Running throughput.py 2048 jobs per iteration, 5 iterations
throughput:     20.8 job/s (script:  20.8 job/s)
1: 99 seconds
throughput:     19.5 job/s (script:  19.4 job/s)
2: 205 seconds
throughput:     19.5 job/s (script:  19.4 job/s)
3: 310 seconds
throughput:     19.7 job/s (script:  19.6 job/s)
4: 415 seconds
throughput:     19.1 job/s (script:  19.1 job/s)
5: 522 seconds
Total of 522 seconds for everything
```

src/common/libflux/message.c
```diff
@@ -1727,47 +1751,43 @@ int flux_msg_sendzsock (void *sock, const flux_msg_t *msg)
...
+    /* N.B. we need to store a zmq_msg_t for each iovec entry so that
+     * the memory is available during the call to iovec_to_msg().  We
+     * use the msg_iovec's "transport_data" field to store the entry
+     * and then clear/free it later.
+     */
     handle = zsock_resolve (sock);
     while (true) {
-        if (recvzsock_frame (handle, zmsg) < 0)
+        struct msg_iovec *tmp;
+        zmq_msg_t *transport_data;
```
garlick, line 1781: Shorter name for local variable?   Maybe `zmsg`?

src/common/libflux/message.c
```diff
@@ -1727,47 +1751,43 @@ int flux_msg_sendzsock (void *sock, const flux_msg_t *msg)
...
+            goto error;
+        iov = tmp;
+        if (!(transport_data = malloc (sizeof (zmq_msg_t))))
+            goto error;
+        zmq_msg_init (transport_data);
+        if (zmq_recvmsg (handle, transport_data, 0) < 0) {
+            zmq_msg_close (transport_data);
+            free (transport_data);
             goto error;
+        }
```
garlick: I think you need to preserve errno in this block

garlick (approved):
Seems like this is close.  Couple fairly nitpicky comments inline.

src/common/libflux/message.c
chu11: hehe, I intentionally tried to avoid `zmsg` b/c it could imply `zmsg_t`, which we just got rid of.  perhaps just `msgdata` or something like that.

chu11:
re-pushed for fixes per review comments above

garlick (approved):
LGTM!

PR #3778, opened 2021-07-16
Allow io encode/decode of binary data

### LINKED ISSUE #3775: flux-shell: io decoding/encoding errors with zeros/random stdout
Kind of a pathological case, but also kind of like fuzzing the job-shell. Not really sure how important handling this edge case is or what is going wrong.

EDIT: Maybe in the decode case, the issue is that the [output does not have a trailing newline](https://github.com/flux-framework/flux-core/blob/4321d8ef735af86c7070eeba0b2b972d7c88eca0/src/common/libeventlog/eventlog.c#L166)?

```
herbein1@lassen11 ~/workspace/2021-08-gmd-OOM
❯ flux mini run -N1 -n1 dd if=/dev/zero bs=1K count=1                                      19:27:16 ()
1+0 records in
1+0 records out
flux-job: eventlog_entry_decode: Invalid argument

(1) herbein1@lassen11 ~/workspace/2021-08-gmd-OOM
❯ flux mini run -N1 -n1 dd if=/dev/urandom bs=1K count=1                                   19:27:25 ()
1+0 records in
1+0 records out
1024 bytes (1.0 kB) copied, 3.4292e-05 s, 29.9 MB/s
0.127s: flux-shell[0]: ERROR: ioencode: Invalid argument
0.127s: flux-shell[0]: ERROR: write stdout task 0: Invalid argument
0.127s: flux-shell[0]: ERROR: ioencode: Invalid argument
0.127s: flux-shell[0]: ERROR: write stdout task 0: Invalid argument
0.127s: flux-shell[0]: ERROR: ioencode: Invalid argument
0.127s: flux-shell[0]: ERROR: write stdout task 0: Invalid argument
0.127s: flux-shell[0]: ERROR: ioencode: Invalid argument
0.127s: flux-shell[0]: ERROR: write stdout task 0: Invalid argument
0.127s: flux-shell[0]: ERROR: ioencode: Invalid argument
0.127s: flux-shell[0]: ERROR: write stdout task 0: Invalid argument
0.127s: flux-shell[0]: ERROR: ioencode: File exists
0.127s: flux-shell[0]: ERROR: write stdout task 0: File exists
0.127s: flux-shell[0]: stderr: flux-shell: ERROR: ioencode: Invalid argument
0.127s: flux-shell[0]: stderr: flux-shell: ERROR: write stdout task 0: Invalid argument
0.127s: flux-shell[0]: stderr: flux-shell: ERROR: ioencode: Invalid argument
0.127s: flux-shell[0]: stderr: flux-shell: ERROR: write stdout task 0: Invalid argument
0.127s: flux-shell[0]: stderr: flux-shell: ERROR: ioencode: Invalid argument
0.127s: flux-shell[0]: stderr: flux-shell: ERROR: write stdout task 0: Invalid argument
0.127s: flux-shell[0]: stderr: flux-shell: ERROR: ioencode: Invalid argument
0.127s: flux-shell[0]: stderr: flux-shell: ERROR: write stdout task 0: Invalid argument
0.128s: flux-shell[0]: stderr: flux-shell: ERROR: ioencode: Invalid argument
0.128s: flux-shell[0]: stderr: flux-shell: ERROR: write stdout task 0: Invalid argument
0.128s: flux-shell[0]: stderr: flux-shell: ERROR: ioencode: File exists
0.128s: flux-shell[0]: stderr: flux-shell: ERROR: write stdout task 0: File exists
```

Happens at much smaller sizes too:
```
❯ flux mini run -N1 -n1 dd if=/dev/urandom bs=32 count=1                                   19:30:50 ()
1+0 records in
1+0 records out
32 bytes (32 B) copied, 3.0001e-05 s, 1.1 MB/s
0.127s: flux-shell[0]: stderr: flux-shell: ERROR: ioencode: Invalid argument
0.128s: flux-shell[0]: stderr: flux-shell: ERROR: write stdout task 0: Invalid argument
0.127s: flux-shell[0]: ERROR: ioencode: Invalid argument
0.127s: flux-shell[0]: ERROR: write stdout task 0: Invalid argument
```

grondo:
As described in #3775, if a job input or output stream contains data that cannot be encoded in utf8, `ioencode()` and `iodecode()` currently fail.

Along with a minor adjustment in RFC 24 (flux-framework/rfc#296), this PR falls back to encoding I/O eventlog data events with base64 if `json_pack()` fails on the original data. The new encoding is added to the context object so that the decoder knows that the data is in base64.

Instead of using the libsodium base64 routines, a base64 implementation is added from CCAN, since we want to get away from using libsodium at this point. Other base64 users in the codebase were not updated, that is left for a future exercise.

This is a WIP for now since it uses an as-yet unapproved change to RFC 24.

Fixes #3775

### DIFF
--- a/src/common/libioencode/ioencode.c
+++ b/src/common/libioencode/ioencode.c
@@ -20,8 +20,32 @@
 #include <jansson.h>
 
 #include "src/common/libutil/macros.h"
+#include "src/common/libccan/ccan/base64/base64.h"
 
 #include "ioencode.h"
+
+
+static json_t *data_encode_base64 (const char *stream,
+                                   const char *rank,
+                                   const char *data,
+                                   int len)
+{
+    ssize_t n;
+    json_t *o = NULL;
+    char *dest = NULL;
+    size_t destlen = base64_encoded_length (len) + 1; /* +1 for NUL */
+
+    if ((dest = calloc (1, destlen))
+        && (n = base64_encode (dest, destlen, data, len)) >= 0) {
+            o = json_pack ("{s:s s:s s:s s:s#}",
+                           "stream", stream,
+                           "rank", rank,
+                           "encoding", "base64",
+                           "data", dest, n);
+    }
+    free (dest);
+    return o;
+}
 
 json_t *ioencode (const char *stream,
                   const char *rank,
@@ -46,8 +70,15 @@ json_t *ioencode (const char *stream,
         if (!(o = json_pack ("{s:s s:s s:s#}",
                              "stream", stream,
                              "rank", rank,
-                             "data", data, len)))
-            goto error;
+                             "data", data, len))) {
+            /*
+             *  json_pack() may fail if there are bytes that cannot
+             *   be encoded as UTF-8 - try again, this time
+             *   encoding the data in base64:
+             */
+            if (!(o = data_encode_base64 (stream, rank, data, len)))
+                goto error;
+        }
     }
     else {
         if (!(o = json_pack ("{s:s s:s}",
@@ -64,6 +95,22 @@ error:
     return rv;
 }
 
+static int decode_data_base64 (char *src,
+                               size_t srclen,
+                               char **datap,
+                               size_t *lenp)
+{
+    ssize_t rc;
+    size_t size = base64_decoded_length (srclen);
+    if (!(*datap = calloc (1, size)))
+        return -1;
+    if ((rc = base64_decode (*datap, size, src, srclen)) < 0)
+        return -1;
+    *lenp = rc;
+    return 0;
+}
+
+
 int iodecode (json_t *o,
               const char **streamp,
               const char **rankp,
@@ -73,8 +120,10 @@ int iodecode (json_t *o,
 {
     const char *stream;
     const char *rank;
-    char *bin_data = NULL;
+    const char *encoding = NULL;
     size_t bin_len = 0;
+    char *data = NULL;
+    size_t len = 0;
     int eof = 0;
     bool has_data = false;
     bool has_eof = false;
@@ -85,11 +134,12 @@ int iodecode (json_t *o,
         return -1;
     }
 
-    if (json_unpack (o, "{s:s s:s}",
+    if (json_unpack (o, "{s:s s:s s?s}",
                      "stream", &stream,
-                     "rank", &rank) < 0)
+                     "rank", &rank,
+                     "encoding", &encoding) < 0)
         goto cleanup;
-    if (json_unpack (o, "{s:s%}", "data", &bin_data, &bin_len) == 0) {
+    if (json_unpack (o, "{s:s%}", "data", &data, &len) == 0) {
         has_data = true;
     }
     if (json_unpack (o, "{s:b}", "eof", &eof) == 0)
@@ -106,11 +156,17 @@ int iodecode (json_t *o,
         (*rankp) = rank;
     if (datap) {
         *datap = NULL;
-        if (bin_data) {
-            if (!(*datap = malloc (bin_len)))
-                goto cleanup;
-            memcpy (*datap, bin_data, bin_len);
-            bin_data = NULL;
+        if (data) {
+            if (encoding && strcmp (encoding, "base64") == 0) {
+                if (decode_data_base64 (data, len, datap, &bin_len) < 0)
+                    goto cleanup;
+            }
+            else {
+                bin_len = len;
+                if (!(*datap = malloc (bin_len)))
+                    goto cleanup;
+                memcpy (*datap, data, bin_len);
+            }
         }
     }
     if (lenp)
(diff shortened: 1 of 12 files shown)


src/common/libioencode/ioencode.c
```diff
@@ -20,9 +20,33 @@
 #include <jansson.h>
```
chu11, line 20: commit message typo "butes" -> "bytes"

src/common/libioencode/ioencode.c
```diff
@@ -20,9 +20,33 @@
...
+                                   const char *rank,
+                                   const char *data,
+                                   int len)
+{
+    ssize_t n;
+    json_t *o = NULL;
+    char *dest = NULL;
+    size_t destlen = base64_encoded_length (len);
+
+    if ((dest = calloc (1, destlen))
```
chu11, line 38: can call `malloc` instead of `calloc`?  Since all data will be overwritten in the `base64_encode()` below.

Also on the decode path below in two places.

src/common/libioencode/ioencode.c
```diff
@@ -64,6 +95,22 @@ json_t *ioencode (const char *stream,
...
+    if (!(*datap = calloc (1, size)))
+        return -1;
+    if ((rc = base64_decode (*datap, size, src, srclen)) < 0)
+        return -1;
+    *lenp = rc;
+    return 0;
+}
+
+
 int iodecode (json_t *o,
```
chu11, line 114: consideration b/c of this change, should `char **data` be `void **data` now?

src/common/libioencode/ioencode.c
grondo: This is true for `base64_encoded_length()`, but `base64_decoded_length()` returns the _minimum_ buffer size for decode, so some bytes would otherwise be uninitialized and it seemed safer to set those to zero.

I can change the calloc() here if you feel that is important.

src/common/libioencode/ioencode.c
grondo: I chose not to change the function prototypes. Even `base64_encode()` takes `char *src` as the buffer to encode, so I'd be inclined to leave things as is unless we hit an issue (though maybe you see an issue I don't?)

src/common/libioencode/ioencode.c
grondo: Actually I just realized this comment from `base64_encoded_length()`:
```c
/**
 * base64_encoded_length - Calculate encode buffer length
 * @param srclen the size of the data to be encoded
 * @note add 1 to this to get null-termination
 * @return Buffer length required for encode
 */
```

So this code needs to be fixed to add 1 to destlen. Then, to ensure NUL termination, it is probably safest to use `calloc` here just in case `base64_encode()` doesn't write the NUL.

grondo:
Forced a push with squashed fixup for off-by-one `destlen` and the commit typo caught by @chu11.

src/common/libioencode/ioencode.c
chu11: nope, was just for consideration.

src/common/libioencode/ioencode.c
chu11: > I can change the calloc() here if you feel that is important. 

Nah, if it does need to an initialize, then its fine.  I thought the entire buffer would be overwritten right after.  I think one decode path does a `calloc` then `memcpy`, so that could probably be a `malloc` instead.

grondo: Ah, you're right. That was the existing code (and the common code path), so I should probably change that back to a `malloc()`. It was changed inadvertently when I refactored some code.

grondo: Ok, I forced a push changing the useless use of `calloc()` back to `malloc()`. Thanks!

garlick:
Not really up to speed on this, but wanted to ask about this comment?

> since we want to get away from using libsodium at this point.

Were you referring to #3617?  libsodium is not subject to the FIPS self-test penalty at load time, as far as I know.

grondo:
Oh, yeah sorry! I totally misspoke on that one. If someone wants they could pull the ccan/base64 out of this PR and replace with the equivalent libsodium. Sorry about that! If not, I can fix after vacation.

garlick:
Thinking about it a bit, the CCAN version may be preferable since it's standalone and I think base64 is the only thing we are using out of libsodium (except transitively through libzmq).  We might find a libsodium dependency to be a liability at some point, for example if the FIPS patches that exist for it find their way into RHEL.

garlick (approved):
This looks good to me.  I just approved and set MWP on the RFC update.

@chu11, any objection to this going in as is?  I think @grondo addressed all your comments.

The CCAN base64 seems like the right way to go.

chu11:
@garlick, LGTM!

PR #3782, opened 2021-07-16
Globally drop unnecessary 0MQ includes

### LINKED ISSUE #3624: drop unnecessary 0MQ includes
As noted in ~~#3312~~ (#3617), we should try to minimize/localize our dependence on 0MQ libraries.

Problem: `src/bindings/lua/zmsg-lua.h` and  `src/cmd/builtin/proxy.c` include `[c]zmq.h` unnecessarily.

chu11:
Problem: czmq.h and/or zmq.h was included in a number of locations
that was unnecessary.

Solution: Remove the unnecessary includes.  In some cases, replace with
include of czmq_containers.h or other appropriate headers that czmq/zmq
happened to include.

Fixes #3624

### DIFF
--- a/src/bindings/lua/zmsg-lua.h
+++ b/src/bindings/lua/zmsg-lua.h
@@ -12,7 +12,6 @@
 #define HAVE_ZMSG_LUA_H
 
 #include <lua.h>
-#include <zmq.h>
 #include <jansson.h>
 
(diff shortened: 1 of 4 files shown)


grondo (approved):
Nice cleanup! Thanks!

Issue #3772, opened 2021-07-13
python job validator: only enforces canonical jobspec, not V1


SteVwonder:
In the current `Validator` class, there is following code: 

```
class Validator(ValidatorPlugin):
    def __init__(self, parser):
        self.require_version = 1
        parser.add_argument(
            "--require-version",
            metavar="V",
            type=int,
            default=1,
            help="Require jobspec version V",
        )

    def configure(self, args):
        self.require_version = args.require_version
        if args.require_version < 1:
            raise ValueError(
                f"Required jobspec version too low: {args.require_version} is < 1"
            )
        elif args.require_version > 1:
            raise ValueError(
                f"Required jobspec version too high: {args.require_version} is > 1"
            )

    def validate(self, args):
        validate_jobspec(json.dumps(args.jobspec))
```

There are two problems with this setup.  First, `args.require_version` is not passed down through to `validate_jobspec` so the jobspec is only ever checked that it is canonical jobspec, and V1 is never enforced.  The second (potential issue) is that once the previous issue is fixed, it isn't clear that it will be possible to relax the validator to not require V1 and just validate canonical jobspec since the `args.require_version` defaults to `1` and the `configure` code prevents any other value.  Maybe we need to allow `0` as the `canonical` version?

grondo:
How about either allowing `--require-version=any` or add a new option `--no-require-version`?

Thanks for catching this dumb error!

SteVwonder:
#3784 fixed this. Closing.

PR #3784, opened 2021-07-17
python: fix version requirement in jobspec validator plugin

grondo:
This PR addresses issue #3772 -- the `jobspec` `flux job-validator` plugin was supposed to be validating V1 jobspec only by default (and with `--require-version=1`), but unfortunately was completely ignoring this option.

This PR fixes the bug in the plugin, and also adds support for `--require-version=any` for disabling the jobspec version check (which is currently the default)

Some tests needed to updated now that V1 is the default requirement.


SteVwonder (approved):
LGTM! Thanks @grondo!

SteVwonder:
CC @milroy since this might affect his K8s stuff going forward.

Issue #3776, opened 2021-07-14
drop unnecessary CFLAGS / LIBS in Makefiles


chu11:
Remove unnecessary ZMQ_CFLAGS / ZMQ_LIBS includes throughout Makefiles.  It appears some were just lazily cut & pasted.  Some are probably lingering from before `libczmqcontainers` existed.

Related work to #3617 
partially related to #3624

chu11:
also found cases of LIBUUID_CFLAGS / LIBUUID_LIBS and HWLOC_CFLAGS

chu11:
While you technically do not need to add ZMQ_CFLAGS / ZMQ_LIBS or various other CFLAGS/LIBS, the fact that many times `libflux-core` and `libutil` (and maybe others) are linked, maybe you should.

perhaps splitting up `libutil` or other convenience libs may be prudent.  Perhaps this should come after #3617 rather than before.  Work TBD.

PR #3785, opened 2021-07-18
Consistently use CFLAGS / LIBS in Makefiles

chu11:
Per #3776, I found that ZMQ_CFLAGS and ZMQ_LIBS were used excessively throughout the Makefiles.  Libraries /modules that have no need to link to zmq (e.g. `liboptparse`) were linked to it.  Some of these were likely old additions that were no longer needed after refactoring with `libczmqcontainers`.   Others were likely lazy cut & paste issues.   As I went through, realized that many ZMQ_LIBS additions were unnecessary b/c other libs (e.g. `libflux-internal`) already links to it.  So cleaned those up too.

Along the way, noticed more unnecessary inclusions and just decided to go nuts, cleaning up LIBUUID, LIBSODIUM, HWLOC_CFLAGS, LIBDL, LIBRT, and JANSSON macros as well.  Some cases were stupid, where variables like `test_ldadd` included `FOO_LIBS`, and every test binary that used `test_ldadd` also linked to `FOO_LIBS`.

I may have missed some, but minimally its much more cleaned up.

### DIFF
--- a/src/broker/Makefile.am
+++ b/src/broker/Makefile.am
@@ -72,7 +72,8 @@ flux_broker_LDADD = \
 	$(ZMQ_LIBS) \
 	$(LIBUUID_LIBS) \
 	$(LIBSODIUM_LIBS) \
-	$(PMIX_LIBS)
+	$(PMIX_LIBS) \
+	$(LIBDL)
 
 flux_broker_LDFLAGS =
 
--- a/src/broker/Makefile.am
+++ b/src/broker/Makefile.am
@@ -69,6 +69,7 @@ flux_broker_LDADD = \
 	$(top_builddir)/src/common/libflux-internal.la \
 	$(top_builddir)/src/common/libflux-optparse.la \
 	$(ZMQ_LIBS) \
+	$(LIBUUID_LIBS) \
 	$(PMIX_LIBS)
 
 flux_broker_LDFLAGS =
--- a/src/broker/Makefile.am
+++ b/src/broker/Makefile.am
@@ -12,6 +12,7 @@ AM_CPPFLAGS = \
 	$(PMIX_CFLAGS) \
 	$(ZMQ_CFLAGS) \
 	$(LIBUUID_CFLAGS) \
+	$(LIBSODIUM_CFLAGS) \
 	$(VALGRIND_CFLAGS)
 
 fluxcmd_PROGRAMS = flux-broker
@@ -70,6 +71,7 @@ flux_broker_LDADD = \
 	$(top_builddir)/src/common/libflux-optparse.la \
 	$(ZMQ_LIBS) \
 	$(LIBUUID_LIBS) \
+	$(LIBSODIUM_LIBS) \
 	$(PMIX_LIBS)
 
 flux_broker_LDFLAGS =
--- a/src/broker/Makefile.am
+++ b/src/broker/Makefile.am
@@ -68,6 +68,7 @@ flux_broker_LDADD = \
 	$(top_builddir)/src/common/libpmi/libpmi_client.la \
 	$(top_builddir)/src/common/libflux-internal.la \
 	$(top_builddir)/src/common/libflux-optparse.la \
+	$(ZMQ_LIBS) \
 	$(PMIX_LIBS)
 
 flux_broker_LDFLAGS =
@@ -87,6 +88,7 @@ test_ldadd = \
 	$(top_builddir)/src/common/libpmi/libpmi_client.la \
 	$(top_builddir)/src/common/libflux-internal.la \
 	$(top_builddir)/src/common/libtap/libtap.la \
+	$(ZMQ_LIBS) \
 	$(PMIX_LIBS)
 
 test_ldflags = \
--- a/src/broker/Makefile.am
+++ b/src/broker/Makefile.am
@@ -13,6 +13,7 @@ AM_CPPFLAGS = \
 	$(ZMQ_CFLAGS) \
 	$(LIBUUID_CFLAGS) \
 	$(LIBSODIUM_CFLAGS) \
+	$(JANSSON_CFLAGS) \
 	$(VALGRIND_CFLAGS)
 
 fluxcmd_PROGRAMS = flux-broker
@@ -72,6 +73,7 @@ flux_broker_LDADD = \
 	$(ZMQ_LIBS) \
 	$(LIBUUID_LIBS) \
 	$(LIBSODIUM_LIBS) \
+	$(JANSSON_LIBS) \
 	$(PMIX_LIBS) \
 	$(LIBDL)
 
@@ -93,6 +95,7 @@ test_ldadd = \
 	$(top_builddir)/src/common/libflux-internal.la \
 	$(top_builddir)/src/common/libtap/libtap.la \
 	$(ZMQ_LIBS) \
+	$(JANSSON_LIBS) \
 	$(PMIX_LIBS)
 
 test_ldflags = \
--- a/src/cmd/Makefile.am
+++ b/src/cmd/Makefile.am
@@ -10,7 +10,6 @@ AM_CPPFLAGS = \
 	-I$(top_srcdir)/src/include \
 	-I$(top_builddir)/src/common/libflux \
 	$(FLUX_SECURITY_CFLAGS) \
-	$(LIBUUID_CFLAGS) \
 	$(LIBSODIUM_CFLAGS) \
 	$(HWLOC_CFLAGS) \
 	$(PMIX_CFLAGS)
@@ -22,7 +21,6 @@ fluxcmd_ldadd = \
 	$(top_builddir)/src/common/libflux-core.la \
 	$(top_builddir)/src/common/libflux-optparse.la \
 	$(FLUX_SECURITY_LIBS) \
-	$(LIBUUID_LIBS) \
 	$(LIBPTHREAD) \
 	$(LIBDL) \
 	$(HWLOC_LIBS) \
--- a/src/cmd/Makefile.am
+++ b/src/cmd/Makefile.am
@@ -21,7 +21,6 @@ fluxcmd_ldadd = \
 	$(top_builddir)/src/common/libflux-optparse.la \
 	$(FLUX_SECURITY_LIBS) \
 	$(LIBPTHREAD) \
-	$(LIBDL) \
 	$(HWLOC_LIBS)
 
 LDADD = $(fluxcmd_ldadd)
--- a/src/cmd/Makefile.am
+++ b/src/cmd/Makefile.am
@@ -11,7 +11,8 @@ AM_CPPFLAGS = \
 	-I$(top_builddir)/src/common/libflux \
 	$(FLUX_SECURITY_CFLAGS) \
 	$(HWLOC_CFLAGS) \
-	$(PMIX_CFLAGS)
+	$(PMIX_CFLAGS) \
+	$(JANSSON_CFLAGS)
 
 fluxcmd_ldadd = \
 	$(top_builddir)/src/common/libkvs/libkvs.la \
@@ -21,7 +22,8 @@ fluxcmd_ldadd = \
 	$(top_builddir)/src/common/libflux-optparse.la \
 	$(FLUX_SECURITY_LIBS) \
 	$(LIBPTHREAD) \
-	$(HWLOC_LIBS)
+	$(HWLOC_LIBS) \
+	$(JANSSON_LIBS)
 
 LDADD = $(fluxcmd_ldadd)
 
--- a/src/cmd/Makefile.am
+++ b/src/cmd/Makefile.am
@@ -23,7 +23,6 @@ fluxcmd_ldadd = \
 	$(top_builddir)/src/common/libflux-core.la \
 	$(top_builddir)/src/common/libflux-optparse.la \
 	$(FLUX_SECURITY_LIBS) \
-	$(ZMQ_LIBS) \
 	$(LIBUUID_LIBS) \
 	$(LIBPTHREAD) \
 	$(LIBDL) \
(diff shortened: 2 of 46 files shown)


garlick:
As noted in #3617, a _module_ picks up all of the objects in a libtool convenience library in its LIBADD, and thus requires all external dependencies of the convenience library.  However, an _executable_ only pulls in objects that resolve symbols in the executable, and would only require the external dependencies of those objects.

If we go with this cleanup, then our commands will possibly end up linked against external libraries they don't use.

OTOH if we go the other way and drop external libraries from the convenience library's LIBADD, those eight(?) libraries will have to be added to every module's LIBADD, and a subset to every executable's LDADD.  A lot of repetition.  Grr.

Maybe we could go with this cleanup, then see if we can reduce libflux-internal's external dependencies by breaking out a few key bits?   Obviously the key one is going to be `$(ZMQ_LIBS)`.  We may be able to get rid of `$(LIBSODIUM_LIBS)` after #3785 gets merged.   Maybe `$(LIBUUID_LIBS)`.

garlick:
Just did an experiment: build master and run
```
$ cd src/cmd/.libs
$ for i in *; do echo -n "$i "; ldd $i|wc -l; done
flux 33
flux-event 30
flux-exec 30
flux-job 30
flux-keygen 27
flux-kvs 30
flux-logger 29
flux-module 30
flux-ping 30
flux-queue 30
flux-R 11
flux-start 30
flux-terminus 30
```
I then repeated that for this branch, and got the same results.  So this PR does no harm relative to current master.

chu11:
> Maybe we could go with this cleanup, then see if we can reduce libflux-internal's external dependencies by breaking out a few key bits? Obviously the key one is going to be $(ZMQ_LIBS). We may be able to get rid of $(LIBSODIUM_LIBS) after #3785 gets merged. Maybe $(LIBUUID_LIBS).

This is sort of my feeling, it's step one on the path to something better.  Many of the CFLAGS/LIBS were cut & paste errors or forgotten removals after refactoring.  It made it hard to break out ZMQ_LIBS b/c *everything* was linking to it.

src/connectors/shmem/Makefile.am
```diff
@@ -23,5 +23,4 @@ shmem_la_LDFLAGS = -module $(san_ld_zdef_flag) \
 shmem_la_LIBADD = \
 	$(top_builddir)/src/common/libflux-internal.la \
 	$(top_builddir)/src/common/libflux-core.la \
-	$(ZMQ_LIBS) \
```
garlick, line 26: shmem connector uses the `zsock` class, so this is still required.  Also it looks like `$(ZMQ_CFLAGS)` is missing so that should be added.

src/test/Makefile.am
```diff
@@ -8,8 +8,7 @@ AM_LDFLAGS = \
 AM_CPPFLAGS = \
 	-I$(top_srcdir) \
 	-I$(top_srcdir)/src/include \
-	-I$(top_builddir)/src/common/libflux \
-	$(ZMQ_CFLAGS)
+	-I$(top_builddir)/src/common/libflux
```
garlick, line 11: There are no more c targets in src/test, so AM_CPPFLAGS, AM_LDFLAGS, AM_CFLAGS, and LDADD can all go away.

src/common/libjob/Makefile.am
```diff
@@ -56,8 +56,7 @@ test_ldadd = \
 	$(LIBPTHREAD) \
 	$(LIBRT) \
         $(FLUX_SECURITY_LIBS) \
-	$(LIBDL) \
-	$(LIBSODIUM_LIBS)
```
garlick, line 64: libjob/sign_none.c uses libsodium, so this should remain.

Another option would be to convert it to use the CCAN base64 just merged.

src/cmd/Makefile.am
```diff
@@ -21,7 +21,6 @@ fluxcmd_ldadd = \
 	$(top_builddir)/src/common/libflux-optparse.la \
 	$(FLUX_SECURITY_LIBS) \
 	$(LIBPTHREAD) \
-	$(LIBDL) \
```
garlick, line 29: flux-module.c includes dlcfn.h but doesn't appear to use it.  Maybe drop that too?

t/Makefile.am
```diff
@@ -454,204 +451,167 @@ mpi_mpich_basic_simple_LDADD = $(MPI_CLDFLAGS) $(test_ldadd)
...
 request_rpc_stream_CPPFLAGS = $(test_cppflags)
-request_rpc_stream_LDADD = \
-	$(test_ldadd) $(LIBDL)
+request_rpc_stream_LDADD = $(test_ldadd)
 
 module_parent_la_SOURCES = module/parent.c
 module_parent_la_CPPFLAGS = $(test_cppflags)
 module_parent_la_LDFLAGS = $(fluxmod_ldflags) -module -rpath /nowher
-module_parent_la_LIBADD = \
-	$(test_ldadd) $(LIBDL)
```
garlick, line 555: t/module/parent.c calls `dlopen()` so this should remain.

src/common/libjob/Makefile.am
```diff
@@ -54,7 +54,6 @@ test_ldadd = \
         $(top_builddir)/src/common/libtap/libtap.la \
```
garlick, line 54: Commit message:   "macro was more than necessary'.  Missing "used"?

That was commit 6234eb20fae7ced9ada4c22ddf76a7d6230c7afa

src/common/libjob/Makefile.am
```diff
@@ -52,7 +52,6 @@ test_ldadd = \
         $(top_builddir)/src/common/libflux/libflux.la \
         $(top_builddir)/src/common/libflux-internal.la \
         $(top_builddir)/src/common/libtap/libtap.la \
-	$(JANSSON_LIBS) \
```
garlick, line 59: I think we need this one.  jansson is used in several .c files in libjob.

src/common/libkvs/Makefile.am
```diff
@@ -61,7 +61,6 @@ test_ldadd = \
 	$(top_builddir)/src/common/libflux/libflux.la \
 	$(top_builddir)/src/common/libflux-internal.la \
 	$(top_builddir)/src/common/libtap/libtap.la \
-	$(JANSSON_LIBS) \
```
garlick, line 68: jansson is used in libkvs/kvs_commit.c so presumably we should keep this and add missing `$(JANSSON_CFLAGS)`.

The jansson header is included in kvs.c, but I think it could be removed.

src/common/librlist/Makefile.am
```diff
@@ -26,7 +26,6 @@ test_ldadd = \
 	$(top_builddir)/src/common/libtap/libtap.la \
 	$(top_builddir)/src/common/libflux-internal.la \
 	$(LIBPTHREAD) \
-	$(JANSSON_LIBS) \
```
garlick, line 31: librlist uses jansson extensively.  Should add the missing `$(JANSSON_CFLAGS)`.

src/modules/job-exec/Makefile.am
```diff
@@ -58,7 +58,7 @@ test_ldadd = \
 	$(top_builddir)/src/common/libtap/libtap.la \
 	$(top_builddir)/src/common/libflux-internal.la \
 	$(top_builddir)/src/common/libflux-core.la \
-	$(LIBPTHREAD) $(JANSSON_LIBS)
+	$(LIBPTHREAD)
```
garlick, line 61: job-exec uses jansson directly, but is missing a `$(JANSSON_LIBS)` in its LIBADD.

src/modules/job-manager/Makefile.am
```diff
@@ -112,7 +112,7 @@ test_ldadd = \
 	$(top_builddir)/src/common/libjob/libjob.la \
 	$(top_builddir)/src/common/libflux-internal.la \
 	$(top_builddir)/src/common/libflux-core.la \
-	$(LIBPTHREAD) $(JANSSON_LIBS)
```
garlick: Same here, $(JANSSON_LIBS) Is missing from LIBADD.

Also, what's the deal with `$(fluxmod_libadd)`?  All the modules add it, but it doesn't seem to be defined anywhere?  (possibly I'm missing something).

garlick (changes requested):
I found a few places where I didn't think removal was appropriate - where a module/command/whatever was using the dependency directly, as opposed to needing it because it linked with a convenience library that uses it.

chu11:
> Edit: by the way I'm assuming that if a library is used directly, it should have both the CFLAGS and the LIBS def for it in its makefile, even if it uses a convenience library that also uses that library. In that case the LIBS might be redundant, but at least if the convenience library stops using the library, other things won't need to be updated.

So I did the opposite.  B/c `libflux-core` and `libflux-internal` effectively link to everything, and a very large swath of stuff always ldadd `libflux-core` and `libflux-internal`, the redundancy sort of existed everywhere.  Which is why I may have removed it in more places than you originally thought I should have.  And (unsurprisingly) we're not consistent.  There are locations that we should probably add ZMQ_LIBS (or some other libs) for consistency.

Still thinks we should add the XXX_LIBS back?

> The CFLAGS is needed regardless, so the include file can be found if it was installed to a non-standard prefix.

Agreed, I (atleast intended) to do this, but could have made a mistake or two (i looks like I did given above).

src/common/libjob/Makefile.am
chu11: working on #3787, perhaps that should go in first before this

garlick:
> Still thinks we should add the XXX_LIBS back when a lib/module/executable uses it?

If we need the CFLAGS anyway, and the libs that the convenience libs depend on may be changing, then it seems to me to make the most sense to add/keep both where the dependency is used directly.

chu11:
Another comment.  Lots of ldadds (especially test_ldadds) are through `common/libflux-core.la`.  This is different than using `common/libflux/libflux.la`, the latter which does not yet link to ZMQ_LIBS and a bunch of other libs.

Another approach is to convert internal ldadds to always use the internal libs (e.g. `librouter/librouter.la`) instead of the wrapper that already does all the linking (`libflux-internal.la`).

It would force us to put ZMQ_LIBS, LIBUUID_LIBS, etc. into the right places.

chu11:
> If we need the CFLAGS anyway, and the libs that the convenience libs depend on may be changing, then it seems to me to make the most sense to add/keep both where the dependency is used directly.

Ok, I'll make another run through it, since this approach will be different.

garlick:
> Lots of ldadds (especially test_ldadds) are through common/libflux-core.la. This is different than using common/libflux/libflux.la

I think it's good to link against `libflux-core.la` where needed.  The only places where there may be an issue is in subdirectories of `src/common`, when iteratively working on something that becomes part of `libflux-core.la`.  You want the test to use the latest version of the component under development, but there is no makefile rule to rebuild `libflux-core.la` at that level.  There IMHO it makes sense to reference the component directly.

Are there other reasons to go the other way?  (I'm not thinking of them right now)

chu11:
> Are there other reasons to go the other way? (I'm not thinking of them right now)

Its just to reduce redundancy of linking, and to some extent force us to only link what we need to link.

garlick:
> Its just to reduce redundancy of linking, and to some extent force us to only link what we need to link.

If it decreases the `make check` time I'm all for it.  Otherwise, meh?  (My opinion, and not a cherished one :-)

src/modules/job-manager/Makefile.am
chu11: > Also, what's the deal with $(fluxmod_libadd)? All the modules add it, but it doesn't seem to be defined anywhere? (possibly I'm missing something).

I'm going to take a guess.  Perhaps `fluxmod_ldflags` was slowly copied around and cutters/pasters assumed `fluxmod_libadd` or `fluxmod_ldadd` was also defined?  Apparently this issue has existed as far back as `0.1.0`,  I'll do a bit of cleanup with this too.

chu11:
re-pushed addressing the comments / discussion above.

- 3 random cleanup commits before the beefier stuff
- generally use the rule to add/use CFLAGS / LIBS when code uses stuff from that lib.  So in addition to the removals I had before, there are now commits to add CFLAGS / LIBS in places.
- adjust commit messages as a result

src/broker/Makefile.am
```diff
@@ -93,14 +95,16 @@ test_ldadd = \
...
 	$(PMIX_LIBS)
 
 test_ldflags = \
 	-no-install
 
 test_cppflags = \
         -I$(top_srcdir)/src/common/libtap \
-        $(AM_CPPFLAGS)
+        $(AM_CPPFLAGS) \
+	$(JANSSON_CFLAGS)
```
garlick, line 107: Inconsistent whitespace here (use tabs I think).

Also `$(JANSSON_CFLAGS)` is also being added to `$(AM_CPPFLAGS)`, which is included here, so this is a dup.

src/modules/job-list/Makefile.am
```diff
@@ -9,7 +9,8 @@ AM_CPPFLAGS = \
 	-I$(top_srcdir) \
 	-I$(top_srcdir)/src/include \
 	-I$(top_builddir)/src/common/libflux \
-	$(FLUX_SECURITY_CFLAGS) $(YAMLCPP_CFLAGS)
+	$(FLUX_SECURITY_CFLAGS) $(YAMLCPP_CFLAGS) \
```
garlick: What's YAMLCPP_CFLAGS doing here?  Leftover from days gone by?  It's not defined.  Maybe drop that from here and other places it was cut and paste (job-archive, job-info, job-ingest, and kvs-watch).

garlick (approved):
Just spotted a couple minor things but generally looking good!

src/modules/job-list/Makefile.am
chu11: oh, haha!  I just assumed we were using it :P  i'll yoink it

chu11:
re-pushed per comments above, there's one extra commit in there to deal w/ the YAMLCPP_CFLAGS.  @garlick wanna take a quick skim before I label MWP.

garlick (approved):
Looks good after the small update.  Thanks!

Issue #3783, opened 2021-07-16
Python FluxExecutor: add `attach` method for job IDs


jameshcorbett:
As part of the [PSI-J project](https://github.com/ExaWorks/psi-j-python) @andre-merzky requested that the `FluxExecutor` add an ``attach(jobid)`` method that would return a Future like ``FluxExecutor.submit(*args)``, the difference being that in ``attach`` the executor does not have a Jobspec to submit---it can just use the provided job ID.

The use-case is that the Python interpreter exits, and you fire up a new interpreter and want to re-connect to some existing job IDs.

PR #3790, opened 2021-07-21
python: allow executor to attach to jobs

jameshcorbett:
As mentioned in #3783, FluxExecutor instances only work with jobs that they submit themselves, not jobs that are submitted by other instances or by other mechanisms. This PR adds a `FluxExecutor.attach(jobid)` method that takes a job ID and returns a future. The futures behaves identically to those that are returned by `FluxExecutor.submit(jobspec)`. The only difference is that the job ID of the `attach` futures is available immediately, so `attach_future.jobid(0)` should never fail.

Tagging @andre-merzky .

### DIFF
--- a/src/bindings/python/flux/job/executor.py
+++ b/src/bindings/python/flux/job/executor.py
@@ -26,6 +26,22 @@ from flux.job.submit import submit_async, submit_get_id
 from flux.job.event import event_watch_async, JobException, MAIN_EVENTS
 
 
+_SubmitPackage = collections.namedtuple(
+    "_SubmitPackage", ["submit_args", "submit_kwargs", "future"]
+)
+
+
+class _AttachPackage:  # pylint: disable=too-few-public-methods
+    """Namedtuple-esque class. Constructor sets jobid on future."""
+
+    __slots__ = ("jobid", "future")
+
+    def __init__(self, jobid, fut):
+        self.jobid = jobid
+        self.future = fut
+        self.future._set_jobid(jobid)  # pylint: disable=protected-access
+
+
 class _FluxExecutorThread(threading.Thread):
     """Thread that submits jobs to Flux and waits for event updates.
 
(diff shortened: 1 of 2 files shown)


src/bindings/python/flux/job/executor.py
```diff
@@ -514,14 +568,36 @@ def submit(self, *args, **kwargs):
...
+    def attach(self, jobid):
+        """Attach a ``FluxExecutorFuture`` to an existing job ID and return it.
+
+        Returned futures will behave identically to futures returned by the
+        ``FluxExecutor.submit`` method.
+
+        This method is primarily useful for monitoring jobs that have been
+        submitted through other mechanisms.
+
+        :param jobid: jobid to attach to.
```
SteVwonder, line 582: Comment from a super quick pass between meetings: can this be (does it have to be) a `flux.job.JobID`?  Either way, can you add the type to the function signature?

SteVwonder (commented):
Thanks @jameshcorbett for putting this together so quickly!

src/bindings/python/flux/job/executor.py
jameshcorbett: No restrictions on type, but if the type isn't accepted by `flux.job.event_watch_async` the future will error out. I added `:type jobid: int` since I think that's the most general type accepted.

andre-merzky:
Thanks @jameshcorbett , this looks great to me!

SteVwonder (approved):
Thanks @jameshcorbett!

jameshcorbett:
@SteVwonder could you re-approve? Sorry, I wasn't thinking and rebased with master (no other changes though).

grondo (approved):
LGTM too. @jameshcorbett, I can set MWP if this is ready to merge.

jameshcorbett:
Yup, thanks @grondo!

Issue #3623, opened 2021-04-29
don't use zmq_strerror() in libutil/log.c


garlick:
As noted in #3617, we should try to minimize/localize our dependence on 0MQ libraries.

Problem: PR #3620 got rid of `zmq_strerror()` in libflux/flog.c but there is another user in `libutil/log.c` that should be similarly converted to `strerror()`.

chu11:
aww crud, this is fixed via #3628, I put the wrong issue number in the commit: 53c86c7ae6ed4dfa2d4fa7c777c614874ff7f552

PR #3797, opened 2021-07-24
Move libzmq dependent functions out of libflux-core into libzmqutil

chu11:
It's been a massive journey to try and de-link `libzmq` from `libflux-core` (#3620 #3622, #3623, #3624, #3697, #3701, #3742, #3746, #3773, #3776), but hopefully this is the end :-)

This PR will move the few `libzmq` dependent functions out of `libflux-core` and into a new convenience library `libzmqutil`.  The functions are:

```
flux_msg_sendzsock() / fluxmsg_sendzsock_ex()
flux_msg_recvzsock()
flux_zmq_watcher_create()
flux_zmq_watcher_get_zsock()
```

Only those few dependent modules/binaries (`broker`, `shmem`, `libtestutil`) that need them will use `libzmqutil` and link to `libzmq`.  We have to create a few private helper headers along the way (`message_private.h`, `reactor_private.h`) that exports a few macros, helper functions, structs, etc.

You'll notice two `fixup` commits in this PR.  They both rename functions.

```
flux_msg_sendzsock -> zmqutil_msg_sendzsock
flux_msg_recvzsock -> zmqutil_msg_recvzsock
flux_zmq_watcher_create -> zmqutil_zmq_watcher_create
flux_zmq_watcher_get_zsock -> zmqutil_zmq_watcher_get_zsock
```

renaming the functions is of course completely optional.  I went back and forth on it.  So it's just fixup patches for now, we can squash or remove based on discussion.  Or perhaps folks don't like how I renamed them, I struggled with that as well.  I didn't want to prefix anything `zmq_` to avoid potential confusion with it being a `libzmq` function.  But `zmqutil_` is not the prettiest prefix either.

Performance tests on `fluke` (multiple runs showed consistent results similar to these)

current master

```
Running throughput.py 1024 jobs per iteration, 1 iterations
throughput:     22.9 job/s (script:  22.7 job/s)
1: 46 seconds

Running throughput.py 2048 jobs per iteration, 1 iterations
throughput:     21.9 job/s (script:  21.8 job/s)
1: 94 seconds
```

this PR branch

```
>src/cmd/flux start ./throughput_loop.sh 1024 1
Running throughput.py 1024 jobs per iteration, 1 iterations
throughput:     25.3 job/s (script:  25.1 job/s)
1: 41 seconds


Running throughput.py 2048 jobs per iteration, 1 iterations
throughput:     24.0 job/s (script:  23.9 job/s)
1: 85 seconds
```

generally speaking, around a 10% improvement across many runs


garlick:
Whew!  Nice.

Quick comment on function naming.  What about:

```
flux_msg_sendzsock -> zmqutil_msg_send
flux_msg_recvzsock -> zmqutil_msg_recvz
flux_zmq_watcher_create -> zmqutil_watcher_create
flux_zmq_watcher_get_zsock -> zmqutil_watcher_get_zsock
```
just reducing the redundancy a bit.  Same with the file names - could probably drop the `zmqutil_` prefix since they are already in a zmqutil directory.

chu11:
re-pushed per comments.  Comments suggested rename of functions were liked, so removed the squashed fixes.

garlick:
I checked `src/shell/.libs/flux-shell` and it is still linked against libzmq and friends, but I can't for the life of me figure out how!

chu11:
@garlick I'm not seeing it on a TOSS3 cluster.  Are you ldd-ing `.libs/flux-shell` or `.libs/lt-flux-shell`?  b/c the former will link to the `/usr/lib64/libflux-core.so`, which still has the czmq dependency.

garlick:
Ah! that was the problem.  I'm not getting an `lt-flux-shell` in my build (Ubuntu 20.04 LTS) but after installing this version over the top of the old one I had installed, no more libzmq!
```
ldd flux-shell
	linux-vdso.so.1 (0x00007fffba7e3000)
	libflux-core.so.2 => /usr/local/lib/libflux-core.so.2 (0x00007f0321102000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f03210df000)
	libflux-optparse.so.1 => /usr/local/lib/libflux-optparse.so.1 (0x00007f03210c9000)
	liblua5.1.so.0 => /lib/x86_64-linux-gnu/liblua5.1.so.0 (0x00007f0321098000)
	libhwloc.so.15 => /lib/x86_64-linux-gnu/libhwloc.so.15 (0x00007f0321047000)
	libjansson.so.4 => /lib/x86_64-linux-gnu/libjansson.so.4 (0x00007f0321038000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f0320ee7000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f0320cf5000)
	libuuid.so.1 => /lib/x86_64-linux-gnu/libuuid.so.1 (0x00007f0320cec000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f0320ce6000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f03211ee000)
	libudev.so.1 => /lib/x86_64-linux-gnu/libudev.so.1 (0x00007f0320cb9000)
	libltdl.so.7 => /lib/x86_64-linux-gnu/libltdl.so.7 (0x00007f0320cae000)
```
Yay!

chu11:
@garlick i think you need to run `src/shell/flux-shell` one time for it to generate the `lt-flux-shell`.

garlick:
> i think you need to run src/shell/flux-shell one time

I think that's dependent on the libtool version?  My system doesn't produce executables with the `lt-` prefix.  They are the same name as the original.  This is libtool 2.4.6-14.

src/common/libflux/message.c
```diff
@@ -427,9 +384,9 @@ static int msg_append_route (flux_msg_t *msg,
     return 0;
 }
 
-static int iovec_to_msg (flux_msg_t *msg,
-                         struct msg_iovec *iov,
-                         int iovcnt)
+int flux_iovec_to_msg (flux_msg_t *msg,
```
garlick, line 387: Why do these need to have the `flux_` prefix?  This will make them available publicly.

grondo:
> I think that's dependent on the libtool version? 

It is dependent on whether "fast-install" mode is enabled in libtool. I guess on some distros fast-install mode isn't needed, so libtool turns it off by default? I could have sworn it was enabled on my system (Ubuntu 18.04), but now I see it is set to `needless`:
```$ grep enable_fast_install config.status
enable_fast_install='needless'
fast_install=$enable_fast_install
```

I did verify that on CentOS 7 (and thus presumably TOSS 3), fast-install is enabled by default:
```bash-4.2$ grep enable_fast_install config.status
enable_fast_install='yes'
fast_install=$enable_fast_install

```

src/common/libflux/reactor_private.h
```diff
@@ -0,0 +1,67 @@
...
+    unsigned int errflag:1;
+};
+
+struct flux_watcher {
+    flux_reactor_t *r;
+    flux_watcher_f fn;
+    void *arg;
+    struct flux_watcher_ops *ops;
+    void *data;
+};
```
garlick, line 33: It's a bit unfortunate that we need to expose these things outside of reactor.c, given that we theoretically are providing an API to implement custom watchers.  I guess it's because ev_zmq was implemented as a pure libev watcher, and libev isn't exposed directly.   We _might_ be able to reimplement this in terms of the exported functions, but I'm thinking that could be hairy and time consuming, and this works.

I guess it's for the greater good, and it's still private to flux-core :-)

garlick (approved):
Generally very pleased with how this turned out!   Just the one question below.

src/common/libflux/message.c
chu11: This was done b/c `libflux-core.so` doesn't export any `flux_` prefixed functions, so the iovec functions weren't available in some places.  But now that I think about it, any code that needs the iovec functions could ldadd `src/common/libflux/libflux.la` instead of `src/common/libflx-core.la`.  Let me see how that works out.

chu11: Well, this experiment didn't go as well as I thought it would.  There's so many symbols to resolve, and a nice chunk of "what to ldadd first" to ensure dependencies get resolved accordingly.

```
diff --git a/src/common/librouter/Makefile.am b/src/common/librouter/Makefile.am
index 756f5bb..4dcdd41 100644
--- a/src/common/librouter/Makefile.am
+++ b/src/common/librouter/Makefile.am
@@ -63,9 +63,15 @@ test_ldadd = \
         $(top_builddir)/src/common/librouter/librouter.la \
         $(top_builddir)/src/common/libtestutil/libtestutil.la \
         $(top_builddir)/src/common/libflux-internal.la \
-        $(top_builddir)/src/common/libflux-core.la \
-        $(top_builddir)/src/common/libtap/libtap.la \
+        $(top_builddir)/src/common/libflux/libflux.la \
+        $(top_builddir)/src/common/liblsd/liblsd.la \
+        $(top_builddir)/src/common/libev/libev.la \
+        $(top_builddir)/src/common/libutil/libutil.la \
+        $(top_builddir)/src/common/libccan/libccan.la \
+        $(top_builddir)/src/common/libtomlc99/libtomlc99.la \
         $(top_builddir)/src/common/libzmqutil/libzmqutil.la \
+        $(top_builddir)/src/common/libczmqcontainers/libczmqcontainers.la \
+        $(top_builddir)/src/common/libtap/libtap.la \
        $(ZMQ_LIBS)
 
 test_cppflags = \
```

worse yet I'm hitting segfaults in sharness tests.  I haven't been able to figure it out yet, although the backtrace of the segfault points to an unknown function.  A symbol collision wouldn't surprise me given what's going on above.

Hmmm.  Let me mull over this tonight and maybe something obviously simpler can come to mind.

src/common/libflux/reactor_private.h
chu11: i didn't think about the idea of implementing via exposed functions.  But yeah, the fact `libev` wasn't public did create problems.

src/common/libflux/message.c
chu11: i can't think of an obviously clean solution here.  We don't want to move iovec into yet another convenience lib.  Would it be worthwhile to prefix the functions to something to indicate they are private and make sure they are accessible in the `libflux-core.map` file?  Like prefix them with `fprivate_` or something instead?

Although unintended, we get away with this on the reactor side b/c we static inline those two simple functions.  We'd have that problem there too if the functions were not inlined.

garlick: Can we just add `libflux/libflux.la` right after `libflux-core.la` in the LDADD rules?  I didn't try to work through any tests, but the following patch resulted in a working broker and shmem connector for me:
```diff
diff --git a/src/broker/Makefile.am b/src/broker/Makefile.am
index f25cd2eaf..3c02fd731 100644
--- a/src/broker/Makefile.am
+++ b/src/broker/Makefile.am
@@ -67,6 +67,7 @@ flux_broker_LDADD = \
        $(builddir)/libbroker.la \
        $(top_builddir)/src/common/libflux-core.la \
        $(top_builddir)/src/common/libzmqutil/libzmqutil.la \
+       $(top_builddir)/src/common/libflux/libflux.la \
        $(top_builddir)/src/common/libpmi/libpmi_client.la \
        $(top_builddir)/src/common/libflux-internal.la \
        $(top_builddir)/src/common/libflux-optparse.la \
diff --git a/src/common/libflux/message.c b/src/common/libflux/message.c
index 51e402d2e..c6f0db869 100644
--- a/src/common/libflux/message.c
+++ b/src/common/libflux/message.c
@@ -384,7 +384,7 @@ static int msg_append_route (flux_msg_t *msg,
     return 0;
 }
 
-int flux_iovec_to_msg (flux_msg_t *msg,
+int iovec_to_msg (flux_msg_t *msg,
                        struct msg_iovec *iov,
                        int iovcnt)
 {
@@ -507,7 +507,7 @@ flux_msg_t *flux_msg_decode (const void *buf, size_t size)
         iovcnt++;
         p += n;
     }
-    if (flux_iovec_to_msg (msg, iov, iovcnt) < 0)
+    if (iovec_to_msg (msg, iov, iovcnt) < 0)
         goto error;
     free (iov);
     return msg;
@@ -1622,7 +1622,7 @@ void flux_msg_fprint (FILE *f, const flux_msg_t *msg)
     flux_msg_fprint_ts (f, msg, -1);
 }
 
-int flux_msg_to_iovec (const flux_msg_t *msg,
+int msg_to_iovec (const flux_msg_t *msg,
                        uint8_t *proto,
                        int proto_len,
                        struct msg_iovec **iovp,
diff --git a/src/common/libflux/message_private.h b/src/common/libflux/message_private.h
index 7923b8c10..8872d3082 100644
--- a/src/common/libflux/message_private.h
+++ b/src/common/libflux/message_private.h
@@ -61,11 +61,11 @@ struct msg_iovec {
     void *transport_data;
 };
 
-int flux_iovec_to_msg (flux_msg_t *msg,
+int iovec_to_msg (flux_msg_t *msg,
                        struct msg_iovec *iov,
                        int iovcnt);
 
-int flux_msg_to_iovec (const flux_msg_t *msg,
+int msg_to_iovec (const flux_msg_t *msg,
                        uint8_t *proto,
                        int proto_len,
                        struct msg_iovec **iovp,
diff --git a/src/common/libzmqutil/msg_zsock.c b/src/common/libzmqutil/msg_zsock.c
index 42e5b9c54..ad71f51fe 100644
--- a/src/common/libzmqutil/msg_zsock.c
+++ b/src/common/libzmqutil/msg_zsock.c
@@ -40,7 +40,7 @@ int zmqutil_msg_send_ex (void *sock, const flux_msg_t *msg, bool nonblock)
         return -1;
     }
 
-    if (flux_msg_to_iovec (msg, proto, PROTO_SIZE, &iov, &iovcnt) < 0)
+    if (msg_to_iovec (msg, proto, PROTO_SIZE, &iov, &iovcnt) < 0)
         goto error;
 
     if (nonblock)
@@ -117,7 +117,7 @@ flux_msg_t *zmqutil_msg_recv (void *sock)
 
     if (!(msg = flux_msg_create (FLUX_MSGTYPE_ANY)))
         goto error;
-    if (flux_iovec_to_msg (msg, iov, iovcnt) < 0)
+    if (iovec_to_msg (msg, iov, iovcnt) < 0)
         goto error;
     rv = msg;
 error:
diff --git a/src/connectors/shmem/Makefile.am b/src/connectors/shmem/Makefile.am
index 82ade229b..1c64c6553 100644
--- a/src/connectors/shmem/Makefile.am
+++ b/src/connectors/shmem/Makefile.am
@@ -22,6 +22,7 @@ shmem_la_LDFLAGS = -module $(san_ld_zdef_flag) \
 shmem_la_LIBADD = \
        $(top_builddir)/src/common/libflux-internal.la \
        $(top_builddir)/src/common/libflux-core.la \
+       $(top_builddir)/src/common/libflux/libflux.la \
        $(top_builddir)/src/common/libzmqutil/libzmqutil.la \
        $(ZMQ_LIBS)

Comment (chu11) [reply src/common/libflux/message.c:387]: I thought of that last night but it felt sort of wrong.  I guess the linker just does magic to resolve and make sure all symbols are good?  I'll give it a shot and see how it turns out.

Comment (garlick) [reply src/common/libflux/message.c:387]: There's plenty of bad libtool juju to contend with, so IMHO when the magic works, go with it. :-)

File: src/common/libflux/reactor_private.h (hunk 2 above)
Comment (garlick) [reply src/common/libflux/reactor_private.h:33]: Not worth the effort at this time IMHO.

File: src/broker/module.c [hunk 3]
```diff
@@ -23,6 +23,7 @@
 #endif
```
garlick, line 23: Commit message for c9096f74abbaa31d852de3e4c88a960d1f1d62f5

This might be the commit that people will find in the history when they ask "why isn't libzmqutil part of libflux"?  So it might be good to mention the FIPS startup penalty and its adverse effect on job throughput in the commit message problem statement, even though the issue reference does provide that background.   Something like:

> Isolate 0MQ functions that depend on libzmq in a non-exported convenience library, to make it possible to minimize the number of components linked with libzmq.   On systems configured for FIPS compliance, the cyrpto libs used by libzmq perform a self-test in the dso init function, which slows down library loading.  When 0MQ functions were part of libflux-core.so, this significantly impaired job throughput.

grondo:
Sorry, I'm not up to date on this PR. However, I was going to ask if it would be useful to add some kind of test that ensures the build (maybe just `flux-shell`) hasn't linked against libcrypto? It would be trivial to add and might catch and accidental regression down the line (though I'm not sure how likely that actually is).

Just thought I throw that idea out there.

src/common/libflux/message.c
chu11: adding `common/libflux/libflux.la` is causing the segfault in the sharness tests again for me.  It commonly comes with `t0009-dmesg.t`.

```
(gdb) bt
#0  0x00002aaaaf010a30 in ?? ()
#1  0x00002aaaaad2938e in fzlistx_delete (self=self@entry=0x2aaab4001240, handle=handle@entry=0x0) at zlistx.c:408
#2  0x00002aaaaad2957a in fzlistx_purge (self=self@entry=0x2aaab4001240) at zlistx.c:465
#3  0x00002aaaaad295ce in fzlistx_destroy (self_p=0x2aaab40096a0) at zlistx.c:131
#4  0x00002aaaaacef75f in flux_msglist_destroy (l=0x2aaab40096a0) at msglist.c:66
#5  0x00002aaaaace92ae in flux_handle_destroy (h=0x2aaab4001170) at handle.c:393
#6  0x000000000040ae72 in module_thread (arg=0x650ef0) at module.c:231
#7  0x00002aaaab16bea5 in start_thread (arg=0x2aaaaefd4700) at pthread_create.c:307
#8  0x00002aaaac2b39fd in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:111
```

I'm pretty perplexed by this, as the last function pointer is off into undefined space.  My assumption was that there was a symbol collision during linking.  Or there is some memory corruption going on otherwise not related to this.  But when I look at the output from `nm`, this kinda stuff bothers me:

```
0000000000426e10 T flux_msg_get_seq
0000000000426f40 T flux_msg_get_status
0000000000427b20 T flux_msg_get_string
0000000000427eb0 T flux_msg_get_topic
00000000004268a0 T flux_msg_get_type
0000000000426aa0 T flux_msg_get_userid
                 U flux_msg_handler_addvec
                 U flux_msg_handler_allow_rolemask
                 U flux_msg_handler_create
                 U flux_msg_handler_delvec
                 U flux_msg_handler_destroy
                 U flux_msg_handler_start
                 U flux_msg_handler_stop
0000000000427880 T flux_msg_has_payload
00000000004262b0 T flux_msg_incref
0000000000426a50 T flux_msg_is_noresponse
0000000000426990 T flux_msg_is_private
00000000004269f0 T flux_msg_is_streaming
0000000000427db0 T flux_msg_last_error
0000000000427a90 T flux_msg_pack
0000000000426fc0 T flux_msg_route_clear
```

Some symbols in the `libflux-core.la` shared object are bring brought in and some aren't.

I think the best course of action might be to make a `src/common/libflux/libmessageprivate.la` that `libzmqutil` can ldadd for itself.

chu11:
> Sorry, I'm not up to date on this PR. However, I was going to ask if it would be useful to add some kind of test that ensures the build (maybe just flux-shell) hasn't linked against libcrypto? It would be trivial to add and might catch and accidental regression down the line (though I'm not sure how likely that actually is).

Good idea.  I just added a simple:

```
if ldd ${FLUX_BUILD_DIR}/src/shell/.libs/lt-flux-shell | grep -q zmq                                                                        
then                                                                                                                                        
    exit 1                                                                                                                                  
fi                                                                                                                                          
exit 0  
```

regression test, until I remembered our conversation above regarding the default of `enable_fast_install`.  I'm not sure the above test is portable.  But presumably checking `ldd src/shell/.libs/flux-shell` isn't portable or safe as well.  hmmm.

src/common/libflux/message.c
garlick: Hmm, apparently `message.o` was statically linked and `msg_handler.o` will be dynamically linked.  Maybe that is because symbols in `message.o` are referenced by the new convenience library, and `msg_handler.o` symbols are being referenced by the link target?

On the segfault, looks like purge is calling delete with a NULL list entry handle.  The final symbol _would_ be a destructor function pointer, but maybe that pointer is pointing somewhere unresolveable because the list is corrupted.  No idea why rearranging dynamic loading would uncover this but in case it's a real bug, it might be good to run that test under valgrind and see what pops up?  (Happy to help if you want to share your development branch)

Interesting that the stack trace is for a broker module but the test uses the `minimal` personality, which I thought didn't load any modules?

garlick:
> presumably checking ldd src/shell/.libs/flux-shell isn't portable or safe

Would `libtool e ldd src/shell/flux-shell` work here?

src/common/libflux/message.c
chu11: So digging into the segfault, I'm quite perplexed.  valgrind didn't catch anything and nothing in gdb looked out of whack.

The fact that gdb says the pointer goes off to nowhere concerned me.  I added a bunch of printfs just to try and see what was going on and eventually came to this, which I've never seen before:

```
flux_msglist_create:66 msglist = 0x2aaab00096c0, msg_destructor = 0x2aaaaf010b10
flux_handle_create:336 handle=0x2aaab0001190 msglist=0x2aaab00096c0 msglist destructor=0x2aaaaf010b10

<snip>

flux_handle_destroy:394 handle=0x2aaab0001190 msglist=0x2aaab00096c0 msglist destructor=0x2aaaaf010b10
flux_msglist_destroy:74 msglist = 0x2aaab00096c0, msg_destructor = 0x2aaaaacef740
fzlistx_purge:465 - zlistx destructor = 0x2aaaaf010b10
```

Basically just printing out some pointer values in some create/destroy paths (and pointers to data structures so I match up creates & destroys).  Note the changing value for `msg_destructor` above.  This is the value of the `msg_destructor()` function in `libflux/msglist.c`.   It's not a stored value in some struct.  I don't think the address location of a function should ever change?  The pointer to that function that was stored probably is illegal, thus the segfault.

Then I added more printf debugging and saw this horror:

```
flux_msglist_pollfd:189 msglist = 0x2aaab00096c0, msg_destructor = 0x2aaaaacef740
flux_msglist_pollevents:198 msglist = 0x2aaab00096c0, msg_destructor = 0x2aaaaf010b10
flux_msglist_pollevents:198 msglist = 0x2aaab00096c0, msg_destructor = 0x2aaaaf010b10
flux_msglist_pollevents:198 msglist = 0x2aaab00096c0, msg_destructor = 0x2aaaaf010b10
flux_msglist_pollevents:198 msglist = 0x2aaab00096c0, msg_destructor = 0x2aaaaf010b10
flux_msglist_count:182 msglist = 0x2aaab00096c0, msg_destructor = 0x2aaaaf010b10
flux_msglist_pollevents:198 msglist = 0x2aaab00096c0, msg_destructor = 0x2aaaaacef740
flux_msglist_pollevents:198 msglist = 0x2aaab00096c0, msg_destructor = 0x2aaaaacef740
```

The address of the `msg_destructor()` function changed over time.

Theory ... could the same symbol have been linked in twice, and there are issues depending on which function is being called?

chu11:
re-pushed, fixing up the commit message per comments above and adding the extra test.  This re-push is mostly to see if the new test works correctly under the docker images.  None of the non-prefixing of the iovec functions is in this re-push.

src/common/libflux/message.c
garlick: That reproduces for me.

It also makes no sense!  Function addresses are supposed to be constant.  Ug.  I'm going to need to sleep on this one.

chu11: I don't know internally how a binary "looks up" a symbol when it calls a function, but I'm wondering if the binary linked symbol and the symbol in the shared object are both looked up at times.  Doing another debugging run, the function address of `flux_msglist_pollevents()` also changes.

garlick: Yeah, no further enlightenment here.:q


Is it possible to inline those iovec functions as a quick solution?

Also, do the PROTO offsets and stuff need to be in that header?  Seems like only PROTO_SIZE needs to be exported?

chu11: Yeah, the proto export is just for the PROTO_SIZE.

chu11: Speaking to some DEG folks, there is some scanning stuff that is done internally to find symbols and that scanning can change over time, such as if libraries are used in a certain order / loaded in certain order.  Setting LDDEBUG=all, I noticed this:

```
    161371:	binding file /g/g0/achu/chaos/git/flux-framework/flux-core/src/common/.libs/libflux-core.so.2 [0] to /g/g0/achu/chaos/git/flux-framework/flux-core/src/common/.libs/libflux-core.so.2 [0]: normal symbol `flux_msglist_create'
    161452:	binding file /g/g0/achu/chaos/git/flux-framework/flux-core/src/common/.libs/libflux-core.so.2 [0] to /g/g0/achu/chaos/git/flux-framework/flux-core/src/common/.libs/libflux-core.so.2 [0]: normal symbol `flux_msglist_create'
    161452:	binding file /g/g0/achu/chaos/git/flux-framework/flux-core/src/broker/.libs/lt-flux-broker [0] to /g/g0/achu/chaos/git/flux-framework/flux-core/src/common/.libs/libflux-core.so.2 [0]: normal symbol `flux_msglist_create'
```

I don't know exactly how to read this output, but it does look like `flux_msglist_create()` was found in the .so file and the `lt-flux-broker`.

chu11: So  I'm trying the `libmessageprivate.la` approach now.  Unfortunately we have to put more stuff into `message_private.h` and `message_private.c` for things to work, but it seems to more correct solution.

chu11:
re-pushed.  To resolve not exporting iovec functions, I created a `libmessageprivate.la" convenience library within `libflux`.  This will export the iovec functions and a collection of helper functions that are needed.  Unfortunately, this approach did increase the size of message private more than before.  `struct flux_msg` and `struct route_id` and `msg_proto_setup()` also had to be added to message private.

note I added these functions to message private.

```
int msg_route_push (flux_msg_t *msg,
                    const char *id,
                    unsigned int id_len);

int msg_route_append (flux_msg_t *msg,
                      const char *id,
                      unsigned int id_len);

void msg_route_clear (flux_msg_t *msg);

int msg_route_delete_last (flux_msg_t *msg);
```

basically I had to make a call about making `route_id_create()` and `route_id_destroy()` callable in `message_private` OR to hide all of their usage behind "route id helper functions".  I did the latter.

garlick:
OK, finally getting close to done I think.  The `message_private.c` seems like maybe it should be split up topically.  Without going too crazy with refactoring, I do have a modest set of changes I was playing with on a copy of your branch.  Let me try to clean that up and push it somewhere that you can have a look at.

garlick:
Well FWIW I moved some things around on my `issue3617_zmqutil_refactor` branch.

The changes are just on top of this branch and do not have proper commit messages since they probably would just be squashed if you decide to include them.  The changes are
* rename `libmessageprivate.la` to `libmessage.la`
* include it in `libflux-internal.la`
* split `libmessage_private.[ch]` into `_iovec`, `_proto`, and `_route` source modules

Let me know what you think @chu11.  I'm open to whatever - we need to get this done and move on.

chu11:
@garlick Ahhh, I didn't think about putting it into libflux-internal.  Makes sense. I think it works.  Let me yoink your commits and squash am where appropriate and make sure all commits rebuild/pass tests accordingly.

chu11:
re-pushed, squashing and adjusting commits in the process, including the commit messages.

chu11:
hmmm, one builder failed with.

```
  FAIL: test_plugin.t 100 - flux_plugin_load worked
  #   Failed test 'flux_plugin_load worked'
  #   at test/plugin.c line 419.
  Bail out!  Failed to load test plugin: dlopen: test/.libs/plugin_foo.so: undefined symbol: flux_kvs_commit
  ERROR: test_plugin.t - Bail out! Failed to load test plugin: dlopen: test/.libs/plugin_foo.so: undefined symbol: flux_kvs_commit
```

i'm perplexed, as nothing in the unit tests or `src/common/libflux` uses the kvs.  And it's only one builder.  Going to restart.

chu11:
re-pushed with a fix for the builder failure described above, just had to add a `libkvs/libkvs.la` dependency addition.

garlick (approved):
I think we're good here.  Thanks for the long slog!  Set MWP when you're ready.

Issue #3517, opened 2021-02-17
idea: instance wide stats collection service


grondo:
Currently we do have some high-level metrics available for a Flux instance, e.g. we can discern some information from the various job eventlogs, maybe `flux dmesg`, and there are per-module built in message counters (though those are overridden if a module provides customs stats as in the `kvs` module and `job-info`)

Trying to diagnose a couple recent performance issues made clear that it would be nice if we had a more formal way to gather metrics in a flux instance. This issue describes one idea for that interface.

The basic idea is to have an instance wide metrics or stats aggregation service, inspired by existing projects like Etsy's [StatsD](https://codeascraft.com/2011/02/15/measure-anything-measure-everything/) or GitHub's [brubeck](https://github.com/github/brubeck) (See also https://github.blog/2015-06-15-brubeck/)

As an experiment we could try just running a brubeck instance alongside Flux and experimentally send metrics from all brokers and modules directly to its UDP server. Supposedly a brubeck instance can handle 4M metrics per second.

However, a more flux-like stats aggregation service might be to include an aggregator with every flux handle. Metrics could be locally aggregated and forwarded "upstream" when appropriate, perhaps on a timer or idle watcher for flux handles being used with a reactor, maybe some other clever way for handles that aren't being used with a reactor (if that is even necessary). If nothing else, even being able dump aggregated statistics from a handle used in a utility may be insightful.

The end result would be that metrics for the entire instance would be aggregated at the root, perhaps in a new module, which can offer services to dump all statistics or be configured to send stats periodically to graphite or other well-developed stats presentation software (e.g. graphite).

A new metrics API would be added to libflux to push metrics into a handle. We could start with supporting the standard statsd [metric types](https://github.com/statsd/statsd/blob/master/docs/metric_types.md)

Probably starting with some simple counters and timers could go a long way to opening up observability into a flux instance.

As an example, take the recent case of a broker that was observed via `top` to be using >100% CPU, though the instance was not making much progress in handling job throughput.

If modules by default appended to a counter on every wakeup, then we might be able to check our stats service and easily see which module is consuming all the CPU time. We could reinstate the builtin module prep/check watchers to observe the amount of time every module spends sleeping in the reactor vs doing work, and perhaps in an instance that is being slow we could pinpoint the module that is taking longer to handle events over time.

[...]

PR #3806, opened 2021-07-29
add stats api for internal metric collection

garrettbslone:
This adds a simple metric collection & sender api to `libflux` as suggested in issue #3517. The public api in `src/common/libflux/stats.h` defines the functions for storing metrics and changing the internal flush interval (which is `1s` by default). There are three types of supported metrics: `counters`, `gauges`, and `timers`. `Counters` send integers values which are aggregated in Brubeck over its flush interval. `Gauges` take an arbitrary integer value and maintain that value until next set, like a snapshot at a certain point. `Timers` take an amount of time in miliseconds. 

Right now there are no CI tests (something I need some suggestions on), but to test locally first set up Brubeck and Graphite which I have some instructions on setting up in my [https://github.com/flux-framework/fripp/blob/main/journals/week1.md](journal). Next, the stats will try getting the address to send to by reading the `FLUX_FRIPP_STATSD={IP}:{PORT}` environment variable. If you use the default config file from brubeck it should be set to `0.0.0.0:8126`. This branch has the stats running in the `content-cache`, but another use case is to use the [https://github.com/flux-framework/fripp/blob/main/code/jobtap-plugins/state-counts.c](state-counts plugin) which can be built like:

```bash
$ gcc -shared -I$HOME/flux-core -I$HOME/flux-core/src/include -fPIC \
    -o state-counts.so ../fripp.h ../fripp.c state-counts.c
```

and assuming that the plugin is in `../fripp/code/jobtap-plugins/`, it can be loaded into a running instance like this:

```bash
$ flux jobtap load $(pwd)/../fripp/code/jobtap-plugins/state-counts.so
```

### DIFF
--- a/src/broker/content-cache.c
+++ b/src/broker/content-cache.c
@@ -781,9 +781,26 @@ static void cache_purge (struct content_cache *cache)
     }
 }
 
+static void update_stats (struct content_cache *cache)
+{
+    flux_stats_gauge_set (cache->h, "content-cache.count",
+        (int) zhashx_size (cache->entries));
+    flux_stats_gauge_set (cache->h, "content-cache.valid",
+        cache->acct_valid);
+    flux_stats_gauge_set (cache->h, "content-cache.dirty",
+        cache->acct_dirty);
+    flux_stats_gauge_set (cache->h, "content-cache.size",
+        cache->acct_size);
+    flux_stats_gauge_set (cache->h, "content-cache.flush-batch-count",
+        cache->flush_batch_count);
+}
+
 static void sync_cb (flux_future_t *f, void *arg)
 {
     struct content_cache *cache = arg;
+
+    if (flux_stats_enabled (cache->h, NULL))
+        update_stats (cache);
 
     cache_purge (cache);
 
@@ -918,25 +935,6 @@ void content_cache_destroy (struct content_cache *cache)
         free (cache);
         errno = saved_errno;
     }
-}
-
-static void timer_cb (flux_reactor_t *r, 
-                      flux_watcher_t *w, 
-                      int revents, 
-                      void *arg)
-{
-    struct content_cache *cache = arg;
-
-    flux_stats_gauge (cache->h, "flux.content-cache.count",
-        (int) zhashx_size (cache->entries), false);
-    flux_stats_gauge (cache->h, "flux.content-cache.valid",
-        cache->acct_valid, false);
-    flux_stats_gauge (cache->h, "flux.content-cache.dirty",
-        cache->acct_dirty, false);
-    flux_stats_gauge (cache->h, "flux.content-cache.size",
-        cache->acct_size, false);
-    flux_stats_gauge (cache->h, "flux.content-cache.flush-batch-count",
-        cache->flush_batch_count, false);
 }
 
 struct content_cache *content_cache_create (flux_t *h, attr_t *attrs)
--- a/src/broker/content-cache.c
+++ b/src/broker/content-cache.c
@@ -920,6 +920,25 @@ void content_cache_destroy (struct content_cache *cache)
     }
 }
 
+static void timer_cb (flux_reactor_t *r, 
+                      flux_watcher_t *w, 
+                      int revents, 
+                      void *arg)
+{
+    struct content_cache *cache = arg;
+
+    flux_stats_gauge (cache->h, "flux.content-cache.count",
+        (int) zhashx_size (cache->entries), false);
+    flux_stats_gauge (cache->h, "flux.content-cache.valid",
+        cache->acct_valid, false);
+    flux_stats_gauge (cache->h, "flux.content-cache.dirty",
+        cache->acct_dirty, false);
+    flux_stats_gauge (cache->h, "flux.content-cache.size",
+        cache->acct_size, false);
+    flux_stats_gauge (cache->h, "flux.content-cache.flush-batch-count",
+        cache->flush_batch_count, false);
+}
+
 struct content_cache *content_cache_create (flux_t *h, attr_t *attrs)
 {
     struct content_cache *cache;
--- a/src/common/libflux/fripp.c
+++ b/src/common/libflux/fripp.c
@@ -1,0 +1,468 @@
+/************************************************************\
+ * Copyright 2021 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include <arpa/inet.h>
+#include <netinet/in.h>
+#include <stdbool.h>
+#include <sys/socket.h>
+#include <sys/stat.h>
+#include <syslog.h>
+#include <time.h>
+#include <unistd.h>
+
+#include "attr.h"
+#include "flog.h"
+#include "fripp.h"
+#include "reactor.h"
+
+#include "src/common/libczmqcontainers/czmq_containers.h"
+#include "src/common/libutil/iterators.h"
+
+#define FRIPP_MAX_PACKET_LEN 1440
+#define INTERNAL_BUFFSIZE 128
+#define DEFAULT_AGG_PERIOD 1.0
+
+typedef enum {
+    BRUBECK_COUNTER,
+    BRUBECK_GAUGE,
+    BRUBECK_TIMER
+} metric_type;
+
+union val {
+    ssize_t l;
+    double d;
+};
+
+struct metric {
+    union val cur;
+    union val prev;
+    bool inc;
+    metric_type type;
+};
+
+struct fripp_ctx {
+    struct sockaddr_in si_server;
+    int sock;
+    int buf_len;
+    int tail;
+    char *buf;
+    char prefix[INTERNAL_BUFFSIZE];
+
+    zhashx_t *metrics;
+    zlist_t *done;
+    flux_watcher_t *w;
+    double period;
+};
+
+bool fripp_enabled (struct fripp_ctx *ctx, const char *metric)
+{
+    if (ctx && metric)
+        return zhashx_lookup (ctx->metrics, metric) != NULL;
+
+    return getenv ("FLUX_FRIPP_STATSD") != NULL;
+}
+
+void fripp_set_prefix (struct fripp_ctx *ctx, const char *prefix)
+{
+    strncpy (ctx->prefix, prefix, INTERNAL_BUFFSIZE - 1);
+}
+
+static int split_packet (char *packet, int len)
+{
+    for (int i = len - 1; i >= 0; --i) {
+        if (packet[i] == '\n' || packet[i] == '\0') {
+            packet[i] = '\0';
+            return i;
+        }
+    }
+
+    return -1;
+}
+
+static int fripp_send_metrics (struct fripp_ctx *ctx)
+{
+    int len, sock_len = sizeof (ctx->si_server);
+    bool split;
+    char *packet = ctx->buf;
+
+    do {
+        split = false;
+        len = strlen (packet);
+
+        if (len > FRIPP_MAX_PACKET_LEN)
+            split = true;
+        if ((len = split_packet (packet, len > FRIPP_MAX_PACKET_LEN ?
+                FRIPP_MAX_PACKET_LEN : len)) == -1)
+            return -1;
+        (void) sendto (ctx->sock, packet, len, 0,
+                    (void *) &ctx->si_server, sock_len);
+    } while (split && (packet = &packet[len + 1]));
+
+    return 0;
+}
+
+int fripp_packet_appendf (struct fripp_ctx *ctx, const char *fmt, ...)
+{
+    va_list ap, cpy;
+    va_start (ap, fmt);
+    va_copy (cpy, ap);
+
+    char buf[INTERNAL_BUFFSIZE];
+    int len, rc = 0;
+
+    if ((len = vsnprintf (buf, INTERNAL_BUFFSIZE, fmt, ap))
+            >= INTERNAL_BUFFSIZE) {
+        rc = -1;
+        goto done;
+    }
+
+    ctx->tail += len;
+
+    if (ctx->tail >= ctx->buf_len) {
+        char *tmp;
+        if (!(tmp = realloc(ctx->buf, (ctx->tail + 1) * sizeof (char)))) {
+            rc = -1;
+            goto done;
+        }
+
+        ctx->buf_len = ctx->tail + 1;
+        ctx->buf = tmp;
+    }
+
+    strcat (ctx->buf, buf);
+
+done:
+    va_end (ap);
+    va_end (cpy);
+
+    return rc;
+}
+
+int fripp_sendf (struct fripp_ctx *ctx, const char *fmt, ...)
+{
+    va_list ap, cpy;
+    va_start (ap, fmt);
+    va_copy (cpy, ap);
+
+    int len, rc = 0;
+
+    if ((len = vsnprintf (ctx->buf, ctx->buf_len, fmt, ap))
+         >= ctx->buf_len) {
+
+        char *tmp;
+        if (!(tmp = realloc (ctx->buf, (len + 1) * sizeof (char)))) {
+            rc = -1;
+            goto done;
+        }
+
+        ctx->buf = tmp;
+        ctx->buf_len = len + 1;
+        (void) vsnprintf (ctx->buf, ctx->buf_len, fmt, cpy);
+    }
+
+    rc = fripp_send_metrics (ctx);
+
+done:
+    va_end (ap);
+    va_end (cpy);
+
+    return rc;
+}
+
+int fripp_count (struct fripp_ctx *ctx,
+                 const char *name,
+                 ssize_t count)
+{
+    if (ctx->period == 0)
+        return fripp_sendf (ctx, "%s.%s:%zd|C\n", ctx->prefix, name, count);
+
+    struct metric *m;
+
+    if (!(m = zhashx_lookup (ctx->metrics, name))) {
+        if (!(m = malloc (sizeof (*m))))
+            return -1;
+
+        zhashx_insert (ctx->metrics, name, (void *) m);
+        m->prev.l = -1;
+    }
+
+    m->type = BRUBECK_COUNTER;
+    m->inc = false;
+    m->cur.l = count;
+
+    flux_watcher_start (ctx->w);
+
+    return 0;
+}
+
+int fripp_gauge (struct fripp_ctx *ctx,
+                 const char *name,
+                 ssize_t value,
+                 bool inc)
+{
+    if (ctx->period == 0)
+        return fripp_sendf (ctx, inc && value > 0 ?
+                            "%s.%s:+%zd|g\n" : "%s.%s:%zd|g\n",
+                            ctx->prefix, name, value);
+
+    struct metric *m;
+
+    if (!(m = zhashx_lookup (ctx->metrics, name))) {
+        if (!(m = calloc (1, sizeof (*m))))
+            return -1;
+
+        zhashx_insert (ctx->metrics, name, (void *) m);
+        m->prev.l = -1;
+    }
+
+    m->type = BRUBECK_GAUGE;
+    m->inc = inc;
+    m->cur.l = inc ? m->cur.l + value : value;
+
+    flux_watcher_start (ctx->w);
+
+    return 0;
+}
+
+int fripp_timing (struct fripp_ctx *ctx,
+                  const char *name,
+                  double ms)
+{
+    if (ctx->period == 0)
+        return fripp_sendf (ctx, "%s.%s:%lf|ms\n", ctx->prefix, name, ms);
+
+    struct metric *m;
+
+    if (!(m = zhashx_lookup (ctx->metrics, name))) {
+        if (!(m = malloc (sizeof (*m))))
+            return -1;
+
+        zhashx_insert (ctx->metrics, name, (void *) m);
+        m->prev.d = -1.0;
+    }
+
+    m->type = BRUBECK_TIMER;
+    m->inc = false;
+    m->cur.d = ms;
+
+    flux_watcher_start (ctx->w);
+
+    return 0;
+}
+
+static void metric_destroy (void **item)
+{
+    if (item) {
+        free (*item);
+        *item = NULL;
+    }
+}
+
+static void timer_cb (flux_reactor_t *r,
+                      flux_watcher_t *w,
+                      int revents,
+                      void *arg)
+{
+    struct fripp_ctx *ctx = arg;
+    const char *name;
+    struct metric *m;
+
+    int rc = 0;
+
+    if (!ctx->buf[0] && zhashx_size (ctx->metrics) == 0) {
+        flux_watcher_stop (ctx->w);
+        return;
+    }
+
+    FOREACH_ZHASHX (ctx->metrics, name, m) {
+        if ((m->type == BRUBECK_COUNTER || m->type == BRUBECK_GAUGE)
+                && m->cur.l == m->prev.l) {
+            zlist_append (ctx->done, (void *) name);
+            continue;
+        }
+        if (m->type == BRUBECK_TIMER && m->cur.d == m->prev.d) {
+            zlist_append (ctx->done, (void *) name);
+            continue;
+        }
+
+        m->prev.l = m->cur.l;
+
+        switch (m->type) {
+            case BRUBECK_COUNTER:
+                rc = fripp_packet_appendf (ctx,
+                                           "%s.%s:%zd|C\n",
+                                           ctx->prefix,
+                                           name,
+                                           m->cur.l);
+                break;
+            case BRUBECK_GAUGE:
+                rc = fripp_packet_appendf (ctx,
+                                           "%s.%s:%zd|g\n",
+                                           ctx->prefix,
+                                           name,
+                                           m->cur.l);
+                break;
+            case BRUBECK_TIMER:
+                rc = fripp_packet_appendf (ctx,
+                                           "%s.%s:%lf|ms\n",
+                                           ctx->prefix,
+                                           name,
+                                           m->cur.d);
+                break;
+            default:
+                break;
+        }
+
+        if (rc)
+            continue;
+    }
+
+    fripp_send_metrics (ctx);
+
+    char *s;
+    while ((s = zlist_pop (ctx->done)))
+        zhashx_delete (ctx->metrics, s);
+
+    memset (ctx->buf, 0, ctx->buf_len);
+
+    ctx->tail = 0;
+}
+
+void fripp_set_agg_period (struct fripp_ctx *ctx, double period)
+{
+    if (period <= 0) {
+        flux_watcher_stop (ctx->w);
+        ctx->period = 0;
+        return;
+    }
+
+    ctx->period = period;
+    double after = flux_watcher_next_wakeup (ctx->w) - (double) time (NULL);
+
+    flux_timer_watcher_reset (ctx->w, after, ctx->period);
+}
+
+void fripp_ctx_destroy (void *arg)
+{
+    struct fripp_ctx *ctx = arg;
+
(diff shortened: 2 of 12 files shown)


garlick:
Maybe some variant of this would work as the basis for testing in sharness?
```
$  FLUX_FRIPP_STATSD=0.0.0.0:9999 src/cmd/flux start socat UDP-RECVFROM:9999 -
flux.content-cache.dirty:0|g
flux.content-cache.valid:10|g
flux.content-cache.size:1380|g
flux.content-cache.count:10|g
flux.content-cache.flush-batch-count:0|g
```

garrettbslone:
Force pushed to fix spelling of gauge in a few places..

src/common/libflux/stats.h
```diff
@@ -0,0 +1,67 @@
...
+ * next flush.
+ */
+int flux_stats_count (flux_t *h, const char *name, ssize_t count);
+
+/* Update (or create) and store 'value' for 'name' to be sent on the
+ * next flush. The'inc' indicates wether or not 'value' is some delta on 
+ * the previous value. If 'inc' is set and 'name' was not previously stored, 
+ * then the value is stored directly.
+ */
+int flux_stats_gauge (flux_t *h, const char *name, ssize_t value, bool inc);
```
grondo, line 50: I think I argued for keeping the `inc` parameter here on the :coffee: call, but I was confused.

I do think there should be a way to increment, decrement, or set a gauge, but I would suggest/prefer 2 separate functions for clarity and ease of use in the code, e.g as a first cut, something like:

```c
int flux_stats_gauge_set (flux_t *h, const char *name, size_t val);
int flux_stats_gauge_inc (flux_t *h, const char *name, ssize_t val);
```

I guess `flux_stats_gauge_inc (h, "mystat", -1)` would then work to decrement a gauge by 1. Optionally, if preferred you could have a separate `flux_status_gauge_dec()`.

As we also talked about it might be nice to provide a "getter", but probably not necessary at this time.

(This means we could probably drop the `flux_stats_inc()` and `flux_stats_dec()` above?)

src/common/libflux/fripp.c
```diff
@@ -0,0 +1,438 @@
+/************************************************************\
+ * Copyright 2014 Lawrence Livermore National Security, LLC
```
chu11, line 2: just a reminder to update all copyright years to 2021 :-)

src/broker/content-cache.c
```diff
@@ -953,6 +976,15 @@ struct content_cache *content_cache_create (flux_t *h, attr_t *attrs)
...
         goto error;
+    if (!(cache->w = flux_timer_watcher_create (
+              flux_get_reactor(h),
+              1,
+              1,
+              timer_cb,
+              cache)))
+        goto error;
+    flux_watcher_start (cache->w);
+
```
garlick, line 987: Currently this timer fires unconditionally (whether or not we are collecting stats).  On a compute node, this might be stealing cycles from a job, or evicting cache lines, etc.. Is there a way we can disable the timer if we are not collecting stats?

I think it would be OK to piggy back on the `sync_cb()` which gets called every heartbeat (default T=2s).  See also: [flux_sync_create(3)](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man3/flux_sync_create.html).  Maybe just change the `timer_cb()` function into one that gets called from `sync_cb()`.

src/broker/content-cache.c
```diff
@@ -963,4 +995,4 @@ struct content_cache *content_cache_create (flux_t *h, attr_t *attrs)
 
 /*
  * vi:tabstop=4 shiftwidth=4 expandtab
- */
+ */
```
garlick: Please fix trailing white space (I'll just call out this one - there are some others).

src/common/libflux/fripp.c
```diff
@@ -0,0 +1,438 @@
...
+    char *uri, *port_s = NULL, *host = NULL;
+
+    if (!(ctx = calloc (1, sizeof (*ctx)))) {
+        flux_log_error (h, "fripp_ctx_create");
+        goto error;
+    }
+    if (!(uri = getenv ("FLUX_FRIPP_STATSD"))) {
+        flux_log_error (h, "FLUX_FRIPP_STATSD env var not set");
+        goto error;
+    }
```
garlick, line 364: Currently an error is logged if FLUX_FRIPP_STATSD is not set.  Moreover, it's logged for every metric, every time a stats function is called.  FRIPP should be silent if stats collection is not enabled.

Suggestion:  add a function for testing whether stats are enabled:
```c
bool fripp_enabled (flux_t *h, const char *metric);
```
This function can immediately return false if the environment variable is not set, else true.  Later we can use `metric` to test whether a particular metric prefix is enabled, but for now it could just be ignored and set to NULL by callers.

Then call it early to reduce overhead when stats collection is disabled.  For example, in the `flux_stats_*()` functions, return immediately, before trying to access/create context if `fripp_enabled(name)` return false.

src/common/libflux/stats.h
```diff
@@ -0,0 +1,55 @@
...
+#ifndef _FLUX_CORE_STATS_H
+#define _FLUX_CORE_STATS_H
+
+#include <stdarg.h>
+
+#include "handle.h"
+
+/* Format a metric name and increment its value by '1'.
+ */
+int flux_stats_inc (flux_t *h, const char *fmt, ...)
```
garlick, line 32: Just wondering, should all these functions return `void`?  It seems unlikely that callers will want to be burdened with handling any potential error.

garrettbslone:
Force pushing to address some of the suggestions above. 

Starting with the `content-cache` what has changed is the `timer watcher` where the stats were being collected was removed, and they are now being collected inside of the `sync_cb`

In `stats.c/h`:
- removed the `flux_stats_inc/dec` functions
- removed the `inc` option from `flux_stats_gauge` and replaced it with two functions, one to set a gauge and one to increment one
- added a function to set an optional prefix for all metrics sent from the handle
- added a function to check if stats collection is enabled
- removed return value from stats functions
- pulled out some common code into a function
- if stats collection is not enabled return right away instead of trying to get `fripp_ctx`

In `fripp.c/h`:
- added a prefix to be prepended to all metrics in the `fripp_ctx` with a default of `flux.{rank}`
- added a function to check if stats collection is enabled

src/broker/content-cache.c
```diff
@@ -920,6 +920,25 @@ void content_cache_destroy (struct content_cache *cache)
     }
 }
 
+static void timer_cb (flux_reactor_t *r, 
+                      flux_watcher_t *w, 
+                      int revents, 
+                      void *arg)
+{
```
garlick: Oops, should this be in this commit?

src/common/libflux/stats.c
```diff
@@ -0,0 +1,109 @@
...
+{
+    if (!fripp_enabled (h, NULL))
+        return NULL;
+
+    struct fripp_ctx *ctx;
+    if (!(ctx = flux_aux_get (h, FRIPP_AUX_TAG))) {
+        if (!(ctx = fripp_ctx_create (h)))
+            return NULL;
+        if (flux_aux_set (h, FRIPP_AUX_TAG, ctx, fripp_ctx_destroy) == -1)
+            return NULL;
```
garlick, line 33: leaks ctx on aux_set failure

src/common/libflux/stats.c
```diff
@@ -0,0 +1,109 @@
...
+    struct fripp_ctx *ctx;
+    if (!(ctx = get_fripp_ctx (h)))
+        return;
+
+    va_list ap;
+    va_start (ap, fmt);
+
+    char prefix[128];
+    if (vsnprintf (prefix, sizeof (prefix), fmt, ap) >= 128)
+        return;
```
garlick, line 96: be sure va_end() is called on all exit paths after va_start() is called.

src/common/libflux/stats.h
```diff
@@ -0,0 +1,79 @@
...
+/* Metric types:
+ * Counter - An integer value that will, on the backend (brubeck) send the
+ *           count and reset to 0 at each flush. It calculates the change
+ *           from the value sent at the previous flush.
+ *           An example of where to use a counter is the builtin msgcounters
+ *           that are part of each flux_t handle. The counts will continually
+ *           increase and brubeck will handle calculating the count sent in
+ *           the interval.
+ * Gauge   - An integer that takes an arbitrary value and maintains its
+ *           it is set again. Gauges can also take incremental values
```
garlick, line 25: did you mean "maintains its value"?

Really appreciate the inline docs here!

src/common/libflux/stats.h
```diff
@@ -0,0 +1,79 @@
...
+ * next flush. If'name' was not previously stored, then the value is stored
+ * directly (i.e. assumed 0 previous value).
+ */
+void flux_stats_gauge_inc (flux_t *h, const char *name, ssize_t inc);
+
+
+/* Update (or create) and store 'ms' for 'name' to be sent on the
+ * next flush.
+ */
+void flux_stats_timing (flux_t *h, const char *name, double ts);
```
garlick, line 59: Comment references `ms` but argument is `ts`.  (Its not a timestamp, so ts might be misleading)

Name of function is actually `flux_stats_timer()` right?

src/common/libflux/stats.h
```diff
@@ -0,0 +1,79 @@
...
+
+/* Update the internal aggregation period over which metrics accumulate
+ * before being set. A 'period' of '0' indicates the metrics should be
+ * sent immediately.
+ */
```
garlick, line 64: might be useful to state the default aggregation period here.

src/common/libflux/stats.h
```diff
@@ -0,0 +1,79 @@
...
+void flux_stats_set_period (flux_t *h, double period);
+
+/* Set the prefix {prefix}.{metric} to be preprended to all metrics
+ * sent from the handle. The prefix has a max limit of 127 characters.
+ * The default prefix is flux.{rank}.
+ */
+void flux_stats_set_prefix (flux_t *h, const char *fmt, ...);
```
garlick, line 71: We use mustache templates elsewhere in Flux so the curly braces might be a bit confusing.  For a second there I thought maybe `{rank}` would be expanded if I included it in the prefix.  Hmm, should it?

Also "set the {prefix}.{metric}" is a bit confusing.  Aren't we just setting the prefix?

src/common/libflux/stats.h
```diff
@@ -0,0 +1,79 @@
...
+
+bool flux_stats_enabled (flux_t *h, const char *metric);
```
garlick, line 73: Missing header doc?

src/common/libflux/fripp.c
```diff
@@ -0,0 +1,456 @@
...
+    char *buf;
+    char prefix[INTERNAL_BUFFSIZE];
+
+    zhashx_t *metrics;
+    zlist_t *done;
+    flux_watcher_t *w;
+    double period;
+};
+
+bool fripp_enabled (flux_t *h, const char *metric)
```
garlick, line 68: Why does this function take a `flux_t` handle?

src/common/libflux/fripp.c
```diff
@@ -0,0 +1,456 @@
...
+#include "attr.h"
+#include "flog.h"
+#include "fripp.h"
+#include "reactor.h"
+
+#include "src/common/libczmqcontainers/czmq_containers.h"
+#include "src/common/libutil/iterators.h"
+
+#define FRIPP_MAX_PACKET_LEN 1440
+#define INTERNAL_BUFFSIZE 128
```
garlick, line 33: Shouldn't the buffer be larger than the max packet size?

src/common/libflux/fripp.c
```diff
@@ -0,0 +1,456 @@
...
+
+        if (len > FRIPP_MAX_PACKET_LEN)
+            split = true;
+        if ((len = split_packet (packet, len > FRIPP_MAX_PACKET_LEN ?
+                FRIPP_MAX_PACKET_LEN : len)) == -1)
+            return -1;
+        if (sendto (ctx->sock, packet, len, 0,
+                    (void *) &ctx->si_server, sock_len) < 0)
+            fprintf (stderr, "packet %s dropped: %s\n", packet,
+                     strerror (errno));
```
garlick, line 108: the fprintf to stderr should be dropped (leftover debugging?)

It's not great practice (though sometimes the only way) to print stuff on stdio from a library.  Here the entire stats collection mechanism ought to just be "best effort" IMHO.

garlick (commented):
A few comments based on a quick skim.  I'm headed out on a walk and will be back later, but wanted to get these out there.

src/common/libflux/stats.h
grondo: >  I thought maybe {rank} would be expanded if I included it in the prefix. Hmm, should it?

Mustache templates have double curly braces, e.g. `{{rank}}`, but I agree that curly braces could be confusing (esp. when you've recently worked in Python :wink:).

I like the idea of allowing expansion in metric names to ease the burden of adding metrics to any site. W
We do have a bare bones mustache renderer currently in the flux-shell source [here](https://github.com/flux-framework/flux-core/blob/master/src/shell/mustache.h). We could eventually move that into libutil for use in other Flux code.

However, it seems fine to leave template expansion out of this particular PR IMO.

garlick: Oh yeah :-)  I was thinking of `flux jobs --format`.  Agreed, expansion could be done in another PR, if we want to go that way.

src/common/libflux/fripp.c
garrettbslone: It takes a `flux_t` handle instead of a `fripp_ctx` to avoid trying to get/create a `fripp_ctx` each time it checks if collection is on.

src/common/libflux/stats.h
garrettbslone: Function is named `flux_stats_timing` (and `fripp_timing`) I believe to make it clear that fripp/stats don't keep track of the timer internally.

src/common/libflux/fripp.c
garrettbslone: Maybe instead it should take a`fripp_ctx` and then if the `fripp_ctx` is `NULL` it can check if it's enabled on the instance.

src/common/libflux/stats.h
garlick: My bad, I thought I read "timer" in the C code and "timing" in the header.  You're right.

src/common/libflux/fripp.c
garlick: That function isn't using the `flux_t` though:
```c
bool fripp_enabled (flux_t *h, const char *metric)
{
    return getenv ("FLUX_FRIPP_STATSD") != NULL;
}
```

src/common/libflux/fripp.c
garrettbslone: Should the `fprintf` on line 314 (in the `timer_cb`) also be dropped?

garlick: Missed that one.  Yep, IMHO.

src/common/libflux/fripp.c
garrettbslone: The buffer is dynamically sized. It is initialized to the `INTERNAL_BUFFSIZE`, but that size is also used for the formatting of metrics in `fripp_packet_appendf` and for the `prefix` buffer.

garlick: Ah OK, sorry I was confused there.

garrettbslone: Maybe it could benefit from a better name or a comment?

t/t1200-stats-basic.t
```diff
@@ -0,0 +1,14 @@
...
+#
+
+test_description='Test stats collection and sending'
+
+. `dirname $0`/sharness.sh
+
+test_under_flux 2
+
+test_expect_success 'content-cache packets recieved' '
+	FLUX_FRIPP_STATSD=0.0.0.0:9999 flux start socat UDP-RECVFROM:9999 -
```
grondo, line 11: Oops, `socat` is not installed in the ci images (nor on my system), so this test is failing.

Also, how will the `socat` know to exit so the test stops?

Here's one idea. You can create a UDP listener in Python for use with the tests. Here's a basic demo that waits for just one packet and exits. (You'd want to add a timeout to this test to detect no packets and fail.

This version doesn't check the format of the packet, but this could easily be added. The script or different scripts could be enhanced to check for properly formatted or expected stats strings, so it might be a good base to start from:

```diff
diff --git a/t/t1200-stats-basic.t b/t/t1200-stats-basic.t
index 1e9f2aa3f..2bfef452b 100755
--- a/t/t1200-stats-basic.t
+++ b/t/t1200-stats-basic.t
@@ -7,8 +7,18 @@ test_description='Test stats collection and sending'
 
 test_under_flux 2
 
+test_expect_success 'create udp client program' '
+       cat <<-EOF >listen.py &&
+       import socket
+       s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+       s.bind(("0.0.0.0", 9999))
+       print(s.recvfrom(1024))
+       EOF
+       chmod +x listen.py
+'
+
 test_expect_success 'content-cache packets recieved' '
-       FLUX_FRIPP_STATSD=0.0.0.0:9999 flux start socat UDP-RECVFROM:9999 -
+       FLUX_FRIPP_STATSD=0.0.0.0:9999 flux start flux python listen.py
 '
 
 test_done

```

grondo:
```
 grondo@asp:~/git/f.git/t$ ./t1200-stats-basic.t  -d -v
sharness: loading extensions from /home/grondo/git/f.git/t/sharness.d/01-setup.sh
sharness: loading extensions from /home/grondo/git/f.git/t/sharness.d/flux-sharness.sh
sharness: loading extensions from /home/grondo/git/f.git/t/sharness.d/01-setup.sh
sharness: loading extensions from /home/grondo/git/f.git/t/sharness.d/flux-sharness.sh
expecting success: 
	cat <<-EOF >listen.py &&
	import socket
	s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
	s.bind(("0.0.0.0", 9999))
	print(s.recvfrom(1024))
	EOF
	chmod +x listen.py

ok 1 - create udp client program

expecting success: 
	FLUX_FRIPP_STATSD=0.0.0.0:9999 flux start flux python listen.py

(b'flux.0.content-cache.size:1360|g\nflux.0.content-cache.dirty:0|g\nflux.0.content-cache.count:10|g\nflux.0.content-cache.flush-batch-count:0|g\nflux.0.content-cache.valid:10|g', ('127.0.0.1', 46371))
ok 2 - content-cache packets recieved

# passed all 2 test(s)
1..2

```

garrettbslone: `socat` was exiting after receiving the first packet when I was testing. Not sure if it's safe to assume that'll always happen.

grondo: Ah, apologies I didn't know that (and didn't have socat to test with :wink:)

If we can get the python script to parse and validate the packets, then that will really solidify the testing.
You could add options to wait for N packets, or wait for a value to appear (e.g. if you wanted to ensure 1 running job was counted, etc.)

src/common/libflux/stats.h
```diff
@@ -0,0 +1,82 @@
...
+ *           increase and brubeck will handle calculating the count sent in
+ *           the interval.
+ * Gauge   - An integer that takes an arbitrary value and maintains its
+ *           value until it is set again. Gauges can also take incremental
+ *           values in the form of a + or - in front of the value which
+ *           increments the previously stored value.
+ *           An example of where to use a gauge is to track the current
+ *           size of the broker's content-cache. At each point, the cache's
+ *           sizes are independent of each other.
+ * Timer   - A double value which represents the time taken for a given
```
grondo, line 31: I think @garlick had already commented on this, but I can't find it.

The function to set a "timer" metric is called `flux_stats_timing()`, but the docs here reference "Timer". Should one of the two be updated? Maybe just document this stat as a "Timing" metric?

t/t1200-stats-basic.t
```diff
@@ -0,0 +1,92 @@
...
+test_description='Test stats collection and sending'
+
+. `dirname $0`/sharness.sh
+
+test_under_flux 2
+
+export FLUX_FRIPP_STATSD=0.0.0.0:9999
+
+test_expect_success 'create udp client program' '
+	cat <<-EOF >listen.py &&
```
grondo, line 13: Since this script is larger than a few lines now, feel free to include it directly in `t/scripts/stats-listen.py` or something to ease future development. (Sorry, this is my fault, I should have suggested that initially)

t/t1200-stats-basic.t
```diff
@@ -0,0 +1,92 @@
...
+	parser.add_argument("-t", "--timeout", help="timeout in s", type=float, default=10.0)
+	parser.add_argument("-V", "--validate", help="validate packet form", action="store_true")
+	parser.add_argument("-w", "--wait-for", metavar="N",
+						help="wait for N packets to be recieved. Should be used along with the timeout arg",
+						type=int, default=1)
+	args = parser.parse_args()
+
+	s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+	s.bind(("0.0.0.0", 9999))
+	s.settimeout(args.timeout)
```
grondo, line 31: Instead of trying to internally track the timeout within `listen.py` it may be easier to run the program under our builtin `run_timeout` function. Then you don't have to worry about  tracking an overall timeout with blocking network code, which is difficult.

You could still keep the `-w, --wait-for=N` and `-s, --search-for=METRIC` options, just exit with zero status after N packets are received or you find the matching metric. On timeout, `run_timeout` will kill the script and trigger a failure.

grondo:
I just threw in a couple on-off comments on the `listen.py` test program, which may be helpful to get it working in CI.

Also, I noticed a typo in the commit message for b6991a00fe5918128ca8f7e02001584ab73c0813 ("validtion")

grondo:
It case it isn't clear why the `python format` check is failing for you -- we use the `black` formatter to consistently format all Python, even test scripts in `t/scripts`. Install the required version of `black` with `python3 -m pip install 'black==20.08.b1' --force-reinstall`, and then either run `black` on your script or run `./scripts/format` to run the formatter on all python scripts.

garrettbslone:
squashed down the commit history and force pushed

t/stats/stats-immediate.c
```diff
@@ -0,0 +1,79 @@
...
+    struct cb_data *data;
+    flux_t *h;
+    if (!(h = flux_jobtap_get_flux (p)))
+        return -1;
+    if (!(data = calloc (1, sizeof (*data))))
+        return -1;
+
+    flux_stats_set_prefix (h, "flux.job.state.immediate");
+    flux_stats_set_period (h, 0.0);
+
```
grondo, line 73: Whitespace error here and at end of file.

t/stats/stats-immediate.c
```diff
@@ -0,0 +1,79 @@
...
+        return -1;
+    if (!(data = calloc (1, sizeof (*data))))
+        return -1;
+
+    flux_stats_set_prefix (h, "flux.job.state.immediate");
+    flux_stats_set_period (h, 0.0);
+    
+    flux_stats_set_prefix (h, "aQmi173rvgumStDdMZdwtJtpLLVJOUXol2aDndev/XsH/gM\
+wlPz/vMZhajJWGctwJZa1uFoAoINjwITPvezGoQDb/9DD3vkPcknN+f/u3vSc0tg/+3aFTONhUIomK\
+B4qiSKSotbtZl3Ewe2Oh+wI+nuG3/ebqIXSoEXjIFOB7vvGA==");
```
grondo, line 76: What is the purpose of this interesting prefix? If nothing else, I'd suggest at least a comment to help future code spelunkers.

grondo: Also I notice prefix is already set above in L71, maybe this was debugging code?

t/stats/stats-immediate.c
```diff
@@ -0,0 +1,79 @@
...
+    return 0;
+}
+
+int flux_plugin_init (flux_plugin_t *p)
+{
+    struct cb_data *data;
+    flux_t *h;
+    if (!(h = flux_jobtap_get_flux (p)))
+        return -1;
+    if (!(data = calloc (1, sizeof (*data))))
```
grondo, line 68: `data` is leaked. You can tie its destruction to the plugin with `flux_plugin_aux_set(3)`

src/common/libflux/fripp.c
```diff
@@ -0,0 +1,468 @@
...
+#include "src/common/libutil/iterators.h"
+
+#define FRIPP_MAX_PACKET_LEN 1440
+#define INTERNAL_BUFFSIZE 128
+#define DEFAULT_AGG_PERIOD 1.0
+
+typedef enum {
+    BRUBECK_COUNTER,
+    BRUBECK_GAUGE,
+    BRUBECK_TIMER
```
grondo, line 39: Question: would these be better named STATSD, or are these metric types specific to brubeck?

t/stats/stats-basic.c
```diff
@@ -0,0 +1,73 @@
...
+    return 0;
+}
+
+int flux_plugin_init (flux_plugin_t *p)
+{
+    struct cb_data *data;
+    flux_t *h;
+    if (!(h = flux_jobtap_get_flux (p)))
+        return -1;
+    if (!(data = calloc (1, sizeof (*data))))
```
grondo, line 66: `data` is leaked (tie to plugin with `flux_plugin_aux_set()`)

grondo (commented):
Ok, just a few comments this round.

Also noticed a typo in the PR title: "interal" missing `n`

t/stats/stats-immediate.c
garrettbslone: This was to make sure that a prefix longer than the limit (127 characters) wouldn't be set and leave the prefix unchanged.

src/common/libflux/fripp.c
garrettbslone: Well, they are all Statsd types and the gauge and timer use the same syntax for brubeck and statsd, but the counter uses a slightly different syntax for brubeck (`C`) than statsd (`c`).

garrettbslone: maybe better for consistency for now?

garrettbslone:
Force pushing to address comments above, removed wip tag from title, and fixed spelling in title.

t/stats/stats-immediate.c
grondo: Ah! Thanks :-)

t/t1200-stats-basic.t
```diff
@@ -0,0 +1,54 @@
...
+udp=$SHARNESS_TEST_SRCDIR/scripts/stats-listen.py
+timeout=$SHARNESS_TEST_SRCDIR/scripts/run_timeout.py
+plugin_i=${FLUX_BUILD_DIR}/t/stats/.libs/stats-immediate.so
+plugin_b=${FLUX_BUILD_DIR}/t/stats/.libs/stats-basic.so
+
+test_expect_success 'load jobtap plugin' '
+	flux jobtap load $plugin_i && flux jobtap load $plugin_b &&
+	flux jobtap list > plugins &&
+	grep stats-basic plugins && grep stats-immediate plugins
+'
```
grondo, line 19: One more comment. I don't think we need this test to run under `test_under_flux`, nor have the jobtap plugins loaded here. I may have missed something but the rest of the functionality tests all use `flux start`, so the "outer" instance created by `test_under_flux` actually isn't used, nor are the loaded plugins.

If that is true then you coudl just remove `test_under_flux 2` and the `load jobtap plugin` test.

src/common/libflux/fripp.c
grondo: This is all internal to `fripp.c` so I guess I don't have a strong feeling at this point.

grondo:
Might as well also rebase against current master. :grin:

grondo:
BTW, you could probably quiet that LGTM error by binding to `127.0.0.1` instead of `0.0.0.0`.

grondo (approved):
LGTM! Nice work @garrettbslone!

garlick:
Yes thanks for all your hard work this summer.  The finished produce is great!

Ready to merge?  If so I'll set MWP and babysit that tempormental coverage test.

garrettbslone:
I think it's ready to merge!

garlick:
Oops I had just set MWP before seeing your comments @grondo.  OK, @garrett, want to take care of that change first?

garlick:
Setting it again - sorry, I was looking at an older comment.

garlick:
@mergifyio rebase

grondo:
:tada: nice work @garrettbslone!

Issue #1804, opened 2018-11-05
audit use of zhash_t for upgrade to zhashx_t


garlick:
It was noted in #1797 that zhash_t has a largish memory footprint for use with a small number of items, and that it would appear zhashx_t might be a little smarter when presented with this sort of usage.

zhashx_t is also more configurable - I used it in the job-manager module because it allowed the hash function to be overridden so that integer jobid's could be used as keys, unlike zhash_t, which would have required string keys to be constructed from the integers.

We should look at how zhash_t is used in flux-core and see if there are any obvious places where a zhashx_t would be a better choice.

garlick:
It seems like the "fastpath" in `libflux/msg_handler.c` is a prime candidate for a zhashx_t.  Before this code was added, all message handlers were appended to a zlist that is searched linearly for every received message.  This is because the `struct match` doesn't lend itself to other lookup methods, and was fine for a small number of message handlers, but became a performance problem as the number of concurrent RPCs went up.  Each RPC adds an ephemeral message handler that is destroyed when the RPC completes. The fastpath code was added to address the performance problem, and is based on a lookup table, only for response messages, that maps matchtags to message handlers in  O(1).

The fastpath code relies on the fact that matchtags are immediately recycled, and so generally their value is numerically low.  It then maintains a dynamically sized array that fits the highest numbered matchtag in use.  I guess I didn't use a zhash_t before because it seemed like a large memory footprint for the common case of low concurrency, and matchtags had to be "stringified" in order to be used as keys.  I question that decision now.  However, zhashx_t solves both of those problems, and would reduce code complexity over the dynamically resized lookup table.

For fun I quickly prototyped this last night and it was a simple change with a [sloccount](https://dwheeler.com/sloccount/) LOC delta of -66.  If it wouldn't be objectionable given our other priorities, I can turn this into a PR that also includes some cleanup and tests.  The prototype code is on my `dispatch_fastpath` branch, specifically 5d6536fa6b4a5e9ce4cbdbe786b8eb7128202a85 in case @grondo wants to retry his test from #1797 and see what the effect is on that test.  (It was his mention of it that got me looking at the code).

grondo:
Nice, your fix saves almost all the 5.4MiB attributed to `fastpath_get` in my contrived reproducer vs current master!

Current master 1956120553b44ce7205e8d0bb9b21c30f9ea3c67:
[...]

PR #3807, opened 2021-07-31
libkvs: compact only if ops len > 1

chu11:
I noticed this will doing some looking at issue #1804 awhile back.  Forgot that I had this sitting on a branch.  It's just a simple / dumb fix.

```
Problem: The txn_compact() function would attempt to compact
transactions if there are > 0 operations.  But it is impossible to
compact just 1 operation, you need atleast 2 operations to conceivably
compact.

Solution: Check for > 1 instead of > 0 operations.
```

### DIFF
--- a/src/common/libkvs/kvs_txn_compact.c
+++ b/src/common/libkvs/kvs_txn_compact.c
@@ -210,7 +210,7 @@ int txn_compact (flux_kvs_txn_t *txn)
     }
 
     len = json_array_size (txn->ops);
-    if (!len)
+    if (len < 2)
         return 0;
 
     if (!(ops_new = json_array ())) {


src/common/libkvs/kvs_txn_compact.c
```diff
@@ -210,7 +210,7 @@ int txn_compact (flux_kvs_txn_t *txn)
     }
 
     len = json_array_size (txn->ops);
-    if (!len)
+    if (!(len > 1))
```
garlick, line 213: Good catch!

Slightly odd comparison there - what about `len < 2`?

garlick (approved):

src/common/libkvs/kvs_txn_compact.c
chu11: haha,yeah I guess that would be simpler :-)

chu11:
re-pushed per @garlick comment above.  I'll set MWP after this passes.

Issue #3813, opened 2021-08-04
flux R append should consider GPUs when collapsing node types


garlick:
Problem: when R's with and without GPUs that are otherwise identical are passed to `flux R append`, they are collapsed without the GPUs.

to recreate:
```
flux R encode -r 0-1 -c 0-3 -H fluke[0-1] >R1
flux R encode -r 2-3 -c 0-7 -H fluke[2-3] >R2
flux R encode -r 4-5 -c 0-7 -g 0-3 -H fluke[4-5] >R3
cat R1 R2 R3 | flux R append >R
-
```
End result is
```json
{
  "version": 1,
  "execution": {
    "R_lite": [
      {
        "rank": "0-1",
        "children": {
          "core": "0-3"
        }
      },
      {
        "rank": "2-5",
        "children": {
          "core": "0-7"
        }
      }
    ],
    "starttime": 0,
    "expiration": 0,
    "nodelist": [
      "fluke[0-5]"
    ]
  }
}
```
I did verify that the GPUs were present in R2.

grondo:
Oops, yeah, that is definitely a bug!

garlick:
Mentioning @javawolfpack since he might bump into this and require a hand edit to work around when configuring his jetsons.

javawolfpack:
Thanks for the tag/heads up.

grondo:
I found the issue - when comparing two resource nodes, the present children of the first node are iterated and compared with the second node. If the first node doesn't have gpus then gpus aren't compared and the rank is considered identical.
Fix:
```diff
diff --git a/src/common/librlist/rnode.c b/src/common/librlist/rnode.c
index da7c12587..849770212 100644
--- a/src/common/librlist/rnode.c
+++ b/src/common/librlist/rnode.c
@@ -659,8 +659,15 @@ int rnode_cmp (const struct rnode *a, const struct rnode *b)
 {
     int rv = 0;
     /*  All avail idsets must match */
-    struct rnode_child *ca = zhashx_first (a->children);
+    struct rnode_child *ca;
     struct rnode_child *cb;
+
+    /* If two nodes do not have the same number of children types,
+     *  then they are not identical. Order doesn't matter here...
+     */
+    if (zhashx_size (a->children) != zhashx_size (b->children))
+        return -1;
+    ca = zhashx_first (a->children);
     while (ca) {
         if (!(cb = zhashx_lookup (b->children, ca->name)))
             return -1;
```

```console
[...]

PR #3814, opened 2021-08-05
rlist: fix erroneous collapse of nodes with different resource children when emitting R

grondo:
This fixes the problem described in #3813.

The problem was an embarrassing error in `rnode_cmp()`, which compares two resources nodes to determine if they have identical child resources. The function iterates each child resource type in the first node and compares idsets with the same resource type in the second node. However, if the 2nd node has extra resources, like gpus, then the node is considered equal and is collapsed into the first node and the gpus are lost.

The fix is to first compare the number of resource types and ensure they are equal before continuing on to further comparison.

### DIFF
--- a/src/common/librlist/rnode.c
+++ b/src/common/librlist/rnode.c
@@ -659,8 +659,15 @@ int rnode_cmp (const struct rnode *a, const struct rnode *b)
 {
     int rv = 0;
     /*  All avail idsets must match */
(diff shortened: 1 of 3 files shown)


garlick (approved):
Looks great!  Thanks for the quick fix.

grondo:
Thanks! Setting MWP

Issue #3800, opened 2021-07-27
broker: track RPC state and send error responses for lost peers


garlick:
When a broker peer is lost, there is no possibility for responses to be received for pending RPCs.  Brokers should track outstanding RPCs and send error responses when the next hop dies to avoid hangs.

PR #3822, opened 2021-08-15
broker: fail pending RPCs when TBON child goes down

garlick:
This creates a generic RPC tracking class in librouter, and uses it on TBON children in the broker, as described briefly in #3800 and in more detail [here](https://github.com/flux-framework/flux-core/blob/master/src/broker/doc/resiliency.md).

Each broker tracks the state of all RPC requests passing through it to its children.  In a multi-hop RPC, each hop in that direction creates state.  If a broker goes down, only that broker's TBON parent generates the error response, which then causes the state in other brokers to be freed like it would have been had the intended destination responded.

The overhead is:
- a zhashx entry for each sent request, containnig a reference on sent `flux_msg_t`, valid while the RPC is in flight
- zhashx lookup for each received response (key=uuid+matchtag)
-  full hash walk on receipt of a disconnect request

It feels to me like this should be pretty light overhead, both in terms of memory and cpu, but I'll do some testing to see if I can quantify it.

BTW, the `rpc_track` class was written with the idea that it could be reused for:
- loss of parent (causing pending rpcs to fail fast during subtree panic)
- temporary loss of broker connection in shell

Leaving this WIP for now as I work on testing.

### DIFF
--- a/src/broker/overlay.c
+++ b/src/broker/overlay.c
@@ -621,24 +621,24 @@ const char *overlay_get_bind_uri (struct overlay *ov)
     return ov->bind_uri;
 }
 
-/* When sending to a TBON child, zmq_sendmsg() failed with EHOSTUNREACH.
- * Since ROUTER socket has ZMQ_ROUTER_MANDATORY set, EHOSTUNREACH on a
- * connected peer signifies a disconnect.  See zmq_setsockopt(3).
- */
-static void overlay_sendmsg_child_noroute (struct overlay *ov,
-                                           const flux_msg_t *msg)
+static void overlay_child_status_update (struct overlay *ov,
+                                         struct child *child,
+                                         int status)
 {
-    const char *uuid;
-    struct child *child;
+    if (child->status != status) {
+        if (subtree_is_online (child->status)
+            && !subtree_is_online (status)) {
+            zhashx_delete (ov->child_hash, child->uuid);
+        }
+        else if (!subtree_is_online (child->status)
+            && subtree_is_online (status)) {
+            zhashx_insert (ov->child_hash, child->uuid, child);
+        }
 
-    if ((uuid = flux_msg_route_last (msg))
-        && (child = child_lookup (ov, uuid))
-        && subtree_is_online (child->status)) {
-        child->status = SUBTREE_STATUS_LOST;
+        child->status = status;
         monotime (&child->status_timestamp);
+
         subtree_status_update (ov);
-        overlay_child_log (ov, child, "send failed, child has disconnected");
-        zhashx_delete (ov->child_hash, child->uuid);
         overlay_monitor_notify (ov);
     }
 }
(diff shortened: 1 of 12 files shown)


garlick:
OK, I'm done playing with tests for a while, so this is probably ready for a review.

I'm running some throughput tests and thus far see no performance diff at all.  That may be because RPCs in the "downstream" direction are actually relatively few and far between with the thoughput tester workload.  (While monitoring rank 0, I haven't seen more than 5 at a time).

grondo:
> I'm running some throughput tests and thus far see no performance diff at all. That may be because RPCs in the "downstream" direction are actually relatively few and far between with the thoughput tester workload. (While monitoring rank 0, I haven't seen more than 5 at a time).

Interesting. I suppose that makes sense. The main downstream RPCs would likely be from the job-exec  module's use of the exec protocol to tell downstream brokers to launch job-shells for each job. If you are running in "test" execution mode, you wouldn't even have those RPCs..

Could a benchmark with the ping protocol give a better idea of any impact here?

chu11 (commented):
two nits i saw

garlick:
> Could a benchmark with the ping protocol give a better idea of any impact here?

Good thought!   like `time ./flux ping --count=1000000 -i 0 1`.  I'll try that.

garlick:
That was a good idea.  I can get about 300 sustained concurrent RPCs to one peer, peaking around 5000 or so, with that test.  

Master runs about 20s +/-1s.
This branch is maybe 20.5s +/- 1s.

That seems pretty reasonable.

grondo:
Yeah, nice work!

garlick:
Thanks!  One builder timed out on the shell input test, so restarted.

grondo (approved):
LGTM!

garlick:

garlick:
Coverage timed out at 40m.  I don't know if it's useful but I captured the last bit showing a gap in the output of around 35m:
```
2021-08-23T15:40:02.1448825Z make  check-TESTS
2021-08-23T15:40:02.2747643Z             test_service.t:  PASS: N=18  PASS=18  FAIL=0 SKIP=0 XPASS=0 XFAIL=0
2021-08-23T15:40:02.2756870Z                test_attr.t:  PASS: N=53  PASS=53  FAIL=0 SKIP=0 XPASS=0 XFAIL=0
2021-08-23T15:40:02.3392370Z             test_liblist.t:  PASS: N=6   PASS=6   FAIL=0 SKIP=0 XPASS=0 XFAIL=0
2021-08-23T15:40:02.3432096Z             test_pmiutil.t:  PASS: N=9   PASS=9   FAIL=0 SKIP=0 XPASS=0 XFAIL=0
2021-08-23T15:40:02.4152338Z         test_boot_config.t:  PASS: N=54  PASS=54  FAIL=0 SKIP=0 XPASS=0 XFAIL=0
2021-08-23T15:40:03.7729908Z             test_overlay.t:  PASS: N=78  PASS=78  FAIL=0 SKIP=0 XPASS=0 XFAIL=0
2021-08-23T16:15:06.4572336Z ##[error]The operation was canceled.
2021-08-23T16:15:06.4635510Z ##[group]Run src/test/checks-annotate.sh
2021-08-23T16:15:06.4636384Z [36;1msrc/test/checks-annotate.sh[0m
2021-08-23T16:15:06.4675075Z shell: /usr/bin/bash -e {0}
2021-08-23T16:15:06.4675479Z env:
2021-08-23T16:15:06.4675863Z   TAP_DRIVER_QUIET: 1
2021-08-23T16:15:06.4676273Z   FLUX_TEST_TIMEOUT: 300
2021-08-23T16:15:06.4676739Z   DOCKER_REPO: fluxrm/flux-core
2021-08-23T16:15:06.4677233Z   DOCKER_USERNAME: travisflux
2021-08-23T16:15:06.4677694Z   DOCKER_PASSWORD: 
2021-08-23T16:15:06.4679182Z   GITHUB_TOKEN: ***
2021-08-23T16:15:06.4679599Z   COVERAGE: t
2021-08-23T16:15:06.4679971Z ##[endgroup]
2021-08-23T16:15:06.7000682Z Found 0 errors from 126 tests in testsuite
2021-08-23T16:15:06.7192144Z ls: cannot access 't/*.trs': No such file or directory
2021-08-23T16:15:06.7240413Z ##[error]Detected 132 missing missing tests:
```
The tests that completed are broker unit tests.  All of them completed except `test_runat.t`.  Hmm...

Issue #3820, opened 2021-08-13
export a method to get task rank information in shell API


grondo:
Problem: Shell plugins (and possibly other users) may want to be able to get task rank information from the shell. At present, only the number of tasks on a shell rank is available via the `flux_shell_get_rankinfo(3)` JSON payload. Currently, job task ranks are assigned in a simple reproducible manner, but if there are ever alternate methods for rank assignment, then a method to get the rank mapping will be required.

Solution: Offer a new shell API call (or add additional information to `flux_shell_get_rankinfo(3)`) that unambiguously exports the task rank mapping to consumers. We may even want to save this information somehow in the job record so that task rank IDs can easily be mapped back to shell ranks, broker ranks, and hostnames after the fact.

grondo:
This also came up in  flux-framework/flux-pmix#14. In that PR @garlick built up an idset of task ranks for the current shell rank. This seems like a good approach. So how about we add a `taskids` member to the payload for `flux_shell_get_rank_info(3)` and `flux_shell_rank_info_unpack(3)`, i.e.:
```
/*  Return rank and task info for given shell rank as JSON string.
 *  {
 *   "broker_rank":i,
 *   "ntasks":i
 *   "taskids"s     # RFC 22 IDset string of task ids on this shell rank
 *   "resources": { "cores":s, ... }
 *  }
 */
```

garlick:
Yes I think that would be useful!

Issue #2414, opened 2019-09-30
support MPMD execution


garlick:
Rich Drake for SNL's SIERRA code:
> Searching Sierra's launch scripts, I only see the use of these:

> -n

> --multi-prog

> And the ":" delimiter for MPMD execution.

> Of course, if binding needs to be specified, then we will use that.

Seems MPMD support is a gap... any idea how easy or difficult to match this srun option?

_Originally posted by @dongahn in https://github.com/flux-framework/flux-core/issues/2150#issuecomment-492781807_

SteVwonder:
Other comments from that thread on SIERRA's specific use of MPMD:

>> We use --multi-prog on ATS-1 for all our MPMD-type coupled executions (I think all examples are two codes in a partitioned core set).  If there was a way to run these without using --multi-prog, I'm sure we could make that work.

@SteVwonder:
> @garlick and I just chatted briefly about that. We could probably mimic the exact behavior with a wreck plugin, but we could also leverage nested instances to achieve the co-scheduling. In the nested instance case, we (or the user) would just need to make sure that the total amount of work submitted to the sub-instance is not greater than the resources allocated to the instance can handle (i.e., ntasks cannot be greater than ncores). We could probably wrap that logic up in a flux multiprog command, or even add it as a flag to flux capacitor

@dongahn:
> Yeah it seems like Rich is alluding that he can make use of a general co-scheduling capability. For oneoff option like this, it would be wise to invite the users like Rich to firm up our solution as co-design. Looks like we will have to divide and conquer across different SNL teams a bit for effective communications going forward.

@garlick:
> Great info we are accumulating here. It is sort of difficult to decide what options to support. The goal is to provide a stable porting target not an srun clone.

> My suggestion is to start with the options that are supported in flux jobspec and make a super simple wrapper for the plumbing commands in master, and backport those to a 0.11 script that translates to wreckrun options.

> Possibly this will help us identify some missing plumbing in master for synchronization, I/O, etc. that will be good short term work items.

SteVwonder:
In a recent concall with the Radical Pilot folks, they expressed interest in being able to specify whether the various programs are a part of a single global communicator or if each program has its own communicator.  I'm not sure how feasible that latter bit is, but I at least wanted to document it before forgetting.

grondo:
[...]

PR #3826, opened 2021-08-16
Shell: libpals interface plugin

### LINKED ISSUE #3825: Cray PMI (libpals) support
In order for Flux to support Cray's PMI on Shasta machines like rznevada, Perlmutter, and El Cap, Flux will need to interface with Cray's libpals. From the Cray folks:

> To support Cray PMI, the launcher must perform these tasks:
> 1. Assign an apid to the application (arbitrary string, unique per-application)
> 2. On each compute node, create a spool directory owned by the application's user
> 3. On each compute node, write an apinfo file in the spool directory (format below)
> 4. Set environment variables for each spawned process (listed below)
> 5. Remove the spool directory on each compute node when the application is complete
>  
> The apinfo file uses C structures written directly to the file. It starts with a header, followed by a number of structure arrays.
>  
> These environment variables should be set for each process:
> PALS_APID - Application ID (arbitrary string, mostly used for logging)
> PALS_APINFO - Full path to the apinfo file
> PALS_RANKID - Global rank ID for this process
> PALS_NODEID - Node index for this process (e.g. head compute node is 0, next compute node is 1, etc)
> PALS_SPOOL_DIR - Application-specific directory for keeping runtime files
> PMI_CONTROL_PORT - Port number for PMI to bind on each compute node. Must be different for each concurrent application running on the same node(s).

jameshcorbett:
Closes #3825.

Many thanks to @grondo for helping me with the build issues I was having, and for contributing the two commits. We were discussing in Slack how best to make the plugin optional and he said:

> We don't really have any .so plugins for the shell yet, so we could remove that line that loads all plugins by default.
> 
> Another option would be to allow the plugin to be loaded, but require a shell option to enable the pals plugin operation. This is how many of the builtin plugins work, since they are all always loaded.
> 
> A third option would be to separately package the pals plugin and only install it on systems where it will be used (I think that was the original vision of how things would work, but now I realize that may not be possible in all cases)

Since this plugin should always run on certain systems, I think it would be best to load it on an instance-wide level or at build time.

Tests are still needed, plus a good way to determine the number of cores per task for the job. Also, there are some VNI/NIC information that may still need to be added---will have to consult with HPE people.

The plugin makes assumptions about task placement, so would be affected by #3820. Also, some of the information exported to libpals would change if MPMD (#2414) was an option.


jameshcorbett:
Demo:

```
[corbett8@rznevada15:flux-core]$ git checkout cray-mpi && make clean > /dev/null && make -j > /dev/null && src/cmd/flux start flux mini run -n2 ../mpi_hello
Hello from task 1 on rznevada15!
Hello from task 0 on rznevada15!
MASTER: Number of MPI tasks is: 2
[corbett8@rznevada15:flux-core]$ git checkout master && make clean > /dev/null && make -j > /dev/null && src/cmd/flux start flux mini run -n2 ../mpi_hello
Hello from task 0 on rznevada15!
Hello from task 0 on rznevada15!
MASTER: Number of MPI tasks is: 1
MASTER: Number of MPI tasks is: 1
```

SteVwonder:
> Since this plugin should always run on certain systems, I think it would be best to load it on an instance-wide level or at build time.

I'll defer to you and the other flux-core devs as to the best way to handle optionally including/building/installing this plugin,  but if we want to go the machine-specific route (i.e., you install this plugin instance-wide only on specific machines), then https://github.com/flux-framework/flux-coral2 might be a good place for this plugin to live.

grondo:
>  https://github.com/flux-framework/flux-coral2 might be a good place for this plugin to live.

I was also just thinking that perhaps this plugin does not necessarily belong in flux-core (Sorry for dropping this on you after you already went through all the work of integrating the plugin into our autotools build, @jameshcorbett!)

Packaging the plugin separately may have the following benefits (besides machine-specific installation as already noted by @SteVwonder):

 * if for some reason you want to release versions of the pals plugin more rapidly than the flux-core release cycle, it will be simple to do that
 * setting up a libpals specific CI environment for all of flux-core may be onerous, this may be simpler (if even possible) in a standalone package or under flux-coral2 as @SteVwonder suggests.


Sorry I didn't consider this while we were struggling with the flux-core build system @jameshcorbett!

src/shell/plugins/pals.c
```diff
@@ -0,0 +1,493 @@
...
+#include <flux/hostlist.h>
+#include <flux/shell.h>
+
+#include "jansson.h"
+
+#include "jobspec.h"
+
+/*  PALS --- interface with HPE/Cray's PMI.
+ *
+ * To support Cray PMI, the launcher must perform these tasks:
```
garlick, line 26: Some general comments:
* perhaps the actual libpals could be used to write tests for this?
* maybe include cray in the name?  Like `cray-pals`?  Otherwise the name doesn't provide much useful context.
* a bit more description in the header about the overall architecture would be useful, since usually PMI is a service provided by the launcher; in this case it is a proprietary service that we are just handing off job information to (as I understand it!)
* it would possibly be good to push back on cray on the naming of the port variable since a) env vars beginning with `PMI_` are used by PMI and this is operating at a different level, and b) apparently it's going to be  comma-separated list of ports, so it should be plural.
* seems like it may be a good idea to locate this in the other repo as suggested by @SteVwonder.  That would allow you to update it frequently during early deployment, without requiring a flux-core release.

jameshcorbett: Thanks for the comments, I agree with all of them. Cray would definitely have the upper hand in the environment variable argument because they're already [built into Slurm](https://github.com/SchedMD/slurm/blob/6cb01bd89a92c7eaa80624891c034b2af85727d4/src/plugins/mpi/cray_shasta/mpi_cray_shasta.c#L171) though.

jameshcorbett:
Sounds like everyone is on board with moving the PR over to flux-coral2 so I will go ahead and do that.

> Sorry I didn't consider this while we were struggling with the flux-core build system @jameshcorbett!

No problem, you didn't have all the context. I had figured this might happen anyway.

PR #3827, opened 2021-08-17
add bootstrap.ipv6_enable = true config option

### LINKED ISSUE #3824: broker: zeromq ipv6 flag prevents interface binding if ipv6 is off
On fluke, setting the bind address to e.g. `tcp://mgmt:8050` fails with ENODEV (No such device).

This error is coming from `zsock_bind()`, which internally calls `zmq_bind()`.  When `zmq_bind()` looks up an interface, it calls `getifaddrs()` and then walks the result with
```c
    //  Find the corresponding network interface.
    bool found = false;
    for (const ifaddrs *ifp = ifa; ifp != NULL; ifp = ifp->ifa_next) {
        if (ifp->ifa_addr == NULL)
            continue;

        const int family = ifp->ifa_addr->sa_family;
        if (family == (_options.ipv6 () ? AF_INET6 : AF_INET)
            && !strcmp (nic_, ifp->ifa_name)) {
            memcpy (ip_addr_, ifp->ifa_addr,
                    (family == AF_INET) ? sizeof (struct sockaddr_in)
                                        : sizeof (struct sockaddr_in6));
            found = true;
            break;
        }
    }
```

Since we call `zsys_set_ipv6 (1)` unconditionally in the broker, and the interface named `mgmt` has no IPV6 address assigned, this code doesn't find it.

The ipv6 option was set in PR #3489, commit 0d3ca76e827ad1f2952d58c484792d186e587cd4

Solution: provide a way to disable the ipv6 option in the system instance config, as this seems like the main scenario where you'd be binding to an interface, as opposed to using the "primary address" heuristic used by PMI bootstrap.

garlick:
Problem: as noted in #3824, the current default of enabling ipv6 for all zeromq sockets prevents the flux system instance from binding to an interface that only has ipv4 addresses.  We just hit this on fluke, and in fact this bug is a blocker for getting the system instance updated on fluke.

This PR disables ipv6 for config bootstrap only, and adds a config setting for overriding the default, should that be necessary:
  ```toml  
 [bootstrap]
 enable_ipv6 = true
``` 
    
Fixes #3824

### DIFF
--- a/doc/man5/flux-config-bootstrap.rst
+++ b/doc/man5/flux-config-bootstrap.rst
@@ -29,6 +29,11 @@ expected to bootstrap from an identical config file.
 KEYWORDS
 ========
 
+enable_ipv6
+   (optional) Boolean value for enabling IPv6.  By default only IPv4 is
+   enabled.  Note that setting this to true prevents binding to a named
+   interface that only supports IPv4.
+
 curve_cert
    (optional) Path to a CURVE certificate generated with flux-keygen(1).
    The certificate should be identical on all broker ranks.
--- a/doc/test/spell.en.pws
+++ b/doc/test/spell.en.pws
@@ -534,3 +534,5 @@ afternotok
 parsedatetime
 cancelall
 raiseall
+IPv4
+IPv6
--- a/src/broker/boot_config.c
+++ b/src/broker/boot_config.c
@@ -510,10 +510,8 @@ int boot_config (flux_t *h, struct overlay *overlay, attr_t *attrs)
                                        bind_uri,
                                        sizeof (bind_uri)) < 0)
             goto error;
-        if (overlay_bind (overlay, bind_uri) < 0) {
-            log_err ("overlay_bind");
+        if (overlay_bind (overlay, bind_uri) < 0)
             goto error;
-        }
         if (overlay_authorize (overlay,
                                overlay_cert_name (overlay),
                                overlay_cert_pubkey (overlay)) < 0) {
--- a/src/broker/boot_config.c
+++ b/src/broker/boot_config.c
@@ -183,13 +183,14 @@ int boot_config_parse (const flux_conf_t *cf,
     memset (conf, 0, sizeof (*conf));
     if (flux_conf_unpack (cf,
                           &error,
-                          "{s:{s?s s?i s?s s?s s?o}}",
+                          "{s:{s?s s?i s?s s?s s?o s?b}}",
                           "bootstrap",
                             "curve_cert", &conf->curve_cert,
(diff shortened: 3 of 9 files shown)

### CODE REVIEW

File: doc/man5/flux-config-bootstrap.rst [hunk 1]
```diff
@@ -29,6 +29,11 @@ expected to bootstrap from an identical config file.
 KEYWORDS
 ========
 
+enable_ipv6
+   (optional) Boolean value for enabling IPv6.  By default only IPv4 is
+   enabled.  Note that setting this to true prevents binding to a named
+   interface that only supports IPv4.
```
grondo, line 35: Do you feel the current behavior of zeromq is a bug? If so, maybe we should open an issue against zeromq and reference that issue here (Saying "on versions of zeromq affected by the following issue...")

grondo (approved):
A naive question (besides the other naive question inline) - Does this change the behavior of non-system instances as well to be ipv4 only by default?

Besides these two questions, LGTM.

doc/man5/flux-config-bootstrap.rst
garlick: Not sure.  I think maybe I just didn't anticipate this particular outcome?  The extent of 0MQ's documentation appears to be (From zmq_setsockopt(3)):
```
   ZMQ_IPV6: Enable IPv6 on socket
       Set the IPv6 option for the socket. A value of 1 means IPv6 is enabled
       on the socket, while 0 means the socket will use only IPv4. When IPv6
       is enabled the socket will connect to, or accept connections from, both
       IPv4 and IPv6 hosts.
```
I could see the argument that explicitly enabling ipv6 on the 0MQ socket, then binding it to an _interface_ that doesn't support ipv6 should be a hard error.  Note that it's not an error to bind to an IPv4 _address_ with ipv6 enabled.

To answer your other question, everything in the `[bootstrap]` stanza only applies to the config bootstrap method, not to the PMI bootstrap method.  Maybe that should be made clearer in the man page update.

grondo: > To answer your other question, everything in the [bootstrap] stanza only applies to the config bootstrap method, not to the PMI bootstrap method. Maybe that should be made clearer in the man page update.

Thanks, that may not be necessary. I misread the changes to `boot_pmi.c` (or didn't actually look at them) and assumed something in the PMI bootstrap had changed. Excuse my mistake!

garlick:
Coverage is low mainly due to the increased error logging verbosity.

OK to set MWP?

grondo:
Yes, sorry I should have mentioned set MWP when ready

PR #3835, opened 2021-08-24
WIP: shell: add functions to access jobspec summary information

grondo:
This is a WIP PR which adds a simple interface to the jobs shell API to fetch jobspec "summary" information (as described in #3831).

This is pretty simple, but I wanted to put a WIP PR up before adding any more testing, basically to get an ack that the interface is ok. @jameshcorbett

PR #3839, opened 2021-08-24
python: fix bug in FluxExecutor.attach method

jameshcorbett:
Problem: when a nonexistent jobid is passed to FluxExecutor.attach(), the executor should catch a `FileNotFoundException` and error out the future. However, along that line of control flow a variable is undefined (`UnboundLocalError`), causing the executor to crash. Fix: a single-line change to define the variable. Add tests to catch the problem.

For an example of the bug see https://github.com/flux-framework/flux-core/pull/3835#issuecomment-904735399


grondo (approved):
Wow, quick work! I appreciate the additional tests!
Thanks!

PR #3840, opened 2021-08-25
testsuite: make valgrind test opt-in

grondo:
The `t5000-valgrind.t` test is prone to spurious failures on new systems and distros due to false positives in system libraries, etc. This can cause confusion when new users run `make check`.

Disable `t5000-valgrind.t` by default, unless it is run with the `-d, --debug` flag (this allows it to be run by hand as `./t5000-valgrind.t -d`), or if `FLUX_ENABLE_VALGRIND_TEST` is set in the environment. The environment variable is then set in `checks_run.sh` so that the valgrind test is always run under `docker-run-checks.sh` and therefore the CI.


garlick (approved):
Thanks for this!  This should avoid a lot of busy work adding suppressions for not-our-problems.

Issue #3805, opened 2021-07-29
need ability to force disconnect from broker peer


garlick:
As we debug broker resiliency for the system instance, it may be useful for testing or to trigger recovery in a hung instance, to be able to force a broker peer to "panic" and restart its subtree, as though its parent had crashed.

This may be useful in conjunction with #2797

PR #3843, opened 2021-08-27
broker: fail pending RPCs when TBON parent goes down

garlick:
This implements broker RPC tracking in the upstream direction, analogous to the downstream tracking added in #3822.  The `rpc_track` class added in that PR was reused. 

With this component completed, we now have the machinery in place to avoid RPC hangs resulting from lost broker connections.  The next piece is to activate the machinery when we detect an uncommunicative broker or other unrecoverable broker failure mode.

A `flux overlay disconnect RANK` command was added that allows a sys admin to force disconnect a broker,  as proposed in #3805.   Tests for this PR are built upon this tool.

Code was added as part of `flux overlay disconnect` to determine the parent of the target rank by querying the topology from rank 0 and searching it.  To make this easier to test, a subcommand was added that exposes the topology lookup directly.  Maybe it'll be useful in other contexts:
```
$ flux overlay parentof 4
1
```
Also, while doing a bit of refactoring in the command, I updated `flux overlay status` to map ranks to hostnames using `resource.R` if `config.hostlist` is unavailable, as suggested by @grondo.  That mapping is exposed directly for testing in another subcommand:
```
$ flux overlay gethostbyrank 4
system76-pc
```
I've left this as a WIP just so I have a chance to cogitate on it a bit over the weekend, and decide if I've missed anything.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -322,6 +322,7 @@ AM_CONDITIONAL([ENABLE_PYLINT], [test "x$PYLINT" = "xpylint"])
 AX_PROG_LUA([5.1],[5.5])
 AX_LUA_HEADERS
 AX_LUA_LIBS
+# N.B oldest in CI are bionic=libzmq-4.2.5, centos7=zeromq-4.1.4
 PKG_CHECK_MODULES([ZMQ], [libczmq >= 3.0.0 libzmq >= 4.0.4])
 X_AC_JANSSON
 PKG_CHECK_MODULES([HWLOC], [hwloc >= 1.11.1], [], [])
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -234,9 +234,6 @@ int main (int argc, char *argv[])
     zsys_set_logstream (stderr);
     zsys_set_logident ("flux-broker");
     zsys_handler_set (NULL);
-    zsys_set_linger (5);
-    zsys_set_rcvhwm (0);
-    zsys_set_sndhwm (0);
 
     /* Set up the flux reactor with support for child watchers.
      * Associate an internal flux_t handle with the reactor.
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -225,7 +225,19 @@ int main (int argc, char *argv[])
         || sigaction (SIGTERM, NULL, &old_sigact_term) < 0)
         log_err_exit ("error setting signal mask");
 
-    /* Initailize zeromq context
+    /* Initialize libczmq zsys class.
+     *
+     * zsys_init() creates a global 0MQ context and starts the 0MQ I/O thread.
+     * The context is implicitly shared by users of the zsock class within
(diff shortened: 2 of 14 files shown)


garlick:
I tried to do a little bit of performance testing to see if this has an impact.  As with the downstream RPC tracking, there doesn't seem to be a measurable impact on job throughput.  I resorted to a ping test with 16 ranks lined up in a fanout=1, pinging from rank 15 t0 0 with `-i0 -c100000` .  With this branch, around 12s.  Master, about 11.5s.  (repeated 3x each).

So maybe a little impact but not much, for that test anyway.

garlick:
BTW, it might be worth noting that none of the traffic that stays on rank 0 that is critical to job throughput, such as among `job-manager`, `sched-*`, `job-exec`, `kvs`, and `content-cache` is subject to this tracking.

grondo:
Just experimenting with this branch and everything LGTM so far!

A couple things I found in testing that I just want to ensure are expected behavior:

1. When disconnecting a subtree, this error is always printed. Seems like it could confuse users, but not sure how it could be suppressed. I wonder if the disconnecting client is trying to log something _after_ being disconnected?
  ```
  $ flux overlay disconnect 3
  flux-overlay: asking rank 1 to disconnect child rank 3
  2021-09-01T19:42:50.205047Z broker.err[1]: DROP downstream request topic log.append from e2e01592-9964-4a7e-825e-264be5c1e5e3: message from unknown peer
 ```
2. Should `flux overlay disconnect RANK` ensure that `RANK` is not already disconnected before attempting to send a new disconnect request?
 ```
 $ flux overlay disconnect 3
 flux-overlay: asking rank 1 to disconnect child rank 3
 flux-overlay: overlay.disconnect-subtree: failed to send KEEPALIVE_DISCONNECT message
 ``` 
 Alternately, just printing `flux-overlay: child rank 3 is already disconnected` might be slightly preferrable.

garlick:
> 2021-09-01T19:42:50.205047Z broker.err[1]: DROP downstream request topic log.append from e2e01592-9964-4a7e-825e-264be5c1e5e3: message from unknown peer

> I wonder if the disconnecting client is trying to log something after being disconnected?

I think these are messages that the disconnecting node had sent before receiving the disconnect request.

In the next PR (#3845)  4ebe51c8331f3a9fcdf7bf285583a9822408bcbd suppresses them.

> Alternately, just printing flux-overlay: child rank 3 is already disconnected might be slightly preferrable.

That sounds like a nice improvement.  I'll do that and push it here.

grondo:
> the next PR (#3845) 4ebe51c suppresses them.

Oops! Sorry I missed that :facepalm:

garlick:
No worries, and thanks!  I'll squash those last couple changes down, then set MWP if that's OK.

I have a commit message change queued up for #3845, which I'll rebase and push after this goes in.

grondo:
Yeah, squash rebase and MWP away!

PR #3844, opened 2021-08-28
ci: add ability to run tests under system instance

grondo:
This PR adds a tweak to the fluxorama docker image along with a simple script to make iterative development and test easier within this systemd based docker container.

Currently, the script only launches an interactive shell within the container after launching it. In the future, though, the script could be extended to run a suite of multi-user specific tests.

I'm posting this a little early since @chu11 may find the script useful in his current work.

The script first optionally rebuilds the `fluxrm/flux-core:centos8` image from the current build directory (only if `--rebuild` option is used). It then builds the `systest` stage of the `fluxorama` image (an image without ttyd enabled) and starts the container in the background. An interactive shell is started in the container as the current user (just like `docker-run-checks.sh -I`). Once the shell exits, the running container is stopped with `docker exec shutdown -r now`)

example:
```console
grondo@asp:~/git/f.git$ ./src/test/docker/docker-run-systest.sh 
Building system image for user grondo 1000 group=1000
Sending build context to Docker daemon  30.21kB
Step 1/20 : FROM fluxrm/flux-core:centos8 AS systest
[snip]...

Successfully built 1f9a4a0f1e07
Successfully tagged fluxorama:systest

af1147e6f79cb51450b23ff62006eb36d30630fe551584063eaa379900a4110a
 grondo@fluxorama:/usr/src$ flux resource list
      free      1        4        0 fluxorama
      down      0        0        0 
 grondo@fluxorama:/usr/src$ sudo systemctl status flux
   Loaded: loaded (/etc/systemd/system/flux.service; enabled; vendor preset: di>
   Active: active (running) since Sat 2021-08-28 02:47:49 UTC; 2min 27s ago
  Process: 36 ExecStartPre=/bin/chmod 0700 /var/lib/flux (code=exited, status=0>
  Process: 35 ExecStartPre=/bin/chown flux:flux /var/lib/flux (code=exited, sta>
  Process: 27 ExecStartPre=/bin/mkdir -p /var/lib/flux (code=exited, status=0/S>
 Main PID: 37 (flux-broker-0)
    Tasks: 17 (limit: 26213)
   Memory: 7.6M
   CGroup: /system.slice/snap.docker.dockerd.service/system.slice/flux.service
           └─37 broker --config-path=/etc/flux/system/conf.d -Stbon.fanout=256 >

Aug 28 02:47:49 fluxorama flux[37]: job-manager.info[0]: restart: 0 jobs
Aug 28 02:47:49 fluxorama flux[37]: job-manager.info[0]: restart: 0 running jobs
Aug 28 02:47:49 fluxorama flux[37]: job-manager.info[0]: restart: no checkpoint>
Aug 28 02:47:49 fluxorama flux[37]: broker.info[0]: rc1.0: running /etc/flux/rc>
Aug 28 02:47:49 fluxorama flux[37]: sched-simple.info[0]: resource update: {"re>
Aug 28 02:47:49 fluxorama flux[37]: broker.info[0]: rc1.0: /bin/bash -c /etc/fl>
Aug 28 02:47:49 fluxorama flux[37]: broker.info[0]: rc1-success: init->quorum 0>
Aug 28 02:47:49 fluxorama flux[37]: broker.info[0]: quorum-full: quorum->run 4.>
Aug 28 02:47:49 fluxorama flux[37]: broker.info[0]: rc2-none: ignored in run
Aug 28 02:50:15 fluxorama systemd[1]: flux.service: Failed to reset devices.lis>

```

### DIFF
--- a/src/test/docker/docker-run-systest.sh
+++ b/src/test/docker/docker-run-systest.sh
@@ -168,4 +168,24 @@ if test $RC -ne 0; then
     die "system tests failed with rc=$RC"
 fi
 
+if test -n "$CI" -a -n "$COVERAGE"; then
+    curl -Os https://uploader.codecov.io/latest/linux/codecov &&
+    chmod +x codecov &&
+    docker exec \
+      -e GCOV \
+      -e USER \
+      -e CI \
+      -e PROJECT \
+      -e HOME=/home/$USER \
+      -e GITHUB_ACTIONS \
+      -e GITHUB_HEAD_REF \
+      -e GITHUB_REF \
+      -e GITHUB_REPOSITORY \
+      -e GITHUB_RUN_ID \
+      -e GITHUB_SERVER_URL \
+      -e GITHUB_SHA \
+      -e GITHUB_WORKFLOW \
+      -w $WORKDIR flux-system-test-$$ ./codecov -F ci-system
+fi
+
 # vi: ts=4 sw=4 expandtab
--- a/src/test/docker/docker-run-systest.sh
+++ b/src/test/docker/docker-run-systest.sh
@@ -1,0 +1,171 @@
+#!/bin/bash
+#
+#  Build fluxorama image with builddir mounted as /usr/src for testing.
+#
+
+PROJECT=flux-core
+WORKDIR=/usr/src
+MOUNT_HOME_ARGS="--volume=$HOME:/home/$USER -e HOME"
+JOBS=2
+IMAGE="fluxrm/flux-core:centos8"
+
+declare -r prog=${0##*/}
+die() { echo -e "$prog: $@"; exit 1; }
+
+declare -r long_opts="help,no-home,no-cache,rebuild,jobs:,image:"
+declare -r short_opts="hrj:i:"
+declare -r usage="
+Usage: $prog [OPTIONS]\n\
+Build fluxorama system test docker image for CI builds\n\
+\n\
+Options:\n\
+ -h, --help                    Display this message\n\
+ -j, --jobs=N                  Value for make -j (default=$JOBS)\n\
+ -i, --image=NAME              Base image (default=$IMAGE)\n\
+     --rebuild                 Rebuild base fluxorama image from source\n\
+     --no-home                 Skip mounting the host home directory\n\
+     --no-cache                Run docker build with --no-cache option\n\
+"
+
+# check if running in OSX
+if [[ "$(uname)" == "Darwin" ]]; then
+    # BSD getopt
+    GETOPTS=`/usr/bin/getopt $short_opts -- $*`
+else
+    # GNU getopt
+    GETOPTS=`/usr/bin/getopt -u -o $short_opts -l $long_opts -n $prog -- $@`
+    if [[ $? != 0 ]]; then
+        die "$usage"
+    fi
+    eval set -- "$GETOPTS"
+fi
+
+while true; do
+    case "$1" in
+      -h|--help)     echo -ne "$usage";          exit 0  ;;
+      -j|--jobs)     JOBS="$2";                  shift 2 ;;
+      -i|--image)    IMAGE="$2";                 shift 2 ;;
+      --rebuild)     REBUILD_BASE_IMAGE=t;       shift   ;;
+      --no-home)     MOUNT_HOME_ARGS="";         shift   ;;
+      --no-cache)    NOCACHE="--no-cache";       shift   ;;
+      --)            shift; break;                       ;;
+      *)             die "Invalid option '$1'\n$usage"   ;;
+    esac
+done
+
+if test $# -eq 0; then
+    set "bash"
+fi
+
+TOP=$(git rev-parse --show-toplevel 2>&1) \
+    || die "not inside $PROJECT git repository!"
+which docker >/dev/null \
+    || die "unable to find docker binary!"
+
+. ${TOP}/src/test/checks-lib.sh
+
+if test "$REBUILD_BASE_IMAGE" = "t"; then
+    checks_group "Rebuilding fluxrm/flux-core:centos8 from source" \
+      $TOP/src/test/docker/docker-run-checks.sh \
+        -j $JOBS \
+        -i centos8 \
+        -t $IMAGE \
+        --install-only
+fi
+
+checks_group "Building system image for user $USER $(id -u) group=$(id -g)" \
+  docker build \
+    ${NOCACHE} \
+    --build-arg IMAGE=$IMAGE \
+    --build-arg USER=$USER \
+    --build-arg UID=$(id -u) \
+    --build-arg GID=$(id -g) \
+    -t fluxorama:systest \
+    --target=systest \
+    ${TOP}/src/test/docker/fluxorama \
+    || die "docker build failed"
+
+NAME=flux-system-test-$$
+checks_group "Launching system instance container $NAME" \
+  docker run -d --rm \
+    --hostname=fluxorama \
+    --workdir=$WORKDIR \
+    $MOUNT_HOME_ARGS \
+    --volume=$TOP:$WORKDIR \
+    --volume=/sys/fs/cgroup:/sys/fs/cgroup:ro \
+    --tmpfs=/run \
+    --cap-add SYS_PTRACE \
+    --name=flux-system-test-$$ \
+    --network=host \
+    fluxorama:systest \
+    || die "docker run of fluxorama test container failed"
+
+until docker exec -u $USER:$GID \
+    flux-system-test-$$ flux mini run hostname 2>/dev/null; do
+    echo "Waiting for flux-system-test-$$ to be ready"
+    sleep 1
+done
+
+#  Start user@uid.service service for unit tests:
+#
+checks_group "Starting user service user@$(id -u).service" \
+  docker exec flux-system-test-$$ \
+    systemctl start user@$(id -u).service \
+    || die "docker start user@$(id -u).service failed"
+
+if test -t 0; then
+    INTERACTIVE="-ti"
+fi
+
+checks_group "Executing tests under system instance container" \
+  docker exec \
+    "${INTERACTIVE}" \
+    -u $USER:$GID \
+    ${CC+-e CC=$CC} \
+    ${CXX+-e CXX=$CXX} \
+    ${LDFLAGS+-e LDFLAGS=$LDFLAGS} \
+    ${CFLAGS+-e CFLAGS=$CFLAGS} \
+    ${CPPFLAGS+-e CPPFLAGS=$CPPFLAGS} \
+    -e PS1 \
+    -e GCOV \
+    -e CCACHE_CPP2 \
+    -e CCACHE_READONLY \
+    -e COVERAGE \
+    -e TEST_INSTALL \
+    -e CPPCHECK \
+    -e DISTCHECK \
+    -e RECHECK \
+    -e UNIT_TEST_ONLY \
+    -e chain_lint \
+    -e JOBS \
+    -e USER \
+    -e PROJECT \
+    -e CI \
+    -e TAP_DRIVER_QUIET \
+    -e TEST_CHECK_PREREQS \
+    -e FLUX_TEST_TIMEOUT \
+    -e FLUX_TEST_SIZE_MAX \
+    -e PYTHON_VERSION \
+    -e PRELOAD \
+    -e POISON \
+    -e INCEPTION \
+    -e ASAN_OPTIONS \
+    -e BUILD_DIR \
+    -e S3_ACCESS_KEY_ID \
+    -e S3_SECRET_ACCESS_KEY \
+    -e S3_HOSTNAME \
+    -e S3_BUCKET \
+    -e HOME=/home/$USER \
+    -e XDG_RUNTIME_DIR=/run/user/$(id -u) \
+    -e DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/$(id -u)/bus \
+    -w $WORKDIR \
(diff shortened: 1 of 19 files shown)


chu11:
Added a commit on top of this PR to add a `--shell` option to the systest script, so us poor souls who default to non-bash can tell the script to launch with bash :-)

grondo:
I pushed a fixup commit here that fixes an issue @chu11 found (The `docker exec` command was setting `CC` `CFLAGS`, etc. as empty in the environment of the container)

Additionally, the script now allows an arbitrary command to be passed instead of running an interactive shell. The shell also defaults to `bash` instead of `$SHELL`.

grondo:
I've updated this PR to include a lot of groundwork to get a build in CI running under the system test docker container (fluxorama). Essentially, a new `--system` option is added to `docker-run-checks.sh` which results in the following steps:

 1. The current working tree is built and installed in the `centos8` docker image, so that the system instance flux-core code matches the working tree
 2. The `docker-run-checks.sh` script then runs `docker-run-systest.sh src/test/checks_run.sh` on the resulting image, which:
 3.  A temporary fluxorama image is created using the "systest" stage
 4. The system test container is started, and a user@uid.service service enabled for the current user
 5. Finally, `checks_run.sh` is run inside the system instance

For now, it is probably enough that systemd exists and a `user@uid.service` systemd instance is available to move our system instance testing along. 

Later, we'll have to figure out a way to run multiuser tests within this container. (All the pieces are there except how exactly to develop these tests and run them)

For now I've enabled a centos8 "system" builder to the CI including coverage. There was a coverage drop when switching to centos8 which I have not been able to diagnose. So I've left the ubuntu coverage builder enabled since supposedly codecov supports collecting multiple coverage reports for the same build into one report. (We'll see how that goes)

grondo:
Oops, looks like I missed a need for `math.floor`
```
cron-5: stats.count=1, wanted 1
  lua: /usr/src/src/cmd/flux-cron:99: bad argument #2 to 'fmt' (number has no integer representation)
```

Also, this is a little concerning:
```
flux-module: /usr/src/src/modules/job-manager/plugins/.libs/alloc-bypass.so: undefined symbol: flux_jobtap_job_set_flag
```

:confused:

grondo:
Clearly codecov doesn't merge the two coverage reports (see above, the first comment states 0.01% drop in the coverage, and the second -0.43%, but 0.43% is what is reported to github)

Possibly the drop in coverage is due to the fact we can't use `--enable-pmix-bootstrap` in the centos8 image, though I see no evidence of that in the actual coverage report. The issue could also just be differences due to a different compiler... 

Since the reports are not merged, I wonder if we should get rid of the current coverage build since it is duplicating the new system test coverage builder?

grondo:
I think this is ready for a review. I'm thinking the coverage drop is due to compiler and/or gcov differences when moving from gcc 7.x on bionic to gcc 8.5 on centos8. I haven't yet found an exact explanation, and I didn't want to spend too much time on it, but I did spot check some files that Codecov marked as dropping >4% coverage, and all the missed lines were still in error handling paths (it didn't appear that any tests were missed). I suppose this could be due to optimization differences  or what lines are emitted in gcov output or which are not.

I did an experiment with `-O0` and this caused a more precipitous drop in coverage of ~4%. I think this is because compilation without any optimization probably results in more "lines" emitted that would have been combined by optimization, and many of the "new" lines were not covered due to error paths, etc. (I'm just guessing, definitely have no expertise in this area)

We could keep running the coverage build on bionic, but many new features will likely require a system-instance to test, so we would not get coverage reports for those PRs. The main benefit of coverage tests is getting a clue when some new code is not covered, and moving the coverage testing to the "system" test should still give us that. (If not, then we could also bring back the other coverage builder)

garlick (approved):
This will be super helpful I think!

I ran  `docker-run-systest.sh`  manually and just came up with one suggestion, commented inline.

It looks like the bionic coverage builder is still enabled - was that intended?

Just to be sure  I understand, right now tests are not using the actual fluxorama system instance, correct?  Although the system instance is running in the container and the sharness tests change paths so that the installed paths are found before the source tree ones, all the flux instances under test are still ephemeral ones started with `test_under_flux`?  If I wanted to write a sharness test, say to make sure that guests can't do certain things in the fluxorama instance, what would that look like?

Approving since my one comment is a nit and the above is a question not a critique.

src/test/docker/docker-run-systest.sh
```diff
@@ -0,0 +1,171 @@
...
+checks_group "Building system image for user $USER $(id -u) group=$(id -g)" \
+  docker build \
+    ${NOCACHE} \
+    --build-arg IMAGE=$IMAGE \
+    --build-arg USER=$USER \
+    --build-arg UID=$(id -u) \
+    --build-arg GID=$(id -g) \
+    -t fluxorama:systest \
+    --target=systest \
+    src/test/docker/fluxorama \
```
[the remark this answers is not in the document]
grondo: Yeah, good catch. I only ever run `docker-run-checks.sh` from the top level so I didn't discover this (and in fact I actually assumed the script only worked from the top level). Will fix.

grondo:
> Just to be sure I understand, right now tests are not using the actual fluxorama system instance, correct? Although the system instance is running in the container and the sharness tests change paths so that the installed paths are found before the source tree ones, all the flux instances under test are still ephemeral ones started with test_under_flux? If I wanted to write a sharness test, say to make sure that guests can't do certain things in the fluxorama instance, what would that look like?

Good question. First, yes, all the existing tests that use `test_under_flux` should be running single-user test instances. One difference being that systemd is running so that any sharness tests requiring systemd can make use of it.

I've kind of lost my context on how we might run system instance testing. Tests could directly target the executing fluxorama system instance, but one gotcha is that all concurrently executing tests would share the instance, which could result in a deadlock.

Maybe there is a clever way we can start per-test system instances. O/w, we should figure out how to run system instance tests serially -- perhaps there is a single sharness test for all tests which are _required_ to run against the fluxorama system instance directly, i.e. where `test_under_flux` is not sufficient.

grondo:
> It looks like the bionic coverage builder is still enabled - was that intended?

I was kind of hoping Codecov could combine coverage tests (it supposedly does, but doesn't seem to be working the way I expect). I'll just disable the previous coverage builder once I hear that project members are ok with swapping for running in the system instance builder (so that we can run _all_ the tests in one builder)

garlick:
> perhaps there is a single sharness test for all tests which are required to run against the fluxorama system instance directly,

That would be an excellent start IMHO.  Would it make sense to add a sharness test in this PR that just makes sure the system instance can run a job submitted by a guest and is skipped if the enclosing instance is not fluxorama?

Would code paths traveled in the actual running fluxorama instance contribute to coverage?

grondo:
> That would be an excellent start IMHO. Would it make sense to add a sharness test in this PR that just makes sure the system instance can run a job submitted by a guest and is skipped if the enclosing instance is not fluxorama?

That is a good idea. I was thinking of adding a SYSTEM or similar sharness prereq, so maybe I can add that here, then add a small set of tests driven from a master sharness test.

> Would code paths traveled in the actual running fluxorama instance contribute to coverage?

I think the idea was yes, but this currently isn't the case. I'm not sure how to have the code running as user `flux` dump coverage into the source directory of the user running in the container. We may have to see if there is a way to get that to work. Maybe something using a bind mount?

Ah! or maybe we can add this `gcc` option
```
fprofile-dir=path

Set the directory to search for the profile data files in to path. This option affects only the profile data generated by -fprofile-generate, -ftest-coverage, -fprofile-arcs and used by -fprofile-use and -fbranch-probabilities and its related options. Both absolute and relative paths can be used. By default, GCC uses the current directory as path, thus the profile data file appears in the same directory as the object file. In order to prevent the file name clashing, if the object file name is not an absolute path, we mangle the absolute path of the sourcename.gcda file and use it as the file name of a .gcda file.

When an executable is run in a massive parallel environment, it is recommended to save profile to different folders. That can be done with variables in path that are exported during run-time:

 %p 
  process ID.

%q{VAR}
value of environment variable VAR
```

Maybe this would speed up our `make check` as well!

grondo:
I've pushed an experimental `t9000-system.t` sharness test which, similar to `t4000-issues.t`, loads a set of test from a directory and runs them serially.

`t9000-system.t` only runs if a system instance is detected (`security.owner` attribute exists and differs from current userid). If a system instance is detected it _sources_ all tests in `t/system/*` unlike the issues test driver. This is done so that individual tests can call `test_expect_success()` directly to delineate multiple tests and their success/failure in TAP output.

In light of that, all tests _could_ just be inserted directly into `t9000-system.t`, but this allows disjoint tests to be split up into multiple files, and in the future if we discover a way to run multiple system instances per host (it seems like this should be easily doable), then the tests could eventually be run in parallel.

t/t9000-system.t
```diff
@@ -0,0 +1,53 @@
...
+#
+
+test_description='Run tests against a system instance of Flux'
+
+. `dirname $0`/sharness.sh
+
+owner=$(flux getattr security.owner)
+if test -n "$owner" -a "$owner" != "$(id -u)"; then
+    	test_set_prereq SYSTEM
+fi
```
garlick, line 11: Looks great!

It might also be good to check for  `FLUX_TEST_INSTALLED_PATH` before setting the SYSTEM prereq, so we don't end up testing a mixture of flux versions here.

grondo: Good point!

Note that in the `system` CI image, the version of Flux installed in the fluxorama image should be exactly the same as the version in the build directory. However, testing from builddir is not necessarily a good test that everything works properly when installed (and is installed in the right place), so probably best to set up the environment to use installed paths.

grondo: For `t9000-system.t` I think we always want to test the installed path so how about this?
```diff
diff --git a/t/t9000-system.t b/t/t9000-system.t
index 4ee53083f..0b820de04 100755
--- a/t/t9000-system.t
+++ b/t/t9000-system.t
@@ -3,6 +3,7 @@
 
 test_description='Run tests against a system instance of Flux'
 
+FLUX_TEST_INSTALLED_PATH=${FLUX_TEST_INSTALLED_PATH:-/usr/bin}
 . `dirname $0`/sharness.sh
 
 owner=$(flux getattr security.owner)

```

I "tested" that this runs the installed `flux` by running the test after `make clean` completed at the top level:

```
 grondo@fluxorama:/usr/src/t$ ../src/cmd/flux
bash: ../src/cmd/flux: No such file or directory
 grondo@fluxorama:/usr/src/t$ ./t9000-system.t  -d -v
sharness: loading extensions from /usr/src/t/sharness.d/01-setup.sh
sharness: loading extensions from /usr/src/t/sharness.d/flux-sharness.sh
expecting success: 
	jobid=$(flux mini submit id -u) &&
	result=$(flux job attach $jobid) &&
	test_debug "echo Job ran with userid $result" &&
	test $result -eq $(id -u) &&
	test $(flux getattr security.owner) -ne $result

Job ran with userid 1000
ok 1 - 0001-basic.t: system instance runs job as current uid

expecting success: 
	test $(flux jobs -no {userid} $jobid) -eq $(id -u)

ok 2 - 0001-basic.t: flux jobs lists job with correct userid

# passed all 2 test(s)
1..2

```

garlick: I was just thinking about running `make check` on a system that happens to have a different versin flux system instance running.

grondo: Ah, I see, and we wouldn't want to run this test which might not be appropriate for the system installed version.
Hm, I'm not certain how to work around that one. Maybe a separate environment variable that is required for force the test to run, kind of like we have for the valgrind test?

garlick: Could we just skip the test if FLUX_TEST_INSTALLED is unset?

grondo: I could be confused. Is that set by default for the system instance testing?
AFAIK, that is set by the `test-install` builder onlly.

Maybe it should be set for the system instance testing, though we'd have to figure out if coverage works for installed binaries and libraries.

garlick: Yeah I was thinking just set that for the system instance testing.  Sorry, I wasn't thinking about the coverage issue.  Your proposal of an "opt-in" variable is ok with me!

garlick: Yep, agreed!

grondo:
Hm, I'm seeing this in the `system, coverage` builder:

```
./src/common/libccan/ccan/bitmap/.libs/bitmap.gcno:no functions found
./src/common/libccan/ccan/bitmap/.libs/bitmap.gcda:version 'A85R', prefer version 'A93*'
./src/common/libccan/ccan/str/.libs/debug.gcno:version 'A85R', prefer 'A93*'
find: ‘gcov’ terminated by signal 11
```

Perhaps the problem is that the codecov uploader is running outside of the docker container, with a different version of `gcov` than was used to build flux-core inside the container.

I'll see if I can download and run the codecov uploader inside of the docker container.

grondo:
I think I've convinced myself that any coverage changes here are due to the newer compiler reporting more lines than the older compiler. E.g.  notice in the coverage diff that 7745 more lines were detected (not sure how that seems like a lot) in this PR vs master:

In at least one case I noticed that the older compiler doesn't seem to be able to detect `return` lines, so maybe that is part of it.

```@@            Coverage Diff             @@
##           master    #3844      +/-   ##
==========================================
- Coverage   83.50%   80.17%   -3.33%     
==========================================
  Files         372      372              
  Lines       53772    61517    +7745     
==========================================
+ Hits        44902    49321    +4419     
- Misses       8870    12196    +3326     

```

Anyway, this probably ready to go in if I can get a final ACK.

garlick:
Oh OK, that makes sense.

Yes let's get this merged!

grondo:
Thanks!

PR #3846, opened 2021-08-31
broker: improve logging of 0MQ socket events

garlick:
0MQ does things like connect and authentication in its I/O thread without exposing it in its API, which is convenient, though sometimes when not working, you wish you knew what was going on in there. 

0MQ does offer [zmq_socket_monitor(3)](http://api.zeromq.org/master:zmq-socket-monitor) which we've employed in the past for counting peers, but it has a bad bug in the version of libzmq that is still packaged on centos7, and has a few other caveats that you have to be very careful with, such as the ability to deadlock the 0MQ I/O thread if the monitor socket is not properly managed.  So we stopped using it.

This PR brings back socket monitoring for the broker overlay network, but only for logging purposes, and only for modern 0MQ versions.  The sorts of messages that might appear can be seen by running the overlay unit test, e.g.
```
# debug: child sockevent ipc://@trio-0 disconnected
# debug: child sockevent ipc://@trio-0 accepted
# err: child sockevent ipc://@trio-0 handshake failed protocol: ZMTP mechanism mismatch
# err: child sockevent ipc://@trio-0 handshake failed auth: ZAP status code 400
```
This is not super useful on its own, but it might provide useful context around other overlay messages in the logs when things are not working right.

The monitoring API in 0MQ is evolving, so I wrapped it up in a `libzmqutil_monitor` class so we can avoid exposing those details in `broker/overlay.c`.

### DIFF
--- a/src/broker/overlay.c
+++ b/src/broker/overlay.c
@@ -22,6 +22,7 @@
 #include "src/common/libzmqutil/msg_zsock.h"
 #include "src/common/libzmqutil/reactor.h"
 #include "src/common/libzmqutil/zap.h"
+#include "src/common/libzmqutil/monitor.h"
 #include "src/common/libczmqcontainers/czmq_containers.h"
 #include "src/common/libutil/log.h"
 #include "src/common/libutil/kary.h"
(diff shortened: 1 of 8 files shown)


grondo (approved):
LGTM. Same comment applies about hardening the semi-public interface of the `zmq_monitor` class. Completely optional and could be done down the road if there are more use cases.

garlick:
OK, added the param checks and small unit test, and rebased on current master.  I'll go ahead and set MWP.   Thanks!

PR #3847, opened 2021-09-01
libzmqutil: [cleanup] generalize the zeromq authentication protocol server

garlick:
A while back we stopped using the czmq ZAP handler, and implemented our own in `broker/overlay.c` that is integrated with the broker event loop.  The code in `overlay.c` is getting more complex, and this is a chunk that could be abstracted elsewhere.

This PR creates a new class `zmqutil_zap` with some simple interfaces, and then converts the broker to use it.

For the most part, this is just shoveling code from one directory to another, so not much going on here beyond cleanup.

### DIFF
--- a/src/broker/overlay.c
+++ b/src/broker/overlay.c
@@ -21,6 +21,7 @@
 
 #include "src/common/libzmqutil/msg_zsock.h"
 #include "src/common/libzmqutil/reactor.h"
+#include "src/common/libzmqutil/zap.h"
 #include "src/common/libczmqcontainers/czmq_containers.h"
 #include "src/common/libutil/log.h"
 #include "src/common/libutil/kary.h"
@@ -38,7 +39,6 @@
 #endif
 
 #define FLUX_ZAP_DOMAIN "flux"
-#define ZAP_ENDPOINT "inproc://zeromq.zap.01"
 
 #define DEFAULT_FANOUT 2
 
@@ -114,9 +114,7 @@ static const double idle_max = 30.0;
 
 struct overlay {
     zcert_t *cert;
-    zcertstore_t *certstore;
-    zsock_t *zap;
-    flux_watcher_t *zap_w;
+    struct zmqutil_zap *zap;
     int enable_ipv6;
 
     flux_t *h;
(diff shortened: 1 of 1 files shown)


src/common/libzmqutil/zap.c
```diff
@@ -0,0 +1,222 @@
...
+
+/* Create a zcert_t and add it to in-memory zcertstore_t.
+ */
+int zmqutil_zap_authorize (struct zmqutil_zap *zap,
+                           const char *name,
+                           const char *pubkey)
+{
+    uint8_t public_key[32];
+    zcert_t *cert;
+
```
grondo, line 159: Since this code has been moved "external" to the broker, should all public functions here check for valid arguments? (.e. return EINVAL on `!zap || !name || !pubkey`)

I realize that will add extra development and testing time, so I'm fine with leaving that out for now, but thought I'd ask.

grondo (approved):
LGTM! Just one question inline.

src/common/libzmqutil/zap.c
garlick: I thought not.  Maybe we could add that and a local unit test if a second user emerges?  For now I wanted to keep moving forward.

garlick:
If this is OK with you @grondo I'll set MWP.

src/common/libzmqutil/zap.c
grondo: Should this file have just been kept under `src/broker` if there is no other user current or planned?

I guess it is moot at this point if we're just moving forward.

grondo:
Sounds fine with me.

src/common/libzmqutil/zap.c
garlick: It does seem like a useful standalone thing if we ever want to have a process other than the broker implement a zeromq-based server that uses 0MQ security, so I'd prefer we keep it here.  I think you've convinced me that I should take a small amount of time and make the interfaces more robust and add a unit test, so I'll go ahead and do that.

garlick:
I went ahead and made the suggested parameter checks and added a small unit test.  The unit test could do more, but I think that can wait.  I'll set MWP per earlier comment.

Thanks!

PR #3849, opened 2021-09-03
update flux-security version to v0.5.0 in docker images

grondo:
This PR updates the default flux-security version in docker images created by `docker-run-checks.sh` to the latest: v0.5.0. This means all published `fluxrm/flux-core` images will include the newer flux-security.


garlick (approved):
Good idea.

grondo:
Thanks! I've set MWP.

Issue #3852, opened 2021-09-08
rpc: fulfilled RPCs don't release reactor active watcher count


grondo:
A non-streaming Flux RPC only expects one response, but currently the msg_handler is left "active" even after that response is received. This makes it appear as if the reactor has active watchers even if no more message or other activity is expected. This is ordinarily not a problem if the future associated with the RPC is destroyed, since this destroys the associated msg_handler(s), allowing the reactor to exit once all RPCs have been processed.

If, however, the contents of the RPC or future need to be used *outside* the reactor, e.g. when integrating with a foreign eventloop in Python or Julia, then the future cannot be destroyed, the reactor will never report a zero active watcher count or exit.

One fix would be to stop the msg_handler associated with an RPC once no more responses are expected. For a non-streaming RPC this is obviously after the first response. For a streaming RPC, the msg_handler can presumably be stopped after the first error is received.

The following test program demonstrates the issue:
```C
#include <unistd.h>
#include <stdio.h>
#include <errno.h>
#include <string.h>

#include <flux/core.h>

static int ready = 0;

static void cb (flux_future_t *f, void *arg)
{
    ready = 1;
}

int main (int ac, char **av)
{
    int rc;
    int loopcount = 0;
    flux_future_t *f;
    flux_kvs_txn_t *txn;

    flux_t *h = flux_open (NULL, 0);
    if (h == NULL) {
        fprintf (stderr, "flux_open: %s\n", strerror (errno));
        exit (1);
    }
    if (!(txn = flux_kvs_txn_create ())) {
        fprintf (stderr, "flux_kvs_txn_create: %s\n", strerror (errno));
        exit (1);
    }
    if (flux_kvs_txn_put (txn, 0, "foo", "test") < 0) {
        fprintf (stderr, "flux_kvs_txn_put: %s\n", strerror (errno));
        exit (1);
    }
    if (!(f = flux_kvs_commit (h, NULL, 0, txn))) {
        fprintf (stderr, "flux_kvs_commit: %s\n", strerror (errno));
        exit (1);
    }
    flux_kvs_txn_destroy (txn);

    if (flux_future_then (f, -1., cb, NULL) < 0) {
        fprintf (stderr, "flux_future_then: %s\n", strerror (errno));
        exit (1);
    }
    while ((rc = flux_reactor_run (flux_get_reactor (h),
                                   FLUX_REACTOR_NOWAIT))) {
        if (ready && loopcount++ > 1) {
            fprintf (stderr,
[...]

PR #3853, opened 2021-09-08
libflux: deactivate RPC message handlers after final response

grondo:
This PR fixes the issue described in #3852, a problem first described in flux-framework/FluxRM.jl#17.

In general, the problem is that RPCs do not drop their active references on an event loop until the future is _destroyed_, instead of dropping them when future is _fulfilled_. This means an RPC future can't be used outside of an event loop since `flux_reactor_run(3)` will not return until all RPC futures are destroyed.

This PR just stops the internal RPC msg_handler once the final expected response is received. In the case of a normal RPC, this is when the future is first fulfilled, but for streaming RPCs this is at the first error.

A couple spots were missing the FLUX_RPC_STREAMING flag (as evidenced by hangs). These were fixed here.

The libsubprocess protocol still doesn't send an error response as the terminating message in the protocol. However, I think this is OK for now since, internally, the multi-response future is destroyed whenever the "final" response is received.

### DIFF
--- a/src/common/libflux/rpc.c
+++ b/src/common/libflux/rpc.c
@@ -179,6 +179,7 @@ static void response_cb (flux_t *h, flux_msg_handler_t *mh,
                          const flux_msg_t *msg, void *arg)
 {
     flux_future_t *f = arg;
+    struct flux_rpc *rpc = flux_future_aux_get (f, "flux::rpc");
     flux_msg_t *cpy;
     int saved_errno;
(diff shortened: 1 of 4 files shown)


garlick (approved):
Thanks for fixing this oversight!  Looks great, and also thanks for taking the time to enhance the unit test!

grondo:
Thanks! I set MWP.

Issue #3855, opened 2021-09-09
Spectrum MPI personality does not support 10.4


SteVwonder:
In Spectrum 10.4, we need to use the `ucx` pml plugin as opposed to `yalla` (which is used in 10.3).  So we need A) someway for the user to optionally specify which version of Spectrum they are running and B) someway to key off of that within the personality to change the behavior.

On the ☕ call today, @grondo suggested extending the `-o mpi=` syntax to support something like `-o mpi=spectrum@10.4` or `-o mpi=spectrum=10.4`.  The `spectrum` personality just needs to split based on the chosen delimiter (if it exists) and modify its behavior based on that version.

dongahn:
I like the version augmented personality idea.

grondo:
We could change the current test in MPI plugins to use `if not shell.options.mpi:find("^spectrum")` (I think this is the equivalent of `startswith` in Lua, e.g.
```diff
diff --git a/src/shell/lua.d/spectrum.lua b/src/shell/lua.d/spectrum.lua
index 78b21b524..379cb162c 100644
--- a/src/shell/lua.d/spectrum.lua
+++ b/src/shell/lua.d/spectrum.lua
@@ -10,7 +10,7 @@
 
 -- Set environment specific to spectrum_mpi (derived from openmpi)
 --
-if shell.options.mpi ~= "spectrum" then return end
+if not shell.options.mpi:find("^spectrum") then return end
 
 local posix = require 'posix'
 
```

Then to test for a specific version later, you can use an exact match, i.e. `if shell.options.mpi == "spectrum@10.4"`

At some point we may want a way to add some convenience functions to these Lua plugins, e.g. `string:startswith` and perhaps a way to get the version number and do `<=` or `>=` comparison, etc.

grondo:
Hm, one thing we could easily do to collect a "library" of functions for the shell lua plugins is to make a `00-lib.lua` plugin. E.g. I pulled some of the reusable code out of `spectrum.lua` and added some functions that could be useful:

```lua

function string.startswith (str, start)
    if not str or str == "" then
        return false
    end
    return string.sub(str,1,string.len(start)) == start
end

function shell.get_mpi_option ()
    local mpi = shell.options.mpi
    if mpi then
        return mpi:match("^[^@]+"), mpi:match("@(.+)$")
    end
    return nil
end

function prepend_path (env_var, path)
    local val = shell.getenv (env_var)

[...]

PR #3856, opened 2021-09-10
shell: add library of Lua convenience functions for use in plugins 

grondo:
As described in #3855, it would be helpful if there was a way to abstract commonly useful functionality for the MPI personality plugins in the shell (and other Lua-plugins). This PR creates and installs a `00-lib.lua` "plugin", and collects some new and existing common functions that will be available in all other shell Lua plugins (assuming that `00-lib.lua` is sourced before other lua.d files)

A new `shell.getopt_with_version (name)` function is introduced, which automatically splits a top-level shell option `name` into a `value` and `version` parts, separated by `@` (assuming that is the separator of choice).
E.g. if `-o mpi=spectrum@10.4` then
```lua
    local mpi, version = shell.getopt_with_version ("mpi")
```
would return `"spectrum", "10.4"`. The `spectrum.lua` plugin could then take version-dependent action by checking if `version` is set (not `nil`). 

Some of the other functions in `spectrum.lua` were preemptively moved to `00-lib.lua`, even though there are not yet other users.

### DIFF
--- a/src/shell/Makefile.am
+++ b/src/shell/Makefile.am
@@ -22,6 +22,7 @@ dist_shellrc_SCRIPTS = \
 	initrc.lua
 
 dist_shelllua_SCRIPTS = \
+	lua.d/00-lib.lua \
 	lua.d/mvapich.lua \
 	lua.d/intel_mpi.lua \
 	lua.d/openmpi.lua \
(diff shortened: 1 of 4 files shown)


SteVwonder (approved):
Very nice refactor and additions! LGTM! Thanks @grondo!

grondo:
Thanks! I'll set MWP.

PR #3858, opened 2021-09-11
WIP: add pmix shell plugin

garlick:
OK, this is an experiment, and more WIP than most WIPs.  It adds a pmix plugin to the shell so that possibly we'll be better at launching openmpi someday (like in version 5 when they drop all other ways to launch it).

To replicate my environment, install openpmix 3.2.3 somewhere like `/usr/local` and configure flux with `--enable-pmix-bootstrap`.  Note that `libpmix2-3.1.5-1` packaged wtih Ubuntu 20.04 doesn't work.  I don't know whose fault it is, mine or theirs.  Avoid that version for now.  A `src/shell/plugins/pmix.so` module will be built that the shell currently loads automatically.   Automatic loading is not the desired endpoint with this for a variety of reasons, not least that it brings in libzmq and its dependencies, undoing all the work that @chu11 did to avoid the FIPS self-test penalty on TOSS 4.

Because our `openmpi.lua` plugin subverts openmpi's use of pmix, there is a temporary commit here that just comments that stuff out.  We'll need a long term solution for that.

The main innovation in this attempt is a mechanism to synchronize pmix callbacks/upcalls, which are made in the context of the pmix server thread, with the reactor in the shell thread. The real callbacks/upcalls are wrapped so that they create a message and send it over a shared memory zeromq socket.  The message is received by a reactor watcher for the other end of the socket, and the callback is reconstructed with the original signature.  This shenanigans is all neatly contained in `pmix/server.c` so for the most part, those functions still work as described in the pmix spec.

There are two pmix upcalls implemented, which I assumed were fundamental.  One is the "direct modex" callback in which we let the pmix server thread on one shell asks the pmix server thread on another shell for a particular rank's posted KVS keys.  This is implemented as a shell RPC. I'm not sure I totally got it right as I'm ignoring the `pmix.req.key=value` option that the server passes telling us to synchronize with the remote server to wait for specific KVS keys to be posted before asking for the keys in bulk.  But I don't know how that's supposed to work so for now, I just fetch the keys and hope they are there.

The other callback of interest is "fence_nb" which is the combo barrier/exchange operation as in  "normal" pmi.  I just mapped it to the tree-based exchange code in the pmi builtin plugin, which is not a great match in terms of its data handling, but gets the job done with a little extra base64 encoding.

I am able to run  an openmpi hello world across two shells with this.

Much more to do for this to be reasonable, but wanted to park it here in case it stimulated some discussion.

[...]

PR #3859, opened 2021-09-12
shell: fixes and enhancements for plugin loading

grondo:
This PR is a set of changes and fixes to the shell plugin loader found while trying to allow plugin loading from subdirectories for #3858.

This includes:

 - link any in-tree plugins from `.libs` to the parent so that `.libs` doesn't need to be used in `INTREE_SHELL_PLUGINPATH`. This allows plugins intree to be loaded from subdirectories to support "optional" plugins.
 - It is kind of painful to remember to pass a table with `file` member when loading a plugin with no configuration. Update `plugin.load` to optionally take a string as its only argument and consider that the pattern to use for plugin loading. That is `plugin.load ("*.so")` now works.
 - Fix `plugin.load` so it raises an error on invalid argument


garlick (approved):
Nice, thank you!

Tested by rebasing  #3858 on top of this and adding `plugin.load ("pmix/pmix.so")` to `openmpi.lua`.  Works!  Also does indeed throw an error when the plugin isn't there.

grondo:
Thanks! I've set MWP.

PR #3860, opened 2021-09-14
libflux/future: fix comment typo

chu11:
`<=` should be `<`, otherwise it'd be the `==` case listed below.


grondo (approved):
Good catch.

PR #3861, opened 2021-09-14
allow jobtap plugins to subscribe to job events

grondo:
This work was done in anticipation of future requirements for some jobtap plugins to get callbacks on job events and not just job state transitions (and priority/urgency events). This could probably be implemented by separately watching the job eventlog, but this has high development cost and is wasteful, since all events go through the job manager anyway.

Issuing a callback for all loaded plugins on every job event would probably be high cost, therefore this approach requires plugins that want these extra  callbacks to "subscribe" to events for the job. This means that in the common case, this change should add no overhead, and only plugins which are interested in all job events will have `job.event.*` callbacks called.

Unfortunately the resulting code was a little more complicated than I had hoped. Since a plugin subscribed to jobs could be unloaded before the jobs are inactive, and jobs can go inactive before a plugin is unloaded, each plugin has to keep a list of its subscriptions in addition to the job's list of subscribed plugins. Then on destruction, each entity has to be removed from the alternate list.

Finally, to make good use of the new functionality, the dependency-after plugin uses a job subscription for `after` dependencies to release jobs after the `start` event, instead of after the transition to the `RUN` state (technically after the `alloc` event)

(Thinking of release notes, perhaps this last change should be moved to a separate PR)

This is still WIP because a more comprehensive test plugin should be added that exercises the new functionality.

### DIFF
--- a/doc/man7/flux-jobtap-plugins.rst
+++ b/doc/man7/flux-jobtap-plugins.rst
@@ -161,6 +161,14 @@ job.state.*
   eventlog, but before any action has been taken on that state (since the
   action could involve immediately transitioning to a new state)
 
+job.event.*
+  The ``job.event.*`` callbacks are only made for plugins that have explicitly
+  subscribed to a job with ``flux_jobtap_job_subscribe()``. In this case,
+  all job events result in this callback being invoked on all subscribed
+  plugins. This may be useful for plugins to get notification of events
+  that do not necessarily result in a state transition, e.g. the ``start``
+  event or a non-fatal ``exception``.
+
 job.state.depend
   The callback for ``FLUX_JOB_STATE_DEPEND`` is the final place from which
   a plugin may add dependencies to a job. Dependencies are added via
--- a/src/modules/job-manager/event.c
+++ b/src/modules/job-manager/event.c
@@ -650,6 +650,20 @@ static int event_jobtap_call (struct event *event,
                               json_t *entry,
                               flux_job_state_t old_state)
 {
+
+    /*  Notify any subscribers of all events, separately from
+     *   special cases for state change and urgency events below.
+     */
+    if (jobtap_notify_subscribers (event->ctx->jobtap,
+                                   job,
+                                   name,
+                                   "{s:O}",
+                                   "entry", entry) < 0)
+            flux_log (event->ctx->h, LOG_ERR,
+                      "jobtap: event.%s callback failed for job %ju",
+                      name,
+                      (uintmax_t) job->id);
+
     if (job->state != old_state) {
         /*
          *  Call plugin callback on state change
--- a/src/modules/job-manager/jobtap.c
+++ b/src/modules/job-manager/jobtap.c
@@ -875,6 +875,45 @@ out:
     return rc;
 }
 
+int jobtap_notify_subscribers (struct jobtap *jobtap,
+                               struct job *job,
+                               const char *name,
+                               const char *fmt,
+                               ...)
+{
+    flux_plugin_arg_t *args;
+    char topic [64];
+    int topiclen = 64;
+    va_list ap;
+    int rc;
+
+    if (!job->subscribers)
+        return 0;
+
+    if (snprintf (topic, topiclen, "job.event.%s", name) >= topiclen) {
+        flux_log (jobtap->ctx->h, LOG_ERR,
+                  "jobtap: %s: %ju: event topic name too long",
+                  name,
+                  (uintmax_t) job->id);
+        return -1;
+    }
+
+    va_start (ap, fmt);
+    args = jobtap_args_vcreate (jobtap, job, fmt, ap);
+    va_end (ap);
+    if (!args) {
+        flux_log (jobtap->ctx->h, LOG_ERR,
+                  "jobtap: %s: %ju: failed to create plugin args",
+                  topic,
+                  (uintmax_t) job->id);
+        return -1;
+    }
+
+    rc = jobtap_stack_call (jobtap, job->subscribers, job, topic, args);
+    flux_plugin_arg_destroy (args);
+    return rc;
+}
+
 int jobtap_call (struct jobtap *jobtap,
                  struct job *job,
                  const char *topic,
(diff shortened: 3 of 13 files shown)


grondo:
I realize now the `after:JOBID` dependency changes here have a couple issues:
 
 * It violates the letter of [RFC 26](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_26.html#after)
 * There is race on job submission if the antecedent job is in `RUN` state but hasn't emitted the `start` event

I will pull the `dependency-after` changes here.

Releasing jobs after the `start` event instead of the `alloc` event _feels_ like the right thing to do. In particular,  this will allow better synchronization between non-traditional jobs (e.g. service jobs or KubeFlux jobs) which may use alloc-bypass or may want to synchronize with actual application startup, where an external tool will presumably post a `start`  event when the service/container has _started_ not just been allocated resources.

So maybe I'll propose a PR update to RFC 26 for that change, and future PR that updates the semantics of `after:JOBID`.

grondo:
Ok, the changes to `after:JOBID` dependencies has been removed and a test plugin added to the testsuite.

The coverage is a little low due to various error conditions that can't be forced under CI.

grondo:
Hm, `t3304-system-rpctrack-down.t` failed in a builder with:
```
  expecting success: 
  	count=$(flux module stats --parse=child-rpc overlay) &&
  	echo $count &&
  	test $count -gt 0
  
  0
  not ok 10 - broker 0 tracked child rpc count is nonzero
```

I'll restart the CI

garlick:
Glad you mentioned that test failure @grondo - I was thinking that sitaution was not possible, so I went ahead and opened an issue in case there's a mistaken assumption in the test.

src/modules/job-manager/event.c
```diff
@@ -650,6 +650,20 @@ static int event_jobtap_call (struct event *event,
                               json_t *entry,
                               flux_job_state_t old_state)
 {
+
```
garlick, line 653: Commit message: s/whie/which/

src/modules/job-manager/event.c
```diff
@@ -650,6 +650,20 @@ static int event_jobtap_call (struct event *event,
...
+     */
+    if (jobtap_notify_subscribers (event->ctx->jobtap,
+                                   job,
+                                   name,
+                                   "{s:O}",
+                                   "entry", entry) < 0)
+            flux_log_error (event->ctx->h,
+                            "jobtap: event.%s callback failed for job %ju",
+                            name,
+                            (uintmax_t) job->id);
```
garlick, line 665: assumes errno is set by `jobtap_notify_subscribers()` but it appears that is not the case.

src/modules/job-manager/jobtap.c
```diff
@@ -875,6 +875,44 @@ int jobtap_check_dependencies (struct jobtap *jobtap,
...
+
+    va_start (ap, fmt);
+    if (!(args = jobtap_args_vcreate (jobtap, job, fmt, ap))) {
+        flux_log (jobtap->ctx->h, LOG_ERR,
+                  "jobtap: %s: %ju: failed to create plugin args",
+                  topic,
+                  (uintmax_t) job->id);
+        return -1;
+    }
+    va_end (ap);
```
garlick, line 909: va_start / va_end not balanced in error case

doc/man7/flux-jobtap-plugins.rst
```diff
@@ -161,6 +161,14 @@ job.state.*
   eventlog, but before any action has been taken on that state (since the
   action could involve immediately transitioning to a new state)
 
+job.event.*
```
garlick, line 164: commit message s/int/in/

garlick (approved):
LGTM!  just a couple of minor things inline.

Is the case of job manager restart with plugins that subscribe to job events handled satisfactorily here?  I'm not sure the plugins would get past job events replayed.  Is that an issue for recreating plugin state?

I'll go ahead and approve.  If there is any work to do for the restart case then we could always open a bug and return to it later.

grondo:
>
>
> Is the case of job manager restart with plugins that subscribe to job
> events handled satisfactorily here? I'm not sure the plugins would get past
> job events replayed. Is that an issue for recreating plugin state?
>
No, events are not replayed, but then neither are job state callbacks. When
a job is reloaded, plugins will get a callback for the job's current state.
If some state needs to be acquired, it could be done at that point. PR
#3683 was meant to help here, but unfortunately that one sounds like a non
starter.

It will be difficult to support event tracing without a way to check for
any events that already occurred before the trace started, so I'm wondering
if both these PRs should be dropped.

garlick:
It wasn't my intent to put up road blocks to expanding the in-memory footprint if we need it!

In fact, weren't we considering caching the entire eventlog in memory for active jobs at one point?

garlick:
You commented in the other PR about the per-job zlistx here, but isn't it created/destroyed on demand?   So just the overhead of a pointer when not in use, which sounds fine to me.   I suppose you could move it to the aux list and save even that :-)

grondo:
> You commented in the other PR about the per-job zlistx here, but isn't it created/destroyed on demand? So just the overhead of a pointer when not in use, which sounds fine to me. I suppose you could move it to the aux list and save even that :-)

Oh yeah, good point. I had it on the aux list at first but I wanted to make the likely case (no subscribers) fast, and checking a pointer for NULL is probably faster than an aux_item lookup, though I'm fine moving it back there if you'd prefer.

garlick:
> Oh yeah, good point. I had it on the aux list at first but I wanted to make the likely case (no subscribers) fast, and checking a pointer for NULL is probably faster than an aux_item lookup, though I'm fine moving it back there if you'd prefer.

No that makes sense - this is fine.

grondo:
rebased so that the follow on PRs can also be rebased.

grondo:
rebased again. This time added a cleanup commit to remove `envz.h` and `argz.h` from `restart.c` since they were unused (unrelated).

Otherwise, nothing has changed.

grondo:
Thanks! Setting MWP now that the 2 follow-on PRs have been approved. I had to restart CI due to some random ffi errors in one of the builders (if it reproduces I'll open an issue)

PR #3863, opened 2021-09-15
jobtap: allow jobtap plugins to query posted events for jobs

grondo:
This PR adds a new interface `flux_jobtap_job_event_count()` to jobtap.

This allows plugins to determine which and how many of a named event have been posted to a job. This allows, for instance, more fine-grained determination of a job's current "state" (beyond `flux_job_state_t`), e.g. has a job in `RUN` state seen a `start` event or `release` event or `exception`. This is also provides a much easier method by which a plugin can determine a job's place in the state diagram that checking the current state (if the job has seen a `start` event, then we know it is in the RUN state or later, instead of having to check whether current state is `CLEANUP` or `INACTIVE` or `RUN`)

Unfortunately, this requires a hash added to each job struct. However, in real-world testing this didn't seem to have much impact to job throughput, though there will be a runtime memory usage penalty. Since effort has been made to keep the size of the `struct job` small, I wonder if this approach will be acceptable.

This interface will be used by future changes to the `after:JOBID` dependencies to trigger the release of the dependent job on the `start` event instead of transition to the `RUN` state.

This PR is built on top of #3861 (mainly to avoid conflicts)

### DIFF
--- a/src/modules/job-manager/job.c
+++ b/src/modules/job-manager/job.c
@@ -25,6 +25,8 @@
 #include "job.h"
 #include "event.h"
 
+#define EVENTS_BITMAP_SIZE 64
+
 static void subscribers_destroy (struct job *job);
 
 void job_decref (struct job *job)
@@ -37,6 +39,7 @@ void job_decref (struct job *job)
         json_decref (job->annotations);
         grudgeset_destroy (job->dependencies);
         subscribers_destroy (job);
+        free (job->events);
         aux_destroy (&job->aux);
         free (job);
         errno = saved_errno;
@@ -57,6 +60,10 @@ struct job *job_create (void)
 
     if (!(job = calloc (1, sizeof (*job))))
         return NULL;
+    if (!(job->events = bitmap_alloc0 (EVENTS_BITMAP_SIZE))) {
+        free (job);
+        return NULL;
+    }
     job->refcount = 1;
     job->userid = FLUX_USERID_UNKNOWN;
     job->urgency = FLUX_JOB_URGENCY_DEFAULT;
@@ -360,6 +367,31 @@ int job_events_subscribe (struct job *job, flux_plugin_t *p)
         return -1;
     }
 
+    return 0;
+}
+
+int job_event_id_set (struct job *job, int id)
+{
+    if (id < 0) {
+        errno = EINVAL;
+        return -1;
+    }
+    if (id >= EVENTS_BITMAP_SIZE) {
+        errno = ENOSPC;
+        return -1;
+    }
+    bitmap_set_bit (job->events, id);
+    return 0;
+}
+
+int job_event_id_test (struct job *job, int id)
+{
+    if (id < 0 || id >= EVENTS_BITMAP_SIZE) {
+        errno = EINVAL;
+        return -1;
+    }
+    if (bitmap_test_bit (job->events, id))
+        return 1;
     return 0;
 }
 
--- a/configure.ac
+++ b/configure.ac
@@ -165,7 +165,10 @@ AC_CHECK_HEADERS( \
 ##
 # Checks for typedefs, structures, and compiler characteristics
 ##
-AC_C_BIGENDIAN
+AC_C_BIGENDIAN(
+  [AC_DEFINE([HAVE_BIG_ENDIAN], [1], [big endian])],
+  [AC_DEFINE([HAVE_LITTLE_ENDIAN], [1], [little endian])]
+)
 AC_C_CONST
 AC_TYPE_SIZE_T
 AX_COMPILE_CHECK_SIZEOF(int)
--- a/src/common/libccan/Makefile.am
+++ b/src/common/libccan/Makefile.am
@@ -23,5 +23,7 @@ libccan_la_SOURCES = \
 	ccan/bitmap/bitmap.h \
 	ccan/bitmap/bitmap.c \
 	ccan/endian/endian.h \
+	ccan/compiler/compiler.h \
+	ccan/ptrint/ptrint.h \
 	ccan/build_assert/build_assert.h \
 	ccan/container_of/container_of.h
--- a/src/common/libccan/Makefile.am
+++ b/src/common/libccan/Makefile.am
@@ -20,5 +20,8 @@ libccan_la_SOURCES = \
 	ccan/list/list.h \
 	ccan/base64/base64.c \
 	ccan/base64/base64.h \
+	ccan/bitmap/bitmap.h \
+	ccan/bitmap/bitmap.c \
+	ccan/endian/endian.h \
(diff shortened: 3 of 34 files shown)


garlick:
> Unfortunately, this requires a hash added to each job struct. However, in real-world testing this didn't seem to have much impact to job throughput, though there will be a runtime memory usage penalty. Since effort has been made to keep the size of the struct job small, I wonder if this approach will be acceptable.

A zhashx per job is a bit heavy.  Is it necessary to have a count of events as opposed to a flag, and do arbitrary events have to be countable or could we just count the RFC-defined ones that drive the state machine?

grondo:
I don't think counting events here is necessary. However, limiting
supported events to a finite set would limit long term usefulness. E.g.
plugins could not synchronize on any custom event names.

I can go back to the drawing board on this one. Maybe for now we track only
the start event for jobs, which is the main issue being solved here (can't
tell if a job in the RUN state has started or not) Maybe the finish event
too?

garlick:
> Maybe for now we track only the start event for jobs, which is the main issue being solved here (can't tell if a job in the RUN state has started or not) Maybe the finish event too?

So just expand that  `struct job` bitfield to include some more bits?  That makes sense to me if it works for the use case at hand.

grondo:
> A zhashx per job is a bit heavy. 

Here's another idea: what we really want is fast insertions, but lookups do not have to be that fast.
Could we use a list of events or `struct aux_item` instead to support the full set of events?

grondo:
We should probably consider closing #3861 as well, since it adds a zlistx to every job in order to track which plugins are tracing events for which jobs. Without the ability to query whether individual events have already occured, per-event callbacks may have limited actual usefulness.

grondo:
> Could we use a list of events or struct aux_item instead to support the full set of events?

Or a `char **argz, size_t argz_len` might require even less memory, and `argz_add` is probably pretty fast.
For multiple duplicate events the argz vector might grow large compared to keeping a count of each event, however this would be uncommon, and and interesting side effect is you keep the event ordering... I can't think of a smaller data structure to store a set of strings.

garlick:
Ooh yeah, the argz is a neat idea.

grondo:
Just for fun, I ran the following test program under massif to get some idea of heap utilization for these two disparate approaches:

```c
#include <czmq.h>
#include <stdio.h>
#include <argz.h>

int main (void)
{
    zhashx_t *h = zhashx_new ();
    char *argz = NULL;
    size_t argz_len = 0;

    zhashx_insert (h, "submit", (void *) 0x1);
    zhashx_insert (h, "dependency-add", (void *) 0x1);
    zhashx_insert (h, "dependency-remove", (void *) 0x1);
    zhashx_insert (h, "depend", (void *) 0x1);
    zhashx_insert (h, "priority", (void *) 0x1);
    zhashx_insert (h, "alloc", (void *) 0x1);
    zhashx_insert (h, "start", (void *) 0x1);
    zhashx_insert (h, "finish", (void *) 0x1);
    zhashx_insert (h, "release", (void *) 0x1);
    zhashx_insert (h, "free", (void *) 0x1);
    zhashx_insert (h, "clean", (void *) 0x1);


    argz_add (&argz, &argz_len, "submit");
    argz_add (&argz, &argz_len, "dependency-add");
    argz_add (&argz, &argz_len, "dependency-remove");
    argz_add (&argz, &argz_len, "depend");
    argz_add (&argz, &argz_len, "priority");
    argz_add (&argz, &argz_len, "alloc");
    argz_add (&argz, &argz_len, "start");
    argz_add (&argz, &argz_len, "finish");
    argz_add (&argz, &argz_len, "release");
    argz_add (&argz, &argz_len, "free");
    argz_add (&argz, &argz_len, "clean");
    sleep (1);
    zhashx_destroy (&h);
    free (argz);
}
```

If I'm reading the results correctly:
```
zhashx_new: 128B
zhashx_insert: 440B
??? (unknown czmq function): 136B
strdup (I assume for zhashx key duplication): 94B
---
Total: 798B for zhashx


argz_add: 94B
```

So quite a savings there, even if we store duplicate events (e.g. `urgency` `priority`) multiple times, argz would still be a pretty memory efficient way to cache events posted to jobs.

grondo:
Maybe I'll redo this PR with the argz approach and see how it goes.

This will clean things up a little since the ccan additions can be dropped.

grondo:
I've taken a different approach here to see if the result is more acceptable. In this version of the PR, I tried to reduce the amount of memory added to each job, while still allowing plugins to query for arbitrary event names (assuming in the future plugins may emit custom events to the eventlog for consumption, e.g. in the case of the CRAY pals port distribution plugin)

The first change is that counts of events are no longer provided to plugins. If a plugin needs to determine the number of any event, or if e.g. `*-add` and `*-remove` events are unbalanced, then it will need to fetch the eventlog.

Since event names are mostly shared between all jobs, it seemed like a shame to keep copies of event strings in memory for each job. Therefore, this PR now introduces a global events hash which creates a unique integer identifier for each new event name. In order to track the set of events posted to each job, we now only have to store a set of integers, which is a much more space efficient problem.

Since there is a disturbing lack of bitmaps in Flux, this PR pulls in a resizeable bitmap implementation from CCAN and assigns a 64b bitmap to each job for tracking events. This allows for 64 unique global event names across a running Flux instance, which seems like plenty for now. This implementation therefore only adds 8 bytes (plus a pointer?) to each job structure, and should have pretty fast additions and lookups (especially once the event id hash is "primed" by the first job)

The bitmap structure is resizeable, so if the fixed size becomes a problem, that can be handled at runtime in the future. It just didn't seem necessary at this point. (addition of a event with an id >= 64 will now fail)

garlick:
Finally we have the bitmap technology!

src/modules/job-manager/job.c
```diff
@@ -62,6 +65,7 @@ struct job *job_create (void)
     job->urgency = FLUX_JOB_URGENCY_DEFAULT;
     job->priority = -1;
     job->state = FLUX_JOB_STATE_NEW;
+    job->events = bitmap_alloc0 (EVENTS_BITMAP_SIZE);
```
garlick, line 68: check for malloc failure?

garlick (approved):
Nice job on the scalability here!  I just noticed one small nit.  Feel free to set MWP when fixed.

src/modules/job-manager/job.c
grondo: Thanks, not sure how I forgot that.

grondo:
Thanks, I've fixed up that missing check for malloc failure and force-pushed.

I will wait until #3865 is approved to set MWP since it will be good to get these 3 chained PRs merged in sequence.

grondo:
rebased after #3861. Setting MWP.

Issue #3815, opened 2021-08-05
job-exec: replace fork/waitpid with a mechanism which allows "rediscovery" of broker subprocesses


grondo:
## Problem

Currently, job shells are launched in bulk via libsubprocess from the job-exec module on rank 0 of the system instance. If a broker or subtree needs to be restarted, the flux-shell processes will be signaled (or an attempt to signal them will be made, this may not work for multi-user jobs), and on restart all knowledge of the subprocesses will be lost. Additionally, the job-exec module will receive no further RPC responses in the libsubprocess protocol and will assume the shell processes are still running, or at best will receive an error response and assume the shells are dead and will terminate the job.

As part of [L2 Resiliency](https://github.com/flux-framework/flux-core/wiki/System-Instance-L2-Resiliency) a broker that is restarted will need the ability to leave these launched flux-shell processes running after shutdown, as well as a method to discover any actively running job shells on restart.

A restarted broker will additionally need to be able to do the equivalent of `waitpid()` on processes for which it is no longer the parent.

This is one step in allowing jobs to survive a broker restart.

## Experiment: run job shells as transient service units in a user systemd instance (session manager)

A potential solution, specific to systems running systemd, is to leverage systemd to manage containment and execution of job shells. Thus, the systemd instance will become the parent of all job shells. The flux would use the DBus API to request launch of jobs, and would also use DBus to get asynchronously notified of status changes in the transient unit (e.g. PID, exit status, etc). The broker can also set the `RemainAfterExit` property on the unit, so that information about a job shell that exits is not lost if it occurs while a broker is disconnected or restarting.

The first requirement here is that a user systemd instance for the user `flux` is started when the `flux.service` is started. This is not as easy as it seems, ideally we'd be able to add:

```
Wants: user@flux.service
```

Unfortunately, the correct service name for user systemd instance is `user@UID.service`, and there does not seem to be a way to look up a UID in a unit file, so we have to settle for starting systemd via `ExecStartPre`:
```
# Ensure the user systemd instance for user 'flux' is started.
#  The system instance will use this to run job shells as
#  transient units.
ExecStartPre=bash -c 'systemctl start user@$(id -u flux).service'
```

We also need to ensure that the Flux instance will have access to the DBus connection. Again, since the `flux` userid can't be determined in the unit file, we have to set the requisite environment variables on the `ExecStart` command line:

```
ExecStart=bash -c '\
  XDG_RUNTIME_DIR=/run/user/$UID \
[...]

PR #3864, opened 2021-09-16
Support new libsdprocess to launch processes under systemd

chu11:
per some discussion in https://github.com/flux-framework/flux-core/issues/3815 I'd thought I'd post my initial implementation of `libsdprocess` to see if there are any initial comments.  Probably the header file `sdprocess.h` is the most important thing to skim over.

### DIFF
--- a/src/common/libsdprocess/sdprocess.c
+++ b/src/common/libsdprocess/sdprocess.c
@@ -1,0 +1,1785 @@
+/************************************************************\
+ * Copyright 2021 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* libsdprocess
+ *
+ * The libsdprocess library is a library to launch and monitor
+ * processes under systemd.  It abstracts away the underlying sd-bus
+ * API calls into a higher level API.  Monitoring of a process is
+ * integrated into the flux reactor by using a fd watcher and the file
+ * descriptor provided by sd_bus_get_fd().
+ *
+ * A good portion of the code here is modeled after the code found in
+ * the tools `systemd-run` and `systemctl`.
+ *
+ * Additional information can be found in the following documentation:
+ *
+ * DBUS spec: https://dbus.freedesktop.org/doc/dbus-specification.html
+ *
+ * org.freedesktop.systemd1:
+ * https://www.freedesktop.org/software/systemd/man/org.freedesktop.systemd1.html
+ *
+ */
+
+#if HAVE_CONFIG_H
+# include "config.h"
+#endif
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdarg.h>
+#include <string.h>
+#include <stdint.h>
+#include <unistd.h>
+#include <limits.h>
+#include <flux/core.h>
+#include <assert.h>
+
+#include <systemd/sd-bus.h>
+
+#include "sdprocess.h"
+#include "strv.h"
+
+struct sdprocess {
+    flux_t *h;
+    flux_reactor_t *reactor;
+    char *unitname;
+    char **argv;
+    char **envv;
+    int stdin_fd;
+    int stdout_fd;
+    int stderr_fd;
+
+    sd_bus *bus;
+    /* save previous bus in reconnect case */
+    sd_bus *bus_prev;
+    /* <unitname>.service */
+    char *service_name;
+    /* /org/freedesktop/systemd1/unit/<unitname>_2eservice */
+    char *service_path;
+
+    /* state watcher */
+    flux_watcher_t *w_state;
+    flux_watcher_t *w_state_prep;
+    flux_watcher_t *w_state_idle;
+    flux_watcher_t *w_state_check;
+    bool active;
+    bool active_sent;
+    bool exited;
+    bool exited_sent;
+    sdprocess_state_f state_cb;
+    void *state_cb_arg;
+    uint32_t exec_main_status;
+    uint32_t exec_main_code;
+    char *active_state;
+    char *result;
+    int wait_status;
+};
+
+static bool sdprocess_log_enable = true;
+
+static int sdprocess_state_setup (sdprocess_t *sdp,
+                                  sdprocess_state_f state_cb,
+                                  void *arg,
+                                  flux_reactor_t *reactor);
+
+void sdprocess_logging (bool enable)
+{
+    sdprocess_log_enable = enable;
+}
+
+static void sdp_log (flux_t *h, int level, const char *fmt, ...)
+{
+    va_list ap;
+    va_start (ap, fmt);
+    if (sdprocess_log_enable)
+        flux_vlog (h, level, fmt, ap);
+    va_end (ap);
+}
+
+static void sdp_log_error (flux_t *h, const char *fmt, ...)
+{
+    va_list ap;
+    va_start (ap, fmt);
+    if (sdprocess_log_enable)
+        flux_log_verror (h, fmt, ap);
+    va_end (ap);
+}
+
+static void set_errno_log (flux_t *h,
+                           int ret,
+                           const char *prefix)
+{
+    errno = -ret;
+    sdp_log_error (h, "%s", prefix);
+}
+
+static void set_errno_log_errmsg (flux_t *h,
+                                  int ret,
+                                  sd_bus_error *error,
+                                  const char *prefix)
+{
+    assert (error);
+    errno = -ret;
+    if (error->message)
+        sdp_log (h, LOG_ERR, "%s: %s", prefix, error->message);
+    else
+        sdp_log_error (h, "%s", prefix);
+}
+
+void sdprocess_destroy (sdprocess_t *sdp)
+{
+    if (sdp) {
+        int save_errno = errno;
+        free (sdp->unitname);
+        strv_destroy (sdp->argv);
+        strv_destroy (sdp->envv);
+
+        sd_bus_close (sdp->bus);
+        sd_bus_unref (sdp->bus);
+        sd_bus_close (sdp->bus_prev);
+        sd_bus_unref (sdp->bus_prev);
+        free (sdp->service_name);
+        free (sdp->service_path);
+
+        flux_watcher_destroy (sdp->w_state);
+        flux_watcher_destroy (sdp->w_state_prep);
+        flux_watcher_destroy (sdp->w_state_idle);
+        flux_watcher_destroy (sdp->w_state_check);
+
+        free (sdp->active_state);
+        free (sdp->result);
+        free (sdp);
+        errno = save_errno;
+    }
+}
+
+static char *calc_service_path (const char *unitname)
+{
+    /* note that lower case should be used, not upper case hex.  Apparently
+     * a part of dbus standard.
+     */
+    char hex2char[16] = "0123456789abcdef";
+    const char *service_path_prefix = "/org/freedesktop/systemd1/unit/";
+    int service_path_prefix_len = strlen (service_path_prefix);
+    /* _2e is escape of . (period) */
+    const char *service_path_suffix = "_2eservice";
+    int service_path_suffix_len = strlen (service_path_suffix);
+    const char *ptr;
+    char *service_path;
+    int index = 0;
+    int len = 0;
+
+    /* we must escape all special chars in the service path b/c dbus
+     * cannot handle them.
+     *
+     * escape rule is <special char> -> _<hex bytes of char>.
+     * e.g. . -> _2e
+     *
+     * note that a-f should be lower case
+     *
+     * also escape 0-9 if numeral is first character in unitname.
+     */
+
+    /* buffer length for path is:
+     *
+     * path prefix length
+     * + length unitname * 3 (*3 if all chars need to be escaped)
+     * + path suffix length
+     * + 1 (NUL byte)
+     */
+    len = service_path_prefix_len;
+    len += strlen (unitname) * 3;
+    len += service_path_suffix_len;
+    len += 1;
+
+    if (!(service_path = calloc (1, len)))
+        return NULL;
+
+    strcpy (service_path, service_path_prefix);
+    index += service_path_prefix_len;
+
+    ptr = unitname;
+    while (*ptr) {
+        if (!((*ptr) >= 'A' && (*ptr) <= 'Z')
+            && !((*ptr) >= 'a' && (*ptr) <= 'z')
+            && !(ptr > unitname && (*ptr) >= '0' && (*ptr) <= '9')) {
+            service_path[index++] = '_';
+            service_path[index++] = hex2char[*ptr >> 4];
+            service_path[index++] = hex2char[*ptr & 0xF];
+        }
+        else
+            service_path[index++] = *ptr;
+        ptr++;
+    }
+
+    strcpy (service_path + index, service_path_suffix);
+
+    return service_path;
+}
+
+static sdprocess_t *sdprocess_create (flux_t *h,
+                                      const char *unitname,
+                                      char **argv,
+                                      char **envv,
+                                      int stdin_fd,
+                                      int stdout_fd,
+                                      int stderr_fd)
+{
+    struct sdprocess *sdp = NULL;
+    int ret;
+
+    if (!(sdp = calloc (1, sizeof (*sdp))))
+        return NULL;
+    if (!(sdp->unitname = strdup (unitname)))
+        goto cleanup;
+    if (argv) {
+        if (strv_copy (argv, &sdp->argv) < 0)
+            goto cleanup;
+    }
+    if (envv) {
+        if (strv_copy (envv, &sdp->envv) < 0)
+            goto cleanup;
+    }
+    sdp->h = h;
+    if (!(sdp->reactor = flux_get_reactor (sdp->h)))
+        goto cleanup;
+    sdp->stdin_fd = stdin_fd;
+    sdp->stdout_fd = stdout_fd;
+    sdp->stderr_fd = stderr_fd;
+
+    if ((ret = sd_bus_open_user (&sdp->bus)) < 0) {
+        errno = -ret;
+        goto cleanup;
+    }
+    if (asprintf (&sdp->service_name, "%s.service", sdp->unitname) < 0)
+        goto cleanup;
+    if (!(sdp->service_path = calc_service_path (sdp->unitname)))
+        goto cleanup;
+
+    return sdp;
+
+cleanup:
+    sdprocess_destroy (sdp);
+    return NULL;
+}
+
+static int transient_service_set_stdio_properties (sdprocess_t *sdp,
+                                                   sd_bus_message *m)
+{
+    int ret;
+
+    if (sdp->stdin_fd >= 0) {
+        if ((ret = sd_bus_message_append (m,
+                                          "(sv)",
+                                          "StandardInputFileDescriptor",
+                                          "h",
+                                          sdp->stdin_fd)) < 0)
+            goto error;
+    }
+
+    if (sdp->stdout_fd >= 0) {
+        if ((ret = sd_bus_message_append (m,
+                                          "(sv)",
+                                          "StandardOutputFileDescriptor",
+                                          "h",
+                                          sdp->stdout_fd)) < 0)
+            goto error;
+    }
+
+    if (sdp->stderr_fd >= 0) {
+        if ((ret = sd_bus_message_append (m,
+                                          "(sv)",
+                                          "StandardErrorFileDescriptor",
+                                          "h",
+                                          sdp->stderr_fd)) < 0)
+            goto error;
+    }
+
+    return 0;
+
+error:
+    set_errno_log (sdp->h, ret, "error setup stdio properties");
+    return -1;
+}
+
+static int transient_service_set_environment_properties (sdprocess_t *sdp,
+                                                         sd_bus_message *m)
+{
+    int ret;
+
+    if (sdp->envv) {
+
+        if ((ret = sd_bus_message_open_container (m, 'r', "sv")) < 0)
+            goto error;
+
+        if ((ret = sd_bus_message_append (m, "s", "Environment")) < 0)
+            goto error;
+
+        if ((ret = sd_bus_message_open_container (m, 'v', "as")) < 0)
+            goto error;
+
+        if ((ret = sd_bus_message_append_strv (m, sdp->envv)) < 0)
+            goto error;
+
+        if ((ret = sd_bus_message_close_container (m)) < 0)
+            goto error;
+
+        if ((ret = sd_bus_message_close_container (m)) < 0)
+            goto error;
+    }
+
+    return 0;
+
+error:
+    set_errno_log (sdp->h, ret, "error setup environment properties");
+    return -1;
+}
+
+static int transient_service_set_cmdline_properties (sdprocess_t *sdp,
+                                                     sd_bus_message *m)
+{
+    int ret;
+
+    if ((ret = sd_bus_message_open_container (m, 'r', "sv")) < 0)
+        goto error;
+
+    if ((ret = sd_bus_message_append (m, "s", "ExecStart")) < 0)
+        goto error;
+
+    if ((ret = sd_bus_message_open_container (m, 'v', "a(sasb)")) < 0)
+        goto error;
+
+    if ((ret = sd_bus_message_open_container (m, 'a', "(sasb)")) < 0)
+        goto error;
+
+    if ((ret = sd_bus_message_open_container (m, 'r', "sasb")) < 0)
+        goto error;
+
+    if ((ret = sd_bus_message_append (m, "s", sdp->argv[0])) < 0)
+        goto error;
+
+    if ((ret = sd_bus_message_append_strv (m, sdp->argv)) < 0)
+        goto error;
+
+    if ((ret = sd_bus_message_append (m, "b", false)) < 0)
+        goto error;
+
+    if ((ret = sd_bus_message_close_container (m)) < 0)
+        goto error;
+
+    if ((ret = sd_bus_message_close_container (m)) < 0)
+        goto error;
+
+    if ((ret = sd_bus_message_close_container (m)) < 0)
+        goto error;
+
+    if ((ret = sd_bus_message_close_container (m)) < 0)
+        goto error;
+
+    return 0;
+
+error:
+    set_errno_log (sdp->h, ret, "error setup cmdline properties");
+    return -1;
+}
+
+static int transient_service_set_properties (sdprocess_t *sdp,
+                                             sd_bus_message *m)
+{
+    int ret;
+
+    if ((ret = sd_bus_message_append (m,
+                                      "(sv)",
+                                      "Description",
+                                      "s",
+                                      "libsdprocess")) < 0) {
+        set_errno_log (sdp->h, ret, "sd_bus_message_append");
+        return -1;
+    }
+
+    /* achu: no property assignments for the time being */
+
+    if ((ret = sd_bus_message_append (m, "(sv)", "AddRef", "b", 1)) < 0) {
+        set_errno_log (sdp->h, ret, "sd_bus_message_append");
+        return -1;
+    }
+
+    if ((ret = sd_bus_message_append (m,
+                                      "(sv)",
+                                      "RemainAfterExit",
+                                      "b",
+                                      true)) < 0) {
+        set_errno_log (sdp->h, ret, "sd_bus_message_append");
+        return -1;
+    }
+
+    /* Stdio */
+
+    if (transient_service_set_stdio_properties (sdp, m) < 0)
+        return -1;
+
+    /* Environment */
+
+    if (transient_service_set_environment_properties (sdp, m) < 0)
+        return -1;
+
+    /* Cmdline */
+
+    if (transient_service_set_cmdline_properties (sdp, m) < 0)
+        return -1;
+
+    return 0;
+}
+
+static int start_transient_service (sdprocess_t *sdp)
+{
+
+    sd_bus_message *m = NULL, *reply = NULL;
+    sd_bus_error error = SD_BUS_ERROR_NULL;
+    int save_errno, ret, rv = -1;
+
+    if ((ret = sd_bus_message_new_method_call (
+                                        sdp->bus,
+                                        &m,
+                                        "org.freedesktop.systemd1",
+                                        "/org/freedesktop/systemd1",
+                                        "org.freedesktop.systemd1.Manager",
+                                        "StartTransientUnit")) < 0) {
+        set_errno_log (sdp->h, ret, "sd_bus_message_new_method_call");
+        goto cleanup;
+    }
+
+    /* Name and mode */
+    if ((ret = sd_bus_message_append (m,
+                                      "ss",
+                                      sdp->service_name,
+                                      "fail")) < 0) {
+        set_errno_log (sdp->h, ret, "sd_bus_message_append");
+        goto cleanup;
+    }
+
+    /* Properties */
+    if ((ret = sd_bus_message_open_container (m, 'a', "(sv)")) < 0) {
+        set_errno_log (sdp->h, ret, "sd_bus_message_open_container");
+        goto cleanup;
+    }
+
+    if ((ret = transient_service_set_properties (sdp, m)) < 0)
+        goto cleanup;
+
+    if ((ret = sd_bus_message_close_container (m)) < 0) {
+        set_errno_log (sdp->h, ret, "sd_bus_message_close_container");
+        goto cleanup;
+    }
+
+    /* Auxiliary units */
+    if ((ret = sd_bus_message_append (m, "a(sa(sv))", 0)) < 0) {
+        set_errno_log (sdp->h, ret, "sd_bus_message_append");
+        goto cleanup;
+    }
+
+    if ((ret = sd_bus_call (sdp->bus, m, 0, &error, &reply)) < 0) {
+        set_errno_log_errmsg (sdp->h, ret, &error, "sd_bus_call");
+        goto cleanup;
+    }
+
+    rv = 0;
+cleanup:
+    save_errno = errno;
+    sd_bus_message_unref (m);
+    sd_bus_message_unref (reply);
+    sd_bus_error_free (&error);
+    errno = save_errno;
+    return rv;
+}
+
+sdprocess_t *sdprocess_exec (flux_t *h,
+                             const char *unitname,
+                             char **argv,
+                             char **envv,
+                             int stdin_fd,
+                             int stdout_fd,
+                             int stderr_fd)
+{
+    struct sdprocess *sdp = NULL;
+
+    if (!h
+        || !unitname
+        || strlen (unitname) == 0
+        || !argv) {
+        errno = EINVAL;
+        return NULL;
+    }
+
+    if (!(sdp = sdprocess_create (h,
+                                  unitname,
+                                  argv,
+                                  envv,
+                                  stdin_fd,
+                                  stdout_fd,
+                                  stderr_fd)))
+        return NULL;
+
+    if (start_transient_service (sdp) < 0)
+        goto cleanup;
+
+    return sdp;
+
+cleanup:
+    sdprocess_destroy (sdp);
+    return NULL;
+}
+
+static int check_exist (sdprocess_t *sdp)
+{
+    sd_bus_error error = SD_BUS_ERROR_NULL;
+    char *load_state = NULL;
+    int save_errno, ret, rv = -1;
+
+    if ((ret = sd_bus_get_property_string (sdp->bus,
+                                           "org.freedesktop.systemd1",
+                                           sdp->service_path,
+                                           "org.freedesktop.systemd1.Unit",
+                                           "LoadState",
+                                           &error,
+                                           &load_state)) < 0) {
+        set_errno_log_errmsg (sdp->h, ret, &error, "sd_bus_get_property_string");
+        goto cleanup;
+    }
+
+    if (!strcmp (load_state, "not-found")) {
+        errno = ENOENT;
+        goto cleanup;
+    }
+
+    rv = 0;
+cleanup:
+    save_errno = errno;
+    sd_bus_error_free (&error);
+    free (load_state);
+    errno = save_errno;
+    return rv;
+}
+
+sdprocess_t *sdprocess_find_unit (flux_t *h, const char *unitname)
+{
+    struct sdprocess *sdp = NULL;
+
+    if (!h || !unitname) {
+        errno = EINVAL;
+        return NULL;
+    }
+
+    if (!(sdp = sdprocess_create (h,
+                                  unitname,
+                                  NULL,
+                                  NULL,
+                                  -1,
+                                  -1,
+                                  -1)))
+        return NULL;
+
+    if (check_exist (sdp) < 0)
+        goto cleanup;
+
+    return sdp;
+
+cleanup:
+    sdprocess_destroy (sdp);
+    return NULL;
+}
+
+static int get_final_properties (sdprocess_t *sdp)
+{
+    sd_bus_error error = SD_BUS_ERROR_NULL;
+    uint32_t exec_main_status;
+    char *result = NULL;
+    int save_errno, ret, rv = -1;
+
+    if ((ret = sd_bus_get_property_trivial (sdp->bus,
+                                            "org.freedesktop.systemd1",
+                                            sdp->service_path,
+                                            "org.freedesktop.systemd1.Service",
+                                            "ExecMainStatus",
+                                            &error,
+                                            'i',
+                                            &exec_main_status)) < 0) {
+        set_errno_log_errmsg (sdp->h, ret, &error, "sd_bus_get_property_trivial");
+        goto cleanup;
+    }
+
+    if ((ret = sd_bus_get_property_string (sdp->bus,
+                                           "org.freedesktop.systemd1",
+                                           sdp->service_path,
+                                           "org.freedesktop.systemd1.Service",
+                                           "Result",
+                                           &error,
+                                           &result)) < 0) {
+        set_errno_log_errmsg (sdp->h, ret, &error, "sd_bus_get_property_string");
+        goto cleanup;
+    }
+
+    sdp->exec_main_status = exec_main_status;
+    if (sdp->result)
+        free (sdp->result);
+    sdp->result = result;
+    rv = 0;
+cleanup:
+    save_errno = errno;
+    sd_bus_error_free (&error);
+    if (rv < 0)
+        free (result);
+    errno = save_errno;
+    return rv;
+}
+
+static void calc_wait_status (sdprocess_t *sdp)
+{
+    assert (sdp->result);
+    if (!strcmp (sdp->result, "signal"))
+        sdp->wait_status = __W_EXITCODE (0, sdp->exec_main_status);
+    else
+        sdp->wait_status = __W_EXITCODE (sdp->exec_main_status, 0);
+}
+
+static bool check_exec_main_code_unit_done (int exec_main_code)
+{
+    if (exec_main_code == CLD_EXITED
+        || exec_main_code == CLD_KILLED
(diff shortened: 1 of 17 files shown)


chu11:
thought i'd park a checkpoint of work here.

- the job-exec module has a new `testsdexec` launches, which tests running jobs under `libsdprocess`.  It's simplified at the moment, dunno if we'd want to pull things out of `exec` to make some common code (e.g. path to job-shell, use default vs. in conf vs in jobspec ... etc.).  still need to work under multiuser / imp.

- The `libsdprocess` api has been refined further, with the most important addition being:

```
/* non-blocking wait, call callback when active / exited
 * - may never be active
 */
int flux_sdprocess_state_callback (flux_sdprocess_t *sdp,
                                   flux_sdprocess_state_f state_cb,
                                   void *arg);
```

chu11:
just parked a checkpoint in which minimal imp support appears to work! (took me a long time to figure out had to close stdin after sending `J`, doh! :-).  Perhaps we're nearing a checkpoint of something that is commitable, before moving onto next steps, such as API polishing, recovery after broker crash, etc.  minimally need to setup tests.

grondo:
One minor idea: maybe the systemd-based exec should be enabled only via a configuration option, so we don't require all users and ci containers to have `libsystemd-devel`?

chu11:
> One minor idea: maybe the systemd-based exec should be enabled only via a configuration option, so we don't require all users and ci containers to have libsystemd-devel?

yup!  I added this into configure for conditional builds (of course could have forgotten a place to put it, will resolve as i iterate).

```
PKG_CHECK_MODULES([LIBSYSTEMD], [libsystemd >= 0.23],
                                [have_libsystemd=yes], [have_libsystemd=no])
AC_CHECK_HEADER([systemd/sd-bus.h], [have_sd_bus_h=yes], [have_sd_bus_h=no])
AM_CONDITIONAL([HAVE_LIBSYSTEMD],
        [test "$have_libsystemd" = yes -a "$have_sd_bus_h" = yes])

```

chu11:
Just pushed what I believe is a good checkpoint for `libsdprocess` that can be considered for merging.

Also added is a new job-exec plugin I call `testsdexec` ("test-systemd-exec") for running jobs under it.

Note built on top of #3844 & #3872

Review considerations:

- errnos I chose in `libsdprocess` ok?
- any extra helper functions that would be useful?
- any "polishing" that would be worthwhile for this round?

Note that there are still many things to do after this:

- functions to discover already running jobs under systemd by unitname/servicename
- running under imp tests
- support configuring cgroup properties / other systemd properties
- signaling under imp
- setup / configure stdio better

chu11:
hmm inception tests failing, probably b/c `test-inception.sh` doesn't take into account conditional testing based on environment.

`flux mini bulksubmit --watch -q ./{} ::: t[0-9]*.t python/t*.py lua/t*.t`

hmmm, what would be the easiest way to conditional test this?  Perhaps in the Makefile all tests could be output to a file and that could be read in via `test-inception.sh`?

grondo:
> hmm inception tests failing, probably b/c test-inception.sh doesn't take into account conditional testing based on environment.

Not sure what you mean by conditional testing, but if my assumption is right, for sharness tests the conditional should be in the test itself, not done via the makefile or other test driver.

If the correct environment doesn't exist then the sharness test should skip everything. For example, grep for any tests that use `skip_all`.

chu11:
> If the correct environment doesn't exist then the sharness test should skip everything. For example, grep for any tests that use skip_all.

Ahh, I was are of `skip_all_unless_have` but not the generic `skip_all`.

```
#  Don't run if systemd environment variables not setup
if test -z "$DBUS_SESSION_BUS_ADDRESS" \
     -o -z "$XDG_RUNTIME_DIR"; then  
    skip_all='DBUS_SESSION_BUS_ADDRESS and/or XDG_RUNTIME_DIR not set'
    test_done
fi

uid=`id -u`
userservice="user@${uid}.service"
if ! systemctl list-units | grep ${userservice}
then
    skip_all='systemd user service not running'
    test_done
fi
```

also added a sorta hackish check for the `systemctl list-units` check into `libsdprocess/tests` as well.

chu11:
re-pushed with a few extra commits on top.  These commits support:
- `flux_sdprocess_find_unit()` - find a unit in systemd so that it can be monitored / check status / cleaned up, etc.
- `flux_sdprocess_list()` - list current user's systemd units, with an optional pattern to glob for
   - for consideration, do we need additional filters / things to filter on?

grondo:
> for consideration, do we need additional filters / things to filter on?

I don't think so. Seems like `job-exec` could discover all job shell service units with `flux_sdprocess_list()` using a pattern that matches job units belonging to the system instance, then attach to running units with `flux_sdprocess_find_unit()`?

This pretty much solves the issue of reattaching to running job shells for job-exec.

grondo:
I don't know why, but I thought we were going to be able to start up a systemd user instance in one of our existing container builds (instead of only in the fluxorama/system test image). Of course systemd refuses to start a user instance if the system has not been booted with systemd, so I guess it isn't possible.

We'll probably have to add a build to the ci that runs under the system instance in order to run the sharness test added here. In that case I'm not sure if we want to run all the tests, or just the "system instance" related tests which require systemd. Requires more thought.

chu11:
re-pushed, added a stupid comment typo fix, but more importantly a new set of tests that simulate a broker going down / back up.

chu11:
re-pushed.  The big change is instead of having a `testsdexec` job-exec plugin, its now just the `sdexec` plugin.  There's also a new commit that splices out the imp path, job-shell, etc. configuration that's common to `sdexec` and `exec`.

chu11:
just pushed a `fixup` that adds a check for the flux configuration:

```
[exec]
sdexec = "true"
```

which would tell the `sdexec` plugin to be used if the above is enabled.  I also removed the `sdexec` enabling via jobsec.  This is for #3970.  I didn't squash it since I wasn't sure if this fix should be in this PR or not, but I figure it can be.

grondo:
> just pushed a fixup that adds a check for the flux configuration:

Nice! I would suggest a more sysadmin friendly configuration, e.g. `use-systemd = true`. Note also that TOML supports boolean values so the `true` should not be quoted.

chu11:
> Nice! I would suggest a more sysadmin friendly configuration name, e.g. use-systemd = true. Note also that TOML supports boolean values so the true should not be quoted.

Ahh, I tried the boolean support yesterday but couldn't get it to work and figured maybe toml didn't support it.  Probably just did something stupid.

re-pushed with your suggestion and squashed the fixups.  The only way to enable `sdexec` now is via the flux config:

```
[exec]
use-systemd = true
```

chu11:
just re-pushed, just to fix up some upstream merge conflicts (#3996 and #3872)

grondo:
Now that #3844 was finally merged, this will need a rebase. Sorry!

chu11:
thanks @grondo, rebased & re-pushed

chu11:
offline discussion w/ @grondo - a convenience function to convert unitname to PID would be useful.  This should be easy:

```
MainPID and ControlPID contain the main and control PID of the service. The main PID is the current main PID of the service and is 0 when the service currently has no main PID. The control PID is the PID of the current start/stop/reload process running and is 0 if no such process is currently running. That means that ExecMainPID and MainPID differ in the way that the latter immediately reflects whether a main process is currently running while the latter possible contains information collected from the last run even if the process is no longer around.
```

chu11:
offline discussion, @grondo saw a busy loop 

```
Thread 16 (Thread 0x7efc8bfff700 (LWP 21173)):
#0  0x00007efca10c56c9 in getpid_cached () from /usr/lib64/libsystemd.so.0
#1  0x00007efca10411e5 in bus_pid_changed () from /usr/lib64/libsystemd.so.0
#2  0x00007efca1044de8 in bus_process_internal ()
   from /usr/lib64/libsystemd.so.0
#3  0x00007efca1383253 in watcher_properties_changed_cb (r=<optimized out>, 
    w=0x7efc84028220, revents=<optimized out>, arg=0x7efc84019950)
    at sdprocess.c:933
#4  0x00007efccb58ba93 in ev_invoke_pending (loop=0x7efc84007bf0) at ev.c:3372
#5  0x00007efccb58fa35 in ev_run (flags=0, loop=0x7efc84007bf0) at ev.c:3775
#6  ev_run (loop=0x7efc84007bf0, flags=0) at ev.c:3623
#7  0x00007efccb55ea1b in flux_reactor_run (r=0x7efc84001d20, 
    flags=flags@entry=0) at reactor.c:128
#8  0x00007efca13554a8 in mod_main (h=0x7efc84006180, argc=<optimized out>, 
    argv=0x7efc84008440) at job-exec.c:1269
#9  0x000000000040d666 in module_thread (arg=0xfe3500) at module.c:195
#10 0x00007efccb32a17a in start_thread () from /usr/lib64/libpthread.so.0
#11 0x00007efcc94f5dc3 in clone () from /usr/lib64/libc.so.6
```


```
index bcc9bb17f..464f93d07 100644
--- a/src/common/libsdprocess/sdprocess.c
+++ b/src/common/libsdprocess/sdprocess.c
@@ -927,10 +927,11 @@ static void watcher_properties_changed_cb (flux_reactor_t 
*r,
                                            int revents,
                                            void *arg)
 {
+    int n = 0;
     flux_sdprocess_t *sdp = arg;
 
     if (revents & FLUX_POLLIN)
-        while ( sd_bus_process (sdp->bus, NULL) ) { }
+        while ( sd_bus_process (sdp->bus, NULL) && ++n < 100 ) { }
     else {
         flux_log (sdp->h, LOG_ERR, "Unexpected revents: %X", revents);
         return;
```

he saw during a restart flux shutdown, so I'm assuming perhaps a corner case where `sd_bus_process()` is regularly returning an error?  So we need to check `sd_bus_process()` for an error to not busy loop.

chu11:
have been unable to reproduce the busy loop that @grondo saw, i'm assuming its a racy scenario.  Unsure if can be reproduced more easily with some of the changes grondo was experimenting with.  Simply wanted to try and get more debugging via

```
diff --git a/src/common/libsdprocess/sdprocess.c b/src/common/libsdprocess/sdprocess.c
index bcc9bb1..a3a35a6 100644
--- a/src/common/libsdprocess/sdprocess.c
+++ b/src/common/libsdprocess/sdprocess.c
@@ -929,8 +929,19 @@ static void watcher_properties_changed_cb (flux_reactor_t *r,
 {
     flux_sdprocess_t *sdp = arg;
 
-    if (revents & FLUX_POLLIN)
-        while ( sd_bus_process (sdp->bus, NULL) ) { }
+    if (revents & FLUX_POLLIN) {
+        int rc;
+#if 0
+        while ( (rc = sd_bus_process (sdp->bus, NULL)) && rc > 0) { }
+        if (rc < 0)
+            flux_log (sdp->h, LOG_ERR, "sd_bus_process: %d", rc);
+#else
+        while ( (rc = sd_bus_process (sdp->bus, NULL))) {
+            if (rc < 0)
+                flux_log (sdp->h, LOG_ERR, "sd_bus_process: %d", rc);
+        }
+#endif
+    }
     else {
         flux_log (sdp->h, LOG_ERR, "Unexpected revents: %X", revents);
         return;
```

chu11:
repushed with some additions per some offline / above discussions

- added `flux_sdprocess_pid()` to get the main PID of the process launched by systemd
- `sdexec_kill()` will now call `flux imp <signal> <pid>` if job-exec in run in multiuser mode
  - AFAICT we needed the kill to finish before `sdexec_kill()` returns, so there were multiple options on how to do this.  I elected to use `libsubprocess` and create a temporary / local reactor to handle the wait for the command to complete (logging stderr/stdout if there's an error).  Not sure if this was a good idea or not.

grondo:
> AFAICT we needed the kill to finish before sdexec_kill() returns, 

No, the kill should be asynchronous, e.g. see the `exec_kill()` implementation for "normal" execution:

 https://github.com/flux-framework/flux-core/blob/22720e131610ca8cf3ea297f32cdfbd3b402368e/src/modules/job-exec/exec.c#L324-L351

Note, however, that at least in the current code, the kill should take a reference on the job until the future is fulfilled to avoid trying to reference a destructed job on error.

chu11:
> No, the kill should be asynchronous, e.g. see the exec_kill() implementation for "normal" execution:

Ahhh, I had misread the code.  The `kill` function (here `exec_kill`) should return immediately, but the "kill action" can be asynchronous.

chu11:
argh, i spent most of the day hunting a race condition.  Ends up I have something racy within `libsdprocess`.  When the sd-bus library runs off to do / start some of its callbacks, there's a chance `job-exec` will end the job (i.e. it finishes, gets canceled, etc.) and destroys a bunch of data structures.  But systemd doesn't know this once it enters the callback, b/c it effectively runs outside of the flux reactor.  So probably gotta do some reference counting or something.

grondo:
I'm probably oversimplifying, but can you use `jobinfo_incref()` to take a reference on the `struct jobinfo *` with sdexec until all systemd callbacks are known to be complete?

chu11:
> I'm probably oversimplifying, but can you use jobinfo_incref() to take a reference on the struct jobinfo * with sdexec until all systemd callbacks are known to be complete?

Ahh, that's a good idea, b/c the `exit` routine won't be called until `jobinfo`'s refcount goes to zero.

chu11:
hmmm, so while my fix above solved one problem (removed SEGV races), it appears to have caused another problem.  For some unknown reason, the bus connections occasionally die (ECONNRESET).  No known reason for it at the moment.  Still digging.

chu11:
re-pushed

- fix issue with `sd_bus_progress()` returning ECONNRESET if a job takes longer than 25 seconds.  This is believed to be due to the default BUS_DEFAULT_TIMEOUT of 25 seconds.  Unfortunately changing this is supported in a newer version of systemd, so the workaround is to reconnect the bus if ECONNRESET is hit.

- add some extra tests to cover some corner cases for jobs running > 25 seconds.  These are set to LONGTEST so they won't normally run.

- add a few more corner case tests on top :-)

chu11:
re-pushed, adding a configure check for `sd_bus_set_method_timeout()` and using it if it is available.  Verified this works on my ubuntu box, but admittedly with just a dumb reproducer.

chu11:
re-pushed, adding manpage entry for `use-systemd` into the new `flux-config-exec`

src/modules/job-exec/sdexec.c
```diff
@@ -0,0 +1,584 @@
...
+
+static int sdexec_start (struct jobinfo *job)
+{
+    struct sdexec *se = job->data;
+    char *unitname = NULL;
+    char **cmdv = NULL;
+    char **envv = NULL;
+    int rv = -1;
+
+    if (asprintf (&unitname, "flux-sdexec-%ju", (uintmax_t)job->id) < 0) {
```
grondo, line 348: Something I just considered: The unitname for jobs should include the current rank as well as jobid in case we are running tests with multiple brokers per node.

chu11: good idea!

chu11:
re-pushed, adding @grondo's suggested tweak above.

grondo:
I haven't fully digested this PR quite yet, but I've updated my branch that combines this PR and #4016 and we still require this patch (or something like it) to fully support reattaching job-exec to running jobs after broker restart:



```diff
--- a/src/modules/job-exec/sdexec.c
+++ b/src/modules/job-exec/sdexec.c
@@ -380,10 +380,23 @@ static int sdexec_start (struct jobinfo *job)
                                          cmdv,
                                          envv,
                                          se->stdin_fds[1],
-                                         se->stdout_fds[1],
-                                         se->stderr_fds[1]))) {
-        jobinfo_fatal_error (job, errno, "flux_sdprocess_exec");
-        goto cleanup;
+                                         -1,
+                                         -1))) {
+        if (errno == EEXIST) {
+            flux_log (se->h, LOG_INFO, "%ju: attaching to unit=%s",
+                      (uintmax_t) job->id,
+                      unitname);
+            if (!(se->sdp = flux_sdprocess_find_unit (se->h, unitname)))
+                jobinfo_fatal_error (job, errno,
+                                     "flux_sdprocess_find %s",
+                                     unitname);
+            flux_log (se->h, LOG_INFO, "%ju: found unit=%s",
+                      (uintmax_t) job->id, unitname);
+        }
+        else {
+            jobinfo_fatal_error (job, errno, "flux_sdprocess_exec");
+            goto cleanup;
+        }
     }
 
     if (flux_sdprocess_state (se->sdp, state_cb, se) < 0) {
```

Should we plan to first merge #4106, plus minor required changes to the job shell to support reconnect, then update this PR with full restart functionality and tests?

chu11:
@grondo this PR didn't add any "re-attach" code.  I had initially thought of it as independent of this PR, although that was a long time ago and things are different now.

> Should we plan to first merge #4016, plus minor required changes to the job shell to support reconnect, then update this PR with full restart functionality and tests?

That sounds like a good plan.  This was my initial stab at reattach:

https://github.com/chu11/flux-core/commit/00e7ede88b2dd9ebe00445d30862f9ca876de751

which logically isn't really that different from what you did, but does look at the `job->reattach` flag instead of checking for `EEXIST`.

chu11:
> Merge this PR now, then add a separate PR with the reconnect and reattach support

I like this idea.  I sort of like the idea that `libsdprocess` is sort of its own PR.  Since we don't support reattach yet, the `sdexec` plugin doesn't really offer much different than what's already there.  Of course we may hit a corner case or two later on, but I think that's something we can deal with later as needed?

grondo:
> I like this idea. I sort of like the idea that libsdprocess is sort of its own PR.

Ok, are you suggesting to remove the `sdexec` job-exec implementation from this PR?

>  Since we don't support reattach yet, the sdexec plugin doesn't really offer much different than what's already there.

Good point, which is why it would be nice merge when we've got an entire new feature, instead of bit by bit.

However, I could go either way on this one. If we can merge libsdprocess (even though by itself it doesn't add any new feature nor have any users) then that may help us develop the actual features, which is a good thing.

chu11:
> Ok, are you suggesting to remove the sdexec job-exec implementation from this PR?

I wasn't, although now that you mention it, it sort of makes sense.  

> However, I could go either way on this one. If we can merge libsdprocess (even though by itself it doesn't add any new feature nor have any users) then that may help us develop the actual features, which is a good thing.

Sounds good.  I'll re-push with `sdexec` removed.

chu11:
re-pushed, removing the `sdexec` implementation + tests, but leaving the "simulated broker crash" tests, which only use `libsdprocess`.

grondo:
Perhaps all the `job-exec` changes should  be moved to a new `sdexec` PR (some refactoring still in here which doesn't "fit" with libsdprocess)?

Also, coverage is a little low, there are whole functions in `libsdprocess` that don't have coverage -- I know this can be difficult but any way to bump coverage a little?

chu11:
> Perhaps all the job-exec changes should be moved to a new sdexec PR (some refactoring still in here which doesn't "fit" with libsdprocess)?

Ahh, good thought.

> Also, coverage is a little low, there are whole functions in libsdprocess that don't have coverage -- I know this can be difficult but any way to bump coverage a little?

Lemme see if I can get it up.  Perhaps some error cases (like `EEXIST`) might be able to bump it up.

Codecov show's several functions that are completely missed, which should be impossible (both in the tests pass and the fact that there is coverage of the calls to those functions).  Not sure why codecov is misses those paths.

chu11:
re-pushed, adding a EEXIST test, which hopefully will get atleast one of the set_error functions to be called.  I'm not sure why the last codecov missed two big `get_properties` functions.  Will see if it was a fluke.

Also (similar to work in #4060, log to LOG_DEBUG instead of LOG_ERR.

chu11:
got coverage up a tad, but not much.  For some reason `get_properties_int()` and `get_properties_string()` aren't covered in the diff coverage, even though the calls to those functions are covered.  Dunno if this is a corner case with codecov?

grondo:
I can look into this. Not sure if it is the case here, but sometimes a strategic `_gcov_flush()` call is required to get full coverage when process exits abnormally or calls exec...

chu11:
> I can look into this. Not sure if it is the case here, but sometimes a strategic _gcov_flush() call is required to get full coverage when process exits abnormally or calls exec...

Hmmm, doubt its the case here, since this code is ultimately "monitoring" a process, its not in a process that's exiting.

Some of this code is called in a path that originates via a `systemd` callback routine.  No idea if that could be the cause, like the code gets optimized in a funky way b/c of that fact.

grondo:
Ok, can you try adding this commit (on my fork of your `libsdprocess_B` branch)?

```diff
commit c519c7f5ef46b49e9e9c4ea9698bd01b4d7e8bb6 (HEAD -> libsdprocess_B, grondo/libsdprocess_B)
Author: Mark A. Grondona <mark.grondona@gmail.com>
Date:   Tue Jan 25 08:55:25 2022 -0800

    ci: clear CFLAGS on coverage runs
---
 src/test/checks_run.sh | 2 ++
 1 file changed, 2 insertions(+)

diff --git a/src/test/checks_run.sh b/src/test/checks_run.sh
index a4b2737e6..fb7ce62e4 100755
--- a/src/test/checks_run.sh
+++ b/src/test/checks_run.sh
@@ -116,6 +116,8 @@ if test "$COVERAGE" = "t"; then
 
        rm -f .coverage .coverage*
 
+       #  Disable any optimization in CFLAGS
+       CFLAGS=
        ARGS="$ARGS --enable-code-coverage"
        CHECKCMDS="\
        ENABLE_USER_SITE=1 \

```

chu11:
re-pushed with your CI change, `get_properties()` still seems to be missed :-(

grondo:
Ah, I wonder if it is due to the way the system tests are run. Let me try a couple other things. Sorry!

grondo:
There are ~7 minor whitespace errors in this PR. Perhaps while I give the code a quick review, you can find and fix those?

Actually, you could just try:
```console
$ git rebase --whitespace=fix origin/master
```

chu11:
> There are ~7 minor whitespace errors in this PR.

hmm, how are you seeing these? 

`git diff master --check`

doesn't give me any errors.

src/common/libsdprocess/sdprocess.c
```diff
@@ -0,0 +1,1830 @@
...
+                                           "org.freedesktop.systemd1.Service",
+                                           "Result",
+                                           &error,
+                                           &result)) < 0) {
+        set_errno_log_error (sdp->h, ret, &error, "sd_bus_get_property_string");
+        goto cleanup;
+    }
+    
+    sdp->exec_main_status = exec_main_status;
+    sdp->result = result;
```
grondo, line 679: Valgrind caught a leak here. Do you need to `free (sdp->result)` before setting the new value?

src/common/libsdprocess/test/sdprocess.c
```diff
@@ -0,0 +1,1815 @@
...
+    char *cpy = NULL;
+    char *ptr;
+    int len = 0;
+
+    if (!(cpy = strdup (cmdline)))
+        BAIL_OUT ("strdup");
+
+    while ((ptr = strtok (cpy, " "))) {
+        len++;
+        cpy = NULL;
```
grondo, line 69: Since `cpy` is set to NULL here, it is leaked on function exit.

src/common/libsdprocess/test/sdprocess.c
```diff
@@ -0,0 +1,1815 @@
...
+        BAIL_OUT ("calloc");
+
+    if (!(cpy = strdup (cmdline)))
+        BAIL_OUT ("strdup");
+
+    i = 0;
+    while ((ptr = strtok (cpy, " "))) {
+        if (!(strv[i++] = strdup (ptr)))
+            BAIL_OUT ("strdup");
+        cpy = NULL;
```
grondo, line 99: Since `cpy` is set to NULL here it is leaked on exit from this function.

grondo:
In my copy they were highlighted red in the diff output. However, let me make absolutely sure I'm up to date.

grondo (commented):
Made an initial quick pass. There were a couple memleaks discovered by valgrind.
Hint: valgrind can be run directly on the unit test(s).

For the sdexec job-exec implementation, we may also want to add a workload to valgrind that runs a few jobs under sdexec.

grondo:
> However, let me make absolutely sure I'm up to date.

Strange, I tried to reset my branch to yours, but I still get errors:
```
 grondo@asp:~/git/f.git$ git diff --check origin/master
src/common/libsdprocess/sdprocess.c:232: trailing whitespace.
+            sdp->service_path[index++] = hex2char[*ptr >> 4]; 
src/common/libsdprocess/sdprocess.c:677: trailing whitespace.
+    
src/common/libsdprocess/sdprocess.c:963: trailing whitespace.
+                       
src/common/libsdprocess/sdprocess.c:1440: trailing whitespace.
+    int save_errno, ret, rv = -1; 
src/common/libsdprocess/test/sdprocess.c:1715: trailing whitespace.
+    /* Instead of checking if user service running via some equivalent to 
src/common/libsdprocess/test/sdprocess.c:1807: trailing whitespace.
+    
t/sdprocess/sim-broker.c:152: trailing whitespace.
+    
t/sdprocess/sim-broker.c:207: trailing whitespace.
+    
t/sdprocess/sim-broker.c:277: trailing whitespace.
+    
t/t2410-exec-systemd.t:10: trailing whitespace.
+     -o -z "$XDG_RUNTIME_DIR"; then  
```

Can you spot check those and let me know if I'm still out-of-date somehow?

chu11:
> Can you spot check those and let me know if I'm still out-of-date somehow?

Weird.  Some spot checks, there's no whitespace errors on my end.  Even looking with `:set list` in `vi` didn't show em and my emacs didn't highlight anything.

chu11:
re-pushed with the memleak fixes you noted above. sorry, totally forgot to run a valgrind check.  Only remaining one is a memleak in `libgcrypt` that i suspect is out of our control.

grondo:
Ok, not sure what is going on with my local copy. Ignore my comments on whitespace :facepalm:

grondo:
BTW, I couldn't get to the bottom of the coverage issue. Likely this is something to do with the "system" coverage builder or reports. We'll have to address that in general later if it starts to become more of a problem (we'll also have to address how to get coverage of code that only runs in the system instance itself. An issue for another day), but for now we'll just move on.

grondo:
I was also going to suggest we transition to using something like `llog.h` here instead of `flux_log()` directly, however I don't really think that will be necessary, since the only real use case for libsdprocess at this time will be from the job-exec module.

grondo:
One other quick suggestion though:

It will be useful to determine from `flux version` if Flux was built with systemd support. Would it make sense to add `+libsystemd==VERSION` to the `build-options:` output of `flux version` when systemd support is available?

Then in the testsuite, libsdprocess and sdexec tests should be skipped if this option isn't available, even if systemd user instance is running and appropriate env variables are set.

chu11:
> I was also going to suggest we transition to using something like llog.h here instead of flux_log() directly, however I don't really think that will be necessary, since the only real use case for libsdprocess at this time will be from the job-exec module.

I'll attach to my other issue for `libsubprocess` (#4065) for a todo later on

chu11:
> It will be useful to determine from flux version if Flux was built with systemd support. Would it make sense to add +libsystemd==VERSION to the build-options: output of flux version when systemd support is available?

I like this idea!

chu11:
re-pushed, per @grondo suggestion above `flux version` now outputs if `libsystemd` was part of the build.

```
fluxorama|/usr/src 1>flux version
commands:    		0.33.0-109-g5e67241f4
libflux-core:		0.33.0-109-g5e67241f4
broker:  		0.33.0-109-g5e67241f4
FLUX_URI:		local:///tmp/flux-WaqnBH/local-0
build-options:		+libsystemd+hwloc==2.1.0+zmq==4.3.4
```

and added a check into the tests to also check for this (but still check for the env and user service).

```
if ! flux version | grep libsystemd; then
    skip_all='flux not built with libsystemd'
    test_done
fi
```

chu11:
> codecov/patch Failing after 1s — 71.58% of diff hit (target 83.50%)

restarting codecov, not sure why it says failed after 1s.

grondo:
Sorry, now that I see we didn't include `lib` on `libhwloc` and `libzmq`, lets drop `lib` and just use `+systemd` for consistency. This was totally my fault, so let me know what I can do to make it up to you  :stuck_out_tongue_winking_eye:

chu11:
@grondo haha, no worries.  re-pushed with fixes.

src/common/libsdprocess/sdprocess.c
```diff
@@ -0,0 +1,1835 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <stdint.h>
+#include <unistd.h>
+#include <limits.h>
+#include <assert.h>
+
+#include <systemd/sd-bus.h>
+
+#include "src/common/libflux/reactor.h"
```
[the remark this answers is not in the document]
garlick: oops, sorry I was starting a review and clicked the wrong button.

src/common/libsdprocess/sdprocess.c
```diff
@@ -0,0 +1,1835 @@
...
+        char **ptr = strv;
+        while (*ptr) {
+            len++;
+            ptr++;
+        }
+    }
+    return len;
+}
+
+static int strv_copy (char **strv, char ***cpyp)
```
garlick, line 147: Would argz work here instead of the strv functions?

If not, maybe the strv functions should be in their own source module with a unit test?

src/common/libsdprocess/sdprocess.c
```diff
@@ -0,0 +1,1835 @@
...
+        free (sdp->service_path);
+
+        flux_watcher_stop (sdp->w_state);
+        flux_watcher_destroy (sdp->w_state);
+        flux_watcher_stop (sdp->w_state_prep);
+        flux_watcher_destroy (sdp->w_state_prep);
+        flux_watcher_stop (sdp->w_state_idle);
+        flux_watcher_destroy (sdp->w_state_idle);
+        flux_watcher_stop (sdp->w_state_check);
+        flux_watcher_destroy (sdp->w_state_check);
```
garlick, line 125: `flux_watcher_destroy()` calls stop internally, so no need to do it here too.

src/common/libsdprocess/sdprocess.c
```diff
@@ -0,0 +1,1835 @@
...
+    /* we must escape all special chars b/c dbus cannot handle them.
+     * escape rule is <special char> -> _<hex bytes of char>.
+     * e.g. . -> _2e
+     *
+     * note that a-f should be lower case
+     *
+     * also escape 0-9 if numeral is first character in unitname.
+     */
+    len = service_path_prefix_len;
+    len += strlen (sdp->unitname) * 3;
```
garlick, line 219: Suggestion: break out the escaping of special characters to a "string_copy_with_escape()" function or similar.

This code is a bit hard to follow to ensure that it doesn't exceed its buffer space and that the resulting string is always null terminated.  The use of `strcpy()` tends to raise a red flag.

src/common/libsdprocess/sdprocess.c
chu11: My understanding is that `argz` is a single buffer w/ `'\0` separating strings?  So it wouldn't work in this case.

A `strv` file sounds like a good idea, I could use it in the `sdprocess` tests as well.

garlick: Yes but there are convenience functions for building argv style arrays of pointers and the reverse.  Not sure if it's a net win complexity to switch though - you'd probably need to consider that.

(See `argz_extract()` for example)

src/common/libsdprocess/sdprocess.c
garlick: (I just mean the use of `strcpy()` invites closer scrutiny - so the code that uses it should be fairly clear/commented so that the scrutiny isn't too time consuming)

chu11:
re-pushed with a lot of fix ups for the issues commented by @garlick above.

chu11:
Oh whoops, i accidentally hit "closed with comment", re-opened of course :-)

grondo:
fixups LGTM! There's a couple places in the new `strv.c` that aren't covered by any tests. Now that there's a unit test it should be possible to ensure these paths are covered? Specifically `strv_len_delim()` is not covered (or at least codecov doesn't think so, maybe just another optimization issue I didn't actually check yet.)

src/common/libsdprocess/sdprocess.h
```diff
@@ -0,0 +1,158 @@
...
+/* Find process already launched under systemd
+ *
+ * Setup of XDG_RUNTIME_DIR and DBUS_SESSION_BUS_ADDRESS environment
+ * variables are assumed.  If not, systemd will return error.
+ */
+flux_sdprocess_t *flux_sdprocess_find_unit (flux_t *h,
+                                            const char *unitname);
+
+flux_sdprocess_t *flux_sdprocess_local_find_unit (flux_reactor_t *r,
+                                                  const char *unitname);
```
garlick, line 93: These two functions seem to differ only in that one takes a `flux_t` handle and the other a `flux_reactor_t`.  How is one "local" and the other not?  Could the one that takes a `flux_t` handle be dropped and callers could just get the reactor from the handle themselves?

Same with `flux_sdprocess_local_exec()` and `flux_sdprocess_exec()`.

src/common/libsdprocess/sdprocess.c
```diff
@@ -0,0 +1,1835 @@
...
+    int ret;
+
+    if ((ret = sd_bus_message_append (m,
+                                      "(sv)",
+                                      "Description",
+                                      "s",
+                                      "libsdprocess")) < 0) {
+        set_and_log_errno (sdp->h, ret, "sd_bus_message_append");
+        return -1;
+    }
```
garlick, line 260: There is so much repetition in this function....does each error need to be separately logged or could they be consolidated with a goto that doesn't call out the function that failed?

garlick (commented):
Couple more comments, somewhat high level so wanted to get your take before finishing my review.  BTW I don't know how @grondo feels about it but it might be good to squash the fixups so far.

src/common/libsdprocess/sdprocess.h
grondo: I could be wrong but this is modeled after libsubprocess, which has `flux_exec(3)` and `flux_local_exec(3)`, which takes a reactor and therefore only invokes processes "locally", i.e. as a direct child of the calling process.

I could be wrong, but for `libsdprocess` the requirement for a function that takes a reactor instead of a handle is probably only for testing purposes, i.e. so that unit tests can be written without connecting to a broker. Since it isn't expected that the "local" functions will ever be used, maybe there is a better way to create the unit tests (using libtestutil?) that would allow @chu11 to remove the redundant functions and reduce the amount of maintained but unused code.

grondo:
> BTW I don't know how @grondo feels about it but it might be good to squash the fixups so far.

Yes, that sounds good to me.

src/common/libsdprocess/sdprocess.h
garlick: I figured - just thought maybe the "local" part ought to be taken out of the API since everything is actually local.  But if there's a need I'm not seeing, it is not worth creating a kerfuffle.

Also: it looks like `flux_t` is only used internally so `sd_and_log_errno()` can call `flux_log( LOG_DEBUG)` when an error occurs.  I wonder if maybe the logging should be dropped or made an optional callback?  Meh, maybe it's fine.

grondo: Yeah, it would be a good if we excise the local functions, only to have less code to maintain. I was having a similar thought, but then again this API isn't exported publicly and is only for internal use by job-exec..

chu11:
> fixups LGTM! There's a couple places in the new strv.c that aren't covered by any tests. Now that there's a unit test it should be possible to ensure these paths are covered? Specifically strv_len_delim() is not covered (or at least codecov doesn't think so, maybe just another optimization issue I didn't actually check yet.)

Ugh, yeah I think there must be something up with codecov.  If you look a few lines down, the call to that function is literally covered.

image

While its always possible I royally screwed up my tests and nothing is being tested (the rest of `strv_create()` is covered after the call), the `int len = 0`, `if (str)`, and `return len` lines should minimally be covered.

src/common/libsdprocess/sdprocess.h
chu11: ahhh yes, unlike `libsubprocess` the `local` may not make as much sense.  Perhaps I should stick into a `_private.h` or similar.

grondo:
> Ugh, yeah I think there must be something up with codecov.

Yeah, probably the same issue as before. I'll have to get back to investigating that at some point.

src/common/libsdprocess/sdprocess.h
grondo: Or maybe the unit tests can just use `loopback_create()` from `libtestutil` to get a local `flux_t` and you can drop the duplicate functions? (Didn't check if that is feasible)

chu11:
just re-pushed.  squashed the prior batch of fixups, the fixups address the comments above.

- went with moving some functions into a `sdprocess_private.h` file for testing.  I don't fully understand how the `loopback_create()` test handle works, but it does seem to have some behavior differences that lead to some quirks.

- i created a number of helper functions for previous "long chunks" in `transient_service_set_properties()`, there by removing some of the error logging repetition, but still keeping the log messages a bit helpful.

garlick:
> but then again this API isn't exported publicly and is only for internal use by job-exec..

It is exported actually.  That should probably be fixed (remove `flux_` function prefixes and don't install header file).

> I don't fully understand how the loopback_create() test handle works, but it does seem to have some behavior differences that lead to some quirks.

Hmm, in the loopback handle,  there is no message handler for the log messages, and no separate thread to handle them asynchronously.  That can be resolved, for example register a `flux_log_f` function with `flux_log_set_redirect()` like [here](https://github.com/flux-framework/flux-core/blob/f749c06621ca09588b99d66850458298de0c3b92/src/broker/test/overlay.c#L648).  But my vote would be to replace the direct calls to `flux_log()` with a callback interface so the user of libsdprocess can control it, e.g.
```c
typedef void (*sdprocess_error_f)(int errnum, const char *errmsg, void *arg);

void sdprocess_set_error_handler (struct sdprocess *sdp, sdprocess_error_f cb, void *arg);
```
Then the main functions just take a reactor not a `flux_t` handle and there would be no logging if a handler is not registered.  In the test you can register a handler that calls `diag()`.

grondo:
> It is exported actually. That should probably be fixed (remove flux_ function prefixes and don't install header file).

Oh, I didn't even notice that the header file was being installed. Good catch!

Also I assumed the libsdprocess library wasn't being linked into libflux-core.so, but I like the idea of removing the `flux_` prefix and just having all the functions called `sdprocess_*`. (The naming is probably my fault since I suggested we make a "clone" of libsubprocess, which at this point seems like a bit of a bad idea on my part)

> But my vote would be to replace the direct calls to flux_log() with a callback interface so the user of libsdprocess can control it, e.g.

We also have the llog interface (which was suggested earlier), but what you've proposed is fine too.

chu11:
> It is exported actually. That should probably be fixed (remove flux_ function prefixes and don't install header file).

Unlike `libsubprocess` which a bit more of a generic library, would non `job-exec` executor modules be developed by users in the future? and might need `libsdprocess`?  Otherwise yes, we should remove `flux_` prefixes.

grondo:
>  would non job-exec executor modules be developed by users in the future? and might need libsdprocess? Otherwise yes, we should remove flux_ prefixes.

I don't foresee that need at this point, but easy enough to create `flux_` wrappers if necessary in the future.

garlick:
> Also I assumed the libsdprocess library wasn't being linked into libflux-core.so

Oh I think you're right about that, now that I look closer.

garlick:
> We also have the llog interface (which was suggested earlier), but what you've proposed is fine too.

Sorry again, I forgot that was mentioned.  If it's a fit here it would probably be preferable to reinventing a new callback...

grondo:
> Sorry again, I forgot that was mentioned. If it's a fit here it would probably be preferable to reinventing a new callback...

Eh... you still have to have a subsystem-specific callback with the llog interface, it just has the same prototype. I kind of have mixed feelings about it just now. It is a better fit if you want to be able to enable debug, verbose, error, and critical messages (and include file, function, and line info). But a bit heavyweight for just logging errors.

chu11:
> Then the main functions just take a reactor not a flux_t handle and there would be no logging if a handler is not registered. In the test you can register a handler that calls diag().

While writing tests, it occurred to me that alot of (possibly most) logging cases would occur in `sdprocess_exec()` and `sdprocess_find_unit()` (not to mention `sdprocess_list()`).  In other words, before the `sdp` context is created for the user.

Do we want to pass in a `error_cb` in those functions?  It's not the style we've done before and not what I would prefer.

Would a global `sdprocess_set_error_handler (sdprocess_error_f cb, void *arg)` be offensive?

OR

perhaps there should be a far more simple / basic `sdprocess_create_context ()` function done before `sdprocess_exec()`, etc.

chu11:
> Would a global sdprocess_set_error_handler (sdprocess_error_f cb, void *arg) be offensive?

Thinking about it a bit more, perhaps this is best, but should be viewed more in the vein of "for debugging", not really for the "average user".  One of the reasons I do somewhat excessive debugging in this library is b/c the documentation for `sd-bus` is not the best.  I don't have the greatest confidence as we move to different systems or different versions of systemd.

grondo:
>  Ok, wait a second. We've already agreed that libsdprocess is an internal lib to be almost exclusively used by job-exec, so is removing logging via flux_log that important?

Yeah, you make a good point. We shouldn't band over backwards to remove internal logging directly with `flux_log()` when the only intended use case is a broker module.

I think the original comment to remove logging was only suggested because it would improve the interface a bit and remove function parameters that are only used to enable logging.

You also make a _very_ good point about debugging. I have noticed myself adding lots of temporary logging to sdexec while working on that branch. It would be nice to have trace-level logs in the libsdprocess code that could be enabled at runtime if necessary. (I could be wrong, but I think `flux_log` always forwards all log messages to the broker), with a different internal log). At this point, this could be added in a future PR, and actually `llog` might be a good candidate for this because the client could drop any messages locally instead of forwarding them along to be dropped later. `llog` also has an `llog_debug()` function that is compiled out of the code when `LLOG_DEBUG` is not defined, which would allow even more verbose logging as a compile-time option.

It is not difficult to wire `llog` into a library. For `libsdprocess` what I would do is add the logging function to `flux_sdprocess_t` as

```
    sdprocess_log_f llog;
    void *llog_data;
```

Then have a `flux_sdprocess_set_log (sdp, my_logger, data);` to set a logging handler per sdprocess. No need for a global context.

If the logger isn't set, then no logging is done.

However, as I said, this could be done another day as your argument about it not being as important for `libsdprocess` has merit.

chu11:
> It is not difficult to wire llog into a library. For libsdprocess what I would do is add the logging function to flux_sdprocess_t as

Understood, the issue was not wiring it in, the issue was:

```
sdp = sdprocess_exec (h, unitname, ...);

sdprocess_set_error_handler (sdp, error_cb, arg);
```

a significant, if not majority, of potential error cases is in the `sdprocess_exec()` path, those errors wouldn't be logged under this scenario.  Thus the proposal should we do:

```
sdp = sdprocess_context_handler (h, unitname, ...);

sdprocess_set_error_handler (sdp, error_cb, arg); // or combine into context handler

sdprocess_exec (sdp);
```

or do a global initializer.

grondo:
> a significant, if not majority, of potential error cases is in the sdprocess_exec() path, those errors wouldn't be logged under this scenario. Thus the proposal should we do:

Got it. Good point! I would vote for a global log handler, if we implement anything at all at this point..

chu11:
Just re-pushed.  Per discussion above (and discounted "stupid" fixes)

- refactor the set properties function to have fewer "set errno and log" repeated calls.
- don't install `sdprocess.h` header file
- remove `flux_` prefix in all functions
- support a `sdprocess_logging()` helper function to enable/disable flux broker logging.  I made this a simple private function just for unit testing instead of setting up alternate logging functions.
- remove the "local_" functions
- update tests to use `loopback_create()` and disable flux broker logging so everything works

all the changes are in fixups, which I added minor comments so there's an idea what each fixup does.  Maybe should just squash since its a lot at this point.

grondo:
Nice set of improvements. I haven't reviewed fully yet, but it is fine with me if you want to squash all the fixups.

chu11:
> Nice set of improvements. I haven't reviewed fully yet, but it is fine with me if you want to squash all the fixups.

Re-pushed, squashed all of the fixups

grondo (approved):
Thanks for all the work @chu11! I think this is ready to go in.

grondo:
The pylint build had a spurious error (network error with pip install). That might prevent mergify from attempting a rebase.

PR #3865, opened 2021-09-16
job-manager: release after:JOBID dependencies after "start" instead of "alloc" event

grondo:
This PR builds on top of #3863 and changes the behavior of the dependency-after plugin to release `after:JOBID` jobs after the `start` event instead of on transition to `RUN` state (i.e. after the `alloc` event). This will allow for better synchronization of dependent jobs when there may be delay between job allocation and job start (e.g. for service jobs or k8s containers, etc)

Still WIP because I need to work on testing. However, putting it up now for early feedback.

### DIFF
--- a/src/modules/job-manager/plugins/dependency-after.c
+++ b/src/modules/job-manager/plugins/dependency-after.c
@@ -311,9 +311,7 @@ static int dependency_after_cb (flux_plugin_t *p,
      *   lists below
      */
     if (type == AFTER_START
-        && (target_state == FLUX_JOB_STATE_RUN
-            || target_state == FLUX_JOB_STATE_CLEANUP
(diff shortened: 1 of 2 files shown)


grondo:
Rebased on #3863 without any changes to code.

However, this version adds a test which makes use of job-exec override to ensure `after:JOBID` jobs aren't released until after the `start` event, not the `alloc` event.

garlick (approved):
Looks good!

I guess with the infrastructure you've added here, it would be trivial to do something like add a way for a "service job" to emit a specialized "service ready" event, e.g. once it's bound to a service endpoint, and build a simple service job dependency scheme around that.

grondo:
Rebased after #3863. Set MWP

PR #3586, opened 2021-04-06
libutil: add minor optimizations to aux container

garlick:
The aux container used to implement `_aux_get()` and `_aux_set()` in several Flux API classes was implicated in a performance issue described in #3583.  This PR implements two optimizations affecting the linear search of `aux_get()`: 1) allocate the aux item key with the item struct, to reduce cache overhead, and 2) store anonymous entries at the end of the list and stop searching once the first anonymous entry is encountered.

The first change may or may not have an effect but it seems like good citizenry to to collapse two mallocs into one anyway.  The second change seemed obvious after seeing @chu11's fix in #3585, where a large number of anonymous entries in a `flux_t` handle were slowing down access to keyed items.

PR #3867, opened 2021-09-18
mpi: add -ompi=openmpi@5 support

garlick:
Problem: upstream openmpi nixed our flux plugins in the v5 series, as mentioned in #3586, but our `openmpi.lua` plugin sets the environment to require those plugins, unconditionally.  This prevents flux from bootstrapping openmpi-v5, even with a pmix shell plugin.

Use the new mpi personality version support to change the logic if `-ompi=openmpi@5` is supplied by the user.  Otherwise, behavior does not change.

### DIFF
--- a/src/shell/lua.d/openmpi.lua
+++ b/src/shell/lua.d/openmpi.lua
@@ -8,8 +8,15 @@
 -- SPDX-License-Identifier: LGPL-3.0
 -------------------------------------------------------------
 
-shell.setenv ("OMPI_MCA_pmix", "flux")
-shell.setenv ("OMPI_MCA_schizo", "flux")
+local mpi, version = shell.getopt_with_version ("mpi")
+
+if mpi == "openmpi" and version and tonumber (version) >= 5 then
+    shell.setenv ("OMPI_MCA_pmix", "^flux")
+    shell.setenv ("OMPI_MCA_schizo", "^flux")
+else
+    shell.setenv ("OMPI_MCA_pmix", "flux")
+    shell.setenv ("OMPI_MCA_schizo", "flux")
+end
 
 -- OpenMPI needs a job-unique directory for vader shmem paths, otherwise
 -- multiple jobs per node may conflict (see flux-framework/flux-core#3649).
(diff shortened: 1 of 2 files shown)


src/shell/lua.d/openmpi.lua
```diff
@@ -8,8 +8,15 @@
...
-shell.setenv ("OMPI_MCA_schizo", "flux")
+local mpi, version = shell.getopt_with_version ("mpi")
+
+if mpi == "openmpi" and version == "5" then
+    shell.setenv ("OMPI_MCA_pmix", "^flux")
+    shell.setenv ("OMPI_MCA_schizo", "^flux")
+else
+    shell.setenv ("OMPI_MCA_pmix", "flux")
+    shell.setenv ("OMPI_MCA_schizo", "flux")
+end
```
grondo, line 19: Do you know if this env var has to be set to `^flux` explicitly for openmpi v5+, or can it be set to nothing at all?

grondo: Also, should the test be for `version >= 5`?

grondo (approved):
Seems good to me.

Going forward, I wonder how to best allow version specific extensions to the built-in mpi "personality" plugins provided by flux-core? One idea is that add-on plugins could be loaded in deterministic order such that the last version-specific plugin is loaded last, and thus it could override any settings made automatically by the more generic personality plugins.

Another idea would be to treat the shell "mpi" option specially, and after all plugins have been loaded, attempt to load `"<mpi>/<version>.lua"` if an mpi option is specified with a version. Then a openmpi v5 specific plugin could go in `$rc_path/openmpi/5.lua`?

Sorry for thinking out loud in this PR. If this gets things moving forward, I'm fine with merging it now and working on a more generic solution later.

src/shell/lua.d/openmpi.lua
garlick: > Do you know if this env var has to be set to ^flux explicitly for openmpi v5+, or can it be set to nothing at all?

pmix v5 doesn't have those plugins, but I thought maybe it would be OK to have "try this option" in our back pocket for opempi  v4 when it's not working.

> Also, should the test be for version >= 5?

OK, that makes sense.  The flux mca plugins aren't coming back, I'm sure!

garlick:
I'll set mwp, thanks!

No deep thoughts from me on how to structure these mpi personality plugins, sorry.

Issue #2924, opened 2020-04-30
idea: exec system bypass


grondo:
Summarizing a discussion on the call this AM:

It would be useful in a few cases to allow a job to be submitted and scheduled, but select to bypass the execution system. In this case, the submitter would be taking over the exec system protocol in some to-be-determined way.

This would be useful in providing an allocate-only interface for on-node microtasking workloads (like `make -j`), in which running a task or tasks through job-shell may be unnecessary (and unwanted) overhead.

@SteVwonder noted that this might also work for the simulator, so that at least it wouldn't have to unload the existing exec module.

@garlick made a great suggestion that a disconnect handler could generate a `release` event when processes that allocate in this manner exit.

grondo:
Another use case came up for an exec bypass feature:

On the :coffee: call, @SteVwonder was describing the k8s/flux integration where an exec-bypass feature would be useful. In this project, the execution of the "job" is being handed of to Kubernetes. Currently, in place of real work, the k8s integration runs a `sleep inf` job per k8s job. It would be much better if the k8s plugin/service could takeover exec system functionality. This would have the benefit of allowing the k8s integration to:

 1. delay the "start" event until the kubernetes workload(?) has actually started
 2. emit a finish event when the k8s workload is destructed
 3. avoid the accumulation of sleep jobs on rank 0

@SteVwonder, let me know if I made any incorrect statements here.

It may be pretty easy to implement an exec-bypass flag for jobs as was done for alloc-bypass.

grondo:
This feature was nominally provided by #3868. I'm going to close this for now. We can open a new issue when we have more concrete use cases that are not satisfied by the existing implementaiton.

PR #3868, opened 2021-09-20
job-exec: enable manual override option for mock execution jobs

grondo:
Testing of future work that uses internal job events of the `RUN` state for synchronization may require better control of how the `start` and `finish` events are emitted for test jobs. 

This PR extends the job-exec module's `testexec` implementation to add an `override` option. When this option is enabled via `attributes.system.exec.test.override=1` in jobspec, the module does not immediately issue the `start` response to the job manager, but instead waits for a manual request from the job owner via a new `job-exec.override` service. If the job does not have a `duration` or `attributes.system.exec.test.run_duration` set (or they are `0.`) then the job will continue to "run" until a manual `finish` event is requested, or a job exception raised.

A new plumbing command `flux job-exec-override` is provided to ease the use of this new interface.

This will allow, for instance, a test that ensures a dependency waiting on the `start` event is not released after the `alloc` event (and thus `RUN` state), and instead waits explicitly for the `start` event.

Also, it occurs to me that, though not ideal, this interface could be used as a stopgap solution for #2924 (exec system bypass). E.g. a driver script could request resources with an `exec.test.override` jobspec, wait for the alloc event, execute work outside of the Flux exec system and manually issue a `start` event when the work has started, then when the work is complete, manually issue a `finish` event with the correct status, effectively bypassing the normal execution system.

This could also be useful for launching k8s containers using Flux alloc-bypass jobs. The container can be associated with an alloc-bypass, job-exec override job so that the container takes no resources from the instance. The `start` and `finish` events can be manually issued when the container is started and stopped, and other Flux jobs can set `after` and `afterany` dependencies on these container jobs. (Once #3865 goes in)

### DIFF
--- a/src/modules/job-exec/testexec.c
+++ b/src/modules/job-exec/testexec.c
@@ -28,6 +28,10 @@
  *   "wait_status":i        - report this value as status in the "finish" resp
  *   "mock_exception":s     - mock an exception during this phase of job
  *                             execution (currently "init" and "run")
+ *   "override":i           - exec override mode: wait for RPC to emit start
+ *                             event a testexec job. If job has unlimited
+ *                             duration, then also wait for finish RPC or
+ *                             job exception for job finish event.
  * }
  *
  */
@@ -37,26 +41,43 @@
 #endif
 
 #include "src/common/libutil/fsd.h"
+#include "src/common/libjob/job_hash.h"
+#include "src/common/libczmqcontainers/czmq_containers.h"
 #include "job-exec.h"
+
+struct testexec_ctx {
+    flux_t *h;
+    flux_msg_handler_t *mh;  /* testexec RPC handler */
+    flux_msg_handler_t *dh;  /* disconnect handler */
+    zhashx_t *jobs;
+};
 
 struct testconf {
     bool                  enabled;          /* test execution enabled       */
+    int                   override;         /* wait for RPC for start event */
     double                run_duration;     /* duration of fake job in sec  */
     int                   wait_status;      /* reported status for "finish" */
     const char *          mock_exception;   /* fake excetion at this site   */
 };
 
 struct testexec {
+    struct jobinfo *job;
     struct testconf conf;
     struct idset *ranks;
     flux_watcher_t *timer;
 };
 
-static struct testexec * testexec_create (struct testconf conf)
+
+static struct testexec_ctx *testexec_ctx = NULL;
+
+
+static struct testexec * testexec_create (struct jobinfo *job,
+                                          struct testconf conf)
 {
     struct testexec *te = calloc (1, sizeof (*te));
     if (te == NULL)
         return NULL;
+    te->job = job;
     te->conf = conf;
     return (te);
 }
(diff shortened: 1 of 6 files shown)


grondo:
Oh, a simple example:
```
ƒ(s=1,d=0,builddir) grondo@asp:~$ flux mini submit --setattr=system.exec.test.override=1 true
ƒnL5fjYc7
ƒ(s=1,d=0,builddir) grondo@asp:~$ flux job eventlog ƒnL5fjYc7
1632153515.226417 submit userid=1000 urgency=16 flags=0
1632153515.239306 depend
1632153515.239354 priority priority=16
1632153515.241314 alloc annotations={"sched":{"resource_summary":"rank0/core0"}}
```

*Commence idle whistling*

```
ƒ(s=1,d=0,builddir) grondo@asp:~$ flux job-exec-override start ƒnL5fjYc7
ƒ(s=1,d=0,builddir) grondo@asp:~$ flux job eventlog ƒnL5fjYc7
1632153515.226417 submit userid=1000 urgency=16 flags=0
1632153515.239306 depend
1632153515.239354 priority priority=16
1632153515.241314 alloc annotations={"sched":{"resource_summary":"rank0/core0"}}
1632153538.568192 start
ƒ(s=1,d=0,builddir) grondo@asp:~$ flux job-exec-override finish ƒnL5fjYc7 0
ƒ(s=1,d=0,builddir) grondo@asp:~$ flux job eventlog ƒnL5fjYc7
1632153515.226417 submit userid=1000 urgency=16 flags=0
1632153515.239306 depend
1632153515.239354 priority priority=16
1632153515.241314 alloc annotations={"sched":{"resource_summary":"rank0/core0"}}
1632153538.568192 start
1632153552.467833 finish status=0
1632153552.470122 release ranks="all" final=true
1632153552.470923 free
1632153552.470975 clean
```

garlick:
This sounds great - thanks for the super clear explanation and the k8s use case example.

grondo:
Force-pushed a testing update to handle more corner cases (and one testing error)

Apologies if anyone was in the midst of reviewing.

src/modules/job-exec/testexec.c
```diff
@@ -215,20 +245,151 @@ static int testexec_kill (struct jobinfo *job, int signum)
...
+}
+
+static void testexec_ctx_destroy (struct testexec_ctx *ctx)
+{
+    if (ctx) {
+        int saved_errno = errno;
+        flux_msg_handler_destroy (ctx->mh);
+        flux_msg_handler_destroy (ctx->dh);
+        if (ctx->jobs)
+            zhashx_destroy (&ctx->jobs);
```
garlick, line 334: Test for non-null not required here.

src/modules/job-exec/testexec.c
```diff
@@ -28,6 +28,10 @@
  *   "wait_status":i        - report this value as status in the "finish" resp
  *   "mock_exception":s     - mock an exception during this phase of job
```
garlick, line 29: commit message for 6d68af3679ae515b6a242ab0c0b:
s/wheck/when/
s/reponse/response/

garlick (approved):
LGTM!  Just a couple minor nits.  Set MWP  when ready.

Issue #3866, opened 2021-09-17
Environment to load external plugins


eleon:
* Goal: Use the environment to load external plugins in Flux without having to write them into the Flux plugins directory. 
* Problem: In Spack, or using system installations of Flux, one cannot change the Flux installation. Having an environment variable or something similar to point Flux to additional plugins to load could address this problem.

grondo:
For example, the job shell will automatically load all `*.so` plugins in the default shell `pluginpath` via the default shell `initrc.lua`. For a Flux spack package, the default pluginpath will by definition not contain any external plugins since external packages are not supposed to place anything in the flux-core package root.

Ideally the spack environment exported by packages that want to extend the job shell would be able to append to the built in shell pluginpath or otherwise extend the default list of loaded plugins.

At this time, I'm not certain of the most efficacious method to do this, but it does seem like a reasonable request. I'll come back with a proposal if someone else doesn't beat me to it.

PR #3869, opened 2021-09-20
allow default shell rc path to be extended via FLUX_SHELL_RC_PATH

grondo:
This is an attempt at a simple solution for #3866.

The default `initrc.lua` for the job shell loads all `*.lua` rc scripts from `${rcpath}/lua.d/*`, where `rcpath` is the directory in which `initrc.lua` was found. On a normal system, the default shell initialization can be influenced by dropping another Lua rc script in this directory. (This is how the mpi 'personality' plugins work).

However, in systems like spack, where every package has its own root, add-on packages like mpibind can't just drop a new file into the shell `rcpath`, since that path is in the flux-core package's root. This PR adds support for a new environment variable `FLUX_SHELL_RC_PATH` which, when set in a job's environment, will extend the search path for `*.lua` files to automatically load at runtime.

Therefore, in spack, packages can append a path to a directory where they keep a Lua rc file which can then load plugins, further change environment, etc.

For example, the mpibind spack package, which wants to always load the `mpibind.so` shell plugin, can install a one-liner `mpibind.lua` like:
```Lua
plugin.load ("/full/path/to/mpibind.so")
```
and in the environment manipulation for the spack package, extend the environment to append the directory in which this Lua file is installed to `FLUX_SHELL_RC_PATH`.

On a non-spack system, the `mpibind.lua` file can just be installed directly into the default rcpath, e.g. `/etc/flux/shell/lua.d`.

### DIFF
--- a/src/shell/initrc.lua
+++ b/src/shell/initrc.lua
@@ -3,7 +3,11 @@
 plugin.load { file = "*.so", conf = {} }
 
 -- Source all rc files under shell.rcpath/lua.d/*.lua of shell rcpath:
-source (shell.rcpath .. "/lua.d/*.lua")
+local rcpath = shell.rcpath.."/lua.d" ..
+               ":" .. (shell.getenv("FLUX_SHELL_RC_PATH") or "")
+for path in rcpath:gmatch ("[^:]+") do
+    source (path .. "/*.lua")
+end
 
 -- If userrc is set in shell options, then load the user supplied initrc here:
 if shell.options.userrc then source (shell.options.userrc) end
--- a/t/t2603-job-shell-initrc.t
+++ b/t/t2603-job-shell-initrc.t
@@ -42,6 +42,15 @@ test_expect_success 'flux-shell: initrc: plugin.searchpath set via broker attr' 
 	flux setattr conf.shell_pluginpath "${old_pluginpath}"
 	
 '
+test_expect_success 'flux-shell: default initrc obeys FLUX_SHELL_RC_PATH' '
+	mkdir test-dir.d &&
+	cat >test-dir.d/test.lua <<-EOF &&
+	shell.log ("plugin loaded from test-dir.d")
+	EOF
+	FLUX_SHELL_RC_PATH=$(pwd)/test-dir.d \
+	  flux mini run hostname >rcpath.log 2>&1 &&
+	grep "plugin loaded from test-dir.d" rcpath.log
+'
 test_expect_success 'flux-shell: initrc: generate 1-task jobspec and matching R' '
 	flux jobspec srun -N1 -n1 echo Hi >j1 &&
 	cat >R1 <<-EOT
--- a/t/t2603-job-shell-initrc.t
+++ b/t/t2603-job-shell-initrc.t
@@ -4,6 +4,8 @@ test_description='Test flux-shell initrc.lua implementation'
 
 . `dirname $0`/sharness.sh
 
+test_under_flux 1
+
 FLUX_SHELL="${FLUX_BUILD_DIR}/src/shell/flux-shell"
 
 INITRC_TESTDIR="${SHARNESS_TEST_SRCDIR}/shell/initrc"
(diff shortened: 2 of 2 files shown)


garlick:
That seems like it would work. 

So shell plugins supplied by external packages would supply a `packagename.lua` script.

The flux pmix  plugin should probably be enabled/disabled by the mpi personalities.  Is there a way for  `openmpi.lua` and `spectrum.lua` to set an option that a `flux-pmix.lua` script could test?

grondo:
> So shell plugins supplied by external packages would supply a packagename.lua script.

If you want your shell plugin always loaded, the `packagename.lua` is optional if you are not in an environment like spack - you can just install the `.so` to the default `fluxshellpluginpath`. Otherwise, yes an external package would install a `package.lua` which could then optionally or always load the plugin.

> The flux pmix plugin should probably be enabled/disabled by the mpi personalities. Is there a way for openmpi.lua and spectrum.lua to set an option that a flux-pmix.lua script could test?

They could, though I'm not sure if it is good practice for flux-core builtin mpi personalities to assume existence of any external packages (this would require an external package developer to get some change into flux-core before their package would work (?))

Can the `flux-pmix.lua` rc script be considered an "extension" of the built-in MPI personalities, and take action independently if `shell.options.mpi` is set to some set of supported MPIs?

grondo:
It occurs to me that one issue here is that this laissez faire approach to deferring the handling of `shell.options.mpi` is that an invalid option is ignored. E.g. I can request the `awesome` MPI personality explicitly, and my plea will be silently ignored:

```
$ flux mini run -o mpi=awesome hostname
asp
```

Eventually, we may want to make some shell options first class and handled specifically in the default initrc.
E.g. If `-o mpi` is used, the shell could try to load a specific plugin from `${rcpath}/mpi/${name}.lua`, then if a version is provided, try `mpi/${name}-version.lua`. If neither of these files are found, it could be an error.

grondo:
Another thought -- since this issue really only affects the spack package of flux-core, another solution would be to modify the default shell initrc similarly in the flux-core spack package. Then no changes here would be necessary.

One benefit of this change though is that you can more easily guarantee order of rc scripts during shell startup, since the FLUX_SHELL_RC_PATH directories are traversed in order (starting with the built in one)

garlick:
> One benefit of this change though is that you can more easily guarantee order of rc scripts during shell startup, since the FLUX_SHELL_RC_PATH directories are traversed in order (starting with the built in one)

Yeah, this change seems good to me as is.

grondo:
@SteVwonder or @eleon, do you concur that this will be a good enough solution for #3866?

If I can get an ACK or review from one of you then I will set MWP.

eleon:
Thank you for the prompt solution, @grondo. This looks good to me. I will try it out with `mpibind` in Spack when I get an opportunity and report back.

grondo:
Thanks @eleon. I will set this PR to merge. Once that is done if you build flux@master in spack you should have this feature. If there are any issues, please open another GitHub issue and we'll address from there. 

Thank you!

eleon:
Will do, @grondo. Thank you!

grondo:
Oh still waiting for an officieal review from @SteVwonder or @garlick

garlick (approved):
LGTM!

grondo:
Thanks!

eleon:
@grondo , nice work! I successfully installed mpibind+flux through Spack: 
```
spack install mpibind@master+cuda+flux ^lua@5.1.4
```
```
$ flux mini run -n2 -c18 ./affinity/mpi
pascal25   Task   0/  2 runing on 9 CPUs: 0-8
           Task   0/  2 has 1 GPUs: 0x4 
pascal25   Task   1/  2 runing on 9 CPUs: 9-17
           Task   1/  2 has 1 GPUs: 0x7 
```
A few notes: 
* I needed a specific version of `lua` that matches my Slurm configuration
* The updated mpibind recipe is not upstream yet
* `mpibind` works out of the box with Flux (but can be disabled with `-o mpibind=off`)!

Question: 
* Is it safe to assume that the new environment variable is present starting with Flux version 0.29?

grondo:
> Is it safe to assume that the new environment variable is present starting with Flux version 0.29?

No, unfortunately it will only be supported with Flux > 0.29.0

eleon:
Okay. I will leave the spack recipe with `flux@master` for now and change to '>0.29' when the Flux spack recipe is updated.

PR #3871, opened 2021-09-21
docker: add --build-arg to docker-run-checks, systemd-devel to centos8

grondo:
A couple of small docker/test updates here.

In flux-pmix we needed a `--build-arg` option to `docker-run-checks.sh`, so pull that back up here.

For systemd-based job shell execution @chu11 needs `systemd-devel` in the centos8 image, so add that.

I've already pushed the new `fluxrm/testenv:centos8` image to docker hub.


garlick (approved):
LGTM!

grondo:
Verified this worked for @chu11, so setting MWP.

PR #3872, opened 2021-09-23
systemd: start flux systemd user service

chu11:
Problem: Some future flux services may wish to run processes
under systemd. This would require a flux user service to be
setup under systemd.

Solution: Update flux service file to setup flux user service
when the broker is started.  Add appropriate environment
variables for user service to environment before launching flux broker.

### DIFF
--- a/etc/flux.service.in
+++ b/etc/flux.service.in
@@ -5,7 +5,10 @@ Wants=munge.service
 [Service]
 TimeoutStopSec=90
 KillMode=mixed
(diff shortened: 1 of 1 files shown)


grondo:
@chu11, hope you don't mind, I rebased, fixed the conflict and force-pushed the result to your branch so we can consider getting this PR merged.

garlick (approved):
I'm new to systemd user instances but I know you've tested this, and it looks fine to me.  As discussed, let's get this in our next release so we can do some testing on fluke.

PR #3873, opened 2021-09-23
shell: add taskids idset to flux_shell_get_rank_info(3)

grondo:
As described in #3820, a couple shell plugins have needed to gather the list of taskids assigned to shell ranks in a job. Currently, plugins can assume "block" allocation of task ids and recreate the mapping themselves, but not only is this cumbersome, it will be wrong if the shell ever chooses a different mapping for task ids across shell ranks.

This PR adds a new member of the `flux_shell_get_rank_info(3)` payload - a `taskids` entry which is an RFC 22 idset string of task ids the shell has assigned to the target shell rank.

### DIFF
--- a/doc/man3/flux_shell_get_info.rst
+++ b/doc/man3/flux_shell_get_info.rst
@@ -60,7 +60,14 @@ string with the following layout:
 
    "broker_rank":i,
    "ntasks":i
+   "taskids":s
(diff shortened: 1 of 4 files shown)


garlick (approved):
Perfect!  Thanks so much.

grondo:
Thanks, setting MWP.

PR #3875, opened 2021-09-24
Fix a couple build and test issues found on ppc64le with clang 6.0+

grondo:
This PR fixes a couple issues that popped up when building on a ppc64le TOSS system with clang 11.0.x.

The first issue is a build problem discussed in #3874.

The second issue isn't ppc64le/clang specific, but fixes a test error when the Python interpreter used by `flux python` is a symlink.

### DIFF
--- a/src/common/libccan/ccan/base64/base64.c
+++ b/src/common/libccan/ccan/base64/base64.c
@@ -31,7 +31,7 @@ static int8_t sixbit_from_b64(const base64_maps_t *maps,
 	int8_t ret;
 
 	ret = maps->decode_map[(unsigned char)b64letter];
-	if (ret == (char)0xff) {
+	if (ret == (int8_t)0xff) {
 		errno = EDOM;
 		return -1;
 	}
@@ -41,7 +41,7 @@ static int8_t sixbit_from_b64(const base64_maps_t *maps,
 
 bool base64_char_in_alphabet(const base64_maps_t *maps, const char b64char)
 {
-	return (maps->decode_map[(const unsigned char)b64char] != (char)0xff);
+	return (maps->decode_map[(const unsigned char)b64char] != (int8_t)0xff);
 }
 
 void base64_init_maps(base64_maps_t *dest, const char src[64])
--- a/t/t0001-basic.t
+++ b/t/t0001-basic.t
@@ -61,7 +61,8 @@ test_expect_success 'flux-python command runs a python that finds flux' '
 
 test_expect_success 'flux-python command runs the configured python' '
 	define_line=$(grep "^#define PYTHON_INTERPRETER" ${FLUX_BUILD_DIR}/config/config.h) &&
-	expected=$(echo ${define_line} | sed -E '\''s~.*define PYTHON_INTERPRETER "(.*)"~\1~'\'') &&
+	pypath=$(echo ${define_line} | sed -E '\''s~.*define PYTHON_INTERPRETER "(.*)"~\1~'\'') &&
+	expected=$(${pypath} -c "import sys; print(sys.executable)") &&
 	actual=$(flux python -c "import sys; print(sys.executable)") &&
 	test "${expected}" = "${actual}"
 '


autogen.sh
```diff
@@ -5,7 +5,7 @@
 # macros are not in default aclocal search path.
 #
 echo "Running libtoolize --automake --copy ... "
-libtoolize --automake --copy || exit
+libtoolize --force --automake --copy || exit
```
chu11, line 8: did these libtoolize changes accidentally get put in?

grondo: Yes, must have accidentally done `git commit -a`. I'll fix. Thanks!

src/common/libccan/ccan/base64/base64.c
```diff
@@ -31,7 +31,7 @@ static int8_t sixbit_from_b64(const base64_maps_t *maps,
 	int8_t ret;
 
 	ret = maps->decode_map[(unsigned char)b64letter];
-	if (ret == (char)0xff) {
+	if (ret == (int8_t)0xff) {
```
chu11, line 34: i was initially confused why this fix even worked, b/c `int8_t` is typedefed as `signed char` in stdint.h.  Then I realized `decode_map` is a `signed char decode_map[256];`.

Would it be worthwhile to cast to `signed char` just for consistency's sake?

autogen.sh
grondo: This should be fixed now

src/common/libccan/ccan/base64/base64.c
grondo: We're comparing to an `int8_t` so it seems like the most robust solution is to cast to the object of comparison?

I have to admit I don't follow your reasoning here. I think I remember reading once that `char` is not guaranteed to be exactly 8 bits. My guess is _that_ is the source of the complaint. Casting `0xff` to `char` isn't guaranteed to be "8 bits, all 1s" which is what I took as the intent of this and the other use of `(char)0xff`.

grondo: > (i.e. signed char and char should be the same?)

So I learned something new today. The `char` type is special in that it is not `signed` or `unsigned` and is compatible with neither. So I guess `signed char` and `char` are _not_ the same (and I guess that is specifically the case on ppc64le?)

https://stackoverflow.com/questions/2054939/is-char-signed-or-unsigned-by-default/2054941#2054941

For example the following test program:
```c
#include <limits.h>
#include <stdio.h>
int main (int ac, char **av)
{
    printf ("CHAR_MIN = %d\n", CHAR_MIN);
}
```
returns this on ppc64le:

```console
$ ./test
CHAR_MIN = 0
```

and this on x86:
```console
$ ./test
CHAR_MIN = -128
```
Therefore I think your suggestion to use `signed char` works, but I still prefer to cast the `0xff` to the type being compared.

dongahn (approved):

src/common/libccan/ccan/base64/base64.c
chu11: oh i'm sorry, my comment must have been super confusing b/c I assumed base64 was using "signed char" vs "int8_t" consistently.  But I don't think they are.  

In this one we are comparing to `int8_t` (i.e. `int8_t ret;`).  But in the other change:

```maps->decode_map[(const unsigned char)b64char] != (int8_t)0xff```

decode_map is a `signed char`, so I thought casting to `signed char` instead of `int8_t` would be better just for consistency on this line.

Soooo I guess it dosn't really matter b/c it appears base64 code is already inconsistent about it :-)

Comment (grondo) [reply src/common/libccan/ccan/base64/base64.c:34]: > int8_t is typedefed as signed char in stdint.h.

BTW, as of C99 (I think) these types are part of the standard (not defined to `signed char`)

https://en.cppreference.com/w/c/types/integer

On ppc64le headers, at least, `signed char` is defined in terms of `int_least8_t`:

```c
/* Signed.  */
typedef signed char     int_least8_t;
```

Which is defined as the "smallest signed integer type of at least width 8" :shrug:

grondo:
@chu11, if you approve of my reasoning for use of `int8_t` then I will set MWP.

chu11 (approved):

chu11:
approved, sorry. my comments ended up just being a distraction :P

grondo:
> sorry. my comments ended up just being a distraction :P

No I didn't really understand why the "fix" worked as well, so at least I got a little educated. (Still not 100% certain I understand though)

Issue #3877, opened 2021-09-25
Cannot remove a module when it returned with flux_module_set_running 


SteVwonder:
@FrankD412 and @bhatiaharsh encountered a hang in Flux during bootstrap on Summit.  They use Spack to build/install/load Flux, and the flux-sched Spack package sets `FLUX_RC_EXTRA` so that Fluxion is loaded by default when the Spack `flux-sched` package is loaded.

Unfortunately, due to a double loading bug, the `flux-sched` package is loaded twice and the `FLUX_RC_EXTRA` points to the flux-sched rc directory twice.  This leads to the flux-sched rc scripts being run twice on setup.  On the second run through, a hang occurs.  Seemingly when the resource module gets reloaded the second time (this time with qmanager also loaded).

When reproduced on a single node (either a docker container on a laptop or a single node on Summit), only the following error occurs during bootstrap:

```
2021-09-24T22:50:15.506817Z sched-fluxion-qmanager.err[0]: update_on_resource_response: exiting due to sched-fluxion-resource.notify failure: Operation canceled
```

When run on multiple nodes, a hang occurs.

Will update this issue with additional details as we collect them.

grondo:
I think the problem is not that you "cannot remove a module when it has used `flux_module_set_running()`", but rather that a module cannot be removed before it has entered the reactor so that the `<modname>.shutdown` request from the broker can be processed.

The `flux_module_set_running()` call is provided so that a module can allow `flux module load` to return before the reactor is entered, however, the module should still enter the reactor at some point so that it can handle the builtin module services `.stats.{get,clear}` `.shutdown` `.debug` and `.ping`. I guess I'm still wondering why `sched-fluxion-resource` never enters the reactor in this case.

I think a good long term solution would be a solution to #2010 which would allow asynchronous initialization when necessary. In the case of `sched-fluxion-resource` though, initialization is "released" early so is there some reason the module can't enter the reactor instead and perform asynchronous initialization?

Even so, I wonder why `sched-fluxion-resource` is permanently blocked in a synchronous call outside the flux reactor during initialization? Shouldn't this eventually finish, the module enter the reactor, then the unload succeed?

>  But perhaps what we can do is to add a simple check into flux-core's rc1 script: for example, if FLUX_RC_EXTRA points to the base rc directory, make it a NOOP or similar.

[...]

PR #3878, opened 2021-09-27
remove duplicate directories from FLUX_RC_EXTRA, FLUX_SHELL_RC_PATH

grondo:
As described in #3877, spack induced duplicate entries in FLUX_RC_EXTRA can cause difficult to diagnose problems, since rc scripts in these directories are run twice with unpredictable results. Other, colon separated PATH-like variables may suffer a similar issue.

It is easy for the `flux(1)` command driver to deduplicate colon-separated paths, so add calls to `environment_from_env()` for any variables of this form including `FLUX_RC_EXTRA` and `FLUX_SHELL_RC_PATH` (the only two I could find at this time)

Enhance `environment_from_env()` to take a `NULL` default so that if either of these environment variables are unset, they stay unset.


src/cmd/flux.c
```diff
@@ -208,6 +208,12 @@ int main (int argc, char *argv[])
                           "FLUX_PMI_LIBRARY_PATH",
                           flux_conf_builtin_get ("pmi_library_path", flags), 0);
 
+    /*  De-duplicate any other FLUX_*_PATH environment variables by
```
garlick, line 211: Commit message: inconsistent use of _deduplicate_ vs _de-duplicate_.

garlick (approved):
Nice!

Just a nit picky english comment which you can safely ignore!

src/cmd/flux.c
grondo: Oh, good catch. That would have annoyed me. Deduplicate is a perfectly fine english word without the dash so I will remove it.

dongahn (approved):
LGTM! 

This will be very useful. Thanks @grondo. So that I understand `FLUX_RC_EXTRA` will be deduplicated before https://github.com/flux-framework/flux-core/blob/master/etc/rc1#L35 get executed correct?

PR #3879, opened 2021-09-28
shell: require FLUX_SHELL_PLUGIN_NAME in plugins to fix logging component discovery

grondo:
The current mechanism for discovery of shell plugin name in the logging facility only works for plugin code which is directly invoked as a result of a `plugstack_call()`. Shell plugins that log from reactor-based callbacks, therefore, do not currently have a "component" name set on their log messages, which makes output confusing and will thwart future efforts to filter log messages on component name.

Drop the "current stack" of plugins and `plugstack_get_current()` functionality which provided the current plugin and instead add a `const char *component` argument to all logging functions. To avoid changing all logging call sites, set the component to a new _required_ define `FLUX_SHELL_PLUGIN_NAME` in all the convenience macros `shell_log()` `shell_debug()` etc.

To ensure out-of-tree plugins define `FLUX_SHELL_PLUGIN_NAME`, error out in `shell.h` if this preprocessor macro is not defined.

Unfortunately this breaks the shell plugin ABI and API. We could have defined new alternate shell logging functions, but since we're still early in the game, I decided to keep things simple and just tolerate the breakage this time.

I will submit PRs to all current out-of-tree shell plugins - currently in flux-sched, flux-coral2, and mpibind. (I think flux-pmix will already be handled)

Meanwhile, someone ensure I'm not going off the deep end here and stop me if so.

### DIFF
--- a/doc/man3/flux_shell_log.rst
+++ b/doc/man3/flux_shell_log.rst
@@ -13,7 +13,8 @@ SYNOPSIS
 
 ::
 
-   void flux_shell_log (int level,
+   void flux_shell_log (const char *component,
+                        int level,
                         const char *file,
                         int line,
                         const char *fmt,
@@ -21,7 +22,8 @@ SYNOPSIS
 
 ::
 
-   int flux_shell_err (const char *file,
+   int flux_shell_err (const char *component,
+                       const char *file,
                        int line,
                        int errnum,
                        const char *fmt,
@@ -29,7 +31,8 @@ SYNOPSIS
 
 ::
 
-   void flux_shell_fatal (const char *file,
+   void flux_shell_fatal (const char *component,
+                          const char *file,
                           int line,
                           int errnum,
                           int exit_code,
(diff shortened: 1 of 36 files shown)


garlick:
Seems good to me!  I just tested with flux-pmix, and my copious log messges all have a lovely prefix now.  Yay!

grondo:
Ok, we'll just need to make sure all other dependent PRs are merged first:

 * https://github.com/LLNL/mpibind/pull/25
 * https://github.com/flux-framework/flux-coral2/pull/9
 * https://github.com/flux-framework/flux-sched/pull/864

grondo:
Restarted CI now that downstream changes have been merged.

grondo:
Well I missed a required update to a test in the flux-sched PR, so a follow on PR posted :facepalm: 

 * flux-framework/flux-sched#865

grondo:
Ok, restarted CI again. This PR is probably ready for an official review now.

garlick (approved):
All looks good!  Thanks for doing this - nice improvement.

grondo:
Thanks! I've set MWP.

PR #3885, opened 2021-09-30
resource: fail get-xml request on quorum subset

### LINKED ISSUE #3884: consider failing `resource.get-xml` if broker quorum is a subset of the instance
This was brought up as an idea in flux-framework/flux-sched#866, where a system instance's configured R was missing JGF augmentation, and the system deadlocked, blocked on this RPC.  A proposed solution is to fail the RPC with a nice human readable error message if the `broker.quorum` is set to a subset of the instance, since that could mean that we need to  begin operations with some nodes down, and the `get-xml` RPC could block indefinitely.

garlick:
Problem: if R is missing augmented graph information, fluxion
might block forever on resource.get-xml request for resource
discovery results.

Resource discovery is the natural fallback if resources not
not configured.  But fluxion cannot tell the difference between
R that lacks configured JGF augmentation because it was misconfigured,
and R allocated by a scheduler that doesn't support it.

If discovered hwloc results are not immediately available at the time
of the request, fail the request if the broker.quorum attribute is
less than the full instance, as this indicates that the instance
should begin operation with some ranks offline, and R should contain
all needed information.

Tack a test onto t3301-system-latestart.t

Fixes #3884.

### DIFF
--- a/src/modules/resource/inventory.c
+++ b/src/modules/resource/inventory.c
@@ -470,6 +470,31 @@ done:
     return rc;
 }
 
+static int check_broker_quorum (flux_t *h, bool *is_subset)
+{
+    const char *quorum;
+    uint32_t size;
+    char buf[32] = "0";
+
+    if (!(quorum = flux_attr_get (h, "broker.quorum"))
+        || flux_get_size (h, &size) < 0)
+        return -1;
+    if (size > 1) {
+        if (snprintf (buf,
+                      sizeof (buf),
+                      "0-%d",
+                      (int)size - 1) >= sizeof (buf)) {
+            errno = EOVERFLOW;
+            return -1;
+        }
+    }
+    if (strcmp (buf, quorum) == 0)
+        *is_subset = false;
+    else
+        *is_subset = true;
+    return 0;
+}
+
 /* If xml collection is still in progress, this function delays responding
  * until it completes.
  */
@@ -489,7 +514,20 @@ static void resource_get_xml (flux_t *h,
         errstr = "this RPC only works on rank 0";
         goto error;
     }
+    /* xml is still being collected.  If broker.quorum is not the entire
+     * instance, then this may never complete, so fail the request in that
+     * case; otherwise save the request in inv->waiters for later response.
+     */
     if (!inv->xml) {
+        bool is_subset;
+        if (check_broker_quorum (h, &is_subset) < 0)
+            goto error;
+        if (is_subset) {
+            errno = EINVAL;
+            errstr = "Request may block forever. "
+                     "If R is incomplete, there may be a config error.";
+            goto error;
+        }
         if (zlist_append (inv->waiters, (void *)flux_msg_incref (msg)) < 0) {
             flux_msg_decref (msg);
             errno = ENOMEM;
--- a/t/t3301-system-latestart.t
+++ b/t/t3301-system-latestart.t
@@ -17,6 +17,10 @@ test_under_flux 2 system
 
 startctl="flux python ${SHARNESS_TEST_SRCDIR}/scripts/startctl.py"
 
+get_hwloc () {
+        flux python -c "import flux; print(flux.Flux().rpc(\"resource.get-xml\",nodeid=$1).get_str())"
+}
+
 test_expect_success 'broker.quorum was set to 0 by system test personality' '
 	echo 0 >quorum.exp &&
 	flux getattr broker.quorum >quorum.out &&
@@ -53,6 +57,12 @@ test_expect_success 'single node job can run with only rank 0 up' '
 
 test_expect_success 'two node job is accepted although it cannot run yet' '
 	flux mini submit -N2 -n2 echo Hello >jobid
+'
+
+# bug 3884
+test_expect_success 'resource.get-xml RPC fails with reasonable error' '
+	test_must_fail get_hwloc 0 2>get_hwloc.err &&
+	grep "may block forever" get_hwloc.err
 '
 
 test_expect_success 'start rank 1' '


garlick:
This does feel a bit hacky.

I might be forgetting the details, but wasn't the hwloc access implemented because the "Rv1 reader" code had not yet been developed in fluxion?  Could this hack, and `get-hwloc` RPC for that matter, be avoided by falling back the Rv1 reader if the JGF was not inserted into R?

It might be better to unwind some hacks rather than add more :-)

src/modules/resource/inventory.c
```diff
@@ -470,6 +470,24 @@ static int get_from_kvs (struct inventory *inv, const char *key)
...
+{
+    const char *quorum;
+    uint32_t size;
+    char buf[32] = "0";
+
+    if (!(quorum = flux_attr_get (h, "broker.quorum"))
+        || flux_get_size (h, &size) < 0)
+        return false;
+    if (size > 1)
+        snprintf (buf, sizeof (buf), "0-%d", (int)size - 1);
```
dongahn, line 485: Super minor nit: check the return code for an error?

src/modules/resource/inventory.c
```diff
@@ -470,6 +470,24 @@ static int get_from_kvs (struct inventory *inv, const char *key)
...
+    if (strcmp (buf, quorum) == 0)
```
dongahn, line 486: If a very unlikely failure from `snprintf` above, `buf` may not be well-formed and buffer overflow could happen? `strncmp` since this is a known buffer size?

src/modules/resource/inventory.c
```diff
@@ -489,7 +507,17 @@ static void resource_get_xml (flux_t *h,
         errstr = "this RPC only works on rank 0";
         goto error;
     }
+    /* xml is still being collected.  If broker.quorum is not the entire
+     * instance, then this may never complete, so fail the request in that
+     * case; otherwise save the request in inv->waiters for later response.
+     */
     if (!inv->xml) {
+        if (quorum_subset (h)) {
+            errno = EINVAL;
```
dongahn, line 516: `EWOULDBLOCK` or `EAGAIN`? Since it is not like the RPC argument was invalid but the state of the flux instance cannot service the RPC?

dongahn (approved):
LGTM. Just minor nits.

dongahn:
>  might be forgetting the details, but wasn't the hwloc access implemented because the "Rv1 reader" code had not yet been developed in fluxion?

Yap. 

The Rv1 reader code is currently python as part of `flux ion-R`. But no C++ code yet. My biggest fear going down on the path of integrating on-the-fly `rv1 reader` into fluxion is that `.scheduling` key will soon need to contain much more information than what `rv1` currently contains (e.g., Rabbit multitiered storage info) and extending the `rv1 reader` feature for future use will be pretty brittle and seems a bit like a premature optimization. 

Having this seems like a minimum viable for now, we then perfect it when more use cases come to our ways in the future?

src/modules/resource/inventory.c
garlick: `snprintf()` does ensure the string is terminated even on overflow so that should not be required.

src/modules/resource/inventory.c
garlick: Given that we're returning a textual error, all we really need for errno is _something_ nonzero.  The intent was that fluxion would detect the error and log the human readable string at LOG_ERR level to leave a big clue for someone.  It shouldn't need to do anything with the errno value.

src/modules/resource/inventory.c
garlick: Its not really possible to overflow the buffer unless there were a programming error here, but maybe this function ought to have a chance to fail instead of just returning the logical results.  Let me fix that.

src/modules/resource/inventory.c
dongahn: Nice to know. thx

src/modules/resource/inventory.c
dongahn: Fluxion use case is fine. Just caught my eyes in case this RPC used by something other than fluxion in the future. Ok as is.

src/modules/resource/inventory.c
dongahn: I don't even know if I actually check the return code. But I thought I would mention in case you folks do that in `flux-core`. BTW, you do also guard some of the low level system functions from returning prematurely when being attached to a debugger?

garlick:
> My biggest fear going down on the path of integrating on-the-fly rv1 reader into fluxion is that .scheduling key will soon need to contain much more information than what rv1 currently contains (e.g., Rabbit multitiered storage info) and extending the rv1 reader feature for future use will be pretty brittle and seems a bit like a premature optimization.

I'm not sure I get why it would be brittle.  Rv1 is [well defined](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_20.html) and incorporated into the design of several components.  It shouldn't change at this point without becoming Rv2, which I think would be a different reader?

It's our only standardized abstract resource definition at this point, it handles the majority of our production systems, and it's compact and simple to parse if you have jansson, idset, and hostlist libraries avaialble.    My vote would be we go ahead and implement it.  If we did, then we would not require the extra sys admin step of augmenting R on most systems - extra steps for exotic systems only :-)

But sure we have lots of things on our plate and we have to prioritize.

dongahn:
> I'm not sure I get why it would be brittle. Rv1 is well defined and incorporated into the design of several components. It shouldn't change at this point without becoming Rv2, which I think would be a different reader?

Good point. Brittle was a bad choice of word. Maybe "agile" might have been a better choice...

When we start to test an exotic system like El Cap testbed in a few months, we will soon need a way to inject resources that are not defined by `RV1` but still develop and test things pretty quickly. An ability to inject that as an JGF into the initial R can help with this. Yes, it is true we can define Rv2 to cover those resources which we will at some point. My observation from the past has been it could take significant amounts of time to define it formally. 

Having said that, I can see us implementing on-the-fly RV1 reader and use it only when `.scheduling` key isn't defined.

garlick:
> When we start to test an exotic system like El Cap testbed in a few months, we will soon need a way to inject resources that are not defined by RV1 but still develop and test things pretty quickly. An ability to inject that as an JGF into the initial R can help with this. Yes, it is true we can define Rv2 to cover those resources which we will at some point. My observation from the past has been it could take significant amounts of time to define it formally.

I think inserting JGF in Rv1 under the opaque scheduling key is a fine stopgap for exotic systems until we have time to work on Rv2, which might be a while.  IMHO the Rv1 reader is desirable in addition to that, to keep on track with the overall design. 

There may be a startup performance benefit at scale too, because we are only gathering the bulky hwloc XML in order to support  the `get-xml` RPC for fluxion.  The "resource discovery" algo in the resource module is to gather/reduce Rv1 fragments which are much more compact than the hwloc XML.  The XML was  tacked on so that fluxion wouldn't need to implement the Rv1 reader right away.  A lot of data is being moved unnecessarily during instance bootstrap as a result.

grondo (approved):
This seems like a good stopgap for now to me.

dongahn:
@garlick: all good points and I agree. Let's plan this and using rv1 reader when we don't have `.scheduling` key seems like a good compromise. For immediate fluke testing, we should probably plan to land this PR?

garlick:
I'll go ahead and set MWP - thanks for the review and discussion!

PR #3888, opened 2021-09-30
t: fix incorrect method call in test-terminal.perl

### LINKED ISSUE #3887: test-terminal.perl fails due to method not found
The exact error is `Can't locate object method "follower" via package "IO::Pty"`.  As noted in https://github.com/chriscool/sharness/issues/105, this is because `IO::Pty` only has a `slave` method and not a `follower` method.  The uses of `follower` need to be reverted back to `slave` until `IO::Pty` changes their terminology.

SteVwonder:
Problem: The test-terminal.perl script fails with
`Can't locate object method "follower" via package "IO::Pty"`.
As noted in chriscool/sharness#105, this is because IO::Pty only has a
slave method and not a follower method.

Solution: Revert the changes in 5f2fb21 until IO::Pty changes their
terminology.

Closes #3887

### DIFF
--- a/t/test-terminal.perl
+++ b/t/test-terminal.perl
@@ -67,14 +67,14 @@ sub copy_stdio {
 if ($#ARGV < 1) {
 	die "usage: test-terminal program args";
 }
-my $leader_out = new IO::Pty;
-my $leader_err = new IO::Pty;
-$leader_out->set_raw();
-$leader_err->set_raw();
-$leader_out->follower->set_raw();
-$leader_err->follower->set_raw();
-my $pid = start_child(\@ARGV, $leader_out->follower, $leader_err->follower);
-close $leader_out->follower;
-close $leader_err->follower;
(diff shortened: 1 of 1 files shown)


grondo:
Should this instead just revert commit 5f2fb215c78b15b029533383bb4c269f2d490120?
i.e.
```
$ git revert 5f2fb215c78b15b029533383bb4c269f2d490120
```

chu11:
> Should this instead just revert commit 5f2fb21?

Agreed.  Since this is what upstream did, we should probably just copy what they did and revert until it can be changed in the future.

SteVwonder:
Thanks @grondo and @chu11!  Just force-pushed a revert commit.

grondo (approved):
LGTM. It might be nice to add an explanation to the commit message, but not necessary, so I leave that up to you.

SteVwonder:
Ok.  Just updated that commit message.  Thanks @grondo!

PR #3889, opened 2021-09-30
job-manager: small fixes for the alloc-bypass plugin

grondo:
Some small mistakes were found in the `alloc-bypass` plugin during a review of flux-framework/flux-k8s#14.

Since early on it may be common practice to start from one of the existing jobtap plugins by copying an example, we should fix the issues discovered so they aren't copied again. This work was done quickly so an extra set of eyes will be valued.

Additionally during that review it was discovered that the Python tests have no coverage of `flux_respond_error()` (which would also be a good example usage for someone to copy), so that is tacked on here even though it is not relevant.

### DIFF
--- a/src/modules/job-manager/plugins/alloc-bypass.c
+++ b/src/modules/job-manager/plugins/alloc-bypass.c
@@ -88,6 +88,7 @@ static int alloc_start (flux_plugin_t *p,
                         flux_jobid_t id,
                         const char *R)
 {
+    int saved_errno;
     flux_future_t *f = NULL;
     flux_jobid_t *idptr = NULL;
 
@@ -102,8 +103,10 @@ static int alloc_start (flux_plugin_t *p,
     *idptr = id;
     return 0;
 error:
+    saved_errno = errno;
     flux_future_destroy (f);
     free (idptr);
+    errno = saved_errno;
     return -1;
 }
 
@@ -227,11 +230,12 @@ static int validate_cb (flux_plugin_t *p,
                                     "alloc-bypass::R",
                                     s,
                                     free) < 0) {
+        int saved_errno = errno;
         free (s);
         return flux_jobtap_reject_job (p,
                                        args,
                                        "failed to capture alloc-bypass R: %s",
-                                       strerror (errno));
+                                       strerror (saved_errno));
     }
     return 0;
 }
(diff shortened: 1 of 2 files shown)


src/modules/job-manager/plugins/alloc-bypass.c
```diff
@@ -40,12 +40,14 @@ static void alloc_continuation (flux_future_t *f, void *arg)
                                      *idptr,
```
garlick, line 40: commit message s/Skipp/Skip/

src/modules/job-manager/plugins/alloc-bypass.c
```diff
@@ -61,35 +61,52 @@ static void alloc_continuation (flux_future_t *f, void *arg)
...
+    if (!(idptr = malloc (sizeof (*idptr)))
+        || flux_future_aux_set (f, "jobid", idptr, free) < 0)
+        goto error;
+
     *idptr = id;
-    flux_kvs_txn_destroy (txn);
     return 0;
+error:
+    flux_future_destroy (f);
+    free (idptr);
```
dongahn, line 106: `free` can be wrapped to save `errno`? Same comment from my review on @milroy's

dongahn (approved):
LGTM. Just one minor comment.

src/modules/job-manager/plugins/alloc-bypass.c
```diff
@@ -142,11 +142,13 @@ static int sched_cb (flux_plugin_t *p,
```
garlick, line 142: commit message s/failes/fails/

garlick (approved):
Looks good to me, just a couple commit typos.

src/modules/job-manager/plugins/alloc-bypass.c
grondo: Yeah, though `errno` is not used on return from this function (a generic error is logged in the exception without `strerror`) However, maybe it would be better to capture the errno (though probably not helpful because we won't know which function exactly got the error)

grondo: I actually found another case where `errno` was clobbered by `free`. I fixed both cases with an add-on commit.

grondo:
Thanks! I think I addressed all the feedback. Setting MWP.

PR #3890, opened 2021-10-01
shell: improve initrc extensibility, support version specific mpi plugin loading

grondo:
This PR makes some slight improvements to the shell's initrc environment.

The kludgy `lua.d/00-lib.lua` is replaced with a builtin `flux.shell` Lua module which is loaded _before_ sourcing `initrc.lua` so that functions defined in that module are available at the start of `initrc.lua` and/or if `initrc.lua` is replaced by the user. This allows the functions defined in this module to become part of the shell standard initrc interface.

Two new sourcing convenience functions are added to the `flux.shell` module:

 - `shell.source_rcpath (pattern)`: Source all files matching `pattern` from rcpath (the builtin path + `lua.d/` with `FLUX_SHELL_RC_PATH` appended)
 - `shell.source_rcpath_option (opt)`: Source all files matching any of the following in rcpath:
     - `<opt>/<value>.lua`
     - `<opt>/<value>@<version>.lua` (if `-o opt=value@version`)

Use `shell.source_rcpath_option ("mpi")` in the default shell `initrc.lua` so that specific mpi plugins can be loaded from an `mpi/` subdirectory.

Move the `spectrum.lua` plugin to `lua.d/mpi/spectrum.lua` to take advantage of these new features.

Edit: In summary, new functionality is that `-o mpi=name@version` will try to load the following files, in order, after the "default" versions provided in the top-level `lua.d/*.lua` files:

  - `<rcpath>/mpi/name.lua`
  - `<rcpath>/mpi/name@version.lua`

Where `rcpath` is the standard shell rcpath (directory of the shell `initrc.lua` + `/lua.d`) plus any `FLUX_SHELL_RC_PATH` environment variable appended.

This means that 3rd party mpi Lua plugins can be dropped into `/etc/shell/lua.d/mpi/<name>.lua` to extend the default behavior for that MPI in general, or `/etc/shell/lua.d/mpi/<name>@<version>.lua` for a specific version. Since these files are guaranteed to be loaded after the generic MPI "plugins", the version specific plugins can override settings, environment variables, etc.


garlick:
Excellent!

Doing a little testing here and seems to work as advertised!

At this point could we fail the job if the user requests an mpi personality that cannot be loaded?  For example, I could imagine `-ompi=openmpi@5` ending up on a FAQ somewhere, but that will not work if the flux-pmix package is not installed.

grondo:
> At this point could we fail the job if the user requests an mpi personality that cannot be loaded? For example, I could imagine -ompi=openmpi@5 ending up on a FAQ somewhere, but that will not work if the flux-pmix package is not installed.

We could require that a `mpi/<name>@<version>.lua` file be present if `-o mpi=name@version` is used. That would mean that there would need to be a file per version of each mpi you want to support (not sure how to do it for the generic `<name>.lua` plugins, or mpi plugins that are always loaded like mvapich and the generic openmpi plugins). Maybe the way to do this would be to drop the `@version` thing and just give the different MPIs different names, e.g. `-o mpi=openmpi` vs `-o mpi=openmpi5`. Then if an `mpi` option is received by the shell, and no `mpi/<value>.lua` exists, a fatal error is generated.

Additionally, the job obviously wouldn't fail until the shell starts, which is a bummer. If we eventually add the `flux-shell --validate` option, then maybe that could help?

We might have to rethink how to better handle these shell options which are a bit ad-hoc now.

garlick:
OK, that's fine, let's file that thought away for future consideration.  This is a good improvement IMHO.  Let me make a review pass and then my vote is in favor of this going in.

garlick (approved):
This looks good.

BTW I did confirm that a script specified by `-ouserrc` does load last, so a user can use that to override plugin-set environment variables if necessary.  (I had mentioned offline that I wasn't sure that was working last week, but that was my error).

With this in place, my earlier changes to `openmpi.lua` can be reverted.  Did you want to do that here?  That was c003f4f7cb631d8da193755a20a75037534a82dd and 7a046ad24c7ac63a41e088270fdd690e705fc666.

garlick:
Looks good, thanks!

grondo:
Ok, setting MWP.

Issue #3894, opened 2021-10-02
Flux missing stderr during an MPI_Abort


SteVwonder:
A program that writes to stderr right before calling MPI_Abort produces no program output from Flux (just some flux-shell errors).  Here is a simple reproducer:

```c
❯ cat mpi-abort.c                                                                                                                                                                                                              17:40:39 ()
#include <stdio.h>
#include <mpi.h>

main(int argc, char **argv)
{
   int node;

   MPI_Init(&argc,&argv);
   MPI_Comm_rank(MPI_COMM_WORLD, &node);

   printf("Hello World from Node %d\n",node);

   if (node == 0) {
     fprintf(stderr, "Rank 0 is going to MPI_Abort now\n");
     MPI_Abort(MPI_COMM_WORLD, 42);
   }

   MPI_Finalize();
}
```

Here is the output when run under the latest Flux release:
```
herbein1@quartz1154 ~/workspace/mpi-examples
❯ flux start flux mini run ./mpi-abort-openmpi-3.0.1-x86                                                                                                                                                                       17:39:19 ()
0.575s: flux-shell[0]: FATAL: MPI_Abort: N/A
0.575s: flux-shell[0]: stderr: flux-shell: FATAL: MPI_Abort: N/A
0.576s: job.exception type=exec severity=0 MPI_Abort: N/A
flux-job: task(s) exited with exit code 42
2021-10-02T00:39:25.454216Z broker.err[0]: rc2.0: flux mini run ./mpi-abort-openmpi-3.0.1-x86 Exited (rc=42) 1.3s

(42) herbein1@quartz1154 ~/workspace/mpi-examples
❯ which flux                                                                                                                                                                                                                   17:39:26 ()
/usr/global/tools/flux/toss_3_x86_64_ib/flux-c0.29.0-s0.18.0/bin/flux
```

Here is the output when run under slurm (note the program stderr, the MPI stderr, and even the program stdout):
```
❯ srun -N1 -n1 -ppdebug ./mpi-abort-openmpi-3.0.1-x86                                                                                                                                                                          17:39:46 ()
srun: job 7464392 queued and waiting for resources
srun: job 7464392 has been allocated resources
Rank 0 is going to MPI_Abort now
--------------------------------------------------------------------------
MPI_ABORT was invoked on rank 0 in communicator MPI_COMM_WORLD
with errorcode 42.

NOTE: invoking MPI_ABORT causes Open MPI to kill all MPI processes.
[...]

PR #3898, opened 2021-10-05
shell: avoid dropping stderr after a PMI abort

grondo:
The `shell_die ()` call currently used in the job shell's PMI abort handler causes the shell to exit immediately without any ability to process pending events or I/O in the reactor. This results in the loss of stdout/err which might occur around the time of the abort.

This PR introduces a new shell API call: `flux_shell_raise(3)` which allows a shell plugin to easily raise a job exception, and replaces the `shell_die()` call with `flux_shell_raise(3)` in the PMI plugin's abort handler. This allows the shell to reenter the reactor after the exception is raised. The exec system will then terminate the job due to the exception.

A similar test to the reproducer from #3894 is also added to the testsuite.

This changes the overall wait or finish status for the job, since there is currently no way to propagate the MPI_Abort requested exit code down through the exec system at this time. A separate issue will be created to track that.

The new `flux_shell_raise(3)` function does a synchronous get to ensure the exception was successfully created before moving on, because this is what `shell_die(3)` did. If this is a problem in other contexts, maybe the function should get a flag to run in synchronous mode, or fire-and-forget mode (as @garlick was currently doing in the pmix implementation)

Finally, the shell `doom` plugin has a similar issue (calls `shell_die()` from the `exit-on-error` handler), so there is a potential for output loss there as well. I decided not to address that here, and wait for a solution to the global exit code problem, since that plugin also expects to be able to set the exit code for the shell.

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -135,6 +135,7 @@ MAN3_FILES_SECONDARY = \
 	man3/flux_shell_setopt_pack.3 \
 	man3/flux_shell_err.3 \
 	man3/flux_shell_fatal.3 \
+	man3/flux_shell_raise.3 \
 	man3/flux_shell_log_setlevel.3 \
 	man3/flux_shell_task_info_unpack.3 \
 	man3/flux_shell_task_cmd.3 \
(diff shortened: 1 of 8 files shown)


garlick:
Just to be sure I understand - is the sequence:
- fatal exception is raised
- job-exec sends shells a SIGTERM
- SIGTERM handler kills job processes
- shell exits once children have been reaped AND stdio EOFs received

grondo:
Yes, though I think there is a 5s timeout (configurable) after which SIGKILL is sent to shells. This is, of course, just how it works now and could be redesigned as needed. I do kind of like that a sev 0 exception has some guarantee of cleaning up the job at a higher level.

garlick (approved):
This looks good to me!  Thanks.

grondo:
Thanks! I've set MWP.

Issue #3912, opened 2021-10-19
bad node(s) cause voluminous errors on fluke


grondo:
There must be one or more nodes with some kind of issue on fluke, though I haven't tracked down what it is yet. 

I can run a subinstance of Flux on 25 of the 50 nodes, but trying all 50 nodes results in some `bind` and `ZMTP` errors.

```
 grondo@fluke108:~$ flux mini alloc -N25 -n25 flux resource list
     STATE NNODES   NCORES    NGPUS NODELIST
      free     15       15        0 fluke[51-52,56,58,60,64-65,67-70,74,76-78]
 allocated      0        0        0 
      down     10       10        0 fluke[80,85-86,88,91,95,99-101,103]
[detached: session exiting]
 grondo@fluke108:~$ flux mini alloc -N50 -n50 flux resource list
2021-10-19T02:16:57.912345Z broker.err[1]: child sockevent 192.168.128.7:49152 bind failed: Address already in use
2021-10-19T02:16:57.912450Z broker.err[1]: child sockevent 192.168.128.7:49153 bind failed: Address already in use
2021-10-19T02:16:57.912541Z broker.err[1]: child sockevent 192.168.128.7:49154 bind failed: Address already in use
2021-10-19T02:16:57.912821Z broker.err[6]: child sockevent 192.168.128.13:49152 bind failed: Address already in use
2021-10-19T02:16:57.913394Z broker.err[5]: child sockevent 192.168.128.12:49152 bind failed: Address already in use
2021-10-19T02:16:57.913039Z broker.err[3]: child sockevent 192.168.128.10:49152 bind failed: Address already in use
2021-10-19T02:16:57.913017Z broker.err[4]: child sockevent 192.168.128.11:49152 bind failed: Address already in use
2021-10-19T02:16:57.949021Z broker.err[2]: child sockevent tcp://[::ffff:192.168.128.8]:49152 handshake failed protocol: ZMTP cryptographic
2021-10-19T02:16:57.988784Z broker.err[1]: child sockevent tcp://[::ffff:192.168.128.7]:49155 handshake failed protocol: ZMTP cryptographic
2021-10-19T02:16:57.995391Z broker.err[2]: child sockevent tcp://[::ffff:192.168.128.8]:49152 handshake failed protocol: ZMTP cryptographic
[...]

PR #3913, opened 2021-10-19
improve broker overlay logging and debugging capability

garlick:
A work in progress PR for some overlay debug changes, etc, while working on #3912


garlick:
Oh good idea, and thanks for asking that.  I'm realizing that the two attributes added in this PR should probably have on/off values rather than using present=on, absent=off.  That not only allows it to be immutable but to be listed with `flux lsattr`.

PR #3926, opened 2021-10-29
broker: ignore missing ldconfig when searching for libpmi.so

grondo:
Problem: Some installations of flux-core do not have ldconfig(1)
present in PATH, and this results in a senseless warning

  sh: ldconfig: command not found

Adjust the shell command used with popen(3) to first check for ldconfig
in PATH before trying to run it, thus silently ignoring the missing
command.


garlick (approved):
LGTM!  Thanks for taking care of this.

grondo:
Thanks! I've set MWP

Issue #3448, opened 2021-01-04
automatically drain unresponsive nodes


garlick:
Brokers send a keepalive message to their parent on every heartbeat, if they have not sent any other messages since the last heartbeat.  If a broker has not been heard from in some threshold number of heartbeats, it is marked "Idle" by its parent broker.  Transition to and from "idle" is only _logged_ at the moment.  No action is taken.

A flux instance should be configurable to:
* auto-drain idle nodes
* auto-undrain un-idle nodes
* set the idle threshold

Since idle detection is in the broker, and the resource module takes care of drain/undrain, we need an interface between the two, or perhaps an eventing scheme where the set of idle nodes is announced instance wide when it changes.

garlick:
Couple of thoughts on this design:

**Background**

The broker overlay tracks its direct TBON descendant peers:
- time since last message (idle time)
- subtree state (_offline_ if peer shut down nicely or hasn't yet connected; _lost_ if disconnected due to error; other states _partial_, _degraded_, and _full_ all mean the peer is online and reflect the status of the peer's descendants

An online peer can transition to _lost_ if there is an error sending a message to it, or if it is administratively disconnected with `flux overlay disconnect RANK`.  (#3897 proposes to also disconnect after idle time reaches configurable threshold.)  The subtree state is reduced at each level to a single value, so the root (rank 0) doesn't know the full set of ranks that are in each state, only the subtree state.

The resource module independently monitors nodes for purposes of including/excluding nodes in scheduling.   When a node loads or unloads the resource module it emits a message.  These messages go through a timed reduction to rank 0, and is used to genreate online/offline `resource.acquire` responses to the scheduler, taking into account other things like whether the node was administratively drained previously.  (If a node's configured resources do not match its hwloc topology, then the resource module fails to load and the message at load time is not sent).

**Design Thoughts**
* The important state information that needs to be shared from overlay to resource module is the transition to _lost_.  All other state transitions are already covered, since a node has to be online at the broker level to load the resource module, and will have unloaded the resource module before transitioning to offline at the broker level.
* The broker can also provide the subtree topology, so if a node transitioning to _lost_ is a router, it can provide affected descendant ranks
* If the resource module is going to use this info to drain nodes, it would be nice if some human readable text describing how the node(s) were lost could be provided, e.g. "disconnected due to administrative action" or "disconnected due to idle threshold".
[...]

PR #3930, opened 2021-11-02
resource: avoid scheduling on nodes that have crashed

garlick:
This PR adds a broker "groups" abstraction that replaces some duplicate idset "timed batch" reduction code in the broker state machine and in the resource module.   A module or broker service may make a `group.join` RPC to join a group, and `group.leave` to leave a group.   The rank that joins or leaves is the local broker rank, so this service is explicitly for "one member per broker" groups.  If a disconnect is received for a client that joined some groups, an automatic LEAVE is generated.  If a broker subtree transitions to  _lost_ or _offline_, automatic LEAVEs are generated for the subtree ranks.  A `group.get` RPC is available, on rank 0 only, for either streaming or non-streaming access to the current group membership.

Migrating the state machine to this server was trivial.  A `broker.online` group was added.  Brokers join this group when they enter QUORUM state.  The state machine on rank 0 runs a streaming `group.get`, and posts the `quorum-full` event once this group reaches quorum.  No action is taken when brokers leave the group.

At first I figured I would add another group for the resource module and have the module on each rank join and leave when the module is loaded/unloaded, but on second thought, I felt it was an improvement to reuse the above `broker.online` group, since that group reflects brokers that have fully completed rc1, as opposed to merely loading the resource module.

This PR does not automatically drain nodes that have crashed, as was proposed in #3448.  It is a first step that just prevents new jobs from starting on nodes that are down.  If the nodes come back up, the scheduler can immediately use them again.

I did not add anything to the API or commands to "export" access to groups, as it didn't seem all that useful to do so.  However some ideas that might be interesting for the future
- add a `--group NAME` option to `flux exec`, e.g. to only exec on "up" nodes, run `flux exec --group broker.online`
- automatically create a group per broker module to track which ranks have loaded them, and then show in `flux module list` like we had long ago
- add a group to track which nodes have exceeded responsiveness thresholds and use that to drain sad nodes
- add a group per job for tracking where the job shells are (might be challenging)

For now I felt it best to keep it simple.


src/broker/groups.c
```diff
@@ -0,0 +1,750 @@
...
+                            "error constructing groups.get response");
+        }
+        request = flux_msglist_next (group->watchers);
+    }
+}
+
+/* Process a groups.get request for group membership.  To avoid bugs arising
+ * from users making the request with nodeid=FLUX_NODEID_ANY on rank > 0,
+ * which would return that broker's subtree membership, reject such requests.
+ * If the group doesn't exist, it is created, empty.  This request may
```
grondo, line 547: Minor question: I'm curious about the decision to create groups that do not exist with a `get`. Since groups are never destroyed, does this have the potential to accumulate unneeded group objects in the broker from accidental or `gets` with typo'd group names?

Also does this mean a client can't tell the difference between a group that doesn't exist and one that is empty (or is there no difference between the two right now?)

To conserve resources at least, would it make sense to destroy empty groups on `get` client disconnect? (Maybe that is already done and I just haven't gotten that far)

grondo (approved):
This ended up being really nice cleanup! Nice work!
I couldn't find much on which to comment -- just a couple typos and one (probably naive) question inline.

src/broker/groups.c
garlick: The reason for creating a group on get and not distinguishing between empty and nonexistent was so that a streaming get request could be attached to a group before its creation.  That's also the reason why it's not removed when empty or when the joiner disconnects - there might still b a get waiting on it.

Those issues could be addressed by moving the get requests to a common list in the main groups struct.  I thought maybe it was not a huge deal given that groups are for instance owner only?  However, I may have landed in that position due to "design fatigue"!   So what do you think, should it change?

garlick:
Repushed with typo fixes and rebased on current master.

src/broker/groups.c
grondo: No need to change. I was just curious.

> That's also the reason why it's not removed when empty or when the joiner disconnects - there might still b a get waiting on it.

Can you check if a get is still waiting?

This could easily be fixed in the future if it even becomes a problem, so not a big deal at all. It does seem like you could easily deal with the (admittedly unlikely) `get` to an invalid group by removing groups that are empty when the last "getter" disconnects.

But no need to do anything now I think! We need to keep forward momentum

garlick:
@cmoussa1, were you able to walk through this in the fluxorama docker image to double check this procedure before we ask our admins to use it on fluke?

Also, was thinking you might want to mention `flux-accounting` in the first paragraph of [package installation](https://flux-framework.readthedocs.io/en/latest/adminguide.html#package-installation) where we introduce `flux-core`, `flux-sched`, and `flux-security`.

src/broker/groups.c
garlick: That's true.  OK, I'm for moving forward also, as this has taken way too long already and there is more to do in the resource module to properly handle nodes that bounce up and down or become idle.  Thanks!  I'll set MWP.

PR #3932, opened 2021-11-02
broker: fail gracefully when rundir or local-uri exceed AF_UNIX path limits

garlick:
Problem: if `local-uri` or TBON unix domain sockets exceed the max allowed path length, the instance fails with obscure errors.  This may occur if
- TMPDIR is set to a long path name
- rundir is set on the command line with -Srundir=path
- local-uri is set on the command line with -Slocal-uri=uri

Add some checks to catch this early and print reasonable errors.  Also update flux-broker-attributes(7) to discourage the use of `rundir` and warn about the AF_UNIX path limit, which is quite small and may surprise some.

### DIFF
--- a/doc/man7/flux-broker-attributes.rst
+++ b/doc/man7/flux-broker-attributes.rst
@@ -23,11 +23,14 @@ size
    The number of broker ranks in the flux instance
 
 rundir
-   A temporary directory available for scratch storage within the session.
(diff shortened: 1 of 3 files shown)


grondo (approved):
Great improvement! LGTM!

garlick:

Issue #3811, opened 2021-08-03
kvs: support mechanism to checkpoint  and restore guest namespaces


garlick:
Problem: once the exec system is capable of preserving running jobs across an instance restart (#3801), KVS guest namespaces for those jobs will need to be preserved as well.

Note that the primary kvs namespace final root reference is written out to the content backing store when flux is shut down.  This may be extended to include defensive checkpoints of the primary namespace as discussed in #3552.

A design should consider that we want the capability to garbage collect un-referenced blobs from the content backing store while the instance is shut down (#258).   A hypothetical utility would create a new content backing store and only copy over the content blobs from the old backing store that are referenced from the most recent primary namespace checkpoint.  Blobs referenced only by guest namespaces would be lost.  This means that it is not helpful for, say, the rank 0 job shell for a job to track the most recent rootref returned from a guest namespace KVS commit, since what it points to might be garbage collected.  (It might be helpful for detecting that data has been lost however).

The job-exec module is responsible for linking the final snapshot of the guest namespace into the primary namespace  once the job completes.  Maybe for an initial cut, the job-exec module could do the same for any running jobs during its shutdown, and then on startup, recreate the guest namespaces from the snapshots (after first discovering which jobs are supposed to still be running, method TBD).

The initial cut could then be extended for defensive checkpointing some way, to allow for recovery from a rank 0 crash.

An alternative to defensive checkpointing of guest namespaces, which might have scalability ramifications, is to have the rank 0 shell "back up" any critical job data that has been written to the guest namespace and restore it upon request.

grondo:
Just curious: since the kvs is already writing out final root reference on shutdown, why not have it also checkpoint currently active namespaces, so they can be resurrected at startup? i.e. wouldn't it be simpler for the kvs just to recreate namespaces from a checkpoint, than to have job-exec have to go through the process of linking incomplete namespaces into the main namespace, only to undo that on restart? (I'm guessing there is some underlying reason the proposed solution is better that I'm not seeing)

One benefit of taking care of this in the kvs is that namespaces created for any other purpose would be preserved, though admittedly the only use case now is for running jobs...

garlick:
[...]

PR #3941, opened 2021-11-04
kvs: optionally initialize namespace to a root reference

chu11:
While playing around with #3811 I realized it was impossible to initialize a namespace to a specific root reference.  So "checkpointing" was sort of impossible.

This updates the KVS namespace create RPC to allow specification of a root reference to initialize the namespace to.

Edit: Note that I did break ABI on `libkvs` to allow user to pass a `const char *rootref`.  To my knowledge, no flux-framework callers outside of `flux-core`.

### DIFF
--- a/src/modules/kvs/kvs.c
+++ b/src/modules/kvs/kvs.c
@@ -2230,15 +2230,12 @@ static void stats_clear_request_cb (flux_t *h, flux_msg_handler_t *mh,
 }
 
 static int namespace_create (struct kvs_ctx *ctx, const char *ns,
-                             uint32_t owner, int flags, const char **errmsg)
+                             const char *rootref, uint32_t owner,
+                             int flags, const char **errmsg)
 {
     struct kvsroot *root;
-    json_t *rootdir = NULL;
-    char ref[BLOBREF_MAX_STRING_SIZE];
-    void *data = NULL;
     flux_msg_t *msg = NULL;
     char *topic = NULL;
-    int len;
     int rv = -1;
 
     /* If namespace already exists, return EEXIST.  Doesn't matter if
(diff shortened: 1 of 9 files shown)


garlick (changes requested):
This seems like a good step, but can we do it without breaking ABI?  I'm concerned we're a bit past the point where we can just do that at will, despite our 0.x.x. semantic version.  Could we just add a function like `flux_kvs_namespace_create_with()` with the new function signature?

Another thought is that it might simplify this a bit if the `kvs.namespace-create` RPC always required a treeobj parameter, and `flux_kvs_namespace_create()` just internally created one and supplied it.

chu11:
re-worked per @garlick comments above.   I redid the KVS namespace create RPC to always take a rootref.  In the common case, `flux_kvs_namespace_create()` creates an empty treeobj dir and than gets the rootref of that before sending it to the KVS.  I elected to do it this way so we could get the relatively costly `blobref_hash()` function out of the KVS.

There is a potential corner case here once we do garbage collection.  It is always assumed that the "empty directory" reference exists in the content store b/c that's how all KVSes begin.  Once we do garbage collection, perhaps there will have to be a few "always keep around" references list.

After that added a new `flux_kvs_namespace_create_with()` per comment above.

src/modules/kvs/kvs.c
```diff
@@ -2262,23 +2259,7 @@ static int namespace_create (struct kvs_ctx *ctx, const char *ns,
         return -1;
     }
 
-    if (!(rootdir = treeobj_create_dir ())) {
```
garlick, line 2265: I like the negative lines of code in the kvs module :-)

src/modules/kvs/kvs.c
```diff
@@ -2262,23 +2259,7 @@ static int namespace_create (struct kvs_ctx *ctx, const char *ns,
...
-    }
-    len = strlen (data);
-
-    if (blobref_hash (ctx->hash_name, data, len, ref, sizeof (ref)) < 0) {
-        flux_log_error (ctx->h, "%s: blobref_hash", __FUNCTION__);
-        goto cleanup;
-    }
-
-    setroot (ctx, root, ref, 0);
+    setroot (ctx, root, rootref, 0);
```
garlick, line 2262: Just to veirfy - do we know for sure that the object pointed to by the blobref (the empty treeobj) is valid in the content cache?  I'm assuming we must know that because this didn't change from before AFAIKT

garlick (approved):
Thanks for making those changes!  Looks good, just one question inline.

src/modules/kvs/kvs.c
chu11: yes, at the moment its guaranteed b/c that's how the primary KVS is initialized (see `store_initial_rootdir()`).  Although it would be good to add a note about it.  I'll add a comment.

chu11:
re-pushed, just adding a comment covering the question @garlick brought up above

garlick:
Thanks!  Feel free to set MWP when you're ready.

PR #3945, opened 2021-11-05
cron: load crontabs from directory if cron.directory attribute is set

garlick:
cron: load crontabs from directory
    
Problem: sys admins need a way to set up cron jobs for the system instance only, that persist across an instance reboot.
    
Set a `cron.directory`  attribute in the systemd unit file.
    
In rc1, on rank 0 only, if the cron.directory attribute is set, iterate over files in the directory, loading each into the cron service with `flux cron tab <$file`.


grondo (approved):
LGTM!

garlick:
Thanks!  Setting MWP.

PR #3947, opened 2021-11-09
job-manager / job-exec: checkpoint and restore guest KVS namespaces

chu11:
Thought I'd throw up this WIP if there are any high level thoughts / comments.  This is an prototype set of changes for checkpointing KVS namespaces when a broker goes down and "re-attaching" to an already running job.  What this PR does is:

A) upon unloading the job-exec module, checkpoint the rootref of the KVS namespaces of still running jobs
B) upon re-load, if the job is believed to still be running, send `job-exec` a "re-attach" flag from `job-manager`
C) upon re-load, have `job-exec` re-build KVS namespace with checkpointed rootref
D) upon "re-attach", `testexec` will simulate remaining runtime of the job based on the starttime of the job (i.e. perhaps `testexec` thinks the job has run 90 seconds out of 100 seconds, simulate a run for just 10 more seconds)

This currently works with tests like this:

```
flux mini submit --setattr=system.exec.test.run_duration=100s hostname

<wait a little>

# remove all job modules and KVS
flux module remove job-exec; flux module remove sched-simple; flux module remove job-list; flux module remove job-info; flux module remove job-manager; flux module remove job-ingest; flux module remove kvs

<wait a little>

# re-add KVS and all job modules
sleep 5; flux module load kvs; flux module load job-manager; flux module load job-info; flux module load job-list; flux module load job-ingest; flux module load job-exec; flux module load sched-simple
```

the eventlogs for the above test look like this (we would need RFC changes to document any new events we put in here).  Some of the guest event log stuff is just for debugging / info.

```
>flux job eventlog f5xBPt79
1636493577.871553 submit userid=8556 urgency=16 flags=0
1636493577.885256 depend
1636493577.885322 priority priority=16
1636493577.887893 alloc annotations={"sched":{"resource_summary":"rank0/core0"}}
1636493577.890752 start
1636493590.477205 flux-reattach
1636493590.757046 reattach
1636493607.762088 finish status=0
1636493607.766412 release ranks="all" final=true
1636493607.767615 free
1636493607.767666 clean

>flux job eventlog -p guest.exec.eventlog f5xBPt79
1636493577.888897 init
1636493577.890580 starting
1636493577.890601 timer timerrun=30.0
1636493590.755381 reattach
1636493590.756492 re-starting
1636493590.756854 note timeleft=17
1636493590.756869 timer timerrun=17.0
1636493607.761802 timercb
1636493607.761841 complete status=0
1636493607.761891 done


```

need to cleanup code, try and make some code non-synchronous, need to make tests, but I think this is a first good step

### DIFF
--- a/src/modules/job-exec/checkpoint.c
+++ b/src/modules/job-exec/checkpoint.c
@@ -16,6 +16,9 @@
  * primary KVS.  This will allow the namespaces to be recreated if
  * a job manager is brought down than back up.
  *
+ * Also provide helper functions to get rootrefs from the checkpointed
+ * object.
+ *
  * OPERATION
  *
  * Get the KVS rootrefs for all running jobs and commit to
@@ -28,12 +31,76 @@
 #endif
 #include <assert.h>
 #include <unistd.h>
+#include <errno.h>
 #include <flux/core.h>
 
 #include "src/common/libczmqcontainers/czmq_containers.h"
 
 #include "job-exec.h"
 #include "checkpoint.h"
+
+flux_future_t *checkpoint_get_rootrefs (flux_t *h)
+{
+    if (!h) {
+        errno = EINVAL;
+        return NULL;
+    }
+
+    return flux_kvs_lookup (h,
+                            NULL,
+                            0,
+                            "job-exec.kvs-namespaces");
+}
+
+char *checkpoint_find_rootref (flux_future_t *f,
+                               flux_jobid_t id,
+                               uint32_t owner)
+{
+    int saved_errno;
+    flux_t *h = flux_future_get_flux (f);
+    char *rv = NULL;
+    const char *rootrefs;
+    json_error_t error;
+    json_t *o = NULL;
+    size_t index;
+    json_t *value;
+
+    if (flux_kvs_lookup_get (f, &rootrefs) < 0) {
+        flux_log_error (h, "checkpoint_get_rootref: flux_kvs_lookup_get");
+        goto cleanup;
+    }
+
+    if (!(o = json_loads (rootrefs, 0, &error))) {
+        flux_log (h, LOG_ERR, "json_loads rootrefs: %s", error.text);
+        goto cleanup;
+    }
+
+    json_array_foreach (o, index, value) {
+        flux_jobid_t l_id;
+        uint32_t l_owner;
+        const char *rootref;
+
+        if (json_unpack (value,
+                         "{s:I s:i s:s}",
+                         "id", &l_id,
+                         "owner", &l_owner,
+                         "kvsroot", &rootref) < 0) {
+            flux_log (h, LOG_ERR, "json_unpack rootref: %s", error.text);
+            goto cleanup;
+        }
+        if (l_id == id && l_owner == owner) {
+            if (!(rv = strdup (rootref)))
+                goto cleanup;
+            break;
+        }
+    }
+
+cleanup:
+    saved_errno = errno;
+    json_decref (o);
+    errno = saved_errno;
+    return rv;
+}
 
 static int lookup_nsroots (flux_t *h, zhashx_t *jobs, flux_future_t **fp)
 {
--- a/src/modules/job-exec/checkpoint.c
+++ b/src/modules/job-exec/checkpoint.c
@@ -1,0 +1,186 @@
+/************************************************************\
+ * Copyright 2021 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* Prototype checkpoint of running jobs KVS root refs
+ *
+ * DESCRIPTION
+ *
+ * Handle checkpoint of running job's guest KVS namescapes into the
+ * primary KVS.  This will allow the namespaces to be recreated if
+ * a job manager is brought down than back up.
+ *
+ * OPERATION
+ *
+ * Get the KVS rootrefs for all running jobs and commit to
+ * "job-exec.kvs-namespaces".
+ *
+ */
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+#include <assert.h>
+#include <unistd.h>
+#include <flux/core.h>
+
+#include "src/common/libczmqcontainers/czmq_containers.h"
+
+#include "job-exec.h"
+#include "checkpoint.h"
+
+static int lookup_nsroots (flux_t *h, zhashx_t *jobs, flux_future_t **fp)
+{
+    struct jobinfo *job = zhashx_first (jobs);
+    flux_future_t *fall = NULL;
+    flux_future_t *f = NULL;
+
+    while (job) {
+        if (job->running) {
+            if (!fall) {
+                if (!(fall = flux_future_wait_all_create ()))
+                    goto cleanup;
+                flux_future_set_flux (fall, h);
+            }
+            if (!(f = flux_kvs_getroot (h, job->ns, 0)))
+                goto cleanup;
+            if (flux_future_aux_set (f, "jobinfo", job, NULL) < 0)
+                goto cleanup;
+            if (flux_future_push (fall, job->ns, f) < 0)
+                goto cleanup;
+            f = NULL;
+        }
+        job = zhashx_next (jobs);
+    }
+
+    (*fp) = fall;
+    return 0;
+
+cleanup:
+    flux_future_destroy (f);
+    flux_future_destroy (fall);
+    return -1;
+}
+
+static json_t *get_nsroots (flux_t *h, flux_future_t *fall)
+{
+    const char *child;
+    json_t *nsdata = NULL;
+    int saved_errno;
+
+    if (!(nsdata = json_array ())) {
+        errno = ENOMEM;
+        return NULL;
+    }
+
+    child = flux_future_first_child (fall);
+    while (child) {
+        flux_future_t *f = flux_future_get_child (fall, child);
+        struct jobinfo *job;
+        const char *blobref = NULL;
+        json_t *o = NULL;
+        if (!f)
+            goto cleanup;
+        if (!(job = flux_future_aux_get (f, "jobinfo")))
+            goto cleanup;
+        if (flux_kvs_getroot_get_blobref (f, &blobref) < 0)
+            goto cleanup;
+        if (!(o = json_pack ("{s:I s:i s:s}",
+                             "id", job->id,
+                             "owner", job->userid,
+                             "kvsroot", blobref))) {
+            errno = ENOMEM;
+            goto cleanup;
+        }
+        if (json_array_append (nsdata, o) < 0) {
+            json_decref (o);
+            errno = ENOMEM;
+            goto cleanup;
+        }
+        json_decref (o);
+        child = flux_future_next_child (fall);
+    }
+
+    return nsdata;
+cleanup:
+    saved_errno = errno;
+    json_decref (nsdata);
+    errno = saved_errno;
+    return NULL;
+}
+
+static int checkpoint_commit (flux_t *h, json_t *nsdata)
+{
+    flux_future_t *f = NULL;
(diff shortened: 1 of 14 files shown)


garlick:
Haven't had a chance to review but, yay!  Nice!

grondo:
Yeah, me either, but nice work getting us moving forward on this @chu11!

It seems like you are very close to "restarting" from an existing sqlite content backing store. Any idea what prevents that currently (i.e. why not restart an instance entirely from the same `content.backing-path` rather than unload/reload modules).

Excited to try this one out! (and double points for getting the testexec jobs to have the correct duration even after a reattach) :slightly_smiling_face:

chu11:
> It seems like you are very close to "restarting" from an existing sqlite content backing store. Any idea what prevents that currently (i.e. why not restart an instance entirely from the same content.backing-path rather than unload/reload modules).

It was mostly the fact that jobs are killed in rc1/rc3.  I think its doable, but just haven't tried yet :-)

chu11:
just re-pushed with various cleanup so the code isn't quite as big a mess.  Also verified that having a set content store works too.  i.e. this works

```
rm -f /tmp/achu/content.sqlite ; src/cmd/flux start -o,--setattr=content.backing-path=/tmp/achu//content.sqlite

flux mini submit --setattr=system.exec.test.run_duration=60s hostname

src/cmd/flux start -o,--setattr=content.backing-path=/tmp/achu//content.sqlite

```

And after this the job is still listed in `flux jobs -A`.  Note that we have to modify `etc/rc1` to remove the cancellation of jobs when exiting an instance.  For tests I write, I'll add an environment variable or something like that if I need to tell `rc1` to not do that.

chu11:
re-pushed, just adding some tests and some instrumentation to try and get more test coverage

chu11:
removing WIP as I think this has reached a point of being reasonably reviewable

garlick:
Just to double check before digging in - if the namespace already exists, we don't retore a potentially older rootref on top of it do we?

chu11:
> Just to double check before digging in - if the namespace already exists, we don't restore a potentially older rootref on top of it do we?

No, if the namespace already exists you'll just get a normal EEXIST error upon the namespace create, I check for it and fallthrough to the rest of the normal setup code.

```
            if (!job->reattach || errno != EEXIST) {
                jobinfo_fatal_error (job, errno, "failed to create guest ns");
                goto done;
            }
```

although that is a good point that I should write a test for this.

chu11:
just re-pushed, doing some cleanup and prefixing some of the eventlog events with `debug` and only outputting them when job debugging is setup.

chu11:
re-pushed with some minor cleanups, but most importantly added a test for when the guest KVS namespace is not destroyed during a "reattach".

src/modules/job-exec/job-exec.c
```diff
@@ -1094,10 +1094,170 @@ static int configure_implementations (flux_t *h, int argc, char **argv)
...
+{
+    struct jobinfo *job = zhashx_first (ctx->jobs);
+    flux_future_t *fall = NULL;
+    flux_future_t *f = NULL;
+
+    while (job) {
+        if (job->running) {
+            if (!fall) {
+                if (!(fall = flux_future_wait_all_create ())) {
+                    flux_log_error (ctx->h, "flux_future_wait_all_create");
```
garlick, line 1228: Error messages could be improved in this function.

src/modules/job-exec/job-exec.c
```diff
@@ -1094,10 +1094,170 @@ static int configure_implementations (flux_t *h, int argc, char **argv)
...
+}
+
+static json_t *get_namespace_roots (struct job_exec_ctx *ctx,
+                                    flux_future_t *fall)
+{
+    const char *child;
+    json_t *nsdata = NULL;
+
+    if (!(nsdata = json_array ())) {
+        flux_log (ctx->h, LOG_ERR, "json_array");
```
garlick, line 1265: or maybe "out of memory"?

src/modules/job-manager/restart.c
```diff
@@ -266,8 +266,21 @@ int restart_from_kvs (struct job_manager *ctx)
...
             }
         }
-        else if ((job->state & FLUX_JOB_STATE_RUNNING) != 0)
+        else if ((job->state & FLUX_JOB_STATE_RUNNING) != 0) {
             ctx->running_jobs++;
+            job->reattach = 1;
+            if ((job->flags & FLUX_JOB_DEBUG)) {
+                if (event_job_post_pack (ctx->event,
+                                         job,
+                                         "debug.flux-reattach",
```
garlick, line 275: why is "flux" in the name?

At the moment I'm not finding  the event that signifies completion but if it's still there, prob should balance the names, like `debug.exec-reattach-start` and `debug.exec-reattach-finish`.  (those are the event suffixes used by prolog/epilog events)

garlick (commented):
I added a few comments, maybe not super helpful at this point.  I'll make another pass in the morning when I have more than half a brain :-)

src/modules/job-manager/restart.c
chu11: good point, i was just following the `flux-restart` event from earlier.  start and finish seem better.

src/modules/job-exec/job-exec.c
```diff
@@ -1094,10 +1094,170 @@ static int configure_implementations (flux_t *h, int argc, char **argv)
     return 0;
 }
 
-static int unload_implementations (void)
+static flux_future_t *lookup_namespace_roots (struct job_exec_ctx *ctx)
```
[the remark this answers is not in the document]
chu11: don't think i can abstract it at that level of detail b/c the `job->running` flag has to be looked at to determine which jobs are running.  but can definitely be abstracted more, perhaps just passing the jobs hash around.

src/modules/job-exec/job-exec.c
garlick: I meant to say "this commit" rather than "this function".  The error messages all look like they not contain enough context to really run anything down.

chu11:
re-pushed with fixes based on comments above.  I had to squash everything since the fixes were a little odd.

- spliced out the checkpoint code into new files `job-exec/checkpoint.[ch]`
- the `debug.exec-reattach-start/finish` events are added in a new commit, since they were previously across two different commits

src/modules/job-exec/checkpoint.c
```diff
@@ -0,0 +1,261 @@
...
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* Prototype flux job exec service
+ *
+ * DESCRIPTION
+ *
+ * This module implements the exec interface as described in
+ * job-manager/start.c, but does not currently support execution of
```
garlick: Header cut & pasted from job-exec.c.

A paragraph or two describing what this does (and its prototype nature) would be good actually!  (I got momentarily excited :-)

src/modules/job-exec/checkpoint.c
```diff
@@ -0,0 +1,261 @@
...
+    flux_future_t *f = NULL;
+
+    while (job) {
+        if (job->running) {
+            if (!fall) {
+                if (!(fall = flux_future_wait_all_create ())) {
+                    flux_log_error (h, "lookup_nsroots: "
+                                            "flux_future_wait_all_create");
+                    goto cleanup;
+                }
```
garlick, line 36: Looks like all the error cases in this function are equally unlikely and uninteresting to distinguish.  Suggest you simply log one message down in `checkpoint_running()` where the function is called, like "failed to initiate KVS requests for namespace checkpoint".

garlick (commented):
Oops started reviewing again but you are still updating.  I'll just make these two comments and ask you let me know when you're ready.

src/modules/job-exec/checkpoint.c
garlick: Was going to make the same comment for the function after that one.  Maybe whole file needs error logs audited.

chu11:
> Oops started reviewing again but you are still updating. I'll just make these two comments and ask you let me know when you're ready.

Oops, sorry!  Shortly after I pushed I realized I didn't remove that cut & pasted header stuff.  So I tried to sneak it their removal. 

Updated the headers in `checkpoint.[ch]`, and consolidated the log messages per your comments above.

src/modules/job-manager/job.h
```diff
@@ -40,6 +40,7 @@ struct job {
     uint8_t free_pending:1; // free request sent to sched
```
garlick, line 40: In the description, could you indicate that a boolean flag was added to the start request payload?  it's kind of vague with the current wording.  Also, it's for any job that's in the RUN state per its replayed event log, not a "belief" per se (you're giving job manager too much credit!)

garlick (commented):
Sorry to give you dribs and drabs.  Apparently Sue ordered dinner and it's here so I'll be offline for an hour or so.  Couple more comments for ya.

src/modules/job-exec/checkpoint.c
```diff
@@ -0,0 +1,190 @@
...
+
+static int checkpoint_commit (flux_t *h, json_t *nsdata)
+{
+    flux_future_t *f = NULL;
+    flux_kvs_txn_t *txn = NULL;
+    char *s = NULL;
+    int rv = -1;
+
+    if (!nsdata || json_array_size (nsdata) <= 0)
+        return 0;
```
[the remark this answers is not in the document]
chu11: actually, i can just remove this check now, the logic is different than some initial work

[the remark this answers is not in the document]
garlick: Oh duh, that's not an error is it.

chu11:
re-pushed with some minor fixes per discussion above and offline.

- checkpointing now also removes namespaces that were checkpointed.
- minor tweaks to some checkpoint logic
- don't output error when there is nothing to checkpoint (logic error I had earlier)
- update some descriptions

src/modules/job-exec/job-exec.c
```diff
@@ -845,15 +846,111 @@ static flux_future_t *ns_create_and_link (flux_t *h,
...
         flux_log_error (h, "namespace_move: flux_future_and_then");
         flux_future_destroy (f);
         return NULL;
     }
     return f2;
 }
 
+static void get_rootref_cb (flux_future_t *fprev, void *arg)
+{
+    int saved_errno;
```
garlick, line 869: Should this code for restoring namespaces be in the same source file as the code to save them?

garlick (approved):
This seems pretty close.  Just one comment about location of source.

src/modules/job-exec/job-exec.c
chu11: i assume you meant "removing namespaces".  Good point, perhaps it should be in `job-exec.c` b/c that's where we create/setup namespaces too.  I'll move it in there.

chu11:
re-pushed, the remove of recently checkpointed namespaces is now within `job-exec.c`.  I spliced it out into its own commit within the series

garlick:
Well, my thought was that if `checkpoint_running()` in checkpoint.c writes the checkpoint object, that `ns_get_rootref()` which reads the checkpoint object should also be in checkpoint.c.   That way the details of the checkpoint object wouldn't have to be exposed outside of that source file.  Would that potentially work?

chu11:
> Well, my thought was that if checkpoint_running() in checkpoint.c writes the checkpoint object, that ns_get_rootref() which reads the checkpoint object should also be in checkpoint.c. That way the details of the checkpoint object wouldn't have to be exposed outside of that source file. Would that potentially work?

Ohh sorry, I didn't know you were referring to that half of this PR.  Yes, I think would work work and is a good idea too.  Let me do that.

I do think that the "remove namespaces" can be in job-exec though, since its independent of the checkpointing.

garlick:
No my bad - I was not very clear about what code I was referring to!

chu11:
re-pushed, I added

```
flux_future_t *checkpoint_get_rootrefs (flux_t *h);

char *checkpoint_find_rootref (flux_future_t *f,
                               flux_jobid_t id,
                               uint32_t owner);
```

as helper functions to get the checkpoint object and find a rootref within it.  B/c of all of the flux future compositions / continuations, I didn't want to move that code into the `checkpoint` lib.

garlick:
Thanks @chu11 - this looks good.  Let's get this merged so we can keep the momentum up!

Nice work!

PR #3948, opened 2021-11-10
broker/overlay: misc cleanup

garlick:
Suppress a couple of concerning but actually benign error messages @grondo noticed when he was testing flux on a node that was running out of memory.

Plus some other minor quasi-related stuff I had queued up, the most substantial of which is changing an RPC that is used by `flux overlay status` to be more useful for other things TBD.  (This commit came from an earlier monitoring branch so the use case is not there anymore, but I think this is probably a better way to structure the RPC before expanding it to include idle time).


grondo (approved):
LGTM!

garlick:
Thanks!  I'll set MWP

PR #3954, opened 2021-11-11
README: update LLNL-CODE

garlick:
Problem: README lists original LLNL release ID, but the ID was updated when when Flux was relicensed as LGPL-3.0.

Update to the new release ID.

Thanks to @slabasan fpr catching that!


grondo (approved):
Thanks!

PR #3956, opened 2021-11-11
cmd: improve flux-mini bulksubmit --dry-run output with --cc

grondo:
This improves the broken `flux-mini bulksubmit --dry-run` output when the `--cc` option is used.

Instead of just dropping the option in output, this simple change includes `--cc` in the output if the option was used.

```console
$ flux mini bulksubmit --dry-run --cc=1-2 echo {} ::: 1 2 3
flux-mini: submit --cc=1-2 echo 1
flux-mini: submit --cc=1-2 echo 2
flux-mini: submit --cc=1-2 echo 3
```

### DIFF
--- a/src/cmd/flux-mini.py
+++ b/src/cmd/flux-mini.py
@@ -386,7 +386,7 @@ class Xcmd:
 
             #  For better verbose and dry-run output, capture mutable
             #   args that were actually changed:
-            if val != newval:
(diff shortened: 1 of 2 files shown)


chu11 (approved):
LGTM

grondo:
Thanks!

Issue #3946, opened 2021-11-05
resource module exits after duplicate instance broker started


grondo:
On fluke I was playing with starting another multi-user instance as a job in the system instance and ran with `--config-path=/etc/flux/system/conf.d`. This caused the brokers of the job to try to connect up with the existing instance, which in turn seems to have caused a series of inexplicable errors:

```2021-11-05T20:40:28.837851Z sched-fluxion-qmanager.debug[0]: alloc success (queue=default id=26188784882157568)
2021-11-05T20:40:28.872565Z barrier.debug[0]: create shell-26188784882157568-init 2
2021-11-05T20:40:28.876618Z barrier.debug[0]: destroy shell-26188784882157568-init 2
2021-11-05T20:40:28.879742Z barrier.debug[0]: create shell-26188784882157568-start 2
2021-11-05T20:40:28.881580Z barrier.debug[0]: destroy shell-26188784882157568-start 2
2021-11-05T20:40:28.906849Z broker.debug[0]: child sockevent tcp://192.168.64.1:8050 accepted
2021-11-05T20:40:28.913855Z broker.info[0]: overlay auth fluke1 OK
2021-11-05T20:40:28.913952Z broker.debug[0]: child sockevent tcp://192.168.64.1:8050 accepted
2021-11-05T20:40:28.913968Z broker.info[0]: overlay auth fluke1 OK
2021-11-05T20:40:28.914066Z broker.debug[0]: child sockevent tcp://192.168.64.1:8050 handshake succeeded
2021-11-05T20:40:28.914115Z broker.debug[0]: child sockevent tcp://192.168.64.1:8050 handshake succeeded
2021-11-05T20:40:28.914397Z broker.err[0]: rank 100 uuid 309885dc-e25c-4ac0-b2cb-b0da6cc0a25c lost, to be superceded by new hello
2021-11-05T20:40:28.914679Z broker.err[0]: hello from rank 100 uuid 5901d745-5230-4347-b8e0-0b1873b2dec5 status full
2021-11-05T20:40:28.914834Z broker.err[0]: rank 98 uuid 77c38dbc-d199-4685-a4ee-55f6117f510f lost, to be superceded by new hello
2021-11-05T20:40:28.915124Z broker.err[0]: hello from rank 98 uuid 52ccb3eb-d742-489c-98ba-30a4becbf1a4 status full
[...]

PR #3958, opened 2021-11-12
resource: improve topo-reduce error handling

garlick:
This addresses a fatal resource module error discovered by @grondo in #3946.  The error handling code in the `resource.topo-reduce` request handler was not quite correct with respect to handling (expected) duplicate requests.

Also, an earlier commit 66d55f32dfa7e7c132a9b57dfa2056d20415e1e4 was supposed to suppress ENOSYS errors from the `groups.get` RPC, but it handled  `groups.leave` only.  Fix that too.

### DIFF
--- a/t/t2310-resource-module.t
+++ b/t/t2310-resource-module.t
@@ -8,8 +8,7 @@ test_description='Test resource module'
 # Then we will reload after adding TOML to cwd
 export FLUX_CONF_DIR=$(pwd)
 
-# min SIZE=4
-SIZE=$(test_size_large)
+SIZE=4
 test_under_flux $SIZE kvs
 
 # Usage: grep_event event-name <in >out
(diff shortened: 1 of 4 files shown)


garlick:
I've noticed this valgrind test frequently timing out in CI and just realized it was caused by an error in the conversion from `flux jobspec srun`, where a 1m timeout became 1s.  Fixed that here.

t/t2310-resource-module.t
```diff
@@ -8,8 +8,7 @@ test_description='Test resource module'
 # Then we will reload after adding TOML to cwd
 export FLUX_CONF_DIR=$(pwd)
 
-# min SIZE=4
-SIZE=$(test_size_large)
+SIZE=4
```
grondo, line 11: commit title typo: "cover duplicate _top_-reduce messages"

grondo (approved):
LGTM! Just one commit message typo noted inline.

Issue #3960, opened 2021-11-15
improve job-exec host unreachable errors


garlick:
Problem: when a broker becomes disconnected from the overlay network, `job-exec` may print "host unreachable" errors without calling out the broker rank or hostname.

Example [here](https://github.com/flux-framework/flux-core/blob/34f894d4ae471fb0295080f6dd4760748b48c836/src/modules/job-exec/exec.c#L306).

That specific error was  noted when reviewing broker logs from a failed AHA Moles dedicated application time run, but we could probably do a little audit of job-exec and see if there are other places where error messages could be made more useful.

Bonus points if we can include both the hostname and the broker rank, since correlating these logs with other system failures is easier with the hostname.

grondo:
Just a note: there may be diminishing returns fixing the current job-exec implementation, though this improvement seems easy enough.

> Bonus points if we can include both the hostname and the broker rank, since correlating these logs with other system failures is easier with the hostname.

Including this kind of thing in general will require that all modules and services in Flux be made aware of hostname/rank mapping for the instance.

Would it be better to detect these failures at a lower level, and call out the hostname/rank once at that level, then correlating rank-specific module/service failures back to the originally reported failing rank?

garlick:
> Would it be better to detect these failures at a lower level, and call out the hostname/rank once at that level, then correlating rank-specific module/service failures back to the originally reported failing rank?

We should certainly log the problem at the lower level IMHO.

Given the diminishing returns you noted, I agree that we shouldn't invest in a huge change here, but if it's relatively simple to provide more context in the error message, it may help in a post mortem to be able to classify that error as fallout from the lower level one and not something else that has to be run down.

grondo:
> but if it's relatively simple to provide more context in the error message, it may help in a post mortem to be able to classify that error as fallout from the lower level one and not something else that has to be run down.

Agreed. I was mainly arguing against requirement to include the _hostname_ as well as rank at these higher levels, which would require hostname to rank mapping be kept within every module that might report these errors, and if the hostname is reported at a lower level this extra state and effort might be considered wasteful.

Reporting the rank in error messages like this will currently require that the target rank of the RPC be stashed in the future as an aux_item. I wonder if we could extend the `flux_rpc_t` API to add a getter for the target rank, so that the boilerplate of stashing an integer in a `flux_future_t` could be avoided, making it easy and idiomatic to report the target rank for failed RPCs?

garlick:
[...]

PR #3962, opened 2021-11-15
job-exec: improve errors and job exception message on broker disconnect

grondo:
As discussed in #3960, the remote execution and kill errors from the job-exec module do not include any rank information, which is unhelpful in diagnosing problems from logs alone.

This PR adds a `flux_rpc_get_nodeid(3)` API call to make it convenient to print the target rank in rpc errors, and uses this to add rank information to many remote subprocess and job-exec error messages.

I didn't go as far as caching the entire request for every RPC, since I didn't have a use case for anything other than the `nodeid` at this time, and was worried about the extra memory required when many, many RPCs are in flight in a client or module. For now, a `nodeid` field is added to the `struct flux_rpc` which simply caches the `nodeid` argument to `flux_rpc(3)`, making it conveniently available during error handling on the `flux_future_get(3)` or `flux_rpc_get(3)` side.

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -224,6 +224,8 @@ MAN3_FILES_SECONDARY = \
 	man3/flux_rpc_get.3 \
 	man3/flux_rpc_get_unpack.3 \
 	man3/flux_rpc_get_raw.3 \
+	man3/flux_rpc_get_matchtag.3 \
+	man3/flux_rpc_get_nodeid.3 \
 	man3/flux_kvs_lookupat.3 \
 	man3/flux_kvs_lookup_get.3 \
 	man3/flux_kvs_lookup_get_unpack.3 \
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -286,6 +286,8 @@ man_pages = [
     ('man3/flux_rpc', 'flux_rpc_get_unpack', 'perform a remote procedure call to a Flux service', [author], 3),
     ('man3/flux_rpc', 'flux_rpc_get_raw', 'perform a remote procedure call to a Flux service', [author], 3),
     ('man3/flux_rpc', 'flux_rpc', 'perform a remote procedure call to a Flux service', [author], 3),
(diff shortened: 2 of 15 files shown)


grondo:
Hm, a couple builders are suddenly failing `t5000-valgrind.t` here:

```
Running job-info
  ++ flux mini submit -t 1 -n 1 /bin/true
  + id=f9a4aD2X
  + flux job attach f9a4aD2X
  1.129s: job.exception type=timeout severity=0 resource allocation expired
  job-info: Failed with rc=1
```

I'll look into it (probably timing related) and perhaps push a workaround or fix here if I can figure it out.

grondo:
Turns out it was likely due to an unnecessary time limit placed on the job submitted under valgrind. I've removed the timelimit (though I'm a little worried the fact that the job is now timing out indicates this PR has introduced some kind of slowdown)

garlick:
No that one was my bad when converting from `flux jobspec srun`!  I had submitted a fixup as part of pr #3958 (e9f5514).

I think it's probably better to drop the timeout, so we can go with this version.

grondo:
Ugh, sorry I had somehow missed that! I'll drop the fix from this PR and go ahead and review #3958 so it can go in first.

garlick (approved):
Looks great!  Thanks for coming up with this so quickly.  I think it's quite reasonable to just cache the nodeid argument - nice that it has the same footprint as the matchtag accessor.

grondo:
I've added a few commits here to further improve job-exec errors on broker disconnect:

 - Improve the resulting exception note to "lost contact to job shell on broker rank N" as suggested by @garlick 
 - Duplicate all job exception errors generated by the job-exec module to the dmesg log
 - Add a test to ensure killing a broker rank results in a job exception with the expected error message

grondo:
Restarted CI since a builder timed out.

I'll set MWP once someone ACKs the commits I added after the last approval.

garlick (approved):
The updates looks great!  I'll kickstart CI.

grondo:
Thanks! I've set MWP

Issue #3961, opened 2021-11-15
provide mechansim to log broker hostnames


garlick:
Problem: brokers usually log errors with the broker rank, but hostnames would be useful when "bad nodes" need to be tracked down.

As noted in #3959, #3960, and #3897, a bad node can ruin your whole flux instance, and when that happens, it's useful to know the culprit.  We do have `flux overlay gethostbyrank`, but that's not very convenient and it's not available when looking at logs post mortem.

We have `config.hostlist` broker attribute, and a nodelist in `resource.R` which `flux overlay gethostbyrank` falls back to if not a system instance.  It may be helpful if we standardized a single broker attribute name for the hostlist (maybe `hostlist`) and had resource fill it in if not already set.   Code that needs to log hostnames could fetch this attribute and then hold the cached mapping.

A complication is that a dynamically discovered `resource.R` doesn't show up until QUORUM is reached, so it would not be useful for pre-QUORUM errors (like in #3959) when launched by slurm.  Maybe the SLURM_JOB_NODELIST environment variable could be used as the initial value of the hostlist if available.

Finally, rather than require every module that needs the hostlist to fetch this attribute, I wonder if we could find a way to map ranks to hostnames automatically in `flux_log()` if a special token is logged like `@gethostbyrank:RANK@` or `@gethostsbyranks:IDSET@`?

Edit: RANK is just a degenerate IDSET so second token not needed!

grondo:
> Finally, rather than require every module that needs the hostlist to fetch this attribute, I wonder if we could find a way to map ranks to hostnames automatically in flux_log() if a special token is logged like @gethostbyrank:RANK@ or @gethostsbyranks:IDSET@?

This is a great idea and I was having a similar thought. I wonder if we can use [mustache(5)](https://mustache.github.io/mustache.5.html) templates for this. Then we could, perhaps, add other template substitutions in the logs (though I can't think of what that would be at the moment).

Unfortunately, I think the mustache template to render a "callable" would be a little more verbose, e.g.
```
{{#gethostsbyranks}}IDSET{{/gethostsbyranks}}
```
and currently our simplistic mustache template implementation doesn't support "sections". But it would be nice to use one templating system in Flux instead of multiple...

garlick:
Although mustache templates might still be a cool thing to do, the main problem of this issue was solved by #3966 and #3971, so closing.

Issue #3344, opened 2020-11-16
implement flux hostlist command


chu11:
discussed several times in other issues.  this could be convenient for scripting, should be done after PR #3341 is done.

grondo:
We also had a `flux hostlist` command in the days of wreckrun which had a different kind of use case. We should look back at that command and see if we can usefully combine the two use cases.

chu11:
I thought of cranking this one out only to realized I would just be re-implementing https://github.com/grondo/lua-hostlist

perhaps could be copied in and tweaked?

grondo:
This one has languished since there isn't really a current use case.

However, some local issues have been opened up against `lua-hostlist` that could be solved by implementing a `flux hostlist`. We can't just copy it in an tweak it because `libflux-hostlist` API and the original hostlist API are quite different.

Let's pause on this one until I decide if it is easier to fix the issues with `lua-hostlist` by updating that code, or whether we should move the code here.

grondo:
A user asked how to create a hostlist file for a job and I realized there isn't an easy way at the moment (without resorting to `lua-hostlist`).

I've prototyped a simple `flux-hostlist` command which takes an option set of "sources", including:
 - "local" - get a hostlist from the enclosing job or instance
 - jobid - get hostlist for named jobid
 - "stdin" or "-" - get a hostlist from stdin
 - raw hosts - treat argument as hostlist

The default would be "local", so that `flux hostlist` gives you a hostlist for the enclosing instance/job:
```console
$ src/cmd/flux hostlist
pi[3,0-2,4]
```

But you can also target jobids so you can do things like find the common hosts for the last 5 failed jobs:
```console
$ src/cmd/flux hostlist --intersect $(flux jobs -f failed -c 5  -no {id})
pi[0,3]
```

Otherwise, the command could be used similarly to lua-hostlist (with more limited options for now)
```
$ src/cmd/flux hostlist --help
usage: flux-hostlist [-h] [-e] [-d S] [-c] [-n NTH] [-L LIMIT] [-S] [-u] [-i]
                     [--fallback]
                     [SOURCES ...]

positional arguments:
  SOURCES            One or more hostlist sources: 'local' (job or current
                     instance), '-' or 'stdin' (read hosts on stdin), 'jobid',
                     or 'hosts' (default: local)

optional arguments:
  -h, --help         show this help message and exit
  -e, --expand       Expand hostlist using defined output delimiter
[...]

PR #3963, opened 2021-11-16
WIP: add broker.hostlist attribute

garlick:
Following up on #3961, this renames the `config.hostlist` attribute to `broker.hostlist`, and has the resource module set it from _R_ if it is not already set by the broker config bootstrap.  Then `flux overlay` is converted to always use this instead of falling back to reading _R_ from the KVS.

We didn't have future-based libflux API calls for manipulating broker attributes so I refactored that code a little to add them.  I included a `flags` parameter which lets you optionally direct the operation to rank 0, which is helpful here because `broker.host` is only set on rank 0 when the resource module sets it.

This still feels a little half baked and some tests are needed so adding the WIP prefix.

### DIFF
--- a/doc/man7/flux-broker-attributes.rst
+++ b/doc/man7/flux-broker-attributes.rst
@@ -190,7 +190,7 @@ hello.hwm
 CONFIG ATTRIBUTES
 =================
 
-config.hostlist
+broker.hostlist
    The rank-ordered hosts specified in the ``bootstrap`` section of
    the Flux configuration.  Hosts are listed in RFC29 hostlist format.
 
--- a/src/broker/boot_config.c
+++ b/src/broker/boot_config.c
@@ -277,10 +277,10 @@ int boot_config_attr (attr_t *attrs, json_t *hosts)
         goto error;
 
     if (attr_add (attrs,
-                  "config.hostlist",
+                  "broker.hostlist",
                   s,
                   FLUX_ATTRFLAG_IMMUTABLE) < 0) {
-        log_err ("attr_add config.hostlist %s", s);
+        log_err ("attr_add broker.hostlist %s", s);
         goto error;
     }
 
--- a/src/broker/test/boot_config.c
+++ b/src/broker/test/boot_config.c
@@ -466,9 +466,9 @@ void test_attr (const char *dir)
     ok (rc == 0,
         "boot_config_attr works NULL hosts");
     errno = 0;
-    ok (attr_get (attrs, "config.hostlist", NULL, NULL) < 0
+    ok (attr_get (attrs, "broker.hostlist", NULL, NULL) < 0
         && errno == ENOENT,
-        "attr_get cannot find config.hostlist after NULL hosts");
+        "attr_get cannot find broker.hostlist after NULL hosts");
 
     hosts = json_array ();
     if (hosts == NULL)
@@ -476,9 +476,9 @@ void test_attr (const char *dir)
     rc = boot_config_attr (attrs, hosts);
     ok (rc == 0,
         "boot_config_attr works empty hosts");
-    ok (attr_get (attrs, "config.hostlist", NULL, NULL) < 0
+    ok (attr_get (attrs, "broker.hostlist", NULL, NULL) < 0
         && errno == ENOENT,
-        "attr_get cannot find config.hostlist after hosts");
+        "attr_get cannot find broker.hostlist after hosts");
     json_decref (hosts);
     hosts = NULL;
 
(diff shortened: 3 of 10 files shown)


grondo:
I'm guessing the attribute is `broker.hostlist` instead of just `hostlist` to differentiate that this is a direct mapping of broker ranks to hostnames, and not necessarily a list of all hosts in the instance (though for now at least I think they're the same)?

I ask only because there are currently other attributes which we should perhaps think about moving under `broker.` namespace (e.g. `rank`, `size`) if we're moving to this idiom.

I liked your idea of potentially capturing a hostlist early in startup from the environment or other means when possible. I wonder if it would make sense to allow a simple "plugin" interface in the broker to allow for that, with "from config", "from provided R" and "from `SLURM_JOB_NODELIST`" being a few of the  first builtin plugins.

garlick:
Now that you mention it maybe `hostlist` would be fine.  I was vaguely thinking I needed to distinguish a broker rank ordered hostlist from some other rank ordered hostlist (like job or shell) but that's not really an issue now that you mention it.  You're right, "broker" is a dumb prefix since these are all "broker attributes".  Some other attributes names could probably be cleaned up too.

I think we need a slurm environment variable that applies to the job step not the job.  For example, if one or two nodes from the job are excluded because they are slow/bad.  Maybe there is some combination of environment variables that would let us figure it out. 

I'll think about plugins.  Sort of hurts my brain in the _R_ case.

grondo:
> I'll think about plugins. Sort of hurts my brain in the R case.

I agree. To me it also feels strange to have a module set a broker attribute.

Maybe the job shell could set `FLUX_JOB_HOSTLIST` so that the broker could set the attribute itself without fetching R from the enclosing instance? Then it is just a matter of trying a couple different environment variables? (I have a feeling that for "other" situations, we still may want a plugin system that could do more than just check an environment variable)

grondo:
> I think we need a slurm environment variable that applies to the job step not the job. For example, if one or two nodes from the job are excluded because they are slow/bad. Maybe there is some combination of environment variables that would let us figure it out.

Yeah, I meant "whatever Slurm variable gives us the right list of nodes". A short search turns up:
```
   SLURM_STEP_NODELIST   List of nodes allocated to the step.
```

However, it doesn't feel like we'd want to hard code every possible environment variable for every possible foreign launcher here, thus the suggestion of offering a plugin interface. (Though probably for now handling Flux, Slurm, and LSF would be enough I guess if you prefer hard coding)

garlick:
> SLURM_STEP_NODELIST   List of nodes allocated to the step.

Duh!  How'd I miss that one?

I think you're right that a pluggable mechanism for obtaining the hostlist in one go from the enclosing instance (whatever it may be) would be simpler than leaning on R.   We could ensure it happens early, to support better log messages before QUORUM is reached, and we could probably do it on all broker ranks instead of only rank 0.

PMI might be a useful way to pass the hostlist from job shell to broker when flux launches flux, perhaps preferable to an environment variable?   If the key is not present, we could then try a stack of plugins until one succeeds, and maybe just fail the launch if none succeed, thus encouraging people wanting support for a different launcher to contribute a plugin.

garlick:
I think this PR is going to need a do-over so I'll go ahead and close it.  I appreciate the feedback - we'll land in a better place if we do this right!

grondo:
I actually thought this was a good step forward if we made a hostlist attribute available in most cases.
Note: possibly related is #3344, though I can't remember the old use case for `flux hostlist`, maybe a `flux hostlist` command would provide a better UI for things like converting between rank and hostname. (Though maybe we if don't have a good use case for it we should just leave it out and close the issue)

grondo:
>  OTOH, if we add the hostname to the broker's PMI business card, we can get this done without any special cases.

Great, that does sound like a clean solution. I'm assuming brokers already exchange business cards in an all-to-all so there would be no huge extra impact there?

We should also evaluate the relative importance of having the the hostlist attribute available. It is really only a small convenience in the error log case, to avoid an extra step of manually translating a rank to hostname. If this starts to become relatively difficult to get right or has an impact on efficient launch then we can perhaps back away for now.

garlick:
Our PMI implementation does an all to all exchange internally, but that just means it's a local operation to pull things out of the PMI KVS.  Since PMI KVS fetches are synchronous RPCs that cannot be parallelized, sequentially fetching a key for each rank does add some latency that scales linearly with instance size.

I do like the simplicity of this approach though.  I'll do some tests and we can decide if it's worth it.

There are for sure other ways to do this, and we could always deploy one of those at a later date as an optimization.  Perhaps right now though, we're less worried about instance startup time than about the general job throughput case.

PR #3966, opened 2021-11-16
broker: set hostlist attribute when Flux is boostrapped from PMI

garlick:
Problem: a single-user flux instance boostrapped from flux or a foreign launcher does not have a hostlist, which would be useful for logging as suggested in #3961.

This renames  the `config.hostlist` broker attribute to `hostlist` and removes an assumption in the resource module that a configured _R_ must be re-ranked if the attribute is present (really it should only be re-ranked if the `[bootstrap]` config table exists).

Then the `hostlist` attribute is set during PMI bootstrap by adding the hostname to each node's business card and reading all the business cards on each rank.

In this attempt, the hostlist is guaranteed to be set after the broker exists bootstrap, regardless of the bootstrap method, which makes it useful for logging during wire-up.

This is a revamp of PR #3963.

Labeled WIP pending performance testing.

### DIFF
--- a/doc/man7/flux-broker-attributes.rst
+++ b/doc/man7/flux-broker-attributes.rst
@@ -39,6 +39,9 @@ content.backing-path
    If file exists, its content is imported into the instance.
    If it doesn't exist, it is created.
 
+hostlist
+   An RFC29 hostlist in broker rank order.
+
 
 TOPOLOGY ATTRIBUTES
 ===================
@@ -185,14 +188,6 @@ hello.hwm
    The reduction high water mark for the broker wireup protocol,
    normally calculated based on the topology.
    Set to 0 to disable the high water mark.
-
-
-CONFIG ATTRIBUTES
-=================
-
-config.hostlist
-   The rank-ordered hosts specified in the ``bootstrap`` section of
-   the Flux configuration.  Hosts are listed in RFC29 hostlist format.
 
 
 RESOURCES
--- a/src/broker/boot_config.c
+++ b/src/broker/boot_config.c
@@ -277,10 +277,10 @@ int boot_config_attr (attr_t *attrs, json_t *hosts)
         goto error;
 
     if (attr_add (attrs,
-                  "config.hostlist",
+                  "hostlist",
                   s,
                   FLUX_ATTRFLAG_IMMUTABLE) < 0) {
-        log_err ("attr_add config.hostlist %s", s);
+        log_err ("attr_add hostlist %s", s);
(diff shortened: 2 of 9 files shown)


garlick:
Quick check on my desktop, FWIW

With this PR:
```
$ /flux start -s256 -o,-v
[snip]
flux-broker: boot: rank=0 size=256 time 0.794s
[snip]
```

Current master:
```
$ /flux start -s256 -o,-v
[snip]
flux-broker: boot: rank=0 size=256 time 0.298s
[snip]
```

garlick:
Re-pushed with mem leak fixed.  I keep forgetting to run that valgrind test manually before pushing.

grondo:
> Re-pushed with mem leak fixed. I keep forgetting to run that valgrind test manually before pushing.

Set `FLUX_ENABLE_VALGRIND_TEST=t` in your `.bash_profile` :wink: 

I went ahead and ran some quick tests on fluke with 200 brokers:

master:
```console
$ time flux mini run -n200 -o pty src/cmd/flux start -o,-v flux getattr size
lt-flux-broker: boot: rank=0 size=200 time 0.034s
lt-flux-broker: parent: none
lt-flux-broker: child: tcp://[::ffff:192.168.128.6]:49153
200
[detached: session exiting]
                           
real	0m8.089s
user	0m0.116s
sys	0m0.025s
```

this PR branch:
```console
$   time flux mini run -n200 -o pty src/cmd/flux start -o,-v flux getattr size
lt-flux-broker: boot: rank=0 size=200 time 0.090s
lt-flux-broker: parent: none
lt-flux-broker: child: tcp://[::ffff:192.168.128.6]:49152
200
[detached: session exiting]
                           
real	0m13.872s
user	0m0.129s
sys	0m0.014s

```

garlick:
My fluke tests mostly back up your results  - about a 3x slowdown in the boot timing, but barely taking 0.1s at 200 nodes so not really worried about it. I'm not seeing much difference in overall runtime between master and this branch, at least after the first run.

grondo (approved):
FWIW, LGTM!

As noted, this PR was tested under system instance of Flux as well as under Slurm.

I'm sure this is already in progress, but in case not, the name of the attribute still needs to be updated in `flux-broker-attributes(7)`.

garlick:
Thanks!  Just pushed the man page change.  I had forgotten.  I'll go ahead and set MWP.

grondo:
You'll also need to remove WIP from title :slightly_smiling_face:

PR #3968, opened 2021-11-17
testsuite: sanitize environment and fix hang in t2607-job-shell-input.t

grondo:
Just a couple small testsuite fixes in one PR.

This PR unsets `FLUX_RC_EXTRA` and `FLUX_SHELL_RC_PATH` for all sharness tests since these variables from the environment could interfere with test reproducibility.

Also improve a couple flaky tests in `t2607-job-shell-input.t`. The main improvement is more efficient submission of duplicate jobs and addition of a timeout, which _should_ turn the hangs we sometimes see in CI into hard failures (at which point we could investigate further)

### DIFF
--- a/t/sharness.d/flux-sharness.sh
+++ b/t/sharness.d/flux-sharness.sh
@@ -375,4 +375,9 @@ fi
 for var in $(env | grep ^PMI); do unset ${var%%=*}; done
 for var in $(env | grep ^SLURM); do unset ${var%%=*}; done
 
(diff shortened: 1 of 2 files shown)


garlick (approved):
Excellent!   That test fails pretty often for me on my desktop, and I'm embarassed to say that despite this reproducibility, I never spent enough time to run it down.  Thank you.

grondo:
This may only fix the hangs, not failures fo the test, which I'm unable to reproduce 😞

PR #3971, opened 2021-11-18
add flux_get_hostbyrank() and improve caching of broker attributes

garlick:
Now that the broker `hostlist` attribute is always set, add a function for translating a broker rank to hostname, and use it to make a few log messages more useful.

This also makes a small change to the broker so that "immutable" broker attributes like `hostlist` are automatically cached in the broker flux_t handle and in all the module flux_t handles.   In other contexts, `flux_get_hostbyrank()` might make an RPC on the first call to fetch the hostlist attribute.

### DIFF
--- a/src/broker/overlay.c
+++ b/src/broker/overlay.c
@@ -1016,6 +1016,7 @@ static void hello_request_handler (struct overlay *ov, const flux_msg_t *msg)
     int version;
     const char *errmsg = NULL;
     char errbuf[128];
+    const char *reason = "";
     flux_msg_t *response;
     const char *uuid;
     int status;
@@ -1038,6 +1039,7 @@ static void hello_request_handler (struct overlay *ov, const flux_msg_t *msg)
                   (unsigned long)ov->parent.rank);
         errmsg = errbuf;
         errno = EINVAL;
+        reason = "not a peer of this broker - mismatched config?";
         goto error_log;
     }
     /* Oops, child was previously online, but is now saying hello.
@@ -1046,7 +1048,8 @@ static void hello_request_handler (struct overlay *ov, const flux_msg_t *msg)
      */
     if (subtree_is_online (child->status)) { // crash
         flux_log (ov->h, LOG_ERR,
-                  "rank %lu uuid %s lost, to be superceded by new hello",
+                  "%s (rank %lu) reconnected after crash, dropping old uuid %s",
+                  flux_get_hostbyrank (ov->h, child->rank),
                   (unsigned long)child->rank,
                   child->uuid);
         overlay_child_status_update (ov, child, SUBTREE_STATUS_LOST);
@@ -1056,6 +1059,7 @@ static void hello_request_handler (struct overlay *ov, const flux_msg_t *msg)
     if (!version_check (version, ov->version, errbuf, sizeof (errbuf))) {
         errmsg = errbuf;
         errno = EINVAL;
+        reason = "flux-core version mismatch - needs software update?";
         goto error_log;
     }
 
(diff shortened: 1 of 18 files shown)


src/common/libflux/attr.c
```diff
@@ -184,6 +189,34 @@ int flux_get_rank (flux_t *h, uint32_t *rank)
...
+        const char *val;
+        if (!(val = flux_attr_get (h, "hostlist"))
+            || !(c->hostlist = hostlist_decode (val)))
+            goto error;
+    }
+    if (!(result = hostlist_nth (c->hostlist, rank)))
+        goto error;
+    return result;
+error:
+    snprintf (c->hostbuf,
```
grondo, line 213: Wow, it is kind of a bummer the attr cache has to contain 64 bytes `hostbuf` just to print `rank=N` when `hostlist_nth()` fails. Perhaps not a big deal, but what if `flux_get_hostbyrank()` returns NULL on error with errno set? Most of the existing use cases _also_ print rank, and I think `%s` in a format will usually print `(null)` if the string argument is NULL...

Eh, not a big deal in the end, I just thought I'd bring it up.

src/broker/overlay.c
```diff
@@ -963,7 +963,9 @@ static void parent_cb (flux_reactor_t *r, flux_watcher_t *w,
                 logdrop (ov, OVERLAY_UPSTREAM, msg, "malformed keepalive");
             }
             else if (ktype == KEEPALIVE_DISCONNECT) {
-                flux_log (ov->h, LOG_ERR, "parent disconnect");
+                log_msg ("%s (rank %lu) sent disconnect control message",
```
grondo, line 966: I was under the impression that `flux_log (h, LOG_ERR, ..)` is already sent to stderr if the severity is <= `log-stderr-level`?

grondo (approved):
Wow, nice little improvement here! Feel free to set MWP. All my comments are merely questions or optional suggestions.

src/common/libflux/attr.c
garlick: Now that I look at that, it does seem like overkill!

In case this gets used in other contexts besides printf, maybe returning "(null)" would be OK.

src/broker/overlay.c
garlick: Well not currently.  Maybe we should consider switching this default to `local`
```
log-stderr-mode
   If set to "leader" (default), broker rank 0 emits forwarded logs from
   other ranks to stderr, subject to the constraints of log-forward-level
   and log-stderr-level.  If set to "local", each broker emits its own
   logs to stderr, subject to the constraints of log-stderr-level.
```

garlick: Actually there is also this:
```
log-critical-level
   Log entries at syslog(3) level at or below this value are copied
   to stderr on the logging rank, for capture by the enclosing instance.
```
Which currently defaults to LOG_CRIT.  I could log this message at LOG_CRIT.

Perhaps in a future PR we can adjust log-critical-level and log-stderr-mode.

garlick:
Repushed with changes that should address comments above.  Thanks!

src/broker/overlay.c
grondo: Sounds good to me! Thanks for the clarification.

Something to consider: for all single user instances stderr of rank 0 is _probably_ the same destination as all other ranks, since we're running in a parallel job where stderr is captured from all job ranks, or we're running with `flux start -s` where `stderr` is shared.

In the system instance case, stderr is captured in the local systemd journal, so it would probably be really helpful if local `LOG_ERR` and higher severity error messages were captured here by default because that might be the first place a sysadmin would look.

(Not sure if that actually clarifies anything, except that copying error messages to stderr might be best for most circumstances, unless it causes the messages to be duplicated at rank0)

In any event, nothing needs changing in this PR afaics.

garlick: Thanks!  I opened that as a separate issue.

grondo:
Feel free to set MWP as far as I'm concerned. Thanks!

PR #3974, opened 2021-11-19
flux-overlay status: rework options

garlick:
Problem:  flux -overlay requires a bunch of options to get the most useful output.

The inadequacy of the current default output was brought home to me in our discussion with the AHA moles team.  I suggested the command, and Jeff ran it on the call and said "it just prints full".   If you're running the tool you probably want the detail.  If you don't want the detail, then you can learn the options to dial it down.

So, rework the options so that one doesn't have to specify `-vvv --ghost --pretty --color` to get the "best" output.  Now the default includes all the brokers with full gaudiness, which can be dialed down with `--no-pretty --no-color` if desired.  Add two options `--summary` (the former default - terse one word summary of subtree status), and `--down`, which only shows down/offline brokers and their TBON parents.

Repurpose `-v` to show time in current state and `-vv` to add RPC round trip time instead of the `--since` and `--times` options.

Finally, add hostnames to the output unconditionally.

Example of default output for a `--test-size=16` instance with rank 3 disconnected:

Screenshot from 2021-11-19 11-58-53

### DIFF
--- a/src/cmd/builtin/overlay.c
+++ b/src/cmd/builtin/overlay.c
@@ -42,9 +42,6 @@ static struct optparse_option status_opts[] = {
     { .name = "timeout", .key = 't', .has_arg = 1, .arginfo = "FSD",
       .usage = "Set RPC timeout (default none)",
     },
-    { .name = "hostnames", .key = 'H', .has_arg = 0,
-      .usage = "Display hostnames instead of ranks",
-    },
     { .name = "times", .key = 'T', .has_arg = 0,
       .usage = "Show round trip RPC times",
     },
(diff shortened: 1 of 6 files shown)


grondo:
Ah, yeah, great idea!

Any sense of how difficult it would be to support similar style output to something like `systemd-cgls` (it is very difficult to figure out which broker lines up where in the current output)

Screenshot from 2021-11-19 12-05-39

garlick:
I do really like that output format!   As discussed offline, I think we would need to build up a data structure representing the overlay tree in one phase, and then render the output in another, if outputting it line by line, because otherwise you can't know whether to start an edge that will connected to a line that would be output later.  I think it would actually be totally fine to make that change, but I have a feeling it would take me a couple of days, so for now I'll suggest that we save that one for a rainy day.

I forgot that we copied `t3301-system-latestart.t` over to flux-sched so it will need a small PR over there.  The use of `flux overlay status` is actually not needed for test correctness so I should be able simply remove it and avoid making flux-sched require the newest flux-core.

grondo (approved):
Nice! I really like these changes!

garlick:
Thanks! OK, seeting MWP.

Issue #3973, opened 2021-11-19
Timer watcher callbacks are garbage-collected by Python prematurely


jameshcorbett:
Investigating an error brought to my attention by Jeff Mast, I think I have narrowed down the problem to unreliable timer-watcher behavior.

I have this print-debugging bit of code from [the Python FluxExecutor](https://github.com/flux-framework/flux-core/blob/f8679d81a1bb5405b8e1751f77fced3915c00ea7/src/bindings/python/flux/job/executor.py#L92):

```python

    def __run(self):
        """Loop indefinitely, submitting jobspecs and fetching jobids."""
        self.__flux_handle.timer_watcher_create(
            self.__poll_interval, self.__submit_new_jobs, repeat=self.__poll_interval
        ).start()
        while self.__work_remains():
            print("ITERATING WHILE LOOP", flush=True)
            self.__submit_new_jobs(reactor_run=False)
            if self.__flux_handle.reactor_run() < 0:
                raise RuntimeError("reactor start failed")

    def __submit_new_jobs(self, *_, reactor_run=True):
        """Pull jobspecs from the queue and submit them.

        Invoked on a timer, and passed several arguments, none
        of which are used---hence the "*_"
        """
        print("CALLING IN TO SUBMIT NEW JOBS", flush=True)
```

I would expect to see a print statement every poll_interval (== 0.1) seconds. I do, at first, but once some jobs are submitted, it seems like my timer watcher is forgotten about and the reactor only cares about satisfying my futures related to those jobs. Furthermore, ``reactor_run`` exits after satisfying those job-related futures, again as if it had forgotten about the timer. I'll try to imitate my terminal output:

```
CALLING IN TO SUBMIT NEW JOBS
CALLING IN TO SUBMIT NEW JOBS
CALLING IN TO SUBMIT NEW JOBS
submitted 0 to flux executor
CALLING IN TO SUBMIT NEW JOBS
submitted 1 to flux executor
submitted 2 to flux executor
CALLING IN TO SUBMIT NEW JOBS
CALLING IN TO SUBMIT NEW JOBS
CALLING IN TO SUBMIT NEW JOBS
CALLING IN TO SUBMIT NEW JOBS
flux_jobid_available: script_id: 0 flux_jobid: fZk98zj
flux_jobid_available: script_id: 1 flux_jobid: fZmd8H5
flux_jobid_available: script_id: 2 flux_jobid: fZmd8H6

[...]

Issue #2485, opened 2019-10-25
Infinite recursion in wrapper.Wrapper python object


jameshcorbett:
When I constructed a wrapper.Wrapper instance with no arguments, I got an infinite recursion:
>>> flux.wrapper
<module 'flux.wrapper' from '/usr/lib64/flux/python2.7/flux/wrapper.pyc'>
>>> flux.wrapper.Wrapper()
Exception RuntimeError: 'maximum recursion depth exceeded' in <bound method Wrapper.__del__ of <flux.wrapper.Wrapper object at 0x2aaaad357c10>> ignored
Exception RuntimeError: 'maximum recursion depth exceeded' in <bound method Wrapper.__del__ of <flux.wrapper.Wrapper object at 0x2aaaad357c10>> ignored
...

I checked, and the issue exists in Flux 0.13 too. I fixed the issue, and I was going to create a pull request but I couldn't seem to push a branch to this repo. This is how I fixed it (explanation below)

@@ -279,7 +279,7 @@ class Wrapper(WrapperBase):
    def __getattr__(self, name):
        fun = None
        if re.match("__.*__", name):
            # This is a python internal name, skip it
             raise AttributeError
         # try it bare
        llib = getattr(self, "lib")
         try:
             fun = getattr(llib, name)
         except AttributeError:
@@ -310,8 +310,9 @@ class Wrapper(WrapperBase):
 
     def __clear(self):
         # avoid recursion
-        handle = self._handle
-        if handle is None:
+        if "_handle" in dir(self) and self._handle is not None:
+            handle = self._handle
+        else:
             return
         if self.destructor is None:
             return

When you create a Wrapper object with no arguments, the interpreter immediately calls __del__, which calls clear(), which tries to access self._handle, which calls __getattr__ because self._handle doesn't exist because the object wasn't initialized. Getattr tries to access self.lib, which (like self._handle) doesn't exist, so it calls __getattr__(self, "lib")... and then the cycle repeats and you're stuck. You might consider changing the line llib = getattr(self, "lib") to just llib = self.lib, since it's the same thing.

dongahn:
Thanks @jameshcorbett for reporting this and your fix. 

We have a specific protocol we have to follow with a PR and such. If you have a W/R for now, you would say you use that to make progress. If not, I can update the installation with this patch.

Next time we meet (maybe next Wed), I will show you how we create a PR so that you can do the same thing for future issues like this.

jameshcorbett:
[...]

PR #3975, opened 2021-11-20
python: avoid early garbage collection of Watcher objects

grondo:
This PR is an attempt at a general fix for #3973.

The problem of `Watcher` objects being subject to early GC is solved by adding them automatically to an active watchers set in the `Flux` handle from which they are created. To avoid a circular reference, the internally stored `flux_handle` in `Watcher` objects becomes a `weakref.ref()`. At `destroy()` time, a best effort is made to remove the `Watcher` from the `flux_handle` watcher set.

Unfortunately, I was getting an `AttributeError` with this approach at interpreter exit for `TimerWatcher has no attribute "get_flux"`, which is baffling, so I wrapped the call to get the `Watcher` flux handle in a `try/except` block. It probably means I've done something subtly wrong, but I couldn't find it.

### DIFF
--- a/t/python/t0007-watchers.py
+++ b/t/python/t0007-watchers.py
@@ -121,10 +121,11 @@ class TestSignal(unittest.TestCase):
         sigw.destroy()
 
     def test_signal_watcher(self):
-        cb_called = [False]
+        cb_called = False
 
         def cb(handle, watcher, signum, args):
-            cb_called[0] = True
+            nonlocal cb_called
+            cb_called = True
             handle.reactor_stop()
 
         def raise_signal(handle, wathcer, revents, args):
@@ -141,7 +142,7 @@ class TestSignal(unittest.TestCase):
             to2.start()
             rc = self.f.reactor_run()
             self.assertTrue(rc >= 0, msg="reactor exit")
-            self.assertTrue(cb_called[0], "Signal Watcher Called")
+            self.assertTrue(cb_called, "Signal Watcher Called")
 
     def test_signal_watcher_exception(self):
         def signal_cb(handle, watcher, signum, args):
--- a/t/python/t0007-watchers.py
+++ b/t/python/t0007-watchers.py
@@ -12,6 +12,7 @@
 
 import unittest
 import os
+import gc
 import signal
 import flux
 from subflux import rerun_under_flux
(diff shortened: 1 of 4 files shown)


grondo:
Well, I thought I had run the entire testsuite, but perhaps not with the most recent changes because this broke a lot of tests. Adding `WIP` for now until I figure out what broke.

jameshcorbett:
I'll take a look at the mysterious AttributeError later tonight. I'll also check whether the GC would just work as-is without any weakref stuff, what I read in the PEP made it sound like it might.

jameshcorbett:
I think I have figured out the mysterious ``AttributeError``. If ``__init__`` raises an exception, ``__del__`` will still be called. In this case, the source of the problem is ``test_timer_add_negative`` which raises an exception (as the test expects) in ``TimerWatcher.__init__`` before the ``super()`` call which defines ``self.get_flux``. So I think your ``try / except AttributeError`` is the right way to go.

(Fun fact: my running into this exact same problem resulted in [the first issue](https://github.com/flux-framework/flux-core/issues/2485) I posted to this repo.)

In any case I think we can avoid messing with weakrefs and so on completely because CPython does seem to be able to resolve the circular reference and garbage collect anyway:

```python
import gc

class A:

    def __init__(self, b):
        self.b = b

    def __del__(self):
        print(f"Calling destructor for {self}")


class B:

    def __init__(self):
        self.a = A(self)

    def __del__(self):
        print(f"Calling destructor for {self}")



b = B()
del b
gc.collect()
print("Garbage collected")
```

Example output:
```
[corbett8@ruby964:test_exec]$ python3 ../gc_test.py 
Calling destructor for <__main__.B object at 0x2aaaac14b1d0>
Calling destructor for <__main__.A object at 0x2aaaac14b208>
Garbage collected
[corbett8@ruby964:test_exec]$ 
```

t/python/t0007-watchers.py
```diff
@@ -206,6 +207,33 @@ def timer_cb(handle, watcher, revents, _args):
...
 
 
+class TestIssue3973(unittest.TestCase):
+    @classmethod
+    def setUpClass(self):
+        self.f = flux.Flux()
+
+    def test_watcher_gc(self):
+
+        cb_called = [0]
```
jameshcorbett, line 217: This is the right thing to do for Python 2, and I know it is in some of the other tests in this module, but in Python 3 the standard thing to do is to use the ``nonlocal`` keyword instead of wrapping the value in a list. Something like:

```python
cb_called = 0

    def cb(handle, watcher, revents, _args):
          nonlocal cb_called
          cb_called += 1
```

grondo: Thanks! I was completely unaware and just copied other examples in the test (which were probably from the time of Python2)

grondo:
> I think I have figured out the mysterious AttributeError

Impressive sleuthing!

> In any case I think we can avoid messing with weakrefs and so on completely because CPython does seem to be able to resolve the circular reference and garbage collect anyway:

Ok, so I _did_ actually waste half a day yesterday :facepalm: :laughing: 

Thanks for figuring that out. I'll redo this PR by early next week without using the weakref solution.

jameshcorbett:
> Ok, so I _did_ actually waste half a day yesterday 🤦 😆

Sorry... I probably should have told you not to try weakrefs (which are always a pain) until I/we could confirm whether we actually needed to do anything. The first things I read suggested that Python couldn't GC circular references if the objects defined ``__del__`` (which in our case they do) but I later learned that Python changed that in 3.4.

Thanks for doing this!

grondo:
Ok, @jameshcorbett, I've gone back to the simpler version. Thanks for your help!

I've removed the WIP label so go ahead and review away!

jameshcorbett (approved):
LGTM! I also pulled this branch and then double-checked that it fixes the ``FluxExecutor`` bug that Jeff reported. I will let him know you fixed it.

grondo:
Thanks for all your help @jameshcorbett! I'll set MWP.

PR #3979, opened 2021-11-24
add flux top command

garlick:
This adds a first cut at a `flux top` command.  Usage is `flux top [jobid]`.

There is quite a bit of explanation in the commit that adds the command so I won't duplicate that here.  Instead here are a couple of screen shots.  The first example is of monitoring a batch job running the throughput test:
```
ƒ(s=8,d=0,builddir) $ cat foo.sh
#!/bin/bash
/home/garlick/proj/flux-core/src/test/throughput.py -x -n 10000
ƒ(s=8,d=0,builddir) $ flux mini batch -n16 -t 10m foo.sh
ƒ3BBeCNddq
ƒ(s=8,d=0,builddir) $ flux top ƒ3BBeCNddq
```
The two turtles on the left indicate that this job has an instance depth of 1 (hierarchy of 2).  The hourglass on the right shows the time left in the instance, because it was submitted with a time limit.
Screenshot from 2021-11-24 10-14-29

The second example shows fluke, with some nodes down graphed in red.  There is no expiration time in the system instance, so there is an infinity symbol in the upper right.  Its not shown in these screenshots, but a little heartbeat indicator flashes for each system heartbeat event message, letting you know the instance and `flux top`'s connection to it are functioning.

Screenshot from 2021-11-24 11-29-42

### DIFF
--- a/src/cmd/top/summary_pane.c
+++ b/src/cmd/top/summary_pane.c
@@ -1,0 +1,412 @@
+/************************************************************\
+ * Copyright 2021 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+#include <sys/types.h>
+#include <unistd.h>
+#include <time.h>
+#include <jansson.h>
+
+#include "src/common/libutil/fsd.h"
+#include "src/common/librlist/rlist.h"
+
+#include "top.h"
+
+static const struct dimension win_dim = { 0, 0, 80, 5 };
+static const struct dimension level_dim = { 0, 0, 2, 1 };
+static const struct dimension jobid_dim = { 36, 0, 16, 1 };
+static const struct dimension timeleft_dim = { 70, 0, 10, 1 };
+static const struct dimension resource_dim = { 4, 1, 36, 3 };
+static const struct dimension heart_dim = { 77, 3, 1, 1 };
+static const struct dimension stats_dim = { 60, 1, 15, 3 };
+
+static const double heartblink_duration = 0.5;
+
+struct resource_count {
+    int total;
+    int down;
+    int used;
+};
+
+struct stats {
+    int depend;
+    int priority;
+    int sched;
+    int run;
+    int cleanup;
+    int inactive;
+    int total;
+};
+
+struct summary_pane {
+    struct top *top;
+    WINDOW *win;
+    unsigned long instance_level;
+    flux_jobid_t jobid;
+    double expiration;
+    struct stats stats;
+    struct resource_count node;
+    struct resource_count core;
+    struct resource_count gpu;
+    flux_watcher_t *heartblink;
+    bool heart_visible;
+};
+
+static void draw_timeleft (struct summary_pane *sum)
+{
+    double now = flux_reactor_now (flux_get_reactor (sum->top->h));
+    double timeleft = sum->expiration - now;
+    char buf[32] = "";
+
+    if (timeleft > 0)
+        fsd_format_duration_ex (buf, sizeof (buf), timeleft, 2);
+
+    mvwprintw (sum->win,
+               timeleft_dim.y_begin,
+               timeleft_dim.x_begin,
+               "%*s%s",
+               timeleft_dim.x_length - 2,
+               buf,
+               timeleft > 0 ? "⌚" : "∞");
+}
+
+static void draw_level (struct summary_pane *sum)
+{
+    const char *sup[] = { "", "²", "³", "⁴", "⁵", "⁶", "⁷", "⁸", "⁹", "⁺" };
+    int size = sizeof (sup) / sizeof (sup[0]);
+    unsigned long level = sum->instance_level;
+
+    wattron (sum->win, COLOR_PAIR (TOP_COLOR_YELLOW));
+    mvwprintw (sum->win,
+               level_dim.y_begin,
+               level_dim.x_begin,
+               "%s%s",
+               "ƒ",
+               sup[level < size ? level : size - 1]);
+    wattroff (sum->win, COLOR_PAIR (TOP_COLOR_YELLOW));
+}
+
+static void draw_jobid (struct summary_pane *sum)
+{
+    if (sum->jobid != FLUX_JOBID_ANY) {
+        char buf[jobid_dim.x_length + 1];
+        flux_job_id_encode (sum->jobid, "f58", buf, sizeof (buf));
+        wattron (sum->win, A_BOLD);
+        mvwprintw (sum->win, jobid_dim.y_begin, jobid_dim.x_begin, "%s", buf);
+        wattroff (sum->win, A_BOLD);
+    }
+}
+
+static void draw_stats (struct summary_pane *sum)
+{
+    mvwprintw (sum->win,
+               stats_dim.y_begin,
+               stats_dim.x_begin,
+               "%*d pending",
+               stats_dim.x_length - 10,
+               sum->stats.depend + sum->stats.priority + sum->stats.sched);
+    mvwprintw (sum->win,
+               stats_dim.y_begin + 1,
+               stats_dim.x_begin,
+               "%*d running",
+               stats_dim.x_length - 10,
+               sum->stats.run + sum->stats.cleanup);
+    mvwprintw (sum->win,
+               stats_dim.y_begin + 2,
+               stats_dim.x_begin,
+               "%*d inactive",
+               stats_dim.x_length - 10,
+               sum->stats.inactive);
+}
+
+/* Create a little graph like this that fits in bufsz:
+ *     name [||||||||||        |||32/128]
+ * "used" grows from the left in yellow; "down" grows from the right in red.
+ * Fraction is used/total.
+ */
+static void draw_bargraph (WINDOW *win, int y, int x, int x_length,
+                           const char *name, struct resource_count res)
+{
+    char prefix[16];
+    char suffix[16];
+
+    if (x_length > 80)
+        x_length = 80;
+    if (res.used > res.total)
+        res.used = res.total;
+
+    snprintf (prefix, sizeof (prefix), "%5s [", name);
+    snprintf (suffix, sizeof (suffix), "%d/%d]", res.used, res.total);
+
+    int slots = x_length - strlen (prefix) - strlen (suffix) - 1;
+    mvwprintw (win,
+               y,
+               x,
+               "%s%*s%s",
+               prefix,
+               slots, "",
+               suffix);
+    /* Graph used */
+    wattron (win, COLOR_PAIR (TOP_COLOR_YELLOW));
+    for (int i = 0; i < ((double)res.used / res.total) * slots; i++)
+        mvwaddch (win, y, x + strlen (prefix) + i, '|');
+    wattroff (win, COLOR_PAIR (TOP_COLOR_YELLOW));
+
+    /* Graph down */
+    wattron (win, COLOR_PAIR (TOP_COLOR_RED));
+    for (int i = slots - 1;
+         i >= slots - ((double)res.down / res.total) * slots; i--) {
+        mvwaddch (win, y, x + strlen (prefix) + i, '|');
+    }
+    wattroff (win, COLOR_PAIR (TOP_COLOR_RED));
+}
+
+static void draw_resource (struct summary_pane *sum)
+{
+    draw_bargraph (sum->win,
+                   resource_dim.y_begin,
+                   resource_dim.x_begin,
+                   resource_dim.x_length,
+                   "nodes",
+                   sum->node);
+    draw_bargraph (sum->win,
+                   resource_dim.y_begin + 1,
+                   resource_dim.x_begin,
+                   resource_dim.x_length,
+                   "cores",
+                   sum->core);
+    draw_bargraph (sum->win,
+                   resource_dim.y_begin + 2,
+                   resource_dim.x_begin,
+                   resource_dim.x_length,
+                   "gpus",
+                   sum->gpu);
+}
+
+static void draw_heartbeat (struct summary_pane *sum)
+{
+    mvwprintw (sum->win,
+               heart_dim.y_begin,
+               heart_dim.x_begin,
+               "%s",
+               sum->heart_visible ? "♡" : " ");
+}
+
+/* Fetch expiration time (abs time relative to UNIX epoch) from resource.R.
+ * If unavailable (e.g. we are a guest in the system instance), return 0.
+ */
+static double get_expiration (flux_t *h)
+{
+    flux_future_t *f;
+    double val = 0;
+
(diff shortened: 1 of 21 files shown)


src/common/libutil/fsd.c
```diff
@@ -87,20 +87,30 @@ int fsd_parse_duration (const char *s, double *dp)
     return 0;
```
chu11, line 87: commit message typo "alloews"

grondo:
Super neat!

chu11:
I could not see turtle in my use of `flux top` w/ gnome-terminal.  Just printing it on the command line worked (`printf "\U0001F422\n"`).  I could see the turtle in `vi` and `emacs`, suggesting it wasn't a terminal or font setting issue.  After messing around, I was able to see the turtle on a RHEL8 based system (fluke) but not a RHEL7 based system (catalyst).  Worked on my personal ubuntu box too.  As an aside on an LC system had display issues until I unset `LC_ALL=C`.

I have no idea why it works on RHEL8 but not RHEL7, dunno if it's a `ncurses` issue (RHEL7 - 5.9, RHEL8 - 6.1, ubuntu - 6.2).  I put in some other chars (copyright symbol) and that worked on RHEL7.  So I'm assuming a display issue with newer unicode chars (vs infinity symbol, which is older, and displayed fine).

perhaps some simpler chars should be used to be on the safe side?  Or just a "depth = ..." and "timeleft = ..."?

src/common/libutil/fsd.c
garlick: pushed a fix for that, thanks

garlick:
I'm really out of my depth here, but if "older" versions of unicode are more portable, some alternative choices might be
- ⌚ https://www.compart.com/en/unicode/U+231A
- ♡ https://www.compart.com/en/unicode/U+2661
- ☯ https://www.compart.com/en/unicode/U+262F

chu11:
@garlick the flux f & heart show up now.  Although I see a line of `q` instead of the normal line.

Screenshot from 2021-11-29 10-07-05

i see the q line on both rhel7 & rhel8.

As an aside, I can barely see the heart flicker.  Perhaps just leave it and only remove it if haven't received a heartbeat after some timeout period?  Or perhaps if haven't received a heartbeat in awhile could replace with another emoji?  I would recommend broken heart 💔, but that seems to be a newer unicode.  Perhaps a simple ❌ (274C)?

garlick:
I pushed some changes this morning including lengthening the display time for the heart.  Did that  help?

The horizontal line issue is surprising because it purely uses the ncurses API to draw the line.  However, it's probably not really needed at all and just wastes a line of prime screen real estate, so let's try without.

grondo:
I gave this a quick whirl and found that job names (and perhaps other fields?) are not truncated, resulting in long names messing up the job listing display, e.g.

Screenshot from 2021-11-29 11-34-59

It would be cool if the jobs listing supported the same format specifier as `flux jobs` so that custom fields could also be listed, plus the truncation is already taken care of...

Another idea would be to switch the field order to be more top-like and list the job name last.

Also, not sure if anyone else is experiencing this, but I get flicker when the job listing updates (I though ncurses took care of that, so not sure what is going on, but `top` doesn't flicker in my terminal)

garlick:
Thanks @grondo - I pushed fixes for the field width and moved the NAME field to the end of the line with no width limit.  Looks better!  If you have a chance to try again, please report back on whether the new (old) unicode characters are working for you.

I'm puzzled by the flicker you're reporting.  I'm surely doing something wrong but offhand I can't see what it might be.  I have seen the screen blank momentarily a couple of times on fluke, but I assumed it had to do with a slow connection.  Will investigate further.

I had thought that a `--format` option could be deferred to a future PR.  Does that seem reasonable?

garlick:
Here's a current screen from fluke (eh sorry about the tmux intrusion and the gaudy terminal wiindow!)

Screenshot from 2021-11-29 13-34-46

garlick:
And a batch job with a timeout.

Screenshot from 2021-11-29 13-39-31

grondo:
Nice! :+1: looks good!

> I had thought that a --format option could be deferred to a future PR. Does that seem reasonable?

Yeah, that is fine. I guess I was making the point that a lot work went into the `JobInfoFormat` class in Python and duplicating all that in C sounds like a big chore. Could the job listing window just come from the output of `flux jobs`? I guess a nice feature down the road would be selection of individual job to take some action (like attach to the job, cancel the job, or recursively run `flux top` again on the job), and getting the job listing from a subprocess might make that impossible.

We have to start somewhere though!

chu11:
> I pushed some changes this morning including lengthening the display time for the heart. Did that help?

Doesn't seem to help much.  I upped it to 0.5s and it seems to be alot better at that number.

garlick:
Thanks @chu11 - OK then, 0.5s it is!  :shrug:

@grondo, no great ideas right now.   Worst case,  we could convert the tool to python and retain most of the design.  I don't think that necessarily means this couldn't go in as a stop gap though.

grondo:
> I don't think that necessarily means this couldn't go in as a stop gap though.

Agreed!

grondo:
Ok, I finally solved my flicker issue. The solution comes in two parts:

 1. Use `wnoutrefresh ()` instead of `wrefresh ()` for each pane, followed by `doupdate ()` once. This results in one refresh of the actual terminal screen instead one per pane/window 
See https://invisible-island.net/ncurses/ncurses-intro.html#flush

2.  Use `werase ()` in place of `wclear ()` when drawing a pane. From the `curs_clear(3x)` man page:
>     The  clear and wclear routines are like erase and werase, but they also
>     call clearok, so that the screen is cleared completely on the next call
>     to wrefresh for that window and repainted from scratch.

The repainting from scratch and extra updates to the screen seem to be causing multiple extra repaints of the terminal per loop, resulting in the flicker on my terminal.

```diff
diff --git a/src/cmd/top/joblist_pane.c b/src/cmd/top/joblist_pane.c
index 78342ac18..e15e50865 100644
--- a/src/cmd/top/joblist_pane.c
+++ b/src/cmd/top/joblist_pane.c
@@ -40,7 +40,7 @@ void joblist_pane_draw (struct joblist_pane *joblist)
     size_t index;
     json_t *job;
 
-    wclear (joblist->win);
+    werase (joblist->win);
     wattron (joblist->win, A_BOLD);
     mvwprintw (joblist->win,
                0,
@@ -128,7 +128,7 @@ void joblist_pane_query (struct joblist_pane *joblist)
 
 void joblist_pane_refresh (struct joblist_pane *joblist)
 {
-    wrefresh (joblist->win);
+    wnoutrefresh (joblist->win);
 }
 
 struct joblist_pane *joblist_pane_create (struct top *top)
diff --git a/src/cmd/top/summary_pane.c b/src/cmd/top/summary_pane.c
index 28989ca2a..8ced3742c 100644
--- a/src/cmd/top/summary_pane.c
+++ b/src/cmd/top/summary_pane.c
@@ -359,7 +359,7 @@ void summary_pane_query (struct summary_pane *sum)
 
 void summary_pane_draw (struct summary_pane *sum)
 {
-    wclear (sum->win);
+    werase (sum->win);
     draw_level (sum);
     draw_jobid (sum);
     draw_timeleft (sum);
@@ -370,7 +370,7 @@ void summary_pane_draw (struct summary_pane *sum)
 
 void summary_pane_refresh (struct summary_pane *sum)
 {
-    wrefresh (sum->win);
+    wnoutrefresh (sum->win);
 }
 
 struct summary_pane *summary_pane_create (struct top *top)
diff --git a/src/cmd/top/top.c b/src/cmd/top/top.c
index 6713ccd9c..b27e17a23 100644
--- a/src/cmd/top/top.c
+++ b/src/cmd/top/top.c
@@ -73,6 +73,7 @@ void refresh_cb (flux_reactor_t *r,
 
     summary_pane_refresh (top->summary_pane);
     joblist_pane_refresh (top->joblist_pane);
+    doupdate ();
 }
 
 static int job_get_state (flux_t *h, flux_jobid_t id, int *statep)

```

grondo:
Another thing I noticed is that `top(1)` doesn't change the default background color on my terminal, while `flux top` does. It is minor, but I think calling `use_default_colors ()` before `start_color ()` would be slightly preferable, though then it does make the gauges look a bit strange, as they only get the black background on my terminal behind the fill characters.

garlick:
Thanks @grondo!  Pushing those changes now.

P.s. apparently -1 as the background color for the two color pairs causes it to use the default.

grondo:
Great! That worked. Feel free to squash that commit.

grondo:
One last suggestion: How about if we copy `top`'s use of reverse video for the joblist column header? I think this looks pretty good:

Screenshot from 2021-11-29 19-08-25

Though I'm not sure the best way to calculate the width of the final NAME field...

```diff
diff --git a/src/cmd/top/joblist_pane.c b/src/cmd/top/joblist_pane.c
index e15e50865..428f0f8b8 100644
--- a/src/cmd/top/joblist_pane.c
+++ b/src/cmd/top/joblist_pane.c
@@ -40,14 +40,21 @@ void joblist_pane_draw (struct joblist_pane *joblist)
     size_t index;
     json_t *job;
 
+    /*  Get the total width of the NAME field so that the reverse fill
+     *   goes one less than the edge of the terminal.
+     */
+    int w = getmaxx (joblist->win) - (12 + 8 + 2 + 6 + 6 + 7 + 6);
+
     werase (joblist->win);
     wattron (joblist->win, A_BOLD);
+    wattron (joblist->win, A_REVERSE);
     mvwprintw (joblist->win,
                0,
                0,
-               "%12s %8s %2s %6s %6s %7s %s",
-               "JOBID", "USER", "ST", "NTASKS", "NNODES", "RUNTIME", "NAME");
+               "%12s %8s %2s %6s %6s %7s %-*s",
+               "JOBID", "USER", "ST", "NTASKS", "NNODES", "RUNTIME", w, "NAME");
     wattroff (joblist->win, A_BOLD);
+    wattroff (joblist->win, A_REVERSE);
     if (joblist->jobs == NULL)
         return;
     json_array_foreach (joblist->jobs, index, job) {

```

src/cmd/top/summary_pane.c
```diff
@@ -0,0 +1,422 @@
...
+               suffix);
+    /* Graph used */
+    wattron (win, COLOR_PAIR (TOP_COLOR_YELLOW));
+    for (int i = 0; i < ((double)res.used / res.total) * slots; i++)
+        mvwaddch (win, y, x + strlen (prefix) + i, '|');
+    wattroff (win, COLOR_PAIR (TOP_COLOR_YELLOW));
+
+    /* Graph down */
+    wattron (win, COLOR_PAIR (TOP_COLOR_RED));
+    for (int i = slots - 1; i >= slots - ((double)res.down / res.total) * slots; i--)
```
chu11, line 171: line looks to be > 80 chars

src/cmd/top/summary_pane.c
```diff
@@ -0,0 +1,422 @@
...
+static int get_instance_level (flux_t *h, int *level)
+{
+    const char *s;
+    int lev = 0;
+
+    if (!(s = flux_attr_get (h, "instance-level"))) {
+        log_err ("error fetching instance-level broker attribute");
+        return -1;
+    }
+    lev = strtol (s, NULL, 10);
```
chu11, line 241: should we check for parse error?  b/c there could be issues if `level` returns 0.

src/cmd/top/ucache.c
```diff
@@ -0,0 +1,91 @@
...
+#endif
+#include <sys/types.h>
+#include <stdlib.h>
+#include <pwd.h>
+#include <stdio.h>
+
+#include "top.h"
+
+struct ucache_entry {
+    int id;
```
chu11, line 25: should use `uid_t` everything in this file instead of int?

chu11 (commented):
i know almost nothing about `ncurses`, so just a skim to see if I noticed anything general coding wise.

garlick:
Thanks @chu11 and @grondo - just pushed those changes.  I'll try to get some minimal testing added tomorrow AM.

t/t2801-top-cmd.t
```diff
@@ -0,0 +1,47 @@
...
+test_expect_success 'flux-top fails if job is not running' '
+	test_must_fail flux top $(cat jobid) 2>notrun.err &&
+	grep "job is not running" notrun.err
+'
+test_expect_success 'flux-top fails if stdin is not a tty' '
+	test_must_fail flux top --test-exit </dev/null 2>notty.err &&
+	grep "stdin is not a terminal" notty.err
+'
+test_expect_success 'flux-top --test-exit works with a pty' '
+	$runpty flux top --test-exit >/dev/null
```
grondo, line 44: Note that the `runpty` script already supports a "quit character" sent when the script receives `SIGUSR1`.
Instead of adding a test-only `--test-exit` option, you could use something like the following:
```sh
$runpty --format=asciicast -o log.asciicast --quit-char=q src/cmd/flux top &
pid=$! &&
kill -USR1 $pid &&
wait $pid
```

This could be a more flexible way to test `flux top` since you can wait for an event to occur before quitting the program (and it actually tests using `q` to quit the program), and additionally the screen cast is captured in the asciicast v2 output so you can actually check for (some kinds of) expected output (though with ncurses that could be challenging).

It would also be feasible to eventually extend `runpty.py` to add a scripted list of input if we want to ever add more testing based on keystrokes.

grondo (commented):
Just some quick comments about testing on my first pass.

garlick:
Just forced pushed the following:
- fix ncursesw package name on bionic builder
- update sharness test with some of @grondo's ideas (thanks!)
- make `--test-exit` option invisible
- some #include tidying

garlick:
Just force pushed a change to add a `fatal()` function that calls curses `endwin()` before printing its message.  (Duh, why didn't I do that before?)  Now fatal error messages can actually be seen by the user.

I changed a some internal functions to just call `fatal()` directly instead of returning an error to the caller to eliminate untestable error paths.   But I also separated some error cases that had been lumped together, so net coverage may be a wash.

garlick:
Bionic 32 bit builder seems to have gotten stuck and timed out - couldn't really tell where it was stuck.  Restarted.

src/cmd/top/ucache.c
```diff
@@ -0,0 +1,91 @@
...
+        free (ucache);
+        errno = saved_errno;
+    }
+}
+
+struct ucache *ucache_create (void)
+{
+    struct ucache *ucache;
+
+    if (!(ucache = calloc (1, sizeof (ucache))))
```
grondo, line 86: `ucache` is being allocated sizeof a pointer instead of sizeof `struct ucache`:

```suggestion
    if (!(ucache = calloc (1, sizeof (*ucache))))
```

grondo (approved):
Just a few minor memory-related issues. Otherwise, this LGTM! This will be a neat addition to our release!

src/cmd/top/ucache.c
garlick: Eek!

garlick:
OK given that this was approved, I'm going to set MWP.  Thanks for the great feedback @grondo and @chu11!

grondo:
Nice work!

garlick:
Thanks!

garlick:
Oops - I had forgotten that this PR adds a dependency on ncurses, and included the provisional addition to `checks/Dockerfile` (902534657d1324227ee9c068d650c224c4e4a137) that was intended to be removed before merging.

grondo:
Ah, yeah. I had put it on my list to create new docker images and push them up. I can do that now.

Issue #3959, opened 2021-11-15
flux logs no errors during instance startup if some brokers don't start


garlick:
Problem: sometimes a "bad node" will prevent slurm from starting a flux broker.  After a timeout on the order of minutes, slurm will abort the job launch, but won't print which hostname failed to start.   This was called out as a problem by the AHA moles team during dedicated application time - it is a pain to figure out what node failed so that it can be excluded and a slightly smaller flux instance can be launched without it.

It would be useful if we could log something at LOG_ERR level if the instance is stuck at QUORUM state for longer than some timeout, and include the hostname(s) that are taking too long.

grondo:
> After a timeout on the order of minutes, slurm will abort the job launch, but won't print which hostname failed to start. 

I wonder if the `srun --no-kill` option would be helpful here:

> -k, --no-kill[=off]
> Do not automatically terminate a job if one of the nodes it has been allocated fails. This option applies to job and step allocations. The job will assume all responsibilities for fault-tolerance. Tasks launch using this option will not be considered terminated (e.g. -K, --kill-on-bad-exit and -W, --wait options will have no effect upon the job step). The active job step (MPI job) will likely suffer a fatal error, but subsequent job steps may be run if this option is specified.
> Specify an optional argument of "off" disable the effect of the SLURM_NO_KILL environment variable.
> 
> The default action is to terminate the job upon node failure.
```

I don't know if `srun` will still terminate a job that fails to launch on all nodes with this option, but it might be worth a try. 

It seems like this option would be useful in any case for long running Flux instances, along with something like `-S tbon.arity=256`.

garlick:
That is a good suggestion in addition to logging the slow hostnames!

PR #3980, opened 2021-11-28
broker: log slow nodes during startup

garlick:
This is an attempt to improve logging somewhat when faced with the problem described in #3959, where flux gets stuck starting up because of a bad node, and doesn't provide any clues as to which node is the bad one.  A timer is added to the state machine which is started on rank 0 upon entering QUORUM state, and fires every 60s, logging the brokers that haven't checked in yet, until rank 0 exits the QUORUM state.  Here's an example with the timeout cranked down to 1s:
```
$ flux start -s8 -o,-Sbroker.quorum-timeout=1s
2021-11-28T00:09:52.631500Z broker.err[0]: quorum delayed: waiting for system76-pc (rank 7)
2021-11-28T00:09:52.680774Z broker.err[0]: quorum reached
ƒ(s=8,d=0,builddir) $
```

Unfortunately, if the bad node is a router on the overlay network, any of its children will likely be logged as well, so it may still be ambiguous which node is the root cause.  Still, probably better than nothing.

### DIFF
--- a/etc/flux.service.in
+++ b/etc/flux.service.in
@@ -15,7 +15,8 @@ ExecStart=@X_BINDIR@/flux broker \
   -Slog-stderr-mode=local \
   -Scontent.backing-path=@X_LOCALSTATEDIR@/lib/flux/content.sqlite \
   -Sbroker.rc2_none \
-  -Sbroker.quorum=0
+  -Sbroker.quorum=0 \
+  -Sbroker.quorum-timeout=none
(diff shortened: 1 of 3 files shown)


grondo (approved):
Sorry, seemed to have missed giving this one a review! LGTM!

I tried it out on fluke and it works as advertised. A nice improvement.

garlick:
Thanks for the review and fluke testing!  I'll set MWP.

PR #3981, opened 2021-12-01
NEWS: add release notes for 0.32.0

garlick:
Here are draft notes for flux-core 0.32.0.  As always, feel free to push changes here.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,45 @@
+flux-core version 0.32.0 - 2021-12-05
+-------------------------------------
+
+This release adds early groundwork for recovering running jobs across a
+Flux restart.  It also includes improved log messages based on user feedback
+about Flux failures on real workflow runs, a first draft of a new `flux top`
+tool, and a critical fix for system deployments of Flux (#3958).
+
+### Fixes
+
+ * python: fix reference counting for Python Message objects (#3983)
+ * python: avoid early garbage collection of Watcher objects (#3975)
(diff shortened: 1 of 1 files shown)


grondo (approved):
LGTM!

garlick:
Bionic builders are failing with
```
2021-12-03T23:39:39.8818771Z configure: error: could not find python module jsonschema, version 2.3.0+ required
```
Just remembered that we still need to revert 902534657d1324227ee9c068d650c224c4e4a137 to drop the install of ncurses from checks/Dockerfile, so I'll push a PR for that.

grondo:
Something clearly went wrong with the bionic image I pushed up to Dockerhub. I'll have to investigate tomorrow :disappointed:

grondo:
Ok, the problem is that `pip install jsonschema` picked up a version that is too _new_ (apparently it doesn't have the `.__version__` attribute anymore, which causes the configure check to fail) Gotta love pip!

I'll fix this in the morning by pinning jsonschema to the older version in the Dockerfile. I'll also see if there is a better way to check the version of python modules in autoconf??

grondo:
I've pushed an update here to incorporate #3990 and #3991 assuming it can get merged.

grondo:
The fedora34 failure will be fixed when #3991 is merged.

garlick:
Setting MWP

grondo:
@mergifyio rebase

Issue #3982, opened 2021-12-01
python: unable to easily save message requests for later use


grondo:
@jameshcorbett reported an issue trying to respond to a request outside of the request message handler in a Python service-as-a-process. He boiled down the issue to this nice reproducer:

service:
```python
import flux
from flux.constants import FLUX_MSGTYPE_REQUEST

_SAVED_MESSAGES = []


def create_cb(fh, t, msg, arg):
    _SAVED_MESSAGES.append(msg)
    #fh.respond(msg, {"success": True})


def watch_cb(fh, *args, **kwargs):
    try:
        msg = _SAVED_MESSAGES.pop()
    except IndexError:
        pass
    else:
        fh.respond(msg, {"success": True})
        print("responded to message")
        pass


fh = flux.Flux()
serv_reg_fut = fh.service_register("dws")

w = fh.msg_watcher_create(
    create_cb, FLUX_MSGTYPE_REQUEST, "dws.create"
)
w.start()
serv_reg_fut.get()
timer_fh_watch = fh.timer_watcher_create(
    1, watch_cb, repeat=1
)
timer_fh_watch.start()
fh.reactor_run()
```

client:
```python
import flux


def cb(msg):
    print(msg.get())


h = flux.Flux()

h.rpc("dws.create", {"foo": "bar"}).then(cb)
h.reactor_run()
```

The problem is that, of course, the underlying C request message is destroyed once the callback returns. The workaround is to manually incref/decref the message, which in python requires `fh.msg_incref(msg)` `fh.msg_decref(msg)`.

However, it seems like the Python bindings should take a reference on the message for every `Message` object and let the Python garbage collector handle calling `decref`. Then Python programmers do not have to worry about how the underlying C object is managed -- as long as the Python `Message` instance is around, the underlying C `flux_msg_t` should still be valid.

grondo:
I'll see if I have the wherewithal to implement a solution :confused:

garlick:
Strangely, I observed no change in behavior with the reproducer when I modified `flux_msg_destroy()` to set the message type to `FLUX_MSGTYPE_ANY` or even zeroing the message struct before freeing.  Oh well maybe this will be clearer tomorrow.

grondo:
This is from the reproducer reworked in PR #3983, with a `flux_msg_fprintf()` in the `groups.get` response handler (at the site of the stray error message), in case it is helpful. (Assuming this is what you were referring to by the behavior of the reproducer):
```
[...]

PR #3983, opened 2021-12-02
python: fix reference counting for Python Message objects

grondo:
This appears to fix the issue described in #3982.

The `Message` class was probably written before `flux_msg_incref()`, and it tried to be smart about whether or not to "destruct" the underlying `flux_msg_t` on `Message` garbage collection. This led to possible use-after-free of `flux_msg_t` if a message is stashed from a msg_handler in Python for later use.

With the availablity of `flux_msg_incref()`, the handling of `Message` objects becomes more sane. We always call `flux_msg_decref()` when a `Message` is GC'd, and call `flux_msg_incref()` always, unless the `Message` constructor itself creates the underlying `flux_msg_t`.

### DIFF
--- a/src/bindings/python/flux/message.py
+++ b/src/bindings/python/flux/message.py
@@ -33,7 +33,6 @@ class Message(WrapperPimpl):
             self,
             type_id=flux.constants.FLUX_MSGTYPE_REQUEST,
             handle=None,
-            destruct=False,
         ):
             super(Message.InnerWrapper, self).__init__(
                 ffi,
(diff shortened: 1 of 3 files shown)


jameshcorbett (approved):
LGTM! I'm not very familiar with CFFI but it looks reasonable. I also checked that this fixes the problem in my more complicated setup (the one that I simplified into the reproducer).

grondo:
Thanks @jameshcorbett!

PR #3984, opened 2021-12-02
Support job memo events

grondo:
This PR ostensibly adds support for the new RFC 21 job `memo` events.

Currently, `memo` events are used to manager the `user` annotations within the job-manager and job-list modules. It will eventually be cleaner to separate the two, using "annotations" just for scheduler annotations, and job "memo" events for "user" memos, but I didn't go that far in this PR because it would probably be another week of work to disentangle the two.

This PR replaces `flux job annotate` and the associated `job-manager.annotate` RPC with `flux job memo` and a `job-manager.memo` service. This was done for better alignment with the concept of a `memo` event, but also because I was having a lot of trouble touching anything in the annotations code without breaking the tests that use `flux job annotate`, so it was easier to just remove all that and start over with new tests.

`flux job memo` has a slightly different interface in order to more closely match the interface of `flux kvs put`, i.e. it supports dotted "paths" for assignment, and multiple `key=value` pairs on the commandline in order to allow complex objects to be built up in a single command, e.g. `flux job memo JOBID a.b.c=42 a.d=foo`

Normally, `memo` requests sent to the job manager are posted to the job's eventlog, so that the memo can be recovered on restart, but the new `memo` RPC supports an optional `volatile` flag which bypasses the eventlog. This is meant for memos with high frequency updates, and which do not need to be recovered from the eventlog on restart.

Along the way, I noticed a problem with the `EVENT_NO_COMMIT` flag, which required a rework of these event flags in the job-manager. Since the job-list module assumes that the journal eventlog sequence number matches the virtual sequence number of events in the job eventlog (for restart), all events posted with `EVENT_NO_COMMIT` should not have a journal sequence number, *except* for the `submit` event, which is already in the eventlog at the time the job manager processes the event. Therefore, in order to use `EVENT_NO_COMMIT` for volatile `memo` events, a rework of these event flags was required. I expanded the comments on the flags to hopefully explain the situation, but in short `EVENT_JOURNAL_ONLY` is dropped and a new flag was added to force a sequence number for a `EVENT_NO_COMMIT` event (for the `submit` evnet)

A (probably good) side effect of the above is that jobtap plugins can now get a callback on events that were previously journal-only (`annotations` events).

Sorry for the rambling description. I thought I should probably explain the changes to the job-manager event handling in more detail to give some context.


src/cmd/flux-job.c
```diff
@@ -3023,6 +3039,76 @@ int cmd_annotate (optparse_t *p, int argc, char **argv)
...
+    id = parse_jobid (argv[optindex]);
+
+    /*  Build memo object from multiple values */
+    for (i = optindex + 1; i < argc; i++) {
+        void *valbuf = NULL;
+        char *key;
+        char *value;
+        json_t *val;
+
+        key = strdup (argv[i]);
```
garlick, line 3001: check `strdup()` for NULL

src/modules/job-manager/annotate.c
```diff
@@ -167,6 +167,54 @@ void annotate_handle_request (flux_t *h,
...
+                                        "memo", &memo) < 0
+        || flux_msg_get_cred (msg, &cred) < 0)
+        goto error;
+    if (!(job = zhashx_lookup (ctx->active_jobs, &id))) {
+        errstr = "unknown job id";
+        errno = ENOENT;
+        goto error;
+    }
+    if (flux_msg_cred_authorize (cred, job->userid) < 0) {
+        errstr = "guests can only add a memo to their own jobs";
```
garlick, line 157: s/add a memo to/memoize?

garlick (approved):
This looks really good to me!  I left a few minor comments inline.

src/common/libutil/jpath.c
```diff
@@ -155,4 +231,29 @@ json_t *jpath_get (json_t *o, const char *path)
...
+        errno = EINVAL;
+        return -1;
+    }
+    json_object_foreach_safe (o, tmp, key, value)  {
+        if (json_is_object (value)) {
+            if (jpath_clear_null (value) < 0)
+                return -1;
+            if (json_object_size (value) == 0
+                && json_object_del (o, key) < 0)
+                return -1;
```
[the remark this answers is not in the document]
grondo: Ah good catch. Actually `json_object_del()` only returns error if `key` is not found, which is not possible (and we don't care if it happens), so I'll just ignore the return code here. Sound ok?

src/common/libutil/jpath.c
```diff
@@ -155,4 +231,29 @@ json_t *jpath_get (json_t *o, const char *path)
...
+        }
+        else if (json_is_null (value)
+                 && json_object_del (o, key) < 0)
+                return -1;
```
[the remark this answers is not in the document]
grondo: Same as above.

src/common/libutil/jpath.c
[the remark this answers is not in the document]
garlick: Sure.

src/modules/job-manager/annotate.c
grondo: I actually thought about calling it memoize, but [memoize](https://en.wikipedia.org/wiki/Memoization) already has a different definition, which might make the resulting error message confusing (I thought anyway)

garlick: OK :-)  Your message is just fine and I should probably not have bothered you with that one.

grondo:
Ok, addressed @garlick's comments, added a couple more tests to the `flux job memo` sharness test, and squashed and pushed the result (hope that was ok)

PR #3985, opened 2021-12-02
libflux: improve safety against refcounting bugs in low-level message functions

garlick:
This PR extends the NULL safety check in many `flux_msg_*()` functions to include checking that the message reference count is nonzero.  I verified that the reproducer for #3982 no longer sends a misdirected response to the broker groups service (it just hangs now).  Hopefully these checks will make message refcounting bugs more straigtforward to track down.

There's also a bit of gratuitous cleanup in here (couldn't resist).


grondo (approved):
LGTM! Nice cleanup!

PR #3986, opened 2021-12-02
broker: record child instance URIs as job memo

grondo:
This PR (built on top of #3984) adds a feature to the broker to record the rank 0 URI as a job "memo" when the current instance is detected to be a Flux job.

This makes the URI available from `job-list` in the annotations attribute or from `flux jobs` as `{user.uri}`. This will make it more convenient to fetch sub-job URIs from job listing tools (and allows detection of Flux instances  in the job listing via presence of the user.uri annotation, rather than requiring a separate lookup from the job's kvs namespace.


src/broker/broker.c
```diff
@@ -834,6 +839,73 @@ static int init_local_uri_attr (struct overlay *ov, attr_t *attrs)
...
+    }
+    if (!(f = flux_rpc_pack (h,
+                             "job-manager.memo",
+                             FLUX_NODEID_ANY,
+                             0,
+                             "{s:I s:{s:s}}",
+                             "id", id,
+                             "memo",
+                               "uri", uri)))
+        return -1;
```
garlick, line 898: need to call `flux_close()` on `h` before returning

src/broker/broker.c
```diff
@@ -834,6 +839,73 @@ static int init_local_uri_attr (struct overlay *ov, attr_t *attrs)
...
+
+    if (flux_rpc_get (f, NULL) < 0) {
+        log_err ("job-manager.memo uri");
+        return -1;
```
garlick, line 902: call `flux_close(h)` and `flux_future_destroy (f)` before returning

garlick (approved):
LGTM!

Note that since #3979 was merged, [`open_flux_instance()`](https://github.com/flux-framework/flux-core/blob/89825bc430e816185004b9f06326663e127a3f9c/src/cmd/top/top.c#L129) should be updated.  Maybe it could be combined with the `job-list` query in `job_get_state()` right above it?

grondo:
> Note that since #3979 was merged, open_flux_instance() should be updated. Maybe it could be combined with the job-list query in job_get_state() right above it?

Yes that was the idea, you can list a job and get its uri (if a child instance) with a single call. When I do the rebase I'll make your suggested change.

src/broker/broker.c
grondo: Got it, good catch.

grondo:
I've pushed an update to this PR with the necessary change to flux-top's `open_flux_instance()`. Please have a look @garlick and ACK the changes there. (also addressed comments above and squashed the result)

garlick (approved):
The top changes look great, thanks!

garlick:
Clang is cranky about that change though:
```
2021-12-03T15:49:23.7894966Z   CC       top/flux_top-top.o
2021-12-03T15:49:23.7896159Z top/top.c:110:9: error: variable 'state' is used uninitialized whenever '||' condition is true [-Werror,-Wsometimes-uninitialized]
2021-12-03T15:49:23.7897263Z     if (!(f = flux_rpc_pack (h,
2021-12-03T15:49:23.7897602Z         ^~~~~~~~~~~~~~~~~~~~~~~
2021-12-03T15:49:23.7898068Z top/top.c:130:11: note: uninitialized use occurs here
2021-12-03T15:49:23.7898604Z     if (!(state & FLUX_JOB_STATE_RUNNING))
2021-12-03T15:49:23.7898968Z           ^~~~~
2021-12-03T15:49:23.7899621Z top/top.c:110:9: note: remove the '||' if its condition is always false
2021-12-03T15:49:23.7900120Z     if (!(f = flux_rpc_pack (h,
2021-12-03T15:49:23.7900476Z         ^~~~~~~~~~~~~~~~~~~~~~~
2021-12-03T15:49:23.7901165Z top/top.c:108:14: note: initialize the variable 'state' to silence this warning
2021-12-03T15:49:23.7901671Z     int state;
2021-12-03T15:49:23.7901978Z              ^
2021-12-03T15:49:23.7902252Z               = 0
2021-12-03T15:49:23.7902596Z 1 error generated.
2021-12-03T15:49:23.7903255Z Makefile:1048: recipe for target 'top/flux_top-top.o' failed
```

grondo:
Thanks, I'll fix that.

grondo:
The issue was likely because the `fatal()` function wasn't marked as "noreturn". I've added the appropriate attribute. We'll see if that fixes the clang errors.

grondo:
Setting MWP.

PR #3989, opened 2021-12-04
Revert "docker: add ncurses to checks/Dockerfile"

garlick:
This reverts commit 902534657d1324227ee9c068d650c224c4e4a137.

PR #3991, opened 2021-12-05
docker: fix and update ci dockerfiles

grondo:
During rebuild of docker images for #3989 wonderful `pip` in the bionic image pulled in latest `jsonschema`, which broke the build since `configure` can't detect the version for that module after 4.0 since they removed the `__version__` attribute on the module. Additionally, the newest cffi module no longer built with python3.7 because it now appears to require `libffi-dev`.

Additionally, I noticed that `src/test/docker/README.md` hadn't been updated with instructions for how to build the multiarch `bionic` image, which requires `docker buildx`.

This PR includes the fix for #3989. It seems like that commit can also fit in this PR to speed up the merge of these fixes.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -14,7 +14,7 @@ queue_rules:
       - status-success="bionic - test-install"
       - status-success="focal - py3.8"
       - status-success="centos7"
-      - status-success="centos8 - py3.7"
+      - status-success="centos8"
       - status-success="fedora33 - gcc-10,py3.9"
       - status-success="fedora34 - gcc-11.2,py3.9"
       - status-success="coverage"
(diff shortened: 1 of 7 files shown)


grondo:
It would be good to get this in before we tag so that the extra apt-update from the checks builder isn't present in the final docker images produced by the tag, but that is pretty minor.

The resulting docker images from these changes have already been pushed up to docker hub.

grondo:
Ok, maybe something still isn't quite right here since tests are failing. I'll keep looking and mark this WIP.

grondo:
Ok, I've worked through the problems and found a couple more issues due to the updated container images.
This should be good to go now.

garlick (approved):
Thanks for working it out @grondo!

grondo:
Thanks, I've set MWP

grondo:
Ugh. Forgot to update mergify config. Pushed that here but now this will need a manual merge.

grondo:
Thanks!

PR #3997, opened 2021-12-07
connectors: avoid embedded synchronous RPC for subscribe/unsubscribe

garlick:
Problem: connectors have event subscribe/unsubscribe "operations" from early times when 0MQ was the primary transport, but even in the one extant 0MQ-based connector (shmem), the operations make hidden synchronous RPCs.

In addition to being unnecessary, the embedded RPCs are inflexible. If handled like other RPCs, then:
- subscriptions can be made asynchronously, where appropriate
- subscriptions can be made with FLUX_RPC_NORESPONSE, where appropriate
- In the future, RPC tracking for automatic reconnect when a broker restarts could track subscription RPCs like any other

This PR changes `flux_event_subscribe()` and `flux_event_unsubscribe()` to wrap a synchronous RPC rather than call a connector operation that might wrap a synchronous RPC or might be a no-op.  I figured it was safest to leave the semantics of these functions unchanged, rather than try to change their prototype to return a future.    Instead, two new functions `flux_event_subscribe_ex()` and `flux_event_unsubscribe_ex()` are added.  In addition to returning futures, these functions accept RPC flags.

There was a bit of special care needed to avoid deadlock in the broker, and in loopback test scenarios.  Those are documented in commit messages, and IMHO not concerning because they are not representative of our primary use cases for pub/sub.   A new handle flag FLUX_O_TEST_NOSUB was added to avoid blocking RPCs to self in a loopback test.  Not usually required - only when events are part of a test, synchronous RPCs are used, and the test doesn't install RPC handlers for `event.subscribe` and `event.unsubscribe`.

### DIFF
--- a/doc/man3/flux_open.rst
+++ b/doc/man3/flux_open.rst
@@ -31,6 +31,9 @@ the value of $FLUX_URI is used, if set.
 FLUX_O_TRACE
    Dumps message trace to stderr.
 
+FLUX_O_CLONE
+   Used internally by ``flux_clone()`` (see below).
+
 FLUX_O_MATCHDEBUG
    Prints diagnostic to stderr when matchtags are leaked, for example when
    a streaming RPC is destroyed without receiving a error response as
--- a/doc/man3/flux_open.rst
+++ b/doc/man3/flux_open.rst
@@ -42,6 +42,11 @@ FLUX_O_MATCHDEBUG
 FLUX_O_NONBLOCK
    The ``flux_send()`` and ``flux_recv()`` functions should never block.
 
+FLUX_O_TEST_NOSUB
+   Make ``flux_event_subscribe()` and ``flux_event_unsubscribe()`` no-ops.
+   This may be useful in specialized situations with the ``loop://`` connector,
+   where no message handler is available to service subscription RPCs.
+
 ``flux_clone()`` creates another reference to a ``flux_t`` handle that is
 identical to the original in all respects except that it does not inherit
 a copy of the original handle's "aux" hash, or its reactor and message
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -357,13 +357,6 @@ int main (int argc, char *argv[])
         goto cleanup;
     }
 
-    /* Create content cache.
-     */
-    if (!(ctx.cache = content_cache_create (ctx.h, ctx.attrs))) {
-        log_err ("content_cache_create");
-        goto cleanup;
-    }
-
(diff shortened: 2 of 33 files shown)


src/broker/broker.c
```diff
@@ -357,13 +357,6 @@ int main (int argc, char *argv[])
         goto cleanup;
     }
 
-    /* Create content cache.
```
grondo, line 361: Commit message typo `Add flux_event_subsribe_ex()`

src/common/libflux/event.c
```diff
@@ -22,6 +22,68 @@
...
 #include "message.h"
 
+flux_future_t *flux_event_subscribe_ex (flux_t *h,
+                                        const char *topic,
+                                        int flags)
+{
+    if (!h || !topic) {
+        errno = EINVAL;
+        return NULL;
+    }
```
grondo, line 32: If `flux_rpc_pack(3)` checks `h` and `topic` then this check could be dropped.

src/common/libflux/event.c
```diff
@@ -22,6 +22,68 @@
...
+}
+
+flux_future_t *flux_event_unsubscribe_ex (flux_t *h,
+                                          const char *topic,
+                                          int flags)
+{
+    if (!h || !topic) {
+        errno = EINVAL;
+        return NULL;
+    }
```
grondo, line 63: similar to above, if `flux_rpc_pack()` also checks args we're checking twice here (no harm though)

doc/man3/flux_open.rst
```diff
@@ -42,6 +42,11 @@ FLUX_O_MATCHDEBUG
 FLUX_O_NONBLOCK
    The ``flux_send()`` and ``flux_recv()`` functions should never block.
 
+FLUX_O_TEST_NOSUB
```
grondo, line 45: commit message typo: "flag is not document." :arrow_right: "flag is not documented."

grondo (approved):
LGTM! Just a few comments inline, none of which are critical.

src/common/libflux/event.c
garlick: h is checked but topic is not (it's not the rpc topic but part of the payload).
i'll drop the h check.

garlick:
Pushed those changes, and added some unit tests.  A lot of the diff here is error handling code that is hard to cover though.

grondo:
:+1: Feel free to set MWP as far as I'm concerned.

garlick:
OK thanks!

Issue #3900, opened 2021-10-05
handle: need a way to "reconnect" a flux_t handle


grondo:
As part of the solution to #3801, the job shell will need to attempt to reconnect to a local broker after it is in offline mode. After a reconnect, established service names will need to be re-registered/updated along with any subscriptions.

@garlick had proposed a flag on `flux_open(3)` that would request a disconnected handle be automatically reconnected. Maybe the service registration and subscriptions could be automatically handled in this case?

For the job shell, we may want to block when waiting to reconnect, so that other shell reactor watchers are not active during offline mode, but some users may want to asynchronously reconnect, so consider that in the design.

The host program will probably want some sort of notification that a reconnect has occurred. This could be handled with a specific callback somehow, or the handle could send a message and the program could establish a message watcher for whatever topic we decide.

garlick:
Thanks for opening this @grondo!  A couple of thoughts:
- maybe `FLUX_O_RECONNECT` vs (later) `FLUX_O_RECONNECT | FLUX_O_NONBLOCK`?
- the connector (client side) supporting `FLUX_O_RECONNECT` could employ [rpctrack.c](https://github.com/flux-framework/flux-core/blob/master/src/common/librouter/rpc_track.h) to ensure that any pending RPCs receive appropriate errors upon reconnect
- need to send disconnect messages for those RPCs also, in case entire instance was not restarted and the service provider still holds state for the rpc
- we will then need to deal with the fallout of any failing RPCs in the shell

grondo:
Great input!

> need to send disconnect messages for those RPCs also, in case entire instance was not restarted and the service provider still holds state for the rpc

This is probably a dumb idea, but is there no way to update routes to clients that are reconnecting and therefore have them essentially restart where they left off, rather than needing to restart all RPCs which establish state in the server? (i.e. a "reconnect" event similar to "disconnect") In general, it seems like reestablishing all that state on the client side will be pretty onerous (especially when dealing with things like the eventlog watch which will replay all events that have already been seen).

Hm, maybe for the shell the only streaming RPC that would have state is the input eventlog watcher, so for our current use case, maybe not a big deal.

garlick:
The trouble is that streaming response messages may be lost if a router hop restarts (including the router the shell is connected to).  I think the safe thing to do is to fail them, then the problem is a big obvious one not a tricky little one.

For stdin maybe we could just track the number of eventlog entries already consumed and ignore that many on restart?

[...]

PR #3998, opened 2021-12-10
WIP: add rlocal:// connector that can persist over broker restart

garlick:
This is an experimental connector that tolerates a broker restart, as discussed in #3900.   When the connection is lost, any pending RPCs get error responses.  The reconnect attempts are synchronous, effectively putting a reactive client to sleep until connection is successful.  There is more detail in the commit message that adds the connector.

There is also some minor cleanup in other connectors here.

I'll add some further observations, but wanted to get this parked before going offline to travel tomorrow.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -521,6 +521,7 @@ AC_CONFIG_FILES( \
   src/shell/Makefile \
   src/connectors/Makefile \
   src/connectors/local/Makefile \
+  src/connectors/rlocal/Makefile \
   src/connectors/shmem/Makefile \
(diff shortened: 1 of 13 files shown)


garlick:
I just figured out that this is not quite the right approach.  The librouter code for reading and writing the connector's fd is not compatible with it being wrapped in an epoll fd.  However, there is already an epollfd  at the flux_t level, so I think I can maybe add a connector callback for reconnect that swaps out the disconnected fd for a new one at that level.  So closing this.

PR #3999, opened 2021-12-10
add FluxURIResolver Python class and flux-uri command for job URI discovery

grondo:
This PR introduces a `FluxURIResolver` Python class which can resolve target URIs to Flux job URIs. Support for resolving URIs in different contexts is provided by resolver "plugins" which are loaded based on the target URI _scheme_, though if no _scheme_ is provided, then a scheme of `jobid` is assumed.

A new `flux-uri(1)` utility is provided to access this Python class from the commandline.

The PR includes three resolver plugins:

 * `jobid`: resolve job URIs for a Flux jobid in the current instance. This resolver works recursively if a "path" of jobids is provided, e.g. `jobid:f1234/f3456` will attempt to resolve the URI for jobid `f3456` running as  child job of `f1234`. Since `jobid` is the default URI resolver scheme, `f1234/f3456` would also work. e.g.
   ```console
   $ flux uri ƒR3S77XQb
   ssh://quartz18/var/tmp/grondo/flux-jJ8CaZ/local-0
   ```
 * `pid:` resolve a URI for a local process id. This plugin attempts to read the `FLUX_URI` value from `/proc/PID/environ`. As a convenience, if the process appears to be a `flux-broker`, then the `FLUX_URI` of one of its children is used. This allows determination of the URI for a running flux-broker by providing its PID (useful in implementation of other resolvers)
   ```console
   $ flux uri pid:$(pidof flux-broker)
   local:///tmp/flux-TBU7RM/local-0
   ```
 * `slurm:` an experimental resolver for Flux instances run as Slurm jobs. The `slurm` plugin works by using `srun` to run `scontrol listpids` on the first node of a Slurm job. For each PID that is also a direct child of slurmstepd (i.e. localid == 0) the `pid` resolver is used to resolve the FLUX_URI locally.
   ```console
   $ squeue -u grondo
             JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
           7843494    pdebug interact   grondo  R      55:36      2 quartz[18-19
   $ flux uri slurm:7843494
   ssh://quartz18/var/tmp/grondo/flux-MpnytT/local-0
   ```

URI resolver plugins are loaded from the `flux.uri.resolvers` namespace, so it is trivial to add third party or override existing resolvers on systems where needed. E.g. I assume someone could provide an `lsf` plugin at some point.

The next step is to extend `flux proxy` to call out to `flux uri` if its argument is not already a `local://` or `ssh://` URI. I actually have this working, but figured that would be better for a separate PR, release notes wise.

Here's a preview:
```console
$ src/cmd/flux proxy slurm:7843731 flux resource list
      free      4       72        0 quartz[1,1-2,2]
      down      0        0        0 

```

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -46,8 +46,28 @@ source_suffix = '.rst'
 
 extensions = [
     'sphinx.ext.intersphinx',
-    'sphinx.ext.napoleon'
+    'sphinx.ext.napoleon',
+    'domainrefs'
 ]
+
+domainrefs = {
+    'man1': {
+        'text': "%s(1)",
+        'url': "../man1/%s.html"
+    },
+    'man3': {
+        'text': "%s(3)",
+        'url': "../man3/%s.html"
+    },
+    'man5': {
+        'text': "%s(5)",
+        'url': "../man5/%s.html"
+    },
+    'man7': {
+        'text': "%s(7)",
+        'url': "../man7/%s.html"
+    }
+}
 
 # Disable "smartquotes" to avoid things such as turning long-options
 #  "--" into en-dash in html output, which won't make much sense for
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -147,6 +147,7 @@ man_pages = [
     ('man1/flux-hwloc', 'flux-hwloc', 'Control/query resource-hwloc service', [author], 1),
     ('man1/flux-jobs', 'flux-jobs', 'list jobs submitted to Flux', [author], 1),
     ('man1/flux-jobtap', 'flux-jobtap', 'List, remove, and load job-manager plugins', [author], 1),
+    ('man1/flux-uri', 'flux-uri', 'resolve Flux URIs', [author], 1),
     ('man1/flux-keygen', 'flux-keygen', 'generate keys for Flux security', [author], 1),
     ('man1/flux-kvs', 'flux-kvs', 'Flux key-value store utility', [author], 1),
     ('man1/flux-logger', 'flux-logger', 'create a Flux log entry', [author], 1),
--- a/doc/man1/flux-uri.rst
+++ b/doc/man1/flux-uri.rst
@@ -1,0 +1,147 @@
+===========
+flux-uri(1)
+===========
+
+
+SYNOPSIS
+========
+
+**flux** *uri* [OPTIONS] *TARGET*
+
+DESCRIPTION
+===========
+
+Connections to Flux are established via a Uniform Resource Indicator
+(URI) which is passed to the :man3:`flux_open` API call. These *native*
+URIs indicate the "connector" which will be used to establish the
+connection, and are typically either *local*, with a  ``local`` URI
+scheme, or *remote*, with a ``ssh`` URI scheme. These URIs are considered
+fully-resolved, native Flux URIs.
+
+Processes running within a Flux instance will have the ``FLUX_URI``
+environment variable set to a native URI which :man3:`flux_open` will
+use by default, with fallback to a compiled-in native URI for the system
+instance of Flux. Therefore, there is usually no need to specify a URI when
+connecting to the enclosing instance. However, connecting to a *different*
+Flux instance will require discovery of the fully-resolved URI for that
+instance.
+
+**flux uri** attempts to resolve its *TARGET* argument to a native local
+or remote URI. The *TARGET* is itself a URI which specifies the method
+to use in URI resolution via the scheme part, with the path and query
+parts passed to a plugin which implements the resolution method.
+
+As a convenience, if *TARGET* is specified with no scheme, then the scheme
+is assumed to be ``jobid``.  This allows ``flux uri`` to be used to look
+up the URI for a Flux instance running as a job in the current enclosing
+instance with:
+
+::
+
+   $ flux uri JOBID
+
+Depending on the *TARGET* URI scheme and corresponding plugin, specific
+query arguments int *TARGET* may be supported. However, as a convenience,
+all *TARGET* URIs support the special query arguments ``local`` or
+``remote`` to force the resulting URI into a local (``local://``) or remote
+(``ssh://``) form. For example:
+
+::
+
+   $ flux uri JOBID?local
+
+would return the ``local://`` URI for *JOBID* (if the URI can be resolved).
+
+A list of supported URI schemes will be listed at the bottom of
+``flux uri --help`` message. For a description of the URI resolver schemes
+included with Flux, see the URI SCHEMES and EXAMPLES sections below.
+
+OPTIONS
+=======
+
+**-remote**
+   Return the *remote* (``ssh://``)  equivalent of the resolved URI.
+
+**--local**
+   Return the *local* (``local://``) equivalent of the resulved URI.
+   Warning: the resulting URI may be invalid for the current system
+   if the network host specified by an ``ssh`` URI is not the current
+   host.
+
+URI SCHEMES
+===========
+
+The following URI schemes are included by default:
+
+jobid:ID[/ID...]
+   This scheme attempts to get the URI for a Flux instance running as a
+   job in the current enclosing instance. This is the assumed scheme if no
+   ``scheme:`` is provided in *TARGET* passed to ``flux uri``, so the
+   ``jobid:`` prefix is optional. A hierarchy of Flux jobids is supported,
+   so ``f1234/f3456`` will resolve the URI for job ``f3456`` running in
+   job ``f1234`` in the current instance.
+
+pid:PID
+  This scheme attempts to read the ``FLUX_URI`` value from the process id
+  *PID* using ``/proc/PID/environ``. If *PID* refers to a ``flux-broker``,
+  then the scheme reads ``FLUX_URI`` from the broker's initial program or
+  another child process since ``FLUX_URI`` in the broker's environment
+  would refer to *its* parent (or may not be set at all in the case of a
+  test instance started with ``flux start --test-size=N``).
+
+slurm:JOBID
+  This scheme makes a best-effort to resolve the URI of a Flux instance
+  launched under Slurm. It invokes ``srun`` to run ``scontrol listpids``
+  on the first node of the job, and then uses the ``pid`` resolver until
+  it finds a valid ``FLUX_URI``.
+
+
+EXAMPLES
+========
+
+To get the URI of a job in the current instance in its ``local://`` form:
+
+::
+
+   $ flux uri --local ƒN8Pz2xVu
+   local:///tmp/flux-zbVtVg/jobtmp-0-ƒN8Pz2xVu/flux-59uf5w/local-0
+
+or
+
+::
+
+   $ flux uri ƒN8Pz2xVu?local
+   local:///tmp/flux-zbVtVg/jobtmp-0-ƒN8Pz2xVu/flux-59uf5w/local-0
+
+
+Get the URI of a nested job:
+
+::
+
+   $ flux uri ƒqxxTiZBM/ƒr2XFWP?local
+   local:///tmp/flux-zbVtVg/jobtmp-0-ƒqxxTiZBM/flux-EPgSwk/local-0
+
+.. note::
+   With  the ``jobid`` resolver, ``?local`` only needs to be placed on
+   the last component of the jobid "path" or hierarchy. This will resolve
+   each URI in turn as a local URI.
+
+Get the URI of a local flux-broker
+
+::
+
+   $ flux uri pid:$(pidof -s flux-broker)
+   local:///tmp/flux-sLuBkZ/local-0
+
+Get the URI for a Flux instance running as a Slurm job:
+
+::
+
+   $ flux uri slurm:7843494
+   ssh://cluster42/var/tmp/user/flux-MpnytT/local-0
+
+
+RESOURCES
+=========
+
+Github: http://github.com/flux-framework
--- a/src/bindings/python/flux/Makefile.am
+++ b/src/bindings/python/flux/Makefile.am
@@ -44,6 +44,7 @@ nobase_fluxpy_PYTHON = \
 	uri/__init__.py \
 	uri/resolvers/jobid.py \
 	uri/resolvers/pid.py \
+	uri/resolvers/slurm.py \
 	utils/parsedatetime/__init__.py \
 	utils/parsedatetime/parsedatetime.py \
 	utils/parsedatetime/warns.py \
--- a/src/bindings/python/flux/Makefile.am
+++ b/src/bindings/python/flux/Makefile.am
@@ -9,6 +9,7 @@ nobase_fluxpy_PYTHON = \
 	future.py \
 	memoized_property.py \
 	debugged.py \
+	importer.py \
 	core/__init__.py \
 	core/watchers.py \
 	core/inner.py \
--- a/src/bindings/python/flux/Makefile.am
+++ b/src/bindings/python/flux/Makefile.am
@@ -42,6 +42,7 @@ nobase_fluxpy_PYTHON = \
 	progress.py \
 	uri/uri.py \
 	uri/__init__.py \
+	uri/resolvers/jobid.py \
 	utils/parsedatetime/__init__.py \
 	utils/parsedatetime/parsedatetime.py \
 	utils/parsedatetime/warns.py \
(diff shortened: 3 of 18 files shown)


grondo:
BTW, `flux-uri` lists available plugins in `--help` output:
```console
$ flux uri  --help
usage: flux-uri [-h] [--remote] [--local] TARGET

Resolve TARGET to a Flux URI

positional arguments:
  TARGET      A Flux jobid or URI in scheme:argument form (e.g. jobid:f1234)

optional arguments:
  -h, --help  show this help message and exit
  --remote    convert a local URI to remote
  --local     convert a remote URI to local

Supported resolver schemes:
  jobid		Get URI for a given Flux JOBID
  pid		Get FLUX_URI for a given local PID
  slurm		Get URI for a Flux instance launched under Slurm

```
it can also be used as a trivial way to convert local URIs to remote and vice versa:
```console
$ flux uri --remote $FLUX_URI
ssh://asp/tmp/flux-TBU7RM/local-0
$ flux uri --local ssh://asp/tmp/flux-TBU7RM/local-0
local:///tmp/flux-TBU7RM/local-0
```

garlick:
This is a nice addition to our tool set!   Couple of quick comments:

It might be good to add examples of input accepted by the built-in methods to the man page.

I would expect this to be commonly used, so could we add the magic `flux-help-command` comment to the man rst source so this pops up in `flux help` output?

Should this parse?
```
$ flux uri pid://1234
flux-uri: ERROR: [Errno 2] No such file or directory: '/proc////status'
```
(I expected the `//` to be optional when the authority is missing, but I could be mistaken about that.)

In another PR we could drop the jobid resolving code from `flux top` and have it popen `flux-uri` instead.   Then `flux top slurm:JOBID` just works.  Nice!  On that note, it would be handy if the`--local` and `--remote` options could be specified in the URI as query or fragment, e.g. `pid:1234?local`.

grondo:
> Should this parse?

No, `pid://1234` is a URI with authority `1234`, the authority is only "empty" when there is a following slash as in the `local` scheme `local:///tmp/foo/bar`

grondo:
> I would expect this to be commonly used, so could we add the magic flux-help-command comment to the man rst source so this pops up in flux help output?

Sure, though I was thinking that those need a holistic redo since we still reference lesser used commands like `keygen` and `logger`

grondo:
> In another PR we could drop the jobid resolving code from flux top and have it popen flux-uri instead.

Would it be better to encourage `flux proxy URI flux top` instead?

In any event, in the next PR I'll add a helper function to call out to `flux uri` and we can add it to commands that make sense.

grondo:
>  On that note, it would be handy if the--local and --remote options could be specified in the URI as query or fragment, e.g. jobid:1234?local.

Ooh, great idea!

garlick:
> Would it be better to encourage `flux proxy URI flux top` instead?

Eh, I'd prefer to call your helper :-)

> Sure, though I was thinking that those need a holistic redo since we still reference lesser used commands

True.  I'll open an issue on that.

garlick:
> Would it be better to encourage flux proxy URI flux top instead?

The problem with encouraging `flux-proxy` for running a single command is it adds an unnecessary layer of message routing, which feels slightly wrong to me when a simple environment modification and exec would suffice.

What about adding a `--remote=URI` option to `flux(1)`?

grondo:
> The problem with encouraging flux-proxy for running a single command is it adds an unnecessary layer of message routing, which feels slightly wrong to me when a simple environment modification and exec would suffice.

Ah, that should have been obvious, sorry. I was thinking of the case of FLUX_URI set in a shell session with many commands.

> What about adding a --remote=URI option to flux(1)?

I'm not sure how exactly that would work, but might be more user friendly to selectively extend commands that can take a URI to allow resolver URIs instead of just Flux job URIs. (I could just be missing the use case for `flux --remote=URI` -- how is it different from `FLUX_URI=URI flux...`?)

However, what if `flux proxy` worked more like `tmux` or `screen` or ssh control sockets? Instead of opening a random socket, flux-proxy could open a socket that uniquely identifies the remote instance. Then if another process has FLUX_URI set to the same remote, the proxy connection could be reused?

garlick:
I was just thinking `flux --remote URI  ...` would be shorthand for `FLUX_URI=$(flux uri URI) flux ...` (a bit less typing), but was  just brainstorming.   Maybe after we have `flux uri` merged, certain commands (like `flux jobs`?) may emerge as frequently used with a `FLUX_URI=` prefix, and then we can decide whether it makes sense to add a convenience option to them.

The tmux idea is great!  Now I think I've forgotten how `flux terminus` works. I don't want to get us too far off topic in this PR and I have to hit the road soon, so I'll just take that thought "to go". :-)

grondo:
I've pushed an update here that adds support for common `?local` `?remote` query params on the URI resolver URIs passed to `flux uri` and `FluxURIResolver` as suggested.

If the final id in a `jobid:` lookup contains `?local`, then all intermediate jobid URIs will be resolved to `local:` URIs, so this can be used to resolve a nested job without `ssh` (useful in testing)

The manpage was also expanded with a description of all 3 included resolver plugins, and examples of each.

Finally, a few more tests were added to cover the new code for handing query parameters.

grondo:
> The tmux idea is great! Now I think I've forgotten how flux terminus works. I don't want to get us too far off topic in this PR and I have to hit the road soon, so I'll just take that thought "to go". :-)

I actually shouldn't have mentioned tmux, as I was thinking more along the lines of how ssh control sockets work. When you ssh to host, if control sockets are enabled ssh first looks for a socket in a well known location based on the the target host and options (like username, port, etc). Could the `ssh` connector do something similar to re-use an existing proxy connector before launching another ssh? If one doesn't exist, flux-proxy could be started running daemonized.

grondo:
I'd actually like to clean up the terminology a bit here and could use some help.

I've been calling the URI schemes into which resolver URIs are resolved (i.e. `ssh` and `local` for now) job URIs, but that probably isn't correct, since a `flux start --test-size` instance isn't a job, but it has a URI. Should these be called "connector URIs" or something else more specific?

It gets a little confusing talking about URIs that resolve to other URIs. I wish we had a better name for the URI argument to `flux uri`. If there's any suggestions, I could clean up the docs in this PR. Additionally, maybe an RFC describing "URI resolver URIs" (or something) would help here?

garlick:
I guess the distinction is that the former can be passed to `flux_open(3)` while the latter has to be resolved first (kind like domain names).

Probably not great, but could we call the new one a URN and the original a URL (both URIs)?

Another thought, is there really any reason why we couldn't allow the new URIs to be passed to `flux_open(3)`?  It's just more schemes, so could we fall through to the proposed C helper if `flux_open(3)` can't find a scheme to dlopen directly...

grondo:
> Another thought, is there really any reason why we couldn't allow the new URIs to be passed to flux_open(3)? It's just more schemes, so could we fall through to the proposed C helper if flux_open(3) can't find a scheme to dlopen directly...

Well, there's a `popen2()` call which might be surprising and could possibly cause problems in some cases, but maybe it could be opt-in via a new `flux_open(3)` flag? It would also be nice to encourage Python users of `flux.Flux()` to resolve the URI first to avoid calling in to `C` to `popen()` another Python process.

garlick:
I think this PR could be independent of adding name resolution to `flux_open(2)`, which seems like it requires more thought, if it should be done at all.

One quick thought: do you think renaming the comand to `flux-resolve-uri` would clarify its purpose?

On terminology, maybe "high level" or "unresolved" Flux URIs get resolved to "native" Flux URIs?

grondo:
> I think this PR could be independent of adding name resolution to flux_open(2), which seems like it requires more thought, if it should be done at all.

:+1:

> One quick thought: do you think renaming the comand to flux-resolve-uri would clarify its purpose?

Yeah, good thought. I had originally called it `flux-uri-resolve` but thought the shorter command would be appreciated when used in scripts and such. Now that you mention it, since the command will be more of a helper or plumbing command it doesn't make sense to drop the verb, so I'll add it back.

> On terminology, maybe "high level" or "unresolved" Flux URIs get resolved to "native" Flux URIs?

Yeah, I like that and it will do for now.

grondo:
> I would expect this to be commonly used, so could we add the magic flux-help-command comment to the man rst source so this pops up in flux help output?

After discussion it seems like this would be more of a plumbing script, so I'll leave it out of the automated `flux help` generation for now.

grondo:
Ok, I've pushed some fixup commits that rename `flux uri` to `flux uri-resolve`, though I have to say after doing that, I miss the shorter command (it is like a literal "flux URI for JOBID" or "flux uri for PID") but no big deal either way.

garlick:
I am fine both ways too.  If you feel that shorter is better here feel free to revert that fixup.

grondo:
I actually ended up keeping the shorter name for now (easy to change if we decide against it). Even though `flux uri` itself might not be commonly used by normal users, I added some introductory text about "URIs"  to the `flux-uri(1) ` manpage, and it seemed like it would be nice to reference this from other locations as "see flux-uri(1)".

Another idea would be to add a `flux uri resolve` subcommand to `flux uri` if the shorter name `flux uri` is too confusing. Then the intro to Flux URIs could still be in `flux-uri(1)`.

The intro was written a little hastily and still needs improvement, but at least it is somewhere.

grondo:
Oh, I also got annoyed at how awkward it is to reference a manpage from in the reStructuredText. I found a random "extension" to easily add new domain refs to sphinx, and added special `manN` refs using a new `domainrefs` setting in `conf.py`. Now you can reference a manpage with, e.g. `` :man3:`flux_open` ``

doc/man1/flux-uri.rst
```diff
@@ -0,0 +1,144 @@
...
+included with Flux, see the URI SCHEMES and EXAMPLES sections below.
+
+OPTIONS
+=======
+
+**-remote**
+   Return the _remote_ (``ssh://``)  equivalent of the resolved URI.
+
+**--local**
+   Return the _local_ (``local://``) equivalent of the resulved URI.
```
garlick, line 64: Use rst emphasis not markdown for "local" and "remote"

doc/man1/flux-uri.rst
```diff
@@ -0,0 +1,144 @@
...
+   ``scheme:`` is provided in *TARGET* passed to ``flux uri``, so the
+   ``jobid:`` prefix is optional. A hierarchy of Flux jobids is supported,
+   so ``f1234/f3456`` will resolve the URI for job ``f3456`` running in
+   job ``f1234`` in the current instance.
+
+pid:PID
+  This scheme attempts to read the ``FLUX_URI`` value from the process id
+  *PID* using ``/proc/PID/environ``. If *PID* refers to a ``flux-broker``,
+  then a child of the broker is used instead in order to obtain the
+  URI for that broker (instead of its parent)
```
garlick, line 86: Maybe instead of "then a child of the broker is used instead" say "then the scheme reads FLUX_URI from the broker's initial program or another child process since FLUX_URI in the broker's environment would refer to _its_ parent.

I'm not sure if that makes it any clearer.  Feel free to ignore.

Missing period at end of sentence.

doc/man1/flux-uri.rst
```diff
@@ -0,0 +1,144 @@
...
+::
+
+   $ flux uri ƒqxxTiZBM/ƒr2XFWP?local
+   local:///tmp/flux-zbVtVg/jobtmp-0-ƒqxxTiZBM/flux-EPgSwk/local-0
+
+.. note::
+   With  the ``jobid`` resolver, ``?local`` only needs to be placed on
+   the last component of the jobid "path" or hierarchy. This will resolve
+   each URI in turn as a local URI, so it is only useful if all jobs
+   are actually running on the local system.
```
garlick, line 124: If the last job is running on the local system, then all enclosing instances have to be running on the local system, so maybe the "so..." explanation could just be dropped?

doc/man1/flux-uri.rst
```diff
@@ -0,0 +1,144 @@
...
+
+Get the URI of a local flux-broker
+
+::
+
+   $ flux uri pid:$(pidof flux-broker | cut -d' ' -f1)
```
garlick, line 130: maybe use `pidof -s` instead of the cut?

garlick (commented):
Here are some notes on the man page.  I'll make a pass through the code and then finish my review but wanted to get you these first.

src/bindings/python/flux/Makefile.am
```diff
@@ -9,6 +9,7 @@ nobase_fluxpy_PYTHON = \
 	future.py \
```
garlick, line 9: commit message: s/validatory.py/validator.py/

doc/man1/flux-uri.rst
grondo: Also added "(or may not be set at all in the case of an instance started with `flux start --test-size=N`)"

src/bindings/python/flux/Makefile.am
```diff
@@ -40,6 +40,8 @@ nobase_fluxpy_PYTHON = \
 	hostlist.py \
```
garlick, line 40: commit message s/hiearchical/hierarchical/

> if a URI is resolved with no scheme, it is assumed to be a Flux jobid

Did you mean "if a URI is parsed with no scheme..."?  And is the default scheme an optimization or a convenience?

doc/man1/flux-uri.rst
grondo: Yeah, that sounds better, though more of what I was going for is that in resolving each job along the way the jobid resolver plugin has to connect to the parent job to resolve its child, and when `?local` is placed on the last jobid _all_ connections will use the `local` connector. This may not work if for example the first jobid is on another host (the first `flux_open()` will fail).

Also, jobid URIs are resolved to the URI of rank 0. If not all enclosing instances have rank 0 running locally, then your statement, while true, doesn't apply to resolution of URIs.

However, perhaps that is too much detail for the manpage.

doc/man1/flux-uri.rst
grondo: Cool, didn't know about `pidof -s`

doc/conf.py
```diff
@@ -46,9 +46,29 @@
 
 extensions = [
     'sphinx.ext.intersphinx',
-    'sphinx.ext.napoleon'
+    'sphinx.ext.napoleon',
+    'domainrefs'
```
garlick, line 50: This is awesome!

src/bindings/python/flux/Makefile.am
```diff
@@ -44,6 +44,7 @@ nobase_fluxpy_PYTHON = \
 	uri/__init__.py \
```
garlick, line 44: Commit message: s/direclty/directly/

garlick (approved):
Just a few superficial comments - I didn't spot anything in the code to comment in.

Great work to improve our usability here.

doc/man1/flux-uri.rst
garlick: Up to you, but if you leave it in you might want to include the rank 0 requirement.

> If not all enclosing instances have rank 0 running locally

The top enclosing instance wouldn't have to be local, correct?

src/bindings/python/flux/Makefile.am
grondo: > And is the default scheme an optimization or a convenience?

Yeah, a convenience. I'll just drop the `As an optimization,` -- no need for filler phrases in the commit message

doc/man1/flux-uri.rst
grondo: No, I just left it out.

> The top enclosing instance rank 0 isn't required to be local, correct?

no I meant the first jobid in the path component, it will be forced to a `local://` uri if `?local` is used. Probably the only use case for `?local` is developers working with test instances, so it is probably better to just sweep this under the rug for now.

grondo:
Seeing some unexpected errors in the centos8 builder, only in `t3000-mpi-basic.t`:
```
./t/t3000-mpi-basic.log
  2021-12-13T17:28:40.977607Z broker.err[0]: rc1.0: flux-module: /usr/src/src/modules/job-manager/plugins/.libs/perilog.so: undefined symbol: flux_jobtap_epilog_start
  2021-12-13T17:28:40.977801Z broker.err[0]: rc1.0: flux-module: /usr/src/src/modules/job-manager/plugins/.libs/alloc-bypass.so: undefined symbol: flux_jobtap_job_set_flag
```

Which concerns me that we're executing the wrong version of Flux with `flux-module`? But flux shouldn't be installed in the image so I'm a bit confused. Couldn't reproduce this locally either, so maybe I'll just try restarting the build.

(suspiciously, centos7 seems to have timed out, but got a similar error from the t3000-mpi-basic.t test)

grondo:
Ok, rerunning the CI fixed the errors noted above, still a little strange, but I'll set MWP now.

Issue #4005, opened 2021-12-15
attach to jobs with `-o pty` hangs in some circumstances


grondo:
Recently noticed a couple cases where `flux job attach` hangs on jobs with `-o pty` in a state uninterruptible by Ctrl-C

 1. When an exec exception occurs, e.g. `flux mini run -o pty badcommand`
 2. With `flux mini alloc` when the instance terminates with an error. Reproducer:
   ```
   $ flux mini alloc -o exit-timeout=1 -N2 -n2 
   $ flux overlay disconnect 1
   flux-overlay: asking rank 0 to disconnect child rank 1
   $ 6.343s: flux-shell[0]: stderr: flux-shell: FATAL: doom: rank 1 exited and exit-timeout=1s has expired
   6.343s: flux-shell[0]: FATAL: doom: rank 1 exited and exit-timeout=1s has expired
   6.344s: job.exception type=exec severity=0 rank 1 exited and exit-timeout=1s has expired
   flux-job: task(s) exited with exit code 1
   flux-job: client.c:370: data_write_cb: data_write: No service matching 6885-shell-4933826904064.terminus.0 is regi
   flux-job: client.c:370: data_write_cb: data_write: No service matching 6885-shell-4933826904064.terminus.0 is regi]
 ```

PR #4010, opened 2021-12-17
fix uninterruptible hang when attached to terminated jobs with -o pty

grondo:
This PR fixes the issues described in #4005 with two updates to the pty client code:

 1. Terminate and detach the pty client on error response from a pty write message. It is unlikely things will get any better, and since the forced detach character `^\` requires a response from the server, trying to stay "connected" just causes the client to get stuck with no way out.
 2. Add periodic zero-length writes from the client to act as keepalives. This prevents a `flux job attach` to a job with a pty from hanging forever when there is no client activity and the remote is killed (e.g. after a timeout or other error)

Now this construct doesn't hang:
```console
$ flux mini alloc -N2 -n2 -o exit-timeout=1 sh -c 'flux overlay disconnect 1 && sleep inf'
flux-overlay: asking rank 0 to disconnect child rank 1
4.063s: flux-shell[0]: stderr: flux-shell: FATAL: doom: rank 1 exited and exit-timeout=1s has expired
                     4.063s: flux-shell[0]: FATAL: doom: rank 1 exited and exit-timeout=1s has expired
                      4.065s: job.exception type=exec severity=0 rank 1 exited and exit-timeout=1s has expired
[detached: remote pty disappeared]
$
```

### DIFF
--- a/src/common/libterminus/client.c
+++ b/src/common/libterminus/client.c
@@ -296,12 +296,19 @@ static void pty_client_resize (struct flux_pty_client *c)
         llog_error (c, "flux_future_then: %s", flux_strerror (errno));
 }
 
-static void pty_die (struct flux_pty_client *c, const char *message)
+static void pty_die (struct flux_pty_client *c, int code, const char *message)
 {
+    flux_pty_client_stop (c);
     if (c->attached && (c->flags & FLUX_PTY_CLIENT_NOTIFY_ON_DETACH)) {
         printf ("\033[999H[detached: %s]\033[K\n\r", message);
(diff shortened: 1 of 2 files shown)


grondo:
Hm, looks like I broke the basic `flux terminus` tests. I'll look into that.

grondo:
Ok, fixed the issue causing failures in the basic `flux terminus` tests.

Coverage is low because a larger than average amount of error handling code was touched.

garlick (approved):
LGTM!

grondo:
Thanks!

PR #4011, opened 2021-12-17
flux-top: allow top to recursively call itself

grondo:
This started as an experiment but turned out to be useful, so posting it as a PR to get comments.

Now that instance URIs are available in job listing, add the `annotations` attribute to the `flux top` `job-list` query and have jobs that are instances colored blue (or made bold if there's no color support).

Introduce the idea of a current or "selected" job. The selection is made by traversing the `flux top` job listing with the `j` and `k` or arrow keys. The currently selected job is highlighted. On the "enter" key, `flux top` recursively calls itself on the currently selected job, if that job is an instance (nothing happens for a normal job).

Since the new function `top_run()` is called recursively, hitting the `q` key in the nested `top` call pops you back up to the previous job, while `Ctrl-c` still quits immediately.

This turns `flux top` into a very limited instance explorer:

It might be a little hard to follow, but here's a screencast of an example:

test

BTW, in order to assist in testing this there's a couple underlying enhancements

 * A new `FLUX_URI_RESOLVE_LOCAL` environment variable causes `uri_resolve()` and `flux uri` to always return the `local://` equivalent of URIs.
 * The `runpty.py` script was given a `--input` option which reads input in asciicast v2 format and feeds it the child pty. The cast above in fact was created from the following input:
   ```json
   { "version": 2 }
   [2.00, "i", "j"]
   [2.50, "i", "j"]
   [2.80, "i", "j"]
   [3.50, "i", "\n"]
   [3.55, "i", "\\x0c"]
   [3.80, "i", "j"]
   [5.00, "i", "\n"]
   [5.05, "i", "\\x0c"]
   [7.00, "i", "q"]
   [7.05, "i", "\\x0c"]
   [9.00, "i", "q"]
   [9.05, "i", "\\x0c"]
   [9.55, "i", "k"]
   [10.0, "i", "\n"]
   [10.05, "i", "\\x0c"]
   [12.50, "i", "q"]
   [12.55, "i", "\\x0c"]
   [14.0, "i", "q"]
   ```

### DIFF
--- a/src/cmd/top/joblist_pane.c
+++ b/src/cmd/top/joblist_pane.c
@@ -104,7 +104,7 @@ void joblist_pane_draw (struct joblist_pane *joblist)
                          "annotations",
                            "user",
                              "uri", &uri) < 0)
-            fatal (errno, "error decoding a job record from job-list RPC");
+            fatal (0, "error decoding a job record from job-list RPC");
         if (flux_job_id_encode (id, "f58", idstr, sizeof (idstr)) < 0)
             fatal (errno, "error encoding jobid as F58");
         if (fsd_format_duration_ex (run, sizeof (run), now - t_run, 2) < 0)
--- a/src/cmd/top/joblist_pane.c
+++ b/src/cmd/top/joblist_pane.c
@@ -20,7 +20,7 @@
 
 #include "top.h"
 
-static const struct dimension win_dim = { 0, 5, 80, 60 };
+static const struct dimension win_dim = { 0, 6, 80, 60 };
 
 struct joblist_pane {
     struct top *top;
--- a/src/cmd/top/joblist_pane.c
+++ b/src/cmd/top/joblist_pane.c
@@ -142,8 +142,12 @@ static void joblist_continuation (flux_future_t *f, void *arg)
     struct joblist_pane *joblist = arg;
     json_t *jobs;
 
-    if (flux_rpc_get_unpack (f, "{s:o}", "jobs", &jobs) < 0)
-        fatal (errno, "error decoding job-list.list RPC response");
+    if (flux_rpc_get_unpack (f, "{s:o}", "jobs", &jobs) < 0) {
+        if (errno != ENOSYS)
+            fatal (errno, "error decoding job-list.list RPC response");
+        flux_future_destroy (f);
+        return;
+    }
     json_decref (joblist->jobs);
     joblist->jobs = json_incref (jobs);
     joblist_pane_draw (joblist);
--- a/src/cmd/top/joblist_pane.c
+++ b/src/cmd/top/joblist_pane.c
@@ -27,7 +27,39 @@ struct joblist_pane {
     WINDOW *win;
     json_t *jobs;
     struct ucache *ucache;
+
+    /*  Currently selected jobid. Ironically FLUX_JOBID_ANY means
+     *   no current selection.
+     */
+    flux_jobid_t current;
 };
+
+static int lookup_jobid_index (json_t *jobs, flux_jobid_t id)
+{
+    int result = -1;
+    size_t index;
+    json_t *job;
+
+    if (jobs) {
+        json_array_foreach (jobs, index, job) {
+            flux_jobid_t jobid;
+
+            if (json_unpack (job, "{s:I}", "id", &jobid) < 0)
+                continue;
+            if (jobid == id)
+                return (int) index;
+        }
+    }
+    return result;
+}
+
+static json_t *get_current_job (struct joblist_pane *joblist)
+{
+    int index = lookup_jobid_index (joblist->jobs,
+                                    joblist->current);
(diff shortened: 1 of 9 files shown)


garlick:
Woo, neat!  I wonder if we ought to change the "title" (jobid) at the top to a slash-separated path?

grondo:
> I wonder if we ought to change the "title" (jobid) at the top to a slash-separated path?

Good idea. That might be a good visual aid for the user to see how deep they've gone before they go too far...

grondo:
How's this?

test2


I notice on my system and in the asciinema player, many of the superscripts don't render properly in `flux top` :slightly_frowning_face: 

I'm not sure how to reliably detect that superscript is supported by the current font, so perhaps we should use a normal number in some way (maybe we can also provide instance `size` which could be interesting as well, though perhaps confusing for normal users)

garlick:
That looks great!!!

Sure feel free to replace the ƒ-superscript with something that can be more reliably rendered.   The size might actually be pretty useful - I was thinking it would be the same as the node count most of the time, but not so if nodes are excluded by the config, probably the norm for a system instance.

src/cmd/top/joblist_pane.c
```diff
@@ -104,6 +150,29 @@ static void joblist_continuation (flux_future_t *f, void *arg)
...
+    if (json_unpack (job,
+                     "{s:{s:{s:s}}}",
+                     "annotations",
+                       "user",
+                         "uri", &uri) < 0)
+        return;
+    if (uri == NULL)
+        return;
+    if (!(top = top_create (uri)))
+        fatal (errno, "failed to create sub-top for job");
```
garlick, line 181: Does this mean browsing the system instance and selecting another guest's batch job is fatal to top?
It might be nice if the interface didn't allow those jobs to be selected, or if the error was non fatal.  Not sure what that should like like in the UI...

grondo: Yeah, that's a good point. For now it behaves like you ran `flux top JOBID`. 

I can try to pop up a dialog box, I'm sure curses must have support for that. Or for now we could have "nothing silently happen" like it does for non-instance jobs :laughing: (I thought maybe in the future hitting enter on a non-instance job could attach to it in a new window or something, and there could be other commands, like get more detailed information, jobspec browser, cancel, etc.)

garlick:
"Nothing happens" might be the best bang for the buck at this stage :-)

grondo:
Ok, pushed a fixup here that simply returns if `top_create()` or `top_run()` fail when called recursively, rather than aborting. I also added a commit to ignore `ENOSYS` errors in RPC responses from `job-list` and resource queries, since this can happen when the instance is running, but not yet fully initialized.

I also _experimentally_ added an info line to the summary panel that shows the instance `size`, current `depth` and the value of the `version` attribute (since there was room). It doesn't look great IMO -- I tried a couple other layouts though and this was the best I came up with. Too bad superscripts don't display because that was an elegant way to hint about the depth!

I'd actually be fine dropping that commit for now, since it doesn't have much to do with the original intent of this PR. It might be good to revisit this separately later when we have more time, but I threw this up in case it was acceptable we could just use it for now.

example:

test2

garlick:
I think the new depth, size, and version display are great and we should include them.  It's not like we're locked in, if we manage to come up with a better representation or other use for the screen space later.

src/cmd/top/joblist_pane.c
```diff
@@ -60,21 +93,32 @@ void joblist_pane_draw (struct joblist_pane *joblist)
...
                          "state", &state,
                          "name", &name,
                          "nnodes", &nnodes,
                          "ntasks", &ntasks,
-                         "t_run", &t_run) < 0)
+                         "t_run", &t_run,
+                         "annotations",
+                           "user",
+                             "uri", &uri) < 0)
             fatal (errno, "error decoding a job record from job-list RPC");
```
garlick, line 107: `json_unpack()` won't reliably set errno on failure

garlick (approved):
I gave this is a spin, ran under valgrind, and walked through the code.  LGTM!  Nice improvement.

I know you've still got this marked WIP so sorry if this feedback was premature.

src/cmd/top/joblist_pane.c
grondo: I don't think that line was part of this PR, but I'll add a commit to fix it.

grondo:
> I know you've still got this marked WIP so sorry if this feedback was premature.

No feedback is fine. It was only marked WIP to give others time to comment on the added feature.

I've rebased and removed the WIP label.

garlick:
AFAIC you could set MWP on this.

grondo:
Thanks, setting MWP.

grondo:
Looks like maybe dockerhub is down. Will retry later

chu11:
re-kicked tests, dockerhub back up (or atleast it was for my PR :-)

grondo:
Thanks!

Issue #3993, opened 2021-12-06
"flux jobs -A -f pending,running" lists jobs whose states are neither pending nor running


ofaaland:
[root@elmerfudd1:~]# which flux
/bin/flux

[root@elmerfudd1:~]# rpm -qf /bin/flux
flux-core-0.30.0-1.t4.x86_64

[root@elmerfudd1:~]# flux jobs -A -f pending,running | head
       JOBID USER     NAME       ST NTASKS NNODES  RUNTIME NODELIST
 ?A7nDuT6HFm faaland1 flux        R      2      1    1.37m elmerfudd15
 ?A7k3EEmWPZ faaland1 flux       CD      2      1   4.642m elmerfudd15
 ?A78U4ZGT4s faaland1 flux       CD      1      1   0.129s elmerfudd15
 ?A78AwqfMWo faaland1 flux       CD      1      1   33.29s elmerfudd15
 ?A786dALnq5 faaland1 flux       CD      1      1   0.152s elmerfudd15
 ?A77dkxPdrP faaland1 flux       CD      1      1   56.23s elmerfudd15
 ?A73df7Yz9u faaland1 flux       CD      1      1   8.715m elmerfudd15
 ?6FZfjVWbYj faaland1 flux       CD      1      1   14.01m elmerfudd15
 ?6FZj7RDgbq faaland1 flux       CD      1      1   0.052s elmerfudd15

ofaaland:
It may be that the man page is just out of date.  It says:
```
       -f, --filter=STATE|RESULT
              List jobs with specific job state or result. Multiple states or  results  can  be
              listed  separated  by  comma.  See  JOB  STATUS below for additional information.
              Defaults to pending,running.
```

which led me to believe that a comma-delimited list of states was valid input.  However, it doesn't actually say that anywhere, nor does the output of ```flux jobs -h```.

ofaaland:
Actually, it's not just the comma-delimited list of states that is at issue.  state "running" also doesn't seem to work.
```
[root@elmerfudd1:~]# flux jobs -A -f running | head
       JOBID USER     NAME       ST NTASKS NNODES  RUNTIME NODELIST
 ?A7nDuT6HFm faaland1 flux        R      2      1    12.7m elmerfudd15
 ?A7k3EEmWPZ faaland1 flux       CD      2      1   4.642m elmerfudd15
 ?A78U4ZGT4s faaland1 flux       CD      1      1   0.129s elmerfudd15
[...]

PR #4012, opened 2021-12-17
cmd/flux-jobs: re-work -A option and clarify -a option

chu11:
Per #3993 

- make `-A` option be a convenience for `--user=all` and nothing else.  This allows `-A` to be used alongside `--filter`.
- re-word documentation that `-a` lists jobs in all states.  The documentation that it lists all jobs for the current user is redundant b/c the current user is always the default.  It is basically now a shortcut for `--filter=pending,running,inactive`
- as a result, `flux jobs -a -A` may be commonly used as a shortcut for listing all jobs, in all states, for all users

### DIFF
--- a/doc/man1/flux-jobs.rst
+++ b/doc/man1/flux-jobs.rst
@@ -24,12 +24,11 @@ OPTIONS
 =======
 
 **-a**
-   List all jobs of the current user, including inactive jobs.
-   Equivalent to specifying *--filter=pending,running,inactive*.
+   List jobs in all states, including inactive jobs.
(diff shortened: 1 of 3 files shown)


garlick (approved):
This sounds like a good change! 

Might want to get an ack from @ryanday36 and @grondo too.

ryanday36:
This seems like a good change to me.

grondo:
This is an excellent change, thanks!

chu11:
thanks, setting MWP

chu11:
oh whoops, i guess i have a python formatting issue.  i'm MWP after I fix that tomorrow.

PR #4013, opened 2021-12-17
connector cleanup

garlick:
Some connector cleanup that was part of the withdrawn #3998.

### DIFF
--- a/src/connectors/local/local.c
+++ b/src/connectors/local/local.c
@@ -11,19 +11,9 @@
 #if HAVE_CONFIG_H
 #include "config.h"
 #endif
-#include <assert.h>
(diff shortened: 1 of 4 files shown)


chu11 (approved):
LGTM.

Question, should we remove `ctx->magic == MAGIC` checks more globally?  Could create an issue for it (it might be a nice easy issue for the holidays :-)

garlick:
> Question, should we remove ctx->magic == MAGIC checks more globally? Could create an issue for it (it might be a nice easy issue for the holidays :-)

It seems like those assertions have minimal value these days, so I'm for getting rid of the ones in native flux code.

PR #4014, opened 2021-12-17
libflux: check reactor flags

garlick:
I happened to notice that `flux proxy` is calling `flux_reactor_create (SIGCHLD)`  but the flag should be `FLUX_REACTOR_SIGCHLD`.

Fix that, and add flag validation to `flux_reactor_create()` and `flux_reactor_run()`.


chu11 (approved):
LGTM!

PR #4015, opened 2021-12-19
fix segfault in uri_resolve() when argument has "/" prefix

grondo:
This PR fixes an embarrassing error in the `uri_resolve()` implementation which causes `flux proxy /JOBID` to segfault. :facepalm:

For a minute, I thought it would be cool if `/` resolved to the depth=0 instance (root) in the current hierarchy, and if we also supported `..` as "look in parent", but I'm pretty sure we don't want to go there, so `/JOBID` is treated the same as `JOBID`.


garlick (approved):
Good catch!

grondo:
Thanks!

PR #4016, opened 2021-12-21
libflux: add flux_reconnect(), revamp flux fatal error callback

garlick:
This is take 2 on `flux_t` handle reconnect machinery.  In this attempt, the flux fatal error callback (registered with `flux_fatal_set()` is reworked so that it is only called when there is a connector send/receive error, and it can return success or failure, with success meaning that the error was corrected and the send/receive should be retried.

Then a new `flux_reconnect()` function is added that calls a connector `reconnect()` operation, if implemented.  It is implemented in the local connector.  Pending RPCs are tracked, and are sent ECONNRESET error responses upon a reconnect.

Finally, as an example implementation, `flux-proxy` gets a new `--reconnect` option.  Its handle error callback attempts to reconnect forever (retrying every 2s), meanwhile blocking the reactor.  A sharness test validates that this works across a broker restart.  `flux-proxy`'s callback looks like this
```c
static int fatal_error (flux_t *h, void *arg)
{
    struct proxy_command *ctx = arg;

    if (optparse_hasopt (ctx->opts, "reconnect")) {
        log_msg ("broker: %s", strerror (errno));
        log_msg ("reconnecting");
        while (flux_reconnect (h) < 0) {
            if (errno == ENOSYS)
                log_msg_exit ("reconnect not implemented by connector");
            sleep (2);
        }
        log_msg ("reconnected");
        return 0;
    }
    log_msg_exit ("broker: %s", strerror (errno));
}
```

Loose ends
- RPCs are tracked even when there is no reconnect operation implemented in the connector
- no mechanism for re-establishing pub/sub event subscriptions
- no mechanism for re-establishing service registration
- no way to tune the number of connect retries or the interval between them

Probably at least the first issue should be addressed before this could be considered for merging, but wanted to get this out there for any feedback, especially on co-opting the "fatal error" mechanism, and leaving the "reconnect loop" up to the user.


chu11:
high level comment, should the "fatal_error()" function be renamed to something like "handle_disconnect_error()" or something?  "fatal_error" sounds .... well, fatal :-)

grondo:
This seems like a good approach to me!

Just so I'm sure I understand, what we'd need to do in the shell to add the reconnect ability would be to add a `fatal_error()` handler similar to what you've done for `flux-proxy`?

At first glance I was a little worried that blocking the reactor might cause problems for the job shell, but actually I think this might be the right approach. It will greatly simplify implementation of shell plugins if they don't have worry that callbacks might occur while offline. In fact, even if job tasks exit, the shell won't be able to do anything about it until a pending reconnect has succeeded.

> and leaving the "reconnect loop" up to the user.

￼It does seems like users of `flux_reconnect ()` will always want to do so in a loop. That is, I can't imagine a case where the caller will try to reconnect once and then give up. So on the one hand it might be useful to move the retry-until logic into the function. However, leaving it up to the caller does give a bit more flexibility, e.g. a caller could use a reactor timer to do the retries if desired so that the reactor is _not_ blocked.

You could imagine a retry loop for `flux_reconnect()` on a local connector that uses `inotify` to detect when a Unix domain socket has reappeared. I'm not sure if that kind of retry would be possible from the caller, since it is connector-dependent. That argues slightly for putting the retry into the `flux_reconnect()` function, but I'm not sure I'd strongly argue that approach presently. (Not even sure if it would work for our main use case, since if the broker dies unexpectedly it may leave the local connector socket lying around)

Sorry, just thinking aloud for a moment there.

garlick:
@chu11 - good point.   Maybe "connector error"?  e.g. `flux_connector_error_set()`, `flux_connector_error_f` etc?

@grondo - exactly.

My thought was at least this way there's an opportunity to get work done during the delay between connect attempts, as long a nobody attempts to send messages.  If we want to push the retries into the local connector, the reconnect operation uses the same parameters for retry as was used in the initial connect.  This can be overridden with the FLUX_LOCAL_CONNECTOR_RETRY_COUNT environment variable.  We could define a value that means "infinite" (like -1).  We could also add a "setopt" key for the number of retries to make the interface nicer.  Not sure if that's terribly useful though.

garlick:
Just pushed some changes
- adding an RPC to wait for broker to enter RUN state
- preventing the  error callback from being re-entered if it makes failing RPCs
- fix a bad bug (!) where the tracker would generate error responses for RPCs made within the error callback
- change flux-proxy to wait for RUN state

grondo:
FYI - I used this patch for the shell in the demo of single broker restart, and it worked great!

```diff
diff --git a/src/shell/shell.c b/src/shell/shell.c
index 95a266886..f246836c5 100644
--- a/src/shell/shell.c
+++ b/src/shell/shell.c
@@ -58,6 +58,8 @@ static struct optparse_option shell_opts[] =  {
     OPTPARSE_TABLE_END
 };
 
+static void shell_events_subscribe (flux_shell_t *shell);
+
 /* Parse optarg as a jobid rank and assign to 'jobid'.
  * Return 0 on success or -1 on failure (log error).
  */
@@ -220,6 +222,29 @@ static void shell_parse_cmdline (flux_shell_t *shell, int argc, char *argv[])
     shell->p = p;
 }
 
+static int reconnect (flux_t *h, void *arg)
+{
+    flux_future_t *f;
+    flux_shell_t *shell = arg;
+    while (flux_reconnect (h) < 0) {
+        if (errno == ENOSYS)
+            shell_die (1, "reconnect not implemented by connector");
+        sleep (2);
+    }
+    if (!(f = flux_rpc (h, "state-machine.wait", NULL, FLUX_NODEID_ANY, 0)))
+        shell_die (1, "could not send state-machine.wait RPC");
+    if (flux_rpc_get (f, NULL) < 0)
+        shell_die (1, "state-machine.wait: %s", future_strerror (f, errno));
+    flux_future_destroy (f);
+    shell_events_subscribe (shell);
+    if (!(f = flux_service_register (h, shell_svc_name(shell->svc))))
+        shell_die (1, "could not re-register shell service name");
+    if (flux_rpc_get (f, NULL) < 0)
+        shell_die (1, "flux_service_register: %s", future_strerror (f, errno));
+    shell_log ("reconnected");
+    return 0;
+}
+
 static void shell_connect_flux (flux_shell_t *shell)
 {
     if (shell->standalone)
@@ -229,6 +254,8 @@ static void shell_connect_flux (flux_shell_t *shell)
     if (!shell->h)
         shell_die_errno (1, "flux_open");
 
+    flux_fatal_set (shell->h, reconnect, shell);
+
     /*  Set reactor for flux handle to our custom created reactor.
      */
     flux_set_reactor (shell->h, shell->r);
```

garlick:
Just pushed another checkpoint of this with the following changes
* renamed "fatal error" to "comms error" per @chu11's note
* now require FLUX_O_RPCTRACK to enable RPC tracking, so we don't have that on when it's not needed
* minor commit cleanup and squashing
* rebase on current master

grondo:
Great, I will attempt to update the proof-of-concept patch I had against the shell to use this code. I'll paste the result here in case you want to include it in the PR so there is at least one user of the interface.

garlick:
> Great, I will attempt to update the proof-of-concept patch I had against the shell to use this code. I'll paste the result here in case you want to include it in the PR so there is at least one user of the interface.

This PR adds a `--reconnect` option to `flux-proxy` so there is technical a user, although it's somewhat contrived I guess.

Release notes wise, maybe the shell reconnect should stand alone?  (Your thoughts?)

I'll see if I can improve the testing here.

grondo:
> Release notes wise, maybe the shell reconnect should stand alone? (Your thoughts?)

Yeah, that sounds right. Sorry, I looked right at the `flux-proxy --reconnect` and still said the dumb thing!

garlick:
Added some docs and some validation of flags arguments in handle.c (with tests)..  I think this is getting closer so I dropped the WIP.

src/cmd/builtin/proxy.c
```diff
@@ -289,6 +289,56 @@ static void proxy_command_destroy_usock_and_router (struct proxy_command *ctx)
...
+        return -1;
+    }
+    if (!(f = flux_rpc (h, "state-machine.wait", NULL, FLUX_NODEID_ANY, 0))) {
+        log_err ("could not send state-machine.wait RPC");
+        goto done;
+    }
+    if (flux_rpc_get (f, NULL) < 0) {
+        log_msg ("state-machine.wait: %s", future_strerror (f, errno));
+        goto done;
+    }
```
grondo, line 319: I'm confused by the comment above:

> ` state-machine.wait RPC fails if broker is past the RUN state`

and the code here which skips `router_renew()` when `state-machine.wait` fails.

If a broker restarts and `flux-proxy` attempts a reconnect, but by the broker is already in RUN state by the time the `state-machine.wait` RPC is sent, shouldn' t the `router_renew()` call still be made?

I'm probably just missing something subtle.

Also, is `router_renew()` only useful for `flux-proxy`? It would be nice if the shell didn't have to re-establish its own subscriptions and service registrations after a reconnect, but it isn't a router so cannot use `router_renew()` AFAICS. Probably it is actually impossible to handle this case for clients connected using the local connector, but I thought I'd ask for clarity's sake.

grondo (approved):
Besides the one clarification already noted, this LGTM and I've separately tested using a proof-of-concept reconnect handler in the shell (to be a future PR).

The only other suggestion is perhaps to mention that subscriptions, service registrations, etc. have to be re-established after `flux_reconnect(3)`. Also optionally mention that the reconnect may have to wait for the broker to enter RUN state? (i.e. enough info for a user to implement the reconnect correctly without needing to look at example code)

Edit: And sorry if that documentation was already there and I just missed it. Just realized I didn't look too carefully at the man page yet.

src/cmd/builtin/proxy.c
garlick: I'll add some more inline commentary as this is not very clear, but the idea was that the `state-machine.wait` RPC would fail if the broker is _past_ the RUN state, e.g. it is shutting down.   The reconnect logic then does its timed retry, giving the broker a chance to be restarted by systemd.

grondo: Ugh, sorry. My fault. I had read that as "RUN state or beyond".

garlick: Yeah, `router_renew()` is only for librouter users like `flux-proxy`.

It might be possible to add a shell plugin that acts as a router for processes spawned by the shell, if that's useful.  This would not take much code if we wanted to try it.

I hadn't considered whether we could somehow use librouter from the shell thread itself.   librouter is meant to be used to build a component that is inserted in the messsage path between clients and the broker (or another router).   Might be worth thinking about that.

grondo: This is a very intriguing idea!

We already have some messages that shell internal only. For example, the rank 0 output plugin sends messages to itself. It would be neat if that could be handled internal to the shell somehow.

src/common/libflux/connector.h
```diff
@@ -30,6 +30,7 @@ struct flux_handle_ops {
     int         (*pollevents)(void *impl);
     int         (*send)(void *impl, const flux_msg_t *msg, int flags);
     flux_msg_t* (*recv)(void *impl, int flags);
+    int         (*reconnect)(void *impl);
```
chu11, line 33: should commit message say "comms error" callback now?

garlick:
Re-pushed with some improvements to testing, and documentation (man page and inline) suggested by @grondo's comments.  

I'll go ahead and set MWP.  Thanks for the review!

src/cmd/builtin/proxy.c
```diff
@@ -322,6 +375,11 @@ static int cmd_proxy (optparse_t *p, int ac, char *av[])
     if (flux_set_reactor (ctx.h, r) < 0)
         log_err_exit ("flux_set_reactor");
 
+    /* Register handler for loss of broker connection if --reconnect
+     */
+    if (optparse_hasopt (p, "reconnect"))
```
chu11, line 382: need to add `--reconnect` to `flux-proxy` manpage?

chu11 (commented):
few minor nits

garlick:
Thanks @chu11 - forced a push with those changes

PR #4022, opened 2022-01-04
flux-jobs: support instance-specific fields in output

grondo:
This PR adds a small set of `instance.` fields to the `JobInfo` class, and therefore also to `flux-jobs` output

By default, the extra information is not fetched for `JobInfo` objects. Instead a new class method `.get_instance_info()` is added which can optionally be run to fetch Flux instance data if the job is indeed an instance, and if the job is currently running.

The instance-specific data is encapsulated in a new `InstanceInfo` class, which currently contains `JobStats` and `SchedResourceList` objects at the time the data was fetched. The class also provides some possibly useful high-level attributes including:

 * `instance.utilization`: the number of cores used divided by the total number of cores in the instance
 * `instance.gpu_utilization`: same as above but for GPU resources, if any
 * `instance.progress`: number of inactive jobs divided by the total number of jobs
 * `instance.stats`: returns a string job stats summary of the form `PD:N R:N CD:N F:N`, where `PD` is pending job count, `R` is running job count, `CD` is number of successfully completed jobs and `F` is number of failed jobs.

To assist in output of percentage based fields like `instance.utilization` and `instance.progress`, a new
conversion character `P` is added that converts a floating-point number to a percentage. (See below for examples)

In `flux-jobs`, the extra instance info is only fetched if any requested field starts with `instance.`

e.g.:

```console
$ flux jobs -Rf running -o '{id.f58:>12} {instance.stats:^26} {instance.progress!P:>4h} {instance.resources.all.ncores:>5} {instance.utilization!P:>4}'
       JOBID           STATS            PROG CORES CPU%
    ƒ3tuqfCB     PD:2 R:1 CD:0 F:0        0%     1 100%

ƒ3tuqfCB:
     ƒTXt9Ww     PD:0 R:1 CD:0 F:0        0%     1 100%

ƒ3tuqfCB/ƒTXt9Ww:
     ƒT5CNSj   PD:678 R:1 CD:321 F:0     32%     1 100%

ƒ3tuqfCB/ƒTXt9Ww/ƒT5CNSj:
    ƒ2dnH5wj                               -           
```

While working on this PR, I noticed while testing in a multi-user instance that `flux jobs -R` is slow when there are multi-user jobs on the system and `-A` is used. This is because the command will try to recurse all instance jobs, not just those for the current user, and it does so serially, which when is slow when `flux_open(3)` fails for other users' jobs.

That is fixed here by:

 * Do not recurse into other users' jobs  by default. This can be disabled by using a new `--recurse-all` option.
 * Update flux-jobs to use a ThreadPoolExecutor when querying instance info or recursing. This does indeed speed up the fetch of sub-job listing or instance info since timeouts or other errors can occur in parallel. A new `--threads=N` option is provided in case the default ThreadPool max workers needs to be adjusted.

I could move these latter fixes to a separate PR if requested.

### DIFF
--- a/src/bindings/python/flux/job/info.py
+++ b/src/bindings/python/flux/job/info.py
@@ -419,6 +419,19 @@ class JobInfoFormat(flux.util.OutputFormat):
                         value = ""
                     else:
                         raise
+            elif conv == "P":
+                #  Convert a floating point to percentage
+                try:
+                    value = value * 100
+                    if value < 100:
+                        value = f"{value:.2g}%"
+                    else:
+                        value = f"{value:3.0f}%"
+                except (TypeError, ValueError):
+                    if orig_value == "":
+                        value = ""
+                    else:
+                        raise
             else:
                 value = super().convert_field(value, conv)
             return value
--- a/src/bindings/python/flux/job/info.py
+++ b/src/bindings/python/flux/job/info.py
@@ -59,6 +59,29 @@ class ExceptionInfo:
         self.note = note
 
 
+class EmptyObject:
+    """Convenience "empty" object for use with string.format
+
+    This class can be used in place of a real class but returns
+    appropriate empty or unset value for various conversions, or
+    for string.format() calls.
+    """
+
+    def __getattr__(self, attr):
+        return EmptyObject()
+
+    def __repr__(self):
+        return ""
+
+    def __str__(self):
+        return ""
+
+    def __format__(self, spec):
+        # Strip trailing specifier (e.g. d, f)
+        spec = spec.rstrip("bcdoxXeEfFgGn%")
+        return "".__format__(spec)
+
+
 # AnnotationsInfo is a wrapper for a namedtuple.  We need this
 # object so that we can we detect when an attribute is missing and
 # ultimately return an empty string (e.g. when an attribute does not
@@ -86,9 +109,10 @@ class AnnotationsInfo:
         try:
             return object.__getattribute__(self.atuple, attr)
         except AttributeError:
-            # We return an empty AnnotationsInfo so that we can recursively
+            # We return an empty object so that we can recursively
             # handle errors.  e.g. annotations.user.illegal.illegal.illegal
-            return AnnotationsInfo({})
+            return EmptyObject()
+
 
 
 class InfoList(list):
--- a/src/bindings/python/flux/job/info.py
+++ b/src/bindings/python/flux/job/info.py
@@ -191,7 +191,7 @@ class JobInfo:
             raise AttributeError
         try:
             return getattr(self, "_{0}".format(attr))
-        except KeyError:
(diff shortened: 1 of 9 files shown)


garlick:
Really nice!

Although I may be suffering from the need to have python spoon fed to me in little bites due to...personal failings, it does seem like the changes to speed up `flux jobs` recursion ought to go in another PR.

grondo:
Ok, I've chopped off the commits that modify generally how `flux jobs -R` works and I'll submit those in a follow-on PR.

src/bindings/python/flux/job/info.py
```diff
@@ -19,6 +19,8 @@
 import flux.constants
```
chu11, line 19: commit message type "speficic"

t/t2800-jobs-instance-info.t
```diff
@@ -0,0 +1,50 @@
...
+. $(dirname $0)/sharness.sh
+
+test_under_flux 4 job
+
+export FLUX_URI_RESOLVE_LOCAL=t
+waitfile="${SHARNESS_TEST_SRCDIR}/scripts/waitfile.lua"
+
+flux getattr broker.pid
+
+#  Launch some test instance jobs with varying characteristics
```
chu11, line 14: like prior PR, perhaps a comment on what you're launching, since it can take a bit to grok

t/t2800-jobs-instance-info.t
```diff
@@ -0,0 +1,50 @@
...
+	id=$(flux mini submit flux start /bin/true) &&
+	id2=$(flux mini submit -n2 -c1 flux start \
+		"flux mini run /bin/false ; \
+		 flux mini submit --cc=1-4 sleep inf && \
+		 touch ready && \
+		 flux queue idle") &&
+	flux job wait-event $id clean &&
+	$waitfile -t 60 ready
+'
+test_expect_success 'flux-jobs can get instance info' "
```
chu11, line 26: perhaps we need some `!P` percentage tests (just check atleast `%` is output?")

chu11 (commented):
nothing stands out to me, i think its a great addition.  just a few minor comments / nits.

t/t2800-jobs-instance-info.t
grondo: Good suggestion. I was struggling with how to make a good test for `!P`, but your idea is reasonable.

grondo:
I've addressed above comments and force-pushed the result. Thanks!

chu11 (approved):
LGTM

garlick (approved):
I mistakenly left some comments for commits in this PR over in my review of #4024.  But anyway, all looking good here otherwise.

grondo:
Thanks! I've addressed the comments relevant to this PR and force-pushed the result. I'll set MWP now.

PR #4023, opened 2022-01-04
testsuite: minor testsuite fixes

chu11:
hit some corner cases on my machine, minor fixes

### DIFF
--- a/t/t2801-top-cmd.t
+++ b/t/t2801-top-cmd.t
@@ -87,6 +87,7 @@ test_expect_success NO_CHAIN_LINT 'flux-top quits on q keypress' '
 	wait $pid
 '
 test_expect_success NO_CHAIN_LINT 'flux-top can call itself recursively' '
+	SHELL=/bin/sh &&
 	flux jobs &&
 	flux proxy $(cat jobid2) flux jobs -c1 -no {id} >expected.id &&
 	cat <<-EOF >recurse.in &&
--- a/t/t2802-uri-cmd.t
+++ b/t/t2802-uri-cmd.t
@@ -31,7 +31,7 @@ test_expect_success 'flux-uri passes through ssh and local URIs unchanged' '
 	test "$result" = "$local_uri" &&
 	result=$(flux uri --remote $local_uri) &&
 	test_debug "echo flux uri --remote $local_uri returns $result" &&
-	test "$result" = "ssh://$(hostname -s)/tmp/flux-xyzzy/local-0"
+	test "$result" = "ssh://$(hostname)/tmp/flux-xyzzy/local-0"
 '
 test_expect_success 'flux-uri pid resolver works' '
 	test "$(flux uri pid:$$)" =  "$FLUX_URI"


t/t2801-top-cmd.t
```diff
@@ -88,7 +88,7 @@ test_expect_success NO_CHAIN_LINT 'flux-top quits on q keypress' '
 '
 test_expect_success NO_CHAIN_LINT 'flux-top can call itself recursively' '
 	flux jobs &&
-	flux proxy $(cat jobid2) flux jobs -c1 -no {id} >expected.id &&
+	flux proxy $(cat jobid2) flux jobs -c1 -no "\{id\}" >expected.id &&
```
grondo, line 91: Just curious what about your environment caused the quoting to be necessary?

It might be nice to include that in the commit description, since it isn't clear when `{id}` would be expanded to something other than `{id}` by a POSIX shell or even bash.

t/t2802-uri-cmd.t
```diff
@@ -31,7 +31,7 @@ test_expect_success 'flux-uri passes through ssh and local URIs unchanged' '
 	test "$result" = "$local_uri" &&
 	result=$(flux uri --remote $local_uri) &&
 	test_debug "echo flux uri --remote $local_uri returns $result" &&
-	test "$result" = "ssh://$(hostname -s)/tmp/flux-xyzzy/local-0"
+	test "$result" = "ssh://$(hostname)/tmp/flux-xyzzy/local-0"
```
grondo, line 34: Nice catch! The test should probably accept either long or short hostname, but since the code that builds the remote uri is probably using the same result as `$(hostname)` I think your fix will work.

grondo (approved):
Nice finds! Sorry about the errors here.

Just one minor suggestion that the commits be filled in with `Problem:` commit body, since especially with the quoting of `{id}` it wasn't clear why that line in particular needed to be quoted (vs all the other uses of `-o {id}` in the testsuite).

t/t2801-top-cmd.t
chu11: I had just assumed we quoted the format string in most tests. Hmmm.

chu11: ohh i'm guessing its b/c we're passing the command to flux proxy, vs just running `flux jobs` itself.

chu11:
re-pushed with new commit messages per @grondo comment above

t/t2801-top-cmd.t
grondo: I can't get it to fail though.. ah, maybe the shell invoked by flux-proxy on your machine is not POSIX sh?

grondo: Maybe another good "fix" for the tests where `flux proxy` is used would be to `export SHELL=/bin/sh`?

chu11: yeah, i default to tcsh.  lets go with `/bin/sh` as the goto instead.

chu11:
re-pushed, went with a `SHELL=/bin/sh` fix instead of what I did before and updated commit message.

grondo:
LGTM! Feel free to set MWP whenever you're ready, as far as I'm concerned.

PR #4024, opened 2022-01-04
flux-jobs: improve recursive job listing

grondo:
This PR makes some improvements to recursive job listing by

 * only attempting to recurse into current user's jobs by default, unless `--recurse-all` is used.
 * connect and fetch all child job info in parallel using a ThreadPoolExecutor. A new `--threads` option is added to optionally adjust the `max_workers` value for this recursive jobs as well as instance-info gathering ThreadPools.

This is based on top of #4022

### DIFF
--- a/doc/man1/flux-jobs.rst
+++ b/doc/man1/flux-jobs.rst
@@ -91,6 +91,14 @@ OPTIONS
    equivalent to ``flux jobs`` without ``-R``, and ``--level=1`` would
    limit recursive job listing to child jobs of the current instance.
 
+**--threads**\ *=N*
+   When ``flux jobs`` recursively queries job lists (with ``--recursive``)
+   or fetches info for jobs that are also instances (see
+   ``instance.*`` fields), a pool of threads is used to parallelize
+   the required RPCs. Normally, the default number of ThreadPoolExecutor
+   threads is used, but by using the ``--threads``, a specific number
+   of threads can be chosen.
+
 
 JOB STATUS
 ==========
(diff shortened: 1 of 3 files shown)


src/bindings/python/flux/job/info.py
```diff
@@ -191,7 +191,7 @@ def __getattr__(self, attr):
             raise AttributeError
```
garlick, line 253: In commit message for c4a3555c53a1ca7ce53b57c13b589590fdf4f233, missed a "t" in "AtributeError".

doc/man1/flux-jobs.rst
```diff
@@ -78,6 +78,12 @@ OPTIONS
    option is used, then each child instance in the hierararchy is listed
    with its stats.
 
+**--recurse-all**
+   By default, jobs not owned by the user running ``flux jobs`` are
+   skipped with ``-R, --recurseive``. This option forces the command
+   to attempt to recurse into the jobs of other users.  Implies
+   ``--recursive``.
+
```
garlick, line 86: Maybe add a note that "By default, Flux instances can only permit the instance owner to connect"?

garlick (approved):
Wow, there are a lot of great improvements in here!  Approving, just a couple of minor comments.

doc/man1/flux-jobs.rst
grondo: Great idea! I added that comment, rebased on current master and pushed the result.

grondo:
Rebased after #4022 landed. Ok to set MWP?

grondo:
went ahead and set MWP.

PR #4026, opened 2022-01-05
add flux-pstree command

grondo:
This PR adds a `flux-pstree(1)` command which offers [`pstree(1)`](https://man7.org/linux/man-pages/man1/pstree.1.html)  like output for Flux job hierarchies. 

This is still a WIP because there isn't yet any documentation or tests. It seemed a good idea to get some early feedback before taking a pass at either, in case the interface needs to be changed. It is built on top of the two pending PRs #4024 and #4022.

Currently this command only targets jobs for the calling user, as it didn't seem too useful to attempt to list other user's jobs at this time. Perhaps in the future a `-A, --user=all` option could be added if necessary.

The commit that adds the command has a good overview, so I won't repeat that here. Mainly I'll just show some examples:

```console
$ flux jobs
  ƒ7vazapGzs grondo   flux        R      2      2   1.807m asp,asp
$ flux pstree --help
usage: flux-pstree [-h] [-a] [-c N] [-f STATE|RESULT] [-x] [-l] [-L N] [-p]
                   [-n] [-X] [-o FORMAT] [--parent-label FORMAT]
                   [--detail-format FORMAT] [-d NAME] [-C] [--ascii]
                   [--no-skip-root]
                   [JOBID [JOBID ...]]

positional arguments:
  JOBID                       Limit output to specific jobids

optional arguments:
  -h, --help                  show this help message and exit
  -a, --all                   Include all jobs for current user in all states
  -c, --count=N               Limit to N jobs per instance (default 1000)
  -f, --filter=STATE|RESULT   list jobs with specific job state or result
  -x, --extended              print extended details before tree output
  -l, --long                  do not truncate long lines to terminal width
  -L, --level=N               only descend N levels
  -p, --parent-ids            include jobids in parent labels
  -n, --no-header             Suppress header with -x, --extended
  -X, --no-combine            disable combination of identical children
  -o, --label=FORMAT          change label format (default='{name}')
      --parent-label=FORMAT   change label format for parent only
      --detail-format=FORMAT  specify output format for extended details
                              (implies -x, --extended)
  -d, --details=NAME          Select a named extended details format
                              (default,resources,progress,stats)
  -C, --compact               Use compact tree connectors
      --ascii                 Use ASCII character tree connectors
      --no-skip-root          Include root in output

$ flux pstree
flux
├── flux
│   └── 2*[sleep]
├── flux
│   └── sleep
└── flux
    └── 4*[sleep]
$ flux pstree --no-combine
flux
├── flux
│   ├── sleep
│   └── sleep
├── flux
│   └── sleep
└── flux
    ├── sleep
    ├── sleep
    ├── sleep
    └── sleep
$ flux pstree -x
  ƒ7vazapGzs grondo    R      2      2   2.313m flux
     ƒejEiwz grondo    R      1      1   2.275m ├── flux
    ƒ2QsbLdY grondo    R      1      1   0.790s │   ├── sleep
    ƒ2QsbLdX grondo    R      1      1   0.830s │   └── sleep
     ƒejEiwy grondo    R      1      1   2.275m ├── flux
    ƒ2UNheEj grondo    R      1      1   1.017s │   └── sleep
     ƒehkjfd grondo    R      1      1   2.275m └── flux
    ƒ2VJ5CPF grondo    R      1      1   0.454s     ├── sleep
    ƒ2VJ5CPE grondo    R      1      1   0.736s     ├── sleep
    ƒ2VJ5CPD grondo    R      1      1   0.784s     ├── sleep
    ƒ2VJ5CPC grondo    R      1      1   0.795s     └── sleep
$ flux pstree --details=progress
       JOBID NTASKS  NJOBS PROG CPU% GPU%    RUNTIME
  ƒ7vazapGzs      2      3   0%  88%         0:02:34 flux
     ƒejEiwz      1   1000  26% 100%         0:02:32 ├── flux
    ƒ2R11H3F      1                          0:00:01 │   ├── sleep
    ƒ2R11H3E      1                          0:00:01 │   └── sleep
     ƒejEiwy      1   1000  13% 100%         0:02:32 ├── flux
    ƒ2VGbD6o      1                          0:00:01 │   └── sleep
     ƒehkjfd      1   1000  53% 100%         0:02:32 └── flux
    ƒ2WuUR8w      1                          0:00:01     ├── sleep
    ƒ2WuUR8v      1                          0:00:01     ├── sleep
    ƒ2WuUR8u      1                          0:00:01     ├── sleep
    ƒ2WuUR8t      1                          0:00:01     └── sleep
$ flux pstree --details=stats
       JOBID           STATS              RUNTIME
  ƒ7vazapGzs     PD:0 R:3 CD:0 F:0        0:02:53 flux
     ƒejEiwz   PD:702 R:2 CD:296 F:0      0:02:51 ├── flux
    ƒ2S2JnK1                              0:00:01 │   ├── sleep
    ƒ2R8RDSz                              0:00:01 │   └── sleep
     ƒejEiwy   PD:850 R:1 CD:149 F:0      0:02:51 ├── flux
    ƒ2WLrgwK                              0:00:01 │   └── sleep
     ƒehkjfd   PD:397 R:4 CD:599 F:0      0:02:51 └── flux
    ƒ2YA7owm                              0:00:01     ├── sleep
    ƒ2Y8dpfT                              0:00:01     ├── sleep
    ƒ2Y8dpfS                              0:00:01     ├── sleep
    ƒ2Y8dpfR                              0:00:01     └── sleep
$ flux pstree --details=resources
       JOBID NODES  USED CORES  USED  GPUS  USED
  ƒ7vazapGzs     2     2     8     7     0     0 flux
     ƒejEiwz     1     1     2     2     0     0 ├── flux
    ƒ2S3nmbQ                                     │   ├── sleep
    ƒ2S3nmbP                                     │   └── sleep
     ƒejEiwy     1     1     1     1     0     0 ├── flux
    ƒ2XAJHxK                                     │   └── sleep
     ƒehkjfd     1     1     4     4     0     0 └── flux
    ƒ2YTufKu                                         ├── sleep
    ƒ2YPThUv                                         ├── sleep
    ƒ2YPThUu                                         ├── sleep
    ƒ2YPThUt                                         └── sleep
$ flux pstree -a
flux
├── flux
│   ├── 676*[sleep:PD]
│   ├── 2*[sleep:R]
│   └── 322*[sleep:CD]
├── flux
│   ├── 836*[sleep:PD]
│   ├── sleep:R
│   └── 163*[sleep:CD]
└── flux
    ├── 341*[sleep:PD]
    ├── 4*[sleep:R]
    └── 655*[sleep:CD]
$ flux pstree -a --no-skip-root
.
├── flux
│   ├── flux
│   │   ├── 664*[sleep:PD]
│   │   ├── 2*[sleep:R]
│   │   └── 334*[sleep:CD]
│   ├── flux
│   │   ├── 829*[sleep:PD]
│   │   ├── sleep:R
│   │   └── 170*[sleep:CD]
│   └── flux
│       ├── 314*[sleep:PD]
│       ├── 3*[sleep:R]
│       └── 683*[sleep:CD]
├── 5*[flux:CA]
├── 6*[flux:F]
└── 4*[flux:CD]
```

You can watch a complex hierarchy form using `watch -n1 flux pstree` (here I've used the `--ascii` option since Unicode box connectors don't show up with `asciicast2gif`:

output

### DIFF
--- a/doc/man1/flux-pstree.rst
+++ b/doc/man1/flux-pstree.rst
@@ -1,0 +1,233 @@
+.. flux-help-include: true
+
+==============
+flux-pstree(1)
+==============
+
+
+SYNOPSIS
+========
+
+**flux** **pstree** [*OPTIONS*] [JOBID ...]
+
+DESCRIPTION
+===========
+
+flux-pstree(1) displays a tree of running jobs by job name, similar to
+what the :linux:man1:`pstree` command does for system processes.
+
+Like pstree(1), identical leaves of the job tree are combined, which
+results in a more compact output when many jobs within a Flux instance
+share the same job name.
+
+The flux-pstree(1) command supports custom labels for jobs, including
+separately labeling parent jobs, using the same format string syntax
+supported by :man1:`flux-jobs`.
+
+The command lists actively running jobs by default, but a ``-a, --all``
+option lists all jobs in all states for the current user. In the case
+that ``-a`` is used, the job labels will automatically be amended to
+include the job status (i.e. ``{name}:{status_abbrev}``), though this
+can be overridden on the command line.
+
+The flux-pstree(1) command additionally supports listing extended
+job information before the tree display with the ``-x, --extended``,
+``-d, --details=NAME``, or ``--detail-format=FORMAT``  options, e.g.
+
+::
+
+  $ flux pstree -x
+       JOBID USER     ST NTASKS NNODES  RUNTIME
+  ƒJqUHUCzX9 user1     R      2      2   10.68s flux
+     ƒe1j54L user1     R      1      1   8.539s ├── flux
+     ƒuusNLo user1     R      1      1   5.729s │   ├── sleep
+     ƒutPP4T user1     R      1      1   5.731s │   └── sleep
+     ƒe1j54K user1     R      1      1   8.539s └── flux
+    ƒ2MYrwzf user1     R      1      1   4.736s     └── sleep
+
+Several detail formats are available via the ``-d, --details=NAME``
+option, including progree, resources, and stats. For example, the
+``progress`` display attempts to show the overall progress and
+utilization of all Flux instances in a hierarchy by displaying the
+total number of jobs in that instance (``NJOBS``), the "progress"
+(``PROG`` - inactive/finished jobs divided by total jobs), and
+core and GPU utilization (``CORE%`` and ``GPU%`` - number of used
+resources divided by total available resources):
+
+::
+
+  $ flux pstree --details=progress
+
+       JOBID NTASKS  NJOBS  PROG CORE%  GPU%    RUNTIME
+  ƒJqUHUCzX9      2      3    0% 37.5%          0:02:15 flux
+     ƒe1j54L      1   1000   23%  100%          0:02:13 ├── flux
+    ƒ2HmSnHd      1                             0:00:01 │   ├── sleep
+    ƒ2Hjxo1K      1                             0:00:01 │   └── sleep
+     ƒe1j54K      1   1000 11.5%  100%          0:02:13 └── flux
+    ƒ2b6cPMS      1                             0:00:01     └── sleep
+
+
+By default, flux-pstree(1) truncates lines that exceed the current
+value of the ``COLUMNS`` environment variable or the terminal width
+if ``COLUMNS`` is not set. To disable truncation, use the ``-l, --long``
+option.
+
+
+By default, the enclosing Flux instance, or root of the tree, is included
+in output, unless extended details are displayed as when any of the
+``-x, --extended``, ``-d, --details=NAME`` or ``--detail-format=FORMAT``
+options are used, or if one or more jobids are directly targetted with
+a ``JOBID`` argument. This behavior can be changed via the
+``--skip-root=[yes|no]`` option.
+
+
+OPTIONS
+=======
+
+**-a, --all**
+   Include jobs in all states, including inactive jobs.
+   This is shorthand for *--filter=pending,running,inactive*.
+
+**-c, --count**\ *=N*
+   Limit output to N jobs at every level (default 1000).
+
+**-f, --filter**\ *=STATE|RESULT*
+   Include jobs with specific job state or result. Multiple states or
+   results can be listed separated by comma. See the JOB STATUS section
+   of the :man1:`flux-jobs` manual for more detail.
+
+**-l, --long**
+   Do not truncate long lines at ``COLUMNS`` characters.
+
+**-p, --parent-ids**
+   Prepend jobid to parent labels.
+
+**-L, --level** *=N*
+   Only descend *N* levels of the job hierarchy.
+
+**-x, --extended**
+   Print extended details before tree output. This is the same as
+   ``--details=default``.
+
+**-d, --detail**\ *=NAME*
+   Select a named extended details format. The list of supported names
+   can be seen in ``flux pstree --help`` output.
+
+**-n, --no-header**
+   For output with extended details, do not print header row.
+
+**-X, --no-combine**
+   Typically, identical child jobs that are leaves in the tree display
+   are combined as ``n*[label]``. With this option, the combination of
+   like jobs is disabled.
+
+**-o, --label**\ *=FORMAT*
+   Specify output format for node labels using Python format strings.
+   Supports all format fields supported by :man1:`flux-jobs`.
+
+**--parent-label**\ *=FORMAT*
+   Label tree parents with a different format than child jobs.
+
+**--detail-format**\ *=FORMAT*
+   Specify an explicit details format to display before the tree part.
+   Care should be taken that each line of the format is the same width
+   to ensure that the tree display is rendered correctly (i.e. by judicious
+   use of format field widths, e.g. ``{id.f58:>12}`` instead of just
+   ``{id.f58}``.
+
+**--skip-root**\ *=yes|no*
+   Excplicitly skip (yes)  or force (no) display of the enclosing instance,
+   or root of the tree, in output.
+
+**-C, --compact**
+   Use compact tree connectors. Usefully for deep hieararchies.
+
+**--ascii**
+   Use ascii tree connectors.
+   
+
+EXAMPLES
+========
+
+The default output of flux-pstree(1) shows all running jobs for the
+current user by name, including any running sub-jobs. If there are
+currently no running jobs for the current user, only the enclosing
+instance is displayed as a ``.``, to indicate the root of the tree:
+
+::
+
+  $ flux pstree
+  .
+
+
+If there is a running job, it is displayed under the root instance,
+and includes all child jobs. Identical children are combined:
+
+::
+
+  $ flux pstree
+  .
+  └── flux
+      ├── flux
+      │   └── 2*[sleep]
+      └── flux
+          └── sleep
+  
+
+Extra information can be added to parents, which are instances of
+flux. For example, summary job stats can be easily added:
+
+::
+
+  $ flux pstree --skip-root=yes --parent-label='{name} {instance.stats}'
+  flux PD:1 R:2 CD:0 F:0
+  ├── flux PD:592 R:2 CD:406 F:0
+  │   └── 2*[sleep]
+  └── flux PD:794 R:1 CD:205 F:0
+      └── sleep
+  
+Or utilization:
+
+::
+
+  $ flux pstree --skip-root=yes \
+    --parent-label='cores={instance.resources.all.ncores} {instance.utilization!P}' \
+  cores=8 37.5%
+  ├── cores=2 100%
+  │   └── 2*[sleep]
+  └── cores=1 100%
+
+Displaying jobs in all states automatically adds the job *status* to the
+display, which offers a compact representation of the state of jobs
+throughout a hierarchy:
+
+::
+
+  $ flux pstree -a
+  .
+  ├── flux
+  │   ├── flux:PD
+  │   ├── flux
+  │   │   ├── 824*[sleep:PD]
+  │   │   ├── 2*[sleep:R]
+  │   │   └── 174*[sleep:CD]
+  │   └── flux
+  │       ├── 914*[sleep:PD]
+  │       ├── sleep:R
+  │       └── 85*[sleep:CD]
+  ├── flux:CA
+  ├── 36*[flux:CD]
+  ├── hostname:CA
+  └── hostname:CD
+  
+
+
+RESOURCES
+=========
+
+Flux: http://flux-framework.org
+
+SEE ALSO
+========
+
+:man1:`flux-jobs`
--- a/src/bindings/python/flux/util.py
+++ b/src/bindings/python/flux/util.py
@@ -19,8 +19,10 @@ import argparse
 import traceback
 import signal
 import threading
+import shutil
 from datetime import datetime, timedelta
 from string import Formatter
+from collections import namedtuple
 
 from flux.core.inner import ffi, raw
 from flux.utils.parsedatetime import Calendar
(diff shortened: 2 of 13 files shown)


src/bindings/python/flux/util.py
```diff
@@ -380,3 +382,163 @@ def format(self, obj):
...
+        if pstack is None:
+            pstack = []
+        if level and level < len(pstack):
+            return
+
+        if self.count > 1:
+            label = f"{self.count}*[{self.label}]"
+        else:
+            label = self.label
+
```
chu11, line 485: took me awhile to grok this recursion.  I'm sure it was hard to write :-)

I think the `pstack[:-1]` sort of threw me for a loop b/c we're not using the last piece of the pstack, so it sort of doesn't make sense until you walk yourself threw it a bit.  Like one level of recursion down, the append to the pstack isn't used at all.  

I'm sure some comments would help.

One random idea (possibly not good, but just brainstorming), since the last append of the `pstack` isn't used when you recurse only one level down, instead of appending to `pstack`, how about just pass a variable down.  Like hypothetically:

```
def _render(
        self,
        pstack=None,
        next_in_pstack=None
        connector="",
        style="box",
        level=None,
        truncate=None,
    ):
```

So like if `next_in_pstack != None` then you append to pstack, then set `next_in_pstack` to the next recursive step?

src/bindings/python/flux/util.py
```diff
@@ -380,3 +382,163 @@ def format(self, obj):
...
+    def __init__(self, label, prefix="", combine_children=False):
+
+        self.label = label
+        self.prefix = prefix
+        if self.prefix:
+            self.prefix += " "
+
+        self.duplicates = {}
+        self.children = []
+        self.count = 1
```
chu11, line 413: would suggest `combine_count` or `duplicate_count` or something like that for this variable, as I thought we might be counting levels or something.  Below

```
      if self.count > 1:
            label = f"{self.count}*[{self.label}]"
```

was initially confusing, I had no idea what was going on until I figured out "ohh this is the duplicate count label".

perhaps `increment` below would be renamed as well.

src/bindings/python/flux/util.py
```diff
@@ -380,3 +382,163 @@ def format(self, obj):
...
+        result = f"{self.prefix}{prefix}{connector}{label}"
+        if truncate and len(result) > truncate:
+            result = result[: truncate - 1] + "+"
+        print(result)
+
+        connectors = self.connector_styles[style]
+        for child in self.children:
+            if child == self.children[-1]:
+                pstack.append(connectors.BLANK)
+                connector = connectors.ELBOW
```
chu11, line 496: not sure if `connector` is the best variable since you have the `connectors` variable.  Although unsure of a better variable name.

src/cmd/flux-pstree.py
```diff
@@ -0,0 +1,366 @@
...
+        "--ascii",
+        action="store_true",
+        help="Use ASCII character tree connectors",
+    )
+    parser.add_argument(
+        "--no-skip-root",
+        action="store_false",
+        dest="skip_root",
+        help="Include root in output",
+    )
```
chu11, line 319: very first time I ran this I got:

```
flux
flux
```

and I thought something was broken.  But running `flux pstree --no-skip-root`, I got what I thought I would see

```
.
├── flux
└── flux
```

dunno if default should be with root?

chu11 (commented):
this looks awesome, just some initial few comments

src/bindings/python/flux/util.py
```diff
@@ -380,3 +382,163 @@ def format(self, obj):
...
+        self,
+        pstack=None,
+        connector="",
+        style="box",
+        level=None,
+        truncate=None,
+    ):
+        if pstack is None:
+            pstack = []
+        if level and level < len(pstack):
```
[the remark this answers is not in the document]
grondo: No, we want to stop recursion when the actual depth (represented by the length of `pstack`) is _greater than_ the requested `level`. Perhaps the variable names are confusing here, I'll fix that (`level` is the requested maximum depth to recurse, not the current level of the tree recursion)

src/bindings/python/flux/util.py
grondo: Does it help if you think of the stack as the stack of characters required in before the tree connector if you need to recurse another level down? Your suggestion to pass the next character required as a separate argument seems good, but I don't see much difference between the current approach and that (except adding more parameters to the function). Perhaps some comments will help.

src/bindings/python/flux/util.py
grondo: Ok, I'll try to think of something better.

src/cmd/flux-pstree.py
grondo: I struggled with this as well, and had it the way you suggest for awhile. The problem I ran into is that including root doesn't add much useful data, except when you catch a couple instances before they have run any jobs as you have here, and makes it more likely the tree will overflow the terminal.

However, maybe my contrived examples were deeper than in the common case, so possibly it won't be a problem in reality.

Another idea would be to include the "root" in the normal case, but exclude in when printing "detailed" information, in which case the output would be less likely to be confusing and we'd be more likely to exceed the current terminal width.

src/bindings/python/flux/util.py
grondo: That seems like a valid suggestion.

src/bindings/python/flux/util.py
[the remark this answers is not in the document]
chu11: ahhh, i think that was it, I was thinking current level not max level.

src/bindings/python/flux/util.py
chu11: Yeah, some comments would help.  The way I eventually thought about it is that you are passing a connector and an additional prefix into the next level of recursion.  It just so happens that the additional prefix is never used in the next recursive level, but is only used 2 recursions later.

Another brainstorming idea (again, possibly not good), you could do something like:

```
if pstack[-1] == connectors.TEE
   pstack[-1] = connectors.PIPE
pstack.append(connects.TEE)
```

kinda thing in the recursive step so that the whole `pstack` is just output everytime you recurse.  I'm not sure if it's a net win logic wise or not.  Perhaps simplifying the output code but complicating the recursive step.

src/cmd/flux-pstree.py
chu11: > Another idea would be to include the "root" in the normal case, but exclude in when printing "detailed" information, in which case the output would be less likely to be confusing and we'd be more likely to exceed the current terminal width.

I like this idea.

In time perhaps a jobid input would go along with `flux pstree`?  In that case I could see `skip-root` being normal (sort of like when you do `pstree PID`).  But for the general listing of all jobs of a user, listing root seems right?  Sort of like the default output of `pstree`?

grondo: `flux pstree` can already take a list of JOBIDs.

The real problem with "root" is that we can't easily create a `JobInfo` object for it, and for test instances or root instances run under a foreign RM there isn't even a jobid, so the comparison to `pstree` falls flat there. However, since the root is handled specially anyway and the label is always forced to `.`, maybe all that is fine.

I can also try to make a "fake" JobInfo object for the enclosing instance and see how that goes.

src/bindings/python/flux/util.py
grondo: I don't know, it feels like we're getting into a matter of opinion and style here and I'd prefer to keep the code the way I wrote it since I understand that approach a little better. Let me push my fixup commit with additional comments and see what you think.

src/bindings/python/flux/util.py
grondo: fixed in most recent push

src/bindings/python/flux/util.py
grondo: Changed `connectors` to `cstyle` in latest fixup commit.

src/bindings/python/flux/util.py
```diff
@@ -380,3 +382,206 @@ def format(self, obj):
...
+    def _render(
+        self,
+        pstack=None,
+        connector="",
+        style="box",
+        max_level=None,
+        truncate=None,
+    ):
+
+        #  `pstack` is a stack of prfix characters used during
```
chu11, line 488: typo `prfx`

src/bindings/python/flux/util.py
```diff
@@ -380,3 +382,206 @@ def format(self, obj):
...
+        #   recursive walk of the Tree. This stack maintains the
+        #   extra prefix characters required before printing the
+        #   current tree connector and node label.
+        #
+        #  Note: the connector at the top of the stack is always
+        #   the connector required for the _next_ level down in the
+        #   tree. This is because we know what the prefx connector
```
chu11, line 495: typo `prefx`

src/bindings/python/flux/util.py
chu11: acknowledged, its a style thing at this point.  The comments you added definitely help.

src/bindings/python/flux/util.py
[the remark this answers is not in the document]
grondo: I renamed `level` to `max_level` in `_render()` in the most recent fixup command (and reversed the sense of the test, you were right it was pretty confusing)

grondo:
@chu11: The most recent fixup commit here does the following

 * Replace `--no-skip-root` with `--skip-root=yes|no` option. The default is `--skip-root=no` unless `-d, --details=NAME` or `-x, --extended` are used
 * If we're not skipping the root of the tree, then the program attempts to create a sensible `JobInfo` object for the enclosing instance, either by determining the current jobid and using `job-list` in the parent, or by mocking up some information. Either way "instance info" fields will be available for the "root", so this is probably a net improvement, even though the approach is a little kludgy.

Let me know what you think.

grondo:
After testing on fluke I found that `flux pstree` was really slow when it has to connect to jobs with the `ssh` connector (instead of forcing all URIs to `local://` like I do on my home test machine)

To remedy, I added support for using a ThreadPoolExecutor to `flux-pstree` as was done for `flux jobs`. This greatly speeds up operation of the tool, since jobs are gathered in  separate threads simultaneously at each level. However, the current solution is just a stopgap for now, as it creates a ThreadPoolExecutor at each level of the recursive search, which means we've got ThreadPools being created from threads. There is probably a nice way to create a global ThreadPool for the pstree process and use that, but I don' t know it. There is a comment in the code that this should be fixed someday.

chu11:
> Let me know what you think.

It's lookin pretty good!

One high level comment.  Do you imagine the `Tree` class you added to Python util to be needed in other places?  It seems sort of specific to `pstree`, especially with all the rendering code in there.

Second high level comment (now that I read the above), `Tree` seems sort of generically named.  Potential collision in the future?  Granted this is Python and not C, but still.

grondo:
> One high level comment. Do you imagine the Tree class you added to Python util to be needed in other places? It seems sort of specific to pstree, especially with all the rendering code in there.

My first thought was to redo `flux overlay` in Python and reuse the `Tree` class there for `flux overlay status`, but it ended up being easier to modify the command in C than rewrite in Python. However, there are many things in Flux that can be displayed with a hierarchical representation besides the TBON and job hierarchies. For instance, job dependencies, compute resources, jobspec etc. It could perhaps also be useful in flux-accounting when displaying a fairshare hierarchy (actually that last one is a stretch, I don't really know if it would be useful there)

Additionally, having the class in the bindings allows easier unit testing.

> Second high level comment (now that I read the above), Tree seems sort of generically named. Potential collision in the future? Granted this is Python and not C, but still.

That's a good question. Since Python namespaces everything, the class is actually `flux.util.Tree` not just `Tree`, so I didn't feel too bad about the naming. However, I'm not really a Python expert. Honestly, I was following the naming example from the Python 
[rich](https://github.com/Textualize/rich/blob/master/rich/tree.py) project (which I wish was ubiquitous enough to use here)

grondo:
I've added some tests and a `flux-pstree(1)` manpage, and removed WIP label.

doc/man1/flux-pstree.rst
```diff
@@ -0,0 +1,231 @@
+==============
```
chu11, line 1: Is this manpage worthy of being added to flux help output?  I would think so?

doc/man1/flux-pstree.rst
```diff
@@ -0,0 +1,231 @@
...
+
+RESOURCES
+=========
+
+Flux: http://flux-framework.org
+
+SEE ALSO
+========
+
+:man1:`flux-jobs`
```
chu11, line 231: should add `pstree` to `flux jobs`?

t/t2803-flux-pstree.t
```diff
@@ -0,0 +1,285 @@
...
+	.
+	├── flux
+	│   ├── flux
+	│   │   └── sle+
+	│   └── flux
+	│       └── sle+
+	└── flux:CD
+	EOF
+	test_cmp ${name}.expected ${name}.output
+'
```
chu11, line 169: duplicated test? (or if its different and i'm missing the difference, the title of the test is the same atleast)

chu11 (approved):
overall LGTM, just a few comments below

doc/man1/flux-pstree.rst
grondo: I wasn't sure at this point if it rose to that level?

doc/man1/flux-pstree.rst
grondo: Good idea. Will do.

t/t2803-flux-pstree.t
grondo: Good catch. Accidental cut-and-paste. Will fix.

doc/man1/flux-pstree.rst
chu11: Its certainly a "user tool", so not in the same class as lets say `flux keygen`.  But it's clearly a more advanced user tool.  I dunno, I would have leaned yes, but its definitely borderline.

grondo: I went ahead and added it. Thanks for the suggestion!

grondo:
Ok, I've addressed @chu11's comments and will now set MWP.

PR #4028, opened 2022-01-05
flux-overlay: enhance tree for overlay status

grondo:
Problem: The flux-overlay status command output doesn't connect
parent ranks to all their children, making output slightly difficult
to read.

Add support for a prettier tree output using compact Unicode tree
"connectors". To properly maintain current node indent and prefix,
a "stack" of connectors is maintained throughout the traversal,
where the appropriate parent prefix is pushed on descending to a child,
and popped when recursive descent is complete.

Unfortunately, this doesn't work prefectly with the --down option
since in that case there is no way to know when the traversal has
encountered the last child of a parent, but still the output even
in this case is slightly improved.

before:
```
0 asp: full
└1 asp: full
 └3 asp: full
  └7 asp: full
   └15 asp: full
  └8 asp: full
 └4 asp: full
  └9 asp: full
  └10 asp: full
└2 asp: full
 └5 asp: full
  └11 asp: full
  └12 asp: full
 └6 asp: full
  └13 asp: full
  └14 asp: full
```
after:
```
0 asp: full
├─ 1 asp: full
│  ├─ 3 asp: full
│  │  ├─ 7 asp: full
│  │  │  └─ 15 asp: full
│  │  └─ 8 asp: full
│  └─ 4 asp: full
│     ├─ 9 asp: full
│     └─ 10 asp: full
└─ 2 asp: full
   ├─ 5 asp: full
   │  ├─ 11 asp: full
   │  └─ 12 asp: full
   └─ 6 asp: full
      ├─ 13 asp: full
      └─ 14 asp: full
```

### DIFF
--- a/src/cmd/builtin/overlay.c
+++ b/src/cmd/builtin/overlay.c
@@ -15,6 +15,7 @@
 #include <flux/core.h>
 
 #include "src/common/libccan/ccan/str/str.h"
+#include "src/common/libccan/ccan/ptrint/ptrint.h"
 #include "src/common/libutil/monotime.h"
 #include "src/common/libutil/fsd.h"
 #include "src/common/libhostlist/hostlist.h"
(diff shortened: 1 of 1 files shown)


garlick (approved):
Very nice!   I couldn't quite figure out how to do this.  Your solution is elegant.

grondo:
Thanks!

PR #4031, opened 2022-01-06
fluxorama: allow user to sudo to flux user, add common systemd environment vars to flux user's bashrc

chu11:
Add a few conveniences to the `fluxorama` image:

- allow primary user to sudo to root or the `flux` user.  Since we launch the broker under `flux`, can be useful for debugging
- setup common systemd environment vars for `flux` user in their `.bashrc`, so that they are already setup when `su`-ing or `sudo`-ing to the `flux` user.


grondo (approved):
Looks reasonable to me!

PR #4032, opened 2022-01-07
t2800-jobs-recursive.t: add missing FLUX_SECURITY prereq to tests

### LINKED ISSUE #4029: t2800-jobs-recursive.t: not ok 12 - cancel alternate user job
Problem:  I'm noticing this new failure on my desktop:
```
$ ./t2800-jobs-recursive.t  -v
[snip]
skipping test: submit fake instance job as another user 
	submit_fake_user_instance

ok 2 # skip submit fake instance job as another user (missing FLUX_SECURITY)

[snip]
expecting success: 
	flux job cancel $(cat altid)

cat: altid: No such file or directory
Usage: flux-job cancel [OPTIONS] id [message ...]
Cancel a job
not ok 12 - cancel alternate user job
#	
#		flux job cancel $(cat altid)
#	
```
Possibly `altid` is not created because the function that creates it, `submit_fake_user_instance()` is only called if the FLUX_SECURITY prereq is set, so a fix would be to add the FLUX_SECURITY prereq to the failing test.

I'm not sure why we're not hitting this in CI though.

grondo:
Problem: The FLUX_SECURITY prereq is missing on some tests that
require it, causing t2800-jobs-recursive.t to fail on flux-core
that isn't built --with-flux-security.

Add the necessary prereq to sharness tests that require flux-security.

Fixes #4029

### DIFF
--- a/t/t2800-jobs-recursive.t
+++ b/t/t2800-jobs-recursive.t
@@ -90,12 +90,13 @@ test_expect_success 'flux jobs --recursive JOBID works' '
 	test_debug "cat recursive-jobid.out" &&
 	test $(grep -c : recursive-jobid.out) -eq 3
 '
-test_expect_success 'flux jobs --recurse-all tries to recurse other user jobs' '
+test_expect_success FLUX_SECURITY \
+	'flux jobs --recurse-all tries to recurse other user jobs' '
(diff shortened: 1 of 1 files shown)


chu11 (approved):
just hit this while reviewing your pstree PR, it works!

PR #4035, opened 2022-01-07
systemd: set linger on flux user

### LINKED ISSUE #4030: fluxorama: keep flux user session lingering
In some systemd user instances, if the user logs out, systemd may kill any lingering user services.  This isn't any issue for the developer's user session (e.g. `chu11` for me) but can be an issue for the `flux` user service that runs the flux broker.

Under some testing, I found that the `flux` user service can die while doing some minor debugging like via sudo (`sudo -u flux ...`) or changing to the flux user and exiting (e.g. `sudo su flux; do something; exit`).

Temporarily, we can stop this from happening via:

`loginctl enable-linger flux`

chu11:
Problem: Systemd user services can be stopped after the last
user session for that user has exited.  This can become a problem
during development / testing / debugging if someone runs something
under the flux user (e.g. 'sudo -u flux ...`) or su's to the flux
user and then exits.

Solution: Set linger on the flux user to ensure systemd user
services will not be stopped.

Fixes #4030

### DIFF
--- a/etc/flux.service.in
+++ b/etc/flux.service.in
@@ -32,6 +32,7 @@ PermissionsStartOnly=true
 ExecStartPre=-/bin/mkdir -p @X_LOCALSTATEDIR@/lib/flux
 ExecStartPre=/bin/chown flux:flux @X_LOCALSTATEDIR@/lib/flux
 ExecStartPre=/bin/chmod 0700 @X_LOCALSTATEDIR@/lib/flux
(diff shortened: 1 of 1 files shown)


chu11:
side note: can go in after 0.33 tag if we view this as a tad "dangerous" to go in right before a release

grondo (approved):
This looks reasonable to me. I don't see a better way to do this.

Issue #4036, opened 2022-01-07
Job timeouts when submitting 10,000 jobs


Treece-Burgess:
Upon submitting 10,000 short lived jobs to Flux, some of the jobs managed by Flux fail. 
We also do see a relationship of how many jobs fail depending on the job length, with shorter jobs being more likely to fail.

For example:
For a job length of 0, we observe a median of **860** failed jobs
For a job length of 5, we observe a median of **214** failed jobs
For a job length of 10, we observe a median of **98** failed jobs

Steps to replicate:
1. Have the batch script and the reproducer within the same directory
2. Submit the batch script to the LSF scheduler to run 

The job failures will be at the top of stdout.0.log.

We observe the problem for the following configuration: 
- Flux version c0.29.0 - s0.18.0
- 32 nodes @ 40 cores on Lassen
- Topology of 1

The following zip file will contain the reproducer.py which can reproduce this issue, and its subsequent batch script: [Reproducer_and_Batch.zip](https://github.com/flux-framework/flux-sched/files/7829634/Reproducer_and_Batch.zip)

grondo:
I've transferred this issue to flux-core since I was able to reproduce with the simple scheduler.

grondo:
@Treece-Burgess, we've added a fix in #4037 which will help avoid many, but not all, of the spurious timeouts we see when running this reproducer. The fix will be included in v0.33 of flux-core.

Even with the fix, some jobs still time out if the run during the initial submission phase. I believe this is because the instance is bogged down handling the 10k job submissions and this slows handling of events and messages related to running jobs.

Until we can address this issue in some more elegant manner, it might be wise to disable the job queue while doing large bulk submissions of jobs so that no jobs run during the submission phase. The queue can then be enabled once all jobs have been submitted.

This can be done with the `flux queue stop` and `flux queue start` commands, or special RPCs to the `job-manager.alloc-admin` service. For example, here's a patch to the reproducer which disables the queue during job submission using RPCs. I don't get any job failures with this change.

```diff
--- reproducer.py.orig	2022-01-07 20:44:17.615100868 -0800
+++ reproducer.py	2022-01-07 20:52:57.862814124 -0800
@@ -25,6 +25,17 @@
 subs_in_flight = 0  # current job submissions in flight
 job_ids = deque()
 
+def queue_stop(h):
+    h.rpc(
+        "job-manager.alloc-admin",
+        {"query_only": False, "enable": False, "reason": "benchmark"}
[...]

PR #4037, opened 2022-01-08
add workaround for invalid job timeouts when system is busy

grondo:
This is a possible fix for #4036.

When running a high throughput series of jobs with very short durations, a delay between when R is created by the scheduler and the job is actually started by the job-exec module can be long enough to cause jobs to be prematurely terminated due to timeout, or for very short durations jobs can even get an exception due to starting after the expiration set in R.

This simple change adds an adjustment in the job-exec module to take into account the time difference between R starttime and the time at which the job-exec module is actually starting the job.

Long term there may be a better solution, but this should work around the specific issue in #4036.

I ran the reproducer several times with this fix and didn't hit any timeout failures.

### DIFF
--- a/src/modules/job-exec/job-exec.c
+++ b/src/modules/job-exec/job-exec.c
@@ -321,26 +321,6 @@ static void jobinfo_complete (struct jobinfo *job, const struct idset *ranks)
     }
 }
 
-void jobinfo_started (struct jobinfo *job)
-{
-    flux_t *h = job->ctx->h;
-    if (h && job->req) {
-        job->running = 1;
-        if (jobinfo_respond (h, job, "start") < 0)
-            flux_log_error (h, "jobinfo_started: flux_respond");
-    }
-}
-
-void jobinfo_reattached (struct jobinfo *job)
-{
-    flux_t *h = job->ctx->h;
-    if (h && job->req) {
-        job->running = 1;
-        if (jobinfo_respond (h, job, "reattached") < 0)
-            flux_log_error (h, "jobinfo_reattach: flux_respond");
-    }
-}
-
 static void kill_timer_cb (flux_reactor_t *r, flux_watcher_t *w,
                            int revents, void *arg)
 {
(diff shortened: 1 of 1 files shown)


grondo:
I still do get some timeouts when running the reproducer with 10000 sleep 0 jobs with a 3s duration set in jobspec. In my runs I got 9 timeouts, this histogram shows that these are outliers:

```
 4337|    o                                                                    
 4109|    o                                                                    
 3881|    o                                                                    
 3653|    o                                                                    
 3424|    o                                                                    
 3196|    o                                                                    
 2968|   oo                                                                    
 2740|   oo                                                                    
 2511|   oo                                                                    
 2283|   oo                                                                    
 2055|   oo                                                                    
 1827|   oo                                                                    
 1598|   oo                                                                    
 1370|   ooo                                                                   
 1142|   ooo                                                                   
  914|   ooo                                                                   
  685|   ooo                                                                   
  457|   ooo                                                                   
  229|  ooooo                                                                  
    1| oooooooo                                  ooo                          o
      ------------------------------------------------------------------------

------------------------------------
|             Summary              |
------------------------------------
|       observations: 10000        |
|       min value: 0.152997        |
|         mean : 0.343446          |
|       max value: 5.762063        |
------------------------------------

```

Here's the eventlog for the maximum runtime:
```
$ flux job eventlog ƒCSt57uNZ
1641600030.145318 submit userid=1000 urgency=16 flags=0
1641600030.329695 depend
1641600030.329725 priority priority=16
1641600031.667550 alloc annotations={"sched":{"resource_summary":"rank3/core0"}}
1641600034.617852 start
1641600037.429612 exception type="timeout" severity=0 userid=-1 note="resource allocation expired"
1641600037.434898 finish status=14
1641600037.927737 release ranks="all" final=true
1641600037.929883 free
1641600037.929909 clean
```
and the `exec.eventlog`:

```
ƒ(s=4,d=0,builddir) grondo@asp:~/test$ flux job eventlog -p guest.exec.eventlog ƒCSt57uNZ
1641600031.677094 init
1641600034.427007 starting
1641600037.434806 complete status=14
1641600037.434850 done
```

grondo:
I've slightly modified the fix here to move the start of expiration timer in the job-exec module to just before the `start` event is emitted. Since the expiration time is also adjusted based on the difference between the R starttime and the current time, this should further protect against start delays in the system.

However, this still does not fully fix the issue reported in #4036. I've noticed the first 8 or jobs still timeout with `sleep 0` jobs and a 3s duration. I believe this is due to general slowness since the 10K jobs are still being ingested during that time.

A 10s duration seems to be enough grace time for even the slowest of these jobs to complete in my testing.

garlick (approved):
This seems reasonable to me.   I can't think of a better way to deal with this.

grondo:
Thanks!

Issue #4038, opened 2022-01-08
flux --parent has wrong FLUX_EXEC_PATH when grandparent is system instance


grondo:
`t2802-uri-cmd.t` currently fails when run as a job in a system instance:

```
expecting success: 
	jobid2=$(flux jobs -no {user.jobid} $jobid) &&
	test_debug "echo attempting to resolve jobid:${jobid}/${jobid2}" &&
	uri=$(FLUX_SSH=$testssh flux uri --local jobid:${jobid}/${jobid2}) &&
	test_debug "echo jobid:${jobid}/${jobid2} is ${uri}" &&
	uri=$(FLUX_SSH=$testssh flux uri --local ${jobid}/${jobid2}) &&
	test_debug "echo ${jobid}/${jobid2} is ${uri}"

attempting to resolve jobid:f3vhcnvK/fP1ULdu
hostname=fluke26
cmd="LD_LIBRARY_PATH=/usr/lib64/flux /g/g0/grondo/git/flux-core.git/src/cmd/flux relay /var/tmp/grondo/flux-PWkMjy/local-0"
flux-uri: ERROR: [Errno 38] No service matching job-list.list-id is registered
flux-uri: DEBUG: Traceback (most recent call last):
  File "/g/g0/grondo/git/flux-core.git/src/bindings/python/flux/util.py", line 203, in __call__
    main_func()
  File "/g/g0/grondo/git/flux-core.git/src/cmd/flux-uri.py", line 57, in main
    uri = resolver.resolve(args.uri)
  File "/g/g0/grondo/git/flux-core.git/src/bindings/python/flux/uri/uri.py", line 192, in resolve
    result = JobURI(self.resolvers[scheme].resolve(resolver_uri))
  File "/g/g0/grondo/git/flux-core.git/src/bindings/python/flux/uri/resolvers/jobid.py", line 72, in resolve
    return self._do_resolve(uri, flux.Flux(), force_local=force_local)
  File "/g/g0/grondo/git/flux-core.git/src/bindings/python/flux/uri/resolvers/jobid.py", line 64, in _do_resolve
    resolver_uri, flux.Flux(uri), force_local=force_local
  File "/g/g0/grondo/git/flux-core.git/src/bindings/python/flux/uri/resolvers/jobid.py", line 45, in _do_resolve
    flux_handle, jobid, attrs=["state", "annotations"]
[...]

PR #4045, opened 2022-01-11
env: fix prepend of colon-separated paths in reverse order

grondo:
This fixes issue #4038, among other possible issues, in which `flux --parent` could end up incorrectly pushing the system path onto the front of `FLUX_EXEC_PATH` when running from a builddir. This occurs because when a path with multiple elements is pushed onto `FLUX_EXEC_PATH`, the elements are pushed in reverse order because they are iterated _in order_ and pushed onto the front individually.

This fix just iterates the path elements in reverse order when pushing onto the front.

For an unknown reason, this then required reversing the order of the intree PYTHONPATH elements (seemed to be a Python bug to me, but I didn't spend too much time on it since the order shouldn't matter.)


garlick (approved):
Nice job running this subtle one down.

grondo:
Thanks! Setting MWP

PR #4046, opened 2022-01-12
ci: fix use of poison-libflux.sh and add poison fllux-* commands

grondo:
This PR fixes the use of the `poison-libflux.sh` script during `docker-run-checks.sh`. It turns out the script wasn't actually being used because it needs to be run under `sudo`.

Once that was fixed, it turns out many tests were being linked with system `libflux-*.so` libraries under some distros. This required some further fixes and experimental workarounds to the build system.

Finally, some poisoned `flux-*` commands are added to a fake `FLUX_EXEC_PATH` to ensure we never call out to commands in the system or previously registered `FLUX_EXEC_PATH` during the testsuite (as was seen with #4038)

Along the way a couple other minor testsuite issues were fixed.

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -369,7 +369,8 @@ AM_CFLAGS = \
 	$(CODE_COVERAGE_CFLAGS)
 
 AM_LDFLAGS = \
-	$(CODE_COVERAGE_LIBS)
+	$(CODE_COVERAGE_LIBS) \
+	-no-install
 
 AM_CPPFLAGS = \
 	-I$(top_srcdir) \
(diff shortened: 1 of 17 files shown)


doc/Makefile.am
```diff
@@ -369,7 +369,8 @@ AM_CFLAGS = \
 	$(CODE_COVERAGE_CFLAGS)
```
garlick, line 369: commit message for c34a429065514815a0b2a2974e1535350993048f

s/system-install/system-installed/
s/wil/will/

garlick (approved):
Wow, lot of great fixes here.  Thanks for the great commit messages explaining the vudu!  Jeez, what a pain.

grondo:
Fixed the commit message typo noted by @garlick and force-pushed the result.

Will set MWP. Thanks!

PR #4051, opened 2022-01-14
broker: avoid leaking Cray environment

garlick:
Problem: if flux is launched by slurm on a cray system with
the cray_shasta plugin loaded, and then flux tries to launch
cray MPI, the job may start with incorrect job information.

Flux cannot launch cray MPI on a shasta system without its
own similar plugin.  When that is written it will likely overwrite
any existing values.  However, it may not be loaded by default,
so these changes are still useful.

Add cray shasta runtime variables to runat skiplist, which
prevents them from being set in the Flux initial program,
which is often the initial environment for work submitted to flux.

### DIFF
--- a/src/broker/runat.c
+++ b/src/broker/runat.c
@@ -83,6 +83,16 @@ static const char *env_blocklist[] = {
     "PMIX_SERVER_URI2",
     "PMIX_SERVER_URI",
 #endif
+    /* Cray shasta plugin for slurm/flux sets these variables.
+     * If set in flux's environment, don't leak through to applications
+     * launched by flux or incorrect job info could be conveyed.
+     */
+    "PALS_APID",
+    "PALS_RANKID",
+    "PALS_NODEID",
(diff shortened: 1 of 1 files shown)


grondo:
Just one concern: It seems unsustainable to have every possible influential environment variable compiled into a broker env blocklist. Perhaps environment variables sanitized in the broker could be supplemented via toml config? Then the list could be customized on a site-by-site basis, without requiring a new version of flux-core when it is discovered we missed one.

Or, if the environment variables only affect jobs launched by Flux, this seems more easily maintained in a shell plugin. A default one that sanitizes the same variables could be distributed with Flux. If it is always loaded before any other Cray MPI specific plugin, then it wouldn't remove any environment variables set by said plugin.

garlick:
Great point, and it sounds like @jameshcorbett is already doing that in the flux-shasta2 plugin for launching cray MPI jobs.  So maybe I had better withdraw this PR.

PR #4052, opened 2022-01-14
resource: drain torpid nodes

garlick:
Here is a first cut at changes to drain nodes that are "torpid" (unresponsive).

A broker group `broker.torpid` is set up to track torpid nodes based on the existing code for logging them.  Then the resource module watches that group.  Any broker ranks (er execution targets) entering the group are drained with the message `broker was unresponsive` and will remain drained until someone manually runs `flux resource undrain target`.

The parameters affecting this behavior can be set via config file or broker attribute.  The broker attribute can be set on the fly.  In case it becomes annoying during a DAT or something, it can be disabled with `flux exec flux setattr tbon.torpid_max 0`.

I just realized while writing this that I didn't update man pages so adding WIP and will take care of that later.

### DIFF
--- a/doc/man7/flux-broker-attributes.rst
+++ b/doc/man7/flux-broker-attributes.rst
@@ -76,6 +76,19 @@ tbon.prefertcp
    with PMI, tcp:// endpoints will be used instead of ipc://, even if all
    brokers are on a single node.
 
+tbon.torpid_min
+   The amount of time (in RFC 23 Flux Standard Duration format) that a broker
+   will allow the connection to its TBON parent to remain idle before sending a
+   keepalive message.  This value may be adjusted on a live system.
+
+tbon.torpid_max
+   The amount of time (in RFC 23 Flux Standard Duration format) that a broker
+   will wait for an idle TBON child connection to send messages before
+   declaring it torpid (unresponsive).  A value of 0 disables torpid node
+   checking.  Torpid nodes are automatically drained and require manual
+   undraining with :man1:`flux-resource`.  This value may be adjusted on a
+   live system.
+
 
 SOCKET ATTRIBUTES
 =================
@@ -197,6 +210,8 @@ RESOURCES
 =========
 
 Flux: http://flux-framework.org
+
+RFC 23: Flux Standard Duration: https://github.com/flux-framework/rfc/blob/master/spec_23.rst
 
 
 SEE ALSO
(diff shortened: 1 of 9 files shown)


garlick:
I added documentation to the flux-broker-attributes(7) man page.  Was going to add to the config file docs in section 5 but realized some reorganization is probably needed there, and that should be done in a different PR to fit release notes, after some discussion of how we'd like it to look.

Dropping the WIP prefix as I think this is ready for a review.

src/broker/overlay.c
```diff
@@ -349,45 +349,58 @@ bool overlay_peer_is_torpid (struct overlay *ov, uint32_t rank)
...
-                    if (child->torpid) {
-                        flux_log (ov->h,
-                                  LOG_ERR,
-                                  "broker on %s (rank %lu) is responsive now",
-                                  flux_get_hostbyrank (ov->h, child->rank),
-                                  (unsigned long)child->rank);
-                        child->torpid = false;
-                    }
+            }
+            else { // duration < toprid_max OR torpid_max == 0
```
grondo, line 397: minor typo: `toprid_max`

doc/man7/flux-broker-attributes.rst
```diff
@@ -76,6 +76,19 @@ tbon.prefertcp
...
+tbon.torpid_min
+   The amount of time (in RFC 23 Flux Standard Duration format) that a broker
+   will allow the connection to its TBON parent to remain idle before sending a
+   keepalive message.  This value may be adjusted on a live system.
+
+tbon.torpid_max
+   The amount of time (in RFC 23 Flux Standard Duration format) that a broker
+   will wait for an idle TBON child connection to send messages before
+   declaring it torpid (unresponsive).  A value of 0 disables torpid node
+   checking.  Torpid nodes may be automatically drained and require manual
```
grondo, line 88: Instead of "Torpid nodes _may_ be automatically drained..." should it be "Torpid nodes that are not already drained are drained automatically..." or something similar?

grondo (approved):
Looks great! This feels like a big leap forward in functionality with relatively little work, which speaks highly of the broker and groups architecture!

It also feels like a big deal that this feature is available to single user/batch jobs. For example, long running DAT jobs can now start the TBON configured as a flat tree, a few extra nodes, and have down nodes automatically drained so they can survive a few node crashes. That seems like a revolutionary feature!

garlick:
Thanks @grondo!  I fixed your suggestions and rebased on current master.  Setting MWP.

PR #4054, opened 2022-01-19
flux-perilog-run: improve usefulness of logging when prolog/epilog fails

grondo:
When attempting to set up prolog/epilog support on a system, it was discovered that errors from remote prolog/epilog execution were being copied to the logs without an identifying rank or hostname, which made diagnosis of errors unnecessarily difficult.

This PR improves the logging of the `flux perilog-run` script when used to execute per-rank `prolog` and `epilog` by including the rank and hostname in any lines of stderr generated by these scripts, along with other small improvements.


garlick (approved):
Looks reasonable.

grondo:
Thanks! Had to fix a broken `&&` chain in the test and force pushed.

PR #4056, opened 2022-01-19
configure: Add python docutils check

### LINKED ISSUE #4055: FAIL: t0001-basic.t 5 - we can find a flux binary
recently hit this.

```
Usage: flux [OPTIONS] COMMAND ARGS
  -v, --verbose          Be verbose about environment and command search
  -V, --version          Display command and component versions
  -p, --parent           Set environment of parent instead of current instance

For general Flux documentation, please visit
    https://flux-framework.readthedocs.io

flux: /g/g0/achu/chaos/git/flux-framework/flux-core/etc/flux/help.d/core.json::1: '[' or '{' expected near end of file
flux: /g/g0/achu/chaos/git/flux-framework/flux-core/etc/flux/help.d/core.json: failed to read content
: Success
```

`core.json` is empty.

i suspect the issue is the machine I got this on doesn't have a python3 `docutils` installed, thus an issue with `doc/conf.py`.

Multiple options:

- `configure.ac` - require docutils? don't generate help if `docutils` not found?
- help code should check if `core.json` is non-empty and avoid error?
- don't generate options if `conf.py` doesn't work?

perhaps first option is easiest, just require it when building.  Should update readme as well.

Edit: as an aside, `not ok 97 - flux-help command list can be extended` fails too, same core problem

chu11:
Problem: docutils is required for building flux help output. Test
errors will occur if it is not available.

Solution: Add a configure check for python docutils.  Also add
it as a package requirement in README.md.

Fixes #4055


chu11:
note that the version I added in the configure check is the one in the RHEL7 package.  Not sure how much the version number matters for `docutils`.

grondo:
Isn't it only required if building docs? So added to the following test?
```m4
#  If --enable-docs=yes, but no doc generator found,
#   then error immediately:
#
AS_IF([test "x$enable_docs" = "xyes" -a "x$sphinx" = "xfalse"],[
  AC_MSG_ERROR([--enable-docs used but no document generator found!])
])
```

Also included in the footnote about docs in the `README.md` (Apologies if I missed something obvious)

grondo:
Ah! Sorry, I missed that there is more detail in an issue. I'll go look over there.

grondo:
Let's try the strategy proposed in #4055 to allow `gen-cmdhelp.py` to read `conf.py` without `docutils`, then in this PR make `docutils` only a requirement when building docs. Does that sound reasonable?

chu11:
ok, redid this PR.  Put a `docutils` check in, although slightly tweaked from what I had before.

I split out `man_pages` from `conf.py` into its own file, so that the `gen-cmdhelp.py` script not longer requires `docutils`.

chu11:
hmmm, `ci / bionic - gcc-8,content-s3,distcheck` builder failed with

```
   Making all in doc
  Making all in .
    BUILD     manpages
  Makefile:1517: recipe for target 'man1/flux.1' failed
  make[3]: *** [man1/flux.1] Error 2
  Makefile:1229: recipe for target 'all-recursive' failed
  make[2]: *** [all-recursive] Error 1
  Makefile:587: recipe for target 'all-recursive' failed
  make[1]: *** [all-recursive] Error 1
  Makefile:793: recipe for target 'distcheck' failed
  make: *** [distcheck] Error 1
```

It's the only builder that failed.  Hmmmm.

chu11:
oh duh! it's probably b/c I didn't add `manpages.py` to the `EXTRA_DIST`, re-pushed (went ahead and squashed since the change is so tiny).

chu11:
re-pushed, i had a stupid whitespace fix that had been sitting in a branch for awhile.  just tacked it on.

chu11:
rebased and re-pushed, fixing up the conflict from #4057

grondo (approved):
LGTM! I tested a few scenarios and everything worked as expected.

sphinx packages require docutils, so I had to install sphinx then manually remove docutils to get into a state where one is available and the other is not. But even then, the configure logic worked great.

PR #4057, opened 2022-01-19
doc: add man pages for config files

garlick:
This PR fills in some missing man pages in section 5.  For a start it adds `flux-config(5)` which is an overview of config files, and pages for a couple of additional tables, including for `tbon` so we can document the new "torpid" knobs.  I wanted to see if this pattern looks OK before adding more (hence WIP).  I was thinking we could have a man page for each top level table, named with the pattern `flux-config-NAME.5` where NAME is the table name.  That seem reasonable?

Then we can reference these pages from the configuration section of the [Administrator's Guide](https://flux-framework.readthedocs.io/en/latest/adminguide.html).

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -288,7 +288,9 @@ MAN5_FILES_PRIMARY = \
 	man5/flux-config-tbon.5 \
 	man5/flux-config-exec.5 \
 	man5/flux-config-resource.5 \
-	man5/flux-config-archive.5
+	man5/flux-config-archive.5 \
+	man5/flux-config-job-manager.5
+
 
 MAN7_FILES = $(MAN7_FILES_PRIMARY)
 
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -285,7 +285,8 @@ MAN5_FILES_PRIMARY = \
 	man5/flux-config.5 \
 	man5/flux-config-access.5 \
 	man5/flux-config-bootstrap.5 \
-	man5/flux-config-tbon.5
+	man5/flux-config-tbon.5 \
+	man5/flux-config-exec.5
 
 MAN7_FILES = $(MAN7_FILES_PRIMARY)
 
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -287,7 +287,8 @@ MAN5_FILES_PRIMARY = \
 	man5/flux-config-bootstrap.5 \
 	man5/flux-config-tbon.5 \
 	man5/flux-config-exec.5 \
-	man5/flux-config-resource.5
+	man5/flux-config-resource.5 \
+	man5/flux-config-archive.5
 
 MAN7_FILES = $(MAN7_FILES_PRIMARY)
 
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -286,7 +286,8 @@ MAN5_FILES_PRIMARY = \
 	man5/flux-config-access.5 \
 	man5/flux-config-bootstrap.5 \
 	man5/flux-config-tbon.5 \
-	man5/flux-config-exec.5
+	man5/flux-config-exec.5 \
+	man5/flux-config-resource.5
 
 MAN7_FILES = $(MAN7_FILES_PRIMARY)
 
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -284,7 +284,8 @@ MAN5_FILES = $(MAN5_FILES_PRIMARY)
 MAN5_FILES_PRIMARY = \
 	man5/flux-config.5 \
 	man5/flux-config-access.5 \
-	man5/flux-config-bootstrap.5
+	man5/flux-config-bootstrap.5 \
+	man5/flux-config-tbon.5
 
 MAN7_FILES = $(MAN7_FILES_PRIMARY)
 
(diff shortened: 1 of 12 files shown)


grondo:
This sounds like a good strategy to me!

garlick:
OK, thanks!  I dropped the WIP and filled in the other missing pages.

doc/man5/flux-config-exec.rst
```diff
@@ -0,0 +1,46 @@
...
+
+KEYS
+====
+
+imp
+   (optional) Set the path to the IMP (Independent Minister of Privilege)
+   helper program, as described in RFC 15, so that jobs may be launched with
+   the credentials of the guest user that submitted them.  If unset, only
+   jobs submitted by the instance owner may be executed.
+
```
grondo, line 24: `job-exec` also supports a `job-shell` key to configure a different path for the job shell (unlikely to be necessary) and a `kill-timeout` key to configure the time between SIGTERM andSIGKILL when a job is terminated.

I'm fine updating this in a future PR.

doc/man5/flux-config.rst
```diff
@@ -0,0 +1,79 @@
...
+
+TOML: Tom's Obvious Minimal Language: https://toml.io/en/
+
+
+SEE ALSO
+========
+
+:man1:`flux-broker`, :man5:`flux-config-access`, :man5:`flux-config-bootstrap`,
+:man5:`flux-config-tbon`, :man5:`flux-config-exec`,
+:man5:`flux-config-resource`
```
grondo, line 79: We could also add `flux-config-job-manager(5)` and `flux-config-archive(5)` (for the `job-archive` module). I'm fine with adding those in another PR if we want to get the basic stuff in.

grondo (approved):
LGTM. I noted some missing docs, but it seems like those could be added in another PR if you are done writing docs for now :wink:

doc/man5/flux-config-exec.rst
garlick: Ah we can get that in here.  I was working backward from the admin guide and so I missed that.

doc/man5/flux-config.rst
garlick: Yeah lets add that

doc/man5/flux-config-exec.rst
grondo: FYI - I did 
```console
$ git grep flux_conf_unpack | grep -v test/
src/broker/boot_config.c:    if (flux_conf_unpack (cf,
src/broker/broker.c:    if (flux_conf_unpack (conf, NULL, "{s:{}}", "bootstrap") == 0) {
src/broker/brokercfg.c:    if (flux_conf_unpack (flux_get_conf (h), NULL, "o", &conf) < 0)
src/broker/overlay.c:        if (flux_conf_unpack (flux_get_conf (ov->h),
src/common/libflux/conf.c:int flux_conf_unpack (const flux_conf_t *conf,
src/common/libflux/conf.h:int flux_conf_unpack (const flux_conf_t *conf,
src/modules/connector-local/local.c:    if (flux_conf_unpack (conf,
src/modules/content-s3/content-s3.c:    if (flux_conf_unpack (conf,
src/modules/job-archive/job-archive.c:    if (flux_conf_unpack (flux_get_conf (ctx->h),
src/modules/job-exec/exec.c:    if (flux_conf_unpack (flux_get_conf (h),
src/modules/job-exec/exec.c:    if (flux_conf_unpack (flux_get_conf (h),
src/modules/job-exec/job-exec.c:    if (flux_conf_unpack (flux_get_conf (h),
src/modules/job-manager/jobtap.c:    if (flux_conf_unpack (conf,
src/modules/job-manager/journal.c:    if (flux_conf_unpack (flux_get_conf (ctx->h),
src/modules/job-manager/plugins/perilog.c:    if (flux_conf_unpack (flux_get_conf (h),
src/modules/resource/inventory.c:    if (flux_conf_unpack (flux_get_conf (h),
src/modules/resource/resource.c:    if (flux_conf_unpack (conf,
```

garlick:
I added a few more pages, although I did not include the perilog config or the content-s3 config as I felt like I was already being a bit invasive in the very good doc space of the jobtap plugins, and I thought it might be a bit premature to start documenting content-s3.  Those could be added in future PRs as time permits I guess.

doc/Makefile.am
```diff
@@ -284,7 +284,8 @@ MAN5_FILES = $(MAN5_FILES_PRIMARY)
 MAN5_FILES_PRIMARY = \
 	man5/flux-config.5 \
 	man5/flux-config-access.5 \
-	man5/flux-config-bootstrap.5
+	man5/flux-config-bootstrap.5 \
```
grondo, line 287: Commit message refers to `[access]` table instead of `[tbon]`

doc/Makefile.am
```diff
@@ -285,7 +285,8 @@ MAN5_FILES_PRIMARY = \
 	man5/flux-config.5 \
 	man5/flux-config-access.5 \
 	man5/flux-config-bootstrap.5 \
-	man5/flux-config-tbon.5
+	man5/flux-config-tbon.5 \
+	man5/flux-config-exec.5
```
grondo: very minor: commit message uses lower case `problem` when all other commit messages use upper-case.

grondo (approved):
Thanks for doing this! LGTM, just a couple very minor comments on commit messages.

garlick:
Thanks!  I fixed those things and forced a push.

grondo:
Feel free to set MWP. Then @chu11 can rebase #4056 since it moves the `man_pages` to a separate file.

garlick:
OK, pushing one small correction then MWP.  Thanks!

Issue #2139, opened 2019-04-25
flux-proxy leaves shell is bad shape after instance shutsdown


SteVwonder:
If I start a flux instance, connect into it via `flux-proxy` (locally or via ssh), and then kill the flux instance, the shell that ran `flux-proxy` is left in bad shape.  I've had this happen under both bash and zsh.  Under bash, I can usually run `reset` and everything is ok.  Under zsh, the shell will happily consume characters and then display seemingly random characters as well as mutate previously typed characters. It is quite bizarre. [Video of the experience](https://gfycat.com/CreamyBriefDaddylonglegs).  This makes typing multi-character commands like `reset` or `exit` impossible.  I'm not sure exactly what is causing it, but I'm wondering if there is something we can do on our end to prevent zsh/bash from ending up in such a tragic state.

grondo:
This was annoying me lately so I looked into it for a bit.

I don't think it is the shell from which you ran `flux-proxy` that ends up in bad shape, it is the current shell *spawned* by `flux-proxy` that becomes unusable.

This is actually a generic issue with (at least) bash when its parent is killed. E.g.:
```
 grondo@fluke108:~$ bash
 grondo@fluke108:~$ bash
 grondo@fluke108:~$ kill -9 $PPID
 grondo@fluke108:~$ Killed
 grondo@fluke108:~$ 
 grondo@fluke108:~$ llogout
Shared connection to fluke closed.
```

I'm guessing this is because when the parent of the subshell forcibly exits it is reparented to init and results in terminal confusion if not loss 

E.g., after `bash; bash` I see:
```
USER      PPID   PID  PGID  SESS CMD
grondo   13414 24790 24790 24790 -bash
grondo   24790 24906 24906 24790 bash
grondo   24906 24910 24910 24790 bash
```

If I kill my parent from the current foreground shell, the interactive shell is reparented to init:
```
grondo   13414 24790 24790 24790 -bash
grondo       1 24910 24910 24790 bash
```

And if I try to strace that shell (24910), it immediately gets `EIO` reading stdin, prints `exit` and exits:

```
 grondo@fluke108:/proc$ strace -p 24910
strace: Process 24910 attached
read(0, 0x7fffffffc9e7, 1)              = -1 EIO (Input/output error)
rt_sigprocmask(SIG_BLOCK, [INT], [], 8) = 0
[snip]
write(2, "exit\n", 5)                   = 5
[...]

PR #4058, opened 2022-01-20
flux-proxy: improve experience when proxied Flux instance terminates

grondo:
This PR contains a few fixes (already described in #2139) to improve the experience of `flux proxy` users when the proxied Flux instance terminates.

Currently, `flux proxy` exits, leaving the child process (usually an interactive shell) to be reparented to PID 1 and lose its terminal. This typically causes the user's tty to get into a bad state. Sometimes I've even had myself logged out of my connection to the host from which I originally ran `flux proxy`.

This PR adds a couple mitigations for these problems:

  * Attempt to restore terminal when `flux proxy` gets an error. This seems to help me in most cases, though nothing is guaranteed. The code makes a best-effort attempt to do the restore.
  * Attempt to keep `flux proxy` running until the child process exits. This makes it so that the child process is not reparented to init and doesn't lose its controlling tty. This possibly obviates the need for the first approach above, but since restoring the terminal shouldn't hurt anything, we keep that for now.
  * Add a `-k, --kill` option that causes `flux-proxy` to send `SIGTERM` to the child process on error. This could be useful to ensure cleanup of the child when the proxied Flux instance terminates.

This can be considered a proposal PR. I'm open to any changes or adjustments.

### DIFF
--- a/src/cmd/builtin/proxy.c
+++ b/src/cmd/builtin/proxy.c
@@ -281,6 +281,14 @@ static void version_check (flux_t *h, bool force)
     }
 }
 
+static void proxy_command_destroy_usock_and_router (struct proxy_command *ctx)
+{
+    usock_server_destroy (ctx->server); // destroy before router
+    ctx->server = NULL;
+    router_destroy (ctx->router);
+    ctx->router = NULL;
+}
+
 static int cmd_proxy (optparse_t *p, int ac, char *av[])
 {
     int n;
@@ -351,13 +359,26 @@ static int cmd_proxy (optparse_t *p, int ac, char *av[])
      */
     save_terminal_state ();
     if (flux_reactor_run (r, 0) < 0) {
+        if (errno == ECONNRESET)
+            log_msg ("Lost connection to Flux");
+        else
+            log_err ("flux_reactor_run");
+        if (!optparse_hasopt (p, "nohup")) {
+            log_msg ("Sending SIGHUP to child processes");
+            flux_future_destroy (flux_subprocess_kill (ctx.p, SIGHUP));
+            flux_future_destroy (flux_subprocess_kill (ctx.p, SIGCONT));
+        }
+        proxy_command_destroy_usock_and_router (&ctx);
+
+        /*
+         * Wait for child to normally terminate
+         */
+        flux_reactor_run (r, 0);
         restore_terminal_state ();
-        log_err ("flux_reactor_run");
         goto done;
     }
 done:
-    usock_server_destroy (ctx.server); // destroy before router
-    router_destroy (ctx.router);
+    proxy_command_destroy_usock_and_router (&ctx);
 
     if (ctx.exit_code)
         exit (ctx.exit_code);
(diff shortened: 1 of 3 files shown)


chu11 (approved):
nothing stood out to me

src/cmd/builtin/proxy.c
```diff
@@ -351,13 +359,26 @@ static int cmd_proxy (optparse_t *p, int ac, char *av[])
      */
     save_terminal_state ();
     if (flux_reactor_run (r, 0) < 0) {
+        if (errno == ECONNRESET)
+            log_msg ("Lost connection to Flux");
+        else
+            log_err ("flux_reactor_run");
+        if (optparse_hasopt (p, "kill")) {
+            log_msg ("Sending SIGTERM to child processes");
+            flux_future_destroy (flux_subprocess_kill (ctx.p, SIGTERM));
```
garlick, line 368: I'm not sure of myself here, but would SIGHUP be a better choice than SIGTERM?  Bash ignores SIGTERM but exits on SIGHUP after resending SIGHUP to all jobs.  For an interactive shell, it seems a little nicer to have everything clean up when the connection goes away, instead of being left with a shell with a non-functional FLUX_URI in its `env` (making my nice ƒ(s=4,d=0,builddir) PS1 sad.

Along those lines, what about reversing the sense of the kill option so that signaling is the default, but let it be disabled in cases where it's causing problems?

I realize SIGHUP is supposed to mean "interactive user went away" not "some application that happens to be the center of this developer's universe bu likely not everyone's went away" so it's possible this is a bad idea.

garlick (approved):
Generally great improvement!  I did have a couple of questions about the signal policy, but honestly this is fine if you disagree so I'm marking approved.

src/cmd/builtin/proxy.c
grondo: Hm, good question. `SIGTERM` seemed to work for me (it causes my interactive shell to immediately exit). I didn't even consider `SIGHUP`, but now that I look at the `bash(1)` man page I see you are right (at least for bash), so I think I agree with your assessment. I'll test it out.

I'm fine with reversing the sense of the `-k, --kill` option. I'll rename it `-k, --no-kill` if that sounds ok?

grondo: Ah, there's a nice history delivering `SIGHUP` to processes when their "controlling" process exits of which I was only vaguely aware:

https://pubs.opengroup.org/onlinepubs/9699919799/functions/_Exit.html#tag_16_01_08_02

I wonder if we should send SIGHUP+SIGCONT and make believe we're a shell or kernel?

grondo: Ok, I made the suggested change (and added `SIGCONT` in case the proces is stopped).

The option to avoid `SIGHUP+SIGCONT` is now `-n, --nohup`.

Updated tests and the manpage. I added the following spiel to the manpage:

```

       If the connection to the Flux instance is lost, for  example  when  the
       target instance terminates, flux proxy will emit an error message, send
       SIGHUP and SIGCONT to the spawned shell or other process, and wait  for
       it  to  terminate  before exiting.  The delivery of signals can be dis‐
       abled with the -n, --nohup option, but be aware that Flux commands run‐
       ning  under  a  flux  proxy  which  has lost its connection will likely
       result in errors.
```

garlick: I like the way `--nohup` turned out there.  Nice.

grondo:
I've set MWP.

PR #4059, opened 2022-01-20
sharness: unset FLUX_CONF_DIR for all tests

jameshcorbett:
Problem: if FLUX_CONF_DIR is set in the current environment,
tests which assume that the broker has an empty config
will fail. This includes t0013-config-file.t.

Remove FLUX_CONF_DIR from all tests in flux-sharness.sh.
Tests can individually set this variable if they need it.

The following test was breaking:

```
corbett8@docker-desktop:~/flux-core/t$ ./t0013-config-file.t -d -v
sharness: loading extensions from /home/corbett8/flux-core/t/sharness.d/01-setup.sh
sharness: loading extensions from /home/corbett8/flux-core/t/sharness.d/flux-sharness.sh
expecting success: 
	flux broker ${ARGS} flux lsattr -v | grep config.path >noconfig.out &&
	egrep "^config.path.*-$" noconfig.out

not ok 1 - flux broker works with no config
#	
#		flux broker ${ARGS} flux lsattr -v | grep config.path >noconfig.out &&
#		egrep "^config.path.*-$" noconfig.out
#
```


grondo (approved):
Nice find! Thanks!

jameshcorbett:
Thanks, setting MWP.

Issue #3964, opened 2021-11-16
libsubprocess: do not log errors with flux_log()


grondo:
As mentioned in #3906, libsubprocess currently logs errors internally with `flux_log_error()`, the result of which is perhaps not that useful, e.g.:
```
2021-11-16T15:26:48.028464Z job-exec.err[0]: remote_exec_cb: flux_rpc_get_unpack: rank 3: No route to host
```
In this case the error will be returned to the job-exec module which will terminate the job with an appropriate error message, so the message above including internal function names is redundant.

Additionally, when running non-module clients (like `flux exec`) it probably doesn't make much sense to dump errors to the flux dmesg.

chu11:
are you thinking no logging?  or logging to stderr or something else?

garlick:
My $0.02 would be to avoid logging (especially at LOG_ERR) in cases where an error is returned to a client/user who may be in a better position to log with useful context.

PR #4060, opened 2022-01-20
libsubprocess: fix excess logging and logging corner cases

chu11:
Initially spawned by issue #3964, ending up fixing many minor things.

- some lines > 80 chars
- some invalid error messages
- do not log common EHOSTUNREACH errors
- remove duplicate logging
- fix potential errno clobbering error
- (and then finally) when error is returned to the user, do not log to LOG_ERR, log to LOG_DEBUG to avoid likely redundant error logs

### DIFF
--- a/src/common/libsubprocess/local.c
+++ b/src/common/libsubprocess/local.c
@@ -165,7 +165,7 @@ static int channel_local_setup (flux_subprocess_t *p,
     int buffer_size;
 
     if (!(c = channel_create (p, output_f, name, channel_flags))) {
-        flux_log_error (p->h, "calloc");
+        flux_log_error (p->h, "channel_create");
         goto error;
     }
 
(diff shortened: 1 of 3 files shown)


grondo (approved):
This seems like good cleanup. I wonder if we should go even further and switch libsubprocess to something like `src/common/libutil/llog.h`, so that the caller can decide how log messages should be handled. Even though log messages have been demoted to LOG_DEBUG, it seems wrong to have a user library logging to the Flux broker log, e.g. in the case of `flux exec`.

However, this is nice enough cleanup it would be nice to merge this now and deal with user defined logging later.

chu11:
> This seems like good cleanup. I wonder if we should go even further and switch libsubprocess to something like src/common/libutil/llog.h, so that the caller can decide how log messages should be handled. Even though log messages have been demoted to LOG_DEBUG, it seems wrong to have a user library logging to the Flux broker log, e.g. in the case of flux exec.

Good point.  The early days of `libsubprocess` were predominantly for `job-exec` and broker-ish kinda things, which is why `flux_log()` was used.  And I don't think we had `llog` at the time.  Lets merge this and we'll put `llog` as a TODO item.

grondo:
>  Lets merge this and we'll put llog as a TODO item.

Sounds good!

chu11:
@mergifyio refresh

PR #4062, opened 2022-01-21
improve error message from flux-proxy and flux-jobs for invalid and unknown jobids

### LINKED ISSUE #3882: flux-jobs: failure to parse jobid error is generic
I accidentally used the `flux jobs` command rather than `flux job` command when trying to cancel all my jobs, and I got this error:
```
ƒ(s=1,d=0) fluxuser@0159d36f7779:/workspaces/flux-coral2/t$ flux jobs cancelall -f
flux-jobs: ERROR: [Errno 22] EINVAL: Invalid argument
ƒ(s=1,d=0) fluxuser@0159d36f7779:/workspaces/flux-coral2/t$ flux jobs cancelall
flux-jobs: ERROR: [Errno 22] EINVAL: Invalid argument
ƒ(s=1,d=0) fluxuser@0159d36f7779:/workspaces/flux-coral2/t$ flux jobs cancelall -h
flux-jobs: ERROR: [Errno 22] EINVAL: Invalid argument
```

It might be more instructive to have an error that calls out the exact argument (`cancelall`) that cause the error and that it is an invalid/malformed jobid.  Something like:
```
flux-jobs: ERROR: Failed to parse jobid (cancelall)
```

chu11:
Problem: When the user inputs an illegal job id the error
message is a very generic:

flux-jobs: ERROR: [Errno 22] EINVAL: Invalid argument

which can be confusing and unclear what was done wrong.

Solution: Improve error message to:

flux-jobs: error: argument JOBID: illegal JobID <invalid arg>

Add a test case.

Fixes #3882

### DIFF
--- a/src/bindings/python/flux/job/JobID.py
+++ b/src/bindings/python/flux/job/JobID.py
@@ -55,7 +55,10 @@ class JobID(int):
         if isinstance(value, int):
             jobid = value
         else:
-            jobid = id_parse(value)
+            try:
+                jobid = id_parse(value)
+            except OSError:
+                raise ValueError(f"{value} is not a valid Flux jobid")
         self = super(cls, cls).__new__(cls, jobid)
         self.orig_str = str(value)
         return self
--- a/src/cmd/flux-pstree.py
+++ b/src/cmd/flux-pstree.py
@@ -164,13 +164,21 @@ def load_tree(
     #
     attrs = flux.job.list.VALID_ATTRS
     try:
-        jobs = (
-            JobList(flux.Flux(uri), ids=jobids, filters=filters, attrs=attrs)
-            .fetch_jobs()
-            .get_jobs()
-        )
+        jobs_rpc = JobList(
+            flux.Flux(uri), ids=jobids, filters=filters, attrs=attrs
+        ).fetch_jobs()
+        jobs = jobs_rpc.get_jobs()
     except (OSError, FileNotFoundError):
         return tree
+
+    #  Print all errors accumulated in JobList RPC:
+    #  fetch_jobs() may not set errors list, must check first
+    if hasattr(jobs_rpc, "errors"):
+        try:
+            for err in jobs_rpc.errors:
+                print(err, file=sys.stderr)
+        except EnvironmentError:
+            pass
 
     #  Since the executor cannot be used recursively, start one per
     #   loop iteration. This is very wasteful but greatly speeds up
--- a/t/t2800-jobs-cmd.t
+++ b/t/t2800-jobs-cmd.t
@@ -299,7 +299,7 @@ test_expect_success 'flux jobs can take specific IDs in any form' '
 	test_cmp ids.specific.expected ids.specific.out
 '
 
-test_expect_success 'flux-jobs error on bad IDs' '
+test_expect_success 'flux-jobs error on unknown IDs' '
 	flux jobs --suppress-header 0 1 2 2> ids.err &&
 	count=`grep -i unknown ids.err | wc -l` &&
 	test $count -eq 3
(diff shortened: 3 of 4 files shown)


chu11:
it just occurred to me, could maybe make the `JobID_check()` function global and use in other places (`flux-pstree` minimally)

src/cmd/flux-jobs.py
```diff
@@ -198,6 +198,13 @@ def __call__(self, parser, namespace, values, option_string=None):
         setattr(namespace, "filtered", True)
 
 
+def JobID_check(arg):
```
grondo, line 201: It would be better to do this in the `JobID` constructor instead. If we raise `ValueError` directly, then `argparse` will kind of do the right thing automatically wherever the type of an argument is `JobID`, e.g.

```diff
diff --git a/src/bindings/python/flux/job/JobID.py b/src/bindings/python/flux/job/JobID.py
index 2debc6f3e..a0d8dd3e5 100644
--- a/src/bindings/python/flux/job/JobID.py
+++ b/src/bindings/python/flux/job/JobID.py
@@ -55,7 +55,10 @@ class JobID(int):
         if isinstance(value, int):
             jobid = value
         else:
-            jobid = id_parse(value)
+            try:
+                jobid = id_parse(value)
+            except OSError:
+                raise ValueError(f"{value} is not a valid Flux jobid")
         self = super(cls, cls).__new__(cls, jobid)
         self.orig_str = str(value)
         return self
```

Then:

```console
$ flux jobs g1234
usage: flux-jobs [-h] [-a] [-A] [-c N] [-f STATE|RESULT] [-n]
                 [-u [USERNAME|UID]] [-o FORMAT] [--color WHEN] [-R] [-L N]
                 [--recurse-all] [--threads N] [--stats] [--stats-only]
                 [JOBID [JOBID ...]]
flux-jobs: error: argument JOBID: invalid JobID value: 'g1234'
```

grondo (commented):
I think the `JobID` class should be modified instead (suggestion inline)

chu11:
re-pushed, went with your suggested fix @grondo.  While adding tests for `flux-pstree`, I noticed that `flux-pstree` did not have a test for "unknown jobid".  Added a test, and noticed that `flux-pstree` was silent when an unknown / invalid jobid was input by the user.  Figured we might want to output an error similar to how we do in `flux-jobs`.

Fix was a little tricky, b/c `JobList().fetch_jobs().get_jobs()` will not necessarily set `.errors`, while (in `flux-jobs`) the approach of calling `JobList().jobs()` will set `.errors`.  

I went back and forth on several approaches, and eventually went with renaming `.jobs()` to `.jobinfos()`, which I think is more consistent with other naming we do in the python list code.  Then created a new `.jobs()` function, which will set `.errors` appropriately.

src/bindings/python/flux/job/list.py
```diff
@@ -251,15 +251,18 @@ def fetch_jobs(self):
...
         If the Future object returned by JobList.fetch_jobs has not yet been
         fulfilled (e.g. is_ready() returns False), then this call may block.
-        Otherwise, returns a list of JobInfo objects for all jobs returned
-        from the underlying job listing RPC.
         """
         rpc = self.fetch_jobs()
         jobs = rpc.get_jobs()
         if hasattr(rpc, "errors"):
             self.errors = rpc.errors
+        return jobs
```
grondo, line 263: While I agree with your assessment that the new naming is more consistent, we do have existing users of this code and I don't think we should change the interface on them.

chu11: fair enough.  alternate ideas?

Perhaps we add `jobsinfo()` but leaves `jobs()` as a legacy interface?  (unclear how standard the `@deprecated` decorator is at the moment).  Then in the case when the caller whats the raw jobs data .. or `jobdicts()`?

grondo: I think fetching the raw list of jobs is not going to be a common case. How about we leave the interface as is for now and just fix `flux pstree`

chu11:
re-pushed, added specific fix into `flux-pstree` to handle unknown / invalid job ids.

PR #4064, opened 2022-01-21
testsuite: fix non-bourne shell test failure

chu11:
Problem: Test in t1105-proxy failed because of quoting/escape
issues with non-bourne shells.

Solution: Set SHELL to /bin/sh in test.


grondo (approved):
Good catch, sorry about that!

Makes me wonder if we shouldn't have one of the CI builds run with SHELL=/bin/csh or something...

PR #4069, opened 2022-01-25
improve flux-config-bootstrap(5)

garlick:
The documentation in flux-config-bootstrap(5) is really minimal and probably not all that useful.  This PR expounds a bit and adds a simpler example.

### DIFF
--- a/doc/man5/flux-config-bootstrap.rst
+++ b/doc/man5/flux-config-bootstrap.rst
@@ -30,7 +30,9 @@ enable_ipv6
 curve_cert
    (optional) Path to a CURVE certificate generated with
    :man1:`flux-keygen`.  The certificate should be identical on all
-   broker ranks.  It is required for instance sizes > 1.
+   broker ranks.  It is required for instance sizes > 1.  The file should
+   be owned by the instance owner (e.g. `flux` user) and only readable by
+   that user.
 
 default_port
    (optional) The value is an integer port number that is substituted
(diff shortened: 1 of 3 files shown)


doc/man5/flux-config-bootstrap.rst
```diff
@@ -61,6 +63,35 @@ hosts
    substitutions work here as well.
 
 
+ZEROMQ ENDPOINTS
+================
+
+In this context, ZeroMQ endpoint URIs normally use the :man7:`zmq_tcp`
```
grondo: Did you mean `:linux:man7:` (flux-core doesn't have a `zmq_tcp(7)` manpage AFAICS)

grondo (approved):
Nice improvement!

doc/man5/flux-config-bootstrap.rst
garlick: Just caught that and forced a push - you were too quick! :-)

Although I note that :linux: doesn't have the zmq pages, so the link is broken in the readthedocs preview.  Better than hitting the flux-undocumented page tough I guess.

grondo: Ah, sorry I thought I had deleted that comment but it must have been cached somehow.

I wonder if it is our choice of "linux" manpage provider. `zmq_tcp(7)` does seem to be available [here](https://linux.die.net/man/7/zmq_tcp)

garlick: I'll try changing it and see how that goes, thanks!

grondo:
Seems fine to me! I tried to run `sphinx-build -M linkcheck` with limited success. However, it did catch a broken manpage linke to `inotify(2)`, but that appears to be because `inotify` is in section 7 (in `man3/flux_stat_watcher_create.rst`.

garlick:
Good catch, I'll tack on a fix for that.

garlick:
Thanks for the review - setting MWP.

Issue #3346, opened 2020-11-17
job-exec module needs a rewrite and new protocol


grondo:
There are several (more than several) open issues against the job-exec module. This module really needs a complete re-implementation, perhaps first with a thoughtful RFC describing the protocol so we avoid corner cases.

Outstanding issues in the current design include:

 * restart of rank 0 broker not supported (#2874)
 * job epilog support is not currently enabled and does not support "partial release" (#2317, #2943)
 * job shell and early standard output/error is lost (#2206, #3185)
 * launching all shells from rank 0 of instance is not scalable

The following use case should also be considered in the design:

 * exec system bypass for special workloads (#2924)

I think moving the launch of job shells off of rank 0 to the first rank of the job will help a lot with job throughput, as well as allowing the exec module a way to possible restart on rank 0. We'll have to think about how the job-exec system rediscovers running jobs at restart.

grondo:
This work still needs to be done, but some of the requirements have changed.

 * Now that the strategy to support broker restart includes `libsdprocess`, moving execution of the job shell directly to the involved broker ranks is required because the `sdexec` exec implementation can only work locally. (not only to allow restart of rank 0 nor only for scalability). It is an open question whether normal execution should also work this way
 * Some of the referenced issues above have already been solved, but we should not regress them (e.g. reporting of early launch errors, handling job shell output) 


Some other items to think about in the redesign:

 * Partial release (#2204)
 * Locally initiated prolog/epilog replacing need for ad-hoc solution

I'm sure there are other things I'm missing at the moment.

chu11:
Wanted to document a side discussion had with @grondo, we also will want a way to configure which "exec" mechanism we want to use.  There should be a:

- default mechanism (subprocess, systemd, etc.)
- allow instance owner to use a non-default (e.g. testexec or something else)
- but do not allow guests to chose non-default

the current mechanism in `job-exec` for selecting which exec method to use makes the above somewhat tricky to do, and thus has temporarily not been done

(Related to #3970 but not quite the same)

PR #4070, opened 2022-01-25
job-exec: support new sdexec job launch plugin 

chu11:
Built on top of #3864.  Hope is that we can add @grondo's `job-shell` reconnect on top of this, then hopefully add a simple test that will allow flux to restart and the job to be "re-attached".

### DIFF
--- a/doc/man5/flux-config-exec.rst
+++ b/doc/man5/flux-config-exec.rst
@@ -22,6 +22,16 @@ imp
    the credentials of the guest user that submitted them.  If unset, only
    jobs submitted by the instance owner may be executed.
 
+method
+   (optional) Run job shell under a specific mechanism other than the default
+   forked subprocesses.  Potential configurations:
+
+   systemd
+
+   Run job shells are run under systemd, the job shell may be able to
+   survive an unexpected broker shutdown and be recovered when the
+   broker is restarted.
+
 job-shell
    (optional) Override the compiled-in default job shell path.
 
--- a/etc/rc1
+++ b/etc/rc1
@@ -89,10 +89,14 @@ if test $RANK -eq 0 -a "${FLUX_SCHED_MODULE}" != "none" \
     flux module load ${FLUX_SCHED_MODULE:-sched-simple}
 fi
 
-if test $RANK -eq 0 -a -z "${FLUX_DISABLE_JOB_CLEANUP}"; then
+if test $RANK -eq 0; then
+    method=$(flux config get --default=notset exec.method)
+    if test "${method}" != "systemd" \
+            -a -z "${FLUX_DISABLE_JOB_CLEANUP}"; then
 	flux admin cleanup-push <<-EOT
 	flux queue stop --quiet
 	flux job cancelall --user=all --quiet -f --states RUN
 	flux queue idle --quiet
 	EOT
+    fi
 fi
--- a/src/modules/job-exec/Makefile.am
+++ b/src/modules/job-exec/Makefile.am
@@ -37,6 +37,10 @@ job_exec_la_SOURCES = \
 	testexec.c \
 	exec.c
 
+if HAVE_LIBSYSTEMD
+job_exec_la_SOURCES += sdexec.c
+endif
+
 job_exec_la_LDFLAGS = \
 	$(fluxmod_ldflags) \
 	-module
@@ -47,6 +51,10 @@ job_exec_la_LIBADD = \
 	$(top_builddir)/src/common/libflux-core.la \
 	$(JANSSON_LIBS) \
 	libbulk-exec.la
+
+if HAVE_LIBSYSTEMD
+job_exec_la_LIBADD += $(top_builddir)/src/common/libsdprocess/libsdprocess.la
+endif
 
 bulk_exec_SOURCES = \
 	test/bulk-exec.c
--- a/src/modules/job-exec/Makefile.am
+++ b/src/modules/job-exec/Makefile.am
@@ -30,6 +30,8 @@ job_exec_la_SOURCES = \
 	job-exec.c \
 	checkpoint.h \
 	checkpoint.c \
+	exec_config.h \
+	exec_config.c \
 	rset.c \
 	rset.h \
 	testexec.c \
--- a/src/modules/job-exec/exec.c
+++ b/src/modules/job-exec/exec.c
@@ -33,13 +33,11 @@
 #include <unistd.h>
 
 #include "job-exec.h"
+#include "exec_config.h"
 #include "bulk-exec.h"
 #include "rset.h"
 
 extern char **environ;
-static const char *default_cwd = "/tmp";
-static const char *default_job_shell = NULL;
-static const char *flux_imp_path = NULL;
 
 struct exec_ctx {
     const char * mock_exception;   /* fake exception */
@@ -72,40 +70,6 @@ static const char * exec_mock_exception (struct bulk_exec *exec)
     if (!ctx || !ctx->mock_exception)
         return "none";
     return ctx->mock_exception;
-}
-
-static const char *jobspec_get_job_shell (json_t *jobspec)
-{
-    const char *path = NULL;
-    (void) json_unpack (jobspec, "{s:{s:{s:{s:s}}}}",
-                                 "attributes", "system", "exec",
-                                 "job_shell", &path);
-    return path;
-}
-
-static const char *job_shell_path (struct jobinfo *job)
-{
-    const char *path = jobspec_get_job_shell (job->jobspec);
-    return path ? path : default_job_shell;
-}
-
-static const char *jobspec_get_cwd (json_t *jobspec)
-{
-    const char *cwd = NULL;
-    (void) json_unpack (jobspec, "{s:{s:{s:s}}}",
-                                 "attributes", "system",
-                                 "cwd", &cwd);
-    return cwd;
-}
-
-static const char *job_get_cwd (struct jobinfo *job)
-{
-    const char *cwd;
-    if (job->multiuser)
-        cwd = "/";
-    else if (!(cwd = jobspec_get_cwd (job->jobspec)))
-        cwd = default_cwd;
-    return (cwd);
 }
 
 static void start_cb (struct bulk_exec *exec, void *arg)
@@ -300,7 +264,7 @@ static int exec_init (struct jobinfo *job)
     struct bulk_exec *exec = NULL;
     const struct idset *ranks = NULL;
 
-    if (job->multiuser && !flux_imp_path) {
+    if (job->multiuser && !config_get_imp_path ()) {
         flux_log (job->h,
                   LOG_ERR,
                   "unable run multiuser job with no IMP configured!");
@@ -333,18 +297,18 @@ static int exec_init (struct jobinfo *job)
         goto err;
     }
     if (job->multiuser) {
-        if (flux_cmd_argv_append (cmd, flux_imp_path) < 0
+        if (flux_cmd_argv_append (cmd, config_get_imp_path ()) < 0
             || flux_cmd_argv_append (cmd, "exec") < 0) {
             flux_log_error (job->h, "exec_init: flux_cmd_argv_append");
             goto err;
         }
     }
-    if (flux_cmd_argv_append (cmd, job_shell_path (job)) < 0
+    if (flux_cmd_argv_append (cmd, config_get_job_shell (job)) < 0
         || flux_cmd_argv_appendf (cmd, "%ju", (uintmax_t) job->id) < 0) {
         flux_log_error (job->h, "exec_init: flux_cmd_argv_append");
         goto err;
     }
-    if (flux_cmd_setcwd (cmd, job_get_cwd (job)) < 0) {
+    if (flux_cmd_setcwd (cmd, config_get_cwd (job)) < 0) {
         flux_log_error (job->h, "exec_init: flux_cmd_setcwd");
         goto err;
     }
@@ -431,7 +395,7 @@ static int exec_kill (struct jobinfo *job, int signum)
     flux_future_t *f;
 
     if (job->multiuser)
-        f = bulk_exec_imp_kill (exec, flux_imp_path, signum);
+        f = bulk_exec_imp_kill (exec, config_get_imp_path (), signum);
     else
         f = bulk_exec_kill (exec, signum);
     if (!f) {
@@ -467,54 +431,9 @@ static void exec_exit (struct jobinfo *job)
     job->data = NULL;
 }
 
-/*  Configure the exec module.
- *  Read the default job shell path from config. Allow override on cmdline
- */
 static int exec_config (flux_t *h, int argc, char **argv)
 {
-    flux_error_t err;
-
-    /*  Set default job shell path from builtin configuration,
-     *   allow override via configuration, then cmdline.
-     */
-    default_job_shell = flux_conf_builtin_get ("shell_path", FLUX_CONF_AUTO);
-
-
-    /*  Check configuration for exec.job-shell */
-    if (flux_conf_unpack (flux_get_conf (h),
-                          &err,
-                          "{s?:{s?s}}",
-                          "exec",
-                            "job-shell", &default_job_shell) < 0) {
-        flux_log (h, LOG_ERR,
-                  "error reading config value exec.job-shell: %s",
-                  err.text);
-        return -1;
-    }
-
-    /*  Check configuration for exec.imp */
-    if (flux_conf_unpack (flux_get_conf (h),
-                          &err,
-                          "{s?:{s?s}}",
-                          "exec",
-                            "imp", &flux_imp_path) < 0) {
-        flux_log (h, LOG_ERR,
-                  "error reading config value exec.imp: %s",
-                  err.text);
-        return -1;
-    }
-
-    /* Finally, override values on cmdline */
-    for (int i = 0; i < argc; i++) {
-        if (strncmp (argv[i], "job-shell=", 10) == 0)
-            default_job_shell = argv[i]+10;
-        else if (strncmp (argv[i], "imp=", 4) == 0)
-            flux_imp_path = argv[i]+4;
-    }
-    flux_log (h, LOG_DEBUG, "using default shell path %s", default_job_shell);
-    if (flux_imp_path)
-        flux_log (h, LOG_DEBUG, "using imp path %s", flux_imp_path);
-    return 0;
+    return config_init (h, argc, argv);
 }
 
 struct exec_implementation bulkexec = {
--- a/src/modules/job-exec/exec_config.c
+++ b/src/modules/job-exec/exec_config.c
@@ -1,0 +1,123 @@
+/************************************************************\
+ * Copyright 2021 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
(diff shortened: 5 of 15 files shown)


chu11:
re-pushed

- updated `sdprocess` API class for what was eventually commited in #3864 (almost exclusively search & replace, nothing notable here)

- support a new `attributes.system.exec.sd.test` attribute to force `sdexec` to be used under some test scenarios, such as ...

- add `sdexec` valgrind workload, had to do some tricks here to get things to work, see the workload comments

- added some more sharness tests

- defaulted stdout/stderr fds of systemd launched processes to `-1`.  We could debate if this should be configurable or not, but under most configurations this means stdout/stderr will goto the systemd journal.  It does appear there is some raciness / bugs in systemd that stdout/stderr can be lost if there is an error an the process exits quickly.  It seems best to let them goto the systemd journal, seems best / highest odds of things getting logged somewhere.

chu11:
hmmm focal py3.8 and el8 builders are timing out on my `sdexec` valgrind tests.

```
  Running job-sdexec
  flux timed out after 500.0s
```

I increased the timeout from 300 to 500s assuming it was legitimately timing out, but seems like this timeout is due to a hang or something.  I'll add some debugging, see if we can figure out something.

grondo:
Sometimes you can trigger those intermittent hangs by forcing the valgrind test onto a single core locally, e.g.
```
$ taskset -c 0 ./t5000-valgrind.t -d -v
```

chu11:
re-pushed, @grondo surmised that the issue was `job-sdexec` should not be run in all builders, just the `system` one, which is a really good guess and probably right :-)

chu11:
Ahh, figured out why in the below case `sdexec` was not cleaning up the job.  Needed to implement the `job-exec` `cancel` callback.

```
  Running job-sdexec
  1644382844.242704 submit userid=1001 urgency=16 flags=0
  1644382844.413431 depend
  1644382844.415442 priority priority=16
  1644382844.702991 alloc annotations={"sched":{"resource_summary":"rank0/core1"}}
  1644382844.933744 exception type="exec" severity=0 userid=-1 note="sdprocess_exec: No such file or directory"
```

chu11:
re-pushed, since the last time I squash

- add `sdexec_cancel()`, to handle corner case when `sdprocess_exec()` fails.
- re-add stdout/stderr logging to `exec.eventlog`, make logging configurable and fix bug in which logging didn't work
- did some refactoring
- add tests for much fo the above
- added basic system / multiuser tests for `sdexec`

Note that the system / multiuser tests use `system.exec.sd.test=true` attribute to make sure tests are run under sdexec / systemd.  This can be altered later if we make `sdexec` the default.  Unsure if we'd want to update the default `exec.toml` or create one in the fluxorama image that makes `sdexec` the default.   Thoughts?

grondo:
>  Unsure if we'd want to update the default exec.toml or create one in the fluxorama image that makes sdexec the default. Thoughts?

Good question. I think as we move forward with the system instance testing it will make sense at some point to switch the `system` test in ci to use systemd by default. This would probably best be done via the `docker-run-systest.sh` script at first, though eventually it will make sense to always have the system instance use sdexec.

grondo:
Just coming back around to check on this.

A little while ago, I pushed 4 commits to my copy of your `job_exec_sdexec` branch  which add support for a job shell `--reconnect` option, uses this option for sdexec launched jobs, and includes a fix for the sdexec implementation trying to re-send J to reattached jobs.

If you'd like I can push these directly on top of this branch.

Then, I think before we merge we should add a real test that runs in the "system" instance testsuite, and ensure that a test system instance broker can be restarted without losing running sleep jobs.

grondo:
BTW, now that we have `flux config get` support, I wonder if we should update rc1 to check for `use-systemd = true` instead of an environment variable to avoid registering the cleanup scriptlets that kill all jobs?

chu11:
re-pushed

- note that I put all of the minor cleanups into a different PR #4197 since the commit count was getting a tad high
- added @grondo's commits from his branch (note @grondo, i found a code formatting issue in one of your commits, I just went ahead and fixed it).
- added a fix for a systemd race condition (I handled under the single user case, not the multiuser case)
- "reduced" raciness possibility, but AFAICT cannot eliminate it.  Added fixes into tests to resolve.
  - FYI, the race is the fact that to launch job, we call systemd, to launch imp, which launches job-shell, which runs user task.  if you send a signal shortly after launching job, signal may not actually be sent to the user task, it could be sent do something "higher up" in the chain.  There's nothing wrong with this generally speaking, but expected exit values in tests can be racy.
- updated `etc/rc1` to not cancel jobs when we're running under the system instance w/ systemd
- added system instance tests, including basic ones and system instance restart ones

chu11:
re-pushed

- new commit, noticed that a call in `libsdprocess` probably should be synchronous and not asynchronous given how it is used.  Did some benchmark tests and job throughput did drop maybe 5%, but given the potential for job state changes to be completely missed, I felt fix was necessary.
- fixed a race condition in which some events might not be logged to the job eventlog.  Basically was a race condition based on whether the flux reactor would call the "I have output" or "job has exited" routine first.  Resolved by setting stdout/stderr socketpairs to NON_BLOCKING and doing a final dump of stdout/stderr before job completion.

grondo:
> new commit, noticed that a call in libsdprocess probably should be synchronous and not asynchronous given how it is used. 

There is no chance this call could block?

chu11:
> There is no chance this call could block?

Hmmm, good point.  I suppose it is always possible it could block.

In this particular case, there's a decent amount of synchronous stuff in this path b/c its basically setting up for the systemd exec before returning to the user.  If we want to go down the asynchronous path to have higher job throughput, I think we'd need to think of a redesign or try to figure out some "bulk submit" kinda thing to do.

chu11:
per an offline discussion with @grondona, the configuration to enable `sdexec` is now under 

```
[systemd]
enable = true
```

this is because we expect additional systemd configurations to come in time.

As a result of this change a new `flux-config-systemd` manpage is now available.

grondo:
Actually I was suggesting `exec.systemd` as I think a top level `[systemd]` table will be confusing. Then the man page will be `flux-config-exec-systemd(5)`. This makes it clear that we're talking about systemd as used by the exec system.

chu11:
Somehow I just didn't like `[exec.systemd]`.  Thinking about it, I think it was due to the suggestion of:

```
[exec.systemd]
enable = true
```

so a user would enable the "default" execution mechanism outside of the main `[exec]` table.   So the main "default" setting is probably in a different file and a different manpage than `flux-config-exec(5)`.  Of course we can put it all in one manpage but as we add cgroups stuff, I felt it would have to go into a new manpage.  So somehow in my mind I can to using `[systemd]` as better.

now that I think about it a bit more, I'll suggest the following.

```
[exec]
default_exec = {subprocess,systemd,(maybe "test")}

[exec.systemd]
future cgroups stuff
```

now it sort of makes more sense to me.  you pick the default executor (and in the future there could be non-systemd ones).  so now the sub-table and manpage for `[exec.systemd]` make more sense.  its extra configs for the executor that's used.

Looking at the above, the default executors could possibly be renamed?  instead of `sdexec` it should be `systemd`?  All the names we can debate a bit.  Dunno if changing `job-exec`'s config to something like this should be a PR before this one.

grondo:
> so a user would enable the "default" execution mechanism outside of the main [exec] table.

TOML is hierarchical, so this is actually within the `[exec]` table.

> default_exec = {subprocess,systemd,(maybe "test")}

I don't really understand having an array of "default" execution implementations. There can currently only be one default, and you would never want to choose the `test` implementation as the default. However, this would work for me:

```toml
[exec]
method = "systemd" # or, default-method if you prefer

[exec.systemd]

```

Which I think follows the spirit of your suggestion

grondo:
You did convince me that we don't want `exec.systemd.enabled = true`. You are right that is awkward, and if another execution method was added in the future, we don't want to have a way for them both to be "enabled" a the same time. So I fully understand and agree with your suggestion for a top-level `[exec]` key to configure the "default" execution method.

chu11:
> I don't really understand having an array of "default" execution implementations. There can currently only be one default, and you would never want to choose the test implementation as the default. However, this would work for me:

Oh sorry, those were "options" for the user to pick not an array :-)  

```
[exec]
method = "systemd" # or, default-method if you prefer
```

so yeah, this is exactly what i'm thinking of.  Let me add this via a different PR, since I think it's outside of this PR.

chu11:
> you would never want to choose the test implementation as the default

wasn't sure if some weird test scenarios might want it.  Maybe?  But we can obviously choose not to do that.

grondo:
Ah, ok, sorry! Sounds good to me then.

chu11:
Talked to @grondo offline and a full implementation of

```
[exec]
method = ...
```

would be difficult to do given how job-exec is implemented internally right now.  So I've documented it in #3346 for later.

For this PR, we will just support ```method = "systemd"``` and nothing else.

Re-pushed the PR to support just this.

chu11:
re-pushed, removed some errant copy & paste in the sdexec tests

chu11:
re-pushed, just rebasing on top of #4252

grondo:
I was running a sort of (unfair) stress test on this branch, essentially something like:
```
$ flux mini submit --cc=1-100 sleep 1
```
Then doing `sudo systemctl restart flux` at random intervals. I'm not sure if we should fully expect this to work, but after doing this test for a little while, I had several failed jobs:
```
grondo@fluxorama:/usr/src/src/modules/job-exec$ flux jobs -f failed
    ƒDGKGR8t grondo   sleep       F      1      1   5.247s fluxorama
    ƒDGNEPha grondo   sleep       F      1      1   5.230s fluxorama
    ƒDGNEPhZ grondo   sleep       F      1      1   5.227s fluxorama
    ƒDGKGR8s grondo   sleep       F      1      1   5.226s fluxorama
    ƒDGEpTHq grondo   sleep       F      1      1   7.851s fluxorama
    ƒDGHnRrX grondo   sleep       F      1      1   7.839s fluxorama
    ƒDG8tWAU grondo   sleep       F      1      1   7.899s fluxorama
    ƒDGANVSo grondo   sleep       F      1      1   7.830s fluxorama
    ƒDF7aztg grondo   sleep       F      1      1   4.715s fluxorama
    ƒDFgCj6F grondo   sleep       F      1      1   4.611s fluxorama
    ƒDFYnngX grondo   sleep       F      1      1   4.635s fluxorama
    ƒDF7aztf grondo   sleep       F      1      1   4.731s fluxorama
    ƒDE3KX4B grondo   sleep       F      1      1   6.663s fluxorama
    ƒDDnVeEj grondo   sleep       F      1      1   6.665s fluxorama
    ƒDDoydX5 grondo   sleep       F      1      1   6.656s fluxorama
    ƒDDuuaeT grondo   sleep       F      1      1   6.637s fluxorama
```

All of these hit the same job exception though:
```
$ for id in $(flux jobs -f failed -no {id}); do flux job attach $id 2>&1; done | grep job.exception
37.955s: job.exception type=exec severity=0 read: Bad file descriptor
37.946s: job.exception type=exec severity=0 read: Bad file descriptor
37.939s: job.exception type=exec severity=0 read: Bad file descriptor
37.931s: job.exception type=exec severity=0 read: Bad file descriptor
32.679s: job.exception type=exec severity=0 read: Bad file descriptor
32.671s: job.exception type=exec severity=0 read: Bad file descriptor
32.672s: job.exception type=exec severity=0 read: Bad file descriptor
32.654s: job.exception type=exec severity=0 read: Bad file descriptor
22.637s: job.exception type=exec severity=0 read: Bad file descriptor
22.604s: job.exception type=exec severity=0 read: Bad file descriptor
22.604s: job.exception type=exec severity=0 read: Bad file descriptor
22.607s: job.exception type=exec severity=0 read: Bad file descriptor
10.391s: job.exception type=exec severity=0 read: Bad file descriptor
10.393s: job.exception type=exec severity=0 read: Bad file descriptor
10.386s: job.exception type=exec severity=0 read: Bad file descriptor
10.366s: job.exception type=exec severity=0 read: Bad file descriptor
```

This appears to be from this line in `sdexec.c`:
```c
   if ((len = read (fd, buf, 1024)) < 0) {
        if (errno != EWOULDBLOCK)
            jobinfo_fatal_error (se->job, errno, "read");
    }
    else if (len > 0)
        jobinfo_log_output (se->job,
                            0,
                            basename (cmd),
                            stream,
                            buf,
                            len);
```

A couple points: It might be helpful if the job exception note included some other context in this case, since it is what appears in user output. However, I also wonder if a failure reading output from the job shell should be fatal here? Maybe just log an error and move on? Though in this case we get EBADF, so maybe not? I guess we need to get the bottom of why this is happening.

chu11:
>  I guess we need to get the bottom of why this is happening.

Ahh good catch.  Seems to be a corner case in here:

```
    else if (state == SDPROCESS_EXITED
             && !se->jobinfo_tasks_complete_called) {

        /* Since we are calling jobinfo_tasks_complete(), the
         * stdout/stderr fd watchers may be stopped.  So if there is
         * any lingering data on the stdout/stderr streams, they may
         * not be output.  This is possible because the reactor may
         * call this state callback before stdio callbacks below.
         */

        if (se->stdoutlog == SDEXEC_LOG_EVENTLOG)
            output_stream (se, se->stdout_fds[0], "stdout");
        if (se->stderrlog == SDEXEC_LOG_EVENTLOG)
            output_stream (se, se->stderr_fds[0], "stderr");

        jobinfo_tasks_complete (se->job,
                                resource_set_ranks (se->job->R),
                                sdprocess_wait_status (se->sdp));
        se->jobinfo_tasks_complete_called = true;
    }
```

so the process is exited, so the call to `output_stream()` is trying to read stdout/stderr from a process thats exited.  But if the other side of the socketpair is closed, should only lead to a EOF.  Will have to see if there's something racy going on that I"m currently not seeing.  I guess the fact it was a bunch of `sleep 1` instead of `sleep 100` is what is catching this.

chu11:
> I guess we need to get the bottom of why this is happening.

ahhh, I'm stupid.  When we "reattach" a job due a flux restart, the stdout/stderr fds will not have been initialized, thus the EBADFD.  On a "reattach" there is no way to get stdout/stderr (unless we let it all go to the systemd journal).

The reason this is racy (as eluded to in the comment in the above code snippet) is that this is a randomness factor about which callback will be called first.  Will the fd-watcher callback (i.e. stdout is available to read) or will it be the "state change in the process" callback.

I still see the following exception sometimes:

`284.395s: job.exception type=exec severity=0 rank 0 exited and exit-timeout=30s has expired`

I'm not 100% if this is avoidable?  I don't understand the doom code perfectly yet, but the assumption is the shell started, tried to do its "doom" initialization, broker restarted at the wrong time, and then shell job (sleep 1) completed before broker was brought back up?

grondo:
> I'm not 100% if this is avoidable? I don't understand the doom code perfectly yet, but the assumption is the shell started, tried to do its "doom" initialization, broker restarted at the wrong time, and then shell job (sleep 1) completed before broker was brought back up?

Yeah, the  `doom` plugin is completely in the job shell, so I don't think anything in the sdexec exec implementation could be changed to avoid this error. It does seem to have initialized correctly, since it is emitting the job exception. However, we should probably look into this one further since I don't really expect this to happen. 

My guess is that the task exits while the broker is not connected. Since the job shell should be blocked in the reconnect handler, I would expect that the shell doesn't notice the task exits until the reconnect is complete, then we'd expect the shell to just exit normally at that time. However, something is going wrong that is getting the shell stuck. It might be useful to run all the jobs with `-o verbose=2`, then dump the output when one of these errors  occur. Also the exec eventlog might be of interest as well.

chu11:
> My guess is that the task exits while the broker is not connected. Since the job shell should be blocked in the reconnect handler, I would expect that the shell doesn't notice the task exits until the reconnect is complete, then we'd expect the shell to just exit normally at that time. However, something is going wrong that is getting the shell stuck. It might be useful to run all the jobs with -o verbose=2, then dump the output when one of these errors occur. Also the exec eventlog might be of interest as well.

Hmmm, so I don't see a `shell.task-exit` event (side note, i restart flux every ten seconds in my test, thus the 3 restarts during the 30s doom timeout)

```
fluxorama|/usr/src 7>flux job eventlog -p guest.exec.eventlog ƒ3bWdp5rX
1650045657.927009 init
1650045657.934450 starting
1650045657.964564 shell.init leader-rank=0 size=1 service="8556-shell-5729754808320"
1650045657.968631 shell.start task-count=1
1650045665.545121 reattach
1650045665.548844 re-starting
1650045676.895165 reattach
1650045676.900576 re-starting
1650045688.130869 reattach
1650045688.137609 re-starting
1650045688.983061 complete status=256
1650045688.983114 done
```

I don't see any error messages that seem related.  Just` in case messages were lost going to exec.eventlog, tried logging to the systemd journal instead and still no error messages of note.

I'm guessing the missing `shell.task-exit` is simply b/c the broker is down when we try to log that event.  Would we want anything there to try and retry that eventlog commit?

And why the job doesn't recognize the task exit, don't know about that yet.  No specific clues (don't know the exit that well, trying to decipher)

grondo:
Well, I thought the shell would be blocked in the reconnect handler while it is disconnected, so it shouldn't be trying to post any events or anything during that time. However, perhaps the problem is that a message or messages is lost when the disconnect occurs?

chu11:
I added some debugging to dump the `completion_refs` hash when the doom timer expired (ignore the `1: doom_task_exit`, random debugging I stuck in there):

```
132.752s: job.exception type=exec severity=0 rank 0 exited and exit-timeout=30s has expired: 1: doom_task_exit: key output.txn=1,key eventlogger.txn=1,
```

It seems the two refs related to the eventloggers weren't unrefed.

I'm not 100% sure how the reconnect fully works.  Could flux future then callbacks not be called b/c of a reconnect?

grondo:
> I'm not 100% sure how the reconnect fully works. Could flux future then callbacks not be called b/c of a reconnect?

Yeah, if a request is sent to the local broker, but the shell is disconnected before there is a response, the reply will be dropped. This might require more thought, but perhaps for now there should be a set of completion references which are automatically removed after a  reconnect. At least this way the shell won't hang.

chu11:
> Yeah, if a request is sent to the local broker, but the shell is disconnected before there is a response, the reply will be dropped. This might require more thought, but perhaps for now there should be a set of completion references which are automatically removed after a reconnect. At least this way the shell won't hang.

Went a set of plugin callbacks be a good idea here?  e.g. perhaps a `reconnect_init()` and a `reconnect_cleanup()` set of callbacks?  Although there's no need for `reconnect_init()` at the moment.

Perhaps should be a follow up PR, since it's handling a very specific case in the shell that is perhaps not related to this PR.

chu11:
re-pushed, fixing up the EBADFD issue and a rebase issue due to #4278.  Will work on `doom` in shell issue in a follow up PR.

chu11:
re-pushed, i stupidly didn't clean up a test, which affected work in follow up PRs (or if I wanted to run the test again)

chu11:
rebase + repush, absorbing the initial implementation I put into #4293, setting WIP

grondo:
I've actually been testing this PR branch a bit lately and was thinking we might want to just go ahead and merge soon, then fold into the distributed execution module as needed. That way you would not have to keep rebasing at least.

What do you think?

chu11:
@grondo, sure, since the shell stuff is TBD, I assume you're saying I should axe it in this PR?

chu11:
re-pushed, given @grondo comment above plus the growth in complexity of #4293, I'm going to split it off and re-open #4293.  Removing WIP.

grondo:
Sorry I let this PR languish, @chu11!

I think it would be ok to merge this now, so you don't have to keep rebasing. We'll transition the sdexec implementation to the job-exec replacement when it is ready.

grondo (approved):

PR #4075, opened 2022-01-27
shell: enhance pty support

grondo:
This is something I had sitting around for a bit and thought I'd throw up here for now as a WIP for comments.

The job shell pty support is currently pretty limited. It only allows allocating a pty to rank 0, and only supports "interactive" use, where the pty is interactively attached immediately after start. All data sent from the task running under a pty is lost, meaning it is not logged anywhere.

The above is fine for the `flux mini alloc` use case, however there is currently no way to allocate a pty for tasks when run non-interactively or to allocate a pty to all tasks. This may be useful when running asynchronous jobs that expect a pty, e.g. a testsuite that wants to emit color output.

This PR modifies the job shell's `pty` option to be a JSON object with multiple, optional keys, including:

 * `"ranks"` (int or string): an integer rank or RFC 22 idset of ranks for which a pty should be allocated
 * `"capture"` (int): If nonzero, capture pty output and send it to the same location as `stdout`
 * `"interactive"` (int): If nonzero, emit the rank 0 pty service endpoint along with the `shell.init` event. This allows `flux job attach` or other entities to know that an interactive attach to the rank 0 pty is desired.

If the supplied `pty` option is _not_ an object, then a default of `{"ranks":"all","capture":1,"interactive":0}` is used. This means that the default behavior of `-o pty` is changed, it will now run all tasks in a job with a pty, and send output to the same location as `stdout` (the kvs by default). This seems like the right default, since attaching to jobs interactively is probably not the common case, except as used by `flux mini alloc`, in which case we can set up the pty for interactive use within the command.

If only `interactive` is set, then `ranks` defaults to `0` as does `capture`. Therefore to get the old behavior of allocating a pty on rank 0 for interactive use `-o pty.interactive` can be used. `flux mini alloc` is updated here to use this mechanism.

The main use case I had in mind here was running the flux-core testsuite in bulk with color output. That is, the following now works

```console
 $ flux mini bulksubmit --watch --progress -o pty -n1 ./{} ::: t*.t python/t*.py lua/t*.t
```

and the testsuite is run in parallel with color output. I figure this may be useful in similar scenarios.

This is a WIP since tests still would need to be written, if the approach seems acceptable.

The main issue is probably the change in default behavior with `-o pty`.  My assumption is that this option is really only used by `flux mini alloc` at this time, however people could be using it for something like:

```console
$ flux mini run -o pty -n8 flux start
```
and uses like this would have to be updated to use `-o pty.interactive`. However I feel like in the long run, the non-interactive case is likely to be most common so I wanted to make that the easiest to use.

Anyone have any opinions?

### DIFF
--- a/doc/man1/flux-shell.rst
+++ b/doc/man1/flux-shell.rst
@@ -215,7 +215,34 @@ see the documentation for the specific plugin in question. The following
 options are supported by the builtin plugins of ``flux-shell``:
 
 **pty**
-  Allocate a pty for the first task rank.
+  Allocate a pty to all task ranks for non-interactive use. Output
+  from all ranks will be captured to the same location as ``stdout``.
+  This is the same as setting **pty.ranks=all** and **pty.capture**.
+  (see below).
+
+**pty.ranks**\ =\ *OPT*
+  Set the task ranks for which to allocate a pty. *OPT* may be either
+  an RFC 22 IDset of target ranks, an integer rank, or the string "all"
+  to indicate all ranks.
+
+**pty.capture**
+  Enable capture of pty output to the same location as stdout. This is
+  the default unless **pty.interactive** is set.
+
+**pty.interactive**
+  Enable a a pty on rank 0 that is set up for interactive attach by
+  a front-end program (i.e. ``flux job attach``). With no other **pty**
+  options, only rank 0 will be assigned a pty and output will not
+  be captured. These defaults can be changed by setting other
+  **pty** options after **pty.interactive**, e.g.
+
+  .. code-block:: console
+
+    $  flux mini run -o pty.interactive -o pty.capture ...
+
+  would allocate an interactive pty on rank 0 and also capture the
+  pty session to the KVS (so it can be displayed after the job exits
+  with ``flux job attach``).
 
 **cpu-affinity**\ =\ *OPT*
   Adjust the operation of the builtin shell ``affinity`` plugin.
--- a/src/shell/pty.c
+++ b/src/shell/pty.c
@@ -19,12 +19,15 @@
 #include <fcntl.h>
 #include <sys/ioctl.h>
 #include <termios.h>
+#include <jansson.h>
 
 #include <flux/core.h>
 #include <flux/shell.h>
+#include <flux/idset.h>
 
 #include "src/common/libterminus/pty.h"
 #include "src/common/libterminus/terminus.h"
+#include "src/common/libccan/ccan/ptrint/ptrint.h"
 #include "builtins.h"
 #include "log.h"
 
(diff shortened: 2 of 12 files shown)


garlick:
Nice!  This seems like a good improvement

How would you feel about adding a top level make`check_in_flux` (or similar)  target for the bulksubmit line quoted above?

If someone uses `-o pty` when they really meant `-o pty.interactive`, it seems like the plumbing changes (rank >  0 output going to KVS etc) but they probably woudn't notice any difference?   Give that slurm `--pty` works like current `-o pty`, I could see that happening accidentally when people transition.  But if impact is low then it probably won't matter?

grondo:
> How would you feel about adding a top level makecheck_in_flux (or similar) target for the bulksubmit line quoted above?

Yeah, there is already `t/test-inception.sh` and the idea was to add `-o pty` to that.

> If someone uses -o pty when they really meant -o pty.interactive, it seems like the plumbing changes (rank > 0 output going to KVS etc) but they probably woudn't notice any difference? Give that slurm --pty works like current -o pty, I could see that happening accidentally when people transition. But if impact is low then it probably won't matter?

This is actually my main worry. There will actually be a big difference because without `-o pty.interactive` the current terminal (on the client) is not put into raw mode, *and* stdin is not sent to the pty on any rank by default. (This could be improved later).

So the user experience would indeed change, unfortunately, in that the user will be unable to interact with the remote pty(s) in the `-o pty` case. I assumed this is probably not a common case outside of `flux mini alloc`, but I could be wrong. I'm open to other ideas.

grondo:
> in that the user will be unable to interact with the remote pty(s) in the -o pty case.

If this case becomes a problem, I wonder if a `--pty` option would be warranted for `flux mini run` only.
This could not only set `-o pty.interactive` on behalf of the user, but could invoke the underlying `flux job attach` with a new option that would force a pty attach to the remote job, putting local stdio into raw mode, etc.

garlick:
> If this case becomes a problem, I wonder if a --pty option would be warranted for flux mini run only.

Could always wait and see - easy to add later.

grondo:
I've added some tests here (we'll see where coverage ends up) and removed WIP.

src/shell/pty.c
```diff
@@ -62,6 +65,163 @@ shell_terminus_server_start (flux_shell_t *shell, const char *shell_service)
     return t;
 }
 
+static void pty_monitor (struct flux_pty *pty, void *data, int len)
+{
+    flux_plugin_arg_t *args = flux_plugin_arg_create ();
```
garlick, line 70: check for NULL return from `flux_plugin_arg_create()`

Edit: oh, I guess `flux_plugin_arg_pack()` will fail with EINVAL in that case so maybe OK.

doc/man1/flux-shell.rst
```diff
@@ -215,7 +215,37 @@ see the documentation for the specific plugin in question. The following
 options are supported by the builtin plugins of ``flux-shell``:
 
 **pty**
```
garlick, line 217: Would it improve clarity if there were separate entries for:
- pty
- pty.interactive
- pty.ranks=OPT
- pty.capture

I'm not sure casual users will be thinking of how the shell options map to json objects in the jobspec, so presenting them as distinct options might make them more accessible?

garlick (approved):
Great job on the test coverage.

I poked at this and found everything to work as described.  Just a couple of comments that aren't really significant so feel free to not address if you don't agree.

doc/man1/flux-shell.rst
grondo: Ah, that is a good idea! I will make that change.

grondo:
I've been seeing this issue intermittently in the centos7 builder:

```
 expecting success: 
  	flux mini run -n2 --label-io -o pty.ranks=0,1 tty > ptyss.out &&
  	grep 0: ptyss.out &&
  	grep 1: ptyss.out &&
  	test_must_fail grep "not a tty" ptyss.out
  
  0.056s: flux-shell[0]: stderr: flux-shell: ERROR: pty: monitor: failed to send 12 bytes of shell.output: Cannot allocate memory
  0.056s: flux-shell[0]: stderr: flux-shell: ERROR: pty: monitor: failed to send 12 bytes of shell.output: Cannot allocate memory
  0.056s: flux-shell[0]: ERROR: pty: monitor: failed to send 12 bytes of shell.output: Cannot allocate memory
  0.056s: flux-shell[0]: ERROR: pty: monitor: failed to send 12 bytes of shell.output: Cannot allocate memory
  not ok 7 - pty: pty.ranks can take an idset
```

Not sure why it only occurs in that builder.. I'll have to investigate.

grondo:
Found a possible integer type issue. Force-pushed a potential fix.

grondo:
@garlick helped me find the _actual_ problem. Fix pushed.

grondo:
I tacked on a small patch to add `-o pty` to `t/test-inception.sh` so we get color output in CI, and set MWP.

grondo:
Thanks!

Issue #4033, opened 2022-01-07
idea: allow users to query "uptime" of enclosing instance


grondo:
It is a small thing, but it might be of interest to be able to query the uptime of the current enclosing instance.

This came up in development of `flux pstree`, where we try to approximate a `JobInfo` for the current enclosing instance, even if that instance isn't technically a Flux job (otherwise we can query the `job-list` module of the parent)

One simple approach would be to add a `broker.starttime` or similar attribute, though that would track the last broker restart and not necessarily the uptime of something like a system instance, which may persist across many restarts (though perhaps the broker starttime is of interest in and of itself)

Another idea would be to fill in the starttime member of `R` whenever it is not already set when the resource module initializes from configuration or by gathering R from hwloc. Then, I think, users could get at this information by querying the resource list from the scheduler.

There's probably many other acceptable ways to do this as well.

garlick:
Good idea.  Maybe we should have a simplistic `broker.starttime` attribute that just tracks an individual broker start time (with rank 0's acting as a kind of stand in for "instance (re-)start time"), AND fill in the starttime member of `resource.R` in the KVS so that we have an instance incept time that persists across a system instance restart?

(Actually I think `resource.R` gets overwritten on each rank 0 restart in the system instance, so we might need to add something to preserve an existing starttime in that case)

PR #4076, opened 2022-01-28
add broker.starttime; add uptime to flux-top, flux-pstree

garlick:
This adds a `broker.starttime` attribute as discussed in #4033, and a convenience function for fetching it from rank 0 so it can be used to calculate instance uptime.  Then the uptime is displayed in `flux-top`.

There are a couple of small cleanups tacked on also.

### DIFF
--- a/src/cmd/top/summary_pane.c
+++ b/src/cmd/top/summary_pane.c
@@ -15,6 +15,7 @@
 #include <unistd.h>
 #include <time.h>
 #include <jansson.h>
+#include <math.h>
 
 #include "src/common/libutil/fsd.h"
 #include "src/common/librlist/rlist.h"
@@ -164,14 +165,14 @@ static void draw_bargraph (WINDOW *win, int y, int x, int x_length,
                suffix);
     /* Graph used */
     wattron (win, COLOR_PAIR (TOP_COLOR_YELLOW));
-    for (int i = 0; i < ((double)res.used / res.total) * slots; i++)
+    for (int i = 0; i < ceil (((double)res.used / res.total) * slots); i++)
         mvwaddch (win, y, x + strlen (prefix) + i, '|');
     wattroff (win, COLOR_PAIR (TOP_COLOR_YELLOW));
 
     /* Graph down */
     wattron (win, COLOR_PAIR (TOP_COLOR_RED));
     for (int i = slots - 1;
-         i >= slots - ((double)res.down / res.total) * slots; i--) {
+         i >= slots - ceil (((double)res.down / res.total) * slots); i--) {
         mvwaddch (win, y, x + strlen (prefix) + i, '|');
     }
     wattroff (win, COLOR_PAIR (TOP_COLOR_RED));
--- a/src/cmd/top/summary_pane.c
+++ b/src/cmd/top/summary_pane.c
@@ -53,6 +53,7 @@ struct summary_pane {
     WINDOW *win;
     int instance_level;
     int instance_size;
+    double starttime;
     const char *instance_version;
     double expiration;
     struct stats stats;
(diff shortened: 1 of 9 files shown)


src/cmd/top/summary_pane.c
```diff
@@ -417,6 +429,8 @@ struct summary_pane *summary_pane_create (struct top *top)
     sum->instance_level = get_instance_attr_int (top->h, "instance-level");
     sum->instance_size = get_instance_attr_int (top->h, "size");
     sum->instance_version = flux_attr_get (top->h, "version");
+    if (flux_get_instance_starttime (top->h, &sum->starttime) < 0)
+        fatal (errno, "error fetching broker.starttime attribute");
```
grondo, line 433: Should `flux top` tolerate a missing broker.starttime?
```
$ src/cmd/flux top
flux-top: error fetching broker.starttime attribute: No such file or directory
```

garlick: OK, that's easy enough.

garlick: Pushed a fix for that - it will just use the start time of `top` if the instance starttime is unavailable.  That OK?

grondo: That seems fine. It would be an increasingly rare corner case anyway.

grondo (approved):
Nice!  I like it.

I actually wonder if a `N days, HH:MM:SS` would look better for the uptime display in top, but current approach is absolutely fine.

This will also allow us to display a RUNTIME for enclosing instance that is not a job in `flux pstree` if you want to also include this:

```diff
diff --git a/src/cmd/flux-pstree.py b/src/cmd/flux-pstree.py
index 950dca23a..58284b599 100755
--- a/src/cmd/flux-pstree.py
+++ b/src/cmd/flux-pstree.py
@@ -426,6 +426,10 @@ def get_root_jobinfo():
             nodelist=nodelist,
             annotations={"user": {"uri": uri}},
         )
+        try:
+            info['t_run'] = float(handle.attr_get("broker.starttime"))
+        except OSError:
+            pass
 
     #  If 'ranks' idset came from parent, it could be confusing,
     #   rewrite ranks to be relative to current instance, i.e.
```

Or I can make a follow on PR or push a commit directly to this branch.

garlick:
Stepping out for a bit so feel free to push to this branch!

grondo:
Ok, I pushed a tiny change to flux-pstree, which now uses `broker.starttime` to set a fake `t_run` for an enclosing instance that is not also a Flux job.

src/common/libflux/attr.h
```diff
@@ -66,6 +66,13 @@ int flux_get_size (flux_t *h, uint32_t *size);
  */
 const char *flux_get_hostbyrank (flux_t *h, uint32_t rank);
 
+/* Look up the broker.starttime attribute on rank 0.
+ * Subtract from flux_reactor_now() get Flux instance uptime.
+ * N.B. if the instance has been restarted, this value is the most
+ * recent restart time.
+ */
```
chu11, line 73: this comment doesn't seem to match the code, we're not subtracting from `flux_reactor_now()`?

chu11 (commented):
also (just to be sure), we don't need to document `broker.starttime`?

src/common/libflux/attr.h
chu11: oh wait, I think you mean, "Subtract the starttime from flux_reactor_now() to get the Flux instance uptime", not that the function does this.  Maybe minor reformat / wording, or maybe it's just me being stupid.

garlick: Oh yeah I see now that the wording is a bit confusing.  I'll fix.

garlick:
Fixed inline comment and added a short bit to `flux-broker-attributes(7)`.

PR #4078, opened 2022-01-30
flux-mini: add --wait-event option to submit/bulksubmit

grondo:
Had this one sitting around and there's been a few times it would have be useful for testing, so I thought I'd go ahead and propose a PR.

This PR adds a `--wait-event=NAME` option to `flux mini submit` and `flux mini bulksubmit`. The current `--wait` option thus becomes equivalent to `--wait-event=clean`. This option reduces the need for constructs like 

```shell
    id=$(flux mini submit ...)
    flux job wait-event ${id} start
```
which can be replaced with simply
```
   id=$(flux mini submit --wait-event=start ...)
```
and allows waiting for _all_ submitted jobs to receive the indicated event when used with `bulksubmit` or `flux mini submit --cc`.

(It does not, however, provide a timeout option like `flux job wait-event`, so if a timeout is desired `wait-event` should be used.)

The `--wait-event` option optionally allows events from the exec eventlog when the event starts with `exec.`, e.g. `flux mini submit --wait-event=exec.shell.init`.

There's some other cleanup in this PR, namely adding a `stop=True` parameter to `future.cancel()` which allows further callbacks for a streaming RPC to be stopped immediately from Python (as if `flux_future_destroy()` had been called in C), a fix for the `--log` and `--log-stderr` options, and a small fix for `progressbar`.

### DIFF
--- a/src/bindings/python/flux/progress.py
+++ b/src/bindings/python/flux/progress.py
@@ -294,6 +294,7 @@ class ProgressBar(Bottombar):
         length = width - len(before) - len(after) - 2
 
         #  Calculate amount of bar fully filled and buil string
+        fraction = min(fraction, 1.0)
         fill = int(length * fraction)
         filled = style[-1] * fill
 
--- a/src/bindings/python/flux/progress.py
+++ b/src/bindings/python/flux/progress.py
@@ -293,7 +293,7 @@ class ProgressBar(Bottombar):
         #  Calculate remaining length for progress bar:
         length = width - len(before) - len(after) - 2
 
-        #  Calculate amount of bar fully filled and buil string
+        #  Calculate amount of bar fully filled and build string
         fraction = min(fraction, 1.0)
         fill = int(length * fraction)
         filled = style[-1] * fill
--- a/src/cmd/flux-mini.py
+++ b/src/cmd/flux-mini.py
@@ -760,6 +760,7 @@ class SubmitBulkCmd(SubmitBaseCmd):
 
         #  dictionary of open logfiles for --log, --log-stderr:
         self._logfiles = {}
+        self.t0 = None
 
         super().__init__()
         self.parser.add_argument(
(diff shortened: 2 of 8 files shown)


src/bindings/python/flux/progress.py
```diff
@@ -294,6 +294,7 @@ def _formatter(self, bbar, width):
         length = width - len(before) - len(after) - 2
 
         #  Calculate amount of bar fully filled and buil string
```
chu11, line 296: not related to this PR but noticed typo here, "buil" -> "build"?

src/cmd/flux-mini.py
```diff
@@ -903,13 +917,14 @@ def event_watch_cb(self, future, args, jobinfo, label=""):
             )
         elif event.name == "alloc":
             jobinfo["state"] = "running"
-        elif event.name == "start" and args.watch:
+        elif event.name == "start" and (args.watch or args.wait.startswith("exec.")):
             #
             #  Watch the exec eventlog if the --watch option was provided:
```
chu11, line 922: minor comment typo: probably should colon end of line since the sentence continues now

chu11 (commented):
just noticed a few nits, technically none of them are from this PR :-)

grondo:
Addressed @chu11's comments and forced a push

chu11 (approved):

grondo:
Thanks @chu11, setting MWP.

chu11:
Perhaps we should create an issue to cleanup a lot of tests that could now use this option.  I imagine there's a nice chunk of cleanup that could be done.

garlick:
Looks like the shell pty test got stuck here in the coverage run.  I'll restart.
```
   expecting success: 
  	id=$(flux mini submit -o pty.interactive -o pty.capture tty) &&
  	flux job attach $id >ptyim.out 2>&1 &&
  	flux job attach $id
```

grondo:
> Looks like the shell pty test got stuck here in the coverage run. I'll restart.

Hm, I saw that one hang another time too and thought it was something else. I'll open an issue and look into it.

PR #4080, opened 2022-01-31
ci: rename centos images to el, and use rockylinux for el8 image

grondo:
It appears centos8 repos have shutdown or moved, and `yum update` is failing in repos where it is required in CI where extra packages are necessary.

This PR renames the "centos" images as a more generic "el" and updates all references throughout the code, including mergify config.

Then, the `el8` base image is updated from `centos:8` to `rockylinux:8`.

After this is merged and new `flux-core:el*` images are built, all dependent flux-framework projects' CI will also need to be updated.


src/test/generate-matrix.py
```diff
@@ -196,23 +196,23 @@ def __str__(self):
...
     docker_tag=True,
 )
 
 # Centos8, system, coverage
 matrix.add_build(
-    name="centos8 - system,coverage",
-    image="centos8",
+    name="el8 - system,coverage",
+    image="el8",
     coverage=True,
```
chu11, line 216: do some of these "Centos" comments need to be updated as well?  May be good to do a grep for it and double check.

chu11:
just one high level comment.  While irritating to search/replace "centos8" to "rocky8", what are the odds that this will change / issue will happen again?  Skimming code and seeing `el8` is not the clearest thing to me, `rocky8` is clearer.

grondo:
The reason to use `el8` is that this generically means "RHEL8 clone". If we have to update away from rockylinux8 in the future then this global search/replace won't be necessary, instead the base image could just be updated from `rockylinux:8` to `almalinux:8` or whatever. The specific naming of everything bit us this time, I didn't want to suffer that again.

src/test/generate-matrix.py
grondo: Yeah, good catch. I had only searched (case sensitively) for `centos`. I'll fix these. Thanks!

grondo:
I force-pushed an update to rename most other references to CentOS in comments etc. (Thanks @chu11!).

I also added the reasoning behind the `el7`,`el8` naming in the commit that makes the change.

chu11:
> The specific naming of everything bit us this time, I didn't want to suffer that again.

Understood, just wondering if you thought odds low/high to justify the change vs. just a search / replace.

chu11 (approved):

grondo:
> Understood, just wondering if you thought odds low/high to justify the change vs. just a search / replace.

Heh, probably even odds. All the centos replacements are relatively new at this time so I'm not sure what will happen. The real pain will be the next step where every project that uses the `flux-framework/flux-core:centos8` image has to _also_ be updated...

grondo:
Ah, this will need to be manually merged since the mergify config has been modified as part of the rename.

chu11:
hitting the big button!

grondo:
Thanks!

PR #4085, opened 2022-02-02
shell: fix duplicate logging after evlog plugin is initialized

grondo:
Error messages from the job shell are currently duplicated to the `evlog` plugin and `stderr`, and since job-exec copies `stderr` to the output eventlog, these errors are also duplicated in the output of `flux job attach`, which makes for a bit more output than is strictly necessary, e.g.:

```
ƒ(s=1,d=0) grondo@fluke108:~$ flux mini run badprog
0.060s: flux-shell[0]: FATAL: task 0: start failed: badprog: No such file or directory
0.060s: flux-shell[0]: stderr: flux-shell: FATAL: task 0: start failed: badprog: No such file or directory
0.061s: job.exception type=exec severity=0 task 0: start failed: badprog: No such file or directory
```

This PR adds a way to suppress all `stderr` log messages once the `evlog` plugin is initialized. It is unlikely we'll lose important errors, since the `evlog` plugin copies log messages that fail to get committed to the eventlog to `stderr` anyway.

### DIFF
--- a/src/shell/evlog.c
+++ b/src/shell/evlog.c
@@ -79,6 +79,9 @@ static int log_eventlog (flux_plugin_t *p,
 
 static void evlog_destroy (struct evlog *evlog)
 {
+    /*  Redirect future logging to stderr */
+    flux_shell_log_setlevel (evlog->level, "stderr");
+
     eventlogger_flush (evlog->ev);
     eventlogger_destroy (evlog->ev);
     free (evlog);
--- a/src/shell/evlog.c
+++ b/src/shell/evlog.c
@@ -206,7 +206,8 @@ static int log_eventlog_start (flux_plugin_t *p,
                                    evlog) < 0)
         goto err;
 
-    flux_shell_log_setlevel (FLUX_SHELL_ERROR, "stderr");
+    /*  Disable stderr logging */
+    flux_shell_log_setlevel (FLUX_SHELL_QUIET, "stderr");
     return 0;
 err:
     evlog_destroy (evlog);
--- a/src/shell/evlog.c
+++ b/src/shell/evlog.c
@@ -79,6 +79,7 @@ static int log_eventlog (flux_plugin_t *p,
 
 static void evlog_destroy (struct evlog *evlog)
 {
+    eventlogger_flush (evlog->ev);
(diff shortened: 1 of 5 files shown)


src/shell/log.c
```diff
@@ -337,7 +337,7 @@ static int log_setlevel (flux_shell_t *shell, const char *dest, int level)
 
 int flux_shell_log_setlevel (int level, const char *dest)
 {
-    if (level < FLUX_SHELL_FATAL || level > FLUX_SHELL_TRACE) {
+    if (level < -1 || level > FLUX_SHELL_TRACE) {
```
garlick, line 340: Might be more clear if something like `FLUX_SHELL_DISABLE = -1` were added to the enum.

garlick (approved):
Looks like a good change to me.  Just one minor suggestion.

src/shell/log.c
grondo: Good point, but can you use a negative value with an enum?

garlick: It would appear so :-)

grondo: Ok, I'll add that.

grondo:
Ok, I addressed @garlick's wise comment (I chose `FLUX_SHELL_QUIET` though) and force pushed the result.

I also realized that the `evlog` plugin should enable `stderr` logging again as it is exiting  (though it is unlikely there will be any log messages at this point) so tacked that fix on as well.

grondo:
Oh, I also need to update the `flux_shell_log_setlevel(3)` manpage, one moment.

src/shell/evlog.c
```diff
@@ -79,6 +79,7 @@ static int log_eventlog (flux_plugin_t *p,
 
 static void evlog_destroy (struct evlog *evlog)
 {
+    eventlogger_flush (evlog->ev);
```
garlick, line 85: Looking good!

I did just notice a typo in 16d45797c47d403998d6b18c656000631dc00d90

s/evenlogger/eventlogger/

grondo:
I've updated the manpage, fixed the typo in commit message found by @garlick, and also added a small missing test to the shell logging tests.

grondo:
Setting MWP

PR #4094, opened 2022-02-03
flux resource: allow limited guest access

garlick:
This PR ensures that `flux resource drain` (no args) can be used on any rank to list drained nodes, not just rank 0.

Also, allow guests to access the `resource.status` RPC, thus allowing them to use the above command plus `flux resource status`.  This eliminates an unnecessary sudo for sys admins, and also may be useful to users debugging a problem with their jobs.  There is no reason to restrict access that I can think of.

Update tests and man page.

### DIFF
--- a/doc/man1/flux-resource.rst
+++ b/doc/man1/flux-resource.rst
@@ -58,13 +58,12 @@ COMMANDS
    *-o,--format=FORMAT*, customizes output formatting (see below).
    *-s,--states=STATE,...* limits output to specified resource states, where
    valid states are "online", "offline", "avail", "exclude", "drain", and "all".
-   This command is restricted to the Flux instance owner.
 
 **drain** [targets] [reason ...]
    If specified without arguments, list drained nodes.  The *targets* argument
    is an IDSET or HOSTLIST specifying nodes to drain.  Any remaining arguments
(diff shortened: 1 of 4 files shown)


grondo (approved):
Great!

garlick:
Thanks!  bionic failed here, which I don't think could be related so I'll restart:
```
 rexec ps works
  
  expecting success: 
          ${FLUX_BUILD_DIR}/t/rexec/rexec -r 1 sleep 100 &
          pid1=$!
          ${FLUX_BUILD_DIR}/t/rexec/rexec -r 1 sleep 100 &
          pid2=$!
  	sleep 1 &&
          ${FLUX_BUILD_DIR}/t/rexec/rexec_ps -r 1 > output &&
          count=`cat output | wc -l` &&
          test "$count" = "2" &&
  	sleep 1 &&
  	kill -TERM $pid1 &&
  	kill -TERM $pid2 &&
          ${FLUX_BUILD_DIR}/t/rexec/rexec_ps -r 1 > output &&
          count=`cat output | wc -l` &&
          test "$count" = "0"
  
  not ok 27 - disconnect terminates all running processes
```

garlick:
Weird, same failure in bionic.  Another try.

Issue #4090, opened 2022-02-02
flux overlay disconnect <hostname> tries to act on rank 0


garlick:
Problem: flux overlay disconnect should accept a hostname, not just a rank.  Or at minimum it should not happly translate "fluke69" to 0.

PR #4095, opened 2022-02-03
flux-overlay: improve timeouts, hostname handling

garlick:
This fixes a couple of issues with `flux overlay` identified during testing on fluke yesterday.

### DIFF
--- a/src/cmd/builtin/overlay.c
+++ b/src/cmd/builtin/overlay.c
@@ -23,6 +23,8 @@
 
 #include "builtin.h"
 
+static double default_timeout = 0.5;
+
 static const char *ansi_default = "\033[39m";
 static const char *ansi_red = "\033[31m";
 static const char *ansi_yellow = "\033[33m";
@@ -39,7 +41,7 @@ static struct optparse_option status_opts[] = {
                " 2=show round-trip RPC times."
     },
     { .name = "timeout", .key = 't', .has_arg = 1, .arginfo = "FSD",
-      .usage = "Set RPC timeout (default none)",
+      .usage = "Set RPC timeout, 0=disable (default 0.5s)",
     },
     { .name = "summary", .has_arg = 0,
       .usage = "Show only the root subtree status."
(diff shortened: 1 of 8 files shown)


garlick:
Oops posted prematurely.   I should tack on a fix for #4090 to this PR.

grondo:
Oh, I really like the rename of `gethostbyrank` to `flux overlay lookup`.

I know you aren't quite done with this PR, but I did have one quick question.

It seems like on production clusters, disconnected brokers may be a common case. Does a 5s timeout for `flux overlay status` seem excessive, given that (if I'm correct about the common case) this will be the minimum time to run the command? Would it hurt to add a much smaller timeout by default (like maybe 200ms or less), and you already have the `--timeout` option to extend it? (Caveat: I'm probably missing some subtlety here where a 5s timeout is reasonable, and I apologize for bringing it up if so)

garlick:
The timeout really only comes into play with unresponsive nodes that have
not disconnected.  I kind of picked 5s out of nowhere though, and it is a
bit long.  Maybe a few hundred ms would be better, given that there is an
override in case that turns out to be too small.

garlick:
OK this is ready for a review.   I dropped the default `flux overlay status` timeout to 500ms and added the fix for #4090.

src/cmd/builtin/overlay.c
```diff
@@ -39,7 +41,7 @@ static struct optparse_option status_opts[] = {
                " 2=show round-trip RPC times."
     },
     { .name = "timeout", .key = 't', .has_arg = 1, .arginfo = "FSD",
-      .usage = "Set RPC timeout (default none)",
+      .usage = "Set RPC timeout, 0=disable (default 5s)",
```
grondo, line 44: Update default

grondo (approved):
This looks good to me. The `flux overlay lookup` command turned out nicely.

Just one small nit called out inline.

src/cmd/builtin/overlay.c
garlick: Good catch!  Forcing a push with that change and setting MWP.

Issue #4089, opened 2022-02-02
flux overlay disconnect on unresponsive node causes parent segfault


garlick:
Problem: flux overlay disconnect causes the parent rank to hang if the target node is unresponsive.

garlick:
It actually appears that this cause a segfault on rank 0.  The systemd journal says:
```
Feb 02 15:04:06 fluke1 bash[3520468]: broker.err[0]: broker on fluke69 (rank 66) has been unresponsive for 31.9987s
Feb 02 15:17:07 fluke1 bash[3520468]: job-exec.info[0]: job-exception: id=125187084783716352: lost contact with job shell on broker fluke69 (rank 66)
Feb 02 15:17:07 fluke1 systemd[1]: flux.service: Main process exited, code=killed, status=11/SEGV
Feb 02 15:17:07 fluke1 systemd[1]: flux.service: Failed with result 'signal'.
```

Meanwhile the downstream brokers are all still running, so apparently _they_ didn't lose their connection with the absentee parent.

garlick:
Bringing the rank 0 broker back up causes all the other nodes' brokers to exit with a nozero exit code, so systemd does not restart them.

Logs from fluke's login node:
```
Feb 02 15:51:58 fluke108 bash[2512908]: broker.crit[2]: fluke1 (rank 0) sent disconnect control message
Feb 02 15:51:58 fluke108 bash[2512908]: broker.err[2]: error sending groups.update request: No route to host
Feb 02 15:51:58 fluke108 bash[2512908]: broker.err[2]: state-machine.monitor: No route to host
Feb 02 15:51:58 fluke108 bash[2512908]: broker.info[2]: rc2-abort: run->cleanup 1.27607h
Feb 02 15:51:58 fluke108 bash[2512908]: broker.info[2]: cleanup-none: cleanup->shutdown 4.6392e-05s
Feb 02 15:51:58 fluke108 bash[2512908]: broker.info[2]: children-none: shutdown->finalize 3.6663e-05s
Feb 02 15:51:58 fluke108 bash[2512908]: broker.info[2]: rc3.0: running /etc/flux/rc3.d/01-sched-fluxion-qmanager-stop
Feb 02 15:51:58 fluke108 bash[2512908]: broker.info[2]: rc3.0: running /etc/flux/rc3.d/02-sched-fluxion-resource-stop
[...]

PR #4096, opened 2022-02-03
Fix broker segfault when an exception is raised on a job after completed prolog

grondo:
This fixes the issue described in #4089, where the job-manager perilog plugin will segfault on a job-exception raised for a job which completed a prolog/epilog.


grondo:
Hm, the bionic distcheck builder hit the same rexec test failure seen in #4094. Will open an issue.

garlick (approved):
LGTM!

grondo:
Thanks. Sweet MWP

Issue #4077, opened 2022-01-28
Interactive pty doesn't wait on attach for short-lived processes


grondo:
Problem: Something like `flux mini run -o pty.interactive tty` has an inherent race condition. Once the task exits the session is torn down and the job exits, thus there is no pty to which to attach, and in this case the output is lost.

Interactive ptys should probably wait for attach before closing the pty session and exiting the shell.

This might require disabling the doom shell plugin by default, and establishing a completion reference for the interactive pty, which would need to be released when the remote user detaches (or something).

PR #4098, opened 2022-02-03
libterminus: fix memory leaks in pty code

grondo:
This was split off from work on #4077. Still don't have the fix working, but these fixes for memory leaks seemed like they should go in sooner rather than later.

### DIFF
--- a/src/common/libterminus/client.c
+++ b/src/common/libterminus/client.c
@@ -312,7 +312,10 @@ static void pty_die (struct flux_pty_client *c, int code, const char *message)
      */
     if (code)
         c->wait_status = code << 8;
-    c->exit_message = strdup (message);
+    if (message) {
+        free (c->exit_message);
+        c->exit_message = strdup (message);
+    }
     notify_exit (c);
 }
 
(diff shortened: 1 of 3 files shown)


src/common/libterminus/client.c
```diff
@@ -316,6 +312,10 @@ static void pty_die (struct flux_pty_client *c, int code, const char *message)
         free (c->exit_message);
         c->exit_message = strdup (message);
     }
+    if (c->attached && (c->flags & FLUX_PTY_CLIENT_NOTIFY_ON_DETACH)) {
+        printf ("\033[999H[detached: %s]\033[K\n\r", c->exit_message);
+        fflush (stdout);
+    }
```
chu11, line 318: not knowing the code super well, now that `message` can be NULL, is there a scenario where `exit_message` has not yet been set?  Perhaps a small comment to indicate it will have been set by this point is all that's important.

grondo: Good point. Maybe it would be better to handle the case where the `exit_message` is NULL by printing `unknown reason`, though it is likely that this case can't be covered in testing.

grondo: Ok, I think I addressed your comment in the latest push.

garlick (approved):
LGTM

grondo:
I'll wait for @chu11 to approve to ensure I addressed his insightful comment.

chu11 (approved):
LGTM

grondo:
Thanks! Sweet MWP

PR #4099, opened 2022-02-04
broker: use TCP level keepalives

### LINKED ISSUE #4086: powered off nodes do not trigger disconnect
As noted offline by @ofaaland, powering off a node doesn't result in its disconnection, so a job that was running on the powered off node will remain in RUN state indefinitely, and RPCs to the node will hang.  As a consequence, irritatingly, `flux overlay status` hangs.

This is in contrast to the behavior when stopping a broker, where the overlay socket has a chance to be torn down.

As of 0.34, these nodes are drained, e.g.:
```
TIMESTAMP            RANK     REASON                         NODELIST
2022-02-02T15:04:06  66       broker was unresponsive        fluke69
```

Running `flux overlay disconnect` on the troubled node causes its parent rank to hang.  Boy what a horror show, I just hung fluke's rank 0.

garlick:
Enable TCP keepalives so Flux can disconnect from nodes that have been turned off.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -68,6 +68,10 @@ domainrefs = {
         'text': '%s(7)',
         'url': 'https://linux.die.net/man/7/%s',
     },
+    'linux:man8': {
+        'text': '%s(8)',
+        'url': 'https://linux.die.net/man/8/%s',
+    },
 }
 
 # Disable "smartquotes" to avoid things such as turning long-options
--- a/src/broker/overlay.c
+++ b/src/broker/overlay.c
@@ -114,6 +114,17 @@ static const double sync_max = 5.0;
 static const double default_torpid_min = 5.0;
 static const double default_torpid_max = 30.0;
 
+/* TCP keepalives for child connections.
+ * These are distinct from the application level keepalives described above
+ * and are needed to detect nodes that are turned off before they can send FIN.
+ * The values of -1 below mean "use OS default" and are 0MQ-specific.
+ * The OS defaults can be tuned with sysctl.
+ */
+static int default_keepalive_enable = 1;    // 0=disable, 1=enable
+static int default_keepalive_count = -1;
+static int default_keepalive_idle = -1;
+static int default_keepalive_interval = -1;
+
 struct overlay_monitor {
     overlay_monitor_f cb;
     void *arg;
@@ -138,6 +149,10 @@ struct overlay {
     int zmqdebug;
     double torpid_min;
     double torpid_max;
+    int keepalive_enable;
+    int keepalive_count;
+    int keepalive_idle;
+    int keepalive_interval;
 
     struct parent parent;
 
@@ -1223,10 +1238,12 @@ int overlay_connect (struct overlay *ov)
         zsock_set_unbounded (ov->parent.zsock);
         zsock_set_linger (ov->parent.zsock, 5);
         zsock_set_ipv6 (ov->parent.zsock, ov->enable_ipv6);
+
         zsock_set_zap_domain (ov->parent.zsock, FLUX_ZAP_DOMAIN);
         zcert_apply (ov->cert, ov->parent.zsock);
         zsock_set_curve_serverkey (ov->parent.zsock, ov->parent.pubkey);
         zsock_set_identity (ov->parent.zsock, ov->uuid);
+
         if (zsock_connect (ov->parent.zsock, "%s", ov->parent.uri) < 0)
             goto nomem;
         if (!(ov->parent.w = zmqutil_watcher_create (ov->reactor,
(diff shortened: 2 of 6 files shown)


src/broker/overlay.c
```diff
@@ -1755,13 +1780,24 @@ static int overlay_configure_torpid (struct overlay *ov)
...
                               &error,
-                              "{s?{s?s s?s}}",
+                              "{s?{s?s s?s s?i s?i s?i}}",
                               "tbon",
                                 "torpid_min", &min_fsd,
-                                "torpid_max", &max_fsd) < 0) {
+                                "torpid_max", &max_fsd,
+                                "keepalive_cnt", &ov->keepalive_cnt,
+                                "keepalive_idle", &ov->keepalive_idle,
+                                "keepalive_intvl", &ov->keepalive_intvl) < 0) {
```
grondo, line 1789: For config files, I'd argue against abbreviating these values unnecessarily (only saving 2 and 3 characters), i.e. `keepalive_count`, `keepalive_idle`, and `keepalive_interval`. However, that is just my opinion.

src/broker/overlay.c
```diff
@@ -1793,6 +1829,25 @@ static int overlay_configure_torpid (struct overlay *ov)
                          set_torpid,
                          ov) < 0)
         return -1;
+
+    /* Only allow command line override, not runtime on these attrs
+     */
+    if (overlay_configure_attr_int (ov->attrs,
+                                    "tbon.keepalive_cnt",
```
grondo, line 1836: Same comment about the broker attributes - I don't see the value in abbreviating `count` to `cnt`.

Also it seems like it would be really handy to be able to tune these values at runtime without requiring a broker restart, but I'm guessing this is difficult because of the way the `zsock` options have to be set?

grondo (commented):
I know you're still working on this, but a couple high level comments came to mind when peeking at this so I thought I'd offer them up to be shot down sooner rather than later.

src/broker/overlay.c
garlick: Good idea, I was just using the zeromq names for these things, which are made up anyway.  Will do.

src/broker/overlay.c
garlick: Unfortunately these have to be set before the socket is bound, so there's no way to change it at runtime.

garlick:
Changed the names as suggested and updated man pages.

garlick:
I realized my guesses for reasonable keepalive timing parameters probably should not be the default values for Flux.  Some sites might tune the system defaults on a cluster (`sysctl -a|grep keepalive`), so it is probably better to accept those values as the default and let them be overridden by the flux config / broker attributes.

The downside is that the linux defaults would take hours to discover a turned off node, so for example on fluke where the defaults are not altered, we will definitely want to suggest some tighter values for the flux config.

garlick:
This is probably ready for review, although I'll leave the WIP there until i can verify it's working as advertised on fluke.

src/broker/overlay.c
grondo: Ah I didn't realize that you were using the zeromq names, and that those in turn were named after the sysctl settings. I kind of feel dumb for my suggestion now..

garlick: Actually zeromq names are a little tweaked compared to the system names, e.g. system "probes" vs zmq "cnt", system "time" vs "idle".  I thought your suggestion was a good one.

src/broker/overlay.c
```diff
@@ -114,6 +114,17 @@ static const double sync_max = 5.0;
 static const double default_torpid_min = 5.0;
 static const double default_torpid_max = 30.0;
 
+/* TCP keepalives for child connections.
```
grondo, line 117: typo in commit message for 2c0bb0e: `grpe`

doc/conf.py
```diff
@@ -68,6 +68,10 @@
         'text': '%s(7)',
         'url': 'https://linux.die.net/man/7/%s',
     },
+    'linux:man8': {
```
grondo, line 71: Commit message typo in 6e78a76: s/references/referenced/

t/t0013-config-file.t
```diff
@@ -212,4 +212,75 @@ test_expect_success '[bootstrap] curve_cert file must be mode o-r' '
...
+		-Stbon.keepalive_enable=1 \
+		-Stbon.keepalive_idle=30 \
+		-Stbon.keepalive_interval=1 \
+		-Stbon.keepalive_count=5 \
+		flux lsattr -v >attrs3.out &&
+	grep "^tbon.keepalive_" attrs3.out | sort >attrs_keepalive3.out &&
+	test_cmp attrs_keepalive3.exp attrs_keepalive3.out
+'
+
+test_expect_success 'tbon.keepalive_enable attr value check works' '
```
grondo, line 269: Not sure if adding checks for values of other keepalive attributes/config would be too onerous here.

grondo (approved):
LGTM! Just called out a couple typos in commit messages.

t/t0013-config-file.t
garlick: Yeah, I was being a bit lazy there :-)   OK, will add.

garlick:
OK, fixed up those things.  I'll go ahead and set MWP.  Thanks!

PR #4100, opened 2022-02-04
systemd: set restart=always

### LINKED ISSUE #3818: change systemd unit file to restart=always
Problem: brokers currently are not restarted by systemd if they exit with a non-zero return code.
    
The "subtree panic" scheme could result in brokers exiting with non-zero return codes, through no fault of their own.  The desired behavior is the subtree immediately restarts and attempts to connect to the instance.
    
Change the systemd unit file from 'restart=on-success' to 'restart=always'.

@grondo commented:

From [systemd.service(5)](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Restart=):
> If set to on-success, it will be restarted only when the service process exits cleanly. In this context, a clean exit means any of the following:
>
> * exit code of 0;
> * for types other than Type=oneshot, one of the signals SIGHUP, SIGINT, SIGTERM, or SIGPIPE;
> * exit statuses and signals specified in `SuccessExitStatus=`.

Maybe we could define an exit status for "subtree panics" which could then be used with `SuccessExitStatus=` so that failing brokers are only restarted in that situation?

I'd be worried about a bad config or other fatal condition resulting in systemd restarting flux continually.

_Originally posted by @grondo in https://github.com/flux-framework/flux-core/pull/3816#discussion_r687052781_

garlick:
Problem: when an upstream broker fails, downstream brokers need to
be manually restarted.

This is sort of as designed but it's a pain.

Switch to Restart=always mode so the broker always tries again,
and sys admins won't have to mop up when something bad happens.

Fixes #3818

### DIFF
--- a/etc/flux.service.in
+++ b/etc/flux.service.in
@@ -22,7 +22,7 @@ ExecStart=/bin/bash -c '\
   -Sbroker.quorum-timeout=none \
 '
 ExecReload=@X_BINDIR@/flux config reload
-Restart=on-success
+Restart=always
 RestartSec=5s
 User=flux
 Group=flux


grondo (approved):
LGTM!

Issue #4097, opened 2022-02-03
t0005-rexec.t: not ok 27 - disconnect terminates all running processes


grondo:
This test recently started failing often in bionic builders:
```
expecting success:
 ${FLUX_BUILD_DIR}/t/rexec/rexec -r 1 sleep 100 &
 pid1=$!
 ${FLUX_BUILD_DIR}/t/rexec/rexec -r 1 sleep 100 &
 pid2=$!
 sleep 1 && 
 ${FLUX_BUILD_DIR}/t/rexec/rexec_ps -r 1 > output && 
 count=`cat output | wc -l` && 
 test "$count" = "2" && 
 sleep 1 && 
 kill -TERM $pid1 && 
 kill -TERM $pid2 && 
 ${FLUX_BUILD_DIR}/t/rexec/rexec_ps -r 1 > output && 
 count=`cat output | wc -l` && 
 test "$count" = "0"

 not ok 27 - disconnect terminates all running processes
```

chu11:
Thinking about this test failure, I think that this:

```
 kill -TERM $pid1 && 
 kill -TERM $pid2 && 
 ${FLUX_BUILD_DIR}/t/rexec/rexec_ps -r 1 > output && 
 count=`cat output | wc -l` && 
 test "$count" = "0"
```

is simply racy on a busy system.  the child watcher has to wake up and recognize process has died and completion callback has to be called as well (done via reactor too).  So putting the count check in a loop with a "timeout" should be more than sufficient.

grondo:
Great! This will be a good one to get fixed!

chu11:
```
        sleep 1 &&                                                                                                                          
        ${FLUX_BUILD_DIR}/t/rexec/rexec_ps -r 1 > output &&                                                                                 
        count=`cat output | wc -l` &&                                                                                                       
        test "$count" = "2" &&        
```

in addition we can remove these sleeps, which I imagine were to resolve "raciness" before.

Issue #4083, opened 2022-02-01
t2612-job-shell-pty.t: hang in test for pty.interactive+pty.capture


grondo:
@garlick noticed in #4078:

> Looks like the shell pty test got stuck here in the coverage run. I'll restart.
>
>   expecting success: 
>  	id=$(flux mini submit -o pty.interactive -o pty.capture tty) &&
>  	flux job attach $id >ptyim.out 2>&1 &&
> 	flux job attach $id

Perhaps there is a small race here that could cause a hang.

grondo:
I was able to reproduce this locally by running the following in a loop, after about 250 iterations:
```sh
#!/bin/sh
id=$(flux mini submit -o pty.interactive -o pty.capture tty) &&
flux job attach $id 
```

In the reproducer,  `flux job attach` is hanging in `flux_pty_client_attach()` waiting for an RPC response from `<uid>-shell-<jobid>.terminus.0`. I'm guessing this hang occurs because the request is delivered before the shell exits (so no `ENOSYS`), but since these tests are using short-lived processes (see also #4077), the shell exits before a response can be sent.

To make matters worse, SIGINT is somehow disabled at this point so you can't Ctrl-C out of the `flux job attach`.

It seems likely a fix for #4077 would address this issue, since the pty server would keep the shell actively connected until at least one client connects for an "interactive" pty, and we could also install a completion reference to ensure the shell doesn't go away until the pty service has been removed.

PR #4101, opened 2022-02-04
shell: do not allow instance owner to use guest shell services

grondo:
The system instance owner is by design an unprivileged user which should not have access to guest user services provided by the job shell, but currently most shell services allow FLUX_ROLE_OWNER.

Lock down access to shell-provided services to the job owner only.


garlick (approved):
Oooh, _very_ good catch.

grondo:
Thanks! Setting MWP.

grondo:
I've had to restart tests 3 times -- hitting either #4097 or #4083. Trying again now...

PR #4102, opened 2022-02-04
shell: fix race in attach to interactive job pty

grondo:
This PR reworks the pty code to close a race condition when attaching to a short-lived process which has been allocated a pty.

The major visible change here is that, when running a job with `-o pty.interactive`, the shell will keep the rank 0 pty available until at least one client attaches, even if the task has exited. If no client attaches the shell `doom` plugin will terminate the job after 30s by default. This can be disabled via `-o exit-timeout=-1`, in which case the shell could hang until the job times out waiting for a client to attach.

### DIFF
--- a/src/common/libterminus/terminus.c
+++ b/src/common/libterminus/terminus.c
@@ -69,6 +69,7 @@ extern char **environ;
 
 struct terminus_session {
     struct flux_terminus_server *server;
+    int refcount;
 
     char *name;
     char topic[128];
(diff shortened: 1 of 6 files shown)


grondo:
I have introduced one issue in the refactor of the shell pty plugin. Hold off on a review until I can push the fix.

grondo:
Ok, this is probably ready for a review

src/common/libterminus/terminus.c
```diff
@@ -207,6 +212,11 @@ terminus_session_create (struct flux_terminus_server *ts,
     if (!(s->pty = flux_pty_open ()))
         goto error;
     flux_pty_set_flux (s->pty, ts->h);
+    flux_pty_aux_set (s->pty, "terminus_session", s, NULL);
```
garlick, line 215: there's a malloc inside `flux_pty_aux_set()` so check for failure?

garlick (approved):
Just spotted one minor nit, o/w LGTM!

src/common/libterminus/terminus.c
grondo: Good catch, thanks!

grondo:
I fixed up the missing check for failure (and similarly a missing check for return code from `flux_pty_set_flux()`) and squashed the result.

There also appeared to be a new test hang in one of the tests that unnecessarily ran `bash` under a pty as the test program, and then canceled the job to get `flux job attach` to exit. I think something about using `bash` here was the cause of the hangs, so I've switched to the simpler `tty` program, which has a side benefit of exiting with error if the setup of the pty fails, plus it should exit nearly immediately which is more in the spirit of the test.

We'll see if the hangs are resolved.

PR #4104, opened 2022-02-05
docs: add Python overview documentation

jameshcorbett:
Adds documentation covering some basic usage for the Python package, I wanted to get some feedback before I add much more. Currently covers some of the major ``flux.job`` functions. Still missing is a description of:

- RPCs
- The Flux reactor
- Watchers

Let me know what else you think should be added.


chu11:
can't speak for others, but I always think that more documentation is better and I think this is a good start!

only high level question is how all of this will go into the read the docs?  It feels like some of the documentation here is sort of "overview" or "beginners guide"-ish.  So it should go under the top level flux-framework read the docs page.  But some feels like it will become more "formal" documentation and should go under the flux-core sub-project of read the docs.

So perhaps some of this should go into https://github.com/flux-framework/flux-docs instead?

grondo:
I'm definitely not a sphinx expert, but I assume the Python docs have to stay with this project so that the `autoclass` and related magic works.

Probably there is a way with sphinx to link to this documentation from the top level in the flux-docs site, once this PR has been merged?

BTW, I see no reason this should not go in as is. If there's more detail or changes to add those could go in as future PRs IMO.

jameshcorbett:
I don't know whether it's feasible to get the ``autoclass`` magic to work in ``flux-docs``, but I think it's definitely possible. I think we would need to build ``flux-docs`` in a place where the ``flux`` python package was importable, which means building flux-core.

I'm pretty sure there is a way to link flux-docs and flux-core without hardcoding links but I don't know how to do it (although I think it's called intersphinx), I'll look into it.

Marked as ready for review, will rebase with master though.

doc/python/basics.rst
```diff
@@ -0,0 +1,58 @@
...
+Note however that if you import the ``flux`` package from outside a Flux
+instance, you will need to specify a Flux instance to communicate with,
+or most of the package's operations will fail.
+
+
+Additional Examples
+-------------------
+
+In addition to the examples given in these pages, there is
+a set of detailed examples maintained `here <https://flux-framework.readthedocs.io/projects/flux-workflow-examples/en/latest/>`_.
```
chu11, line 31: instead of `here` how about something like "amongst the workflow examples"

doc/python/job_submission.rst
```diff
@@ -0,0 +1,196 @@
...
+* Cancelling jobs
+* Viewing job info and events
+
+
+Job submission
+--------------
+
+Job submission is performed by creating a ``flux.job.Jobspec`` object,
+populating it with attributes, and then passing it to one of the submission
+functions, which will be described later. Jobspec objects define everything
```
chu11, line 20: probably can shorten, "one of the submission functions described below"?

doc/python/job_submission.rst
```diff
@@ -0,0 +1,196 @@
...
+
+The ``FluxExecutor`` interface
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Basic ``FluxExecutor`` usage consists of creating an executor,
+submitting jobspecs to it, and then attaching callbacks to those
+futures or waiting for them to complete. Executors must have
+``.shutdown()`` called when they are no longer to be used. However,
+they also support the context-manager protocol (i.e. ``with executor ...:``)
+which will calls ``shutdown``.
```
chu11, line 120: English tyop: calls -> call

src/bindings/python/flux/job/Jobspec.py
```diff
@@ -299,20 +301,24 @@ def _create_slot(cls, label, count, with_child):
 
     @property
     def duration(self):
+        """Job's time limit.
+
+        The duration may be:
+
+        * an int or float in seconds
+        * a string in Flux Standard Duration
```
chu11, line 309: perhaps can note RFC23

chu11 (approved):
LGTM, although it'd probably be good to have another person give it a once over since it is a lot of new documentation

grondo:
> I think we would need to build flux-docs in a place where the flux python package was importable, which means building flux-core.

It looks like building flux-core isn't necessarily required, since the readthedocs [pr build](https://readthedocs.org/projects/flux-core/builds/16013032/) somehow magically works with just the sources.

I'm still not sure if moving any of the documentation that uses autoclass to flux-docs is a good idea, because this could let the documentation in flux-docs get out of date unless we somehow trigger a rebuild on every flux-core update (in case there was a change or addition to the docstrings in the python documentation)

We do already have some form of intersphinx set up in the flux-docs repo, so maybe linking to these docs at the top-level will already "just work"

grondo:
I already went through the new docs once and they seem like an excellent addition! Thanks @jameshcorbett!

I'll take another editing pass, but I doubt I'll find much of anything.

Some high-level, very trivial comments though:

 * It would be better if the changes that are not documentation were split out of the `docs: add high-level..` commit. There are some changes in there that aren't clear on first look and a commit message with `Problem:` statement would really help (e.g. addition of `-T` to apidoc options). The removal of the `.gitignore` should perhaps even be its own commit in this case (since it probably causes a specific problem)
 * All commits should have a body, preferably with `Problem:` statement explaining the issue being resolved.

jameshcorbett:
OK to set MWP? I think I improved the commits.

grondo:
Yes, nice work! That is very helpful.

I just made a small tweak to the title since it will automatically be incorporated in release notes. Feel free to set MWP

PR #4105, opened 2022-02-05
flux-ping: support hostnames in TARGET

garlick:
This changes `flux-ping` to accept a hostname as part of the TARGET instead of a rank, e.g.
```
$ ./flux ping system76-pc
0!broker.ping pad=0 seq=0 time=0.200 ms (1c36d225!b4330dc7!b6026d87)
0!broker.ping pad=0 seq=1 time=0.198 ms (1c36d225!b4330dc7!b6026d87)
^C
```
or
```
$ ./flux ping 'system76-pc!kvs'
0!kvs.ping pad=0 seq=0 time=0.232 ms (85d88d89!b4330dc7!b6026d87!e92cbac2)
0!kvs.ping pad=0 seq=1 time=0.363 ms (85d88d89!b4330dc7!b6026d87!e92cbac2)
^C
```
It also adds a general function for resolving hosts to ranks:
```c
/* Find the lowest numbered broker rank running on 'host', by consulting
 * the "hostlist" attribute.
 * Returns rank on success, -1 on failure with errno set.
 */
int flux_get_rankbyhost (flux_t *h, const char *host);
```

### DIFF
--- a/src/cmd/flux-ping.c
+++ b/src/cmd/flux-ping.c
@@ -214,48 +214,56 @@ void timer_cb (flux_reactor_t *r, flux_watcher_t *w, int revents, void *arg)
     }
 }
 
-int parse_nodeid (const char *s, uint32_t *np)
+int parse_nodeid (struct ping_ctx *ctx, const char *input, uint32_t *nodeid)
 {
-    uint32_t n;
+    int rank;
     char *endptr;
 
+    if (!strcmp (input, "any")) {
+        *nodeid = FLUX_NODEID_ANY;
+        return 0;
+    }
+    if (!strcmp (input, "upstream")) {
+        *nodeid = FLUX_NODEID_UPSTREAM;
+        return 0;
+    }
+    if ((rank = flux_get_rankbyhost (ctx->h, input)) >= 0) {
+        *nodeid = rank;
+        return 0;
+    }
     errno = 0;
-    n = strtoul (s, &endptr, 10);
-    if (errno != 0 || *endptr != '\0')
-        return -1;
-    *np = n;
-    return 0;
+    rank = strtol (input, &endptr, 10);
+    if (errno == 0 && *endptr == '\0' && rank >= 0) {
+        *nodeid = rank;
+        return 0;
+    }
+    return -1;
 }
 
-/* Parse 's' to obtain the ping topic string and (optionally) nodeid.
- * If nodeid is non-NULL, then try to parse a nodeid from 's'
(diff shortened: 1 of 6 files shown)


chu11:
High level comment

> 0!broker.ping pad=0 seq=1 time=0.198 ms (1c36d225!b4330dc7!b6026d87)

If user input `hostname` should that be output instead of the rank `0`?  Might make things clearer?

If we're consistent to how `ping` works, then it should output rank.  But it does help that `ping` outputs a "header":

```
PING www.google.com (142.250.81.228) 56(84) bytes of data.
64 bytes from lga25s74-in-f4.1e100.net (142.250.81.228): icmp_seq=1 ttl=111 time=76.5 ms
```

so perhaps

```
FLUX-PING systemd76-pc (rank 0)
0!broker.ping pad=0 seq=1 time=0.198 ms (1c36d225!b4330dc7!b6026d87)
```

??? just a thought

src/cmd/flux-ping.c
```diff
@@ -279,6 +288,12 @@ int main (int argc, char *argv[])
         log_msg_exit ("optparse_add_option_table");
     if ((optindex = optparse_parse_args (opts, argc, argv)) < 0)
         exit (1);
+    if (optindex != argc - 1) {
+        optparse_print_usage (opts);
+        exit (1);
+    }
+    if (!(target = strdup (argv[optindex])))
```
grondo, line 295: minor: is `target` leaked? (didn't see a `free (target)` but may have missed it.

grondo (approved):
Nice improvement!
Just a couple very minor things noted inline.

src/cmd/flux-ping.c
garlick: Oops, yup!

grondo:
>  But it does help that ping outputs a "header":

I like that idea, `flux ping` could always output rank and hostname (when available) in a header. (though I'd drop the `FLUX-PING` and use `flux-ping: ` to match logging output of all other commands)

This would be interesting information when pinging by rank, though it wouldn't work when pinging a service on "any" rank (in that case maybe the right approach is to output the hostname along with rank/service in each line.)

Since we're trying to push out a release, maybe we can open an issue on this and add the improvement later. Up to @garlick of course.

garlick:
Yeah, I like where that discussion is going but in the interest of time lets defer.  I'll open an issue.  Thanks!

(pushing fixes for the review comments)

garlick:
I'll go ahead and set MWP.

PR #4110, opened 2022-02-07
broker: use RPC not broker module sync/status

garlick:
Problem: the broker and its modules communicate status using
special control messages, but this communication would be more
clear if regular RPCs were used, with no disadvantage.

Replace low level message code with regular RPCs.  Register
a broker service named `broker.module-status` instead of hard-wiring
module status updates in the broker's low level message callback.

In addition, update `flux_module_set_running()` to use the RPC.

Update tests.

### DIFF
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -1153,6 +1153,59 @@ error:
         flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
 }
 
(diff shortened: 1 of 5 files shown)


grondo (approved):
LGTM! Nice touch adding the EPROTO and other expected failure test.

garlick:
Thanks!  I"ll go ahead and MWP this one.

PR #312, opened 2015-08-03
modules/resource_hwloc: new hwloc resource grokker

trws:
Currently in for discussion, not ready for general use as yet.  When loaded on
all ranks, `flux module load -r all resource_hwloc`, and loaded, `flux event
pub resource.load`, populates the KVS with resource information for the flux
instance.  The full topology for each node is stored as xml at
`resource.hwloc.xml.<rank>` in the KVS, and a complete tree of resources,
including IO and caches, by individual key and value is available at
`resource.hwloc.<rank>.[<resource_type>_<resource_logical_id>...]`. It has not
been instructed to look outside of administrative limitations, but that hasn't
been tested either.

PR #4112, opened 2022-02-08
rename keepalive messages to control messages

garlick:
This changes all references to application level keepalives and the Flux keepalive message type to "control messages" per the proposed change to RFC 3 in flux-framework/flux-core#312.

This should avoid some confusion between TCP keepalives, which we now support, and these control messages, now only used between brokers after #4110.


grondo (approved):
LGTM!

garlick:
Thanks!  setting MWP.

PR #4119, opened 2022-02-09
doc: update flux-broker-attributes(7)

garlick:
This adds missing broker attributes to flux-broker-attributes(7) and reorganizes its sections.

Also there's a fix for an error in flux-config-tbon(5).

### DIFF
--- a/doc/man7/flux-broker-attributes.rst
+++ b/doc/man7/flux-broker-attributes.rst
@@ -173,29 +173,12 @@ log-level
 CONTENT ATTRIBUTES
 ==================
 
-content.acct-dirty
-   The number of dirty cache entries on this rank.
-
-content.acct-entries
-   The total number of cache entries on this rank.
-
-content.acct-size
-   The estimated total size in bytes consumed by cache entries on
-   this rank, excluding overhead.
-
-content.acct-valid
-   The number of valid cache entries on this rank.
-
 content.backing-module
    The selected backing store module, if any. This attribute is only
    set on rank 0 where the content backing store is active.
 
 content.blob-size-limit
    The maximum size of a blob, the basic unit of content storage.
-
-content.flush-batch-count
-   The current number of outstanding store requests, either to the
-   backing store (rank 0) or upstream (rank > 0).
 
 content.flush-batch-limit
    The maximum number of outstanding store requests that will be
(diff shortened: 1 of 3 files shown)


doc/man7/flux-broker-attributes.rst
```diff
@@ -34,22 +37,86 @@ rundir
...
+
+broker.rc3_path
+   The path to the broker's rc3 script.
+
 broker.starttime
    Timestamp of broker startup from :man3:`flux_reactor_now`.
 
+conf.connector_path
+   A colon-separated list of directories that are searched by :man3:`flux_open`
+   to resolve the scheme component of a URI.
```
grondo, line 93: This might be confusing since the attribute does not actually influence `flux_open(3)` but is instead inherited from the `FLUX_CONNECTOR_PATH` environment variable (and I'm not sure anything is even done with it after that). Therefore it is in some ways informational only.

Similar comment to the other `conf.*` attributes (except `pmi_library_path` I think), though `conf.exec_path` is used by `flux(1)`, but only when used with the `--parent` option.

Also, I note that these attributes can be set. Maybe a separate issue would be whether that should be possible.

grondo (approved):
This LGTM.

I just have one comment inline, but whether we address that now or open an issue before we decide what to do is up to you.

doc/man7/flux-broker-attributes.rst
garlick: Good point - I'll update the documentation to reflect that.

Just opened #4120 about making them immutable since that probably should be a separate release notes item.

garlick: I ended up auditing all of the entries to add defaults, and annotations showing whether each attribute can be updated on the command line or at runtime.

grondo: Nice!

garlick: I noticed a readthedocs rendering issue in the RESOURCE section so if it's OK I'd like to squash the fixup and force push.

grondo: Yeah, sounds good

garlick: OK, pushed.

grondo:
Checked over the most recent changes and looks really nice! Feel free to set MWP as far as I'm concerned.

garlick:
Thanks!

Issue #4117, opened 2022-02-09
shell: log messages from the shell don't make it into file with --output/--error


grondo:
This is a usability issue:
```console
$ flux mini batch -o pty.boop --wrap -n1 echo what the what
ƒHwvRQqf48o
$ cat flux-ƒHwvRQqf48o.out
$ flux jobs ƒHwvRQqf48o
       JOBID USER     NAME       ST NTASKS NNODES  RUNTIME NODELIST
 ƒHwvRQqf48o grondo   echo        F      1      1   0.265s fluke102
 $ flux job attach ƒHwvRQqf48o
0.296s: job.exception type=exec severity=0 invalid shell pty option: 1 object item(s) left unpacked: boop
flux-job: task(s) exited with exit code 1
0: stdout redirected to flux-ƒHwvRQqf48o.out
0: stderr redirected to flux-ƒHwvRQqf48o.out
0.294s: flux-shell[0]: FATAL: pty: invalid shell pty option: 1 object item(s) left unpacked: boop
```

The default output file for my batch job is empty, but the job has failed. The errors can only be seen in the output eventlog.

When the `--output` and/or `--error` options are used, the shell should also redirect logging to the same destination as stderr.

grondo:
Possibly this worked as expected before but was broken by 2baa3c4161f33c01822dfc075751e5d8bec8d94f, though I don't think the shell's stderr ever was redirected to the `--output` destination.

PR #4125, opened 2022-02-11
shell: capture job shell error messages in designated output file

grondo:
This PR is a stopgap fix for #4117. 

When an output or error file is specified, re-route log messages to the same output file as stderr, instead of sending them to the output eventlog.

Since the output files aren't opened until after the shell has connected to Flux and initialized, some log messages and errors still end up in the output eventlog, e.g. if something goes wrong before shell.init, but at least there isn't likely to be an _empty_ output file when a job fails after it gets that far, so the spirit of the issue is resolved.

Eventually, perhaps a better fix would be to have the exec system read the output and error options from the jobspec and, if the output type is `file`, pass this along in new `--output` and `--error` options to the job shell. This would allow the job shell to open output file immediately, and even the earliest of errors could be captured. However, I'm not sure at this point if that is the "right" fix.

before:
```console
$ id=$(flux mini submit --wait --output=example.out nosuchcommand) && flux job status $id; ls -l example.out; cat example.out
ƒJSBRFaBzRm: exception: type=exec note=task 0: start failed: nosuchcommand: No such file or directory
-rw-r--r-- 1 grondo grondo 0 Feb 11 07:43 example.out
``` 

after:
```console
$ id=$(flux mini submit --wait --output=example.out nosuchcommand) && flux job status $id; ls -l example.out; cat example.out
ƒ2uW3UFh: exception: type=exec note=task 0: start failed: nosuchcommand: No such file or directory
-rw-r--r-- 1 grondo grondo 85 Feb 11 07:44 example.out
flux-shell[0]: FATAL: task 0: start failed: nosuchcommand: No such file or directory
```

### DIFF
--- a/src/shell/output.c
+++ b/src/shell/output.c
@@ -455,6 +455,23 @@ static int shell_output_file (struct shell_output *out)
     return 0;
 }
 
+static void shell_output_eof (struct shell_output *out,
+                              flux_msg_handler_t *mh)
+{
+    if (--out->eof_pending == 0) {
+        flux_msg_handler_stop (mh);
+        if (flux_shell_remove_completion_ref (out->shell, "output.write") < 0)
+            shell_log_errno ("flux_shell_remove_completion_ref");
+
+        /* no more output is coming, flush the last batch of output */
+        if ((out->stdout_type == FLUX_OUTPUT_TYPE_KVS
+            || (out->stderr_type == FLUX_OUTPUT_TYPE_KVS))) {
+            if (eventlogger_flush (out->ev) < 0)
+                shell_log_errno ("eventlogger_flush");
+        }
+    }
+}
+
 static int shell_output_write_leader (struct shell_output *out,
                                       const char *type,
                                       json_t *o,
@@ -494,20 +511,8 @@ static int shell_output_write_leader (struct shell_output *out,
         shell_log_error ("json_array_clear failed");
         goto error;
     }
-    if (eof) {
-        if (--out->eof_pending == 0) {
-            flux_msg_handler_stop (mh);
-            if (flux_shell_remove_completion_ref (out->shell, "output.write") < 0)
-                shell_log_errno ("flux_shell_remove_completion_ref");
-            /* no more output is coming, flush the last batch of
-             * output */
-            if ((out->stdout_type == FLUX_OUTPUT_TYPE_KVS
-                 || (out->stderr_type == FLUX_OUTPUT_TYPE_KVS))) {
-                if (eventlogger_flush (out->ev) < 0)
-                    shell_log_errno ("eventlogger_flush");
-            }
-        }
-    }
+    if (eof)
+        shell_output_eof (out, mh);
     return 0;
 error:
     return -1;
--- a/src/shell/output.c
+++ b/src/shell/output.c
@@ -184,7 +184,7 @@ static int shell_output_redirect_stream (struct shell_output *out,
                                          const char *stream,
                                          const char *path)
 {
-    struct idset *idset;
+    struct idset *idset = NULL;
     json_t *entry = NULL;
     char *entrystr = NULL;
     int saved_errno, rc = -1;
@@ -236,6 +236,7 @@ error:
     json_decref (entry);
     free (entrystr);
     free (rankptr);
+    idset_destroy (idset);
     errno = saved_errno;
     return rc;
 }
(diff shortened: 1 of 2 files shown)


src/shell/output.c
```diff
@@ -409,15 +409,17 @@ static int shell_output_file (struct shell_output *out)
...
                                       json_t *o,
                                       flux_msg_handler_t *mh) // may be NULL
 {
     bool eof = false;
     json_t *entry;
 
-    if (iodecode (o, NULL, NULL, NULL, NULL, &eof) < 0)
+    if (strcmp (type, "data") == 0
+        && iodecode (o, NULL, NULL, NULL, NULL, &eof) < 0)
         goto error;
```
garlick: errno is not set if the strcmp condition is true but `shell_output_write_leader() ` is expected to set errno or `flux_respond_error()` will fail with EINVAL.

src/shell/output.c
```diff
@@ -409,15 +409,17 @@ static int shell_output_file (struct shell_output *out)
 }
```
garlick, line 453: in commit message for 8e13b6ee8e057b2c0a879672e5d357f5867ce76b s/feild/field/ and s/commont/common/

src/shell/output.c
```diff
@@ -345,6 +345,47 @@ static int shell_output_write_fd (int fd, const void *buf, size_t len)
...
+        output_type = out->stderr_type;
+        ofp = &out->stderr_file;
+    }
+    if ((output_type == FLUX_OUTPUT_TYPE_FILE) && len > 0) {
+        if (ofp->label
+            && shell_output_label (ofp, rank) < 0)
+            return -1;
+        if (shell_output_write_fd (ofp->fdp->fd, data, len) < 0)
+            return -1;
+    }
```
garlick, line 384: Is `data` leaked on error?

src/shell/output.c
```diff
@@ -84,7 +84,6 @@ struct shell_output {
     struct eventlogger *ev;
```
garlick, line 84: In commit message for 31277bad4117442931d3adcf0e7a6a1981751990 s/refernence/reference/

garlick (commented):
Whoops I started making review comments but I think you fixed some of them as I was making them.  I'll submit what I have so far in case anything is useful and then try again when you're ready.

src/shell/output.c
garlick: Pretty sure you just made this a non-error.

grondo: Yeah, iirc we don't need to separately check for `eof` anymore so this conditional could be dropped.

src/shell/output.c
grondo: yes, good catch!

grondo: I was going to add a separate commit for this fix since it was present in existing code, however I just squashed it into the refactor commit. Hope that was ok!

src/shell/output.c
```diff
@@ -458,7 +459,7 @@ static void shell_output_decref (struct shell_output *out,
                                  flux_msg_handler_t *mh)
 {
     if (--out->refcount == 0) {
-        flux_msg_handler_destroy (mh);
+        flux_msg_handler_stop (mh);
```
chu11, line 462: i think this flux_msg_handler_destroy() -> flux_msg_handler_stop() is in the wrong commit?

grondo: Good catch, it should have never been destroyed and I tried to fix up a development typo in a rebase and must have messed that up. Let me see if I can fix it up. Thanks!

grondo: Ah, yeah it was supposed to be squashed into 4d1e7f5

grondo: Ok, I think I've fixed that.

chu11:
re-pushed, since the last time I squash

- add `sdexec_cancel()`, to handle corner case when `sdprocess_exec()` fails.
- re-add stdout/stderr logging to `exec.eventlog`, make logging configurable and fix bug in which logging didn't work
- did some refactoring
- add tests for much fo the above
- added basic system / multiuser tests for `sdexec`

Note that the system / multiuser tests use `system.exec.sd.test=true` attribute to make sure tests are run under sdexec / systemd.  This can be altered later if we make `sdexec` the default.  Unsure if we'd want to update the default `exec.toml` or create one in the fluxorama image that makes `sdexec` the default.   Thoughts?

grondo:
rebased.

garlick (approved):
This LGTM!

grondo:
Thanks! Set MWP

PR #4126, opened 2022-02-11
resource: emit a more specific error when rlist_rerank() fails

grondo:
This PR adds an `rlist_error_t` parameter to `rlist_rerank()` which offers a more specific error when the rerank operation fails.

The parameter is then used in the `resource` module to print a specific error along with the general `error reranking R`.

Finally, to avoid the confusing error message when the `resource` module exits due to `rlist_rerank`'s use of `errno` like `ENOSPC` and `EOVERFLOW`, reset `errno` to `EINVAL` when `rlist_rerank()` fails.

### DIFF
--- a/src/cmd/flux-R.c
+++ b/src/cmd/flux-R.c
@@ -562,7 +562,7 @@ int cmd_rerank (optparse_t *p, int argc, char **argv)
         log_err_exit ("Failed to transform R objects on stdin");
     if (argc != 2)
         log_err_exit ("Must provide a hostlist for re-ranking");
-    if (rlist_rerank (rl, argv[1]) < 0) {
+    if (rlist_rerank (rl, argv[1], NULL) < 0) {
         if (errno == ENOENT)
             log_msg_exit ("failed to find one or more provided hosts in R");
         else if (errno == EOVERFLOW)
--- a/src/broker/module.c
+++ b/src/broker/module.c
@@ -184,7 +184,7 @@ static void *module_thread (void *arg)
         mod_main_errno = errno;
         if (mod_main_errno == 0)
             mod_main_errno = ECONNRESET;
-        flux_log (p->h, LOG_CRIT, "fatal error: %s", strerror (errno));
+        flux_log (p->h, LOG_CRIT, "module exiting abnormally");
     }
 
     /* Before processing unhandled requests, ensure that this module
--- a/src/common/librlist/rlist.c
+++ b/src/common/librlist/rlist.c
@@ -330,34 +330,60 @@ struct rnode * rlist_find_host (const struct rlist *rl, const char *host)
     return NULL;
 }
 
-static int rlist_rerank_hostlist (struct rlist *rl, struct hostlist *hl)
+static int rlist_rerank_hostlist (struct rlist *rl,
+                                  struct hostlist *hl,
+                                  rlist_error_t *errp)
 {
     uint32_t rank = 0;
     const char *host = hostlist_first (hl);
     while (host) {
         struct rnode *n = rlist_find_host (rl, host);
-        if (!n)
+        if (!n) {
+            if (errp) {
+                int saved_errno = errno;
+                snprintf (errp->text, sizeof (errp->text),
+                          "Host %s not found in resources",
+                          host);
+                errno = saved_errno;
+            }
             return -1;
+        }
         n->rank = rank++;
         host = hostlist_next (hl);
     }
     return 0;
 }
 
-int rlist_rerank (struct rlist *rl, const char *hosts)
+int rlist_rerank (struct rlist *rl, const char *hosts, rlist_error_t *errp)
 {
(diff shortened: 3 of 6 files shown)


src/cmd/flux-R.c
```diff
@@ -562,7 +562,7 @@ int cmd_rerank (optparse_t *p, int argc, char **argv)
         log_err_exit ("Failed to transform R objects on stdin");
```
chu11, line 562: Seem to be missing a end paren in commit message, probably goes after "rerank operation"?

```
(too many or too few hosts specified for the rerank operation, but the resulting error message from users
can therefore be vague and confusing, e.g. "rlist_rerank: No space left of device".
```

chu11 (approved):
LGTM, just this comment nit

src/cmd/flux-R.c
grondo: Thanks! Fixed the commit message and pushed the result.

garlick (approved):
Looks good!  I misconfigured my system instance to see how the logs turned out and this is what I get from `journalctl -u flux` after the broker fails to start on rank 0:
```
Feb 12 18:38:21 picl0 bash[24554]: resource.err[0]: error reranking R: Number of hosts (4) is less than node count (8)
Feb 12 18:38:21 picl0 bash[24554]: resource.crit[0]: fatal error: Invalid argument
Feb 12 18:38:21 picl0 bash[24554]: broker.err[0]: rc1.0: flux-module: broker.insmod: Invalid argument
Feb 12 18:38:21 picl0 bash[24554]: broker.err[0]: rc1.0: /bin/bash -c /usr/local/etc/flux/rc1 Exited (rc=1) 0.2s
Feb 12 18:38:21 picl0 bash[24554]: broker.info[0]: rc1-fail: init->shutdown 0.200315s
Feb 12 18:38:21 picl0 bash[24554]: broker.info[0]: children-none: shutdown->finalize 0.00039453s
Feb 12 18:38:21 picl0 bash[24554]: broker.info[0]: rc3.0: /bin/bash -c /usr/local/etc/flux/rc3 Exited (rc=0) 0.2s
Feb 12 18:38:21 picl0 bash[24554]: broker.info[0]: rc3-success: finalize->exit 0.173281s
Feb 12 18:38:21 picl0 systemd[1]: flux.service: Main process exited, code=exited, status=1/FAILURE
Feb 12 18:38:21 picl0 systemd[1]: flux.service: Failed with result 'exit-code'.
```

grondo:
> `resource.crit[0]: fatal error: Invalid argument`

I wonder if we should also simplify this generic error on abnormal module termination. It might not be helping to include a generic `fatal error` message along with `strerror (errno)`. In most cases the module will have (hopefully) already printed a more descriptive error of what actually went wrong, and, if not, then it is unlikely the errno will help pinpoint the problem (except perhaps in the exception case of `ENOMEM`).

Perhaps just `module exiting abnormally` or similar?

grondo:
> That works for me.

I went ahead and tacked that on here.

garlick:
Looks like there may be a fluxion test that depends on the current behavior:
```
2022-02-14T18:42:08.1548425Z PASS: t4000-match-params.t 1 - loading resource module with a tiny machine GRUG works
2022-02-14T18:42:08.1552903Z PASS: t4000-match-params.t 2 - loading resource module with an XML works
2022-02-14T18:42:08.1556859Z PASS: t4000-match-params.t 3 - loading resource module with no option works
2022-02-14T18:42:08.1560752Z PASS: t4000-match-params.t 4 - loading resource module with a nonexistent GRUG fails
2022-02-14T18:42:08.1564600Z PASS: t4000-match-params.t 5 - loading resource module with a nonexistent XML fails
2022-02-14T18:42:08.1568812Z FAIL: t4000-match-params.t 6 - loading resource module with incorrect reader fails
2022-02-14T18:42:08.1572710Z PASS: t4000-match-params.t 7 - loading resource module with known policies works
2022-02-14T18:42:08.1576557Z PASS: t4000-match-params.t 8 - loading resource module with unknown policies is tolerated
2022-02-14T18:42:08.1580677Z PASS: t4000-match-params.t 9 - removing resource works
2022-02-14T18:42:08.1584412Z ERROR: t4000-match-params.t - exited with status 1
```

grondo:
Interesting. I'll remove the last commit and save that for a future PR.

grondo:
> Edit: looks like maybe we could just change the test to grep for the new message:

Yeah, that test isn't so good anyway since `Invalid argument` could appear anywhere in the logs and it would pass. I already have a PR open on the flux-sched testsuite, so I'll fix it there and make the test less sensitive to the actual content of the flux-core logging. Once that is fixed I can re-propose the one liner fix I just removed here.

grondo:
Actually, since the fix for flux-sched was merged (thanks @garlick), I pushed the commit again to simplify the module abnormal exit error.

PR #4127, opened 2022-02-12
broker: allow tbon.zmqdebug to be set in config file and make sure it's really off if set to 0

garlick:
This fixes a bug in the way the `tbon.zmqdebug` attribute setting was interpreted, and adds support for configuring it via the TOML config file.


grondo (approved):
LGTM! Only minor suggestion would be to add more context to the commit message for 5ccdbf4 (`testsuite: add tests` -> `testsuite: add tbon.zmqdebug tests`?)

garlick:
Thanks!  sorry about the overly brief commit message. I fixed that, and I'll set MWP.

Issue #4124, opened 2022-02-11
broker: network partition leaves downstream end hanging


garlick:
Problem: if a network cable is unplugged long enough for the upstream node to give up and disconnect, then the network cable is plugged back in, the downstream node never recovers.

Running with `tbon.zmqdebug = 1`, the upstream says the following when the network is plugged back in
```
2022-02-11T01:41:38.400874Z broker.debug[0]: child sockevent tcp://192.168.88.252:8050 accepted                                                                
2022-02-11T01:41:38.416575Z broker.info[0]: overlay auth picl0 OK
2022-02-11T01:41:38.417755Z broker.debug[0]: child sockevent tcp://192.168.88.252:8050 handshake succeeded        
```
and the downstream says
```
2022-02-11T01:41:38.276134Z broker.debug[2]: parent sockevent tcp://picl0:8050 disconnected                                                                    
2022-02-11T01:41:38.276435Z broker.debug[2]: parent sockevent tcp://192.168.88.252:8050 connect retried                                                        
2022-02-11T01:41:38.397151Z broker.debug[2]: parent sockevent tcp://192.168.88.252:8050 connect delayed                                                        
2022-02-11T01:41:38.397757Z broker.debug[2]: parent sockevent tcp://192.168.88.252:8050 connected                                                              
2022-02-11T01:41:38.415294Z broker.debug[2]: parent sockevent tcp://192.168.88.252:8050 handshake succeeded                           
```

So the connection is actually re-established as far as  zeromq is concerned.  However, the upstream broker has given up on the downstream node and removed its flux connection state.  What we expect to happen is the downstream node will send a message, and the upstream node will not show it as connected and return errors, causing downstream to close its connection and exit.  This is not happening.

@grondo suggested offline that maybe the client just isn't saying anything, and I think that is likely correct.

garlick:
However, contrary to that theory, `flux ping 0` on the downstream node not cause rank 0 to disconnect it.  If the ping message had gotten through, rank 0 (upstream) should have taken this code path, but there is nothing in the logs:
```c
        /* New peer is trying to communicate without first saying hello.
         * This likely means that *this broker* restarted without getting a
         * message through to the child, and the child still thinks it is
         * communicating with the same broker (since 0MQ hides reconnects).
         * Send CONTROL_DISCONNECT to force subtree panic.
         */
        else {
            logdrop (ov, OVERLAY_DOWNSTREAM, msg,
                     "didn't say hello, sending disconnect");
[...]

PR #4130, opened 2022-02-14
broker: handle network partition

garlick:
Problem:   If a network cable is unplugged long enough for a TBON child to be disconnected, and then re-plugged, the TBON child does not recover, as noted in #4124.

What happens is this:
- parent times out waiting for TCP ACK, and connection to child is dropped
- child is marked "lost" (now it must restart in order to reconnect)
- child is blocked waiting for connection to parent to be re-established
- when network is plugged back in, zeromq tries to send a message, gets RST which triggers disconnect
- zeromq reconnects
- queued messages are sent
- parent receives queued messages but they come from child marked disconnected
- current code silently discards those messages (resulting in hang)

Solution: in the last step, instead of silently dropping those messages, we log them and send a disconnect control message.  The child then exits and systemd restarts it.

This was tested on little test cluster.

Some changes were also made to log messages to improve the signal to noise ratio in this scenario.

### DIFF
--- a/src/broker/groups.c
+++ b/src/broker/groups.c
@@ -221,9 +221,9 @@ static void batch_apply (struct groups *g)
             char *s = idset_encode (group->members, IDSET_FLAG_RANGE);
             flux_log (g->ctx->h,
                       LOG_DEBUG,
-                      "groups: set %s %s",
+                      "groups: %s=%s",
                       name,
-                      s ? s : "NULL");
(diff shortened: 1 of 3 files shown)


grondo (approved):
LGTM, nice job testing and running this down!

garlick:
Thanks!  Setting MWP.

Issue #3614, opened 2021-04-23
python: proposal that `flux.job.submit` return a `JobID` not `int`


SteVwonder:
Would turn this:
```
compute_jobreq = JobspecV1.from_command(
    command=["./compute.py", "120"], num_tasks=4, num_nodes=2, cores_per_task=2
)
compute_jobreq.cwd = os.path.expanduser("~/flux-workflow-examples/job-submit-api/")
print(JobID(flux.job.submit(f, compute_jobreq)).f58)
```

Into this:
```
compute_jobreq = JobspecV1.from_command(
    command=["./compute.py", "120"], num_tasks=4, num_nodes=2, cores_per_task=2
)
compute_jobreq.cwd = os.path.expanduser("~/flux-workflow-examples/job-submit-api/")
print(flux.job.submit(f, compute_jobreq).f58)
```

Not sure what kind of fallout that would have.

Relevant code to change: https://github.com/flux-framework/flux-core/blob/master/src/bindings/python/flux/job/submit.py#L89

grondo:
This seems reasonable to me

jameshcorbett:
That seems reasonable to me too, and since you're still technically returning an int (a subclass) I don't think it would break anything unless it was a hard type-check ``type(variable) == int``. But if you can convert JobIDs to ints and vice-versa, I also don't see that much would be gained.

PR #4134, opened 2022-02-15
python: return `JobID` from flux.job.submit, not `int`

grondo:
As described in #3614, it would be much more convenient to return a `JobID` instead of `int` from `flux.job.submit` and `flux.job.SubmitFuture.get_id`, with no discernable drawbacks.

This PR updates these Python calls to return a `JobID` object. This should not break any existing code, since `JobID()` on an instance of `JobID` should work fine.

Once the calls are converted, unnecessary use of `JobID()` is removed in `flux-mini.py` and `throughput.py`.

### DIFF
--- a/src/bindings/python/flux/job/submit.py
+++ b/src/bindings/python/flux/job/submit.py
@@ -12,6 +12,7 @@ import errno
 from flux import constants
 from flux.util import check_future_error
 from flux.future import Future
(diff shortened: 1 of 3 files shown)


chu11 (approved):
LGTM

grondo:
Thanks, @chu11!

Issue #4128, opened 2022-02-13
automatically disable/stop queue on instance restart after crash


garlick:
Problem: if the rank 0 broker segfaults, systemd restarts it immediately.  The other ranks then fail and restart also, and the end result looks exactly like a healthy instance restart, except the KVS will have reverted to the last checkpiont, and the system loses track of any running jobs.

It seems like we need a way for the restarted rank 0 broker to detect that it was not shut down properly and prevent the system from starting back up again until a sys admin can intervene.

Related: #4092

grondo:
> It seems like we need a way for the restarted rank 0 broker to detect that it was not shut down properly and prevent the system from starting back up again until a sys admin can intervene.

Are you thinking of something like a "safe mode" (where brokers can start but perhaps queue and some other features are disabled), or preventing the brokers from starting at all (which I think is what we had before a systemd config change)?

If the former, I wonder if there's a good way to notify users and sysadmins that the instance is operating in a safe mode?

garlick:
I was thinking more of a safe mode at the instance level, where  scheduling and job submission would be disabled.

One idea would be to add an eventlog for instance restarts, where events are added for each instance start and stop.  If during startup, the last event posted was a start, safe mode could be entered.   This would need to be in conjunction with #4092 since otherwise the KVS would roll back to the previous clean shutdown.

grondo:
> I wonder if we need another utility to provide at-a-glance status for a flux instance?

Since `broker.starttime` was added in #4076, I was thinking a `flux uptime` command might be a useful addition. In addition to the rank 0 "uptime" perhaps it could provide some other related instance status information?

garlick:
I renamed this issue to focus it on a mechanism for preventing business as usual after an instance crash.   We can open another issue on `flux uptime` or other tooling.

grondo:
Well I guess a trivial implementation would be an rc1 script that enables safe mode if a file exists in a predetermined location, creating the file but not activating safe mode if the file does not exist. 

A paired rc3 script would remove the file on normal shutdown.

This would also allow an administrator to setup safe mode before starting the broker (spiritual equivalent of `/etc/nologin`? The file could even contain an optional message).

PR #4136, opened 2022-02-15
kvs: add date to kvs-primary checkpoint

### LINKED ISSUE #3580: kvs: log date of restored checkpoint
It would be handy if this log message at startup
```
kvs.info[0]: restored kvs-primary from checkpoint
```
included the date the checkpoint was written.

Possibly related to #2783

chu11:
Thought I'd tackle this old issue.  I did a relatively simple thing for the check pointing, writing out a serialized json object storing the rootref + timestamp instead of just the rootref.

Multiple other possibilities, but thought this was the simplest / reasonable (i.e. no need to change content store implementation).  Other options include writing two keys out instead of one.

Note that I intentionally did not support backwards compatibility.  Figured use of this was rare at the moment.

```
Problem: It'd be convenient if we knew the date when the kvs
primary checkpoint was checkpointed.

Solution: When checkpointing the primary KVS, store a json
object with both the rootref and timestamp, instead of just
the rootref string.  On retrieval, parse appropriately and
retrieve timestamp for output in logs.

Fixes #3580
```

### DIFF
--- a/src/modules/kvs/kvs.c
+++ b/src/modules/kvs/kvs.c
@@ -2707,14 +2707,62 @@ static void process_args (struct kvs_ctx *ctx, int ac, char **av)
     }
 }
 
+static int checkpoint_get_version0 (const char *value, char *buf, size_t len)
+{
+    /* if value is a blobref, its verison 0 checkpoint */
+    if (blobref_validate (value) == 0) {
+        if (strlen (value) >= len)
+            return -1;
+        strcpy (buf, value);
+        return 0;
+    }
+    return -1;
+}
+
+static int checkpoint_get_version1 (const char *value,
+                                    char *buf, size_t len,
+                                    double *timestamp)
+{
+    const char *rootref = NULL;
+    json_t *o = NULL;
+    int version;
+    int rv = -1;
+
+    if (!(o = json_loads (value, 0, NULL)))
+        goto error;
+
+    if (json_unpack (o, "{s:i}", "version", &version) < 0)
+        goto error;
+    /* only can handle version 1 right now */
+    if (version == 1) {
+        if (json_unpack (o, "{s:s s:f}",
+                         "rootref", &rootref,
+                         "timestamp", timestamp) < 0)
+            goto error;
+        if (strlen (rootref) >= len)
+            goto error;
+        strcpy (buf, rootref);
+    }
+    else
+        goto error;
+
+    rv = 0;
+error:
+    json_decref (o);
+    return rv;
+}
+
 /* Synchronously get string value by key from checkpoint service.
  * Copy value to buf with '\0' termination.
  * Return 0 on success, -1 on failure,
  */
 static int checkpoint_get (flux_t *h, const char *key, char *buf, size_t len)
 {
-    flux_future_t *f;
-    const char *value;
+    flux_future_t *f = NULL;
+    const char *value = NULL;
+    double timestamp = 0.;
+    char datestr[128];
+    int rv = -1;
 
     if (!(f = flux_rpc_pack (h,
                              "kvs-checkpoint.get",
@@ -2726,24 +2774,59 @@ static int checkpoint_get (flux_t *h, const char *key, char *buf, size_t len)
         return -1;
     if (flux_rpc_get_unpack (f, "{s:s}", "value", &value) < 0)
         goto error;
-    if (strlen (value) >= len) {
+
+    if (checkpoint_get_version1 (value, buf, len, &timestamp) < 0
+        && checkpoint_get_version0 (value, buf, len) < 0) {
         errno = EINVAL;
         goto error;
     }
-    strcpy (buf, value);
-    flux_future_destroy (f);
-    return 0;
+
+    if (timestamp > 0.) {
+        time_t sec = timestamp;
+        struct tm tm;
+        gmtime_r (&sec, &tm);
+        strftime (datestr, sizeof (datestr), "%FT%T", &tm);
+    }
+    else
+        snprintf (datestr, sizeof (datestr), "N/A");
+
+    flux_log (h, LOG_INFO,
+              "restored kvs-primary from checkpoint on %s", datestr);
+    rv = 0;
 error:
     flux_future_destroy (f);
-    return -1;
+    return rv;
 }
 
 /* Synchronously store key-value pair to checkpoint service.
  * Returns 0 on success, -1 on failure.
  */
-static int checkpoint_put (flux_t *h, const char *key, const char *value)
+static int checkpoint_put (flux_t *h, const char *key, const char *rootref)
 {
-    flux_future_t *f;
+    flux_future_t *f = NULL;
+    double timestamp;
+    json_t *o = NULL;
+    char *value = NULL;
+    int save_errno, rv = -1;
+
+    timestamp = flux_reactor_now (flux_get_reactor (h));
+
+    /* version 0 checkpoint
+     * - blobref string only
+     * version 1 checkpoint object
+     * - {"version":1 "rootref":s "timestamp":f}
+     */
+    if (!(o = json_pack ("{s:i s:s s:f}",
+                         "version", 1,
+                         "rootref", rootref,
+                         "timestamp", timestamp))) {
+        errno = ENOMEM;
+        goto error;
+    }
+    if (!(value = json_dumps (o, JSON_COMPACT))) {
+        errno = ENOMEM;
+        goto error;
+    }
 
     if (!(f = flux_rpc_pack (h,
                              "kvs-checkpoint.put",
@@ -2754,13 +2837,17 @@ static int checkpoint_put (flux_t *h, const char *key, const char *value)
                              key,
                              "value",
                              value)))
-        return -1;
-    if (flux_rpc_get (f, NULL) < 0) {
-        flux_future_destroy (f);
-        return -1;
-    }
+        goto error;
+    if (flux_rpc_get (f, NULL) < 0)
+        goto error;
+    rv = 0;
+error:
+    save_errno = errno;
     flux_future_destroy (f);
-    return 0;
+    json_decref (o);
+    free (value);
+    errno = save_errno;
+    return rv;
 }
 
 /* Store initial root in local cache, and flush to content cache
@@ -2855,9 +2942,7 @@ int mod_main (flux_t *h, int argc, char **argv)
         /* Look for a checkpoint and use it if found.
          * Otherwise start the primary root namespace with an empty directory.
          */
-        if (checkpoint_get (h, "kvs-primary", rootref, sizeof (rootref)) == 0)
-            flux_log (h, LOG_INFO, "restored kvs-primary from checkpoint");
-        else {
+        if (checkpoint_get (h, "kvs-primary", rootref, sizeof (rootref)) < 0) {
             if (store_initial_rootdir (ctx, rootref, sizeof (rootref)) < 0) {
                 flux_log_error (h, "storing initial root object");
                 goto done;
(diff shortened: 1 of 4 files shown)


chu11:
hmmmm few builders failing with

```
  not ok 36 - unloaded kvs module
  FAIL: t0017-security.t 36 - unloaded kvs module
  #	
  #		flux module remove kvs
  #	
  not ok 37 - flux content flush not allowed for guest user
  FAIL: t0017-security.t 37 - flux content flush not allowed for guest user
  #	
  #		! FLUX_HANDLE_ROLEMASK=0x2 flux content flush 2>content.flush.err &&
  #		grep -q "Operation not permitted" content.flush.err
  #	
  not ok 38 - flux content load/store allowed for guest user
  FAIL: t0017-security.t 38 - flux content load/store allowed for guest user
  #	
  #		echo Hello >content.store.value &&
  #		FLUX_HANDLE_ROLEMASK=0x2 \
  #		    flux content store <content.store.value >content.store.ref &&
  #		FLUX_HANDLE_ROLEMASK=0x2 \
  #		    flux content load $(cat content.store.ref) >content.load.value &&
  #		test_cmp content.store.value content.load.value
  #	
  # failed 3 among 38 test(s)
```

not entirely sure how what i did could have mucked up these tests, and only on these builders. hmmm

garlick:
I like the idea of json here since we can add other metadata as it comes up without necessarily breaking backwards compatibility.

However, breaking backwards compatibility today will require fluke and elmerfudd to blow away their content.sqlite file and start over.   It might be good to make an effort to avoid that, for example if unpacking the JSON fails, fall back to the old method?   It may also be a good idea to include a version number in the json object to make parsing multiple versions easier going forward.

garlick:
Although.... if were were to have an eventlog in the KVS that's posted to on shutdown, right before the checkpoint is written, we could get the same effect and already have tooling for it.   I was already thinking about something like that in conjunction with #4128.  Can we pause and ponder?

garlick:
Thinking about this a bit more, this seems like a perfectly fine solution to #3580.   The use of an eventlog for tracking intermediate checkpoints or whatever could be a totally separate thing handled at a higher level than this.  (And in fact reading an eventlog within the kvs itself seems like a level violation of some kind.

I think you should carry on but do add a version, and do support restoring from the current checkpoint format if that's not too hard.

chu11:
re-pushed, addressing the comments made by @garlick above

src/modules/kvs/kvs.c
```diff
@@ -2707,14 +2707,50 @@ static void process_args (struct kvs_ctx *ctx, int ac, char **av)
     }
```
garlick, line 2707: Commit message for f80b1a513481c21f53d2b194e33393bebaaa3d9b:

Maybe should read "...when the primary _namespace_ was checkpointed."

and "When checkpointing the primary _namespace_..."

chu11:
doh! 

```
  ./t2010-kvs-snapshot-restore.t: 3: eval: /usr/bin/sqlite3: not found
  ./t2010-kvs-snapshot-restore.t: 3: eval: /usr/bin/sqlite3: not found
  ```

well try and redo test using python sqlite3

File: src/modules/kvs/kvs.c [hunk 2]
```diff
@@ -2707,14 +2707,50 @@ static void process_args (struct kvs_ctx *ctx, int ac, char **av)
...
 }
 
+static int checkpoint_get_version0 (flux_t *h, const char *value,
```
garlick, line 2710: Drop unused `flux_t *` parameter.

src/modules/kvs/kvs.c
```diff
@@ -2707,14 +2707,50 @@ static void process_args (struct kvs_ctx *ctx, int ac, char **av)
...
+{
+    if (strlen (value) >= len) {
+        errno = EINVAL;
+        return -1;
+    }
+    strcpy (buf, value);
+    return 0;
+}
+
+static int checkpoint_get_version1 (flux_t *h, json_t *o,
```
garlick, line 2721: Drop unused `flux_t  *` parameter.

garlick: Suggestion:  take string value not `json_t *` and do the `json_loads()` in this function.
Similarly, don't waste effort on errno here.

src/modules/kvs/kvs.c
```diff
@@ -2726,24 +2762,79 @@ static int checkpoint_get (flux_t *h, const char *key, char *buf, size_t len)
...
+    /* if value is a blobref, its verison 0 checkpoint */
+    if (blobref_validate (value) == 0)
+        return checkpoint_get_version0 (h, value, buf, len);
+
+    if (!(o = json_loads (value, 0, NULL))) {
         errno = EINVAL;
         goto error;
     }
-    strcpy (buf, value);
-    flux_future_destroy (f);
```
garlick, line 2734: With above changes, this can just be
```c
if (checkpoint_get_version1() < 0 && checkpoint_get_version0() < 0 {
    // error interpreting checkpoint object
}
```

src/modules/kvs/kvs.c
```diff
@@ -2726,24 +2762,79 @@ static int checkpoint_get (flux_t *h, const char *key, char *buf, size_t len)
...
+}
+
+static int get_timestamp_now (double *timestamp)
+{
+    struct timespec ts;
+    if (clock_gettime (CLOCK_REALTIME, &ts) < 0)
+        return -1;
+    *timestamp = (1E-9 * ts.tv_nsec) + ts.tv_sec;
+    return 0;
 }
```
garlick, line 2807: Could `flux_reactor_now()` be used instead of adding this function?

src/modules/kvs/kvs.c
```diff
@@ -2757,10 +2848,14 @@ static int checkpoint_put (flux_t *h, const char *key, const char *value)
...
         flux_future_destroy (f);
-        return -1;
+        goto error;
     }
+    rv = 0;
+error:
     flux_future_destroy (f);
-    return 0;
+    json_decref (o);
+    free (value);
```
garlick, line 2857: Potentially clobbers errno.

src/modules/kvs/kvs.c
```diff
@@ -2850,13 +2945,26 @@ int mod_main (flux_t *h, int argc, char **argv)
...
+            time_t sec = timestamp;
+            struct tm tm;
+            if (timestamp > 0.) {
+                gmtime_r (&sec, &tm);
+                strftime (datestr, sizeof (datestr), "%FT%T", &tm);
+            }
+            else
+                snprintf (datestr, sizeof (datestr), "N/A");
+            flux_log (h, LOG_INFO,
+                      "restored kvs-primary from checkpoint on %s", datestr);
```
garlick, line 2966: This should probably be in its own function, if not folded into `checkpoint_get()`.

garlick (changes requested):
Seems like code footprint could be greatly reduced here.

chu11:
re-pushed addressing all of @garlick's comments + fixing the `/usr/bin/sqlite3` CI issue and fixing up some stupid things along the way, including a mem corruption i accidentally introduced that only showed up on CI el8!!

I kept everything as fixups, lemme know if it would be preferred to just squash all the fixups.

garlick:
Thanks!  I think you can squash the fixups, then I'll give it another pass.

chu11:
squashed all the fixups and re-pushed

as an aside, I tried to generate the `change-checkpoint.py` script into the middle of the test to avoid creating yet another helper script.  But there were so many quoting & escaping issues, it became a pain, thus putting it off into a separate script.

src/modules/kvs/kvs.c
garlick: `json_decref()` and `free()` potentially clobber errno.

grondo:
> as an aside, I tried to generate the change-checkpoint.py script into the middle of the test to avoid creating yet another helper script. But there were so many quoting & escaping issues, it became a pain, thus putting it off into a separate script.

Sometimes it is easier to use a here-doc outside of any `test_expect_success()`, then you don't have to deal with the quoting issues. (e.g. see `t0016-cron-faketime.t`)

garlick (approved):
Looking good.  Just noticed one minor errno clobber looks like it's still there.

Couple questions:
- where does the python sqlite3 module come from?  Is there a dependency that configure.ac should be checking for?  (I realize it wasn't introduced in this PR, so just wondering)
- I have another PR going right now that calls `flux_kvs_getroot()` then writes a checkpoint.  I wonder if it would be a good idea to add some non-exported functions to `libkvs` for reuse within flux-core?

I could actually take your functions and turn them into shared ones in my PR if that helps keep things moving along.

chu11:
just did a minor re-push, i realized `flux kvs getroot | jq -r .data[0]` could be replaced with `flux kvs getroot -b` and thus `jq` is no longer required for those tests.  Hopefully didn't clobber anything you were reviewing.

chu11:
> where does the python sqlite3 module come from? Is there a dependency that configure.ac should be checking for? (I realize it wasn't introduced in this PR, so just wondering)

Good question, I initially assumed we didn't have it installed, then to my surprise its actually packaged with the base `python3-libs` package!

> I have another PR going right now that calls flux_kvs_getroot() then writes a checkpoint. I wonder if it would be a ood idea to add some non-exported functions to libkvs for reuse within flux-core?

I had pondered about this too, b/c we checkpoint the guest namespaces in `job-exec`.  Although that checkpointing is a bit of a different beast, was unsure if it would be worthwhile to put all those checkpoints in the same format.

> I could actually take your functions and turn them into shared ones in my PR if that helps keep things moving along.

If you're thinking they'd be useful, perhaps some higher level functions would be worthwhile.  Are you mostly thinking about functions that would create the checkpoint object for users?  and presumably parse said checkpoint object?

chu11:
re-pushed, fixing up the potential `errno` clobber @garlick noticed above.  Just did a `save_errno = errno` kinda thing.

garlick:
What I have at the moment is a `flux kvs flush` command that calls
1. flux_kvs_getroot()
2. RPC to content.flush on rank 0 (flush KVS to backing store)
3. content.checkpoint-put the rootref obtained in 1.

Calling this from cron, for example, would let us recover more data in a crash.    Also, if we append to an eventlog on startup and on exit, we can detect when an instance starts up that was not properly shut down.   But we need to update the checkpoint after the startup append, otherwise the instance always reverts to the last valid shutdown.

chu11:
>    flux_kvs_getroot()
>    RPC to content.flush on rank 0 (flush KVS to backing store)
>    content.checkpoint-put the rootref obtained in 1.

would it worthwhile for `content.checkpoint-get` and `content.checkpoint-put` to be "smarter" than just taking a string and reading/writing it?  Perhaps not for this PR, but could be a refactoring so that it's better in general.  Like handle the version & timestamp for the user?

garlick:
Well, I do think we want to keep the backends fairly "dumb" so any future changes happen in one place instead of three (or more).  However, one nice change we could make would be to use a json "o" as the value, and allow it to be either a json string or a json object, so we don't have to encode the json object as a json string that we put inside a json object :-)

chu11:
> However, one nice change we could make would be to use a json "o" as the value, and allow it to be either a json string or a json object, so we don't have to encode the json object as a json string that we put inside a json object :-)

Ahh that's a good idea.  I was going to do a `flux_rpc_pack ("{s:o}")` in this PR until I realized I couldn't.  Perhaps that could be a follow up PR, and would make your code easier / better too?

garlick:
Maybe the next PR could introduce some reusable (but not public) functions in libkvs and make that protocol change?

chu11:
> Maybe the next PR could introduce some reusable (but not public) functions in libkvs and make that protocol change?

sounds good!, will make an issue after this PR goes in (not sure why the flux-sched build is taking forever)

chu11:
re-kicking the tests, the sched builder got stuck in git clone for some reason

Issue #3064, opened 2020-07-22
Flux on Cray xc40


dongahn:
We have users who want to run OpenMPI on Cray systems at LANL and NERSC. This ticket is opened to track the status of this support. 

Current status:
- @SteVwonder and LANL were able to build flux-core and flux-shed on Trinitite.
-  If you build an MPI with OpenMPI, LANL configuration is such that it adds RPATH to cray PMI library which is not compatible with Flux

dongahn:
From LANL:

>    Yes libpmi.so is rpath'ed in into the executable this is done by the openmpi wrapper. Below I tried to set the LD_PRELOAD to point your libpmi2.so , as it does
have the missing pmi2 symbols , but "alps" sublayer will not initialize within openmpi based on your pmi2 library .
   As a note, i have built up the just released 0.17 of flux-core and 0.9 release of flux-scheduler. and will try them once i can get them pushed to "gadget" .  Will the new version of flux now launch a pmix execuable directly, possibly using the srun provided pmix server ? As this is the other working job launch mechinisim we have on the crays other than the cray pmi2 . If i am not mistaken , openmpi4 is "suppose" to dlopen the library exposed in in the FLUX_PMI_LIBRARY_PATH. I do not believe i have your libraries in my LD_LIBRARY_PATH ( shown below) .  Is there a way to disable the preloading / your flux mini might be doing so it run with its craypmi and pmix libraries intact ? or will the job launch ultimately fail.
 
```console 
mcoyne@nid00027:~> env  | grep LD_LIBRARY
CRAYPAT_LD_LIBRARY_PATH=/opt/cray/pe/gcc-libs:/opt/cray/gcc-libs:/opt/cray/pe/perftools/7.0.6/lib64
[...]

PR #4142, opened 2022-02-17
WIP: libpmi2: implement bits needed for Cray MPI

garlick:
As discussed in #3064, our libpmi2.so was insufficient to support Cray MPI.  This PR collects some changes needed to get it working.


garlick:
Huh, sched failed here.  restarting....
```
2022-02-17T03:19:28.0372307Z FAIL: t1017-rv1-bootstrap.t 4 - rv1-bootstrap: resource idempotency preserved
```

dongahn:
Does that test fail because of a timeout (600 secs)?

grondo:
That's a known intermittent test failure, fyi. 

It would probably be simple to write a test here that exercises the new PMI2 interfaces and get some coverage (sorry if you were already doing that)

dongahn:
I just asked because I think I fixed one potential issue with nondetermism: https://github.com/flux-framework/flux-sched/pull/900/commits/544f327a8aa949552431f26d915e8533e8d8b8ec. Apparently that intermittent problem seems still there.

dongahn:
I wonder if issuing `flux mini submit` immediately after flux module load of `qmanager` could be the source of intermittent error: https://github.com/flux-framework/flux-sched/blob/master/t/t1017-rv1-bootstrap.t#L34. Since the qmanager init could return because it becomes ready to accept a request...

garlick:
Just pushed some tests.  I changed the shell pmi plugin hack so that it stores the `local::` prefix with the key so that "node scope" and "global scope" are separate key namespaces.   No idea if that's important for actual use cases but it sees consistent with the intent.

It's worth being aware that "node scope" is really "shell scope" here.  When there are multiple brokers (and therefore potentially shells) per node, the the procs under two shells on the same node won't see each others node scope keys, yet PMI_process_mapping indicates to the app that they should.  If we build flux-core with Cray MPI in the environment, it's possible we'll run into problems when `make check` runs the MPI tests.

If all goes well I'll drop the WIP and ask @jameshcorbett to try one more test on perlmutter before we call it mergeworthy.

garlick:
I dropped the WIP b/c I think this is in pretty good shape and is ready for review.  @jameshcorbett - if you have a chance to try this on perlmutter and verify one more time that it works (after my recent changes), I'd appreciate it.

I'm not sure what to make of the codecov report.  It seems totally wrong.  I think the coverage ought to be pretty good for the diff, and since I added coverage for some things that weren't covered before, there should be an overall improvement.

jameshcorbett (approved):
LGMT. Test suite came back clean on Perlmutter and I was able to run single-node MPI jobs. I was _not_ able to run multi-node MPI jobs, or even ``--test-size=2 flux mini run -N2`` pseudo-multi-node jobs. I wasn't able to run multi-node jobs with Slurm either so there may be something deeper going on, but here are the logs from the pseudo-multi-node job:

```
22.279s: flux-shell[1]: TRACE: pmi: 1: C: cmd=get kvsname=ƒvwAscb key=local::-bcast-1-0
22.279s: flux-shell[1]: TRACE: pmi: 1: S: cmd=get_result rc=4
23.279s: flux-shell[1]: TRACE: pmi: 1: C: cmd=get kvsname=ƒvwAscb key=local::-bcast-1-0
23.279s: flux-shell[1]: TRACE: pmi: 1: S: cmd=get_result rc=4
24.280s: flux-shell[1]: TRACE: pmi: 1: C: cmd=get kvsname=ƒvwAscb key=local::-bcast-1-0
24.280s: flux-shell[1]: TRACE: pmi: 1: S: cmd=get_result rc=4
25.280s: flux-shell[1]: TRACE: pmi: 1: C: cmd=get kvsname=ƒvwAscb key=local::-bcast-1-0
25.280s: flux-shell[1]: TRACE: pmi: 1: S: cmd=get_result rc=4
```

Repeating seemingly endlessly.

src/common/libpmi/pmi2.c
```diff
@@ -248,10 +254,43 @@ int PMI2_Info_GetSize (int *size)
...
+            sleep (1);
+        result = pmi_simple_client_kvs_get (pmi_global_ctx,
+                                            kvsname,
+                                            local_name,
+                                            value,
+                                            valuelen);
+        if (result != PMI2_ERR_INVALID_KEY && result != PMI2_SUCCESS)
+            return result;
+        tries++;
+    } while (result == PMI2_ERR_INVALID_KEY && waitfor != 0);
```
jameshcorbett, line 288: Probably best to have a limit to this to prevent endlessly looking for an invalid key?

garlick: The reason I did it that way was this comment in argonne's pmi2.h:
> If waitfor is non-zero, the function will never return with found set to zero.

Meh, I agree with you though.  I'll add a 30s timeout.

grondo:
> I'm not sure what to make of the codecov report. It seems totally wrong. I think the coverage ought to be pretty good for the diff, and since I added coverage for some things that weren't covered before, there should be an overall improvement.

Oh, if you are talking about Codecov's claim that this "will **decrease** coverage by 3.22%", that is an issue with our dual coverage reports, one from a normal build and one from our `system` builder on el8. The el8 coverage always results in a ~3% decrease, and for some reason codecov reports this, even though they ostensibly support _merging_ multiple coverage reports together (maybe that doesn't work when there is overlap).

Your diff coverage looks great at >90% so it wouldn't make any sense for overall coverage to drop.

FYI, the non-`system` coverage report is [here](https://codecov.io/github/flux-framework/flux-core/commit/752c221d0b8b7ae853b5b65cb0c00fcd07821c48), and it appears your changes increased coverage by ~+0.4%.

garlick:
Thanks @grondo and @jameshcorbett for your comments.

I added a 30s timeout for `PMI2_Info_GetNodeAttr()` as a separate commit - probably best to leave it separate with explanation since it appears to contradict the canonical pmi2.h comment.  I think it's the right thing though.

jameshcorbett:
FWIW, can confirm that a simple MPI application still works on Perlmutter with the latest commit.

garlick:
I'll set MWP in the morning unless there are further comments.  Thanks!

grondo:
> CI had timed out on on the flux-sched tests, restarting.

Was the hang in `t1011-dynstate-change.t`? I've been seeing that problem in my flux-sched PR and opened issue flux-framework/flux-sched#907

garlick:
Sorry, I couldn't spot what was failing, although maybe I didn't look hard enough because I was fairly sure this PR couldn't have caused a failure over there.

garlick:
EL7 builder failed here:
```
2022-02-18T15:00:13.9697517Z Making all in python
2022-02-18T15:00:13.9787939Z Making all in _flux
2022-02-18T15:00:14.0386819Z   GEN      _rlist_clean.h
2022-02-18T15:00:14.0821567Z   GEN      _security_clean.h
2022-02-18T15:00:14.0910984Z   GEN      _rlist_preproc.h
2022-02-18T15:00:14.0985503Z   GEN      _idset_clean.h
2022-02-18T15:00:14.1111904Z   GEN      _security_preproc.h
2022-02-18T15:00:14.1156227Z   GEN      _hostlist_clean.h
2022-02-18T15:00:14.1160495Z   GEN      _idset_preproc.h
2022-02-18T15:00:14.1667153Z   GEN      _core_clean.h
2022-02-18T15:00:14.1943146Z   GEN      _rlist.c
2022-02-18T15:00:14.1946182Z make[4]: *** [Makefile:1032: _rlist.c] Error 1
2022-02-18T15:00:14.1951021Z make[4]: *** Waiting for unfinished jobs....
```
Not sure why that would be happening all of a sudden.

I rebased on current master and forced a push to kick off testing again (and possibly not require a rebase before merging)

Issue #4146, opened 2022-02-18
add flux-uptime


garlick:
> I wonder if we need another utility to provide at-a-glance status for a flux instance?

Since `broker.starttime` was added in #4076, I was thinking a `flux uptime` command might be a useful addition. In addition to the rank 0 "uptime" perhaps it could provide some other related instance status information?

_Originally posted by @grondo in https://github.com/flux-framework/flux-core/issues/4128#issuecomment-1040602966_

garlick:
While working on stuff related to #4128, I prototyped a little uptime command.  It looks like this:
```
$ flux uptime
 08:51:45 up 5.7m, owner, depth 0, online 16/16
$
```
The info comes from the `broker.starttime`, `security.owner`, and  `instance-level` broker attributes.

Also the `broker.online` broker group.

This was just a quick hack - any thought on how to improve?

garlick:
Good suggestion! 

Maybe instead of "online 16/16" I should just tack on the number of drained nodes if > 0?  (Since offline nodes are automatically drained)?  That'd be in keeping with tacking on queue status if disabled.

PR #4148, opened 2022-02-19
add flux-uptime command

garlick:
OK here's a crack at a `flux-uptime` command per discussion in #4146.  Here's some example output on a single user instance:
```
$ flux uptime
 17:29:27 up 3.7s, owner garlick, depth 0, size 4
```

Here's some output on a system instance, that happens to have a drained node, some offline nodes, and both scheduling and job submission disabled
```
$ flux uptime
 17:31:46 up 46m, owner flux, depth 0, size 8, drained 1, offline 5, submit disabled, sched disabled
```

It gets a little long if everything is showing, but most of the time that won't be the case.

### DIFF
--- a/doc/man1/flux-uptime.rst
+++ b/doc/man1/flux-uptime.rst
@@ -1,0 +1,58 @@
+.. flux-help-description: Tell how long Flux has been up and running
+
+==============
+flux-uptime(1)
+==============
+
+
+SYNOPSIS
+========
+
+**flux** **uptime**
+
+
+DESCRIPTION
+===========
+
+The ``flux-uptime`` displays the following information about the
+current Flux instance, on one or two lines:
+
+- The current wall clock time.
+
+- The elapsed time the Flux instance has been running, in RFC 23 Flux Standard
+  Duration format.  This is derived from the ``broker.starttime`` attribute
+  on the rank 0 broker.
+
+- The Flux instance owner.  On a system instance, this is probably the
+  ``flux`` user.
+
+- The Flux instance depth.  If the Flux instance was not launched as a job
+  within another Flux instance, the depth is zero.
+
+- The instance size.  This is the total number of brokers, which is usually
+  also the number of nodes.
+
+- The number of drained nodes, if greater than zero.  Drained nodes are
+  not eligible to run new jobs, although they may be online, and may currently
+  be running a job.
+
+- The number of offline nodes, if greater than zero.  A node is offline if
+  its broker is not connected to the instance overlay network.
+
+- A notice if job submission is disabled.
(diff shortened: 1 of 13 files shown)


grondo:
Nice!

> It gets a little long if everything is showing, but most of the time that won't be the case.

I wonder if would help if the "optional" queue status strings appear on a separate line?
e.g.

```console
$ flux uptime
 17:31:46 up 46m, owner flux, depth 0, size 8, drained 1, offline 5
 submit disabled
```

This would be more likely to avoid overflowing 80 columns.

Also, one thing I notice from `uptime(1)` is that it uses 2 spaces between "fields" for better separation (at least it appears to). I bet that would improve the output here.

garlick:
Oops, last push broke the uptime sharness test.  Just pushed a fix.

grondo:
> `18:38:01 up 1h,  owner flux,  depth 0,  size 8,  drained 1,  offline 5`
> ` submit disabled,  scheduler disabled`

Very minor, but I think it would read better as `1 drained,  5 offline` instead of `drained 1,  offline 5`

Otherwise, looks great to me!

doc/man1/flux-uptime.rst
```diff
@@ -0,0 +1,58 @@
...
+
+Flux: http://flux-framework.org
+
+RFC 23: Flux Standard Duration: https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_23.html
+
+
+SEE ALSO
+========
+
+:man1:`flux-resource`
```
grondo, line 58: Not critical, but might as well also reference `flux-queue` and `flux-getattr`?

grondo (approved):
LGTM! Just one minor suggestion inline which you could take or leave IMO.

doc/man1/flux-uptime.rst
grondo: Oh, duh. Maybe because `flux-queue(1)` doesn't exist yet. Sorry, you can ignore me...

garlick: There's no man page for `flux-queue` yet.  I started to add one but maybe that's best left for another PR.

garlick:
Thanks @grondo - I forced a push with those changes and I'll go ahead and set MWP.

doc/man1/flux-uptime.rst
grondo: Yeah, I caught myself after I posted that comment. Apologies...

garlick: Oh for some reason I didn't see that even  a half an hour after you posted it.  No worries!

Issue #4144, opened 2022-02-18
content-sqlite / content-files : checkpoint functions should put/get object instead of string


chu11:
Per discussion in https://github.com/flux-framework/flux-core/pull/4136, it would be convenient if the checkpoint functions put/get a json object instead of a string

Issue #4145, opened 2022-02-18
libkvs: checkpoint convenience functions


chu11:
Per discussion in #4136, it may be convenient if some common "kvs checkpointing" functions were put into `libkvs`.  They could be private, need not be public.  Off the top of my head:

```C
json_t *kvs_checkpoint_create (const char *rootref);
kvs_checkpoint_parse_rootref (json_t *o, char *buf, size_t len);
kvs_checkpoint_parse_timestamp (json_t *o, double *timestamp);
```

might work out, with internal setting version / timestamp or getting only if version valid.

Edit: OR possibly wrappers around these to do the rpc calls / gets, TBD

garlick:
I like the RPC idea, something like?
```c
flux_future_t *kvs_checkpoint_update (flux_t *h, const char *name, const char *blobref);
flux_future_t *kvs_checkpoint_lookup (flux_t *h, const char *name);
const char *kvs_checkpoint_lookup_get_blobref (flux_future_t *f);
const char *kvs_checkpoint_lookup_get_formatted_timestamp (flux_future_t *f);
```

PR #3025, opened 2020-06-25
content-s3: add alternate backing store

garrettbslone:
This is a work in progress pr that adds an s3 backing store. It is very light weight and doesn't have much error checking at this point. This pr is based on #2992. 

For testing, I suggest pulling down [Minio](https://www.google.com/search?client=ubuntu&channel=fs&q=minio+quickstart+guide&ie=utf-8&oe=utf-8) and following these steps:

1. Install the s3 library: `apt/yum install libs3-dev`
2. From your flux build directory spin up the Minio server: `minio server ./s3-data &`
3. Set the necessary env vars: 
        `S3_HOSTNAME=127.0.0.1:9000`
        `S3_SECRET_ACCESS_KEY={secretkey}`
        `S3_ACCESS_KEY_ID={accesskey}`
        `S3_BUCKET=flux-bucket`
4. Invoke the content-s3 backing store: `src/cmd/flux start -o,-Scontent.backing-module=content-s3`

TODOs:
1. Add error checking
2. Add TOML config hook
3. Add sharness tests

PR #4149, opened 2022-02-19
 content-{sqlite,files,s3}: refactor checkpoint 

chu11:
Problem: We would like to checkpoint more complex json data structures to the content backing service, but since the content checkpoint services only take string values, it is inconvenient to have to encode/decode all json objects into/from strings.

Solution: Update the checkpoint protocol to instead take / send a json object.  Support backwards compatibility by sending a json string when the data stored is not valid json.  Update all callers accordingly and add additional tests.

Fixes https://github.com/flux-framework/flux-core/issues/4144

Couple of side notes:

- no new tests via s3 content backing, as I'm not entirely sure how to test at the moment.  Perhaps there's some notes somewhere I'm not looking?

- i pondered updating `value` to `data` everywhere, b/c `value` suggests just a string.  But the `sqlite` database uses `value` as the column header, so I kept it as is.


garlick:
> no new tests via s3 content backing, as I'm not entirely sure how to test at the moment. Perhaps there's some notes somewhere I'm not looking?

ci testing for it was added with #3025, although I'm having a bit of trouble unwinding how to run it manually.   At least if you add tests to `t0024-content-s3.t` and push here, it will get run.

From raw logs of `ci / bionic - gcc-8,content-s3,distcheck`

```
2022-02-19T01:44:26.0389891Z         t0024-content-s3.t:  PASS: N=28  PASS=25  FAIL=0 SKIP=3 XPASS=0 XFAIL=0
```

src/modules/content-sqlite/content-sqlite.c
```diff
@@ -374,20 +387,25 @@ void checkpoint_put_cb (flux_t *h,
...
-                             &value) < 0)
+                             &o) < 0)
         goto error;
     if (strlen (key) == 0) {
         errno = EINVAL;
         goto error;
     }
+    if (!(value = json_dumps (o, JSON_ENCODE_ANY))) {
+        errno = EINVAL;
+        goto error;
```
garlick, line 407: We really only need to support reading the version 0 checkpoint, not writing it, so JSON_ENCODE_ANY is not needed in the `json_dumps()` .  While you're there, may as well add JSON_COMPACT.  The strlen check can be dropped too.

src/modules/content-sqlite/content-sqlite.c
```diff
@@ -353,18 +356,28 @@ void checkpoint_get_cb (flux_t *h,
         errno = ENOENT;
         goto error;
     }
+    s = (char *)sqlite3_column_text (ctx->checkpt_get_stmt, 0);
+    if (!(o = json_loads (s, JSON_DECODE_ANY, NULL))) {
+        /* version 0 checkpoint may have been just a blobref string */
```
garlick, line 361: s/may have been just/is/

Could also add a note that version 0 was used prior to flux-core 0.36?

Didn't think of this before, but instead of `json_string(s)` maybe we should just do
```c
o = json_pack ("{s:s s:s s:f}", "version", 0, "value", "rootref", s, "timestamp", 0.);
```
That would simplify the code on the other end.

garlick (commented):
Couple of quick comments inline.

I think we probably only need to support "version 0" in content-sqlite, since I seriously doubt anybody has a running system instance using the other backing stores at this point.

src/modules/content-sqlite/content-sqlite.c
chu11: > o = json_pack ("{s:s s:s s:f}", "version", 0, "value", "rootref", s, "timestamp", 0.);

I'm not following.  Perhaps you messed up the json_pack?  I think there's more args than the format can take.

garlick: Ooops, right - delete "value".  I mean why not construct a consistent "checkpoint object" for version 0 for outside users since it doesn't add any more complexity really.

chu11: oh wait, i think i know what you meant ... 

```
if (!(o = json_loads (...))) {
      /* if data is a version 0 blobref, respond with version 1 format */
      if (blobref_validate (str))
         o = json_pack ("{s:i s:s s:f}", "version", 0, "rootref", s, "timestamp", 0.);
}
```

garlick: The title of this PR might be better for the release notes if it ready "enhance KVS checkpoint protocol."

src/modules/content-sqlite/content-sqlite.c
chu11: ahhh i remember why i did JSON_ENCODE_ANY before, it's b/c tons of tests like in `t0012-content-sqlite` still write general strings instead of objects.  Do we still want to support generically writing strings to checkpoint?  I assume no b/c this is specifically the kvs checkpoint put/get rpc.

chu11:
re-pushed, correctly things from above + more                                                                                               
                                                                                                                                            
- fixed a memleak / possible corruption found by valgrind (`s:O` instead of `s:o`)                                                          
- use JSON_COMPACT instead of JSON_ENCODE_ANY                                                                                               
  - and similarly no need to use JSON_DECODE_ANY                                                                                            
- re-work checkpoint protocol, use `o = json_pack ("{s:s ...}",                                                                             
  "version", 0, ...)` instead of sending raw string when supporting                                                                         
  backwards compatibility.                                                                                                                  
- update tests for changes                                                                                                                  
- add some s3 tests (i hope they work in the CI, presently untested)                                                                        
- remove version 0 support in `content-{files,s3}`                                                                                          
- remove some unnecessary tests

garlick:
Ooops I missed that this was ready for another review.  Do you want to squash the fixups and I'll make another pass?

chu11:
@garlick lemme squash and also update given #4153

chu11:
re-pushed, squashed fixes and fixed up the new `startlog` code from #4153.  Code in `startlog.c` is not the prettiest, figure will be fixed up when #4145 is done (which I'm currently working on)

garlick:
> Code in startlog.c is not the prettiest, figure will be fixed up when https://github.com/flux-framework/flux-core/issues/4145 is done (which I'm currently working on)

Maybe it would be a good idea to combine this and that?  It's an internal function so probably not vital that it appear standalone in release notes.

chu11:
> Maybe it would be a good idea to combine this and that? It's an internal function so probably not vital that it appear standalone in release notes.

Sure, we can do that!

chu11:
re-pushed, adding the new kvs checkpoint helper functions in `libkvs`, and then updating usage in `startlog` and `modules/kvs`.

chu11:
hmmm, hit a concerning build error on `bionic`

```
   expecting success: 
  	today=`date --iso-8601` &&
  	grep checkpoint dmesgfiles.out | grep ${today}
  
  not ok 17 - verify date in flux logs (files)
```

haven't seen this locally yet.  will be adding some debug commits, perhaps `date` outputs something unexpected on this builder.

chu11:
one builder also failed with this in t3200-instance-restart.  I actually saw a few times under a system with heavy load but couldn't reproduce after awhile, thought maybe system was crazy busy for a short burst.  But hmmm.

```
  expecting success: 
  	grep $(cat files_id1.out) files_list.out
  
  not ok 15 - inactive job list contains job from before restart
```

chu11:
Finally got a good piece of debug info from a local run on `fluke`.

`2022-02-26T01:39:38.377231Z kvs.err[0]: kvs_checkpoint_lookup_get_rootref: Invalid argument`

then looked at `content-files/kvs-primary`

`{"version":1,"rootref":"sha1-2b30390bf27cf7bba1222dc032797d8345b2669a","timestamp":1645839598.682987}}`

huh?  why is there an extra curly brace in there?  That would explain the EINVAL.  Some buffer overflow or non-NULL termination?

thought about it and came to following theory, `content-files` originally created files based on the sha1-hash, so effectively all filenames are unique.  So

```
    if ((fd = open (path, O_WRONLY | O_CREAT, 0666)) < 0)
        return -1;
    if (write_all (fd, data, size) < 0) {
        ERRNO_SAFE_WRAP (close, fd);
        return -1;
    }
    if (close (fd) < 0)
        return -1;
```

is ok.  But the `kvs-primary` key, the filename is always the same and the checkpoint can be re-written over and over again into the same file.  This wasn't a problem when we wrote the sha1-hash rootref over and over again, since its the same length everytime.

With the new json object format, the `timestamp` field can be of variable length, thus we're overwriting a prior entry, and if they are of different lengths ... badness.  Thus the possibility of an invalid json object eventually being stored in the `kvs-primary` file.

So I think we need to add a `O_TRUNC` above.  Will try later this weekend.

Whew, this was a tough one.  Hopefully this is correct :P

chu11:
re-pushed, added a comment for adding `O_TRUNC` to the `open()` call and that seemed to fix things.

Before adding `O_TRUNC`, running `make -j16 check` 10 times on catalyst, I got two failures of `not ok 15 - inactive job list contains job from before restart`.

After adding `O_TRUNC`, running `make -j16 check` 25 times on catalyst, I saw no kvs checkpoint related errors (although I did see #4169 a few times).

grondo:
If this is close, don't forget to update the PR title for release notes (and I'll update release notes).

garlick:
Ah O_TRUNC looks like it was the right thing.  Good job.  OK, starting a review pass.

src/modules/content-sqlite/content-sqlite.c
```diff
@@ -353,18 +356,33 @@ void checkpoint_get_cb (flux_t *h,
...
+         * reply with updated object representation.
+         */
+        if (!(o = json_pack ("{s:i s:s s:f}",
+                             "version", 0,
+                             "rootref", s,
+                             "timestamp", 0.))) {
+            errno = EINVAL;
+            goto error;
+        }
+    }
```
garlick, line 371: In the event that data becomes corrupted in the sqlite database, it may be good to have a little tighter error checking here and not return json-encoded garbage to the caller.   Maybe something like
```c
if (!(o = json_loads (s, 0, &error))) {
   if (blobref_validate (s) < 0) {
      errstr = error.text;
      errno = EINVAL;
      goto error;
   }
   /* assume "version 0" if value is a bare blobref and return it in a json envelope */
   ...
}

File: src/modules/content-sqlite/content-sqlite.c [hunk 4]
```diff
@@ -374,16 +392,21 @@ void checkpoint_put_cb (flux_t *h,
...
                              "key",
                              &key,
                              "value",
-                             &value) < 0)
+                             &o) < 0)
         goto error;
+    if (!(value = json_dumps (o, JSON_COMPACT))) {
+        errno = EINVAL;
+        goto error;
+    }
```
garlick, line 409: Might want to set errstr = "failed to encode checkpoint value"; here.

src/modules/kvs/kvs.c
```diff
@@ -2707,62 +2707,18 @@ static void process_args (struct kvs_ctx *ctx, int ac, char **av)
...
-    rv = 0;
-error:
-    json_decref (o);
-    return rv;
-}
-
-/* Synchronously get string value by key from checkpoint service.
- * Copy value to buf with '\0' termination.
+/* Synchronously get checkpoint databy key from checkpoint service.
+ * Copy rootref buf with '\0' termination.
```
garlick, line 2712: s/databy/data by/

src/common/libkvs/kvs_checkpoint_private.h
```diff
@@ -0,0 +1,33 @@
+/************************************************************\
```
garlick, line 1: Since there isn't a conflicting public version of this header like say `message.h` in libflux, maybe this can just be called `kvs_checkpoint.h`?

src/common/libkvs/kvs_checkpoint.c
```diff
@@ -0,0 +1,157 @@
...
+#include <stdlib.h>
+#include <jansson.h>
+#include <flux/core.h>
+#include <time.h>
+
+#include "kvs_checkpoint_private.h"
+
+flux_future_t *kvs_checkpoint_update (flux_t *h,
+                                      const char *key,
+                                      const char *rootref)
```
garlick, line 25: Would `kvs_checkpoint_commit()` sound slightly better?  I dunno, just a thought.

src/modules/content-files/filedb.c
```diff
@@ -87,7 +87,7 @@ int filedb_put (const char *dbpath,
             *errstr = "key name too long for internal buffer";
         return -1;
     }
-    if ((fd = open (path, O_WRONLY | O_CREAT, 0666)) < 0)
+    if ((fd = open (path, O_WRONLY | O_CREAT | O_TRUNC, 0666)) < 0)
```
garlick, line 90: Good find, this is just a straight up bug (and my bad).

garlick (changes requested):
I added a few comments - feel free to push back if you feel I'm off base.

Also: is there a way we could restructure the commits to avoid the history churn within this PR at the call points of the `kvs_checkpoint*()` helpers?  For example, introduce the helpers first but implement the old protocol, then switch users to the helpers, then change protocol?

chu11:
> Also: is there a way we could restructure the commits to avoid the                                                                        
> history churn within this PR at the call points of the                                                                                    
> kvs_checkpoint*() helpers? For example, introduce the helpers first                                                                       
> but implement the old protocol, then switch users to the helpers, then                                                                    
> change protocol?                                                                                                                          
                                                                                                                                            
Hmmm, not too easily, as I'd be changing a lot of code, sort of starting from scratch.  Unfortunately started this before your                                                                              
`startlog` PR, so doing it the other way around wasn't as obvious at the time.

chu11:
Re-pushed addressing all of the comments above.  Went ahead and squashed since everything is tiny changes.

garlick (approved):
Thanks for all the changes, looking good!
I just had a couple more suggestions for cleanup, but they're fairly trivial so I'll tentatively mark approved.

src/common/libkvs/kvs_checkpoint.c
```diff
@@ -0,0 +1,157 @@
...
+    if (!(f = flux_rpc_pack (h,
+                             "kvs-checkpoint.put",
+                             0,
+                             0,
+                             "{s:s s:O}",
+                             "key",
+                             key,
+                             "value",
+                             o)))
+        goto error;
```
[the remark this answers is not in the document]
chu11: oh duh, that's obvious now.  I think I could do in a few other places too.

chu11:
re-pushed addressing the comments above.  @garlick wanna do one more quick skim before I set MWP?

chu11:
hmmm, hit one build error that i'm 99% sure is not releated to this PR.  restarted builders.

```
  expecting success: 
          ${FLUX_BUILD_DIR}/t/rexec/rexec -r 1 sleep 100 &
          pid1=$!
          ${FLUX_BUILD_DIR}/t/rexec/rexec -r 1 sleep 100 &
          pid2=$!
  	sleep 1 &&
          ${FLUX_BUILD_DIR}/t/rexec/rexec_ps -r 1 > output &&
          count=`cat output | wc -l` &&
          test "$count" = "2" &&
  	sleep 1 &&
  	kill -TERM $pid1 &&
  	kill -TERM $pid2 &&
          ${FLUX_BUILD_DIR}/t/rexec/rexec_ps -r 1 > output &&
          count=`cat output | wc -l` &&
          test "$count" = "0"
  
  not ok 27 - disconnect terminates all running processes
```

grondo:
We've been seeing that one lately: #4097

src/common/libkvs/kvs_checkpoint.c
```diff
@@ -0,0 +1,132 @@
...
+
+/* returns "N/A" if not available */
+int kvs_checkpoint_lookup_get_formatted_timestamp (flux_future_t *f,
+                                                   char *buf,
+                                                   size_t len)
+{
+    int version;
+    double timestamp;
+
+    if (!f || !buf || len <= 0) {
```
garlick, line 102: I don't think len can be less than zero as `size_t` is unsigned.

src/common/libkvs/kvs_checkpoint.c
```diff
@@ -0,0 +1,132 @@
...
+    return 0;
+}
+
+/* returns "N/A" if not available */
+int kvs_checkpoint_lookup_get_formatted_timestamp (flux_future_t *f,
+                                                   char *buf,
+                                                   size_t len)
+{
+    int version;
+    double timestamp;
```
garlick, line 100: I would initialize timestamp to 0. for good measure since it's conditionally assigned.  It's also conditionally used, but matching those two conditions up might be too much to ask of some compilers.

garlick (approved):
Looks good!  This time just a couple of minor things.  If ci isn't complaining, no biggie I guess.

src/common/libkvs/kvs_checkpoint.c
garlick: Oh, actually, probably the right thing is to ignore `len` since zero isn't the only value that's too short.  Instead, check the return value of `snprintf()` and `strftime()` to catch any overflow.

chu11:
just re-pushed, just cleaned up the two things @garlick noted above

```
diff --git a/src/common/libkvs/kvs_checkpoint.c b/src/common/libkvs/kvs_checkpoint.c
index abc7567..78e640d 100644
--- a/src/common/libkvs/kvs_checkpoint.c
+++ b/src/common/libkvs/kvs_checkpoint.c
@@ -97,9 +97,9 @@ int kvs_checkpoint_lookup_get_formatted_timestamp (flux_future_t *f,
                                                    size_t len)
 {
     int version;
-    double timestamp;
+    double timestamp = 0.;
 
-    if (!f || !buf || len <= 0) {
+    if (!f || !buf) {
         errno = EINVAL;
         return -1;
     }
@@ -119,11 +119,17 @@ int kvs_checkpoint_lookup_get_formatted_timestamp (flux_future_t *f,
         time_t sec = timestamp;
         struct tm tm;
         gmtime_r (&sec, &tm);
-        strftime (buf, len, "%FT%T", &tm);
+        if (strftime (buf, len, "%FT%T", &tm) == 0) {
+            errno = EINVAL;
+            return -1;
+        }
+    }
+    else { /* version == 0 */
+        if (snprintf (buf, len, "N/A") >= len) {
+            errno = EINVAL;
+            return -1;
+        }
     }
-    else /* version == 0 */
-        snprintf (buf, len, "N/A");
-
     return 0;
 }
```

garlick:
Ready for MWP on this one?

chu11:
yup! thanks for review.

chu11:
hmm, builder failed with 

```
ERROR: t2402-job-exec-dummy.t - exited with status 141 (terminated by signal 13?)
```

All tests passed in `t2402-job-exec-dummy`, so this is on shutdown.  13 = SIGPIPE, which I'm not entirely sure how that could happen.  Perhaps something slow going on during shutdown?

chu11:
@mergifyio refresh

PR #4153, opened 2022-02-23
add flux-startlog and enter safe mode after crash

garlick:
This adds a  `flux-startlog` utility that operates on `admin.eventlog`.

The rc1 adds a `start` event to the eventlog, and rc3 adds a `finish` event.  If the last two events are both starts, then flux is restarting after a crash, since it means rc3 didn't execute.   In that case, run `flux queue stop` and `flux queue disable` (later in rc1) to avoid further damage.

The command provides a history of flux restarts similar to last(1):
```
$ sudo flux startlog
2022-02-23 19:08 - running (58s)
$ sudo systemctl restart flux
$ sudo flux startlog
2022-02-23 19:08 - 2022-02-23 19:10 (1.3m)
2022-02-23 19:10 - running (2.6s)
$ sudo kill -9 89249  # kill the broker
$ sudo systemctl start flux
$ sudo flux startlog
2022-02-23 19:08 - 2022-02-23 19:10 (1.3m)
2022-02-23 19:10 - crashed
2022-02-23 19:10 - running (21s)
$ flux uptime
 11:12:37 up 1.9m,  owner flux,  depth 0,  size 8,  4 offline
  submit disabled,  scheduler disabled
$ flux queue status
flux-queue: Job submission is disabled: Flux is in safe mode due to an incomplete shutdown.
flux-queue: Scheduling is disabled
```
Marking as WIP since feedback on the interface would be useful before drilling down on remaining test coverage.  Also, oops - I need to print time in the local timezone.

### DIFF
--- a/doc/man1/flux-startlog.rst
+++ b/doc/man1/flux-startlog.rst
@@ -1,0 +1,64 @@
+.. flux-help-description: Show Flux instance start and stop times
+
+================
+flux-startlog(1)
+================
+
+
+SYNOPSIS
+========
+
+**flux** **startlog**
+
+
+DESCRIPTION
+===========
+
+List the Flux instance's start and stop times, by interpreting the contents
+of the KVS ``admin.eventlog``.
+
+A ``start`` event is posted to the eventlog at startup, and a ``finish`` event
+is posted to the eventlog at finalization.  The timestamps on the two events
+are used to calculate the instance run time, which is shown in the listing
+in Flux Standard Duration format.
+
+If the current ``start`` event is not immediately preceded by a ``finish``
+event (unless it is the first entry in the eventlog), then the Flux instance
+may have crashed and data may have been lost.  If this is detected on instance
+startup, Flux is placed into a safe mode where scheduling and job submission
+are disabled, to avoid further damage pending recovery.
+
+This command is not available to guest users.
+
+
+OPTIONS
+=======
+
+**-h, --help**
+   Summarize available options.
+
+**--check**
+   If the instance has most recently restarted from a crash, exit with a
+   return code of 1, otherwise 0.
+
+**--quiet**
+   Suppress non-error output.
+
+**-v, --show-version**
+   Show the flux-core software version associated with each start event.
+
+
+RESOURCES
+=========
+
+Flux: http://flux-framework.org
+
+RFC 18: KVS Event Log Format: https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_18.html
+
+RFC 23: Flux Standard Duration: https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_23.html
+
+
+SEE ALSO
+========
+
+:man1:`flux-uptime`, :man1:`flux-kvs`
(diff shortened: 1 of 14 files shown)


garlick:
Dropping the WIP after:
- add flux-core version to the eventlog (so we know what release crashed)
- slightly better test coverage
- address LGTM warning (valid!)
- add man page

etc/rc1
```diff
@@ -58,9 +58,10 @@ if test $RANK -eq 0 -a "${FLUX_SCHED_MODULE}" != "none" \
     flux module load ${FLUX_SCHED_MODULE:-sched-simple}
 fi
 
-test $RANK -ne 0 -o "${FLUX_INSTANCE_RESTART}" = "t" \
```
grondo, line 57: typo in commit message for c5b89ed: `it's` should be `its`

grondo: Also :+1: on this change!

grondo (approved):
Wow, this is pretty neat and seems useful!

I guess it doesn't catch a problem that happens between when the `finish` event is posted and `rc3` eventually does exit, but I guess that doesn't matter because if a problem does occur at that point, the KVS data probably won't make it to the content store anyway, and the problem will indeed be detected on the next startup!

etc/rc1
garlick: Pushed a fix for that and a dictionary addition for el7.

Also, just wanted to note that a KVS checkpoint is forced out after the start event but not the finish event, so the finish event won't likely be commited to the backing store until the kvs module is unloaded.

src/cmd/builtin/uptime.c
```diff
@@ -219,8 +219,7 @@ static void append_count_if (char *buf,
         int n = strlen (buf);
         snprintf (buf + n,
```
chu11, line 220: commit message, minor english, "a comma and space", not plural spaces

doc/man1/flux-startlog.rst
```diff
@@ -0,0 +1,64 @@
...
+List the Flux instance's start and stop times, by interpreting the contents
+of the KVS ``admin.eventlog``.
+
+A ``start`` event is posted to the eventlog at startup, and a ``finish`` event
+is posted to the eventlog at finalization.  The timestamps on the two events
+are used to calculate the instance run time, which is shown in the listing
+in Flux Standard Duration format.
+
+If the current ``start`` event is not the first entry in the eventlog and
+is not immediately preceded by a ``finish`` event, then the Flux instance
```
chu11, line 26: I had to read this several times to grok, perhaps instead: 

`If the current ``start`` event is not immediately preceded by a ``finish`` event (unless it is the first entry in the eventlog), then ...`

basically highlighting the common case and noting the uncommon case vs. listing the uncommon case first.

chu11 (approved):
LGTM

note to myself, #4149 will have to be updated to update this as well.

src/cmd/builtin/uptime.c
garlick: It's actually one comma and two spaces.

garlick:
Force pushed the man page change suggested by @chu11, and also corrected a typo I noticed in a commit message.  I'll go ahead and set MWP.  Thanks!

PR #4157, opened 2022-02-24
libflux: add flux_hostmap_lookup(3)

grondo:
This PR abstracts the functionality provided by `flux overlay lookup` into a function `flux_hostmap_lookup(3)` exported from libflux-core for convenience.

The function can take either a set of ranks in Idset form or a set of hostnames in Hostlist form and will return the reverse as an allocated string.

I didn't add docs since there presently any for `flux_get_hostbyrank(3)` and `flux_get_rankbyhost(3)`, but I could if that is desired. (I'm guessing `flux_hostmap_lookup(3)` will be slightly less useful than the other two, but there is one use case in #4155, so it made sense to avoid the code duplication)

### DIFF
--- a/src/common/libflux/attr.c
+++ b/src/common/libflux/attr.c
@@ -17,6 +17,8 @@
 
 #include "src/common/libczmqcontainers/czmq_containers.h"
 #include "src/common/libhostlist/hostlist.h"
+#include "src/common/libidset/idset.h"
+#include "src/common/libutil/errprintf.h"
 
 #include "attr.h"
 #include "rpc.h"
@@ -188,24 +190,33 @@ int flux_get_rank (flux_t *h, uint32_t *rank)
     return 0;
 }
 
-const char *flux_get_hostbyrank (flux_t *h, uint32_t rank)
+static struct hostlist * get_hostlist (flux_t *h)
 {
     struct attr_cache *c;
+
+    if (!(c = get_attr_cache (h)))
+        return NULL;
+    if (!c->hostlist) {
+        const char *val;
+        if (!(val = flux_attr_get (h, "hostlist"))
+            || !(c->hostlist = hostlist_decode (val)))
+            return NULL;
+    }
+    return c->hostlist;
+}
+
+const char *flux_get_hostbyrank (flux_t *h, uint32_t rank)
+{
+    struct hostlist *hl;
     const char *result;
 
     if (rank == FLUX_NODEID_ANY)
         return "any";
     if (rank == FLUX_NODEID_UPSTREAM)
         return "upstream";
-    if (!(c = get_attr_cache (h)))
+    if (!(hl = get_hostlist (h)))
         goto error;
-    if (!c->hostlist) {
-        const char *val;
-        if (!(val = flux_attr_get (h, "hostlist"))
-            || !(c->hostlist = hostlist_decode (val)))
-            goto error;
-    }
-    if (!(result = hostlist_nth (c->hostlist, rank)))
+    if (!(result = hostlist_nth (hl, rank)))
         goto error;
     return result;
 error:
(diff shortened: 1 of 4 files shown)


grondo:
Probably will be a bit difficult to get higher coverage on this one, since there are so many error cases in the new function that can't easily be tested.

src/common/libflux/attr.c
```diff
@@ -17,6 +17,7 @@
```
garlick, line 17: Commit message for 79de48c309af39342080b63437f4affbe8e59bf1
s/fucntion/function/

src/common/libflux/attr.h
```diff
@@ -72,6 +72,20 @@ const char *flux_get_hostbyrank (flux_t *h, uint32_t rank);
  */
 int flux_get_rankbyhost (flux_t *h, const char *host);
 
+typedef struct {
+    char text[128];
+} flux_lookup_error_t;
+
```
garlick, line 78: Not sure if this is a great idea but I'll suggest it anyway - should we just call this `flux_error_t` and define it in <libflux/types.h> along with `flux_free_f`?  Then maybe we could stop proliferating similar typedefs like
```c
typedef struct {
    char filename[80];  // if unused, will be set to empty string
    int lineno;         // if unused, will be set to -1
    char errbuf[160];
} flux_conf_error_t;

typedef struct {
    char text[128];
} rlist_error_t;

typedef struct jobtap_error {
    char text [128];
} jobtap_error_t;

typedef struct {
    char text[160];
} job_list_error_t;

typedef struct {
    char text[160];
} flux_jobspec1_error_t;
```
Could also maybe define the useful `errprintf()` function.

src/common/libflux/attr.h
```diff
@@ -72,6 +72,20 @@ const char *flux_get_hostbyrank (flux_t *h, uint32_t rank);
...
+/* Return a list of hosts/ranks in hostlist/idset form given 'targets'
+ *  in idset/hostlist form. Caller must free returned value.
+ *
+ * Returns NULL on failure with error message in errp->text if errp
+ *   is non-NULL.
+ */
+char *flux_hostmap_lookup (flux_t *h,
+                           const char *targets,
+                           flux_lookup_error_t *errp);
```
garlick, line 87: It may be good to also point out
- the source of the mapping is the broker rank-ordered broker `hostlist` attribute
- idset is a set (unordered, no duplicates), while hostlist is a list (ordered, may contain duplicates)
- if there are multiple ranks per host, this function only maps a host to the first rank on it

garlick (approved):
Great!  I had a couple of suggestions for you to consider but nothing major.

src/common/libflux/attr.h
grondo: I had a similar thought, but didn't go there here (esp when it was tied up in another pr).

In general a good idea I think, I'll give it a try here and see how it goes.

grondo: Oh, I remember the other thought I was having -- this will break existing users. Maybe better to do now than later though, but I'm not sure of the exact fallout...

garlick: Maybe just name this with the general name and we can convert other APIs later?

garlick: Of course I let myself off the hook when working on other PRs that did similar things, so it's a bit rude of me to call it out here :-)

grondo: I could add the general name, and all existing cases could be typedef'd except for `flux_conf_error_t` which uses `errbuf` instead of `text` for the error text.

grondo:
rebased on #4165 and removed `flux_lookup_error_t` in favor of `flux_error_t`.

Also improved comment on `flux_hostmap_lookup(3)` as suggested by @garlick.

garlick:
Great!  Already approved, so freel free to set MWP aftter #4165 goes in.

PR #4159, opened 2022-02-24
shell: fix delay in completion of jobs with a single shell rank

grondo:
Commit 1d08bf76fd773fedb4a1956a500760254ca27fdf modified how the shell output plugin detects completion, but it introduced a performance issue for single rank jobs, which now wait at least `output.batch-timeout` seconds before the output plugin flushes pending events and exits the reactor.

This PR fixes that issue by having the rank 0 shell add a reference count per task - so that once all tasks are complete the output eventlogger can be flushed and exit immediately.

Without this fix single rank jobs have a minimum runtime of ~500 ms, which affects single node job throughput.

### DIFF
--- a/src/shell/output.c
+++ b/src/shell/output.c
@@ -1123,6 +1123,7 @@ struct shell_output *shell_output_create (flux_shell_t *shell)
     if (!(out->pending_writes = zlist_new ()))
         goto error;
     if (shell->info->shell_rank == 0) {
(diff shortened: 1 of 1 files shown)


garlick (approved):
LGTM!  Good find!

grondo:
Rebased. Thanks! I'll set MWP.

PR #4162, opened 2022-02-25
libflux/future: set missing errno in flux_future_wait_for() 

### LINKED ISSUE #4160: libflux: flux_future_wait_for does not set errno on unfulfilled future that fails to initialize
Had a simple unit test for #4145 to make sure an unfulfilled future failed correctly, and to my surprise `flux_future_wait_for()` does not set errno when the future is unfulfilled and a timeout is not set.  I think `EAGAIN` is probably the right errno.  Manpage should be updated too.

chu11:
Problem: When a negative timeout is passed to flux_future_wait_for()
and a future is not properly initialized to be tied to a localized
'now' reactor, flux_future_wait_for() can return -1 without setting
an errno.

Set errno of EINVAL when the above corner case occurs.  Add unit
test.

Fixes #4160

### DIFF
--- a/src/common/libflux/future.c
+++ b/src/common/libflux/future.c
@@ -480,8 +480,14 @@ int flux_future_wait_for (flux_future_t *f, double timeout)
             flux_dispatch_requeue (f->now->h);
         f->now->running = false;
     }
-    if (!future_is_ready (f))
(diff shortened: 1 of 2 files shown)


grondo (approved):
Seems good to me. Nice find BTW!

PR #4165, opened 2022-02-25
consolidate multiple *_error_t structures into a common flux_error_t

grondo:
As suggested in #4157, replace many instances of `<class>_error_t` with a common `flux_error_t`.

Add a common `errprintf()` to libutil and replace duplicate definitions of this convenience function in `librlist`, `libjobspec1` and `jobtap`.

In `libjobspec1`, use a typedef to avoid breaking existing users.

This PR doesn't touch `flux_conf_error_t` since its structure contains `file` and `line` members which would require `errprintf()` to contain parameters for these members unnecessarily in the other modules where they are unused. (it also uses `error.errbuf` instead of `error.text` so would unconditionally break existing callers)


src/common/libflux/types.h
```diff
@@ -17,6 +17,13 @@ extern "C" {
 
 typedef void (*flux_free_f)(void *arg);
```
garlick, line 18: commit message:
s/somethign/something/

src/common/libutil/Makefile.am
```diff
@@ -90,7 +90,9 @@ libutil_la_SOURCES = \
 	jpath.c \
 	jpath.h \
 	uri.c \
```
garlick, line 92: commit message:
missing "and" in problem statement?

garlick (approved):
Nice cleanup!
Couple commit message typos to fix but feel free to set MWP after that.

chu11 (approved):
LGTM too

grondo:
Thanks @chu11 and @garlick! I've set MWP.

PR #4166, opened 2022-02-25
add flux-config get

garlick:
Here's a new command and corresponding RPC:
```
$ flux config get -h
Usage: flux config get [--type=TYPE] [--quiet] [PATH]
Query broker configuration values
  -t, --type=TYPE        Set expected type (any, string, int, real, boolean,
                         object, array, fsd, fsd-int, fsd-real)
  -q, --quiet            Suppress printing of "[key] is not set" errors.
```

Parking this as a WIP until I have a chance to write tests and man page.  Meanwhile, how's the usage look?  I was mainly intending to support usage from shell scripts like rc1.

### DIFF
--- a/src/cmd/builtin/config.c
+++ b/src/cmd/builtin/config.c
@@ -8,6 +8,10 @@
  * SPDX-License-Identifier: LGPL-3.0
 \************************************************************/
 
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+
 #include "builtin.h"
 
 static int internal_config_reload (optparse_t *p, int ac, char *av[])
--- a/src/cmd/builtin/config.c
+++ b/src/cmd/builtin/config.c
@@ -11,8 +11,147 @@
 #if HAVE_CONFIG_H
 #include "config.h"
 #endif
+#include <jansson.h>
+
+#include "src/common/libutil/jpath.h"
+#include "src/common/libutil/fsd.h"
 
 #include "builtin.h"
+
+typedef enum {
+    FSD_NONE,
+    FSD_INTEGER,
+    FSD_REAL,
+    FSD_STRING,
+} fsd_subtype_t;
+
+struct map {
+    const char *s;
+    json_type type;
+    fsd_subtype_t fsd_subtype;
+};
+
+static struct map typemap[] = {
+    { "object", JSON_OBJECT, FSD_NONE },
+    { "array", JSON_ARRAY, FSD_NONE },
+    { "string", JSON_STRING, FSD_NONE },
+    { "integer", JSON_INTEGER, FSD_NONE },
+    { "real", JSON_REAL, FSD_NONE },
+    { "boolean", JSON_TRUE, FSD_NONE }, // special case
+    { "any", JSON_NULL, FSD_NONE },     // special case
+    { "fsd", JSON_STRING, FSD_STRING },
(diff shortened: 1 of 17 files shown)


grondo:
Usage looks good to me. I guess the common case is to check for a configured value, but a value may not match or it may be not set, which might make script-based usage a little awkward (though maybe I'm missing some obvious trick). 

Would it be to kludgy to add a `--default=VALUE` option, in which case `flux config get` would print `VALUE` (converted via `--type` option) if the config `PATH` is not set? Then testing for e.g. `exec.use-systemd` would just be
```sh
if test $(flux config get --default=false exec.use-systemd) != "true"; then
   ...
fi
```
Another option would be to add a corresponding `flux config test`, though this might be a little tricky to allow anything other than a strict comparison since the flux-config command is currently in C.

garlick:
I went ahead and added `--default=VAL`.  Great idea.

A test subcommand would be neat, but I was figuring test(1) probably had us covered for basic stuff now - maybe good enough for now?

Sharness test coming soon.

grondo:
> A test subcommand would be neat, but I was figuring test(1) probably had us covered for basic stuff now - maybe good enough for now?

Yeah, with the addition of `--default=VAL` I don't think `flux config test` is required at this time.

src/cmd/builtin/config.c
```diff
@@ -11,9 +11,148 @@
...
+    }
+    if (!(h = builtin_get_flux_handle (p)))
+        log_err_exit ("flux_open");
+    if (!(f = flux_rpc (h, "config.get", NULL, FLUX_NODEID_ANY, 0))
+        || flux_rpc_get_unpack (f, "o", &o) < 0)
+        log_msg_exit ("Error fetching config object: %s",
+                      future_strerror (f, errno));
+    typestr = optparse_get_str (p, "type", "any");
+    if (parse_json_type (typestr, &type, &fsd_subtype) < 0)
+        log_msg_exit ("Unknown type");
```
grondo, line 143: Minor suggestion: `log_msg_exit ("Unknown type: %s", typestr)`

Also, could add a test that ensures a bad type provided here fails with the expected error.

grondo (approved):
LGTM! One pretty minor suggestion inline.

It would also be nice to have a command line interface to the "builtin" configuration values (i.e. `flux_conf_builtin_get(3)` - I wonder if this could be a `--builtin` flag on `flux config get` or if it should be a separate command.

This is only a nice to have (mostly for testing at this point), and we can use a small Python script to implement it, so not necessary to add to this PR.

src/cmd/builtin/config.c
garlick: > It would also be nice to have a command line interface to the "builtin" configuration values (i.e. `flux_conf_builtin_get(3)` - I wonder if this could be a `--builtin` flag on `flux config get` or if it should be a separate command.

Sure, good idea.  I'll add that too.

garlick:
Just forced a push, adding a `flux config builtin` subcommand, and the suggested error message fix + test.

I probably went too far trying to document all the builtin keys, but we can always move that to another place later as we fill out our docs.

grondo:
Great! We can remove a [custom script](https://github.com/flux-framework/flux-sched/blob/master/t/scripts/flux-builtin-conf.py) from the flux-sched testsuite once this goes in.

garlick:
Thanks, I'll go ahead with MWP.

PR #4170, opened 2022-02-28
drop unused project meta files

### LINKED ISSUE #4135: remove unused project meta-files
Problem: recently we committed cleanups to some other flux-framework projects to set automake foreign mode and remove unused project meta files like ChangeLog and README, but somehow we missed `flux-core`.

While we're at it, I don't think we need the symlink from NOTICE -> NOTICE.LLNS.

Also probably INSTALL could be removed, since we cover that in the README.md.

garlick:
Problem: automake "foreign mode" was enabled in a previous commit,
but unused GNU project-specific meta files that were required before
that were not cleaned up.

Drop ChangeLog and README which have no content.
Drop the NOTICE link to NOTICE.llns as it is not helping much.
Drop INSTALL as Flux-specific installation instructions are
in the README.md and in readthedocs.

Fixes #4135


grondo (approved):
Great! LGTM.

PR #4172, opened 2022-03-01
flux-uptime: provide useful output for slow/stuck broker state

garlick:
Problem: my system instance was taking a long time to get through `rc1`, but some tools fail in that state and others hang on rank 0.  I thought it might be useful for `flux-uptime` to report how long the broker has been in the current state if it's not in _run_ state.  Here's an example:
```
 garlick@picl0:~/proj/flux-conf$ flux uptime
 13:03:07 init 6.2s,  owner flux,  depth 0,  size 8
```
Meanwhile, a compute node reports:
```
 garlick@picl2:~$ flux jobs
flux-jobs: ERROR: [Errno 11] Upstream Flux broker is offline. Try again later.
 garlick@picl2:~$ flux uptime
 12:53:51 join 21m,  owner flux,  depth 0,  size 8
```
After rank 0 finally gets to run state, it looks like before, except it says _run_ instead of _up_:
```
 garlick@picl0:~/proj/flux-conf$ flux uptime
 13:06:21 run 3.3m,  owner flux,  depth 0,  size 8,  2 offline

I updated the flux-uptime(1) man page with a description of the broker states that might appear in the output, FWIW.

### DIFF
--- a/src/cmd/builtin/uptime.c
+++ b/src/cmd/builtin/uptime.c
@@ -229,20 +229,42 @@ static void append_count_if (char *buf,
  */
 static void default_summary (flux_t *h)
 {
-    double t_start = attr_get_starttime (h);
     double t_now = flux_reactor_now (flux_get_reactor (h));
     int userid = attr_get_int (h, "security.owner");
     int size = attr_get_int (h, "size");
     int level = attr_get_int (h, "instance-level");
-    int drained = resource_status_drained (h);
-    int online = groups_get_count (h, "broker.online");
-    int offline = size - online;
+    int drained = 0;
+    int offline = 0;
+    bool submit_is_disabled = false;
+    bool sched_is_disabled = false;
+    flux_future_t *f;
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/cmd/builtin/uptime.c [hunk 1]
```diff
@@ -229,29 +229,52 @@ static void append_count_if (char *buf,
  */
```
chu11, line 229: commit message typo

```Problem: if the broker gets stuck in rc1, cleanup, or rc3,
it May not be evident what is going on without examining logs.
```

lower case `M` in `May`.

chu11 (approved):
LGTM, just noticed a commit message typo.

grondo (approved):
Nice! LGTM, too. This is a pretty useful tool now!

garlick:
Thanks!  I fixed that typo and also made another editing pass on the broker state descriptions in the man page, to make them more useful from the perspective of someone debugging a hung system.

PR #4173, opened 2022-03-01
[WIP] libflux: support flux_watcher_is_actve()

### LINKED ISSUE #4156: libflux: add flux_watcher_is_active()
A few times I've been hit by bugs / issues where it would be convenient to know if a watcher is presently started / stopped.  So I end up setting flags to debug the situation.

It would be convenient to add this to `libflux/reactor` (perhaps in a private api?).  Something like:

`bool flux_watcher_is_started (flux_watcher_t *w);`

internally a maintained flag set / unset when the watcher is started / stopped is probably all that is needed.

chu11:
Problem: Presently there is no way to know if a watcher has
been started or stopped.  This can cause difficulty debugging
some situations in which watchers are regularly started / stopped.

Solution: Add a flux_watcher_is_active() function to determine the
start / stop state of a watcher.

Fixes #4156

### DIFF
--- a/src/common/libflux/reactor.c
+++ b/src/common/libflux/reactor.c
@@ -206,17 +206,28 @@ struct flux_watcher_ops * flux_watcher_get_ops (flux_watcher_t *w)
 void flux_watcher_start (flux_watcher_t *w)
 {
     if (w) {
-        if (w->ops->start)
+        if (w->ops->start) {
             w->ops->start (w);
+            w->is_active = true;
+        }
     }
 }
 
 void flux_watcher_stop (flux_watcher_t *w)
 {
     if (w) {
-        if (w->ops->stop)
+        if (w->ops->stop) {
             w->ops->stop (w);
+            w->is_active = false;
+        }
     }
+}
+
+bool flux_watcher_is_active (flux_watcher_t *w)
+{
+    if (w)
+        return w->is_active;
+    return false;
 }
 
 void flux_watcher_destroy (flux_watcher_t *w)
--- a/src/common/libflux/reactor.h
+++ b/src/common/libflux/reactor.h
@@ -80,6 +80,7 @@ typedef void (*flux_watcher_f)(flux_reactor_t *r, flux_watcher_t *w,
 
 void flux_watcher_start (flux_watcher_t *w);
 void flux_watcher_stop (flux_watcher_t *w);
+bool flux_watcher_is_active (flux_watcher_t *w);
 void flux_watcher_destroy (flux_watcher_t *w);
 double flux_watcher_next_wakeup (flux_watcher_t *w);
 
--- a/src/common/libflux/reactor_private.h
+++ b/src/common/libflux/reactor_private.h
@@ -30,6 +30,7 @@ struct flux_watcher {
     void *arg;
     struct flux_watcher_ops *ops;
     void *data;
+    bool is_active;
 };
 
 static inline int events_to_libev (int events)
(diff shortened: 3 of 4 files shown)


chu11:
As a side note, in #4156 it was suggested that `ev_is_active()` might be used vs using some flags.  `ev_is_active()` can only be passed data structures native to the `libev` like `ev_io`.  So it'd be difficult to use `ev_is_active()` b/c we'd have to figure out the type of the watcher, determine if its a native `libev` data structure, or we gotta dig into the data structure to dig out a `libev` data structure stored in it, etc. etc.  So using flags was just easier.

garlick:
How do you handle a watcher that stops itself, like a one-shot timer?

I would think you could add a new `is_active` function pointer to `struct flux_watcher_ops`, then implement a callback for each watcher.  The ones that are just wrapping ev watchers would just call `ev_is_active()` on the `data` member like the way start and stop call `ev_io_stop()` and `ev_io_start()`.  That's as far as I got thinking about it - presumably the "compound watchers" (made up of multiple ev watchers) would need a custom callback that might do more inside...

chu11:
> How do you handle a watcher that stops itself, like a one-shot timer?

ahh i did not think about those, I guess they do not call `flux_watcher_stop()` ?

> I would think you could add a new is_active function pointer to struct flux_watcher_ops, then implement a callback for each 
> watcher. The ones that are just wrapping ev watchers would just call ev_is_active() on the data member like the way start and > stop call ev_io_stop() and ev_io_start(). That's as far as I got thinking about it - presumably the "compound watchers" (made 
> up of multiple ev watchers) would need a custom callback that might do more inside...

given the above case, then we probably have to go down this path.

garlick:
> ahh i did not think about those, I guess they do not call flux_watcher_stop() ?

Right, our timers are just a thin wrapper around `ev_timer` and there is no mechanism that I know of to get a callback when the timer stops because it has fired once and there is no repeat configured.

grondo:
> given the above case, then we probably have to go down this path.

Possibly dumb question: Will it be worth it to implement this (sounds like a fair bit of work now) for what is possibly a debug-only case? Seems like if we want to debug a specific use case, a flag could be added at a layer above the `flux_watcher_t` with much less effort.

 If implemented, we'd probably need to add `is_active()` tests for every derived watcher type too...

(I may not understand the exact use case here, so ignore me if the question was naive)

chu11:
> Possibly dumb question: Will it be worth it to implement this (sounds like a fair bit of work now) for what is possibly a debug-only case? Seems like if we want to debug a specific use case, a flag could be added at a layer above the flux_watcher_t with much less effort.

Good point.  I've needed something not so dissimilar to what I proposed here several times, so just figured I'd add it.  But you're right, we'd probably have to go down a much tougher path to implement "is_active" correctly for all watchers.

Willing to drop this.  But here's a dumb idea.  Instead of `flux_watcher_is_active()`, it could be `flux_watcher_is_started()`?  It determines if the user literally called `flux_watcher_started() / flux_watcher_stopped()`?  Also could be made private as another option.

garlick:
My only objection is we don't want to create a foot-gun for someone who tries to use it with a timer.  I suppose a compromise could be that you implement the flag for the default case, but also implement an optional watcher callback.  The timer problem could be solved by just implementing the callback and having it call `ev_is_active()`.

Still maybe not worth it.

chu11:
> My only objection is we don't want to create a foot-gun for someone who tries to use it with a timer. I suppose a compromise could be that you implement the flag for the default case, but also implement an optional watcher callback. The timer problem could be solved by just implementing the callback and having it call ev_is_active().

Perhaps a middle ground technique could be

```
bool flux_watcher_is_active()
{
    if (ops.is_active)
      return ops.is_active();
    else
        /* do flag stuff */
}
```

but a @grondo says above, that's still a lot of testing.

I'll table this for now.  If I need to do this again someday, then I figure it's worthwhile to implement for real.  I'll update issue with some of these comments.

PR #4179, opened 2022-03-02
testsuite: fix racy tests in t0005-rexec

chu11:
Problem: Several potential races exist in t0005-rexec.  In one
case, processes launched in the background may have not yet started when
`rexec_ps` is run to see the process listing.  In the second case,
`rexec_ps` may still list killed processes because internal flux
watchers have not yet noticed the killed processes / updated process listings.

Solution: Run `rexec_ps` in a loop to wait for the process listing to
reach the count we are expecting.  Timeout after 30 seconds if the
process listing is not working.

Fixes https://github.com/flux-framework/flux-core/issues/4097


grondo (approved):
LGTM. Thank you for finding and fixing this one!

garlick:
Good job!

PR #4187, opened 2022-03-04
add assertions that rank, size, hostlist broker attributes are cacheable

garlick:
Problem: rank, size, and hostlist are required to be in the `flux_t` handle attribute cache in the broker with the IMMUTABLE flag set, or deadlock could result in strange corners of the broker.

Add some assertions that will make this fail fast rather than hang, and a better comment.  Also, while puzzling about broker attributes, I found and fixed a couple of unchecked mallocs.

Thanks to @grondo for triggering this subtle footgun and wondering aloud about it, although I'm sure he enjoyed it so much that thanks will not be necessary.

### DIFF
--- a/src/broker/attr.c
+++ b/src/broker/attr.c
@@ -97,7 +97,8 @@ int attr_add (attr_t *attrs, const char *name, const char *val, int flags)
         errno = EEXIST;
         return -1;
     }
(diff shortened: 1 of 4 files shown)


grondo (approved):
LGTM! Thanks!

Issue #4182, opened 2022-03-03
job shell/resource module assigns ranks incorrectly for heterogeneous allocations


grondo:
Problem: The job shell (for unknown reason) sorts nodes in the allocated R by the number of available cores instead of their assigned rank. For an R with differing number of cores per rank, this can cause the ranks assigned by PMI to mismatch the ranks in R, which results in, at a minimum, errors about rank hostname mismatches during verification.

For now, the shell should just sort the allocation in rank order.

Eventually, `rlist_rerank ()` should probably take a hostlist argument and re-rank based on the provided order, which will allow more flexibility in the future.

grondo:
Correction: `rlist_rerank ()` does re-rank based on a hostlist, but it is only used when there is a bootstrap configured hostlist. The `resource` module uses `rlist_remap ()` when starting in a job with an existing _R_, which simply remaps all ids (rank and core ids I believe) from zero origin.

A simple fix here will be to use `rlist_rerank ()` before `rlist_remap ()`, when the `hostlist` attribute exists _and_ does not contain duplicates.

grondo:
Ok, here is what I believe is a working patch. This simply uses `rlist_rerank()` to correctly map hostnames to ranks in the `resource` module when possible:

```diff
 src/modules/resource/inventory.c | 38 ++++++++++++++++++++++++++++----
 1 file changed, 34 insertions(+), 4 deletions(-)

diff --git a/src/modules/resource/inventory.c b/src/modules/resource/inventory.c
index 8e916037d..b3a4fef27 100644
--- a/src/modules/resource/inventory.c
+++ b/src/modules/resource/inventory.c
@@ -384,6 +384,22 @@ error:
     return -1;
 }
 
+static bool no_duplicates (const char *hosts)
+{
+    struct hostlist *hl = hostlist_decode (hosts);
+    bool result = false;
+
+    if (hl) {
+        int count = hostlist_count (hl);
+        hostlist_uniq (hl);
+        result = hostlist_count (hl) == count;
+    }
+
+    hostlist_destroy (hl);
+
+    return result;
+}
+
 /* Derive resource object from R, normalizing broker ranks to origin.
  * Return success (0) but leave *Rp alone if conversion cannot be performed,
  * thus *Rp should be set to NULL before calling this function.
@@ -394,7 +410,10 @@ static int convert_R (flux_t *h, const char *job_R, int size, json_t **Rp)
     struct rlist *rl;
     json_t *R;
[...]

PR #4188, opened 2022-03-04
resource: assign ranks in R provided by job-info based on hostlist attribute

grondo:
As discussed in #4182, the job shell may not assign ranks in an order that follows the parent instance rank ordering. This occurs when there is a heterogeneous allocation of cores (i.e. different numbers of cores per execution target). This behavior is a hold over from a different time, and there is probably an argument that the shell should not do this, but there is a slightly better argument that the `hostlist` attribute is the canonical mapping of hosts to ranks for an instance, and R should follow this mapping not its own. (also, providing this solution is much easier than changing the shell behavior for now).

This PR adds support for re-ranking R in the resource module when R is provided by the parent, as it already does when R is provided by a bootstrap configuration. The re-ranking is only attempted when the `hostlist` attribute exists (which it always should), and there are no duplicates in `hostlist` (i.e. we're not running multiple brokers per node). Also, if re-ranking fails, this is not an error, since this can occur in testing when using a fake R. However, the re-ranking should fix the issue described in #4182, since there will always be a valid `hostlist` when the shell reorders hosts during a job launch of a heterogeneous allocation.

For testing purposes, the ability to set a `hostlist` on the command line is also introduced in this PR, and the test simply runs a job with a re-ordered set of hosts and ensures that the `resource.R` attribute in the resulting instance was properly rank-ordered.

### DIFF
--- a/src/broker/boot_pmi.c
+++ b/src/broker/boot_pmi.c
@@ -21,6 +21,7 @@
 #include "src/common/libutil/cleanup.h"
 #include "src/common/libutil/ipaddr.h"
 #include "src/common/libutil/kary.h"
+#include "src/common/libutil/errno_safe.h"
 #include "src/common/libpmi/pmi.h"
 #include "src/common/libpmi/pmi_strerror.h"
 #include "src/common/libpmi/clique.h"
@@ -153,6 +154,28 @@ overflow:
     return -1;
 }
 
+static int set_hostlist_attr (attr_t *attrs, struct hostlist *hl)
+{
+    const char *value;
+    char *s;
+    int rc = -1;
+
+    /*  Allow hostlist attribute to be set on command line for testing.
+     *  The value must be re-added if so, so that the IMMUTABLE flag can
+     *   be set so that the attribute is properly cached.
+     */
+    if (attr_get (attrs, "hostlist", &value, NULL) == 0) {
+        s = strdup (value);
+        (void) attr_delete (attrs, "hostlist", true);
+    }
+    else
+        s = hostlist_encode (hl);
+    if (s && attr_add (attrs, "hostlist", s, FLUX_ATTRFLAG_IMMUTABLE) == 0)
+        rc = 0;
+    ERRNO_SAFE_WRAP (free, s);
+    return rc;
+}
+
 int boot_pmi (struct overlay *overlay, attr_t *attrs)
 {
     int fanout = overlay_get_fanout (overlay);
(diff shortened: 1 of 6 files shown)


src/broker/boot_pmi.c
```diff
@@ -153,6 +153,28 @@ static int format_bind_uri (char *buf, int bufsz, attr_t *attrs, int rank)
...
+     */
+    if (attr_get (attrs, "hostlist", &value, NULL) == 0) {
+        s = strdup (value);
+        (void) attr_delete (attrs, "hostlist", true);
+    }
+    else
+        s = hostlist_encode (hl);
+    if (s && attr_add (attrs, "hostlist", s, FLUX_ATTRFLAG_IMMUTABLE) == 0)
+        rc = 0;
+    free (s);
```
garlick, line 174: `free()` potentially clobbers errno, which needs to be set so `log_err()` prints something sensible on failure.

src/modules/resource/inventory.c
```diff
@@ -384,6 +384,22 @@ static int convert_R_conf (flux_t *h, json_t *conf_R, json_t **Rp)
...
 
+static bool no_duplicates (const char *hosts)
+{
+    struct hostlist *hl = hostlist_decode (hosts);
+    bool result = false;
+
+    if (hl) {
+        int count = hostlist_count (hl);
+        hostlist_uniq (hl);
+        result = hostlist_count (hl) == count;
```
garlick, line 395: The condutional would be slightly clearer IMHO as
```c
if (hostlist_count (hl) == count)
    result = true;
```

src/modules/resource/inventory.c
```diff
@@ -410,19 +429,36 @@ static int convert_R (flux_t *h, const char *job_R, int size, json_t **Rp)
                   size);
         goto noconvert;
     }
+    /*  If we have an assigned hostlist and there is no more than
+     *   on broker per rank (i.e. no duplicates), then rerank R
+     *   based on the assigned hostlist.
+     */
```
garlick, line 435: s/on/one/

garlick (approved):
LGTM, just a couple minor things.  Nice job coming up with this fix so quickly.

dongahn:
@grondo: thanks. I don't see any red flag. So https://github.com/flux-framework/flux-sched/pull/916 should wait until this gets merged?

grondo:
>  So flux-framework/flux-sched#916 should wait until this gets merged?

No need to wait IMO.

grondo:
I've addressed @garlick's comments and force-pushed the result.

garlick:
Thanks!  Feel free to set MWP.

grondo:
Thanks!

Issue #4181, opened 2022-03-02
resource verification of configured GPUs can fail due to misconfigured hwloc


grondo:
When the _system_ installed libhwloc is not configured to properly detect GPUs as `CoProc` objects in hwloc, local resource discovery will fail to detect these GPUs and the resource module verification step will fail, resulting in drained ranks.

Perhaps we need a way to allow discovery of any GPU object during resource verification, so that when GPUs are configured in the `system.R`, we have the best chance of finding local GPUs, even when libhwloc is not configured to detect them as CoProcs.

grondo:
I'm closing this for now after merge of #4192. It didn't seem possible to expand discovery of GPUs with hwloc so for now we just don't verify GPUs already configured in R.

Users will have to ensure they have a properly built libhwloc for single user mode and dynamic resource discovery.

PR #4192, opened 2022-03-04
resource: restrict resource verification to cores/hostnames only

grondo:
As a stopgap solution for #4181, drop the verification of "gpu" resources (actually any resource except "core") by the flux-core `resource` module in all cases. The `rlist_verify()` used by the resource module will still verify that rank:hostname mapping is correct.

This should allow explicit configuration of GPUs by sysadmins without additionally requiring the `resource.noverify` option, for which we do not have a good plan for propagation to child instances.

Additionally, verification of cores and hostnames has proven valuable lately in the detection of incorrect rank ordering, so it would be nice to keep that part of it for now.

If at some point in the future the system-installed default libhwloc can detect GPUs reliably, then GPU devices can be added back into the verification step. (But we aren't holding our breath)

### DIFF
--- a/src/modules/resource/topo.c
+++ b/src/modules/resource/topo.c
@@ -87,26 +87,43 @@ static int drain_self (struct topo *topo, const char *reason)
 static int topo_verify (struct topo *topo, json_t *R, bool nodrain)
 {
     json_error_t e;
-    struct rlist *rl;
+    struct rlist *rl = NULL;
+    struct rlist *rl_cores = NULL;
+    struct rlist *r_local_cores = NULL;
     flux_error_t error;
-    int rc;
+    int rc = -1;
 
     if (!(rl = rlist_from_json (R, &e))) {
         flux_log (topo->ctx->h, LOG_ERR, "R: %s", e.text);
         errno = EINVAL;
         return -1;
     }
-    rc = rlist_verify (&error, rl, topo->r_local);
+
+    /*  Only verify cores (and rank hostname) for now.
+     *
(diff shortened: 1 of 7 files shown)


t/t2315-resource-system.t
```diff
@@ -101,4 +101,20 @@ test_expect_success 'invalid exclude hosts cause instance failure' '
     grep "nosuchhost: Invalid argument" ${name}/logfile
```
garlick, line 101: Commit message: "doesn't not" - extra negative there?

src/modules/resource/topo.c
```diff
@@ -87,26 +87,35 @@ static int drain_self (struct topo *topo, const char *reason)
 static int topo_verify (struct topo *topo, json_t *R, bool nodrain)
```
garlick, line 87: It might be helpful later to have a comment block here noting that gpus are purposefully not verified with a reference to #4181?

garlick (approved):
LGTM!    Comments are documentation only.

t/t2315-resource-system.t
grondo: Ok, will fix to "doesn't not not" 😉

src/modules/resource/topo.c
grondo: Yes, good call! I'll add that.

grondo:
Thanks! I fixed the commit message typo and added suggested comment to topo_verify() (wise suggestion).

PR #4194, opened 2022-03-05
flux-top: add man page, minor bug fixes

garlick:
`flux-top` was missing a man page, so here's one.

Also, some small fixes for when `flux-top` is run on a system instance early in system startup:
- don't die if time drift causes a negative FSD, just print zero
- update summary job count even if there is no job activity (in case first update silently failed with ENOSYS during startup)
- if `job-list` is slow (like during startup with many inactive jobs), don't pile on more and more requests, just wait patiently.

These are not likely scenarios for users, but maybe for system administrators.


src/cmd/top/summary_pane.c
```diff
@@ -216,7 +216,9 @@ static void draw_info (struct summary_pane *sum)
     double now = flux_reactor_now (flux_get_reactor (sum->top->h));
     char fsd[32] = "";
 
-    (void)fsd_format_duration_ex (fsd, sizeof (fsd), now - sum->starttime, 2);
+    (void)fsd_format_duration_ex (fsd,
+                                  sizeof (fsd),
+                                  fabs (now - sum->starttime), 2);
```
chu11, line 221: the 2 should be indented on the next line?

chu11 (approved):
LGTM, always good to get more documentation.  Noted one nit.

garlick:
Fixed that, and reworded some silliness in the man page.

garlick:
Coverage builder seems to have failed early so I restarted CI.

PR #4196, opened 2022-03-06
sched-simple: fix counting bug that can cause scheduler to fail after a restart

grondo:
This fixes a counting bug in librlist where cores allocated from ranks that are down caused an underflow in count of available cores. This caused sched-simple to think there were less than zero cores available after a restart with running jobs, so no further jobs could be scheduled.

### DIFF
--- a/src/common/librlist/rlist.c
+++ b/src/common/librlist/rlist.c
@@ -1660,7 +1660,8 @@ static int rlist_alloc_rnode (struct rlist *rl, struct rnode *n)
     }
     if (rnode_alloc_idset (rnode, n->cores->avail) < 0)
         return -1;
-    rl->avail -= idset_count (n->cores->avail);
+    if (rnode->up)
+        rl->avail -= idset_count (n->cores->avail);
     return 0;
 }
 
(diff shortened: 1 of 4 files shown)


grondo:
You can adapt the reproducer script from this PR, pasted here for convenience.
Just replace the `flux module reload sched-simple` with the appropriate commands to reload the Fluxion scheduler. Then run the script on a system with Fluxion as the default scheduler simply as:

```console
$ flux start -s 1 ./test.sh
```

```sh
#!/bin/sh -e

which flux

NCORES=$(flux resource list -s up -no {ncores})
jobids=$(flux mini submit --cc=0-1 -n $NCORES sleep 600)
id=$(echo $jobids | cut -d ' ' -f1)
id2=$(echo $jobids | cut -d ' ' -f2)
flux job wait-event $id start

flux resource drain 0

flux module reload sched-simple

flux resource list
flux resource undrain 0

echo "t4184: waiting for resources to be back up:"
while test $(flux resource list -s up -no {ncores}) -ne $NCORES; do
  sleep 0.25
done

echo "t4184: canceling $id"
flux job cancel ${id}

echo "t4184: waiting for $id to end"
flux job wait-event $id clean

echo "t4184: waiting for $id2 to start..."

flux job wait-event -t 15 $id2 start

echo "t4184: canceling $id2..."
flux job cancel $id2

echo "t4184: waiting for $id2 to end..."
flux job wait-event -t 15 $id2 clean
```

Issue #4195, opened 2022-03-06
early exec failure causes job hang


garlick:
Problem: a job seems to be stuck in RUN state after failing to launch.

On a system instance with 5 compute nodes, I tried to run `stress` across 5 nodes, but it was not installed on all the of the nodes.    The job became stuck in RUN state.  I was unable to cancel the job, and when I tried to stop flux with `systemctl stop flux` on the rank 0 node, it got stuck in rc3 running `queue idle --quiet`.  To avoid losing data when systemd eventually times out (#4158), I killed that process manually and rc3 completed.
```
$ flux mini run -N5 -n5 -t 5s stress --cpu=1
0.778s: flux-shell[1]: FATAL: task 0: start failed: stress: No such file or directory
0.779s: flux-shell[0]: FATAL: task 0: start failed: stress: No such file or directory
0.784s: job.exception type=exec severity=0 task 0: start failed: stress: No such file or directory
0.791s: job.exception type=exec severity=0 task 0: start failed: stress: No such file or directory
0.788s: flux-shell[2]: FATAL: task 0: start failed: stress: No such file or directory
0.802s: job.exception type=exec severity=0 task 0: start failed: stress: No such file or directory
5.488s: job.exception type=timeout severity=0 resource allocation expired
^Cflux-job: one more ctrl-C within 2s to cancel or ctrl-Z to detach

flux-job: stdin not accepted by job
$ flux jobs
       JOBID USER     NAME       ST NTASKS NNODES  RUNTIME NODELIST
  ƒCqRCsckGP garlick  stress      R      5      5   0.756s picl[2-3,5-7]

$ flux job eventlog ƒCqRCsckGP
1646582397.433282 submit userid=5588 urgency=16 flags=0
1646582397.453018 depend
1646582397.453103 priority priority=16
1646582397.461234 alloc annotations={"sched":{"resource_summary":"rank[2-3,5-7]/core0"}}
1646582397.461562 prolog-start description="job-manager.prolog"
1646582397.898796 prolog-finish description="job-manager.prolog" status=0
1646582397.916531 start
1646582398.217496 exception type="exec" severity=0 userid=5588 note="task 0: start failed: stress: No such file or directory"
1646582398.224415 exception type="exec" severity=0 userid=5588 note="task 0: start failed: stress: No such file or directory"
1646582398.234865 exception type="exec" severity=0 userid=5588 note="task 0: start failed: stress: No such file or directory"
[...]

PR #4199, opened 2022-03-08
job-exec: fix more potential hangs after early shell failure

grondo:
As described in #4195 early exec failures _after_ the initial shell barrier may still cause job hangs since shells could block stuck at a barrier if for some reason an exception can't be raised, or even if signals can't be delivered due to a IMP bug or other issue.

Since the exec system knows that shells have exited during or before a barrier, we might as well have the module abort the barrier at the time, rather than waiting for "something else" to clean up the shells. This will result in a more timely job exit when everything is working correctly, and will also result in a proper cleanup when things are not working well (as was the case in #4195).

### DIFF
--- a/src/modules/job-exec/exec.c
+++ b/src/modules/job-exec/exec.c
@@ -45,6 +45,7 @@ struct exec_ctx {
     const char * mock_exception;   /* fake exception */
     int barrier_enter_count;
     int barrier_completion_count;
+    int exit_count;
 };
 
 static void exec_ctx_destroy (struct exec_ctx *tc)
@@ -156,6 +157,19 @@ static int exec_barrier_enter (struct bulk_exec *exec)
         ctx->barrier_enter_count = 0;
         ctx->barrier_completion_count++;
     }
+    else if (ctx->barrier_enter_count == 1 && ctx->exit_count > 0) {
+        /*
+         *  Terminate barrier with error immediately when a barrier is
+         *   started after one or more shells have already exited. The
+         *   case where a shell exits while a barrier is already in progress
+         *   is handled in exit_cb().
+         */
+        if (bulk_exec_write (exec,
+                             "FLUX_EXEC_PROTOCOL_FD",
+                             "exit=1\n",
+                             7) < 0)
+            return -1;
+    }
     return 0;
 }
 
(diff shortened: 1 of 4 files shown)


src/modules/job-exec/exec.c
```diff
@@ -239,17 +265,23 @@ static void exit_cb (struct bulk_exec *exec,
...
 
-        /*  Terminate barrier with failed exit status.
-         *  This will allow any shells that do get to the barrier to exit
-         *   immediately, instead of waiting to be killed by exec system.
-         */
+    /*  If a shell exited before the first barrier or there is a
+     *   barrier in progresss (enter_count > 0), then terminate the
+     *   current/next barrier immediately with error. This will allow
+     *   shells currently waiting or entering the barrier in the future
+     *   to exit immediately, rather than being killed by exec system.
```
garlick, line 274: s/progresss/progress/

src/modules/job-exec/exec.c
```diff
@@ -45,6 +45,7 @@ struct exec_ctx {
     const char * mock_exception;   /* fake exception */
```
garlick, line 45: commit message for 9652288a830450c8979be49eb279fcebfa587b8d:

I'm not quite parsing this (can an in-progress barrier also be complete?)
> - a shell exits while an in-progress barrier is complete

src/shell/shell.c
```diff
@@ -1318,7 +1318,9 @@ int main (int argc, char *argv[])
             else if (errno == ENOENT)
                 ec = 127;
             shell_die (ec, "task %d: start failed: %s: %s",
-                       i, flux_cmd_arg (task->cmd, 0), strerror (errno));
+                       task->rank,
+                       flux_cmd_arg (task->cmd, 0),
+                       strerror (errno));
```
garlick, line 1323: If this is where we end up when an executable cannot be started on one of the nodes, it might also be helpful to call out the hostname, if that's not too inconvenient.

e.g. this is what I get when I repeat the test from #4165:
```
 garlick@picl0~$ flux mini run -N4 -n16 stress --timeout 60 --cpu 1
0.794s: flux-shell[0]: FATAL: task 0: start failed: stress: No such file or directory
0.795s: job.exception type=exec severity=0 task 0: start failed: stress: No such file or directory
```

garlick (approved):
Looks great, and tests great too!  (I repeated original scenario from #4195 on my test cluster several times.)

I had a few comments...about comments and one about a log message, but minor.  This LGTM.

src/modules/job-exec/exec.c
grondo: Sorry, that must be a typo (I don't understand it either!)

src/shell/shell.c
garlick: Sorry, I should have said I realize that would be an enhancement, not a deficiency of the proposed pr which is fixing a bug.  Feel free to skip handling it here.

grondo: Since rank doesn't seem to be useful in log messages, maybe RFC 24 "log" events should include the hostname of the shell which generated them (instead of or in addition to the shell rank), and then front end tools could display the hostname for all log messages?

grondo: Talking offline, it was decided that since the rank->hostname mapping already exists in R, log consumers that want to display hostnames can do so by fetching R first.

src/modules/job-exec/exec.c
grondo: Fixed to `a shell exits before an in-progress barrier is complete`

PR #4200, opened 2022-03-09
shell: add hostname to a couple key log messages

grondo:
This PR adds the hostname to the log message (and exception note) when a task fails to start (e.g. due to missing executable), or when the shell is unable to `chdir()` to the indicated `cwd` in jobspec.

It will be helpful to have hostnames associated with these errors readily available in the logs or exception note when only a subset of job shells encounter them. We don't go as far as adding the hostname to every error or log message, however, as the rank to hostname mapping is available after the fact for other, more rare errors.

We could eventually _optionally_ display hostnames instead of shell ranks from `flux job attach` by having the command fetch _R_ and use that to lookup the mapping when displaying log messages. That wasn't done for now since it doesn't seem like a critical enhancement at this time.

### DIFF
--- a/src/shell/internal.h
+++ b/src/shell/internal.h
@@ -14,6 +14,7 @@
 #include <flux/core.h>
 #include <flux/optparse.h>
 #include <flux/shell.h>
+#include <sys/param.h>
 
 #include "src/common/libutil/aux.h"
 #include "src/common/libczmqcontainers/czmq_containers.h"
@@ -24,6 +25,7 @@
 struct flux_shell {
     flux_jobid_t jobid;
     int broker_rank;
+    char hostname [MAXHOSTNAMELEN + 1];
     int protocol_fd;
 
     optparse_t *p;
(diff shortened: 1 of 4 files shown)


chu11 (approved):
LGTM, just a typo in commit message "prooblem" -> "problem".

garlick (approved):
I concur, looks great, and thanks for doing this.

chu11:
tests LGTM, just one sentence in the commit message didn't quite sound right:

```
but some errors may contain text that is difficult to match
or small changes such as a hostname, which will make exact matches
difficult.
```

did you mean something like "but some errors may contain text that is difficult to match (such as hostname)"?

grondo:
>  just one sentence in the commit message didn't quite sound right:

Yeah, the wording is awkward, sorry. I'll improve. (The intent is to protect the tests from breaking due to small future changes, such as addition of hostname or similar)

grondo:
Ok, I improved the problem statement to:

    Problem: The exception tests in python/t0010-job.py require exact
    matches, but this has the potential for tests that break with small
    changes to the log messages in the future, such as the addition of
    a hostname.

Hopefully an improvement?

Issue #4185, opened 2022-03-03
'flux resource drain' doesn't distinguish between drained nodes that are still running jobs vs. fully idle


ryanday36:
Currently, when an administrator drains a node with `flux resource drain`, the node immediately goes into a 'drain' state, regardless of whether or not there are still active jobs on the node. LC operations has use cases (e.g. some hardware replacements) where they would like a simple way to be able to tell if a node is drained and idle ('drain') vs. drained but not all running jobs have completed ('draining').

PR #4205, opened 2022-03-09
flux-resource: differentiate drained vs draining ranks

grondo:
As described in #4185, `flux resource drain` and `flux resource status` do not differentiate between ranks that are drained and still have jobs allocated, vs ranks that are "idle" (and thus ready to be taken down for service).

This PR enhances `flux-resource` to fetch  the `sched.resource-status` RPC in addition to `resource.status` and uses the result to create an idset of "draining" ranks by taking the intersection of `drain` ranks with those the scheduler has allocated.

### DIFF
--- a/doc/man1/flux-resource.rst
+++ b/doc/man1/flux-resource.rst
@@ -34,6 +34,9 @@ A few notes on drained nodes:
 - When an instance is restarted, drained nodes remain drained.
 - The scheduler may determine that a job request is *feasible* if the total
   resource set, including drained nodes, would allow it to run.
+- In ``flux resource status`` and ``flux resource drain``, the drain state
+  of a node will be presented as "drained" if the node has no job allocations,
+  and "draining" if there are still jobs running on the node.
 
 Some further background on resource service operation may be found in the
 RESOURCE INVENTORY section below.
@@ -57,7 +60,9 @@ COMMANDS
    increase output verbosity.  *-n,--no-header* suppresses header from output.
    *-o,--format=FORMAT*, customizes output formatting (see below).
    *-s,--states=STATE,...* limits output to specified resource states, where
-   valid states are "online", "offline", "avail", "exclude", "drain", and "all".
+   valid states are "online", "offline", "avail", "exclude", "draining",
+   "drained", and "all". The special "drain" state is also supported, and
+   selects both draining and drained resources.
 
 **drain** [-f] [-u] [targets] [reason ...]
    If specified without arguments, list drained nodes.  The *targets* argument
--- a/src/cmd/flux-resource.py
+++ b/src/cmd/flux-resource.py
@@ -20,6 +20,7 @@ from flux.idset import IDset
 from flux.hostlist import Hostlist
 from flux.resource import ResourceSet, resource_list, SchedResourceList
 from flux.rpc import RPC
+from flux.future import WaitAllFuture
 
 
 def reload(args):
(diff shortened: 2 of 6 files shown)


grondo:
An example:
```
ƒ(s=4,d=0,builddir) grondo@asp:~/git/issues/draining$ flux mini submit sleep 300ƒ4yPBPvw
ƒ(s=4,d=0,builddir) grondo@asp:~/git/issues/draining$ flux resource drain 0-1 testing
ƒ(s=4,d=0,builddir) grondo@asp:~/git/issues/draining$ flux resource drain
TIMESTAMP            STATE    RANK     REASON                         NODELIST
2022-03-09T15:38:05  draining 0        testing                        asp
2022-03-09T15:38:05  drain    1        testing                        asp
ƒ(s=4,d=0,builddir) grondo@asp:~/git/issues/draining$ flux resource status
    STATUS NNODES RANKS           NODELIST
     avail      3 0,2-3           asp,asp,asp
     drain      1 1               asp
  draining      1 0               asp
```

I just realized I still need to update the `flux-resource(1)` man page, so I'll get to that next.

garlick:
Great!  Quick thought - maybe `drain` should be renamed to `drained`?  (-ed to complement the -ing)

grondo:
> Great! Quick thought - maybe drain should be renamed to drained?

I actually did that at first, but it seemed to require more changes than I was willing to do at first. However, now there is a bit more churn anyway, so I'll go ahead and do it as suggested.

grondo:
Still a WIP, but now based on top of #4209 to get updated `IDset` methods.

Also, there's a fixup commit which changes `drain` to `drained`, while redefining a special `drain` state which includes both `drained` and `draining` ranks. I wanted to get feedback on that approach before squashing the work (unfortunately, due to the process the changes might have to be one commit).

The idea is that `flux resource status -s drain -no {nnodes}` will still give you the count of draining+drained nodes, which may be useful sometimes (and indeed, prevents breaking several tests). The `drain` state is never displayed by default, only when requested in the tool. `flux resource drain` never displays the drain state so its output should not be confusing.

Docs are still a todo item here.

garlick (commented):
Should `flux resource drain` somehow indicate whether the drained nodes are running jobs?

Just asking, I'm fine if you think not.

grondo:
> Should flux resource drain somehow indicate whether the drained nodes are running jobs?

I could be confused, but that is what `draining` vs `drained` is supposed to indicate. I might be missing something in your question though?

```
ƒ(s=4,d=0,builddir) grondo@asp:~/git/issues/draining$ flux resource drain 0-1
ƒ(s=4,d=0,builddir) grondo@asp:~/git/issues/draining$ flux resource drain
TIMESTAMP            STATE    RANK     REASON                         NODELIST
2022-03-11T09:34:44  draining 0                                       asp
2022-03-11T09:34:44  drained  1                                       asp
```

garlick:
Oops I think I was thrown off course by this comment

> flux resource drain never displays the drain state so its output should not be confusing.

I forgot that this was already done in addition to the `flux resrouce status` changes.  Sorry, being too hasty with my comments!

grondo:
> Oops I think I was thrown off course by this comment

No problem, what I meant there (and failed to sensibly convey) is that `flux resource drain` will only show ranks in either `drained` or `draining`, not the virtual `drain` state which means either `drained` or `draining` in `flux resource status`.

grondo:
This is probably ready for a review.

One caveat that bears considering: When a scheduler is not loaded, nothing in Flux really knows what ranks are currently allocated, so `flux resource status` and `flux resource drain` will report all "drain" ranks as `drained`. (it currently assumes no rank are allocated in this state)

Maybe `flux-resource` should flag this condition somehow (like maybe by calling out the state as `drain?` or similar). Alternatively, all drain ranks could be flagged as `draining` in this state since that is most safe.

garlick (approved):
LGTM!

grondo:
Thanks! Setting MWP.

Issue #4202, opened 2022-03-09
kvs dump/restore utilities needed


garlick:
The following was proposed in #258

> Let's say that this issue can be closed if we can garbage collect the content store on the way up from an instance restart, based on following the last-written `kvs-primary` checkpoint, and deleting everything that's not referenced.

A naive approach is to build a list of all blobrefs, then walk them from the checkpoint, adding reference counts, then delete all the blobs with a reference count of zero.  Each content back end would need to grow a `delete` method, and an optional `rebuild` method, since sqlite (at least) requires a [vacuum](https://www.sqlite.org/lang_vacuum.html) operation to actually reclaim disk space after deletions.  A less naive approach might involve a specialized utility for each back end, or specialized builtin methods for performing the walk within the backend itself.

A simpler idea is to add serial dump/restore utilities.  Dump walks blobs from the the checkpoint (interpreting RFC 11 objects) and writes data in a standard archive format like tar, with each kvs key represented as a file.  Restore reads the archive, recreates RFC 11 metadata for the "files" on the archive, writes new blobs to the content store, and restores the checkpoint itself.  Offline garbage collection could be accomplished by dumping the kvs, removing the back-end database (e.g. `content.sqlite`) and restoring it from the archive.

Besides simplicity, this approach has some additional merits:
- it would optimize the metadata representation of eventlogs (blobref per event converted to single blobref)
- ability to dump from one back end and restore to another (e.g. migrating from content-sqlite to content-s3)
- ability to dump from one hash algo and restore to another (e.g. migrating from sha-1 to sha-256)
- portability of the archive across flux versions as the archive itself would not contain any RFC 11 metadata
- ability to manipulate the archive outside of flux with standard tools
- space efficient, as a serial archive can be compressed

PR #4208, opened 2022-03-10
add flux-dump and flux-restore

garlick:
As proposed in #4202, here are dump/restore utilities that walk KVS metadata using the content interface so they can be used when the KVS is not loaded.  They use [libarchive](https://www.libarchive.org/) to support various portable archive formats.

Still playing with these and plan to add tests and man pages but wanted to get this checkpointed.

### DIFF
--- a/doc/man1/flux-dump.rst
+++ b/doc/man1/flux-dump.rst
@@ -1,0 +1,124 @@
+.. flux-help-description: Dump kvs to portable archive format
+
+============
+flux-dump(1)
+============
+
+
+SYNOPSIS
+========
+
+**flux** **dump** [*OPTIONS*] *OUTFILE*
+
+
+DESCRIPTION
+===========
+
+The ``flux-dump`` command writes a snapshot of the Flux key value store
+to a file in a portable archive format.
+
+The snapshot is of the primary namespace of the current KVS root by default.
+If *--checkpoint* is specified, the snapshot is of the last KVS checkpoint
+written to the content backing store.
+
+The format of the archive and compression algorithm, if any, is determined
+by the output file name extension.  The supported formats and compression
+algorithms are a capability of :linux:man3:`libarchive`.  Some common
+extensions that are supported by modern versions of that library are:
+
+.tar
+   POSIX *ustar* format, compatible with :linux:man1:`tar`.
+
+.tgz, .tar.gz
+   POSIX *ustar* format, compressed with :linux:man1:`gzip`.
+
+.tar.bz2
+   POSIX *ustar* format, compressed with :linux:man1:`bzip2`.
+
+.tar.xz
+   POSIX *ustar* format, compressed with :linux:man1:`xz`.
+
+.zip
+   ZIP archive, compatible with :linux:man1:`unzip`.
+
+.cpio
+   POSIX CPIO format, compatible with :linux:man1:`cpio`.
+
+.iso
+   ISO9660 CD image
+
+The output filename may be *-* to indicate standard output.  In that case,
+uncompressed tar format is used.
+
+
+OPTIONS
+=======
+
+**-h, --help**
+   Summarize available options.
+
+**-v, --verbose**
+   List keys on stderr as they are dumped instead of a periodic count of
+   dumped keys.
+
+**-q, --quiet**
+   Don't show periodic count of dumped keys on stderr.
+
+**--checkpoint**
+   Generate snapshot from the latest checkpoint written to the content
+   backing store, instead of from the current KVS root.
+
+**--no-cache**
+   Bypass the broker content cache and interact directly with the backing
+   store.  This may be slightly faster, depending on how frequently the same
+   content blobs are referenced by multiple keys.
+
+
+OTHER NOTES
+===========
+
+KVS commits are atomic and propagate to the root of the namespace.  Because of
+this, when ``flux-dump`` archives a snapshot of a live system, it reflects one
+point in time, and does not include any changes committed while the dump is
+in progress.
+
+Since ``flux-dump`` generates the archive by interacting directly with the
+content store, the *--checkpoint* option may be used to dump the most recent
+state of the KVS when the KVS module is not loaded.
+
+Only regular values and symbolic links are dumped to the archive.  Directories
+are not dumped as independent objects, so empty directories are omitted from
+the archive.
+
+KVS symbolic links represent the optional namespace component in the target
+as a *NAME::* prefix.
+
+The KVS path separator is converted to the UNIX-compatible slash so that the
+archive can be unpacked into a file system if desired.
+
+The modification time of files in the archive is set to the time that
+``flux-dump`` is started if dumping the current KVS root, or to the timestamp
+of the checkpoint if *--checkpoint* is used.
+
+The owner and group of files in the archive are set to the credentials of the
+user that ran ``flux-dump``.
+
+The mode of files in the archive is set to 0644.
+
+
+RESOURCES
+=========
+
+Flux: http://flux-framework.org
+
+RFC 10: Content Storage Service: https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_10.html
+
+RFC 11: Key Value Store Tree Object Format v1: https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_11.html
+
+
+
+
+SEE ALSO
+========
+
+:man1:`flux-restore`, :man1:`flux-kvs`
--- a/doc/man1/flux-restore.rst
+++ b/doc/man1/flux-restore.rst
@@ -1,0 +1,72 @@
+.. flux-help-description: Restore kvs from portable archive format
+
+===============
+flux-restore(1)
+===============
+
+
+SYNOPSIS
+========
+
+**flux** **restore** [*OPTIONS*] *INFILE*
+
+
+DESCRIPTION
+===========
+
+The ``flux-restore`` command reads a KVS snapshot from a portable archive
+format, usually written by :man1:`flux-dump`.
+
+The archive source may be specified as a file path or *-* for standard input.
+The format of the archive may be any of the formats supported by
+:linux:man3:`libarchive` and is determined on the fly based on the archive
+content.
+
+The snapshot may be restored to a KVS key if *--key=NAME* is used and the
+KVS service is running, or as a checkpoint in the content backing store
+if *--checkpoint* is used, without the KVS running.  One of those two options
+is required.
+
+
+OPTIONS
+=======
+
+**-h, --help**
+   Summarize available options.
+
+**-v, --verbose**
+   List keys on stderr as they are restored instead of a periodic count of
+   restored keys.
+
+**-q, --quiet**
+   Don't show a periodic count of restored keys on stderr.
+
+**--checkpoint**
+   After restoring the archived content, write the final root blobref
+   to the KVS checkpoint area in the content backing store.  The checkpoint
+   is used as the initial KVS root when the KVS module is loaded.  Unload
+   the KVS module before restoring with this option.
+
+**--key**\ =\ *NAME*
+   After restoring the archived content, write the final root blobref
+   to a KVS key, so the key becomes the restored root directory.
+
+**--no-cache**
+   Bypass the broker content cache and interact directly with the backing
+   store.  Performance will vary depending on the content of the archive.
+
+
+RESOURCES
+=========
+
+Flux: http://flux-framework.org
+
+RFC 10: Content Storage Service: https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_10.html
+
+RFC 11: Key Value Store Tree Object Format v1: https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_11.html
+
+
+SEE ALSO
+========
+
+:man1:`flux-dump`, :man1:`flux-kvs`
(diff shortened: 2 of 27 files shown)


garlick:
Some numbers from my test cluster, FWIW:
- 22094 inactive jobs
- 1234 MB content.sqlite
- 122s  to load job manager
- 102s to load job-list

With flux stopped I ran this script, which took 5m+57s to run
```bash
#!/bin/bash
SQLITE_FILE=/net/frankenfurter/pool/scratch/flux/content.sqlite

reader() {
        ./flux start -o,-Sbroker.rc1_path= \
                -o,-Sbroker.rc3_path=\
                -o,-Scontent.backing-path=$SQLITE_FILE.bak \
                bash -c "\
                        flux module load content-sqlite && \
                        flux dump --checkpoint - \
                "
}

writer() {
        ./flux start -o,-Sbroker.rc1_path= \
                -o,-Sbroker.rc3_path= \
                -o,-Scontent.backing-path=$SQLITE_FILE \
                bash -c "\
                        flux module load content-sqlite && \
                        flux restore -v --checkpoint - \
                "
}

mv $SQLITE_FILE $SQLITE_FILE.bak
reader | writer
```
Afterwards:
- 22094 inactive jobs (no change)
- 731 MB content.sqlite (0.59 x)
- 40s  to load job manager (0.32 x)
- 54s to load job-list (0.52 x)

grondo:
> Other things I thought of but felt could be deferred are:

Another idea we had

 * A filter or transform step before restore which would allow - among other things like pruning jobs - restoring to an instance running on a different resource set. This would perhaps allow for easy checkpoint/restore of Flux instances without requiring the instance restart on the same allocation. (We could call it the "everlasting jobstopper")

chu11 (commented):
just did a skim, overall looks good, didn't notice much out of place.  just a lot of nits / minor things.

doc/man1/flux-restore.rst
```diff
@@ -0,0 +1,67 @@
...
+KVS service is running, or as a checkpoint in the content backing store
+if *--checkpoint* is used, without the KVS running.
+
+OPTIONS
+=======
+
+**-h, --help**
+   Summarize available options.
+
+**-v, --verbose**
```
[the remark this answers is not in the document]
garlick: The level is not used so I'll make it boolean.

doc/man1/flux-dump.rst
```diff
@@ -0,0 +1,120 @@
...
+uncompressed tar format is used.
+
+
+OPTIONS
+=======
+
+**-h, --help**
+   Summarize available options.
+
+**-v, --verbose**
```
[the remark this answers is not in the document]
garlick: Same here

src/common/libkvs/kvs_checkpoint.c
```diff
@@ -91,45 +91,21 @@ int kvs_checkpoint_lookup_get_rootref (flux_future_t *f, const char **rootref)
...
                                 "value",
                                 "version", &version,
-                                "timestamp", &timestamp) < 0)
-        return -1;
-
-    if (version != 0 && version != 1) {
-        errno = EINVAL;
+                                "timestamp", &ts) < 0)
         return -1;
-    }
```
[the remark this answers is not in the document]
garlick: Sorry I think you are right.

The thought I was having was that we don't want to artificially prevent a system administrator from reverting a flux-core upgrade that happens to raise  the checkpoint version number here, when the timestamp is probably not changing; however, I think that could be more of an issue for deciding when to bump the version number.  It probably shouldn't be bumped at all unless the new version introduces a breaking change.

So I'll fix that, thanks.

doc/man1/flux-dump.rst
```diff
@@ -0,0 +1,120 @@
...
+   List keys on stderr as they are dumped.
+
+**--checkpoint**
+   Generate snapshot from the latest checkpoint written to the content
+   backing store, instead of from the current KVS root.
```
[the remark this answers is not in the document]
garlick: Dropping the optional arg so the man page will be correct.

doc/man1/flux-restore.rst
```diff
@@ -0,0 +1,67 @@
...
+   List keys on stderr as they are restored.
+
+**--checkpoint**
```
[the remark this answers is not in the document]
garlick: Dropping the optional arg so the man page will be correct.

garlick:
Just pushed an update with changes squashed (hope that's OK):
- made `--verbose` and `--snapshot` boolean options consistent with man page
- changed kvs snapshot helpers to use the `kvs-primary` if the key is NULL since that's all we support now
- restored snapshot object version check
- fixed other typos and whitespace issues called out by @chu11
- improved testing to ensure that "valref" objects are covered, and to count blobs before and after a restore to show GC benefit
- print number of keys and blobs restored on `flux-restore` completion
- fixed a  json memory leak in `flux-restore` found with valgrind
- rebased on current master

garlick:
Oops found some formatting errors in the man pages so forced a push to correct those, and while I was in there, make an editing pass for clarity.

garlick:
I'm hitting a surrogate error in @chu11's job-archive sqlite query tool that I reused here to count blobs in the sqlite database, only  in the inception builder.  It looks like it may be fixed by @grondo's pr #4212.  Fortuitous!  I'll try rebasing this PR on that one and see how it goes.

grondo:
Just trying out this branch on my size=90 full system instance on fluke with 40K+ jobs.

One thing that might be nice to have is a progress meter in `flux dump/restore`. I'm not sure if that is even possible, but a command running for many minutes with no visible output is somewhat disconcerting, and `-v, --verbose` is way too much information. Anything in-between might be helpful, even just a count of data dumped or restored?

garlick:
Argh, sorry, just re-pushed again - I had let a fix for #4210 leak onto this branch that was not quite right.
Also rebased on @grondo's latest copy of his branch.

garlick:
Looks like #4212 fixed my problem.  Thanks!

I added a periodic count of dumped keys to stderr to both `flux-dump` and `flux-restore`.  It is disabled if a `-q,--quiet` option is specified, or if `-v,--verbose`. 

Once #4212 gets merged, I'll rebase this on master and push the progress stuff too.

I think the coverage here could be better, but only if we simulate bad  KVS metadata in the store.  That's not going to be fun.  My vote is to save that for another day.

garlick:
OK, re-pushed after rebasing, and with new progress display (overwrites line on stderr every 100 keys dumped/restored) and  `--quiet` option.

grondo:
Nice! The progress display is very nice, IMO.

chu11:
> Since flux-dump generates the dump by interacting directly with the
content store, a dump may be created when the KVS is not loaded, as
long as the --checkpoint option is supplied so it has a starting blobref.
Otherwise, the starting blobref is obtained by requesting the current
root from the KVS.

Saw this in a commit message and made me wonder, would a `--checkpoint-blobref` option to reasonable for `flux -dump`?  Like if a user just saved off the value after calling `flux kvs getroot`?

Was just a thought.

chu11:
Another random thought / idea.  Should the key a user restores to be a symlink?  Or possibly in a different namespace?  I'm not entirely sure if there's a realistic use case.  I'm imagining like a user wants to re-start their instance from a specific point in time?  Just a random brainstorm, feel free to ignore.

t/t2807-dump-cmd.t
```diff
@@ -0,0 +1,172 @@
...
+test_expect_success 'repeat dump with --no-cache' '
+	flux dump --no-cache --checkpoint foo.tar
+'
+test_expect_success 'unload content-sqlite' '
+	flux content flush &&
+	flux content dropcache &&
+	flux module remove content-sqlite
+'
+# Small values are contained in dirent rather than separate objects, so expect:
+#  rootdir-v0 + (rootdir-v1 + a + b + x) + rootdir-v2 + rootdir-v3 + rootdir-v4
```
chu11, line 69: it took me a while to grok this comment and get to how you go to 8.  Perhaps more clear (I think the +x above is a cut & paste error?)

```
rootdir-v0 (initial rootdir) + (rootdir-v1 (add a) + dir a + dir b) + (rootdir-v2 (add y)) + (rootdir-v3 (add x) + valref x) + (rootdir-v4 (add z)) = 8 blobrefs
```

t/t2807-dump-cmd.t
```diff
@@ -0,0 +1,172 @@
...
+test_expect_success 'repeat restore with --no-cache' '
+	flux restore --no-cache --checkpoint foo.tar
+'
+test_expect_success 'unload content-sqlite' '
+	flux content flush &&
+	flux content dropcache &&
+	flux module remove content-sqlite
+'
+# Intermediate rootdir versions are not restored so expect:
+#  rootdir-v4 + a + b + x
```
chu11, line 97: for consistency to my 0above comment, dir a, dir b, valref x?

doc/man1/flux-dump.rst
```diff
@@ -0,0 +1,124 @@
...
+SYNOPSIS
+========
+
+**flux** **dump** [*OPTIONS*] *OUTFILE*
+
+
+DESCRIPTION
+===========
+
+The ``flux-dump`` command write a snapshot of the Flux key value store
```
chu11, line 17: "writes a snapshot"? (plural write)

doc/man1/flux-restore.rst
```diff
@@ -0,0 +1,82 @@
...
+
+
+CAVEATS
+=======
+
+The KVS module writes its final root checkpoint to the backing store when
+it is unloaded, replacing whatever was there before.  Therefore, restoring
+the KVS checkpoint from archive with the KVS module loaded and then unloading
+the KVS module causes the restored blobref to be overwritten.  Please ensure
+the KVS module is unloaded when using the *--checkpoint* option.
```
chu11, line 66: idea: could kvs-restore check that the module isn't loaded?

doc/man1/flux-restore.rst
```diff
@@ -0,0 +1,82 @@
...
+The format of the archive may be any of the formats supported by
+:linux:man3:`libarchive` and is determined on the fly based on the archive
+content.
+
+The snapshot may be restored to a KVS key if *--key=NAME* is used and the
+KVS service is running, or as a checkpoint in the content backing store
+if *--checkpoint* is used, without the KVS running.  One of those two options
+is required.
+
+
```
chu11, line 30: perhaps want to add (or add an example?) of how dump + restore garbage collects inactive references?  But on other hand it may not be super wise to advertise, as users may abuse it?

garlick:
Thanks for the additional comments @chu11!

On the two non-inlined ideas:

> would a --checkpoint-blobref option to reasonable for flux -dump? Like if a user just saved off the value after calling flux kvs getroot?

I had a `--blobref` option originally so it would be easy to add back in, but it seemed like any gainful use of it would require too much knowledge about how the KVS internals work, e.g. that you'd want a blobref that refers to a directory that you could get from a commit, or other tricky ways.  I thought the more user-friendly option would be `--key` (a directory name), but that was a little harder to implement, or at least I had a hard time figuring it out at the time.   So...maybe a follow on PR if a use case develops?

> Could the user restoring want the key to be a symlink? Or possibly in a different namespace? I'm not entirely sure if there's a realistic use case. I'm imagining like a user wants to re-start their instance from a specific point in time?

It definitely seems like dump/restore from/to a namespace root makes sense. I actually listed that above as something I thought we might want to add later.   I think a symlink can easily be added after the restore, if the idea there is to  add one that could be followed into another namespace from the primary.

Good ideas IMHO, but could be deferred to if/when a usecase develops?

t/t2807-dump-cmd.t
garlick: sorry, those comments weren't updated when the script changed!  I'll update and expand the comment so the casual reader won't be totally mystified!

doc/man1/flux-restore.rst
garlick: Oooh good idea.

doc/man1/flux-restore.rst
garlick: I think it may be best to hold off on that for a bit.  For now, the things that are omitted from the archive are described in `flux-dump(1)`.  If we're successful with using it for garbage collection, we could revisit.

garlick:
Pushed some fixups (squashed) for @chu11's comments.

The test for the kvs being loaded was a great idea - I was able to remove the CAVEATS section from `flux-restore(1)` now that it is automatically detected.

garlick:
Well something is not right here.  When I dump and restore my system at home, I can no longer run jobs.  I get the following error

```
$ flux mini run hostname
0.475s: job.exception type=exec severity=0 failed to create guest ns: No such file or directory
$ sudo flux dmesg
[snip]
2022-03-15T04:32:12.313805Z kvs.err[0]: content_load_completion: flux_content_load_get: No such file or directory
2022-03-15T04:32:12.317248Z job-exec.info[0]: job-exception: id=2317853860560896: failed to create guest ns: No such file or directory
```
So it looks like there is a dangling blobref there somewhere.

chu11 (approved):
updates look good, LGTM!

garlick:
Thanks @chu11.

Just pushed  some more fixups (squashed) after running down the problem above:
- add work around and test for #4222
- fixed minor typos in job-exec.c, content-sqlite.c
- flush content store after restoring, unless `--no-cache`
- slightly adjust progress output and make `--quiet` really quiet
- cherry-pick @grondo's commit to add libarchive to Dockerfiles and drop temporary `checks/Dockerfile` commit

grondo:
Sorry the fedora33 docker image build or push failed for unknown reason. I had to restart the build so it will be up to an hour before that is fixed.

PR #4209, opened 2022-03-10
python: add missing methods and improve efficiency of IDset class

grondo:
This PR was split off from some work on #4205.

The Python `IDset` class was implemented before the C libidset API exposed important set operations directly, and so the class converted IDsets to Python `set` objects and used that class for set operations. While we don't use Python because it is fast, it turns out that the libidset equivalent set operations (added since the original `IDset` class implementation, I think) are much faster (~11x faster).

This PR converts many of the `IDset` class methods to use the underlying C implementation. Also, a missing `difference` method is added, and some Python operators are overloaded to match the Python `set` conventions (`|/+` for union, `-` for difference, `&` for intersect).

Finally, the `IDset()` constructor is improved to allow more types (`int` now works to create an idset of one element, lists of integers can be out of order, and `IDset()` on an idset creates a copy)

The following times an intersection operation on an idset of size 1024 1000x:

```python
import time
from flux.idset import IDset
ids = IDset("0-1023")
t = time.time()
for i in range(1000):
    result = ids.intersect("100-200,300-400")
dt = time.time() - t
print(f"1000 intersect took {dt:<5.3f}s")
````

Before:
```
1000 intersect took 4.412s
```

After:
```
1000 intersect took 0.372s
```

`IDset` objects are iterable, so if a programmer prefers they can still do set operations the slow Python way by converting to a set with `set(idset)`.

### DIFF
--- a/src/bindings/python/flux/idset.py
+++ b/src/bindings/python/flux/idset.py
@@ -79,6 +79,8 @@ class IDset(WrapperPimpl):
             arg = str(arg)
         elif isinstance(arg, collections.Iterable) and not isinstance(arg, str):
             arg = ",".join(str(i) for i in sorted(arg))
+        elif isinstance(arg, int):
+            arg = str(arg)
 
         self.default_flags = flags
         try:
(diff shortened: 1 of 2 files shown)


garlick (approved):
Sounds like a good change to me!

I didn't spot anything to comment on looking over the commits.  Looks pretty contained and well tested.

grondo:
Thanks, setting MWP.

PR #4212, opened 2022-03-12
testsuite: ensure tests can run concurrently with --root=$FLUX_JOB_TMPDIR 

grondo:
A good stress test for the testsuite is to run multiple copies of every test across a flux instance via
```
$ flux mini bulksubmit --quiet --shuffle --watch --progress -o pty --cc=1-4  --job-name={./%} \
   sh -c './{} -d -v --root=$FLUX_JOB_TMPDIR' ::: t*.t python/t*.py lua/t*.t
```
This command submits 4 copies of every test in the testsuite as a job to the current instance, watching progress and output as the tests progress.

However, when attempting this on fluke, I ran into several issues, including:

 - Because the testsuite runs with `LANG=C` and `FLUX_JOB_TMPDIR` contains the jobid in f58 encoding, a few test scripts and one core flux Python module broke due to Unicode errors.
 - at least one test broke when `FLUX_JOB_CC` was set in the current environment
 - a couple tests used static files or ports outside of the trash directory, so could not tolerate being run concurrently
 - the python tests don't accept or ignore options like `-d, --debug` or `--root=PATH` and instead abort with unknown option errors

This PR fixes all the little issues above.

I was then able to grab a large allocation on fluke and run the above command (even using `--cc=1-10` to run 10 copies of every test) successfully. 

```
PD:0    R:0    CD:1699 F:1    │██████████████████████████████████│100.0% 0:01:48
$ flux jobs -f failed -o {name}
NAME
t0000-sharness
```

There are still a couple issues I hope to run down:

 - `t0000-sharness.t` intermittently fails in this mode
 -  Some Lua tests need to get the ability to run in a unique directory
 -  Some tests emit special pty characters when running, which messes up the top-level `--progress` meter

Therefore this is still a WIP

### DIFF
--- a/src/bindings/python/flux/util.py
+++ b/src/bindings/python/flux/util.py
@@ -88,9 +88,11 @@ def encode_payload(payload):
     if payload is None or payload == ffi.NULL:
         payload = ffi.NULL
     elif isinstance(payload, str):
-        payload = payload.encode("UTF-8")
+        payload = payload.encode("UTF-8", errors="surrogateescape")
     elif not isinstance(payload, bytes):
-        payload = json.dumps(payload, ensure_ascii=False).encode("UTF-8")
+        payload = json.dumps(payload, ensure_ascii=False).encode(
+            "UTF-8", errors="surrogateescape"
+        )
     return payload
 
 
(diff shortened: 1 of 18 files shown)


grondo:
I just pushed a few more fixes and now all tests appear to run successfully when concurrently, at least with the `flux mini bulksubmit`  command above and `--cc=1-12` in a large instance on fluke.

garlick (approved):
This all looks great!  Very cool way to shake out the flakey tests!

I just noted one commit message typo.

grondo:
Thanks! I fixed the typo and set MWP.

PR #4215, opened 2022-03-13
content-sqlite: check that file has rw permission

### LINKED ISSUE #4210: content-sqlite: lack of write permission on sqlite file causes hang
In some testing, I accidentally started flux (as user flux) with `content.sqlite` file readable but not writable.  rc1 got stuck, which wasn't detected by systemd.  I ran `systemctl stop flux` and it hung for a couple of minutes and ultimately probably kill -9'ed the broker.

Meanwhile the `content-sqlite` module logged:
```
2022-03-11T02:18:57.923470Z broker.debug[0]: insmod content-sqlite
2022-03-11T02:18:57.953269Z broker.debug[0]: content backing store: enabled content-sqlite
2022-03-11T02:18:57.981048Z broker.debug[0]: insmod kvs
2022-03-11T02:18:57.984055Z kvs.info[0]: restored kvs-primary from checkpoint on 2022-03-11T01:51:43
2022-03-11T02:18:58.006304Z broker.debug[0]: insmod kvs-watch
2022-03-11T02:18:58.020766Z content-sqlite.err[0]: store: executing stmt: attempt to write a readonly database(8)
2022-03-11T02:18:58.021244Z broker.err[0]: content store: Operation not permitted
2022-03-11T02:18:58.021436Z content-sqlite.err[0]: store: executing stmt: attempt to write a readonly database(8)
2022-03-11T02:18:58.021747Z broker.err[0]: content store: Operation not permitted
2022-03-11T02:18:58.021866Z content-sqlite.err[0]: store: executing stmt: attempt to write a readonly database(8)
2022-03-11T02:18:58.022181Z broker.err[0]: content store: Operation not permitted
```

Seems like that "Operation not permitted" error should be fatal.

garlick:
Problem: when restarting Flux with an existing content.sqlite
file, sqlite does not raise an error if the db is open read-write
but the file is not writable.  Instead, it fails store RPCs one
by one.

Add an access(2) call before opening the database with the sqlite3 API.

Fixes #4210

### DIFF
--- a/src/modules/content-sqlite/content-sqlite.c
+++ b/src/modules/content-sqlite/content-sqlite.c
@@ -13,6 +13,7 @@
 #if HAVE_CONFIG_H
 #include "config.h"
 #endif
+#include <unistd.h>
 #include <sqlite3.h>
 #include <lz4.h>
 #include <flux/core.h>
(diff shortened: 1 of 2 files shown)


grondo (approved):
LGTM!

garlick:
Thanks!

PR #4216, opened 2022-03-14
restrict access to content service used as KVS blob layer

garlick:
As discussed in #4210, it may be a good idea to drop content functions from the public API because they've not been found to be generally useful, and they now will have the caveat that content not linked to the KVS could evaporate on instance checkpoint/restart.   This PR makes the content functions private, and also limits access to content to the instance owner.

It is still possible to store arbitrary content from a job to the job's private KVS namespace, and thereby obtain access to the content store, but at least in that case the access is associated with a job and a user.  With content load + store open to guests, access was effectively anonymous.

I was going to move the  `load`, `store`, and `spam` subcommands from `flux content` to a test-only utility, but just moved `spam`.  My thought was that restricting access to only the instance owner made the existence of the load and store commands less of a foot gun?


grondo (approved):
This seems like a very wise change!

chu11 (approved):
LGTM, just saw two typos

garlick:
Rebased on current master and fixed the typos @chu11 found.  Also I left content.[ch] intact in libcontent rather than merging into libcontent-util.[ch] which makes the diff cleaner.

I'll go ahead and set MWP - thanks!

PR #4223, opened 2022-03-16
sched-simple: set expiration of jobs with no duration to instance lifetime

grondo:
This PR is pretty trivial. If there is no duration set in a jobspec, and the current instance has a nonzero `expiration`, then propagate that expiration to the _R_ generated for the job.

This change does _not_ check if a job with a nonzero duration would exceed the current instance lifetime. That will be in a separate PR.

### DIFF
--- a/src/modules/sched-simple/sched.c
+++ b/src/modules/sched-simple/sched.c
@@ -152,13 +152,20 @@ static struct simple_sched * simple_sched_create (void)
     return ss;
 }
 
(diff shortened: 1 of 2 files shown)


garlick (approved):
LGTM!

Looks like a missing HAVE_JQ is causing the new test to fail on the bionic builder.

grondo:
Thanks, I've set MWP.

grondo:
I had to tweak the testsuite test so it would work as a job under the inception tests with a single core. Force-pushed the result.

PR #4224, opened 2022-03-16
job-manager: add builtin job duration validator plugin

grondo:
There are several ways to validate that the requested time limit of a job (as set in the `duration` key of jobspec) will not exceed the current instance lifetime (as set in `R.execution.expiration`). The approach used here is to add  a builtin jobtap plugin which fetches `resource.R` and rejects jobs where the submitted `duration` would exceed any instance `expiration` as set in _R_.

Another possible approach would be to do the validation in the scheduler "feasibility" RPC. But the feasibility RPC is only optionally enabled, and enabling it by default causes lots of tests to break. Also the feasibility plugin is only able to reject jobs when the scheduler is loaded, leaving a gap when the scheduler is unloaded or being reloaded for some reason.

A job validator plugin could also be developed which would run alongside the jobspec validator and optional scheduler feasibility validator. However, each time this plugin starts up it would have to fetch `resource.R` to get the current instance expiration and since the validator script is started on demand I don't think we want to pay the extra Python startup cost.

So, the approach taken in this PR is to have a jobtap plugin, loaded by default, which will validate job duration so that we get that functionality in all instances of Flux, regardless of scheduler.

### DIFF
--- a/src/modules/job-manager/Makefile.am
+++ b/src/modules/job-manager/Makefile.am
@@ -65,7 +65,8 @@ libjob_manager_la_SOURCES = \
 	jobtap.c \
 	plugins/priority-default.c \
 	plugins/dependency-after.c \
-	plugins/begin-time.c
+	plugins/begin-time.c \
+	plugins/validate-duration.c
 
 fluxinclude_HEADERS = \
 	jobtap.h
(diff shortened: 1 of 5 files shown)


garlick (approved):
This seems like the most sensible design.   Nice clear error message for the user!
```
(s=1,d=1,builddir) garlick@system76-pc:~ flux mini run -t 60 hostname
flux-mini: ERROR: [Errno 22] job duration (1m) exceeds remaining instance lifetime (35s)
```

Just curious, and the KVS watch is fine as it is, but is there a scenario you foresee where `resource.R` would change?

grondo:
> Just curious, and the KVS watch is fine as it is, but is there a scenario you foresee where resource.R would change?

Well, this was mainly for testing for now, however this could occur if a running job gets its time limit updated. I assume at some point we'll need to do that and there will be a way to update _R_.

grondo:
Thanks! I've set MWP.

PR #4225, opened 2022-03-16
clarify descriptions in flux-dump(1), flux-restore(1)

garlick:
Problem: `flux-dump` and `flux-restore` are a matched pair of utilities but their man pages are somewhat disjoint.

Rework the descriptions so that the symmetry is more evident.

Fix a couple of typos and reword some awkward sentences.


chu11 (approved):
pretty straight forward cleanup, LGTM

Issue #4222, opened 2022-03-15
kvs should not assume empty directory exists in content store


garlick:
Problem: when garbage collecting the content store offline, as proposed in #258, commits to new namespaces fail due to a dangling blobref.

This was reported in #4208 as an inability to launch jobs after GC.

It can be reproduced more simply by creating a new namespace and committing a key there, e.g.
```
$ flux kvs namespace create testns
$ flux kvs put --namespace=testns a=b
flux-kvs: flux_kvs_commit: No such file or directory
$ flux dmesg
[snip]
kvs.err[0]: content_load_completion: flux_content_load_get: No such file or directory
```

The problem is resolved if `flux-restore` stores an unreferenced emtpy directory to the content store.

Probably the KVS should no longer assume that an empty directory is already stored just because the KVS necessarily started out empty at some point.  That assumption is subverted by the GC.

garlick:
I think your alternate idea is a pretty good quick fix.  Just do a `content_store()` on the empty directory when the KVS is loaded and put a nice comment above that code stating something like:
```c
/* Store an empty RFC 11 dir object.  The code for creating a new (empty)  kvs
 * namespace assumes that this object was stored when the primary namespace
 * was initially created, but offline garbage collection foiled that plan.
 * See also flux-framework/flux-core#4208
 */
```

PR #4229, opened 2022-03-16
kvs: always store empty directory object to content store

chu11:
Problem: The kvs has always assumed an empty directory treeobj
has been stored in the content store, because that is how the first
kvs root was originally initialized.  The introduction of kvs dump / restore
can prune non-referenced objects in the content store, such as the
empty directory treeobj.  This can lead to errors.

Solution: Always store an empty directory object into the content
store when loading the kvs module.

Fixes https://github.com/flux-framework/flux-core/issues/4222

### DIFF
--- a/src/cmd/builtin/restore.c
+++ b/src/cmd/builtin/restore.c
@@ -295,26 +295,6 @@ static json_t *restore_snapshot (struct archive *ar, flux_t *h)
     return rootref;
 }
 
-/* Work around #4222.
- */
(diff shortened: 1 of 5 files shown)


chu11:
offline discussion w/ @garlick, I was perhaps too clever with my changes.  I made it far simpler.  With the exception of an additional comment, the entirety of the change is simply moving when we call `store_initial_rootdir()`.

garlick (approved):
This is easier for me to grok, thank you!

PR #4231, opened 2022-03-17
broker: fine tune logging and enable core dump on SIGSEGV

garlick:


src/broker/broker.c
```diff
@@ -217,10 +217,13 @@ int main (int argc, char *argv[])
...
     parse_command_line_arguments (argc, argv, &ctx);
 
-    /* Block all signals, saving old mask and actions for SIGINT, SIGTERM.
+    /* Block all signals but SIGSEGV, SIGFPE.
+     * Sae old mask and actions for SIGINT, SIGTERM.
      */
     sigset_t sigmask;
     sigfillset (&sigmask);
+    sigdelset (&sigmask, SIGSEGV);
+    sigdelset (&sigmask, SIGFPE);
```
grondo, line 226: `SIGQUIT`, `SIGILL`, and `SIGABRT` are also listed as generating a coredump. I can see an argument for blocking `SIGQUIT`, but should `SIGILL` and `SIGABRT` be removed from the signal set here as well, unlikely though their delivery may be in the short time between here and setting up the real signal handlers?

src/broker/broker.c
```diff
@@ -217,10 +217,13 @@ int main (int argc, char *argv[])
 
     parse_command_line_arguments (argc, argv, &ctx);
 
-    /* Block all signals, saving old mask and actions for SIGINT, SIGTERM.
+    /* Block all signals but SIGSEGV, SIGFPE.
+     * Sae old mask and actions for SIGINT, SIGTERM.
```
grondo, line 221: Typo in comment, `Sae`->`Save`

grondo (approved):
LGTM!

A typo and one question inline, but feel free to ignore the question as it doesn't seem critical to me.

src/broker/broker.c
garlick: The old signal mask doesn't get restored until after the reactor exits so I think those are actually blocked for most of the time the broker is running.  Wow, I'm not sure what I was thinking.  Unblocking those as well.

grondo: I forgot these as well
```
       SIGBUS      10,7,10     Core    Bus error (bad memory access)
       SIGSYS      12,31,12    Core    Bad system call (SVr4);
                                       see also seccomp(2)
       SIGTRAP        5        Core    Trace/breakpoint trap
       SIGXCPU     24,24,30    Core    CPU time limit exceeded (4.2BSD);
                                       see setrlimit(2)
       SIGXFSZ     25,25,31    Core    File size limit exceeded (4.2BSD);
                                       see setrlimit(2)
```

grondo: Sounds good, feel free to set MWP as far as I'm concerned.

garlick:
Thanks - setting MWP

PR #4235, opened 2022-03-21
libjob: improve flux_job_statetostr(), flux_job_resulttostr() interface

garlick:
This is a small "upgrade" to the interfaces for converting job states to strings that fell out of some other work I was doing.
```c
/* Convert state to string.  'fmt' may be:
 * "s" - lower case short name
 * "S" - upper case short name
 * "l" - lower case long name
 * "L" - upper case long name
 * This function always returns a valid string, though it may
 * be something like "(unknown)".
 */
const char *flux_job_statetostr (flux_job_state_t state, const char *fmt);
```
Identical change here, just to keep interfaces similar
```c
const char *flux_job_resulttostr (flux_job_result_t result, const char *fmt);
```

### DIFF
--- a/src/common/libjob/strtab.c
+++ b/src/common/libjob/strtab.c
@@ -1,0 +1,63 @@
+/************************************************************\
+ * Copyright 2022 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+#include <flux/core.h>
+
+#include "strtab.h"
+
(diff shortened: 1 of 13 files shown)


src/common/libjob/strtab.c
```diff
@@ -0,0 +1,63 @@
...
+        case 's':
+            return entry->short_lower;
+        case 'S':
+            return entry->short_upper;
+        case 'l':
+            return entry->long_lower;
+        case 'L':
+        default:
+            return entry->long_upper;
+    }
```
chu11, line 30: curious why default to long upper when there is bad input from the user?  Why not EINVAL or "unknown"?

chu11 (approved):
overall LGTM, just one question below

src/common/libjob/strtab.c
garlick: Since these are often used to build printf style arguments, it's inconvenient to have to check for error, so I was just thinking having a default was better.  Also it's consistent with the behavior that was there before when an unknown state was passed.

garlick:
Thanks, setting MWP

PR #4236, opened 2022-03-22
add support for resource properties

grondo:
This PR adds support for resource properties as specified in [RFC 20](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_20.html), along with basic support for [RFC 31](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_31.html) constraints.

Thus far, most of the groundwork has been completed, so I thought I'd post now to allow @dongahn an early comment and synchronization period. Still TODO is to add support for constructing RFC 31 constraints in jobspec from the `flux mini` command line tools (tentative support for selecting resources based on constraints is in sched-simple already), along with associated testing.

In summary, this PR contains the following enhancements:
 - support for RFC 20 `properties` key on encode and decode of Rv1 objects
 - new property-specific options to `flux R encode/decode`, along with a new `flux R set-property` command
 - a simple, librlist-based constraint matching function to support "filtering" resources with an RFC 31 constraint object
 - initial proof-of-concept change to sched-simple to support  `attributes.system.constraints` set in jobspec.
   
While RFC 31 only requires support for the `properties` operation in a job constraints object, it was simple enough to add basic logical operation support for `and`, `or`, and `not`, so that is included here.

In this implementation, all ranks inherit a default property of their hostname. This allows us to get support for excluding hosts via the `^hostname` property (though hostlist isn't supported). In retrospect, maybe it would be better to support hostnames as another kind of constraint (which could support hostlists).

You can experiment with this branch in a test instance by using `flux R set-property` to set some properties, e.g.:

```
 $ flux start -s 4
 $ flux kvs put resource.R="$(flux kvs get resource.R | flux R set-property foo:2-3)"
 $ flux module reload resource && flux module load sched-simple
2022-03-22T21:28:03.578409Z sched-simple.err[0]: exiting due to resource update failure: the resource module was unloaded
 $ flux mini run flux getattr rank
0
 $ flux mini run --setattr=system.constraints.properties='["foo"]' flux getattr rank
2

```

### DIFF
--- a/src/bindings/python/flux/job/Jobspec.py
+++ b/src/bindings/python/flux/job/Jobspec.py
@@ -66,6 +66,33 @@ def _validate_dependency(dep):
         raise TypeError("dependency value must be a string")
 
 
+def _validate_property_query(name):
+    invalid_chars = set("&'\"`|()")
+    if any((x in invalid_chars) for x in name):
+        raise TypeError(f"invalid character in property '{name}'")
+
+
+def _validate_constraint_op(operator, args):
+    if not isinstance(operator, str):
+        raise TypeError(f"constraint operation {operator} is not a string")
+    if not isinstance(args, abc.Sequence):
+        raise TypeError(f"argument to constraint {operator} must be a sequence")
+    if operator in ["and", "or", "not"]:
+        for constraint in args:
+            _validate_constraint(constraint)
+    elif operator in ["properties"]:
+        for name in args:
+            _validate_property_query(name)
+
+
+def _validate_constraint(constraints):
+    """Validate RFC 31 Constraint object"""
+    if not isinstance(constraints, abc.Mapping):
+        raise TypeError("constraints must be a mapping")
+    for operator, arg in constraints.items():
+        _validate_constraint_op(operator, arg)
+
+
 def validate_jobspec(jobspec, require_version=None):
     """
     Validates the jobspec by attempting to construct a Jobspec object.  If no
@@ -277,6 +304,8 @@ class Jobspec(object):
                 raise TypeError("attributes.system.dependencies must be a list")
             for dependency in system["dependencies"]:
                 _validate_dependency(dependency)
+        if "constraints" in system:
+            _validate_constraint(system["constraints"])
 
     @staticmethod
     def _create_resource(res_type, count, with_child=None):
--- a/src/common/librlist/rlist.c
+++ b/src/common/librlist/rlist.c
@@ -25,6 +25,7 @@
 #include "src/common/libutil/errprintf.h"
 
 #include "rnode.h"
+#include "match.h"
 #include "rlist.h"
 #include "rhwloc.h"
 
@@ -303,6 +304,34 @@ err:
     rlist_destroy (result);
     return NULL;
 }
+
+struct rlist *rlist_copy_constraint (const struct rlist *orig,
+                                     json_t *constraint,
+                                     flux_error_t *errp)
+{
+    if (rnode_match_validate (constraint, errp) < 0)
+        return NULL;
+    return rlist_copy_internal (orig,
+                                (rnode_copy_f) rnode_copy_match,
+                                (void *) constraint);
+}
+
+struct rlist *rlist_copy_constraint_string (const struct rlist *orig,
+                                            const char *constraint,
+                                            flux_error_t *errp)
+{
+    struct rlist *rl;
+    json_error_t error;
+    json_t *o = json_loads (constraint, 0, &error);
+    if (!o) {
+        errprintf (errp, "%s", error.text);
+        return NULL;
+    }
+    rl = rlist_copy_constraint (orig, o, errp);
+    json_decref (o);
+    return rl;
+}
+
 
 int rlist_remove_ranks (struct rlist *rl, struct idset *ranks)
 {
--- a/src/common/librlist/rlist.c
+++ b/src/common/librlist/rlist.c
@@ -882,6 +882,160 @@ static int rlist_assign_nodelist (struct rlist *rl, json_t *nodelist)
     return rc;
 }
 
+static void property_destructor (void **arg)
+{
+    if (arg) {
+        idset_destroy (*(struct idset **) arg);
+        *arg = NULL;
+    }
+}
+
+static char * property_string_invalid (const char *s)
+{
+    return strpbrk (s, "^&'\"`|()");
+}
+
+int rlist_add_property (struct rlist *rl,
+                        flux_error_t *errp,
+                        const char *name,
+                        const char *targets)
+{
+    unsigned int i;
+    int count;
+    struct rnode *n;
+    struct idset *ids = NULL;
+    struct idset *unknown = NULL;
+    int rc = -1;
+    char *p;
+
+    if ((p = property_string_invalid (name))) {
+        errprintf (errp,
+                   "Invalid character '%c' in property \"%s\"",
+                   *p,
+                   name);
+        errno = EINVAL;
+        goto out;
+    }
+
+    if (!targets || !(ids = idset_decode (targets))) {
+        errprintf (errp,
+                   "Invalid idset string '%s'",
+                   targets ? targets : "(null)");
+        errno = EINVAL;
+        goto out;
+    }
+
+    /*  Check for invalid ranks first, so we can fail early before
+     *   applying any properties.
+     */
+    if (!(unknown = idset_create (0, IDSET_FLAG_AUTOGROW))) {
+        errprintf (errp, "Out of memory");
+        goto out;
+    }
+    i = idset_first (ids);
+    while (i != IDSET_INVALID_ID) {
+        if (!rlist_find_rank (rl, i)
+            && idset_set (unknown, i) < 0) {
+            errprintf (errp, "unknown rank %u", i);
+            errno = ENOENT;
+            goto out;
+        }
+        i = idset_next (ids, i);
+    }
+    if ((count = idset_count (unknown)) > 0) {
+        p = idset_encode (unknown, IDSET_FLAG_RANGE);
+        errprintf (errp,
+                   "%s%s not found in target resource list",
+                   p ? (count == 1 ? "rank " : "ranks ") : "some ranks",
+                   p ? p : "");
+        free (p);
+        errno = ENOENT;
+        goto out;
+    }
+
+    i = idset_first (ids);
+    while (i != IDSET_INVALID_ID) {
+        if ((n = rlist_find_rank (rl, i)))
+            if (rnode_set_property (n, name) < 0) {
+                errprintf (errp,
+                           "Failed to set property %s on rank %u",
+                           name,
+                           i);
+                goto out;
+            }
+        i = idset_next (ids, i);
+    }
+    rc = 0;
+out:
+    idset_destroy (ids);
+    idset_destroy (unknown);
+    return rc;
+}
+
+int rlist_assign_properties (struct rlist *rl,
+                             json_t *properties,
+                             flux_error_t *errp)
+{
+    json_t *val;
+    const char *name;
+    char *p;
+    int rc = -1;
+
+    if (!rl || !properties) {
+        errprintf (errp, "Invalid argument");
+        errno = EINVAL;
+        return -1;
+    }
+
+    if (!json_is_object (properties)) {
+        errprintf (errp, "properties must be an object");
+        errno = EINVAL;
+        return -1;
+    }
+
+    /*  First, iterate all property inputs to ensure they are valid.
+     *  This avoids the need to undo any applied properties on failure.
+     */
+    json_object_foreach (properties, name, val) {
+        struct idset *ids;
+        if (!json_is_string (val)) {
+            char *s = json_dumps (val, JSON_COMPACT | JSON_ENCODE_ANY);
+            errprintf (errp, "properties value '%s' not a string", s);
+            free (s);
+            errno = EINVAL;
+            goto error;
+        }
+        if ((p = property_string_invalid (name))) {
+            errprintf (errp,
+                       "invalid character '%c' in property \"%s\"",
+                        *p,
+                        name);
+            errno = EINVAL;
+            goto error;
+        }
+        if (!(ids = idset_decode (json_string_value (val)))) {
+            errprintf (errp,
+                       "invalid idset '%s' specified for property \"%s\"",
+                       json_string_value (val),
+                       name);
+            errno = EINVAL;
+            goto error;
+        }
+        idset_destroy (ids);
+    }
+
+    json_object_foreach (properties, name, val) {
+        /*  Note: validity of json_string_value (val)
+         *   checked above, no need to do it again here.
+         */
+        if (rlist_add_property (rl, errp, name, json_string_value (val)) < 0)
+            goto error;
+    }
+    rc = 0;
+error:
+    return rc;
+}
+
 struct rlist *rlist_from_json (json_t *o, json_error_t *errp)
 {
     int i, version;
@@ -890,16 +1044,19 @@ struct rlist *rlist_from_json (json_t *o, json_error_t *errp)
     json_t *R_lite = NULL;
     json_t *nodelist = NULL;
     json_t *scheduling = NULL;
+    json_t *properties = NULL;
     double starttime = -1.;
     double expiration = -1.;
+    flux_error_t error;
 
     if (json_unpack_ex (o, errp, 0,
-                        "{s:i s?O s:{s:o s?o s?F s?F}}",
+                        "{s:i s?O s:{s:o s?o s?o s?F s?F}}",
                         "version", &version,
                         "scheduling", &scheduling,
                         "execution",
                           "R_lite", &R_lite,
                           "nodelist", &nodelist,
+                          "properties", &properties,
                           "starttime", &starttime,
                           "expiration", &expiration) < 0)
         goto err;
@@ -927,6 +1084,11 @@ struct rlist *rlist_from_json (json_t *o, json_error_t *errp)
     }
     if (nodelist && rlist_assign_nodelist (rl, nodelist) < 0)
         goto err;
+    if (properties && rlist_assign_properties (rl, properties, &error) < 0) {
+        if (errp)
+            snprintf (errp->text, sizeof (errp->text), "%s", error.text);
+        goto err;
+    }
     return (rl);
 err:
     rlist_destroy (rl);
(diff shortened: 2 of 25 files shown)


src/common/librlist/rlist.c
```diff
@@ -890,16 +1043,19 @@ struct rlist *rlist_from_json (json_t *o, json_error_t *errp)
...
     json_t *nodelist = NULL;
     json_t *scheduling = NULL;
+    json_t *properties = NULL;
     double starttime = -1.;
     double expiration = -1.;
+    flux_error_t error;
 
     if (json_unpack_ex (o, errp, 0,
-                        "{s:i s?O s:{s:o s?o s?F s?F}}",
+                        "{s:i s?O s:{s:o s?o s?o s?F s?F}}",
```
dongahn, line 1097: `nodelist` is mandatory and must exist?

src/common/librlist/rlist.c
```diff
@@ -939,7 +1102,8 @@ struct rlist *rlist_from_R (const char *s)
     struct rlist *rl = NULL;
     json_t *o = json_loads (s, 0, &err);
     if (o)
-        rl = rlist_from_json (o, &err);
+        if (!(rl = rlist_from_json (o, &err)))
+            fprintf (stderr, "%s", err.text);
```
dongahn, line 1151: Just checking... do we really want this to go `stderr`?

src/common/librlist/rlist.c
```diff
@@ -1271,17 +1435,117 @@ static json_t *rlist_json_nodelist (struct rlist *rl)
...
+    json_t *result = NULL;
+    struct idset *ids;
+
+    if (!(properties = rlist_properties (rl)))
+        return NULL;
+
+    /*  Do not bother returning an empty JSON object
+     */
+    if (zhashx_size (properties) == 0)
+        goto out;
```
dongahn, line 1546: Do we need `errno = EINVAL`?

src/common/librlist/rlist.c
```diff
@@ -1271,17 +1435,117 @@ static json_t *rlist_json_nodelist (struct rlist *rl)
...
+        errno = ENOMEM;
+        goto out;
+    }
+    ids = zhashx_first (properties);
+    while (ids) {
+        const char *name = zhashx_cursor (properties);
+        char *s = idset_encode (ids, IDSET_FLAG_RANGE);
+        json_t *val;
+        if (!s || !(val = json_string (s))) {
+            free (s);
```
dongahn, line 1558: `free` can clobber `errno`?

src/common/librlist/rlist.c
```diff
@@ -1271,17 +1435,117 @@ static json_t *rlist_json_nodelist (struct rlist *rl)
...
+            goto out;
+        }
+        if (json_object_set_new (o, name, val) < 0) {
+            free (s);
+            json_decref (val);
```
dongahn, line 1563: `free` and `json_decref` can clobber `errno`? Better to set `errno = ENOMEM`?

src/common/librlist/rlist.c
```diff
@@ -304,6 +305,34 @@ struct rlist * rlist_copy_ranks (const struct rlist *rl, struct idset *ranks)
...
+
+struct rlist *rlist_copy_constraint_string (const struct rlist *orig,
+                                            const char *constraint,
+                                            flux_error_t *errp)
+{
+    struct rlist *rl;
+    json_error_t error;
+    json_t *o = json_loads (constraint, 0, &error);
+    if (!o) {
+        errprintf (errp, "%s", error.text);
```
dongahn, line 327: `errno = ENOMEM`?

src/bindings/python/flux/job/Jobspec.py
```diff
@@ -66,6 +66,33 @@ def _validate_dependency(dep):
         raise TypeError("dependency value must be a string")
 
 
+def _validate_property_query(name):
```
dongahn, line 69: Commit: erorrs -> errors

dongahn (changes requested):
@grondo: this looks great! I just added a few change requests and other comments/questions.

dongahn:
@grondo: Just one high level question. If you set a scheduler specific property like `amd-mi50@gpu : 0-1` and  users specified, `--setattr=system.constraints.properties='["amd-mi50"]'`, will `sched-simple` be able to match this with rank 0 or 1? Or does the specification must be `--setattr=system.constraints.properties='["amd-mi50@gpu"]'`?

dongahn:
Also do we need to document `attributes.system.constraints` in both Canonical Jbspec RFC and V2 Jobspec RFC?

grondo:
Like flux-framework/rfc#318?

I think since `attributes.system.constraints` is defined in RFC 14 it is covered by 

> Common system keys are listed below, and their definitions can be found in RFC14. 

in RFC 25?

grondo:
>  If you set a scheduler specific property like amd-mi50@gpu : 0-1 and users specified, --setattr=system.constraints.properties='["amd-mi50"]', will sched-simple be able to match this with rank 0 or 1? Or does the specification must be --setattr=system.constraints.properties='["amd-mi50@gpu"]'?

It is the latter case, all of flux-core including sched-simple only know about per rank/execution target properties. (i.e. the same applies to `flux R decode --property=NAME`, you must specify `amd-mi50@gpu`, the `@gpu` is reserved for scheduler use but is currently not stripped from the property name assigned to the rank(s))

dongahn:
>the @gpu is reserved for scheduler use but is currently not stripped from the property name assigned to the rank(s))

Yeah I thought so but I wanted to double check. 

Since `@gpu` is scheduler-specific per RFC 20, I think it would be a bit better if each scheduler treats only the property string that appears before `@` as the actual property name. And it is up to each scheduler to do what it wants to do with the suffix string - it could use it for an optimization and/or for structuring the code for future extension.

Then `--setattr=system.constraints.properties='["amd-mi50"]'` should be how to specify the property?

dongahn:
>Like https://github.com/flux-framework/rfc/pull/318?

> I think since attributes.system.constraints is defined in RFC 14 it is covered by

>Common system keys are listed below, and their definitions can be found in RFC14.

Ah yes. This is great. But I would like it a bit better if RFC 31 is listed in the Related standard section and perhaps have "for example, "properties". I did keyword search on RFC14 and couldn't find anything and mistakenly thought that is not covered -- and I was wrong ;-)

> in RFC 25?

Yup.

src/common/librlist/rlist.c
```diff
@@ -882,6 +882,159 @@ static int rlist_assign_nodelist (struct rlist *rl, json_t *nodelist)
...
+        return -1;
+    }
+
+    /*  First, iterate all property inputs to ensure they are valid.
+     *  This avoids the need to undo any applied properties on failure.
+     */
+    json_object_foreach (properties, name, val) {
+        struct idset *ids;
+        if (!json_is_string (val)) {
+            char *s = json_dumps (val, JSON_COMPACT | JSON_ENCODE_ANY);
```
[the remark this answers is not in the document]
grondo: Sorry, my working copy changed (sorry this was marked WIP still) and this particular code is no longer there.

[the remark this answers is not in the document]
grondo: Oh, actually it just moved. I don't think there's much we can do if s is NULL here. Most `sprintf` implementations seem to substitute `(null)` when the `char *` is `NULL`, I could add explicit handling for this if desired? In any case I think we want the error to be "properties value must be a string" and not "Out of memory"

src/common/librlist/rlist.c
```diff
@@ -882,6 +882,159 @@ static int rlist_assign_nodelist (struct rlist *rl, json_t *nodelist)
...
+                       name);
+            errno = EINVAL;
+            goto error;
+        }
+        idset_destroy (ids);
+    }
+
+    json_object_foreach (properties, name, val) {
+        const char *s = json_string_value (val);
+        if (!s || rlist_add_property (rl, errp, name, s) < 0)
```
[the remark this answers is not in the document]
grondo: Thanks, this code was present before the extra validation was added for enhanced error reporting. I can remove the redundant check so it doesn't cause confusion.

src/common/librlist/rlist.c
grondo: yes, from [RFC 20](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_20.html)

> The value of the execution key SHALL contain at least the keys R_lite, and nodelist, with optional keys properties, starttime and expiration. Other keys are reserved for future extensions.

I don't think the `nodelist` requirement changed in this PR?

src/common/librlist/rlist.c
grondo: This isn't an error condition, so I don't think so.

src/common/librlist/rlist.c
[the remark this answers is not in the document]
dongahn: I was thinking the major case of `NULL` return would be out of memory in which case we return to the caller right away. But `json_dumps` can return `NULL` on other errors so this code is fine as is.

src/common/librlist/rlist.c
dongahn: Yes, this PR doesn't change `nodelist` requirement so it is okay not to change. I just noticed the the unpack was treating 'nodelist` as optional.

src/common/librlist/rlist.c
grondo: Hm, currently errors from this function (and `rlist_json_nodelist()`) are ignored. I think I'll tack a commit on the end to fix that (and fix errno clobbering)

grondo:
> Since @gpu is scheduler-specific per RFC 20, I think it would be a bit better if each scheduler treats only the property string that appears before @ as the actual property name. And it is up to each scheduler to do what it wants to do with the suffix string - it could use it for an optimization and/or for structuring the code for future extension.

> Then `--setattr=system.constraints.properties='["amd-mi50"]'` should be how to specify the property?

The constraints are satisfied by the scheduler, so this is how it could work for Fluxion, but you are suggesting in flux-core everything after `@` should be dropped? That feels like it might be a bit surprising to users and a little fussy to implement? Is there some reason properties can't be treated literally in flux-core and then matched using scheduler-specific part in specific schedulers?

I guess I could be convinced, but the implementation I have to stash properties in a hash will have to be replaced with something else that can tolerate this other scheme where `@*` suffix must be preserved in Rv1 but ignored when applying constraints.

grondo:
> This is great. But I would like it a bit better if RFC 31 is listed in the Related standard section and perhaps have "for example, "properties". I did keyword search on RFC14 and couldn't find anything and mistakenly thought that is not covered -- and I was wrong ;-)

If you have a chance, can you open an issue in the rfc repo so I don't forget to do that? Thanks :slightly_smiling_face:

src/common/librlist/rlist.c
grondo: Ah! ok I see. It might be that works out well for testing, but we can look at improving it to required in the future.

src/common/librlist/rlist.c
grondo: That's just some leftover debug. Will remove.

dongahn:
>The constraints are satisfied by the scheduler, so this is how it could work for Fluxion, but you are suggesting in flux-core everything after @ should be dropped? That feels like it might be a bit surprising to users and a little fussy to implement? Is there some reason properties can't be treated literally in flux-core and then matched using scheduler-specific part in specific schedulers?

No, I think the string can be treated literally alright. I was just looking at the RFC:

>Additionally, the @ character is reserved for scheduler specific property use. Any suffix that appears after the @ character in a property name SHALL be a scheduler-specific string. For example, amd-mi50@gpu, amd-mi50 SHALL be the property string, but a scheduler MAY use the gpu suffix to perform scheduling optimization for gpus of the corresponding ranks.

It says the generic string SHALL be the property string, not the whole thing. Maybe I was the one who suggested the prose and added ambiguity. 

>I guess I could be convinced, but the implementation I have to stash properties in a hash will have to be replaced with something else that can tolerate this other scheme where @* suffix must be preserved in Rv1 but ignored when applying constraints.

Ah I didn't think about this complexity. From fluxion point of view, once `scheduling` key is populated the property to right graph vertices, this info will be propagated into nested instanced through the graph so this isn't too much hassle. (IOW, rv1 for a nested job only needs to have the generic piece). But for the case you have to propagate the property through execution key only, this does add complexity. W/ that, it makes sense to treat the whole thing as the property string. (We may still need a revision to RFC20 though)?

dongahn:
>If you have a chance, can you open an issue in the rfc repo so I don't forget to do that? Thanks 

https://github.com/flux-framework/rfc/issues/319

grondo:
> It says the generic string SHALL be the property string, not the whole thing. Maybe I was the one who suggested the prose and added ambiguity.

Ah, ok. I apologize as I didn't quite internalize that, but I think you are right, currently the implementation doesn't match the wording in the RFC.

If it helps, I was thinking of properties with a "scheduler-specific part" as still having the literal property, i.e. `amd-mi50@gpu` applying to the execution target, but the scheduler MAY use the scheduler-specific part to apply the property to individual resources under that execution target. For example, in this case `amd-mi50@gpu` would be the literal property of the execution target, but `amd-mi50` might be applied by the scheduler to the gpu(s) of the execution target. Thus with sched-simple a user may specify `amd-mi50@gpu` in property constraints to restrict to ranks with this property, but on an advanced scheduler perhaps `amd-mi50` property will be supported and will match specifically the GPU.

I'm sorry if I misinterpreted the use case here.

grondo:
>  (We may still need a revision to RFC20 though)?

@dongahn, please check out flux-framework/rfc#320 and see if the new description is less ambiguous. I'm afraid I'm still not certain of the use case you envision for the scheduler-specific strings, so let me know if I've missed the mark.

dongahn:
> I'm afraid I'm still not certain of the use case you envision for the scheduler-specific strings, so let me know if I've missed the mark.

Thanks. Essentially I wanted to see if this can be used to explore support fine grain property attachment as a future extension. I can imagine a future system having different types of resources of the same kind like different core types (e.g., xeon cores on a socket and arm-v9 cores on a GPU device). One way to allow users to select a specific type is by doing a property string machine at the fine-grained vertex class.

dongahn:
>Ok, thanks. That is similar to what I was thinking in that you could perhaps specify properties like arm-a15@core[0-3] and arm-a7@core[4-7].

Exactly!

grondo:
Ok, I've updated this branch with full end-to-end support and testing of resource properties.
Updates from the previous checkpoint include:

 * a `flux-mini --requires` option to specify property based constraints. This option basically just sets `attributes.system.constraints.properties` to the array of property names set via `--requires=NAME`. So only a set of properties with logical _and_ are supported. (though exclusion of a property can be specified by prefixing a property name with `^`)
 * a new sharness test `t2276-job-requires.t` which reloads the scheduler with some properties specified and esnures that job requests with constraints are assigned to expected ranks.
 * I've also addressed all @dongahn's comments (and squashed result)

grondo:
I've removed `WIP` since this PR is now pretty much feature complete (pending changes and fixes based on any feedback)

garlick:
I'll give this a quick review pass now.

src/common/librlist/rlist.c
```diff
@@ -927,6 +1085,13 @@ struct rlist *rlist_from_json (json_t *o, json_error_t *errp)
     }
     if (nodelist && rlist_assign_nodelist (rl, nodelist) < 0)
         goto err;
+    if (properties && rlist_assign_properties (rl, properties, &error) < 0) {
+        if (errp) {
+            snprintf (errp->text, sizeof (errp->text), "%s", error.text);
+            errp->text[sizeof (errp->text) - 1] = '\0';
```
garlick, line 1136: snprintf already guarantees that  `errp->text` will be null terminated.

src/common/librlist/rlist.c
```diff
@@ -1271,17 +1436,117 @@ static json_t *rlist_json_nodelist (struct rlist *rl)
...
+            keys = NULL;
+        }
+        n = zlistx_next (rl->nodes);
+    }
+
+    return properties;
+error:
+    saved_errno = errno;
+    zhashx_destroy (&properties);
+    if (keys)
```
garlick, line 1526: test for non-NULL argument is not needed

src/common/librlist/rlist.c
```diff
@@ -1271,17 +1436,117 @@ static json_t *rlist_json_nodelist (struct rlist *rl)
...
+    if (!(o = json_object ())) {
+        errno = ENOMEM;
+        goto out;
+    }
+    ids = zhashx_first (properties);
+    while (ids) {
+        const char *name = zhashx_cursor (properties);
+        char *s = idset_encode (ids, IDSET_FLAG_RANGE);
+        json_t *val;
+        if (!s || !(val = json_string (s))) {
```
garlick: set `errno = ENOMEM` before `goto out`

src/common/librlist/rlist.c
```diff
@@ -1271,17 +1436,117 @@ static json_t *rlist_json_nodelist (struct rlist *rl)
...
+            free (s);
+            goto out;
+        }
+        if (json_object_set_new (o, name, val) < 0) {
+            free (s);
+            json_decref (val);
+            goto out;
+        }
```
garlick, line 1566: need to set `errno = ENOMEM`  before `goto out`

garlick (approved):
Nice!  This is a great capability.

I had a couple of nit picky comments mostly on error paths.  Actually I started having a feeling of deja vu later in my commit by commit review and was wondering if you might have fixed a couple of things I called out in commits that came before - however, I didn't see my comments inline so I left them in?  Ignore those comments if you already addressed them.

grondo:
>  Actually I started having a feeling of deja vu later in my commit by commit review and was wondering if you might have fixed a couple of things I called out in commits that came before 

No, that's because I keep making the same dumb mistakes! :disappointed:

src/common/librlist/rnode.c
```diff
@@ -40,6 +41,8 @@ void rnode_destroy (struct rnode *n)
         int saved_errno = errno;
         free (n->hostname);
         zhashx_destroy (&n->children);
+        if (n->properties)
+            zhashx_destroy (&n->properties);
```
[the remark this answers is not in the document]
grondo: Ugh, why can I never get that straight?

src/common/librlist/rlist.c
```diff
@@ -882,6 +882,161 @@ static int rlist_assign_nodelist (struct rlist *rl, json_t *nodelist)
...
+        goto out;
+    }
+
+    if (!targets || !(ids = idset_decode (targets))) {
+        errprintf (errp,
+                   "Invalid idset string '%s'",
+                   targets);
+        errno = EINVAL;
+        goto out;
+    }
```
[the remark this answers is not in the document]
grondo: Ok, easily addressed, thanks.

src/common/librlist/rlist.c
grondo: I _think_ this was fixed by 555ff3251e4918cfe8170c31190a134ed698dec3, but the line numbers referenced here don't match what is in the actual branch so maybe I'm confused...

src/common/librlist/rlist.c
grondo: same as above

src/common/librlist/rlist.c
garlick: Yeah that's what was confusing me too.  I think we're good.

grondo: This may be my fault. I introduced the bug in one commit and fixed in another. However, in this particular case that was cleaner because initially errors were ignored in the calling function. The follow-up commit fixed that along with ignored errors from the similar function that generates the `nodelist` key. Sorry about that! I'll try to refrain from doing that in the future.

grondo:
I've addressed all @garlick's comments (Thanks!) and force-pushed.

garlick:
Thanks!  I'm for MWP if you are.

grondo:
I think I need to wait for @dongahn to clear his request for changes. I'll try out his Fluxion properties branch first as well to make sure no further changes are required.

dongahn (approved):
LGTM. I also tested this to some extend w/ my branch. All good.

grondo:
Ok, rebased and I'll set MWP.

Issue #3895, opened 2021-10-02
A mechanism to gracefully shut down a Flux instance


dongahn:
One of LLNL major workflow teams runs instances with `[s|js]run ... flux start sleep inf` one on Lassen for ML and the other on Ruby for simulations. And then uses a domain specific workflow adapter to submit workflow jobs dynamically into those instances for their producer-consumer workflow. 

They are currently looking for a good way to gracefully shut down flux when all molecules are processed. Without this, this workflow can lead to some idling of resources. 

I gave them the following scripts for now. But this is a kludge at best and it seems a flux shutdown command can go a long way for them to do this in a portable way.

```bash
#! /bin/bash
#set -x

declare -r prog=${0##*/}

die() { echo -e "${prog}:" "$@"; exit 1; }

if [[ $# -ne 1 ]]
then
    echo "Usage: kill_gracefully JOBID"
    exit 1
fi

if [[ ${SYS_TYPE} = "blueos_3_ppc64le_ib_p9" ]]
then
    output=$(/usr/tcetmp/bin/attach-helper $1)
    if [[ $? -ne 0 ]]
    then
        die "attach-helper failed: wrong JOBID?"
    fi

    head_node=$(echo ${output} | awk -F ":" '{print $1}')
    launcher_pid=$(echo ${output} | awk -F ":" '{print $2}')
    ssh ${head_node} "kill ${launcher_pid}"
elif [[ ${SYS_TYPE} = "toss_3_x86_64_ib" ]]
then
    squeue -j $1 -tr -o "%.7i %B" > /dev/null
    if [[ $? -ne 0 ]]
    then
        die "squeue report failed: wrong JOBID?"
    fi
    output=$(squeue -j $1 -tr -o "%.7i %B" | tail -n1)
    head_node=$(echo ${output} | awk '{print $2}')
    sleep_pid=$(rsh ${head_node} pgrep sleep)
    rsh ${head_node} kill ${sleep_pid}
else
    die "unsupported SYS_TYPE"
fi
```

garlick:
That's a good idea.  The current way for a sys admin to shut down the system instance is to tell systemd to stop the rank 0 broker, which may be less obvious than a nice imperative command, particularly since the effect of doing it on a rank > 0 is to shutdown just the subtree rooted at that rank, which people might find counterintuitive?

[...]

PR #4237, opened 2022-03-24
flux-dmesg: add --new option, plus logging cleanup

garlick:
Problem: `flux-dmesg(1)` (and associated RPC service) can "follow" broker log messages, but always requires the entire ring buffer to be dumped first.

This PR adds a flag to the RPC to skip dumping the ring buffer, and then adds the `--new` option to `flux-dmesg(1)`:
```
Usage: flux dmesg [OPTIONS...]
Print or control log ring buffer
  -C, --clear            Clear the ring buffer
  -c, --read-clear       Clear the ring buffer contents after printing
  -f, --follow           Track new entries as are logged
  -n, --new              Show only new log messages
```
I actually needed this capability in the RPC for work on a `flux-shutdown(1)` command as discussed in #3895, where I thought it would be useful to have an option to show log messages as the instance is shutting down.  Touching that old code also resulted in some gratuitous cleanup and modernization, and ultimately I figured it was best as a standalone PR.

### DIFF
--- a/src/broker/log.c
+++ b/src/broker/log.c
@@ -51,7 +51,7 @@ typedef struct {
     zlist_t *buf;
     int ring_size;
     int seq;
-    zlist_t *followers;
+    struct flux_msglist *followers;
 } logbuf_t;
 
 struct logbuf_entry {
@@ -97,28 +97,6 @@ static void logbuf_trim (logbuf_t *logbuf, int size)
     }
 }
 
-static void flux_msg_decref_wrapper (void *data)
-{
-    const flux_msg_t *msg = data;
-    flux_msg_decref (msg);
-}
-
-
-static int logbuf_follow (logbuf_t *logbuf, const flux_msg_t *msg)
-{
-    if (zlist_append (logbuf->followers,
-                      (flux_msg_t *)flux_msg_incref (msg)) < 0) {
-        flux_msg_decref (msg);
-        errno = ENOMEM;
-        return -1;
-    }
-    zlist_freefn (logbuf->followers,
-                  (flux_msg_t *)msg,
-                  flux_msg_decref_wrapper,
-                  true);
-    return 0;
-}
-
 static int append_new_entry (logbuf_t *logbuf, const char *buf, int len)
 {
     struct logbuf_entry *e;
@@ -136,11 +114,11 @@ static int append_new_entry (logbuf_t *logbuf, const char *buf, int len)
             errno = ENOMEM;
             return -1;
         }
-        msg = zlist_first (logbuf->followers);
+        msg = flux_msglist_first (logbuf->followers);
         while (msg) {
             if (flux_respond (logbuf->h, msg, e->buf) < 0)
                 log_err ("error responding to log.dmesg request");
-            msg = zlist_next (logbuf->followers);
+            msg = flux_msglist_next (logbuf->followers);
         }
     }
     return 0;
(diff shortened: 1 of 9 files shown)


src/common/libflux/msg_handler.c
```diff
@@ -334,7 +334,7 @@ static void call_handler (flux_msg_handler_t *mh, const flux_msg_t *msg)
             && matchtag != FLUX_MATCHTAG_NONE) {
             const char *errmsg;
             if (mh->rolemask == 0 || mh->rolemask == FLUX_ROLE_OWNER)
-                errmsg = "Request requires owner credentals";
+                errmsg = "Request requires owner credentials";
```
grondo, line 337: Oh wow, good catch!

src/broker/log.c
```diff
@@ -244,54 +243,40 @@ static int logbuf_set_filename (logbuf_t *logbuf, const char *destination)
     return 0;
 }
 
+static const char *int_to_string (int n)
+{
+    static char s[32];
+    (void)snprintf (s, sizeof (s), "%d", n);
```
grondo, line 249: Someone (not me!) will probably want a comment that `%d` of an `int` can never require more than 32 bytes (actually I think it is only 12 or 13 (for 4 byte int) but it doesn't hurt to have more than double)

t/t0017-security.t
```diff
@@ -137,7 +137,7 @@ test_expect_success 'flux logger not allowed for non-owner' '
 
 test_expect_success 'flux dmesg not allowed for non-owner' '
 	! FLUX_HANDLE_ROLEMASK=0x2 flux dmesg 2>dmesg.err &&
-	grep -q "Operation not permitted" dmesg.err
+	grep -q "Request requires owner credentials" dmesg.err
```
grondo, line 140: Looks like this change snuck into the wrong commit?

grondo (approved):
LGTM! Just a couple minor comments inline.

t/t0017-security.t
garlick: No that one belongs in the "simplify code" commit because I started printing the error with `log_msg_exit ("...%s", future_strerror (f, errno))` instead of just `log_err_exit ("...")`. 

So the output switched from the canned errno string to the Flux human readable message.

Which is how I noticed it was misspelled :-)

garlick:
re-pushed with fix for typo

t/t0017-security.t
grondo: Ah! got it, sorry about that!

src/common/libflux/msg_handler.c
garlick: Thanks - I'll not take a victory lap since I'm pretty sure I introduced and proliferated the typo :-/

src/cmd/builtin/dmesg.c
```diff
@@ -26,122 +26,66 @@ static struct optparse_option dmesg_opts[] = {
...
+
+        if (!(f = flux_rpc_pack (h,
+                                 "log.dmesg",
+                                 FLUX_NODEID_ANY,
+                                 FLUX_RPC_STREAMING,
+                                 "{s:b}",
+                                 "follow", optparse_hasopt (p, "follow"))))
+            log_err_exit ("error sending log.dmesg request");
+        while (flux_rpc_get (f, &buf) == 0) {
+            dmesg_fprint (stdout, buf, strlen (buf));
```
chu11, line 76: AFAICT we only output to stdout and no other file streams are possible?  We could save ourselves the extra parameter to `dmesg_fprintf()` and remove the `FILE *f` parameter.  Thus we could just use `printf` in `dmesg_fprintf()` ... and I guess `dmesg_fprintf()` could be renamed into `dmesg_output()` or something.

garlick:
OK to set MWP?

grondo:
Fine with me!

PR #4238, opened 2022-03-25
job-ingest: add TOML config 

grondo:
This PR adds a TOML config for `job-ingest` module in order to make configuration of validator plugins in the system possible using config files.

`flux config reload` is supported for `ingest.validator.plugins` and `ingest.validator.args`, but it would have been a little too much development to allow disabling the validator on reload, so that is not supported now (and this can be done by reloading the job-ingest module anyway).

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -290,7 +290,8 @@ MAN5_FILES_PRIMARY = \
 	man5/flux-config-exec.5 \
 	man5/flux-config-resource.5 \
 	man5/flux-config-archive.5 \
-	man5/flux-config-job-manager.5
+	man5/flux-config-job-manager.5 \
+	man5/flux-config-ingest.5 
 
 
 MAN7_FILES = $(MAN7_FILES_PRIMARY)
--- a/doc/man5/flux-config-ingest.rst
+++ b/doc/man5/flux-config-ingest.rst
@@ -1,0 +1,69 @@
+=====================
+flux-config-ingest(5)
+=====================
+
+
+DESCRIPTION
+===========
+
+The Flux **job-ingest** service verifies and validates job requests
+before announcing new jobs to the **job-manager**. Configuration of the
+**job-ingest** module can be accomplished either via the module command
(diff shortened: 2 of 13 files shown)


src/modules/job-ingest/job-ingest.c
```diff
@@ -914,10 +914,12 @@ int job_ingest_ctx_init (struct job_ingest_ctx *ctx,
             return -1;
```
garlick: commit message:
s/woker_create/worker_create/

src/modules/job-ingest/validate.c
```diff
@@ -165,28 +160,37 @@ struct validate *validate_create (flux_t *h,
...
+        if (worker_set_cmdline (v->worker[i], argc, argv) < 0)
             goto error;
     }
-    free (argv);
-    free (argz);
-    return v;
+    rc = 0;
 error:
     free (argv);
     free (argz);
```
garlick, line 183: need to save/restore errno here (since errors from this function are reported with `flux_log_error()` which expects errno to be set)

doc/Makefile.am
```diff
@@ -290,7 +290,8 @@ MAN5_FILES_PRIMARY = \
 	man5/flux-config-exec.5 \
```
garlick, line 290: commit message:
s/support/supports/

doc/man5/flux-config-ingest.rst
```diff
@@ -0,0 +1,69 @@
...
+=====================
+
+
+DESCRIPTION
+===========
+
+The Flux **job-ingest** service verifies and validates job requests
+before announcing new jobs to the **job-manager**. Configuration of the
+**job-ingest** module can be accomplished either via the module command
+line or a ``ingest`` TOML table. See the KEYS section below for supported
```
garlick, line 12: s/a/an/

garlick (approved):
LGTM, with a couple minor suggestions

grondo:
Thanks! I've made the suggested changes, rebased, squashed, and pushed the result.

PR #4239, opened 2022-03-26
add 'require-instance' job validator plugin

grondo:
This PR adds a validator plugin which attempts to require that all submitted jobs will result in a new instance of Flux. The idea is that we may not want to allow regular parallel jobs (as submitted by `flux mini run` and `flux mini submit` for example) in a system instance, since this results in potential extra burden on the system instance TBON and KVS.

At first it was thought that admins could just drop a validator plugin into the correct path if this was required, but in the end it seemed like offering a native solution (which also serves as another validator plugin example) was probably wise.


garlick (approved):
LGTM!

grondo:
Thanks! I'll set MWP.

PR #4240, opened 2022-03-26
ci: run 32bit build under linux32 personality

### LINKED ISSUE #4226: ci: 32bit builds aren't detecting 32bit architecture in configure
At some point, our 32bit CI builds started detecting the arch as `x86_64`. The distro is still 32bit, i.e. all binaries and libraries are 32bit, but configure is reporting:

```
  checking build system type... x86_64-pc-linux-gnu
  checking host system type... x86_64-pc-linux-gnu
  checking target system type... x86_64-pc-linux-gnu
```

I _think_ docker used to run `linux/386` images under a 32bit personality using `setarch`, but that doesn't seem to be the case anymore.

Related issue: https://github.com/moby/moby/issues/42732

From that issue, it doesn't look like developers are keen to fix this problem (They claim `uname -m` isn't the correct way to detect image architecture, a possibly true statement from a docker-centric point of view, but completely useless since we can't change how autotools detects the arch afaik).

We'll probably have to somehow add  `setarch i686` or similar somewhere in the build process to restore previous functionality.

grondo:
Problem: At some point the docker 32bit builder (platform = linux/i386)
stopped setting the i386 personality(2), so uname -m reports x86_64
instead of i686, and thus configure detects an x86_64 target.

Check for a platform of linux/i386 in checks_run.sh and re-invoke the
script under `setarch i386` to force the "correct" uname.

Fixes #4226

### DIFF
--- a/src/test/checks_run.sh
+++ b/src/test/checks_run.sh
@@ -26,6 +26,17 @@
 #  CC, CXX, LDFLAGS, CFLAGS, etc.
 #
 
+
+#  Ensure uname -m reports 32bit architecture if platform was specified
+#   as 386:
+#
+case $PLATFORM in *386)
+  unset PLATFORM
+  echo "Rexecuting under linux32 personality"
+  exec setarch i386 $0 "$@"
+  ;;
+esac
+
 # if make is old, and scl is here, and devtoolset is available and not turned
 # on, re-exec ourself with it active to get a newer make
 if make --version | grep 'GNU Make 4' 2>&1 > /dev/null ; then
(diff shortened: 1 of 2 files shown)


grondo:
Hold off on this one. The changes didn't have the intended effect in CI so I have to do more investigation.

I've set WIP for now.

grondo:
Found a solution that seems to work more reliably in GH actions and pushed the result.

Note: The current approach won't work when using `docker-run-checks.sh -I`. In that case you'll have to manually use `setarch`.

garlick (approved):
LGTM.

grondo:
Thanks! I've set MWP.

PR #4241, opened 2022-03-28
broker: don't leave shutdown state prematurely

garlick:
Problem: broker immediately posts "children-none" on entry to STATE_SHUTDOWN, even if it has TBON children.

`state_machine->child_count` is always set to zero on entry to STATE_SHUTDOWN because it hasn't been updated since it was
initialized, when no peers were  online.  There is actually no reason to be saving this value - just call the `overlay_get_child_peer_count()` accessor when the count is needed instead of caching it.

Add a failing test to the broker state machine sharness test script.

This bug means that leaves-to-root rc3 execution has been running without synchronization.  The rc3's effectively all start at the same time across the instance, and upstream could complete before downstream, which violates a primary invariant of our distributed service design.  Most likely what has been saving us from data loss is the "cleanup" scripts which ensure all jobs cancel or complete before any rc3 scripts begin executing, and the fact that rank 0 has more to do than the other ranks so it takes longer.

This bug seems to have been introduced in release 0.26.0 (2021-04-22).

### DIFF
--- a/src/broker/state_machine.c
+++ b/src/broker/state_machine.c
@@ -62,8 +62,6 @@ struct state_machine {
     struct monitor monitor;
     struct quorum quorum;
 
-    int child_count;
-
     struct flux_msglist *wait_requests;
 };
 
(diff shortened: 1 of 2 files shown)


garlick:
Thanks!   Setting MWP.

Issue #4081, opened 2022-02-01
flux.service: use StateDirectory instead of manual creation and chown of /var/lib/flux


grondo:
The flux service unit file currently explicitly creates, chowns, and chmods the state directory on startup:

https://github.com/flux-framework/flux-core/blob/d80c652e576a734eb33b51b929bc44251b8bd59f/etc/flux.service.in#L32-L34

However, it appears that the [StateDirectory](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#RuntimeDirectory=)  setting in the unit file is meant for this purpose, so we may want to use that instead.

The one difference is that the value for the state directory in the current  unit file uses the expansion of `LOCALSTATEDIR` which is dependent on `--prefix`. There might be an argument that the system instance should ignore `LOCALSTATEDIR` and instead use the system localstatedir under `/var/lib`. I note in the systemd reference above that the `STATE_DIRECTORY` environment variable will be set when `StateDirectory` is utilized, so we could use the following changes to `flux.service.in`:

```diff
diff --git a/etc/flux.service.in b/etc/flux.service.in
index bf9a90cfa..4893bf6b8 100644
--- a/etc/flux.service.in
+++ b/etc/flux.service.in
@@ -16,7 +16,7 @@ ExecStart=/bin/bash -c '\
   -Slocal-uri=local://@X_RUNSTATEDIR@/flux/local \
   -Slog-stderr-level=6 \
   -Slog-stderr-mode=local \
-  -Scontent.backing-path=@X_LOCALSTATEDIR@/lib/flux/content.sqlite \
+  -Scontent.backing-path=${STATE_DIRECTORY}/content.sqlite \
   -Sbroker.rc2_none \
   -Sbroker.quorum=0 \
   -Sbroker.quorum-timeout=none \
@@ -28,10 +28,9 @@ User=flux
 Group=flux
 RuntimeDirectory=flux
 RuntimeDirectoryMode=0755
+StateDirecotry=flux
+StateDirectoryMode=0700
 PermissionsStartOnly=true
-ExecStartPre=-/bin/mkdir -p @X_LOCALSTATEDIR@/lib/flux
-ExecStartPre=/bin/chown flux:flux @X_LOCALSTATEDIR@/lib/flux
-ExecStartPre=/bin/chmod 0700 @X_LOCALSTATEDIR@/lib/flux
 ExecStartPre=/usr/bin/loginctl enable-linger flux
[...]

PR #4242, opened 2022-03-28
systemd: make directory init failure non-fatal

garlick:
Problem: the `content.sqlite` file is only created on rank 0 but if the systemd unit file ExecStartPre fails to set the owner and mode of `${localstatedir}/lib/flux` on any rank, then flux fails to start.

Example: `${localstatedir}/lib/flux` is a symbolic link pointing to `/scratch/flux` on all ranks, but `/scratch` is only mounted on rank 0.  Flux fails to start on the nonzero ranks.

Add a "-" in front of the commands so that the unit succeeds even if those commands fail.

### DIFF
--- a/etc/flux.service.in
+++ b/etc/flux.service.in
@@ -31,8 +31,8 @@ RuntimeDirectory=flux
 RuntimeDirectoryMode=0755
 PermissionsStartOnly=true
 ExecStartPre=-/bin/mkdir -p @X_LOCALSTATEDIR@/lib/flux
-ExecStartPre=/bin/chown flux:flux @X_LOCALSTATEDIR@/lib/flux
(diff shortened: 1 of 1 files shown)


grondo:
After working on this, any opinion on #4081? I can't remember any details, but we should close that issue with a description of why we can't use the systemd provided `StateDirectory` and `StateDirectoryMode` if they aren't a tenable solution for the flux service unit (e.g. because of the reason listed above: state directory may not exist on all nodes)

garlick:
Oh I had lost track of that issue!  I will have a look.

garlick:
Closing this as I think the solution in #4081 puts us on a better path than doing "whatever" in ExecStartPre as we do now.

Furthermore, upon consideration,  I think my whining about flux not starting due to dangling symlinks is a problem of my on creation and not something we have to work around.

PR #4243, opened 2022-03-28
rc3: ensure exit code reflects any errors

garlick:
Problem: failures in the rc3 script are not reflected in the instance exit code, so shutdown-related failures could be missed
in test.

The rc3 script previously ran with "#!/bin/bash -e" so any error caused the script to immediately exit, but the "-e" was dropped to address concerns that a premature exit due to a trivial error could prevent the KVS data from being dumped out to stable storage.

Add some logic so that the script runs to completion but any errors that occur are reflected in the script exit code.

### DIFF
--- a/etc/rc3
+++ b/etc/rc3
@@ -1,12 +1,13 @@
 #!/bin/bash
 
 RANK=$(flux getattr rank)
+exit_rc=0
 
 # Usage: modrm {all|<rank>} modname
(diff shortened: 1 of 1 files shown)


grondo (approved):
LGTM!

garlick:
Thanks - setting MWP.

PR #4245, opened 2022-03-29
support node exclusive allocations

### LINKED ISSUE #4228: set ntasks to nnodes if ntasks isn't given in flux mini command (minor issue)
I noticed when testing node exclusive scheduling that `flux mini run -N2 hostname` produces an error:
```
[day36@fluke108:~]$ flux mini run -N 2 hostname
flux-mini: ERROR: node count must not be greater than task count
[day36@fluke108:~]$
```
This also occurs without node exclusive scheduling, but it seems more likely for users to hit it with node exclusive scheduling on. It seems like it would be reasonable for the task count to be set to the node count if it's not specified by the user.

grondo:
This PR adds an `--exclusive` flag in the `flux-mini` set of commands, which only works when `-N, --nodes` is also specified, and sets the `exclusive` key to `true` on the node vertex in jobspec. Support for node exclusive requests is added to sched-simple (mainly for testing), and finally the number of tasks/slots in `flux mini` commands is defaulted to the number of nodes, and the `exclusive` flag enabled, when number of tasks is not otherwise specified on the command line (fixes #4228).

As noted above, the new `--exclusive` flag only works if `-N, --nodes` is also specified. I was having trouble deciding if Jobspec V1 could even support setting the exclusive flag on nodes without also specifying the number of required nodes, though I admittedly didn't spend too much time on it. The current approach addresses the main concern here, which was defaulting ntasks/slots to nnodes in a coherent manner, so that seemed good enough for now.

There is a difference in behavior between Fluxion and sched-simple when allocating nodes exclusively. Fluxion will emit R with only the minimum resources required by the jobspec on each node (but all resources on the node will show as allocated), whereas sched-simple currently emits all exclusive node resources in the emitted R. Thus, for sched-simple:

```console
$ flux mini alloc -N4 flux resource list
      free      4       16        0 fluke[7-10]
      down      0        0        0 
```

whereas with Fluxion:
```console
$ flux mini alloc -N4 flux resource list
      free      4        4        0 fluke[7-10]
      down      0        0        0 
```

I'm not sure, but this seems like something we'll want to fix. Though I may be missing some subtlety here, it seems like it would be surprising behavior if users ask for nodes exclusively but only get one core from each node in their child instance. @dongahn?

### DIFF
--- a/src/common/librlist/rlist.c
+++ b/src/common/librlist/rlist.c
@@ -2012,9 +2012,9 @@ rlist_alloc_constrained (struct rlist *rl,
     return result;
 }
 
-struct rlist *rlist_alloc_ex (struct rlist *rl,
-                              const struct rlist_alloc_info *ai,
-                              flux_error_t *errp)
+struct rlist *rlist_alloc (struct rlist *rl,
+                           const struct rlist_alloc_info *ai,
+                           flux_error_t *errp)
 {
     struct rlist *result = NULL;
 
@@ -2044,35 +2044,6 @@ struct rlist *rlist_alloc_ex (struct rlist *rl,
     }
 
     return result;
-}
-
-struct rlist *rlist_alloc (struct rlist *rl, const char *mode,
-                          int nnodes, int slots, int slotsz)
-{
-    struct rlist *result = NULL;
-    struct rlist_alloc_info ai = {
-        .nnodes = nnodes,
-        .nslots = slots,
-        .slot_size = slotsz,
-        .mode = mode
-    };
-
-    if (alloc_info_check (rl, &ai, NULL) < 0)
-        return NULL;
-
-    /*
-     *   Try allocation. If it fails with not enough resources (ENOSPC),
-     *    then try again on an empty copy of rlist to see the request could
-     *    *ever* be satisfied. Adjust errno to EOVERFLOW if not.
-     */
-    result = rlist_try_alloc (rl, &ai);
-    if (!result && (errno == ENOSPC)) {
-        if (rlist_alloc_feasible (rl, mode, nnodes, slots, slotsz))
-            errno = ENOSPC;
-        else
-            errno = EOVERFLOW;
-    }
-    return (result);
 }
 
 static int rlist_free_rnode (struct rlist *rl, struct rnode *n)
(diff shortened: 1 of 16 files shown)


grondo:
It occurs to me I should get @trws and @ryanday36's blessing on the end result here, i.e. that `flux mini <command> --nodes=X` now doesn't throw an error. Instead it acts the same as `flux mini <command> --nodes=X --nslots=X --exclusive`. I hope that is generally what we agreed to as the most reasonable solution in #4228. (I may have missed the mark though!)

trws:
Sounds right to me, good stuff!
________________________________

grondo:
Great! Thanks @ryanday36 and @trws!

src/common/librlist/rlist.c
```diff
@@ -1809,18 +1809,19 @@ static zlistx_t *rlist_get_nnodes (struct rlist *rl, int nnodes)
  *  Works by getting the first N least utilized nodes and spreading
  *  the nslots evenly across the result.
  */
-static struct rlist *rlist_alloc_nnodes (struct rlist *rl, int nnodes,
-                                         int cores_per_slot, int slots)
+static struct rlist *rlist_alloc_nnodes (struct rlist *rl,
+                                         struct rlist_alloc_info *ai)
```
garlick, line 1815: Nice cleanup!

Not sure if this is feasible, but in functions where the `struct rlist_alloc_info` parameter is not modified, can it be declared `const`?  Parameters that were obviously "in" before could now be sneakily "in/out", so the extra declaration could clarify that.

garlick (approved):
I just had one comment, not a show stopper either.  Great work here!

I did give this a manual poke and LGTM.

Truth be known, I seem to type `-N` without `-n` a lot, despite knowing better.  It's hard to unlearn that muscle memory.  So at minimum this change is going reduce reduce my annoyed-with-self time.

src/common/librlist/rlist.c
grondo: Yeah, that sounds like a good suggestion!

grondo: Ok, I've made that change and force-pushed the result. Thanks!

garlick:
MWP?

grondo:
Yes, setting MWP now.

PR #4246, opened 2022-03-29
broker: prevent systemd restart if rc1 fails

garlick:
Problem:  if rc1 fails, for example, due to a misconfiguration, systemd will restart the broker over and over, filling up logs with useless noise until someone notices and fixes the problem.

If rc1 fails, most likely it is going to fail again on restart.   It seems more appropriate to not restart in this case.

This PR configures a special "don't restart" exit code using  `RestartPreventExitStatus`, and then uses it in place of the rc1 script exit code when rc1 fails, but only when run from systemd.

This was peeled off of work on #3895 - it seemed like it should stand alone for the release notes.

### DIFF
--- a/doc/man7/flux-broker-attributes.rst
+++ b/doc/man7/flux-broker-attributes.rst
@@ -100,6 +100,10 @@ broker.rc1_path [Updates: C]
 broker.rc3_path [Updates: C]
    The path to the broker's rc3 script.  Default: ``${prefix}/etc/flux/rc1``.
 
+broker.exit-restart [Updates: C, R]
+   A numeric exit code that the broker uses to indicate that it should not be
+   restarted.  This is set by the systemd unit file.  Default: unset.
+
 broker.starttime
    Timestamp of broker startup from :man3:`flux_reactor_now`.
 
--- a/etc/flux.service.in
+++ b/etc/flux.service.in
@@ -20,11 +20,13 @@ ExecStart=/bin/bash -c '\
   -Sbroker.rc2_none \
   -Sbroker.quorum=0 \
   -Sbroker.quorum-timeout=none \
+  -Sbroker.exit-norestart=42 \
 '
 SyslogIdentifier=flux
 ExecReload=@X_BINDIR@/flux config reload
 Restart=always
 RestartSec=5s
+RestartPreventExitStatus=42
 User=flux
 Group=flux
 RuntimeDirectory=flux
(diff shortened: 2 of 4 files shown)


garlick:
Got a failure on f33 that I haven't seen before:
```
  expecting success: 
  	cat <<-EOF >limit.sh &&
  	#!/bin/sh -e
  	round() { printf "%.0f" \$1; }
  
  	expiration=\$(flux kvs get resource.R | jq .execution.expiration)
  	echo "expiration is \$expiration"
  
  	id1=\$(flux mini submit --wait-event=start sleep 300)
  	flux jobs -no "{id.f58} {expiration}"
  	test "\$(flux jobs -no {expiration} \$id1)" = "\$expiration"
  	flux job cancelall -f
  
  	id2=\$(flux mini submit --wait-event=start -t 1m sleep 300)
  	flux jobs -no "{id.f58} {expiration}"
  	exp2=\$(flux jobs -no {expiration} \$id2)
  	test \$(round \${exp2}) -lt \$(round \${expiration})
          flux job cancelall -f
  	EOF
  	chmod +x limit.sh &&
  	flux mini run --time-limit=10m flux start ./limit.sh
  
  expiration is 1648588064.5174201
  fMgNyyy 1648588064.51742
  2022-03-29T20:57:46.405543Z broker.err[0]: rc2.0: /bin/bash -c ./limit.sh Exited (rc=1) 0.7s
  2022-03-29T20:57:46.499055Z broker.err[0]: cleanup.1: flux-job: Canceled 1 jobs (0 errors)
  not ok 2 - job timelimits are propagated
```

grondo (approved):
LGTM! Should `broker.exit-norestart` get an entry in `flux-broker-attributes(7)`?

Also, very minor suggestion. It might help to split the commit that adds and uses `broker.exit-norestart` into one that adds `norestart_configure()` (support a special exit code for the broker to. communicate that it should not auto-restart), and one that sets the exit code to `broker.exit-norestart` on rc1 failure (prevent restart on rc1 failure). Don't do that unless you really want to, things are fine as they are.

grondo:
> Got a failure on f33 that I haven't seen before:

Hm, looks like Python decided to round the value for expiration returned from `flux jobs` vs that returned by `jq`. Probably have to adjust the test to more consistently format those floating point values..

garlick:
Thanks for the review @grondo - I made all the suggested changes.

grondo:
Looks good! I already approved so feel free to set MWP.

PR #4248, opened 2022-03-30
broker: add statedir attribute, drop content.backing-path

garlick:
This adds a `statedir` broker attribute that is set by the systemd unit file to the path of a directory that is preserved across broker restarts.  The  `content.backing-path` broker attribute is no more.  Its users now place files in `statedir`, if set, with fallback to `rundir`.

Instructions can be made more clear:  set `statedir` if you want flux to be restartable, and leave the details to us.  If you don't set it, any files go to `rundir`, which is always cleaned up.

Also TODO if this is merged:
* Update [this faq entry](https://flux-framework.readthedocs.io/en/latest/faqs.html#memory-exhaustion-on-a-node-when-running-large-ensembles-with-flux) about redirecting sqlite to another file system
* Follow up with a PR to `job-archive` to use `statedir` for its database location by default, and to define another way of activating it, since right now configuration of the dbpath is what causes it to run.


src/broker/broker.c
```diff
@@ -735,6 +738,11 @@ static int checkdir (const char *name, const char *path)
         log_err ("cannot stat %s %s", name, path);
         return -1;
     }
+    if (sb.st_uid != getuid ()) {
```
grondo, line 741: Should `sb.st_uid == 0` also be allowed? (Actually, I don't know the reason for the directory ownership requirement, instead of just write access. Is it for security purposes (to ensure an existing content store could not have been modified by another user), or just to make it simpler to check for write access?)

If this requirement stays, then we might want to update the documentation of `statedir` to include something like: "If set, `statedir` must exist and be owned by the instance owner"

grondo (approved):
This looks good to me. I question the requirement that `statedir` be strictly owned by the instance owner below, but this could easily be addressed later if it even becomes an issue, so that comment is more of a question. (It might be nice to update the doc though)

src/broker/broker.c
garlick: My thought was that the tests for u=rwx does not have the desired effect if "u" is not the instance owner.  So we could just make the test more sophisticated, looking for write access to the directory by `getuid()`?   That seemed a bit involved without a use case in mind.  Does that make sense?

grondo: > My thought was that the tests for u=rwx does not have the desired effect if "u" is not the instance owner.

Yes, that part was clear. My thought is that it wasn't clear you can't specify a `statedir` to which you have write access but are not the owner. A scenario that is unlikely but was the first thing I tried (`statedir=/tmp`). It also might be the case that sysadmins might prefer a directory ownership of `root` for a statedir location for the system instance.

However, I agree there probably isn't a real use case at this time, and if we need to fix it the code already exists in `flux-security` to check a directory for secure permissions exhaustively, so we can just bring that code in when necessary.

My main suggestion was just to add documentation of the requirement at this point.

garlick: Got it.  OK, I'll make the documentation change.

garlick:
Added the suggested documentation change:
```
      statedir [Updates: C]
              A  directory in which persistent state such as the content back‐
              ing store data is stored  by  the  Flux  broker,  to  facilitate
              restarts.   If  unset,  this  data  goes  to  rundir where it is
              cleaned up on instance shutdown.  If set,  this  directory  must
              exist and be owned by the instance owner.  Default: unset.
```
Also some ci tests were failing due to a typo I introduced when fixing a && chain.  Fixed.  Sorry for the very public whack-a-mole!

src/modules/content-sqlite/content-sqlite.c
```diff
@@ -657,7 +647,7 @@ int mod_main (flux_t *h, int argc, char **argv)
         flux_log_error (h, "content_sqlite_create failed");
         return -1;
     }
-    if (content_sqlite_opendb(ctx) < 0)
+    if (content_sqlite_opendb (ctx) < 0)
         goto done;
     if (content_register_backing_store (h, "content-sqlite") < 0)
         goto done;
```
chu11, line 653: not a big deal, but dunno if these whitespace / code style changes were supposed to go into this commit or not.

doc/man7/flux-broker-attributes.rst
```diff
@@ -33,15 +33,23 @@ version
...
+   set on the command line, beware exceeding the UNIX domain socket path limit
+   described in :linux:man7:`unix`, as low as 92 bytes on some systems.  To
+   support the :man1:`flux-start` ``--test-size`` option where multiple brokers
+   share a ``rundir``, if ``rundir`` is set to a pre-existing directory, the
+   directory is not removed by the broker on exit.  In most cases this
+   attribute should not be set by users.
+
+statedir [Updates: C]
+   A directory in which persistent state such as the content backing store data
+   is stored by the Flux broker, to facilitate restarts.  If unset, this data
```
chu11, line 49: first sentence seems a little run on.  Perhaps

A directory in which persistent state is stored by the Flux broker.  For example, content backing store data will be stored here to facilitate restarts.

garlick:
Pushed those changes.   I'm going to go ahead and set MWP.  Hope that's OK.

PR #4249, opened 2022-03-30
flux-resource: support properties when listing resources

grondo:
It occurred to me that while we added support for execution target properties in #4236, there was no way to actually query what properties were available from the current instance.

This PR adds some missing calls to the librlist API to make this possible, adds similar methods to the Python `ResourceSet` class, and and finally makes a `{properties}` field available from `flux resource list`. This field is now displayed by default if any properties exist in the current instance:

```console
$ flux resource list
      free                 1        4        0 asp
      free xx              1        4        0 asp
      free foo,xx          1        3        0 asp
      free foo             1        4        0 asp
 allocated foo,xx          1        1        0 asp
      down                 0        0        0 
```

The list of all properties available can easily be reported with `flux resource list -s all -no {properties}`:
```console
$ flux resource list -s all -no {properties}
foo,xx
```

This required a small rework of how `flux-resource list` collects similar lines together, which is explained in the commit message.

### DIFF
--- a/src/bindings/python/flux/resource/ResourceSet.py
+++ b/src/bindings/python/flux/resource/ResourceSet.py
@@ -107,7 +107,9 @@ class ResourceSet:
 
     def copy(self):
         """Return a copy of a ResourceSet"""
-        return ResourceSet(self.impl.copy())
+        rset = ResourceSet(self.impl.copy())
+        rset.state = self.state
+        return rset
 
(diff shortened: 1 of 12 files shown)


garlick (approved):
This LGTM.

grondo:
Thanks! setting MWP.

PR #4250, opened 2022-03-31
add flux-shutdown command

garlick:
This adds `flux-shutdown(1)` and a mechanism for synchronizing shutdown activities so the command can monitor logs on the way down, and not exit until after `rc3` completes.
```
$ flux shutdown -h
Usage: flux shutdown [OPTIONS] [TARGET]
Shut down the Flux instance
      --background       Exit the command immediately after initiating shutdown
      --quiet            Show only log messages <= LOG_WARNING level
  -v, --verbose=[LEVEL]  Increase log verbosity: 0=show log messages <= LOG_INFO
                         level (default), 1=show all log messages
```
Example, shutting down my system instance:
```
$ sudo flux shutdown
broker.info[0]: cleanup.0: /bin/bash -c flux queue stop --quiet Exited (rc=0) 0.0s
broker.info[0]: cleanup.1: /bin/bash -c flux job cancelall --user=all --quiet -f --states RUN Exited (rc=0) 0.0s
broker.info[0]: cleanup.2: /bin/bash -c flux queue idle --quiet Exited (rc=0) 0.0s
broker.info[0]: cleanup-success: cleanup->shutdown 0.18846s
broker.info[0]: quorum-full: ignored in shutdown
sched-simple.info[0]: resource update: {"down":"2"}
broker.info[0]: quorum-full: ignored in shutdown
sched-simple.info[0]: resource update: {"down":"5"}
broker.info[0]: quorum-full: ignored in shutdown
sched-simple.info[0]: resource update: {"down":"3"}
broker.info[0]: children-complete: shutdown->finalize 0.454286s
broker.info[0]: quorum-full: ignored in finalize
sched-simple.info[0]: resource update: {"down":"7"}
broker.info[0]: quorum-full: ignored in finalize
broker.info[0]: rc3.0: /bin/bash -c /usr/local/etc/flux/rc3 Exited (rc=0) 1.0s
broker.info[0]: rc3-success: finalize->goodbye 0.995555s
$
```
Yeah, some of those logs could use some fine tuning, but that's a matter for another PR.

I envisioned adding other options to the shutdown command such as:
- scheduling a shutdown time in the future
- disable the queue N minutes before shutdown
- kill all running jobs and wait for them to become inactive (maybe with some control over the exception message)
- trigger a `flux-dump` for later restart
- trigger a dump/restore on the `content.sqlite` to remove unreferenced blobrefs

There may be other shutdown-time options that would be useful to have.  But for now, just the bare bones.

This is WIP pending the addition of tests, but I wanted to get any feedback on the overall design.


garlick:
Re-pushed with some tests, and some fixes that came out of writing the tests.

garlick:
Repushed with test race fix.

src/common/libflux/message.c
```diff
@@ -1446,16 +1446,14 @@ int flux_match_asprintf (struct flux_match *m, const char *fmt, ...)
```
chu11, line 1446: commit message typo "Modfiy" -> "Modify"

garlick: Pushed that fix.

grondo:
Cool! Tested this out on fluke and it works to remotely shut down a job. Nice!

```console
$ src/cmd/flux shutdown ƒSk7twdWVfm
broker.err[0]: rc2.0: /bin/bash Exited (rc=130) 36.7s
broker.info[0]: rc2-fail: run->cleanup 36.6892s
broker.info[0]: cleanup.0: /bin/bash -c flux queue stop --quiet Exited (rc=0) 0.1s
broker.info[0]: cleanup.1: /bin/bash -c flux job cancelall --user=all --quiet -f --states RUN Exited (rc=0) 0.1s
broker.info[0]: cleanup.2: /bin/bash -c flux queue idle --quiet Exited (rc=0) 0.1s
broker.info[0]: cleanup-success: cleanup->shutdown 0.230415s
broker.info[0]: children-complete: shutdown->finalize 0.491639s
sched-simple.info[0]: resource update: {"down":"1"}
broker.info[0]: rc3.0: /bin/bash -c /g/g0/grondo/git/flux-core.git/etc/rc3 Exited (rc=0) 1.0s
broker.info[0]: rc3-success: finalize->goodbye 1.03809s
```

However, when I do try to "accidentally" run `flux shutdown` within an instance, it just hangs for me. Not sure if this is reproducible for someone else, perhaps I've done something wrong:

```
grondo@fluke108:~/git/flux-core.git$ flux mini run -o pty.interactive -N2 -n2 src/cmd/flux start
ƒ(s=2,d=1,builddir) grondo@fluke7:~/git/flux-core.git$ FLUX_HANDLE_TRACE=1 flux shutdown -vvv
> 0.00000
> shutdown.start
> flags=topic,payload,route,streaming userid=unknown rolemask=none nodeid=any matchtag=1
> {"loglevel":7}

```

garlick:
Uh oh, let me see if I can reproduce that.  I do have a test for it but the test is rather simple minded.

garlick:
Just pushed a fix for that hang.  The problem is that we are using SIGTERM generally to abort rc scripts spawned by the broker.  Most of them are shell scripts, but the shell (at least `bash`) ignores SIGTERM when a subprocess is running.  In this case the shutdown command requests to abort rc2, but the initial program shell ignores the signal because sleep is running.

SIGHUP has the same default action (terminate process) as SIGTERM, but the shell distributes SIGHUP to any children, so it seems like a much better choice.  The fix is to switch over to that signal.  Possibly in the future we may find that in some extreme cases this signal needs to be followed with a SIGKILL after a timeout, but for now I think this is fine.

grondo (approved):
This looks good to me! I think the manpage needs to be updated to indicate the switch to `SIGHUP`, but I may have just missed that change.

I tested some corner cases, but everything seems to work as advertised.

The only thing I wonder, and I wouldn't argue that it needs to be added now, is if we want a `TIME` argument like `shutdown(8)`, with a default of waiting at least a few seconds and an option to cancel the shutdown with a Ctrl-C?

Honestly, we can probably wait until someone accidentally says "What's this command do?" and types `flux shutdown` as root, but I just wanted to bring it up (so I can later declare I once had forethought)

garlick:
OK, I pushed with a fix for the man page.

Yes, I agree we may want some options like shutdown(8) in the future.  Just wanted to get the basics in first.

I'll go ahead and set MWP.

PR #4252, opened 2022-03-31
libsdprocess: minor fixups

chu11:
Pulling out some minor libsdprocess fixups from #4070 and a forthcoming branch about systemd properties.

### DIFF
--- a/src/common/libsdprocess/sdprocess.c
+++ b/src/common/libsdprocess/sdprocess.c
@@ -1013,16 +1013,15 @@ static int setup_state_watcher (sdprocess_t *sdp, flux_reactor_t *reactor)
 
     /* Setup callback when `sd_bus_process ()` is called on properties
      * changed */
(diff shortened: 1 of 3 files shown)


chu11:
re-pushed, fixing up a LGTM issue

grondo (approved):
This looks fine to me. Only suggestion would be to add a  comment about why the `sd_bus_match_signal_async()` function is not safe to use (so we don't forget), but I consider that as optional, so approving.

chu11:
> This looks fine to me. Only suggestion would be to add a comment about why the sd_bus_match_signal_async() function is not safe to use (so we don't forget), but I consider that as optional, so approving.

Thanks!  I'll leave out a comment.  It ends up that most of `libsdprocess` is synchronous, the lingering async was more of an accidental addition (assumed cut & paste).

grondo:
Ok, that makes sense!

Issue #2032, opened 2019-02-20
usability: check for tty in flux-start in "interactive shell" mode


grondo:
We've seen several early users try
```
$ srun flux start
```

and report that it is "hanging" when really the lack of pty is causing no prompt to be displayed.

At least in the case of `flux-start`'s "default" behavior when no initial program is specified, we should check for pty and error out if none is found (unless there's a case I'm not thinking of where this is desired behavior)

dongahn:
Related, I've seen cases where users just followed our `README.md` on an mxterm window

```
srun --pty --mpi=none -N64 src/cmd/flux start
```

and happened to launch too many brokers per node.  Mxterm sets num-tasks which affect the number of processes you want to launch with the above command. 



We should be better off changing this to?
```
srun --pty --mpi=none -N64 -n64 src/cmd/flux start
```

grondo:
Yes, good idea!

PR #4253, opened 2022-03-31
broker: emit error when running interactive shell without tty

grondo:
While going through some old issues I found one that seemed easy and valuable to fix

 * #2032

This PR has the broker check for a tty on stdin when invoking an interactive shell. If there is no tty the broker exits with an error:

```console
$ flux mini run flux start                                                                              
lt-flux-broker: stdin is not a tty - can't run interactive shell               
```

This prevents wasting time when you think an interactive job is taking a long time to start, but really you've forgotten `-o pty`/`-o pty.interactive` or the `--pty` option with `srun`.

There was a little test fallout from this one, and most of my time was actually spent trying to fix Python encoding errors in the `runpty` test script (as well as working around timing issues getting EOF to the shell in the one case that is required for testing)

### DIFF
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -661,6 +661,11 @@ static void set_proctitle (uint32_t rank)
 static int create_runat_rc2 (struct runat *r, const char *argz, size_t argz_len)
 {
     if (argz == NULL) { // run interactive shell
+        /*  Check if stdin is a tty and error out if not to avoid
+         *   confusing users with what appears to be a hang.
+         */
+        if (!isatty (STDIN_FILENO))
+            log_msg_exit ("stdin is not a tty - can't run interactive shell");
         if (runat_push_shell (r, "rc2") < 0)
             return -1;
     }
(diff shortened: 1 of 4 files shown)


garlick (approved):
Nice!  Sorry the tests were a nuisance.

grondo:
Thanks, and no problem!

vsoch:
Can I run it without interactive? E.g., I'm trying to run flux in a container base on github actions and I'm seeing this error - I can't figure out a way to disable or, or to run flux (and test with snakemake) with needing to start flux.

grondo:
@vsoch: Yes, as discussed offline, this check and error for an interactive tty only occurs when there are no arguments provided to `flux start` or `flux broker` and thus an interactive shell is run. This can be a problem if a user doesn't allocate a tty to the instance, then the command appears to hang, when really the shell is just not displaying a prompt.

You can run non-interactively by providing any argument to `flux start` or `flux broker`, even `flux start bash` if you really want to start an interactive shell without a tty.

vsoch:
For future us, the workflow that worked wound up being:

```yaml
      - name: Start Flux and Test Workflow
        run: |
          su fluxuser
          cd examples/flux
          which snakemake
          flux start snakemake --flux --jobs=1
```

Issue #3202, opened 2020-09-09
system instance: broker exposes exec service on rank 0


garlick:
Problem: there is no way to prevent root or flux from running arbitrary commands (as flux) on the node running the rank 0 broker.  If that happens to be a cluster management node, some sites have a policy of restricting access to limit the effect of a compromised compute or login node.
```
$ sudo flux exec -r 0 id
uid=500(flux) gid=500(flux) groups=500(flux)
```
(where invoking user is root or flux).

There should probably be some way to prevent this by configuration.

PR #4258, opened 2022-04-01
broker: reject remote exec requests on rank 0

grondo:
This PR addresses #3202. A simple approach is taken for now: rather than offering configuration, remote subprocess exec requests to rank 0 are always rejected.

It is unlikely that use of `flux exec` or `flux_rexec()` to rank 0 from another broker is a current use case, so this approach was considered the most safe for now.

### DIFF
--- a/src/broker/overlay.c
+++ b/src/broker/overlay.c
@@ -31,6 +31,7 @@
 #include "src/common/libutil/errno_safe.h"
 #include "src/common/libutil/monotime.h"
 #include "src/common/librouter/rpc_track.h"
+#include "src/common/libccan/ccan/ptrint/ptrint.h"
 
 #include "overlay.h"
 #include "attr.h"
@@ -424,6 +425,14 @@ static struct child *child_lookup (struct overlay *ov, const char *id)
             return child;
     }
     return NULL;
+}
+
+bool overlay_msg_is_local (const flux_msg_t *msg)
+{
+    /*  Return true only if msg is non-NULL and the message does not
+     *  have the overlay supplied "remote" tag.
+     */
+    return (msg && flux_msg_aux_get (msg, "overlay::remote") == NULL);
 }
 
 /* Given a rank, find a (direct) child peer.
@@ -856,6 +865,13 @@ static void child_cb (flux_reactor_t *r, flux_watcher_t *w,
 
     if (!(msg = zmqutil_msg_recv (ov->bind_zsock)))
         return;
+    /* Flag this message as remotely received. This allows efficient
+     * operation of the overlay_msg_is_local() function.
+     */
+    if (flux_msg_aux_set (msg, "overlay::remote", int2ptr (1), NULL) < 0) {
+        logdrop (ov, OVERLAY_DOWNSTREAM, msg, "failed to tag msg as remote");
+        goto done;
+    }
     if (flux_msg_get_type (msg, &type) < 0
         || !(uuid = flux_msg_route_last (msg))) {
         logdrop (ov, OVERLAY_DOWNSTREAM, msg, "malformed message");
@@ -950,6 +966,13 @@ static void parent_cb (flux_reactor_t *r, flux_watcher_t *w,
 
     if (!(msg = zmqutil_msg_recv (ov->parent.zsock)))
         return;
+    /* Flag this message as remotely received. This allows efficient
+     * operation of the overlay_msg_is_local() function.
+     */
+    if (flux_msg_aux_set (msg, "overlay::remote", int2ptr (1), NULL) < 0) {
+        logdrop (ov, OVERLAY_UPSTREAM, msg, "failed to tag msg as remote");
+        goto done;
+    }
     if (flux_msg_get_type (msg, &type) < 0) {
         logdrop (ov, OVERLAY_UPSTREAM, msg, "malformed message");
         goto done;
--- a/src/common/libsubprocess/subprocess.h
+++ b/src/common/libsubprocess/subprocess.h
@@ -124,6 +124,19 @@ flux_subprocess_server_t *flux_subprocess_server_start (flux_t *h,
                                                         const char *local_uri,
                                                         uint32_t rank);
 
+
+typedef int (*flux_subprocess_server_auth_f) (const flux_msg_t *msg,
+                                              void *arg);
+
+/*   Register an authorization function to the subprocess server
+ *
+ *   The registered function should return 0 to allow the request to
+ *    proceed, and -1 with errno set to deny the request.
+ */
+void flux_subprocess_server_set_auth_cb (flux_subprocess_server_t *s,
+                                         flux_subprocess_server_auth_f fn,
+                                         void *arg);
+
 /*  Stop a subprocess server / cleanup flux_subprocess_server_t.  Will
  *  send a SIGKILL to all remaining subprocesses.
  */
--- a/src/broker/exec.c
+++ b/src/broker/exec.c
@@ -23,6 +23,7 @@
 
 #include "attr.h"
 #include "exec.h"
+#include "overlay.h"
 
 #define EXEC_TERMINATE_TIMEOUT 5.0
 
@@ -75,6 +76,15 @@ int exec_terminate_subprocesses_by_uuid (flux_t *h, const char *id)
     return 0;
 }
 
+static int reject_nonlocal (const flux_msg_t *msg, void *arg)
+{
+    if (!overlay_msg_is_local (msg)) {
+        errno = EPERM;
+        return -1;
+    }
+    return 0;
+}
+
 int exec_initialize (flux_t *h, uint32_t rank, attr_t *attrs)
 {
     flux_subprocess_server_t *s = NULL;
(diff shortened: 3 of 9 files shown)


src/broker/overlay.c
```diff
@@ -426,6 +426,29 @@ static struct child *child_lookup (struct overlay *ov, const char *id)
...
+        errno = EINVAL;
+        return false;
+    }
+
+    /*  If the last pushed route is a peer or parent, then this message
+     *   was definitely not local.
+     */
+    if (strcmp (route, ov->parent.uuid) == 0
+        || child_lookup_online (ov, route)
+        || child_lookup (ov, route))
```
garlick, line 447: `child_lookup_online()` could be dropped if `child_lookup()` is called.

garlick (approved):
This looks right.

src/broker/overlay.c
grondo: I was thinking that would be faster most of the time, but that was probably dumb. Sorry.

garlick: Oh sorry, I didn't think of that.  Maybe just add a comment to that effect?

grondo: Ok, I added that comment.

It also occurred to me another way to do this might be to flag messages that are received from the overlay (e.g. `flux_msg_aux_set (msg, "overlay::is_remote", int2ptr (1), NULL)`. This would avoid a guaranteed loop through all children on every local message, though I'm not sure that's a real issue in this limited case.

grondo: Since the implementation is simple: here's what would be the diff between what we have here and using an aux "flag" on messages. I'm not sure, though, if this is premature optimization, or if it is even a fully sufficient solution. What do you think? (dumb tests pass anyway)

```diff
diff --git a/src/broker/overlay.c b/src/broker/overlay.c
index e5d093f70..99f67fc8e 100644
--- a/src/broker/overlay.c
+++ b/src/broker/overlay.c
@@ -31,6 +31,7 @@
 #include "src/common/libutil/errno_safe.h"
 #include "src/common/libutil/monotime.h"
 #include "src/common/librouter/rpc_track.h"
+#include "src/common/libccan/ccan/ptrint/ptrint.h"
 
 #include "overlay.h"
 #include "attr.h"
@@ -438,15 +439,7 @@ bool overlay_request_is_local (struct overlay *ov, const flux_msg_t *msg)
         errno = EINVAL;
         return false;
     }
-
-    /*  If the last pushed route is a peer or parent, then this message
-     *   was definitely not local. Note: we call child_lookup_online()
-     *   before child_lookup() becuase the former is potentially much
-     *   faster.
-     */
-    if (strcmp (route, ov->parent.uuid) == 0
-        || child_lookup_online (ov, route)
-        || child_lookup (ov, route))
+    if (flux_msg_aux_get (msg, "overlay::remote"))
         return false;
     return true;
 }
@@ -881,6 +874,8 @@ static void child_cb (flux_reactor_t *r, flux_watcher_t *w,
 
     if (!(msg = zmqutil_msg_recv (ov->bind_zsock)))
         return;
+    if (flux_msg_aux_set (msg, "overlay::remote", int2ptr (1), NULL) < 0)
+        goto done;
     if (flux_msg_get_type (msg, &type) < 0
         || !(uuid = flux_msg_route_last (msg))) {
         logdrop (ov, OVERLAY_DOWNSTREAM, msg, "malformed message");
@@ -975,6 +970,8 @@ static void parent_cb (flux_reactor_t *r, flux_watcher_t *w,
 
     if (!(msg = zmqutil_msg_recv (ov->parent.zsock)))
         return;
+    if (flux_msg_aux_set (msg, "overlay::remote", int2ptr (1), NULL) < 0)
+        goto done;
     if (flux_msg_get_type (msg, &type) < 0) {
         logdrop (ov, OVERLAY_UPSTREAM, msg, "malformed message");
         goto done;
```

garlick: That seems very close to me!

One problem here though is that the aux container isn't part of the serialized message, so it gets cleared when the message is forwarded (for example to a broker module).   It seems like the right thing to do here would be to define a new message flag in RFC 3 and set it in the flags field of the message "proto frame'?

grondo: My thought was that currently `overlay_request_is_local()` is only exported to the broker code specifically, so perhaps this limited support is ok for now (and perhaps better than checking the parent and child uuids directly each time?)

When we want to export this functionality to modules or via libflux, then I think the idea of a new flag is great. Does that make sense?

garlick: Ah yes that does make sense.  OK, this looks good then!

I would suggest adding comments where the message is annotated stating that it's for use by `overlay_request_is_local()`.  Or maybe `overlay_message_is_remote()` would be more appropriate now.

grondo: Good suggestions!

grondo: Ok, I made this change and forced a push.

I kept the `is_local()` api for this function. This is because then the function can "fail safe" by returning false if `msg == NULL`.

Also, the function can work on any message now I assumed, not just requests, so it is renamed `overlay_msg_is_local()`. Let me know what you think.

src/broker/overlay.c
```diff
@@ -950,6 +965,11 @@ static void parent_cb (flux_reactor_t *r, flux_watcher_t *w,
 
     if (!(msg = zmqutil_msg_recv (ov->parent.zsock)))
         return;
+    /* Flag this message as remotely received. This allows efficient
+     * operation of the overlay_msg_is_remote() function.
+     */
+    if (flux_msg_aux_set (msg, "overlay::remote", int2ptr (1), NULL) < 0)
+        goto done;
```
garlick, line 972: Call `logdrop()` on error?

grondo:
Had to restart the coverage builder since it hung during the coverage setup job.

src/broker/overlay.c
```diff
@@ -426,6 +427,15 @@ static struct child *child_lookup (struct overlay *ov, const char *id)
     return NULL;
 }
 
+bool overlay_msg_is_local (const flux_msg_t *msg)
+{
+    if (!msg) {
+        errno = EINVAL;
+        return false;
+    }
```
garlick, line 435: This function could just return the result of `flux_msg_aux_get()` since that function also returns NULL if msg is NULL.  Also, the explicit setting of errno won't be helpful since the caller won't know whether errno is valid on false return.

src/broker/overlay.c
```diff
@@ -856,6 +866,11 @@ static void child_cb (flux_reactor_t *r, flux_watcher_t *w,
 
     if (!(msg = zmqutil_msg_recv (ov->bind_zsock)))
         return;
+    /* Flag this message as remotely received. This allows efficient
+     * operation of the overlay_msg_is_remote() function.
+     */
+    if (flux_msg_aux_set (msg, "overlay::remote", int2ptr (1), NULL) < 0)
+        goto done;
```
garlick, line 873: Call `logdrop()` on error?

src/common/libsubprocess/subprocess.h
```diff
@@ -124,6 +124,16 @@ flux_subprocess_server_t *flux_subprocess_server_start (flux_t *h,
                                                         const char *local_uri,
                                                         uint32_t rank);
 
+
+typedef int (*flux_subprocess_server_auth_f) (const flux_msg_t *msg,
+                                              void *arg);
```
garlick, line 129: Add comment stating that callback should return 0 to allow, -1 to deny?

src/broker/broker.c
```diff
@@ -432,7 +432,7 @@ int main (int argc, char *argv[])
         log_err ("heaptrace_initialize");
         goto cleanup;
     }
-    if (exec_initialize (ctx.h, ctx.rank, ctx.attrs) < 0) {
+    if (exec_initialize (ctx.h, ctx.overlay, ctx.attrs) < 0) {
```
garlick, line 435: Shouldn't be necessary to change `exec_initialize()` prototype, since authorization doesn't need `struct overlay`?

garlick (changes requested):
Found a few small things to whine about.

src/broker/overlay.c
grondo: Yeah, the only reason that is there is to satisfy a unit test. You can test if a function set `errno` by zeroing `errno` first (e.g. as in `strtol(3)`, but I agree no caller would actually do this so will remove.

src/broker/overlay.c
grondo: Oops, sorry, didn't look carefully enough to see that `logdrop()` is required/encouraged here.

src/broker/broker.c
grondo: Thanks, good catch. That was left over from the previous attempt. I should have just started over :facepalm:

src/broker/overlay.c
grondo: > This function could just return the result of flux_msg_aux_get() since that function also returns NULL if msg is NULL. 

Ah, but I just noticed that we want to return `false` for `msg == NULL` not `true`, so the `if (!msg)` will have to stay, or am I missing something?

grondo:
I've pushed a bunch of fixup commits to address some of the comments. Perhaps this one is now getting closer?

src/broker/overlay.c
garlick: Sorry, you're right!

garlick (approved):
LGTM, thanks!

grondo:
Ok, I've squashed all fixups, forced a push, and will set MWP. Thanks.

Issue #3275, opened 2020-10-22
python JobspecV1: allow resource with count of 0


SteVwonder:
From @andre-merzky:

specifying
```json
{ 
  "count": 0, 
  "type": "gpu" 
}
```
in the resource requirements triggers an error (count needs to be >0):
```
1602870315.836 : agent_scheduling.0000 : 13293 : 139644234123008 : ERROR   : flux submission failed
Traceback (most recent call last):
 File "/home/merzky/radical/radical.pilot.test/ve3/lib/python3.8/site-packages/radical/pilot/agent/scheduler/flux.py", line 93, in work
   spec = self._task_to_flux(uid, cud, sbox)
 File "/home/merzky/radical/radical.pilot.test/ve3/lib/python3.8/site-packages/radical/pilot/agent/scheduler/flux.py", line 246, in _task_to_flux
   js = flux_job.JobspecV1(tasks=spec['tasks'],
 File "/home/merzky/projects/flux/install/lib/python3.6/site-packages/flux/job/Jobspec.py", line 548, in __init__
   super(JobspecV1, self).__init__(resources, tasks, **kwargs)
 File "/home/merzky/projects/flux/install/lib/python3.6/site-packages/flux/job/Jobspec.py", line 143, in __init__
   self._validate_resource(res)
 File "/home/merzky/projects/flux/install/lib/python3.6/site-packages/flux/job/Jobspec.py", line 202, in _validate_resource
   raise ValueError("count must be > 0")
ValueError: count must be > 0
```

I would suggest to allow 0 and treat it accordingly, as that simplifies template based job specs.  Otherwise one needs to either assembly templates on the fly from parts, depending on what resources are needed, or switch template types altogether.

PR #4259, opened 2022-04-02
python: allow resource count of 0 in Jobspec v1

grondo:
When going through old issues I noticed #3275, which seemed like a reasonable and easy fix.

This change allows a count of 0 for core and gpu resources (not sure if core is all that useful, but this addresses the main point of the issue)

### DIFF
--- a/src/bindings/python/flux/job/Jobspec.py
+++ b/src/bindings/python/flux/job/Jobspec.py
@@ -219,6 +219,7 @@ class Jobspec(object):
             raise ValueError("operator must be one of {}".format(valid_operator_values))
 
     @classmethod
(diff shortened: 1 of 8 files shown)


grondo:
I force-pushed an update that reduces the changes here:

 - dropped the change to allow "complex" resource count specification (min, max) to be zero (this was probably gratuitous)
 - restrict the allowable resource types that can have a count of 0 to anything that is not `node`, `slot` or `core`

Also added a test to ensure a `gpu` count of 0 passes validation.

garlick (approved):
LGTM!

grondo:
Thanks! Setting MWP.

Issue #4251, opened 2022-03-31
job-archive should use statedir path by default


garlick:
Problem: currently the path to `job-archive`'s database file has to be explicitly set, placing an unnecessary burden on system administrators.

In most cases this file could just be co-located with `content.sqlite` in `statedir`.

Instead of unconditionally loading the module in rc1, we could conditionally load it only if it has a TOML configuration
```sh
if test $RANK -eq 0 && flux config get --type=object archive >/dev/null; then
   flux module load job-archive
fi
```
Then  in `job-archive`, instead being inactive if the `dbpath` key is not set, try to use `$statedir/job-archive.sqlite` and fail if a `dbpath` is not configured AND `statedir` is not set.

It looks like flux accounting also hardwires this path, based on our instructions [here](https://flux-framework.readthedocs.io/en/latest/adminguide.html#automatic-updates).  Maybe the `flux account update-usage` script could make the ` --job-archive_db_path` argument optional and by default try to use `$(flux getattr statedir)/job-archive.sqlite`?

chu11:
> Instead of unconditionally loading the module in rc1, we could conditionally load it only if it has a TOML configuration 

Hmmm, so one of the issues here is that that `dbpath` is really the only "important" config in the archive table.  there's some minor configs like period & busytimeout, but most will probably just use the defaults and not want to setup a toml configuration.

ideas

- add a `enable` or similar setting.  doesn't necessarily have to be under `[archive]` could be under `job-manager` or something (like a `archive-jobs`).

- similar to https://github.com/flux-framework/flux-core/pull/4070/commits/4bb85dd3c053072b6e04eea59c40e814415c5ac9, detect if we're the system instance?  perhaps we'd need a new config for the system instance

will ponder more

garlick:
Maybe the period could be required?

grondo:
> Maybe the period could be required?

FWIW, I like that idea.

chu11:
i hadn't thought of that, ok, lets go with making `period` required

chu11:
hmm should we remove config via the module args?  seems like it has little value if we're now requiring config via toml.

grondo:
I don't think config via toml is required, just that a defined `period` would be required. Therefore in a test instance you could still `flux module load job-archive period=10m`, but I suppose if you left off `period=` then the module load would fail.

chu11:
[...]

PR #4260, opened 2022-04-02
job-archive: use statedir path if dbpath is not set

chu11:
Per discussion in #4251.

- make `period` required in `job-archive` to make it do its thing
- if `dbpath` is not set, assume a path in `statedir` for job-archive, otherwise error
- in `rc1`, don't load `job-archive` module unless we're rank 0 & `archive.period` has been set

i have one commit below that is purely anal retentive, it can be dropped :-)

### DIFF
--- a/etc/rc1
+++ b/etc/rc1
@@ -30,7 +30,10 @@ modload 0 cron sync=heartbeat.pulse
 modload 0 job-manager
 modload all job-info
 modload 0 job-list
-modload 0 job-archive
+period=`flux config get --default= archive.period`
+if test $RANK -eq 0 -a -n "${period}"; then
+    flux module load job-archive
+fi
 
 if test $RANK -eq 0 && ! flux startlog --check --quiet; then
     flux queue stop
--- a/src/modules/content-files/content-files.c
+++ b/src/modules/content-files/content-files.c
@@ -277,8 +277,10 @@ static struct content_files *content_files_create (flux_t *h)
      */
     if (!(dbdir = flux_attr_get (h, "statedir")))
         dbdir = flux_attr_get (h, "rundir");
-    if (!dbdir)
+    if (!dbdir) {
         flux_log_error (h, "neither statedir nor rundir are set");
+        goto error;
+    }
     if (asprintf (&ctx->dbpath, "%s/content.files", dbdir) < 0)
         goto error;
     if (mkdir (ctx->dbpath, 0700) < 0 && errno != EEXIST) {
--- a/src/modules/job-archive/job-archive.c
+++ b/src/modules/job-archive/job-archive.c
@@ -58,8 +58,8 @@ const char *sql_since = "SELECT MAX(t_inactive) FROM jobs;";
 
 struct job_archive_ctx {
     flux_t *h;
-    char *dbpath;
     double period;
+    char *dbpath;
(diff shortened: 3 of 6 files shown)


src/modules/content-files/content-files.c
```diff
@@ -277,8 +277,10 @@ static struct content_files *content_files_create (flux_t *h)
      */
     if (!(dbdir = flux_attr_get (h, "statedir")))
         dbdir = flux_attr_get (h, "rundir");
-    if (!dbdir)
+    if (!dbdir) {
         flux_log_error (h, "neither statedir nor rundir are set");
+        goto error;
+    }
```
garlick, line 283: Thanks for catching this!  Wow don't know how I managed to do that not once but twice.

etc/rc1
```diff
@@ -30,7 +30,10 @@ modload 0 cron sync=heartbeat.pulse
 modload 0 job-manager
 modload all job-info
 modload 0 job-list
-modload 0 job-archive
+period=`flux config get --default= archive.period`
+if test $RANK -eq 0 -a -n "${period}"; then
+    modload 0 job-archive
```
garlick, line 35: Since the rank has already been checked (and that's all `modload()` does), probably better to just call `flux module load job-archive` here.

src/modules/job-archive/job-archive.c
```diff
@@ -543,10 +555,11 @@ int mod_main (flux_t *h, int ac, char **av)
...
-    process_config (ctx);
+    if (process_config (ctx) < 0)
+        goto done;
 
-    /* we do nothing if no dbpath specified */
-    if (ctx->dbpath) {
+    /* we do nothing if period not set */
+    if (ctx->period > 0.0) {
         if (job_archive_init (ctx) < 0)
             goto done;
```
garlick, line 564: Shouldn't it be an error if the period is not set?  Is loading the module but having it do nothing a valid use case?

garlick (changes requested):
Thanks @chu11 - looking good.  Just a couple of things that I think need to be addressed and/or discussed.

src/modules/job-archive/job-archive.c
chu11: ahh yes, I was keeping the style that was previously there, but we should error out instead.

chu11:
re-pushed, addressing comments above.  the issues were pretty tiny so just squashed.  The check for a valid period was put into the `process_config()` function.

```
--- a/src/modules/job-archive/job-archive.c
+++ b/src/modules/job-archive/job-archive.c
@@ -544,6 +544,12 @@ static int process_config (struct job_archive_ctx *ctx)
         else
             ctx->busy_timeout = (int)(1000 * tmp);
     }
+
+    /* period is required to be set */
+    if (ctx->period == 0.0) {
+        flux_log_error (ctx->h, "period not set");
+        return -1;
+    }
     return 0;
 }
```

garlick (approved):
LGTM!  Thanks.

chu11:
re-pushed, added 1 more test just trying to get more coverage.  If all is good I'll set MWP.

chu11:
one more re-push with another test, we're getting closer to 80% :-) (but might not get there)

chu11:
whew, hit 80% exactly

PR #4262, opened 2022-04-02
flux-hwloc: remove command

garlick:
`flux hwloc topology` is using the `resource.get-xml` RPC which is otherwise no longer needed because fluxion now has an Rv1 reader (see flux-framework/flux-sched#921)

It would be trivial to re-implement `flux hwloc topology` to obtain XML from each rank on demand, but since it can no longer create an actual aggregate machine description as of hwloc 2.0, it seems like this command is no longer very useful.  All it does is emit a concatenation of XML documents for each node.  `flux hwloc info` seems like it is also not that useful to us since we can get similar information for the instance from other sources.

This PR removes the whole `flux hwloc` command.


grondo:
I'm all for this! :+1:

There are still some tutorials floating around out there that reference `flux hwloc`. We should update those that we can.

e.g.

 * https://hpc-tutorials.llnl.gov/flux/section1/
 * https://flux-framework.readthedocs.io/en/latest/quickstart.html
 * https://hpc.llnl.gov/banks-jobs/running-jobs/batch-system-cross-reference-guides

garlick:
Now that we have `flux resouce info` (thanks @grondo) perhaps this can be merged?  Then we'll have until next month's release to complete documentation updates.

grondo (approved):
Yes, that sounds good to me!

PR #4268, opened 2022-04-04
libsubprocess: remove prefix on server setup

### LINKED ISSUE #4176: libsubprocess: make `flux_rexec()` topic configurable OR remove server topic configuration
In `libsubprocess` the `flux_rexec()` function sends remote execs to the `broker.rexec` topic.

The `flux_server_start()` function takes a prefix to determine what the server target `<prefix>.rexec` will be called.  The current flux broker sets the prefix to `broker`, so everything works.

This is inconsistent, presumably one of these should be adjusted.  The removal of the `prefix` seems the easier path.  I cannot find a reason why the `prefix` is supported, it may have been a legacy reason that is now forgotten.

chu11:
Problem: The flux_subprocess_server_start() function takes
a prefix for setting up server targets.  However the flux_rexec()
function is not configurable on what target should be used to
launch a remote subprocess.  This inconsistency is likely due
to a forgotten design decision / hope from long ago.

Solution: Remove the prefix for server setup and always internally
assume the prefix is "broker".

Fixes #4176

### DIFF
--- a/src/broker/exec.c
+++ b/src/broker/exec.c
@@ -92,7 +92,7 @@ int exec_initialize (flux_t *h, uint32_t rank, attr_t *attrs)
 
     if (attr_get (attrs, "local-uri", &local_uri, NULL) < 0)
         goto cleanup;
-    if (!(s = flux_subprocess_server_start (h, "broker", local_uri, rank)))
+    if (!(s = flux_subprocess_server_start (h, local_uri, rank)))
         goto cleanup;
     if (rank == 0)
         flux_subprocess_server_set_auth_cb (s, reject_nonlocal, NULL);
--- a/src/common/libsubprocess/server.c
+++ b/src/common/libsubprocess/server.c
@@ -608,45 +608,37 @@ error:
     json_decref (procs);
 }
 
-int server_start (flux_subprocess_server_t *s, const char *prefix)
+int server_start (flux_subprocess_server_t *s)
 {
     /* rexec.processes is primarily for testing */
     struct flux_msg_handler_spec htab[] = {
-        { FLUX_MSGTYPE_REQUEST, "rexec",        server_exec_cb, 0 },
-        { FLUX_MSGTYPE_REQUEST, "rexec.write",  server_write_cb, 0 },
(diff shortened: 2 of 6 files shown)


chu11:
an aside, this inconsistency has been annoying me, so just decided to fix it.  can revisit in the future if we ever need to add this support for real (would probably require change to both `flux_subprocess_server_start()` and `flux_rexec()`).

chu11:
> I have a branch from awhile ago that uses this functionality.

Ohhh I didn't realize that you were actually using it.  How are you handling config of the target via `flux_rexec()`?

grondo:
I had forgotten too until I remembered. I added a `flux_cmd_setopt` option, e.g.

```c
    flux_cmd_setopt (cmd, "service", "shell-123456.rexec");
```

chu11:
> I had forgotten too until I remembered. I added a flux_cmd_setopt option, e.g.

ahh ok.  Well, I'll leave it to you.  Do you think you'll need it with reasonable probability in near future?  We can drop this PR of course.

grondo:
I'm not sure there's a use case for it. We were thinking it could maybe be useful for co-locating services like debuggers, but I'm not sure that's needed since we also have "alloc-bypass" capability. We can merge this PR if you'd like and I'll just add a revert to my stack of commits. If we never do add the exec functionality to the shell, it might be nice to have this cleanup in place anyway...

chu11:
> it might be nice to have this cleanup in place anyway...

Yeah, and if we really need this functionality, I figure we need to think a bit on how to do this with `flux_rexec()`.  I sorta didn't want to think about it right now, which is why I went in this direction vs "allow alternate targets with rexec" path.

grondo (approved):
Seems good to me. We can easily add the prefix back if we ever need it.

grondo:
>  I figure we need to think a bit on how to do this with flux_rexec(). I sorta didn't want to think about it right now, which is why I went in this direction vs "allow alternate targets with rexec" path.


Yeah, and I wouldn't want to add the flux_rexec() changes without a use case, along with all the extra required tests, etc. Let's just obliterate this for now and resurrect it if necessary in the future.

Issue #4266, opened 2022-04-04
add flux-resource info subcommand


garlick:
Problem: #4262 proposed to remove `flux hwloc` but `flux hwloc info`, which prints a one line summary of resources, is a pretty useful command that is prominent in our documentation.

@grondo had the idea that maybe a `flux resource info` command that produces similar output could be provided, and that this is pretty close:
```
 grondo@fluke108:~$ flux resource list -s all -no '{nnodes} Nodes, {ncores} Cores, {ngpus} GPUs'
101 Nodes, 404 Cores, 0 GPUs
```

IMHO this is a good idea and will be a simpler transition for documentation too.

PR #4272, opened 2022-04-06
add flux-resource info subcommand

grondo:
This PR adds a `flux resource info` subcommand as described in #4266.
e.g.
```
grondo@fluke108:~/git/flux-core.git$ src/cmd/flux resource info
101 Nodes, 404 Cores, 0 GPUs
grondo@fluke108:~/git/flux-core.git$ src/cmd/flux resource info -s up
89 Nodes, 356 Cores, 0 GPUs
grondo@fluke108:~/git/flux-core.git$ src/cmd/flux resource info -s down
12 Nodes, 48 Cores, 0 GPUs
```

The implementation is just a thin wrapper around `flux resource list`, i.e. the `sched.resource-status` RPC. I thought at first it would be more appropriate to use the `resource.status` RPC but due to the structure of the code it was much easier to use the scheduler-based call. Therefore, this command won't work if a scheduler is not loaded.

### DIFF
--- a/doc/man1/flux-resource.rst
+++ b/doc/man1/flux-resource.rst
@@ -55,6 +55,12 @@ COMMANDS
    as "down" due to its simplified interface with the resource service defined
    by RFC 27.
 
+**info** [-s STATE,...]
+   Show a brief, single line summary of scheduler view of resources.
+   With *-s, --states=STATE,...*, limit the output to specified resource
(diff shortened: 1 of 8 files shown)


garlick (approved):
Nice!  Thanks!

grondo:
Thanks! Setting MWP

PR #4275, opened 2022-04-07
testsuite: fix failing tests on parallel testsuite runs

chu11:
I was regularly hitting two failures when I did `make -j16 check`.  Here are suggested fixes.

Note, that I sort of "jumped" big on my timeouts / input times.  Could be tweaked lower if people want them to be lower.


garlick (approved):
Lgtm!

Issue #4276, opened 2022-04-09
broker should not accept new connections during and after SHUTDOWN state


garlick:
Problem: while testing purging of inactive jobs during `job-manager` module unload (which caused the module to take a bit to unload), I ran into this problem where nodes that have disconnected in SHUTDOWN state manage to reconnect during FINALIZE state:
```
$ sudo flux shutdown
broker.info[0]: cleanup.0: flux queue stop --quiet Exited (rc=0) 0.0s
broker.info[0]: cleanup.1: flux job cancelall --user=all --quiet -f --states RUN Exited (rc=0) 0.0s
broker.info[0]: cleanup.2: flux queue idle --quiet Exited (rc=0) 0.0s
broker.info[0]: cleanup-success: cleanup->shutdown 0.247065s
broker.info[0]: online: picl[0,2,5-7] (ranks 0,2,5-7)
broker.info[0]: online: picl[0,2,6-7] (ranks 0,2,6-7)
broker.info[0]: online: picl[0,2,6] (ranks 0,2,6)
broker.info[0]: children-complete: shutdown->finalize 0.241157s
broker.info[0]: online: picl[0,2] (ranks 0,2)
broker.info[0]: online: picl0 (ranks 0)
broker.info[0]: overlay auth picl0 OK
broker.info[0]: accepting connection from picl7 (rank 7) uuid eba0c2f8-1724-48df-a830-80297ff579a4 status full
broker.info[0]: overlay auth picl0 OK
broker.info[0]: overlay auth picl0 OK
broker.info[0]: accepting connection from picl2 (rank 2) uuid 866306e8-c381-4c2b-a85a-d794702426e8 status full
broker.info[0]: accepting connection from picl6 (rank 6) uuid c1216d87-15fa-4653-9756-957df1864f5a status full
broker.info[0]: overlay auth picl0 OK
broker.info[0]: accepting connection from picl3 (rank 3) uuid a655dab1-ad90-4b5c-b8d2-6b6e7d17c1be status full
broker.info[0]: overlay auth picl0 OK
broker.info[0]: accepting connection from picl5 (rank 5) uuid 482fccfb-abc8-4b6d-94e7-b40a5f998fb5 status full
job-manager.info[0]: purge: removed 10005 of 10006 inactive jobs
broker.info[0]: rc3.0: /usr/local/etc/flux/rc3 Exited (rc=0) 12.8s
broker.info[0]: rc3-success: finalize->goodbye 12.8915s
```
[...]

PR #4277, opened 2022-04-09
broker: prevent downstream peers from connecting during shutdown

garlick:
This PR addresses #4276, where in a system instance, downstream brokers are observed to partially wire up during shutdown.  The problem is that systemd may respawn a broker that was just shut down before the its parent has completely finalized, which creates extra log noise on parent and child.

Call `zsock_unbind()` upon entering the FINALIZE state, where we know all downstream peers have already disconnected, and we start to execute the rc3 script.

In addition, this adds a little state machine cleanup mainly for the purpose of making the logs look more intuitive during the shutdown process (which is more visible now that we have [flux-shutdown(1)](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man1/flux-shutdown.html).


grondo (approved):
LGTM!

garlick:
Thanks!

PR #4278, opened 2022-04-09
libflux: convert flux_conf_error_t to flux_error_t

garlick:
Problem: #4165 introduced a common type for returning human readable errors from Flux API calls, but `flux_conf_unpack()` and related functions were not converted.  Furthermore, the file and line structure members in `flux_conf_error_t` are pretty useless after the initial parsing of TOML files since most of the time these functions access the JSON config object and any unpack error would not generate a useful file and line.

Drop the `flux_conf_error_t` type from the API and convert the `flux_conf_*()` functions to use `flux_error_t`.

Update tests and users.

I did not find any users in `flux-sched`, `flux-accounting`, `flux-coral2`, or `flux-pmix` so at least nothing breaks in those projects as a result of this change.

### DIFF
--- a/src/broker/boot_config.c
+++ b/src/broker/boot_config.c
@@ -175,7 +175,7 @@ int boot_config_parse (const flux_conf_t *cf,
                        struct boot_conf *conf,
                        json_t **hostsp)
 {
-    flux_conf_error_t error;
+    flux_error_t error;
     const char *default_bind = NULL;
     const char *default_connect = NULL;
     json_t *hosts = NULL;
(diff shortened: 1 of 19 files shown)


grondo (approved):
Ah, good idea to just include file and line info in the textual error when it is available. Apologies for not taking this extra step when `flux_error_t` was introduced! :facepalm: 

Seems reasonable and LGTM!

garlick:
Thanks!  I tacked on some tests to improve the diff coverage a bit. I'll go ahead and set MWP.

PR #4279, opened 2022-04-10
job-manager: introduce config callback

garlick:
Problem: the job manager has multiple subsystems that consume the TOML configuration object from the broker, but there is no mechanism for them to receive a callback when the configuration changes (e.g. sys admin calls `flux config reload` or `systemctl reload flux`).

To prepare to support a dynamically configurable purge policy in a future PR, this PR introduces a way for job manager subsystems to register a callback for configuration handling:
```c
/* Return value:
 *   0=success, one-shot
 *  -1=falure (set 'error' but not errno)
 *   1=success, continue to invoke callback on config updates
 */
typedef int (*conf_update_f)(const flux_conf_t *conf,
                             flux_error_t *error,
                             void *arg);
/* Immediately call 'cb' on current config object, and then on config updates
 * as indicated by initial callback's return value (see above).
 */
int conf_register_callback (struct conf *conf,
                            flux_error_t *error,
                            conf_update_f cb,
                            void *arg);
void conf_unregister_callback (struct conf *conf, conf_update_f cb);
```
The two existing users of TOML config are changed to use this.  As a demo the max journal size (renamed from `job-manager.events_maxlen`  to `job-manager.journal-size-limit`) can now be changed on the fly.

This could be generalized for all broker modules, but without a compelling use case elsewhere, I thought it best to hold off doing anything like that for now, minimizing the surface area of the configuration API so it will be easier to fix/replace later.  We can always generalize later if a need arises.

### DIFF
--- a/src/modules/job-manager/jobtap.c
+++ b/src/modules/job-manager/jobtap.c
@@ -32,6 +32,7 @@
 
 #include "annotate.h"
 #include "prioritize.h"
+#include "conf.h"
 #include "event.h"
 #include "jobtap.h"
 #include "jobtap-internal.h"
@@ -391,10 +392,11 @@ static int jobtap_conf_entry (struct jobtap *jobtap,
     return 0;
 }
 
-static int jobtap_parse_config (struct jobtap *jobtap,
-                                const flux_conf_t *conf,
-                                flux_error_t *errp)
+static int jobtap_parse_config (const flux_conf_t *conf,
+                                flux_error_t *errp,
+                                void *arg)
 {
+    struct jobtap *jobtap = arg;
     json_t *plugins = NULL;
     flux_error_t error;
     json_t *entry;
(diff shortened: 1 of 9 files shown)


src/modules/job-manager/journal.c
```diff
@@ -20,7 +20,9 @@
 #include "src/common/libczmqcontainers/czmq_containers.h"
 #include "src/common/libeventlog/eventlog.h"
 #include "src/common/libutil/errno_safe.h"
+#include "src/common/libutil/errprintf.h"
```
grondo, line 23: typo (double "now") in commit message for 18bb02cad4cf372ab27b481a5e6c9e0f82a45680:

>  but now job-manager now supports

src/modules/job-manager/jobtap.c
```diff
@@ -32,6 +32,7 @@
 
 #include "annotate.h"
 #include "prioritize.h"
+#include "conf.h"
```
grondo, line 35: typo (double "now") in commit message for 4c3545e7d91180293c733250900f74440add2292

> but now job-manager now supports

grondo (approved):
LGTM, just noted a couple very minor typos in commit messages.

garlick:
Fixed those typos, forced a push, and set MWP.  Thanks!

PR #4282, opened 2022-04-13
libev: update to version 4.33

garlick:
Problem: libev has undergone some development and bugfixes since the last time we updated our embedded copy.

Update from 4.27 to 4.33.

Changes are:

4.33 Wed Mar 18 13:22:29 CET 2020
        - no changes w.r.t. 4.32.

4.32 (EV only)
        - the 4.31 timerfd code wrongly changed the priority of the signal
          fd watcher, which is usually harmless unless signal fds are
          also used (found via cpan tester service).
        - the documentation wrongly claimed that user may modify fd and events
          members in io watchers when the watcher was stopped
          (found by b_jonas).
        - new ev_io_modify mutator which changes only the events member,
          which can be faster. also added ev::io::set (int events) method
          to ev++.h.
        - officially allow a zero events mask for io watchers. this should
          work with older libev versions as well but was not officially
          allowed before.
        - do not wake up every minute when timerfd is used to detect timejumps.
        - do not wake up every minute when periodics are disabled and we have
          a monotonic clock.
        - support a lot more "uncommon" compile time configurations,
          such as ev_embed enabled but ev_timer disabled.
        - use a start/stop wrapper class to reduce code duplication in
          ev++.h and make it needlessly more c++-y.
        - the linux aio backend is no longer compiled in by default.
        - update to libecb version 0x00010008.

4.31 Fri Dec 20 21:58:29 CET 2019
        - handle backends with minimum wait time a bit better by not
          waiting in the presence of already-expired timers
          (behaviour reported by Felipe Gasper).
        - new feature: use timerfd to detect timejumps quickly,
          can be disabled with the new EVFLAG_NOTIMERFD loop flag.
        - document EV_USE_SIGNALFD feature macro.

4.30 (EV only)
        - change non-autoconf test for __kernel_rwf_t by testing
          LINUX_VERSION_CODE, the most direct test I could find.
        - fix a bug in the io_uring backend that polled the wrong
          backend fd, causing it to not work in many cases.

4.29 (EV only)
        - add io uring autoconf and non-autoconf detection.
        - disable io_uring when some header files are too old.

4.28 (EV only)
        - linuxaio backend resulted in random memory corruption
          when loop is forked.
        - linuxaio backend might have tried to cancel an iocb
          multiple times (was unable to trigger this).
        - linuxaio backend now employs a generation counter to
          avoid handling spurious events from cancelled requests.
        - io_cancel can return EINTR, deal with it. also, assume
          io_submit also returns EINTR.
        - fix some other minor bugs in linuxaio backend.
        - ev_tstamp type can now be overriden by defining EV_TSTAMP_T.
        - cleanup: replace expect_true/false and noinline by their
          libecb counterparts.
        - move syscall infrastructure from ev_linuxaio.c to ev.c.
        - prepare io_uring integration.
        - tweak ev_floor.
        - epoll, poll, win32 Sleep and other places that use millisecond
          reslution now all try to round up times.
        - solaris port backend didn't compile.
        - abstract time constants into their macros, for more flexibility.

### DIFF
--- a/src/common/libev/LICENSE
+++ b/src/common/libev/LICENSE
@@ -1,0 +1,37 @@
+All files in libev are
+Copyright (c)2007,2008,2009,2010,2011,2012,2013 Marc Alexander Lehmann.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
+
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+
+    * Redistributions in binary form must reproduce the above
+      copyright notice, this list of conditions and the following
+      disclaimer in the documentation and/or other materials provided
+      with the distribution.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+"AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+Alternatively, the contents of this package may be used under the terms
+of the GNU General Public License ("GPL") version 2 or any later version,
+in which case the provisions of the GPL are applicable instead of the
+above. If you wish to allow the use of your version of this package only
+under the terms of the GPL and not to allow others to use your version of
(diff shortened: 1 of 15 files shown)


chu11 (approved):
seems like a good idea to update!

garlick:
Hmm, just hit this in bionic and el8 builders.  I'll restart just to see if it's reproduceable or not.

```
expecting success: 
      $set_faketime 2016-06-04 15:30:00 &&
      id=$(echo "15 * * * * flux event pub t.cron.complete" | flux_cron tab) &&
      next=$(date +%s --date="2016-06-04 16:15:00") &&
      cron_entry_check ${id} type datetime &&
      cron_entry_check ${id} stopped false &&
      cron_entry_check ${id} typedata.next_wakeup ${next} &&
      echo sleeping at $(date) &&
      ${event_trace} t.cron t.cron.complete \
          $set_faketime 2016-06-04 16:15:00 &&
      echo done at $(date) &&
      cron_entry_check ${id} stats.count 1 &&
      flux cron delete ${id}
  
  timestamp file: @2016-06-04 15:30:00: date is now Sat Jun 4 15:29:59 UTC 2016
  cron-1: type=datetime, wanted datetime
  cron-1: stopped=false, wanted false
  cron-1: typedata.next_wakeup=1465056900, wanted 1465056900
  sleeping at Sat Jun 4 15:29:59 UTC 2016
  timestamp file: @2016-06-04 16:15:00: date is now Sat Jun 4 16:14:59 UTC 2016
  /usr/src/t/scripts/event-trace.lua timed out after 5.0s
  not ok 3 - flux-cron tab works for set minute
```

grondo (approved):
LGTM, too.

I'll see if I can reproduce the libfaketime/cron test failures locally. Cron does make use of reactor watchers heavily, and libfaketime also has to be able to influence the time in a way that will get ev timer and periodic watchers to wake up, so hopefully the libev updates didn't perturb this tenuous testing situation.

grondo:
I could not reproduce locally, so I went ahead and restarted the failed builders. Hope that was ok!

garlick:
I was able to reproduce it locally in a fedora35 docker image, running
```
$ while ./t0016-cron-faketime.t -v; do /bin/true; done
```
in just 2 or 3 iterations.

grondo:
Ok, I did try a few iterations, but will look again tomorrow.

garlick:
Looks like the tests all passed, thanks @grondo for the fix!

I reversed the order of the commits so git bisect won't land on a non-working commit, forced a push, and set MWP.

PR #4283, opened 2022-04-13
fix hang when job with an interactive pty fails early

grondo:
This PR fixes an issue that can cause `flux job attach` to hang when attaching to an interactive pty and the job gets an exception before the pty client is fully "attached". This occurs specifically when a non-existent command is supplied, for example:
```console
$ flux mini run -o pty.interactive foo
<hang>
```

This PR reworks the `attach_pty()` code so that the attach can be performed asynchronously, and adds an accessor so that the client can query if it is attached yet. When a job exception occurs while the pty is not yet attached, we can avoid the hang by destroying the pty client.

Now the command above no longer hangs:
```console
$ flux mini run -o pty.interactive foo
0.079s: flux-shell[0]: FATAL: task 0 (host eel): start failed: foo: No such file or directory
0.083s: job.exception type=exec severity=0 task 0 (host eel): start failed: foo: No such file or directory
```

Some other minor rework was required to allow the pty attach to occur asynchronously.

### DIFF
--- a/src/cmd/flux-job.c
+++ b/src/cmd/flux-job.c
@@ -2098,6 +2098,17 @@ void attach_event_continuation (flux_future_t *f, void *arg)
                          type,
                          severity,
                          note ? note : "");
+
+        /*  If this job has an interactive pty and the pty is not yet attached,
+         *   destroy the pty to avoid a potential hang attempting to connect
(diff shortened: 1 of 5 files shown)


garlick (approved):
Nicely done, LGTM!

PR #4284, opened 2022-04-14
libutil: support "ms" suffix for Flux Standard Duration

grondo:
This PR adds support for the `ms` suffix in `fsd_parse_duration(3)` and `fsd_format_duration(3)` now that the changes to [RFC 23](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_23.html) allowing that suffix have been merged. Mainly this is useful when small durations are printed, e.g. in logs:

before
```
 parent-none: join->init 3.312e-05s
```

after
```
 parent-none: join->init 0.169518ms
```

(Note these were from real examples, so the numbers are not meant to match)

The `ms` suffix may be useful in other situations as well, so it seemed prudent to go ahead and add support for it.

### DIFF
--- a/src/common/libutil/fsd.c
+++ b/src/common/libutil/fsd.c
@@ -18,6 +18,7 @@
 #include <stdlib.h>
 
 #include "fsd.h"
+#include "src/common/libccan/ccan/str/str.h"
 
 static int is_invalid_duration (double d)
 {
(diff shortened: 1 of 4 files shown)


src/common/libutil/fsd.c
```diff
@@ -86,7 +86,9 @@ int fsd_format_duration_ex (char *buf,
         errno = EINVAL;
         return -1;
     }
-    if (duration < 60.)
+    if (duration < 0.1 && duration != 0.)
```
garlick, line 89: Is there a reason the threshold is 0.1 and not 1?

I can't decide if I'd rather see 0.750s or 750ms but I kind of lean towards the latter.  I guess it could go either way!

garlick (approved):
LGTM!  Just one question in line that I really don't have a strong opinion on, so feel free to ignore and set MWP.

src/common/libutil/fsd.c
grondo: I vacillated as well. However, the reason for the addition of ms was for very small durations, and since we've converted most small durations to seconds until recently, it seemed like minimizing the conversion to `ms` until it really mattered made the most sense here. (and `.5s` is much better than `500ms` IMO).

I guess at least a comment would make sense.

grondo: I added the following comment above this statement:
```c
    /*  We'd rather present a result a seconds if possible, since that
     *  is the base unit of FSD. However, if the duration is very small,
     *  present in milliseconds since the result will be easier for a
     *  human to read. E.g. 62.1ms vs 0.0621s, or more importantly
     *  0.0123ms vs 1.23e-05s.
     */
```

grondo: Thanks! Will fix!

grondo:
Thanks! Setting MWP.

Issue #4273, opened 2022-04-06
implement prototype job-db module


garlick:
Problem: the `job-list` module's memory cache grows without bound, and tracks KVS which we need to start pruning.  The `job-archive` module is mainly structured to feed data to `flux-accounting` and offers no on-line query support.

We want to answer the question:  is it feasible to develop a new module that answers `job-list` queries exactly as the `job-list` module does now, but instead of _only_ caching jobs in memory, retains inactive job data in a persistent database?

Some notes from our discussions
- It could be called `job-db`
- It could register the `job-list` service so everything just works if it is loaded in place of `job-list` (even `job-archive` as is)
- It could consume the job manager journal and maintain an in-memory cache of active jobs
- Upon job transition to INACTIVE state, job data (now immutable) could be added to a sqlite database and dropped from the memory cache.
- Queries would seamlessly access the cache as well as the database
- Sqlite JSON capability may be of use for answering queries against R and jobspec without needing to break fields out to separate rows (see #4234)

The goal of prototyping here would be to quickly get something functional that we can discuss.

Automated testing could be deferred until we decide whether to move forward.

garlick:
A suggestion would be to leave off `guest.exec.eventlog`, stdio, annotations, and memos for this experiment.  We can always add later.

Also, a simplifying assumption may be that (unlike `job-list`) no kvs scan is needed on restart.  I think we can make that work by adding some way to flush the job manager journal to consumers and end with an ENODATA during the broker CLEANUP state, before rc3 runs.

chu11:
I hit a snag during my prototyping today and thought I'd do a thought dump before moving forward:

- the major snag: the job-list/job-db module can be sent the journal event for "job inactive" (eventlog "clean" event) BEFORE the eventlog "clean" event is written to the KVS.  So there's a race condition here, which could lead to further complications down the road.
  - e.g. i thought if I implemented a "inactive cache list", that might solve the problem, i.e. don't dump to archive until later once you know all the data is in the KVS.  But if there was a system crash, we might lose inactive data b/c that data hasn't been flushed to the archive yet.  But we gotta make sure data is flushed to the archive before we can remove data from the KVS, yadda yadda yadda.

notes on technical things to resolve
  
- will have to redo how we do job stats, probably not hard, anything related to the inactive job list has to be converted into a sql query probably (could be complicated with inactive cache list given above).

[...]

PR #4286, opened 2022-04-15
job-manager: add inactive job purge capability

garlick:
This adds two methods of purging inactive jobs:

One can purge interactively, e.g.
```
$ flux job purge --num-limit=10
purged 90 inactive jobs, 10 remaining
```

or one can configure automatic purging via TOML
```toml
[job-manager]

inactive-age-limit = "7d"
inactive-num-limit = 10000
```

The job manager retains inactive jobs in a hash and places them in a list ordered by the time they became inactive.  When jobs are purged, the entry in the hash is removed along with the corresponding KVS directory.

### DIFF
--- a/src/modules/job-manager/purge.c
+++ b/src/modules/job-manager/purge.c
@@ -58,6 +58,20 @@ int purge_enqueue_job (struct purge *purge, struct job *job)
     return 0;
 }
 
+static int purge_publish (struct purge *purge, json_t *jobs)
+{
+    flux_future_t *f;
+
+    if (!(f = flux_event_publish_pack (purge->ctx->h,
+                                       "job-purge-inactive",
+                                       0,
+                                       "{s:O}",
+                                       "jobs", jobs)))
+        return -1;
+    flux_future_destroy (f);
+    return 0;
+}
+
 /* Return true if candidate job is eligible for purging based on
  * provided limits.  A job is purgeable if either limit is exceeded.
  */
@@ -107,12 +121,17 @@ static flux_future_t *purge_inactive_jobs (struct purge *purge,
     double now = flux_reactor_now (flux_get_reactor (purge->ctx->h));
     struct job *job;
     flux_kvs_txn_t *txn;
+    json_t *jobs = NULL;
     char key[64];
     flux_future_t *f = NULL;
     int count = 0;
 
     if (!(txn = flux_kvs_txn_create ()))
         return NULL;
+    if (!(jobs = json_array ())) {
+        errno = ENOMEM;
+        goto error;
+    }
     while ((job = zlistx_first (purge->queue)) && count < max_purge_count) {
         if (!purge_eligible (now - job->t_clean,
                              age_limit,
@@ -122,6 +141,12 @@ static flux_future_t *purge_inactive_jobs (struct purge *purge,
         if (flux_job_kvs_key (key, sizeof (key), job->id, NULL) < 0
             || flux_kvs_txn_unlink (txn, 0, key) < 0)
             goto error;
+        json_t *o = json_integer (job->id);
+        if (!o || json_array_append_new (jobs, o)) {
+            json_decref (o);
+            errno = ENOMEM;
+            goto error;
+        }
         (void)zlistx_delete (purge->queue, job->handle);
         job->handle = NULL;
         zhashx_delete (purge->ctx->inactive_jobs, &job->id);
@@ -132,9 +157,11 @@ static flux_future_t *purge_inactive_jobs (struct purge *purge,
         goto error;
     }
     if (!(f = flux_kvs_commit (purge->ctx->h, NULL, 0, txn))
-        || flux_future_aux_set (f, "count", int2ptr (count), NULL) < 0)
+        || flux_future_aux_set (f, "count", int2ptr (count), NULL) < 0
+        || purge_publish (purge, jobs) < 0)
         goto error;
     flux_kvs_txn_destroy (txn);
+    json_decref (jobs);
     /* N.B. if kvs commit fails, jobs are still removed from the hash/list.
      * It doesn't seem worth the effort to structure the code to avoid this
      * due to high complexity of solution, low probability of error, and
@@ -143,6 +170,7 @@ static flux_future_t *purge_inactive_jobs (struct purge *purge,
     return f;
 error:
     flux_kvs_txn_destroy (txn);
+    json_decref (jobs);
     flux_future_destroy (f);
     return NULL;
 }
--- a/src/modules/job-manager/purge.c
+++ b/src/modules/job-manager/purge.c
@@ -1,0 +1,431 @@
+/************************************************************\
+ * Copyright 2022 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* purge.c - remove old inactive jobs
+ */
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+#include <flux/core.h>
+#include <assert.h>
+
+#include "src/common/libutil/errprintf.h"
+#include "src/common/libutil/fsd.h"
+#include "src/common/libutil/errprintf.h"
+#include "src/common/libccan/ccan/ptrint/ptrint.h"
+
+#include "job-manager.h"
+#include "job.h"
+#include "purge.h"
+#include "conf.h"
+
+#define INACTIVE_NUM_UNLIMITED  (-1)
+#define INACTIVE_AGE_UNLIMITED  (-1.)
+
+struct purge {
+    struct job_manager *ctx;
+    double age_limit;
+    int num_limit;
+    zlistx_t *queue;
+    flux_future_t *f_sync;
+    flux_future_t *f_purge;
+
+    flux_msg_handler_t **handlers;
+    struct flux_msglist *requests;
+};
+
+static const int purge_batch_max = 100; // max KVS ops per txn
+
+/* Add an inactive job to the "purge queue".
+ * The queue is ordered by the time the job became inactive, so
+ * the first job in the queue is the oldest.
+ */
+int purge_enqueue_job (struct purge *purge, struct job *job)
+{
+    assert (job->handle == NULL);
+    if (!(job->handle = zlistx_insert (purge->queue, job, false))) {
+        errno = ENOMEM;
+        return -1;
+    }
+    return 0;
+}
+
+/* Return true if candidate job is eligible for purging based on
+ * provided limits.  A job is purgeable if either limit is exceeded.
+ */
+static bool purge_eligible (double age,
+                            double age_limit,
+                            int num,
+                            int num_limit)
+{
+    if (age_limit != INACTIVE_AGE_UNLIMITED && age > age_limit)
+        return true;
+    if (num_limit != INACTIVE_NUM_UNLIMITED && num > num_limit)
+        return true;
+    return false;
+}
+
+static int purge_eligible_count (struct purge *purge,
+                                 double age_limit,
+                                 int num_limit)
+{
+    double now = flux_reactor_now (flux_get_reactor (purge->ctx->h));
+    struct job *job;
+    int count = 0;
+
+    job = zlistx_first (purge->queue);
+    while (job) {
+        if (!purge_eligible (now - job->t_clean,
+                             age_limit,
+                             zlistx_size (purge->queue) - count,
+                             num_limit))
+            break;
+        count++;
+        job = zlistx_next (purge->queue);
+    }
+    return count;
+}
+
+/* Send a KVS commit containing unlinks for one or more inactive jobs.
+ * Return future if successful, with 'count' added to aux hash.
+ * Return NULL on failure with errno set.
+ * N.B. Failure with errno=ENODATA just means there were no eligible jobs.
+ */
+static flux_future_t *purge_inactive_jobs (struct purge *purge,
+                                           double age_limit,
+                                           int num_limit,
+                                           int max_purge_count)
+{
+    double now = flux_reactor_now (flux_get_reactor (purge->ctx->h));
+    struct job *job;
+    flux_kvs_txn_t *txn;
+    char key[64];
+    flux_future_t *f = NULL;
+    int count = 0;
+
+    if (!(txn = flux_kvs_txn_create ()))
+        return NULL;
+    while ((job = zlistx_first (purge->queue)) && count < max_purge_count) {
+        if (!purge_eligible (now - job->t_clean,
+                             age_limit,
+                             zlistx_size (purge->queue),
+                             num_limit))
+            break;
+        if (flux_job_kvs_key (key, sizeof (key), job->id, NULL) < 0
+            || flux_kvs_txn_unlink (txn, 0, key) < 0)
+            goto error;
+        (void)zlistx_delete (purge->queue, job->handle);
+        job->handle = NULL;
+        zhashx_delete (purge->ctx->inactive_jobs, &job->id);
+        count++;
+    }
+    if (count == 0) {
+        errno = ENODATA;
+        goto error;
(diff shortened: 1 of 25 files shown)


grondo:
This is great!

This does seem to have an impact on job throughput, though I'm not sure if it is something to worry about at this point.
For example, on master:
```console
$ flux jobs --stats-only
0 running, 32768 completed, 0 failed, 0 pending
$ src/test/throughput.py -sn 4096
  4096 jobs:   4096 submitted,      0 running, 4096 completed
number of jobs: 4096
submit time:    3.690 s (1110.1 job/s)
script runtime: 29.725s
job runtime:    29.180s
throughput:     140.4 job/s (script: 137.8 job/s)
```

vs this branch:
```console
$ flux jobs --stats-only
0 running, 32768 completed, 0 failed, 0 pending
$ src/test/throughput.py -sn 4096
  4096 jobs:   4096 submitted,      0 running, 4096 completed
number of jobs: 4096
submit time:    3.444 s (1189.3 job/s)
script runtime: 35.197s
job runtime:    34.764s
throughput:     117.8 job/s (script: 116.4 job/s)
```

Purging jobs (down to 1024) in this case restores the throughput:
(though of course job-list still shows all 36K jobs)

```console
$ flux job purge --num-limit=1024 --batch=100
purged 35840 inactive jobs, 1024 remaining
$ flux jobs --stats-only
0 running, 36864 completed, 0 failed, 0 pending
$ src/test/throughput.py -sn 4096
  4096 jobs:   4096 submitted,      0 running, 4096 completed
number of jobs: 4096
submit time:    3.719 s (1101.5 job/s)
script runtime: 27.975s
job runtime:    27.574s
throughput:     148.5 job/s (script: 146.4 job/s)
```

Since the purge is really only meant for the system instance, which is by design the least common instance type, I wonder if we should consider only enabling the job-manager inactive cache by configuration?

Having said that, the throughput numbers here may not even be relevant, since the throughput of real jobs is limited in other ways due to the reality of executing actual processes. However, I did want to bring this up since we've worked so hard to keep this numbers high in the past, I just wanted us to keep aware of the issue.

garlick:
Thank you for testing that!  That result is a little surprising to me.  I wonder if the list insertion into `purge->queue` is the problem.  Let me poke at that a bit.

grondo:
Some other random thoughts:

 * The dependency plugins in the job-manager could also make great use of the inactive job cache to allow jobs to have dependencies on other, completed jobs (right now jobs can only have dependencies on active jobs). This might only work if the job "result" for inactive jobs is available.
 * Apropos of the previous bullet, it might be nice to have a function to lookup a jobid that checks active hash and falls back to inactive hash. Callers can check the job state to determine if the job was active vs inactive to determine from which hash the job was found.
 * The job-list module of course doesn't purge its cache of inactive jobs. So, after a purge (currently) a job can be listed, but `flux job info JOBID` would fail (since the jobid is no longer in the kvs). That might be confusing, but probably isn't a big issue at this point.

grondo:
>  That result is a little surprising to me. 

Hm, you may also want to make sure the result is reproducible for you!

garlick:
Good observations!

> The job-list module of course doesn't purge its cache of inactive jobs. So, after a purge (currently) a job can be listed, but flux job info JOBID would fail (since the jobid is no longer in the kvs). That might be confusing, but probably isn't a big issue at this point.

I do have a development branch in which the journal is modified to include inactive jobs (optionally) so that job-list doesn't have to scan the KVS.  In that branch, the entire eventlog for each job is retained in memory  and so a new journal consumer can completely  "catch up" with the job manager and track ongoing job events in one stream.

In that scenario, it would make sense to add a message to the journal stream when a job gets purged even though it's not an event per se.  That would let a journal consumer like job-list keep the size of its caches in check without the need for something separate like #3688.  This is an "open loop" or eventually consistent design, so the situation you describe could still arise although it would be far less likely.

I should probably rebase that branch on top of this and post it as a WIP.  If nothing else it's something concrete to contribute to the discussion about #4273 (job-db).

grondo:
Thanks for that update! I apologize, I now realize my comment was a bit off-topic for this specific PR.

garlick:
> I now realize my comment was a bit off-topic for this specific PR.

Not at all - that was a good observation and a loose end I failed to articulate.

> Hm, you may also want to make sure the result is reproducible for you!

I'm not really seeing it.  Or at least the variability in results is too high to really conclude anything.  Here are throughput results (`src/test/throughput.py -sn 4096`) for master and purge branches;
```
                total           master TP       purge TP

+4k jobs        4k              288.4           300.6
+4k jobs        8k              235.7           280.3
+4k jobs        12k             233.3           224.7
+4k jobs        16k             269.8           244.7
+4k jobs        20k             225.7           229.9
+4k jobs        24k             224.6           247.7
+4k jobs        28k             252.5           226.2
+4k jobs        32k             235.7           246.1
```

I might be missing something or maybe I should have tried with more jobs?

grondo:
Yeah, I was not getting that much variability on my system. You could try more jobs since your system is clearly double the speed of mine. But if it is not reproducible for you, then this might not be an actual issue.

chu11:
high level comment, is this something we want to hold off on until we have job archiving (such as #4273) more in place?

or is the idea that since this configurable, the default behavior stays the same and we don't have to worry about archiving.  If users want to lose inactive data in the kvs permanently, its ok.

garlick:
> or is the idea that since this configurable, the default behavior stays the same and we don't have to worry about archiving. If users want to lose inactive data in the kvs permanently, its ok.

Yeah this is what I was thinking.  I might go a little further and recommend that our system instance early adopters purge at 1 week or so just to ensure that we don't have an unsustainable situation.  Then we'll add in some form of offline garbage collection and we will be on a healthy trajectory.  job-archive will grab all the data pretty aggressively so as long as the  configured purge age is on the order of days, we shouldn't be losing data as far as flux-accounting is concerned.

However it's up for debate.   Purging would limit the ability to ask what or who ran on a "bad node" beyond the purge period for example.  Maybe we can swoop in with a good job-db solution in the near term that would allow longer term preservation and querying of job data?

garlick:
I did reproduce the slowdown by testing with more jobs.  Here's some data that shows it on my system:
```
                total           master TP       purge TP
+32k jobs       32k             203.7           207.7
+8k jobs        40k             210.4           223.1
+8k jobs        48k             228.9           197.0
+8k jobs        56k             210.8           178.8
+8k jobs        64k             206.0           150.9
+8k jobs        72k             203.0           134.4
+8k jobs        80k             205.4           113.3
+8k jobs        88k             216.9           100.5
```

garlick:
I was using the list backwards so I had to scan the whole thing to insert the next job!  With that fixed (last column) the numbers at scale are better:
```
		total	master	purge	nolist	revlist
+32k jobs	32k	203.7	207.7	198.7	204.5
+8k jobs	40k	210.4	223.1	237.9	225.0
+8k jobs	48k	228.9	197.0	225.5	208.5
+8k jobs	56k	210.8	178.8	217.6	216.9
+8k jobs	64k	206.0	150.9	217.1	211.0
+8k jobs	72k	203.0	134.4	215.5	211.5
+8k jobs	80k	205.4	113.3	197.5	207.4
+8k jobs	88k	216.9	100.5	191.9	209.5
```
I'll post the fix shortly.  I wanted to add a test to ensure the oldest jobs are purged first since I was purging the youngest and no tests were catching it.

grondo:
Nice catch!

garlick:
Hmm, the bionic - py3.7,clang-6.0  builder seems to have gotten stuck in the broker unit tests (on `test_runat.t`)specifically, then hung for a while.  Restarting since this doesn't seem likely to be related to this pr.
```
2022-04-15T22:46:21.3560385Z make  check-TESTS
2022-04-15T22:46:21.5422684Z                test_attr.t:  PASS: N=53  PASS=53  FAIL=0 SKIP=0 XPASS=0 XFAIL=0
2022-04-15T22:46:21.5439307Z             test_pmiutil.t:  PASS: N=9   PASS=9   FAIL=0 SKIP=0 XPASS=0 XFAIL=0
2022-04-15T22:46:21.5592137Z             test_service.t:  PASS: N=18  PASS=18  FAIL=0 SKIP=0 XPASS=0 XFAIL=0
2022-04-15T22:46:21.5598217Z             test_liblist.t:  PASS: N=6   PASS=6   FAIL=0 SKIP=0 XPASS=0 XFAIL=0
2022-04-15T22:46:21.6522470Z         test_boot_config.t:  PASS: N=54  PASS=54  FAIL=0 SKIP=0 XPASS=0 XFAIL=0
2022-04-15T22:46:22.7412790Z             test_overlay.t:  PASS: N=83  PASS=83  FAIL=0 SKIP=0 XPASS=0 XFAIL=0
2022-04-15T23:42:25.5427971Z ##[error]The operation was canceled.
```

garlick:
It failed again but in the purge sharness test.  I realize I have a test that depends on a previous test that has the NO_CHAIN_LINT conditional, so adding that to the other dependent test.

garlick:
> The job-list module of course doesn't purge its cache of inactive jobs. So, after a purge (currently) a job can be listed, but flux job info JOBID would fail (since the jobid is no longer in the kvs). That might be confusing, but probably isn't a big issue at this point.

I had purging set to 8h on my test system and just spent a minute confused by the fact that `flux jobs` was still showing jobs that should have been purged.  Well I guess that brought home this point for me!

Hmm, I wonder if it would be OK to do something simple for now like publish the list of purged jobs so caches in job-list can be pruned?

garlick:
It was pretty easy to implement so I went ahead and pushed a change to have the job manager publish an event listing job IDs being purged, and then have job-list subscribe to that message and remove jobs.

This invites comparison to PR #3688 (pending forever), where a lot of work was done to implement job-list purging.  I think the job manager is the correct place to perform the purge since it also also solves the problem of the KVS growing without bound and is the "source of truth" for a lot of other job information.  But some good discussion happened over in that PR, e.g.
- `flux job list-purge` lists job info that is being purged (like dmesg -c)
- purged jobids are added by job-list to a hash so they can still be distinguished from invalid jobids

I'm not sure either of those points makes as much sense when job-manager is doing the purging, but I thought I'd raise those points in case there should be some discussion, or if other aspects of that PR come to mind that would be applicable here.

src/modules/job-manager/purge.c
```diff
@@ -0,0 +1,408 @@
...
+    if (!(f = flux_kvs_commit (purge->ctx->h, NULL, 0, txn))
+        || flux_future_aux_set (f, "count", int2ptr (count), NULL) < 0
+        || purge_publish (purge, jobs) < 0)
+        goto error;
+    flux_kvs_txn_destroy (txn);
+    /* N.B. if kvs commit fails, jobs are still removed from the hash/list.
+     * It doesn't seem worth the effort to structure the code to avoid this
+     * due to high complexity of solution, low probability of error, and
+     * minor consequences.
+     */
```
grondo, line 147: I didn't have a chance to fully review this PR yet, but `FLUX_TEST_VALGRIND=t ./t2809-job-purge.t` caught a leak of `jobs` on successful return here.

```
==89443== 1,168 (320 direct, 848 indirect) bytes in 8 blocks are definitely lost in loss record 21 of 29
==89443==    at 0x4848899: malloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==89443==    by 0x4A25070: json_array (in /usr/lib/x86_64-linux-gnu/libjansson.so.4.13.0)
==89443==    by 0xA39A93C: purge_inactive_jobs (purge.c:110)
==89443==    by 0xA39AC50: purge_request_cb (purge.c:283)
==89443==    by 0x487B8B7: call_handler (msg_handler.c:344)
==89443==    by 0x487BC49: dispatch_message (msg_handler.c:380)
==89443==    by 0x487BC49: handle_cb (msg_handler.c:481)
==89443==    by 0x48AACE2: ev_invoke_pending (ev.c:3770)
==89443==    by 0x48AE02F: ev_run (ev.c:4190)
==89443==    by 0x48AE02F: ev_run (ev.c:4021)
==89443==    by 0x487A9FE: flux_reactor_run (reactor.c:128)
==89443==    by 0xA392C11: mod_main (job-manager.c:220)
==89443==    by 0x119B20: module_thread (module.c:183)
==89443==    by 0x4AC4B42: start_thread (pthread_create.c:442)
```

garlick: Ah, most likely that was in the code I just added for updating job-list.  Thank you for catching that!

garlick: Fixed (squashed into 39664788f18df39be3c279d2c49c32d147a40126)

src/modules/job-manager/purge.c
```diff
@@ -0,0 +1,381 @@
...
+
+    if (flux_request_unpack (msg,
+                             NULL,
+                             "{s:f s:i s:i}",
+                             "age_limit", &age_limit,
+                             "num_limit", &num_limit,
+                             "batch", &batch) < 0)
+        goto error;
+    if (age_limit != INACTIVE_AGE_UNLIMITED && age_limit < 0) {
+        errmsg = "age limit must be >= 0 (0=unlimited)";
```
grondo, line 293: Is `errmsg` correct? (`INACTIVE_AGE_UNLIMITED` is `-1.` not `0`, but perhaps there's some subtlety I missed) FWIW, `0` would be a find value for unlimited

src/modules/job-manager/purge.c
```diff
@@ -0,0 +1,381 @@
...
+        errno = EINVAL;
+        goto error;
+    }
+    if (num_limit != INACTIVE_NUM_UNLIMITED && num_limit < 0) {
+        errmsg = "num limit must be >= 0 (0=unlimited)";
```
grondo, line 298: Same question as above, isn't `-1` unlimited?

src/modules/job-manager/purge.c
```diff
@@ -0,0 +1,381 @@
...
+        double t;
+        if (fsd_parse_duration (fsd, &t) < 0)
+            return errprintf (error,
+                              "job-manager.inactive-age-limit: invalid FSD");
+        age_limit = t;
+    }
+    if (num_limit != INACTIVE_NUM_UNLIMITED) {
+        if (num_limit < 0)
+            return errprintf (error,
+                              "job-manager.inactive-num-limit: must be >= 0");
```
grondo, line 363: Should the error message be `job-manager.inactive-num-limit: must be >= -1`?

grondo (approved):
This LGTM! I ran through some random manual testing and didn't find any issues. Even tried running a set of `flux job purge` commands in parallel with a small batch count and everything worked great!

Since `flux job purge` is an unrecoverable operation, I wonder if we want to require a `--force` option, but otherwise report how many jobs would be purged. I'm not sure how difficult that would be to add. Another option would be an `are you sure? (yes/no)` type prompt, unless `--force` is given. I don't think this has to be done before merging this PR, it was just something that occurred to me.

src/modules/job-manager/purge.c
garlick: oops yeah, I changed the values but didn't update the message!

garlick:
I like the ideal of reporting number of jobs purged only, unless `--force` is given!

src/modules/job-manager/purge.c
garlick: Actually, for the TOML config my thought was that users wouldn't specify a limit if they don't want one.  Does that sound OK?

O/w we should probably define a special value for unlimited age too, like "unlimited".

grondo: That makes sense to me.

garlick:
Just force pushed with the fixes discussed, adding `flux job purge --force`,  and squashing some incremental development (hope that's OK).  I left the publication of the purged job ids and subscription in `job-list` broken out as separate commits since that might be a temporary solution, and it should be made clear what to back out later if we come up with something better.

grondo:
Nice! I tested this out and works as described! Feel free to set MWP when you're ready.

grondo:
BTW, this feature satisfies one of my minor use cases: When running the Flux testsuite as a set of jobs within an instance, it is convenient to use `flux jobs --stats-only` to get a report if any jobs failed. The only way to do this multiple times, though, is to restart the instance to get the stats and previously completed jobs cleared.

Now `flux job purge` may be used as an alternative (e.g. when you do not want to launch another large instance) :+1:

Issue #4285, opened 2022-04-13
flux mini run fails on Spack built flux-sched


wkharold:
Built flux-sched on CentOS 7 with Spack 0.17.1

running ```flux mini run -n4 --label-io hostname``` failed with

    2022-04-13T16:46:43.100676Z job-ingest.err[0]: validator[0]: exited with rc=1
    flux-mini: ERROR: [Errno 71] Unrecoverable error: validator unexpectedly exited

It failed after a Slurm `salloc` and after `flux start --test-size=4`

grondo:
Ok, I rebuilt flux-core on the 0.17.1 branch and was not able to reproduce the problem. 

Here's the output of `spack spec flux-core` on my system.

```
$ spack spec flux-core
Input spec
--------------------------------
flux-core

Concretized
--------------------------------
flux-core@0.30.0%gcc@11.2.0~cuda~docs patches=0329f4b38b5ce4e1a3dee212cb93ce40388d1bde47c82a7a8e73bc7c38274c81 arch=linux-rhel7-broadwell
    ^czmq@4.1.1%gcc@11.2.0 arch=linux-rhel7-broadwell
        ^autoconf@2.69%gcc@11.2.0 patches=35c449281546376449766f92d49fc121ca50e330e60fefcfc9be2af3253082c2,7793209b33013dc0f81208718c68440c5aae80e7a1c4b8d336e382525af791a7,a49dd5bac3b62daa0ff688ab4d508d71dbd2f4f8d7e2a02321926346161bf3ee arch=linux-rhel7-broadwell
            ^m4@1.4.19%gcc@11.2.0+sigsegv patches=9dc5fbd0d5cb1037ab1e6d0ecc74a30df218d0a94bdd5a02759a97f62daca573,bfdffa7c2eb01021d5849b36972c069693654ad826c1a20b53534009a4ec7a89 arch=linux-rhel7-broadwell
                ^libsigsegv@2.13%gcc@11.2.0 arch=linux-rhel7-broadwell
            ^perl@5.34.0%gcc@11.2.0+cpanm+shared+threads arch=linux-rhel7-broadwell
                ^berkeley-db@18.1.40%gcc@11.2.0+cxx~docs+stl patches=b231fcc4d5cff05e5c3a4814f6a5af0e9a966428dc2176540d2c05aff41de522 arch=linux-rhel7-broadwell
[...]

PR #4288, opened 2022-04-16
python: fix return from flux.importer.import_plugins() when no plugins found

grondo:
Problem: When flux.importer.import_plugins() does not find any
plugins to import, it returns an empty list, but the function should
always return a dict.

Return an empty dict when no plugins are found in import_plugins().

As discovered by @wkharold in #4285


garlick (approved):
Good catch!

grondo:
Felt bad that there was no test here, so I added a simple one to the job-validator sharness tests.

garlick (commented):
Oh, even nicer.

PR #4289, opened 2022-04-18
flux-dmesg: colorize output and add -H, --human option

grondo:
This is a little side branch I've been working on, so I thought I'd go ahead and get it to the point of usefulness today and post it as a PR for comments.

There are times I've wanted to check `flux dmesg` output and compare the time between events, but the default output isn't entirely conducive to that. The timestamps are in ISO 8601 UTC with sub-second precision, and while these can be compared it requires a little bit of mental effort.

In contrast, the Linux `dmesg` command has a nice `-H, --human` mode which displays a human-readable timestamp for every new minute (e.g. `Apr18 12:30`), then a delta for each new message. I've borrowed that approach here (with a slight tweak by default) for `flux-dmesg(1)`.

`flux dmesg -H` displays a date-time timestamp at every new minute, then each subsequent message is prefixed with an offset from the last "time break". Additionally, the logs are colored by default in the style of Linux `dmesg(1)` (timebreak in bold green, timestamps in green, facility in yellow) and additionally the log severity is dropped in favor of message colorization. E.g.:

Screenshot from 2022-04-18 13-47-40

Color can be controlled with the semi-standard `--color=WHEN` option (a useful invocation may be `flux dmesg --color --human | less -R` for example). Also, a `-d, --delta` option is provided to present sub-minute timestamps as deltas between messages instead of offset from the last timebreak.

```console
$ flux dmesg --human --delta
[Apr18 07:18] broker[0]: insmod connector-local
[  +0.000095] broker[0]: start: none->join 12.7597ms
[  +0.000030] broker[0]: parent-none: join->init 0.023014ms
[  +0.001156] broker[0]: online:  (ranks )
[  +0.000026] connector-local[0]: allow-guest-user=false
[  +0.000029] connector-local[0]: allow-root-owner=false
[  +0.072795] broker[0]: insmod barrier
```

Finally, since all the colorization support was already there, I tacked on a commit that adds color to the `flux dmesg` default output as well. This will be easy to drop if there are objections - I didn't even add any tests for it.

Screenshot from 2022-04-18 15-12-29

### DIFF
--- a/doc/man1/flux-dmesg.rst
+++ b/doc/man1/flux-dmesg.rst
@@ -34,6 +34,17 @@ OPTIONS
 **-n, --new**
    Follow only new log entries.
 
+**-H, --human**
+   Display human-readable output. See also **--color** and **--delta**.
+
+**-d, --delta**
+   With **--human**, display the time delta between messages instead
+   of a relative offset since the last absolute timestamp.
+
+**-L, --color**\ *[=WHEN]*
+   Colorize output. The optional argument *WHEN* can be *auto*, *never*,
+   or *always*. If *WHEN* is omitted, it defaults to *always*. The default
+   value when the **--color** option is not used is *auto*.
 
 EXAMPLES
 ========
--- a/src/cmd/builtin/dmesg.c
+++ b/src/cmd/builtin/dmesg.c
@@ -328,6 +328,7 @@ static int cmd_dmesg (optparse_t *p, int ac, char *av[])
 
     tzset ();
 
+    log_init ("flux-dmesg");
     if ((n = optparse_option_index (p)) != ac)
         log_msg_exit ("flux-dmesg accepts no free arguments");
 
--- a/src/cmd/builtin/dmesg.c
+++ b/src/cmd/builtin/dmesg.c
@@ -270,7 +270,7 @@ void dmesg_print_human (struct dmesg_ctx *ctx, const char *buf, int len)
     fflush (stdout);
 }
 
-void dmesg_print (const char *buf, int len)
+void dmesg_print (struct dmesg_ctx *ctx, const char *buf, int len)
 {
     struct stdlog_header hdr;
     const char *msg;
@@ -282,12 +282,18 @@ void dmesg_print (const char *buf, int len)
     else {
         nodeid = strtoul (hdr.hostname, NULL, 10);
         severity = STDLOG_SEVERITY (hdr.pri);
-        printf ("%s %s.%s[%" PRIu32 "]: %.*s\n",
-                 hdr.timestamp,
-                 hdr.appname,
-                 stdlog_severity_to_string (severity),
-                 nodeid,
-                 msglen, msg);
+        printf ("%s%s%s %s%s.%s[%" PRIu32 "]%s: %s%.*s%s\n",
+                dmesg_color (ctx, DMESG_COLOR_TIME),
+                hdr.timestamp,
+                dmesg_color_reset (ctx),
+                dmesg_color (ctx, DMESG_COLOR_NAME),
+                hdr.appname,
+                stdlog_severity_to_string (severity),
+                nodeid,
+                dmesg_color_reset (ctx),
+                severity_color (ctx, severity),
+                msglen, msg,
+                dmesg_color_reset (ctx));
     }
     fflush (stdout);
 }
@@ -352,7 +358,7 @@ static int cmd_dmesg (optparse_t *p, int ac, char *av[])
             if (optparse_hasopt (p, "human"))
                 dmesg_print_human (&ctx, buf, strlen (buf));
             else
-                dmesg_print (buf, strlen (buf));
+                dmesg_print (&ctx, buf, strlen (buf));
             flux_future_reset (f);
         }
         if (errno != ENODATA)
--- a/src/cmd/builtin/dmesg.c
+++ b/src/cmd/builtin/dmesg.c
@@ -13,8 +13,65 @@
 #endif
 #include "builtin.h"
 #include <inttypes.h>
+#include <unistd.h>
+#include <time.h>
 
 #include "src/common/libutil/stdlog.h"
+#include "ccan/str/str.h"
+
+#define ANSI_COLOR_RED        "\x1b[31m"
+#define ANSI_COLOR_GREEN      "\x1b[32m"
+#define ANSI_COLOR_YELLOW     "\x1b[33m"
+#define ANSI_COLOR_BLUE       "\x1b[34m"
+#define ANSI_COLOR_MAGENTA    "\x1b[35m"
+#define ANSI_COLOR_CYAN       "\x1b[36m"
+#define ANSI_COLOR_GRAY       "\x1b[37m"
+
+#define ANSI_COLOR_RESET      "\x1b[0m"
+#define ANSI_COLOR_BOLD       "\x1b[1m"
+#define ANSI_COLOR_HALFBRIGHT "\x1b[2m"
+#define ANSI_COLOR_REVERSE    "\x1b[7m"
+
+#define ANSI_COLOR_RESET      "\x1b[0m"
+
+struct dmesg_ctx {
+    optparse_t *p;
+    unsigned int color:1;
+    unsigned int delta:1;
+    struct tm last_tm;
+    struct timeval last_tv;
+};
+
+enum {
+    DMESG_COLOR_MODE_AUTO,
+    DMESG_COLOR_MODE_NEVER,
+    DMESG_COLOR_MODE_ALWAYS
+};
+
+enum {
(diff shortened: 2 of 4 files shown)


garlick:
Love it :+1: !

src/cmd/builtin/dmesg.c
```diff
@@ -25,9 +82,184 @@ static struct optparse_option dmesg_opts[] = {
...
+        return -1;
+
+    if (!(localtime_r (&t, tm)))
+        return -1;
+    tv->tv_sec = t;
+
+    if (extra[0] == '.') {
+        char *endptr;
+        double d = strtod (extra, &endptr);
+        tv->tv_usec = d * 1000000;
```
chu11, line 153: should check `endptr`, otherwise no need to pass in?

t/t0009-dmesg.t
```diff
@@ -111,5 +111,35 @@ test_expect_success 'logged non-ascii characters handled ok' '
 test_expect_success 'dmesg request with empty payload fails with EPROTO(71)' '
 	${RPC} log.dmesg 71 </dev/null
 '
+test_expect_success 'dmesg -H, --human works' '
+	flux dmesg --human | sed -n 1p \
+	    | grep "^\[[A-Z][a-z][a-z][0-3][0-9] [0-9][0-9]:[0-9][0-9]\]" &&
+	flux dmesg --human | sed -n 2p \
+	    | grep "^\[ *+[0-9]*\.[0-9]*\]"
+
+'
```
chu11, line 120: empty line added by mistake?

t/t0009-dmesg.t
```diff
@@ -111,5 +111,35 @@ test_expect_success 'logged non-ascii characters handled ok' '
...
+test_expect_success 'dmesg -H, --human --delta works' '
+	flux dmesg --human --delta | sed -n 1p \
+	    | grep "^\[[A-Z][a-z][a-z][0-3][0-9] [0-9][0-9]:[0-9][0-9]\]" &&
+	flux dmesg --human --delta | sed -n 2p \
+	    | grep "^\[ *+[0-9]*\.[0-9]*\]"
+
```
chu11, line 126: empty line added by mistake?

t/t0009-dmesg.t
chu11, line 120: perhaps add a small comment indicating what the regex is for.  Took me a few seconds to grok intent :-)

doc/man1/flux-dmesg.rst
```diff
@@ -34,6 +34,17 @@ OPTIONS
...
+   Display human-readable output. See also **--color** and **--delta**.
+
+**-d, --delta**
+   With **--human**, display the time delta between messages instead
+   of a relative offset since the last absolute timestamp.
+
+**-L, --color**\ *[=WHEN]*
+   Colorize output. The optional argument *WHEN* can be *auto*, *never*,
+   or *always*. If *WHEN* is omitted, it defaults to *always*. The default
+   value if **--color** is not used is *auto*.
```
chu11, line 47: value if -> value of?

grondo: No, "The value of --color is not used is _auto_" won't make sense either. Obviously the wording is confusing though, so how about simply "The default value is `auto`"?

grondo:
Thanks for the review @chu11! I've addressed comments I believe, squashed, and force pushed the result.

t/t0009-dmesg.t
```diff
@@ -111,5 +111,39 @@ test_expect_success 'logged non-ascii characters handled ok' '
 test_expect_success 'dmesg request with empty payload fails with EPROTO(71)' '
 	${RPC} log.dmesg 71 </dev/null
 '
+test_expect_success 'dmesg -H, --human works' '
+	#
+	#  Note: --human option should format first timestamp of dmesg output
+	#   as [MmmDD HH:MM] and second line should be an offset thereof
+        #   e.g. [  +0.NNNNNN]. The following regexes attempt to verify
```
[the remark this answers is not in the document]
grondo: Yeah, weird, I don't know why that one line got spaces instead of a tab, but I'll fix.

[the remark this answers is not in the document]
grondo: Ok, fixed. Thanks for catching that!

src/cmd/builtin/dmesg.c
```diff
@@ -25,9 +82,190 @@ static struct optparse_option dmesg_opts[] = {
...
+        case LOG_CRIT:
+            return dmesg_color (ctx, DMESG_COLOR_CRIT);
+        case LOG_ERR:
+            return dmesg_color (ctx, DMESG_COLOR_ERR);
+        case LOG_WARNING:
+            return dmesg_color (ctx, DMESG_COLOR_WARNING);
+        case LOG_DEBUG:
+            return dmesg_color (ctx, DMESG_COLOR_DEBUG);
+        default:
+            return "";
```
garlick, line 239: Were LOG_NOTICE and LOG_INFO left off on purpose?

On my system DEBUG, INFO, and NOTICE all look the same.

garlick (approved):
This is great!   Just some minor quibbles inline.

src/cmd/builtin/dmesg.c
grondo: Yes, LOG_NOTICE and LOG_INFO are left in the default color. I could add them to be more explicit if that is preferrable. On my system LOG_DEBUG is gray compared to white for the default text. (e.g. see above screenshots)

So if you run this:
```
$ printf "hello\n\x1b[37mhello\x1b[0m\n"
```

Both lines look the same? I guess this is why util-linux supports [`terminal-colors.d(5)`](https://manpages.ubuntu.com/manpages/xenial/man5/terminal-colors.d.5.html). I was thinking one day we could support that, but it seemed like overkill for now.

Maybe for now we could try this?

```diff
diff --git a/src/cmd/builtin/dmesg.c b/src/cmd/builtin/dmesg.c
index f349a18a3..c9db85950 100644
--- a/src/cmd/builtin/dmesg.c
+++ b/src/cmd/builtin/dmesg.c
@@ -69,7 +69,7 @@ static const char *dmesg_colors[] = {
     [DMESG_COLOR_CRIT]          = ANSI_COLOR_BOLD ANSI_COLOR_RED,
     [DMESG_COLOR_ERR]           = ANSI_COLOR_RED,
     [DMESG_COLOR_WARNING]       = ANSI_COLOR_BOLD,
-    [DMESG_COLOR_DEBUG]         = ANSI_COLOR_GRAY,
+    [DMESG_COLOR_DEBUG]         = ANSI_COLOR_HALFBRIGHT ANSI_COLOR_BLUE,
 };
```

garlick: Yeah pretty close to the same.  One's a tiny bit brighter.

Screenshot from 2022-04-19 09-56-22

With your changes - better!

Screenshot from 2022-04-19 09-58-56

grondo: Uh, halfbright blue was unreadable in iTerm on my Mac, so that probably isn't great.

grondo: I'll push a change with just normal blue and see how that goes. Looks fine on the two terminals I tried. Not sure if it ends up being too busy though.

src/cmd/builtin/dmesg.c
```diff
@@ -25,9 +82,190 @@ static struct optparse_option dmesg_opts[] = {
...
+
+    if (extra[0] == '.') {
+        char *endptr;
+        double d = strtod (extra, &endptr);
+
+        /*  In this implementation, the rest of the timestamp must be
+         *   "Z" to indicate UTC or "Zulu time":
+         */
+        if (endptr == NULL || !streq (endptr, "Z"))
+            return -1;
```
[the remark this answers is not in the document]
grondo: Ok, fixed and force-pushed. Also fixed the use of `endptr` (`endptr == NULL` wasn't the right test here, now we just ensure `*endptr == 'Z'`).

src/cmd/builtin/dmesg.c
grondo: Ok, I pushed the result. Maybe try this

```sh
for severity in emerg alert crit err warning notice info debug;
    do flux logger --severity=$severity This is a message at level $severity;
done
```
And ensure it looks ok on your terminal with `flux dmesg -H`

garlick: That looks good!

Screenshot from 2022-04-19 10-15-36

grondo: Ok, I squashed the color change and force pushed. Hope that was ok!

grondo:
@mergifyio refresh

grondo:
I'm not sure why mergify is stuck on this one. Maybe I'll try another push to see if that unsticks it.

Issue #4290, opened 2022-04-19
sched-simple allocates a node that is not online


garlick:
Problem: 8 node cluster has  rank 0 excluded, and rank 4 offline (it does not exist but it is configured), however, rank 4 seems to have been allocated to a job by sched-simple even though it never appears in a resource.acquire "up" response.

This should have hung waiting for another node to come online:
```
$ flux mini run -N7 hostname
0.495s: job.exception type=prolog severity=0 prolog failed with status=17408
```

```
$ sudo flux dmesg
2022-04-19T01:42:34.473718Z broker.debug[0]: insmod connector-local
2022-04-19T01:42:34.473932Z broker.info[0]: start: none->join 52.3214ms
2022-04-19T01:42:34.474068Z broker.info[0]: parent-none: join->init 0.088333ms
2022-04-19T01:42:34.477071Z broker.info[0]: online:  (ranks )
2022-04-19T01:42:34.488180Z connector-local.debug[0]: allow-guest-user=true
2022-04-19T01:42:34.488217Z connector-local.debug[0]: allow-root-owner=true
2022-04-19T01:42:34.744635Z broker.debug[0]: insmod barrier
2022-04-19T01:42:34.880016Z broker.debug[0]: insmod content-sqlite
2022-04-19T01:42:34.926633Z broker.debug[0]: content backing store: enabled content-sqlite
2022-04-19T01:42:35.019088Z broker.debug[0]: insmod kvs
2022-04-19T01:42:35.021593Z kvs.info[0]: restored KVS from checkpoint on 2022-04-19T01:29:02Z
2022-04-19T01:42:35.097672Z broker.debug[0]: insmod kvs-watch
2022-04-19T01:42:35.203546Z broker.debug[0]: insmod resource
2022-04-19T01:42:35.307399Z resource.debug[0]: reslog_cb: resource-init event posted
2022-04-19T01:42:35.310529Z resource.debug[0]: reslog_cb: resource-define event posted
2022-04-19T01:42:35.360924Z broker.debug[0]: insmod cron
2022-04-19T01:42:35.361763Z cron.info[0]: synchronizing cron tasks to event heartbeat.pulse
2022-04-19T01:42:35.429630Z broker.debug[0]: insmod job-manager
2022-04-19T01:42:35.915093Z job-manager.info[0]: restart: 103 jobs
2022-04-19T01:42:35.915129Z job-manager.info[0]: restart: 0 running jobs
[...]

PR #4292, opened 2022-04-19
sched-simple: do not allocate down nodes with --exclusive

grondo:
As noted in #4290, librlist and thus sched-simple ignores up/down state of nodes when the `--exclusive` flag is used (implicitly or explicitly). This is because the `exclusive` branch in `rlist_alloc_nnodes()` simply ignored the state of nodes while trying to allocate them exclusively.

This patch fixes the algorithm to sort down nodes to the end of the list, then abort the search on encountering the first used _or_ down node.

A unit test is added to ensure basic operation of the fix.


garlick (approved):
Thanks for the quick fix!

PR #4293, opened 2022-04-20
shell: fix reconnect hang

chu11:
Per discussion in #4070, this fixes a hang in the shell due to reconnects occurring at the wrong time and shell "completion references" never getting cleared.

Note that I created a new callback for when we reconnect, called `reconnect_init()`.  This is presently unused.  I only added it because adding only a `reconnect_cleanup()` callback seemed "unbalanced".  Can nix if people think it should be nixed.  There are things that could be done with a `reconnect_init()` though.  Perhaps we want to re-try writing to an eventlog by caching unconfirmed eventlog writes?  That's just an idea and not for this PR.

This is built on top of #4070.  Note that this is independent of #4255, does not have to go in before / after it.

Note I don't have an issue created for this, b/c it seems weird to create an issue for a bug that can't exist yet b/c of in progress PR.  Will create an issue later.

### DIFF
--- a/src/common/libeventlog/eventlog.c
+++ b/src/common/libeventlog/eventlog.c
@@ -341,7 +341,7 @@ char *eventlog_entry_encode (json_t *entry)
     return buf;
 }
 
-char *eventlog_encode (json_t *array)
+char *eventlog_encode (json_t *a)
 {
     json_t *value;
     size_t index;
@@ -349,11 +349,11 @@ char *eventlog_encode (json_t *array)
     int bufsz = 0;
     int used = 0;
 
-    if (!array || !json_is_array (array)) {
+    if (!a || !json_is_array (a)) {
         errno = EINVAL;
         return NULL;
     }
-    json_array_foreach (array, index, value) {
+    json_array_foreach (a, index, value) {
         char *s = json_dumps (value, JSON_COMPACT);
 
         if (!s || (used = append_string_nl (&buf, &bufsz, used, s)) < 0) {
--- a/src/common/libeventlog/eventlog.c
+++ b/src/common/libeventlog/eventlog.c
@@ -369,6 +369,39 @@ char *eventlog_encode (json_t *a)
     return buf;
 }
 
+int eventlog_contains_event (const char *s, const char *name)
+{
+    json_t *a = NULL;
+    size_t index;
+    json_t *value;
+    int rv = -1;
+
+    if (!s || !name) {
+        errno = EINVAL;
+        return -1;
+    }
+
+    if (!(a = eventlog_decode (s)))
+        return -1;
+
+    json_array_foreach (a, index, value) {
+        double t;
+        const char *n;
+        json_t *c;
+        if (eventlog_entry_parse (value, &t, &n, &c) < 0)
+            goto out;
+        if (!strcmp (name, n)) {
+            rv = 1;
+            goto out;
+        }
+    }
+
+    rv = 0;
+out:
+    json_decref (a);
+    return rv;
+}
+
 /*
  * vi:tabstop=4 shiftwidth=4 expandtab
  */
--- a/src/common/libeventlog/eventlog.h
+++ b/src/common/libeventlog/eventlog.h
@@ -52,6 +52,11 @@ json_t *eventlog_entry_vpack (double timestamp,
 
 char *eventlog_entry_encode (json_t *entry);
 
+/* Convenience function to search eventlog for event with name.
+ * Returns 1 if found, 0 if not, -1 on error.
+ */
+int eventlog_contains_event (const char *s, const char *name);
+
 #ifdef __cplusplus
 }
 #endif
--- a/src/common/libeventlog/eventlog.h
+++ b/src/common/libeventlog/eventlog.h
@@ -28,7 +28,7 @@ int eventlog_entry_parse (json_t *entry,
 json_t *eventlog_decode (const char *s);
 
 /* encode json array of event objects into an eventlog */
-char *eventlog_encode (json_t *o);
+char *eventlog_encode (json_t *a);
 
 /* decode a single eventlog entry into a json object */
 json_t *eventlog_entry_decode (const char *entry);
--- a/src/common/libeventlog/test/eventlog.c
+++ b/src/common/libeventlog/test/eventlog.c
@@ -479,6 +479,34 @@ void eventlog_entry_encoding_errors (void)
         "eventlog_entry_vpack context=\"[\"foo\"]\" fails with EINVAL");
 }
 
+void eventlog_contains_event_test (void)
+{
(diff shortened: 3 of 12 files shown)


grondo:
This might require more careful thought. A transaction request could be dropped in the shell due to the reconnect, or even in the local broker or any broker along the way to the KVS if they've been restarted. In that case, it might be good to store in-flight transactions and attempt them again after the receonnect. We'd then have to ensure transactions were idempotent, or check after the fact if the eventlog entries already made it to the respective eventlog(s).

Though, now that I mention that out loud, perhaps this half measure is a good first step. Then we can deal with fallout like missing eventlog entries later. I don't see why this can't go as part of #4070 though, since it is directly related to the job shell reconnect support, which is already included in that PR. It seems preferable to have PR in as much a working state as possible before merging.

chu11:
> This might require more careful thought. A transaction request could be dropped in the shell due to the reconnect, or even in the local broker or any broker along the way to the KVS if they've been restarted. In that case, it might be good to store in-flight transactions and attempt them again after the receonnect. We'd then have to ensure transactions were idempotent, or check after the fact if the eventlog entries already made it to the respective eventlog(s).

I was actually pondering the idempotent thing last night.  Do we require idempotent eventlogs?  I'm sure we have parts of flux that assume it, dunno if that could be worked around?  like would most of flux care if we had two `finish` events?  idempotent user output could be configurable or something since that could be large and take up a lot of memory.

> Though, now that I mention that out loud, perhaps this half measure is a good first step. Then we can deal with fallout like missing eventlog entries later. 

Ya know, let me just experimentally try the "cache unconfirmed eventlog entries" thing I mentioned above.  See if it'll work out.  It might not be horrible.  I won't bother with a retry of user output, that's a different story.

> I don't see why this can't go as part of https://github.com/flux-framework/flux-core/pull/4070 though, since it is directly related to the job shell reconnect support, which is already included in that PR. It seems preferable to have PR in as much a working state as possible before merging.

Ok, I'll put into that PR instead.

grondo:
>  like would most of flux care if we had two finish events? 

Good question! There might be some code, e.g. in `flux job attach`,  which perform some action that is not idempotent on seeing a specific event. (e.g. attaching to a pty on the `shell.init` event comes to mind). It does seem easier to fix those call sites than to somehow make eventlog appends de-duplicate or somesuch...

grondo:
> I won't bother with a retry of user output, that's a different story.

Yeah, this is complicated by the fact that the eventlogger collects things in batches. It is probably ok if lines of output might be dropped on a restart. Note that for the system instance, if all jobs are "batch jobs" then almost certainly output is redirected to a file anyway, so probably not much effort should be expended here.

chu11:
closing will put into #4070 instead

chu11:
re-opening, the number of commits started getting larger and its becoming more complex.  Decided it would be best to split off into its own PR.

chu11:
re-pushed, this is a WIP with atleast 1 ugly implementation detail to figure out.

- call `shell_remove_completion_ref()` upon a reconnect for the shell eventlogger in `events.c` (missed this one before).

- setup a `reconnect.init` for the eventlogger in `events.c`.  It will lookup the `exec.eventlog` and ensure any events emitted by the `shell` have not been lost due to a reconnect.  Not presently tested, as it would probably require a lot of instrumentation to catch this specific race.  I struggled on how to implement this.  Would love the lookup of a hash, but need the in-order of a list.  Eventually settled on a list, but using `zlistx` requires a lot of "setup" for such a simple piece of code.

- the ugly part, the eventlogger in `events.c` is not a plugin.  So the `reconnect_cleanup()` and `reconnect_init()` calls can't be called through the plugstack.  For the mean time, I just manually call them in `reconnect ()`.  I'm not a huge fan of this, we could ...
  - make the eventlogger to `exec.eventlog` a plugin?  Probably not, there's probably a lot of logging that is expected to work / not work not connected to the plugstack
  - just manually create a plugin, put it on the plugstack, and use the plagstack to register the `reconnect.init` and `reconnect.cleanup` callbacks? - not pretty, but maybe ok?
  - add some mechanism to the plugstack to say "call these functions too", even though they are not specifically part of the plugstack.
  - leave what i have currently, it's not the prettiest, but perhaps ok ....

grondo:
I was surprising to me that `reconnect.cleanup` was before a reconnect is started, and `reconnect.init` is after. It might be clearer if we named these `shell.disconnect` and `shell.reconnect`. Also, it occurs to me that the "cleanup" could be done in `shell.reconnect` (after the shell has reconnected), so do we only need one new plugin callback? (this would nicely mirror the existing `shell.connect` callback.

What you've done for `events.c` looks fine to me, no reason that everything has to be a plugin IMO. However, if it bothers you it would be simple to create a `flux_plugin_t` handle inside of `events.c` and register it with `plugstack_push()`. (I guess this is one of your suggested solutions already. We can make it "prettier" by making a nicer interface to do it.)

grondo:
> Not presently tested, as it would probably require a lot of instrumentation to catch this specific race. 

Hm, I was wondering if a specially crafted test shell plugin could help test that at least the code is working.

Another idea would be to abstract the eventlog comparison code into its own module which could at least be unit tested. Maybe it could even be in libeventlog.

chu11:
> I was surprising to me that reconnect.cleanup was before a reconnect is started, and reconnect.init is after. It might be clearer if we named these shell.disconnect and shell.reconnect. Also, it occurs to me that the "cleanup" could be done in shell.reconnect (after the shell has reconnected), so do we only need one new plugin callback? (this would nicely mirror the existing shell.connect callback.

Ahh, I like the idea of calling it `shell.reconnect` and having only the one function.  It would make my hack a little less offensive in my eyes too :P

chu11:
> Hm, I was wondering if a specially crafted test shell plugin could help test that at least the code is working.

Yeah, I just wasn't sure how I could trigger a `reconnect` in the shell for testing, as well as make sure an eventlog message was in flight too.

> Another idea would be to abstract the eventlog comparison code into its own module which could at least be unit tested. Maybe it could even be in libeventlog.

Ahh that's a good idea.

chu11:
re-pushed, per some above discussion

- "reconnect.cleanup" and "reconnect.init" are now just a single "reconnect" callback.

- to get some mild increased coverage, I added a new convenience eventlog function `eventlog_contains_event (const char *s, const char *name);`, which takes an eventlog string and just checks if an event name is in it.
  - If you have multiple event names to check, this can be slower because you gotta convert the string to array multiple times.  But I figure its a small cost in exchange for just a slightly simpler interface and given that the number of shell events emitted to the exec.eventlog is currently 2 events, figure its not a big deal.

chu11:
rebased on master, since it'd been awhile

chu11:
@grondo thanks! re-pushed with fixes.  Setting MWP

Issue #4152, opened 2022-02-22
content-sqlite may leave on-disk data inconsistent on a crash


garlick:
Problem: The `content-sqlite` content backing store module sets [SQLite pragmas](https://www.sqlite.org/pragma.html) that improve performance, but relax consistency guarantees of SQL commits, like `journal_mode=OFF` and `synchronous=OFF`.  Consequently, the backing store could become corrupted after a rank 0 crash, complicating recovery.

We should consider an option (especially for system instance) that disables these performance hacks.

After quantifying the before and after performance, we may want to investigate some mitigations.  For example, right now each blob is committed separately.  We could add multiple blobs to a commit, then after a threshold size, or a timer expiration, commit them and send the responses for all once the commit finishes.

garlick:
An example:
```
Apr 03 13:07:06 picl0 flux[14320]: content-sqlite.err[0]: setting sqlite 'journal_mode' pragma: database disk image is malformed(11)
```

garlick:
Another example.  This time the instance was shut down cleanly with `sudo flux shutdown`.  I'm not sure what caused the database to become malformed.
```
Apr 16 16:22:30 picl0 flux[86356]: broker.info[0]: shutdown: run->cleanup 38.429m
Apr 16 16:22:30 picl0 flux[86356]: broker.info[0]: cleanup.0: flux queue stop --quiet Exited (rc=0) 0.0s
Apr 16 16:22:30 picl0 flux[86356]: broker.info[0]: cleanup.1: flux job cancelall --user=all --quiet -f --states RUN Exited (rc=0) 0.0s
Apr 16 16:22:30 picl0 flux[86356]: broker.info[0]: cleanup.2: flux queue idle --quiet Exited (rc=0) 0.0s
Apr 16 16:22:30 picl0 flux[86356]: broker.info[0]: cleanup-success: cleanup->shutdown 0.12144s
Apr 16 16:22:30 picl0 flux[86356]: broker.info[0]: children-none: shutdown->finalize 0.437481ms
Apr 16 16:22:30 picl0 flux[86356]: broker.info[0]: rc3.0: /usr/local/etc/flux/rc3 Exited (rc=0) 0.3s
Apr 16 16:22:30 picl0 flux[86356]: broker.info[0]: rc3-success: finalize->goodbye 0.338493s
Apr 16 16:22:30 picl0 flux[86356]: broker.info[0]: goodbye: goodbye->exit 1.144ms
[...]

Issue #4267, opened 2022-04-04
support content.backing-module=none


garlick:
Problem: the default instance configuration loads `content-sqlite` but removes the database file when the instance terminates and locates the database file in `/tmp`, which on many systems is a ram disk.  If it's memory either way - are we just expending effort moving it around? 

Perhaps we could change rc1 to only load `content-sqlite` when `statedir` is defined, and advise workflow/DAT users to use `flux start -o,-Sstatedir=/a/persistent/directory`

garlick:
I'm not sure about making this is the default, since there is unexplained job throughput degradation in this mode.  However it is not currently possible to even select this mode.  Let's allow this issue to be closed once setting `content.backing-module=none` works as you'd expect.

Extra credit: allow KVS checkpoints to be saved in memory after the KVS module is unloaded so that `flux dump --checkpoint` works after the KVS module is unloaded, and `flux shutdown --dump=FILE` works as expected.

garlick:
Well that got me wondering how that would go.   Here's a quick patch that would use `:memory:` if `statedir` is not defined.  IOW, whenever the instance is not restartable anyway.   Compared to just making `content-cache` work standalone, an upside is that larger blobs are compressed.  A downside is the unnecessary overhead of moving blobs back and forth between content-cache and content-sqlite.  Here's a quick throughput test:

Using `:memory:` (rank 0 broker reached 1.2g)
```
ƒ(s=1,d=0,builddir) $ src/test/throughput.py  -x  -n 8192
number of jobs: 8192
submit time:    2.901 s (2824.2 job/s)
script runtime: 368.868s
job runtime:    368.634s
throughput:     22.2 job/s (script:  22.2 job/s)

```

Compared to master  (rss reached 745m)
```
ƒ(s=1,d=0,builddir) $ src/test/throughput.py  -x  -n 8192
number of jobs: 8192
submit time:    2.947 s (2779.9 job/s)
script runtime: 322.580s
job runtime:    322.333s
throughput:     25.4 job/s (script:  25.4 job/s)
```

```diff
diff --git a/src/modules/content-sqlite/content-sqlite.c b/src/modules/content-sqlite/content-sqlite.c
index 6d9ad61f4..5fb46b734 100644
--- a/src/modules/content-sqlite/content-sqlite.c
+++ b/src/modules/content-sqlite/content-sqlite.c
@@ -539,11 +539,18 @@ static json_t *pack_tstat (tstat_t *ts)
     return o;
[...]

PR #4294, opened 2022-04-20
select safer content.sqlite consistency pragmas

garlick:
Problem: content.sqlite sets journal_mode=OFF, synchronous=OFF for performance, but we do occasionally see database corruption in the system instance, as noted in #4152.

This PR changes the content-sqlite pragmas to journal_mode=WAL, synchronous=NORMAL which is safer at the expense of some performance.  It also adds metrics to content-sqlite so we can get a window into the performance.   Below is a comparison of performance in both modes while running the throughput test with 32768 jobs.  Job throughput decreased by 23%.  This is likely an acceptable trade-off for the system instance, but perhaps not for single user instances.

Marking as a WIP for now.

**TL;DR**

Snippets from https://www.sqlite.org/pragma.html 

**journal_mode=OFF, synchronous=OFF**

> If the application crashes in the middle of a transaction when the OFF journaling mode is set, then the database file will very likely [go corrupt](https://www.sqlite.org/howtocorrupt.html#cfgerr). Without a journal, there is no way for a statement to unwind partially completed operations following a constraint error. This might also leave the database in a corrupted state.

> With synchronous OFF (0), SQLite continues without syncing as soon as it has handed data off to the operating system. If the application running SQLite crashes, the data will be safe, but the database [might become corrupted](https://www.sqlite.org/howtocorrupt.html#cfgerr) if the operating system crashes or the computer loses power before that data has been written to the disk surface. On the other hand, commits can be orders of magnitude faster with synchronous OFF.

**journal_mode=WAL, synchronous=NORMAL**

> The WAL journaling mode uses a [write-ahead log](https://www.sqlite.org/wal.html) instead of a rollback journal to implement transactions. The WAL journaling mode is persistent; after being set it stays in effect across multiple database connections and after closing and reopening the database.

 > When synchronous is NORMAL (1), the SQLite database engine will still sync at the most critical moments, but less often than in FULL mode.  [WAL mode](https://www.sqlite.org/wal.html) is always consistent with synchronous=NORMAL, but WAL mode does lose durability. A transaction committed in WAL mode with synchronous=NORMAL might roll back following a power loss or system crash. Transactions are durable across application crashes regardless of the synchronous setting or journal mode. The synchronous=NORMAL setting is a good choice for most applications running in WAL mode.

**journal_mode=WAL, synchronous=NORMAL**
```
$ src/test/throughput.py -sn 32768
 32768 jobs:  32768 submitted,      0 running, 32768 completed
number of jobs: 32768
submit time:    8.383 s (3908.9 job/s)
script runtime: 472.976s
job runtime:    206.300s
throughput:     158.8 job/s (script:  69.3 job/s)
$ flux module stats content
{
 "count": 30855,
 "valid": 30855,
 "dirty": 0,
 "size": 16775199,
 "flush-batch-count": 0
}
$ flux module stats content-sqlite
{
 "object_count": 1029997,
 "dbfile_size": 997396480,
 "dbfile_free": 277364133888,
 "load_time": {
  "count": 463050,
  "min": 0.0014220000000000001,
  "max": 0.26876100000000003,
  "mean": 0.0065742753590326119,
  "stddev": 0.0037543449026159067
 },
 "store_time": {
  "count": 1029997,
  "min": 0.0057210000000000004,
  "max": 40.314458999999999,
  "mean": 0.07438751751801416,
  "stddev": 0.91053893559580146
 }
}
```

**journal_mode=OFF, synchronous=OFF**
```
$ src/test/throughput.py -sn 32768
 32768 jobs:  32768 submitted,      0 running, 32768 completed
number of jobs: 32768
submit time:    7.961 s (4116.1 job/s)
script runtime: 468.344s
job runtime:    160.295s
throughput:     204.4 job/s (script:  70.0 job/s)
$ flux module stats content
{
 "count": 24508,
 "valid": 24508,
 "dirty": 0,
 "size": 16766936,
 "flush-batch-count": 0
}
$ flux module stats content-sqlite
{
 "object_count": 1041628,
 "dbfile_size": 1014226944,
 "dbfile_free": 277350973440,
 "load_time": {
  "count": 464516,
  "min": 0.001433,
  "max": 0.40983799999999998,
  "mean": 0.0063714025975424998,
  "stddev": 0.0037928408896549066
 },
 "store_time": {
  "count": 1041628,
  "min": 0.0032560000000000002,
  "max": 8.1697399999999991,
  "mean": 0.016616053997204328,
  "stddev": 0.042195470281613467
 }
}
```

### DIFF
--- a/src/modules/content-sqlite/content-sqlite.c
+++ b/src/modules/content-sqlite/content-sqlite.c
@@ -608,6 +608,7 @@ static int content_sqlite_opendb (struct content_sqlite *ctx)
 {
     int flags = SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE;
     char s[128];
+    int count;
 
     if (sqlite3_open_v2 (ctx->dbfile, &ctx->db, flags, NULL) != SQLITE_OK) {
         log_sqlite_error (ctx, "opening %s", ctx->dbfile);
@@ -687,6 +688,21 @@ static int content_sqlite_opendb (struct content_sqlite *ctx)
         log_sqlite_error (ctx, "preparing checkpt_put stmt");
         goto error;
     }
+    if (sqlite3_exec (ctx->db,
+                      sql_objects_count,
+                      set_count,
+                      &count,
+                      NULL) != SQLITE_OK) {
+        log_sqlite_error (ctx, "querying objects count");
+        goto error;
+    }
+    flux_log (ctx->h,
+              LOG_DEBUG,
+              "%s (%d objects) journal_mode=%s synchronous=%s",
+              ctx->dbfile,
+              count,
+              ctx->journal_mode,
+              ctx->synchronous);
     return 0;
 error:
     set_errno_from_sqlite_error (ctx);
--- a/src/modules/content-sqlite/content-sqlite.c
+++ b/src/modules/content-sqlite/content-sqlite.c
@@ -72,6 +72,8 @@ struct content_sqlite {
     size_t lzo_bufsize;
     void *lzo_buf;
     struct content_stats stats;
+    const char *journal_mode;
+    const char *synchronous;
 };
 
 static void log_sqlite_error (struct content_sqlite *ctx, const char *fmt, ...)
@@ -605,21 +607,24 @@ error:
 static int content_sqlite_opendb (struct content_sqlite *ctx)
 {
     int flags = SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE;
+    char s[128];
 
     if (sqlite3_open_v2 (ctx->dbfile, &ctx->db, flags, NULL) != SQLITE_OK) {
         log_sqlite_error (ctx, "opening %s", ctx->dbfile);
         goto error;
     }
+    snprintf (s, sizeof (s), "PRAGMA journal_mode=%s", ctx->journal_mode);
     if (sqlite3_exec (ctx->db,
-                      "PRAGMA journal_mode=OFF",
+                      s,
                       NULL,
                       NULL,
                       NULL) != SQLITE_OK) {
         log_sqlite_error (ctx, "setting sqlite 'journal_mode' pragma");
         goto error;
     }
+    snprintf (s, sizeof (s), "PRAGMA synchronous=%s", ctx->synchronous);
     if (sqlite3_exec (ctx->db,
-                      "PRAGMA synchronous=OFF",
+                      s,
                       NULL,
                       NULL,
                       NULL) != SQLITE_OK) {
@@ -720,6 +725,8 @@ static struct content_sqlite *content_sqlite_create (flux_t *h)
         goto error;
     ctx->lzo_bufsize = lzo_buf_chunksize;
     ctx->h = h;
+    ctx->journal_mode = "WAL";
+    ctx->synchronous = "NORMAL";
 
     /* Some tunables:
      * - the hash function, e.g. sha1, sha256
@@ -732,10 +739,15 @@ static struct content_sqlite *content_sqlite_create (flux_t *h)
     }
 
     /* Prefer 'statedir' as the location for content.sqlite file, if set.
-     * Otherwise use 'rundir'.
+     * Otherwise use 'rundir', and enable pragmas that increase performance
+     * but risk database corruption on a crash (since rundir is temporary
+     * and the database is not being preserved after a crash anyway).
      */
-    if (!(dbdir = flux_attr_get (h, "statedir")))
+    if (!(dbdir = flux_attr_get (h, "statedir"))) {
         dbdir = flux_attr_get (h, "rundir");
+        ctx->journal_mode = "OFF";
+        ctx->synchronous = "OFF";
+    }
     if (!dbdir) {
         flux_log_error (h, "neither statedir nor rundir are set");
         goto error;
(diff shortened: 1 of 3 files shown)


garlick:
> This is likely an acceptable trade-off for the system instance, but perhaps not for single user instances.

Would it be too disruptive to NOT load a content backing store unless `statedir` is explicitly configured?  IOW, accumulate content in the rank 0 broker memory?  As noted in #4267, when /tmp is a `tmpfs`, the backing store in `rundir` may just be shifting content from one portion of memory to another.

I repeated the above 32K job throughput test without a backing store and noted that broker rss grew to 4.3G, compared to 2.2G with a backing store and the new pragmas.  Throughput is slightly improved over a backing store with old pragmas, more so with the new pragmas.  I observed (anecdotally) performance variation during the run, probably due to stepped efficiency of zhashx as it grew and optimized itself.

```
$ src/test/throughput.py -sn 32768
 32768 jobs:  32768 submitted,      0 running, 32768 completed
number of jobs: 32768
submit time:    7.774 s (4214.9 job/s)
script runtime: 490.101s
job runtime:    156.350s
throughput:     209.6 job/s (script:  66.9 job/s)
```

We are already telling workflow users in our [FAQ](https://flux-framework.readthedocs.io/en/latest/faqs.html#memory-exhaustion-on-a-node-when-running-large-ensembles-with-flux) (which needs updating) to offload content storage to a real file system.

grondo:
I'll just throw this idea out there: have we considered offering a possibly more performant, reliable content backing store for use with large system instances, e.g. mysql? Keep in mind my database knowledge is weak, so this may be a dumb question.

garlick:
> I'll just throw this idea out there: have we considered offering a possibly more performant, reliable content backing store for use with large system instances, e.g. mysql?

A definite possibility down the road.   However, I think that's a bigger project that we can undertake right now, and might be better tackled by someone who has prior experience with one or more of those.  A couple of thoughts though
- since sqlite is embeddable, it's available for workflow users to just turn on, while an external db like mysql requires more setup and adminstration
- we currently operate with only one database client (the content-sqlite module) so that allows us to turn on exclusive mode, bypassing all manner of race prevention that would necessarily be there for a "real" database.

At least we have a pluggable back-end so bringing on a database expert would be relatively easy since they wouldn't have to learn a lot of flux internals.

grondo:
> since sqlite is embeddable, it's available for workflow users to just turn on, while an external db like mysql requires more setup and adminstration

Yeah, sorry, I meant to say that this would be for system instance use only. Sorry for the OT question.

garlick:
I suppose the least disruptive thing we could do here is use the safe pragmas only if `statedir` is defined, and leave things as is otherwise.

grondo:
That seems like a good idea. Though I personally would not be opposed to disabling content-sqlite by default unless requested by configuration (or when statedir is set?). Then users have a way to re-enable it if there are issues and we could always change default behavior in a future release if there are problems. I don't have a strong feeling either way for what is best though.

chu11:
high level comment: should this be tried in `job-archive` as well?

garlick:
Why don't we do the least disruptive option in this PR to hopefully improve the database corruption situation. It seems like going "memory only" as the default might need to be combined with some optimization of the content-cache code, since we have the performance variability and the unexplained script runtime numbers.  Or at least further analysis.  That could come as a separate PR that would "fix" #4267.

garlick:
> high level comment: should this be tried in job-archive as well?

Hmm, yeah, although it is not running in exclusive mode, so there may be other considerations.   The safest thing would just be to use the defaults (not set journal_mode or synchronous) and see if we have a performance problem to address and go from there.

chu11:
> The safest thing would just be to use the defaults (not set journal_mode or synchronous) and see if we have a performance problem to address and go from there.

While I was working on my `job-db` prototype (#4273) I actually tried just removing those two configs, but hit issues.  Eventually just gave up and moved on.  It was a prototype and not the original `job-archive`, so other potential bugs / issues were in play.  But something to think about for later.

garlick:
> While I was working on my job-db prototype (https://github.com/flux-framework/flux-core/issues/4273) I actually tried just removing those two configs, but hit issues.

I tihnk the default mode is `journal_mode=DELETE synchronous=FULL`

I just ran a 8192 job throughput test with those modes set and minutes later it's still writing out the cache with a backlog of 255K objects.  I think it's doing about 100 objects (commits) per second.  Here's a dump of the stats so far:
```{
 "object_count": 23727,
 "dbfile_size": 6258688,
 "dbfile_free": 278189137920,
 "load_time": {
  "count": 0,
  "min": 0.0,
  "max": 0.0,
  "mean": 0.0,
  "stddev": 0.0
 },
 "store_time": {
  "count": 23711,
  "min": 8.4326059999999998,
  "max": 52.818275,
  "mean": 10.710714572139546,
  "stddev": 1.6567295147468295
 }
}
```
If you compare the mean store time to the stats above, you can see it's about three orders of magnitude slower.  One object per commit is probably not the ideal workload for this mode.

Anyway, I don't know if that gives you any insight into how this would play out in job archive or db.

garlick:
Re-pushed with the following changes
- add journal_mode=VAL and synchronous=VAL module options for testing
- log the modes and the number of objects at debug level when database is opened
- use journal_mode=OFF synchronous=OFF when running out of `${rundir}` as before
- use journal_mode=WAL synchronous=NORMAL when running out of `${statedir}`
- fix an integer overflow bug in the stats reporting for the content cache that I happened across
- more tests

src/modules/content-sqlite/content-sqlite.c
```diff
@@ -213,7 +226,9 @@ static int content_sqlite_store (struct content_sqlite *ctx,
     uint8_t hash[BLOBREF_MAX_DIGEST_SIZE];
     int hash_len;
     int uncompressed_size = -1;
+    struct timespec t0;
 
+    monotime (&t0);
```
grondo, line 233: I could be reading the diff wrong, but it looks like `t0` isn't used elsewhere in this function? Apologies if I just missed it.

grondo (approved):
LGTM! I noted one place that looked like an unused variable, but I'm probably just reading the code incorrectly.

src/modules/content-sqlite/content-sqlite.c
garlick: Oh good catch!

garlick:
Fixed and forced a push.

grondo:
I've already approved, so feel free to set MWP whenever you're ready.

garlick:
Thanks!  Setting MWP.

PR #4298, opened 2022-04-23
broker: suppress online message with no members

grondo:
Problem: When a response to the groups.get request for the
broker.online group reports no members, the broker prints a confusing
log message:

```
 [  +0.002432] broker[0]: online:  (ranks )
```

Suppress this log message if the groups.get response contains an
empty set for members. This also avoids creating and destroying an
empty hosts string.

### DIFF
--- a/src/broker/state_machine.c
+++ b/src/broker/state_machine.c
@@ -721,9 +721,11 @@ static void broker_online_cb (flux_future_t *f, void *arg)
         return;
     }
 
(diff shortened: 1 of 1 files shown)


garlick (approved):
LGTM!

garlick:
This is ready for a review if someone has a moment.

garlick:
Oops, sorry my comment was in the wrong PR!

PR #4299, opened 2022-04-24
content: track RFC 10 protocol changes

garlick:
This implements the protocol change proposed in flux-framework/rfc#330, in which blobref strings are replaced with the raw hash in content load/store RPCs.  The broker content-cache also now uses hashes instead of blobrefs as the zhashx key.

The RPC wrappers in libcontent were updated, so their users including `flux-content`, `flux-dump`, `flux-restore`, and the KVS module did not really have to change, although some of the libcontent functions were renamed.

One thing that superficially worked before but no longer works at all after this change is modifying the hash algorithm across an instance restart.  If we want to enable a system instance to migrate from sha1 to sha256, a dump/restore will be required and the content back ends will need to treat an attempt to restart under a different hash as a fatal error.  Since we don't yet have TOML support for changing the hash algorithm, I think we don't necessarily need to fix that here.

### DIFF
--- a/src/common/libcontent/content.c
+++ b/src/common/libcontent/content.c
@@ -20,7 +20,9 @@
 
 #include "src/common/libutil/blobref.h"
 
-flux_future_t *content_load (flux_t *h, const char *blobref, int flags)
+flux_future_t *content_load_byblobref (flux_t *h,
+                                       const char *blobref,
+                                       int flags)
 {
     const char *topic = "content.load";
     uint32_t rank = FLUX_NODEID_ANY;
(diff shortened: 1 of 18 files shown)


src/common/libcontent/content.c
```diff
@@ -37,14 +39,27 @@ flux_future_t *content_load_byblobref (flux_t *h,
...
+        return NULL;
+    return content_load_byhash (h, hash, hash_size, flags);
 }
 
 int content_load_get (flux_future_t *f, const void **buf, int *len)
 {
     return flux_rpc_get_raw (f, buf, len);
 }
 
+
```
chu11, line 62: nit: errant extra line added?

chu11 (approved):
looked through everything and didn't see anything other than just this 1 nit!

grondo (approved):
I had gone through this PR and did some local testing with no issues, but forgot to come back and approve. LGTM!

garlick:
Thanks guys, setting MWP.

Issue #258, opened 2015-07-10
kvs object store needs garbage collection


garlick:
The kvs object store currently grows without bound.

This is hard to address in the current design.  Reference counting would add overhead to the system.  Periodically walking the namespace to find disconnected objects ignores the fact that eventually consistent slaves may still be using those objects, or clients may be traversing old versions of the name space after `kvsdir_t` is turned into a "snapshot reference" per issue #64.

trws:
This is potentially becoming relevant again.  A long-run job I have testing longevity is, at 200,000 jobs none of which use the kvs for IO, up to 26GB of sqlite database at this point.

Do you think it would be reasonable to do something like etcd, where older versions can be traversed, but only up to a certain distance back? In their case it's a certain number of updates to a value, on the order of 1000 or something, but we might be able to establish a limit at which a client couldn't expect to be able to look back in time to ease this a bit.

trws:
An idle thought, but if we could tell the kvs explicitly that nothing will ever look at a given key again, would it be reasonable to clear out all data that was ever used to represent that key?  I'm thinking specifically of something we could include with purge to take care of old lwj data that we're explicitly deleting.

garlick:
Hmm, not sure how that would work since content blobs can be pointed to from multiple keys/directories and the content store is inherently deduplicating, with no refcount/back reference data kept with the blobs.

It seems like we need something like [git-gc](https://git-scm.com/docs/git-g) here, to identify "unreachable objects".    For us this is complicated by multiple namespaces sharing one content store, the possible existence of content references outside of any KVS namespace, and the difficulty of taking the KVS offline for any length of time to walk every reference.

I like the etcd idea.  I wonder as a first cut if we could add an epoch to each content blob and then periodically walk the namespace(s), updating the epoch for all currently-referenced blobs?  Then purge all blobs whose epoch is older than some threshold.

The other thing that seems worth pursuing is to add some "persistence flags"  to a namespace to handle
* namespaces that need no persistence at all, like the per-job PMI namespace
* namespaces that might only need their "final" snapshot captured (some jobs maybe?)
* namespaces that require strong persistence (on disk after every commit, say)

Another thought is that the current "write back" cache on rank 0 might have some opportunity to avoid writing some objects to the backing store completely, e.g. if they are "dereferenced" before being written.  Content blobs in flight could have some additional flags that affect their persistence. 

Just thinking out loud really, more discussion/thought needed.

SteVwonder:
[...]

PR #4303, opened 2022-04-26
flux-shutdown: add --gc garbage collection option

garlick:
This adds some logic to rc1 and rc3 to enable offline garbage collection of a system instance using `flux-dump(1)` and `flux-restore(1)`.  If requested by `flux shutdown --gc`, a dump is produced in ${statedir} during shutdown with a RESTORE symlink pointed to it.  On startup, if RESTORE exists, the current backing store is truncated and content is restored from the archive file.  Then the RESTORE symlink is removed.

For example on my test system:
```
$ sudo flux shutdown --gc
flux-shutdown: shutdown will dump KVS (this may take some time)
broker.info[0]: cleanup.0: flux queue stop --quiet Exited (rc=0) 0.0s
broker.info[0]: cleanup.1: flux job cancelall --user=all --quiet -f --states RUN Exited (rc=0) 0.0s
broker.info[0]: cleanup.2: flux queue idle --quiet Exited (rc=0) 0.0s
broker.info[0]: cleanup-success: cleanup->shutdown 0.104717s
broker.info[0]: children-none: shutdown->finalize 0.179421ms
broker.info[0]: rc3.0: dumping content to /var/lib/flux/dump-20220425_174130.tgz
broker.info[0]: rc3.0: /usr/local/etc/flux/rc3 Exited (rc=0) 0.6s
broker.info[0]: rc3-success: finalize->goodbye 0.578407s
$ sudo ls -l /var/lib/flux
total 39684
-rw-r--r-- 1 flux flux  1073152 Apr 25 17:41 content.sqlite
-rw-r--r-- 1 flux flux    48183 Apr 25 17:41 dump-20220425_174130.tgz
-rw-r--r-- 1 flux flux 39510016 Apr 24 18:09 job-archive.sqlite
lrwxrwxrwx 1 flux flux       24 Apr 25 17:41 RESTORE -> dump-20220425_174130.tgz
$ sudo systemctl start flux
$ sudo ls -l /var/lib/flux
total 40896
-rw-r--r-- 1 flux flux     4096 Apr 25 17:42 content.sqlite
-rw-r--r-- 1 flux flux  2307232 Apr 25 17:42 content.sqlite-wal
-rw-r--r-- 1 flux flux    48183 Apr 25 17:41 dump-20220425_174130.tgz
-rw-r--r-- 1 flux flux 39510016 Apr 24 18:09 job-archive.sqlite
```

It's also possible to checkpoint/restart an instance with:
```
$ flux shutdown --dump=foo.tar.bz2
$ flux start -o,-Scontent.restore=foo.tar.bz2
```

although this is only practical if R hasn't changed at the moment.

Marking a WIP as I wanted to get feedback on the approach before writing tests.

I had thought maybe garbage collection could be automated somehow by tracking some metric that could be used as an indicator of the need.  However, maybe getting a bit of experience with doing it manually first makes sense.

### DIFF
--- a/etc/rc1
+++ b/etc/rc1
@@ -3,10 +3,6 @@
 # Allow connector-local more time to start listening on socket
 RANK=$(FLUX_LOCAL_CONNECTOR_RETRY_COUNT=30 flux getattr rank)
 
-if ! content_backing=$(flux getattr content.backing-module 2>/dev/null); then
-    content_backing=content-sqlite
-fi
-
 # Usage: modload {all|<rank>} modname [args ...]
 modload() {
     local where=$1; shift
@@ -16,7 +12,34 @@ modload() {
 }
 
 modload all barrier
-modload 0 ${content_backing}
+
+if test $RANK -eq 0; then
+    backingmod=$(flux getattr content.backing-module 2>/dev/null) || :
+    backingmod=${backingmod:-content-sqlite}
+    dumpfile=$(flux getattr content.restore 2>/dev/null) || :
+    if test -n "${dumpfile}"; then
+        if test "${dumpfile}" = "auto"; then
+            statedir=$(flux getattr statedir 2>/dev/null) || :
+            dumplink="${statedir:-.}/dump/RESTORE"
+            if test -h "${dumplink}"; then
+                dumpfile=$(readlink -f ${dumplink}) || :
+            else
+                dumpfile=""
+                dumplink=""
+            fi
+        fi
+    fi
+    if test -n "${dumpfile}"; then
+        flux module load ${backingmod} truncate
+        echo "restoring content from ${dumpfile}"
+        flux restore --quiet --checkpoint ${dumpfile}
+        if test -n "${dumplink}"; then
+            rm -f ${dumplink}
+        fi
+    else
+        flux module load ${backingmod}
+    fi
+fi
 
 modload all kvs
 modload all kvs-watch
--- a/etc/rc3
+++ b/etc/rc3
@@ -45,7 +45,25 @@ modrm all kvs
 
 flux content flush || exit_rc=1
 
-backingmod=$(flux getattr content.backing-module 2>/dev/null)
-modrm 0 ${backingmod:-content-sqlite}
+if test $RANK -eq 0; then
+    backingmod=$(flux getattr content.backing-module 2>/dev/null)
+    backingmod=${backingmod:-content-sqlite}
+    dumpfile=$(flux getattr content.dump 2>/dev/null)
+    if test $exit_rc -eq 0 -a -n "${dumpfile}"; then
+        if test "${dumpfile}" = "auto"; then
+            statedir=$(flux getattr statedir 2>/dev/null)
+            mkdir -p "${statedir:-.}/dump"
+            dumpfile="${statedir:-.}/dump/$(date +%Y%m%d_%H%M%S).tgz"
+            dumplink="${statedir:-.}/dump/RESTORE"
+        fi
+        echo "dumping content to ${dumpfile}"
+        if flux dump --quiet --checkpoint ${dumpfile}; then
+            test -n "$dumplink" && ln -s $(basename ${dumpfile}) ${dumplink}
+        else
+            exit_rc=1
+        fi
+    fi
+    flux module remove ${backingmod} || exit_rc=1
+fi
 
 exit $exit_rc
--- a/src/modules/content-sqlite/content-sqlite.c
+++ b/src/modules/content-sqlite/content-sqlite.c
@@ -807,7 +807,7 @@ static int process_args (struct content_sqlite *ctx,
             *truncate = true;
         }
         else {
-            flux_log_error (ctx->h, "Unknown module option: '%s'", argv[i]);
+            flux_log (ctx->h, LOG_ERR, "Unknown module option: '%s'", argv[i]);
             return -1;
         }
     }
--- a/src/modules/content-sqlite/content-sqlite.c
+++ b/src/modules/content-sqlite/content-sqlite.c
@@ -76,6 +76,7 @@ struct content_sqlite {
     struct content_stats stats;
     const char *journal_mode;
     const char *synchronous;
+    bool truncate;
 };
 
 static void log_sqlite_error (struct content_sqlite *ctx, const char *fmt, ...)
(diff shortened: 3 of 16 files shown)


garlick:
I went ahead and added test coverage here and have been testing this on my home test system, so removing the WIP.  I'm still open to doing this another way if people have better ideas!

grondo:
Sorry, I haven't had time to take a peek at this. I can't think of a any better interface, given that garbage collection must occur as part of a dump/restore. I can't remember, does GC happen just as a natural result of the restore? 

As a more general point (and I guess unrelated to this PR), I'm a little worried there will be confusion when to use `flux shutdown` vs systemd for stopping a Flux system instance. I'm not sure there are many systemd services that are stopped via a different command.

garlick:
Yes a dump/restore walks the KVS metadata starting with the last root hash, so when it is restored, all the unreferenced data is left behind.  In addition, the archive only contains "files", unlike a file system archive created with tar (for example).  So empty job directories are removed also.

One can still run `systemctl stop flux` on the rank 0 onde to shut down the instance.  And one could trigger a dump/restore as part of that by manually setting the `content.dump` attribute to `auto`.  However I didn't want to encourage that because then the dump would be subject to the systemd `TimeoutStopSec`  and would risk getting killed before the dump is complete.  Hence adding the option to `shutdown` so it's tied to that way of bringing down the instance.  I anticipate that we will add other things to `flux-shutdown(1)` like options to stop the queue and let running jobs complete, or to shut down in the future.  So maybe it will become a natural way to stop flux.

One other weakness I see here is the dump files aren't removed and will pile up after a while.  I was vaguely thinking that this could be valuable if we wanted to revert to a previous checkpoint if the db was corrupted or whatever, and that sys admins could manage the dump files with log rotation tools.  Is that reasonable?

I guess the other question - is `--gc`  an annoyingly terse option?  It's just a shorthand for `--dump=auto`.  Should we just go with that?  The purpose of `--gc` was to provide an option that matched the desired end effect (garbage collecting the KVS).  I'm open to better  names.

grondo:
Ah, thanks for that refresher, that was helpful. Given the above, this approach seems just fine IMO. I like the idea of extending shutdown semantics in the future. Also, if dumpfiles tend to accumulate, maybe logrotate or systemd-tmpfiles could be configured to automatically clean things up?

grondo:
And FWIW, I don't have a problem with `--gc`. It is close enough to `git gc` where I understand what it is meant to do.

garlick:
Pushed a tmpfiles.d config file and moved "auto" dumps to `$statedir/dump` since that made the tmpfiles rule easier to write.

Tested on my test system instance by running `flux shutdown --gc` a few times and reducing the age setting for dump files, then observing that they were purged with `systemd-tmpfiles --clean`

garlick:
Repushed with reference to #258

src/modules/content-sqlite/content-sqlite.c
```diff
@@ -807,7 +807,7 @@ static int process_args (struct content_sqlite *ctx,
             *truncate = true;
         }
         else {
-            flux_log_error (ctx->h, "Unknown module option: '%s'", argv[i]);
+            flux_log (ctx->h, LOG_ERR, "Unknown module option: '%s'", argv[i]);
```
chu11, line 810: perhaps should stylize do a similar change in `content-files`?  (`content-files` sets errno = EINVAL to make it not as bad)

chu11: oh and I guess w/ `content-s3` too (given follow up commit to this one)

t/t0018-content-files.t
```diff
@@ -73,6 +73,10 @@ kvs_checkpoint_get() {
 # Tests of the module by itself (no content cache)
 ##
 
+test_expect_success 'content-files module load fails with unknown option' '
+	test_must_fail flux module load content-files notoption
+'
+
```
chu11, line 79: nit, should this test be a different commit? not really related to ` content-files: add truncate module option `

etc/rc3
```diff
@@ -45,7 +45,25 @@ modrm all kvs
...
+            dumplink="${statedir:-.}/dump/RESTORE"
+        fi
+        echo "dumping content to ${dumpfile}"
+        if flux dump --quiet --checkpoint ${dumpfile}; then
+            test -n "$dumplink" && ln -s $(basename ${dumpfile}) ${dumplink}
+        else
+            exit_rc=1
+        fi
+    fi
+    flux module remove ${backingmod} || exit_rc=1
```
chu11, line 66: should use `modrm`?

etc/rc1
```diff
@@ -16,7 +12,34 @@ modload() {
...
+            if test -h "${dumplink}"; then
+                dumpfile=$(readlink -f ${dumplink}) || :
+            else
+                dumpfile=""
+                dumplink=""
+            fi
+        fi
+    fi
+    if test -n "${dumpfile}"; then
+        flux module load ${backingmod} truncate
```
chu11, line 33: use modload for consistency?

etc/rc1
```diff
@@ -16,7 +12,34 @@ modload() {
...
+        echo "restoring content from ${dumpfile}"
+        flux restore --quiet --checkpoint ${dumpfile}
+        if test -n "${dumplink}"; then
+            rm -f ${dumplink}
+        fi
+    else
+        flux module load ${backingmod}
```
chu11, line 40: use modload for consistency?

etc/rc1
```diff
@@ -16,7 +12,34 @@ modload() {
...
+            fi
+        fi
+    fi
+    if test -n "${dumpfile}"; then
+        flux module load ${backingmod} truncate
+        echo "restoring content from ${dumpfile}"
+        flux restore --quiet --checkpoint ${dumpfile}
+        if test -n "${dumplink}"; then
+            rm -f ${dumplink}
+        fi
```
chu11, line 38: should only remove the link if the restore is successful?

chu11 (commented):
overall lgtm, just a few comments / nits I found

etc/rc3
garlick: `modrm` tests $RANK before running `flux module remove`, but this is within a block that is already conditional on $RANK.   I thought it kind of weird to trade a straightforward one-liner for a function call to do same when it was necessary to repeat the rank constraint.  Does that make sense?

etc/rc1
garlick: same deal here

etc/rc1
garlick: ibid

etc/rc1
garlick: Yes but the rc1 shebang is `#!/bin/bash -e` so the script aborts before the remove if the restore is unsuccessful.

garlick:
Just pushed fixes based on @chu11's comments, and also rebased on current master.

etc/rc3
chu11: ahh that makes sense, you have the `rank == 0` check above this.

chu11 (approved):
LGTM!

garlick:
Thanks!  I'll set MWP.

PR #4307, opened 2022-04-28
job-archive: use safer pragmas

chu11:
Following up PR #4294 where `content-sqlite` used safer pragmas, this updates `job-archive` to do the same.

I wasn't entirely sure of the best way to test performance with `job-archive`.  Because it "polls" `job-list` data, there's a non trivial amount of wall clock time dependent on that polling.

So I ended up instrumenting `job-archive` with the following:

```
+    if (clock_gettime (CLOCK_MONOTONIC, &tp_start) < 0)
+        flux_log_error (ctx->h, "clock_gettime start");
     while (sqlite3_step (ctx->store_stmt) != SQLITE_DONE) {
         /* due to rounding errors in sqlite, duplicate entries could be
          * written out on occassion leading to a SQLITE_CONSTRAINT error.
@@ -381,6 +387,10 @@ void job_info_lookup_continuation (flux_future_t *f, void *arg)
             goto out;
         }
     }
+    ctx->total += monotime_since (tp_start);
+    ctx->count++;
+    if (ctx->count % 25 == 0)
+        flux_log (ctx->h, LOG_DEBUG, "count = %u, total = %f, average time = %f", ctx->count, ctx->total, ctx->total / ctx->count);
```

Basically, for every job archived, count up the time it takes to write an entry to the sqlite archive.  So no measurement time wasted on polling or job-info lookups in the KVS or anything else like that.

The results on 4096 jobs run (note these are all milliseconds, run on my laptop).

current master (`journal_mode=OFF` && `synchronous=OFF`)

`2022-04-28T00:14:15.180731Z job-archive.debug[0]: count = 4075, total = 327.133517, average time = 0.080278`

`journal_mode=WAL` && `synchronous=NORMAL`

`2022-04-28T00:39:40.350984Z job-archive.debug[0]: count = 4075, total = 541.424963, average time = 0.132865`

`journal_mode=DELETE` && `synchronous=FULL`

`2022-04-28T00:57:53.516332Z job-archive.debug[0]: count = 4075, total = 30441.640082, average time = 7.470341`

no pragmas set

`2022-04-28T01:21:22.601982Z job-archive.debug[0]: count = 4075, total = 30171.204437, average time = 7.403977a

So WAL & NORMAL slow things down 65%-ish.  But the default of DELETE & FULL is awful, slowing it down 100x-ish.

So I went with updating the pragmas with WAL & NORMAL, just like in `content-sqlite`.

### DIFF
--- a/src/modules/job-archive/job-archive.c
+++ b/src/modules/job-archive/job-archive.c
@@ -177,7 +177,7 @@ int job_archive_init (struct job_archive_ctx *ctx)
     }
 
     if (sqlite3_exec (ctx->db,
-                      "PRAGMA journal_mode=OFF",
+                      "PRAGMA journal_mode=WAL",
                       NULL,
                       NULL,
                       NULL) != SQLITE_OK) {
@@ -185,7 +185,7 @@ int job_archive_init (struct job_archive_ctx *ctx)
         goto error;
     }
     if (sqlite3_exec (ctx->db,
-                      "PRAGMA synchronous=OFF",
+                      "PRAGMA synchronous=NORMAL",
                       NULL,
                       NULL,
                       NULL) != SQLITE_OK) {
--- a/src/modules/job-archive/job-archive.c
+++ b/src/modules/job-archive/job-archive.c
@@ -26,6 +26,8 @@
 
 #include "src/common/libutil/log.h"
 #include "src/common/libutil/fsd.h"
+#include "src/common/libutil/tstat.h"
+#include "src/common/libutil/monotime.h"
 
 #define BUSY_TIMEOUT_DEFAULT 50
 #define BUFSIZE              1024
@@ -65,6 +67,7 @@ struct job_archive_ctx {
     sqlite3_stmt *store_stmt;
     double since;
     int kvs_lookup_count;
+    tstat_t sqlstore;
 };
 
 static void log_sqlite_error (struct job_archive_ctx *ctx, const char *fmt, ...)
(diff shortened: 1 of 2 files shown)


garlick:
This seems reasonable.  Couple of notes come to mind
- might as well bring over the `tstat_t` performance measurement stuff from `content-sqlite` as it captures min/max/stddev, and there will be performance variation especially in WAL mode
- there is more of a performance hit here than in `content-sqlite` since `synchronous` mode is not OFF
- WAL mode also affects read performance for users like flux-accounting
- It might be interesting also to measure overall wall clock for archiving N jobs

garlick:
OH my bad, I meant `locking_mode=EXCLUSIVE` in `content-sqlite`.

chu11:
Added a commit for some basic module stats using `tstat`.  Dunno if I should encapsulate within an object for future stat flexibility?  I ended up not doing so for now.

Stats are in-line with above.  note that I captured more sql "overhead" in this stat gathering than my instrumentation above.

OFF/OFF

```
 "count": 4097,
 "min": 0.067011000000000001,
 "max": 5.6384309999999997,
 "mean": 0.13771830070783483,
 "stddev": 0.14469457659048435
```

WAL/NORMAL

```
 "count": 4097,
 "min": 0.056492000000000001,
 "max": 41.101869999999998,
 "mean": 0.20180585672443221,
 "stddev": 1.4494089776207182
```

DELETE/FULL

```
 "count": 4097,
 "min": 3.6344460000000001,
 "max": 53.730946000000003,
 "mean": 7.6258716375396771,
 "stddev": 4.2748313000976479
```

garlick (approved):
This looks good - thanks for adding the stats!

Although it may be interesting to explore performance in more depth and see if there is any measurable impact on the accounting scripts, I think avoiding data corruption here is likely more important, so I'm good with this going in as is.

Pinging @cmoussa1 so he is aware that this might have a (likely small) performance impact on his scripts.

Issue #2795, opened 2020-03-04
content-sqlite declares hash keys as CHAR(20) but hashes can be longer


garlick:
The content-sqlite object table is created with this SQL command

```sql
CREATE TABLE if not exists objects(
    hash CHAR(20) PRIMARY KEY, size INT, object BLOB
)
```

20 bytes is the size of the sha1 digest.  If sha256 is selected, it looks like the hash would be truncated to 20 bytes.  Need to double check this.

garlick:
Actual bug, needs to be fixed.

grondo:
I've set the `confirmed` label which will be ignored by stalebot.

garlick:
Poking at this I noticed that sha256 hashes do not appear to be truncated in the database after all.  After running an instance with `content.hash=sha256`, I opened the database with `sqlite3`:
```
sqlite> select hex(hash) from objects;
03ADFF92E9BA85F30E86BED289F306F382D34EA55D9665C2FDC780123D45DF5B
078223D5ACC2BE2323F40E3E7A5328325AF1FCFD15CEA27FB9E9171B23029A26
090051AA98AF0B08D8B02498DD08402D174FD2350A7BF15E005FDF6AD8387AB8
0A7DA187AE93FCA3F8C5EBDBC2B1F5787A9168000147F170508E1742368AB0A1
0DE5395C30BFDC16C69343CDE115C7510BB74BFAA5D9BC356D793A116A99FE11
0FEB1FA1F50C0A76813CBAB946701EF784EEC66799DE5FDCE35DDA980BBC2310
1027F471D2ED1CCA59303A543DAD18EA687CC00FB5E4CBDBA27CC01F18B9C2FB
[snip]
```
These lines are 64 hex characters long, so 32 byte values were stored in the hash column.  sha256 hashes are 32 bytes long.  As explained [here](https://www.sqlite.org/datatype3.html), sqlite ignores values in parenthesis following a type name.  All sqlite values can have a dynamic length.

We should just drop the length from our code since it looks wrong and accomplishes nothing.

Also, I noted in that document that declaring the type of this key to be TEXT causes it to be stored using the database character encoding (e.g. UTF-8) and associates a collating function with the column for key comparisons.  BLOB on the other hand appears to work just fine as a primary key, and bypasses all that, using `memcmp()` for comparisons.  In addition, in a quick test, switching to BLOB resulted n a modest decrease in database size (like 10 percent).

garlick:
[...]

PR #4315, opened 2022-05-03
use correct type in content-sqlite, misc. content test cleanup

garlick:
This PR modifies the schema of the content-sqlite database to use BLOB instead of CHAR(20) as the type for the hash key in the objects table, as discussed in #2795.

Although this is a schema change, it shouldn't cause problems for system instances on upgrade.  The change only affects the command used to create the objects table if does not already exist.  So if the database already has an objects table, it will continue to use CHAR(20).  The next time the database is recreated (for example after `flux shutdown --gc`) the new column type will be used.

There is also some minor test cleanup here.


chu11 (approved):
LGTM

garlick:
Thanks!   Setting MWP.

PR #4319, opened 2022-05-04
README: add libarchive prerequisite

garlick:
Problem: README does not list libarchive build prerequisite.

Add libarchive to prerequiste table and os-specific package lists.


grondo (approved):
Thanks!

PR #4320, opened 2022-05-04
python: uri: fix intermittent failure of fallback pid resolver

### LINKED ISSUE #4168: test failure: t2802-uri-cmd.t 8 - flux-uri pid resolver works on flux-broker with fallback
everyonce in awhile I get this failure when I run `make -j16 check`.  

```
test_expect_success 'flux-uri pid resolver works on flux-broker with fallback' '
        (
          export FLUX_FORCE_BROKER_CHILD_FALLBACK=t &&
          test "$(flux uri pid:$(flux getattr broker.pid))" =  "$FLUX_URI"
        )
'
```

I assume something racy, although what it could be escapes me.

Edit: side note, tests that occasionally fail but have a `-t 30` or some similar timeout, its easy to just accept the failure.  But in this test I can't tell if a timeout would be reasonable.

grondo:
Problem: The fallback PID URI resolver was failing intermittently
during testing due to what was presumed whitespace in the command
string found in /proc/[pid]/stat. Since the /proc/[pid]/stat contents
were split into fields on whitespace, the incorrect field was passed
to int() and an exception thrown, resulting in the inscrutable error:

 flux-uri: ERROR: invalid literal for int() with base 10: 'S'

To fix, use a regex instead of simply splitting the contents on
whitespace. This allows whitespace to possibly occur in the command
name, since we now consume everything between parentheses as the
command name.

Fixes #4168

### DIFF
--- a/src/bindings/python/flux/uri/resolvers/pid.py
+++ b/src/bindings/python/flux/uri/resolvers/pid.py
@@ -37,9 +37,15 @@ def _get_broker_child_fallback(broker_pid):
             #  of a TOU-TOC race here, so just ignore all errors
             pass
         else:
-            ppid = int(data.split()[3])
-            if ppid == broker_pid:
-                return pid
+            match = re.match(r"^[0-9]+ \(.*\) \w+ ([0-9]+)", data)
+            #  Attempt to convert match to integer. On regex match failure,
+            #   or integer conversion failure, just skip this entry
+            try:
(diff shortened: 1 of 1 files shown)


chu11 (approved):
LGTM! Thanks, glad to finally have this one fixed.

grondo:
Thanks for the help running it down!

PR #4321, opened 2022-05-04
testsuite: increase test timeout

chu11:
While testing the patch in #4320, I did hit:

`FAIL: t2607-job-shell-input.t 7 - flux-shell: multiple jobs, each want stdin`

once or twice.  Assuming it was a timeout issue.  Increase the job timeout in this test.


grondo (approved):
Sounds good to me!

PR #4322, opened 2022-05-05
flux-overlay: replace --no-color with --color=WHEN option, add -H, --highlight=TARGETS

grondo:
While I was working on something recently, I wanted an option like the [pstree(1)](https://linux.die.net/man/1/pstree) `--highlight` option for `flux overlay status` to highlight one or more ranks and their ancestors in the overlay. For example, for the new distributed job-exec module, it would be nice to easily know which ranks handled messages for a job after knowing the job's assigned idset or hostlist.

This PR adds a new `-H, --highlight=TARGETS` to `flux overlay status` which will highlight the given ranks and their ancestors (i.e. the subtree involving TARGETS) in color if color is enabled or via the special characters `<< >>` if not (useful for testing, and also _almost_ as visible as color).

This PR also replaces the `--no-color` option with the more standard `--color=WHEN` option we've used in a couple other commands.  Since `--color=auto` is the default, color is now disabled by default when stdout is not connected to a tty, which means `--no-color` can simply be dropped where it was required in test.

This was a little more work than I initially thought when I started the endeavour, however since the feature is nominally working I elected to share it here. I realize the feature only fulfills a rare corner case, so I would leave it up to others if we want to actually merge it.

Examples:

Screenshot from 2022-05-04 21-13-01

Screenshot from 2022-05-04 21-16-23

```console
ƒ(s=16,d=0,builddir) grondo@asp:~$ flux overlay status -H 10-12 --color=never
<<0 asp>>: full
├─ <<1 asp>>: full
│  ├─ 3 asp: full
│  │  ├─ 7 asp: full
│  │  │  └─ 15 asp: full
│  │  └─ 8 asp: full
│  └─ <<4 asp>>: full
│     ├─ 9 asp: full
│     └─ <<10 asp>>: full
└─ <<2 asp>>: full
   ├─ <<5 asp>>: full
   │  ├─ <<11 asp>>: full
   │  └─ <<12 asp>>: full
   └─ 6 asp: full
      ├─ 13 asp: full
      └─ 14 asp: full
```

### DIFF
--- a/src/cmd/builtin/overlay.c
+++ b/src/cmd/builtin/overlay.c
@@ -11,6 +11,7 @@
 #if HAVE_CONFIG_H
 # include <config.h>
 #endif
+#include <unistd.h>
 #include <jansson.h>
 #include <flux/core.h>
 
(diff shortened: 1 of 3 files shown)


garlick (approved):
Just played with this a bit.  Very neat!  Changes LGTM.

Issue #4313, opened 2022-05-02
job-list: support attribute to grab all job-list data for a specific job


chu11:
Per discussion in #4234, it would be convenient if there was a `job-list` attribute to grab all data associated with a job.  Hypothetically the attribute could be "dump " or "all".  It has multiple benefits:

- its easier to do this than requesting like 25+ different attributes from `job-list` when we want a lot of the job data
  - in some frontend commands this is already done
  - could be useful when needing to archive lots of job data (#4273)
- we could archive just this one JSON object and re-load it semi-quickly (#4273)
  - and this is easier than having a sql table with like 25+ columns and inserts with 25+ column names
- flexible for future queryable things in job-list that don't exist today

Issue #3572, opened 2021-03-25
adding new job "attribute" is painful


grondo:
Currently, in order to add a new job "attribute" to Flux you have to update at least 5 locations:

 * `const char *attrs[]` list of attributes in job-list `list_attrs_cb()`
 * `const char *list_attrs` string in `flux-job.c` for `cmd_list()`
 * header format dictionary in `flux/job/info.py` in the Python bindings
 * `VALID_ATTRS` constant list of valid attributes in `flux/job/list.py`in the Python bindings
 * The format to attrs mapping in `flux-jobs.py`

It would be nice if this was a little easier, though I haven't thought of exactly how to do that yet.

chu11:
#4313 may be able to solve a few of these, which helps

chu11:
after merge of #4324 this list is shorter / modified

 * `const char *attrs[]` list of attributes in job-list `job-list.c`
 * header format dictionary in `flux/job/info.py` in the Python bindings
 * the expected list-attrs list in `t/python/t0010-job.py`
 * The format to attrs mapping in `flux-jobs.py`

PR #4324, opened 2022-05-05
job-list: support new "all" attribute to get all job attributes

chu11:
Per discussion in #4313, support a new "all" attribute to get all job-list attrs from a job.

- its much easier than listing / maintaining like 25+ keywords
- has potential benefits when we want to archive all data in 1 big blob
- flexibility for future new attributes and need to update less stuff

side notes:

- i didn't support this in `flux-jobs` as I felt it wasn't necessary
- i pondered ways to make it so we didn't need to maintain a list of attrs in `t/python/t0010-job.py`.  Like hypothetically stick all attrs in a text file and have a script that generates C code / python list.  But I'll leave that for another day. (#3572)

### DIFF
--- a/src/bindings/python/flux/job/list.py
+++ b/src/bindings/python/flux/job/list.py
@@ -17,35 +17,6 @@ from flux.job.info import JobInfo
 from flux.rpc import RPC
 
 
-VALID_ATTRS = [
-    "userid",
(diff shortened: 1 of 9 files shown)


grondo (approved):
This LGTM! Nice addition.

PR #4326, opened 2022-05-11
job-list: support `states_mask` attribute

chu11:
Problem: It would be convenient to know which job states were actually reached during a job's lifecycle.  However there is
no way to retrieve this information, even though a "states_mask" is already stored internally in `job-list`.

Solution: Support a new 'states_mask' attribute to retrieve the 'states_mask' from job-list.

This is mostly in preparation for future `job-db` support (#4273), as we'd need to store it and later retrieve it.  I suppose it is generally potentially useful too.

Thoughts:

- didn't add support in `flux-jobs`, as I feel it is low probability needed by users, atleast at the moment
- I went back and forth on the name of this.  It is called `states_mask` internally in `job-list`. I just kept it b/c I couldn't think of any name better than this.  I'm open to any suggestions.

### DIFF
--- a/t/job-manager/job-conv.py
+++ b/t/job-manager/job-conv.py
@@ -25,6 +25,43 @@ def statetostr(args):
         print(raw.flux_job_statetostr(int(state), fmt).decode("utf-8"))
 
 
+# return value of highest state bitmask, which is INACTIVE state
+def get_highest_powerof2():
+    state = ffi.new("flux_job_state_t [1]")
+    raw.flux_job_strtostate("INACTIVE", state)
+    return (int)(state[0])
+
+
+# convert bitmask to list of powers of 2, in increasing order
+def bitmask2powersof2(n):
+    l = []
+    p = get_highest_powerof2()
+    while p > 0 and n > 0:
+        if p <= n:
+            l.insert(0, p)
+            n = n - p
+        p = (int)(p / 2)
+    return l
+
+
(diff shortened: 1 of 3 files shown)


garlick:
> This is mostly in preparation for future job-db support (https://github.com/flux-framework/flux-core/issues/4273), as we'd need to store it and later retrieve it. I suppose it is generally potentially useful too.

Is this still needed now that job-db is turning into a replacement for job-list?

I should probably know, but what's it used for internally?

chu11:
> Is this still needed now that job-db is turning into a replacement for job-list?

Yeah, its necessary to reconstruct internal job state.

> I should probably know, but what's it used for internally?

Basically its used to determine what data is available for the user.  For example, if a job was never run (e.g. resources can't be satisfied) we know that data related to a job's run (nodelist, etc.) are not valid.

garlick:
Who is consuming this information outside of job-list then?  It seems like it is for job-db but you said job-db was going to be integrated with job-list.

chu11:
> Who is consuming this information outside of job-list then? It seems like it is for job-db but you said job-db was going to be integrated with job-list.

Ahhh, i see your point.  I guess it could be a "hidden" job-list attribute only within `job-list`, since there are no external consumers.  That makes sense.

garlick:
OK, I was just trying to understand why it was needed.  If it's not needed externally then it probably makes sense to keep it internal only.

PR #4327, opened 2022-05-12
job-list: consistently return job attributes

chu11:
Problem: The job-list module was inconsistent in its return of
job data to the user.  In some cases, "defaults" were returned
when data was never initialized.  In other cases, the data is
simply not returned to the user.

Solution: Consistently do not return data to the user when it
is not available / not set.  The attributes adjusted are:

name
ntasks
nnodes
ranks
nodelist
expiration

Adjust tests accordingly.

In the Python JobInfo class, set display defaults for name and ntasks.

### DIFF
--- a/src/bindings/python/flux/job/info.py
+++ b/src/bindings/python/flux/job/info.py
@@ -199,6 +199,8 @@ class JobInfo:
         "t_cleanup": 0.0,
         "t_inactive": 0.0,
         "expiration": 0.0,
+        "name": "",
+        "ntasks": "",
         "nnodes": "",
         "priority": "",
         "ranks": "",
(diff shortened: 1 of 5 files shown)


chu11:
i threw a couple of cleanups on top of this PR, and also rebased it based on #4326 so that we have a consistent order to the PRs.  (note, that it is independent on #4326).

chu11:
pushed two more stupid cleanups on top of this PR

chu11:
agh, more cleanups are coming.  screw it, all those cleanups will go into a new PR.  This is now a 1 commit PR :-)

garlick (approved):
This seems reasonable to me (the attr consistency commit, not the commits from #4326 which I think you're reworking)

chu11:
re-pushed, just the one commit, no longer based on #4326

PR #4330, opened 2022-05-17
shell: improve handling of TMPDIR

grondo:
This PR fixes a couple `TMPDIR` issues discovered recently by @bhatiaharsh, specifically:

 * If `TMPDIR` is set to `FLUX_JOB_TMPDIR`, as may be the case when a job is submitted from another job, clear `TMPDIR` instead of trying to reuse and/or create it.
 * If `TMPDIR` can't be created, e.g. the parent directory does not exist, do not raise a fatal error. Instead fall back to `/tmp` with a warning.

### DIFF
--- a/src/shell/tmpdir.c
+++ b/src/shell/tmpdir.c
@@ -99,10 +99,8 @@ static int tmpdir_init (flux_plugin_t *p,
     cleanup_push_string (cleanup_directory_recursive, jobtmp);
 
     /* Set/change FLUX_JOB_TMPDIR to jobtmp.
-     * If TMPDIR is unset, set it to $FLUX_JOB_TMPDIR.
      */
-    if (flux_shell_setenvf (shell, 1, "FLUX_JOB_TMPDIR", "%s", jobtmp) < 0
-        || flux_shell_setenvf (shell, 0, "TMPDIR", "%s", jobtmp) < 0)
+    if (flux_shell_setenvf (shell, 1, "FLUX_JOB_TMPDIR", "%s", jobtmp) < 0)
         shell_die_errno (1, "error updating job environment");
 
     return 0;
--- a/src/shell/tmpdir.c
+++ b/src/shell/tmpdir.c
@@ -61,10 +61,11 @@ static int mkjobtmp_rundir (flux_shell_t *shell, char *buf, size_t size)
     return 0;
 }
 
-static int mkjobtmp_tmpdir (flux_shell_t *shell, char *buf, size_t size)
+static int mkjobtmp_tmpdir (flux_shell_t *shell,
+                            const char *tmpdir,
+                            char *buf,
+                            size_t size)
 {
-    const char *tmpdir = flux_shell_getenv (shell, "TMPDIR");
-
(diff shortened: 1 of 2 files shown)


grondo:
I'm looking into why these changes break some of the inception tests that run subinstances. Better to hold off on a review until I find the problem.

grondo:
Well, that took a little while to track down.

The problem was that the first solution _unset_ `TMPDIR` if it was equal to `FLUX_JOB_TMPDIR` (i.e. this job was submitted from within another job). This then caused the shell `tmpdir` plugin to set `TMPDIR` to the newly constructed `FLUX_JOB_TMPDIR`, which is the default behavior if `TMPDIR` is not set.

If the shell was launching a new instance, the broker `rundir` was set under `TMPDIR`. Since `FLUX_JOB_TMPDIR` is constructed under `rundir` the path of the job tmpdir would start to get long and if one of the child jobs in this instance was another instance, the AF_UNIX path length would overflow since this socket is placed under `rundir`.

This didn't occur before this fix since once `TMPDIR` was set to `FLUX_JOB_TMPDIR` the first time, that became the new `TMPDIR` for that entire job hierarchy from that point on. This didn't cause a problem in all our testing since we always run single node tests. :disappointed: 

The solution is to reset `TMPDIR=/tmp` when `TMPDIR == FLUX_JOB_TMPDIR`. This short-circuits the recursive growth of `TMPDIR`. Though it basically means `TMPDIR` will effectively be `/tmp` for jobs in all batch jobs, unless `TMPDIR` is previously set to something else. Perhaps we should ponder that a moment before considering this issue to be solved.

garlick:
That seems workable!

Do you remember the reason why we are setting TMPDIR in the first place?  I can't for the life of me remember, although I'm pretty sure there was a good practical reason.

grondo:
> Do you remember the reason why we are setting TMPDIR in the first place? I can't for the life of me remember, although I'm pretty sure there was a good practical reason.

I was trying to remember, but I cannot. I seem to vaguely remember it was a nice property that `runder` was under a job-specific tmpdir, but that only works to about one level deep... Maybe it would be fine to just leave `TMPDIR` unset instead of setting it to `FLUX_JOB_TMPDIR`. That would likely be another way to solve this issue.

garlick:
I'd be tempted to say let's try leaving it alone, although I don't want to waste too much of your time when there was probably a reason we were setting it.  Your call!

grondo:
Well, the current "fix" here is not sitting well with me anyhow. You get `TMPDIR` == `FLUX_JOB_TMPDIR` at level 1 but after that `TMPDIR` is effectively `/tmp` due to some special case workaround.

I've revised the PR to just leave `TMPDIR` alone, which actually removes a line of code instead of adding several. It seems a little easier to follow too.

garlick (approved):
Thanks for trying that!  Sounds good to me.  I hope we don't soon discover the reason we were setting TMPDIR before, but then we have some time before the next release.

grondo:
Ok, I'll set MWP and we'll see how it goes.

PR #4334, opened 2022-05-18
job-manager: do not send purged events

### LINKED ISSUE #4331: job-manager: purge does not clear relevant journal events
While working on #4273 I think I hit a bug independent of that work.

if you purge events in `job-manager` (subsequently clearing them from the KVS), events in the journal for those jobs may still exist.  When reloading the `job-list` module, the job-list may get job ids that are illegal/have no data for them.

I think the correct fix would be to have the `job-manager` to clear those events in the journal.  `job-list` could handle this as well by checking if a jobid is legal and just not processing it.  But since there's typically many events for a job, it'd be a lot of ID checking.

chu11:
Problem: If a job has been purged from the job-manager, events in the journal may still exist for those purged jobs.  They can be sent to journal requestors.
    
Solution: When a new request is made for the journal, check that the events in the journal are still valid before sending them.
    
Fixes #4331

### DIFF
--- a/src/modules/job-manager/journal.c
+++ b/src/modules/job-manager/journal.c
@@ -185,14 +185,21 @@ static void journal_handle_request (flux_t *h,
     wrapped_entry = zlist_first (journal->events);
     while (wrapped_entry) {
         const char *name;
+        flux_jobid_t id;
 
         if (json_unpack (wrapped_entry,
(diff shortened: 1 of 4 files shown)


garlick (approved):
Nicely done.  That turned out to be pretty simple.

PR #4335, opened 2022-05-19
python: fix confusing error message from pid URI resolver

grondo:
This PR fixes a minor error in the pid URI resolver. An incorrect variable name was used in an error message, so a Python `NameError` exception was thrown with confusing error message `flux-uri: ERROR: name 'broker_pid' is not defined`, instead of a `ValueError` with the real error message `PID <pid> is a flux-broker and no child found`.


chu11 (approved):
LGTM!

grondo:
Thanks! Setting MWP.

PR #4336, opened 2022-05-19
job-list: store inactive job data in job database

chu11:
Per discussion in #4273, some notes on what this does:

- add an sqlite database that stores all inactive data to it.

- db table is:

```
const char *sql_create_table = "CREATE TABLE if not exists jobs("
                               "  id CHAR(16) PRIMARY KEY,"
                               "  t_inactive REAL,"
                               "  jobdata JSON,"
                               "  eventlog TEXT,"
                               "  jobspec JSON,"
                               "  R JSON"
```

Where `jobdata` is a json blob containing all job data (the `all` `job-list` attribute from #4324, but added a `states_mask` which is used internally)

The main reason we have a `t_inactive` column is b/c this is backwards compatible to sqlite versions that don't have json support.

- database archiving is only enabled if a dbpath is specified in a `job-list` toml config or if `statedir` is configured.  So in other words, if no db is configured, it behaves just like before.

Ideas:

- should the `jobdata` have a version embedded?  if things like job states bitmask or things like that change?

### DIFF
--- a/src/modules/job-list/Makefile.am
+++ b/src/modules/job-list/Makefile.am
@@ -10,7 +10,8 @@ AM_CPPFLAGS = \
 	-I$(top_srcdir)/src/include \
 	-I$(top_builddir)/src/common/libflux \
 	$(FLUX_SECURITY_CFLAGS) \
-	$(JANSSON_CFLAGS)
+	$(JANSSON_CFLAGS) \
+	$(SQLITE_CFLAGS)
 
 fluxmod_LTLIBRARIES = job-list.la
 
@@ -21,6 +22,8 @@ job_list_la_SOURCES = \
 	job_state.c \
 	job_data.h \
 	job_data.c \
+	job_db.h \
+	job_db.c \
 	list.h \
 	list.c \
 	job_util.h \
@@ -28,7 +31,9 @@ job_list_la_SOURCES = \
 	idsync.h \
 	idsync.c \
 	stats.h \
-	stats.c
+	stats.c \
+	util.h \
+	util.c
 
 job_list_la_LDFLAGS = $(fluxmod_ldflags) -module
 job_list_la_LIBADD = \
@@ -38,4 +43,5 @@ job_list_la_LIBADD = \
 	$(top_builddir)/src/common/libflux-optparse.la \
 	$(top_builddir)/src/common/librlist/librlist.la \
 	$(JANSSON_LIBS) \
-	$(HWLOC_LIBS)
+	$(HWLOC_LIBS) \
+	$(SQLITE_LIBS)
--- a/src/modules/job-list/Makefile.am
+++ b/src/modules/job-list/Makefile.am
@@ -19,6 +19,8 @@ job_list_la_SOURCES = \
 	job-list.h \
 	job_state.h \
 	job_state.c \
+	job_data.h \
+	job_data.c \
 	list.h \
 	list.c \
 	job_util.h \
--- a/src/modules/job-list/job-list.c
+++ b/src/modules/job-list/job-list.c
@@ -17,6 +17,7 @@
 
 #include "job-list.h"
 #include "job_state.h"
+#include "job_data.h"
 #include "list.h"
 #include "idsync.h"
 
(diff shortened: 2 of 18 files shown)


cmoussa1:
I'm sorry if this isn't related to this PR but I figured I would double check. Is the plan to also adjust the job-archive DB to match the schema you have proposed in the PR description? I only ask because I currently use this query to fetch data from the job-archive DB:

```sql
SELECT userid,id,t_submit,t_run,t_inactive,ranks,R,jobspec FROM jobs
```

If so, that's totally fine - I will just need to tweak the query slightly to get the data I need.

chu11:
> I'm sorry if this isn't related to this PR but I figured I would double check. Is the plan to also adjust the job-archive DB to match the schema you have proposed in the PR description?

The plan is that we would leave the current `job-archive` the way it is until `flux-accounting` can adjust to the new database, everything gets deployed, etc.  then we can eventually retire `job-archive`.

cmoussa1:
Sounds good! Thanks for clarifying.

chu11:
oh crap, i accidentally did `git push origin :issue4273_job_db`, thus deleting remotely deleting the branch :P

chu11:
just re-based and and re-pushed fixing up merge conflicts due to recent changes

- adjust for changes for `job-list.list-inactive` removal from #4513
- adjust tests for `flux job list` needing to be on tty from #4499
- added some additional tests for `flux job list --since`

chu11:
rebased - fixing conflicts from #4542

chu11:
repushed, rebased on PR #4575

chu11:
rebased & re-pushed given #4575 going in

chu11:
rebased / refactored / cleanup based on recent cleanup PRs in job-list (#4639, #4644)

chu11:
Recently the topic of storing inactive and possibly purged jobs in a database and then being able to retrieve them later on via `flux-jobs` was brought up.  Some of that work was done in this PR.  This PR was initially submitted nearly a year ago, so I think it was just forgotten.  So hopefully this message restarts conversation on it :-)

Admittedly needs to be re-looked at given possible new needs / requirements / changes in flux-core / etc.

See additional conversation given #4914

chu11:
making this WIP, just wanted to point out a few points mostly from #4914

- an updated version of this PR this should go in after the constraint support and some additional features are added from #4914, would hate to have a sqlite database table that we need to re-do

- there is concern of creating a database table that is ultimately one we don't like and have to re-do several times.  I do see two longer out solutions:
  - use a nosql database, but one that supports sql queries than can be converted from jsonlogic
  - use sqlite with json support, but dumping all data into one json blob in a field.  By doing this we sort of get a "sql table doesn't matter" effect (i.e. if there is only 1 column, we don't care about making a bad table), although the devil is in the details until prototyping is done
    - bad part, json support in sqlite isn't supported in most distro's versions of sqlite right now

- several of the database points I mention above are why I decided adding "constraints" support to `job-list` would be a better "medium term" solution than going with a everything in `job-list` in a database approach.  right choice?  we will see.

- Sooo I'm going to keep this PR open for the time being, make it WIP, could be someone we still want to merge medium term if some we just deem the pros outweigh the cons.

chu11:
This job-list w/ a database has been re-worked:

- the DB now has a bunch of data stored in various columns, this is so ...

- constraints are now mapped into SQL queries using these new columns (e.g. `WHERE ((userid = 42) AND (queue = 'pbatch'))`)

  - per discussion in https://github.com/flux-framework/flux-core/pull/5847 we could have the SQL query use JSON and dig into the jobspec, but that requires us to use a localized copy of sqlite since the JSON support isn't in many distributed versions of sqlite.  This could be an alternate implementation path to take.  Either way we need a `constraint2sql()` function, so I just prototyped a non-JSON one to start.

- I should add some stress testing of sorts where the db holds like thousands of jobs from different users, to see if some performance impact might exist.  Although if we assume users won't increase the max number of returned jobs to (lets say) INT_MAX, maybe this isn't a huge deal.

  - We could also limit things in other ways, like guest users cannot access user != themselves in the older archived jobs (only in memory ones, i.e. jobs that are pending) ... although now that I think about it, perhaps a user should not be able to see != any inactive job.

garlick:
Thanks for the reminder of this PR.  We should probably give this one a look before exploring other ideas since it's fully implemented!

Issue #4338, opened 2022-05-23
rc3 hang after startup with content-sqlite corruption


garlick:
Problem: test system instance hung during startup because `content-sqlite` encountered a corrupt database, then rc3 hung running `flux content flush`.
```
2022-05-23T15:10:34.666532Z content-sqlite.debug[0]: /var/lib/flux/content.sqlite (556 objects) journal_mode=WAL synchronous=NORMAL
2022-05-23T15:10:34.666805Z broker.debug[0]: content backing store: enabled content-sqlite
2022-05-23T15:10:34.743113Z broker.debug[0]: insmod kvs
2022-05-23T15:10:34.744392Z content-sqlite.err[0]: store: executing stmt: database disk image is malformed(11)
2022-05-23T15:10:34.744725Z broker.err[0]: content store: Invalid argument
2022-05-23T15:10:34.745462Z kvs.info[0]: restored KVS from checkpoint on 2022-05-23T00:19:48Z
2022-05-23T15:10:34.830264Z broker.debug[0]: insmod kvs-watch
2022-05-23T15:10:34.845895Z kvs.err[0]: content_load_completion: content_load_get: No such file or directory
2022-05-23T15:10:34.846620Z broker.err[0]: rc1.0: flux: Error commiting start event: No such file or directory
2022-05-23T15:10:34.848588Z broker.err[0]: rc1.0: /usr/local/etc/flux/rc1 Exited (rc=1) 0.7s
2022-05-23T15:10:34.848993Z broker.info[0]: rc1-fail: init->shutdown 0.700172s
2022-05-23T15:10:34.849191Z broker.info[0]: children-none: shutdown->finalize 0.162685ms
2022-05-23T15:10:35.025620Z broker.debug[0]: rmmod barrier
2022-05-23T15:10:35.025985Z broker.debug[0]: module barrier exited
2022-05-23T15:10:35.038375Z kvs.err[0]: content_load_completion: content_load_get: No such file or directory
2022-05-23T15:10:35.039105Z broker.err[0]: rc3.0: flux: Error commiting finish event: No such file or directory
2022-05-23T15:10:35.053194Z broker.debug[0]: rmmod kvs-watch
2022-05-23T15:10:35.053527Z broker.debug[0]: module kvs-watch exited
2022-05-23T15:10:35.067434Z broker.debug[0]: rmmod kvs
2022-05-23T15:10:35.077639Z broker.debug[0]: module kvs exited
```
```console
[...]

PR #4340, opened 2022-05-24
content-sqlite: verify database integrity during module load

garlick:
Problem: if the content-sqlite database is corrupted, this is typically discovered at runtime when the KVS attempts to load/store a blob, which causes secondary fallout.  It would be better to catch that condition early.

Add the [quick_check](https://www.sqlite.org/pragma.html#pragma_quick_check) sqlite pragma.  Since pragmas are executed before the module enters the reactor, if this check fails, the module load fails, and rc1 aborts before the higher level services are loaded.

This addresses #4338

### DIFF
--- a/src/modules/content-sqlite/content-sqlite.c
+++ b/src/modules/content-sqlite/content-sqlite.c
@@ -641,6 +641,14 @@ static int content_sqlite_opendb (struct content_sqlite *ctx, bool truncate)
         goto error;
     }
     if (sqlite3_exec (ctx->db,
+                      "PRAGMA quick_check",
+                      NULL,
+                      NULL,
+                      NULL) != SQLITE_OK) {
+        log_sqlite_error (ctx, "setting sqlite 'quick_check' pragma");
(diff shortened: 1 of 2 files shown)


grondo (approved):
LGTM!

garlick:
Thanks, I tacked on a bit of testing to try to improve the numbers.

chu11:
Would it be worthwhile to copy the `quick_check` over to `job-archive`?  I will copy over to PR #4336 of course.

chu11:
> Be aware the overhead (when the pragma is evaluated during initialization) is proprtional to the number of rows in the database.

ahhh, which there would no cleanup in the job-archive dbs long term.  I guess its ok to leave this out for the time being in `job-archive` and `job-db`.

garlick:
I couldn't quite hit the code path I wanted by selectively mangling the database file, although I did cover another uncovered path.  Oh well, let's call this good.  Thanks for the review!  Setting MWP.

Issue #4341, opened 2022-05-24
system instance cert contains hostname that created key, confusing the logs


garlick:
Problem: each time a peer authenticates to its TBON parent in a system instance, the hostname that created the shared certificate appears in the auth log message:

For example, picl2 is authenticating  to picl0, but the auth message says:
```
2022-05-24T19:37:04.755724Z broker.info[0]: overlay auth picl0 OK
2022-05-24T19:37:04.759009Z broker.info[0]: accepting connection from picl2 (rank 2) uuid 39033208-1e3d-4df8-bf07-3755ee81afd1 status full
```

To make this slightly less confusing:
- change the log message to be more descriptive, e.g. `overlay auth cert-name=NAME OK`
- provide a `flux-keygen` option to set the cert name to something other than the hostname when the cert is generated (such as cluster name)
- update the admin guide to suggest that this option be used (or make it mandatory)

PR #4342, opened 2022-05-24
improve logging of overlay peer authentication

garlick:
This works on the problem described in #4341, where logs that look like this:
```
2022-05-24T22:40:04.389347Z broker.info[0]: overlay auth picl0 OK
```
are potentially misleading, since `picl0` is the hostname where the cert was generated not the peer that is authenticating.

We change the log slightly so it at least indicates what that name is:
```
2022-05-24T22:40:04.389347Z broker.info[0]: overlay auth cert-name=picl0 OK
```
and then add a `flux keygen --name=NAME` option that lets the certificate name be set to something other than the local hostname.  I was thinking we would recommend the cluster name in the admin guide.

### DIFF
--- a/doc/man1/flux-keygen.rst
+++ b/doc/man1/flux-keygen.rst
@@ -6,7 +6,7 @@ flux-keygen(1)
 SYNOPSIS
 ========
 
-**flux** **keygen** *PATH*
+**flux** **keygen** [*--name=NAME*] *PATH*
 
 
 DESCRIPTION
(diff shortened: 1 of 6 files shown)


grondo (approved):
This seems reasonable to me! Seeing those messages I did always kind of assume it was logging the name of the authenticating host.

A little late, but would a simplification be to only log the cert name if the auth fails? O/w, the cert-name is possibly superfluous since it worked we know it is the right one?

This is all minor though, the PR here is definitely a good improvement!

garlick:
Thanks - well, I kind of hate to remove those log messages since they are showing that our authentication service is actually being used by libzmq.  They are also fairly infrequent.  I can drop the log level to debug though.

grondo:
Sorry, I think the log messages should be kept, I was only questioning if the cert-name is necessary on success if it is causing confusion...

garlick:
Oh gotcha, that's a decent suggestion.  I still would feel slightly better printing all the info in the log message if it doesn't bother you too much.

grondo:
>  I still would feel slightly better printing all the info in the log message if it doesn't bother you too much.

Nope, it was just meant to be a suggestion!

PR #4345, opened 2022-05-25
libflux: return better errno in future wait path

### LINKED ISSUE #4343: libflux: return more clear error on uninitialized reactor in `flux_future_wait_for()`
This is similar to #4160.  I basically did:

```
        flux_future_t *f2 = flux_future_create (NULL, NULL);
        flux_future_set_flux (f2, h);
        if (flux_future_get (f2, NULL) < 0) {
            perror ("flux_future_get");
            exit (1);
        }
```

and you get: `flux_future_get: Invalid argument`.  The reason is in `flux_future_wait_for()` (called from `flux_future_get()`), if the timeout is < 0, a timeout watcher is never created.  So the now reactor has nothing on it, so when `flux_reactor_run()` is later executed, it just returns immediately.

Now that I've hit this twice, I'm thinking this is probably a corner case that needs to be corrected.

I see possible solutions.  An experiment putting an idle watcher into the now reactor generally speaking worked (did not test thoroughly).

```
+        if (timeout < 0.) {       // disable
            flux_watcher_stop (now->timer);
+            if (!now->idle) {
+                now->idle = flux_idle_watcher_create (now->r, NULL, NULL);
+                if (!now->idle)
+            }
+            flux_watcher_start (now->idle);
+        }
```

although I'm not 100% if having a reactor with only the idle watcher is ok?  Is the reactor basically just spinning at that point?

another option might be to set the timeout watcher to `DBL_MAX` timeout length, which would timeout in bajillions of years.

chu11:
Problem: In flux_future_wait_for() the errno EINVAL is returned
when a flux future is not initialized properly.  This errno is
generic and confusing.

Solution: Return EDEADLOCK instead.  Update comments to give
more details on how a hang / deadlock may occur.  Update unit
tests accordingly.

Fixes #4343

### DIFF
--- a/src/common/libflux/future.c
+++ b/src/common/libflux/future.c
@@ -481,11 +481,17 @@ int flux_future_wait_for (flux_future_t *f, double timeout)
         f->now->running = false;
     }
     if (!future_is_ready (f)) {
-        /* Typically this error should be "impossible", but can occur
-         * if the future does not get initialized properly to use the
-         * now reactor.
+        /* This block of code is reached b/c flux_reactor_run() on the
+         * "now" reactor returned >= 0 AND the future was
+         * not fulfilled, e.g.
+         * - no init callback was registered when future was created
+         * - the init callback didn't register any watchers
+         * - the registered watchers stopped themselves or called
+         *   flux_reactor_stop() without fulfilling the future.
+         * Return EDEADLOCK indicating that the future cannot be
+         * fulfilled.
          */
-        errno = EINVAL;
+        errno = EDEADLOCK;
         return -1;
     }
     return 0;
--- a/doc/man3/flux_future_get.rst
+++ b/doc/man3/flux_future_get.rst
@@ -129,6 +129,9 @@ ETIMEDOUT
    A timeout passed to ``flux_future_wait_for()`` expired before the future
    was fulfilled.
 
+EDEADLOCK
+   ``flux_future_wait_for()`` would likely deadlock due to an
+   improperly initialized future.
 
 RESOURCES
 =========
--- a/doc/test/spell.en.pws
+++ b/doc/test/spell.en.pws
@@ -626,3 +626,4 @@ num
 gc
 tgz
 tmpfiles
+EDEADLOCK
--- a/src/common/libflux/test/future.c
+++ b/src/common/libflux/test/future.c
@@ -122,8 +122,8 @@ void test_simple (void)
     ok (!flux_future_is_ready (f),
         "flux_future_is_ready returns false");
     errno = 0;
-    ok (flux_future_wait_for (f, -1.0) < 0 && errno == EINVAL,
-        "flux_future_wait_for fails with EINVAL with timeout < 0");
+    ok (flux_future_wait_for (f, -1.0) < 0 && errno == EDEADLOCK,
+        "flux_future_wait_for fails with EDEADLOCK with timeout < 0");
     ok (!flux_future_is_ready (f),
         "flux_future_is_ready returns false");
     errno = 0;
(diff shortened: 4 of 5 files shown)


src/common/libflux/future.c
```diff
@@ -481,11 +483,21 @@ int flux_future_wait_for (flux_future_t *f, double timeout)
...
+         *
+         * However, if the future does not get initialized properly
+         * (i.e. via flux_future_create()'s init callback) the reactor
+         * may not have any watchers on it and return immediately,
+         * thus the future is still not ready.
+         *
+         * This is also why an "infinite timer watcher" is not created
+         * for the reactor in now_context_set_timeout().  In the event
+         * the flux future does not get initialized properly, a
+         * "infinite timer watcher" would lead to a hang / deadlock.
```
grondo, line 498: The comment about an "infinite timer watcher" might be confusing here. At least I'm not sure what prompted this part of the comment and it would make me scratch my head. I would consider just dropping this section or replacing with something like "If the reactor did block without any watchers registered in a "now" context, then this would be deadlock. Therefore we return EDEADLOCK here to warn the user".

grondo (commented):
Nice! One very minor comment inline.

It might be good to also update the `flux_future_get(3)` man page `ERRORS` section to explicitly document what `EDEADLOCK` means in this case.

chu11:
re-pushed with suggested fixes, please take a quick skim to see if the wording looks good

grondo (approved):
LGTM! Thanks! We may want @garlick to also weigh in on the wording.

chu11:
re-pushed, forgot to add EDEADLOCK to manpage allowlist of words

src/common/libflux/future.c
```diff
@@ -481,11 +483,21 @@ int flux_future_wait_for (flux_future_t *f, double timeout)
...
+         * However, if the future does not get initialized properly
+         * (i.e. via flux_future_create()'s init callback) the reactor
+         * may not have any watchers on it and return immediately,
+         * thus the future is still not ready.
+         *
+         * If the reactor blocked without any other watchers
+         * registered in the "now" context, then this would likely
+         * lead to a deadlock. Therefore we return EDEADLOCK here to
+         * warn the user.
          */
```
garlick, line 499: This comment is getting a little long.  May I suggest the following edit:
```c
/* This block of code is executed if the future was not fulfilled AND
 * flux_reactor_run() on the "now" reactor returned 0 immediately, e.g.
 * - no init callback was registered
 * - the init callback didn't register any watchers
 * - the registered watchers stopped themselves or called flux_reactor_stop()
 *   without fulfilling the future.
 * Return EDEADLOCK indicating that the future cannot be fulfilled.
 */
```

src/common/libflux/future.c
```diff
@@ -108,6 +108,8 @@ static int now_context_set_timeout (struct now_context *now,
                                     double timeout, void *arg)
 {
     if (now) {
+        /* Why not create a timer with timeout set to DBL_MAX?  See
+         * deadlock error notes in flux_future_wait_for() */
```
garlick, line 112: This comment seems like it's warding off something that would not be intuitive to do.

However, if you think it helps, then I'm fine with leaving it.

garlick (approved):
LGTM!   I just had comments about comments - you can take them or leave them.

src/common/libflux/future.c
chu11: yeah, i was debating if its better or worse without it.  i'll remove it since you mentioned it :-)

chu11:
re-pushed given @garlick comments above.  Did a little bit of a tweak to what Jim suggested, if you could please give it a quick skim.

garlick:
Looks good - thanks!

PR #4346, opened 2022-05-31
job-manager: post submit event, instead of job-ingest

garlick:
Problem: job-ingest posts the submit event, but every other job
    event is posted by the job manager. This results in some special
    cases in the job manager that can be confusing.
    
Since job-ingest now waits for the job manager to "approve" each
    job before responding to the user, this division of labor is no
    longer required.
    
Have the job manager post the submit event.  Maintain the invariant
    that the user does not receive the job ID until the submit event
    has been committed to the KVS by deferring the response to job-ingest
    until the KVS batch completes.
    
Impact on submit latency:  latency is increased because job-ingest does
    not send its request to the job manager until the jobspec is committed
    to the KVS, and now the job manager does not respond until the submit
    event is committed to the KVS.  An individual submit must now wait for
    two back to back KVS commits instead of one.
    
Impact on job throughput:  any impact on job throughput should be
    mitigated by KVS commit batching.  For example, the submit event is
    likely now batched with the depend event that usually follows it,
    whereas before it was batched with the jobspec.

This PR includes a little bit of ancillary cleanup as well.

### DIFF
--- a/src/modules/job-ingest/job-ingest.c
+++ b/src/modules/job-ingest/job-ingest.c
@@ -232,7 +232,8 @@ static void batch_response_destroy (struct batch_response *bresp)
 static void *jobid_duplicator (const void *item)
 {
     flux_jobid_t *id = calloc (1, sizeof (flux_jobid_t));
-    *id = *((flux_jobid_t *)item);
+    if (id)
+        *id = *((flux_jobid_t *)item);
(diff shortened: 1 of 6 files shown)


garlick:
Thanks, I'll go ahead and set MWP.

Issue #4308, opened 2022-05-02
kvs: add FLUX_KVS_SYNC or similar


garlick:
What would be really nice, but possibly a bit of work, would be to add a new KVS flag that could be added to `flux_kvs_commit()` so that once the commit finishes, we are guaranteed the result is on the backing store.   Then we could judiciously use this in places where "durability" of the commit is critical.   It would mean adding the above sequence to the KVS commit logic for the primary namespace only.  If @chu11 thinks this might be feasible, perhaps we could open a separate issue on that and let this one be closed with something simple like the cron job.

_Originally posted by @garlick in https://github.com/flux-framework/flux-core/issues/4301#issuecomment-1108925864_

Edit: "above sequence" references:

```
    calls flux_kvs_getroot()
    makes the content.flush RPC on rank 0
    calls kvs_checkpoint_commit()
```

chu11:
as a side note, this should probably override #3330, i'll close that one

garlick:
Oops sorry I didn't catch that.

chu11:
per a @garlick comment in #4343, should this be done inside the kvs module?  or should it / can it be done in userspace in the `flux_kvs_commit()` function?

I initially thought it should be in the module, but upon realizing that there were convenient userspace calls for everything, I thought it worthwhile to just put in userspace in `flux_kvs_commit()`.

garlick:
I _think_ there is an atomicity issue if this is done on the client side.  E.g. threads A and B both commit with the sync flag
1. thread A commit completes
2. thread B commit completes
3. thread B checkpoint completes (rootref from commit B)
4. thread A checkpoint completes (rootref from commit A)
5. crash

thread B's data is lost

PR #4348, opened 2022-06-01
kvs: support FLUX_KVS_SYNC

chu11:
Per discussion in #4308.

Thought I'd park this to get any initial thoughts.  I have some very simple tests but could use more, perhaps after I see how coverage is after this PR starts.

- Basically added several states to the "kvs transaction" state machine to do the `content.flush`  and `checkpoint`.  I don't think the code change is shocking.  Perhaps the most interesting refactoring was to change the state machine's "switch" statement into a loop, b/c doing a "fallthrough" between states no longer makes sense with the new states.  Right now it sort of looks weird that the primary state machine is a `while (1)` loop and exits are all based on `break`, `goto`, and `return`.

- Only works with primary namespace, but could work with another namespaces.  We'd just have to come up with a consistent key naming scheme for the checkpoints of other namespaces?
  - OR `flux_kvs_commit()` could be altered to take a key for the checkpoint location?

- Only works with `flux_kvs_commit()` and not `flux_kvs_fence()` at the moment.  Dunno if we need to do both?  In some ways doesn't make sense for fence?  But in some ways might as well add it?

### DIFF
--- a/src/common/libkvs/kvs_commit.h
+++ b/src/common/libkvs/kvs_commit.h
@@ -24,10 +24,23 @@ extern "C" {
  * operations will be illegal to compact and result in an error.  Most
  * notably, if a key has data appended to it, then is overwritten in
  * the same transaction, a compaction of appends is not possible.
+ *
+ * FLUX_KVS_SYNC will ensure all data flushed to the backing store and
+ * the root reference is checkpointed.  It effectively performs a:
+ *
+ * content.flush on rank 0
+ * checkpoint on the new root reference from the commit
+ *
+ * FLUX_KVS_SYNC only works against the primary KVS namespace.  If any
+ * part of the content.flush or checkpoint fails an error will be
+ * returned and the entire commit will fail.  For example, if a
+ * content backing store is not loaded, ENOSYS will returned from this
+ * commit.
  */
 enum kvs_commit_flags {
     FLUX_KVS_NO_MERGE = 1, /* disallow commits to be mergeable with others */
     FLUX_KVS_TXN_COMPACT = 2, /* try to combine ops on same key within txn */
+    FLUX_KVS_SYNC = 4, /* flush and checkpoint after commit is done */
 };
 
 flux_future_t *flux_kvs_commit (flux_t *h, const char *ns, int flags,
--- a/src/modules/kvs/kvs.c
+++ b/src/modules/kvs/kvs.c
@@ -88,6 +88,7 @@ static void transaction_prep_cb (flux_reactor_t *r, flux_watcher_t *w,
 static void transaction_check_cb (flux_reactor_t *r, flux_watcher_t *w,
                                   int revents, void *arg);
 static void start_root_remove (struct kvs_ctx *ctx, const char *ns);
+static void kvstxn_apply (kvstxn_t *kt);
 
 /*
  * kvs_ctx functions
@@ -895,6 +896,12 @@ static void work_queue_check_append (struct kvs_ctx *ctx,
         work_queue_append (ctx, root);
 }
 
+static void kvstxn_apply_cb (flux_future_t *f, void *arg)
+{
+    kvstxn_t *kt = arg;
+    kvstxn_apply (kt);
+}
+
 /* Write all the ops for a particular commit/fence request (rank 0
  * only).  The setroot event will cause responses to be sent to the
  * transaction requests and clean up the treq_t state.  This
@@ -997,6 +1004,34 @@ static void kvstxn_apply (kvstxn_t *kt)
         }
 
         assert (wait_get_usecount (wait) > 0);
+        goto stall;
+    }
+    else if (ret == KVSTXN_PROCESS_SYNC_CONTENT_FLUSH) {
+        /* N.B. futre is managed by kvstxn, should not call
+         * flux_future_destroy() on it */
+        flux_future_t *f = kvstxn_sync_content_flush (kt);
+        if (!f) {
+            errnum = errno;
+            goto done;
+        }
+        if (flux_future_then (f, -1., kvstxn_apply_cb, kt) < 0) {
+            errnum = errno;
+            goto done;
+        }
+        goto stall;
+    }
+    else if (ret == KVSTXN_PROCESS_SYNC_CHECKPOINT) {
+        /* N.B. futre is managed by kvstxn, should not call
+         * flux_future_destroy() on it */
+        flux_future_t *f = kvstxn_sync_checkpoint (kt);
+        if (!f) {
+            errnum = errno;
+            goto done;
+        }
+        if (flux_future_then (f, -1., kvstxn_apply_cb, kt) < 0) {
+            errnum = errno;
+            goto done;
+        }
         goto stall;
     }
     /* else ret == KVSTXN_PROCESS_FINISHED */
--- a/src/modules/kvs/kvs.c
+++ b/src/modules/kvs/kvs.c
@@ -113,6 +113,7 @@ static struct kvs_ctx *kvs_ctx_create (flux_t *h)
 
     if (!(ctx = calloc (1, sizeof (*ctx))))
         return NULL;
+    ctx->h = h;
     if (!(ctx->hash_name = flux_attr_get (h, "content.hash"))) {
         flux_log_error (h, "getattr content.hash");
         goto error;
@@ -121,8 +122,7 @@ static struct kvs_ctx *kvs_ctx_create (flux_t *h)
         goto error;
     if (!(ctx->krm = kvsroot_mgr_create (ctx->h, ctx)))
         goto error;
-    ctx->h = h;
-    if (flux_get_rank (h, &ctx->rank) < 0)
+    if (flux_get_rank (ctx->h, &ctx->rank) < 0)
         goto error;
     if (ctx->rank == 0) {
         ctx->prep_w = flux_prepare_watcher_create (r, transaction_prep_cb, ctx);
--- a/src/modules/kvs/kvs.c
+++ b/src/modules/kvs/kvs.c
@@ -44,12 +44,13 @@
 #include "kvsroot.h"
 #include "kvssync.h"
 
-/* sync_cb() is called periodically to manage cached content and namespaces.
- * Synchronize with the system heartbeat if possible, but keep the time between
- * checks bounded by 'sync_min' and 'sync_max' seconds.
+/* heartbeat_sync_cb() is called periodically to manage cached content
+ * and namespaces.  Synchronize with the system heartbeat if possible,
+ * but keep the time between checks bounded by 'heartbeat_sync_min'
+ * and 'heartbeat_sync_max' seconds.
  */
-const double sync_min = 1.;
-const double sync_max = 30.;
+const double heartbeat_sync_min = 1.;
+const double heartbeat_sync_max = 30.;
 
 /* Expire cache_entry after 'max_lastuse_age' seconds.
  */
@@ -1222,7 +1223,7 @@ static int heartbeat_root_cb (struct kvsroot *root, void *arg)
     return 0;
 }
 
-static void sync_cb (flux_future_t *f, void *arg)
+static void heartbeat_sync_cb (flux_future_t *f, void *arg)
 {
     struct kvs_ctx *ctx = arg;
 
@@ -2874,7 +2875,7 @@ int mod_main (flux_t *h, int argc, char **argv)
 {
     struct kvs_ctx *ctx;
     flux_msg_handler_t **handlers = NULL;
-    flux_future_t *f_sync = NULL;
+    flux_future_t *f_heartbeat_sync = NULL;
     int rc = -1;
 
     if (!(ctx = kvs_ctx_create (h))) {
@@ -2929,8 +2930,11 @@ int mod_main (flux_t *h, int argc, char **argv)
         flux_log_error (h, "flux_msg_handler_addvec");
         goto done;
     }
-    if (!(f_sync = flux_sync_create (h, sync_min))
-            || flux_future_then (f_sync, sync_max, sync_cb, ctx) < 0) {
+    if (!(f_heartbeat_sync = flux_sync_create (h, heartbeat_sync_min))
+            || flux_future_then (f_heartbeat_sync,
+                                 heartbeat_sync_max,
+                                 heartbeat_sync_cb,
+                                 ctx) < 0) {
         flux_log_error (h, "error starting heartbeat synchronization");
         goto done;
     }
@@ -2958,7 +2962,7 @@ int mod_main (flux_t *h, int argc, char **argv)
     }
     rc = 0;
 done:
-    flux_future_destroy (f_sync);
+    flux_future_destroy (f_heartbeat_sync);
     flux_msg_handler_delvec (handlers);
     kvs_ctx_destroy (ctx);
     return rc;
--- a/src/modules/kvs/kvstxn.c
+++ b/src/modules/kvs/kvstxn.c
@@ -63,6 +63,24 @@ struct kvstxn {
     zlist_t *dirty_cache_entries_list;
     int internal_flags;
     kvstxn_mgr_t *ktm;
+    /* State transitions
+     *
+     * INIT - perform initializations / checks
+     * LOAD_ROOT - load KVS root
+     *           - if needed, report missing refs to caller and stall
+     * APPLY_OPS - apply changes to KVS
+     *           - if needed, report missing refs to caller and stall
+     * STORE - generate dirty entries for caller to store
+     * PRE_FINISHED - stall until stores complete
+     *              - generate keys modified in txn
+     * FINISHED - end state
+     *
+     * INIT -> LOAD_ROOT
+     * LOAD_ROOT -> APPLY_OPS
+     * APPLY_OPS -> STORE
+     * STORE -> PRE_FINISHED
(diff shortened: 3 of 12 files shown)


garlick:
Cool that you got this done so quickly @chu11!  I haven't had a chance to review yet but wanted to respond to some questions
- If it's not a huge pain, I do think FLUX_KVS_SYNC should work with fence.  That's really just a variant of a commit, and it's a little bit weird for it not to accept the same flags
- I think in the absence of a strategy for checkpointing namespaces other than the primary, a commit that specifies this flag should fail.

src/common/libkvs/kvs_commit.c
```diff
@@ -61,12 +61,23 @@ flux_future_t *flux_kvs_fence (flux_t *h, const char *ns, int flags,
             return NULL;
     }
 
+    if (flags & FLUX_KVS_SYNC
+        && strcmp (ns, KVS_PRIMARY_NAMESPACE) != 0) {
+        errno = EINVAL;
+        return NULL;
+    }
+
```
garlick, line 69: This check should be in the kvs module, not in the client, in case the RPCs are made directly, for example by a language binding that cannot readily wrap Flux futures.

src/common/libkvs/kvs_commit.c
```diff
@@ -61,12 +61,23 @@ flux_future_t *flux_kvs_fence (flux_t *h, const char *ns, int flags,
...
         if (txn_compact (txn) < 0)
             return NULL;
         flags &= ~FLUX_KVS_TXN_COMPACT;
     }
 
+    /* FLUX_KVS_SYNC implies FLUX_KVS_NO_MERGE, as we will checkpoint
+     * after this specific commit completes */
+    if (flags & FLUX_KVS_SYNC)
+        flags |= FLUX_KVS_NO_MERGE;
+
```
garlick, line 80: Should be in the kvs module also.

src/common/libkvs/kvs_commit.c
```diff
@@ -113,12 +124,23 @@ flux_future_t *flux_kvs_commit (flux_t *h, const char *ns, int flags,
             return NULL;
     }
 
+    if (flags & FLUX_KVS_SYNC
+        && strcmp (ns, KVS_PRIMARY_NAMESPACE) != 0) {
+        errno = EINVAL;
+        return NULL;
+    }
```
garlick, line 131: move to kvs module

src/common/libkvs/kvs_commit.c
```diff
@@ -113,12 +124,23 @@ flux_future_t *flux_kvs_commit (flux_t *h, const char *ns, int flags,
...
         if (txn_compact (txn) < 0)
             return NULL;
         flags &= ~FLUX_KVS_TXN_COMPACT;
     }
 
+    /* FLUX_KVS_SYNC implies FLUX_KVS_NO_MERGE, as we will checkpoint
+     * after this specific commit completes */
+    if (flags & FLUX_KVS_SYNC)
+        flags |= FLUX_KVS_NO_MERGE;
+
```
garlick, line 143: move to kvs module

src/modules/kvs/kvs.c
```diff
@@ -999,6 +1006,30 @@ static void kvstxn_apply (kvstxn_t *kt)
...
+        if (!f) {
+            errnum = errno;
+            goto done;
+        }
+        if (flux_future_then (f, -1., kvstxn_apply_cb, kt) < 0) {
+            errnum = errno;
+            goto done;
+        }
+        goto stall;
+    }
```
garlick, line 1032: A comment that the future _f_ is managed in kvstxn.c and need not be freed here might be helpful.

src/common/libkvs/kvs_commit.c
```diff
@@ -163,6 +163,15 @@ static int decode_response (flux_future_t *f, const char **rootrefp,
...
 
+int flux_kvs_commit_get_rootref (flux_future_t *f, const char **rootref)
+{
+    if (!f || !rootref) {
+        errno = EINVAL;
+        return -1;
+    }
+    return decode_response (f, rootref, NULL);
+}
+
```
garlick, line 196: Correct me if I'm wrong but this appears to not be used anymore by the sync implementation in this PR.  It's still handy though I guess - split to another PR?

garlick (commented):
Nice work!  I've put a few initial comments below.

One question: what happens if a commit is requested with the SYNC flag while another sync is in progress?  I wasn't too clear on whether the entire commit "pipeline" must complete before another commit can start.  I see the kvstxn stuff caches one future each for the flush and the sync RPCs, not a list of them.

garlick:
Another comment is that if kvs commits with the sync flag are serialized internally in the kvs module, no other flux components should be making checkpoint calls..  I see two users:  `flux-startlog(1)` and `flux-restore(1)`.  It looks like the former can trivially drop some code and just add the new FLUX_KVS_SYNC flag.  The latter is a special case since it insists that the KVS module not be loaded when it writes the checkpoint so it should be fine.

We should probably add a comment to `kvs_checkpoint.h` like
```c
/* It is racy to call kvs_checkpoint_commit() when the KVS module is loaded.
 * In most cases, use the FLUX_KVS_SYNC flag with flux_kvs_commit()
 * to ensure committed data survives a Flux restart.
 */
```

chu11:
> One question: what happens if a commit is requested with the SYNC flag while another sync is in progress? I wasn't too clear on whether the entire commit "pipeline" must complete before another commit can start. I see the kvstxn stuff caches one future each for the flush and the sync RPCs, not a list of them.

For each namespace, all commits eventually get serialized in the kvs module, so it shouldn't matter if multiple SYNC at the same time.  It just will slow down the pipeline / throughput of all commits.

garlick:
What is the behavior of SYNC if a backing store is not loaded?

src/common/libkvs/kvs_commit.c
chu11: yeah, was thinking of splitting it off anyways, so i'll do that right now!

chu11:
> What is the behavior of SYNC if a backing store is not loaded?

Ahhh good question.  It should just error out b/c the RPC to the backing store should fail, and since `setroot()` hasn't been called yet, no change to the kvsroot of the namespace.  But this is a good test to add.

And as I think about this more, some comments should probably be added to make it clear that after a successful checkpoint, we need to just finish and call `setroot()`, nothing can be done in between that could potentially cause an error (i.e. don't want our checkpoint to point to a rootref that never existed).

src/modules/kvs/kvs.c
chu11: probably should return const for the future too!

chu11: oh doh!  lots of funcs don't expect const futures.  comment is probably enough then.

chu11:
>  What is the behavior of SYNC if a backing store is not loaded?

Ahh it ends up that it hangs.  The reason is that `content.flush` will simply not respond until the flush completes, like hypothetically it returns if a content backing store module is loaded later.   This feels wrong to me.  Think `content.flush` should return an error when a backing store isn't around?

chu11:
re-pushed

- due to bug #4375, this is now built on top of PR #4376
- fixed up per comments above
  - extra comments in `kvs_checkpoint.h` and added comments about calling `setroot()` right after a successful checkpoint.
  - checks previously done in `libkvs` now done in module
    - side note, instead of setting FLUX_KVS_NO_MERGE in the flags when FLUX_KVS_SYNC is set, when merging is done in the kvs module, just check for FLUX_KVS_NO_MERGE OR FLUX_KVS_SYNC, i.e. treating them equally to each other.

chu11:
re-pushed, rebased on #4376

So for those following at home, PR order should be

#4380
#4376 
#4348 (this one)

chu11:
ugh, I was just reminded that we have `libflux/sync.h`

```
/* Synchronize future to the system heartbeat.
 * Set minimum > 0. to establish a minimum time between fulfillments.
 * Use a continuation timeout to establish a maximum time between fulfillments.
 */
flux_future_t *flux_sync_create (flux_t *h, double minimum);
```

so lots of "sync" stuff already exists, including in the KVS (including the "sync_cb")

Perhaps should we rename to `FLUX_KVS_CHECKPOINT`?  Lemme ponder a better name.

chu11:
> Perhaps should we rename to FLUX_KVS_CHECKPOINT? Lemme ponder a better name.

Hmmm, `sync` just seems to make sense b/c I tend to think of `sync(3)`   Should `flux_sync_create()` perhaps be renamed?  Perhaps `flux_heartbeat_sync_create()`?  That might make it a bit more clear and differentiated from this PRs "sync".

grondo:
My opinion shouldn't matter much here, but I don't find it likely that the KVS sync and heartbeat sync would be confused. I mean the KVS sync related flag is `FLUX_KVS_SYNC` right not just `FLUX_SYNC`? It makes sense that the term sync might be used differently in different contexts. (Though I wouldn't be opposed to renaming `flux_sync_create(3)` to `flux_heartbeat_sync_create(3)`)

chu11:
> My opinion shouldn't matter much here, but I don't find it likely that the KVS sync and heartbeat sync would be confused. I mean the KVS sync related flag is FLUX_KVS_SYNC right not just FLUX_SYNC? It makes sense that the term sync might be used differently in different contexts.

Perhaps you're right.  It is quite confusing in the KVS b/c we already have `f_sync` and `sync_cb()` and things like that that are related to the heartbeat sync, making me think a name change was warranted.  But perhaps just renaming all that stuff in the KVS should be enough.

chu11:
re-pushed, I just added 1 commit into the series to rename "sync" variables / funcs to "heartbeat_sync" in the kvs module, to avoid future confusion.

garlick:
> I don't find it likely that the KVS sync and heartbeat sync would be confused

Yeah that doesn't worry me either, as far as the public API goes.  It seems like good cleanup to rename internal kvs stuff to avoid confusion though.

src/modules/kvs/kvstxn.c
```diff
@@ -63,6 +63,14 @@ struct kvstxn {
...
     int internal_flags;
     kvstxn_mgr_t *ktm;
+    /* State transitions
+     *
+     * INIT -> LOAD_ROOT
+     * LOAD_ROOT -> APPLY_OPS
+     * APPLY_OPS -> STORE
+     * STORE -> PRE_FINISHED
+     * PRE_FINISHED -> FINISHED
+     */
```
garlick, line 80: Since the state machine is _linear_ how about just:
```
INIT -> LOAD_ROOT -> APPLY_OPS -> STORE -> PRE_FINISHED -> FINISHED
```
Would it be useful to have a sentence or two describing each state also?

src/modules/kvs/kvstxn.c
```diff
@@ -68,15 +68,15 @@ struct kvstxn {
      * INIT -> LOAD_ROOT
      * LOAD_ROOT -> APPLY_OPS
      * APPLY_OPS -> STORE
-     * STORE -> PRE_FINISHED
-     * PRE_FINISHED -> FINISHED
+     * STORE -> POST_STORE
```
garlick, line 75: Is there a better name that would describe what's actually happening in the state rather than just "the state that comes after store"?

src/common/libkvs/kvs_commit.h
```diff
@@ -24,10 +24,19 @@ extern "C" {
...
  * notably, if a key has data appended to it, then is overwritten in
  * the same transaction, a compaction of appends is not possible.
+ *
+ * FLUX_KVS_SYNC will ensure all data flushed to the backing store and
+ * the root reference is checkpointed.  It effectively performs a:
+ *
+ * content.flush on rank 0
+ * checkpoint on the new root reference from the commit
+ *
+ * FLUX_KVS_SYNC only works against the primary KVS namespace.
```
garlick, line 34: It may be good to mention that users can expect an ENOSYS error if FLUX_KVS_SYNC is used but no backing store is loaded, and that this leaves KVS content unchanged (the whole commit fails).

garlick (approved):
LGTM!  Just some minor comments mostly about comments.

src/modules/kvs/kvstxn.c
chu11: yeah, i can write a bit about each state.  I'll keep the separate on each line though, b/c it was with knowledge of non-linear in a future commit.

src/modules/kvs/kvstxn.c
chu11: good idea, lets go with "generate keys"

chu11:
re-pushed with fixes for @garlick comments above, just some extra comments and renaming POST_STORE to GENERATE_KEYS

```
diff --git a/src/common/libkvs/kvs_commit.h b/src/common/libkvs/kvs_commit.h
index 4548867..01cd957 100644
--- a/src/common/libkvs/kvs_commit.h
+++ b/src/common/libkvs/kvs_commit.h
@@ -31,7 +31,11 @@ extern "C" {
  * content.flush on rank 0
  * checkpoint on the new root reference from the commit
  *
- * FLUX_KVS_SYNC only works against the primary KVS namespace.
+ * FLUX_KVS_SYNC only works against the primary KVS namespace.  If any
+ * part of the content.flush or checkpoint fails an error will be
+ * returned and the entire commit will fail.  For example, if a
+ * content backing store is not loaded, ENOSYS will returned from this
+ * commit.
  */
 enum kvs_commit_flags {
     FLUX_KVS_NO_MERGE = 1, /* disallow commits to be mergeable with others */
diff --git a/src/modules/kvs/kvstxn.c b/src/modules/kvs/kvstxn.c
index d0ada07..3f61cd7 100644
--- a/src/modules/kvs/kvstxn.c
+++ b/src/modules/kvs/kvstxn.c
@@ -69,12 +69,24 @@ struct kvstxn {
     kvstxn_mgr_t *ktm;
     /* State transitions
      *
+     * INIT - perform initializations / checks
+     * LOAD_ROOT - load KVS root
+     *           - if needed, report missing refs to caller and stall
+     * APPLY_OPS - apply changes to KVS
+     *           - if needed, report missing refs to caller and stall
+     * STORE - generate dirty entries for caller to store
+     * GENERATE_KEYS - stall until stores complete
+     *               - generate keys modified in txn
+     * SYNC_CONTENT_FLUSH - call content.flush (for FLUX_KVS_SYNC)
+     * SYNC_CHECKPOINT - call kvs_checkpoint_commit (for FLUX_KVS_SYNC)
+     * FINISHED - end state
+     *
      * INIT -> LOAD_ROOT
      * LOAD_ROOT -> APPLY_OPS
      * APPLY_OPS -> STORE
-     * STORE -> POST_STORE
-     * POST_STORE -> FINISHED
-     * POST_STORE -> SYNC_CONTENT_FLUSH
+     * STORE -> GENERATE_KEYS
+     * GENERATE_KEYS -> FINISHED
+     * GENERATE_KEYS -> SYNC_CONTENT_FLUSH
      * SYNC_CONTENT_FLUSH -> SYNC_CHECKPOINT
      * SYNC_CHECKPOINT -> FINISHED
      */
@@ -83,7 +95,7 @@ struct kvstxn {
         KVSTXN_STATE_LOAD_ROOT = 2,
         KVSTXN_STATE_APPLY_OPS = 3,
         KVSTXN_STATE_STORE = 4,
-        KVSTXN_STATE_POST_STORE = 5,
+        KVSTXN_STATE_GENERATE_KEYS = 5,
         KVSTXN_STATE_SYNC_CONTENT_FLUSH = 6,
         KVSTXN_STATE_SYNC_CHECKPOINT = 7,
         KVSTXN_STATE_FINISHED = 8,
@@ -231,7 +243,7 @@ json_t *kvstxn_get_keys (kvstxn_t *kt)
 void kvstxn_cleanup_dirty_cache_entry (kvstxn_t *kt, struct cache_entry *entry)
 {
     if (kt->state == KVSTXN_STATE_STORE
-        || kt->state == KVSTXN_STATE_POST_STORE) {
+        || kt->state == KVSTXN_STATE_GENERATE_KEYS) {
         char ref[BLOBREF_MAX_STRING_SIZE];
         const void *data;
         int len;
@@ -993,7 +1005,7 @@ kvstxn_process_t kvstxn_process (kvstxn_t *kt, const char *rootdir_ref)
             /* cache now has ownership of rootcpy, we don't need our
              * rootcpy anymore.  But we may still need to stall user.
              */
-            kt->state = KVSTXN_STATE_POST_STORE;
+            kt->state = KVSTXN_STATE_GENERATE_KEYS;
             json_decref (kt->rootcpy);
             kt->rootcpy = NULL;
 
@@ -1008,7 +1020,7 @@ kvstxn_process_t kvstxn_process (kvstxn_t *kt, const char *rootdir_ref)
             kt->newroot_entry = entry;
             cache_entry_incref (kt->newroot_entry);
         }
-        else if (kt->state == KVSTXN_STATE_POST_STORE) {
+        else if (kt->state == KVSTXN_STATE_GENERATE_KEYS) {
             /* Caller didn't call kvstxn_iter_dirty_cache_entries() */
             if (zlist_first (kt->dirty_cache_entries_list)) {
                 kt->blocked = 1;
@@ -1133,7 +1145,7 @@ int kvstxn_iter_dirty_cache_entries (kvstxn_t *kt,
 {
     struct cache_entry *entry;
 
-    if (kt->state != KVSTXN_STATE_POST_STORE) {
+    if (kt->state != KVSTXN_STATE_GENERATE_KEYS) {
         errno = EINVAL;
         return -1;
     }

```

PR #4351, opened 2022-06-01
job-manager: unwind job event posting recursion

garlick:
Problem: when a job event is posted to a job in the job manager, the "action" for the event might post more events.  Since each event is processed immediately, this can lead to re-entry to the event posting code, and myriad corner cases.

This PR does not switch to a full asynchronous event posting model like the broker's state machine, where events are posted to a queue and always processed in another reactor handler. I went there first and found there was quite a bit of code that assumed that the effects of posting an event are immediate.  Instead this takes a more conservative step: an event queue is added to each job.  Any events that are posted while another event is in progress are added to the queue.  Then when the event is finished, the queue is processed in order.  The queue is always left empty when the original event posting returns.

Effectively this means that some events happen immediately, but others happen "later", depending on whether or not they are posted from an event "action".  But it does keep the events in order and avoids the recursion that was hard to think about.  The current set of tests are all passing, but this difference in event posting semantics could pose problems down the road.  We may be able to convert to a fully asynchronous model at some point and then we would have consistency.

Marking as WIP because I'd like a bit more time to think about this and decide whether more tests are needed.

### DIFF
--- a/src/modules/job-manager/event.c
+++ b/src/modules/job-manager/event.c
@@ -42,6 +42,7 @@
 
 #include "src/common/libczmqcontainers/czmq_containers.h"
 #include "src/common/libeventlog/eventlog.h"
+#include "src/common/libutil/errno_safe.h"
 #include "ccan/ptrint/ptrint.h"
 #include "ccan/str/str.h"
 
@@ -765,10 +766,10 @@ static int event_job_cache (struct event *event,
     return job_event_id_set (job, id);
 }
 
-int event_job_post_entry (struct event *event,
-                          struct job *job,
-                          int flags,
-                          json_t *entry)
+int event_job_process_entry (struct event *event,
+                             struct job *job,
+                             int flags,
+                             json_t *entry)
 {
     int rc;
     flux_job_state_t old_state = job->state;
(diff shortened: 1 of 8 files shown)


garlick:
I updated this PR since, somewhat shockingly, the tests are all passing now, but it is still very much a WIP.

There are a lot of changes packed in here and I may need to invest some effort into splitting it into multiple PRs.

There are also some implied changes to RFC 21, for example, adding a `valid` event to transition the job from NEW->DEPEND instead of `submit`, and allowing a job to transition back to PRIORITY from ALLOC when urgency is updated.

Please hold off reviewing while I sort things out.

garlick:
Rebased on #4366 which simplifies this PR quite a bit.

garlick:
Dropping WIP from this one because I think it will be ready for a review once #4366 goes in.

garlick:
Ugh, coverage is rather sad here.  I'll work on that.

src/modules/job-manager/event.c
```diff
@@ -771,7 +771,6 @@ int event_job_process_entry (struct event *event,
                              int flags,
```
chu11, line 755: commit message typo "explaning" -> "explaining"

src/modules/job-manager/event.c
```diff
@@ -637,7 +637,8 @@ int event_job_update (struct job *job, json_t *event)
     }
     else if (streq (name, "exception")) {
         int severity;
-        if (job->state != FLUX_JOB_STATE_INACTIVE) {
+        if (job->state != FLUX_JOB_STATE_INACTIVE
+            && job->state != FLUX_JOB_STATE_NEW) {
```
chu11, line 640: in the prior commit (` job-manager: relax state machine `) you changed:

```
        if (job->state == FLUX_JOB_STATE_NEW
            || job->state == FLUX_JOB_STATE_INACTIVE)
            goto inval;
```

to

```
        if (job->state != FLUX_JOB_STATE_INACTIVE) {
```

so here you add `job->state != FLUX_JOB_STATE_NEW` back.   Was there a commit series mistake somewhere in here?

chu11 (commented):
nothing big caught my eye, just small things

src/modules/job-manager/event.c
garlick: Yeah, thanks - will fix.

garlick:
Fixed those two things and pushed.

chu11 (approved):
LGTM, although I'm not too familiar with the jobtap stuff, dunno if you'd like for @grondo to take a skim too.

grondo (approved):
Yeah, this LGTM! Nice work!

garlick:
Oh excellent, OK thanks guys!  Setting MWP.

grondo:
Cool! A big improvement!

Issue #3445, opened 2020-12-31
shell: use of killpg(2) to terminate processes may break cleanup for some applications


grondo:
This came up from a discussion with @FrankD412 on slack.

When a job is canceled, a `SIGTERM` is sent to all job shell processes which each then forward that signal to all tasks with `flux_subprocess_kill(3)`. By default I believe, `flux_subprocess_kill` uses `killpg(2)` to send the signal to the entire process group to ensure as much cleanup as possible (similar to a Unix shell forwarding SIGINT to the current process group which owns the terminal).

This is causing problems for @FrankD412, however, because their main process spaws worker subprocesses which unexpectedly receive SIGTERM on `flux job cancel`, interrupting a clean exit which requires that the main process instruct the workers to clean up. In this case, it would be better if the initial `SIGTERM` were sent _only_ to the main task, and the subsequent `SIGKILL` (after grace period) was sent to the process group.

Unfortunately, this would cause problems in the common case. That is, if a task creates children but _doesn't_ block SIGTERM, the main task would terminate on `flux job cancel` and the children would be left around. This is specifically the case that the use of `killpg(2)` is trying to handle.

However, maybe a shell option could be provided to change this behavior, e.g. `-o no-killpg` or similar could disable the use of `killpg(2)` in general by the shell could be used in cases where like this. The option would be ignored, of course, when the signal is `SIGKILL`.

Note that there are also other mechanisms than `flux job cancel` (which, in general, would be expected to kill jobs immediately) to tell a job is should gracefully exit. One idea would be for jobs to subscribe to exceptions and then generate a "job shutdown" exception with nonzero severity. This would allow the job to gracefully shutdown without the use of signals, but requires more work than a signal handler, so may not be preferred by application developers.

There may be other ideas for alternate non-fatal ways to trigger job cleanup as well.

dongahn:
@FrankD412's email description was that those subprocesses are hung when the parent process is killed. @FrankD412: are the subprocesses also install SIGTERM and ignore it?

FrankD412:
[...]

PR #4355, opened 2022-06-06
shell: execute job tasks in their own process group

grondo:
This PR makes a slight change in the way the job shell launches tasks so that each task is in its own process group. This allows the shell to use `killpg(2)` (via libsubprocess) when forwarding signals to a job, so that even children of job tasks are delivered a signal. This should result in better cleanup when job tasks call `fork()`, though note that this may affect an existing use case: #3445.

IMO, it makes sense that the job shell would treat tasks similar to the real shell and use process groups to manage sets of tasks. However, since this could break existing users, maybe we need a shell option to disable it?

I'll work on that next since it should be fairly trivial to implement. Then we'll need some tests. I'm thinking `-o nosetpgrp` but if someone has a better idea please let me know.

### DIFF
--- a/src/shell/task.c
+++ b/src/shell/task.c
@@ -193,7 +193,7 @@ int shell_task_start (struct shell_task *task,
                       shell_task_completion_f cb,
                       void *arg)
 {
-    int flags = 0;
+    int flags = FLUX_SUBPROCESS_FLAGS_SETPGRP;
     flux_subprocess_hooks_t hooks = {
         .pre_exec = subproc_preexec_hook,
         .pre_exec_arg = task,
--- a/src/shell/task.c
+++ b/src/shell/task.c
@@ -44,6 +44,7 @@
 
 #include "src/common/libczmqcontainers/czmq_containers.h"
 
+#include "internal.h"
 #include "task.h"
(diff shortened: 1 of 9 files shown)


garlick:
Just tried this with the two `stress` tests discussed in flux-framework/flux-security#142 (one where job is killed due to expiration, one where job is terminated interactively with ctrl-C.  Both cleaned up properly.  Thanks!

src/shell/task.c
```diff
@@ -193,7 +193,7 @@ int shell_task_start (struct shell_task *task,
                       shell_task_completion_f cb,
```
garlick, line 193: commit message s/caould/could/

garlick (approved):

grondo:
Just pushed some commits on top that add a new `-o setpgrp` option to the job shell which reverts to the old behavior of running job tasks in the job shell's process group.

grondo:
I had to fix a testing bug (use of `python` instead of `flux python`). Squashed and pushed again.

garlick (approved):
Option changes LGTM!

grondo:
Thanks! Setting MWP.

PR #4360, opened 2022-06-07
t2201-job-cmd.t: fix bug in UTF-8 test

### LINKED ISSUE #4359: not ok: flux-job: f58 can use multibyte prefix
This test is failing in the buildfarm:
```
UTF8_LOCALE=$(locale -a | grep UTF-8)
test -n "$UTF8_LOCALE" && test_set_prereq UTF8_LOCALE
test_expect_success UTF8_LOCALE 'flux-job: f58 can use multibyte prefix' '
        jobid=$(LC_ALL=${UTF8_LOCALE} flux job id --to=f58 1) &&
        test "$jobid" = "ƒ2"
'
```

with the error:
```
/bin/sh: warning: setlocale: LC_ALL: cannot change locale (en_US.utf8@ampm.UTF-8
en_US@ampm.UTF-8)
```

The issue is likely that there is more than one match to `locale -a | grep UTF-8`, so we're trying to assign a string with newlines to `UTF8_LOCALE`.

grondo:
Problem: Before the test "f58 can use multibyte prefix", the test
script attempts to find a valid utf8 locale using locale -a piped
to grep.  However, if multiple locales match the pattern, then the
UTF8_LOCALE variable used to set LC_ALL is invalid since it will
contain newlines.  This breaks the test on machines with many locales.

Use head(1) to ensure UTF8_LOCALE only contains a single line.

Add a line of debug output to the test in case there are further
problems in the future.

Fixes #4359

### DIFF
--- a/t/t2201-job-cmd.t
+++ b/t/t2201-job-cmd.t
@@ -169,9 +169,10 @@ test_expect_success 'flux-job: id --to=f58 works' '
 	test "$jobid" = "$MINJOBID_F58"
 '
 
-UTF8_LOCALE=$(locale -a | grep UTF-8)
+UTF8_LOCALE=$(locale -a | grep UTF-8 | head -n1)
 test -n "$UTF8_LOCALE" && test_set_prereq UTF8_LOCALE
 test_expect_success UTF8_LOCALE 'flux-job: f58 can use multibyte prefix' '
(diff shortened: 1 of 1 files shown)


garlick (approved):
LGTM!

grondo:
Thanks! Setting MWP.

PR #4361, opened 2022-06-07
wait for 'finish' instead of 'clean' event in flux-mini run and flux-job attach

grondo:
This PR changes `flux job attach` (and thus `flux mini run`) to wait only for the job `finish` event instead of `clean` to avoid unnecessarily waiting in these interactive commands when there is a slow job epilog or some other  delay between the finish and clean events.

A new `--wait-event` option is added to `flux job attach` and exposed in `flux mini run`. `flux mini run --wait-event=clean ...` will restore the current behavior for example.

### DIFF
--- a/src/cmd/flux-job.c
+++ b/src/cmd/flux-job.c
@@ -229,6 +229,10 @@ static struct optparse_option attach_opts[] =  {
     { .name = "show-exec", .key = 'X', .has_arg = 0,
       .usage = "Show exec events on stderr",
     },
+    { .name = "wait-event", .key = 'w', .has_arg = 1, .arginfo = "NAME",
+      .usage = "Wait for event NAME before detaching from eventlog "
+               "(default=finish)"
+    },
     { .name = "label-io", .key = 'l', .has_arg = 0,
       .usage = "Label output by rank",
     },
@@ -1345,6 +1349,7 @@ struct attach_ctx {
     flux_jobid_t id;
     bool readonly;
     const char *jobid;
+    const char *wait_event;
     flux_future_t *eventlog_f;
     flux_future_t *exec_eventlog_f;
     flux_future_t *output_f;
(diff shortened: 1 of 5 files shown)


garlick (approved):
This worked out nicely.  Looks like it was handy to have `flux_job_event_watch()` so whoever came up with that - good job!

LGTM

t/t2274-manager-perilog.t
```diff
@@ -111,7 +111,7 @@ test_expect_success 'perilog: load perilog.so plugin' '
 test_expect_success 'perilog: configured prolog/epilog works for jobs' '
 	undrain_all &&
 	jobid=$(flux mini submit --job-name=works hostname) &&
-	flux job attach -vE $jobid > submit1.log 2>&1 &&
+	flux job attach -vE -w clean $jobid > submit1.log 2>&1 &&
```
chu11, line 114: commit message typo `Howevetr` -> "However"

grondo:
Thanks! I've set MWP.

This behavior was gratifyingly easy to change. Hopefully this doesn't break any existing users.

src/cmd/flux-job.c
```diff
@@ -2211,13 +2211,13 @@ int cmd_attach (optparse_t *p, int argc, char **argv)
         log_err_exit ("flux_get_reactor");
 
     /*  Check for the event name that attach should wait for in the
-     *   main job eventlog. The default is the "clean" event.
+     *   main job eventlog. The default is the "finish" event.
```
chu11, line 2214:
```
On systems with long running job or manager epilogs, or where
otherwise there might be a long delay between a job finishing and the
job's clean event, it is probably unexpected to have flux-job attach
and therefore flux-mini run hang until these post job operations
complete.
```

Commit message seems a little run on.  Perhaps

```
Sometimes there might be a long delay between a job finishing and the
job's clean event, such as with a long running manager epilog.
It is probably unexpected to have flux-job attach, and therefore flux-mini run,
hang until these post job operations  complete.
```

chu11 (approved):
LGTM, just some comment nits

t/t2274-manager-perilog.t
grondo: Oops, sorry! Set MWP too soon! :frowning_face:

PR #4363, opened 2022-06-08
testsuite: fix perilog sanity test

garlick:
Problem: now that flux job attach returns once the finish event
is posted, the prolog/epilog test in t2212-job-manager-plugins.t
does not see the epilog-finish event and a test bug lets it succeed
anyway.

The exit code of the first command in a pipeline is ignored by the
shell, so replace grep|cut expressions in the test with a shell
function that fails if either command fails.

Add `--wait-event clean` to the flux-job attach command so it captures
the epilog-finish event once again.


chu11 (approved):
LGTM!

grondo (approved):
Ugh, sorry about that!

I hope there aren't too many other test bugs like this one lurking about :-(

PR #4364, opened 2022-06-09
job-manager:  transition back to PRIORITY state on urgency update

garlick:
Problem:  #4351 changes the order of job events when processing one event triggers another to be posted.  This broke the urgency update so that the queue was reordered _before_ the priority value (that determines the order) was set.

Since we now support a cycle back from SCHED to PRIORITY  for the `flux-restart` event, it seemed like it would be trivial to allow the same when the urgency is updated in SCHED state.  This not only fixes the above problem but simplifies the control flow. Furthermore, if a priority plugin takes a long time to update the priority (perhaps it needs to phone home) then the job is listed in PRIORITY state instead of SCHED, providing greater transparency to the user.

This broke the `priority-hold` plugin so it was changed to post an `urgency` event in the `job.state.depend` callback instead, and was renamed to `submit-hold` (see commit message for more detail).

This was peeled off of #4351 and can be applied independently.  It should be accompanied by an update to  [RFC 21](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_21.html)'s state diagram.


src/modules/job-manager/event.c
```diff
@@ -343,7 +343,10 @@ int event_job_action (struct event *event, struct job *job)
              * In the event we have re-entered this state from the
              * SCHED state, dequeue the job first.
              */
-            alloc_dequeue_alloc_request (ctx->alloc, job);
+            if (job->alloc_pending)
```
grondo, line 346: Commit typo: "Change hte state machine"

grondo (commented):
I'm sure I'm missing something obvious, but if a priority plugin now won't be able to return a raw priority value from the `job.state.priority` callback, will this break the flux-accounting multi factor priority plugin? The "hold" plugin was merely a simple example of returning a priority from this state, and if that doesn't work, how can any other priority plugin?

Apologies if I'm just being dense, I know you've been working on this for quite awhile...

garlick:
> I'm sure I'm missing something obvious, but if a priority plugin now won't be able to return a raw priority value from the job.state.priority callback, will this break the flux-accounting multi factor priority plugin? The "hold" plugin was merely a simple example of returning a priority from this state, and if that doesn't work, how can any other priority plugin?

I hope I'm not missing something either, but the intent is not to change how the  `job.state.priority` callback works, but to replace the hard wired `urgency` event code path with a transition back to PRIORITY state.  Before this change, the `urgency` event would trigger a call to plugin code that posts a `priority` event and reorders the queue.  After #4351, the queue would be reordered and _theN_ the `priority` event would be processed so the reorder was effectively a no-op.   Rather than straighten out that code, it seemed simpler to just transition back to PRIORITY state and let the priority plugin do its thing once more.

Does that make sense?

grondo:
Yes, I think that makes sense.

Ah, and I think I just understood why this broke the `priority-hold` plugin. Since th `priorty-hold` plugin _always_ returns a `0` priority, now that jobs return to PRIORITY state on urgency, there was no way to release jobs with this plugin loaded, since it continually returns `0`. Therefore a strictly "priority" plugin can no longer implement the feature we want here, to hold all submitted jobs until they are released by an urgency update.

Is that correct?

garlick:
Exactly.   So to get the same effect, I implemented a plugin that posts the `urgency` event (setting the value to 0) on entry to the DEPEND state.  I'll try to reword that commit description since it's not as obvious as the way you restated it.

grondo:
Thanks! The fix was clever, nice work.

garlick:
Thanks!  I reworded that commit slightly and fixed the spelling error.

grondo (approved):
LGTM!

garlick:
Setting MWP

Issue #3812, opened 2021-08-03
jobtap: call job.validate on plugin load


grondo:
As noted in flux-framework/flux-accounting#136, it is painful to handle recreation of state on a plugin reload. This is because on reload only the `job.new` callback is called. Therefore if state is accumulated in `job.validate` this effort currently needs to be duplicated in `job.new`, but only for jobs that are not in the NEW state. This puts an undue burden on plugin developers.

We should consider if calling `job.validate` on all active jobs on plugin load would make this easier.

Active jobs being rejected could have an exception raised, for example if a plugin with more restrictive limits was loaded.

PR #4366, opened 2022-06-13
job-manager: transition NEW->DEPEND on "validate" event

garlick:
Problem: the job "submit" event is defined as the first job event and also signifies that the job is valid.  This leaves no room for e.g. "jobspec-update" events to be posted before the job is validated.  Such events are needed to set default values for queue name or duration, and those values affect validation.

Alter the job state machine to transition NEW->DEPEND on a new "valid" event.  The order is now
- post "submit" event
- (future: post "jobspec-update" events)
- `job.validate` jobtap callback
- `job.new` jobtap callback
- post "valid" event
- job transitions to DEPEND state

Or if the job is invalid:
- post "submit" event
- (future: post "jobspec-update" events)
- `job.validate` jobtap callback
- post "invalid" event (journal only)
- job is removed

This will require a change to [RFC 21](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_21.html).

One problem this creates is "submit" is not captured by plugins that subscribe to events in `job.new`, so it seems like some thought might need to go into how jobtap plugins interact with the job manager during job creation, and whether we need some additional callbacks/interfaces as a result of this change (or maybe as part of this change).


grondo:
Once potential gotcha: The naming of the "valid" event could be confused with any valid event (see, even that statement is confusing) It actually confused me for a second until it clicked. The difference is obvious in hindsight, but I wonder if it will be troublesome when discussing this particular event by name in other contexts?

The only suggestion I have is to rename the event to something a bit more specific, like perhaps `validate` (`invalidate` would also work for jobs that are rejected to avoid talking about jobs that are rejected on "invalid" events..)

However, I admit maybe this is only a problem for me, so I defer to your final judgement.

garlick:
> The only suggestion I have is to rename the event to something a bit more specific, like perhaps validate (invalidate would also work for jobs that are rejected to avoid talking about jobs that are rejected on "invalid" events..)

Great idea, that will avoid some awkward sentences.  I'll make that change.

Any thoughts on the ramifications of calling `job.new` before the submit event is posted so that plugins can have access to all events?  Right now I believe `job.new` is interpreted as "here is a new, valid job".   A plugin _could_ subscribe to events and wait for "invalidate" or "validate" events to determine how to proceed from there.  Is that putting too much burden on the plugin author?

grondo:
> Any thoughts on the ramifications of calling job.new before the submit event is posted so that plugins can have access to all events? Right now I believe job.new is interpreted as "here is a new, valid job". A plugin could subscribe to events and wait for "invalidate" or "validate" events to determine how to proceed from there. Is that putting too much burden on the plugin author?

I'm actually pondering if we could get rid of the `job.new` callback. Since addition of `job.validate`, it has turned into a bit of a kludge. The only reason this callback is currently necessary (I _think_), is because the `job.validate` callback is not invoked on reload. However, this causes an awkward situation for plugins that need to validate jobs _and_ have some internal state for active jobs, because they need to register both the `job.validate` callback _and_ a `job.new` callback to handle the case of reload. Thus we have the open issue #3812, which proposes that we go ahead and call `job.validate` on reload. If we fixed this, and updated any users of `job.new` (probably only the multi-factor priority plugin), then perhaps we could drop `job.new`.

That being said, does that actually help?

grondo:
Hm, reading back what you've written, I don't think my line a reasoning above helps at all. We have a couple use cases that are tricky now:

 * A jobtap plugin that wants to subscribe to events misses the `submit` event (and possibly jobspec-update events?) because of the proposed order of events in the new scheme
 * A jobtap plugin that wants to only track new, valid jobs will have to register state for a job in the `job.new` callback, but then watch for an `invalidate` event to purge the state of that job.

Maybe we could treat `job.new` as you propose above. All submitted jobs make this callback, even if they are later rejected by a `job.validate` callback. This nicely handles the first case as you note.

Then to handle "a new, valid job is submitted", we add a new callback, name TBD, that is triggered by the `validate` event?

garlick:
Yes I think you captured the issues that concern me with your two items above.

So if we introduce `job.validated` callback (basically an alias for `job.event.validate` with no subscription required) would that do it?  We probably also need a `job.invalidated` callback for the case where a job is rejected after `job.new` - or maybe `job.validate` just has a boolean parameter?

> Edit: and we also fix https://github.com/flux-framework/flux-core/issues/3812 to handle the case of setting up state in both job.validate and job.new?

Were you thinking this fixes that problem?

grondo:
> So if we introduce job.validated callback (basically an alias for job.event.validate with no subscription required) would that do it? 

Yeah, though we have to make sure that callback is also made for jobs during plugin and job manager reload.
I'm not sure there is enough of a use case for `job.invalidated` that we would need to support that.

> > Edit: and we also fix #3812 to handle the case of setting up state in both job.validate and job.new?
> Were you thinking this fixes that problem?

Not necessarily this particular problem. However, it would simplify updating the multifactor priority plugin I think. I should probably look at how that plugin uses `job.new` first, though. 

I.e. we could first support `job.validate` on reload, then update `mf_priority.so` to do its job tracking completely in `job.validate`, then finally change the behavior of `job.new`.

This is just one proposal of how we could make these changes, perhaps after looking at `mf_priority` a different approach will be evident.

garlick:
> I'm not sure there is enough of a use case for job.invalidated that we would need to support that.

I guess my worry would be that per-job state created in `job.new` might leak if there is no callback to indicate that the job was rejected by the validator, since in that case there is no exception and no transition to inactive.

grondo:
> and no transition to inactive.

Ah, I hadn't realized that! Good point.. Hm, it would be nice if there was a matching callback to `job.new` that was always called before a job was destroyed/inactive.

grondo:
> I guess my worry would be that per-job state created in job.new might leak if there is no callback to indicate that the job was rejected by the validator, since in that case there is no exception and no transition to inactive.

In most cases, the creation of per-job state in the `job.new` callback should probably be discouraged. It will likely be just wasted cycles for rejected jobs. It might even cause unintended consequences if jobs that aren't technically "active" are counted as such. I'm kind of contemplating if we can get rid of `job.new` if we enable `job.validate` on reload... Needs more thought though.

garlick:
We'll want some jobtap callback that comes before `job.validate` so jobspec-update events can be posted to set defaults.  The modified jobspec would then be checked by `job.validate`.

I guess that doesn't say anything about whether `job.new` should be kept.  It would be somewhat awkward to subscribe to events in `job.new`, then post jobspec-update events from `job.event.submit`, although that is vaguely how I thought it might work.   Maybe it would be better to have a purpose built callback for this.

garlick:
It doesn't look like this PR would break the mf_priority plugin or either of the coral2 plugins, if we don't change the semantics for `job.new`.  The fact that the submit event cannot be captured does not affected those plugins, as far as I can tell.

It seems like `job.validate` is the wrong place to create job state, because if there are multiple plugins loaded that implement `job.validate` the callbacks themselves won't know if the job is accepted or not since all of them would need to pass before the "validate" event is posted and the job is accepted.

Maybe we need `job.create` and `job.destroy` callbacks that are always called as a pair, one as early as possible (e.g. before the "submit" event or during restart), and the other as late as possible (after the "clean" or "invalidate" event).

`job.create` would be able to subscribe to events and receive all of them.  It could distinguish a job that has been accepted (the restart case) from one that hasn't by looking for a job state of NEW.  If a plugin needs to do something on acceptance of a job in NEW state, it could subscribe to `job.event.validate` or `job.state.depend`.

garlick:
Yay.  Next question is should those be added in this PR or another?  Since this PR corresponds to the proposed state machine change over in flux-framework/rfc#337, perhaps it makes sense for this to remain standalone for release notes clarity?

grondo:
> perhaps it makes sense for this to remain standalone for release notes clarity?

Yes, that makes sense to me!

grondo:
I've opened #4368 for the proposed jobtap changes.

src/modules/job-manager/submit.c
```diff
@@ -104,11 +104,26 @@ static int submit_job (struct job_manager *ctx,
...
     return 0;
+error_post_invalid:
+    /* Let journal consumers know this job did not pass validation and
+     * all data concerning it should be expunged.
+     */
+    (void)event_job_post_pack (ctx->event,
+                               job,
+                               "invalidate",
+                               EVENT_NO_COMMIT,
+                               NULL);
```
grondo, line 126: I might not be understanding the `invalidate` event, but should this event also be posted if a `job.validate` callback fails or  the `jobtap_check_dependencies()` function fails? Currently, this event is only posted if `event_job_post_pack ()` of the `valid` event fails.

grondo (commented):
One question and one small typo noted inline.

src/modules/job-manager/submit.c
garlick: Ugh yeah, it should _primarily_ be posted in the validation failure case.   When the job fails, the ingest module removes the job directory (including the eventlog) but "invalid" is needed to tell journal consumers that the job is DOA.   Big error, thank you!

garlick:
Fixed those things, rebased on current master, and forced a push.

I'll look at adding a test tomorrow to ensure the invalidate event is posted when validation fails.

grondo:
One builder appeared to hang before the docker image was created, so I've restarted it.

garlick:
Thanks!

PR #4367, opened 2022-06-13
libsdprocess: support cgroup memory configuration

chu11:
In this PR support `MemoryHigh` and `MemoryMax` configuration via systemd in launched processes.  They both can take bytes or percentages as input.

Built on top of #4255.  Decided to separate this out since it's a completely different type of property being supported in `sdprocess`.  However, the changes are much smaller than anticipated, could conceive of merging into #4255 if we wanted to.

### DIFF
--- a/doc/man5/flux-config-exec.rst
+++ b/doc/man5/flux-config-exec.rst
@@ -52,6 +52,31 @@ cpu_set_allowed
 
    (optional) Set to true to set limit execution to only allocated cores.
 
+MemoryHigh
+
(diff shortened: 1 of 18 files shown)


grondo:
I'd be fine merging this PR with its predecessor. I think then the PR could be titled, `sdexec: enable resource control properties: MemoryMax, MemoryHigh, AllowedCPUs, etc`

chu11:
ok, i'll merge into the prior PR

PR #4373, opened 2022-06-15
shell: obtain hwloc XML from enclosing instance

grondo:
Putting up this WIP PR before I head out of town (feel free to add early comments). This is a WIP because it still needs some tests added.

This change significantly speeds up throughput of many simultaneous jobs on systems with many cores like corona. Loading hwloc topology directly appears to be essentially serialized on these systems, so we go from a throughput of 6-8 job/s to 40-50 job/s by avoiding the redundant reload of topology via hwloc.

Loading hwloc topology is also very slow on corona (on the order of 1s), so it should help speed up short-lived jobs as well.

### DIFF
--- a/src/shell/shell.c
+++ b/src/shell/shell.c
@@ -421,6 +421,16 @@ int flux_shell_unsetenv (flux_shell_t *shell, const char *name)
     return json_object_del (shell->info->jobspec->environment, name);
 }
 
+int flux_shell_get_hwloc_xml (flux_shell_t *shell, const char **xmlp)
+{
+    if (!shell || !shell->info || !xmlp) {
+        errno = EINVAL;
+        return -1;
+    }
+    *xmlp = shell->info->hwloc_xml;
+    return 0;
+}
+
 static json_t *flux_shell_get_info_object (flux_shell_t *shell)
 {
     json_error_t err;
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -61,6 +61,7 @@ MAN3_FILES_PRIMARY = \
 	man3/flux_shell_aux_set.3 \
 	man3/flux_shell_current_task.3 \
 	man3/flux_shell_get_flux.3 \
+	man3/flux_shell_get_hwloc_xml.3 \
 	man3/flux_shell_get_info.3 \
 	man3/flux_shell_get_jobspec_info.3 \
 	man3/flux_shell_getenv.3 \
--- a/doc/man3/flux_shell_get_hwloc_xml.rst
+++ b/doc/man3/flux_shell_get_hwloc_xml.rst
@@ -1,0 +1,49 @@
+===========================
+flux_shell_get_hwloc_xml(3)
+===========================
+
+
+SYNOPSIS
+========
+
+::
+
+   #include <flux/shell.h>
+   #include <errno.h>
+
+::
+
+   int flux_shell_get_hwloc_xml (flux_shell_t *shell,
+                                 const char **hwloc_xml);
+
+
+DESCRIPTION
+===========
+
+``flux_shell_get_hwloc_xml()`` returns an hwloc XML string which has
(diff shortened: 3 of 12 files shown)


src/shell/shell.c
```diff
@@ -433,14 +433,15 @@ static json_t *flux_shell_get_info_object (flux_shell_t *shell)
...
-                            "{ s:I s:i s:i s:i s:s s:O s:O s:{ s:i s:b }}",
+                            "{ s:I s:i s:i s:i s:s s:O s:O s:s s:{ s:i s:b }}",
                             "jobid", shell->info->jobid,
                             "rank",  shell->info->shell_rank,
                             "size",  shell->info->shell_size,
                             "ntasks", shell->info->total_ntasks,
                             "service", shell_svc_name (shell->svc),
                             "jobspec", shell->info->jobspec->jobspec,
                             "R", shell->info->R,
+                            "hwloc", shell->info->hwloc_xml,
```
garlick, line 444: This is likely a minor consideration compared with the big performance gains attained here, but since the shell info object is accessed as a string, it would avoid unnecessary json encoding/decoding to store the hwloc xml under a separate aux key and add a specialized accessor that just passes the string through.

Since the hwloc xml can be large, and any access to the shell info object from a plugin would trigger its decode, it might be worth it.

garlick (commented):
Nice optimization!  Just one preliminary comment.

src/shell/shell.c
garlick: Oops, sorry I misread the code - the `json_dumps()` only happens if a plugin calls `flux_shell_get_info()` to access the object for parsing with another JSON library.  The unpack functions access the raw json object directly.

If the XML string is merely attached to a json object that is never serialized, then I'm not sure there is any downside at all to doing it this way,  other than jansson checking that the string is utf8 and scanning for the null terminator.  Even that could be avoided by packing with `s#` or `s%` if the XML content encoding says utf8.  Meh, premature optimization.

grondo:
Given concerns with efficiency of packing a giant XML object into a JSON object, I went ahead and modified this PR to add a new accessor `flux_shell_get_hwloc_xml(3)` instead of adding a new member to the shell "info" object. It seemed like the hwloc XML is probably special enough to have its own function. This avoids any concerns with shell plugins accessing the shell object, Jansson scanning the xml for utf-8 characters etc.

I've also added documentation and a test for invalid args, so I'll remove WIP.

garlick (approved):
LGTM!

grondo:
Hm, one of the builders hit this error I haven't seen before:
``` 
expecting success: 
  	flux module reload content-sqlite
  
  flux-module: broker.rmmod content-sqlite: No such file or directory
  not ok 34 - flux module reload content-sqlite
  
  expecting success: 
  	test $(flux module stats \
  	    --type int --parse store_time.count content-sqlite) -eq 0 &&
  	test $(flux module stats \
  	    --type int --parse object_count content-sqlite) -eq 1

  flux-module: flux_open: Connection refused
  ../../../t/t0012-content-sqlite.t: 2: test: -eq: unexpected operator
  not ok 35 - database survives module reload

```

The broker log shows:

```
 2022-06-20T22:24:05.171135Z content-sqlite.err[0]: /tmp/flux-3t0mRQ/content.sqlite: Permission denied
  2022-06-20T22:24:05.171223Z content-sqlite.err[0]: content_sqlite_create failed: Permission denied
  2022-06-20T22:24:05.171261Z content-sqlite.crit[0]: module exiting abnormally
  corrupted double-linked list
  flux-start: 0 (pid 166485) Aborted
  flux-start: 3 (pid 166507) Killed
  flux-start: 2 (pid 166497) Killed
  flux-start: 1 (pid 166493) Killed
```

So perhaps an unexpected condition on the particular docker container running the test?

Clearly not related to this PR, so I'll restart.

grondo:
Thanks! I've set MWP.

Issue #4375, opened 2022-06-16
broker: content.flush will hang forever if backing store not loaded / never loaded


chu11:
while working on #4348 noticed that `content.flush` will hang forever if a backing store isn't loaded.  This is:

A) not desirable and seems wrong
B) would cause the KVS w/ FLUX_KVS_SYNC to hang

returning an error (most likely ENOSYS) would be far more suitable.

Note that `content_unregister_backing_request()` should return errors on presently attempted content.flush futures too.

PR #4376, opened 2022-06-17
broker: return error in content.flush if no backing store exists 

chu11:
Problem: The content.flush target will never respond to a request if a backing store is never loaded.

Solution: Return error with ENOSYS if content.flush cannot be completed.  In addition, return ENOSYS if a content.flush is in progress when the backing store is unregistered.  Add regression test.

Fixes https://github.com/flux-framework/flux-core/issues/4375

### DIFF
--- a/src/broker/content-cache.c
+++ b/src/broker/content-cache.c
@@ -703,6 +703,16 @@ static void content_unregister_backing_request (flux_t *h,
         flux_log_error (h, "error responding to unregister-backing request");
     if (cache->acct_dirty > 0)
         flux_log (h, LOG_ERR, "%d unflushables", cache->acct_dirty);
+    /* If backing store is unloaded with pending flush requests, ensure that
+     * they receive an error response.
+     */
+    if (cache->flush_requests) {
+        request_list_respond_error (&cache->flush_requests,
+                                    cache->h,
+                                    ENOSYS,
+                                    NULL,
+                                    "flush");
+    }
     return;
 error:
     if (flux_respond_error (h, msg, errno, errstr) < 0)
@@ -785,6 +795,10 @@ static void content_flush_request (flux_t *h, flux_msg_handler_t *mh,
 {
     struct content_cache *cache = arg;
 
+    if (cache->rank == 0 && !cache->backing) {
+        errno = ENOSYS;
+        goto error;
+    }
     if (cache->acct_dirty > 0) {
         if (cache_flush (cache) < 0)
             goto error;
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -294,6 +294,7 @@ dist_check_SCRIPTS = \
 	issues/t4184-sched-simple-restart.sh \
 	issues/t4222-kvs-assume-empty-dir.sh \
 	issues/t4331-job-manager-purged-events.sh \
+	issues/t4375-content-flush-hang.sh \
 	issues/t4378-content-flush-force.sh \
 	issues/t4379-dirty-cache-entries-flush.sh \
 	python/__init__.py \
--- a/t/issues/t4375-content-flush-hang.sh
+++ b/t/issues/t4375-content-flush-hang.sh
@@ -1,0 +1,13 @@
+#!/bin/bash -e
+
+flux content flush
+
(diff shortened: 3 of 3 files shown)


chu11:
hmmm a lot of builders failing with

```
  ok 28 - t4375-content-flush-hang.sh
  PASS: t4000-issues-test-driver.t 28 - t4375-content-flush-hang.sh
  # passed all 28 test(s)
  1..28
  ERROR: t4000-issues-test-driver.t - exited with status 137 (terminated by signal 9?)
```

perhaps i'm not cleaning something up with my new regression test?  Not sure what could be causing this.

garlick:
Not sure if it is related to the test failure, but rc1 should still run the flush on all ranks b/c that ensures that all cache writes to rank 0 have completed on the other ranks.   The flush rpc only requires the backing store flush when it is run on rank 0 (sort of like the backing store were an extension of the cache hierarchy).

chu11:
> Not sure if it is related to the test failure, but rc3 should still run the flush on all ranks b/c that ensures that all cache writes to rank 0 have completed on the other ranks.

Hmmmm.  Ok, perhaps the ENOSYS should only be returned on rank 0 then?  On non-rank 0's it can still return success if the backing module doesn't exist?

garlick:
> perhaps the ENOSYS should only be returned on rank 0 then? On non-rank 0's it can still return success if the backing module isn't loaded?

Yes that fits the current design, such as it is.

chu11:
ok, think I figured out the issue causing all these test failures.  This might be an additional corner case in the `content-cache`.  My regression test is basically:

```
flux module remove content-sqlite
flux kvs put issue4375=1
! flux content flush
flux module load content-sqlite
```

When the `content-sqlite` module is loaded, it is re-registered with the content cache and `cache_flush()` is eventually called.  However, I don't think that the `put` listed above is ever flushed.  It's because in `content_store_request()`

```
    if (e->dirty) {
        if (cache->rank > 0 || cache->backing) {
            if (cache_store (cache, e) < 0)
                goto error;
            if (cache->rank > 0) {  /* write-through */
                if (msgstack_push (&e->store_requests, msg) < 0)
                    goto error;
                return;
            }
        }
    }
```

basically `cache_store()` never gets called, dirty entry never gets put onto the "flush list", thus never gets flushed.

I think the correct fix would be to do a scan, and make sure all dirty entries on flushable after a registration.  Lemme look into doing that.  If might be a per-requisite fix to this fix :-)

chu11:
agh, there wass another issue.

When the `content-sqlite` module is re-registered (`content_register_backing_request()`), it eventually calls `cache_flush()`, which eventually will call `content_store()` to flush all content that needs to be stored.

BUT ... `content-sqlite` hasn't completed setting itself up, thus you get a ENOSYS error.

I noticed in the manpage for `flux content` that it says "The **flux content flush** command initiates store requests
for any dirty entries in the local cache and waits for them to complete."

But that's not really the case, it just waits for all present dirty entries in the local cache to be flushed, but doesn't necessarily initiate the flush.  Perhaps needs to be remedied?  Will have to think about it.

chu11:
re-pushed, rebasing on #4380, which should fix some of the above issues

src/broker/content-cache.c
```diff
@@ -703,6 +703,11 @@ static void content_unregister_backing_request (flux_t *h,
         flux_log_error (h, "error responding to unregister-backing request");
     if (cache->acct_dirty > 0)
         flux_log (h, LOG_ERR, "%d unflushables", cache->acct_dirty);
+    request_list_respond_error (&cache->flush_requests,
+                                cache->h,
+                                ENOSYS,
+                                NULL,
+                                "flush");
```
garlick, line 710: Even though `request_list_respond_error()` is a no-op if there are no pending flush requests, having it run unconditionally looks suspicious.  Perhaps wrapping it in something like this would make it clearer:
```c
/* If backing store is unloaded with pending flush requests, ensure that
 * they receive an error response.
 */
if (cache->flush_requests != NULL) {
    request_list_respond_error (...);
}
```

garlick (approved):
Just a couple cosmetic comments, looks good!

Issue #4379, opened 2022-06-21
broker: some dirty content-cache entries will never be flushed


chu11:
As discussed in #4376, there is a corner case where dirty cache entries will never be flushed b/c they are never stored onto the flushables list.  example reproducer:

```
flux module remove content-sqlite
flux kvs put abc=1
flux module load content-sqlite
flux content flush
```

(Note that this issue is independent on #4378)

Issue #4378, opened 2022-06-21
broker: content.flush does not force flush


chu11:
While working on #4375 / #4376, it was determined that `content.flush` does not actually flush the cache.  It only waits if any flushes are correctly ongoing to finish.

This is probably not what is desired, leads to issues, and runs counter to documentation of `flux content flush`
.

chu11:
Note, I believe the core change was introduced in 5295581c

PR #4380, opened 2022-06-21
broker: content cache corner case corrections

chu11:
Fixes for #4378 and #4379.  Decided to try and fix both in one PR b/c they are very tightly integrated with each other.  One requires the other for a regression test, the other sort of doesn't appear without the other.

### DIFF
--- a/src/broker/content-cache.c
+++ b/src/broker/content-cache.c
@@ -774,6 +774,8 @@ static void content_flush_request (flux_t *h, flux_msg_handler_t *mh,
     struct content_cache *cache = arg;
 
     if (cache->acct_dirty > 0) {
+        if (cache_flush (cache) < 0)
+            goto error;
         if (msgstack_push (&cache->flush_requests, msg) < 0)
             goto error;
         return;
--- a/src/broker/content-cache.c
+++ b/src/broker/content-cache.c
@@ -611,7 +611,6 @@ static int cache_flush (struct content_cache *cache)
 {
     struct cache_entry *e;
     int last_errno = 0;
-    int count = 0;
     int rc = 0;
 
     while (cache->flush_batch_count < cache->flush_batch_limit) {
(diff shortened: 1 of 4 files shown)


src/broker/content-cache.c
```diff
@@ -774,6 +774,8 @@ static void content_flush_request (flux_t *h, flux_msg_handler_t *mh,
     struct content_cache *cache = arg;
 
     if (cache->acct_dirty > 0) {
+        if (cache_flush (cache) < 0)
```
grondo, line 787: Commit message typo: ..."when a cache flush has not yet initiated" -- "been initiated"?

t/Makefile.am
```diff
@@ -294,6 +294,8 @@ dist_check_SCRIPTS = \
 	issues/t4184-sched-simple-restart.sh \
 	issues/t4222-kvs-assume-empty-dir.sh \
 	issues/t4331-job-manager-purged-events.sh \
+	issues/t4378-content-flush-force.sh \
```
grondo, line 297: Commit msg: consider adding a commit body and re-wording commit subject to remove reference to github issue numbers. Not a requirement, but I do find that github issue numbers make a nice supplement to the commit message, but aren't as nice when they are the sole context. (E.g. when browsing `git log` locally you will have to separately look up issues online)

E.g. something like this (I may have gotten description wrong)
```
testsuite: add regression tests for content flush issues

Problem: No tests in the testsuite ensure that content flush is forced after
`flux content flush` nor that any dirty cache entries are flushed after a content
backing module is loaded. (Issues #4378 and #4379)

Add regression tests for these specific issues.
```

As I said, just a suggestion and it may be more personal preference, so feel free to ignore.

grondo (approved):
This LGTM. The solutions here seem pretty straightforward, but I do wonder if we should also wait for @garlick to approve when he returns from vacation since I'm pretty unfamiliar with this code.

I did have some suggestions inline for the commit messages, but feel free to ignore those.

grondo:
Also, please consider retitling this PR  to `content cache corner case corrections` :stuck_out_tongue_closed_eyes:

chu11:
re-pushed, just added a little extra comments / commit message info.

garlick (approved):
LGTM, thanks!

PR #4382, opened 2022-06-25
job-manager: fix case where a job manager epilog can't be configured without a prolog

grondo:
I noticed this while doing some testing. If only a `[job-manager.epilog]` is configured, without also configuring a `[job-manager.prolog]`, then the epilog does not get initiated by the job manager `perilog.so` plugin.

The problem is that the plugin only subscribes to job events during the prolog handler. So, if a prolog is not configured, this subscription doesn't happen and the `job.event.finish` callback isn't received by the plugin.

The fix is to always subscribe to events in the `job.state.run` callback if either a prolog or epilog action is configured.

### DIFF
--- a/src/modules/job-manager/plugins/perilog.c
+++ b/src/modules/job-manager/plugins/perilog.c
@@ -314,11 +314,6 @@ static int run_command (flux_plugin_t *p,
         return -1;
     }
 
-    if (prolog) {
-        if (flux_jobtap_job_subscribe (p, FLUX_JOBTAP_CURRENT_JOB) < 0)
-            flux_log_error (h, "prolog: flux_jobtap_job_subscribe");
-    }
-
     return 0;
 }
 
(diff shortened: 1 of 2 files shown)


src/modules/job-manager/plugins/perilog.c
```diff
@@ -327,6 +322,21 @@ static int run_cb (flux_plugin_t *p,
...
 {
+
+    /*  If an epilog command is registered, then we need to subscribe
+     *   to all events for this job to catch job.event.finish.
+     */
+    if (perilog_config.epilog_cmd != NULL) {
+        if (flux_jobtap_job_subscribe (p, FLUX_JOBTAP_CURRENT_JOB) < 0) {
+            flux_jobtap_raise_exception (p,
+                                         FLUX_JOBTAP_CURRENT_JOB,
+                                         "prolog",
```
chu11, line 336: "epilog" instead of "prolog"?

chu11: oh wait, b/c of the other comment I made, if this was a "fix in a wrong commit", then "prolog" might be fine here.

src/modules/job-manager/plugins/perilog.c
```diff
@@ -323,10 +323,13 @@ static int run_cb (flux_plugin_t *p,
...
-     *   to all events for this job to catch job.event.finish.
+    /*
+     *  Subscribe to job events if an epilog or prolog command is
+     *   registered. This is needed to allow this plugin to subscribe
+     *   to the finish event for the epilog, and any exception events
+     *   for the prolog (so it can be canceled).
      */
-    if (perilog_config.epilog_cmd != NULL) {
+    if (perilog_config.epilog_cmd || perilog_config.prolog_cmd) {
         if (flux_jobtap_job_subscribe (p, FLUX_JOBTAP_CURRENT_JOB) < 0) {
```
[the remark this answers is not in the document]
grondo: Yes, thanks for catching that! This commit should only touch the testsuite

grondo:
Ok @chu11, I moved the fixup to the right commit and this PR should make more sense now :-)

garlick (approved):
LGTM!

src/modules/job-manager/plugins/perilog.c
```diff
@@ -327,6 +322,24 @@ static int run_cb (flux_plugin_t *p,
                    flux_plugin_arg_t *args,
                    void *arg)
 {
+
```
chu11, line 325: dumb commit message typo, an extra space here, `unless a  prolog`

chu11 (approved):
just saw an extra space in commit message, otherwise all looks good

src/modules/job-manager/plugins/perilog.c
grondo: Thanks! Fixed.

grondo:
Thanks! I've set MWP.

Issue #4301, opened 2022-04-25
kvs: add defensive checkpoints


garlick:
Problem: currently the kvs root reference is not periodically updated while an instance runs, so a  rank 0 crash could result in significant loss of data.

The root reference is currently written out in the following circumstances
- what the start event is posted to `admin.eventlog` in rc1
- when the kvs is unloaded in rc3

Edit: note that the content cache on rank 0 must be flushed to the backing store before the checkpoint is written, otherwise a mistimed crash could result in dangling references.

garlick:
One thought is as a first cut, we could add a `flux kvs checkpoint` subcommand which
- calls `flux_kvs_getroot()`
- makes the `content.flush` RPC on rank 0
- calls  `kvs_checkpoint_commit()`

And then add a flux cron job for doing this periodically.

What would be really nice, but possibly a bit of work, would be to add a new KVS flag that could be added to `flux_kvs_commit()` so that once the commit finishes, we are guaranteed the result is on the backing store.   Then if we could judiciously use this in places where "durability" of the commit is critical.   It would mean adding the above sequence to the KVS commit logic for the primary namespace only.  If @chu11 thinks this might be feasible, perhaps we could open a separate issue on that and let this one be closed with something simple like the cron job.

garlick:
I opened a separate issue on the KVS flag so we can consider it more carefully there.

Meanwhile, let's say we can close this issue if we implement a mechanism for the kvs itself to periodically update the checkpoint if there has been commit activity and no checkpoint for some period of time.   This code would be useful even after a flag is added.

chu11:
pondering this a bit yesterday. thinking about defensive checkpointing in a range, from most aggressive to least

1) after every commit
2) after every stdio commits
3) after every eventlog entry
4) after every job completes
5) every X time period (i.e. via cron)
6) every time user / sysadmin requests (i.e. checkpoint command per comments above)
7) after flux shuts down

could be missing some other options that I'm not thinking about right now.

I think 1 & 2 are nuts, and probably 3 too.  But for the system instance, I could see 4 as potential option, b/c high throughput jobs from users should be through their own instance (presumably user checkpoints will not be as often).  So the checkpointing shouldn't be nuts?

Another approach might be to support a checkpointing option within the system instance, letting admins configure for the safety level they wish, sorta like with verbosity in logs.

Nevertheless, might be worthwhile to do some experiments to see how much slower throughput is after #4348 goes in with a number of the options above.

[...]

PR #4383, opened 2022-06-29
kvs: add defensive checkpoint via sync configuration option

chu11:
Per discussion in #4301, add a `sync = <time>` option to the KVS, so that the primary kvs namespace can be checkpointed on a regular basis.

### DIFF
--- a/src/modules/kvs/kvs.c
+++ b/src/modules/kvs/kvs.c
@@ -1044,6 +1044,7 @@ static void kvstxn_apply (kvstxn_t *kt)
 done:
     if (errnum == 0) {
         json_t *names = kvstxn_get_names (kt);
+        int internal_flags = kvstxn_get_internal_flags (kt);
         int count;
         if ((count = json_array_size (names)) > 1) {
             int opcount = 0;
@@ -1051,8 +1052,10 @@ done:
             flux_log (ctx->h, LOG_DEBUG, "aggregated %d transactions (%d ops)",
                       count, opcount);
         }
-        setroot (ctx, root, kvstxn_get_newroot_ref (kt), root->seq + 1);
-        setroot_event_send (ctx, root, names, kvstxn_get_keys (kt));
+        if (!(internal_flags & KVSTXN_INTERNAL_FLAG_NO_PUBLISH)) {
+            setroot (ctx, root, kvstxn_get_newroot_ref (kt), root->seq + 1);
+            setroot_event_send (ctx, root, names, kvstxn_get_keys (kt));
+        }
     } else {
         fallback = kvstxn_fallback_mergeable (kt);
 
--- a/src/modules/kvs/kvs.c
+++ b/src/modules/kvs/kvs.c
@@ -1570,7 +1570,7 @@ static void relaycommit_request_cb (flux_t *h, flux_msg_handler_t *mh,
         goto error;
     }
 
-    if (kvstxn_mgr_add_transaction (root->ktm, name, ops, flags) < 0) {
+    if (kvstxn_mgr_add_transaction (root->ktm, name, ops, flags, 0) < 0) {
         flux_log_error (h, "%s: kvstxn_mgr_add_transaction",
                         __FUNCTION__);
         goto error;
@@ -1651,7 +1651,8 @@ static void commit_request_cb (flux_t *h, flux_msg_handler_t *mh,
         if (kvstxn_mgr_add_transaction (root->ktm,
                                         treq_get_name (tr),
                                         ops,
-                                        flags) < 0) {
+                                        flags,
+                                        0) < 0) {
             flux_log_error (h, "%s: kvstxn_mgr_add_transaction",
                             __FUNCTION__);
             goto error;
@@ -1749,7 +1750,8 @@ static void relayfence_request_cb (flux_t *h, flux_msg_handler_t *mh,
         if (kvstxn_mgr_add_transaction (root->ktm,
                                         treq_get_name (tr),
                                         treq_get_ops (tr),
-                                        treq_get_flags (tr)) < 0) {
+                                        treq_get_flags (tr),
+                                        0) < 0) {
             flux_log_error (h, "%s: kvstxn_mgr_add_transaction",
                             __FUNCTION__);
             goto error;
@@ -1856,7 +1858,8 @@ static void fence_request_cb (flux_t *h, flux_msg_handler_t *mh,
             if (kvstxn_mgr_add_transaction (root->ktm,
                                             treq_get_name (tr),
                                             treq_get_ops (tr),
-                                            treq_get_flags (tr)) < 0) {
+                                            treq_get_flags (tr),
+                                            0) < 0) {
                 flux_log_error (h, "%s: kvstxn_mgr_add_transaction",
                                 __FUNCTION__);
                 goto error;
--- a/src/modules/kvs/kvs.c
+++ b/src/modules/kvs/kvs.c
@@ -29,11 +29,13 @@
 #include "src/common/libutil/monotime.h"
 #include "src/common/libutil/tstat.h"
 #include "src/common/libutil/timestamp.h"
+#include "src/common/libutil/errprintf.h"
 #include "src/common/libkvs/treeobj.h"
 #include "src/common/libkvs/kvs_checkpoint.h"
 #include "src/common/libkvs/kvs_txn_private.h"
 #include "src/common/libkvs/kvs_util_private.h"
 #include "src/common/libcontent/content.h"
+#include "src/common/libutil/fsd.h"
 
 #include "waitqueue.h"
 #include "cache.h"
@@ -43,6 +45,7 @@
 #include "kvstxn.h"
 #include "kvsroot.h"
 #include "kvs_wait_version.h"
+#include "kvs_checkpoint.h"
 
 /* heartbeat_sync_cb() is called periodically to manage cached content
  * and namespaces.  Synchronize with the system heartbeat if possible,
@@ -73,6 +76,7 @@ struct kvs_ctx {
     bool events_init;            /* flag */
     const char *hash_name;
     unsigned int seq;           /* for commit transactions */
+    kvs_checkpoint_t *kcp;
     struct list_head work_queue;
 };
 
@@ -89,6 +93,8 @@ static void transaction_prep_cb (flux_reactor_t *r, flux_watcher_t *w,
 static void transaction_check_cb (flux_reactor_t *r, flux_watcher_t *w,
                                   int revents, void *arg);
 static void start_root_remove (struct kvs_ctx *ctx, const char *ns);
+static void work_queue_check_append (struct kvs_ctx *ctx,
+                                     struct kvsroot *root);
 static void kvstxn_apply (kvstxn_t *kt);
 
 /*
@@ -103,9 +109,17 @@ static void kvs_ctx_destroy (struct kvs_ctx *ctx)
         flux_watcher_destroy (ctx->prep_w);
         flux_watcher_destroy (ctx->check_w);
         flux_watcher_destroy (ctx->idle_w);
+        kvs_checkpoint_destroy (ctx->kcp);
         free (ctx);
         errno = saved_errno;
     }
+}
+
+static void work_queue_check_append_wrapper (struct kvsroot *root,
+                                             void *arg)
+{
+    struct kvs_ctx *ctx = arg;
+    work_queue_check_append (ctx, root);
 }
 
 static struct kvs_ctx *kvs_ctx_create (flux_t *h)
@@ -138,6 +152,13 @@ static struct kvs_ctx *kvs_ctx_create (flux_t *h)
             goto error;
         flux_watcher_start (ctx->prep_w);
         flux_watcher_start (ctx->check_w);
+        ctx->kcp = kvs_checkpoint_create (h,
+                                          NULL, /* set later */
+                                          0.0,  /* default 0.0, set later */
+                                          work_queue_check_append_wrapper,
+                                          ctx);
+        if (!ctx->kcp)
+            goto error;
     }
     ctx->transaction_merge = 1;
     list_head_init (&ctx->work_queue);
@@ -2704,6 +2725,30 @@ error:
         flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
 }
 
+static void config_reload_cb (flux_t *h,
+                              flux_msg_handler_t *mh,
+                              const flux_msg_t *msg,
+                              void *arg)
+{
+    struct kvs_ctx *ctx = arg;
+    const flux_conf_t *conf;
+    const char *errstr = NULL;
+    flux_error_t error;
+
+    if (flux_conf_reload_decode (msg, &conf) < 0)
+        goto error;
+    if (kvs_checkpoint_reload (ctx->kcp, conf, &error) < 0) {
+        errstr = error.text;
+        goto error;
+    }
+    if (flux_respond (h, msg, NULL) < 0)
+        flux_log_error (h, "error responding to config-reload request");
+    return;
+ error:
+    if (flux_respond_error (h, msg, errno, errstr) < 0)
+        flux_log_error (h, "error responding to config-reload request");
+}
+
 /* see comments above in event_subscribe() regarding event
  * subscriptions to kvs.namespace */
 static const struct flux_msg_handler_spec htab[] = {
@@ -2741,8 +2786,21 @@ static const struct flux_msg_handler_spec htab[] = {
                             setroot_pause_request_cb, FLUX_ROLE_USER },
     { FLUX_MSGTYPE_REQUEST, "kvs.setroot-unpause",
                             setroot_unpause_request_cb, FLUX_ROLE_USER },
+    { FLUX_MSGTYPE_REQUEST, "kvs.config-reload", config_reload_cb, 0 },
     FLUX_MSGHANDLER_TABLE_END,
 };
+
+static int process_config (struct kvs_ctx *ctx)
+{
+    flux_error_t error;
+    if (kvs_checkpoint_config_parse (ctx->kcp,
+                                     flux_get_conf (ctx->h),
+                                     &error) < 0) {
+        flux_log (ctx->h, LOG_ERR, "%s", error.text);
+        return -1;
+    }
+    return 0;
+}
 
 static void process_args (struct kvs_ctx *ctx, int ac, char **av)
 {
@@ -2897,6 +2955,8 @@ int mod_main (flux_t *h, int argc, char **argv)
         flux_log_error (h, "error creating KVS context");
         goto done;
     }
+    if (process_config (ctx) < 0)
+        goto done;
     process_args (ctx, argc, argv);
     if (ctx->rank == 0) {
         struct kvsroot *root;
@@ -2940,6 +3000,8 @@ int mod_main (flux_t *h, int argc, char **argv)
             flux_log_error (h, "event_subscribe");
             goto done;
         }
+
+        kvs_checkpoint_update_root_primary (ctx->kcp, root);
     }
     if (flux_msg_handler_addvec (h, htab, ctx, &handlers) < 0) {
         flux_log_error (h, "flux_msg_handler_addvec");
@@ -2953,6 +3015,7 @@ int mod_main (flux_t *h, int argc, char **argv)
         flux_log_error (h, "error starting heartbeat synchronization");
         goto done;
     }
+    kvs_checkpoint_start (ctx->kcp);
     if (flux_reactor_run (flux_get_reactor (h), 0) < 0) {
         flux_log_error (h, "flux_reactor_run");
         goto done;
--- a/src/modules/kvs/kvs_checkpoint.c
+++ b/src/modules/kvs/kvs_checkpoint.c
@@ -1,0 +1,233 @@
+/************************************************************\
+ * Copyright 2019 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdbool.h>
+#include <ctype.h>
+#include <errno.h>
+#include <jansson.h>
+#include <flux/core.h>
+
+#include "src/common/libutil/errprintf.h"
+#include "src/common/libutil/fsd.h"
+
+#include "kvs_checkpoint.h"
+#include "kvsroot.h"
+
+struct kvs_checkpoint {
+    flux_t *h;
+    struct kvsroot *root_primary;
+    double checkpoint_period;   /* in seconds */
+    flux_watcher_t *checkpoint_w;
+    kvs_checkpoint_txn_cb txn_cb;
+    void *txn_cb_arg;
+    int last_checkpoint_seq;
+};
+
+static int checkpoint_period_parse (const flux_conf_t *conf,
+                                    flux_error_t *errp,
+                                    double *checkpoint_period)
+{
+    flux_error_t error;
+    const char *str = NULL;
+
+    if (flux_conf_unpack (conf,
+                          &error,
+                          "{s?{s?s}}",
+                          "kvs",
+                          "checkpoint-period", &str) < 0) {
+        errprintf (errp,
+                   "error reading config for kvs: %s",
+                   error.text);
+        return -1;
+    }
+
+    if (str) {
+        if (fsd_parse_duration (str, checkpoint_period) < 0) {
+            errprintf (errp,
+                       "invalid checkpoint-period config: %s",
+                       str);
+            return -1;
+        }
+    }
+
+    return 0;
+}
+
+int kvs_checkpoint_config_parse (kvs_checkpoint_t *kcp,
+                                 const flux_conf_t *conf,
+                                 flux_error_t *errp)
+{
+    if (kcp) {
+        double checkpoint_period = kcp->checkpoint_period;
+        if (checkpoint_period_parse (conf, errp, &checkpoint_period) < 0)
+            return -1;
+        kcp->checkpoint_period = checkpoint_period;
+    }
+    return 0;
+}
+
+int kvs_checkpoint_reload (kvs_checkpoint_t *kcp,
+                           const flux_conf_t *conf,
+                           flux_error_t *errp)
+{
+    if (kcp) {
+        double checkpoint_period = kcp->checkpoint_period;
+        if (checkpoint_period_parse (conf,
+                                     errp,
+                                     &checkpoint_period) < 0)
+            return -1;
+
+        if (checkpoint_period != kcp->checkpoint_period) {
+            kcp->checkpoint_period = checkpoint_period;
+            flux_watcher_stop (kcp->checkpoint_w);
+
+            if (kcp->root_primary
+                && kcp->checkpoint_period > 0.0) {
+                flux_timer_watcher_reset (kcp->checkpoint_w,
+                                          kcp->checkpoint_period,
+                                          kcp->checkpoint_period);
+                flux_watcher_start (kcp->checkpoint_w);
+            }
+        }
+    }
+    return 0;
+}
+
+
+static void checkpoint_cb (flux_reactor_t *r,
+                           flux_watcher_t *w,
+                           int revents,
+                           void *arg)
+{
+    kvs_checkpoint_t *kcp = arg;
+    char name[64];
+    json_t *ops = NULL;
+
+    /* if no changes to root since last checkpoint-period, do
+     * nothing */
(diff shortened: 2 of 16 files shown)


garlick:
On a quick skim, it looks like the timer wakes up even when there has been no kvs activity, and also that it commits actual data to the KVS each time it syncs.  I don't think either of those things are desirable or necessary.

Maybe a timer is the wrong approach - could the commit logic check how long it's been since the last sync and just do one if periodic sync is configured AND it's been long enough?

chu11:
> On a quick skim, it looks like the timer wakes up even when there has been no kvs activity, and also that it commits actual data to the KVS each time it syncs. I don't think either of those things are desirable or necessary.

Yup, it does wake up if there hasn't been any activity, but it maintains the last sequence number of the root so that it won't do another commit if unnecessary.

```
    /* if no changes to root since last sync, don't do it */
    if (ctx->sync_last_rootseq == root->seq)
        return;
```

> Maybe a timer is the wrong approach - could the commit logic check how long it's been since the last sync and just do one if periodic sync is configured AND it's been long enough?

Hmmm, that's an interesting idea.  Perhaps if there hasn't been a checkpoint in awhile, just add `FLUX_KVS_SYNC` to that particular commit?  It's worth trying out, see if it's cleaner.

garlick:
It's probably a non starter to have the code periodically commit something, especially something _different_ each time since that contributes to content.sqlite growth.

garlick:
If a sync fails, something is really going wrong with the system or the backing store, and KVS data is in jeopardy with no real recovery options, so it probably would be fine IMHO to fail a commit in that case.

garlick:
Oh, duh - the way I proposed (not using a timer) could leave dirty data unsynced longer the configured time if another commit doesn't come along.  Sorry about that, your intuition was right that a timer is the way to go!

The spurious wakeup can be avoided by not starting the timer watcher until a commit comes in.  E.g. each commit checks if timer is active, and if not , sets the time and starts it.  Timer expires, sync happens, and the system could be completely idle until another commit comes in and starts the timer again.

You're the expert on the best way to do this inside the KVS, so take this with a grain of salt but instead of a flag could we just have the timer post a regular commit with no ops (empty txn) and the FLUX_KVS_SYNC flag, and then just ensure that the state machine does the right thing with that if it doesn't already?  It seems like it would be nice to allow that anyway, from outside the KVS module.

chu11:
>  post a regular commit with no ops (empty txn) and the FLUX_KVS_SYNC flag

I was just thinking about this.  For some reason, I felt that a txn must not be empty.  But if it is empty, that should totally be allowed.  Then this would be easy and I can perhaps add some "state machine jumps" to skip states that are no longer important.

garlick:
Could you just do (on every commit):
```c
if (!timer_flag) {
    flux_timer_watcher_reset (w, sync_time, 0.);
    flux_timer_watcher_start (w);
    timer_flag = true;
}
```
Then in the timer callback, start the sync and set `timer_flag = false`.

So you'd get approximately a sync every `sync_time` as long as there is activity.

If the sync time is less than some threshold we could always bump it up and document this.   Say 1m?

chu11:
> Could you just do (on every commit):

You mean not do the checks I was referring to above (or just not do the `strcmp` for the primary namespace)?  I suppose, although we'd probably be activating the timer way more often than we want to?

garlick:
Oh yes, sorry, those checks would be needed.  Commits arriving together are normally combined so maybe those checks would effectively be rate limits to the batch timeout?  Can't imagine they would amount to much overhead in any case, but famous last words!

src/modules/kvs/kvs.c
```diff
@@ -1223,6 +1229,68 @@ static int heartbeat_root_cb (struct kvsroot *root, void *arg)
...
+    flux_kvs_txn_t *txn = NULL;
+    flux_future_t *f = NULL;
+
+    /* prior sync not completed, skip this turn in the sync */
+    if (ctx->sync_f)
+        return;
+
+    /* N.B. no actual changes in this transaction */
+    if (!(txn = flux_kvs_txn_create ())) {
+        flux_log_error (ctx->h, "%s: flux_kvs_txn_create", __FUNCTION__);
```
garlick, line 1270: The log messages in this function are all equally unlikely and don't provide much info.  Maybe all the failure cases could use one better message like "periodic checkpoint setup failed: ..."

garlick (commented):
Nice - a few minor comments for you.

src/modules/kvs/kvs.c
```diff
@@ -1223,6 +1229,68 @@ static int heartbeat_root_cb (struct kvsroot *root, void *arg)
...
+    root = kvsroot_mgr_lookup_root (ctx->krm, KVS_PRIMARY_NAMESPACE);
+    assert (root);
+
+    /* if there are no commits to be processed, shut down timer, we'll
+     * wake it up next time the primary KVS namespace has work to be
+     * done on it */
+    if (kvstxn_mgr_ready_transaction_count (root->ktm) == 0) {
+        flux_watcher_stop (ctx->sync_w);
+        ctx->sync_timer_running = false;
+    }
```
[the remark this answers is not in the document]
chu11: Hmmmm, ok, I see your point, instead of the root check here and the queue-ing checks everywhere else, effectively do some timer math to see if the timer should restarted after every commit.  That might be simpler, lemme give it a try.

As an aside, I realized I do have a corner case, in that the current code guarantees to checkpoint within "sync time" after the first commit (or if the KVS has been non-busy, the first commit in awhile in which the timer is woken up), which is probably not correct.  It should be "sync time" after the KVS started (or sync immediately if there hasn't been a sync in > "sync time").

Also, it occurred to me, to remove the wasteful "strcmp" on every namespace string, I could just add a special flag in the `kvsroot` struct to indicate "i am the primary namespace".  I think that'd be useful.

chu11:
re-pushed

- re-did the sync logic, doing:
  - one shot timers
  - start timer after commit completes vs when commit queued up
  - fix logic to ensure timer launched earlier if necessary
- add optimization commit to remove constnat `strcmp()` on the KVS primary namespace name
- add flux-config-kvs manpage, which i just completely forgot
- minor cleanups per comments above

src/modules/kvs/kvs.c
```diff
@@ -903,6 +911,28 @@ static void kvstxn_apply_cb (flux_future_t *f, void *arg)
...
+        && !ctx->sync_started) {
+        /* monotime_since() returns in milliseconds, have to adjust to
+         * seconds */
+        double t = monotime_since (ctx->last_sync) / 1000.0;
+        double after = 0.0;
+        if (t < ctx->sync)
+            after = ctx->sync - t;
+        /* else we haven't synced in a long time, so set timer
+         * to set off immediately (i.e. after = 0.0)
+         */
```
garlick, line 928: This seems a bit weird to force a sync on the first commit after the system has been idle, when this commit is likely to be immediately followed by others if it 's a job (the common case).  Maybe we could describe the sync parameter as "the maximum  length of time dirty KVS data will remain in memory before being flushed to the backing store" and then just set the timer to the fixed configured value?

That seems more intuitive to me.

chu11: Is the comment perhaps misleading?  I initialize `last_sync` to when the module was loaded, so the first sync should occur "sync period" time after the first commit.

garlick: Maybe I'm confused, or maybe I wasn't clear - I thought `t` was tracking the time since the last sync and would shorten the timer according to how long it's been since the last sync.   So for example the sync period is 30m and the system is idle for a day and then someone runs a job, the first commit would trigger a sync.  But since nothing else has been committed, there really isn't any increased urgency to do the sync right then.

If instead we think about the sync time period as placing an upper bound on how long after a change we wait to checkpoint it, then we would just start the timer with a fixed duration when ever something changes.

chu11: ok, i see your point now.  So instead of "checkpoint every X time", you're thinking it should be "checkpoint if anything older than X time"

chu11:
re-pushed, tweaking sync logic per @garlick comments above.  Tweaked KVS manpage language as a result.  As fixed up some manpage generation / spell check fails that I messed up.

src/modules/kvs/kvs.c
```diff
@@ -1223,6 +1243,47 @@ static int heartbeat_root_cb (struct kvsroot *root, void *arg)
...
 
+static void sync_continuation (flux_future_t *f, void *arg)
+{
+    struct kvs_ctx *ctx = arg;
+
+    if (flux_rpc_get (f, NULL) < 0)
+        flux_log_error (ctx->h, "%s: flux_rpc_get", __FUNCTION__);
+    flux_future_destroy (f);
+    ctx->sync_f = NULL;
+    ctx->sync_started = false;
```
garlick, line 1254: Should this be set false in `sync_cb()`?  If another commit completes before an in-progress sync does, then it wouldn't start the timer again, and could sit there until another commit comes along.

chu11: I don't think so.  The timer could be restarted before the last sync completes.  Here's the way I think about it.  Here's the transaction queue where a bunch of commits come in and are put on the queue.

`putA -> putB -> putC -> NULL`

`putA` finishes, leading to the sync timer to be started.  At some point in the future the `sync_cb()` will be called and add a transaction to the end of the queue to do the sync.  So perhaps its looks like this at that point (`putB` has completed before the `sync` was added):

`putC -> putD -> putE -> sync -> NULL`

if we set `sync_started` to false in `sync_cb()`, that means the timer can be restarted after `putC` completes, which is not what we want.  We'd want the sync timer to be restarted on a `putF` that will come later after the sync.

But your comment does make me think, is it possible a `putF` could complete before the `sync_started` flag is set to false in the continuation?  The odds are low, but I'm not sure if it could be guaranteed.  There's a lot of reactor stuff going on in the background.

I suppose the safe thing is `sync_started` would maybe have to be set to false when the sync is simply completed, perhaps via some callback mechanism.

These corner cases are tough to think about wrap around, and I definitely had to ponder them a bit.  Is it worth considering going back to something closer to the initial implementation I proposed?  Where its simply a timer that sets off every X time?  It largely avoids thinking about a lot of these corner cases.

garlick:
I think it is safe as long as 1) timer is running or started when each
commit finishes, and 2) timer is stopped when sync is started.  When a sync
starts it is capturing the root ref at that moment, so it needs to be
possible for a commit that occurs right after that to start the timer again.

Does that make sense?  Sorry all I have is my phone this weekend and so I
might be missing some detail I would catch if I could review code....

chu11:
re-pushed, missed a "&&" chain in a test

src/modules/kvs/kvs.c
```diff
@@ -2882,6 +2975,8 @@ int mod_main (flux_t *h, int argc, char **argv)
         flux_log_error (h, "error creating KVS context");
         goto done;
     }
+    if (process_config (ctx) < 0)
```
garlick, line 2978: Maybe for a future PR: support `kvs.config-reload` so this tunable can be changed on the fly.

src/modules/kvs/kvs.c
```diff
@@ -1053,6 +1072,7 @@ static void kvstxn_apply (kvstxn_t *kt)
         }
         setroot (ctx, root, kvstxn_get_newroot_ref (kt), root->seq + 1);
         setroot_event_send (ctx, root, names, kvstxn_get_keys (kt));
+        start_sync_timer (ctx, root);
```
garlick, line 1075: By chance is the timer re-armed here by the sync (empty commit) itself?  I'm noting that even when the system should be idle I'm getting repeated syncs as though the timer were periodic.

chu11: hmm that makes sense, although I'm not sure how this isn't being caught by tests.

chu11: ahhh, my tests only ensure that the rootref hasn't changed, which doesn't mean syncs won't happen.  I assume you're putting in printfs or something and notice the sync callback being called more times than it should?

We need another flag to ensure this doesn't happen, one for the sync timer and one for a sync being active.

But b/c of the racy thing I mentioned above, (a commit could complete before the `sync_continuation()` is called), I think this would involve more hackery than I'd like.  Lets talk on the coffee time about this.

This has certainly exceeded the complexity I originally imagined this PR to be, I'm not sure if you had other ideas as well.

garlick: OK.  No coffee time today though (it's tuesday).  Did you want to chat now?  LMK when you're available.

They only race I can imagine is maybe if the sync commit took longer than a concurrent timer.

I didn't entirely understand  how the empty commit is handled, but it would seem like it must be publishing the root if it's also starting the timer since that's right next to it in the code?  It probably shouldn't be doing that either, so maybe it needs to short circuit out of there earlier?

chu11:
A side note, @garlick and I chatted and we realized perhaps the problem is that we are using FLUX_KVS_SYNC here.  Effectively,
we are trying to manage the "sync" from something that is sort of outside of the KVS transactions's purview.

Instead, will make an attempt to try and implement as a "feature" within the KVS transaction manager.

If that fails, we may fall back and just do a normal cron-like-timer.

chu11:
Re-pushed, altering the logic of how the syncs are done.

1) In the KVS transaction manager, add a new function `kvstxn_mgr_prepend_sync()` that will fake prepend a "sync transaction" onto the head of the transaction queue.  Why prepend instead of append?  It's to ensure the "sync" is done ASAP, instead of being done at some unknown time in the future if there is a large queue of commits waiting to be processed.

This was important so we don't have to deal with the complexity from before of "what if there's a sync transaction already on the transaction queue".  It's now the next transaction all the time, keeping this simpler.

We also add `kvstxn_is_sync()` as a helper function to determine if the transaction is a special "sync" one initiated by the KVS module.

2) `sync_cb()` now calls `kvstxn_mgr_prepend_sync()` instead of creating a sync commit via `flux_kvs_commit()`.  Why is this important?  B/c now we don't need wait for a future to know the sync is complete.  The KVS transaction code will know when a sync is done and it is complete (using `kvstxn_is_sync()`).  So we don't need to deal with the complexity of waiting for the future to complete to know the sync is done.

3) We also have the ability to identify when a special KVS module sync occurs, so that when one occurs, we don't have to perform a `setroot()` or `setroot_event_send()` after the transaction is complete.  We know this by using the `kvstxn_is_sync()` function.

4) The timer is now activated / whenever a commit is done and sync timer is inactive AND a sync is not in progress.  The flags / logic of this is easier now that we don't need to deal with a future and deal with syncs at the back end of the queue.

5) put the cleanups in PR #4389, so hopefully PR is easier to review as a result.

chu11:
rebased and re-pushed, fixed a stupid thing I noticed

```
diff --git a/src/modules/kvs/kvs.c b/src/modules/kvs/kvs.c
index 1708dd9..f206e40 100644
--- a/src/modules/kvs/kvs.c
+++ b/src/modules/kvs/kvs.c
@@ -2800,8 +2800,7 @@ static int process_config (struct kvs_ctx *ctx)
 
     if (syncstr) {
         if (fsd_parse_duration (syncstr, &ctx->sync) < 0) {
-            flux_log_error (ctx->h, "invalid sync config: %s", syncstr);
-            errno = EINVAL;
+            flux_log (ctx->h, LOG_ERR, "invalid sync config: %s", syncstr);
             return -1;
         }
     }


```

chu11:
since we haven't merged yet, went ahead and added the `kvs.config-reload` support.  It's a fixup for the time being, since its a tad on the non-simple side.  I may have "over handled" corner cases.  Could have perhaps just let certain situations be.

garlick:
Quick comment on the dynamic config bit:
- if you're failing the config update, make sure the user gets a human readable message (which will be seen on `flux config reload` stderr)
- I think you could just keep it simple:  stop current timer, start new one with new times (or not)

chu11:
rebased, re-pushed, squashed, fixing up a few of the things mentioned by @garlick above

src/modules/kvs/kvstxn.c
```diff
@@ -1268,19 +1276,59 @@ int kvstxn_mgr_add_transaction (kvstxn_mgr_t *ktm,
     return 0;
 }
 
+int kvstxn_mgr_prepend_sync (kvstxn_mgr_t *ktm, unsigned int seq)
+{
+    kvstxn_t *kt = NULL;
+    char *name = NULL;
+
+    /* N.B. we "fake" prepend to the ready list by storing the sync
+     * transactino in a variable in the kvstxn_mgr_t.  If we really
```
garlick, line 1285: typo "transactino"

garlick (commented):
Here are a few comments.

I'm not sure if it's the right thing but I'll throw it out there:  would it make sense to create a sync.c source file for this stuff, just to keep it contained?

src/modules/kvs/kvs.c
```diff
@@ -2689,6 +2752,140 @@ static void setroot_unpause_request_cb (flux_t *h, flux_msg_handler_t *mh,
...
+
+    if (flux_conf_unpack (conf,
+                          &err,
+                          "{s?{s?s}}",
+                          "kvs",
+                          "sync", &syncstr) < 0) {
+        snprintf (errbuf,
+                  errbufsize,
+                  "error reading config for kvs: %s",
+                  err.text);
```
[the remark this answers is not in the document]
chu11: ahhh i happened to copy & paste from `resource`, perhaps i should update there too while I'm at it.

src/modules/kvs/kvs.c
```diff
@@ -73,6 +74,11 @@ struct kvs_ctx {
     bool events_init;            /* flag */
     const char *hash_name;
     unsigned int seq;           /* for commit transactions */
+    double sync;                /* in seconds */
+    flux_watcher_t *sync_w;
+    bool sync_timer_started;
+    bool sync_txn_submitted;
```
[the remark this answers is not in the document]
chu11: ahh good catch, i don't allow two "sync prepends" to the kvstxn queue, so it literally would do nothing if a second were attempted to be added by the user.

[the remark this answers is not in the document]
chu11: oh wait ... this was there to catch a corner case.  a "sync prepend" may occur while a commit is in progress, so the checkoutpoint would occur right after the current commit in progress is complete.

If we removed this flag, the sync timer could go off for the commit right before the checkpoint is done.  I guess we could allow it?   But it seems sort of silly to me to allow this.  We'd want the sync timer to go off for the commit right after the sync/checkpoint.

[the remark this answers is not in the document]
garlick: I'm having a hard time following this, and that's not a good indication that someone coming along later will be able to quickly understand it enough to make changes.  I'm trying to think what would cordon this feature off from the rest of the complex kvs stuff, so it could be grokked in isolation.

What if we just create a new source module sync.c and each time we _complete_ a transaction on the root namespace, we call a function like `sync_setroot()`.  Then in that module (without hooking into any other part of the KVS except maybe a configuration callback) we implement the content flush + checkpoint and timer?  Maybe sync.c could contain functions that are also used by the FLUX_KVS_SYNC flag handling in the commit code?

Too much of a do-over?

src/modules/kvs/kvs.c
```diff
@@ -2689,6 +2752,140 @@ static void setroot_unpause_request_cb (flux_t *h, flux_msg_handler_t *mh,
...
+                    struct kvsroot *root;
+                    root = kvsroot_mgr_lookup_root (ctx->krm,
+                                                    KVS_PRIMARY_NAMESPACE);
+                    assert (root);
+                    run_sync_timer (ctx);
+                }
+            }
+        }
+
+        ctx->sync = sync;
```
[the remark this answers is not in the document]
chu11: not sure I can simplify it that much, but you're right that it could be simplified.  The above did have some dead code (`kvsroot_mgr_lookup_root()` looks up something that is not used).  If I used `start_sync_timer_if_needed()`, I'd have to do the lookup.

I did clean it up to the following.

```
    if (ctx->rank == 0 && sync != ctx->sync) {
        ctx->sync = sync;
        if (!ctx->sync_txn_submitted) {
            if (ctx->sync_timer_started) {
                flux_watcher_stop (ctx->sync_w);
                if (ctx->sync > 0.0) {
                    flux_timer_watcher_reset (ctx->sync_w, ctx->sync, 0.);
                    flux_watcher_start (ctx->sync_w);
                }
                else /* ctx->sync == 0.0 */
                    ctx->sync_timer_started = false;
            }
            else {
                if (ctx->primary_commit_since_last_sync)
                    start_sync_timer (ctx);
            }
        }
    }
```

src/modules/kvs/kvs.c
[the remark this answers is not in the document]
chu11: another thought, would it be worthwhile to go back to the earlier timer based method?  Simply wake up every X time and perform a sync using FLUX_KVS_SYNC?  So much of the complexity is eliminated, but at the cost of an occasional timer wakeup and the setroot event that goes out every once in awhile.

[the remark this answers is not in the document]
garlick: It  seems wrong to broadcast the setroot event when  the KVS content has not changed.

Doing this standalone really shouldn't be that bad - in fact, couldn't it just subscribe to the setroot event and use that to start the timer?

chu11:
talked to @garlick, we both agree that this has become more complicated than either of as initially imagined.  So we're going to backtrack.  We're going to go with the timer loop idea that I originally proposed.  But we'll create a special internal flag like KVSTXN_NO_PUBLISH or something like that, where it will not publish the setroot event after a transaction has completed.

we'll come back to this someday if this ends up being a performance issue, which we both agree is unlikely if the user sticks to sensible checkpoint periods (like 30 minutes), but can be a problem if the period goes down.

chu11:
rebased & re-pushed

- renamed "sync" config to "checkpoint-period"
- went back to timer loop mechanism, wakeup once every X configured time
- support a NO_PUBLISH flag in `kvstxn` to avoid sending out `setroot`  events after an internal checkpoint is done.

src/modules/kvs/kvs.c
```diff
@@ -1239,6 +1245,60 @@ static void heartbeat_sync_cb (flux_future_t *f, void *arg)
     flux_future_reset (f);
 }
 
+static void checkpoint_period_cb (flux_reactor_t *r,
```
garlick, line 1248: Any reason this stuff was not moved to a new source file?   It seems like it could be fairly self-contained and kvs.c is already really long.

chu11: it seemed just below the threshold for a new file, but it's close.  I'll splice out into a new file.

chu11:
re-pushed, splicing all the checkpoint-period stuff into its own file.  There were two tricky parts

1) the need to update the "work queue" of roots with work in them.  So I had to add a callback function to deal with that.

2) how to deal with the fact this only matters in rank == 0 w/o sticking a ton of ```if (ctx->rank == 0)``` everywhere.

I just made it so the checkpoint-period "struct" is set to NULL on rank != 0, and all of the checkpoint-period functions
do nothing when the context pointer is NULL.

src/modules/kvs/kvstxn.c
```diff
@@ -55,7 +55,8 @@ struct kvstxn {
     json_t *ops;
     json_t *keys;
     json_t *names;
-    int flags;
+    int flags;                  /* kvs flags from request caller */
+    int api_flags;       /* special kvstxn api flags */
```
garlick, line 59: I'm not sure the `api_` prefix is communicating that it's an internal flag.  Would `internal_flags` be better?

src/modules/kvs/kvs.c
```diff
@@ -138,6 +152,16 @@ static struct kvs_ctx *kvs_ctx_create (flux_t *h)
...
+        /* ctx->kcp will not be created on rank != 0, all kvs_checkpoint
+         * functions will do nothing when ctx->kcp is NULL.
+         */
+        ctx->kcp = kvs_checkpoint_create (h,
+                                          NULL, /* set later */
+                                          0.0,  /* default 0.0, set later */
+                                          work_queue_check_append_wrapper,
+                                          ctx);
+        if (!ctx->kcp)
+            goto error;
```
garlick, line 164: Since this code is inside an `if (ctx->rank == 0)` block, maybe the comment is not needed here?  Might be better to just document in `kvs_checkpoint.h` that each function is a no-op if called with a NULL `kvs_checkpiont_t` argument to avoid excessive rank 0 tests in the code.

src/modules/kvs/kvs.c
```diff
@@ -2703,6 +2727,30 @@ static void setroot_unpause_request_cb (flux_t *h, flux_msg_handler_t *mh,
         flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
 }
 
+static void reload_cb (flux_t *h,
```
garlick, line 2730: maybe naming this `config_reload_cb()` would be better in such a large code base as the kvs?

src/modules/kvs/kvs_checkpoint.c
```diff
@@ -0,0 +1,230 @@
...
+    kvs_checkpoint_t *kcp = arg;
+    char *name = NULL;
+    json_t *ops = NULL;
+
+    /* if no changes to root since last checkpoint-period, do
+     * nothing */
+    if (kcp->last_checkpoint_seq == kcp->root_primary->seq)
+        return;
+
+    if (asprintf (&name, "checkpoint-period.%u", kcp->root_primary->seq) < 0
```
garlick, line 123: It would probably be better to use a static buffer instead of `asprintf()` here since the string is of predictable length.

src/modules/kvs/kvs_checkpoint.c
```diff
@@ -0,0 +1,230 @@
...
+        kcp->txn_cb (kcp->root_primary, kcp->txn_cb_arg);
+
+    /* N.B. "last_checkpoint_seq" protects against unnecessary
+     * checkpointing when there is no activity in the primary KVS.
+     */
+    kcp->last_checkpoint_seq = kcp->root_primary->seq;
+
+ error:
+    free (name);
+    json_decref (ops);
```
garlick, line 157: Clobbers errno

src/modules/kvs/kvs_checkpoint.h
```diff
@@ -0,0 +1,66 @@
...
+ * txn_cb - callback after each checkpoint transaction submitted
+ * txn_cb_arg - will be passed to txn_cb
+ */
+kvs_checkpoint_t *kvs_checkpoint_create (flux_t *h,
+                                         struct kvsroot *root_primary,
+                                         double checkpoint_period,
+                                         kvs_checkpoint_txn_cb txn_cb,
+                                         void *txn_cb_arg);
+
+/* will update internal checkpoint_period setting as needed */
```
garlick, line 37: I think you can drop "will" in all the comments in this header file.

t/t1011-kvs-checkpoint-period.t
```diff
@@ -0,0 +1,165 @@
...
+            i=$((i + 1))
+        done
+        return 1
+}
+
+test_expect_success 'configure bad checkpoint-period timer in kvs' '
+        cat >kvs.toml <<EOF &&
+[kvs]
+checkpoint-period = "1Z"
+EOF
```
garlick, line 47: Suggestion: use `<<-EOF` and the hereis content can be indented wtih tab characters.

garlick (approved):
This looks great @chu11.  I think breaking the functionality out to its own source file was a good move.  At  least for a relative outsider to the code base, it's way easier to grok when it can be isolated.

I made some suggestions but they are all pretty superficial at this point.

Oh I'll set this up on my test system and give it a quick "real life" test.

src/modules/kvs/kvs_checkpoint.c
chu11: does it matter in this case?  the function returns void.

src/modules/kvs/kvstxn.c
chu11: i wasn't the biggest fan of "api_flags" either.  The issue was there is already an "internal_flags" used internally within the `kvstxn` api (internal means not externally exposed to users).  

perhaps could be called `kvstxn_flags`?  perhaps the other flags needs to renamed to `op_flags`?  lemme think about it.

src/modules/kvs/kvs_checkpoint.c
garlick: Oh sorry, I thought I was following a different call stack.  This is the timer watcher callback (duh).

Given that the function cannot return an error, maybe it would make sense to rename the goto label to "done".

src/modules/kvs/kvstxn.c
chu11: ahh i had a good idea, the "internal_flags" didn't really have to be bitmasks, they were just there as "bools" for certain conditions.  So I replaced them with a bunch of bool flags.  Then api_flags could be "internal_flags" instead.

garlick:
I set my test system up for 30m checkpoints, put something in the kvs, then after an hour ran `flux kvs dump --checkpoint - | tar tvf -` and there was my test key.  Yay!

chu11:
re-based and re-pushed with fixes given comments above.  Perhaps want to give it one more skim @garlick?   The big change is making the previously named `internal_flags` into three separate bools.

```
-    int internal_flags;
+    bool processing;            /* kvstxn is being processed */
+    bool merged;                /* kvstxn is a merger of transactions */
+    bool merge_component;       /* kvstxn is member of a merger */
```

and the previously named `api_flags` is now `internal_flags`.

garlick (approved):
Looks good after a final skim!

PR #4385, opened 2022-06-29
python: add LSF URI resolver plugin

wihobbs:
Though I've gotten this to run for `lalloc` sessions I started with Flux, I'm marking this as work-in-progress until I can test more multi-node allocations and edge cases for error handling.

What I'm most concerned about here is the string parsing logic associated with getting the compute node hostname (see L#54-58). This won't work if Flux doesn't span the entire allocation of a user's compute resources, and may not necessarily resolve broker-0 in the allocation (not sure about this, still working on checking multi-node allocations). If anyone has a better way of doing this that'd be excellent, but we tried `bjobs -json` and several variations of that to no success.

In spite of these limitations, I hope this provides something useful to users that we don't currently have.

### DIFF
--- a/src/bindings/python/flux/uri/resolvers/lsf.py
+++ b/src/bindings/python/flux/uri/resolvers/lsf.py
@@ -1,0 +1,109 @@
+###############################################################
+# Copyright 2022 Lawrence Livermore National Security, LLC
+# (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+#
+# This file is part of the Flux resource manager framework.
+# For details, see https://github.com/flux-framework.
+#
+# SPDX-License-Identifier: LGPL-3.0
+###############################################################
+
+"""
+This module supports the Flux URI resolver for IBM Spectrum LSF.
+
+It can be used like:
+  `flux uri lsf:<JOBID>`.
+
+"""
+
+import getpass
+import os
+import subprocess
+from pathlib import PurePath
+
+import yaml
+
+import flux.hostlist as hostlist
+from flux.uri import URIResolverPlugin, FluxURIResolver
+
+
+def lsf_find_compute_node(jobid):
+    """Figure out where the job is being run using YAML output from IBM Cluster Systems Manager."""
+
+    csm_path = os.getenv(
+        "CSM_ALLOCATION_QUERY", "/opt/ibm/csm/bin/csm_allocation_query"
+    )
+
+    sp = subprocess.run(
+        [csm_path, "-j", str(jobid)],
+        stdout=subprocess.PIPE,
+    )
+    try:
+        hosts = yaml.safe_load(sp.stdout.decode("utf-8"))["compute_nodes"]
+        return str(hostlist.decode(hosts).sort()[0])
+    except Exception as exc:
+        raise ValueError(
+            f"Unable to find a compute node attached to job {jobid}"
+        ) from exc
+
+
+def check_lsf_jobid(pid, jobid):
+    with open(f"/proc/{pid}/environ", encoding="utf-8") as envfile:
+        if f"LSB_JOBID={jobid}" in envfile.read():
+            return True
+    return False
+
+
+def lsf_job_pid(jobid):
+    """Resolve the pids of an LSF job using ssh to the compute node"""
+
+    sp = subprocess.run(
+        ["ps", "-Ho", "pid,comm", "-u", getpass.getuser()],
+        stdout=subprocess.PIPE,
+        check=True,
+    )
+    for line in sp.stdout.decode("utf-8").split("\n"):
+        if "flux-broker" in line:
+            pid = line.split()[0]
+            if check_lsf_jobid(pid, jobid):
+                return pid
+    raise ValueError(f"Unable to find a flux session attached to your job")
+
+
+def lsf_get_uri(hostname, jobid):
+    """
+    Resolve the URI of a Flux instance by running
+
+      `ssh <HOSTNAME> flux uri lsf:<PID>?is_compute`
+
+    and converting the result from bytecode to a string.
+    """
+    ssh_path = os.getenv("FLUX_SSH", "ssh")
+    sp = subprocess.run(
+        [ssh_path, hostname, "flux", "uri", "--remote", f"lsf:{jobid}?is_compute"],
+        stdout=subprocess.PIPE,
+        check=True,
+    )
+    return sp.stdout.decode("utf-8")
+
+
+class URIResolver(URIResolverPlugin):
+    """A URIResolver that can fetch a FLUX_URI from an LSF job id"""
+
+    def describe(self):
+        return "Get URI for a Flux instance launched under IBM Spectrum LSF"
+
+    def resolve(self, uri):
+        jobid = str(PurePath(uri.path).parts[0])
+
+        if "is_compute" in uri.query_dict:
+            uri = FluxURIResolver().resolve(f"pid:{lsf_job_pid(jobid)}")
+            return uri.remote
+
+        try:
+            hostname = lsf_find_compute_node(jobid)
+            return lsf_get_uri(hostname, jobid)
+        except OSError as exc:
+            raise ValueError(
+                f"LSF Job {jobid} doesn't seem to have a FLUX_URI"
+            ) from exc
--- a/t/t2802-uri-cmd.t
+++ b/t/t2802-uri-cmd.t
@@ -4,9 +4,13 @@ test_description='Test flux uri command'
 
 . $(dirname $0)/sharness.sh
 
+# Required to be able to test LSF resolver
+export LSB_JOBID=12345
+
 test_under_flux 2
 
 testssh="${SHARNESS_TEST_SRCDIR}/scripts/tssh"
+
 
 test_expect_success 'flux-uri -h prints list of resolvers' '
 	flux uri --help >help.out >help.out 2>&1 &&
@@ -144,6 +148,22 @@ test_expect_success 'flux-uri mock testing of slurm resolver works' '
 	test "$result" = "$FLUX_URI" &&
 	( export PATH=$(pwd):$PATH REMOTE=t SRUN_FAIL=t &&
 	  test_expect_code 1 flux uri slurm:1234 )
+'
+test_expect_success 'setup fake csm_allocation_query for mock lsf testing' '
+	cat <<-EOF >csm_allocation_query &&
+	#!/bin/sh
+	test -n "\$LSF_FAIL" && exit 4
+	echo "compute_nodes:"
+	echo " - lassen9"
+	echo " - lassen10"
+	EOF
+	chmod +x csm_allocation_query &&
+	export CSM_ALLOCATION_QUERY=$(pwd)/csm_allocation_query &&
+	export FLUX_SSH=${SHARNESS_TEST_SRCDIR}/scripts/tssh
+'
+test_expect_success 'flux-uri mock testing of lsf resolver works' '
+	result=$(flux uri --local lsf:12345) &&
+	test "$result" = "$FLUX_URI"
 '
 test_expect_success 'cleanup jobs' '
 	flux job cancelall -f &&
--- a/src/bindings/python/flux/Makefile.am
+++ b/src/bindings/python/flux/Makefile.am
@@ -47,6 +47,7 @@ nobase_fluxpy_PYTHON = \
 	uri/resolvers/jobid.py \
 	uri/resolvers/pid.py \
 	uri/resolvers/slurm.py \
+	uri/resolvers/lsf.py \
 	utils/parsedatetime/__init__.py \
 	utils/parsedatetime/parsedatetime.py \
 	utils/parsedatetime/warns.py \


grondo (commented):
Nice work! Just a few comments and questions inline. I can't really comment on the LSF usage, but calling out to a job listing utility seems reasonable here to me.

src/bindings/python/flux/uri/resolvers/lsf.py
```diff
@@ -0,0 +1,102 @@
...
+            hostname = node[1]
+    if hostname != "":
+        return hostname
+    raise ValueError("Unable to resolve the hostname of the compute job.")
+
+
+def lsf_job_pid(jobid, hostname) -> str:
+    """Resolve the pids of an LSF job using ssh to the compute node"""
+    sp = subprocess.run(
+        ["ssh", hostname, "ps", "-Hu", "`whoami`", "|", "grep", "-m1", "flux-broker"],
```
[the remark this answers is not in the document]
wihobbs: Ah. `ps -H` is used to make sure that we're getting the parent instance of any flux session running. From there if a user wants to connect to a child session, they can resolve that URI using the pid URI resolver. I'll check - maybe the command you've got would be compatible with `-H` too.

src/bindings/python/flux/uri/resolvers/lsf.py
```diff
@@ -0,0 +1,102 @@
...
+def lsf_find_compute_node(jobid) -> str:
+    """Figure out where the job is being run using JSON output from `bjobs -l`."""
+
+    """
+    To get the name of the system (lassen or sierra or nid) without the node number,
+    pull the hostname and strip off anything that's not alphabetic.
+    """
+    host_system = subprocess.run(["hostname"], capture_output=True).stdout.decode(
+        "utf-8"
+    )
```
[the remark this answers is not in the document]
jameshcorbett: I would second the ``socket.gethostname`` comment. If you find yourself ever running a process to do some simple thing in Python, do a web search first to make sure there isn't a function somewhere in the standard library to do it for you. I don't think I have ever needed to run processes simply to replicate UNIX commands.

src/bindings/python/flux/uri/resolvers/lsf.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def resolve(self, uri):
+        jobid = str(PurePath(uri.path).parts[0])
+
+        try:
+            hostname = lsf_find_compute_node(jobid)
+            return lsf_get_uri(hostname, lsf_job_pid(jobid, hostname))
+        except (ValueError, OSError):
+            pass
+        raise ValueError(f"LSF Job {jobid} doesn't seem to have a FLUX_URI")
```
jameshcorbett, line 102: I'm not sure I understand the point of suppressing an exception and then raising a different one. You lose the traceback that way, and the actual error could be completely unrelated to jobids/FLUX_URI. I would suggest either not trying to catch the exception at all or something like:
```python
try:
    ...
except (ValueError, OSError) as exc:
    raise ValueError("LSF job doesn't have FLUX_URI") from exc
```
https://stackoverflow.com/questions/24752395/python-raise-from-usage

src/bindings/python/flux/uri/resolvers/lsf.py
```diff
@@ -0,0 +1,102 @@
...
+"""
+This module supports the Flux URI resolver for IBM Spectrum LSF.
+
+It can be used like:
+  `flux uri --remote lsf:<JOBID>`.
+
+"""
+
+
+def lsf_find_compute_node(jobid) -> str:
```
jameshcorbett, line 27: As with ATS I think it's better not to add type annotations unless the project has infrastructure (e.g. mypy) to support them.

src/bindings/python/flux/uri/resolvers/lsf.py
```diff
@@ -0,0 +1,102 @@
...
+from flux.uri import FluxURIResolver, URIResolverPlugin
+
+
+"""
+This module supports the Flux URI resolver for IBM Spectrum LSF.
+
+It can be used like:
+  `flux uri --remote lsf:<JOBID>`.
+
+"""
```
jameshcorbett, line 24: Module docstrings go at the top of the file, before any code (including imports).

src/bindings/python/flux/uri/resolvers/lsf.py
grondo: This could be my fault if the practice was copied from the `slurm` jobid resolver. Apologies if so!

src/bindings/python/flux/uri/resolvers/lsf.py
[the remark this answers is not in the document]
grondo: Good point, I had forgotten that there might be a hierarchy of Flux instances. In that case, your use of `-H` here is clever.  ;-)

Also, I forgot that `ps` only prints processes in the current session (associated with the same terminal) so my suggestion won't work anyway. What you have here seems fine, though I wish there was a better way to select all processes for the calling user.

Actually, since the local and remote username must match (I assume this is required for LSF), then maybe the result of `os.getlogin()` could just be substituted, e.g.:

```suggestion
        ["ssh", hostname, "ps", "-Hu", os.getlogin(), "|", "grep", "-m1", "flux-broker"],
```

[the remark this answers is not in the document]
grondo: > if a user wants to connect to a child session, they can resolve that URI using the pid URI resolver.

To clarify, once you have connected to a Flux instance, all sub-instances within that instance can most easily be resolved with the `jobid` resolver, which is more efficient, since the URI for jobs is stashed as a `memo` event in every job that is also an instance. Also, the `jobid` resolver is the default, so you can just use `flux uri JOBID` (or pass `JOBID` directly to any command that takes a high-level URI like `flux top`, `flux proxy`, `flux pstree`, etc) :slightly_smiling_face:

src/bindings/python/flux/uri/resolvers/lsf.py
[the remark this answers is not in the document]
wihobbs: So, while thinking about this last night, I wondered if there was a python library that could support SSH. Turns out, there is (https://www.paramiko.org/) but I don't think its worth adding to the python modules used in Flux. And I'm not sure it would play nice with BlueOS systems, so definitely not worth the trouble. But this is a good note about best practices in Python (and general) development, thanks @jameshcorbett and @grondo.

src/bindings/python/flux/uri/resolvers/lsf.py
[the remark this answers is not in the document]
wihobbs: Hmm, that's interesting. I'm getting the subprocess call to do all the options with: 
```python3
    sp = subprocess.run(
        ["ssh", hostname, "ps", "-Ho", "pid,comm", "-u", os.getlogin()],
        capture_output=True,
    )
```
returns:
```python3
(Pdb) print(sp.stdout.decode("utf-8"))
   PID COMMAND
 60425 sshd
 60426   ps
 52926 sshd
 52928   flux-broker-0
 53342     bash
 52828 jsmd
```

wihobbs:
Thanks for all the comments @grondo and @jameshcorbett. I'm removing WIP because I think this is ready for a final review. The last few concerns I have are 
1. @ryanday36 gave an excellent command which gets the compute nodes in YAML, and can be parsed with the pyyaml library flux already has as a dependency. I'm not sure if this will work on non-LC systems (ORNL may put their `csm_allocation_query` in a different place. But I assume this will work on Sierra (I cannot test this) and definitely works on Lassen.
2. Code coverage - not sure how to achieve this since this code would need a running LSF job in order to be tested. Maybe add a test about the error handling?

src/bindings/python/flux/uri/resolvers/lsf.py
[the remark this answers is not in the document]
grondo: Ah, sorry. I don't know why it was failing for me. User error probably!

src/bindings/python/flux/uri/resolvers/lsf.py
```diff
@@ -0,0 +1,84 @@
...
+import subprocess
+from pathlib import PurePath
+
+import yaml
+
+from flux.uri import URIResolverPlugin
+
+
+def lsf_find_compute_node(jobid):
+    """Figure out where the job is being run using JSON output from `bjobs -l`."""
```
grondo, line 30: Is output in JSON or YAML? (the comment says JSON but the code uses `yaml.safe_load`)

grondo (commented):
Getting closer, a couple more comments inline.

For testing purposes, you could try making a mock test environment via a test `csm_allocation_query` script. It would at least give the code a sanity check.. I'm not sure we bothered doing testing of the slurm jobid resolver (I'll have to look), so this isn't critical in my mind.

src/bindings/python/flux/uri/resolvers/lsf.py
grondo: Oh, I see, this is probably just a stale comment from when you used `bjobs -l`?

wihobbs:
I think I might've found an edge case. Is `broker-0` the only broker with a URI?

```
[hobbs17@lassen10:~]$ ps -u hobbs17
   PID TTY          TIME CMD
 39697 ?        00:00:00 jsmd
 39794 ?        00:00:00 flux-broker-1
[hobbs17@lassen10:~]$ flux uri --remote pid:39794
flux-uri: ERROR: name 'broker_pid' is not defined
[hobbs17@lassen10:~]$ flux uri --local pid:39794
flux-uri: ERROR: name 'broker_pid' is not defined
```
but on `lassen9`
```
hobbs17@lassen9:~]$ ps -u hobbs17
   PID TTY          TIME CMD
 10339 ?        00:00:00 jsmd
 10445 ?        00:00:01 flux-broker-0
[hobbs17@lassen9:~]$ flux uri --remote pid:10445
ssh://lassen9/var/tmp/flux-44wR9n/local-0
```

If so, I'll have to amend to account for the case where a user is allocated nodes lassen9-14 or similar. Currently, the YAML file outputs `lassen10` first and `lassen9` last since 9>1 even though 9 is the head node. (just an example, I'm sure there are other similar edge cases)

A better way might be to do `pdsh` which can list processes across nodes, list all the flux brokers on all the nodes, and then resolve the URI using PID on the node that has `broker-0`. Or, I could just be misunderstanding something about the way this works...

grondo:
All the brokers have a URI, but the only way to resolve a flux-broker process' URI is to find a child of the broker and look for FLUX_URI (at least that is how we do it now). Unless some jobs are running, only the flux-broker-0 has children...

You could try loading the list of hosts into the flux.hostlist module and sort it numeric suffix.

grondo:
e.g. 
```console
$ flux python -c 'import flux.hostlist as hostlist; hl = hostlist.decode("lassen10,lassen9"); print (hl.sort()[0])'
lassen9

```

src/bindings/python/flux/uri/resolvers/lsf.py
jameshcorbett: I'd say it's definitely not good practice to put error handling wherever possible. Often the best thing to do is to _not_ catch exceptions, especially in lower-level (i.e. simpler) functions. Off the top of my head I would say you should only catch an exception if

- You can work around the error and still give the desired functionality. For instance, you're pinging some unreliable service, and it fails. The proper thing to do might be to catch the error and try again rather than letting the error propagate.
- The error message won't make any sense to the user of the function and so you want to change the error message and type. But in this case I think it's good to `raise ... from` because then you can see both the original error and the final error with the more intelligible message.

I think you would develop a feel for it as you write more Python, and you see cases where error messages are not helpful at all, and some that are quite helpful.

jameshcorbett: Since this is a user tool I would think about what kinds of error messages would be helpful to users, who might not know much about programming, or about Python. I can think of a few:

- jobid is not valid
- there isn't LSF on the cluster (i.e. it's Slurm or something else. probably very rare)
- there isn't a Flux instance running under the job

If you can detect one of those conditions, I think it would be nice to throw a simple error and don't try to catch it anywhere, just let it propagate and terminate Python. Otherwise, if it's something weird and nitty-gritty going wrong, like the YAML returned by CSM doesn't have the right entry, I would just let the 'normal' error happen because the true source of the issue is going to be  opaque to users no matter what, and at least if you let the normal error happen (as opposed to catching it and rethrowing  a generic "something went wrong!" error) you'll get a message that's useful for developers.

jameshcorbett: Having said so much I'm pretty sure I'm bound to have said something wrong so if someone else can correct me (@grondo ) please do.

src/bindings/python/flux/uri/resolvers/lsf.py
```diff
@@ -0,0 +1,84 @@
...
+
+    sp = subprocess.run(
```
jameshcorbett, line 32: I think it would be nice to add `check=True` to any of your `subprocess.run` calls, because that will throw an error with the captured stdout and stderr if the process exits with nonzero return code.

For instance, it makes sense to me that before trying to parse the YAML output of the CSM command, you would check whether the CSM command exited normally. That way you can spit out whatever error message CSM gave you, rather than throwing a YAML decode error.

jameshcorbett (commented):
I think it's looking good!

src/bindings/python/flux/uri/resolvers/lsf.py
```diff
@@ -0,0 +1,84 @@
...
+def lsf_get_uri(hostname, pid):
+    """
+    Resolve the URI of a Flux instance by running
+
+      `ssh <HOSTNAME> flux uri --remote pid:<PID>`
+
+    and converting the result from bytecode to a string.
+    """
+    sp = subprocess.run(
+        ["ssh", hostname, "flux", "uri", "--remote", f"pid:{pid}"], capture_output=True
```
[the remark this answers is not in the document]
jameshcorbett: I like this idea. I think it's a bit confusing though so I wish it were possible to relaunch the plugin with some additional command-line option or something to say that it's on the target host and so can find the PID and use the PID resolver (rather than doing a hostname check), but it doesn't look like that's possible given the way the URI resolver works.

The clearest thing I think would be to have some external script that would do nothing but find PIDs and URIs in one go. But adding a standalone script just for that would be overkill probably.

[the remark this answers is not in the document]
grondo: You can pass options to the resolver by using URI `query` parameters as documented in [flux.uri.uri.FluxURIResolver.resolve](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/python/autogenerated/flux.uri.uri.html#flux.uri.uri.FluxURIResolver.resolve). The query parameters should be available from the underlying [flux.uri.uri.URI](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/python/autogenerated/flux.uri.uri.html#flux.uri.uri.URI) `query_dict` dictionary.

After saying all that though, I wonder if the least confusing way to do this would be to add a separate `first-local-broker` resolver or something, which would return the URI of the first broker found on the node for the given user. Then the steps used by the `lsf` resolver would be:

 1. determine first compute node of allocation
 2. run `ssh <host> flux uri resolve first-local-broker:` (The `:` is required to hint this is a URI not a jobid)

Either approach would be fine with me.

src/bindings/python/flux/uri/resolvers/lsf.py
```diff
@@ -0,0 +1,84 @@
...
+    try:
+        return str(yaml.safe_load(sp.stdout.decode("utf-8"))["compute_nodes"][0])
+    except:
+        raise ValueError("Unable to find a compute node attached to your job.")
+
+
+def lsf_job_pid(jobid, hostname):
+    """Resolve the pids of an LSF job using ssh to the compute node"""
+    sp = subprocess.run(
+        ["ssh", hostname, "ps", "-Ho", "pid,comm", "-u", os.getlogin()],
```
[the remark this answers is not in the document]
grondo: This was my suggestion, and all my Python suggestions should be suspect :wink:. I just suggested the first thing I found online for the query "How to get the current username in Python".

Since the Python documentation says `getpass.getuser` should be prefered over `os.getlogin`, I would go with that. (With apologies to @wihobbs for the incorrect suggestion)

wihobbs:
Thanks @jameshcorbett and @grondo. After our coffee discussion today, I think I've addressed all your comments. I removed a lot of the error handling but kept it in two places:
1. The job has not yet been allocated compute resources. CSM's output on this doesn't make it's way up to the user without some help and may not be particularly useful.
2. Flux is not running on the cluster (James suggested this in an earlier comment).

This should also handle the lassen[9-14] or similar edge case thanks to Mark's suggestion.

src/bindings/python/flux/uri/resolvers/lsf.py
```diff
@@ -0,0 +1,89 @@
...
+import flux.hostlist as hostlist
+from flux.uri import URIResolverPlugin
+
+def lsf_find_compute_node(jobid):
+    """Figure out where the job is being run using YAML output from IBM Cluster Systems Manager."""
+
+    if os.getenv("CSM_ALLOCATION_QUERY"):
+        csm_path = str(os.getenv("CSM_ALLOCATION_QUERY"))
+    else:
+        csm_path = "/opt/ibm/csm/bin/csm_allocation_query"
```
jameshcorbett, line 35: The above four lines could be simplified to: `csm_path = os.getenv("CSM_ALLOCATION_QUERY", "/opt/ibm/csm/bin/csm_allocation_query")`.

jameshcorbett (approved):
Code LGTM! One minor comment below. I also think @grondo might like one or two simple tests but I will let him comment.

src/bindings/python/flux/uri/resolvers/lsf.py
```diff
@@ -0,0 +1,91 @@
...
+
+import flux.hostlist as hostlist
+from flux.uri import URIResolverPlugin
+
+
+def lsf_find_compute_node(jobid):
+    """Figure out where the job is being run using YAML output from IBM Cluster Systems Manager."""
+
+    csm_path = os.getenv(
+        "CSM_ALLOCATION_QUERY", "/opt/ibm/csm/bin/csm_allocation_query"
```
grondo, line 34: Since we can override `csm_allocation_query` now, we could add basic sanity tests to `make check` by setting `CSM_ALLOCATION_QUERY` to a script which returns known output and ensuring the "LSF" jobid resolves to what is expected. (you may also have to add a fake `ssh` script, there are examples of this in the testsuite).

This is what is done for the `slurm` uri resolver  in `t2802-uri-cmd.t`.

grondo (commented):
This LGTM, one trivial comment inline. Also mentioned how we might get some sanity tests in the testsuite for this resolver.

Also, let's retitle the PR and commit `python: add LSF URI resolver plugin` for consistency with the other uri resolver commits.

t/t2802-uri-cmd.t
```diff
@@ -145,6 +145,21 @@ test_expect_success 'flux-uri mock testing of slurm resolver works' '
...
+	echo " - lassen9"
+	echo " - lassen10"
+	EOF
+	chmod +x csm_allocation_query
+	export CSM_ALLOCATION_QUERY=$(pwd)/csm_allocation_query
+	export FLUX_SSH=${SHARNESS_TEST_SRCDIR}/scripts/tssh
+	'
+
+test_expect_success 'flux-uri mock testing of lsf resolver works' '
+	test_expect_code 0 flux uri lsf:12345
```
jameshcorbett, line 161: I think `test_expect_code 0` is the same as `test_expect_success`, but in any case I don't think you need either, I think this could just be `flux uri lsf:12345` with no change to behavior. 

How come you don't check that the URI returned is the same as $FLUX_URI?

t/t2802-uri-cmd.t
```diff
@@ -145,6 +145,22 @@ test_expect_success 'flux-uri mock testing of slurm resolver works' '
...
+test_expect_success 'setup fake csm_allocation_query for mock lsf testing' '
+	cat <<-EOF >csm_allocation_query &&
+	#!/bin/sh
+	echo "compute_nodes:"
+	echo " - lassen9"
+	echo " - lassen10"
+	EOF
+	chmod +x csm_allocation_query
+	export CSM_ALLOCATION_QUERY=$(pwd)/csm_allocation_query
+	export FLUX_SSH=${SHARNESS_TEST_SRCDIR}/scripts/tssh
```
grondo, line 157: broken `&&` chain here. Each command in a test should be chained with `&&`.

t/t2802-uri-cmd.t
```diff
@@ -145,6 +145,22 @@ test_expect_success 'flux-uri mock testing of slurm resolver works' '
...
+'
+test_expect_success 'flux-uri mock testing of lsf resolver works' '
+	result=$(flux uri --local lsf:12345)
```
grondo, line 160: Missing `&&` here.

t/t2802-uri-cmd.t
```diff
@@ -145,6 +145,22 @@ test_expect_success 'flux-uri mock testing of slurm resolver works' '
...
+	test "$result" = "$FLUX_URI" &&
+	test_expect_code 0 flux uri --local lsf:12345
```
grondo, line 162: This line could be removed since `flux uri --local lsf:1234` is already tested above.

t/t2802-uri-cmd.t
```diff
@@ -145,6 +145,22 @@ test_expect_success 'flux-uri mock testing of slurm resolver works' '
...
+'
```
grondo, line 163: Also consider adding a test for some failure cases, like invalid LSF jobid. This should raise the coverage of your code.

grondo (commented):
Some comments inline. Nice job on the tests.

grondo:
Unfortunately, the test here fails when run as a Flux job. This is because there is (at least) two independent Flux instances, the parent instance under which the job is being run, and the `test_under_flux` instance which is being used for the sharness test.

I'm not exactly sure how to fix this without some LSF way to associate processes with a job. We may need some kind of trickery in the sharness test. I'll ponder...

grondo:
@wihobbs, what do you think of this? We can verify that a flux-broker actually belongs to an LSF job by checking for a matching `LSB_JOBID` environment variable. Then we set this in the test script to ensure the mock testing picks up the correct flux-broker:

```diff
diff --git a/src/bindings/python/flux/uri/resolvers/lsf.py b/src/bindings/python/flux/uri/resolvers/lsf.py
index ef8d9a674..33dccb9f2 100644
--- a/src/bindings/python/flux/uri/resolvers/lsf.py
+++ b/src/bindings/python/flux/uri/resolvers/lsf.py
@@ -49,6 +49,13 @@ def lsf_find_compute_node(jobid):
         ) from exc
 
 
+def check_lsb_jobid(pid, jobid):
+    with open(f"/proc/{pid}/environ") as envfile:
+        if f"LSB_JOBID={jobid}" in envfile.read():
+            return True
+    return False
+
+
 def lsf_job_pid(jobid, hostname):
     """Resolve the pids of an LSF job using ssh to the compute node"""
 
@@ -61,8 +68,12 @@ def lsf_job_pid(jobid, hostname):
     )
     for line in sp.stdout.decode("utf-8").split("\n"):
         if "flux-broker" in line:
-            return line.split()[0]
-    raise ValueError(f"Unable to find a flux session on host {hostname}")
+            pid = line.split()[0]
+            if check_lsb_jobid(pid, jobid):
+                return pid
+    raise ValueError(
+        f"Unable to find a flux session for LSF job {jobid} on host {hostname}"
+    )
 
 
 def lsf_get_uri(hostname, pid):
diff --git a/t/t2802-uri-cmd.t b/t/t2802-uri-cmd.t
index 5af6d7660..730630188 100755
--- a/t/t2802-uri-cmd.t
+++ b/t/t2802-uri-cmd.t
@@ -4,6 +4,12 @@ test_description='Test flux uri command'
 
 . $(dirname $0)/sharness.sh
 
+
+#  Required for lsf resolver tests below so that the mock testing can
+#   find the correct flux-broker
+#
+export LSB_JOBID=12345
+

```

grondo:
Oh, wait, the above only works if the resolver and the PID are on the same host. Let me try again tomorrow. We may have to alter the resolver to run itself on the remote host with an option as we talked about before.

grondo:
Ok, here's a different approach. The idea is that the plugin now would

 1. Look up a compute hostname for the LSF `jobid`
 2. If successful, then run `ssh hostname flux uri resolve lsf:<jobid>?is_compute
 3. If the plugin is invoked with the `is_compute` flag, then find a local flux-broker via the `ps -H` trick, ensure it the right process by looking for matching `LSB_JOBID` in the environment, then return the result of the `pid` resolver for that process

Hopefully that makes sense. This has the side benefit of only requiring a single ssh to the compute node.

This should be considered a rough draft and needs to be tested on a real LSF system.

One other note, I removed `stderr=subprocess.PIPE` from at least one `subprocess.run` invocation, because otherwise the error message is lost for any processes that fail.

```diff
diff --git a/src/bindings/python/flux/uri/resolvers/lsf.py b/src/bindings/python/flux/uri/resolvers/lsf.py
index ef8d9a674..870978e8b 100644
--- a/src/bindings/python/flux/uri/resolvers/lsf.py
+++ b/src/bindings/python/flux/uri/resolvers/lsf.py
@@ -24,7 +24,7 @@ from pathlib import PurePath
 import yaml
 
 import flux.hostlist as hostlist
-from flux.uri import URIResolverPlugin
+from flux.uri import URIResolverPlugin, FluxURIResolver
 
 
 def lsf_find_compute_node(jobid):
@@ -49,35 +49,43 @@ def lsf_find_compute_node(jobid):
         ) from exc
 
 
-def lsf_job_pid(jobid, hostname):
+def check_lsb_jobid(pid, jobid):
+    with open(f"/proc/{pid}/environ", encoding="utf-8") as envfile:
+        if f"LSB_JOBID={jobid}" in envfile.read():
+            return True
+    return False
+
+
+def lsf_job_pid(jobid):
     """Resolve the pids of an LSF job using ssh to the compute node"""
 
-    ssh_path = os.getenv("FLUX_SSH", "ssh")
     sp = subprocess.run(
-        [ssh_path, hostname, "ps", "-Ho", "pid,comm", "-u", getpass.getuser()],
+        ["ps", "-Ho", "pid,comm", "-u", getpass.getuser()],
         stdout=subprocess.PIPE,
         stderr=subprocess.PIPE,
         check=True,
     )
     for line in sp.stdout.decode("utf-8").split("\n"):
         if "flux-broker" in line:
-            return line.split()[0]
-    raise ValueError(f"Unable to find a flux session on host {hostname}")
+            pid = line.split()[0]
+            if check_lsb_jobid(pid, jobid):
+                return pid
+    raise ValueError(
+        f"Unable to find a flux session for LSF job {jobid} on host {hostname}"
+    )
 
 
-def lsf_get_uri(hostname, pid):
+def lsf_get_uri(hostname, jobid):
     """
     Resolve the URI of a Flux instance by running
 
-      `ssh <HOSTNAME> flux uri pid:<PID>`
+      `ssh <HOSTNAME> flux uri lsf:<JOBID>?is_compute`
 
-    and converting the result from bytecode to a string.
     """
     ssh_path = os.getenv("FLUX_SSH", "ssh")
     sp = subprocess.run(
-        [ssh_path, hostname, "flux", "uri", "--remote", f"pid:{pid}"],
+        [ssh_path, hostname, "flux", "uri", f"lsf:{jobid}?is_compute"],
         stdout=subprocess.PIPE,
-        stderr=subprocess.PIPE,
         check=True,
     )
     return sp.stdout.decode("utf-8")
@@ -92,9 +100,17 @@ class URIResolver(URIResolverPlugin):
     def resolve(self, uri):
         jobid = str(PurePath(uri.path).parts[0])
 
+        #  If 'is_compute' option is present, then assume we can find
+        #  a flux-broker associated with the LSF job locally.
+        if "is_compute" in uri.query_dict:
+            pid = lsf_job_pid(jobid)
+            uri = FluxURIResolver().resolve(f"pid:{pid}")
+            return uri.remote
+
+        #  Otherwise, determine hostname of the job and reinvoke this
+        #  resolver plugin on the remote host
         try:
-            hostname = lsf_find_compute_node(jobid)
-            return lsf_get_uri(hostname, lsf_job_pid(jobid, hostname))
+            return lsf_get_uri(lsf_find_compute_node(jobid), jobid)
         except OSError as exc:
             raise ValueError(
                 f"LSF Job {jobid} doesn't seem to have a FLUX_URI"
```

wihobbs:
@grondo Thanks for all the help, that looks good to me. I just have one question -- do you think all the stderr checks should be removed from subprocess calls, or just the one you removed above? I removed them all in the most recent commit.

Pushed this to get it off LC onto my laptop, will try the docker containers soon to see if this resolves the issue.

jameshcorbett:
Oh yeah that's my fault. I had told @wihobbs to capture stdout/stderr but add `check=True` because I thought (based on what I read in the docs) it would throw an exception with the captured stderr if the process has a nonzero exit code, but actually the docs say that the exception has the captured stderr as an _attribute_ but it isn't given in the message.

wihobbs:
@jameshcorbett, just FYI I actually removed the check from the `csm_allocation_query` (but left it in both other places) call so that a more accurate error could be raised -- the call fails when a compute node has not been allocated resources but this isn't entirely clear just from that command's error message.

grondo:
Capturing stderr was a good thought, I didn't even realize it would be a problem until I was trying it out to see why the resolver was failing under a Flux job. Of course, if there is a more Pythonic way to make sure the error printed by a subprocess is returned to the user, I'm all for it.

wihobbs:
@grondo I implemented the changes you suggested, (which I think are important, and tested on Lassen) but the Sharness tests still don't pass when evaluating `$result = $FLUX_URI`. I believe this is due to the hierarchy of instances present in the Flux session. By default, the resolver resolves the URI of the topmost Flux instance, and in a Docker container I just tested:
```
FLUX_URI = local:///tmp/flux-7kprBH/local-0
hostname=lassen9
cmd="flux uri --remote lsf:12345?is_compute"
  PID TTY          TIME CMD
    1 pts/0    00:00:00 bash
 6292 pts/0    00:00:00   flux-broker-0
 6879 pts/0    00:00:00     bash
 6884 pts/0    00:00:00       flux-broker-0
 7471 pts/0    00:00:00         bash
47120 pts/0    00:00:00           flux-start
47207 pts/0    00:00:00             flux-broker-0
48022 pts/0    00:00:00               sh
51344 pts/0    00:00:00                 ps
49633 pts/0    00:00:00               flux-shell
49661 pts/0    00:00:00                 flux-broker-0
50243 pts/0    00:00:00                   flux-script-255
50362 pts/0    00:00:00                     flux-job
50273 pts/0    00:00:00                   flux-shell
50301 pts/0    00:00:00                     flux-broker-0
51181 pts/0    00:00:00                       flux-job
51214 pts/0    00:00:00                       python3
51247 pts/0    00:00:00                       flux-shell
51274 pts/0    00:00:00                         sleep
50992 pts/0    00:00:00               python3
47208 pts/0    00:00:00             flux-broker-1
51068 pts/0    00:00:00               flux-shell
51110 pts/0    00:00:00                 sleep
34023 pts/0    00:00:00   flux-broker-0
34799 pts/0    00:00:00     rc3
34842 pts/0    00:00:00       flux-module
result = local:///tmp/flux-zNJCu9/local-0
ok 24 - flux-uri mock testing of lsf resolver works

expecting success: 
	flux job cancelall -f &&
	flux queue drain

```

I think what's happening here is that $FLUX_URI is the instance the tests are run in, while $result is the parent all the way at the top. So there are many layers in between, but even two would probably fool the resolver in this test, which is built on the premise that any user will only be looking for the topmost Flux instance on a node. 

Could we remove this evaluation and just `test_expected_success` for the return code of the `flux uri lsf:12345` call?

grondo:
I verified that, at least locally, the problem with the current PR is that the fake `LSB_JOBID` is not set in the test Flux instance:
```diff
diff --git a/t/t2802-uri-cmd.t b/t/t2802-uri-cmd.t
index 5af6d7660..4f57460e9 100755
--- a/t/t2802-uri-cmd.t
+++ b/t/t2802-uri-cmd.t
@@ -4,6 +4,10 @@ test_description='Test flux uri command'
 
 . $(dirname $0)/sharness.sh
 
+#  Export a fake LSB_JOBID for the test_under_flux instance
+#  to allow the LSF resolver mock tests to "find" this instance:
+export LSB_JOBID=12345
+
 test_under_flux 2
 
 testssh="${SHARNESS_TEST_SRCDIR}/scripts/tssh"
```

wihobbs:
Ha. yep, I noticed that too -- put the `LSB_JOBID` in the wrong line. I tested this with bionic and el7, hopefully it works now.

wihobbs:
The check 
```
             t2802-uri-cmd.t:  PASS: N=25  PASS=25  FAIL=0 SKIP=0 XPASS=0 XFAIL=0
```
passes in el8 but the tests hang and take longer than an hour.

grondo:
It was actually the `el8 - system,coverage` build that timed out. I have noticed that today on my builds as well. It just seems that the coverage builds are taking a lot longer today. I did recently update the `el8` base docker image, so that could have something to do with it, though since both coverage builds are affected in my experience, it could be something else. Coverage does slow everything down since it has to write the gcno/gcda files while running, so maybe we've found a new race or things are just slow.

When this happens, you can tell Github CI to just re-run the failed build and it might complete on a second try.

src/bindings/python/flux/uri/resolvers/lsf.py
```diff
@@ -0,0 +1,109 @@
...
+"""
+
+import getpass
+import os
+import subprocess
+from pathlib import PurePath
+
+import yaml
+
+import flux.hostlist as hostlist
```
jameshcorbett, line 26: Just FYI you could also have done `from flux import hostlist`.

grondo (approved):
This LGTM! Nice work. Go ahead and set the merge-when-passing flag when you are ready!

jameshcorbett (approved):
LGTM! Thanks @wihobbs !

wihobbs:
Just gave this a final spin on lassen and we should be good. Thanks so much for all the help @jameshcorbett and @grondo!

wihobbs:
Just wanted to add -- I learned a *lot* from all the comments and changes on this. So thanks for the great feedback, too.

Issue #4368, opened 2022-06-14
jobtap: replace `job.new` with `job.create`, `job.destroy` and add valid job callback


grondo:
As noted in #4366 [this comment]((https://github.com/flux-framework/flux-core/pull/4366#issuecomment-1155229093)), we should replace `job,new` with a `job.create` and `job.destroy` pair. This will give jobtap plugins a place to get notified of all jobs before validation and subscribe to job events, set job defaults, and create job state for jobs even when they might be marked as invalid by another jobtap plugin.

We should also make the common case of a plugin only wanting to be notified of _validated_ jobs easy, so I'd suggest a callback for jobs that have been validated (`job.validated`?) so that a plugin could subscribe to `job.validated` to setup state and `job.destroy` to remove any of said state. Both `job.create` and `job.validated` would need to be called on plugin and job manager reload.

This would supersede the need for #3812, so I will close that issue. As @garlick pointed out, it is bad practice to set up any state in `job.validate` callback since a plugin later in the stack could reject the job, resulting in state for a rejected job lingering in the plugin.

This approach would make implementing a plugin which validates _and_ creates state for jobs much cleaner to implement because the job is _only_ validated in the `job.validate` callback, and state creation is isolated to `job.validated`. It is also much preferable to _not_ call `job.validate` on reload since those jobs have already been validated.

BTW, if we also clear jobtap plugin aux data at `job.destroy` then we've also got a solution for #4296.

grondo:
No I can't think of a distinction between either `job.state.depend` or `job.event.validate` except that the naming would stay consistent even if a new state appeared between the `validate` event and the `DEPEND` state, and plugins could subscribe to this callback without needed to set up event subscription in `job.create`. Additionally, this callback would always be called for reloaded jobs.

This gives plugin authors less chances to get it wrong and reduces duplicated code between plugins. (If that makes sense)

PR #4386, opened 2022-07-01
allow configurable defaults for jobspec system attributes

garlick:
I wanted to get this posted for early feedback.  It's a new builtin jobtap plugin called `jobspec-default` that parses the policy table described in RFC 33 and can set configured defaults for jobs where they are missing.

For example one could configure
```toml
queue = "batch"
duration = 3600
```
and if either of those are missing from a job, the defaults are set by posting a `jobspec-update` event.

(Note that a duration of 0 in the submitted jobspec is considered the same as "missing", a quirk of RFC 14).

The `jobspec-update` event is posted from a `job.create` jobtap callback as described in #4368.   The event must be posted early so that `job.validate` can access it, for example to check the correct queue limits if no queue was specified by the user.  If necessary the code to add `job.create` and `job.destroy` can be split off to a separate PR.

This is based on top of PR #4351.

### DIFF
--- a/src/modules/job-manager/jobtap.c
+++ b/src/modules/job-manager/jobtap.c
@@ -42,6 +42,7 @@
 #define FLUX_JOBTAP_PRIORITY_UNAVAIL INT64_C(-2)
 
 extern int priority_default_plugin_init (flux_plugin_t *p);
+extern int jobspec_default_plugin_init (flux_plugin_t *p);
 extern int after_plugin_init (flux_plugin_t *p);
 extern int begin_time_plugin_init (flux_plugin_t *p);
 extern int validate_duration_plugin_init (flux_plugin_t *p);
@@ -53,6 +54,7 @@ struct jobtap_builtin {
 
 static struct jobtap_builtin jobtap_builtins [] = {
     { ".priority-default", priority_default_plugin_init },
+    { ".jobspec-default", jobspec_default_plugin_init },
     { ".dependency-after", after_plugin_init },
     { ".begin-time", &begin_time_plugin_init },
     { ".validate-duration", &validate_duration_plugin_init },
--- a/src/modules/job-manager/plugins/jobspec-default.c
+++ b/src/modules/job-manager/plugins/jobspec-default.c
@@ -1,0 +1,410 @@
+/************************************************************\
+ * Copyright 2022 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* jobspec-default.c - supply defaults for missing jobspec system attrs
+ *
+ * First set missing keys using values from [policy.jobspec.defaults],
+ * then override with values from [queues.<name>.policy.defaults].
+ * Post a jobspec-update event to apply the changes.
+ *
+ * See also:
+ *  RFC 33/Flux Job Queues
+ *  RFC 21/Job States and Events
+ */
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+#include <jansson.h>
+#include <flux/core.h>
+#include <flux/jobtap.h>
+
+#include "src/common/libutil/fsd.h"
+#include "src/common/libutil/errprintf.h"
+#include "src/common/libutil/errno_safe.h"
+#include "ccan/str/str.h"
+
+/* Fetch general defaults and assign to 'defaults' (caller must not free).
+ * If no defaults are configured, set defaults to NULL and return success.
+ * Return 0 on success.  On parse failure, set 'error' and return -1.
+ */
+static int get_general_defaults (json_t *conf,
+                                 json_t **defaults,
+                                 flux_error_t *error)
+{
+    json_t *o = NULL;
+    json_error_t e;
+
+    if (conf) {
+        if (json_unpack_ex (conf,
+                            &e,
+                            0,
+                            "{s?{s?{s?{s?o}}}}",
+                            "policy",
+                              "jobspec",
+                                "defaults",
+                                  "system", &o) < 0) {
+            errprintf (error, "policy.jobspec.defaults.system: %s", e.text);
+            return -1;
+        }
+    }
+    *defaults = o;
+    return 0;
+}
+
+/* Fetch defaults for queue 'name' and assign to 'defaults' (caller must NOT
+ * free).  If no defaults are configured, set defaults to NULL and return
+ * success.  Return 0 on success.  On parse failure, set 'error' and return -1.
+ */
+static int get_queue_defaults (json_t *conf,
+                               const char *name,
+                               json_t **defaults,
+                               flux_error_t *error)
+{
+    json_t *o = NULL;
+    json_error_t e;
+
+    if (conf && name) {
+        /* N.B. If a queue was named, it is an error if 'name' is missing
+         * from [queues], or if [queues] itself is missing;  however, it is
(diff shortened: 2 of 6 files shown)


garlick:
Re-pushed, fixing a broken && chain in the test, and allowing duration to be configured as an FSD per RFC 33.

garlick:
Rebased on current master.

garlick:
This is now based on #4392   (original work here is the last 3 commits)

This `jobspec-default` plugin now parses both the general defaults and the queue specific defaults described in [RFC 33](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_33.html), such that if a queue is selected either by the user or by default, the queue specific defaults are used in place of the general ones.

This is just to fill in missing bits from the user's jobspec.  I was considering limit enforcement and full queue support to be out of scope for this PR.

garlick:
Oh dear, I have broken fluxion.  Some of its tests set the `queue` system attribute in job requests, and this plugin insists that requested queues have entries in the [queues] table defined by RFC 33.
```
2022-07-06T20:07:40.279496Z job-manager.err[0]: jobspec-default: Error parsing queue1 queue policy: Object item not found: queues
[Errno 22] Error parsing queue1 queue policy: Object item not found: queues
not ok 9 - qmanager: job submitted to queue=queue1 (conservative)
```

Maybe for now I'll relax that requirement and then when we do "queues for real" we can revisit.

garlick:
Rebased on current master now that #4392 was merged.

src/modules/job-manager/plugins/jobspec-default.c
```diff
@@ -0,0 +1,376 @@
...
+        json_object_foreach (queues, queue, value) {
+            if (get_queue_defaults (h, queue, &queue_defaults, &e) < 0) {
+                errprintf (error,
+                           "Error parsing %s queue policy: %s",
+                           queue,
+                           e.text);
+                return -1;
+            }
+        }
+    }
```
chu11, line 180: is this check redundant?  the foreach iterates over every value of `queues.X`.  `get_queue_defaults()` makes sure that `queues.X` exists.  Or am I missing something.

src/modules/job-manager/plugins/jobspec-default.c
```diff
@@ -0,0 +1,376 @@
...
+         */
+        if (val != NULL
+            && streq (key, "duration")
+            && json_real_value (val) == 0)
+            val = NULL;
+
+        if (val == NULL) {
+            snprintf (nkey, sizeof (nkey), "attributes.system.%s", key);
+
+            /* Special: allow duration to be configured as FSD per RFC 33
```
chu11, line 254: this comment tripped me up, "allow duration" sounds like you're allowing fsd, but what you're really saying is "duration may be configured as FSD", so we have to handle as a special case.

t/t2219-job-manager-jobspec-default.t
```diff
@@ -0,0 +1,175 @@
...
+	jm_getattr $(cat id9) jobspec >getattr9.json &&
+	jq -e ".jobspec.attributes.system.duration == 111" <getattr9.json
+'
+test_expect_failure 'submit a job to unknown queue fails' '
+	test_must_fail flux mini submit \
+		--setattr=system.queue=pbatch /bin/true 2>submit.err &&
+	grep "Error parsing pbatch queue policy" submit.err
+'
+
+test_done
```
chu11, line 175: i think one test variant we're missing is a default queue has been specified.  So I think ...

```
queue = pbatch
[queues.pbatch.policy.jobspec.defaults.system]
duration = 123
[queues.pdebug.policy.jobspec.defaults.system]
duration = 456
```

(hopefully i did that right)

src/modules/job-manager/plugins/jobspec-default.c
garlick: `[queues]` is not required so that's why the foreach is in a block that tests for it being non-NULL.

Then each entry can optionally have `policy.jobspec.default.system` defined as a table, so we get that if available, and we throw an error if it cannot be parsed (e.g. what if `queue.pdebug.policy` had a value of 42 instead of a table?)

src/modules/job-manager/plugins/jobspec-default.c
garlick: Right, I should improve that comment.  It may be _configured_ as an FSD but it has to be a number in jobspec, so we can't just copy it over there like other attributes.

t/t2219-job-manager-jobspec-default.t
garlick: Ah good catch.

garlick:
Pushed a couple of fixups to address those comments.  Thanks @chu11!

garlick:
Actually, please hold off on any review.  I think it would be a good idea to tackle #4393 first since that will affect the way this plugin processes its config.

garlick:
I repushed with the addition of a `conf.update` jobtap callback that gets called once on plugin load and every time the config is updated with `flux config reload`, then converted the `jobspec-default` to validate its config when it changes.

src/modules/job-manager/jobtap.c
```diff
@@ -505,6 +500,11 @@ struct jobtap *jobtap_create (struct job_manager *ctx)
         goto error;
     }
 
+    if (load_builtins (jobtap) < 0) {
```
grondo, line 571: It would be nice to have a commit message for 706db0b507d1672724a018c14f39bdb1307124a7 for completeness.

grondo (approved):
This LGTM. I had a couple other minor comments but dropped them because I decided they were more questions of style.

My only other comment is that perhaps some of the missing coverage in handling some jobspec attributes could be covered if the validator was removed during the testing to allow bad values to get through to the job manager? Not very important IMO though.
This is a great improvement and well done!

src/modules/job-manager/jobtap.c
garlick: Style consistency comments are welcome BTW.

grondo: A separate PR might be nice to get an extra entry in the release notes. However, I'd say it is up to you, if it causes any issues splitting this work up, then I'd say just push the doc and test updates to this PR.

src/modules/job-manager/plugins/jobspec-default.c
```diff
@@ -0,0 +1,414 @@
...
+        }
+    }
+done:
+    json_decref (defaults);
+    json_decref (update);
+    return 0;
+error:
+    json_decref (defaults);
+    json_decref (update);
+    return -1;
```
grondo, line 373: Here was my one style comment: This code duplication stuck out to me. Would it be cleaner to initialize an `rc` variable at the top and then remove the duplication via something like

```c
int rc = -1;
```

```c
    rc = 0;
error:    
    json_decref (defaults);
    json_decref (update);
    return rc;
```

Or do people not like seeing a normal return after an `error` label? (In that case maybe rename the label `out`.)

garlick: No, that's a fair point. To be honest I've been avoiding the "rc" idiom in all but the simplest cases after I spent some time unwinding the error handling in some flux-sched functions a long time ago that "re-used" rc to hold internal function call results and fell through to the end without gotos.  When the control flow is clear, it is not a problem so I probably overreacted.

grondo: That is a good point, and I've probably been guilty of that `rc` misuse in trying to avoid the duplicate `free` blocks (hopefully not too often though!). We just need a version of C with the [`defer`](https://open-std.org/jtc1/sc22/wg14/www/docs/n2895.htm) keyword.

garlick:
Rebased on master and reworked `create_cb()` exit path to avoid code duplication.

grondo (approved):
Still LGTM!

PR #4388, opened 2022-07-02
job-manager: allow dependencies on inactive jobs

grondo:
This PR allows jobtap plugins to lookup inactive jobs (where that makes sense). This in turn allows the dependency-after plugin to utilize jobs that are inactive as the antecedent for the various `after*` dependency schemes.


garlick (approved):
LGTM!

grondo:
Thanks! I've set MWP.

PR #4389, opened 2022-07-05
kvs: misc cleanups

chu11:
Peeling off from #4383, that PR is going to pivot in a new direction and these cleanups wouldn't necessarily be relevant to the shift.  But they are still decent cleanups.


garlick (approved):
LGTM!

Issue #4296, opened 2022-04-20
job-manager: aux items for jobs are now kept forever or until job purge


grondo:
This occurred to me while assisting on flux-framework/flux-accounting#231.

The jobtap interface allows plugins to stash arbitrary data along with jobs via `flux_jobtap_job_aux_set(3)`. When the job-manager destroyed all jobs that became inactive, this data was destroyed (if the plugin registered a destructor) as the jobs transition to inactive since the job structures were immediately destroyed after the clean event.

Now the job manager keeps jobs in an inactive has until a purge is done, which means this plugin data can stay around for the entire lifetime of the job.

This isn't a problem per se, but it does seem slightly wasteful since there are no further jobtap plugin callbacks after the job is inactive. We should decide if this is a problem, and then potentially we could special case jobtap `aux_item` entries and delete them before jobs are placed on the inactive hash.

garlick:
As @grondo noted in #4392, jobtap plugins are the only users of `job->aux`, so this issue might be resolved by simply clearing that out  `job.cleanup`  returns.

garlick:
Just realized there is a gotcha when aux item destructors are located within a plugin.  In that case we have to be sure they are called when the plugin is unloaded or we may try to execute code that isn't there anymore.

PR #4392, opened 2022-07-06
jobtap: add job.create, job.destroy callbacks

garlick:
This was split off from #4386.

It just adds the job.create, job.destroy jobtap callbacks and related stuff as discussed in #4368.

### DIFF
--- a/doc/man7/flux-jobtap-plugins.rst
+++ b/doc/man7/flux-jobtap-plugins.rst
@@ -115,16 +115,40 @@ CALLBACK TOPICS
 The following callback "topic strings" are currently provided by the
 *jobtap* interface:
 
+job.create
+  The ``job.create`` topic notifies a jobtap plugin about a newly introduced
+  job. This call may be made in three different situations:
+
+    1. on job submission
+    2. when the job manager is restarted and has reloaded a job from the KVS
+    3. when a new jobtap plugin is loaded
+
+  In case 1 above, the job state will always be ``FLUX_JOB_STATE_NEW``, while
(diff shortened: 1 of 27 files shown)


grondo:
I didn't fully review yet, but I did have one thought:

> Also fixes #4296 since job.destroy is a hook for plugins to destroy their job-specific state.

I was actually thinking that the job manager could destroy per-job jobtap state after `job.destroy` on behalf of the plugins. I _think_ jobtap plugins are the only users of the `job->aux` container, so perhaps we could rename that `plugin_aux` and destroy it after `job.destroy` is called?

Otherwise, we might want to rework the aux interface for plugins, e.g. remove the last parameter of `flux_jobtap_job_aux_set(3)` since it wouldn't be useful to plugins:

```c
int flux_jobtap_job_aux_set (flux_plugin_t *p,
                             flux_jobid_t id,
                             const char *name,
                             void *val,
                             flux_free_f free_fn);
```

and add a `flux_jobtap_job_aux_free(3)` call.

garlick:
Good idea - that would reduce boiler plate needed in plugins.  I had thought of doing something similar but didn't realize that jobtap was the only user of job->aux.  In any case, adding a second aux to the struct costs very little if the need arises in the future.

Would that be OK to do as a separate PR?    This PR could retain the destroy callback, but drop the changes I made to free state in a couple plugins and also drop the reference to fixing #4296?

grondo:
Sure, a separate PR sounds appropriate to me.

garlick:
OK, the stuff related to #4296 has been removed here.

grondo (approved):
LGTM, nice improvement IMO.

Issue #4401, opened 2022-07-11
process historical kvs dumps in CI


garlick:
> here's a script we could add to CI which can generate a dumpfile for any flux-core version, by default the most recent tag. It includes a very basic workload for now, but this could be extended over time.

> The idea would be to run this in a CI step before the build and save the dumpfile in the `valid` dumpfiles directory automatically. This way we wouldn't have to remember to generate new dumpfiles for each tag. Of course the drawback is the tests would only run in CI...

```sh
#!/bin/bash
#
#  Run a workload in the docker container from a given flux-core tag
#   (previous tag by default), and preserve a dumpfile.
#
OUTPUTDIR="$(pwd)/output"
TAG=$(git describe --abbrev=0)
WORKLOAD="\
#!/bin/sh
#
#  Basic workload used to create a set of varied eventlogs in KVS
#
flux mini bulksubmit --quiet --cc=1-4 {} ::: true false nocommand &&
id=\$(flux mini submit --urgency=hold hostname) &&
flux job cancel \$id &&
id=\$(flux mini submit --wait-event=start sleep 30) &&
flux mini submit --quiet --dependency=afternotok:\$id true &&
flux job cancel \$id &&
flux queue drain &&
flux jobs -a
"
declare -r prog=${0##*/}
die() { echo -e "$prog: $@"; exit 1; }

declare -r long_opts="help,tag:,workload:,statedir:"
declare -r short_opts="ht:w:d:"
declare -r usage="
Usage: $prog [OPTIONS]\n\
Run a workload under the latest flux-core tag docker image and save\n\
the content file to a state directory.\n\
\n\
Options:\n\
 -h, --help             Display this message\n\
 -t, --tag=NAME         Run against flux-core tag NAME instead of latest tag\n\
 -w, --workload=SCRIPT  Run workload SCRIPT instead of a default\n\
 -d, --output=DIR       Save dumpfile to DIR (default=$OUTPUTDIR)
"

# check if running in OSX
if [[ "$(uname)" == "Darwin" ]]; then
    # BSD getopt
    GETOPTS=`/usr/bin/getopt $short_opts -- $*`
else
    # GNU getopt
    GETOPTS=`/usr/bin/getopt -u -o $short_opts -l $long_opts -n $prog -- $@`
    if [[ $? != 0 ]]; then
        die "$usage"
    fi
    eval set -- "$GETOPTS"
fi

while true; do
    case "$1" in
[...]

PR #4402, opened 2022-07-12
ci: create kvs dumpfile from previous tag for testing

grondo:
This PR adds a script like the one proposed in #4401 to the GitHub workflow. This runs a small workload in the most recent tag and saves the dumpfile to `t/job-manager/dumps/valid` which will cause the dumpfile to get testing during `make check`.

This should, in theory, automatically catch PRs proposed that can't restart the previous tagged version KVS.

The workload here is very minimal, but could be easily extended if needed.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     if: github.event_name == 'pull_request'
     steps:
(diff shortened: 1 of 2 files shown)


grondo:
Oh, this is based on #4399 since that is where the dumpfile tests were added by @garlick.

garlick (approved):
This is a really excellent test to have going.  Thanks!

Issue #4405, opened 2022-07-12
job-manager: dependency does not get added until after job exits job.state.depend


cmoussa1:
After updating my flux-core Docker images on my local machine to the latest versions of flux-core, I noticed some sharness tests failing with the priority plugin, specifically the tests that deal with `max_running_jobs` limits. Currently, this limit is enforced by adding a job dependency to the job in `job.state.depend`. After updating flux-core, however, it seems the dependency does not get added until after the job exits `job.state.depend` and enters SCHED state. Here's a specific case I ran into:

For example, let’s say I have a `max_running_jobs` limit of 1 and I submit 2 jobs. The first job should run fine, and the second job should have a dependency added to it (and thus held in DEPEND state) because the user has hit their `max_running_jobs` limit. But because the dependency does not get added by the time the job exits DEPEND state, the job gets scheduled and the dependency is not enforced and the job runs. 

Because the job that was supposed to have a `max_running_jobs` dependency on it is now running, when it transitions to `job.state.inactive` and it looks to see if there are any held jobs, I _think_ it sees its own job ID under the plugin's internal list of held jobs and then tries to remove the dependency from itself, which maybe causes some unexpected behavior. 

As an aside, I can (and should) prevent some of this unexpected behavior in the plugin by introducing some additional checks before previously-held jobs are removed from the plugin's internal list. But from the perspective of adding dependencies, I wonder if this might be a flux-core bug.

grondo:
Thanks for opening this issue @cmoussa1! @garlick and I ran down the problem to a difference introduced by the new event queueing mechanism.  If a job enters DEPEND state without any dependencies, `event_job_action()` immediately issues the `depend` event, transitioning the job out of DEPEND state, event if `dependency-add` events are pending in `job->event_queue`. 

Though a strong argument can be made that this is the proper behavior, unfortunately the use of `job.state.depend` to add dependencies is documented in [flux-jobtap-plugins(7)](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man7/flux-jobtap-plugins.html), and is not only used by flux-accounting and the `mf_priority.so` plugin, but also by the DWS jobtap plugin in flux-coral2. Therefore, I think our best course of action is to restore the previous behavior.

I'll put up a PR momentarily.

PR #4406, opened 2022-07-12
job-manager: fix dependency-add from job.state.depend callback

grondo:
As noted in #4405, recent changes to how posted events are handled by the job manager broke the addition of dependencies in the `job.state.depend` callback. This PR fixes that issue by avoiding post of the `depend` event if there is a `dependency-add` event queued.

A test is also added, since that was clearly missing coverage. (All the tests apparently used `job.dependency.*` callbacks instead)

Finally, to avoid merging PRs which break flux-accounting, a CI check similar to the existing flux-sched check is added for flux-accounting.

Since this modifies mergify config it will likely need a manual merge.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -80,6 +80,24 @@ jobs:
         cd flux-sched &&
         src/test/docker/docker-run-checks.sh -j 4 -i bionic
 
+  check-accounting:
+    needs: [ python-format, python-lint, mypy ]
+    name: flux-accounting check
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+        fetch-depth: 0
+    - run: >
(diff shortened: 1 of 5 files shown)


cmoussa1 (approved):
This LGTM! Thank you for getting this change up so quickly, and thanks for adding the flux-accounting check to CI too. I need to be better on my end about keeping my flux-core Docker images up to date, so I'm sorry that I'm a little late on bringing this up.

Since you added the check to `.mergify.yml`, I think this will need to be merged manually once all the checks pass, correct?

grondo:
> Since you added the check to .mergify.yml, I think this will need to be merged manually once all the checks pass, correct?

Yes, I'd also like to get @garlick's approval.

garlick (approved):
LGTM!  Sorry to you and @cmoussa1 for not catching this when the change was made.

LMK when you'd like the button pressed.

grondo:
Feel free to press the button when required checks all pass. Thanks!

grondo:
Thanks!

PR #4407, opened 2022-07-13
Use flux_error_t and errprintf() over char buf and snprintf()

chu11:
While working on #4383, review comments noted that I did not use `flux_error_t` and `errprintf()` for writing out error messages in the `kvs` module.  It ended up I had copied the style of usage from the `resource` module.

I noticed that style was still used in a number of places, so updated its usage in lots of places.  Hopefully I did not go too overboard and update more than I should have :-|  This was not a super deep audit.

Note that in a few locations the char buf size was larger than the 160 bytes of `flux_error_t` (like 256).  We could make `flux_error_t` larger?  But I suspect the 256 bytes was arbitrarily chosen and unlikely to matter.

### DIFF
--- a/src/broker/brokercfg.c
+++ b/src/broker/brokercfg.c
@@ -19,6 +19,7 @@
 #include <assert.h>
 
 #include "src/common/libutil/log.h"
+#include "src/common/libutil/errprintf.h"
 
 #include "attr.h"
 #include "module.h"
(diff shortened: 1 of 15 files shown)


garlick:
`errprintf()` needn't be cast to (void).  It always returns -1 to allow the idiom `return errprintf ...` but it doesn't have a meaningful return code that you need to tell people you're ignoring.  Pretty sure compilers don't care.

chu11:
> errprintf() needn't be cast to (void). It always returns -1 to allow the idiom return errprintf ... but it doesn't have a meaningful return code that you need to tell people you're ignoring. Pretty sure compilers don't care.

Yeah, I left them there just b/c there was `(void) snprintf( ...)` before.  Not sure why those void casts were there.  I'll go ahead and just remove them then.

garlick:
I probably put them there to telegraph that yes I know I'm ignoring this return code that you're typically supposed to check.

chu11:
re-pushed, removing those excess void casts.  went ahead and squashed

chu11:
re-pushed, fixing a `content-s3` build issue.  Also noticed a logic error in the content-s3 code, so tacked on a minor fix to the top of the commits.

garlick (approved):
Well that was a bit of work  - and a nice result!

PR #4410, opened 2022-07-15
kvs: rename kvs.sync target to kvs.wait-version

chu11:
Problem: The kvs.sync target has been used with the
flux_kvs_wait_version() function (and earlier incarnations) since
commit:

https://github.com/flux-framework/flux-core/commit/20555806079d8c5922547cc1f7229d39aab59b7c kvs: replace 'version' key with kvs.sync,getroot

The decision to use the name "kvs.sync" vs. "kvs.wait-version"
appears somewhat arbitrary.  A likely decision based on the then
alpha development flux-core was in.  A comment even exists to inform
readers the sync_request_cb() function is related to "wait_version()".

The use of the term "sync" can be confusing now given the FLUX_KVS_SYNC
flag and heartbeat syncs.

Solution: Rename the kvs.sync target to kvs.wait-version.  Adjust all
other related code accordingly (variables, function names, helper code,
filenames, etc.).

---

Additional comments:

I admittedly wasn't 100% sure if this was necessary, perhaps still aren't 100% on it.  But both myself and @garlick forgot about the fact `kvs.sync` existed. It also lead to @garlick and I debating name changes for the "sync" config in #4383.

So I decided to look up the history, like why is this called `kvs.sync`.  It appears to be a somewhat arbitrary choice in January 2014 (predates v0.1.0 by two years), so during kvs super experimental time period.   I found it the funniest that a comment reminds the reader that `kvs.sync` is really for `kvs_wait_version()`.

So I decided, maybe its for the best.

As an aside, the best name I could come up with to replace the stat `#syncers` was `#versionwaiters`, welcome to alternate name suggestions.


t/t1000-kvs.t
```diff
@@ -1243,8 +1243,8 @@ test_expect_success 'kvs: get --label works' '
 test_expect_success 'getroot request with empty payload fails with EPROTO(71)' '
 	${RPC} kvs.getroot 71 </dev/null
 '
-test_expect_success 'sync request with empty payload fails with EPROTO(71)' '
-	${RPC} kvs.sync 71 </dev/null
+test_expect_success 'wait-vesion request with empty payload fails with EPROTO(71)' '
```
garlick, line 1246: typo: "wait-vesion"

garlick (approved):
LGTM, just one typo.

PR #4411, opened 2022-07-15
jobtap: add conf.update callback

garlick:
Problem: jobtap plugins have no simple way to be notified when TOML configuration changes. 

This adds a `conf.update` callback topic.  It is called 
- when the plugin is initially loaded
- each time `flux config reload` is run

The "proposed" config object is found in the callback input arguments under `conf`.  The plugin may reject the configuration by returning  `flux_jobtap_error()` with a human readable error (preferably calling out the offending TOML section).  If the config is rejected at plugin load time, the plugin load fails, e.g.
```
$ flux jobtap load config.so
flux-jobtap: ERROR: [Errno 22] Error parsing [testconfig]: Object item not found: testconfig
```
or if rejected at config reload time, `flux config reload` fails, e.g.
```
$ flux config reload
flux-config: reload: job-manager: Error parsing [testconfig]: Expected string, got integer
```

Note that this callback is not mandatory for access to configuration, since plugins can always fetch the current configuration object by calling
```c
flux_get_conf (flux_jobtap_get_flux (plugin))
```

Split from #4386

### DIFF
--- a/src/modules/job-manager/jobtap.c
+++ b/src/modules/job-manager/jobtap.c
@@ -66,6 +66,7 @@ struct jobtap {
     zhashx_t *plugins_byuuid;
     zlistx_t *jobstack;
     char last_error [128];
+    bool configured;
 };
 
 struct dependency {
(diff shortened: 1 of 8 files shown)


src/modules/job-manager/jobtap.c
```diff
@@ -66,6 +66,7 @@ struct jobtap {
     zhashx_t *plugins_byuuid;
     zlistx_t *jobstack;
     char last_error [128];
+    bool configured;
```
grondo, line 69: Very minor typo in commit message: `A conf.update jobtap callbacks is added` -> ` conf.update jobtap callback is added`

src/modules/job-manager/jobtap.c
```diff
@@ -398,6 +398,58 @@ static int jobtap_conf_entry (struct jobtap *jobtap,
     return 0;
 }
 
+static int jobtap_call_conf_update (flux_plugin_t *p,
```
grondo, line 401: Very minor commit message typo: configuraiton

t/Makefile.am
```diff
@@ -429,6 +429,7 @@ check_LTLIBRARIES = \
 	job-manager/plugins/create-event.la \
 	job-manager/plugins/create-reject.la \
 	job-manager/plugins/perilog-test.la \
+	job-manager/plugins/config.la \
```
grondo, line 432: Minor: commit body trails off: `Add test coverage to`

grondo (approved):
Nice work! I appreciate the improvements to the error logging from jobtap plugins, and this whole PR is done in a very thoughtful manner. Thanks!

grondo:
flux-sched check failed with known failure:
```
rv1-bootstrap: resource idempotency preserved
```

I'll restart it when all other tests complete.

garlick:
Actually - just spotted a problem.  I'll remove MWP for now.  Thanks though.

garlick:
I inserted a new commit:  a50a1c0584ddbf35360137df1cc55fd7aaad7598.

While rebasing #4386 on this PR and adding tests, I noticed that failure of a builtin plugin isn't detected, so the other patch here that adds logging for those failures is effectively dead code.

There's no test in this pr for failing a builtin plugin (could see how to do that easily) but there will be test coverage for this in #4386.  If that's not acceptable  I could move both commits to the other PR.

grondo:
The current state of the PR seems acceptable to me

PR #4412, opened 2022-07-17
job-manager: fix race in job eventlog commit and job shell start

grondo:
As described in #4409, this PR fixes the race between eventlog creation in the job manager and job startup by adding a flag which queues events until the initial submit event in the eventlog is committed to the KVS.

Additionally, the event batch timeout is made tunable at runtime via a new `job-manager.set-event-timeout` RPC (much easier to use than a module option), and that is used in a reproducer in the tesuite using @chu11's trick of extending the timeout to 1s.

### DIFF
--- a/src/modules/job-manager/event.c
+++ b/src/modules/job-manager/event.c
@@ -59,10 +59,10 @@
 
 #include "event.h"
 
-const double batch_timeout = 0.01;
-
 struct event {
     struct job_manager *ctx;
+    flux_msg_handler_t **handlers;
+    double batch_timeout;
     struct event_batch *batch;
     flux_watcher_t *timer;
     zlist_t *pending;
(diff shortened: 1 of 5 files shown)


garlick (approved):
Nicely done!  Really impressed with @chu11 and @grondo ran this down and proposed a solution so quickly!

grondo:
> Nicely done! Really impressed with @chu11 and @grondo ran this down and proposed a solution so quickly!

Thanks! This was mostly due to @chu11's intuition and reproducer, I was down in the deep weeds.

I've pushed a small modification to the tests to get slightly better coverage of the new RPC. If the tests are passing I'll set MWP.

grondo:
I'll wait to set MWP until we've decided this is right fix (discussion in #4409)

grondo:
Sounds like we decided we should merge this stopgap fix for now since the alternate solution proposed in #4409 didn't seem to be an easy fix. I'll set MWP.

Issue #3652, opened 2021-05-13
enforcing per-job limits on duration and job size


cmoussa1:
_If I need to instead open this issue in a different repository, please let me know and I'll move it._

The first sub-issue that came up as a result of breaking up [flux-sched #638](https://github.com/flux-framework/flux-sched/issues/638) is the need to enforce per-job limits both in terms of a max wall time and a max job size.

It was noted in #638 that these two limits could be handled by the job-ingest validator:

> I'm not sure if we've documented it anywhere, but we can pretty easily handle the walltime and job size limits at the job-ingest validator. Note: if we enable modifying jobspec post-submission, we will need to re-validate. If anyone disagrees, we should open a separate issue to discuss there.

I would assume that the limits would need to be defined at the queue level (e.g. I can only request up to 4 nodes and run my job for 30 minutes at a time in one queue and an unlimited amount of nodes and wall time in another queue), and we would need to store these definitions in a configuration file somewhere.

garlick:
Closed by #4415

Issue #4309, opened 2022-05-02
job manager needs new interface for job "limits"


grondo:
As we discuss user job and queue limits (related #4302 #4306), we realized that it may be impossible to handle certain use cases for limits given current job manager design, assuming limits are evaluated in the `job.validate` plugin callback (as I believe is currently implemented in the job-accounting `mf_priority.so` plugin).

To review, there are a few classes of limits which could be handled within the job manager and thus via jobtap plugins, including:

 * user or queue submission limits, e.g. a maximum number of allowed active jobs. If a user submits a job that would exceed this limit, the job should be rejected.
 * user or queue running job limits, e.g. a maximum number of running jobs. If a user submits a job that exceeds these limits, the job should be held but not rejected.
 * resource limits - the scheduler _may_ be the only entity that can enforce these limits. E.g. if there is a limit on total node count, but a job is submitted that only requests cores, the job manager cannot make the determination if this job would exceed some sort of limit. The scheduler must evaluate the job and ensure it does not exceed the limit before allocating resources.

The discussion here only relates to the first two bullets (at least for now). The mechanism for a scheduler to apply resource limits is outside of the scope of this issue.

A problem with the current approach of using existing jobtap plugin callbacks to implement these limits is that each plugin is operating in isolation, and therefore there is no way to implement a plugin that overrides either fatal or hold limits. One idea floated by @garlick is to treat limits like dependencies, with add/remove events that can add or remove specific limits. This could reuse the current DEPEND state, or perhaps we would want to add a new state specific to limits.

Instead of rejecting or holding jobs, plugins that are enforcing limits would instead _add_ a fatal or nonfatal "limit" to the job. A plugin that _overrides_ limits could remove limits that have been added up to that point in the plugin call stack (perhaps we could also allow a plugin to push an override event that clears even future limits, so that plugin order doesn't matter). After all plugin 'limit' callbacks have been called, the current state of limits is applied. If there are outstanding nonfatal limits then the job stays in the LIMIT (or DEPEND) state. The list of outstanding limits would be available via job listing utilities, similar to dependencies. One or more fatal limit events would cause the job to rejected.

This scheme allows a lot of flexibility in how limits, at least those limits which can be enforced by the job manager, are applied and removed. Mainly, it allows limits to be added and removed by separate plugins, and allows some insight into which limits may be holding up a job.

garlick:
I like the fact that with this design
[...]

PR #4415, opened 2022-07-19
job-manager: add limits plugins for duration and job size

garlick:
This adds a plugin that implements general and per-queue limits as described in RFC 33, based on just checking the jobspec against the TOML config.

As I was looking for an issue to reference when posting this, I came across #4309 and realized I dove into this today without remembering the idea of using a dependency like scheme to add/remove limits.  Which was a good idea!  Tomorrow I'll pause this and spend some considering that approach instead.

### DIFF
--- a/doc/man5/flux-config-policy.rst
+++ b/doc/man5/flux-config-policy.rst
@@ -1,0 +1,108 @@
+=====================
+flux-config-policy(5)
+=====================
+
+
+DESCRIPTION
+===========
+
+The ``policy`` table captures general site preferences for job request defaults
+and limits, as described in RFC 33.
+
+Each queue defined in the ``queues`` table described in
+:man5:`flux-config-queues` may have a ``policy`` sub-table that follows the
+same rules as the main table.  The per-queue table overrides the general table
+for jobs submitted to that queue.
+
+DEFAULTS
+========
+
+Default values for job requests may be configured in the
+``policy.jobspec.defaults.system`` table.  If a job request does not specify
+a value for a system attribute that is configured in the table, the configured
+value is substituted.  Some common examples are:
+
+policy.jobspec.defaults.system.duration (float seconds or FSD string)
+   (optional) If a job is submitted without specifying a job duration,
+   this value is substituted.
+
+policy.jobspec.defaults.system.queue (string)
+   (optional) If a job is submitted without specifying a queue name,
+   this value is substituted.
+
+LIMITS
+======
+
+Limits may be configured in the ``policy.limits`` table.  Job requests are
+rejected at submission time if they violate configured limits.
+
+.. note::
+   It is possible to set a queue-specific limit to an `unlimited` value,
+   and thereby defeat a configured general limit for jobs submitted to that
+   queue.  The actual value used to indicate `unlimited` varies by limit
+   type and is noted below.
+
+policy.limits.duration (float seconds or FSD string)
+   (optional) Maximum duration that a job may request (0 = unlimited).
+
+.. note::
+   Limit checks take place before the scheduler sees the request, so it is
+   possible to bypass a node limit by requesting only cores, or the core limit
+   by requesting only nodes (exclusively) since this part of the system does
+   not have detailed resource information.  Generally node and core limits
+   should be configured in tandem to be effective on resource sets with
+   uniform cores per node.  Flux does not yet have a solution for node/core
+   limits on heterogeneous resources.
+
+policy.limits.job-size.max.nnodes (integer)
+   (optional) Maximum number of nodes that may be requested (-1 = unlimited).
+
+policy.limits.job-size.max.ncores (integer)
+   (optional) Maximum number of cores that may be requested (-1 = unlimited).
+
+policy.limits.job-size.max.ngpus (integer)
+   (optional) Maximum number of GPUs that may be requested (-1 = unlimited).
+
+policy.limits.job-size.min.nnodes (integer)
+   (optional) Minimum number of nodes that may be requested (-1 = unlimited).
+
+policy.limits.job-size.min.ncores (integer)
+   (optional) Minimum number of cores that may be requested (-1 = unlimited).
+
+policy.limits.job-size.min.ngpus (integer)
+   (optional) Minimum number of GPUs that may be requested (-1 = unlimited).
+
+
+
+EXAMPLE
+=======
+
+::
+
+   [policy.defaults]
+   duration = "1h"
+   queue = "pbatch"
+
+   [policy.limits]
+   duration = "4h"
+   job-size.max.nnodes = 8
+   job-size.max.gpus = 4
+
+   [queues.pdebug.policy.limits]
+   duration = "30m"
+   job-size.max.gpus = -1  # unlimited
+
+
+RESOURCES
+=========
+
+Flux: http://flux-framework.org
+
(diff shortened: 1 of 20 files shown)


garlick:
I'm not making a lot of progress seeing how to build up a more comprehensive limit architecture using the ideas in #4309.  These plugins are rather simple minded and are only capable of rejecting a job submission if the request overtly asks for more than the configured maximum duration or over/under the configured maximum/minimum job size.

I guess the question is whether it makes any sense to put these in as a stop gap until we can circle back and do something smarter.  If not, I'll withdraw them and we'll need to probably have some design discussions as a next step.

grondo:
> I guess the question is whether it makes any sense to put these in as a stop gap until we can circle back and do something smarter. If not, I'll withdraw them and we'll need to probably have some design discussions as a next step.

It may make sense to use this as a stopgap.

Perhaps we should first ask what is the minimum functionality we'd need for a system with multiple users and then see if the features added address a large portion of those features. cc: @ryanday36.

ryanday36:
rejecting jobs that exceed a node or time limit is definitely useful functionality. I don't think we need to do anything more complicated than that, especially not initially.

grondo:
It should possibly be mentioned that in order to set a limit on nnodes, a corresponding limit for ncores must also be set, otherwise a user can get around the nnodes limit by requesting cores only (at least currently):

```console
$ flux mini run -N4 hostname
flux-mini: ERROR: [Errno 22] requested nnodes exceeds policy limit of 2
$ flux mini run -N4 flux getattr size
flux-mini: ERROR: [Errno 22] requested nnodes exceeds policy limit of 2
$ flux mini run -n4 -c4 flux getattr size
4
4
4
4
```

This will be tricky on heterogeneous clusters where a request for cores only could correspond to a range of nnodes depending on what the scheduler chooses.

garlick:
I expanded the man page a bit and added a note about the limitations of the node/core limits per the suggestion (thanks)!

garlick:
Testing this on my test system instance.   I set the following limits:
```
[policy.limits]
duration = "1h"
job-size.max.nnodes = 4
job-size.max.ncores = 16
```
And when I exceed them I get e.g.
```
$ flux mini run -n17 stress --cpu 1
flux-mini: ERROR: [Errno 22] requested ncores exceeds policy limit of 16
$ flux mini run -t 2h hostname
flux-mini: ERROR: [Errno 22] requested duration exceeds policy limit of 1h
```
Incidentally, although it's not something we would likely do in LC, the node/core limits are handy on my test system since my R defines some nodes that are down long term.  It prevents me from submitting a job that will not run anytime soon because it's "feasible" but too big to run on the nodes/cores I have online.

garlick:
linking #3652

src/modules/job-manager/plugins/limit-duration.c
```diff
@@ -0,0 +1,323 @@
...
+                        flux_plugin_arg_t *args,
+                        void *arg)
+{
+    struct limit_duration *ctx = flux_plugin_aux_get (p, auxkey);
+    flux_job_state_t state;
+    double duration = DURATION_UNLIMITED;
+    const char *queue = NULL;
+    flux_error_t error;
+
+    if (flux_plugin_arg_unpack (args,
```
chu11, line 238: any reason not to have the same comments / checks as in the other limits plugin?

```


    /* If no limits are configured, return immediately.  This is the common
     * case for a non-system instance and since this plugin is always loaded,
     * don't waste time.
     */
    if (!limits_isset (&ctx->general_limits)
        && zhashx_size (ctx->queues) == 0)
        return 0;
```

and

```
    /* Parse (optional) jobspec attributes.system.queue.
     * Leave queue NULL if unspecified.
     * Throw an error if it's the wrong type or related.
     */
```

doc/man5/flux-config-policy.rst
```diff
@@ -0,0 +1,108 @@
...
+========
+
+Default values for job requests may be configured in the
+``policy.jobspec.defaults.system`` table.  If a job request does not specify
+a value for a system attribute that is configured in the table, the configured
+value is substituted.  Some common examples are:
+
+policy.jobspec.defaults.system.duration (double or FSD string)
+   (optional) If a job is submitted without specifying a job duration,
+   this value is substituted.
```
chu11, line 27: should mention units?   just to be clear.

src/modules/job-manager/plugins/limit-duration.c
```diff
@@ -0,0 +1,323 @@
...
+ *  [policy.limits.duration]
+ * Queue-specific limit:
+ *  [queues.<name>.policy.limits.duration]
+ *
+ * N.B. a queue limit may override the general limit with a higher or lower
+ * limit, even "unlimited" (0).
+ *
+ * See also:
+ *  RFC 33/Flux Job Queues
+ */
```
chu11, line 23: would it be worth mentioning `policy.jobspec.defaults.system.duration`, just to say?

doc/man5/flux-config-policy.rst
```diff
@@ -0,0 +1,108 @@
...
+rejected at submission time if they violate configured limits.
+
+.. note::
+   It is possible to set a queue-specific limit to an `unlimited` value,
+   and thereby defeat a configured general limit for jobs submitted to that
+   queue.  The actual value used to indicate `unlimited` varies by limit
+   type and is noted below.
+
+policy.limits.duration (double or FSD string)
+   (optional) Maximum duration that a job may request (0 = unlimited).
```
chu11, line 46: units?

t/t2221-job-manager-limit-duration.t
```diff
@@ -0,0 +1,124 @@
...
+	EOT
+	test_must_fail flux config reload
+'
+test_expect_success 'configure an invalid queue duration limit' '
+	cat >config/policy.toml <<-EOT &&
+	[queues.pdebug.policy.limits]
+	duration = "xyz123"
+	EOT
+	test_must_fail flux config reload
+'
```
chu11, line 123: worthwhile to add test that also adds in `policy.jobspec.defaults.system.duration`?

t/t2222-job-manager-limit-job-size.t
```diff
@@ -0,0 +1,126 @@
...
+
+test_expect_success 'unload all built-in plugins' '
+        flux jobtap remove ".*"
+'
+test_expect_success 'configure an invalid job-size limit' '
+	cat >config/policy.toml <<-EOT &&
+	[policy.limits]
+	job-size.max.nnodes = -42
+	EOT
+        flux config reload
```
chu11, line 21: errant tab/spacing here?

chu11 (commented):
minor things i found on initial skim

doc/man5/flux-config-policy.rst
garlick: Oh yeah and "double" is not a toml or JSON type.  TOML uses "real".

t/t2221-job-manager-limit-duration.t
garlick: Well that uncovered a bug where I was using `json_real_value()` to parse a type = json_integer value and getting 0.0, so yes it was worthwhile!

garlick:
Thanks for the great comments @chu11.  I addressed those and force pushed after resolving the conflicts with master.

garlick:
Let's say this fixes #3652.

garlick:
I'm going to go ahead and set MWP here.

chu11 (approved):

Issue #4413, opened 2022-07-17
flux_rpc_get_unpack on empty string returns ENOMEM


garlick:
@grondo noted in #4409:

> I can't even figure out how to create an empty eventlog. Any hints? If I use flux_kvs_txn_put (txn, 0, key, "") then I get
```
$ flux mini run hostname
2022-07-17T02:25:18.141047Z job-info.err[0]: get_main_eventlog_continuation: flux_rpc_get_unpack: Cannot allocate memory
0.050s: job.exception type=exec severity=0 flux_job_event_watch_get: Cannot allocate memory
flux-job: task(s) exited with exit code 1
0.049s: flux-shell[0]: FATAL: input: flux_job_event_watch_get: Cannot allocate memory
````
> ENOMEM from flux_rpc_get_unpack(3) seems like another bug.

The `flux_kvs_txn_put ("")` _should_ create a kvs value containing an empty string (e.g. just a NULL character).

The `flux_rpc_get_unpack()` in `flux_job_event_watch_get()` attempts to unpack the format string `{s:s}` and fails appropriately, but the ENOMEM error is wrong.

**Comment by garlick** [comment]
Note, this works:
```
$ FLUX_HANDLE_TRACE=1 flux kvs put a=""
--------------------------------------
> 0.00000
> kvs.commit
> flags=topic,payload,route userid=unknown rolemask=none nodeid=any matchtag=1
> {"namespace":"primary","flags":0,"ops":[{"key":"a","flags":0,"dirent":{"ver":1,"type":"val","data":""}}]}
--------------------------------------
< 0.00042
< kvs.commit
< flags=topic,payload,route userid=5588 rolemask=owner errnum=0 matchtag=1
< {"rootref":"sha1-7880715848f056a10f0168c7ebd1febf46d6a846","rootseq":7}

$ FLUX_HANDLE_TRACE=1 flux kvs get a
--------------------------------------
> 0.00000
> kvs.lookup
> flags=topic,payload,route userid=unknown rolemask=none nodeid=any matchtag=1
> {"key":"a","namespace":"primary","flags":0}
--------------------------------------
< 0.00029
< kvs.lookup
< flags=topic,payload,route userid=5588 rolemask=owner errnum=0 matchtag=1
< {"val":{"data":"","type":"val","ver":1}}
```

**Comment by grondo** [comment]
This simpler reproducer shows that the problem might be in the job-info module:

```
[...]

PR #4416, opened 2022-07-19
job-info: handle invalid eventlog entry errors more carefully

chu11:
Per discussion in #4413, there was a corner case where ENOMEM was returned to a caller when an eventlog was empty.  The ENOMEM came from assumptions that an eventlog can never be empty (RFC18 says that an empty eventlog should be impossible).  Handle several corner cases around this fact.

I decided that since an empty eventlog should be impossible, instead of returning EINVAL, I returned EPROTO.  Right choice?

As a fallout, I updated other code to also return EPROTO if the eventlog was in error.

### DIFF
--- a/src/modules/job-info/allow.c
+++ b/src/modules/job-info/allow.c
@@ -36,10 +36,14 @@ static int eventlog_get_userid (struct info_ctx *ctx, const char *s,
 
     if (!(a = eventlog_decode (s))) {
         flux_log_error (ctx->h, "%s: eventlog_decode", __FUNCTION__);
+        /* if eventlog improperly formatted, we'll consider this a
+         * protocol error */
+        if (errno == EINVAL)
+            errno = EPROTO;
         goto error;
     }
     if (!(entry = json_array_get (a, 0))) {
-        errno = EINVAL;
+        errno = EPROTO;
         goto error;
     }
     if (eventlog_entry_parse (entry, NULL, &name, &context) < 0) {
@@ -48,7 +52,7 @@ static int eventlog_get_userid (struct info_ctx *ctx, const char *s,
     }
     if (strcmp (name, "submit") != 0 || !context) {
         flux_log (ctx->h, LOG_ERR, "%s: invalid event: %s", __FUNCTION__, name);
-        errno = EINVAL;
+        errno = EPROTO;
         goto error;
     }
     if (json_unpack (context, "{ s:i }", "userid", &userid) < 0) {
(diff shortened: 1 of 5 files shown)


t/issues/t4413-empty-eventlog.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash -e
+
+# create a fake job eventlog in the KVS, no normal mechanism will
+# create an empty eventlog
+
+jobidhex=`flux job id --to=dothex 123456789`
+flux kvs put "job.${jobidhex}.eventlog"=""
```
grondo, line 7: `flux job id` supports `--to=kvs`

t/t2232-job-info-security.t
```diff
@@ -279,4 +279,61 @@ test_expect_success 'flux job wait-event guest.exec.eventlog fails via -p (live
...
+	jobidhex=`flux job id --to=dothex 123456789` &&
+	echo $submitstr | flux kvs put --raw "job.${jobidhex}.eventlog"=-
+'
+
+test_expect_success 'flux job info eventlog fails (user)' '
+	set_userid 9000 &&
+        flux job info 123456789 eventlog 2>&1 | grep "Protocol error" &&
+	unset_userid
+'
+
```
grondo, line 338: Also might be a good idea to add tests with bad data (poorly formatted JSON, binary data, etc) here.

grondo (commented):
Just a couple minor comments I noticed before any kind of full review.

chu11:
re-pushed with tweaks given @grondo comments, and I realized a few minor things along the way that i tweaked (one test was wrong, fixed up a comment, etc.)

src/modules/job-info/allow.c
```diff
@@ -47,7 +47,7 @@ static int eventlog_get_userid (struct info_ctx *ctx, const char *s,
         goto error;
     }
     if (strcmp (name, "submit") != 0 || !context) {
-        flux_log_error (ctx->h, "%s: invalid event", __FUNCTION__);
+        flux_log (ctx->h, LOG_ERR, "%s: invalid event", __FUNCTION__);
```
grondo, line 54: optional suggestion, add actual event name for easier debugging if this (highly unlikely) error is logged, i.e. `invalid event <name>`

grondo (approved):
LGTM. One comment inline, but the error is so unlikely it is completely optional.

chu11:
thanks @grondo, re-pushed with the suggested minor fix

Issue #4414, opened 2022-07-19
flux-dump: job/.../eventlog: missing blobref 0: Protocol error


garlick:
@grondo noted the following failure:

>  in a newly built I'm getting a really weird error in the newly built fluxrm/testenv:el8 image in t4000-issues-test-driver.t. It obviously didn't occur before, but seems reproducible in this particular image:
```
expecting success: run_timeout 120 /usr/src/t/issues/t4222-kvs-assume-empty-dir.sh
flux-dump: archived 5 keys
flux-dump: job/0000/009e/d300/0000/eventlog: missing blobref 0: Protocol error
not ok 26 - t4222-kvs-assume-empty-dir.sh
#	run_timeout 120 /usr/src/t/issues/t4222-kvs-assume-empty-dir.sh
```

grondo:
This is embarrassing. I was accidentally running `make check` in a repo in which I had some experimental work staged for testing the addition of an empty job eventlog in the `job-ingest` module (see #4409). Apparently the empty eventlog (or kvs commit containing the empty eventlog?) was causing this issue here (that is why it wasn't reproducible elsewhere duh).

I tried editing the failing test script `t4222-kvs-assume-empty-dir.sh` to create some empty entries in the kvs, but this didn't seem to reproduce the problem, so I'm at a loss as to what was going on here.

(Except to verify the issue is coming from here
https://github.com/flux-framework/flux-core/blob/9b97e065e1d788ab858b18515ea4176ff209b685/src/cmd/builtin/dump.c#L137-L146

where `flux_msg_get_payload()` returns `EPROTO` because the message has no payload.)

grondo:
Reproduced, apparently the APPEND is key:
```diff

diff --git a/t/issues/t4222-kvs-assume-empty-dir.sh b/t/issues/t4222-kvs-assume-empty-dir.sh
index 8fc0767ae..21f7567f6 100755
--- a/t/issues/t4222-kvs-assume-empty-dir.sh
+++ b/t/issues/t4222-kvs-assume-empty-dir.sh
@@ -1,5 +1,9 @@
 #!/bin/bash -e
 
+flux kvs put test=
+flux kvs put --append test=foo
+
+
 flux module remove kvs-watch
 flux module remove kvs
```

```console
$ ./src/cmd/flux start ./t/issues/t4222-kvs-assume-empty-dir.sh -d -v
flux-dump: archived 3 keysflux-dump: test: missing blobref 0: Protocol error
[...]

PR #4418, opened 2022-07-21
flux-dump: fix handling of empty blobref value

grondo:
This fixes the issue described in #4414 by switching `flux-dump` to use `flux_response_decode_raw()` instead of `flux_msg_get_payload()`, the former of which handles responses with an "empty" payload in the expected way.

An attempt to add the precipitating condition to the tests in `t2807-dump-cmd.t`was also made, though I stumbled around a bit trying to figure out the correct blob counts and description of what is happening in the KVS under the covers for the empty, then appended, key. Please forgive and correct any mistakes therein.

### DIFF
--- a/src/cmd/builtin/dump.c
+++ b/src/cmd/builtin/dump.c
@@ -138,7 +138,7 @@ static void dump_valref (struct archive *ar,
                                           treeobj_get_blobref (treeobj, i),
                                           content_flags))
             || flux_future_get (f, (const void **)&msg) < 0
-            || flux_msg_get_payload (msg, &data, &len) < 0) {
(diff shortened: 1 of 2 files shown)


src/cmd/builtin/dump.c
```diff
@@ -138,7 +138,7 @@ static void dump_valref (struct archive *ar,
                                           treeobj_get_blobref (treeobj, i),
```
garlick, line 138: commit message: s/lenght/length/

garlick (approved):
I think your test explanation is exactly right.
One spelling error in a commit message - o/w looks good!

grondo:
Thanks, setting MWP.

PR #4426, opened 2022-07-25
flux-queue(1): add man page

garlick:
Problem: flux-queue(1) man page is missing, but this is a useful administrator command.

Add a simple page.


grondo (approved):
LGTM!

PR #4427, opened 2022-07-25
libflux: handle flux_respond_error (errnum=0)

### LINKED ISSUE #3036: flux_respond_error: accept an errnum of 0
It was discussed on Slack that `flux_respond_error` considers `errnum` == 0 to be invalid and errors out with `EINVAL`.    From the man page: 
> flux_respond_error() returns an error response to the sender. errnum must be non-zero. If errmsg is non-NULL, an error string payload is included in the response. The error string may be used to provide a more detailed error message than can be conveyed via errnum.

This was surprising to both @grondo and I.  One suggestion was to allow an `errnum` of 0 for `flux_respond_error` since we've been shifting to using descriptive `errmsg`s anyways.  Maybe we only allow an `errnum` of 0 when an `errmsg` is provided?

garlick:
Problem: if `flux_respond_error()` is called with errnum of 0, no response is sent, which can cause hard to diagnose problems.

Use EINVAL if errnum is zero.

Fixes #3036

### DIFF
--- a/doc/man3/flux_respond.rst
+++ b/doc/man3/flux_respond.rst
@@ -51,7 +51,7 @@ building the payload using variable arguments with a format string in
 the style of jansson's ``json_pack()`` (used internally).
 
 ``flux_respond_error()`` returns an error response to the sender.
-*errnum* must be non-zero. If *errmsg* is non-NULL, an error string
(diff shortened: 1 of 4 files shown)


chu11 (approved):
LGTM

garlick:
Added a spot of test coverage and rebased on current master.

garlick:

Issue #4300, opened 2022-04-24
Flux job IDs could be reused if job manager checkpoint is lost


garlick:
Problem: the job manager writes the largest job ID used to date to  `checkpoint.job-manager` when unloading and reads it back upon restart, but if that data is lost, the job ID epoch is reset to zero.  The possibility of duplicate job IDs could cause problems for retrieval of job archive data.

It may also be useful if  *(cluster-name, jobid)* were a center wide unique identifier for a chunk of work run at the system instance level.

Idea: add an optional FLUID epoch configuration, intended only for the system instance.   For example:
```toml
[job-manager]

job-id-epoch = 2022-01-01T00:00:00-00:00
```
then if that is set, compute the FLUID starting epoch based on the current wall clock minus `job-id-epoch`.

My worry with that approach is that if the wall clock is not synchronized, flux could start issuing job IDs for the wrong epoch.  But then again, if the time is off, queries against the job archive are going to be pretty messed up anyway.

Another problem would be how to detect and/or handle a change in epoch.  Maybe the checkpoint could include the configured epoch, if any.

grondo:
Since the job manager now tracks inactive jobs, does it also process inactive jobs in the kvs on restart? Even if there are no active jobs on a restart, could the job manager therefore compute the largest jobid during restart?

I guess the problem here is if there was a purge of _all_ jobs before a restart. In that case your proposal is more foolproof, though I do worry that it is an odd bit of configuration to require.

garlick:
This is not an inactive job  purge issue since `checkpoint.job-manager` includes the last job ID issued even if all inactive jobs are purged.   I was worried about either losing the content database due to corruption, or if a sys admin decides to "start from scratch" for whatever reason and removes the content store without also clearing the job archive db.

Another thought is to have the job archive db or its replacement provide the last used job id, or (maybe slightly cleaner dependency wise) refuse to load if its last used job id is greater than the job manager's?

grondo:
Yeah, sorry - I kind of meant why does the job manager have to checkpoint the last job id when it is also reading _all_ the jobids from the kvs. Can't it just use the largest jobid after restart is complete?

I like the idea that the job archive could provide the last used jobid, but that won't work if we're restarting with running jobs, whereas using the recovered jobids would, even if rank 0 broker has crashed. I'm probably just missing some important detail.

garlick:
> Could a purge of all jobs trigger the write of the job-manager's last use jobid to the checkpoint?

[...]

PR #4443, opened 2022-07-27
job-manager: improve robustness of max job id recovery on restart

garlick:
Problem:  as noted in #4300, the job-manager only writes out the max job ID to the KVS when it is unloading, so despite best efforts to preserve KVS content (e.g. with defensive checkpoints added in #4383) when there is a crash or a failed shutdown, we are in danger of starting from an old max ID value and issuing duplicate job IDs.

This just uses the greater of the checkpoint value or the maximum job ID replayed from the KVS to improve robustness.

Also, since purging can eliminate all jobs in the KVS, we add code to update the checkpoint value in the KVS if  the job matching the current max job ID happens to be purged, as suggested by @grondo.

### DIFF
--- a/src/modules/job-manager/job-manager.c
+++ b/src/modules/job-manager/job-manager.c
@@ -219,8 +219,8 @@ int mod_main (flux_t *h, int argc, char **argv)
         flux_log_error (h, "flux_reactor_run");
         goto done;
     }
-    if (checkpoint_to_kvs (&ctx) < 0) {
-        flux_log_error (h, "checkpoint_to_kvs");
+    if (restart_save_state (&ctx) < 0) {
+        flux_log_error (h, "error saving job manager state to KVS");
         goto done;
     }
     rc = 0;
--- a/src/modules/job-manager/job-manager.c
+++ b/src/modules/job-manager/job-manager.c
@@ -76,12 +76,12 @@ static void stats_cb (flux_t *h, flux_msg_handler_t *mh,
 {
     struct job_manager *ctx = arg;
     int journal_listeners = journal_listeners_count (ctx->journal);
-    if (flux_respond_pack (h, msg, "{s:{s:i} s:i s:i}",
+    if (flux_respond_pack (h, msg, "{s:{s:i} s:i s:i s:I}",
                            "journal",
                              "listeners", journal_listeners,
                            "active_jobs", zhashx_size (ctx->active_jobs),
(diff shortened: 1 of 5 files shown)


src/modules/job-manager/restart.h
```diff
@@ -20,6 +20,8 @@ int restart_from_kvs (struct job_manager *ctx);
 /* exposed for unit testing only */
 int restart_count_char (const char *s, char c);
 
+int checkpoint_to_txn (struct job_manager *ctx, flux_kvs_txn_t *txn);
```
chu11, line 23: maybe it's just me, but this sounds you're checkpointing into the txn?  which makes no sense.  Hmmm.

`save/write/add_checkpoint_to_txn()`?  I'd say add is the best of this bunch.
`txn_checkpoint_max_jobid()`?

maybe there's a better one.

garlick: Yeah I wasn't too happy with the function names either.  OK, I tacked on an earlier commit that renames `checkpoint_to_kvs()` to `restart_save_state()` and then made the new function `restart_save_state_to_txn()`.  Hopefully that works a little better, and also normalizes all the public functions in restart.c to have a `restar_*` prefix, and avoids the use of "checkpoint" which recently has become a bit of a loaded term.

src/modules/job-manager/job-manager.c
```diff
@@ -219,8 +219,8 @@ int mod_main (flux_t *h, int argc, char **argv)
         flux_log_error (h, "flux_reactor_run");
         goto done;
     }
-    if (checkpoint_to_kvs (&ctx) < 0) {
-        flux_log_error (h, "checkpoint_to_kvs");
+    if (restart_save_state (&ctx) < 0) {
```
grondo, line 222: Commit message typo: "confusign"

src/modules/job-manager/restart.c
```diff
@@ -197,6 +197,8 @@ static int restart_map_cb (struct job *job, void *arg, flux_error_t *error)
                    (uintmax_t)job->id);
         return -1;
     }
+    if (ctx->max_jobid < job->id)
```
grondo, line 200: Given the previous rename of "checkpoint" to "restart", should/could the commit message here be updated too?

grondo (approved):
Nice! LGTM. Just a couple minor comments inline (one is more of a question)

garlick:
Thanks!  I fixed those things and forced a push.  I'll set MWP.

PR #4444, opened 2022-07-28
flux-config-bootstrap(5): improve hosts description

### LINKED ISSUE #4408: doc: flux-config-bootstrap(7) description of `hosts` key is unclear
A user attempting to configure a small system instance was confused by the description of the `boostrap.hosts` key in `flux-config-bootstrap(5)`:
```
       hosts  (optional) A rank-ordered array of host entries. Each host entry
              is  a  TOML table containing at minimum the host key. The broker
              determines its rank by matching its hostname in the hosts  array
              and  taking  the  array  index.  An empty or missing hosts array
              implies a standalone (single broker) instance. The entry  for  a
              broker  with downstream peers must either assign the bind key to
              a ZeroMQ endpoint URI, or the default_bind URI  described  above
              is  used. The entry for a broker with downstream peers must also
              either assign the connect key to a ZeroMQ endpoint URI,  or  the
              default_connect  URI described above is used. The same %h and %p
              substitutions work here as well.

```

The situation could perhaps be improved by moving the description of `hosts` array entries to its own section, and then only briefly describing the _purpose_ of the `hosts` array in the description entry in the KEYS section.

garlick:
Problem: there is too much prose crammed into the key table
hosts entry.

Add some background information to the DESCRIPTION section,
so that the key descriptions can be made more concise, then
break out a section on host entry keys.

Convert the WARNING paragraph in the description to a restructured
text warning box.

Fixes #4408

### DIFF
--- a/doc/man5/flux-config-bootstrap.rst
+++ b/doc/man5/flux-config-bootstrap.rst
@@ -10,11 +10,33 @@ The system instance requires that the overlay network configuration be
 statically configured.  The ``bootstrap`` TOML table defines these details
 for a given cluster.
 
-WARNING:  Although ``flux config reload`` works on a live system, these
-settings do not take effect until the next broker restart.  As such, they
-must only be changed in conjunction with a full system instance restart in
-order to avoid brokers becoming desynchronized if they are independently
-restarted before the next instance restart.
+.. warning::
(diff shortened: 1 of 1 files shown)


chu11 (approved):
LGTM

garlick:
Thanks - setting MWP.

PR #4449, opened 2022-07-29
flux-top: support split of inactive jobs into completed and failed

grondo:
@woodard had requested splitting the flux-top(1) inactive job stats into successful vs failed for easier monitoring of a test workload.

I already had this lying around, so I thought I'd propose it. I think @garlick had something similar but in a different form? We could use that instead, but I could not find it.

This version keeps the default display unless you are connecting to the instance as the owner, then details are displayed by default. In either case, the display can be toggled with the `d` key. The end result is fairly simple:

Screenshot from 2022-07-26 09-53-17

### DIFF
--- a/doc/man1/flux-top.rst
+++ b/doc/man1/flux-top.rst
@@ -34,6 +34,9 @@ j, down-arrow
 k, up-arrow
    Move cursor up in the job listing.
 
+d
+   Toggle display of inactive job details (failed vs successful jobs).
+
 enter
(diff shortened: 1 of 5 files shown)


garlick:
Just tried this and works great!  I quite like this.  It _does_ seem like I prototyped this once, but I can't find it either :shrug:

garlick (approved):
LGTM!  I checked on the system instance to confirm the reverse behavior.  All god!

grondo:
Thanks! setting MWP.

PR #4450, opened 2022-07-29
add flux_open_ex(3)

grondo:
This stack of changes is prereq for some other, much simpler, work, but the addition of `flux_open_ex(3)` seemed like a good place to stop to ensure this new function gets documented in release notes.

Currently, `flux_open(3)` may emit errors to stderr, especially when using the ssh connector, since all errors from `ssh` will go to the current terminal's `stderr`. This can be inconvenient for some users of libflux, since libraries typically aren't expected to write directly to `stderr`.

This PR first augments `popen2()` to allow capture of `stderr` (along with other minor cleanup), adds some other helper functions, then adds a new `flux_open_ex(3)` interface to libflux which takes an optional `flux_error_t` parameter in which errors can be returned to the caller. Then, the `ssh` connector is updated to capture stderr of the underlying `ssh` command on failure with the new `popen2()` functionality, and copies this back to the user (possibly truncated). Specific errors are added in a few other scenarios, but this wasn't done exhaustively for now to save time.

Finally, the Python `flux.Flux()` implementation is updated to use `flux_open_ex()`, which is probably the right thing, and also has the side effect of testing the interface (though presently there are no specific tests, the existing `flux_open()` tests are assumed to handle most of the testing since it calls `flux_open_ex()`).

This is a WIP because I'm still pondering how to add some specific tests.

### DIFF
--- a/src/common/libflux/handle.c
+++ b/src/common/libflux/handle.c
@@ -32,6 +32,7 @@
 #include "src/common/libutil/aux.h"
 #include "src/common/libutil/errno_safe.h"
 #include "src/common/libutil/monotime.h"
+#include "src/common/libutil/errprintf.h"
 
 #include "handle.h"
 #include "reactor.h"
@@ -218,7 +219,9 @@ static char *find_file (const char *name, const char *searchpath)
     return path;
 }
 
-static connector_init_f *find_connector (const char *scheme, void **dsop)
+static connector_init_f *find_connector (const char *scheme,
+                                         void **dsop,
+                                         flux_error_t *errp)
 {
     char name[PATH_MAX];
     const char *searchpath = getenv ("FLUX_CONNECTOR_PATH");
@@ -229,18 +232,22 @@ static connector_init_f *find_connector (const char *scheme, void **dsop)
     if (!searchpath)
         searchpath = flux_conf_builtin_get ("connector_path", FLUX_CONF_AUTO);
     if (snprintf (name, sizeof (name), "%s.so", scheme) >= sizeof (name)) {
+        errprintf (errp, "Connector path too long");
         errno = ENAMETOOLONG;
         return NULL;
     }
     if (!(path = find_file (name, searchpath))) {
+        errprintf (errp, "Unable to find connector name '%s'", scheme);
         errno = ENOENT;
         goto error;
     }
     if (!(dso = dlopen (path, RTLD_LAZY | RTLD_LOCAL | FLUX_DEEPBIND))) {
+        errprintf (errp, "dlopen: %s: %s", path, dlerror ());
         errno = EINVAL;
         goto error;
     }
     if (!(connector_init = dlsym (dso, "connector_init"))) {
+        errprintf (errp, "%s: missing connector_init symbol", path);
         errno = EINVAL;
         goto error_dlopen;
     }
@@ -262,7 +269,7 @@ static char *strtrim (char *s, const char *trim)
     return *s ? s : NULL;
 }
 
-flux_t *flux_open (const char *uri, int flags)
+flux_t *flux_open_ex (const char *uri, int flags, flux_error_t *errp)
 {
     char *default_uri = NULL;
     char *path = NULL;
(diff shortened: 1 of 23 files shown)


garlick (commented):
This is excellent cleanup!  Really glad you took the time to do this.
I just gave this a skim without finding anything.  When you take the WIP off I'll make a more detailed pass.

chu11:
high level comment, why `popen2_ex()`?  why not break ABI and add flags to `popen2()` since its already a utility function and used in just a few places in flux?

src/common/libutil/strstrip.c
```diff
@@ -0,0 +1,78 @@
...
+        return NULL;
+
+    size = strlen (result);
+
+    end = result + size - 1;
+    while (end >= result && isspace ((unsigned char) *end))
+        end--;
+    *(end + 1) = '\0';
+
+    return result;
```
chu11, line 74: i think this is safe, but size could be 0 after the front striping of spaces (i.e. string of all spaces), so we're sort of doing pointer arithmetic with an invalid pointer.  Perhaps add a ` if (s[0] == '\0')` check?

src/common/libutil/errprintf.h
```diff
@@ -29,6 +30,12 @@ int errprintf (flux_error_t *errp, const char *fmt, ...)
 int verrprintf (flux_error_t *errp, const char *fmt, va_list ap);
 
 
+static inline void err_init (flux_error_t *errp)
+{
+    if (errp)
+        memset (errp->text, 0, sizeof (errp->text));
+}
+
```
chu11, line 38: need a unit test?

src/common/libutil/strstrip.c
grondo: Yeah, if `size == 0` then `end = result - 1` so `*(end + 1) = '\0'` just reterminates `result`. I see what you mean, we might as well do this:
```c
    if ((size = strlen (result)) == 0)
        return result;
```

grondo:
> high level comment, why popen2_ex()? why not break ABI and add flags to popen2() since its already a utility function and used in just a few places in flux?

Eh, I was probably just being lazy - adding the new interface allowed me to not have to touch the existing unit, tests where there are quite a few calls. However, in retrospect you're probably right so I'll just add flags to `popen2()`

grondo:
Ok, I've addressed @chu11's comments and pushed the result. Included now are some basic tests for `flux_open_ex(3)` so I'll remove WIP.

src/common/libflux/handle.c
```diff
@@ -278,6 +285,8 @@ flux_t *flux_open (const char *uri, int flags)
                           | FLUX_O_TEST_NOSUB
                           | FLUX_O_RPCTRACK;
 
+    err_init (errp);
+
```
garlick, line 289: Are we going to make callers do something like`error.text[0] ? error.text : strerror (errno)` to decode errors?

Seems like the `flux_error_t` should always be filled when provided and there is an error.

garlick (commented):
Big improvement.

I did want to get your take on my comment about allways filling `flux_error_t` before marking approved though.

src/common/libflux/handle.c
grondo: Good point. Sorry that was short sighted. I'll fix it if possible

src/common/libutil/popen2.c
```diff
@@ -106,6 +131,10 @@ struct popen2_child *popen2 (const char *path, char *const argv[])
         saved_errno = errno;
         goto error;
     }
+    if ((p->flags & POPEN2_CAPTURE_STDERR) && pipe (p->efd) < 0) {
+        saved_errno = errno;
+        goto error;
+    }
```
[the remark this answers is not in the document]
grondo: Not intentionally, though all fds are closed before exec in the child process by 
```c
    if (fdwalk (popen2_child_close_fd, p)) {
```
so it could be redundant to also set close-on-exec. I'll add it for consistency though.

src/common/libutil/uri.c
```diff
@@ -65,7 +65,7 @@ char *uri_resolve (const char *uri)
         goto out;
     nullify_newline (result);
 out:
-    if (pclose2 (child) < 0) {
+    if (pclose2 (child) != 0) {
```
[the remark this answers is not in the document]
grondo: Yeah, that makes sense

grondo:
Thanks for the good comments @garlick and @chu11. I've now addressed @garlick's comments and force-pushed the result (hope that was ok!). There was some rebasing so take a quick look to verify everything's in its place.

garlick (approved):
LGTM!

grondo:
Thanks! I'll set MWP

PR #4452, opened 2022-07-29
flux-top: avoid premature exit on recursive top error

grondo:
Problem: When recursive top_create() fails due to a `flux_open()` error, e.g. the user cannot access the job's URI, top exits with an error when the intention is just to fall back to the current top instance.

This PR removes some calls to `fatal()` in `top_create()` and instead passes errors back via a `flux_error_t`. On error, a popup dialog is displayed instead of raising a fatal error. The dialog is dismissed after 4s or if any key is pressed.

This PR makes use of `flux_open_ex(3)`, so is based on top of #4450. In addition, the internal `uri_resolve()` function is augmented similarly to `flux_open_ex(3)` so that errors to `stderr` from `popen2()` of `flux uri` are captured.

Example of error popup:

Screenshot from 2022-07-29 11-10-17

### DIFF
--- a/src/cmd/builtin/proxy.c
+++ b/src/cmd/builtin/proxy.c
@@ -361,7 +361,7 @@ static int cmd_proxy (optparse_t *p, int ac, char *av[])
         optparse_fatal_usage (p, 1, "URI argument is required\n");
 
     target = av[optindex++];
-    if (!(uri = uri_resolve (target)))
+    if (!(uri = uri_resolve (target, NULL)))
         log_msg_exit ("Unable to resolve %s to a URI", target);
 
     if (optparse_hasopt (p, "reconnect"))
--- a/src/cmd/builtin/shutdown.c
+++ b/src/cmd/builtin/shutdown.c
@@ -53,7 +53,7 @@ static int subcmd (optparse_t *p, int ac, char *av[])
     }
 
     if (target) {
-        char *uri = uri_resolve (target);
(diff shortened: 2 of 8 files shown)


src/common/libflux/handle.h
```diff
@@ -77,11 +77,11 @@ enum {
...
 flux_t *flux_open (const char *uri, int flags);
 
-/* Like flux_open(), but if optional flux_error_t parameter is non-NULL,
+/* Like flux_open(), but if optional 'error' parameter is non-NULL,
  * then any errors normally emitted to stderr will instead be returned
- * in errp->text.
+ * in error->text.
  */
-flux_t *flux_open_ex (const char *uri, int flags, flux_error_t *errp);
+flux_t *flux_open_ex (const char *uri, int flags, flux_error_t *error);
```
chu11, line 84: looks like a rebase error that lead to this change?  minimally should not be in this commit.

grondo: Yes, thanks for catching that! Fixed.

grondo:
Rebased after merge of #4450. Also added a test that ensures desired functionality. Hopefully the test is robust enough to survive ci.

garlick (approved):
LGTM!  I was able to get the error box to show up on my test cluster by starting a batch job then removing its  local-0 socket  and navigating to it with top.  Works as advertised!

grondo:
Thanks, setting MWP.

PR #4454, opened 2022-08-01
flux-relay: initialize log prefix to hostname when possible

grondo:
When the ssh connector fails to connect to a remote uri, but the ssh was successfully able to execute `flux-relay`, an error such as the following is displayed
```
flux-relay: local:///tmp/flux-xoZRkf/local-0: Permission denied
```
The `flux-relay` prefix probably isn't all that helpful for users, and may be confusing, since `flux-relay` was only indirectly run as part of the `ssh` connector. Instead it may be more useful to use the hostname as the command prefix. This way, the user is made aware of the host on which the local connection failed e.g

```
fluke89: local:///tmp/flux-xoZRkf/local-0: Permission denied
```

### DIFF
--- a/src/cmd/builtin/relay.c
+++ b/src/cmd/builtin/relay.c
@@ -131,8 +131,18 @@ static int cmd_relay (optparse_t *p, int ac, char *av[])
     flux_t *h;
     int optindex;
     char *uri;
+    char hostname [HOST_NAME_MAX + 1];
 
-    log_init ("flux-relay");
+    /*  If possible, initalize logging prefix as local hostname. (In the
+     *  unlikely event gethostname(3) fails, use "uknown-host".)
+     *
+     *  This will be more helpful than a literal "flux-relay" logging prefix
+     *  for end users that may be uknowningly using flux-relay as part of
+     *  the ssh connector.
+     */
+    log_init ("uknown-host");
+    if (gethostname (hostname, sizeof (hostname)) == 0)
+        log_init (hostname);
 
     optindex = optparse_option_index (p);
     if (optindex == ac)


src/cmd/builtin/relay.c
```diff
@@ -131,8 +131,18 @@ static int cmd_relay (optparse_t *p, int ac, char *av[])
...
+    char hostname [MAXHOSTNAMELEN + 1];
 
-    log_init ("flux-relay");
+    /*  If possible, initalize logging prefix as local hostname.
+     *
+     *  This will be more helpful than a literal "flux-relay" logging prefix
+     *  for end users that may be uknowningly using flux-relay as part of
+     *  the ssh connector.
+     */
+    if (gethostname (hostname, MAXHOSTNAMELEN) < 0)
```
garlick, line 142: By my read of [gethostname(1)](https://man7.org/linux/man-pages/man2/gethostname.2.html) the length parameter should be the size of the buffer not the maximum string length, so use MAXHOSTNAMELEN + 1, or maybe better, `sizeof (hostname)` here?

Also, not sure if this is important but the man page references HOST_NAME_MAX not MAXHOSTNAME.  (Might be an alias but the former is apparently the POSIX one).  Looking around our code base we use both :-(

Another thought is to use "unknown-host" or similar as the fallback if `gethostname()` fails (very unlikely).  Could just statically initialize hostname[] to that value and avoid the duplicate call to `log_init()`.

garlick (approved):
Good idea!  This is fine as is, although I did make some nitpicky comments about `gethostname()` that likely won't amount to much improvement.

src/cmd/builtin/relay.c
grondo: Yeah, I did also see the `HOST_NAME_MAX` reference in the man page as well, but when I spot checked our codebase I saw use of `MAXHOSTNAMELEN`. Rather than fix it now, I chose `MAXHOSTNAMELEN` for consistency (though I guess I missed the uses of `HOST_NAME_MAX`)

grondo:
@garlick, I addressed your comments. I switched to `HOST_NAME_MAX` assuming that at some point we might want to standardize on that, though now reading [POSIX - limits.h](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/limits.h.html), I see that `HOST_NAME_MAX` is in the section "Runtime Invariant Values (Possibly Indeterminate)" which has the statement: 

> A definition of one of the symbolic constants in the following list shall be omitted from <limits.h> on specific  implementations where the corresponding value is equal to or greater than the stated minimum, but is unspecified.
>

Therefore it seems like existence of `HOST_NAME_MAX` is not guaranteed. I suppose we could handle that in configure.
> This indetermination might depend on the amount of available memory space on a specific instance of a specific implementation. The actual value supported by a specific instance shall be provided by the sysconf() function.

garlick (approved):
Sheesh, they make a simple thing so complicated.
Oh well, this looks good to me!

grondo:
Thanks! Setting MWP.

Issue #4456, opened 2022-08-01
content-{sqlite,files,s3}: register backing store after registering services


chu11:
While working on #4267, realized that ENOSYS errors encountered are due to the fact that the `content-{sqlite,files,s3}` modules all register themselves as backing stores BEFORE registering their services.  This makes it so that `cache_flush()` and similar things in the `content-cache` cannot work during the registration process.

The calls should be reversed.

The issues target in PR #4380 were specifically different, but highly related to this issue.

PR #4458, opened 2022-08-01
content-sqlite,files,s3: register backing module after setup is complete

chu11:
Problem: In the content-sqlite, content-files, and content-s3 modules,
the backing module is registered before setup of module services is
complete.  This can cause problems if the broker's content-cache
wants to make calls into the module during registration time.

Solution: Ensure that all setup is complete before registering the
backing module.

Update regression test which would fail with this change.

Fixes https://github.com/flux-framework/flux-core/issues/4456

### DIFF
--- a/src/modules/content-s3/content-s3.c
+++ b/src/modules/content-s3/content-s3.c
@@ -497,12 +497,12 @@ int mod_main (flux_t *h, int argc, char **argv)
         flux_log_error (h, "content_s3_create failed");
         return -1;
     }
+    if (content_register_service (h, "content-backing") < 0)
+        goto done;
+    if (content_register_service (h, "kvs-checkpoint") < 0)
+        goto done;
     if (content_register_backing_store (h, "content-s3") < 0)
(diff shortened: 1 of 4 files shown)


src/modules/content-s3/content-s3.c
```diff
@@ -500,17 +500,16 @@ int mod_main (flux_t *h, int argc, char **argv)
...
+        goto done_unreg;
     if (content_register_service (h, "kvs-checkpoint") < 0)
-        goto done;
+        goto done_unreg;
     if (flux_reactor_run (flux_get_reactor (h), 0) < 0) {
         flux_log_error (h, "flux_reactor_run");
-        goto done;
+        goto done_unreg;
     }
-    if (content_unregister_backing_store (h) < 0)
```
garlick, line 510: I think this is already done automatically (unregistering services on module unload).
However it doesn't hurt to do it explicitly.

garlick (approved):
LGTM!

PR #4459, opened 2022-08-02
flux-overlay: add man page, open to guest users

garlick:
Problem: `flux overlay status` is restricted to guests for no particular reason, requiring sys admins to run under sudo.

Make the necessary RPCs available to guest users.

Also, add missing man page, and fix a bug in the `flux overlay disconnect` output (still restricted to instance owner).

### DIFF
--- a/src/broker/overlay.c
+++ b/src/broker/overlay.c
@@ -1984,7 +1984,7 @@ static const struct flux_msg_handler_spec htab[] = {
     },
     {
         FLUX_MSGTYPE_REQUEST,
-        "overlay.disconnect",
+        "overlay.disconnect", // clean up after 'flux overlay status --wait'
         disconnect_cb,
         FLUX_ROLE_USER,
     },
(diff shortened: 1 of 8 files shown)


src/broker/overlay.c
```diff
@@ -1974,25 +1974,25 @@ static const struct flux_msg_handler_spec htab[] = {
...
         overlay_health_cb,
-        0
+        FLUX_ROLE_USER,
     },
     {
         FLUX_MSGTYPE_REQUEST,
         "overlay.disconnect",
         disconnect_cb,
-        0
+        FLUX_ROLE_USER,
```
grondo, line 1989: Am I reading this wrong, or does this allow a user to force a broker disconnect?

grondo (commented):
Great! More documentation always good!

I did have one question about the `overlay.disconnect` RPC before setting approval.

src/broker/overlay.c
garlick: No,  that is for cleaning up `flux overlay status --wait=STATE` clients that disconnect before STATE is reached.

The RPC topic for `flux overlay disconnect` is `overlay.disconnect-subtree`.

I'll add a comment b/c that is confusing!

grondo: Ah! That makes sense now (I forgot momentarily about the `*.disconnect` handlers)

grondo (approved):
LGTM!

garlick:
Thanks!  I'll set MWP.

PR #4463, opened 2022-08-03
content-{sqlite,files,s3}: route checkpoint-get and checkpoint-put through broker

chu11:
As per-requisite support for issue #4267, we will need to route checkpoints through the broker into the content backing modules.

There is a nice chunk of cleanup to do before this can happen, mostly surrounding topic names.  So this is a bit of a pre-requisite PR I'm splitting off from the main PR since it doesn't really do much but touches a lot of lines of code.

Note that in the last commit, there is some code that is "common" and would look like I should add a common function.  But they will be modified shortly so I left them separate.  I could clean that up if desired.

### DIFF
--- a/src/broker/content-cache.c
+++ b/src/broker/content-cache.c
@@ -340,13 +340,12 @@ static void cache_entry_remove (struct content_cache *cache,
 {
     assert (e->load_requests == NULL);
     assert (e->store_requests == NULL);
+    assert (!e->dirty);
     list_del (&e->list);
     if (e->valid) {
         cache->acct_size -= e->len;
         cache->acct_valid--;
     }
-    if (e->dirty)
-        cache->acct_dirty--;
     zhashx_delete (cache->entries, e->hash);
 }
 
--- a/src/broker/content-cache.c
+++ b/src/broker/content-cache.c
@@ -605,6 +605,124 @@ error:
         flux_log_error (h, "content store: flux_respond_error");
 }
 
+static void checkpoint_get_continuation (flux_future_t *f, void *arg)
+{
+    struct content_cache *cache = arg;
+    const flux_msg_t *msg = flux_future_aux_get (f, "msg");
+    const char *s;
+
+    assert (msg);
+
+    if (flux_rpc_get (f, &s) < 0)
+        goto error;
+
+    if (flux_respond (cache->h, msg, s) < 0)
+        flux_log_error (cache->h, "%s: flux_respond", __FUNCTION__);
+    flux_future_destroy (f);
+    return;
+
+error:
+    if (flux_respond_error (cache->h, msg, errno, NULL) < 0)
+        flux_log_error (cache->h, "flux_respond_error");
+    flux_future_destroy (f);
+}
+
+void content_checkpoint_get_request (flux_t *h, flux_msg_handler_t *mh,
+                                     const flux_msg_t *msg, void *arg)
+{
+    struct content_cache *cache = arg;
+    const char *topic = "content-backing.checkpoint-get";
+    const char *s = NULL;
+    const flux_msg_t *msgcpy = flux_msg_incref (msg);
+    flux_future_t *f = NULL;
+
+    /* Temporarily maintain ENOSYS behavior */
+    if (!cache->backing) {
+        errno = ENOSYS;
+        goto error;
+    }
+
+    if (flux_request_decode (msg, NULL, &s) < 0)
+        goto error;
+
+    if (!(f = flux_rpc (h, topic, s, 0, 0))
+        || flux_future_aux_set (f,
+                                "msg",
+                                (void *)msgcpy,
+                                (flux_free_f)flux_msg_decref) < 0
+        || flux_future_then (f, -1, checkpoint_get_continuation, cache) < 0) {
+        flux_log_error (h, "error starting checkpoint-get RPC");
+        goto error;
+    }
+
+    return;
+
+error:
+    if (flux_respond_error (h, msg, errno, NULL) < 0)
+        flux_log_error (h, "error responding to checkpoint-get request");
+    flux_future_destroy (f);
+    flux_msg_decref (msgcpy);
+}
+
+static void checkpoint_put_continuation (flux_future_t *f, void *arg)
+{
+    struct content_cache *cache = arg;
+    const flux_msg_t *msg = flux_future_aux_get (f, "msg");
+    const char *s;
+
+    assert (msg);
+
+    if (flux_rpc_get (f, &s) < 0)
+        goto error;
+
+    if (flux_respond (cache->h, msg, s) < 0)
+        flux_log_error (cache->h, "%s: flux_respond", __FUNCTION__);
+    flux_future_destroy (f);
+    return;
+
+error:
+    if (flux_respond_error (cache->h, msg, errno, NULL) < 0)
+        flux_log_error (cache->h, "flux_respond_error");
+    flux_future_destroy (f);
+}
+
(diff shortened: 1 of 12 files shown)


src/broker/content-cache.c
```diff
@@ -340,13 +340,12 @@ static void cache_entry_remove (struct content_cache *cache,
...
     assert (e->load_requests == NULL);
     assert (e->store_requests == NULL);
+    assert (!e->dirty);
     list_del (&e->list);
     if (e->valid) {
         cache->acct_size -= e->len;
         cache->acct_valid--;
     }
-    if (e->dirty)
-        cache->acct_dirty--;
```
garlick, line 349: Ah sorry, dead code there.

Note "decrement" is misspelled in the commit description.

src/common/libkvs/kvs_checkpoint.c
```diff
@@ -37,7 +37,7 @@ flux_future_t *kvs_checkpoint_commit (flux_t *h,
         timestamp = flux_reactor_now (flux_get_reactor (h));
 
     if (!(f = flux_rpc_pack (h,
-                             "kvs-checkpoint.put",
+                             "content-backing.checkpoint-put",
```
garlick: Ah good cleanup.  This was initially created as a separate service since it seemed like it ought to be independent of the "content" service, but that's not how it has played out so this change make sense and in fact has been bugging me.

src/broker/content-cache.c
```diff
@@ -605,6 +605,133 @@ static void content_store_request (flux_t *h, flux_msg_handler_t *mh,
...
+    const char *s;
+
+    assert (msg);
+
+    if (flux_rpc_get (f, &s) < 0)
+        goto error;
+
+    if (flux_respond (cache->h, msg, s) < 0) {
+        flux_log_error (cache->h, "%s: flux_respond", __FUNCTION__);
+        goto error;
```
garlick, line 621: Don't call `flux_respond_error()` to handle a `flux_respond()` error (it is likely to fail).

src/broker/content-cache.c
```diff
@@ -605,6 +605,133 @@ static void content_store_request (flux_t *h, flux_msg_handler_t *mh,
...
+        flux_log_error (cache->h, "flux_respond_error");
+    flux_future_destroy (f);
+}
+
+/* flux_free_f type */
+static void flux_msg_decref_wrapper (void *arg)
+{
+    const flux_msg_t *msg = arg;
+    flux_msg_decref (msg);
+}
```
garlick, line 638: Just cast `flux_msg_decref()` to flux_free_f instread?

src/broker/content-cache.c
```diff
@@ -605,6 +605,133 @@ static void content_store_request (flux_t *h, flux_msg_handler_t *mh,
...
+    if (flux_request_decode (msg, NULL, &s) < 0)
+        goto error;
+
+    if (!(f = flux_rpc (h, topic, s, 0, 0))
+        || flux_future_aux_set (f,
+                                "msg",
+                                (void *)flux_msg_incref (msg),
+                                flux_msg_decref_wrapper) < 0
+        || flux_future_then (f, -1, checkpoint_get_continuation, cache) < 0) {
+        flux_log_error (h, "%s: checkpoint backing get", __FUNCTION__);
```
garlick, line 663: The error message would be "broker: content_checkpoint_get_request: checkpoint backing get: ...".

Maybe just have it be "error starting checkpoint-get RPC: ..."

Also: this leaks `msg` if the rpc succeeds but the aux_set fails.

src/broker/content-cache.c
```diff
@@ -605,6 +605,133 @@ static void content_store_request (flux_t *h, flux_msg_handler_t *mh,
...
+    const char *s;
+
+    assert (msg);
+
+    if (flux_rpc_get (f, &s) < 0)
+        goto error;
+
+    if (flux_respond (cache->h, msg, s) < 0) {
+        flux_log_error (cache->h, "%s: flux_respond", __FUNCTION__);
+        goto error;
```
garlick, line 688: When `flux_respond()` fails, don't try to send an error response (it is likely to fail).  Logging the unlikely error is sufficient.

src/broker/content-cache.c
```diff
@@ -605,6 +605,133 @@ static void content_store_request (flux_t *h, flux_msg_handler_t *mh,
...
+        || flux_future_then (f, -1, checkpoint_get_continuation, cache) < 0) {
+        flux_log_error (h, "%s: checkpoint backing get", __FUNCTION__);
+        goto error;
+    }
+
+    return;
+
+error:
+    if (flux_respond_error (h, msg, errno, NULL) < 0)
+        flux_log_error (h, "flux_respond_error");
```
garlick, line 671: Instead of "broker: flux_respond_error..." which would be infuriatingly non specific, maybe "error responding to checkpoint-get request" or similar?  I realize there's a lot of examples of the former in the code base but I think in this case it's better to be inconsistent and start doing better rather rather than defer fixing these until we can fix all of them.

src/broker/content-cache.c
```diff
@@ -605,6 +605,133 @@ static void content_store_request (flux_t *h, flux_msg_handler_t *mh,
...
+        goto error;
+
+    if (!(f = flux_rpc (h, topic, s, 0, 0))
+        || flux_future_aux_set (f,
+                             "msg",
+                             (void *)flux_msg_incref (msg),
+                             flux_msg_decref_wrapper) < 0
+        || flux_future_then (f, -1, checkpoint_put_continuation, cache) < 0) {
+        flux_log_error (h, "%s: checkpoint backing put", __FUNCTION__);
+        goto error;
```
garlick, line 724: Same comments as other function: leaks msg on error, tries to send an error response in response to a response error, unnecessarily vague log message if `flux_respond_error()` fails.

garlick (commented):
Given that the KVS checkpiont functionality is being folded into the content service, does it make sense to prefix the functions with `content_` instead of `kvs_` and add a flags option to allow the cache to be bypassed like the content functions?

The dump/restore commands have a `--no-cache` option for bypassing the broker cache.  This would let them trivially follow suit with the checkpoint functions.

I'll stop short of suggesting that libcontent be merged into libkvs since it only contains non-public helpers and the content service now more or less exists only for the kvs, but maybe that's food for thought for a future cleanup.

chu11:
> Given that the KVS checkpiont functionality is being folded into the content service, does it make sense to prefix the functions with content_ instead of kvs_ and add a flags option to allow the cache to be bypassed like the content functions?

You're referring to the `libkvs` checkpoint functions, right?

Hmmm.  I like the idea of adding the bypass flag, but wondering if we should keep them prefixed as `kvs_`, solely b/c of the `KVS_DEFAULT_CHECKPOINT` key.  Effectively they act as a wrapper for this special case checkpointing.

chu11:
> Hmmm. I like the idea of adding the bypass flag, but wondering if we should keep them prefixed as kvs_, solely b/c of the KVS_DEFAULT_CHECKPOINT key. Effectively they act as a wrapper for this special case checkpointing.

I'm going to actually do this in a separate PR leaving this one isolated to what it currently does.  I'll write up an issue.

src/broker/content-cache.c
chu11: ugh, get lots of warnings b/c the `flux_msg_incref/decref()` functions are const (which is probably why I wrote the wrapper to begin with and forgot)

chu11:
re-pushed with fixes for comments from above, went ahead and squashed them since they were all relatively minor

garlick (approved):
Sorry, should have approved last time!  LGTM!

Issue #4465, opened 2022-08-04
job-list: segfault on fluke after `flux jobs -aA` on login node


grondo:
After running `flux jobs -aA` on the login node, the rank 0 broker segfaulted in `job-list`:
```
(gdb) where                                                                                                                                                                  
#0  0x00007ffff6330cc7 in __strlen_avx2 () from /lib64/libc.so.6                                                                                                             
#1  0x00007ffff6bb7a12 in json_string () from /lib64/libjansson.so.4                                                                                                         
#2  0x00007fffd15e886f in store_attr (job=job@entry=0x7fffb4592f60, attr=<optimized out>, o=o@entry=0x7fffb49c3b10, errp=errp@entry=0x7fffcaeb9720) at job_util.c:83         
#3  0x00007fffd15e8d4c in job_to_json (job=job@entry=0x7fffb4592f60, attrs=attrs@entry=0x7fffb43007b0, errp=errp@entry=0x7fffcaeb9720) at job_util.c:239                     
#4  0x00007fffd15e73c6 in get_jobs_from_list (jobs=jobs@entry=0x7fffb48bd5b0, errp=errp@entry=0x7fffcaeb9720, list=0x7fffb40103b0, max_entries=max_entries@entry=1000,       
    attrs=attrs@entry=0x7fffb43007b0, userid=userid@entry=4294967295, states=126, results=15) at list.c:70                                                                   
#5  0x00007fffd15e7539 in get_jobs (ctx=ctx@entry=0x7fffb4010140, errp=errp@entry=0x7fffcaeb9720, max_entries=1000, attrs=0x7fffb43007b0, userid=4294967295, states=126,     
    results=15) at list.c:139                                                                                                                                                
#6  0x00007fffd15e7732 in list_cb (h=0x7fffb40061d0, mh=<optimized out>, msg=0x7fffb44d6be0, arg=0x7fffb4010140) at list.c:207                                               
#7  0x00007ffff7b61d07 in call_handler (mh=0x7fffb40101e0, msg=msg@entry=0x7fffb44d6be0) at msg_handler.c:344                                                                
#8  0x00007ffff7b6233b in dispatch_message (type=1, msg=0x7fffb44d6be0, d=0x7fffb400f1b0) at msg_handler.c:380                                                               
#9  handle_cb (r=0x7fffb400e8a0, hw=<optimized out>, revents=<optimized out>, arg=0x7fffb400f1b0) at msg_handler.c:481                                                       
[...]

PR #4462, opened 2022-08-03
github: fixes for issues found when pushing a tag

grondo:
This PR (now) fixes a couple issues found when pushing the last flux-core tag:

 - There was still one remaining issue with the dumpfile creation step when run on a tag
 - The auto-deploy steps seem to trigger when pushing a tag to a fork, not just the flux-framework/flux-core repo

PR #4470, opened 2022-08-04
job-list: avoid segfault after job purge

garlick:
@grondo and I ran this fix to #4465, which crashed the rank 0 broker on fluke twice this morning.

I'll see if I can come up with a test.  It's a use-after-free so the test may not be a reliable reproducer unless run under valgrind.  That's still probably useful I guess.

Edit: I did confirm that the valgrind test described in the issue (which involves restart from fluke's kvs dump) no longer produces errors with this fix.

### DIFF
--- a/src/modules/job-list/job_state.c
+++ b/src/modules/job-list/job_state.c
@@ -1146,6 +1146,22 @@ done:
     return rc;
 }
 
+/* Call zlistx_sort() on the list, then reacquire the list handles since
+ * they are no longer valid after sorting (flux-framework/flux-core#4465).
+ */
+static void sort_job_list (zlistx_t *list)
+{
+    struct job *job;
+
+    zlistx_sort (list);
+
+    job = zlistx_first (list);
+    while (job) {
+        job->list_handle = zlistx_cursor (list);
+        job = zlistx_next (list);
+    }
+}
+
 /* Read jobs present in the KVS at startup. */
 int job_state_init_from_kvs (struct list_ctx *ctx)
 {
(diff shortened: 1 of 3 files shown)


src/modules/job-list/job_state.c
```diff
@@ -1146,6 +1146,22 @@ static int depthfirst_map (struct list_ctx *ctx, const char *key,
     return rc;
 }
 
+/* Call zlistx_sort() on the list, then reacquire the list handles since
```
chu11, line 1149: commit message typo "referend" -> "referenced"?

chu11 (approved):
aghh!  My bad!

src/modules/job-list/job_state.c
garlick: I think we all agree that this `zlistx` API choice was a massive foot-gun!

garlick:
Add a test and fixed that commit message typo.

Not sure what people think about running valgrind in this one sharness subtes  - the test completes in around 15s on my desktop, but might be pretty slow in CI.   It does seem to do a pretty good job of reproducing the problem (before the fix).

chu11:
> Not sure what people think about running valgrind in this one sharness subtest - the test completes in around 15s on my desktop, but might be pretty slow in CI. It does seem to do a pretty good job of reproducing the problem (before the fix).

Why not put in `t/issues` as a regression test?  Not doable in there?

garlick:
Maybe that would be better.  It expects all the tests to run under one flux instance but so does the sharness test that I added my test to...

garlick:
Ok, that was cleaner, thanks for that suggestion.

grondo:
Oof, the docker image for flux-core v0.42.0 was never uploaded due to the issue fixed by #4462, so now the dumpfile fetch is failing which is causing all PRs to fail. The fix in #4462 will work around this because it ignores missing dumpfile (I believe). So we'll have to get that PR approved and merged, then re-run CI here. Sorry!

grondo:
Restarting CI after merge of #4462

grondo:
@mergifyio rebase

grondo (approved):
LGTM too! Thanks!

One problem with `t4000-issues-test-driver.t` is that the individual tests are run serially and it is already a really long test. Maybe I'll look into mitigating this somehow after this PR lands.

PR #4473, opened 2022-08-05
docker: add pam development package to images

grondo:
Problem: The PAM development package is not included in the
fluxrm/testenv docker images, but this package will be necessary
for flux-security.

Add the pam development package to all docker images.


grondo:
Docker images resulting from the above changes have already been pushed to `fluxrm/testenv:*`

garlick (approved):
Sounds good!

Issue #3799, opened 2021-07-26
system instance:  need ability to map TBON topology to cluster topology


garlick:
Problem:  a system instance may benefit from configuring "router nodes" (interior tree nodes) to be the service nodes within a scalable unit, but the current configuration which maps a k-ary tree of broker ranks to a flat hostlist does not directly support that.

Design a way to express the desired mapping in configuration.

Allow the TBON levels to have differing numbers of descendants instead of one _k_ for the entire tree.

An example is we might want to have a primary management node as rank 0, the RPS nodes for each scalable unit as the first tree level, other service nodes within the scalable unit as a second tree level, and compute nodes as leaves.

garlick:
This is probably going to be needed sooner rather than later so attaching to next release milestone.

garlick:
Dropping from the next release milestone since this is now tracked in a feature tracker, and we likely won't need it for a few months given current rollout plans which have us on sub-128 node systems through at least September 2022.

Just jotting down one idea about how to represent this in a convenient and compact way:  an optional `parent` key in each hosts array entry.

Example: 256 node cluster consisting of 64 node scalable units and three level TBON with router mapped to first node in each unit, and "test2" (arbitrarily) designated as the management node:
```toml
hosts = [
    { host = "test2" },
    { host = "test[0,64,128,192]", parent = "test2" },
    { host = "test[2-63]",    parent = "test1" },
    { host = "test[65-127]",  parent = "test64" },
    { host = "test[129-191]", parent = "test128" },
    { host = "test[193-255]", parent = "test192" }
]
```
It would be an all or nothing thing - if `parent` is specified at all, the tree would need to be fully specified.  Otherwise topology would be generated as a function of fanout with ranks in breadth first order as they are now.

PR #4474, opened 2022-08-08
refactor broker overlay for topology flexibility

garlick:
Problem: the broker overlay network code assumes a "complete" k-ary tree topology.

This PR prepares the broker to support a custom configured tree topology as discussed in #3799 by removing assumptions about the topology from the overlay code and pushing the details into a new class.   Later, we can add support to that class for setting up trees that are wired any way we like, for example with all interior nodes (overlay message routers) mapped to cluster service nodes that are less likely to be destabilized by running resource intensive user applications.

The most user-visible change here is that the `tbon.fanout` broker attribute has changed:
- it now may be set to a value of 0 on the command line to select a flat topology, regardless of instance size
- it is reset after the topology is instantiated to reflect the number of direct child peers at each broker

In addition, the default topology is now flat for an instance bootstrapped from configuration (normally the system instance).

### DIFF
--- a/doc/man7/flux-broker-attributes.rst
+++ b/doc/man7/flux-broker-attributes.rst
@@ -146,7 +146,9 @@ TREE BASED OVERLAY NETWORK
 ==========================
 
 tbon.fanout [Updates: C]
-   Branching factor of the tree based overlay network.  Default: ``2``.
+   Branching factor of the tree based overlay network.  A value of ``0``
+   means the topology is "flat" (rank 0 is the parent of all other ranks).
+   Default: ``2``.
 
 tbon.descendants
    Number of descendants "below" this node of the tree based
--- a/etc/flux.service.in
+++ b/etc/flux.service.in
@@ -11,7 +11,6 @@ ExecStart=/bin/bash -c '\
   @X_BINDIR@/flux broker \
   --config-path=@X_SYSCONFDIR@/flux/system/conf.d \
   -Scron.directory=@X_SYSCONFDIR@/flux/system/cron.d \
-  -Stbon.fanout=256 \
   -Srundir=@X_RUNSTATEDIR@/flux \
   -Sstatedir=${STATE_DIRECTORY:-/var/lib/flux} \
   -Slocal-uri=local://@X_RUNSTATEDIR@/flux/local \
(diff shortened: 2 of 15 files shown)


garlick:
Pushed fixups to various ci failures...

grondo:
> it is reset after the topology is instantiated to reflect the number of direct child peers at each broker

I don't necessarily have a problem with this change, but I wondered if you considered _adding_ an attribute to track the current number of broker tbon children and keeping `tbon.fanout` as a constant reflection of the chosen configuration?

garlick:
> I don't necessarily have a problem with this change, but I wondered if you considered adding an attribute to track the current number of broker tbon children and keeping tbon.fanout as a constant reflection of the chosen configuration?

Now that you raise the question, the change does seem  unjustified since this is likely to only affect the system instance.  I just didn't know what to set it to when the tree shape is arbitrary.  Maybe we should just leave it unset if the tree is not  a [complete k-ary](https://en.wikipedia.org/wiki/M-ary_tree).

I didn't have a particular use case for an attribute containing the number of direct peers, so I'd vote we add that later if needed (it would be easy).

grondo:
> Now that you raise the question, the change does seem unjustified since this is likely to only affect the system instance. I just didn't know what to set it to when the tree shape is arbitrary. Maybe we should just leave it unset if the tree is not a complete k-ary.

That seems reasonable to me. 

> I didn't have a particular use case for an attribute containing the number of direct peers, so I'd vote we add that later if needed (it would be easy).

That was my main thought. If there isn't a use case, keep `tbon.fanout` set to what was configured. If there is a use case later, then we could add a `tbon.nchildren` or whatever.

garlick:
Ok, pushed that change, as well as a fix for a ci failure in the topology test that I missed in the last round.

garlick:
oops, I realized I need to update the flux-broker-attributes(7) man page and fix a couple commit message errors.  Force pushing again - hope that doesn't mess up any reviews in progress.

grondo (approved):
LGTM. Just spotted one commit message typo

garlick:
Fixed that typo and rebased on current master - i'll set mwp.

PR #4475, opened 2022-08-08
kvs: add root sequence number to checkpoint object

### LINKED ISSUE #4446: kvs: add root sequence number to checkpoint object
Problem:  the KVS root sequence number, mentioned in #4311 as an indicator of how much change has been committed to the kvs,  isn't persistent across restarts (e.g. it's not stored with the root hash in the checkpoint object).   It therefore isn't a great indicator since it resets to zero whenever the instance restarts, whether a GC dump occurs across that reset or not.

Add the root sequence number to the checkpoint object.  It should not be necessary to bump the checkpoint object version.  We can just tolerate a missing sequence number when parsing and have it read as zero.

chu11:
Problem: The KVS root sequence number isn't persisted across
restarts.  It restarts at zero everytime the KVS module is loaded.

Solution: Add the root sequence number to the checkpoint object, read
it in and initialize the KVS primary namespace to that sequence
number.

Fixes #4446

### DIFF
--- a/src/cmd/builtin/restore.c
+++ b/src/cmd/builtin/restore.c
@@ -371,7 +371,8 @@ static int cmd_restore (optparse_t *p, int ac, char *av[])
                      shortblobref_length (blobref),
                      blobref);
         }
-        if (!(f = kvs_checkpoint_commit (h, NULL, blobref, restore_timestamp))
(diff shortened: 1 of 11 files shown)


garlick (approved):
LGTM!

Issue #4476, opened 2022-08-08
libkvs: add cache bypass option to `kvs_checkpoint` functions


chu11:
Per discussion in #4463, it would be nice to add a cache bypass option to the `kvs_checkpoint` functions.  It can be used in dump/restore to go directly to the backing store as well.

PR #4477, opened 2022-08-08
libkvs: Support KVS_CHECKPOINT_FLAG_CACHE_BYPASS flag

chu11:
Problem: The kvs_checkpoint functions checkpoint data to the backing
store through the broker's content-cache.  There is no way to
bypass the cache, similar to libcontent's CONTENT_FLAG_CACHE_BYPASS
flag.

Solution: Support a new KVS_CHECKPOINT_FLAG_CACHE_BYPASS, that will
get/put directly to the backing store.  The flag is very similar to
libcontent's CONTENT_FLAG_CACHE_BYPASS flag.

Fixes https://github.com/flux-framework/flux-core/issues/4476

### DIFF
--- a/src/broker/content-cache.c
+++ b/src/broker/content-cache.c
@@ -617,13 +617,13 @@ static void checkpoint_get_continuation (flux_future_t *f, void *arg)
         goto error;
 
     if (flux_respond (cache->h, msg, s) < 0)
-        flux_log_error (cache->h, "%s: flux_respond", __FUNCTION__);
+        flux_log_error (cache->h, "error responding to checkpoint-get");
     flux_future_destroy (f);
     return;
 
 error:
     if (flux_respond_error (cache->h, msg, errno, NULL) < 0)
-        flux_log_error (cache->h, "flux_respond_error");
+        flux_log_error (cache->h, "error responding to checkpoint-get");
     flux_future_destroy (f);
 }
 
@@ -676,13 +676,13 @@ static void checkpoint_put_continuation (flux_future_t *f, void *arg)
         goto error;
 
     if (flux_respond (cache->h, msg, s) < 0)
-        flux_log_error (cache->h, "%s: flux_respond", __FUNCTION__);
+        flux_log_error (cache->h, "error responding to checkpoint-put");
     flux_future_destroy (f);
     return;
 
 error:
     if (flux_respond_error (cache->h, msg, errno, NULL) < 0)
-        flux_log_error (cache->h, "flux_respond_error");
+        flux_log_error (cache->h, "error responding to checkpoint-put");
     flux_future_destroy (f);
 }
 
--- a/src/cmd/builtin/dump.c
+++ b/src/cmd/builtin/dump.c
@@ -371,7 +371,7 @@ static int cmd_dump (optparse_t *p, int ac, char *av[])
         const char *blobref;
         double timestamp;
 
-        if (!(f = kvs_checkpoint_lookup (h, NULL))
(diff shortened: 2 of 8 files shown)


garlick:
Any reason not to just put a `content_`  prefix in front  of the functions and use the same flag?

chu11:
my reasoning was that these function wrap stuff that is KVS specific

A) the default KVS checkpoint key of ` KVS_DEFAULT_CHECKPOINT`

B) the current (version == 1) KVS checkpoint format of `"{"key":<key>"value":{"version":<version> "rootref":<ref> "sequence":<seq> "timestamp":<timestamp>}"

So we probably do not want to generalize it as a `content_checkpoint` function as it is right now.

I suppose we could generalize it if we wanted, define the KVS_DEFAULT_CHECKPOINT key elsewhere and use it everywhere instead of it being the default?  perhaps instead of "rootref" it should be "blobref"?  Not sure if "sequence" is generic or not.

garlick:
I'd vote for keeping the checkpiont API minimized given that it is not intended to be used outside of the KVS or tools like dump/restore that play wtih KVS internal metadata.  But I'll defer to you on how you'd like that to look.  I was thinking mostly about how to fulfill the current internal use cases with the minimum API cross section (that's where my comments were coming from).

chu11:
Hmmm, my initial thought was to rename to `content_kvs_checkpoint_commit/lookup`,  therefore we can reuse the flags from `libcontent` which does make sense.  But we can make it clear this group of functions is sorta kvs specific.

BUT, then I felt that we should honor all `libcontent` flags.  But honoring `CONTENT_FLAG_UPSTREAM` doesn't seem to make much sense for the checkpointing.

So, I think I'd lean to keeping them named `kvs_` and have the new flag.

src/broker/content-cache.c
```diff
@@ -617,13 +617,13 @@ static void checkpoint_get_continuation (flux_future_t *f, void *arg)
         goto error;
 
     if (flux_respond (cache->h, msg, s) < 0)
-        flux_log_error (cache->h, "%s: flux_respond", __FUNCTION__);
+        flux_log_error (cache->h, "error routing checkpoint-get response");
```
garlick, line 620: Why are some of these message "error routing" and others "error responding"?   There is not really any routing going on in the flux_respond call so IMHO the latter is more appropriate.

chu11:
rebased and fixed up conflicts from #4475

garlick (approved):
LGTM.

chu11:
re-pushed, just fixing up those few log messages

```
diff --git a/src/broker/content-cache.c b/src/broker/content-cache.c
index d5c6390..d2ef184 100644
--- a/src/broker/content-cache.c
+++ b/src/broker/content-cache.c
@@ -617,7 +617,7 @@ static void checkpoint_get_continuation (flux_future_t *f, void *arg)
         goto error;
 
     if (flux_respond (cache->h, msg, s) < 0)
-        flux_log_error (cache->h, "error routing checkpoint-get response");
+        flux_log_error (cache->h, "error responding to checkpoint-get");
     flux_future_destroy (f);
     return;
 
@@ -676,7 +676,7 @@ static void checkpoint_put_continuation (flux_future_t *f, void *arg)
         goto error;
 
     if (flux_respond (cache->h, msg, s) < 0)
-        flux_log_error (cache->h, "error routing checkpoint-put response");
+        flux_log_error (cache->h, "error responding to checkpoint-put");
     flux_future_destroy (f);
     return;
 

```

PR #4480, opened 2022-08-09
testsuite: misc valgrind cleanup

garlick:
This cleans up the test I added for #4465 to respect FLUX_ENABLE_VALGRIND_TEST and load the valgrind suppressions.  The latter is required for the test to pass on aarch64.

I also removed an obsolete environment variable from the mian valgrind test.

### DIFF
--- a/t/issues/t4465-job-list-use-after-free.sh
+++ b/t/issues/t4465-job-list-use-after-free.sh
@@ -4,6 +4,7 @@ VALGRIND=`which valgrind`
 if test -z "${VALGRIND}"; then
     exit 0
 fi
+VALGRIND_SUPPRESSIONS=${SHARNESS_TEST_SRCDIR}/valgrind/valgrind.supp
 
 STATEDIR=issue4470-statedir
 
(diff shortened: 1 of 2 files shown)


grondo (approved):
Maybe we should setup a `VALGRIND` prereq in `sharness.d/flux-sharness.sh` using the checks from `t5000-valgrind.t` (e.g. I think valgrind tests need to be skipped if ASAN is enabled).

Then this test could use the `t4000-issues-test-driver.t` idiom for adding a prereq, 
```sh
# test-prereqs: VALGRIND
```
So that the test is recorded add as a `SKIP` instead of a `PASS`.

However, that's only suggested and probably not necessary at this point, so LGTM!

garlick:
Those changes sound good but if you don't mind I'd like to save that for another time.

Note that the entire issues test is already skipped when ASAN is enabled
```
if test_have_prereq ASAN; then
    skip_all='skipping issues tests under AddressSanitizer'
    test_done
fi
```

grondo:
> Those changes sound good but if you don't mind I'd like to save that for another time.

As noted above, that's fine with me. I can fix the pass vs skipped test when I get around to redoing `t4000-issues-test-driver.t`

PR #4483, opened 2022-08-10
flux-mini: --setattr: place keys in `attributes.system` by default with and default values to 1

grondo:
This PR attempts to simplify the `flux-mini` `--setattr` interface by placing keys in `attributes.system` by default (unless the key starts with `system.`, `user.` or `.`), since almost all use cases for `--setattr` are targeting `attributes.system.` anyway.

Also, similar to `--setopt`, if a value is not provided, have it default to `1`. This allows, for example, `attributes.system.prolog.clear-ssd = 1` to be set simply with `flux mini batch --setattr=prolog.clear-ssd` instead of the more cumbersome `flux mini batch --setattr=system.prolog.clear-ssd=1`.

### DIFF
--- a/doc/man1/flux-mini.rst
+++ b/doc/man1/flux-mini.rst
@@ -365,12 +365,14 @@ OTHER OPTIONS
    otherwise VAL is interpreted as a string. If VAL is not set, then the
    default value is 1. See SHELL OPTIONS below.
 
-**--setattr=KEY=VAL**
+**--setattr=KEY[=VAL]**
    Set jobspec attribute. Keys may include periods to denote hierarchy.
-   VAL may be valid JSON (bare values, objects, or arrays), otherwise VAL
-   is interpreted as a string. If KEY starts with a ``^`` character, then
-   VAL is interpreted as a file, which must be valid JSON, to use as the
-   attribute value.
+   If KEY does not begin with ``system.``, ``user.``, or ``.``, then
+   ``system.`` is assumed.  VAL is optional and may be valid JSON (bare
+   values, objects, or arrays), otherwise VAL is interpreted as a string. If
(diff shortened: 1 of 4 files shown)


garlick (approved):
Nice usability enahncement!  LGTM.

doc/man1/flux-mini.rst
```diff
@@ -365,12 +365,14 @@ OTHER OPTIONS
...
 
-**--setattr=KEY=VAL**
+**--setattr=KEY[=VAL]**
    Set jobspec attribute. Keys may include periods to denote hierarchy.
-   VAL may be valid JSON (bare values, objects, or arrays), otherwise VAL
-   is interpreted as a string. If KEY starts with a ``^`` character, then
-   VAL is interpreted as a file, which must be valid JSON, to use as the
-   attribute value.
+   If KEY does not begin with ``system.``, ``user.``, or ``.``, then
+   KEY is set in ``attributes.system.`` by default.  VAL is optional and
```
chu11, line 371: perhaps is confusing to say "does not begin with `system.` ...` then later say defaults to `attributes.system`.  Perhaps just say `system.` is assumed.

chu11 (approved):
LGTM, just one nit

doc/man1/flux-mini.rst
grondo: You mean something like
```suggestion
   If KEY does not begin with ``system.``, ``user.``, or ``.``, then
   `system.` is assumed.  VAL is optional and
```

Yes, that sounds better.

chu11: yeah, something like that.

grondo: Ok, I made that change and forced a push. Thanks!

grondo:
Thanks! Setting MWP.

Issue #4482, opened 2022-08-09
broker: content-cache flush list can be corrupted


chu11:
While working on #4267 (and leading to #4472) I saw `flux content flush` hang during some regression tests.  Debugging I eventually realized that the `acct_dirty` in the `content-cache` and the length of the flush list were not consistent.  With `acct_dirty` never reaching zero, `flux content flush` would hang never getting a reply.

After trial and error I determined regression test for issue #1760 to be the problem when the content backing module was not loaded.

In the `content-cache`, if a duplicate piece of data arrives, a cache entry can be added to the flush list twice and corrupt the flush list.  This happens b/c `list_add_tail()` does not check if the node is already on the list.

Duplicate data can be sent twice by issuing a `flux kvs dropcache` between KVS writes that are identical, thus the KVS has no cached data and can send "previously sent" data to the content cache again.

While this bug is conceptually possible when there is a loaded content backing module, it would be very racy. (Update: duplicate data would have to be sent after a KVS dropcache AND the entry has not yet been flushed to the backing store in the content-cache).  It is far easier to reproduce when the content backing module isn't loaded.

Here's a reproducer and some debug output I put into the broker to illustrate the badness (some spacing edits for readability):

```
cat <<-EOF >t4500.sh
#!/bin/sh -e

flux kvs put issue4500A.a="abcdefghijk"
flux kvs put issue4500A.b="lmnopqrstuv"
flux kvs put issue4500A.c="wxyz0123456"
flux kvs put issue4500A.d="7890ABCDEFG"
flux kvs put issue4500A.e="HIJKLMNOPQR"
flux kvs put issue4500A.f="STUVWXYZ!!!"
flux kvs put issue4500A.g="<<<<<:>>>>>"

flux kvs dropcache

flux kvs put issue4500B.a="abcdefghijk"
flux kvs put issue4500B.b="lmnopqrstuv"
flux kvs put issue4500B.c="wxyz0123456"
flux kvs put issue4500B.d="7890ABCDEFG"
flux kvs put issue4500B.e="HIJKLMNOPQR"
flux kvs put issue4500B.f="STUVWXYZ!!!"
flux kvs put issue4500B.g="<<<<<:>>>>>"

flux module load content-sqlite

flux content flush

EOF

chmod +x t4500.sh

flux start -s 1 \
    -o,--setattr=broker.rc1_path=${SHARNESS_TEST_DIRECTORY}/rc/rc1-issue4500 \
[...]

PR #4484, opened 2022-08-10
broker: fix content-cache flush list corruption

chu11:
Problem: A dirty cache entry has to potential to be added onto the flush list twice.  This double addition can lead to list corruption.
The observed side effect was a list that was shortened and no longer accurate with respects to the `acct_dirty` counter.  This could lead to hangs with content flush, missed flushes to the backing store, and segfault/memory corruption in the worst case.

Solution: Check if the cache entry is already on the flush list before adding it.

Fixes https://github.com/flux-framework/flux-core/issues/4482

### DIFF
--- a/t/issues/t4482-flush-list-corruption.sh
+++ b/t/issues/t4482-flush-list-corruption.sh
@@ -1,0 +1,40 @@
+#!/bin/sh -e
+
+# How this test works
+#
+# add some unique data to the content cache, we do several stores to
+# build up a decent length internal list of flushable cache entries.
+#
+# write some of the same data again, if error present internal flush
+# list will be messed up and length of flush list < number of dirty
+# entries (acct_dirty).
+#
+# before fix, flux content flush will hang b/c number of dirty entries
+# (acct_dirty) never reaches zero.
+
+cat <<-EOF >t4482.sh
+#!/bin/sh -e
+
+echo "abcde" | flux content store
+echo "fghij" | flux content store
+echo "klmno" | flux content store
+echo "pqrst" | flux content store
+echo "tuvwx" | flux content store
+
+echo "fghij" | flux content store
+echo "klmno" | flux content store
+
+flux module load content-sqlite
+
+flux content flush
+
+flux module remove content-sqlite
+
+EOF
+
+chmod +x t4482.sh
+
+flux start -s 1 \
+    -o,--setattr=broker.rc1_path= \
+    -o,--setattr=broker.rc3_path= \
+    ./t4482.sh
--- a/src/broker/content-cache.c
+++ b/src/broker/content-cache.c
@@ -524,6 +524,18 @@ error:
     cache_resume_flush (cache);
 }
 
+/* Issue #4482, there is a small chance a dirty entry could be added
+ * to the flush list twice which can lead to list corruption.  As an
+ * extra measure, perform a delete from the list first.  If the node
+ * is not on the list, the delete is a no-op.
+ */
+static void flush_list_append (struct content_cache *cache,
+                               struct cache_entry *e)
+{
+    list_del (&e->list);
+    list_add_tail (&cache->flush, &e->list);
+}
+
 static int cache_store (struct content_cache *cache, struct cache_entry *e)
 {
     flux_future_t *f;
@@ -535,7 +547,7 @@ static int cache_store (struct content_cache *cache, struct cache_entry *e)
         return 0;
     if (cache->rank == 0) {
         if (cache->flush_batch_count >= cache->flush_batch_limit) {
-            list_add_tail (&cache->flush, &e->list);
+            flush_list_append (cache, e);
             return 0;
         }
         flags = CONTENT_FLAG_CACHE_BYPASS;
@@ -595,7 +607,7 @@ static void content_store_request (flux_t *h, flux_msg_handler_t *mh,
          * during purge or dropcache, so this does not alter
          * behavior */
         if (cache->rank == 0 && !cache->backing)
-            list_add_tail (&cache->flush, &e->list);
+            flush_list_append (cache, e);
     }
     if (flux_respond_raw (h, msg, hash, hash_size) < 0)
         flux_log_error (h, "content store: flux_respond_raw");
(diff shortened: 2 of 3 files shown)


chu11:
nit for debate: is "corruption" the right word to use here?  Lets say we got a list like:

`a -> b -> c -> d -> e -> NULL`

lets say `c` gets double appended, I think the result is

`a-> b -> c -> NULL`

b/c code would set `e->next = c`, `c.prev = e`, `c.next = NULL`, but b/c `b`'s pointer still points to `c`, we got a shortened list as a result.  We've lost pointers to `d` and `e`.

it's not really "corrupted" in the usual sense of the word.

chu11:
hmmm, one builder failed my regression test b/c of:

```
  expecting success: run_timeout 120 /usr/src/flux-core-0.42.0-46-g410dedf74/t/issues/t4482-flush-list-corruption.sh
  2022-08-10T20:30:35.659903Z broker.err[0]: rc1.0: /bin/bash: /usr/src/flux-core-0.42.0-46-g410dedf74/_build/sub/t/rc/rc1-issue4482: No such file or directory
  2022-08-10T20:30:35.660029Z broker.err[0]: rc1.0: /usr/src/flux-core-0.42.0-46-g410dedf74/_build/sub/t/rc/rc1-issue4482 Exited (rc=127) 0.0s
  2022-08-10T20:30:35.662138Z broker.err[0]: rc3.0: /bin/bash: /usr/src/flux-core-0.42.0-46-g410dedf74/_build/sub/t/rc/rc3-issue4482: No such file or directory
  2022-08-10T20:30:35.662266Z broker.err[0]: rc3.0: /usr/src/flux-core-0.42.0-46-g410dedf74/_build/sub/t/rc/rc3-issue4482 Exited (rc=127) 0.0s
```

note sure why every other builder works.  Lemme try using `FLUX_SOURCE_DIR` instead of `SHARNESS_TEST_DIRECTORY`.

garlick:
Well first, excellent job tracking this down, and it seems like the effect is actually pretty insidious.

Here's a thought.  In at least one other place In content-cache.c, I see we call
```c
        list_del_from (&cache->lru, &e->list);
        list_add (&cache->lru, &e->list);
```
would doing that be sufficient rather than creating a new `list_` function? 

If we need to add a new function, we may want to add it directly to libccan and submit the change upstream.  Or if it's really only useful to us, then possibly add it with a "namespace" other than "list_" so it is evident to casual perusers of our code that it's not part of the original class.  But if we can do something simple to use the class as designed without adding anything then maybe it's better to do that.

chu11:
> would doing that be sufficient rather than creating a new list_ function?

Hmmm, I don't think that will specifically work b/c it requires the entry to be on a list already (there's an assert in the ccan code that checks for this fact). but `list_del()` + `list_add_tail()` should be sufficient.

> If we need to add a new function, we may want to add it directly to libccan and submit the change upstream.

I went down this path just b/c I remember needing the check in the KVS.  But doing something similar to what you suggested should be fine.

chu11:
re-pushed, doing `list_del()` and `list_add_tail()` together instead of the previous solution.  so the PR is now just one commit :-)

t/issues/t4482-flush-list-corruption.sh
```diff
@@ -0,0 +1,52 @@
...
+flux kvs put issue4482A.g="<<<<<:>>>>>"
+
+flux kvs dropcache
+
+flux kvs put issue4482B.a="abcdefghijk"
+flux kvs put issue4482B.b="lmnopqrstuv"
+flux kvs put issue4482B.c="wxyz0123456"
+flux kvs put issue4482B.d="7890ABCDEFG"
+flux kvs put issue4482B.e="HIJKLMNOPQR"
+flux kvs put issue4482B.f="STUVWXYZ!!!"
```
garlick, line 38: I'm not sure this is doing what it looks like at face value, although it still may work.  Those short values will be cached inside the directory entry for "issue4482B" so really what you're doing is creating multiple versions of that directory and the root in the content store.

Suggestion: use `flux content store` since the problem this pokes at has nothing to do with the kvs.

garlick (approved):
LGTM.  Thanks, simple is good.
Just a suggestion for improving the test.

t/issues/t4482-flush-list-corruption.sh
chu11: ahhh yeah, you're right my description is not correct, but I the the effect is identical, the multiple versions of the directory are the "data", not the junk I'm writing.

Let me try with `flux content store`

t/issues/t4482-flush-list-corruption.sh
```diff
@@ -0,0 +1,52 @@
...
+
+flux module load content-sqlite
+
+flux content flush
+
+EOF
+
+chmod +x t4482.sh
+
+flux start -s 1 \
```
grondo, line 49: Since the test script is generated in place, it might be easier (and keep all test components together) to generate the custom `rc1` and `rc3` scripts here as well. I'd also hate to proliferate the one-off rc scripts in `t/rc` and end up with many `rc1-issue*` files in there in the future.

grondo: Or, you could do away with the rc scripts and just load and unload necessary modules directly in the test script as you are doing with `content-sqlite`

grondo (commented):
Not a full review, but I did want to point something out in how the test is organized.

Also, in practice we usually split additional tests into a separate commit unless code changes break tests in keeping with the idea that commits should "do one thing". However, I don't feel strongly about that so this is fine with me.

t/issues/t4482-flush-list-corruption.sh
garlick: My vote would be to make the test self-contained as well.  It should probably drop the bash -e option and handle errors explicitly then, so that it doesn't bail out leaving modules unloaded for the next test.

chu11:
re-pushed, cleaned up the test a ton, it looks far simpler / better now, and split it off into its own commit.

garlick (approved):
LGTM!

Issue #4486, opened 2022-08-11
uids should be treated as unsigned longs


wkharold:
Google's [OS Login guest environment mechanism](https://github.com/GoogleCloudPlatform/guest-oslogin) creates uids in the range > 2^31. Some flux code treats these values as signed values with unfortunate consequences:

```bash
$ id
uid=3420080244(#####) gid=3420080244(#####) groups=3420080244
$ flux mini alloc -N1
0.207s: flux-shell[0]: FATAL: shell_svc_create
0.212s: job.exception type=exec severity=0 shell_svc_create
flux-job: task(s) exited with exit code 1
```

At least the `build_topic` function in src/shell/svc.c needs to be modified: per garlick1@llnl.gov

>  You could change the %d to a %lu there and cast the uid to a (unsigned long) and see how that goes

That does in fact fix the error show above.

grondo:
Thanks for catching this! I started a docker container with a similarly large UID/GID and then ran our `make check` and found a few more failures. We'll get these fixed (and maybe think about how to recreate this in CI).

PR #4489, opened 2022-08-11
fix formatting issues with large UIDs

grondo:
As identified by @wkharold in #4486, uids are treated as signed 32bit integers in several places, including packing into JSON objects like message payloads and eventlogs, as well as when the shell service string is rendered.

I ran `make check` in a docker container as a user with a large UID and this PR is the result of fixing test failures.

I tried to update all places where a userid is packed into a JSON object to pack as `json_int_t` (with a cast) (`json_pack ("I")`), while still unpacking as `i` (since the destination is still 32 bits). @garlick and I could not think of why this wouldn't work.

It would be nice to run one of the CI builds with a VLU (very large userid), but I wasn't quite sure how to get that done quickly and wanted to get this fix in first.


src/modules/job-manager/urgency.c
```diff
@@ -97,7 +97,7 @@ void urgency_handle_request (flux_t *h,
     orig_urgency = job->urgency;
     if (event_job_post_pack (ctx->event, job,
                              "urgency", 0,
-                             "{ s:i s:i }",
+                             "{ s:I s:i }",
                              "userid", cred.userid,
                              "urgency", urgency) < 0)
```
garlick, line 102: missed the typecast on this one

garlick (approved):
LGTM - just one nit

src/modules/job-manager/urgency.c
grondo: Nice catch! Thanks, I've fixed that one and pushed the result.

grondo:
Setting MWP.

PR #4490, opened 2022-08-11
docker: typo in path to Dockerfile

vsoch:
There is a typo in the path to the Dockerfile we missed.


grondo (approved):
Oops, sorry I missed that!
I'll set the merge-when-passing label which will auto-merge this one once all checks pass (even though here the checks are irrelevant)

PR #4492, opened 2022-08-13
etc: support content.backing-module=none

chu11:
Per discussion in #4267, always loading the backing module content-sqlite can lead to performance issues, especially for shorter lived single user instances.   Default to not loading a backing module.

- to enable this, cache checkpoint put/get in the broker even when a backing module does not exist.  This allows a number of checkpoint operations to still work even if a backing module isn't loaded.
- update rc1/rc3 to not load the checkpoint module by default
  - update systemd service file to default configure loading `content-sqlite`
  - update tests that require content-sqlite to specifically configure for it

one possible remaining gotcha is that if the user doesn't set a backing module, they are still allowed to load one and that becomes the backing module.  Dunno if we'd like to support a special backing module config of "none" (or equivalent word) to say "do not allow a backing module to be loaded"?

### DIFF
--- a/src/broker/content-checkpoint.c
+++ b/src/broker/content-checkpoint.c
@@ -15,6 +15,7 @@
 #endif
 #include <inttypes.h>
 #include <assert.h>
+#include <jansson.h>
 #include <flux/core.h>
 
 #include "src/common/libczmqcontainers/czmq_containers.h"
@@ -27,21 +28,95 @@ struct content_checkpoint {
     flux_msg_handler_t **handlers;
     uint32_t rank;
     struct content_cache *cache;
+    zhashx_t *hash;
+    unsigned int hash_dirty;
 };
+
+struct checkpoint_data {
+    struct content_checkpoint *checkpoint;
+    json_t *value;
+    uint8_t dirty:1;
+    bool in_progress;
+    int refcount;
+};
+
+static struct checkpoint_data *
+checkpoint_data_incref (struct checkpoint_data *data)
+{
+    if (data)
+        data->refcount++;
+    return data;
+}
+
+static void checkpoint_data_decref (struct checkpoint_data *data)
+{
+    if (data && --data->refcount == 0) {
+        if (data->dirty)
+            data->checkpoint->hash_dirty--;
+        json_decref (data->value);
+        free (data);
+    }
+}
+
+/* zhashx_destructor_fn */
+static void checkpoint_data_decref_wrapper (void **arg)
+{
+    if (arg) {
+        struct checkpoint_data *data = *arg;
+        checkpoint_data_decref (data);
+    }
+}
+
+static struct checkpoint_data *
+checkpoint_data_create (struct content_checkpoint *checkpoint,
+                        json_t *value)
+{
+    struct checkpoint_data *data = NULL;
+
+    if (!(data = calloc (1, sizeof (*data))))
+        return NULL;
+    data->checkpoint = checkpoint;
+    data->value = json_incref (value);
+    data->refcount = 1;
+    return data;
+}
+
+static int checkpoint_data_update (struct content_checkpoint *checkpoint,
+                                   const char *key,
+                                   json_t *value)
+{
+    struct checkpoint_data *data = NULL;
+
+    if (!(data = checkpoint_data_create (checkpoint, value)))
+        return -1;
+
+    zhashx_update (checkpoint->hash, key, data);
+    data->dirty = 1;
+    checkpoint->hash_dirty++;
+    return 0;
+}
 
 static void checkpoint_get_continuation (flux_future_t *f, void *arg)
 {
     struct content_checkpoint *checkpoint = arg;
     const flux_msg_t *msg = flux_future_aux_get (f, "msg");
-    const char *s;
+    const char *key;
+    json_t *value = NULL;
 
     assert (msg);
 
-    if (flux_rpc_get (f, &s) < 0)
+    if (flux_request_unpack (msg, NULL, "{s:s}", "key", &key) < 0)
         goto error;
 
-    if (flux_respond (checkpoint->h, msg, s) < 0)
+    if (flux_rpc_get_unpack (f, "{s:o}", "value", &value) < 0)
+        goto error;
+
+    if (checkpoint_data_update (checkpoint, key, value) < 0)
+        goto error;
+
+    if (flux_respond_pack (checkpoint->h, msg, "{s:O}", "value", value) < 0)
         flux_log_error (checkpoint->h, "error responding to checkpoint-get");
+
     flux_future_destroy (f);
     return;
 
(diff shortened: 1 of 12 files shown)


garlick:
In #4267 
> I'm not sure about making this is the default, since there is unexplained job throughput degradation in this mode. However it is not currently possible to even select this mode. Let's allow this issue to be closed once setting `content.backing-module=none` works as you'd expect.

Could we cut this PR there and save the discussion about making it the default for another time?

chu11:
> Could we cut this PR there and save the discussion about making it the default for another time?

ahhh, I missed that.  I'll tweak the PR.  Should be simpler as a result as many testsuite updates don't need to happen now.

garlick:
One other thought is would it make sense to split the broker checkpoint stuff off to another source file?  content-cache.c is already long and complex and this seems to be somewhat disjoint.

chu11:
re-pushed

- split out content checkpoint code into another file
  - the "context' that the `content-checkpoint` API creates is actually stored within `struct content_cache` and not `struct broker`.  I 
    didn't initially intend for this, but `content-cache` needs to call `content-checkpoint` at one point, so I felt it better to do it this way.
- rc scripts do not load `none` by default now

garlick (changes requested):
Thanks!

I'm noting that while I can shut down a "none" instance with `flux shutdown --dump=foo.tgz`, I can't restart from that file:
```
$ lux start -o,-Scontent.backing-module=none,-Scontent.restore=foo.tgz
2022-08-21T14:05:21.398977Z broker.err[0]: rc1.0: flux-restore: error flushing content cache: Function not implemented
```
It might be a good idea to pause and question whether the "flush" operation should fail when there is no backing store.  Maybe we should redefine it as "flush to backing store, if any"?  I know that would require other content tests to be updated.   I'm not sure if that's the right answer or not, but if it is, it could also enable the "startlog" stuff to remain in the rc files, and an instance that is restarted several times using dump/restore could have a startlog like one that restarts with a backing store.

Another thing that's probably worth testing is that you can reload the kvs module in  "none" instance, and its content remains valid.

chu11:
> Edit: just realized this checkpoint service is created on all ranks, but I think it should only be created on rank 0? If for some reason it is accessed from other ranks, it should just let the broker forward the requests to rank 0 (which is what happens if the message handlers are only registered on rank 0).

Oh good catch, although I think we want to load the checkpoint service all of the time.  We just don't want the checkpoint caching service used on rank != 0.  This was actually a bug back in #4463, ENOSYS should only be returned when the backing module isn't loaded on rank == 0.

chu11:
re-pushed, addressing all of the comments above except the discussion about if `content.flush` should be an error or not.  Of particular note:

- created a bunch of new functions to deal with the cleanup paths better.  Those mem-leaks / double frees were embarrasing :P
- add some more tests per comments above (reload kvs, forwarding from rank != 0 works as intended)

chu11:
> It might be a good idea to pause and question whether the "flush" operation should fail when there is no backing store. Maybe we should redefine it as "flush to backing store, if any"? I know that would require other content tests to be updated. I'm not sure if that's the right answer or not, but if it is, it could also enable the "startlog" stuff to remain in the rc files, and an instance that is restarted several times using dump/restore could have a startlog like one that restarts with a backing store.

Hmmmm, my initial feeling is that when someone does `content.flush` there should be an error returned.  Presumably you'd want to know that things weren't backed up properly?

Here's a thought, could `content.flush` take an argument that is something like `--quiet`?   i.e. don't return an error if the backing isn't there?

Alternately, we could simply add some options to startlog, restore, etc. to not flush when the option is set, and we could set the option when backing module == `none`.

chu11:
argh, re-pushed, fixed up a mem-leak and fixed some bash-sism in my tests that were affecting the CI

chu11:
and re-pushed, fixing a s3 CI issue

src/broker/content-checkpoint.c
```diff
@@ -51,45 +126,72 @@ static void checkpoint_get_continuation (flux_future_t *f, void *arg)
...
         goto error;
-    }
 
-    if (flux_request_decode (msg, NULL, &s) < 0)
+    msgref = flux_msg_incref (msg);
+    if (flux_future_aux_set (f,
+                             "msg",
+                             (void *)msgref,
+                             (flux_free_f)flux_msg_decref) < 0) {
+        flux_msg_decref (msgref);
```
garlick, line 150: The `msgref` temporary variable isn't really accomplishing anything at this point.   Might as well just use 
```c
    if (flux_future_aux_set (f,
                             "msg",
                             (void *)flux_msg_incref (msg),
                             (flux_free_f)flux_msg_decref) < 0) {
        flux_msg_decref (msg);
```
Same comment applies to `put`

src/broker/content-checkpoint.c
```diff
@@ -51,45 +126,72 @@ static void checkpoint_get_continuation (flux_future_t *f, void *arg)
...
+        if (flux_respond_pack (h, msg,
+                               "{s:O}",
+                               "value", data->value) < 0)
+            flux_log_error (h, "error responding to checkpoint-get");
+        return;
     }
 
+    if (content_checkpoint_get_backing (checkpoint, msg, key, &errstr) < 0)
+        goto error;
+
```
garlick, line 189: For rank 0, this is forcing the RPC to use the backing store, so it would get ENOSYS in the "none" case.  The request should go to the broker RPC.  Same comment applies to `put`.

It looks like `t0028-content-backing-none.t` includes a test that expects this behavior.  Shouldn't it work?

garlick (commented):
Much improved though I still had a couple of comments/questions

src/broker/content-checkpoint.c
garlick: I hate to keep dragging this out.  Do we have a use case for _not_ returning ENOSYS to the checkpoint operations on rank > 0?    This is primarily used internally by the kvs on rank 0 only, and by the dump/restore and startlog tools, all of which one would expect  to run on rank0 I think.  Up to you but we could just prune that case and call it good if it cuts down the code.  I maybe shouldn't have brought it up, but I was thinking that we could just load the service on rank 0 and let requests be forwarded "naturally", forgetting that the service name is now shared with the content load/store ops that must work on all ranks.

chu11: My initial thought upon reading your comments was that I screwed up.  We should have `content.checkpoint-{get,put}` forward to rank 0's `content.checkpoint-{get,put}` on non-rank 0 brokers.  But your comment makes sense that maybe we don't have to, we should just `ENOSYS` right off the bat for ranks != 0.

For myself, I tend to lean towards "consistency", b/c I just dislike seeing something different in the code.

Let me have rank > 0 forward appropriately to rank 0.  I'll spin that off into another PR since it's sort of independent of this and really a mistake from #4463

chu11:
re-pushed, building on top of #4519, adjust some tests as a result of #4519

chu11:
rebased and re-pushed now that #4519 is merged, this PR is a lot smaller now :-)

garlick (approved):
LGTM, thanks for sticking it out through all the review comments :-)

chu11:
re-started one builder that had a ton of failures not related to this PR.  assumption is workflow/container borked and affected bunch of tests.

chu11:
rebased & re-pushed, mergifyio seemed to get stuck on something

PR #4493, opened 2022-08-15
testsuite: misc cleanup and activate valgrind suppression needed on aarch64

garlick:
Problem: a recently added issue test runs valgrind unconditionally, and the test fails on aarch64.

Honor the FLUX_ENABLE_VALGRIND_TEST in that issue test and activate the valgrind suppressions there, since the test was failing due to a false positive in libev that we already have a suppression for.

Then a tiny bit of random cleanup noticed along the way.


garlick:
Argh, this PR was already merged.  My branch was just out of date and I had forgotten we did it last week.  Sorry for the noise!

PR #4499, opened 2022-08-16
flux-job: point users to flux-jobs(1)

### LINKED ISSUE #4497: flux job list should probably not be installed
Problem: it's easy to confuse `flux job` with `flux jobs`.

Perhaps `flux job list`, which outputs JSON, should  just be a test program that is not installed (or a hidden command).

garlick:
Problem: flux job list is too easy to confuse with flux-jobs(1),
but the former is a test tool that emits JSON.

Omit the list, list-inactive, and list-ids subcommands from help output.
If output is a tty, assume a human is trying to ingest JSON (blech!)
and throw a fatal error that instructs them to try flux-jobs(1).

Fixes #4497

### DIFF
--- a/src/cmd/flux-job.c
+++ b/src/cmd/flux-job.c
@@ -360,21 +360,21 @@ static struct optparse_subcommand subcommands[] = {
       "[OPTIONS]",
       "List jobs",
       cmd_list,
-      0,
+      OPTPARSE_SUBCMD_HIDDEN,
       list_opts
     },
     { "list-inactive",
       "[OPTIONS]",
       "List Inactive jobs",
       cmd_list_inactive,
-      0,
+      OPTPARSE_SUBCMD_HIDDEN,
       list_inactive_opts
     },
     { "list-ids",
       "[OPTIONS] ID [ID ...]",
       "List job(s) by id",
       cmd_list_ids,
-      0,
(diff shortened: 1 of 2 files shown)


chu11 (approved):
LGTM

chu11:
I think this is fine for the short term, but for the long term should these subcommands be split out into a test tool, like `flux jobtest` or something?  While the average person wouldn't want to output the JSON to the command line, I have done this on occasion.

Also, would this effect debugging efforts in sharness when one uses the `-v` and/or `-d` options?  Maybe?

garlick:
Meh, I think this is probably enough on this.  If we have issues down the road we can deal with it then?

chu11:
> Meh, I think this is probably enough on this. If we have issues down the road we can deal with it then?

Sure, can deal with when it becomes an issue.

grondo:
> Edit: yeah, I guess it would if you did some dumb debugging like this

I ran all the current tests that contain `flux job list` with `-d -v` and didn't find any that failed currently, so I think it is probably fine for now.

grondo (approved):
LGTM!

garlick:
Oops, broken && chain!  I force pushed a fix and I'll set MWP.  Thanks!

PR #4500, opened 2022-08-16
doc: add TIMEOUT result to flux-jobs(1)

### LINKED ISSUE #4498: flux-jobs: document job status / result can report TIMEOUT
it would be useful if job resutls also showed jobs that timed out.

TBD: do we need a new color for timeouts in the job output display?  Could be the same color as CANCELED, as they are both in that group of jobs that are "indeterminate" to some extent.

chu11:
Problem: flux-jobs was missing information about TIMEOUT being
a valid result / status.

Solution: Update documentation accordingly.

Fixes #4498


garlick (approved):
This pertains to the man page not the command, so maybe use flux-jobs(1) or "flux-jobs manual page" in PR title, commit title, and commit description?

chu11:
@garlick oh good point, updated all to `flux-jobs(1)`

PR #4501, opened 2022-08-16
testsuite: add timeout job tests

chu11:
So I'm admittedly not sure if the pros outweigh the cons on this.  This was initially going to be just comments to point readers to the tests in `t2900-job-limits`, but as I was skimming things I realized "oh this isn't covered ... and this isn't covered ... and ..." and I slowly started adding things.

The addition of the timed out job to the job list tests is probably racy, which we can add things to try and limit that.

I maybe added too many permutations to the job list filtering tests, which can be whittled down.

Do the pros outweigh the cons?

### DIFF
--- a/t/t2800-jobs-cmd.t
+++ b/t/t2800-jobs-cmd.t
@@ -30,7 +30,7 @@ runpty="${SHARNESS_TEST_SRCDIR}/scripts/runpty.py --line-buffer -f asciicast"
 
 # Return the expected jobids list in a given state:
 #   "all", "run", "sched", "active", "inactive",
-#   "completed", "canceled", "failed"
+#   "completed", "canceled", "failed", "timeout'
 #
 state_ids() {
 	for f in "$@"; do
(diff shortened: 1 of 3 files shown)


grondo:
The added tests seem pretty reliable actually, and it is good that we add tests for querying timeout jobs. I'd say since you've done the work here already, let's go ahead and merge this. Thanks!

t/t2800-jobs-cmd.t
```diff
@@ -102,6 +102,14 @@ test_expect_success HAVE_JQ 'submit jobs for job list testing' '
 	echo $jobid >> inactiveids &&
 	echo $jobid > failed.ids &&
 	#
+	#  Run a job that will timeout, copy its JOBID to both inactive and
+	#   timeout lists.
+	#
+	jobid=`flux mini submit --time-limit=0.5s sleep 30` &&
+	echo $jobid >> inactiveids &&
+	echo $jobid > timeout.ids &&
+	fj_wait_event `tail -n 1 inactiveids` clean &&
```
grondo, line 111: Not a big deal, but why get the jobid using `tail` when it is already in `$jobid`? (Assuming I missed something obvious, or this was just cut-and-paste from other tests)

grondo (approved):
LGTM, just one question inline.

grondo:
Suggestion: s/timed out/timeout/ in PR title since that more closely matches the literal status string being tested.

t/t2800-jobs-cmd.t
chu11: oh haha, it's cut and paste from the other tests.

chu11:
re-pushed fixing the nit that @grondo noted above.  removed WIP, will set MWP

PR #4503, opened 2022-08-17
flux-jobs: improve bad username error message

chu11:
Problem: When an invalid username is provided to flux-jobs,
the error message is:

flux-jobs: ERROR: invalid literal for int() with base 10: 'fasdfsdafd'

which is not helpful.

Solution: Improve error message to:

flux-jobs: ERROR: Invalid user fasdfsdafd specified

Update test to check for correct error output message.


grondo (approved):
Nice fix!

garlick:
Hope this is OK but I'm just going to set MWP on this one since Al is on vacation right now.  Nice fix.

PR #4505, opened 2022-08-18
update fluid check to check explicitly for utf-8

### LINKED ISSUE #4504: ƒ may be misinterpreted
Found by Peter Thompson, ƒ is interpreted as Esc-_ by some terminals, and possibly also by at least one VNC client.  Might be worth considering if we wan to use that quite so ubiquitously.

trws:
Some multibyte locales do not support the florin symbol, and we were
checking for width rather than encoding.  This check looks explicitly
for UTF-8 support in the current locale.

For normal examples this is the same as before, but for "interesting" character sets like some of the Japanese or Korean varieties avoids printing something outside the base ascii range.

fixes #4504


grondo:
This seems like it should work, but it doesn't pass the "simple test" for me for some reason
(Note, I added debugging here to print the result of `nl_langinfo()`)

```console
$ flux jobs -ano {id.f58}nl_langinfo = UTF-8
ƒ3yFsZ6o
$ LANG=C flux jobs -ano {id.f58}
nl_langinfo = UTF-8
ƒ3yFsZ6o
$ LANG=C locale charmap
ANSI_X3.4-1968
$ FLUX_F58_FORCE_ASCII=1 flux jobs -ano {id.f58}
nl_langinfo = UTF-8
f3yFsZ6o
```

trws:
Ok that's interesting, I'm seeing the same thing, but if I set LC_ALL then that same check works.  Would you mind checking if you see the same @grondo? I'm trying to figure out why LC_all would impact it but not LANG, for this they both should I think.

trws:
Well, learning something new today.  Because python3 uses unicode, utf-8, heavily it interprets a `LANG` of C or POSIX as C.UTF-8 rather than C, and only overrides that if LC_ALL is used:

```console
~/Projects/flux/flux-core/build utf8-check *9 !2 ?30                 scogland1@chimera 
❯ LANG=C /usr/bin/python3 -c 'import locale ; print(locale.setlocale(locale.LC_ALL))'
LC_CTYPE=C.UTF-8;LC_NUMERIC=C;LC_TIME=C;LC_COLLATE=C;LC_MONETARY=C;LC_MESSAGES=C;LC_PAPER
=C;LC_NAME=C;LC_ADDRESS=C;LC_TELEPHONE=C;LC_MEASUREMENT=C;LC_IDENTIFICATION=C

~/Projects/flux/flux-core/build utf8-check *9 !2 ?30                 scogland1@chimera 
❯ LANG=POSIX /usr/bin/python3 -c 'import locale ; print(locale.setlocale(locale.LC_ALL))'

LC_CTYPE=C.UTF-8;LC_NUMERIC=C;LC_TIME=C;LC_COLLATE=C;LC_MONETARY=C;LC_MESSAGES=C;LC_PAPER
=C;LC_NAME=C;LC_ADDRESS=C;LC_TELEPHONE=C;LC_MEASUREMENT=C;LC_IDENTIFICATION=C

~/Projects/flux/flux-core/build utf8-check *9 !2 ?30                 scogland1@chimera 
❯ LC_ALL=C /usr/bin/python3 -c 'import locale ; print(locale.setlocale(locale.LC_ALL))'
C
```

Because `flux jobs` is a python command, it's inheriting that behavior, and now that instead of checking for multibyte we're checking for "do we have utf8" it says we do.  To be conservative, I'm thinking we add the old check back in, in addition to the nl_langinfo check.  Good with you @grondo?

grondo:
> To be conservative, I'm thinking we add the old check back in, in addition to the nl_langinfo check. Good with you @grondo?

Wow, thanks for the locale lessons today @trws!

I was going to suggest the same thing, as it seems we need to be very conservative here. So, sounds good to me!

Thanks!

grondo (approved):
LGTM! Want to just squash the two commits together (preserving interesting recap in the commit messages of course)?

trws:
Squashed.

grondo:
Excellent! Ok to set MWP?

trws:
Go for it. 👍

Issue #4506, opened 2022-08-18
add flux jobtap query subcommand


garlick:
Problem: it's hard to debug jobtap plugins.

In flux-framework/flux-accounting#264, an RPC for dumping internal state of the accounting `mf_priority` jobtap plugin is proposed, along with a command.

It might be wise to add a standardized way to extract a json object representing internal state from any plugin that implements a well known service or even a jobtap callback, and add a subcommand to `flux jobtap` that dumps the result to stdout.

Another related idea is to provide a way to query a version string, for cases like accounting where the plugin has a version independent of flux-core.

PR #4507, opened 2022-08-19
add flux jobtap query subcommand

grondo:
As described in #4506, this PR adds support for a new `flux jobtap query` subcommand which can fetch extended data about jobtap plugins (currently `name` and `path` by default). Optionally, plugins can register for a `plugin.query` callback topic and add arbitrary data to the response payload. This gives plugins a generic way to export internal state for debugging or other purposes.

As an example implementation, the `dependency-after` plugin is extended to support this interface and adds a `dependencies` key the query response which includes all active dependencies, e.g.:

```
$ flux jobtap query .dependency-after | jq
{
  "name": ".dependency-after",
  "path": "builtin",
  "dependencies": [
    {
      "id": 500699234304,
      "depid": 505581404160,
      "type": "after-success",
      "description": "after-success=fE9r5TG3"
    },
    {
      "id": 519590379520,
      "depid": 524053118976,
      "type": "after-success",
      "description": "after-success=fEedRWaK"
    },
    {
      "id": 533515468800,
      "depid": 537894322176,
      "type": "after-success",
      "description": "after-success=fF1qwEUX"
    }
  ]
}
```

If a plugin was loaded from a file, the path reflects the actual path of the original DSO:

```
$ flux jobtap query alloc-bypass.so | jq
{
  "name": "alloc-bypass.so",
  "path": "/home/grondo/git/flux-core.git/src/modules/job-manager/plugins/.libs/alloc-bypass.so"
}
```

### DIFF
--- a/doc/man7/flux-jobtap-plugins.rst
+++ b/doc/man7/flux-jobtap-plugins.rst
@@ -237,6 +237,21 @@ The callback should return 0 on success, and -1 on failure.  On failure,
 it may optionally set a human readable error string in the ``errstr`` output
 argument.  The ``flux_jobtap_error()`` convenience function may be useful here.
 
+
+PLUGIN CALLBACK TOPICS
+======================
+
+plugin.query
+  The job manager calls the ``plugin.query`` callback topic to give a plugin
+  the opportunity to provide extra data in response to a ``jobtap-query``
+  request (as used by the ``flux jobtap query PLUGIN`` command). This can be
+  used by a plugin to export internal plugin state for inspection by an admin
+  or user by placing the data in the output arguments of the callback, e.g.::
+
+    flux_plugin_arg_pack (p, FLUX_PLUGIN_ARG_OUT,
+                          "{s:O}"
+                          "data", internal_data);
+
 .. _priority:
 
 PRIORITY
--- a/src/modules/job-manager/jobtap.c
+++ b/src/modules/job-manager/jobtap.c
@@ -1425,6 +1425,90 @@ void jobtap_handler (flux_t *h,
         flux_log_error (h, "jobtap_handler: flux_respond");
 }
 
+static int jobtap_query_plugin (flux_plugin_t *p,
+                                char **json_str,
+                                flux_error_t *errp)
+{
+    int rc = -1;
+    flux_plugin_arg_t *args;
+    const char *path = flux_plugin_get_path (p);
+    const char *name = jobtap_plugin_name (p);
+
+    if (path == NULL)
+        path = "builtin";
+
+    if (!(args = flux_plugin_arg_create ()))
+        return errprintf (errp,
+                          "flux_plugin_arg_create: %s",
+                          strerror (errno));
+
+    if (flux_plugin_arg_pack (args,
+                              FLUX_PLUGIN_ARG_OUT,
+                             "{s:s s:s}",
+                             "name", name,
+                             "path", path) < 0) {
+        errprintf (errp, "%s", flux_plugin_arg_strerror (args));
+        goto out;
+    }
+
+    if (flux_plugin_call (p, "plugin.query", args) < 0) {
+        errprintf (errp, "plugin.query failed");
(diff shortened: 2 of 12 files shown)


cmoussa1 (approved):
This LGTM! Thanks @grondo for working on this. I think this will be a useful utility especially for the multi-factor priority plugin. I also think it should be somewhat straightforward for me to adjust flux-framework/flux-accounting#264 as a result of this PR, which I will do once this lands.

Thanks!

grondo:
@garlick noticed some `errno` inconsistencies in this PR and I've just forced a push to address those (largely by not using `errno` anymore)
 - The query RPC no longer sets the `errno` parameter to `flux_respond_error(3)` since I think in all cases a human-readable error is used
 -  `flux_plugin_get_path(3)` no longer sets EINVAL when its parameter is `NULL`

I also fixed a few tests that failed to supply the `jq` `-e` option, and added a couple tests for good measure.

src/modules/job-manager/jobtap.c
```diff
@@ -1425,6 +1425,93 @@ void jobtap_handler (flux_t *h,
...
+    if (!found) {
+        errprintf (&error, "%s: plugin not found", name);
+        goto error;
+    }
+
+    rc = flux_respond (h, msg, result);
+    ERRNO_SAFE_WRAP (free, result);
+    if (rc < 0)
+        goto error;
+    return;
```
garlick, line 1509: If `flux_respond()` fails, best to just log the error and return since the error response will likely fail too.

src/modules/job-manager/plugins/dependency-after.c
```diff
@@ -617,6 +623,63 @@ static int inactive_cb (flux_plugin_t *p,
...
+    }
+
+    if (flux_plugin_arg_pack (args,
+                              FLUX_PLUGIN_ARG_OUT,
+                              "{s:o}",
+                              "dependencies",
+                              o) < 0)
+        flux_log_error (flux_jobtap_get_flux (p),
+                        "dependency-after: query_cb: flux_plugin_arg_pack: %s",
+                        flux_plugin_arg_strerror (args));
```
garlick, line 679: Leaks `o` on `flux_plugin_arg_pack()` failure.

doc/man7/flux-jobtap-plugins.rst
```diff
@@ -237,6 +237,21 @@ The callback should return 0 on success, and -1 on failure.  On failure,
...
+plugin.query
+  The job manager calls the ``plugin.query`` callback topic to give a plugin
+  the opportunity to provide extra data in response to a ``jobtap-query``
+  request (as used by the ``flux jobtap query PLUGIN`` command). This can be
+  used by a plugin to export internal plugin state for inspection by an admin
+  or user by placing the data in the output arguments of the callback, e.g.::
+
+    flux_plugin_arg_pack (p, FLUX_PLUGIN_ARG_OUT,
+                          "{s:o}"
+                          "data", internal_data);
```
garlick, line 253: Since the args are destroyed after the callback returns, maybe the example should use `{s:O}` (big O)?

I'm not sure if that makes it any clearer.  This is a fertile ground for bugs though.

garlick (approved):
Nicely done!  Just a couple of nits.

src/modules/job-manager/jobtap.c
grondo: Duh, good point.

doc/man7/flux-jobtap-plugins.rst
grondo: Ah, very good point. Given my mistake above, it is probably always better to pack with "O" and decref manually. (I don't know how many times I have to relearn this until it gets through my head)

grondo:
I addressed @garlick's comments above and force-pushed the result.

garlick:
Thanks - feel free to set MWP!

grondo:
Thanks!

PR #4509, opened 2022-08-22
testsuite: unset FLUX_F58_FORCE_ASCII during testsuite

grondo:
Problem: If FLUX_F58_FORCE_ASCII is set in the environment before
running `make check`, a few tests will fail.

Add FLUX_F58_FORCE_ASCII to the list of environment variables to
unset before the testsuite.


garlick (approved):
LGTM!

grondo:
Thanks!

PR #4512, opened 2022-08-22
testsuite: use flux jobs in valgrind workload

chu11:
Problem: In commit

ad5cd8bf1d2c2a3e4226d7318683bc08fdae505b

the "flux job list" command will give a warning / error to the user
to use "flux jobs" when the output is a tty.  The valgrind workload
still uses "flux job list", which can lead to errors when running the
valgrind unit tests by hand on the command line.

Solution: Use "flux jobs" instead of "flux job list" in the valgrind
workload.


grondo (approved):
Nice catch!

PR #4513, opened 2022-08-23
job-list: remove job-list.list-inactive RPC

grondo:
This PR moves some of the special functionality of `job-list.list-inactive` into the main `job-list.list` RPC, updates existing users to use `job-list.list` instead, then removes the now duplicated `job-list.list-inactive` RPC and its associated code from `job-list`.

This is a precursor to adding `--since` functionality to `flux jobs`, which will come in a follow-on PR.


garlick (approved):
LGTM.  Just one small commit message typo.

chu11:
LGTM, although I was expecting us to remove the `list_inactive()` functions as well.  Do we have enough callers to make it worthwhile to keep?  vs. forcing all callers to update to the primary list function instead?

grondo:
>  Do we have enough callers to make it worthwhile to keep? vs. forcing all callers to update to the primary list function instead?

I doubt we have any callers of the C API `flux_job_list()` or `flux_job_list_inactive()` outside of the utilities that now use them. However, we don't have a method for deprecating functions so I left them alone for now.

As you had pointed out, the flux-sched `flux-tree-helper.py` utility is an external user of `flux.job.job_list_inactive()`, so I left that in too to avoid breaking flux-sched.

grondo:
Pushed a fix for the commit message typo. Will set MWP.

PR #4515, opened 2022-08-23
cmd/flux-jobs: include job state in status output

### LINKED ISSUE #4495: flux jobs does not display CLEANUP state by default
Problem: if a job is stuck completing for whatever reason (such as the painfully slow epilog on LC systems, but another one would be a hung node), `flux jobs` shows the job as `R` (running).  This abbreviated state is obscuring what is really going on.

Just now I tried to run a program with the wrong path on corona and got the appropriate error from `flux mini run`.   It's now about 10 minutes later and it's still showing as `R`.  If it showed as completing, it would be more obvious what is happening.

Most of the job states exist because they convey information about the job that could be useful to users.  We may want to re-evaluate the decision to only show the "high level" states by default.

chu11:
Problem: Currently the job status outputs "PENDING"
when the job is in the DEPEND or SCHED state and "RUNNING" when the
job is in the RUN or CLEANUP states.  This abstraction of job states
in the job status doesn't give enough detail about what is going
on with a job.

Solution: Update the job status to output the job state when a job
is not inactive.  Output the job result when it is inactive.  In
other words, the list of outputs for the job status is now:

DEPEND, SCHED, RUN, CLEANUP, COMPLETED, FAILED, CANCELED, or TIMEOUT

Update tests for new output accordingly.

Update documentation accordingly.

Fixes #4495

### DIFF
--- a/doc/man1/flux-jobs.rst
+++ b/doc/man1/flux-jobs.rst
@@ -130,9 +130,9 @@ CANCELED, TIMEOUT. Under the *result_abbrev* field name, these are
 abbreviated as CD, F, CA, and TO respectively.
 
 The job status is a user friendly mix of both, a job is always in one
-of the following five statuses: PENDING, RUNNING, COMPLETED, FAILED,
-CANCELED, or TIMEOUT. Under the *status_abbrev* field name, these are
-abbreviated as P, R, CD, F, CA, and TO respectively.
+of the following statuses: DEPEND, SCHED, RUN, CLEANUP, COMPLETED,
+FAILED, CANCELED, or TIMEOUT. Under the *status_abbrev* field name,
+these are abbreviated as D, S, R, C, CD, F, CA, and TO respectively.
 
 
 OUTPUT FORMAT
@@ -227,7 +227,8 @@ The field names that can be specified are:
    list of any currently outstanding job dependencies
 
 **status**
-   job status (PENDING, RUNNING, COMPLETED, FAILED, CANCELED, or TIMEOUT)
+   job status (DEPEND, SCHED, RUN, CLEANUP, COMPLETED, FAILED,
+   CANCELED, or TIMEOUT)
 
 **status_abbrev**
    status but in a max 2 character abbreviation
--- a/src/bindings/python/flux/job/info.py
+++ b/src/bindings/python/flux/job/info.py
@@ -36,11 +36,13 @@ def resulttostr(resultid, fmt="L"):
     return raw.flux_job_resulttostr(resultid, fmt).decode("utf-8")
 
 
+# Status is the job state when pending/running (i.e. not inactive)
+# status is the result when inactive
 def statustostr(stateid, resultid, fmt="L"):
-    if stateid & flux.constants.FLUX_JOB_STATE_PENDING:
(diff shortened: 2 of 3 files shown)


grondo:
Well that was a lot simpler than I anticipated! :-)

I'm a little worried that changing from "PD" to "S" (for the SCHED state) is a little more confusing that it needs to be. It feels like a bit of a bad idea, but it would make the change here less drastic -- and give users what they are used to -- if we kept `PD` as the abbreviation for the SCHED state.

Just bringing it up as something to discuss. Any thoughts?

chu11:
> I'm a little worried that changing from "PD" to "S" (for the SCHED state) is a little more confusing that it needs to be. It feels like a bit of a bad idea, but it would make the change here less drastic -- and give users what they are used to -- if we kept PD as the abbreviation for the SCHED state.

I see your point, although it just feels wrong to keep "PD" :-)

An idea, which I'm not sure if I ever created an issue for it.

1) create a way to specify default `flux-jobs` output, perhaps via an environment variable or something.  Sites can define what should be their default.

2) Create a `status_abbrev_trad` for "traditional" abbreviations?

chu11:
re-pushed, fixing up python black issue

garlick:
> Just bringing it up as something to discuss. Any thoughts?

Hmm you're probably right, both because of the slurm state and because of what our tools look like now.

A downside is that we define the virtual "pending" state as "The job is in DEPEND, PRIORITY, or SCHED states." in [RFC 21](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_21.html) and we let people filter `flux jobs` output based on those virtual states.  :shrug:

chu11:
> A downside is that we define the virtual "pending" state as "The job is in DEPEND, PRIORITY, or SCHED states." in [RFC 21](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_21.html) and we let people filter flux jobs output based on those virtual states. shrug

I hate to even suggest it, but we could make a new virtual state, like uhhh "PRE_PROCESS" that means NEW, DEPEND, PRIORITY, and "PENDING" is an alias for "SCHED"??

grondo:
> A downside is that we define the virtual "pending" state as "The job is in DEPEND, PRIORITY, or SCHED states." in RFC 21 and we let people filter flux jobs output based on those virtual states.

Yeah, that is very good point, though we aren't reporting a virtual "state" in the `flux jobs` output per se, but a more granular "status", so it may be ok.

Eh, "`S`" may end up working fine actually, it is kind of mnemonically not only "SCHED" but also like `ps` process state code `S`, the job is "Sleeping waiting for resources". (Though we shouldn't push that resemblance too far).

Hm, the `ps` `STAT` column gives me another idea that we could eventually add modifiers to the "state" abbreviation eventually if we want or need to do so. Just something to keep in mind. The derived "status" field is pretty flexible in this way...

garlick:
I like `S` then for the default output.  We can pursue other ideas for output customization or a friendlier UX separately perhaps?

grondo:
> We can pursue other ideas for output customization or a friendlier UX separately perhaps?

Yeah, that's fine with me!

grondo (approved):
LGTM! Thanks!

PR #4517, opened 2022-08-23
flux-jobs: add --since=WHEN and --name=NAME options

grondo:
Now that `job-list.list` supports the `since` and `name` parameters, this PR adds support to `flux-jobs` for these options.

The `--name` option restricts returned jobs to those matching a supplied name, and the `--since=WHEN` option restricts jobs to jobs that became inactive after WHEN. If no other filtering options are supplied, then `--since` implies `-a` (because obviously pending and running jobs will become inactive after any supplied time).

The argument to `--since` supports raw timestamp, (negative) offset in FSD, or anything handled by the `flux.util.parse_datetime()` function (e.g. `yesterday`, `an hour ago`, `last monday`).

Using `flux jobs --since` we can now easily determine which jobs were pending or running since yesterday
```console
$ flux jobs --since=8am,yesterday
```
jobs that failed in the last 24hrs
```console
$ flux jobs -f failed --since=-24h
```
or jobs that completed (regardless of status) the over the weekend
```console
$ flux jobs -f completed,failed,timeout --since 'last fri night'
```

### DIFF
--- a/doc/man1/flux-jobs.rst
+++ b/doc/man1/flux-jobs.rst
@@ -36,8 +36,22 @@ OPTIONS
 **-u, --user**\ *=[USERNAME|UID]*
    List jobs for a specific username or userid. Specify *all* for all users.
 
+**--name**\ *=[JOB NAME]*
+   List jobs with a specific job name.
+
 **-c, --count**\ *=N*
    Limit output to N jobs (default 1000)
+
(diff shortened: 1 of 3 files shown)


chu11 (approved):
Awesome, LGTM!

grondo:
Thanks! I'll set MWP.

PR #4518, opened 2022-08-24
testsuite: add more checkpoint sequence tests

chu11:
splitting out from #4492, as that PR has evolved this cleanup / extra tests sort of don't belong there anymore.


garlick (approved):
LGTM!

PR #4519, opened 2022-08-24
broker: forward content.checkpoint-{get,put} RPCs to rank 0

chu11:
Per a discussion in #4492, there was a mistake in PR #4463 were non-rank 0 brokers did not forward checkpoint RPCs to rank 0.  Instead, they were forwarded to rank 0's backing module.  It has little effect / meaning before the work in #4492, but felt it wise to split it out into its own PR as it solves a specific thing.  I also put some of the work from #4492 into this PR, since it made sense.  Most notably, the movement of the content-checkpoint code out of `content-cache` and into its own `content-checkpoint` files.  

Also added a helper lib for the content tests, b/c I was getting irritated constantly modifying all three content tests all of the time.

### DIFF
--- a/src/broker/content-cache.c
+++ b/src/broker/content-cache.c
@@ -27,6 +27,7 @@
 
 #include "attr.h"
 #include "content-cache.h"
+#include "content-checkpoint.h"
 
 /* A periodic callback purges the cache of least recently used entries.
  * The callback is synchronized with the instance heartbeat, with a
@@ -99,6 +100,8 @@ struct content_cache {
     uint64_t acct_size;             // total size of all cache entries
     uint32_t acct_valid;            // count of valid cache entries
     uint32_t acct_dirty;            // count of dirty cache entries
+
+    struct content_checkpoint *checkpoint;
 };
 
 static void flush_respond (struct content_cache *cache);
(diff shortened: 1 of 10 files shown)


src/broker/content-cache.c
```diff
@@ -645,7 +645,7 @@ void content_checkpoint_get_request (flux_t *h, flux_msg_handler_t *mh,
     struct content_cache *cache = arg;
     const char *topic = "content-backing.checkpoint-get";
     const char *s = NULL;
-    const flux_msg_t *msgcpy = flux_msg_incref (msg);
+    const flux_msg_t *msgref = flux_msg_incref (msg);
```
garlick: I suggest instead of renaming the variable, we get rid of it and fix the refcounting problem on the error paths, as discussed in the other PR.

While you're at it, may as well fix the future leak that occurs when aux_set or then fails.

t/t0024-content-s3.t
```diff
@@ -127,6 +127,20 @@ test_expect_success HAVE_JQ 'checkpoint-get foo returned rootref bar' '
...
+
+test_expect_success HAVE_JQ 'checkpoint-get on rank 1 forwards to rank 0' '
+       echo rankref >rankref.exp &&
+       o=$(checkpoint_get_msg rankone) &&
+       jq -j -c -n ${o} \
+	   | flux exec -r 1 ${RPC} content.checkpoint-get \
+	   | jq -r .value | jq -r .rootref > rankref.out &&
+       test_cmp rankref.exp rankref.out
+'
+
```
garlick, line 143: These tests really don't need to be repeated in every backing store since the code they are testing is in the broker not the backing stores...  Just pick one?

garlick (approved):
Couple of nits, otherwise good.

chu11:
re-pushed, removed those extra tests per comments above.  Instead of the `msgcpy` -> `msgref` commit, I just added a commit that refactored things to remove the temp variable and fixup the flux future double decref issue.

garlick (approved):
LGTM!

Issue #2657, opened 2020-01-19
Add Python black formatter into docker images


dongahn:
Black formatter is required for our python code style. We might as well install python black formatter into docker. It is possible this has already been added since I'm using a slightly old docker image though.

grondo:
The black formatter is pulled in by `pip` before docker-run-checks in travis, and is installed with `--upgrade` so that the latest version is always run. Installing into the docker image isn't a bad idea, but I'm afraid the version in the docker image will quickly become out of sync with the version used to do the actual testing...

SteVwonder:
Not sure if this would be helpful, but we could create a new Dockerfile/image for development that includes things like `gdb` and `black`.   It could be another "flavor" alongside `bionic` and `centos7`, or it could be built on top of those flavors and not pushed to docker hub (i.e., devs would just build it locally whenever they need it).  I *think* (could be way wrong) that either of these methods would keep the tools up-to-date w.r.t. the code contained within the same images.

dongahn:
I like that idea. I also certainly want one with totalview dependecies.

grondo:
> It could be another "flavor" alongside bionic and centos7

Just to save space, it should probably be built on top of one of the images that already contains other dependencies for flux (then you only have to save one extra fs layer with devel environment).

This could be treated like the `travis` image we have now.

SteVwonder:
> Just to save space, it should probably be built on top of one of the images that already contains other dependencies for flux (then you only have to save one extra fs layer with devel environment).

Sounds good.  From this thread, looks like the `pylint`, `black`, and `gdb` packages could be added.  If we are adding `gdb` I wonder if it is worth adding the debug symbol packages for our various dependencies.

vsoch:
Just wondering - what is the goal to install black into the image? Do people use it for development? Usually a black --check is more suitable during CI (that then builds the image) over providing it in the container. What is the use case here?

grondo:
I'm guessing that development tools like `black` `pylint` `gdb` `strace` etc would be valuable for people using the container for development (e.g. on a mac)

vsoch:
[...]

PR #4521, opened 2022-08-24
adding deps to docker bases for python debugging

vsoch:
This is a first shot to add some Python development libraries into (what I think are?) the base containers.
This includes black, pylint, and pyflakes, along with gdb and strace.

This will close https://github.com/flux-framework/flux-core/issues/2657.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,10 @@ libtool
 .libs/
 libltdl/
 
+# python environments
+env
+.eggs
+
 # generated library version header
 /src/common/libflux/version.h
 
--- a/src/bindings/python/Dockerfile
+++ b/src/bindings/python/Dockerfile
@@ -1,0 +1,48 @@
+FROM python:3.9
+
+ARG FLUX_RELEASE=0.42.0
+ENV ${FLUX_RELEASE}=${FLUX_RELEASE}
+
+# from root:
+# docker build -t ghcr.io/flux-framework/flux-python -f src/bindings/python/Dockerfile .
+# docker run -it ghcr.io/flux-framework/flux-python
+
+RUN apt-get update \
+ && apt-get -qq install -y --no-install-recommends \
+        libsodium-dev \
+        libzmq3-dev \
+        libczmq-dev \
+        libjansson-dev \
+        libmunge-dev \
+        libncursesw5-dev \
+        lua5.4 \
+        liblua5.4-dev \
+        liblz4-dev \
+        libsqlite3-dev \
+        uuid-dev \
+        libhwloc-dev \
+        libmpich-dev \
+        libs3-dev \
+        libevent-dev \
+        libarchive-dev \
+        libpam-dev && \
+        ldconfig && \
+        rm -rf /var/lib/apt/lists/*
+
+RUN wget https://github.com/flux-framework/flux-core/releases/download/v${FLUX_RELEASE}/flux-core-${FLUX_RELEASE}.tar.gz && \
+    tar -xzvf flux-core-${FLUX_RELEASE}.tar.gz && \
+    mv flux-core-${FLUX_RELEASE} /code
+WORKDIR /code
+
+# Flux needs to find "system python" here or burps an error
+RUN ln -s $(which python) /usr/bin/python && \
+    ./configure --prefix=/usr/local --without-python && \
+    rm -rf /code/src/bindings/python
+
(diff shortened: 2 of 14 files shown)


grondo:
We had talked about creating separate devel docker images with these packages added, since they unnecessarily add to the size of the base images used in CI (though admittedly this is convenient, it is rare to use all these containers for development, and in that case the required packages can be dynamically added.)

It probably isn't the end of the world if we add these though, but I wonder if we should choose just one image to enhance to start?

vsoch:
@grondo I'm creating a set of images under /etc/docker that are multistage build with spack - and I wasn't looking closely at these yet because I thought "they are production don't change" - but you are saying you want these to also be multistage (but not use spack)?

grondo:
I was only saying that the existing containers built from `src/test/docker/*` are mostly for CI test, so they don't necessarily need to have devel packages like gdb, python black and pylint, etc. installed. The Dockerfiles there form the base for the `fluxrm/testenv` images which are then pulled during flux-core CI and subsequently create the `fluxrm/flux-core` images for use in `flux-sched` build etc.

That being said, I'm not an expert here and if there is a better way, I'm all for that! Though I think we do want to keep the CI running against the current set of distros, and not spack only, but that shouldn't influence how we create development containers.

vsoch:
oh derp! Where are the "production flux" images found?

vsoch:
oh I see, I found this script that generates a build matrix - https://github.com/flux-framework/flux-core/runs/8001881082?check_suite_focus=true so there aren't actually Dockerfile for the final builds, it's all in code... :sob: 

How would someone ever reproduce building a flux image with so many layers? Unless they are an ogre and have layers. :onion:

grondo:
> oh derp! Where are the "production flux" images found?

Maybe there is some confusion. We don't have production images afaik. We generate docker images mainly to facilitate testing in CI and to some extent standalone on a specific distro.

Flux is intended to be system software and not run out of containers.

I agree though the CI built containers are messy and probably do "bad things", but nobody has had the time (or more likely the knowledge) to clean up that.

> How would someone ever reproduce building a flux image with so many layers? 

I admit I'm unclear on the use case to "reproduce building a flux image". Maybe if we figure out what the use case is we can find a better way?

vsoch:
We could also just leave the CI as is, and I could integrate these updates in the TBA spack images. I'm still waiting on that view bug to be fixed so we get our first base container!

grondo:
> I have preferences maybe about the base image. So I need a FROM for my build. Do I use the CI images for that?

I think you can do that, but the images are unfortunately large, so maybe as you said the spack-based images would be better for that.

If you do use `FROM fluxrm/flux-sched`, then you could add the extra development files there, so we don't need to add them to the CI images?

@trws also had the idea long ago to create an "end user" image in CI by building flux-core in the CI image, then installing it to a different container that only has the the runtime requirements, so it would be nice and slim with less ogre-like layers. I don't know how to do that though :-) If we did it, it would probably be just another action in the CI workflow?

vsoch:
> If we did it, it would probably be just another action in the CI workflow?

I think for anything a user is going to go off of, I'd like to see an actual Dockerfile that can be used. There is nothing more frustrating as a container developer as going to a repository looking for a clear Dockerfile, and either:

1. not finding it, period
2. finding the wrong ones (as I did here)

Do you know who made this original matrix thing? I think what I'd like to see are a core set of Dockerfile bases, on per OS,with build args that populate specifics. And then we could have two levels - bases and multistage builds, each of which would be provided as Dockerfile. But I'm not familar with the setup here so I'm not sure what it's doing / what we'd need to extend to this idea.

grondo:
> Do you know who made this original matrix thing?

The original matrix thing was developed by me when we moved from Travis to Github CI. It just generates the matrix of builds for CI and has nothing to do with Docker (except that the builds occur in docker via our `docker-run-checks.sh` script). (note that multistage builds did not exist when we started this.)

@trws was the one that originally developed the `testenv` base images for our CI, and then we evolved it from there.

If we want to make a new Dockerfile for end users, that would be great. As I said, Dockerfiles isn't a focus for us since this is system software. Maybe we can do this work in a `flux-framework/docker` repo or `flux-framework/flux-in-docker`? So it is easier to find?

vsoch:
I've hijacked this branch to work on something else, so closing here.

PR #4523, opened 2022-08-26
fetch J instead of jobspec in the job shell, support flux job info --original jobspec

grondo:
This PR is prepares some consumers of jobspec for the possibility that Flux instance may store redacted and/or modified jobspec in the "jobspec" KVS key.

First, a new `flux job info --original` option is added which, when used with the `jobspec` key, fetches and "unwraps" J into jobspec. If the jobspec in the KVS is redacted, this will be the only way to get the originally submitted jobspec.

Similarly, the job shell is modified to fetch and verify J instead of the jobspec from the job-info service.

In order to support decoding J into a jobspec string in two places, a new function `unwrap_string()` is added (to libjob since there didn't seem to be a better place) to avoid cut and pasting the code to unwrap signed data into a NUL terminated string (when built with or without flux-security).

Finally, some modifications and additions to the testsuite are made to test and adapt to these changes.

### DIFF
--- a/src/cmd/Makefile.am
+++ b/src/cmd/Makefile.am
@@ -105,6 +105,8 @@ flux_start_LDADD = \
 flux_job_LDADD = \
 	$(fluxcmd_ldadd) \
 	$(top_builddir)/src/shell/libmpir.la \
+	$(top_builddir)/src/common/libjob/libjob.la \
+	$(top_builddir)/src/common/libutil/libutil.la \
 	$(top_builddir)/src/common/libczmqcontainers/libczmqcontainers.la \
 	$(top_builddir)/src/common/libdebugged/libdebugged.la \
 	$(top_builddir)/src/common/libterminus/libterminus.la
--- a/src/common/libjob/unwrap.c
+++ b/src/common/libjob/unwrap.c
@@ -1,0 +1,116 @@
+/************************************************************\
+ * Copyright 2022 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include <errno.h>
+#include <string.h>
+#include <stdint.h>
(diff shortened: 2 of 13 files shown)


grondo:
I had to add a fix for `TEST_CHECK_PREREQS` handling.

src/cmd/Makefile.am
```diff
@@ -105,6 +105,8 @@ flux_start_LDADD = \
 flux_job_LDADD = \
```
garlick, line 105: Commit message:

s/that a submitted/that a user submitted/

src/common/libjob/unwrap.c
```diff
@@ -0,0 +1,116 @@
...
+    if (sign_none_unwrap (s, (void **) &data, &len, &uid) < 0) {
+        errprintf (errp, "sign-none-unwrap failed: %s", strerror (errno));
+        return NULL;
+    }
+    if (verify && uid != getuid ()) {
+        errprintf (errp,
+                   "sign-none-unwrap: signing userid %lu != current %lu",
+                   (unsigned long) uid,
+                   (unsigned long) getuid ());
+        free (data);
```
garlick, line 48: free potentially clobbers errno and I _think_ functions that call this one might expect errno to be valid (e.g. in the shell).

garlick (approved):
Just a couple of minor nits.  LGTM!

src/common/libjob/unwrap.c
grondo: Hm, I tried to explicitly _not_ depend on errno here (since what detail can errno give in this case), e.g. the comment on `unwrap_string()`:

```c
 *  If verify is true, then fail if signing mechanism is invalid or
 *   signing user does not match current uid. On failure, error.text
 *   is filled in with an error message. (errno not necessarily
 *   guaranteed to be valid).
```

The job shell uses `shell_log_error()` which logs an error but doesn't call `strerror()` (as `shell_log_errno()` would I believe).

Did you see a place where errno is expected to be valid?

garlick: My mistake, sorry.  I saw `shell_log_error()` and assumed.  Also I didn't read your inline doc carefully.  Never mind!

grondo: Yeah, unfortunate name collision between `shell_log_error()` and `log_error()`

grondo:
Fixed commit message typo and I'll set MWP.

PR #4524, opened 2022-08-26
[WIP] broker: do not drop dirty cache entries on error

### LINKED ISSUE #4472: broker: content-cache `acct_dirty` can be greater than the number of flushable entries
While working on #4267, I noticed that sometimes the `acct_dirty` field can be greater than the length of the `flush` list, even when a backing module is not loaded.  a `flux content flush` will therefore hang, b/c `acct_dirty` will never become zero, so the content flush will never respond to the original request.

I'm not sure how this happened, but here are several possibilities when looking through code.  Some of this is duplicated text from #4267.

1) one potential path is in `cache_store()` and `cache_store_continuation()`.  An entry from the flush list is sent to the content backing module for backing and removed from the flush list.  If an error occurs, `cache_store_continuation()` will see the error but not put it back on the flush list and `acct_dirty` will not be decremented.

2) Another potential (similar) path may be when we unload the backing-module, so that backing store requests get lost.  `acct_dirty` will stay incremented even though flush entries have been taken off the list. 

(Update, now that I think about it #2 is a strong possibility.  I noticed this in the regression tests, which one test fix of #4267 went into affect, content-sqlite was being loaded / reloaded regularly in the regression tests.)

chu11:
Problem: If an error occurs during a content store, a dirty cache entry can be lost forever as it not on the flush list or the lru list.  As a result, the "dirty count" of entries will be inconsistent to the known dirty entries (i.e. entries in the flush list or in the process of being stored).

Solution: If a content store fails, add the entry to a new flush_errors list so it can be tried again during a forced content flush.

Fixes #4472

Notes:

While I'm in the middle of all this broker cache code, I thought it would be wise to try and fix this up before forgetting about this chunk of the broker.  I'm marking WIP b/c:

A) this is based on code review while trying to figure out #4482.  We actually haven't had this be an issue.

B) while not very complex, it maybe adds a bit more complexity than I wanted to.  I couldn't put the dirty entries back on the flush list b/c that could lead to "infinite loop" as dirty cache entries get retried over and over again to be stored (assuming error on-going).

B2) we could make more optimal by checking for certain error conditions, but elected not to do that for now

C) I don't know how to test this without introducing a lot of instrumentation.  So punted on that for now.

So I'm thinking either:

1) we consider merging based on just code review

2) We park this code as "for future knowledge" in case we ever need to look into it more seriously down the road

### DIFF
--- a/src/broker/content-cache.c
+++ b/src/broker/content-cache.c
@@ -89,6 +89,8 @@ struct content_cache {
 
     struct list_head lru;           // LRU is for valid, clean entries only
     struct list_head flush;         // dirties queued due to batch limit
+    struct list_head flush_errors;  // flush entries that hit error
+    unsigned int flush_errors_len;
 
     uint32_t blob_size_limit;
     uint32_t flush_batch_limit;
@@ -484,10 +486,21 @@ error:
  */
 static void cache_resume_flush (struct content_cache *cache)
 {
-    if (cache->acct_dirty == 0 || (cache->rank == 0 && !cache->backing))
+    if (cache->acct_dirty == 0
+        || (cache->acct_dirty - cache->flush_errors_len) == 0
+        || (cache->rank == 0 && !cache->backing))
         flush_respond (cache);
     else
         (void)cache_flush (cache); /* resume flushing, subject to limits */
+}
+
+/* Call list_del() just in case entry is already on the list. */
+static void flush_errors_list_append (struct content_cache *cache,
+                                      struct cache_entry *e)
+{
+    list_del (&e->list);
+    list_add_tail (&cache->flush_errors, &e->list);
+    cache->flush_errors_len++;
 }
 
 static void cache_store_continuation (flux_future_t *f, void *arg)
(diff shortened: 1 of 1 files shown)


garlick:
I'd wager that the most likely cause of failure would be a full file system, and no amount of retries is going to fix that.  For now, I would suggest we log a LOG_CRIT message to be sure the first error is captured  by the systemd journal (ENOSPC for example) and just stop the reactor.   It would be OK IMHO to forgo testing for this for now.

chu11:
> I'd wager that the most likely cause of failure would be a full file system, and no amount of retries is going to fix that. For now, I would suggest we log a LOG_CRIT message to be sure the first error is captured by the systemd journal (ENOSPC for example) and just stop the reactor. It would be OK IMHO to forgo testing for this for now.

Hmmm, I wasn't contemplating going down this path.

Alternate idea, could we unregister the backing module? (this goes along with PR #4492, things could keep on working for awhile).

chu11:
> We may want to just park this for a while though. Maybe a distributed/replicated content store (as @trws was alluding to in slack earlier) would reduce the odds of getting here and then we might prefer the "fail fast" alternative in the unlikely event that we do.

Yeah, lets park it for the time being.  I do like your idea of logging to LOG_CRIT instead.  Perhaps I'll just do that and call it a day.  I'll put that in a different PR since I want to preserve this branch for the future.

PR #4526, opened 2022-08-26
broker: log content store errors to LOG_CRIT

chu11:
Problem: In the event of content store errors, such as ENOSPC
if the disk is full, error messages are logged to LOG_ERR.  This
does not properly signify the severity of the error, as something
like ENOSPC would mean no data is being backed up to disk.

Solution: Log to LOG_CRIT, to signify the importance of this
issue and the need for it to be resolved.

### DIFF
--- a/src/broker/content-cache.c
+++ b/src/broker/content-cache.c
@@ -505,7 +505,8 @@ static void cache_store_continuation (flux_future_t *f, void *arg)
             flux_log (cache->h, LOG_DEBUG, "content store: %s",
                       "backing store service unavailable");
         else
-            flux_log_error (cache->h, "content store");
+            flux_log (cache->h, LOG_CRIT, "content store: %s",
+                      strerror (errno));
         goto error;
     }
     if (hash_size != content_hash_size


src/broker/content-cache.c
```diff
@@ -505,7 +505,7 @@ static void cache_store_continuation (flux_future_t *f, void *arg)
             flux_log (cache->h, LOG_DEBUG, "content store: %s",
                       "backing store service unavailable");
         else
-            flux_log_error (cache->h, "content store");
+            flux_log (cache->h, LOG_CRIT, "content store");
```
garlick, line 508: Should be
```c
flux_log (cache->h, LOG_CRIT, "content store: %s", strerror (errno));
```

chu11: oh yeah, no strerror output when you take out the error ... my bad, sloppy :P

chu11:
re-pushed, fixing up the 1 line ... in my 1 line patch :P

garlick (approved):
:+1:

Issue #4311, opened 2022-05-02
automate KVS garbage collection


garlick:
Problem: assuming #4303 is merged, we have a mechanism for offline KVS garbage collection that has to be manually invoked by running `flux shutdown --gc`.  We need a way to trigger this automatically based on an estimate of how productive it would be.

For example, maybe if `flux shutdown` is invoked without `--gc`, and the KVS root sequence number is past some threshold, OR the number of jobs purged is beyond some threshold, garbage collection could be automatically selected.  Possibly with a prompt e.g.
```
$ sudo flux shutdown
OK to perform dump/restore for offline KVS garbage collection (Y/N)?
```

garlick:
Couple of notes:
- once #4446 is resolved, the sequence number counting from 0 to some threshold may be used as an indicator of how "busy" the KVS has been since the last dump.
- the root sequence number resets to zero across GC because the checkpoint is not part of the dump
- The `flux restore --checkpoint` called from rc1 ensures that a new checkpoint (seq=0) is written during startup after the dump is restored.  The kvs module on rank 0 initializes from this when is loaded.
- The sequence threshold should be configurable via TOML `[kvs]` table and be disabled by default
- The actual GC should be triggered by `flux shutdown`, to be carried out in `rc1`.  If the broker is stopped by `systemctl stop flux` instead, GC should not be attempted because it risks getting the rank 0 broker killed on systemd timeout.

PR #4528, opened 2022-08-30
kvs: support gc-threshold config 

chu11:
Problem: KVS garbage collection is only done when an administrator runs flux-shutdown and chooses to garbage collect via the --dump or --gc options.

Solution: Support a kvs gc-threshold configuration option.  This configuration will take an integer count of KVS changes (the KVS version number or sequence number).  Once the threshold has been crossed, flux-shutdown will ask the user if they wish to garbage collect.  This offers an easy way for administrators to be reminded of garbage collection on a regular basis.

Fixes https://github.com/flux-framework/flux-core/issues/4311

### DIFF
--- a/doc/man5/flux-config-kvs.rst
+++ b/doc/man5/flux-config-kvs.rst
@@ -22,6 +22,13 @@ checkpoint-period
    primary namespace.  The checkpoint is used to protect against data
    loss in the event of a Flux broker crash.
 
+gc-threshold
+   (optional) Sets the number of KVS commits (distinct root snapshots)
+   after which offline garbage collection is performed by
+   :man1:`flux-shutdown`. A value of 100000 may be a good starting
+   point. (Default: garbage collection must be manually requested with
+   `flux-shutdown --gc`).
+
 
 EXAMPLE
 =======
@@ -30,7 +37,7 @@ EXAMPLE
 
    [kvs]
    checkpoint-period = "30m"
-
+   gc-threshold = 100000
 
 RESOURCES
 =========
@@ -43,4 +50,4 @@ RFC 23: Flux Standard Duration: https://flux-framework.readthedocs.io/projects/f
 SEE ALSO
 ========
 
-:man5:`flux-config`
+:man1:`flux-shutdown`,:man5:`flux-config`
--- a/src/cmd/builtin/shutdown.c
+++ b/src/cmd/builtin/shutdown.c
@@ -12,12 +12,87 @@
 # include <config.h>
 #endif
 #include <unistd.h>
+#include <jansson.h>
 #include <flux/core.h>
 
 #include "src/broker/state_machine.h"
+#include "src/common/libkvs/kvs_checkpoint.h"
 #include "src/common/libutil/uri.h"
 
 #include "builtin.h"
+
+static void get_kvs_version (flux_t *h, int *version)
+{
+    (*version) = 0;
+    if (flux_kvs_get_version (h, NULL, version) < 0
+        && errno != ENOSYS)
+        log_err_exit ("Error fetching KVS version");
+}
+
+static void get_gc_threshold (flux_t *h, int *gc_threshold)
+{
+    flux_future_t *f;
+    json_t *o;
+    (*gc_threshold) = 0;
+    if (!(f = flux_rpc (h, "config.get", NULL, FLUX_NODEID_ANY, 0))
+        || flux_rpc_get_unpack (f, "o", &o) < 0)
+        log_msg_exit ("Error fetching flux config: %s",
+                      future_strerror (f, errno));
+    (void)json_unpack (o, "{s:{s:i}}", "kvs", "gc-threshold", gc_threshold);
+}
+
+static int askyn (char *prompt, bool default_value, bool *result)
+{
+    while (1) {
+        char buf[16];
+        printf ("%s [%s]? ", prompt, default_value ? "Y/n" : "y/N");
+        fflush (stdout);
+        if (fgets (buf, sizeof (buf), stdin) == NULL)
+            return -1;
+        if (buf[0] == '\n')
+            break;
+        if (buf[0] == 'y' || buf[0] == 'Y') {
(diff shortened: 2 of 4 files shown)


src/cmd/builtin/shutdown.c
```diff
@@ -12,13 +12,54 @@
...
 
+static void get_checkpoint_sequence (flux_t *h, int *seq) {
+    flux_future_t *f;
+    (*seq) = 0;
+    if (!(f = kvs_checkpoint_lookup (h, NULL, 0))
+        || (kvs_checkpoint_lookup_get_sequence (f, seq) < 0
+            && errno != ENOENT))
+        log_msg_exit ("Error fetching checkpoint sequence: %s",
+                      future_strerror (f, errno));
+}
```
garlick, line 32: Quick comment as I head out to a doctor's apt:
- open brace for the function should be on a line by itself per project style
- I think we'll want to check the current live kvs root sequence number, not the stored checkpoint since the latter will be out of date

doc/man5/flux-config-kvs.rst
```diff
@@ -22,6 +22,13 @@ checkpoint-period
    primary namespace.  The checkpoint is used to protect against data
    loss in the event of a Flux broker crash.
 
+gc-threshold
+   (optional) Sets the KVS garbage collection sequence number
+   threshold.  Once this threshold is crossed in the KVS checkpoint,
+   it will inform :man1:`flux-shutdown` to ask the user to perform
+   offline KVS garbage collection.  It is recommended the
+   ``checkpoint-period`` configuration also be configured if this is
+   set.
```
garlick, line 31: Suggested change:
> (optional) Sets the number of KVS commits (distinct root snapshots) after which offline garbage collection is performed by flux-shutdown(1).    A value of 100000 may be a good starting point.  (Default: garbage collection must be manually requested with flux-shutdown --gc).

chu11:
re-pushed with fixes per comment above.

src/cmd/builtin/shutdown.c
```diff
@@ -12,13 +12,54 @@
...
+
+    get_kvs_version (h, &version);
+    get_gc_threshold (h, &gc_threshold);
+
+    if (gc_threshold > 0 && version > gc_threshold) {
+        char *s = NULL;
+        printf ("gc threshold exceeded, do you want to perform garbage collection (Y/n)? ");
+        scanf ("%ms", &s);
+        if (!s || strncasecmp (s, "y", 1) == 0)
+            rc = true;
```
garlick, line 57: If I just type a carriage return here, I get no feedback but scanf is still waiting for input.  Maybe something like the following would work better, and would handle stdin not a tty also (e.g. when scripted)
```c
bool askyn (char *prompt, bool default_value)
{
    char buf[16];
    int i;

    for (;;) {
        printf ("%s [%s]? ", prompt, default_value ? "Y/n" : "y/N");
        fflush (stdout);
        if (!isatty (STDIN_FILENO)
            || fgets (buf, sizeof (buf), stdin) == NULL
            || buf[0] == '\n')
            break;
        if (buf[0] == 'y' || buf[0] == 'Y')
            return true;
        if (buf[0] == 'n' || buf[0] == 'N')
            return false;
        printf ("Please answer y or n\n");
    };
    return default_value;
}
```

garlick (commented):
This is great!
Just one comment about the prompt machinery.

src/cmd/builtin/shutdown.c
chu11: looks good, and it seems that several builders didn't like the newer `%ms` that (I thought) was supported in all newer `scanf`s

garlick: on second thought, it might be a little weird to take the default on stdin EOF or `!isatty()`.  Perhaps this function should be modified so that it can return an error in those cases, and the caller could do something like abort and suggest a `-y` option?

Also, this seems like it could be a useful libutil function for reuse...

chu11:
re-pushed, I went with the approach of

- if tty, `flux shutdown` outputs the question for the user and they need to answer yes/no/<enter>
- if no tty, returns error unless user specifies `-y` or `-n`.
- add more tests to cover (i hope) all scenarios

src/cmd/builtin/shutdown.c
```diff
@@ -12,13 +12,88 @@
...
+    json_t *o;
+    (*gc_threshold) = 0;
+    if (!(f = flux_rpc (h, "config.get", NULL, FLUX_NODEID_ANY, 0))
+        || flux_rpc_get_unpack (f, "o", &o) < 0)
+        log_msg_exit ("Error fetching flux config: %s",
+                      future_strerror (f, errno));
+    (void)json_unpack (o, "{s:{s:i}}", "kvs", "gc-threshold", gc_threshold);
+}
+
+int askyn (char *prompt, bool default_value, bool *result)
```
garlick, line 44: Suggestion: make `askyn()` static like the other local functions.

src/cmd/builtin/shutdown.c
```diff
@@ -116,6 +193,14 @@ static struct optparse_option opts[] = {
...
                " 1=show all log messages",
     },
+    { .name = "yes", .key = 'y', .has_arg = 0,
+      .usage = "If garbage collection threshold exceeded, "
+               "perform garbage collection",
+    },
+    { .name = "no", .key = 'n', .has_arg = 0,
+      .usage = "If garbage collection threshold exceeded, "
+               "do not perform garbage collection",
+    },
```
garlick, line 203: How about for the usage just "answer y to all y/n questions"?

garlick (approved):
Just a couple of minor nits o/w LGTM!
(I did test it out on my test system).

I think this is an important feature so thanks for pushing it through!

src/cmd/builtin/shutdown.c
chu11: and will tweak manpage as well

chu11:
re-pushed with fixes per comments above

Issue #4522, opened 2022-08-25
flux-mini: add --no-attach option to alloc subcommand


grondo:
A use case that came up from early users is to allocate resources without attaching to the resulting instance then submit work to that instance from "outside". This is the equivalent of `salloc --no-shell` and `srun --jobid=JOBID`.

The equivalent now is something like
```console
$ jobid=$(flux mini batch -N2 --wrap sleep inf)
$ flux job wait-event $jobid memo
$ flux proxy $jobid
```

But the `wait-event` and `sleep inf` are a bit of a hassle. We should consider adding an option to `flux mini alloc` to skip the `flux job attach` step, do the equivalent of `sleep inf` (perhaps set rc2_none?), and instead wait until the URI is available from the subinstance, e.g. the above would become:

```console
$ jobid=$(flux mini alloc -N2 --no-attach)
$ flux proxy $jobid
```

grondo:
Something came up while trying to implement something similar in the current flux version.

The `memo` event for the instance URI is emitted early in broker startup, and thus the instance may not yet be ready to accept jobs -- i.e. `flux proxy $jobid flux mini submit ...` may fail immediately after the `memo` event. To rectify, we'll need some way in `flux mini alloc --no-attach` to block until the instance is ready to accept work. (@garlick suggests the `state-machine.wait` RPC).

Or, we could look into delaying the `memo` event until the instance is ready.

PR #4531, opened 2022-08-31
flux-mini: add --bg option to flux-mini alloc

grondo:
This PR adds a new `--bg` option to `flux mini alloc` which causes the program to print a jobid on stdout and exit once the new instance is ready, instead of interactively attaching. This should fix the use case described in #4522 (chose a slightly shorter option name).

With `-v`, the command will log what it is waiting on to stderr:
```console
$ flux mini alloc --bg -n1 -v 
ƒkdmkPjqR:  0.000s: waiting for resources
ƒkdmkPjqR:  0.026s: resources allocated
ƒkdmkPjqR:  0.094s: waiting for instance
ƒkdmkPjqR:  1.255s: ready
ƒkdmkPjqR
```

### DIFF
--- a/doc/man1/flux-mini.rst
+++ b/doc/man1/flux-mini.rst
@@ -47,9 +47,10 @@ file is read and submitted along with the job. If no SCRIPT is
 provided, then one will be read from *stdin*.
 
 **flux-mini alloc** works similarly to **batch**, but instead blocks until
-the job has started and interactively attaches to the new Flux instance.
-By default, a new shell is spawned as the initial program of the instance,
-but this may be overridden by supplying COMMAND on the command line.
+the job has started and interactively attaches to the new Flux instance
+(unless the ``--bg`` option is used).  By default, a new shell is spawned
+as the initial program of the instance, but this may be overridden by
+supplying COMMAND on the command line.
 
 The intent is for the "mini" commands to remain simple with stable interfaces
 over time, making them suitable for use in scripts.
@@ -393,6 +394,16 @@ OTHER OPTIONS
 **--debug**
    Enable job debug events, primarily for debugging Flux itself.
    The specific effects of this option may change in the future.
+
+**--bg**
+   *(alloc only)* Do not interactively attach to the instance. Instead,
+   print jobid on stdout once the instance is ready to accept jobs. The
+   instance will run indefinitely until a time limit is reached, the
+   job is canceled, or it is shutdown with ``flux shutdown JOBID``
+   (preferred). If a COMMAND is given then the job will run until COMMAND
+   completes. Note that ``flux job attach JOBID`` cannot be used to
+   interactively attach to the job (though it will print any errors or
+   output).
 
 **-B, --broker-opts=OPT**
    *(batch only)* For batch jobs, pass specified options to the Flux brokers
--- a/src/cmd/flux-mini.py
+++ b/src/cmd/flux-mini.py
@@ -19,6 +19,7 @@ import re
 import random
 import itertools
 import atexit
+import time
 from itertools import chain
 from string import Template
 from collections import ChainMap
@@ -31,6 +32,7 @@ from flux import util
 from flux import debugged
 from flux.idset import IDset
 from flux.progress import ProgressBar
+from flux.uri import JobURI
 
 
 class Dependency:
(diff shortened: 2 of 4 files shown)


src/cmd/flux-mini.py
```diff
@@ -1611,13 +1627,88 @@ def init_jobspec(self, args):
...
+        ts = event.timestamp - self.t0
+
+        if args.verbose and event.name == "alloc":
+            self.log(jobid, ts, "resources allocated")
+        if event.name == "memo" and "uri" in event.context:
+            if args.verbose:
+                self.log(jobid, ts, "waiting for instance")
+
+            #  Wait for child instance to start rc2 using state-machine.wait
+            #  Then stop the reactor to return to caller.
```
garlick, line 1671: It would be slightly more accurate to say "Wait for child instance to complete rc1...", since after the RPC succeeds we just know that the RUN state was entered (rc2 would be started  in that state, if configured, but we don't know if it's happened yet).

Note that the RPC fails if the broker is in a state that cannot lead to RUN (e.g. if rc1 fails, it skips to CLEANUP)

doc/man1/flux-mini.rst
```diff
@@ -394,6 +395,10 @@ OTHER OPTIONS
    Enable job debug events, primarily for debugging Flux itself.
    The specific effects of this option may change in the future.
 
+**--bg**
+   *(alloc only)* Do not interactively attach to the instance. Instead,
+   print jobid on stdout once the instance is ready to accept jobs.
+
```
garlick, line 401: Is it worth noting that in this mode, `flux job attach` cannot be used to interactively attach later?

garlick (approved):
LGTM and seems to work as advertised.
Just had a couple of cosmetic suggestions.

doc/man1/flux-mini.rst
grondo: Yes, good idea. 

Some background:

I actually had the command starting an interactive shell at first so that users could later `flux job attach` to their job. In this case, you can still cancel the job or tell it to shutdown and the shell will get SIGHUP or killed and the job will exit. However, the jobs then are recorded as FAILED.

With `broker.rc2_none` set, it seemed like a nice side effect that you can `flux shutdown $jobid` and the job will complete successfully instead of failing.

src/cmd/flux-mini.py
grondo: Thanks! Note that when this RPC fails an exception should be raised from Python (i.e. the "return value" of the RPC is implicitly checked).  I should add a test for this in the testsuite (right now the only test ensures that the command fails when the broker is given an invalid argument)

doc/man1/flux-mini.rst
garlick: That makes good sense.  Maybe it's useful to mention that `flux shutdown $jobid` is the expected way the instance should be stopped then?

grondo: Yeah, probably. 

Though that command issues a mass of broker log messages on stdout by default, I was thinking we should suppress that by default without -v if normal users were going to run it. I wasn't sure what you thought of that in the typical case of a sysadmin using it on the system instance though.

src/cmd/flux-mini.py
grondo: Just verifying this works: the RPC `.get()` throws an exception on failure, but the message would probably confuse a normal user:
```
FileNotFoundError: [Errno 2] broker has surpassed RUN state
```
This error should probably be caught and rewritten to something more clear.

grondo:
Ok, I think I've addressed the couple comments above and forced a push.

doc/man1/flux-mini.rst
garlick: What about suggesting `flux shutdown --quiet`?

I'd be a bit worried about changing the default on that one just because if something hangs, it's nice to know where in the process you are.  I guess I'm primarily concerned about system instances there.

grondo: Yeah, I was worried about changing default behavior as well. It isn't a big enough problem to worry about at the moment.

garlick (approved):
Thanks, LGTM!

grondo:
Thanks, I'll set MWP presuming tests pass.

grondo:
Hm, many of the builds are timing out, likely a race condition or something in the tests :-( (or maybe in the command itself) This should be fun to track down.

Issue #3731, opened 2021-06-17
flux-proxy breaks under tcsh when forwarding LD_LIBRARY_PATH


SteVwonder:
The `LD_LIBRARY_PATH` forwarding that was added in https://github.com/flux-framework/flux-core/pull/3458 seemingly breaks under tcsh.  A user reported `LD_LIBRARY_PATH=/usr/tce/packages/…/lib/intel64 command not found` when trying to use `flux-proxy` with `tcsh` as their shell.

`strace`ing the flux-proxy reveals what the generated command line looks like:
```
❯ strace -f -e execve /usr/global/tools/flux/toss_3_x86_64_ib/flux-c0.26.0-s0.16.0/bin/flux proxy ssh://quartz28/tmp/ hostname                                                                                                 17:25:14 ()
execve("/usr/global/tools/flux/toss_3_x86_64_ib/flux-c0.26.0-s0.16.0/bin/flux", ["/usr/global/tools/flux/toss_3_x8"..., "proxy", "ssh://quartz28/tmp/", "hostname"], 0x7fffffffc5a0 /* 107 vars */) = 0
strace: Process 47913 attached
[pid 47913] execve("/usr/bin/rsh", ["/usr/bin/rsh", "quartz28", "LD_LIBRARY_PATH=/usr/lib64", "/collab/usr/global/tools/flux/to"..., "relay", "/tmp/"], 0x648a70 /* 114 vars */) = 0
```

I believe the issue just comes down to how bash/zsh interpret inline env vars vs tcsh:
```
❯ bash -c 'LD_LIBRARY_PATH=/foo printenv LD_LIBRARY_PATH'                                                                                                                                                                      17:33:53 ()
/foo

❯ tcsh -c 'LD_LIBRARY_PATH=/foo printenv LD_LIBRARY_PATH'                                                                                                                                                                      17:33:59 ()
LD_LIBRARY_PATH=/foo: Command not found.
```

We could fix in a few ways.  The two obvious ones to me are:

- We could revert the PR so that LD_LIBRARY_PATH is not forwarded and just make sure our packaging (mainly in /usr/global) does not produce flux binaries that rely on `module load`ed libraries (e.g., `libcudart`).
- We could use something like `env` or `bash` to ensure that this works regardless of the users shell:
  - `rsh quartz28 bash -c 'LD_LIBRARY_PATH=/usr/lib64 /path/to/flux relay ...'`
[...]

PR #4532, opened 2022-08-31
fix ssh connector with csh/tcsh shells

grondo:
This is a proposed fix for #3731. I still have to ensure it works for a user with `csh` or `tcsh` and find some way to add a regression test. However, I'm posting as an early WIP to get any feedback.

### DIFF
--- a/src/connectors/ssh/ssh.c
+++ b/src/connectors/ssh/ssh.c
@@ -152,6 +152,8 @@ int build_ssh_command (const char *uri_path,
 
     /* LD_LIBRARY_PATH */
     if (ld_lib_path) {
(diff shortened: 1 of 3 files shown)


garlick:
Hey @chu11, aren't you a tcsh user?  Maybe you could give this a quick test?

grondo:
Thanks! I think I thought of a way to test this in the testsuite as well. Testsuite updates pushed, we'll see how this goes in CI (not sure if any builders have `/bin/tcsh`, and I changed `t/scripts/tssh`), but it seemed to work locally.

garlick (approved):
LGTM!

grondo:
Thanks!

Issue #4536, opened 2022-09-01
jobs are unnecessarily constrained to cores assigned to system instance broker


grondo:
Sites may choose to run the system instance broker on a subset of cores, e.g. it may be policy to run all system processes on a constrained list of resources to preserve other resources for user tasks. If this is the case, then the XML loaded by the flux-core `resource` module is also constrained to those same cores, since by default we restrict the hwloc topology to the current cpumask. Since the shell reads the topology XML directly from the `resource` module, it is also restricted to the cpumask of the system instance broker and so on down the line.

e.g. on corona the system instance broker is restricted to the second sibling PU of each core as inherited from PID 1 (unsure if this is site or systemd policy):
```console
$ taskset -cp $(flux getattr broker.pid)
pid 835678's current affinity list: 48-95
$ taskset -cp 1
pid 1's current affinity list: 48-95
```

This causes jobs to be restricted to the same cpumask:

```console 
$ flux mini alloc -N1 sh -c 'taskset -cp $$'
pid 3324859's current affinity list: 48-95
```

And thus jobs run within this instance can only use the second PU of each core:
```console
$ flux mini alloc -N1 flux mini run -n1 -c4 sh -c 'taskset -cp $$'
pid 1382482's current affinity list: 92-95
```

It seems useful to allow the system instance broker to be constrained to a subset of resources, but this should not affect jobs run within the instance by default. I propose the following changes to better handle this case:

 1. Add an option to the flux-core `resource` module to avoid calling `hwloc_topology_restrict()` to restrict the topology to the current available resources. We could suggest this option be used for the system instance.
 2. If the job shell is able to fetch the topology XML from the enclosing instance, then avoid further restricting the topology to the current cpumask. Rationale: The enclosing instance XML has already been (optionally) restricted by the `resource` module. If a "norestrict" option has been used, then we want to inherit this for jobs, which will then set their cpumask based on assigned resources, if not, then restricting again is redundant anyway.
 3. If the job shell is not able to fetch topology XML from the enclosing instance, then fall back to manually restricting  to the current cpumask.

These changes should allow the system instance broker to be constrained to any set of cores (via cpu affinity), but still allow jobs _full_ access to the assigned cores.

grondo:
I think this should have been closed by #4538

PR #4538, opened 2022-09-01
resource: add norestrict option to avoid restricting hwloc topology XML

grondo:
This PR ostensibly "solves" #4536 by allowing the `resource` module to skip the `hwloc_topology_restrict()` step when generating the hwloc XML with a new `norestrict` option in the config.

The idea is that sysadmins would set this option when the system flux instance is constrained by policy to a subset of available cores (e.g. all system processes are constrained to a system set). Then the XML in the resource module will contain the entire of set of system resources instead of a restricted set.

The shell then also avoids the topology restrict when pulling the XML from its enclosing instance, since this is redundant no matter what, and nicely solves the case of avoiding a restricted resource XML when the system instance broker is restricted by policy.


doc/man5/flux-config-resource.rst
```diff
@@ -32,6 +32,12 @@ exclude
    the scheduler, but will still be used to determine satisfiability of job
    requests until the instance is restarted.
 
+norestrict
+   (optional) Disable restricting of the loaded HWLOC topology XML to the
+   current cpu affinity mask of the Flux broker. This option should be use
```
garlick, line 37: s/use/used/

garlick (approved):
I'm a bit out of my depth with the hwloc api stuff but FWIW this looks good to me.
Just one minor typo noted.

grondo:
> I'm a bit out of my depth with the hwloc api stuff

Without mentioning the hwloc api: what happens without `norestrict` is that the `resource` module will restrict the topology XML gathered to the current process CPU affinity mask. Since the job shell fetches this XML for its own use, this then applies that same restriction to all jobs, since the job shell will use this XML to calculate the affinity mask it should use for the job.

Thus, if the system instance broker is pinned to just a few cores, all jobs will inherit this affinity by default.

Note that the job shell inherits the broker's cpu affinity mask by virtue of fork/exec or `posix_spawn(3)`. So it also should not restrict the topology to its current affinity, but instead should use the instance-provided XML as its guide as to which resources are actually available within the instance.

grondo:
Ok, I fixed the doc typo and will set MWP if the above general description of the problem solution sounds good?

garlick:
Yes that makes sense and is more or less what I got from reading your code/comments so all good here!

grondo:
Thanks! I'll set MWP.

PR #4539, opened 2022-09-01
flux-mini: fix bulksubmit help message

jameshcorbett:
Problem: 'flux mini bulksubmit' does not describe itself
accurately when passed the '--help' option.

Fix the help message to more accurately describe usage.

### DIFF
--- a/src/cmd/flux-mini.py
+++ b/src/cmd/flux-mini.py
@@ -1374,6 +1374,8 @@ class BulkSubmitCmd(SubmitBulkCmd):
         for xargs in commands:
             total += len(self.cc_list(xargs))
 
+        if total == 0:
+            raise ValueError("no jobs provided for bulk submission")
         #  Initialize progress bar if requested:
         if args.progress:
             if not args.dry_run:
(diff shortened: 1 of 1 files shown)


src/cmd/flux-mini.py
```diff
@@ -1374,6 +1374,8 @@ def main(self, args):
         for xargs in commands:
             total += len(self.cc_list(xargs))
 
+        if total == 0:
+            raise ValueError("no jobs provided for bulk submission")
```
jameshcorbett, line 1378: Without this check, when `total == 0`, a ZeroDivisionError is raised when printing the progress bar.

chu11: hey james, it seems like there are three fixes in this one commit.  Could they be split up?  AFAICT, there's a typo fix, there's the `total == 0` fix, and the help message fix.

jameshcorbett: I thought someone might comment on that. I made it two commits, because I think the typo falls under the help message fix (it's part of the bulksubmit help message)

chu11 (approved):
LGTM

PR #4541, opened 2022-09-02
shell: inherit FLUX_F58_FORCE_ASCII from job environment

### LINKED ISSUE #4540: FLUX_F58_FORCE_ASCII=1 does not change f in output file name
One of our users on tioga is not getting the "fancy f" in his jobids:

> So, I know that first char in the jobid (tioga) should be 'f' but my terminal is printing that weird char.  Have set TERM to a few different things to try and have done an stty sane.  Any other settings or env vars or things that anyone knows of to fix this?
> was able to flux job cancel by replacing that 'Æ' with an 'f'.  But annoying
> Also shows up in output  files from batch jobs like so:
> flux-ÆyGKufZ6VM.out

I pointed him at setting FLUX_F58_FORCE_ASCII=1 to work around this, and it is making his jobids start with 'f' in the terminal, but he is still getting 'Æ' in the FLUID in his default output file names:

> I am using that now, and the stdout is fixed, but the 'fancy f' is sill showing up in filenames created by flux, such that I can't cut-n-paste them quickly without accounting for that 'fancy f' 
> here is the example.  And do you know  a solution for this?
> flux mini batch -N1 -n4 ./tioga.sh
> f2B2uSNXMmR
> ls -la *out
> flux-Æ2B2uSNXMmR.out
> A cut-n-paste of that out file to 'tail' or 'cat' for instance fails with
> cat flux-Æ2B2uSNXMmR.out
> cat: 'flux-'$'\306''2B2uSNXMmR.out': No such file or directory

It seems like the FORCE_ASCII env variable should affect the jobid in file names. Also, if you have any suggestions for fixing the  'Æ', those would be welcome.

grondo:
As noted in #4540, setting `FLUX_F58_FORCE_ASCII=1` for a job does not change the encoding of a job output file with `{{id}}` in it. This is because the job shell isn't using the job environment, but the environment of the broker or the that created by the IMP (in multi-user mode).

This PR sets up the job shell to inherit a static list of environment variables from the job environment, starting with `FLUX_F58_FORCE_ASCII`.

Note that this can be fixed in current Flux versions by an addition to the shell `initrc.lua` or `lua.d` (see #4540 for the example). However, it seemed like some environment variables should always be inherited by the shell, so thus this PR.

Fixes #4540

### DIFF
--- a/src/shell/shell.c
+++ b/src/shell/shell.c
@@ -1288,6 +1288,25 @@ static int shell_register_event_context (flux_shell_t *shell)
     return 0;
 }
 
+/*  Export a static list of environment variables from the job environment
+ *   to the current shell environment. This is important for variables like
+ *   FLUX_F58_FORCE_ASCII which should influence some shell behavior.
+ */
+static int shell_export_environment_from_job (flux_shell_t *shell)
+{
+    const char *vars[] = {
+        "FLUX_F58_FORCE_ASCII",
+        NULL,
+    };
+
+    for (int i = 0; vars[i] != NULL; i++) {
+        const char *val = flux_shell_getenv (shell, vars[i]);
+        if (val && setenv (vars[i], val, 1) < 0)
+            return shell_log_errno ("setenv (%s)", vars[i]);
+    }
+    return 0;
+}
+
 int main (int argc, char *argv[])
 {
     flux_shell_t shell;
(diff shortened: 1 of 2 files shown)


garlick (approved):
LGTM!

grondo:
Thanks! Setting MWP.

Issue #4530, opened 2022-08-30
display nnodes for pending jobs in 'flux jobs'


ryanday36:
'flux jobs' currently only displays 'nnodes' for running and completed jobs, but that information is important to understanding when a pending job might run, especially on node scheduled clusters. E.g. the following two jobs are very different, but appear to the the same in the `flux jobs` output:
```
[day36@corona211:~]$ flux mini batch -N96 -n96 -t 1d --wrap hostname
ƒ86Z4duezHM
[day36@corona211:~]$ flux mini batch -N2 -n96 -t 1d --wrap hostname
ƒ86Z8NyRqPd
[day36@corona211:~]$ flux jobs
       JOBID USER     NAME       ST NTASKS NNODES  RUNTIME NODELIST
 ƒ86Z4duezHM day36    hostname   PD     96      -        - -
 ƒ86Z8NyRqPd day36    hostname   PD     96      -        - -
 ƒ86Nznv7K6w day36    flux        R      1      1    22.3m corona177
[day36@corona211:~]$
```
Would it be possible to display the requested NNODES for pending jobs?

grondo:
One problem is you can't tell how many nodes a user is requesting from jobspec alone, i.e. they may have only requested cores, and it will be up to the scheduler to decide how many nodes that is. 
However, maybe we can make an estimate, or some other resources count that will have the desired effect? Also, in some cases the number of nodes will be known..

chu11:
if the user specifies a number of nodes (`flux mini -NX`) I think the jobspec will always specify a number nodes.  I suppose in that case we could always list `nnodes` for that job.

The only question for me is if it is acceptable for some pending jobs to have `nnodes` listed and others not.  I feel it should be ok.  If the user didn't require a node count, then they are trusting the scheduler to do what is best.  This differs from slurm, but perhaps users would learn over time this is the expected behavior?

Another thought is if this should be a configurable option, dunno if we want this to be default behavior in job-list.

> However, maybe we can make an estimate, or some other resources count that will have the desired effect?

Would it be possible to tell the scheduler to "pretend" schedule the job and return an estimated resources list?  like, ask the scheduler to hypothetically schedule this job if all resources were available, what would you do?  I imagine this could be useful in other ways too.

grondo:
[...]

PR #4542, opened 2022-09-02
job-list: return nnodes if jobspec specifies nodes

chu11:
Problem: job-list does not return nnodes until a job is running.
When a job specifically requests a number of nodes, it may be
useful for job-list to return nnodes so it can be listed in
job listing tools such as flux-jobs.

Solution: If the jobspec specifically lists nodes as a requested
resource, have job-list return nnodes even before it is running.

Fixes https://github.com/flux-framework/flux-core/issues/4530

### DIFF
--- a/src/modules/job-list/job_util.c
+++ b/src/modules/job-list/job_util.c
@@ -91,9 +91,9 @@ static int store_attr (struct job *job,
         val = json_integer (job->ntasks);
     }
     else if (!strcmp (attr, "nnodes")) {
-        /* job->nnodes potentially < 0 if R invalid */
-        if (!(job->states_mask & FLUX_JOB_STATE_RUN)
-            || job->nnodes < 0)
+        /* job->nnodes < 0 if not set yet or R invalid, may be set in
+         * DEPEND or RUN state */
+        if (job->nnodes < 0)
             return 0;
         val = json_integer (job->nnodes);
     }
--- a/src/modules/job-list/job_state.c
+++ b/src/modules/job-list/job_state.c
@@ -376,8 +376,7 @@ static int parse_res_level (struct list_ctx *ctx,
 
 /* Return basename of path if there is a '/' in path.  Otherwise return
  * full path */
-static const char *
-parse_job_name (const char *path)
+static const char *parse_job_name (const char *path)
 {
     char *p = strrchr (path, '/');
     if (p) {
--- a/src/modules/job-list/job_state.c
+++ b/src/modules/job-list/job_state.c
@@ -397,6 +397,7 @@ static int jobspec_parse (struct list_ctx *ctx,
     json_error_t error;
     json_t *jobspec = NULL;
     json_t *tasks, *resources, *command, *jobspec_job = NULL;
+    struct res_level res[3];
     int rc = -1;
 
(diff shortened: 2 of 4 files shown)


chu11:
oh yeah, in the tests I do

```
	id1=$(flux mini submit hostname) &&
	id2=$(flux mini submit -N1 hostname) &&
	id3=$(flux mini submit -N3 hostname) &&
```

to generate pending jobs that should and should not have nodes specifically specified.  I'm not sure if this is a "safe" way to test.  First the first job above, I suppose the default could change some day, or perhaps its depends on a system wide config.

grondo:
> Edit: just thought of this, would it better to hand create a jobpsec to use for this test?

That is a good idea. Or, maybe just leave out the test for "nnodes not specified". This isn't something the PR "fixes" (i.e. were there tests for that before? I didn't check the changes yet), so why test for it? In the future if someone adds a fix or feature (e.g. to display a nnodes estimate), they'll not only have to add tests for their feature, but remove the previous tests.

chu11:
> That is a good idea. Or, maybe just leave out the test for "nnodes not specified". This isn't something the PR "fixes" (i.e. were there tests for that before? I didn't check the changes yet), so why test for it?

Ahh good point, the "nnodes not specified" is basically already covered.

chu11:
re-pushed, tweaking tests per @grondo comment above

src/modules/job-list/job_util.c
```diff
@@ -91,9 +91,20 @@ static int store_attr (struct job *job,
...
+             * resource specifically requested by user and is listed
+             * in jobspec.
+             */
+            if ((job->states_mask & FLUX_JOB_STATE_DEPEND)
+                && job->nnodes > 0) {
+                val = json_integer (job->nnodes);
+                goto out;
+            }
+            return 0;
+        }
```
grondo, line 105: This should probably be obvious, but why is this extra block needed? Why can't you just always return `nnodes` if it is `> 0`, now that it may be `> 0` even before `RUN` state?

chu11: So right now `nnodes` is re-calculated after it starts running, so `nnodes` can change if something goes wrong during the run.  Not sure if that is wise or not.

If `nnodes > 0` should `nnodes` not be re-calculated?  If that's probably for the best, then the above can just become `if nnodes > 0`, as you suggest.

grondo: Ah, sorry, maybe this is because I don't know the code well enough.

I thought store_attr() was called from a job-list request to create the JSON response payload for each struct job *, and that if ``nodeschanges, thenjob->nnodes` would be updated, so you could always just grab `job->nnodes` and always get the "updated" value.

I think it is good that nnodes is getting updated at the RUN state.

chu11: oh wait, you're right ... the check shouldn't matter, b/c the job state is irrelevant, `nnodes` is updated the regardless.  The check for job state is simply to see if the data is available / relevant.  doh!

chu11:
re-pushed, simplifying logic per discussion above

grondo (approved):
LGTM

chu11:
hmmm second time recently that mergifyio has gotten stuck ...

image

rebased and re-pushed to kick things

PR #4543, opened 2022-09-02
testsuite: fix non-bourne shell test failure

chu11:
Problem: Test in t2802-uri-cmd failed because of quoting
issues with non-bourne shells.  Test was recently broken due to
change in:

testsuite: make sure test ssh runs $SHELL

c6a760ce61de5c3228d0a452a77b52eb0784558c

Solution: Set SHELL to /bin/sh in test.


grondo (approved):
Thanks!

Issue #4533, opened 2022-08-31
need equivalent of --ntasks-per-node


grondo:
I don't believe there is an easy way to request a number of nodes and set the number of tasks to run on those nodes outside of the shell-provided backdoor `-o per-resource.type=node -o per-resource.count=N`. If it is possible to make this request with jobspec v1, we should consider how to add this option to `flux mini run`. If not, then maybe hiding the shell option w/a behind such a `flux mini run` option would be better than telling users to set those options directly.

trws:
I can't remember entirely how our `per-resource` support is currently implemented, but jobspec V1 includes support for `per_slot` in `count` rather than `total`, so we could construct a `nodes[count]>slot` resource spec and then use `count:per_slot:tasks_per_node` to get there.  We could probably also do the math locally as long as we require the number of nodes to be explicit, and only allow exclusive nodes with that option, then `count:total:count * tasks_per_node` should do it.

grondo:
> I can't remember entirely how our per-resource support is currently implemented,

Sadly, this option causes the job shell to completely ignore the `tasks` section of jobspec and do exactly what the option states, run N tasks per resource type noted in the option (only nodes, cores supported at this time)

> so we could construct a nodes[count]>slot resource spec and then use count:per_slot:tasks_per_node to get there.

I'm probably wrong, but [RFC 25](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_25.html#v1-specific-resource-graph-restrictions) lists the only allowed specs as:

 * slot>core
 * node>slot>core
 * slot>(core,gpu)
 * node>slot>(core,gpu)

i.e. `slot` always has to contain a `core` and/or `gpu`. If we could relax this constraint, then I can see how `node>slot` could be used with `count:per_slot` (note that the job shell currently does not support `count:per_slot` afaik, so that would need to be addressed as well). I'm not sure why we limited the spec as it is now though...

trws:
Ah, then the easiest is probably to make `--ntasks-per-node` generate the `per-resource` form with `resource.type=node` and just prevent them from supplying other task options.  Would that work or cause other issues do you think?

grondo:
That would work, and is a good solution since we can change how it is implemented in jobspec without breaking the "interface"

grondo:
> and just prevent them from supplying other task options. 

[...]

PR #4544, opened 2022-09-06
cmd: add "per-resource" allocation options to flux-mini run and submit

grondo:
This is a request for comments PR (and a WIP) since I'm not sure if this will be an acceptable way to solve #4533. I first made an attempt to add `--tasks-per-node` support to the base `flux mini run,submit` commands and `JobspecV1.from_command()`, but I fear "that way madness lies" and I eventually gave up. Trying to get consistent and expected behavior when trying to mix the existing `--ntasks`, `--nodes`, `--cores-per-task` and new `--tasks-per-node` and `--cores` options was too much for me.

Instead, the approach taken here is to add a new `JobspecV1.per_node()` constructor:
```python
    def per_node(
        cls,
        command,
        ncores=None,
        nnodes=None,
        tasks_per_node=None,
        tasks_per_core=None,
        gpus_per_node=None,
        exclusive=False,
    ):
        """
        Factory function that builds a v1 jobspec from an explicit count
        of nodes or cores and a number of tasks per one of these resources.

        Use setters to assign additional properties.

        Args:
            ncores: Total number of cores to allocate
            nnodes: Total number of nodes to allocate
            tasks_per_node: Force the number of tasks per node
            tasks_per_core: Force > 1 tasks per core
            gpus_per_node: With nnodes, request a number of gpus per node
            exclusive: with nnodes, request whole nodes exclusively
        """
```

This function uses the shell `per-resource` option to satisfy the user request in most cases, but note that it bails out instead of trying to solve many corner cases, since it seemed best at this time to solve only the obvious cases for now.

If any of the new options `--cores`, `--tasks-per-node`, `--tasks-per-core`, `--gpus-per-node` are used, then `flux mini run` and `flux mini submit` use `JobspecV1.per_node()` to construct the jobspec, instead of `JobspecV1.from_command()`. These new options are split off in the `--help` now to make it clear they operate in a different "mode":
```
Common resource options:
  -N, --nodes=N             Number of nodes to allocate
      --exclusive           With -N, --nodes, allocate nodes exclusively

Per task options:
  The following options allow per-task specification of resources, and
  should not be combined with per-resource options.

  -n, --ntasks=N            Number of tasks to start
  -c, --cores-per-task=N    Number of cores to allocate per task
  -g, --gpus-per-task=N     Number of GPUs to allocate per task

Per resource options:
  The following options allow per-resource specification of tasks, and
  should not be used with per-task options above

      --cores=N             Request a total number of cores
      --tasks-per-node=N    Force number of tasks per node
      --tasks-per-core=N    Force number of tasks per core
      --gpus-per-node=N     Request a number of GPUs per node. Only with
                            --nnodes, --tasks-per-node
```

This allows us to avoid changing the `from_command()` implementation, and only invoke the new behavior when the new command options are used. This should make maintenance of the two "modes" a little easier going forward.

This satisfies the use case of explicitly specify a number of tasks per node without specifying the exact number of total nodes, while also allowing some other cases (like oversubscribing tasks to cores), however, it seems like it might be a dangerous step extending our "minimal" interface in this way. However, I'm unsure how to do a better job at this time. Thus, before I go any further adding more testing and documentation, I'm pausing to get some feedback.

### DIFF
--- a/src/bindings/python/flux/job/Jobspec.py
+++ b/src/bindings/python/flux/job/Jobspec.py
@@ -648,6 +648,124 @@ class JobspecV1(Jobspec):
             raise ValueError("attributes.system.duration must be a number")
 
     @classmethod
+    # pylint: disable=too-many-branches,too-many-locals,too-many-statements
+    def per_resource(
+        cls,
+        command,
+        ncores=None,
+        nnodes=None,
+        per_resource_type=None,
+        per_resource_count=None,
+        gpus_per_node=None,
+        exclusive=False,
+    ):
+        """
+        Factory function that builds a v1 jobspec from an explicit count
+        of nodes or cores and a number of tasks per one of these resources.
+
+        Use setters to assign additional properties.
+
+        Args:
+            ncores: Total number of cores to allocate
+            nnodes: Total number of nodes to allocate
+            per_resource_type: (optional) Type of resource over which to
+                               schedule a count of tasks. Only "node" or
+                               "core" are currently supported.
+            per_resource_count: (optional) Count of tasks per
+                                `per_resource_type`
+            gpus_per_node: With nnodes, request a number of gpus per node
+            exclusive: with nnodes, request whole nodes exclusively
+        """
+
+        #  Handle per-resource specification:
+        #  It is an error to specify one of per_resource_{type,count} and
+        #   not the other:
+        #
+        per_resource = None
+        if per_resource_type is not None and per_resource_count is not None:
+            if not isinstance(per_resource_type, str):
+                raise ValueError("per_resource_type must be a string")
+            if per_resource_type not in ("node", "core"):
+                raise ValueError(
+                    f"Invalid per_resource_type='{per_resource_type}' specified"
+                )
+            if not isinstance(per_resource_count, int):
+                raise ValueError("per_resource_count must be an integer")
+            if per_resource_count < 1:
+                raise ValueError("per_resource_count must be >= 1")
+
+            per_resource = {"type": per_resource_type, "count": per_resource_count}
+        elif per_resource_type is not None:
+            raise ValueError("must specify a per_resource_count with per_resource_type")
+        elif per_resource_count is not None:
+            raise ValueError("must specify a per_resource_type with per_resource_count")
+
+        if ncores is not None:
+            if not isinstance(ncores, int) or ncores < 1:
+                raise ValueError("ncores must be an integer >= 1")
+        if gpus_per_node is not None:
+            if not isinstance(gpus_per_node, int) or gpus_per_node < 0:
+                raise ValueError("gpus_per_node must be an integer >= 0")
+            if not nnodes:
+                raise ValueError("gpus_per_node must be specified with nnodes")
+        if nnodes is not None:
+            if not isinstance(nnodes, int) or nnodes < 1:
+                raise ValueError("nnodes must be an integer >= 1")
+        elif exclusive:
+            raise ValueError("exclusive can only be set with a node count")
+
+        nslots = None
+        slot_size = 1
+        if nnodes and ncores:
+            #  Request ncores across nnodes, actually running a given
+            #   number of tasks per node or core
+            if ncores < nnodes:
+                raise ValueError("number of cores cannot be less than nnodes")
+            if ncores % nnodes != 0:
+                raise ValueError(
+                    "number of cores must be evenly divisible by node count"
+                )
+            #
+            #  With nnodes, nslots is slots/node (total_slots=slots*nodes)
+            nslots = 1
+            slot_size = int(ncores / nnodes)
+        elif ncores:
+            #  Request ncores total, actually running a given
+            #   number of tasks per node or core
+            #
+            #  Without nnodes, nslots is total number of slots:
+            nslots = ncores
+            slot_size = 1
+        elif nnodes:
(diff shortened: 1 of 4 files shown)


garlick:
The approach seems sane to me, in fact I can't imagine doing it any better given the constraints we are dealing with.  The help output makes it pretty clear how these should be used.  Hopefully error messages are similarly clear (I didn't look) since I imagine people hacking on batch scripts do a fair bit of cutting and pasting without reading documentation.

Would it improve clarity if the new options had a common prefix?   I was thinking `--force-tasks-per-node=N` and `--force-tasks-per-core=N` but I guess that doesn't really work for the others...  Hmm.

src/bindings/python/flux/job/Jobspec.py
```diff
@@ -647,6 +647,123 @@ def _v1_validate(resources, tasks, kwargs):
         if not isinstance(system["duration"], numbers.Number):
             raise ValueError("attributes.system.duration must be a number")
 
+    @classmethod
+    # pylint: disable=too-many-branches,too-many-locals,too-many-statements
+    def per_node(
```
trws, line 652: Maybe per_resource? Since this can do per_{core,node}?

trws:
> Would it improve clarity if the new options had a common prefix? I was thinking `--force-tasks-per-node=N` and `--force-tasks-per-core=N` but I guess that doesn't really work for the others... Hmm.

We could do something like this, but have options that specify which *mode* it's working in that would work a bit like subcommands.  Have the default be something like `--total-resources` and new `--per-resource` (terrible names, need bikeshedding), put them both in a mutually exclusive group so only one can be specified, and gate the options on which mode it's in? That could help make it obvious how they group, and make argparse enforce more invariants for us.

On that note, I like this interface in general, but have thoughts on refactoring for simplicity/readability, especially if I remember correctly that we are requiring python 3.6 these days.  The only real fundamental question I have though is whether we should consider refactoring from the constructor taking both `per-core` and `per-node` to taking `per={node or core}` like how the jobspec works.  We could still have the arguments `--tasks-per-node` etc. for slurm-like comfort, but internally we work on a kinda generic target resource, at least in principle, so it might be worth factoring that way.

trws:
Should say, short version, I have no issues with the command line interface if folks are happy with it, and we could always tweak the constructor interface if we want to after release if that's useful.

src/bindings/python/flux/job/Jobspec.py
grondo: Good point!

grondo:
> We could do something like this, but have options that specify which mode it's working in that would work a bit like subcommands.

I was actually going to say I did think about adding a `--mode=` option (similar to `valgrind`) to force users to specify which mode of the command they want to use. It would actually be easy to subclass `SubmitCmd` to make yet another `flux mini` subcommand, but I couldn't think of a good name for it and it seemed like perhaps it would go a little too far.

In this case, since use of any of the listed options forces the "per-resource mode", I felt like making it easier for the user was probably best, but I could easily be convinced otherwise, especially since we're supposed to be designing a "stable interface" for the flux-mini command set.

> whether we should consider refactoring from the constructor taking both per-core and per-node to taking per={node or core} like how the jobspec works

Ah, yes, since per_node and per_core are mutually exclusive, are you thinking something like:
```python
    def per_resource(
        cls,
        command,
        ncores=None,
        nnodes=None,
        per_resource_type=None,
        per_resource_count=None,
        gpus_per_node=None,
        exclusive=False,
    ):
```

That does seem a bit better

trws:
Yup, that’s what I had in mind. Maybe even merge out nnodes and ncores if they are exclusive, if not then just like that. That way if we add “sockets” or something later it kinda just falls out.







whether we should consider refactoring from the constructor taking both per-core and per-node to taking per={node or core} like how the jobspec works

Ah, yes, since per_node and per_core are mutually exclusive, are you thinking something like:

    def per_resource(
        cls,
        command,
        ncores=None,
        nnodes=None,
        per_resource_type=None,
        per_resource_count=None,
        gpus_per_node=None,
        exclusive=False,
    )

grondo:
I've pushed a couple fixup commits here that

 * Rename `JobspecV1.per_node()` to `JobspecV1.per_resource()`, and rework the constructor function to take `per_resource_type` and `per_resource_count` options instead of the mutually exclusive `tasks_per_node` and `tasks_per_core` arguments. The function was reworked accordingly.
 * `flux mini run,submit` keep the same options, but now use the new `JobspecV1.per_resource()` function. Also some other fixes and cleanup were added here.


Tests are unchanged and still pass, for what little that means. I'll assume only minor changes will be required on further review, so I'll update the documentation and start adding a few more tests. Perhaps we can get this into our release today.

grondo:
I've force-pushed the new versions of `JobspecV1.per_resource()` and `flux-mini.py`. This is ready for a full review  so I've removed the RFC and WIP prefix.

garlick (approved):
Looks great to me!  Nice work on the tests and the docs.

grondo:
@trws: this is where we ended up. The `flux mini run,submit` options are the same, but there is now a `JobspecV1.per_resource()` constructor used by `flux-mini`. That did simplify things a bit so thanks for that suggestion. Want to give a final ACK before I set MWP?

grondo:
Thanks! In order to tag the release I'll set MWP now and we can open issues on any problems found by anyone.

trws:
LGTM, thanks @grondo

Issue #3744, opened 2021-06-25
design shell plugin for file broadcast


garlick:
Following up on a coffee time discussion with @jameshcorbett 

There is a need for an C API for copying file(s) to a job.  (What other requirements are there?)

Some notes from the discussion
* a shell plugin could implement a distributed file copy service
* plugins can register job-specific service "methods"
* plugins can also get plugin specific options from the jobspec attributes (which can be set on `flux mini` command line)
* plugins can address RPCs to shell ranks within the job
* see the PMI plugin for an example of a virtual TBON (in this case to gather PMI keys)
* the job specific service name can be found in the job eventlog (see MPIR plugin for an example)
* files should be sent in chunks (say 4K) to avoid head of line blocking in the broker
* it may be handy to use the per-job tmpdir established by the tmpdir job plugin (FLUX_JOB_TMPDIR environment variable should be set in the job)

Additional thoughts:
* `src/common/libutil/kary.h` provides some helper function for determining virtual TBON peers etc.
* Use a streaming RPC to send the chunks to the destination.
* Note that the shells are only running while the job is running, so this wouldn't work for stage-in while the job is still pending
* The service could actually be the same on each shell rank, with the semantics of "copy to local file and TBON subtree".

garlick:
Couple of more thoughts:
* Need to handle open errors on remote files (e.g. file exists)
* Need to handle write errors to remote files (e.g. file system full).
* Errors should be propagated back to the API
* A `flux_future_t` based interface is desirable to enable reactive programming
* May want a way to pass open flags into the API (e.g. O_TRUNC)
* API should have a way to select a subset of job ranks as targets
* API should have a way to select any destination path (e.g. escape FLUX_JOB_TMPDIR if desired)
* Idea: option to let service select the destination path and set a configurable environment variable to point to it

I hate to even mention this because I really like the  idea of having this capability in Flux, but I think this problem could also be solved with  [mpifileutils](https://github.com/hpc/mpifileutils) or similar with  "alloc bypass" from #3740, copying R from the target job.  Two advantages of that approach are 1) uses RDMA, and 2) portability.

garlick:
[...]

Issue #3631, opened 2021-05-03
CTI support: shipping files to a running job


ardangelo:
To support running a tool alongside a job, CTI will ship over the tool's binaries and other support files to the nodes assigned to the user's job. Then, the tool binaries will be started on the nodes and start analysis on the job.

As discussed in https://github.com/flux-framework/flux-core/issues/3291, Flux would need a plugin to support shipping files after a job has been started.

Furthermore, providing a job-specific temporary directory that is cleaned up upon job exit would be very useful in this context, to be able to ship files to a known path on nodes that are reliably cleaned up.

jameshcorbett:
Status of CTI, from @ardangelo :

> We have implemented all the rest of CTI functionality using the Flux API, but are still relying on SSH for the file transfer functionality.
> While this works in testing, it requires SSH access to compute nodes. On some of our internal systems, and nearly all customer systems, SSH access to nodes is restricted or disabled entirely for normal users.
> So while we can verify our Flux implementation for CTI, we won't be able to actually use it on systems where SSH access to nodes is restricted.
> Implementing runtime file shipping in Flux will remove the SSH requirement for CTI's Flux implementation.

Similarly from an earlier email:

> CTI shipped with preliminary Flux support in PE 21.12, our December release.
> As is, it still requires getting SSH keys set up to ship files to nodes, we are just waiting on support for file shipping for running jobs in the Flux API.
> Once we have that, we'll be able to strip out the SSH key requirement and have Flux fully implemented in CTI.
> You should still be able to use CTI with Flux for now, but you would need passwordless SSH key access set up to compute nodes for the running job to ship files out correctly.

Motivation for the request, from @ardangelo :

> Regarding the file transfers: the point is to make files accessible to the job and its tool daemons... Some of the systems we run on do not have any form of cross-mounted storage to share data between the login / compute nodes, and file shipping allows us to run on these systems. Even on systems that do have shared storage, we don't want to have to run a heuristic to detect the location every time. So, shipping support files to job-specific temporary directories is the way we've decided to go.

I think a function like this would satisfy the requirements.

```C

/*
 * Broadcast the contents of file descriptor 'fd' to a Flux job given by 'id'.
 * The contents of the FD will be written to 'destpath' on every compute node
 * in the job. 
 *  
 * If 'destpath' is a relative path, it will be interpreted as relative to the job's
 * temporary directory (given by the FLUX_JOB_TMPDIR environment variable).
 *
[...]

PR #4546, opened 2022-09-07
WIP: broker: add ability to mmap files into content cache

garlick:
I'm parking this weekend experiment pending some time to test scalability.

This adds a feature I had been thinking about for some time - support for mmapping a file into the broker content cache on rank 0, and reading it through the cache on the other broker ranks of the instance, leveraging the tree based overlay network fanout for scalability.

For preliminary tooling I added
- `flux content mmap FILENAME [BLOBSIZE]` which maps a file and lists blobrefs for chunks on stdout
- `flux content mumnap FILE` which unmaps the file
- a new  `flux content load` mode which reads blobrefs on stdin and contcatenates the blobs on stdout

With this one can do something like:
```
ƒ(s=64,d=0,builddir) $ flux content mmap 4g.img | flux exec sh -c "flux content load | md5sum"
[snip]
ƒ(s=64,d=0,builddir) $ flux content munmap 4g.img
```
In that command, the list of blobrefs is being distributed to the stdin of the remote commands through flux exec, then the file contents is being loaded through the cache.

On the relevance to the file broadcast tool being discussed in #3744 and #3631, I'm not sure if this is a direct fit because
- this service is only available to the instance owner
- the source file must be on broker rank 0
- the source must be a file (it cannot be a stream)

If we did want to go forward with such a tool though, I was thinking it could be made standalone so it could map the file, perform the copy by using the broker subprocess exec service directly, and then unmap the file when done.  In the above example, the file has to be manually unmapped after the command completes.

Mostly though, in this PR, I just wanted to see if we could implement the basic capability since it's been on the back burner for so long, and then we'll at least have something to compare other ideas to.  Sometimes these demons must be exorcised.

### DIFF
--- a/src/broker/Makefile.am
+++ b/src/broker/Makefile.am
@@ -42,6 +42,8 @@ libbroker_la_SOURCES = \
 	content-cache.c \
 	content-checkpoint.h \
 	content-checkpoint.c \
+	content-mmap.h \
+	content-mmap.c \
 	runat.h \
 	runat.c \
 	state_machine.h \
--- a/src/broker/content-cache.c
+++ b/src/broker/content-cache.c
@@ -60,8 +60,9 @@ struct msgstack {
 };
 
 struct cache_entry {
-    void *data;
+    const void *data;
     int len;
+    void *data_container;
     void *hash;                     // key storage is contiguous with struct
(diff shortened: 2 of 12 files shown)


garlick:
Reworked PR coming soon, closing this one.

PR #4548, opened 2022-09-07
doc: document {id.dec} in flux-jobs(1)

chu11:
Problem: The output format of {id.dec} was not documented in flux-jobs(1).

Add it to the list of id output formats.


garlick (approved):
Groovy.

Issue #4436, opened 2022-07-26
add job notification


grondo:
This issue tracks the development of support for job notification (e.g. email on job events)  for Flux jobs as required for the rollout of Flux as the native resource manager on production systems.

 - [ ] #4435
 - [ ] Implement job notification system as designed

PR #4555, opened 2022-09-08
job-list: handle per-resource ntasks special case

chu11:
Problem: Under several special cases to work around Jobspec V1
limitations, the shell per-resource options are used to specify
number of tasks in a job.  This case was not handled by job-list.

Solution: Handle this special case and set ntasks accordingly.

### DIFF
--- a/src/modules/job-list/job_data.c
+++ b/src/modules/job-list/job_data.c
@@ -19,6 +19,7 @@
 
 #include "src/common/libczmqcontainers/czmq_containers.h"
 #include "src/common/librlist/rlist.h"
+#include "src/common/libccan/ccan/str/str.h"
 
 #include "job_data.h"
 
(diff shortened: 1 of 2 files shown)


src/modules/job-list/job_state.c
```diff
@@ -390,6 +391,34 @@ static const char *parse_job_name (const char *path)
...
+                        "count",
+                        count) < 0) {
+        flux_log (ctx->h, LOG_ERR,
+                  "%s: job %ju invalid jobspec: %s",
+                  __FUNCTION__, (uintmax_t)job->id, error.text);
+        return -1;
+    }
+
+    return 0;
+}
```
garlick, line 420: It feels like this is the sort of code that should be in a separate source module with unit tests.

Have you thought about whether it would be feasible to abstract the code that parses jobspec and R into a `struct job` into a  standalone and potentially reusable class or set of classes?

chu11: i've thought about it on and off.  Lemme consider it more deeply with this PR and also considering #4336

chu11: i already went down the path to do some split out in #4436, so perhaps its a good idea to try and do a split out now rather than later.  I'll try to do that in a separate PR.

chu11:
rebased on PR #4575

chu11:
rebased & re-pushed b/c #4575 put merged

grondo (approved):
This looks simple enough. We should get an issue open to encapsulate @garlick's suggestion to refactor all the R and jobspec parsing into a testable module which wasn't completed here nor in the previous refactoring PR, so we don't lose track of it.

PR #4559, opened 2022-09-08
python: uri: use path to current flux executable in lsf resovler

### LINKED ISSUE #4558: python: uri resolvers that `ssh flux uri` should pass the full path to `flux`
Problem: When trying to use a resolver from a side installed, spack built, or builddir flux-core, resolvers (such as the `lsf` resolver) may run `ssh HOST flux uri <resolver>?args`, but since `ssh` does not forward the environment, this runs the system installed flux. If the system installed flux doesn't support the resolver scheme, then the command fails.

grondo:
Problem: The lsf URI resolver plugin executes `ssh flux uri resolve..` on a compute host during URI resolution, but this may not work if the remote host has a different version of Flux installed as the system instance.

Assume a shared filesystem by default and replace `flux` with the path to the local flux executable by using `shutil.which("flux")`. As with the ssh connector, allow this to be overridden with `FLUX_SSH_RCMD`.

Fixes #4558

### DIFF
--- a/src/bindings/python/flux/uri/resolvers/lsf.py
+++ b/src/bindings/python/flux/uri/resolvers/lsf.py
@@ -19,6 +19,7 @@ It can be used like:
 import getpass
 import os
 import subprocess
+import shutil
 from pathlib import PurePath
 
 import yaml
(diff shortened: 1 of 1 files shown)


garlick (approved):
Makes sense.  LGTM.
(I did not test this one).

PR #4560, opened 2022-09-09
improve signal/noise ratio in systemd journal logs

garlick:
Problem: there is a lot of garbage to sift through when trying to do a post-mortem on a flux problem by looking at `journalctl -u flux` on the rank 0 node.

Eliminate, rate-limit, and reduce severity of several offending messages.


grondo (approved):
LGTM!

PR #4561, opened 2022-09-09
spack: add flux-core container build

vsoch:
Not multi-stage build yet because we do not have copy view, only symlink! I tested running a piece of the tutorial in here and the basics worked!

### DIFF
--- a/.github/workflows/build-container.yaml
+++ b/.github/workflows/build-container.yaml
@@ -1,4 +1,4 @@
-name: Build and Deploy Container
+name: Build and Deploy Containers
 
 on:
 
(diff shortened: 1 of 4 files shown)


grondo:
Thanks! Before merging the branch will need to be rebased to avoid the merge commit and please squash the two commits with identical messages. 

Also, for consistency we try to use the imperative for commit messages "add" instead of "adding" for example. See [RFC 1 Patch Requirements ](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_1.html#patch-requirements)

vsoch:
Gotcha! It’s not ready for review yet but I’ll fix it up when it is.

vsoch:
Okay this is ready for review! And I think I did the commit right this time. It looks like they are some "warning: this will deprecate with spack 0.0.19" - specifically the format for the concretizer in the spack.yaml, but we're not there yet, so my preference is to maintain the current convention (which we know works) until it's forced.

grondo (approved):
LGTM! You are the spack expert so we trust your judgement there!

I'm going to tweak the PR title since it will be used in the release notes, then set the MWP flag. Thanks!

PR #4566, opened 2022-09-13
resource: support configuration of resources in TOML config

grondo:
After lamenting that configuration of resources via a hand edited _R_ or via use of `flux R encode` was confusing and error prone, @garlick and I had the idea that a simple configuration of resources in the `[resource]` TOML table might be a huge improvement.

This PR introduces support for a `resource.config` array which can be used to configure resources instead of `resource.path` pointing to the system _R_. The format of each entry in `resource.config` is a required `hosts` key pointing to a hostlist of node names to which the entry applies, then optional `cores`, `gpus` (in idset form), and `properties` (an array of properties to assign to `hosts`). It is not an error to specify a hostname more than once - in that case the host is updated not overwritten. E.g.:

```toml
[[resource.config]]
hosts = "corona[171-207,211-296]"
cores = "0-47"
gpus =  "0-7" 

[[resource.config]]
hosts = "corona[171-207,211-280]"
properties = ["batch"]

[[resource.config]]
hosts = "corona[281-296]"
properties = ["debug"]
```

The main benefit of this format is that the admin does not have to worry about ranks, which would be reassigned by the `resource` module based on the `[bootstrap]` hostlist anyway, and can focus on assigning resources and properties to hosts by name.

### DIFF
--- a/src/common/librlist/rnode.c
+++ b/src/common/librlist/rnode.c
@@ -187,6 +187,7 @@ fail:
 
 int rnode_add (struct rnode *orig, struct rnode *n)
 {
+    int rc = 0;
     struct rnode_child *c;
     if (!orig || !n)
         return -1;
@@ -197,7 +198,21 @@ int rnode_add (struct rnode *orig, struct rnode *n)
             return -1;
         c = zhashx_next (n->children);
     }
-    return 0;
+    if (n->properties) {
+        zlistx_t *l = zhashx_keys (n->properties);
+        if (l != NULL) {
+            const char *property = zlistx_first (l);
+            while (property) {
+                if (rnode_set_property (orig, property) < 0)
+                    rc = -1;
+                property = zlistx_next (l);
+            }
+            zlistx_destroy (&l);
+        }
+        else
+            rc = -1;
+    }
+    return rc;
 }
 
 struct rnode *rnode_create (const char *name, uint32_t rank, const char *ids)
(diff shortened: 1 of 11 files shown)


src/common/librlist/rnode.c
```diff
@@ -197,7 +198,17 @@ int rnode_add (struct rnode *orig, struct rnode *n)
             return -1;
         c = zhashx_next (n->children);
     }
-    return 0;
+    if (n->properties) {
+        zlistx_t *l = zhashx_keys (n->properties);
```
garlick, line 202: Need to check `zhashx_keys()` for NULL return here.

grondo:
Since @ryanday36 has configured a few systems with the `flux R encode` method, it will be good to get his feedback on this new method to see if anything is missing or could be improved.

garlick (approved):
It should be noted that we still have a way to put JGF in the configured R for advanced scheduling use cases because the `path` config key still allows an external R  object to be configured.

I really like this.  I quickly converted my test cluster to use it and everything worked as expected on the first try (with properties for queues).

Big usability improvement IMHO.

src/common/librlist/rnode.c
grondo: Thanks! I addressed this and force pushed the result.

ryanday36:
I like it! Much more human readable and looks like it would play nicely with templates in ansible. Do hosts need to be mapped to ranks somewhere, or are they implied by the order of the hosts?

garlick:
> Do hosts need to be mapped to ranks somewhere, or are they implied by the order of the hosts?

The hosts are mapped to ranks by the `bootstrap.hosts` array.  That was true before - the ones in R were overridden by `bootstrap.hosts` if the order was different.

grondo:
Ok given @ryanday36's approval I'll set MWP. Thanks all!

Issue #2627, opened 2020-01-11
flux-jobs: Support verbose default option


chu11:
As time goes on, we will learn about more information that users like to get, but just not as much as the default output.  Some possibilities:

- exit code
- priority
- node count
- time limit

chu11:
from #4594

just to get the ball rolling on ideas for `--format=long`, I'll throw this out as a contender

```
flux jobs -a -o '{id.f58:>12} {username:<8.8} {name:<10.10} {state:>8} {result:>9h} {ntasks:>6} {nnodes:>6h} {runtime\!F:>8h} {t_remaining\!F:>12h} {nodelist:h}'
```

basically status_abbrev -> state & result (no abbrev, full names)
add t_remaining

initially added expiration, but felt it had less value.

figured priority wouldn't matter to average user?

TBH, this isn't much more than the default. Although I think `t_remaining` is useful for running jobs.

grondo:
I'd suggest the addition of `t_submit`, perhaps `{t_submit!d:%b%d %R}` (except that padding and alignment are supported for `datetime` formats. Hmm, maybe introduction of a new conversion specifier that converts a datetime to this form would be useful)

Also, I don't find that separating `{status}` into `{state}` and `{result}` adds much. Maybe just replace `{status_abbrev}` with the full `{status}`?

Finally, I would also suggest adding `sched.t_estimate` when it exists. Maybe we could introduce a new format field that returns `t_estimate` for pending jobs or the nodelist for other jobs.

chu11:
> Maybe just replace {status_abbrev} with the full {status}?

Yeah I agree.

> I'd suggest the addition of t_submit

This surprises me, what is the average use case / interest?  Perhaps to try and cancel jobs submitted a long time ago?

> Finally, I would also suggest adding sched.t_estimate when it exists.

Oh yeah, that's a good idea.  We don't list it in`flux-jobs(1)` so totally forgot about it.  Especially once the #4591 goes in so it works with schedulers that don't support it.

grondo:
> This surprises me, what is the average use case / interest? Perhaps to try and cancel jobs submitted a long time ago?

Well otherwise you have no idea how long a job has been sitting in the queue or "been around". You might see 5 jobs in the queue, but not know that one has been there for 5 days. This would be the equivalent to the `ps(1)` `START` column which is displayed with `ps aux`. Given that our model for `flux jobs` was `ps` in some ways, it seems reasonable to include this field.

[...]

PR #4571, opened 2022-09-15
WIP: add configurable named formats to flux-jobs

grondo:
This one has finally gotten to a state where I can post a working PR. I thought I'd post it now as a WIP before moving on to testing, etc. to make sure everyone is ok with the design and other choices (such as the choice of supported config file formats, vendoring the `tomli` library and so on)

This PR adds support for configurable, named formats to `flux jobs`, e.g. `flux jobs --format=deps`. Alternate format strings can be configured internally, via system config file(s), or via user config in `~/.config/flux/flux-jobs.{toml,yaml,json}`. The location of config files follows the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) mainly for lack of a better standard to follow. Although XDG is ostensibly for desktop systems, it does seem other software is starting to adopt this standard set of paths.

In this implementation `flux-jobs` reads all config files found in the system default directory `/etc/xdg/flux/flux-jobs.*`, the `XDG_CONFIG_DIRS` environment variable, user config in `~/.config/flux/flux-jobs.*` and `XDG_CONFIG_HOME` in order of increasing precedence, merging the configuration each time. Thus, a sysadmin can override the default `flux-jobs` output format by configuring a `default` format in system config, and a user could override _that_ in user config.

The config file format is simple here. `flux-jobs` looks for keys in a `formats` table where the key represents the format name. Each entry in the table has a `format` key for the format string and an optional `description` which gives a string displayed if a user uses `flux jobs --format=help`:

```toml
[formats.myformat]
description = "My useful format"
format = "{id.f58:>12} {name:>8.8} {t_submit!D:<19} {t_run!D:<19} {t_remaining!F}
```

```console
$ flux jobs -o help

Configured flux-jobs output formats:

  default      Default flux-jobs format string
  deps         Show job urgency, priority, and dependencies
  myformat     My useful format

$ flux jobs -a -o myformat
       JOBID     NAME T_SUBMIT            T_RUN               T_REMAINING
  ƒ8fn1HSnC7    sleep 2022-09-15T07:10:26 2022-09-15T07:10:26 4.948m
   ƒAGg5hwdZ    sleep 2022-09-14T15:15:11 2022-09-14T15:15:11 0s
  ƒ3jKEaH943    sleep 2022-09-14T20:42:06 2022-09-14T20:42:06 0s
  ƒ3jJU5AhRh    sleep 2022-09-14T20:42:04 2022-09-14T20:42:04 0s
   ƒALZqYaZ5    sleep 2022-09-14T15:15:20 2022-09-14T15:15:20 0s
   ƒAJam8Krf    sleep 2022-09-14T15:15:15 2022-09-14T15:15:15 0s
    ƒDQxNEPR    sleep 2022-09-14T14:55:19 2022-09-14T14:55:19 0s

```


garlick:
Nice!  That will simplify this [faq entry](https://flux-framework.readthedocs.io/en/latest/faqs.html#why-is-my-job-not-running)!

grondo:
> Nice! That will simplify this faq entry!

Thanks, I copied your suggested format entry from there as the first non-default named format.

I force pushed some fixes for LGTM comments. Apparently it isn't safe to call an overridden method from `__init__`, so I had to refactor the `UtilConfig` class to have a `load()` method which actually loads (and optionally validates) configuration.

chu11:
I was reminded of this issue I made eons ago:  https://github.com/flux-framework/flux-core/issues/2627.    Doesn't have to be for this PR, but we could consider for this PR.

grondo:
> I was reminded of this issue I made eons ago: https://github.com/flux-framework/flux-core/issues/2627.

Good idea. I had that same thought to support a builtin "long" or "verbose" format. If the approach taken here so far is good we could add that, or simply add it to a follow on PR.

chu11:
something i was thinking about, if we support a `--format=verbose` output, it's so common for users to expect `--verbose` or `-v` to be an option, we may want to add `--verbose` to be equal to `--format=verbose`.

Soooo ... would we want a `--deps` option?  I'd lean no b/c its not as common an expectation, but just through I'd throw this out there.

grondo:
> if we support a --format=verbose output, it's so common for users to expect --verbose or -v to be an option, we may want to add --verbose to be equal to --format=verbose.

This is why I'd lean toward a `--format=long` option, not `verbose` (plus it is less to type).

As a user I'd expect `--verbose` to make the command itself emit verbose logging message, not change the output format of the command.

grondo:
I've added documentation and some tests here so this is ready for a review.

grondo:
Force pushed a fix for a broken &&-chain in the tests.

grondo:
One issue that still needs to be addressed is that if the default `flux-jobs` format is overridden in the hard-coded system path `/etc/xdg/flux/flux-jobs.*` or in `~/.config/flux/flux-jobs.*`, then this format would be used even in a builddir, which could cause tests to fail. There should be a way to disable inclusion of these directories. Perhaps an environment variable that could be set by sharness.

Additionally, sharness should unset the `XDG_*` environment variables as well, just in case.

doc/man1/flux-jobs.rst
```diff
@@ -356,6 +359,56 @@ include:
...
+=============
+
+The ``flux-jobs`` command supports registration of named output formats
+in configuration files. The command loads configuration files from
+``flux-jobs.EXT`` from the following paths in order of increasing precedence:
+
+ * ``/etc/xdg/flux``
+ * ``$XDG_CONFIG_DIRS/flux``
+ * ``$HOME/.config/flux``
+ * ``XDG_CONFIG_HOME/flux``
```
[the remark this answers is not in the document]
grondo: Thanks for finding that!

Yeah, it makes more sense that the defaults are used only when the XDG variables aren't set, not in addition to the XDG variables like I assumed at first (now I'm scratching my head wondering why I thought that in the first place).

> I wonder just using that package makes sense?

It is tempting with Python to just use available libraries add a `requirements.txt` and have users `pip install` away, but I'm unsure if this is good practice for system software. Though in this case the package is already available on TOSS 4, so perhaps pyxdg is more ubiquitous than other non-standard python libraries. Still it is one more requirement to add, and this case seems pretty trivial (literally it would replace a couple lines of code?) so I admit I'm on the fence. I'll use it if you feel strongly about it.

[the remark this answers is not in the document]
garlick: > I'll use it if you feel strongly about it.

I do not.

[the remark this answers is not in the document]
grondo: Ok, I pushed a fixup commit that I think fix the config directory search path (and also gets the XDG_CONFIG_DIRS in the right order). Let me know what you think and I can squash.

I also added a commit that sets the XDG_CONFIG_* variables to nonexistent directories for the testsuite so that system or user config does not interfere with tests.

src/bindings/python/flux/util.py
```diff
@@ -596,3 +606,114 @@ def render(
...
+
+                dict_merge(self.dict, conf)
+
+        return self
+
+    def validate(self, path, conf):
+        """
+        Validate config file as it is loaded before merging it with the
+        configuration. This function does nothing in the base class,
+        specific sub classes should implement this method.
```
chu11, line 719: "should" in ietf is "recommended", you use "may" in commit message, suggesting "optional".  So perhaps reword "should" to "may" (adjust sentence as needed of course).

src/bindings/python/flux/util.py
```diff
@@ -596,3 +606,114 @@ def render(
...
+
+
+def get_searchpath(var, suffix=""):
+    value = os.getenv(var)
+    if value is None:
+        return []
+    return [f"{x}{suffix}" for x in value.split(":")]
+
+
+#  Slightly modified from https://stackoverflow.com/posts/7205107/timeline
```
chu11, line 618: i think url is wrong, you put the modification history timeline?

src/bindings/python/flux/util.py
```diff
@@ -596,3 +606,114 @@ def render(
...
+            if isinstance(src[key], dict) and isinstance(new[key], dict):
+                dict_merge(src[key], new[key], path + [str(key)])
+            elif src[key] == new[key]:
+                pass  # same leaf value
+            else:
+                #  Default is to override:
+                src[key] = new[key]
+        else:
+            src[key] = new[key]
+    return src
```
chu11, line 634: for your implementation, I don't think `path` is needed.  In the original it appears it was just used for debugging.

src/cmd/flux-jobs.py
```diff
@@ -290,8 +348,10 @@ def parse_args():
         "-o",
         "--format",
         type=str,
+        default="default",
         metavar="FORMAT",
-        help="Specify output format using Python's string format syntax",
+        help="Specify output format by name (use 'help' to get a list of names)"
+        + ", or using Python's string format syntax",
```
chu11, line 361: "or using" -> "or use"

src/bindings/python/flux/util.py
```diff
@@ -663,14 +663,18 @@ def __init__(self, name, initial_dict=None):
...
             self.dict = dict(initial_dict)
 
-        #  Build config search path in order of reverse precedence
-        #  based on XDG specification:
-        self.searchpath = (
-            ["/etc/xdg/flux"]
-            + get_searchpath("XDG_CONFIG_DIRS", "/flux")
-            + [f"{Path.home()}/.config/flux"]
-            + get_searchpath("XDG_CONFIG_HOME", "/flux")
-        )
```
chu11: with the fixup i don't think you need `get_searchpath()` anymore

src/bindings/python/flux/util.py
```diff
@@ -663,14 +663,18 @@ def __init__(self, name, initial_dict=None):
...
+        confdirs = [os.getenv("XDG_CONFIG_HOME") or f"{Path.home()}/.config"]
+
+        #  Append XDG_CONFIG_DIRS as colon separated path (or /etc/xdg)
+        confdirs += (os.getenv("XDG_CONFIG_DIRS") or "/etc/xdg").split(":")
+
+        #  Append "/flux" to all members of confdirs to build searchpath:
+        self.searchpath = [Path(directory, "flux") for directory in confdirs]
+
+        #  Reorder searchpath into reverse precedence order
+        self.searchpath.reverse()
```
chu11, line 677: it took me awhile to figure out why you did this, b/c to the casual observer it makes no sense why you'd put the dirs in a list then just reverse it (vs just putting it in the list in reverse order).  Perhaps just add note about how its b/c XDG_CONFIG_DIRS has higher precedence with paths listed first.

chu11:
with our impending `queue` support coming, I couldn't help but think we should have a `flux-jobs.toml` template in `etc/flux` and it could be something like.

```
# [format.default]
# description = "Default ..."
# format = "{id.f58:>12} {username:<8.8} {queue:>8} ....`
```

i.e. it might be good to provide some default template alternatives for the system instance that admins can just uncomment if they want it.

Not necessarily for this PR, but we may want to add that as a TODO somewhere.

chu11:
oh whoops, I was referring the source code `etc/flux` directory.  But my comment was more generally, would it be a good idea to have a `flux-jobs.toml` template installed by default so people generally know "oh, maybe I might want to look in here and consider configuring something."

Looking around a bit more, I guess we don't have those toml templates for a lot of flux installed by default.  Maybe that's something we should consider doing.

src/cmd/flux-jobs.py
grondo: I think "or using" is correct here (I think the real problem is the unnecessary comma), because "or using Python's string format syntax" still follows "Specify output format..." as the other half of the "or", but I agree it is awkward. Maybe the order should be reversed here so the parenthetical comes at the end, i.e.

> Specify output format using Python's string format syntax or a defined format by name (use 'help' to get a list of names).

src/bindings/python/flux/util.py
grondo: Yeah, sorry, still a work in progress!

src/bindings/python/flux/util.py
grondo: Good point, we could put `[::-1]` at the end of  the line that splits XDG_CONFIG_DIRS to insert in reverse order,  but I found that even more inscrutable. I'll add a comment that reminds readers that colon-separated paths are typically in order of precedence and we want to reverse that here.

grondo:
> But my comment was more generally, would it be a good idea to have a flux-jobs.toml template installed by default so people generally know "oh, maybe I might want to look in here and consider configuring something."

A good point, but in this case I think the most likely configuration would be in a user's `~/.config/flux` and we can't drop a template in there anyway. :shrug:

chu11:
> A good point, but in this case I think the most likely configuration would be in a user's ~/.config/flux and we can't drop a template in there anyway. shrug

Couldn't one go into `/etc/xdg/flux`?  I thought that was sort of like the "system wide" available ones.  Possibly overriding the default ones?

garlick:
> Couldn't one go into /etc/xdg/flux? I thought that was sort of like the "system wide" available ones. Possibly overriding the default ones?

Maybe that area should be reserved for sites that have extended flux in a way we did not anticipate.  If we anticipate formats that some users might like, then maybe the right approach is to include them as named format strings so they work the same at all sites.

Unrelated: would it make sense to have something like a `--show-format-config` option that emits TOML on stdout for the currently selected format?  Then someone could redirect that to  `~/.config/flux/flux-jobs/foo.toml` and edit to suit?

Another thought: should we suggest or even enforce namespaces on the format strings to keep user formats separate from the system-provided ones?  It might be confusing to have the hotline tell you to type `flux jobs -o default` and have that produce different results depending on the user's config.

garlick:
I'm not sure my angst is justified but that was my thought, yes.

grondo:
Not allowing override is going to be tricky. I'll have to withdraw this PR and come back to it at a later time I think.

Issue #4404, opened 2022-07-12
job-list: handle jobspec-update event


garlick:
Problem:  Missing jobspec attributes may be filled in by the `jobspec-update` event, but `job-list` does not parse that event.

Edit: reference pr #4386

grondo:
Oh, good thinking!

I had in the back of my mind that `jobspec-update` events would be for the instance only and would be ignored by the shell, though the default shell options use case is appealing. If we could do signed `jobspec-update` events, then this could be used by users to modify jobspec after submission, and in that case it would be nice if job-info could apply the updates so that a request for the jobspec would return the current version.

This reminds me, though, that we still need to address #2814

garlick:
Over in #4494 we came to the conclusion that, at least for the current need of assigning jobspec defaults, we would modify the jobspec before it lands in the KVS or the job-manager.  The original, signed jobspec would be retained in `J` unmodified.

In PR #4523, the job shell was modified to unwrap the original jobspec from `J`.

Then in #4520 it is proposed that we drop the environment from the jobspec stored in the KVS.

PR #4529 relocates the jobspec-default plugin to job-ingest.

So in summary, our design for jobspec defaults drifted a bit since this issue was opened and it is no longer urgent for that use case that we support the `jobspec-update` event in the tools.  However, we should leave this issue open as we still may need `jobspec-update` for other user cases like #2799.

grondo:
Pinging this issue since we're looking into modifying a limited set of jobspec attributes post-submission.
So I think the priority of this issue should be bumped up again.

chu11:
Hadn't thought about this issue in awhile, mostly b/c of the short/medium term need.  But if need is super apparent given recent PRs, can obviously bump up.

grondo:
The main use case for now is detailed in #4844

PR #4575, opened 2022-09-15
job-list: split out job information into new files                                                                                          

chu11:
This is a general refactoring PR.                                                                                                           
                                                                                                                                            
I actually did some similar refactoring in #4336

I did some similar refactoring in a (now tabled) PR for #4404

@garlick commented on some potential refactoring in #4555                                                                                   

So thus this refactoring, and why I put it in a separate PR.
                                                                                                                                            
I could add some unit tests to capture some worst offending failure cases (e.g. like jobspec that is just a curly brace), but didn't given its limited coverage increase.  Some of this is already handled in sharness tests.

### DIFF
--- a/src/modules/job-list/Makefile.am
+++ b/src/modules/job-list/Makefile.am
@@ -20,6 +20,8 @@ job_list_la_SOURCES = \
 	job-list.h \
 	job_state.h \
 	job_state.c \
+	job_data.h \
+	job_data.c \
 	list.h \
 	list.c \
 	job_util.h \
--- a/src/modules/job-list/job-list.c
+++ b/src/modules/job-list/job-list.c
@@ -17,6 +17,7 @@
 
 #include "job-list.h"
 #include "job_state.h"
+#include "job_data.h"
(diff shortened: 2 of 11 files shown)


grondo:
I thought the idea proposed in #4555 was to split out some of the common R and jobspec "parsing" into a reusable library (i.e. that could eventually be reused by some other facility in flux-core if necessary) Is this just a stepping stone to that goal since here it seems like all that is done so far is to move some code into a new source file?

chu11:
yeah, this is just moving in that general direction.  The main reason I did this is b/c I sort of had the same "move data structures into new files" in two separate PRs, so I felt this refactoring was sort of necessary to avoid more conflicts going forward.

grondo:
Ok, that seems reasonable. We should indeed try to refactor the jobspec/R parsing into 'struct job' into reusable code with unit tests at some point though.

grondo (approved):
Since the only purpose here is to move code to a new file, seems good to me.

PR #4579, opened 2022-09-16
flux-jobs: support getting job queue and filtering by job queue

chu11:
built on top of #4575 and #4555 b/c there were going to be conflicts and just wanted to avoid them.

Note I didn't support a queue filtering support, which will come in another PR (see new issue #4580)


chu11:
so to my surprise, filtering jobs by queue name seemed to be pretty trivial to add.  I think it's basically the same as filtering by job name.  There's a part of me that thinks I'm missing something .. and maybe I am.

So pushed to this PR instead of starting a new PR.

src/modules/job-list/job_data.c
```diff
@@ -224,20 +249,39 @@ int job_parse_jobspec (struct job *job, const char *s)
...
+            goto nonfatal_error;
+
+        /* if nodes specified, per-resource.type == "node", and
+         * per-resource.count > 0 there is an adjustment on ntasks.
+         */
+        if (type && streq (type, "node") && count > 0)
+            job->ntasks = res[0].count * count;
+    }
+
+    /* Set job->ntasks if not yet set
```
garlick, line 290: Does per-resource type=core need to be handled here as well?

I wonder if this would be slightly clearer if `parse_per_resource()` returned success when it has a valid per-resource setting, and then the code for altering ntasks could be conditional on that rather than on the values of type and count?

src/modules/job-list/job_data.c
```diff
@@ -117,6 +118,30 @@ static const char *parse_job_name (const char *path)
...
+                        "{s:{s:{s?:{s?:{s?:{s?:s s?:i}}}}}}",
+                        "attributes",
+                          "system",
+                            "shell",
+                              "options",
+                                "per-resource",
+                                  "type", type,
+                                  "count",count) < 0) {
+        flux_log (job->h, LOG_ERR,
+                  "%s: job %ju invalid jobspec: %s",
```
garlick, line 137: See `get_per_resource_option()` in the shell info.c - per-resource must contain a type key or it's invalid.

garlick (commented):
Couple of off the cuff comments from reading diffs, but I"ll also try this on my test system to see how it works on a system instance with queues :-)

src/modules/job-list/job_data.c
chu11: doh, this code is from the prior PR #4555 this was built on top of :-)  But it's a good suggestion.  I'll do that as a cleanup commit in this PR (need to rebase now that #4555 is merged).  And I can cleanup the thing you noticed below as well.

garlick:
Tried this on my test system and seems to work well!

```
$ flux jobs  -a -o '{id.f58:>12} {name:<10.10} {status:>2.2} {ntasks:>6} {nnodes:>6h} {queue:>8} {nodelist:h}'
       JOBID NAME       ST NTASKS NNODES    QUEUE NODELIST
 ƒAn6CGSkTo5 hostname   CO      2      2    batch picl[1-2]
 ƒAn67sz1jQK sleep      CO      4      4    batch picl[1-4]
 ƒAn67UrJwNF sleep      CO      4      4    batch picl[1-4]
 ƒAn66v5gp6B sleep      CO      4      4    batch picl[1-4]
 ƒAn4smfyn5d hostname   CO      1      1    debug picl6

$ flux jobs --queue=batch  -a -o '{id.f58:>12} {name:<10.10} {status:>2.2} {ntasks:>6} {nnodes:>6h} {queue:>8} {nodelist:h}'
       JOBID NAME       ST NTASKS NNODES    QUEUE NODELIST
 ƒAn6CGSkTo5 hostname   CO      2      2    batch picl[1-2]
 ƒAn67sz1jQK sleep      CO      4      4    batch picl[1-4]
 ƒAn67UrJwNF sleep      CO      4      4    batch picl[1-4]
 ƒAn66v5gp6B sleep      CO      4      4    batch picl[1-4]

$ flux jobs --queue=debug -a -o '{id.f58:>12} {name:<10.10} {status:>2.2} {ntasks:>6} {nnodes:>6h} {queue:>8} {nodelist:h}'
       JOBID NAME       ST NTASKS NNODES    QUEUE NODELIST
 ƒAn4smfyn5d hostname   CO      1      1    debug picl6
```

src/modules/job-list/job_data.c
chu11: > Does per-resource type=core need to be handled here as well?

Yeah, I think it does.  I was initially under the impression that `--tasks-per-core` would set ntasks the normal way, but that doesn't seem to be the case.  Good catch.

chu11:
re-pushed, bunch of builders failed with an errant `&&` on the end of a line

chu11:
re-pushed, fixed a logic error that was causing issues.  The logic error also existed in the job name filter (which I copied the queue filter from).

grondo (approved):
This LGTM!

PR #4584, opened 2022-09-19
configure: remove --without-python

vsoch:
The `--without-python` flag does not make sense if Flux cannot build without it. This PR tests removing the flag (and if checks) for building the bindings. Discussion in https://github.com/flux-framework/flux-core/pull/4577#issuecomment-1250493250. It could be that we want to keep the flag, but have it only control building the Python wrappers (and not the core logic for getting python working for generating commands, etc.) I'm happy to try both - let me know.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -204,14 +204,6 @@ X_AC_CHECK_COND_LIB(dl, dlerror)
 X_AC_MALLOC
 AC_CHECK_LIB(m, floor)
 
-
-AC_ARG_WITH([python],
-  [AS_HELP_STRING([--without-python],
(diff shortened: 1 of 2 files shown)


vsoch:
Woohoo increasing coverage without writing tests!! (deleting things is the best)

grondo:
Thanks! I think if we need `--without-python` then we can re-implement it in the future with more care, ensuring it works. For now, clearly there is no use case for this so it seems best to remove it. Thanks!

I'm just going to update the PR title before mergify merges it to make it more clear this applies to the build (since PR title automatically goes into release notes)

grondo (approved):
Thank you!

grondo:
I'll set MWP!

PR #4586, opened 2022-09-19
job-manager: print better errors on inactive move

### LINKED ISSUE #4585: error preserving inactive job: No such file or directory
Problem: the following log messages are reliably produced by fluxion's `t1014-annotations.t` sharness test:
```
expecting success: 
    active_jobs=$(flux job list --state=active | jq .id) &&
    for job in ${active_jobs}; do flux job cancel ${job}; done &&
    for job in ${active_jobs}; do flux job wait-event -t 10 ${job} clean; done

2022-09-19T15:01:18.980576Z job-manager.err[0]: 10351542272: error preserving inactive job: No such file or directory
2022-09-19T15:01:18.982506Z job-manager.err[0]: 12012486656: error preserving inactive job: No such file or directory
2022-09-19T15:01:18.984429Z job-manager.err[0]: 13706985472: error preserving inactive job: No such file or directory
```

As noted in in [the code](https://github.com/flux-framework/flux-core/blob/4113f741ea171b70b15b3bf5b595ca3c3c49e8d3/src/modules/job-manager/event.c#L32), `event_job_action()` is supposed to be idempotent, and while the code for adding jobs to the inactive jobs hash as part of the INACTIVE action is idempotent, these log messages would seem to indicate that there is a problem when the action is run more than once.

garlick:
Problem: we get 'error preserving inactive job: no such file or directory' errors on occasion.

The event_job_action() function is documented as idempotent so expect that the INACTIVE state action may be executed multiple times, and do not log this as an error.

Fixes #4585

### DIFF
--- a/src/modules/job-manager/event.c
+++ b/src/modules/job-manager/event.c
@@ -430,11 +430,17 @@ int event_job_action (struct event *event, struct job *job)
             job->eventlog_readonly = 1;
             if ((job->flags & FLUX_JOB_WAITABLE))
                 wait_notify_inactive (ctx->wait, job);
-            if (zhashx_insert (ctx->inactive_jobs, &job->id, job) < 0
-                || purge_enqueue_job (ctx->purge, job) < 0) {
-                flux_log_error (event->ctx->h,
-                                "%ju: error preserving inactive job",
-                                (uintmax_t) job->id);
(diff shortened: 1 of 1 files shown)


grondo (approved):
LGTM!

PR #4587, opened 2022-09-19
ingest: set configured queue constraints

garlick:
This adds an ingest "frobnicator" plugin (thanks @grondo) that works with the RFC 33 `queues.NAME.requires` spec proposed in flux-framework/rfc#342. The broker policy/queue config validator is also modified to allow `queues.NAME.requires`.

This was sufficient to get my test system queues working again with this config
```toml
[ingest.validator]
plugins = ["feasibility", "jobspec" ]

[ingest.frobnicator]
plugins = [ "defaults", "constraints" ]

[[resource.config]]
hosts = "picl[0-7]"
cores = "0-3"

[[resource.config]]
hosts = "picl[1-5]"
properties = ["batch"]

[[resource.config]]
hosts = "picl[6-7]"
properties = ["debug"]

[queues.debug]
requires = [ "debug" ]

[queues.batch]
requires = [ "batch" ]

queue = "debug"
```
Marked as WIP pending

- [x] more testing
- [x] add flux-config-queues(5)

### DIFF
--- a/doc/man5/flux-config-queues.rst
+++ b/doc/man5/flux-config-queues.rst
@@ -1,0 +1,88 @@
+=====================
+flux-config-queues(5)
+=====================
+
+
+DESCRIPTION
+===========
+
+The ``queues`` table configures job queues, as described in RFC 33.
+Normally, Flux has a single anonymous queue, but when queues are configured,
+all queues are named, and a job submitted without a queue name is rejected
+unless a default queue is configured.
+
+Each key in the ``queues`` table is a queue name, whose value is a table
+consisting of the following optional keys:
+
+requires (array)
+   A list of resource property names that selects a subset of the total
+   configured resources for the queue.  Each property name is a string, as
+   defined in RFC 20.  If multiple properties are listed, they are combined
+   in a logical *and* operation.  Properties are attached to resources in the
+   ``resource`` table as described in :man5:`flux-config-resource`.
+
+policy (table)
+   A policy table as described in :man5:`flux-config-policy` that overrides
+   the general system policy for jobs submitted to this queue.
+
+A default queue name may be configured by setting
+``policy.jobspec.defaults.system.queue`` as described in
+:man5:`flux-config-policy`.
+
+
+EXAMPLE
+=======
+
+::
+
+   [ingest]
+   validator.plugins = ["feasibility", "jobspec"]
+   frobnicator.plugins = ["defaults", "constraints"]
+
+   [[resource.config]]
+   hosts = test[0-7]
+   properties = ["debug"]
+
+   [[resource.config]]
+   hosts = test[8-127]
+   properties = ["batch"]
+
+   [queues.debug]
+   policy.limits.duration = "30m"
+   requires = [ "debug" ]
+
+   [queues.batch]
+   policy.limits.duration = "8h"
+   requires = [ "batch" ]
+
+   [policy.jobspec.defaults.system]
+   queue = "batch"
+
+   [sched-fluxion-qmanager]
+   queues = "batch debug"
+   queue-policy-per-queue="batch:easy debug:fcfs"
+
+   [sched-fluxion-resource]
+   match-policy = "lonodex"
+   match-format = "rv1_nosched"
+
+
+CAVEATS
+=======
+
+Queue resources should not overlap.
+
+
+RESOURCES
+=========
+
+Flux: http://flux-framework.org
+
+RFC 20: Resource Set Specification Version 1: https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_20.html
+
+RFC 33: Flux Job Queues: https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_33.html
+
+SEE ALSO
+========
+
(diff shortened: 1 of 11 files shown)


src/broker/brokercfg.c
```diff
@@ -286,25 +286,37 @@ static int validate_queues_config (flux_conf_t *conf, flux_error_t *error)
...
-                return -1;
+                goto inval;
             }
             if (policy) {
                 char key[1024];
                 snprintf (key, sizeof (key), "queues.%s.policy", name);
                 if (validate_policy_json (policy, key, error) < 0)
                     return -1;
             }
+            if (requires) {
```
grondo, line 306: One thing we might want to do here is validate that each configured property is valid according to [RFC 20](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_20.html), i.e. does not contain whitespace or any characters from the set `! & ' " ^ ` | ( )`.

grondo (commented):
LGTM so far. I had just one thought while doing a first pass (sorry the comment was meant to be an individual comment, not a review, but I had already clicked "Start a Review")

garlick:
Force pushed with a new `flux-config-queues(5)` man page and changes to validate property names, per  @grondo's comments.

src/broker/brokercfg.c
grondo: Ah, I was wrong! Actually `^` should be allowed since that is currently our shorthand for `not`, e.g. a sysadmin should be able to configure a `queue.NAME.requires` of `[ "^batch" ]` which means that queue  `NAME` excludes all resources with the `batch` property.

I'm sorry for messing that one up.

grondo (approved):
LGTM. I unfortunately gave bad advice in my previous comment  (the `^` character at least should be allowed in `requires` (sorry about that), and had one trivial suggestion for the `flux-config-queues(5)` man page.

garlick:
NP!  I should have caught it - I did skim RFC 20 after your comment but missed it too.

Any thoughts for further testing of the constraints plugin before we merge?

grondo:
> Any thoughts for further testing of the constraints plugin before we merge?

Yeah, here's some ideas:

 * ensure the constraints plugin can be specified even if no queues are configured
 * ensure the constraints plugin can be specified when the queue does not have a `requires` key

Also, I noticed that the constraints plugin I gave you allows for a non-list comma separated string to be specified for `requires` I think. That code could be removed.

garlick:
> Also if you drive the constraints plugin from flux job-frobnicator directly you could test that some of the invalid configuration is handled correctly there as well (e.g. requires not a list, invalid queue specified...

Since the config comes from the broker, and the broker now checks the config, this is not really possible.

Just squashed the following changes
- fix up a missed check in the broker queues validation
- add some more tests based on your suggestions
- tweak man page opening paragraph per suggestion
- allow ^ in property names
- eliminate support for spitting comma-separated properties in constraints plugin
- rebase on current master

garlick:
oops I need to try that again - I accidentally squashed the wrong pair of commits

grondo:
> Since the config comes from the broker, and the broker now checks the config, this is not really possible.

Ah, yes, then I suppose the code to check validity of the config in the plugin is actually dead code and can be removed?

grondo (approved):
LGTM!

garlick:
> Ah, yes, then I suppose the code to check validity of the config in the plugin is actually dead code and can be removed?

Well, that prompted me to look through what was being checked and discover a couple of cases that were missed by the broker check (default queue not in `[queues]`, queue policy contains default queue) and cover those.  A few tests in `t2112-job-ingest-frobnicator.t` were replaced with tests in `t2241-policy-config.t` (which checks the broker config check).

I also discovered that the man page example didn't have the default queue in `[queues]` and fixed taht (doh!)

However, the checks in the frobnicator plugins are pretty minimal and do serve as documentation of expectations, if nothing else (like an assertion).  Maybe OK to leave them in?

doc/man5/flux-config-queues.rst
```diff
@@ -0,0 +1,88 @@
...
+
+   [sched-fluxion-resource]
+   match-policy = "lonodex"
+   match-format = "rv1_nosched"
+
+
+CAVEATS
+=======
+
+Queue resources should not overlap.
```
grondo, line 73: Oh, I was going to ask about this caveat. Why can't queue resources overlap (is this a Fluxion constraint?).

garlick: Fluxion allows it, but some issues were identified in flux-framework/flux-sched#939.

Some use cases might work with Fluxion as is, like opening up a *stopped* queue to select users before a DAT, then when the DAT comes around, stop/drain the normal queues and start the DAT queue.   That particular one would require the job manager to support multiple queues though. 

Maybe there are other cases, but I thought it best to not advertise it at this point since the near term goal is partitioned resources.

grondo: Ah, thanks, I had forgotten about that. I do know we were talking about allowing a queue with access to all resources (and we do have a `pall` partition on many clusters), and this caveat prevents that option. I agree we can ignore it for now as long as the concept of non-overlapping resource sets isn't baked into the design.

garlick:
Setting MWP, thanks!

PR #4588, opened 2022-09-19
Fix obsolete autotools warnings

woodard:
Fix a bunch warnings running autogen.sh that show up on Fedora 36. Many of these fixes were automated with autoupdate tool. However, this also made some non-human readable formatting. This fixes the those formatting problems.

ax_compile_check_sizeof.m4 was updated from upstream. lx_find_mpi.m4 seems to be written by LLNL and used in many packages. All of the rest of those packages will need to be updated as well.

Signed-off-by: Ben Woodard <woodard@redhat.com>


grondo:
Instead of merging master into PR branches, we always rebase (i.e. `git rebase origin/master`) so that we don't have any extra merge commits. (This is why the `validate commits` action is failing on this PR)

grondo:
The el7 build seems to have failed very early here:
```
Making all in libpmi
    CC       simple_client.lo
  simple_client.c: In function 'pmi_simple_client_abort':
  simple_client.c:376:9: error: 'for' loop initial declarations are only allowed in C99 mode
           for (char *cp = cpy; *cp != '\0'; cp++) {
           ^
  simple_client.c:376:9: note: use option -std=c99 or -std=gnu99 to compile your code
  make[3]: *** [Makefile:975: simple_client.lo] Error 1
  make[3]: *** Waiting for unfinished jobs....
```

It appears that somehow the changes here have dropped whatever flags enable C99 in the EL7 build. I'm unsure what that would be though...

configure.ac
```diff
@@ -94,7 +94,7 @@ AC_SUBST([LIBFLUX_HOSTLIST_VERSION_INFO])
 ##
 # Checks for programs
 ##
-AC_PROG_CC_C99
```
woodard, line 97: This is obsolete but it may be the way that you forced the RHEL7 compilers into C99 mode.

grondo (approved):
LGTM and seems to working in all the builders now. Thanks!

PR #4591, opened 2022-09-20
flux-jobs: support collapsible format fields

grondo:
This PR adds a way to have format fields in flux-jobs that are optionally completely removed when they render as an empty string for all jobs. So-called "collapsible" fields are indicated in a format  by prepending `?:` before the format, e.g. `?:{name}`. The main use case here is so that we can optionally list jobs queues in the system instance in the default format, but not have this empty, useless field present for all single user instances which are most likely using anonymous queues. E.g.:

```console
$ flux jobs -ao '{id.f58:>12} ?:{exception.type:>16} {nodelist}'
       JOBID   EXCEPTION-TYPE NODELIST
    ƒEiwyRCb             exec asp
    ƒC5m5dL3          timeout asp
    ƒ4H3iwEo                  asp
    ƒ3uw9ATy                  asp
$ flux jobs -f completed -ao '{id.f58:>12} ?:{exception.type:>16} {nodelist}'
       JOBID NODELIST
    ƒ4H3iwEo asp
    ƒ3uw9ATy asp
```

There's no documentation here yet, I wanted to get this up in case anyone had strong objections to the approach.

### DIFF
--- a/doc/man1/flux-jobs.rst
+++ b/doc/man1/flux-jobs.rst
@@ -149,6 +149,15 @@ following is the format used for the default format:
 
    {id.f58:>12} {username:<8.8} {name:<10.10} {status_abbrev:>2.2} {ntasks:>6} {nnodes:>6h} {runtime!F:>8h} {nodelist:h}
 
+If a format field is preceded by the special string ``?:`` this will
+cause the field to be removed entirely from output if the result would
+be an empty string for all jobs in the listing. E.g.::
+
+   {id.f58:>12} ?:{exception.type}
+
+would eliminate the EXCEPTION-TYPE column if no jobs in the list received
+an exception.
+
 The special presentation type *h* can be used to convert an empty
 string, "0s", "0.0", or "0:00:00" to a hyphen. For example, normally
 "{nodelist}" would output an empty string if the job has not yet run.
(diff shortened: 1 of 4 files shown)


garlick (commented):
Neat!

Does it make sense to add something to flux-jobs(1) OUTPUT FORMAT section?

grondo:
> Does it make sense to add something to flux-jobs(1) OUTPUT FORMAT section?

Yes, sorry I probably should have made this a WIP and noted I didn't add any documentation (in case there were objections to the approach)

chu11:
I like the approach too.  I suppose the only bad thing is the extra O(n) loop through job entries.  Which in practice shouldn't matter too much given we have a max entries output count.

grondo:
>  I suppose the only bad thing is the extra O(n) loop through job entries. Which in practice shouldn't matter too much given we have a max entries output count.

Yeah, that worried me a bit too so I timed it. The extra loop over 1000 jobs took about 19ms while the overall time to display 1000 jobs was well over a second.

garlick (approved):
Works for me!  Not sure if @chu11 had further thoughts?

chu11 (approved):
nope, LGTM too

grondo:
Thanks @chu11 and @garlick! Setting MWP.

PR #4594, opened 2022-09-21
flux-jobs: support --format=NAME

grondo:
This is the pared down version of #4571. Basically this allows extra internally defined named formats, starting with `default` and `deps`. Pretty underwhelming, but we could add `long` here as well if we can agree on an internal long format.


grondo:
Good thoughts. Maybe we should open an issue to iterate?

grondo:
Closing this alternate PR since #4595 was approved.

PR #4595, opened 2022-09-21
flux-jobs: add configurable named formats

grondo:
It turns out it wasn't so difficult to reject overrides of internal default formats in the `flux jobs` configuration support.

Unfortunately, I can't re-open #4571, so we'll have to start over here.

This updates #4571 such that any internally defined formats for `flux-jobs` can't be overridden by system or user configuration.
```conf
$ flux jobs 
flux-jobs: ERROR: /home/grondo/.config/flux/flux-jobs.toml: override of builtin format 'default' not permitted
```

It also tentatively adds support for `--format=get-config=NAME` as requested:
```console
$ flux jobs -o get-config=default
[formats.default]
description = "Default flux-jobs format string"
format = "{id.f58:>12} {username:<8.8} {name:<10.10} {status_abbrev:>2.2} {ntasks:>6} {nnodes:>6h} {runtime!F:>8} {nodelist:h}"
```

This feature is added and tested in separate commits for now in case we want to drop that feature.

The ability to list what config file defined a given format was not added at this time, because I could not determine an easy way to do that. If that is a showstopper then PR #4594 should be merged instead and we can table this one again.

### DIFF
--- a/doc/man1/flux-jobs.rst
+++ b/doc/man1/flux-jobs.rst
@@ -61,9 +61,12 @@ OPTIONS
    results can be listed separated by comma. See JOB STATUS below for
    additional information. Defaults to *pending,running*.
 
-**-o, --format**\ *=FORMAT*
-   Specify output format using Python's string format syntax. See OUTPUT
-   FORMAT below for field names.
+**-o, --format**\ *=NAME|FORMAT*
+   Specify a named output format *NAME* or a format string using Python's
+   format syntax. See OUTPUT FORMAT below for field names. Named formats
+   may be listed via ``--format=help``. Additional named formats may be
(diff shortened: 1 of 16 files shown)


grondo:
Pushed a fix for wording issue in `flux-jobs(1)` caught by @chu11. I also added a note that `XDG_CONFIG_DIRS` is traversed in reverse order so that the precedence order of colon-separated paths is preserved.

chu11 (approved):
LGTM, nothing else caught my eye.

garlick (approved):
This looks great to me!

grondo:
Thanks! I'll set MWP and delete the other PR.

PR #4597, opened 2022-09-22
flux-jobs: support width and alignment with `!d` conversion specifier

grondo:
The `flux-jobs` `!d` conversion specifier converts job timestamps to Python `datetime` objects so that `strftime(3)` format specifiers can be used to generate flexible representations of time. However, a drawback of this is that the `datetime` class' `__format__` method does not support width and alignment specifiers (i.e. `:>20`), which makes it impossible to use with other fields because the placement of the field (and the header) can't be controlled.

This PR introduces a `JobDatetime` class which subclasses `datetime` and replaces the `__format__` method with a version that allows the Python format spec to occur after any time format separated by `::`. The `!d` conversion specifier is then updated to convert to `JobDatetime` instead of `datetime`.

Now this works:
```console
$ flux jobs -ao '{id.f58:>12} {t_submit!d:%b%d %R::>12}'
       JOBID     T_SUBMIT
  ƒ5C6TPdeHm  Sep21 19:32
  ƒ5C52XDnfq  Sep21 19:32
  ƒ3CqkPSKDR  Sep21 15:19
```

### DIFF
--- a/src/bindings/python/flux/job/info.py
+++ b/src/bindings/python/flux/job/info.py
@@ -366,6 +366,31 @@ def fsd(secs):
     return strtmp
 
 
+class JobDatetime(datetime):
+    """
+    Subclass of datetime but with a __format__ method that supports
+    width and alignment specification after any time formatting via
+    two colons `::` before the Python format spec, e.g::
+
+    >>> "{0:%b%d %R::>16}".format(JobDatetime.now())
+    '     Sep21 18:36'
+    """
+
+    def __format__(self, fmt):
+        # The string "::" is used to split the strftime() fromat from
+        # any Python format spec:
+        vals = fmt.split("::", 1)
+
+        # Call strftime() to get the formatted datetime as a string
+        result = self.strftime(vals[0])
+
+        # If there was a format spec, apply it here:
+        try:
+            return f"{{0:{vals[1]}}}".format(result)
+        except IndexError:
+            return result
+
+
 class JobInfoFormat(flux.util.OutputFormat):
     """
     Store a parsed version of an output format string for JobInfo objects,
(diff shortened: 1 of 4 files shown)


src/bindings/python/flux/job/info.py
```diff
@@ -366,6 +366,31 @@ def fsd(secs):
     return strtmp
 
 
+class JobDatetime(datetime):
```
chu11, line 369: commit message typo "timestapms" -> "timestamps"

t/t2800-jobs-cmd.t
```diff
@@ -805,6 +805,23 @@ test_expect_success 'flux-jobs --format={expiration!d:%FT%T},{t_remaining!H} wor
...
 '
+test_expect_success 'flux-jobs --format={expiration!d:%FT%T::>20.20} works' '
+	cat expiration.in | \
+	    flux jobs --from-stdin -o "{expiration!d:%b%d %R::>20.20}" \
+	    >exp-fmt.out &&
+	test_debug "cat exp-fmt.out" &&
+	grep "          EXPIRATION" exp-fmt.out &&
+	grep "         $(date --date=@${exp} +%b%d\ %R)" exp-fmt.out
+'
+test_expect_success 'flux-jobs --format={expiration!d:%FT%T::^=20} works' '
```
chu11, line 816: neat, I didn't know about "fill" formatting.  typo "^=" -> "=^"

chu11 (approved):
LGTM, just two nits

grondo:
Good catches @chu11. I'll pushed the fixes and will set MWP.

grondo:
I don't know why LGTM and readthedocs builds keep failing on this branch :confused: Will try another push in a bit, or let mergify do it if another PR is merged first.

PR #4601, opened 2022-09-22
job-list: handle per-resource.type=core corner case when setting ntasks

chu11:
#4555 handled the `per-resource.type=node` corner case, but missed the `per-resource.type=core` case.  This adds the core case.

### DIFF
--- a/src/modules/job-list/job_data.c
+++ b/src/modules/job-list/job_data.c
@@ -123,20 +123,31 @@ static int parse_per_resource (struct job *job,
                                int *count)
 {
     json_error_t error;
+    json_t *o = NULL;
 
     if (json_unpack_ex (job->jobspec, &error, 0,
-                        "{s:{s:{s?:{s?:{s?:{s?:s s?:i}}}}}}",
+                        "{s:{s:{s?:{s?:{s?:o}}}}}",
                         "attributes",
                           "system",
                             "shell",
                               "options",
-                                "per-resource",
-                                  "type", type,
-                                  "count",count) < 0) {
+                                "per-resource", &o) < 0) {
         flux_log (job->h, LOG_ERR,
                   "%s: job %ju invalid jobspec: %s",
                   __FUNCTION__, (uintmax_t)job->id, error.text);
         return -1;
+    }
+
+    if (o) {
+        if (json_unpack_ex (o, &error, 0,
+                            "{s:s s?:i}",
+                            "type", type,
+                            "count", count) < 0) {
+            flux_log (job->h, LOG_ERR,
+                      "%s: job %ju invalid per-resource spec: %s",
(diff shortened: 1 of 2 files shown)


src/modules/job-list/job_data.c
```diff
@@ -244,6 +244,11 @@ static int parse_jobspec_ntasks (struct job *job,
             && res[2].type != NULL && !strcmp (res[2].type, "core")
             && res[2].with == NULL)
             job->ntasks = res[0].count * count;
+        else if (streq (type, "core")
+                 && res[0].type != NULL && !strcmp (res[0].type, "slot")
+                 && res[1].type != NULL && !strcmp (res[1].type, "core")
+                 && res[1].with == NULL)
+            job->ntasks = res[0].count * count;
```
grondo, line 251: I'm not sure this is correct. With `per-resource.type=core` the number of tasks is the number of total cores (not slots) multiplied by `per-resource.count`. E.g.
```
$ flux mini run -c 2 -o per-resource.type=core -o per-resource.count=2 hostname
asp
asp
asp
asp
ƒ(s=1,d=0,builddir) grondo@asp:~/git/flux-core.git$ flux jobs -a
    ƒdgFNVaF grondo   hostname   CD      2      1   0.048s asp
```

`NTASKS` here should be 4 not 2.

grondo: There's another case that doesn't work, but this one is trickier. Sometimes we don't know the number of tasks until the job has resources assigned, e.g.
```
$ flux mini run -N1 --tasks-per-core=2 hostname
asp
asp
asp
asp
asp
asp
asp
asp
ƒ(s=1,d=0,builddir) grondo@asp:~/git/flux-core.git$ flux jobs -ac1
   ƒ5BuNW95Z grondo   hostname   CD      1      1   0.055s asp
```

`NTASKS` should be 8 here because 1 node with 4 cores was allocated exclusively with `--tasks-per-core=2`. In this case job-list doesn't know how many total tasks there will be until the job is allocated resources, so IMO NTASKS should remain unset until _R_ is available.

chu11: thanks, i clearly misunderstood the parsing of the jobspec + per-resource=core.  Lemme work on these.

chu11:
i'm going to close this, it appears the original `ntasks` logic is also out of date.  I think we originally developed this logic back when `-n` and `-N` and maybe nothing else was supported.

```
>flux mini submit -n 2 -c 2 hostname
ƒFf5DETCK
>flux jobs ƒFf5DETCK
   ƒFf5DETCK achu     hostname   CD      2      1   0.062s catalyst160
```

which is getting me confused on a number of these corner cases.  So lets revisit this one later.

PR #4605, opened 2022-09-23
flux-top: add minimal support for job queues

grondo:
This PR simply adds a QUEUE column to `flux-top` when any jobs in the job list have a queue defined.

I thought about adding a `--queue=NAME` option to filter jobs by queue, but this needs more thought since the resource list and job statistics will still apply the entire instance so this may cause more confusion than insight. 

(I'll open a new issue on this so that the original issue tied to our milestone can be closed)

### DIFF
--- a/src/cmd/top/joblist_pane.c
+++ b/src/cmd/top/joblist_pane.c
@@ -29,6 +29,8 @@ struct joblist_pane {
     json_t *jobs;
     struct ucache *ucache;
 
+    bool show_queue;
+
     /*  Currently selected jobid. Ironically FLUX_JOBID_ANY means
      *   no current selection.
      */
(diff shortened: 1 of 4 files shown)


src/cmd/top/joblist_pane.c
```diff
@@ -207,6 +207,11 @@ static void joblist_continuation (flux_future_t *f, void *arg)
     json_decref (joblist->jobs);
     joblist->jobs = json_incref (jobs);
     joblist_pane_draw (joblist);
+    if (joblist->top->test_exit) {
```
chu11, line 210: commit message "change the behavior of the --test-only" -> "--test-exit" or perhaps "change the behavior of the test option".

chu11 (approved):

src/cmd/top/joblist_pane.c
grondo: D'oh! Thanks! Yeah it is the `--test-exit` option :facepalm:

grondo:
Ok, I fixed that silly mistake and pushed the result.

PR #4608, opened 2022-09-23
ingest: enable frobnicator when needed by [queues] or [policy]

garlick:
Problem: the configuration of queues and jobspec defaults is made more error prone by the need to configure the ingest frobnicator.

Since we know the frobnicator is needed when those other things are configured, just enable if implicitly when those configs are present.  Also, to simplify this, implement a default plugin list that includes `defaults` and `constraints`.

A small bug fix to the defaults plugin is included, and a little bit of test cleanup.

### DIFF
--- a/src/bindings/python/flux/job/frobnicator/frobnicator.py
+++ b/src/bindings/python/flux/job/frobnicator/frobnicator.py
@@ -66,6 +66,7 @@ class JobFrobnicator:
     namespace.
     """
 
+    default_frobnicators = ["defaults", "constraints"]
     plugin_namespace = "flux.job.frobnicator.plugins"
 
     def __init__(self, argv, pluginpath=None, parser=None):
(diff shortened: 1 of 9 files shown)


src/bindings/python/flux/job/frobnicator/frobnicator.py
```diff
@@ -66,6 +66,7 @@ class JobFrobnicator:
     namespace.
     """
 
+    default_frobnicators = ["defaults", "constraints"]
```
grondo, line 69: commit message nit: "Loads 'defaults' and 'constraints' plugins..." :arrow_right: "Load 'defaults' and 'constraints' plugins..."

grondo (approved):
Great improvement!

src/bindings/python/flux/job/frobnicator/frobnicator.py
garlick: fixed that and forced a push

garlick:
I'll go ahead and set MWP here then, thanks!

Issue #4596, opened 2022-09-21
t2808-shutdown-cmd.t hangs in CI


grondo:
I've been seeing this problem occur in CI routinely enough to open an issue. It appears that all tests finish, but then the sharness script hangs at the end somehow and times out. Needs more investigation.

```
./t2808-shutdown-cmd.t timed out after 300.0s
  2022-09-[21](https://github.com/flux-framework/flux-core/actions/runs/3100715735/jobs/5021298718#step:9:22)T20:27:08.453793Z broker.err[0]: cleanup.1: flux-job: Canceled 4 jobs (0 errors)
  ok 1 - flux-shutdown -h prints custom usage
  PASS: t2808-shutdown-cmd.t 1 - flux-shutdown -h prints custom usage
  ok 2 - flux-shutdown fails on unknown option
  PASS: t2808-shutdown-cmd.t 2 - flux-shutdown fails on unknown option
  ok 3 - flux-shutdown fails if FLUX_URI is set wrong
  PASS: t2808-shutdown-cmd.t 3 - flux-shutdown fails if FLUX_URI is set wrong
  ok 4 - flux-shutdown fails if job argument is not a valid URI
  PASS: t2808-shutdown-cmd.t 4 - flux-shutdown fails if job argument is not a valid URI
  ok 5 - flux-shutdown fails if job argument is unknown
  PASS: t2808-shutdown-cmd.t 5 - flux-shutdown fails if job argument is unknown
  ok 6 - run a test job to completion
  PASS: t2808-shutdown-cmd.t 6 - run a test job to completion
  ok 7 - flux-shutdown fails if job is not running
  PASS: t2808-shutdown-cmd.t 7 - flux-shutdown fails if job is not running
  ok 8 - submit batch script and wait for it to start
  PASS: t2808-shutdown-cmd.t 8 - submit batch script and wait for it to start
  ok 9 - flux-shutdown JOBID works
  PASS: t2808-shutdown-cmd.t 9 - flux-shutdown JOBID works
  ok 10 - shutdown output contains rc3 exit code (0)
  PASS: t2808-shutdown-cmd.t 10 - shutdown output contains rc3 exit code (0)
  ok 11 - shutdown output contains state transition to goodbye
  PASS: t2808-shutdown-cmd.t 11 - shutdown output contains state transition to goodbye
  ok 12 - job exit code indicates SIGHUP termination
[...]

PR #4609, opened 2022-09-23
testsuite: remove errant test_done call

chu11:
Problem: An debugging test_done call was left in t2800-jobs-cmd.t, therefore not running any tests after the call.

Remove the errant call.


chu11:
I think two PRs were merged with this left in by accident, hopefully no regression failures :P

grondo (approved):
Eek! Sorry about that!

chu11:
restarted a builder that hit #4596

chu11:
hmmm the mergify queue seems to be stuck, rebased and forced a push

Issue #4610, opened 2022-09-23
job-list: use `libjj` to avoid changes in nnodes or ntasks calculation logic


chu11:
it appears the logic to calculate `ntasks` is out of date.  Logic was developed a long time ago and all the options present in flux today may not be covered.  Extra test coverage should also be done.

example

```
>flux mini submit -n 2 --cores-per-task=2 hostname
ƒJ1MTBPD1
>flux jobs ƒJ1MTBPD1
       JOBID USER     NAME       ST NTASKS NNODES  RUNTIME NODELIST
   ƒJ1MTBPD1 achu     hostname   CD      2      1   0.063s catalyst160
```

clearly ntasks should be 4 in this case.

chu11:
as @grondo mentions in #4601 there are cases where we don't know the number of cores until the node is allocated

```
$ flux mini run -N1 --tasks-per-core=2 hostname
asp
asp
asp
asp
asp
asp
asp
asp
ƒ(s=1,d=0,builddir) grondo@asp:~/git/flux-core.git$ flux jobs -ac1
       JOBID USER     NAME       ST NTASKS NNODES  RUNTIME NODELIST
   ƒ5BuNW95Z grondo   hostname   CD      1      1   0.055s asp
```

chu11:
edit: no i was wrong in the above

```
>flux mini submit -n 2 --cores-per-task=2 hostname
ƒJ1MTBPD1
>flux jobs ƒJ1MTBPD1
       JOBID USER     NAME       ST NTASKS NNODES  RUNTIME NODELIST
   ƒJ1MTBPD1 achu     hostname   CD      2      1   0.063s catalyst160
```

ntasks is correct here, `--cores-per-task` is not the same as `--tasks-per-core`

PR #4611, opened 2022-09-24
job-list: use libjj to parse jobspec

chu11:
Problem: The initial implementation of job-list's ntasks parsing
was copied from the the job-shell in v0.14.0.  As flux moves forward,
there were questions if the parsing logic in job-list still was
correct.  Even if still correct, there was potential for logic
changes in the future.

Solution: To simplify things, use libjj to parse nnodes and ntasks
from the jobspec.  In the event any logic changes happen underneath
the covers, job-list will automatically pick them up from libjj.

Fixes https://github.com/flux-framework/flux-core/issues/4610


src/modules/job-list/job_data.c
```diff
@@ -123,22 +123,33 @@ static int parse_per_resource (struct job *job,
...
-                                  "count",count) < 0) {
+                                "per-resource", &o) < 0) {
         flux_log (job->h, LOG_ERR,
                   "%s: job %ju invalid jobspec: %s",
                   __FUNCTION__, (uintmax_t)job->id, error.text);
         return -1;
     }
 
+    if (o) {
+        if (json_unpack_ex (o, &error, 0,
```
grondo, line 175: Looks like the job shell initializes `count` to `1` when the `per-resources` key exists, but `count` is not set.  If I'm reading the code correctly, this code would leave `count` set to `0`.

src/modules/job-list/job_data.c
```diff
@@ -118,6 +118,71 @@ static const char *parse_job_name (const char *path)
...
+                  __FUNCTION__, (uintmax_t)job->id, error.text);
+        return -1;
+    }
+
+    if (!json_is_array (command)) {
+        flux_log (job->h, LOG_ERR,
+                  "%s: job %ju invalid jobspec",
+                  __FUNCTION__, (uintmax_t)job->id);
+        return -1;
+    }
```
grondo, line 113: Now that this code is in its own function, it is clear this block of code is unused when an explicit job name is set in jobspec. This stuff could be moved into the `if (!job->name)` block and therefore only executed when the job does not otherwise already have a name set.

grondo (commented):
Nice cleanup by switching to libjj. Just found one potential issue and one trivial suggestion.

src/modules/job-list/job_data.c
chu11: ahhh i didn't realize that subtlety.  So it's more like count is optionally specified, assumed `1` is not specified.

grondo: Yeah, sorry. In retrospect the current behavior seems like a bad idea, but sadly it is what it is.

chu11:
re-pushed with tweaks per @grondo comments above

grondo:
I probably should have thought of this sooner, but would it make sense to move the parsing of `per-resource` key in jobspec (when set) into libjj, then use that here and in the job shell via libjj?

It is fine if you want to save that for another PR. I haven't looked at how much work it might be, but if it is quick and easy, it might be good to get it landed in a single PR.

chu11:
> I probably should have thought of this sooner, but would it make sense to move the parsing of per-resource key in jobspec (when set) into libjj, then use that here and in the job shell via libjj?

I guess in my mind I thought of `libjj` as "permanent" while `per-resource` is a temporary workaround?  But this workaround may linger for quite some time.

And now that you mention it, maybe it doesn't matter so much as `libjj` is an internal library (although it is placed within `libjob` which is not internal, thus maybe why I thought of that way).

Why don't we save that for another PR since it would involve digging into the shell as well.  I can write up an issue.

grondo:
Oh, yeah, and I was getting the shell and sched-simple confused, so nevermind.

chu11:
> Oh, yeah, and I was getting the shell and sched-simple confused, so nevermind.

Nevermind as in not a good idea?  As I skim the shell code now, I can see it sort of has its own parser for `jobspec` as well, not sure if `libjj` can just be stuffed in there.

grondo:
Yeah, libjj is more for the scheduler and the shell has to parse the jobspec in context of R so a bit different. Probably good to just ignore my suggestion for now. :-)

grondo (approved):
Ok, LGTM!

Issue #4573, opened 2022-09-15
job-exec: allow fatal job exception on EHOSTUNREACH to be optionally disabled


grondo:
Problem: The job-exec module terminates all jobs in which it loses contact to the job shell, such as when a node goes down, but this prevents execution of resilient child instances which can tolerate a down node.

As a first step, the job-exec module should support a jobspec attribute to disable this behavior.

@garlick had suggested this option should eventually support an idset of ranks for which lost contact should be a fatal job exception. This would let a child instance register any router nodes as critical. (However, I think we'll open a separate issue on that and some other child instance resiliency issues)

Issue #4417, opened 2022-07-19
use case: SCR needs allocations to tolerate single node failure


ofaaland:
A job that needs many nodes has a higher likelihood that a node will fail during the run.  A user may handle this by requesting an extra node (N+1 total nodes).  The user launches the job on N nodes, leaving one idle, and the job stores checkpoints periodically.  When one of the N nodes fails, the user re-launches the job on the remaining N good nodes (using the formerly spare node), which reads in the latest checkpoint file and restarts.

Without node-local storage or without SCR, this is still desirable to users as requesting a new allocation after the failure may put them well back in the scheduler's queue and delay completion of the job significantly.

With node-local storage and SCR, the checkpoints will be stored on node-local storage.  Re-starting the job can be accomplished by calculating the checkpoint data that will be needed on the "formerly spare" node (from the N-1 nodes that are still up) and writing it to node-local storage there, and then re-launching the job.  This is much, much faster than copying all the checkpoint data to a global file system and then reading it back on a new set of nodes.

Note that users might actually allocate more than one extra node, this is just an example.

ofaaland:
Currently a node failure is fatal to the allocation, although there is a few-second period after the downed node is drained and before the allocation is lost:
```
[faaland1@fluke6 ~] $rpm -qa | grep ^flux | sort
flux-accounting-0.17.0-1.t4.x86_64
flux-core-0.41.0-4.t4.x86_64
flux-core-python38-0.41.0-4.t4.x86_64
flux-pam-0.1.0-1.t4.x86_64
flux-sched-0.23.0-1.t4.x86_64
flux-security-0.7.0-1.t4.x86_64
[faaland1@fluke6 ~] $flux resource status
    STATUS NNODES NODELIST
     avail      8 fluke[6-13]

### fluke13 is powered off here
2022-07-19T16:58:49.000465Z broker.err[3]: broker on fluke13 (rank 7) has been unresponsive for 31.9998s

[faaland1@fluke6 ~] $flux resource status
    STATUS NNODES NODELIST
     avail      7 fluke[6-12]
   drained      1 fluke13
[faaland1@fluke6 ~] $flux mini run --ntasks 28 hostname | sort | uniq -c
      4 fluke10
      4 fluke11
      4 fluke12
      4 fluke6
      4 fluke7
      4 fluke8
      4 fluke9
[...]

Issue #4589, opened 2022-09-19
child instance should be able to notify parent of critical ranks


grondo:
A first step to child instance resiliency is described in #4573, where the job-exec feature to terminate a job on node failure can be completely disabled. However, in reality some ranks are critical in a child instance, obviously rank 0 and router brokers if there are any. Therefore there should be a way for a child instance to send the parent a list of ranks on which a failure should be considered a fatal error. (Thanks @garlick for this idea, and a reminder of it on the coffee call.)

Issue #4581, opened 2022-09-19
child instance occasionally does not exit cleanly after node failure


grondo:
When experimenting with a solution for #4573, it was found that a child instance which was allowed to continue after a node failure would stay in the RUN state even after the initial program has exited. In order for resilient child instances to be useful, they must exit cleanly (with the exit code of the initial program most likely) even after node failure. This will allow the parent instance to fully hand over fault tolerance to the child without a potential hang.

PR #4615, opened 2022-09-26
make child instances resilient to non-critical node failures

grondo:
This is a WIP, more of an experiment or proof-of-concept really, to address #4417 and #4573. (So no need for code review just yet, but looking for comments on the general approach.)

I was able to run down what was causing instances started with the experimental `exec.ignore-lost-ranks` to hang around after the rank 0 broker exits. It turns out the Flux instance was exiting normally, but the `output` plugin in the rank 0 job shell was not releasing its "completion reference" because it was still waiting for EOF from the lost shell rank.

This PR adds support for notifying the rank 0 job shell that a shell has been lost (actually through a couple different methods), along with an experimental `shell.lost` plugin callback, which the `output` plugin uses to decrement its internal refcount since it won't need to wait for that job shell anymore.

A new `exception` job shell plugin handles waiting for the appropriate exception and calling the `shell.lost` plugin callbacks. At first this plugin watched the job eventlog for the exception, but this had an impact on job throughput since it adds a new eventlog watcher for every job even though this particular exception will be very rare.

The second attempt adds an `exception` shell service endpoint to the `exception` plugin via which the job-exec module can notify the rank 0 shell of the non-fatal exception. This obviously has much better performance, but in order for that to work, the shell service handler had to be extended to allow messages from the instance owner as well as from the job owner.

I've kept these two solutions separate in the commit history for now until we've settled on the right approach.

I also think @garlick's suggestion of having the broker in the child notify the job-exec service of "critical ranks" is a superior method. In fact, maybe this should just be the default for all instances and users could select the level of resiliency by selecting a larger fanout or fanout=0 for the maximum resiliency? This will take a little longer to implement, and we may want to get at least basic support into the next release which is why I stopped here.

The following script is how I'm testing this feature:

```sh
#!/bin/bash
#
# 

log ()
{
    printf "free-range-test: $@"
}

list_descendants ()
{
    local children=$(ps -o pid= --ppid "$1")

    for pid in $children; do
        list_descendants "$pid"
    done

    echo "$children"
}


#  Check if we need to start parent job, if so, reexec under flux-start:
if test "$FREE_RANGE_TEST_ACTIVE" != "t"; then
    export FREE_RANGE_TEST_ACTIVE=t
    log "Re-launching test script under flux-start\n"
    exec flux start -s 4 --test-exit-mode=leader -o -Stbon.fanout=0 $0
fi

#  Start a free range job
log "Starting a free range child instance\n"
jobid=$(flux mini alloc -N4 --bg \
	--setattr=exec.ignore-lost-ranks \
	--broker-opts=-Stbon.fanout=0)

log "Started job $jobid\n"

#  Run a job on all ranks of child job
log "Current overlay status of $jobid:\n"
flux proxy $jobid flux overlay status 

log "Launch a sleep job within $jobid:\n"
flux proxy $jobid flux mini submit -N4 sleep inf

flux pstree -x --skip-root=no

#  Now simulate a node failure by killing a broker in parent
#  instance, along with all its children
broker_pid=$(flux exec -r 3 flux getattr broker.pid)

log "Killing rank 3 (pid %d) and all children\n" $broker_pid
kill -9 $(list_descendants $broker_pid) $broker_pid

log "Wait for exception event in $jobid\n"
flux job wait-event -t 5 $jobid exception

log "But running a 3 node job in $jobid still works:\n"
flux proxy $jobid flux mini run -t 5s -N3 hostname

log "Overlay status of $jobid should show rank offline:\n"
flux proxy $jobid flux overlay status 

log "Call flux shutdown on $jobid\n"
flux shutdown --quiet $jobid

log "job $jobid should exit cleanly (no hang) and a zero exit code:\n"
flux job wait-event -t 2 $jobid finish

log "dump output from job:\n\n"
flux job attach $jobid

printf "\n"
log "flux-job attach exited with code=$?\n"
```
E.g. on this branch:

```console
$ src/cmd/flux ./free-range-test.sh 
free-range-test: Re-launching test script under flux-start
free-range-test: Starting a free range child instance
free-range-test: Started job ƒjFeXq1
free-range-test: Current overlay status of ƒjFeXq1:
0 asp: full
├─ 1 asp: full
├─ 2 asp: full
└─ 3 asp: full
free-range-test: Launch a sleep job within ƒjFeXq1:
ƒw4ap2K
           . grondo    R      4      4   5.556s .
     ƒjFeXq1 grondo    R      4      4   3.198s └── flux
     ƒw4ap2K grondo    R      4      4   0.260s     └── sleep
free-range-test: Killing rank 3 (pid 4765) and all children
free-range-test: Wait for exception event in ƒjFeXq1
flux-start: 3 (pid 4765) Killed
2022-09-26T00:37:05.183617Z broker.err[0]: asp (rank 3) transitioning to LOST due to EHOSTUNREACH error on send
1664152625.184043 exception type="node-failure" severity=2 userid=4294967295 note="lost contact with job shell on asp (shell rank 3)"
free-range-test: But running a 3 node job in ƒjFeXq1 still works:
asp
asp
asp
free-range-test: Overlay status of ƒjFeXq1 should show rank offline:
0 asp: degraded
├─ 1 asp: full
├─ 2 asp: full
└─ 3 asp: lost
free-range-test: Call flux shutdown on ƒjFeXq1
free-range-test: job ƒjFeXq1 should exit cleanly (no hang) and a zero exit code:
1664152632.296247 finish status=0
free-range-test: dump output from job:

4.447s: job.exception type=node-failure severity=2 lost contact with job shell on asp (shell rank 3)
2022-09-26T00:37:05.666283Z broker.err[0]: asp (rank 3) transitioning to LOST due to EHOSTUNREACH error on send

free-range-test: flux-job attach exited with code=0
```

### DIFF
--- a/doc/man7/flux-broker-attributes.rst
+++ b/doc/man7/flux-broker-attributes.rst
@@ -89,6 +89,13 @@ broker.mapping
    the instance has one broker per node on 16 nodes, and ``(vector,(0,1,16))``
    means it has 16 brokers on one node.
 
+broker.critical-ranks [Updates: C]
+   An RFC 22 idset representing broker ranks that are considered critical
+   to instance operation. The broker notifies the job execution system in
+   the parent instance of these ranks such that a fatal job exception
+   is raised when a failing node or other error occurs affecting any rank
+   in this set. Default: rank 0 plus any other overlay network routers.
+
 broker.pid
    The process id of the local broker.
 
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -104,7 +104,7 @@ static void init_attrs_starttime (attr_t *attrs, double starttime);
 
 static int init_local_uri_attr (struct overlay *ov, attr_t *attrs);
 
-static int set_uri_job_memo (attr_t *attrs);
+static int execute_parental_notifications (struct broker *ctx);
 
 static const struct flux_handle_ops broker_handle_ops;
 
@@ -397,7 +397,7 @@ int main (int argc, char *argv[])
     if (init_local_uri_attr (ctx.overlay, ctx.attrs) < 0) // used by runat
         goto cleanup;
 
-    if (ctx.rank == 0 && set_uri_job_memo (ctx.attrs) < 0)
+    if (ctx.rank == 0 && execute_parental_notifications (&ctx) < 0)
         goto cleanup;
 
     if (create_runat_phases (&ctx) < 0)
@@ -921,39 +921,22 @@ static int init_local_uri_attr (struct overlay *ov, attr_t *attrs)
     return 0;
 }
 
-static int set_uri_job_memo (attr_t *attrs)
+static flux_future_t *set_uri_job_memo (flux_t *h,
+                                        flux_jobid_t id,
+                                        attr_t *attrs)
 {
-    const char *jobid = NULL;
-    const char *parent_uri = NULL;
     const char *local_uri = NULL;
     const char *path;
     char uri [1024];
     char hostname [MAXHOSTNAMELEN + 1];
-    flux_jobid_t id;
-    flux_t *h;
-    flux_future_t *f;
-    int rc = -1;
 
-    /* Skip if "jobid" or "parent-uri" not set, this is probably
-     *  not a child of any Flux instance.
-     */
-    if (attr_get (attrs, "parent-uri", &parent_uri, NULL) < 0
-        || parent_uri == NULL
-        || attr_get (attrs, "jobid", &jobid, NULL) < 0
-        || jobid == NULL)
-        return 0;
-
-    if (flux_job_id_parse (jobid, &id) < 0) {
-        log_err ("Unable to parse jobid attribute '%s'", jobid);
-        return -1;
-    }
     if (attr_get (attrs, "local-uri", &local_uri, NULL) < 0) {
         log_err ("Unexpectedly unable to fetch local-uri attribute");
-        return -1;
+        return NULL;
     }
     if (gethostname (hostname, sizeof (hostname)) < 0) {
         log_err ("gethostname failure");
-        return -1;
+        return NULL;
     }
     path = local_uri + 8; /* forward past "local://" */
     if (snprintf (uri,
@@ -961,31 +944,61 @@ static int set_uri_job_memo (attr_t *attrs)
                  "ssh://%s%s",
                  hostname, path) >= sizeof (uri)) {
         log_msg ("buffer overflow while checking local-uri");
+        return NULL;
+    }
+    return flux_rpc_pack (h,
+                          "job-manager.memo",
+                          FLUX_NODEID_ANY,
+                          0,
+                          "{s:I s:{s:s}}",
+                          "id", id,
+                          "memo",
+                            "uri", uri);
+}
+
+static int execute_parental_notifications (struct broker *ctx)
+{
+    const char *jobid = NULL;
+    const char *parent_uri = NULL;
+    flux_jobid_t id;
+    flux_t *h = NULL;
+    flux_future_t *f = NULL;
+    int rc;
+
+    /* Skip if "jobid" or "parent-uri" not set, this is probably
+     *  not a child of any Flux instance.
+     */
+    if (attr_get (ctx->attrs, "parent-uri", &parent_uri, NULL) < 0
+        || parent_uri == NULL
+        || attr_get (ctx->attrs, "jobid", &jobid, NULL) < 0
+        || jobid == NULL)
+        return 0;
+
+    if (flux_job_id_parse (jobid, &id) < 0) {
+        log_err ("Unable to parse jobid attribute '%s'", jobid);
         return -1;
     }
 
-    /*  Open connection to parent instance and post "uri" user annotation
+    /*  Open connection to parent instance:
      */
     if (!(h = flux_open (parent_uri, 0))) {
         log_err ("flux_open to parent failed");
         return -1;
     }
-    if (!(f = flux_rpc_pack (h,
-                             "job-manager.memo",
-                             FLUX_NODEID_ANY,
-                             0,
-                             "{s:I s:{s:s}}",
-                             "id", id,
-                             "memo",
-                               "uri", uri))
-        || flux_rpc_get (f, NULL) < 0) {
+
+    /*  Perform any RPCs to parent in parallel */
+    if (!(f = set_uri_job_memo (h, id, ctx->attrs)))
+        goto out;
+
+    /*  Wait for RPC results */
+    if (flux_future_get (f, NULL) < 0) {
         log_err ("job-manager.memo uri");
         goto out;
     }
     rc = 0;
 out:
-    flux_future_destroy (f);
     flux_close (h);
+    flux_future_destroy (f);
     return rc;
 }
 
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -104,6 +104,8 @@ static void init_attrs_starttime (attr_t *attrs, double starttime);
 
 static int init_local_uri_attr (struct overlay *ov, attr_t *attrs);
 
+static int init_critical_ranks_attr (struct overlay *ov, attr_t *attrs);
+
 static int execute_parental_notifications (struct broker *ctx);
 
 static const struct flux_handle_ops broker_handle_ops;
@@ -394,7 +396,8 @@ int main (int argc, char *argv[])
 
     set_proctitle (ctx.rank);
 
-    if (init_local_uri_attr (ctx.overlay, ctx.attrs) < 0) // used by runat
+    if (init_local_uri_attr (ctx.overlay, ctx.attrs) < 0 // used by runat
+        || init_critical_ranks_attr (ctx.overlay, ctx.attrs) < 0)
         goto cleanup;
 
     if (ctx.rank == 0 && execute_parental_notifications (&ctx) < 0)
@@ -919,6 +922,49 @@ static int init_local_uri_attr (struct overlay *ov, attr_t *attrs)
         }
     }
     return 0;
+}
+
+static int init_critical_ranks_attr (struct overlay *ov, attr_t *attrs)
+{
+    int rc = -1;
+    const char *val;
+    char *ranks = NULL;
+    struct idset *critical_ranks = NULL;
+
+    if (attr_get (attrs, "broker.critical-ranks", &val, NULL) < 0) {
+        if (!(critical_ranks = overlay_get_default_critical_ranks (ov))
+            || !(ranks = idset_encode (critical_ranks, IDSET_FLAG_RANGE))) {
+            log_err ("unable to calculate critical-ranks attribute");
+            goto out;
+        }
+        if (attr_add (attrs,
+                      "broker.critical-ranks",
+                      ranks,
+                      FLUX_ATTRFLAG_IMMUTABLE) < 0) {
+            log_err ("attr_add critical_ranks");
+            goto out;
+        }
+    }
+    else {
+        if (!(critical_ranks = idset_decode (val))
+            || idset_last (critical_ranks) >= overlay_get_size (ov)) {
+            log_msg ("invalid value for broker.critical-ranks='%s'", val);
+            goto out;
+        }
+        /*  Need to set immutable flag when attr set on command line
+         */
+        if (attr_set_flags (attrs,
+                            "broker.critical-ranks",
+                            FLUX_ATTRFLAG_IMMUTABLE) < 0) {
+            log_err ("failed to make broker.criitcal-ranks attr immutable");
+            goto out;
+        }
+    }
+    rc = 0;
+out:
+    idset_destroy (critical_ranks);
+    free (ranks);
+    return rc;
 }
 
 static flux_future_t *set_uri_job_memo (flux_t *h,
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -38,6 +38,7 @@
 #include "src/common/libidset/idset.h"
 #include "src/common/libutil/ipaddr.h"
 #include "src/common/libpmi/pmi.h"
+#include "src/common/libpmi/clique.h"
 #include "src/common/libpmi/pmi_strerror.h"
 #include "src/common/libutil/fsd.h"
 #include "src/common/libutil/errno_safe.h"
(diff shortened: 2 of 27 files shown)


grondo:
> I also think @garlick's suggestion of having the broker in the child notify the job-exec service of "critical ranks" is a superior method. In fact, maybe this should just be the default for all instances and users could select the level of resiliency by selecting a larger fanout or fanout=0 for the maximum resiliency

I've reworked this PR to use this approach instead of the previous experimental solutions.

The PR now puts the following changes together to make _all_ child jobs resilient to some extent:

 * Add a set of critical ranks for jobs kept by the `job-exec` module, which may be updated by _any_ job.
 * Allow the rank 0 broker of instances with a parent to calculate the set of critical ranks and update the job-exec module of the parent.
 * Add a shell `exception` plugin which adds a shell `exception` RPC through which the parent instance can notify the job shell of exceptions directly. (Note: this could also be done by having the job shell watch the job eventlog, but I found an performance impact here since every job will add an eventlog watcher for the rare case of a nonfatal exception) This is used to decrement references held by a lost shell (e.g. output EOF).
 * Do not raise a fatal job exception in the job-exec module when a lost job shell is not in the critical ranks idset. Instead notify the rank 0 job shell of the condition.

When this all in place, all child instances now survive a loss of a non-critical rank by default. Therefore, to get maximum resiliency a user need only set `tbon.fanout=0` to minimize critical ranks. (This could still be part of a `--resiliency` option or similar for `flux mini batch` and `flux mini alloc` in a future PR).

grondo:
I've updated `flux-broker-attributes(7)` and removed a couple superfluous commits (related to experiments 1 and 2) and removed WIP. This is probably ready for a review. (There was a lot of rebasing away of previous attempts, so hopefully history is not too messy)

src/modules/job-exec/job-exec.c
```diff
@@ -158,6 +158,32 @@ static struct jobinfo * jobinfo_new (void)
     return job;
 }
 
+flux_future_t *jobinfo_shell_rpc_pack (struct jobinfo *job,
```
chu11, line 162: commit message typo "send and RPC" -> "send an RPC"

src/modules/job-exec/job-exec.c
```diff
@@ -158,6 +158,32 @@ static struct jobinfo * jobinfo_new (void)
...
+                                       const char *topic,
```
chu11, line 163: commit message "Add jobinfo_shell_rpc_pack() as a wrapper around flux_rpc_pack() which sends to a shell topic on the first rank of a job."  Not wrong, but reads funny to me, maybe a little run on?

"flux_rpc_pack().  Function will send an RPC to the designated shell topic on the first rank of a job."?

src/modules/job-exec/exec.c
```diff
@@ -163,12 +163,53 @@ static void output_cb (struct bulk_exec *exec, flux_subprocess_t *p,
...
+                                      "type", "lost-shell",
+                                      "severity", 2,
+                                      "shell_rank", shell_rank)))
+            return -1;
+    if (flux_future_then (f, -1., lost_shell_continuation, job) < 0) {
+        flux_future_destroy (f);
+        return -1;
+    }
+    return 0;
+}
```
chu11, line 204: instead of hard coded 2, perhaps define some "NON_FATAL_SEVERITY" macro or something?

src/broker/broker.c
```diff
@@ -104,7 +104,7 @@ static void init_attrs_starttime (attr_t *attrs, double starttime);
 
 static int init_local_uri_attr (struct overlay *ov, attr_t *attrs);
 
-static int set_uri_job_memo (attr_t *attrs);
+static int execute_parental_notifications (struct broker *ctx);
```
chu11, line 110: commit message typo "witht" -> "with"

src/broker/broker.c
```diff
@@ -104,6 +104,8 @@ static void init_attrs_starttime (attr_t *attrs, double starttime);
 
 static int init_local_uri_attr (struct overlay *ov, attr_t *attrs);
 
+static int init_critical_ranks_attr (struct overlay *ov, attr_t *attrs);
```
chu11, line 108: commit message typo "queryry" -> "query"

src/broker/broker.c
```diff
@@ -38,6 +38,7 @@
 #include "src/common/libidset/idset.h"
 #include "src/common/libutil/ipaddr.h"
 #include "src/common/libpmi/pmi.h"
+#include "src/common/libpmi/clique.h"
```
chu11, line 41: `fixup` left over in commit message

chu11 (commented):
just did a high level skim to try and get the gist of this review first, found a bunch of nits

Wondering, should some of the fixes be split off?  It seems atleast the fixes for #4581 could be in its own PR?   #4573 and #4589 maybe are a bit intertwined to be split out.

grondo:
> Wondering, should some of the fixes be split off? It seems atleast the fixes for #4581 could be in its own PR? #4573 and #4589 mayb eare a bit intertwined to be split out.

I can see your point, but I'd rather not split them out (and #4581 would be difficult to test standalone, and this PR would be broken without it)

src/modules/job-exec/exec.c
grondo: I'm not sure we've standardized where to put that? Perhaps the numbering should just follow syslog severities? That feels like a whole other issue/RFC etc though

grondo:
Thanks @chu11, great catches. I went ahead and fixed all the typos and force-pushed the result.

src/modules/job-exec/exec.c
chu11: yeah, I don't think we do, but it just felt weird to see the `2` in there all alone without context (had to verify it was severity).  Perhaps just "int non_fatal_severity = 2;` in the function to give it context?

grondo: Ok, good point. I've force pushed a change to define a `FLUX_JOB_EXCEPTION_CRIT` in that file to `2`. At least we'll be able to search the code for it if we ever define these in the API or somewhere more centralized.

garlick:
This is great!

This is just something to be aware of, not necessarily anything we need to deal with here, but one thing I have noticed is that the batch jobs have the default `tcp.user-timeout` of 20m or so, while my system instance tunes it down to 2s.  So when I power off a node, it takes a loooong time for the node to be marked as _lost_ in the batch job.  If I shut down before then, the job hangs on the way down and gets killed by the doom plugin.  I wonder if we should reduce the default from 20m to something less than the 30s doom timeout to close that window?

grondo:
Oh, good catch! I wonder if there is some sensible way for the child instance to watch for the node failure job exception in the parent and trigger a check for liveness (I think you had suggested that before). Watching every job's eventlog is probably too high impact to handle the relatively rare case, and non instance owner users can't seem to subscribe to `job-exception` events. Maybe the rank 0  broker should be able to subscribe to non-fatal exception notifications, either via the job-manager or the job-exec service?

Note: I'm also fine with tuning the default `tcp.user-timeout`, just thought I'd throw out the above idea too.

garlick:
I had a similar thought.  It does seem like a job should have access to its own exceptions somehow!

grondo:
Oh, also it would be useful if exceptions had a way to add other context besides a note.

garlick:
Hmm, if a non-critical node is shutdown (as in `sudo shutdown -h now`) instead of powered off, the job seems to always get killed immediately:
```
 garlick@picl0:~/proj/flux-core$ flux mini alloc -t10m --broker-opts=-Stbon.fanout=0  -N4 -q batch
ƒ(s=4,d=1) garlick@picl1:~/proj/flux-core$ flux getattr hostlist
picl[1-3,5]
ƒ(s=4,d=1) garlick@picl1:~/proj/flux-core$ 81.394s: job.exception type=exec severity=0 flux_job_event_watch_get: Function not implemented

2022-09-30T05:09:40.386926Z broker.err[0]: rc2.0: /bin/bash Hangup (rc=129) 76.1s
2022-09-30T05:09:40.389796Z broker.err[0]: picl5 (rank 3) transitioning to LOST due to EHOSTUNREACH error on send
[detached: session exiting]
```
This must be the shell watching the `guest.input` eventlog?   I tried the same thing with a batch instead of alloc and it failed in a slightly different way:
```
$ flux job attach ƒB2fZJVC6Py
0: stdout redirected to flux-ƒB2fZJVC6Py.out
0: stderr redirected to flux-ƒB2fZJVC6Py.out
35.887s: job.exception type=exec severity=0 flux_job_event_watch_get: Function not implemented
flux-job: task(s) exited with exit code 137
```
So maybe a different code path to investigate here.  Again, not necessarily in this PR.

grondo:
Ah, this might be because the broker is terminated or exits before the shell (presumably by systemd), so any outstanding RPCs from the shell get an error. It is a little confusing, though, how the shell is still able to generate a fatal exception with `flux_job_raise(3)` if the broker is at a point where it is rejecting messages, so I probably don't fully understand what is going on here :thinking:

grondo:
Force-pushed to resolve conflict

src/modules/job-exec/job-exec.c
```diff
@@ -158,6 +158,32 @@ static struct jobinfo * jobinfo_new (void)
...
+                  topic) < 0)
+        return NULL;
+    if ((rank = resource_set_nth_rank (job->R, 0)) == IDSET_INVALID_ID)
+        return NULL;
+    va_start (ap, fmt);
+    f = flux_rpc_vpack (job->ctx->h, shell_topic, rank, 0, fmt, ap);
+    va_end (ap);
+    flux_future_aux_set (f, "jobinfo", job, (flux_free_f) jobinfo_decref);
+    jobinfo_incref (job);
+    return f;
```
garlick, line 185: `shell_topic` is leaked

src/modules/job-exec/job-exec.c
```diff
@@ -1176,6 +1188,47 @@ static void exception_cb (flux_t *h, flux_msg_handler_t *mh,
...
+
+    if (!(job = zhashx_lookup (ctx->jobs, &id))) {
+        errno = ENOENT;
+        goto error;
+    }
+    if (flux_msg_get_userid (msg, &userid) < 0
+        || userid != job->userid) {
+        errno = EPERM;
+        goto error;
+    }
```
garlick, line 1217: Suggestion: use the higher level functions for access control, something like
```c
struct flux_msg_cred cred;

if (flux_request_unpack (msg, ...) < 0
    || flux_request_get_cred (msg, &cred) < 0)
    goto error;
// lookup job
if (flux_msg_cred_authorize (cred, job->userid) < 0)
    goto error;
```
That will let the instance owner in too, as came up elsewhere.

src/broker/overlay.c
```diff
@@ -1542,6 +1542,20 @@ const char *overlay_get_subtree_status (struct overlay *ov, int rank)
...
+    /* For now, return all internal ranks plus rank 0
+     */
+    if (!(ranks = topology_get_internal_ranks (ov->topo))
+        || idset_set (ranks, 0) < 0) {
+        idset_destroy (ranks);
+        return NULL;
+    }
+    return ranks;
+}
+
```
garlick, line 1558: Suggestion: take "critical" out of the name of this function since "what's critical" is to be configurable by the caller.  Maybe `overlay_get_nonleaf_ranks()`?

src/broker/broker.c
```diff
@@ -921,6 +924,41 @@ static int init_local_uri_attr (struct overlay *ov, attr_t *attrs)
...
+            goto out;
+        }
+    }
+    else {
+        if (!(critical_ranks = idset_decode (val))
+            || idset_last (critical_ranks) >= overlay_get_size (ov)) {
+            log_msg ("invalid value for broker.critical-ranks='%s'", val);
+            goto out;
+        }
+    }
```
garlick, line 955: Ugh so sorry about this broker attr api  :frowning_person: 

You'll want to set the IMMUTABLE flag on the attribute if it came from the command line so that people won't think it can be set on the fly.
```c
if (attr_set_flags (attrs, "broker.critical_ranks", FLUX_ATTRFLAG_IMMUTABLE) < 0)
    goto out;
```

doc/man7/flux-broker-attributes.rst
```diff
@@ -89,6 +89,13 @@ broker.mapping
    the instance has one broker per node on 16 nodes, and ``(vector,(0,1,16))``
    means it has 16 brokers on one node.
 
+broker.critical-ranks [Updates: C]
+   An RFC 22 idset representing broker ranks that are considered critical
+   to instance operation. The broker notifies the job execution system in
+   the parent instance of these ranks such that a fatal job exception
+   is raised when a failing node or other error occurs affecting any rank
+   in this set. Default: rank 0 plus any other internal ranks.
```
garlick, line 97: "internal" is used here without much context.  Would "overlay network routers" be any clearer?

garlick (approved):
All looks good!  Just a few minor suggestions.

src/broker/overlay.c
grondo: Good point, but the function will always return rank 0 even for the degenerate case of an instance of size=1. This is why I added `critical` instead of nonleaf or internal. Any other suggestions?

doc/man7/flux-broker-attributes.rst
grondo: "internal" is taken directly from tree data structure terminology, but yeah "overlay network router" is probably better here :+1:

src/modules/job-exec/job-exec.c
grondo: Thanks I had forgotten about this one.

What about just `flux_msg_authorize(msg, job->userid)`? Is there something that makes pulling out the `cred` required here? BTW, there is no reason I can think of that the instance owner would use this RPC, but it doesn't hurt to allow it.

src/modules/job-exec/job-exec.c
grondo: Good catch! Thanks!

src/broker/overlay.c
grondo: I'm not sure what to call this function, reading up a bit the terminology seems to indicate that a tree of size=1 is a leaf by itself, so I hesitate to use the term "nonleaf" ranks since that is technically incorrect.

Here's my thinking on keeping the name "critical". The set of critical nodes is a property of the overlay, if a "caller" or user of the overlay wants to change what is in the critical set then that should be a separate method of the overlay. We may also eventually allow the overlay to be hand-configured, in which case the critical set could be part of the configuration. In these cases, we still need a getter to return the critical set of ranks so could this function stay as-is and and perhaps the implementation changes based on the underlying overlay object properties?

garlick: How about `overlay_get_default_critical_ranks()`?  Then it communicates that these are not the final critical ranks.  However, this is an internal function so if you still prefer the original, I'm OK with that!

src/modules/job-exec/job-exec.c
garlick: Yup i think that's fine too - I just wasn't thinking about that one :-)

src/broker/overlay.c
grondo: I'm confused - where are the "final critical ranks"? I'm missing something...

garlick: Isn't this providing the default setting for the `broker.critical-ranks` attribute, which can be overridden on the command line?

grondo: Ah, ok, yes. But I was considering the `struct overlay` as a class which knows nothing about broker attributes, and the critical-ranks idset as an attribute of an instance of that class. The broker then queries the overlay for what it considers critical ranks if the attribute isn't set on the command line. I guess calling it "default" makes an assumption it is only being used for the one purpose, but that is true so I'm fine changing the name.

grondo:
I addressed @garlick's comments above and force-pushed the result with a rebase

src/broker/overlay.c
grondo: Ok, I changed the name of the function, squashed and force-pushed the result.

garlick: Meh, my suggestion was probably lame.  You pick.

grondo: OMG, this was my fault. I was somehow looking at the wrong code (the code added to overlay.c), so I apologize for all my confusing nonsense!

grondo:
Setting MWP.

grondo:
Restarted the test-install builder which seemed to hang in the `test_runat.t` unit test.

grondo:
hit #4596. Restarted that builder

PR #4616, opened 2022-09-26
ingest: require job queue if [queues] are configured

garlick:
Problem: if queues are configured, but no default queue is configured, and a job is submitted without a queue, the job is ingested and allowed to fail later.

Modify the 'defaults' frobnicator plugin to reject such a job at ingest.


grondo (approved):
LGTM!

Issue #4613, opened 2022-09-25
flux-jobs: should duration be used in RUNTIME column for non-running jobs?


garlick:
Problem: the RUNTIME column doesn't convey useful information for non-running jobs:
```
       JOBID QUEUE    USER     NAME       ST NTASKS NNODES  RUNTIME NODELIST
 ƒAnP2EUJgNB batch    garlick  sleep       S      4      4       0s -
 ƒAnNpvHtZ67 debug    garlick  stress      R      1      1   30.58s picl6
 ƒAnNnmJF73u batch    garlick  stress      R      2      2   35.48s picl[4-5]
 ƒAnNmGEhHDR batch    garlick  stress      R      2      2   38.87s picl[2-3]
 ƒAnNjh4TT1R batch    garlick  stress      R      1      1   42.45s picl1
```
Could we show the jobspec duration in this column when the job is pending?

Edit: if that's thought to be a good idea then maybe DURATION would be a better column heading?

grondo:
I think that seems like a reasonable idea. Note in the common, non-system-instance case a duration may not be set either so the column would still be empty.

I had the idea last week to make some contextual properties for the `JobInfo` class. I had prototyped an attribute to replace the NODELIST column that displays something like `priority-wait` for jobs in PRIORITY, the dependencies for jobs in DEPEND, any `sched.t_estimate` or `sched.reason_pending` if available in SCHED state, or otherwise the NODELIST.

We could do something similar here for RUNTIME/DURATION?

grondo:
Oh, one prerequisite here is that the `job-list` module capture the duration from jobspec, when set. Currently the duration doesn't appear to be an attribute provided by the service.

chu11:
> Edit: if that's thought to be a good idea then maybe DURATION would be a better column heading?

Since they provide two different pieces of information, I like the idea of keeping them separate just in case.

I'll throw out the idea that we simply use a new field name, perhaps generically named TIME, that can represent the RUNTIME/DURATION.  It'd be similar to the `status` field,which is a combination of job state and job result.

PR #4617, opened 2022-09-26
job-list: support retrieval of job duration

chu11:
Per comment in #4613, it would be useful to grab job duration in `job-list` / `flux-jobs`.

This is built on top of #4611, as `libjj` conveniently parses `duration` for us already.


t/t2260-job-list.t
```diff
@@ -797,6 +797,37 @@ test_expect_success HAVE_JQ 'flux job list lists nnodes for pending jobs correct
         flux queue start
 '
 
+test_expect_success 'reload the job-list module' '
```
grondo, line 800: Just a small thought: The commit message for this change just says tests aren't "where they should be", but it would be nice to know _why_ the tests should be in a different place, not just that they are in the wrong place.

doc/man1/flux-jobs.rst
```diff
@@ -187,7 +187,7 @@ the following conversion flags are supported by *flux-jobs*:
 
 **!F**
    convert a duration in floating point seconds to Flux Standard Duration (FSD).
-   string.  Defaults to empty string if duration field does not exist.
```
grondo, line 190: My guess is that this was supposed to say something like "to a Flux Standard Duration (FSD) string" since it is important to know the conversion results in a string, not a number or other object that has special format specifiers. Definitely not a big deal, but it would be nice to keep the emphasis that the result of the conversion is a string.

grondo (approved):
Nice! Just a couple very minor comments inline.

chu11:
re-pushed tweaks per comments above, setting MWP

doc/man1/flux-jobs.rst
grondo: Yeah, also a missing `a` or something. Definitely needed a fix.

PR #4621, opened 2022-09-27
flux-shell: always use pmi=pershell by default

### LINKED ISSUE #4619: broker.mapping (PMI_process_mapping) values are incorrect
@grondo noted the following:
```
 grondo@fluke108:~$  flux mini run -N1 --tasks-per-node=4 flux start flux mini run -N4 flux start flux getattr broker.mapping
(vector,(0,1,4))
 grondo@fluke108:~$  flux mini run -N2 --tasks-per-node=4 flux start flux mini run -N4 flux start flux getattr broker.mapping
(vector,(0,4,1))
```
Why 1 node vs 4 nodes?  Shouldn't they both be 4 nodes?

The parent mappings look correct:
```
 grondo@fluke108:~$  flux mini run -N1 --tasks-per-node=4 flux start flux getattr broker.mapping
(vector,(0,1,4))
 grondo@fluke108:~$  flux mini run -N2 --tasks-per-node=4 flux start flux getattr broker.mapping
(vector,(0,2,4))
```
Note: There was a time when we needed the mapping to show the "true" node mapping so that openmpi wouldn't assume the same shared memory path should work on ranks that are not co-located in a clique.  But that was deemed unnecessary  in #3673.

See also [RFC 13](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_13.html#id14)

garlick:
Problem: when a Flux instance has multiple brokers on one node, PMI cliques are generated using "single" mode, but when an instance has multiple brokers on multiple nodes, cliques are generated using "pershell" mode.  Propagating "single" through sub-instances causes information to be lost in the broker.mapping attribute, rendering it useless for determining the hierarchical relationships between brokers.

Always use "pershell" (unless overridden).

Fixes #4619

### DIFF
--- a/src/shell/pmi/pmi.c
+++ b/src/shell/pmi/pmi.c
@@ -375,17 +375,12 @@ static void pmi_fd_cb (flux_shell_task_t *task,
     }
 }
 
-/* Query broker to see if instance mapping is known, then use that information
- * to select whether process mapping should be "none", "single", or "pershell".
- * If the instance mapping is unknown, use "pershell".  The choice of default
- * was a process of trial and error:
+/* Generate 'PMI_process_mapping' key (see RFC 13) for MPI clique computation.
  *
  * PMI_process_mapping originated with MPICH, which uses it to determine
  * whether it can short circult the comms path between local ranks with shmem.
  * MPICH allows the key to be missing or its value to be empty, and in those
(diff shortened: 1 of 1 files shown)


grondo (approved):
Great! LGTM!

Issue #4612, opened 2022-09-25
job-info: broker crash when watched eventlog is replaced


grondo:
I hit a broker segfault in the testsuite while testing a shell plugin that placed a watch on the job eventlog.
I ran the problem down to a test in  `./t2232-job-info-security.t` that rewrites a running job's eventlog -- standalone reproducer below:

```sh
#!/bin/bash
jobid=$(flux mini submit --wait-event=start sleep 300)
kvsdir=$(flux job id --to=kvs $jobid)

flux job attach $jobid &

flux kvs get --raw ${kvsdir}.eventlog \
	| sed -e s/submit/foobar/ \
	| flux kvs put --raw ${kvsdir}.eventlog=-

wait
```

```console
$ src/cmd/flux start --wrap=libtool,e,valgrind ./test-eventlog-rewrite.sh
==30006== Memcheck, a memory error detector
==30006== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==30006== Using Valgrind-3.13.0 and LibVEX; rerun with -h for copyright info
==30006== Command: /home/grondo/git/f.git/src/broker/.libs/flux-broker ./test-die.sh
==30006== 
hwloc x86 backend cannot work under Valgrind, disabling.
==30006== Thread 12:
==30006== Invalid read of size 1
==30006==    at 0x61A5E60: __strchr_avx2 (strchr-avx2.S:45)
==30006==    by 0x9F94CAC: eventlog_parse_next (watch.c:197)
==30006==    by 0x9F94CAC: watch_continuation (watch.c:268)
==30006==    by 0x4E8B9D2: ev_invoke_pending (ev.c:3770)
==30006==    by 0x4E8F728: ev_run (ev.c:4190)
==30006==    by 0x4E5BA3A: flux_reactor_run (reactor.c:128)
==30006==    by 0x9F93BDE: mod_main (job-info.c:136)
==30006==    by 0x1180B8: module_thread (module.c:183)
==30006==    by 0x50D36DA: start_thread (pthread_create.c:463)
==30006==    by 0x613961E: clone (clone.S:95)
==30006==  Address 0x0 is not stack'd, malloc'd or (recently) free'd
==30006== 
==30006== 
==30006== Process terminating with default action of signal 11 (SIGSEGV)
==30006==  Access not within mapped region at address 0x0
==30006==    at 0x61A5E60: __strchr_avx2 (strchr-avx2.S:45)
[...]

PR #4624, opened 2022-09-27
job-info: return error on invalid eventlog append

chu11:
Problem: If an eventlog is illegally overwritten, it can result in the appearance of a "zero length" append.  If a user is watching
this eventlog, this "zero length" append can result in a segfault in the job-info module.

Solution: If a "zero length" append occurs in an eventlog being watched, consider the eventlog malformed and return an error.

Issue https://github.com/flux-framework/flux-core/issues/4612

Side notes: 

EINVAL the best error to return?  I considered EPROTO as well, but that didn't seem quite right.  Open to suggestions.

### DIFF
--- a/src/modules/job-info/watch.c
+++ b/src/modules/job-info/watch.c
@@ -246,10 +246,21 @@ static void watch_continuation (flux_future_t *f, void *arg)
     const char *input;
     const char *tok;
     size_t toklen;
+    const char *errmsg = NULL;
 
     if (flux_kvs_lookup_get (f, &s) < 0) {
         if (errno != ENOENT && errno != ENODATA && errno != ENOTSUP)
             flux_log_error (ctx->h, "%s: flux_kvs_lookup_get", __FUNCTION__);
+        goto error;
+    }
+
+    /* Issue #4612 - zero length append illegal for an eventlog.  This
+     * most likely occurred through an illegal overwrite of the whole
+     * eventlog.
+     */
+    if (!s) {
+        errmsg = "illegal append of zero bytes";
+        errno = EINVAL;
         goto error;
     }
 
@@ -310,7 +321,7 @@ error_cancel:
     }
 
 error:
-    if (flux_respond_error (ctx->h, w->msg, errno, NULL) < 0)
+    if (flux_respond_error (ctx->h, w->msg, errno, errmsg) < 0)
         flux_log_error (ctx->h, "%s: flux_respond_error", __FUNCTION__);
 
     /* flux future destroyed in watch_ctx_destroy, which is called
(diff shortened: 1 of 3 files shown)


grondo:
> EINVAL the best error to return? 

`flux_respond_error()` allows an error message to be sent in addition to an error number, how about adding one of those to differentiate this error from other EINVAL errors at least.

grondo:
While you're thinking about this code, it might also be a good time to consider what else a user might do in the guest namespace that could cause the broker some trouble. Just an idle thought, because I can't think of anything specific.

chu11:
> While you're thinking about this code, it might also be a good time to consider what else a user might do in the guest namespace that could cause the broker some trouble. Just an idle thought, because I can't think of anything specific.

Yeah, I couldn't think of anything either.  If an overwrite of the eventlog was short on bytes, there would already be an error sent (via `kvs-watch`, b/c it can't handle appends that "shorten" the value).  If an overwrite of the eventlog was long on bytes, then then extraneous extra data would be forwarded on, and the receiver would have to handle that data accordingly (most likely parse error of bad data).

I began thinking about what if someone injected rogue events.  Like what if:

`flux kvs put --append ${kvsdir}.eventlog` `{"timestamp":1234, "name":"clean"}`

i.e. inject a `clean` event. as the last event no matter what state the job is in (skip finish, release, free states).  

Now that I think about, such "rogue event insertions" may be worthwhile to add for testing, to see how well we protect against this stuff in code.  But I probably wouldn't put as high priority.  There's a mountain of permutation we could potentially cover.

chu11:
re-pushed with a small tweak, adding an errmsg to the EINVAL that is returned.

grondo:
> i.e. inject a clean event. as the last event no matter what state the job is in (skip finish, release, free states).

A guest can't write to the main eventlog so probably not as important an issue. What about invalid JSON  or invalid eventlog entries and garbage? I guess we're dependending on Jansson to handle those for us?

chu11:
> A guest can't write to the main eventlog so probably not as important an issue. What about invalid JSON or invalid eventlog entries and garbage? I guess we're dependending on Jansson to handle those for us?

Right now `job-info` doesn't deal with invalid eventlog entries so they would be passed up to callers and handle by things like `eventlog_entry_decode()`.  I did a quick test last night and it appeared to be handled gracefully.

chu11:
hmmm, regression test is always failing in inception CI.

chu11:
re-pushed, I think the backgrounded process was racing with foreground process in the reproducer.  So tweaked the regression test to eliminate the racing.

src/modules/job-info/watch.c
```diff
@@ -246,13 +246,24 @@ static void watch_continuation (flux_future_t *f, void *arg)
     const char *input;
     const char *tok;
     size_t toklen;
+    const char *errmsg = NULL;
```
grondo, line 249: nit: was the commit message supposed to have "Fixes #4612" not just "Issue #4612"?

grondo (approved):
LGTM!

src/modules/job-info/watch.c
chu11: oops, yeah

chu11:
re-pushed with that minor commit message fixes, setting MWP

PR #4626, opened 2022-09-29
WIP: python: add contextual_info jobinfo field

grondo:
This PR adds a new `contextual_info` property to `JobInfo` which returns different data depending on what state the job is in. In PRIORITY it returns the arbitrary string `priority-wait` (somewhat redundant to the job state, but could be helpful nonetheless), if the job is in DEPEND, it returns `depends:{dependencies}`, if the job is in SCHED it returns `eta:` + time until `sched.t_estimate` if `sched.t_estimate` is set (which I guess it isn't anywhere so oh well), and if in RUN or later then the nodelist is returned.

This new property is meant to replace `{nodelist}` in the default `flux-jobs` output. If this works out we can add other `contextual_*` properties (e.g. `time` which returns `duration` until the job is running as proposed in #4613)

I'm just posting a kind of working prototype here for discussion. Silly example:

```
       JOBID USER     NAME       ST NTASKS NNODES  RUNTIME INFO|NODELIST
  ƒ3TJG44bks grondo   hostname    S      1      -       0s 
  ƒ3TJPADA6j grondo   hostname    D      1      -       0s depends:after-finish=ƒ3TJ7micT9
  ƒ3TJW31puV grondo   hostname    D      1      -       0s depends:after-success=ƒ3TJPADA6j
  ƒ3TJ7micT9 grondo   sleep       R     16      4   1.106s asp,asp,asp,asp
```


src/bindings/python/flux/job/info.py
```diff
@@ -349,6 +349,75 @@ def returncode(self):
             code = os.WEXITSTATUS(status)
         return code
 
+def job_fields_to_attrs(fields):
+    # Note there is no attr for "id", its always returned
```
chu11, line 377: lol, my fault from original code "its" -> "it's" :-)

src/cmd/flux-jobs.py
```diff
@@ -34,7 +34,7 @@ class FluxJobsConfig(UtilConfig):
             "format": (
                 "{id.f58:>12} ?:{queue:<8.8} {username:<8.8} {name:<10.10} "
                 "{status_abbrev:>2.2} {ntasks:>6} {nnodes:>6h} "
-                "{runtime!F:>8} {nodelist:h}"
+                "{runtime!F:>8} {contextual_info}"
```
chu11, line 37: don't forget to update manpage

chu11 (commented):
I think this is great!  Two small nits.

grondo:
Forced a push with tests and doc added. Also fixed a `UnicodeEncodeError` in flux-jobs that can occur when displaying dependencies

src/bindings/python/flux/job/info.py
```diff
@@ -349,6 +349,35 @@ def returncode(self):
             code = os.WEXITSTATUS(status)
         return code
 
+    @memoized_property
```
chu11, line 352: commit message typo "woudl" - "would"

doc/man1/flux-jobs.rst
```diff
@@ -382,6 +383,17 @@ include:
...
+the state of the job or other context:
+
+**contextual_info**
+   Returns selected information based on the job's current state.  If the
+   job is in PRIORITY state, then the string ``priority-wait`` is returned,
+   if the job is in DEPEND state, then a list of outstanding  dependencies
+   is returned, if the job is in SCHED state then an estimated time the
+   job will run is returned (if the scheduler supports it). Otherwise,
+   the assigned nodelist is returned (if resources were assigned).
+
```
chu11, line 396: idea: move `status` and `status_abbrev` into this area too?  fields that are "combo" fields?  There is already a `Job status` section, so not sure if it would be more confusing or not.

chu11 (commented):
high level comment, i'm sure you thought about it, but `contextual_info` isn't the greatest name for a field??

I almost want to suggest the super generic "job_info"? (although it may confuse developers with the module).

Or perhaps "info_nodelist", since that's the header of the field.  It's a nodelist, but provides info when nodelist isn't available.

grondo:
> high level comment, i'm sure you thought about it, but contextual_info isn't the greatest name for a field??

Well, I was thinking all the fields that change based on job state or other context could have the same prefix, e.g. `contextual_*`. This gives a hint to the user that the data displayed will not be the same for all jobs -- i.e. it is contextual. So the next one I'd propose is `contextual_time` that displays the job duration until the job is running in which case it will show the runtime. Maybe we can come up with another naming scheme for this.

doc/man1/flux-jobs.rst
grondo: I was thinking of contextual fields as displaying a different JobInfo property depending on job state or some other information. Since `status` is itself a JobInfo property, it doesn't fit into this category. (Does that even make sense?)

chu11: `status` is sort of the "decrepit" contextual field, show `state` except when state == inactive, show `result`.  It's obviously the most minimal contextual one.

And isn't `contextual_info` also a JobInfo property?  You also added it to the class.  Or did you mean a uhh "native" properties.

grondo: Yeah, I can see that it is actually a gray area. I was thinking of `status` as more well defined -- the job has a "status" (it is just a conglomeration of state or result), but it doesn't really have this new field which is just a convenience property for use in `flux-jobs` or other `JobInforFormat` use cases.

garlick:
This seems great!  The only note I have is that the header could be shortened, e.g. instead of "INFO|NODELIST", maybe just INFO would be fine?
```
 garlick@picl0:~/proj/flux-core/src/cmd$ ./flux jobs
       JOBID QUEUE    USER     NAME       ST NTASKS NNODES  RUNTIME INFO|NODELIST
 ƒBKHa1BV51q debug    garlick  sleep       S      1      -       0s 
 ƒBKHaMAqPPV debug    garlick  sleep       S      1      -       0s 
 ƒBKHaiWm3hR debug    garlick  sleep       S      1      -       0s 
 ƒBKHZcnmv9m debug    garlick  sleep       R      1      1   10.62s picl7
 ƒBKHZAk4XzP debug    garlick  sleep       R      1      1   11.63s picl6
```
When the QUEUE field is there, I'm finding the header is spilling over 80 columns as well.

grondo:
Good catch thanks! I fixed the header as suggested and I'll remove WIP.

chu11 (approved):
nothing caught my eye in previous reviews except for those nits

grondo:
Thanks! I'll set MWP.

PR #4634, opened 2022-10-01
consolidate python linting and formatting

### LINKED ISSUE #4631: Review Python linting strategy and possible automation

vsoch:
**update** this PR will be replaced by #4636 . It seems that isort and/or flake8 (to be determined!) actually broke all-the-things. We will want to do these changes incrementally (and in separate PRs) to determine the culprit! Closing here, will update the other PR with the notes here.

Problem: the developer experience is challenging because linting and formatting are hard to debug locally, and then the separation in the CI means 3+ clicks to see what is going on! To fix this, this changeset adds the use of pre-commit, which will always run linting/formatting (and with automated fixes for a subset of the tools) right before commit, so the developer can have some confidence that passing locally == passing in the CI. In the case that there is a bug in the CI, we will now just have one place to look instead of 3+. This should be a start of a new setup, including addition of pre-commit and consolidation of settings into a pyproject.toml (and setup.py for flake8) and we can further tweak as we establish our linting preferences. In summary:

- I've added pre-commit and a scripts/requirements.dev.txt to install all linting/ formatting deps
- I've added a developer.rst to the Python docs to show how to use this
- I've removed the custom linting shell scripts in scripts (in practice I found running them gave me different results than the CI, likely because of versioning)
- I've added a pyproject.toml with preferences for isort, mypi, and black
- The setup.cfg holds flake8 (does not support pyproject.toml I don't think)
- To get the linting to pass, I had to run isort and do fixes on several files (fixes here)
- To make the developer.rst complete, I included the instructions I use to develop Python (and removed the --without-python in the container).

Here it is in action! Note that it's running the checks just before I commit, so I can fix any issues before I do.

image

And pre-commit  is super cool - it's like being in a candy store! https://github.com/pre-commit/pre-commit-hooks
I also tested with rebase to ensure it works! This will close #4631 


vsoch:
Hit this error in setting up the environment:

```
      Complete output (6 lines):
        File "/home/runner/.cache/pre-commit/repokurmseju/setup.py", line 1
          from __future__ import annotations
          ^
      SyntaxError: future feature annotations is not defined
```
I read that it's not supported under 3.7, so I bumped it up.

vsoch:
If anyone has insight to why the docker-run-checks are failing, it would be helpful! The logs aren't super helpful:
```
Run cd .. && git clone https://github.com/flux-framework/flux-accounting && cd flux-accounting && src/test/docker/docker-run-checks.sh -j 4 
Cloning into 'flux-accounting'...
Building image el8 for user runner 1001 group=121
mounting /home/runner as /home/runner
mounting /home/runner/work/flux-core/flux-accounting as /usr/src
git fetch tags
build setup
autogen.sh
configure 
make clean...
make --output-sync=target --no-print-directory -j 4
make --output-sync=target --no-print-directory -j 4 check
docker-run-checks.sh: docker run failed
Error: Process completed with exit code 1.
```

https://github.com/flux-framework/flux-core/actions/runs/3163340737

grondo:
You can expand many of the sections you pasted above in the Github actions log window. The first thing I usually check is the check "annotations" which tries to show you the output of failed tests. For example here:

Screenshot from 2022-10-01 07-42-32

It looks like this PR broke all the Python commands. I have no idea why, there's a lot of churn introduced here

I would try running `make check` or `src/test/docker/docker-run-checks.sh`locally until tests pass which will give you better turnaround time and easier debugging than pushing to a PR to have CI run. You can pass `-i [el7|el8|focal|fedora34|fedora35]` to the script to make sure changes will work with all the distros we support. Once all those pass you can then push your branch to Github and hopefully have to debug less through the output logs.

vsoch:
The only changes of any significance (as the rest were linting) was trying to add exception types to two blocks, and I suspect that's what causing the issues. I'll try restoring those (and then figuring out how to handle for the linter) and hopefully that should work. I'm not sure the extent to which I can run the docker tests because I don't actually build flux on a local machine, always in a container. So I need to use the CI for testing that.

vsoch:
I removed the changes to exceptions, and the only other thing I can think of is either the import matter was important (yikes!) or the pin of testing with Python 3.6 for linting made some difference.

vsoch:
It would be nice if there was a clear command to just run Python tests instead of needing to figure out what is going on through the tangles of Makefile.am, etc. I'm not sure I could figure it out.

vsoch:
OK tried running the docker test locally - it failed because of the version issue.

vsoch:
I just don't know how to debug this :/

grondo:
> It would be nice if there was a clear command to just run Python tests instead of needing to figure out what is going on through the tangles of Makefile.am, etc. I'm not sure I could figure it out.

You can run individual tests by just running them. Run under `flux start` so that you ensure you are running tests against the build tree not the installed bindings:

```console
$ cd t
$ ../src/cmd/flux start
$ ./python/t0001-handle.py 
TAP version 13
ok 1 __main__.TestHandle.test_anonymous_handle_rpc_ping
ok 2 __main__.TestHandle.test_attr_get
ok 3 __main__.TestHandle.test_create_handle
ok 4 __main__.TestHandle.test_get_rank
ok 5 __main__.TestHandle.test_log
ok 6 __main__.TestHandle.test_rpc_null_payload
ok 7 __main__.TestHandle.test_rpc_ping
ok 8 __main__.TestHandle.test_rpc_ping_unicode
ok 9 __main__.TestHandle.test_rpc_with
1..9
```

In this case you could have also tried a basic check like
```
$ src/cmd/flux start flux mini run hostname
```

Since `flux-mini.py` is a Python script making use of the Python bindings, etc. This also means you can't just "run the Python tests" since a lot of the bindings are tested via the commands that utilize them, and all those commands and other interfaces are tested via the many other sharness tests under `t`.

grondo:
> OK tried running the docker test locally - it failed because of the version issue.

I apologize if I missed it, what is the version issue?

vsoch:
The bug that building from a commit doesn't work:

```
 #define FLUX_CORE_VERSION_MAJOR     af37d1164
                                     ^
version.c:25:18: note: in expansion of macro ‘FLUX_CORE_VERSION_MAJOR’
         *major = FLUX_CORE_VERSION_MAJOR;
                  ^~~~~~~~~~~~~~~~~~~~~~~
version.h:25:37: note: each undeclared identifier is reported only once for each function it appears in
 #define FLUX_CORE_VERSION_MAJOR     af37d1164
                                     ^
version.c:25:18: note: in expansion of macro ‘FLUX_CORE_VERSION_MAJOR’
         *major = FLUX_CORE_VERSION_MAJOR;
                  ^~~~~~~~~~~~~~~~~~~~~~~
version.c:29:41: error: expected expression before ‘;’ token
         *patch = FLUX_CORE_VERSION_PATCH;
                                         ^
```

vsoch:
I've been (sort of) successful to run with pytest in my development container, and I keep seeing:

```
        #  Instantiate implementation from version
        #  note: only version 1 supported for now
        if version == 1:
            self.version = 1
>           self.impl = Rlist(arg)
E           TypeError: 'module' object is not callable
```
for different objects. I'm worried that the import order matters (but why? where?) and isort has broken something because of that.

grondo:
Since the problem ostensibly being addressed here is that the python linting and formatting are difficult for developers to debug and implement locally, perhaps the PR could be pared back to avoid adding any new linters and formatters. This would avoid the need to change any of the Python source in the project (likely removing the introduced CI errors). Instead we could just experiment with the pre-commit hook and other great additions that automate trivial linting and formatting fixups. Then, at a later time we could address the addition of new linting tools which will make it much easier to see what is causing problems (and will have a better history for the release notes, etc.)

Does that make any sense?

vsoch:
yeah I was just thinking that - to test my theory I'm going to make "the same" pr but without isort, and I'm going to add exceptions for all the changes I made. Then if that doesn't work we know it's not that.

grondo:
> The bug that building from a commit doesn't work:

You can just fetch the tags to fix this (that is what is done in CI), or I believe we committed your fix to also accept `FLUX_VERSION`.

PR #4635, opened 2022-10-01
consolidate python linting and formatting

vsoch:
Problem: the developer experience is challenging because linting and formatting are hard to debug locally, and then the separation in the CI means 3+ clicks to see what is going on! To fix this, this changeset adds the use of pre-commit, which will always run linting/formatting (and with automated fixes for a subset of the tools) right before commit, so the developer can have some confidence that passing locally == passing in the CI. In the case that there is a bug in the CI, we will now just have one place to look instead of 3+. This should be a start of a new setup, including addition of pre-commit and consolidation of settings into a pyproject.toml (and setup.py for flake8) and we can further tweak as we establish our linting preferences.

This is the same as #4634 but without isort and flake8.

Actually - I just noticed files removed from a makefile.am - not sure how that happened - going to look into that too.


src/bindings/python/flux/Makefile.am
```diff
@@ -34,10 +34,6 @@ nobase_fluxpy_PYTHON = \
 	job/validator/plugins/schema.py \
 	job/validator/plugins/feasibility.py \
 	job/validator/plugins/require-instance.py \
-	job/frobnicator/__init__.py \
```
vsoch, line 37: Hmm this is the piece I don't understand - I must have messed something up with creating this second version to delete these lines.

vsoch:
Yeah I don't know what's going on here with the git history, I guess I need to start over.

vsoch:
okay going to close this one - I'm worried about the extra changes that I don't know where they came from, most likely my incompetence. :laughing:

Issue #4520, opened 2022-08-24
store redacted jobspec in kvs


garlick:
We currently store `J` and `jobspec` in the KVS for each job.  `J` includes the jobspec and other submission artifacts signed by the user.  `jobspec` is exactly as unwrapped from `J`.

The size of the jobspec is usually dominated by `attributes.system.environment`.  We've found it useful to pass around a "redacted jobspec" where the environment has been removed, because it is really only the job shell that requires access to this part of the jobspec.

Since the environment portion is already present in `J` it is somewhat redundant to include it in `jobspec` in the KVS, especially since it is not widely required.  This issue proposes to
-  remove the environment from `<jobid>.jobspec`
-  have the shell request `<jobid>.J` from job-info and get the jobspec from there rather than request `<jobid>.jobspec` directly
-  add a `flux job info --original jobspec` (or similar) option for fetching the unredacted jobspec on the command line

See also #4494 where we discuss moving `jobspec-default` functionality to `job-ingest`.  Presumably we can write the modified _and_ redacted jobspec to the KVS now that the original/signed one is being used where appropriate

We should also update [RFC 16](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_16.html) to track this change.

PR #4637, opened 2022-10-01
drop environment from KVS jobspec

garlick:
As discussed in #4520, this drops `attributes.system.environment` from the `jobspec` stored in the KVS.  The original jobspec is still available via `J` (signed).


grondo (approved):
LGTM!

garlick:
Thanks!  Setting MWP.

Issue #4440, opened 2022-07-27
job request should be rejected if it specifies an unknown queue


garlick:
Problem: currently a job may be submitted with `attributes.system.queue` set to a name not listed in the TOML queue configuration, and the job is accepted.

This was temporarily allowed to avoid breaking fluxion's `t1006-qmanager-multiqueue.t` test.  Fluxion has its own module command line queue configuration for queues that predates RFC 33.  This is proposed to be addressed in flux-framework/flux-sched#950.

`t2220-job-manager-jobspec-default.t` contains a test that current expects failure
```
test_expect_failure 'submit a job to unknown queue fails' '
        test_must_fail flux mini submit \
                --setattr=system.queue=pbatch /bin/true 2>submit.err &&
        grep "Error parsing pbatch queue policy" submit.err
'
```

garlick:
Since #4529 was merged, jobs submitted to an unknown queue are rejected, as long as the frobnicator is enabled.

However it might still be a good idea to reject a jobs submitted with unknown queue names when the frobnicator is not enabled (e.g. in a typical batch job).  So leaving this one open.

PR #4640, opened 2022-10-03
make queue state persist across instance restart

garlick:
This adds the queue enable/disable state to the job manager KVS checkpoint so that it persists across an instance restat.

This PR is currently based on #4440 (only the most recent 2 commits are part of this PR).


grondo (approved):
Nice! LGTM!

garlick:
Thanks!  Rebased on master after #4440 was merged, and now setting MWP.

PR #4641, opened 2022-10-03
python: add contextual_time jobinfo field

grondo:
This PR follows up the `contextual_info` with a `contextual_time` field. As discussed in #4613, it might be useful to display the job duration when available for pending jobs in the default `flux jobs` output. To allow that feature easily, a new contextual jobinfo property is introduced, `contextual_time`, which displays runtime for jobs that hit the RUN state, but job duration (if set) otherwise. The heading is simply `TIME` as proposed in that issue.

### DIFF
--- a/t/t2800-jobs-cmd.t
+++ b/t/t2800-jobs-cmd.t
@@ -571,6 +571,20 @@ test_expect_success 'flux-jobs --format={runtime:0.3f} works' '
 	[ "$(grep -E "\.[0-9]{3}" runtime-dotRI.out | wc -l)" = "15" ]
 '
 
+test_expect_success 'flux-jobs --format={contextual_time} works' '
+	flux jobs --filter=pending -c1 -no "{contextual_time:h}" \
+		>ctx_timeP.out &&
+	flux jobs --filter=running -c1 -no "{contextual_time:h}" \
+		>ctx_timeR.out &&
+	flux jobs --filter=completed -c1 -no "{contextual_time:0.3f}" \
+		>ctx_timeCD.out &&
+	echo "300.0" >duration.expected &&
+	test_cmp duration.expected ctx_timeP.out &&
+	test_must_fail test_cmp duration.expected ctx_timeR.out &&
+	test_must_fail test_cmp duration.expected ctx_timeCD.out &&
+	grep -E "\.[0-9]{3}" ctx_timeCD.out
+'
+
 test_expect_success 'flux-jobs emits useful error on invalid format' '
 	test_expect_code 1 flux jobs --format="{runtime" >invalid.out 2>&1 &&
 	test_debug "cat invalid.out" &&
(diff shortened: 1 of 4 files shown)


garlick (approved):
LGTM!

t/t2800-jobs-cmd.t
```diff
@@ -571,6 +571,20 @@ test_expect_success 'flux-jobs --format={runtime:0.3f} works' '
...
+	flux jobs --filter=running -c1 -no "{contextual_time:h}" \
+		>ctx_timeR.out &&
+	flux jobs --filter=completed -c1 -no "{contextual_time:h}" \
+		>ctx_timeCD.out &&
+	echo "300.0" >duration.expected &&
+	test_cmp duration.expected ctx_timeP.out &&
+	test_must_fail test_cmp duration.expected ctx_timeR.out &&
+	test_must_fail test_cmp duration.expected ctx_timeCD.out &&
+	grep -E "\.[0-9]{3}" ctx_timeCD.out
+'
```
chu11, line 586: perhaps also add a test for some failed job too.  Off the top of my head, a canceled pending job should have runtime = 0.  The timed out job should have a runtime > 0 as well.  I can't remember if a `nosuchcommand` job failure hits the running state or not, but that's another one to try.

I notice that such tests were not added to the `{runtime}` tests above this, so dunno if worth copying over there.  B/c of `contextual_time`'s logic, want to make sure inactive jobs are picking up the runtime instead of the duration.

t/t2800-jobs-cmd.t
```diff
@@ -571,6 +571,20 @@ test_expect_success 'flux-jobs --format={runtime:0.3f} works' '
...
+		>ctx_timeP.out &&
+	flux jobs --filter=running -c1 -no "{contextual_time:h}" \
+		>ctx_timeR.out &&
+	flux jobs --filter=completed -c1 -no "{contextual_time:h}" \
+		>ctx_timeCD.out &&
+	echo "300.0" >duration.expected &&
+	test_cmp duration.expected ctx_timeP.out &&
+	test_must_fail test_cmp duration.expected ctx_timeR.out &&
+	test_must_fail test_cmp duration.expected ctx_timeCD.out &&
+	grep -E "\.[0-9]{3}" ctx_timeCD.out
```
chu11, line 585: this grep is b/c the `{runtime}` tests above formatted with `0.3f`.  I suppose could have a spurious failure if some tests runtime was 1 second on the dot?  So perhaps update to "0.3f" too for consistency.  Or maybe just check runtime < 300.0?

chu11 (commented):
overall LGTM, just some test & coverage considerations

t/t2800-jobs-cmd.t
grondo: Ooh, good catch.

t/t2800-jobs-cmd.t
grondo: Yeah, the test already demonstrates that `{runtime}` is used for an inactive job and I had figured the `{runtime}` rendering logic was already tested elsewhere, so seemed like it would be a redundant test.

t/t2800-jobs-cmd.t
grondo: Ok, fixed that one!

chu11 (approved):
LGTM

grondo:
Thanks @chu11 for catching that testing issue, fixed and repushed, and I'll set MWP

Issue #4638, opened 2022-10-03
job-list: support ncores attribute?


grondo:
The job-list module currently only supports the `ntasks` and `nnodes` attributes for determining the "size" of a job request. But this can make a job that requests 4 tasks with 1 core per task and a job with 4 tasks that requests 4 cores per task look the same in `flux jobs` output.

It might be useful to capture an `ncores` attribute when possible along with `nnodes`. We could then attempt to introduce a `size` field in `flux jobs`  that summarizes a pending job size with the known data from jobspec, perhaps replacing the separate `NTASKS`, `NNODES` fields we have now (since `NTASKS` alone doesn't tell us that much)

chu11:
what are thoughts on if `ncores` should be available before it has hit the RUN state?  Similar to `nnodes`, the information can be available early, although the parsing logic is a bit trickier than the `nnodes` case (and maybe `ntasks` case).  

Wondering if I should do #4592 first.

grondo:
If it is possible it should be made available. The idea of the combined `size` field would be to use `nnodes` if possible, but fall back to `ncores` if not (e.g. in the case of `flux mini submit -n16 ...`) It could change once in the run state, e.g. `flux mini submit -n1 --exclusive ...` is a request for 1 core but since the node is allocated exclusively, multiple cores would end up in R.

chu11:
oh crud, I forgot to add a "Fixes" line in #4664, this is completed by that PR

PR #4642, opened 2022-10-03
flux-jobs: add --format=long

### LINKED ISSUE #2627: flux-jobs: Support verbose default option

grondo:
This simple PR adds a `flux jobs --format=long` as described in #2627.

Built on top of #4641 since it uses the `contextual_time` field.

Fixes #2627.

### DIFF
--- a/src/cmd/flux-jobs.py
+++ b/src/cmd/flux-jobs.py
@@ -37,6 +37,15 @@ class FluxJobsConfig(UtilConfig):
                 "{contextual_time!F:>8h} {contextual_info}"
             ),
         },
+        "long": {
+            "description": "Extended flux-jobs format string",
+            "format": (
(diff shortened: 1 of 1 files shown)


chu11:
looks solid to me, although I'm wondering if we'd like to add ncores after that's done? #4638

grondo:
> although I'm wondering if we'd like to add ncores after that's done? 

Once we have ncores I was thinking of replacing separate NTASKS and NNODES with a job SIZE which is combination of ncores and nnodes. In any event, this could be updated after that feature is available, but I thought it would be good to get `--format=long` into this release.

grondo:
Forced a push since #4641 was merged.

PR #4643, opened 2022-10-04
testsuite: document FLUX_TEST_VALGRIND

chu11:
I can never remember what to set, so document it in the testsuite README.  Note that in the sharness script it appears setting it to non-empty is fine, but I just documented to set to "t" since that's we seem to do the most in the issues.


grondo (approved):
Seems like a good addition to me!

PR #4644, opened 2022-10-04
job-list: refactor to abstract "idsync" logic better

chu11:
refactor round two, abstract the "idsync" logic better, instead of the logic being spread around and throughout all the code.


grondo (approved):
Seems reasonable to me!

Issue #4646, opened 2022-10-04
job-list: libjj does not parse `total` tasks correctly


chu11:
PR https://github.com/flux-framework/flux-core/pull/4611 updated `job-list` to use `libjj` to parse ntasks, but `libjj` does not appear to parse the "total" field for tasks correctly:

```
>flux mini submit -N4 -n10 --cores-per-task=2 hostname
f2CrbhzxkX

>flux job info f2CrbhzxkX jobspec | jq
{
  "resources": [
    {
      "type": "node",
      "count": 4,
      "with": [
        {
          "type": "slot",
          "count": 3,
          "with": [
            {
              "type": "core",
              "count": 2
            }
          ],
          "label": "task"
        }
      ]
    }
  ],
  "tasks": [
    {
      "command": [
        "hostname"
      ],
      "slot": "task",
      "count": {
        "total": 10
      }
    }
  ],
  "attributes": {
    "system": {
      "duration": 0,
      "cwd": "/g/g0/achu/chaos/git/flux-framework/flux-core/src/common/libjob"
    }
  },
  "version": 1
}

>flux jobs f2CrbhzxkX
       JOBID USER     NAME       ST NTASKS NNODES     TIME INFO
  f2CrbhzxkX achu     hostname   CD     12      4   0.065s catalyst[160,160,160,160]
```

ntasks should be ~~10~~ ~~20 (forgot about `--cores-per-task`)~~ 10 (oops, confused --cores-per-task with --tasks-per-core), not 12.

chu11:
hmmm, actually maybe `libjj` should not be handling this.  perhaps this is just a special case that `job-list` needs to handle.

grondo:
:grimacing: Yeah, the job shell doesn't even currently support `total` I don't think (this could be part of the reason). (none of the existing front end utilities support it either AFAIK) So there's a larger problem here.

grondo:
Oh, oops. I was mistaken, the shell does support `total` obviously or the simple case of `flux mini run -N4 -n10 ..` wouldn't work. I think the problem was that `libjj` was originally for the scheduler, where `tasks` is irrelevant. A fix would be good. Note there is already a set of test for libjj in `t0022-jj-reader.t`

PR #4651, opened 2022-10-04
job-list: parse tasks with total count in jobspec

chu11:
Problem: Sometimes tasks are listed as a "total" count in the
jobspec.  This case is not parsed by libjj and subsequently
an invalid number of tasks can be returned by job-list.

Add additional parsing to parse this specific case.

Fixes https://github.com/flux-framework/flux-core/issues/4646

Also noticed invalid testing errors in t2260, fixed those up here as well.


grondo (approved):
LGTM! Will be good to get this one in the release too, thanks!

chu11:
restarting a builder that hit #4596

PR #4654, opened 2022-10-05
Flux reactor backend timeout

trws:
This interface mirrors the one used by libuv, which offers:

* uv_backend_fd
* uv_backend_timeout

For flux this adds flux_reactor_resume_timeout, which when combined with
flux_pollfd and some extra timer logic can be used to drive a handle
to completion with relatively efficient blocking behavior.  The design
is that the external event loop should set up watchers much like ev_flux
does to handle IO events, but additionally in the check callback
should invoke flux_reactor_resume_timeout to determine what blocking
mode flux expects.  The logic is:
* If the reactor has a prepare, check, or idle watcher *currently
  active* then it will return 0, and expect the outer loop to call back
  into it on the next iteration of the outer loop.  The outer loop
  should use an idle watcher or similar to ensure this makes progress.
* If the reactor has either timer or periodic watchers registered, then
  it will return the minimum registered relative timeout or pending
  fire time.  The outer loop is expected to call back into the flux
  reactor at that timeout, and possibly before.  Using a timer or
  periodic call into flux should both work.
* If the reactor finds neither pending timeouts nor per-iteration
  watchers, then it will return -1, indicating that the outer loop
  can block on the pollfd of the handle being embedded until an IO
  event arrives.

For the case of embedding a handle's pollfd and associated continuations
and reactor events into an external loop this appears to work well.  It
does not handle *all* reactor events well necessarily.  Using a pollfd, all IO watchers,
signal and child watchers are suspect, and should not be used with this
interface.  The per-iteration and time event watchers should all be
safe.
(NOTE: updated with corrections for pollfd rather than backend_fd)

Here's a small(ish) application that embeds a handle and reactor in a libuv event loop.  It seems to work well with a kvs lookup with a continuation on the future as well as with a timer.  More experimentation would be good, but I think this will let us manage both embedding in other event loops and embedding flux handles in other flux handles for asyncio.  

```c
/************************************************************\
 * Copyright 2014 Lawrence Livermore National Security, LLC
 * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
 *
 * This file is part of the Flux resource manager framework.
 * For details, see https://github.com/flux-framework.
 *

#include <flux/core.h>
#include <stdbool.h>
#include <stddef.h>
#include <stdio.h>
#include <uv.h>

#include "src/common/libev/ev.h"

struct uv_flux;

typedef void (*uv_flux_f) (uv_loop_t *loop, struct uv_flux *w, int revents);
double get_flux_timeout (flux_reactor_t *r);

struct uv_flux {
    uv_poll_t io_w;
    uv_prepare_t prepare_w;
    uv_idle_t idle_w;
    uv_check_t check_w;
    uv_timer_t timer_w;
    flux_t *h;
    int pollfd;
    int events;
    uv_flux_f cb;
    void *data;
    uv_loop_t loop;
    flux_future_t *fut;
};

/* Get flux poll events, converted to libev
 */
static int get_pollevents (flux_t *h)
{
    int e = flux_pollevents (h);
    int events = 0;
    if (e < 0 || (e & FLUX_POLLERR))
        events |= UV_DISCONNECT;
    if ((e & FLUX_POLLIN))
        events |= UV_READABLE;
    if ((e & FLUX_POLLOUT))
        events |= UV_WRITABLE;
    return events;
}

void null_cb (uv_handle_t *h)
{
}

static void prepare_cb (uv_prepare_t *w)
{
    struct uv_flux *fw =
        (struct uv_flux *)((char *)w - offsetof (struct uv_flux, prepare_w));
    int events = get_pollevents (fw->h);
    printf ("prepare_cb events: %d\n", events);
    get_flux_timeout (flux_get_reactor (fw->h));

    if ((events & fw->events) || (events & EV_ERROR))
        uv_idle_start (&fw->idle_w, null_cb);
    else
        uv_poll_start (&fw->io_w, UV_READABLE, null_cb);
}
static void timer_cb (uv_timer_t *w)
{
    struct uv_flux *fw =
        (struct uv_flux *)((char *)w - offsetof (struct uv_flux, timer_w));
    printf ("timer_cb\n");
    get_flux_timeout (flux_get_reactor (fw->h));
    fw->cb (&fw->loop, fw, 0);
    get_flux_timeout (flux_get_reactor (fw->h));
}

static void check_cb (uv_check_t *w)
{
    struct uv_flux *fw =
        (struct uv_flux *)((char *)w - offsetof (struct uv_flux, check_w));

    // if (ev_is_pending (&fw->io_w)
    //         && ev_clear_pending (loop, &fw->io_w) & EV_ERROR) {
    //     fw->cb (loop, fw, EV_ERROR);
    //     return;
    // }
    int events = get_pollevents (fw->h);
    printf ("check_cb events: %d\n", events);
    fw->cb (&fw->loop, fw, events);

    double to = get_flux_timeout (flux_get_reactor (fw->h));
    uv_poll_stop (&fw->io_w);
    uv_idle_stop (&fw->idle_w);

    if (to < 1e-100 && to > -1e-100) {
        uv_idle_start (&fw->idle_w, null_cb);
    } else if (to > 0)
        uv_timer_start (&fw->timer_w, timer_cb, to * 1000, 0);
}

int uv_flux_init (struct uv_flux *w, uv_flux_f cb, flux_t *h, int events)
{
    w->cb = cb;
    w->h = h;
    w->events = events;
    if ((w->pollfd = flux_pollfd (h)) < 0)
        return -1;

    w->prepare_w.data = w;
    w->check_w.data = w;
    uv_prepare_init (&w->loop, &w->prepare_w);
    uv_timer_init (&w->loop, &w->timer_w);
    uv_check_init (&w->loop, &w->check_w);
    uv_idle_init (&w->loop, &w->idle_w);
    uv_poll_init (&w->loop, &w->io_w, w->pollfd);

    return 0;
}

void uv_flux_start (uv_loop_t *loop, struct uv_flux *w)
{
    uv_prepare_start (&w->prepare_w, prepare_cb);
    uv_check_start (&w->check_w, check_cb);
}

void uv_flux_stop (uv_loop_t *loop, struct uv_flux *w)
{
    uv_prepare_stop (&w->prepare_w);
    uv_check_stop (&w->check_w);
    uv_poll_stop (&w->io_w);
    uv_idle_stop (&w->idle_w);
    uv_timer_stop (&w->timer_w);
}

void run_loop_cb (uv_loop_t *loop, struct uv_flux *w, int revents)
{
    flux_reactor_t *r = flux_get_reactor (w->h);
    flux_reactor_run (r, FLUX_REACTOR_ONCE);
}
double get_flux_timeout (flux_reactor_t *r)
{
    double ret = flux_reactor_resume_timeout (r);
    printf ("timeout: %lf\n", ret);
    return ret;
}

void tw_cb (flux_reactor_t *r, flux_watcher_t *w, int revents, void *data)
{
    struct uv_flux *fw = data;
    printf ("We made it into timer!\n");
    get_flux_timeout (flux_get_reactor (fw->h));
    uv_flux_stop (&fw->loop, fw);
    uv_stop (&fw->loop);
}

void then_cb (flux_future_t *fut, void *data)
{
    struct uv_flux *ef = data;
    printf ("We made it!\n");
    flux_future_destroy(fut);
    // uv_flux_stop(&ef->loop, ef);
    // uv_stop(&ef->loop);
}

int main (int argc, char *argv[])
{
    struct uv_flux ef = {0};
    flux_t *f = flux_open (NULL, 0);
    flux_reactor_t *r = flux_get_reactor (f);
    uv_loop_init (&ef.loop);
    uv_flux_init (&ef, run_loop_cb, f, UV_READABLE);
    uv_flux_start (&ef.loop, &ef);
    get_flux_timeout (r);
    flux_future_t *fut = flux_kvs_lookup (f, NULL, 0, "resource.R");
    ef.fut = fut;
    flux_future_then (fut, -1, then_cb, &ef);
    get_flux_timeout (r);
    flux_watcher_t *timer_watcher = flux_timer_watcher_create (r, 5, 0, tw_cb, &ef);
    flux_watcher_start (timer_watcher);
    get_flux_timeout (r);
    return uv_run (&ef.loop, UV_RUN_DEFAULT);
}

/*
 * vi:tabstop=4 shiftwidth=4 expandtab
 */
```

@vchuravy, you might want to check this out.

### DIFF
--- a/src/common/libflux/reactor.c
+++ b/src/common/libflux/reactor.c
@@ -16,7 +16,10 @@
 #include <errno.h>
 #include <stdbool.h>
 #include <fcntl.h>
+#include <float.h>
+#include <math.h>
 
+#include "config.h"
 #include "handle.h"
 #include "reactor.h"
 #include "ev_flux.h"
@@ -111,6 +114,42 @@ flux_reactor_t *flux_get_reactor (flux_t *h)
         }
     }
     return r;
+}
+
+static __thread double timeout_seconds = 0;
+
+static void check_timeouts(struct ev_loop *current_loop, int t, void *vw) {
+  ev_watcher *w = vw;
+  ev_tstamp remaining = 0;
+  if (t == EV_TIMER)
+    remaining = ev_timer_remaining(current_loop, (ev_timer *)w);
+  else if (t == EV_PERIODIC)
+    remaining = ev_periodic_at(w) - ev_now(current_loop);
+
+  // take the minimum
+  if (isnan(timeout_seconds) || timeout_seconds > remaining)
+    timeout_seconds = remaining;
+}
+
+static __thread bool any_active = false;
+
+static void check_any_active(struct ev_loop *current_loop, int t, void *vw) {
+  ev_watcher *w = vw;
+  if (ev_is_active(w))
+    any_active = true;
+}
+
+double flux_reactor_resume_timeout(flux_reactor_t *r) {
+  any_active = false;
+  ev_walk(r->loop, EV_CHECK | EV_IDLE | EV_PREPARE, check_any_active);
+  if (any_active)
+    return 0;
+
+  timeout_seconds = NAN;
+  ev_walk(r->loop, EV_TIMER | EV_PERIODIC, check_timeouts);
+  if (isnan(timeout_seconds))
+    return -1;
+  return timeout_seconds;
 }
 
 int flux_reactor_run (flux_reactor_t *r, int flags)
--- a/configure.ac
+++ b/configure.ac
@@ -26,6 +26,8 @@ AM_MAINTAINER_MODE([enable])
 
 AC_DEFINE([_GNU_SOURCE], 1,
           [Define _GNU_SOURCE so that we get all necessary prototypes])
+AC_DEFINE([EV_WALK_ENABLE], 1,
+          [Define EV_WALK_ENABLE to enable ev_walk])
 
 ##
 # Generate project versions from PACKAGE_VERSION (set from git describe above)
--- a/src/common/libflux/reactor.h
+++ b/src/common/libflux/reactor.h
@@ -53,6 +53,12 @@ void flux_reactor_destroy (flux_reactor_t *r);
 flux_reactor_t *flux_get_reactor (flux_t *h);
 int flux_set_reactor (flux_t *h, flux_reactor_t *r);
 
+/* seconds until the loop should be resumed to make progress efficiently
+   <0 for blocking ok
+   0 for return as soon as possible
+   >0 for time in floating point seconds */
+double flux_reactor_resume_timeout(flux_reactor_t *r);
+
 int flux_reactor_run (flux_reactor_t *r, int flags);
 
 void flux_reactor_stop (flux_reactor_t *r);


src/common/libflux/reactor.c
```diff
@@ -113,6 +116,42 @@ flux_reactor_t *flux_get_reactor (flux_t *h)
...
+
+static void check_any_active(struct ev_loop *current_loop, int t, void *vw) {
+  ev_watcher *w = vw;
+  if (ev_is_active(w))
+    any_active = true;
+}
+
+double flux_reactor_resume_timeout(flux_reactor_t *r) {
+  any_active = false;
+  ev_walk(r->loop, EV_CHECK | EV_IDLE | EV_PREPARE, check_any_active);
```
garlick, line 144: Really excellent job finding the undocumented `ev_walk()` interface @trws!  Apparently you looked at libev code and did not go blind!

I think I may have misspoken in our chat - I believe only the `ev_idle` watcher prevents the loop from blocking when active.  The prepare and check watchers can remain active and the loop will still block.  So I think you'd want to drop EV_CHECK and EV_PREPAR from the `ev_walk()` call highlighted above.  (I believe this makes it more correct although in the particular case of the check watcher in the future continuation implementation, it looks like we don't leave it running when we expect the loop to block, so I don't expect it will make a difference in testing)

If I understand correctly, the strategy is to never block in `ev_run()`, so the only event sources that will cause the flux loop to run are the timeout and the `flux_pollfd()`.  So that really means _none_ of the other watchers besides the flux_t handle and the idle watcher will work as asynchronous event sources in the flux reactor.   For example a file descriptor watcher in the flux reactor cannot make the parent event loop wake up.  Have I got that right?  It seems like your description is saying otherwise.

Due to the limitations, regardless of what they are, I was suggesting in the chat that we maybe build a flag in to the flux reactor to indicate to `flux_reactor_create()` that it's running the special "adjunct reactor" mode.  If the flag is set, fail any attempt to create watchers that are unsupported.  Conversely, fail `flux_reactor_resume_timeout()` if the flag is clear.  Your proposed interface appears to be complete for embedding a `flux_t` in a foreign loop.  However if someone decided to use some lesser used API functions, libsubprocess, for example that makes use of child watchers and buffer watchers (which internally use fd watchers), it would just fail and there would be no intermittent hangs to track down.

When we get a little further along, it would probably make sense to add a unit test to libflux/test that embeds the flux reactor in...another flux reactor (or in libev)?

This is real progress, thanks so much for pushing it forward!!

trws: Thanks @garlick, and you're absolutely right about the support, for some reason I was thinking about using the libev backend_fd as the watched file descriptor rather than a pollfd and messed up what will and will not work.  Using this interface along with a pollfd will get you the messages and associated continuations, timers, prepare, check and idle watchers on the underlying reactor.  None of the others will work without some additional work.  If we're willing to extend libev a bit to expose the backend_fd, everything else would work when using epoll on linux, with some unfortunate properties for signalfd.  I'm not sure if I want to go that far, at least for this step, so we'll be limited to a handle, messages, continuations and timers.

With that being the limitation, I'm considering how we should build the interface to mark a reactor in "embedded" mode.  It could be directly a property of trying to use `flux_reactor_resume_timeout`, checking for any unsupported watcher and bailing out or we could make it happen when a pollfd is requested from the handle but I'm not sure if we could actually do it on create of the reactor.  Mainly I'm thinking of the situation where a handle gets created in bindings, and we don't know if the user intends to use it directly with `reactor_run` in blocking mode or if they intend to embed it.  What do you think?

garlick: Performing the `ev_walk()` sweep each time  `flux_reactor_resume_timeout()` is called and skipping the flag entirely seems appealing since it ought to catch any rogue watchers added to the event loop, and makes this solution self contained.  Although on a critical path of sorts, that scan ought to be quick since in this use case there's likely to only be a handful of watchers.

One thing to note (not necessarily a problem but might come up in test) is that the sweep won't catch watchers that are created but not started.  Libev doesn't track watchers that aren't started.

Upon looking at reactor code i was regretting suggesting we test a flag in the watcher create path.  it seems like we'd either have to test  it in every watcher implementation or refactor the generic watcher interface.  And we've exposed that interface in the public API.

trws: Yeah, I think making it self contained is worth the possible cost at least for now.  If we later found it was prohibitive for some reason we could re-work without breaking anything, either by tracking a different way or adding something to libev to have a more efficient walk that doesn't require thread-local memory, but I don't expect it to be performance critical or terribly expensive.

I was slightly surprised by the behavior of non-started watchers being missing at first, but in the end I think it actually helps us to some extent.  This way if something creates a watcher because it might use it later, but doesn't actually use it, we aren't pessimizing and breaking on that case. It also means we should be iterating over less watchers.

garlick: Since there's already an `ev_walk()` in there to look for idle watchers, maybe the callback could just be extended to also look for the rogue watchers.

Using the callback`s `void *` argument to point to a struct on the stack of `flux_reactor_resume_timeout()` could let you pass in the watcher and flags and avoid thread-local storage for flags, unless I'm missing something.

trws: ev_walk doesn't offer a void * argument in its callback that we can control unfortunately, I originally thought it did but it passes in the watcher there and expects the callback to determine the type from the int.  If we had everything we need in the watcher's `data` pointer that would work, but it puts us in a chicken and egg situation.   Actually... I could save the data pointer from the loop, put a stack variable in there for state that will get passed in, then restore it.  That might actually work, will give it a shot.

garlick: Ah I misread that, sorry.

`data` is member of `flux_watcher_t` not `ev_watcher`, unless I'm not spotting another `data` member, as it's not all defined in one place it seems.

Probably no big deal if we have to stick with the thread local storage, at least for nw.

garlick:
This was a valiant attempt but I think since a couple of years have gone by without action due to other priorities, it's appropriate to withdraw the PR for now.

trws:
Yeah, agreed, thanks for the triage @garlick.  I think something like this would be useful, but the "right" approach would be to hack in the "get a pollable fd from libev" as suggested to someone on the libev mailing list rather than doing this.  It would give us the ability to truly catch all the events rather than what we have now, but needed a redo. Mostly adding this here as context in case we come back to it.

PR #4655, opened 2022-10-05
test: job-ingest: skip guest tests when default sign-type fails

grondo:
Problem: The job-ingest test_job.t unit test fails if the build environment is not able to sign jobspec -- e.g. if sign-type=munge is configured but the munge daemon is not running. But the purpose of this unit test is not to test flux-security but to test job-ingest interfaces, so forcing failure due to misconfigured environment is not necessary.

Just skip the "guest" tests when the default sign-type is not functional.

(This PR fixes a failure in toss buildfarm, where the munge daemon is not running)


garlick (approved):
Sorry about that!  I completely missed that this would be an issue.
Solution looks good.

grondo:
It is not a straightforward one to remember. I wonder if we should disable munge in one of the CI builds to catch this kind of thing.

Issue #4657, opened 2022-10-06
broker: runat broker tests hang sometimes on TOSS3


chu11:
After #4653 was merged, the `runat` tests in the broker are hanging for me sometimes:

```
# info: test6.0: local://notreally
# info: test6.0: printenv FLUX_URI Exited (rc=0) 0.0s
ok 36 - completion called once
ok 37 - exit code zero
ok 38 - FLUX_URI was set for subprocess
ok 39 - pushed /bin/true;sleep 3600
ok 40 - runat_start works
ok 41 - runat_abort works
```

for me they're hanging at perhaps 50-70% of the time, more often than not.  It appears its hanging in a `flux_reactor_run()` call:  

```
    ok (runat_abort (r, "test7") == 0,
        "runat_abort works");
    completion_called = 0;
    ok (flux_reactor_run (flux_get_reactor (h), 0) >= 0
        && completion_called == 1,
        "completion called once");
    ok (runat_get_exit_code (r, "test7", &rc) == 0 && rc == 129,
        "exit code 129 (= signal 15 + 1)");
    ok (match_list (logs, "Hangup") == 1,
        "process termination was logged");
    rc = -1;

```

Which would suggest some watcher has been leftover?  Hopefully not in `libsubprocess`.

This is on catalyst a TOSS3 system.  But does not appear to trigger on corona, a TOSS4 system.  I have no idea why the OS would matter.  Or perhaps its inherently racy and catalyst just triggers it somehow.

Edit: an aside, typo fix to be done here in runat tests (inidicating -> indicating)

```
    /* N.B. if sleep has started, the abort function kills it.
     * If it is not yet started, the subprocess state callback kills it
     * when it transitions to running.  Either way we should see an
     * exit code inidicating terminated.
     */

```

chu11:
I had forgotten about `posix_spawn` support in `libsubprocess` and took a wild guess that there are differences in it with TOSS3 vs TOSS4.  Commented it out in `libsubprocess` as a test and hang disappears.  So perhaps a SETPGRP subtlety in `posix_spawn` implementation or perhaps a RHEL7 vs RHEL8 subtlety, neither of which would surprise me.

chu11:
given the nature of this test, which is basically to run something and then immediately abort it, it wouldn't shock me that there is something racy in `posix_spawn` on RHEL7, we are looking at an 8 year older glibc/kernel/whereever the issue is.

as nothing more than a guess

```
[...]

PR #4660, opened 2022-10-06
broker/test: Fix runat race on older glibc versions

chu11:
Problem: On older glibc versions, the setting of the child process group via POSIX_SPAWN_SETPGROUP can be racy.  A call to killpg() immediately after a spawned process is launched may not kill a child process.  In one runat test, this can give the appearance of a hang as a 3600 second sleep call is being waited on.

- support a new flag in libsubprocess to force the use of FORK/EXEC
- refactor runat code to take flags instead of bool for log_stdio
- have runat flags take a flag for forced use for FORK/EXEC
- have test use FORK/EXEC if using older versions of glibc

Fixes https://github.com/flux-framework/flux-core/issues/4657

### DIFF
--- a/src/common/libsubprocess/local.c
+++ b/src/common/libsubprocess/local.c
@@ -396,7 +396,9 @@ static void child_watch_cb (flux_reactor_t *r, flux_watcher_t *w,
 
 static int create_process (flux_subprocess_t *p)
 {
-    if (!p->hooks.pre_exec && !flux_cmd_getcwd (p->cmd))
+    if (!(p->flags & FLUX_SUBPROCESS_FLAGS_FORK_EXEC)
+        && !p->hooks.pre_exec
+        && !flux_cmd_getcwd (p->cmd))
         return create_process_spawn (p);
     return create_process_fork (p);
 }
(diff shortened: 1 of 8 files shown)


src/common/libsubprocess/local.c
```diff
@@ -396,7 +396,9 @@ static void child_watch_cb (flux_reactor_t *r, flux_watcher_t *w,
 
 static int create_process (flux_subprocess_t *p)
 {
-    if (!p->hooks.pre_exec && !flux_cmd_getcwd (p->cmd))
+    if (!(p->flags & FLUX_SUBPROCESS_FLAGS_FORK_EXEC)
+        && !p->hooks.pre_exec
+        && !flux_cmd_getcwd (p->cmd))
```
grondo, line 401: This seems fine, but is it perhaps a bit overkill to add this flag for just one test? Or were you considering other use cases? Based on the code changed here, I'm sure you saw that you can also force fork()/exec() by setting a cwd of `.`.

grondo (approved):
This fix seems ok to me and very well analyzed! It is a little disturbing that all this is needed to fix one test on one OS 😞 . I'm a little worried there will be no other users ever of these new flags ...

src/common/libsubprocess/local.c
chu11: yeah, it is a bit overkill :-)  But I figured there's a decent chance we'll need to force fork/exec again in the future for one reason or another.  Maybe I'll be wrong :P

grondo: Well let's hope not (at least not from the broker)

PR #4661, opened 2022-10-06
etc: add functional bash completions

grondo:
This PR reimplements the existing bash completion script to be much more fully featured. Not _all_ commands have completions after the first level `flux <TAB>` completion, but many of the most common commands have been added (see the file for details). Additionally, some experimental support for completing jobids and queues was added. Also notably, this fixes the problem we had before where the completion was disabling bash filename completion under any command that started with `flux`.

It is difficult to test bash completions except by hand, so these should be considered experimental and there could be lingering issues.

### DIFF
--- a/etc/completions/flux.pre
+++ b/etc/completions/flux.pre
@@ -9,56 +9,1247 @@
 #
 shopt -s extglob
 
+# return success if first argument is in remaining args
+_flux_contains_word() {
+    local w word=$1; shift
+    for w in "$@"; do
+        [[ $w = "$word" ]] && return
+    done
+}
+
+# Determines the first non-option word of the command line. This
+# is usually the command. Returns empty string if first non-option
+# word is $cur, since that means we're still completing the command.
+#
+_flux_get_cmd() {
+    local firstword i
+
+    firstword=
+    for ((i = 1; i < ${#COMP_WORDS[@]}; ++i)); do
+        if [[ ${COMP_WORDS[i]} != -* && ${COMP_WORDS[i]} != $cur ]]; then
+            firstword=${COMP_WORDS[i]}
+            break
+        fi
+    done
+
+    echo $firstword
+}
+
+# Determines the first non-option word after another given word,
+# This is usually a sub-command
+# Returns the word if result = $cur since that means we're still
+# completing the current subcommand
+#
+_flux_get_subcmd() {
+    local firstword i word=$1; shift
+
+    firstword=$word
+    found=0
+    for ((i = 1; i < ${#COMP_WORDS[@]} - 1; ++i)); do
+        local w="${COMP_WORDS[i]}"
+        if [[ $w != -*  && -n $w && $w != $cur ]]; then
+            if ((found == 1)); then
+                firstword=$w
+                break
+            fi
+            if [[ $w == $word ]]; then
+                found=1
+            fi
+        fi
+    done
+
+    echo $firstword
+}
+
+# Autocomplete queue name for `-q, --queue=NAME`
+# returns failure if cur or prev option is not -q,--queue
+#
+_flux_complete_queue() {
+    #  Autocomplete queue names
+    if [[ "$prev" == "--queue" \
+        || "$prev" == "-q" \
+        || "$cur" == "--que"* ]]; then
+        local queues=$(flux queue status | sed -n 's/^\(.*\): .*$/\1/p')
+        if [[ "$cur" == "--que"* ]]; then
+            queues=$(printf -- '--queue=%s ' $queues)
+        fi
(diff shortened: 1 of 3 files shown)


etc/completions/flux.pre
```diff
@@ -9,56 +9,963 @@
...
+            # Add space if there is not a '=' in suggestions
+            compopt -o nospace
+        fi
+    else
+        COMPREPLY=( $(compgen -W "${subcmds}" -- "$cur") )
+    fi
+    return 0
+}
+
+# flux-admin completions
```
chu11, line 690: flux-admin(8) for consistency of other comments

etc/completions/flux.pre
```diff
@@ -9,56 +9,963 @@
...
+            return 0
+        fi
+    else
+        COMPREPLY=( $(compgen -W "${subcmds}" -- "$cur") )
+    fi
+    return 0
+}
+
+
+# flux-kvs eventlog completions
```
chu11, line 460: flux-kvs(1) for consistency to other comments?

etc/completions/flux.pre
chu11: oh wait, seems to be installed in /usr/bin?  is that what we want?

chu11 (commented):
just a few high level comments, obviously a lot of bash-completion specific stuff that i'm not too knowledgeable about.  But played with it a bit and it seems to work.  i think the tab completion for job ids is especially useful for testing.

etc/completions/flux.pre
```diff
@@ -9,56 +9,963 @@
...
+			    firstword=$w
+			    break
+            fi
+            if [[ $w == $word ]]; then
+                found=1
+            fi
+		fi
+	done
+
+	echo $firstword
```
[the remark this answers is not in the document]
grondo: Thanks I'll clean that up (Thus the addition of `vi` modeline :facepalm:)

etc/completions/flux.pre
grondo: > oh wait, seems to be installed in /usr/bin? is that what we want?

I'm not 100% sure what this is referring to, but it doesn't sound right :slightly_smiling_face: Want to explain more?

etc/completions/get_builtins.sh
```diff
@@ -34,3 +37,5 @@ get_builtins() {
 }
 
 get_builtins $1
+
+# vi: ts=4 sw=4 expandtab
```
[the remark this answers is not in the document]
grondo: I might just drop the commit I found it isn't working anyway and `flux restore` also conflicts with completion of `flux resource` more than `flux relay` so the change isn't really helpful.

etc/completions/flux.pre
grondo: Yeah, there's no man page for `flux-admin` but I put `flux-admin(8)` anyway. Note that `flux-admin.py` should not be in `/usr/bin` but `/usr/libexec/flux/cmd/flux-admin.py`.

grondo:
Ok, I force pushed a new version here that fixes an issue where e.g. `flux mini<TAB>` didn't autocomplete by adding a trailing space, added a bunch of new commands (with attribute completion for `flux getattr <TAB>`). Also forgot to mention that we have loaded module completion with `flux module [remove|reload|info|stats] <TAB>`.

Based on @garlick's feedback I also updated the installation dir for the completions file from hard-coded `${sysconfdir}/bash_completion.d` to use pkg-config based on the bash-completions [FAQ](https://github.com/scop/bash-completion/blob/master/README.md#faq).

The change to the `get_builtins.sh` script was also dropped.

etc/completions/get_builtins.sh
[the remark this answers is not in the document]
grondo: commit dropped

grondo:
I also fixed a problem with the completions file prereqs in `etc/Makefile.am`

grondo:
Oh, oops `el7` doesn't seem to have `PKG_CHECK_VAR` so I'll have to figure out a different way to do that.

grondo:
Restarting a build due to 
>  503 Egress is over the account limit.

I've fixed the pkg-config autoconf stuff and this is ready for a real review.

garlick (approved):
These work great for me. I confirmed that they are now installed into `/usr/share/bash-completion/completions` when I do `make install` to the default prefix of `/usr/local` and that this is picked up by bash on my system, and it seems to work as it should.  Thanks for doing this - great time saver.

grondo:
Thanks! I'll set MWP.

grondo:
Belatedly I thought: how is installing always to the same bashcompdir going to work for spack or `make install` off to the side? I'd better open an issue.

PR #4665, opened 2022-10-07
rc1: reload configuration on rank > 0

### LINKED ISSUE #4663: brokers may use old config after a flux restart orchestrated wtih flux shutdown
Problem:  Running `flux shutdown` on the management node, then pushing out config changes, then running `systemctl start flux` on the management node does not apply the config changes on all nodes.

This outcome seems counter-intuitive, but what happens is the brokers on the other other nodes are immediately restarted by systemd after the shutdown.  They then block waiting for upstream node to come online.  If the configuration changes _after_ they restart, it is not picked up unless `flux config reload` is run.

A solution would be to have the broker automatically reload the config after the upstream rank comes online.

garlick:
Problem: although flux shutdown stops the rank 0 broker in a system instance, the other ranks are immediately restarted by systemd. This means that altering the configuration while flux is "down" may be too late for the brokers that have already restarted but have not yet rejoined the instance.

To avoid confusing people, run flux config reload in rc1 on rank > 0 to ensure it starts with the latest configuration.

Fixes #4663

### DIFF
--- a/etc/rc1
+++ b/etc/rc1
@@ -16,6 +16,12 @@ backing_module() {
     echo ${backingmod:-content-sqlite}
 }
 
+# Get the latest config in case it changed while upstream broker was down.
+# See also: flux-framework/flux-core#4663
+if test $RANK -gt 0; then
+    flux config reload
+fi
+
 modload all barrier
 
 if test $RANK -eq 0; then


grondo (approved):
LGTM!

garlick:
OK, I'll set MWP on this one then, thanks.

PR #4667, opened 2022-10-07
build: fix use of system bash-completion dir

### LINKED ISSUE #4666: bash completion file is always installed to system bash completion path
Problem: commit ee47ced454f384b296b4e5fdb2be5134110326c1 changed the installation path for the flux bash completion file to the result of `pkg-config --variable=completionsdir bash-completion`, ignoring `--prefix`, but this will break side-install packages and possibly spack builds since the system bash-completion path is always used.

We may want to revert this commit, or perhaps we can make use of the system bashcompdir a new configure option, and use the default of `${sysconfdir}/etc/bash_completion.d/` otherwise (this preserves existing behavior, but allows a user installing to `/usr/local` for instance to still direct the completion file to the right system location)

grondo:
Problem: commit ee47ced changed the installation path for the flux bash completion file to the result of

  pkg-config --variable=completionsdir bash-completion

ignoring `--prefix`. However, this will break `make install` to side install locations, temporary installs, and possibly spack builds since the system bash-completion path is always used.

Set `bashcompdir=${sysconfdir}/etc/bash_completion.d` by default and add a new option `--with-system-bash-completion-dir` for users that want to force installation of the completion file to the system-defined bash-completion path.

Fixes #4666

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -341,9 +341,17 @@ PKG_CHECK_MODULES([LIBUUID], [uuid], [], [])
 PKG_CHECK_MODULES([CURSES], [ncursesw], [], [])
 PKG_CHECK_MODULES([LIBARCHIVE], [libarchive], [], [])
 
-PKG_CHECK_EXISTS([bash-completion],
-    [bashcompdir=`$PKG_CONFIG --variable=completionsdir bash-completion`],
-    [bashcompdir="${sysconfdir}/bash_completion.d"])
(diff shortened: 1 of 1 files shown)


garlick:
Nice thank you!  I will give this a quick test on my system.

garlick (approved):
I tested this with and without the option - all good!

grondo:
Great, thanks!

PR #4669, opened 2022-10-10
python: move and rename some classes for reusability

grondo:
This PR moves some code around in `flux.util` and `flux.job.info` to enhance reusability (required for some follow up PRs), including some other cleanup.

In summary:

 * The `JobFormatter` and `HeaderFormat` class are embedded in the `JobInfoFormat`class, but these classes are not job specific. Move `JobFormatter` to `flux.util.UtilFormatter` and embed `HeaderFormat` in the more general `OutputFormat` class.
 *  Move `JobDatetime`, which is also not really job-specific, to `flux.util.UtilDatetime`, and the `fsd()` function from `flux.job.info` into `flux.util`.
 * Move `filter_empty()` to `OutputFormat` and modify its return type to be a new format string instance of an instance of the `OutputFormat` class for reusability.
 * Fix a bug in `OutputFormat` constructor when a format string that already has `0.` prefixes is used.

### DIFF
--- a/src/bindings/python/flux/job/info.py
+++ b/src/bindings/python/flux/job/info.py
@@ -13,7 +13,6 @@ import time
 import pwd
 import json
 import string
-from datetime import datetime, timedelta
 from collections import namedtuple
 
 import flux.constants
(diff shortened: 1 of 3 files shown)


chu11 (approved):
lgtm!

grondo:
Thanks! I'll set MWP.

PR #4674, opened 2022-10-11
Add devcontainer environment for vscode

vsoch:
Problem: it is hard to develop flux-core locally.

Solution: This addition will allow VSCode users to quickly run (and work in) a development container, defined under .devcontainer in the root, which contains a devcontainer.json file and a Dockerfile that uses the fluxrm/testenv:focal base to quickly pull and
spin up the container. We need the Dockerfile to support a post start command to fix a git security issue, and also add developer tools like formatters, linters, etc. This is a vanilla install, meaning it just will allow for build of the core flux, and as other developers try this out we can easily add other extensions that make the development experience better!

The one sticky bit was with git permissions - if you sign you can't commit from inside the container (no keys) and then when you go outside the container, having commit inside the container, you get a permissions issue (that can be fixed with a chown). Likely the best solution is just not to write to git (commit) from inside the container for now, and people can play around and see if there is a better way to go about it. I don't personally mind bringing up another terminal.

My thinking is that we can get a vanilla version in first, and then those that try it with VSCode that have preferences for extensions can propose adding them separately.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,13 @@
+FROM fluxrm/testenv:focal
+
+LABEL maintainer="Vanessasaurus <@vsoch>"
+
+# For easier Pyton development. 
+# TODO add pre-commit installs here when added!
+RUN python3 -m pip install IPython && \
+    python3 -m pip install 'black==20.08.b1' --force-reinstall && \
+    python3 -m pip install 'pylint==2.4.4' --force-reinstall && \
+    python3 -m pip install 'mypy==0.770' --force-reinstall
+
+# Assuming installing to /usr/local
+ENV LD_LIBRARY_PATH=/usr/local/lib
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,16 @@
+{
+    "name": "Flux Core Python 3.6",
+    "dockerFile": "Dockerfile",
+  
+    "customizations": {
+      "vscode": {
+        "settings": {
+          "terminal.integrated.defaultProfile.linux": "bash"
+        }, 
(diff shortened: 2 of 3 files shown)


grondo (approved):
This seems good to me. I'm not a VScode user so can't really comment on implementation.

The new vscode doc in the README is pretty vscode+developer specific, so I thought maybe it should be in a collapsible section, but then again there isn't much in there so it doesn't really matter at this point.

vsoch:
@grondo we probably need a strategy for having explicitly developer documentation - I think for Python (my pre-commit PR) it worked well to put under docs/python, but for this I wasn't sure where it would best go, because most of the content here is maaaaaaan pages! :bat: :man:

grondo:
Yeah, works for me! My only suggestion was maybe to put in a collapsible field, but it certainly isn't critical at this point. Feel free to set MWP unless someone else has objections? (I'm not sure if there are any other developers that work on flux-core and use vscode to test this out)

vsoch:
@trws I thought you mentioned you might?

I would argue to _not_ put it in a collapsible section because:

1. It will remind us we need a strategy for developer docs (ideas?)
2. People that could be looking for it won't easily find it
3. It looks super up-to-date and cool to be using them!
4. Probably most people don't scroll that far down in the README anyway... :laughing: 

And do you mean like this?

<details>

<summary>Hello this is details!</summary>

```console
echo raining meatballs!
```
</details>

I guess I've never thought to use that in a readme - didn't think it would work! Definitely need to try now!

grondo:
I only know about collapsible sections because we already use it in the README. Like I said I'm fine without it.

vsoch:
> I only know about collapsible sections because we already use it in the README. Like I said I'm fine without it.

I am probably good either way - if someone else wants to chime in for a third vote I'll be happy to just do that!

PR #4675, opened 2022-10-12
broker: allow custom topology to be configured

garlick:
As discussed in #3799, this adds an optional "parent" key to `[bootstrap]` hosts array entries so that a custom topology can be expressed.  Here's an example from a tiny test cluster:
```toml
hosts = [
    { host = "picl0" },
    { host = "picl[1-2]", parent = "picl0" },
    { host = "picl[3-5]", parent = "picl1" },
    { host = "picl[6-7]", parent = "picl2" },
]
```
```
 garlick@picl0:~$ flux overlay status
0 picl0: full
├─ 1 picl1: full
│  ├─ 3 picl3: full
│  ├─ 4 picl4: full
│  └─ 5 picl5: full
└─ 2 picl2: full
   ├─ 6 picl6: full
   └─ 7 picl7: full
```
A larger cluster that preserves the mapping of ranks to hostname suffixes while using non consecutive nodes as routers as might be done in a system with scalable units:
```toml
[[bootstrap.hosts]]
host = "test[0-127]"

[[bootstrap.hosts]]
host = "test[1,64]"
parent = "test0"
[[bootstrap.hosts]]
host = "test[2-63]"
parent = "test1"
[[bootstrap.hosts]]
host = "test[65-127]"
parent = "test64"
```

### DIFF
--- a/src/broker/boot_config.c
+++ b/src/broker/boot_config.c
@@ -480,40 +480,6 @@ int boot_config_geturibyrank (json_t *hosts,
     return 0;
 }
 
-/* Get the value of the tbon.fanout attribute.
- * If it was set, set the immutable flag.
- * If it wasn't set, set a default value (if non-NULL).
- */
-static const char *get_tbon_fanout (attr_t *attrs, const char *default_value)
-{
-    const char *key = "tbon.fanout";
-    const char *val = NULL;
-    int flags;
-
-    if (attr_get (attrs, key, &val, &flags) == 0) {
-        if (!(flags & FLUX_ATTRFLAG_IMMUTABLE)
-            && attr_set_flags (attrs, key, FLUX_ATTRFLAG_IMMUTABLE) < 0)
-            return NULL;
-    }
-    else {
-        val = default_value;
-        if (val && attr_add (attrs, key, val, FLUX_ATTRFLAG_IMMUTABLE) < 0)
-            return NULL;
-    }
-    return val;
-}
-
-static bool have_parent_directives (json_t *hosts)
-{
-    size_t index;
-    json_t *entry;
-    json_array_foreach (hosts, index, entry) {
-        if (json_object_get (entry, "parent"))
-            return true;
-    }
-    return false;
-}
-
 int boot_config (flux_t *h, struct overlay *overlay, attr_t *attrs)
 {
     struct boot_conf conf;
@@ -522,6 +488,7 @@ int boot_config (flux_t *h, struct overlay *overlay, attr_t *attrs)
     json_t *hosts = NULL;
     struct topology *topo = NULL;
     flux_error_t error;
+    const char *topo_uri;
 
     /* Ingest the [bootstrap] stanza.
      */
@@ -553,33 +520,17 @@ int boot_config (flux_t *h, struct overlay *overlay, attr_t *attrs)
         rank = 0;
     }
 
-    if (have_parent_directives (hosts)) {
-        if (get_tbon_fanout (attrs, NULL)) {
-            errno = EINVAL;
-            log_msg ("tbon.fanout conflicts with custom topology config");
-            goto error;
-        }
-        topology_hosts_set (hosts);
-        topo = topology_create ("custom:", size, &error);
-        topology_hosts_set (NULL);
-        if (!topo) {
-            log_msg ("%s", error.text);
-            goto error;
-        }
+    // N.B. overlay_create() sets the tbon.topo attribute
+    if (attr_get (attrs, "tbon.topo", &topo_uri, NULL) < 0) {
+        log_msg ("error fetching tbon.topo attribute");
+        goto error;
     }
-    else {
-        const char *fanout;
-        char uri[16];
-
-        if (!(fanout = get_tbon_fanout (attrs, "0"))) {
-            log_err ("Error manipulating tbon.fanout attribute");
-            goto error;
-        }
-        snprintf (uri, sizeof (uri), "kary:%s", fanout);
-        if (!(topo = topology_create (uri, size, &error))) {
-            log_msg ("Error creating %s topology: %s", uri, error.text);
-            goto error;
-        }
+    topology_hosts_set (hosts);
+    topo = topology_create (topo_uri, size, &error);
+    topology_hosts_set (NULL);
+    if (!topo) {
+        log_msg ("Error creating %s topology: %s", topo_uri, error.text);
+        goto error;
     }
     if (topology_set_rank (topo, rank) < 0
         || overlay_set_topology (overlay, topo) < 0)
--- a/src/broker/boot_config.c
+++ b/src/broker/boot_config.c
@@ -141,11 +141,22 @@ static json_t *boot_config_expand_hosts (json_t *hosts)
     }
     json_array_foreach (hosts, index, value) {
         struct hostlist *hl = NULL;
+        json_error_t error;
         const char *host, *s;
+        const char *bind = NULL;
+        const char *connect = NULL;
+        const char *parent = NULL;
 
-        if (json_unpack (value, "{s:s}", "host", &host) < 0) {
-            log_msg ("Config file error [bootstrap]: missing host field");
-            log_msg ("Hint: hosts entries must be table containing a host key");
+        if (json_unpack_ex (value,
+                            &error,
+                            0,
+                            "{s:s s?s s?s s?s !}",
+                            "host", &host,
+                            "bind", &bind,
+                            "connect", &connect,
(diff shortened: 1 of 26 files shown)


garlick:
Re-pushed with minor changes to error messages and enhanced testing.  Also rebased on current master.

garlick:
Rebased on current master and fixed a small bug.  Please hold off reviewing for just a bit - @grondo had a great idea about topology plugins with a URI style scheme.  I'd like to make sure this is at least compatible with that concept going forward.

trws:
This looks great @garlick.  When discussing some of the results with Bronis earlier today he brought up the idea of trying a binomial tree to see if we get better barrier and broadcast etc. performance out of the overlay.  I think we can try that with this without any modifications as long as the leaf depth doesn't need to be the same for all leaves.

garlick:
Just pushed with some refactoring to support a URI for topology creation that can fit with a (future) plugin scheme for topology generation.  In this PR there are two "builtin plugins", one called  `kary`  and one called `custom` for the parent directives in the bootstrap.hosts array.  Probably we need to allow the topology to be set either on command line or in config.

It should be pretty easy to add code that auto-generates variosu topologies.

garlick:
Repushed wtih `tbon.fanout` replaced wtih `tbon.topo`, and now it can be configured via TOML as well.
To facilitate transition, I left code in place to make `tbon.fanout=K` an alias for `tbon.topo=kary:K`.

This is probably ready for review.

chu11:
going through the commits now, but wondering at a high level, if a parent is not listed for a node, is the parent assumed to be rank 0?

like what if I did the following (which I stole from your commit message example)

```
hosts = [
    { host = "test1" },
    { host = "test[0,64,128,192]", parent = "test1" },
    { host = "test[2-63]",    parent = "test0" },
    { host = "test[65-127]",  parent = "test64" },
    { host = "test[129-191]", parent = "test128" },
    { host = "test[193-255]" }
]
```
would the parent of test[193-255] be test1?  I didn't see a test listed, but might be a good test to add.

side note, i'm only through 2 commits so far, so possible this becomes clear in later commits.

t/t0013-config-file.t
```diff
@@ -177,6 +201,33 @@ test_expect_success 'start size=2 instance with ipc://' '
...
+	host = "fake0"
+	bind = "ipc://${BINDDIR}/fake0"
+	connect = "ipc://${BINDDIR}/fake0"
+	[[bootstrap.hosts]]
+	host="fake1"
+	bind = "ipc://${BINDDIR}/fake1"
+	connect = "ipc://${BINDDIR}/fake1"
+	[[bootstrap.hosts]]
+	host="fake2"
+	parent = "fake1"
```
chu11, line 223: nit: why spaces around equal signs sometimes but not others? :-)

chu11 (commented):
review through commit "[testsuite: add coverage for custom topology](https://github.com/flux-framework/flux-core/pull/4675/commits/1bc6d0c67ab2bf07fc33ca916bf84d407bea29fc)"

src/broker/overlay.c
```diff
@@ -1922,6 +1947,45 @@ static int overlay_configure_zmqdebug (struct overlay *ov)
...
+    }
+    /* Treat tbon.fanout=K as an alias for tbon.topo=kary:K.
+     */
+    const char *fanout;
+    char buf[16];
+    if (attr_get (ov->attrs, "tbon.fanout", &fanout, NULL) == 0) {
+        snprintf (buf, sizeof (buf), "kary:%s", fanout);
+        topo_uri = buf;
+    }
+    if (overlay_configure_attr (ov->attrs,"tbon.topo", topo_uri, NULL) < 0) {
```
chu11, line 1982: nit: space needed here `ov->attrs,"tbon.topo"`. :-)

src/broker/boot_config.c
```diff
@@ -522,6 +488,7 @@ int boot_config (flux_t *h, struct overlay *overlay, attr_t *attrs)
     json_t *hosts = NULL;
     struct topology *topo = NULL;
     flux_error_t error;
+    const char *topo_uri;
```
chu11, line 491: typo commit message "prcedence"

chu11, line 491: Shouldn't the precedences be listed from highest to lowest b/c you list "1, 2, 3" next to them?

```
Add a new attribute: tbon.topo which can be set to a topology URI.
The value is determined by the following in ascending prcedence:
1. compiled-in default ('custom' for config bootstrap, 'kary:2' for PMI)
2. TOML configuration
3. command line
```

src/broker/overlay.c
```diff
@@ -1922,6 +1947,45 @@ static int overlay_configure_zmqdebug (struct overlay *ov)
     return 0;
 }
 
+/* Configure tbon.topo attribute.
+ * Precedence is: compiled-in default, TOML config, command line.
```
chu11, line 1951: perhaps should say increasing precedence or something?

chu11:
high level comment, I think having the default configuration be "custom" even if the user doesn't specify "parent" might sound a tad confusing?  Like if I don't specify a "parent" at all in the config, but the user sees "custom" as the "tbon.topo" config, would that be confusing?

If the user doesn't specify a parent at all, perhaps could have a special default configuration of "flat" (or similar word")?

chu11 (commented):
review; through "[doc: describe tbon.topo and custom topology setup](https://github.com/flux-framework/flux-core/pull/4675/commits/ad3c8a16f70dbbbd82993d4087683da32a823ccb)"

garlick:
> if a parent is not listed for a node, is the parent assumed to be rank 0?

Yes - I hope it's clear in flux-config-bootstrap(5):
> In a `custom` topology, rank 0 is the upstream peer of all other ranks,
unless `parent` keys are present in the `hosts` array to define a
different tree shape.

src/broker/boot_config.c
```diff
@@ -95,29 +95,34 @@ int boot_config_format_uri (char *buf,
...
-    }
-    if (json_array_append_new (hosts, nentry) < 0)
-        goto nomem;
+
+    if (!(nentry = json_deep_copy (entry))
+        || set_string (nentry, "host", hostname) < 0
+        || json_array_append_new (hosts, nentry) < 0) {
+        json_decref (nentry);
+        return -1;
+    }
```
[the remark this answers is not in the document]
garlick: Yeah, the error detail isn't needed by the one function that calls this (internal) one.

t/t0013-config-file.t
garlick: Typo!

src/broker/boot_config.c
garlick: I'll drop the numbers if that's confusing.

garlick:
> high level comment, I think having the default configuration be "custom" even if the user doesn't specify "parent" might sound a tad confusing? Like if I don't specify a "parent" at all in the config, but the user sees "custom" as the "tbon.topo" config, would that be confusing?

I waffled on "custom" vs "kary:0" for that case, but ultimately it seemed simpler (both in code and documentation) to say that the default for config file bootstrap is "custom" unless overridden.

garlick:
Fixed the issues raised so far and forced a push (thanks @chu11!)

chu11:
> I waffled on "custom" vs "kary:0" for that case, but ultimately it seemed simpler (both in code and documentation) to say that the default for config file bootstrap is "custom" unless overridden.

Ahhh it'd be difficult to convey "kary:0" default in some cases but "kary:2" default in other cases.  Yeah, pros and cons on each approach.  If you're not won over with my "flat" idea, then the current is ok.

garlick:
Yeah, I think this way is slightly less confusing so I'll leave it as is unless somebody feels more strongly than me :-)

One thing I do not like in retrospect about the code for `custom` is that I added `topology_hosts_set()` so that the bootstrap code could make the expanded hosts array available to the `custom` plugin.  Instead,  I should add a topology aux container so that arbitrary data can be made available.  Since there's already a `topology_aux_set()` for associating data with ranks, I'll rename that to `topology_rank_aux_set()` which is clearer anyway.  (I think an aux container is probably better in this case than making broker stuff globally available, since this topology class is otherwise standalone.  And anyway the expanded hosts array is ephemeral data so it wouldn't fit that model anyway)

chu11 (approved):
nothing else caught my eye.  this area of the code is not my expertise though, dunno if you want a once over from someone else too.

garlick:
Thanks!  I think this is probably OK and if we get it merged we do have time to tweak it before the next release, if needed.  I'll go ahead and set MWP.

PR #4680, opened 2022-10-12
broker: ignore ENOSYS from parent job-exec.critical-ranks RPC

### LINKED ISSUE #4679: can't start instance >= 0.44.0 under older version of flux
If you try to start an instance of flux >= 0.44.0 under an older version of Flux it fails with:
```
flux-broker: job-exec.critical-ranks: Function not implemented
```

ENOSYS failure from the `job-exec.critical-ranks` RPC to the parent instance should probably just be ignored.

grondo:
Problem: A broker that supports sending its critical ranks to the parent via the job-exec.critical-ranks RPC can't be launched under an older Flux instance that doesn't support this RPC.

Ignore ENOSYS errors from the job-exec.critical-ranks RPC. It should not be a fatal error if this RPC fails.

Fixes #4679

### DIFF
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -1117,7 +1117,7 @@ static int execute_parental_notifications (struct broker *ctx)
         log_err ("job-manager.memo uri");
         goto out;
     }
-    if (flux_future_get (f2, NULL) < 0) {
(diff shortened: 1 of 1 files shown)


garlick (approved):
LGTM

grondo:
Setting MWP (since this bug prevents launching a test buildtree instance within most of our existing system instances)

PR #4683, opened 2022-10-13
Vscode setup

trws:
New setup for vscode and similar.  The commits and messages provide much more detail, but this builds on the devcontainer setup so we get a pretty full featured semi-IDE with all of flux's paths and idioms working and correctly identifying things.

If and when we get the python linting PR in, it would be good to refactor a bit of this, like to have the dockerfile use the requirements from the top level, and to better integrate the settings that has for flake8 and mypy (they are on in here but not well configured since I didn't want to duplicate all that good work).

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -6,9 +6,20 @@ LABEL maintainer="Vanessasaurus <@vsoch>"
 RUN apt-get update
 COPY scripts/requirements-dev.txt /requirements.txt
 
-# For easier Python development. 
+# For easier Python development.
 RUN python3 -m pip install IPython && \
     python3 -m pip install -r /requirements.txt
 
 # Assuming installing to /usr/local
 ENV LD_LIBRARY_PATH=/usr/local/lib
+
+# Assuming installing to /usr/local
+ENV LD_LIBRARY_PATH=/usr/local/lib
+
+# extra interactive utilities and compilation_database generation
+RUN apt-get update \
+ && apt-get -qq install -y --no-install-recommends \
+    bear \
+    ripgrep \
+    fd-find \
+    gdb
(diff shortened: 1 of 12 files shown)


trws:
Bah, forgot to run my rebase before pushing. Should be cleaned up.

grondo (approved):
I can't really review the devcontainer and devenv changes but everything else looks fine to me!

vsoch:
If there is a dependency here on the pre-commit stuff - can we review / merge that first and update here, and then save the double PR? @grondo are you still busy with top secret duties? :secret: :ninja:

vsoch:
I'd like to give this a spin too (after rebase). @trws what is missing for me is a brief description (in the developer docs somewhere maybe?) about how to use this environment, what you've added, etc. Right now not knowing I'd be afraid to just open it up in case it's going to compile something on my local machine without asking!

trws:
Rebased and tweaked the readme to add some more information.  It doesn't build anything, but it does install some extra packages that are generally useful as well as the extensions I pre-configured to give semantic completion for everything.

vsoch:
Okay here are the recommended I see:

image

It gave me the option to install them, so as long as I'm not developing anything locally my preference is to say no, so I'm okay with this setup. Just curious - why should we have a local setup when we have the devcontainer?

I can give more feedback as I use it in the future, but I don't see any alarming issues.

trws:
It's mainly a preference thing, if someone wants to say use the remote extension to develop on one of the clusters, which is already set up as a dev environment, then it will use the local setup and you can install or not install the plugins you want.  Also potentially good for using debugging features, since it's likely we'll end up needing to debug outside of the devcontainer at least sometimes.

trws:
Just realized this is still pending, did a rebase to bring this up to date, cleaned up the one conflict.  Hopefully this will pass through now.

Issue #4604, opened 2022-09-23
idea: collect per-queue job statistics in job-list


grondo:
In job-list we have the `job-list.job-stats` RPC which produces statistics for all the jobs (active and inactive) within an instance. Now that we are close to having official queues support, it might be useful to also have per-queue statistics. The job-list module could dynamically create queue stats on the fly for jobs as soon as the queue attribute is gleaned. The only drawback is that there would not be an entry for a queue that hasn't had any jobs yet, but perhaps callers could deal with that (stats are assumed all zeros)

chu11:
hmmmm, any thoughts on if the current object returned from `job-list.job-stats` should stay the same / backwards compatible or can it change?

```
{"job_states":{"depend":0,"priority":0,"sched":0,"run":0,"cleanup":0,"inactive":0,"total":0},"failed":0,"canceled":0,"timeout":0}
```

isn't the most extensible object for this.  would prefer to do something like:

```
{"all: {job stats ...}, "batch": {job stats ...}, "debug": {job stats ....}}
```

and could perhaps limit returned values if user requests specific queue(s).

a backwards compatible version could be

```
{
/* original above */
"job_states":{"depend":0,"priority":0,"sched":0,"run":0,"cleanup":0,"inactive":0,"total":0},"failed":0,"canceled":0,"timeout":0,

/* queue specific above, but now under "queues" keyword */
"queues": {"all: {job stats ...}, "batch": {job stats ...}, "debug": {job stats ....}}
}
```

but don't like that.

or could just have a new `job-list.queue-stats` rpc.

I think breaking backwards compatibility would be best, I don't think we should expect any users to be getting the raw json of it?

garlick:
Given that there may not be any named queues, it's sort of awkward to return an object with queue names as the keys, since it forces you to hardwire and make invalid a key name for the anonymous queue.  What I settled on in the job manager was returning an array where each entry contained an optional queue name.  So instead of
```json
"queues": {"all": {}, "batch": {}, "debug": {}}
```
use something like
```json
"queues":[
  {"name":"all", "stats":{}},
  {"name":"batch", "stats":{}}, 
  {"name":"debug", "stats":{}}
]
```
the anon queue case would be
```json
"queues":[
   {"stats":{}}
]
```

grondo:
[...]

PR #4684, opened 2022-10-13
job-list: support queue specific stats

chu11:
per #4604, get queue specific job stats.  in `flux-jobs`, the `--queue` option can now also select which queue `--stats` outputs.

### DIFF
--- a/doc/man1/flux-jobs.rst
+++ b/doc/man1/flux-jobs.rst
@@ -74,23 +74,27 @@ OPTIONS
    Defaults to *auto*.
 
 **--stats**
-   Output a summary of global job statistics before the header.
-   May be useful in conjunction with utilities like
-   :linux:man1:`watch`, e.g.::
+   Output a summary of job statistics before the header.  By default
+   shows global statistics.  If ``--queue`` is specified, shows
+   statistics for the specified queue.  May be useful in conjunction
+   with utilities like :linux:man1:`watch`, e.g.::
 
       $ watch -n 2 flux jobs --stats -f running -c 25
 
-   will display a summary of global statistics along with the top 25
+   will display a summary of statistics along with the top 25
    running jobs, updated every 2 seconds.
 
 **--stats-only**
-   Output a summary of global job statistics and exit.
-   ``flux jobs`` will exit with non-zero exit status with ``--stats-only``
-   if there are no active jobs. This allows the following loop to work::
+   Output a summary of job statistics and exit.  By default shows
+   global statistics.  If ``--queue`` is specified, shows statistics
+   for the specified queue.  ``flux jobs`` will exit with non-zero
+   exit status with ``--stats-only`` if there are no active jobs. This
+   allows the following loop to work::
 
        $ while flux jobs --stats-only; do sleep 2; done
 
-   All other options are ignored when ``--stats-only`` is used.
+   All options other than ``--queue`` are ignored when
+   ``--stats-only`` is used.
 
 **-R, --recursive**
    List jobs recursively. Each child job which is also an instance of
--- a/src/modules/job-list/stats.c
+++ b/src/modules/job-list/stats.c
@@ -42,11 +42,11 @@ void job_stats_update (struct job_stats *stats,
                        struct job *job,
                        flux_job_state_t newstate)
 {
-    stats->state_count[state_index(newstate)]++;
+    stats->state_count[state_index (newstate)]++;
 
     /*  Stats for NEW are not tracked */
     if (job->state != FLUX_JOB_STATE_NEW)
-        stats->state_count[state_index(job->state)]--;
+        stats->state_count[state_index (job->state)]--;
 
     if (newstate == FLUX_JOB_STATE_INACTIVE && !job->success) {
         stats->failed++;
@@ -65,7 +65,7 @@ void job_stats_purge (struct job_stats *stats, struct job *job)
 {
     assert (job->state == FLUX_JOB_STATE_INACTIVE);
 
-    stats->state_count[state_index(job->state)]--;
+    stats->state_count[state_index (job->state)]--;
 
     if (!job->success) {
         stats->failed--;
--- a/src/modules/job-list/stats.c
+++ b/src/modules/job-list/stats.c
@@ -18,6 +18,14 @@
 #include "stats.h"
 #include "job_data.h"
 
+static void free_wrapper (void **item)
+{
+    if (item) {
+        free (*item);
+        (*item) = NULL;
+    }
+}
+
 struct job_stats_ctx *job_stats_ctx_create (flux_t *h)
 {
     struct job_stats_ctx *statsctx = NULL;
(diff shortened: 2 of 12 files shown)


chu11:
re-pushed, fixed up && chain in tests, valgrind memleak find, and python linting

src/modules/job-list/stats.c
```diff
@@ -157,12 +210,69 @@ static json_t *stats_encode (struct job_stats *stats)
         return NULL;
     }
     json_decref (states);
+
+    if (name) {
+        json_t *no = json_string (name);
+        if (!no || json_object_set (o, "name", no) < 0) {
```
garlick, line 216: This leaks `no` on success.  I think you meant `json_object_set_new()`?

src/modules/job-list/stats.c
```diff
@@ -157,12 +210,69 @@ static json_t *stats_encode (struct job_stats *stats)
...
+        errno = ENOMEM;
+        return -1;
+    }
+
+    stats = zhashx_first (statsctx->queue_stats);
+    while (stats) {
+        const char *name = zhashx_cursor (statsctx->queue_stats);
+        json_t *qo = stats_encode (stats, name);
+        if (!qo) {
+            json_decref (queues);
```
garlick, line 242: json_decref() clobbers errno

src/modules/job-list/stats.c
```diff
@@ -157,12 +210,69 @@ static json_t *stats_encode (struct job_stats *stats)
...
+            json_decref (o);
+            errno = ENOMEM;
+            return NULL;
+        }
+    }
     return o;
 }
 
+static int queue_stats_encode (struct job_stats_ctx *statsctx,
+                               json_t *o)
```
garlick, line 227: The function name does not suggest that the stats are going to be inserted in `o`.  Maybe have the caller do that?

src/modules/job-list/stats.c
```diff
@@ -103,9 +151,14 @@ static void stats_purge (struct job_stats *stats, struct job *job)
  */
 void job_stats_purge (struct job_stats_ctx *statsctx, struct job *job)
 {
+    struct job_stats *stats;
+
     assert (job->state == FLUX_JOB_STATE_INACTIVE);
```
garlick, line 156: An assertion here kills the rank 0 broker.  Appropriate in this case?

doc/man1/flux-jobs.rst
```diff
@@ -74,23 +74,27 @@ OPTIONS
    Defaults to *auto*.
 
 **--stats**
-   Output a summary of global job statistics before the header.
-   May be useful in conjunction with utilities like
-   :linux:man1:`watch`, e.g.::
+   Output a summary of job statistics before the header.  By default will
+   output global statistics.  If ``--queue`` is specified, will output
+   statistics for the specified queue.  May be useful in conjunction with
+   utilities like :linux:man1:`watch`, e.g.::
```
garlick, line 80: Maybe s/will output/show/?

doc/man1/flux-jobs.rst
```diff
@@ -74,23 +74,27 @@ OPTIONS
...
 
 **--stats-only**
-   Output a summary of global job statistics and exit.
-   ``flux jobs`` will exit with non-zero exit status with ``--stats-only``
-   if there are no active jobs. This allows the following loop to work::
+   Output a summary of job statistics and exit.  By default will output
+   global statistics.  If ``--queue`` is specified, will output
+   statistics for the specified queue.  ``flux jobs`` will exit with
+   non-zero exit status with ``--stats-only`` if there are no active
+   jobs. This allows the following loop to work::
```
garlick, line 92: s/will output/show/

garlick (commented):
I tried running this branch on my test system that has two queues (batch and debug).  Initially, I ran no new jobs and got no per queue stats reported even though there were inactive jobs in the KVS.  Then I ran a job in the batch queue that happened to fail, and got these stats
```
 garlick@picl0:~/proj/flux-core$ flux jobs --stats --queue=batch
0 running, -1 completed, 1 failed, 0 pending
```
so something's a little off there.

chu11:
> so something's a little off there.

hmmm.  maybe a bug related to restarting of a flux instance, which I may have not added a test for.  Thanks!

src/modules/job-list/stats.c
chu11: i think that assert was meant as a double check that this is only called on inactive jobs, but I see

```
    if (flux_event_unpack (msg, NULL, "{s:o}", "jobs", &jobs) < 0)                                                                          
        flux_log_error (h, "job-purge-inactive message");                                                                                   
    json_array_foreach (jobs, index, entry) {                                                                                               
        flux_jobid_t id = json_integer_value (entry);                                                                                       
        struct job *job;                                                                                                                    
                                                                                                                                            
        if ((job = zhashx_lookup (ctx->jsctx->index, &id))) {                                                                               
            job_stats_purge (ctx->jsctx->statsctx, job);                                                                                    
            if (job->list_handle)                                                                                                           
                zlistx_delete (ctx->jsctx->inactive, job->list_handle);                                                                     
            zhashx_delete (ctx->jsctx->index, &id);                                                                                         
            count++;                                                                                                                        
        }                                                                                                                                   
    }                                                          
```

that we do not verify this fact.  it trusts the fact that every job id passed to it is inactive.  We should double check that within the loop.

I'll create an issue, b/c this is independent of this PR.

src/modules/job-list/stats.c
```diff
@@ -18,6 +18,24 @@
...
+        return NULL;
+    }
+    statsctx->h = h;
+    return statsctx;
+}
+
+void job_stats_ctx_destroy (struct job_stats_ctx *statsctx)
+{
+    if (statsctx)
+        free (statsctx);
```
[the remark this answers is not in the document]
chu11: ahh, the `if (statsctx)` was a little bit "with future expectations of need", should be in a later commit instead.

chu11:
re-pushed, fixing up the issues @garlick found above

fixed queue specific stats on reload of the job-list module and added some tests & extra coverage for it

However, I was not able to figure out how

> 0 running, -1 completed, 1 failed, 0 pending

occurred though.  It's perplexing.  Basically in the python code the number of completed jobs is:

`self.successful = self.inactive - self.failed`

suggesting inactive job count was < the failed job count to get the `-1` output.  Which I can't see a way in which that could happen.  

My only guess is it could have been a counting issue with a job that was active when the job-list module was reloaded and one of the counters may have been under-counted.  But I don't think that's what happened here.

garlick:
> My only guess is it could have been a counting issue with a job that was active when the job-list module was reloaded and one of the counters may have been under-counted. But I don't think that's what happened here.

This might be unrelated to this PR.  Watch this:
```
 garlick@picl0:~$ flux jobs --stats-only
0 running, 87 completed, 13 failed, 0 pending
 garlick@picl0:~$ flux mini run -q debug /bin/false
 garlick@picl0:~$ flux jobs --stats-only
0 running, 86 completed, 14 failed, 0 pending
```
Somehow the completed count dropped.  This is on master right after #4687 (emoji) was merged.

garlick (approved):
Just one minor comment.  I tried this again on my system with queues configured and it seems to behave as expected!

src/modules/job-list/stats.c
```diff
@@ -18,19 +18,62 @@
...
+
+    if (!job->queue)
+        return NULL;
+
+    stats = zhashx_lookup (statsctx->queue_stats, job->queue);
+    if (!stats) {
+        if (!(stats = calloc (1, sizeof (*stats))))
+            return NULL;
+        if (zhashx_insert (statsctx->queue_stats, job->queue, stats) < 0) {
+            flux_log_error (statsctx->h, "%s: zhashx_insert", __FUNCTION__);
```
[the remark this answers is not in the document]
chu11: oops, it seems I do that all over the place and with `zlistx` functions too.  I'll leave this here for now, do a bigger cleanup in a followup PR.

PR #4688, opened 2022-10-14
README: update flux help output

chu11:
Just noticed the flux help output was out of date.

I didn't remove the readthedocs pointer in the helper output to keep it 100% identical, but maybe should remove?


grondo (approved):
Nice catch! Seems fine to leave the `flux help` output verbatim.

PR #4636, opened 2022-10-01
ci: consolidate python linting and formatting

vsoch:
Problem: the developer experience is challenging because linting and formatting are hard to debug locally, and then the separation in the CI means 3+ clicks to see what is going on! To fix this, this changeset adds the use of pre-commit, which will always run linting/formatting (and with automated fixes for a subset of the tools) right before commit, so the developer can have some confidence that passing locally == passing in the CI. In the case that there is a bug in the CI, we will now just have one place to look instead of 3+. This should be a start of a new setup, including addition of pre-commit and consolidation of settings into a pyproject.toml (and setup.py for flake8) and we can further tweak as we establish our linting preferences. In summary:

- I've added pre-commit and a scripts/requirements.dev.txt to install all linting/ formatting deps
- I've added a developer.rst to the Python docs to show how to use this
- I've removed the custom linting shell scripts in scripts (in practice I found running them gave me different results than the CI, likely because of versioning)
- I've added a pyproject.toml with preferences for mypi, and black
- To make the developer.rst complete, I included the instructions I use to develop Python (and removed the --without-python in the container).

Note that isort and flake8 are currently not being added, as we learned in #4634 that they broke the CI! My proposal is to get this into master, and then do incremental changes to figure out what broke the testing. If it's an issue that import order is important, it's fine to leave as is, but I think we need to understand why (and document it).

And here is a previous screenshot in action! Note that isort/flake8 have since been disabled. Note that it's running the checks just before I commit, so I can fix any issues before I do.

image

And pre-commit  is super cool - it's like being in a candy store! https://github.com/pre-commit/pre-commit-hooks
I also tested with rebase to ensure it works! This will close #4631 

Signed-off-by: vsoch <vsoch@users.noreply.github.com>

PR #4689, opened 2022-10-14
Pre-commit extensions and multi-version setup

trws:
Following on to #4636 this will do the following:

* Enable flake8, fix associated issues
* Allow running pre-commit with any compatible version of python, due to version range constraints I think the limit is 3.6 to 3.9 due to dependency compatibility
* Add a vermin check to ensure that python 3.6 keeps working without actually having to *use* 3.6, and add a comment to ignore the one issue we had where the analysis fails
* Update versions of *everything* other than what @vsoch added, because some of our tools are old enough they can't run on newer versions of python and are much more likely to miss things or give us false positives.  I think these versions were chosen to support python 2, we need to fix them.  Also solving associated issues (though I've found very few).


trws:
Next thing after this will be to add in the c/c++ linters and just have it run everywhere, trick there will be excluding all the external projects we don't want it to process.  Is there a central list of those somewhere?

grondo:
> Next thing after this will be to add in the c/c++ linters and just have it run everywhere, trick there will be excluding all the external projects we don't want it to process. Is there a central list of those somewhere?

You can reference the code coverage ignore list I think:

https://github.com/flux-framework/flux-core/blob/7fec12d79f7dd7f38a82f0ab800a5fa1dd9a6bed/Makefile.am#L17-L32

grondo (approved):
This is working for me!

trws:
Sweet! Then if this manages to pass, I can work the other tools in with the vscode PR I had started (it needs some of the stuff anyway, doing it there will cut the duplication down a good bit).  I'm good with MWP here if you are @grondo.

Also thanks for the coverage ignore list idea, I completely forgot about that but that's perfect.

PR #4690, opened 2022-10-15
lint: update devcontainer to work with pre-commit

vsoch:
I had wanted this to be a scoped PR to just update line length (to the suggested 100 by Todd) but the devcontainer didn't work for me - the black version installed was broken with ubuntu focal. I have restored the container to its working state and added the line length fixes!

After this I'll tackle isort.


grondo:
Is there a reason to update to 100?  I prefer the current default and had thought 100 was specific to spack.

vsoch:
I like them a little longer, and I thought we were trying to follow Todd’s suggestions?

grondo:
I think @trws had suggested the current default would be fine, and the spack line length was chosen to minimize the changes when first applying a new default maximum. To minimize changes here we can just keep the `black` default? 🙂

vsoch:
Ok, I’ll update the PR to delete those lines then (I must have missed this conversation!)

vsoch:
okay that should do it!

vsoch:
_Narrator: The dinosaur tried, with all her might, to do the commit message correctly this time, but she failed to realize her carefully chosen "fixup" was in fact, a terrible choice._

grondo:
Thanks! I'm fine with this if @trws is! I had thought he had a way to get the `focal` container to work, but I'm not sure, so I'll let him approve and set MWP.

trws:
Yup, I broke it. 😭 

Sorry @vsoch, when I tried from a clean container I got basically the same issue.  It's because I missed propagating the black update from the requirements-dev.txt to the container.  If you're amenable I'd prefer to handle it like this:
* move the requirements-dev.txt to the .devcontainer folder
* symlink that back to somewhere less annoying (I tried doing it the other direction but docker refuses to read through it, and vscode refuses to run build from the root of the repo)
* use the requirements file from the dockerfile, with a COPY or ADD or whatever, so we don't have to manually keep them up to date

The root problem is that black <22.3 has the wrong version constraint for click, and it breaks, as long as we have the black >=22.4 the problem goes away and we don't need bionic.  What do you think @vsoch?

Would probably be a good thing to have an action that builds the devcontainer, but that's definitely a new PR.

vsoch:
@trws I actually tried a copy of the file but it didn't work for me - thinking about it now, it likely was an issue of the "scripts" folder being outside of the build context. I can probably tweak the context to be able to use the shared file, so let me try that and I'll get back to you!

vsoch:
yeaaaah dawg that totally worked! Will update shortly.

vsoch:
okay all set! This maintains the focal image but now the linting checks work for me!

.devcontainer/devcontainer.json
```diff
@@ -1,7 +1,8 @@
 {
     "name": "Flux Core Python 3.6",
     "dockerFile": "Dockerfile",
-  
+    "context": "../",
```
trws, line 4: Excellent! This is what I didn't find, much better solution.

trws (approved):
Looks great, thanks @vsoch!

.devcontainer/devcontainer.json
vsoch: For future us - the walk through guides aren't very detailed in what you can put in this file, but this guide (assuming you mostly know what the variables mean) is! I'll leave here for future us to find: https://containers.dev/implementors/json_reference/

vsoch: And you know what' really funny - I almost bought this domain back when .dev was new!! https://containers.dev/ Glad I didn't, I would have probably wasted it on something silly. :laughing:

PR #4691, opened 2022-10-15
ci: add isort to pre-commit and linting

vsoch:
I intend to incrementally test isort - and because the ultimate source of truth are the tests here (and I couldn't reproduce locally) it makes most sense to test in the CI on GitHub. I [tried opening this on my repository](https://github.com/researchapps/flux-core/pull/2) to not add traffic here, but the tests already failed, so unless my first change hit the sorting order issue, I think I'll need to test here. Understandably this might use a lot of CI juice so I'll try to do it in off hours and weekends! You can safely ignore this branch for the time being - I need to carefully change files in small groups until tests break (and maybe they already have, lol).

### Current Failures

The following, when isort-ed, break tests.

- flux/resources:
  - `__init__.py`
- flux/job:
  - `__init__.py`
  - kvs.py
- flux/kvs.py (this one really doesn't make sense to me, but failed twice!)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,11 +20,12 @@ repos:
         types: [python]
         entry: black
 
-#      - id: isort
-#        name: isort
-#        language: python
-#        types: [python]
-#        entry: isort
+      - id: isort
+        name: isort
+        args: [--filter-files]
+        language: python
+        types: [python]
+        entry: isort
 
       - id: flake8
         name: flake8
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,8 +4,11 @@ target-version = ['py36']
 exclude = ["^env/", "^src/bindings/python/flux/utils/*/*.py", "^src/bindings/python/flux/utils/*/*/*.py"]
 
 [tool.isort]
-profile = "isort"
-exclude = ["^env/"]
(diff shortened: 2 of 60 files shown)


vsoch:
okay first potential finding - we had two docker failures when I isort-ed the resources directory: https://github.com/flux-framework/flux-core/actions/runs/3257626868/jobs/5349009794 

I just pushed a change to undo that, and we will see if the undo fixes this break!

trws (approved):
Looks great!  It would probably be good to look into why those three fail in a separate PR, just makes me curious, but I'm happy to see this working.

trws:
@grondo, want to take a glance before MWP?

vsoch:
You beat me! I was waiting for all tests to pass before converting from draft state - I'll do that now!

vsoch:
@trws after this PR let's go back to your devcontainers work, rebase with this, and we can make sure the whole pre-commit flow works with your changes! And I'd like to give it a test run too in my VSCode (VS means Vanessa Saurus Code).

trws:
That's a good idea, make sure we have the right isort versions for the range and everything.  Thanks @vsoch!

grondo (approved):
This looks good to me! Might as well get this done sooner rather than later.

PR #4694, opened 2022-10-17
update clang-format and format some files as test

trws:
Update clang-format using some options from versions up to 11, and format some files with version 12 so we can see what they look like.  This is not really meant to be merged, but to pick a few large-ish files with various code patterns so we can figure out if this is what we want.

The only thing that really stood out to me in here is the `ok ()` and similar functions/macros seem like they are almost always formatted in a non-standard way.  Should those go in with the pack/unpack functions so they're left alone?

### DIFF
--- a/src/common/libsubprocess/server.c
+++ b/src/common/libsubprocess/server.c
@@ -9,7 +9,7 @@
 \************************************************************/
 
 #if HAVE_CONFIG_H
-# include "config.h"
+#include "config.h"
 #endif
 
 #include <sys/types.h>
@@ -37,8 +37,8 @@
 static const char *auxkey = "flux::rexec";
 
 struct rexec {
-    const flux_msg_t *msg;          // rexec request message
-    flux_subprocess_server_t *s;    // server context
+    const flux_msg_t *msg;        // rexec request message
+    flux_subprocess_server_t *s;  // server context
 };
 
 static void rexec_destroy (struct rexec *rex)
@@ -49,8 +49,7 @@ static void rexec_destroy (struct rexec *rex)
     }
 }
 
-static struct rexec *rexec_create (const flux_msg_t *msg,
-                                   flux_subprocess_server_t *s)
+static struct rexec *rexec_create (const flux_msg_t *msg, flux_subprocess_server_t *s)
 {
     struct rexec *rex;
 
@@ -155,7 +154,8 @@ static void rexec_completion_cb (flux_subprocess_t *p)
         /* no fallback if this fails */
         if (flux_respond_pack (rex->s->h, rex->msg, "{s:s s:i}",
                                "type", "complete",
-                               "rank", rex->s->rank) < 0)
+                               "rank", rex->s->rank)
+                               < 0)
             flux_log_error (rex->s->h, "%s: flux_respond_pack", __FUNCTION__);
     }
 
@@ -182,8 +182,7 @@ static void internal_fatal (flux_subprocess_server_t *s, flux_subprocess_t *p)
     }
 }
 
-static void rexec_state_change_cb (flux_subprocess_t *p,
-                                   flux_subprocess_state_t state)
+static void rexec_state_change_cb (flux_subprocess_t *p, flux_subprocess_state_t state)
 {
     struct rexec *rex = flux_subprocess_aux_get (p, auxkey);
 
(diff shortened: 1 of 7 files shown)


src/common/libsubprocess/server.c
```diff
@@ -155,7 +154,8 @@ static void rexec_completion_cb (flux_subprocess_t *p)
         /* no fallback if this fails */
         if (flux_respond_pack (rex->s->h, rex->msg, "{s:s s:i}",
                                "type", "complete",
-                               "rank", rex->s->rank) < 0)
+                               "rank", rex->s->rank)
+                               < 0)
```
trws, line 158: Looking back at this now, I'm wondering what was left to adjust to make this good enough. The thing I remember trying to work around is the break before trailing `< 0` in if statements, which is why this is a comment on these lines as an example.  Oddly though, looking back at it I kinda don't hate it.  It makes it visually obvious what's being compared here, it can be avoided by wrapping in parentheses, and if we're happy with the rest I'm not sure it's worth blocking on.  @garlick, @grondo, @chu11, thoughts?

grondo: I really don't like the break before `< 0` which is why I suppose we don't do it in our existing formatting.

trws: Odd as it seems, we might also be able to address it with a bit of scripting.  It's a very specific spot, and a vim script of `%g/^\s+\(\!=|==|<|>\) \d.{0,5}$/normal kJ` seems to fix it 100% of the time.  The only downside is that it would mean in a few cases (less than a dozen total I think) we'll end up with lines up to 5 chars longer than the limit set in the clang-format file.  I'm going to throw that into a wrapper script, maybe convert it to sed, and throw up the update.

trws: Updated versions of the test files pushed, I made a small mistake so there's a blank line at the beginning of the files, but the trailing comparison operator is taken care of.  Need to figure out how to make it reasonably nice to use this, awk makes the process a bit slow, but does that otherwise take care of it?

trws:
Huh, one other option I didn't know existed (not sure it did last I checked), but that makes things more consistent and actually works a lot like the black format for arguments would be to switch AlignAfterOpenBracket from Align to BlockIndent.  That would mean that instead of aligning all function arguments to the indent of the open parentheses, they would all get the continuation indent, which looks like this:

```c
 if ((port1 = get_port (triple->range)) < 0 || (port2 = get_port (triple->range)) < 0
 ┆   || !(arr = json_pack ("[I, I]", port1, port2))
 ┆   || flux_jobtap_event_post_pack (
 ┆   ┆   ┆  triple->plugin,
 ┆   ┆   ┆  triple->jobid,
 ┆   ┆   ┆  "cray_port_distribution",
 ┆   ┆   ┆  "{s:O, s:I}",
 ┆   ┆   ┆  "ports",
 ┆   ┆   ┆  arr,
 ┆   ┆   ┆  "random_integer",
 ┆   ┆   ┆  random
 ┆   ┆  ) < 0
 ┆   || flux_jobtap_job_aux_set (
 ┆   ┆   ┆  triple->plugin,
 ┆   ┆   ┆  triple->jobid,
 ┆   ┆   ┆  CRAY_PALS_AUX_NAME,
 ┆   ┆   ┆  arr,
 ┆   ┆   ┆  (flux_free_f)json_decref
 ┆   ┆  ) < 0) {
 ┆   if (arr)
 ┆   ┆   json_decref (arr);
 ┆   prolog_status = 1;
 ┆   flux_log_error (
 ┆   ┆   h,
 ┆   ┆   PLUGIN_NAME
 ┆   ┆   ": "
 ┆   ┆   "Failed to post ports to job"
 ┆   );
 ┆   goto cleanup;
 }
```

I'm oddly fond of this and interested in what everyone else thinks.  Having gotten used to it with black (and how it avoids extra diff gak in argument lists, literals and so-forth) it works for me more than it once would have, and it naturally keeps the `< 0` attached to the close-peren for some reason.

With the current value it looks like this (with the fixup script):

```c
 if ((port1 = get_port (triple->range)) < 0 || (port2 = get_port (triple->range)) < 0
 ┆   || !(arr = json_pack ("[I, I]", port1, port2))
 ┆   || flux_jobtap_event_post_pack (triple->plugin,
 ┆   ┆   ┆   ┆   ┆   ┆   ┆   ┆   ┆   triple->jobid,
 ┆   ┆   ┆   ┆   ┆   ┆   ┆   ┆   ┆   "cray_port_distribution",
 ┆   ┆   ┆   ┆   ┆   ┆   ┆   ┆   ┆   "{s:O, s:I}",
 ┆   ┆   ┆   ┆   ┆   ┆   ┆   ┆   ┆   "ports",
 ┆   ┆   ┆   ┆   ┆   ┆   ┆   ┆   ┆   arr,
 ┆   ┆   ┆   ┆   ┆   ┆   ┆   ┆   ┆   "random_integer",
 ┆   ┆   ┆   ┆   ┆   ┆   ┆   ┆   ┆   random) < 0
 ┆   || flux_jobtap_job_aux_set (triple->plugin,
 ┆   ┆   ┆   ┆   ┆   ┆   ┆   ┆   triple->jobid,
 ┆   ┆   ┆   ┆   ┆   ┆   ┆   ┆   CRAY_PALS_AUX_NAME,
 ┆   ┆   ┆   ┆   ┆   ┆   ┆   ┆   arr,
 ┆   ┆   ┆   ┆   ┆   ┆   ┆   ┆   (flux_free_f)json_decref) < 0) {
 ┆   if (arr)
 ┆   ┆   json_decref (arr);
 ┆   prolog_status = 1;
 ┆   flux_log_error (h,
 ┆   ┆   ┆   ┆   ┆   PLUGIN_NAME
 ┆   ┆   ┆   ┆   ┆   ": "
 ┆   ┆   ┆   ┆   ┆   "Failed to post ports to job");
 ┆   goto cleanup;
 }
```

garlick:
@trws, even though the source changes are massively out of date here, the .clang-format changes are an improvement over what's there.

Any objection if I push to this branch to limit the changes to only the .clang-format file?  Then we can get that merged and it can be incrementally improved from there.

garlick:
Eh, let's close this and I'll just propose the new .clang-format in a new PR referencing this one and with authorship attributed to you.

Issue #4678, opened 2022-10-12
add flux config load


garlick:
> Yeah, we should get a separate issue open for the `flux config reload --stdin` (or should it just be `flux config load`?) option. This is the only way you can configure an instance that has already started without a configuration directory set. For instance, this would give users a way to configure their batch jobs from within the batch script.

_Originally posted by @grondo in https://github.com/flux-framework/flux-core/issues/4676#issuecomment-1276659750_

grondo:
I'd argue for a `flux config load` since it obviates the need for `--stdin` option, and if you are sending configuration over stdin you aren't really "reloading". Also, it might be nice if this command explicitly _did not_ reload from config files, so you could temporarily override configuration until the next `flux config reload`.

PR #4695, opened 2022-10-18
flux config: add load subcommand

garlick:
As discussed in #4678, this adds a new `flux config load` subcommand that takes a configuration object on stdin (JSON or TOML) and directly updates the broker.   E.g. you can do stuff like this:
```sh
flux config load <<EOF
tbon.tcp_user_timeout = "2s"
EOF
```
or
```sh
flux config get | jq '.tbon.tcp_user_timeout = "3s"' | flux config load
```
In this PR, the entire config object is replaced with each invocation of `flux config load`, which seems like the safest way to use this in test, since it makes it clear that no config lingers from earlier tests.   A simple extension (perhaps for a future PR?) would be to add an `--update` option to let changes be applied incrementally, e.g.  [json_object_update()](https://jansson.readthedocs.io/en/latest/apiref.html#c.json_object_update) style, which would be a bit more natural than the  `jq` example above.

### DIFF
--- a/src/cmd/builtin/config.c
+++ b/src/cmd/builtin/config.c
@@ -11,10 +11,13 @@
 #if HAVE_CONFIG_H
 #include "config.h"
 #endif
+#include <unistd.h>
 #include <jansson.h>
 
 #include "src/common/libutil/jpath.h"
 #include "src/common/libutil/fsd.h"
+#include "src/common/libutil/read_all.h"
+#include "src/common/libutil/tomltk.h"
 #include "ccan/array_size/array_size.h"
 
 #include "builtin.h"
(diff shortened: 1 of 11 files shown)


chu11:
somehow "config load" doesn't sound right to me.  I think the main thing is it sounds like it "goes with" "config reload".  Like someone might normally want to do:

```
flux config load < foo
flux config reload
```

I would guess this would scan the toml files and update the config with whatever is in the toml files in the directory?  (as an aside, might be a good test to add)

idea, "flux config overwrite"?

chu11 (changes requested):
LGTM except for my comment about is "flux config load" the right word / name to use.

DAMMIT, I clicked "requested changes" by accident as I was scrolling down to submit ... how do I undo this...

chu11 (dismissed):

garlick:
NP!

> I would guess this would scan the toml files and update the config with whatever is in the toml files in the directory? (as an aside, might be a good test to add)

In the `--config-path=DIR` directory, yes.  It's a no-op if one wasn't set on the broker command line.

I thought load was OK but I kind of see your point about load/reload.  Well, we do have a `flux config get` that (by default) dumps the config to stdout.  Maybe `flux config set` that reads the config from stdin makes sense?   What do you think @grondo?

chu11:
> Maybe flux config set that reads the config from stdin makes sense?

I like the idea of `flux config set` b/c it does sound like it "goes with" `flux config get`.  I could see users use `flux config get` to get the current config and modify it by hand.

Given this potential use case, should `flux config get` also take a filename as input instead of just stdin?

grondo:
> Given this potential use case, should flux config get also take a filename as input instead of just stdin?

I assume you meant `flux config set` here? 

I still think `flux config load` is better, but would it be less confusing if `flux config load` took an optional `confdir` on the commandline, and if given, updated the internal `confdir` so that a subsequent `reload` reloaded from the new path?

grondo:
After reading this back I can see an argument for `flux config set` vs `flux config load`, but I still can't get past the fact that `flux config get` can get one individual value, but `flux config load` overwrites everything, so really they are not counterparts at all. I say we reserve `flux config set` for when one value can be set instead of everything?

chu11:
> I assume you meant flux config set here?

Oops yeah, `flux config set`

chu11:
> I still think flux config load is better, but would it be less confusing if flux config load took an optional confdir on the commandline, and if given, updated the internal confdir so that a subsequent reload reloaded from the new path?

i like this idea, then `load` and `reload` go together better.

garlick:
Ah that sounds pretty good.  OK I'll try adding `flux config load --config-path=DIR` (matching the broker option).

garlick:
>  updated the internal confdir so that a subsequent reload reloaded from the new path

I'm a bit worried about this part of it though.  If a sys admin was tempted to `flux config load --config-dir=DIR` on a cluster node to test out a config and forgets about it, they might be surprised to find later that this broker ignored new configs pushed out via config management (because it no longer reloads from the controlled directory).  My vote would be to let a one-off TOML directory be loaded using `flux config load --config-path=DIR` but _not_ update the internal confdir.

Thoughts?

grondo:
>  My vote would be to let a one-off TOML directory be loaded using flux config load --config-path=DIR but not update the internal confdir.

That seems fine to me. I wasn't sure which case would be more confusing honestly so I'm personally fine with either way.

grondo:
> `flux config load --config-path=DIR`

This seems fine and I do like the parity with the broker option, but you could also just support `flux config load DIR` -- is the worry that this is really just a test option at this point and we don't want to make it _too_ easy?

garlick:
I think  `flux config load DIR` is fine.  I'll do that.

chu11:
> flux config load DIR

This seems fine, and I guess stdin if no DIR specified.

src/cmd/builtin/config.c
```diff
@@ -192,6 +195,60 @@ static int internal_config_reload (optparse_t *p, int ac, char *av[])
...
+        flux_conf_t *conf;
+        flux_error_t error;
+
+        if (!(conf = flux_conf_parse (path, &error))
+            || flux_conf_unpack (conf, &error, "O", &obj) < 0)
+            log_msg_exit ("Error parsing config: %s", error.text);
+        flux_conf_decref (conf);
+    }
+    else {
+        if ((n = read_all (STDIN_FILENO, &buf)) < 0)
```
chu11, line 224: can move `buf` and `size` vars into this block b/c its only used here

chu11 (approved):
overall LGTM, just one nit below

src/cmd/builtin/config.c
garlick: Good catch.  OK, setting MWP - thanks!

Issue #3817, opened 2021-08-07
idea: add a method to determine the context within Flux in which a process is currently running


grondo:
As noted in #3744, it would be useful to have some way for a process to determine the context in which it is running as it relates to Flux jobs, initial program, etc. Off the top of my head, I can think of a few different contexts we might want to delineate:

 1. Not within any Flux instance (`flux_open ()` fails with `ENOENT`)
 2. Enclosing instance is the multi-user system instance (`instance-level` attribute is `0`, `security.owner` != current UID, `jobid` attribute not set)
 3. Enclosing instance is a job in a foreign RM or `flux start --test-size` session, and process is running as part of initial program (same as above, but `security.ower` == current uid) 
 4. Enclosing instance is a Flux job and process is running as part of initial program
 5. Enclosing instance is a Flux job and process is part of a job within that instance

AFAICT, there is not a good way to easily determine the difference between 4 and 5 above. Perhaps less importantly, there is not a _clean_ way to tell the difference between 2 and 3 either (in the case a process is running with the UID of the `flux` user for example)

It might be nice if we could add a function that would return "something" to allow a process to differentiate between these different contexts. Since "context" is actually a bit of an overloaded term, we might need something different, but the only idea I've come up with so far is to have a set of named process "scopes".
This should be just considered an early idea at this point and we can iterate as much as people desire, or even throw out this idea as unnecessary if it will cause too much confusion.

Here's a first cut at names for the "scopes" outlined above:

 - 1: none
 - 2: system
 - 3,4: initial program (I suppose instance-level could be used to differentiate these two)
 - 5:  job

We could add an API call `flux_get_process_scope(3)` which would return one of these strings, and would allow programs to alter behavior based on their current context. For the example of `flux bcast` it could abort with a warning if run in `job` scope since it likely doesn't make sense to run that command as a job.

A `flux scope` command could simply print the result of `flux_get_process_scope(3)` for use in scripts, etc.

grondo:
In discussing the repercussions of our inability to determine if a process is running in the "scope" of a Flux instance or a job within a Flux instance with @ofaaland, we had the idea to use a simple environment variable set by the job shell, but cleared by the `flux-broker`. Keying off this environment variable would allow `flux_get_process_scope()` or similar to determine whether the scope is `job` or `initial program` (perhaps `instance` is a better name for that one, I don't know)

This would be trivial to implement and would assist @ofaaland's use case immediately.

[...]

PR #4699, opened 2022-10-19
libflux: support flux_get_process_scope()

chu11:
Per #3817.  This is super WIP with no testing yet, but wanted to post to get initial high level feedback.

- basic logic within function for the 5 categories discussed in #3817
- should be 4 categories instead of 5?  some discussion for this in #3817 too
- API style / enum names
- location of the function, I put it in new files `libflux/util.[ch]` ... but could go in `libjob` instead? but that's not the super right place either ... 

TODO
- the command
- tons of tests
- do we need a nice python wrapper?
- document it / add nice comments & examples

### DIFF
--- a/src/common/libflux/util.h
+++ b/src/common/libflux/util.h
@@ -1,0 +1,56 @@
+/************************************************************\
+ * Copyright 2022 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+#ifndef _FLUX_CORE_UTIL_H
+#define _FLUX_CORE_UTIL_H
+
+#include <flux/core.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef enum flux_process_scope {
+    FLUX_PROCESS_SCOPE_NONE = 0,
+    FLUX_PROCESS_SCOPE_SYSTEM_INSTANCE = 1,
+    FLUX_PROCESS_SCOPE_INITIAL_PROGRAM = 2,
+    FLUX_PROCESS_SCOPE_JOB = 3,
+} flux_process_scope_t;
+
+/* Retrieve information on the scope of a job
+ *
+ * NONE - process is not running under flux
+ *
+ * SYSTEM_INSTANCE - process is running under the system instance
+ * - e.g. flux mini run my_process.sh
+ *
+ * INITIAL_PROGRAM - process is running as the initial program
+ * of a user flux instance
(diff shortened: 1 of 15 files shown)


src/common/libflux/util.h
```diff
@@ -0,0 +1,39 @@
...
+#endif
+
+typedef enum flux_process_scope {
+    FLUX_PROCESS_SCOPE_NONE = 0,
+    FLUX_PROCESS_SCOPE_SYSTEM_INSTANCE = 1,
+    FLUX_PROCESS_SCOPE_USER_INSTANCE_INITIAL_PROGRAM = 2,
+    FLUX_PROCESS_SCOPE_FLUX_JOB_INSTANCE_INITIAL_PROGRAM = 3,
+    FLUX_PROCESS_SCOPE_FLUX_JOB_INSTANCE_JOB = 4,
+    FLUX_PROCESS_SCOPE_UNKNOWN = 99,
+} flux_process_scope_t;
```
grondo, line 27: Not a full review, just wanted to get this comment in for consideration:

I would start with a pared down list of scopes with less egregiously long names.  :wink: I _think_ this is all our use cases need at this point? Callers that need to differentiate can do so via other attributes, etc.

```suggestion
typedef enum flux_process_scope {
    FLUX_PROCESS_SCOPE_NONE = 0,
    FLUX_PROCESS_SCOPE_SYSTEM = 1,
    FLUX_PROCESS_SCOPE_INITIAL_PROGRAM = 2,
    FLUX_PROCESS_SCOPE_JOB = 3,
    FLUX_PROCESS_SCOPE_UNKNOWN = 99,
} flux_process_scope_t;
```

Also, I didn't look at the code yet, is there really a need for `FLUX_PROCESS_SCOPE_UNKNOWN`? Shouldn't the function just return an error if it cannot determine the scope?

chu11: > Shouldn't the function just return an error if it cannot determine the scope?

Because we don't cover every permutation (per my comment in #3817), I wanted to have some flexibility to allow for future additions, thus the UNKNOWN and why I set that to 99.  We could just return an error instead.  Hmmmm EPERM?  Will have to think about it.

grondo: Aren't the extra permutations just special cases of the already covered scopes as described above?

chu11: yeah, that's true.  given the more limited set, the potential for future extensions is less likely.

chu11:
re-pushed with a hopefully reviewable version of this

- there's `flux_get_process_scope()`.  Again, I stuck it in a new `libflux/util.[ch]`, which I dunno if that's the best place.  In `libjob` would maybe be reasonable too, I waffled back and forth on it.  Could be in `scope.[ch]` too as `util` might be too generic?
- added some tests
- added a `flux scope` command
- added documentation

left to debate, do we need a python utility function for this?  in `util.py` seems wrong.  Maybe that'll determine where it goes in the C code too.

chu11:
per comment in #3817 from @garlick, closing

PR #4705, opened 2022-10-20
Create codeql.yml

trws:
Initial attempt at configuring building and setup for the github version of LGTM scanning. Moving to my own fork.


etc/docker/ubuntu/install_deps.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
```
vsoch, line 1: This is much neater! Can we do bash strict mode so a failure will kill the build?

trws: Oh absolutely, I just wasn't thinking about it.

trws: Done, I'll push it up once we get some results back from the currently enqueued CodeQL run so I can see if anything is still breaking.  Hitting some delays on getting the jobs picked up.

vsoch: Sounds good! My experience with codeQL (I've just used for Python) is that it's fairly slow. I have it on a few scattered projects and it has never hit any issues so I'm undecided about if I like it / if it's worth it - if it's not any slower than the CI here it shouldn't add time, but if it extends the running time we might want to consider just the scheduled run, or similar.

trws: Could be, I'm hoping it will be an adequate replacement for LGTM, since it's literally the same code query system by the same team under the hood.

grondo (approved):
Great! Thanks for taking care of this. I just found one thing I had a question on, otherwise it will be interesting to see how this action "comments" on PRs vs LGTM.
(Since LGTM is going away, though, I suppose that is neither here nor there..)

grondo:
This PR could also probably remove `.lgtm.yml`.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,82 @@
...
+        # Prefix the list here with "+" to use these queries and those in the config file.
+        
+        # Details on CodeQL's query packs refer to : https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#using-queries-in-ql-packs
+        # queries: security-extended,security-and-quality
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        if [ -f requirements.txt ];
+        then pip install -r scripts/requirements-dev.txt;
+        fi
```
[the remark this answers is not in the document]
vsoch: In which context would it not be there?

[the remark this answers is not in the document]
grondo: Oh, that's a good point. The conditional could just be removed as well. (I'm guessing it was there for portability to other flux-framework projects?)

vsoch:
Thank you for your service, LGTM! :vulcan_salute:

grondo:
> Thank you for your service, LGTM!

Oh, I see what you did there!

vsoch:
@trws the errors are because, depending on whether you are building the container vs. in a container, the two different cases will differ in requiring sudo (required in the run here) or not (a traditional container build). So likely we want an envar that comes from the action that controls that.

vsoch:
Yes! Likely that will be the fix since this particular action will always require it. The case I linked needed the dynamic-ness because it could be run in either context as an action (e.g., using it within a user's container or outside of directly on the runner).

trws:
Ok, rebases done.  Assuming I didn't break anything, this should be ready for a clean review.

trws:
I removed the LGTM config file, but the result is not what I expected.  It's still getting run, just without configuration.  Seems like we'll have to deal with that manually and turn off the integration.

PR #4709, opened 2022-10-22
add devcontainer autocomplete

vsoch:
Problem: if we are using the devcontainer to develop Python scripts that use Flux, we cannot take advantage of autocomplete without the Python plugin installed (must be installed in the container) along with having the custom Flux paths included to be seen. This fix will ensure the Python extension is installed, and that all search paths are provided where the module could be (pre and post compile, with preference for installed locations if flux is ultimately built)

Examples (this is newly starting the dev container):

Screenshot from 2022-10-21 18-28-09
Screenshot from 2022-10-21 18-28-23

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -6,10 +6,24 @@
     "customizations": {
       "vscode": {
         "settings": {
-          "terminal.integrated.defaultProfile.linux": "bash"
+          "terminal.integrated.defaultProfile.linux": "bash",
+
+          // Ensure that Python autocomplete works out of the box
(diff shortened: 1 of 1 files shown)


alecbcs:
Is there a way to set the python paths to be a bit more general/use an environment variable so we won't have to update this config with every new version of python?

vsoch:
I think we'd need to do some fu to programatically derive the path, and somehow add as an environment variable here (not sure if that's possible though)? So my current answer, I think, is I don't know.

grondo (approved):
I don't use the devcontainer so can't do a real review, but this seems fine to me!
Feel free to set MWP if you'd like.

PR #4712, opened 2022-10-25
flux-proxy: require owner == proxy user

### LINKED ISSUE #4648: flux proxy to system instance results in sign-type=none error
@ryanday36 reported this:

If you try to use `flux proxy` to connect to the system instance and submit a job, you get an error about the use of `sign-type=none`:
```
$ flux proxy local:///var/run/flux/local flux mini run hostname
flux-mini: ERROR: [Errno 1] only instance owner can use sign-type=none
```
The connector must be returning the wrong owner ID and fooling `flux_job_submit(3)` into thinking it is submitting to a single user instance in the case of flux-proxy?

garlick:
Problem: if proxy user is a guest in the Flux instance, `flux_job_submit()` can get confused about job request signing requirements.

For now just don't allow flux-proxy to connect as a guest to an instance, unless `--force` is specified.

Fixes #4648

### DIFF
--- a/src/cmd/builtin/proxy.c
+++ b/src/cmd/builtin/proxy.c
@@ -281,6 +281,33 @@ static void version_check (flux_t *h, bool force)
     }
 }
 
+static void owner_check (flux_t *h, bool force)
+{
+    const char *owner;
+    unsigned long userid;
+
+    if (!(owner = flux_attr_get (h, "security.owner")))
+        log_err_exit ("flux_attr_get security.owner failed");
+    errno = 0;
+    userid = strtoul (owner, NULL, 10);
+    if (errno != 0)
+        log_msg_exit ("error converting security.owner=%s to integer", owner);
+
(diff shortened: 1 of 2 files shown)


grondo (commented):
This LGTM, but it does seem like a test should/could be added under `t/system` to ensure that proxy to the system instance fails as expected.

garlick:
Oh, silly me, I had lost track of `t/system` and was kind of stumped with how to concoct a test for this.  Perfect, I'll add that.

garlick:
Code coverage is apparently not reflected in the codecov report because it was the system test?    Anyway, I think this is ready for further review.

grondo:
It should be, however it seems like the system tests are not getting folded into the codecov reports today on all our PRs ...:confused:

grondo (approved):
LGTM!

garlick:
Thanks, I'll set MWP.

PR #4713, opened 2022-10-25
flux-resource: fix scalability issues with large sets of resources

grondo:
This PR fixes an embarrassing scaling issue with `flux resource list`. The most severe issue is that `flux-resource`  uses the `ResourceSet.union` method to build up resource sets with matching lines of output. Since the `union` method returns a new resource set on each iteration, this causes an exponential slowdown as the program iterates over all ranks in the returned resource set.

This PR adds an `add` method to librlist and `ResourceSet`, which is like `union` but modifies the target resource set instead of returning a copy. This solves 90% of the issue since NRANKS copy and destroys are now avoided.

The other issue is that lookups by rank in `librlist` is a linear search and this is done frequently in `flux resource` since it iterates over all ranks at least once. To address this issue, a rank hash is added to `librlist` to allow fast lookups of resources by rank. This fixes another 5% of the scaling issue.

Before these changes, it would take almost 2 minutes for `flux resource list` on a cluster with 3000 nodes. After this PR the operation takes <1s.

master:
```console
$ time flux resource list --from-stdin <  t.json 
      free   2988   143424        0 quartz[11-499,501-998,1000-3000]
 allocated     13      624        0 quartz[0-10,500,999]
      down      0        0        0 

real	1m50.031s
user	1m49.967s
sys	0m0.064s
```

This branch:
```console
$ time flux resource list --from-stdin <  t.json 
      free   2988   143424        0 quartz[11-499,501-998,1000-3000]
 allocated     13      624        0 quartz[0-10,500,999]
      down      0        0        0 

real	0m0.854s
user	0m0.792s
sys	0m0.066s
```

### DIFF
--- a/src/bindings/python/flux/resource/ResourceSet.py
+++ b/src/bindings/python/flux/resource/ResourceSet.py
@@ -104,6 +104,16 @@ class ResourceSet:
             self.impl.append(arg.impl)
         return self
 
+    def add(self, *args):
+        """
+        Add resources to a ResourceSet that are not already members
+        """
+        for arg in args:
+            if not isinstance(arg, ResourceSet):
+                arg = ResourceSet(arg, version=self.version)
+            self.impl.add(arg.impl)
+        return self
+
     def copy(self):
         """Return a copy of a ResourceSet"""
         rset = ResourceSet(self.impl.copy())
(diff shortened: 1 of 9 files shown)


garlick:
Great result!

garlick (approved):
LGTM!

grondo:
Thanks! Will set MWP.

Issue #4716, opened 2022-10-25
FLUX_JOB_ID is set in batch scripts


garlick:
Problem: the broker allows its FLUX_JOB_ID to leak through to the initial program, which makes it hard to distinguish whether a program is being run from a batch script or from a job.  It looks like a job, but the job is really in the enclosing instance not the instance that started the initial program.
                               
FLUX_JOB_ID confuses our OpenMPI schizo plugin, preventing singleton MPI jobs from being runnable from the batch script.  See https://github.com/flux-framework/flux-core/issues/4671#issuecomment-1274853242

It also confuses [libyogrt](https://github.com/LLNL/libyogrt) as mentioned in #3817

PR #4717, opened 2022-10-25
unset job environment variables in initial program

garlick:
Problem: as noted in #4716, we shouldn't let an instance's job environment variables leak through to the initial program.

This adds the following variables to the environment variable block list in the code that spawns rc1, rc3, and the initial program:
- FLUX_JOB_SIZE
- FLUX_JOB_NNODES
- FLUX_JOB_TMPDIR
- FLUX_TASK_RANK
- FLUX_TASK_LOCAL_ID
- FLUX_URI (somewhat redundantly since this is set later on to a new value)
- FLUX_KVS_NAMESPACE (also redundantly with code that reads and clears this in the broker but for completeness)


grondo (approved):
Great! I was pleased that no existing tests had to be modified after this change.

PR #4721, opened 2022-10-26
flux-job: support multiple jobids for `cancel`, `raise`, and `kill`

grondo:
I thought I'd throw this implementation up in a PR for comments since it seems to be working.

This PR allows the `flux job cancel`, `flux job raise`, and `flux job kill` subcommands to target multiple jobids per invocation. The ability to set an option message in the free arguments is preserved by stopping the processing of jobids at the first argument that doesn't look like a FLUID, or when the string `--` is encountered. This heuristic could get easily confused if the first word of the message could be interpreted as a jobid, e.g. if it starts with the letter `f` or is a number. The result would be that the command would error on that one jobid, and the message set in the exception note would be missing one or more words. Therefore, it will probably be best practice to use `--` to separate jobids from the message, or to use the new `-m, --message=NOTE` option added here to `flux job raise` and `flux job cancel`.

Given the rarity with which the an exception note is actually used with these commands, this solution and its caveats may be acceptable given the useful outcome.

### DIFF
--- a/doc/man1/flux-job.rst
+++ b/doc/man1/flux-job.rst
@@ -8,15 +8,15 @@ flux-job(1)
 SYNOPSIS
 ========
 
-**flux** **job** **cancel** *id* [*message...*]
+**flux** **job** **cancel** [*OPTIONS*] *ids...* [*--*] [*message...*]
 
 **flux** **job** **cancelall** [*OPTIONS*] [*message...*]
 
-**flux** **job** **kill** [*--signal=SIG*] *id*
+**flux** **job** **kill** [*--signal=SIG*] *id* [*id...*]
 
 **flux** **job** **killall** [*OPTIONS*]
 
-**flux** **job** **raise** [*OPTIONS*] *id* [*message...*]
+**flux** **job** **raise** [*OPTIONS*] *ids...* [*--*] [*message...*]
 
 **flux** **job** **raiseall** [*OPTIONS*] *type* [*message...*]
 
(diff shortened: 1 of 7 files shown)


doc/man1/flux-job.rst
```diff
@@ -30,7 +30,11 @@ flux-job(1) performs various job related housekeeping functions.
 CANCEL
 ======
 
-A single job may be canceled with ``flux job cancel``.
+One or more jobs by may be canceled with ``flux job cancel``.  An optional
+message included with the cancel exception may be provided via the *-m,
+--message=NOTE* option or after the list of jobids. The special argument
+*"--"* forces the end of jobid processing and can be used to separate the
+exception message from the jobids when necessary.
```
chu11, line 37: need to add `-m` description just like with exception.

t/t2201-job-cmd.t
```diff
@@ -229,6 +229,14 @@ test_expect_success 'flux-job: raise fails with invalid jobid' '
...
 '
 
+test_expect_success 'flux-job: raise fails with invalid jobids' '
+	test_must_fail flux job raise foo fee
+'
+
+test_expect_success 'flux-job: raise fails if note set twice' '
+	test_must_fail flux job raise --message=hi ${validjob} -- hello
+'
+
```
chu11, line 239: not necessarily related to this PR, but AFAICT (maybe I'm not seeing it) we don't have any tests that ensure the message is actually posted in the eventlog correctly.  Perhaps should add some while we're at it.

Or perhaps I'm not seeing them, I did a great for "flux job cancel" and didn't see an obvious place they were tested.

chu11 (commented):
overall LGTM, just two small things

garlick (commented):
This approach seems good to me!

I did find something not quite right with the message parsing demonstrated by
```
$ flux mini submit sleep 120
ƒjwULY4F
$ flux job raise ƒjwULY4F -m "test message"
$ flux job attach ƒjwULY4F 
90.545s: job.exception type=cancel severity=0 -m test message
```
(The `-m` got in the message somehow)

grondo:
That's because the `-m` comes after the jobid (by default option parsing stops at the first non-option argument)  in your example above. (the same thing would have happened before) I'm not quite sure how to fix this one.

garlick (approved):
Oh duh, yeah that makes sense. That's on me!

t/t2201-job-cmd.t
grondo: You're right. I'll open an issue on this one.

grondo:
Thanks! I'll set MWP.

Issue #4668, opened 2022-10-08
`flux mini` `--exclusive` flag should have a short option


grondo:
@ryanday36 suggested a short option `-x` for `flux mini --exclusive`. We've tended to err on the side of not adding short options, but this seems like a good suggestion. `-xN` has a nice mnemonic for `exclusive nodes` ...

PR #4726, opened 2022-10-26
flux-mini: add -x short option for --exclusive

grondo:
Not much to see here, this PR just adds a `-x` short option equivalent for `--exclusive` as suggested in #4668


garlick (approved):
I like reviewing the easy ones.  Today I reviewed one PRs!

PR #4735, opened 2022-10-27
testsuite: fix broken proxy system test

garlick:
Problem: there is a typo in the flux-proxy test added to t/system/0001-basic.t.

Fix typo.


grondo:
Ah, failures in the `system` builder are not caught because `make check-code-coverage` always exits with success. I'll try to fix that as well.

grondo (approved):
LGTM, thanks!

PR #4737, opened 2022-10-28
[EXAMPLE] add proto example using Python

vsoch:
Problem: we want to eventually support a more standard rpc interface, and using Captain Proto (cue superhero music) can likely help us. This is not intended to merge in, but is an example of creating a simple client/server model using the Python library for captain proto, and it uses a schema for a FluxMessage to create the message, and then send to the server, and the server creates the flux RPC and returns the result (serialized as a string, which likely we can improve upon.

This PR is intended for discussion - pinging @trws to discuss next steps and @alecbcs to join in on the fun! :unicorn: 

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -2,8 +2,8 @@ FROM fluxrm/testenv:focal
 
 LABEL maintainer="Vanessasaurus <@vsoch>"
 
-# Pip not provided in this version
-RUN apt-get update
+# Captain proto (this can eventually go in testenv container)
+RUN apt-get update && apt-get install -y capnproto
 COPY scripts/requirements-dev.txt /requirements.txt
 
 # For easier Python development. 
--- a/etc/proto/README.md
+++ b/etc/proto/README.md
@@ -1,0 +1,36 @@
+# Captain Proto
+
+> No... Captain Flux! He's a hero! 👩‍🚒️
+
+I am following the [installation](https://capnproto.org/install.html) instructions here.
+The tool should be installed to the devcontainers environment. Then we define our protocol buffer
+file as [schema.capnp](schema.capnp) and use within Python. This is a fairly simple setup
+(to test things out and learn about this library!). First, you'll need flux bindings on your
+path and to start flux:
+
+```consoel
+export PYTHONPATH=/usr/local/lib/python3.8/site-packages
+flux start --test-size=4
+```
+
+Note that the path will be default added in the devcontainers environment soon!
+To start the server portion:
+
+```console
+$ python3 run-server.py
+Starting Flux Message Interface! pkill python3 in another terminal to 🔴
+```
+
+Yes - nothing elegant here - we are using a pkill to python3 to kill the server! 💀 
+That will start the server running on `127.0.0.1:12345`. Then we can test the client with [run-client.py](run-client.py):
+
+```console
+$ python3 run-client.py 
+🍓 Created client <capnp.lib.capnp.TwoPartyClient object at 0x7f071c5e3220>
+{'fee': 'fi', 'route': 'de5d06ad!e058728f!0ae58f31!ee699da9', 'userid': 0, 'rolemask': 1}
+```
+
+What we basically do via the client is prepare the request, send it to the server, and then
+print what is returned. The server handles created a Flux RPC and running "get" for it.
+It's super simple, but kind of neat! I think we can make cooler stuff off of this 
+basic idea.
--- a/etc/proto/run-client.py
+++ b/etc/proto/run-client.py
@@ -1,0 +1,35 @@
+import capnp
+import json
+
+# Load the schema (this method is more reliable)
+capnp.remove_import_hook()
+schema_capnp = capnp.load('schema.capnp')
+
+
+# Create a new message
+# message = schema_capnp.FluxMessage.new_message()
+ 
+
+def main():
+
+    client = capnp.TwoPartyClient('127.0.0.1:12345')
+    print(f"🍓 Created client {client}")
+
+    # Bootstrap the server capability and cast it to the FluxMessageInterface
+    msg = client.bootstrap().cast_as(schema_capnp.FluxMessageInterface)
+    request = msg.get_request()
+    
+    # Prepare the flux message!
+    request.m.topic = "kvs.ping"
+    request.m.payload = json.dumps({"fee": "fi"})
+    promise = request.send()
+
+    # asynchronous
+    # promise.then(lambda ret: print(ret)).wait()
+
+    # synchronous
+    result = promise.wait()
+    print(json.loads(result.x))
+    
+if __name__ == "__main__":
+    main()
--- a/etc/proto/run-server.py
+++ b/etc/proto/run-server.py
@@ -1,0 +1,43 @@
+#!/usr/bin/env python3
+
+import capnp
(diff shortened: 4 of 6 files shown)


garlick:
Is the eventual goal to come up with a capnproto schema for the flux message protocol (unix domain stream socket framing) described in [RFC 3](https://flux-framework--350.org.readthedocs.build/projects/flux-rfc/en/350/spec_3.html)?  Or just something similar that would have a gateway?

Don't forget there are possibly multiple flux instances running on a single node (and sometimes multiple brokers from the same instance).  We worked through all those issues with the unix domain socket that we currently offer (unique paths, security, uri resolvers, etc) so there's some benefit in reusing that socket and not inventing a new protocol.

vsoch:
I think we are mostly just exploring at this point - @trws and @alecbcs and I chat about Captain Proto, and decided the first step was to "give it a try!" This isn't intended to be anything final, but rather a proof of concept, and the simplest route possible. @trws I'm awake and functioning if/when you want to talk more about what we want to try next!

garlick:
Gotcha.  I realized after I spoke that I have no idea if CP lends itself to creating schemas for existing protocols.

It would be neat if it did though, since then you could build "native" flux clients for any language supported by CP, without requiring a protocol gateway or having to think about security or port/path collisions between Flux instances.

vsoch:
> It would be neat if it did though, since then you could build "native" flux clients for any language supported by CP, without requiring a protocol gateway or having to think about security or port/path collisions between Flux instances.

yes yes yes! That!!

trws:
So, if we want to do that we could, but CP would not be the tool. CP is all about its own serialization format, which is actually really good but really not compatible with what we currently do.

[zproto](https://github.com/zeromq/zproto) on the other hand is by the zeromq people,  supports arbitrary binary protocols and I think has native handling for zmq message frames which would save a lot of complexity.  If you're up for it @vsoch, that would be an option to skip the gateway entirely and go directly to everything else.  We would still have the issue of exposing our more internal RPC-based APIs, but it would integrate natively that way.  It also supports c, clojure, c#, f#, python and go natively.  In principle if we did it right, once the protocol is implemented, we could use the generated c from there to *replace* the existing C marshaling code so we could evolve the protocol for all outputs together.  This is actually what I was hoping to do years ago, but never had time for. 😞 

Not sure if that can actually get us around the authentication and encryption issues though, since those still take place in the C library client code don't they @garlick?

trws:
Ok cool, so that might be a really nice way to go.  Now that ssh can forward domain sockets we could leverage that for remote without too much trouble too if it helps.

vsoch:
okay I've chat with @trws and I have an idea for a next step... going to try zproto! I will likely ping y'all with an update / use this branch as our working discussion space for these ideas. If I don't get to it today, definitely will have time over the weekend. HOoooooray Friday!!

garlick:
zproto _appears_ to generate stubs for clients/servers that use 0MQ sockets.  We do pass messages over 0MQ sockets in Flux, but the socket offered to users (e.g. `FLUX_URI=local://...`) isn't one.  Just wanted to be sure you're aware of that caveat up front, in case zproto insists on running the whole show, which it certainly _can_ do.

garlick:
To summarize :coffee: conversation with @trws -
- zproto probably not gonna work because it produces 0MQ messages that we can't use here
- [flatbuffers](https://google.github.io/flatbuffers/) focuses on serialization to/from buffer and so may be a better alternative
- still, it's not going to be able to conform with rfc3

So if we need a new protocol, but still want to use the unix domain socket, there are a couple of different approaches:
- support two packet formats, with version negotiation per client connection.  This way you only have to deal with the server side in flux-core, since the client side would continue to work.
- clone the local connector to a new name and use a new socket path for the client.  No disruption of existing code, but need to translate FLUX_URI to new scheme/path in clients
- just switch everything to the new protocol in a branch and hack hack hack to vicotry

In all approaches the generated C codec would get added to flux-core.

Ultimately we still need the "message parts"  abstraction described in rfc3  if we want to avoid changing higher level stuff in flux.  OTOH to maximize the benefit of a generated codec, you might like to violate that abstraction a bit and have it generate a nice flux message class for you with accessors for things not exposed at that level.  Slightly annoying if we have to change the abstractions in flux-core to get there.

More thought required.   I really like the idea of having a way to do the lowest level message send/receive (at least) natively in other languages.  If only there were some tool that would let us describe the current format abstractly and generate a codec in a variety of languages.   Seems like there's a gap in available tools here.

If any further documentation would help LMK.

trws:
Yeah, I agree about the gap, and thanks again for the chat today by the way.  I thought Kaitai struct would be the thing TBH, it can represent our entire format and is *only* for the actual representation part, no send/recv, but it's also only for *reading* formats right now.  That said, I think we can get really, really close to what's in RFC3 with what FlatBuffers has, it just can't manage bit-for-bit identical.  It can however do packed bitfields, a sized frame so we get a u32 prefix size on each one, and represent optional and defaulted fields of any basic c type, arrays of any of same, validated strings, enums (as bit flags or not), and unions.  We may have to play with it a bit, but I think we could actually cover the whole thing and come out with only order 5 words overhead in message size.

What not exposed things are you thinking of @grondo? You have a much better handle on how the message parts get used in flux than I do, but I could see (for example) leveraging union fields or similar to make a common case easier to tool by expressing a more specific option, encoded string say, or a blob of bytes.

garlick:
We have a nice "class" in libflux for operating on a `flux_msg_t`.  You'd want to generate that as completely as possible in the target language to minimize the amount of manual coding labor neededf or each target.  I was thinking this is at odds with the layering we have in flux-core, where we serialize a message object first to an array of blobs, then hand that off to another abstract layer (which is either zeromq, or the local connector)

But the more I think about it, it won't be possible to generate that entire class since there is logic in there that could not be expressed in the abstract description (like the accessor for a topic string checks the message type and flags before iterating to the right blob offset in the message and decoding it as a string).

It starts to make more sense to me if we talk about two codecs, one for the array of blobs at the connector level and one for the PROTO blob which is part of the libflux `flux_msg_t` class.  That's a good chunk of what you need to build out a new message class.  The other blobs in the blob array are either just strings or raw payloads that are interpreted elsewhere.

trws:
We should be able to literally do that. Have the proto blob be a separate struct or table that we either refer to in the other one, or we don’t refer to it and just mark that blob as containing generic flat buffer data (that doesn’t change the format, just adds a convenience accessor that makes interpreting it easier).


---

On October 28, 2022 at 7:18:36 PM PDT, Jim Garlick ***@***.***> wrote:

We have a nice "class" in libflux for operating on a flux_msg_t. You'd want to generate that as completely as possible in the target language to minimize the amount of manual coding labor neededf or each target. I was thinking this is at odds with the layering we have in flux-core, where we serialize a message object first to an array of blobs, then hand that off to another abstract layer (which is either zeromq, or the local connector)

But the more I think about it, it won't be possible to generate that entire class since there is logic in there that could not be expressed in the abstract description (like the accessor for a topic string checks the message type and flags before iterating to the right blob offset in the message and decoding it as a string).

It starts to make more sense to me if we talk about two codecs, one for the array of blobs at the connector level and one for the PROTO blob which is part of the libflux flux_msg_t class. That's a good chunk of what you need to build out a new message class. The other blobs in the blob array are either just strings or raw payloads that are interpreted elsewhere.

PR #4738, opened 2022-10-28
job-list: ensure purged jobs are inactive

### LINKED ISSUE #4711: job-list: verify purged jobs are inactive
As discussed in #4684, the `job-purge-inactive` event handler does not verify that the jobs passed in are inactive.  That could lead to badness in the job-list module, most notably a assertion bug.

chu11:
Problem: The handler for the job-purge-inactive event does not check if the requested job ids are inactive.  Purging a non-inactive job id can lead to various issues such as stats miscalculations and (worst case) an assertion could be triggered.

Add check to ensure job is inactive before doing purge.

Fixes #4711

### DIFF
--- a/src/modules/job-list/job-list.c
+++ b/src/modules/job-list/job-list.c
@@ -99,6 +99,8 @@ static void purge_cb (flux_t *h,
         struct job *job;
 
         if ((job = zhashx_lookup (ctx->jsctx->index, &id))) {
+            if (job->state != FLUX_JOB_STATE_INACTIVE)
+                continue;
             job_stats_purge (&ctx->jsctx->stats, job);
             if (job->list_handle)
                 zlistx_delete (ctx->jsctx->inactive, job->list_handle);
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -313,6 +313,7 @@ dist_check_SCRIPTS = \
 	issues/t4482-flush-list-corruption.sh \
 	issues/t4583-free-range-test.sh \
 	issues/t4612-eventlog-overwrite-crash.sh \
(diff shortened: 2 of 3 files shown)


grondo (approved):
Looks good! Any chance of a test?

chu11:
> Looks good! Any chance of a test?

I couldn't think of a way to test easily.  We can't use anything through the `job-manager` because that would only send inactive jobs.

Do we have some type of "event" equivalent to `t/request/rpc`?

grondo:
> Do we have some type of "event" equivalent to t/request/rpc?

Probably could be a Python one-liner...

chu11:
oh wait i'm stupid, we have the `flux event ` command :P

chu11:
added a test, wanna take a quick skim @grondo?  I'm not sure how to handle the raciness bit the best.  If there's a bit that's racy and you need to make sure that it didn't crash, you have to spin atleast a little bit to make sure?  Couldn't come up with a better idea.

grondo:
> If there's a bit that's racy and you need to make sure that it didn't crash, you have to spin atleast a little bit to make sure?

Yeah, I don't have a better idea. I'd say it is ok to just run the test and know that we'll detect a crash during shutdown if it occurs (i.e. don't worry about sleeping for a second)?

chu11:
ok, re-pushed just calling `flux job stats` at the end of the test.  Just added a comment about the raciness instead.

chu11:
hmmm the address sanitizer failed "Error: Detected 168 missing missing tests:", i guess something went really wrong with the builder. restarting.

chu11:
doh, forgot to add HAVE_JQ prereq, re-push

PR #4739, opened 2022-10-29
job-list: add successful job count stat

chu11:
built on top of #4684 

we had failed, timeout, and canceled counters, but not "successful" for some reason.  Probably some random "what is our current need" decision from long ago.

### DIFF
--- a/src/bindings/python/flux/job/stats.py
+++ b/src/bindings/python/flux/job/stats.py
@@ -68,14 +68,13 @@ class JobStats:
 
         for state, count in resp["job_states"].items():
             setattr(self, state, count)
-        for state in ["failed", "timeout", "canceled"]:
+        for state in ["successful", "failed", "timeout", "canceled"]:
             setattr(self, state, resp[state])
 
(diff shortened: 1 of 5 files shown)


chu11:
rebased & re-pushed now that #4684 is in, and added 1 extra cleanup commit

src/bindings/python/flux/job/stats.py
```diff
@@ -46,10 +46,10 @@ def __init__(self, handle, queue=None):
             "run",
             "cleanup",
             "inactive",
+            "successful",
             "failed",
             "timeout",
             "canceled",
-            "successful",
```
grondo: Maybe the "reordering" done here should be moved to the first commit in this PR that also does reordering? Alternately, that commit could be moved last so that all the "reordering" is done in a single commit.

chu11: ahhh that's a good idea, instead of having two commits do it

grondo (approved):
Great! LGTM!

PR #4742, opened 2022-10-31
etc: update bash completion functions

grondo:
Some options were changed and removed recently without corresponding updates in our bash completion functions. This PR addresses that issue. It also adds tab completion for named formats with `-o, --format=` with `flux resource` and `flux jobs`.

Additionally, some strange behavior with long options was found with an older bash version. The bash-completion initialization function seems to help with that (along with some `COMP_WORDBREAKS` machinations), but we don't yet require bash-completion so only call that project's `_init_completion()` function when it is available.

### DIFF
--- a/etc/completions/flux.pre
+++ b/etc/completions/flux.pre
@@ -61,6 +61,17 @@ _flux_get_subcmd() {
     echo $firstword
 }
 
+# Handle compgen for words that may contain an '=', so we need to remove
+# any `=` from COMP_WORDBREAKS
+_flux_compgen_with_equals() {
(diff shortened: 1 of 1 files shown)


chu11 (approved):
played with this a bit and works for me.  Just like the original PR, don't understand subtleties of how bash completions work, so review is mostly a skim / sanity check.

grondo:
Thanks! I don't even understand the subtleties of bash completion which is a little worrisome. At some point we may want to add `bash-completion` as a dependency and use more of what that project offers,

PR #4744, opened 2022-10-31
job-list: cleanup error logging, remove excess logging

### LINKED ISSUE #4741: job-list: audit / cleanup logging of zlistx / zhashx errors
In many parts of `job-list` `flux_log_error()` is called after errors in zlistx / zhashx functions.  But those functions may not set errno.  Should (generally speaking) call `flux_log()` instead.

`errno = ENOMEM` may need to be set as well, dependent on the function and if errno setting to caller is necessary.

chu11:
Problem: A number of error logging functions assumed zhashx and zlistx functions set errno and thus flux_log_error() was called to log an error.  However it is not a safe assumption to assume they set errno on error.

Solution: Replace flux_log_error() calls with flux_log() calls after zhashx or zlistx errors.  When necessary, set errno = ENOMOM or remove setting of errno = ENOMEM when it is not necessary to set.

Fixes #4741

### DIFF
--- a/src/modules/job-list/job_state.c
+++ b/src/modules/job-list/job_state.c
@@ -169,9 +169,9 @@ static void revert_job_state (struct job_state_ctx *jsctx,
     }
 }
 
-static void job_insert_list (struct job_state_ctx *jsctx,
-                             struct job *job,
-                             flux_job_state_t newstate)
+static int job_insert_list (struct job_state_ctx *jsctx,
+                            struct job *job,
+                            flux_job_state_t newstate)
 {
     /* Note: comparator is set for running & inactive lists, but the
      * sort calls are not called on zlistx_add_start() */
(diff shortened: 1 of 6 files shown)


garlick (commented):
These errors aren't going to be very helpful if we see them.  Maybe at least tack on "out of memory" if that can be presumed?

It's also worth checking whether logging is appropriate in these internal function if they are capable of returning an error to their caller who may also handle it.  For example,  the first one I checked doesn't appear to need to log anything:  `idsync_check_id_valid()` is only called (indirectly) by `get_job_by_id()` which logs an error message when it fails.   The actual stack is `get_job_by_id ()` -> `check_id_valid()` -> `idsync_check_id_valid()`.

And at least some callers of `get_job_by_id()` also log errors so maybe some additional cleanup is need there.

chu11:
yup, I was admittedly a little lazy with this PR, just trying to fix up only this specific thing.  I'll put in a little more work :-)

chu11:
re-pushed, doing a bit more cleanup than before.  Not sure if I caught every single case, but hopefully got most of them.

src/modules/job-list/job_state.c
```diff
@@ -181,18 +181,27 @@ static void job_insert_list (struct job_state_ctx *jsctx,
         if (!(job->list_handle = zlistx_insert (jsctx->pending,
                                                 job,
                                                 search_direction (job))))
-            flux_log (jsctx->h, LOG_ERR, "%s: zlistx_insert", __FUNCTION__);
+            flux_log (jsctx->h,
+                      LOG_ERR,
+                      "%s: zlistx_insert: out of memory",
```
garlick, line 186: Well...  this function has three possible failures that are all ENOMEM.  The error messages are going to be lke
```
job-list: job_insert_list: zlistx_insert: out of memory
```
Not much context.  Also, distinguishing the three separate ENOMEM cases with unique log messages is also probably unhelpful.

Suggestion: change  `job_insert_list()` to return success/failure and set errno = ENOMEM on failure.  Then handle the logging in the callers.   One caller is `job_change_list()` which could supply more context like:
```
job-list: error moving job to new list on state transition to <state>: <strerror>
```
The other caller is `depthfirst_map_one()`, maybe:
```
job-list: error moving job to new list during restart from kvs:  <strerror>
```

garlick (commented):
Looking better!  I still think the error messages are not providing a lot of context though.

I don't want to hold this PR up forever but it seems like the log messages that are being touched should at least be fixed up a bit if it doesn't trigger too much unrelated refactoring.

src/modules/job-list/job_state.c
chu11: good idea, i did a similar "collapse" in another function.  I'll do it here too.

chu11:
re-pushed.

- regarding the `job_insert_list()` comment above, i added a commit to "goto enomem" along with the prior commit that did the same and have `job_insert_list()` return error on enomem.
  - I only added a log message in `job_change_list()`, b/c a ENOMEM error would have eventually been logged in the `depthfirst_map_one()` path via an error returned from in `job_state_init_from_kvs()`.

- regarding the `idsync_data_destroy()` comment above, I added a commit that preserves errnos in all destroy functions and cleaned up `save_errno` in more places.

garlick (approved):
Thanks @chu11  - cleanup looks good!

chu11:
shall we try to sneak into the release? I can hold off setting MWP for a day.

garlick:
Let's get it in.  I'll go ahead and set MWP.  Thanks!

Issue #4743, opened 2022-10-31
ulimits not propagated 


ryanday36:
We were getting reports of unexpected seg faults while compiling in a Flux instance, and found that the stack size was being limited in the flux instance:
```
[day36@rzwhippet17:~]$ ulimit -s 
unlimited
[day36@rzwhippet17:~]$ flux mini alloc -N1 --exclusive
[day36@rzwhippet2:~]$ ulimit -s
8192
[day36@rzwhippet2:~]$
```
Is there something that we should be configuring to get ulimits propagated into the Flux instance?

grondo:
We don't currently support propagating ulimits, so resource limits are being inherited from the brokers.
Until we can implement this, a more reasonable stack size could be set via `/etc/security/limits.conf` for logins and Flux jobs if we enable PAM config for the IMP (See [flux-config-security-imp(5)](https://flux-framework.readthedocs.io/projects/flux-security/en/latest/man5/flux-config-security-imp.html#flux-config-security-imp-5).

grondo:
Implementation of resource limit propagation could be a little tricky. It would be trvial to add an `atttributes.system.shell.options.rlimits` section to jobspec, and have a simple job shell plugin that would attempt to set any limits defined therein. However, the job shell doesn't run with any elevated privileges, so it would be limited by all current hard limits. Therefore, we'd also have to configure PAM for the IMP as suggested above to increase the hard limits to something reasonable for user jobs.

If that sounds like a reasonable first approach, I could probably propose something soon.

garlick:
What does it mean to propagate ulimits?  Capture the user's limits in the jobspec at submit time, and then in the job
- set initial limits with the IMP PAM config as mentioned above
- have a shell plugin attempt to apply the jobspec limits

I get why that would be desirable for jobs submitted from a batch script because then you could put `ulimit` commands in the batch script, but would it be a bad thing to propagate someone's login node settings when the batch job is submitted (they might set an unlimited core during development, but not want thousands of large core files in the job)?

grondo:
I think this is another "we're used to what Slurm does, so everything else should behave the same" issue. Slurm propagates all current rlimits by default to all jobs, unless the user specifies otherwise via the `--propagate=LIST` option. (At least that is how I remember it working).

ryanday36:
Thanks @grondo and @garlick! I've asked Clay to look at enabling the IMP PAM configs and setting the limits to match what users get by default. I think that'll go a long way toward fixing this.

[...]

PR #4745, opened 2022-11-01
propagate process resource limits to jobs

grondo:
Ok, here's a proposed solution to #4743. This PR adds a new `rlimit` shell plugin and corresponding `rlimit` shell option where process soft rlimits and the values to propagate can be set in jobspec. Then, `flux-mini` commands are all extended to add a set of default rlimits to propagate.

A new `--rlimit=RULE` option is added which works similarly to the existing `--env=RULE` option, i.e. if `RULE` starts with a `-` then that name is explicitly excluded from rlimit propagation, otherwise the name is explicitly included, or with `name=value` an value is set. For example, `--rlimit=-*` propagates nothing, `--rlimit=core=1024` sets `RLIMIT_CORE` to `1024` and `--rlimit=-*,core` propagates only the current core rlimit.

I wasn't sure if the `rlimit` key should be directly in `attributes.system` like the `environment`. For now it is treated as a shell option since setting the rlimits is done by a shell plugin.

### DIFF
--- a/src/shell/rlimit.c
+++ b/src/shell/rlimit.c
@@ -1,0 +1,124 @@
+/************************************************************\
+ * Copyright 2022 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* shell rlimit propagations
+ *
+ * Call setrlimit(2) for any resource limits defined in
+ * attributes.system.shell.options.rlimit
+ */
+#define FLUX_SHELL_PLUGIN_NAME "rlimit"
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+#include <signal.h>
+#include <sys/resource.h>
+#include <jansson.h>
+
+#include <flux/core.h>
+#include <flux/shell.h>
+
+#include "ccan/str/str.h"
+
+#include "internal.h"
+#include "builtins.h"
+
+static int rlimit_name_to_string (const char *name)
+{
+    if (streq (name, "cpu"))
+        return RLIMIT_CPU;
+    if (streq (name, "fsize"))
+        return RLIMIT_FSIZE;
+    if (streq (name, "data"))
+        return RLIMIT_DATA;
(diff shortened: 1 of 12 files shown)


src/shell/rlimit.c
```diff
@@ -0,0 +1,122 @@
...
+                             s ? s : "");
+            free (s);
+            rc = -1;
+            continue;
+        }
+        if (getrlimit (resource, &rlim) < 0)
+            return shell_log_errno ("getrlimit %s", key);
+        rlim.rlim_cur = json_integer_value (value);
+        if (setrlimit (resource, &rlim) < 0)
+            return shell_log_errno ("setrlimit %s", key);
```
garlick, line 110: I guess logging this error and continuing makes sense, b/c a soft value that works on a login node may not be permitted on a compute node.

garlick (approved):
Great!  That came together well.

I think your approach of using shell options for this is fine (I would have been fine going the other way as well), and the command line options for limiting propagation look neatly flexible.

Seems to work great:
```
ƒ(s=1,d=0,builddir) $ ulimit -c
0
ƒ(s=1,d=0,builddir) $ flux mini run bash -c "ulimit -c"
0
ƒ(s=1,d=0,builddir) $ ulimit -c unlimited
ƒ(s=1,d=0,builddir) $ ulimit -c
unlimited
ƒ(s=1,d=0,builddir) $ flux mini run bash -c "ulimit -c"
unlimited
ƒ(s=1,d=0,builddir) $ flux mini run --rlimit=-*,core bash -c "ulimit -c"
unlimited
ƒ(s=1,d=0,builddir) $ flux mini run --rlimit=-* bash -c "ulimit -c"
0
```

src/shell/rlimit.c
grondo: Good point! I'll fix that.

grondo:
I've forced a push here that addresses @garlick's comments, adds tests and documentation, and also makes a change to the `attributes.system.shell.options.rlimit` dictionary to use short names (e.g. `core`, `memlock`) for the dictionary keys instead of the full `RLIMIT_*` names (saves a little space in jobspec).

I'll remove WIP now, and hopefully @ryanday36 can give a review to verify that the solution will meet our needs.

ryanday36:
This looks like a nice solution to me. Thanks @grondo!

garlick (approved):
Checked again after updates and looks good!

grondo:
Thanks!

PR #4747, opened 2022-11-01
testsuite: add flux job raise/cancel/kill tests

### LINKED ISSUE #4736: testsuite: add tests for exception message for flux-job raise/cancel
As noted by @chu11 in #4721 
>  AFAICT (maybe I'm not seeing it) we don't have any tests that ensure the message is actually posted in the eventlog correctly. Perhaps should add some while we're at it.

chu11:
Problem: A number of more basic flux job raise / cancel / kill tests are not covered.  In some cases they may be covered by use in other testsuite areas, but a more basic and clear coverage of usage and options doesn't exist.

Add missing raise, cancel, and kill coverage.

Fixes #4736


chu11:
re-pushed, fixing a && chain error

grondo (approved):
This seems reasonable to me. My only (optional) suggestion would be to combine the submission of all the `flux mini submit sleep 100` jobs into a single `flux mini submit --bcc=` to save a bit of time since back to back mini submits are slow (because, Python). However, this could just be done at a later date to improve the test runtime overall.

chu11:
thanks, setting MWP

PR #4752, opened 2022-11-02
shell: rlimit: improve handling of limits over hard limit

grondo:
Problem: The job shell rlimit plugin issues a possibly confusing message when the requested soft limit exceeds the hard limit. Also, the limit is left alone instead of raising the limit as much as possible, which may be more helpful.

Check to see if the requested soft limit exceeds the hard limit and issue an explicit warning that the shell cannot set the requested value and instead will raise the soft limit to the hard limit.

Update a test in the testsuite with new expected error message.


garlick (approved):
Good improvement!

Issue #4719, opened 2022-10-26
flux jobs completed count drops when a failed job is run


garlick:
Problem: this counting bug was noted during a review of #4684 but it reproduces on current master.
```
 garlick@picl0:~$ flux version
commands:    		0.44.0-113-ge549e5afd
libflux-core:		0.44.0-113-ge549e5afd
libflux-security:	0.8.0-2-g8da4e73
build-options:		+systemd+hwloc==2.4.0+zmq==4.3.4
 garlick@picl0:~$ flux jobs --stats-only
0 running, 86 completed, 14 failed, 0 pending
 garlick@picl0:~$ flux mini run -q debug /bin/false
flux-job: task(s) exited with exit code 1
 garlick@picl0:~$ flux jobs --stats-only
0 running, 85 completed, 15 failed, 0 pending
```
After the failed job, the completed count dropped from 86 to 85.

chu11:
haven't been able to reproduce under simple `flux start` local instances.

@garlick, could you run `flux job stats` before and after the failed job.  So we can atleast isolate if its a `job-list` issue or a `flux-jobs` issue.

garlick:
```console
 garlick@picl0:~$ flux jobs --stats-only
0 running, 83 completed, 17 failed, 0 pending

 garlick@picl0:~$ flux mini run -q batch /bin/false
flux-job: task(s) exited with exit code 1
 garlick@picl0:~$ flux jobs --stats-only
0 running, 82 completed, 18 failed, 0 pending

 garlick@picl0:~$ flux mini run -q batch /bin/false
flux-job: task(s) exited with exit code 1
 garlick@picl0:~$ flux jobs --stats-only
0 running, 81 completed, 19 failed, 0 pending

 garlick@picl0:~$ flux mini run -q batch /bin/true
 garlick@picl0:~$ flux jobs --stats-only
0 running, 81 completed, 19 failed, 0 pending

 garlick@picl0:~$ flux mini run -q batch /bin/true
 garlick@picl0:~$ flux jobs --stats-only
0 running, 81 completed, 19 failed, 0 pending
```

grondo:
Would dumping the stats object directly with `flux job stats` also be helpful here?

garlick:
```
 garlick@picl0:~$ flux job stats
{"job_states":{"depend":0,"priority":0,"sched":0,"run":0,"cleanup":0,"inactive":100,"total":100},"failed":19,"canceled":1,"timeout":8}
[...]

PR #4753, opened 2022-11-02
WIP: job-list: add timestamp of recent purge to job stats                                                                                   

chu11:
Before adding tests, wanted to throw this experiment out for comment.

In #4719 @garlick and I were quite confused about some job stats until it was realized that `inactive-num-limit` was set.  I doubt this is the last time something like this is going to happen.

So that got me thinking monitoring job stats will be very confused if stats change in some unexpected way.

So I added a `t_last_purge` into job-list stats to indicate the time of the last purge.  So people can monitor this if they are concerned.  I think this is a good idea in general for the `job-list` stats RPC.

The part that is perhaps a tad more debatable.  In `flux-jobs`, I output a warning to stderr if there has been a purge in the last hour.

```
>flux jobs --stats-only
Warning: job(s) purged at 2022-11-02T10:34:15, stats affected
0 running, 3 completed, 0 failed, 0 pending
```

could output timestamp everytime on the stats line?  although i dislike this idea.


grondo:
The warning doesn't seem necessary to me. Also the 1hr time period seems a bit arbitrary, though I understand what you're trying to do, but on a system with `inactive-num-limit` and any sort of activity, the warning will always be there.

Maybe it would be more useful to also include purge stats and display this in flux-jobs stats as a reminder that there was a purge which may affect your stats?

chu11:
> Maybe it would be more useful to also include purge stats and display this in flux-jobs stats as a reminder that there was a purge which may affect your stats?

Ahh that's a good idea.  Like:

```
0 running, 3 completed, 0 failed, 0 pending, 1 purged
```

I guess the one negative is on a instance restart, the purge count disappears.  But perhaps that's not something we care about, b/c we're just trying to give the user a warning that some stats may have changed since they last looked.

grondo:
Yeah that's what I was thinking. To avoid confusion maybe use `inactive purged`. I'm not sure if it is best to do a running count of total jobs purged, or those purged in the last hour, etc... :thinking:

chu11:
closing, we'll try the purge count method

PR #4756, opened 2022-11-03
job-list: add inactive purge count to job-list stats

chu11:
Following up #4753 for issue #4719, instead of adding a purge timestamp to the job stats, add a count of jobs purged.  Definitely like this approach better.

pro:
- can do purge counts per queue as well as total
- cleaner, not as much refactoring needed to be done in job-list

con:
- modifies default flux-jobs stats output

but overall, I like this better than the prior attempt that I went ahead and added tests thinking this a worthwhile candidate for merging.

### DIFF
--- a/src/bindings/python/flux/job/stats.py
+++ b/src/bindings/python/flux/job/stats.py
@@ -50,6 +50,7 @@ class JobStats:
             "failed",
             "timeout",
             "canceled",
+            "inactive_purged",
             "pending",
             "running",
             "active",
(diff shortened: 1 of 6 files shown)


chu11:
huh ... why is the CI not starting for this after like 8 hours.  perhaps stuck.  i'll force push something to try and kick it.

chu11:
Thanks, setting MWP

chu11:
rebased and re-pushed, the CI seemed to get stuck or something

Issue #4514, opened 2022-08-23
libjob: deprecate and remove flux_job_list() and flux_job_list_inactive()


grondo:
The C wrapper functions for the `job-list.list` RPC: `flux_job_list()` and `flux_job_list_inactive()` are likely unused outside of test utilities, offer only a subset of functionality of the RPC, and can't be updated or extended without breaking ABI.

We should probably just somehow mark these functions as deprecated and remove them.

chu11:
thought I'd try to deprecate `flux_job_list()` and `flux_job_list_inactive()` really quick.  Updating the function calls with rpcs in `cmd/flux-job` relatively easy and pretty much all tests pass (1 test adjustment to deal with error check no longer handled by `libjob`).  Warnings from tests in `libjob/test` can be dealt with via removing the tests.

Tricky part is how to deal with the warnings coming from the python bindings.  Not sure if we'd want to remove the cffi inclusion of these functions or not, although I'd lean to their removal since people should be using the properly .  Right now the binding scripts gather all appropriate headers, scrape off comments and what not, and put them all together for cffi to do its thing.  

Not sure of the best way to deal with this.  If we had some type of comment field like `NO_FLUX_PYTHON_BINDING`, could sed/awk those functions away.   Or could do some pre-processor trickery, like `#ifdef NO_FLUX_BINDING` and only define when we're doing the cffi magic.

Will leave this for another day.

chu11:
> Not sure of the best way to deal with this. If we had some type of comment field like NO_FLUX_PYTHON_BINDING, could sed/awk those functions away. Or could do some pre-processor trickery, like #ifdef NO_FLUX_BINDING and only define when we're doing the cffi magic (Edit: this is probably preferred, b/c we do a pre-processor step to generate the python bindings anyways).

Was just thinking about this and had a nifty idea on how to do this that might be easier, we could:

A) use some code formatter to make sure all functions are written on one line after the current pre-processing is done.
B) "grep -v deprecated" 

only negative is build would require said code formatting tool to be available.  Dunno how good/bad that is.

chu11:
not sure why we didn't list `flux_job_list_id()` in the above, but I suspect that could be deprecated too.

garlick:
What about adding `__attribute__ ((deprecated))` to the C functions, and then logic in the python binding generation script to not generate bindings for functions with that attribute?

chu11:
[...]

PR #4757, opened 2022-11-04
WIP: job-list: return all attrs by default in job-list-id

chu11:
By default, job-list-id returns no attributes if the user does not specify any.  This is probably not what most users expect.  This is especially true with the python interface, where the average user while likely do:

`job = flux.job.job_list_id(conn, job_id)`

vs.

`job = flux.job.job_list_id(conn, job_id, attrs=["id", "state", "nodelist", ...])`

So I went with job-list-rpc now taking "attrs" as an optional json key and if not-specified, returns all attributes.  The python interface follows accordingly.

I contemplated only having the python interface set `attrs=["all"]` by default, but that felt inconsistent so I did it on the main RPC too.  

Also contemplated having the python interface default to a set of "common" attrs like `attrs=["id", "state", "nodelist", ...]` which would be a reduced list and not have all of the less common attrs (`t_depend`, `annotations`, etc.).  But again, felt inconsistent.

Edit: Oh yeah, and another option is to make `attrs` in the Python `job_list_id()` a required positional input.  Thus a user would be required to set `attrs=["all"]`, but given we're trying to make the API easier, decided against that as well.

So by doing this:

pro:
- users get all data by default, which is probably what they expect

con:
- more data sent back by default (more processing, etc.)
- inconsistent with primary job listing "job-list" rpc.  But we probably do not want "all" by default there, b/c that's probably too much data coming back.  Where as this is specific to just the id requested by the user.

any initial thoughts / comments?  think is good idea?  or perhaps one of the other ideas I list above?

### DIFF
--- a/src/bindings/python/flux/job/list.py
+++ b/src/bindings/python/flux/job/list.py
@@ -89,8 +89,11 @@ class JobListIdRPC(RPC):
 
 # list-id is not like list or list-inactive, it doesn't return an
 # array, so don't use JobListRPC
-def job_list_id(flux_handle, jobid, attrs=[]):
-    payload = {"id": int(jobid), "attrs": attrs}
+def job_list_id(flux_handle, jobid, attrs=None):
+    if attrs:
+        payload = {"id": int(jobid), "attrs": attrs}
+    else:
+        payload = {"id": int(jobid)}
     rpc = JobListIdRPC(flux_handle, "job-list.list-id", payload)
     #  save original JobId argument for error reporting
     rpc.jobid = jobid
--- a/src/cmd/flux-job.c
+++ b/src/cmd/flux-job.c
@@ -1386,7 +1386,8 @@ int cmd_list_ids (optparse_t *p, int argc, char **argv)
     for (i = 0; i < ids_len; i++) {
         flux_jobid_t id = parse_jobid (argv[optindex + i]);
         flux_future_t *f;
-        if (!(f = flux_job_list_id (h, id, "[\"all\"]")))
+        /* by default flux_job_list_id gets all attrs */
+        if (!(f = flux_job_list_id (h, id, NULL)))
             log_err_exit ("flux_job_list_id");
         if (flux_future_then (f, -1, list_id_continuation, NULL) < 0)
             log_err_exit ("flux_future_then");
--- a/src/common/libjob/job.h
+++ b/src/common/libjob/job.h
@@ -168,7 +168,7 @@ flux_future_t *flux_job_list_inactive (flux_t *h,
                                        const char *json_str);
 
 /* Similar to flux_job_list(), but retrieve job info for a single
- * job id */
+ * job id.  `json_str` may be NULL, to retrieve all attributes. */
 flux_future_t *flux_job_list_id (flux_t *h,
                                  flux_jobid_t id,
                                  const char *json_str);
--- a/src/common/libjob/list.c
+++ b/src/common/libjob/list.c
@@ -90,15 +90,21 @@ flux_future_t *flux_job_list_id (flux_t *h,
     json_t *o = NULL;
     int saved_errno;
 
-    if (!h || !json_str
-           || !(o = json_loads (json_str, 0, NULL))) {
+    if (!h || (json_str
+               && !(o = json_loads (json_str, 0, NULL)))) {
         errno = EINVAL;
         return NULL;
     }
-    if (!(f = flux_rpc_pack (h, "job-list.list-id", FLUX_NODEID_ANY, 0,
-                             "{s:I s:O}",
-                             "id", id,
-                             "attrs", o)))
+    if (o)
+        f = flux_rpc_pack (h, "job-list.list-id", FLUX_NODEID_ANY, 0,
+                           "{s:I s:O}",
+                           "id", id,
+                           "attrs", o);
+    else
+        f = flux_rpc_pack (h, "job-list.list-id", FLUX_NODEID_ANY, 0,
+                           "{s:I}",
+                           "id", id);
+    if (!f)
         goto error;
 
     json_decref (o);
--- a/src/common/libjob/test/job.c
+++ b/src/common/libjob/test/job.c
@@ -162,11 +162,6 @@ void check_corner_case (void)
         "flux_job_list_id h=NULL fails with EINVAL");
 
     errno = 0;
-    ok (flux_job_list_id (h, 0, NULL) == NULL
-        && errno == EINVAL,
-        "flux_job_list_id json_str=NULL fails with EINVAL");
-
-    errno = 0;
     ok (flux_job_list_id (h, 0, "wrong") == NULL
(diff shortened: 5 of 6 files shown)


chu11:
> Also contemplated having the python interface default to a set of "common" attrs like attrs=["id", "state", "nodelist", ...] which would be a reduced list and not have all of the less common attrs (t_depend, annotations, etc.). 

Actually, thinking about this a bit, perhaps this isn't a bad idea if we do it for all the job-list retrieval functions (list, list-inactive, list-id).  That way it'd be consistent and users can always expect something even if they don't request specific attributes.

grondo:
I don't think interfaces should return a "reduced" list. I should be either all or none. I think the reduced list is going to be even more confusing because more things will just work but some random JobInfo properties will fail or return incorrect results.

chu11:
> I don't think interfaces should return a "reduced" list. I should be either all or none. I think the reduced list is going to be even more confusing because more things will just work but some random JobInfo properties will fail or return incorrect results.

I guess my concern was that as the number of attributes grows, then we're returning more and more.  But perhaps that's something that can be dealt with in the long future if it gets unwieldy.  Like perhaps a special attribute of "core" could be added to return the "core list" of attributes and that becomes the new default.

chu11:
hmmmm, would it make sense for Python `job_list()` and `job_list_inactive()` to require the `attrs` parameter?  I think the odds low that a user would call one of those functions and not want any attributes other than the job id?  Then there would be some consistency to the C functions / RPCs that require it.  I'm wondering if the initial default setting of `attrs=[]` in the Python API was not a good idea.

grondo:
At first I thought that might break a lot of things, but then you're right, `attrs=[]` is pretty useless, so I wonder if there are any existing callers that don't pass attrs?

garlick:
I wonder if it would just be useful to intrepid explorers to be able to send an RPC with no attribute list and then see what they get?    If all our tools are sending an attribute list, is there any reason to allow this?

chu11:
> At first I thought that might break a lot of things, but then you're right, attrs=[] is pretty useless, so I wonder if there are any existing callers that don't pass attrs?

It's just hard to imagine a case, since it is pretty useless.  If we made it a required positional argument, it could of course break some users.  Thus the debate if we should just set some better default (i.e. my "core values" argument above).  

> I wonder if it would just be useful to intrepid explorers to be able to send an RPC with no attribute list and then see what they get?

Just today @vsoch hit the:

`job_list_id(conn, job_id)`

doesn't return any data question, so I'm beginning to think that it confuses users more than it helps.

vsoch:
> doesn't return any data question, so I'm beginning to think that it confuses users more than it helps.

Can confirm - I autocompleted the functions, saw that one, and then called it (with the handle and job id) and then was confused to get another object back to explore. Adding the attrs with "all" worked beautifully but (as a user) I wouldn't have been able to figure it out. In that APIs should be intuitive, I would err on the side of returning what the user expects (and making the object not the default). If you do keep the object, I would add an obvious to_dict() (akin to_str()) so it's stupidly obvious how to "get all the data." 

I'm not sure about the naming convention, but it even might make sense to provide a courtesy function get_job(jobid) that wraps that call and json.loads() the string representation so the calling user gets back a dictionary of attributes and values (which is what I'd expect is the most common use case!)

grondo:
> but it even might make sense to provide a courtesy function get_job(jobid) that wraps that call and json.loads() the string representation so the calling user gets back a dictionary of attributes and values (which is what I'd expect is the most common use case!)

There's [JobList](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/python/autogenerated/flux.job.list.html#flux.job.list.JobList), which is a little nicer. The default dictionary of raw attributes returned by the `job-list` module is not all that useful on its own, which is why we have the `JobInfo` class. However, maybe we can return a dictionary from an object of that class for when only-a-dict-will-do.

grondo:
> Thus the debate if we should just set some better default (i.e. my "core values" argument above).

I think this is premature optimization. Just return all attributes if the attrs dict is empty. Inevitably someone is going to expect to get some property of job and it won't work even though 90% of the other properties do because one "attribute" is missing. This is only the case with the Python API and so its use will be rare. The `JobList` interface already returns all attributes by default...

chu11:
ok, I just want to gather my thoughts into one place here.

`job = flux.job.job_list_id(conn, job_id)`

results in no data returned except the job id, which is strongly believed to not be what users expect and IMO not good.

Idea 1: Python `job_list_id()` does default `attrs=["all"]`

- pro: least amount of change

- con 1: this is inconsistent with Python `job_list()` and `job_list_inactive()`.

- con 2: this would be inconsistent to the C version of "job_list_id()` which expects a list of attrs.  And b/c its a C function, it can require an input from the user.

Idea 2: Update both Python & C version of `job_list_id()` to assume `attrs=["all"]` if not specified (this is what is currently in this PR).

- pro: isolated to just `job_list_id()`, which is sort of unique / different than `job_list()` and `job_list_inactive()` since it only returns info on one job.

- con: inconsistency to `job_list()` and `job_list_inactve()` in Python & C

Idea 3: defaults `attrs=["all"]` in `job_list()`, `job_list_inactive()` and `job_list_id()` in Python only

- pro: Python API all consistent

- con 1: too much data returned in `job_list()` and `job_list_inactive()`?

- con 2: inconsistency to C api

Idea 4: defaults `attrs=["all"]` Python & C

- pro: is consistent everywhere

- con: too much data returned in `job_list()` and `job_list_inactive()`?

Idea 5: defaults `attrs=["id", "userid", ...]` some "core set"

- pro: solves some concerns of too much data above

- pro: can have good general defaults in both Python & C

- con: users surprised by some missing data they were expecting?

Idea 6: Python - require attrs argument via positional requirement

- pro: consistent to C API

- con 1: break some current users?

- con 2: probably not most user friendly API usage

Idea 7: leave it alone

- pro: we leave it alone!

- con: confirmed confusing users

For me, I'm not that concerned about Python and C APIs being consistent to each other, but it does bother me when `job_list_id()` is inconsistent to `job_list()` (in C or Python land).

I dunno, all have pros and cons.  I liked "Idea 5", but @grondo mentioned above he wasn't a big fan thinking its a net negative.  

Perhaps "Idea 3" above probably the best approach overall?  And we just accept the negative that comes with it?  If it begins to get unwieldy users will have to whittle down to just the attrs they want?

chu11:
> I think this is premature optimization. Just return all attributes if the attrs dict is empty. Inevitably someone is going to expect to get some property of job and it won't work even though 90% of the other properties do because one "attribute" is missing. This is only the case with the Python API and so its use will be rare. The JobList interface already returns all attributes by default...

Oh, didn't see your post as I was gathering my thoughts.  So I guess we're in agreement on "idea 3".

grondo:
> So I guess we're in agreement on "idea 3".

Makes sense to me, especially since we plan to remove the C APIs from libjob (#4514)

PR #4761, opened 2022-11-05
python: add flux.job.get_job

### LINKED ISSUE #4760: python: supply to_dict() function to JobInfo
via @vsoch in #4757 

> Can confirm - I autocompleted the functions [edit: `job_list_id()`], saw that one, and then called it (with the handle and job id) and then was confused to get another object back to explore. Adding the attrs with "all" worked beautifully but (as a user) I wouldn't have been able to figure it out. In that APIs should be intuitive, I would err on the side of returning what the user expects (and making the object not the default). If you do keep the object, I would add an obvious to_dict() (akin to_str()) so it's stupidly obvious how to "get all the data."

vsoch:
Problem: most users of the Python API (or any API in general) ideally want an intuitive user interface. An early likely interaction is to do a flux.job.submit, and then (for many use cases) the next desire is to get more information about the job. As currently implemented, the user needs to stumble on flux.job.job_list_id and decide to try it, and then further to look at the rpc object returned and figure out they can do get_str and json loads that, OR a less intuitive "get" to do the same. OR they could figure out some interaction with a JobInfo object, which to be frank, I do not see how they would know to go there because it is located in a totally separate place, and requires knowing how to instantiate (and is not just calling an easy function). I think the most ideal solution here, from the point of user expectation, is to place an obvious "get_job" directly on the same module the user is already interacting with. That way, it is quickly found and returns the expected data structure without further parsing needed.

- Fixes #4760 

### DIFF
--- a/doc/python/job_submission.rst
+++ b/doc/python/job_submission.rst
@@ -81,6 +81,49 @@ of the asynchronous interfaces.
 
 .. autofunction:: flux.job.wait
 
+.. autofunction:: flux.job.get_job
+
+
+``get_job``
+~~~~~~~~~~~
+
+After submitting a job, if you quickly want to see information for it,
+you can use ``flux.job.get_job``. Here is an example:
+
+.. code:: python
+
+    import flux
+    import flux.job
+
+    # It's encouraged to create a handle to use across commands
+    handle = flux.Flux()
+
+    jobspec = flux.job.JobspecV1.from_command(["sleep", "60"])
+    jobid = flux.job.submit(handle, jobspec)
+    job_meta = flux.job.get_job(handle, jobid)
+
+    {
+        "job": {
+            "id": 676292747853824,
+            "userid": 0,
+            "urgency": 16,
+            "priority": 16,
+            "t_submit": 1667760398.4034982,
+            "t_depend": 1667760398.4034982,
+            "state": "SCHED",
+            "name": "sleep",
+            "ntasks": 1,
+            "ncores": 1,
+            "duration": 0.0
+        }
+    }
+
+If the jobid you are asking for does not exist, ``None`` will be returned.
+For the interested user, this is a courtesy function that wraps using the identifier
+to create an RPC object, serializing that to string, and loading as JSON. 
+Since it is likely you, as the user, will be interacting with ``flux.job``, it
+is also logical you would look for this function to retrieve the job on the same
+module.
 
 ``result`` vs ``wait``
 ~~~~~~~~~~~~~~~~~~~~~~
--- a/src/bindings/python/flux/job/list.py
+++ b/src/bindings/python/flux/job/list.py
@@ -13,7 +13,7 @@ import pwd
 
 import flux.constants
 from flux.future import WaitAllFuture
-from flux.job.info import JobInfo
+from flux.job.info import JobInfo, statetostr
 from flux.rpc import RPC
 
 
@@ -95,6 +95,47 @@ def job_list_id(flux_handle, jobid, attrs=["all"]):
     #  save original JobId argument for error reporting
     rpc.jobid = jobid
     return rpc
+
+
+# get_job is the single variant of job_list_id, and returns the
+# expected data structure (dict) to describe one job (or None)
+def get_job(flux_handle, jobid):
+    """Get job information dictionary based on a jobid
+
+    This is a courtesy, blocking function for users looking for
+    details about a job after submission. The dictionary includes
+    the job identifier, userid that submit it, urgency, priority,
+    t_submit, t_depend, (and others when finished), state, name,
+    ntasks, ncores, duration, nnodes, result, runtime, returncode,
+    waitstatus, nodelist, and exception type, severity, and note.
+    """
+    payload = {"id": int(jobid), "attrs": ["all"]}
+    rpc = JobListIdRPC(flux_handle, "job-list.list-id", payload)
+    try:
+        jobinfo = rpc.get()
+
+    # The job does not exist!
+    except FileNotFoundError:
+        return None
+
+    jobinfo = jobinfo["job"]
+
+    # User friendly string from integer
+    state = jobinfo["state"]
+    jobinfo["state"] = statetostr(state)
+
+    # Get job info to add to result
+    info = rpc.get_jobinfo()
+    jobinfo["nnodes"] = info._nnodes
+    jobinfo["result"] = info.result
+    jobinfo["returncode"] = info.returncode
+    jobinfo["runtime"] = info.runtime
+    jobinfo["priority"] = info._priority
+    jobinfo["waitstatus"] = info._waitstatus
+    jobinfo["nodelist"] = info._nodelist
+    jobinfo["nodelist"] = info._nodelist
+    jobinfo["exception"] = info._exception.__dict__
+    return jobinfo
 
 
 class JobListIdsFuture(WaitAllFuture):
--- a/t/python/t0010-job.py
+++ b/t/python/t0010-job.py
@@ -672,7 +672,41 @@ class TestJob(unittest.TestCase):
         # synchronous job.result() test
         self.assertEqual(job.result(self.fh, ids[3]), result[ids[3]].get_info())
 
+    def test_33_get_job(self):
+        self.sleep_jobspec = JobspecV1.from_command(["sleep", "5"])
+        jobid = job.submit(self.fh, self.sleep_jobspec)
+        meta = job.get_job(self.fh, jobid)
+        self.assertIsInstance(meta, dict)
+        for key in ['id',
+                    'userid',
+                    'urgency',
+                    'priority',
+                    't_submit',
+                    't_depend',
+                    'state',
+                    'name',
+                    'ntasks',
+                    'ncores',
+                    'duration',
+                    'nnodes',
+                    'result',
+                    'runtime',
+                    'returncode',
+                    'waitstatus',
+                    'nodelist',
+                    'exception']:
+            self.assertIn(key, meta)
 
+        self.assertEqual(meta['id'], jobid)
+        self.assertEqual(meta['name'], "sleep")
+        self.assertTrue(meta['state'] in ["SCHED", "DEPEND", "RUN"])
+        self.assertEqual(meta['ntasks'], 1)
+        self.assertEqual(meta['ncores'], 1)
+
+        # Test a job that does not exist
+        meta = job.get_job(self.fh, 123456)
+        self.assertIsNone(meta)
+        
 if __name__ == "__main__":
     from subflux import rerun_under_flux
 
--- a/src/bindings/python/flux/job/__init__.py
+++ b/src/bindings/python/flux/job/__init__.py
@@ -14,7 +14,7 @@ from flux.job.kvs import job_kvs, job_kvs_guest
 from flux.job.kill import kill_async, kill, cancel_async, cancel
 from flux.job.submit import submit_async, submit, submit_get_id
 from flux.job.info import JobInfo, JobInfoFormat, job_fields_to_attrs
-from flux.job.list import job_list, job_list_inactive, job_list_id, JobList
+from flux.job.list import job_list, job_list_inactive, job_list_id, JobList, get_job
 from flux.job.wait import wait_async, wait, wait_get_status, result_async, result
 from flux.job.event import (
     event_watch_async,


vsoch:
okay the underlying error from my test is a FileNotFound error - does anyone know what is going on here with respect to the RPC? 
```console
  FAIL: python/t0010-job.py 37 __main__.TestJob.test_33_get_job
    ---
      message: |
          File "./python/t0010-job.py", line 684, in test_33_get_job
            meta = job.get_job(self.fh, 123456)
          File "/usr/src/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/src/bindings/python/flux/job/list.py", line 108, in get_job
            return rpc.get()
          File "/usr/src/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/src/bindings/python/flux/rpc.py", line 77, in get
            resp_str = self.get_str()
          File "/usr/src/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/src/bindings/python/flux/util.py", line 90, in func_wrapper
            retval = func(future, *args, **kwargs)
          File "/usr/src/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/src/bindings/python/flux/rpc.py", line 71, in get_str
            self.pimpl.flux_rpc_get(payload_str)
          File "/usr/src/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/src/bindings/python/flux/wrapper.py", line 347, in wrap_class
            return new_fun(self_renamed, *args, **kwargs)
          File "/usr/src/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/workdir/src/bindings/python/flux/util.py", line 66, in func_wrapper
            raise EnvironmentError(error.errno, errmsg.decode("utf-8")) from None
        FileNotFoundError: [Errno 2] No such file or directory
    ...
  1..37
```
I'll play around with it locally to see if I can get insights!

vsoch:
ahh derp I figured it out! Fix coming soon.

vsoch:
okay fixed! The bug in the test was the second part of it - asking for a non-existing job (e.g., calling rpc.get()) raises that FileNotFoundError. Since a `get_job` for a job that doesn't exist should (I think) return None, we catch that error and return None instead.

chu11:
as @grondo mentions in #4760, the info returned from the `job-list` module isn't the most user friendly.  Using your example in the documentation:

```
    {
        "job": {
            "id": 5332587773952,
            "userid": 0,
            "urgency": 16,
            "priority": 16,
            "t_submit": 1667674976.7376702,
            "t_depend": 1667674976.7376702,
            "t_run": 1667674976.749891,
            "state": 16,
            "name": "sleep",
            "ntasks": 1,
            "ncores": 1,
            "duration": 0.0,
            "nnodes": 1,
            "ranks": "0",
            "nodelist": "58449938df2e",
            "expiration": 0.0,
            "annotations": {
                "sched": {
                    "resource_summary": "rank0/core0"
                }
            }
        }
    }
```

you'll notice `state` just returns the bitmask value instead of "RUN".

I think that's why in #4760 I suggested to add `to_dict()` in the JobInfo class.  So perhaps it'd be

```
def get_job(flux_handle, jobid):
    payload = {"id": int(jobid), "attrs": ["all"]}
    rpc = JobListIdRPC(flux_handle, "job-list.list-id", payload)
    try:
         return JobInfo(self.get_job()).to_dict()
```        
        
so users will get all of the information but prettied up.  `state` would be "RUN", `state_abbrev` would be "R", `runtime` would be calculated and available for the user (you'll notice its not in the above), etc.

Some minor commit message things:

commit message subject should mention the general area of flux-core, in this case `python:` would suffice.

We try to do commit messages with a "Problem" and "Solution" sections, so try to split up commit message into those chunks.

Be sure to add "Fixes #4760" into your commit message, so the issue will be closed when the PR is merged.

vsoch:
And JobInfo doesn't actually have a to_dict? How would I do that?

I'm trying my best for commits - I am no longer doing "adding" but instead "add" and I'm remembering to use Problem:. And typically I don't put the fixes (number reference) in the commit directly but the PR text. The commit history, imho, should not be tied to GitHub, but we get the automation of closing the issue by mentioning it in the description.

I really am doing my best... I feel like the rules are constantly getting more strict and I'll try my best but nobody has convinced me these extra details are much more than nit picks that push contributors away, to be honest.

vsoch:
okay trying out writing to_dict now - I see how JobInfo works.

vsoch:
@chu11 how are you calling self.get_job from within a module function? What is self?

vsoch:
oh it's the RPC! I should really think about things before posting. Sorry!

vsoch:
okay here are the attributes (after json dumping) on the JobInfo class:

```
{
    "t_depend": 1667694884.4836743,
    "t_run": 1667694884.4956093,
    "t_cleanup": 1667694889.5312598,
    "t_inactive": 1667694889.5334642,
    "duration": 0.0,
    "expiration": 0.0,
    "name": "sleep",
    "queue": "",
    "ntasks": 1,
    "ncores": 1,
    "nnodes": 1,
    "priority": 16,
    "ranks": "0",
    "nodelist": "58449938df2e",
    "success": true,
    "waitstatus": 0,
    "id": 333909749399552,
    "userid": 0,
    "urgency": 16,
    "t_submit": 1667694884.4836743,
    "exception_occurred": false,
    "annotations": {
        "sched": {
            "resource_summary": "rank0/core0"
        }
    },
    "state_id": 64,
    "result_id": 1,
    "exception": {
        "occurred": false,
        "severity": "",
        "type": "",
        "note": ""
    },
    "sched": {
        "resource_summary": "rank0/core0"
    },
    "user": {},
    "dependencies": {}
}
```
I think a lot of the info is the same? The added bits are the exception info and sched, and if there were dependencies it might be that I have to parse those specifically too. Do you really want me to change it for this different implementation?

chu11:
> @chu11 how are you calling self.get_job from within a module function? What is self?

Oh sorry, I was just cutting and pasting for example, but yeah, it's supposed to be the rpc.

chu11:
> okay changed to your spec @chu11 ! I guess I'm wondering - why not just return a job from the submit functions, and then the person can call to_dict() on it?

Good question, maybe that's what we should have done.  Although we could be breaking users at this point so not sure if it would break a bunch of people.

```
{
    "t_depend": 1667694884.4836743,
    "t_run": 1667694884.4956093,
    "t_cleanup": 1667694889.5312598,
    "t_inactive": 1667694889.5334642,
    "duration": 0.0,
    "expiration": 0.0,
    "name": "sleep",
    "queue": "",
    "ntasks": 1,
    "ncores": 1,
    "nnodes": 1,
    "priority": 16,
    "ranks": "0",
    "nodelist": "58449938df2e",
    "success": true,
    "waitstatus": 0,
    "id": 333909749399552,
    "userid": 0,
    "urgency": 16,
    "t_submit": 1667694884.4836743,
    "exception_occurred": false,
    "annotations": {
        "sched": {
            "resource_summary": "rank0/core0"
        }
    },
    "state_id": 64,
    "result_id": 1,
    "exception": {
        "occurred": false,
        "severity": "",
        "type": "",
        "note": ""
    },
    "sched": {
        "resource_summary": "rank0/core0"
    },
    "user": {},
    "dependencies": {}
}
```

hmmm fields like `state` and `result` and "runtime" aren't there.  I guess they haven't been created within the object, they are created on demand as they are requested.  

I'd hate to create some "master list" of fields to be stored in the dict, but an obvious solution is coming to me at the moment.

@grondo created the JobInfo object a long time ago using memoization to not generate data until required / requested.   If we consider this an important use case going forward, maybe its JobInfo that needs to be updated in some way.

vsoch:
@grondo please disperse your wisdom to us!!

vsoch:
I think looking at this again, I like the idea of rethinking the design of the JobInfo, and having it be returned by get_job. And then have a clear to_dict function. That way, we can expand on the idea (JobInfo being the unit to provide a job) without breaking anything (e.g., returning a dict that changes, or returning a dict and then a class).

grondo:
>  I guess I'm wondering - why not just return a job from the submit functions, and then the person can call to_dict() on it?

The reason a jobid and not a "job" is returned from the submit functions is because this is the way Flux works. It is a distributed system and ingesting new jobs is handled by a separate service from getting information about a job. To return a "job" object from job.submit would require a separate RPC to the job-list service after the jobid is returned, which would slow down job submission greatly. Of course, there is nothing preventing a user from combining the two calls into a custom job submission function, it just doesn't make sense to that in the "standard" API.

> created the JobInfo object a long time ago using memoization to not generate data until required / requested. If we consider this an important use case going forward, maybe its JobInfo that needs to be updated in some way.

Yeah, the `JobInfo` class was created specifically by use for tools like `flux-jobs` which are formatting or getting information on many jobs at once. We haven't really had a use case for returning a raw dictionary of values until now, I wonder if a different interface might be required. I thought the original `get_job()` function here was a nice addition for when you want a dict of data for a single job. Maybe that could be augmented to add some of the derived values that are already available in `JobInfo`? Obviously some of the fields provided by `JobInfo` are not of interest in the `to_dict()` use case, i.e. some of the abbreviated names, emoji status, etc.
Therefore, even if we do provide a `to_dict()` method for `JobInfo`, there will likely need to be some cached list of properties to generate anyway...

vsoch:
Ok - so you'd like me to go back to the first implementation? If so - what fields aren't default there that we should add (and where should I get them from?)

grondo:
> Ok - so you'd like me to go back to the first implementation?

I think it is up to you and what your needs are at this point!

> what fields aren't default there that we should add (and where should I get them from?)

I'd say it depends on what you need now. If it were me, I'd just convert the state (and any other) ids to strings (e.g. see `flux.job.info.statetostr`) and start there. I have no idea what the dict is being used for so I'm not the right person to ask what a `get_job()` like interface should return.

vsoch:
Ok I’ll go back to that and add state! But I’d like to hear from @chu11 first because he asked me to not use that implementation.

vsoch:
Nevermind went ahead and restored it back - I like this solution better. I don't think we should be digging into JobInfo.

chu11:
> Ok I’ll go back to that and add state! But I’d like to hear from @chu11 first because he asked me to not use that implementation.

I'd go through the `flux-jobs` manpage and see what fields you feel are important to you and the average user.  I think the emoji ones aren't for example :P

Just skimming the manpage here's the ones I think are "core" important, but feel free to add / subtract if you feel otherwise.

note, some of these are default returned from the job-list module, i'm just listing everything I think is "core" or likely to be important to the average user.

```
**id**
**urgency**
**priority**
**name**
**ntasks**
**ncores**
**nnodes**
**state**
**result**
**waitstatus**
**returncode**
**exception.occurred**
**exception.severity**
**exception.type**
**exception.note**
```

and these are perhaps a bit borderline ...

```
**username**  - if user submitted, likely won't care
**queue** - doesn't care once submitted?
**ranks** - doesn't care once running?
**nodelist** - doesn't care once running?
**success** - will know from result
```

as another note, the `job-list` module defaults to not return data that is not yet relevant, i.e. "result" is not returned until the job is inactive.  Not sure if you'd want to set defaults in the dict you return to the user?

vsoch:
okay I'm at the point where I want to parse my job info into a table - so time to finish this up! Will work on something shortly!

vsoch:
okay here is what I'm testing!

```
{'id': 521419096064,
 'userid': 0,
 'urgency': 16,
 'priority': 16,
 't_submit': 1668125307.640829,
 't_depend': 1668125307.640829,
 'state': 'SCHED',
 'name': 'sleep',
 'ntasks': 1,
 'ncores': 1,
 'duration': 0.0,
 'state_emoji': '📅',
 'result': '',
 'returncode': '',
 'runtime': 0.0,
 'waitstatus': '',
 'nodelist': '',
 'exception': {'occurred': '', 'severity': '', 'type': '', 'note': ''}}
```
And I love the emoji and want it for my table, so I'm pro "include the emoji" in this output!

vsoch:
okay here is the slightly improved interaction. I don't love that I always have to carry around my own handle, but this is much easier than the current!

```python
import flux.job

jobspec = flux.job.JobspecV1.from_command(['sleep', '1200'])
handle = flux.Flux()

job = flux.job.submit(handle, jobspec)

flux.job.get_job(handle, job)
{'id': 21125887164416,
 'userid': 0,
 'urgency': 16,
 'priority': 16,
 't_submit': 1668126535.7681446,
 't_depend': 1668126535.7681446,
 't_run': 1668126535.781823,
 'state': 'RUN',
 'name': 'sleep',
 'ntasks': 1,
 'ncores': 1,
 'duration': 0.0,
 'nnodes': 1,
 'ranks': '0',
 'nodelist': '1b7da4385a56',
 'expiration': 0.0,
 'annotations': {'sched': {'resource_summary': 'rank0/core0'}},
 'state_emoji': '🏃',
 'result': '',
 'returncode': '',
 'runtime': 13.023864269256592,
 'waitstatus': '',
 'exception': {'occurred': '', 'severity': '', 'type': '', 'note': ''}}
```

I think (my preference) for a wrapper would be kind of what I did for the RESTful client, like:

```
cli = FluxJobSubmit()
job = cli.submit(["sleep", 10])
job.info()
```
```console
{'id': 21125887164416,
 'userid': 0,
 'urgency': 16,
 'priority': 16,
 't_submit': 1668126535.7681446,
 't_depend': 1668126535.7681446,
 't_run': 1668126535.781823,
 'state': 'RUN',
 'name': 'sleep',
 'ntasks': 1,
 'ncores': 1,
 'duration': 0.0,
 'nnodes': 1,
 'ranks': '0',
 'nodelist': '1b7da4385a56',
 'expiration': 0.0,
 'annotations': {'sched': {'resource_summary': 'rank0/core0'}},
 'state_emoji': '🏃',
 'result': '',
 'returncode': '',
 'runtime': 31.784000873565674,
 'waitstatus': '',
 'exception': {'occurred': '', 'severity': '', 'type': '', 'note': ''}}
```

chu11:
two comments

1) if `state_emoji` should `result_emoji` be there too?  maybe? maybe not?

2) The original idea was `to_dict()`, which is more obvious about what is being returned.  is `get_job()` too generic?  
i'm struggling to come up with anything better though..  `get_job_info()`? or `get_job_status()`?  The problem is `get_job_info()` or `get_job_status()` could imply other things (job info object or actual job status).  `get_job_data()`??  I dunno.  Just throwing that out there.

src/bindings/python/flux/job/list.py
```diff
@@ -97,6 +97,39 @@ def job_list_id(flux_handle, jobid, attrs=[]):
     return rpc
 
 
+# get_job is the single variant of job_list_id, and returns the
+# expected data structure (dict) to describe one job (or None)
+# This is a courtesy function for the user that looks for it
+# in the flux.job namespace after submitting a job and wanting
+# information about it.
+def get_job(flux_handle, jobid):
```
jameshcorbett, line 105: I think this comment should be formalized and converted into a docstring.

t/python/t0010-job.py
```diff
@@ -672,6 +672,40 @@ def cancel_on_start(future, jobid):
...
+                    'waitstatus',
+                    'nodelist',
+                    'exception']:
+            assert key in meta 
+
+        assert meta['id'] == jobid
+        assert meta['name'] == "sleep"
+        assert meta['state'] in ["SCHED", "DEPEND", "RUN"]
+        assert meta['ntasks'] == 1
+        assert meta['ncores'] == 1
```
jameshcorbett, line 704: We don't use pytest so I think these should be unittest assertions like ``self.assertEqual(meta['ncores'], 1)``

t/python/t0010-job.py
```diff
@@ -672,6 +672,40 @@ def cancel_on_start(future, jobid):
...
+
+        # Test a job that does not exist
+        meta = job.get_job(self.fh, 123456)
+        assert not meta
```
jameshcorbett, line 708: ``is None`` rather than just ``not``?

src/bindings/python/flux/job/list.py
```diff
@@ -97,6 +97,39 @@ def job_list_id(flux_handle, jobid, attrs=[]):
...
+    payload = {"id": int(jobid), "attrs": ["all"]}
+    rpc = JobListIdRPC(flux_handle, "job-list.list-id", payload)
+    try:
+        jobinfo = rpc.get()["job"]
+
```
jameshcorbett, line 110: Better to move everything below this line outside of the ``try`` block?

jameshcorbett (approved):
FWIW I think this looks cool, some minor comments below.

> I think (my preference) for a wrapper would be kind of what I did for the RESTful client, like:
> 
> cli = FluxJobSubmit()
> job = cli.submit(["sleep", 10])
> job.info()

That looks and sounds a lot like the FluxJobExecutor to me.

doc/python/job_submission.rst
```diff
@@ -81,6 +81,46 @@ of the asynchronous interfaces.
 
 .. autofunction:: flux.job.wait
 
+.. autofunction:: flux.job.get_job
+
+
+``get_job``
+~~~~~~~~~~~
+
+After submitting a job, if you quickly want to see stats for it,
```
chu11, line 90: I wouldn't use the word "stats", "data" or "information".

chu11 (commented):
few high level comments:

please add "python: " for the pull request title as well.  That title will be used by a script to generate the release notes.

commit message - empty line between problem and solution

So I see your point about not putting "Fixes ..." in the commit message b/c its github specific.  But, that "Fixes ..." is also a pointer to the original issue and all of its comments and notes, which is useful if one later comes upon the commit and wants to know why it was done.  I'll let others comment if they disagree, but if you are against "Fixes ..." in the commit message, perhaps just a URL to the issue should be done instead.

doc/python/job_submission.rst
```diff
@@ -81,6 +81,46 @@ of the asynchronous interfaces.
...
+you can use ``flux.job.get_job``. Here is an example:
+
+.. code:: python
+
+    import flux
+    import flux.job
+
+    jobspec = flux.job.JobspecV1.from_command(["sleep", "60"])
+    jobid = flux.job.submit(flux.Flux(), jobspec)
+    job_meta = flux.job.get_job(flux.Flux(), jobid)
```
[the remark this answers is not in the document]
vsoch: Sure, I can do that here instead - I definitely agree we want to encourage good practices. When I use this in my restful app, I have one handle created on init that I'm just passing to these.

src/bindings/python/flux/job/list.py
vsoch: yeah that works, I'm indifferent about the two implementations (both have pros and cons).

doc/python/job_submission.rst
```diff
@@ -81,6 +81,49 @@ of the asynchronous interfaces.
...
+After submitting a job, if you quickly want to see information for it,
+you can use ``flux.job.get_job``. Here is an example:
+
+.. code:: python
+
+    import flux
+    import flux.job
+
+    # It's encouraged to create a handle to use across commands
+	handle = flux.Flux()
```
jameshcorbett, line 99: Indentation

src/bindings/python/flux/job/list.py
```diff
@@ -97,6 +97,44 @@ def job_list_id(flux_handle, jobid, attrs=[]):
...
 
 
+# get_job is the single variant of job_list_id, and returns the
+# expected data structure (dict) to describe one job (or None)
+def get_job(flux_handle, jobid):
+    """Get job information dictionary based on a jobid
+
+    This is a courtesy function for the user that looks for it
+    in the flux.job namespace after submitting a job and wanting
+    information about it.
```
jameshcorbett, line 107: I realize this is a lot of stuff to write, but it would probably be good to describe the contents of the dictionary. Especially since the docstring will be pulled into the docs

t/python/t0010-job.py
```diff
@@ -672,6 +672,40 @@ def cancel_on_start(future, jobid):
         # synchronous job.result() test
         self.assertEqual(job.result(self.fh, ids[3]), result[ids[3]].get_info())
 
+    def test_33_get_job(self):
+        self.sleep_jobspec = JobspecV1.from_command(["sleep", "5"])
+        jobid = job.submit(self.fh, self.sleep_jobspec)
+        meta = job.get_job(self.fh, jobid)
+        assert isinstance(meta, dict)
```
jameshcorbett, line 679: `self.assertIsInstance(meta, dict)`

t/python/t0010-job.py
```diff
@@ -672,6 +672,40 @@ def cancel_on_start(future, jobid):
...
+                    'ncores',
+                    'duration',
+                    'nnodes',
+                    'result',
+                    'runtime',
+                    'returncode',
+                    'waitstatus',
+                    'nodelist',
+                    'exception']:
+            assert key in meta
```
jameshcorbett, line 698: `self.assertIn(key, meta)`

t/python/t0010-job.py
```diff
@@ -672,6 +672,40 @@ def cancel_on_start(future, jobid):
...
+
+        self.assertEqual(meta['id'], jobid)
+        self.assertEqual(meta['name'], "sleep")
+        self.assertTrue(meta['state'] in ["SCHED", "DEPEND", "RUN"])
+        self.assertEqual(meta['ntasks'], 1)
+        self.assertEqual(meta['ncores'], 1)
+
+        # Test a job that does not exist
+        meta = job.get_job(self.fh, 123456)
+        self.assertEqual(meta, None)
```
jameshcorbett, line 708: `self.assertIsNone(meta)`

src/bindings/python/flux/job/list.py
```diff
@@ -97,6 +97,44 @@ def job_list_id(flux_handle, jobid, attrs=[]):
     return rpc
 
 
+# get_job is the single variant of job_list_id, and returns the
+# expected data structure (dict) to describe one job (or None)
+def get_job(flux_handle, jobid):
+    """Get job information dictionary based on a jobid
+
+    This is a courtesy function for the user that looks for it
```
jameshcorbett, line 105: Suggestion: "This is a courtesy, blocking function for users looking for details about a job after submission."

src/bindings/python/flux/job/list.py
```diff
@@ -97,6 +97,44 @@ def job_list_id(flux_handle, jobid, attrs=[]):
...
+    information about it.
+    """
+    payload = {"id": int(jobid), "attrs": ["all"]}
+    rpc = JobListIdRPC(flux_handle, "job-list.list-id", payload)
+    try:
+        jobinfo = rpc.get()
+
+    # The job does not exist!
+    except FileNotFoundError:
+        return None
```
jameshcorbett, line 116: Is there any chance that `None` will be returned if this function is called immediately after submission, or is that guaranteed not to happen? Maybe a question for @grondo

chu11: as long as the ID is valid it shouldn't happen.  If the `job-list` module has not yet gotten job data to return to the user, it will wait until it is available.

jameshcorbett:
@vsoch you should feel free to set MWP as far as I'm concerned.

vsoch:
Woot! @grondo how soon will this be available in the fluxrm/flux-sched containers? That's the base I use for the restful API (where I currently implemented the same function but will very happily delete code and use this!)

jameshcorbett:
My guess would be the next release, so the beginning of December?

grondo:
I restarted the flux-sched ci which should result in new containers built
from latest flux-core.

vsoch:
Woot! Thank you! I'll test this out next round of work on the RESTful API. In other news - adding Flux to Nextflow! I can't believe I wrote in a language I didn't know and it kind of worked? lol.

https://github.com/nextflow-io/nextflow/pull/3412

It will probably be a few rounds of tweaks / rebase per Paolo's feedback - I will ping y'all when it looks good from his side for feedback on how I handled the Flux interactions!

PR #4762, opened 2022-11-08
python: by default return all attrs in job listing functions

chu11:
Following up #4757, all of the python list functions return all attributes by default, instead of no entries.  This is the more obvious / expected behavior users should expect.

### DIFF
--- a/src/common/libjob/job.h
+++ b/src/common/libjob/job.h
@@ -138,7 +138,7 @@ int flux_job_wait_get_id (flux_future_t *f, flux_jobid_t *id);
 
 /* Request a list of jobs.
  * If 'max_entries' > 0, fetch at most that many jobs.
- * 'json_str' is an encoded JSON array of attribute strings, e.g.
+ * 'attrs_json_str' is an encoded JSON array of attribute strings, e.g.
  * ["id","userid",...] that will be returned in response.
  *
  * Process the response payload with flux_rpc_get() or flux_rpc_get_unpack().
@@ -155,7 +155,7 @@ int flux_job_wait_get_id (flux_future_t *f, flux_jobid_t *id);
  */
 flux_future_t *flux_job_list (flux_t *h,
                               int max_entries,
-                              const char *json_str,
+                              const char *attrs_json_str,
                               uint32_t userid,
                               int states);
 
@@ -165,13 +165,13 @@ flux_future_t *flux_job_list (flux_t *h,
 flux_future_t *flux_job_list_inactive (flux_t *h,
                                        int max_entries,
                                        double since,
-                                       const char *json_str);
+                                       const char *attrs_json_str);
 
 /* Similar to flux_job_list(), but retrieve job info for a single
  * job id */
 flux_future_t *flux_job_list_id (flux_t *h,
                                  flux_jobid_t id,
-                                 const char *json_str);
+                                 const char *attrs_json_str);
 
 /* Raise an exception for job.
  * Severity is 0-7, with severity=0 causing the job to abort.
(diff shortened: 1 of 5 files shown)


t/python/t0013-job-list.py
```diff
@@ -74,170 +74,239 @@ def waitForConsistency(self, jobs_list_length):
...
+        jobid = self.submitJob(["hostname"])
+
+        self.waitForConsistency(17)
+
+        rpc_handle = flux.job.job_list_id(self.fh, jobid)
+
+        job = rpc_handle.get_job()
+        self.assertEqual(job["id"], jobid)
+        # numbert of attributes should be atleast 10
+        self.assertGreater(len(job), 10)
```
jameshcorbett, line 283: At one point it might be good to test that all expected attributes are there? I'm thinking something like `self.assertLessEqual({"userid", "status", "starttime", "id", ...}, set(job.keys()))`

t/python/t0013-job-list.py
```diff
@@ -74,170 +74,239 @@ def waitForConsistency(self, jobs_list_length):
...
+
+    # flux job list-inactive - alternate attrs requested
+    def test_05_list_inactive_success(self):
+        rpc_handle = flux.job.job_list_inactive(self.fh, time.time() - 3600, 10, attrs=["userid"])
+
+        jobs = self.getJobs(rpc_handle)
+
+        self.assertEqual(len(jobs), 1)
+        # returns id, userid only
+        self.assertEqual(len(jobs[0]), 2)
```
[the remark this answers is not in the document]
jameshcorbett: just to be sure that only the attributes requested are present

jameshcorbett (approved):
Taking it for granted that this is the right approach (as opposed to other approaches discussed in #4757), this LGTM.

src/common/libjob/job.h
```diff
@@ -138,7 +138,7 @@ int flux_job_wait_get_id (flux_future_t *f, flux_jobid_t *id);
 
 /* Request a list of jobs.
  * If 'max_entries' > 0, fetch at most that many jobs.
- * 'json_str' is an encoded JSON array of attribute strings, e.g.
+ * 'attrs_json_str' is an encoded JSON array of attribute strings, e.g.
```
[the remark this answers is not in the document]
chu11: yeah, just a minor cleanup

t/python/t0013-job-list.py
chu11: note to myself, noticed a typo in the above "numbert"

chu11:
re-pushed with fixes from comments above

t/python/t0013-job-list.py
```diff
@@ -74,170 +74,266 @@ def waitForConsistency(self, jobs_list_length):
...
+            "urgency",
+            "priority",
+            "state",
+            "t_submit",
+            "t_run",
+            "t_inactive",
+            "name",
+            "ranks",
+        },
+    )
```
jameshcorbett, line 107: What you have here is a 1-tuple with a single element, a set. I don't think the tests will pass because `self.assertLessEqual(core_attrs, set(job.keys()))` should give you something like `TypeError: '<' not supported between instances of 'tuple' and 'set'` unless I'm missing something

jameshcorbett: If you removed the parentheses and the comma I think it should work. Technically you could just remove the comma but I think better to remove the parentheses too

chu11: oops!  I didn't push my last changes :P  somehow `black` didn't like my formatting and did that.

jameshcorbett:
Not sure if you wanted another review, but as far as I'm concerned, you should feel free to set MWP.

chu11:
thanks!

PR #4770, opened 2022-11-10
content: cleanup and minor tooling updates

garlick:
This splits off some broker content service cleanup from #4546, and also adds support to flux-content(1) for "split content" which will be used in tests to be written for #4546 and shouldn't break any existing uses of the command.


chu11 (approved):
LGTM

garlick:

Issue #4629, opened 2022-09-29
flux queue start|stop does not work with named queues


garlick:
Problem: after #4627, `flux queue disable --queue=NAME` works to prevent job submission on a queue basis.  It would be helpful to also allow `flux queue stop --queue=NAME` to prevent jobs in a particular queue from being scheduled.

One use case for this would be allowing jobs to be submitted to an `all` partition in advance of a DAT.

Edit: this administrative state should be preserved across instance restart like in #4628

chu11:
hacked away on a prototype for the flue queue start/stop with named queues as it took a little time to mentally grasp what was going on.  some refactoring steps that maybe should be done:

- should consider splitting up `alloc-admin` RPC into an `alloc-admin` and `alloc-query` or something.  Logically I think makes things more complicated than probably want.  Would lead to an extra rpc call in `flux-queue` commands I think.

- how to format of the response from `alloc-admin` (or future `alloc-query`)?  I temporarily decided to follow what I did in `job-list` with job stats.  I added a `queues` array into the response that will list enable/disable and reason for each queue, if named queues exist.  I think this generally works, although the "reason" field makes no sense if named queues exist (I just put in reasons like "see queue specifics").

- i copied the `jobq` data structure from `job-manager/queue` into `job-manager/alloc`.  Could be made into a util library or something maybe?  Thinking about it.

- the `job-manager` "alloc queue" could probably use a rename, b/c with "job queue" support, code could get confusing.  Although no alternate name came to mind that would be easier.  maybe "schedqueue" or "schedq"?  Still thinking about it.

chu11:
> I could be missing some details but I was hoping we could just add a started flag to jobq in queue.c. There's already a data structure in there that handles one anon queue vs N named queues and saving/restoring state to the KVS. I would have thought the "admin" rpc for start/stop could be patterned after the enable/disable one that's already there.

Yup, in my prototype I largely patterned it off the enable/disable from the `queue-admin`.   Are you suggesting we move `alloc-admin` into `queue.c`?  I just didn't think to do it that way.

```
    stop by scanning the alloc queue for jobs with that queue name and remove them from the queue
    start by scanning the active job hash for jobs with that queue name that are not in the alloc queue and add them
```

that's pretty much what i've done, except in the stop case we also should cancel pending jobs that have been sent to the scheduler.  We don't want those to be allocated resources, either, correct?

garlick:
[...]

PR #4773, opened 2022-11-18
job-manager: fix memory corruption due to json reference counting bug

chu11:
While working on #4629 found a memory corruption bug that somehow has never caused us any grief :-) (Edit: So far ... )

threw in a whole bunch of other cleanups / minor fixes that I had in my branch too.


garlick (approved):
Good catch on the json_pack refcount!  Other cleanup looks good too.

PR #4775, opened 2022-11-19
 job-manager: use allocation terms consistently

chu11:
Problem: The terms enable/disable and start/stop are used interchangeably when referencing if a scheduler can allocate resources or not.

Consistently use the terms start/stop when referring to scheduler allocations being "allowed" vs "not-allowed".  This includes changes to variables, comments, rpc parameters, tool output, and tests.


garlick (approved):
That was a good cleanup!  Thanks.

PR #4776, opened 2022-11-21
job-manager: support start / stop of individual queues

chu11:
Per discussion in #4629.  Some thoughts:

- For the rpc `job-manager.queue-admin`, I have it so the user should specify either `enable` or `start` but it's an error to specify both.  Ok?  I decided it was better than a `which=<enable|start>` and a  `command=<on|off>` kinda thing.  Or I could have done a `enable=<on|off|keep-same>`.  But didn't like those options.  The resulting code is not to my full liking.  Couldn't think of a better idea.

- To align with old behavior, "stop reason" not required but "disable reason" stays required.  Could make it required?  Left it unrequired mainly b/c it would break tests (i.e. we use `flux queue stop` in a lot of places just to keep jobs pending).

- I preserve backwards compatibility with queue state checkpointing by keeping the enable/disable "reason" field called "reason" instead of "disable_reason".  Remaining code (like variables) stays "reason" too as a result.  Don't have to maintain backwards compatibility?  Or could rename going forward, only have to read "reason" just in case for backwards compatibility?

TODO - need some priority tests, to make sure when queue stopped / restarted - higher proiority jobs run first.


garlick:
Regarding the `queue-admin` RPC: could we simplify things by splitting it up into  `queue-start` and `queue-enable` or similar?  The logic for supporting both in one RPC seems more complicated than it should be on both the client and the server.  IMHO it is better to have a little code duplication if the end result follows familiar patterns that are obvious at a glance.  (Sorry, I created that mess, and now I regret it)

> To align with old behavior, "stop reason" not required but "disable reason" stays required. Could make it required?

I think let's leave it optional and avoid changing the tests.

>  preserve backwards compatibility with queue state checkpointing by keeping the enable/disable "reason" field called "reason" instead of "disable_reason".

Yes, we should be sure that the next release can read the checkpoint info written by the previous one.  I would make all the new fields optional and provide defaults values for when values are missing from the checkpoint that match the previous behavior.  It would be good add a test of this also.

grondo:
> It would be good add a test of this also.

I'm not certain, but is this already tested in CI by virtue of the job-manager restart tests in `t2219-job-manager-restart.t`?

chu11:
> I'm not certain, but is this already tested in CI by virtue of the job-manager restart tests in t2219-job-manager-restart.t?

I think @garlick was referring to a backwards compatibility test.  I assume I would probably hand edit the checkpoint in the dump output or something to see that backwards compatibility worked?

chu11:
> I'm probably not understanding, but a set of valid dump files from previous versions of flux-core is included in the "valid" directory, and additionally a dumpfile is created from the last known tag in CI before running the testsuite. Is the checkpoint not included in the dumpfile? If not then maybe it should be added and we get the testing for free?

Ohhhh I didn't know about the older dump files.

```
./t/job-manager/dumps/valid
./t/job-manager/dumps/valid/dump-0-40-0.tar.bz2
./t/job-manager/dumps/valid/dump-0-41-0.tar.bz2
```

I haven't looked into how these are generated, but I just gotta "hand create" them from earlier builds and stick them in there?

grondo:
One is automatically created in CI for the last known tag:
https://github.com/flux-framework/flux-core/blob/9734f4fcf72f7acc8463c6e8e34e8438be2e47a3/.github/workflows/main.yml#L135-L140

Is that not sufficient?

chu11:
that sounds sufficient then, unless we want to store one permanently for long term backwards compatibility?  Which I don't think is necessary.

garlick:
p.s. I'd be fine if we made the "disable" reason optional to match "stop", if the asymmetry is bothersome.

Another related thought is that we _could_ make the initial state of a named queue disabled and stopped so that we could add them on the fly without there being a race where they might be used before that is intended.   For example, the admins could add a DAT queue to the config and then enable but not start it to allow jobs to be submitted before the DAT.  It would not be good to do that for the anonymous queue though, since that would break batch jobs.  Just a thought to consider.

chu11:
> Another related thought is that we could make the initial state of a named queue disabled and stopped so that we could add them on the fly without there being a race where they might be used before that is intended. For example, the admins could add a DAT queue to the config and then enable but not start it to allow jobs to be submitted before the DAT. It would not be good to do that for the anonymous queue though, since that would break batch jobs. Just a thought to consider.

i'll add that as a "issue to write up" post this PR

chu11:
oh i forgot, we could split this PR into two PRs, one for converting `alloc-admin` to `queue-admin` (or now `queue-enable` / `queue-start`).  Then follow that up with per-queue stop/start.  I think the PR is manageable and not too big, but we could.

chu11:
hmmm concerning a few builders failed with `flux-uptime` constantly segfaulting.

```
 Warning: Found 1 errors from 342 tests in testsuite
Error: not ok 1 - flux-uptime works on rank 0
Error: not ok 2 - flux-uptime works on rank 1
Error: not ok 3 - flux-uptime works as guest
Error: not ok 4 - flux-uptime reports correct size
Error: not ok 6 - flux-uptime reports scheduler stopped
Error: not ok 7 - flux-uptime reports drained node count
Error: ERROR: t2804-uptime-cmd.t - exited with status 1
```

valgrind on TOSS3 system doesn't seem to capture this. hmmm

garlick:
> we could split this PR into two PRs

My vote would be to keep it as one.  As you say it's manageable and we probably don't need to call out changes separately in the release notes.

chu11:
re-pushed and removed WIP

- redid PR to have `queue-enable` and `queue-start` RPC targets
- internally renamed "reason" to "disable_reason" when appropriate
- added more tests, most notably tests to ensure queue stop/start don't messup job priorities and which jobs run next

src/modules/job-manager/alloc.c
```diff
@@ -66,9 +66,11 @@ static void requeue_pending (struct alloc *alloc, struct job *job)
         job->handle = NULL;
```
garlick, line 66: commit message for "job-manager: support per queue start/stop":

The alloc requests are canceled not the jobs so maybe better would be

> Problem: Individual queues cannot be started/stopped.
> 
> Solution: Update the queue-start RPC to allow individual queues to be started/stopped.
When a queue is stopped, alloc requests for jobs in that queue are canceled, and
the jobs are removed from the alloc queue and prevented from re-entering it.

src/modules/job-manager/queue.c
```diff
@@ -311,16 +311,45 @@ int queue_submit_check (struct queue *queue,
...
+{
+    json_error_t error;
+    const char *qname = NULL;
+
+    if (!job || !job->jobspec_redacted) {
+        errno = EINVAL;
+        return -1;
+    }
+
+    if (json_unpack_ex (job->jobspec_redacted, &error, 0,
```
garlick, line 344: It might be more efficient to parse this on job creation into a `job->queue` member.  Then you could follow the existing error handling of the "create"  functions if something is wrong
- if job is created from eventlog, propagate the json parse error (which is ignored here)
- if job is created from json, return generic failure since jobspec was already validated at ingest.

And this parsing step would be done once not possibly multiple times per job.

src/modules/job-manager/queue.c
```diff
@@ -547,12 +576,61 @@ static void queue_enable_cb (flux_t *h,
         flux_log_error (h, "error responding to job-manager.queue-enable");
 }
 
-static void queue_stop_allocations (struct queue *queue) {
-    if (alloc_pending_count (queue->ctx->alloc) > 0) {
+int job_in_queue (struct queue *queue,
+                  struct job *job,
+                  const char *name,
+                  bool *in_queue)
```
garlick, line 602: This name is confusing on so many levels :smile:: `job_` prefix should be reserved for `job.[ch]`, all jobs are "in" a queue although it may be the anonymous one, or are we talking about the "alloc queue", and it sounds like a boolean test but it's not.

Further argument for preparsing `job->queue` and then this function can disappear and callers can just test for NULL and do string compares.

src/modules/job-manager/queue.c
```diff
@@ -547,12 +576,61 @@ static void queue_enable_cb (flux_t *h,
...
+                    return -1;
+            }
+        }
+    next:
+        job = zhashx_next (queue->ctx->active_jobs);
+    }
+    return 0;
+}
+
+static void queue_stop_allocations (struct queue *queue, const char *name) {
```
garlick, line 639: Should this and the previous function just be called `queue_stop()` and `queue_start()`?

Also should be a line break before the opening brace.

garlick (commented):
Nice work.  Just a few comments on first pass.  I'll circle back later and give this a test.

Edit: not sure if I"ll get back to this today so no problem if you end up force pushing to this branch to fix stuff.

src/modules/job-manager/queue.c
```diff
@@ -453,6 +547,47 @@ static void queue_enable_cb (flux_t *h,
...
 
+static void queue_stop_allocations (struct queue *queue) {
+    if (alloc_pending_count (queue->ctx->alloc) > 0) {
+        struct job *job = zhashx_first (queue->ctx->active_jobs);
+        while (job) {
+            if (job->alloc_pending)
+                alloc_cancel_alloc_request (queue->ctx->alloc, job);
+            job = zhashx_next (queue->ctx->active_jobs);
+        }
+    }
```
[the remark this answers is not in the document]
garlick: Yep, derp!

src/modules/job-manager/queue.c
chu11: as an aside to this comment, I wanted to rename the "alloc->queue", b/c the word "queue" was overused, but I couldn't come up with a better name.  Then of course we'd have to adjust the "alloc_enqueue" and "alloc_dequeue" function names too ...

src/modules/job-manager/queue.c
chu11: hmmm there's a `queue_start_all()` function already, which was based on `queue_enable_all()`.  Perhaps `queue_enable_all()` probably should be `jobq_enable_all()`, which follows `jobq_enable()`, then the code for `queue_start` follows after that.

garlick:
Just tried to put this branch on my test cluster (which has two queues configured) and Flux refused to start with:
```
Nov 27 07:10:41 picl0 flux[168061]: job-manager.err[0]: restart: checkpoint.job-manager: Invalid argument
Nov 27 07:10:41 picl0 flux[168061]: job-manager.crit[0]: module exiting abnormally
```

My `checkpoint.job-manager` contains
```json
{
  "max_jobid": 104661775630204930,
  "queue": [
    {
      "name": "batch",
      "enable": true
    },
    {
      "name": "debug",
      "enable": true
    }
  ]
}
```

grondo:
Oh, that is probably my fault! I failed to think through the fact that all the existing dump/reload tests do not have any queues configured whey I suggested that existing tests should be sufficient. :facepalm: Sorry @chu11 and @garlick!

garlick:
> Oh, that is probably my fault! I failed to think through the fact that all the existing dump/reload tests do not have any queues configured when I suggested that existing tests should be sufficient. facepalm Sorry @chu11 and @garlick!

I missed that too!

chu11:
doh, i forgot that

```
        if (json_unpack (entry,                                                                                                             
                         "{s?s s:b s?s s?s s:b s?s}",                                                                                       
                         "name", &name,                                                                                                     
                         "enable", &enable,                                                                                                 
                         "reason", &reason,                                                                                                 
                         "disable_reason", &disable_reason,                                                                                 
                         "start", &start,                                                                                                   
                         "stop_reason", &stop_reason) < 0) {                                                                                
            errno = EINVAL;                                                                                                                 
            return -1;                                                                                                                      
        } 
```

"start" would be optional when restarting with backwards compatibility.

garlick:
Just a thought but it might be a good idea to add a `version` key to the job manager checkpoint object.

Another case to consider (maybe with tests) is that the queue configuration could change while Flux is offline, so invalid queue names in the checkpoint should be ignored, and missing queue names in the checkpoint should get default stop/enable values.

chu11:
> Just a thought but it might be a good idea to add a version key to the job manager checkpoint object.

> Another case to consider (maybe with tests) is that the queue configuration could change while Flux is offline, so invalid queue names in the checkpoint should be ignored, and missing queue names in the checkpoint should get default stop/enable values.

Both good ideas.  Probably should be done as mini pre-requisite PRs to this one.

src/modules/job-manager/queue.c
chu11: can the `jobspec-update` event be used to alter a job's queue?  Right now, it appears the answer is yes, so would have to handle this case.

garlick: No, we don't currently support that.

chu11:
Re-pushed, addressing comments above and from #4876 which was brought into here.  Some "bigger" highlights

- add 'queue' to `job` struct for easier access, this results in some cleaner code elsewhere
- add `alloc_pending_jobs_first()/next()`, which results in some cleaner code elsewhere
- add version number to job-manager checkpoint object
- add additional coverage for older checkpoint dumps w/ and w/o queues, and for future regression tests, add dumps for the new "version 1" dumps

src/modules/job-manager/queue.c
```diff
@@ -270,6 +294,15 @@ int queue_restore_state (struct queue *queue, int version, json_t *o)
...
         }
+        if (version == 1) {
+            if (json_unpack (entry,
+                             "{s:b s?s}",
+                             "start", &start,
+                             "stop_reason", &stop_reason) < 0) {
+                errno = EINVAL;
+                return -1;
+            }
+        }
```
garlick, line 305: Suggestion: don't share code between versions.  Have one unpack for all of version 0, and one unpack for all of version 1.  The fact that we have a version number means that there doesn't have to be any compatibility at all between versions, so assuming that in the code doesn't telegraph the design properly IMHO.

src/modules/job-manager/restart.c
```diff
@@ -220,7 +222,8 @@ int restart_save_state_to_txn (struct job_manager *ctx, flux_kvs_txn_t *txn)
     if (flux_kvs_txn_pack (txn,
                            0,
                            checkpoint_key,
-                           "{s:I s:O}",
+                           "{s:i s:I s:O}",
+                           "version", CHECKPOINT_VERSION,
```
garlick, line 226: This hunk can be deleted and  setting the version key can be introduced with version 1.    The version 0 in the wild does not set the version key so we shouldn't start doing that here.

t/t2219-job-manager-restart.t
```diff
@@ -78,6 +88,22 @@ test_expect_success 'verify that named queue status persists across restart' '
...
+	flux start -o,--config-path=$(pwd)/conf.d \
+	    -o,-Scontent.dump=dump_queue_stopped.tar \
+	    flux queue stop --queue batch xyzzy &&
+	flux start -o,--config-path=$(pwd)/conf.d \
+	    -o,-Scontent.restore=dump_queue_stopped.tar \
+	    flux queue status >dump_queue_stopped.out &&
+	grep "^debug: Scheduling is started" dump_queue_stopped.out &&
+	grep "^batch: Scheduling is stopped: xyzzy" dump_queue_stopped.out
+'
+
```
garlick, line 106: do we also need targeted tests for
- configured queue is missing from checkpoint (enable/stop should get default values)
- checkpointed queue is missing from config (it should be ignored)

garlick (commented):
Looking good!  Just a few minor suggestions.
I'll give this a re-test on my system instance with queues in the morning.

src/modules/job-manager/queue.c
```diff
@@ -430,7 +430,7 @@ static void queue_enable_cb (flux_t *h,
             errno = EINVAL;
             goto error;
         }
-        if (queue_enable_all (queue, enable, disable_reason))
+        if (jobq_enable_all (queue, enable, disable_reason))
```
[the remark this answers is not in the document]
chu11: This is mostly related to your comment in the previous review that `queue_stop_allocations()` and `queue_start_allocations()` should be abbreviated to `queue_start()` and `queue_stop()`.  Which I agreed with.

The functions were named with the suffix `_allocations()` b/c there were other functions `queue_start_X()`, and the reason `queue_start_X()` existed was b/c of functions like `queue_enable_all()`, so the change was to get the function naming pattern down for later functions.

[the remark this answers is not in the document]
garlick: Oh I think I see - thanks!

garlick:
Oh crud, it just dawned on me that Flux always stops the queues on shutdown, so making the start state persistent means that a system instance will always come up all queues stopped, defeating the point of saving the state.

To review:  the use case this PR enables is the ability to have an "all" queue for DATs that can be enabled and stopped before a DAT, then when the scheduled DAT time arrives, other queues can be stopped and the "all" queue started.  And if Flux is restarted in the interim, the current start state of all queues should persist.  This seems like an important use case.

Maybe we could add a `flux queue --nosave` flag or similar for use in the rc script?  Then we'd have to make sure to preserve the last state set without that flag...  Other suggestions welcome.

Sorry I didn't think of this!

garlick:
Another minor point - it seems odd that this prints the status of all queues, as opposed to just the target queue:
```
$ sudo flux queue start -q batch
batch: Scheduling is started
debug: Scheduling is stopped
```
and that `flux queue enable/disable` prints nothing.

A suggestion would be to have enable/disable/start/stop print the same output as status, e.g. both the enable and start status for the target queue, e.g.
```
$ sudo flux queue start -q batch
batch: Job submission is enabled
batch: Scheduling is started
```

chu11:
> Maybe we could add a flux queue --nosave flag or similar for use in the rc script? Then we'd have to make sure to preserve the last state set without that flag... Other suggestions welcome.

Oh, that's a good idea.  I've been setting `FLUX_DISABLE_JOB_CLEANUP=t` all over the place in tests, but that's a much better idea.

src/modules/job-manager/queue.c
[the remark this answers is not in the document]
chu11: doh!  `alloc->pending_jobs` is only used in `limited` mode.  So we have to iterate through all of the `active_jobs`.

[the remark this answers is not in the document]
garlick: oh.... I had forgotten that.  I wonder if there is much downside to keeping the pending_jobs list regardless of scheduling mode?  I agree that iterating the hash is right for this PR though.  Sorry about that.

chu11:
> A suggestion would be to have enable/disable/start/stop print the same output as status, e.g. both the enable and start status for the target queue, e.g.

I'll do this in a follow up PR.  enable/disable don't have the `--quiet` option, and the change in output might break a few tests, so better to separate that one out.

chu11:
> Maybe we could add a flux queue --nosave flag or similar for use in the rc script? Then we'd have to make sure to preserve the last state set without that flag... Other suggestions welcome.

Argh!  This has been posing a problem.  Implementation was simple enough but took me awhile to realize we always checkpoint when the job-manager is shutdown.

```
    if (flux_reactor_run (r, 0) < 0) {
        goto done;
    }
    if (restart_save_state (&ctx) < 0) {
        flux_log_error (h, "error saving job manager state to KVS");
        goto done;
    }
```

so the option sort of does nothing.

I experimented with this idea:

```
diff --git a/src/modules/job-manager/job-manager.c b/src/modules/job-manager/job-manager.c
index e20415908..3a5bb5a6c 100644
--- a/src/modules/job-manager/job-manager.c
+++ b/src/modules/job-manager/job-manager.c
@@ -226,12 +226,12 @@ int mod_main (flux_t *h, int argc, char **argv)
     }
     if (restart_from_kvs (&ctx) < 0) // logs its own error messages
         goto done;
-    if (flux_reactor_run (r, 0) < 0) {
-        flux_log_error (h, "flux_reactor_run");
+    if (restart_save_state (&ctx) < 0) {
+        flux_log_error (h, "error saving initial job manager state to KVS");
         goto done;
     }
-    if (restart_save_state (&ctx) < 0) {
-        flux_log_error (h, "error saving job manager state to KVS");
+    if (flux_reactor_run (r, 0) < 0) {
+        flux_log_error (h, "flux_reactor_run");
         goto done;
     }
     rc = 0;
```

i.e. only checkpoint before `flux_reactor_run()` to ensure one checkpoint exists, and then all checkpointing afterwards is responsibility of things that care about checkpoints to checkpoint when a change occurs.  Works except for:

```
FAIL: t2219-job-manager-restart.t 9 - and max_jobid is still greater than zero
```

One possible solution would be to have multiple checkpoint objects in the `job-manager`, so not everything is checkpointed on exit, i.e. `checkpoint.job-manager.max_jobid` and `checkpoint.job-manager.queues` or something.

Another solution would be to cache some type of "checkpoint object" in the `job-manager` context, so that we checkpoint that object and not checkpoint the "live data".  So we can choose to update the "checkpoint object" when we want and that is the data that the job-manager will checkpoint when it exits.

Will have to mull over this some more, maybe another option will come to me.

garlick:
I thought that might be annoying.  I think the checkpoint should only be written when the job manager is unloaded and read when it is loaded.  I would guess we would need a shadow value for started that reflects the persistent state.

chu11:
> I thought that might be annoying. I think the checkpoint should only be written when the job manager is unloaded and read when it is loaded. I would guess we would need a shadow value for started that reflects the persistent state.

Yeah, that was my thinking with the cached "checkpoint object", although a shadow value might be easier.

chu11:
re-pushed, with lots of fixes given above comments / discussion.  Besides the code cleanup, the big change is how to solve the `flux queue stop` during shutdown problem.  Added a `nocheckpoint` flag to the `queue-start` RPC and a `--nocheckpoint` option to `flux queue stop`.  Implemented a "shadow value" to store the checkpointing value per discussion above.   It ended up being way simpler than I thought it would be.

chu11:
uh oh, it appears some `flux-sched` multiqueue tests fail b/c of how the command changes (i assume, still need to investigate).  

Perhaps should remove tests in `flux-sched`, then add them back later?

chu11:
rebased & re-pushed, the changes to support consistent output between `flux queue enable/disable` and `flux queue start/stop` wasn't as bad as I thought, so I decided to stick it on top of this PR.  Hopefully this PR isn't getting too weighty, we're at 28 commits now.

garlick:
Just poking at this.  I noticed that the `checkpoint.job-manager` gets updated every time I start/stop/enable/disable a queue.  It should only be updated when the job manager is unloaded like the `max_jobid` value.

Those states do seem to persist across instance restart!

chu11:
The checkpoint after every queue change to enable/disable (and followed with start/stop) was added via:

```
commit 1fc00e95d3f15e2a74b073892ce81da1cf15c36b
Author: Jim Garlick <garlick.jim@gmail.com>
Date:   Mon Oct 3 06:31:10 2022 -0700

    job-manager: save/restore queue state in KVS
    
    Problem: queue enable/disable state is not persistent across an
    instance restart, which may surprise system administrators.
    
    Add queue state to the job manager KVS checkpoint that is saved/restored
    at module unload/load.  Also, save a checkpint after a successful
    queue-admin RPC to capture any changes immediately.
    
    Fixes #4628
```

It appears it was intentionally added for a reason, do you recall why?  Or maybe was a mistake?  Maybe should punt to new PR if you think should change?

garlick (approved):
Looks good @chu11!

chu11:
awesome thanks!

chu11:
re-pushed, supporting backwards compatibility where `flux queue start/stop` does `--all` implicitly if neither `--queue` or `--all` is specified.    The entirety of the differences compared to before is below (note the first few lines are a cleanup possibility I found where I could move some vars inside a block., it does nothing).

Perhaps biggest debate will be is if the warning message text is good.

```
diff --git a/src/cmd/flux-queue.c b/src/cmd/flux-queue.c
index c98e0524f..aeddbb49b 100644
--- a/src/cmd/flux-queue.c
+++ b/src/cmd/flux-queue.c
@@ -387,8 +387,6 @@ static void queue_status (flux_t *h,
 {
     if (!name || verbose) {
         json_t *queues;
-        size_t index;
-        json_t *value;
         flux_future_t *f;
 
         f = flux_rpc (h, "job-manager.queue-list", NULL, 0, 0);
@@ -397,6 +395,8 @@ static void queue_status (flux_t *h,
                                        "queues", &queues))
             log_msg_exit ("%s", future_strerror (f, errno));
         if (json_array_size (queues) > 0) {
+            size_t index;
+            json_t *value;
             json_array_foreach (queues, index, value) {
                 queue_status_one (h, json_string_value (value), out_cb);
             }
@@ -488,6 +488,28 @@ static void print_start_status (const char *name,
             stop_reason ? stop_reason : "");
 }
 
+void check_legacy_all (flux_t *h, const char *name, bool *allp, bool quiet)
+{
+    /* to ensure backwards compatibility, we assume --all if neither
+     * --name or --all are specified.  However, only output warning
+     * if multiple queues are configured.
+     */
+    if (!name && !(*allp)) {
+        json_t *queues;
+        flux_future_t *f;
+
+        if (!quiet
+            && (f = flux_rpc (h, "job-manager.queue-list", NULL, 0, 0))
+            && flux_rpc_get_unpack (f,
+                                    "{s:o}",
+                                    "queues", &queues) == 0
+            && json_array_size (queues) > 0)
+            fprintf (stderr,
+                     "warning: --queue/--all not specified, assuming --all\n");
+        (*allp) = true;
+    }
+}
+
 int cmd_start (optparse_t *p, int argc, char **argv)
 {
     flux_t *h;
@@ -501,6 +523,7 @@ int cmd_start (optparse_t *p, int argc, char **argv)
     }
         log_err_exit ("flux_open");
+    check_legacy_all (h, name, &all, optparse_hasopt (p, "quiet"));
     queue_start (h, name, true, NULL, all, false);
     if (!optparse_hasopt (p, "quiet")) {
         queue_status (h,
@@ -527,6 +550,7 @@ int cmd_stop (optparse_t *p, int argc, char **argv)
         reason = parse_arg_message (argv + optindex, "reason");
         log_err_exit ("flux_open");
+    check_legacy_all (h, name, &all, optparse_hasopt (p, "quiet"));
     queue_start (h, name, false, reason, all, nocheckpoint);
     if (!optparse_hasopt (p, "quiet")) {
         queue_status (h,
diff --git a/t/t2240-queue-cmd.t b/t/t2240-queue-cmd.t
index 4b08c728b..2b8e37e89 100755
--- a/t/t2240-queue-cmd.t
+++ b/t/t2240-queue-cmd.t
@@ -439,15 +439,19 @@ test_expect_success 'flux-queue enable can do one queue' '
 '
 
 
-test_expect_success 'flux-queue stop without --queue or --all fails' '
-       test_must_fail flux queue stop test reasons
-'
 test_expect_success 'flux-queue stop --all affects all queues' '
        flux queue stop --all test reasons > mqstop.out &&
        test $(grep -c "Scheduling is stopped: test reasons" mqstop.out) -eq 2 &&
        flux queue status >mqstatus_stop.out &&
        test $(grep -c "Scheduling is stopped: test reasons" mqstatus_stop.out) -eq 2
 '
+test_expect_success 'flux-queue stop w/o --all affects all queues but outputs warning' '
+       flux queue start --all &&
+       flux queue stop test reasons 2>mqstatus_stop2.err &&
+       grep "warning" mqstatus_stop2.err &&
+       flux queue status >mqstatus_stop2.out &&
+       test $(grep -c "Scheduling is stopped: test reasons" mqstatus_stop2.out) -eq 2
+'
 test_expect_success 'jobs may be submitted to either queue' '
        flux mini submit --wait-event=priority -q batch /bin/true > job_batch1.id &&
        flux mini submit --wait-event=priority -q debug /bin/true > job_debug1.id
@@ -476,9 +480,6 @@ test_expect_success 'submitted jobs are not running' '
        wait_state $(cat job_debug1.id) SCHED &&
        flux jobs -n -o "{state}" $(cat job_debug1.id) | grep SCHED
 '
-test_expect_success 'flux-queue start without --queue or --all fails' '
-       test_must_fail flux queue start
-'
 test_expect_success 'flux-queue start --all affects all queues' '
        flux queue start -a > mqstart.out &&
        test $(grep -c "Scheduling is started" mqstatus.out) -eq 2 &&
@@ -491,6 +492,14 @@ test_expect_success 'submitted jobs ran and completed' '
        wait_state $(cat job_debug1.id) INACTIVE &&
        flux jobs -n -o "{state}" $(cat job_debug1.id) | grep INACTIVE
 '
+test_expect_success 'flux-queue start w/o --all affects all queues but outputs warning' '
+       flux queue stop --all &&
+       flux queue start 2>mqstatus_start2.err &&
+       grep "warning" mqstatus_start2.err &&
+       flux queue status >mqstatus_start2.out &&
+       test $(grep -c "Scheduling is started" mqstatus_start2.out) -eq 2
+'
+
 test_expect_success 'flux-queue stop can do one queue' '
        flux queue stop -q batch nobatch > mqstop_batch.out &&
        test $(grep -c "Scheduling is" mqstop_batch.out) -eq 1 &&
```

garlick:
Nicely done, keeping the diff to a reviewable minimum!  If the sched test passes then I'm for merging (already approved).

chu11:
rebased, guess the queue is busy, no need to have this go through CI twice :P

chu11:
woo-hoo, flux-sched passed, setting MWP, restarted the coverage builder that timed out

chu11:
restarted the flux-accounting builder, seemed to be stuck, was running for about 2 hours.

chu11:
crud, this is the second time that `flux-accounting`'s builder has hung, something must be up ... there are some calls to `flux queue stop` in `flux-accounting`'s tests.

chu11:
yup, got my docker built with the right tree and it does seem to be `t1013-mf-priority-queues.t`, i think it's hanging in the `flux queue idle` in rc1's cleanup path.  Don't know why right now.  Atleast 1 test fails in that file too, not sure if that's the reason why.

chu11:
ok, i think I figured it out, I'll have to send a PR to `flux-accounting`

```
test_expect_success 'stop the queue' '                                                                                                      
        flux queue stop                                                                                                                     
'

<snip>

# Include "foo" queue that accounting doesn't know about for test below                                                                     
test_expect_success 'configure flux with those queues' '                                                                                    
        cat >conf.d/queues.toml <<-EOT &&                                                                                                   
        [queues.standby]                                                                                                                    
        [queues.expedite]                                                                                                                   
        [queues.bronze]                                                                                                                     
        [queues.silver]                                                                                                                     
        [queues.gold]                                                                                                                       
        [queues.foo]                                                                                                                        
        EOT                                                                                                                                 
        flux config reload                                                                                                                  
'
```

before this PR, the start/stop of queues was global, so all of the newly configured queues were stopped by default b/c of a prior `flux queue stop`.  But now that queues are individually started/stopped, all of the queues that were just configured were automaticaly started.  So jobs were being run, which also affected a test later on in this file.  so i think we need to do

```
diff --git a/t/t1013-mf-priority-queues.t b/t/t1013-mf-priority-queues.t
index 39f38d0..adbcdce 100755
--- a/t/t1013-mf-priority-queues.t
+++ b/t/t1013-mf-priority-queues.t
@@ -84,6 +84,7 @@ test_expect_success 'configure flux with those queues' '
        [queues.foo]
        EOT
        flux config reload
+       flux queue stop
 '
```

and some other cleanup should be done as well.

chu11:
kicked the flux-accounting builder, hopefully will no longer hang after https://github.com/flux-framework/flux-accounting/pull/302

chu11:
flux-accounting passes now!  but seems mergifyio needs to be kicked (probably timeout from previous builder hang)

chu11:
@mergifyio rebase

PR #4781, opened 2022-11-23
broker: fix startup failure when broker.mapping is not set

grondo:
This PR fixes an issue introduced by #4615: When the `broker.mapping` attribute isn't set, the broker can't calculate the instance critical ranks and this is treated as a fatal error when it should not be. This can occur when `-o pmi.clique=none` is used or if `PMI_process_mapping` overflows the PMI max value length.

This PR simply makes an error sending the critical ranks to the upstream job-exec service a non-fatal error.


chu11 (approved):
LGTM!

grondo:
Thanks! I set MWP.

Issue #4466, opened 2022-08-04
need option for restarting from kvs dump for debug


garlick:
Problem: it's currently possible to run `flux kvs dump` on a running instance to obtain a dump of kvs content, and load that in a test instance for offline debug, but rc1 has to be hand edited in the test instance to avoid loading modules like `resource` that will cause the instance to fail to start:
```
$ flux start -o,-Scontent.restore=/g/g0/garlick/kvs.tgz
[garlick@fluke1:flux-core]$ src/cmd/flux start -o,-Scontent.restore=kvs.tgz
2022-08-04T16:04:45.543257Z resource.err[0]: problem replaying eventlog drain state: Invalid argument
2022-08-04T16:04:45.543268Z resource.crit[0]: module exiting abnormally
2022-08-04T16:04:45.543841Z broker.err[0]: rc1.0: flux-module: broker.insmod: Invalid argument
2022-08-04T16:04:45.544863Z broker.err[0]: rc1.0: /g/g0/garlick/proj/flux-core/etc/rc1 Exited (rc=1) 2.1s
[garlick@fluke1:flux-core]$ 
```

garlick:
Current procedure is apply this diff to rc1 (assuming you're running the test instance from the source tree)
```diff
diff --git a/etc/rc1 b/etc/rc1
index dfb6b615d..72c5351e0 100755
--- a/etc/rc1
+++ b/etc/rc1
@@ -48,11 +48,12 @@ if test $RANK -eq 0; then
     flux startlog --post-start-event
 fi
 
-modload all resource
+#modload all resource
 modload 0 cron sync=heartbeat.pulse
 modload 0 job-manager
 modload all job-info
 modload 0 job-list
+exit 0
 period=`flux config get --default= archive.period`
 if test $RANK -eq 0 -a -n "${period}"; then
     flux module load job-archive
```
Then start as above.  FWIW, the script I was using to start the test instance under valgrind to chase #4465 is
```sh
#!/bin/bash
src/cmd/flux start \
        --wrap=libtool,e,valgrind \
        --wrap=--log-file=valgrind.out \
        --wrap=--tool=memcheck \
        --wrap=--leak-check=full \
        --wrap=--gen-suppressions=all \
        --wrap=--trace-children=no \
        --wrap=--child-silent-after-fork=yes \
        --wrap=--num-callers=30 \
        --wrap=--leak-resolution=med \
        --wrap=--error-exitcode=1 \
[...]

PR #4783, opened 2022-11-28
add flux start --recovery

garlick:
Problem: when a system instance won't start, debugging is a chore.

This adds a `flux start --recovery` mode that _should_ let rank 0 of a system instance be started interactively when something goes wrong with systemd startup, e.g.
```console
$ sudo -u flux ./flux start --recovery
flux@picl0:/nfshome/garlick/proj/flux-core/src/cmd$ 
```
It only works for the system instance right now, since it hardwires the system instance configuration path and doesn't work with PMI bootstrap.

Thought I'd post it to see what people thought before digging in deeper.  It seems like it's happened several times that we try to start Flux after a configuration update or new release, and the first problem we encounter is that the instance won't start.  If there's something in the KVS (like the `checkpoint.job-manager` problem I noted in #4776) then in theory it could be manually corrected using this mode.

### DIFF
--- a/src/cmd/flux-start.c
+++ b/src/cmd/flux-start.c
@@ -27,6 +27,7 @@
 #include <flux/core.h>
 #include <flux/optparse.h>
 
+#include "ccan/str/str.h"
 #include "src/common/libczmqcontainers/czmq_containers.h"
 #include "src/common/libutil/xzmalloc.h"
 #include "src/common/libutil/log.h"
@@ -87,8 +88,16 @@ static void setup_profiling_env (void);
 static void client_wait_respond (struct client *cli);
 static void client_run_respond (struct client *cli, int errnum);
 
+const char *default_config_path = X_SYSCONFDIR "/flux/system/conf.d";
+const char *default_statedir = "/var/lib/flux";
+
 const char *usage_msg = "[OPTIONS] command ...";
 static struct optparse_option opts[] = {
+    { .name = "recovery",   .key = 'r', .has_arg = 2, .arginfo = "[TARGET]",
+      .flags = OPTPARSE_OPT_SHORTOPT_OPTIONAL_ARG,
+      .usage = "Start instance in recovery mode with dump file or statedir", },
+    { .name = "sysconfig",  .has_arg = 0,
+      .usage = "Load system configuration", },
     { .name = "verbose",    .key = 'v', .has_arg = 2, .arginfo = "[LEVEL]",
       .usage = "Be annoyingly informative by degrees", },
     { .name = "noexec",     .key = 'X', .has_arg = 0,
@@ -430,6 +439,18 @@ void add_args_list (char **argz, size_t *argz_len, optparse_t *opt, const char *
             log_err_exit ("argz_add");
 }
 
+void add_argzf (char **argz, size_t *argz_len, const char *fmt, ...)
+{
+    va_list ap;
+    char arg[1024];
+
+    va_start (ap, fmt);
+    (void)vsnprintf (arg, sizeof (arg), fmt, ap);
+    va_end (ap);
+    if (argz_add (argz, argz_len, arg) != 0)
+        log_err_exit ("argz_add");
+}
+
 char *create_rundir (void)
 {
     char *tmpdir = getenv ("TMPDIR");
@@ -482,6 +503,19 @@ int execvp_argz (char *argz, size_t argz_len)
     return -1;
 }
 
+bool system_instance_is_running (void)
+{
+    flux_t *h;
+    bool running = false;
+
+    unsetenv ("FLUX_URI");
+    if ((h = flux_open (NULL, 0))) {
+        running = true;
+        flux_close (h);
+    }
+    return running;
+}
+
 /* Directly exec() a single flux broker.  It is assumed that we
  * are running in an environment with an external PMI service, and the
  * broker will figure out how to bootstrap without any further aid from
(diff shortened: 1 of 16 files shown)


grondo:
This seems useful!

If possible, a message printed just before dropping to the prompt with some terse instructions might be helpful. (I'm thinking like the recovery/single user mode shell in linux)

garlick:
Here's what I came up with for a message.  I reworked a couple of bits including reducing the stderr log level to "warn".
```console
$ sudo -u flux flux start -r
2022-11-28T19:53:44.970486Z job-manager.err[0]: restart: checkpoint.job-manager: Invalid argument
2022-11-28T19:53:44.971955Z job-manager.crit[0]: module exiting abnormally
2022-11-28T19:53:44.974528Z broker.err[0]: rc1.0: flux-module: broker.insmod: Invalid argument
2022-11-28T19:53:44.976154Z broker.err[0]: rc1.0: /usr/local/etc/flux/rc1 Exited (rc=1) 0.7s
+-----------------------------------------------------
|  Entering Flux system instance recovery mode.
|  All resources will be offline during recovery.
|  Any rc1 failures noted above may result in
|   reduced functionality until manually corrected.
|  KVS changes made during recovery are persitent.
|
|  broker.rc1_path    /usr/local/etc/flux/rc1
|  broker.rc3_path    /usr/local/etc/flux/rc3
|  config.path        /usr/local/etc/flux/system/conf.d
|  statedir           /var/lib/flux
|
|  Exit this shell when finished.
+-----------------------------------------------------
$ 
```

grondo:
Oh nice! I wasn't thinking of that much detail but that is pretty nice :+1:

grondo:
Oh, _cool_! Can this also be used as an easier method to solve #4466?

garlick:
Ah thanks for finding that issue!  (I was looking for it)  Yes, we'll just need  to set recovery mode but restore to an instance that puts `content.sqlite` in `rundir` not `statedir`.  I guess a different `flux start` option would be needed.  Thinking up a name will be the hardest part :-)

garlick:
I tentatively just made the `--recover[=ARG]` optional argument accept a dump file.

I hope I'm not making this too confusing by conflating two use cases with one option.  If ARG is a statedir (or is missing and the statedir is presumed to be `/var/lib/flux`), then changes are persistent.  For example, you can change the KVS or submit jobs (which won't run until the resources are online when the instance is running normally).  However, if ARG is a dump file, nothing you do gets preserved - there is no statedir, and `content.sqlite` is created in rundir instead, which is cleaned up on exit.

Furthermore, if `--recover` with no ARG is specified, the system instance configuration is loaded, but if you load from a system instance dump file, you must add the `--sysconfig` option (or `-o,--config-path=PATH` for a path other than the default).

Confusing?

I did modify the startup banner so if you're running from a dump, you get something like
```
+-----------------------------------------------------
| Entering Flux recovery mode.
| All resources will be offline during recovery.
| Any rc1 failures noted above may result in
|  reduced functionality until manually corrected.
|
| broker.rc1_path    -
| broker.rc3_path    -
| config.path        -
| statedir           none, any changes will not be preserved
|
| Exit this shell when finished.
+-----------------------------------------------------
```

garlick:
This is probably ready for a review when someone has a moment.

src/cmd/flux-start.c
```diff
@@ -482,6 +503,21 @@ int execvp_argz (char *argz, size_t argz_len)
     return -1;
 }
 
+bool system_instance_is_running (void)
```
chu11, line 506: This is fine, but i couldn't help but wonder if there's a better way or "correct way" to do this, like perhaps calling systemd directly?

chu11 (approved):
nothing really caught my eye, just one side comment

src/cmd/flux-start.c
garlick: Heh, maybe, although I don't have the first idea of how to do that, and I would imagine it would be more complex...

grondo: FWIW, this seems fine to me since it would work even if the system instance is started without systemd or via some other process management scheme (e.g. in cloud scenarios if that is ever a thing?)

What struck me here is that this duplicates code that is already present in `flux_open(3)` -- would it be simpler to just unset `FLUX_URI` and then call `flux_open (NULL, 0)`?

t/t2315-resource-system.t
```diff
@@ -163,7 +163,7 @@ test_expect_success 'resources can be configured in TOML' '
 	hosts = "foo[3-10]"
 	properties = ["batch"]
 	EOF
-	flux start -s 1 \
+	flux start -s 11 \
```
grondo, line 166: This change will make this test take significantly longer, especially in CI and with coverage enabled. My suggestion would be this instead:
```diff
diff --git a/t/t2315-resource-system.t b/t/t2315-resource-system.t
index b3a0c665f..10ab4c0d1 100755
--- a/t/t2315-resource-system.t
+++ b/t/t2315-resource-system.t
@@ -165,12 +165,12 @@ test_expect_success 'resources can be configured in TOML' '
        EOF
        flux start -s 1 \
                -o,--config-path=$(pwd)/${name},-Slog-filename=${name}/logfile \
-               flux resource list -s up > ${name}/output &&
+               flux resource list -s all > ${name}/output &&
        test_debug "cat ${name}/output" &&
        cat <<-EOF >${name}/expected &&
             STATE PROPERTIES NNODES   NCORES    NGPUS NODELIST
-               up debug           3       12        0 foo[0-2]
-               up batch           8       32        4 foo[3-10]
+              all debug           3       12        0 foo[0-2]
+              all batch           8       32        4 foo[3-10]
        EOF
        test_cmp ${name}/expected ${name}/output
```

grondo (approved):
Nicely done, just had one extra suggestion (besides the other comment about `flux_open()`)

src/cmd/flux-start.c
garlick: Good point!  Will fix.

garlick:
Fixed those things, and I guess I'll set MWP!  Thanks guys.

PR #4786, opened 2022-11-29
job-manager: add version to checkpoint object and extra checkpoint coverage

chu11:
Per comment in #4776 it might be a good idea to add a version to the job-manager checkpoint object.

Also add some extra tests.  I created the dump files by manually doing:

```
src/cmd/flux start
do some queue enable/disable/config
flux dump <file>
```

on the flux 0.45.0 tag.  Tests caught the issue @garlick mentioned in #4776 within that branch.

This admittedly was a smaller change than I anticipated.  I could just add it to #4776???


chu11:
it occurred to me that i should probably add some valid checkpoints/dump tests for the new "version 0" checkpoint object.  I generated by doing:

```
#!/bin/sh

flux mini run /bin/true

flux queue disable "foof"

flux queue enable

flux kvs get checkpoint.job-manager

flux dump t/job-manager/dumps/valid/dump-0-46-0-queue-anon-v0.tar.bz2

flux config load <<-EOT
[queues.batch]
[queues.debug]
EOT

flux queue disable --all "foobar"

flux queue enable --all

flux kvs get checkpoint.job-manager

flux dump t/job-manager/dumps/valid/dump-0-46-0-queues-named-v0.tar.bz2
```

note that this may not be necessary.  If we're confident this and PR #4776 is merged before a release is done, could do this in that PR instead.

garlick:
Is the version going to be changed to 1 in #4776?  If not, then maybe we should wait on this until we need a version 1?  It seems unnecessary to put version:0 in the object that otherwise hasn't changed, when we have to be able to parse version 0 without it for backwards compatibility.

Also, it does make the job manager restart test a bit "queue centric" when each dump is run with and without a queue config.  Anticipating that other things will be added, maybe ti would be best to just add the historical dumps with queues to make sure they can be parsed, but not worry about how that interacts with the config in the "bulk" dump testing.  Instead, add some focused tests that specifically ensure enable/stop state is handled properly for the various cases.

chu11:
> Is the version going to be changed to 1 in https://github.com/flux-framework/flux-core/pull/4776? If not, then maybe we should wait on this until we need a version 1? It seems unnecessary to put version:0 in the object that otherwise hasn't changed, when we have to be able to parse version 0 without it for backwards compatibility.

Yeah, it'll be version 1 in #4776.  I guess this was more of a "pre-req" commit / PR.  We could easily just put the version commit into that PR instead.

> Also, it does make the job manager restart test a bit "queue centric" when each dump is run with and without a queue config. Anticipating that other things will be added, maybe ti would be best to just add the historical dumps with queues to make sure they can be parsed, but not worry about how that interacts with the config in the "bulk" dump testing. Instead, add some focused tests that specifically ensure enable/stop state is handled properly for the various cases.

I guess what you're saying is I may have "permutated" a bit much with these new tests.  Which is fair.  Perhaps should not add the extra tests in `t2219-job-manager-restart`, just add the new dumps only.

Given this is now a one commit PR, I'm just going to close absorb this into #4776.

PR #4788, opened 2022-11-30
libtaskmap: support RFC 34 unknown task maps

grondo:
RFC 34 denotes that a task map with no task blocks indicates that the mapping is unknown.

Update libtaskmap to treat empty mappings as unknown, e.g. `taskmap_nnodes()` now returns `-1` instead of `0` for an "unknown" mapping. A `taskmap_unknown()` function is added for querying if a mapping is unknown.

Explicitly detect cases where an unknown mapping is provided to the job shell and issue an error.


src/common/libtaskmap/taskmap.c
```diff
@@ -618,6 +618,9 @@ static char *taskmap_encode_pmi (const struct taskmap *map)
     zlistx_t *l;
```
garlick, line 618: In commit message for "libtaskmap: encode an unknown PMI_..."
s/uknown/unknown/

garlick (approved):
LGTM!

grondo:
Thanks, fixed that and I'll set MWP.

PR #4789, opened 2022-11-30
add file broadcast toolset

garlick:
This is a redo of #4546 with some proposed better tooling.

To review, the idea was that we could mmap(2) a (possibly large) file into the rank 0 broker, then use the hierarchical scalability properties of the content cache to _pull_ the file to local storage on each node of a job.  The basic functionality is to create the mapping and emit an object that includes some file metadata and an array of blobrefs.  The blobrefs refer to regions of the actual external file - this data does not land in the backing store.   The object can be transferred to a "client" who can then use it to recreate the file through the content cache.

That PR only supported mapping individual files.  I was thinking of a use case like transferring a large file system image.   That worked well, but the tooling was cumbersome if you wanted to transfer a small data set or an executable and its dependent DSOs, which seem like they would be nice use cases to handle.  So in this PR the tooling supports a "tar" like utility for walking trees of files and mapping them, and the concept of "tags".  All mapped files are assigned a tag, and once mapped, they are referred to _only_ by tag.

A group of files is transferred by passing in one or more tags to a query RPC, and receiving a file list in the form of arrays of blobrefs in a multi-response RPC.  The blobrefs refer to _fileref_ objects that match the tags, which in turn contain the blobrefs for the file content as described above.  The list can be further filtered at the source by applying an optional glob pattern on the file names.  By loading all the blobrefs, the client recreates the original files.  A shell plugin is available that performs this function within a job.

**Examples**

The following command maps the flux-core source tree with the tag `flux`, and an Ubuntu install image with the tag `image`:
```console
$ flux file map -C /home/garlick/proj --tags=flux flux-core
$ flux file map -C /home/garlick/Downloads --tags=image ubuntu-22.04-desktop-amd64.iso
$ flux file list -v --tags=image
f 0664 3654957056 ubuntu-22.04-desktop-amd64.iso
$ flux file list -v --tags=flux
f 0664     1106 flux-core/.pre-commit-config.yaml
f 0664      554 flux-core/codecov.yml
f 0664    59254 flux-core/aclocal.m4
[snip]
```

This runs a job with the `image` tagged file copied to $FLUX_JOB_TMPDIR on each node:
```console
$ flux mini run -o copyin.tags=image -o verbose=2 bash -c 'ls -l $FLUX_JOB_TMPDIR'
[snip]
0.036s: flux-shell[0]: DEBUG: copyin: => /tmp/flux-qgI86c/jobtmp-0-ƒ31mXGAzT
0.040s: flux-shell[0]: TRACE: copyin: f 0664 3654957056 ubuntu-22.04-desktop-amd64.iso
12.477s: flux-shell[0]: DEBUG: copyin: 1 files 293.8MB/s
[snip]
total 3569300
-rw-rw-r-- 1 garlick garlick 3654957056 Nov 30 14:11 ubuntu-22.04-desktop-amd64.iso
[snip]
```

This runs a job with the `flux` taggged files copied to $FLUX_JOB_TMPDIR on each node:
```console
$ flux mini run -o copyin.tags=flux -o verbose=1 bash -c 'ls -l $FLUX_JOB_TMPDIR'
[snip]
0.037s: flux-shell[0]: DEBUG: copyin: => /tmp/flux-qgI86c/jobtmp-0-ƒ6A7vUd9R
3.050s: flux-shell[0]: DEBUG: copyin: 13391 files 114.7MB/s
[snip]
total 4
drwxrwxr-x 13 garlick garlick 4096 Nov 30 14:18 flux-core
[snip]
```

The general idea was that in a batch script, which runs in your own private instance, you could map some files from a global file system at the top of the script, organizing them with tags, then run jobs that that specify tags of interest to that job.

It's possible to copy files outside of a job also.   For example, this copies the `flux` files to /tmp on broker ranks 1-7:
```console
$ flux exec -r1-7 flux file get --tags=flux -C /tmp
```
Other details are:
- there is a default tag named `main` which is assumed if a tag is not specified (maybe a batch job only has one set of files and then why bother with tags)
- running a job with `-ocopyin` by itself copies the `main` (default) tag
- running a job with `-ocopyin.destdir=/tmp/xyz` copies files to the specified directory rather than $FLUX_JOB_TMPDIR
- running a job wtih `-ocopyin.pattern=foo/*` copies only the files whose paths match the pattern
- sparse files are transferred efficiently
- symlinks and directories are supported too
- as an optimization, small files (up to 4K) store data directly in the fileref object.

Known limitations:
- Only the rank 0 broker can map files
- Streaming data can't be mapped without copying it to files first
- Only the rank 0 broker can answer the list query

There are some more todos below, but I thought maybe it would be a good idea to seek feedback on the general approach before going there.

**todo**

- [x] docs
- [ ] scalability testing on a real cluster

Edit: dropped improving performance of "list" transfer within job shells for a future PR if needed.

### DIFF
--- a/doc/man1/flux-filemap.rst
+++ b/doc/man1/flux-filemap.rst
@@ -1,0 +1,148 @@
+.. flux-help-description: Map files into a Flux instance
+
+===============
+flux-filemap(1)
+===============
+
+
+SYNOPSIS
+========
+
+**flux** **filemap** **map** [*--tags=LIST*] [*-C DIR*] *PATH* ...
+
+**flux** **filemap** **unmap** [*--tags=LIST*]
+
+**flux** **filemap** **list** [*--tags=LIST*] [*--long*] [*PATTERN*]
+
+**flux** **filemap** **get** [*--tags=LIST*] [*-C DIR*] [*PATTERN*]
+
+
+DESCRIPTION
+===========
+
+``flux-filemap`` uses :linux:man2:`mmap` to map files into the rank 0 broker
+*content cache*.  After mapping, the files may be extracted on any broker rank,
+taking advantage of scalability properties of the distributed cache to move the
+data.  The files are treated as read-only and must not change while mapped.
+
+``flux-filemap map`` maps one or more file *PATH* arguments.  It must be run
+on the rank 0 broker, such as within a batch script, and the files must be
+directly accessible by the rank 0 broker.  If a PATH refers to a directory,
+the directory is recursively mapped.  If a file is encountered that is not
+readable, or has a type other than regular file, directory, or symbolic link,
+a fatal error occurs.  Sparse files such as file system images for virtual
+machines are mapped efficiently.  File discretionary access permission are
+preserved, but file attributes, ACLs, and group ownership are not.
+
+``flux-filemap list`` lists mapped files.  Optionally, a :man7:`glob` pattern
+may be specified to filter the list.
+
+``flux-filemap get`` extracts mapped files and may be run on any broker or
+across all brokers using :man1:`flux-exec`.  Optionally, a :man7:`glob` pattern
+may be specified to filter the list.  When extracting mapped files in parallel,
+take care to specify a *--directory* that is not shared and is not on a network
+file system without considering the ramifications.
+
+``flux-filemap unmap`` unmaps mapped files.
+
+The ``stage-in`` shell plugin described in :man1:`flux-shell` may be used to
+extract previously mapped files into $FLUX_JOB_TMPDIR or another directory.
+
+OPTIONS
+=======
+
+**-h, --help**
+   Display options and exit
+
+**-T, --tags=LIST**
+   Specify a comma separated list of *tags*.  If no tags are specified,
+   the *main* tag is assumed.
+
+**-C, --directory=DIR**
+   Change to the specified directory before performing the operation
+   (*map* and *get* subcommands only).
+
+**-v, --verbose=[LEVEL]**
+   Increase output verbosity (*map* and *get* subcommands only).
+
+**-l, --long**
+   Include more detail in file listing (*list* subcommand only).
+
+**--direct**
+   Avoid indirection through the content cache when fetching the top level
+   data for each file.  This may be fastest for a single or small number of
+   clients, but will scale poorly when performed in parallel (*get* subcommand
+   only).
+
+**--blobref**
+   List blobrefs (*list* subcommand only).
+
+**--fileref**
+   List fileref objects (*list* subcommand only).
+
+EXAMPLE
+=======
+
+Example 1:  a batch script that copies data from ``/project/dataset1`` to a
+temporary directory on each node of the batch allocation where it can be used
+by multiple jobs.
+
+.. code-block:: console
+
+  #!/bin/bash
+
+  flux filemap map -C /project dataset1
+  flux exec -r all mkdir -p /tmp/project
+  flux exec -r all flux filemap get -C /tmp/project
+
+  # app1 and app2 have access to local copy of dataset1
+  flux mini run -N1024 app1 --input /tmp/project/dataset1
+  flux mini run -N1024 app2 --input /tmp/project/dataset1
+
+  # clean up
+  flux exec -r all rm -rf /tmp/project
+  flux filemap unmap
+
+Example 2: a batch script that maps two data sets with tags, then uses the
+``stage-in`` shell plugin to selectively copy them to $FLUX_JOB_TMPDIR,
+which is automatically cleaned up after each job.
+
+.. code-block:: console
+
+  #!/bin/bash
+
+  flux filemap map --tags=ds1 -C /project dataset1
+  flux filemap map --tags=ds2 -C /project dataset2
+
+  # App0 uses $FLUX_JOB_TMPDIR/dataset1 and $FLUX_JOB_TMPDIR/dataset2
+  flux mini run -N1024 -o stage-in.tags=ds1,ds2 App0
+
+  # App1 uses only $FLUX_JOB_TMPDIR/dataset1
+  flux mini run -N1024 -o stage-in.tags=ds1 App1
+
+  # App2 uses only $FLUX_JOB_TMPDIR/dataset2
+  flux mini run -N1024 -o stage-in.tags=ds2 App2
+
+  # clean up
+  flux filemap unmap --tags=ds1,ds2
+
+CAVEATS
+=======
+
+The rank 0 Flux broker may die with a SIGBUS error if a mapped file is removed
+or truncated, and subsequently accessed, since that renders pages mapped into
+the brokers address space invalid.
+
+If mapped file content changes, access may fail if the original data is not
+cached.  Under no circumstances will the new content be returned.
+
+RESOURCES
+=========
+
+Flux: http://flux-framework.org
+
+
+SEE ALSO
+========
+
+:man1:`flux-shell`,
--- a/doc/man1/flux-shell.rst
+++ b/doc/man1/flux-shell.rst
@@ -320,6 +320,24 @@ options are supported by the builtin plugins of ``flux-shell``:
   this will also be passed to the invoked plugin. Normally, this option will
   be set by the ``flux mini --taskmap`` option.
 
+**stage-in**
+  Copy files to $FLUX_JOB_TMPDIR that were previously mapped using
+  :man1:`flux-filemap`.
+
+**stage-in.tags**
+  Select files to copy by specifying their tags.  By default the ``main``
+  tag is used.
+
+**stage-in.pattern**
+  Further filter the selected files to copy using a :man7:`glob` pattern.
+
+**stage-in.destdir**
+  Copy files to the specified directory instead of $FLUX_JOB_TMPDIR.  Careful!
+  The $FLUX_JOB_TMPDIR is cleaned up when the job exits, is guaranteed to
+  be unique, and is generally on fast local storage.  Make sure ``destdir``
+  does not specify a network file system or a shared directory without
+  considering the ramifications.
+
 SHELL INITRC
 ============
 
--- a/src/broker/content-mmap.c
+++ b/src/broker/content-mmap.c
@@ -1,0 +1,766 @@
+/************************************************************\
+ * Copyright 2022 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* content-mmap.c - map files into content cache on rank 0
+ *
+ * Each file is represented by a 'struct content_region' that includes
+ * a 'fileref' object containing the file's metadata and blobrefs for content.
+ * The region also contains a pointer to mmap(2)ed memory for the file's
+ * content.
+ *
+ * All files have one or more tags, so the regions are placed in a
+ * hash-of-lists where the list names are tags, and the entries are struct
+ * content_regions.  When files are listed or removed, the requestor provides
+ * one or more tags.  (flux filemap has an implicit "main" tag, but that
+ * is entirely implemented in the tools, not here).
+ *
+ * The general idea is that one makes a list query with tags and an optional
+ * glob and gets back a list of blobrefs.  Each blobref represents a serialized
+ * 'fileref' object that corresponds to a file/content_region.  Within the
+ * fileref is optionally a vector of blobrefs representing data content.
+ * Having obtained this list, the client can use it to pull the fileref, and
+ * subsequently its data through the content cache.
+ *
+ * The content-cache calls content_mmap_region_lookup() on rank 0 when it
+ * doesn't have a requested blobref in cache, and only consults the backing
+ * store when that fails.  If content_mmap_region_lookup() succeeds, the
+ * content-cache takes a reference on the struct content_region.  When we
+ * request to unmap a region, the munmap(2) and free of the struct is delayed
+ * until all content-cache references are dropped.
+ *
+ * Slightly tricky optimization:
+ * To speed up content_mmap_region_lookup() we have mm->cache, which is used
+ * to find a content_region given a hash.  The cache contains hash keys for
+ * both mmapped data (fileref.blobvec) and serialized fileref objects.
+ * A given hash may appear in multiple files or parts of the same file,
+ * so when a file is mapped, we put all its hashes in mm->cache except those
+ * that are already mapped.  If nothing is unmapped, then we know all the
+ * blobrefs for all the files will remain valid.  However when something is
+ * unmapped we could be losing pieces of unrelated files.  Since unmaps are
+ * bulk operations involving tags, we just walk the entire hash-of-lists
+ * at that time and restore any missing cache entries.
+ *
+ * Safety issue:
+ * The content addressable storage model relies on the fact that once hashed,
+ * data does not change.  However, this cannot be assured when the data is
+ * mmapped from a file that may not be protected from updates.  To avoid
+ * propagating bad data in the cache, content_mmap_validate() is called each
+ * time an mmapped cache entry is accessed.  This function recomputes the
+ * hash to make sure the content has not changed.  If the data has changed,
+ * the content-cache returns an error to the requestor.  In addition, mmapped
+ * pages could become invalid if the size of a mapped file is reduced.
+ * Accessing invalid pages could cause the broker to crash with SIGBUS.  To
+ * mitigate this, content_mmap_validate() also calls stat(2) on the file to
+ * make sure the memory region is still valid.  This is not foolproof because
+ * it is inherently a time-of-check-time-of-use problem.  In fact we rate
+ * limit the calls to stat(2) to avoid a "stat storm" when a file with many
+ * blobrefs is accessed, which increases the window where it could have
+ * changed.  But it's likely better than not checking at all.
+ */
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+#include <sys/mman.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <limits.h>
+#include <stdlib.h>
+#include <fnmatch.h>
+#include <jansson.h>
+#include <assert.h>
+#include <flux/core.h>
+
+#include "src/common/libczmqcontainers/czmq_containers.h"
+#include "src/common/libutil/hola.h"
+#include "src/common/libutil/errno_safe.h"
+#include "src/common/libutil/errprintf.h"
+#include "src/common/libutil/blobref.h"
+#include "src/common/libutil/fileref.h"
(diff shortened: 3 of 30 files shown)


garlick:
Pushing again to fix a memory leak in hola.c found running sharness tests under valgrind, and to fix a broken && chain found by CI and one remaining CodeQL problem.  Some incremental development was squashed.

garlick:
On 64 nodes of corona, a 1024 "node" flux instance (cyclic allocation):

Test file: 2.6G tar file of devel work (uncompressed)

Test command `flux mini run -N$size -o stage-in /bin/true`

Times for each size run, with rate calculated as 2.9G*$size / time:

- size=1, time=9.1s, rate=285MB/s
- size=2, time=19.0s, rate=273MB/s
- size=4, time=21.0s, rate=495MB/s
- size=8, time=23.8s, rate=873MB/s
- size=16, time=30.5s, rate=1363MB/s
- size=32, time=35.6s, rate=2337MB/s
- size=64, time=36.6s, rate=4546MB/s
- size=128, time=37.9s, rate=9781MB/s
- size=256, time=37.5s, rate=17749MB/s
- size=512, time=126.3s, rate=10539MB/s
- size=1024, time=171.2s, rate=15551MB/s

garlick:
This testing on corona was all going swimmingly but I did encounter an error I cannot explain where during stage-in (of about 128K files), I got a content.load failure on a blobref.  That's a big uh oh.  Working to reproduce that now and get some debug.

chu11:
things were going swimmingly for me on catalyst, I just did a `time` on

```
time flux exec -r all flux filemap get -C /tmp/achu/filemaptest
```

similar to your results, basically no time difference from 2 to 256 brokers.

but when I got to 512 brokers (128 nodes, 4 tasks-per-node) got:

```
flux-filemap: enwiki.txt: Can't unlink already-existing object                                                                              
flux-filemap: enwiki.txt: Can't unlink already-existing object                                                                              
flux-filemap: enwiki.txt: Can't create 'enwiki.txt'                                                                                         
flux-filemap: enwiki.txt: Can't create 'enwiki.txt'                                                                                         
flux-filemap: enwiki.txt: Can't stat existing object                                                                                        
flux-filemap: enwiki.txt: Can't create 'enwiki.txt'                                                                                         
flux-filemap: enwiki.txt: Can't create 'enwiki.txt'                                                                                         
flux-filemap: enwiki.txt: Can't unlink already-existing object                                                                              
flux-filemap: enwiki.txt: Can't unlink already-existing object                                                                              
flux-filemap: enwiki.txt: Can't create 'enwiki.txt'     
```

Dunno if that's caused by a similar underlying problem to what you saw.  My file is ~9.4G.

garlick:
Hmm those would seem to be extraction errors from libarchive.  Did you have muliple brokers per node extracting to the same directory?

garlick:
>  I did encounter an error I cannot explain where during stage-in (of about 128K files), I got a content.load failure on a blobref.

I was somewhat stupidly mapping the tree I was running out of.  My best theory (having tried in vain to reproduce this) is that something that was mapped changed, and then the textual error that's generated from the content.load when this happens was dropped when it was transferred to a rank > 0 via internal RPC.   I'll correct that error message handling and try this again tomorrow the same way I ran today to see if it reproduces.

The scaling seems pretty good, especially when there are duplicate blobrefs in the file as I found when I tried to share a mostly empty 1G ext4 file system image.  Lots of 1MB blobs of all zeros in there, except where the superblock backsups are!

Anyway, I'm pleased with the scaling results so far.  Just need to run down these problems.

chu11:
> Hmm those would seem to be extraction errors from libarchive. Did you have muliple brokers per node extracting to the same directory?

Oh duh!  I had no problems 1 broker per node, but once I ran out of nodes I went to > 1 broker per node and was writing to the same dir.  Should script something a little smarter.

garlick:
Force pushed with some error handling fixes, additional testing, and warnings in the man page about file content changing after mapping.

A big caveat added to the docs is if a file is removed or truncated the broker can get a SIGBUS.  I don't think that's a show stopper since this stuff is instance owner only and you can shoot yourself in the foot if you like, but it is a big penalty if you screw up or if you map a dataset from /usr/gapps or similar and somebody else does something. 

At least due to the content addressable nature of the cache, you'll either get the original data or die trying.  No silently passing on data that's part way through being updated (unless this happens during the short window when `flux content filemap map` is run).

chu11 (commented):
noticed you updated the commit series, so just parking this review which is only through the first commit (which I don't believe changed given the commit hash)

src/common/libutil/fileref.c
```diff
@@ -0,0 +1,356 @@
...
+                        "{s:s s:I s:I s:I s:i s?s s:o}",
+                        "path", &path,
+                        "size", &size,
+                        "mtime", &mtime,
+                        "ctime", &ctime,
+                        "mode", &mode,
+                        "data", &data,
+                        "blobvec", &blobvec) < 0) {
+        n = snprintf (buf, bufsize, "invalid fileref");
+    }
```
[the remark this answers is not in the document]
garlick: Meh, having to check for a missing argument error when it's obviously there at the call site is going to look silly.  I'll just have it avoid crashing in that case and assume if you didn't pass in a buffer, you didn't want the result. :-)

src/common/libutil/fileref.c
```diff
@@ -0,0 +1,356 @@
...
+                break;
+            goto error;
+        }
+        offset = cursor;
+        if (offset < size) {
+            // N.B. If there are no holes, sets cursor == size
+            if ((cursor = lseek (fd, offset, SEEK_HOLE)) == (off_t)-1)
+                goto error;
+            if (blobsize > cursor - offset)
+                blobsize = cursor - offset;
```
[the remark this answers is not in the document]
garlick: You're right, it's a little difficult to figure out as is.  Let me see if I can simplify the code a bit.

src/common/libutil/fileref.c
```diff
@@ -0,0 +1,356 @@
...
+    if (chunksize < 0) {
+        errprintf (error, "chunksize cannot be negative");
+        goto inval;
+    }
+    if (!fullpath)
+        fullpath = path;
+    /* Refuse to put absolute paths in a fileref object.  Strip leading '/'.
+     */
+    while (*path == '/')
+        path++;
```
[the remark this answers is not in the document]
garlick: Correct - it would be a bit too easy to do something like
```
$ flux filemap /g/gapps/mydata
$ flux mini run -o copyin ...
```
and then parallel overwrite the data in /usr/gapps.  I'll expand that comment a bit!

src/common/libutil/fileref.c
```diff
@@ -0,0 +1,356 @@
...
+    if (strlen (path) == 0)
+        path = ".";
+    /* Avoid TOCTOU in S_ISREG case by opening before checking its type.
+     */
+    if ((fd = open (fullpath, O_RDONLY | O_NOFOLLOW)) < 0)
+        rc = lstat (fullpath, &sb);
+    else
+        rc = fstat (fd, &sb);
+    if (rc < 0) {
+        errprintf (error, "%s: %s", path, strerror (errno));
```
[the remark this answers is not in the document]
garlick: Agreed that errors should not use the modified (relativized) path but instead should use the original `path` as specified by the user.   The `fullpath` is `path` post-processed by `realpath()`.  It is always fully qualified and may look different from what the user specified.

src/common/libutil/hola.h
```diff
@@ -0,0 +1,53 @@
...
+\************************************************************/
+
+#ifndef _UTIL_HOLA_H
+#define _UTIL_HOLA_H
+
+#include "src/common/libczmqcontainers/czmq_containers.h"
+
+enum {
+    HOLA_AUTOCREATE = 1,    // create list on first addition
+    HOLA_AUTODESTROY = 2,   // destroy list on last removal
```
chu11, line 18: some comments would be good here to indicate they make it so user doesn't have to call `hola_hash_add()` or `hola_hash_delete()`.  I was initially confused what their purpose was, b/c I thought this behavior would be the default behavior.

src/common/libutil/test/hola.c
```diff
@@ -0,0 +1,363 @@
...
+
+    if (!(checklist = calloc (len, sizeof (checklist[0]))))
+        BAIL_OUT ("out of memory");
+    h = hola_create (HOLA_AUTOCREATE);
+    if (!h)
+        BAIL_OUT ("hola_create failed");
+    for (int i = 0; i < len; i++) {
+        if (!hola_list_add_end (h, t[i].key, t[i].val))
+        BAIL_OUT ("could not populate test object for iteration");
+    }
```
chu11, line 220: missing indent :-)

chu11 (commented):
through the big `broker/content-mmap: map files into content cache ` commit.  I think I understand the code now.  Maybe as I get to other commits other parts will become more clear.

src/common/libutil/test/fileref.c
```diff
@@ -0,0 +1,524 @@
...
+            memset (buf, *cp, blocksize);
+            size_t n = write (fd, buf, blocksize);
+            if (n < 0)
+                BAIL_OUT ("error writing to %s: %s", name, strerror (errno));
+            if (n < blocksize)
+                BAIL_OUT ("short write to %s", name);
+        }
+    }
+    if (ftruncate (fd, blocksize*strlen (spec)) < 0)
+        BAIL_OUT ("error truncating %s: %s", name, strerror (errno));
```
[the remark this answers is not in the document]
garlick: Hmm, must be left over from an earlier version of that function.  Good catch!

src/common/libutil/hola.h
garlick: I was thinking there may be a use case for the class where the lists are only created intentionally and it would be desirable to fail an add to an unknown list.  Then of course when the list is emptied you wouldn't necessarily want it to be deleted.  Also, you may not want to delete empty lists if you expect a limited set of lists to transition back and forth from zero items and nonzero items, for performance.

I'll add a bit more doc.

src/common/libutil/hola.c
```diff
@@ -0,0 +1,287 @@
...
+        hola->have_list_destructor = 1;
+    }
+}
+void hola_set_list_duplicator (struct hola *hola, zlistx_duplicator_fn fun)
+{
+    if (hola) {
+        hola->list_duplicator = fun;
+        hola->have_list_duplicator= 1;
+    }
+}
```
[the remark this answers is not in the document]
garlick: I just dropped the `have_list_X` flags.  They were there to record that the user wanted functions installed, including if they wanted them set to NULL (thinking that setting to NULL was overwriting an internal default).  NULL is actually the default so the flags aren't necessary.

t/t0029-filemap-cmd.t
```diff
@@ -0,0 +1,244 @@
...
+	flux filemap map ./testfile
+'
+test_expect_success 'truncate mapped test file' '
+	cp /dev/null testfile
+'
+test_expect_success 'size reduction should cause an error' '
+	rm -f copydir/testfile &&
+	test_must_fail flux filemap get -C copydir 2>reduced.err &&
+	grep changed reduced.err
+'
```
chu11, line 242: just for coverage, increase file and make sure still works?

chu11: also possibly relevant given my prior perms question, test coverage of change in permissions?

chu11 (commented):
through the filemap tool tests, nearing the end

src/broker/content-mmap.c
```diff
@@ -0,0 +1,766 @@
...
+    assert (reg->data != NULL);
+    assert (data >= reg->data);
+    assert (data + data_size <= reg->data + reg->data_size);
+
+    if (monotime_since (reg->last_check)/1000 >= max_check_age) {
+        struct stat sb;
+
+        if (stat (reg->fullpath, &sb) < 0
+            || sb.st_size < reg->data_size)
+            return false;
```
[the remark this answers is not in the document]
garlick: What problem were you thinking this would solve?  The stat already fails if we can no longer read the file.  Since we're taking a copy from a specific point in time, the main thing we need to protect against is SIGBUS errors (file shrinks) and violating the content addressable storage invariant that blobs never change IMHO.

src/cmd/builtin/filemap.c
```diff
@@ -0,0 +1,644 @@
...
+    ctx.tags = get_list_option (p, "tags", "main");
+    ctx.h = builtin_get_flux_handle (p);
+    if (!ctx.h)
+        log_err_exit ("flux_open");
+    while (n < ac) {
+        const char *path = av[n++];
+        struct stat sb;
+
+        if (lstat (path, &sb) < 0)
+            log_err_exit ("%s", path);
```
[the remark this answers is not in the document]
garlick: I think `dirwalk()` can only be called on a directory so the lstat is necessary to determine if we have a directory or something else.  Possibly I'm not understanding the suggestion (maybe a diff would help?)

t/t0029-filemap-cmd.t
garlick: That turns out to be hard to test because we cannot force a re-stat(2) of the file until the rate limit runs time (5s) runs out.  Anyway, recovery after an error has been detected is not really part of the design.  In fact we might want to unmap the file in that case and invalidate the other cache entries at some point.

src/broker/content-mmap.c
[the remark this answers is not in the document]
chu11: oh yeah, i'm stupid ... the `stat()` is effectively a perms check :P

src/cmd/builtin/filemap.c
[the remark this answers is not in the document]
chu11: I was just thinking this ...

```
diff --git a/src/cmd/builtin/filemap.c b/src/cmd/builtin/filemap.c
index 2f40ea769..9bf2bac33 100644
--- a/src/cmd/builtin/filemap.c
+++ b/src/cmd/builtin/filemap.c
@@ -245,9 +245,9 @@ static int subcmd_map (optparse_t *p, int ac, char *av[])
         const char *path = av[n++];
         struct stat sb;
 
-        if (lstat (path, &sb) < 0)
-            log_err_exit ("%s", path);
         if (S_ISDIR (sb.st_mode)) {
+            if (lstat (path, &sb) < 0)
+                log_err_exit ("%s", path);
             if (dirwalk (path, flags, map_visitor, &ctx) < 0)
                 log_err_exit ("%s", path);
         }
```

b/c in the `map_add` path you'd be calling `lstat` twice.  This is not a huge deal of course.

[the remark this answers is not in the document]
garlick: the S_ISDIR check has to be after the lstat() call though...

doc/man1/flux-filemap.rst
```diff
@@ -0,0 +1,148 @@
+.. flux-help-description: Map files into a Flux instance
```
chu11, line 1: before i forget, the `flux --help` output in README will now be out of date,  should update.

doc/man1/flux-filemap.rst
```diff
@@ -0,0 +1,148 @@
...
+DESCRIPTION
+===========
+
+``flux-filemap`` uses :linux:man2:`mmap` to map files into the rank 0 broker
+*content cache*.  After mapping, the files may be extracted on any broker rank,
+taking advantage of scalability properties of the distributed cache to move the
+data.  The files are treated as read-only and must not change while mapped.
+
+``flux-filemap map`` maps one or more file *PATH* arguments.  It must be run
+on the rank 0 broker, such as within batch script, and the files must be
```
chu11, line 29: "within batch" -> "within a batch"

doc/man1/flux-shell.rst
```diff
@@ -320,6 +320,24 @@ options are supported by the builtin plugins of ``flux-shell``:
   this will also be passed to the invoked plugin. Normally, this option will
   be set by the ``flux mini --taskmap`` option.
 
+**stage-in**
+  Copy files to $FLUX_JOB_TMPDIR that were previously mapped using
+  :man1:`flux-filemap`.
```
chu11, line 325: we seem to list "common" or "useful" shell options in `flux-mini`.  Do some of these pass the bar to be included there.

chu11 (commented):
skipped the shell plugin commits and went to the manpages first :P

doc/man1/flux-filemap.rst
garlick: AFAIK `flux --help` is automatically updated?

doc/man1/flux-shell.rst
garlick: Maybe a brief mention with references like this?
```diff
diff --git a/doc/man1/flux-mini.rst b/doc/man1/flux-mini.rst
index 002d26a25..4f15125d1 100644
--- a/doc/man1/flux-mini.rst
+++ b/doc/man1/flux-mini.rst
@@ -758,6 +758,9 @@ overridden in some cases:
    Configure the PMI plugin's built-in key exchange algorithm to use a
    virtual tree fanout of ``N`` for key gather/broadcast.  The default is 2.
 
+**stage-in**
+   Copy files previously mapped with :man1:`flux-filemap` to $FLUX_JOB_TMPDIR.
+   See :man1:`flux-shell` for more *stage-in* options.
 
 RESOURCES
 =========
```

doc/man1/flux-filemap.rst
chu11: yes, but not the example output in README.  It's stupid ... I only mention b/c I just randomly noticed it recently.  https://github.com/flux-framework/flux-core/pull/4688

Maybe we need a way to automatically update this.

doc/man1/flux-shell.rst
chu11: sounds good

src/shell/stage-in.c
```diff
@@ -0,0 +1,477 @@
...
+        void *buf;
+        ssize_t n;
+
+        if (!(buf = calloc (1, bufsize)))
+            return -1;
+        if ((n = base64_decode (buf, bufsize, s, len)) < 0) {
+            free (buf);
+            errno = EINVAL;
+            return -1;
+        }
```
chu11, line 102: nit, malloc instead of calloc, since you're going to overwrite the buffer immediately

(i guess this was done in cmd/builtin/filemap.c too)

src/shell/stage-in.c
```diff
@@ -0,0 +1,477 @@
...
+    }
+    archive_write_free (archive);
+    free (orig_dir);
+    return 0;
+error:
+    if (archive)
+        archive_write_free (archive);
+    free (orig_dir);
+    return -1;
+}
```
chu11, line 413: should try to `chdir()` to orig dir on error, in case caller wants to survive the error?

chu11 (commented):
reivew of the stage-in shell commits.  I notice some similarity between code in shell plugin and filemap command.  I guess it just crossed the boundary of "create shared util library"?  Probably just inconvenient enough not to do.

src/shell/stage-in.c
garlick: Regarding the code duplication, yeah, since the two implementations have different styles of error handling, it seemed like a high effort to refactor.  I could be convinced if a few more things need to be changed in both places :-)

garlick:
Force pushed fixups for your  comments @chu11 - thanks.

chu11 (approved):
LGTM!

garlick:
Thanks for all the hard work reviewing this @chu11!  I'll set MWP.

Issue #4790, opened 2022-12-01
broker segfault in prolog_kill_cb() -> flux_subprocess_kill()


garlick:
The rank 0 broker segfaulted on tiogai.
```
[root@tioga6:garlick]# flux version
commands:               0.43.0
libflux-core:           0.43.0
libflux-security:       0.8.0
build-options:          +hwloc==2.8.0+zmq==4.3.4
```
The `coredumpctl info` output is:
```
          PID: 2219049 (flux-broker-0)
           UID: 763 (flux)
           GID: 752 (flux)
        Signal: 11 (SEGV)
     Timestamp: Wed 2022-11-30 16:19:54 PST (2h 7min ago)
  Command Line: broker --config-path=/etc/flux/system/conf.d -Scron.directory=/>
    Executable: /usr/libexec/flux/cmd/flux-broker
 Control Group: /
         Slice: -.slice
       Boot ID: 4501d7661d624d88ba85fa850a884587
    Machine ID: 2ca30a0f38df47cb8f7969147587a154
      Hostname: tioga6
       Storage: /var/lib/systemd/coredump/core.flux-broker-0.763.4501d7661d624d>
       Message: Process 2219049 (flux-broker-0) of user 763 dumped core.
                
                Stack trace of thread 2219096:
                #0  0x00007ffff7b7f991 flux_subprocess_kill (libflux-core.so.2)
                #1  0x00007fffd1fd4b9f prolog_kill_cb (perilog.so)
                #2  0x00007ffff7b92053 ev_invoke_pending (libflux-core.so.2)
                #3  0x00007ffff7b960f8 ev_run (libflux-core.so.2)
                #4  0x00007ffff7b60f3f flux_reactor_run (libflux-core.so.2)
                #5  0x00007fffd223ce30 mod_main (job-manager.so)
                #6  0x0000000000410041 module_thread (flux-broker)
                #7  0x00007ffff79291cf start_thread (libpthread.so.0)
                #8  0x00007ffff629ddd3 __clone (libc.so.6)
                
                Stack trace of thread 2219066:
                #0  0x00007ffff6393a07 epoll_wait (libc.so.6)
                #1  0x00007ffff7b9296f epoll_poll (libflux-core.so.2)
                #2  0x00007ffff7b95546 ev_run (libflux-core.so.2)
                #3  0x00007ffff7b60f3f flux_reactor_run (libflux-core.so.2)
[...]

PR #4793, opened 2022-12-02
broker: fix potential crash after sending SIGKILL to job-manager prolog

grondo:
This PR is a probable fix for the crash described in #4790. I am not 100% certain of the root cause, since I don't see how the subprocess kill future callback could get called after the subprocess handle is destroyed (the subprocess and kill future are destroyed together when the subprocess or job completes), but ensuring the future is only called once on timeout would have prevented this particular crash (and not destroying the future in this case was definitely a bug)


garlick (approved):
Looks good to me!

grondo:
rebased on current master to get the CI fix. I've set MWP.

PR #4794, opened 2022-12-02
ci: revert to ubuntu-20.04 for workflow jobs that fail on ubuntu-latest

grondo:
Problem: The python linting job fails on ubuntu-latest because python 3.6 is no longer supported.

Use runs-on: ubuntu-20.04 in the python-lint job so that python 3.6 can still be used.

See https://github.com/actions/setup-python/issues/355#issuecomment-1335042510

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ jobs:
 
   python-lint:
     name: python linting
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
     - uses: actions/setup-python@v4
       with:
@@ -83,7 +83,7 @@ jobs:
 
   ci-checks:
     needs: [generate-matrix]
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     env:
       TAP_DRIVER_QUIET: 1
       FLUX_TEST_TIMEOUT: 300


garlick (approved):
OK!

grondo:
Thanks! Set MWP

grondo:
Hm, the system test build is getting stuck here:
```
Waiting for flux-system-test-85846 to be ready
Waiting for flux-system-test-85846 to be ready
Waiting for flux-system-test-85846 to be ready
Waiting for flux-system-test-85846 to be ready
Waiting for flux-system-test-85846 to be ready
Waiting for flux-system-test-85846 to be ready
Waiting for flux-system-test-85846 to be ready
Waiting for flux-system-test-85846 to be ready
Waiting for flux-system-test-85846 to be ready
Waiting for flux-system-test-85846 to be ready
```

We've seen these issues before, but never in CI, and I didn't think the change here would affect any of the other builds, but I'm looking into it.

grondo:
My guess is that the `ubuntu-latest` tag was updated to 22.04 today, precipitating the failures in the python linting and the system coverage builds (I guess docker-run-systest.sh doesn't work on ubuntu 22.04). I'll update this PR to revert both builders to `ubuntu-20.04`.

grondo:
Oops, I had forgotten to remove MWP when adding ubuntu-20.04 to the `ci-checks` builder. Oh, well, it seems to have worked...

Issue #4771, opened 2022-11-11
`flux start bash` hangs


grondo:
Starting after flux-core v0.44.0 (I think) `flux start bash` hangs. This was first noted by a hang in the `fluxrm/flux-sched` and `fluxrm/flux-core` latest docker images, which have a default `CMD` of `/usr/bin/flux start /bin/bash`.

I believe this is due to 28facc6d4489227b81f0c296a58cefb960666e41, which sets a new process group for initial program for "non interactive" shell programs. I guess because there is an argument to `flux start` the broker assumes a non-interactive script is being run (have to look at the runat code more carefully), but in this case the shell thinks it should be interactive. When bash sees that it is not in the terminal process group, it issues `kill (0, SIGTTIN)` and hangs.

The fix for the container case is to run `docker run -ti fluxrm/flux-sched flux start`, which replaces the default `CMD`. It will require a little bit of study to determine how we fix the determination of interactive vs non-interactive case in the broker.

grondo:
Oops, should have been closed by #4795.

PR #4795, opened 2022-12-04
broker: improve detection of interactive shells

grondo:
This PR nominally solves #4771 while at the same time  avoiding most hangs when the broker fails to detect that an initial program is supposed to be interactive.

The root of the problem is that the broker runs non-interactive initial programs in a separate process group to aid cleanup, but if the child tries to read from the tty it gets SIGTTIN and ends up stopped in what is effectively an uninterruptible state (requires SIGKILL to be terminated). Since the broker currently only treats lack of an argument as an "interactive" shell, `flux start bash` or `flux start <any shell>` causes this behavior.

To fix this issue the broker has to have a good idea of what arguments might specify an interactive shell. So, instead of just checking for `argz == NULL` to detect an interactive shell, the broker now reads `/etc/shells` and compares the entries and their basenames to the first argument (when there is only a single argument) to detect if the program will be interactive. This will work for cases like `flux start bash` and `flux start /bin/bash`, but not `flux start bash -l`.

Since the solution here isn't 100%, it would be good to also cause an error instead of a hang when an interactive process is run non-interactively by the broker. To detect this case, libsubprocess is extended to add a `FLUX_SUBPROCESS_STOPPED` when the child watcher returns a status of `WIFSTOPPED`. The broker then can issue an error and kill the child process. Since the child is being run non-interactively, I don't think there is a case where this would be unwanted behavior.

Of course, all this would not be necessary if we didn't run non-interactive rc2 in a separate process group. However, at least for now, I think the benefits of doing that (i.e. more reliable cleanup) outweigh the costs of the extra code here.

### DIFF
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -695,7 +695,7 @@ static int create_runat_rc2 (struct runat *r, const char *argz, size_t argz_len)
          */
         if (!isatty (STDIN_FILENO))
             log_msg_exit ("stdin is not a tty - can't run interactive shell");
-        if (runat_push_shell (r, "rc2", 0) < 0)
+        if (runat_push_shell (r, "rc2", argz, 0) < 0)
             return -1;
     }
     else if (argz_count (argz, argz_len) == 1) { // run shell -c "command"
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -661,9 +661,35 @@ static void set_proctitle (uint32_t rank)
     (void)prctl (PR_SET_NAME, proctitle, 0, 0, 0);
 }
 
+static bool is_interactive_shell (const char *argz, size_t argz_len)
+{
+    bool result = false;
+    /*  If no command is specified, then an interactive shell will be run
+     */
+    if (argz == NULL)
+        return true;
+
+    /*  O/w, if command is plain "$SHELL", e.g. bash, zsh, csh, etc.
+     *   then assume interactive shell.
+     */
+    if (argz_count (argz, argz_len) == 1) {
+        char *shell;
+        char *cmd = argz_next (argz, argz_len, NULL);
+        while ((shell = getusershell ())) {
+            if (strcmp (cmd, shell) == 0
+                || strcmp (cmd, basename (shell)) == 0) {
+                result = true;
+                break;
+            }
+        }
+        endusershell ();
+    }
+    return result;
+}
+
 static int create_runat_rc2 (struct runat *r, const char *argz, size_t argz_len)
 {
-    if (argz == NULL) { // run interactive shell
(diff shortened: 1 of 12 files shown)


src/broker/broker.c
```diff
@@ -661,9 +662,63 @@ static void set_proctitle (uint32_t rank)
     (void)prctl (PR_SET_NAME, proctitle, 0, 0, 0);
 }
 
+int etc_shells_to_argz (char **argz, size_t *argz_len)
+{
+    void *buf = NULL;
+    ssize_t len;
+    int fd = -1;;
```
chu11, line 669: typo, `;;`

src/broker/broker.c
```diff
@@ -661,9 +662,63 @@ static void set_proctitle (uint32_t rank)
...
+    int rc = -1;
+    int saved_errno;
+
+    if ((fd = open ("/etc/shells", O_RDONLY)) < 0
+        || (len = read_all (fd, &buf)) < 0
+        || (argz_create_sep (buf, '\n', argz, argz_len)) != 0)
```
chu11, line 675: on my box there is a comment in `/etc/shells`, which i guess just gets slurped up here and compared below?  No big deal, but maybe want to mention as a comment?

src/common/libsubprocess/test/subprocess.c
```diff
@@ -1459,6 +1459,49 @@ void test_state_change (flux_reactor_t *r)
...
+{
+    diag ("state_change_stopped: state = %s",
+          flux_subprocess_state_string (state));
+    if (state == FLUX_SUBPROCESS_STOPPED) {
+        flux_future_t *f = flux_subprocess_kill (p, SIGKILL);
+        ok (true, "subprocess state == STOPPED in state change handler");
+        flux_future_destroy (f);
+    }
+    state_change_cb_count++;
+}
```
chu11, line 1473: the way this test is written I don't think its guaranteed that FLUX_SUBPROCESS_STOPPED was seen.  Perhaps instead set a global counter `stopped_count` and update that counter here.  Then in `test_state_change_stopped()` check that `stopped_count == 1`.

chu11 (approved):
overall LGTM, just a few small nits

src/broker/broker.c
grondo: Oh good catch, I should have noticed that. It should be easy to skip lines that start with `#`

grondo: Hm, I just discovered [`getusershell(3)`](https://linux.die.net/man/3/getusershell), maybe I should just use that here.

grondo: Ok, pushed a new version using `getusershell(3)` which simplifies this code.

chu11:
lgtm!

chu11:
system coverage timed out, kicked it

Issue #4778, opened 2022-11-22
shell: pmi: increase maximum PMI KVS value size


grondo:
From [#4772](https://github.com/flux-framework/flux-core/pull/4772#discussion_r1029682424), @garlick suggests:

> We should consider increasing the PMI max KVS value size to allow cyclic on systems we have today.

We have at least one system with 128 cores, so it would be good to support something like 64 nodes with 128 tasks-per-node.

Also, we wanted to use cyclic task mapping as a possible way to get a shorter hostlist when scale testing flux instance bootstrap times. In that case we were trying to test up to something like 8 tasks-per-node across 3000 nodes.

grondo:
In the end this was not feasible since MPICH and MVAPICH2 seem to use 1024 byte buffers internally which causes errors and even segfaults when the max KVS value size is increased.

A workaround is now in place which sets a `flux.taskmap` key in the PMI KVS which the Flux PMI client can use in place of `PMI_process_mapping` when available. This solves the original impetus for this issue which is to allow a 128 tasks-per-node cyclic distribution since an RFC 34 cyclic task map is more compact than a `PMI_process_mapping`.

Closing this issue.

PR #4796, opened 2022-12-05
pmi: increase max value length to 2048

grondo:
As suggested in #4778, this simple PR increases the  max KVS value length for PMI-1 and 2 in the simple server and therefore the job shell PMI implementation (I believe this is all that is necessary, please let me know if not).

This doubles the size to 2048, which should allow PMI_process_mapping up to about 169 tasks per node. If we want to support 256 tasks per node, we'll need to increase this further. I'm not sure what the repercussions would be for increasing vallen_max further to 3192 or 4096. (Seems like 3192 would be large enough for 256 tasks per node)

### DIFF
--- a/src/common/libpmi/pmi2.h
+++ b/src/common/libpmi/pmi2.h
@@ -33,8 +33,8 @@ extern "C" {
 #define PMI2_ERR_OTHER              14
 
 #define PMI2_MAX_KEYLEN             64
-#define PMI2_MAX_VALLEN             1024
-#define PMI2_MAX_ATTRVALUE          1024
+#define PMI2_MAX_VALLEN             2048
+#define PMI2_MAX_ATTRVALUE          2048
 #define PMI2_ID_NULL                -1
 
 
--- a/src/common/libpmi/simple_server.h
+++ b/src/common/libpmi/simple_server.h
@@ -14,7 +14,7 @@
 struct pmi_simple_server;
 
 #define SIMPLE_KVS_KEY_MAX         64
-#define SIMPLE_KVS_VAL_MAX         1024
+#define SIMPLE_KVS_VAL_MAX         2048
 #define SIMPLE_KVS_NAME_MAX        64
 
 #define SIMPLE_MAX_PROTO_OVERHEAD  64


garlick:
I _think_ this is all that's needed in the server.

I suppose it would not be feasible to concoct an MPI test for CI that would try to exceed 1024 bytes in PMI_process_mapping to see what the mpis do with this.  (Maybe a LONGTEST?)

grondo:
> I suppose it would not be feasible to concoct an MPI test for CI that would try to exceed 1024 bytes in PMI_process_mapping to see what the mpis do with this. (Maybe a LONGTEST?)

Hm, that was a good suggestion. While we don't have an MPI test that can be used to check that the cyclic task map was actually detected (yet), I did try running 128 tasks per "node" with the MPI hello test and found that on my ubuntu system it seems to trigger a bug in MPI:
```console
$ flux mini run -N2 --taskmap=cyclic --tasks-per-node=128 ./hello
double free or corruption (top)
double free or corruption (top)
double free or corruption (top)
double free or corruption (top)
double free or corruption (top)
[snip]
flux-job: task(s) exited with exit code 134
```

Though 123 tasks per node works:
```console
$ flux mini run -N2 --taskmap=cyclic --tasks-per-node=123 ./hello
ƒ9LTaVEyD: completed MPI_Init in 4.077s.  There are 246 tasks
ƒ9LTaVEyD: completed first barrier in 1.494s
ƒ9LTaVEyD: completed MPI_Finalize in 0.993s
```

I'll give some other MPI versions a try, this version was `MPICH Version:	3.3a2`

grondo:
Same thing happens with mvapich2-2.3.6. Perhaps this change is a bit futile.

garlick:
Ugh, yeah a cursory look in the mpich code shows that their simple PMI client implementation hardwires 1024 in the pmi  protocol handler (`simple_pmiutil.c`).  It's probably getting a protocol error because the server response is truncated and it no longer gets the terminating newline (expected by the protocol).  And it must not handle that well.

Interesting that in PMI-2 argonne gave up on a server-defined max value length and just #define the value to 1024 in the pmi2.h header.    I'm not sure there's even a way to get it to look at our pmi2.h during the build.  Maybe that value is considered part of the ABI.

Well I guess this PR could at least leave a reference  to this issue in the places where 1024 is hard wired to aid future explorers.

garlick:
Slightly unrelated: should we store `flux.taskmap` as a PMI key and use it (if available) instead of `PMI_process_mapping` in the simple_client.c implementation?

Then FWIW our libpmi.so `PMI_Get_clique_size()` and `PMI_Get_clique_ranks()` could work with a large cyclic mapping.

The broker could also preferentially use it I guess.

grondo:
This seemed to work pretty well, so closing this PR in favor of @garlick's suggestion.

PR #4803, opened 2022-12-07
libhostlist: fix 8K limit on encoded hostlists

grondo:
This PR fixes an artificial 8K limit on the return value from `hostlist_encode()` by replacing an internal statically sized buffer with the use of a POSIX memstream to dynamically allocate memory for the encoded result.

Additionally, some unused code is removed.


chu11 (approved):
lgtm, and the cleanup looks good!

grondo:
Thanks! I'll set MWP.

PR #4806, opened 2022-12-08
cleanup: remove some unused functions

grondo:
I ran code coverage without any unit tests to find libutil and other functions that are used only by their own unit tests.

I thought I'd find a bit more, but it was only a couple functions after all.


chu11 (approved):
LGTM

PR #4808, opened 2022-12-08
flux-job: remove finish_mpir_interface()

### LINKED ISSUE #4797: flux-job: remove finish_mpir_interface()
The actions taken in `flux-job.c` `finish_mpir_interface()` seem to be unnecessary and may cause debuggers grief.
Remove this function and just let the aborting process exit.

grondo:
Problem: The finish_mpir_interface() function is not necessary and may cause problems in some paralllel debuggers (verified with TotalView).

Remove this function. Just let flux-job exit at the end of a job to notify the debugger that the job is complete.

Fixes #4797

### DIFF
--- a/src/cmd/flux-job.c
+++ b/src/cmd/flux-job.c
@@ -2031,14 +2031,6 @@ static void setup_mpir_interface (struct attach_ctx *ctx, json_t *context)
     }
 }
 
-static void finish_mpir_interface ()
-{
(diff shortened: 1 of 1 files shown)


chu11 (approved):
if confirmed causes totalview grief, seems to make sense

grondo:
Thanks, setting MWP.

Issue #4810, opened 2022-12-09
t0029-filemap-cmd: not ok 30 - holes were preserved


chu11:
Saw

```
--- blocks.exp	2022-12-09 01:20:01.149304000 +0000
+++ blocks.out	2022-12-09 01:20:01.152313000 +0000
@@ -1 +1 @@
-0
+16
not ok 30 - holes were preserved
#	
#		stat --format="%b" testfile3 >blocks.exp &&
#		stat --format="%b" copydir/testfile3 >blocks.out &&
#		test_cmp blocks.exp blocks.out
#	
```

the file was created in a previous test

```
	truncate --size=8192 testfile3
```

happened to run this test on an NFS mount and wondering if its simply subtlety in fs behaviors with libarchive.  Ran on `/tmp/` which is a tmpfs and there were no issues.

Not sure of the best way to deal with at the moment.

chu11:
i've confirmed that tmpfs vs nfs return different things during the mapping of the file, in particular the returns from `lseek` w/ SEEK_DATA and SEEK_HOLE

nfs

```
offset = 0, size = 8192
append = 0, blobsize = 8192
ok 28 - map test file

expecting success: 
	flux filemap get -C copydir &&
	test_cmp testfile3 copydir/testfile3

extract blob [[0, 8192, "sha1-0631457264ff7f8d5fb1edc2c0211992a67c73e6"]]
offset = 0u, size = 8192u
ok 29 - test file can be read through content cache
```

vs

tmpfs

```
offset = 0, size = 8192
enxio
ok 28 - map test file

expecting success: 
	flux filemap get -C copydir &&
	test_cmp testfile3 copydir/testfile3

extract blob []
ok 29 - test file can be read through content cache
```

the return of `ENXIO` from `lseek` w/ SEEK_DATA is what I'd expect under the case of a whole file that is a "hole".  But that's not happening on this nfs mount, leading to blobrefs being created and the end result of a file that is 16 blocks instead of 0.  Why the initial `truncate` gives the file `0` blocks is beyond me.  Perhaps some internal trickery with nfs when a file is initially created that `libarchive` doesn't know about??

I'd say if we can add some logic like

```
if orig_blocks == 0 and newfile_blocks != 0
     check that orig_size == newfile_size
```

and simply document that some filesystems do this.

chu11:
oh hey what do ya know

https://serverfault.com/questions/731632/does-nfs-and-smb-support-sparse-files

```
[...]

Issue #4809, opened 2022-12-09
t0029-filemap-cmd: not ok 9 - file permissions are correct


chu11:
Saw this

```
--- access.exp	2022-12-09 01:18:26.491671000 +0000
+++ access.out	2022-12-09 01:18:26.495667000 +0000
@@ -1 +1 @@
-744
+700
not ok 9 - file permissions are correct
#	
#		stat --format="%a" testfile >access.exp &&
#		stat --format="%a" copydir/testfile >access.out &&
#		test_cmp access.exp access.out
#	
```

After a bit of debugging realized this:

```
before test_under_flux umask
0022
sharness: loading extensions from /g/g0/achu/chaos/git/flux-framework/flux-core/t/sharness.d/01-setup.sh
sharness: loading extensions from /g/g0/achu/chaos/git/flux-framework/flux-core/t/sharness.d/flux-sharness.sh
before test_under_flux umask
0077
after test_under_flux umask
0077
```

So the umask is somehow changing when `test_under_flux` is called.  My suspicion is there may be different configurations between different shells so there's no way to know if the umask when this test was run will end up being the umask used in the test.

Tried setting `umask 022` at the beginning of the test, this test then passes.  So perhaps needs to be manually set for this test since we are explicitly testing for specific permissions and cannot assume a certain umask exists before the test is run.

PR #4812, opened 2022-12-09
testsuite: fix several corner cases in t0029-filemap-cmd

chu11:
proposed solutions for issue #4809 and #4810


chu11:
per some discussion in #4810, fixed in an alternate way.  Note that @garlick is author on last commit here.

garlick (approved):
Tried it out and tests are working for me.  Thanks!

PR #4817, opened 2022-12-09
flux-pmi: add installed PMI test tool

garlick:
Problem: when debugging MPI or other parallel launch problems, it is sometimes useful to be able to run a stripped down  PMI client  to verify that at least that level of functionality is working in the failing environment.

I've found myself building the clunky pmi test programs in libpmi for this purpose, when really I wanted something like `mpi_hello` but at the pmi level.  I can imagine it would be useful to be able to suggest that someone try this without requiring them to have a built source tree.

This adds a built-in command that can be run as a parallel job to verify PMI, e.g.
```
$ flux mini run -n32 flux pmi barrier
ƒ4mC1JWT: completed pmi barrier on 32 tasks in 0.001s.

$ flux mini run flux pmi get flux.taskmap
{"version":1,"map":[[0,5,6,1],[5,1,2,1]]}
```

### DIFF
--- a/src/cmd/Makefile.am
+++ b/src/cmd/Makefile.am
@@ -20,6 +20,7 @@ AM_CPPFLAGS = \
 fluxcmd_ldadd = \
 	$(top_builddir)/src/common/libkvs/libkvs.la \
 	$(top_builddir)/src/common/librlist/librlist.la \
+	$(top_builddir)/src/common/libpmi/libpmi_client.la \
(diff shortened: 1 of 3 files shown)


grondo (approved):
This looks useful to me, great idea!

LGTM! (fixing the so-called "python-linting" error should be trivial)

garlick:
Thanks - fixed that and also added a small bit of missing test coverage.  I'll set MWP.

PR #4818, opened 2022-12-10
shell/stage-in: allow destination scope to be specified

garlick:
Problem: the stage-in plugin cannot specify a global file system as the destination, since the copy always occurs on all shells of the job.

Rename the `stage-in.destdir` option to `stage-in.destination` and allow an optional scope prefix of `global:` (denoting a global file system) or `local:` (denoting a local file system).   If the scope is _global_, the copy is only performed on the first shell.


grondo (approved):
LGTM! Thoughtful improvement.

garlick:
Thanks, I guess I'll set MWP then.

PR #4819, opened 2022-12-10
shell: allow users to map specific cpus to tasks with `map` cpu-affinity option

grondo:
This PR adds a new mode for shell's `cpu-affinity` option which allows users to map specific cpus to tasks using a semicolon separated list. List members can be any of hwloc's supported bitmap formats, including list (like an idset), bitmap (hwloc specific bitmap format) or taskset (taskset hex bitmask format). If less entries exist in the list than tasks on the node, then entries are reused, e.g.:

```console
$ flux mini run -n3 -o cpu-affinity=map:'0-3;0xf,0x0;0xffff000000000' -N1 --exclusive sh -c 'taskset -cp $$'
1: pid 1621419's current affinity list: 32-35
0: pid 1621418's current affinity list: 0-3
2: pid 1621420's current affinity list: 36-51
```

### DIFF
--- a/src/shell/affinity.c
+++ b/src/shell/affinity.c
@@ -32,6 +32,31 @@ struct shell_affinity {
     hwloc_cpuset_t *pertask;
 };
 
+static void cpuset_array_destroy (hwloc_cpuset_t *set, int size)
+{
+    if (set) {
+        for (int i = 0; i < size; i++) {
+            if (set[i] != NULL)
+                hwloc_bitmap_free (set[i]);
+        }
+        free (set);
+    }
+}
+
+static hwloc_cpuset_t *cpuset_array_create (int size)
+{
+    hwloc_cpuset_t *set = calloc (size, sizeof (hwloc_cpuset_t));
+    if (!set)
+        return NULL;
+    for (int i = 0; i < size; i++)
+        if (!(set[i] = hwloc_bitmap_alloc ()))
+            goto error;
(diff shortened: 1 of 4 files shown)


grondo:
Inception tests were failing due to missing use of `MULTICORE` prereq where necessary. Fixed and pushed the result.

src/shell/affinity.c
```diff
@@ -41,6 +66,89 @@ static int topology_restrict (hwloc_topology_t topo, hwloc_cpuset_t set)
...
+        if (strstarts (arg, "0x")) {
+            /*  If string starts with 0x then parse as a bitmask. If a
+             *  comma is present in the string, then this is likely a
+             *  hwloc-style bitmap string, otherwise, try the taskset
+             *  style bitmaps, which are simpler.
+             */
+            if (strchr (arg, ','))
+                rc = hwloc_bitmap_sscanf (cpusets[i], arg);
+            else
+               rc = hwloc_bitmap_taskset_sscanf (cpusets[i], arg);
```
garlick, line 104: indent problem there?

garlick (approved):
Just spotted one minor cosmetic issue, otherwise LGTM!

src/shell/affinity.c
grondo: Nice catch! Fixed, rebased, and pushed. I'll set MWP.

PR #4823, opened 2022-12-10
libflux/message: cleanup with macros and CCAN pushpull class

garlick:
Problem: the Flux message PROTO frame (described in [RFC 3](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_3.html)) codec is not self contained.

I came across the CCAN "pushpull" class and it seemed like a good way to clean up the PROTO codec.  Now it just looks like this:
```c
int proto_encode (const struct proto *proto, void *buf, size_t size)
{
    char *cp = buf;
    size_t len = 0;

    fixed_size = size;
    push_set_realloc (proto_realloc);

    if (!push_u8 (&cp, &len, PROTO_MAGIC)
        || !push_u8 (&cp, &len, PROTO_VERSION)
        || !push_u8 (&cp, &len, proto->type)
        || !push_u8 (&cp, &len, proto->flags)
        || !push_u32 (&cp, &len, proto->userid)
        || !push_u32 (&cp, &len, proto->rolemask)
        || !push_u32 (&cp, &len, proto->aux1)
        || !push_u32 (&cp, &len, proto->aux2)
        || len < size) {
        errno = EINVAL;
        return -1;
    }
    return 0;
}

int proto_decode (struct proto *proto, const void *buf, size_t size)
{
    const char *cp = buf;
    size_t len = size;
    uint8_t magic, version;

    if (!pull_u8 (&cp, &len, &magic)
        || magic != PROTO_MAGIC
        || !pull_u8 (&cp, &len, &version)
        || version != PROTO_VERSION
        || !pull_u8 (&cp, &len, &proto->type)
        || !pull_u8 (&cp, &len, &proto->flags)
        || !pull_u32 (&cp, &len, &proto->userid)
        || !pull_u32 (&cp, &len, &proto->rolemask)
        || !pull_u32 (&cp, &len, &proto->aux1)
        || !pull_u32 (&cp, &len, &proto->aux2)
        || len > 0) {
        errno = EPROTO;
        return -1;
    }
    return 0;
}
```
I also added some private macros to `message_private.h` to clean up the flags and message type checks which are highly repetitive and tedious around the message code.

### DIFF
--- a/src/common/libflux/message_proto.c
+++ b/src/common/libflux/message_proto.c
@@ -32,7 +32,7 @@ static void proto_set_u32 (uint8_t *data, int index, uint32_t val)
 void msg_proto_setup (const flux_msg_t *msg, uint8_t *data, int len)
 {
     assert (len >= PROTO_SIZE);
-    assert (msg->proto.type != FLUX_MSGTYPE_ANY);
+    assert (msg_type_is_valid (msg));
     memset (data, 0, len);
     data[PROTO_OFF_MAGIC] = PROTO_MAGIC;
     data[PROTO_OFF_VERSION] = PROTO_VERSION;
--- a/src/common/libflux/message_proto.c
+++ b/src/common/libflux/message_proto.c
@@ -8,51 +8,72 @@
  * SPDX-License-Identifier: LGPL-3.0
 \************************************************************/
 
+/* message_proto.c - marshal RFC 3 PROTO frame */
+
 #if HAVE_CONFIG_H
 #include "config.h"
 #endif
 #include <errno.h>
-#include <stdbool.h>
-#include <string.h>
-#include <arpa/inet.h>
-#include <assert.h>
-#include <inttypes.h>
 
-#include "message.h"
-#include "message_private.h"
+#include "ccan/pushpull/pushpull.h"
+
 #include "message_proto.h"
 
-static void proto_set_u32 (uint8_t *data, int index, uint32_t val)
-{
-    uint32_t x = htonl (val);
-    int offset = PROTO_OFF_U32_ARRAY + index * 4;
-    memcpy (&data[offset], &x, sizeof (x));
-}
-
-void msg_proto_setup (const flux_msg_t *msg, uint8_t *data, int len)
-{
-    assert (len >= PROTO_SIZE);
-    assert (msg_type_is_valid (msg));
-    memset (data, 0, len);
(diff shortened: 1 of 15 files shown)


src/common/libflux/message_proto.c
```diff
@@ -32,7 +32,7 @@ static void proto_set_u32 (uint8_t *data, int index, uint32_t val)
 void msg_proto_setup (const flux_msg_t *msg, uint8_t *data, int len)
 {
     assert (len >= PROTO_SIZE);
-    assert (msg->proto.type != FLUX_MSGTYPE_ANY);
+    assert (msg_type_is_initialized (msg));
```
grondo: I don't see where `msg_type_is_initialized ()` is defined, should this be `msg_type_is_valid ()`? (That is what was used in `flux_msg_create()` above. Sorry if I missed something obvious

src/common/libflux/message_proto.c
```diff
@@ -8,51 +8,70 @@
...
-
-#include "message.h"
-#include "message_private.h"
+
+#include "ccan/pushpull/pushpull.h"
+
 #include "message_proto.h"
 
-static void proto_set_u32 (uint8_t *data, int index, uint32_t val)
+static size_t fixed_size;
```
grondo, line 22: This static variable could use a comment IMO. I'm still a little confused on how this will work in the broker with multiple threads possibly calling these protocol functions at the same time. I have no doubt it does work, I'm just confused how. A comment to soothe amateurs like myself might go a long way.

grondo (approved):
Nice cleanup! Just a couple questions inline (sorry for the potentially dumb questions)

src/common/libflux/message_proto.c
garlick: Uh oh that was a rebase error - will fix!  (The original name of that function was `msg_type_is_initialized()` but I later changed it to `msg_type_is_valid()`)

src/common/libflux/message_proto.c
garlick: Good point - this code would have to assume the `fixed_size` never changes.  It happens not to but that's not a great assumption given that the function takes a variable size buffer.  Eek bad.  Will fix.

garlick:
Fixed those things - thanks for finding them.

grondo:
LGTM! Feel free to set MWP when you're ready.

PR #4824, opened 2022-12-11
build: fix installation of libpmi*.so

grondo:
Problem: Commit 90d4af5fdd357d180dafa7ed101e8b99896a0962 attempted to fix a race condition during `make install` that could cause relinking of libpmi*.so to fail when used with `make -j`, but the fix wasn't implemented properly and instead the PMI libraries were not installed at all.

The problem with the original fix appears to be that it registers a duplicate rule:

 install-fluxlibLTLIBRARIES: $(install-libLTLIBRARIES)

which prevents the _real_ install rule from appearing in Makefile.in. Instead, this fix uses what appears to be proposed in the suggested workaround here:

  https://debbugs.gnu.org/cgi/bugreport.cgi?bug=7328

which I don't completely understand, but first assigns the rule name to a variable, then uses the variable to make the new rule with a dependency on install-libLTLIBRARIES. This appears to work around the issue of the disappearing real installation rule.

### DIFF
--- a/src/common/Makefile.am
+++ b/src/common/Makefile.am
@@ -84,7 +84,8 @@ fluxlib_LTLIBRARIES = libpmi.la libpmi2.la
 #
 #  ref: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=7328
 #
-install-fluxlibLTLIBRARIES: $(install-libLTLIBRARIES)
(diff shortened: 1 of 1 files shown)


garlick (approved):
LGTM!

grondo:
Thanks for the quick review, setting MWP.

PR #4825, opened 2022-12-11
testsuite: fix failure on a system with fully-qualified hostname

grondo:
Problem: In t2616-job-shell-taskmap.t an expected hostname is compared to `hostname -s` which breaks the test on systems with a fully-qualified hostname.

Drop the `-s` argument to hostname to make the test more reliable.


grondo:
This issue was detected when trying to build RPMs in the TOSS buildfarm.

garlick (approved):
LGTM

PR #4828, opened 2022-12-13
WIP: ci: speed up coverage builds

grondo:
This PR updates the coverage builds with two main improvements:

 - run the main coverage build on fedora35 to get a more modern compiler and ostensibly better coverage results.
 - run only the necessary tests in the "system" coverage build to avoid excessively long build times

There is a coverage drop here, my guess is because of differences between total lines detected or which lines are optimized out in the newer GCC vs what we were using on bionic. The PR is a WIP while I convince myself I haven't dropped any tests or anything like that.

Before this PR the `system` coverage build was often running over the 60 minute time limit. Now the system build should run in < 10 minutes and the regular coverage build is ~30 minutes. This should reduce CI turn around time significantly. (We may just have to live with a one time coverage statistic drop)


grondo:
I removed WIP. I think I've found most of the missing coverage bits, though codecov seems to be a little confused at the moment.

grondo:
Thanks! Setting MWP.

PR #4829, opened 2022-12-14
add internal universal PMI client library

garlick:
Problem: the one-off code in the broker for selecting among several PMI client implementations  is a bit of a mess and not sharable with other parts of Flux such as `flux-pmi`.

Create a new libupmi internal library that contains this logic and convert the broker and `flux-pmi` to use it.

This is tagged a WIP until unit tests can be written.


src/common/libpmi/upmi_libpmi.c
```diff
@@ -87,8 +87,6 @@ static void *op_create (struct upmi *upmi,
         return NULL;
     }
     ctx->upmi = upmi;
-    if (!path)
-        path = getenv ("PMI_LIBRARY");
```
grondo: Commit message: "allows takes" :arrow_right: "takes the" or "allows a"?

grondo (approved):
LGTM!

I also tested launching an instance on a TOSS 3 cluster running Slurm and no issues :+1:

garlick:
Thanks!  I fixed the commit message and forced a push.  Will set MWP.

PR #4833, opened 2022-12-14
flux-top: add extra test coverage

chu11:
in preparation for queue support in `flux-top` I would need some extra testing ability.  Wanted to add testing coverage under non-queue circumstances first, but hit some snags, thus this PR resulting in some cleanup, fixes to allow easier / more testing, add more tests.

I think the only iffy commit here may be the `test-exit after summary output refreshed` commit.  I admittedly took a very lazy approach to dealing with this.  I felt adding some type of "completion_ref" count system like done in the shell would have been overkill.

### DIFF
--- a/src/cmd/top/joblist_pane.c
+++ b/src/cmd/top/joblist_pane.c
@@ -159,7 +159,7 @@ void joblist_pane_draw (struct joblist_pane *joblist)
             wattron (joblist->win, A_REVERSE);
         if (uri != NULL)
             wattron (joblist->win, COLOR_PAIR(TOP_COLOR_BLUE) | A_BOLD);
-        if (joblist->show_queue)
+        if (joblist->show_queue) {
             mvwprintw (joblist->win,
                        1 + index,
                        0,
@@ -174,7 +174,19 @@ void joblist_pane_draw (struct joblist_pane *joblist)
                        name_width,
                        name_width,
                        name);
-        else
+            if (joblist->top->testf)
+                fprintf (joblist->top->testf,
+                         "%s %s %s %s %d %d %s %s\n",
+                         idstr,
+                         queue,
+                         username,
+                         flux_job_statetostr (state, "S"),
+                         ntasks,
+                         nnodes,
+                         run,
+                         name);
+        }
+        else {
             mvwprintw (joblist->win,
                        1 + index,
                        0,
@@ -188,6 +200,17 @@ void joblist_pane_draw (struct joblist_pane *joblist)
                        name_width,
                        name_width,
                        name);
+            if (joblist->top->testf)
+                fprintf (joblist->top->testf,
+                         "%s %s %s %d %d %s %s\n",
+                         idstr,
+                         username,
+                         flux_job_statetostr (state, "S"),
+                         ntasks,
+                         nnodes,
+                         run,
+                         name);
+        }
         wattroff (joblist->win, A_REVERSE);
         wattroff (joblist->win, COLOR_PAIR(TOP_COLOR_BLUE) | A_BOLD);
     }
--- a/src/cmd/top/joblist_pane.c
+++ b/src/cmd/top/joblist_pane.c
@@ -210,7 +210,7 @@ static void joblist_continuation (flux_future_t *f, void *arg)
     if (joblist->top->test_exit) {
         /* Ensure joblist window is refreshed before exiting */
         wrefresh (joblist->win);
-        flux_reactor_stop (flux_future_get_reactor (f));
+        test_exit_check (joblist->top);
     }
     flux_future_destroy (f);
 }
--- a/t/t2801-top-cmd.t
+++ b/t/t2801-top-cmd.t
@@ -78,8 +78,9 @@ test_expect_success 'flux-top fails if TERM is not supported' '
 test_expect_success 'submit batch script and wait for it to start' '
 	cat >batch.sh <<-EOT &&
 	#!/bin/sh
+	flux mini submit --wait-event=start sleep 300
 	touch job2-has-started
-	flux mini run sleep 300
+	flux queue drain
 	EOT
 	chmod +x batch.sh &&
 	flux mini batch -t30m -n1 batch.sh >jobid2 &&
--- a/t/t2801-top-cmd.t
+++ b/t/t2801-top-cmd.t
@@ -33,8 +33,33 @@ test_expect_success 'flux-top fails if job argument is unknown' '
 	test_must_fail $runpty -n --stderr=unkjobid.err flux top 12345 &&
 	grep "jobid 12345 not found" unkjobid.err
 '
+test_expect_success 'flux-top --test-exit works with a pty' '
+	$runpty flux top --test-exit >/dev/null
+'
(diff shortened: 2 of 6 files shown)


chu11:
re-pushed a fix to fix a chain-&& issue.  Two builders failed .. i'm worried that some of the lines are wrapped funny and my greps aren't working.  Ugh ... not sure how to do with that otherwise.  maybe need to ncurses `fflush()` equivalent or something.

chu11:
re-pushed with a somewhat ugly way to get around the multi-line asciicast output.

chu11:
ugh ... ok i think this has to be done a different way.  having to deal with all the weird parsing possibilities is probably a bad idea.  I'm now thinking some simple `--test-exit-output=file` is probably best.  Just have some dummy file with the data that will ultimately be displayed is good enough.

chu11:
a completely different idea for testing, I added a new function "--test-exit-dump", which dumps basically the same information from `flux-top`, but into a test file that doesn't include any escape codes, etc.  And I test against that.  I dunno, maybe it's not a great idea?  At some point I think writing something to handle all the corner cases parsing runpty's `asciicast` output isn't worth it.  This is less error prone and makes the testing in `t2801-top-cmd.t` easier to understand.

grondo:
I like the idea of a dump feature, but you have to ask what it is actually testing. `flux top` is using standard RPCs to gather its data and those RPCs and their results are already tested elsewhere, so mainly what we want to test is that these values are displayed properly and as expected. If we only test a hidden test-only option, then we are limited in what we've actually tested.

I also agree there is diminishing returns to testing all the corner cases with the raw terminal output. I'm not sure the right solution there, but also keep in mind we don't want to make the tests so onerous that it is painful for people to make changes to flux-top in the future. I'm not saying the extra testing here is not valuable, but just to keep in mind that we should have a balanced approach.

grondo:
A related idea: [top](https://man7.org/linux/man-pages/man1/top.1.html) has a "batch mode" `-b` which dumps screens of output with (I think) no escape codes to stdout. This is actually useful because the output of `top` can be sent to a file or other programs. I wonder how difficult it would be to refactor the `--test-exit-dump` option into a batch option?

I just thought I'd throw that idea out there, it is definitely not high priority since we have a lot of other more important work to do.

chu11:
> A related idea: top has a "batch mode" -b which dumps screens of output with (I think) no escape codes to stdout. This is actually useful because the output of top can be sent to a file or other programs. I wonder how difficult it would be to refactor the --test-exit-dump option into a batch option?

(oops missed this response yesterday)

Hmmm that's a good idea.  Although I have no idea how `top` does its batch mode.  Not sure if there's a trick inside curses or something.  But worthwhile to atleast investigate a bit.

t/t2801-top-cmd.t
```diff
@@ -128,6 +128,12 @@ test_expect_success NO_CHAIN_LINT 'flux-top does not exit on recursive failure'
 	        --input=recurse-fail.in flux top &&
 	grep -qi "error connecting to Flux" recurse-fail.log
 '
+test_expect_success 'cleanup running jobs' '
+	flux job cancel $(cat jobid2) &&
+	flux job wait-event $(cat jobid2) clean &&
+	flux job cancel $(cat jobid3) &&
+	flux job wait-event $(cat jobid3) clean
```
grondo, line 191: FYI - cancel can now take multiple jobids.

src/cmd/top/joblist_pane.c
```diff
@@ -174,7 +174,19 @@ void joblist_pane_draw (struct joblist_pane *joblist)
                        name_width,
                        name_width,
                        name);
-        else
+            if (joblist->top->test_exit && joblist->top->testf)
```
grondo, line 177: Would `top->testf` ever be set without `top->test_exit`? (i.e. are both checks really necessary here?)

grondo (approved):
Looks good, nice testing improvements!

Just some minor comments inline. Not critical to address them IMO.

src/cmd/top/joblist_pane.c
chu11: ahhh i did tweak something along the way, so you're right, both checks are no longer necessary.

chu11:
> Hmmm that's a good idea. Although I have no idea how top does its batch mode. Not sure if there's a trick inside curses or something. But worthwhile to atleast investigate a bit.

Well, I don't grok the code in `top` perfectly after a skim.  But it appears that they do a mixture of windowing with curses and just write out stuff to STDOUT without curses.  So I think things are refactored in there far more easily than in our flux-top.

We can perhaps keep it in mind for the future.

chu11:
re-pushed with some cleanup per comments above.  also one extra commit, i realized that was a racy test in t2801-top-cmd.

t/t2801-top-cmd.t
```diff
@@ -79,11 +79,15 @@ test_expect_success 'submit batch script and wait for it to start' '
 	cat >batch.sh <<-EOT &&
 	#!/bin/sh
 	touch job2-has-started
-	flux mini run sleep 300
+	jobid=\$(flux mini submit --flags waitable sleep 300)
+	flux job wait-event \$jobid start
```
grondo, line 83: The separate `submit` and `wait-event` could be replaced with `flux mini submit --wait-event=start --flags waitable sleep 300`

t/t2801-top-cmd.t
```diff
@@ -79,11 +79,15 @@ test_expect_success 'submit batch script and wait for it to start' '
...
+	touch job2-batch-sleep-has-started
+	flux job wait --all
 	EOT
 	chmod +x batch.sh &&
 	flux mini batch -t30m -n1 batch.sh >jobid2 &&
-	$waitfile job2-has-started
+	$waitfile job2-has-started &&
+	$waitfile job2-batch-sleep-has-started
```
grondo, line 90: It would probably be simpler to just wait for `job2-batch-sleep-has-started` (or just move `job2-has-started` to after `flux job wait-event` above), since appearance of `job2-batch-sleep-has-started` implies `job2-has-started`...

grondo (approved):
Still seems good to me. Just a couple comments regarding cleanup of `testsuite: fix job race`

t/t2801-top-cmd.t
```diff
@@ -50,28 +75,63 @@ test_expect_success 'flux-top fails if TERM is not supported' '
...
-'
 test_expect_success 'submit batch script and wait for it to start' '
 	cat >batch.sh <<-EOT &&
 	#!/bin/sh
 	touch job2-has-started
-	flux mini run sleep 300
+	jobid=\$(flux mini submit --flags waitable sleep 300)
+	flux job wait-event \$jobid start
+	touch job2-batch-sleep-has-started
+	flux job wait --all
```
grondo, line 85: Oh, also, a more idiomatic way to wait for all jobs to complete at the end of a batch script would be
```
flux queue drain
```
Then you don't need to add the `waitable` flag to all jobs.

chu11: oh that's a good idea

chu11:
tweaked the last commit per your comments, I'll set MWP

PR #4838, opened 2022-12-15
libtaskmap: add TASKMAP_ENCODE_RAW_DERANGED

garlick:
Problem: flux-pmix needs the taskmap in raw form, without the idset ranges.
    
Add a flag to trivially generate the raw taskmaps without ranges.   

This will allow some [code](https://github.com/flux-framework/flux-pmix/blob/main/src/shell/plugins/maps.c#L51) over there to just fetch the taskmap directly and encode it without the extra machinations.


grondo (approved):
LGTM!

garlick:
Thanks!  Setting MWP.

PR #4841, opened 2022-12-16
flux-shell: add -opmi=off option

garlick:
Problem: there is no way to disable the flux-shell's PMI-1 server, but this may be useful when testing other process manager interfaces like pmix.

Add `-opmi=off` and improve the robustness of the pmi shell plugin option parsing.


grondo (approved):
Good improvement! LGTM.

PR #4845, opened 2022-12-16
add tools for querying job remaining time: `flux_job_timeleft(3)`, python `flux.job.timeleft()` and `flux-job timeleft`

grondo:
This WIP PR proposes a new `flux_job_timeleft(3)` function which fetches the remaining time for the current process, abstracting the implementation from callers because (as we've seen) the determination of _how_ to get the current time left is not exactly straightforward in all situations.

This function doesn't take a jobid but is still placed in libjob with the `flux_job` prefix since only jobs have a duration/expiration, so it doesn't seem to really belong under libflux.

Opening early as a WIP to get feedback (nothing about the interface is set in stone) and to allow @garlick to attempt addition of a `flux timeleft` utility, which can also act as a test vehicle for the function.

### DIFF
--- a/doc/man3/flux_job_timeleft.rst
+++ b/doc/man3/flux_job_timeleft.rst
@@ -1,0 +1,43 @@
+====================
+flux_job_timeleft(3)
+====================
+
+
+SYNOPSIS
+========
+
+::
+
+   #include <flux/core.h>
+
+::
+
+   int flux_job_timeleft (flux_t *h,
+                          flux_error_t *error,
+                          double *timeleft);
+
+DESCRIPTION
+===========
+
+The ``flux_job_timeleft()`` function determines if the calling process
+is executing within the context of a Flux job (either a parallel job or
+a Flux instance running as a job), then handles querying the appropriate
+service for the remaining time in the job.
+
+RETURN VALUE
+============
+
+``flux_job_timeleft()`` returns 0 on success with the remaining time in
+floating point seconds stored in ``timeleft``. If the job does not have
+an established time limit, then ``timeleft`` is set to ``inf``. If the job
+time limit has expired or the job is no longer running, then ``timeleft``
+is set to ``0``.
+
+If the current process is not part of an active job or instance, or another
+error occurs, then this function returns ``-1`` with an error string set in
+``error->text``.
+
+RESOURCES
+=========
+
+Flux: http://flux-framework.org
--- a/src/cmd/flux-job.c
+++ b/src/cmd/flux-job.c
@@ -21,6 +21,7 @@
 #include <time.h>
 #include <stdio.h>
 #include <string.h>
+#include <math.h>
 #include <ctype.h>
 #include <pwd.h>
 #include <assert.h>
@@ -43,6 +44,7 @@
 #include "src/common/libjob/unwrap.h"
 #include "src/common/libutil/read_all.h"
 #include "src/common/libutil/monotime.h"
+#include "src/common/libutil/fsd.h"
 #include "src/common/libidset/idset.h"
 #include "src/common/libeventlog/eventlog.h"
 #include "src/common/libioencode/ioencode.h"
@@ -99,6 +101,7 @@ int cmd_wait (optparse_t *p, int argc, char **argv);
 int cmd_memo (optparse_t *p, int argc, char **argv);
 int cmd_purge (optparse_t *p, int argc, char **argv);
 int cmd_taskmap (optparse_t *p, int argc, char **argv);
+int cmd_timeleft (optparse_t *p, int argc, char **argv);
 
 int stdin_flags;
 
@@ -398,6 +401,13 @@ static struct optparse_option taskmap_opts[] = {
     OPTPARSE_TABLE_END
 };
 
+static struct optparse_option timeleft_opts[] = {
+    { .name = "human", .key = 'H', .has_arg = 0,
+      .usage = "Output in Flux Standard Duration instead of seconds.",
+    },
+    OPTPARSE_TABLE_END
+};
+
 static struct optparse_subcommand subcommands[] = {
     { "list",
       "[OPTIONS]",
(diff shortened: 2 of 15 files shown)


grondo:
Removed WIP for an initial review. This PR now includes a python interface in `flux.job.timeleft()`. Possibly the last bit to do here is add a manpage for `flux_job_timeleft(3)`

grondo:
Hold off. I found an issue with querying timeleft for pending jobs. I'll have to come back to that later today.

grondo:
Ok, fixed the issue for pending jobs and added a brief man page. Ready for a review probably.

garlick (approved):
Nice work, couldn't spot anything wrong.

grondo:
We should get approval from @ofaaland before merging probably. I'll request a review.

src/cmd/flux-job.c
```diff
@@ -3536,6 +3546,34 @@ int cmd_taskmap (optparse_t *p, int argc, char **argv)
...
+
+    if (optindex < argc) {
+        if (setenv ("FLUX_JOB_ID", argv[optindex++], 1) < 0)
+            log_err_exit ("setenv");
+    }
+    if (flux_job_timeleft (h, &error, &t) < 0)
+        log_msg_exit ("%s", error.text);
+    if (fsd_format_duration (buf, sizeof (buf), t) < 0)
+        log_err_exit ("fsd_format_duration");
+    printf ("%s\n", buf);
```
chu11, line 3572: maybe can be for a follow up PR, but perhaps option to output in plain ole seconds w/o the fsd suffix?  I can imagine users wanting to `sec=$(flux job timeleft)` and then immediately use the seconds in something else time related ... or possibly just check for `-eq 0`.  i.e. not waste time converting from fsd to a units they care about.

t/t2201-job-cmd.t
```diff
@@ -767,6 +767,13 @@ test_expect_success 'flux job: timeleft returns 0 for completed job' '
 	flux job timeleft $id > timeleft7 &&
 	grep "0s" timeleft7
 '
+test_expect_success 'flux job: timeleft fails for pending job' '
+	flux queue stop &&
+	id=$(flux mini submit -t 10m true) &&
+	test_expect_code 1 flux job timeleft $id > timeleft8 2>&1 &&
+	flux queue start &&
+	grep "has not started" timeleft8
+'
```
chu11, line 776: this test is in the ` testsuite: add tests for flux.job.timeleft() ` commit.  accidentally put here and not in the other commit with all of the other tests for `t2201-job-cmd`?

src/cmd/flux-job.c
grondo: Yeah, @garlick and I talked about that, but there didn't seem to be a known use case at this point and adding more development and test effort at this point seemed like a poor choice.

t/t2201-job-cmd.t
grondo: Ooh, good catch! Thanks, will fix

doc/man3/flux_job_timeleft.rst
```diff
@@ -0,0 +1,42 @@
...
+
+::
+
+   #include <flux/core.h>
+
+::
+
+   int flux_job_timeleft (flux_t *h,
+                          flux_error_t *error,
+                          double *timeleft);
```
ofaaland, line 17: I see a comment below that says *timeleft is in units of seconds, but I don't see that specified in this man page.

src/cmd/flux-job.c
ofaaland: I don't know if there's an existing fsd <-> seconds convenience function.  Absent that, I personally would prefer seconds rather than fsd, because I can so easily convert seconds to minutes/hours/days but the reverse is not so easy.

That said, for SCR I'll be using python and using flux.job.timeleft(), so I have no concrete use case either.

ofaaland (commented):
Great!  Thanks for taking care of this so quickly.  Looks good to me.  A couple of nits below.

src/cmd/flux-job.c
grondo: Ok, since there is two votes for seconds, I"ll go ahead and implement that, with a `-H, --human` option to return FSD.

What do people think should be returned for the `infinity` case: `-1` or `UINT_MAX`?

doc/man3/flux_job_timeleft.rst
grondo: Great catch! Thanks, will fix.

src/cmd/flux-job.c
garlick: I'd vote for UINT_MAX.  Then a batch script could do something like

```
thing_runtime=3600
while test $(flux job timeleft) -gt $thing_runtime; do
    flux mini run -t $thing_runtime $thing
done
```

grondo:
Thanks for the review @ofaaland! That was very helpful. I've made the following changes here:

 - `flux job timleft` now reports integral seconds left in the current or specified job's time limit. No time limit is reported with `ULONG_MAX`.
 - A `-H, --human` option was added to `flux job timeleft`
 - An entry for `timeleft` was added to `flux-job(1)` for completeness
 - The `flux_job_timeleft(3)` man page was updated to specify that the returned time left is in floating point seconds.

I squashed all the new stuff into the existing commits, hope that was ok.

garlick:
All good here!

grondo:
Ok, @ofaaland want to give an ACK? If so, then I'll set MWP.

grondo:
Well, I think since @ofaaland already commented we can merge this one and get it out there. Setting MWP.

ofaaland:
> Well, I think since @ofaaland already commented we can merge this one and get it out there. Setting MWP.

Shoot, sorry.  I was about to approve and then got distracted.  Thank you!

grondo:
No problem! Thanks @ofaaland!

PR #4846, opened 2022-12-16
libutil: support "infinity" in FSD

grondo:
This PR updates the C and Python interfaces in flux-core to support RFC 23 "infinity" special case.

### DIFF
--- a/src/bindings/python/flux/util.py
+++ b/src/bindings/python/flux/util.py
@@ -226,6 +226,10 @@ class CLIMain(object):
 
 
 def parse_fsd(fsd_string):
+    # Special case for RFC 23 "infinity"
+    if fsd_string in ["inf", "infinity", "INF", "INFINITY"]:
+        return float("inf")
+
     match = re.match(r"(.*?)(s|m|h|d|ms)$", fsd_string)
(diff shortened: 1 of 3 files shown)


chu11 (dismissed):
LGTM (assuming RFC23 update is approved, which I think we agree it will be)

grondo:
Some changes here based on offline feedback from @garlick. It turns out `strtod()` parses case-insensitive `inf` and `infinity` and stopping at the right place. So we do not need to special case `infinity` on the parsing end, all while supporting whatever representations of infinity are supported by `strtod(3)`. I think @garlick will update the RFC 23 changes to stipulate that infinite values SHALL NOT support units, so checking for trailing characters (along with allowing `FP_INFINITE` is the only change needed in `fds_parse_duration(3)`.

On the encoding side, an infinite value is always represented as `infinity`. Let me know if you'd rather `inf`.

grondo:
Updated the fsd unit test to add test vectors proposed in the RFC 23 update.

garlick (approved):
All looks good to me!

grondo:
Thanks! I've set MWP

Issue #4606, opened 2022-09-23
flux-top: add queue-specific top


grondo:
Add a `--queue=NAME` option and/or dynamic queue filtering to `flux-top` so that specific queues can be watched.

This would include:

 * filtering jobs to a specific queue (or queues for extra credit)
 * displaying only those resources matching the queue constraint in the resource summary pane
 * displaying job stats specific to the chosen queues

PR #4847, opened 2022-12-20
flux-top: limit jobs and summary to specific queue

chu11:
per #4606.  I'm really glad I added that testing in #4833, caught a couple of issues :P

- i assume `--queue` doesn't recursively goto into sub-instances.  Possibly sub-instances may have same queue names setup, but it's not what i think is desired by the `--queue` option.

- right now `flux-top` determines if the "QUEUE" column should be displayed if any jobs have been submitted to any queue.  Maybe this should be changed to just determining if any queues have been configured, i.e. if `queue.*` actually is configured?  While what is implemented is perhaps more "correct", its heavier weight and not sure its really needed except outside of testing scenarios (i.e. queue configuration changed between job submissions).  But for the time being I left it in.

- will probably have to add support in `flux-sched` for the `resource-status` RPC.

- i somewhat lazily chose not to add "queues" (plural) support.  Didn't think there was a big use case at the moment.
  - big issue, `job-list` only filters jobs by single queue, so would have to update job-list module or do some type of "build up display of multiple responses". Edit: OR get all jobs and just filter them in `flux-top`, which dunno if I'd want to do that.

  - getting resources of multiple queues should be easy, just add to the constraints list

  - tiny annoying to support in job-stats, b/c gotta add the queue stats together

  - I'm thinking I'll put into an issue as a longer-term todo.

### DIFF
--- a/src/cmd/top/joblist_pane.c
+++ b/src/cmd/top/joblist_pane.c
@@ -18,6 +18,7 @@
 #include <jansson.h>
 
 #include "src/common/libutil/fsd.h"
+#include "ccan/str/str.h"
 
 #include "top.h"
 
@@ -146,6 +147,10 @@ void joblist_pane_draw (struct joblist_pane *joblist)
                            "user",
                              "uri", &uri) < 0)
             fatal (0, "error decoding a job record from job-list RPC");
+
+        if (joblist->top->queue && !streq (joblist->top->queue, queue))
+            continue;
+
         if (flux_job_id_encode (id, "f58", idstr, sizeof (idstr)) < 0)
             fatal (errno, "error encoding jobid as F58");
         (void)fsd_format_duration_ex (run, sizeof (run), fabs (now - t_run), 2);
@@ -311,7 +316,7 @@ void joblist_pane_enter (struct joblist_pane *joblist)
     /*  Lazily attempt to run top on jobid, but for now simply return to the
      *   original top window on failure.
      */
-    if ((top = top_create (uri, title, &error)))
+    if ((top = top_create (uri, title, NULL, &error)))
         top_run (top, 0);
     else
         error_popup (joblist, error.text);
--- a/src/cmd/top/summary_pane.c
+++ b/src/cmd/top/summary_pane.c
@@ -422,24 +422,29 @@ static void resource_continuation (flux_future_t *f, void *arg)
 static void stats_continuation (flux_future_t *f, void *arg)
 {
     struct summary_pane *sum = arg;
+    json_t *o = NULL;
 
-    if (flux_rpc_get_unpack (f,
-                             "{s:i s:i s:i s:i s:{s:i s:i s:i s:i s:i s:i s:i}}",
-                             "successful", &sum->stats.successful,
-                             "failed", &sum->stats.failed,
-                             "canceled", &sum->stats.canceled,
-                             "timeout", &sum->stats.timeout,
-                             "job_states",
-                               "depend", &sum->stats.depend,
-                               "priority", &sum->stats.priority,
-                               "sched", &sum->stats.sched,
-                               "run", &sum->stats.run,
-                               "cleanup", &sum->stats.cleanup,
-                               "inactive", &sum->stats.inactive,
-                               "total", &sum->stats.total) < 0) {
+    if (flux_rpc_get_unpack (f, "o", &o) < 0) {
         if (errno != ENOSYS)
-            fatal (errno, "error decoding job-list.job-stats RPC response");
+            fatal (errno, "error getting job-list.job-stats RPC response");
     }
+
+    if (json_unpack (o,
+                     "{s:i s:i s:i s:i s:{s:i s:i s:i s:i s:i s:i s:i}}",
+                     "successful", &sum->stats.successful,
+                     "failed", &sum->stats.failed,
+                     "canceled", &sum->stats.canceled,
+                     "timeout", &sum->stats.timeout,
+                     "job_states",
+                       "depend", &sum->stats.depend,
+                       "priority", &sum->stats.priority,
+                       "sched", &sum->stats.sched,
+                       "run", &sum->stats.run,
+                       "cleanup", &sum->stats.cleanup,
+                       "inactive", &sum->stats.inactive,
+                       "total", &sum->stats.total) < 0)
+        fatal (0, "error decoding job-list.job-stats object");
+
     flux_future_destroy (f);
     sum->f_stats = NULL;
     draw_stats (sum);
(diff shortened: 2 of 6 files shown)


grondo:
Haven't taken a look yet, but I did want to comment on this:

> will probably have to add support in flux-sched for the resource-status RPC.

Hm, this might be non-trivial and also means this PR can't be merged until that support has landed in flux-sched.

Did you considering doing the filtering in flux-top instead? This could greatly simplify the implementation, and would work without changing the scheduler status RPC.

grondo:
> i somewhat lazily chose not to add "queues" (plural) support. Didn't think there was a big use case at the moment.

Yeah, I don't think there is a use case for multiple queues besides the obvious one of displaying all queues.

> OR get all jobs and just filter them in flux-top, which dunno if I'd want to do that.

I think it would actually be very advantageous to filtering jobs in `flux top` directly. We may want to add support for a `Q` key (or I was thinking left/right  arrow keys) to "page" between queue displays, so that a user doesn't have to quit and restart `top` to view different queues. It would be much easier to do this if the set of displayed jobs is filtered rather than mucking about with sending a new RPC...

chu11:
> Hm, this might be non-trivial and also means this PR can't be merged until that support has landed in flux-sched.

> Did you considering doing the filtering in flux-top instead? This could greatly simplify the implementation, and would work without changing the scheduler status RPC.

i hadn't considered it, but after skimming `flux-sched` this morning ... this does seem like a good idea.

chu11:
> I think it would actually be very advantageous to filter jobs in flux top directly. We may want to add support for a Q key (or I was thinking left/right arrow keys) to "page" between queue displays, so that a user doesn't have to quit and restart top to view different queues. It would be much easier to do this if the set of displayed jobs is filtered rather than mucking about with sending a new RPC...

that's a good idea, and it would actually allow us to "flip" through queues on sub-instances that are selected.  let me look into doing it this way.

chu11:
re-pushed, implementing the "filtering" within `flux-top` instead of in `sched-simple` or `job-list`.  The code ended up simpler in some cases.

I don't think I'll do the 'Q' (jump between queues) support in this PR, punt that to a follow up PR.

grondo:
We could probably punt the runtime queue selection for awhile. I was thinking to allocate left-arrow/right-arrow (maybe `h`/`l`) to flip through queues, but not critical at all.

src/cmd/top/joblist_pane.c
```diff
@@ -146,6 +146,11 @@ void joblist_pane_draw (struct joblist_pane *joblist)
                            "user",
                              "uri", &uri) < 0)
             fatal (0, "error decoding a job record from job-list RPC");
+
+        if (joblist->top->queue
+            && strcmp (joblist->top->queue, queue) != 0)
```
garlick, line 151: Might want to conver to `streq()` here (top.c already uses it).

src/cmd/top/summary_pane.c
```diff
@@ -373,12 +376,24 @@ static int resource_count (json_t *o,
...
-    if (!(rl = rlist_from_json (R, NULL)))
+    if (!(rl_all = rlist_from_json (R, NULL)))
         return -1;
+    if (sum->top->queue_constraint) {
+        flux_error_t error;
+        rl_constraint = rlist_copy_constraint (rl_all,
+                                               sum->top->queue_constraint,
+                                               &error);
+        if (!rl_constraint)
+            fatal (errno, "failed to create constrained rlist");
```
garlick, line 387: Consider including `error.text` as part of the failure message?

src/cmd/top/summary_pane.c
```diff
@@ -358,13 +358,16 @@ static int get_instance_attr_int (flux_t *h, const char *attr)
 }
 
 
-static int resource_count (json_t *o,
+static int resource_count (struct summary_pane *sum,
+                           json_t *o,
                            const char *name,
                            int *nnodes,
                            int *ncores,
                            int *ngpus)
```
garlick, line 366: Minor, but since this function is counting resources returned from `sched.resource-status` maybe the function declaration would be more clear with the response payload first (as before), then add a new `json_t *constraint` argument rather than having the caller pass in all of `struct summary_pane` when it's only used to find the constraint?

src/cmd/top/summary_pane.c
```diff
@@ -429,6 +447,31 @@ static void stats_continuation (flux_future_t *f, void *arg)
...
+            fatal (EPROTO, "job-list.job-stats queue not an array");
+        json_array_foreach (queues, index, value) {
+            const char *name;
+            if (json_unpack (value, "{s:s}", "name", &name) < 0)
+                fatal (EPROTO, "job-list.job-stats queue name not found");
+            if (!strcmp (name, sum->top->queue)) {
+                q = value;
+                break;
+            }
+        }
```
garlick, line 467: Suggestion: move his  block to its own function, something like
```c
json_t *parse_queue_stats (json_t *o, const char *name)
```
and have it return NULL on failure, then handle the failure in one place as a protocol error rather than having three separate error message for what is basically the same unlikely problem.

src/cmd/top/top.c
```diff
@@ -237,8 +238,49 @@ static char * build_title (struct top *top, const char *title)
...
+{
+    flux_future_t *f;
+    json_t *o;
+    json_t *tmp;
+    json_t *requires = NULL;
+
+    if (!(f = flux_rpc (top->h, "config.get", NULL, FLUX_NODEID_ANY, 0))
+        || flux_rpc_get_unpack (f, "o", &o) < 0)
+        fatal (errno, "Error fetching flux config");
+
```
garlick, line 251: Suggestion: fetch the config object and store it within `struct top` before calling `setup_constraint()` ?   Seems like it might come in handy later e.g. when multiple queues/constraints are supported.

src/cmd/top/top.c
```diff
@@ -237,8 +238,49 @@ static char * build_title (struct top *top, const char *title)
...
+        if (!json_is_array (requires))
+            fatal (0, "queue requires configuration invalid");
+        if (!(array = json_array ()))
+            fatal (0, "Error creating json array");
+        json_array_foreach (requires, index, value) {
+            if (json_array_append (array, value) < 0)
+                fatal (0, "Error appending to json array");
+        }
+        if (!(top->queue_constraint = json_pack ("{s:o}", "properties", array)))
+            fatal (0, "Error creating queue constraints");
```
garlick, line 276: Suggestion: instead of copying `requires`, just take a reference on it e.g.
```c
constraint = json_pack ("{s:O}", "properties", requires);
```

garlick (commented):
Just tested on my system with two queues and seems to work well!
A few comments inline.

chu11:
rebased and re-pushed with fixups

garlick:
I've lost my context on this one.  Want to squash the fixups and I'll give it another pass?

chu11:
squashed and repushed

garlick (approved):
This looks good.  Thanks!

PR #4849, opened 2022-12-22
Experimental WIP: add flux-pgrep(1) and flux-pkill(1)

grondo:
I started prototyping a `flux pgrep` utility, but won't have time to wrap this up before I leave for a few weeks, so I'm posting something experimental to get people thinking about it, discuss the interface, etc. 

The code is in _very_ prototype form, but the main experiment here is to use the `Constraint` query syntax proposed in #4677 (along with using a similar internal representation as in the RFC 31 Job Constraints Specification) to allow powerful and flexible job matching instead of just resource matching.

The `flux-pgrep` command here takes some args similar to `flux-jobs`, with non-option arguments forming a pgrep "expression".
```
usage: flux-pgrep [-h] [-a] [-A] [-u [USERNAME|UID]] [-f STATE|RESULT]
                  [--queue QUEUE] [-c N] [--max-entries N] [-n] [-o FORMAT]
                  [-k] [--wait]
                  EXPRESSION [EXPRESSION ...]

positional arguments:
  EXPRESSION                 job name pattern or match expression

optional arguments:
  -h, --help                 show this help message and exit
  -a                         Target jobs in all states
  -A                         Target all users
  -u, --user=[USERNAME|UID]  Limit output to specific username or userid
                             (Specify "all" for all users)
  -f, --filter=STATE|RESULT  List jobs with specific job state or result
      --queue=QUEUE          Limit output to specific queue
  -c, --count=N              Limit output to the first N matches
      --max-entries=N        Limit number of searched jobs to N entries
                             (default: 1000)
  -n, --suppress-header      Suppress printing of header line
  -o, --format=FORMAT        Specify output format using Python's string
                             format syntax or a defined format by name (use
                             'help' to get a list of names)
  -k, --kill                 Cancel instead of printing matching jobs
      --wait                 Wait for jobs to finish after cancel
```

A `flux-pgrep` expression is a query similar in form to gmail or github issues search: a set of terms which may have the form `operator:value` and may be joined by `and`, `or`, or `not`. By default, terms are joined with `and`, so `foo:1 bar:2` is `foo:1 and bar:2`. If an operator is not specified, then `name` is implied, which matches job names by regex (like `pgrep(1)`).

Therefore, `flux pgrep ^flux` would match all your currently active jobs that start with `flux`.

Currently supported operators include `jobid:`, `status:`, `ranks:`, `host|hostlist:`, `runtime`, `start`, `end`, `before`, `since`, `is`, and `nnodes`. Some of these could be dropped or names amended (e.g. `before` and `since` operate on submit time, it might be clearer to use `submitted`)

Operators that take numeric values additionally support conditional and range syntax. For example, `start:>8am` queries for jobs  with start time after 8am, and `start:8am..9am` queries for jobs that started between 8am and 9am inclusive. If you leave off either end of the `..` it works as expected, e.g. `start:8am..` matches jobs that started after 8am and `start:..8am` matches jobs that started before 8am. (`start:8am` specifies exactly 8am).

Because the `<` and `>` require shell quoting, I was thinking of letting `+` and `-` be an alternative to these characters, and that is what is implemented now.

There are also a couple instances of syntactic sugar to make more common queries easier. The form `ID..ID` can be used as an alternate for `jobid:ID1..ID2` and will match all jobids in a range (a requested feature). A term prefixed with an `@` is shorthand for querying when a job was "running", e.g. `@8am host:fluke123` queries what job was running on fluke123 at 8am. The `@` operator can also take a range: `@12am..8am` would list all jobs that were running in that time period.

The `is` operator tries to be smart about allowing matches for job states, results and status. It can take any job state, plus the virtual states `running`, `pending`, `complete`, etc, as well as results `success`, `failure`, `timeout`, `canceled`, etc. This allows you to quickly query the job running on a given host `is:running host:corona123`, or all the jobs that ran in the last day that failed in less than 5 minutes: `is:failure start:>12am runtime:<5m`.

By default, `flux pgrep` only outputs the matching jobids, similar to the `pgrep(1)` utility. However, you can select `--format=full` to get `flux-jobs` like output:
```
 grondo@corona212:~$ flux pgrep -c 5 -a ^flux
ƒVvgZo1MUyd
ƒVvgRoE49s5
ƒVvAbmCR4jZ
ƒVteoaztHCT
ƒVgCMmbgXCK
 grondo@corona212:~$ flux pgrep -c 5 -o full -a ^flux
 ƒVvgZo1MUyd grondo   flux       TO      1      1   30.01m corona191
 ƒVvgRoE49s5 grondo   flux        F      2      2   31.47s corona[189-190]
 ƒVvAbmCR4jZ grondo   flux       CD      1      1   3.079m corona174
 ƒVteoaztHCT grondo   flux       CD      2      2   18.56m corona[173-174]
 ƒVgCMmbgXCK grondo   flux       CD      2      2   11.17m corona[173-174]
grondo@corona212:~$ flux pgrep -c 5 -o full -a ^flux 'runtime:>15m'
 ƒVvgZo1MUyd grondo   flux       TO      1      1   30.01m corona191
 ƒVteoaztHCT grondo   flux       CD      2      2   18.56m corona[173-174]
 ƒUtgSrCvybM grondo   flux       TO      2      2   30.01m corona[173-174]
 ƒUS2n9MNnVD grondo   flux       TO      2      2   30.02m corona[171,173]
 ƒUQpgKmARXm grondo   flux       TO      2      2   30.02m corona[171,173]
 grondo@corona212:~$ flux pgrep -c 5 -o full -a ^flux 'runtime:>15m' is:failed host:corona[160-170]
  ƒQ8HGmJruD grondo   flux        F     48      1   6.313h corona160
```

The `flux pkill` command is just a link to `flux pgrep` and is equivalent to calling `flux pgrep` with the `-k, --kill` option, though what the command actually does now is cancel jobs. It currently reports the number of jobs killed, and includes a `--wait` option to wait until the jobs complete before returning to the command line:

```
f(s=2,d=1,builddir) grondo@corona189:~$ flux mini submit --quiet --cc=1-1000 sleep 100
f(s=2,d=1,builddir) grondo@corona189:~$ flux pkill sleep
flux-pkill: INFO: Canceled 1000 jobs
```

On problem here is that `flux pgrep|pkill` uses the default limit of fetching at most 1000 entries from the `job-list` service. There's a `--max-entries` option to increase that, but this is something to keep in mind since your `pkill` command may not find all possible matches. Maybe eventually we could move `job-list` to use a database and instead of compiling the `flux-pgrep` expression locally, we could compile it a SQL statement or job-list could support the JsonLogic-like RFC 31 constraint object format.

This branch is currently based on #4677 since it uses the `flux.job.Constraint` class. If this PR were ever merged, the current `JobPgrep` class would need a rewrite. If you dare look at the code, remember: prototype.

### DIFF
--- a/src/bindings/python/flux/job/Jobspec.py
+++ b/src/bindings/python/flux/job/Jobspec.py
@@ -15,6 +15,7 @@ import json
 import math
 import numbers
 import os
+import re
 
 import yaml
 from _flux._core import ffi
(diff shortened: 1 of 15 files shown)


grondo:
Closing this PR in favor of a simpler approach in #4867

PR #4850, opened 2022-12-22
 doc: document format fields in flux-resource(1) 

chu11:
tackling the most difficult issues before the holiday break!  

Hopefully I got all of this right, had to go through code to figure out what was available / allowed.

### DIFF
--- a/doc/man1/flux-resource.rst
+++ b/doc/man1/flux-resource.rst
@@ -109,6 +109,68 @@ named format may be generated with ``--format=get-config=NAME``.  See
 :man1:`flux-jobs` *OUTPUT FORMAT* section for a detailed description of this
 syntax.
 
+Resources are combined into a single line of output when possible depending on
+the supplied output format.  Resource counts are not included in the
+determination of uniqueness.  Therefore, certain output formats will alter the
+number of lines of output.  For example:
+
+::
+
+   $ flux resource list -no {nnodes}
+
+Would simply output a single of output containing the total number of nodes.
+The actual state of the nodes would not matter in the output.
+
+The following field names can be specified for the **status** and **drain**
+subcommands:
+
+**state**
+   State of node(s): "online", "offline", "avail", "exclude", "drain",
+   "draining", "drained", "all"
+
+**nnodes**
+   number of nodes
+
+**ranks**
+   ranks of nodes
+
+**nodelist**
+   node names
+
+**timestamp**
+   If node(s) in drain/draining/drained state, timestamp of node(s)
+   set to drain.
+
+**reason**
+   If node(s) in drain/draining/drained state, reason node(s) set to
+   drain.
+
+The following field names can be specified for the **list** subcommand:
+
+**state**
+   State of node(s): "up", "down", "allocated", "free", "all"
+
+**properties**
+   Properties associated with resources.
+
+**nnodes**
+   number of nodes
+
+**ncores**
+   number of cores
+
+**ngpus**
+   number of gpus
+
+**ranks**
+   ranks of nodes
+
+**nodelist**
+   node names
+
+**rlist**
+   Short form string of all resources.
+
 
 CONFIGURATION
 =============
--- a/doc/man1/flux-resource.rst
+++ b/doc/man1/flux-resource.rst
@@ -103,9 +103,12 @@ OUTPUT FORMAT
 =============
 
 The *--format* option can be used to specify an output format using Python's
-string format syntax or a defined format by name. For a list of built-in
-and configured formats use ``-o help``. See :man1:`flux-jobs` *OUTPUT FORMAT*
-section for a detailed description of this syntax.
+string format syntax or a defined format by name. For a list of built-in and
+configured formats use ``-o help``.  A configuration snippet for an existing
+named format may be generated with ``--format=get-config=NAME``.  See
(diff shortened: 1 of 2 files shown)


doc/man1/flux-resource.rst
```diff
@@ -102,11 +102,14 @@ COMMANDS
 OUTPUT FORMAT
 =============
 
-The *--format* option can be used to specify an output format using Python's
-string format syntax or a defined format by name. For a list of built-in
-and configured formats use ``-o help``. See :man1:`flux-jobs` *OUTPUT FORMAT*
+The *--format* option can be used to specify an output format using
```
grondo, line 105: I'm assuming it was accidental that this was formatted to ~70 columns width instead of 78 like the rest of the file.

doc/man1/flux-resource.rst
```diff
@@ -109,6 +109,56 @@ snippet for an existing named format may be generated with
 ``--format=get-config=NAME``.  See :man1:`flux-jobs` *OUTPUT FORMAT*
 section for a detailed description of this syntax.
 
+The field names that can be specified for the **status** and **drain**
+commands are:
```
grondo, line 113: format at 78 columns instead of ~70?

doc/man1/flux-resource.rst
```diff
@@ -109,6 +109,56 @@ snippet for an existing named format may be generated with
...
+
+**state**
+   State of node(s): "online", "offline", "avail", "exclude", "drain",
+   "draining", "drained", "all"
+
+**nnodes**
+   The number of nodes in state.
```
grondo, line 120: I don't think this is just the number of nodes in a given state. It reports the number of nodes in the line being printed. If the format is only `{nnodes}`, then the total number of nodes is printed, e.g.
```
$ flux resource list -no {nnodes}
121
```

I'd suggest it is clearest to just say "The number of nodes"

doc/man1/flux-resource.rst
```diff
@@ -109,6 +109,56 @@ snippet for an existing named format may be generated with
...
+
+**ranks**
+   The ranks of the nodes in state.
```
grondo, line 123: See above note for `nnodes`

doc/man1/flux-resource.rst
```diff
@@ -109,6 +109,56 @@ snippet for an existing named format may be generated with
...
+
+**nodelist**
+   The nodes in state.
```
grondo, line 126: See above note for `nnodes`

doc/man1/flux-resource.rst
```diff
@@ -109,6 +109,56 @@ snippet for an existing named format may be generated with
...
+
+**timestamp**
+   If node(s) in drain/draining/drained state, timestamp of node(s)
+   set to drain.
+
+**reason**
+   If node(s) in drain/draining/drained state, reason node(s) set to
+   drain.
+
+The field names that can be specified for the **list** are:
```
grondo, line 136: `s/for the **list** are/for the **list** subcommand/`?

doc/man1/flux-resource.rst
```diff
@@ -109,6 +109,56 @@ snippet for an existing named format may be generated with
...
+
+**state**
+   State of node(s): "up", "down", "allocated", "free", "all"
+
+**properties**
+   Properties associated with resources in state.
```
grondo, line 142: see above note for `nnodes`

doc/man1/flux-resource.rst
```diff
@@ -109,6 +109,56 @@ snippet for an existing named format may be generated with
...
+   The number of cores in state.
+
+**ngpus**
+   The number of gpus in state.
+
+**ranks**
+   The ranks of the nodes in state.
+
+**nodelist**
+   The nodes in state.
```
grondo, line 157: These all have the same comments as above.

Since this is duplicated information, I wonder if there is some way to share the list between the `status` and `list` subcommands and maybe just explain the differences? (Just an idea, if that is awkward, ignore it)

grondo (commented):
Initial comments inline.

doc/man1/flux-resource.rst
grondo: BTW, this is how `flux resource info` works, it is just 
```
$ flux resource list -no "{nnodes} Nodes, {ncores} Cores, {ngpus} GPUs"
121 Nodes, 5808 Cores, 968 GPUs
```

chu11: ahh, i see that there is a `resources_uniq_lines()` function that alters the output dependent on how many lines there will be.  I wasn't aware of that quirk.

grondo: Yeah, I wonder if that should be added to the documentation for `flux resource list`, `status` and `drain`. I'm struggling with a concise description of what the commands do though, maybe something like "These commands combine resources into a single line of output when possible depending on the supplied output format. Resource counts are not included in the determination of uniqueness."

doc/man1/flux-resource.rst
chu11: I had considered that approach, going with perhaps something like:

```
**nnodes**
   (list, status, drain) ....
   
**reason**
   (status, drain) ...   
```

but the `state` being so different between status and list, and them having a few fields that don't intersect, felt best to separate them out.

chu11:
re-pushed, try again with some tweaks and adding a chunk of text w/ example to explain how number of lines of output can be different given the output format specified.

Issue #4852, opened 2022-12-22
job-list: some jobs do not have t_submit attribute


grondo:
I noticed this on corona. Some (older) jobs are missing the `t_submit` attribute. E.g.
``` 
grondo@corona212:~/git/flux-core.git$ flux jobs -no {t_submit} ƒkEuFtiMs5
flux-jobs: ERROR: invalid JobInfo attribute 't_submit'
 grondo@corona212:~/git/flux-core.git$ flux job list-ids ƒkEuFtiMs5 | jq
{
  "id": 5537405654270976,
  "userid": 6885,
  "urgency": 16,
  "priority": 159,
  "t_run": 1658937548.6476562,
  "t_cleanup": 1658938152.5519431,
  "t_inactive": 1658938164.2931011,
  "state": 64,
  "name": "sleep",
  "ntasks": 1,
  "ncores": 48,
  "duration": 0.0,
  "nnodes": 1,
  "ranks": "4",
  "nodelist": "corona293",
  "success": true,
  "exception_occurred": false,
  "result": 1,
  "expiration": 1659542348.0,
  "annotations": {
    "sched": {
      "queue": "default"
    },
    "user": {
      "uri": "ssh://corona293/var/tmp/grondo/flux-JpUjo4/local-0"
    }
  },
  "waitstatus": 0
}
```

AFAIK, all jobs should have the `t_submit` attribute. Since this seems to appear only on older jobs, I wonder if `job-list` drops this attribute for jobs on restart, or if a restart is at least somehow related.

chu11:
hmmm.  So in commit 9f7aa094e3533cb49c6bbd02c8535e7209aef31e

```
@@ -932,6 +932,8 @@ static struct job *eventlog_restart_parse (struct list_ctx *ctx,
         if (!strcmp (name, "submit")) {
             if (submit_context_parse (ctx->h, job, context) < 0)
                 goto error;
+        }
+        else if (!strcmp (name, "validate")) {
             update_job_state (ctx, job, FLUX_JOB_STATE_DEPEND, timestamp);
         }
         else if (!strcmp (name, "depend")) {
```

The `validate` event transitions to the DEPEND state and sets `t_submit`.  The eventlog for the job you list above has no `validate` event, so I guess run before we updated the system with this new change to the state machine.

So I guess the question is:

[...]

PR #4853, opened 2022-12-28
 job-list: separate t_submit / t_depend  calculation

chu11:
Problem: Prior to the introduction of the validate event in

https://github.com/flux-framework/flux-core/commit/9f7aa094e3533cb49c6bbd02c8535e7209aef31e

the t_submit and t_depend timestamps returned by job-list were
identical because the 'submit' timestamp lead to the DEPEND
state transition.  Both attribute names were supported for
convenience.

However, after the introduction of the 'validate' event, these two
values should be different.  t_submit should be associated with the
'submit' event.  t_depend should be associated with the 'validate'
event and transition to the DEPEND state.

This lack of a split between the two values resulted in data
errors for older jobs without a 'validate' event.  Without the
'validate' event job-list would never process that a t_submit/t_depend
timestamp ever occurred.  So job-list would not return a value of
t_submit/t_depend for any job without a 'validate' event.

Solution: Set t_submit and t_depend independently from each other.
t_submit is set all the time from the 'submit' event.  t_depend
is set on transition to the DEPEND state.  Therefore if a transition
to a DEPEND state does not occur, it will not be set.

On older jobs without a 'validate' event, t_depend may not be set as a
result.  However t_submit will be set on every job.

Fixes https://github.com/flux-framework/flux-core/issues/4852

### DIFF
--- a/src/modules/job-list/job_data.h
+++ b/src/modules/job-list/job_data.h
@@ -21,7 +21,8 @@
  *
  * associated eventlog entries when restarting
  *
- * t_depend - "submit"
+ * t_submit = "submit"
+ * t_depend - "validate"
  * t_priority - "priority" (not saved, can be entered multiple times)
  * t_sched - "depend" (not saved, can be entered multiple times)
  * t_run - "alloc"
@@ -36,8 +37,7 @@ struct job {
     int urgency;
     int64_t priority;
     double t_submit;
-    // t_depend is identical to t_submit
-    // double t_depend;
+    double t_depend;
     double t_run;
     double t_cleanup;
     double t_inactive;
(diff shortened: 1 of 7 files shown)


garlick:
Maybe if version is not set in the submit event (implying version < 1),
t_depend should be set to t_submit?

chu11:
> Maybe if version is not set in the submit event (implying version < 1), t_depend should be set to t_submit?

Ahh good idea, we could do that.

chu11:
re-pushed, for backwards compatibility, `t_submit == t_depend`

garlick (approved):
LGTM!

PR #4855, opened 2023-01-03
libjob: deprecate flux_job_list() and flux_job_list_inactive()

chu11:
Per #4514 

I gave this issue another go around late last week.  The major gotcha is the auto-generation of python bindings to deprecated C functions, which we probably don't want to do.

Going back and forth with several options, I decided on the following approach.  Set

```
#ifndef FLUX_DEPRECATED
...
#endif
```

around deprecated C functions, so that the python scripts wouldn't generate those functions.

The big pro was that it was relatively easy, not a lot of changes, no major script updates to python bindings stuff.

The con is, well, that this is all it is.  Using the python `deprecated` decorator appears more for non-auto-generated stuff and would probably require a non-trivial amount of scripting hacks to add that in, especially for auto generated stuff.

Considered a special "blocklist" file for the python bindings, but given how cffi generates things (just slurps up the .h files) didn't seem the easiest to add either.

### DIFF
--- a/src/cmd/flux-job.c
+++ b/src/cmd/flux-job.c
@@ -1357,10 +1357,19 @@ int cmd_list (optparse_t *p, int argc, char **argv)
     else
         userid = getuid ();
 
-    if (!(f = flux_job_list (h, max_entries, "[\"all\"]", userid, states)))
-        log_err_exit ("flux_job_list");
+    if (!(f = flux_rpc_pack (h,
+                             "job-list.list",
+                             FLUX_NODEID_ANY,
+                             0,
+                             "{s:i s:[s] s:i s:i s:i}",
+                             "max_entries", max_entries,
+                             "attrs", "all",
+                             "userid", userid,
+                             "states", states,
+                             "results", 0)))
+        log_err_exit ("flux_rpc_pack");
     if (flux_rpc_get_unpack (f, "{s:o}", "jobs", &jobs) < 0)
-        log_err_exit ("flux_job_list");
+        log_err_exit ("flux job-list.list");
     json_array_foreach (jobs, index, value) {
         char *str;
         str = json_dumps (value, 0);
@@ -1399,10 +1408,20 @@ int cmd_list_inactive (optparse_t *p, int argc, char **argv)
     if (!(h = flux_open (NULL, 0)))
         log_err_exit ("flux_open");
 
-    if (!(f = flux_job_list_inactive (h, max_entries, since, "[\"all\"]")))
-        log_err_exit ("flux_job_list_inactive");
+    if (!(f = flux_rpc_pack (h,
+                             "job-list.list",
+                             FLUX_NODEID_ANY,
+                             0,
+                             "{s:i s:f s:i s:i s:i s:[s]}",
+                             "max_entries", max_entries,
+                             "since", since,
+                             "userid", FLUX_USERID_UNKNOWN,
+                             "states", FLUX_JOB_STATE_INACTIVE,
+                             "results", 0,
+                             "attrs", "all")))
+        log_err_exit ("flux_rpc_pack");
     if (flux_rpc_get_unpack (f, "{s:o}", "jobs", &jobs) < 0)
-        log_err_exit ("flux_job_list_inactive");
+        log_err_exit ("flux job-list.list");
     json_array_foreach (jobs, index, value) {
         char *str;
         str = json_dumps (value, 0);
--- a/src/modules/job-archive/job-archive.c
+++ b/src/modules/job-archive/job-archive.c
@@ -499,12 +499,26 @@ void job_archive_cb (flux_reactor_t *r,
                      void *arg)
 {
     struct job_archive_ctx *ctx = arg;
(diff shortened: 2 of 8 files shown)


t/t2260-job-list.t
```diff
@@ -438,9 +438,9 @@ test_expect_success 'flux job list-inactive w/ since -1 leads to error' '
 	grep "Invalid argument" list_inactive_error1.out
 '
 
-test_expect_success 'flux job list-inactive w/ count -1 leads to error' '
+test_expect_success HAVE_JQ 'flux job list-inactive w/ count -1 leads to error' '
```
grondo, line 441: We don't need `HAVE_JQ` anymore since b63aefdd5b20762e5fc05e52073ca16f7a9776cb.

But also, the test doesn't appear to use `jq` anyway?

src/modules/job-list/list.c
```diff
@@ -226,6 +226,11 @@ void list_cb (flux_t *h, flux_msg_handler_t *mh,
         errno = EPROTO;
         goto error;
     }
+    if (since < 0.) {
```
grondo, line 229: commit message suggestion: "ensure since is >= 0" doesn't parse without more context. Suggest "ensure since parameter >= 0"?

t/t2260-job-list.t
```diff
@@ -433,14 +433,14 @@ test_expect_success 'flux job list-inactive w/ count limits output of inactive j
 	test $count -eq 14
 '
 
-test_expect_success 'flux job list-inactive w/ since -1 leads to error' '
+test_expect_success HAVE_JQ 'flux job list-inactive w/ since -1 leads to error' '
```
grondo, line 436: Again, `HAVE_JQ` not necessary, but also appears no use of `jq` here?

src/cmd/flux-job.c
```diff
@@ -1357,10 +1357,19 @@ int cmd_list (optparse_t *p, int argc, char **argv)
     else
         userid = getuid ();
 
-    if (!(f = flux_job_list (h, max_entries, "[\"all\"]", userid, states)))
-        log_err_exit ("flux_job_list");
+    if (!(f = flux_rpc_pack (h,
+                             "job-list.list",
```
grondo, line 1361: Commit message: suggestion: `flux-job: do not call deprecated functions`

src/modules/job-archive/job-archive.c
```diff
@@ -499,12 +499,26 @@ void job_archive_cb (flux_reactor_t *r,
                      void *arg)
 {
     struct job_archive_ctx *ctx = arg;
-    char *attrs = "[\"userid\", \"ranks\", \"t_submit\", " \
-                   "\"t_run\", \"t_cleanup\", \"t_inactive\"]";
     flux_future_t *f;
 
-    if (!(f = flux_job_list_inactive (ctx->h, 0, ctx->since, attrs))) {
-        flux_log_error (ctx->h, "%s: flux_job_list_inactive", __FUNCTION__);
+    if (!(f = flux_rpc_pack (ctx->h,
```
grondo, line 504: Similar suggestion for the other commit message - instead of "remove flux_job_list() warnings", just "do not use flux_job_list()"? Eh, take or leave these suggestions.

src/common/libjob/test/job.c
```diff
@@ -115,46 +115,6 @@ void check_corner_case (void)
     ok (flux_job_submit_get_id (NULL, NULL) < 0 && errno == EINVAL,
         "flux_job_submit_get_id with NULL args fails with EINVAL");
 
-    /* flux_job_list */
```
grondo, line 118: Hm, it seems like we should be able to still run these tests against even deprecated functions. Maybe the right solution here is to suppress warnings about deprecated functions for this test only?

grondo:
Also, just noticed in the PR title: s/deprecated/deprecate/

t/t2260-job-list.t
chu11: agh sorry, rebase error, this PR was originally from before that HAVE_JQ removal

chu11:
re-pushed, fixing up things per the comments above.  The big notable change is the last commit, `-Wno-deprecated-declaration` is added ot the build of `libjob` `job` unit tests, to avoid warnings when running those tests.

grondo (approved):
LGTM!

vsoch:
hey @chu11 ! I'm working on the bindings tonight and found this commit https://github.com/flux-framework/flux-core/commit/0d08fb8340e29f03acab78ae8f708b01e1bccab6 and am going to try and mimic what I see in the Makefile (but I might have some questions!)

vsoch:
OK I think I figured it out - this is one of those preprocess flags for gcc, and it would look like this:
```console
Writing stage 1 clean header to /workspaces/flux-python/src/_core_clean.h
/usr/bin/gcc -E -D __attribute__(...)= -DFLUX_DEPRECATED(...)= /workspaces/flux-python/src/_core_clean.h -o /workspaces/flux-python/src/_core_preproc.h
```

vsoch:
And the warnings went away! It's magical! :sparkles:

Issue #4804, opened 2022-12-07
job-manager: updates checkpoint.job-manager each time queue state changes


garlick:
Problem: As noted in #4776, 1fc00e95d3f15e2a74b073892ce81da1cf15c36b added support to save/restore job manager queue state with the job manager checkpoint.  For some reason it also forces a KVS commit of the checkpoint each time the state changes, which seems unnecessary.

We should probably only save the state when the module unloads.

PR #4856, opened 2023-01-05
job-manager: do not checkpoint on every queue state change

chu11:
Per #4804, the number of checkpoints to the KVS seems a bit excessive, as it occurs on every change.

Instead, do it only when the user asks to via a `sync` flag.  So that users can say "hey preserve this specific change just in case of catastrophic failure", as well as to help with testing.

I made one additional change that I think is important, when checkpointing in the `job-manager`, make sure that the FLUX_KVS_SYNC flag is set, to ensure the data is on the backing store as well.  Would have slowed things down before, but now that syncs are optional, I think it more than offsets it.

### DIFF
--- a/src/modules/job-manager/queue.c
+++ b/src/modules/job-manager/queue.c
@@ -619,8 +619,6 @@ static void queue_enable_cb (flux_t *h,
         if (jobq_enable (q, enable, disable_reason) < 0)
             goto error;
     }
-    if (restart_save_state (queue->ctx) < 0)
-        flux_log_error (h, "problem saving checkpoint after queue change");
     if (flux_respond (h, msg, NULL) < 0)
         flux_log_error (h, "error responding to job-manager.queue-enable");
     return;
@@ -719,8 +717,6 @@ static void queue_start_cb (flux_t *h,
         else
             queue_stop (queue, name);
     }
-    if (restart_save_state (queue->ctx) < 0)
-        flux_log_error (h, "problem saving checkpoint after queue change");
     if (flux_respond (h, msg, NULL) < 0)
         flux_log_error (h, "error responding to job-manager.queue-start");
     return;
--- a/doc/man1/flux-queue.rst
+++ b/doc/man1/flux-queue.rst
@@ -84,7 +84,7 @@ OPTIONS
    Be taciturn.
 
 **-a, --all**
-   Use with *enable*, *disable*, *stop, or *start* subcommands to
+   Use with *enable*, *disable*, *stop*, or *start* subcommands to
    signify intent to affect all queues, when queues are configured but
    *--queue* is missing.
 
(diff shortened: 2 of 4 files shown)


src/modules/job-manager/restart.c
```diff
@@ -241,7 +241,7 @@ int restart_save_state (struct job_manager *ctx)
 
     if (!(txn = flux_kvs_txn_create ())
         || restart_save_state_to_txn (ctx, txn) < 0
-        || !(f = flux_kvs_commit (ctx->h, NULL, 0, txn))
+        || !(f = flux_kvs_commit (ctx->h, NULL, FLUX_KVS_SYNC, txn))
```
garlick, line 244: Again - high cost operation, and we already have `kvs.checkpoint-period` periodically saving the KVS root.

src/cmd/flux-queue.c
```diff
@@ -49,6 +49,9 @@ static struct optparse_option stop_opts[] = {
     { .name = "all", .key = 'a', .has_arg = 0,
       .usage = "Force command to apply to all queues if none specified",
     },
+    { .name = "sync", .has_arg = 0,
+      .usage = "Flush current queue state to KVS after update",
+    },
```
garlick, line 54: As noted previously I think this option is not something an admin needs/wants.

garlick (changes requested):
As noted inline, I think the KVS_SYNC flag is inappropriate, and the user-facing option is probably not the right way to go.   However, maybe the change suggested in #4856 is going too far in the opposite direction?

If we left the KVS commit of queue state changes as is, then at least `kvs.checkpoint-period` would make sure the backing store is kept reasonably current in the event of a crash.  Thoughts?

src/modules/job-manager/queue.c
```diff
@@ -610,8 +612,10 @@ static void queue_enable_cb (flux_t *h,
         if (jobq_enable (q, enable, disable_reason) < 0)
             goto error;
     }
-    if (restart_save_state (queue->ctx) < 0)
-        flux_log_error (h, "problem saving checkpoint after queue change");
+    if (sync) {
+        if (restart_save_state (queue->ctx) < 0)
+            flux_log_error (h, "problem saving checkpoint after queue change");
+    }
```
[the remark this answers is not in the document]
chu11: Yeah, I basically went down this path solely to keep some current tests working (they check `checkpoint.job-manager` directly after a `flux queue enable/disable/start/stop`.  Will re-do and just re-work the tests instead!

chu11:
re-pushed, going with an alternate approach.  Simply do not checkpoint on every queue state change, only checkpoint when the `job-manager` module is unloaded.

Had to update some tests as a result.

garlick (approved):
LGTM!

garlick:
Any reason not to set MWP on this?

chu11:
nope, was just waiting for your approval :-)

Issue #4822, opened 2022-12-10
job-manager: default queues to disabled and stopped


chu11:
from a discussion in #4776

> Another related thought is that we could make the initial state of a named queue disabled and stopped so that we could add them on the fly without there being a race where they might be used before that is intended. For example, the admins could add a DAT queue to the config and then enable but not start it to allow jobs to be submitted before the DAT. It would not be good to do that for the anonymous queue though, since that would break batch jobs. Just a thought to consider.

chu11:
Note that queues automatically being enabled/started led to this annoyance in some flux-accounting tests

https://github.com/flux-framework/flux-accounting/pull/302

newly created queues were automatically enabled/started and jobs began to run, leading to errors in tests.

chu11:
hmmm, its sort of annoying having to do

```
flux queue enable
flux queue start
```

after every new queue that is setup.  Admittedly this is mostly annoying in testing b/c we're making new queues all of the time.  Wondering if we could combine into a single command.  like

`flux queue enable --start`

chu11:
OR ... should queues be defaulted to enabled and stopped?  i.e. jobs can be submitted, but none schedulable?

garlick:
Or maybe the reverse (disabled, running)?  If no jobs are enqueued then none can be scheduled right?

We're only talking about brand new queues since their state is saved/restored on each Flux restart after that.

chu11:
> (disabled, running)

I assume you mean `disabled/started`.

My assumption is the common situation is gonna be for a DAT, so the admin sets up a new queue named `DAT_queue_jan_5_2023` or something.    We allow the user(s) of the DAT to start submitting, and when the DAT begins, the admin just starts it (`flux queue start`)??

garlick:
Oops, yeah I meant `disabled/started`.

I had assumed there would be a single, long-lived "all" queue for DATs and hadn't considered queues created for each one, but that seems like something people might want to do since it would allow preparations for multiple DATs (or DAT phases) to be made in parallel.

Pondering a bit more, you're  probably right about `enabled/stopped` being the preferred initial state.  I was thinking the admin would want control over when to allow jobs to be submitted, but it might be surprising if the queue is enabled and the jobs start running!  Yeah, I'm voting for your original proposal now.

PR #4857, opened 2023-01-07
job-manager: default queues to disabled and stopped

chu11:
Per #4822

Per a comment in #4822, I would like to throw up for discussion if the default should be "enabled / stopped" instead of "disabled / stopped".  So potential we'll want to tweak this.


garlick:
Does @ryanday36 have an opinion on what the initial state of a newly created queue should be?

The initial thought was "disabled" (users cannot submit) and "stopped" (no scheduling).  @chu11 is suggesting maybe "enabled" (users can submit) and "stopped" (no scheduling).

As noted in #4822, the current enabled/started initial state seems like it could result in...surprises.

ryanday36:
A default of enabled/stopped seems reasonable to me. It might also be useful to allow the admin to set the state in the config somewhere. e.g. we currently have various `policy.limit` set by queue in `policy.toml`. Could we add `policy.state` to set default disabled/enabled and stopped/started? I guess then we'd also have to decide on what takes precedence when things get restarted, so maybe that's more than we need to worry about right now.

garlick:
Thanks @ryanday36 !  I'll open an issue on config file support as I think that could be discussed independently of this.

chu11:
per discussion here and in #4822, lets go with enabled / stopped as the default.

chu11:
oops, had some missing `&&` chains, squashed and re-pushed

garlick (approved):
LGTM.

I assume since the accounting and sched tests are working in CI that this doesn't break anything over there...

chu11:
> I assume since the accounting and sched tests are working in CI that this doesn't break anything over there...

That's the assumption (although flux-sched builder seems hung on something non-testing related.  but I built and tested against this tree w/ flux-sched and all was good)

chu11:
looks like we'll need to do some `flux-sched` updates, i accidentally tested against master locally, oops

garlick:
I'd guess the sched updates are just adding `flux queue start` which would be no-ops if the queue is already started, so we could merge that change in sched before this goes in?

chu11:
> I'd guess the sched updates are just adding flux queue start which would be no-ops if the queue is already started, so we could merge that change in sched before this goes in?

Yup, that's my plan.  Just making sure it works against `master` right now.

Issue #4765, opened 2022-11-09
add QUEUE column to flux resource list


garlick:
Problem: `flux resource list` does not break down the resource list by queue

After #4764 is fixed, we could have `flux-resource` pull down the config object.  If  queues are configured, it could add a QUEUE column to its output.

Possibly it could also remove the configured queue constraints from the PROPERTIES column, since that is likely redundant.

PR #4859, opened 2023-01-12
flux-resource: support QUEUE output in resource list

chu11:
Per #4765

Just throwing this up here to see if there's any initial feedback.

This is a little tricky to support because:

A) "queue" is not returned from a `sched.resource-status` response ... thus

B) "queue" is not in the ResourceSet that is ultimately constructed from said response and fed into output code

I decided to create a new class currently called "ResourceSetPlusQueue", which just adds a "queue" property on top of "ResourceSet".  You initialize it with the flux configuration, and some logic is done to find queue(s) that meet the property requirements of the resources.

maybe some corner cases still as I haven't written any tests, but thought I'd just throw this up here if there's any initial thoughts on if this approach is sensible.  And maybe I'll come up with a better class name :P

my initial dumb test

```
#!/bin/sh

flux R encode -r 0-3 -p batch:0-1 -p debug:2-3 | tr -d '\n' | flux kvs put -r resource.R=-

flux config load <<-EOT
[queues.batch]
requires = [ "batch" ]

[queues.debug]
requires = [ "debug" ]

EOT

flux module unload sched-simple
flux module reload resource
flux module load sched-simple

flux resource list -o "{state:>10} {queue:>10} ?:{properties:<10.10+} {nnodes:>6} {ncores:>8} {ngpus:>8} {nodelist}"
```

```
     STATE      QUEUE PROPERTIES NNODES   NCORES    NGPUS NODELIST
      free      batch batch           1        1        0 catalyst160
 allocated                            0        0        0 
      down      batch batch           1        1        0 catalyst160
      down      debug debug           2        2        0 catalyst[160,160]
```

### DIFF
--- a/src/cmd/flux-resource.py
+++ b/src/cmd/flux-resource.py
@@ -56,14 +56,14 @@ class FluxResourceConfig(UtilConfig):
         "default": {
             "description": "Default flux-resource list format string",
             "format": (
-                "{state:>10} ?:{properties:<10.10+} {nnodes:>6} "
+                "{state:>10} ?:{queue:<8.8} ?:{properties:<10.10+} {nnodes:>6} "
                 "{ncores:>8} {ngpus:>8} {nodelist}"
             ),
         },
         "rlist": {
             "description": "Format including resource list details",
             "format": (
-                "{state:>10} ?:{properties:<10.10+} {nnodes:>6} "
+                "{state:>10} ?:{queue:<8.8} ?:{properties:<10.10+} {nnodes:>6} "
                 "{ncores:>8} {ngpus:>8} {rlist}"
             ),
         },
@@ -432,7 +432,7 @@ class ResourceSetExtra(ResourceSet):
     @property
     def queue(self):
         queues = ""
-        if self.flux_config and self.flux_config["queues"]:
+        if self.flux_config and "queues" in self.flux_config:
             properties = json.loads(self.get_properties())
             for key, value in self.flux_config["queues"].items():
                 if value["requires"] and set(value["requires"]).issubset(
--- a/src/cmd/flux-resource.py
+++ b/src/cmd/flux-resource.py
@@ -430,6 +430,17 @@ class ResourceSetExtra(ResourceSet):
             super().__init__(arg, version)
 
     @property
+    def propertiesx(self):
+        properties = json.loads(self.get_properties())
+        queues = self.queue
+        if self.queue:
+            queues = queues.split(",")
+            for q in queues:
+                if q in properties:
+                    properties.pop(q)
+        return ",".join(properties.keys())
+
+    @property
     def queue(self):
         queues = ""
         if self.flux_config and "queues" in self.flux_config:
@@ -448,7 +459,7 @@ def resources_uniq_lines(resources, states, formatter, config):
     the ResourceSet formatter argument. Include only the provided states
     """
     #  uniq_fields are the fields on which to combine like results
-    uniq_fields = ["state", "properties", "queue"]
+    uniq_fields = ["state", "properties", "propertiesx", "queue"]
 
     #
     #  Create the uniq format by combining all provided uniq fields:
@@ -504,6 +515,7 @@ def list_handler(args):
         "state": "STATE",
         "queue": "QUEUE",
         "properties": "PROPERTIES",
+        "propertiesx": "PROPERTIES",
         "nnodes": "NNODES",
         "ncores": "NCORES",
         "ngpus": "NGPUS",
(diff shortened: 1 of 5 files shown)


src/cmd/flux-resource.py
```diff
@@ -446,14 +468,15 @@ def resources_uniq_lines(resources, states, formatter):
 
     #  Create a mapping of resources sets that generate uniq "lines":
     lines = {}
+    config = flux.Flux().rpc("config.get").get()
```
grondo, line 471: This seems like a reasonable approach to me! Possibly this is just prototype code at this point, and if so ignore my comment, but the fetch of the config should probably go outside of this function into `list_handler()` directly, at least so that only a single flux handle is required.

On another note, it is a bit unfortunate that queues and their properties will now be duplicated in output -- an unfortunate side effect of the design that resources assigned to a queue are selected by properties. I guess we could try removing properties that are the same as the queue name, then the PROPERTIES column would go away when they match exactly those in QUEUE,  but this seems like it could end up being more confusing...

chu11: > On another note, it is a bit unfortunate that queues and their properties will now be duplicated in output -- an unfortunate side effect of the design that resources assigned to a queue are selected by properties. I guess we could try removing properties that are the same as the queue name, then the PROPERTIES column would go away when they match exactly those in QUEUE, but this seems like it could end up being more confusing...

yeah, @garlick mentioned that in #4765.  There's a part of me that wants to leave them there, b/c while we often name the properties the same names as the queues, you don't have to.  I dunno, something we can debate as I flush this out more.

grondo: What if you only remove the queue property from the properties list if it matches the queue name. I think keeping the output of `flux resource list` as simple as possible should be a goal here, not just to satisfy the exact request.

Then, this could allow us to rethink how queues are configured to make this less confusing, without eventually breaking or changing the output of `flux resource list`.

chu11: hmmm, i guess this is where it gets tricky.  Should it be removed from PROPERTIES only if the user is going to output QUEUE?  Or should it not matter if the user is outputting QUEUE?

grondo: Good question. A user or sysadmin may want to be able to query the set of all properties with `flux resource list -no {properties}`, or `flux resource list -no '{properties:<25} {nodelist}'`. Always removing queue properties from PROPERTIES would somewhat thwart this use case. However, I can see why you would lean toward doing this since it might _also_ be confusing if the listed PROPERTIES change based on what headings are used. However, if we take that approach, then we are effectively making queue properties "hidden". Of course, we could add an option to `flux resource` to change this behavior, and then pick one as the default.

chu11: perhaps we could have an alternate `properties` key?  like `properties_simple` or something.

grondo: That is a pretty good idea! :+1: Maybe `properties_redacted` or `properties_other` or just `propertiesx` (for properties-excluded). Anything is probably fine since this would likely be in the default format, but not used too often when specifying a format manually.

chu11:
re-pushed, added tests, incorporated comments from above.  it's a good first candidate for review and merge, the default output of `flux resource list` is now:

```
{state:>10} ?:{queue:<8.8} ?:{propertiesx:<10.10+} {nnodes:>6} {ncores:>8} {ngpus:>8} {nodelist}
```

grondo (approved):
LGTM! Just a couple suggestions inline.

chu11:
re-pushed with a few fixups for comments above and one that should hopefully allow `flux-sched` to work (mistake was a C programming programming in Python :P)

```
-        if self.flux_config and self.flux_config["queues"]:
+        if self.flux_config and "queues" in self.flux_config:

```

grondo:
It just occurred to me that we could add additional fields related to queues in the `ResourceSetExtra` class, for example various policy limits. This makes me wonder if we want `queue` to be a class so that we can have fields like `{queue.limits.duration}`. Of course, if we promote `queue` to a class later we could always have its repr be equivalent to `{queue.name}` so perhaps nothing to do here for now.

chu11:
> It just occurred to me that we could add additional fields related to queues in the ResourceSetExtra class, for example various policy limits. This makes me wonder if we want queue to be a class so that we can have fields like {queue.limits.duration}. Of course, if we promote queue to a class later we could always have its repr be equivalent to {queue.name} so perhaps nothing to do here for now.

Lets log that for an issue for now.

chu11:
woot, flux-sched passes!  I'll squash, re-push, and set MWP

PR #4870, opened 2023-01-20
clean up little used broker attribute functionality

garlick:
Problem: the broker attribute design is outdated and contains bits that we no longer use.

Cull the broker attribute internal design back a bit so it is perhaps more comprehensible to newcomers.


grondo (approved):
Nice cleanup!

PR #4871, opened 2023-01-20
python: add flux.constraint.parser for RFC 35 Constraint Query Syntax

grondo:
This PR adds a `ConstraintParser` class to the Flux Python bindings which can parse the generic query syntax being proposed for RFC 35 in flux-framework/rfc#360. I'm posting the implementation early since it may help reviewers of the RFC PR to experiment with the syntax and give feedback in either PR.

The `flux.constraint.parser` module supports a `__main__` function, so you can run the parser like this for testing:
```console
$ src/cmd/flux python -m flux.constraint.parser --default-op=name 'a|b|c'
{"or": [{"name": ["a"]}, {"name": ["b"]}, {"name": ["c"]}]}
```

### DIFF
--- a/src/bindings/python/flux/Makefile.am
+++ b/src/bindings/python/flux/Makefile.am
@@ -53,6 +53,9 @@ nobase_fluxpy_PYTHON = \
 	uri/resolvers/pid.py \
 	uri/resolvers/slurm.py \
 	uri/resolvers/lsf.py \
+	constraint/parser.py \
+	constraint/parsetab.py \
+	constraint/__init__.py \
 	utils/parsedatetime/__init__.py \
 	utils/parsedatetime/parsedatetime.py \
 	utils/parsedatetime/warns.py \
@@ -79,7 +82,30 @@ if HAVE_FLUX_SECURITY
 nobase_fluxpy_PYTHON += security.py
 endif
 
+
+BUILT_SOURCES = \
+	constraint/parsetab.py
+
+CLEANFILES = \
+	constraint/parser.out \
+	$(BUILT_SOURCES)
+
+STDERR_DEVNULL = $(stderr_devnull_$(V))
+stderr_devnull_ =  $(stderr_devnull_$(AM_DEFAULT_VERBOSITY))
+stderr_devnull_0 = >/dev/null 2>&1
+
+#  Ensure that python-ply's parsetab.py is generated. Create
+#  parent directory if necessary in builddir. Note: this also
+#  creates a parser.out, which is explicitly removed in
+#  CLEANFILES above. If in builddir, constraint directory
+#  should be removed. That is done below in clean-local.
+#
+constraint/parsetab.py: constraint/parser.py
+	$(AM_V_GEN)$(MKDIR_P) constraint && $(PYTHON) $< \
+	  --outputdir $(builddir)/constraint $(STDERR_DEVNULL)
+
 clean-local:
+	-rmdir constraint 2>/dev/null || :
 	-rm -f *.pyc */*.pyc *.pyo */*.pyo
 	-rm -rf __pycache__ */__pycache__
 
--- a/src/bindings/python/flux/constraint/parser.py
+++ b/src/bindings/python/flux/constraint/parser.py
@@ -1,0 +1,456 @@
+##############################################################
+# Copyright 2023 Lawrence Livermore National Security, LLC
+# (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+#
+# This file is part of the Flux resource manager framework.
+# For details, see https://github.com/flux-framework.
+#
+# SPDX-License-Identifier: LGPL-3.0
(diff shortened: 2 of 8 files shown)


grondo:
This PR would add `python3-ply` as an explicit requirement for flux-core. Currently it is somewhat of an implicit dependency since flux-core depends on `cffi` which depends on `pycparser` which uses `ply`. In `bionic`, it appears that `pycparser` used an internal version of `ply`, so that module had to be explicitly added to the docker image.

(I did have some commits in my branch to update README and Dockerfiles, but I seem to have dropped those. I'll find it and push here)

grondo:
I'm removing WIP here because I'm basically done working on the implementation, barring any adjustments to https://github.com/flux-framework/rfc/pull/360.

grondo:
I've pushed an addition here to optimize the `flux mini --requires` use case. Like terms can now be optionally combined, which preserves backwards compatibility of multiple-use of `--requires`, e.g. `--requires=foo --requires=bar` can now generate:
```json
{"properties": ["foo", "bar"]}
```
instead of 
```json
{"and": [{"properties": ["foo"]}, {"properties": ["bar"]}]}
```

In general, when terms within an `and` can be combined like this, it is also more efficient to do so, so this feature will presumably be useful in the general case as well.

src/bindings/python/flux/constraint/parser.py
```diff
@@ -0,0 +1,453 @@
...
+        pos = self.quote_start
+        raise ConstraintSyntaxError(f"Unclosed quote '\"' at position {pos}")
+
+
+class ConstraintParser:
+    r"""
+    Base constraint query syntax parser class.
+
+    This class implements the simple constraint query language with
+    the following grammar:
```
garlick, line 177: Should this be introduced as RFC 35 constraint syntax in the comment block so future hackers know that the implementation is constrained by that document?

Also, not sure if "value" should be referred to as "operand" here?

t/Makefile.am
```diff
@@ -73,6 +73,7 @@ TESTSCRIPTS = \
 	t0024-content-s3.t \
 	t0028-content-backing-none.t \
 	t0029-filemap-cmd.t \
+	t0030-marshall.t \
```
garlick, line 76: Yikes!  My bad, sorry!

garlick (approved):
This looks good to me!  I'm finding the new dependency is already everywhere I build/test flux  so that's an easy one to swallow.  My python-fu is a bit weak so maybe @chu11 could give this a quick look as well?

src/bindings/python/flux/constraint/parser.py
grondo: Good point. I forgot to go back and reword the class description after RFC 35 was finalized.

grondo: Description updated.

src/bindings/python/flux/Makefile.am
```diff
@@ -53,6 +53,9 @@ nobase_fluxpy_PYTHON = \
 	uri/resolvers/pid.py \
 	uri/resolvers/slurm.py \
 	uri/resolvers/lsf.py \
+	constraint/parser.py \
```
chu11, line 56: commit message typo "Constrant" -> "Constraint"

chu11: and also "propsed" -> "proposed"

src/bindings/python/flux/Makefile.am
```diff
@@ -79,7 +82,30 @@ if HAVE_FLUX_SECURITY
...
+
+CLEANFILES = \
+	constraint/parser.out \
+	$(BUILT_SOURCES)
+
+STDERR_DEVNULL = $(stderr_devnull_$(V))
+stderr_devnull_ =  $(stderr_devnull_$(AM_DEFAULT_VERBOSITY))
+stderr_devnull_0 = >/dev/null 2>&1
+
+#  Ensure that pyton-ply's parsetab.py is generated. Create
```
chu11, line 97: pyton-> python

src/bindings/python/flux/constraint/parser.py
```diff
@@ -0,0 +1,453 @@
...
+    to split ``value`` into multiple entries. E.g. if
+    ::
+
+        split_values = { "op": "," }
+
+    Then ``op:foo,bar`` would result in ``{"op":["foo","bar"]}``.
+
+    Terms are joined by AND unless OR is specified, such that ``a b c``
+    is the same as ``a && b && c``. A term can be negated with ``-``,
+    but to negate a whole expression, NOT must be used.
```
chu11, line 213: given you have an example for the first statement, perhaps an example(s) for the second statement?

chu11 (commented):
similar to @garlick, python-fu is a little weak given lack of knowledge of ply, but looks well tested.  Just noticed a few typos.

grondo:

Issue #4873, opened 2023-01-21
t0001-basic.t: not ok 17 - and broker.boot-method=single


chu11:
this is on catalyst

```
not ok 17 - and broker.boot-method=single
#	
#		test libpmi = single
#	
```
appears to have started after PR #4865 was merged.

grondo:
Is this the test from `t0001-basic.t`? (If so, let's add that to the issue description)

However, this doesn't reproduce for me on catalyst. Can you try this in your environment?
```console
$ src/cmd/flux start -o,-Sbroker.rc1_path=,-Sbroker.rc3_path= flux getattr broker.boot-method
single
```

chu11:
oops yeah, was `t0001-basic.t`.

```
>src/cmd/flux start -o,-Sbroker.rc1_path=,-Sbroker.rc3_path= flux getattr broker.boot-method
libpmi
```

could be some weird pathing thing?

grondo:
Might be helpful to try with `FLUX_PMI_DEBUG=1` set, e.g.:
```console
$ FLUX_PMI_DEBUG=1 src/cmd/flux start -o,-Sbroker.rc1_path=,-Sbroker.rc3_path= flux getattr broker.boot-method
boot_pmi: flux-pmi-client: trying 'simple'
boot_pmi: flux-pmi-client: trying 'libpmi'
boot_pmi: flux-pmi-client: trying 'single'
boot_pmi: single: initialize: rank=0 size=1 name=single: success
boot_pmi: single: get key=flux.instance-level: key not found
boot_pmi: single: finalize: successs
single
```

garlick:
Haven't tried yet, but if slurm's libpmi.so is installed in ${libdir} it might be hitting that.
Perhaps this test should set FLUX_PMI_CLIENT_SEARCHPATH="simple single" to prevent that...

chu11:
it does seem to be finding `libpmi` somewhere.

```
>FLUX_PMI_DEBUG=1 src/cmd/flux start -o,-Sbroker.rc1_path=,-Sbroker.rc3_path= flux getattr broker.boot-method
boot_pmi: flux-pmi-client: trying 'simple'
boot_pmi: flux-pmi-client: trying 'libpmi'
boot_pmi: libpmi: initialize: rank=0 size=1 name=0.0: success
boot_pmi: libpmi: get key=flux.instance-level: invalid key argument
boot_pmi: libpmi: finalize: successs
libpmi
```

[...]

PR #4876, opened 2023-01-22
testsuite: drop fragile broker.boot-method test

garlick:
Problem: t0001-basic.t "broker.boot-method=single" test fails if a libpmi.so is found that falls through to singleton mode.

Drop the test.  It wasn't doing much.

Fixes issue #4873


garlick:
re-pushed with some noise removed from one of the commits.

grondo (approved):
LGTM!

PR #4880, opened 2023-01-23
docker: add site-packages to default python3 path

### LINKED ISSUE #4877: Feature request: Flux python bindings on Python path
Currently when I am in the flux base container, I can't see flux python bindings without starting a flux instance:

```bash
docker run --entrypoint bash -it fluxrm/flux-sched:focal
python3 -c "import flux"
```
```console
 fluxuser@20f8658bb2ea:~$ python3 -c "import flux"
ModuleNotFoundError: No module named 'flux'
```
I suspect that starting the instance handles adding everything to python path or similar, so not starting it means we don't have that. Here is starting the instance:

```bash
flux start --test-size=4
python3 -c "import flux" # works
```
This is typically logical because I couldn't do anything with flux without the instance! However I have a case of starting a flux instance via the broker, and then having a (sort of "sidecar") service that I want to be able to interact with it. If I start the service via any kind of flux start / submit, then I either get only a small set of resources, OR I get WAY too many (e.g., imagine running a huge cluster and giving that many processes to a dinky web server? Probably a bad idea!) So in my testing, when I simply add the flux location to the Pythonpath in the container:

```bash
docker run --entrypoint bash -it fluxrm/flux-sched:focal
export PYTHONPATH=/usr/lib/flux/python3.8
python3 -c "import flux"
```
That actually works! I think as we get more complex use cases for this base container the flux bindings should generally be on the path, and the user can still learn they need to start the flux instance to use them. I did this for my base containers for the Flux RestFul API and didn't realize it wasn't present in the base image (until now).

And my use case isn't actually working yet - still strugglebussing along but I'm not giving up.

grondo:
Problem: On debian based docker containers (bionic, focal), the flux python module is not found by default because

  /usr/lib/python3.x/site-packages

is not in the default search path. This is because debian based distributions only search 'dist-packages' instead of 'site-packages'. However, `make install` puts our Flux modules in 'site-packages' because this is the module install directory understood by autotools.

To work around this issue, place a 'site-packages.pth' file in 'dist-packages' to get Python to read site-packages as well as dist-packages.

Fixes #4877

### DIFF
--- a/src/test/docker/bionic/Dockerfile
+++ b/src/test/docker/bionic/Dockerfile
@@ -68,6 +68,9 @@ RUN for PY in python3.6 python3.7 python3.8 ; do \
 	    "markupsafe==2.0.0" \
             coverage cffi six pyyaml "jsonschema>=2.6,<4.0" \
             sphinx sphinx-rtd-theme sphinxcontrib-spelling; \
+	sudo mkdir -p /usr/lib/${PY}/dist-packages; \
+	echo ../site-packages >/tmp/site-packages.pth; \
+	sudo mv /tmp/site-packages.pth /usr/lib/${PY}/dist-packages; \
     done ; \
     apt-get -qq purge -y python3-pip \
  && apt-get -qq autoremove -y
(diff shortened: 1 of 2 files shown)


chu11 (approved):
seems reasonable to me

grondo:
Great, thanks @chu11!

Issue #4868, opened 2023-01-19
add flux mini batch --dump option (or similar)


garlick:
Problem: currently the KVS content for a batch job instance is lost when the job completes, but it might be useful for preserving debug info or for restarting the instance later when we can do that.

This is currently possible by setting the `flux mini batch --broker-opts=-Scontent.dump=foo.tgz` or similar.  Might be nice if there were a quick option for that which set the archive name to something with the job ID in it if that's possible.

grondo:
Was this completed by #4881?

garlick:
Yes, thanks for catching that!

PR #4881, opened 2023-01-23
support flux mini batch,alloc --dump[=FILE]

garlick:
In #4868, it was suggested (er by me) that an option to dump kvs contents when a batch job instance exits might be useful.

This extends the `content.dump=auto` code in rc3 to use the job ID and produce `flux-<jobid>-kvs.tgz` if requested when a `statedir` is not defined.

I started looking at `flux-mini.py` to see about adding a shorthand option for this and realized it might be better if such an option worked like `--output` and supported mustache templates.  So maybe this is the wrong approach and support for this should be added to the shell `batch` plugin instead.  Parking this pending some time to think more about it.  Meanwhile, input welcome!

### DIFF
--- a/doc/man1/flux-mini.rst
+++ b/doc/man1/flux-mini.rst
@@ -672,6 +672,14 @@ OTHER OPTIONS
    to separate files (including stdin) by newline. To separate by
    consecutive whitespace, specify ``--sep=none``.
 
+**--dump=[FILE]**
+   *(batch,alloc)* When the job script is complete, archive the Flux
+   instance's KVS content to ``FILE``, which should have a suffix known
+   to :linux:man3:`libarchive`, and may be a mustache template as described
+   above for ``--output``.  The content may be unarchived directly or examined
+   within a test instance started with the :man1:`flux-start` ``--recovery``
+   option.  If ``FILE`` is unspecified, ``flux-{{jobid}}-dump.tgz`` is used.
+
 .. _bulksubmit:
 
 BULKSUBMIT
--- a/doc/test/spell.en.pws
+++ b/doc/test/spell.en.pws
@@ -696,3 +696,4 @@ pkill
 deps
 queuem
 timelimit
+unarchived
--- a/src/cmd/flux-mini.py
+++ b/src/cmd/flux-mini.py
@@ -1705,6 +1705,13 @@ def add_batch_alloc_args(parser):
         help="Pass options to flux brokers",
     )
     parser.add_argument(
+        "--dump",
+        nargs="?",
+        const="flux-{{jobid}}-dump.tgz",
+        metavar="FILE",
+        help="Archive KVS on exit",
+    )
+    parser.add_argument(
         "-n",
         "--nslots",
         type=int,
@@ -1804,6 +1811,10 @@ class BatchCmd(MiniCmd):
             args.nslots = args.nodes
             args.exclusive = True
 
+        if args.dump:
+            args.broker_opts = args.broker_opts or []
+            args.broker_opts.append("-Scontent.dump=" + args.dump)
+
         jobspec = JobspecV1.from_batch_command(
             script=self.read_script(args),
             jobname=args.SCRIPT[0] if args.SCRIPT else "batchscript",
@@ -1868,6 +1879,10 @@ class AllocCmd(MiniCmd):
         if args.bg and not args.COMMAND:
             args.broker_opts = args.broker_opts or []
             args.broker_opts.append("-Sbroker.rc2_none=1")
+
+        if args.dump:
+            args.broker_opts = args.broker_opts or []
+            args.broker_opts.append("-Scontent.dump=" + args.dump)
 
         jobspec = JobspecV1.from_nest_command(
             command=args.COMMAND,
--- a/src/shell/batch.c
+++ b/src/shell/batch.c
@@ -53,10 +53,8 @@ batch_info_create (flux_shell_t *shell, json_t *batch)
 
     if (flux_shell_info_unpack (shell,
                                "{s:i s:I}",
-                               "rank",
-                               &b->shell_rank,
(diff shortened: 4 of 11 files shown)


garlick:
Here's a redo with a `-odump[=FILE]` option added to the batch plugin.

If just `-odump` is specified, then the dump file defaults to `flux-JOBID-dump.tgz`.
If FILE is specified, it may contain {{id}} or {{jobid}} like the output file specification.

Is this getting warmer?

garlick:
In case it wasn't evident, this enables this sort of thing:
```
$ flux mini batch -N1 -odump  --wrap /bin/true
ƒDDRjpHu
$ flux start --recover=flux-ƒDDRjpHu-dump.tgz
+-----------------------------------------------------
| Entering Flux recovery mode.
| All resources will be offline during recovery.
| Any rc1 failures noted above may result in
|  reduced functionality until manually corrected.
|
| broker.rc1_path    /home/garlick/proj/flux-core/etc/rc1
| broker.rc3_path    /home/garlick/proj/flux-core/etc/rc3
| config.path        -
| statedir           changes will not be preserved
|
| Exit this shell when finished.
+-----------------------------------------------------
ƒ(s=1,d=0,builddir) $ 
```
whereupon one could use the regular tools within the "recovery instance" to look at job eventlogs, access job output, etc..

garlick:
Not sure if the fact that this only works for batch jobs (and not for `flux mini alloc` for example) will cause confusion...

garlick:
Sounds good.  OK, reworked with `flux mini batch --dump[=FILE]` option and man page entry.

grondo:
Great, I'm going to try this one out. Don't forget to update the PR title (for release notes)

garlick:
Pushed a fixup for python linting (forgot to run  `scripts/format` on `flux-mini.py`.)

grondo:
Some things occurred to me after looking at this for a minute:
 - I wonder if it would be useful (and not dangerous) for the job shell to always expand job arguments as a mustache template, then this PR could be simplified and --dump could be supported for `flux mini alloc` as well.
 - it might be useful to have a strategy to make any batch job "dumpable" at runtime. Actually, can this already be done via something like: `flux proxy JOBID flux setattr content.dump=flux-JOBID-dump.tar.gz`? If so maybe a `flux job dumpable JOBID` or similar command could be easily developed.

garlick:
> I wonder if it would be useful (and not dangerous) for the job shell to always expand job arguments as a mustache template, then this PR could be simplified and --dump could be supported for flux mini alloc as well.

Could you elaborate on that?  Did you mean doing mustache expansion on the argv array that gets passed to libsubprocess, or on shell options in the jobspec before the plugins parse them, or...?

> it might be useful to have a strategy to make any batch job "dumpable" at runtime. Actually, can this already be done via something like: flux proxy JOBID flux setattr content.dump=flux-JOBID-dump.tar.gz? If so maybe a flux job dumpable JOBID or similar command could be easily developed.

Yes I think so, and good idea!

garlick:
At minimum it seems like `shell_outfile_mustache_render ()` was a poor naming choice if we intend it to be used in other contexts.

grondo:
> At minimum it seems like shell_outfile_mustache_render () was a poor naming choice if we intend it to be used in other contexts.

Yeah, there should be one shell mustache renderer created a startup. If we wanted to make the renderer extensible, we could actually call into the plugin stack for each template substitution `mustache.render.<name>`, which would allow plugins to add new substitution keys. In fact a user could supply substitution keys within a shell initirc which could be interesting. However, I can't think of a use case at the moment so probably best to delay any kind of extra effort there.

grondo:
Ooh, that could actually be used to implement a kind of multi program support, e.g. `flux mini run {{prog}}` and a shell intirc with something like (pseudocode):

```lua
plugin.register {
    name = "testrenderer",
    handlers = {
        { topic = "mustache.render.name",
           fn = function()
               if task.info.rank == 0 then
                   return "appmain"
               end
               return "appworker"
           end
    }
}
```

Actually I don't think lua plugins can return stuff in the plugin out args currently, and it would be better long term to support multi-prog in the jobspec tasks section, but you get the idea...

garlick:
I'll push an interim version that moves the shared mustache support to the shell proper and keeps the renderer alive for the duration of the shell.  It also does expansion on all the command line args as discussed.

What I didn't do yet was try to move `--dump` support from the batch plugin to `flux mini`, or add more general tests.

grondo:
>  It also does expansion on all the command line args as discussed.

Neat!

grondo:
I wasn't sure if was clear that now `flux mini alloc/batch --dump[=FILE]` coule be just a convenience option for appending `-Scontent.dump=[flux-{{id}}-dump.tar.gz|FILE]` to the `args.broker_opts` array.. that should remove the need for a special `batch.dump` key in shell options I think.

garlick:
Sorry yes that was clear!  I had pushed an intermediate step and hadn't context switched back to this PR to finish the job yet.

grondo:
Ok, no problem! I just didn't know if you were waiting on an ACK to move forward :+1:

garlick:
Just pushed an update which moves `--dump` arg handling to `flux mini` (for `submit` and `alloc`) and adds  tests.

garlick:
oops, broken && chain.  Re-pushed.

garlick:
I gave this PR another pass and fixed up some dumb things.  It's ready for a review.

grondo (approved):
This looks great to me! Tested it out and everything works well.

The mustache rendering of job arguments seems like it could be useful more generally, I wonder where it should be documented? For now, since only the jobid can be rendered, it is probably fine to keep it our little secret.

garlick:
Great! Yeah, I couldn't figure out where that documentation should go.  Maybe for later.   Thanks - I'll set MWP.

PR #4883, opened 2023-01-23
job-list: add jobspec and R parsing unit tests

chu11:
something that's been on the TODO list for awhile.

I'm sure there's some cases I've missed, but it puts a good amount there that wasn't there before.  And most important, the skeleton is there for future additions and regression testing.


grondo:
I didn't look at this yet, but an initial question occurred to me: there's already a bunch of valid and invalid jobspec under `t` -- would it make sense to have a sharness test that uses the test program developed here to drive a set of tests based on those instead of creating duplicates or new test input within src/modules/job-info?

chu11:
> I didn't look at this yet, but an initial question occurred to me: there's already a bunch of valid and invalid jobspec under t -- would it make sense to have a sharness test that uses the test program developed here to drive a set of tests based on those instead of creating duplicates or new test input within src/modules/job-info?

Yeah, I hand made the jobspecs for several (possible not good) reasons.

- a lof the sharness tests are in yaml, but the unit tests would require json.  Obviously could convert, etc. etc.
- i begin writing very specific jobspecs/Rs to hit corner cases in `job-list` paths, so i eventually sort of just continued down that path.

Not the greatest argument(s) to go down this path ... it just nudged me towards that direction.

grondo (approved):
This seems like a good idea to merge. We can decide later if we want to consolidate valid and invalid jobspecs and R for testing in one place project-wide and move things around then.

chu11:
re-pushed, forgot to EXTRA_DIST those new test files

chu11:
oops, re-pushed, a "top_builddir" should have been "top_srcdir"

PR #4890, opened 2023-01-25
fix message encoding problem introduced in v0.46.1

garlick:
As noted in issue#4886, some cleanup included in flux-core v0.46.1 inadvertently changed the message encoding so it no longer conforms to [RFC 3](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_3.html).  When mixing this version with prior versions, the mismatched version detection isn't triggered.  The logs are pretty useless.

This PR fixes the bug, adds tests that should catch this sort of problem should it happen again, and adds a helpful log message in case is v0.47 and beyond is running on a management node with v0.46.1 on computes at some point.


grondo (approved):
Great job on all the testing here! Didn't see anything else on which to comment, so approving.

garlick:
Thanks!  I noticed a commit message typo and one place where I have an superflous return statement so I'll go ahead and force push those changes.

garlick:
setting mwp, thanks again!

Issue #4891, opened 2023-01-26
flux-resource: ERROR: "Invalid format field 'requires' for <class '__main__.ResourceSetExtra'>


garlick:
I had an "all" queue configured (though not started) in my test system which overlapped the debug and batch queues, and found I got this error:
```
$ flux resource list
flux-resource: ERROR: "Invalid format field 'requires' for <class '__main__.ResourceSetExtra'>"
```
Dropping "all" from the config made the error go away.

grondo:
There probably just needs to be a default empty `requires` member in the ResourceSetExtra class (and we should add a test for this situation)

chu11:
```
        if self.flux_config and "queues" in self.flux_config:                                                                               
            properties = json.loads(self.get_properties())                                                                                  
            for key, value in self.flux_config["queues"].items():                                                                           
                if value["requires"] and set(value["requires"]).issubset(                                                                   
                    set(properties)                                                                                                         
                ):                                                                                                                          
                    queues = queues + "," + key if queues else key                                                                          
        return queues                                                    
```

yeah, my assumption is "requires" didn't exist for the "all" queue b/c it was "all", thus leading to a KeyError and the listed error:

```
raise KeyError(f"Invalid format field {exc} for {typestr}")
```

so if no "requires" is listed for a queue, then I guess its ok for all resources?

garlick:
I did confirm that [RFC 33](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_33.html#queue-configuration) says `requires` is optional.

grondo:
> so if no "requires" is listed for a queue, then I guess its ok for all resources?

An empty constraint matches everything.

chu11:
if a queue has no constraints, whats people's opinion on this output:

```
     STATE QUEUE    NNODES   NCORES    NGPUS NODELIST
      free all,batc      2        2        0 catalyst[159,159]
      free all,debu      2        2        0 catalyst[159,159]
 allocated all           0        0        0 
      down all           0        0        0 
```

its right, b/c "all" applies to allocated/down, even though no resources exist for it, but it sorta looks weird.

it would be annoying to check if resources != 0 in the other columns before adding the "queue" to the output list.

i could go either way on it

PR #4893, opened 2023-01-26
flux-resource: handle queues with no configured constraints

chu11:
Per discussion in #4891.

Only minor thing to debate is what I did in commit https://github.com/flux-framework/flux-core/commit/f4cbdd5dc6a5c5872c38331eb5928172c3da81ac.  I like it, but I can see others may not.

### DIFF
--- a/t/t2350-resource-list.t
+++ b/t/t2350-resource-list.t
@@ -236,4 +236,31 @@ test_expect_success 'flux resource lists no properties in propertiesx (extraprop
 	flux resource list -o "{state} {nnodes} {propertiesx}" > listpropx_extraprop.out &&
 	grep "free 4 foo" listpropx_extraprop.out
 '
+# N.B. we use the queue "every" b/c the word "all" happens to be in
+# the word "allocated", annoyingly messing up greps
+test_expect_success 'configure queues and resource split amongst queues, add every queue' '
+	flux R encode -r 0-3 -p batch:0-1 -p debug:2-3 \
+	   | tr -d "\n" \
+	   | flux kvs put -r resource.R=- &&
+	flux config load <<-EOT &&
+	[queues.every]
+	[queues.batch]
+	requires = [ "batch" ]
+	[queues.debug]
+	requires = [ "debug" ]
+	EOT
+	flux queue start --all &&
+	flux module unload sched-simple &&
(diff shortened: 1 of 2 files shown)


grondo:
> Only minor thing to debate is what I did in commit https://github.com/flux-framework/flux-core/commit/f4cbdd5dc6a5c5872c38331eb5928172c3da81ac. I like it, but I can see others may not.

That seems like the right choice to me!

chu11:
re-pushed, just saw a typo in a commit message

t/issues/t4891-flux-resource-empty-requires.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash -e
```
grondo, line 1: Question: is the specific issues test required, or is this already covered in `t2350-resource-list.t`? If it is already covered, then I'd vote to just drop this test.

BTW, if you keep this test, then it needs to restore the confguration of the t4000-issues instance, since this could cause subsequent tests to fail. Or, it should run in a subinstance.

chu11: yeah, its a duplicate.  We can drop.

chu11:
re-pushed, dropping the regression test that is duplicated in a new addition to t2350-resource-list.t

t/t2350-resource-list.t
```diff
@@ -168,7 +168,7 @@ test_expect_success 'flux resource lists expected queues in states (single)' '
 test_expect_success 'cleanup jobs' '
 	flux job cancel $(cat job1A.id) $(cat job1B.id)
 '
-test_expect_success 'configure queues and resource split amongst queues' '
```
grondo, line 171: commit message typo: "t23t0-resource-list" :arrow_right: "t2350-resource-list.t"

grondo: Oops that was supposed to be part of a review, sorry.

grondo (approved):
This looks good to me.

Since the PR title gets incorporated into release notes, I'd vote for a simpler title that describes the change to the uninitiated, e.g. `flux-resource: handle queues with no configured constraints`

chu11:
thanks, re-pushed with typo fix in commit message, setting MWP

Issue #4862, opened 2023-01-17
crashing with running jobs can lead to problems


garlick:
Problem: if the flux rank 0 broker crashes with running jobs and then restarts, the fluxion scheduler may fail when processing the hello protocol. 

```
2023-01-17T16:10:17.222950Z sched-fluxion-qmanager.err[0]: hello: error loading R for id=202046916802905088: No such file or directory
```

To get fluxion loaded, those jobs have to be removed from the KVS.

Until we can handle recovering running jobs, we should probably just force these jobs into inactive state.

Caveat:  tasks belonging to the job could still be running.

Edit: the fluxion bug that is triggered here is flux-framework/flux-sched#991

garlick:
I think #4894 addresses the short term issue here so closing.

PR #4894, opened 2023-01-28
work around fluxion inbability to recover running jobs

garlick:
Here's a workaround for the problem's we have with fluxion refusing to start when running jobs are in the KVS, described in #4862.

In this one, a failure of the scheduler's `hello` callback causes a fatal `scheduler-restart` job exception to be posted rather than a scheduler tear-down.  When that particular exception is raised, the job manager clears the flag that indicates that resources are allocated, so the `sched.free` RPC is skipped when the job enters cleanup.

This is a WIP pending writing tests.

### DIFF
--- a/src/common/libschedutil/hello.c
+++ b/src/common/libschedutil/hello.c
@@ -18,6 +18,26 @@
 #include "init.h"
 #include "hello.h"
 
+
+static void raise_exception (flux_t *h, flux_jobid_t id, const char *note)
+{
+    flux_future_t *f;
+
+    flux_log (h,
+              LOG_INFO,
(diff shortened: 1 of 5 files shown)


garlick:
If we can get flux-framework/flux-sched#1000 (ding ding ding! what did I win?) merged first, then this PR should start passing the sched CI test.

I'll try to add a test here that works with sched simple - something like 
- start a job
- unload sched simple
- exclude a node the job was using (at the `resource` level)
- load sched simple
- observe that the job received a fatal exception

grondo (approved):
LGTM! Nice solution.

garlick:
Thanks!  I'll set MWP.

PR #4889, opened 2023-01-25
flux-queue: rewrite in python

chu11:
Per discussion in #4763 / #4885, lets start by re-writing `flux-queue` in python before moving onto more things.

This is pretty much a direct translation into Python with python-izing a bunch of code style.

Maybe debate could be on the error or usage output?  Interface and output should be identical.

Issue #4763, opened 2022-11-08
need a way for users to query limits


garlick:
Problem: there is not currently an interface for users to query the configured policy limits (by queue if queues are enabled).

ryanday36:
My suggestion here, since I tend to just reach for the familiar, would be to add a couple of columns for queue limits to the `flux resource list` output a la the Slurm sinfo command. Alternatively, a `flux queue list` command could be added that summarizes the resources and limits in each queue.

garlick:
We had an offline discussion about possibly adding a `flux policy` command that would pull down the `[policy]` and `[queues]` config objects for a start and display the policy (global and per-queue) in a nice way for the user.

I'll open a separate issue on making the flux config object accessible by guests (currently a road block for this idea)

After getting that working, we could consider whether we can work out how to give this command access to the per-user policy/limits imposed by flux-accounting.

grondo:
From #4659 (not much new, but that seems to be a duplicate, so moving this comment here before closing it):

> We could make a flux queue list using similar interfaces built on flux.util.OutputFormat like flux jobs and flux resource list. This would present a consistent and familiar API for users. We'd need to make a Queue class and have name/limits/defaults/etc available as attributes of each queue object. We might want to wait to implement this until we're certain we don't have to redo anything about the current queues implementation.

chu11:
did we reach a conclusion on how to approach this?  I believe there are three approaches discussed over time:

1) new `flux policy` command
2) add `flux queue list` (or similar option)
3) add in `flux resource list` (related to comment in #4863)

one negative of 2 is that `flux-queue` is written in C, so all of the output format stuff we support in `flux-jobs` and `flux-resource` can't be used.  It wouldn't be the end of the world to convert it to python first.  `flux resource` seems like a possibly better place to put it?  And unlike `flux queue`, `flux resource` is already user facing vs admin facing.

garlick:
I'm not sure if it should be a new command or a new `flux resource` subcommand.  I'd slightly lean towards a new one but I could be convinced otherwise.

I think initially the command could just fetch the config object from the broker and dump all the `policy` attributes from  [RFC 33](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_33.html).  A `--queue` option could apply the queue overrides, if any.  So maybe a usage something like
```
$ flux policy [--queue=NAME] [attribute list]
```
[...]

PR #4896, opened 2023-01-30
flux-queue: support flux queue list

chu11:
Per discussion in #4763, this is built on top of #4889.  Just thought I'd throw it up here, see if there are any initial thoughts.

These are the currently supported fields:

```
+    headings = {
+        "queue": "QUEUE",
+        "limits.duration": "DURATION",
+        "limits.min.nnodes": "MIN NNODES",
+        "limits.max.nnodes": "MAX NNODES",
+        "limits.min.ncores": "MIN NCORES",
+        "limits.max.ncores": "MAX NCORES",
+        "limits.min.ngpus": "MIN NGPUS",
+        "limits.max.ngpus": "MAX NGPUS",
+    }
```

A) for consistency to `flux-jobs` and `flux-resource`, use `{queue}` for the queue name vs `{name}`.

B) b/c these are all queue specific, didn't feel like I needed the "queue" prefix on the limit fields (i.e. not `{queue.limits.duration}`).

C) I felt having fields with all of the depth like in the policy config (i.e. `limits.job-size.min.nnodes`) was excessive.  And also since python f-strings don't support hyphens, that just messes things up too.

I currently have a default output like so:

```
   "?:{queue:<8.8} {limits.duration!F:>8} {limits.max.nnodes:>10} "
   "{limits.max.ncores:>10} {limits.max.ngpus:>9}"
```

since I doubt the average user cares about the minimums.

Some example outputs:

anon queue w/ some settings

```
DURATION MAX NNODES MAX NCORES MAX NGPUS
      4h         10        inf       inf
```

some queues w/ some settings

```
QUEUE    DURATION MAX NNODES MAX NCORES MAX NGPUS
batch          4h         10        inf       inf
debug         30m          8          8         8
```

Perhaps the biggest question is how to deal with "unlimited"/"infinity".  The current output is "inf" per `math.inf`.  I think its a decent choice and gets the point across.

Although I haven't done it yet, I think it might be convenient to add a conversion flag, like hypothetically `!I`, that will output an empty string instead of "inf" if the value is infinity.  Since a number of clusters don't have gpus, we could then make the default output for ngpus `?:{limits.max.ngpus!I:>9}`, so along with the `?:` formatter, there will be no gpus output on the no-gpu clusters.


grondo:
Quick very early thoughts:

> Perhaps the biggest question is how to deal with "unlimited"/"infinity". The current output is "inf" per math.inf. I think its a decent choice and gets the point across.

`inf` works well since it is valid FSD, but I like your idea of displaying nothing when there is no configured limit so that `?:` can be used by default. `!F` could be used to force a value to FSD and display `inf` for unset values if desired.

Other quick comments:

 * Instead of DURATION, TIMELIMIT may be a more appropriate field name.
 * Is the default duration for each queue available as a field?
 * It might be just me, but I dislike headings with spaces, it makes it difficult to differentiate the columns. Instead of fields like `MAX NNODES`, I'd suggest simply "MAXNODES`. 
 * We may want to set up some convenient fields that include `min-max` in one field for users that are interested in minimums. The fields could then be simply labeled `NODES`, `CORES`, `GPUS`, example output might be:
   ```
   QUEUE    TIMELIMIT NODES  CORES   GPUS
   batch           4h  1-10  1-inf  1-inf
   debug          30m   1-8    1-8    1-8
   ```
 * I wonder if there should be a way to mark the default queue? Maybe just a `*` on the name?

chu11:
> Is the default duration for each queue available as a field?

You mean the default put into the jobspec?  Yeah, we could do that too.  So perhaps we should have `{limits.timelimit}` and `{default.timelimit}`.

> We may want to set up some convenient fields that include min-max in one field for users that are interested in minimums. 

Oh good idea.  Could have `{nnodes_range}`, and that would make more sense as the default output.

> I wonder if there should be a way to mark the default queue? Maybe just a * on the name?

Another good idea.  Could perhaps add as another field, `{queuem}`?? for queue-markup?  Perhaps I'll think of something better.

grondo:
> Oh good idea. Could have {nnodes_range}, and that would make more sense as the default output.

Just an idea: Since you have `{limits.min.nnodes}`, `{limits.max.nnodes}`, would `{limits.range.nnodes}` work? I haven't looked at the implementation to see how inconvenient that would be though. However, it may be more extensible in the long run...

chu11:
>   Just an idea: Since you have {limits.min.nnodes}, {limits.max.nnodes}, would {limits.range.nnodes} work? I haven't looked at the implementation to see how inconvenient that would be though. However, it may be more extensible in the long run...

Hmmmm, would minorly inconvenient b/c the "limits.range" object wouldn't have parental access to the "limits.min/max" objects.   Lemme think about it.

chu11:
re-pushed with a bunch of suggestions from above

- added `{queuem}` for "queue markup", add an asterisk next to the default queue
- added `{defaults.timelimit}`
- added `{limits.range.nnodes/ncores/ngpus}`
- struggling with header for `{defaults.timelimit}` vs `{limits.timelimit}`, I did "TIMELIMIT*" for defaults for now.  Any good suggestions?
- adjusted headers for MIN/MAX per comment above

example output


```
QUEUE    TIMELIMIT*  TIMELIMIT     NNODES     NCORES      NGPUS
batch*           2h         4h       2-10      1-inf        inf
debug            1h        30m        1-8        1-8        1-8
```

note for the ranged ones, I did "inf" if the range is "inf-inf", since I think that looks better.

chu11:
Struggling a bit with some of the presentation.

First I did this:

`{limit.ngpus:>10i}`

where the `i` would convert "inf" to hyphen like with the `h` special presentation.  Simple enough, a decent output option to indicate that nothing was set.

then I wanted to try and do:

`?:{limit.ngpus:>10I}`

where `I` converts "inf" to empty string and `?:` could eliminate a column if it was all empty string.

But it seems the code for `?:` only works before conversion / special presentation adjustments.  It (rightfully) did not take into account intentionally outputting an empty string if there was a non-empty string to begin with.

I dunno, I like the idea that "inf" should be output by default to make it clear to the user that the limit is infinity.  Perhaps there could be a special `limits.min.nnodes_E` or something to say output empty string if none set?

Another option is to create a new converter like `!I` that assumes inifinity if the output was an empty string?

I dunno, I'm going back and forth on this.  Welcome thoughts.

grondo:
> struggling with header for {defaults.timelimit} vs {limits.timelimit}, I did "TIMELIMIT*" for defaults for now. Any good suggestions?

Since it isn't a configured limit, but a default, maybe just `DEFAULTTIME`

grondo:
> I dunno, I like the idea that "inf" should be output by default to make it clear to the user that the limit is infinity.

FWIW, I'm fine with `inf` output by default! If it is causing consternation, let's just keep it that way for now and if necessary we can come back and try again in the future.

chu11:
need to add tests, but I sort of like where I'm at now.  We're going to go with the `i` special presentation character, which converts "inf" to "-", which I think is a good way to alert the user that there are no limits configured.

Example output:

```
QUEUE    DEFAULTTIME  TIMELIMIT     NNODES     NCORES      NGPUS
batch*            2h         4h       2-10      1-inf          -
debug             1h        30m        1-8        1-8        1-8
```

with the current default output of

```
                "?:{queuem:<8.8} {defaults.timelimit!F:>11i} {limits.timelimit!F:>10i} {limits.range.nnodes:>10i} "
                "{limits.range.ncores:>10i} {limits.range.ngpus:>10i}"
```

grondo:
That turned out pretty nice!

chu11:
re-pushed with some testing, removed WIP, I think this is in good shape for a review

perhaps one thing to debate is if there should be any other named format to support other than "default".  I can't think of anything obvious though.

src/bindings/python/flux/util.py
```diff
@@ -419,6 +419,9 @@ def format_field(self, value, spec):
             basecases = ("", "0s", "0.0", "0:00:00", "1970-01-01T00:00:00")
             value = "-" if str(value) in basecases else str(value)
             spec = spec[:-1] + "s"
+        elif spec.endswith("i"):
```
grondo, line 422: After seeing this duplicated code it took me a second to remember that we can't just handle `inf` via `"h"` because `i` is needed to specifically handle limits which may have a valid and presentable value of `0`. To remind us of this I'd suggest a comment here describing why `i` is being added.

Also, the code could be deduplicated a bit by handling both `h` and `i` in the same code block, e.g. something like
```diff
diff --git a/src/bindings/python/flux/util.py b/src/bindings/python/flux/util.py
index deb13746f..2cb42f42b 100644
--- a/src/bindings/python/flux/util.py
+++ b/src/bindings/python/flux/util.py
@@ -338,6 +338,13 @@ def fsd(secs):
 class UtilFormatter(Formatter):
     # pylint: disable=too-many-branches
 
+    # supported presentation characters which turn a base case
+    # value, e.g. '0' or an empty string, into a single hyphen.
+    hyphenators = {
+        "h": ("", "0s", "0.0", "0:00:00", "1970-01-01T00:00:00"),
+        "i": ("inf"),
+    }
+
     def convert_field(self, value, conv):
         """
         Flux utility-specific field conversions. Avoids the need
@@ -410,8 +417,8 @@ class UtilFormatter(Formatter):
             denote_truncation = True
             spec = spec[:-1]
 
-        if spec.endswith("h"):
-            basecases = ("", "0s", "0.0", "0:00:00", "1970-01-01T00:00:00")
+        if spec and spec[-1] in self.hyphenators:
+            basecases = self.hyphenators[spec[-1]]
             value = "-" if str(value) in basecases else str(value)
             spec = spec[:-1] + "s"
         retval = super().format_field(value, spec)
```

t/t2241-queue-cmd-list.t
```diff
@@ -0,0 +1,371 @@
...
+# duration = "4h"
+# job-size.min.nnodes = 2
+# job-size.max.nnodes = 10
+# job-size.min.ncores = 1
+# job-size.max.ncores = -1
+# job-size.min.ngpus = 1
+# # job-size.max.ngpus left out to test default of infinity
+#
+test_expect_success 'create policy defaults config for tests' '
+	cat <<-EOF > policy_defaults_config.in
```
grondo, line 99: you use `-EOF` here but do not indent contents.

t/t2241-queue-cmd-list.t
```diff
@@ -0,0 +1,371 @@
...
+# [queues.debug.policy.limits]
+# duration = "2h"
+# job-size.max.nnodes = 4
+# job-size.min.ncores = 2
+#
+# N.B. Note that many values in queues.debug.policy.limits left out to
+# test that system wide config is parsed correctly
+test_expect_success 'create queue defaults config for tests' '
+	cat <<-EOF > queues_config.in
+{
```
grondo, line 216: Wow, this makes me wonder if it would be easier to use a TOML config (in comments above) along with `flux config load` and let `flux queue list` read config from the broker?

grondo (approved):
This looks pretty nice! I just had some suggestions inline and a couple testing questions.
O/w this looks good to me.

t/t2241-queue-cmd-list.t
chu11: Yeah, i initially tried to use a linux command line toml->json converter, but didn't know if such a tool could safely be assumed to be installed.

To me, its under that threshold of annoyance to support reading config in toml format, like adding a `--from-stdin-toml` or whatever.  Or I guess I could try and detect toml vs json.

grondo: Oh, I just meant that the broker reads toml then returns the config in json, so why not use it since the test is running under Flux anyway (i.e. do not use `--from-stdin` at all). However, I may have missed what the test is doing.

chu11: ohhh just load it via toml normally.  yeah, i guess that would be cleaner :-)

chu11:
thanks for the review, re-pushed with fixes per comments

Issue #4861, opened 2023-01-17
safe mode should be more sophisticated


garlick:
Problem: when Flux rc1 detects that the instance crashed (by examining the startlog in the KVS), it enters a "safe mode", where all queues are disabled and stopped.  It's not intuitive how deep "safe mode" goes or how to get out of it.

To exit start mode, one currently must run: 
```
$ sudo flux queue enable --all
$ sudo flux queue start --all
```

Moreover, this might start/enable queues that were disabled/stopped before safe mode was entered.

One thought is that maybe Flux (on rank 0) should leave the queues alone and simply abort rc1 with instructions on how to run `flux start --recovery` and fix up the startlog.

PR #4898, opened 2023-01-30
eliminate "safe mode" after improper shutdown

garlick:
Problem: safe mode is confusing as noted in #4861, and more so now that we have multiple queues with state that persists across a reboot.

All an admin can really do after an improper shutdown is re-enable the queues and hope for the best.  This PR changes the behavior after an improper shutdown is detected.  Instead of stopping/disabling all queues, it just logs the improper shutdown and continues.  If there are troubles, this will be useful information to have in the logs.

### DIFF
--- a/etc/rc1
+++ b/etc/rc1
@@ -75,9 +75,7 @@ fi
 if test $RANK -eq 0; then
     if test "$(backing_module)" != "none"; then
         if ! flux startlog --check --quiet; then
-            flux queue stop --all
-            flux queue disable --all \
-                "Flux is in safe mode due to an incomplete shutdown."
+	    echo "Flux was not shut down properly.  Data may have been lost."
         fi
     fi
 fi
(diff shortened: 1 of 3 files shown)


etc/rc1
```diff
@@ -75,9 +75,7 @@ fi
 if test $RANK -eq 0; then
     if test "$(backing_module)" != "none"; then
         if ! flux startlog --check --quiet; then
-            flux queue stop --all
-            flux queue disable --all \
-                "Flux is in safe mode due to an incomplete shutdown."
+	    echo "Flux was not shut down properly.  Data may have been lost."
```
grondo, line 78: commit message typo: s/when flux is not restart cleanly/when flux is not restarted cleanly/

grondo (approved):
LGTM!

garlick:
Thanks! I fixed that typo and will set MWP.

Issue #4900, opened 2023-01-31
job-manager: alloc-bypass plugin isn't working


grondo:
@jameshcorbett noticed that the `alloc-bypass.so` plugin no longer works. A simple demo is sufficient to show the problem:

```console
$ flux mini submit --setattr=alloc-bypass.R="$(flux R encode -r 0 -c 0-3)" sleep inf
ƒETCj3cB
2023-01-31T01:16:38.554415Z job-manager.err[0]: sched.alloc-response: id=512090963968 already allocated
$ flux resource list
     STATE NNODES   NCORES    NGPUS NODELIST
      free      4       15        0 asp,asp,asp,asp
 allocated      1        1        0 asp
      down      0        0        0 
```

It appears that the alloc-bypass job is still resulting in an alloc request to the scheduler, resulting in one core allocated (which can now never be freed)

Also, clearly our tests of alloc-bypass were worthless so let's fix that too :facepalm:

grondo:
Ok, the problem here is because the `set-flags` event that the `alloc-bypass.so` plugin uses to set the alloc-bypass flag is posted in the `job.state.sched` callback. This used to work when job events were handled directly by the function that posted them. However, this recursion was fixed in de9d43460f5e50bcaef3338b295bb6df0083ccc4 and so I think the alloc-bypass plugin has been broken since then!.

The easiest fix is to just post the set-flags event in the `validate` callback once the job is considered valid. I can't think of any issues that will cause.

grondo:
And the existing test doesn't fail because it first starts a normal job that allocates all resources. Therefore the alloc-bypass job _does not_ have resources allocated by the scheduler, since the incorrect alloc request is still sitting in its queue.

I'll add a test that simulates what @jameshcorbett was doing: run the alloc-bypass job first, _then_ try to run a normal job that allocates all resources.

PR #4901, opened 2023-01-31
job-manager: fix alloc-bypass plugin

grondo:
This fixes the issue with the `alloc-bypass` plugin described in #4900 and fixes the test to catch this error or errors like this in the future.


garlick (approved):
LGTM!

grondo:
Thanks! Setting MWP.

Issue #4899, opened 2023-01-30
inconsistency with --no-header vs --suppress-header


chu11:
It seems we do `--suppress-header` with `flux-jobs` and `flux-pgrep`.  We do `--no-header` w/ `flux-resource` and `flux-pstree`.

Would be easy enough to support backwards compatibility through a HIDDEN option.

Any strong opinions on which one to keep?

I like "--no-header" better, since the short option is `-n` anyways.

PR #4903, opened 2023-01-31
flux-jobs/flux-pgrep: rename --suppress-header to --no-header

chu11:
Per #4899, make the `--suppress-header` / `--no-header` consistent between tools.  Chose to common-ify on `--no-header`.  

Note that I legacy support `--suppress-header` in `flux-jobs` but not in `flux-pgrep`, b/c the latter has not officially been released yet, so I assumed no users of it yet.

### DIFF
--- a/src/cmd/flux-jobs.py
+++ b/src/cmd/flux-jobs.py
@@ -256,9 +256,16 @@ def parse_args():
     )
     parser.add_argument(
         "-n",
-        "--suppress-header",
+        "--no-header",
         action="store_true",
         help="Suppress printing of header line",
+    )
+    # --suppress-header, older legacy name
+    parser.add_argument(
+        "--suppress-header",
+        action="store_true",
+        dest="no_header",
+        help=argparse.SUPPRESS,
     )
     parser.add_argument(
         "-u",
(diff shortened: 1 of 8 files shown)


src/cmd/flux-jobs.py
```diff
@@ -256,10 +256,16 @@ def parse_args():
...
         "-n",
-        "--suppress-header",
+        "--no-header",
         action="store_true",
         help="Suppress printing of header line",
     )
+    # --suppress-header, older legacy name
+    parser.add_argument(
+        "--suppress-header",
+        action="store_true",
```
grondo, line 266: you can use the [dest](https://docs.python.org/3/library/argparse.html#dest) option to `add_argument` to store true in `no_header` instead of `suppress_header` so that this just becomes a hidden alternate for `--no-header`, e.g.

```python
parser.add_argument(
    "--suppress-header",
    action="store_true",
    dest="no_header",
    help=argparse.SUPPRESS
)
```

Then remove references to `args.suppress_header` everwhere.

grondo (commented):
Looks good. One suggested change inline.

chu11:
thanks!  re-pushed with the tiny cleanup and setting MWP

grondo (approved):
LGTM! (you can put merge-when-passing on a PR without approving review since the approval is one of the things mergify checks before merging)

PR #4905, opened 2023-02-01
flux-resource: increase width of queue field

### LINKED ISSUE #4904: flux resource list - give QUEUE a little more space?
I'm not sure if "debug", "batch" and "all" are likely to become defacto standard queue names in LC, but if so, could we spare _one more character_ in the queue column? :smile:
```
$ flux resource list
     STATE QUEUE    PROPERTIES NNODES   NCORES    NGPUS NODELIST
      free debug,al                 1        4        0 picl1
      free debug,al 4g              1        4        0 picl2
      free batch,al                 3       12        0 picl[3-4,6]
      free batch,al 4g              1        4        0 picl5
      free batch,al testprope+      1        4        0 picl7
 allocated                          0        0        0 
      down                          0        0        0 
```
Edit: also we are flagging overflow with "+" in the properties column but apparently not in the QUEUE column.  Should that be fixed?

chu11:
Problem: Unlike other tools such as flux-queue and flux-jobs, it is possible for multiple queues to be listed with the "flux resource list" "queue" field.  The default output width is 8 characters, which is a little short for multiple queues.

Increase from 8 characters to 10 characters.

Fixes #4904

---

We can debate the number of chars, I just didn't want to go too big b/c the common case is of course 1 queue.

Also, I elected not to make the heading "QUEUES" (plural).  The field name is "queue" and I figure one queue is still the common case.

Edit: or we could go with QUEUE(S)?

### DIFF
--- a/src/cmd/flux-resource.py
+++ b/src/cmd/flux-resource.py
@@ -56,7 +56,7 @@ class FluxResourceConfig(UtilConfig):
         "default": {
             "description": "Default flux-resource list format string",
             "format": (
-                "{state:>10} ?:{queue:<8.8} ?:{propertiesx:<10.10+} {nnodes:>6} "
+                "{state:>10} ?:{queue:<10.10} ?:{propertiesx:<10.10+} {nnodes:>6} "
                 "{ncores:>8} {ngpus:>8} {nodelist}"
             ),
         },


grondo (approved):
Thanks!

Issue #4906, opened 2023-02-02
job-list will not ingest alloc bypass jobs


garlick:
Problem: @jameshcorbett ran into a problem on hechy when using the alloc-bypass plugin:

```
Feb 01 15:18:46 hetchy7 flux[250503]: job-list.err[0]: eventlog_restart_parse: alloc context for 149949730573516800 invalid
Feb 01 15:18:46 hetchy7 flux[250503]: job-list.err[0]: initialization from kvs error: Protocol error
Feb 01 15:18:46 hetchy7 flux[250503]: job-list.crit[0]: module exiting abnormally
```

The offending eventlog was
```
[flux@hetchy7:/]$ flux job eventlog 149949730573516800
1675288442.910732 submit userid=763 urgency=16 flags=0 version=1
1675288442.924381 validate
1675288442.936043 depend
1675288442.936067 priority priority=16
1675288442.936102 set-flags flags=["alloc-bypass"]
1675288442.937575 alloc bypass=true
1675288442.937724 prolog-start description="job-manager.prolog"
1675288442.937733 prolog-start description="cray-pals-port-distributor"
1675288442.941202 prolog-finish description="cray-pals-port-distributor" status=0
1675288443.432181 prolog-finish description="job-manager.prolog" status=0
1675288443.435883 start
1675288445.517003 finish status=256
1675288445.517265 epilog-start description="job-manager.epilog"
1675288445.517305 free
1675288445.523132 release ranks="all" final=true
1675288445.523173 clean
```

Using `flux start --recovery` to remove the job from the KVS was the solution.

grondo:
This wasn't caught by the testsuite because the problem only comes up when the `job-list` is replaying jobs from the kvs. Reloading `job-list` at the end of `t2273-job-alloc-bypass.t` reproduces the issue. I wonder why it works for jobs via the journal but not from the kvs?

In any case, I think the jobtap system allows plugins to add arbitray context to any job event, so the job-list module should be more permissive here. Maybe a good test to add would be to have a jobtap plugin that adds context to every event and make sure nothing goes :boom:.

grondo:
Looks like job-list requires that if the alloc event has a context object, it contains `annotations`:

https://github.com/flux-framework/flux-core/blob/604eb888344a169df7d03b6d96f2a47b3cf68b58/src/modules/job-list/job_state.c#L601-L611

[...]

PR #4907, opened 2023-02-02
job-list: do not assume alloc event context always contains annotations

chu11:
Per discussion in #4906

I only added a regression test, no additional wider coverage tests yet per discussion in #4906.

Hopefully will have those wider extra coverage tests in soon.  But just in case that extra wider coverage can't be completed in a timely manner, just wanted to make sure we atleast get this fix in before the next release, we can always add the extra tests later.

### DIFF
--- a/src/modules/job-list/job_state.c
+++ b/src/modules/job-list/job_state.c
@@ -602,15 +602,10 @@ static struct job *eventlog_restart_parse (struct job_state_ctx *jsctx,
             /* context not required if no annotations */
             if (context) {
                 json_t *annotations;
-                if (!(annotations = json_object_get (context, "annotations"))) {
-                    flux_log (jsctx->h, LOG_ERR,
-                              "%s: alloc context for %ju invalid",
-                              __FUNCTION__, (uintmax_t)job->id);
-                    errno = EPROTO;
-                    goto error;
+                if ((annotations = json_object_get (context, "annotations"))) {
+                    if (!json_is_null (annotations))
+                        job->annotations = json_incref (annotations);
                 }
-                if (!json_is_null (annotations))
-                    job->annotations = json_incref (annotations);
             }
 
             if (job->state == FLUX_JOB_STATE_SCHED)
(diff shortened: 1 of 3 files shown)


grondo:
Sounds good to me!

chu11:
added a coverage test on top, instead of the dump file or jobtap plugin approach as mentioned in #4904, I just "hand" wrote an eventlog over an existing one.  That way we can easily add extra events in the future (vs the dump file approach).

t/issues/t4906-job-list-alloc-bypass.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash -e
```
grondo, line 1: Since you have the follow on commit which adds a regression test to the job-list module, maybe this separate regression test isn't required? The reason I ask is that the t4000-issues test runs all the reproducers serially and at this point is really slow, so avoiding another `flux start` test therein would keep the testsuite a little leaner.

chu11: actually the follow on test doesn't test it perfectly, as the regression that tests the event `alloc {"bypass":1}`, whereas the new test added `alloc {"annotations}:<something>, "etc": 1}`.

But I could tweak it accordingly.

grondo: Ah, that might be good just to save some unnecessary work for the issues test driver.

chu11: ahhh this was a good idea, ended up there was a corner case with exception notes (which are optional).

chu11:
re-pushed, removing the regression test and adding a new test in t2260-job-list.t instead.  Decided to add another test to check for optional context fields.  The only other one I found was userid/note in exceptions, and it ended up there was a corner case in job-list for that case, so fixed that too.

grondo (approved):
This LGTM. Nice job catching the exception note corner case!

Issue #4234, opened 2022-03-21
job-archive: store jobspec/R as JSON type in sqlite when available


grondo:
Recent versions of sqlite have embedded support for [JSON](https://www.sqlite.org/json1.html) via the `JSON` column type. Using this type allows queries to reference arbitrary keys in the stored JSON, which sounds pretty appealing if we can one day open up access to the job-archive database for more advanced queries.

As a test, here's a Python script which will ingest all the current jobs in an instance and store them in a local `test.db` using JSON columns for `annotations`, `eventlog`, `jobspec` and `R`.

```python
#!/usr/bin/python3

import sys
import sqlite3
import json
from datetime import datetime, timedelta

import flux
from flux.job import JobList
from flux.hostlist import Hostlist

handle = flux.Flux()
alljobs = JobList(handle, attrs=flux.job.list.VALID_ATTRS, max_entries=0)

conn = sqlite3.connect("test.db")
c = conn.cursor()
c.execute(
    "CREATE TABLE if not exists jobs("
    "  id CHAR(16) PRIMARY KEY,"
    "  userid INT,"
    "  ranks TEXT,"
    "  t_submit REAL,"
    "  t_run REAL,"
    "  t_cleanup REAL,"
    "  t_inactive REAL,"
    "  name TEXT,"
    "  nnodes INT,"
    "  ntasks INT,"
    "  waitstatus INT,"
    "  success BOOLEAN,"
    "  result TEXT,"
    "  annotations JSON,"
    "  eventlog JSON,"
    "  jobspec JSON,"
    "  R JSON"
    ")"
)

insert_sql = """
    INSERT INTO
      jobs(
        id,
        userid,
        ranks,
        t_submit,
        t_cleanup,
        t_inactive,
        name,
        nnodes,
        ntasks,
        waitstatus,
        success,
        result,
        annotations,
        eventlog,
        R,
        jobspec)
      VALUES(?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?)
"""


def lookup_cb(future, job):
    try:
        result = future.get()
        job.jobspec = result["jobspec"]
        job.R = result["R"]
        job.eventlog = result["eventlog"]
    except Exception as exc:
        print(f"{job.id}: {exc} (ignoring)")
        pass


jobs = []
for job in alljobs.jobs():
    jobs.append(job)
    keys = [ "jobspec", "eventlog" ]
    if job.nnodes:
        keys.append("R")
    handle.rpc(
[...]

Issue #3066, opened 2020-07-22
flux jobs:  sorting support with configurable lexicographical order


dongahn:
This ticket is created to put in a request to add sorting support within flux jobs.

As Fluxion scheduler will pass important, yet-scheduler-specific info ("which queue each job is running") through the annotation API, it would be great to customize the sorting order of flux-jobs. 

One immediate use case would be for multi-queue support. 

Under multi queue setting (e.g., system instance) it would be great to configure flux jobs such a  way that it sorts its jobs in queue name order first so that we can see jobs in batch queue first before debug queue or vice versa. 

In another case, users may want to sort the jobs by queue name first and then by start time estimate...

The general solution seems like giving users and/or admin a way to set lexicographical sorting order of the display fields.

grondo:
Would you also want to only fetch the top N pending jobs from each "queue"? `flux-jobs` by default is going to fetch the "top" N jobs overall, these may include all jobs from one queue and none of the other...

chu11:
i think we can declare this closed with #6380

PR #4908, opened 2023-02-02
add `flux job last`

garlick:
I thought these trivial commands would be useful.  `flux-job last` prints the job ID of the most recently submitted job.  `flux job history` prints the job ID and "name" for the most recently submitted 10 jobs.
```
$ flux job last
flux-job: job history is empty
$ flux mini submit -N1 true
ƒ7bbUqDZZ
$ flux job last
ƒ7bbUqDZZ
$ flux job history
ƒ7bbUqDZZ	true
ƒ4a15vsCs	hostname
ƒ4Zj7Lc5M	hostname
ƒ4ZXi88PZ	hostname
ƒ3HPLDNQ7	hostname
ƒoFzsGdV	sleep
ƒhtPAE7h	hostname
```
I couldn't figure out how to get a submission-ordered list of jobs (including all states) out of `job-list` so I added a little jobtap plugin that tracks the last 10 job ids for each user and registers a service that returns the list for the asking user.

I don't know if others will find this useful, so before I did any real work writing tests and such I thought I'd run it up the flag pole.

The choice to write a jobtap plugin might be controversial since probably the real solution is to allow more general queries of the  data stored in job-list.  Or maybe I'm missing some other obvious way to get this data.  However, writing jobtap plugins is fun, and it didn't take much effort to make this work (IOW it could be considered a throw-away if a real query solution emerged)

### DIFF
--- a/src/cmd/flux-job.c
+++ b/src/cmd/flux-job.c
@@ -103,6 +103,7 @@ int cmd_memo (optparse_t *p, int argc, char **argv);
 int cmd_purge (optparse_t *p, int argc, char **argv);
 int cmd_taskmap (optparse_t *p, int argc, char **argv);
 int cmd_timeleft (optparse_t *p, int argc, char **argv);
+int cmd_last (optparse_t *p, int argc, char **argv);
 
 int stdin_flags;
 
@@ -575,6 +576,13 @@ static struct optparse_subcommand subcommands[] = {
       cmd_timeleft,
       0,
       timeleft_opts,
+    },
+    { "last",
+      "SLICE",
+      "List my most recently submitted job id(s)",
+      cmd_last,
+      0,
+      NULL,
     },
     { "purge",
       "[--age-limit=FSD] [--num-limit=N]",
@@ -3723,6 +3731,61 @@ int cmd_timeleft (optparse_t *p, int argc, char **argv)
             sec = 1;
         printf ("%lu\n", sec);
     }
+    flux_close (h);
+    return 0;
+}
+
+int cmd_last (optparse_t *p, int argc, char **argv)
+{
+    int optindex = optparse_option_index (p);
+    flux_future_t *f;
+    flux_t *h;
+    char buf[32];
+    json_t *jobs;
+    size_t index;
+    json_t *entry;
+    const char *slice = "[:1]";
+    char sbuf[16];
+
+    if (optindex < argc) {
+        slice = argv[optindex++];
+        // if slice doesn't contain '[', assume 'flux job last N' form
+        if (!strchr (slice, '[')) {
+            errno = 0;
+            char *endptr;
+            int n = strtol (slice, &endptr, 10);
+            if (errno != 0 || *endptr != '\0') {
+                optparse_print_usage (p);
+                exit (1);
+            }
+            snprintf (sbuf, sizeof (sbuf), "[:%d]", n);
+            slice = sbuf;
+        }
+    }
+    if (optindex < argc) {
+        optparse_print_usage (p);
+        exit (1);
+    }
+    if (!(h = flux_open (NULL, 0)))
+        log_err_exit ("flux_open");
+    if (!(f = flux_rpc_pack (h,
+                             "job-manager.history.get",
+                             FLUX_NODEID_ANY,
+                             0,
+                             "{s:s}",
+                             "slice", slice))
+        || flux_rpc_get_unpack (f, "{s:o}", "jobs", &jobs) < 0) {
+        log_msg_exit ("%s", future_strerror (f, errno));
+    }
+    if (json_array_size (jobs) == 0)
+        log_msg_exit ("job history is empty");
+    json_array_foreach (jobs, index, entry) {
+        flux_jobid_t id = json_integer_value (entry);
+        if (flux_job_id_encode (id, "f58", buf, sizeof (buf)) < 0)
+            log_err_exit ("error encoding job ID");
+        printf ("%s\n", buf);
+    }
+    flux_future_destroy (f);
     flux_close (h);
     return 0;
 }
--- a/src/common/libutil/Makefile.am
+++ b/src/common/libutil/Makefile.am
@@ -97,6 +97,8 @@ libutil_la_SOURCES = \
 	fileref.h \
 	hola.c \
 	hola.h \
+	slice.c \
+	slice.h \
 	strstrip.c \
 	strstrip.h
 
@@ -130,7 +132,8 @@ TESTS = test_sha1.t \
 	test_errprintf.t \
 	test_fileref.t \
 	test_hola.t \
-	test_strstrip.t
+	test_strstrip.t \
+	test_slice.t
 
 test_ldadd = \
 	$(top_builddir)/src/common/libutil/libutil.la \
@@ -273,3 +276,7 @@ test_fileref_t_LDADD = $(test_ldadd)
 test_hola_t_SOURCES = test/hola.c
 test_hola_t_CPPFLAGS = $(test_cppflags)
 test_hola_t_LDADD = $(test_ldadd)
+
+test_slice_t_SOURCES = test/slice.c
+test_slice_t_CPPFLAGS = $(test_cppflags)
+test_slice_t_LDADD = $(test_ldadd)
--- a/src/common/libutil/hola.c
+++ b/src/common/libutil/hola.c
@@ -287,6 +287,31 @@ void *hola_list_next (struct hola *hola, const void *key)
     return item;
 }
 
+void *hola_list_prev (struct hola *hola, const void *key)
+{
+    void *item = NULL;
+
+    if (hola && key) {
+        zlistx_t *l;
+        if ((l = zhashx_lookup (hola->hash, key)))
+            item = zlistx_prev (l);
+    }
+    return item;
+}
+
+void *hola_list_last (struct hola *hola, const void *key)
+{
+    void *item = NULL;
+
+    if (hola && key) {
+        zlistx_t *l;
+        if ((l = zhashx_lookup (hola->hash, key)))
+            item = zlistx_last (l);
+    }
+    return item;
+}
+
+
 void *hola_list_cursor (struct hola *hola, const void *key)
 {
     void *handle = NULL;
--- a/src/common/libutil/hola.c
+++ b/src/common/libutil/hola.c
@@ -24,6 +24,7 @@ struct hola {
     zlistx_t *keys; // for iteration
     zlistx_destructor_fn *list_destructor;
     zlistx_duplicator_fn *list_duplicator;
+    zlistx_comparator_fn *list_comparator;
     unsigned int keys_valid:1;
     unsigned int flags;
 };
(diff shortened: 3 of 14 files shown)


chu11:
> The choice to write a jobtap plugin might be controversial since probably the real solution is to allow more general queries of the data stored in job-list.

Just linking to this issue,which I suddenly remembered

https://github.com/flux-framework/flux-core/issues/3066

chu11:
> The choice to write a jobtap plugin might be controversial since probably the real solution is to allow more general queries of the data stored in job-list. 

Thinking about it a bit, there would be a number of corner cases trying to get last N jobs submitted from `job-list`, mostly surrounding purged jobs and probably some trickiness with jobs that are running or inactive (or if they are all inactive).  So this seems like a sensible approach.  I'll just mention #4336, which could solve the inactive jobs case if we ever believe we need to go down the job-db path soon.

A dumb alternative is if we wanted to implement this at the `flux mini` or python API level, where job ids are cached in the user's home directory, like `.bash_history`, but `.flux_jobs_history` or something instead.  With this approach its simpler, less corner case prone than querying `job-list`, but not all job submissions would probably support this mechanism.

grondo:
> A dumb alternative is if we wanted to implement this at the flux mini or python API level, where job ids are cached in the user's home directory, like .bash_history, but .flux_jobs_history or something instead. With this approach its simpler, less corner case prone than querying job-list, but not all job submissions would probably support this mechanism. 

Unfortunately, that can't work because a user can have many flux instances active at once across a center, a cluster, or even within a job.

> Thinking about it a bit, there would be a number of corner cases trying to get last N jobs submitted from job-list, mostly surrounding purged jobs and probably some trickiness with jobs that are running or inactive (or if they are all inactive). 

I think it would be helpful if job-list could be given a sort order to apply to `max_entries` such that `flux jobs -no {id} -c 10` would give you your last 10 submitted jobs, instead of whatever the order is now. I think this would solve 99% of cases, if one of your last 10 jobs has been purged then you're in trouble if you wanted to do something with it anyway.

garlick:
I was secretly thinking "if job-list accepted SQL..."

Then stuff like [this](https://www.ibm.com/support/pages/how-select-last-n-rows-table-ascending-order) is possible.

grondo:
My first reaction is that if we need a `flux job history` command, then something in `flux jobs -a` is broken. Though there is something appealing in these simple commands, I can imagine someone will want to get the 11th job, or will ask for `flux jobs` like fields to be available in `flux job history`, etc.

Could we take a look at what it would take for `flux jobs -ano {id} -c 10` to return the last 10 submitted jobs? I'm running into similar problems with the advanced `flux pgrep` queries...

As we've discussed before, long term it would be best if job-list used a db internally. Then perhaps job-list could assign a per-user monotonically increasing jobid to each job to store in the database so users can query their nth job easily. If the maxid was also stored for each user, the jobids could be indexed from the end as well (give me the 5th most recent job I submitted).

chu11:
> I think it would be helpful if job-list could be given a sort order to apply to max_entries such that flux jobs -no {id} -c 10 would give you your last 10 submitted jobs, instead of whatever the order is now. I think this would solve 99% of cases, if one of your last 10 jobs has been purged then you're in trouble if you wanted to do something with it anyway.

Yeah, I could have swore we had an issue about this but I can't find it.  #3066 is the closest one I can find (it appears more specific to `flux-jobs` than `job-list`).  My recollection is @dongahn was looking for a way to sort jobs by time remaining until job will be run (so it'd be sorting `annotations.t_remaining` or something).

Perhaps we should restart said discussion.  Dunno if we'd want to go down the path of some mega-query syntax or something far simpler.

grondo:
> I was secretly thinking "if job-list accepted SQL..."

This would be ideal. Recent versions of sqlite support JSON as well as shown in #4234.

grondo:
Another only slightly related idea was to have job-list support RFC 31 style constraint objects. This is what I'm using in the full `flux pgrep` implementation, and theres a nice query syntax (RFC 35) for it. For now `flux pgrep` just pulls down `max_entries` jobs and filters locally, but perhaps there is a way to turn an RFC 31 Constraint object into a SQL query...

garlick:
I feel like a simple way to get the most recently submitted job would be pretty helpful.  I'm more ambivalent about `flux job history` - I guess you're right it duplicates `flux jobs` and should go away.

Just a thought  but would it make sense to modify this jobtap plugin to track all jobs and then change `flux job last` so that it takes an argument of
- (none) print the last submitted job id
- `N` print the nth submitted job id
- `-N` print the nth ago submitted job id
- `M-N` print the m through nth submitted job ids

Then we essentially have the monotonic integer jobid.  You just have to throw a `$(flux job last 42)`  on the command line where the job ID goes...

Once `job-list` could return a submit ordered list of job IDs that is unaffected by purging then the jobtap plugin could go away.

chu11:
> Another only slightly related idea was to have job-list support RFC 31 style constraint objects. This is what I'm using in the full flux pgrep implementation, and theres a nice query syntax (RFC 35) for it. For now flux pgrep just pulls down max_entries jobs and filters locally, but perhaps there is a way to turn an RFC 31 Constraint object into a SQL query...

I like this idea, it would give us probably 99% of what we want but not go full on SQL-ish. 

One thought is we'd need an ascending/descending thingy in the query, although that could also be a flag outside of the query.

grondo:
Yes, this is more compelling without `flux job history` IMO. Especially if the jobtap plugin can track all jobs. I'm still a bit unclear about what would happen in the jobtap plugin after a purge. 

> I feel like a simple way to get the most recently submitted job would be pretty helpful

If that is the main use case, the solution could be simplified to keep only the last jobid for each user and `flux job last` could take no arguments. I'm guessing there was a use case also for getting the last N jobs too, in which case I find the `flux job last N` syntax to get the Nth job a bit awkward. What if we added two new commands `flux job last ` and `flux job range`, where `flux job last` is a synonym for `flux job range` but indexes from the end of the submitted jobs list?

We could also consider using Python's [slice](https://python-reference.readthedocs.io/en/latest/docs/brackets/slicing.html)
syntax for ranges, which is established and might be more familiar to many users, and allows a negative _step_ to start from the end e.g.

 * `flux job last` is a synonym for `flux job range -1` and returns the last job
 * `flux job last N` is a synonym for `flux job range -N:` and returns the Nth most recently submitted job
 * `flux job range N` returns the Nth submitted job (if 1 origin, then that is different than Python, but may make more sense here)
 * `flux job range :N` returns the _first_ N jobs
 * `flux job range N:M` gives the jobs in range N to M
 * `flux job range N:` returns jobs from job N to the end of the list

Perhaps this syntax could be added to `flux pgrep/pkill` as well which could become another client of this jobtap plugin. Just throwing out ideas that would make this a little more palatable imo.

garlick:
On purging in the jobtap plugin, I was originally thinking we could ignore it and return purged IDs which would then fail if used, but in retrospect, that was not a great plan.  Idea: register a destructor on the list entry using `flux_jobtap_job_aux_set()`.  When the job goes away, the list entry could be removed.

While it's straightforward to assign an integer alias to jobs, preserving the mapping across an instance restart (given purging and non-deterministic KVS traversal order) would require that we store the mappings in the job eventlog and job manager checkpoint.  Would we require that the mappings never change once assigned?

Assuming we need persistent mappings, I'm thinking either that should be out of scope for this PR and this should just add `flux job last` where only the relative position is needed, OR we nix this as a failed experiment.

grondo:
> Assuming we need persistent mappings,

I'm not sure we need persistent mappings, though I haven't come up with a use case for this feature yet, so may be missing something.

Maybe it would be simplest to start with just `flux job last` but allow getting the last N jobs or M-N jobs, then you do not have to worry about persistent mappings (did I get that right?). Does that satisfy the use case? I feel like we should actually be discouraging the use case of "i want the nth job I ever submitted on this cluster" and instead add job search and query tools that satisfy these kinds of requirements.

garlick:
> Maybe it would be simplest to start with just flux job last but allow getting the last N jobs or M-N jobs, then you do not have to worry about persistent mappings (did I get that right?). Does that satisfy the use case?

Yep, it satisfies my use case.  

> I feel like we should actually be discouraging the use case of "i want the nth job I ever submitted on this cluster" and instead add job search and query tools that satisfy these kinds of requirements.

Works for me.

garlick:
This latest incarnation of `flux job last` accepts an optional python slice expression as discussed above.  The slice applies to the user's list of submitted jobs in reverse order (newest first) with an origin of zero.  `flux job last` is alias for `flux job [:1]` (list most recent job); `flux job last [:]` lists all the user's jobs, newest first; and  `flux job last [::-1]` lists all the user's jobs, oldest first.  It's a pretty flexible way to query an ordered list - neat idea @grondo!

The list is currently not purged, so that will need to be handled before this should be considered for merging, so I'll leave the WIP on there.

src/cmd/flux-job.c
```diff
@@ -576,6 +577,13 @@ static struct optparse_subcommand subcommands[] = {
       0,
       timeleft_opts,
     },
+    { "last",
+      "SLICE",
+      "List my most recently submitted job ID(s)",
```
grondo, line 582: Other subcommands in `flux job` reference jobids as `jobid(s)` or `job id(s)`. Instead of adding a third way, maybe we can settle on one (my preference is "jobid" "jobids". This is consistent with the `flux-job(1)` man page as well, though I haven't checked the rest of docs, --help, etc.)

grondo (commented):
Looking good to me. I had one really minor comment so far.

I also wonder if the command should allow the brackets to be optional on a slice expression such that an arg with `:` but not starting with `[` is wrapped in `[]` on behalf of the user. This would allow the slice syntax to be used without quoting to avoid glob match (I had a file `1` in my directory for instance)

garlick:
Thanks - I added a bit more testing, and fixed up the usage message for `flux job last`.

I couldn't quite decide if I'm comfortable dropping the requirement of parens around slices.  Could we maybe add that later if it becomes a problem?

Dropping the WIP as I think this is probably close to ready.

grondo:
> I couldn't quite decide if I'm comfortable dropping the requirement of parens around slices. Could we maybe add that later if it becomes a problem?

Yep, that's fine. We have the same issue with hostlists (though much less likely to hit a glob match when prefixed with a hostname than matching a file named `1`). I did hit this issue immediately, but lesson learned.

grondo:
Though if it makes it more palatable (for the future) the Python slice syntax is probably technically `start:stop:step` which _may_ occur inside an sequence index operator `[ ]`. So no reason the `[]` should be required in order to call it an implementation inspired by slice syntax.

It does kind of feel like since `[]` is supported with `flux job last` then `flux job last [2]` should give you the 3rd most recent job. But I think that can wait for later as well, since the main use case is to get the last job or last N jobs.

grondo (approved):
This looks good to me!

grondo:
Setting MWP (I swear I have permission from @garlick)

garlick:
Good points.   I just needed a little convincing I guess.

> It does kind of feel like since [] is supported with flux job last then flux job last [2] should give you the 3rd most recent job.

Doh!  Why didn't I think of that.  Duh.

Well perhaps for the next release.

PR #4913, opened 2023-02-03
flux-resource: mark drained+offline nodes with asterisk

### LINKED ISSUE #4909: flux resource drain should indicate if node is also offline
Problem: sys admins require a way to see at a glance which drained nodes are running flux vs not.

In LC, nodes do not start flux until they have a clean bill of health.  Thus a drained that is offline is one that is not fixed, and a drained node that is online is one that has been fixed and can be undrained. 

It would be a help to the sys admin workflow if `flux resource drain` denoted which nodes are offline.

Here picl4 is both drained and turned off:
```
$ flux resource list
     STATE QUEUE    PROPERTIES NNODES   NCORES    NGPUS NODELIST
      free debug,al                 1        4        0 picl1
      free debug,al 4g              1        4        0 picl2
      free batch,al                 2        8        0 picl[3,6]
      free batch,al 4g              1        4        0 picl5
      free batch,al testprope+      1        4        0 picl7
 allocated                          0        0        0 
      down batch,al                 1        4        0 picl4
```
I think what's being requested is for offline to be incorporated into the `flux resource drain` output?
```
$ flux resource drain
TIME         STATUS   REASON                         NODELIST
Feb02 18:19  drained                                 picl4
```

grondo:
Proposing this a bit early so that @garlick can test on his system and give feedback.
This is the simplest solution to #4909, so we can sneak it into the next release.

Problem: It is difficult to tell which drained nodes are offline vs not in the output of flux-resource drain and status.

Take a simple step at first and flag offline drained nodes with an asterisk, i.e. `drained*`, in flux-resource `status` and `drain` output.

Update the expected output in one test.

Fixes #4909

### DIFF
--- a/doc/man1/flux-resource.rst
+++ b/doc/man1/flux-resource.rst
@@ -37,6 +37,7 @@ A few notes on drained nodes:
 - In ``flux resource status`` and ``flux resource drain``, the drain state
   of a node will be presented as "drained" if the node has no job allocations,
   and "draining" if there are still jobs running on the node.
+- If a node is drained and offline, then "drained*" will be displayed.
 
 Some further background on resource service operation may be found in the
 RESOURCE INVENTORY section below.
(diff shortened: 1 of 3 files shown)


chu11:
high level comment, would it make more sense to create a new field like `{statem}` that marks states in a special way that's different than `{state}`?  That's what was done with `flux queue list` and the `{queue}` vs `{queuem}` field.

garlick:
Just did a quick test of this on my system and looks good!
```
$ ./flux resource drain
TIME         STATUS   REASON                         NODELIST
Feb03 05:33  drained  testing                        picl5
Feb02 18:19  drained*                                picl4
$ ./flux resource status
    STATUS NNODES NODELIST
     avail      5 picl[1-3,6-7]
   offline      1 picl4
   exclude      1 picl0
   drained      1 picl5
  drained*      1 picl4
```

grondo:
Thanks! I'll set MWP.

doc/man1/flux-resource.rst
```diff
@@ -61,15 +62,15 @@ COMMANDS
...
-   be forced with the ``--skip-empty`` option.
+   Show system view of resources.  The *-n,--no-header* suppresses
+   header from output, *-o,--format=FORMAT* customizes output formatting
+   (see below), and *-s,--states=STATE,...* limits output to specified
+   resource states, where valid states are "online", "offline", "avail",
+   "exclude", "draining", "drained", "drained*", and "all". The special
+   "drain" state is also supported, and selects both draining and drained
+   resources. Normally, ``flux resource status`` skips lines with no
+   resources, unless the ``-s, --states`` option is used. Suppression of
+   empty lines can always be forced with the ``--skip-empty`` option.
```
chu11, line 73: should we mention what `drain*` means?

grondo: Hm, none of the states are described in this section. The description of `drained*` is in the section that describes drained nodes. We could reference that here?

chu11: ahhh, i missed the description of `drained*` in the other section.  This is fine then.  Sorry for the noise :P

grondo: NP, a pretty good suggestion IMO

PR #4923, opened 2023-02-07
flux-mini: do not combine hostlist/ranks terms joined by AND

grondo:
The flux-mini RFC 35 constraint parser inadvertently added `hostlist` and `ranks` operators to the `combinded_terms` set in the  parser, which means that `ranks:1 and ranks:2`is joined to `ranks:1-2`. But `ranks:1-2` means the rank can be in the set `1-2`, which means they are joined by OR not AND.

Remove `hostlist` and `ranks` from the `combined_terms` set to avoid this incorrect behavior.


grondo:
Thanks, setting MWP.

PR #4924, opened 2023-02-07
suggest use of --force with `flux resource drain` when target is already drained

grondo:
This PR simply adds a note to try `--force` in `flux resource drain` when the target is already drained, in case the user wants to update the drain reason.

### DIFF
--- a/t/t2311-resource-drain.t
+++ b/t/t2311-resource-drain.t
@@ -67,6 +67,13 @@ test_expect_success 'drain detects subset of already drained targets' '
 	test $(flux resource status -s drain -no {nnodes}) -eq 1
 '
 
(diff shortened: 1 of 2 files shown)


t/t2311-resource-drain.t
```diff
@@ -67,6 +67,13 @@ test_expect_success 'drain detects subset of already drained targets' '
 	test $(flux resource status -s drain -no {nnodes}) -eq 1
 '
 
+test_expect_success 'drain suggests --force with existing reason' '
```
chu11, line 70: commit message typo "outout"

chu11 (approved):
LGTM, just a commit message typo

PR #4925, opened 2023-02-07
build: avoid continuous regeneration of constraint/parsetab.py

grondo:
Problem: The Python ply module is smart enought to avoid regenerating parsetab.py when its tables have not changed (via checksum), but make will always try to regenerate parsetab.py if it is older than constraint/parser.py. This generates unnecessary work for make on every run when parser.py is updated, but the parsing has not changed.

Touch constraint/parser.py after running constraint/parser.py to indicate to make that this file is up to date whether it was rewritten or not.


garlick (approved):
LGTM

PR #4928, opened 2023-02-07
etc: update bash completions

grondo:
Problem: Some new commands and subcommands were added without updates to the completions file.

Add support for completion of `flux job timeleft`, `flux pgrep` and `flux pkill`.


grondo:
I'm just going to set MWP now on this one so it will get auto-merged if approved.

chu11 (approved):
LGTM!

PR #4929, opened 2023-02-07
flux-queue: adjust default output of min/max/range

### LINKED ISSUE #4927: should flux queue list be renamed to flux queue policy?
Problem: it's not obvious that the NNODES, NCORES, and NGPUS colums in `flux-queue` list are limits, as opposed to resources assigned to the queues.
```
$ flux queue list
QUEUE    DEFAULTTIME  TIMELIMIT     NNODES     NCORES      NGPUS
debug*           30s        30m          -          -          -
batch             1h         8h          -          -          -
all              30s          -          -          -          -
```
Would this output make more sense if the subcommand were renamed from "list" to "policy"?

chu11:
Problem: The current default outputs of node, core, and gpu values
and ranges in flux queue list is not considered very user friendly.

Solution: If the value of nnodes, ncores, or ngpus is not configured,
instead of always returning "inf", return "inf" only if it is the max.
Return 0 if it is the min.  As a result, the default output from the
range values of nodes, cores, and gpus becomes 0-inf.

Fixes #4927

### DIFF
--- a/doc/man1/flux-jobs.rst
+++ b/doc/man1/flux-jobs.rst
@@ -173,13 +173,11 @@ has a queue assigned in the default format shown above).
 The special presentation type *h* can be used to convert an empty
 string, "0s", "0.0", or "0:00:00" to a hyphen. For example, normally
 "{nodelist}" would output an empty string if the job has not yet run.
-By specifying, "{nodelist:h}", a hyphen would be presented instead.  The
-special presentation type *i* is similar to *h*, but can convert
-"inf" (infinity) to a hyphen.
+By specifying, "{nodelist:h}", a hyphen would be presented instead.
 
 The special suffix *+* can be used to indicate if a string was truncated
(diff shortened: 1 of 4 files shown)


grondo (approved):
Wow, that ended up cleaning up a lot of special case code. Nice work!
LGTM, though you may want to also get @garlick's feedback.

garlick (approved):
I like it!
```
QUEUE    DEFAULTTIME  TIMELIMIT     NNODES     NCORES      NGPUS
debug*           30s        30m      0-inf      0-inf      0-inf
batch             1h         8h      0-inf      0-inf      0-inf
all              30s        inf      0-inf      0-inf      0-inf
```
This makes more sense to me.

grondo:
@chu11: ok to set MWP?

chu11:
yup, setting MWP

PR #4939, opened 2023-02-14
doc: add shell help on flux-jobs(1) formatting

### LINKED ISSUE #4938: tcsh steals '!' in custom output strings
tcsh appears to be interpreting '!H', '!F', etc. in flux command custom output strings as attempts to get commands from the command history rather than as strings to pass to flux. Users can get around this by escaping the '!' with, e.g. '\!H', but this should probably at least be documented in the appropriate man pages.

Example:
```
rzwhippet18{testnewm}21: flux jobs -A -o "{contextual_time!H}"
H: Event not found.
rzwhippet18{testnewm}24: flux jobs -A -o '{contextual_time!H}'
H: Event not found.
rzwhippet18{testnewm}25: flux jobs -A -o '{contextual_time\!H}'
TIME
4:00:00
12:00:00
12:00:00
3:50:00
3:50:00
3:50:00
1:39:46
rzwhippet18{testnewm}26:
```

chu11:
Problem: The exclamation point used in flux-jobs(1) conversion flags is interpreted by some shells, such as csh/tcsh.  It leads to the following error:

$ flux jobs -A -o "{contextual_time!H}"
H: Event not found.

Solution: Add a note in flux-jobs(1) reminding some users that some shells will interpret the exclamation point and it needs to be escaped.

Fixes #4938

### DIFF
--- a/doc/man1/flux-jobs.rst
+++ b/doc/man1/flux-jobs.rst
@@ -170,6 +170,9 @@ would eliminate the EXCEPTION-TYPE column if no jobs in the list received
 an exception. (Thus the job queue is only displayed if at least one job
 has a queue assigned in the default format shown above).
 
+As a reminder to the reader, some shells will interpret braces
+(``{`` and ``}``) in the format string.  They may need to be quoted.
+
 The special presentation type *h* can be used to convert an empty
 string, "0s", "0.0", or "0:00:00" to a hyphen. For example, normally
 "{nodelist}" would output an empty string if the job has not yet run.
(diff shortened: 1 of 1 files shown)


chu11:
@ryanday36 looks ok?

ryanday36 (commented):
I would be fully explicit and say "The exclamation point may need to be escaped (``\!``)."

ryanday36 (approved):
looks good. Thanks @chu11!

chu11:
oh crap ... 

```
       As  a reminder to the reader, some shells may interpret special characters in Python's string format syntax.  The format may need
       to be quoted or escaped to work under certain shells.
```

this was a little bit later in the manpage.  Let me combine them.

chu11:
re-tweaked the changes, instead of having that generic reminder lower in the section, I now have two more explicit sentences talking about quoting braces and escaping exclamation points.

garlick (approved):
LGTM

PR #4949, opened 2023-02-17
broker: allow a file argument to `-c, --config-path` in TOML or JSON

grondo:
Currently, the broker can only read configuration as a set of `*.toml` files from a configuration directory. This is because the underlying `flux_conf_parse(3)` call always appends `/*.toml` to its argument and reads all matching files.

This PR extends `flux_conf_parse(3)` to detect if its path argument is a directory, and if not, attempts to read the file directly. Additionally, if the file has the extension `.json`, parse the config as JSON instead of TOML.

If the path argument is a directory, `flux_conf_parse(3)` still only reads TOML from that directory, to avoid inadvertently parsing some JSON files that wer ignored before.

Thi ability to parse a  file allows more convenient use of simple config files for the broker (you don't have to put your config file in an otherwise empty directory). Parsing JSON allows the output of `flux config get` from one instance to be read by another (since we can easily convert TOML to JSON, but not the other way around). It will also make it easier to stash configuration in jobspec in the future.

I tried to find and update most references to the prior requirement that `--config-path=PATH` point to a dir.

### DIFF
--- a/doc/man5/flux-config.rst
+++ b/doc/man5/flux-config.rst
@@ -6,21 +6,27 @@ DESCRIPTION
 ===========
 
 Flux normally operates without configuration files.  If configuration is
-needed, the :man1:`flux-broker` **--config-path=DIR** option may be used
-to instruct Flux to parse all files matching the glob **DIR/*.toml**.
-Alternatively, the FLUX_CONF_DIR environment variable may be set the config
-directory path.  If both are set, the command line argument takes precedence.
-The option value must be a directory, not a file.
+needed, the :man1:`flux-broker` **--config-path=PATH** option may be used
+to instruct Flux to parse a config file or, if **PATH** is a directory, all
+files matching the glob **PATH/*.toml**.  Alternatively, the FLUX_CONF_DIR
+environment variable may be used to set the configuration file or directory
+path. If both are set, the command line argument takes precedence.
 
 The Flux systemd unit file starts the system instance broker with
 ``--config-path=${sysconfdir}/flux/system/conf.d``.  Further discussion of the
 system instance configuration as a whole may be found in the Flux
 Administrator's Guide.
 
-Flux configuration files follow the TOML file format, with configuration
-subdivided by function into separate TOML tables.  The tables may all appear
-in a single ``.toml`` file or be fragmented in multiple files.  For example,
-the Flux Administrator's Guide suggests one file per top level TOML table.
+Flux configuration files typically follow the TOML file format,
+with configuration subdivided by function into separate TOML tables.
+The tables may all appear in a single ``.toml`` file or be fragmented in
+multiple files.  For example, the Flux Administrator's Guide suggests one
+file per top level TOML table.
+
+If the **PATH** value is a regular file, the configuration may optionally be
+provided as JSON. This allows the output of ``flux config get`` to be read
+by another Flux instance. A JSON config file must have the file extension
+``.json``, otherwise it is assumed to be TOML.
 
 Each Flux broker parses configuration files independently, however it is
 assumed that config files are identical across the brokers of a given instance.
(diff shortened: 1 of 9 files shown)


chu11:
high level comment

> If the path argument is a directory, flux_conf_parse(3) still only reads TOML from that directory, to avoid inadvertently parsing some JSON files that wer ignored before.

Understand the desire for backwards compatibility, but there's the part of me that says we should just "eat it" and allow both .json and .toml.  Because inevitably someone will put .json files in the directory and wonder why that doesn't work.

src/common/libflux/test/conf.c
```diff
@@ -120,9 +133,10 @@ void test_basic (void)
 
     /* Add files
      */
-    create_test_file (dir, "01", path1, sizeof (path1), t1);
-    create_test_file (dir, "02", path2, sizeof (path2), tab2);
-    create_test_file (dir, "03", path3, sizeof (path3), tab3);
+    create_test_file (dir, "01", NULL, path1, sizeof (path1), t1);
+    create_test_file (dir, "02", NULL, path2, sizeof (path2), tab2);
+    create_test_file (dir, "03", NULL, path3, sizeof (path3), tab3);
```
chu11, line 138: instead of NULL extension meaning "toml", perhaps just pass "toml", so its clear which ones are toml and which ones are json.  adjust `create_test_file()` as a result (i think).

grondo:
> Understand the desire for backwards compatibility, but there's the part of me that says we should just "eat it" and allow both .json and .toml. Because inevitably someone will put .json files in the directory and wonder why that doesn't work.


Good point. I wonder if @garlick has an opinion and we if we all agree we make that change. I'm definitely on the fence.

src/common/libflux/test/conf.c
grondo: Yeah, that's a good suggestion.

garlick:
>> Understand the desire for backwards compatibility, but there's the part of me that says we should just "eat it" and allow both .json and .toml. Because inevitably someone will put .json files in the directory and wonder why that doesn't work.

> Good point. I wonder if @garlick has an opinion and we if we all agree we make that change. I'm definitely on the fence.

That seems wrong to me.  It feels as though we would be encouraging sys admins to prepare config in JSON form instead of TOML or to mix them, which we're not, and probably shouldn't because JSON is not especially human friendly.

grondo:
>  It feels as though we would be encouraging sys admins to prepare config in JSON form instead of TOML or to mix them,

Yeah, I especially don't like the idea of ending up with them mixed.

>  It feels as though we would be encouraging sys admins to prepare config in JSON form instead of TOML

Just to clarify, are you ok with `flux_conf_parse(3)` allowing JSON when PATH is a single file with `.json` extension? I'm not sure that encourages users and admins to prepare config in JSON instead of TOML or not, but it does seem wrong that  the output of `flux config get` can't be parsed as a config file...

garlick:
I'm OK with that change and I think the bit you added to `flux-config(5)` makes it fairly clear that the JSON support is for ingesting the entire config object when dumped by software, not for humans.

src/common/libflux/test/conf.c
grondo: Fixed that and force-pushed.

grondo:
Thanks @garlick and @chu11. Force-pushed a fix for the file extension handling in the unit test and I'll set MWP.

doc/man5/flux-config.rst
```diff
@@ -6,21 +6,27 @@ DESCRIPTION
 ===========
```
garlick, line 7: "documentaition" is misspelled in the commit title of 07457e450b9f8f4be16445fc15f62f2191095c9a

garlick (approved):
LGTM!  Just noticed one commit message misspelling

grondo:
Thanks! Fixed that typo caught by @garlick and will set MWP.

PR #4934, opened 2023-02-10
Fix minor documentation errors and typos

chu11:
Problem: "sub-instance" should not be hyphenated

Fix spelling.

There's more misuse in `flux-docs`, but saw there was one dumb one here :P  The remaining ones were in NEWS.md, but didn't touch those.

PR #4950, opened 2023-02-18
improve error message in Python frontend utilities when broker is not running

grondo:
This PR is a stab at improving the error message from most utilities when `flux_open()` fails. Specifically, when the local connector socket is not found, an error more descriptive than "No such file or directory" is set in the `flux_error_t` of `flux_open_ex(3)` as suggested by @garlick, e.g.:
```
flux-jobs: ERROR: Unable to connect to Flux: broker socket /run/flux/local was not found
```

In other cases, `strerror()` is still used, e.g.
```console
$ FLUX_URI=local:///tmp src/cmd/flux jobs
flux-jobs: ERROR: Unable to connect to Flux: Connection refused
```
which seems about as clear as you can get.

This only works for Python utilities since many C based commands have not switched to `flux_open_ex(3)`.

In addition, error messages from most Python utilities are simplified by attempting to drop the unnecessary `[Error N]` prefix in the `CLIMain()` utility decorator.

### DIFF
--- a/src/bindings/python/flux/util.py
+++ b/src/bindings/python/flux/util.py
@@ -218,7 +218,13 @@ class CLIMain(object):
             exit_code = ex
         except Exception as ex:  # pylint: disable=broad-except
             exit_code = 1
-            self.logger.error(str(ex))
+            # Prefer '{strerror}[: {filename}' error message over default
+            # OSError string representation which includes useless
+            # `[Error N]` prefix in output.
+            errmsg = getattr(ex, "strerror", None) or str(ex)
+            if getattr(ex, "filename", None):
+                errmsg += f": '{ex.filename}'"
+            self.logger.error(errmsg)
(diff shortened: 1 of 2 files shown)


garlick (approved):
Looks good, thanks!

grondo:
Thanks!

src/bindings/python/flux/util.py
```diff
@@ -218,7 +218,13 @@ def __call__(self, main_func):
             exit_code = ex
         except Exception as ex:  # pylint: disable=broad-except
             exit_code = 1
-            self.logger.error(str(ex))
+            # Prefer '{strerror}[: {filename}' error message over default
```
chu11, line 221: is that `[` in here a typo?  Could just tack on top of #4934 if it is.

grondo: Yes. Good catch

PR #4951, opened 2023-02-20
Support {{tmpdir}} in shell mustache templates and simplify batch jobspec construction

grondo:
The shell `batch` plugin needs to rewrite the job arguments because the batch script is written to a random filename in TMPDIR. This unforutunately requires different handling of "nested" jobspec (i.e. jobspec that creates a subinstance) between `flux mini alloc` and `flux mini batch`.

Now that there is a `FLUX_JOB_TMPDIR`, write the job script there as `$FLUX_JOB_TMPDIR/script` so the batch script is always in a well known location.

Extend the shell mustache rendere to allow plugins to fulfill templates by subscribing to the callback topic `mustache.render.<name>`. Extend the `tmpdir` plugin to substitute `{{tmpdir}}` with `FLUX_JOB_TMPDIR`.

Finally, refactor the Jobspec `from_batch_command()` method to use `from_nest_command()`, a batch jobspec is now just a special case of a `flux mini alloc` jobspec. The command line becomes `flux broker [broker_opts] {{tmpdir}}/script [options]...`.

The batch plugin now checks for initial arguments of `flux broker` or `flux start` and skips rewriting args in that case. Otherwise it behaves as before to handle running older jobspecs which may be queued after a flux-core upgrade.

This PR could be split into 2 or 3 separate PRs if that would be clearer, just wasn't sure if the extra work that would entail would be worth it.

Also, this is built on top of #4942 to avoid conflicts in `flux-mini.py`.

### DIFF
--- a/src/bindings/python/flux/job/Jobspec.py
+++ b/src/bindings/python/flux/job/Jobspec.py
@@ -902,20 +902,18 @@ class JobspecV1(Jobspec):
         if not script.startswith("#!"):
             raise ValueError(f"{jobname} does not appear to start with '#!'")
         args = () if args is None else args
-        jobspec = cls.from_command(
-            command=[jobname, *args],  # argv[0] will be replaced with the script
-            num_tasks=num_slots,
-            cores_per_task=cores_per_slot,
-            gpus_per_task=gpus_per_slot,
+        jobspec = cls.from_nest_command(
+            command=["{{tmpdir}}/script", *args],
+            num_slots=num_slots,
+            cores_per_slot=cores_per_slot,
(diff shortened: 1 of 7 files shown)


grondo:
Rebased on current master after merge of #4942

grondo:
Added some extra tests to bump the coverage of some new (and moved) code.

We should probably get this merged sooner rather than later since the refactor of `flux-mini.py` into separate commands is imminent.

garlick (approved):
LGTM!  Sorry for the delay getting this reviewed.

grondo:
Thanks!

PR #4955, opened 2023-02-23
libsubprocess: drop FLUX_SUBPROCESS_EXEC_FAILED state

garlick:
Problem: the FLUX_SUBPROCESS_EXEC_FAILED state is not needed

This state is a terminal state that indicates that a remote `flux_exec()` call failed before a PID could be returned.  Other remote failures, before and after a PID is returned, are covered by FLUX_SUBPROCESS_FAILED, so the extra state seems to serve no purpose.

Drop the state and update users.

Also, fix logic surrounding a last-ditch SIGTERM sent to subprocesses on failure and improve unit testing on remote subprocess channels.

### DIFF
--- a/src/broker/runat.c
+++ b/src/broker/runat.c
@@ -179,7 +179,6 @@ static void state_change_cb (flux_subprocess_t *p,
 
     switch (state) {
         case FLUX_SUBPROCESS_INIT:
-        case FLUX_SUBPROCESS_EXEC_FAILED:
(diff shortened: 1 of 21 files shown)


garlick:
Repushed with a little test program instead of `socat(1)` as suggested by @grondo offline.

grondo (approved):
Great cleanup! LGTM!

PR #4959, opened 2023-02-23
flux-resource: suppress NGPUS on systems without GPUs

grondo:
This PR extends the `?:` prefix in the `OutputFormat` class to suppress columns in which all lines have a representation of zero in addition to suppressing all empty lines. This then allows use of `?:` in the `ngpus` column of `flux resource list`. Displaying an NGPUS column on systems without GPUs was identified as an important issue by at least one user.


garlick (approved):
Seems to have the desired effect on my gpuless test system!  LGTM

PR #4960, opened 2023-02-24
libsubprocess: rework internal logging

garlick:
This eliminates the calls to `flux_log()` inside libsubprocess, which is not appropriate in all cases.  It is replaced with an llog function callback, which has been used in the shell.

Since some log messages occur right when the subprocess is created, it would not work to have a "setter" on the subprocess object to set up the logging - that would be too late.  So rather than add logger parameters to `flux_rexec()` and friends, I added a single function `flux_set_default_subprocess_log()` that sets a log function and argument in the `flux_t` handle aux container.  Then when a subprocess or the subprocess server is created, it uses that if it was set.

A llog-compatible callback that logs to the broker is provided, and that is used in the broker-resident subprocess users for now.

The existing llog-compatible shell logger is used in the flux shell.

### DIFF
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -409,6 +409,11 @@ int main (int argc, char *argv[])
     if (ctx.rank == 0 && execute_parental_notifications (&ctx) < 0)
         goto cleanup;
 
+    /* Set up internal logging for libsubprocesses.
+     */
+    if (flux_set_default_subprocess_log (ctx.h, flux_llog, ctx.h) < 0)
+        goto cleanup;
+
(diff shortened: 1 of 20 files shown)


garlick:
For some reason the change to subprocess logging in the flux shell was one of the shell tests to fail in CI.  Since that would have been a change in behavior for the shell (subprocess was not logging at all before) I dropped that commit for now.

grondo:
I thought I had approved this one but I guess I never submitted it. Let me take another look, but overall great cleanup and well needed!

grondo (approved):
Nice cleanup, LGTM!

PR #4961, opened 2023-02-24
make all flux-mini subcommands available as top level flux commands

vsoch:
Problem: The flux user interface hides the main commands that a user needs to submit jobs.
Solution: Step 1, is to move the flux mini sub commands up to the top level.

This is marked as a WIP until I can verify all tests work! I have been trying to run them to test locally buy it turns my computer to kaput. A few notes:

- each mini sub-command is refactored into a new command script.
- updated: bases are all put with flux python bindings in flux.cli.
- the new docs also share the common bulk of content from a new shared/ folder, with customization only based on minor differences I saw between the commands.
- I changed the gen-cmdline.py to not do an `exec(open(read))` because I think we can accomplish the same by just adding the module to the path and importing.
- I can follow up in another PR with maybe some docs about how those python commands magically appear there. It wasn't as straightforward as I thought.
- I added the tests by refactor each of the subcommand tests into a new test that uses it. The original mini tests remain the same.
- I tried to make as few changes as possible to the Python code, the goal being this PR should just be moving things around into this new structure, and not modifying beyond that.
- Any other aesthetic changes (e.g., extra spaces) are likely from black, which has been adding newlines to most of my projects.
- I've still never been able to run a test locally so I'm going to have to debug here.
- I have no idea how to split this artificially into separate commits and then possibly need to work on just one of them. I think if I do that I'll break the branch.

I've marked this as a WIP because I need to see if the tests pass and make changes as needed! I'll flag it ready when that is done.


vsoch:
okay I need to ask for help here: https://github.com/flux-framework/flux-core/actions/runs/4260277944/jobs/7413330755 because when I start a flux instance and run a lot of these tests on the command line manually, there are no errors. But when I run the file as a script, almost everything fails. Let me know what I can try or what I'm doing wrong.

garlick:
I'm not sure about the approach - I thought we would probably start by creating aliases for the new command names and not necessarily break the old ones, then shift tests to the new names, then figure out some way to deprecate/hide the old `mini` name, and decide whether we want to do that in a later release or right away.  Ideally without forcing a re-review of all that code.

vsoch:
I do t think my approach is relevant to my question or the tests… ? I disagree and think this approach is best.

garlick:
> I'm not sure about the approach - I thought we would probably start by creating aliases for the new command names and not necessarily break the old ones, then shift tests to the new names, then figure out some way to deprecate/hide the old mini name, and decide whether we want to do that in a later release or right away.

Sorry, I misunderstood - I see that `flux-mini` still works and so we can still do that.

garlick:
Could `base.py` go in `src/bindings/python/flux/submit/base.py` or similar?

garlick:
I actually have no idea what the right location for `base.py` is!   Just thought maybe that would work to get past the failing tests.

Could `flux-mini` just call the new commands instead of duplicating all that code?  I If `mini` remains for a while it will become a problem to maintain in two places, and presumably it won't be covered by tests anymore?  (I didn't check that).

Minor at this stage but in case you are reworking stuff good to keep in mind:
- we don't do copyright ranges in this project (just leave the original dates as is)
- drop the whitespace changes
- docs changes should be broken out to (at least) one commit
- test changes should be broken out to (especially since the code changes shouldn't break anything)

vsoch:
I just rebased so the warning above goes away. Can someone (maybe that wrote and understands) these tests help? I don't even know where to start.

grondo:
This is great! I'm really glad you decided to refactor the commands instead of doing wrappers. This should really help development of the commands going forward. Calling the new `flux.cli` version from `flux-mini.py` will also help us ensure backwards compatibility (while we need it). Awesome!

Thank you so much for getting this started. And thanks for making suggested changes thus far, super helpful!

I'll try to check out what's going wrong in the testsuite. Meanwhile, one high level comment for now:

 - maybe a `black` misconfiguration on your development system has a bunch of unrelated whitespace and formatting changes throughout many Python files in the project. This makes it a little bit harder to review the PR and typically formatting changes unrelated to new features or fixes would go in their own commit anyway. If there is any way, can we drop those changes here?

Thank you!

vsoch:
@grondo thank you for your kindness I'm fairly discouraged with the tests here because I don't understand them.

> maybe a black misconfiguration on your development system has a bunch of unrelated whitespace and formatting changes throughout many Python files in the project. This makes it a little bit harder to review the PR and typically formatting changes unrelated to new features or fixes would go in their own commit anyway. If there is any way, can we drop those changes here?

I have the right version of black in my flux environment:

```
$ which black
/home/vanessa/Desktop/Code/flux/flux-core/env/bin/black

$ black --version
black, 22.8.0 (compiled: yes)
Python (CPython) 3.9.12
(env) (base) vanessa@vanessa-ThinkPad-T490s:~/Desktop/Code/flux/core$ 
```
so it must have been pulled in at some point. Let me manually add those lines back.

grondo:
> thank you for your kindness I'm fairly discouraged with the tests here because I don't understand them and the testing setup has many layers and I have never gotten them working locally.

No problem! The testing has to simulate many environments we use in production as well as the `make distcheck` target which is sometimes a pain to deal with! I have no problem helping out with that mystery!

vsoch:
@grondo the only way I've been able to test locally is manually copy pasting those commands after manually changing the configure.ac to have a fake version and then starting a flux instance. I thought those first commands were supposed to fail because each of flux submit and run do require something (not just an empty directive).

grondo:
Next issue: Some of the failures in `t2713-python-cli-bulksubmit.t` are because the tests check for dry-run output of `flux: submit foo` when the command still outputs `flux-mini: submit foo`. This might be a little tricky to handle since we want the `flux mini` tests to succeed as well, but I'm fine just updating the `flux mini` tests -- the prefix of the dry-run output isn't that important.

Maybe a next step here is to search for `flux-mini` inside the new `flux.cli` sources and fix those.

grondo:
BTW, I'd be happy to fix these little test issues and push fixup commits to your branch along the way. However, it seems like you are still actively developing here and I don't want to cause conflicts!

vsoch:
oh no I'm totally not! I switched to working on the operator and learning about terraform because I needed a new thing for the rest of today. Please commit and push away. The only reason I left the last commit (and didn't fully rebase) is so we have the two versions (that are fairly different, organization wise) in the history, if needed. So please feel free to edit the git history however you see fit - I suspect you are much better at that than me!

I really appreciate the help! I have a hard time contributing to Flux because of the testing setup.

grondo:
No problem! I am super impressed this is all working and that is so great!

I'm happy to take things from here and I'll get your feedback on any questions I have along the way.

vsoch:
:raised_hands: :bow: :sob: :joy: 

thank-you

grondo:
> I really appreciate the help! I have a hard time contributing to Flux because of the testing setup.

When you have a chance (no rush), can you summarize your development environment and how you are running the testsuite and the problems you hit? I'd like to attempt to resolve any issues you are having to make the process easier for you and any other contributors that come in running the same environment. In general, our small coterie of developers thus far has not had too much issue with debugging the testsuite, but we want to make it easy for everyone!

vsoch:
yeah! So here is what I do:

- I open in VSCode, this gives me the option to "Open in a container"
- In the container, I need to edit the configure.ac to have a hard coded version (the fallback to the environment FLUX_VERSION doesn't seem to work)
- Then I autogen, configure, make, make install (assuming everything is ready to go to test)
- At that point I've tried running the container test command (outside of my VSCode on a separate terminal) but then that pollutes the space with a build that, for example, has flux security (and I have to rebuild in my development environment).
- So instead what I do is start a flux instance, `flux-start --test-size=4`
- And then I go into the tests folder and try running individual files. If it's possible I add IPython.embed() to various places to debu interactively. If not, I do a lot of printing.
- In the case like this one where mostly everything failed, I started copy pasting commands one by one into my terminal. But then I see them work in the terminal, but they don't work when I run the test.

When I run `make check` or similar it just takes eons, and sometimes just freezes my computer or never finishes, so I can't really do that.

So TLDR: I'm using the .devcontainers environment and just trying to run tests, and when that doesn't work I have to manually try things. And then finally, even when I think something might be working, I will push and it will fail in the CI. (e.g., the distutils thing).

grondo:
Thanks! I haven't used vscode before, but I'll put it on my list to get that setup so I can reproduce:
> In the container, I need to edit the configure.ac to have a hard coded version (the fallback to the environment FLUX_VERSION doesn't seem to work)

This is because you are developing with a shallow clone? If you `git fetch --tags` this issue should go away, but I'll look into why FLUX_VERSION isn't working.

> Then I autogen, configure, make, make install (assuming everything is ready to go to test)

The testsuite is made to test the build directory, so there is no need to `make install`. There is actually a target `make check-prep` that will prep the builddir for testing so you do not have to run a full `make check`. You can then go into the `t` directory and run tests by hand. I usually use the `-d -v` flags to get full debug output and keep the "trash directory" around for analysis.

The tests are also shell scripts, so to debug you can edit them by hand to get extra information if needed.

vsoch:
>  The testsuite is made to test the build directory, so there is no need to make install. There is actually a target make check-prep that will prep the builddir for testing so you do not have to run a full make check. You can then go into the t directory and run tests by hand. I usually use the -d -v flags to get full debug output and keep the "trash directory" around for analysis.

> The tests are also shell scripts, so to debug you can edit them by hand to get extra information if needed.

Oh nice! Is this advice written up somewhere in developer docs (even in the readme) because future me is going to forget again.

trws:
You can also subset the tests with `make check TESTS=<test-names>` and run most of them directly as @grondo mentioned.  Some docs on the system used to tie it all together [here](https://www.gnu.org/software/automake/manual/html_node/Parallel-Test-Harness.html).

grondo:
Or, best of all, run the entire testsuite as jobs under Flux:
```sh
flux mini bulksubmit -o pty --watch --quiet \
	sh -c './{} --root=$FLUX_JOB_TMPDIR' \
	::: t[0-9]*.t python/t*.py lua/t*.t
```
This is what the test "inception" builder does.

garlick:
I don't think anyone mentioned the `make -j N` "go fast" option.  If you push `make check` too hard you might get a spurious failure here and there but a little parallelism really helps.

grondo:
@garlick refactored the doc changes here and I've found and fixed all the very minor issues that were causing the test suite to fail (along with some refactoring of commits), and I've just pushed the current result. At least locally and in the CI branch for my fork, this is now all working.

I'll now remove WIP from the PR title

vsoch:
Holy cow this is amazing! That is SO many commits!

vsoch:
I'm looking at this and I'm fairly confident I'm not capable of that. E.g., I don't actually move through a PR making changes in a logic fashion, I'm working on everything at once.

My suggestion for what I might do moving forward is doing tiny PRs. E.g., if I want to refactor something large, I just do one trivial tiny thing at once. It will mean a lot more PRs to review, but I think that's the only way I can come close to this commit style requirement.

grondo:
I just want to say thanks for doing the work here @vsoch. I don't think we'd have taken this step and gotten over the activation energy to get it done so quickly without your initial push! And the end result is great! So :bow: :bow: :bow:

grondo:
> I think that's the only way I can come close to this commit style requirement.

I think some of the commits here are broken up a little more than strictly necessary. However, a commit that does one thing (like the gen-cmdhelp.py changes that were unrelated to the other stuff) is the main requirement. Then down the road we remember *why* that file was last changed and the problem that was being solved.

vsoch:
If / when Flux explodes into a larger open source project (with a potentially larger contributor base) we need to talk about a balance for requirements for a new contributor. My 0.02 is that this level of detail is too much, so if it's wanted it can be something a maintainer does after the fact, or it can be lessened a bit. If someone comes here to contribute to the first time and they find the process arduous they will not be likely to come back.

vsoch:
And I really appreciate  you doing this for me - I would not have been capable of it!

grondo:
I've went ahead and update the whole testsuite (besides the `flux-mini` specific tests) to drop the use of `flux mini X` for `flux X`. Testing locally showed that everything worked, so hopefully CI passes as well!

garlick (approved):
Awesome.  Thanks again @vsoch for bringing this up at just the right time following through so quickly.  I'm feeling really good about this change!

grondo:
Great! Let's set MWP and get this merged assuming the testsuite changes do not break anything.

garlick:
Whee! :roller_coaster:

vsoch:
After this can we look at the discussion again https://github.com/flux-framework/flux-core/discussions/4958 and discuss how to hide / move down one level of the commands on the top level that are plumbing? My concern with this merge is that the top level is a bit busy now, so I think this is step 2 (and shouldn't be too hard, knock on wood, granted that I don't mess up the git again).

grondo:
I've got something in progress for `flux help` that should satisfy this I think

garlick:
Apropos of your comments on PR building:
```
$ fortune
The last thing one knows in constructing a work is what to put first.
		-- Blaise Pascal
```

PR #4964, opened 2023-02-25
.devcontainer permissions fix

vsoch:
Problem: writing files in the DevContainer is done as root. 
Solution: have the default user be uid 1000, gid 1000, and they are written as the same user on the host.

To demonstrate this working, here I am making lasagna in the container:

image

And the same file viewed from the host - it's still owned by me!

image

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -2,6 +2,14 @@ FROM fluxrm/testenv:focal
 
 LABEL maintainer="Vanessasaurus <@vsoch>"
 
+# Match the default user id for a single system so we aren't root
+ARG USERNAME=vscode
+ARG USER_UID=1000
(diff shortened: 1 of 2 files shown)


vsoch:
Oh you know what we should add vscode to sudoers too. Give me one minute (finishing another PR).

vsoch:
okay just added! the vscode user should now be able to do `sudo make install`. This isn't essential because we can use flux before it's moved into a system bin, but I would expect the developer to want to potentially make other changes to the environment (e.g as I do sometimes install vim or similar).

garlick (approved):
OK by me (caveat: I'm not adding much here, just assuming you know what you're doing)

vsoch:
Absolutely! 
image

garlick:
(already approved so set merge when passing when ready)

PR #4968, opened 2023-02-27
allow flux-core to be configured in ascii-only mode

grondo:
This PR adds a new option to flux-core configure: `--enable-broken-locale-mode`, which when used causes flux-core to avoid using the non-ascii `ƒ` character as a FLUID prefix (and in `flux top` etc). It is offered as more fool-proof way to disable the "fancy" character for install where locales or terminals are known to be misconfigured such that `ƒ` does not display correctly.

In these environments, RPMs can be built with this option and the question of when/how to set `FLUX_F58_FORCE_ASCII` can be avoided.

Set as a WIP in case others have better ideas, and because I'm considering if we need a CI builder with this option enabled.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -416,6 +416,15 @@ AS_IF([test "x$enable_content_s3" = "xyes"], [
 
 AM_CONDITIONAL([ENABLE_CONTENT_S3], [test "x$enable_content_s3" = "xyes"])
 
+AC_ARG_ENABLE(
+    [broken-locale-mode],
+    AS_HELP_STRING([--enable-broken-locale-mode],
+                   [Assume broken locale config and use ascii only]))
+
+if test "$enable_broken_locale_mode" = "yes"; then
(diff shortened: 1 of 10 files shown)


garlick:
Yeah I think this is probably a good idea.

grondo:
I went ahead and added an el8 build to CI with `--enable-broken-locale-mode`. This should help us catch added tests that expect `ƒ` in F58 encoded jobids and thus fail with this option enabled (which we wouldn't otherwise catch until building RPMs).

However, if this extends the total checks runtime, it may be better to just add this flag to one of the existing builds instead of adding a new one.

grondo:
Didn't seem to extend the overall runtime of CI by adding another build, so we'll keep that for now. I'll remove WIP.

garlick (approved):
LGTM!

grondo:
Thanks, assuming tests pass I'll set MWP

PR #4969, opened 2023-02-27
README: trim it down now that we have readthedocs

garlick:
Problem: now that we have extensive docs on the readthedocs site, the info in the README is prone to getting stale.

Prune it a bit and reorganize.


grondo (approved):
Nice! This LGTM!

Issue #4956, opened 2023-02-23
stdin redirection options


ryanday36:
A user is asking for more control over how they interact with their job via stdin:

> pF3D requires that stdin only be connected to MPI rank zero and it wants I/O to be unbuffered. This is done by passing "-i0 -u" to srun. I need to know the equivalent flags for flux.

ryanday36:
More from Slack:

grondo
  [11 minutes ago](https://llnl-performance.slack.com/archives/CBNV6RG8Y/p1677177960133539?thread_ts=1677177460.522789&cid=CBNV6RG8Y)
Hm, I'm not sure if we can do that at the moment. :grimacing: I wonder if running with a pty would work in this case. I'm trying to think of a test case we can use.
grondo
  [10 minutes ago](https://llnl-performance.slack.com/archives/CBNV6RG8Y/p1677178046772799?thread_ts=1677177460.522789&cid=CBNV6RG8Y)
We better open an issue on this one. I wonder if pF3D is one of those programs that has an interpreter like interface on rank 0?
Ryan
  [8 minutes ago](https://llnl-performance.slack.com/archives/CBNV6RG8Y/p1677178166368899?thread_ts=1677177460.522789&cid=CBNV6RG8Y)
:+1: I can open an issue and try to get more details from Steve.
grondo
  [6 minutes ago](https://llnl-performance.slack.com/archives/CBNV6RG8Y/p1677178278748059?thread_ts=1677177460.522789&cid=CBNV6RG8Y)
running with -o pty.interactive  might work. This allocates a pty on rank 0 and attaches to it with flux job attach and thus the standard i/o is all unbuffered...
grondo
  [5 minutes ago](https://llnl-performance.slack.com/archives/CBNV6RG8Y/p1677178346519089?thread_ts=1677177460.522789&cid=CBNV6RG8Y)
but unbuffered io should be doable without a pty as well and currently isn't. And this won't work if flux job attach is run outside of a pty like in a batch job, so maybe nevermind.

ryanday36:
More details on pF3D from Steve Langer:

pF3D is actually a custom version of Yorick. Basically we just tell Yorick to dyload a set of physics libraries. Yorick is an interpreter that uses a C-like language. An important feature is that arrays are first class data types, so you can perform element-wise operations on arrays without an interpreted loop. Yorick predates the serious use of Python for scientific research. When numpy was written, it had a goal of including all the array operations supported by Yorick. 

[...]

PR #4973, opened 2023-03-01
cmd: add -u, --unbuffered option to submission commands

grondo:
This PR partially addresses #4956. A new `-u, --unbuffered` option is added to `flux run`, `flux submit`, etc which attempts to disable input and output buffering "as much as practical". This option does a few things:

 - Sets `output.*.buffer.type = "none"` in jobspec to disable line buffering of application `stdout` in the job shell (and ensure the same for `stderr`)
 - Sets `output.batch-timeout = 0.05` to minimize batching of output eventlog events so output arrives in the eventlog in a timely manner
 - For `flux run`, a new `--unbuffered` option is passed to `flux job attach` which disables line buffering of `stdin`.

The next piece for #4956 is to allow stdin to be directed to only a subset of tasks, possibly redirecting input from `/dev/null` for the remainder. That work is slightly more involved though, and it is possible that this option alone may satisfy many use cases. Also, this seemed like a good stopping point for the release notes entry.

### DIFF
--- a/doc/man1/common/submit-standard-io.rst
+++ b/doc/man1/common/submit-standard-io.rst
@@ -25,6 +25,24 @@ emitting the job's I/O to its stdout and stderr.
    Redirect stderr to the specified filename *TEMPLATE*, bypassing the KVS.
    *TEMPLATE* is expanded as described above.
 
+**-u, --unbuffered**
+   Disable buffering of standard input and output as much as practical.
+   Normally, stdout from job tasks is line buffered, as is stdin when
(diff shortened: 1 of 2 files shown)


grondo:
FYI - the Python coverage module can't capture coverage before an `os.execvp` call, so that is why there is low coverage in `run.py` reported here.

garlick (approved):
LGTM!

grondo:
Thanks! I'll set MWP.

PR #4976, opened 2023-03-03
testsuite: unset FLUX_F58_FORCE_ASCII in some tests

grondo:
Problem: The FLUX_F58_FORCE_ASCII environment variable is sanitized in all sharness tests, but not in unit and python tests.

Remove the FLUX_F58_FORCE_ASCII environment variable as needed in tests where it breaks expected output or results.


garlick (approved):
:+1:

PR #4977, opened 2023-03-04
flux-run: allow stdin to be directed to a subset of tasks

grondo:
This PR adds support for directing standard input of `flux run` to a subset of task ranks.

This is a stopgap for #4956 until we can get better support for input redirection directly in the job shell.

This PR works by adding a new `--stdin-ranks=RANKS` option to `flux job attach`. Then, the `--input` option (in `flux run` only) is extended so that it a supplied idset is passed along to `flux job attach` instead of setting a filename in jobspec.

In addition, there's some minor cleanup in the job shell to detect if a writer attempts to write to closed tasks.

In a future PR, a new scheme will be proposed to allow more generic redirection of standard input, but `--input=RANKS` should still work in this case, so that we won't be changing behavior.

### DIFF
--- a/doc/man1/common/submit-standard-io.rst
+++ b/doc/man1/common/submit-standard-io.rst
@@ -7,8 +7,10 @@ KVS, where they may be accessed with the ``flux job attach`` command.
 In addition, :man1:`flux-run` processes standard I/O in real time,
 emitting the job's I/O to its stdout and stderr.
 
-**--input=FILENAME**
+**--input=FILENAME|RANKS**
    Redirect stdin to the specified filename, bypassing the KVS.
+   As a special case for ``flux run``, the argument may specify
+   an idset of task ranks in to which to direct standard input.
 
 **--output=TEMPLATE**
    Specify the filename *TEMPLATE* for stdout redirection, bypassing
--- a/src/bindings/python/flux/cli/base.py
+++ b/src/bindings/python/flux/cli/base.py
@@ -775,7 +775,15 @@ class MiniCmd:
             jobspec.setattr("system.job.name", args.job_name)
 
         if args.input is not None:
-            jobspec.stdin = args.input
+            #  Note: temporary stopgap until per-task input is supported
+            #  by the job shell:
+            #  Check if --input specified an IDset. If not, then assume
+            #  a file, otherwise, do not modify jobspec, input will be
+            #  handled by `flux job attach`.
+            try:
+                IDset(args.input)
+            except (ValueError, OSError):
+                jobspec.stdin = args.input
 
         if args.output is not None and args.output not in ["none", "kvs"]:
             jobspec.stdout = args.output
@@ -1446,6 +1454,17 @@ class SubmitBulkCmd(SubmitBaseCmd):
             #  substitute any {cc} in args (only if --cc or --bcc):
             xargs = Xcmd(args, cc=i) if i else args
             jobspec = self.jobspec_create(xargs)
+
+            #  For now, an idset argument to args.input is not supported
+            #  in submit:
+            if xargs.input:
+                try:
+                    IDset(xargs.input)
+                    LOGGER.error("per-task input not supported for submit")
+                    sys.exit(1)
+                except (ValueError, OSError):
+                    # --input was not an idset, just continue:
+                    pass
 
             if args.cc or args.bcc:
                 label = f"cc={i}: "
--- a/src/cmd/flux-job.c
+++ b/src/cmd/flux-job.c
@@ -1588,7 +1588,7 @@ struct attach_ctx {
     flux_jobid_t id;
     bool readonly;
     bool unbuffered;
-    const char *stdin_ranks;
+    char *stdin_ranks;
     const char *jobid;
     const char *wait_event;
     flux_future_t *eventlog_f;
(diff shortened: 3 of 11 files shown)


src/shell/input.c
```diff
@@ -375,6 +425,11 @@ struct shell_input *shell_input_create (flux_shell_t *shell)
     in->shell = shell;
     in->stdin_type = FLUX_INPUT_TYPE_SERVICE;
     in->ntasks = shell->info->rankinfo.ntasks;
+    if (!(in->open_tasks = idset_create (0, IDSET_FLAG_AUTOGROW))
+        || idset_range_set (in->open_tasks, 0, shell->info->total_ntasks - 1)) {
+        free (in);
+        return NULL;
+    }
```
garlick, line 432: leaks `in->open_tasks` if `idset_range_set()` fails?

src/bindings/python/flux/cli/base.py
```diff
@@ -775,7 +775,15 @@ def jobspec_create(self, args):
             jobspec.setattr("system.job.name", args.job_name)
```
garlick, line 775: "to a a subset" (delete one a) in commit message for 0b89b136153c2cba93f6813b

doc/man1/common/submit-standard-io.rst
```diff
@@ -7,8 +7,10 @@ KVS, where they may be accessed with the ``flux job attach`` command.
 In addition, :man1:`flux-run` processes standard I/O in real time,
```
garlick, line 7: s/documents/documented/ in commit message for 49115b9b3e065

garlick (commented):
Found a couple things on a quick review pass.
I'll give this a try later this weekend

grondo:
Fixed the issues noted in first review pass above and forced a push

garlick:
This seems to work great!

One thing I did notice is that if I specify an `--input` idset that include ranks the job, I get a SIGPIPE error when input is processed.
```console
$ flux run --label-io --input=0-2 -N2 cat
foo
flux-job: attach_send_shell: Broken pipe
```
which is fatal to the job.  This seems like a reasonable reaction, but I thought I'd mention it in case there's a way to call out the fact that the `--input` argument was not specified correctly, to give the user a better hint about what went wrong.

grondo:
This was a tough one. The input plugin doesn't support easily returning an error message in that error path, though it could be extended (I just wondered if it was worth it at the time). We could print a more reasonable error in `flux job attach`, but not until the `shell.start` event when the taskmap is posted to the eventlog, and that is too late :thinking: 

I do think I see another issue in `flux job attach`, so let me try to fix that and ponder this one.

Oh, duh, the job shell can just log an error message when input is sent to a closed or invalid taskid...

grondo:
> which is fatal to the job. 

Oh, actually this is not fatal to the job, but it is fatal to `flux job attach`. That could be a potential problem since it might be unexpected in the `flux run` case that the job is potentially still running...

src/common/libtaskmap/taskmap.c
```diff
@@ -457,7 +457,7 @@ const struct idset *taskmap_taskids (const struct taskmap *map, int nodeid)
     struct taskmap_block *block;
     struct idset *taskids;
 
-    if (!map || nodeid < 0 || taskmap_unknown (map)) {
+    if (!map || nodeid < -1 || taskmap_unknown (map)) {
```
garlick, line 460: Would this be slightly more clear if we defined
```c
#define TASKMAP_NODEID_ALL (-1)
```
in the header for users?

doc/man1/common/submit-standard-io.rst
```diff
@@ -7,8 +7,10 @@ KVS, where they may be accessed with the ``flux job attach`` command.
...
 emitting the job's I/O to its stdout and stderr.
 
-**--input=FILENAME**
+**--input=FILENAME|RANKS**
    Redirect stdin to the specified filename, bypassing the KVS.
+   As a special case for ``flux run``, the argument may specify
+   a subset of task ranks to which to direct standard input.
```
garlick, line 13: Should this mention that RANKS is an idset?

garlick (approved):
Looks great!  I just a couple of minor suggestions for you to consider but I'd be fine if this went in as is.  Feel free to set MWP once you are happy with it.

src/common/libtaskmap/taskmap.c
grondo: On second thought, if that convenience parameter is confusing I'll just drop the change to libtaskmap and move the functionality into `flux job`.

grondo:
> Oh, actually this is not fatal to the job, but it is fatal to flux job attach. That could be a potential problem

How about if we delay the stdin watcher until the shell.start event in the case `--stdin-ranks` != "all", then `flux job attach` can just adjust any error on the user's part at runtime, avoiding the error exit:
```console
$ flux run -n2 --input=0-4 cat 
flux-job: warning: adjusting --stdin-ranks from 0-4 to 0-1
```

src/common/libtaskmap/taskmap.c
garlick: Sorry, the -1 is not particularly confusing, so I probably shouldn't have suggested that!

grondo: No, in retrospect that was a little kludgy and I didn't even add tests. Since there's only one use case, better to just do it by hand in `flux-job.c`. What I really want is a way to create a "filled" idset (combine idset_create() and idset_set_range() into a single call)

src/cmd/flux-job.c
```diff
@@ -1588,7 +1588,7 @@ struct attach_ctx {
     flux_jobid_t id;
```
garlick, line 1588: commit message for  095ed41af417cf78801bc611404b3c7f4b23fb5a:
s/prodided/provided/

garlick (approved):
looks good - just found one typo in a commit message.

src/cmd/flux-job.c
grondo: Thanks for catching that! I've fixed that and force pushed, and will set MWP.

PR #4978, opened 2023-03-04
flux-mini: emit deprecation warning

garlick:
Problem: the deprecation warning added to flux-mini in #4961 is not working.

This fixes it.  Does this seem appropriate/ok?
```
$  flux mini 
flux-mini: WARNING: ⚠️ flux-mini is deprecated, use flux-batch, flux-run, etc.⚠️
usage: flux-mini [-h] {run,submit,bulksubmit,batch,alloc} ...
flux-mini: error: the following arguments are required: subcommand
```
I was wondering about the non-ascii glyphs but they do get your attention and it's not like people are going to whine too much about a command they should be migrating away from :-)

### DIFF
--- a/src/cmd/flux-mini.py
+++ b/src/cmd/flux-mini.py
@@ -31,6 +31,8 @@ def main():
     sys.stderr = open(
         sys.stderr.fileno(), "w", encoding="utf8", errors="surrogateescape"
     )
+    if sys.stderr.isatty():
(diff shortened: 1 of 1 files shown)


grondo (approved):
Nice! I wondered how to do this without breaking everything, but I think the idea to only emit the warning when stderr is a tty is a great compromise!

garlick:
Thanks! I'll set MWP then.

PR #4979, opened 2023-03-05
shell: fix memory leak in doom plugin

grondo:
Problem: The incorrect JSON object is decref'd in doom_post() in the doom shell plugin, resulting in a leak of the outer `entry` object.

Call json_decref() on the correct object.

### DIFF
--- a/src/shell/doom.c
+++ b/src/shell/doom.c
@@ -113,7 +113,7 @@ static void doom_post (struct shell_doom *doom, json_t *task_info)
 
     flux_future_destroy (f); // fire and forget
     free (entrystr);
-    json_decref (task_info);
+    json_decref (entry);
     flux_kvs_txn_destroy (txn);
 }
 


garlick (approved):
Good catch!  Also a use after free when  `doom_notify_cb()` returns and `task_info` is freed along with the rest of the message payload!

grondo:
>  Also a use after free when doom_notify_cb() returns and task_info is freed along with the rest of the message payload!

I'm not sure there is a use after free there - the reference on `task_info` is incremented in `doom_post()` and decremented by `json_decref (entry)`. Thus, the `task_info` object should have a refcount of 1 when returning from `doom_notify_cb()`. However, I could be missing something. I did verify no errors from valgrind on this PR branch.

I'll set MWP.

garlick:
I might have misread then, sorry!

Issue #4947, opened 2023-02-16
broker.err[0]: proc_state_change_cb: illegal state: Stopped


garlick:
Problem: when a subprocess started by the broker subprocess server receives SIGSTOP, the server complains about an illegal state and destroys it.

I think STOPPED was added to libsubprocess (for use in the shell maybe?) but the remote side was not updated.

PR #4981, opened 2023-03-06
libsubprocess:  don't abort remote processes that receive SIGSTOP

garlick:
Problem: a remote subprocess that receives SIGSTOP is killed for entering an "illegal state" (#4947)

Handle the STOPPED state properly for remote processes.


grondo (approved):
LGTM!

PR #4982, opened 2023-03-06
flux-top: initialize f character before drawing panes

grondo:
Problem: The initial draw of the summary pane shows "(null)" instead of "ƒ" or "f" because the top->f_char is not initialized until *after* the panes are created.

Move the initial draw of top summary and job panes to after the f_char member is initialized.


garlick (approved):
Ah I noticed that and meant to look into it.  Great!

PR #4983, opened 2023-03-07
add flux-cancel(1)

grondo:
Since we now have commands to submit jobs at the top level of the `flux` command hierarchy, it seems a bit of an omission that we do not have a top level command to cancel jobs.

This PR adds a fairly simple `flux cancel` command. `flux cancel` contains a combination of the functionality from `flux job cancel` and `flux job cancelall`:
```
usage: flux cancel [OPTIONS] [JOBIDS...]

Cancel pending or running jobs with an optional exception note.

positional arguments:
  JOBIDS...            jobids to cancel

optional arguments:
  -h, --help           show this help message and exit
      --all            Target all jobs
  -n, --dry-run        Do not cancel any jobs
  -q, --quiet          Suppress output when no jobs match
  -u, --user=USER      Set target user or 'all' (instance owner only)
  -S, --states=STATES  List of job states to target (default=active)
  -m, --message=NOTE   Set optional cancel exception note
```

Note that the `--force` option has been replaced with `--dry-run`. This is probably a little more user friendly for the common case, since users will expect `flux cancel --all` to mean cancel all, and not have to doubly confirm that they want to cancel all their jobs.

For now, `flux job cancel` and `flux job cancelall` are left alone. They are probably much faster than the Python equivalent, so I'd hate to replace all uses in the testsuite and rc3 script with `flux cancel`. However, I'll attempt that now to determine if there's any fallout.

### DIFF
--- a/doc/man1/flux-cancel.rst
+++ b/doc/man1/flux-cancel.rst
@@ -1,0 +1,56 @@
+.. flux-help-description: cancel one or more jobs
+.. flux-help-section: jobs
+
+==============
+flux-cancel(1)
+==============
+
+
+SYNOPSIS
+========
+
+**flux** **cancel** [*OPTIONS*] [*JOBID...*]
+
+DESCRIPTION
+===========
+
+flux-cancel(1) cancels one or more jobs by raising a job exception of
+type=cancel. An optional message included with the cancel exception may be
+provided via the *-m, --message=NOTE* option. Canceled jobs are immediately
+sent SIGTERM followed by SIGKILL after a configurable timeout (default=5s).
+
+flux-cancel(1) can target multiple jobids by either taking them on the
+command line, or via the selection options *--all*, *-u, --user*, or *-S,
+--states=STATES*. It is an error to provide jobids on the command line
+and use one or more of the selection options.
+
+By default *--all* will target all jobs for the current user. To target all
+jobs for all users, use *--user=all* (only the instance owner is allowed
+to use *--user=all*). To see how many jobs flux-cancel(1) would kill,
+use the *-n --dry-run* option.
+
+OPTIONS
+=======
+
+**-n, --dry-run**
+   Do not cancel any jobs, but print a message indicating how many jobs
+   would have been canceled.
+
+**-m, --message=NOTE**
+   Set an optional exception note.
+
+**-u, --user=USER**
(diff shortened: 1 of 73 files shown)


doc/man1/flux-cancel.rst
```diff
@@ -0,0 +1,56 @@
...
+
+==============
+flux-cancel(1)
+==============
+
+
+SYNOPSIS
+========
+
+**flux** **cancel** [*OPTIONS*] [*JOBIDS..*]
```
garlick, line 12: Elipses has a missing period.

Maybe it would be good to use the form `[JOBID ...]` like in [flux-jobs(1)](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man1/flux-jobs.html)?

garlick (approved):
Looks straightforward to me!  Thanks!

grondo:
@ryanday36, would you take a quick look at the interface for `flux cancel` here and approve? Just let me know if you have any suggestions or changes (it will be easy to add new options going forward, but we probably can't change anything after this makes it into a release)

doc/man1/flux-cancel.rst
grondo: Fixed and force-pushed the result.

grondo:
Ok, I've updated `etc/rc1` and the testsuite to use `flux cancel` instead of `flux job cancel` and `flux job cancelall`. Tests are passing locally but we'll see about CI...

Pinging @vsoch also to approve interface of `flux cancel`...

src/cmd/flux-cancel.py
```diff
@@ -0,0 +1,187 @@
...
+        except OSError as exc:
+            errors += 1
+            LOGGER.error(f"{jobid}: {exc.strerror}")
+    if errors > 0:
+        sys.exit(1)
+
+
+def cancelall(args):
+
+    STATES = {
```
vsoch, line 101: This seems like logic that might be needed elsewhere - any reason to one-off define it here, or any reason to not define it in flux.constants? If it's just used here this should be OK - but if we find we are reproducing this elsewhere (which could lead to bugs) lets consider a single place to import from.

vsoch (approved):
Absolutely fantastic! I made a minor comment (a "think about for the future"), but I think it's good to go.

grondo:
Thanks! Setting MWP

PR #4984, opened 2023-03-08
completions: remove flux-mini and other updates

grondo:
It seems gauche to offer tab completion for `flux mini` when that command is deprecated, so this PR removes it from the bash tab completion config distributed with flux.

I found some other minor missing things and added those here as well.


garlick (approved):
LGTM!

PR #4990, opened 2023-03-10
prevent orphaned job processes when terminating jobs due to exception

### LINKED ISSUE #4987: not all job tasks are terminated after MPI_Abort
A user is reporting that after an `MPI_Abort` they need to manually clean up job processes.

> When our application dies ungracefully (generally an exception that gets caught and calls MPI_Abort), we end up with many processes still spinning. I can then run kill -s KILL $(ps -C [cmd] -h -o pid) and that cleans up like ~100 lingering processes (8 ranks + a bunch of threads/rank). I'm having trouble reproducing this outside of our application. Any suggestions for tracking down the cause of this? I don't see this behavior under the same circumstances with LSF or SLURM.

The application in question does have a SIGTERM handler which prints a backtrace then exits.

I'll try to make a standalone reproducer.

grondo:
This PR should fix #4987. It is WIP for now while I consider how to add a test.

Currently, the job shell is always terminated with SIGKILL after the configurable kill timeout in the job-exec module. This PR modifies the exec system to instead _request_ that the job shell send SIGKILL to the job tasks. This should result in many less orphan processes in the common case. Tasks can still be missed if they escape the process group set for them by the shell, but they will eventually be caught by future improvements to run first-level system jobs in a container.

The SIGKILL is sent open loop (since it is based on an event anyway), so I wonder if the exec system should keep trying as long as the job is still around. :thinking:

### DIFF
--- a/src/modules/job-exec/job-exec.c
+++ b/src/modules/job-exec/job-exec.c
@@ -359,11 +359,19 @@ static void kill_timer_cb (flux_reactor_t *r, flux_watcher_t *w,
                            int revents, void *arg)
 {
     struct jobinfo *job = arg;
+    flux_future_t *f;
     flux_log (job->h,
               LOG_DEBUG,
               "Sending SIGKILL to job %ju",
               (uintmax_t) job->id);
-    (*job->impl->kill) (job, SIGKILL);
+    if (!(f = flux_job_kill (job->h, job->id, SIGKILL))) {
+        flux_log_error (job->h,
+                        "flux_job_kill (%ju, SIGKILL)",
+                        job->id);
+        return;
+    }
+    /* Open loop */
+    flux_future_destroy (f);
 }
 
 
--- a/src/modules/job-exec/job-exec.c
+++ b/src/modules/job-exec/job-exec.c
@@ -136,6 +136,7 @@ void jobinfo_decref (struct jobinfo *job)
         idset_destroy (job->critical_ranks);
         eventlogger_destroy (job->ev);
         flux_watcher_destroy (job->kill_timer);
+        flux_watcher_destroy (job->kill_shell_timer);
         flux_watcher_destroy (job->expiration_timer);
         zhashx_delete (job->ctx->jobs, &job->id);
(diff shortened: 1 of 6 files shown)


garlick:
Excellent - thanks for running this down.  This ought to be really helpful!

I was wondering if we need a SIGKILL for the shell as a last resort but if we know the shell isn't blocking SIGTERM, then if it's still around, then it's probably blocked in the kernel and nothing's going to help anyway until that's resolved.

grondo:
>  was wondering if we need a SIGKILL for the shell as a last resort but if we know the shell isn't blocking SIGTERM, then if it's still around, then it's probably blocked in the kernel and nothing's going to help anyway until that's resolved.

This is a good point. I realized this could happen even if the shell is misbehaving or if it is replaced maliciously by a user. Then the shell would never get killed and the user may have continued access to resources (via PAM module for instance) until the issue is manually resolved.

Therefore, I've added a second timer here which sends SIGKILL to the job shell after a longer timeout (5x the initial timer or after another 20s by default) along with a test.

grondo:
Oh, I also did add a test for the original issue being resolved here, so removing WIP.

grondo:
I think there's a race condition in the new test that is causing it to fail in some of the CI builds. Let me look into that.

grondo:
The test issue seems to be resolved, so setting MWP.

src/modules/job-exec/job-exec.c
```diff
@@ -136,6 +136,7 @@ void jobinfo_decref (struct jobinfo *job)
         idset_destroy (job->critical_ranks);
```
garlick, line 136: In commit message for "terminate job shell with SIGKILL after timeout",
s/rpocess/process/

garlick (approved):
Looks good!  Just one commit message typo.

src/modules/job-exec/job-exec.c
grondo: Fixed that and forced a push. Will set MWP.

PR #4993, opened 2023-03-11
python: improve Hostlist class indexing

grondo:
This simple PR improves the indexing of `Hostlist` objects by

 * returning a new `Hostlist` when more than one host is indexed, e.g. by slice (before, a `list` of hosts was returned)
 * allowing any iterable to be used with a Hostlist index, most notably a tuple, e.g. `hl[1,3,5]` and an idset `hl[ids]`.


garlick (approved):
Wow that's really neat!

grondo:
Thanks! I'll set MWP.

PR #4995, opened 2023-03-15
testsuite: require jq(1)

grondo:
This PR makes `jq` a hard requirement for all sharness tests, and then removes all the related `HAVE_JQ`, `jq` wrappers, and `TEST_CHECK_PREREQS` uses in the testsuite, which were all there just to make sure `HAVE_JQ` was used appropriately.

Now test authors can use `jq` in tests without having to worry if they've set prereqs properly etc.


garlick (approved):
LGTM!

PR #4996, opened 2023-03-16
add deploy to GitHub packages

vsoch:
Problem: Docker might be taking away OSS project repos 
Solution: Do a double deploy to include ghcr.io as a backup!

I'm not super familiar with how the push is done - but I found a script that generates the build matrix, and the deploy shell script, and made changes to both.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -91,6 +91,8 @@ jobs:
       DOCKER_USERNAME: travisflux
       DOCKER_PASSWORD: ${{ secrets.DOCKER_HUB_TRAVISFLUX_TOKEN }}
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      GITHUB_ACTOR: ${{ github.actor }}     
+      GITHUB_PACKAGE_REPO: ghcr.io/flux-framework/flux-core
     timeout-minutes: 60
     strategy:
       matrix: ${{fromJson(needs.generate-matrix.outputs.matrix)}}
--- a/src/test/docker-deploy.sh
+++ b/src/test/docker-deploy.sh
@@ -10,26 +10,32 @@ if test "$GITHUB_REPOSITORY" != "flux-framework/flux-core"; then
     exit 0
 fi
 
-test -n "$DOCKER_REPO" ||     die "DOCKER_REPO not set"
-test -n "$DOCKER_PASSWORD" || die "DOCKER_PASSWORD not set"
-test -n "$DOCKER_USERNAME" || die "DOCKER_USERNAME not set"
-test -n "$DOCKER_TAG" ||      die "DOCKER_TAG not set"
+test -n "$DOCKER_REPO"         || die "DOCKER_REPO not set"
+test -n "$DOCKER_PASSWORD"     || die "DOCKER_PASSWORD not set"
+test -n "$DOCKER_USERNAME"     || die "DOCKER_USERNAME not set"
+test -n "$DOCKER_TAG"          || die "DOCKER_TAG not set"
+test -n "$GITHUB_TOKEN"        || die "GITHUB_TOKEN not set"
+test -n "$GITHUB_ACTOR"        || die "GITHUB_ACTOR not set"
+test -n "$GITHUB_PACKAGE_REPO" || die "GITHUB_PACKAGE_REPO not set"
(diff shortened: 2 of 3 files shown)


grondo:
> I'm not super familiar with how the push is done - but I found a script that generates the build matrix, and the deploy shell script, and made changes to both.

I think you got it right. Thanks! Should we move the `fluxrm/testenv` images to ghcr as well? These are manually created and pushed to docker hub as necessary. Then, in another PR we could switch the CI to pull from github instead of docker hub.

vsoch:
Yeah! So if they are manually created - do you want to push them to `ghcr.io/flux-framework/testenv` there and then I'm guessing we change the FROM in the Dockerfile headers? I'm good to update this PR or just do another one.

vsoch:
It looks like it's mostly the .devcontainer and then various documentation:
```
$ grep -R testenv
vscode.md:and select your cloned Flux repository root. This will build a development environment from [fluxrm/testenv](https://hub.docker.com/r/fluxrm/testenv/tags)
grep: src/bindings/lua/flux.so: No such file or directory
src/test/docker/README.md:#### fluxrm/testenv Docker images
src/test/docker/README.md:under the `fluxrm/testenv:bionic`, `fluxrm/testenv:focal`,
src/test/docker/README.md:`fluxrm/testenv:el7`, and `fluxrm/testenv:el8` respectively, and
src/test/docker/README.md:to customize the `fluxrm/testenv` before building. Without this secondary
src/test/docker/README.md:would not make sense to be constantly updating the base `fluxrm/testenv`
src/test/docker/README.md:Later, a flux-core maintainer can move the dependency into the `testenv`
src/test/docker/README.md:pushed to DockerHub at `fluxrm/testenv:bionic` and
src/test/docker/README.md:`fluxrm/testenv:el7`. Be sure to test that the `docker-run-test.sh`
src/test/docker/README.md:script still runs against the new `testenv` images, e.g.:
src/test/docker/README.md:    docker build --no-cache -t fluxrm/testenv:$i src/test/docker/$i &&
src/test/docker/README.md:    docker push fluxrm/testenv:$i
src/test/docker/README.md:$  docker buildx build --push --platform=linux/386,linux/amd64 --tag fluxrm/testenv:bionic src/test/docker/bionic
src/test/docker/README.md:docker build -t fluxrm/testenv:$image src/test/docker/$image
src/test/docker/docker-run-checks.sh:BASE_DOCKER_REPO=fluxrm/testenv
NEWS.md: * docker: small updates for testenv images, travis builds (#3058)
.devcontainer/Dockerfile:FROM fluxrm/testenv:focal
```

grondo:
It will take about a day to stage the base testenv images to ghcr because of my ADSL connection. Let me see if I can do that first.

BTW, does ghcr.io allow read only, unauthenticated access to images?

vsoch:
> BTW, does ghcr.io allow read only, unauthenticated access to images?

Yes it does. When you push it will default to private in most cases (unless you change the default) and then you can change the visibility in the container settings. Also note that if you manually push a container and then actions needs to push, the actions runner will need to be given permissions. If you push first from actions, given the actions runner in the repository permissions has that setting, it should work the first time.

I guess in simpler terms if there is any issue just ping me! :laughing:

vsoch:
Just circling back - can I rebase here and then set merge on pass? Is this still desired?

I noticed the same setup for the flux-sched repo - I could do the same thing I did here there (if there is interest)!

vsoch:
@grondo do you want to do this soon? I propose we finish up this work and then remove the ubuntu builds here: https://github.com/flux-framework/flux-core/actions/runs/4592479440/jobs/8109553376

grondo:
Sorry, this fell off my todo list since Docker retracted their plan to delete OSS project repos! I can try to work on this after our release this week.

vsoch:
Ping! @grondo I think when we have separate testing repos, we will want these ghcr images, and built nightly so we can regularly test against the latest flux here. Do you have time to take a look sometime this week?

grondo:
I was just thinking that maybe these ghcr actions could take care of pulling the docker hub `fluxrm/testenv:*` images and push them to ghcr.io? Then we'd only have to keep the base images up to date in one place?

vsoch:
Yes! We absolutely could do that. We wouldn't just want to push, but maybe have the testenv build in another automated place. Where are they currently built from?

PR #4998, opened 2023-03-17
fix link dependencies for job-manager modules

trws:
Not sure why this only appears on nixos, but the submit-hold, alloc-bypass and perilog modules were all missing necessary link dependencies.  I think this was hiding behind our normal ignore missing symbols behavior for modules and the load order on other systems happening to load the libjob and libflux-internal libraries before the modules on most platforms.

This is the first part of the fixup stuff from my poking at flux on NixOS, but regardless of whether we want the other changes to let our tests pass without a FHS-compatible root, I'm pretty sure we'll want this.

### DIFF
--- a/src/modules/job-manager/Makefile.am
+++ b/src/modules/job-manager/Makefile.am
@@ -12,6 +12,12 @@ AM_CPPFLAGS = \
 	-I$(top_srcdir)/src/common/libccan \
 	$(JANSSON_CFLAGS)
 
+COMMON_LIBS = \
+	libjob-manager.la \
+	$(top_builddir)/src/common/libjob/libjob.la \
+	$(top_builddir)/src/common/libflux-internal.la \
+	$(top_builddir)/src/common/libflux-core.la
+
 fluxmod_LTLIBRARIES = \
 	job-manager.la
 
@@ -100,17 +106,13 @@ plugins_submit_hold_la_LDFLAGS = \
 	$(fluxplugin_ldflags) \
 	-module
 plugins_submit_hold_la_LIBADD = \
-	$(top_builddir)/src/common/libjob/libjob.la \
-	$(top_builddir)/src/common/libflux-internal.la \
-	libjob-manager.la
+	${COMMON_LIBS}
 
 plugins_alloc_bypass_la_SOURCES = \
 	plugins/alloc-bypass.c
 plugins_alloc_bypass_la_LIBADD = \
-	$(top_builddir)/src/common/libflux-internal.la \
 	$(top_builddir)/src/common/librlist/librlist.la \
-	$(top_builddir)/src/common/libjob/libjob.la \
-	libjob-manager.la
+	${COMMON_LIBS}
 plugins_alloc_bypass_la_LDFLAGS = \
 	$(fluxplugin_ldflags) \
 	-module
@@ -118,10 +120,8 @@ plugins_alloc_bypass_la_LDFLAGS = \
 plugins_perilog_la_SOURCES = \
 	plugins/perilog.c
 plugins_perilog_la_LIBADD = \
-	$(top_builddir)/src/common/libflux-internal.la \
 	$(top_builddir)/src/common/librlist/librlist.la \
-	$(top_builddir)/src/common/libjob/libjob.la \
-	libjob-manager.la
+	${COMMON_LIBS}
 plugins_perilog_la_LDFLAGS = \
 	$(fluxplugin_ldflags) \
 	-module
--- a/src/modules/job-manager/Makefile.am
+++ b/src/modules/job-manager/Makefile.am
@@ -99,12 +99,18 @@ plugins_submit_hold_la_SOURCES = \
 plugins_submit_hold_la_LDFLAGS = \
 	$(fluxplugin_ldflags) \
 	-module
+plugins_submit_hold_la_LIBADD = \
+	$(top_builddir)/src/common/libjob/libjob.la \
+	$(top_builddir)/src/common/libflux-internal.la \
+	libjob-manager.la
 
 plugins_alloc_bypass_la_SOURCES = \
 	plugins/alloc-bypass.c
 plugins_alloc_bypass_la_LIBADD = \
 	$(top_builddir)/src/common/libflux-internal.la \
-	$(top_builddir)/src/common/librlist/librlist.la
+	$(top_builddir)/src/common/librlist/librlist.la \
+	$(top_builddir)/src/common/libjob/libjob.la \
+	libjob-manager.la
 plugins_alloc_bypass_la_LDFLAGS = \
 	$(fluxplugin_ldflags) \
 	-module
@@ -114,7 +120,8 @@ plugins_perilog_la_SOURCES = \
 plugins_perilog_la_LIBADD = \
 	$(top_builddir)/src/common/libflux-internal.la \
 	$(top_builddir)/src/common/librlist/librlist.la \
-	$(top_builddir)/src/common/libjob/libjob.la
+	$(top_builddir)/src/common/libjob/libjob.la \
+	libjob-manager.la
 plugins_perilog_la_LDFLAGS = \
 	$(fluxplugin_ldflags) \
 	-module


grondo:
I haven't figured out why yet, but this PR seems to break both the flux-sched and flux-accounting builds. Distressingly, the flux-sched tests hang forever.

The only necessary symbols for these plugins should be from `libflux-core.so`, i.e. all these plugins _should_ be able to be built out-of-tree actually, so I think all the libtool convenience libraries should probably be dropped from LIBADD. This might work now everywhere else because the broker is linked with `libflux-core.so`, so the plugins can find those symbols on `dlopen(3)`.

If an explicit link needs to be made for libflux-core.so, then maybe `$(top_builddir)/src/common/libflux-core.la` would be sufficient?

What's a bit interesting about the job-manager plugins is that we have a DSO dlopening another DSO and I can't think that through on a Friday, nor am I expert enough on this subject to make any other salient comments at present.

grondo:
Well, I think the CI issues with flux-sched/accounting is probably because this branch is fairly out of date.

Maybe retry after rebase on current master.

But I do also think we should only need to link these plugins with libflux-core.so. Since I can't reproduce the issue (we haven't seen this issue in any other environment), I can't test it.

trws:
I’ll rebase to it, no problem there. Maybe that by itself will change the linking issues, but there were symbols in several of them that came from czmq (without linking it) libjob and others. Will send something easier to reproduce if it keeps up after the rebase I suppose, though one simple way to see what’s missing would be to turn off the flag that allows undefined symbols and see what blows up.


---

On March 18, 2023 at 3:42:03 PM PDT, Mark Grondona ***@***.***> wrote:

Well, I think the CI issues with flux-sched/accounting is probably because this branch is fairly out of date.

Maybe retry after rebase on current master.

But I do also think we should only need to link these plugins with libflux-core.so. Since I can't reproduce the issue (we haven't seen this issue in any other environment), I can't test it.

grondo:
Oh, yeah. I forgot about the libutil stuff. Sorry!

src/modules/job-manager/Makefile.am
```diff
@@ -12,6 +12,12 @@ AM_CPPFLAGS = \
 	-I$(top_srcdir)/src/common/libccan \
 	$(JANSSON_CFLAGS)
 
+COMMON_LIBS = \
+	job-manager.la \
```
grondo, line 16: was this supposed to be `libjob-manager.la`?

The asan build is failing with 
```
 *** Warning: Linking the executable test_job.t against the loadable module
  *** job-manager.so is not portable!
  /usr/bin/ld: test/job_t-job.o: in function `test_create':
  /usr/src/src/modules/job-manager/test/job.c:25: undefined reference to `job_create'
  /usr/bin/ld: /usr/src/src/modules/job-manager/test/job.c:28: undefined reference to `ok_at_loc'
```

Looks like somehow the link to libtap for the tests was dropped as well?

src/modules/job-manager/Makefile.am
```diff
@@ -129,11 +136,8 @@ TESTS = \
 	test_annotate.t
 
 test_ldadd = \
-	libjob-manager.la \
+	${COMMON_LIBS}
 	$(top_builddir)/src/common/libtap/libtap.la \
```
trws, line 140: Test link to libtap is here.

src/modules/job-manager/Makefile.am
trws: This was intentionally `job-manager.la` so that the modules would like to the dynamic library.  If I'm reading this right, `libjob-manager.la` is the static library that gets linked into `job-manager.la`, and the linker script exposes all the symbols the modules need so this seemed safer to avoid duplication.

src/modules/job-manager/Makefile.am
trws: Oh no, I see what you mean, not sure why that worked, I'll add the backslash and amend this.  Thanks!

src/modules/job-manager/Makefile.am
grondo: Yes, I can see the sense in that, but the test programs specifically should be linked against the `libjob-manager.la` since they are just unit tests.

Also, I've never tried to link one plugin DSO against another so I'd be surprised if there weren't issues there. I'm still wondering what symbols internal to the job-manager anything besides the unit tests are using? (Since that breaks the separation between the job-manager and its jobtap plugins, I should probably fix that.)

grondo: I think you could probably just revert the changes to `test_ldadd`. The unit tests are purposefully built to pull in code from the current directory and avoid dynamic linking as much as practical. Were there any specific issues with the unit tests on nixos? Seems like the issue was instead with the 3 external job-manager plugins.

grondo: > I can definitely use the static lib in the tests, but to clarify you want the static lib there to embed the symbols in the binary rather than using the SO like we use in production?

I admit I don't really understand the issue you are having (a bit above my level of understand of dynamic loading), so if you say the job-manager/plugins need to be linked against the `job-manager.so` I trust you. However, the warning is a bit concerning (and annoying since it comes up in every build). Is there really no other way? What symbols were missing on dlopen()?

> I can definitely use the static lib in the tests, but to clarify you want the static lib there to embed the symbols in the binary rather than using the SO like we use in production?

Yes, the tests are unit tests so they are specifically testing the internal API of the individual source files.

grondo: Do you have a recipe to reproduce the issue on NixOS?

I just remembered we have some other job-manager plugins used in in test under `t/job-manager/plugins` -- any sense of why those do not have the same issue exhibited by the plugins that are under `src/modules/job-manager/plugins`?

trws:
I can give you the flake and lockfile which should reproduce it.  The 
symbols that come up missing are listed out here, this is attempting to 
run `./t2208-job-manager-wait.t -v`

```
2023-03-17T06:03:19.239325Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/submit-hold.so: 
undefined symbol: flux_jobtap_event_post_pack
2023-03-17T06:03:19.239362Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/alloc-bypass.so: 
undefined symbol: flux_jobtap_raise_exception
2023-03-17T06:03:19.239375Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/perilog.so: 
undefined symbol: flux_jobtap_raise_exception
2023-03-17T06:03:19.310158Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/submit-hold.so: 
undefined symbol: flux_jobtap_event_post_pack
2023-03-17T06:03:19.310208Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/alloc-bypass.so: 
undefined symbol: flux_jobtap_raise_exception
2023-03-17T06:03:19.310219Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/perilog.so: 
undefined symbol: flux_jobtap_raise_exception
2023-03-17T06:03:19.352520Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/submit-hold.so: 
undefined symbol: flux_jobtap_event_post_pack
2023-03-17T06:03:19.352565Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/alloc-bypass.so: 
undefined symbol: flux_jobtap_raise_exception
2023-03-17T06:03:19.352577Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/perilog.so: 
undefined symbol: flux_jobtap_raise_exception
2023-03-17T06:03:19.390158Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/submit-hold.so: 
undefined symbol: flux_jobtap_event_post_pack
2023-03-17T06:03:19.390196Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/alloc-bypass.so: 
undefined symbol: flux_jobtap_raise_exception
2023-03-17T06:03:19.390214Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/perilog.so: 
undefined symbol: flux_jobtap_raise_exception
2023-03-17T06:03:19.461891Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/submit-hold.so: 
undefined symbol: flux_jobtap_event_post_pack
2023-03-17T06:03:19.461948Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/alloc-bypass.so: 
undefined symbol: flux_jobtap_raise_exception
2023-03-17T06:03:19.461971Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/perilog.so: 
undefined symbol: flux_jobtap_raise_exception
2023-03-17T06:03:19.506164Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/submit-hold.so: 
undefined symbol: flux_jobtap_event_post_pack
2023-03-17T06:03:19.506218Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/alloc-bypass.so: 
undefined symbol: flux_jobtap_raise_exception
2023-03-17T06:03:19.506234Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/perilog.so: 
undefined symbol: flux_jobtap_raise_exception
2023-03-17T06:03:19.580037Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/submit-hold.so: 
undefined symbol: flux_jobtap_event_post_pack
2023-03-17T06:03:19.580099Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/alloc-bypass.so: 
undefined symbol: flux_jobtap_raise_exception
2023-03-17T06:03:19.580125Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/perilog.so: 
undefined symbol: flux_jobtap_raise_exception
2023-03-17T06:03:19.616724Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/submit-hold.so: 
undefined symbol: flux_jobtap_event_post_pack
2023-03-17T06:03:19.616791Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/alloc-bypass.so: 
undefined symbol: flux_jobtap_raise_exception
2023-03-17T06:03:19.616816Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/perilog.so: 
undefined symbol: flux_jobtap_raise_exception
2023-03-17T06:03:19.700886Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/submit-hold.so: 
undefined symbol: flux_jobtap_event_post_pack
2023-03-17T06:03:19.700937Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/alloc-bypass.so: 
undefined symbol: flux_jobtap_raise_exception
2023-03-17T06:03:19.700956Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/perilog.so: 
undefined symbol: flux_jobtap_raise_exception
2023-03-17T06:03:19.738438Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/submit-hold.so: 
undefined symbol: flux_jobtap_event_post_pack
2023-03-17T06:03:19.738489Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/alloc-bypass.so: 
undefined symbol: flux_jobtap_raise_exception
2023-03-17T06:03:19.738514Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/perilog.so: 
undefined symbol: flux_jobtap_raise_exception
2023-03-17T06:03:19.778193Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/submit-hold.so: 
undefined symbol: flux_jobtap_event_post_pack
2023-03-17T06:03:19.778249Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/alloc-bypass.so: 
undefined symbol: flux_jobtap_raise_exception
2023-03-17T06:03:19.778280Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/perilog.so: 
undefined symbol: flux_jobtap_raise_exception
2023-03-17T06:03:19.878696Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/submit-hold.so: 
undefined symbol: flux_jobtap_event_post_pack
2023-03-17T06:03:19.878754Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/alloc-bypass.so: 
undefined symbol: flux_jobtap_raise_exception
2023-03-17T06:03:19.878781Z broker.err[0]: rc1.0: flux-module: 
/olddisk/home/scogland1/Repos/flux-core/src/modules/job-manager/plugins/.libs/perilog.so: 
undefined symbol: flux_jobtap_raise_exception
expecting success:
```

src/modules/job-manager/Makefile.am
trws: I can use libjob-manager for all of these and it will work, that gets rid of the warning because they're no longer linking to an so with a non-canonical name.  That means we may end up with some amount of binary bloat, but the loader will de-duplicate at runtime and everything will continue to work.  Here, I'll push that version so we can see what the CI output looks like.

trws:
@grondo: any idea what's going on with codecov here?  It's showing indirect drops in coverage all over the place, increases in others in ways this patch can't have caused.  I've seen this happen on spack when there are issues with multiple coverage submissions, but haven't seen it with flux, is this normal?

grondo:
And yes, we also seem to get random small changes in coverage all over the place on a normal basis. I'm not sure what the issue is there, but we usually ignore "impossible" coverage changes.

trws:
Thanks @grondo that makes sense.  If the coverage stuff is ok, then this is probably ready for a final review, this version works with the static lib and seems to pass all the tests without warnings, so if you're happy with it I'm good.

grondo:
> I can give you the flake and lockfile which should reproduce it.  The
symbols that come up missing are listed out here, this is attempting to
run `./t2208-job-manager-wait.t -v`

I'd love to get the flake that reproduces this error. Looking back at the errors you pasted above, they are coming from `flux module` and not `flux jobtap`. I think these errors are meaningless, and are just an artifact of `flux module` traversing _all_ `.so` files when it tries to find a module, since it is recursively searching all directories under `$builddir/src/common` when running tests.

Therefore, I actually think these errors could be ignored, unless they caused some failure in the actual test?

This also explains why the plugins built in `t/job-manager/plugins` do not generate any errors, since they are not in the `FLUX_MODULE_PATH`.

Perhaps we can try a modification to the testsuite `FLUX_MODULE_PATH` instead as a workaround for this issue.

trws:
On 24 Mar 2023, at 8:20, Mark Grondona wrote:

> I'd love to get the flake that reproduces this error. Looking back at the errors you pasted above, they are coming from `flux module` and not `flux jobtap`. I think these errors are meaningless, and are just an artifact of `flux module` traversing _all_ `.so` files when it tries to find a module, since it is recursively searching all directories under `$builddir/src/common` when running tests.
>
> Therefore, I actually think these errors could be ignored, unless they caused some failure in the actual test?

They did, the tests fail unless they are fixed.

> This also explains why the plugins built in `t/job-manager/plugins` do not generate any errors, since they are not in the `FLUX_MODULE_PATH`.
>
> Perhaps we can try a modification to the testsuite `FLUX_MODULE_PATH` instead as a workaround for this issue.
If that works that would be fine by me.

flake to come in a second.

trws:
This has a bunch of extra dev stuff in it, but I normally keep both the flake.nix and flake.lock in `flux-core/nix` and use it like this:

```console
# for a non-FHS compatible shell, NixOS native
$ nix develop "path:$(pwd)/nix"
# for a FHS-compatible shell, NixOS in non-NixOS clothing
$ nix develop "path:$(pwd)/nix#fhs"
```

It has some extra stuff in it you probably don't need, but I'm hesitant to mess with it when the goal is a reproducer.

[flake.zip](https://github.com/flux-framework/flux-core/files/11064732/flake.zip)

grondo:
Ok, I found the real source of the issue here. We're hitting what's described in this NixOS issue: 

 https://github.com/NixOS/nixpkgs/issues/25796

NixOS by default adds the `-z now` flag, so our use of `RTLD_LAZY` when using `dlopen` on these modules is ignored. Thus the dynamic linker errors.

I verified by running in a shell with `hardeningDisable = "bindnow"` set, e.g.

```
 nix-shell -E '((import <nixpkgs> {}).mkShell { hardeningDisable = ["bindnow"]; })'
```

which prevents these errors.

I'm also thinking the errors are only causing testsuite failures because they get mixed in with expected output and thus trigger false failures. I'll look into that and fix it. For now, could you add the above `hardeningDisable` flag to the flux-core nix flake? I couldn't figure out exactly where it goes.

Long term, we should probably set the `FLUX_MODULE_PATH` more explicitly  in the build tree to point to just the module directories, which would allow builds with `-z now` to work without these errors.

grondo:
Besides the removal of `-z now` I did find some other spurious test failures under nix. Proposed fixes in #5015. Many of the tests that use systemd were also failing for me under nix, so I suggest also disabling systemd by unsetting `XDG_RUNTIME_DIR` for the duration of the testsuite.

trws:
Sounds good, and I got the hardening change added to the flake.  I'll close this in favor of #5015 and some subsequent PRs.  The flake is actually more or less working all the way for me locally now.  Will post some of the more orthogonal test changes, then maybe the flake itself and test changes so we can discuss.

PR #5000, opened 2023-03-18
flux-resource: support overwrite of drain timestamp with `--force --force`

grondo:
This PR adds a `force-overwrite` "mode" to the job-manager drain RPC which overwrites the drain reason _and_ timestamp.

Use of `--force` twice in `flux resource drain` then sets this mode, such that `flux resource drain -ff TARGETS REASON` will update both timestamp and reason, instead of just the drain reason.

### DIFF
--- a/doc/man1/flux-resource.rst
+++ b/doc/man1/flux-resource.rst
@@ -83,7 +83,8 @@ COMMANDS
    By default, **flux resource drain** will fail if any of the *targets*
    are already drained. To change this behavior, use either of the
    *-f, --force* or *-u, --update* options. With *--force*, the *reason* for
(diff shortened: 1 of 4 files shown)


garlick (approved):
Looks good!  However, the commit title and description for "job-manager: drain: add force-overwrite mode" should really reference the resource module not the job manager.

grondo:
> the commit title and description for "job-manager: drain: add force-overwrite mode" should really reference the resource module not the job manager.

D'oh! Sorry! Will fix.

grondo:
Thanks, I've fixed the issues in the commit messages and will set MWP now.

PR #5001, opened 2023-03-21
cleanup: improve reusability of common message handlers

garlick:
This adds a new variant of `flux_msg_handler_addvec()` that accepts a service name:
```c
int flux_msg_handler_addvec_ex (flux_t *h,
                                const char *service_name,
                                const struct flux_msg_handler_spec tab[],
                                void *arg,
                                flux_msg_handler_t **hp[]);
```
The service name (if non-NULL) is prepended to the topic strings defined as `const` strings in the handler array at registration.  This makes it easier to create reusable services that register an array of methods under different service names.

Broker modules now use this to register the common methods like `stats-get`, `ping`, and `rusage` under the module name.  Those methods are moved into libflux (though not exposed in the public API) so that things like the shell could add them if it were considered helpful.

P.S.  the `stats` protocol was also modernized a bit:  Avoiding crazy JSON key names like `#request (tx)`, and replacing the period with a hyphen in method names.

P.P.S. this will be useful if we decide to embed the subprocess server in the shell since it registers an array of methods internally.


garlick:
Thanks!  Per offline discussion, I'll move the methods to a new local library "libfluxutil" where we can drop shared bits that depend on libflux-core.so.

PR #5002, opened 2023-03-22
fix: warning message to user should be actual command

vsoch:
Problem: the deprecation warning for flux-mini tells the user about the underlying command, which is not actually what they would execute!

Solution: Fix the warning to reference "flux mini" and "flux (submit|batch|run) etc. so it makes sense.

I was doing some testing of flux with merlin tonight and was (very happy) to see this warning, but I think we could improve it slightly by showing the user the command they would actually run, and not the underlying actual filename. 
image

### DIFF
--- a/src/cmd/flux-mini.py
+++ b/src/cmd/flux-mini.py
@@ -34,10 +34,10 @@ def main():
     if sys.stderr.isatty():
         cmd = sys.argv[1] if len(sys.argv) > 1 else None
         if cmd in ["submit", "run", "batch", "alloc", "bulksubmit"]:
-            suggestion = f"flux {sys.argv[1]} "
+            suggestion = f"flux {cmd} "
         else:
             suggestion = "flux submit, flux run, etc."
-        LOGGER.warning(f"⚠️ flux-mini is deprecated, use {suggestion}⚠️")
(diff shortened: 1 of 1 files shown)


src/cmd/flux-mini.py
```diff
@@ -34,10 +34,10 @@ def main():
     if sys.stderr.isatty():
         cmd = sys.argv[1] if len(sys.argv) > 1 else None
         if cmd in ["submit", "run", "batch", "alloc", "bulksubmit"]:
-            suggestion = f"flux {sys.argv[1]} "
```
vsoch, line 37: Also I'm wondering why the warning suggestion still showed `flux-batch` (with the hyphen) when "cmd" checks for the term without it.

grondo (approved):
It is a little strange that the screenshot in your initial comment contains output from an older version, but the changes here look good.

Though you might want to update the commit message to avoid confusion since the warning message already does reference `flux batch`, `flux run` etc.

garlick:
@vsoch - this is approved so if you want to set the merge-when-passing label, Mergify will rebase/merge automatically.

vsoch:
looks like the merge train failed - so I rebased manually.

garlick:
I restarted the asan builder - one of our tests failed that is probably a race of some kind, clearly nothing to do with this PR.

PR #5003, opened 2023-03-22
cleanup: allow subprocess service name to be configured

garlick:
This PR modifies the subprocess server so a service name (currently `rexec`) can be specified when the server is instantiated and then adds  `flux_rexec_ex()`, a variant of `flux_rexec()` with an additional service name parameter.

Internally, the "remote" portion of libsubprocess is cleaned up so that the RPC protocol is handled exclusively in client.c and server.c

This is prep work for
- embedding the subprocess server in the shell (needed once shells are launched in a limited cgroups environment)
-  #5004
- allowing subprocess clients to communicate with a new compatible server that starts processes using systemd

### DIFF
--- a/src/common/libsubprocess/test/iochan.c
+++ b/src/common/libsubprocess/test/iochan.c
@@ -232,7 +232,7 @@ int main (int argc, char *argv[])
 
     plan (NO_PLAN);
 
-    h = rcmdsrv_create ();
+    h = rcmdsrv_create ("rexec");
 
     if (flux_set_default_subprocess_log (h, tap_logger, NULL) < 0)
         BAIL_OUT ("could not set logger");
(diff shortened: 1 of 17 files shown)


src/common/libsubprocess/test/iochan.c
```diff
@@ -232,7 +232,7 @@ int main (int argc, char *argv[])
 
     plan (NO_PLAN);
 
-    h = rcmdsrv_create ();
```
grondo, line 235: commit message typo: Updat -> Update

grondo (approved):
LGTM! Just spotted one typo in a commit message.

garlick:
Thanks!  Fixed that one and will set MWP.

PR #5006, opened 2023-03-23
cleanup: improve interface for subprocess logging

garlick:
Problem: it was probably a misstep to add `flux_set_default_subprocess_log()` instead of adding logger parameters directly to the subprocess functions.

Add `flux_local_exec_ex()` that takes an `llog` function.

Drop the `hooks` argument from `flux_local_exec()` since it is not widely used (leave it in the `_ex()` version)

Drop `flux_exec()` since the only difference from `flux_local_exec()` was taking a `flux_t *` instead of a reactor, so that it could call `flux_log()` internally, which it no longer does.

Add the logger parameters to `subprocess_server_create()`, and have subprocesses inherit the logger from the server.

### DIFF
--- a/src/common/libsubprocess/subprocess.c
+++ b/src/common/libsubprocess/subprocess.c
@@ -481,12 +481,13 @@ static int subprocess_setup_completed (flux_subprocess_t *p)
     return 0;
 }
 
-static flux_subprocess_t * flux_exec_wrap (flux_t *h, flux_reactor_t *r, int flags,
-                                           const flux_cmd_t *cmd,
-                                           const flux_subprocess_ops_t *ops,
-                                           const flux_subprocess_hooks_t *hooks,
-                                           subprocess_log_f log_fn,
-                                           void *log_data)
+flux_subprocess_t *flux_local_exec_ex (flux_reactor_t *r,
+                                       int flags,
+                                       const flux_cmd_t *cmd,
+                                       const flux_subprocess_ops_t *ops,
+                                       const flux_subprocess_hooks_t *hooks,
+                                       subprocess_log_f log_fn,
+                                       void *log_data)
 {
     flux_subprocess_t *p = NULL;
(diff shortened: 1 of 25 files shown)


src/common/libsubprocess/subprocess.c
```diff
@@ -546,12 +546,20 @@ flux_subprocess_t * flux_exec (flux_t *h, int flags,
     return flux_exec_wrap (h, r, flags, cmd, ops, hooks);
 }
 
+flux_subprocess_t *flux_local_exec_ex (flux_reactor_t *r,
```
grondo: > Problem: flux_local_exec() takes a "hooks" parameter that is rarely used.

I only see 4 places (besides tests) that use `flux_local_exec()`. Two use the hooks parameter and two do not. I may be missing something that makes half the usage "rarely used", but is the churn here really worth it? Most of the change is introduced in the unit tests, but maybe this change makes some future changes cleaner?

grondo: Ah, maybe this was necessary to add a function that takes the llog parameters anyway... (now that I'm viewing the next commit)

grondo (approved):
This looks like good cleanup to me!

src/common/libsubprocess/subprocess.c
garlick: That's a fair point.  I guess it would have been fine to leave the hooks parameter in `flux_local_exec()` and my claim of rarity is on thin ice.  Should I move it back there?

garlick: Just pushed with typo fixed and a more honest claim in the commit that creates `flux_local_exec_ex()`.  I could easily move `hooks` back to flux_local_exec() if that's better.

grondo: Nah, after seeing the other ex-tra parameters added to `flux_local_exec_ex()`, it seems fine to me!

garlick:
Thanks!  I'll go ahead and set MWP then.

PR #5007, opened 2023-03-23
librlist: drop V1 flag from hwloc XML generation

### LINKED ISSUE #5005: drop `HWLOC_TOPOLOGY_EXPORT_XML_FLAG_V1` when exporting hwloc XML
The librlist  `rhwloc_local_topology_xml()` function passes the `HWLOC_TOPOLOGY_EXPORT_XML_FLAG_V1` flag when exporting the topology to XML under hwloc 2.x, thus forcing the XML for all users of this function to V1

https://github.com/flux-framework/flux-core/blob/f168dec941c46458a216486e06c3e2abf1bcea0e/src/common/librlist/rhwloc.c#L114-L119

The original reason for this flag is lost to us, but it was probably put in place when first updating Flux to support hwloc 2.x, and at the time perhaps there were external users (e.g. flux-sched) which could still be using hwloc 1.x, and it was important to allow those users to continue to read the topology gathered by Flux.

Now that hwloc 2.x is more ubiquitous and the XML gathered by the flux `resource` module has less users, it may be possible to drop this flag.

In fact, we now have external users like `mpibind` which assume hwloc 2.x and the V1 flag may be detrimental to those integrations.

We should consider dropping this flag unless there are objections. 

cc; @trws 
cc: @eleon

grondo:
Problem: The librlist rhwloc_local_topology_xml() function passes the HWLOC_TOPOLOGY_EXPORT_XML_FLAG_V1 flag when exporting the topology to XML under hwloc 2.x, thus forcing the XML for all users of this function to the hwloc v1 format. This includes the XML stored in the resource module and available via resource.topo-get.

Now that hwloc 2.x is more ubiquitous and the XML gathered by the flux resource module has less users, it may be possible to drop this flag.

In fact, we now have external users like mpibind which assume hwloc 2.x and the V1 flag may be detrimental to those integrations.

Drop the HWLOC_TOPOLOGY_EXPORT_XML_FLAG_V1 flag when generating XML from a v2 topology so that v2 XML is available.

Fixes #5005

### DIFF
--- a/src/common/librlist/rhwloc.c
+++ b/src/common/librlist/rhwloc.c
@@ -112,8 +112,7 @@ char *rhwloc_local_topology_xml (rhwloc_flags_t rflags)
     if (topo == NULL)
         return (NULL);
 #if HWLOC_API_VERSION >= 0x20000
-    int flags = HWLOC_TOPOLOGY_EXPORT_XML_FLAG_V1;
-    if (hwloc_topology_export_xmlbuffer (topo, &buf, &buflen, flags) < 0) {
+    if (hwloc_topology_export_xmlbuffer (topo, &buf, &buflen, 0) < 0) {
 #else
     if (hwloc_topology_export_xmlbuffer (topo, &buf, &buflen) < 0) {
 #endif


garlick (approved):
I'm for it if it passes CI (and I think you already confirmed that it does).

grondo:
Thanks! I set MWP.

PR #5010, opened 2023-03-23
testsuite: speed up t4000-issues-test-driver.t

grondo:
Problem: The t4000-issues-test-driver.t runs individual tests serially, and now that there are many issues tests to run, the time to run the test is getting very long - exceeding the 5 minute CI time limit in some cases.

Alter the design of the t4000-issues-test-driver.t test to run each issue reproducer script as a job. Since some scripts themselves run jobs, execute each script under a single-core Flux instance.

Adjust one test that only works with multiple cores to cancel a running job before submitting a new one.

### DIFF
--- a/t/t4000-issues-test-driver.t
+++ b/t/t4000-issues-test-driver.t
@@ -10,7 +10,7 @@ if test_have_prereq ASAN; then
     test_done
 fi
 
-SIZE=4
+SIZE=2
 test_under_flux ${SIZE}
 echo "# $0: flux session size will be ${SIZE}"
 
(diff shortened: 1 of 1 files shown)


grondo:
This gives a >2x speedup of the `t4000-issues-test-driver.t` test on my local 4 core system. It should offer some reduction of runtime in CI, though I'm unsure how much. If there is still an issue with runtime of this test, we could look at splitting up the issues tests into separate sharness tests.

garlick (approved):
LGTM!

grondo:
Thanks, setting MWP.

PR #5011, opened 2023-03-24
job-manager: ensure epilog-start event prevents resource release for job that never started

grondo:
@jameshcorbett reported an issue where an `epilog-start` event was emitted for a job in CLEANUP, but the `free` and `clean` events were still emitted before `epilog-finish`. This only seems to occur if a job enters CLEANUP without ever having started, e.g. when a job exception occurs during the job prolog.

The problem occurs because the `epilog-start` event is still pending in the job manager event queue when the action for CLEANUP runs. Since the job didn't start, and `perilog_active` is still 0, the `free` event is emitted immediately instead of waiting for `epilog-finish`.

This PR simply adds a check for a pending `epilog-finish` event that is enqueued.

A new test is added to replicate the failing case.

### DIFF
--- a/src/modules/job-manager/event.c
+++ b/src/modules/job-manager/event.c
@@ -407,6 +407,7 @@ int event_job_action (struct event *event, struct job *job)
              * it is safe to release all resources to the scheduler.
              */
             if (job->has_resources
+                && !job_event_is_queued (job, "epilog-start")
                 && !job->perilog_active
                 && !job->alloc_bypass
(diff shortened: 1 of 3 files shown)


jameshcorbett (approved):
My flux-coral2 tests are passing and the code changes look reasonable. Thanks @grondo !

garlick (approved):
LGTM!

grondo:
Thanks all, will set MWP.

PR #5012, opened 2023-03-24
fix rare `Unable to connect to JOBID` error from `flux alloc --bg`

grondo:
Problem: The broker posts a job memo containing the instance remote uri too early during startup. This creates a race condition where the job uri is available in the eventlog before the local connector has created the socket, resulting in a rare failure to connect using tools such as `flux alloc --bg`.

Move the function that notifies any parent instance of subinstance particulars (including instance uri) to after the connector is loaded.


garlick (approved):
That makes sense. Thanks!

PR #5014, opened 2023-03-24
improve broker debugging on overlay connect failure

garlick:
Problem: if a broker cannot connect to its upstream peer, `flux dmesg` doesn't work, and if the problem is a misconfigured hostname, nothing is logged anyway.

Add dmesg to the list of services that are available "early" (before upstream connect).

Add a small check on the upstream URI before handing it over to zmq_connect().

If I comment out the upstream peer in /etc/hosts on my test cluster, now I can do this:
```
$ sudo flux dmesg
2023-03-24T22:52:26.438913Z broker.err[7]: getaddrinfo picl2: No address associated with hostname
2023-03-24T22:52:26.459197Z broker.debug[7]: insmod connector-local
2023-03-24T22:52:26.459352Z broker.info[7]: start: none->join 18.249ms
2023-03-24T22:52:26.460967Z connector-local.debug[7]: allow-guest-user=true
2023-03-24T22:52:26.460997Z connector-local.debug[7]: allow-root-owner=true
```
And incidentally if I fix /etc/hosts, the broker is able to connect and continue on without a restart.

### DIFF
--- a/src/broker/overlay.c
+++ b/src/broker/overlay.c
@@ -32,6 +32,7 @@
 #include "src/common/libutil/errprintf.h"
 #include "src/common/librouter/rpc_track.h"
 #include "src/common/libccan/ccan/ptrint/ptrint.h"
+#include "src/common/libyuarel/yuarel.h"
 
 #include "overlay.h"
 #include "attr.h"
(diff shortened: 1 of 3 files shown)


src/broker/overlay.c
```diff
@@ -1249,13 +1250,44 @@ static void parent_monitor_cb (struct zmqutil_monitor *mon, void *arg)
...
+        || !streq (u.scheme, "tcp"))
+        goto done;
+    /* N.B. this URI will be used for zmq_connect(), therefore it must
+     * be a valid peer address, not an interface name or wildcard.
+     */
+    if ((e = getaddrinfo (u.host, NULL, NULL, &result)) == 0) {
+        freeaddrinfo (result);
+        goto done;
+    }
+    flux_log (h, LOG_ERR, "getaddrinfo %s: %s", u.host, gai_strerror (e));
```
grondo, line 1278: Just one quick comment before a full review: Users and admins may appreciate a more descriptive message here like: "Warning: unable to resolve upstream peer %s: %s"?

garlick: Yeah probably :-)  Will fix.

garlick: Force pushed a fix to that one.  Good call!

grondo (approved):
LGTM!

PR #5015, opened 2023-03-26
testsuite: fix test issues under nix

grondo:
This PR fixes a few spurious errors in the testsuite caused when running the nix package manager. These are mostly fixes that would have popped up elsewhere anyway.

This does not go as far as replacing every use of absolute paths to commands (`/bin/sh`, `/bin/bash`) yet, but this allows most tests to pass under the flake shared by @trws in #4998.

Some systemd tests still fail for unknown reasons. It is best to disable the tests that use systemd when running under nix by unsetting `XDG_RUNTIME_DIR` during the testsuite, e.g.
```
XDG_RUNTIME_DIR= make -j 12 check
```

### DIFF
--- a/t/t2613-job-shell-batch.t
+++ b/t/t2613-job-shell-batch.t
@@ -51,11 +51,14 @@ test_expect_success 'flux-shell: historical batch jobspec still work' '
 	for spec in $SHARNESS_TEST_SRCDIR/batch/jobspec/*.json; do
 		input=$(basename $spec) &&
 		cat $spec |
-		    jq -S ".attributes.system.environment.PATH=\"$PATH\"" \
+		    jq -S ".attributes.system.environment.PATH=\"$PATH\"" |
+		    jq -S ".attributes.system.environment.PYTHONPATH=\"$PYTHONPATH\"" |
+		    jq -S ".attributes.system.environment.HOME=\"$HOME\"" |
+		    jq -S ".attributes.system.cwd=\"$(pwd)\"" \
 		    >$input &&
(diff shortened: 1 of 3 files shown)


trws:
I have fixes for the systemd tests too I think, though we might actually want to do a bit more there.  The trick I ran into was after the first failure the systemd tests the suite didn't clean out the unit, and was then failing because it couldn't re-create it.  Will post that stuff.

Also, THANK YOU! The issue with t2613 was the last one I couldn't figure out after the rebase.  What was it?

t/t2613-job-shell-batch.t
```diff
@@ -51,11 +51,13 @@ test_expect_success 'flux-shell: historical batch jobspec still work' '
 	for spec in $SHARNESS_TEST_SRCDIR/batch/jobspec/*.json; do
 		input=$(basename $spec) &&
 		cat $spec |
-		    jq -S ".attributes.system.environment.PATH=\"$PATH\"" \
+		    jq -S ".attributes.system.environment.PATH=\"$PATH\"" |
+		    jq -S ".attributes.system.environment.PATH=\"$PATH\"" |
```
trws, line 55: Dup? Otherwise this looks great.

grondo:
Rebased and force-pushed a fix to also add `HOME` to the set of environment variables added to the old batch jobspec in `t2613` (based on comments in #5100)

trws (approved):

grondo:
Thanks @trws! Setting MWP.

PR #5017, opened 2023-03-27
testsuite: increase test timeouts

chu11:
Not sure what happened in the last month but nice chunk of tests were failing / timing out for me when doing big parallel `make -j16 check` runs.

I'm sure it has nothing to do with the tests I updated here specifically, but its whatever other tests are running in parallel and eating up tons of CPU.  Perhaps some new tests are really CPU heavy and thus hitting these specifically.  The timeout of 180 was picked somewhat randomly based on surrounding tests.

If this is approved, I'll actually wait a few days to hit MWP.  See if any more pop up.

Edit: maybe the t4000 tests now running in parallel is causing this?  It launches up a flux instance of two nodes within it, and each node can have many CPUs, thus running many t4000 tests in parallel?

### DIFF
--- a/t/t2701-mini-batch.t
+++ b/t/t2701-mini-batch.t
@@ -101,7 +101,7 @@ test_expect_success MULTICORE 'flux-mini batch: exclusive flag worked' '
 
 test_expect_success 'flux-mini batch: --output=kvs directs output to kvs' '
 	id=$(flux mini batch -n1 --flags=waitable --output=kvs batch-script.sh) &&
-	run_timeout 60 flux job attach $id > kvs-output.log 2>&1 &&
+	run_timeout 180 flux job attach $id > kvs-output.log 2>&1 &&
 	test_debug "cat kvs-output.log" &&
 	grep "size=1 nodes=1" kvs-output.log
 '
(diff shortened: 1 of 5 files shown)


grondo:
> Edit: maybe the t4000 tests now running in parallel? It launches up a flux instance of two nodes within it, and each node can have many CPUs, thus running many t4000 tests in parallel?

Yes, see #5010. In CI we were actually seeing the t4000 test run over the 5 minute overall timeout per-sharness test that `tap-driver.sh` enforces.

chu11:
re-pushed, test in `t2702` also timed out recently.  so bumped timeouts in there as well.

grondo (approved):
Seems good to me, though I haven't experienced these timeouts.
Still not sure what was the question about the t4000 tests?

chu11:
>  Still not sure what was the question about the t4000 tests?

Basically I never hit these timeouts a month ago, so I was wondering would might have started causing them.  My best guess at something increasing business was the parallelization of t4000 tests.  It's just a guess of course.  For all I know the systems I'm running on are just busier too :P

Just for kicks, I did a `make -j16 check` on the commit right before the t4000 change and hit these timeouts.  So it wasn't that change.

chu11:
for kicks i tried to go hunting to see if some testsuite change caused this for me.   It appears #4961 was the offender.  Not sure what it could be.  Perhaps underneath every call to the new commands python is searching a bunch of paths for python libs to load?  And that's putting some stress on stuffs?

PR #5018, opened 2023-03-27
cli: adjust description of `--begin-time` submission option

grondo:
As suggested by @v and @ryanday36, the documentation of the `--begin-time` option doesn't make it clear that the argument can be either a relative offset or an absolute time. This PR adjusts both the help output and manual page to hopefully make that more clear.


garlick (approved):
LGTM!

PR #5020, opened 2023-03-27
libflux: remove extraneous +1s used in buffers

### LINKED ISSUE #3880: libflux/flog: remove unnecessary + 1 on buffers
Noticed in libflux/flog that code like this is used a lot of places.

```
    char buf[FLUX_MAX_LOGBUF + 1];
    (void)vsnprintf (buf, sizeof (buf), fmt, ap);
```

Typically the +1 is used to ensure that the buffer will always be NUL terminated, since `snprintf()` (and family) may not NUL terminate.  But if the size of the buffer passed to `snprintf()` is always `sizeof (buf)`, then it doesn't matter.  Instead `FLUX_MAX_LOGBUF` should be passed in.

Edit:

Not sure what would be best.  Probably?

```
char buf[LOGBUF + 1];
(void)vsnprintf (buf, LOGBUF, fmt, ap);
buf[LOGBUF] = '\0';
```

chu11:
Problem: An internal buffer added +1 to the buffer, but didn't use it in any meaningful way.

Remove the extra +1 that isn't needed.

Fixes #3880

### DIFF
--- a/src/common/libflux/flog.c
+++ b/src/common/libflux/flog.c
@@ -33,7 +33,7 @@
 typedef struct {
     char appname[STDLOG_MAX_APPNAME + 1];
     char procid[STDLOG_MAX_PROCID + 1];
-    char buf[FLUX_MAX_LOGBUF + 1];
+    char buf[FLUX_MAX_LOGBUF];
     flux_log_f cb;
     void *cb_arg;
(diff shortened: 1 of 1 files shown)


garlick (approved):
Thanks!

PR #5022, opened 2023-03-27
flake and non-fhs test rework

trws:
This is mainly here for discussion right now, everything seems to work except t2613-job-shell-batch.t including building, installing and non-fhs-wrapped development shell usage for flux core.


garlick:
This is pretty out of date at this point.  Let's close the PR.  Please resubmit an updated one if any of the commits here should have been pulled in.

PR #5023, opened 2023-03-27
doc: document FLUX_URI_RESOLVE_LOCAL

chu11:
Problem: The environment FLUX_URI_RESOLVE_LOCAL is not documented.

Add info about it in flux-uri(1).

----

I couldn't remember the environment variable and had to dig to find it.  So decided we should stick it in `flux-uri(1)`.


grondo (approved):
LGTM, thanks for catching this! One suggestion inline, but take it or leave it.

chu11:
@grondo thanks for tweak, added it and re-pushed.  Will set MWP.

PR #5024, opened 2023-03-27
doc: document job environment variables

### LINKED ISSUE #4568: Document environment variables set for jobs
A user was asking if we have a documented list of environment variables similar to

 https://slurm.schedmd.com/srun.html#SECTION_OUTPUT-ENVIRONMENT-VARIABLES

Flux probably has way less environment variables than that, but we should collect and document them somewhere.

chu11:
Problem: The environment variables that Flux creates when a job is run is not documented.

Add documentation into flux-run(1) and flux-submit(1).

Fixes #4568

----

I largely cut and pasted the text that @grondo had written up in the issue.  Only debate might be where I put this in the manpages.  Could be higher/lower than where I put it?

### DIFF
--- a/doc/man1/common/submit-job-environment-variables.rst
+++ b/doc/man1/common/submit-job-environment-variables.rst
@@ -1,0 +1,29 @@
+JOB ENVIRONMENT VARIABLES
+=========================
+
+Flux creates several environment variables for every job task.  They are as follows:
+
+**FLUX_JOB_ID**
+    The jobid for this job
+
+**FLUX_URI**
+    The URI of the enclosing Flux instance
+
+**FLUX_JOB_TMPDIR**
+    Path to temporary directory created for job
+
+**FLUX_TASK_LOCAL_ID**
+    The task id local to the node
+
+**FLUX_TASK_RANK**
+    The global rank for this task
+
+**FLUX_JOB_NNODES**
+    The total number of nodes in the job
+
+**FLUX_JOB_SIZE**
+    The count of job shells in the job, typical the same value as FLUX_JOB_NNODES
+
+**FLUX_KVS_NAMESPACE**
+    The identifier of the KVS guest namespace of the job. This redirects KVS command and
+    API calls to the job guest namespace instead of the root namespace.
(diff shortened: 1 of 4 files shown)


chu11:
oops, had a spelling error, re-pushed

doc/man1/common/submit-job-environment-variables.rst
```diff
@@ -0,0 +1,29 @@
...
+
+**FLUX_JOB_NNODES**
+    The total number of nodes in the job
+
+**FLUX_JOB_SIZE**
+    The count of job shells in the job, typical the same value as FLUX_JOB_NNODES
+
+**FLUX_KVS_NAMESPACE**
+    The identifier of the KVS guest namespace of the job. This redirects KVS command and
+    API calls to the job guest namespace instead of the root namespace.
```
grondo, line 29: Should we also document `FLUX_JOB_CC`, set if `--cc` was used? It is slightly different from these, and is already mentioned in the `--cc` documentation (so doesn't apply to `flux run`), so maybe not, but thought I'd bring it up.

grondo (approved):
Nice! LGTM!

garlick:
Not sure if this is a bad idea or not but I'll throw it out there - do we want a man page just for "flux environment variables"?  Like `flux-environment(7)`?

I seem to recall somebody asked "what environment vars does flux use" or similar at one point.  One stop shopping for that question might not be a bad thing.

chu11:
> Not sure if this is a bad idea or not but I'll throw it out there - do we want a man page just for "flux environment variables"? Like flux-environment(7)?
>
> I seem to recall somebody asked "what environment vars does flux use" or similar at one point. One stop shopping for that question might not be a bad thing.

I guess it couldn't hurt, although it might be a pain to maintain.  Would said manpage also document uhhh "input" environment variables?  like `FLUX_URI_RESOLVE_LOCAL`?

chu11:
oops, re-pushed fixing a distcheck bug

garlick:
> I guess it couldn't hurt, although it might be a pain to maintain. Would said manpage also document uhhh "input" environment variables? like FLUX_URI_RESOLVE_LOCAL?

True.  Let's table that idea for now.

PR #5025, opened 2023-03-27
python: recognize local timezone epoch adjustment

chu11:
Problem: The "h" output presentation type will output a hyphen (-) if a timestamp field is equal to the epoch of 1970-01-01T00:00:00. However, this does not account for any local timezone offsets that may occur when outputting the epoch time.

Solution: In addition to checking 1970-01-01T00:00:00, get the local epoch time and add it to the list of potential "zero" things that
can be converted to a hyphen.


grondo (approved):
Nice!

Issue #5021, opened 2023-03-27
flux-top: queue filter outputs on wrong line


chu11:
Noticed that `flux-top` output can output jobs on the wrong line. e.g.

All queues listed

![image](https://user-images.githubusercontent.com/274859/228038986-9ecea81d-3990-45a5-ab9f-1483a2984222.png)

filtering just batch queue.

![image](https://user-images.githubusercontent.com/274859/228039075-faace29c-af44-48ee-9ae4-7d243d7a9331.png)

Notice the empty line between the header and the first job.  That's where the "debug" queue job would be.

There's some logic errors because some code still operates on indices of the "whole joblist" vs the filtered joblist specific to a queue.

PR #5032, opened 2023-03-28
flux-top: fix queue specific output display

chu11:
Per #5021, the listing of jobs when filtered by specific queue did not display correctly because job indexing was not done correctly.

### DIFF
--- a/src/cmd/top/joblist_pane.c
+++ b/src/cmd/top/joblist_pane.c
@@ -370,15 +370,33 @@ void joblist_pane_set_current (struct joblist_pane *joblist, bool next)
     if (joblist->current != FLUX_JOBID_ANY)
         index = lookup_jobid_index (joblist->jobs, joblist->current);
 
+    /* find next valid index */
     njobs = json_array_size (joblist->jobs);
-    if (next && index == njobs - 1)
-        index = -1;
-    else if (!next && index <= 0)
-        index = njobs;
+    while (1) {
+        int next_index = next ? index + 1 : index - 1;
+        if (next_index == njobs)
+            next_index = 0;
+        else if (next_index < 0)
+            next_index = njobs - 1;
 
-    if (!(job = json_array_get (joblist->jobs,
-                                next ? index + 1 : index - 1)))
-        return;
+        if (!(job = json_array_get (joblist->jobs, next_index)))
+            return;
+
+        if (joblist->top->queue) {
+            const char *queue = "";
+            if (json_unpack (job, "{s?s}", "queue", &queue) < 0)
+                fatal (0, "error decoding a job record from job-list RPC");
+            if (!streq (joblist->top->queue, queue)) {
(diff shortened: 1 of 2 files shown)


chu11:
re-pushed, tweaking how FLUX_URI_RESOLVE_LOCAL is set/unset per a comment in #5052

src/cmd/top/joblist_pane.c
```diff
@@ -370,15 +370,33 @@ void joblist_pane_set_current (struct joblist_pane *joblist, bool next)
     if (joblist->current != FLUX_JOBID_ANY)
         index = lookup_jobid_index (joblist->jobs, joblist->current);
 
+    /* find next valid index */
     njobs = json_array_size (joblist->jobs);
-    if (next && index == njobs - 1)
-        index = -1;
-    else if (!next && index <= 0)
-        index = njobs;
+    while (1) {
```
grondo, line 375: Just an idle question. Would it be simpler to create a separate "filtered" list of jobs and use the simpler selection code to select among the displayed jobs? It might be useful going forward if `flux top` ever filters jobs with something other than queues (e.g. by user), since the same selection code will work no matter how the list is filtered.

grondo (commented):
One suggestion inline. However, I haven't examined the code carefully to know if my suggestion is feasible.

src/cmd/top/joblist_pane.c
chu11: at the time I just didn't want to make a copy of the whole joblist, b/c currently its an incref copy from what was returned from the joblist query.

```
    if (flux_rpc_get_unpack (f, "{s:o}", "jobs", &jobs) < 0) {                                                                              
        if (errno != ENOSYS)                                                                                                                
            fatal (errno, "error decoding job-list.list RPC response");                                                                     
        flux_future_destroy (f);                                                                                                            
        return;                                                                                                                             
    }                                                                                                                                       
    json_decref (joblist->jobs);                                                                                                            
    joblist->jobs = json_incref (jobs);         
```

But you make a good point that such a filtered list is a good "investment" for easier future filtering.  Lemme take a crack at it.

grondo: Could you use a zlistx as a filtered "view" into the list of all jobs? Not at a computer, sorry if that's a silly suggestion

chu11: Looking through the code, I'd like to look at the "filtered job list" as a refactor after PR #5052.  With the queue abstraction that I placed in that PR I think it'd be easier.  I did ponder refactoring both this PR and that PR to put the queue abstraction "earlier" in the commit series, but I figure it's probably less work overall to do it after the two PRs together.

grondo: Seems reasonable to me!

grondo (approved):
LGTM!

PR #5039, opened 2023-03-29
resource: do not allow ranks to be both drained and excluded

### LINKED ISSUE #4999: resource: do not allow a rank to be both excluded and drained
Currently, excluded resources can also be drained with `flux resource drain`. While this has no effect, it causes problems for #4997, which assumes that `exclude`, `draining`, and `drained` are all exclusive states.

Since draining resources is a dynamic way to remove them from consideration by the scheduler, perhaps it does not make sense to allow draining of excluded nodes.

Offline, @garlick noted the other piece here would be to handle the case where a drained node becomes excluded by a future configuration change. Perhaps the resource module can just ignore the previous drain state in this case. (Possibly related: #4971)

grondo:
This PR attempts to prevent excluded ranks from being drained since `flux resource status` now assumes a node/rank is only in one of the states avail, exclude, drained, or draining.

Ensuring this required changes in multiple places - I'm not sure if there was a cleaner way to handle this:

 * exclude idset is now checked in the drain request handler and an error generated if any target is excluded
 * if a node is excluded on replay, any drain events for that rank are ignored
 * if a node is excluded at runtime, any drain state for that rank is cleared

Since drain initialization now requires access to the exclude idset, the resource module initialization had to be reordered a bit. It was actually a bit difficult to determine the dependencies between the resource module bits, so I tried to add some comments on what I found. 

Fixes #4999

### DIFF
--- a/doc/man1/flux-resource.rst
+++ b/doc/man1/flux-resource.rst
@@ -109,6 +109,10 @@ COMMANDS
    only those ranks that are not already drained or do not have a *reason* set
    have their *reason* updated.
 
+   Resources cannot be both excluded and drained, so **flux resource drain**
+   will also fail if any *targets* are currently excluded by configuration.
+   There is no option to force an excluded node into the drain state.
+
    This command, when run with arguments, is restricted to the Flux instance
    owner.
 
--- a/doc/man5/flux-config-resource.rst
+++ b/doc/man5/flux-config-resource.rst
@@ -62,6 +62,9 @@ exclude
    the scheduler, but will still be used to determine satisfiability of job
    requests until the instance is restarted.
 
+   If a drained node is subsequently excluded, the drain state of the node
+   is cleared since nodes cannot be both excluded and drained.
+
 norestrict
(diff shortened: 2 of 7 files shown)


garlick (approved):
Nice work.  I'm fine with this going in as proposed.
I will try a few things on my test cluster just for fun though if you want to hold off a bit.

garlick:
OK  I tried the following on my test cluster
- drain an excluded node (fails as expected)
- drain a node then exclude it (no longer drained)
- drain a node, shutdown flux, exclude the node, then start (no longer drained)
- shutdown flux, unexclude the node, then start (still not drained)

Seems like all is well!

The one thing I noticed was this INFO message may be a little out of place, although it's harmless if too difficult to suppress
```
 garlick@picl0:~/proj/flux-core$ sudo flux resource drain picl0 xyz
flux-resource: ERROR: [Errno 17] rank 0 excluded
flux-resource: INFO: Use --force to overwrite existing drain reason
```

grondo:
Setting MWP. Thanks for the extra testing and review @garlick!

PR #5043, opened 2023-03-29
flux-pgrep: fix warning about sre_constants on Python 3.11+

grondo:
The `sre_constants` module has been depreacated in latter Python versions. Turns out we didn't need it anyway.


chu11 (approved):
LGTM!

Issue #4488, opened 2022-08-11
support a flux overlay whatsup command


chu11:
Per discussion in #4487.  From @garlick

> I like the idea of flux overlay whatsup as another way to view cluster health, if they would find that helpful given the caveats.

> An idset and/or hostlist representing up and down ranks can be easily generated from groups info tracked at rank 0.

debate on if it should be called `flux overlay whatsaup` or `flux whatsup` or `flux up` (revived old command)

chu11:
per #5044 closing, as it is obsolete given newer things added to `flux resource status`

Issue #4487, opened 2022-08-11
support flux "up vs down" module so whatsup can work with flux


chu11:
I was suddenly reminded that we discussed this like 6 or 7 years ago.  it would be nice if whatsup (https://github.com/chaos/whatsup) could determine node up vs. down status via flux, since flux will be sending around a heartbeat very similarly to other mechanisms that whatsup works with (ganglia, cerebro).  On systems with TOSS this may allow us to remove the need for cerebro.

It might even be possible for users to use whatsup in sub-instances too if they'd like.

Off the top of my head:

A) overlay health might be close to having everything whatsup needs, although I'm not sure if this is the right mechanism to use long term, since it's possible people may set up multiple brokers per node.  Or if a broker goes down and takes down children brokers, whatsup won't have the right information.

B) what might be best is to write a module that simply listens for flux heartbeats then supplies that info to whatsup along w/ most recent heartbeat timestamp and provides that to whatsup.  Could also write a streaming rpc to supply up/down info so whatsup can support the `--monitor` option with flux.  Not sure if module would need to map rank to hostname or if whatsup could do that outside of the module via other flux calls, TBD later.

garlick:
Maybe it would be useful to add a whatsup(1) compatible subcommand to [flux-overlay(1)](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man1/flux-overlay.html)?

An idset and/or hostlist representing up and down ranks can be easily generated from groups info tracked at rank 0.

However, besides the issues with using flux that you mention, I think whatsup(1) and cerebro's simplicity are their main advantage.  Flux might be part of the problem that whatsup(1) helps diagnose.

garlick:
I like the idea of `flux overlay whatsup` as another way to view cluster health, if they would find that helpful given the caveats.

BTW,  this statement is not quite accurate:

> since flux will be sending around a heartbeat very similarly to other mechanisms that whatsup works with (ganglia, cerebro).

The heartbeat is sent by rank 0 to the other ranks over the tree, and it's for synchronization not monitoring.  It's not like cerebro where each rank sends out a multicast beacon.

grondo:
> I like the idea of `flux overlay whatsup` as another way to view cluster health, if they would find that helpful given the caveats.

I'd actually suggest `flux whatsup` in this case if there isn't a strong reason to put it under the `flux overlay` command

chu11:
[...]

PR #5044, opened 2023-03-29
flux-overlay: add whatsup subcommand

chu11:
Per discussion in #4487 and #4488

Nothing too fancy, get the group set from the rank 0 broker and do some idset math to figure out up vs down.  Maybe don't need to support ALL of those options, that's what leads to a lot of the extra output logic.  But kept it in there for consistency to `whatsup(1)`.

### DIFF
--- a/src/cmd/builtin/overlay.c
+++ b/src/cmd/builtin/overlay.c
@@ -13,6 +13,7 @@
 #endif
 #include <unistd.h>
 #include <jansson.h>
+#include <math.h>
 #include <flux/core.h>
 
 #include "src/common/libccan/ccan/str/str.h"
@@ -21,6 +22,8 @@
 #include "src/common/libutil/fsd.h"
 #include "src/common/libhostlist/hostlist.h"
 #include "src/common/librlist/rlist.h"
+#include "src/common/libidset/idset.h"
+#include "src/common/libhostlist/hostlist.h"
 
 #include "builtin.h"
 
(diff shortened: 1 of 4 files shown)


grondo:
Those issue numbers don't seem to point to the right issues?

chu11:
oops! yeah, adjusting numbers

src/cmd/builtin/overlay.c
```diff
@@ -919,6 +938,136 @@ static int subcmd_disconnect (optparse_t *p, int ac, char *av[])
...
+
+    if (ranks) {
+        up_str = idset_encode (idset_up, IDSET_FLAG_RANGE);
+        down_str = idset_encode (idset_down, IDSET_FLAG_RANGE);
+    }
+    else {
+        hl_up = idset2hostlist (h, idset_up);
+        hl_down = idset2hostlist (h, idset_down);
+        up_str = hostlist_encode (hl_up);
+        down_str = hostlist_encode (hl_down);
```
grondo, line 1022: Use `flux_hostmap_lookup(3)` (you'll have to create `up_str` and `down_str` for the hostlist case, but then you can remove the local `idset2hostlist()` function which mostly duplicates the functionality of `flux_hostmap_lookup(3)`.

grondo (commented):
One note inline about removing an internal function that duplicates and existing API call.

Also, there's some valid critical warnings about non-constant format strings.

Maybe for this implementation we don't need the in-place computed formats and can just a constant width for the count of up/down nodes to silence those warnings and simplify the code? (Also drop the need to check for `log10`)

chu11:
re-pushed with fixes per comments above.  Thanks, I didn't see `flux_hostmap_lookup()`.  

I decided against the fixed width, it just looked too weird to me.  I just went with unaligned output instead.  e.g. 

```
up: 15: fake[0-14]
down: 0:
```

also added a manpage entry, forgot about that with the first cut.

chu11:
re-pushed, missed some `&&` in test chaining

chu11:
> Just coming back to this, and I apologize if this is a naive question, but what is this command providing vs flux resource status, 

The original idea was to offer a simple "whatsup" based on just overlay data, not anything else as a quick status check.  https://github.com/flux-framework/flux-core/issues/4487#issuecomment-1212127903

The issue was brought up in August 2022, long before many `flux resource status` work was done.  .  So maybe it doesn't provide a super ton other than just something simple / familiar now.   I think the online/offline calculation in `module/resource` effectively does the same logic?

Won't be super offended if we consider this duplicate and should drop it.  If a user really would like this as an alternate "whats up or down" in this format, could also be added into `whatsup(1)` as a backend option there instead.

garlick:
We should probably decide if this is in or out rather than let it languish and accumulate merge conflicts.

I guess I would vote :-1: because I can't see how it offers any information that's not already in `flux resource status`, and we've already marketed that one in our docs.

If that ends up being the consensus then I do apologize for suggesting this a while ago..

chu11:
ok, lets close it then.

PR #5045, opened 2023-03-29
mpi: drop mvapich.lua plugin

### LINKED ISSUE #5041: setting LD_LIBRARY_PATH to flux's PMI directory considered harmful?
Problem: on cray systems, cray MPI and other runtimes need to find the cray pmi library before flux's, but flux (via the mvapich.lua plugin) sets LD_LIBRARY_PATH to include flux's PMI libdir.

Consider not doing that.

It won't affect flux launching flux or anything else configured to use the PMI simple wire protocol since they never go looking for a PMI library to dlopen.

It won't affect openmpi version < 5, since the old [flux mca plugin](https://github.com/open-mpi/ompi/blob/v4.1.x/opal/mca/pmix/flux/pmix_flux.c#L142) gets the dso path from the FLUX_PMI_LIBRARY_PATH environment variable.

It may affect certain mvapich configurations (e.g. on TOSS 3 systems).

It is the mvapich plugin that is setting this environment variable.  Perhaps we can drop this setting from the plugin since it should not be required when mvapich is configured to allow the wire protocol to work.

garlick:
Problem: the mvapich.lua plugin prepends the directory containing Flux's PMI library to LD_LIBRARY_PATH, but this causes problems on Cray systems where the system PMI library must be found first.

The LD_LIBRARY_PATH setting seems to only be required for mvapich linked directly against libpmi2.so, as was the case on TOSS 3 systems where mvapich packages predated Flux.  The mvapich configuration that we recommend in our FAQ works fine without this setting, and indeed without the entire mvapich.lua plugin being loaded.

Drop the mvapich.lua plugin.
Update t3001-mpi-personalities.t test.

Fixes #5041


grondo (approved):
Excellent! Removing code, I love it!

Issue #4801, opened 2022-12-06
add flux job purge ID [ID ...][


garlick:
Problem: if an INACTIVE job entry in the KVS contains something that prevents a restart from occuring, or is for some reason taking up an outsize amount of space, it may be helpful to remove it.  However manually removing it doesn't purge it from the job manager or from `job-list`.  The manual remove is also a little awkward, e.g.  `flux kvs unlink -R $(flux job --to=kvs ID)`

If we extended `flux job purge` to accept job IDs, then we could  add a check that the job really is inactive, and we could post the `job-purge-inactive` journal event and clear it from the job manager inactive hash (and zombie hash in the wait logic, if waitable).

PR #5047, opened 2023-03-30
flux-job: support job ids to purge

chu11:
Per #4801.  I decided to add a new `purge-id` subcommand instead of having `purge` take optional ID arguments.  That way its similar to `list` vs `list-id`.

### DIFF
--- a/doc/man1/flux-job.rst
+++ b/doc/man1/flux-job.rst
@@ -31,7 +31,7 @@ SYNOPSIS
 
 **flux** **job** **timeleft** [*OPTIONS*] [*id*]
 
-**flux** **job** **purge** [*OPTIONS*]
+**flux** **job** **purge** [*OPTIONS*] [*id...*]
 
 DESCRIPTION
 ===========
@@ -247,7 +247,8 @@ PURGE
 =====
 
 Inactive job data may be purged from the Flux instance with ``flux job purge``.
-The following options may be used to add selection criteria:
+Specific job ids may be specified for purging.  If no job ids are
+specified, the following options may be used for selection criteria:
 
 **--age-limit=FSD**
    Purge inactive jobs older than the specified Flux Standard Duration.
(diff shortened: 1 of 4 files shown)


grondo:
Weren't list and list-id deprecated though?
```
$ flux job list
This is not the command you are looking for. Try flux-jobs(1).
```
And flux-jobs optionally takes one or more jobids.

chu11:
> Weren't list and list-id deprecated though?

fair enough :-)

chu11:
re-pushed, re-working the PR so that its a single `flux job purge` command, which can optionally take job ids.

garlick:
Just tried this out and it seems to work well!

One thing I did notice when I tried to purge a running job was the error message was not particularly helpful:
```
$ flux jobs -A
    ƒ3TMJ3PM garlick  sleep       R      1      1   1.849s system76-laptop
$ flux job purge -f ƒ3TMJ3PM
flux-job: job-manager.purge-id: No such file or directory
```
We should return human-readable errors where possible.   Maybe for the ENOENT case we could have the RPC handler check the `active_jobs` hash and if found, send back "can't purge job in RUN state" (or whatever non-INACTIVE state it's in?)

Also I suggest not printing `job-manager.purge-id` since that's not helpful.

chu11:
re-pushed with some error message fixups given @garlick comments above and added/adjusted some tests.  Also fixed up the manpage, I forget to remove stuff related to what was formerly "purge-id".

Went with the following error message outputs:

```
>flux job purge f8Mo4FLb
flux-job: purge: cannot purge active job

>flux job purge f8Mo4FLf
flux-job: purge: id not found
```

Possibly controversial, I went with returning EPERM in the former case.

```
    if (!job) {
        if (!(job = zhashx_lookup (purge->ctx->active_jobs, &id))) {
            (*errmsg) = "id not found";
            errno = ENOENT;
        }
        else {
            (*errmsg) = "cannot purge active job";
            errno = EPERM;
        }
        return NULL;
    }
```

i.e. you cannot purge a job that is action, thus EPERM? ehhh

garlick (approved):
Thanks!

On your comment about the errno choice:  when responding with a textual error response, I think elsewhere we've just used EINVAL as a catch-all when there's not an obvious choice and a textual error is being returned also.

chu11:
> On your comment about the errno choice: when responding with a textual error response, I think elsewhere we've just used EINVAL as a catch-all when there's not an obvious choice and a textual error is being returned also.

Thanks.  I'll tweak that to EINVAL.  Re-push and set MWP.

PR #5031, opened 2023-03-28
job-list: minor code consistency cleanup

chu11:
technically is `job-list` and things related to `job-list`

PR #5048, opened 2023-03-30
job-list: make job stats consistent to job results

chu11:
Problem: The way that job-list stats counts failures is that all non-successful jobs are failures.  This can be a bit confusing b/c job results return jobs as "failed", "canceled", or "timeout".  So the total job failures via "job results" is different than the job failures via job-list stats.

Problem: Make the job-list stat counts consistent to job results.  Update tests and users of job-list stats in flux-top, flux-jobs, and t2260-job-list.t.

This PR is built on top of #5031

### DIFF
--- a/doc/man1/flux-jobs.rst
+++ b/doc/man1/flux-jobs.rst
@@ -104,6 +104,9 @@ OPTIONS
    will display a summary of statistics along with the top 25
    running jobs, updated every 2 seconds.
 
+   Note that all job failures, including canceled and timeout jobs,
+   are collectively counted as "failed" in ``--stats``.
+
 **--stats-only**
    Output a summary of job statistics and exit.  By default shows
    global statistics.  If ``--queue`` is specified, shows statistics
@@ -115,6 +118,9 @@ OPTIONS
 
    All options other than ``--queue`` are ignored when
    ``--stats-only`` is used.
+
+   Note that all job failures, including canceled and timeout jobs,
+   are collectively counted as "failed" in ``--stats-only``.
 
 **-R, --recursive**
    List jobs recursively. Each child job which is also an instance of
--- a/doc/man1/flux-jobs.rst
+++ b/doc/man1/flux-jobs.rst
@@ -162,7 +162,7 @@ states also exist: "pending", an alias for DEPEND,SCHED; "running", an
 alias for RUN,CLEANUP; "active", an alias for "pending,running".
 
 After a job has finished and is in the INACTIVE state, it can be
-marked with one of three possible results: COMPLETED, FAILED,
+marked with one of the possible results: COMPLETED, FAILED,
 CANCELED, TIMEOUT. Under the *result_abbrev* field name, these are
 abbreviated as CD, F, CA, and TO respectively.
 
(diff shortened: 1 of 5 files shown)


garlick:
A downside to this proposal is that `flux top` and `flux jobs --stats-only` "failed" counts will differ, and the latter won't represent TIMEOUT or CANCELED jobs in any of the summary statistics.  That seems a bit confusing too?

Just wondering if maybe it would be better to just call out in the `flux-jobs(1)` man page that the "failed" summary statistic includes  jobs that ended with FAILED, TIMEOUT, or CANCELED job results?

chu11:
I think that I updated the code for `flux top` and `flux jobs --stats-only` to be consistent to each other.  Did you mean `flux job stats`?  The latter is the raw stats coming straight from the job-list stats RPC, but `flux jobs --stats-only` collates it into output it wants to display.

IMO I think I'd like for the RPC of job-list stats and job-list "listing" to be consistent to each other.  How tools choose to display that data is sort of their own choosing?  An alternative might be for `flux-jobs --stats-only` to display timeout and canceled jobs separately, but I don't think we want to change output format at this point?

Your note on clarification in the manpage is worthwhile and we can add that too.  As well as add a fix for #5111 while we're at it.

src/cmd/top/summary_pane.c
```diff
@@ -152,7 +152,7 @@ static void draw_stats (struct summary_pane *sum)
     }
 
     if (sum->show_details) {
-        int failed = sum->stats.failed;
+        int failed = sum->stats.failed + sum->stats.timeout + sum->stats.canceled;
```
grondo, line 155: As in the JobStats class, maybe a comment stating here that `flux top` reports the total count of unsuccessful jobs in the `failed` stats? A minor suggestion since I guess it is pretty obvious.

chu11:
re-pushed, adding a `flux-jobs(1)` clarification per comment above, and stuck a fix for #5111 on top as well

doc/man1/flux-jobs.rst
```diff
@@ -116,6 +119,9 @@ OPTIONS
    All options other than ``--queue`` are ignored when
    ``--stats-only`` is used.
 
+   Note that all job failures, including canceled and timeout jobs
+   are collectively listed as "failed" in ``--stats-only``.
```
grondo, line 123: Maybe s/listed/counted/?

grondo (approved):
I was on the fence on this one.

On one hand, it makes very good sense to match only the jobs reported as FAILED in the `failed` count in these stats. On the other hand, all the use cases seem to combine canceled+timeout+failed in the reported `failed` count, which means as far as stats go, combining all unsuccessful jobs into a `failed` count is the expected result.

However, since this change affects only the underlying payload of the stats response, and since it easier for end users to calculate a total failed count by adding the 3 different failed statuses than to get the FAILED jobs by subtracting, I think I am ok with this approach.

Made some suggestions inline though. We should make sure @garlick agrees before merging.

garlick:
I just realized i was not testing the entire proposed change when I commented earlier.  :facepalm:   If the tools are still reporting counts as before then I have no issue!

chu11:
Re-pushed, fixing up per the comments above.  If I don't hear from anyone, I'll set MWP.  thanks.

chu11:
removing MWP temporarily, want #5112 to be merged first to fix mergify

Issue #5033, opened 2023-03-29
flux job wait: special exit code when no more waitable jobs


chu11:
While working on https://github.com/flux-framework/flux-docs/pull/220 I wanted to write a simple example loop to get jobids and "post process" jobs as they complete.  But AFAICT there's no way to differentiate between a job that exited with 1 and `flux job wait` exiting 1 when there are no more jobs to process.  This is the stupid thing I eventually came up with.

```
    jobid=$(flux job wait 2>&1)
    while ! echo ${jobid} | grep "no more waitable jobs"
    do
        echo "jobid ${jobid} finished, going to post process it"
        jobid=$(flux job wait 2>&1)
    done
    echo "no more waitable jobs, I'm done"
```

perhaps if `flux job wait` exited with a special exit code it'd be easier to know that that exit code means no more waitable jobs?  It could even be configurable by the user?

Or perhaps there's a simple way to do this that I'm just missing.

Edit:

I'm thinking something like the following would be easier and more obvious (didn't test, hopefully I wrote this right)

```
while jobid=`flux job wait`; [[ $? -ne 88 ]]
do
     echo "$jobid finished, do something with it"
done
```

grondo:
I ran across this same issue with `flux job wait` and ended up not using it because of this problem.

Using a special exit code is probably the right approach. I think `flux job wait` doesn't return with the exit status of the job, but just `1` for `!success` and `0` for "success", so this should work.

garlick:
Sounds reasonable.

Note:  [flux-job(1)](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man1/flux-job.html#wait) needs an update - the description misses the point of waitable jobs.

PR #5049, opened 2023-03-31
flux-job: add special exit code to flux job wait when not waitable

chu11:
As discussed in #5033, it is hard to differentiate between a job not being waitable / no more jobs are waitable vs a waitable job had an error.  Because in both circumstances `flux job wait` exits with 1.

This PR has `flux job wait` exit with 100 when either a jobid is non-waitable (invalid, not waitable, etc.) or there are no more waitable jobs.  So we can more easily write a script like this:

```
while jobid=`flux job wait 2> /dev/null`; ec=$?; test $ec -ne 100
do
    echo "$jobid finished"
    if [ $ec -eq 0 ]
    then
        echo "job successful"
        # do some post processing on successful jobs
    else
        echo "job failed"
        # do some post processing on failed jobs
    fi
done
echo "no more jobs to wait on"
```

I picked the exit code 100 somewhat arbitrarily.  It's common to use exit code of 2, 3, 4, ... for other purposes.  I just wanted to make it clear this wasn't an exit code from the job.  Maybe 100 is overkill.  I could make it configurable??

### DIFF
--- a/doc/man1/flux-job.rst
+++ b/doc/man1/flux-job.rst
@@ -119,14 +119,15 @@ results are consumed.
 When run with a jobid argument, ``flux job wait`` blocks until the specified
 job completes.  If the job was successful, it silently exits with a code of
 zero.  If the job has failed, an error is printed on stderr, and it exits with
-a code of one.  It is an error if the job was not submitted with the
-``waitable`` flag.
+a code of one.  If the jobid is invalid or the job is not waitable, ``flux job wait``
+exits with a code of two.  This special exit code of two is used to differentiate
+between a failed job and not being able to wait on the job.
 
 When run without arguments, ``flux job wait`` blocks until the next waitable
 job completes and behaves as above except that the jobid is printed to stdout.
-When there are no more waitable jobs, it exits with a code of one.  Note that
-a ``while flux job wait...`` loop terminates on the first unsuccessful job
-or when there are no more jobs.
+When there are no more waitable jobs, it exits with a code of two.  The exit code
+of two can be used to determine when no more jobs are waitable when using
+``flux job wait`` in a loop.
 
 ``flux job wait --all`` loops through waitable jobs as they complete, printing
 their jobids.  If all jobs are successful, it exits with a code of zero.  If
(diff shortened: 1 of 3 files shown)


chu11:
re-pushed, doing exit code 2 instead of 100.

doc/man1/flux-job.rst
```diff
@@ -119,14 +119,15 @@ results are consumed.
 When run with a jobid argument, ``flux job wait`` blocks until the specified
 job completes.  If the job was successful, it silently exits with a code of
 zero.  If the job has failed, an error is printed on stderr, and it exits with
-a code of one.  It is an error if the job was not submitted with the
-``waitable`` flag.
+a code of one.  If the jobid is invalid or the job is not waitable, ``flux job wait``
+exits with two.  This special exit code of two is used to differentiate between
```
garlick, line 123: To be consistent, should be "a code of two".   I agree that's starting to get a little long winded but consistentcy...

doc/man1/flux-job.rst
```diff
@@ -119,14 +119,15 @@ results are consumed.
...
 job completes and behaves as above except that the jobid is printed to stdout.
-When there are no more waitable jobs, it exits with a code of one.  Note that
-a ``while flux job wait...`` loop terminates on the first unsuccessful job
-or when there are no more jobs.
+When there are no more waitable jobs, it exits with a code of two.  The exit code
+of two can be used to determine when no more jobs are waitable when using
+``flux job wait`` in a loop.
 
 ``flux job wait --all`` loops through waitable jobs as they complete, printing
 their jobids.  If all jobs are successful, it exits with a code of zero.  If
```
garlick, line 133: It might be worth saying "loops through all the waitable jobs" here (adding "all the"), just to be clear.

garlick (approved):
Man page could maybe use some slight changes but otherwise looks good to me,  thanks!

doc/man1/flux-job.rst
chu11: i added an extra commit on top to just add this little clarification

chu11:
@garlick thanks for the review.  added an extra commit on top of your clarification addition and fixed up that minor bit of text consistency.  Setting MWP

chu11:
@Mergifyio rebase

chu11:
@Mergifyio refresh

chu11:
@mergifyio rebase

chu11:
@mergifyio update

chu11:
@Mergifyio requeue

chu11:
@Mergifyio rebase

Issue #5040, opened 2023-03-29
flux start hangs on cray 


garlick:
Problem: As noted in flux-framework/flux-coral2#44, when flux finds the cray libpmi.so, it hangs.

The problem seems to be that this library hangs when a key is requested that is not present in its KVS (probably waiting for it to appear from another node).  This is incompatible with the way flux is using PMI to fetch job information like `flux.instance-level`,  `flux.taskmap`, and `PMI_process_mapping` (although that last one may not be an issue).

Perhaps to avoid this we could look for hints that the enclosing instance is not flux, and then skip fetching the flux specific PMI KVS keys?

garlick:
I tried skipping the kvs fetches of keys that we know won't be there but found that our bizcard values (containing public keys) were being truncated by cray's PMI, and there was no way to obtain the PMI_process_mapping.

Instead I made a PMI2 "upmi" (universal flux client) plugin.  That seems to work the way it ought.  However, we probably still need a heuristic to avoid fetching the flux-specific keys because apparently the library is a bit chatty on stderr when keys aren't found:
```
rzvernal11 /g/g0/garlick/proj/flux-core > echo $LD_LIBRARY_PATH
/opt/cray/pe/lib64:/opt/cray/lib64:/opt/cray/pe/papi/7.0.0.1/lib64:/opt/cray/libfabric/2.0/lib64

rzvernal11 /g/g0/garlick/proj/flux-core > srun -N2 src/cmd/flux start flux getattr broker.mapping
Thu Mar 30 13:59:44 2023: [PE_1]:_pmi2_kvs_get:key [flux.instance-level] was not found.
Thu Mar 30 13:59:44 2023: [PE_1]:PMI2_KVS_Get:_pmi2_kvs_get failed
Thu Mar 30 13:59:44 2023: [PE_1]:_pmi2_kvs_get:key [flux.taskmap] was not found.
Thu Mar 30 13:59:44 2023: [PE_1]:PMI2_KVS_Get:_pmi2_kvs_get failed
Thu Mar 30 13:59:44 2023: [PE_0]:_pmi2_kvs_get:key [flux.instance-level] was not found.
Thu Mar 30 13:59:44 2023: [PE_0]:PMI2_KVS_Get:_pmi2_kvs_get failed
Thu Mar 30 13:59:44 2023: [PE_0]:_pmi2_kvs_get:key [flux.taskmap] was not found.
[...]

PR #5051, opened 2023-03-31
libpmi: enable flux to bootstrap with cray libpmi2.so

garlick:
This works through the problems identified in #5040.  I've been able to test this on rzvernal, where the cray PMI libraries were added to LD_LIBRARY_PATH and seems ok.

```
$ srun -N2 -n4 src/cmd/flux start flux getattr broker.mapping
[[0,2,2,1]]
$ src/cmd/flux start flux getattr broker.mapping
(snip big warning from libhwloc)
[[0,1,1,1]]
```

### DIFF
--- a/src/broker/boot_pmi.c
+++ b/src/broker/boot_pmi.c
@@ -88,7 +88,6 @@ static int set_broker_mapping_attr (struct upmi *upmi,
     else {
         /* First attempt to get flux.taskmap, falling back to
          * PMI_process_mapping if this key is not available.
-         * This should be replaced when #4800 is fixed.
          */
         char *s;
         if (upmi_get (upmi, "flux.taskmap", -1, &s, NULL) == 0
--- a/src/broker/boot_pmi.c
+++ b/src/broker/boot_pmi.c
@@ -84,7 +84,7 @@ static int set_broker_mapping_attr (struct upmi *upmi,
     int rc;
 
     if (size == 1)
-        val = strdup ("{\"version\":1,\"map\":[[0,1,1,1]]}");
+        val = strdup ("[[0,1,1,1]]");
     else {
         /* First attempt to get flux.taskmap, falling back to
          * PMI_process_mapping if this key is not available.
--- a/src/common/libpmi/upmi_libpmi2.c
+++ b/src/common/libpmi/upmi_libpmi2.c
@@ -106,6 +106,7 @@ static void plugin_ctx_destroy (struct plugin_ctx *ctx)
 
 static struct plugin_ctx *plugin_ctx_create (const char *path,
                                              bool noflux,
+                                             bool craycray,
                                              flux_error_t *error)
 {
     struct plugin_ctx *ctx;
@@ -134,7 +135,7 @@ static struct plugin_ctx *plugin_ctx_create (const char *path,
         goto error;
     }
 
-    if (dlsym (ctx->dso, "PMI_CRAY_Get_app_size") != NULL)
+    if (dlsym (ctx->dso, "PMI_CRAY_Get_app_size") != NULL || craycray)
         ctx->flags |= LIBPMI2_IS_CRAY_CRAY;
 
     return ctx;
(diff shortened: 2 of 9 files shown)


garlick:
Probably should hold off here - I'm seeing the occasional segfault when launching under slurm and want to run that down.

garlick:
This is based on top of #5053 now, so that one should go in first.

I've been unable to reproduce the segfault I saw earlier so I'll say (tentatively) that this is tentatively ready for a review.

src/common/libpmi/upmi_libpmi2.c
```diff
@@ -128,6 +131,10 @@ static struct plugin_ctx *plugin_ctx_create (const char *path,
         errprintf (error, "%s:  missing required PMI2_* symbols", path);
         goto error;
     }
+
+    if (dlsym (ctx->dso, "PMI_CRAY_Get_app_size") != NULL)
+        ctx->flags |= LIBPMI2_IS_CRAY_CRAY;
```
grondo, line 139: :joy:

garlick:
Rebased on master after #5053 was merged.

src/broker/boot_pmi.c
```diff
@@ -88,7 +88,6 @@ static int set_broker_mapping_attr (struct upmi *upmi,
     else {
         /* First attempt to get flux.taskmap, falling back to
          * PMI_process_mapping if this key is not available.
-         * This should be replaced when #4800 is fixed.
```
grondo, line 91: Just to clarify, the broker still does need to fall back to PMI_process_mapping here?

t/t3002-pmi.t
```diff
@@ -206,7 +206,7 @@ test_expect_success 'flux-pmi --libpmi-noflux fails w/ flux libpmi.so' '
 	    flux pmi --method=libpmi:$(cat libpmi) --libpmi-noflux barrier"
 '
 test_expect_success 'flux broker refuses the Flux libpmi.so and goes single' '
-	FLUX_PMI_DEBUG=1 \
+	FLUX_PMI_DEBUG=1 FLUX_PMI_CLIENT_METHODS="libpmi single" \
```
grondo, line 209: I'm unclear on why the test should _not_ ensure that both libpmi2 and libpmi plugins refuse to boot from Flux's own library. Maybe the commit message just needs to explain the why of the problem statement. (Apologies if it should be obvious)

grondo (approved):
Nice!

Just a couple questions inline. I didn't find anything major so approving.

src/broker/boot_pmi.c
garlick: Yes, the current code which does
```c
        char *s;
        if (upmi_get (upmi, "flux.taskmap", -1, &s, NULL) == 0
            || upmi_get (upmi, "PMI_process_mapping", -1, &s, NULL) == 0) {
            val = pmi_mapping_to_taskmap (s);
            free (s);
        }
```
is fine I think!  In any case, the issue was closed so the reference probably should be dropped.

t/t3002-pmi.t
garlick: It could check both, but it says its checking libpmi and I add another check for libpmi2 later on, so it was more just an effort to organize tests by plugin and not have the behavior of one plugin affect a test that claims to be checking the other. I'll update the commit message to try to be more clear.

garlick:
Repushed with that commit message changed to
```
    testsuite: narrow libpmi noflux test
    
    Problem: a test that claims to check the libpmi plugin's rejection
    of Flux's libpmi.so is now checking libpmi2 also
    
    Set FLUX_PMI_CLIENT_METHODS to skip over libpmi2 so a failure will
    implicate the correct plugin.
```
and I'll go ahead and set MWP.

t/t3002-pmi.t
grondo: Ah! That makes sense (it should have been obvious, sorry!)

PR #5052, opened 2023-03-31
flux-top: support ability to flip through queues via left/right arrow keys

chu11:
It is inconvenient to have to restart flux-top to filter jobs by a different queue.

Support using the left/right arrow keys or h/l to "rotate" through all of the possible queues, including "all" queues.

Only minor thing is it's a little annoying that you don't know what queue (or "all" queues) you are currently rotating through if no jobs are in that queue.  So I added the queue name along with the fancy f in the upper left.  Could be somewhere else?  Originally I was going to add something right above the "QUEUE" header, until I realized the instance depth is already output there when you recurse down in `flux-top`.  We could also remove this, it's not super important.

This PR is built on top of #5032

### DIFF
--- a/src/cmd/top/joblist_pane.c
+++ b/src/cmd/top/joblist_pane.c
@@ -99,6 +99,21 @@ void joblist_pane_draw (struct joblist_pane *joblist)
     wattroff (joblist->win, A_REVERSE);
     if (joblist->jobs == NULL)
         return;
+
+    if (json_array_size (joblist->jobs) == 0
+        && queues_configured (joblist->top->queues)) {
+        const char *filter_queue = NULL;
+        /* can return NULL filter_queue for "all" queues */
+        queues_get_queue_name (joblist->top->queues, &filter_queue);
+        if (filter_queue)
+            mvwprintw (joblist->win,
+                       5,
+                       25,
+                       "No jobs to display in queue %s",
+                       filter_queue);
+        return;
+    }
+
     json_array_foreach (joblist->jobs, index, job) {
         char *uri = NULL;
         char idstr[16];
--- a/src/cmd/top/joblist_pane.c
+++ b/src/cmd/top/joblist_pane.c
@@ -133,6 +133,7 @@ void joblist_pane_draw (struct joblist_pane *joblist)
         int ntasks;
         int nnodes;
         double t_run;
+        const char *filter_queue;
 
         if (json_unpack (job,
                          "{s:I s:i s:i s:s s?s s:i s:i s:f s?{s?{s?s}}}",
@@ -149,7 +150,9 @@ void joblist_pane_draw (struct joblist_pane *joblist)
                              "uri", &uri) < 0)
             fatal (0, "error decoding a job record from job-list RPC");
 
-        if (joblist->top->queue && !streq (joblist->top->queue, queue))
+        /* can return NULL filter_queue for "all" queues */
+        queues_get_queue_name (joblist->top->queues, &filter_queue);
+        if (filter_queue && !streq (filter_queue, queue))
             continue;
 
         if (flux_job_id_encode (id, "f58", idstr, sizeof (idstr)) < 0)
@@ -363,12 +366,16 @@ void joblist_pane_set_current (struct joblist_pane *joblist, bool next)
     json_t *job;
     flux_jobid_t id = FLUX_JOBID_ANY;
     int njobs;
+    const char *filter_queue;
 
     if (joblist->jobs == NULL)
         return;
 
     if (joblist->current != FLUX_JOBID_ANY)
         index = lookup_jobid_index (joblist->jobs, joblist->current);
+
+    /* can return NULL filter_queue for "all" queues */
+    queues_get_queue_name (joblist->top->queues, &filter_queue);
 
     /* find next valid index */
     njobs = json_array_size (joblist->jobs);
@@ -382,11 +389,11 @@ void joblist_pane_set_current (struct joblist_pane *joblist, bool next)
         if (!(job = json_array_get (joblist->jobs, next_index)))
             return;
 
-        if (joblist->top->queue) {
+        if (filter_queue) {
             const char *queue = "";
             if (json_unpack (job, "{s?s}", "queue", &queue) < 0)
                 fatal (0, "error decoding a job record from job-list RPC");
-            if (!streq (joblist->top->queue, queue)) {
+            if (!streq (filter_queue, queue)) {
                 index = next_index;
                 continue;
             }
--- a/src/cmd/top/joblist_pane.c
+++ b/src/cmd/top/joblist_pane.c
@@ -65,39 +65,17 @@ static json_t *get_current_job (struct joblist_pane *joblist)
 }
 
 
-/*  Return true if any job in joblist->jobs has a queue defined. If
- *  joblist->jobs is NULL or empty, then return the current show_queue
- *  value.
- */
-static bool show_queue (struct joblist_pane *joblist)
-{
-    size_t index;
-    json_t *job;
-
-    if (!joblist->jobs || json_array_size (joblist->jobs) == 0)
-        return joblist->show_queue;
-
-    json_array_foreach (joblist->jobs, index, job) {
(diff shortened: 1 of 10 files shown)


t/t2801-top-cmd.t
```diff
@@ -10,8 +10,6 @@ runpty="${SHARNESS_TEST_SRCDIR}/scripts/runpty.py"
 waitfile="${SHARNESS_TEST_SRCDIR}/scripts/waitfile.lua"
 testssh="${SHARNESS_TEST_SRCDIR}/scripts/tssh"
 
-export FLUX_URI_RESOLVE_LOCAL=t
```
grondo, line 13: This is fine, but might actually be more tractable long term to just fix the one test test that unsets the variable to run in a sub-shell, since every case of resolving a URI in the test should use `FLUX_URI_RESOLVE_LOCAL=t` (and it might not always be clear that a test will end up resolving a uri...)

grondo:
Nice! This is cool.

>  So I added the queue name along with the fancy f in the upper left. 

Was it also meant to be in yellow? It seems to give it greater importance than it should. What about just to the left of the version number, and specify `queue: <name>` like the other data in that row. It could also go in the blank like between resource summary and instance data line.

I also noticed there's a delay between switching the queue and getting an update in the resource summary.

t/t2801-top-cmd.t
chu11: good idea.  This change was from PR #5032 so I'll push a change in that PR.

grondo: Oh, oops sorry!

chu11:
> Was it also meant to be in yellow?

Oops, nope :P

> What about just to the left of the version number, and specify queue: <name> like the other data in that row. It could also go in the blank like between resource summary and instance data line.

Lemme see if I can sneak it in to the left of the version number, don't have a ton of space though.  Perhaps if queues are configured, there could just be an extra line below it.

> I also noticed there's a delay between switching the queue and getting an update in the resource summary.

Hmm yeah I notice it now too.  I draw the joblist pane and summary pane at the same time.  I'm wondering if the summary pane has to go get new queue specific data, then the update comes at the next heartbeat.  So gotta do something to make it update faster.

chu11:
> Hmm yeah I notice it now too. I draw the joblist pane and summary pane at the same time. I'm wondering if the summary pane has to go get new queue specific data, then the summary pane re-draw comes at the next heartbeat. So gotta do something to make it re-draw earlier/immediately.

Ahhh my guess was almost correct.  the summary pane draw function doesn't do a query, so the query comes at the next heartbeat, which updates the summary stats.  Just gotta force the query before the draw.

```
diff --git a/src/cmd/top/keys.c b/src/cmd/top/keys.c
index 49e7fa0..5125d29 100644
--- a/src/cmd/top/keys.c
+++ b/src/cmd/top/keys.c
@@ -45,12 +45,14 @@ static void keys_cb (flux_reactor_t *r,
         case 'h':
         case KEY_LEFT:
             queues_prev (keys->top->queues);
+            summary_pane_query (keys->top->summary_pane);
             summary_pane_draw (keys->top->summary_pane);
             joblist_pane_draw (keys->top->joblist_pane);
             break;
         case 'l':
         case KEY_RIGHT:
             queues_next (keys->top->queues);
+            summary_pane_query (keys->top->summary_pane);
             summary_pane_draw (keys->top->summary_pane);
             joblist_pane_draw (keys->top->joblist_pane);
             break;
```

chu11:
re-pushed, fixing the summary pane update per comment above.

after some trial and error, I liked adding the queue name on a new line in the summary pane

image

So we do lose a line ... and that's an empty line when there are no queues configured (or "all" queues are displayed), which doesn't look bad.  And we now have an extra line if we need to add anything in the future.

As an aside, I went back and forth between showing no queue name when all queues are listed vs listing something like "queue: all" ... any strong opinions on which way to go?  Right now it elects to show nothing when all queues are listed.

grondo:
Just an opinion, but not a fan of the extra line. Why can't `queue`  fit between `uptime` and `version` again?

My opinion is no need to put  `queue: all`. There may be a queue called "all"

chu11:
> Just an opinion, but not a fan of the extra line. Why can't queue fit between uptime and version again?

It just seemed tight to me.  In the above we only have 18 spaces in between, so that's only a few spaces left and right of something like "queue: batch".

but now that I take a closer look "depth: %d" isn't very big and there's a lot of space between that and the "uptime" string.  What if I moved the uptime string a little to the left?

image

That doesn't look too bad, although "uptime" is a little left of center now if there's no queues.

chu11:
I re-pushed, with the version that put the queue name on the same line as the uptime and version.

grondo:
Had another idea, the `queue: foo` entry is redundant whenever there are jobs to display. So how about just putting `No jobs in queue foo` in the job list window when there are no jobs in the selected queue. Then we don't have to clutter the bottom summary line with the queue that will be almost always unused or redundant.

chu11:
> Had another idea, the queue: foo entry is redundant whenever there are jobs to display. So how about just putting No jobs in queue foo in the job list window when there are no jobs in the selected queue. Then we don't have to clutter the bottom summary line with the queue that will be almost always unused or redundant.

Oh that's a good idea.  I'll try to make it "pretty" by centering in in the joblist pane.

chu11:
re-pushed

- I merged the commits from #5065 in here because it would have made things easier.
- per @grondo suggestion above, simply display "no jobs to display" in the joblist pane if there are none to display.  I do this even for any case, anonymous queue, per queue, etc.  I really like this, as it simply informs the user in an obvious way that nothing is broken, there's just nothing to display.

image

grondo:
Nice, I like that better :+1:. However, I think the `no jobs to display` when not viewing a specific queue is an unnecessary change, though that's only my personal opinion. If you don't feel strongly about it, is there any way to suppress that message for the default case?

grondo:
>  I just liked it, b/c when you run flux top and nothing shows up, one might wonder why nothing shows up.

My point was just that this is how it has always been and since there's been no complaints, the change in behavior didn't seem necessary.

chu11:
> My point was just that this is how it has always been and since there's been no complaints, the change in behavior didn't seem necessary.

That's a good point, that I am altering the original behavior.

chu11:
re-pushed, per @grondo suggestion above, under anonymous queue or "all queues", do not display "no jobs to display"

src/cmd/top/joblist_pane.c
```diff
@@ -99,6 +99,21 @@ void joblist_pane_draw (struct joblist_pane *joblist)
...
+
+    if (json_array_size (joblist->jobs) == 0
+        && queues_configured (joblist->top->queues)) {
+        const char *filter_queue = NULL;
+        /* can return NULL filter_queue for "all" queues */
+        queues_get_queue_name (joblist->top->queues, &filter_queue);
+        if (filter_queue)
+            mvwprintw (joblist->win,
+                       10,
+                       25,
```
grondo, line 111: Maybe just leave the note at the top? What if the terminal has < 25 rows?

grondo (approved):
LGTM! A couple minor comments.

src/cmd/top/joblist_pane.c
chu11: yeah, i guess I wanted to center it to be pretty, but perhaps on line 2 or 3 would be preferable.

chu11: oh wait a second, 25 is the width ... it starts on line 10 ... do you feel that's too far down?

grondo: Oh, oops. I felt I had a window that wasn't too small, and the text was near the bottom. But I forget how this works, is that line 10 of the pane? I'm fine with keeping it as is, it was just a suggestion.

chu11: Yeah, line 10 of the pane.  But we can bump it up a bit, maybe line 5 or something.

chu11:
@grondo thanks.  Cleaned up that commit message and I moved up the message from line 10 to line 5 of the window.  It still looks pretty good on a quick test.

PR #5053, opened 2023-03-31
libpmi: improve tracing of dlopened PMI libraries

garlick:
Problem: when debugging flux launch failure with FLUX_PMI_DEBUG=1, trace output does not include the final dlopened library, if any.

This adds a way for the plugin to return a generic note on successful initialization, which is included in the trace output.  The libpmi plugin sets its note to "using .../libpmi.so".  Here's an example when launching under slurm on a cray system (using libpmi2 which isn't merged yet but you get the idea):
```
rzvernal11 /g/g0/garlick/proj/flux-core > FLUX_PMI_DEBUG=1 srun -N2 -n2 src/cmd/flux start /bin/true
boot_pmi: flux-pmi-client: trying 'simple'
boot_pmi: simple: PMI_FD not found in environ
boot_pmi: flux-pmi-client: trying 'libpmi2'
boot_pmi: flux-pmi-client: trying 'simple'
boot_pmi: simple: PMI_FD not found in environ
boot_pmi: flux-pmi-client: trying 'libpmi2'
boot_pmi: libpmi2: using /opt/cray/pe/lib64/libpmi2.so (cray quirks enabled)
boot_pmi: libpmi2: using /opt/cray/pe/lib64/libpmi2.so (cray quirks enabled)
boot_pmi: libpmi2: initialize: rank=0 size=2 name=kvs_74317.0: success
boot_pmi: libpmi2: initialize: rank=1 size=2 name=kvs_74317.0: success
[snip]
```

The "refactor flags setting" commit was brought over from #5051 because it makes a necessary structural change for this tracing to work.

### DIFF
--- a/src/common/libpmi/upmi.c
+++ b/src/common/libpmi/upmi.c
@@ -13,6 +13,7 @@
 #if HAVE_CONFIG_H
 #include "config.h"
 #endif
+#include <jansson.h>
 #include <flux/core.h>
 #include <assert.h>
 #include <argz.h>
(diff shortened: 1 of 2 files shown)


grondo (approved):
LGTM!

Issue #5057, opened 2023-04-01
lt-flux-broker: libpmi2: initialize: init: pmi error 14


chu11:
I'm seeing a mountain of sharness errors on catalyst now.

```
lt-flux-broker: libpmi2: initialize: init: pmi error 14
lt-flux-broker: bootstrap failed
not ok 4 - run a testexec job in persistent instance (exit run)
```

I'm guessing something not quite right after merging #5051?

Edit: confirmed it started after #5051 was merged, pre-that merge tests all passed

Edit2: apologies, I meant after PR 5051 merged, not necessarily the single commit for 5040

Edit3: ok, i can now confirm commit 17c665d800a4e86cf6fe236467563c07405c057a which solves #5040 is causing the problem

chu11:
error is coming from `op_initialize()` in the new `upmi_libpmi2.c`

```
    result = ctx->job_getid (jobid, sizeof (jobid));                                                                                        
    if (result != PMI2_SUCCESS)                                                                                                             
        return upmi_seterror (p, args, "init: %s", pmi_strerror (result));     
```

chu11:
This temporarily fixes things

```
-static const char *default_methods = "simple libpmi2 libpmi single";
+static const char *default_methods = "simple libpmi libpmi2 single";
```

I don't know pmi underneath the covers, but i guess slurm's libpmi2 is somehow picked up and for some reason doesn't make flux happy.

But the fix for #5040 has a situation where libpmi2 is better than libpmi.  Not sure hwo to deal w/ this circular issue.

garlick:
Uh oh.  That test you cited as failing just runs `flux start` so I would guess that whatever libpmi2 is being found is unable to start a singleton.  You may well be correct that slurm's libpmi.so is able to, and slurm's libpmi2.so is not.

Changing the libpmi/libpmi2 order would negate the cray fix that just went in so more thought required...

We could make sure we always use `flux-start --test-size=1` in sharness but we still want singletons to work when users start them on a slurm system.

garlick:
Reproduced:
```
[garlick@catalyst159:flux-core]$ FLUX_PMI_DEBUG=1 src/cmd/flux start -o,-v
boot_pmi: flux-pmi-client: trying 'simple'
boot_pmi: simple: PMI_FD not found in environ
boot_pmi: flux-pmi-client: trying 'libpmi2'
boot_pmi: libpmi2: using /lib64/libpmi2.so
[...]

PR #5058, opened 2023-04-02
PMI: ensure fallthrough when PMI library fails to initialize

garlick:
Problem: a flux standalone singleton no longer works on slurm systems, as reported in #5057

The libpmi2.so from slurm (which is installed in a public libdir) does not "emulate" singleton when called outside of a slurm job, unlike the libpmi.so from slurm which does emulate singleton.

Now that we're searching for libpmi2 ahead of libpmi because of the cray situation, we're hitting this.

Change the libpmi2 (and libpmi) plugins so that they ensure the library is going to work before selecting it.  This lets the pmi client try the next method.

WIP pending some manual testing.

### DIFF
--- a/src/common/libpmi/upmi_libpmi.c
+++ b/src/common/libpmi/upmi_libpmi.c
@@ -36,6 +36,8 @@ struct plugin_ctx {
     int (*kvs_put) (const char *kvsname, const char *key, const char *value);
     int (*kvs_commit) (const char *kvsname);
     int (*kvs_get) (const char *kvsname, const char *key, char *value, int len);
+    int rank;
+    int size;
     char kvsname[1024];
 };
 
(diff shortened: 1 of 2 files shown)


chu11:
confirmed on catalyst that the gajillions of sharness errors no longer happen with this.  thanks!

garlick:
I need to give it a spin on the cray systems to make sure I didn't break anything over there, then it'll be mergeable.  If you have a moment to review the patches @chu11, that'd be helpful.  I'm happy to explain anything that's not obvious.  It's PMI so maybe nothing is obvious :-)

chu11 (approved):
took a look and everything LGTM!

garlick:
Seems good on rzvernal (under slurm and standalone singleton) so I'll set MWP.  Thanks!

PR #5062, opened 2023-04-03
add libabigail checks

### LINKED ISSUE #4396: add libabigail ABI checks to CI
It might be good to add ABI checks to CI to warn if and when PRs break ABI.

See https://github.com/buildsi/libabigail-action for an example of how this might be done.

Since we publish docker images of the latest merged changes, we might be able to do this without the parallel builds for comparison (just fetch the latest docker image and pull `libflux-core.so` from there).

vsoch:
Problem: ABI breaks can happen between versions
Solution: run libabigail against release vs. pr

We do this by grabbing the *.so out of the install folder in the latest `fluxrm/flux-sched:focal` and comparing against using the same container to build the pull request code. Since Flux is finicky about having a tag for the version, we make a faux one.

This will close #4396 

Background for the action: https://youtu.be/2Oet0hGOy7U?t=1998

Watch beyond that segment for my Yoda impression!

### DIFF
--- a/.github/workflows/libabigail.yaml
+++ b/.github/workflows/libabigail.yaml
@@ -1,0 +1,112 @@
+name: Run Libabigail
+on: [pull_request]
+
+jobs:
+  build-release:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Previous Release
+      uses: actions/checkout@v3
+    - name: Retrieve from Container
+      run: |
+       mkdir -p /tmp/libs
+       docker run --rm \
+          --workdir=$(pwd) \
+          --volume=/tmp/libs:/tmp/libs \
+          fluxrm/flux-sched:focal \
+          sh -c "sudo cp /usr/lib/*.so /tmp/libs"
+       sudo chown -R $USER /tmp/libs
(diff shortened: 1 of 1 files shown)


.github/workflows/libabigail.yaml
```diff
@@ -0,0 +1,112 @@
...
+    steps:
+    - name: Build Previous Release
+      uses: actions/checkout@v3
+    - name: Retrieve from Container
+      run: |
+       mkdir -p /tmp/libs
+       docker run --rm \
+          --workdir=$(pwd) \
+          --volume=/tmp/libs:/tmp/libs \
+          fluxrm/flux-sched:focal \
```
grondo, line 16: This is great!

I'm sure there is a good reason, but why use the flux-sched docker image instead of flux-core here? The flux-sched image only gets updated when changes to flux-sched are merged, so it could be out of date?

vsoch: I honestly didn't know there was a flux core one :laughing: I always use flux-sched! Even better, I think we should use the containers (TBA) built via GitHub actions and pushed to ghcr.io #4996

PR #5066, opened 2023-04-03
libpmi: improve error messages from upmi plugins

garlick:
Problem: when libpmi and libpmi2 upmi plugins fail, they do not include the full path the dso in errors.

This was a regression from flux-framework/flux-core#5058.

Restore paths to error messages.


chu11 (approved):
LGTM!

PR #5067, opened 2023-04-03
openmpi: don't force flux MCA plugins

garlick:
Problem: we may want to bootstrap pre-v5 versions of openmpi using the flux-pmix plugin, but the always-loaded openpmi plugin tries to force the flux MCA plugins to load.

Drop that part of the plugin.  It shouldn't really be necessary anyway, since the whole purpose of the ompi "schizo" layer is to determine which bootstrap method to use.  It should find Flux OK unless it's configured wrong.


chu11 (approved):
LGTM

Issue #5063, opened 2023-04-03
idea: normalize pmi shell options


garlick:
Problem: the "simple" PMI server shell plugin occupies the "pmi" shell option name space but there are other PMI implementations available now (cray-pals and pmix) that ignore this space.  For example `-o pmi=off` is ignored by flux-pmix.

Proposal: let the value of `-o pmi` select the implementation, like `-o pmi=pmix` or `-o pmi=cray-pals`, and move  implementation specific options down a level, e.g.  what is now `-o pmi.kvs=native` would become `-o pmi.simple.kvs=native`.

Finally, we may need a way to globally alter the pmi server implementation default (e.g. on cray systems), but still have flux launch flux with its preferred method.

garlick:
Oh right, this doesn't work (duh):
```console
$ flux run -opmi=simple -opmi.simple.kvs=native  true
flux-run: ERROR: 'str' object has no attribute 'setdefault'
```
Perhaps each implementation will need its own top level key for configuration.

Then we could free up `-o pmi=VALUE` to accept a comma separated list of implementations to load.  E.g. on a cray system we could set the default to `-o pmi=simple,cray-pals`.

PR #5069, opened 2023-04-03
shell: support -o pmi=LIST

garlick:
This implements the idea presented in #5063 of interpreting `flux run -o pmi=VALUE` values as a list of PMI implementations to load.   It preserves the meaning of `-o pmi=off` and the behavior of loading the simple server by default, but now if `-opmi` is specified and it doesn't include `simple`, the simple server is not loaded.

This PR also renames the little used simple pmi tunables to be under an implementation specific key.  `-o pmi.kvs=native` becomes `-o pmi-simple.kvs=native`.

This gives users more control over which PMIs are active among `simple`, `pmix`, and `cray-pals`.  It will require those plugins to be modified as well to disable themselves if they are not in the list.

### DIFF
--- a/doc/man1/flux-shell.rst
+++ b/doc/man1/flux-shell.rst
@@ -325,17 +325,22 @@ options are supported by the builtin plugins of ``flux-shell``:
   be set by the :man1:`flux-submit` and related commands --taskmap`` option.
 
 **pmi=off**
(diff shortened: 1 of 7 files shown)


chu11 (approved):
LGTM

PR #5070, opened 2023-04-04
sdbus: enable flux to communicate with a systemd user instance

garlick:
Edit: after some iteration, this PR now only contains the `sdbus` module and its tests.  The `sdexec` portion will be submitted separately.

This is a rework/reimagining of @chu11's libsdprocess work from 2021 with the goal of getting  `job-exec` to launch shells as systemd transient units for the following short term benefits:
- clean up of orphaned processes when flux comes back online
- systemd cgroups containment (especially memory) on job shells

and of course to prepare groundwork for the eventual recovery of running jobs.

Two broker modules are added that enable libsubprocess to work with systemd transient units.  The modules are only loaded in an instance configured with
```toml
[systemd]
enable = true
```

One module is a D-Bus gateway module that translates Flux RPCs into D-Bus RPCs and lets flux subscribe to D-Bus "events" (called signals). It operates reactively and can handle multiple outstanding RPCs that may complete out of order.

The other module is a libsubprocess compatible remote server that launches processes as systemd transient units via the above gateway.

Both modules load on all ranks, so the `job-exec` "bulkexec" method should be able to work like it does now, just using a different subprocess server name.  (But see last todo below)

For testing, `flux-exec(1)` gets an option to use the new subprocess service, and `flux-sdctl(1)` is a C program that behaves like `systemctl(1)`.  (Ew.  Not pretty.  See second todo below)

Todo:

[snip]

This work benefited greatly from the earlier work as @chu11 did a nice job of documenting the issues he ran into and having a working example to start with greatly reduced the inertia of starting this!

My goal in posting this is to get the design out there for any initial thoughts, and to let would-be reviewers have a sneak peek so they won't have to grok the whole thing at once when it's time to review in earnest.

### DIFF
--- a/src/modules/sdbus/message.c
+++ b/src/modules/sdbus/message.c
@@ -1,0 +1,486 @@
+/************************************************************\
+ * Copyright 2023 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* message.c - D-Bus message payload/JSON conversion helpers
+ */
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+#include <jansson.h>
+#include <errno.h>
+#include <systemd/sd-bus.h>
+#include <assert.h>
+
+#include "src/common/libutil/errno_safe.h"
+#include "src/common/libutil/errprintf.h"
+#include "ccan/str/str.h"
+#include "ccan/array_size/array_size.h"
+
+#include "objpath.h"
+#include "message.h"
+
+typedef union {
+    uint8_t u8;
+    int16_t i16;
+    uint16_t u16;
+    int32_t i32;
+    uint32_t u32;
+    int64_t i64;
+    uint64_t u64;
+    int i;
+    double f;
+} number_t;
+
+struct tab {
+    uint8_t type;
+    const char *desc;
+};
+
+static struct tab typetab[] = {
+    { SD_BUS_MESSAGE_METHOD_CALL, "method-call" },
+    { SD_BUS_MESSAGE_METHOD_RETURN, "method-return" },
+    { SD_BUS_MESSAGE_METHOD_ERROR, "method-error" },
+    { SD_BUS_MESSAGE_SIGNAL , "signal" },
+};
+
+const char *sdmsg_typestr (sd_bus_message *m)
+{
+    uint8_t type;
+
+    if (sd_bus_message_get_type (m, &type) >= 0) {
+        for (int i = 0; i < ARRAY_SIZE (typetab); i++)
+            if (typetab[i].type == type)
+                return typetab[i].desc;
+    }
+    return "unknown";
+}
+
+static int sdmsg_put_array (sd_bus_message *m, const char *fmt, json_t *o)
+{
+    int e;
+    size_t index;
+    json_t *entry;
+
+    if ((e = sd_bus_message_open_container (m, 'a', fmt)) < 0)
+        return e;
+    json_array_foreach (o, index, entry) {
+        if ((e = sdmsg_put (m, fmt, entry)) < 0)
+            return e;
+    }
+    if ((e = sd_bus_message_close_container (m)) < 0)
+        return e;
+
+    return e;
+}
+
+static int sdmsg_put_string (sd_bus_message *m, char type, json_t *o)
+{
+    int e;
+    switch (type) {
+        case 'g':
+        case 's': {
+            const char *val = json_string_value (o);
+            if (!val)
+                return -EPROTO;
+            if ((e = sd_bus_message_append_basic (m, type, val)) < 0)
+                return e;
+            break;
+        }
+        case 'o': {
+            char *val = objpath_encode (json_string_value (o));
+            if (!val)
+                return -errno;
+            e = sd_bus_message_append_basic (m, type, val);
+            free (val);
+            if (e < 0)
+                return e;
+            break;
+        }
+        default:
+            return -EPROTO;
+    }
+    return 0;
+}
+
+static int sdmsg_put_basic (sd_bus_message *m, char type, json_t *o)
+{
+    number_t n;
+
+    if (!m || !o)
+        return -EPROTO;
+    if (type == 's' || type == 'g' || type == 'o')
+        return sdmsg_put_string (m, type, o);
+    switch (type) {
+        case 'y':
+            n.u8 = json_integer_value (o);
+            break;
+        case 'b':
+            n.i = json_is_true (o) ? 1 : 0;
+            break;
+        case 'n':
+            n.i16 = json_integer_value (o);
+            break;
+        case 'q':
+            n.u16 = json_integer_value (o);
+            break;
+        case 'i':
+            n.i32 = json_integer_value (o);
+            break;
+        case 'u':
+            n.u32 = json_integer_value (o);
+            break;
+        case 'x':
+            n.i64 = json_integer_value (o);
+            break;
+        case 't':
+            n.u64 = json_integer_value (o);
+            break;
+        case 'h':
+            n.i = json_integer_value (o);
+            break;
+        case 'd':
+            n.f = json_real_value (o);
+            break;
+        default:
+            return -EPROTO;
+    }
+    return sd_bus_message_append_basic (m, type, &n);
+}
+
+static int sdmsg_put_variant (sd_bus_message *m, json_t *o)
+{
+    const char *type;
+    json_t *val;
+    int e;
+
+    if (json_unpack (o, "[so]", &type, &val) < 0)
+        return -EPROTO;
+    if ((e = sd_bus_message_open_container (m, 'v', type)) < 0)
+        return e;
+    if ((e = sdmsg_put (m, type, val)) < 0)
+        return e;
+    if ((e = sd_bus_message_close_container (m)) < 0)
+        return e;
+    return 0;
+}
+
+static int sdmsg_put_struct (sd_bus_message *m, const char *fmt, json_t *o)
+{
+    int e;
+
+    if ((e = sd_bus_message_open_container (m, 'r', fmt)) < 0)
+        return e;
+    if ((e = sdmsg_write (m, fmt, o)) < 0)
+        return e;
+    if ((e = sd_bus_message_close_container (m)) < 0)
+        return e;
+    return 0;
+}
+
+int sdmsg_put (sd_bus_message *m, const char *fmt, json_t *o)
+{
+    int e;
+
+    if (streq (fmt, "a(sv)"))
+        e = sdmsg_put_array (m, "(sv)", o);
+    else if (streq (fmt, "a(sasb)"))
+        e = sdmsg_put_array (m, "(sasb)", o);
+    else if (fmt[0] == 'a' && strlen (fmt) == 2)
+        e = sdmsg_put_array (m, &fmt[1], o);
+    else if (streq (fmt, "(sv)"))
+        e = sdmsg_put_struct (m, "sv", o);
+    else if (streq (fmt, "(sasb)"))
+        e = sdmsg_put_struct (m, "sasb", o);
+    else if (streq (fmt, "v"))
+        e = sdmsg_put_variant (m, o);
+    else if (strlen (fmt) == 1)
+        e = sdmsg_put_basic (m, fmt[0], o);
+    else
+        e = -EPROTO;
+    return e;
+}
+
+int sdmsg_write (sd_bus_message *m, const char *fmt, json_t *o)
+{
+    int e;
+
+    if (!json_is_array (o))
+        return -EPROTO;
+
+    int cursor = 0;
+    for (int i = 0; fmt[i] != '\0';) {
+        json_t *entry;
+        char *efmt = NULL;
+
+        if (!(entry = json_array_get (o, cursor++)))
+            return -EPROTO;
+        if (!strncmp (&fmt[i], "a(sasb)", 7))
+            efmt = strndup (&fmt[i], 7);
+        else if (!strncmp (&fmt[i], "a(sv)", 5))
+            efmt = strndup (&fmt[i], 5);
+        else if (fmt[i] == 'a' && strlen (&fmt[i]) > 1)
+            efmt = strndup (&fmt[i], 2);
+        else
+            efmt = strndup (&fmt[i], 1);
+        if (!efmt)
+            return -ENOMEM;
+        if ((e = sdmsg_put (m, efmt, entry)) < 0) {
+            free (efmt);
+            return e;
+        }
+        i += strlen (efmt);
+        free (efmt);
+    }
+    return 0;
+}
+
+static int sdmsg_get_string (sd_bus_message *m, char type, json_t **op)
+{
+    int e;
+    const char *val = NULL;
+    json_t *o = NULL;
+
+    if ((e = sd_bus_message_read_basic (m, type, &val)) <= 0)
+        return e;
+    switch (type) {
+        case 'g':
+        case 's':
+            o = json_string (val);
+            break;
+        case 'o': {
+            char *tmp;
+            if (!(tmp = objpath_decode (val)))
+                return -EPROTO;
+            o = json_string (tmp);
+            free (tmp);
+            break;
+        }
+        default:
+            return -EPROTO;
+    }
+    if (!o)
+        return -ENOMEM;
+    *op = o;
+    return 1;
+}
+
+static int sdmsg_get_basic (sd_bus_message *m, char type, json_t **op)
+{
+    char peek_type;
+    int e;
+    json_t *o;
+    number_t n;
+
+    if ((e = sd_bus_message_peek_type (m, &peek_type, NULL)) <= 0)
+        return e;
+    if (type == 0)
+        type = peek_type;
+    if (type != peek_type)
+        return -EPROTO;
+    if (type == 'g' || type == 's' || type == 'o')
+        return sdmsg_get_string (m, type, op);
+    if ((e = sd_bus_message_read_basic (m, type, &n)) <= 0)
+        return e;
+    switch (type) {
+        case 'y':
+            o = json_integer (n.u8);
+            break;
+        case 'n':
+            o = json_integer (n.i16);
+            break;
+        case 'q':
+            o = json_integer (n.u16);
+            break;
+        case 'i':
+            o = json_integer (n.i32);
+            break;
+        case 'u':
+            o = json_integer (n.u32);
+            break;
+        case 'x':
+            o = json_integer (n.i64);
+            break;
+        case 't':
+            o = json_integer (n.u64);
+            break;
+        case 'b':
+            o = n.i ? json_true () : json_false ();
+            break;
+        case 'h':
+            o = json_integer (n.i);
+            break;
+        case 'd':
+            o = json_real (n.f);
+            break;
+        default:
+            return -EPROTO;
+    }
+    if (!o)
+        return -ENOMEM;
+    *op = o;
+    return 1;
+}
+
+static int sdmsg_get_array (sd_bus_message *m, const char *fmt, json_t **op)
+{
+    json_t *a;
+    int e;
+
+    if (!(a = json_array ()))
+        return -ENOMEM;
+    if ((e = sd_bus_message_enter_container (m, 'a', fmt)) <= 0) {
+        json_decref (a);
+        return e;
+    }
+    while ((e = sdmsg_read (m, fmt, a)) > 0)
+        ;
+    if (e < 0) {
+        json_decref (a);
+        return e;
+    }
+    if ((e = sd_bus_message_exit_container (m)) < 0) {
+        json_decref (a);
+        return e;
+    }
+    *op = a;
+    return 1;
+}
+
+static int sdmsg_get_unknown (sd_bus_message *m, const char *fmt, json_t **op)
+{
+    int e;
+    if ((e = sd_bus_message_skip (m, fmt)) <= 0)
+        return e;
+    *op = json_null ();
+    return e;
+}
+
+static int sdmsg_get_variant (sd_bus_message *m, json_t **op)
+{
+    const char *contents;
+    char type;
+    json_t *val = NULL;
+    json_t *o = NULL;
+    int e;
+
+    if ((e = sd_bus_message_peek_type (m, &type, &contents)) <= 0)
+        return e;
+    if (type != 'v')
+        return -EPROTO;
+    if ((e = sd_bus_message_enter_container (m, 'v', contents)) <= 0)
+        return e;
+    if (strlen (contents) == 1)
+        e = sdmsg_get_basic (m, contents[0], &val);
+    else if (strlen (contents) == 2 && contents[0] == 'a')
+        e = sdmsg_get_array (m, contents + 1, &val);
+    else
+        e = sdmsg_get_unknown (m, contents, &val);
+    if (e <= 0)
+        return e;
+    if (!(o = json_pack ("[sO]", contents, val))) {
+        json_decref (val);
+        return -ENOMEM;
+    }
+    if ((e = sd_bus_message_exit_container (m)) < 0) {
+        json_decref (val);
+        json_decref (o);
+        return e;
+    }
+    json_decref (val);
+    *op = o;
+    return 1;
+}
+
+static int sdmsg_get_property_dict (sd_bus_message *m, json_t **op)
+{
+    json_t *dict;
+    int e;
+
+    if (!(dict = json_object ()))
+        return -ENOMEM;
+    if ((e = sd_bus_message_enter_container (m, 'a', "{sv}")) <= 0)
+        goto out;
+    while ((e = sd_bus_message_enter_container (m, 'e', "sv")) > 0) {
+        const char *key;
+        json_t *val;
+        int e;
+        if ((e = sd_bus_message_read (m, "s", &key)) <= 0
+            || (e = sdmsg_get_variant (m, &val)) <= 0)
+            goto out;
+        if (json_object_set_new (dict, key, val) < 0) {
+            json_decref (val);
+            goto nomem;
+        }
+        if ((e = sd_bus_message_exit_container (m)) < 0)
+            goto out;
+    }
+    if (e < 0 || (e = sd_bus_message_exit_container (m)) < 0)
+        goto out;
+    *op = dict;
+    return 1;
+nomem:
+    e = -ENOMEM;
+out:
+    json_decref (dict);
+    return e;
+}
+
+int sdmsg_get (sd_bus_message *m, const char *fmt, json_t **op)
+{
+    int e;
+
+    if (streq (fmt, "a{sv}"))
+        e = sdmsg_get_property_dict (m, op);
+    else if (fmt[0] == 'a' && strlen (fmt) > 1)
+        e = sdmsg_get_array (m, fmt + 1, op);
+    else if (streq (fmt, "v"))
+        e = sdmsg_get_variant (m, op);
+    else if (strlen (fmt) == 1)
+        e = sdmsg_get_basic (m, fmt[0], op);
+    else
+        e = -EPROTO;
+    return e;
+}
+
+int sdmsg_read (sd_bus_message *m, const char *fmt, json_t *o)
+{
+    for (int i = 0; fmt[i] != '\0';) {
+        json_t *entry = NULL;
+        char *efmt = NULL;
+        int e;
+
+        if (!strncmp (&fmt[i], "a{sv}", 5))
+            efmt = strndup (&fmt[i], 5);
+        else if (fmt[i] == 'a' && strlen (&fmt[i]) > 1)
+            efmt = strndup (&fmt[i], 2);
+        else
+            efmt = strndup (&fmt[i], 1);
+        if (!efmt)
+            return -ENOMEM;
+
+        if ((e = sdmsg_get (m, efmt, &entry)) <= 0) {
+            free (efmt);
+            if (e == 0 && i > 0)
+                e = -EPROTO;
+            return e;
+        }
+        if (json_array_append_new (o, entry) < 0) {
+            free (efmt);
+            json_decref (entry);
+            return -ENOMEM;
+        }
+        i += strlen (efmt);
+        free (efmt);
+    }
+    return 1;
+}
+
+// vi:ts=4 sw=4 expandtab
--- a/src/modules/sdbus/message.h
+++ b/src/modules/sdbus/message.h
@@ -1,0 +1,44 @@
+/************************************************************\
+ * Copyright 2023 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+#ifndef _SDBUS_MESSAGE_H
+#define _SDBUS_MESSAGE_H
+
+#include <systemd/sd-bus.h>
+#include <jansson.h>
+
+const char *sdmsg_typestr (sd_bus_message *m);
+
+/* Put one value (or container) specified by 'fmt' from json object 'o' to
+ * current cursor position of message 'm'.  Return 0 on success, or -errno
+ * on failure.
+ */
+int sdmsg_put (sd_bus_message *m, const char *fmt, json_t *o);
+
+/* Put list of values specified by 'fmt' from json array 'o' to current cursor
+ * position of message 'm'.  Return 0 on success or -errno on failure.
+ */
+int sdmsg_write (sd_bus_message *m, const char *fmt, json_t *o);
+
+/* Get one value (or container) specified by 'fmt' from message 'm' at the
+ * current cursor position and return in a new json object assigned to 'op'.
+ * Return 1 on success, or -errno on failure.
+ */
+int sdmsg_get (sd_bus_message *m, const char *fmt, json_t **op);
+
+/* Get list of values specified by 'fmt' from message 'm' at the current cursor
+ * position and append them to the json array 'o'.
+ * Return 1 on success, or -errno on falure.
+ */
+int sdmsg_read (sd_bus_message *m, const char *fmt, json_t *o);
+
+#endif /* !_SDBUS_MESSAGE_H */
+
+// vi:ts=4 sw=4 expandtab
--- a/src/modules/sdbus/sdbus.c
+++ b/src/modules/sdbus/sdbus.c
@@ -1,0 +1,640 @@
+/************************************************************\
+ * Copyright 2023 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* sdbus.c - sd-bus bridge for user-mode systemd
+ */
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+#include <flux/core.h>
+#include <systemd/sd-bus.h>
+
+#include "src/common/libutil/errno_safe.h"
+#include "src/common/libutil/errprintf.h"
+#include "ccan/str/str.h"
+
+#include "message.h"
+#include "interface.h"
+#include "watcher.h"
+#include "subscribe.h"
+#include "connect.h"
+#include "sdbus.h"
+
+struct sdbus_ctx {
+    flux_future_t *f_conn; // owns ctx->bus
+    sd_bus *bus;
+    flux_watcher_t *bus_w;
+    flux_msg_handler_t **handlers;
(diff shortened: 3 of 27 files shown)


src/modules/sdbus/sdbus.c
```diff
@@ -0,0 +1,457 @@
...
+                  LOG_ERR,
+                  "unsubscribe failed: %s",
+                  future_strerror (f, errno));
+    }
+}
+
+/* Subscribe when first subscriber is added.
+ * Unsubscribe when last subscriber is removed.
+ */
+static void update_subscribe (struct sdbus_ctx *ctx, int delta)
```
chu11, line 315: medium level comment: this function took me awhile to grok, perhaps just split up into subscribe and unsubscribe separate functions.

chu11 (commented):
a few initial comments through the `modules/sdbus: new service` commit.  Don't have the 100% picture yet, but I think I see where you're going and I like the idea.  You went a level lower in the protocol than I did i.e. `sd_bus_send()` where as I did `sd_bus_call_method()` so many aspects of the protocol that I just assume the protocol is working as you want :-)

one high level comment, I notice that you don't call `sd_bus_set_method_call_timeout()` at any point.  Did you not see any timeouts during your testing?  Granted the systemd version you're testing against might be much improved over the one I initially developed against, but it may be wise to add just in case.  The default timeout from a method call is 25 seconds.

src/modules/sdbus/progress.c
```diff
@@ -0,0 +1,171 @@
...
+        flux_watcher_start (prog->in);
+    if ((events & POLLOUT))
+        flux_watcher_start (prog->out);
+    flux_timer_watcher_reset (prog->tmout, timeout, 0.);
+    if (timeout >= 0.)
+        flux_watcher_start (prog->tmout);
+    return;
+fatal:
+    flux_reactor_stop_error (flux_get_reactor (prog->h));
+}
```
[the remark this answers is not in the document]
chu11: I should also note, I don't think events == 0 and timeout != 0 is possible, but worthwhile to add a check and log if it happens?

src/modules/sdbus/progress.c
```diff
@@ -0,0 +1,171 @@
...
+ * Copyright 2023 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* progress.c - integrate sd_bus with the flux reactor
```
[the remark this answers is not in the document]
chu11: Or sdbus_router?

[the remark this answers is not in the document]
garlick: I used "progress" because that code is letting the sdbus API's event loop have some time to make progress whenever events/timeouts are pending.  "router" or "bridge" is what the whole flux sdbus module is doing so I was looking for something that communicated event loop integration.  Maybe "reactor" or "eventloop" would be a better choice?

src/modules/sdbus/progress.c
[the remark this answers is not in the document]
garlick: [sd_bus_get_fd(3)](https://manpages.ubuntu.com/manpages/focal/man3/sd_bus_get_fd.3.html) describes what I was trying to do here:
>  These three functions are useful to hook up a bus connection
       object with an external or manual event loop involving poll() or
       a similar I/O polling call. Before each invocation of the I/O
       polling call, all three functions should be invoked: the file
       descriptor returned by sd_bus_get_fd() should be polled for the
       events indicated by sd_bus_get_events(), and the I/O call should
       block for that up to the timeout returned by
       sd_bus_get_timeout(). After each I/O polling call the bus
       connection needs to process incoming or outgoing data, by
       invoking sd_bus_process(3).

This is accomplished by setting up flux POLLIN and POLLOUT watchers on the file descriptor, and a flux timer watcher for the timeout, and then whenever any of those is ready, calling `sd_bus_process()` to let the sdbus API event loop make progress and possibly get a received message for flux to process.

I don't think the order of those calls is significant.  We have to asynchronously block waiting for any of POLLIN, POLLOUT, or the timeout to be ready, and that blocking doesn't occur until the reactor callback exits.

src/modules/sdbus/sdbus.c
garlick: Good idea, that whole thing got me confused when I looked at it again also!  I'll refactor some way that makes that clearer.

src/modules/sdbus/progress.c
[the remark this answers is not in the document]
chu11: yeah, perhaps just eventloop.

src/modules/sdbus/progress.c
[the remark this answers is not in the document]
chu11: I'll have to play with this a bit later, but my recollection is this bit me a few times.  As I look through `libsdprocess` some more, I do not setup a timer watcher, only a fd watcher, so checking for timeout == 0 matters a lot more for that code.

chu11:
> My understanding is the timeouts apply to method calls when the sdbus API is matching responses to callbacks, like in sd_bus_call() or sd_bus_call_async(). I think each method-call request message has a timeout associated with it, and when you make one of those higher level calls, you can expect to get a callback either when a response is received or the timeout expires. However I think that timeout is implemented in the layer of code that dispatches messages to callbacks, which we don't use.

Yeah, I was wondering if it applied in your case.  i couldn't find super strong evidence in my limited reading.  If you can launch a process and it monitor it for > 25 seconds, then I think the answer is it doesn't apply!

garlick:
> Yeah, I was wondering if it applied in your case. i couldn't find super strong evidence in my limited reading. If you can launch a process and it monitor it for > 25 seconds, then I think the answer is it doesn't apply!

Quick test of sleep 60:
```console
$ cat sd.toml
[systemd]
enable = true
$ ./flux start -o,-c sd.toml
$ flux exec --service-name=sdexec sleep 60
$ flux dmesg|grep sdbus
2023-04-06T15:55:46.093665Z sdbus.debug[0]: bus send method-call cookie=5 StartTransientUnit
2023-04-06T15:55:46.094629Z sdbus.debug[0]: bus recv method-return cookie=5 StartTransientUnit
2023-04-06T15:55:46.124874Z sdbus.debug[0]: bus recv signal flux-0.service PropertiesChanged
2023-04-06T15:55:46.124987Z sdbus.debug[0]: bus recv signal flux-0.service PropertiesChanged
2023-04-06T15:55:46.125051Z sdbus.debug[0]: bus recv signal flux-0.service PropertiesChanged
2023-04-06T15:55:46.125110Z sdbus.debug[0]: bus recv signal flux-0.service PropertiesChanged
2023-04-06T15:55:46.125149Z sdbus.debug[0]: bus recv signal /org/freedesktop/systemd1/job/68972 PropertiesChanged
2023-04-06T15:55:46.125291Z sdbus.debug[0]: bus recv signal flux-0.service PropertiesChanged
2023-04-06T15:55:46.125389Z sdbus.debug[0]: bus recv signal flux-0.service PropertiesChanged
2023-04-06T15:55:46.125474Z sdbus.debug[0]: bus recv signal flux-0.service PropertiesChanged
2023-04-06T15:55:46.125543Z sdbus.debug[0]: bus recv signal flux-0.service PropertiesChanged
2023-04-06T15:56:46.126386Z sdbus.debug[0]: bus recv signal flux-0.service PropertiesChanged
2023-04-06T15:56:46.126501Z sdbus.debug[0]: bus recv signal flux-0.service PropertiesChanged
2023-04-06T15:56:46.126862Z sdbus.debug[0]: bus send method-call cookie=6 StopUnit
2023-04-06T15:56:46.127481Z sdbus.debug[0]: bus recv signal flux-0.service PropertiesChanged
2023-04-06T15:56:46.127580Z sdbus.debug[0]: bus recv signal flux-0.service PropertiesChanged
2023-04-06T15:56:46.127651Z sdbus.debug[0]: bus recv method-return cookie=6 StopUnit
2023-04-06T15:56:46.127738Z sdbus.debug[0]: bus recv signal flux-0.service PropertiesChanged
2023-04-06T15:56:46.127841Z sdbus.debug[0]: bus recv signal flux-0.service PropertiesChanged
2023-04-06T15:56:46.127927Z sdbus.debug[0]: bus recv signal flux-0.service PropertiesChanged
2023-04-06T15:56:46.128010Z sdbus.debug[0]: bus recv signal flux-0.service PropertiesChanged
2023-04-06T15:56:46.128113Z sdbus.debug[0]: bus recv signal flux-0.service PropertiesChanged
2023-04-06T15:56:46.128193Z sdbus.debug[0]: bus recv signal flux-0.service PropertiesChanged
```
I edited out some noise from the sdbus debug output but you get the idea - StartTransientUnit method call completes immediately, and the unit state is monitored via PropertiesChanged signals.

A difference from the earlier code is that sdbus subscribes to all signals when the module is loaded so there is no method call to subscribe as part of the unit launch.  Also, the unit state changes gleaned from the signals.  There doesn't appear to be a need for the GetAll method calls to request specific properties.

src/modules/sdbus/progress.c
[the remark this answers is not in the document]
garlick: Yes the main thing is that the flux reactor cannot be allowed to block if the sdbus reactor has a pending timeout.  If the timeout is 0, that is treated like any other timeout - the timer watcher becomes ready when the timeout expires, immediately in this case.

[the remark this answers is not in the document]
garlick: In practice, I don't think I've ever seen events set to anything but 1 (POLLIN).

However I don't think there's any requirement that timeout must be 0 if events is 0.  I think your concern must be about events = 0 (no events) and timeout = -1 (no timeout)?  I assumed sdbus wouldn't return that combination but if it did I don't think we would ever call the progress functions again.  Hmm, at minimum we should treat that as a fatal error if it occurs.  And if that ever happens, we might want to convert this code to use prep/check/idle watchers.  I'll have to think about that one - it might be better to use prep/check/idle anyway.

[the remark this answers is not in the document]
garlick: clarification: by timeout == -1 I mean the post-processed timeout in progress.c.  `sd_bus_get_timeout()` returns UINT64_MAX for "no timeout" and that is turned into -1 (compatible with the flux timer watcher) in `get_relative_timeout()`.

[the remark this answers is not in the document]
chu11: > Hmm, at minimum we should treat that as a fatal error if it occurs. 

This is where there are subtleties in `systemd` that I think neither of us truly know :P  I'm wondering if in `libsdprocess` we see events == 0 b/c of my unique setup and it simply isn't possible under yours.

Perhaps a couple of "we assume this is impossible" fatal error checks would be worthwhile and if it ends up biting us later, then code can be re-worked?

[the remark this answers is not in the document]
garlick: Oh maybe events is always 1 because I always have a subscription going.

> Perhaps a couple of "we assume this is impossible" fatal error checks would be worthwhile and if it ends up biting us later, then code can be re-worked?

Wholeheartedly agree!

chu11:
as a complete aside, when I saw your code in `objpath.[ch]` my immediate thought was "oh crap, functions for this exist?", i.e. in `libsdprocess`.   

```
    ptr = unitname;
    while (*ptr) {
        if (!((*ptr) >= 'A' && (*ptr) <= 'Z')
            && !((*ptr) >= 'a' && (*ptr) <= 'z')
            && !(ptr > unitname && (*ptr) >= '0' && (*ptr) <= '9')) {
            service_path[index++] = '_';
            service_path[index++] = hex2char[*ptr >> 4];
            service_path[index++] = hex2char[*ptr & 0xF];
        }
        else
            service_path[index++] = *ptr;
        ptr++;
    }
```

I guess `systemd-run` and stuff weren't updated at the time I worked on `libsdprocess`.  Glad you found them.

garlick:
Yeah that simplified things a little!

chu11:
high level comment, it seems `sdbus` returns ALL events to the caller and the caller is responsible for filtering out what they want?  I think this is fine for now, but perhaps some extra filtering down the road could be done.  Although I'm not sure what that filtering would be.

src/modules/sdexec/start.c
```diff
@@ -0,0 +1,235 @@
...
+        errno = EPROTO;
+        return NULL;
+    }
+    if (!(prop = json_array ())) {
+        errprintf (error, "out of memory");
+        errno = ENOMEM;
+        return NULL;
+    }
+    if (prop_add_execstart (prop, "ExecStart", cmdline) < 0
+        || prop_add_string (prop, "Type", "exec") < 0
```
chu11, line 167: just curious where you found "type=exec" documented?  Or dug up in systemd code?  Struggling to figure out where this field comes from, b/c I didn't do this in `libsdprocess`.

chu11 (commented):
still going through sdexec, just a few high level comments

src/modules/sdexec/start.c
garlick: The various types are described [here](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Type=)

chu11: ahhh i had totally forgotten that this stuff is strewn about various manpages too :-)   Perhaps useful to add documentation pointers when you can recall where they came from.  I probably should have done a better job of that in libsdprocess too.

garlick: No worries - you got it working!

Maybe we could add a markdown page in the sdbus source that includes references and (eventually) some documentation?   I would imagine someone who was not familiar at all with dbus or systemd would find this all pretty mystifying.

chu11: That sounds like a good idea.  I remember having difficulty even remembering where I found things documented, as little bits are found all over the place :-)

src/modules/sdexec/recover.c
```diff
@@ -0,0 +1,117 @@
...
+        flux_future_t *f;
+
+        if (sdexec_parse_unit (entry, &unit) < 0) {
+            flux_log (h, LOG_ERR, "recover: error parsing unit object");
+            goto out;
+        }
+        if (streq (unit.active, "failed"))
+            f = sdexec_reset_failed_unit (h, unit.name);
+        else
+            f = sdexec_stop_unit (h, unit.name, "fail");
```
chu11, line 105: in `libsdprocess` i basically have logic of

```
        if (streq (unit.active, "active")
            f = sdexec_stop_unit (h, unit.name, "fail");
        else if (streq (unit.active, "failed"))
            f = sdexec_reset_failed_unit (h, unit.name);
        else
            /* error unexpected state */
```

I recall on rare occasions getting other states, which I presumed was due to something racy, like perhaps catching a unit at that exact moment it's between two states we care about.  And IIRC reset/fail and stop unit may not work.  Most notably is the use of RemainAfterExit, b/c I think a state can go from inactive back to active.

But your logic might be better, b/c you are doing 

```
        if (streq (unit.active, "failed"))
            f = sdexec_reset_failed_unit (h, unit.name);
        else
            f = sdexec_stop_unit (h, unit.name, "fail");
```

i.e. do stop if not failed, and that might work for all states != "failed".

Not sure if you've seen issues here, but perhaps worthwhile to atleast log if the active state is something other than what we expect??

src/modules/sdexec/sdexec.c
```diff
@@ -0,0 +1,622 @@
...
+            return m;
+        m = flux_msglist_next (msglist);
+    }
+    return NULL;
+}
+
+static void start_properties_changed (struct sdexec_ctx *ctx,
+                                      const flux_msg_t *msg,
+                                      json_t *properties)
+{
```
chu11, line 268: idea, document expected state transitions that can/should occur within this function and log unexpected state transitions?  B/c its systemd, some bad admin may do something that wouldn't be expected normally.

src/modules/sdbus/test/message.c
```diff
@@ -0,0 +1,472 @@
...
+    const char *type;
+    json_t *val;
+    ok (json_unpack (out, "[[so]]", &type, &val) == 0
+        && streq (type, "as")
+        && json_is_array (val)
+        && json_array_size (val) == 3
+        && json_is_string (json_array_get (val, 0))
+        && json_is_string (json_array_get (val, 1))
+        && json_is_string (json_array_get (val, 2)),
+        "string array variant was translated properly");
```
chu11, line 293: the anal part of me says to check the values too :-)

src/modules/sdbus/test/objpath.c
```diff
@@ -0,0 +1,79 @@
...
+}
+
+int main (int argc, char **argv)
+{
+    sd_bus *bus;
+    int e;
+
+    plan (NO_PLAN);
+
+    if ((e = sd_bus_open_user (&bus)) < 0) {
```
chu11, line 64: i added this into `libsdprocess` unit tests as a sanity check b/c a lot of times they may be run under the developer's user name and not the `flux` user and they may have forgot to set things up.

```
    if (!(dbusenv = getenv ("DBUS_SESSION_BUS_ADDRESS"))
        || !(xdgenv = getenv ("XDG_RUNTIME_DIR"))) {
        diag ("DBUS_SESSION_BUS_ADDRESS or XDG_RUNTIME_DIR not set");
        done_testing ();
    }

    /* Instead of checking if user service running via some equivalent to
     *
     * systemctl list-units user@UID.service
     *
     * just make sure the path to XDG_RUNTIME_DIR exists
     *
     * we don't check DBUS_SESSION_BUS_ADDRESS since it often has
     * funky formatting (e.g. unix:path=/run/user/8556/bus)
     */
    if (access (xdgenv, R_OK | W_OK) < 0) {
        diag ("cannot access XDG_RUNTIME_DIR");
        done_testing ();
    }
```

src/modules/sdbus/test/message.c
```diff
@@ -0,0 +1,472 @@
...
+        "sdmsg_read works on message containing complex variant");
+
+    diagjson (o);
+
+    const char *s1, *s2, *type;
+    ok (json_unpack (o, "[s[sn]s]", &s1, &type, &s2) == 0
+        && streq (s1, "eek")
+        && streq (type, "a(yy)")
+        && streq (s2, "ook"),
+        "complex variant was translated to json null");
```
chu11, line 338: took me awhile to grok, but I think you're basically testing if the format is one that sdmsg/interface doesn't know how to handle (b/c you hard code selected ones), it gets converted into json null?

chu11:
Haven't reviewed the last test commit yet but overall I like the approach.  I think my main concern was the amount of messages coming from `sdbus`, but since `sdbus` is going to loaded on every node, and presumably its only going to managing some relatively small-ish finite number of processes local to that node, the number of messages shouldn't be too much.  Some filtering might be ok to eliminate an excess number of "garbage" messages coming out?

t/t2407-sdbus.t
```diff
@@ -0,0 +1,117 @@
+#!/bin/sh
+# ci=system
+
+test_description='Test flux systemd sd-bus gatweay'
```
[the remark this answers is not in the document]
grondo: the tests under `t/system` are meant to be those that run against a system instance of Flux. These tests appear standalone to me, so I don't think they should be moved into `t/system`.

t/t2407-sdbus.t
```diff
@@ -0,0 +1,117 @@
...
+. $(dirname $0)/sharness.sh
+
+if ! flux version | grep systemd; then
+	skip_all="flux was not built with systemd"
+	test_done
+fi
+if ! systemctl --user show --property Version; then
+	skip_all="user systemd is not running"
+	test_done
+fi
```
chu11, line 15: similar to my comment in the unit tests, should there be additional checks?  I imagine common scenario is user running tests and this might be run against a user service which may not be setup.

src/modules/sdbus/test/objpath.c
garlick: The XDG_RUNTIME_DIR isn't required if DBUS_SESSION_BUS_ADDRESS is set, is it?  See  [source code](https://github.com/systemd/systemd/blob/fde55f3a327c4eef3bbdeb199b05e7c385d1f331/src/libsystemd/sd-bus/sd-bus.c#L1333).  One or the other needs to be set in order for the bus open to succeed, but the tests will still fail if the user session isn't actually running, hence the check.

Was your point that you'd like further hints if the test environment isn't set up right?

src/modules/sdbus/test/message.c
garlick: Correct.  Maybe a comment to that effect would be a good thing for me to add.

src/modules/sdexec/sdexec.c
garlick: Yes good idea!

src/modules/sdexec/sdexec.c
```diff
@@ -0,0 +1,622 @@
...
+        if (flux_respond_pack (ctx->h,
+                               msg,
+                               "{s:s s:I}",
+                               "type", "finished",
+                               "status", wait_status) < 0)
+            flux_log_error (ctx->h, "error responding to exec request");
+        channel_start_output (proc->out);
+        channel_start_output (proc->err);
+        proc->state = SD_FINISHED;
+    }
```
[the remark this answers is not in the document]
garlick: That looks wrong.  I guess it would mean all the output gets processed late, and most likely things will hang if there is significant output before that a kernel buffer fills up.  Not sure how that ended up like that - will  investigate/fix.  Thanks for catching it!

src/modules/sdexec/recover.c
garlick: I'll ponder that one.  I think your idea of documenting/logging all the state changes might help this become clearer.

src/modules/sdbus/test/objpath.c
chu11: > Was your point that you'd like further hints if the test environment isn't set up right?

Yup, basically just give extra hints.

You might be right that XDG_RUNTIME_DIR isn't required.  Perhaps it was before?  Or perhaps I just got that from some random guide and have been going with it this whole time.

garlick: Ah actually, calling systemctl does print hints on my system at least:
```
./t*sdbus*t
build-options:		+systemd+hwloc==2.5.0+zmq==4.3.4
Failed to connect to bus: $DBUS_SESSION_BUS_ADDRESS and $XDG_RUNTIME_DIR not defined (consider using --machine=<user>@.host --user to connect to bus of other user)
1..0 # SKIP user systemd is not running
```

garlick:
Pushed an update to the sdbus module and its tests:
- reimplemented progress.c as a "bus watcher", which is less code and more
  familar as a flux abstraction.  It uses prep/check watchers internally and
  so is less fragile with respect to suggested sd_bus_get_timeout()
  and sd_bus_get_events() corner cases/systemd version dependent behavior.
- reimplemented sdbus sharness test standalone, with internal python one liners
  in place of the 'flux sdctl' C test utility and increased coverage
- added notes to sdbus message unit test explaining what is being tested and
  why. Shorten one test by testing for in/out JSON equality after
  json->dbus->json rather than checking individual JSON values.  If sdbus
  access fails and env vars are not set, print diag info before skipping test.
- added README.md with some D-Bus and systemd references
- `sdbus_subscribe()` and `unsubscribe()` composite RPCs refactored without
  confusing helper function

src/modules/sdbus/watcher.c
```diff
@@ -0,0 +1,156 @@
...
+/* called just after poll(2) return */
+static void check_cb (flux_reactor_t *r,
+                      flux_watcher_t *w,
+                      int revents,
+                      void *arg)
+{
+    struct sdbus_watcher *sdw = arg;
+
+    if (sdw->cb)
+        sdw->cb (r, sdw->w, FLUX_POLLIN, sdw->cb_arg);
```
chu11, line 83: this initially threw me for a loop, until i realized the `revents` here doesn't really matter.  Could use a comment.  Also, perhaps could use POLLOUT instead of POLLIN, given this is the subscription callback.

src/modules/sdbus/watcher.c
```diff
@@ -0,0 +1,156 @@
...
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* watcher.c - a flux watcher that becomes ready when sd-bus needs service
+ */
```
chu11, line 12: as discussed offline, perhaps a small snippet to discuss whats going on.  Since many of the watchers don't take callbacks, I was confused what this was inititally doing.

chu11: perhaps should also note, part of my initial confusion was that the `check_cb()` will be called even when a non-systemd related event (e.g. random other flux rpcs) trigger in libev land.  Should be fine, as `sd_bus_process()` should be fast, but that confused me as well.

garlick: On further consideration, I think your intuition is correct: that the fd and timer watchers should call the bus watcher callback, not a check watcher.  That should eliminate the needless callbacks.  I'm not sure why I thought the check watcher would help.

chu11 (commented):
I really like the automatic reconnect stuff, it makes things a lot more simple / obvious.  Only major thought is, what if someone doesn't want it to connect?  i assume the module should never be loaded?

I can't think of a situation where someone would load this module and not want it to connect.   I guess some user is starting a local test instance and accidentally loads it but doesn't want to?  Not sure if some type of "yup I really want this" kinda config would be necessary?

I'm not super sure, but food for thought.

src/modules/sdbus/watcher.c
garlick: I'm redoing this that the fd/timer watcher revents is just passed through to the callback.  I'm hoping that won't require any comment.  It seems clear that revents is not useful in the callback - you have to call `sd_bus_process()` regardless.

garlick:
Thanks for all the review @chu11!  Per offline discussion, I've squashed development and sliced off the sdexec porion to make this a standalone PR that adds the `sdbus` module and tests.  I have to head out for a bit but later I'll do some testing on rhel 8 etc.

garlick:
@chu11 - I tacked on a couple of small tweaks to sharness tests after testing on el8.  I think this is probably in a fairly stable state now if you'd like to make another review pass.

garlick:
>  Only major thought is, what if someone doesn't want it to connect? i assume the module should never be loaded?
>
> I can't think of a situation where someone would load this module and not want it to connect. I guess some user is starting a local test instance and accidentally loads it but doesn't want to? Not sure if some type of "yup I really want this" kinda config would be necessary?

The module doesn't load unless the configuration sets `systemd.enable = true` so I don't think that's much of a worry.  If it did happen, the exponential backoff would level off at a reconnect every 1m and a log message when that fails.  I think this probably is not a big concern.

src/modules/sdbus/test/message.c
garlick: This test now calls `json_equal()` on the input and output objects of json->dbus->json so the values are now checked.

t/t2407-sdbus.t
```diff
@@ -0,0 +1,246 @@
...
+    bus_list_units "$1" >tmp.json || return 1
+    $jq -r ".params[0][][$2]" <tmp.json
+}
+
+# Usage: bus_count_units PATTERN
+bus_count_units() {
+    bus_list_units_parsed "$1" 0 >tmp.list || return 1
+    wc -l <tmp.list
+}
+
```
chu11, line 44: I think some, possibly not all, of your tests below could be racy against these list/count checks.  i.e. you start a unit then check if its running via listing, but it may not be running by the time you check.  IIRC the kill tests can also be racy, no guarantee that the signal has been received by the process by the time you check.

Perhaps it would be safer to wrapper these in a loop and spin if the number / string / whatever isn't what you expect it to be yet.

t/t2408-sdbus-recovery.t
```diff
@@ -0,0 +1,77 @@
...
+	bus_get_manager_prop Version
+'
+test_expect_success 'clear the broker ring buffer' '
+	flux dmesg -C
+'
+# This should trigger a 2s backoff before reconnect is attempted
+test_expect_success 'force bus reconnect' '
+	bus_reconnect
+'
+test_expect_success NO_CHAIN_LINT 'initiate subscription the background' '
```
chu11, line 55: "in the background"?

t/t2407-sdbus.t
```diff
@@ -121,11 +121,13 @@ test_expect_success 'there are no flux-t2406 units loaded' '
 '
 test_expect_success 'create subscribe script' '
 	cat >subscribe.py <<-EOT &&
+	import sys
 	import flux
 	handle = flux.Flux()
 	fut = handle.rpc("sdbus.subscribe",{},flags=flux.constants.FLUX_RPC_STREAMING)
 	while True:
 	    print(fut.get_str())
+	    sys.stdout.flush()
```
chu11: copy over to reconnect tests script as well?

chu11 (commented):
saved review of the giant commit till last, here's the review for everything else first.  Mostly small stuff, but I do think there is potential raciness in the sharness tests.  Would be best to preemptively handle.

src/modules/sdbus/test/objpath.c
```diff
@@ -0,0 +1,83 @@
...
+    const char *path;
+};
+
+static struct testvec opvec[] = {
+    { "/object/path/foo.suffix",
+      "/object/path/foo_2esuffix" },
+    { "/org/freedesktop/systemd1/unit/flux-foo.service",
+      "/org/freedesktop/systemd1/unit/flux_2dfoo_2eservice" },
+    { "/foo/flea-bag",
+      "/foo/flea_2dbag" },
```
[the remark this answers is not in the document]
garlick: I don't quite get the joke? :confused:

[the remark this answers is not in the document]
chu11: doh! https://en.wikipedia.org/wiki/Flea_(musician)

t/t2407-sdbus.t
```diff
@@ -0,0 +1,246 @@
...
+
+test_under_flux 1 minimal
+
+flux setattr log-stderr-level 1
+
+#
+# N.B. ListUnitsByPatterns response payload is a 'params' array whose first
+# and only item (".params[0]") is an array of units.  The jq(1) expression
+# ".params[0][][0]" extracts the first field of each unit array entry (the
+# name is field 0) and prints one entry per line.
```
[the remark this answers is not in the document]
garlick: Still not clear?  OK , I can add a bit more.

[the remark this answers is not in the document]
chu11: you use index 1 later on below, just thought index 1 could be documented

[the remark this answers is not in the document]
garlick: This better?
```sh
# N.B. ListUnitsByPatterns response payload is a 'params' array whose first
# and only item (".params[0]") is an array of units.  The jq(1) expression
# ".params[0][][0]" extracts field 0 of each unit array entry (unit name)
# and prints one entry per line, ".params[0][][1]" extracts field 1 of each
# unit entry (description) and prints one entry per line, etc.
```

t/t2407-sdbus.t
garlick: Ugh, I think you're right.  Let me rework.

src/modules/sdbus/test/objpath.c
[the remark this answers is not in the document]
garlick: Oh duh :smile:

garlick:
Squashed fixups and re-pushed per offline discussion with @chu11

src/modules/sdbus/message.c
```diff
@@ -0,0 +1,486 @@
...
+        else if (fmt[i] == 'a' && strlen (&fmt[i]) > 1)
+            efmt = strndup (&fmt[i], 2);
+        else
+            efmt = strndup (&fmt[i], 1);
+        if (!efmt)
+            return -ENOMEM;
+        if ((e = sdmsg_put (m, efmt, entry)) < 0) {
+            free (efmt);
+            return e;
+        }
```
chu11, line 238: i think json_decref(entry) on error path?

src/modules/sdbus/message.h
```diff
@@ -0,0 +1,29 @@
...
+
+const char *sdmsg_typestr (sd_bus_message *m);
+
+int sdmsg_put (sd_bus_message *m, const char *fmt, json_t *o);
+
+int sdmsg_write (sd_bus_message *m, const char *fmt, json_t *o);
+
+int sdmsg_get (sd_bus_message *m, const char *fmt, json_t **op);
+
+int sdmsg_read (sd_bus_message *m, const char *fmt, json_t *o);
```
chu11, line 25: perhaps add comments on return values, since -errno is not the most common expected return.  and `sdmsg_read()` returns 1 on success instead of 0?  Perhaps follows another function from `sd_bus`?

chu11 (commented):
not through the big commit yet, but a few more comments

src/modules/sdbus/message.c
garlick: No, `entry` is an element of the function parameter `o` which belongs to the caller.  We use `json_array_get()` to access it which does not take a reference.

src/modules/sdbus/message.h
garlick: Great point.  I was following libsystemd's error conventions so these functions can be easily mixed with `sd_bus_*` ones.  I should have some comment to that effect inline.

src/modules/sdbus/sdbus.c
```diff
@@ -0,0 +1,640 @@
...
+}
+
+/* Bus subscribe completed.  Henceforth, sd-bus signals will be forwarded
+ * to subscribers.  Service pending sdbus.call requests.
+ * N.B. servicing pending sdbus.call requests when the connect is finalized
+ * might cause problems for a user that asynchronously subscribes to signals,
+ * then initiates an action and expects the subscription to capture all signals
+ * triggered by the action.
+ */
+static void subscribe_continuation (flux_future_t *f, void *arg)
```
chu11, line 517: nit: rename to avoid confusion with `subscribe_cb`?  bus_subscribe_continuation?

chu11 (commented):
nothing big stands out to me in the rest of the big commit

garlick:
Thanks!  I'll squash the fixups for your comments and force a push.

chu11 (approved):
LGTM!

garlick:
I wonder if this systemd stuff should go in a separate framework project, like `flux-systemd`?
- It is complex yet only helps the system instance (and is not even _required_ for that, at least not now) so it is arguably not "core" functionality
- hacking requires specialized understanding  of dbus, systemd concepts and API, so maybe having a different set of maintainers/contributors would make sense at some point?
- it may benefit from having its own CI matrix since right now I think it's only tested in the el8 system,coverage builder

I'm not sure those benefits outweigh the cost of setup etc.  Just putting the idea out there for consideration before we merge anything.

chu11:
A few points I'm just thinking about out loud.

- i think `sdexec` will be uhhh "protocol connected" with `libsubprocess`, so there's a connection to `libsubprocess` there?
- will all of the resource containment stuff be limited to `flux-systemd-exec`, or will it have to bleed into `flux-core` anyways?
  (side note, it seems that in my early prototype work in #4255 it did bleed into `job-exec`, but the `sdexec` and `libsdprocess` design was super different, haven't really thought too much about it in your design).
  
without thinking about it a lot, the benefit of creating a CI that's specific for systemd stuff is a strong selling point and makes me lean towards another project.  We've setup a non-trivial amount of "special case" testing for systemd already in `flux-core`.

garlick:
> i think sdexec will be uhhh "protocol connected" with libsubprocess, so there's a connection to libsubprocess there?

I think it's probably OK to separate the protocol from the implementation.  In fact, having more than one way to remotely start work could be a thing  (remember BlueGene?)  RFC for the protocol needed I suppose.  I did expend some effort simplifying it a bit before this work began.

> will all of the resource containment stuff be limited to flux-systemd-exec, or will it have to bleed into flux-core anyways?

Probably not but I was thinking at least at the lower levels we could just use `flux_cmd_setopt()` to set containment options that pass through opaquely.   I'm not sure what the rest will look like - we were just going to do simple stuff at first like install a default memory cap.

grondo:
You might actually want to determine what the strategy would be for testing systemd/sdbus interfaces under CI before splitting off a new repo if that the main reason for wanting to do that. It is a bit tricky to start a docker container which runs systemd (this is part of the reason there is only an el8 container that does this), but perhaps if you don't need to run a full systemd system then it could be easier to implement a good matrix of testing.

grondo:
> will all of the resource containment stuff be limited to flux-systemd-exec, or will it have to bleed into flux-core anyways?

Use of systemd for resource containment was originally proposed as a stopgap until we require or need the same support in the IMP. I had somehow thought it would be easier to propagate the resource control options down to systemd. It isn't clear where we should go long term, but it seems like requiring systemd in order to do resource containment is probably not the correct approach long term.

garlick:
I think the stronger reason for doing this in a separate repo is that it's complicated, not necessarily required by all or even most end users, and some specialized knowledge is needed to work on it.  OTOH,  adding more work before we can use this sucks, and I suppose we could always spin it off to another repo later, maybe when we've decided to go another way.  I guess I'm leaning toward the latter at this point.

chu11:
> OTOH, adding more work before we can use this sucks, and I suppose we could always spin it off to another repo later, maybe when we've decided to go another way. I guess I'm leaning toward the latter at this point.

This is my feeling too.  This conversation has us waffling and it'd probably be worse to put it in its own repo only to decide later "oh it should go into flux-core afterall`.

garlick:
Thanks for the discussion guys.  I think we can go ahead and merge this and table the idea of a separate repo for now.  I really appreciate the detailed review @chu11!

chu11:
looks like our typo checker is doing what is supposed to do!

image

PR #5078, opened 2023-04-07
broker: improve systemd integration with sd_notify(3)

garlick:
Systemd's [sd_notify(3)](https://www.freedesktop.org/software/systemd/man/sd_notify.html) provides a mechanism to improve systemd integration for the flux system instance brokers.

This PR contains the following improvements
1) notify systemd of the broker's status during startup/shutdown
2) delay the rank 0 unit transition to "active" by switching to `Type=notify` and sending a notification on rank 0 only once the broker has reached run state (e.g. it's ready to run stuff)

Here's a snapshot of `systemctl status flux` output during startup.  Note the new "Status:" line showing that the broker is running rc1.  Also note that the Active: line shows "activating (start)".  Later it will change to "active (running)".
```
     Loaded: loaded (/usr/local/lib/systemd/system/flux.service; enabled; vendor preset: enabled)
     Active: activating (start) since Fri 2023-04-07 15:19:10 PDT; 2s ago
    Process: 1001585 ExecStartPre=/usr/bin/loginctl enable-linger flux (code=exited, status=0/SUCCESS)
    Process: 1001586 ExecStartPre=bash -c systemctl start user@$(id -u flux).service (code=exited, status
=0/SUCCESS)
   Main PID: 1001588 (flux-broker-0)
     Status: "Running /usr/local/etc/flux/rc1"
      Tasks: 20 (limit: 8755)
        CPU: 2.627s
     CGroup: /system.slice/flux.service
             ├─1001588 broker --config-path=/usr/local/etc/flux/system/conf.d -Scron.directory=/usr/local
/etc/flux/system/cron.d -Srundir=/run/flux -Sstatedir=/var/lib/flux -Slocal-uri=local:///run/flux/local -
Slog-stderr-level=6 -Slog-stderr-mode=local -Sbroker.rc2_none -Sbroker.quorum=0 -Sbroker.quorum-timeout=n
one -Sbroker.exit-norestart=42 -Sbroker.sd-notify=1 -Scontent.restore=auto
             ├─1001592 /bin/bash -e /usr/local/etc/flux/rc1
             ├─1001661 /bin/bash -e /usr/local/etc/flux/rc1.d/01-coral2-rc
             └─1001667 /usr/bin/python3 /usr/local/libexec/flux/cmd/flux-jobtap.py load cray_pals_port_
```

The delayed transition to active also causes `systemctl start flux` to block until the unit is active.   I _think_ this is a good thing since it gives sys admins a better feel for what is happenig when flux starts.  Note that on ranks > 0, we give the notification earlier to avoid blocking startup there, since those brokers cannot reach the run state until their upstream brokers do.

I think this delay will also delay the start up of dependent units like flux-accounting.  Not sure if that is helpful for that particular case, but it might be for future units that want to submit jobs or something.

sd_notify(3) can also be used to extend our `TimeoutStopSec=90` to solve the problem of shutdown getting interrupted if it takes too long.  That's a bit more complicated to orchestrate so I didn't try that here, but this does provide some groundwork to enable it.

This PR is currently based on top of #5077

### DIFF
--- a/etc/flux.service.in
+++ b/etc/flux.service.in
@@ -3,6 +3,7 @@ Description=Flux message broker
 Wants=munge.service
 
 [Service]
+NotifyAccess=main
 TimeoutStopSec=90
 KillMode=mixed
 ExecStart=/bin/bash -c '\
@@ -20,6 +21,7 @@ ExecStart=/bin/bash -c '\
   -Sbroker.quorum=0 \
   -Sbroker.quorum-timeout=none \
   -Sbroker.exit-norestart=42 \
+  -Sbroker.sd-notify=1 \
   -Scontent.restore=auto \
 '
 SyslogIdentifier=flux
--- a/etc/flux.service.in
+++ b/etc/flux.service.in
@@ -3,6 +3,7 @@ Description=Flux message broker
 Wants=munge.service
 
 [Service]
+Type=notify
 NotifyAccess=main
 TimeoutStopSec=90
 KillMode=mixed
--- a/src/broker/Makefile.am
+++ b/src/broker/Makefile.am
@@ -12,6 +12,7 @@ AM_CPPFLAGS = \
 	$(ZMQ_CFLAGS) \
 	$(LIBUUID_CFLAGS) \
 	$(JANSSON_CFLAGS) \
+	$(LIBSYSTEMD_CFLAGS) \
 	$(VALGRIND_CFLAGS)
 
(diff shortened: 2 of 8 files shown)


grondo (approved):
Nice! Tested in fluxorama with no issues!

garlick:
Thank you for the extra testing!  OK, I'll set MWP.

chu11 (approved):

Issue #5083, opened 2023-04-10
libsubprocess: stopped state not handled correct


chu11:
I noticed in the following test in `libsubprocess`:

```
ok 16 - subprocess state == RUNNING after flux_local_exec
ok 17 - flux_subprocess_kill SIGSTOP
# state_change_stopped: state = Stopped
ok 18 - subprocess state == STOPPED in state change handler
# state_change_stopped: state = Running
# state_change_stopped: state = Exited
ok 19 - flux_reactor_run returned zero status
```

Which obviously looks weird, that the state change callback got the "Stopped" state before "Running".

I think it's because in the child watcher we have:

```
    if (WIFSTOPPED (p->status)) {                                                                                                           
        if (p->ops.on_state_change)                                                                                                         
            (*p->ops.on_state_change) (p, FLUX_SUBPROCESS_STOPPED);                                                                         
    }    
```

the `on_state_change()` is called immediately rather than going through the state change prep/check functions, e.g. right below this:

```
    if (WIFEXITED (p->status) || WIFSIGNALED (p->status)) {                                                                                 
                                                                                                                                            
        /* remote/server code may have set FAILED on fatal errors.                                                                          
         */                                                                                                                                 
        if (p->state == FLUX_SUBPROCESS_RUNNING) {                                                                                          
            p->state = FLUX_SUBPROCESS_EXITED;                                                                                              
            state_change_start (p);                                                                                                         
        }  
```

So we need to redo things to go through the state change prep/check functions.  I am not quite sure how at the moment and it may take some re-thought.  See the `state_change_next()` function:

```
static flux_subprocess_state_t state_change_next (flux_subprocess_t *p)                                                                     
{                                                                                                                                           
    assert (p->state != FLUX_SUBPROCESS_FAILED);                                                                                            
                                                                                                                                            
    switch (p->state_reported) {                                                                                                            
    case FLUX_SUBPROCESS_INIT:                                                                                                              
        /* next state must be RUNNING */                                                                                                    
        return FLUX_SUBPROCESS_RUNNING;                                                                                                     
    case FLUX_SUBPROCESS_RUNNING:                                                                                                           
        /* next state is EXITED */                                                                                                          
        return FLUX_SUBPROCESS_EXITED;                                                                                                      
    case FLUX_SUBPROCESS_EXITED:                                                                                                            
    case FLUX_SUBPROCESS_FAILED:                                                                                                            
[...]

Issue #5080, opened 2023-04-09
building without asserts


chu11:
After comments in #5079, I was just curious if flux-core could be built without asserts and if the tests would pass.

There are a number of "unused-but-set-variable" and "unused-variable" errors and some stupid corner cases (e.g. I had a function in `libsubprocess` that ended in an assert, so got a warning that function didn't return a value).  But nothing super horrible (i.e. `assert (rv = something); return rv;` would be really really bad).  Tests pass.

I just thought I'd log my diff here just for note taking purposes and in case anyone in the future wanders into this/needs it.  I think we're a long ways off from wanting to build flux without asserts.

```
diff --git a/src/broker/Makefile.am b/src/broker/Makefile.am
index 4902119..41a9b88 100644
--- a/src/broker/Makefile.am
+++ b/src/broker/Makefile.am
@@ -1,6 +1,9 @@
 AM_CFLAGS = \
        $(WARNING_CFLAGS) \
-       $(CODE_COVERAGE_CFLAGS)
+       $(CODE_COVERAGE_CFLAGS) \
+       -Wno-error=unused-but-set-variable \
+       -Wno-error=unused-variable
+
 AM_LDFLAGS = \
        $(CODE_COVERAGE_LIBS)
 
diff --git a/src/cmd/Makefile.am b/src/cmd/Makefile.am
index 853d987..3e6eff4 100644
--- a/src/cmd/Makefile.am
+++ b/src/cmd/Makefile.am
@@ -1,6 +1,7 @@
 AM_CFLAGS = \
        $(WARNING_CFLAGS) \
-       $(CODE_COVERAGE_CFLAGS)
+       $(CODE_COVERAGE_CFLAGS) \
+       -Wno-error=unused-but-set-variable
 
 AM_LDFLAGS = \
        $(CODE_COVERAGE_LIBS)
diff --git a/src/common/libflux/Makefile.am b/src/common/libflux/Makefile.am
index 1e184ca..c3be6ea 100644
--- a/src/common/libflux/Makefile.am
+++ b/src/common/libflux/Makefile.am
@@ -1,6 +1,7 @@
 AM_CFLAGS = \
        $(WARNING_CFLAGS) \
-       $(CODE_COVERAGE_CFLAGS)
+       $(CODE_COVERAGE_CFLAGS) \
+       -Wno-error=unused-but-set-variable
 
 AM_LDFLAGS = \
        $(CODE_COVERAGE_LIBS)
[...]

PR #5084, opened 2023-04-10
libsubprocess: remove use of assert(0)

chu11:
I noted in #5080 how there were some warnings/errors due to `assert(0)` usage in `libsubprocess`.  In general, shouldn't use `assert(0)`.  Should assert to check for validity of expected things vs doing a `assert(0)` to "error out" on unexpected things.  In one case it was in an user facing API function, which is bad.

Due to my recently created issue #5083 also added a note about how one function doesn't necessarily do the expected thing at the moment.


garlick (approved):
LGTM!

PR #5085, opened 2023-04-11
broker: clean up module infrastructure

garlick:
This cleans up some old code in the broker for loading broker modules, and drops terminology (and code) for "submodules" which are not used.

Two module loading-related functions are dropped from the public API.
```c
char *flux_modname (const char *filename, flux_moderr_f *cb, void *arg);

char *flux_modfind (const char *searchpath, const char *modname,
                    flux_moderr_f *cb, void *arg);
```

Unnecessary fields are dropped from the output of `flux module list` (Digest and Size), which now looks like this:
```
$ flux module list
Module                   Idle  S Service
job-list                    7  R 
job-info                    7  R 
resource                    6  R 
[snip]
```

To work with the simplified module search, broker module DSOs now land in one directory in-tree (`src/modules/.lib`).

WIP pending

- [x] more tests
- [x] editing/culling of [rfc5](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_5.html)
- [x]  fix fluxion "latestart" test that depends on `flux module list` output


garlick:
I pushed a change to add `flux module list -l,--long` which shows the full dso path.  Good idea, and I guess the removed size and digest would have provided clues when debugging such things, but less directly and less conveniently than the path.

Pondering the dynamic naming issue:   I think we could easily pass a name override along with the DSO name/path in `broker.insmod` and then allow the same module to be loaded multiple times under different names.   Some gotchas I can think of are
- `broker.rmmod` now allows a module to be unloaded by path so that "reload" can be implemented as `broker.rmmod` + `broker.insmod` with a path argument.
- the module needs to have a way to find out its name so it can register message handlers under the correct service

I'll see if maybe I can address the first issue in this PR by returning the name in the insmod response and then requiring that the name be used to unload.  I think further development on dynamic naming should be in a separate PR, and should probably be preceded by further discussion on the use case since there may be other, better ways to approach "module loading modules" (like maybe making the module loader a flux standard plugin?)

p.s. this was a little weekend cleanup diversion, not necessarily the beginning of a series in this area.

garlick:
Well it wasn't too hard to add the ability to load a module multiple times under different names and that seems like it could have some nice benefits as suggested by @grondo, so I went ahead:
```
$ flux module load --name=smurf kvs
$ flux module load --name=smurfette kvs
$ flux module list -l | grep kvs
$ flux module list -l |grep kvs.so
kvs                        20  R          /home/garlick/proj/flux-core/src/modules/flux/.libs/kvs.so
smurfette                  12  R          /home/garlick/proj/flux-core/src/modules/flux/.libs/kvs.so
smurf                      16  R          /home/garlick/proj/flux-core/src/modules/flux/.libs/kvs.so
```
Possibly this should be split off to another pr...

garlick:
I think this is ready for a review.   LMK if you think this ought to be split into multiple PRs.  The main areas of change are:
- build system:  link modules in one directory, `src/modules/.libs`, and modfiy the in tree search path to look there (only).
- command changes: `flux module info` is dropped, `flux module (re-)load` now takes a `--name` override, `flux module list` *Digest* and *Size* columns removed, *Path* column added (if `-l,--long`).
- API changes: dropped `flux_modfind()` and `flux_modname()` functions
- Broker internal: determine module name from dso basename not `mod_name` symbol.  Drop "submodule" support.
- Allow the same dso to be loaded multiple times under different names.

t/t0003-module.t
```diff
@@ -48,37 +48,6 @@ test_expect_success 'module: lsmod does not show test module' '
 	! flux module list | grep parent
 '
 
-test_expect_success 'module API: load test module' '
```
grondo, line 51: commit message typo: "some tests is" --> "some tests in"

src/broker/module.c
```diff
@@ -599,16 +600,23 @@ module_t *module_add (modhash_t *mh,
     }
     if (!(p->path = strdup (path))
         || !(p->rmmod = zlist_new ())
-        || !(p->subs = zlist_new ())
-        || !(p->name_buf = strdup (path))
-        || !(p->name = module_name_from_path (p->name_buf)))
+        || !(p->subs = zlist_new ()))
         goto nomem;
+    if (name) {
+        if (!(p->name = strdup (name)))
```
grondo, line 606: Should this be
```c
    if (!(p->name_buf = strdup (name)))
        goto nomem;
    p->name = p->name_buf;
```

To avoid leaking `p->name`? At least I don't see where `p->name` is ever freed.

grondo (approved):
I've done lots of testing on this branch and everything works great. Had only a couple comments on a pass through the code. Looks good!

I think I'm fine with merging the PR as is.

src/broker/module.c
garlick: Oooh, good catch!  In the final code, p->name might be a pointer into p->name_buf or might be a new allocation. So this is a leak.  WIll fix!

garlick: Fixed and force pushed.

grondo:
Looks good. Already approved so feel free to set MWP when you're ready as far as I'm concerned.

PR #5088, opened 2023-04-12
job-exec: improve error message when job shell/imp execution fails

grondo:
This PR modifies the error raised by the job-exec module when the job-shell or IMP failed to execute, which is currently a bit confusing, e.g.:
```
job shell exec error on broker pi3 (rank 0): No such file or directory
```
The new error just adds the path to the failing command, which should make it obvious what went wrong in this case.

When testing, I noticed that `flux job attach` would hang in this scenario (ctr-c works to terminate it though) because of a stray timer that was left on in the reactor, so that is fixed here as well.

For some reason, the test does not work in CentOS 7. Instead of getting an error from exec(2), execution of the job shell from a bad path returns success on this distro, then later exits with exit code 127. I didn't want to spend all day trying to get to the bottom that, since that distro is dwindling in usage. I assume it probably has something to do with limited `posix_spawn` support in that glibc(wild guess). So I've just disabled the test there for now.

I thought I had opened an issue on improving this error message, but I can't find it at the moment.

### DIFF
--- a/src/cmd/flux-job.c
+++ b/src/cmd/flux-job.c
@@ -1628,6 +1628,7 @@ void attach_completed_check (struct attach_ctx *ctx)
         flux_watcher_stop (ctx->sigint_w);
         flux_watcher_stop (ctx->sigtstp_w);
         flux_watcher_stop (ctx->stdin_w);
+        flux_watcher_stop (ctx->notify_timer);
     }
 }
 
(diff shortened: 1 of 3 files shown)


garlick (approved):
Looks like a good improvement.

grondo:
Thanks! I've set MWP.

PR #5090, opened 2023-04-12
testsuite: skip failing test on RHEL7

### LINKED ISSUE #5089: testsuite: not ok 18 - job-exec: path to shell is emitted on exec error
This test is skipped on CENTOS 7, but not surprisingly also fails on RHEL7.

chu11:
Problem: A test in t2400-job-exec-test.t fails on RHEL7.  It is skipped on CENTOS7 and should be skipped on RHEL7 as well.

Skip the test on both RHEL7 and CENTOS7.

Fixes #5089

----

Tested on a RHEL7 and RHEL8 box, hopefully there isn't a corner case in there (that won't be picked up via CI).

### DIFF
--- a/t/t2400-job-exec-test.t
+++ b/t/t2400-job-exec-test.t
@@ -183,13 +183,17 @@ test_expect_success 'job-exec: critical-ranks RPC handles unexpected input' '
             test_must_fail flux python critical-ranks.py $id 0
         )
 '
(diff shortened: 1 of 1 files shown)


garlick (approved):
Looks alright to me!

grondo:
Thanks @chu11!

Issue #5091, opened 2023-04-13
testsuite: do not assume "/usr/bin/env python3" equates to "flux python"


chu11:
Recently hit an issue in the testsuite on corona getting a lot of errors like:

```
Traceback (most recent call last):
  File "/g/g0/achu/chaos/git/flux-framework/flux-core/t/scripts/dmesg-grep.py", line 18, in <module>
    import flux
  File "/g/g0/achu/chaos/git/flux-framework/flux-core/src/bindings/python/flux/__init__.py", line 14, in <module>
    import flux.core.handle
  File "/g/g0/achu/chaos/git/flux-framework/flux-core/src/bindings/python/flux/core/handle.py", line 15, in <module>
    from _flux._core import ffi, lib
ModuleNotFoundError: No module named '_cffi_backend'
```

The issue was that I built / configured `flux python` to be Python 3.6 (i.e. `PYTHON_VERSION=3.6` during configuration).  By some dumb chance, a `python3.7` existed in `/usr/local/bin/python3` which was earlier in my path than `/usr/bin/python3`.  So scripts that did `/usr/bin/env python3` were picking up a different version of python and leading to errors.

I think instead of `/usr/bin/env python3` we should probably be doing `/usr/bin/env -S flux python`.  Not for installed scripts, but atleast within the python scripts in the testsuite.

chu11:
As an aside, in another chat @grondo suggested doing something like:

```
PATH=$(flux python -c 'import os,sys; print(os.path.dirname(sys.executable))'):$PATH
```

in `t/sharness.d/01-setup.sh`.  I gave it a shot but suddenly realized, when `01-setup.sh` is sourced, a test flux instance may not yet be setup, so `flux python` will lead to an error.  There's ways to get around this, but figured just doing `/usr/bin/env -S flux python` might be the safer / correct route.

Or perhaps there's another better way.

grondo:
The PATH is set to the correct flux executable first (otherwise flux start wouldn't work). I tried my suggestion and it worked for the conda build.

One problem I think we ran into with the other solution is that many times the shebang only supports at most two arguments (iirc). If you search back in old issues you might find a very long thread about this problem...

chu11:
Ahhh I think I just had to be slightly less stupid:

```Python
diff --git a/t/sharness.d/01-setup.sh b/t/sharness.d/01-setup.sh
index 84b02f97f..05e1b2630 100644
--- a/t/sharness.d/01-setup.sh
[...]

PR #5093, opened 2023-04-13
build: applicable build and test fixes for conda

grondo:
This PR includes a set of fixes that were necessary under `conda`, excluding the `fwrite` changes required from `attribute_unused` in the CentOS 6 sysroot.

I figured most of these will work around future problems we may hit on other one-off platforms.

### DIFF
--- a/t/sharness.d/01-setup.sh
+++ b/t/sharness.d/01-setup.sh
@@ -47,6 +47,15 @@ if test -n "$FLUX_TEST_INSTALLED_PATH"; then
     PATH=$FLUX_TEST_INSTALLED_PATH:$PATH
     fluxbin=$FLUX_TEST_INSTALLED_PATH/flux
 else # normal case, use ${top_builddir}/src/cmd/flux
+    #
+    #  Ensure that the path to the configured Python is first in PATH
+    #  so the correct version is found by '#!/usr/bin/env python3' in
+    #  several test scripts that use this shebang line.
+    #  N.B.: This is not a complete fix. See flux-core #5091 for details
+    #
+    PATH=$($FLUX_BUILD_DIR/src/cmd/flux python -c \
+           'import os,sys; print(os.path.dirname(sys.executable))'):$PATH
+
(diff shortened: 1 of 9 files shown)


t/sharness.d/01-setup.sh
```diff
@@ -47,6 +47,8 @@ if test -n "$FLUX_TEST_INSTALLED_PATH"; then
     PATH=$FLUX_TEST_INSTALLED_PATH:$PATH
     fluxbin=$FLUX_TEST_INSTALLED_PATH/flux
 else # normal case, use ${top_builddir}/src/cmd/flux
+    PATH=$($FLUX_BUILD_DIR/src/cmd/flux python -c \
+           'import os,sys; print(os.path.dirname(sys.executable))'):$PATH
```
chu11, line 51: perhaps note some of the corner cases I noted in #5091 that this does not fix?  Or perhaps just point to the issue?

chu11 (approved):
LGTM, just one comment below

t/sharness.d/01-setup.sh
grondo: I'll do a little testing and push another commit here if it works out ok.

grondo: Ok, force-pushed this PR branch with an added test in `sharness.d/01-setup.sh` to ensure `flux python` matches `python3`

chu11: :+1:

grondo: Hm, well maybe that didn't work in CI. I'll check it out.

grondo: Ah, in one builder we have two versions of Python installed that are "compatible enough" to both work for `make check`. I'll revert this extra check in favor of a descriptive comment.

grondo: Ok, force-pushed removal of that  failed experiment and added a verbose comment to 01-setup.sh.

chu11: comment looks good to me!

Issue #5094, opened 2023-04-14
job-exec: segfault in libsubprocess remote_output()


grondo:
The broker crashed on fluke yesterday evening with the backtrace:
```gdb
(gdb) bt
#0  0x00007ffff62aaaff in raise () from /lib64/libc.so.6
#1  0x00007ffff627dea5 in abort () from /lib64/libc.so.6
#2  0x00007ffff627dd79 in __assert_fail_base.cold.0 () from /lib64/libc.so.6
#3  0x00007ffff62a3456 in __assert_fail () from /lib64/libc.so.6
#4  0x00007fffc9107823 in fzhash_lookup (self=<optimized out>, 
    key=<optimized out>) at zhash.c:316
#5  0x00007fffc90dc9b9 in remote_output (eof=false, len=80, 
    data=<optimized out>, stream=0x7fff64201b70 "stderr", p=0x7fff640282a0)
    at remote.c:514
#6  rexec_continuation (f=0x7fff64086960, arg=0x7fff640282a0) at remote.c:587
#7  0x00007ffff7b8ea13 in ev_invoke_pending (loop=0x7fff64010640) at ev.c:3770
#8  0x00007ffff7b92ab8 in ev_run (flags=0, loop=0x7fff64010640) at ev.c:4190
#9  ev_run (loop=0x7fff64010640, flags=0) at ev.c:4021
#10 0x00007ffff7b5a12f in flux_reactor_run (r=0x7fff64010620, 
    flags=flags@entry=0) at reactor.c:128
#11 0x00007fffc90d17c8 in mod_main (h=0x7fff64006030, argc=<optimized out>, 
    argv=0x7fff64011410) at job-exec.c:1434
#12 0x00000000004131d1 in module_thread (arg=0x1b58820) at module.c:183
#13 0x00007ffff79221ca in start_thread () from /lib64/libpthread.so.0
#14 0x00007ffff6295e73 in clone () from /lib64/libc.so.6
```

The problem is that `p->channels` is `NULL` in `remote_output()`:

```gdb
(gdb) down
#5  0x00007fffc90dc9b9 in remote_output (eof=false, len=80, 
    data=<optimized out>, stream=0x7fff64201b70 "stderr", p=0x7fff640282a0)
    at remote.c:514
514	in remote.c
(gdb) p *p->channels
Cannot access memory at address 0x0
```

[...]

PR #5095, opened 2023-04-14
libsubprocess: avoid assertion failure in remote_output()

grondo:
Problem: A broker crash was recently observed in libsubprcess remote_output() from an assertion failure in zhash_lookup() due to a NULL target hash.

Diagnosis of the underlying cause is still in progress, but for now avoid the abort by checking that the subprocess has a valid channels hash (and that stream is not NULL) before going on to process the output. Log an error if the condition is detected.

This is a w/a for issue #5094 until we can get to the root cause. I plan to roll this into a 0.49.0-2 RPM if acceptable.

### DIFF
--- a/src/common/libsubprocess/remote.c
+++ b/src/common/libsubprocess/remote.c
@@ -511,6 +511,15 @@ static int remote_output (flux_subprocess_t *p,
 {
     struct subprocess_channel *c;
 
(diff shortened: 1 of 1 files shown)


grondo:
@garlick actually found the underlying cause so I think this PR is now unnecessary.

PR #5101, opened 2023-04-16
build: add `make deb` target to build test debian package

grondo:
This PR adds minimal debian metadata to the `debian` directory along with a `scripts/debbuild.sh` script and `make deb` target for the top level `Makefile`. Packages created in this way are meant for easier installation of flux-core on test systems.

Note that the package depends on an installed flux-security package (There's a `make deb` target there now too).

By default, tests are not run with `make deb`. To enable `make check` during packaging, override the default `DEB_BUILD_OPTIONS` with `DEB_BUILD_OPTIONS=terse make -j4 deb` for example.


garlick (approved):
I've converted my test cluster to use this with the flux-security deb and all is well!  Thanks, big improvement over make install to /usr/local + rsync.

grondo:
Great, thanks!

PR #5102, opened 2023-04-17
testsuite: fix setup error in system tests

chu11:
Problem: In t9000-system.t the FLUX_TEST_INSTALLED_PATH variable is setup if 'debug' is set.  However, the check for 'debug' is done before sharness is sourced, so the `-d` option when running the tests doesn't matter.  This lead to the t9000-system.t tests failing when running make check in a docker container.

Solution: Instead of checking for 'debug' to set FLUX_TEST_INSTALLED_PATH, check for FLUX_ENABLE_SYSTEM_TESTS instead.

### DIFF
--- a/t/t9000-system.t
+++ b/t/t9000-system.t
@@ -3,9 +3,12 @@
 
 test_description='Run tests against a system instance of Flux'
 
-#  Allow this test to be forced to run if debug is set _and_ a flux binary
-#  exists in the FLUX_TEST_INSTALLED_PATH or /usr/bin:
-if test -n "$debug"; then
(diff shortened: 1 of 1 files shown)


chu11:
re-pushed, noticed a typo :-)

grondo (approved):
Great! This should have been how the previous fix was done :facepalm:.

One commit message nit, there's a double use of "Instead" in the Solution statement :-)

chu11:
@grondo thanks, fixed up that typo in the commit message, rebased, and re-pushed.  I'll set MWP.

PR #5019, opened 2023-03-27
Fix minor typos and formatting

chu11:
Just like #4934, the PR for minor typos so we don't have a ton of them littered in every release.

PR #5106, opened 2023-04-21
Add false positives typos config

chu11:
Problem: The current typos workflow check only checks a select number of files because there are so many false positives in flux-core.

Solution: Add a _typos.toml configuration file to eliminate false positives so we can run the typo check on all of flux-core.  The config file either skips files with lots of false positives or lists common false positives we wish to ignore.

----

This works when I run by hand in my build dir, we'll see if this works in CI :shrug: 

Note, this will probably fail typo checking until #5019 is merged

Edit:

yup, seems to work.  These are the ones I have in a fix in #5019

image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,8 +19,6 @@ jobs:
     - uses: actions/checkout@v3
     - name: Check Spelling
       uses: crate-ci/typos@7ad296c72fa8265059cc03d1eda562fbdfcd6df2 # v1.9.0
-      with:
-        files: ./NEWS.md ./README.md ./doc/*.rst ./doc/*/*.rst ./doc/*/*/*.rst ./src/cmd
 
   python-lint:
     name: python linting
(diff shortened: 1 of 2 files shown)


grondo:
Can we do `.typos.toml` instead of `_typos.toml`? Maybe I just don't understand the point of the leading underscore?

chu11:
> Can we do .typos.toml instead of _typos.toml? Maybe I just don't understand the point of the leading underscore?

Yeah, it's stupid.  It's unfortunately what the typo checking advertises for configuration.  I was thinking maybe I need to send a PR to support the period version.

But it appears `.typos.toml` is supported in their code!  Just not documented publicly.

grondo:
Ah, yeah I was about to say it is documented in [the reference](https://github.com/crate-ci/typos/blob/master/docs/reference.md)

chu11:
Ahhh, I only looked at the README.

chu11:
rebased and re-pushed with the typos config renamed

chu11:
removing WIP, this should pass now (hopefully it does!)

grondo (approved):
Nice!

Issue #5108, opened 2023-04-22
broker crash when sched-simple fails to unregister


grondo:
Not sure what happened here, but @vsoch was able to trigger a broker crash by using an older `rc1` with a newer flux:
```
node-1  | broker.info[0]: start: none->join 1.35219ms
node-1  | broker.info[0]: parent-none: join->init 0.017271ms
node-1  | cron.info[0]: synchronizing cron tasks to event heartbeat.pulse
node-1  | job-manager.info[0]: restart: 0 jobs
node-1  | job-manager.info[0]: restart: 0 running jobs
node-1  | job-manager.info[0]: restart: checkpoint.job-manager not found
node-1  | broker.info[0]: rc1.0: running /etc/flux/rc1.d/01-sched-fluxion
node-1  | sched-fluxion-resource.info[0]: version 0.27.0-11-g214aa271
node-1  | sched-fluxion-resource.warning[0]: create_reader: allowlist unsupported
node-1  | sched-fluxion-resource.info[0]: populate_resource_db: loaded resources from core's resource.acquire
node-1  | sched-fluxion-qmanager.info[0]: version 0.27.0-11-g214aa271
node-1  | broker.info[0]: rc1.0: running /etc/flux/rc1.d/02-cron
node-1  | sched-simple.err[0]: service_unregister: Invalid argument
node-1  | sched-simple.err[0]: service_unregister: Invalid argument
node-1  | flux-broker: zlistx.c:234: fzlistx_first: Assertion `self' failed.
node-1  | /home/fluxuser/entrypoint.sh: line 64:    20 Aborted                 (core dumped) FLUX_FAKE_HOSTNAME=$thisHost flux start -o --config /etc/flux/config ${brokerOptions} sleep inf
```

The older `rc1` explains why `sched-simple` was loaded (`flux module list` output changed), but will have to inspect code to see how we got to the `service_unregister` errors and then a crash. (Simply attempting to load `sched-simple` with fluxion modules loaded doesn't reproduce the issue)

vsoch:
You can count on me to come up with strange and mysterious errors! 😆

I was wondering if, given some customization that is needed (and saving an rc file separately from the repository as I have done) there might be a way to separate that logic, so I always get the latest rc shipped with flux but can still add my customization (and don’t run into an issue like this that results from saving and using an older version).

Thanks for posting this!

grondo:
[...]

PR #5109, opened 2023-04-22
sched-simple: avoid assertion failure when trying to load scheduler twice

garlick:
This implements @grondo's fix proposed in #5108, and removes unnecessary service de-registration code from libschedutil that was adding to the log noise when this problem was encountered.

### DIFF
--- a/src/modules/sched-simple/sched.c
+++ b/src/modules/sched-simple/sched.c
@@ -141,21 +141,25 @@ err:
 
 static void simple_sched_destroy (flux_t *h, struct simple_sched *ss)
 {
-    struct jobreq *job = zlistx_first (ss->queue);
-    while (job) {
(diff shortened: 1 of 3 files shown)


src/modules/sched-simple/sched.c
```diff
@@ -141,21 +141,25 @@ jobreq_create (const flux_msg_t *msg)
 
 static void simple_sched_destroy (flux_t *h, struct simple_sched *ss)
 {
```
chu11, line 143: commit message typo: "retore"

chu11 (approved):
LGTM, just one nit

src/modules/sched-simple/sched.c
garlick: Thanks!  Fixed that and forced a push.  I'll set MWP.

Issue #5116, opened 2023-04-25
Job validator reports that no plugins are found when one plugin fails to import


ardangelo:
On one of our test systems, we did not have the `jsonschema` Pip package installed,
This was causing the job validator `schema` plugin to throw an exception when imported.
However, this manifested as the job validator `flux job-validator --list-plugins` reporting that it was unable to load *any* plugins, instead of pointing to `schema` as the failing plugin.

grondo:
Thanks for finding this issue and running down the root cause. Clearly we're catching an `ImportError` at the wrong spot here (though I'm no Python expert). Now that you've figured out the reproducer, I'm sure we'll be able to fix it.

PR #5124, opened 2023-04-27
job-validator: fix empty plugins list when one plugin fails to import

grondo:
This PR fixes an issue where `flux job-validator` reports that no plugins are available when importing any one plugin failed with an `ImportError`. This occurred because a `ModuleNotFoundError` exception was handled at the wrong spot and treated as an empty plugins directory.

Now, any import failure on a plugin will result in an exception being raised to the caller with a sensible error message.

e.g.: in the case reported in #5116 we see this
```console
$ flux job-validator --list-plugins
Available plugins:
```

Now we'll get this:
```console
$ src/cmd/flux job-validator --list-plugins
flux-job-validator: ERROR: failed to import schema plugin: No module named 'jsonschema'
```

### DIFF
--- a/src/bindings/python/flux/importer.py
+++ b/src/bindings/python/flux/importer.py
@@ -36,10 +36,16 @@ def import_plugins(pkg_name, pluginpath=None):
 
     try:
         #  Load 'pkg_name' as a namespace plugin
+        #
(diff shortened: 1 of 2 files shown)


garlick (approved):
LGTM!

PR #5125, opened 2023-04-27
python: fix parsing of special characters in submission directives

grondo:
A user noticed that some batch directives unexpectedly required quoting, e.g.
```console
$ cat test.sh
#!/bin/sh
#flux: --output=test.{{id}}.out
flux run -n8 hostname
$ flux python -m flux.job.directives  test.sh
SETARGS(['--output=test.', '{', '{', 'id', '}', '}', '.out'])
```

On further reading of the `shlex` python documentation, the `whitespace_split` attribute should be set to `True` for "the best shell lexing compatibility", This seems to fix the problem:

```console
$ src/cmd/flux python -m flux.job.directives test.sh
SETARGS(['--output=test.{{id}}.out'])
```

This PR makes that change and adds missing tests that exercise the issue.

### DIFF
--- a/src/bindings/python/flux/job/directives.py
+++ b/src/bindings/python/flux/job/directives.py
@@ -49,6 +49,11 @@ class Directive:
         # quoting and avoid splitting on cmdline args like --foo.
         lexer = shlex.shlex(value, posix=True, punctuation_chars=True)
         #
+        # set whitespace_split to match shell parsing of cmdlines as
+        # closely as possible as documented in the final note here:
+        # https://docs.python.org/3/library/shlex.html
+        lexer.whitespace_split = True
+        #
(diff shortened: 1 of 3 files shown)


src/bindings/python/flux/job/directives.py
```diff
@@ -49,6 +49,11 @@ def __init__(self, value, lineno=-1):
         # quoting and avoid splitting on cmdline args like --foo.
```
chu11, line 49: cut and paste typo "negatel"

chu11 (approved):
LGTM, just that one typo I found

src/bindings/python/flux/job/directives.py
grondo: Thanks! Fixed that and pushed the result.

Issue #4914, opened 2023-02-03
idea: augment job-list "filter" with RFC 31 constraint support


grondo:
Problem: `job-list` is currently limited in the kinds of general "queries" it can support (e.g. it only supports a single userid, limited set of "filters" etc.).

Without re-implementing `job-list` as a database (which still may be the correct long term solution), another idea would be to support a version of the Constraint syntax proposed in [RFC 31](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_31.html). The benefit of this approach is that the implementation is fairly simple but powerful, and we already have a nice string syntax proposed to build Constraint JSON objects easily ([RFC 35](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_35.html)), along with a Python parser which could go into `flux jobs` or `flux pgrep`.

Some prevalent use cases include:
 - list the jobs for > 1 user (`user:bob,alice`)
 - list all jobs that ran or are running on a given node (`host:fluke16`)
 - list all jobs that ran on a bad node since monday morning and failed (`host:fluke16 since:"Mon 8am" is:failed`)
 - list jobs today that failed but were not canceled or timed out (`is:failed not (is:canceled or is:timeout)`)

etc.

The drawback to this approach is more work for the `job-list` module. It is possible that a user could construct a complex query which would act as a DoS for the service.

chu11:
Given discussion in #4908.  could RFC31 support sorting?  (I don't think RFC35 does at the moment, but that could be updated as a result).  I'm imagining something like: `{ "sort-ascending": [ { "properties": [ "t_submit" ] } ] }`.

grondo:
I think it should be a separate option, the way the objects in RFC 31 currently is they apply like a filter, so the constraint is applied to each resource, job, object, etc. A `sort` operation doesn't make sense in that context. However, I'm open to ideas on how that could work.

We could instead add a wrapper around an RFC 31 constraint object that applies other operations to the query, e.g. a sort key or keys, the count of values to return, etc.?

chu11:
curious what folks think, if `job-list` is updated to use support job constraints, should the old filtering mechanisms (and thus RPC format) be supported for backwards compatibility?

Due to how `job-list` currently operates and the refactoring that would be needed, it'd be nice to remove backwards compatibility.  I imagine user facing tools like `flux-jobs` can simply convert the current user options into constraints.

chu11:
[...]

Issue #4186, opened 2022-03-03
filter 'flux jobs' output by Node ID


ryanday36:
Related to #4185, it would be useful to be able to filter the output of `flux jobs` by node ID. I.e. to be able to easily answer questions like who has recently run on node 'fooX'? or how much time is left on the job(s) running on 'fooX'?

Raj-Bagri1:
This is to help us at ops find trends. I.E, 20+ nodes suddenly crashed or OOM'd and they were all running a single or individual jobs by the same user. 
Being able to tell when a job ends will also help us determine when physical work can be initiated on a node so that we may plan accordingly. Especially on time sensitive work.

grondo:
There is a huge opportunity for enhanced tools here since the full "resource set" for every job that ran will be available in the KVS or the job archive database. If there is equivalent functionality you use now in existing resources manager it would be very helpful to delineate some specific scenarios in this issue. Thanks!

grondo:
Here's something I just hacked together that could be the start of a solution here:

```python
#!/usr/bin/python3

import sys
from datetime import datetime

import flux
from flux.job import job_list_inactive
from flux.hostlist import Hostlist
from flux.util import parse_datetime

host = sys.argv[1]
try:
    since = parse_datetime(sys.argv[2]).timestamp()
except:
    since = 0.0

date = datetime.fromtimestamp(since)
print(f"Looking for jobs on {host} since {date:%D %T}")

jobs = job_list_inactive(
    flux.Flux(),
    since=since,
    attrs=["ranks", "nodelist"],
    max_entries=0,
).get_jobinfos()

for job in jobs:
    if host in Hostlist(job.nodelist):
        print(job.id, job.nodelist)

```


e.g. get all the jobs that ran on fluke100 in the last 7 days:

```
 grondo@fluke108:~$ flux python jobs-by-node.py fluke100 -7d
Looking for jobs on fluke100 since 03/06/22 15:21:51
ƒPUpdfzsBNw fluke[6-16,18-23,25-56,58,60,62-65,67-78,80-91,93-103]
ƒPUJoPtLuyR fluke[60,62-65,67-78,80-91,93-103]
ƒPUJ2cxCxBy fluke[60,62-65,67-78,80-91,93-103]
ƒPUHHiqGH4T fluke[60,62-65,67-78,80-91,93-103]
ƒPKG3VMJens fluke[10-16,18-23,25-50,100]
ƒP3mnZZMoPm fluke100
[...]

PR #5126, opened 2023-04-27
WIP: job-list: support job list constraints

chu11:
This is a super early WIP, where I'm just building up the matching library to use within `job-list`, so it isn't used at all for actual queries yet.

The main reason I'm posting is I wanted to highlight these functions in `match.[ch]` and see if anyone sees a problem with it or if its overly dumb with what I want to do.

```
/*  Identical to list_constraint_create() but only cares about
 *  "states" operation.  Effectively tests the 'states' constraint
 *  operation and conditionals with 'states', everything else is
 *  considered true all of the time.
 */
struct list_constraint *state_constraint_create (json_t *constraint,
                                                 flux_error_t *errp);

/* determines if a job in 'state' could potentially return true with
 * the given constraint.
 */
bool state_match (flux_job_state_t state, struct list_constraint *constraint);
```

Basically, given some constraint that is passed in, I'd like to know if it is possible / impossible for any jobs with a specific job state to match it.  e.g. the constraint `states:pending userid:42 queue:batch` returns true if the job state is pending and it doesn't matter what the queue/userid are.  There could be no jobs for userid 42 in `job-list` at all and it will return true.

This is so we can skip iterating and filtering on possibly long lists of pending/running/inactive jobs.  (Note, I also need to add an equivalent one to these for `since` matching to end iterating on the inactive list).

To implement this, this special constraint treats everything that isn't a `states` constraint as true all of the time.  So we are only testing just the `states` constraint and nothing else.  Some special handling of conditionals has to be done too. e.g. `userid:42` and `-userid:42` both need to return true, because we don't care about `userid:42`.

Here's the core code as an example of what I'm doing in this function (removing some error checking to make this example simpler).

```
    json_object_foreach (constraint, op, values) {
        if (streq (op, "userid")
            || streq (op, "name")
            || streq (op, "queue")
            || streq (op, "results")
            || streq (op, "since"))
            return list_constraint_new (errp); // this constraint always returns true
        else if (streq (op, "states"))
            return create_states_constraint (values, errp);
        else if (streq (op, "or") || streq (op, "and") || streq (op, "not"))
            return conditional_constraint (op,
                                           match_or_checkempty,
                                           match_not_checkempty,
                                           state_constraint_create,
                                           values,
                                           errp);
    }
```

As a complete aside as this might come up as a question, "instead of having a new function state_constraint_create() could you just use the normal constraint from before?", and the answer is yes.  That was my round 1 prototype, but I found the code to be too complex / irritating so I shifted to this one.


chu11:
> /*  Identical to list_constraint_create() but only cares about
>  *  "states" operation.  Effectively tests the 'states' constraint
>  *  operation and conditionals with 'states', everything else is
>  *  considered true all of the time.
> */

It doesn't look like my simplistic approach will work, which can be illustrated by this simple example:

`states=running | userid=42`

In my `states_match()` function, this constraint should return true for pending, running, and inactive job states, b/c the userid=42 is the only thing that matters. i.e. userid=42 can be on the pending, running, or inactive lists.  So we have to scan all the lists.

`not (states=running | userid=42)`

This should return true for the pending and inactive job states, as its impossible for a running job (regardless of userid) to pass this constraint match.

I'm going to give it the good 'ol college try on another technique, which I think will basically involve a third potential state of "maybe" a job in a certain state could pass, but shouldn't spend forever on this.  Perhaps an interim solution could be to support job constraints on all the non-job states stuff only (and given discussion in #4914  maybe `since` too).

chu11:
Re-pushed, pushing a first round implementation of constraints based filtering on `job-list.  I effectively implement all the prior filter options into constraints logic.  There is only minor enhancements to this logic, like you can specify multiple userids (i.e. `userid:100,101,102`) which was not possible before.

In follow up work, this will be enhanced:

- the `since` operation should be dropped for a `t_inactive` one, and `t_inactive` should support `<` and `>` (and other timestamps too probably) (NOTE, this could potentially be done in this PR, went back and forth on it but like it to be a follow up PR since supporting `<` and `>` and potentially many timestamps can be tricky and new.)

- I did not add an advanced filter to `flux-jobs`, which could be convenient.  That can be for a follow up PR.  i.e.  `flux jobs --query="userid:1234 name:foo states:pending` per RFC35.

- we will eventually want to add hostlist filtering per #4186

- how this affects inactive jobs in a db in #4336, TBD

- could update #4915 to use the constraints logic (well perhaps need a few of above to be finished first)

- I didn't do this, but pondering if I should.  Should I parse old RPC fields and sent a EPROTO error back if a caller tried to use the old RPC format?  There was some code like the following in the testsuite:

```
  $jq -j -c -n  "{max_entries:1, userid:${id}, states:0, results:0, attrs:[\"all\"]}" \
    | $RPC job-list.list | jq ".jobs[0]" > all_success.out &&
```

It initially shocked me that this worked, only to realize, this basically told `job-list.list` to return all jobs, because it didn't specify a constraint object.  Alternately, a constraint object could be required by `job-list.list` instead of it being optional.

A side note:

- the constraints I create in `src/bindings/python/flux/job.py` are to pass what we currently do in `flux-jobs`.  Issue #5149 is not solved in this.  So I may end up changing the constraint.  In fact, I think this is all it would take to solve #5149.

```
--- a/src/bindings/python/flux/job/list.py
+++ b/src/bindings/python/flux/job/list.py
@@ -56,8 +56,6 @@ def job_list(
     if since:
         constraint["and"].append({ "since": [ since ] })
     if states and results:
-        if states & flux.constants.FLUX_JOB_STATE_INACTIVE:
-            states &= (~flux.constants.FLUX_JOB_STATE_INACTIVE)
         tmp = { "or": [] }
         tmp["or"].append({ "states": [ states ] })
         tmp["or"].append({ "results": [ results ] })
```

chu11:
re-pushed with a small refactor, a "oh, that's still programmed that way because of ... before."

src/common/libjob/list.c
```diff
@@ -59,23 +67,30 @@ flux_future_t *flux_job_list_inactive (flux_t *h,
...
-                             "{s:i s:f s:i s:i s:i s:o}",
+                             "{s:i s:o s:o}",
                              "max_entries", max_entries,
-                             "since", since,
-                             "userid", FLUX_USERID_UNKNOWN,
-                             "states", FLUX_JOB_STATE_INACTIVE,
-                             "results", 0,
-                             "attrs", o))) {
+                             "attrs", o,
+                             "constraints", c))) {
```
grondo, line 90: It looks like there's a mix of `constraint` vs `constraints` throughout this file. What is the difference? It wasn't obvious reading the diff anyway...

grondo (commented):
Just a first pass through.. I'd like to do some further testing on another pass.

One general note: I would have perhaps kept the `since` parameter separate in the protocol. The reason for that is it was meant as a way to stop processing the job lists before you get all the way through, and it isn't clear that the constraint based solution allows that.

Also, eventually I would think `since` would be implemented in the protocol as a time comparison, i.e. something like `t_inactive >= timestamp`. Eh, this all seems fine actually, I just thought I'd bring it up for further discussion.

src/common/libjob/list.c
chu11: That seems to just be an inconsistent I picked somewhat at random.  I'll correct.

chu11:
ya know, maybe `since` should be kept separate outside of constraints.  We have a specific use case for it (job archiving) that does improve performance with it.  It could be kept even after timestamp support is added.  Will think about it.

grondo:
> ya know, maybe since should be kept separate outside of constraints. 

Yeah, that's what I was trying to suggest above :-)

src/common/libjob/list.c
chu11: whew, good catch, at first I was like "how do the tests even work?" only to realize "oh yeah, we deprecated these functions", so I guess its not really tested at all.

chu11:
Re-pushed with the following changes:

- tweaked issues per comments above (notably virtual state abbreviations)
- re-added the "since" `job-list` RPC parameter, we will keep that one outside of constraints
  - and removed "since" constraint support
- added timestamp `job-list` matching constraints, timestamp being `t_run`, `t_inactive`, etc.
  - in hindsight, this didn't have to be added in this PR, as the re-addition of the `since` parameter keeps that filtering the same was before.  But since I programmed it, it's there :-)

src/modules/job-list/match.c
```diff
@@ -0,0 +1,654 @@
...
+            tv = timestamp_value_create_str (str + 1,
+                                             type,
+                                             MATCH_GREATER_THAN,
+                                             errp);
+        else if (strstarts (str, "<"))
+            tv = timestamp_value_create_str (str + 1,
+                                             type,
+                                             MATCH_LESS_THAN,
+                                             errp);
+        else /* if no operation specified >= is default */
```
grondo, line 447: I realize there isn't much point in supporting "=" for timestamps, but it feels weird that `>=` is the default here.
Maybe just support '=' and have that be the default, knowing that in most real use cases a bare timestamp value or `=` would never actually be used?

grondo (commented):
Sorry, I keep going through to review this and end up getting interrupted. Since it is working, my inclination is just get this in and we can deal with any unforeseen issues as we encounter them (though it doesn't appear there will be any)

Before that, I have one issue with the timestamp constraints (noted inline). I'm not sure how I feel about this one so I'm on the fence, but though maybe some discussion would be beneficial.

src/modules/job-list/match.c
chu11: Yeah, I picked `>=` since I figured it was the "best" of all the choices.

Defaulting to `=` does seem like an interesting idea and would make things consistent to everything else in the constraints.  However, I guess my concern would be if someone did a constraint like  `{"t_inactive": [ "SOMETHING" ]}`, basically this constraint would never return anything and a user could be confused.  I dunno if that's better or worse.

I almost feel that just requiring ">", "<", ">=", or "<=" would be better.  What do you think of just making it a requirement?

grondo: I could go either way. I like making it a requirement better than having a default of `>=`. I don't personally think users are going to be constructing these constraints by hand, so I don't think allowing `{"t_inactive": [ "SOMETHING" ]}` would be confusing per se (but it would be difficult to test).

chu11:
rebased and re-pushed an update in which all of the timestamps now require a comparison operator.  A minor fallout of this is that all timestamp values must be strings now, i.e. `">=1234.5", b/c we always require the comparison operator.  I doubt that's a big deal, (it's no big deal in atleast python).  I think it does make the comparison more clear.  Tests updated as a result.

also we should probably merge

https://github.com/flux-framework/rfc/pull/377
and
https://github.com/flux-framework/flux-core/pull/5137

first.  It's mostly a technicality, since I do something in this PR that the RFC clarifies.

garlick (commented):
One general question:  Is there a benefit in converting the RFC 31 JSON constraint object to `struct list_constraint` as opposed to just leaving it as a JSON object and using jansson methods to manipulate it?   A large fraction of the matching library seems to be constructors, destructors, and conversion.

chu11:
> One general question: Is there a benefit in converting the RFC 31 JSON constraint object to struct list_constraint as opposed to just leaving it as a JSON object and using jansson methods to manipulate it? A large fraction of the matching library seems to be constructors, destructors, and conversion.

I was mostly following design pattern used in `librlist'.  Given the current matching that is supported, the use is perhaps a tad limited given its just a bunch of string compares and what not.  However, later on there will be things like hostlist filtering, where we will need to convert possible hostlist ranges and things like that into their appropriate internal data structures.

grondo (approved):
Just went through this again and did some extra testing and LGTM!

chu11:
setting MWP

PR #5113, opened 2023-04-24
ensure license and other informational files appear in distribution tarball

grondo:
@jan-janssen noticed that the flux-core `LICENSE` file was missing from our distribution tarball.

This PR adds that and some other missing licenses and informational files to `EXTRA_DIST` where appropriate so those files are included in the distribution.

PR #5131, opened 2023-04-30
sdbus: return useful errno values to requests on error

garlick:
Problem: sdbus users can't tell the difference between recoverable and non-recoverable errors

Return EAGAIN on recoverable error (sdbus is reconnecting to D-Bus) and ENOSYS when the module is unloading.

There's also a small correction to the EXTRA_DIST additions in #5113

### DIFF
--- a/src/modules/sdbus/sdbus.c
+++ b/src/modules/sdbus/sdbus.c
@@ -603,8 +603,15 @@ void sdbus_ctx_destroy (struct sdbus_ctx *ctx)
 {
     if (ctx) {
         int saved_errno = errno;
-        flux_msglist_destroy (ctx->requests);
-        flux_msglist_destroy (ctx->subscribers);
+        const char *errmsg = "module is unloading";
+        if (ctx->subscribers) {
+            bulk_respond_error (ctx->h, ctx->subscribers, ENOSYS, errmsg);
+            flux_msglist_destroy (ctx->subscribers);
+        }
+        if (ctx->requests) {
+            bulk_respond_error (ctx->h, ctx->requests, ENOSYS, errmsg);
+            flux_msglist_destroy (ctx->requests);
+        }
         flux_msg_handler_delvec (ctx->handlers);
         flux_watcher_destroy (ctx->bus_w);
         flux_future_destroy (ctx->f_subscribe);
--- a/src/modules/sdbus/sdbus.c
+++ b/src/modules/sdbus/sdbus.c
@@ -569,8 +569,8 @@ static void sdbus_recover (struct sdbus_ctx *ctx, const char *reason)
 
     /* Send any pending requests an error.
      */
-    bulk_respond_error (ctx->h, ctx->subscribers, ENOSYS, reason);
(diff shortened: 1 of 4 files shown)


t/t2407-sdbus.t
```diff
@@ -13,6 +13,10 @@ if ! systemctl --user show --property Version; then
 	skip_all="user systemd is not running"
 	test_done
 fi
+if ! busctl --user status >/dev/null; then
```
chu11, line 16: commit message typo - to "skip the tests" (and not sure that has to be in double quotes?)

chu11: a complete aside comment, should these fixes be copied/updated in `t2405-job-exec-sdexec.t` and `t2410-exec-systemd.t` too?  Possibly elsewhere too?

In those tests it checks

```
#  Don't run if systemd environment variables not setup
if test -z "$DBUS_SESSION_BUS_ADDRESS" \
     -o -z "$XDG_RUNTIME_DIR"; then
    skip_all='DBUS_SESSION_BUS_ADDRESS and/or XDG_RUNTIME_DIR not set'
    test_done
fi
```

which is probably why its not a problem, assuming the DBUS env is set correctly.

But perhaps it would be good to make the check consistent across all tests.  Could be a new issue and we can follow up later.

src/modules/sdbus/sdbus.c
```diff
@@ -569,8 +569,8 @@ static void sdbus_recover (struct sdbus_ctx *ctx, const char *reason)
 
     /* Send any pending requests an error.
      */
-    bulk_respond_error (ctx->h, ctx->subscribers, ENOSYS, reason);
-    bulk_respond_error (ctx->h, ctx->requests, ENOSYS, reason);
+    bulk_respond_error (ctx->h, ctx->subscribers, EAGAIN, reason);
```
chu11, line 572: commit message "return EAGAIN instead" :-)

t/t2408-sdbus-recovery.t
```diff
@@ -77,7 +77,20 @@ test_expect_success NO_CHAIN_LINT 'background subscription fails with EAGAIN' '
 	test_must_fail wait $pid &&
 	grep "Errno 11" signals.err
 '
+test_expect_success NO_CHAIN_LINT 'initiate another subscription' '
```
chu11, line 80: commit message "check and add"?  Or perhaps just "add a" .. and i guess technically "test" should be plural :-)

chu11 (approved):
LGTM, just some commit message typos, but one comment below on if the fixes need to be in more places.  But could be a follow up PR.

t/t2407-sdbus.t
garlick: I'm not sure at this point what we're going to end up keeping after we finally get `job-exec` working with the new approach, so I'd prefer we table that one for now.  Helper function seems like overkill to me right now.

garlick:
Thanks for catching the typos!  I've force pushed those changes.

Issue #5132, opened 2023-04-30
parent-uri confuses utilities run under flux-proxy(1)


grondo:
When `flux-proxy` connects to a nested Flux instance, utilities that utilize the `parent-uri` broker attribute do not work because the `parent-uri` is always encoded as a `local://` URI, but the source of the proxy connection is not local.

I can't think of a good way around this besides encoding `parent-uri` as a remote URI, but this means all connections to the parent-uri will use ssh, even if initiated locally.

The use case that uncovered this problem was something like this:
```console
$ flux pstree -x
       JOBID USER     ST NTASKS NNODES  RUNTIME
 ƒ3PY9vN7QXR grondo    R     12      3   12.73s ./instance.sh
     ƒy2jris grondo    R      1      1   10.24s ├── test.py
     ƒkKv1fV grondo    R      9      3   11.07s └── ./instance.sh
     ƒyybQ9d grondo    R      1      1   8.616s     ├── test.py
     ƒkNszEB grondo    R      6      2   9.485s     └── ./instance.sh
     ƒzBTJQP grondo    R      1      1   6.933s         ├── test.py
     ƒmouJ1V grondo    R      3      1   7.771s         └── ./instance.sh
     ƒbL4NU7 grondo    R      1      1   6.221s             └── test.py
$ flux proxy ƒ3PY9vN7QXR/ƒkKv1fV flux job timeleft
flux-job: failed to connect to parent instance: No such file or directory
```

grondo:
Perhaps the best thing to do here is

 1. export an environment variable, e.g. `FLUX_PROXY_REMOTE` that contains the user/hostname part (the "authority") of the URI proxy used to connect
 2. Add a wrapper function for getting the parent URI (e.g. `uri_get_parent()`), which looks for this environment variable, and substitutes the `local://` URI prefix with `ssh://${FLUX_PROXY_REMOTE}/` when it is set.
 3. Update callers of `flux_attr_get (h, "parent-uri")` to instead call `uri_get_parent(3)`
 4. Ensure that `FLUX_PROXY_REMOTE` is cleared from the environment for jobs, flux-exec(1) etc.

A proof of concept seems to fix the `flux job timeleft` case

Perhaps `uri_get_parent()` should actually be added to libflux-core, and additionally a "parent" resolver should be added to `flux-uri(1)` for completeness.

```console
$ flux pstree -x
       JOBID USER     ST NTASKS NNODES  RUNTIME
[...]

PR #5133, opened 2023-05-01
convert parent-uri attribute to remote under flux-proxy(1)

grondo:
This is a WIP fix for #5132. Posting before tests are fully complete (though testing will be trivial) because I'm not sure the implementation will be acceptable.

This PR adds special handling of `parent-uri` directly to `attr_get()`. If there is a request for the `parent-uri` attribute and an environment variable `FLUX_PROXY_REMOTE` is currently set, then the local `parent-uri` is transparently transformed to a remote URI using the remote "authority" in `FLUX_PROXY_REMOTE` and is cached locally as `parent-remote-uri`.

When `flux-proxy(1)` uses a remote URI to connect to an instance, it then sets `FLUX_PROXY_REMOTE` appropriately in the environment of the spawned subprocess. Then, care is taken to unset this environment variable if other subprocesses are started, e.g. as part of a job or flux-exec, etc.

This is not the most satisfying implementation because it pushes knowledge of a specific attribute down into the attr code.
However, the end result is satisfying since even `flux getattr parent-uri` will now return a _usable_ remote URI to the flux-proxy instance, but still the local uri in other cases (i.e. jobs, the broker itself, etc).

The alternative implementation is to add a `uri_get_parent()` function or similar and convert all users of the `parent-uri` attribute to that. This would still leave `flux getattr` and `flux lsattr` and all callers that still directly use `flux_attr_get(3)` susceptible to an issue.

e.g.
```console
$ flux pstree -x
 fovomNPbh3V grondo    R      1      1   25.73s flux
    f8bpyT3u grondo    R      1      1   5.975s └── flux
$ src/cmd/flux proxy fovomNPbh3V/f8bpyT3u printenv FLUX_PROXY_REMOTE
tioga31
$ src/cmd/flux proxy fovomNPbh3V/f8bpyT3u flux getattr parent-uri
ssh://tioga31/var/tmp/grondo/flux-lDWKJi/local-0
$ src/cmd/flux proxy fovomNPbh3V/f8bpyT3u flux exec -r 0 flux getattr parent-uri
local:///var/tmp/grondo/flux-lDWKJi/local-0
```

### DIFF
--- a/src/broker/runat.c
+++ b/src/broker/runat.c
@@ -76,6 +76,7 @@ static const char *env_blocklist[] = {
     "FLUX_TASK_LOCAL_ID",
     "FLUX_URI",
     "FLUX_KVS_NAMESPACE",
+    "FLUX_PROXY_REMOTE",
     "PMI_FD",
     "PMI_RANK",
     "PMI_SIZE",
--- a/src/cmd/builtin/proxy.c
+++ b/src/cmd/builtin/proxy.c
@@ -43,6 +43,7 @@ struct proxy_command {
     flux_subprocess_t *p;
     int exit_code;
     uid_t proxy_user;
+    char *remote_uri_authority;
 };
 
 static const char *route_auxkey = "flux::route";
@@ -140,6 +141,12 @@ static int child_create (struct proxy_command *ctx,
     }
 
     if (flux_cmd_setenvf (cmd, 1, "FLUX_URI", "local://%s", sockpath) < 0)
+        goto error;
+    if (ctx->remote_uri_authority
+        && flux_cmd_setenvf (cmd, 1,
+                             "FLUX_PROXY_REMOTE",
+                             "%s",
+                             ctx->remote_uri_authority) < 0)
         goto error;
 
     /* We want stdio fallthrough so subprocess can capture tty if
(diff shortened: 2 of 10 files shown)


garlick (commented):
I say go ahead with this - I can't think of a better way to solve it!   (my $0.02).

grondo:
Ok, removed WIP and added some tests.

grondo:
Just force-pushed a fix for a broken-&& chain in the tests.

garlick (approved):
LGTM!  One question though.  Does FLUX_PROXY_REMOTE need to be cleared from the job environment too?

E.g. if I submit a job via a proxy session and then that job tries to submit a follow-on job with `--parent`, will it unnecessarily use `ssh://` (and do we care)?

grondo:
> Does FLUX_PROXY_REMOTE need to be cleared from the job environment too?

Yes, it should! I thought I had a commit that had the job shell remove this environment variable, but I must have dropped it.

I'll fix that.

grondo:
Ok I added a commit to have the job shell unset `FLUX_PROXY_REMOTE` if set for all job tasks. It is a bit unsatisfying that this has to be done in a few places (actually the one in flux-rexec probably isn't needed, but it doesn't hurt).

grondo:
I'll go ahead and set MWP here now and add this PR to NEWS.md.

PR #5137, opened 2023-05-02
librlist: fix RFC31 corner cases

chu11:
This updates a corner case and adds tests given updates in RFC31.  Marking this as WIP as this assumes https://github.com/flux-framework/rfc/pull/377 will be accepted, so we shouldn't merge until after it is merged.


chu11:
removing WIP, as I think we're getting closer on the constraints stuff.  But should be merged after  https://github.com/flux-framework/rfc/pull/377 goes in.

PR #5141, opened 2023-05-03
make SchedResourceList optional in flux.job.info

vsoch:
Problem: We currently cannot package rlist.h with the install (at least without more work) and since this is a required import for flux.job.info, without it the Python bindings (installed via pip) will error.
Solution: the flux.resource module is not technically required, so we can fallback to catching the ImportError instead and using the same empty (dummy) class to handle self.resources

I am going to test this with the Python bindings, and would like to get that working first before we finalize it here! Note that I'm going to be taking a break for a few minutes but I'll be back after. I'll update the issue after I have done some testing with this branch!

### DIFF
--- a/src/bindings/python/flux/job/info.py
+++ b/src/bindings/python/flux/job/info.py
@@ -22,8 +22,12 @@ from flux.core.inner import raw
 from flux.job.JobID import JobID
 from flux.job.stats import JobStats
 from flux.memoized_property import memoized_property
-from flux.resource import SchedResourceList
 from flux.uri import JobURI
+
+try:
+    from flux.resource import SchedResourceList
+except ImportError:
+    SchedResourceList = None
 
 
 def statetostr(stateid, fmt="L"):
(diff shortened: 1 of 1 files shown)


vsoch:
@grondo @garlick this is fully working! I think if we want to release older versions of flux for flux-python I can do some custom and manual work to make that happen (it's basically backporting this change, at least I just tested with 0.48.0 and it worked) but if we can get this into a release, it should allow us to build python bindings from here on out. Tomorrow let's chat about  if you want me to do the backported releases - my thinking is that before we do the official ones (because you can't undo/redo a release on pypi) we should do the following:

- decide on a set of python tests for flux-python (if you want to point me to relevant files in core I can figure it out)
- once those are working, find a user / small set of users to test a release candidate for some version they have

When those two things are done I'll be comfortable doing "final" releases for older versions (with the manual work) and then I'll finish up the automated workflow for new releases. It's already written but I need to turn it on.

grondo (approved):
This seems like a good workaround for now, especially if it lets us move forward on the pip install of Python bindings. Thanks!

vsoch:
Thanks! And let me know when you want to discuss the plan above - likely y’all can do the release work first and come back to it. I’m planning on rebuilding my flux images with this new release too.

PR #5144, opened 2023-05-03
job-list: misc cleanup

chu11:
This is some misc cleanup / doc updates peeled off from another branch.  It also include cleanup in `flux-job`, testsuite, and `flux-jobs(1)`.

### DIFF
--- a/doc/man1/flux-jobs.rst
+++ b/doc/man1/flux-jobs.rst
@@ -46,10 +46,11 @@ OPTIONS
    Limit output to N jobs (default 1000)
 
 **--since**\ *WHEN*
-   Limit output to jobs that completed or have become inactive since a
-   given timestamp. This option implies ``-a`` if no other ``--filter``
-   options are specified. If *WHEN* begins with ``-`` character, then
-   the remainder is considered to be a an offset in Flux standard duration
+   Limit output to jobs that have been active since a given timestamp.  In other
+   words, jobs that are currently pending, currently running, or became inactive
+   since the given timestamp.  This option implies ``-a`` if no other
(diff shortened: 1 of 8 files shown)


chu11:
re-pushed with an extra commit from something I just noticed in the JobList class.  I cannot find evidence that we've ever accepted "all" as a filter name.  And "all" in the code only sets pending and running, not pending, running, and inactive.  So it doesn't make much sense.  I suspect its some left over code from development of the class?

Perhaps there is something I just can't remember or some use case I can't find.   But there's a cleanup commit in here for the time being.

garlick (approved):
This all looks good to me.

There was one (slightly fascist) misspelling in the add_filter commit message: "fitler".

chu11:
Thanks.  I would like a double check on the last commit from @grondo.  https://github.com/flux-framework/flux-core/pull/5144/commits/cbc306ba58ea95fd9e7460b1625c2288ec59bab2  Just to make sure I'm not forgetting something or its used in some way I can't find.

grondo (approved):
Yeah, that looks good to me!

grondo:
@chu11 I'm going to set MWP here in a bit unless I hear otherwise, we're trying to see if this mergify issue is affecting all PRs our just some...

grondo:
Ok, we might have to ask the folks at mergify :-(

chu11:
https://github.com/Mergifyio rebase

chu11:
https://github.com/Mergifyio rebase

grondo:
This is approved with MWP so I'll merge since mergify still now working.

PR #5150, opened 2023-05-04
job-manager: make some replay errors non-fatal

### LINKED ISSUE #5147: broker: recovery should be not be aborted on messed up job directory
Problem: a job with a missing eventlog causes flux to refuse to start with
```
flux[273338]: job-manager.err[0]: restart failed: lookup job.0005.b398.c700.0400.eventlog: No such file o>
```

We should probably try to continue starting and ignore jobs that are incomplete.

garlick:
Problem: if a few jobs get messed up in the KVS due to an improper shutdown, recovery is a tedious process involving starting flux in `--recovery` mode, fixing one job, and starting again.
    
When a job cannot be replayed from the KVS and the reason is that the directory is incomplete, log the failure at LOG_ERR level but let replay continue and ultimately the flux restart be successful.

If a job has more serious problems like incorrect content in the eventlog, treat that as a fatal error as before.  This avoids breaking the 'valid' tests that check backwards compatibility with older kvs dumps, which might use an older eventlog format.
    
Update t2219-job-manage-restart.t to expect warnings rather than failure when such jobs are encountered during replay.
    
Fixes #5147


chu11 (approved):

chu11:
something just occurred to me.  Other than a studious admin looking through the logs, how would anyone know there are jobs that are bad in the KVS?  Will they just linger there forever?  Although I didn't look into it thoroughly yet, I'm not sure that `flux job purge` would do anything b/c the jobs aren't known in the job-manager?

Not necessarily in this issue, but perhaps we need some uhh "alert"-ish way to know?  perhaps a "flux job kvs-status" or something?

garlick:
> Not necessarily in this issue, but perhaps we need some uhh "alert"-ish way to know? perhaps a "flux job kvs-status" or something?

Great point!

> Edit: random thought, the job ids could be saved off into a "bad list", and flux job purge --special-bad-ids could handle it?

Heh like a lost+found?  I like it.

garlick:
I'll go ahead and set MWP on this one and we can think about your idea separately.  Thanks for the review!

garlick:
One of the runners was complaining of an uninitialized variable (I don't agree but...) so fixed that and forced a push.

grondo:
I have no idea, I have not seen that one before...

garlick:
@mergifyio refresh

chu11:
@mergifyio rebase

chu11:
with mergify down, going to hit the button

PR #5153, opened 2023-05-05
broker: redefine broker.quorum as a size

### LINKED ISSUE #5145: broker: add ability to configure "quorum" as an instance size threshold
Problem: there is a use case (flux-framework/flux-operator#161) for flux entering RUN state when the instance size reaches a threshold as opposed to the specific set of ranks that can be configured with `broker.quorum`:
```
     broker.quorum [Updates: C]
              An  RFC  22 idset representing broker ranks that are required to
              be online before the rank 0 broker  enters  the  RUN  state  and
              starts the initial program, if any.  Default: all ranks.
```

garlick:
Problem: broker.quorum as an idset is less useful than if it were a size threshold.
    
Redefine broker.quorum as a size.

Fixes #5145

WIP pending some manual testing...

### DIFF
--- a/src/broker/state_machine.c
+++ b/src/broker/state_machine.c
@@ -38,8 +38,8 @@
 #include "shutdown.h"
 
 struct quorum {
-    struct idset *want;
-    struct idset *have; // cumulative on rank 0, batch buffer on rank > 0
+    struct idset *all;
+    struct idset *online; // cumulative on rank 0, batch buffer on rank > 0
     flux_future_t *f;
     double timeout;
     bool warned;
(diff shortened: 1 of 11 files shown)


garlick:
Repushed with some backwards compatibility to avoid breaking fluxion.

garlick:
This works as expected on my test system so I'm removing the WIP.

src/broker/state_machine.c
```diff
@@ -27,6 +27,7 @@
 #include "src/common/libutil/errprintf.h"
 #include "src/common/libsubprocess/server.h"
 #include "ccan/array_size/array_size.h"
+#include "ccan/str/str.h"
```
grondo, line 30: Commit message: I think `broker.quorum=<size-1>` :arrow_right: `broker.quorum=0-<size-1>`

grondo (approved):
LGTM! Just noticed one possible commit message improvement.

garlick:
Thanks I'll set MWP.

Issue #5148, opened 2023-05-04
prevent or mitigate jobs writing large files to kvs stdio


grondo:
A production flux instance became unresponsive recently because the KVS was incapacitated trying to return the output eventlog for a job. The job had apparently gone haywire and was writing the same line of output continuously for some time (I think we estimated about 7 million lines). The kvs was unresponsive for the most part (listing keys seemed to work, but not fetching anything, though I could be mistaken about that)

This is the first time we've seen something like this, but we should definitely have some mitigation for jobs inadvertently writing tons of output to the kvs.

Many ideas were already discussed, including:

 - never write any output to the kvs
 - prevent any `flux submit` or `flux run` jobs at the system instance level using the require-instance validator. If users do this in their own jobs then they only shoot their own two feet.
 - handle this in the job shell output plugin (and possibly `flux job attach` input handler) to set a maximum number of lines or bytes for stdio eventlogs
 - do something similar in the KVS, perhaps an io quota per job or guest namespace

garlick:
Another item for the list:  I think right now a kvs get always returns the entire value in one message, even if it's stored internally in chunks.  It would be more fair to other messages to make kvs get a streaming rpc that returns sequential chunks, since the other messages could be transferred while the big message is streaming.

The kvs design predates streaming rpcs I think.

grondo:
> hypothetically flux job attach just stops outputting after N lines and outputs to the user "too many lines, stopping output" or whatever.

But due to what @garlick said above:

>  I think right now a kvs get always returns the entire value in one message,

I think we're assuming the problem is that the key is already huge, and the problem is that the kvs is assembling it all at once to return the first result to `job-info` watch-eventlog handler, so I'm not sure `flux job attach` can influence this at that time. By the time it can decide "too many lines", the entire object has been fetched from the kvs.

In fact this makes me wonder if there isn't also a problem in `job-info`. It appears that `watch_continuation` goes through every line in an eventlog synchronously after it is fetched from the kvs to return each entry to the caller as a separate RPC. In this case maybe job-info was stuck as well as the kvs. (or maybe I assumed the backtrace was the kvs)

grondo:
I like @garlick's idea of limiting output data in the shell output plugin. RFC 24 even has a [Redirect Event](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_24.html#redirect-event), so the shell could potentially post a redirect event and start directing further output to a file after hitting the limit. (T

[...]

PR #5155, opened 2023-05-05
shell: truncate output to KVS after 10MB

grondo:
This is a WIP pending tests and is meant as a stopgap workaround for #5148.

This PR simply sets a static threshold for KVS output and stops writing to the output eventlog after the threshold is reached.
A warning is issued at the time output is truncated as well as at the end of the job.

Figured I'd post early to get feedback.

E.g.:
```console
$ flux run lptest 79 140000 | wc -c
4.842s: flux-shell[0]:  WARN: output: stdout will be truncated, 10MB limit exceeded
5.240s: flux-shell[0]:  WARN: output: stdout: 714240 of 11200000 bytes truncated
10485760
```

### DIFF
--- a/src/common/libioencode/ioencode.c
+++ b/src/common/libioencode/ioencode.c
@@ -102,11 +102,16 @@ static int decode_data_base64 (char *src,
 {
     ssize_t rc;
     size_t size = base64_decoded_length (srclen);
-    if (!(*datap = malloc (size)))
-        return -1;
-    if ((rc = base64_decode (*datap, size, src, srclen)) < 0)
-        return -1;
-    *lenp = rc;
+    if (datap) {
+        if (!(*datap = malloc (size)))
+            return -1;
+        if ((rc = base64_decode (*datap, size, src, srclen)) < 0)
+            return -1;
+        if (lenp)
+            *lenp = rc;
+    }
+    else if (lenp)
+        *lenp = size;
     return 0;
 }
 
@@ -154,8 +159,9 @@ int iodecode (json_t *o,
         (*streamp) = stream;
     if (rankp)
         (*rankp) = rank;
-    if (datap) {
-        *datap = NULL;
+    if (datap || lenp) {
+        if (datap)
+            *datap = NULL;
         if (data) {
             if (encoding && strcmp (encoding, "base64") == 0) {
                 if (decode_data_base64 (data, len, datap, &bin_len) < 0)
@@ -163,9 +169,11 @@ int iodecode (json_t *o,
             }
             else {
                 bin_len = len;
-                if (!(*datap = malloc (bin_len)))
-                    goto cleanup;
-                memcpy (*datap, data, bin_len);
+                if (datap) {
+                    if (!(*datap = malloc (bin_len)))
+                        goto cleanup;
+                    memcpy (*datap, data, bin_len);
+                }
             }
         }
     }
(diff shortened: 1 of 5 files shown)


src/common/libioencode/ioencode.c
```diff
@@ -102,11 +102,16 @@ static int decode_data_base64 (char *src,
 {
     ssize_t rc;
```
chu11, line 103: commit message typo "unnecessarily", but probably should be "unnecessary" :-)

chu11 (commented):
I suppose in time we'll some of this configurable, like hypothetically a "hard limit" is 50mb, but we default to 10mb or something, and the user could tweak it up if they really want to.

my only random thought is if the threshold for stderr should be bigger than stdout, b/c that might capture critical debugging information.  But, we have to pick a cap so I'm not sure if it matters much.  There's always a chance critical debugging information is missed.

grondo:
> I suppose in time we'll some of this configurable, like hypothetically a "hard limit" is 50mb, but we default to 10mb or something, and the user could tweak it up if they really want to. 

Yeah, I think long term this should be configurable, perhaps via broker configuration as suggested by @garlick so there could be a smaller limit imposed for the system instance vs single user instances. This PR is just meant to be something we can perhaps sneak into the current release to prevent an episode like we saw the other day.

grondo:
Oh, also I should mention the eventual plan is to redirect output to a file instead of truncating, so your other concern will be addressed int he future. I think the output plugin just needs some nontrivial refactoring to enable that at the moment.

chu11 (approved):
this LGTM for the immediate need it serves

grondo:
I added a couple tests. I used the `LONGTEST` prereq since they're a little bit slow.

t/t2606-job-shell-output-redirection.t
```diff
@@ -348,4 +348,15 @@ test_expect_success 'job-shell: shell errors are captured in error file' '
...
+	dd if=/dev/urandom bs=10240 count=800 | base64 --wrap 79 >expected &&
+	flux run cat expected >output 2>truncate.error &&
+	test_debug "cat truncate.error" &&
+	grep truncated truncate.error
+'
+test_expect_success LONGTEST 'job-shell: stderr to kvs is truncated at 10MB' '
+	dd if=/dev/urandom bs=10240 count=800 | base64 --wrap 79 >expected &&
+	flux run sh -c "cat expected >&2"  >truncate2.error 2>&1 &&
+	grep truncated truncate2.error
+'
```
chu11, line 361: minor suggestion, also grep for "stdout" or  "stderr", just to make sure it's truncating the right stream

chu11 (approved):
LGTM, just one suggestion below

src/common/libioencode/ioencode.c
```diff
@@ -154,18 +159,21 @@ int iodecode (json_t *o,
...
                     goto cleanup;
             }
             else {
                 bin_len = len;
-                if (!(*datap = malloc (bin_len)))
-                    goto cleanup;
-                memcpy (*datap, data, bin_len);
+                if (datap) {
+                    if (!(*datap = malloc (bin_len)))
+                        goto cleanup;
```
garlick, line 174: Same deal here - assign `datap` after malloc has succeeded.

garlick (approved):
LGTM!

t/t2606-job-shell-output-redirection.t
grondo: Good suggestion!

grondo:
Ok, repushed with suggested change to the tests.

I also thought I saw a couple comments from @garlick, but now I can't find them...

garlick:
> I also thought I saw a couple comments from @garlick, but now I can't find them...

I decided against making those comments.  There was a mem leak on an error path in iodecode/ioencode but they were pre-existing and I didn't want to hold up this PR.  I'll open an issue to revisit later.

Issue #5119, opened 2023-04-26
How to access a wide range of job information for a user tool


grondo:
Hi all! Just chiming in late with the context for my request.

Post TOSS 4 updates on LC, users have asked about the disappearance of the `checkjob` command on systems running Slurm, which is a handy Moab utility for seeing all sorts of job stats. Since we're not supporting Moab anymore, Jeff Long & I put together a tool that nicely formats some of the output you can get with `squeue` and `sacct`. The result looks something like

```
janeh@pascal83:~$ slurm_jobinfo.py -j 16436_9
JobID        :  <JobID>
JobName      :  Symmetric
State        :  PENDING
User         :  <user>
Group        :  <user's group>
Account      :  <bank>
Partition    :  pbatch
QOS          :  normal
Timelimit    :  1-00:00:00
Submit       :  Tue 04/25 08:49:06
Eligible     :  Tue 04/25 08:49:07
Start        :  Unknown
End          :  Unknown
Elapsed      :  00:00:00
Priority     :  124135
NNodes       :  4
NCPUS        :  144
MinCPUNode   :
NodeList     :  None assigned
WorkDir      :  /p/lustre1/<user directory>
SubmitLine   :  sbatch --array=0-12 Submit_4_nodes.sh
Dependency   :  (null)
EstStart     :  N/A
Reason       :  (AssocMaxJobsLimit)
```

A user requested a similar utility for Flux (as well as more tutorials on how to get job stats with native Flux commands), and I was trying to use the Python APIs to get there. What I have so far:

```
janeh@tioga10:~$ flux_jobinfo.py -j fnwSSLATbTM
Getting info for Flux JobID: fnwSSLATbTM
JobID        :  fnwSSLATbTM
JobName      :  flux
State        :  RUN
User         :  <user>
Partition    :  pdebug
Timelimit    :  4:00:00
Submit       :  Wed 04/26/2023 09:20:44
Elapsed      :  2:14:53
Start        :  Wed 04/26/2023 09:20:44
End          :  Wed 04/26/2023 13:20:44
NNodes       :  1
NCores       :  64
NTasks       :  1
NodeList     :  tioga26
Dependencies  :  []
```

Thanks so much for helping me out @vsoch!

_Originally posted by @xorJane in https://github.com/flux-framework/flux-docs/issues/229#issuecomment-1523885058_

chu11:
link to a brainstorming i just did #5120

grondo:
Some initial comments:

[...]

PR #5156, opened 2023-05-06
job-list: support retrieving current working directory for jobs

chu11:
TSIA

I think the only minor question is A) is `cwd` the right abbreviation to use for this, vs "workdir" or "jobdir" or something else.  I think `cwd` is good personally.  B) is `CWD` a get header for it in `flux-jobs`?  Could be `DIR`? `JOBDIR`?  I'm less sure of the header `CWD`, I guess its fine but certainly wouldn't be wrong to tweak it.


grondo (approved):
LGTM! I think `cwd` and `CWD` header are fine here.

chu11:
Thanks, pinged initiator of #5119 and she felt `cwd` was fine as well.  Setting MWP.

Issue #5158, opened 2023-05-06
libioencode: memleak on error path of libiodecode()


garlick:
Problem: `iodecode()` leaks memory if base64 decode fails and assigns 'datap' even when it returns an error.

These functions set/protect errno in some error cases and not in others.  Whether or not the intent is to set errno should be documented, and then behavior should be made consistent.

(mentioned in #5155)

chu11:
doh! and i see another path in the ioencode side.

PR #5159, opened 2023-05-06
libioencode: fix memleaks on error paths

chu11:
Problem: There are several memory leaks in error paths in libioencode.

Make sure to free memory on error paths.

Fixes https://github.com/flux-framework/flux-core/issues/5158

Also, cleanup some unnecessary goto statements.

### DIFF
--- a/src/common/libioencode/ioencode.c
+++ b/src/common/libioencode/ioencode.c
@@ -53,7 +53,6 @@ json_t *ioencode (const char *stream,
                   bool eof)
 {
     json_t *o = NULL;
-    json_t *rv = NULL;
 
     /* data can be NULL and len == 0 if eof true */
     if (!stream
@@ -76,22 +75,20 @@ json_t *ioencode (const char *stream,
              *   encoding the data in base64:
              */
             if (!(o = data_encode_base64 (stream, rank, data, len)))
-                goto error;
+                return NULL;
         }
     }
     else {
         if (!(o = json_pack ("{s:s s:s}",
                              "stream", stream,
                              "rank", rank)))
-            goto error;
+            return NULL;
     }
     if (eof) {
         if (json_object_set_new (o, "eof", json_true ()) < 0)
-            goto error;
+            return NULL;
     }
-    rv = o;
-error:
-    return rv;
+    return o;
 }
 
 static int decode_data_base64 (char *src,
@@ -131,7 +128,6 @@ int iodecode (json_t *o,
     int eof = 0;
     bool has_data = false;
     bool has_eof = false;
-    int rv = -1;
 
     if (!o) {
         errno = EINVAL;
@@ -142,7 +138,7 @@ int iodecode (json_t *o,
                      "stream", &stream,
                      "rank", &rank,
                      "encoding", &encoding) < 0)
-        goto cleanup;
+        return -1;
     if (json_unpack (o, "{s:s%}", "data", &data, &len) == 0)
         has_data = true;
     if (json_unpack (o, "{s:b}", "eof", &eof) == 0)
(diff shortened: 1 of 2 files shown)


chu11:
> I didn't see a fix for the other two issues mentioned in https://github.com/flux-framework/flux-core/issues/5158:

Doh! Sorry, I completely blanked on the second half of that issue description.

chu11:
re-pushed fixing

- memleaks
- consistently setting errno
- consistently setting return parameters

hopefully that's all the badness in there

src/common/libioencode/ioencode.c
```diff
@@ -42,7 +43,7 @@ static json_t *data_encode_base64 (const char *stream,
                            "encoding", "base64",
                            "data", dest, n);
     }
-    free (dest);
+    ERRNO_SAFE_WRAP (free, dest);
```
garlick, line 46: `json_pack()` does not set errno on failure so probably need to set it in that case.

If we're guaranteeing that errno is set on failure, then IMHO we should state it in the header file so users know what to expect.

garlick (approved):
Looks good, thanks!  I think I spotted one place where errno is not set, and added a request about documenting errno behavior in the header, but O/W looks fine.

src/common/libioencode/ioencode.c
chu11: > json_pack() does not set errno on failure so probably need to set it in that case.

Ahhh, I was only looking at the `malloc` and `base64_encode()` path in that function.  `json_pack()` could fail too.

chu11:
re-pushed with minor tweaks per comments above, nothing too big from what was there before

```C
diff --git a/src/common/libioencode/ioencode.c b/src/common/libioencode/ioencode.c
index 14570d1..e8392f0 100644
--- a/src/common/libioencode/ioencode.c
+++ b/src/common/libioencode/ioencode.c
@@ -37,11 +37,12 @@ static json_t *data_encode_base64 (const char *stream,
 
     if ((dest = malloc (destlen))
         && (n = base64_encode (dest, destlen, data, len)) >= 0) {
-            o = json_pack ("{s:s s:s s:s s:s#}",
-                           "stream", stream,
-                           "rank", rank,
-                           "encoding", "base64",
-                           "data", dest, n);
+        if (!(o = json_pack ("{s:s s:s s:s s:s#}",
+                             "stream", stream,
+                             "rank", rank,
+                             "encoding", "base64",
+                             "data", dest, n)))
+            errno = ENOMEM;
     }
     ERRNO_SAFE_WRAP (free, dest);
     return o;
diff --git a/src/common/libioencode/ioencode.h b/src/common/libioencode/ioencode.h
index a6186bf..6c4a804 100644
--- a/src/common/libioencode/ioencode.h
+++ b/src/common/libioencode/ioencode.h
@@ -17,6 +17,7 @@
 /* encode io data and/or EOF into RFC24 data event object
  * - to set only EOF, set data to NULL and data_len to 0
  * - it is an error to provide no data and EOF = false
+ * - returns RFC24 object on success, NULL on error with errno set
  * - returned object should be json_decref()'d after use
  */
 json_t *ioencode (const char *stream,
@@ -30,6 +31,7 @@ json_t *ioencode (const char *stream,
  * - if no data available, data set to NULL and len to 0
  * - data must be freed after return
  * - data can be NULL and len non-NULL to retrieve data length
+ * - returns 0 on success, -1 on error with errno set
  */
 int iodecode (json_t *o,
               const char **stream,
```

garlick:
Looks good, thanks!

chu11:
oh crud, i meant to fix up the PR title to better describe the collective fixes in here.  dunno if doing it post-merge matters for the release notes?  Couldn't find the release notes generating script.

PR #5160, opened 2023-05-08
ci: add fedora38 builder, update flux-security default to 0.8.0

grondo:
This PR adds a builder for Fedora 38, since this distro has GCC 13 and Python 3.11, a combo we're not yet testing.

The caliper version we've pinned won't build in fedora38, and the newest version isn't compatible with the current code that calls it in flux-core, so I've just disabled Caliper here. Honestly, I'm wondering if we should just remove the caliper support? Eh, not too much of a headache yet, so we can save that decision for another day.

### DIFF
--- a/src/test/docker/fedora38/Dockerfile
+++ b/src/test/docker/fedora38/Dockerfile
@@ -1,0 +1,78 @@
+FROM fedora:38
+
+LABEL maintainer="Mark Grondona <mgrondona@llnl.gov>"
+
+#  Enable PowerTools for development packages
+RUN yum -y update \
+ && yum -y update \
+#  Utilities
+ && yum -y install \
+	wget \
+	man-db \
+	less \
+	git \
+	sudo \
+	munge \
+	ccache \
+	lua \
+	mpich \
+	valgrind \
+	jq \
+	which \
+	file \
+	vim \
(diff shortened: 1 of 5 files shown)


src/test/docker/fedora38/Dockerfile
```diff
@@ -0,0 +1,79 @@
...
+	libtool \
+	autoconf \
+	automake \
+	gcc \
+	gcc-c++ \
+	libasan \
+	make \
+	cmake \
+#  Python
+	python36 \
```
chu11, line 38: dumb question, if python 3.11 is the default python, why install 3.6?  I notice its there for some other fedora images.

grondo: Good catch, I just copied the Fedora dockerfile (probably did that before too :facepalm:). Definitely should remove that before merging this!

grondo: Fixed that, updated the `fluxrm/testenv:fedora38` image, and force-pushed here.

chu11 (approved):
LGTM

chu11:
> Please let me know if you'd rather that change be split off from this PR.

I think it's ok.

PR #5161, opened 2023-05-08
shell: add missing `flux_` prefix to `shell_mustache_render(3)`

grondo:
The `shell_mustache_render(3)` function declared in `shell.h` is missing the `flux_` prefix. Since the shell is built to only export `flux_*` symbols, this means that this function is not available to dynamically loaded shell plugins.

Rename the function `flux_shell_mustache_render(3)`, add protection against a `NULL` `shell` parameter, and add a couple simple tests for invalid arguments to the testsuite.

### DIFF
--- a/src/shell/output.c
+++ b/src/shell/output.c
@@ -814,7 +814,7 @@ shell_output_setup_type_file (struct shell_output *out,
         return -1;
     }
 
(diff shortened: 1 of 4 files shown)


chu11 (approved):
LGTM

Issue #5136, opened 2023-05-02
add mechanism to restrict remote access to services


garlick:
Problem: although the `rexec` service uses a private broker mechanism (commit bbe2b7b1b263747a430bb52db0303a3e290845f6) to prevent remote use on rank 0, the `sdbus` service resides in a broker module and doesn't have access to this mechanism and thus cannot diferentiate local vs remote clients.

Idea: add an [RFC 12](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_12.html) role: FLUX_ROLE_LOCAL.  Set the role when a message is accepted on the unix domain socket in `connector-local`, and clear it in the broker when messages arrive on the overlay network from another rank.   Then the role can be tested in a service request handler.

Edit: this doesn't seem like a perfect fit to the "role" definition, so another option would be to add a `local` message flag by appropriating some bits of the rolemask or perhaps the 4 unused bits of the `type` byte.

PR #5162, opened 2023-05-08
restrict remote access to sdbus on rank 0

garlick:
The flux broker exec service on rank 0 does not allow any user, even the instance owner, to start processes remotely, so that flux cannot be used to attack (albeit only as the unprivileged flux user) a locked down management node from a compromised compute/login node.

Problem: the recently added `sdbus` module (not loaded by default) provides the ability to launch processes as the flux user, but has no restrictions.  Furthermore, the mechanism used to restrict the broker exec service only works within the broker proper and cannot be used in `sdexec`, which is a broker module.

This PR  replaces the broker-only method with a new method (proposed in #5136) that works anywhere, then adds the rank 0 restriction to the `sdbus` module.

### DIFF
--- a/src/common/libflux/message.c
+++ b/src/common/libflux/message.c
@@ -507,6 +507,15 @@ int flux_msg_authorize (const flux_msg_t *msg, uint32_t userid)
     return 0;
 }
 
+bool flux_msg_is_local (const flux_msg_t *msg)
+{
+    uint32_t rolemask;
+    if (flux_msg_get_rolemask (msg, &rolemask) == 0
+        && (rolemask & FLUX_ROLE_LOCAL))
+        return true;
+    return false;
+}
+
 int flux_msg_set_nodeid (flux_msg_t *msg, uint32_t nodeid)
 {
     if (msg_validate (msg) < 0)
--- a/src/common/libflux/message.c
+++ b/src/common/libflux/message.c
@@ -1247,26 +1247,39 @@ static void userid2str (uint32_t userid, char *buf, int buflen)
     assert (n < buflen);
 }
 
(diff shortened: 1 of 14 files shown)


src/common/libflux/message.c
```diff
@@ -1247,26 +1247,39 @@ static void userid2str (uint32_t userid, char *buf, int buflen)
...
+    if (rolemask == FLUX_ROLE_NONE)
+        snprintf (buf, buflen, "none");
+    else if (rolemask == FLUX_ROLE_ALL)
+        snprintf (buf, buflen, "all");
+    else {
+        int offset = 0;
+        for (int i = 0; i < sizeof (rolemask)*8; i++) {
+            if (rolemask & 1<<i) {
+                offset += roletostr (rolemask & 1<<i,
+                                     i > 0 ? "," : "",
```
chu11, line 1277: should this be `offset > 0`, only output comma if something was output previously?  Dunno if unit tests possible.

chu11:
oh i forgot, this needs a RFC12 update?

src/common/libflux/message.c
garlick: Oh you're right, thanks!

garlick: The magic of [open_memstream(3)](https://man7.org/linux/man-pages/man3/open_memstream.3.html) should make a test possible!

garlick:
Fixed that bug, and added a test.  I'll propose adding FLUX_ROLE_LOCAL to RFC 12 shortly.

chu11 (approved):
LGTM

garlick:
Thanks!  Setting MWP.

PR #5163, opened 2023-05-09
cleanup:  use ccan/str in place of strcmp(), strncmp()

garlick:
We've been gradually switching from using `strcmp()` to test for string equality to `streq()`.  This converts a chunk of older code.  It's not the whole enchilada but I figured a PR that touches _everything_ is more likely to conflict with pending PRs, so a chunk is OK?

### DIFF
--- a/src/broker/module.c
+++ b/src/broker/module.c
@@ -14,6 +14,7 @@
 #include <dlfcn.h>
 #include <argz.h>
 #include <czmq.h>
+#undef streq // redefined by ccan/str/str.h below
 #include <uuid.h>
 #include <flux/core.h>
 #include <jansson.h>
@@ -29,6 +30,7 @@
 #include "src/common/libutil/iterators.h"
 #include "src/common/libutil/errprintf.h"
 #include "src/common/libutil/errno_safe.h"
+#include "ccan/str/str.h"
 
 #include "module.h"
 #include "modservice.h"
@@ -300,7 +302,7 @@ int module_sendmsg (module_t *p, const flux_msg_t *msg)
      */
     if (p->muted) {
         if (type != FLUX_MSGTYPE_RESPONSE
-            || strcmp (topic, "broker.module-status") != 0) {
+            || !streq (topic, "broker.module-status")) {
             errno = ENOSYS;
             return -1;
         }
@@ -841,7 +843,7 @@ int module_unsubscribe (modhash_t *mh, const char *uuid, const char *topic)
     }
     s = zlist_first (p->subs);
     while (s) {
-        if (!strcmp (topic, s)) {
+        if (streq (topic, s)) {
             zlist_remove (p->subs, s);
             free (s);
             break;
@@ -858,7 +860,7 @@ static bool match_sub (module_t *p, const char *topic)
     char *s = zlist_first (p->subs);
 
     while (s) {
-        if (!strncmp (topic, s, strlen (s)))
+        if (strstarts (topic, s))
             return true;
         s = zlist_next (p->subs);
     }
--- a/src/common/libkvs/test/kvs_txn.c
+++ b/src/common/libkvs/test/kvs_txn.c
@@ -21,6 +21,7 @@
 #include "treeobj.h"
 
 #include "src/common/libtap/tap.h"
+#include "ccan/str/str.h"
 
 void jdiag (json_t *o)
 {
@@ -107,7 +108,7 @@ int check_string_value (json_t *dirent, const char *expected)
         json_decref (val);
         return -1;
     }
-    if (strcmp (expected, s) != 0) {
+    if (!streq (expected, s)) {
         diag ("%s: expected %s received %s", __FUNCTION__, expected, s);
         json_decref (val);
         return -1;
@@ -193,7 +194,7 @@ void basic (void)
         "1: retrieved");
     ok (txn_decode_op (entry, &key, &flags, &dirent) == 0,
         "1: txn_decode_op works");
-    ok (!strcmp (key, "foo.bar.baz")
+    ok (streq (key, "foo.bar.baz")
         && flags == FLUX_KVS_APPEND
         && check_int_value (dirent, 42) == 0,
         "1: put foo.bar.baz = 42");
@@ -203,7 +204,7 @@ void basic (void)
     jdiag (entry);
     ok (txn_decode_op (entry, &key, &flags, &dirent) == 0,
         "2: txn_decode_op works");
-    ok (!strcmp (key, "foo.bar.bleep")
+    ok (streq (key, "foo.bar.bleep")
         && flags == 0
         && check_string_value (dirent, "foo") == 0,
         "2: put foo.bar.baz = \"foo\"");
@@ -213,7 +214,7 @@ void basic (void)
     jdiag (entry);
     ok (txn_decode_op (entry, &key, &flags, &dirent) == 0,
         "3: txn_decode_op works");
-    ok (!strcmp (key, "a")
+    ok (streq (key, "a")
         && flags == 0
         && json_is_null (dirent),
         "3: unlink a");
@@ -223,7 +224,7 @@ void basic (void)
     jdiag (entry);
     ok (txn_decode_op (entry, &key, &flags, &dirent) == 0,
         "4: txn_decode_op works");
-    ok (!strcmp (key, "b.b.b")
+    ok (streq (key, "b.b.b")
         && flags == 0
         && treeobj_is_dir (dirent) && treeobj_get_count (dirent) == 0,
         "4: mkdir b.b.b");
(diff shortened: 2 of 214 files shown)


grondo:
Seems fine to me! One suggestion though, uses of `strncmp(3)` to check for a string prefix should also be replaced with `strstarts()`.

chu11:
> I figured a PR that touches everything is more likely to conflict with pending PRs

I'm a "pull off the bandaid" kinda person, so I think it'd be ok to do the whole thing.

grondo:
> I'm a "pull off the bandaid" kinda person, so I think it'd be ok to do the whole thing.

If you're willing, then I'm fine with that too as long as the changes stick to updating strcmp() and related calls. It will conflict with some in progress work, but typically I add the update to `streq()` as a separate commit so those commits can easily be dropped.

However, it is a lot of work to do it all at once, so I leave it up to you.

garlick:
Meh, it's a good chunk of the way there.  OK, might as well.

garlick:
Whew finally got all the way through.

chu11:
Are you trying to tweet this PR title?  :-)  (i assume cut & paste typo in title, I'll fix)

src/broker/module.c
```diff
@@ -14,6 +14,7 @@
 #include <dlfcn.h>
 #include <argz.h>
 #include <czmq.h>
+#undef streq
```
chu11, line 17: I like the `// redefined by ccan/str/str.h below` comment you added elsewhere

src/common/libkvs/test/kvs_txn.c
```diff
@@ -278,12 +279,12 @@ void basic (void)
...
+    ok (streq (key, "f.f.f")
         && flags == 0
-        && !strcmp (json_string_value (json_object_get (treeobj_get_data (dirent),
+        && streq (json_string_value (json_object_get (treeobj_get_data (dirent),
                                                         "namespace")),
                     "g.g.g")
-        && !strcmp (json_string_value (json_object_get (treeobj_get_data (dirent),
+        && streq (json_string_value (json_object_get (treeobj_get_data (dirent),
                                                         "target")),
                     "h.h.h"),
```
chu11, line 289: adjust some tabbing here

src/common/libpmi/upmi_libpmi2.c
```diff
@@ -271,7 +271,7 @@ static int op_get (flux_plugin_t *p,
      * employed by Flux to launch Flux.
      */
     else if ((ctx->flags & LIBPMI2_IS_CRAY_CRAY)
-        && (!strncmp (key, "flux.", 5))) {
+        && (strstarts (key, "flux."))) {
```
chu11, line 274: no longer need the extra parens?

src/common/libutil/blobref.c
```diff
@@ -88,20 +89,22 @@ static void sha256_hash (const void *data, int data_len, void *hash, int hash_le
...
-    if (strlen (s1) < len + 1 || s1[len] != '-')
-        return 0;
-    return !strncmp (s1, s2, len);
+    if (!strstarts (s1, s2))
+        return false;
+    s1 += strlen (s2);
+    if (*s1 != '-')
+        return false;
+    return true;
 }
```
chu11, line 100: The old logic ensured that `s1[len]` would not be an array out of bounds, but I think in the new function the array out of bounds is possible.  Didn't look at context surrounding this function yet, just want to double check that array out of bounds isn't possible given how this is used.

chu11 (commented):
TIL that if a diff is so long, eventually github just stops showing the code diff and you gotta click the "load diff" for each file :-)  That was a lot!

I mostly tried to skim for accidental `!strcmp()` -> `!streq()` mistakes, didn't find any.  Just noting a few nits I found along the way.

src/common/libutil/blobref.c
garlick: I'm not seeing how there could be an out of bounds access here?   If strlen(s1) < strlen(2) then strstarts() fails.  Thus strlen (s2) must be >= strlen (s1).  So s1 + strlen (s2) must be in bounds?

chu11: hmmm, ok i think i was thinking of strlen(s1) == strlen(s2), so `strstarts()` passes.  but under that case I guess we just get the NUL delimeter, so all is good.

garlick:
Just forced a push with the changes you suggested @chu11.  We'll see if that F58 error recurs - I wasn't spotting anything wrong in the changes to libutil/fluid.c or the test.

The mistake I made repeatedly was converting something like `strncmp(...) == 0` to `strstarts (...) == 0` (forgetting to drop the == 0), and then figuring out what random thing I broke when I touched _everything_.  Whee!    However I did scan the diff before posting and didn't find any more of that.

grondo:
Noticed a couple spots that could also be converted to `strstarts()` (IMO).
```diff
diff --git a/src/common/libjob/id.c b/src/common/libjob/id.c
index 90c2b01b8..487dc3abe 100644
--- a/src/common/libjob/id.c
+++ b/src/common/libjob/id.c
@@ -16,6 +16,7 @@
 
 #include "job.h"
 
+#include "ccan/str/str.h"
 #include "src/common/libutil/fluid.h"
 
 int flux_job_id_parse (const char *s, flux_jobid_t *idp)
@@ -35,7 +36,7 @@ int flux_job_id_parse (const char *s, flux_jobid_t *idp)
     /*  Ignore any `job.` prefix. This allows a "kvs" encoding
      *   created by flux_job_id_encode(3) to properly decode.
      */
-    if (strncmp (p, "job.", 4) == 0)
+    if (strstarts (p, "job."))
         p += 4;
     return fluid_parse (p, idp);
 }
diff --git a/src/common/libtaskmap/taskmap.c b/src/common/libtaskmap/taskmap.c
index b981869cd..f7ab9ea2c 100644
--- a/src/common/libtaskmap/taskmap.c
+++ b/src/common/libtaskmap/taskmap.c
@@ -22,6 +22,7 @@
 #include "src/common/libutil/errprintf.h"
 #include "src/common/libutil/lru_cache.h"
 #include "src/common/libutil/errno_safe.h"
+#include "ccan/str/str.h"
 #include "taskmap.h"
 
 struct taskmap_block {
@@ -380,7 +381,7 @@ static struct taskmap *taskmap_decode_pmi (const char *s, flux_error_t *errp)
         while (isspace (*tok))
             tok++;
 
-        if (strncmp (tok, "vector,", 7) == 0)
+        if (strstarts (tok, "vector,"))
             got_sentinel = true;
         else if (!is_empty (tok)) {
             if (!got_sentinel) {

```

grondo:
> Hmm wondering if I screwed something up that caused this CI failure in the "ascii" builder:

Looking at that test I wonder if it shouldn't have been skipped when flux-core is compiled in ascii-only mode. I'm not sure how it worked before in that mode exactly, or why the changes here changed the behavior.

grondo:
Ok, the fluid test didn't even include `config.h` so that was a problem :facepalm: . I think this diff will fix the test. Sorry about that!
```diff
diff --git a/src/common/libutil/test/fluid.c b/src/common/libutil/test/fluid.c
index c1121e4d5..38944099e 100644
--- a/src/common/libutil/test/fluid.c
+++ b/src/common/libutil/test/fluid.c
@@ -8,6 +8,9 @@
  * SPDX-License-Identifier: LGPL-3.0
 \************************************************************/
 
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
 #include <errno.h>
 #include <string.h>
 #include <locale.h>
@@ -91,6 +94,7 @@ void test_f58 (void)
         tp++;
     }
 
+#if !ASSUME_BROKEN_LOCALE
     if (setenv ("FLUX_F58_FORCE_ASCII", "1", 1) < 0)
         BAIL_OUT ("Failed to setenv FLUX_F58_FORCE_ASCII");
     ok (fluid_encode (buf, sizeof (buf), f58_tests->id, type) == 0,
@@ -99,6 +103,7 @@ void test_f58 (void)
         "fluid_encode with FLUX_F58_FORCE_ASCII used ascii prefix");
     if (unsetenv ("FLUX_F58_FORCE_ASCII") < 0)
         BAIL_OUT ("Failed to unsetenv FLUX_F58_FORCE_ASCII");
+#endif
 
     ok (fluid_encode (buf, 1, 1, type) < 0 && errno == EOVERFLOW,
         "fluid_encode (buf, 1, 1, F58) returns EOVERFLOW");
```

grondo:
Ah, the reason for the change is that `str.h` includes `config.h`:
https://github.com/flux-framework/flux-core/blob/1a25d9af1ab4f33c39c598e78346f243e8991e39/src/common/libccan/ccan/str/str.h#L1-L4

So that "fixed" the test and uncovered some tests that should have been conditional.

garlick:
Thanks for finding that issue @grondo!  Just pushed a version with your suggested changes.

grondo:
Looks like Github Actions had an issue for a bit his morning. I tried restarting the "failed" builds here FYI

garlick:
Hmm, still failing.  I rebased on current master and forced a push to try again.

grondo (approved):
LGTM!

PR #5166, opened 2023-05-10
testsuite: skip tests that expect COLUMNS to be inherited across calls to flux(1) when that isn't true

grondo:
On some systems, likely those with bash >= 5, tests that make use of a `COLUMNS` environment variable to override the detected terminal width were failing when the test was run with `-d -v`. It isn't obvious why the failures only occur with the
debug and verbose sharness flags, but the cause seems to be invocation of the `flux` libtool wrapper script (which uses `#!/bin/bash`), which then causes the `COLUMNS` variable to be reset.

There isn't a clear workaround for this issue, so instead set a prereq if `COLUMNS` does propagate through a call to `flux`, and use that prereq to skip a couple tests when that isn't true.

### DIFF
--- a/t/sharness.d/flux-sharness.sh
+++ b/t/sharness.d/flux-sharness.sh
@@ -292,6 +292,23 @@ test_on_rank() {
     flux exec --rank=${ranks} "$@"
 }
 
+#  Note: Some versions of bash may cause the `flux` libtool wrapper script
+#  to reset the COLUMNS shell variable even if it is explicitly set for
+#  for testing purposes. This causes tests that check for output truncation
+#  based on COLUMNS to erroneously fail. (Note: this only seems to be the
+#  case when tests are run with --debug --verbose for unknown reasons.
+#
+#  Add a script for tests that use COLUMNS to check if the variable will
+#  be preserved across an invovation of flux(1) so they may set a prereq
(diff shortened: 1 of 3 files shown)


t/sharness.d/flux-sharness.sh
```diff
@@ -292,6 +292,23 @@ test_on_rank() {
     flux exec --rank=${ranks} "$@"
 }
```
chu11, line 294: commit message typo "require"

t/sharness.d/flux-sharness.sh
```diff
@@ -292,6 +292,23 @@ test_on_rank() {
...
 
+#  Note: Some versions of bash may cause the `flux` libtool wrapper script
+#  to reset the COLUMNS shell variable even if it is explicitly set for
+#  for testing purposes. This causes tests that check for output truncation
+#  based on COLUMNS to erroneously fail. (Note: this only seems to be the
+#  case when tests are run with --debug --verbose for unknown reasons.
```
chu11, line 299: need end parens

chu11 (approved):
LGTM except minor nits

grondo:
Fixed the two blunders caught by @chu11 (thanks!) and pushed the result. Setting MWP.

PR #5169, opened 2023-05-11
python: improve handling of `!d` conversion specifier in output formats

grondo:
This PR fixes a few annoyances when using the `!d` conversion specifier in output formats as discussed over in #5055 

 - A `!d` converted timestamp now renders as an empty string for a timestamp of `0.0`
 - The `h` suffix can now be used with `!d` to replace the empty timestamp with a hyphen `-`
 - Currently `!d` used without a format returns an empty string, this PR adds a default format so that a ISO 8601 string is returned instead.

I've also added a set of unit tests for the affected `UtilDatetime` class to ensure the above is working.

### DIFF
--- a/t/t2800-jobs-cmd.t
+++ b/t/t2800-jobs-cmd.t
@@ -1008,7 +1008,7 @@ test_expect_success 'flux-jobs emits empty string for special case t_estimate' '
 	flux jobs -no "${fmt}" >t_estimate_annotations.out 2>&1 &&
 	test_debug "cat t_estimate_annotations.out" &&
 	for i in `seq 1 $(state_count active)`; do
-		echo ",00:00,,,,-,-,-" >> t_estimate_annotations.exp
+		echo ",,,,,-,-,-" >> t_estimate_annotations.exp
 	done &&
 	test_cmp t_estimate_annotations.out t_estimate_annotations.exp
 '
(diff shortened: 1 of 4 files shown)


t/t2800-jobs-cmd.t
```diff
@@ -937,9 +937,15 @@ test_expect_success 'flux-jobs --format={expiration!d:%FT%T::>20.20} works' '
...
+	cat expiration.in | \
+	    flux jobs --from-stdin -o "{expiration!d:%b%d %R::>20.20h}" \
+	    >exp-fmth.out &&
 	test_debug "cat exp-fmt.out" &&
 	grep "          EXPIRATION" exp-fmt.out &&
-	grep "         $(date --date=@${exp} +%b%d\ %R)" exp-fmt.out
+	grep "         $(date --date=@${exp} +%b%d\ %R)" exp-fmt.out &&
+	test_debug "cat exp-fmt.out" &&
+	grep "          EXPIRATION" exp-fmth.out &&
+	grep "         $(date --date=@${exp} +%b%d\ %R)" exp-fmth.out
```
chu11, line 948: i think we need to add one test for `!d` without any format being specified `{expiration!d}`, perhaps in the one above `test_expect_success 'flux-jobs --format={expiration!d:%FT%T},{t_remaining!H} works' '`

chu11 (approved):
LGTM, just one note below on probably one more test to add.

t/t2800-jobs-cmd.t
grondo: Ok, added a separate test for that and force pushed the result.

grondo:
Since this is approved I'll set MWP.

PR #5171, opened 2023-05-12
support base256 (emoji) jobid encoding

grondo:
This PR adds support for the so-called [base256](https://github.com/Equim-chan/base256) encoding for FLUIDs and thus Flux jobids. Base256 encodes bytes of input to emoji using two tables, described further at the link above.

A `FLUID_STRING_BASE256` encoding is introduced which is adapted from base256, but adds a special prefix character of  🇫  (U+1F1EB) as proposed by @garlick, and much of the code was changed to adhere to flux-core calling convention and coding standards.

The end result is support for an `id.emoji` field in output formats, support for standard base256 input in `flux_jobid_parse(3)`, and an `emoji` type in `flux job id --to`:

e.g.
```console
$ src/cmd/flux jobs -ac 10 -no {id.emoji}
🇫😀🐨😀🌺🚣🍼🏌🐨
🇫😀🐨😀🎂🐰🍂🏌🐨
🇫😀🐨😀🌎😀🍁🏌🐨
🇫😀🐨😀🎪💓🚖💇🐨
🇫😀🐨😀🎠💅🚖💇🐨
🇫😀🐨😀💺🤘🚖💇🐨
🇫😀🐨😀🍤💁🚖💇🐨
🇫😀🐨😀🚄😢🚖💇🐨
🇫😀🐨😀🚗🐕🚕💇🐨
🇫😀🐨😀🍭👈🚕💇🐨
```

```console
$ src/cmd/flux jobs -ac 10 -no {id.emoji} | src/cmd/flux job id --to=f58
ƒ5iWbyUPpQF
ƒ5i1UAPAaYo
ƒ5hzUrkGeyu
ƒ5K8DFzpK2P
ƒ5K8C8Ta4UF
ƒ5K8AuopHZ9
ƒ5K86gPuAvT
ƒ5K7yx2UHAF
ƒ5K7pbR18xF
ƒ5K7fJP6GCX
```


src/common/libutil/Makefile.am
```diff
@@ -98,7 +98,9 @@ libutil_la_SOURCES = \
 	slice.c \
```
garlick, line 98: commit message for "libutil: add base256 ..."

s/distict/distinct/

src/common/libutil/base256.h
```diff
@@ -0,0 +1,52 @@
...
+#endif
+
+/*  Decode string `data` in base256 into result `buf` of size `buflen`,
+ *  which should be at least size BASE256_DECODED_SIZE(strlen(in))
+ *
+ *  Returns 0 on success, -1 on error with errno set:
+ *  EINVAL: invalid arguments
+ *  ENOENT: an invalid character appears in data
+ */
+extern int base256_decode (void *buf,
```
garlick, line 32: AFAIK `extern` has no effect here and probably should be dropped for consistency wtih the rest of flux.

src/common/libutil/test/base256.c
```diff
@@ -0,0 +1,110 @@
...
+
+struct b256_test {
+    uint64_t data;
+    const char *result;
+} b256_tests[] =  {
+    { 0,                 "🇫😀🐨😀🐨😀🐨😀🐨" },
+    { 1,                 "🇫😁🐨😀🐨😀🐨😀🐨" },
+    { 1234,              "🇫👢🐔😀🐨😀🐨😀🐨" },
+    { 12342435,          "🇫👅🍳🗨🐨😀🐨😀🐨" },
+    { 21900760568561664, "🇫😀🐨😀🌳👣🚘😹🐨" },
```
garlick, line 35: F58 does not populate significant digits if they are zero.  Any thoughts on whether that would be a good idea here?
```
$ flux job id --to=f58 1
ƒ2
```

garlick (approved):
LGTM!  Ha ha!  Just minor comments.

src/common/libutil/base256.h
grondo: Oops, sorry, I probably missed some things when "converting" the original code ...

src/common/libutil/test/base256.c
grondo: I was thinking about that too! I don't see why not, but I didn't make an attempt yet...

grondo: I was actually thinking about that for the `words` mnemonic encoding as well. That might  be nice to shorten those too, but I haven't looked into what would be involved.

src/common/libutil/base256.h
```diff
@@ -0,0 +1,52 @@
+/************************************************************\
+ * Copyright 2016 Lawrence Livermore National Security, LLC
```
chu11, line 2: 2023

src/common/libutil/base256.h
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+#ifndef UTIL_BASE256_H
+#define UTIL_BASE256_H
+
+#include <stdbool.h>
+
+#define BASE256_ENCODED_SIZE(x) ((x * 4) + 4 + 1)
```
chu11, line 16: perhaps a comment to explain length, which i assume `🇫` in front and NUL char

src/common/libutil/base256.c
```diff
@@ -0,0 +1,149 @@
...
+                    int datalen)
+{
+    register unsigned char table = 0;
+    int i = 0;
+    int n = 0;
+
+    if (buf == NULL || data == NULL || buflen <= 0 || datalen < 0) {
+        errno = EINVAL;
+        return -1;
+    }
```
chu11, line 87: should make make sure datalen can hold the encoded amount via `BASE256_ENCODED_SIZE`?  like what if datalen % 4 != 0?  Oh ... and what if buf = 1 and datalen = 1, then the memcpy right below could overflow.

src/common/libutil/base256.c
```diff
@@ -0,0 +1,149 @@
...
+    int in_size;
+    const char *cur;
+    int i = 0;
+    int n = 0;
+
+    if (buf == NULL || in == NULL || buflen <= 0) {
+        errno = EINVAL;
+        return -1;
+    }
+
```
chu11, line 128: similar point about checking buf lens, what if length of `in` was 5, then there will be a `memcmp` of bad memory?

src/common/libutil/base256.c
chu11: oh i now see that you made a comment in the header, I know this is just a util library for a unique purpose, maybe its overkill

src/common/libutil/fluid.h
```diff
@@ -23,6 +23,7 @@ typedef enum {
     FLUID_STRING_DOTHEX = 1,    // x.x.x.x
     FLUID_STRING_MNEMONIC = 2,  // mnemonicode x-x-x--x-x-x
     FLUID_STRING_F58 = 3,       // FLUID base58 enc: ƒXXXX or fXXXX
+    FLUID_STRING_BASE256 = 4,   // FLUID base256 enc: emoji
```
chu11, line 26: do we need rfc19 update?

chu11: FWIW, I think we do ... emojis in the rfcs seems like something we need :-)

grondo:
Actually I think I'll close this PR. When actually looking at the base256 implementation I think we can do a little better with something else (maybe using the same emoji set)

PR #5173, opened 2023-05-16
pre-commit: increase vermin version

chu11:
In #5055 some code that was only supported in Python 3.11 did not get caught by the vermin check, which surprised me.  It ended up it was caught in a newer version of vermin.

So lets increase the version and add a `novermin` comment for one place that needs it.  Also add a comment to the place that needs the novermin override.

### DIFF
--- a/src/bindings/python/flux/util.py
+++ b/src/bindings/python/flux/util.py
@@ -30,8 +30,10 @@ from typing import Mapping
 
 import yaml
 
+# tomllib added to standard library in Python 3.11
+# flux-core minimum is Python 3.6.
 try:
-    import tomllib
+    import tomllib  # novermin
 except ModuleNotFoundError:
     from flux.utils import tomli as tomllib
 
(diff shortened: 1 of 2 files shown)


src/bindings/python/flux/util.py
```diff
@@ -30,8 +30,10 @@
 
 import yaml
 
+# tomllib added to standard library in Python 3.10
```
grondo, line 33:
```suggestion
# tomllib added to standard library in Python 3.11
```

grondo (approved):
Great, thanks! Just one nit in a comment, I don't think `tomllib` was added until Python 3.11...

src/bindings/python/flux/util.py
chu11: yeah, i got conflicting information about 3.10 vs 3.11.  Seems we got it wrong in this prior commit too :-(  6298a362324f03700362e44990ab212f12a5374c.  But googling it does seem it is 3.11.

chu11:
thanks, setting MWP.

src/bindings/python/flux/util.py
grondo: Yeah, oops!

PR #5179, opened 2023-05-17
flux-jobs: support W presentation type, update cute output format

chu11:
Problem: `flux-jobs(1)` cute output does not throw enough emojis in your face.

Solution: Update `flux-jobs` `cute` format to use `id.emoji` over `id.f58`.  Support new `W` presentation adjuster to make the output prettier.

image


chu11:
re-pushed, fix spelling and linting issues found by CI.

grondo:
FYI, sadly this doesn't quite work for my font combination I guess 🙁 
image

chu11:
doh!  i wonder if this is even solvable within the framework of how we use output formats.  I suppose in some worst case it could loop through all potential outputs, see what the longest emoji length is, then adjust field width .... ugh

grondo:
Yeah, that's my worry as well. I guess we do already loop through all outputs to check for empty fields to handle `?:` :shrug: If that works here maybe its ok? Meh, doesn't seem worth our time at this point.

chu11:
After working on #7053, I was like "hey, would '+:' help work on this old emoji output PR?" (it wasn't around when I first wrote it)

Had to do some small tweaks, but it seems good!

Screenshot from 2025-09-16 13-53-15

The only gotcha is use of the `wcwidth` python package, so we'd require yet another python dependency.  It appears to be fairly common and was intalled on TOSS clusters by default (but not on my desktop).  I did not add it to the `scripts/install-*` scripts yet.

Edit: oh it just occurred to me the STATUS emoji is not lined up correctly in the above picture.  It's b/c this PR is not rebased on PR #7053.  The width has not been corrected for the header.  I temporarily widened the width for this next screenshot.

Screenshot from 2025-09-16 14-02-17


Edit2: Ok, nevermind, I rebased this on #7053.  It's probably necessary for this to all look good.

Screenshot from 2025-09-16 14-43-10


Edit3: ugh, i guess the python dependency isn't in the images, so all CI fails

Issue #5175, opened 2023-05-17
flux-job:  flux_buffer_read_watcher_create: Invalid argument


grondo:
A user is getting this error in their nightly regression tests which run simply: `/usr/bin/flux run -n 1 ./tstsend_is`

```
flux-job:  flux_buffer_read_watcher_create: Invalid argument
```

This appears to be in `flux job attach` when trying to create the buffered read watcher for stdin. I have asked for more detail on the execution environment. At first I thought maybe stdin was closed, but there are some preceding calls to `fcntl(2)` that I presume would have generated errors if this were the case?

I did try running with stdin closed and get a hang, not an error (so this is a separate problem perhaps)

```console
$ flux run hostname <&- 
pi0
# and hang
```

grondo:
After a bit of investigation in #5177, I don't think we're any closer to understanding what could cause this error from `flux_buffer_read_watcher_create()` - the function appears to return `EINVAL` in the following scenarios:

```c
    if (fd < 0) {
        errno = EINVAL;
        return NULL;
    }
    if ((fd_flags = fd_get_flags (fd)) < 0)
        return NULL;
    if (!(fd_flags & O_NONBLOCK)) {
        errno = EINVAL;
        return NULL;
    }
```

i.e.

 - If `fd < 0`: but `STDIN_FILENO` is being used explicitly here, so that can't be the case
 - If `fcntl(F_GETFL)` fails:  but the same `fcntl(2)` calls are made in the calling function and they do not fail, plus `fcntl(2)` is not documented to return `EINVAL` from `F_GETFL` (instead we should see `EBADF` for a bad file descriptor)
 - if the `O_NONBLOCK` flag is not set on the file descriptor, but the calling function set `O_NONBLOCK` without error.

I don't know if there is a way for `fcntl (STDIN_FILENO, F_SETFL, stdin_flags | O_NONBLOCK)` to succeed without the flag actually being set. That could explain this error, but it seems very unlikely.

grondo:
Closing this one since we haven't heard of a reproducer.

PR #5183, opened 2023-05-18
flux-job: ignore stdin instead of aborting when unable to create stdin watcher

grondo:
This is a possible workaround for #5175 since the cause of that issue isn't really understood. Instead of aborting `flux job attach` when there is an EINVAL error creating the buffered stdin watcher, just log a warning that stdin will be ignored.


chu11 (approved):
LGTM

PR #5186, opened 2023-05-19
job-exec: provide IMP exec helper when possible and use stdin/out for shell exec protocol

### LINKED ISSUE #4220: job-exec/sdexec: implement sd-bus based IPC to implement equivalent to FLUX_EXEC_PROTOCOL_FD
With the job-exec/sdexec implementation (PR #4070) there is no equivalent way to add the barrier added by #4155 to solve a number of corner cases.

At present, it seems impossible to solve it via a shared file descriptor like it was solved with the default `exec` job launch.  A more likely solution would be to use `sd-bus`'s IPC mechanisms.  Have `sdexec` setup a "service" / "object" and have the job-shell communicate back to it.

Some notes for later:

key functions to research to implement:

```
sd_bus_request_name
sd_bus_add_object / sd_bus_add_object_vtable
sd_bus_add_match
```

random example i found online
                                                                                                                                            
https://cpp.hotexamples.com/site/file?hash=0xe0216674f5f8bdea4d4e0cadb97fe202b2a07e88e2ba55b6ae6db30fcd61bda7&fullName=openbmc-master/meta-phosphor/common/recipes-phosphor/obmc-phosphor-example-sdbus/obmc-phosphor-example-sdbus.c&project=bradbishop/openbmc

grondo:
This PR adds an IMP exec helper utility that lets the IMP fetch its input from the enclosing instance instead of requiring the input on stdin. This becomes the default behavior if flux-security >= 0.9.0 is detected (the version in which support for an IMP helper was introduced). 

This should improve scalability of mulituser job launch somewhat, but the main purpose is to free up stdin/out of the job shell to be used to implement the exec barrier protocol instead of a separate socketpair. This will be necessary to support launching multi-node jobs under systemd, under which we can't provide a separate set of fds for the protocol.

Thus, when the IMP helper is used, use stdin/out fds for the exec protocol.

Closes #4220

### DIFF
--- a/src/cmd/Makefile.am
+++ b/src/cmd/Makefile.am
@@ -112,7 +112,8 @@ dist_fluxcmd_SCRIPTS = \
 	flux-pstree.py \
 	flux-pgrep.py \
 	flux-queue.py \
-	flux-cancel.py
+	flux-cancel.py \
+	flux-imp-exec-helper
 
 fluxcmd_PROGRAMS = \
 	flux-terminus \
--- a/src/cmd/flux-imp-exec-helper
+++ b/src/cmd/flux-imp-exec-helper
@@ -1,0 +1,20 @@
+#!/bin/sh
+##############################################################
+# Copyright 2023 Lawrence Livermore National Security, LLC
+# (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+#
+# This file is part of the Flux resource manager framework.
+# For details, see https://github.com/flux-framework.
+#
+# SPDX-License-Identifier: LGPL-3.0
+##############################################################
+#
+# Helper for flux-imp exec functionality.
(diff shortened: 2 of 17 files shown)


grondo:
Pushed a fix here for the valgrind suppression file, I was getting a failure on Debian 11 aarch64 because the existing hwloc suppression wasn't general enough. I can move this to a separate PR if desired.

garlick (approved):
LGTM!

FWIW I installed this along with flux-security-0.9.0 on my test cluster and it works as advertised.  I guess that's worth less these days since you are probably doing the same thing.

Also valgrind is now passing on my aarch64 system, which hasn't happened in a while.  Thanks!

grondo:
Thanks! I'll set MWP

PR #5191, opened 2023-05-22
doc: fix python example

### LINKED ISSUE #5190: Documentation for "Asynchronous event-loop interface" is confusing
The documentation lists the following code snippet: 
```
import flux
import flux.job

def submit_cb(fut, flux_handle):
        # when this callback fires, the jobid will be ready
        jobid = fut.get_id()
        # Create a future representing the result of the job
        result_fut = flux.job.result_async(flux_handle, jobid)
        # attach a callback to fire when the job finishes
        result_fut.then(result_cb)

def result_cb(fut):
        print(fut.get_info())

flux_handle = flux.Flux()
jobspec = flux.job.JobspecV1.from_command(["/bin/true"])
for _ in range(5):
        # submit 5 futures and attach callbacks to each one
        submit_future = flux.job.submit_async(f, jobspec)
        submit_future.then(submit_cb, flux_handle)
# enter the flux event loop (the 'reactor') to trigger the callbacks
# once the futures complete
flux_handle.reactor_run()
```

This is confusing as `f` is not defined in this example.

grondo:
Problem: A job submission Python example in the docs has an error, so it won't run, making it a poor example. It also prints the raw JobInfo object, which isn't very demonstrative.

Update the example to fix the error and print something useful from the returned JobInfo result object.

Fixes #5190

### DIFF
--- a/doc/python/job_submission.rst
+++ b/doc/python/job_submission.rst
@@ -216,13 +216,15 @@ as futures complete.
 		result_fut.then(result_cb)
 
 	def result_cb(fut):
-		print(fut.get_info())
+		job = fut.get_info()
+		result = job.result.lower()
+		print(f"{job.id}: {result} with returncode {job.returncode}")
 
 	flux_handle = flux.Flux()
 	jobspec = flux.job.JobspecV1.from_command(["/bin/true"])
 	for _ in range(5):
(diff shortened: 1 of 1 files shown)


chu11 (approved):
LGTM

grondo:
Thanks! Setting MWP.

Issue #2864, opened 2020-03-20
job-info: list-id service wait for specific state


chu11:
Discussed offline and in #2862, in order to eliminate the need for "wait and check" loops in tests, the `list id` service could not only wait for an jobid to be legal, but it could also wait for a jobid to reach a specific state.  Eliminating the need for all of these wait-and-check loops.

garlick:
Couple of more thoughts on this  idea (complications I guess):
- The sync "event" wouldn't be associated with any job but journal events are wrapped in a object that associates them with a job id.  To keep it simple maybe we could just associate the sync event with FLUX_JOBID_ANY and handle that job ID specially in the journal consumers?
- This I'm not sure if - is it sufficient for job-list to respond immediately when the sync event is received and the events before it are applied to the in-memory job info, or would it need to wait until some async fetches of KVS info triggered by those events to complete?

grondo:
> is it sufficient for job-list to respond immediately when the sync event is received and the events before it are applied to the in-memory job info, or would it need to wait until some async fetches of KVS info triggered by those events to complete?

Oh yeah, darn. I was loving the simplicity of the journal synchronization marker, but I wonder if you're right - it can't work as simply as I was thinking...

chu11:
(agh, I didn't hit submit on this until after grondo's post)

> This I'm not sure of: is it sufficient for job-list to respond immediately when the sync event is received and the events before it are applied to the in-memory job info, or would it need to wait until some async fetches of KVS info triggered by those events to complete?

I didn't think about it much, but you're right, this would be a problem.  Internally in job-list there is a data structure for transitioning from one job state to another.  As you surmise, there are KVS lookups and what not that need to be done between job state transitions so that the data is always available before job-list says a job is in a specific job state.  So we'd need some type of "are any list requests waiting for a job id to reach this job state" kinda hash lookup and response, so perhaps the sync cookie idea wouldn't add much.

PR #5194, opened 2023-05-23
testsuite: fix race in flux top tests

chu11:
Problem: On rare occasions some tests in t2801-top-cmd.t fail when running large parallel runs of the testsuite.  The suspicion is that the job-list module has not yet synced up to have knowledge of the jobs that have been submitted in the tests.

Add a function to wait until jobs in t2801-top-cmd.t are known by job-list and have reached the state we want to ensure tests will pass.

### DIFF
--- a/t/t2801-top-cmd.t
+++ b/t/t2801-top-cmd.t
@@ -12,6 +12,26 @@ testssh="${SHARNESS_TEST_SRCDIR}/scripts/tssh"
 
 export FLUX_URI_RESOLVE_LOCAL=t
 
+# To ensure no raciness in tests below, ensure the job-list
+# module knows about submitted jobs in desired states
+JOB_WAIT_ITERS=100
+job_list_wait_state() {
+	id=$1
+	expected=$2
+	local i=0
+	while [ "$(flux jobs -no {state} ${id})" != "${expected}" ] \
+	       && [ $i -lt ${JOB_WAIT_ITERS} ]
+	do
+		sleep 0.1
+		i=$((i + 1))
+	done
+	if [ "$i" -eq "${JOB_WAIT_ITERS}" ]
+	then
+	    return 1
+	fi
+	return 0
+}
+
 test_expect_success 'flux-top -h prints custom usage' '
 	flux top -h 2>usage &&
 	grep "Usage:.*TARGET" usage
@@ -48,7 +68,8 @@ test_expect_success 'flux-top summary shows no jobs initially' '
 	grep "0 failed" nojobs.out
 '
 test_expect_success 'run a test job to completion' '
-	flux submit --wait -n1 flux start /bin/true >jobid
+	flux submit --wait -n1 flux start /bin/true >jobid &&
+	job_list_wait_state $(cat jobid) INACTIVE
 '
 test_expect_success 'flux-top summary shows one completed job' '
 	nnodes=$(flux resource list --format="{nnodes}") &&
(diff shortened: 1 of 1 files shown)


chu11:
re-pushed fixed typo in commit message and && chaining mistake

t/t2801-top-cmd.t
```diff
@@ -12,6 +12,25 @@ testssh="${SHARNESS_TEST_SRCDIR}/scripts/tssh"
 
 export FLUX_URI_RESOLVE_LOCAL=t
 
+# To ensure no raciness in tests below, ensure the job-list
+# module knows about submitted jobs in desired states
+JOB_WAIT_ITERS=100
+job_list_wait_state() {
+	expected=$1
```
grondo, line 19: I'm not sure I see how `job_list_wait_state()` is supposed to work. It looks like it takes one argument (the expected state?) and assumes there is only one active job, but below it seems to be called with two arguments (`jobid` and `state`). Is there just a missing argument (perhaps to `flux jobs`) here?

Also, I'm seeing 
```
./t2801-top-cmd.t: line 23: [: 0: unary operator expected
```
when running the current version of this test.

Also, a more broad question - the eventual consistency issue seems to come up frequently with job-list. It would be nice if there was a way to ensure job-list state is synchronized without this retry loop. If that is easy to implement it might allow a ton of cleanup in the testsuite.

chu11: The answer is I completely flubbed it by not getting an the id passed in.  I have no idea how this worked when I ran it.

Yeah, I have a helper function in `job-list/job-list-helper.sh` but that is unique for the "large setup of jobs" that I do in `t2260-job-list.t` and `t2800-jobs-cmd.t`, so didn't quite work for this case of just a handful of jobs.

Perhaps I can stick this helper in `job-list/job-list-helper.sh` instead.

grondo: I was more thinking of an RPC to job-list that responds when a job or list of jobs has reached a certain state. I haven't thought about how much trouble it would be to implement, but needing these for loops in tests is a little depressing and makes it seem like there's some interface missing. However, if it just more of a pain to add some interface to the module, then this is fine for now...

chu11: @garlick are you thinking something sort of like `flux kvs wait`?  Insert a "version" or sorts into the stream?

garlick: well not a version per se but a "sync cookie" I guess.  Since many operations in the job manager are synchronous with respect to appending events to the journal, one could 
1) do the thing
2) call job manager sync, get back identifier
3) call job-list sync with identifier, get back a response that indicates job-list is caught up to that point in the journal

chu11:
re-pushed with fix for my very dumb mistake.  I decided not to add this function into `job-list/job-list-helper.sh` b/c this uses `flux jobs` instead of `flux job list`.  The reason I use `flux jobs` is b/c I didn't want to check the job state for a specific number vs a string, which is what `flux job list-ids` would have returned to me.

Hopefully can resolve longer out via #2864.

grondo:
any reason not to implement the idea proposed by @garlick now? It seems like it would neatly handle a bunch of testing cases and would be a more satisfying solution to this issue.

chu11:
I mostly updated the PR just because it was done :-) and would temporarily remove the spurious error I see. 
But the error is rare, can give the other approach a shot instead of going with this.

grondo:
Well its a good point that this fixes an error you're seeing and is already done :slightly_smiling_face: 

I'll go ahead and approve and you can decide if this goes in or not.

t/t2801-top-cmd.t
```diff
@@ -12,6 +12,26 @@ testssh="${SHARNESS_TEST_SRCDIR}/scripts/tssh"
...
 
+# To ensure no raciness in tests below, ensure the job-list
+# module knows about submitted jobs in desired states
+JOB_WAIT_ITERS=100
+job_list_wait_state() {
+	id=$1
+	expected=$2
+	local i=0
+	while [ "$(flux jobs -no {state} ${id})" != "${expected}" ] \
+	       && [ $i -lt ${JOB_LIST_WAIT_ITERS} ]
```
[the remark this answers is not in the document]
chu11: aghh ... i'm just flubbing up this PR.  fixed and re-pushed

grondo (approved):
LGTM!

PR #5197, opened 2023-05-23
add capability to run jobs in systemd

garlick:
This is another step towards enabling jobs to run in systemd so that we can limit their resource consumption and eventually be able to restart the flux system instance without killing the workload.  It adds:
- `sdexec` module which adds a subprocess server, compatible with `libsubprocess`, that runs processes as systemd transient units via the `sdbus` module added in #5070 
- `job-exec` support for running jobs with `--setattr system.exec.bulkexec.service=sdexec` to optionally start job shells in systemd

An internal convenience library `libsdexec` is added which hides some of the gory details of managing transient units via the `sdbus` module.  This simplifies the `sdexec` module a bit.  Even though I'm not sure if there will be other users of it, it seemed helpful to me to think of those interfaces as possibly usable elsewhere and to minimize `sdexec` and `libsubprocess` quirks creeping in.

Testing on this has a way to go and due to taking breaks on this several times I can't remember if there are any gaps I've forgotten about!  But anyway, this is a checkpoint that seems to work.

TODO:

- [x] test on actual cluster
- [x] test job kill, with and without imp
- [x] append broker rank to transient unit name prefix to avoid the unit tracker claiming units of other brokers in test
- [x] libsdexec unit tests to at least cover invalid arguments
- [x] verify/document behavior when systemd and/or dbus restarts (audit code for that in sdbus/sdexec)
- [x] test job with large/complex environment (e.g. when env modules present)
- [x] implement `sdexec.list` and use in tests with `rexec.py` test program
- [x] add ability to configure jobs and perilog to use sdexec
- [x] add some method to control debug logging (also in `sdbus`)
- [x] add containment support as prototyped in #4255

### DIFF
--- a/src/common/libsdexec/parse.c
+++ b/src/common/libsdexec/parse.c
@@ -1,0 +1,81 @@
+/************************************************************\
+ * Copyright 2023 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include <errno.h>
+#include <string.h>
+#include <stdlib.h>
+#include <flux/idset.h>
+
+#include "ccan/str/str.h"
+#include "parse.h"
+
+int sdexec_parse_percent (const char *s, double *dp)
+{
+    char *endptr;
+    double d;
+
+    if (!s || !dp)
+        return -1;
+    errno = 0;
+    d = strtod (s, &endptr);
+    if (errno != 0 || endptr == s || strlen (endptr) != 1 || endptr[0] != '%')
+        return -1;
+    if (d < 0 || d > 100)
+        return -1;
+    *dp = d * 1E-2;
+    return 0;
+}
+
+/* Borrow some macro ideas from ccan/bitmap/bitmap.h.  It cannot be used here
+ * because it apparently swaps bytes of internal words on little endian for
+ * portability, which isn't compatible with the byte array requirement of
+ * systemd/dbus.
+ */
+#define BITMAP_NBYTES(nb)           (((nb) + 8 - 1) / 8)
+#define BITMAP_BYTE(bm,n)           (bm)[(n) / 8]
+#define BITMAP_BIT(n)               (1 << ((n) % 8))
+
+int sdexec_parse_bitmap (const char *s, uint8_t **bp, size_t *sp)
+{
+    struct idset *ids;
+    unsigned int id;
+    unsigned int nbits = 0;
+    uint8_t *bitmap = NULL;
+
+    if (!s || !bp || !sp)
+        return -1;
+    if (!(ids = idset_decode (s)))
+        return -1;
+
+    // allocate a bitmap large enough to fit idset
+    if ((id = idset_last (ids)) != IDSET_INVALID_ID)
+        nbits = id + 1;
+    if (nbits > 0) {
+        if (!(bitmap = calloc (1, BITMAP_NBYTES (nbits)))) {
+            idset_destroy (ids);
+            return -1;
+        }
+        id = idset_first (ids);
+        while (id != IDSET_INVALID_ID) {
+            BITMAP_BYTE (bitmap, id) |= BITMAP_BIT (id);
+            id = idset_next (ids, id);
+        }
+        idset_destroy (ids);
+    }
+    *bp = bitmap;
+    *sp = BITMAP_NBYTES (nbits);
+    return 0;
+}
+
+// vi:ts=4 sw=4 expandtab
--- a/src/common/libsdexec/stop.h
+++ b/src/common/libsdexec/stop.h
@@ -1,0 +1,37 @@
+/************************************************************\
+ * Copyright 2023 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+#ifndef _LIBSDEXEC_STOP_H
+#define _LIBSDEXEC_STOP_H
+
+#include <flux/core.h>
+
+/* See https://www.freedesktop.org/wiki/Software/systemd/dbus/
+ * for more info on mode parameter.
+ * mode maybe one of: replace, fail, ignore-dependencies, ignore-requirements.
+ */
+flux_future_t *sdexec_stop_unit (flux_t *h,
+                                 uint32_t rank,
+                                 const char *name,
+                                 const char *mode);
+
+flux_future_t *sdexec_reset_failed_unit (flux_t *h,
+                                         uint32_t rank,
+                                         const char *name);
+
+flux_future_t *sdexec_kill_unit (flux_t *h,
+                                 uint32_t rank,
+                                 const char *name,
+                                 const char *who, // main/control/all
+                                 int signum);
+
+#endif /* !_LIBSDEXEC_STOP_H */
+
+// vi:ts=4 sw=4 expandtab
--- a/src/common/libsdexec/test/channel.c
+++ b/src/common/libsdexec/test/channel.c
@@ -1,0 +1,258 @@
+/************************************************************\
+ * Copyright 2023 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include <string.h>
+#include <jansson.h>
+#include <flux/core.h>
+
+#include "ccan/array_size/array_size.h"
+#include "src/common/libtap/tap.h"
+#include "src/common/libtestutil/util.h"
+#include "src/common/libioencode/ioencode.h"
+#include "channel.h"
+
+bool input_called;
+bool input_eof_set;
+
+void input_cb (flux_reactor_t *r,
+               flux_watcher_t *w,
+               int revents,
+               void *arg)
+{
+    struct channel *ch = arg;
+    const char *name = sdexec_channel_get_name (ch);
+    int fd = sdexec_channel_get_fd (ch);
+    char buf[64];
+    int n;
+
+    n = read (fd, buf, sizeof (buf));
+    if (n < 0) {
+        diag ("%s: read error: %s", name, strerror (errno));
+    }
+    else if (n == 0) {
+        diag ("%s: EOF", name);
+        input_eof_set = true;
+    }
+    else
+        diag ("%s: read %d chars", name, n);
+    input_called = true;
+}
+
+void test_input (void)
+{
+    flux_t *h;
+    struct channel *ch;
+    flux_watcher_t *w;
+    int fd;
+    json_t *io;
+    json_t *io_eof;
+
+    if (!(h = loopback_create (0)))
+        BAIL_OUT ("could not create loopback flux_t handle for testing");
+    if (flux_attr_set_cacheonly (h, "rank", "0") < 0)
+        BAIL_OUT ("could not set rank for testing");
+    ch = sdexec_channel_create_input (h, "in");
+    ok (ch != NULL,
+        "sdexec_channel_create_input works");
+    fd = sdexec_channel_get_fd (ch);
+    ok (fd >= 0,
+        "sdexec_channel_get_fd works");
+    w = flux_fd_watcher_create (flux_get_reactor (h),
+                                fd,
+                                FLUX_POLLIN,
+                                input_cb,
+                                ch);
+    if (!w)
+        BAIL_OUT ("could not create fd watcher");
+    if (!(io = ioencode ("foo", "0", "hello", 6, false)))
+        BAIL_OUT ("could not create json io object");
+    if (!(io_eof = ioencode ("foo", "0", NULL, 0, true)))
+        BAIL_OUT ("could not create json io_eof object");
+    flux_watcher_start (w);
+
+    input_called = false;
+    input_eof_set = false;
+    ok (sdexec_channel_write (ch, io) == 0,
+        "sdexec_channel_write works");
+    ok (flux_reactor_run (flux_get_reactor (h), FLUX_REACTOR_ONCE) >= 0,
+        "flux_reactor_run ran ONCE");
+    ok (input_called == true,
+        "input callback was called");
+    ok (input_eof_set == false,
+        "eof was not set");
+
+    input_called = false;
+    input_eof_set = false;
+    ok (sdexec_channel_write (ch, io_eof) == 0,
+        "sdexec_channel_write eof works");
+    ok (flux_reactor_run (flux_get_reactor (h), FLUX_REACTOR_ONCE) >= 0,
+        "flux_reactor_run ran ONCE");
+    ok (input_called == true,
+        "input callback was called");
+    ok (input_eof_set == true,
+        "eof was set");
+
+    json_decref (io_eof);
+    json_decref (io);
+    flux_watcher_destroy (w);
+    sdexec_channel_destroy (ch);
+    flux_close (h);
+}
+
+bool error_called;
+bool output_called;
+bool output_eof_set;
+
+void output_cb (struct channel *ch, json_t *io, void *arg)
+{
+    const char *stream;
+    int len;
+
+    if (iodecode (io, &stream, NULL, NULL, &len, &output_eof_set) < 0) {
+        diag ("%s: idoecode error: %s",
+              sdexec_channel_get_name (ch),
+              strerror (errno));
+    }
+    else {
+        diag ("%s output: stream=%s len=%d eof=%s",
+              sdexec_channel_get_name (ch),
+              stream,
+              len,
+              output_eof_set ? "true" : "false");
+    }
+    output_called = true;
+}
+
+void error_cb (struct channel *ch, flux_error_t *error, void *arg)
+{
+    diag ("%s error: %s",
+          sdexec_channel_get_name (ch),
+          error->text);
+    error_called = true;
+}
+
+void test_output (void)
+{
+    flux_t *h;
+    struct channel *ch;
+    int fd;
+
+    if (!(h = loopback_create (0)))
+        BAIL_OUT ("could not create loopback flux_t handle for testing");
+    if (flux_attr_set_cacheonly (h, "rank", "0") < 0)
+        BAIL_OUT ("could not set rank for testing");
+    ch = sdexec_channel_create_output (h, "out", output_cb, error_cb, NULL);
+    ok (ch != NULL,
+        "sdexec_channel_crate_output works");
+    sdexec_channel_start_output (ch);
+    ok (true, "sdexec_channel_start_output called");
+
+    fd = sdexec_channel_get_fd (ch);
+    ok (fd >= 0,
+        "sdexec_channel_get_fd works");
+
+    output_called = false;
+    error_called = false;
+    output_eof_set = false;
+    ok (write (fd, "hello", 6) == 6,
+        "wrote 'hello' from unit");
+    ok (flux_reactor_run (flux_get_reactor (h), FLUX_REACTOR_ONCE) >= 0,
+        "flux_reactor_run ran ONCE");
+    ok (output_called == true,
+        "output callback was called");
+    ok (output_eof_set == false,
+        "eof was not set");
+    ok (error_called == false,
+        "error callback was not called");
+
+    output_called = false;
+    error_called = false;
+    output_eof_set = false;
+    sdexec_channel_close_fd (ch);
+    ok (true, "sdexec_channel_close_fd called");
+    ok (flux_reactor_run (flux_get_reactor (h), FLUX_REACTOR_ONCE) >= 0,
+        "flux_reactor_run ran ONCE");
+    ok (output_called == true,
+        "output callback was called");
+    ok (output_eof_set == true,
+        "eof was set");
+    ok (error_called == false,
+        "error callback was not called");
+
+    sdexec_channel_destroy (ch);
+    flux_close (h);
+}
+
+void test_inval (void)
+{
+    struct channel *ch;
+    flux_t *h;
+    json_t *io;
+
+    if (!(h = loopback_create (0)))
+        BAIL_OUT ("could not create loopback flux_t handle for testing");
+    if (!(io = ioencode ("foo", "0", NULL, 0, true)))
+        BAIL_OUT ("could not create json io object");
+
+    errno = 0;
+    ch = sdexec_channel_create_output (NULL, "foo", output_cb, error_cb, NULL);
+    ok (ch == NULL && errno == EINVAL,
+        "sdexec_channel_create_output h=NULL fails with EINVAL");
+    errno = 0;
+    ch = sdexec_channel_create_output (h, NULL, output_cb, error_cb, NULL);
+    ok (ch == NULL && errno == EINVAL,
+        "sdexec_channel_create_output name=NULL fails with EINVAL");
+
+    errno = 0;
+    ch = sdexec_channel_create_input (NULL, "foo");
+    ok (ch == NULL && errno == EINVAL,
+        "sdexec_channel_create_input h=NULL fails with EINVAL");
+    errno = 0;
+    ch = sdexec_channel_create_input (h, NULL);
+    ok (ch == NULL && errno == EINVAL,
+        "sdexec_channel_create_input name=NULL fails with EINVAL");
+
+    errno = 0;
+    ok (sdexec_channel_write (NULL, io) < 0 && errno == EINVAL,
+        "sdexec_channel_write ch=NULL fails with EINVAL");
+
+    ok (sdexec_channel_get_fd (NULL) == -1,
+        "sdexec_channel_get_fd ch=NULL returns -1");
+
+    ok (sdexec_channel_get_name (NULL) != NULL,
+        "sdexec_channel_get_name ch=NULL returns non-NULL");
+
+    lives_ok ({sdexec_channel_start_output (NULL);},
+              "sdexec_channel_start_output ch=NULL doesn't crash");
+    lives_ok ({sdexec_channel_close_fd (NULL);},
+              "sdexec_channel_close_fd ch=NULL doesn't crash");
+    lives_ok ({sdexec_channel_destroy (NULL);},
+              "sdexec_channel_destroy ch=NULL doesn't crash");
+
+    json_decref (io);
+    flux_close (h);
+}
+
+int main (int ac, char *av[])
+{
+    plan (NO_PLAN);
+
+    test_input ();
+    test_output ();
+    test_inval ();
+
+    done_testing ();
+}
+
+// vi: ts=4 sw=4 expandtab
--- a/src/common/libsdexec/unit.c
+++ b/src/common/libsdexec/unit.c
@@ -1,0 +1,260 @@
+/************************************************************\
+ * Copyright 2023 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* unit.c - translate unit property updates to unit object changes
+ */
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+#include <sys/types.h>
+#include <sys/wait.h>
+#include <flux/core.h>
+
+#include "src/common/libutil/errprintf.h"
+#include "src/common/libutil/aux.h"
+#include "ccan/str/str.h"
+#include "ccan/array_size/array_size.h"
+
+#include "property.h"
+#include "list.h"
+#include "unit.h"
+
+struct unit {
+    char *path;
+    sdexec_state_t state;
+    sdexec_substate_t substate;
+    pid_t exec_main_pid;
+    int exec_main_code;
+    int exec_main_status;
+    bool exec_main_pid_is_set;
+    bool exec_main_status_is_set;
+
+    struct aux_item *aux;
+};
+
+void sdexec_unit_destroy (struct unit *unit)
+{
+    if (unit) {
+        int saved_errno = errno;
+        aux_destroy (&unit->aux);
+        free (unit->path);
+        free (unit);
+        errno = saved_errno;
+    }
+}
+
+void *sdexec_unit_aux_get (struct unit *unit, const char *name)
+{
+    if (!unit) {
+        errno = EINVAL;
+        return NULL;
+    }
+    return aux_get (unit->aux, name);
+}
+
+int sdexec_unit_aux_set (struct unit *unit,
+                         const char *name,
+                         void *aux,
+                         flux_free_f destroy)
+{
+    if (!unit) {
+        errno = EINVAL;
+        return -1;
+    }
+    return aux_set (&unit->aux, name, aux, destroy);
+}
+
+const char *sdexec_unit_name (struct unit *unit)
+{
+    if (unit) {
+        const char *cp;
+        if ((cp = strrchr (unit->path, '/')))
+            return cp + 1;
+        return unit->path;
+    }
+    return "internal error: unit is null";
+}
+
+const char *sdexec_unit_path (struct unit *unit)
+{
+    if (unit)
+        return unit->path;
+    return "internal error: unit is null";
+}
+
+pid_t sdexec_unit_pid (struct unit *unit)
+{
+    if (unit && unit->exec_main_pid_is_set)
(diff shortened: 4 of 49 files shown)


grondo:
I've installed this on my home cluster. I know this is a WIP so none of this is unexpected, but some early observations:
 - if `sdexec` module is not loaded, the exception a job gets is 
   ```
   0.049s: job.exception type=exec severity=0 job shell exec error on broker pi0 (rank 1): /usr/lib/aarch64-linux-gnu/flux/flux-imp: Function not implemented
   ```
   If possible, it would be nice to translate this error to something like `no sdexec service loaded` (low priority)
 - I then tried to load the `sdexec` module, which apparently succeeded (`flux module load` exited with zero exit code), but the module was not actually loaded due to:
   ```
   [May26 14:16] broker[1]: insmod sdexec
   [  +0.001998] sdexec[1]: sdbus.subscribe: No service matching sdbus.subscribe is registered
   [  +0.002056] sdexec[1]: reactor exited abnormally: Resource temporarily unavailable
   [  +0.002100] sdexec[1]: module exiting abnormally
   [  +0.003133] broker[1]: module sdexec exited
   ```
   Could `sdexec` wait synchronously for the first response from `sdbus` to ensure the module is successfully loaded before `flux module load` returns?

After loading first `sdbus` then `sdexec`, I'm not getting hangs trying to run jobs on my cluster with `--setattr=exec.bulkexec.service=sdexec`. The only hint in the logs is:

```
[May26 14:18] broker[1]: insmod sdbus
[  +0.002336] sdbus[1]: connect: No such file or directory
[May26 14:19] sdbus[1]: connect: No such file or directory
[  +2.031491] broker[1]: insmod sdexec
[  +4.002437] sdbus[1]: connect: No such file or directory
[  +8.772299] sdexec[1]: preparing flux-e3a3f109.service
[  +8.772639] sdexec[1]: start flux-e3a3f109.service
[ +10.004261] sdbus[1]: connect: No such file or directory
[ +18.012806] sdbus[1]: connect: No such file or directory
[ +28.023362] sdbus[1]: connect: No such file or directory
[ +40.035943] sdbus[1]: connect: No such file or directory
[ +54.050551] sdbus[1]: connect: No such file or directory
[May26 14:20] sdbus[1]: connect: No such file or directory
[ +18.018547] sdbus[1]: connect: No such file or directory
[ +38.039093] sdbus[1]: connect: No such file or directory
[May26 14:21] sdbus[1]: connect: No such file or directory
```

Is my system instance missing some configuration or maybe one of the requisite environment variables?

Trying to cancel the jobs gets it stuck in CLEANUP.

garlick:
Great feedback there @grondo.  I'll have to work through those error cases.

Is the debian `dbus-user-session` package installed on all ranks?  I wonder if that's what might be causing the sdbus `connect: No such file or directory`.    It might be `/run/user/<flux uid>/bus` that is missing.

All this error reporting needs to be improved so thanks for going down that path! (even if involuntarily!)

grondo:
> Is the debian dbus-user-session package installed on all ranks? I wonder if that's what might be causing the sdbus connect: No such file or directory. It might be `/run/user/<flux uid>/bus` that is missing.

The package is installed:
```console
$ dpkg-query -W dbus-user-session
dbus-user-session	1.12.24-0+deb11u1
```
But your assumption about the missing bus seems to be correct. I'll have to see how I get that created, the user systemd instance seems to be running.
```console
$ id flux
uid=500(flux) gid=500(flux) groups=500(flux)
$ sudo ls -l /run/user/500
total 0
drwx------ 2 flux flux 140 Apr  8 14:58 gnupg
drwxr-xr-x 4 flux flux 120 Apr  8 14:58 systemd

```

grondo:
Hm, I had to restart the flux user systemd session:
```console
$ sudo systemctl restart user@500
$ sudo ls -l /run/user/500
total 0
srw-rw-rw- 1 flux flux   0 May 27 10:58 bus
drwx------ 2 flux flux 140 May 27 10:58 gnupg
drwxr-xr-x 4 flux flux 120 May 27 10:58 systemd
```

garlick:
Just pushed a few changes based on your comments
- improved error message if sdbus socket cannot be opened
- ensure that sdexec module load fails if sdbus is not loaded
- (since that made me think a bit) send sdbus rpcs to the local rank not ANY which could go upstream if sdbus is not loaded

More work to do here including trying to improve the exception reason noted above.

src/common/libsdexec/stop.h
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+#ifndef _LIBSDEXEC_STOP_H
+#define _LIBSDEXEC_STOP_H
+
+#include <flux/core.h>
+
+flux_future_t *sdexec_stop_unit (flux_t *h, const char *name, const char *mode);
```
chu11: comment for modes would be good

src/common/libsdexec/unit.c
```diff
@@ -0,0 +1,574 @@
...
+    struct unit_monitor *um = unit->monitor;
+    json_t *dict;
+
+    if (!(dict = sdexec_property_get_all_dict (f))) {
+        flux_log (um->h, LOG_ERR, "%s", future_strerror (f, errno));
+        if (errno == EAGAIN)
+            unit_monitor_recover (um);
+        else {
+            flux_reactor_stop_error (flux_get_reactor (um->h));
+        }
```
chu11, line 463: nit: braces around single line after branch (didn't do above and below)

chu11 (commented):
just doing light review to get my bearings on all the code, just a bunch of nits so far through the `libsdexec` commit

src/common/libsdexec/unit.c
```diff
@@ -0,0 +1,574 @@
...
+
+struct unit {
+    char *path;
+    unit_state_t state;
+    unit_substate_t substate;
+    pid_t exec_main_pid;
+    int exec_main_code;
+    int exec_main_status;
+    uint8_t exec_main_pid_is_set:1;
+    uint8_t exec_main_status_is_set:1;
```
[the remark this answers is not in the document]
garlick: I guess it's saving us sizeof (bool) * 2 - sizeof (uint8_t) per unit but bah, I'll just use bool.  There were more flags earlier :-)

garlick:
Thanks @chu11!  Good feedback.

The unit code is getting some refactoring as noted above so don't spend too much time in there.

What I am struggling with is how to handle loss of the dbus connection.  Right now, the `sdbus` module fails any "subscribe" streaming RPCs when that happens, so `sdexec` gets a failure.  Currently it exits (wrong!).   A better approach might be to just kill off running work but allow new work to be started.  I'm not completely sure what losing that connection means though
- if systemd is restarted, transient units are lost anyway (need to verify but pretty sure)
- if dbus is restarted, transient units are not lost, but does this ever happen in real life?
- if dbus connection is lost due to error handling (like sending a message that's too large) then killing running work is likely wrong.  However, can we prevent that from happening by catching those cases early?  I don't think I've ever seen it happen in my testing but I haven't really stress tested it.  More thought/experimentation needed.

garlick:
Pushed a checkpoint that adds fixes for review comments above and begins a little refactoring.  Plus rebased on current master.

src/modules/sdexec/sdexec.c
```diff
@@ -0,0 +1,790 @@
...
+                  "/org/freedesktop/systemd1/unit/%s",
+                  proc->name) < 0)
+        goto error;
+    if (!(proc->in = sdexec_channel_create_input (ctx->h, "stdin")))
+        goto error;
+    if ((flags & SUBPROCESS_REXEC_STDOUT)) {
+        if (!(proc->out = sdexec_channel_create_output (ctx->h,
+                                                        "stdout",
+                                                         channel_cb,
+                                                         proc)))
```
chu11, line 156: nit: tabbing looks off here

src/modules/sdexec/sdexec.c
```diff
@@ -0,0 +1,790 @@
...
+ * call this from unit and channel callbacks.
+ */
+static void finalize_exec_request (struct sdexec_ctx *ctx,
+                                   struct sdproc *proc,
+                                   struct unit *unit)
+{
+    if (sdexec_unit_state (unit) == STATE_INACTIVE
+        && sdexec_unit_substate (unit) == SUBSTATE_DEAD
+        && proc->out_eof_sent
+        && proc->err_eof_sent) {
```
chu11, line 321: should be `!proc->out || proc->out_eof_sent` (and stderr too) b/c STDOUT/STDERR channels can be optionally set?

src/modules/sdexec/sdexec.c
```diff
@@ -0,0 +1,790 @@
...
+}
+
+static void reset_continuation (flux_future_t *f, void *arg)
+{
+    struct sdproc *proc = arg;
+
+    if (flux_rpc_get (f, NULL) < 0) {
+        flux_log (proc->ctx->h,
+                  LOG_ERR,
+                  "rest-failed %s: %s",
```
chu11, line 365: rest -> reset

src/modules/sdexec/sdexec.c
```diff
@@ -0,0 +1,790 @@
...
+{
+    const flux_msg_t *msg = flux_future_aux_get (f, "request");
+    struct sdproc *proc = flux_msg_aux_get (msg, "sdproc");
+    struct sdexec_ctx *ctx = arg;
+
+    if (sdexec_start_transient_unit_get (f, NULL) < 0)
+        goto error;
+    sdexec_channel_close_fd (proc->in);
+    sdexec_channel_close_fd (proc->out);
+    sdexec_channel_close_fd (proc->err);
```
chu11, line 495: could probably use small comment to explain why closing, natural thought would be that it stay open til the end.

chu11 (commented):
more skimming to grok things, noticed a few small things

t/t2409-sdexec.t
```diff
@@ -0,0 +1,112 @@
...
+	cat >hello.exp <<-EOT &&
+	Hello world!
+	EOT
+	$sdexec -r 0 sh -c "echo Hello world!" >hello.out &&
+	test_cmp hello.exp hello.out
+'
+test_expect_success 'sdexec stderr works' '
+	$sdexec -r 0 sh -c "echo Hello world! >&2" 2>hello.err &&
+	test_cmp hello.exp hello.err
+'
```
chu11, line 75: test both stdout & stderr at same time?

src/modules/sdexec/sdexec.c
garlick: RIght, that's the systemd end of the fd which we know has been received (and duped) once a response to the StartTransientUnit request is received.  I'll add a comment.

chu11:
> libsubprocess: set channel flag only when needed

While you're at it, want to fix the ENOENT to ESRCH in `libsubprocess`?  It can be forked off into a bug fix PR along with this?  (Not necessarily forked off into own PR right now as you're still experimenting).

> sdbus: allow sdbus signal subscriptions to be filtered by unit path
> sdbus: add subscribe-cancel RPC 

Perhaps should be split off into a `sdbus` PR?  Not right now of course, once things become more settled.

garlick:
> sdbus: add subscribe-cancel RPC

Er just realized this was added earler, not here (doh!)

garlick:
Pushed another checkpoint which adds support for `AllowedCPUs` and adds some unit tests.

I did try playing around a bit manually with `flux exec --setopt=SDEXEC_PROP_MemoryMax` and `MemoryHigh` on my test system.  They work!  I did have to add `cgroup_enable=memory` to the kernel command line to enable memory controls.  But after that I could observe the effects when running `stress --vm`:  when the process exceeded `MemoryHigh`, it would be basically niced into submission, and when it exceeded `MemoryMax`, the kernel oom killer would kill it.

chu11:
> Pushed another checkpoint which adds support for AllowedCPUs and adds some unit tests.

(I thought I commented a few days ago but I guess I didn't hit the comment button)

Just wondering if this work should be split off into a different PR?  Doesn't matter at the moment b/c this is WIP, but when the time comes :-)

garlick:
> I think separate PR just to keep this PR size more reviewable.

Ok, agreed!  I have a bit more things to get done then I'll split it up.

garlick:
Pushed a rebase on current master with conflicts resolved, and made debug logging configurable (off by default).

src/modules/sdexec/sdexec.c
```diff
@@ -0,0 +1,850 @@
...
+    if (!(ctx = calloc (1, sizeof (*ctx))))
+        return NULL;
+    ctx->h = h;
+    if (flux_get_rank (h, &ctx->rank) < 0)
+        goto error;
+    /* Add the broker rank to the unit prefix to avoid name collisions
+     * in test instances.
+     */
+    if (parse_config (h, &s) < 0
+        || asprintf (&ctx->unit_prefix, "%s%d-", s, (int)ctx->rank) < 0)
```
chu11: noticed you put "flux-" as default, but would "flux" be more reasonable and you put a "-" here?  i.e. `%s-%d`.  Doubt average user would think to put "-" in the config.

src/modules/sdexec/sdexec.c
```diff
@@ -0,0 +1,850 @@
...
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* sdexec.c - run subprocesses under systemd as transient units
+ */
```
chu11, line 12: super high level thought: split off libsubprocess rexec server into own module?  or into separate "thingy" (file in broker?) to parallel this module? ehhh maybe is overkill, but thought I'd just mention it.

t/t2409-sdexec.t
```diff
@@ -0,0 +1,144 @@
...
+	grep "not found" killunk.err
+'
+test_expect_success 'rank 0 sdexec fails if remote' '
+        test_must_fail flux exec -r 1 $sdexec -r 0 /bin/true 2>restrict.err &&
+        grep "not permitted" restrict.err
+'
+test_expect_success 'rank 0 kill fails if remote' '
+        test_must_fail flux exec -r 1 $rkill -r 0 15 1234 2>killperm.err &&
+	grep "not allowed on rank 0" killperm.err
+'
```
chu11, line 219: i think we're missing a "kill works" test

t/t2409-sdexec.t
```diff
@@ -0,0 +1,144 @@
...
+}
+
+test_expect_success 'sdexec accepts supersize environment' '
+	count=1024 &&
+	make_exports $count >testenv &&
+	sed -e "s/^export //" <testenv | sort >testenv.exp
+	(. $(pwd)/testenv && $sdexec -r 0 printenv) >testenv.raw &&
+	grep TESTVAR_ testenv.raw | sort >testenv.out &&
+	test_cmp testenv.exp testenv.out
+'
```
chu11, line 124: Test seems a little unexpected.  Is there a bug or something related to this test?  maybe a comment needed here.

chu11 (commented):
did a light skim, just a few minor comments.  Didn't go into the properties suff yet.

t/t2409-sdexec.t
garlick: I inferred somewhere that the dbus server might drop the connection if you send it a message that's too big.  So I was thinking if the user's entire environment is included in a dbus message, there might be some size environment that would trigger a disconnect.  I didn't find any such threshold in experiments, but left the test in just to demonstrate that a pretty big environment works OK.  Maybe a comment to this effect would be helpful.

src/modules/job-exec/exec.c
```diff
@@ -309,6 +320,37 @@ static void exit_cb (struct bulk_exec *exec,
...
+    const char *s = "rexec"; // default
+    json_error_t e;
+
+    if (jobspec) {
+        if (json_unpack_ex (jobspec,
+                            &e,
+                            0,
+                            "{s?{s?{s?{s?{s?s}}}}}",
+                            "attributes",
+                              "system",
```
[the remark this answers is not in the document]
garlick: I think technically only `attributes` is required not `system`

https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_14.html#attributes

but OK I'll make that one required.

src/modules/sdexec/sdexec.c
garlick: To clarify: did you mean moving `libsubprocess/server.[ch]` into `broker/exec.c`?

I think the plan is to reuse the subprocess rexec server in flux-shell someday so it probably should remain where it is.

I don't think there is a strong similar argument to be made for sdexec.  Instead of putting the service portion of the sdexec module into an internal convenience library, I put some useful "sdbus client" functions in libsdexec.

src/common/libsdexec/parse.c
```diff
@@ -0,0 +1,122 @@
...
+    if (!s || !bp || !sp)
+        return -1;
+    if (!(ids = idset_decode (s)))
+        return -1;
+
+    // allocate a bitmap large enough to fit idset (1 byte if none set)
+    if ((id = idset_last (ids)) != IDSET_INVALID_ID)
+        nbits = id + 1;
+    if (nbits < 1)
+        nbits = 1;
```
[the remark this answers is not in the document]
garlick: I was assuming an empty bitmap would be legal input.

`nbits` is used to calculate how many bytes to allocate for the bitmap, so this is saying that the smallest size bitmap is 1 byte.  If the bitmap is empty, the byte would be all zeroes.

Hmm, maybe the function could just set the bitmap to NULL and size to zero in this case, and then the code that calls it in start.c would translate that into an empty json array, which might be more correct.  This needs a test too.

src/modules/job-exec/exec.c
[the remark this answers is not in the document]
chu11: Hmm seems like it is optional.  I should double check other places, bc another place I looked considered it required.

src/modules/sdexec/sdexec.c
chu11: > To clarify: did you mean moving libsubprocess/server.[ch] into broker/exec.c?

Yeah basically.  Wasn't sure if it was a good idea or not.  Definitely can put on hold for now.

garlick:
This branch is a bit out of date - my plan is to push an updated version this morning, so it might be good to wait for that before looking at property stuff.

src/modules/job-exec/exec.c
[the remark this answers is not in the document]
chu11: ahhh its RFC25 (jobspec v1) that says "system" is a "MUST".

https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_25.html

But to be on the safe side for the future, "attributes" is probably the only one that should be required for the future.

src/modules/sdexec/sdexec.c
garlick: Good point. Still pondering how best to manage the clean up of lingering process so holding off on changing anything in that area of code for this iteration though.

garlick:
@chu11 - just pushed an update that addresses your comments I think.

src/modules/sdexec/sdexec.c
garlick: In the latest push I eliminated the configured unit prefix.  If an exec request does not have the SDEXEC_NAME option set, then a UUID is used.  Otherwise, it is used verbatim.  Then in `job-exec`, SDEXEC_NAME is set to either `shell-N-jobid` or `imp-kill-N-jobid` (where N is the broker rank).

garlick:
Pushed an update to allow the default exec method to be configured and for the perilog script to use sdexec.  Example:
```toml
[systemd]
enable = true
sdbus-debug = true
sdexec-debug = true

[exec]
imp = "/usr/lib/aarch64-linux-gnu/flux/flux-imp"
service = "sdexec"

[job-manager.prolog]
command = [
   "flux", "perilog-run", "prolog", "--sdexec",
   "-e", "/usr/lib/aarch64-linux-gnu/flux/flux-imp,run,prolog"
]

[job-manager.epilog]
command = [
   "flux", "perilog-run", "epilog", "--sdexec",
   "-e", "/usr/lib/aarch64-linux-gnu/flux/flux-imp,run,epilog"
]
```

Also, pondering about cleanup, I _think_ maybe the sdexec module does not need to go to great lengths to clean up "stray" units (for example those left over after a crash).  Instead, a prolog or epilog script could just use regular systemd tools to do it, if desired - basically stop all service units except for dbus.

Tests still need to be written for the above config.

garlick:
Rebased and added test coverage for prolog/epilog under sdexec, which also covers changing the default exec service to `sdexec`.  I'm going to drop the WIP on this and request a review.

garlick:
Oops! I messed up that last push.  Stand by for a sec while I sort this out.

garlick:
OK better now.  Sorry about that!

garlick:
Restarting inception builder that failed on this known one:
> test_remote.t - exited with status 137 (terminated by signal 9?)

chu11 (commented):
initial review of first few commits, before the giant libsdexec one

src/modules/job-exec/exec_config.c
```diff
@@ -27,6 +27,7 @@
 static const char *default_cwd = "/tmp";
 static const char *default_job_shell = NULL;
 static const char *flux_imp_path = NULL;
+static const char *exec_service = "rexec";
```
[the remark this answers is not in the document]
garlick: It seemed weird next to the `method` key so maybe we can save this until that goes away?  Alternatively, I could just replace `method` with `service`.  Thoughts?

src/modules/job-exec/bulk-exec.c
```diff
@@ -425,8 +476,8 @@ int bulk_exec_push_cmd (struct bulk_exec *exec,
         return -1;
 
     if (zlist_append (exec->commands, c) < 0) {
-        exec_cmd_destroy (c);
-        return -1;
+        errno = ENOMEM;
+        goto error;
```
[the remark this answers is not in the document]
garlick: I can't remember how that got there now, however it does seem required, at least in the`bulk_exec_imp_kill()` -> `bulk_exec_push_one()` -> `bulk_exec_push_cmd()` call path, errno is expected to be set on failure and it wouldn't necessarily be if we left things as is.

Maybe it should be a separate commit.

src/modules/job-exec/exec_config.c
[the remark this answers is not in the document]
chu11: sounds fine to do it later, just wanted to double check.

garlick:
Thanks for the comments! I pushed fixes to address the comments so far.

One thing to note is that guests can currently opt out of sdexec for their jobs.  I wonder if we ought to allow that by default (because it's handy for testing) but prevent if if `exec.service` is set?

src/modules/job-exec/bulk-exec.c
```diff
@@ -477,6 +477,7 @@ int bulk_exec_push_cmd (struct bulk_exec *exec,
 
     if (zlist_append (exec->commands, c) < 0) {
         exec_cmd_destroy (c);
+        errno = ENOMEM;
```
chu11, line 480: not to nit pick, but is ENOMEM impossible from zlist_append()? b/c of their internal asserts? No worries though, this is everywhere in flux-core, we gotta assign some ERRNO for the time being.

garlick: zlist(3) says it can return -1 on out of memory:
```c
  //  Append an item to the end of the list, return 0 if OK or -1 if this
  //  failed for some reason (out of memory). Note that if a duplicator has
  //  been set, this method will also duplicate the item.
  CZMQ_EXPORT int
      zlist_append (zlist_t *self, void *item);
```
I didn't check the source but coding to the docs seemed like the right thing to do.

chu11: ahhh i forgot that my comment tweak was in a PR

https://github.com/zeromq/czmq/pull/2266

no biggie, we can leave for now

garlick: We do have other places in the code like [here](https://github.com/flux-framework/flux-core/blob/master/src/modules/resource/reslog.c#L154) and [here](https://github.com/flux-framework/flux-core/blob/master/src/modules/kvs/lookup.c#L133) that set `errno` to ENOMEM after a `zlist_append()` failure.  So maybe a future cleanup should address this?

chu11: yeah, a future cleanup.

src/common/libsdexec/unit.c
```diff
@@ -0,0 +1,260 @@
...
+
+bool sdexec_unit_has_finished (struct unit *unit)
+{
+    if (unit) {
+        if (unit->exec_main_status_is_set &&
+            unit->exec_main_status < 200)
+            return true;
+    }
+    return false;
+}
```
chu11, line 140: here (and perhaps below too) could use a comment on why number 200 is special

garlick:
Pushed fixes for those last couple comments in libsdexec.

src/common/libsdexec/test/channel.c
```diff
@@ -0,0 +1,231 @@
...
+    output_called = false;
+    error_called = false;
+    sdexec_channel_close_fd (ch);
+    ok (true, "sdexec_channel_close_fd called");
+    ok (flux_reactor_run (flux_get_reactor (h), FLUX_REACTOR_ONCE) >= 0,
+        "flux_reactor_run ran ONCE");
+    ok (output_called == true,
+        "output callback was called");
+    ok (error_called == false,
+        "error callback was not called");
```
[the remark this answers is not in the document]
garlick: Great idea on checking for EOF.

I don't think we can easily trigger `error_cb`.   It is only called when there is a read error on "our" fd (`ch->fd[0]`), which is not exposed outside of the class.  Closing the other fd (`ch->fd[1]`) which _is_ available externally would just cause an EOF which does not cause `error_cb` to be called when it is consumed on the other end.  Closing it outside of `sdexec_channel_close_fd()` would likely do nothing.

garlick:
Pushed libsdexec unit test changes to address the last two comments.

t/t2409-sdexec.t
```diff
@@ -0,0 +1,282 @@
...
+# but 100% is accurate with respect to /proc/meminfo.  Just use 100% for now.
+test_expect_success 'sdexec can set unit MemoryLow property' '
+	echo MemoryLow=$(memtotal) >memlow.exp &&
+	$sdexec -r 0 \
+	    --setopt=SDEXEC_NAME="t2409-memlow.service" \
+	    --setopt=SDEXEC_PROP_MemoryLow=100% \
+	    systemctl --user show --property MemoryLow \
+	        t2409-memlow.service >memlow.out &&
+	test_cmp memlow.exp memlow.out
+'
```
chu11, line 226: two comments.

- dunno if you saw how i did tests in https://github.com/flux-framework/flux-core/pull/4255, I checked that the memory was within 1% of the expected value, so can test more options.

- Also, dunno if intentionally didn't want to test more like if cpusets set correctly?  Most of the extra testing needs cgroups2 support, so dunno if just want to deal with later?  If want to come back to later, perhaps could add comment or create issue to be done later?

chu11 (commented):
reviewed the testing commits at the end, only the big sdexec module left todo

t/t2409-sdexec.t
```diff
@@ -0,0 +1,282 @@
...
+	cat >hello.exp <<-EOT &&
+	Hello world!
+	EOT
+	$sdexec -r 0 sh -c "echo Hello world!" >hello.out &&
+	test_cmp hello.exp hello.out
+'
+test_expect_success 'sdexec stderr works' '
+	$sdexec -r 0 sh -c "echo Hello world! >&2" 2>hello.err &&
+	test_cmp hello.exp hello.err
+'
```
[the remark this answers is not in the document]
garlick: I think this is already covered by this test:
```
test_expect_success 'sdexec can set unit name' '
        $sdexec -r 0 --setopt=SDEXEC_NAME=t2409-funfunfun.service \
            systemctl --user list-units --type=service >name.out &&
        grep t2409-funfunfun name.out
'
```

t/t2409-sdexec.t
garlick: Yes I did review those tests, and yeah my idea was that we'd expand testing when we get down to using cgroups in anger.  Here I mainly just wanted to show that we can set the control properties on our transient units, not check whether that has a particular effect.  I can add a note to that effect.

garlick:
Pushed changes for comments so far.

src/modules/sdexec/sdexec.c
```diff
@@ -0,0 +1,1031 @@
...
+            if (!(f2 = sdexec_reset_failed_unit (h,
+                                                 proc->ctx->rank,
+                                                 sdexec_unit_name (proc->unit)))
+                || flux_future_then (f2, -1, reset_continuation, proc) < 0) {
+                flux_log_error (h, "error initiating unit reset");
+                flux_future_destroy (f2);
+                f2 = NULL;
+            }
+            proc->f_stop = f2;
+        }
```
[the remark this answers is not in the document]
garlick: Since the user can cause failed units I'm not sure this would be useful in real life?

garlick:
Pushed fixes for comments so far.

src/modules/sdexec/sdexec.c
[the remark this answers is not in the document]
chu11: ahhh were you thinking a global config?  I was thinking maybe a per-job config in the jobspec.  Like some user will have some really odd setup and we screwed something up and we need to debug it.

chu11 (approved):
I think that's it ... took awhile to get through!!!

src/modules/sdexec/sdexec.c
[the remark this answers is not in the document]
garlick: Oh right, I thought you meant a global config.  Maybe something to consider later if a use case develops?  Or possibly a "grab all properties before reaping" sort of option (since I'm not sure what else you'd get out of preserving a failed unit)?  Anyway, later?

garlick:
> One thing to note is that guests can currently opt out of sdexec for their jobs. I wonder if we ought to allow that by default (because it's handy for testing) but prevent if if exec.service is set?

Was just thinking maybe we ought to add an
```toml
[exec]
service-override = true
```
sort of config to allow this in test (or on a test system like fluke) but otherwise not allow it?

grondo:
I agree since when sdexec is enabled it is marginally a security issue if a user could bypass it.

garlick:
I went ahead and pushed some commits to implement that.

src/modules/sdexec/sdexec.c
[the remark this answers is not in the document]
chu11: yeah that's fine, we can write up an issue afterwards so we don't forget.  I figure "something will go wrong" is basically inevitable :-)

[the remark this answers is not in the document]
chu11: >  Or possibly a "grab all properties before reaping" sort of option (since I'm not sure what else you'd get out of preserving a failed unit)? Anyway, later?

I'm just trying to think of when you run `systemctl status` if all of that data is elsewhere if you clean it up.  Like sometimes I think you can see stderr error messages via `systemctl status` if it failed?  But that might be in the journald logs too.  But yeah, can figure it out later if its needed.

src/modules/job-exec/exec.c
```diff
@@ -306,11 +308,19 @@ static int parse_service_option (json_t *jobspec,
...
             return -1;
         }
+        if (s2) {
+            if (!override) {
+                errprintf (error, "exec service override is not permitted");
+                errno = EINVAL;
+                return -1;
+            }
+            s = s2;
+        }
```
chu11, line 323: corner case, should check if `s` != `s2`?

src/modules/job-exec/exec_config.c
```diff
@@ -114,12 +120,13 @@ int config_init (flux_t *h, int argc, char **argv)
...
-    /*  Check configuration for exec.service */
+    /*  Check configuration for exec.service and exec.service-override */
     if (flux_conf_unpack (flux_get_conf (h),
                           &err,
-                          "{s?{s?s}}",
+                          "{s?{s?s s?b}}",
                           "exec",
-                            "service", &exec_service) < 0) {
+                            "service", &exec_service,
+                            "service-override", &exec_service_override) < 0) {
```
chu11: should this be documented somewhere?  although skimming comments at top of files in `job-exec` i'm not sure where.

t/t2411-sdexec-job.t
```diff
@@ -23,6 +23,8 @@ cat >config/config.toml <<EOT
 [systemd]
 sdbus-debug = true
 sdexec-debug = true
+[exec]
```
chu11, line 26: commit message typo "line"

t/t2412-sdexec-perilog.t
```diff
@@ -68,4 +68,9 @@ test_expect_success 'prolog was run under systemd' '
 test_expect_success 'epilog was run under systemd' '
 	grep "System epilog script" dmesg.out
 '
+test_expect_success 'configured exec service cannot be overridden' '
+	test_must_fail flux run \
+	    --setattr system.exec.bulkexec.service=rexec \
+	    /bin/true
+'
```
chu11, line 75: related to prior comment, also test override with default?

chu11 (commented):
looks good to me, just some nits

src/modules/job-exec/exec_config.c
garlick: Well maybe I ought to just go ahead and document all the configuration in man pages.

chu11:
LGTM!

garlick:
Added a fixup to one of the tests for the older systemd on TOSS 4 (it requires commands to be fully qualified, whereas the newer one apparently does not - easy fix).

garlick:
After an offline discussion with @grondo, I'm going to set MWP on this.  Thanks very much for the detailed review @chu11!

At the next release we could enable this on fluke and do a "soak test" of sorts while considering next steps for containment/cleanup using cgroups.  Possibly we can reuse some code from #4255.

After this is merged I'll go through the original sdexec prototype and make darn sure I haven't left out anything that should carry over, and then propose removing it and close #5293 if you are OK with that @chu11.

PR #5202, opened 2023-05-24
broker: foward nonfatal signals to all running jobs

grondo:
This PR changes the behavior of the broker when handling nonfatal symbols `SIGUSR1`, `SIGUSR2`.

Currently, the broker ignores `SIGUSR1` and `SIGUSR2`, so a user attempting to signal a batch job with `SIGUSR1` to achieve some effect will be disappointed - the signal is ignored.

This PR changes the handling of these signals so that they are forwarded to all active jobs. If there is an error forwarding signals (e.g. the job-manager module is not loaded), then the broker falls back to calling `state_machine_kill()`.

This allows the `SIGUSR` and `SIGUSR2` signals to be propagated to jobs -- i.e. the expected behavior. If one or more jobs are subinstances, then the signal is forwarded to the jobs in that instance, and so on. The test added here ensures that a `SIGUSR1` delivered to a batch job makes it all the way down to a job running at depth 3 in a hierarchy.

Edit: removed references to `SIGALRM`, the handling of this signal is unchanged here.

### DIFF
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -45,6 +45,7 @@
 #include "src/common/libfluxutil/method.h"
 #include "ccan/array_size/array_size.h"
 #include "ccan/str/str.h"
+#include "ccan/ptrint/ptrint.h"
 
 #include "module.h"
 #include "brokercfg.h"
@@ -1308,7 +1309,8 @@ static void broker_destroy_sigwatcher (void *data)
 
 static int broker_handle_signals (broker_ctx_t *ctx)
 {
-    int i, sigs[] = { SIGHUP, SIGINT, SIGQUIT, SIGTERM, SIGALRM };
+    int i, sigs[] = { SIGHUP, SIGINT, SIGQUIT, SIGTERM,
+                      SIGALRM, SIGUSR1, SIGUSR2 };
     int blocked[] = { SIGPIPE };
     flux_watcher_t *w;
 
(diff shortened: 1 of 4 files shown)


src/broker/runat.c
```diff
@@ -465,9 +465,9 @@ int runat_push_shell_command (struct runat *r,
 
     /*   For shell commands run the target cmdline in a separate process
```
garlick, line 466: commit message for "always execute rc2 in a separate process group": s/doe/done/

src/broker/runat.c
```diff
@@ -465,9 +465,9 @@ int runat_push_shell_command (struct runat *r,
...
      *   group so that any processes spawned by the shell will be signaled
```
chu11, line 467: commit message typo: `doe` -> `done` (I see garlick already saw it too)

src/broker/broker.c
```diff
@@ -1965,13 +1967,75 @@ static void module_status_cb (module_t *p, int prev_status, void *arg)
...
+                    FLUX_NODEID_ANY,
+                    0,
+                    "{s:b s:i s:i}",
+                    "dry_run", 0,
+                    "userid", FLUX_USERID_UNKNOWN,
+                    "signum", signum))
+        || flux_future_then (f, -1., killall_cb, ctx) < 0) {
+            flux_future_destroy (f);
+            return -1;
+    }
```
garlick, line 2013: We usually indent the body of blocks like this by one tab so it begins where the `||` is.

src/broker/broker.c
```diff
@@ -1965,13 +1967,75 @@ static void module_status_cb (module_t *p, int prev_status, void *arg)
...
 
     flux_log (ctx->h, LOG_INFO, "signal %d", signum);
+
+    if (ctx->rank == 0 && !signal_is_deadly (signum)) {
+        /* Attempt to forward non-deadly signals to jobs. If that fails,
+         * the fall through to state_machine_kill() so the signal is
+         * delivered somewhere.
+         */
+        if (killall_jobs (ctx, signum) == 0)
+            return;
```
garlick, line 2033: Might be good to note that this is expected to fail immediately on rank 0 because of the way the broker's internal flux_t handle is implemented.

Aside: is it intended to forward signals to rank 0 if one is received on a rank > 0 broker?

garlick (commented):
Just a couple comments and a question on a first pass

src/broker/broker.c
```diff
@@ -1965,13 +1967,75 @@ static void module_status_cb (module_t *p, int prev_status, void *arg)
...
 
+static bool signal_is_deadly (int signum)
+{
+    int deadly_sigs[] = { SIGHUP, SIGINT, SIGQUIT, SIGTERM, SIGALRM };
+    for (int i = 0; i < ARRAY_SIZE (deadly_sigs); i++) {
+        if (signum == deadly_sigs[i])
+            return true;
+    }
+    return false;
+}
```
chu11, line 1978: stupid nit, feel free to ignore.  The list of non-deadly signals is shorter.  Could do `signal_is_forwardable()` instead?

chu11 (approved):

src/broker/broker.c
grondo: Yeah, I know, good catch. When these lines were moved to their own function the formatting must have been preserved.

src/broker/broker.c
grondo: It should only fail if there is no job-manager module loaded, which should be the extremely uncommon case.
Oh, do you mean instead of getting ENOSYS from the RPC response?

Note: signals are only forwarded from rank 0.

t/issues/t5105-signal-propagation.sh
```diff
@@ -0,0 +1,36 @@
...
+
+id=$(flux submit --output=log flux start \
+     flux run flux start \
+     flux run flux python ./test.py)
+
+$waitfile -t 30 -v ready
+
+flux job kill -s SIGUSR1 $id
+
+$waitfile -t 30 -v -p "got SIGUSR1" log
```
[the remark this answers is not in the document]
grondo: The "extra stuff" is useful when debugging the test. There is only one job run so no need to check the level.

src/broker/broker.c
grondo: Well the list of fatal signals is the only one we keep... This could allow us to add other signals that the broker handles without updating the list of "forwardable" signals. IOW, the list of fatal signals is known and smaller than the set of all signals, and may be reduced in the future.

src/broker/broker.c
grondo: Actually, don't we line up arguments in indented functions, even in a conditional? e.g.

https://github.com/flux-framework/flux-core/blob/5948b01cd86592f6beef89624fa8d6f7654a7a4c/src/cmd/builtin/filemap.c#L222-L228

Note the `||` is indented by one tab. Still the current indentation is wrong

src/broker/broker.c
garlick: Right -  I just meant it might be good to note that the failure, if it occurs, is immediate on rank 0 due to that quirk.  On rank > 0, the request would be forwarded upstream, which led to my wondering whether this code would ever be called on rank > 0.

grondo: This code is currently within a block `if (ctx->rank == 0 && !signal_is_deadly (signum))`, so it will only be called on rank 0. (Still not understanding your question or request here maybe?)

grondo: FYI - changing the comments to the following
```diff
+
+    if (ctx->rank == 0 && !signal_is_deadly (signum)) {
+        /* Attempt to forward non-deadly signals to jobs. If that fails,
+         * then fall through to state_machine_kill() so the signal is
+         * delivered somewhere.
+         */
+        if (killall_jobs (ctx, signum) == 0)
+            return;
+        /*
+         * Note: flux_rpc(3) in the rank 0 broker to the job manager module
+         *  is expected to fail immediately if the job-manager module is not
+         *  loaded due to the broker internal flux_t handle implementation.
+         */
+        flux_log (ctx->h,
+                  LOG_INFO,
+                  "killall failed, delivering signal %d locally instead",
+                  signum);
+    }
```

garlick: That's what I wanted.  Thank you!

garlick (approved):
LGTM!

grondo:
Thanks!

Issue #5037, opened 2023-03-29
python: support general eventlog watching, not just job eventlog watching


chu11:
Was going to look at tackling #4792 until I realized lots of stuff in python bindings lacking for it (I think, unless I'm missing it).

Would be useful to abstract/refactor some general event stuff out what is in `flux/job/event.py` for non-job specific eventlog watching (most notably `class EventLogEvent`).

Also, does not appear a kvs "watch" can be done given KVS api that exists in bindings/python.  Bindings for KVS might be super out of date??

```
327f240d8f (Tom Scogland          2015-07-09 14:44:42 -0700  31) def get_key_direct(flux_handle, key):
294d0b451a (Tom Scogland          2018-12-21 13:43:06 -0800  32)     valp = ffi.new("char *[1]")
af17da3b00 (Albert Chu            2019-01-29 14:41:54 -0800  33)     future = RAW.flux_kvs_lookup(flux_handle, None, 0, key)
f4c9e2d6d9 (Albert Chu            2018-10-10 14:58:50 -0700  34)     RAW.flux_kvs_lookup_get(future, valp)
410e764911 (Stephen Herbein       2016-09-07 10:12:24 -0400  35)     if valp[0] == ffi.NULL:
b4198fe53d (Tom Scogland          2016-09-22 15:46:44 -0700  36)         return None
327f240d8f (Tom Scogland          2015-07-09 14:44:42 -0700  37) 
294d0b451a (Tom Scogland          2018-12-21 13:43:06 -0800  38)     ret = json.loads(ffi.string(valp[0]).decode("utf-8"))
f4c9e2d6d9 (Albert Chu            2018-10-10 14:58:50 -0700  39)     RAW.flux_future_destroy(future)
f4c9e2d6d9 (Albert Chu            2018-10-10 14:58:50 -0700  40)     return ret
327f240d8f (Tom Scogland          2015-07-09 14:44:42 -0700  41) 
```

Is this function still assuming all values in the KVS are stored as json?  That's super old news.  Even a basic read of an eventlog borked due to that `json.loads()` call.

grondo:
The bindings allow direct calls to underlying C APIs since they're using cffi. What is the actual use case for watching an eventlog from Python? Seems like uses of the Python API might want the higher level functions like those provided in `job/event.py`?

chu11:
For #4792 it was to read `resource.eventlog` and `flux resource` is in python :-)

grondo:
[...]

PR #5204, opened 2023-05-24
python: support non-JSON formatted data in KVS

### LINKED ISSUE #5198: python: kvs still assumes all KVS values must be json
Per comments in #5037 and #5196.  json assumption was largely removed like in 2016 or something.  Since most users use the  "job" python API, this issue has lingered in the Python KVS functions for ages.

Example from #5196:

```
import flux
import flux.kvs

f = flux.Flux()
d = flux.kvs.KVSDir(f)

print(d['a']['b']['c'])
```

```
>flux kvs put a.b.c=foo
>flux python ./test.py
  File "./test.py", line 7, in <module>
    print(d['a']['b']['c'])
  File "/g/g0/achu/chaos/git/flux-framework/flux-core/src/bindings/python/flux/kvs.py", line 175, in __getitem__
    return get(self.fhdl, self.key_at(key))
  File "/g/g0/achu/chaos/git/flux-framework/flux-core/src/bindings/python/flux/kvs.py", line 71, in get
    return get_key_direct(flux_handle, key)
  File "/g/g0/achu/chaos/git/flux-framework/flux-core/src/bindings/python/flux/kvs.py", line 38, in get_key_direct
    ret = json.loads(ffi.string(valp[0]).decode("utf-8"))
  File "/usr/lib64/python3.6/json/__init__.py", line 354, in loads
    return _default_decoder.decode(s)
  File "/usr/lib64/python3.6/json/decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib64/python3.6/json/decoder.py", line 357, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

>flux kvs put a.b.c=\"foo\"
>flux python test.py
foo
```

chu11:
Problem: The KVS python interface writes all data to the KVS in JSON and reads all data back assuming it is formatted in JSON.  This appears to be legacy implementation from the days that all data in the KVS was required to be formatted in JSON.

There are still some benefits to doing this, for example Python lists/dicts can be written/read back easily.  However, there are clear
problems with this, most notably data written in non-JSON format cannot be read from the KVS via the Python interface at all.

Solution: With the goal of not breaking current code, support the following compromise.  If data that is read back is not in JSON format, return it as a string if all the data is UTF-8 valid.  If it is not UTF-8 valid, return it in a Python bytes object.  If data cannot be written in JSON, write it as binary data if the data is a bytes object.

Fixes #5198

### DIFF
--- a/src/bindings/python/flux/kvs.py
+++ b/src/bindings/python/flux/kvs.py
@@ -35,7 +35,12 @@ def get_key_direct(flux_handle, key):
     if valp[0] == ffi.NULL:
         return None
 
-    ret = json.loads(ffi.string(valp[0]).decode("utf-8"))
+    try:
+        ret = json.loads(ffi.string(valp[0]).decode("utf-8"))
+    except json.decoder.JSONDecodeError:
+        ret = ffi.string(valp[0]).decode("utf-8")
+    except UnicodeDecodeError:
+        ret = ffi.string(valp[0])
     RAW.flux_future_destroy(future)
     return ret
 
@@ -78,10 +83,17 @@ def get(flux_handle, key):
 
 
 def put(flux_handle, key, value):
-    json_str = json.dumps(value)
     if flux_handle.aux_txn is None:
         flux_handle.aux_txn = RAW.flux_kvs_txn_create()
-    return RAW.flux_kvs_txn_put(flux_handle.aux_txn, 0, key, json_str)
+    try:
+        json_str = json.dumps(value)
+        return RAW.flux_kvs_txn_put(flux_handle.aux_txn, 0, key, json_str)
+    except TypeError:
+        if isinstance(value, bytes):
+            return RAW.flux_kvs_txn_put_raw(
+                flux_handle.aux_txn, 0, key, value, len(value)
+            )
+        raise TypeError
 
 
 def put_mkdir(flux_handle, key):
--- a/t/python/t0005-kvs.py
+++ b/t/python/t0005-kvs.py
@@ -35,10 +35,13 @@ class TestKVS(unittest.TestCase):
         kd = flux.kvs.KVSDir(self.f)
         kd[key] = value
         kd.commit()
-        nv = kd[key]
+
+        kd2 = flux.kvs.KVSDir(self.f)
+        nv = kd2[key]
         self.assertEqual(value, nv)
         self.assertTrue(isinstance(nv, type))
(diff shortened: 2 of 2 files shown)


grondo (approved):
This LGTM! NIce!

grondo:
On a side note, I wonder if we should open an issue to create a more modern Python KVS api. Maybe just wrapping `put`, `lookup` and `watch` interfaces instead of the old-style KVSDir approach (though that seems to be working fine for now with your updates)

chu11:
> On a side note, I wonder if we should open an issue to create a more modern Python KVS api. Maybe just wrapping put, lookup and watch interfaces instead of the old-style KVSDir approach (though that seems to be working fine for now with your updates)

Yeah, I was originally going to open one, but this seems to working ok now, so I held off.  I figure when I (or someone else) gets to #5037, perhaps something will come together out of need.

chu11:
I was missing a line of coverage and I realized my mistake in a test I added.  Made the following change to the new test.

```
diff --git a/t/python/t0005-kvs.py b/t/python/t0005-kvs.py
index 27e3e67..87c5e8a 100755
--- a/t/python/t0005-kvs.py
+++ b/t/python/t0005-kvs.py
@@ -38,7 +38,10 @@ class TestKVS(unittest.TestCase):
 
         kd2 = flux.kvs.KVSDir(self.f)
         nv = kd2[key]
-        self.assertEqual(value, nv)
+        if isinstance(value, bytes) and type is str:
+            self.assertEqual(str(value), nv)
+        else:
+            self.assertEqual(value, nv)
         if type is not None:
             self.assertTrue(isinstance(nv, type))
 
@@ -71,7 +74,7 @@ class TestKVS(unittest.TestCase):
         )
 
     def test_set_not_legal_json(self):
-        self.set_and_check_context("badjson", "{\"foo\" \"bar\"", str)
+        self.set_and_check_context("badjson", b"{", str)
 
     def test_exists_dir(self):
         with flux.kvs.get_dir(self.f) as kd:
```

If I write illegal json as a string, it'll just be read back as a string.  I need to write illegal json as bytes and when it is read back it'll get a json decode error, so I should end up with a string.  I just write out the `{` character only, making it a non JSON object.

PR #5211, opened 2023-05-26
flux-jobs: reduce likelihood of leaving bad terminal color state on empty output

### LINKED ISSUE #5209: `flux jobs` broken output coloring
I noticed that when I specify a nonexistent output format for `flux jobs`, the terminal coloring breaks. It's best explained through an example:

```bash
# terminal is white, as it should be 
bash-4.4$ flux jobs -o '{user.rabbit_workflowafe}' ƒwL5EX5y
USER.RABBIT_WORKFLOWAFE 
bash-4.4$  # now the terminal is red
```

I'm guessing the logic that ends the red coloring is only triggered after a job is printed, but in this case there is no job to print.

I was able to fix it by providing a correct format that actually printed a line:

```bash
bash-4.4$ flux jobs -o '{user.rabbit_workflow}' ƒwL5EX5y # terminal is still red
USER.RABBIT_WORKFLOW
fluxjob-2068228079616
bash-4.4$ # now the terminal is back to white
```

grondo:
This PR likely fixes #5209, though we'll need @jameshcorbett to test since I was unable to reproduce the original issue report.

However, it was definitely wrong to call the `OutputFormat.print_items()` `pre()` function when a line wasn't going to be printed, so this PR is a valid fix no matter what. By moving the call the `pre()` after the line is formatted, I think we also avoid the need for a `finally` block since most exceptions will be thrown there.


jameshcorbett (approved):
Confirmed that it fixes the issue and the change looks reasonable!

grondo:
Thanks! Will set MWP.

PR #5212, opened 2023-05-26
cli: support option to signal a job a configurable time before expiration

### LINKED ISSUE #5104: support option to signal job a configurable time before expiration/timelimit
There has been a request to support the equivalent of the sbatch `--signal=SIGNAL@TIME` option, where signal SIGNAL is sent TIME before a job is canceled due a timelimit or job expiration.

We could probably accomplish this with a simple shell plugin for `flux run` and `flux submit`, however, there is a slight wrinkle for `flux batch` jobs. This feature is typically used to signal an application that demise is imminent in order to allow the application to create a restart dump or checkpoint. If we "signal" a batch job, the signal will be delivered only to the flux-broker processes, which probably isn't what the user wants. Instead, we need a feature that will ensure the signal gets propagated to the child jobs running in the instance, i.e. the equivalent of `flux job killall -f --signal=SIGNAL`.

We should ponder this for a moment, though. Perhaps there is a way for the brokers to propagate non-fatal signals to any shells still running, which will in turn forward those signals to the job tasks (which may be brokers, so the process works recursively). I'm not sure that is a good idea or a really bad idea. Needs more thought.

In the meantime, we may want to consider the behavior of `flux job kill` for batch and alloc jobs... (Hm, that's a larger issue than the initial feature request described in this issue though)

Edit: I also haven't tested this, so I could be incorrect about the current behavior here.

A prereq here:
 -  [x] #5105

grondo:
This PR adds support for a new `--signal=SIG@TIME` option to cli submission commands to optionally send a signal to a job a configurable time before the job time limit.

This implementation uses a job shell option  and plugin to implement the timer which generates the signal. In retrospect, it may have been better to implement this feature in the job-exec module, since that module delivers the `SIGALRM` signal when a job reaches its expiration, and that module currently adjusts the expiration by the delay between the creation of _R_ (the starttime) and the actual time at which job shells started.

Another potential benefit to moving this signal delivery to the exec system is that we then have only one place to update the job expiration when that feature is supported.

It would be fairly simple to make the change to this PR if that is desired. However, posting this here as is because it currently works, and I think up to a few seconds early delivery of the configured signal is probably not a showstopper here. Therefore, it might be better to have something working than perfect.

Fixes #5104


t/t2618-job-shell-signal.t
```diff
@@ -0,0 +1,77 @@
...
+	test_fails_by_signal 10 flux run -t 5s --signal=@4.5s sleep 15
+'
+test_expect_success 'shell: signal option works with alternate signal' '
+	test_fails_by_signal 12 flux run -t 5s --signal=USR2@4.5s sleep 15
+'
+test_expect_success 'shell: signal shell option can be an integer' '
+	test_fails_by_signal 10 flux run -t 5s -o signal -o verbose sleep 10 \
+		2>sig.log &&
+	grep "will expire in 60.0s" sig.log
+'
```
chu11, line 76: i think this test is just testing the defaults, not really testing if integer input acceptable?   Perhaps should have one of each test.

Also, perhaps a test for when `-t` not specified, so signal option shouldn't do anything.

src/shell/signal.c
```diff
@@ -0,0 +1,257 @@
...
+
+    if (expiration == 0.)
+        return 0;
+
+    /* Note: the exec system adjusts expiration by the difference between
+     * the starttime (the time at which R was created), and when the 'start'
+     * event is posted, (after all job shells have started). This shell
+     * plugin does *not* do a similar adjustment, so the signal may be
+     * delivered early. However, this difference should be no more than
+     * a few seconds, which should be critical in real-world scenarios.
```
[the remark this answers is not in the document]
grondo: "should not be no more than a few seconds" (if I'm looking at the same text) would be a double negative (and would be the opposite of what I meant here). However, maybe "this difference should be at most a few seconds" would be more clear?

src/bindings/python/flux/cli/base.py
```diff
@@ -20,6 +20,7 @@
 import os
 import re
 import resource
+import signal
 import sys
 from collections import ChainMap
 from itertools import chain
```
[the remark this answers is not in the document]
grondo: I started with SIG@TIMELEFT  but it seemed overly verbose. The help description says "TIME before job expiration" and I figured that was good enough.

t/t2618-job-shell-signal.t
grondo: > i think this test is just testing the defaults, not really testing if integer input acceptable? Perhaps should have one of each test.

No, if `-o option` is used on the command like the default is to set `attributes.system.shell.options.option = 1`:

```console
$ src/cmd/flux run -o signal --dry-run hostname | jq .attributes.system.shell.options.signal
1
```

> Also, perhaps a test for when -t not specified, so signal option shouldn't do anything.

Good idea! I'll add that.

src/shell/signal.c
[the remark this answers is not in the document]
grondo: Oh, I see I read the first "should be" and not the second :facepalm: Will fix!

grondo:
I addressed @chu11's comments and force-pushed the result.

chu11 (approved):
LGTM, although don't have much of an opinion on if it should be moved into `job-exec`.  Perhaps someone else has a strong opinion on it?

garlick:
> Perhaps someone else has a strong opinion on it?

I think this is probably fine as is for now or maybe forever. I guess if it becomes a problem we know what to do.  By then maybe the planned exec system upheaval will be complete.

I'll add my review shortly, but at first look it seems great.

garlick:
Your recent work (#5202) makes it so that a signal delivered to a batch instance gets propagated to all jobs running in it, so `--signal` is expected to be specified in the batch script or batch submission line correct?  (As opposed to the `flux run` commands or whatever within the batch script?)

I can't remember if we assign a duration to jobs when the enclosing instance has a known time limit?

src/common/libutil/Makefile.am
```diff
@@ -100,7 +100,9 @@ libutil_la_SOURCES = \
 	strstrip.c \
 	strstrip.h \
```
garlick, line 101: Commit message for "add signal name/number helper routines" has the return value as `int` for `signutil_signame()`.  Should be `const char *`

src/common/libutil/sigutil.c
```diff
@@ -0,0 +1,127 @@
...
+
+#include <errno.h>
+#include <signal.h>
+#include <stdbool.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "ccan/str/str.h"
+#include "ccan/array_size/array_size.h"
+
```
garlick, line 23: sigutil.h should be included to ensure prototypes match definitions.

grondo:
> Your recent work (https://github.com/flux-framework/flux-core/pull/5202) makes it so that a signal delivered to a batch instance gets propagated to all jobs running in it, so --signal is expected to be specified in the batch script or batch submission line correct? (As opposed to the flux run commands or whatever within the batch script?)

It can be either I think, though I believe the more common use case is with `flux batch` (?) #5202  was therefore a prerequisite for this work.

The scheduler does propagate any time limit of the current instance to the job in question when creating R. This can be seen with this test:

```console
$ flux alloc -n1 -t 15s flux run flux job timeleft
12
[detached: session exiting]
```

Though that doesn't require that `--signal` be used on the `flux run` line -- it would be optional in all cases.

src/shell/signal.c
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    /* Destroy any current watcher in case this function is called due
+     * to an update in a job's expiration (future update).
+     */
+    flux_watcher_destroy (sig->watcher);
+    sig->watcher = NULL;
+
+    if (expiration == 0.)
+        return 0;
```
garlick, line 105: If the user specified `--signal` on a job with no expiration, should something be logged like
> job has no expiration time, SIG will not be sent

grondo:
Oh, are you saying that the option is possibly unnecessary for `flux run` and `flux submit`? I guess for most cases that could be true, but since the `-t, --time-limit` option is offered on those commands, I reasoned that the `--signal` option should be present as well.

garlick (approved):
LGTM!   Just a couple minor nits.

garlick:
> Oh, are you saying that the option is possibly unnecessary for flux run and flux submit

No no, I was just not remembering that the scheduler sets the expiration when none is specified, and it worried me that people might think they need to specify it on the run/submit line and it would be silently ignored.  I was just confused, never mind!

src/shell/signal.c
grondo: Great idea, I'll add that.

src/common/libutil/sigutil.c
grondo: Oh, good catch.

grondo:
> might think they need to specify it on the run/submit line and it would be silently ignored.

I like your idea of issuing a warning in this case. I've added that and fixed your other comments and force-pushed the result.

Thanks @chu11 and @garlick!

grondo:
Ok, unless I hear something I'll set MWP soon.

PR #5213, opened 2023-05-27
job-list: allow list-id to wait for job state                                                                                               

### LINKED ISSUE #2864: job-info: list-id service wait for specific state

chu11:
Problem: It'd be convenient if the `job-list.list-id` service not only waited for a job id to be legal but to also waited for the job to pass
 a certain job state.  Then it can also be used to "wait on" `job-list` for a number of eventual consistency checks in the testsuite.
                                                                                                                                            
Fixes #2864                                                                                                                                 
                                                                                                                                                                                                                                                                                       
Also stuck on a testsuite coverage for a prior PR that I just forgot to do.

### DIFF
--- a/src/cmd/flux-job.c
+++ b/src/cmd/flux-job.c
@@ -145,6 +145,13 @@ static struct optparse_option list_inactive_opts[] =  {
     OPTPARSE_TABLE_END
 };
 
+static struct optparse_option list_ids_opts[] =  {
+    { .name = "wait-state", .key = 'W', .has_arg = 1, .arginfo = "STATE",
+      .usage = "Return only after jobid has reached specified state",
+    },
+    OPTPARSE_TABLE_END
+};
+
 static struct optparse_option urgency_opts[] =  {
     { .name = "verbose", .key = 'v', .has_arg = 0,
       .usage = "Output old urgency value on success",
@@ -442,7 +449,7 @@ static struct optparse_subcommand subcommands[] = {
       "List job(s) by id",
       cmd_list_ids,
       OPTPARSE_SUBCMD_HIDDEN,
-      NULL,
+      list_ids_opts,
     },
     { "urgency",
       "[OPTIONS] id urgency",
@@ -1380,6 +1387,8 @@ int cmd_list_ids (optparse_t *p, int argc, char **argv)
     int optindex = optparse_option_index (p);
     flux_t *h;
     int i, ids_len;
+    flux_job_state_t state;
+    const char *state_str;
 
     if (isatty (STDOUT_FILENO)) {
         fprintf (stderr,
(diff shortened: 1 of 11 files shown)


t/python/t0005-kvs.py
```diff
@@ -39,7 +40,9 @@ def set_and_check_context(self, key, value, type):
         kd2 = flux.kvs.KVSDir(self.f)
         nv = kd2[key]
         if isinstance(value, bytes) and type is str:
-            self.assertEqual(value.decode('utf-8'), nv)
+            self.assertEqual(value.decode("utf-8"), nv)
```
grondo, line 43: I think normally these files are not subject to formatting by `black`. That might be why the formatting changed Did you run `black` directly on the file?

Actually, it could be a mistake that the python test files aren't being formatted. The `.pre-commit-config.yaml` has:
```yaml
files: '^(src/bindings/python/flux|src/cmd|t/.*\\.py)'
exclude: "^(src/bindings/python/_flux/|src/bindings/python/flux/utils/)"
```
So it does appear that inclusion of python files under `t/` was intended. I think the double backslash is preventing this regex from matching though. I suppose we should fix this in a separate PR. I leave whether you keep the unrelated formatting change in this PR up to you.

t/python/t0005-kvs.py
```diff
@@ -39,7 +40,9 @@ def set_and_check_context(self, key, value, type):
...
+        elif isinstance(value, bytes) and type is dict:
+            self.assertEqual(ast.literal_eval(value.decode("utf-8")), nv)
```
grondo, line 45: Could use a comment -- at least, I don't understand exactly why `ast.literal_eval` is needed here.
Also instead of `self.assertEqual` should you use `self.assertDictEqual`?

src/modules/job-list/idsync.c
```diff
@@ -174,26 +185,31 @@ void idsync_check_id_valid_cleanup (struct idsync_ctx *isctx,
...
 
-    if (!zlistx_add_end (list_isd, isd))
+    if (!zlistx_add_end (iwl->l, isd))
         goto enomem;
 
     return 0;
 
 enomem:
-    zlistx_destroy (&list_isd);
+    idsync_wait_list_destroy ((void **)&iwl);
```
grondo, line 215: cast to `(void **)` is unnecessary here I think.

src/modules/job-list/idsync.c
```diff
@@ -54,6 +54,7 @@ static struct idsync_data *idsync_data_create (flux_t *h,
                                                flux_jobid_t id,
                                                const flux_msg_t *msg,
                                                json_t *attrs,
+                                               flux_job_state_t state,
```
grondo, line 57: Commit message: it might be fine, but I was stumbling over reading of:
> Solution: Support an option job state in the job-list.list-id service.

Is it supposed to be either "Support an optional job state in the..." or maybe "Support a job state option in the..."?

src/cmd/flux-job.c
```diff
@@ -145,6 +145,13 @@ static struct optparse_option list_inactive_opts[] =  {
     OPTPARSE_TABLE_END
 };
 
+static struct optparse_option list_ids_opts[] =  {
```
grondo, line 148: Commit message:

> However, the flux job list-ids does not support anything to use it.

Suggestion: "the flux job list-ids command does not support any way to use it" ?

src/cmd/flux-job.c
```diff
@@ -1471,12 +1480,27 @@ int cmd_list_ids (optparse_t *p, int argc, char **argv)
     if (!(h = flux_open (NULL, 0)))
         log_err_exit ("flux_open");
 
+    /* if no job state specified by user, pick first job state of
+     * depend, which means will return immediately
```
grondo, line 1484: Would it be more correct to say "will return as soon as the job-list module is aware of the job" instead of "immediately"?

grondo (commented):
Nice cleanup! Just a few minor comments.

src/modules/job-list/idsync.c
chu11: i get a error/warning with it :-(

grondo: Oh, ok, my bad then!

t/python/t0005-kvs.py
grondo: FYI - I have a PR ready that fixes `.pre-commit-config.yaml` and all the fallout...

chu11: Hmmm, I guess I ran black on it, leading to that change.

chu11:
rebased picking up the black fixes on `t/python` and fixed up per comments above

grondo (approved):
Great! LGTM!

PR #5215, opened 2023-05-28
[WIP] start of work to add prometheus plugin

### LINKED ISSUE #5214: Discussion for Prometheus plugin (example use case autoscaling)
We want to develop a flux plugin that is able to deliver metrics to a server (likely Prometheus and using https://github.com/jupp0r/prometheus-cpp) that can then be sent to prometheus and the horizontal pod scaler adapter. In layman's terms, when the Flux queue gets too big and needs more resources, it can tell the autoscaling plugin and get them, and shrink back down the same. Likely we'd want the plugin build (outside of or alongside flux?) and then loaded in an rc file, like `modload 0 prometheus`. Also note that prometheus is interesting to use for other cases outside of autoscaling. The original discussion started here: https://github.com/flux-framework/flux-core/pull/5184#issuecomment-1565737130

Questions I have:

 - Documentation for writing a plugin
 - Where does the plugin live - internal to flux-core or can it be external? Which would be better?

From @garlick 

> It depends on what kind of plugin is needed. I would think resource utilization or queue length or something like that would be the sort of metric you'd want for autoscaling? (Could we move this to the autoscaling discussion?)

I think to try we would just want to get the current queue stats - jobs that are running (using what resources) and jobs in the queue (and what resources are needed). I think I'd probably start with the most basic of things - number of jobs in the queue, maybe in different states, and then add details to that about resources needed vs. being used.

If I can get enough inkling of how to start, this would be fun for me to try.

Update: using https://github.com/digitalocean/prometheus-client-c/ doh, I can't use a c++ library!

vsoch:
Problem: we want to report metrics from flux to Prometheus
Solution: write a plugin!

This will close #5214 

These early steps create the plugin skeleton, adding build logic to enable the plugin given particular flags are present. I next want to try installing this, adding the module load to rc1, and seeing what happens. I think likely we want the metrics to be updated from elsewhere (meaning the trigger for an update) but not fully understanding how modules run I am not sure yet.

I'm opening this as a draft because I have some questions that would be helpful to be answered!  I have the basic structure of my plugin and I am currently wondering:

- is the mod_main only called when the module is initially loaded? What other things trigger a plugin / module and do we need to add logic somewhere else to the code. E.g.,: "if the queue size changes, update the metric if the module is active."
- How do I get queue metrics from flux? 
- How do I use the configure.ac to get the library paths to add later to the prometheus/Makefile.am? Is it just an automatic thing (e.g., the variable LIBPROM and LIBPROMHTTP will exist?) 

And more for discussion with others that care about the operator - what metrics (and thus abstractions for them, namely gauge, counter, and histogram) are we interested in for our autoscaler?

Also note associated with this work I've started a "how to write a plugin" document because I couldn't really find anything. https://gist.github.com/vsoch/756f10b52f7889e1b781ccdc599fa8cc. Also note that I didn't add my faux test directory because it's largely useless.

I'm going to test building this into a container with a modified rc1, and maybe my first question will be partially answered if I only see the print once. I basically want this to fire up and be running (and eventually that will mean the metrics server) when the Flux broker starts. I know there are ways to do it with `flux cron` and a python script, but I'd rather try the harder approach first that will be better (and indeed I'm learning a lot and enjoying that).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -98,6 +98,10 @@ compile_flags.txt
 .idea
 .clangd
 
+# Python
+src/bindings/python/*.so-built
+src/bindings/python/_flux
+
 # Rules to ignore auto-generated test harness scripts
 test_*.t
 !/src/bindings/python/test_commands/test_runner.t
--- a/configure.ac
+++ b/configure.ac
@@ -377,7 +377,7 @@ AS_IF([test "x$with_flux_security" = "xyes"], [
                       [flux_sec_incdir=`$PKG_CONFIG --variable=includedir flux-security`],
                       [flux_sec_incdir=;])
     AS_IF([test "x$flux_sec_incdir" = x],
-          [AC_MSG_ERROR([couldn't find flux-security or include directory])])
+          [AC_MSG_ERROR([could not find flux-security or include directory])])
     AC_CHECK_HEADERS([flux/security/version.h])
     AC_DEFINE([HAVE_FLUX_SECURITY], [1], [Define flux-security is available])
     AC_SUBST(FLUX_SECURITY_INCDIR, $flux_sec_incdir)
@@ -396,6 +396,7 @@ if test "$enable_caliper" = "yes"; then
     AC_DEFINE([HAVE_CALIPER], [1], [Define if you have libcaliper])
 fi
 
+# content-s3 module / plugin
 AC_ARG_ENABLE([content-s3],
     AS_HELP_STRING([--enable-content-s3], [Enable S3 storage backend]))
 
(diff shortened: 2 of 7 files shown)


vsoch:
No longer necessary, anyone interested can see github.com/converged-computing/prometheus-flux and for the actual solution I used, https://github.com/converged-computing/flux-metrics-api - TLDR: the Flux broker serves a Kubernetes custom metrics API directly that can advise an autoscaler/v2. I'll probably do a writeup since I learned a lot (and for future me to not forget). Thanks!

Issue #5216, opened 2023-05-29
zhashx_insert: cannot return ENOMEM


chu11:
While going through #5197 I was surprised to see that `zhashx_insert()`'s comments say it can only return EEXIST on error not ENOMEM.  Looking internally into the code, it appears it `assert()`s on memory errors, thus EEXIST is the only error possible.

Some code in `flux-core` assumes an error on insert is ENOMEM, should clean those cases up.  An audit might be nice too just to double check all the places its used.

Edit : I suppose technically since we built czmqcontainers inside flux-core, and we could turn off asserts, ENOMEM could be possible if there's no segfault or something (didn't dig into the code that closely) .... but I don't think that's how we want to code things, b/c perhaps someday we remove the internal czmqcontainers

garlick:
I used to have a czmq issue (zeromq/czmq#1749) open on "fixing" czmq and had intended to do that work and submit it upstream, which is why I wanted the code in flux to assume ENOMEM was possible.  However, that work kept falling off the priority list.

chu11:
From @garlick in #5217

> These are simple containers that do not change much upstream and that we use quite a bit. Maybe it would be OK to change our copy to not assert on allocation error?

I hadn't really thought of that, as my first thought was that we might get rid of `libczmqcontainers` someday.  It was originally added because I found bugs in the upstream implementation and we knew they wouldn't propagate to distros for some time.  So in my mind we would remove the library .... someday.

It wouldn't hurt to try and remove the asserts and return error on memory allocation instead.  Questions I think of off the top of my head:

- For some functions, like `zhash_insert()`, I think we'd have to begin setting `errno` since we wouldn't know what a `-1` return means anymore (i.e. is it EEXIST or ENOMEM).  Not sure if that could lead to cascading issues where we need/want to set errno on all functions with similar situations.  And hopefully there's no `void` returning functions that we would not to alter to return non-void (didn't look closely yet).

- A lot of the zlist/zhash functions assert on invalid input, do we want to remove asserts there as well?  And set errno = EINVAL?

- and of course we have to audit usage everywhere to see fallout.

- and maybe in `flux-sched` too since we copied `libczmqcontainers` there as well

garlick:
[...]

PR #5217, opened 2023-05-30
zhash_insert()/zhashx_insert() return EEXIST 

chu11:
Problem: zhash_insert() and zhashx_insert() return -1 when a duplicate entry exists, NOT when the process is out of memory.  However, a number of locations assume it returns when out of memory and set errno = ENOMEM as a result.

Do not assume zhash_insert()/zhashx_insert() fail on memory allocation errors.  Always ssume duplicate entry.  When necessary set/return errno = EEXIST instead of ENOMEM.

Fixes https://github.com/flux-framework/flux-core/issues/5216

Notes:

My initial reason for doing this was just to make things consistent b/c there was inconsistency in `flux-core`.   Don't know how necessary this is ... Maybe this was just a pointless excercise ... but since I did it, here's the PR :P

There were obviously borderline cases.  Maybe another approach would be to set errno = ENOMEM when a `zhash_insert()` is done directly after a `zhash_lookup()`?

### DIFF
--- a/src/broker/groups.c
+++ b/src/broker/groups.c
@@ -121,11 +121,7 @@ static struct group *group_lookup (struct groups *g,
         }
         if (!(group = group_create (name)))
             return NULL;
-        if (zhashx_insert (g->groups, group->name, group) < 0) {
-            group_destroy (group);
-            errno = ENOMEM;
-            return NULL;
-        }
+        (void)zhashx_insert (g->groups, group->name, group);
     }
     return group;
 }
--- a/src/broker/runat.c
+++ b/src/broker/runat.c
@@ -432,10 +432,7 @@ static int runat_push (struct runat *r,
     if (!(entry = zhashx_lookup (r->entries, name))) {
         if (!(entry = runat_entry_create (name)))
             return -1;
-        if (zhashx_insert (r->entries, name, entry) < 0) {
-            runat_entry_destroy (entry);
-            return -1;
-        }
+        (void)zhashx_insert (r->entries, name, entry);
     }
     if (zlist_push (entry->commands, cmd) < 0) {
         if (zlist_size (entry->commands) == 0)
--- a/src/broker/service.c
+++ b/src/broker/service.c
@@ -160,10 +160,7 @@ int service_add (struct service_switch *sh, const char *name,
     svc = service_create (uuid);
     svc->cb = cb;
     svc->cb_arg = arg;
-    if (zhash_insert (sh->services, name, svc) < 0) {
-        errno = ENOMEM;
-        goto error;
-    }
+    (void)zhash_insert (sh->services, name, svc);
     zhash_freefn (sh->services, name, (zhash_free_fn *)service_destroy);
     return 0;
 error:
--- a/src/cmd/flux-exec.c
+++ b/src/cmd/flux-exec.c
@@ -117,8 +117,7 @@ void completion_cb (flux_subprocess_t *p)
             if (!(idset = idset_create (rank_range, 0)))
                 log_err_exit ("idset_create");
             (void)zhashx_insert (exitsets, buf, idset);
-            if (!zhashx_freefn (exitsets, buf, idset_destroy_wrapper))
-                log_err_exit ("zhashx_freefn");
+            (void)zhashx_freefn (exitsets, buf, idset_destroy_wrapper);
         }
 
         if (idset_set (idset, rank) < 0)
--- a/src/cmd/flux-exec.c
+++ b/src/cmd/flux-exec.c
@@ -116,8 +116,7 @@ void completion_cb (flux_subprocess_t *p)
         if (!(idset = zhashx_lookup (exitsets, buf))) {
             if (!(idset = idset_create (rank_range, 0)))
                 log_err_exit ("idset_create");
-            if (zhashx_insert (exitsets, buf, idset) < 0)
-                log_err_exit ("zhashx_insert");
+            (void)zhashx_insert (exitsets, buf, idset);
             if (!zhashx_freefn (exitsets, buf, idset_destroy_wrapper))
                 log_err_exit ("zhashx_freefn");
         }
--- a/src/common/librlist/rlist.c
+++ b/src/common/librlist/rlist.c
@@ -1605,16 +1605,7 @@ static zhashx_t *rlist_properties (struct rlist *rl)
                         errno = ENOMEM;
                         goto error;
                     }
-                    /* This zhashx_insert() cannot fail since we are
-                     *  guaranteed that `name` is not set in properties
-                     *  hash. However, check for error return in case
-                     *  zhashx_insert() returns -1 on ENOMEM in the future.
-                     */
-                    if (zhashx_insert (properties, name, ids) < 0) {
-                        idset_destroy (ids);
-                        errno = ENOMEM;
-                        goto error;
-                    }
+                    (void)zhashx_insert (properties, name, ids);
                 }
                 if (idset_set (ids, n->rank) < 0)
                     goto error;
--- a/src/common/librouter/disconnect.c
+++ b/src/common/librouter/disconnect.c
@@ -173,10 +173,7 @@ int disconnect_arm (struct disconnect *dcon, const flux_msg_t *msg)
         flux_msg_t *dmsg;
         if (!(dmsg = disconnect_msg (msg)))
             return -1;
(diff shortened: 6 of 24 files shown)


src/broker/groups.c
```diff
@@ -123,7 +123,7 @@ static struct group *group_lookup (struct groups *g,
             return NULL;
         if (zhashx_insert (g->groups, group->name, group) < 0) {
             group_destroy (group);
-            errno = ENOMEM;
+            errno = EEXIST;
```
garlick, line 126: This is inside a block that has already checked whether the group name is in the hash, so I think the insert return code can be ignored.

src/broker/runat.c
```diff
@@ -433,6 +433,7 @@ static int runat_push (struct runat *r,
         if (!(entry = runat_entry_create (name)))
             return -1;
         if (zhashx_insert (r->entries, name, entry) < 0) {
+            errno = EEXIST;
```
garlick, line 436: same here, just ignore the error

src/broker/service.c
```diff
@@ -161,7 +161,7 @@ int service_add (struct service_switch *sh, const char *name,
     svc->cb = cb;
     svc->cb_arg = arg;
     if (zhash_insert (sh->services, name, svc) < 0) {
-        errno = ENOMEM;
+        errno = EEXIST;
```
garlick, line 164: same here, ignore the error

src/common/librlist/rlist.c
```diff
@@ -1612,7 +1612,7 @@ static zhashx_t *rlist_properties (struct rlist *rl)
                      */
                     if (zhashx_insert (properties, name, ids) < 0) {
                         idset_destroy (ids);
-                        errno = ENOMEM;
+                        errno = EEXIST;
```
garlick, line 1615: Just ignore return value and delete the comment above that explains the check.

src/common/librouter/disconnect.c
```diff
@@ -174,6 +174,7 @@ int disconnect_arm (struct disconnect *dcon, const flux_msg_t *msg)
         if (!(dmsg = disconnect_msg (msg)))
             return -1;
         if (zhashx_insert (dcon->hash, key, dmsg) < 0) {
+            errno = EEXIST;
```
garlick, line 177: Inside block that already checked.  Just ignore return value.

src/modules/job-list/idsync.c
```diff
@@ -183,8 +183,10 @@ static int idsync_add_waiter (struct idsync_ctx *isctx,
             goto enomem;
         zlistx_set_destructor (list_isd, idsync_data_destroy_wrapper);
 
-        if (zhashx_insert (isctx->waits, &isd->id, list_isd) < 0)
-            goto enomem;
+        if (zhashx_insert (isctx->waits, &isd->id, list_isd) < 0) {
+            errno = EEXIST;
+            goto cleanup;
```
garlick, line 188: ignore return value, already checked item is not in hash

src/modules/job-list/stats.c
```diff
@@ -72,7 +72,7 @@ static struct job_stats *queue_stats_lookup (struct job_stats_ctx *statsctx,
         if (zhashx_insert (statsctx->queue_stats, job->queue, stats) < 0) {
             flux_log (statsctx->h,
                       LOG_ERR,
-                      "%s: zhashx_insert: out of memory",
+                      "%s: zhashx_insert: queue already exists",
                       __FUNCTION__);
```
garlick, line 76: Can't happen, already checked.  Just ignore return value.

src/modules/job-manager/event.c
```diff
@@ -1085,13 +1085,7 @@ int event_index (struct event *event, const char *name)
...
         entry = int2ptr (((int) zhashx_size (event->evindex) + 1));
         if (zhashx_insert (event->evindex, name, entry) < 0) {
-            /*
-             *  insertion only fails on duplicate entry, which we know
-             *   is not possible in this case. However, cover ENOMEM
-             *   case here in case assert-on-malloc failure is fixed
-             *   in the future for zhashx.
-             */
-            errno = ENOMEM;
+            errno = EEXIST;
```
garlick, line 1088: already checked, just ignore result

src/modules/job-manager/plugins/limit-duration.c
```diff
@@ -94,7 +94,7 @@ static double queues_lookup (zhashx_t *queues, const char *name)
 static int queues_insert (zhashx_t *queues, const char *name, double duration)
 {
     if (zhashx_insert (queues, name, &duration) < 0) { // dups duration
-        errno = ENOMEM;
+        errno = EEXIST;
```
garlick, line 97: fix the caller too - it hardwires "out of memory".

src/modules/job-manager/plugins/perilog.c
```diff
@@ -91,6 +91,7 @@ static struct perilog_proc * perilog_proc_create (flux_plugin_t *p,
     proc->prolog = prolog;
     proc->sp = sp;
     if (zhashx_insert (perilog_config.processes, &proc->id, proc) < 0) {
+        errno = EEXIST;
         free (proc);
         return NULL;
     }
```
garlick, line 97: Set errno after free to avoid it getting clobbered.

src/modules/job-manager/queue.c
```diff
@@ -458,10 +458,13 @@ static int queue_configure (const flux_conf_t *conf,
...
-                if (!(q = jobq_create (name))
-                    || zhashx_insert (queue->named, name, q) < 0) {
-                    jobq_destroy (q);
+                if (!(q = jobq_create (name)))
                     goto nomem;
+                if (zhashx_insert (queue->named, name, q) < 0) {
+                    jobq_destroy (q);
+                    errprintf (error, "error adding new queue");
+                    errno = EEXIST;
+                    return -1;
```
garlick, line 467: already checked with zhashx_lookup.  Just ignore return value of insert.

src/modules/kvs-watch/kvs-watch.c
```diff
@@ -928,6 +928,7 @@ struct ns_monitor *namespace_monitor (struct watch_ctx *ctx,
             return NULL;
         if (zhash_insert (ctx->namespaces, ns, nsm) < 0) {
             namespace_destroy (nsm);
+            errno = EEXIST;
             return NULL;
         }
```
garlick, line 933: already checked, ignore return value.

garlick (changes requested):
In several cases simply changing the code to `(void)zhashx_insert()` is appropriate since the key is known not to be in the hash.

src/modules/job-manager/plugins/limit-duration.c
chu11: ahh good catch on this one and one below.  Having the `queues_insert()` just return void is probably for best.

chu11:
re-pushed per @garlick comments above, mostly updating to ignore return from `zhash_insert()` after `zhash_lookup()`.  In some job-manager plugins, the result was making a `queue_insert()` function return.  I found one other case of `zhash_insert()` after a `zhash_lookup()` that I corrected in `flux-exec`.  

I also noticed a dumb issue in `cmd/flux-exec` at the same time and fixed it with a commit on top.

src/cmd/flux-exec.c
```diff
@@ -118,7 +118,7 @@ void completion_cb (flux_subprocess_t *p)
                 log_err_exit ("idset_create");
             (void)zhashx_insert (exitsets, buf, idset);
             if (!zhashx_freefn (exitsets, buf, idset_destroy_wrapper))
-                log_err_exit ("zhashx_freefn");
+                log_msg_exit ("zhashx_freefn");
         }
```
garlick, line 122: Seems like this return value should be ignored, since zhashx_freefn() only returns NULL when the item doesn't exist, but we just inserted it so we know it does.

chu11: sounds good, tweaked that last commit to ignore the return value instead.

garlick:
Sorry I guess I didn't get back to this for a final approval.  Want to rebase and fix that conflict and then I'll give it a quick final pass?

chu11:
@garlick doh!  i had missed that there was a conflict given a recent merge, thanks for the heads up.  rebased and re-pushed.  I just realized I re-pushed before doing a run of the testsuite locally.  Hopefully I did not re-push too aggressively :P

garlick (approved):
LGTM! Thanks!

PR #5222, opened 2023-05-31
testsuite: increase test expiration

chu11:
Problem: The regression tests in t/issues are now run in parallel, which can increase the chance of test timeouts / failures when a large parallel run of the test suite is done.

Increase the timelimit of each issue to 10m from 2m.  For extra good measure, increase the timeouts of several waits / etc. in several tests under t/issues.


chu11:
recently hit some other test timeouts, through a commit on top for it

Issue #5223, opened 2023-06-01
FAIL: t2801-top-cmd.t 32 - flux-top shows expected data in batch queue


chu11:
It seems #5194 did not fix every race possibility.

```
>cat trash-directory.t2801-top-cmd/batchq.out
nodes 2/2
cores 2/2
gpus 0/0
0 pending
2 running
0 complete
0 failed
f22sUkpXq batch achu R 1 1 2.2s bash
f22sTGqFV batch achu R 1 1 2.2s bash
nodes 2/2
cores 2/2
gpus 0/0
0 pending
2 running
0 complete
0 failed
f22sUkpXq batch achu R 1 1 2.2s bash
f22sTGqFV batch achu R 1 1 2.2s bash
```

It seems the test output is output twice, thus messing up the test.  No idea why it's output twice.

chu11:
Hmmm wondering if a heartbeat could come in a timely/racy manner, output all that stuff, then the "expected" output happens and then exits.

```
static void heartbeat_cb (flux_t *h,                                                                                                        
                          flux_msg_handler_t *mh,                                                                                           
                          const flux_msg_t *msg,                                                                                            
                          void *arg)                                                                                                        
{                                                                                                                                           
    struct top *top = arg;                                                                                                                  
    summary_pane_heartbeat (top->summary_pane);                                                                                             
    summary_pane_draw (top->summary_pane);                                                                                                  
    joblist_pane_draw (top->joblist_pane);                                                                                                  
    summary_pane_query (top->summary_pane);                                                                                                 
}                                                  
```

chu11:
looking through the code, I think this is a corner case in general with the tests, if a job-list (or other) query is slow but a heartbeat arrives a tad earlier than expected, the test output could be different than what is expected.

However, it doesn't explain how this specific output above occurred.  This output suggests that:

- summary outputs were queried and drawn
- job list queried and drawn
- flux_reactor_stop() called
- heartbeat callback called and everything redrawn again

Which I think should be impossible?  I'm not sure if `flux_reactor_stop()` stops the reactor immediately, or if its possible it completes its current iteration of the internal event loop?  If the latter, it could explain the above output, but I don't think that's the case.

grondo:
What's strange is that we (almost?) never see this failure in CI and I haven't seen it locally.

However, maybe some workarounds:

 - I think the heartbeat interval is configurable. Can we just increase it to a very large number for these tests to avoid the potential for one to occur
[...]

PR #5224, opened 2023-06-01
testsuite: fix potential flux-top race

chu11:
Problem: The flux-top command can redraw output when a heartbeat
is received.  On a slow system, such as during a large parallel
run of the testsuite, its possible the heartbeat could arrive before
a test completes, resulting in unexpected output and a failing test.

Solution: In t2801-top-cmd.t, reload the heartbeat module with a much
longer heartbeat period.

Fixes https://github.com/flux-framework/flux-core/issues/5223

### DIFF
--- a/src/cmd/top/top.c
+++ b/src/cmd/top/top.c
@@ -385,6 +385,11 @@ int main (int argc, char *argv[])
                             optparse_get_str (opts, "queue", NULL),
                             &error)))
         fatal (0, "%s", error.text);
+    /* top_create() call above will call initial "draw" routines.  We
+     * do not want those draws to output anything during testing with
+     * the --test-exit-dump option.  Thus we handle --test-exit and
+     * --test-exit-dump setup after the top_create() call.
+     */
     if (optparse_hasopt (opts, "test-exit")) {
         const char *file;
         top->test_exit = 1;
--- a/t/t2801-top-cmd.t
+++ b/t/t2801-top-cmd.t
@@ -20,6 +20,12 @@ job_list_wait_state() {
 	flux job list-ids --wait-state=$2 $1 > /dev/null
 }
 
+# flux-top will redraw panes when a heartbeat is received, which could
+# lead to racy output with tests below.  Remove the heartbeat module to
+# remove this possibility.
+test_expect_success 'set high heartbeat period' '
+	flux module remove heartbeat
+'
 test_expect_success 'flux-top -h prints custom usage' '
 	flux top -h 2>usage &&
 	grep "Usage:.*TARGET" usage


src/cmd/top/top.c
```diff
@@ -385,6 +385,9 @@ int main (int argc, char *argv[])
                             optparse_get_str (opts, "queue", NULL),
                             &error)))
         fatal (0, "%s", error.text);
+    /* Setup test-exit after top_create() call above, we do not want
+     * the very first "draw" to result in any testing output.
+     */
```
grondo, line 390: This is still slightly confusing. What is the testing output -- test-exit just tells `flux-top(1)` to exit after all panes have been drawn once? Actually now that test-exit ensures each pane has been drawn once, _could_ it be set before `top_create()`? (I assume you tried this and thus the comment, so I'm probably still missing some key information - I didn't look at the code in detail)

t/t2801-top-cmd.t
```diff
@@ -32,6 +32,12 @@ job_list_wait_state() {
 	return 0
 }
 
+# flux-top will redraw panes when a heartbeat is received, which could
+# lead to racy output with tests below.  Set the heartbeat period to
+# something really high to avoid it.
+test_expect_success 'set high heartbeat period' '
+	flux module reload heartbeat period=30m
+'
```
grondo, line 40: Question: what happens if you just unload `heartbeat`?

grondo (commented):
Looks good, but just a couple clarification questions inline.

src/cmd/top/top.c
grondo: I agree the comment will be helpful for future programmers, I was just still a bit confused by the comment (since I didn't look at the code) so wondered if it needed a slight tweak? Maybe just expanding the explanation (and include the use case for --test-exit, i.e. why it is used) would be enough.

t/t2801-top-cmd.t
chu11: ahh i guess that would work too.  I initially thought `flux_event_subscribe()` would fail if the event didn't actually exist, but that's stupid of course, it's an event stream not an rpc.  Do think removing preferable to setting a long period?

chu11:
re-pushed, fixing up the comment and removing the heartbeat module instead of increasing its period to a large number

t/t2801-top-cmd.t
grondo: Eh, I was more just curious. If it worked to remove it then that seems more clear (do not ever generate a heartbeat)

grondo (approved):
Great, looks good!

t/t2801-top-cmd.t
chu11: well hopefully it works, since the race is pretty rare, hard to verify :P

PR #5227, opened 2023-06-05
sdbus: add subscription filtering

garlick:
Problem: sdbus signal subscriptions currently are an all or nothing proposition.

This PR adds optional signal message filtering based on interface, member, and path.  The interface and member must match exactly, while path is treated as a glob.

Also log the anticipated socket path when sdbus can't connect to dbus.

(Peeled off of #5197)


garlick:
Oops - just realized I pushed the wrong version of this branch.  Re-pushed.

t/t2407-sdbus.t
```diff
@@ -140,7 +140,7 @@ test_expect_success 'create subscribe script' '
 	import sys
```
chu11, line 140: commit message typo, "not" -> "no"

t/t2407-sdbus.t
```diff
@@ -140,7 +140,7 @@ test_expect_success 'create subscribe script' '
...
 	import flux
 	handle = flux.Flux()
-	fut = handle.rpc("sdbus.subscribe",{},flags=flux.constants.FLUX_RPC_STREAMING)
+	fut = handle.rpc("sdbus.subscribe",{"member":"PropertiesChanged","interface":"org.freedesktop.DBus.Properties","path":"/org/freedesktop/systemd1/unit/flux-t2406*"},flags=flux.constants.FLUX_RPC_STREAMING)
```
chu11, line 143: perhaps can also test defaults (i.e. no filters) and see that you get extra signals?  Perhaps can just setup another script and launch some unit called "foobar" and make sure its in one signals.out and not the other?

chu11 (commented):
> Oops - just realized I pushed the wrong version of this branch. Re-pushed.

Haha, baby crying delayed me from submitting my initial review :-)

Just small stuff

src/modules/sdbus/sdbus.c
```diff
@@ -16,6 +16,7 @@
 #endif
 #include <flux/core.h>
 #include <systemd/sd-bus.h>
+#include <fnmatch.h>
```
grondo, line 19: trivial nit: extra space after "Problem:" in commit message

grondo (approved):
LGTM, one trivial comment - feel free to ignore!

src/modules/sdbus/sdbus.c
garlick: Fixed, thanks.  I'll go ahead and set MWP.

garlick:
Getting errors like this in the ci /el8 ascii builder
```
TypeError: initializer for ctype 'struct idset *' appears indeed to be 'struct idset *', but the types are different (check that you are not e.g. mixing up different ffi instances)
```
I'll just restart that one.

grondo:
Yes, I have seen that from time to time and have no idea what could cause it. It happens rarely, so a restart usually succeeds. Perhaps @trws could shed some light on a possible cause?

trws:
That's what [this part](https://cffi.readthedocs.io/en/latest/cdef.html?highlight=multiple#ffi-ffibuilder-include-combining-multiple-cffi-interfaces) of the CFFI interface is about. Since the python types are generated dynamically in the interpreter, if you have two completely separate ffi modules load the same C type, they will generate *different types* as far as the interpreter is concerned.  The `ffibuilder.include` stuff is meant to solve that, we must be missing that somewhere. It looks like `_rlist_build` already does this right: https://github.com/flux-framework/flux-core/blob/master/src/bindings/python/_flux/_rlist_build.py#L9

I don't see any of the others including core though, for example, and anywhere types from one ffi/lib touch another where neither includes the other, this will happen.

trws:
To be a bit more precise, looking back at the error, I would bet there's a function dealing with an idset in core or somewhere else that isn't including the ffi from idset during its build.  Really we probably want to make sure *all* of our ffi modules include at least core.  Thankfully the fix should be relatively straightforward, though it will take a bit of dependency tweaking in the build system.

PR #5228, opened 2023-06-05
support RFC 14 `files` file archive in jobspec and add `--add-file` option to cli submission commands

grondo:
This PR first abstracts some of the common code between the shell stage-in plugin and the `flux-filemap(1)` command into a new `libfilemap` convenience library (hope the name is OK), then also uses `libfilemap` to add a new `files` plugin to the job shell to support the RFC 14 `files` attribute, which is defined as a RFC 37 JSON file archive in dictionary form.

With this foundation, a `Jobspec.add_file()` method is added to our `Jobspec` class, and this is used to store the batch script instead of the one-off and undocumented `batch` key in `attributes.system`. (Now the `batch` job shell plugin is unnecessary, however we must keep it to handle legacy jobspec that is still enqueued after an upgrade)

Finally, a new `--add-file` option is added to the cli submission commands, which allows arbitrary files to be stored in jobspec and extracted into the job tmpdir.

This is all preparation for embedding instance configuration in jobspec. Actually, it could work now as an example:
```sh
#!/bin/sh
# flux: -n4
# flux: --add-file=conf.toml="""
# flux: [resource]
# flux: noverify = true
# flux: """
# flux: --broker-opts=-c{{tmpdir}}/conf.toml
flux config get
```
```console
$ flux batch --output=test.out conftest.sh 
ƒBKne5ww6w
$ cat test.out
{"resource":{"noverify":true}}
```


src/common/libfilemap/filemap.h
```diff
@@ -0,0 +1,58 @@
...
+                                  json_t *tags,
+                                  const char *pattern);
+
+/*  Extract an RFC 37 File Archive in either array or dictionary form.
+ *  If 'direct' is true, then avoid indirection through the content cache
+ *  when fetching top level data for each file in 'files'.
+ *
+ *  If 'trace_cb' is set, then it will be called for each extracted file,
+ *
+ *  Returns 0 on success, or -1 with error set in errp when non-NULL.
```
garlick, line 48: Comma with nothing following it.

doc/man1/common/submit-other-options.rst
```diff
@@ -34,6 +34,18 @@ OTHER OPTIONS
    character, then VAL is interpreted as a file, which must be valid JSON,
    to use as the attribute value.
 
+**--add-file=[NAME=]ARG**
+   Add a file to the RFC 37 file archive in jobspec before submission. If no
+   ``NAME`` is provided, then ``ARG`` is assumed to be the path to a local
```
garlick, line 39: For the case where file content comes from a file, it might be helpful emphasize (somehow) that a snapshot of the file is taken at job submission time and that after that changing or removing the file has no effect on the job.

garlick (approved):
Nice!  Moving the shared functions out to `libfilemap` is great cleanup.  I apologize that I just couldn't quite make myself do that in the original submission.

Should we move `libutil/fileref.[ch]` to `libfilemap` so those related things are co-located?

grondo:
> Should we move libutil/fileref.[ch] to libfilemap so those related things are co-located?

I didn't go quite that far here. `fileref.[ch]` and `blobref.[ch]` seemed nicely contained in libutil, and the only reason we have to make a separate libfilemap for the extraction functions is because they depend on `flux_t` and thus libflux-core.

It isn't a bad idea to move those two over, but it does seem a bit like busywork, and will require new things like the broker and libcontent to link with libfilemap. Maybe a separate PR? I'm on the fence.

doc/man1/common/submit-other-options.rst
grondo: How is this?
>    Add a file to the RFC 37 file archive in jobspec before submission. Both
>   the file metatdata and content are stored in the archive, so modification
>   or deletion of a file after being processed by this option will have no
 >  effect on the job.

garlick:
NP, strike that thought then.

doc/man1/common/submit-other-options.rst
garlick: Perfect.

grondo:
Force pushed the changes suggested by @garlick and I'll set MWP.

PR #5229, opened 2023-06-05
libsubprocess: fail with ESRCH when pid cannot be found

garlick:
Problem: when `rexec.kill` cannot look up a pid, it fails with ENOENT (No such file or directory).

Fail with ESRCH (No such process) instead.

Also, don't set the SUBPROCESS_REXEC_CHANNEL flag for a remote subprocess if an `on_channel_out` callback is registered, but no channels are configured.

This was peeled off of #5197.


chu11 (approved):
LGTM!

PR #5231, opened 2023-06-06
testsuite: fix tests when run in debug mode

### LINKED ISSUE #5230: testsuite: redirect to /dev/null when using `flux job list-ids --wait-state`
Doh!  Testsuite cleanup in https://github.com/flux-framework/flux-core/pull/5213 works in CI and when you do `make check`, but it'll fail when doing `-v -d` and stuff is outputting to stdout.  Need to clean this up.

chu11:
Problem: flux job list-ids will error out when output is to stdout instead of a file.  Recent testsuite cleanup did not consider this fact when tests are run in debug mode.

Solution: When using flux job list-ids always redirect to a file. Fix offending cases in t2260-job-list.t, t2261-job-list-update.t, and t2800-jobs-cmd.t by redirecting to /dev/null.

Fixes #5230

### DIFF
--- a/t/t2260-job-list.t
+++ b/t/t2260-job-list.t
@@ -140,9 +140,9 @@ test_expect_success 'submit jobs for job list testing' '
 	#  the job-manager queue.  To ensure no raciness in tests, ensure
 	#  jobs above have reached expected states in job-list before continuing.
 	#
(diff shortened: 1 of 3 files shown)


grondo (approved):
Good catch!

PR #5232, opened 2023-06-06
add `--conf` option for registering instance config in `flux-batch` and `flux-alloc`

grondo:
This PR adds a `--conf` option to `flux-alloc` and `flux-batch` which allows users to iteratively build a JSON config file in the jobspec file archive (i.e. `.attributes.system.files["conf.json"].data`). If this option is used once, then `-c{{tmpdir}}/conf.json` is appended to the broker options so that the config becomes the child instance config. The `--conf=ARG` option takes several possible types of arguments:

 * If `ARG` contains a newline, then it is assume to be a multiline configuration and is parsed as JSON or TOML. This allows config to be embedded directly in a batch directive
 * O/w, if `ARG` contains a `=`, the argument is assumed to be `key=value` where `key` can be dotted to set hierarchical values, and value is parsed as JSON -- e.g. `--conf=resource.noverify=true`
 * O/w, if `ARG` ends in `.json` or `.toml` then the argument is assumed to be a path to a file in JSON or TOML.
 * O/w, an error is raised, but eventually the argument may refer to a "named config" file registered in the system or user Flux config hierarchy. (This is planned as future work).
 
The `--conf` option may be used multiple times to iteratively build a config on the command line.

I'm actually tempted to add the `--conf` option to all submission commands, e.g. it could be useful with `flux run OPTIONS.. flux start -o,-c{{tmpdir}}/conf.json`.

I also wonder if might be useful to add an argument that initializes a job's config from the enclosing instance config.

### DIFF
--- a/doc/man1/common/submit-other-options.rst
+++ b/doc/man1/common/submit-other-options.rst
@@ -48,6 +48,32 @@ OTHER OPTIONS
    encode multiple files.  Note: As documented in RFC 14, the file names
    ``script`` and ``conf.json`` are both reserved.
 
+**--conf=FILE|KEY=VAL|STRING**
+   The ``--conf`` option allows configuration for a Flux instance started
+   via ``flux-batch(1)`` or ``flux-alloc(1)`` to be iteratively built on
+   the command line. On first use, a ``conf.json`` entry is added to the
+   internal jobspec file archive, and ``-c{{tmpdir}}/conf.json`` is added
+   to the flux broker command line. Each subsequent use of the ``--conf``
+   option updates this configuration.
+
+   The argument to ``--conf`` may be in one of several forms:
+
+   * A multiline string, e.g. from a batch directive. In this case the string
+     is parsed as JSON or TOML::
+
+      # flux: --conf="""
+      # flux: [resource]
+      # flux: exclude = "0"
+      # flux: """
+
+   * A string containing a ``=`` character, in which case the argument is
+     parsed as ``KEY=VAL``, where ``VAL`` is parsed as JSON, e.g.::
+
+      --conf=resource.exclude=\"0\"
+
+   * A string ending in ``.json`` or ``.toml``, in which case configuration
+     is loaded from a JSON or TOML file.
+
 **--begin-time=+FSD|DATETIME**
    Convenience option for setting a ``begin-time`` dependency for a job.
    The job is guaranteed to start after the specified date and time.
--- a/doc/man1/common/submit-other-options.rst
+++ b/doc/man1/common/submit-other-options.rst
@@ -48,7 +48,7 @@ OTHER OPTIONS
    encode multiple files.  Note: As documented in RFC 14, the file names
    ``script`` and ``conf.json`` are both reserved.
 
-**--conf=FILE|KEY=VAL|STRING**
+**--conf=FILE|KEY=VAL|STRING|NAME**
    The ``--conf`` option allows configuration for a Flux instance started
    via ``flux-batch(1)`` or ``flux-alloc(1)`` to be iteratively built on
    the command line. On first use, a ``conf.json`` entry is added to the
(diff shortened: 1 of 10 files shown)


doc/man1/common/submit-other-options.rst
```diff
@@ -48,6 +48,32 @@ OTHER OPTIONS
...
+
+   The argument to ``--conf`` may be in one of several forms:
+
+   * A multiline string, e.g. from a batch directive. In this case the string
+     is parsed as JSON or TOML::
+
+      # flux: --conf="""
+      # flux: [resource]
+      # flux: noverify = true
+      # flux: """
```
chu11, line 67: i just noticed that noverify is not documented.  Perhaps should pick something else for the user facing docs?

grondo: Oh, yeah, I probably meant to use `norestrict`

grondo:
I'm addressing some of the issues caught by CodeQL and will do a force-push shortly...

doc/man1/common/submit-other-options.rst
grondo: Went with `resource.exclude` which might actually be useful. :shrug:

grondo:
It was actually straightforward to add the "named" config support, so I've tentatively done that here. (Still need to document it, but thought I'd see what reviewers thought first).

The named config support allows a set of `<name>.toml` or `name.<json>` files to be put in the XDG specfied path for flux, e.g. `/etc/xdg/flux/config/<name>.toml` or `~/.config/flux/config/<name>.toml`, and then refer to these files with `--conf=name`.

The intent is that cluster admins could put, for example, a node-exclusive config in `/etc/xdg/flux/config/node-exclusive.toml` and users can run with it via `flux batch --conf=node-exclusive ...`.

grondo:
Hm, github isn't displaying the new commits perhaps there is an issue b/c I swear I've pushed the latest...

grondo:
Just pushed updated documentation for the `--conf=NAME` usage. I could optionally squash the incremental development of the `--conf` option if that is desired.

t/t2716-python-cli-batch-conf.t
```diff
@@ -0,0 +1,152 @@
...
+test_expect_success 'flux-batch --conf=FILE detects invalid syntax' '
+	cat <<-EOF >conf-bad.toml &&
+	[resource]
+	noverify = foo
+	EOF
+	test_must_fail flux batch --conf=conf-bad.toml -n1 --dry-run \
+		--wrap hostname >parse-error.out 2>&1 &&
+	test_debug "cat parse-error.out" &&
+	grep "parse error" parse-error.out
+'
```
chu11, line 51: should test invalid json syntax too?

src/bindings/python/flux/cli/base.py
```diff
@@ -434,8 +434,32 @@ def update_file(self, path, extension=".toml"):
         self.config = dict_merge(self.config, conf)
         return self
 
+    def _find_config(self, name):
+        # Find a named config as either TOML or JSON in XDG search path
+        for path in util.xdg_searchpath(subdir="config"):
+            # Take the first matching filename preferring TOML:
+            for ext in (".toml", ".json"):
+                filename = f"{path}/{name}{ext}"
+                print(f"checking path {filename}", file=sys.stderr)
```
chu11, line 443: just want to make sure if this is left over debugging?  sort of feels like it could be.

src/bindings/python/flux/cli/base.py
```diff
@@ -388,6 +388,8 @@ class BatchConfig:
     Iteratively build a "config" dict from successive updates.
     """
```
chu11, line 389: commit message "useful to users under /etc/xdg/flux/config/*.toml" should be `{*.toml, *.json}`?

chu11 (commented):
just noticed a few small things

src/bindings/python/flux/cli/base.py
grondo: Yes, thanks for catching that!

grondo:
Thanks @chu11! Made the suggested changes and forced a push.

chu11 (approved):
LGTM!

grondo:
Thanks @chu11 and @garlick! Setting MWP.

PR #5237, opened 2023-06-08
broker: unset SLURM_* in initial program environment

garlick:
Problem: SLURM_ prefixed environment variables that leak through slurm -> flux -> MPI can cause hard to diagnose problems.

Clear all SLURM_ prefixed variables in the environment of the rc scripts spawned by the broker (where rc2 == initial program).

### DIFF
--- a/src/broker/runat.c
+++ b/src/broker/runat.c
@@ -79,6 +79,7 @@ static const char *env_blocklist[] = {
     "FLUX_KVS_NAMESPACE",
     "FLUX_PROXY_REMOTE",
     "PMI_*",
+    "SLURM_*",  // flux-framework/flux-core#5206
     NULL,
 };
 
--- a/src/broker/runat.c
+++ b/src/broker/runat.c
@@ -78,9 +78,7 @@ static const char *env_blocklist[] = {
     "FLUX_URI",
     "FLUX_KVS_NAMESPACE",
     "FLUX_PROXY_REMOTE",
-    "PMI_FD",
-    "PMI_RANK",
-    "PMI_SIZE",
+    "PMI_*",
     NULL,
 };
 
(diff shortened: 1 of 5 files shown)


src/broker/runat.c
```diff
@@ -339,9 +337,14 @@ static int runat_command_modenv (struct runat_command *cmd,
...
-        int i;
-        for (i = 0; blocklist[i] != NULL; i++)
-            flux_cmd_unsetenv (cmd->cmd, blocklist[i]);
+        for (int i = 0; blocklist[i] != NULL; i++) {
+            if (strchr (blocklist[i], '?')
+                || strchr (blocklist[i], '*')
+                || strchr (blocklist[i], '['))
+                flux_cmd_unsetenv_glob (cmd->cmd, blocklist[i]);
+            else
+                flux_cmd_unsetenv (cmd->cmd, blocklist[i]);
```
grondo, line 347: This makes me wonder if `flux_cmd_unsetenv()` should just optionally take a glob pattern and push detection of the pattern inside that function?

Also, elsewhere we have these 3 tests abstracted into an `isa_glob()` function. If it isn't quite enough to warrant moving to libutil at this point (and it is such a small function that I don't think so either), then perhaps using `isa_glob()` here (or in `flux_cmd_unsetenv()` would at least allow the common function to easily be discovered later...

grondo (commented):
LGTM! My one serious comment is whether we want to just extend `flux_cmd_unsetenv()` to optionally take a pattern, since doing the `isa_glob()` test and optionally calling `flux_cmd_unsetenv_glob()` would likely be the/a common use case...

src/broker/runat.c
garlick: Oh duh, excellent idea.  I'll do that.

garlick:
Thanks for that review.  I've implemented your suggestion of just adding glob support to `flux_cmd_unsetenv()` which makes much more sense(!) and fixed up the other stuff.  I did not create `libutil/isa_glob.c` but I did abstract that test out to an identical function so it's more clear.

grondo (approved):
LGTM!

PR #5239, opened 2023-06-08
doc: add usage note for --add-file submission cli option

grondo:
Problem: It may be tempting for users to use the --add-file=PATH job submission cli option to broadcast large files to their job, a use which would be inappropriate.

Add a note to this option explaining when --add-file should be used vs the shell stage-in plugin.


garlick (approved):
LGTM, thanks

PR #5242, opened 2023-06-09
shell/pmi: accept pmi1 and pmi2 as aliases for "simple" pmi

garlick:
Problem: when a user specifies `-o pmi=pmi2`, the effect is the same as `-o pmi=off`.

Make `pmi2` and `pmi1` aliases for `-o pmi=simple`.

This is not too outlandish since the flux `libpmi2.so` only speaks the simple wire protocol anyway.

### DIFF
--- a/src/shell/pmi/pmi.c
+++ b/src/shell/pmi/pmi.c
@@ -599,6 +599,7 @@ static int shell_pmi_init (flux_plugin_t *p,
     struct shell_pmi *pmi;
     json_t *config = NULL;
     const char *pmi_opt = NULL;
+    bool enable = false;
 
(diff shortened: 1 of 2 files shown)


grondo (approved):
Seems reasonable to me!

garlick:
Pushed an update to add
- doc: add usage note for --add-file submission cli option (#5239)
- flux-resource: add `-i, --include=TARGETS` option to `list`  and `info` subcommands (#5236)
- completions: update bash completions with recent command changes (#5241)
- shell/pmi: accept pmi1 and pmi2 as aliases for "simple" pmi (#5242)

PR #5246, opened 2023-06-09
flux-resource: add `R` subcommand

grondo:
It seems like a gap that there's no way for a user to access the underlying resource set being used by the scheduler. In fact, in a system instance, users can't access any kind of instance wide R since access to `resource.R` in the KVS is limited to the instance owner.

This PR adds a simple `flux resource R` command that fetches the scheduler resource set and emits Rv1. By default, the R for all resources is emitted, but there are options to get R for any state supported by `flux resource list`.


chu11 (approved):
LGTM!

grondo:
Thanks! I'll set MWP.

PR #5247, opened 2023-06-10
flux-job: add missing include of signal.h

grondo:
Problem: flux-job.c is missing a necessary include of signal.h.

Add the missing include.

### DIFF
--- a/src/cmd/flux-job.c
+++ b/src/cmd/flux-job.c
@@ -28,6 +28,7 @@
 #include <jansson.h>
 #include <argz.h>
 #include <sys/ioctl.h>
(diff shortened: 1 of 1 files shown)


chu11 (approved):
LGTM, curious did `flux-job.c` not compile on some distro?

grondo:
Yeah, in conda-forge (see [here](https://github.com/conda-forge/flux-core-feedstock/pull/21#issuecomment-1584982680)). It uses a centos6 sysroot with modern everything else, so you get some weird errors that don't happen elsewhere.

grondo:
Thanks! I've set MWP.

PR #5248, opened 2023-06-10
libjob: fix leak in sign_unwrap()

grondo:
While diagnosing a local issue with libflux-security, I noticed that `sign_unwrap()` was suppressing the error from `flux_security_configure(3)` and also had a potential leak.

Fix the leak and grab the `flux_security_last_error()` so that errors are easier to diagnosis.

Add one extra test to the unit tests so that when these tests fail the user has some idea of the actual error.


chu11 (approved):
LGTM

grondo:
Thanks! I'll set MWP.

Issue #5245, opened 2023-06-09
restarting flux from dumpfile with large content fails


grondo:
The Flux instance on rzvernal failed to restart from a dumpfile after `flux shutdown --gc` with the following error:
```
Jun 09 12:11:39 rzvernal6 flux[54603]: broker.err[0]: rc1.0: flux-restore: error storing blob for job/004c/5bd6/5900/0800/guest/output: File too large
```
The cause here was likely related to #5148, as we had seen rogue jobs on this system writing 100s of MB of output.

To work around, the registered dumpfile archive was extracted, jobs with large output (and one input!?) file were removed:
```
$  find job -type f -size +11M | sed 's|/guest/output||' | xargs -ti rm -rf {}
```
Then the archive was recreated and the `RESTORE` link updated.

As a point of interest, removal of the offending jobs reduced the size of the compressed content from 1.2G to 326M.

Then flux was able to restart from the new dumpfile.

Possibly something to fix here is to make `flux restore` more resilient in this situation.

grondo:
Whatever the option is, perhaps that should be the default in `rc1`. If exiting with `0` perhaps `--ignore-if-error` would be a more apt option name...

grondo:
Another idea here that just came to mind would be to have a `flux restore --size-limit=` option and allow the restorer to skip content that is greater than an arbitrary limit. This could help admins "clean up" large content after a `flux shutdown --gc`. This would also make the restore faster when there is exceedingly large content, presuming that running `stat` on the file is faster than attempting to store the content and getting an error...

garlick:
That might be both simpler and easier to reason about.  Good thought!

PR #5254, opened 2023-06-12
Ignore errors when starting flux from a restart dump containing giant blobs

garlick:
Problem: flux could not be restarted on a system after `flux shutdown --gc` due to EFBIG errors from `flux-restore(1)`.

This was more fallout from the earlier lack of any limit on the amount of stdio a job can write to the KVS.   Some large stdio keys were apparently still around and were added to the dump file when flux was shut down.  When flux was brought back, up the  keys could not be restored because they exceed the maximum content cache blob size and manual recovery was necessary.

As discussed in #5245, this adds a new option `--ignore-store-error=LIST` to flux-restore(1) that allows certain store errors to be treated as warnings rather than fatal.  rc1 is modified to call `flux restore --ignore-store-error=EFBIG`.

### DIFF
--- a/doc/man1/flux-restore.rst
+++ b/doc/man1/flux-restore.rst
@@ -53,6 +53,8 @@ OPTIONS
    Bypass the broker content cache and interact directly with the backing
    store.  Performance will vary depending on the content of the archive.
 
+**--size-limit**\ =\ *SIZE*
+   Skip restoring keys that exceed SIZE bytes (default: no limit).
 
 RESOURCES
 =========
--- a/etc/rc1
+++ b/etc/rc1
@@ -47,7 +47,7 @@ if test $RANK -eq 0; then
             flux module load ${backingmod} truncate
         fi
         echo "restoring content from ${dumpfile}"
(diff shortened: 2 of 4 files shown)


garlick:
Aw darn, `strerrorname_np()` is apparently too new.  Well, not too hard to provide a replacement that someday can go away.

garlick:
OK, I pushed a redo based on @grondo's idea of `--size-limit`.  The rc1 sets the size limit to 10mb like the stdio limit.

The one worry with setting such a low limit is that we do have some eventlogs that grow without bound like `resource.eventlog` and `admin.eventlog`.  And I suppose `R` could get quite large when it contains JGF.  Thoughts on a reasonable value?

grondo:
I hope eventlogs don't get this large (or by that time we've implemented a method for "rotating" them), but I wonder if even 100M is a reasonable default value for restore.

garlick:
I bumped the rc1 value to 100m and forced a push.

grondo (approved):
LGTM!

PR #5256, opened 2023-06-13
switch from decimal to f58 jobid encoding in most log messages and shell service name

### LINKED ISSUE #5243: use f58 job encoding instead of integer for shell service, job-exec logging
Problem: some code that predates the ubiquitous use of f58 job id encoding in flux uses integers, which aren't as readily identifiable as a job id at first glance, and consume a little extra screen space in some instances.

For example
``` 
$ flux module list
Module                   Idle  S Service
resource                   26  R 
connector-local             0  R 5588-shell-193529793164083200
job-ingest               idle  R 
kvs-watch                  26  R 
sdbus                      71  R 
job-info                   26  R 
barrier                  idle  R 
kvs                        26  R 
sdexec                   idle  R 
```

or 
```
$ dmesg | grpe job-exec
2023-06-09T16:35:43.881955Z job-exec.info[0]: job-exception: id=193529793164083200: resource allocation expired
2023-06-09T16:35:43.882349Z job-exec.debug[0]: exec_kill: 193529793164083200: signal 14
```
Would there be any harm in converting these cases to f58 as we encounter them?

grondo:
This PR adds an `idf58()` convenience function which uses TLS for the buffer so it is safe to use in broker modules. This function is meant to replace similar local functions used in a few other places in the code base, and also to easily convert jobids to an f58 string in logging functions.

This function is then used to replace common code in a couple places, to render the jobid as f58 in the shell service name, and to update most logging of jobids from decimal to f58.

I put this as a WIP becase I'm not sure if others will agree with the approach, and I may have gone too far converting log messages. Also, maybe there's a better way...

Fixes #5243

### DIFF
--- a/src/modules/job-exec/job-exec.c
+++ b/src/modules/job-exec/job-exec.c
@@ -93,6 +93,7 @@
 
 #include "src/common/libczmqcontainers/czmq_containers.h"
 #include "src/common/libjob/job_hash.h"
+#include "src/common/libjob/idf58.h"
 #include "src/common/libeventlog/eventlog.h"
 #include "src/common/libeventlog/eventlogger.h"
 #include "src/common/libutil/fsd.h"
@@ -290,8 +291,8 @@ static int jobid_exception (flux_t *h, flux_jobid_t id,
 
     flux_log (h,
               LOG_INFO,
-              "job-exception: id=%ju: %s",
-              (uintmax_t) id,
+              "job-exception: id=%s: %s",
+              idf58 (id),
               note);
     return flux_respond_pack (h, msg, "{s:I s:s s:{s:i s:s s:s}}",
                                       "id", id,
@@ -366,8 +367,8 @@ static void kill_shell_timer_cb (flux_reactor_t  *r,
     struct jobinfo *job = arg;
     flux_log (job->h,
               LOG_DEBUG,
-              "Sending SIGKILL to job shell for job %ju",
-              (uintmax_t) job->id);
+              "Sending SIGKILL to job shell for job %s",
+              idf58 (job->id));
     (*job->impl->kill) (job, SIGKILL);
 }
 
@@ -378,8 +379,8 @@ static void kill_timer_cb (flux_reactor_t *r, flux_watcher_t *w,
     flux_future_t *f;
     flux_log (job->h,
               LOG_DEBUG,
-              "Sending SIGKILL to job %ju",
-              (uintmax_t) job->id);
+              "Sending SIGKILL to job %s",
+              idf58 (job->id));
     if (!(f = flux_job_kill (job->h, job->id, SIGKILL))) {
         flux_log_error (job->h,
                         "flux_job_kill (%ju, SIGKILL)",
@@ -1221,7 +1222,7 @@ static void exception_cb (flux_t *h, flux_msg_handler_t *mh,
          *   doesn't dump a duplicate exception into the eventlog.
          */
         job->exception_in_progress = 1;
-        flux_log (h, LOG_DEBUG, "exec aborted: id=%ju", (uintmax_t)id);
+        flux_log (h, LOG_DEBUG, "exec aborted: id=%s", idf58 (id));
         jobinfo_fatal_error (job, 0, "aborted due to exception type=%s", type);
     }
 }
--- a/src/modules/job-exec/job-exec.c
+++ b/src/modules/job-exec/job-exec.c
@@ -174,9 +174,9 @@ flux_future_t *jobinfo_shell_rpc_pack (struct jobinfo *job,
     uint32_t rank;
 
     if (asprintf (&shell_topic,
-                  "%ju-shell-%ju.%s",
+                  "%ju-shell-%s.%s",
                   (uintmax_t) job->userid,
-                  (uintmax_t) job->id,
+                  idf58 (job->id),
                   topic) < 0
         || ((rank = resource_set_nth_rank (job->R, 0)) == IDSET_INVALID_ID))
         goto out;
--- a/src/modules/job-exec/job-exec.c
+++ b/src/modules/job-exec/job-exec.c
@@ -186,7 +186,7 @@ flux_future_t *jobinfo_shell_rpc_pack (struct jobinfo *job,
     flux_future_aux_set (f, "jobinfo", job, (flux_free_f) jobinfo_decref);
     jobinfo_incref (job);
 out:
-    ERRNO_SAFE_WRAP (free ,shell_topic);
+    ERRNO_SAFE_WRAP (free, shell_topic);
     return f;
 }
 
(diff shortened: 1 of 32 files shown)


src/modules/job-exec/job-exec.c
```diff
@@ -173,9 +174,9 @@ flux_future_t *jobinfo_shell_rpc_pack (struct jobinfo *job,
     uint32_t rank;
 
     if (asprintf (&shell_topic,
-                  "%ju-shell-%ju.%s",
+                  "%ju-shell-%s.%s",
                   (uintmax_t) job->userid,
-                  (uintmax_t) job->id,
+                  idf58 (job->id),
                   topic) < 0
```
chu11, line 180: intentional or accidental?   Changing topic string?

src/modules/job-manager/kill.c
```diff
@@ -72,9 +72,7 @@ void kill_handle_request (flux_t *h,
     int sig;
     flux_future_t *f;
     char topic [64];
-    const char *errstr = NULL;
-
-    if (flux_request_unpack (msg, NULL, "{s:I s:i}",
+    const char *errstr = NULL; if (flux_request_unpack (msg, NULL, "{s:I s:i}",
```
chu11, line 75: not sure what happened here :-)

chu11 (commented):
I like this change as well.  Not a thorough review, was just skimming and saw two small things.

src/modules/job-manager/kill.c
grondo: oops, probably accidentally joined two lines while editing. Thanks!

src/modules/job-exec/job-exec.c
grondo: Looks like it is in the right commit, but the commit message doesn't mention that the job-exec module also hand encodes the shell service topic string, so needed an update too. Will fix.

garlick: Oops sorry,  I think I am completely off base.  I was thinking of the service published in the `guest.exec.eventlog` e.g.
```json
{"timestamp":1686328514.0094779,"name":"shell.init","context":{"service":"5588-shell-193529793164083200","leader-rank":1,"size":1}}
```
The misdirected point I was trying to make was that tools that might have different environments etc. should use what they find in the eventlog so should be immune from this problem.

grondo: Yeah, all other entities that use the shell service besides the job-exec module currently use the eventlog. We could update the job-exec module, but I think we figured having the job-exec module watch every job's eventlog to wait for the start event, then watch the exec eventlog for the shell service was an unnecessary burden and would impact job throughput.

grondo: @chu11 does make a good point. However, the job-exec module does directly invoke the job shell through libsubprocess, and uses its current environment as the environment for the job shell. Therefore, I think the likelihood of the job shell running with a different locale or `FLUX_F58_FORCE_ASCII=t` is low.

However, we do allow the job shell to be overridden in jobspec, in that case a user could invoke a job shell that was built in broken locale mode where the ascii `f` is forced. That would be an annoying problem...

Maybe there is something we could do to make it unnecessary for the job-exec module to use the job shell service directly.

grondo: > yeah, since it's a change of topic string, perhaps should be in one commit together?

Is it not?
```diff
commit 99e433f08bd6bc89d7d380bc8260d37c09c6c9eb
Author: Mark A. Grondona <mark.grondona@gmail.com>
Date:   Mon Jun 12 13:21:44 2023 -0700

    shell: use f58 jobid encoding for shell service
    
    Problem: The shell service name uses the decimal encoding of the
    jobid as the suffix, but using the f58 encoding would make it more
    clear that the service corresponds directly to a job.
    
    Encode the jobid in the shell service name as f58.
    
    Since the job-exec module builds the shell service topic string by
    hand, update it as well.
---
 src/modules/job-exec/job-exec.c | 4 ++--
 src/shell/svc.c                 | 6 ++++--
 2 files changed, 6 insertions(+), 4 deletions(-)

diff --git a/src/modules/job-exec/job-exec.c b/src/modules/job-exec/job-exec.c
index c22eb89c3..c363efc73 100644
--- a/src/modules/job-exec/job-exec.c
+++ b/src/modules/job-exec/job-exec.c
@@ -174,9 +174,9 @@ flux_future_t *jobinfo_shell_rpc_pack (struct jobinfo *job,
     uint32_t rank;
 
     if (asprintf (&shell_topic,
-                  "%ju-shell-%ju.%s",
+                  "%ju-shell-%s.%s",
                   (uintmax_t) job->userid,
-                  (uintmax_t) job->id,
+                  idf58 (job->id),
                   topic) < 0
         || ((rank = resource_set_nth_rank (job->R, 0)) == IDSET_INVALID_ID))
         goto out;
diff --git a/src/shell/svc.c b/src/shell/svc.c
index 3f6e4560d..cb0c55d4a 100644
--- a/src/shell/svc.c
+++ b/src/shell/svc.c
@@ -35,6 +35,8 @@
 #include <unistd.h>
 #include <flux/core.h>
 
+#include "src/common/libjob/idf58.h"
+
 #include "internal.h"
 #include "task.h"
 #include "svc.h"
@@ -66,9 +68,9 @@ static int build_topic (struct shell_svc *svc,
 {
     if (snprintf (buf,
                   len,
-                  "%ju-shell-%ju%s%s",
+                  "%ju-shell-%s%s%s",
                   (uintmax_t)svc->uid,
-                  (uintmax_t)svc->shell->info->jobid,
+                  idf58 (svc->shell->info->jobid),
                   method ? "." : "",
                   method ? method : "") >= len) {
         errno = EINVAL;
```

Maybe I'm looking at the wrong thing...

garlick: > Like admin/user just sets something up differently on one node vs another. Using the decimal jobid would just seem safer.

I _think_ since the job-exec module starts the shell with a subprocess rexec, the job-exec and shell environments should match (except for FLUX_URI or other overrides)?

chu11: > Is it not?

Hmm I could have swore I saw one of the changes in another commit.. but maybe I mis-saw.  My bad.

grondo: In any event, it is _probably_ safe to continue with the solution here for now, since it just changes the topic string using the current behavior. Then perhaps we make the change to how the job-exec module gets the shell's topic string in a separate PR, and make sure we do it before the next release. I can open an issue if this is desirable?

garlick: That sounds OK to me.  (Also sorry for not refreshing the discussion before commenting!)

I wonder if shell rank 0 could just listen for fatal exceptions since job-exec publishes one right before sending the "exception" RPC?

grondo: I'm not sure why we didn't do it that way at first. I'd have to review how this works, but not all lost shell exceptions should be fatal I thought.

chu11: sounds good to me

grondo:
Ok, I've pushed an update (found some missing needed conversions in libschedutil, and pushed a change to `flux job info` to print the jobid argument in an error message instead of the decimal jobid) Will open an issue about investigating how to make it unnecessary for job-exec to recreate the shell service string.

grondo:
Note also that I didn't change the job shell event name (seemed unnecessary) nor the sdexec unit name

grondo:
Thanks! I've set MWP.

Issue #5176, opened 2023-05-17
python: add convenience function for `job-info.lookup`


chu11:
While most users will go through `job-list.list` or `job-list.list-id` for their job data needs, some more advanced users may need to get R and jobspec, such as `flux-accounting`.  So it'd be nice to have a convenience function for this instead of calling the RPCs directly.

PR #5259, opened 2023-06-13
python: allow JobID to take a JobID

chu11:
While working on #5176 I noticed a corner case in the `JobList` class.  It assumes that all ids passed in are of type `JobID`, so that later on `jobid.orig` can be output in an error message.  But a jobid could just be an integer.

There are multiple ways to handle this.  I elected to go for a simple:

```
-                    msg = f"JobID {child.jobid.orig} unknown"
+                    id = JobID(child.jobid)
+                    msg = f"JobID {id.orig} unknown"
```

solution.  This also requires the `JobID` class to take a `JobID` as input.  There of course other ways to skin this cat of course.

### DIFF
--- a/src/bindings/python/flux/job/JobID.py
+++ b/src/bindings/python/flux/job/JobID.py
@@ -53,6 +53,8 @@ class JobID(int):
     """
 
     def __new__(cls, value):
+        if isinstance(value, JobID):
+            return cls(value.orig_str)
         if isinstance(value, int):
             jobid = value
         else:
--- a/src/bindings/python/flux/job/list.py
+++ b/src/bindings/python/flux/job/list.py
@@ -13,6 +13,7 @@ import pwd
 
 import flux.constants
 from flux.future import WaitAllFuture
+from flux.job import JobID
 from flux.job.info import JobInfo
 from flux.rpc import RPC
 
@@ -211,7 +212,7 @@ class JobList:
         self.since = since
         self.name = name
         self.queue = queue
-        self.ids = ids
+        self.ids = list(map(JobID, ids)) if ids else None
         self.errors = []
         for fname in filters:
             for x in fname.split(","):
--- a/t/python/t0010-job.py
+++ b/t/python/t0010-job.py
@@ -468,6 +468,12 @@ class TestJob(unittest.TestCase):
                     # Ensure encode back to same type works
                     self.assertEqual(getattr(jobid, key), test[key])
 
+        # JobID can also take a JobID
+        jobid1 = job.JobID(1234)
+        jobid2 = job.JobID(jobid1)
+        self.assertEqual(jobid1, jobid2)
+        self.assertEqual(jobid2.orig, "1234")
+
     def test_25_job_list_attrs(self):
         expected_attrs = [
             "userid",


grondo:
>  This also requires the JobID class to take a JobID as input. There of course other ways to skin this cat of course.

I'm surprised this didn't "just work" since `JobID` is a subclass of `int`.. solution in general seems good to me but I have to study the constructor changes a little more...

src/bindings/python/flux/job/JobID.py
```diff
@@ -53,15 +53,19 @@
...
-        self = super(cls, cls).__new__(cls, jobid)
-        self.orig_str = str(value)
+            if isinstance(value, int):
+                jobid = value
+            else:
+                try:
+                    jobid = id_parse(value)
+                except OSError:
+                    raise ValueError(f"{value} is not a valid Flux jobid")
+            self = super(cls, cls).__new__(cls, jobid)
```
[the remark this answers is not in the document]
chu11: huh ... any ideas one this might be showing up?  It's the same code that was there before.  Not sure if we dismissed it before too.

grondo:
> The issue was the .orig in the error message, which I don't think int class has?? Everything appeared to work fine.

I meant constructing a JobID from a JobID (maybe Python has a specific way to implement a copy constructor)? The changes to the JobID seemed like they should be unnecessary

chu11:
> I meant constructing a JobID from a JobID (maybe Python has a specific way to implement a copy constructor)? The changes to the JobID seemed like they should be unnecessary

Ahhh.  Looking around, perhaps `copy.copy()` might be sufficient.  Although it looks like I still need to check if the input is of type `JobID`.

chu11:
> Ahhh. Looking around, perhaps copy.copy() might be sufficient. Although it looks like I still need to check if the input is of type JobID.

Actually, I think what I did might be correct given what I'm doing ..  i'm certainly not a Python expert though.

`myid = JobID(id)`

is saying "give me a "myid" of class JobID, initialize it given the input "id"` and the changes to `JobID` are to initialize, regardless of what type `id` is (even if an object of type `JobID`).  In contrast to a copy constructor which is typically called for something like:

`myid = obj_of_type_JobID`
or
`myid = copy.copy(obj_of_type_JobID)`

grondo:
Sorry, I should have said I think your approach is correct, just was surprised at the amount of change in the constructor. Then I went for a ride and left you hanging!

Not at a computer at the moment, but will take a look shortly. 

BTW, I agree what you have done is a valid copy constructor. Was just idly wondering if Python offered any neat sugar to help do it.

src/bindings/python/flux/job/JobID.py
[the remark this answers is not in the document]
grondo: Hm, I bet this is just checking changed code, and the underlying checks have been updated.
I think in Python 3 the arguments to `super()` are optional, so we should just change this to 
```suggestion
            self = super().__new__(cls, jobid)
```

src/bindings/python/flux/job/JobID.py
```diff
@@ -53,15 +53,19 @@ class JobID(int):
     """
 
     def __new__(cls, value):
-        if isinstance(value, int):
-            jobid = value
+        if isinstance(value, JobID):
+            self = super(cls, cls).__new__(cls, int(value))
+            self.orig_str = value.orig
```
grondo, line 58: This might be simpler for the copy constructor implementation:
```suggestion
        if isinstance(value, JobID):
            return cls(value.orig_str)
```

Then the rest of the function stays the same.

src/bindings/python/flux/job/list.py
```diff
@@ -140,7 +141,8 @@ def get_jobs(self):
                 jobs.append(child.get_job())
             except EnvironmentError as err:
                 if err.errno == errno.ENOENT:
-                    msg = f"JobID {child.jobid.orig} unknown"
+                    id = JobID(child.jobid)
+                    msg = f"JobID {id.orig} unknown"
```
grondo, line 145: I wonder if elsewhere in the code might assume ids are instances of `JobID`. Maybe the `JobList` constructor should instead attempt to convert all `ids` in the input array to `JobID`, e.g.

```diff
diff --git a/src/bindings/python/flux/job/list.py b/src/bindings/python/flux/job/list.py
index b0ff9710f..c78bd8f2f 100644
--- a/src/bindings/python/flux/job/list.py
+++ b/src/bindings/python/flux/job/list.py
@@ -211,7 +211,7 @@ class JobList:
         self.since = since
         self.name = name
         self.queue = queue
-        self.ids = ids
+        self.ids = list(map(flux.job.JobID, ids))
         self.errors = []
         for fname in filters:
             for x in fname.split(","):
```

grondo (commented):
A couple comments for discussion inline.

src/bindings/python/flux/job/JobID.py
chu11: oh yeah, that is much simpler :-)

src/bindings/python/flux/job/list.py
chu11: I considered this approach as well, but I figured since the only failing part was the `id.orig` I wanted to avoid the unnecessary iteration.  But you are right, it would prevent other issues that we aren't expecting and future issues as well.

chu11:
rebased & re-pushed per comments above

PR #5260, opened 2023-06-14
optparse: add `optparse_get_size()` accessor and use it in `flux restore` and `flux filemap`

grondo:
I couldn't help myself and added an `optparse_get_size()` accessor that takes a string argument of the form `N[bkMG]` to specify a size value. `N` can be a floating-point value with optional suffix `b` (the default) for bytes, `k` for KiB (1024 bytes), `M` for MiB (1024*1024 bytes), and `G` for GiB.

Then a couple commands that take a size argument are updated with this accessor so they can take human readable arguments.

### DIFF
--- a/doc/man1/flux-filemap.rst
+++ b/doc/man1/flux-filemap.rst
@@ -76,7 +76,10 @@ OPTIONS
 
 **--chunksize=N**
    Limit the content mapped blob size to N bytes.  Set to 0 for unlimited.
-   The default is 1048576 (*map* subcommand only).
+   N may be a floating-point number with optional suffix 'b' for bytes (the
+   default), 'k' for kibibytes (KiB, units of 1024 bytes), 'M' for mebibytes
+   (MiB, units of 1024 * 1024 bytes), or 'G' for gibibytes (units of 1024 *
+   1024 * 1024 bytes). The default is 1M (*map* subcommand only).
 
 **--direct**
    Avoid indirection through the content cache when fetching the top level
--- a/doc/man1/flux-restore.rst
+++ b/doc/man1/flux-restore.rst
@@ -54,7 +54,11 @@ OPTIONS
    store.  Performance will vary depending on the content of the archive.
 
 **--size-limit**\ =\ *SIZE*
-   Skip restoring keys that exceed SIZE bytes (default: no limit).
+   Skip restoring keys that exceed SIZE bytes (default: no limit). SIZE is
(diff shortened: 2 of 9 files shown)


garlick:
I just wrote a very similar parser here with unit tests:

https://github.com/garlick/flux-core/blob/sdexec/src/common/libsdexec/parse.c#L40

maybe we ought to combine efforts to make sure parsing behaves the same in all contexts?

grondo:
Oh yeah, sorry I missed that! And I was looking for places where a size in bytes is parsed..

Maybe you can add a generic version of your function to libutil and we can redo this PR based on that after that is merged?

Note that I was basing the argument suffix after what is taken by `find -size` but I like the idea of taking `k` or `K` for KiB so I prefer your existing solution.

grondo:
Actually, looking at some other utilities (e.g. systemd, dd) only uppercase `M`, `G`, and `T` are supported in file size arguments. I assume because especially in the case of `m` this can be confused with the SI milli- prefix (and in our case it could be confused with FSD minutes suffix. I'd propose that `k` or `K` is allowed for KiB, but that other suffixes be required uppercase for clarity. However, I don't feel too strongly on that if there is a good reason you wanted to support lower case.

I'll close this PR since it will need to be completely redone anyway (and isn't critical). We can redo this one once an implementation for parsing a size string has been merged.

garlick:
Sure that sounds reasonable.  Maybe we can document that the [SI metric prefixes](https://www.nist.gov/pml/owm/metric-si-prefixes) are supported and just quietly allow `k` or `K` as you suggest.

grondo:
Good idea, though we should be careful to document that the prefixes in this case are multiples of 1024 not 1000 like the actual SI metric prefixes (i.e. `K` == `KiB`, `M` = `MiB`, etc.). Just as an example `dd` supports the following suffixes:

>  N  and  BYTES may be followed by the following multiplicative suffixes:
>  c=1, w=2, b=512,  kB=1000,  K=1024,  MB=1000 * 1000,  M=1024 * 1024,  xM=M,
>  GB=1000 * 1000 * 1000,  G=1024 * 1024 * 1024, and so on for T, P, E, Z, Y. 
>  Binary prefixes can be used, too: KiB=K, MiB=M, and so on.

We probably don't need to support the two letter suffixes, but it at least does show the interface would be consistent with a couple of extant utilities (`find` and `dd`).

PR #5261, opened 2023-06-14
shell: set correct HOSTNAME in job environment if necessary

### LINKED ISSUE #5258: exclude HOSTNAME from job environment
Problem: The `HOSTNAME` environment variable is currently propagated to jobs, which can result in some very confusing output when used by job tasks or a batch script etc.

The HOSTNAME variable should be excluded by default from the job environment.

grondo:
This PR fixes #5258 by overwriting HOSTNAME in the job environment when it is set.


garlick (approved):
LGTM!

grondo:
Thanks! I'll set MWP.

PR #5262, opened 2023-06-14
libutil: add parse_size()

garlick:
Problem: as discussed in #5260,  a parser for size_t that accepts the common SI metric prefixes could be useful for parsing config and option values.

This PR adds one, combining @grondo's from #5260 and mine from #5197.

### DIFF
--- a/src/common/libutil/Makefile.am
+++ b/src/common/libutil/Makefile.am
@@ -140,7 +140,8 @@ TESTS = test_sha1.t \
 	test_timestamp.t \
 	test_environment.t \
 	test_basemoji.t \
-	test_sigutil.t
+	test_sigutil.t \
+	test_parse_size.t
 
 test_ldadd = \
 	$(top_builddir)/src/common/libutil/libutil.la \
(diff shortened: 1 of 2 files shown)


grondo (approved):
This LGTM. I was going to question the use of `size_t` here, since it can possibly be 32 bits on 32bit platforms and thus `SIZE_MAX` would be only 4G. I had used `off_t` because it is the type of the `st_size` member of `struct stat`, however I wonder if `uint64_t` (as I think you used in your original implementation?) would be safest here?

Since the 32bit CI seems to be passing, perhaps my fears about `size_t` are unfounded, so I'm find with this as is.

grondo:
> BTW the unit tests skip any tests that expect success and would overflow the native size_t so that's probably why CI is passing.

Ah, sorry I missed that! :facepalm:

garlick:
I didn't know about that one.  Would you prefer if we used that and returned a `uintmax_t`?  I *think* the results are identical to `strtoull()` with a `uint64_t` if I'm reading that right.

grondo:
No, I was just pointing out as an interesting find, sorry.

grondo:
I already approved. Go ahead and clean up as desired!

PR #5263, opened 2023-06-14
flux-top: fix formatting with ASCII jobids

grondo:
Problem: When jobids are being encoded with ascii 'f', jobs in the flux-top(1) joblist pane are not aligned with the header row. This occurs because mvwprintw(3) doesn't seem to handle multibyte UTF-8 characters properly. (This appears to be why the JOBID column is formatted with %12s while the jobs are formatted with %13.13s currently).

e.g.
2023-06-14-080621_566x148_scrot

Detect whether ascii 'f' is being used in jobids by testing the result of strlen(3) for jobid 0. If the result is 2 then ascii jobids are being used, so use 13 for the JOBID column width, otherwise the multibyte 'ƒ' is most likely being used so use 12 for the JOBID width.

### DIFF
--- a/src/cmd/top/joblist_pane.c
+++ b/src/cmd/top/joblist_pane.c
@@ -27,6 +27,7 @@ static const struct dimension win_dim = { 0, 6, 80, 60 };
 
 struct joblist_pane {
     struct top *top;
+    int jobid_width;
     WINDOW *win;
     json_t *jobs_all;
     json_t *jobs;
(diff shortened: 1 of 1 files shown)


src/cmd/top/joblist_pane.c
```diff
@@ -413,6 +416,24 @@ void joblist_pane_set_current (struct joblist_pane *joblist, bool next)
     }
 }
 
+/*
+ *  Workaround for mvwprintw(3) issues with multibyte jobid 'ƒ' character.
+ *
+ *  Empirically, the JOBID column must be formatted as %12s when the 'ƒ'
+ *  character appears in Flux jobids, but %13s when ascii 'f' is used.
+ *  Guess at the current jobid encoding by determining if jobid 0 has a
+ *  length of 2 (ascii) or 3 (utf-8).
```
chu11, line 425: commit message and this comment confused me until I read the code.  Perhaps "if f58 encoding of jobid 0 has a ..."  otherwise LGTM

PR #5264, opened 2023-06-14
sdbus: make debug logging configurable

garlick:
Problem: sdbus is prolific in its output to the flux circular log buffer.
    
Disable debug logging by default, and allow it to be enabled by config:
```toml
[systemd]
sdbus-debug = true
````

(This was peeled off of #5197)

### DIFF
--- a/src/modules/sdbus/sdbus.c
+++ b/src/modules/sdbus/sdbus.c
@@ -56,6 +56,20 @@ static void sdbus_recover (struct sdbus_ctx *ctx, const char *reason);
 static const double retry_min = 2;
 static const double retry_max = 60;
 
+static bool sdbus_debug = false;
+
+static __attribute__ ((format (printf, 2, 3)))
+void sdbus_log_debug (flux_t *h, const char *fmt, ...)
+{
+    if (sdbus_debug) {
+        va_list ap;
+
+        va_start (ap, fmt);
+        flux_vlog (h, LOG_DEBUG, fmt, ap);
+        va_end (ap);
+    }
+}
+
 static int authorize_request (const flux_msg_t *msg,
                               uint32_t rank,
                               flux_error_t *error)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/modules/sdbus/sdbus.c [hunk 1]
```diff
@@ -696,18 +759,22 @@ struct sdbus_ctx *sdbus_ctx_create (flux_t *h, flux_error_t *error)
 {
     struct sdbus_ctx *ctx;
 
-    if (!(ctx = calloc (1, sizeof (*ctx)))
-        || !(ctx->f_conn = sdbus_connect (h, true, retry_min, retry_max))
+    if (!(ctx = calloc (1, sizeof (*ctx))))
+        goto error_create;
+    if (sdbus_configure (ctx, flux_get_conf (h), error) < 0)
+        goto error;
+    if (!(ctx->f_conn = sdbus_connect (h, true, retry_min, retry_max))
```
grondo, line 766: I'm probably having trouble inspecting the diff, but is the error message in `error.text` never printed if `sdbus_configure()` fails in `sdbus_ctx_create()`?

grondo (approved):
This LGTM, just one question inline about an error message during initialization (and I could be misreading due to the diff)

src/modules/sdbus/sdbus.c
```diff
@@ -56,6 +56,20 @@ static void sdbus_recover (struct sdbus_ctx *ctx, const char *reason);
 static const double retry_min = 2;
 static const double retry_max = 60;
 
+static bool sdbus_debug;
```
chu11, line 59: initialize to false just to be clear on default?

t/t2407-sdbus.t
```diff
@@ -111,9 +111,22 @@ bus_subscribe_cancel () {
     flux python -c "import flux; flux.Flux().rpc(\"sdbus.subscribe-cancel\",{\"matchtag\":$1},flags=flux.constants.FLUX_RPC_NORESPONSE)"
 }
 
+test_expect_success 'enable sdbus-debug in configuration' '
```
chu11, line 114: commit message "disable" -> "disabled by default"

src/modules/sdbus/sdbus.c
garlick: It is printed by the caller which is in main.c.  `error` is an parameter of `sdbus_ctx_create()`.

grondo: Ok, thanks. sorry I missed that.

garlick:
OK fixed those things and forced a push.  Thanks!

chu11 (approved):
LGTM too

PR #5267, opened 2023-06-14
job-exec: always use stdio for exec barrier

garlick:
Problem: testing with the (future) "sdexec" remote subprocess exec service is complicated by the fact that stdio is only used for the exec barrier when the IMP is invoked.
    
Use stdio by default, falling back to a subprocess channel only if the IMP is too old.

A couple of very minor fixes are also included.

Peeled off of #5197

### DIFF
--- a/src/modules/job-exec/exec.c
+++ b/src/modules/job-exec/exec.c
@@ -350,12 +350,12 @@ static int exec_init (struct jobinfo *job)
         goto err;
     }
     if (job->multiuser) {
-            if (config_use_imp_helper ()
-                && flux_cmd_setenvf (cmd,
-                                     1,
-                                     "FLUX_IMP_EXEC_HELPER",
-                                     "flux imp-exec-helper %ju",
-                                     (uintmax_t) job->id) < 0) {
+        if (config_use_imp_helper ()
+            && flux_cmd_setenvf (cmd,
+                                 1,
+                                 "FLUX_IMP_EXEC_HELPER",
+                                 "flux imp-exec-helper %ju",
+                                 (uintmax_t) job->id) < 0) {
             flux_log_error (job->h, "exec_init: flux_cmd_setenvf");
             goto err;
         }
--- a/src/modules/job-exec/exec.c
+++ b/src/modules/job-exec/exec.c
@@ -56,7 +56,11 @@ struct exec_ctx {
 
 static void exec_ctx_destroy (struct exec_ctx *tc)
 {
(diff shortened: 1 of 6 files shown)


chu11:
My immediate high level thought is, can we get rid of the FLUX_EXEC_PROTOCOL_FD code?  It seems that we just have to make a requirement to use `flux-security` >= 0.9.0?

garlick:
> can we get rid of the FLUX_EXEC_PROTOCOL_FD code? It seems that we just have to make a requirement to use flux-security >= 0.9.0?

Hmm, good point - I guess now that we have a tag for that flux-security development, the conditional code could go away if we just require >= that tag.  Thoughts @grondo?

grondo:
Well my thinking was that we still have systems with flux-security < 0.9.0 installed, and dropping support for it would make it so we can't even build flux-core development versions with flux-security support on these systems.

However, flux-security is now rolling out and most development is done without the `--with-flux-security` flag, so perhaps its fine now.

garlick:
Reworked to add flux-security >= 0.9.0 dependency and drop the conditional code for supporting older versions.

grondo:
Release notes-wise should we have a separate PR for "require flux-security >= 0.9.0" There's no way someone could derive that information from "always use stdio for exec barrier".

I'm actually ambivalent because we ourselves are probably the only users of flux-core+flux-security. Maybe just something to keep in mind for the future.

grondo:
Just FYI, this branch can't be built on corona with `--with-flux-security` since that system is only running flux-security 0.8.0

garlick:
> Just FYI, this branch can't be built on corona with --with-flux-security since that system is only running flux-security 0.8.0

I'll go ahead and split out the flux-security version change as a separate PR and we can decide there whether merging would have a negative impact on development.  (If so it's fine to hold off).

grondo:
I think it is probably fine, I just wanted folks to be aware that just because we tagged flux-security 0.9.0 doesn't mean it is already available everywhere.

garlick:
Rebased on master after #5270 was merged, and improved the main commit message.

grondo (approved):
Nice cleanup!

garlick:
bionic - py3.7,clang-6.0 failed here

> 2023-06-15T21:55:34.8078154Z ERROR: test_iostress.t - exited with status 137 (terminated by signal 9?)

unrelated so I'll restart that one.

chu11 (approved):
LGTM too

Issue #5266, opened 2023-06-14
`flux-resource list -i, --include=HOSTS` is incorrect when some ranks are excluded


grondo:
@ryanday36 noticed that `flux resource -i, --include` was selecting the wrong hosts when run on corona. This is due to a think-o in the implementation -- in order to get the rank of a host in Rv1 you have to find the index of the host in `nodelist`, then use that index to lookup the rank in `ranks`. This second step was forgotten, and happens to work in testing because no tests exclude a host from the scheduler resource set.

PR #5268, opened 2023-06-15
flux-resource: fix `-i, --include=HOSTS` in `list` command when some hosts are excluded

grondo:
This PR fixes the issue described #5266, when hosts are excluded by configuration, they do not appear in the `SchedResourceList` "all" resource set, and so an erroneous assumption that the index into the `nodelist` attribute is the node rank causes the wrong hosts to be included.

### DIFF
--- a/t/python/t0022-resource-set.py
+++ b/t/python/t0022-resource-set.py
@@ -41,6 +41,27 @@ class TestRSet(unittest.TestCase):
       }
     }
     """
+    R2 = """
+    {
+      "version": 1,
+      "execution": {
+        "R_lite": [
+          {
+            "rank": "10-13",
+            "children": {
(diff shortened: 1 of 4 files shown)


t/python/t0022-resource-set.py
```diff
@@ -41,6 +41,27 @@ class TestRSet(unittest.TestCase):
       }
     }
```
chu11, line 42: commit message typo "secondard" -> "second"? "secondary"?

chu11 (approved):
LGTM, just one commit message typo I noticed

t/python/t0022-resource-set.py
grondo: Thanks! Fixed and will set MWP.

PR #5269, opened 2023-06-15
support multiplicative suffixes in command options that take a bytes argument

grondo:
This is the redo of the PR to add support for SI units in several commands:
 * `flux restore --size-limit` option
 * `flux filemap --chunksize` and `--small-file-threshold` options
 * `flux ping --pad` option

This is done by using a new `optparse_get_size()` function that uses the new `parse_size()` function in libutil.

This was a rework of a rebase with much manual fixups, so hopefully not too messy.

### DIFF
--- a/doc/man1/flux-ping.rst
+++ b/doc/man1/flux-ping.rst
@@ -40,9 +40,11 @@ OPTIONS
    respectively. Default: send to “*any*”.
 
 **-p, --pad**\ *=N*
-   Include in the payload a string of length *N* bytes. The payload will be
-   echoed back in the response. This option can be used to explore the
-   effect of message size on latency. Default: no padding.
+   Include in the payload a string of length *N* bytes. The payload will
+   be echoed back in the response. This option can be used to explore
+   the effect of message size on latency. Default: no padding. *N* may be
+   specified as a floating point number with optional multiplicative suffix
+   'k' or 'K' for KiB, 'M' for MiB, and so on.
 
 **-i, --interval**\ *=N*
    Specify the delay, in seconds, between successive requests.
--- a/doc/man1/flux-restore.rst
+++ b/doc/man1/flux-restore.rst
@@ -54,7 +54,11 @@ OPTIONS
    store.  Performance will vary depending on the content of the archive.
 
 **--size-limit**\ =\ *SIZE*
-   Skip restoring keys that exceed SIZE bytes (default: no limit).
+   Skip restoring keys that exceed SIZE bytes (default: no limit). SIZE may
+   be a floating point number with an optional multiplicative suffix 'k'
+   or 'K' for for kibibytes (KiB, units of 1024 bytes), 'M' for mebibytes
+   (MiB, units of 1024 * 1024 bytes), 'G' for gibibytes (GiB, units of
+   1024 * 1024 * 1024 bytes), and so on for 'T', 'P', and 'E'.
 
 RESOURCES
 =========
--- a/src/cmd/builtin/filemap.c
+++ b/src/cmd/builtin/filemap.c
@@ -29,8 +29,8 @@
 #include "src/common/libfilemap/filemap.h"
 #include "src/common/libutil/fileref.h"
 
-static const int default_chunksize = 1048576;
-static const int default_small_file_threshold = 4096;
+static const char *default_chunksize = "1M";
+static const char *default_small_file_threshold = "4K";
 
 static json_t *get_list_option (optparse_t *p,
                                 const char *name,
@@ -175,8 +175,8 @@ static int subcmd_map (optparse_t *p, int ac, char *av[])
     }
     ctx.p = p;
     ctx.verbose = optparse_get_int (p, "verbose", 0);
-    ctx.chunksize = optparse_get_int (p, "chunksize", default_chunksize);
-    ctx.threshold = optparse_get_int (p,
+    ctx.chunksize = optparse_get_size (p, "chunksize", default_chunksize);
+    ctx.threshold = optparse_get_size (p,
                                       "small-file-threshold",
                                       default_small_file_threshold);
     ctx.disable_mmap = optparse_hasopt (p, "disable-mmap");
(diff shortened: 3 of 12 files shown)


garlick (commented):
just a couple quick notes before bed - I'll review in the morning.

doc/man1/flux-restore.rst
```diff
@@ -54,7 +54,11 @@ OPTIONS
    store.  Performance will vary depending on the content of the archive.
 
 **--size-limit**\ =\ *SIZE*
-   Skip restoring keys that exceed SIZE bytes (default: no limit).
+   Skip restoring keys that exceed SIZE bytes (default: no limit). SIZE
+   may be a floating point number with an optional SI unit 'k' or 'K'
+   for for kibibytes (KiB, units of 1024 bytes), 'M' for mebibytes (MiB,
+   unites of 1024 * 1024 bytes), 'G' for gibibytes (GiB, units of 1024 *
+   1024 * 1024 bytes), and so on up to 'E' for exbibytes (EiB).
```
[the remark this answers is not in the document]
grondo: Hm, that's going to be confusing since the letter is actually a 'suffix' in this case. Maybe I'll just drop "suffix" and avoid confusion either way, since it is pretty obvious what is meant? (Or, just do the mention of SI units at all)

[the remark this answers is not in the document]
grondo: Or maybe "followed by an SI unit prefix" would work?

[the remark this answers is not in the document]
garlick: Maybe like dd(1), "multiplicative suffixes" might be the least confusing?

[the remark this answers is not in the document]
grondo: Sounds good to me! I'll also update the PR title. Sorry about that!

garlick (approved):
Code LGTM!

I'm sure whatever you decide on on what to call the suffixes in the docs will be fine - this is likely intuitively understood anyway.

grondo:
Ok, I forced a push after rewording most everything. I've gone through this so many times, perhaps we'd better let @chu11 take an editing and review pass before I set MWP.

src/common/liboptparse/optparse.h
```diff
@@ -381,6 +382,16 @@ int optparse_get_int (optparse_t *p, const char *name, int default_value);
 double optparse_get_duration (optparse_t *p, const char *name,
                               double default_value);
 
+/*   Return the option argument parsed as a file size parameter in
+ *   floating-point 'size[bkMGTPE]', where the optional suffix designates
```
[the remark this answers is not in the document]
grondo: ah, sorry, left over from the previous version :facepalm:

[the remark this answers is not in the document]
grondo: Fixed, but I'll wait to force a push.

src/cmd/builtin/filemap.c
```diff
@@ -29,8 +29,8 @@
 #include "src/common/libfilemap/filemap.h"
```
chu11, line 29: commit message typo "thses" -> "these"

src/cmd/builtin/filemap.c
```diff
@@ -175,8 +175,8 @@ static int subcmd_map (optparse_t *p, int ac, char *av[])
     }
     ctx.p = p;
     ctx.verbose = optparse_get_int (p, "verbose", 0);
-    ctx.chunksize = optparse_get_int (p, "chunksize", default_chunksize);
-    ctx.threshold = optparse_get_int (p,
+    ctx.chunksize = optparse_get_size (p, "chunksize", default_chunksize);
+    ctx.threshold = optparse_get_size (p,
```
chu11, line 179: chunksize and threshold are ints in the context, should be updated?  Not sure of fallout elsewhere in code though.

src/cmd/builtin/filemap.c
```diff
@@ -375,9 +375,9 @@ static struct optparse_option map_opts[] = {
       .usage = "Change to DIR before mapping", },
     { .name = "verbose", .key = 'v', .has_arg = 2, .arginfo = "[LEVEL]",
       .usage = "Increase output detail.", },
-    { .name = "chunksize", .has_arg = 1, .arginfo = "N",
+    { .name = "chunksize", .has_arg = 1, .arginfo = "N[kMG]",
```
chu11, line 378: intentionally only document kMG here?  It does make sense, perhaps don't want to intentionalllly suggest exabyte chunksize.

doc/man1/flux-ping.rst
```diff
@@ -40,9 +40,11 @@ OPTIONS
    respectively. Default: send to “*any*”.
 
 **-p, --pad**\ *=N*
-   Include in the payload a string of length *N* bytes. The payload will be
-   echoed back in the response. This option can be used to explore the
-   effect of message size on latency. Default: no padding.
+   Include in the payload a string of length *N* bytes. The payload will
```
chu11, line 43: commit message, intentionally leave out T, E?

chu11 (commented):
overall LGTM, just a few small things I noticed.

src/cmd/builtin/filemap.c
garlick: Those values are ints all the way through, so I'd say the option parsing code should make sure the returned value is not greater than INT_MAX as opposed to letting it overflow.  Maybe here a helper like this would be useful
```c
int get_size_int (optparse_t *p, const char *name, const char *default_value)
{
    uint64_t val = optparse_get_size (p, name, default_value);
    if (val == (uint64_t)-1)
        return -1;
    if (val > INT_MAX) {
        optparse_fatal_usage (p, name, 1, "%s must be less than %zu\n", INT_MAX);
        return -1;
    }
    return (int)val;
}

grondo: and also redo the documentation, again... since obviously `T`, `P`, `E` suffixes are right out.

doc/man1/flux-ping.rst
grondo: As you noticed above, `pad_bytes` is also an `int` so `T`, `P`, `E` are not supported. At some point I'll redo all this with the `optparse_get_size_int()` helper suggested by @garlick.

grondo:
This needs to be redone completely with an `optparse_get_size_int()` which works more like the original PR and only takes suffixes in the set `kKMG` - I'll withdraw this one since it will need a complete re-review.

PR #5270, opened 2023-06-15
build: require flux-security >= 0.9.0

garlick:
Problem: job-exec includes code to run the pre-0.9.0 IMP.

Require 0.9.0 so we can drop that code.

This was split from #5267 for improved release notes.

Also: since 0.9.0 is only now starting to roll out in LC (and may be delayed for some time on some platforms), we could elect to delay merging this until deployments catch up a bit, if not being able to build flux-core `--with-flux-security` on those platforms would be inconvenient.  Quickly checking some systems:

- corona: flux-security 0.8.0
- fluke, tioga, rzvernal: flux-security 0.9.0


grondo (approved):
This seems good to me. Thanks for splitting this out on its own!

garlick:
I'll go ahead and set MWP.  Thanks for the discussion!

PR #5275, opened 2023-06-16
broker: avoid spurious overlay peer warning

### LINKED ISSUE #5273: Warning: unable to resolve upstream peer [: Name or service not known
Problem:  While working on #5272, @grondo noted this warning (previously suppressed):
```
 grondo@corona211:~$ flux alloc -N2 
f(s=2,d=1) grondo@corona274:~$ flux exec -r 1 flux dmesg -HL
[Jun15 19:00] broker[1]: Warning: unable to resolve upstream peer [: Name or service not known
```
The allocation actually works fine.

This check was intended to catch configuration errors (#5009).  In this case the connect URI is provided by 0mq after resolving a wildcard bind address, and is then communicated over PMI.  With ipv6 enabled, 0mq returns uris that look like this:
```
tcp://[::ffff:134.9.50.69]:49155
```
When the check tries to extract the hostname with libyuarel, it apparently returns only `[`.

This check probably doesn't help in the PMI case.  Perhaps it can be moved to boot_config.c so it only applies to the case that's prone to human error.

garlick:
Problem: the broker complains of being unable to resolve its upstream peer when bootstrapped via PMI, but the peer uri is valid.
 ```  
Warning: unable to resolve upstream peer [: Name or service not known
 ```
The URI is generated by 0mq after concretizing a wildcard bind address.  When ipv6 is enabled, 0mq puts the address in a form that cannot be parsed by yuarel, nor can it be looked up by getaddrinfo(3), e.g.
```
tcp://[::ffff:134.9.50.69]:49155
```   
Move the check into boot_config.c so that it is only performed when the URIs come from a config file.
    
Fixes #5273

### DIFF
--- a/t/t0013-config-file.t
+++ b/t/t0013-config-file.t
@@ -1,6 +1,9 @@
 #!/bin/sh
 #
 
+# Append --logfile option if FLUX_TESTS_LOGFILE is set in environment:
+test -n "$FLUX_TESTS_LOGFILE" && set -- "$@" --logfile
+
 test_description='Test config file overlay bootstrap'
 
 . `dirname $0`/sharness.sh
@@ -258,6 +261,43 @@ test_expect_success 'start size=3 instance with ipc:// and custom topology' '
 		flux getattr tbon.maxlevel >conf8a.out &&
 	echo 2 >conf8a.exp &&
 	test_cmp conf8a.exp conf8a.out
+'
+
+waitgrep() {
+	local pattern=$1
+	local file=$2
+	local iter=$3
+	while test $iter -gt 0; do
+		grep "$pattern" $file 2>/dev/null && return 0
+		sleep 0.3
+		iter=$(($iter-1))
+	done
+	return 1
+}
+
+# RFC 2606 reserves the .invalid domain for testing
+test_expect_success NO_CHAIN_LINT 'a warning is printed when upstream URI has unknown host' '
+	mkdir conf8b &&
+	cat <<-EOT >conf8b/bootstrap.toml &&
+	[bootstrap]
+	curve_cert = "testcert"
+	[[bootstrap.hosts]]
+	host = "fake0"
+	connect = "tcp://foo.invalid:1234"
+	[[bootstrap.hosts]]
+	host = "fake1"
+	EOT
+	FLUX_FAKE_HOSTNAME=fake1 \
+		flux broker -vv -Sbroker.rc1_path=,-Sbroker.rc3_path= \
+		--config-path=conf8b 2>warn.err &
+	echo $! >warn.pid &&
+	waitgrep "unable to resolve upstream peer" warn.err 30
+'
+# In case warn.pid actually refers to a libtool wrapper, try pkill(1) -P
+# first to kill its children, then kill(1).  See flux-framework/flux-core#5275.
+test_expect_success NO_CHAIN_LINT 'clean up broker from previous test' '
+	warnpid=$(cat warn.pid) &&
+	pkill -15 -P $warnpid || kill -15 $warnpid
 '
 
 getport() {
--- a/src/broker/boot_config.c
+++ b/src/broker/boot_config.c
@@ -13,6 +13,9 @@
 #if HAVE_CONFIG_H
 #include "config.h"
 #endif
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <netdb.h>
 #include <unistd.h>
 #include <stdbool.h>
 #include <assert.h>
@@ -22,6 +25,7 @@
 #include <flux/hostlist.h>
 #include <flux/taskmap.h>
 
+#include "src/common/libyuarel/yuarel.h"
 #include "src/common/libutil/log.h"
 #include "src/common/libutil/errno_safe.h"
 #include "ccan/str/str.h"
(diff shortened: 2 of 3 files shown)


grondo (approved):
Great! Much better that this in a separate PR. Thanks!

chu11:
An aside comment:

```
0mq puts the address in a form that cannot be parsed by yuarel, nor can it be looked up by getaddrinfo(3), e.g.

tcp://[::ffff:134.9.50.69]:49155
```

I believe this form is a valid IPv6 address.  So not sure if this is something we should look at handling in the long term future?  Dunno if it'd be with `libyuarel` updates or otherwise.

Not sure why it fails `getaddirinfo()`, possibly the port on the end?

garlick:
I'm guessing that 2s timeout is a little short for CI :-)  Pushing an update.

chu11:
Some researching notes:

image

> According to @grondo getaddrinfo() groks ::ffff:134.9.50.69 but not [::ffff:134.9.50.69].
> libyuarel doesn't get past the [

Looking around the interwebs the suggestion is that `getaddrinfo()` simply must have the address parsed out from the URI before being passed in (no different than it won't accept `http://`).

Skimming the header of `libyuarel`

```
/**
 * The struct where the parsed values will be stored:
 *
 * scheme ":" [ "//" ] [ username ":" password "@" ] host [ ":" port ] [ "/" ] [ path ] [ "?" query ]
 *
 * Note: to make sure that no strings are copied, the first slash "/" in the
 * path will be used to null terminate the hostname if no port is supplied.
 */
```

I guess it just doesn't support the brackets.  But it does support the port already.  So I think this is just a bug in `libyuarel`.  I'll create an issue.  Not critical, but not something we should forget.  Can create issue with `libyuarel` too.

grondo:
Just noticed that 
```sh
# Append --logfile option if FLUX_TESTS_LOGFILE is set in environment:
test -n "$FLUX_TESTS_LOGFILE" && set -- "$@" --logfile
```
Is missing at the top of `t0013-config-file.t` so no `.output` file is produced in CI for this test (though in this case you're probably right that the test is hitting the timeout).

garlick:
I'll add that - thanks.   I raised the timeout to 5s but it's still failing.   Will try 10s but this is starting to make me wonder if something else is going on.

garlick:
I pushed a revision of the test so it doesn't have to wait out the full timeout each time.  Let's see if this does any better.

t/t0013-config-file.t
```diff
@@ -260,6 +263,38 @@ test_expect_success 'start size=3 instance with ipc:// and custom topology' '
...
+	local pattern=$1
+	local file=$2
+	local iter=$3
+	while test $iter -gt 0 && ! grep "$pattern" $file 2>/dev/null; do
+		sleep 0.3
+		iter=$(($iter-1))
+	done
+	test $iter -gt 0 && return 0
+	return 1
+}
```
grondo, line 276: Would `t/scripts/waitfile.lua` work here?

garlick: That's what I started with but it seems to not work without a connection to the broker?  (That surprised me - maybe I've got it wrong)

grondo: Oh, wow, you're right. Although it only needs the handle for the reactor, that's too bad (haven't looked at it in awhile.)

grondo:
FYI - when I run `t0013-config-file.t` on this branch I end up with broker that is still resident after the test. Maybe that has something to do with the CI failures?

```
grondo   3922423  0.2  0.0 228256 12172 pts/12   Sl   14:49   0:00 /g/g0/grondo/git/flux-core.git/src/broker/.libs/lt-flux-broker -vv -Sbroker.rc1_path=,-Sbroker.rc3_path= --config-path=conf8b
```

garlick:
Pushed fixes - hopefully this will do it.  Two things
- the test had an extra argument leftover from earlier `test_must_fail 1 wait $pid` so I would guess the `1` was argv0 of the command and thus the wait was not executed, leaving the broker hanging around
- for some reason using `$pid` as a temporary variable was causing TAP test order failures on bionic.   No idea why!  Changed it to `$warnpid` and now it works there in docker :shrug: 

Now I have to end the day admitting that my butt was kicked by a 6 line shell scriptlet.   Sheesh!

garlick:
Gah, some builders are getting stuck in the new test and are canceled after 1h.  There must be a race in handling the terminating signal.  Just changed it to a SIGKILL so that should be that.

garlick:
This thing is still not right.  I've got it hanging in docker in the el8 container so I'm sure I can run it down but I may not have a lot of time for it this weekend. Another option is just dropping the test so we can move the dependent PR along, and come back with a test later.  LMK @grondo.

grondo:
Looks like you fixed it! What was the problem?

garlick:
In the test I start a broker that has to be killed after it prints the warning.  To do this, I ran it in the background and collected the pid with `$!`, then killed that pid later to clean up.

Apparently at least on the el builders, that pid corresponded to the libtool wrapper not the broker iteself and killing the wrapper did not cause the broker to be killed.

So I changed it to `pkill -P $pid || kill $pid` and that seems to work.

grondo:
Oh, wow. Well good debugging! Feel free to set MWP (my only suggestion would be a comment so we don't forget why the `pkill -P` is used).

PR #5279, opened 2023-06-20
Parse jobspec attributes.system optionally 

chu11:
Problem: Per RFC14, the "system" key under "attributes" is not required.  However, while parsing the jobspec a number of unpacking code instances did something similar to the following:

`json_unpack ("{s:{s:{s?s}}}", "attributes", "system", "foo", &s);`

The code is attempting to unpack the optional field "attributes.system.foo", but is saying "attributes" and "attributes.system"
are required.  This has the potential to lead to unintentional parsing errors.

Note that for JobspecV1 in RFC25, the "system" field is required under attributes.  This is likely why code was written this way.

Solution: Update all code like the above into a style like:

`json_unpack ("{s:{s?{s?s}}}", "attributes", "system", "foo", &s);`

Where the "system" key is optional.

Note that we do NOT update unpacking instances like this:

`json_unpack ("{s:{s:{s:s}}}", "attributes", "system", "foo", &s);`

In instances like this, "attributes.system.foo" is required for that particular circumstance.  Therefore "attributes.system" is also required.

Remove one unit test in job-list which is no longer relevant.

Also anal retentive cleanup, update all uses of `s?:` to `s?` for consistency.  I saw several circumstances where `s?:` and `s?` were both used in one line and it irritated me :P


chu11:
> make dist is failing in CI. I think maybe the test input that was removed was not removed from EXTRA_DIST or similar?

Doh!  Looks like it, thanks for noticing.  Completely missed that one of the builders failed.

> Before we merge this, should we resolve whether we need to update RFC 25?

Sounds like a good idea.  PR coming shortly.

chu11:
> Before we merge this, should we resolve whether we need to update RFC 25? attributes.system.duration is required by RFC 25 hence it states that attributes.system is required, while RFC 14 says it is optional. Maybe a little rewording in RFC 25 could resolve this "conflict"?

Hmmm.  As I was updating it, it sounded/looked weird to make this change.  We are saying `system` is optional, but the `duration` requirement below clearly indicates `system` is a requirement.

Alternate idea, perhaps when we do `json_unpack()`, "system" is required IF we're parsing the duration.  If we aren't parsing the duration, then "system" must be optional?  And can add a comment to that effect.

chu11:
rebased & re-pushed

- fixed up the make dist issue
- as I said above, I wasn't too fond of modifying RFC25 as I think the text looked weird as a result.  So instead, the one location that we parse `attributes.system.duration`, I make that a requirement in `flux-core` and add a comment.  Tweaked the code logic a tad as well to check for version == 1 before parsing `attributes.system.duration`.
- added a commit to also add a unit test to `job-list` too.

src/modules/job-list/Makefile.am
```diff
@@ -113,7 +113,6 @@ EXTRA_DIST = \
 	test/jobspec/invalid_version.jobspec \
 	test/jobspec/job_name_alt.jobspec \
 	test/jobspec/missing_attributes.jobspec \
-	test/jobspec/missing_attributes_system.jobspec \
 	test/jobspec/missing_resources.jobspec \
 	test/jobspec/missing_tasks.jobspec \
 	test/jobspec/missing_version.jobspec
```
garlick, line 119: Maybe the removal of the test input should be a standalone commit that precedes the parsing changes?

garlick (approved):
Generally LGTM.
Just one comment about maybe splitting out the removal of the jobspec test input to a separate commit but up to you on that one.

chu11:
@garlick thanks!  split out that one change into a different commit and re-pushed.

PR #5280, opened 2023-06-20
allow guests to use flux module list, flux module stats

### LINKED ISSUE #5244: flux module: consider allowing guests to use list and status subcommands
Problem: `flux module list` and `flux module stats`  are restricted to instance owner.

I can't see any issue with opening these up to guests so that system administrators don't have to sudo to run them.

garlick:
Problem: the  `flux module list` and `flux module stats` sub-commands are unnecessarily restricted to instance owner.

Open them to guests.

Fixes #5244


chu11 (approved):
LGTM

PR #5274, opened 2023-06-16
do not suppress job shell and broker errors with `flux alloc`

grondo:
Currently `flux job attach` skips watching the output eventlog when attaching to an interactive pty because all job output is presumed to go through the pty. However, this means error messages posted as `log` events to the eventlog are ignored, which can result in confusing behavior as described in #5272. It would also be valuable to see regular output from tasks, since brokers in a flux-alloc instance may also issue errors or warnings that would otherwise be suppressed.

This PR changes `flux job attach` so it always watches the output eventlog. Special handling for the pty case is required:
 - add missing `\r` for output and log messages since the terminal is assumed to be in raw mode
 - skip output for rank 0 when the pty is being captured to avoid duplication

Otherwise, the code is largely the same as the non pty case.

A fix for #5273 is also included since otherwise this error would now be seen for every `flux alloc` invocation.

A side benefit is that you can now enable shell and/or broker debugging in flux-alloc and actually see the log messages:
```console
$ flux alloc -o verbose --broker-opts=-Slog-stderr-mode=local,-Slog-stderr-level=7 -N2 --cwd=/foo
0.066s: flux-shell[1]: ERROR: host corona211: Could not change dir to /foo: No such file or directory. Going to /tmp instead
0.066s: flux-shell[1]: DEBUG: Loading /g/g0/grondo/git/flux-core.git/src/shell/initrc.lua
0.068s: flux-shell[1]: DEBUG: pmi-simple: simple wire protocol is enabled
0.073s: flux-shell[1]: DEBUG: 1: task 1 on cores 0-47                    
0.058s: flux-shell[0]: ERROR: host corona211: Could not change dir to /foo: No such file or directory. Going to /tmp instead
0.058s: flux-shell[0]: DEBUG: Loading /g/g0/grondo/git/flux-core.git/src/shell/initrc.lua
0.061s: flux-shell[0]: DEBUG: pmi-simple: simple wire protocol is enabled
0.062s: flux-shell[0]: DEBUG: output: batch timeout = 0.500s             
0.067s: flux-shell[0]: DEBUG: 0: task_count=2 slot_count=2 cores_per_slot=1 slots_per_node=1
0.067s: flux-shell[0]: DEBUG: 0: task 0 on cores 0-47
broker.debug[0]: insmod connector-local              
broker.info[0]: start: none->join 3.4426ms
broker.debug[0]: overlay auth cert-name=1 OK
broker.info[0]: parent-none: join->init 0.071131ms
[...]

PR #5282, opened 2023-06-20
shell: ensure captured pty data does not come after stdout EOF in output eventlog

grondo:
While testing #5274 (since @garlick said it looked too easy) an issue with the output eventlog was found when `pty.capture` is used with `pty.interactive`. Since the task's original stdout file descriptor is closed when attaching to a pty, the `eof` event could be posted to the output eventlog before all pty data is captured there. This would result in `data` events that follow an `eof` event. This is incorrect, and could also result in lost data if the entity processing the output eventlog stops reading or writing a stream on `eof`.

This PR fixes that by allowing a reference to be taken on subprocess output channels. The EOF from the channel is not produced until the reference count drops to zero, i.e. the actual EOF is has been read from the stream and any extra references have been dropped. This allows the pty plugin to take a reference on stdout when it is capturing pty output to ensure no `eof` event appears in the eventlog until the pty has received all data.

This approach resulted in the simplest solution, even though the reference count implementation had to be put all the way down in the buffered read watcher interface. The alternative was a complex reference counting implementation in the output shell plugin, which required plugins calling plugins with a lot more code.


garlick:
Well I didn't mean that comment as a critique - it was congratulatory.  But I guess further poking was productive :+1:

garlick (approved):
This approach is easy to understand - good solution IMHO.
LGTM.

grondo:
> Well I didn't mean that comment as a critique 

Oh, I didn't take it that way, I just meant that when you say something looked too easy, then it was probably done incorrectly.
Maybe this is similar to Murphy's 8th Law: " If everything seems to be going well, you have obviously overlooked something."

grondo:
Hm, the extra testing or this change has exposed a race in `flux job attach` with `pty.interactive` and `pty.capture`. Looking into it (befuddled at the moment though)

grondo:
> Hm, the extra testing or this change has exposed a race in flux job attach with pty.interactive and pty.capture. Looking into it (befuddled at the moment though)

Found what I thought was an unrelated issue, fixed it, and now the hang no longer reproduces for me, so I guess that was the cause? Anyway, will set MWP.

PR #5283, opened 2023-06-20
broker: load module with DSO version extension

### LINKED ISSUE #5255: loading a versioned DSO as a broker module results in confusing error
A developer was trying to load a broker module that happened to have a version `module.so.0.0.0` and the error message could perhaps be improved:

```
flux-module: load ../../src/flux_pwr_monitor/.libs/flux_pwr_monitor.so: error registering flux_pwr_monitor.so.0.0.0 service
```

The broker module loader is assuming all DSOs have exactly the suffix `.so`, but versioned modules may have extra stuff after this part. The version could just be ignored, or the loader could raise an error (unversioned dsos only please) before going on to use the full file name as the mod name (which isn't allowed because of `.`).

garlick:
broker: load module with DSO version extension
    
Problem: currently when a module is loaded by name, only "name.so" matches, while "name.so.0.0.0" does not.
    
Match "name.so*" so that a DSO can be found with or without suffixes.  The first match is used.
    
Fixes #5255


src/broker/broker.c
```diff
@@ -1246,12 +1211,22 @@ static int load_module (broker_ctx_t *ctx,
...
+                                    NULL,
+                                    NULL))
+            || zlist_size (files) == 0) {
             errprintf (error, "module not found in search path");
             errno = ENOENT;
-            return -1;
+            goto error;
         }
-        path = fullpath;
+        path = zlist_first (files);
```
chu11, line 1229: I could see a developer messing up and having multiple modules of the same name (name.so and name.so.1.0.0) in the same dir and not realizing it.  Would it worthwhile to dump/log some warning just in case?  Something simple like "multiiple name.so found, using <path>" or something?   I know the odds are low and unlikely ... just sort of thinking out loud.

chu11 (approved):
LGTM, just one "thinknig out loud" comment below.  perhaps can be ignored.

src/broker/broker.c
garlick: These files are normally generated by build systems so I would think the common case would be just  `foo.so` or `foo.so.0.0.0` and possibly some additional symlinks pointing to the same place.  If there is confusion, `flux module list -l` can show the full path to the DSO that was loaded.  Maybe enough?

chu11: if `flux module list -l` shows the full path, then that seems like more than enough!

Issue #5290, opened 2023-06-26
FAIL: t3203-instance-recovery.t 1 - start a persistent instance of size 4


chu11:
I get this error half regularly when I run a `make -j16 check` on sharness.  

```
 test_expect_success 'start a persistent instance of size 4' '
       mkdir -p test1 &&
       flux start --test-size=4 -o,-Sstatedir=$(pwd)/test1 /bin/true
'
```

Decided to start trying to debug it but not getting far.  Added some debugging into the test the test eventually gets:

`flux-start: 0 (pid 159979) Killed`

also verified bash exit code is 137, so signal 9 is presumably being received.

Added some broker log debugging

`-o,--setattr=log-filename=/tmp/achu/mylog -o,--setattr=log-level=7 \`

The thing that I notice is:

```
>grep Exited /tmp/achu/mylog
Jun 26 23:12:42.235495 broker.info[0]: rc1.0: /g/g0/achu/chaos/git/flux-framework/flux-core/etc/rc1 Exited (rc=0) 3.8s
Jun 26 23:12:43.909435 broker.info[2]: rc1.0: /g/g0/achu/chaos/git/flux-framework/flux-core/etc/rc1 Exited (rc=0) 1.7s
Jun 26 23:12:43.988406 broker.info[1]: rc1.0: /g/g0/achu/chaos/git/flux-framework/flux-core/etc/rc1 Exited (rc=0) 1.8s
Jun 26 23:12:45.902451 broker.info[3]: rc1.0: /g/g0/achu/chaos/git/flux-framework/flux-core/etc/rc1 Exited (rc=0) 1.9s
Jun 26 23:12:47.076571 broker.info[0]: rc2.0: /bin/true Exited (rc=0) 0.9s
Jun 26 23:12:48.653236 broker.info[0]: cleanup.0: flux queue stop --quiet --all --nocheckpoint Exited (rc=0) 1.6s
Jun 26 23:12:54.541536 broker.info[0]: cleanup.1: flux cancel --user=all --quiet --states RUN Exited (rc=0) 5.9s
Jun 26 23:12:57.951793 broker.info[0]: cleanup.2: flux queue idle --quiet Exited (rc=0) 3.4s
Jun 26 23:13:05.578066 broker.info[2]: rc3.0: /g/g0/achu/chaos/git/flux-framework/flux-core/etc/rc3 Exited (rc=0) 7.6s
[...]

PR #5291, opened 2023-06-27
testsuite: fix occasional broker kill error

chu11:
Problem: flux-start will kill all brokers after the first broker exits.  By default the timeout to kill all the other brokers
is 20s.  A broker in a test in t3203-instance-recovery.t would sometimes be killed due to a slow running rc3 cleanup.  This would
lead to a invalid test error.

Solution: Increase the kill timeout to 300s with the--test-exit-timeout option.

Fixes https://github.com/flux-framework/flux-core/issues/5290

Also increase the timeout in one other test that timed out.


grondo (approved):
LGTM

Issue #5233, opened 2023-06-07
libsubprocess: add interface for human readable errors


garlick:
Problem: the remote exec services (`rexec` in the broker, and `sdexec` proposed #5197) sometimes provide human readable errors in RPC responses but there is no interface in libsubprocess to retrieve them.

For example, in #5197, when setting an unknown unit property, the `sdexec.exec` fails with errno=EROFS, errstr="Cannot set property xyz , or unknown property.", and the EROFS decoded to "Read-only file system" in higher level code is not super helpful here.

Note that when the libsubprocess API was designed, RPCs could not return human readable errors, so this is a gap that has emerged as the internals have been "modernized" rather than a design flaw.

We currently have
```c
/* Returns the errno causing the FLUX_SUBPROCESS_FAILED states to be reached.
 */
int flux_subprocess_fail_errno (flux_subprocess_t *p);
```
Perhaps we could add someting analogous for human readable strings like
```c 
const char *flux_subprocess_fail_error (flux_subprocess_t *p);
```
I got a little confused trying to follow the internal error handling.  There is `p->exec_failed_errno` and `p->failed_errno`.  It wasn't clear to me why we had two separate fields and when each would be set.  (It's also possible that in my rework of the remote stuff I misunderstood how things work and created this confusion).  Anyway could use some design help on this one.

chu11:
> I got a little confused trying to follow the internal error handling. There is p->exec_failed_errno and p->failed_errno. It wasn't clear to me why we had two separate fields and when each would be set.

Going back into the history, I think the goal was to make `flux_exec()` and `flux_rexec()` "look the same" no matter if a subprocess was launched locally or remotely.  But the one corner case on that was on non-exec(2) errors, like lets say EPROTO or ENOSYS errors or something, b/c they are impossible locally.

So I think `exec_failed_errno` and `failed_errno` were a minor attempt to differentiate between those two errors, with the former related to the FLUX_SUBPROCESS_EXEC_FAILED state.

Maybe this was a dumb design choice.  But it does seem to have been refactored away over time, so we can just remove `exec_failed_errno` now.

chu11:
> Anyway could use some design help on this one.

Without looking into this with a lot of detail, if we remove `exec_failed_errno`, I think setting / transmitting the failed error string should largely follow the setting / transmitting of `failed_errno`?

garlick:
OK so maybe the steps are
[...]

PR #5294, opened 2023-06-27
libsubprocess: support user friendly error string 

chu11:
Problem: When a subprocess fails the "failed_errno" value is set and can be returned in `flux_subprocess_fail_errno()`.  It would
be convenient if there was an error message to describe the error in more detail.

Solution: Internally support a new "failed_error" field which can store a text message describing the error in more detail.  Generally speaking, always fill this with a text message when "failed_errno" is also set.

Set user friendly error string in many cases throughout libsubprocess.

Support a new `flux_subprocess_fail_error()` to retrieve this error message.

Add corner case unit tests.

Fixes https://github.com/flux-framework/flux-core/issues/5233

----

Extra notes:

I think the biggest review question in this PR will be my choice of "error message" throughout.  And if I should add it in more places.  We capture the big ones of EACCESS / EPERM on a remotely executed process. e.g.

```
flux-exec: Error: rank 0: ./nosuchprocess: error launching process: No such file or directory
```

### DIFF
--- a/src/common/libsubprocess/subprocess.c
+++ b/src/common/libsubprocess/subprocess.c
@@ -1121,6 +1121,17 @@ int flux_subprocess_fail_errno (flux_subprocess_t *p)
     return p->failed_errno;
 }
 
+const char *flux_subprocess_fail_error (flux_subprocess_t *p)
+{
+    if (!p)
+        return "internal error: subprocess is NULL";
+    if (p->state != FLUX_SUBPROCESS_FAILED)
+        return "internal error: subprocess is not in FAILED state";
+    if (p->failed_error.text[0] == '\0')
+        return strerror (p->failed_errno);
+    return p->failed_error.text;
+}
+
 int flux_subprocess_status (flux_subprocess_t *p)
 {
     if (!p) {
--- a/src/cmd/flux-exec.c
+++ b/src/cmd/flux-exec.c
@@ -157,12 +157,17 @@ void state_cb (flux_subprocess_t *p, flux_subprocess_state_t state)
     if (state == FLUX_SUBPROCESS_FAILED) {
         flux_cmd_t *cmd = flux_subprocess_get_cmd (p);
         int errnum = flux_subprocess_fail_errno (p);
+        const char *errmsg = flux_subprocess_fail_error (p);
         int ec = 1;
 
+        /* N.B. if no error message available from
+         * flux_subprocess_fail_error(), errmsg is set to strerror of
+         * subprocess errno.
+         */
         log_msg ("Error: rank %d: %s: %s",
                  flux_subprocess_rank (p),
                  flux_cmd_arg (cmd, 0),
-                 strerror (errnum));
+                 errmsg);
 
         /* bash standard, 126 for permission/access denied, 127 for
          * command not found.  68 (EX_NOHOST) for No route to host.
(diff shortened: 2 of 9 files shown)


chu11:
oops, spelling error, quick fix, squash, re-push

src/common/libsubprocess/subprocess.c
```diff
@@ -1121,6 +1121,21 @@ int flux_subprocess_fail_errno (flux_subprocess_t *p)
...
+        return NULL;
+    }
+    if (p->state != FLUX_SUBPROCESS_FAILED) {
+        errno = EINVAL;
+        return NULL;
+    }
+    if (p->failed_error.text[0] == '\0')
+        return strerror (p->failed_errno);
+    return p->failed_error.text;
+}
```
garlick, line 1137: I'm not sure I know all the places where this will be used but sometimes it's inconvenient to return NULL from a function like this.  Would it be better to to just return textual errors like
- `internal error: subprocess is NULL`
- `internal error: subprocess is not in FAILED state`

then the user can just use the result in a printf argument without checking?

garlick (approved):
:tada: 

Nice work!  This will be helpful.  I had a few comments inline, questions really.  But this generally looks great.  Thanks!

src/common/libsubprocess/client.c
```diff
@@ -130,7 +130,7 @@ flux_future_t *subprocess_rexec (flux_t *h,
     return NULL;
 }
 
-int subprocess_rexec_get (flux_future_t *f)
+int subprocess_rexec_get (flux_future_t *f, flux_error_t *error)
```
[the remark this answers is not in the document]
chu11: sounds good.  I had it originally but thought "ehhh there could be more detail", guess I went too far.

src/common/libsubprocess/client.c
```diff
@@ -157,7 +157,7 @@ int subprocess_rexec_get (flux_future_t *f)
     }
     else if (!streq (ctx->response.type, "started")
         && !streq (ctx->response.type, "stopped")
-        && !streq (ctx->response.type, "finished")) {
+             && !streq (ctx->response.type, "finished")) {
```
garlick, line 160: whoops?

garlick (approved):
A typo crept in?  o/w looks good!

src/common/libsubprocess/client.c
chu11: doh! good catch

chu11:
re-pushed with fixes per comments above, also ...

- I added an error message for "internal fatal" errors, perhaps is unnecessary but just in case.

```C
@@ -136,6 +137,7 @@ static void proc_internal_fatal (flux_subprocess_t *p)
      */
     p->state = FLUX_SUBPROCESS_FAILED;
     p->failed_errno = errno;
+    errprintf (&p->failed_error, "internal fatal error: %s", strerror (errno));
     state_change_start (p);
```

- I think an error message of "NULL" is different than "empty string", so did this tweak to what was there before

```C
     else if (state == FLUX_SUBPROCESS_FAILED) {
+        const char *errmsg = NULL;
+        if (p->failed_error.text[0] != '\0')
+            errmsg = p->failed_error.text;
         rc = flux_respond_error (s->h,
                                  request,
                                  p->failed_errno,
-                                 p->failed_error.text);
+                                 errmsg);
```

alternately I could call `flux_subprocess_fail_errno(p)` and `flux_subprocess_fail_error(p)` so there would always be a message to
send.

- I noticed English errors in the tests, where I said "inputs" where it should be "input".  Dumb new commit added.

Issue #5257, opened 2023-06-13
job-exec: shell service name reconstruction may be fragile to mismatched locale configuration


garlick:
Problem: as noted in #5256, the job-exec module reconstructs the shell's service name to send the rank 0 shell an exception message when one of the other job shells goes offline, but when f58 encoding is used, this becomes somewhat fragile to mismatched locale configuration, especially as the job-exec module is further developed into a distributed service.

There may be other ways to notify rank 0 or to obtain the service name without reconstructing it.  This issue is for considering alternatives.

grondo:
A few alternatives mentioned in #5256:

 - Make it unnecessary for job-exec to send an RPC directly to the job shell by having the job shell watch for lost shell exception in the job eventlog. The shell would have to parse the affected shell rank from the exception note.
 - The job shell could respond with its service name in the barrier protocol instead of just writing `enter`.
 - The job-exec module could fetch the service name from the exec eventlog only when necessary.

garlick:
Hmm, maybe for non UX sitautions such as this one, we ought to have an "f58 plain" encoding that always uses f instead of ƒ?

I'd also kind of like that so I can use f58 encoded job IDs in system unit names.  (systemd balks at ƒ).

grondo:
That's a great idea!

garlick:
Fixed by #5297

PR #5297, opened 2023-06-29
add f58plain job encoding

garlick:
Problem: an f58 job encoding that does not change depending on locale configuration is potentially useful and a solution to the problem discussed in #5257.

This PR introduces `f58plain` encoding and uses it in the shell job service name.

### DIFF
--- a/src/modules/job-exec/job-exec.c
+++ b/src/modules/job-exec/job-exec.c
@@ -171,12 +171,14 @@ flux_future_t *jobinfo_shell_rpc_pack (struct jobinfo *job,
     char *shell_topic = NULL;
     flux_future_t *f = NULL;
     uint32_t rank;
+    char idbuf[21];
 
-    if (asprintf (&shell_topic,
-                  "%ju-shell-%s.%s",
-                  (uintmax_t) job->userid,
-                  idf58 (job->id),
-                  topic) < 0
+    if (flux_job_id_encode (job->id, "f58plain", idbuf, sizeof (idbuf)) < 0
+        || asprintf (&shell_topic,
+                     "%ju-shell-%s.%s",
+                     (uintmax_t) job->userid,
+                     idbuf,
+                     topic) < 0
         || ((rank = resource_set_nth_rank (job->R, 0)) == IDSET_INVALID_ID))
         goto out;
     va_start (ap, fmt);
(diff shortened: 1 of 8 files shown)


src/modules/job-exec/job-exec.c
```diff
@@ -171,12 +171,14 @@ flux_future_t *jobinfo_shell_rpc_pack (struct jobinfo *job,
     char *shell_topic = NULL;
     flux_future_t *f = NULL;
     uint32_t rank;
-
-    if (asprintf (&shell_topic,
-                  "%ju-shell-%s.%s",
-                  (uintmax_t) job->userid,
-                  idf58 (job->id),
-                  topic) < 0
+    char idbuf[21];
```
chu11, line 174: commit message "porblem"

chu11 (approved):
LGTM except for typo below.

I guess only question is if there's a better word than "f58plain", although I honestly can't think of one.  "f58ascii" sounds worse.

src/modules/job-exec/job-exec.c
```diff
@@ -171,12 +171,14 @@ flux_future_t *jobinfo_shell_rpc_pack (struct jobinfo *job,
...
+
+    if (flux_job_id_encode (job->id, "f58plain", idbuf, sizeof (idbuf)) < 0
```
grondo, line 176: Would it make sense to also have `idf58plain()` convenience function to simplify here and the shell's `build_topic()`?

garlick:
Yeah I could think of a better one either.  I fixed that typo.

grondo (approved):
Seems good to me! I wondered if `f58ascii` would be more specific, but it hardly matters so I don't really have an opinion.

One suggestion inline, but feel free to take it or leave it.

grondo:
> I wondered if f58ascii would be more specific,

Ok I guess two votes that is worse (too literal?)

src/modules/job-exec/job-exec.c
garlick: It felt to me like usage of this would be fairly limited, so probably not worth doing that?   Maybe later if that turns out to not be the case?

garlick:
> Ok I guess two votes that is worse (too literal?)

I don't think either of the two votes was a strong one.  Yeah, seemed kinda literal.  If we're good with what's here then I'll set MWP.

chu11:
> Ok I guess two votes that is worse (too literal?)

Yeah, just sounds too literal.

PR #5299, opened 2023-06-29
shell: document signal race

### LINKED ISSUE #5210: FAIL: t2800-jobs-cmd.t 61 - flux-jobs --format={waitstatus},{returncode}
during large parallel make checks I've recently gotten this error twice, the error comes from the fact that the `returncodeI.out` output is:

```
-,-128
36352,142
32512,127
36608,143
32512,127
0,0
0,0
0,0
0,0
```

but expected is

```
-,-128
36352,142
36608,143
36608,143
32512,127
0,0
0,0
0,0
0,0
```

The invalid `32512,127` says that the submitted command was not found, instead of the command being terminated via an exception (which is `36608,143`).  The failing job is:

```
        #
        #  Run a job that we will end with a user exception, copy its JOBID to both
        #       inactive and failed and exception lists.
        #
        jobid=`flux submit --wait-event=start sleep inf` &&
        flux job raise --type=myexception --severity=0 -m "myexception" $jobid &&
        fj_wait_event $jobid clean &&
        echo $jobid >> inactiveids &&
        flux job id $jobid > exception.ids &&
        flux job id $jobid >> failedids &&
```

I presently don't see how something busy / racy could make this fail with "command not found".

Will just monitor this for the time being.

chu11:
Per discussion in #5210, but I'm not 100% sure this completely fixes #5210, but it is a fix nonetheless.

Problem: If a trapped signal (SIGTERM, SIGINT, SIGALRM) is caught before all shell tasks have been launched, unexpected behavior could happen.  For example, some tasks could be run without being sent the signal.  The signal can also interrupt initial setup and cause other problems.

Solution: In this unlikely situation, kill current tasks via SIGKILL and exit.  If this situation is hit outside of testing conditions, it is assumed the job was messed up and we want to just kill it.

### DIFF
--- a/src/shell/log.c
+++ b/src/shell/log.c
@@ -364,6 +364,7 @@ int shell_log_init (flux_shell_t *shell, const char *progname)
     logger.level = FLUX_SHELL_NOTICE;
     logger.fp_level = FLUX_SHELL_NOTICE;
     logger.active = 0;
+    logger.exception_logged = 0;
     logger.fp = stderr;
     logger.rank = -1;
     if (progname && !(logger.prog = strdup (progname)))
--- a/src/shell/signals.c
+++ b/src/shell/signals.c
@@ -17,6 +17,25 @@
  * SIGTERM - forward
  * SIGALRM - forward
  *
+ * Notes:
+ *
+ * By setting up the signal watchers during "shell.init", there is the
+ * potential for inconsistent exit codes if a signal is received before all
+ * tasks have started.  For example, this could be seen with something
+ * like:
+ *
+ * jobid=`flux submit -n1000 foo.sh`
+ * flux job raise --type=foo --severity=0 $jobid
+ *
+ * i.e. raise sends SIGTERM to job/shell immediately after starting,
+ * but due to the large task count of 1000, the signal is received
+ * before tasks are all setup.  Some tasks could receive SIGTERM while
+ * some (to be created ones) do not.
+ *
+ * Note that the shell should always return an error, but the error
+ * may not be consistent.  This situation is extremely rare and only
+ * seen is testing situations such as the above.  So we elect to not
+ * fix this race.
  */
 #define FLUX_SHELL_PLUGIN_NAME "signals"
 
--- a/src/shell/signals.c
+++ b/src/shell/signals.c
@@ -15,6 +15,7 @@
  *
  * SIGINT  - forward to all local tasks
  * SIGTERM - forward
+ * SIGALRM - forward
  *
  */
 #define FLUX_SHELL_PLUGIN_NAME "signals"


src/shell/signals.c
```diff
@@ -25,15 +25,66 @@
...
+     * all the currently running tasks with the desired signal.
+     * However, given how racy this situation is, we can't be sure the
+     * tasks have setup their own signal handlers yet.  So we elect to
+     * just kill those tasks and exit.  The presumption is that if
+     * this situation were ever hit in normal operating conditions, it
+     * is because the job was messed up and killing the job is the
+     * desired result.
+     */
+    if (taskcount < ntasks) {
+        shell_log_errno ("signal %d caught before task(s) setup, exiting", sig);
```
grondo, line 83: Hm, instead of all this, I wonder if it would be cleaner to just call `shell_die (128+sig, "signal %d sent before tasks started")`, since that has the intended effect without having to unpack shell info while handling a signal, etc.

chu11: Yeah, as I thought about the circumstances more, there were a lot of "ohh but we should really check for ..."

Certainly could do `shell_die()` instead of my `signal_exit()` call though, didn't think to do that.

grondo:
Thinking about this last night, the job shell starts all the tasks serially outside of any `flux_reactor_run()` call. So is it even possible for only a subset of tasks to be started when the signal callback is called? It doesn't seem like it. It seems like the signal could only be handled when either no tasks or all tasks have been started. Needs a bit more research, but this could simplify the implementation a bit.

chu11:
> So is it even possible for only a subset of tasks to be started when the signal callback is called? It doesn't seem like it. It seems like the signal could only be handled when either no tasks or all tasks have been started. Needs a bit more research, but this could simplify the implementation a bit.

Ahhh, I think you're right.  In my mind b/c we pass the reactor into `libsubprocess`, it was being run internally in there.  But it's not for locally launched processes.

And since the signal handler will only be called once the reactor is being run, which is after the tasks are setup, this code sort of doesn't matter.  Thus why this fix didn't seem to fix the issue per my comment in #5210.

Hmmm, my first thought is should we setup "traditional" signal catchers before tasks are setup and then setup the reactor signal handlers after tasks are setup (perhaps via `shell.init` and `shell.start` callbacks)?

chu11:
> Hmmm, my first thought is should we setup "traditional" signal catchers before tasks are setup and then setup the reactor signal handlers after tasks are setup (perhaps via shell.init and shell.start callbacks)? Edit: with this method we don't need to do any of the checking for taskcount < ntasks and stuff.

Doh, this approach did not work out.

```
>flux run nosuchcommand
0.073s: flux-shell[0]: FATAL: task 0 (host catalyst159): start failed: nosuchcommand: No such file or directory
0.075s: flux-shell[0]: FATAL: signals: signal 15 sent before tasks started
0.074s: job.exception type=exec severity=0 task 0 (host catalyst159): start failed: nosuchcommand: No such file or directory
0.076s: job.exception type=exec severity=0 signal 15 sent before tasks started
flux-job: task(s) exited with exit code 143
```

the SIGTERM from the first exception arrives before the shell can exit with 127 for "no such command".

grondo:
> the SIGTERM from the exception arrives before the shell can exit with 127 for "no such command".

It might be better to keep the simple reactor based signal handler? The shell enters the reactor to wait for the barrier, so if you get a signal when there are 0 tasks started, then you know the shell was killed early and you can exit immediately. If you get a signal when all tasks are launched, then you take the normal approach of forwarding the signal to the tasks.

chu11:
(uh oh, did my response get lost, hopefully it doesn't show up twice later)

> It might be better to keep the simple reactor based signal handler? The shell enters the reactor to wait for the barrier, so if you get a signal when there are 0 tasks started, then you know the shell was killed early and you can exit immediately. If you get a signal when all tasks are launched, then you take the normal approach of forwarding the signal to the tasks.

If the signal arrives during `shell_barrier()` I think this is true.  But given the error message from `shell_task_start()`, the assumption is the signal occurred after `shell_barrier()` and before `flux_reactor_run()`.

grondo:
> But given the error message from shell_task_start(), the assumption is the signal occurred after shell_barrier() and before flux_reactor_run().

But we've already assumed that the shell can't handle a signal until `flux_reactor_run()` is called.

Is the signal perhaps being delivered to the actual task process after `fork()` but before `exec()`?

grondo:
> Is the signal perhaps being delivered to the actual task process after fork() but before exec()?

Hm, I guess that can't be possible since before exec() (and maybe after?) the signals should still be blocked. I guess I'm still confused about what exactly is happening, and how adding a new signal handler to the shell would fix it. (Maybe I need to reread above)

chu11:
> But we've already assumed that the shell can't handle a signal until flux_reactor_run() is called. (er, I should say outside of flux_reactor_run())

My assumption is that in the guts of `libev` there is a call to `signal(2)` or equivalent that is capturing the signal.  So the signal is still captured before we enter the reactor, but we don't get a callback for it until the reactor is run.

But since we're not in the reactor that point, anything else that goes on like `read(2)`, etc. is prone to this signal interrupt and could return EINTR.  My assumption is that the ENOENT is just a "fallout" errno from this event and not the "true" error.

Maybe something wrong in these assumptions?  It's why I thought we had to setup a `signal(2)` handler outside of the reactor.

chu11:
just re-pushed my attempt number two

- commit to not be able to call `flux_shell_fatal()` more than once.
- per comment above, `signal(2)` used for signal handling until `shell.start` setup is called.

grondo:
> My assumption is that in the guts of libev there is a call to signal(2) or equivalent that is capturing the signal. So the signal is still captured before we enter the reactor, but we don't get a callback for it until the reactor is run.

> But since we're not in the reactor at the point of the error in https://github.com/flux-framework/flux-core/issues/5210, anything else that goes on like read(2), etc. is prone to this signal interrupt and could return EINTR. My assumption is that the ENOENT is just a "fallout" errno from this and not the "true" error.

I'm pretty sure libev uses `signalfd`. In this case are blocking system calls interrupted? I can't get a clear answer. Anyway, since the `errno` here is `ENOENT` it isn't convincing that an interrupted system call is the problem here.

Even if you install an old-style signal handler, this doesn't solve the possible `EINTR` problem does it? Is the idea that we just avoid the problem completely by terminating the job shell immediately instead of on next entry to the reactor.

Is posix_spawn() being used when this issue occurs? I still wonder if there's a race between starting the child and exec(). If posix_spawn() is using vfork() this could cause random badness. (Though I'm sure that code has been carefully vetted so grasping at sraws here.)

src/shell/log.c
```diff
@@ -364,6 +364,7 @@ int shell_log_init (flux_shell_t *shell, const char *progname)
     logger.level = FLUX_SHELL_NOTICE;
     logger.fp_level = FLUX_SHELL_NOTICE;
     logger.active = 0;
+    logger.exception_logged = 0;
```
grondo, line 386: Commit message: is it supposed to be "initialize field to 0"

src/shell/log.c
```diff
@@ -282,11 +283,29 @@ void flux_shell_fatal (const char *component,
     char buf [4096];
     va_list ap;
 
+    if (logger.fatal_in_progress) {
+        if (shell->verbose > 0) {
```
grondo, line 287: Suggestion: just log the second fatal message so we don't lose any possible debug/error information, but return early before invoking other fatal error actions.

src/shell/signals.c
```diff
@@ -29,6 +29,42 @@
...
+ * flux job raise --type=foo --severity=0 $jobid
+ *
+ * i.e. raise sends SIGTERM sent to job/shell immediately after
+ * starting and before tasks setup.  This should not be seen under
+ * normal operating circumstances.
+ *
+ * This race can lead to unexpected errors due to the signal
+ * interrupting initial setup.  It is hard to predict this situation's
+ * behavior, so we elect to trap these signals and exit, rather than
+ * forward to tasks.
```
grondo, line 51: In the commit message

> Problem: If a trapped signal (SIGTERM, SIGINT, SIGALRM) is sent
> before all shell tasks have been launched, unexpected behavior could
> happen due to signal interruption of task setup (e.g. a write(2)
> call could be interrupted and return EINTR, etc.).

This is a guess at this point so IMO should be left out the commit message. Instead, maybe just say that an error could occur for unknown reason when a signal is delivered before tasks start (or while they are being started), so we paper over the error by exiting immediately.

src/shell/signals.c
```diff
@@ -29,6 +29,42 @@
...
+ * we setup signal handlers for the reactor via the "shell.start" callback.
+ */
+static void set_init_signal_callbacks (void)
+{
+    if (signal (SIGINT, init_signal_cb) == SIG_ERR)
+        shell_log_errno ("failed to set up signal handler");
+    if (signal (SIGTERM, init_signal_cb) == SIG_ERR)
+        shell_log_errno ("failed to set up signal handler");
+    if (signal (SIGALRM, init_signal_cb) == SIG_ERR)
+        shell_log_errno ("failed to set up signal handler");
```
grondo, line 65: combine these to remove `shell_log_errno` repetition?

src/shell/signals.c
```diff
@@ -65,6 +99,25 @@ static int signals_init (flux_plugin_t *p,
...
+    if (!shell)
+        return -1;
+
+    set_init_signal_callbacks ();
+
+    if (flux_plugin_add_handler (p,
+                                 "shell.start",
+                                 trap_signals,
+                                 shell) < 0)
+        shell_die_errno (1, "flux_plugin_add_handler");
```
grondo, line 117: Hm, this introduces a small race between when tasks start and the signal watcher gets set up. Are we just creating another problem? Tasks could be started and the shell will just call shell_die() if a signal is received in this window.

grondo (commented):
Just some early comments.

chu11:
> Even if you install an old-style signal handler, this doesn't solve the possible EINTR problem does it? Is the idea that we just avoid the problem completely by terminating the job shell immediately instead of on next entry to the reactor.

Yeah, the idea was the avoid the problem and make sure we exit with the desired exit code of 143 vs. whatever might result otherwise.

> Is posix_spawn() being used when this issue occurs? I still wonder if there's a race between starting the child and exec(). If posix_spawn() is using vfork() this could cause random badness. (Though I'm sure that code has been carefully vetted so grasping at sraws here.)

I assume it is.  Have seen this on both TOSS3/TOSS4 so not sure if it's an older glibc problem either.

src/shell/signals.c
chu11: Yes, it's something I did realize and it perhaps needs a comment somewhere.  Given the potential unknown error that could occur under the current circumstance, I figure "extending the window" of when we catch SIGTERM to just terminate was an acceptable compromise.  It's also b/c the race is pretty rare and unlikely.

chu11: I should also probably add a note in `signals.c` that `shell_die()` will call `flux_shell_killall()` internally.

grondo:
Ok, well it is hard to argue against these changes if it fixes the problem! I wish we know the root cause, bit we've got bigger problems. 👍

chu11:
> Ok, well it is hard to argue against these changes if it fixes the problem! I wish we know the root cause, bit we've got bigger problems. +1

Well I presume it does ... i've seen the race hit like 5-6 times total over gajillions of runs ... so pretty rare.

src/shell/signals.c
grondo: Makes sense, thanks!

src/shell/log.c
chu11: the reason I did this was b/c something as simple as `flux run nosuchcommand` would trigger two fatal error messages.  Think that's ok?  Or perhaps I can adjust the second fatal output message to indicate "maybe not the immportant one" in some way.

grondo: Oh, ugh :frowning_face: Yeah, I guess suppressing the error is best then.

chu11:
re-pushed with fixes per comments above.  Per a comment I made in #5210, I think this fix makes things better, but doesn't appear to eliminate all races.  So from a certain perspective this could be viewed as "excess" since it's a rare race to begin with.  But I think catching more race possibilities is better than less.

grondo:
Hm, coverage tests hung in `t2608-job-shell-log.t` in a fatal error test.

Also another test failed, but looks unrelated. The hang makes me nervous since that's a lot worse than returning the wrong exit code, but hard to tell if its really related.

chu11:
> Also another test failed, but looks unrelated. The hang makes me nervous since that's a lot worse than returning the wrong exit code, but hard to tell if its really related.

Given what I discovered in my notes in #5210 I'm ok with shelving it for the time being.  It is quite rare and would require herculean efforts to fix 100%.

Perhaps the only really important thing is to add some comments into `signals.c` to indicate that this race does exist so we don't forget about it the next time this is encountered.

chu11:
re-pushed, just decided to add a comment about this rather than try to fix it ... since it is not a complete solution and has question marks around it.  I also added a commit for a dumb documentation thing I noticed.

grondo (approved):
Nice work, sorry you had to go through all that!
This at least documents the situation.

Issue #5301, opened 2023-07-03
flux job wait-event: support a "waitcreate" option


chu11:
Per a discussion in #5210 running this on a pending job:

`flux job wait-event -p guest.output fB5t4XaRo data`

resulted in

`flux-job: eventlog path guest.output not found`

when the job started.  I suspect b/c there is that small window of time where the guest namespace exists but the `output` directory has not yet been created by the job shell.  It would be nice if it could "waitcreate" similar to the `flux kvs get` option.

Per a test in `t2607-job-shell-input.t` a check for `shell.init` first would avoid this.

`flux job wait-event -p guest.exec.eventlog $id shell.init 2> pipe6B.err`

Edit; Or should "waitcreate" be the default????  hmmmm
Edit Note: most tests test against `guest.exec.eventlog`, which I believe is created in `job-manager` land, not the shell, thus avoiding this raciness.

Edit Note2: also added to t2800-jobs-cmd.t via #5302

PR #5302, opened 2023-07-03
testsuite: workaround job start signal race

### LINKED ISSUE #5210: FAIL: t2800-jobs-cmd.t 61 - flux-jobs --format={waitstatus},{returncode}

chu11:
Problem: In several tests in t2800-jobs-cmd.t, a job is sent a signal immediately after the job starts.  There is a rare race that can occur when signaling a job very quickly after it has started.  This race can lead to unexpected results.

Solution: Instead of calling "sleep inf", run a script that will echo some output then call "sleep inf".  Ensure that data has been output from the job before signaling it.  This will ensure that the job has started and the process is fully running before being sent a signal.

Fixes #5210

### DIFF
--- a/t/t2800-jobs-cmd.t
+++ b/t/t2800-jobs-cmd.t
@@ -47,6 +47,15 @@ test_expect_success 'configure testing queues' '
 	flux queue start --all
 '
 
+test_expect_success 'create helper job submission script' '
+	cat >sleepinf.sh <<-EOT &&
+	#!/bin/sh
+	echo "job started"
+	sleep inf
+	EOT
+	chmod +x sleepinf.sh
+'
+
 test_expect_success 'submit jobs for job list testing' '
 	#  Create `hostname` and `sleep` jobspec
 	#  N.B. Used w/ `flux job submit` for serial job submission
@@ -82,7 +91,12 @@ test_expect_success 'submit jobs for job list testing' '
 	#  Run a job that we will end with a signal, copy its JOBID to both inactive and
 	#   failed and terminated lists.
 	#
-	jobid=`flux submit --wait-event=start sleep inf` &&
+	# N.B. sleepinf.sh and wait-event on job data to workaround
+	# rare job startup race.  See #5210
+	#
+	jobid=`flux submit ./sleepinf.sh` &&
+	flux job wait-event -p guest.exec.eventlog $jobid shell.init &&
+	flux job wait-event -p guest.output $jobid data &&
 	flux job kill $jobid &&
 	fj_wait_event $jobid clean &&
 	echo $jobid >> inactiveids &&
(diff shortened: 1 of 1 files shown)


chu11:
Hmmm, for some reason this change doesn't work at all in CI ... i have no idea why right now.

```
   1688410053.485321 shell.init service="1001-shell-f2C9HWHm" leader-rank=0 size=1
  flux-job: event 'data' never received
```

t/t2800-jobs-cmd.t
```diff
@@ -82,7 +91,12 @@ test_expect_success 'submit jobs for job list testing' '
 	#  Run a job that we will end with a signal, copy its JOBID to both inactive and
 	#   failed and terminated lists.
 	#
-	jobid=`flux submit --wait-event=start sleep inf` &&
+	# N.B. sleepinf.sh and wait-event on job data to workaround
+	# rare job startup race.  See #5210
+	#
+	jobid=`flux submit ./sleepinf.sh` &&
```
grondo, line 97: You need to keep the `--wait-event=start` option because the `guest.exec.eventlog` won't appear until the job is started.

chu11: i think the job info module takes care of this.  We have a test for it in `t2231-job-info-eventlog-watch.t`

```
# In order to test watching a guest event log that does not yet exist,
# we will start a job that will take up all resources.  Then start
# another job, which we will watch and know it hasn't started running
# yet. Then we cancel the initial job to get the new one running.

test_expect_success 'job-info: generate jobspec to consume all resources' '
        flux run --dry-run -n4 -c2 sleep 300 > sleeplong-all-rsrc.json
'

test_expect_success NO_CHAIN_LINT 'flux job wait-event -p guest.exec.eventlog works (wait job)' '
```

grondo: Oh, nice. I didn't realize that. Does it support this for the `guest.output` eventlog as well?

grondo: I guess not:
```
$ flux job wait-event -v -p guest.output $(flux submit hostname) shell.init 
flux-job: eventlog path guest.output not found
```
Does that seem kind of confusing that job-info supports waiting for one standard eventlog but not the other?

Supporting this would really simplify support for watching the output eventlog.

chu11: Yeah, I created an issue for it this morning:  https://github.com/flux-framework/flux-core/issues/5301

I believe the issue is b/c `guest.output` is created by the shell not the job-manager, so there's that window of time that the guest namespace exists but that specific eventlog doesn't.

grondo (approved):
Nice! LGTM!

chu11:
that was a very unexpected failure in py3.8 builder

```
 Warning: Found 1 errors from 89 tests in testsuite
Error: ERROR: test_iostress.t - missing test plan
Error: ERROR: test_iostress.t - exited with status 137 (terminated by signal 9?)
./src/common/libsubprocess/test_iostress.log
  ./test_iostress.t timed out after 300.0s
  ERROR: test_iostress.t - missing test plan
  ERROR: test_iostress.t - exited with status 137 (terminated by signal 9?)
ls: cannot access 't/*.trs': No such file or directory
```

maybe something in builder went wrong thus the 300 second timeout? restarting it.

PR #5304, opened 2023-07-04
job-manager: add alloc-check plugin

garlick:
This adds a jobtap plugin called `alloc-check` that raises an exception on a job when it is allocated resources by the scheduler that were allocated to another job and have not yet been freed.  I thought this would be useful while chasing flux-framework/flux-sched#1043.

Example - reproducing the above issue:
```toml
# test.toml
[job-manager]
plugins = [
  { load = "perilog.so" },
  { load = "alloc-check.so" },
]
epilog.command = [ "flux", "perilog-run", "epilog", "-e", "sleep,2" ]
```

```console
$ flux start -o,--config-path=test.toml
ƒ(s=1,d=0) $ flux module remove sched-simple
ƒ(s=1,d=0) $ flux module load sched-fluxion-resource
ƒ(s=1,d=0) $ flux module load sched-fluxion-qmanager

ƒ(s=1,d=0) $ for i in $(seq 2); do flux run -N1 -x -t1s sleep inf; done
1.045s: job.exception type=timeout severity=0 resource allocation expired
flux-job: task(s) exited with exit code 142
0.039s: job.exception type=alloc-check severity=0 resources already allocated
1.059s: job.exception type=timeout severity=0 resource allocation expired
flux-job: task(s) exited with exit code 142
ƒ(s=1,d=0) $
```

### DIFF
--- a/src/modules/job-manager/Makefile.am
+++ b/src/modules/job-manager/Makefile.am
@@ -18,6 +18,7 @@ noinst_LTLIBRARIES = \
 jobtap_plugin_LTLIBRARIES = \
 	plugins/submit-hold.la \
 	plugins/alloc-bypass.la \
+	plugins/alloc-check.la \
 	plugins/perilog.la
 
 libjob_manager_la_SOURCES = \
@@ -87,6 +88,15 @@ plugins_alloc_bypass_la_LIBADD = \
 	$(top_builddir)/src/common/libflux-internal.la \
 	$(top_builddir)/src/common/librlist/librlist.la
 plugins_alloc_bypass_la_LDFLAGS = \
+	$(fluxplugin_ldflags) \
+	-module
+
+plugins_alloc_check_la_SOURCES = \
+	plugins/alloc-check.c
+plugins_alloc_check_la_LIBADD = \
+	$(top_builddir)/src/common/libflux-internal.la \
+	$(top_builddir)/src/common/librlist/librlist.la
+plugins_alloc_check_la_LDFLAGS = \
 	$(fluxplugin_ldflags) \
 	-module
 
--- a/src/modules/job-manager/plugins/alloc-check.c
+++ b/src/modules/job-manager/plugins/alloc-check.c
@@ -1,0 +1,215 @@
+/************************************************************\
+ * Copyright 2023 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* alloc-check.c - plugin to ensure resources are never double booked
+ *
+ * A fatal exception is raised on jobs that are granted resources already
+ * granted to another.
+ *
+ * In order to be sure that the exception can be raised before a short job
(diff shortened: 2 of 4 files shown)


garlick:
I pushed a sharness test.

In testing before this, jobs always failed due to the timelimit exception, and the `alloc-check` plugin would add another exception so jobs failed no matter what.   In the new test, I simulated a double-booking using `alloc-bypass` and found the exception raised by `alloc-check` does not seem to cause the job to fail as I had expected, so that needs more looking.   It seems wrong for a job to have a severity 0 event in its eventlog and not fail so possibly  I'm forgetting some additional step that the plugin should be taking.

grondo:
>  It seems wrong for a job to have a severity 0 event in its eventlog and not fail so possibly I'm forgetting some additional step that the plugin should be taking.

I wonder if the problem is that an exception event is just added to the eventlog, but this doesn't trigger the `job-exception` event that the job-exec module is listening for in order to signal the job. (I wonder if the existing use cases  for `flux_jobtap_raise_exception()` either aren't well tested or all occur before the `start` event.

garlick:
Yeah something pretty fishy is going on here.  The job seems to transition to RUN state when it should have gone to CLEANUP:
```
0.000s: job.submit {"userid":5588,"urgency":16,"flags":0,"version":1}
0.011s: job.validate
0.022s: job.depend
0.022s: job.priority {"priority":16}
0.022s: job.alloc {"annotations":{"sched":{"resource_summary":"rank0/core[0-5],gpu0"}}}
0.022s: job.exception type=alloc-check severity=0 resources already allocated
0.024s: job.start
0.023s: exec.init
0.023s: exec.starting
0.036s: exec.shell.init {"service":"5588-shell-fMM79KV","leader-rank":0,"size":1}
0.037s: exec.shell.start {"taskmap":{"version":1,"map":[[0,1,1,1]]}}
0.038s: exec.shell.task-exit {"localid":0,"rank":0,"state":"Exited","pid":2005996,"wait_status":0,"signaled":0,"exitcode":0}
0.039s: exec.complete {"status":0}
0.039s: exec.done
0.039s: job.finish {"status":0}
```
I'll open a separate issue.

grondo:
I wonder if part of the issue is that posting the exception should prevent the start request being sent to the job execution system, but since the change to linearize events the exception is queued instead of processed immediately.

garlick:
Yeah that seems like a plausible explanation.  I did note that the job is in RUN state when the `job.event.alloc` callback is made, so the start request may have already been sent since that's the action that occurs when the RUN state is entered.  Maybe the job manager should cancel the run request rather than relying on the open loop event exclusively?

I also noticed that `flux job wait` says the above eventlog is a failed job, while `flux job attach` and `flux jobs --format {result}` say the job is OK.  So at least there's an inconsistency there that needs to be addressed.

grondo:
>  I did note that the job is in RUN state when the job.event.alloc callback is made

Ah, that's right, the event callbacks are made at the same spot as the state callbacks: _after_ the event transition.

Well, anyway, you are correct in that exceptions raised in RUN state before the `start` event are not being handled correctly in many places :-(

garlick:
For this PR, I changed the test to just wait for an `alloc-check` exception rather than to expect the job to fail. If this is merged I can submit a fluxion test that reproduces the alloc problem over there, so my vote would be not to hold it up on the exception handling issues.

grondo:
Sounds good to me. I'll give a quick review.

grondo (approved):
LGTM

PR #5310, opened 2023-07-05
job-manager: publish event on jobtap exception

### LINKED ISSUE #5306: job-manager: `flux_jobtap_raise_exception()` does not publish event message
Problem: as noted in #5304, `flux_jobtap_raise_exception()` posts the `exception` event to the job eventlog, but does not publish the `job-exception` event message.

garlick:
Problem: the job manager publishes a job-exception event message when an exception is raised via the job-manager.raise RPC, but not when a jobtap plugin calls flux_jobtap_raise_exception().

Publish the job-exception event when a jobtap plugin calls flux_jobtap_raise_exception().

Fixes #5306

### DIFF
--- a/src/modules/job-manager/raise.c
+++ b/src/modules/job-manager/raise.c
@@ -28,11 +28,13 @@
 #if HAVE_CONFIG_H
 #include "config.h"
 #endif
+#include <unistd.h>
 #include <ctype.h>
 #include <flux/core.h>
 
 #include "src/common/libjob/idf58.h"
 #include "src/common/libczmqcontainers/czmq_containers.h"
+#include "src/common/libutil/errno_safe.h"
 
 #include "job.h"
 #include "event.h"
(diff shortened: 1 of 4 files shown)


grondo:
I did notice that there is another spot where the job-manager emits an exception event without publishing an event (in `alloc.c`):
https://github.com/flux-framework/flux-core/blob/0678fe5a3f5b534b9a12c8946f853536024878a1/src/modules/job-manager/alloc.c#L289-L294

Should the event be published in `event.c` when encountering an exception event in the eventlog instead of at each event source?
(Maybe that would cause trouble on replay)

grondo (approved):
This seems reasonable. It is still an open question if job-execption events must be published before the start event, and if so if the exception in alloc.c also needs the treatment (or if the event publishing should be moved to a common place). However, we can open a separate issue those questions aren't easily answered.

garlick:
Maybe I could quickly add a common function to raise.c that all can use, and add the missing case in alloc.c?  I got slightly lost in the weeds thinking about the replay case when you mentioned doing this for all exceptions in event.c so maybe that's not a good idea to do for now.

I apparently didn't document the event messages (for exceptions or for state transitions) in [rfc 21](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_21.html) which seems unfortunate.  Maybe we ought to have a conversation later about those events and what makes sense going forward, and then document that.  My idea at the time was that published exceptions could be used like the [cifts](https://www.mcs.anl.gov/research/cifts/news/index.php) project, which appears not to have gone anywhere, but which had some good ideas about sharing fault information with an open loop pub/sub mechanism.

grondo:
That all sounds reasonable to me!

garlick:
Just pushed a rework that handles the alloc case and refactors to use a common function for exception raising.

src/modules/job-manager/raise.c
```diff
@@ -63,41 +65,61 @@ int raise_check_severity (int severity)
...
+    }
+    // add optional note key
+    if (note) {
+        json_t *val;
+        if (!(val = json_string (note))
+            || json_object_set_new (evctx, "note", val) < 0) {
+            json_decref (val);
+            goto nomem;
+        }
+    }
```
chu11, line 102: given workaround above, note is always non-null ... or just checking `if (note)` for future expected changes

garlick: Yes.  I figured I'd make it easy to just drop those two lines once the issue is addressed.

garlick:
CI caught a spelling error!  Doh!  Force pushed a fix.

garlick:
Restarted fedora38 builder after encountering this known issue
> ERROR: test_remote.t - exited with status 137 (terminated by signal 9?)

grondo (approved):
This ended up being nice cleanup! LGTM!

garlick:
Thanks! I'll set MWP.

PR #5311, opened 2023-07-06
python: support "original" option in job kvslookup module

chu11:
Support something similar to the `--original` option in `flux job info`

There are multiple approaches and I went with the approach of make the `libjob` `unwrap_string()` function public.  Hopefully not a unpopular decision.

There were several reasons:

0) `libjob`'s `unwrap_string()` function easily handles the `flux-core` built w/ and w/o `flux-security` case.

1) while possible to implement the `unwrap_string()`'s `flux-security` code in python via the security module, in `libjob` there is an implementation of the function `sign_none_unwrap()` for the "no flux security" case, which would probably be hard to re-do in python (and minimally annoying to re-do).

2) I realized `unwrap_string()` is used in the job shell, it made me think there may be users of `unwrap_string()` in the future outside of `flux-core`.

So ... I made `unwrap_string()` public and re-named it `flux_unwrap_string()`.  I declare it in a new `job_util.h` header.  I state in the comments in the header that it is mostly for internal use or associated libraries/tools.

Then adding it into the python job `kvslookup.py` module.  Wasn't too hard logically, although unsure I made the code "Pythonic" enough.

### DIFF
--- a/src/bindings/python/flux/job/kvslookup.py
+++ b/src/bindings/python/flux/job/kvslookup.py
@@ -10,21 +10,26 @@
 import errno
 import json
 
+from _flux._core import ffi, lib
 from flux.future import WaitAllFuture
 from flux.job import JobID
 from flux.rpc import RPC
+
+
+def _decode_field(data, key):
+    try:
+        tmp = json.loads(data[key])
+        data[key] = tmp
+    except json.decoder.JSONDecodeError:
+        # Ignore if can't be decoded
+        pass
 
 
 # a few keys are special, decode them into dicts if you can
 def decode_special_data(data):
     for key in ("jobspec", "R"):
         if key in data:
-            try:
-                tmp = json.loads(data[key])
-                data[key] = tmp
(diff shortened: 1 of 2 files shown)


grondo:
Haven't reviewed yet, but sounds good to me!

I kind of wonder if `export flux_unwrap_string(3)` should be its own PR for release notes? Kinda borderline for me, so if it would be a lot of work, nevermind.

chu11:
> I kind of wonder if export flux_unwrap_string(3) should be its own PR for release notes? Kinda borderline for me, so if it would be a lot of work, nevermind.

Ahhh I didn't think of doing that, but it sounds like a good idea.  I'll split it off.

chu11:
re-pushed building off of the now split off #5312, and also fixed up some python lint failures

grondo (approved):
This LGTM!

chu11:
thanks, rebased off master with the merged in #5312, setting MWP

PR #5312, opened 2023-07-06
libjob: export flux_unwrap_string(3) function

chu11:
Per comment in #5311 splitting this into its own PR.

### DIFF
--- a/src/cmd/flux-job.c
+++ b/src/cmd/flux-job.c
@@ -42,7 +42,6 @@
 #include "src/common/libutil/log.h"
 #include "src/common/libutil/jpath.h"
 #include "src/common/libjob/job.h"
-#include "src/common/libjob/unwrap.h"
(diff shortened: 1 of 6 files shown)


grondo:
I think it is ok to put it in `job.h` for now, I was just more curious why you split it out (no `flux_job` prefix makes sense).
I leave the final decision to you.

Note that if you add `job_util.h` to `src/common/core.h` it should probably also be added to `src/include/flux/core.h`. Then it doesn't have to be explicitly included in `src/cmd/flux-job.c` (that was what kind of made me think it would be simpler to just add the helper function to `job.h`)

chu11:
ok re-worked putting the declaration in `job.h`.  Commit message tweak indicating removing `unwrap.h` in the few places that no longer need it.

grondo:
Feel free to set MWP  as far as I'm concerned!

Issue #5313, opened 2023-07-06
`flux job attach` notification says job is "starting" when prolog is running


grondo:
Problem: The `flux job attach` notify support doesn't count the number of `prolog-start` events, and therefore assumes the job is "starting" after the first `prolog-finish` event is posted. This is confusing since users will think a job is hanging while starting when in reality it is waiting for prolog completion.

PR #5315, opened 2023-07-06
flux-job: fix attach notification with multiple prolog-start events

grondo:
This PR fixes the issue described in #5313. Instead of printing `starting` after the first `prolog-finish` event, a counter is kept so that the attach status prints `waiting for prolog` until the last expected `prolog-finish` event.


garlick (approved):
This LGTM.

grondo:
Thanks! I'll set MWP.

Issue #5305, opened 2023-07-04
flux job info: "J" key and --original may output wrong information


chu11:
It just occurred to me that the example error in #5303 may also output incorrectly once that is fixed (related, but slightly different issue).

```
>flux job info --original fLTC6U8f J jobspec
Jul 03 14:40:23.296787 job-info.err[0]: lookup_key: flux_future_push: File exists
flux-job: flux_rpc_get_unpack: File exists
```

This should output "J" and "jobspec", but due to the `--original` option, it'll probably output "jobspec" and "jobspec".

Issue #5303, opened 2023-07-03
job-info: error if same key requested twice


chu11:
```
>flux job info fLTC6U8f J J
Jul 03 14:38:19.482861 job-info.err[0]: lookup_key: flux_future_push: File exists
flux-job: flux_rpc_get_unpack: File exists
```

Note that with the `flux job info --original` option, this could manifest itself in a non-obvious way since "jobspec" will lead to a request of "J" internally.

```
>flux job info --original fLTC6U8f J jobspec
Jul 03 14:40:23.296787 job-info.err[0]: lookup_key: flux_future_push: File exists
flux-job: flux_rpc_get_unpack: File exists
```

grondo:
Ooh, good catch!

chu11:
hmmm, just pondering this a bit this morning.  Should we return a nicer error to user?  Or ignore duplicate keys?  My current opinion is ignore the duplicates, return the keys requested.  In user-land, can output the original input in the order requested.  i.e.

flux job info: input "J" "J" "R"

job-info module: will return "J" & "R"

flux job info: output "J" "J" "R"

PR #5317, opened 2023-07-07
 job-info: ignore duplicate lookup keys

chu11:
Per #5303.  Multiple approaches on fixing this, I elected to ignore duplicates.  The main reason is because the lookup result is a json object.  So duplicate keys sort of make no sense on a return (i.e. in a json object you can't put the same key in the object twice).

Could have done a "are there duplicates?" check instead and return EINVAL, but figured it's a costly check for a (hopefully) rare corner case.

Note, issue #5305 is highly related to this, but a little different.  Going to solve that in follow up PR.


grondo (approved):
LGTM!

PR #5318, opened 2023-07-07
flux-job: fix invalid --original info output

chu11:
Problem: In flux job info and the --original option, it is assumed that every response of "J" is meant to be "jobspec".  However, that
is not a safe assumption, as the user may have input both "J" and "jobspec".

Solution: When outputting the results from a flux job info lookup, output results based on the keys input by the user, not by the keys used during lookup.

Fixes https://github.com/flux-framework/flux-core/issues/5305

Built on top of #5317

### DIFF
--- a/src/cmd/flux-job.c
+++ b/src/cmd/flux-job.c
@@ -3335,11 +3335,11 @@ struct info_ctx {
     bool original;
 };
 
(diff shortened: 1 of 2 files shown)


chu11:
rebased now that #5317 is merged

grondo (approved):
LGTM

chu11:
thanks setting MWP

PR #5320, opened 2023-07-09
flux-job: fix `flux job status` handling of nonfatal exceptions

grondo:
I noticed that `flux job status` treats all job exceptions as fatal. This is incorrect behavior since a job can get an exception with severity > 0 and still be "successful" i.e. exit with a 0 status.

This PR fixes the behavior and adds a simple test.

### DIFF
--- a/src/cmd/flux-job.c
+++ b/src/cmd/flux-job.c
@@ -2837,6 +2837,7 @@ int cmd_status (optparse_t *p, int argc, char **argv)
         int status = -1;
         int exitcode;
         int exception = 0;
+        int exc_severity = 0;
         const char *exc_type = NULL;
 
         if (json) {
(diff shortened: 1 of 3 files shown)


chu11 (approved):
LGTM.  Just one minor nit, it took me a bit to remember "--urgency=0" means hold.  Perhaps could do "--urgency=hold" instead? Sorry if it leads to extra cleanup work, can leave for another day of course.

grondo:
Sure, I can tack on a commit to fix that up.

grondo:
Ok, I pushed a commit to update uses of `urgency=0` to `urgency=hold` in the testsuite.

garlick (approved):
LGTM!

chu11:
LGTM

grondo:
Oops, changed one use of `--urgency=0` where `hold` isn't supported. Will fix, push and then set MWP.

PR #5321, opened 2023-07-09
job-manager: fix prolog/epilog exception handling

grondo:
This PR fixes a couple issues in the handling of exceptions in the `perilog` plugin:

 - Only kill the prolog when an exception is received and the prolog is running. Ignore exceptions during the epilog.
 - Only kill the prolog for fatal job exceptions, i.e. ignore all job exceptions with severity > 0.

This will require flux-framework/flux-sched#1047 to be merged so that the flux-sched checks won't fail.


garlick (approved):
LGTM!  Thanks.

grondo:
Thanks! Setting MWP.

Issue #5308, opened 2023-07-05
python: KVSDir object should write to path it is initialized with


garlick:
Problem: a KVSDir object acts like a dictionary for a  KVS directory when _reading_ from the KVS, but it discards its relative path and operates directly on the root when _writing_ to the KVS.

For example, this program uses a KVSDir to print a key from a job directory and then update the same key:
```python
#!/usr/bin/env python3

import sys
import flux.job
import flux.job.JobID as JobID

jobid = JobID(sys.argv[1])
key = sys.argv[2]
val = sys.argv[3]

handle = flux.Flux()
jobdir = flux.job.job_kvs(handle, jobid)
# This  refers to a key name in the job's KVS directory
print(jobdir[key])

# But this refers to a key name in the root directory
jobdir[key] = val
jobdir.commit()
```

```
$ flux ./foo.py  ƒ6acjkLQf jobspec xyz
{'resources': [{'type': 'slot', 'count': 1, 'with': [{'type': 'core', 'count': 1}], 'label': 'task'}], 'tasks': [{'command': ['hostname'], 'slot': 'task', 'count': {'per_slot': 1}}], 'attributes': {'system': {'duration': 0, 'cwd': '/nfshome/garlick', 'shell': {'options': {'rlimit': {'cpu': -1, 'fsize': -1, 'data': -1, 'stack': -1, 'core': 16384, 'nofile': 128000, 'as': -1, 'rss': -1, 'nproc': 8192}}}}}, 'version': 1}

# hmm no update
$ flux job info ƒ6acjkLQf jobspec
{"resources":[{"type":"slot","count":1,"with":[{"type":"core","count":1}],"label":"task"}],"tasks":[{"command":["hostname"],"slot":"task","count":{"per_slot":1}}],"attributes":{"system":{"duration":0,"cwd":"/nfshome/garlick","shell":{"options":{"rlimit":{"cpu":-1,"fsize":-1,"data":-1,"stack":-1,"core":16384,"nofile":128000,"as":-1,"rss":-1,"nproc":8192}}}}},"version":1}

# oops unexpected jobspec key in the root!
$ flux kvs get jobspec
"xyz"
```

vsoch:
Thanks for opening this @garlick !

garlick:
Thanks for finding it!

chu11:
[...]

PR #5322, opened 2023-07-10
python: fix writes in KVSDir with initial paths

chu11:
Per #5308.  Fix and add testing coverage that was missing before.

Nothing too extravagant about this fix.  It seems it was an  oversight since the initial implementation.  But as I note in #5308, this has the potential to break any code that was using it in the "buggy" way prior.  i.e.

```
dir = KVSDir(handle, path="foo.bar")
dir["foo.bar.a"] = 1
dir.commit()
```

before this created "foo.bar.a" and if we change it'll create "foo.bar.foo.bar.a".  Same with any other functions that underneath the covers write to the KVS via KVSDir (e.g. `fill()` object function) or anything that gets a KVSDir, such as `job_kvs()` or `kvs.get_dir()`.

I think it's low probability that this will cause a lot of breakage out there, but any breakage would be quite severe for said users.  Not sure if we want to ping some code teams first?  Or alert code teams in release notes with special note?

### DIFF
--- a/t/python/t0005-kvs.py
+++ b/t/python/t0005-kvs.py
@@ -114,9 +114,10 @@ class TestKVS(unittest.TestCase):
             kd.fill({"things": 1, "stuff": "strstuff", "dir.other_thing": "dirstuff"})
             kd.commit()
 
-            self.assertEqual(kd["things"], 1)
-            self.assertEqual(kd["stuff"], "strstuff")
-            self.assertEqual(kd["dir"]["other_thing"], "dirstuff")
+        with flux.kvs.get_dir(self.f) as kd2:
+            self.assertEqual(kd2["things"], 1)
+            self.assertEqual(kd2["stuff"], "strstuff")
+            self.assertEqual(kd2["dir"]["other_thing"], "dirstuff")
 
     def test_set_deep(self):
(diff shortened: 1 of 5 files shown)


t/python/t0005-kvs.py
```diff
@@ -114,9 +114,10 @@ def test_fill(self):
             kd.fill({"things": 1, "stuff": "strstuff", "dir.other_thing": "dirstuff"})
```
garlick, line 114: Commit message for "testsuite: check kvs writes with new context ":  an extra "the"

garlick (approved):
LGTM, just spotted one typo.

This ought to be fixed regardless of whether people are relying on the broken behavior, so my vote is to merge it now, and maybe just note the API breaking behavior in the  release notes.  If we can think of anybody we ought to send a courtesy notice to, we could do that as well.

Optional but the generated docs for this are...barren.  Would it make sense to add some inilne docs (for KVSDir at least) as part of this PR to improve things somewhat?

https://flux-framework.readthedocs.io/projects/flux-core/en/latest/python/autogenerated/flux.kvs.html

chu11:
re-pushed fixing up that commit message typo.  i'll set MWP.

PR #5323, opened 2023-07-12
Python: Derive FluxExecutor from concurrent.futures.Executor

jan-janssen:
In analogy to the `FluxExecutorFuture` being derived from `concurrent.futures.Future` now also the `FluxExecutor` is derived from `concurrent.futures.Executor` . This has two advantages, on the one hand the `__enter__()` and `__exit__()` functions are already defined and on the other hand frameworks which integrate interfaces for multiple Executors can check for classes derived from the `concurrent.futures.Executor` class.

### DIFF
--- a/src/bindings/python/flux/job/executor.py
+++ b/src/bindings/python/flux/job/executor.py
@@ -414,7 +414,7 @@ class FluxExecutorFuture(concurrent.futures.Future):
                 self._invoke_flux_callback(callback, log_entry)
 
 
-class FluxExecutor:
+class FluxExecutor(concurrent.futures.Executor):
     """Provides a method to submit and monitor Flux jobs asynchronously.
 
     Forks threads to complete futures and fetch event updates in the background.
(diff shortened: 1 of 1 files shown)


jameshcorbett:
When I wrote it I made a choice not to derive `FluxExecutor` from `concurrent.futures.Executor`, because the interface isn't the same. You can't take a `FluxExecutor` and treat it like an `Executor`. As the docstring of the class says:

```
    Inspired by the ``concurrent.futures.Executor`` class, with the following
    interface differences:

        - the ``submit`` method takes a ``flux.job.Jobspec`` instead of a
          callable and its arguments, and returns a ``FluxExecutorFuture``
          representing that job.
        - the ``map`` method is not supported, given that the executor consumes
          Jobspecs rather than callables.
```

I think these differences are significant and make it so that `isinstance(FluxExecutor(), Executor)` checks should not pass.

jan-janssen:
> I think these differences are significant and make it so that `isinstance(FluxExecutor(), Executor)` checks should not pass. Can you explain why you think they should pass?

Makes sense - thanks for the explanation.

Issue #5327, opened 2023-07-13
python: update / modernize KVS api


chu11:
IIRC, the current implementation was an initial "good for now" when KVS changes were put into transaction objects.  It was never revisited for improvement.  e.g. in code we have 

```
def put_symlink(flux_handle, key, target):
    if flux_handle.aux_txn is None:
        flux_handle.aux_txn = RAW.flux_kvs_txn_create()
    return RAW.flux_kvs_txn_symlink(flux_handle.aux_txn, 0, key, None, target)


def commit(flux_handle, flags: int = 0) -> int:
    if flux_handle.aux_txn is None:
        return -1
    future = RAW.flux_kvs_commit(flux_handle, None, flags, flux_handle.aux_txn)
    RAW.flux_future_get(future, None)
    RAW.flux_kvs_txn_destroy(flux_handle.aux_txn)
    flux_handle.aux_txn = None
    return 0
```

of course, no one appears to have complained at the moment.  so probably not a high priority

chu11:
I'm updating this issue to be a little more all-encompassing.  The KVS Python API was largely developed from a pretty ancient `libkvs` implementation.  From before KVS transactions existed.  Before `libflux` futures existed.  Before `treeobj` existed.  When everything written to the KVS had to be in `json` format.  And probably other things I'm not remembering.

- Python KVS API only allows one transaction per flux handle at a time
- Workarounds for legacy json requirement in data written to KVS (https://github.com/flux-framework/flux-core/issues/5198)
- Many functions return 0 / -1 instead of (probably) raising an exception

So many it will need an update / modernizing at some point

chu11:
Was pondering this a bit:

- kvs namespace support - could be added via a new optional/named parameter to most functions
- new "KVSTxn" class could be used to handle multiple transactions per flux handle at a time
  - could be optional/named parameter to `mkdir()`, `put()`, etc. functions, although the class would have its own mkdir(), put(), etc. that most would probably end up using anyways.  
  - could be optional/named parameter to `commit()` function, again, the class would probably have its own commit() function anyways.
  
together, this could probably modernize the API, but at the same time not obliterate the old API.

grondo:
Probably even more useful than exposing the transaction interface would be exposing even the simplest interface to KVS watch if that isn't already available.

chu11:
> Probably even more useful than exposing the transaction interface would be exposing even the simplest interface to KVS watch if that isn't already available.

[...]

PR #5328, opened 2023-07-13
python: add extra documentation to kvs.py module

chu11:
Added a bit more in there.  Didn't add it for everything, sorta picked and choosed, notably not adding docs for the direct access to C-bindings stuffs b/c that stuff is mostly just used by `KVSDir`.  The readthedocs shouldn't be quite so bare now.

### DIFF
--- a/src/bindings/python/flux/kvs.py
+++ b/src/bindings/python/flux/kvs.py
@@ -46,6 +46,15 @@ def get_key_direct(flux_handle, key):
 
 
 def exists(flux_handle, key):
+    """Determine if key exists
+
+    Args:
+        flux_handle: A Flux handle obtained from flux.Flux()
+        key: key to check for existence
+
+    Returns:
+        bool: True if key exists, False if not
+    """
     try:
         get_key_direct(flux_handle, key)
         return True
@@ -58,6 +67,15 @@ def exists(flux_handle, key):
 
 
 def isdir(flux_handle, key):
+    """Determine if key is a directory
+
+    Args:
+        flux_handle: A Flux handle obtained from flux.Flux()
+        key: key to check if it is a directory
+
+    Returns:
+        bool: True if key is a directory, False if not
+    """
     try:
         get_key_direct(flux_handle, key)
     except EnvironmentError as err:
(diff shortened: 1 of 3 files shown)


src/bindings/python/flux/kvs.py
```diff
@@ -175,13 +189,16 @@ def __init__(self, flux_handle=None, path=".", handle=None):
...
+        """Commit changes to the KVS, must be called after changes are done"""
         return commit(self.fhdl, flags)
 
     def key_at(self, key):
+        """Get full path to KVS key"""
         p_str = self.pimpl.key_at(key)
         return p_str.decode("utf-8")
 
     def exists(self, name):
+        """Evaluate if key exists in the KVS"""
```
garlick, line 201: Should this say "in the KVS directory"?

garlick (commented):
Despite the C heavy nature of the KVS API in flux-core, I think we're likely to have more python users over time, so focusing documentation efforts on the python API might be a good thing?

So perhaps we could get a bit more detail here?

chu11:
> Despite the C heavy nature of the KVS API in flux-core, I think we're likely to have more python users over time, so focusing documentation efforts on the python API might be a good thing?

That's a fair point.  I choose to skip also because of https://github.com/flux-framework/flux-core/issues/5327, i.e. I'm not entirely sure what this should eventually look like.

But can definitely add more, it'll make documentation easier in the future down the road too.

src/bindings/python/flux/kvs.py
chu11: gahhhhhh, I think I am right at the moment, BUT .... this is another "initial path" bug ... it should be "KVS directory"  .... I will push a separate fix.

```
diff --git a/src/bindings/python/flux/kvs.py b/src/bindings/python/flux/kvs.py
index ded222526..38a91814a 100644
--- a/src/bindings/python/flux/kvs.py
+++ b/src/bindings/python/flux/kvs.py
@@ -200,7 +200,7 @@ class KVSDir(WrapperPimpl, abc.MutableMapping):
 
     def exists(self, name):
         """Evaluate if key exists in the KVS"""
-        return exists(self.fhdl, name)
+        return exists(self.fhdl, self._path + name)
 
     def __getitem__(self, key):
         try:
diff --git a/t/python/t0005-kvs.py b/t/python/t0005-kvs.py
index 20efd9a62..344931aeb 100755
--- a/t/python/t0005-kvs.py
+++ b/t/python/t0005-kvs.py
@@ -90,8 +90,15 @@ class TestKVS(unittest.TestCase):
 
     def test_exists_true(self):
         flux.kvs.put(self.f, "thing", 15)
+        flux.kvs.put_mkdir(self.f, "existsubdir")
+        flux.kvs.put(self.f, "existsubdir.thing2", 14)
         flux.kvs.commit(self.f)
         self.assertTrue(flux.kvs.exists(self.f, "thing"))
+        self.assertTrue(flux.kvs.exists(self.f, "existsubdir.thing2"))
+
+        with flux.kvs.get_dir(self.f, "existsubdir") as kd:
+            self.assertTrue(flux.kvs.exists(self.f, "thing2"))
```

chu11:
re-pushed

- added way more documentation per comments above
- along the way realized some really old comments in `libkvs/kvs.h` were out of date, fixed that up
- along the way realized there were no coverage for commit flags, added some extra coverage
- realized there's an issue (#5333) added a temporary workaround for it in the tests

maybe could split out those testsuite commits into a new PR, but they are pretty tiny

grondo (approved):
LGTM!

PR #5329, opened 2023-07-13
python: do not return int status in kvs functions     

chu11:
The first 4 commits here at coverage / cleanups I noticed while working on the prior python kvs module doc PR.

The final two commits were things I noticed and I believe are wrong.  Probably very early "how should we integrate this all in together" mistakes from 2018 in f4c9e2d6d9bf1da1af0f4a333f81bd436cd7f082

Whether we should do them is a different question, i.e. breaking behavior.  I think very low odds this change matters and would break users.  Also, since we did break some behavior via https://github.com/flux-framework/flux-core/pull/5322, I figure what the heck.

### DIFF
--- a/src/bindings/python/flux/kvs.py
+++ b/src/bindings/python/flux/kvs.py
@@ -87,7 +87,7 @@ def put(flux_handle, key, value):
         flux_handle.aux_txn = RAW.flux_kvs_txn_create()
     try:
         json_str = json.dumps(value)
-        return RAW.flux_kvs_txn_put(flux_handle.aux_txn, 0, key, json_str)
+        RAW.flux_kvs_txn_put(flux_handle.aux_txn, 0, key, json_str)
     except TypeError:
         if isinstance(value, bytes):
             return RAW.flux_kvs_txn_put_raw(
@@ -99,33 +99,32 @@ def put(flux_handle, key, value):
 def put_mkdir(flux_handle, key):
     if flux_handle.aux_txn is None:
         flux_handle.aux_txn = RAW.flux_kvs_txn_create()
-    return RAW.flux_kvs_txn_mkdir(flux_handle.aux_txn, 0, key)
+    RAW.flux_kvs_txn_mkdir(flux_handle.aux_txn, 0, key)
 
 
 def put_unlink(flux_handle, key):
     if flux_handle.aux_txn is None:
         flux_handle.aux_txn = RAW.flux_kvs_txn_create()
-    return RAW.flux_kvs_txn_unlink(flux_handle.aux_txn, 0, key)
+    RAW.flux_kvs_txn_unlink(flux_handle.aux_txn, 0, key)
 
 
 def put_symlink(flux_handle, key, target):
     if flux_handle.aux_txn is None:
         flux_handle.aux_txn = RAW.flux_kvs_txn_create()
-    return RAW.flux_kvs_txn_symlink(flux_handle.aux_txn, 0, key, None, target)
+    RAW.flux_kvs_txn_symlink(flux_handle.aux_txn, 0, key, None, target)
 
 
-def commit(flux_handle, flags: int = 0) -> int:
+def commit(flux_handle, flags: int = 0):
     if flux_handle.aux_txn is None:
-        return -1
+        return
     future = RAW.flux_kvs_commit(flux_handle, None, flags, flux_handle.aux_txn)
     RAW.flux_future_get(future, None)
     RAW.flux_kvs_txn_destroy(flux_handle.aux_txn)
     flux_handle.aux_txn = None
-    return 0
 
 
 def dropcache(flux_handle):
-    return RAW.flux_kvs_dropcache(flux_handle)
+    RAW.flux_kvs_dropcache(flux_handle)
 
 
 class KVSDir(WrapperPimpl, abc.MutableMapping):
@@ -177,8 +176,8 @@ class KVSDir(WrapperPimpl, abc.MutableMapping):
             )
         self.pimpl = self.InnerWrapper(flux_handle, path, handle)
 
-    def commit(self, flags=0) -> int:
-        return commit(self.fhdl, flags)
+    def commit(self, flags=0):
+        commit(self.fhdl, flags)
 
     def key_at(self, key):
         p_str = self.pimpl.key_at(key)
@@ -196,8 +195,7 @@ class KVSDir(WrapperPimpl, abc.MutableMapping):
             )
 
     def __setitem__(self, key, value):
-        if put(self.fhdl, self._path + key, value) < 0:
-            print("Error setting item in KVS")
+        put(self.fhdl, self._path + key, value)
 
     def __delitem__(self, key):
         put_unlink(self.fhdl, key)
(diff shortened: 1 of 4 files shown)


src/bindings/python/flux/kvs.py
```diff
@@ -118,10 +118,10 @@ def commit(flux_handle, flags: int = 0) -> int:
     if flux_handle.aux_txn is None:
         return 0
     future = RAW.flux_kvs_commit(flux_handle, None, flags, flux_handle.aux_txn)
-    RAW.flux_future_get(future, None)
+    rv = RAW.flux_future_get(future, None)
     RAW.flux_kvs_txn_destroy(flux_handle.aux_txn)
     flux_handle.aux_txn = None
-    return 0
+    return rv
```
garlick, line 125: Should this raise an exception of some sort instead?
(Sorry if that is a dumb question!)

garlick (approved):
LGTM - good to get this cleaned up!

src/bindings/python/flux/kvs.py
chu11: Good question.  I kept it in the same style for now b/c other functions that call the KVS C-lib also return -1 on error.

The Python API that currently exists was largely developed based off of the KVS api before KVS transactions existed, before libflux futures existed, and probably other things I'm not remembering.  There's no current need to re-do it, but perhaps it will in time.  Perhaps I'll update #5327 to be a general "modernize Python KVS functions".

src/bindings/python/flux/kvs.py
```diff
@@ -116,7 +116,7 @@ def put_symlink(flux_handle, key, target):
 
 def commit(flux_handle, flags: int = 0) -> int:
     if flux_handle.aux_txn is None:
```
grondo, line 118: I struggled a bit with this since calling `commit()` with no transaction felt more like calling `flux_kvs_commit()` with a `NULL` transaction, which is an error. But because of the current practice of one-transaction-per-handle I think your approach here is correct.

However, it isn't very idiomatic Python to return 0 from a function to test for success. Should the function just return nothing and raise an exception on error? (Apologies if you fix this in a later commit)

src/bindings/python/flux/kvs.py
grondo: I was just writing the same question. I think we do need to modernize these functions and use idiomatic Python. My guess is that nobody acutally uses this interface.

grondo: I don't think any Python developer is going to check the return code of a function for success, so I'd actually say this is borderline a bug. Functions like this should return nothing on success (unless there's a return value) and raise an exception on failure. 

In this case if the `future` here was promoted to a `Future`, then you'd get the exception for free...

chu11: ya know what, i'm dumb.  I just saw the code that was there and assumed it actually returned -1 on error, but it doesn't.

```Python
handle = flux.Flux()
flux.kvs.put(handle, ".", "bar")
flux.kvs.commit(handle)
```

```
>flux python foo.py
  File "foo.py", line 50, in <module>
    flux.kvs.commit(handle)
  File "/g/g0/achu/chaos/git/flux-framework/flux-core/src/bindings/python/flux/kvs.py", line 121, in commit
    rv = RAW.flux_future_get(future, None)
  File "/g/g0/achu/chaos/git/flux-framework/flux-core/src/bindings/python/flux/wrapper.py", line 220, in __call__
    os.strerror(errnum),
OSError: [Errno 22] EINVAL: Invalid argument
```

```Python
handle = flux.Flux()
flux.kvs.put_mkdir(handle, "")
flux.kvs.commit(handle)
```

```
  File "foo.py", line 49, in <module>
    flux.kvs.put_mkdir(handle, "")
  File "/g/g0/achu/chaos/git/flux-framework/flux-core/src/bindings/python/flux/kvs.py", line 102, in put_mkdir
    return RAW.flux_kvs_txn_mkdir(flux_handle.aux_txn, 0, key)
  File "/g/g0/achu/chaos/git/flux-framework/flux-core/src/bindings/python/flux/wrapper.py", line 220, in __call__
    os.strerror(errnum),
OSError: [Errno 22] EINVAL: Invalid argument
```

Soo I think it's already returning an exception on error.  So perhaps we should remove those "returns" and send an exception.

grondo: Oh, yeah, I forgot all the wrapped ffi function calls will raise an exception on `< 0` return from the C function...

chu11:
just re-pushed with a different approach

- all those kvs functions no longer return an int status, they return nothing
- added a bunch of "einval" tests in the tests
- new tests lead to discovery of corner case in KVSDir, so fixed that too

t/python/t0005-kvs.py
```diff
@@ -334,6 +335,50 @@ def test_36_put_symlink(self):
...
         self.assertFalse(flux.kvs.exists(self.f, "txn_symlink"))
 
+    def bad_input(self, func, *args):
+        got_einval = False
+        try:
+            func(*args)
+        except EnvironmentError as e:
+            if e.errno == errno.EINVAL:
+                got_einval = True
+        self.assertEqual(got_einval, True)
```
grondo, line 345: I think the `unittest` way to do this is:
```python
with self.assertRaises(OSError) as ctx:
    func(*args)
self.assertEqual(ctx.exception.errno, errno.EINVAL)
```

See https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises

Also, since 3.3 I think `OSError` is preferred over `EnvironmentError`.

t/python/t0005-kvs.py
```diff
@@ -334,6 +335,50 @@ def test_36_put_symlink(self):
...
+
+    # try to overwrite root dir, will fail on commit
+    def test_45_commit_fail(self):
+        got_einval = False
+        try:
+            flux.kvs.put(self.f, ".", "foof")
+            flux.kvs.commit(self.f)
+        except EnvironmentError as e:
+            if e.errno == errno.EINVAL:
+                got_einval = True
```
grondo, line 379: Same as above, use `self.assertRaises()` with a context to check expected contents of the exception.

grondo (approved):
LGTM, just some minor style things (and unittest corrections) noted inline.

t/python/t0005-kvs.py
```diff
@@ -27,7 +27,7 @@ class TestKVS(unittest.TestCase):
     def setUpClass(self):
         self.f = flux.Flux()
 
-    def test_kvs_dir_open(self):
+    def test_01_kvs_dir_open(self):
```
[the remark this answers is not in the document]
chu11: Yeah, I realized this when looking through the tests.  Due to the pretty high number of tests in this file, I felt ordering them would be a good idea.  I was getting confused when looking at the output and seeing errors jumbled around.

t/python/t0005-kvs.py
chu11: ahhh this pattern is used elsewhere, will try and clean up everywhere

chu11:
re-pushed with fixes for comments above.  Also added a commit that cleaned up a few other cases to use "assertRaises()".

PR #5331, opened 2023-07-13
python: fix exists in KVSDir with initial paths

chu11:
Problem: KVSDir can be passed an initial path when initialized.  This
path is utilized as a contextual path for reads and writes.  However,
it was not used for the exists() class function.

Solution: Use the initial path when checking a key in exists().

---

Notes: this was simply missed in #5322 b/c it wasn't a "write" thingie.  Looked through all the KVSDir functions and I think everything else works as intended.

### DIFF
--- a/src/bindings/python/flux/kvs.py
+++ b/src/bindings/python/flux/kvs.py
@@ -182,7 +182,7 @@ class KVSDir(WrapperPimpl, abc.MutableMapping):
         return p_str.decode("utf-8")
 
     def exists(self, name):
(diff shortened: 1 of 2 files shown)


garlick (approved):
Good catch!

chu11:
while order of this PR #5329 and #5328 doesn't strictly matter, it might be nice to get this and #5329 in first, so the documentation one comes last and is updated with all the changes.

chu11:
@garlick thanks!  setting MWP

PR #5332, opened 2023-07-13
python: add API for job output

grondo:
This PR adds a new set of interfaces to the Python API for watching and fetching job output.
The `FutureExt` class is used here, so this is based on #5330.

The new functions are all documented in the code, but summarized here for convenience. All functions can be called on active or inactive jobs. If the job is inactive or starting, the functions will wait until the output eventlog is available to start watching it. (except for `job_output(nowait=True)` in which case a `FileNotFoundError` exception is raised (since output eventlog does not exist).

 * `job_output(flux_handle, jobid, tasks="*", labelio=False, nowait=False, log_stderr_level=LOG_TRACE)`: simple function for fetching job output synchronously in one call. Returns a `JobOutput` tuple with members `stdout`, `stderr`, and `log`, in which each separate stream is returned in full. By default all log messages are sent to `stderr`, but this behavior can be modified with the `log_stderr_level` parameter. By default this function waits until the output eventlog is complete, but with `nowait=True` only the currently available output is returned.
 * `output_event_watch_async(flux_handle, jobid, labelio=False)`: Watch output eventlog events for `jobid`. Returns a `JobOutputEventWatch` future. Callers call `get_event()` on this future during the callback to get the next available event. All events are subclasses of `EventLogEvent`, and include `OutputHeaderEvent`, `OutputEvent`, `RedirectEvent`, `LogEvent`, or `JobExceptionEvent` (though exceptions are not technically output events, it is useful to pass exceptions along to the user because in most cases they should result in a message to `stderr`. 
 * `output_watch_async(flux_handle, jobid, labelio=False, log_stderr_level=LOG_TRACE)`: This is the line oriented, asynchronous output watch interface. The returned `JobOutputWatch` future will be fulfilled when there is line available to read with the `.getline()` method, which returns a `(stream, line)` tuple where stream can be `stdout`, `stderr`, or `log`. (again, log messages are sent to `stderr`, unless `log_stderr_level` is changed)
 * `output_event_watch()` and `output_watch()` generators based on the async versions above.

### DIFF
--- a/src/bindings/python/flux/job/output.py
+++ b/src/bindings/python/flux/job/output.py
@@ -1,0 +1,982 @@
+###############################################################
+# Copyright 2023 Lawrence Livermore National Security, LLC
+# (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+#
+# This file is part of the Flux resource manager framework.
+# For details, see https://github.com/flux-framework.
+#
+# SPDX-License-Identifier: LGPL-3.0
+###############################################################
+
+import base64
+import errno
+from typing import NamedTuple
+
+from flux.future import FutureExt
+from flux.idset import IDset
+from flux.job import (
+    EventLogEvent,
+    JobException,
+    JobID,
+    event_wait,
+    event_watch,
+    event_watch_async,
+    job_kvs_lookup,
+)
+
+# Log levels from shell.h for use with log_stderr_level:
+LOG_QUIET = -1
+LOG_FATAL = 0
+# Note: LOG_ALERT=1, LOG_CRIT=2 are currently reserved in shell.h
+LOG_ERROR = 3
+LOG_WARN = 4
+LOG_NOTICE = 5
+LOG_DEBUG = 6
+LOG_TRACE = 7
+
+
+class Taskset:
+    """
+    A Taskset represents a specific set of task ranks or "all" tasks.
+    Implements a minimal interface of the IDset class, mainly for
+    testing if one Taskset intersects with another.
+
+    The underlying IDset is only initialized for Taskset objects that
+    do represent all tasks.
+
+    Attributes:
+        all (bool): a Taskset representing all tasks
+        ids (IDset): If self.all is False, the underlying idset
+    """
+
+    def __init__(self, tasks):
+        self.all = False
+        if isinstance(tasks, Taskset):
+            # Copy an existing Taskset
+            self.all = tasks.all
+            self.ids = tasks.ids.copy()
+            return
+        if isinstance(tasks, IDset):
+            self.ids = tasks.copy()
+            return
+        self.ids = None
+        if tasks in ("all", "*"):
+            self.all = True
+            return
+        self.ids = IDset(tasks)
+
+    def intersect(self, tasks):
+        if not isinstance(tasks, Taskset):
+            tasks = Taskset(tasks)
+        if self.all:
+            return Taskset(tasks)
+        elif tasks.all:
+            return Taskset(self)
+        return Taskset(self.ids.intersect(tasks.ids))
+
+    def __bool__(self):
+        return self.all or len(self.ids) > 0
+
+    def __str__(self):
+        if self.all:
+            return "all"
+        return str(self.ids)
+
+
+class OutputEvent(EventLogEvent):
+    """
+    Object representing RFC 24 Job Standard I/O data events
+    Attributes:
+        timestamp (float): timestamp for this event
+        name (str): Name of this event: 'data'
+        rank (Taskset): Set of ranks to which this event applies
+        stream (str): name of output stream ("stdout", "stderr")
+        eof (bool): True if this event marks EOF for stream
+        data (str): output data
+        dict (dict): original event as dict
+    """
+
+    def __init__(self, entry, labelio=False):
+        super().__init__(entry)
+        if self.name != "data":
+            raise ValueError(f"event {self.name} is not a data event")
+
+        self.rank = Taskset(self.context["rank"])
+        self.stream = self.context["stream"]
+        self.data = None
+        self.eof = False
+
+        if "eof" in self.context:
+            self.eof = self.context["eof"]
+
+        if "data" in self.context:
+            data = self.context["data"]
+            if "encoding" in self.context:
+                if self.context["encoding"] == "base64":
+                    data = base64.b64decode(data)
+            if "repeat" in self.context:
+                data *= self.context["repeat"]
+            if labelio:
+                data = [f"{self.rank}: {x}" for x in data.splitlines()]
+                data = "\n".join(data) + "\n"
+            self.data = data
+
+    def render(self):
+        return self.data
+
+
+class LogEvent(EventLogEvent):
+    """
+    Object representing a RFC 24 Job "log" event
+    Attributes:
+        timestamp (float): timestamp for this log event
+        name (str): name of this event: 'log'
+        rank (Taskset): shell rank that produced the log message
+        level (int): log level
+        levelstr (str): log level string
+        message (str): log message
+        component (str): log component if available
+        file (str): source file if available
+        line (int); source line if available
+        dict (dict): original event as dict
+    """
+
+    log_level_string = [
+        "FATAL",
+        "FATAL",
+        "FATAL",
+        "ERROR",
+        " WARN",
+        "",
+        "DEBUG",
+        "TRACE",
+    ]
+
+    @property
+    def levelstr(self):
+        return self.log_level_string[self.level]
+
+    def __init__(self, entry):
+        super().__init__(entry)
+        # LogEvent class is only for 'log' events
+        if self.name != "log":
+            raise ValueError(f"event {self.name} is not a log event")
+        self.level = self.context["level"]
+        self.rank = Taskset(self.context["rank"])
+        self.message = self.context["message"]
+        self.component = self.context.get("component", None)
+        self.file = self.context.get("file", None)
+        self.line = self.context.get("line", -1)
+
+    def render(self, include_file_and_line=False):
+        prefix = f"flux-shell[{self.rank}]: "
+        if self.levelstr:
+            prefix += f"{self.levelstr}: "
+        if self.component:
+            prefix += f"{self.component}: "
+        if include_file_and_line and self.file:
+            prefix += f"{self.file}:{self.line}: "
+        return prefix + self.message
+
+
+class OutputHeaderEvent(EventLogEvent):
+    """
+    Object representing an RFC 24 header event
+    Attributes:
+        timestamp (float): timestamp for this event
+        name (str): Name of this event: 'header'
+        version (int): RFC 24 version
+        stdout_count (int): expected number of stdout streams
+        stderr_count (int): expected number of stderr streams
+        stdout_encoding (str): default encoding for stdout
+        stderr_encoding (str): default encoding for stderr
+        event (dict): original event as dict
+    """
+
+    def __init__(self, entry):
+        super().__init__(entry)
+        if self.name != "header":
+            raise ValueError(f"event {self.name} is not a header event")
+        self.version = self.context["version"]
+        self.stderr_count = self.context["count"]["stderr"]
+        self.stdout_count = self.context["count"]["stdout"]
+        self.stdout_encoding = self.context["encoding"]["stdout"]
+        self.stderr_encoding = self.context["encoding"]["stderr"]
+
+
+class RedirectEvent(EventLogEvent):
+    """
+    Object representing an RFC 24 redirect event
+    Attributes:
+        timestamp (float): timestamp for this event
+        name (str): Name of this event: 'redirect'
+        stream (str): the name of the stream being redirected
+        rank (Taskset): the set of ranks being redirected
+        path (str): redirect path
+        dict (dict): original event as dict
+    """
+
+    def __init__(self, entry):
+        super().__init__(entry)
+        if self.name != "redirect":
+            raise ValueError("event {self.name} is not a redirect event")
+        self.stream = self.context["stream"]
+        self.rank = Taskset(self.context["rank"])
+        self.path = self.context["path"]
+
+    def render(self):
+        return f"{self.rank}: {self.stream} redirected to {self.path}"
+
+
+class JobExceptionEvent(EventLogEvent):
+    """
+    Object representing a JobException event
+    Attributes:
+        timestamp (float): timestamp for this event
+        name (str): name of this event: 'exception'
+        severity (int): exception severity
+        exc_type (str): exception type
+        note (str): exception note
+    """
+
+    def __init__(self, entry):
+        super().__init__(entry)
+        if self.name != "exception":
+            raise ValueError("event {self.name} is not an exception event")
+        self.severity = self.context["severity"]
+        self.exc_type = self.context["type"]
+        self.note = self.context["note"]
+
+    def render(self):
+        result = f"job.exception: type={self.exc_type} severity={self.severity}"
+        if self.note:
+            result += f" {self.note}"
+        return result
+
+
+class JobOutput(NamedTuple):
+    """
+    Tuple containing job output result
+    Attributes:
+        stdout (str): stdout from all tasks
+        stderr (str): stderr from all tasks
+        log (str): log messages
+    """
+
+    stdout: str
+    stderr: str
+    log: str
+
+
+def _parse_output_eventlog_entry(entry, labelio=False):
+    """
+    Parse a single output eventlog entry, returning an object of the
+    appropriate type: OutputEvent, LogEvent, OutputHeaderEvent,
+    RedirectEvent or JobExceptionEvent.
+    """
+    if entry is None:
+        return None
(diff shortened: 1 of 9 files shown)


grondo:
It would be good to have @ofaaland review the interfaces here to see if at least one of the proposed APIs  will work for his use case in #4854. Thanks!

grondo:
Found some missing testing and need to address CodeQL issues so hold off reviewing this one for now.

grondo:
Ok, should have fixed most of the codeql warnings and one stray `@classmethod` that prevented the `RedirectEvent` class from being initialized. Close to ready for further review now.

ofaaland:
> #4854

Hi @grondo, these seem great.  I have a few questions.  For context, in my use case, my code will be launching jobs I have no control over, and so they may start producing output immediately after my call to flux.job.submit(), may fail immediately, etc.  The world is their oyster :).

Some of these questions are likely outside the scope of this API, but I mention them here in case the output API interacts in ways I don't understand with the other interfaces I'm using.  I'm happy to read about it if you can point me to the right docs, or talk about it on slack or elsewhere.

1.  What's "log message" mean here?
2.  After calling flux.job.submit(), followed by calling job_output(), would any output emitted by the job before my call to job_output() be queued and then included in the JobOutput tuple?  Same question for the  output_event_watch() and output_watch() generators?  Does it matter if the job exits after flux.job.submit() returns but before I call job_output() or siblings?
3. When the job ends and there's no more output, would flux.job.wait_async() still allow me to determine the exit status of the job?  Or would a different interface be appropriate since I know the job completed at that point?  Does it matter if the job exits after flux.job.submit() returns but before I call wait_async() or other appropriate interface?
5. If I'm using any of these interfaces to get stderr and stdout streams, will that interfere in any way with someone attempting to debug the job itself (not my code) using tools like strace, TotalView, or STAT?

Thanks!

grondo:
> What's "log message" mean here?

Ooh, sorry that was unclear. Log messages can be emitted by the "job shell", e.g. informational or debug messages when a job is run with `-o verbose` or errors, warnings, etc. The default behavior of `flux job attach` (which is spawned by `flux run` for example), is to copy the log messages to `stderr`, so that is what is done by default in `job_output()` etc. There is an option to instead append those messages to a separate `log` stream in the result in case the caller only wants job output (or only log messages)

> After calling flux.job.submit(), followed by calling job_output(), would any output emitted by the job before my call to job_output() be queued and then included in the JobOutput tuple? Same question for the output_event_watch() and output_watch() generators? Does it matter if the job exits after flux.job.submit() returns but before I call job_output() or siblings?

Even if a job exits, a call to `job_output()`, `output_event_watch()` or `output_watch()` will return all output for the job (that was captured in the KVS anyway, there's no way to get the output if the user used `--output=filename`, though a redirect event will notify the caller of that fact.)

In fact, since output is captured in an eventlog in the KVS, you can call these functions multiple times for an inactive job and expect to get the same result.

It also doesn't matter if the job hasn't started yet. All these interfaces will block (or no callbacks called in the async cases) until output is available except for `job_output(nowait=True)`.

> When the job ends and there's no more output, would flux.job.wait_async() still allow me to determine the exit status of the job? Or would a different interface be appropriate since I know the job completed at that point? Does it matter if the job exits after flux.job.submit() returns but before I call wait_async() or other appropriate interface?

Yes, `wait_async()` uses a separate interface from the job eventlogs. It allows you to get the "success" (True/False) and error string for a job anytime after the job exits (or goes inactive, e.g. it could be canceled). It doesn't matter if the job exits before you call `wait`, the jobid just has to be valid (or if called without a `jobid` it will wait for any job). 

Note that `wait` is like POISX `wait(2)` in that it only works once per job (you are reaping the job in a way). Also, the jobs need to be submitted with the `waitable` flag which requires instance owner privilege. If you need the actual return code or to use the call multiple times per job, there is another (more heavyweight) function `flux.job.result()` (or `result_async()`).

>  If I'm using any of these interfaces to get stderr and stdout streams, will that interfere in any way with someone attempting to debug the job itself (not my code) using tools like strace, TotalView, or STAT?

No. The way KVS output works is that the Flux job shell captures output from all tasks and appends it to an "output eventlog" for the job in the KVS. Any number of entities can "watch" this eventlog for entries and decode them to get the actual job output. So you can easily have multiple watchers that do not interrupt normal job operation. The only caveat is that only the job and instance owner have access to these eventlogs, so in a multi-user instance you can't watch the output of other users' jobs.

Hope this helps. Thanks for taking a look and the intelligent questions!

ofaaland:
> Hope this helps. Thanks for taking a look and the intelligent questions!

Yes, very helpful.  The API looks great. Thanks for the work, I'm looking forward to using it!

t/python/t0015-job-output.py
```diff
@@ -0,0 +1,375 @@
...
+
+        # Now, set log_stderr_level=-1.
+        # output.log should contain all log messages
+        output = job_output(self.fh, jobid, log_stderr_level=-1)
+        self.assertEqual(output.stdout, self.test_stdout)
+        self.assertEqual(output.stderr, self.test_stderr)
+        self.assertTrue(len(output.log) > 0)
+
+    def test_job_output_nowait(self):
+        jobid = self.submit(cmd="echo first; sleep 30; echo second")
```
chu11, line 120: given my many "parallel build" test fixups ... 100 or 300 seconds?

chu11 (commented):
This is awesome!  Just some small review comments as I wrap my head a bit more around everything.

src/bindings/python/flux/job/output.py
```diff
@@ -0,0 +1,782 @@
...
+        _output_eventlog_entry_decode(
+            line, stream_dict, tasks, labelio, log_stderr_level
+        )
+
+    # Join lines and return result
+    results = ["".join(stream_dict.get(k)) for k in ("stdout", "stderr", "log")]
+    return JobOutput(*results)
+
+
+def wait_for_output(flux_handle, jobid):
```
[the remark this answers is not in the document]
grondo: Hm, I'm not sure there's a use case for this function outside of the one place it is used (earlier iterations used it elsewhere), so maybe I'll just remove it and move the code into the function.

> I think the wait is what screwed me up, since that typically suggests "wait for job to finish".

We also have `future_wait_for()` which waits for a future to be ready. That is more along the lines of what I was thinking here -- waiting for output to be "ready" to read. (i.e. similar to the kvs WAITCREATE flag). So I don't think we can say that `wait` shall only mean "wait for job to finish" in Flux.

src/bindings/python/flux/job/output.py
```diff
@@ -0,0 +1,782 @@
...
+        if event.rank.intersect(tasks):
+            stream = "stderr"
+    elif event.name == "exception":
+        stream = "stderr"
+    else:
+        return
+
+    data = event.render()
+    if data:
+        stream_dict.setdefault(stream, []).append(data)
```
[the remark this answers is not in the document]
grondo: I'll change it to `text` if `data` doesn't work.

grondo:
Thanks @chu11 and @ofaaland! I've forced a push that addresses @chu11's early comments. (Just to keep on top of things)

grondo:
I also went ahead and added underscores to all the internal function names in the `JobOutputWatch` and `JobOutputEventWatch` classes.

src/bindings/python/flux/job/output.py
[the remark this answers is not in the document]
chu11: Fair enough, dunno why my brain thought `wait_for_output()` would wait for the job to finish first.  Perhaps it's just me.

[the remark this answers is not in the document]
grondo: Yeah, I can understand. Removed the confusing function since I'm not sure there's not much of a real use case.

grondo:
Ok, I've pushed a slight rework of this PR after attempting to use these interfaces in a "bulk job watcher" implementation.
The main changes include:

 -  `JobOutputWatch` and `output_watch()` now get the raw chunks of output data from the eventlog, and a new class `JobOutputWatchLines` and `output_watch_lines()` function perform the "watch by line" functionality. (As an aside, I'm wondering if the by lines interface is really necessary here, maybe it could be dropped for now until someone requests it?)
 -  A `nowait` parameter is added to `output_watch` and `output_watch_lines` for callers that know the output eventlog already exists. This avoids watching the intermediate eventlogs when the caller is already watching those eventlogs.

src/bindings/python/flux/job/output.py
```diff
@@ -0,0 +1,957 @@
...
+        super().__init__(entry)
+        # LogEvent class is only for 'log' events
+        if self.name != "log":
+            raise ValueError(f"event {self.name} is not a log event")
+        self.level = self.context["level"]
+        self.rank = Taskset(self.context["rank"])
+        self.message = self.context["message"]
+        self.component = self.context.get("component", None)
+        self.file = self.context.get("file", None)
+        self.line = (self.context.get("line", -1),)
```
chu11, line 169: not clear to me why self.line is a tuple?

src/bindings/python/flux/job/output.py
```diff
@@ -0,0 +1,957 @@
...
+            raise ValueError("event {self.name} is not an exception event")
+        self.severity = self.context["severity"]
+        self.exc_type = self.context["type"]
+        self.note = self.context["note"]
+
+    def render(self):
+        result = f"job.exception: type={self.exc_type} severity={self.severity}"
+        if self.note:
+            result += f" {self.note}"
+        return result
```
chu11, line 254: i was debating in my head, given note is a requirement now, do we want to output empty string note if none set??  I'm going back and forth about it ... not sure.

src/bindings/python/flux/job/output.py
```diff
@@ -0,0 +1,957 @@
...
+    def _watch_output(self, future):
+        #  Watch output events and propagate to output_watch_future.
+        #
+        #  Note: handle and propagate OSError from future.get_event() for
+        #  reasons noted in wait_for_start_event() below.
+        #
+        event = None
+        try:
+            event = future.get_event(autoreset=False)
+            # Note: do not fulfill
```
chu11, line 474: "do not fulfill immediately" is what is meant?

src/bindings/python/flux/job/output.py
```diff
@@ -0,0 +1,957 @@
...
+    to get the next available output event from the Future. The event will
+    be of type OutputEvent, LogEvent, OutputHeaderEvent, RedirectEvent, or
+    JobExceptionEvent (check 'name' attribute or type to get type of event)
+
+    If the job has a fatal exception before the 'start' event, then the
+    future will be fulfilled with an error with errno=EIO and message
+    'job {jobid} never started'.
+
+    Args:
+        jobid (int, JobID, str): jobid to watch
```
chu11, line 755: flux_handle arg?  ehh

src/bindings/python/flux/job/output.py
```diff
@@ -0,0 +1,957 @@
...
+    If the job receives an exception while watching output, an appropriate
+    error message is emitted to the stderr stream. Similarly, redirect
+    events generate a "redirected to" message on stderr.
+
+    If the job has a fatal exception before the 'start' event, then the
+    future will be fulfilled with an error with errno=EIO and message
+    'job {jobid} never started'.
+
+    Args:
+        labelio (bool): label lines of output with source tasks (default=False)
```
chu11, line 788: flux handle and jobid arg?  atleast jobid since you did that earlier.  also for many of the below.

chu11 (commented):
overall LGTM, this is what I noticed.

src/bindings/python/flux/job/output.py
grondo: While I think an empty note makes coding around job exception events cleaner, as a message to the user an empty note isn't necessary and may even be a little confusing. It isn't too important, but I'd prefer to leave it off at this point to make the output nicer.

src/bindings/python/flux/job/output.py
grondo: Hm, I think the rest of the comment must have gotten accidentally chopped off at some point. I can't really remember what was being said here, but there's a nice long comment in the following conditional that seems to explain the situation so I'll just delete this stray comment.

t/python/t0015-job-output.py
```diff
@@ -0,0 +1,575 @@
...
+            for stream, data in output_watch(
+                self.fh,
+                jobid,
+                labelio=label,
+                log_stderr_level=log_stderr_level,
+            ):
+                result.setdefault(stream, []).append(data)
+            return {
+                key: "".join(result.get(key, [])) for key in ("stdout", "stderr", "log")
+            }
```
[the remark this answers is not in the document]
grondo: Good catch, should have been passed through to `output_watch()`

t/python/t0015-job-output.py
```diff
@@ -0,0 +1,575 @@
...
+                keepends=keepends,
+                nowait=nowait,
+            ):
+                result.setdefault(stream, []).append(line)
+            if nojoin:
+                return result
+            return {
+                key: "\n".join(result.get(key, [])) + "\n"
+                for key in ("stdout", "stderr", "log")
+            }
```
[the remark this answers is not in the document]
grondo: Good idea!

grondo:
Thanks for the review @chu1! Pushed fixes (squashed) to address your comments, plus found a bug in the testsuite where a thread was using the same Flux handle as the main test program (fixed).

grondo:
Hm, looks like Read the Docs now requires a config file. I've pushed one here based off the minimal config described at

 https://blog.readthedocs.com/migrate-configuration-v2/

chu11 (approved):
those small things from before were the only things I noticed.  LGTM.

t/python/t0015-job-output.py
```diff
@@ -0,0 +1,600 @@
...
+            result=None,
+            label=False,
+            log_stderr_level=LOG_TRACE,
+            keepends=False,
+            nojoin=False,
+            nowait=False,
+        ):
+            if result is None:
+                result = {}
+            count = 0
```
chu11, line 434: just thought about this post my PR approval, separate stdout_count and stderr_count?  Probably not necessary since you check stdout/stderr output below to ensure they match.   Just to be thorough and would allow the additional count check below that you skip b/c of the potential for log output.

grondo: Not a bad idea. Maybe I'll try it to see how messy it is. There is one benefit to checking an overall count though -- it would catch cases where we're getting callbacks we don't expect where stream isn't `stderr` or `stdout`... maybe I'll just leave as is for now.

There is another gotcha with the `*lines` watchers -- if a partial line exists within a given eventlog `data` event, that partial will be treated as a whole line currently. That is, the `JobOutputWatchLines` class doesn't currently buffer lines, so really this is kind of "best effort".

This could be fixed in the future if necessary, or we could drop the "lines" based watchers if that shortcoming is a showstopper. (callers could always use `str.splitlines()` if they want to do something per line), though I figured it is maybe acceptable for now (I think most output is attempted to be line buffered at the job shell).

Because of that, I'm not sure pedantic line counting is worth the effort at this point.

chu11: > There is another gotcha with the *lines watchers -- if a partial line exists within a given eventlog data event, that partial will be treated as a whole line currently. That is, the JobOutputWatchLines class doesn't currently buffer lines, so really this is kind of "best effort".

Is this documented in the class?  Didn't see an obvious mention of it (on phone so could has missed) but might be good to add.

grondo: Yeah, I was still wondering if the watch by lines class should instead just be dropped for now? Unsure if there's a compelling use case at this point? If you think it is fine to keep with the comments, then I'm fine leaving it in too.

chu11: I'd say leave it in, I think it's a "net win" for now.  If it becomes a problem, could enhance the class later.

grondo: Ok, that sounds good to me. I'll add some caveats to the docstrings

src/bindings/python/flux/job/output.py
```diff
@@ -0,0 +1,982 @@
...
+
+    If the job has a fatal exception before the 'start' event, then the
+    future will be fulfilled with an error with errno=EIO and message
+    'job {jobid} never started'.
+
+    Note:
+        The current implementation may return a partial line as a full line
+        from ``getline()`` if a partial line is written into a job output
+        eventlog entry. That is, this implementation does not currently line
+        buffer and wait for full lines to fulfull to JobOutputWatchLines
```
chu11, line 837: fulfull -> fulfill?

grondo: Good catch. Noticed another typo too! :facepalm:

grondo:
I have a branch with a conversion of `flux submit` and `flux bulksubmit` `--watch` to use this interface and a prototype `flux watch` command and was just making sure nothing popped out from that work. I'll make a WIP PR in a bit.

grondo:
Actually, things seem to working fine here and smaller issues that pop up (if any) can be fixed in follow on PRs. I'll set MWP.

grondo:
@mergifyio refresh

PR #5335, opened 2023-07-15
python: clear KVS txns on commit error

### LINKED ISSUE #5333: python/kvs: user can't clear internally staged transactions if there is a commit error
i.e. 

```
flux.kvs.put(handle, ".", "foo")
try:
   flux.kvs.commit() // this raises an exception because can't overwrite root dir
except:
  pass

flux.kvs.put(handle, "something else", "foo")
flux.kvs.commit()
```

the last commit() will always fail b/c the internally staged bad change is still stuck in there.

User would have to have internal knowledge to know to wipe/clear/cleanup/etc.  `handle.aux_txn`.

Maybe this is just a part of issue #5327 ... modernizing this API

chu11:
Problem: When a commit error happens on a KVS transaction, an error is raised.  This means that the internal KVS transaction (created via calls to put(), put_mkdir(), etc.) does not get cleared.  Therefore a subsequent call the commit() would redo the the transaction.  If the transaction is the cause of the commit failure, no subsequent commits can ever work.
    
Solution: If an exeption is raised during the commit, clear the internal transactions before returning the exception to the user.
    
Fixes #5333

### DIFF
--- a/src/bindings/python/flux/kvs.py
+++ b/src/bindings/python/flux/kvs.py
@@ -203,9 +203,12 @@ def commit(flux_handle, flags: int = 0):
     if flux_handle.aux_txn is None:
         return
     future = RAW.flux_kvs_commit(flux_handle, None, flags, flux_handle.aux_txn)
-    RAW.flux_future_get(future, None)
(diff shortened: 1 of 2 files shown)


grondo (approved):
LGTM!

PR #5339, opened 2023-07-17
resource: fix problem with exclusions when R is dynamically discovered

### LINKED ISSUE #5337: resource: exclusion cannot be configured when resources are dynamically discovered
Problem: a test instance cannot be started with `resource.exclude` set.

```
$ cat foo.toml
[resource]
exclude = "0"

$ src/cmd/flux start -o,--config-path=foo.toml
Jul 17 07:18:54.353304 resource.err[0]: error decoding exclude set 0: ��: No such file or directory
Jul 17 07:18:54.353320 resource.crit[0]: module exiting abnormally
Jul 17 07:18:54.353619 broker.err[0]: rc1.0: flux-module: load resource: No such file or directory
Jul 17 07:18:54.353920 broker.err[0]: rc1.0: /home/garlick/proj/flux-core/etc/rc1 Exited (rc=1) 0.2s
```

I think the issue is that _R_ is dynamically discovered in this case, and thus is not available when exclusions are applied.  Presumably the problem also occurs with instances started under a foreign resource manager although I have not tried that.

The error message is using an uninitialized `flux_error_t` for that case so that also needs to be fixed,

garlick:
Problem: when _R_ is acquired dynamically, the resource module refuses to process exclusions.
    
`inventory_targets_to_ranks()` fails if _R_ is not set, but _R_ is only used if the input fails to parse as an idset.
    
 Move the check for NULL R into the hostnames to ranks mapping block where it is used.  This allows an exclusion set specified as an idset  to be decoded when R is dynamically discovered.
    
Fixes #5337

### DIFF
--- a/src/modules/resource/inventory.c
+++ b/src/modules/resource/inventory.c
@@ -206,24 +206,25 @@ struct idset *inventory_targets_to_ranks (struct inventory *inv,
                                           const char *targets,
                                           flux_error_t *errp)
 {
-    struct idset *ids;
+    struct idset *ids = NULL;
 
-    if (!inv->R) {
-        errno = ENOENT;
-        return NULL;
-    }
     if (!(ids = idset_decode (targets))) {
         /*  Not a valid idset, maybe an RFC29 Hostlist
          */
         flux_error_t err;
-        struct rlist *rl = rlist_from_json (inv->R, NULL);
-        ids = rlist_hosts_to_ranks (rl, targets, &err);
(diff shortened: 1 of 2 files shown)


garlick:
oops forgot to push related fix.  Stand by.

grondo (approved):
LGTM!

garlick:
Fixed a small mem leak that @grondo pointed out and rebased on current master.  I'll set MWP - thanks!

PR #5341, opened 2023-07-18
resource: ignore live resource.exclude changes

garlick:
Problem: the resource module contains code to process live changes to the configured exclude set, but as discussed in flux-framework/rfc#383, there is no longer much of a use case now that flux-batch(1) et al have a `--conf` option that can configure exclusions on the command line for subinstances.
    
The extra code includes posting `exclude`/`unexclude` events to the KVS resource.eventlog _synchronously_ by calling the internal function `reslog_sync()`.  This makes the resource module unresponsive while the KVS commit completes.
    
This extra complexity is no longer justified.  Remove.

Update tests.

There are also a couple of incidental man page updates here which could be split off to a separate pr if desired.

### DIFF
--- a/doc/man1/flux-broker.rst
+++ b/doc/man1/flux-broker.rst
@@ -45,7 +45,7 @@ OPTIONS
 **-S, --setattr**\ =\ *ATTR=VAL*
    Set initial value for broker attribute.
 
-**-c, --conf-path=**\ =\ *PATH*
+**-c, --config-path=**\ =\ *PATH*
    Set the PATH to broker configuration. If PATH is a directory, then
    read all TOML files from that directory. If PATH is a file, then load
    configuration as JSON if the file extension is ``.json``, otherwise
--- a/doc/man5/flux-config-resource.rst
+++ b/doc/man5/flux-config-resource.rst
@@ -71,6 +71,9 @@ norestrict
    when the Flux system instance is constrained to a subset of cores,
    but jobs run within this instance should have access to all cores.
 
+noverify
+   (optional) If true, disable the draining of nodes when there is a
+   discrepancy between configured resources and HWLOC-probed resources.
 
 EXAMPLE
 =======
--- a/doc/man5/flux-config-resource.rst
+++ b/doc/man5/flux-config-resource.rst
@@ -57,10 +57,6 @@ config
 exclude
    (optional) A string value that defines one or more nodes to withhold
    from scheduling, either in RFC 22 idset form, or in RFC 29 hostlist form.
-   This value may be changed on a live system by reloading the configuration
-   on the rank 0 broker.  A newly excluded node will appear as "down" to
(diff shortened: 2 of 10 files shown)


garlick:
Aww darn, that made fluxion sad.  I guess a sched pr would need to go in first.
```
2023-07-18T01:44:05.1056845Z PASS: t1013-exclusion.t 1 - exclusion: generate jobspecs
2023-07-18T01:44:05.1063165Z PASS: t1013-exclusion.t 2 - exclusion: load config with resource exclusions
2023-07-18T01:44:05.1068568Z PASS: t1013-exclusion.t 3 - load test resources
2023-07-18T01:44:05.1073753Z PASS: t1013-exclusion.t 4 - exclusion: loading fluxion modules works
2023-07-18T01:44:05.1078865Z FAIL: t1013-exclusion.t 5 - exclusion: all jobs are rejected on satisfiability
2023-07-18T01:44:05.1083962Z PASS: t1013-exclusion.t 6 - exclusion: removing fluxion modules
2023-07-18T01:44:05.1089024Z PASS: t1013-exclusion.t 7 - exclusion: reloading resource with a config file
2023-07-18T01:44:05.1094195Z PASS: t1013-exclusion.t 8 - exclusion: loading fluxion modules works
2023-07-18T01:44:05.1105141Z FAIL: t1013-exclusion.t 9 - exclusion: only 1-node jobs are satisfiable
2023-07-18T01:44:05.1105614Z FAIL: t1013-exclusion.t 10 - exclusion: cancel the running job
2023-07-18T01:44:05.1106012Z PASS: t1013-exclusion.t 11 - cleanup active jobs
2023-07-18T01:44:05.1106412Z PASS: t1013-exclusion.t 12 - exclusion: removing fluxion modules
2023-07-18T01:44:05.1106799Z ERROR: t1013-exclusion.t - exited with status 1
```

grondo (approved):
LGTM. One thought I had is that maybe a warning should be issued if the config is updated with `resource.exclude` set, just in case someone was using this for some reason. However, that's admittedly very unlikely, and we current don't warn about other ignored config updates, so perhaps not necessary.

garlick:
That comment reminded me that we may have given out advice to exclude rank 0 when the rank 0 broker is expected to be busy.  FWIW this [FAQ entry](https://flux-framework.readthedocs.io/en/latest/faqs.html#jobs-questions) recommends draining rank 0 in that case, which until recently was much easier that setting `resource.exclude`.  So I think maybe a warning is not necessary?

Oh but I missed updating `flux-config-resource(5)` which does include verbiage about live updates to `resource.exclude`!  I'll push a fix for that.

grondo:
>  So I think maybe a warning is not necessary?

Yeah, since other resource config updates are silently ignored we might as well be consistent. The warning probably adds little value at this point.

garlick:
Thanks for the review, setting MWP

PR #5346, opened 2023-07-20
python: fix potential gc of Future ffi handle before Future destruction

grondo:
This PR fixes a frustrating issue I was hitting while working on #5332 which uses the `FutureExt` class.

The crux of the issue is that a Python `Future` ffi handle for the `then_cb` (`future.cb_handle`) can be garbage collected before the `Future` itself is. If the C `flux_future_t` can be fufilled multiple times, or if the future was not reset, then the reactor may try to invoke the Python `continuation_callback()` again, and Python throws a little fit and aborts since the ffi handle is no longer valid.

Thanks to help from @trws, the problem was narrowed down to the `finally` block in `continuation_callback()`:
https://github.com/flux-framework/flux-core/blob/837104c07ee7400bb2f3413c9bd0b04b971143bd/src/bindings/python/flux/future.py#L46-L52

Since `py_future.cb_handle` is set to `None`,  the ffi handle could be garbage collected at this point. This works fine for futures that only call the `then_cb` a single time. But Futures with multiple fulfillment could end up calling the `then_cb()` again until the `Future` itself is GC'd (at which point `flux_future_destroy()` is called by the destructor).

The solution as suggested by @trws is to just avoid setting `cb_handle = None` and let the ffi handle lifetime match that of the `Future` object. This should completely eliminate the chances of a Python abort due to this problem. (And I probably spent many hours tracking down various ways this can happen and working around it while working on the job output code)

The other issue resolved here is that if an exception is raised in a callback, there is now a chance that the future is still active and fulfilled with an error (e.g. if the Future still has a reference and is not collected by gc). If the caller restarts the reactor, then the future callback will be called again, the same error raised, etc. Therefore, after stashing the exception in the handle, reset and stop the affected future to avoid this issue.

### DIFF
--- a/src/bindings/python/flux/future.py
+++ b/src/bindings/python/flux/future.py
@@ -46,6 +46,14 @@ def continuation_callback(c_future, opaque_handle):
         flux_handle = py_future.get_flux()
         type(flux_handle).set_exception(exc)
         flux_handle.reactor_stop_error()
+        #
+        # Reset this future so it doesn't immediately call the callback
+        # again (and fail) if the reactor is restarted. (and bypass
+        # future.reset() to avoid unnecessarily incrementing _THEN_HANDLES
+        # count).
+        #
+        py_future.pimpl.reset()
+        py_future.stop()
 
     finally:
         if py_future in _THEN_HANDLES:
--- a/src/bindings/python/flux/future.py
+++ b/src/bindings/python/flux/future.py
@@ -48,12 +48,22 @@ def continuation_callback(c_future, opaque_handle):
         flux_handle.reactor_stop_error()
 
     finally:
-        _THEN_HANDLES[py_future] -= 1
-        if _THEN_HANDLES[py_future] <= 0:
-            # allow future object to be garbage collected now that all
-            # registered callbacks have completed
-            py_future.cb_handle = None
-            del _THEN_HANDLES[py_future]
+        if py_future in _THEN_HANDLES:
+            _THEN_HANDLES[py_future] -= 1
+            if _THEN_HANDLES[py_future] <= 0:
+                #
+                #  Note, a multiply fulfilled future which is not reset
+                #  before leaving the then_cb() will end up here, since a
+                #  call to reset() is the only thing that increments the
+                #  _THEN_HANDLES counter. If py_future.cb_handle is set
+                #  to None at this point, then the handle could be garbage
+                #  collected immediately. If the Future is not also collected,
+                #  then continuation_callback() might be called again with
+                #  the same cb_handle, which is now invalid, and Python will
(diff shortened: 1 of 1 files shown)


grondo:
Oops, this was based on #5330, but there's no reason that's necessary so I've based it on current master.

I was unable to create a standalone issue for the problems being solved, so unfortunately this doesn't come with tests.

grondo:
> The other issue resolved here is that if an exception is raised in a callback, there is now a chance that the future is still active and fulfilled with an error (e.g. if the Future still has a reference and is not collected by gc). If the caller restarts the reactor, then the future callback will be called again, the same error raised

BTW, this issue should not come up much in practice, since an exception in a callback stops the reactor with an error _and_ reraises the exception when the reactor is exiting. A user would have to handle the exception from `handle.reactor_run()` and also restart the reactor instead of raising some other kind of program error. Where this did occur is in the testsuite, where we specifically check for proper handling of an exception in a callback on one test of many.

src/bindings/python/flux/future.py
```diff
@@ -48,12 +48,22 @@ def continuation_callback(c_future, opaque_handle):
...
-            # registered callbacks have completed
-            py_future.cb_handle = None
-            del _THEN_HANDLES[py_future]
+        if py_future in _THEN_HANDLES:
+            _THEN_HANDLES[py_future] -= 1
+            if _THEN_HANDLES[py_future] <= 0:
+                #
+                #  Note, a multiply fulfilled future which is not reset
+                #  before leaving the then_cb() will end up here, since a
+                #  call to reset() is the only thing that incrementsthe
```
garlick, line 65: need a space between increments and the

garlick (approved):
This sounds like it is well understood.  I don't necessarily have the big picture here but I trust this is a step forward and it is very well documented.

Just one minor comment nit inline.

src/bindings/python/flux/future.py
trws:
```suggestion
                #  call to reset() is the only thing that increments the
```

trws (approved):
Very nice, I like it.

src/bindings/python/flux/future.py
grondo: Ah, good catch, thanks!

grondo:
Thanks @trws and @garlick! I'll fix up the typo and repush.

src/bindings/python/flux/future.py
grondo: Ok fixed, thanks!

grondo:
Setting MWP. :sweat_smile:

src/bindings/python/flux/future.py
grondo: I wonder why the typos module didn't catch this one?

PR #5348, opened 2023-07-21
job-exec: remove systemd exec prototype

garlick:
This removes prototype code for running jobs under systemd, as this work has largely been replaced by #5197.

I should note that the prototype demonstrated restarting flux with running jobs, limited to one node.  We will want to consider that design going forward, but it is not compatible with the libsubprocess-centric replacement,  which is implemented as a near term (though reusable for long term) way to do memory containment and more reliable cleanup, with minimal changes to `job-exec`.

In going through this I was reminded that there is not currently valgrind coverage for the replacement code, so I'll propose that later.

Also I wanted to acknowledge the hard work that went into this prototype by @chu11.  The replacement would have been much much harder or impossible had Al not figured out so many of the minute details in the prototype.  The prototype also made clear some deficiencies in the libsystemd reactive API that led directly to the replacement's design - again hard to see that on the first go.


chu11:
No issues with removing in general, but one thought.  Do we want to let it sit around a little longer until the remaining stuffs in:

https://github.com/flux-framework/flux-core/pull/4255

gets pulled in?  Mostly testing in that PR I think.

garlick:
My thought was to use #4255 as reference or "test donor" as is (without merging it).  Unless it's rebased on master after this PR, it should still serve as a functional demo.  Please don't remove it though!  Good stuff in there.

chu11:
Here's a possibly dumb idea.  Merge #4255 so we have it in the history ... then remove everything?  So at least everything is there??  I'm more concerned I accidentally remove the branch in my tree :-)

grondo:
>  Do we want to let it sit around a little longer until the remaining stuffs in:

That PR is a great reference, but I'm not sure there's sense of pulling it in to just pull it back out again?

I'll defer to @garlick on that though. Its been long enough since I looked at that and I didn't review the new sdexec implementation in detail so I don't know if we pull that in if some stuff would stay or not though..

grondo:
D'oh, sorry for the late comments, github didn't update my copy of the page so I hadn't realized there were comments above on this topic already.

chu11:
> That PR is a great reference, but I'm not sure there's sense of pulling it in to just pull it back out again?

ahhh sorry, I now realize my original comment wasn't super clear.  

There's "pull in" as in we add the features, but not necessarily merge that PR.  And there's merge and then delete as an alternate option.

garlick:
I'd vote for the former: we leave the PR there for a bit while we mine it for valuables to pull in.

chu11 (approved):
LGTM

PR #5353, opened 2023-07-26
job-manager: prevent jobs with outstanding epilog-start events from becoming inactive

### LINKED ISSUE #5345: job-manager: epilog should prevent `clean` event
As soon as a rabbit job is submitted, flux-coral2 plugins create some external resources for it. When the job reaches `cleanup` state, a flux-coral2 jobtap plugin places an epilog on the job to clean up those resources. The epilog-start initiates the cleanup, the epilog-remove indicates that the cleanup has completed. However, when the job never has resources allocated to it (because of an exception), the epilog doesn't prevent the job from moving to `inactive` state, and the jobtap plugin hits an error when it tries to remove the epilog. 

On the exception path, the epilog isn't strictly necessary, and I could add logic to the plugins to avoid placing the epilog at all (even though the external cleanup would still be required, we could do it asynchronously). But it would be convenient to be able to handle both cases the same way.

When an exception occurs, the eventlog may look like:

```
1689787285.137467 epilog-start description="dws-epilog"
1689787285.137490 clean
```

whereas through the normal path it looks like

```
1689787300.987355 start
1689787301.106616 finish status=0
1689787301.106683 epilog-start description="dws-epilog"
1689787301.111610 release ranks="all" final=true
1689787315.361298 epilog-finish description="dws-epilog" status=0
1689787315.366933 free
1689787315.367069 clean
```

grondo:
This PR fixes #5345, where a job that is canceled before an `alloc` event could become inactive (have a `clean` event posted) even while an epilog action was in progress (outstanding `epilog-start` event).

This PR simply checks for an enqueued `epilog-start` event or `perilog_active` count before allowing the `clean` event to be emitted. Also adds a test that reproduces the original issue.

### DIFF
--- a/src/modules/job-manager/event.c
+++ b/src/modules/job-manager/event.c
@@ -423,7 +423,9 @@ int event_job_action (struct event *event, struct job *job)
                 && !job->alloc_pending
                 && !job->free_pending
                 && !job->start_pending
-                && !job->has_resources) {
+                && !job->has_resources
+                && !job_event_is_queued (job, "epilog-start")
+                && !job->perilog_active) {
 
                 if (event_job_post_pack (event, job, "clean", 0, NULL) < 0)
                     return -1;
--- a/t/t2212-job-manager-plugins.t
+++ b/t/t2212-job-manager-plugins.t
@@ -482,6 +482,16 @@ test_expect_success 'job-manager: epilog works after exception during prolog' '
 	n_free=$(lineno job.free perilog-exception-test.out) &&
 	test $n_epilog -lt $n_free
 '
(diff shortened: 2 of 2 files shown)


jameshcorbett (approved):
Code and test changes look reasonable, if you'd like I'll confirm it works in my reproducer tomorrow.

grondo:
> if you'd like I'll confirm it works in my reproducer tomorrow.

Sure, that would be great! Thanks.

jameshcorbett:
Sorry for the delay, it fixes the issue with my reproducer!

grondo:
Thanks for the confirmation! I'll set MWP.

PR #5354, opened 2023-07-26
flux-job: suppress the `attach` status line in more situations

grondo:
This PR makes it a little more likely that the `flux job attach` status line will be suppressed if its output is not desirable.

Since we really only want to display the status line if `flux job attach` is actually being run interactively, only enable it if all of `stdout`, `stderr`, and `stdin` are attached to a tty. Also add a `FLUX_ATTACH_NONINTERACTIVE` environment variable that always suppresses the status line for environments like ATS where a pty is used to run jobs but the output is logged. In this case the status line generates a lot of output since theres a timer updated every second.


garlick (approved):
LGTM

grondo:
Thanks! I'll set MWP

PR #5355, opened 2023-07-27
ensure flux utilities agree that a job with a fatal exception has failed

### LINKED ISSUE #5307: flux job attach and flux jobs show job with fatal exception as successful
Problem: when a jobtap plugin raises a fatal exception from the `job.event.alloc` callback, the job might still run to completion and be declared successful by `flux job attach` and `flux jobs`.

Note: `flux job wait` identifies the job as failed.

See #5304

grondo:
As pointed out in #5307, `flux job attach` and `flux jobs` ignore fatal job exceptions if the job has a `finish` event with exit status `0`. This doesn't agree with `flux job wait` and `flux job status`, which exit with failure for any fatal job exception.

Given that a fatal exception by definition probably indicates that the job has failed in some way regardless if the job tasks finished successfully, update the behavior of these commands (as well as `flux submit` and `flux bulksubmit`) to exit with failure for any fatal job exception. Additionally, `flux jobs` now only reports a successful job if the job finished with `status=0` _and_ there were no fatal job exceptions.

Because the "alloc check" tests already generated the failing case here (job with fatal job exception but all tasks finish successfully with `status=0`) add checks to this test to ensure that the various utilities now agree that the job with the alloc-check exception has "failed".

Fixes #5307

### DIFF
--- a/src/cmd/flux-job.c
+++ b/src/cmd/flux-job.c
@@ -1595,6 +1595,7 @@ struct attach_ctx {
     int eventlog_watch_count;
     bool statusline;
     char *last_event;
+    bool fatal_exception;
 };
 
 void attach_completed_check (struct attach_ctx *ctx)
@@ -2570,6 +2571,8 @@ void attach_event_continuation (flux_future_t *f, void *arg)
                          severity,
                          note);
 
+        ctx->fatal_exception = (severity == 0);
+
         /*  If this job has an interactive pty and the pty is not yet attached,
          *   destroy the pty to avoid a potential hang attempting to connect
          *   to job pty that will never exist.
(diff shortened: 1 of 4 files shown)


src/cmd/flux-job.c
```diff
@@ -1595,6 +1595,7 @@ struct attach_ctx {
     int eventlog_watch_count;
```
garlick, line 1595: Commit message for "flux-job: attach: always exit nonzero on fatal job exception"

s/if all job finishes/if a job finishes/

garlick (approved):
LGTM!  Just one typo.

grondo:
Thanks! Fixed the typo and will set MWP.

Issue #5352, opened 2023-07-25
shell: `cpu-affinity` `map` option should be able to override cores assigned in R


grondo:
Current the `-o cpu-affinity=map` option causes a fatal error if a core is assigned that falls outside the core list assigned in R. However, this option is meant for situations where the user wants explicit control of mapping of tasks to cores, so failing in this case reduces the usefulness of this option.

PR #5356, opened 2023-07-28
shell: support `-o gpu-affinity=map:LIST`

grondo:
This PR adds support for `gpu-affinity=map:LIST`. This can come in handy for benchmarking, testing, etc. Also, since `-o cpu-affinity` supports `map:`, it seemed like GPUs should be able to assigned in this way.

As a side effect of the approach, #5352 was also fixed.

### DIFF
--- a/doc/man1/flux-shell.rst
+++ b/doc/man1/flux-shell.rst
@@ -267,9 +267,11 @@ options are supported by the builtin plugins of ``flux-shell``:
   If *OPT* starts with ``map:``, then the rest of the option is taken
   as a semicolon-delimited list of cpus to allocate to each task. Each
   entry in the list can be in one of the :linux:man7:`hwloc` list,
-  bitmask, or taskset formats (See :linux:man3:`hwloc_cpuset_t`).
-  The default is ``on``, which binds all tasks to the assigned set
-  of cores in the job.
+  bitmask, or taskset formats (See
+  `hwlocality_bitmap(3) <https://www.open-mpi.org/projects/hwloc/doc/v2.9.0/a00181.php>`_,
+  especially the ``hwloc_bitmap_list_snprintf()``, ``hwloc_bitmap_snprintf()``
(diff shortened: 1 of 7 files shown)


doc/man1/flux-shell.rst
```diff
@@ -276,7 +276,10 @@ options are supported by the builtin plugins of ``flux-shell``:
   sets ``CUDA_VISIBLE_DEVICES`` to the GPU IDs allocated to the job.
   *OPT* may be set to ``off`` to disable the plugin, or ``per-task``
   to divide allocated GPUs among tasks launched by the shell (sets a
-  different GPU ID or IDs for each launched task)
+  different GPU ID or IDs for each launched task) If *OPT* starts with
```
garlick, line 279: Missing period?

garlick: Also this refers to the cpu-affinity entry in the man page, but over there, the hwloc_cpuset_t(3) reference is a 404 in the [rendered page](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man1/flux-shell.html).

garlick (approved):
This seems like a good idea.  I didn't find anything to comment on except the man page entry.

doc/man1/flux-shell.rst
grondo: Oh, nice catch!

grondo: Hm, that manpage exists on some systems but not on linux,die.net. I guess I'll have to reference the hwloc documentation directly (which won't be too nice in a manpage but oh well)

grondo: Ok, I pushed an extra commit to fix the missing `hwloc_cpuset_t(3)` link. Used `hwlocality_bitmask(3)` instead, which has the actual `*sscanf(3)` functions used in the shell plugin. Since that manpage also does not exist on linux.die.net, an explicit link directly to the latest hwloc documentation is used.

grondo:
Setting MWP on this one.

PR #5357, opened 2023-07-28
python: add JobWatcher class and use for `submit` and `bulksubmit` `--watch` functionality

grondo:
This PR abstracts the `SubmitBulkCmd` "watch" functionality into a `JobWatcher` class, which uses the `flux.job.output` functionality for the output watching, and then uses this new class to support `submit` and `bulksubmit` `--watch` and `--wait` functionality.

This class can then be used for other utilities in the future, e.g. something like a `flux watch`.

There's no extra tests of the `JobWatcher` class at this point. Instead the existing bulksubmit and submit tests are used to test operation of the class.

### DIFF
--- a/src/bindings/python/flux/cli/base.py
+++ b/src/bindings/python/flux/cli/base.py
@@ -38,7 +38,7 @@ import flux
 from flux import debugged, job, util
 from flux.constraint.parser import ConstraintParser, ConstraintSyntaxError
 from flux.idset import IDset
-from flux.job import JobspecV1
+from flux.job import JobspecV1, JobWatcher
 from flux.progress import ProgressBar
 from flux.util import dict_merge, set_treedict
 
@@ -731,6 +731,7 @@ class MiniCmd:
         self.flux_handle = None
         self.exitcode = 0
         self.progress = None
+        self.watcher = None
         self.parser = self.create_parser(prog, usage, description, exclude_io)
 
     @staticmethod
@@ -1312,6 +1313,8 @@ class SubmitBaseCmd(MiniCmd):
 
     def run_and_exit(self):
         self.flux_handle.reactor_run()
+        if self.watcher:
+            self.exitcode = max(self.watcher.exitcode, self.exitcode)
         sys.exit(self.exitcode)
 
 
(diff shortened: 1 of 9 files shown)


src/bindings/python/flux/job/watcher.py
```diff
@@ -0,0 +1,574 @@
...
+            jps (bool): with ``progress=True`` show jobs per second instead of
+                timer on right hand side of progress bar
+            wait (str): Event to wait for before terminating watch of job.
+                (default="clean")
+            log_events (bool): Log all events on stderr (default=False)
+            log_status (bool): Log final status of jobs if applicable
+                (default=False)
+            stdout (TextIOWrapper): Default stdout location (default=sys.stdout)
+            stderr (TextIOWrapper): Default stderr location (default=sys.stderr)
+            labelio (bool): Label lines out output with jobid and taskid
```
chu11, line 363: s/out/of/

src/bindings/python/flux/job/watcher.py
```diff
@@ -0,0 +1,574 @@
...
+                or jobs (if None, then set to same location as ``stdout``)
+            wait: event at which to stop watching the job (default=clean)
+        """
+        job = flux.job.JobInfo(
+            {
+                "id": jobid,
+                "state": flux.constants.FLUX_JOB_STATE_SCHED,
+                "t_submit": time.time(),
+            }
+        )
```
chu11, line 460: i think to a casual observer this looks sort of odd, to be hardcoding this stuff.  I guess what's going on is gotta initialize to something, then the event watcher will take care of the rest.  Perhaps a small comment would suffice.

src/bindings/python/flux/job/watcher.py
```diff
@@ -0,0 +1,574 @@
...
+                #  shell never initialized, then print the exception error
+                #  to stderr:
+                if not job.has_event("shell.init"):
+                    print(
+                        flux.job.output.JobExceptionEvent(event).render(),
+                        file=job.stderr,
+                    )
+        elif event.name == "alloc":
+            job.status = "running"
+
```
chu11, line 514: errant empty line?

src/bindings/python/flux/job/watcher.py
```diff
@@ -0,0 +1,574 @@
...
+        elif event.name == "start":
+            if self.watch or job.wait.startswith("exec."):
+                flux.job.event_watch_async(
+                    self.flux_handle, job.id, eventlog="guest.exec.eventlog"
+                ).then(self._exec_event_cb, job, future)
+
```
chu11, line 520: errant empty line?

chu11 (commented):
through the JobWatch commit, just parking review so far before end of the night.  just small stuff.

src/bindings/python/flux/job/watcher.py
grondo: well it was left there on purpose because I felt it helped readability, but I can remove these lines if they seemed out of place

chu11: if intentional that's fine, i thought possibly mistake

src/bindings/python/flux/job/watcher.py
```diff
@@ -0,0 +1,574 @@
...
+        if status in ("DEPEND", "PRIORITY", "SCHED"):
+            return "pending", 0
+        elif status in ("RUN", "CLEANUP"):
+            return "running", 0
+        elif status == "COMPLETED":
+            return "complete", 0
+        elif status in ("FAILED", "CANCELED", "TIMEOUT"):
+            rc = job.returncode
+            if not rc:
+                rc = 1
```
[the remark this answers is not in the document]
grondo: An empty returncode can occur for a job that failed before a `start` event due to a non cancel type exception.
However, your comment made me realize the fix should really be applied to the  `JobInfo.returncode` property function, since as you said, that result would be surprising. I'll push a commit that does that and remove this little block.

grondo:
While working on `flux watch` I found that the output streams need to be flushed after all output has been handled or data might be lost (not exactly sure why this is, but I've added a flush to the `JobWatcher` to handle this case)

t/python/t0010-job.py
```diff
@@ -683,7 +690,21 @@ def cancel_on_start(future, jobid):
...
+                    "result": flux.constants.FLUX_JOB_RESULT_FAILED,
+                    "t_submit": 0.0,
+                    "exception_occurred": True,
+                    "exception_type": "test",
+                    "exception_note": "",
+                    "exception_severity": 0,
+                }
+            ),
+        )
+        self.assertEqual(result[ids[5]].get_info().returncode, 1)
```
chu11, line 707: why split out the "returncode" check into a separate check, could stick in the first check?  Dunno if subtlety I'm missing.

src/bindings/python/flux/cli/base.py
```diff
@@ -1519,41 +1401,56 @@ def submit_cb(self, future, args, label=""):
...
+            ).start()
+
     def jobs_per_sec(self):
         return (self.progress.count + 1) / self.progress.elapsed
 
     def progress_start(self, args, total):
         """
-        Initialize progress bar if one was requested
+        Initialize job submission progress bar user requested progress
+        without --wait or --watch
```
chu11, line 1442: need to nix second "progress" in this sentence?  Isn't reading quite right.

chu11 (approved):
LGTM, just two small comments below.

t/python/t0010-job.py
grondo: This makes the check explicit since returncode is a derived property

src/bindings/python/flux/cli/base.py
grondo: Yeah, I think it is actually a missing "if", will fix.

t/python/t0010-job.py
grondo: I pushed a version with a comment before this test to state the reason for the separate check.

grondo:
Oops, typo'd one of the comments. Will fix shortly.

grondo:
Thanks! Setting MWP.

PR #5358, opened 2023-07-28
 job-info: support WAITCREATE on eventlog watch 

chu11:
Per #5301

Some comments:

- pondered if this should be the default or if a flag / option was best.  Eventually settled on the latter as there may be scenarios we do not want to "waitcreate"

- the "style" in which tests are written in t2231-job-info-eventlog-watch.t is pretty ancient.  It could be cleaned up.  But for this PR, kept the style as is.   Figured cleanup / update can be for a follow up PR.

### DIFF
--- a/etc/completions/flux.pre
+++ b/etc/completions/flux.pre
@@ -463,6 +463,7 @@ _flux_job()
         -q --quiet \
         -v --verbose \
         -p --path= \
+        -W --waitcreate \
     "
     local info_OPTS="\
         -o --original \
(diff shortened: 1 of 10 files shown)


grondo (approved):
LGTM, but add new `--waitcreate` option to `etc/completions/flux.pre`?

chu11:
@grondo thanks.  re-pushed with extra commit for `--waitcreate` in the bash completions file.

grondo:
Great! Feel free to set MWP as far as I'm concerned.

PR #5360, opened 2023-07-28
cmd: add flux-watch(1)

grondo:
Putting this up as a WIP, in case others have suggestions or comments on this added command.

This PR adds a new command, `flux watch` for lack of a better name (I asked ChatGPT and one of its suggestions was `flux gazemancer` but that was just a bit too outlandish). This command basically does the same thing as `flux submit --watch` but can monitor all running jobs, all inactive jobs, or a specific set of jobs. (It has a similar set of options as `flux jobs`, `flux pstree` etc.).

This allows users to submit many jobs then watch then with `flux watch`, e.g. at the end of a batch script, or if you need to run multiple `submit` or `bulksubmit` commands to submit a set of jobs and then want to monitor them all, e.g. `flux watch --all --progress`...

I still need to think more about the set of options here, write tests, etc. But that should go fairly quickly since this is actually a pretty simple command.

### DIFF
--- a/doc/man1/flux-watch.rst
+++ b/doc/man1/flux-watch.rst
@@ -1,0 +1,92 @@
+.. flux-help-section: jobs
+
+============
+flux-jobs(1)
+============
+
+
+SYNOPSIS
+========
+
+**flux** **watch** [*OPTIONS*] [JOBID ...]
+
+DESCRIPTION
+===========
+
+The flux-watch(1) command is used to monitor the output and state of one
+or more Flux jobs. The command works similarly to the :man1:`flux-submit`
+``--watch`` option, but can be used to monitor even inactive jobs. For
+example, to copy all job output to the terminal after submitting a series
+of jobs with :man1:`flux-submit` or :man1:`flux-bulksubmit`, use
+
+::
+
+  flux watch --all
+
+This command can also be used at the end of a batch script to wait for all
+submitted jobs to complete and copy all output to the same location as the
+batch job.
+
+OPTIONS
+=======
+
+**-a, --active**
+   Watch all active jobs.
+   This is equivalent to  *--filter=pending,running*.
+
+**-A, --all**
+   Watch all jobs. This is equivalent to *--filter=pending,running,inactive*.
+
+**-c, --count**\ *=N*
+   Limit output to N jobs (default 1000). This is a safety measure to
+   protect against watching too many jobs with the ``--all`` option. The
+   limit can be disabled with ``--count=0``.
+
+**--since**\ *=WHEN*
+   Limit output to jobs that have been active since a given timestamp.
+   This option implies ``-a`` if no other ``--filter`` options are specified.
+   If *WHEN* begins with ``-`` character, then the remainder is considered
+   to be a an offset in Flux standard duration (RFC 23). Otherwise,
+   any datetime expression accepted by the Python `parsedatetime
+   <https://github.com/bear/parsedatetime>`_ module is accepted. Examples:
+   "-6h", "-1d", "yesterday", "2021-06-21 6am", "last Monday", etc. It is
+   assumed to be an error if a timestamp in the future is supplied.
+
+**-f, --filter**\ *=STATE|RESULT*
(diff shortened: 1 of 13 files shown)


src/cmd/flux-watch.py
```diff
@@ -0,0 +1,207 @@
...
+        args.filter = {"pending", "running"}
+
+    joblist = JobList(
+        fh,
+        ids=args.jobids,
+        attrs=["state", "result", "t_submit"],
+        filters=args.filter,
+        since=since,
+        user=str(os.getuid()),
+        max_entries=0,
```
chu11, line 181: I can't help but look at the "max_entries = 0" and the "--all" option and think "this will not end well".  Although I can't think of any better alternative, other than perhaps a "be careful what you wish for" in the help/manpage.

grondo: Yeah, I had the same thought and added a `--count=N` option with default of 1000. I'm still adding tests and fixing things as they come up, but will push a non WIP version shortly.

grondo:
Force-pushed a fixup for the tests when run with `--chain-lint`

grondo:
Ah, oops, I forgot that flux-sched and flux-accounting use `help_formatter`, so probably best to leave that one in `flux.util`. I'll fix that.

doc/man1/flux-watch.rst
```diff
@@ -0,0 +1,92 @@
...
+
+**-A, --all**
+   Watch all jobs. This is equivalent to *--filter=pending,running,inactive*.
+
+**-c, --count**\ *=N*
+   Limit output to N jobs (default 1000). This is a safety measure to
+   protect against watching too many jobs with the ``--all`` option. The
+   limit can be disabled with ``--count=0``.
+
+**--since**\ *WHEN*
```
chu11, line 45: "=WHEN"?

t/t2813-flux-watch.t
```diff
@@ -0,0 +1,127 @@
...
+	grep "Watching [12] job" since.out
+'
+test_expect_success 'flux-watch: invalid --since fails' '
+	test_expect_code 2 flux watch --since=-1g 2>since.err &&
+	grep "invalid value" since.err
+'
+test_expect_success 'flux-watch: --since in future fails' '
+	test_expect_code 2 flux watch --since=+1d 2>since2.err &&
+	grep "appears to be in the future" since2.err
+'
```
chu11, line 92: should cover negative offset too?  like `-1h`?

grondo:
Ok, fixed. Just ended up adding the `argparse.Action` subclasses to `flux.util` instead of making a new module, which greatly simplifies the end result (even if `flux.util` is getting a bit large)

src/bindings/python/flux/argparse.py
```diff
@@ -0,0 +1,142 @@
...
+        metavar="[yes|no]",
+    ):
+        super().__init__(
+            option_strings=option_strings, dest=dest, help=help, metavar=metavar
+        )
+
+    def __call__(self, parser, namespace, value, option_string=None):
+        if value not in ["yes", "no"]:
+            raise ValueError(f"{option_string} requires either 'yes' or 'no'")
+        setattr(namespace, self.dest, value == "yes")
```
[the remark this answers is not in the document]
grondo: This is just moved code, not anything new.

src/cmd/flux-watch.py
```diff
@@ -0,0 +1,208 @@
...
+        #  instance starttime
+        starttime = max(since, instance_starttime)
+
+    if args.active:
+        args.filter.update(("pending", "running"))
+
+    if args.all:
+        args.filter.update(("pending", "running", "inactive"))
+        if not starttime:
+            starttime = instance_starttime
```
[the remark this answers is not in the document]
grondo: Eh, "all" includes active, so not sure it is really an error or even needs a warning. Just an extra thing to test.. I'm on the fence, I could do it if you feel strongly.

chu11 (approved):
LGTM

grondo:
Thanks, I'll set MWP to get the ball rolling

PR #5369, opened 2023-08-03
python: clear kvs txn after all commits

chu11:
Problem: The internal KVS transaction is only cleared when commit()
causes an error.  It should be cleared after every call to commit().

Clear the internal KVS transaction after every call to commit().

Also fixed up a corner case I saw, I think the future needs to be destroyed too.

### DIFF
--- a/src/bindings/python/flux/kvs.py
+++ b/src/bindings/python/flux/kvs.py
@@ -206,10 +206,10 @@ def commit(flux_handle, flags: int = 0):
     try:
         RAW.flux_future_get(future, None)
     except OSError:
-        RAW.flux_kvs_txn_destroy(flux_handle.aux_txn)
-        flux_handle.aux_txn = None
         raise
     finally:
+        RAW.flux_kvs_txn_destroy(flux_handle.aux_txn)
+        flux_handle.aux_txn = None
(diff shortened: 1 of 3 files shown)


garlick (approved):
Good catch.  This LGTM.

grondo (commented):
Any way to add a simple test for this? Or is it untestable?

chu11:
> Any way to add a simple test for this? Or is it untestable?

Last night I couldn't think of anything, since we're just repeating the old commits over again and presumably getting the same results.

But thinking a little harder this morning, I think maybe I can trick the code using different relative paths with KVSDir.  Lemme give it a shot.

FWIW, I found this out in my KVS namespace branch b/c I was doing:

```
put("a", 1)
commit(namespace="test1")
put("b", 2)
commit(namespace="test2")
```

and both "a" and "b" were ending up in the "test2" namespace.

chu11:
added a regression test, my KVSDir trick wouldn't have worked.  Then I realized if I "externally" modified a KVS entry via a shell call and "flux kvs", then I could test the desired result in a regression test.

chu11:
oops, noticed a typo, re-pushed

PR #5370, opened 2023-08-03
docker: transition bionic to jammy

trws:
The bionic LTS release of ubuntu is falling out of standard support.

Update all references to bionic from the leaving LTS release to the current LTS release.  Builds of the fluxrm/testenv for arm64 and amd64 are already on dockerhub based on these changes.

### DIFF
--- a/src/test/docker/README.md
+++ b/src/test/docker/README.md
@@ -56,16 +56,17 @@ $ for i in focal el7 el8 fedora33 fedora34 fedora35 fedora38; do
   done
 ```
 
-#### Jammy multiarch images
+#### Bookworm and Jammy multiarch images
 
-Building the jammy images for linux/amd64 and linux/386 requires the
+Building the images for linux/amd64, linux/arm64 and linux/386 requires the
 Docker buildx extensions, see
 
  https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/
 
 and run
 ```
-$  docker buildx build --push --platform=linux/386,linux/amd64 --tag fluxrm/testenv:jammy src/test/docker/jammy
+$  docker buildx build --push --platform=linux/arm64,linux/amd64 --tag fluxrm/testenv:jammy src/test/docker/jammy
+$  docker buildx build --push --platform=linux/386,linux/amd64,linux/arm64 --tag fluxrm/testenv:bookworm src/test/docker/bookworm
 ```
 
 to build and push images to docker hub.


grondo:
Just as a reminder because I just thought of it, we should also update required checks/statuses and `.mergify.yml` to change the required checks as appropriate. Actually, I think mergify supports Github required checks now, so maybe the hand  edited list of `status-success` conditions in `.mergify.yml` could be trimmed back (or eliminated?)

trws:
Yup, that's the plan.  I'm trying to get these all passing, then once I'm sure the names are stable will update all of that.  Not sure about the mergify config but probably.  I actually meant to bring up whether we should switch to the built-in github functionality for that too.  I didn't know until today, but it looks like mergify actually breaks git-blame in some cases, which is not great.

grondo:
>  I didn't know until today, but it looks like mergify actually breaks git-blame in some cases, which is not great.

I hadn't heard that, and can't imagine how that could be with the rebase strategy, but maybe I'm not thinking of some corner case?

trws:
I'll have to try and find it again, I went to use git blame to figure out where enchant came from and every line told me it was last touched by mergify for some reason.

grondo:
Hm, maybe a local git config or git version? I was able to use `git blame` to find that commit I shared over on slack.

src/test/docker/jammy/Dockerfile
```diff
@@ -30,6 +30,12 @@ RUN apt-get update \
         jq \
  && rm -rf /var/lib/apt/lists/*
 
+# ubuntu containers now "minimize" themselves, so manpages aren't installed
+# to warn people about this, /usr/bin/man is a shell script that doesn't bother
```
chu11, line 34: nit: here and in commit message.  I think period after "installed" and upper case "to"?   Took a few reads to figure out.

trws:
```suggestion
# Ubuntu containers now "minimize" themselves, so manpages aren't installed.
# To warn people about this, /usr/bin/man is a shell script that doesn't bother
```

grondo:
@trws, I posted the remaining fix for the 32bit build [here](https://github.com/flux-framework/flux-core/issues/5372#issuecomment-1668044973). I can push a commit directly here or you can just grab that one. That _should_ allow the 32bit builder to finish successfully.

grondo (approved):
LGTM @trws, thanks! I've removed the stale actions from branch protections and will push the merge button shortly.

One final question though, were there some updated instructions for the README on how to deal with the linux/386 build + manifest, or are the current changes to the README sufficient?

Thanks!!

trws:
Good point, I'm tweaking the multiarch image language, just a sec.

trws:
Updated.  The commands are the same, but I hadn't added in the extra arch and the bookworm command so definitely better to add that.

grondo:
Awesome, thank you! I'll push the button once the CI is all green again.

Thanks @trws, this was a heroic effort!

grondo:
Looks good, merging! :tada:

PR #5371, opened 2023-08-03
python: have kvs.isdir() return False if key missing

chu11:
while working on other stuff, noticed that `flux.kvs.isdir()` in the python kvs module returns ENOENT if a key is missing.  I believe this is inconsistent to how one normally thinks of these kinds of functions.  If the key does not exist, I think the function should just return False.

At the same time I noticed this, it ends up there was no `isdir()` coverage in the python tests, so adding some.

If folks think the logic should not be changed, I can still add the tests (although tweaked a bit).

### DIFF
--- a/t/python/t0005-kvs.py
+++ b/t/python/t0005-kvs.py
@@ -98,12 +98,23 @@ class TestKVS(unittest.TestCase):
     def test_14_exists_false(self):
         self.assertFalse(flux.kvs.exists(self.f, "argbah"))
 
-    def test_15_commit_flags(self):
+    def test_15_isdir_true(self):
+        with flux.kvs.get_dir(self.f) as kd:
+            kd.mkdir("testisdir")
+        self.assertTrue(flux.kvs.isdir(self.f, "testisdir"))
+
+    def test_16_isdir_false(self):
+        flux.kvs.put(self.f, "testisdirkey", 15)
+        flux.kvs.commit(self.f)
+        self.assertFalse(flux.kvs.isdir(self.f, "testisdirkey"))
+        self.assertFalse(flux.kvs.isdir(self.f, "not_a_key_i_made"))
+
+    def test_17_commit_flags(self):
         flux.kvs.put(self.f, "flagcheck", 42)
         flux.kvs.commit(self.f, 1)
         self.assertTrue(flux.kvs.exists(self.f, "flagcheck"))
 
-    def test_16_remove(self):
+    def test_18_remove(self):
         kd = self.set_and_check_context("todel", "things to delete", str)
         del kd["todel"]
         kd.commit()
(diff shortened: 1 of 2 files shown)


chu11:
noticed a dumb cleanup that should be done, just threw the commit on top of this PR

t/python/t0005-kvs.py
```diff
@@ -378,7 +385,7 @@ def test_45_commit_fail(self):
 
     # just testing that passing flags work, these are pitiful KVS
     # changes and the flags don't mean much
-    def test_46_commit_flags(self):
+    def test_48_commit_flags(self):
```
garlick, line 388: Er do we really need to renumber all these tests?  How about just adding an "a", "b", "c" to the test inserted in the middle, or drop the numbers in a separate commit?

(Other `unittest` users don't seem to be numbered so I assume it's not a necessary convention)

garlick (approved):
Agreed, seems like the right fix!

Just a comment about the test.

t/python/t0005-kvs.py
chu11: I added the numbers a few PRs ago.  Basically the kvs tests grew enough in size that it was hard to debug when I was adding new tests.  Since the tests get run in lexicographical order, the test output was sort of all jumbled around in the output.  Maybe the right thing to do would be to split the tests up in the future so the file is a more manageable size, like many of the other tests.

garlick: Oh ok.  Well, making the new tests 14a, 14b might work then?  Not a big deal if you feel strongly that you want it this way.

grondo: I stopped numbering unittest tests because of this. The backtrace is usually enough to debug the failing tests

chu11: The re-numbering was annoying ... ok you've convinced me... lets clean up the tests! (I'll do a follow up PR, since this one is already approved)

chu11:
rebased and re-pushed.  Also, given comment above I realized that the new tests should run independently of other tests.  So just created some dirs/keys for those tests.  i.e.

```
--- a/t/python/t0005-kvs.py
+++ b/t/python/t0005-kvs.py
@@ -99,10 +99,14 @@ class TestKVS(unittest.TestCase):
         self.assertFalse(flux.kvs.exists(self.f, "argbah"))
 
     def test_15_isdir_true(self):
-        self.assertTrue(flux.kvs.isdir(self.f, "pytestdir"))
+        with flux.kvs.get_dir(self.f) as kd:
+            kd.mkdir("testisdir")
+        self.assertTrue(flux.kvs.isdir(self.f, "testisdir"))
 
     def test_16_isdir_false(self):
-        self.assertFalse(flux.kvs.isdir(self.f, "thing"))
+        flux.kvs.put(self.f, "testisdirkey", 15)
+        flux.kvs.commit(self.f)
+        self.assertFalse(flux.kvs.isdir(self.f, "testisdirkey"))

```

PR #5373, opened 2023-08-07
python: add namespace support to KVS module

chu11:
This originally started out as an experiment just to see if A) could we add some modernization to the Python kvs module without breaking/obliterating the old API and B) see if anything comes up to surprise us design wise in how we'd want to move forward with a modernized API.

In this PR I add namespace support to the kvs module via an optional namespace argument to all appropriate functions.  Overall I think it's ok.  I think it could go in w/o much fanfare.  Notes:

- Order of parameters still left for intended positional arguments.  My anal retentive side dislikes this, as "flags" is often listed positional before "namespace".  Left as is unless we want to intentionally break things.

- Also did a big cleanup of the t0005-kvs.py tests.  Both grouping/organizing the tests  (i.e. `walk()` tests not near each other, as we kept on appending tests to the bottom fo the file) and removing the global numbering of all the tests into just "groups" of tests (i.e. KVSDir tests get number, namespace tests gets numbering order, etc.).

### DIFF
--- a/src/bindings/python/flux/kvs.py
+++ b/src/bindings/python/flux/kvs.py
@@ -11,9 +11,11 @@
 import collections.abc as abc
 import errno
 import json
+import os
 from typing import Any, Mapping
 
 from _flux._core import ffi, lib
+from flux.rpc import RPC
 from flux.wrapper import Wrapper, WrapperPimpl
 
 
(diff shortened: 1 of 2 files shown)


chu11:
re-pushed with some cleanups / fixes given some pylint warnings

chu11:
re-pushed, rebasing on #5376, that one comes first

grondo (approved):
This all seems fine to me and pretty low risk since it doesn't seem to break any interfaces.

chu11:
@grondo thanks, setting MWP

PR #5376, opened 2023-08-08
python: fix __delitem__ in KVSDir with initial paths

chu11:
GAH, another "initial path" bug in KVSDir.  I missed this one in the `__delitem__` item accessor.

I stole some cleanup from PR #5373 to put at the beginning of this PR since it would make the test numbering update easier.  This one now comes first in the chain of PRs.

### DIFF
--- a/t/python/t0005-kvs.py
+++ b/t/python/t0005-kvs.py
@@ -291,7 +291,23 @@ class TestKVS(unittest.TestCase):
         self.assertEqual(kd5["d.e.f"], 4)
         self.assertEqual(kd5["d"]["e"]["f"], 4)
 
-    def test_kvsdir_10_fill_initial_path(self):
+    def test_kvsdir_10_unlink_initial_path(self):
+        with flux.kvs.get_dir(self.f) as kd:
+            kd.mkdir("unlinkinitialpath")
+            kd["unlinkinitialpath"]["a"] = 1
+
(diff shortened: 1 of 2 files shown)


t/python/t0005-kvs.py
```diff
@@ -311,7 +327,7 @@ def test_kvsdir_10_fill_initial_path(self):
             self.assertEqual(kd4["i.j.k"], "baz")
             self.assertEqual(kd4["i"]["j"]["k"], "baz")
 
-    def test_kvsdir_11_mkdir_initial_path(self):
+    def test_kvsdir_12_mkdir_initial_path(self):
```
grondo, line 330: Still renumbering here?

grondo (approved):
LGTM! One question about renumbering in the tests since I thought you had stopped doing that via the first commit. (However, I could be wrong after just a quick read, so ignore me if so!)

t/python/t0005-kvs.py
chu11: I decided to number within test "groups".  So its more limited than before.

chu11:
huh ... the build just failed in the middle of the python build with _rlist.c in the el8-ascii builder.  I'm not sure what happened.  I'll restart the builder in case it was a weird fluke.

image

grondo:
I've seen that happen before in CI, but not elsewhere, so I don't know the cause. It is rare, so rerunning the build usually resolves it.

PR #5377, opened 2023-08-08
docker-deploy: only push latest when arch is 64-bit

trws:
Apparently had a mistake in there.  On merge to master, the new default bookworm containers all went up, but the 386 version got tagged latest rather than the amd64 version.  This should fix that.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -7,7 +7,7 @@ queue_rules:
       - status-success="flux-sched check"
       - status-success="flux-accounting check"
       - status-success="python linting"
-      - status-success="jammy"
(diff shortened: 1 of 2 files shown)


grondo (approved):
LGTM!

trws:
@grondo, looks like mergify doesn't like this, maybe the coverage?

grondo:
@mergifyio refresh

grondo:
Ah, it looks like there's one status that mergify is waiting for that doesn't have a matching action:
2023-08-08-175453_405x115_scrot

PR #5379, opened 2023-08-09
actions: add merge_group trigger

trws:
Add the marge_group trigger in case we want to test out the github native merge queue functionality.  Just add that in, has no effect until and unless we enable the merge queue branch protection.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,4 +1,4 @@
-on: [pull_request, push]
+on: [pull_request, push, merge_group]
 name: ci
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}


grondo:
I've only done a cursory examination of Github merge queues, but after a quick review, don't PRs enter the merge queue only after all required branch protections have passed? Thus if we add `merge_group` trigger, will this require all status checks to rerun again after all checks are passing and the PR enters the merge queue? (I'm sure I'm not understanding something about how this works, sorry!)

Also, I'll go ahead and add required reviews and status checks to our branch protections so we're ready for this change (or to try it out even)

grondo (approved):
This seems fine to me. If it has unintended consequences we could easily back it so, so let's try it.

trws:
@mergifyio refresh

grondo:
One of the bookworm build failed with:
```
Error: not ok 28 - signal forwarding works
Error: ERROR: t0005-exec.t - exited with status 1
```
Obviously unrelated to this PR.

This same test was failing regularly when building in the TOSS buildfarm, so I guess there's some kind of race condition to fix here. For now I've tried restarting the build.

trws:
I've been getting that one on and off locally, it comes up much more 
frequently when running many tests in parallel, but I failed to figure 
out why.

PR #5381, opened 2023-08-09
mergify: remove status check conditions from config

grondo:
Problem: The required status checks in the mergify config are redundant with GitHub branch protections, and keeping the list of statuses up-to-date in the mergify config is quite painful since updates require a PR and manual merge.

Remove the status checks completely from `.mergify.yml`. Mergify will still ensure that required status checks in branch protections are satified, so this should not change any behavior.

This PR will need a manual merge since `.mergify.yml` was touched. :slightly_smiling_face:


grondo:
FYI - As a precursor to this I ensured all current status checks are correct in branch protections, and added 1 required review. Reviews will now also be dismissed when branches are force-pushed or updated (that one might end up being a bit of a pain though)

garlick (approved):
Sounds good!  ready for me to push button?

grondo:
Sure, once the ci completes!

grondo:
Thanks!

PR #5383, opened 2023-08-10
testsuite: fix race in t0005-exec.t signal test

### LINKED ISSUE #5382: t0005-exec.t - not ok 28 - signal forwarding works
This test has been failing in CI and buildfarm. As a test I ran it 100 times with 4 tasks oversubscribed to each core, e.g.:
```sh
$ flux submit --cc=1-100 -o pty --watch --progress sh -c './t0005-exec.t -d -v --root=$FLUX_JOB_TMPDIR'
```
and the test failed >50% of the time:
```
PD:0   R:0   CD:41  F:59  │██████████████████████████████████████│100.0% 0:07:41
```

grondo:
Problem: If the signal in the signal forwarding test in t0005-exec.t is delivered after the "hi" sentinel is printed but before sleep(1) is executed, then bash ignores the signal and the test will timeout and fail.

To prevent the race use awk instead of a bash script, so that there is no point where the delivered signal will not be handled.  The awk script prints the sentinel then reads from a fifo in order to block indefinitely (stdin in sharness may be closed), so there are no extra subprocesses involved.

Fixes #5382

### DIFF
--- a/t/t0005-exec.t
+++ b/t/t0005-exec.t
@@ -188,9 +188,11 @@ test_expect_success 'signal forwarding works' '
 	cat >test_signal.sh <<-EOF &&
 	#!/bin/bash
 	sig=\${1-INT}
+	mkfifo input.fifo
 	stdbuf --output=L \
-	    flux exec -n sh -c "echo hi; sleep 100" >sleepready.out 2>&1 &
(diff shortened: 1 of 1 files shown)


grondo:
Thanks! Added MWP

PR #5386, opened 2023-08-12
job-manager: add flux_jobtap_jobspec_update_pack(3)

grondo:
This PR adds a convenience function for jobtap plugins that want to post a `jobspec-update` event.

The benefit of this function over using `flux_jobtap_event_post_pack(3)` directly is two-fold:
 - The `jobspec-update` event is delayed until after the callback for the current plugin returns, avoiding an update of the jobspec while the callback may be holding some pointers to it as a result of `flux_plugin_arg_unpack()`.
 - It allows some validation of the keys being updated (e.g. RFC 21 specifies they must start with `attributes.`)

The addition of a basic test here also exercises the processing of the `jobspec-update` event in the job manager, which doesn't seem to be tested at all presently.

### DIFF
--- a/src/modules/job-manager/jobtap.c
+++ b/src/modules/job-manager/jobtap.c
@@ -74,7 +74,6 @@ struct jobtap {
     zhashx_t *plugins_byuuid;
     zlistx_t *jobstack;
     json_t *jobspec_update;
-    char last_error [128];
     bool configured;
 };
 
--- a/src/modules/job-manager/jobtap.c
+++ b/src/modules/job-manager/jobtap.c
@@ -73,6 +73,7 @@ struct jobtap {
     zlistx_t *plugins;
     zhashx_t *plugins_byuuid;
     zlistx_t *jobstack;
+    json_t *jobspec_update;
     char last_error [128];
     bool configured;
 };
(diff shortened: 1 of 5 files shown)


src/modules/job-manager/jobtap.c
```diff
@@ -2078,6 +2111,54 @@ int flux_jobtap_event_post_pack (flux_plugin_t *p,
...
+        errno = EINVAL;
+        return -1;
+    }
+    va_start (ap, fmt);
+    o = json_vpack_ex (&error, 0, fmt, ap);
+    va_end (ap);
+    if (!validate_jobspec_updates (o)) {
+        errno = EINVAL;
+        goto out;
+    }
```
garlick, line 2147: Handle `json_vpack_ex()` failure?

garlick (approved):
Sounds good.  Just noticed one possible unchecked error.

src/modules/job-manager/jobtap.c
grondo: Oh, good catch. Forgot to add that back after a refactor 🤦‍♂️

grondo:
Ok, fixed the issue caught by @garlick above. Also found that the `last_error` member of `struct jobtap` isn't actually used anywhere, so added a commit to remove it.

t/Makefile.am
```diff
@@ -482,6 +482,7 @@ check_LTLIBRARIES = \
 	job-manager/plugins/create-reject.la \
 	job-manager/plugins/perilog-test.la \
 	job-manager/plugins/config.la \
+	job-manager/plugins/jobspec-update.la \
```
chu11, line 485: just double checking, is last sentence of commit message finished?  looks like maybe didn't finish sentence?

also typo, "jobspec-updte" -> "jobspec->update"

grondo: Good catch. I meant to drop that sentence fragment.

grondo:
Going to set MWP on this one, Thanks!

PR #5388, opened 2023-08-14
testsuite: relax systemctl output parsing

garlick:
Problem: the t2412-sdexec-perilog.t test script expects exactly one space between output columns but this fails in some environments.

The failing environment is a Ubuntu 22.04 LTS system running systemd 249.11-0ubuntu3.9.

Allow any number of spaces between the columns.


chu11 (approved):
LGTM

PR #5390, opened 2023-08-15
do not search for module and connector DSOs recursively

### LINKED ISSUE #5387: libflux: each directory in FLUX_CONNECTOR_PATH is searched recursively
Problem: we have "extra" connector loading code in [handle.c](https://github.com/flux-framework/flux-core/blob/d403e533ee222a38b04822f1227701236568dc9f/src/common/libflux/handle.c#L222) to find in-tree connector DSOs that are built in the ir own directory.

As a cleanup we could just build the connectors in `src/connectors` with one top level `Makefile.am` like we now do in `src/modules` (#5085) and then just set the in-tree connector path to `src/connectors/.lib`.   The recursive glob search could then be dropped.

garlick:
Problem: connector DSOs are each built in a separate directory, resulting in unnecessary recursive directory searching when loading them from the source tree.

Switch to a single Makefile.am that puts the build products in one directory, then change the definition of INTREE_CONNECTOR_PATH to `$(abs_top_builddir)/src/connectors/.libs` instead of the directory containing it.

Fixes #5387


grondo (approved):
LGTM. Nice to remove some recursive Makefiles!

Can the recursive directory search also be removed from `connector_find()` in libflux?

Not really necessary I suppose since the connector modules are now always installed into their own directory...

garlick:
Yeah I was thinking the recursive search could be removed too but the absence of a drop-in replacement for `dirwalk_find()` (from libutil/dirwalk.c) that is not recursive was not available?  (I was slightly surprised to see that the broker still used `dirwalk_find()` after the changes I made over there and wasn't remembering why but that must be it?)

Should we add a DIRWALK_NORECURSE option to `dirwalk_find()` and use it in both those situations?

src/common/libutil/dirwalk.c
```diff
@@ -233,7 +233,8 @@ static int dirwalk_traverse (dirwalk_t *d, dirwalk_filter_f fn, void *arg)
                 dirwalk_stop (d, errno);
             continue;
         }
-        if (S_ISDIR (d->current->sb.st_mode)) {
+        if (S_ISDIR (d->current->sb.st_mode)
+            && !(d->flags & DIRWALK_NORECURSE)) {
```
grondo, line 237: There's no use case for this, so I hesitate to even mention it, but will this break the use of `DIRWALK_FIND_DIR` with `DIRWALK_NORECURSE`? We could just issue an error if both flags are used together, or I think if you just `dirwalk_visit()` the directory instead of `dirwalk_traverse()` when `DIRWALK_FIND_DIR` is requested, that might fix it? (after 2 minutes of examining code, so I might have got it wrong)

grondo: Oh, thanks, that should have been obvious sorry!

PR #5394, opened 2023-08-16
flux-top: fix title when connected to instances that are not jobs

grondo:
This PR fixes the window title for `flux top` when connected to an instance that is not a job, e.g. a system instance or test instance started with `flux start`, etc.


chu11 (approved):
LGTM!

grondo:
Thanks! Setting MWP.

Issue #4792, opened 2022-12-01
idea: stream status updates from flux-resource or other tools


chu11:
Partly from a hallway conversation I had with @morrone.

I added a `whatsup` option called `--monitor` a long time ago.  The idea is you run

```
> whatsup --monitor
```

at the end of the day, and it'll output things like "node123 (10/22/22 9:00PM): down" (can't remember the exact format, but that's the basics) when things go down/up during the night.   You come in in the morning and you get a nice mini status in your terminal for what happened when you were gone.

Similar option could be useful with `flux-resource` or some other tools.

BUT, the additional benefit is that if we add this, the events stream that implements this underneath could then also be used as a more friendly events streaming service for #4569.

garlick:
We do have a resource eventlog in the KVS that can be watched, e.g. in raw form:
```console
$ sudo flux kvs eventlog get -w resource.eventlog
1669817412.474601 resource-init {"restart":true,"drain":{},"online":"","exclude":"0"}
1669817412.476516 resource-define {"method":"configuration"}
1669817414.620857 online {"idset":"0"}
1669914360.522472 online {"idset":"1"}
1669914360.672504 online {"idset":"2"}
1669914360.986879 online {"idset":"3-5"}
1669914361.136419 online {"idset":"6"}
1669914417.560495 offline {"idset":"6"}
1669914538.094525 drain {"idset":"7","reason":"testing drain","overwrite":0}
```

grondo:
This has nothing to do with #4569 though, since that issue deals with job events. Were you thinking a utility or service that would aggregate all known eventlogs into a single event stream for a consumer? (would need to be instance owner only)

chu11:
> Were you thinking a utility or service that would aggregate all known eventlogs into a single event stream for a consumer?

Ahh, I missed that #4569 was job events specific.  @morrone and I were talking about the potential of other event streams as well.  Although it wasn't discussed aggregating them all into one, we were discussing just the general availability of them.

PR #5395, opened 2023-08-16
flux-resource: support new "watch" subcommand

chu11:
Built on top of #5389.  Per a random idea I once had in #4792, support a `flux resource watch` command that watches and outputs resource changes as they occur by watching `resource.eventlog`.  The idea is to type it in your terminal when you leave work, and see what changed in your terminal when you get back in the next day.  Sample output:

```
2023-08-16T03:51:53 offline fake3
2023-08-16T03:51:53 drain   fake2
```

This was largely an exercise to make sure the recently added KVSWatch support in the Python kvs module is sorta useful.  And for the most part it works as intended.  Admittedly this is sort of just "neat" rather than a "must have" or "important".  But went ahead and add tests/documentation just in case people like it.

Some thoughts:

- should `job.EventLogEvent` be moved outside of `job/event.py` to a more general location?  I didn't do that here, but for consideration.

- I only support the options `--all` and `--ranks` for now.  An infinite number of output possibilities exist, but I think this is ok for now.

### DIFF
--- a/doc/man1/flux-resource.rst
+++ b/doc/man1/flux-resource.rst
@@ -21,6 +21,8 @@ SYNOPSIS
 
 | **flux** **resource** **reload** [-f] [--xml] *path*
 
+| **flux** **resource** **watch** [-a] [-r]
+
 DESCRIPTION
 ===========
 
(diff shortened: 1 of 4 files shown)


grondo:
Haven't had a chance to check this out, but FWIW I like the idea. Good use case for KVS watch too IMO. Nice.

chu11:
re-based, updating the changes to `flux-resource(1)` per the new formatting.

chu11:
rebased on new lineage, #5627, #5389

chu11:
rebased, now that #5389 is in

chu11:
was rebasing some old PRs, due to 42852ea1516f148ded5ec5e87e5af55a57c6e5a8, this PR / feature no longer works / makes sense.

This was a "neat" feature, but mostly was done as a way to use the new kvs async watcher.  If we really want this feature in the future, we'd have to do it an alternate way.

PR #5396, opened 2023-08-16
matrix: add arm64 install-only builder

trws:
This is a step-1 to supporting arm64 a bit better in CI.  If this is
fast enough, maybe we can get some occasional smoke tests on top of it
or something but it should make getting dev containers going easier at
the least.

### DIFF
--- a/src/test/generate-matrix.py
+++ b/src/test/generate-matrix.py
@@ -210,6 +210,7 @@ matrix.add_build(
     env=dict(
         TEST_INSTALL="t",
     ),
+    platform="linux/amd64",
     docker_tag=True,
 )
 
--- a/src/test/generate-matrix.py
+++ b/src/test/generate-matrix.py
@@ -149,6 +149,16 @@ matrix.add_build(
     docker_tag=True,
 )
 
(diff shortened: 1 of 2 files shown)


src/test/generate-matrix.py
```diff
@@ -210,6 +210,7 @@ def __str__(self):
     env=dict(
         TEST_INSTALL="t",
     ),
+    platform="linux/amd64",
```
grondo, line 213: Is this a stray change? If it is supposed to be part of this commit it might be useful to explain why in the commit message (it isn't clear to me on first view anyway)

grondo (approved):
LGTM. One question inline about a possible unrelated change.
I guess we'll see if the manifest update works when we merge this :-)

trws:
I can add it to the commit message, having the platform set forces the name to switch from "bookworm" to "bookworm-amd64" so we don't have the conflict with the manifest on :bookworm.

grondo:
> having the platform set forces the name to switch from "bookworm" to "bookworm-amd64" so we don't have the conflict with the manifest on :bookworm.

Aha! :+1:

PR #5401, opened 2023-08-17
testsuite: handle job signal race in more tests

### LINKED ISSUE #5399: testsuite: t2260-job-list needs signal race workaround
Same fix as #5210  / #5302

chu11:
Problem: In commit

c851ecbb34d7a741030c25fd12a38e0c11f84b4c

a job signal race was worked around in t2800-jobs-cmd.t.  The same fix should have been applied to t2260-job-list.t but was not.

Solution: Carry over the same fixes to t2260-job-list.t.  Also add a small comment in t2800-jobs-cmd.t for why the signal race workaround isn't needed in one location.

Fixes #5399

### DIFF
--- a/t/t2260-job-list.t
+++ b/t/t2260-job-list.t
@@ -24,6 +24,15 @@ wait_jobid_state() {
 	flux job list-ids --wait-state=$2 $1 > /dev/null
 }
 
+test_expect_success 'create helper job submission script' '
+	cat >sleepinf.sh <<-EOT &&
+	#!/bin/sh
+	echo "job started"
+	sleep inf
+	EOT
+	chmod +x sleepinf.sh
+'
+
 # submit a whole bunch of jobs for job list testing
 #
 # - the first loop of job submissions are intended to have some jobs run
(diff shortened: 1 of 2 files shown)


t/t2260-job-list.t
```diff
@@ -24,6 +24,15 @@ wait_jobid_state() {
 	flux job list-ids --wait-state=$2 $1 > /dev/null
 }
 
+test_expect_success 'create helper job submission script' '
+	cat >sleepinf.sh <<-EOT &&
+	#!/bin/sh
+	echo "job started"
+	sleep inf
```
grondo, line 31: Not sure if it is worth fixing here if there's no demonstrated problem, but in researching https://github.com/flux-framework/flux-core/issues/5382 I found that there can be a race if a signal is delivered to a script like this between the echo and the start of the sleep. In this case the signal was ignored and the test timed out.

See the fix in https://github.com/flux-framework/flux-core/commit/50bb62c42a4b6319ecbcdf64e6745ca3d1d96976 which uses awk and blocks on a FIFO instead of sleeping so there are no extra processes involved, thus avoiding the potential race.

grondo (approved):
This LGTM. Added a comment about a potential race which was triggered with the same kind of test in `t5000-exec.t`. However, the existing w/a in `t2800-jobs-cmd.t` doesn't seem to cause  a problem currently (I could reproduce the `t5000-exec.t` race easily on my pi cluster) so switching the alternate solution is optional IMO.

t/t2260-job-list.t
chu11: Hmmmm, for this case, I'm not sure its been seen.  I guess there'd be a hang if the signal was ignored?

But thinking about this ... does this race technically exist basically for every job that every user will ever run?  like.

```bash
#!/bin/bash
flux submit -N4 foo1.sh
flux submit -N4 foo2.sh
...
```

I dunno if this is generally solvable??

grondo: Signal delivery is generally known to be unreliable. There's no fixing it. Only an issue here because the testsuite depends on a certain behavior. Since we're not seeing the hang in these tests, I'm fine leaving it as is until/if there's a real problem. I was just mentioning it as an fyi

PR #5402, opened 2023-08-17
extend ci-checks timeout for arm64 build

trws:
It got really, really close, but didn't quite finish in time on the master build.  Give it a bit more of a buffer before canceling.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -210,6 +210,7 @@ jobs:
       if: >
         (startsWith(github.ref, 'refs/tags/') || github.ref == 'refs/heads/master')
       run: |
+        echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
         docker manifest create fluxrm/flux-core:bookworm fluxrm/flux-core:bookworm-amd64 fluxrm/flux-core:bookworm-386 fluxrm/flux-core:bookworm-arm64
         docker manifest push fluxrm/flux-core:bookworm
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -101,7 +101,7 @@ jobs:
       DOCKER_USERNAME: travisflux
       DOCKER_PASSWORD: ${{ secrets.DOCKER_HUB_TRAVISFLUX_TOKEN }}
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-    timeout-minutes: 60
+    timeout-minutes: 90
     strategy:
       matrix: ${{fromJson(needs.generate-matrix.outputs.matrix)}}
       fail-fast: false


trws:
Looks like there's one other thing to fix here @grondo.  I'm getting this on the manifest step (the re-try of the arm64 build finished in 45 minutes):
`2023-08-17T22:02:56.4362144Z Get "https://registry-1.docker.io/v2/fluxrm/flux-core/manifests/bookworm-amd64": unauthorized: access token has insufficient scopes`

I'm guessing that means the password for travisflux is a token, and it doesn't have the permissions to pull the existing manifests or something?  I don't have the credentials to manage tokens for travisflux, could you see if you can modify the perms on the token to let it do this? I tested the exact commands from the stage with my authentication locally, and it successfully pushed the multiarch manifest, so that should have that done (and we can re-try the stage if the auth fix doesn't require changing tokens).

grondo:
The travisflux token has read,write,delete scope and I don't see any further scopes to add. It is also an "owner" of the fluxrm repo, so I'm not sure why we're getting this error (and for a `Get`?)

There are very few options when creating a dockerhub access token, I don't see anything other than selecting from read, write, and delete access...

grondo (approved):
Looks fine to me. I also logged in locally to docker as user `travisflux` and at lest the `docker manifest` command worked, so I'm not sure what is wrong in the action. Maybe it will just work this time?

trws:
Thanks for looking. Maybe it will just work now, or maybe it’s a cross-node race condition? If it isn’t reliable I can force it to use digests to avoid needing to do the get part, but was hoping to avoid having to set up the extra artifacts.
---
On August 17, 2023 at 3:38:58 PM PDT, Mark Grondona ***@***.***> wrote:

The travisflux token has read,write,delete scope and I don't see any further scopes to add. It is also an "owner" of the fluxrm repo, so I'm not sure why we're getting this error (and for a Get?)

There are very few options when creating a dockerhub access token, I don't see anything other than selecting from read, write, and delete access...

trws:
It ended up popping the same error on Al's PR, so I'm adding in explicit login, probably don't have the same session in the new action context.  Setting MWP on the theory this is good with you and we'll want to see what happens _after_ it merges more than before anyway.

trws:
Well crud. I'm going to sleep on it for now, but the manifest command is still giving the same strange error. Maybe I'll have to use the buildx method or something else. All the builds and individual pushes are working fine, but we aren't continually updating the main :bookworm container right now so I'll try to figure it out asap.

grondo:
Maybe there's a way to use the [tmate](https://github.com/mxschmitt/action-tmate) action to get interactive access to the container, thereby getting more information about why `docker manifest` is failing. It would work a lot better in a PR though.. maybe as a test you could create a PR that enables the arm64 build  along with a test action that attempts `docker manifest` after all the bookworm builds are done? If that reproduces the problem then this could be debugged in a PR...

trws:
That's a good idea, I can definitely change the condition for a PR and anything that would give some more information would certainly help.  It's especially perplexing because now both you and I have tested this with access tokens, so even though there _is_ weirdness with tokens, it clearly works in other environments.  The only thing coming to mind now is if somehow the registry configuration isn't the same between docker-deploy and this, but we don't configure any of that in deploy and I don't see any env var for it or anything...  Need more info for sure.

PR #5405, opened 2023-08-21
job-manager: fix duration limit check for jobs with unset/unlimited duration

grondo:
This PR fixes an outdated design in the `limit-duration` job manager plugin which explicitly allows jobs with unlimited duration through, even if there is an applicable queue or general limit. This was probably the case so that a default could later be applied to the job, but since default are now applied in the job frobnicator at ingest time, this is no longer necessary and allows jobs with unlimited duration to submitted if there is a limit but no duration default.


garlick (approved):
LGTM

PR #5406, opened 2023-08-22
job-manager: prevent jobspec-update events after a job has resources

grondo:
Almost all jobspec updates don't make any sense after the job has been allocated resources and the job starts running, but as noted by @chu11, there is currently nothing the job manager that prevents a `jobspec-update` event in job state from modifying the redacted jobspec.

This PR makes it an error if a `jobspec-update` event occurs in `RUN`, `CLEANUP` or `INACTIVE` states (actually, any event is prevented if the job is inactive now by checking the `eventlog_inactive` flag)

There is also one commit that adds missing support for bumping a job in `SCHED` back to `PRIORITY` on a `jobspec-update` event so that the priority can be recalculated and the `alloc` request restarted, but now I'm wondering if that commit belongs here or in a future PR that adds support for a job manager update service.

### DIFF
--- a/src/modules/job-manager/event.c
+++ b/src/modules/job-manager/event.c
@@ -556,7 +556,9 @@ static int event_handle_jobspec_update (struct job *job, json_t *context)
     const char *path;
     json_t *val;
 
-    if (!job->jobspec_redacted)
+    if (!job->jobspec_redacted
+        || job->state == FLUX_JOB_STATE_RUN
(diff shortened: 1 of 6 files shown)


grondo:
I went ahead and removed the commit that didn't belong.

chu11 (approved):
LGTM.  Should we add a note/clarification/something to RFC21?

grondo:
Yes, that's a good idea

grondo:
Setting MWP.

PR #5407, opened 2023-08-23
job-list: misc cleanup

chu11:
just splicing out some random cleanup before the bigger job-list PR that's upcoming.

### DIFF
--- a/src/modules/job-list/job_data.c
+++ b/src/modules/job-list/job_data.c
@@ -266,13 +266,12 @@ static int parse_jobspec (struct job *job, const char *s, bool allow_nonfatal)
     json_error_t error;
     json_t *jobspec_job = NULL;
     json_t *tasks;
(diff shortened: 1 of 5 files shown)


src/modules/job-list/job_data.c
```diff
@@ -266,13 +266,12 @@ static int parse_jobspec (struct job *job, const char *s, bool allow_nonfatal)
     json_error_t error;
```
garlick, line 266: In the commit message for "job-list: do not return error when nonfatal set" there's a duplicate jobspec word: "illegal json jobspec jobspec"

garlick (approved):
Looks OK To me!

src/modules/job-list/job_data.c
chu11: oops! thanks.  i'll fix up then set MWP.

Issue #4844, opened 2022-12-16
allow job duration of pending jobs to be updated


garlick:
This is a tracking issue for allowing jobspec duration to be updated on a job after it is submitted, but before it is allocated resources.

Note that the extension of expiration time on running jobs is a separate issue.  See #4175.

The duration is part of jobspec, thus changing it should employ the [jobspec-update](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_21.html#jobspec-update-event) event.  The change should only be permitted if requested by the instance owner or the job owner, and must respect the duration limit (general or for the selected queue) if configured.  If an alloc  request has already been sent to the scheduler, it needs to be canceled and resent with the new jobspec.

- [ ] need job manager service method for updating duration of pending job
- [ ] need front end tool for requesting duration update
- [ ] job-list should process the `jobspec-update` event so that the new duration can be returned to users

grondo:
This is now complete.

PR #5408, opened 2023-08-23
job-list: support jobspec-update event

chu11:
Per #4404, built on top of #5407

There were some tricky parts that took me awhile to refactor through, but overall, the commit series shouldn't look too strange.

I didn't list this as WIP as I think this could go in as is.  But perhaps we want to wait until work on #4844 is further along so we can test in tandem??

### DIFF
--- a/src/modules/job-list/job_data.c
+++ b/src/modules/job-list/job_data.c
@@ -41,6 +41,7 @@ void job_destroy (void *data)
         json_decref (job->jobspec_tasks);
         json_decref (job->R);
         json_decref (job->exception_context);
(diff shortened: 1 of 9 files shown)


grondo:
Actually might be nice to have this merged so the #4844 implementation could be more easily tested (afaik, there's no way to fetch the redacted jobspec via an RPC at this point). So, I'll try to review this and predecessor sooner rather than later.

grondo (approved):
This looks pretty good to me! Nothing stood out and I tested it along with my WIP job update service and everything seemed to just work. :+1:

Issue #4698, opened 2022-10-18
job-list: support retrieving job project name?


chu11:
https://github.com/flux-framework/flux-accounting/issues/293

will need to handle `jobspec-update` event? #4404

Issue #4697, opened 2022-10-18
job-list: support retrieving job bank?


chu11:
https://github.com/flux-framework/flux-accounting/issues/272

will need to handle `jobspec-update` event?  #4404

PR #5409, opened 2023-08-24
WIP: add job update service and new job-update(1) command

grondo:
This is a WIP, but posting early while working on tests and documentation in case there are comments on the overall design or other high level aspects of this proposal.

This PR adds a new job update service to the job manger. A new `job-manager.update` RPC accepts requests for job updates with a payload including the target jobid, and an `updates` object, which has the same form as the `jobspec-update` context of [RFC 21](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_21.html), except that period-delimited keys are not required to begin with `attributes.`.

In order to test if an update is allowed for a given key, the update service will attempt to call a `job.update.KEY` jobtap callback. If no plugin is registered for that callback topic, or if the plugin callback returns -1, then the update is rejected. This allows us to gradually and carefully enable updates, as well as allowing out-of-tree projects to allow updates to keys for which they control (e.g. flux-accounting can allow updates to `bank` and `project`).

Once all keys in a payload have been "allowed" by jobtap plugin callbacks (if multiple keys are present, then they all must be allowed or they all fail), then the update service copies the current jobspec and applies the update, then validates the update by passing the updated jobspec to the `job.validate` plugin stack. This allows the update in full to be validated before it is applied.

Plugins can, however, notify the update service that the plugin already considers the update validated by setting a `validated` flag in the output arguments. This can be used to bypass further validation when it is unnecessary or if a user is being allowed to bypass limits for example. If there are multiple keys in the update request, then all keys must have the `validated` flag set to bypass validation.

Since the first key update we want to allow is `attributes.system.duration`, an `update-duration` builtin plugin is also added to the job manager. This plugin allows users to update duration of pending jobs with validation (so that limits are checked), and without validation by default for instance owner (allowing the instance owner to increase the duration beyond configured limits). A plugin config parameter allows this instance owner capability to be disabled (mostly for testing). Having the duration update be a plugin instead of builtin to the update service also allows updates to be disabled completely by removing the plugin.

Finally, a new `flux-update(1)` command is added for sending update requests to the update service. The command takes a jobid and one or more key=value pairs and constructs an update request. Keys that do not begin with standard top-level keys of jobspec are assumed to be under `attributes.system.`, except for `name` which is an alias for `attributes.system.job.name` (yes, the `flux-update` command is designed for general updates at this point even if the update service does not support them).

Each key can have special value handling by adding a specially named method in the internal `JobspecUpdates` class. Currently, `duration` has special handling of the value to allow values of the form `[+-]FSD`. If a duration adjustment is requested via `+` or `-`, the utility downloads the jobspec and eventlog for the job and gets the current duration (with updates applied) and adjusts. Thus the following all should work:

 - `flux update JOBID duration=1h` -- set duration to 1 hour
 - `flux update JOBID duration=+10m` -- add 10 minutes to current duration
 - `flux update JOBID duration=-10m` -- subtract 10 minutes from current duration
 - `flux update JOBID duration=inf` -- set duration to unlimited

Since the `job.validate` callback is made before applying updates, any duration update request that exceeds limits is rejected with an appropriate error message, e.g.:

```console
$ flux update ƒGT7B136w duration=+1h
flux-update: ERROR: requested duration exceeds policy limit of 1h
```

After an update request succeeds, the `jobspec-update` event is posted to the eventlog. This event now bumps jobs in SCHED back to PRIORITY. This should cancel any outstanding `alloc` requests to the scheduler, and when the `alloc` request is made again, the scheduler should get the new duration as part of the redacted jobspec.


grondo:
Rebased now that #5408 is merged which will allow easier testing and experimentation

chu11:
excellent timing, I had started working on #4697 and #4698 and blanked on the fact that a `flux kvs eventlog append` doesn't go through the journal, and thus `job-list` never sees the update in some tests.  Can give this a whirl.

src/modules/job-manager/update.c
```diff
@@ -0,0 +1,232 @@
...
+    FLUX_MSGHANDLER_TABLE_END,
+};
+
+struct update *update_ctx_create (struct job_manager *ctx)
+{
+    struct update *update;
+
+    if (!(update = calloc (1, sizeof (*update))))
+        return NULL;
+    update->ctx = ctx;
```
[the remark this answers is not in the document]
grondo: I marked this as a false positive, but then thought maybe there's some obvious problem I'm missing here?

src/modules/job-manager/event.c
```diff
@@ -675,6 +675,13 @@ int event_job_update (struct job *job, json_t *event)
     else if (streq (name, "jobspec-update")) {
         if (event_handle_jobspec_update (job, context) < 0)
             goto inval;
+        /*  Transition a job in SCHED state back to PRIORITY to trigger
```
chu11, line 671: commit message typo "PRIOITY"

src/modules/job-manager/plugins/update-duration.c
```diff
@@ -0,0 +1,90 @@
...
+ *  owner-allow-any=0.
+ */
+static int owner_allow_any = 1;
+
+static int duration_update_cb (flux_plugin_t *p,
+                               const char *topic,
+                               flux_plugin_arg_t *args,
+                               void *arg)
+{
+    flux_jobid_t id;
```
chu11, line 38: i don't think id is used in this function for anything

chu11 (commented):
just starting to play with things, two small nits i noticed

chu11:
woo hoo, initial tests seem to work

```
                                                                                                                                            
test_expect_success 'support jobspec updates of project and bank' '                                                                         
        flux jobtap load --remove=all ${PLUGINPATH}/project-bank-validate.so                                                                
'                                                                                                                                           
<snip>
# initially put job on hold, jobspec-updates don't matter after the job is running                                                          
test_expect_success 'flux job list outputs project and bank if one set' '                                                                   
        jobid=`flux submit --urgency=hold /bin/true | flux job id` &&                                                                       
        echo $jobid > jobprojectbank2.id &&                                                                                                 
        flux update $jobid project=foo &&                                                                                                   
        flux update $jobid bank=bar &&                                                                                                      
        flux job urgency $jobid default &&                                                                                                  
        wait_jobid_state $jobid inactive &&                                                                                                 
        flux job list -s inactive | grep $jobid | jq -e ".project == \"foo\"" &&                                                            
        flux job list -s inactive | grep $jobid | jq -e ".bank == \"bar\""                                                                  
'                                                                                
```

src/modules/job-manager/plugins/update-duration.c
grondo: Oh, yeah, good catch. Whatever was using the jobid got eventually moved out of that callback, so no longer necessary.

chu11:
seeing:

```
Aug 24 19:21:46.697069 job-manager.err[0]: queue_started: job foHZJWpX invalid queue: val
Aug 24 19:21:46.697141 job-manager.err[0]: queue_started: job foHZJWpX invalid queue: val
Aug 24 19:21:46.697202 job-manager.err[0]: queue_started: job foHZJWpX invalid queue: val
Aug 24 19:21:46.697271 job-manager.err[0]: queue_started: job foHZJWpX invalid queue: val
Aug 24 19:21:46.697330 job-manager.err[0]: queue_started: job foHZJWpX invalid queue: val
Aug 24 19:21:46.697388 job-manager.err[0]: queue_started: job foHZJWpX invalid queue: val
Aug 24 19:21:46.697444 job-manager.err[0]: queue_started: job foHZJWpX invalid queue: val
```

spin when I run t2260-job-list.t.  

 I haven't figured out cause yet, but I'm thinking something in your branch conflicts with my "jobspec-update-job-list" jobtap plugin.  Perhaps the jobspec update data wasn't validated before, but needs to be validated now, and somehow ends up in some error loop?

chu11:
On nothing more than a hunch I reverted the `job-manager: move jobs from SCHED->PRIORITY on jobspec-update` and the spinning error no longer happens.  I'm guessing there's some corner case in there regarding changes to queues.  The error message saying the queue is "val" is perplexing though.

chu11:
ok, well this is part of the issue

```
    if (job->jobspec_redacted) {                                                                                                            
        /* unit tests assume empty jobspec legal, so all fields                                                                             
         * optional                                                                                                                         
         */                                                                                                                                 
        if (json_unpack (job->jobspec_redacted,                                                                                             
                         "{s?{s?{s?s}}}",                                                                                                   
                         "attributes",                                                                                                      
                           "system",                                                                                                        
                             "queue", &job->queue) < 0) {                                                                                   
            errno = EINVAL;                                                                                                                 
            return -1;                                                                                                                      
        }                                                                                                                                   
    }                                           
```

if the value has changed in the redacted jobspec, then queue is now pointing to whatever junk its old pointer is pointing to.  So that is the reason for that mystery.

grondo:
> if the value has changed in the redacted jobspec, then queue is now pointing to whatever junk its old pointer is pointing to. So that is the reason for that mystery.

Good catch, but updating the queue is not supported yet, maybe we should have changed something else incidental in those tests (maybe job name?). Though I guess it wouldn't hurt to fix this issue in this PR since it seems to somehow introduce a problem...

grondo:
Well I guess an explicit call to update the queue name any time the jobspec_redacted is updated might work, but blech. Also is bumping a job back to PRIORITY enough to ensure moving from a stopped to started queue works and vice versa? (I guess the `job.update.attributes.system.queue` plugin callback would deny updating to a queue that is currently stopped... somehow. Not too pretty though... :confused:

grondo:
Maybe a PR to change the job-list test attribute to something other than `queue` would be best for now? And open an issue to capture what we learned about changing the queue of a pending job?

grondo:
Simple solution for now? I can add that here if it works:
```diff
diff --git a/src/modules/job-manager/event.c b/src/modules/job-manager/event.c
index 7f8c70d76..17dc8744b 100644
--- a/src/modules/job-manager/event.c
+++ b/src/modules/job-manager/event.c
@@ -555,7 +555,8 @@ static int event_handle_jobspec_update (struct job *job, json_t *context)
     if (!job->jobspec_redacted
         || job->state == FLUX_JOB_STATE_RUN
         || job->state == FLUX_JOB_STATE_CLEANUP
-        || jobspec_apply_updates (job->jobspec_redacted, context) < 0)
+        || jobspec_apply_updates (job->jobspec_redacted, context) < 0
+        || job_update_queue (job) < 0)
         return -1;
     return 0;
 }
diff --git a/src/modules/job-manager/job.c b/src/modules/job-manager/job.c
index cbc94101c..737963611 100644
--- a/src/modules/job-manager/job.c
+++ b/src/modules/job-manager/job.c
@@ -172,6 +172,11 @@ static int jobspec_redacted_parse_queue (struct job *job)
     return 0;
 }
 
+int job_update_queue (struct job *job)
+{
+    return jobspec_redacted_parse_queue (job);
+}
+
 struct job *job_create_from_eventlog (flux_jobid_t id,
                                       const char *eventlog,
                                       const char *jobspec,
diff --git a/src/modules/job-manager/job.h b/src/modules/job-manager/job.h
index 4fcbccf94..88f760c25 100644
--- a/src/modules/job-manager/job.h
+++ b/src/modules/job-manager/job.h
@@ -127,6 +127,10 @@ int job_event_peek (struct job *job, int *flagsp, json_t **entryp);
 bool job_event_is_queued (struct job *job, const char *name);
 const char *job_event_queue_print (struct job *job, char *buf, int size);
 
+/*  Update queue in case of jobspec update
+ */
+int job_update_queue (struct job *job);
+
 /*  Validate updates as valid RFC 21 jobspec-update event context:
  */
 bool validate_jobspec_updates (json_t *updates);
```

chu11:
hmmmm, my spiritually similar hack still hangs the test, so perhaps there's another issue somewhere.

```
-int jobspec_apply_updates (json_t *jobspec, json_t *updates)
+int jobspec_apply_updates (json_t *jobspec, json_t *updates, struct job *job)
 {
     const char *path;
     json_t *val;
@@ -612,6 +612,7 @@ int jobspec_apply_updates (json_t *jobspec, json_t *updates)
         if (jpath_set (jobspec, path, val) < 0)
             return -1;
     }
+    jobspec_redacted_parse_queue (job);
     return 0;
 }

```

The change to queue in the job-list tests was specific to test that queue stats were updated correctly in `job-list`.  Obviously that PR went in before specific updates were allowed/not allowed.

could simply remove that part of the test temporarily?

grondo:
BTW, @cmoussa1 should probably comment here if this scheme will work for flux-accounting. The idea would be that the accounting priority plugin could add handlers for `job.update.attributes.system.bank` and `job.update.attributes.system.project` to permit updates of these attributes. These callback should check if the new bank and/or project are valid before allowing the update.

It occurs to me that one new caveat here is that he `job.validate` callback should now be idempotent, i.e. it should only validate the job, not make any updates or changes to internal information about the job. This isn't the case currently, so there may be some `job.validate` callbacks that need to be split up to do some work in `job.new` instead.

In the case of a `bank` or `project` update, the job will be kicked back to `PRIORITY` after an update, so the bank and project for a job need to get updated at that callback somehow, before the priority plugin does its calculation of the new priority. I think this also might be a change in how the priority works now, so we may need to coordinate changes in the accounting plugin before "enabling" updates to those attributes.

grondo:
> hmmmm, my spiritually similar hack still hangs the test, so perhaps there's another issue somewhere.

Yeah, I didn't test mine and it crashed the broker. Kind of like encapsulating the update in a single function as you've done so let me try that. It would be good to figure this out now in case the approach here is untenable for a queue update...

grondo:
Hm, maybe the problem is that the job-list test jobspec-update plugin is updating the jobspec on _every_ callback, since the jobspec-update event kicks the job back to PRIORITY, the job will call the callback again, then get sent back to PRIORITY again, and so on...

grondo:
This helps:

```diff
diff --git a/t/job-manager/plugins/jobspec-update-job-list.c b/t/job-manager/plu
gins/jobspec-update-job-list.c
index 43440fe9b..11bcb52c5 100644
--- a/t/job-manager/plugins/jobspec-update-job-list.c
+++ b/t/job-manager/plugins/jobspec-update-job-list.c
@@ -25,6 +25,9 @@ static int validate_cb (flux_plugin_t *p,
                         flux_plugin_arg_t *args,
                         void *data)
 {
+    static bool updated = false;
+    if (updated)
+        return 0;
     if (flux_jobtap_jobspec_update_pack (p,
                                          "{s:f}",
                                          "attributes.system.duration",
@@ -35,6 +38,7 @@ static int validate_cb (flux_plugin_t *p,
                                      "update failure");
         return -1;
     }
+    updated = true;
     return 0;
 }
 
@@ -43,6 +47,9 @@ static int depend_cb (flux_plugin_t *p,
                       flux_plugin_arg_t *args,
                       void *data)
 {
+    static bool updated = false;
+    if (updated)
+        return 0;
     if (flux_jobtap_jobspec_update_pack (p,
                                          "{s:s}",
                                          "attributes.system.job.name",
@@ -53,6 +60,7 @@ static int depend_cb (flux_plugin_t *p,
                                      "update failure");
         return -1;
     }
+    updated = true;
     return 0;
 }
 
@@ -61,6 +69,9 @@ static int sched_cb (flux_plugin_t *p,
                      flux_plugin_arg_t *args,
                      void *data)
 {
+    static bool updated = false;
+    if (updated)
+        return 0;
     if (flux_jobtap_jobspec_update_pack (p,
                                          "{s:s}",
                                          "attributes.system.queue",
@@ -71,6 +82,7 @@ static int sched_cb (flux_plugin_t *p,
                                      "update failure");
         return -1;
     }
+    updated = true;
     return 0;
 }
```

Might only be necessary for the `job.state.sched` callback though.

chu11:
@grondo ahh good catch, that's sorta obvious now that the sched->priority commit is in there.  Applied the logical fix to (only in the sched callback) and the hang is gone.

chu11:
Does this fix point to a potential need to not call the jobtap callback again after a jobspec update?  Wouldn't be the most obvious thing to do to a casual jobtap plugin programmer.

grondo:
I'll just incorporate that into this PR (before the SCHED->PRIORITY commit)

grondo:
> Does this fix point to a potential need to not call the jobtap callback again after a jobspec update? Wouldn't be the most obvious thing to do to a casual jobtap plugin programmer.

I think the jobtap plugin stack _must_  be called again when the job goes back to SCHED or any other state or else the job may not be properly processed by plugins (and there's no way to "know" what the plugins are doing to avoid some plugins but not all, even if that were possible)

cmoussa1:
> BTW, @cmoussa1 should probably comment here if this scheme will work for flux-accounting. The idea would be that the accounting priority plugin could add handlers for `job.update.attributes.system.bank` and `job.update.attributes.system.project` to permit updates of these attributes. These callbacks should check if the new bank and/or project are valid before allowing the update.

This sounds reasonable to me at the moment! It would make sense for the priority plugin to support this kind of functionality; users can belong to multiple banks and projects, so they should be able to switch between them if need be.

> It occurs to me that one new caveat here is that he `job.validate` callback should now be idempotent, i.e. it should only validate the job, not make any updates or changes to internal information about the job. This isn't the case currently, so there may be some `job.validate` callbacks that need to be split up to do some work in `job.new` instead.

I think this makes sense so far. I could definitely be wrong but I think as it stands right now, the priority plugin does not make any updates or changes to internal information about the job (that would, of course, change with flux-framework/flux-accounting#294), right? Maybe I am misunderstanding your point here, sorry! Are you suggesting that this kind of functionality, where the priority plugin posts a jobspec-update for the default project name (or, eventually the default bank name) be moved to the callback for `job.new`?

> In the case of a `bank` or `project` update, the job will be kicked back to `PRIORITY` after an update, so the bank and project for a job need to get updated at that callback somehow, before the priority plugin does its calculation of the new priority. I think this also might be a change in how the priority works now, so we may need to coordinate changes in the accounting plugin before "enabling" updates to those attributes.

This is good to know. If this is the case, where a job goes back to `PRIORITY` after an update, then we would need to add a `flux_jobtap_jobspec_update_pack ()` call(s) to this callback to update the main eventlog for _any_ accounting attributes that could get updated, yes? If so, then I think these calls could just be placed before the priority for the job gets calculated, i.e before the callback calls `priority_calculation ()`? Just my immediate thoughts without any actual testing or anything. :P

grondo:
> Are you suggesting that this kind of functionality, where the priority plugin posts a jobspec-update for the default project name (or, eventually the default bank name) be moved to the callback for job.new?

I was thinking more about a plugin's internal state for a job -- the `job.validate` callback should not change this state based on the validation arguments, since `job.validate` could now be called with _proposed_ jobspec, and if some other validation fails, the updates might never be applied.

>  where a job goes back to PRIORITY after an update, then we would need to add a flux_jobtap_jobspec_update_pack () call(s) to this callback to update the main eventlog for any accounting attributes that could get updated, yes? If so, then I think these calls could just be placed before the priority for the job gets calculated, i.e before the callback calls priority_calculation ()

`jobspec-update` events only need to be emitted once. They would be emitted from the `job-manager.update` service after the updates are allowed and validated. The result would be that the attributes would be updated in the internal copy of jobspec in the job-manager _before_ the job is sent back to the `PRIORITY` state.

Once the job reaches the `PRIORITY` state, then the `job.state.priority` callback is called. So the accounting plugin `job.state.priority` callback should check if the `bank` or `project` has been updated, and if so update any _internal_ bank or project information for the job. (I can't remember if the accounting plugin has any of this state at the moment)

cmoussa1:
> I was thinking more about a plugin's internal state for a job -- the `job.validate` callback should not change this state based on the validation arguments, since `job.validate` could now be called with _proposed_ jobspec, and if some other validation fails, the updates might never be applied.

Ah, so `job.validate` should not create that internal `bank_info` struct that is associated with each job that contains things like active job counts, fairshare value, what queues that user/bank has access to, etc.? And instead, that should be pushed off to a later state, i.e `job.new`? Sorry if I am misunderstanding you. 

> `jobspec-update` events only need to be emitted once. They would be emitted from the `job-manager.update` service after the updates are allowed and validated. The result would be that the attributes would be updated in the internal copy of jobspec in the job-manager _before_ the job is sent back to the `PRIORITY` state.
>
> Once the job reaches the `PRIORITY` state, then the `job.state.priority` callback is called. So the accounting plugin `job.state.priority` callback should check if the `bank` or `project` has been updated, and if so update any _internal_ bank or project information for the job. (I can't remember if the accounting plugin has any of this state at the moment)

Ohhhh, okay. So `priority_cb ()` would need to add some logic to see if the bank name or project name has been updated and update that internal `bank_info` struct if so.

Yeah, I think that callback still unpacks arguments for the job, but doesn't check to see if it needs to be updated or anything. I believe the only case in which it updates the `bank_info` struct is if the job was held in PRIORITY state while the plugin was waiting for any user/bank information to be loaded from the flux-accounting DB. Perhaps this check could be generalized to handle the proposed case?

grondo:
Something to think about, but I wonder if a new jobtap callback `job.update` would be useful.

grondo:
Ok, I just pushed a commit that adds support for a new `job.update` callback that is issued any time there is an update to the job (currently, after a `jobspec-update` event has applied all updates, but before any state transition callback, i.e. `job.state.priority1). I actually realized this is necessary for plugins to reliably notified of job updates since jobs will only transition to a new state if they are in `SCHED`.

src/cmd/flux-update.py
```diff
@@ -0,0 +1,225 @@
...
+                )
+
+            arg = flux.util.parse_fsd(value[1:])
+            if value.startswith("-"):
+                result = duration - arg
+                if result <= 0.0:
+                    duration = flux.util.fsd(duration)
+                    raise ValueError(
+                        f"current duration for {self.jobid} ({duration})"
+                        f" cannot reduced by {value[1:]}"
```
chu11, line 101: "cannot be reduced"?

src/cmd/flux-update.py
```diff
@@ -0,0 +1,225 @@
...
+        and eventlog (to apply jobspec-updates) if necessary.
+        """
+        if self.jobspec is None:
+            lookup = flux.job.job_kvs_lookup(
+                self.flux_handle, jobid=self.jobid, keys=["jobspec", "eventlog"]
+            )
+            self.jobspec = flux.job.JobspecV1(**lookup["jobspec"])
+            self._apply_jobspec_updates(lookup["eventlog"])
+
+        return self.jobspec.getattr(key)
```
chu11, line 74: ugh, this seems annoying.  Perhaps we need a way to get the redacted jobspec?  Since `job-list` is already building its own version of this, perhaps some mechanism through there?  Perhaps could write up an issue and can be a TODO item later.

src/modules/job-manager/update.c
```diff
@@ -0,0 +1,232 @@
...
+ * Once all updates are validated by callbacks, then updates as applied
+ * to jobspec are validated by passing an updated jobspec to the
+ * `job.validate` jobtap plugin stack.
+ *
+ * Plugins may request that this validation step be skipped by setting
+ * the 'validated' flag to 1 in the FLUX_PLUGIN_OUT_ARGS of the
+ * `job.update.*` callback. The `job.validate` step will only be skipped
+ * all keys in an update have the validated flag set.
+ *
+ * If all steps above are successful, then a `jobspec-update` upddate
```
chu11, line 43: typo upddate

src/modules/job-manager/update.c
```diff
@@ -0,0 +1,232 @@
...
+        if (needs_validation)
+            validate = 1;
+    }
+    if (validate
+        && jobtap_validate_updates (ctx->jobtap,
+                                    job,
+                                    updates,
+                                    errp) < 0)
+            return -1;
+    /*  All updates have been allowed by pluginss and validated as a unit,
```
chu11, line 116: s/pluginss/plugins/

chu11 (commented):
just high level comments as i looked through, most are nits

src/modules/job-manager/update.c
grondo: sheesh did the typos ci action not run?

chu11: i thought the same thing, but i think these are all typos the typo checker is just missing

tried `~/bin/typos/typos src/modules/job-manager` and got nothing.

src/cmd/flux-update.py
grondo: Yeah, at some point someone is going to use `flux job info JOBID jobspec` and get the wrong information. But this is what we have for now.

src/modules/job-manager/update.c
chu11: should add to typos when we find them :-)  

https://github.com/crate-ci/typos/issues/784

grondo: Ah, well I did a good job with my typos then. :-)

t/job-manager/plugins/jobspec-update-job-list.c
```diff
@@ -71,6 +78,7 @@ static int sched_cb (flux_plugin_t *p,
                                      "update failure");
         return -1;
     }
+    updated = true;
     return 0;
 }
```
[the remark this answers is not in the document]
grondo: For these tests it is probably fine to post updates from `job.validate` since they are run in a controlled environment (i.e. no out-of-band updates are coming in via the update service). Updates are only validated when posted through the update service...

t/t2290-job-update.t
```diff
@@ -0,0 +1,130 @@
...
+		2>inactive-jobid.err &&
+	test_debug "cat inactive-jobid.err" &&
+	grep "job is inactive" inactive-jobid.err
+'
+test_expect_success 'invalid jobspec key cannot be updated' '
+	test_expect_code 1 flux update $jobid foo=bar
+'
+test_expect_success 'invalid jobspec key cannot be updated' '
+	test_expect_code 1 flux update $jobid foo=bar
+'
```
chu11, line 43: duplicates :-)

grondo: Thanks! Just as a warning, I'm still working on this WIP PR ;-) (though honestly I probably would have missed this duplicate, so good catch) I just wanted to see the coverage of the current set of tests -- though I don't see a report here so :-(

src/modules/job-manager/event.c
```diff
@@ -813,6 +813,33 @@
...
+    if (streq (name, "jobspec-update")) {
+        json_t *updates;
+        if (json_unpack (entry, "{s:o}", "context", &updates) < 0) {
+            flux_log (event->ctx->h,
+                      LOG_ERR,
+                      "unable to unpack jobspec-update contexto for %s",
+                      idf58 (job->id));
+            return -1;
+        }
+        (void) jobtap_call (event->ctx->jobtap,
```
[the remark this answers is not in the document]
grondo: False positive.

grondo:
Ok, I pushed a minimal manpage for `flux-update(1)` and removed WIP. This is probably ready for a full review. A couple things to consider while reviewing:

 - Is `flux update` an acceptable name for this command? It is kind of generic, but `flux job-update` might be inconvenient because other commands begin with `flux job*`, and `flux job update` won't work since it is convenient to write this command in Python. Also figured we have `flux cancel` not `flux job-cancel` so perhaps `flux update` is ok.
 - The caveat in [this comment](https://github.com/flux-framework/flux-core/pull/5409#issuecomment-1693756265) bothers me. However, I'm not thinking of a good way around it for now. If it is decided that his behavior is ok (a previous update that bypasses limits or validation may prevent future updates to a job), then I can add some notes about this to the manual page.
 - Is the scheme for allowing individual updates by plugin acceptable? As a note, I was thinking that some more common or simple updates could be directly allowed by pattern in `job-manager/update.c` itself (e.g. job name seems like a simple one)

grondo:
Also tagging @ryanday36, who should give his approval before we merge this PR.

chu11:
> Is flux update an acceptable name for this command? It is kind of generic, but flux job-update might be inconvenient because other commands begin with flux job*, and flux job update won't work since it is convenient to write this command in Python. Also figured we have flux cancel not flux job-cancel so perhaps flux update is ok.

Random idea (not entirely sure its good, but throwing it out there), could do extra level subcommand i.e. `flux update job name=foo`, `flux update R expiration=1234`, and maybe there will be others in the future?  and perhaps `flux update job help` can list common options?  and I guess `flux update` itself would list "subcommands"?

grondo:
> Random idea (not entirely sure its good, but throwing it out there), could do extra level subcommand i.e. flux update job name=foo, flux update R expiration=1234

Not a bad idea, but I was thinking expiration for a running job would be done either via special handling of `flux update JOBID duration=+10m` in the update service, or an alias of `flux update JOBID expiration=+10m`, since most users are probably blissfully unaware of R.

ryanday36:
I think that `flux update` is good and like the `flux update JOBID ...` syntax. The issue with validation failing for changes made after a `validated` change has been accepted is a tricky one. I don't think it should hold this up as an initial implementation, but it does seem like something that will need to get addressed eventually. Can you give these plugins a way to pass something to `job.validate` so that it ignores errors from keys that the plugin doesn't change?

grondo:
> Can you give these plugins a way to pass something to job.validate so that it ignores errors from keys that the plugin doesn't change?

That approach may not work in all scenarios because a change in one attribute might affect validation of others. The obvious example would be a change a queue. The change in the queue doesn't affect the duration of the job, however the queue may have a different time limit from the job's previous queue. (Similar problem with bank?) This is why running the entire `job.validate` stack is compelling after any update to the job.

Another foolproof solution would be to just not allow limits to be bypassed when using `flux update`. However, I have been asked at least once for this feature so this ability may be valuable in the near term.

One vague idea would be some way to add a flag that persists to individual attributes of a job. All `job.validate` plugins could optionally check for this flag and ignore the respective attribute if that flag is set. :shrug:

I guess as you said, we could probably come up with some mitigation in the near future if we thought about this for awhile, so thanks for that suggestion.

ryanday36:
That sounds like a nice solution to this problem to me.

grondo:
Ok, I've added a new `immutable` flag to the job manager `struct job` and support for setting it via the `set-flags` event.

Guest jobs that are updated by the instance owner and bypass validation now have this flag set before the `jobspec-update` event is posted. This results in the error:

```
flux-update: ERROR: job is immutable due to previous instance owner update
```

The tests were expanded to cover this case.

garlick:
Nice!

chu11 (approved):
second pass, overall LGTM, just small things

etc/completions/flux.pre
```diff
@@ -159,6 +159,29 @@ _flux_cancel()
...
+#  flux-update(1) completions
+_flux_update()
+{
+    local cmd=$1
+    OPTS="\
+        -v, --verbose \
+        -n, --dry-run \
+    "
+    if [[ $cur != -* ]]; then
+        #  Attempt to substibute a pending jobid
```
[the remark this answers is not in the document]
chu11: as an aside, another one to submit to typo checker :-)

garlick:
I've played with this a bit on my test cluster that has limits set up and it seems to work as advertised!

src/modules/job-manager/job.c
```diff
@@ -588,6 +588,59 @@ const char *job_event_queue_print (struct job *job, char *buf, int size)
...
+
+json_t *job_jobspec_with_updates (struct job *job, json_t *updates)
+{
+    json_t *jobspec;
+
+    if (!job->jobspec_redacted
+        || !(jobspec = json_deep_copy (job->jobspec_redacted))) {
+        errno = ENOMEM;
+        return NULL;
+    }
```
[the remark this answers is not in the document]
grondo: Maybe EAGAIN, though I don't really think this error can happen, I'll fix.

grondo:
Thanks @chu11 and @garlick! I've fixed up the final comments by @chu11 and force pushed the result and will set MWP.

grondo:
Hm, the `el8 - ascii` checks failed with:
```
    File "/usr/src/t/scripts/sign-as.py", line 3, in <module>
      from flux.security import SecurityContext
    File "/usr/src/src/bindings/python/flux/security.py", line 11, in <module>
      from _flux._security import ffi, lib
```

I better check if any `FLUX_SECURITY` prereq is needed on some of these tests.

PR #5410, opened 2023-08-25
actions: update typo checker version

chu11:
nothing fancy here, except I chose to add the british "requestor" to the false positives list, instead of change to use the american "requester".  Could fix of course ... we did away with the british "cancelled" awhile back.

Update: FWIW

```
>git grep -i "requestor" | wc -l
13
>git grep -i "requester" | wc -l
2
```

Edit:

Ok, googling more, I guess in computer programming circles "requestor" is liked more.  But I guess it doesn't really matter.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -587,7 +587,7 @@ Other highlights include:
  * configure.ac: fix error message when running autogen.sh (#4590)
  * job-manager: print better errors on inactive move (#4586)
  * broker: fix use-after-free segfault (#4570)
- * python: uri: use path to current flux executable in lsf resovler (#4559)
+ * python: uri: use path to current flux executable in lsf resolver (#4559)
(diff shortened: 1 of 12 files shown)


NEWS.md
```diff
@@ -587,7 +587,7 @@ Other highlights include:
  * configure.ac: fix error message when running autogen.sh (#4590)
  * job-manager: print better errors on inactive move (#4586)
  * broker: fix use-after-free segfault (#4570)
- * python: uri: use path to current flux executable in lsf resovler (#4559)
```
grondo, line 590: For consistency, this commit message should probably go ahead and use the Problem: format. However, I'm having a hard time caring that much so I leave it up to you

grondo (approved):
Great! Thanks!!

Issue #301, opened 2015-07-30
JSON parser/generator issues/thoughts


trws:
As more and more of flux relies, implicitly or explicitly, on the handling of JSON strings, I thought it might be good to put up a discussion issue for the state of things.  Currently we're using `json-c`, which does the job well for the most part, but has some unfortunate shortcomings where efficiency is concerned in some of our use-cases.

Notably, there is no way to get a size in number of elements for a json object other than to iterate over the object in its entirety and collect a count.  It also ranks as one of the slower and higher memory use json implementations in the [nativejson-benchmark](https://github.com/miloyip/nativejson-benchmark).

It's certainly not the top of the priority list, but some thoughts on alternatives might be worthwhile.

Of those with native C interfaces, Jansson seems to have the most features.  Yajl is pretty fast and actually installed on LC systems, but comes as a callback mechanism with no pre-defined data representation, definitely not batteries included, and ultrajson (oddly a C module for python, but by a wide margin the fastest pure C implementation) is the same way in the guise of ujson4c.

In a perfect world, RapidJson would be the obvious choice being almost always the fastest, lowest memory, most feature-packed and battle-tested, but it has no native C wrapper.  Even so, using it to implement the jsonutil and shortjson primitives might be worth it, and the [JSON Pointers](https://tools.ietf.org/html/rfc6901) interface could be wrapped rather cleanly in C (none of the C-based interfaces seem to provide this yet).

garlick:
Thanks for the excellent run-down of options.

It would be nice if there were a JSON implementation with a decent enough API that we could use without wrappers like "shortjson.h".  Transitioning to the native API would probably be a lot safer than transitioning the shared utility functions, since there are still too many places where the JSON type is used interchangeably with json_object.

I haven't looked at all the options you describe above but I did spend some time looking at jansson at one point and thought it's API was pretty well thought out, a clear improvement over json-c.

garlick:
Actually, a plan of attack could be
1) finish excise of `json_object` from public API's
2) untangle  wrapper and `json_object` usage by redefining JSON type to no longer be an alias for `json_object *`, and fixing users identified by compiler error to be either explicit json-c or wrapper.
3) migrate wrapper implementation to new json
4) migrate explicit json-c to explicit new json
5) convert wrapper users  to explicit new json (unless we decide to keep wrappers)

trws:
[...]

PR #5413, opened 2023-08-25
job-list: support getting job project and bank

chu11:
With the `jobspec-update` support from #5408 and the upcoming `flux update` support in #5409, this was half-easy to add so decided to add it (issues #4697 and #4698).

This is dependent on #5409 going in of course.

It's worth noting that nothing in `flux-accounting` does this yet, so it has no practical use.  We may want to hold off on eventual merge until `flux-accounting` merging some stuff in for this.


chu11:
with #5408 and #5409 now in, rebased and removed WIP.

Note that it may be a bit premature to merge until flux-accounting supports in their plugins (such as #301).  But there is nothing wrong with adding it as is for when that support comes from flux-accounting later.

grondo (approved):
This LGTM.

PR #5416, opened 2023-08-31
flux-perilog-run: support prolog and epilog timeouts (default 30m)

grondo:
This PR switches the `flux-perilog-run.py` script to run local and "per rank" prolog and epilog processes under a timeout. By default the timeout is 30m (maybe should be 1h or unlimited by default?), but a `-t, --timeout=FSD` option is added for timeout customization.

In order to more easily support execution of multiple processes under a timeout in this script, it has pretty much been rewritten to use `asyncio`. This has the added benefit that now all local and per-rank executions are done in parallel. There was a bit of other work necessary to make the signaling of timed out processes more reliable, etc.

### DIFF
--- a/src/modules/job-manager/plugins/perilog.c
+++ b/src/modules/job-manager/plugins/perilog.c
@@ -138,24 +138,49 @@ static void perilog_proc_delete (struct perilog_proc *proc)
     }
 }
 
-static void emit_finish_event (struct perilog_proc *proc, int status)
+static void emit_finish_event (struct perilog_proc *proc,
+                               flux_subprocess_t *sp)
 {
+    int status = flux_subprocess_status (sp);
     if (proc->prolog) {
         /*
          *  If prolog failed, raise job exception before prolog-finish
          *   event is emitted to ensure job isn't halfway started before
          *   the exception is raised:
          */
-        if (status != 0
-            && flux_jobtap_raise_exception (proc->p,
-                                            proc->id,
-                                            "prolog",
-                                            0,
-                                            "prolog failed with status=%d",
-                                            status) < 0)
-                flux_log_error (flux_jobtap_get_flux (proc->p),
-                                "prolog-finish: flux_jobtap_raise_exception");
+        if (status != 0) {
+            int code = flux_subprocess_exit_code (sp);
+            int rc;
+            int sig;
+            char *errmsg;
 
+            if ((sig = flux_subprocess_signaled (sp)) > 0
+                || (sig = (code - 128)) > 0) {
+                rc = asprintf (&errmsg,
+                               "prolog killed by signal %d%s",
+                               sig,
+                               sig == SIGTERM ?
+                               " (timeout or job canceled)" :
+                               "");
+            }
+            else
+                rc = asprintf (&errmsg,
+                               "prolog exited with exit code=%d",
+                               flux_subprocess_exit_code (sp));
+            if (rc < 0)
+                errmsg = NULL;
+            if (flux_jobtap_raise_exception (proc->p,
+                                             proc->id,
+                                             "prolog",
+                                             0,
+                                             "%s",
+                                             errmsg ?
+                                             errmsg :
+                                             "job prolog failed") < 0)
+                    flux_log_error (flux_jobtap_get_flux (proc->p),
+                                    "prolog-finish: jobtap_raise_exception");
+            free (errmsg);
+        }
         if (flux_jobtap_prolog_finish (proc->p,
                                        proc->id,
                                        "job-manager.prolog",
@@ -188,7 +213,7 @@ static void completion_cb (flux_subprocess_t *sp)
 {
     struct perilog_proc *proc = flux_subprocess_aux_get (sp, "perilog_proc");
     if (proc) {
-        emit_finish_event (proc, flux_subprocess_status (sp));
+        emit_finish_event (proc, sp);
         perilog_proc_delete (proc);
     }
 }
@@ -220,7 +245,7 @@ static void state_cb (flux_subprocess_t *sp, flux_subprocess_state_t state)
                   flux_subprocess_state_string (flux_subprocess_state (sp)),
                   code);
 
-        emit_finish_event (proc, code);
+        emit_finish_event (proc, sp);
         perilog_proc_delete (proc);
     }
 }
--- a/src/cmd/flux-job.c
+++ b/src/cmd/flux-job.c
@@ -2481,6 +2481,7 @@ static void attach_notify (struct attach_ctx *ctx,
     if (!event_name)
         return;
     if (ctx->statusline
+        && !ctx->fatal_exception
         && (msg = job_event_notify_string (event_name))) {
         int dt = ts - ctx->timestamp_zero;
         int width = 80;
--- a/src/cmd/flux-perilog-run.py
+++ b/src/cmd/flux-perilog-run.py
@@ -9,6 +9,7 @@
 ##############################################################
 
 import argparse
+import asyncio
 import logging
 import os
 import signal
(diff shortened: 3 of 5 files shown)


garlick (commented):
Nice!  I'll give this a spin on my test cluster.

grondo:
Oops, I found a typo (not sure how that snuck in). Pushed a fix now.

garlick:
For some reason this isn't working for me.  My prolog/epilog config looks like this:
```toml
[job-manager.prolog]
command = [
   "--sdexec",
   "--timeout=10s",
   "-e", "/usr/lib/aarch64-linux-gnu/flux/flux-imp,run,prolog"
]

[job-manager.epilog]
command = [
   "--sdexec",
   "--timeout=10s",
   "-e", "/usr/lib/aarch64-linux-gnu/flux/flux-imp,run,epilog"
]
```
My prolog and epilog scripts each contain a sleep 60.

The observation is that each runs the full 60s despite the 10s timeout.

flux-version reports `0.53.0-131-ge74fad1a5`

grondo:
Hrm, I bet the problem is that the signal forwarded by flux-exec can't be delivered to processes launched by the IMP since they are running as root. This is an unfortunate oversight.. I'm not sure if there's a workaround for this presently. :thinking:

garlick:
Well, when I drop the `--sdexec` I do get a failure after 10s:
```
$ flux run -q batch --requires=host:picl7  hostname
15.636s: job.exception type=prolog severity=0 prolog failed with status=3660815
flux-job: ƒYGpCs8X7M1 starting                                         00:00:15
```
and
```
Aug31 23:48  drained  prolog timed out for jobid ƒY+ picl7
```

Then looks like the epilog does not run (expected?)

grondo:
Well, that surprises me a bit, but good that it is working in the non-sdexec case. 

>  Then looks like the epilog does not run (expected?)

Yes, a failure in the prolog should raise a fatal job exception before the the execution system event launches the job shells.
Since there is no `start` event and thus no `finish` event, the job epilog does not run. (Now that you brought it up, though, I wonder if the epilog should run any time the prolog ran?)

grondo:
Does signal propagation through `flux exec` work in general when sdexec is being used?

garlick:
> Does signal propagation through flux exec work in general when sdexec is being used?

That is my worry.  FWIW this quick test worked:
```
$ sudo -u flux flux exec -r 7 --service=sdexec sleep 60
^C
7: Interrupt
```
(and both flux-exec and the remote unit are no longer running)

grondo:
So there is perhaps only an issue when `flux-imp` is involved. `flux-exec` could be taught to execute `flux imp kill` to signal processes running under the IMP, but to make the determination to do that it would have to examine the command line, which seems a little kludgy and error prone.

grondo:
So the problem is that signals cannot be propagated to processes executed under `flux-imp run` since they have elevated privileges. That is, the Ctrl-C test fails in both of these cases:
```console
$ sudo -u flux flux exec -r 0 --service=sdexec /usr/lib/aarch64-linux-gnu/flux/flux-imp run prolog
^C^Cinterrupt (Ctrl+C) one more time within 1.00 sec to exit
$ sudo -u flux flux exec -r 0 /usr/lib/aarch64-linux-gnu/flux/flux-imp run prolog
^C^Cinterrupt (Ctrl+C) one more time within 1.00 sec to exit
```

`flux-imp kill SIGNAL -PID` seems to work here when run by hand.. trying to think about how to do enable this automatically without being too kludgy... Maybe an option on the command line?

In any event, this PR should be put on hold until `flux exec` is  improved.

grondo:
Ok, this proof-of-concept works, though it isn't very satisfying. Basically if `flux-imp` is found in `argv`, then remotely execute `flux-imp kill SIGNAL -PID` to terminate each subprocess:

```diff
diff --git a/src/cmd/flux-exec.c b/src/cmd/flux-exec.c
index d47e73d8a..9e1b3b776 100644
--- a/src/cmd/flux-exec.c
+++ b/src/cmd/flux-exec.c
@@ -55,6 +55,7 @@ static struct optparse_option cmdopts[] = {
 
 extern char **environ;
 
+flux_t *h = NULL;
 uint32_t rank_range;
 uint32_t rank_count;
 uint32_t started = 0;
@@ -76,6 +77,9 @@ flux_watcher_t *stdin_w;
 struct timespec last;
 int sigint_count = 0;
 
+bool use_imp_kill = false;
+const char *imp_path = NULL;
+
 void output_exitsets (const char *key, void *item)
 {
     struct idset *idset = item;
@@ -239,10 +243,69 @@ static void stdin_cb (flux_reactor_t *r, flux_watcher_t *w,
     }
 }
 
-static void signal_cb (int signum)
+static void kill_completion_cb (flux_subprocess_t *p)
+{
+    flux_subprocess_destroy (p);
+}
+
+static flux_subprocess_t *imp_kill (flux_subprocess_t *p, int signum)
+{
+    flux_cmd_t *cmd;
+    flux_subprocess_ops_t ops = {
+        .on_completion = kill_completion_cb,
+        .on_state_change = NULL,
+        .on_channel_out = NULL,
+        .on_stdout = output_cb,
+        .on_stderr = output_cb,
+    };
+
+    pid_t pid = flux_subprocess_pid (p);
+    int rank = flux_subprocess_rank (p);
+
+    if (!(cmd = flux_cmd_create (0, NULL, environ))
+        || flux_cmd_argv_append (cmd, imp_path) < 0
+        || flux_cmd_argv_append (cmd, "kill") < 0
+        || flux_cmd_argv_appendf (cmd, "%d", signum) < 0
+        || flux_cmd_argv_appendf (cmd, "-%ld", (long) pid)) {
+        fprintf (stderr,
+                 "Failed to create flux-imp kill command for rank %d pid %d\n",
+                 rank, pid);
+        return NULL;
+    }
+    /* Note: subprocess object destroyed in completion callback
+     */
+    return flux_rexec (h, rank, 0, cmd, &ops);
+}
+
+static void killall (zlist_t *subprocesses, int signum)
 {
     flux_subprocess_t *p = zlist_first (subprocesses);
+    while (p) {
+        if (flux_subprocess_state (p) == FLUX_SUBPROCESS_RUNNING) {
+            if (use_imp_kill) {
+                if (!imp_kill (p, signum))
+                    fprintf (stderr,
+                             "failed to signal rank %d: %s\n",
+                             flux_subprocess_rank (p),
+                             strerror (errno));
+            }
+            else {
+                flux_future_t *f = flux_subprocess_kill (p, signum);
+                if (!f) {
+                    if (optparse_getopt (opts, "verbose", NULL) > 0)
+                        fprintf (stderr, "failed to signal rank %d: %s\n",
+                                flux_subprocess_rank (p), strerror (errno));
+                }
+                /* don't care about response */
+                flux_future_destroy (f);
+            }
+        }
+        p = zlist_next (subprocesses);
+    }
+}
 
+static void signal_cb (int signum)
+{
     if (signum == SIGINT) {
         if (sigint_count >= 2) {
             double since_last = monotime_since (last);
@@ -265,19 +328,7 @@ static void signal_cb (int signum)
         fprintf (stderr, "sending signal %d to %d running processes\n",
                  signum, started - exited);
 
-    while (p) {
-        if (flux_subprocess_state (p) == FLUX_SUBPROCESS_RUNNING) {
-            flux_future_t *f = flux_subprocess_kill (p, signum);
-            if (!f) {
-                if (optparse_getopt (opts, "verbose", NULL) > 0)
-                    fprintf (stderr, "failed to signal rank %d: %s\n",
-                             flux_subprocess_rank (p), strerror (errno));
-            }
-            /* don't care about response */
-            flux_future_destroy (f);
-        }
-        p = zlist_next (subprocesses);
-    }
+    killall (subprocesses, signum);
 
     if (signum == SIGINT) {
         if (sigint_count)
@@ -322,11 +373,22 @@ char *split_opt (const char *s, char sep, const char **val)
     return cpy;
 }
 
+static bool check_for_imp_run (int argc, char *argv[], const char **ppath)
+{
+    for (int i = 0; i < argc; i++) {
+        if (streq (basename (argv[i]), "flux-imp")) {
+            *ppath = argv[i];
+            return true;
+        }
+    }
+    return false;
+}
+
+
 int main (int argc, char *argv[])
 {
     const char *optargp;
     int optindex;
-    flux_t *h;
     flux_reactor_t *r;
     struct idset *ns;
     uint32_t rank;
@@ -355,6 +417,10 @@ int main (int argc, char *argv[])
         exit (1);
     }
 
+    use_imp_kill = check_for_imp_run (argc - optindex,
+                                      &argv[optindex],
+                                      &imp_path);
+
     if (!(cmd = flux_cmd_create (argc - optindex, &argv[optindex], environ)))
         log_err_exit ("flux_cmd_create");
 
```

garlick:
This output
```
$ flux run -N2 -q batch --requires=host:picl[6-7]  hostname
10.699s: job.exception type=prolog severity=0 prolog failed with status=3660810
flux-job: ƒYYeumQsa1V starting                                         00:00:10
```
is probably fine as is, but for extra credit it would be nice to decode the wait(2) status and suppress the "starting" message, if feasible.

garlick (approved):
LGTM, and as mentioned this works on my test system with the imp and sdexec.

It's nice that this adds the 30m timeout without immediately requiring the configs to change.

garlick:
> Good idea! I'll take a look at that. (Though my guess is a timeout in the epilog (silent to the user) will be much more common than a prolog timeout)

Could easily be done as a separate PR if it's useful to do so.  I hate to hold this up.

src/modules/job-manager/plugins/perilog.c
```diff
@@ -138,24 +138,41 @@ static void perilog_proc_delete (struct perilog_proc *proc)
...
+            int code = flux_subprocess_exit_code (sp);
+            int sig;
+            char *errmsg;
+            if ((sig = flux_subprocess_signaled (sp)) > 0
+                || (sig = (code - 128)) > 0) {
+                asprintf (&errmsg,
+                          "prolog killed by signal %d%s",
+                          sig,
+                          sig == SIGTERM ? " (timeout or job canceled)" : "");
+            }
```
garlick: I think this leaks `errmsg`?

garlick (approved):
Nice! Just what looks like a small mem leak.

src/modules/job-manager/plugins/perilog.c
grondo: That is fixed in the latest push (sorry I pushed too early)

garlick: Right then this is looking good!

grondo:
Thanks! I'll set MWP then assuming tests pass.

Issue #5391, opened 2023-08-15
crashed nodes are drained with generic epilog failure messages


garlick:
Problem: node crashes tend to result in drained nodes with "reasons" like `epilog failed for jobid f28XEx4Rf73d`, which is not as helpful as it could be.

Request from the sys admins is to somehow indicate that "no route to host" is the reason.  Otherwise they have to go digging in the logs to see what part of the epilog script might have failed.

grondo:
The epilog is currently run via the `flux-perilog-run.py` helper command.

One solution would be to skip attempting to run `flux exec -r{rank}` on nodes that are currently offline at the time the script is started, since `flux exec` will definitely fail, and instead preemptively drain the node with a specific error `node was offline for job ID epilog` or similar.

I assume in our environment the flux broker is not started on nodes until there is manual intervention, so there wouldn't necessarily be a race here where the node would come back in time to run the epilog successfully even though it had crashed.

watson6282:
Correct.  We still require a manual `nodeup` to start the flux broker (and other stuff) after a node reboot.  It takes over ten minutes for a node to reboot back to a point where `nodeup` can be run.  I'm assuming that is long enough for the epilog-running process to complete and drain these down nodes.  We are also very unlikely to `nodeup` nodes that don't have a drain message.

PR #5417, opened 2023-08-31
cmd: flux-perilog-run: avoid running prolog/epilog on offline ranks

grondo:
This PR updates the `flux perilog-run` script to avoid running the prolog or epilog on offline ranks, which is guaranteed to fail anyway. When this occurs, a message is logged and the command exits with nonzero status (as if the prolog/epilog failed on the offline ranks), but by default offline ranks are not drained (or have their drain reason updated). A new `-D, --drain-offline` option is provided which causes the command to update/add a drain reason of, .e.g. `offline for JOBID epilog`.

### DIFF
--- a/src/cmd/flux-perilog-run.py
+++ b/src/cmd/flux-perilog-run.py
@@ -143,6 +143,8 @@ async def run_per_rank(name, jobid, args):
         returncode = 1
         LOGGER.info("%s: %s: ranks %s offline. Skipping.", jobid, name, offline)
         ranks.subtract(offline)
+        if args.drain_offline:
+            drain(handle, offline, f"offline for {jobid} {name}")
 
     if args.verbose:
         LOGGER.info(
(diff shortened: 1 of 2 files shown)


garlick:
> Problem: If ranks are offline then a prolog/epilog executed on those
> ranks in --exec-per-rank mode is guaranteed to fail. But this is
> pointless and may overwrite any existing drain reason unecessarily.

Is that last bit true?  It looks like `flux-perilog-run.py` calls the `resource.drain` RPC with `mode=update`, which according to the resource module source:

> if a node in drain target is already drained, request status depends on setting of optional 'mode' member:
> [snip]
> If mode=update, request succeeds and reason is updated only for
> those target that are not drained or do not have reason set.

grondo:
Good question! That seemed to be the case as reported in #5391 --  maybe the wrong assumption about the root cause was made in that issue.

garlick (approved):
LGTM.  Maybe add Fixes #5391 to the first commit?

grondo:
Added the suggested amendment to first commit message and will set MWP. Thanks!

PR #5418, opened 2023-08-31
job-list: allow updates to all of jobspec

chu11:
Per a recent update to RFC21, all fields in a jobspec can now be updated via a jobspec-update event.

PR #5419, opened 2023-09-01
job-manager: support jobspec update to all fields

chu11:
Problem: An update to RFC21 allows all jobspec fields to be updated by a jobspec-update event.  However not all fields are allowed to be updated via the validate_jobspec_updates() function.

Solution: Allow all fields in current legal jobspecs to be updated.

-- 

A split off from #5418


grondo (approved):
Nice! LGTM!

PR #5420, opened 2023-09-01
shell/pmi: warn if application might be using slurm's libpmi2.so

garlick:
Problem: on a slurm system, it's easy to trip over the libpmi2.so that they install in libdir.

To get around that library's quirk of ignoring version negotiation, we already disabled PMI version negotiation in our server.   This converts a hang into a hard failure.  However, we didn't have a way for the protocol engine to warn the user what we suspect may be the cause.

Add a new callback so that the shell pmi plugin can call `shell_warn()` when the protocol engine encounters this situation:

```
$ flux run  -N2 flux pmi --method=libpmi2
flux-pmi: /lib/x86_64-linux-gnu/libpmi2.so: pmi error 14
flux-pmi: /lib/x86_64-linux-gnu/libpmi2.so: pmi error 14
0.047s: flux-shell[0]:  WARN: pmi-simple: is application unintentionally using slurm libpmi2.so?
```

Note: on the :coffee: discussion today I mentioned that I needed a change to subprocess buffering to parse PMI-2 protocol request.  That would be necessary if we wanted to retain the ability to do protocol negotiation.  This PR just adds the warning to the point where we refuse to negotiate, so no such changes needed.

### DIFF
--- a/src/common/libpmi/simple_server.c
+++ b/src/common/libpmi/simple_server.c
@@ -129,6 +129,21 @@ static void trace (struct pmi_simple_server *pmi,
     }
 }
 
+static void warn (struct pmi_simple_server *pmi,
+                  void *client,
+                  const char *fmt,
+                  ...)
+{
+    char buf[1024];
+    va_list ap;
+
+    va_start (ap, fmt);
+    (void)vsnprintf (buf, sizeof (buf), fmt, ap);
+    va_end (ap);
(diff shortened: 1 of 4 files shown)


grondo (approved):
LGTM!

garlick:
Thanks! I'll set MWP.

PR #5421, opened 2023-09-02
ci: fix failure detection builds with coverage and remove obsolete system tests

grondo:
I noticed that the "system" tests have been failing in ci, but because of a broken &&-chain in `CHECKCMDS`, the failures were going undetected.

This PR fixes `CHECKCMDS`, and also removes some obsolete system tests that test the old sdexec implementation (and were therefore failing).

Finally, I noticed an obsolete check in `etc/rc1` for `exec.method = "systemd"`, so that is removed as well.


chu11 (approved):
LGTM!

grondo:
Thanks, I'll set MWP.

garlick (commented):
LGTM!

PR #5422, opened 2023-09-03
cmd: add --with-imp options to flux-exec(1) and flux-perilog-run(1)

grondo:
As suggested offline by @garlick, add a `--with-imp` convenience option to both `flux exec` and `flux perilog-run` that prepends `/path/to/flux-imp run` (fetched from instance configuration) to the command to be executed. Note this requires that the command be configured in the IMP `run` config. At this time this is mostly for `prolog` and `epilog` scripts.

In `flux-exec(1)`, when the `--with-imp` option is used or if `flux-imp` is the first argument of the command, `flux-imp kill` will be used to forward signals to remote processes instead of `flux_subprocess_kill()`. This allows the signals to actually be delivered to the processes spawned by `flux exec` since they are likely running with privilege.

Tests for this feature are added to the "system" ci tests since the IMP is required for a full functionality test.

### DIFF
--- a/src/cmd/flux-exec.c
+++ b/src/cmd/flux-exec.c
@@ -50,6 +50,8 @@ static struct optparse_option cmdopts[] = {
     { .name = "setopt", .has_arg = 1, .arginfo = "NAME=VALUE",
       .flags = OPTPARSE_OPT_HIDDEN,
       .usage = "Set subprocess option NAME to VALUE (multiple use ok)" },
+    { .name = "with-imp", .has_arg = 0,
+      .usage = "Run args under 'flux-imp run'" },
     OPTPARSE_TABLE_END
 };
 
@@ -75,6 +77,9 @@ flux_watcher_t *stdin_w;
 
 struct timespec last;
 int sigint_count = 0;
+
+bool use_imp = false;
+const char *imp_path = NULL;
 
 void output_exitsets (const char *key, void *item)
 {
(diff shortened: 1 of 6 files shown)


grondo:
Renamed some global and local variables to avoid the CodeQL warnings above (didn't want to go through a full refactoring of `flux-exec` at this time)

garlick:
This seems to be working!

My config is now
```toml
[job-manager.prolog]
command = [
   "--sdexec",
   "--timeout=10s",
   "--with-imp",
   "-e", "prolog"
]

[job-manager.epilog]
command = [
   "--sdexec",
   "--timeout=10s",
   "--with-imp",
   "-e", "epilog"
]
```

When I add a sleep 60 in the prolog on picl7, I get:
```
$ flux run -q batch --requires=host:picl7  hostname
10.685s: job.exception type=prolog severity=0 prolog failed with status=3660810
flux-job: ƒYQGGdcm7V1 starting                                         00:00:10
```

src/cmd/flux-exec.c
```diff
@@ -244,10 +245,69 @@ static void stdin_cb (flux_reactor_t *r, flux_watcher_t *w,
...
+    };
+
+    pid_t pid = flux_subprocess_pid (p);
+    int rank = flux_subprocess_rank (p);
+
+    if (!(cmd = flux_cmd_create (0, NULL, environ))
+        || flux_cmd_argv_append (cmd, imp_path) < 0
+        || flux_cmd_argv_append (cmd, "kill") < 0
+        || flux_cmd_argv_appendf (cmd, "%d", signum) < 0
+        || flux_cmd_argv_appendf (cmd, "-%ld", (long) pid)) {
```
garlick, line 271: For consistency the last `flux_cmd_argv_appendf()` call in the chain should probably check for < 0

src/cmd/flux-exec.c
```diff
@@ -322,6 +327,35 @@ char *split_opt (const char *s, char sep, const char **val)
...
+    if (!(f = flux_rpc (h, "config.get", NULL, FLUX_NODEID_ANY, 0))
+        || flux_rpc_get_unpack (f,
+                                "{s?{s?s}}",
+                                "exec",
+                                "imp", &imp) < 0)
+        fprintf (stderr, "error fetching config object: %s",
+                 future_strerror (f, errno));
+    flux_aux_set (h, NULL, f, (flux_free_f) flux_future_destroy);
+    return imp;
+}
```
garlick, line 404: - `flux_` prefix in a local function could be a tiny bit confusing
- this function fprintf's an error, then its caller prints another one on failure

src/cmd/flux-exec.c
```diff
@@ -394,6 +428,20 @@ int main (int argc, char *argv[])
     if (flux_get_size (h, &rank_range) < 0)
         log_err_exit ("flux_get_size");
 
+    if (optparse_hasopt (opts, "with-imp")) {
+        if (!(imp_path = flux_imp_path (h)))
+            log_err_exit ("--with-imp: exec.imp path not found in config");
+        use_imp = true;
+        if (flux_cmd_argv_insert (cmd, 0, "run") < 0
+            || flux_cmd_argv_insert (cmd, 0, imp_path))
```
garlick, line 487: Second `flux_cmd_argv_insert()` should check for < 0 like the first.

src/cmd/flux-exec.c
grondo: Good point, I'll rename the function `get_flux_imp_path()`. 

I think both errors may be needed because if the path to the IMP is not set `flux_rpc_get_unpack()` will not return an error, but the `--with-imp` option should still fail.

src/cmd/flux-exec.c
grondo: D'oh!

garlick:
Yes, great!

PR #5424, opened 2023-09-05
support update of job queue

grondo:
This is an experimental WIP that allows users to move a pending job to a new queue.

The list of valid queues and their current status (enabled/disabled) is currently private to `job-manager/queue.c`, so it isn't straightforward to add an `update-queue` plugin. The plugin would have to send an RPC back to the job-manager itself to query the list of queues and their states (and there's currently no way to do asynchronous work in one of these plugins), or a `jobtap` API call to "unpack" the queues or otherwise query them would be required.

Instead, this PR proposes a way for job manager components to register builtins directly from within their initialization routines, and pass some opaque data along that can be passed to all jobtap callbacks. This seems a little backwards, since it exposes internals to jobtap plugins, which slightly breaks the design, but it happens to be very useful for the specific case of allowing updates for an attribute of jobspec which is managed by a private job manager component.

This allows the `.update-queue` plugin to easily access queue-specific data without RPCs or ad-hoc jobtap functions that expose some internals of the queue module. It also keeps all the queue related code in one place, and reduces the amount of boiler plate code to a minimum, while still allowing the "plugin" to be unloaded to prevent queue updates, or reloaded to re-enable them. If other queue restrictions are added in the future (e.g. access controls) these can easily be supported in the `.update-queue` plugin since it is internal to `job-manager/queue.c`.

Of course the drawback is that someone could never write a truly external queue-update plugin since we've not exposed those internals for general jobtap use. Since that is a pretty big caveat, I'm just proposing this PR as an experimental WIP at this time to get feedback on this particular design choice.

Either way, the next step here is to add a way for the plugin to request that a feasibility check be issued after the update has been allowed and before it is applied. This will require changes to `job-manager/update.c` itself, since the feasibility check requires an RPC to the scheduler.

Example:
```console
$ jobid=$(flux submit --urgency=hold -q batch hostname)
$ flux jobs -no {queue} $jobid
batch
$ flux update $jobid queue=debug
$ flux jobs -no {queue} $jobid
debug
$ flux update $jobid queue=foo
flux-update: ERROR: 'foo' is not a valid queue
$ flux queue disable -q batch testing update of disabled queue
batch: Job submission is disabled: testing update of disabled queue
$ flux update $jobid queue=batch
flux-update: ERROR: job submission to batch is disabled: testing update of disabled queue
```

### DIFF
--- a/doc/man7/flux-jobtap-plugins.rst
+++ b/doc/man7/flux-jobtap-plugins.rst
@@ -272,6 +272,31 @@ attributes have this flag then this validation step is skipped. This can
 be useful to allow an instance owner to update a job attribute beyond limits
 for example.
 
+Some updates may benefit from a job feasibility check before the updates
+are applied. This prevents a user from inadvertently causing a job that
+was feasible at the time of submission to become infeasible through an
+update.  Because the update plugin is in the best position to determine
+if a feasibility check should be completed for an update, feasibility
+checks are only done if a ``feasibility`` flag in ``FLUX_PLUGIN_OUT_ARGS``
+is set. If any plugin for a set of updates requires a feasibility check,
+then feasibility of the updated jobspec as a whole will be checked. If
+the updated job is determined to be infeasible, then the update is aborted
+and an error returned to the user.
+
+The update of one attribute may require modification of other attributes.
+For example, an update of ``attributes.system.queue`` may require
+modification of ``attributes.system.constraints`` to apply the constraints
+of the new queue. To support this use case, plugins may additionally push
+an ``updates`` object onto ``FLUX_PLUGIN_OUT_ARGS``. This object has the
+same form as the ``jobspec-update`` context defined in RFC 21. For example,
+if a plugin wishes to update ``attributes.system.foo`` to 1, it can set ::
+
+   {"updates": {"attributes.system.foo": 1}}
+
+in the ``FLUX_PLUGIN_OUT_ARGS`` before returning. Updates are applied by
+updating the requested updates, so this method could overwrite other user-
+requested updates and caution is advised.
+
 PLUGIN CALLBACK TOPICS
 ======================
 
--- a/src/modules/job-manager/jobtap-internal.h
+++ b/src/modules/job-manager/jobtap-internal.h
@@ -75,6 +75,10 @@ int jobtap_check_dependencies (struct jobtap *jobtap,
  *  needs_validation will be set nonzero. The caller should be sure to pass
  *  the updated jobspec to `job.validate` before posting updates to the job
  *  eventlog.
+ *
+ *  If the update requires a feasibility check with the scheduler, then
+ *  require_feasibility will be set nonzero. The caller should attempt to
+ *  request a feasibility check before applying updates.
  */
 int jobtap_job_update (struct jobtap *jobtap,
                        struct flux_msg_cred cred,
@@ -82,6 +86,7 @@ int jobtap_job_update (struct jobtap *jobtap,
                        const char *key,
                        json_t *value,
                        int *needs_validation,
+                       int *needs_feasibility,
                        char **errp);
 
 /*  Call the `job.validate` plugin stack, but using an updated jobspec by
--- a/src/modules/job-manager/jobtap-internal.h
+++ b/src/modules/job-manager/jobtap-internal.h
@@ -87,6 +87,7 @@ int jobtap_job_update (struct jobtap *jobtap,
                        json_t *value,
                        int *needs_validation,
                        int *needs_feasibility,
+                       json_t **updates,
                        char **errp);
 
 /*  Call the `job.validate` plugin stack, but using an updated jobspec by
--- a/src/modules/job-manager/jobtap-internal.h
+++ b/src/modules/job-manager/jobtap-internal.h
@@ -109,6 +109,16 @@ flux_plugin_t * jobtap_load (struct jobtap *jobtap,
                              json_t *conf,
                              flux_error_t *errp);
 
+typedef int (*jobtap_builtin_f) (flux_plugin_t *p, void *arg);
+
+/*  Add a new jobtap builtin plugin.
+ *  Allows builtins to be created externally to the jobtap module.
+ */
+int jobtap_register_builtin (struct jobtap *jobtap,
+                             const char *name,
+                             jobtap_builtin_f init_cb,
+                             void *arg);
+
 /*  Job manager RPC handler for loading new jobtap plugins.
  */
 void jobtap_handler (flux_t *h,
--- a/src/modules/job-manager/queue.c
+++ b/src/modules/job-manager/queue.c
@@ -46,6 +46,7 @@ struct jobq {
     bool checkpoint_start;  // may be different that actual start due
                             // to nocheckpoint flag
     char *stop_reason; // reason if stopped (optionally set)
+    json_t *requires;  // required properties array
 };
 
 struct queue {
@@ -62,6 +63,7 @@ static void jobq_destroy (struct jobq *q)
 {
     if (q) {
         int saved_errno = errno;
+        json_decref (q->requires);
         free (q->name);
         free (q->disable_reason);
         free (q);
@@ -78,7 +80,7 @@ static void jobq_destructor (void **item)
     }
 }
 
-static struct jobq *jobq_create (const char *name)
+static struct jobq *jobq_create (const char *name, json_t *config)
 {
     struct jobq *q;
 
@@ -87,6 +89,10 @@ static struct jobq *jobq_create (const char *name)
     if (name && !(q->name = strdup (name)))
         goto error;
     q->enable = true;
+
+    if (config && json_unpack (config, "{s?O}", "requires", &q->requires) < 0)
+        goto error;
+
     if (name) {
         q->start = false;
         q->checkpoint_start = false;
@@ -459,7 +465,7 @@ static int queue_configure (const flux_conf_t *conf,
          */
         json_object_foreach (queues, name, value) {
             if (!zhashx_lookup (queue->named, name)) {
-                if (!(q = jobq_create (name)))
+                if (!(q = jobq_create (name, value)))
                     goto nomem;
                 (void)zhashx_insert (queue->named, name, q);
             }
@@ -469,7 +475,7 @@ static int queue_configure (const flux_conf_t *conf,
         if (queue->have_named_queues) {
             queue->have_named_queues = false;
             zhashx_destroy (&queue->named);
-            if (!(queue->anon = jobq_create (NULL)))
+            if (!(queue->anon = jobq_create (NULL, NULL)))
                 goto nomem;
         }
     }
@@ -776,7 +782,7 @@ struct queue *queue_create (struct job_manager *ctx)
     if (!(queue = calloc (1, sizeof (*queue))))
         return NULL;
     queue->ctx = ctx;
-    if (!(queue->anon = jobq_create (NULL)))
+    if (!(queue->anon = jobq_create (NULL, NULL)))
         goto error;
     if (flux_msg_handler_addvec (ctx->h,
                                  htab,
--- a/src/modules/job-manager/queue.c
+++ b/src/modules/job-manager/queue.c
@@ -26,6 +26,8 @@
 
 #include "alloc.h"
 #include "job-manager.h"
+#include "jobtap-internal.h"
+#include "jobtap.h"
 #include "conf.h"
 #include "restart.h"
 #include "queue.h"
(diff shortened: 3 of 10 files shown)


garlick:
> register builtins directly from within their initialization routines, and pass some opaque data along that can be passed to all jobtap callbacks

Oh that seems like a great idea!  If it makes this type of extension easier to use and/or maintain compared to a non-plugin implementation, then the capability sounds pretty useful.

> the drawback is that someone could never write a truly external queue-update plugin since we've not exposed those internals for general jobtap use

That doesn't seem like a biggie to me?  (Or am I missing some possible use case?)

Just beginning to look at this, but should there be a call to `flux_jobtap_jobspec_update_pack()` in the `.update-queue` plugin?

grondo:
> Just beginning to look at this, but should there be a call to flux_jobtap_jobspec_update_pack() in the .update-queue plugin?

No. The way these `job.update.*` callbacks work is that the callback checks if the update is _allowable_. If it is, the callback should return `0`, i.e. success, if not `-1` is returned with an error message set via `flux_jobtap_error()`.

If there is no `job.update.` callback for the requested update, then the update is rejected. So updates have to be specifically allowed.

Once an update is allowed, then the update service in the job manager calls the full `job.validate` plugin stack to ensure the resulting updates are valid in addition to allowed.

grondo:
> That doesn't seem like a biggie to me? (Or am I missing some possible use case?)

I can't think of one, but wanted to bring it up in case someone else had an objection or could think of a case where it would be better to somehow expose the internal queue states via a `flux_jobtap_*` call or set of calls.

src/modules/job-manager/jobtap-internal.h
```diff
@@ -103,6 +103,16 @@ flux_plugin_t * jobtap_load (struct jobtap *jobtap,
                              json_t *conf,
                              flux_error_t *errp);
 
+typedef int (*jobtap_builtin_f) (flux_plugin_t *p, void *arg);
```
chu11, line 106: commit message typo "opqaque"

chu11:
My 10 second dumb thought, is was why can't current queues and their statuses be passed to jobtap plugins as args?

like could there be a new function `json_t *queue_status(struct queue *queue)` that returns a json dictionary of `queuename:status` pairs, and that gets passed to the jobtap plugins as args?

grondo:
> My 10 second dumb thought, is was why can't current queues and their statuses be passed to jobtap plugins as args?

I had something similar in an initial implementation. Problems as I saw them with this approach is:

 1. It adds a lot more packing work to every jobap callback for a use case currently needed by only one callback that will be rarely used
 2. It adds more boilerplate to the queue update plugin.
 3. As more queue specific internal data is added to `struct queue` and `struct jobq` this would additionally need to be exposed packed in every jobtap callback
 4. Other job manager internals may want to allow updates or register other plugins with similar access to internal structures. This adds a way to do that in general, whereas passing via jobtap plugin args would require adding those packing functions to each location. (Not impossible, just more work and over time the plugin args could get very large)

Of course the benefit of your suggestion is that it would work for externally managed plugins. :shrug:

chu11:
> It adds a lot more packing work to every jobap callback for a use case currently needed by only one callback that will be rarely used

True.  I figured we could cache the statuses b/c queue changes should be rare, leading to minimal extra packing.  But minimal != zero :P

grondo:
Jobtap plugin arguments are constructed for each jobtap plugin callback.

My initial implementation had something like a `flux_jobtap_queue_info_unpack(3)` interface function, but that required a fair bit of work, packing and unpacking on the job manager and plugin side, and I didn't relish the idea that `jobtap.h` might have a large growth of these functions over time that have to reach into otherwise encapsulated interfaces of the job manager.

grondo:
Ok, this WIP PR is now more fully fleshed out with a proposed full implementation.

Changes include:
 - The job update service now allows plugins to request a feasibility check for an update by setting a `feasibility` flag in the plugin out arguments. This not only allows the asynchronous feasibility check to occur, but also ensures that feasibility is tested against the update in its entirety. Note that if the scheduler isn't loaded the feasibility rpc will not succeed and the update will be denied.
 - The update-queue plugin requests a feasibility check of the updated jobspec.
 - Plugins can set additional required updates by adding an `updates` object to the plugin out arguments.
 - The update-queue plugin checks current job constraints, and if the constraints don't exactly match current queue constraints it rejects the update. If the constraints match, then the update-queue plugin overwrites current constraints with new queue constraints via the `updates` plugin output argument.

grondo:
Note: before this can be merged, the flux-accounting priority plugin will have to be updated to support queue updates.

grondo:
Note: This is pretty much done, just have to sync up with flux-accounting. However, removing WIP so we can get the implementation reviewd.

doc/man7/flux-jobtap-plugins.rst
```diff
@@ -272,6 +272,31 @@ attributes have this flag then this validation step is skipped. This can
 be useful to allow an instance owner to update a job attribute beyond limits
 for example.
 
+Some updates may benefit from a job feasibility check before thee updates
```
chu11, line 275: thee -> the

t/t2291-job-update-queue.t
```diff
@@ -0,0 +1,152 @@
...
+test_expect_success 'update to same queue fails' '
+	test_must_fail flux update $jobid queue=debug
+'
+test_expect_success 'update to batch queue works' '
+	flux update $jobid queue=batch &&
+	test_debug "flux job eventlog $jobid" &&
+	flux job eventlog $jobid \
+	  | grep jobspec-update \
+	  | grep attributes.system.queue=\"batch\"
+'
```
chu11, line 87: haven't tried yet, but could test `flux job list` to see if `job-list` got update (here and perhaps few other places)?  I think it should work ... if it doesn't, can be a TODO item for me to minimally add some tests in t2260-job-list.

Note, adding here could be annoying as you gotta spin a bit to wait for job-list eventual consistency.

chu11 (commented):
first pass, nothing big i noticed, mostly smaller stuff

t/t2291-job-update-queue.t
grondo: Yeah, I didn't do it because of the raciness, and job update processing in job-list is tested elsewhere so it could be considered redundant here.

src/modules/job-manager/update.c
```diff
@@ -163,7 +178,147 @@ static void update_job (struct job_manager *ctx,
...
+    flux_future_destroy (f);
+}
+
+static struct update_request *
+pending_request_create (struct update *update,
+                        const flux_msg_t *msg,
+                        struct flux_msg_cred cred,
+                        struct job *job,
+                        json_t *updates,
+                        int validate)
```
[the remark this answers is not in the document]
grondo: `pending_request_create()` calls `update_request_create()`, or perhaps I'm misunderstanding the suggestion?

src/modules/job-manager/queue.c
```diff
@@ -774,6 +776,109 @@ void queue_destroy (struct queue *queue)
...
+        if (!json_equal (constraints, expected)) {
+            flux_jobtap_error (p,
+                               args,
+                               "job appears to have non-queue constraints, "
+                               "unable to update queue to %s",
+                               name);
+            json_decref (expected);
+            return -1;
+        }
+        json_decref (expected);
```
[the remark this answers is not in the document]
grondo: I started pulling this out into its own function for more clarity, then realized that there are some bugs here when a queue does not have a configured constraint. This will require a bit more code and tests, so working on that now.

grondo:
I've addressed comments so far and rebased on current master.

Additionally, fixed handling of queues without constraints and added tests with constraint-less queue.

grondo:
Ok, @cmoussa1 has proposed flux-framework/flux-accounting#389 which implements the update logic for queues (and thus tests depend on this PR getting merged).

src/modules/job-manager/update.c
```diff
@@ -70,50 +68,67 @@
...
+};
+
+struct update_request {
+    void *handle;
+    struct update *update;
+    const flux_msg_t *msg;
+    struct flux_msg_cred cred;
+    struct job *job;
+    json_t *updates;
+    unsigned int validate:1;
```
garlick, line 85: An inline comment that `handle` is a zlistx handle for this item in `update->pending_requests` might be helpful.

src/modules/job-manager/update.c
```diff
@@ -163,7 +178,147 @@ static void update_job (struct job_manager *ctx,
...
+                                           updates,
+                                           validate))
+        || !(f = flux_rpc_pack (update->ctx->h,
+                                "sched.feasibility",
+                                0,
+                                0,
+                                "{s:O}",
+                                "jobspec", jobspec))
+        || flux_future_then (f, -1., feasibility_cb, req) < 0)
+        goto error;
```
garlick, line 258: Would the future be leaked if job-manager were unloaded with a feasibility check in progress?

src/modules/job-manager/update.c
```diff
@@ -224,18 +379,35 @@ static void update_handle_request (flux_t *h,
...
+{
+    struct update_request *req = zlistx_first (update->pending_requests);
+    while (req) {
+        if (flux_respond_error (update->ctx->h,
+                                req->msg,
+                                EAGAIN,
+                                "job manager is shutting down") < 0)
+            flux_log_error (update->ctx->h,
+                            "%s: flux_respond_error",
+                            __FUNCTION__);
```
garlick, line 413: A better log message might be "error responding to job-manager.update request".

src/modules/job-manager/queue.c
```diff
@@ -46,6 +46,7 @@ struct jobq {
     bool checkpoint_start;  // may be different that actual start due
                             // to nocheckpoint flag
```
garlick, line 49: Commit title:   totally minor but I think we've been using "topic/subtopic:" where a subtopic is needed.  "job-manager: queue:" doesn't  fit that pattern.

src/modules/job-manager/queue.c
```diff
@@ -78,7 +80,7 @@ static void jobq_destructor (void **item)
     }
 }
 
-static struct jobq *jobq_create (const char *name)
+static struct jobq *jobq_create (const char *name, json_t *value)
```
garlick, line 85: Suggestion: rename parameter `value` to `config` to indicate that it's the queue configuration for the named queue.

src/modules/job-manager/queue.c
```diff
@@ -26,6 +26,8 @@
 
 #include "alloc.h"
```
garlick, line 27: Excellent commit message!

src/modules/job-manager/queue.c
```diff
@@ -774,6 +776,171 @@ void queue_destroy (struct queue *queue)
...
+ *    (Note: json_object_size (NULL) == 0)
+ *
+ *  - json_equal(a, b) returns true
+ */
+static bool constraints_equal (json_t *c1, json_t *c2)
+{
+    if ((json_object_size (c1) == 0 && json_object_size (c2) == 0)
+        || json_equal (c1, c2))
+        return true;
+    return false;
```
garlick, line 792: I don't think you have to compare the two object sizes since [json_equal()](https://jansson.readthedocs.io/en/latest/apiref.html#c.json_equal) already requires that they have the same keys, per the jansson  [equality](https://jansson.readthedocs.io/en/latest/apiref.html#equality) definition.

t/t2291-job-update-queue.t
```diff
@@ -0,0 +1,152 @@
...
+	[policy.jobspec.defaults.system]
+	queue = "batch"
+	EOT
+	flux queue start --all &&
+	flux module unload sched-simple &&
+	flux module reload resource &&
+	flux module load sched-simple &&
+	flux queue list &&
+	flux resource list -o rlist
+'
```
chu11: should we add a variant test where there are no requires/constraints within a queue?

grondo: Oh, yeah. I swear I did that, I wonder if I just didn't push the result. Let me look around...

grondo: Oops, looks like it got squashed into the wrong commit :facepalm: I'll fix that.

t/t2291-job-update-queue.t
```diff
@@ -28,7 +28,7 @@ submit_held_job_as_guest()
 }
 
 test_expect_success 'config queues and resources' '
-	flux R encode -r 0-3 -p batch:0-3 -p debug:3 \
+	flux R encode -r 0-3 -p batch:0-2 -p debug:3 \
```
garlick: Looks like maybe the test changes should be in a different commit?

t/t2291-job-update-queue.t
grondo: Ok, moved the changes to the right commit and force-pushed the result.

garlick (approved):
Nice work!

I had a few comments but I leave it to you decide what is worth addressing before this gets merged.

t/t2291-job-update-queue.t
```diff
@@ -0,0 +1,183 @@
...
+	test_debug "flux job eventlog $jobid" &&
+	flux job eventlog $jobid \
+	  | grep jobspec-update \
+	  | grep attributes.system.queue=\"other\"
+'
+test_expect_success 'job can still run' '
+	flux job urgency $jobid default &&
+	flux job attach $jobid
+'
+test_expect_success 'job with constraints in queue without can''t be updated' '
```
chu11, line 154: typo, double apostraphe

t/t2291-job-update-queue.t
chu11: changes LGTM!

src/modules/job-manager/queue.c
garlick: Oh I just realized I misread that line of code and you're checking for empty.  Well I would assume two empty objects would be considered equal by `json_equal()` but maybe you've actually found that not to be the case!

src/modules/job-manager/queue.c
grondo: Hm, I think this seems to be personal preference:

```
 grondo@pi3:~/git/flux-core.git$ git log | grep -c job-manager:.*: 
47
 grondo@pi3:~/git/flux-core.git$ git log | grep -c job-manager/ 
57
```

BTW, I don't necessarily like `topic/subtopic` because when you search for `topic: ` thinking you'll find all matches for `topic` you are mistaken :-)

src/modules/job-manager/queue.c
grondo: No, the issue is that `json_object_size()` handles `NULL` (returns 0) and `json_equal()` does not (IIRC). We could check `c1 == NULL && c2 == NULL` but then empty objects would be unnecessarily passed to `json_equal()` for comparison, it seemed better to do that in one go... However, if it is confusing maybe the former is better code!

grondo: Just verified if either argument is `NULL` then `json_equal()` will return false, however here we want `NULL == NULL` to return true. Maybe just explicitly testing for `NULL == NULL` would be least confusing...

src/modules/job-manager/update.c
grondo: Ah, yeah. The whole reason for the pending requests list was to do this cleanup, then I forgot to add the future there! :facepalm:

src/modules/job-manager/update.c
grondo: Sounds good. However, this is basically copying existing idiom in the job-manager, shall we clean up everywhere?

```
 grondo@pi3:~/git/flux-core.git/src/modules/job-manager$ git grep "%s: flux_respond_error"
alloc.c:        flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
alloc.c:        flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
alloc.c:        flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
annotate.c:        flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
drain.c:        flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
drain.c:        flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
drain.c:                flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
getattr.c:        flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
job-manager.c:        flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
job-manager.c:        flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
kill.c:        flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
kill.c:        flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
list.c:        flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
raise.c:        flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
raise.c:        flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
start.c:        flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
submit.c:        flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
update.c:        flux_log_error (ctx->h, "%s: flux_respond_error", __FUNCTION__);
update.c:                            "%s: flux_respond_error",
update.c:        flux_log_error (ctx->h, "%s: flux_respond_error", __FUNCTION__);
update.c:        flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
update.c:                            "%s: flux_respond_error",
urgency.c:        flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
wait.c:                flux_log_error (h, "%s: flux_respond_error", __func__);
wait.c:        flux_log_error (h, "%s: flux_respond_error", __func__);
```

grondo: Although now I see you were pointing out this instance as a special case. Seems reasonable.

t/t2291-job-update-queue.t
grondo: oof!

grondo:
Ok, I think I've addressed the various comments

src/modules/job-manager/update.c
garlick: Relying on __FUNCTION__ to provide the context for the error is not always helpful, so I've been shying away from that style of late, in new code at least.  No big deal though.

grondo: I completely agree. It is especially maddening when you try to search for an error from the logs but get no hits because of the use of `__FUNCTION__`.

It did seem reasonable to fix that one spot, so I did so in the latest commit. The other uses of `__FUNCTION__` are pervasive (almost every return from a request handler uses this idiom in the codebase) so I didn't touch those.

grondo:
Ok to set MWP or @chu11 did you want another pass?

chu11:
Go for it!

PR #5428, opened 2023-09-06
flux-job: get updated version of jobspec

chu11:
The jobspec returned by `flux job info` is the one stored in the KVS, which may differ from what is internally "viewed" by flux components due to `jobspec-update` events.  Notably, the jobspec returned may not be consistent to what is returned via `flux jobs`.

This PR will make `flux job info` return an updated jobspec by default.

Added a `--noupdate` option to get the version stored in the KVS.  I'm not super in love with this option name.  I considered `--kvs`, but that was confusing give Python binding function names like ("job_kvs_lookup()").  `--stored` seemed dumb too.  Ehhh ... any better ideas?  I didn't want the option to be `--update` b/c I wanted the default to be the updated one.

Also added support in the Python bindings.

Update `flux-update`, as it no longer needs to apply `jobspec-update` events manually.

### DIFF
--- a/src/bindings/python/flux/job/kvslookup.py
+++ b/src/bindings/python/flux/job/kvslookup.py
@@ -12,7 +12,8 @@ import json
 
 from _flux._core import ffi, lib
 from flux.future import WaitAllFuture
-from flux.job import JobID
+from flux.job import JobID, JobspecV1
+from flux.job.event import EventLogEvent
 from flux.rpc import RPC
 
 
@@ -53,11 +54,15 @@ def job_info_lookup(flux_handle, jobid, keys=["jobspec"]):
     return rpc
 
 
-def _setup_lookup_keys(keys, original):
-    if original and "jobspec" in keys:
-        keys.remove("jobspec")
-        if "J" not in keys:
-            keys.append("J")
+def _setup_lookup_keys(keys, original, base):
+    if "jobspec" in keys:
+        if original:
+            keys.remove("jobspec")
+            if "J" not in keys:
+                keys.append("J")
+        elif not base:
+            if "eventlog" not in keys:
+                keys.append("eventlog")
 
 
 def _get_original_jobspec(job_data):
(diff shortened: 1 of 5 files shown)


grondo:
>  Ehhh ... any better ideas? I didn't want the option to be --update b/c I wanted the default to be the updated one.

What about `--base`?

chu11:
> What about --base?

That's not bad ... illustrates the idea well.  Lets go with it.

chu11:
repushed using the term "base" over "noupdate"

garlick:
Silly question but we have `flux job info --original ID jobspec` which unwraps the jobspec in `J`.  Do we need this also?

grondo:
>  Do we need --base also?

It seems useful at least for testing if nothing else?

chu11:
> Do we need --base also?

I did ponder this.  I added `--base` with the consideration:

- the output of `--original` and `--base` is different, one is the "original redacted" jobspec while the other is the "original original" retrieved via `J`.

- `--base` can also be used in the future with `flux job info <JOBID> R`, which get the updated R once `resource-update` is supported.  I'm not sure `--original` makes sense for `R` (given how it is used for `J` since there is no equivalent for `R`).

Not to say that these are the strongest reasons, just some reasons :-)

src/cmd/flux-job.c
```diff
@@ -3441,28 +3511,45 @@ void info_lookup (flux_t *h,
...
+                extra_key = "eventlog";
         }
 
         /* N.B. job-info.lookup will ignore duplicate keys, in the
-         * event user specified both "jobspec" and "J" on the command
-         * line with --original.
+         * event user specified duplicate keys or options that lead to
+         * duplicate keys: e.g.
+         * - --original and both "jobspec" and "J"
+         * - --base and both "jobspec" and "eventlog"
```
grondo, line 3539: From the code it appears that "jobspec" and "eventlog" are both fetched only if _not_ `--base` not with `--base`.
Is this comment correct?

src/bindings/python/flux/job/kvslookup.py
```diff
@@ -53,16 +53,11 @@ def job_info_lookup(flux_handle, jobid, keys=["jobspec"]):
     return rpc
 
 
-def _original_setup(keys, original):
```
grondo, line 56: commit message: is "inputted" a word?

grondo (approved):
LGTM. A couple nits inline.

src/bindings/python/flux/job/kvslookup.py
garlick: https://english.stackexchange.com/questions/30684/inputted-or-input

chu11: LOL

https://www.merriam-webster.com/grammar/is-inputted-a-word

> the issue of how to treat the past participle is a contentious one, with much blood being shed on both sides.

But it seems folks have accepted as a word

:shrug:

grondo: :+1: However from now on I will think of it as a golf term.

Issue #5400, opened 2023-08-17
pub/sub job state transition notification is little used


garlick:
Problem: the job manager publishes event messages when jobs transition states.  AFACT nothing subscribes to these messages as they don't provide sufficient detail for real world use cases.

As an experiment I disabled this feature and all our tests continue to pass, except for `t2207-job-manager-bulk-state.t` which specifically ensures that the event messages are published.

garlick:
Correction, `flux-top` uses this event as trigger to go fetch data from `job-list`.  Hmm, well we might want to consider another solution for that given that unconditionally publishing state transition information is pretty disruptive, for little gain.

Anyway, low priority.

chu11:
> Correction, flux-top uses this event as trigger to go fetch data from job-list. Hmm, well we might want to consider another solution for that given that unconditionally publishing state transition information is pretty disruptive, for little gain.

I'm suddenly reminded of this prior conversation

https://github.com/flux-framework/flux-core/issues/4569

could make the events journal stream-able for users? or provide a slightly more stream-lined job-list states streaming RPC for flux-top and future tools?

garlick:
Oh thanks for that reference.

The thought I was having was maybe a streaming version of the  `job-list.job-stats` RPC with a configurable rate limit?  E.g. send a response each time there is a change but no more frequently than the requested limit?

garlick:
Well duh, `flux-top` already polls `job-list.job-stats` every hearbeat, so might as well use a change in stats to trigger the `job-list.list` query.

grondo:
A `job-list.job-stats` streaming RPC variant that only sends updates when something has changed does sound useful though!

garlick:
True and that would eliminate all the heartbeat-based based polling in `flux-top` AFAICT.

chu11:
> A job-list.job-stats streaming RPC variant that only sends updates when something has changed does sound useful though!

On top of this, the streaming RPC could send stats changes per-queue, so the noise could be even lower when users only list specific queues.

garlick:
Linking some history
- #3254
- #2109
- #2094

PR #5430, opened 2023-09-06
job-list: support streaming job-stats RPC

garlick:
As discussed in #5400, a streaming version of `job-list.job-stats` would be useful and could let us deprecate the `job-stats` events, reducing job manager complexity and background noise on the overlay network.

This modifies `job-list.job-stats` so if the FLUX_RPC_STREAMING flag is used, responses are sent whenever job stats change (but no more frequently than every 200ms), and switches `flux-top(1)` to use the streaming responses to detect job activity in place of subscribing to `job-stats` events.

Marked as WIP pending the addition of tests and any initial feedback.

### DIFF
--- a/src/modules/job-list/job-list.c
+++ b/src/modules/job-list/job-list.c
@@ -95,6 +95,15 @@ static void purge_cb (flux_t *h,
     flux_log (h, LOG_DEBUG, "purged %d inactive jobs", count);
 }
 
+static void disconnect_cb (flux_t *h,
+                           flux_msg_handler_t *mh,
+                           const flux_msg_t *msg,
+                           void *arg)
+{
+    struct list_ctx *ctx = arg;
+    job_stats_disconnect (ctx->jsctx->statsctx, msg);
+}
+
 static const struct flux_msg_handler_spec htab[] = {
     { .typemask     = FLUX_MSGTYPE_REQUEST,
       .topic_glob   = "job-list.list",
@@ -130,6 +139,11 @@ static const struct flux_msg_handler_spec htab[] = {
       .topic_glob   = "job-purge-inactive",
       .cb           = purge_cb,
       .rolemask     = 0
+    },
+    { .typemask     = FLUX_MSGTYPE_REQUEST,
+      .topic_glob   = "job-list.disconnect",
+      .cb           = disconnect_cb,
+      .rolemask     = FLUX_ROLE_USER,
     },
     FLUX_MSGHANDLER_TABLE_END,
 };
--- a/src/modules/job-list/job-list.c
+++ b/src/modules/job-list/job-list.c
@@ -65,23 +65,6 @@ error:
         flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
 }
 
-static void job_stats_cb (flux_t *h, flux_msg_handler_t *mh,
-                          const flux_msg_t *msg, void *arg)
-{
-    struct list_ctx *ctx = arg;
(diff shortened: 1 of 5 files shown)


grondo:
Gave this a whirl and it is is pretty neat because the `flux-top` stats update in near-realtime when I'm running a high throughput test...

grondo:
One isle question: is 200ms too small an interval for a busy system instance with dozens of users using flux-top(1)?

garlick:
I think it's probably OK :shrug:.  I tend to think of the system instance as not high throughput.  Even if there were 50 users running `flux-top` and multiple jobs turning over per second, it's still be on the order of 250 messages per second which isn't crazy.

garlick:
Add a sharness test and dropped the WIP.

Should the top changes and the job-list changes be split into two PRs?

src/modules/job-list/stats.c
```diff
@@ -17,6 +17,7 @@
 
 #include "ccan/str/str.h"
 #include "src/common/libczmqcontainers/czmq_containers.h"
+#include "src/common/libutil/errno_safe.h"
```
chu11, line 20: commit message, did you mean "split out the response portion"?

src/modules/job-list/job-list.c
```diff
@@ -131,6 +140,11 @@ static const struct flux_msg_handler_spec htab[] = {
       .cb           = purge_cb,
       .rolemask     = 0
     },
+    { .typemask     = FLUX_MSGTYPE_REQUEST,
+      .topic_glob   = "job-list.disconnect",
+      .cb           = disconnect_cb,
+      .rolemask     = FLUX_ROLE_USER,
+    },
```
chu11, line 149: i suppose its not a big deal, but wanted to double check you intentionally didn't want to add a "cancel stream" RPC target?  We could always add later if we need to, there's no real need for it at the moment.

t/t2262-job-list-stats.t
```diff
@@ -0,0 +1,40 @@
...
+	EOT
+	chmod +x job-stats.py
+'
+test_expect_success 'streaming job-stats returns one response immediately' '
+	run_timeout 30 flux python job-stats.py 1
+'
+test_expect_success 'streaming job-stats returns at least 2 responses when a job runs' '
+	flux submit sleep 1 &&
+	run_timeout 30 flux python job-stats.py 2
+'
```
chu11, line 32: i can't help but look at this and think it's a little racy, like if the system was a little slow to launch `job-stats.py`, the sleep could finish before the job stats are run.

Maybe run job-stats.py in the background, run job to completion, and then make sure background process output 2 lines of stats?

chu11 (approved):
LGTM, just some small nits and maybe a race in one test.

src/modules/job-list/job-list.c
garlick: Right, no use case at the moment so I didn't add that.

t/t2262-job-list-stats.t
garlick: Yeah, let me work on that a bit.

garlick:
Fixed the bad commit message and reworked the test.

garlick:
OK to set MWP @chu11?  I didn't realize until just now that you had approved this.

garlick:
Hmm, actually I think for the release notes, the top changes ought to be a separate PR so I'll drop those.

garlick:
I'll go ahead and set MWP and then resubmit the `flux-top(1)` bits as a new PR.

PR #5431, opened 2023-09-07
testsuite: increase sleep time in tests

chu11:
Problem: Every once in awhile t2602-job-shell.t fails with:

FAIL: t2602-job-shell.t 22 - job-shell: test shell kill event handling: numeric signal

On a busy system.

The test checks for a non-zero status but the logs show

1694098677.862467 finish status=0

The suspicion is that the job, which is only 60 seconds long, may actually run to completion before the signal is sent / received on a busy system.

Solution: Increase the job from 60 seconds to 300 seconds.


garlick (approved):
Seems harmless and if it helps, great!

PR #5432, opened 2023-09-07
flux-top: use streaming job-stats RPC

garlick:
Problem: flux-top(1) subscribes to job-state events to detect job activity and trigger other queries, but now that the `job-stats` RPC is streamable, there is no need.
    
Drop the job-state subscription and instead drive queries with streaming `job-stats` responses.


chu11 (approved):
LGTM!

PR #5437, opened 2023-09-09
job-manager: correct fsd output in error message

### LINKED ISSUE #5414: requested duration exceeds policy limit of 3e+01m
Problem:  with `policy.limits.duration = "30m"`, a job request that exceeds the limit gets an unfortunate error message with the limit FSD expressed as `3e+01m`.

```
$ flux submit -q debug -t 31m true
[Errno 22] requested duration exceeds policy limit of 3e+01m
```

Possibly our FSD formatter could use some fine tuning.

chu11:
Problem: An error message in the limit-duration plugin uses the output from fsd_format_duration_ex() with a precision of 1.  The precision of 1 leads to the average output to be in exponent form. For example, "30m" is output as "3e+01m".

Solution: Increase the precision from 1 to 2, which should make the average output look fine.

Fixes #5414


src/modules/job-manager/plugins/limit-duration.c
```diff
@@ -200,7 +200,7 @@ static int check_limit (struct limit_duration *ctx,
         && limit != DURATION_UNLIMITED
```
garlick, line 200: Commit message:
Using the term "average" when talking about numerical precision is a bit confusing.  I assume you meant something like "typical"?

garlick (approved):
LGTM, just one nit about the commit message.

chu11:
re-pushed with tweak to commit message, setting MWP

PR #5438, opened 2023-09-10
testsuite: handle job signal race in more tests

chu11:
Same fix as in #5302 and #5401 

also threw a increased timeout length for a test

### DIFF
--- a/t/t2611-debug-emulate.t
+++ b/t/t2611-debug-emulate.t
@@ -13,7 +13,7 @@ stop_tasks_test()     {
 
 test_under_flux 2
 
-TIMEOUT=10
+TIMEOUT=100
 
 parse_jobid() {
 	outfile=$1 &&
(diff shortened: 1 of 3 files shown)


chu11:
re-pushed, with an extra timeout increase in a test

t/t2611-debug-emulate.t
```diff
@@ -13,7 +13,7 @@ stop_tasks_test()     {
 
 test_under_flux 2
 
-TIMEOUT=10
+TIMEOUT=100
```
grondo, line 16: The test at the end expects to fail with a timeout, so we don't want to blindly increase the timeout 10x in this test. (Did this actually succeed for you?)

The last test has:
```sh
        test_must_fail flux job wait-event -vt ${TIMEOUT} ${jobid} finish &&
```
In order to check that the job does _not_ finish after attach by `flux job attach --debug`. If you increase the timeout to 100 then the test will take a minimum of 100 seconds to run... and other timeouts may come into play ...

grondo (commented):
One comment inline (probably the reason CI is failing...)

t/t2611-debug-emulate.t
chu11: hmmm, well it's failing for me now.  I have no idea how this passed for me before.  I clearly did something stupid before.  But yeah, lets update that one line to use a timeout of 2.

chu11:
re-pushed with the suggested tweak from @grondo above, thanks!

grondo (approved):
Ok, LGTM!

PR #5439, opened 2023-09-10
libflux: drop flux_panic()

garlick:
Problem: flux_panic() is no longer needed.

This is left over from early broker prototyping and has outlived its usefulness.

Drop API call, broker RPC handler, and tests.


chu11 (approved):
agreed!

garlick:
Thanks!  I just started one of the ci builders that hit the faketime failure that I think we've seen very occasionally before, and I'll set MWP.

Issue #5440, opened 2023-09-10
src/modules: module arg parsing


chu11:
While reviewing #5435 I noticed that the PR returns an error/-1 if the module load has a argument that is unknown.

In my mind, I thought most modules ignored unknown args .... after a tiny bit of auditing, it appears that most modules return -1 on an unknown arg.  While looking at modules, `kvs`, and `cron` seem to ignore unknown args.

Atleast those two modules should be updated to return -1 on error to be consistent with the other modules.

PR #5442, opened 2023-09-11
modules: consistently return error on invalid module arguments

chu11:
Per #5440, noticed there was inconsistency where some modules returned error on invalid module args, while others ignored invalid args.

Also noticed a few modules did not set errno = EINVAL when an invalid arg was specified.


garlick (approved):
LGTM!

PR #5446, opened 2023-09-11
flux-keygen: drop libsodium requirement

### LINKED ISSUE #5445: drop libsodium-dev requirement
We currently only use libsodium-dev in `flux-keygen` to add the sodium version to the cert metadata, but according to @garlick in #5444 that data is not used anymore.

We should pull out the sodium bits from `flux-keygen` and drop the libsodium requirement.

garlick:
Problem: libsodium is required by flux-keygen but it is not checked in configure.ac.

All we do with libsodium at this point is add its version to certicate metadata.  This was just "provenance", never actually used for anything, and not worth the extra dependency.  Rather than fix configure.ac, drop the libsodium version from flux-keygen(1) certificates.

Fixes #5445

### DIFF
--- a/src/cmd/flux-keygen.c
+++ b/src/cmd/flux-keygen.c
@@ -15,7 +15,6 @@
 #include <flux/core.h>
 #include <flux/optparse.h>
 #include <czmq.h>
-#include <sodium.h>
 
 #include "src/common/libutil/log.h"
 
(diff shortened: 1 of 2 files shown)


grondo (approved):
Thanks! This LGTM.

I did also notice we have `libsodium-dev` in:
 - `debian/control`
 - `etc/docker/ubuntu/install_deps.sh`
from which it could now be removed.

Also, we could remove it from all the Dockerfiles, but not worry about updating the images until the next time we have a new dep that forces us to do so. (?)

grondo:
> Edit: also flux-security requires it, so we shouldn't remove it from images where that is built.

Oh yeah, :facepalm: ! Thanks!

src/cmd/flux-keygen.c
```diff
@@ -15,7 +15,6 @@
 #include <flux/core.h>
 #include <flux/optparse.h>
 #include <czmq.h>
-#include <sodium.h>
```
grondo, line 18: Commit message typo: certicate

grondo (approved):
LGTM! Thanks for taking care of this!

garlick:
Thanks!  fixed typo and rebased on current master.  I'll set MWP.

Issue #5452, opened 2023-09-14
`/etc/flux/rc1` fails with >1 file in `/etc/flux/rc1.d`


grondo:
The flux-sched `test-install` tests were failing with the error:
```
Sep 14 00:51:42.030111 broker.err[0]: rc1.0: /etc/flux/rc1.d/02-cron: No such file or directory
```

To reproduce, I launched the flux-sched test container with:
```console
$ src/test/docker/docker-run-tests.sh -i bookworm -I
```
and then built and installed flux-sched configured with `--prefix=/usr --sysconfdir=/etc`.

By adding `set -x` to `rc1` I think the problem is occurring in the new `listfiles()` function, which appears to quote the entire result of the glob, so that the script is trying to run
```
"etc/flux/rc1.d/01-sched-fluxion
/etc/flux/rc1.d/02-cron"
```

I _think_ the problem is that `IFS` is still set to `:` before the for loop. Removing that at least fixes the issue...

grondo:
Oh, the `IFS=:` is probably necessary because we're iterating the `FLUX_RC_EXTRA`. Maybe more digging needed here for a fix.

grondo:
Ok, not listing the rc files in a shell function seems to work:
```diff
diff --git a/etc/rc1 b/etc/rc1
index 1b23fd397..cddac74c2 100755
--- a/etc/rc1
+++ b/etc/rc1
@@ -3,14 +3,6 @@
 # Allow connector-local more time to start listening on socket
 RANK=$(FLUX_LOCAL_CONNECTOR_RETRY_COUNT=30 flux getattr rank)
 
-listfiles() {
-    local glob="$1"
-    for file in $glob; do
-        [ $file = "$glob" ] && continue
-        echo $file
-    done
-}
-
 # Usage: modload {all|<rank>} modname [args ...]
 modload() {
     local where=$1; shift
@@ -101,7 +93,8 @@ core_dir=$(cd ${0%/*} && pwd -P)
 all_dirs=$core_dir${FLUX_RC_EXTRA:+":$FLUX_RC_EXTRA"}
 IFS=:
 for rcdir in $all_dirs; do
-    for rcfile in $(listfiles "$rcdir/rc1.d/*"); do
+    for rcfile in $rcdir/rc1.d/*; do
+       [ -e $rcfile ] || continue
        echo running $rcfile
         $rcfile
     done
```

If @garlick approves I can propose a PR, still not sure if I'm missing something here.

PR #5453, opened 2023-09-14
rc: fix `flux start` failure when multiple scripts are present in `rc1.d`

grondo:
This is a potential fix for #5452.

### DIFF
--- a/etc/rc1
+++ b/etc/rc1
@@ -3,14 +3,6 @@
 # Allow connector-local more time to start listening on socket
 RANK=$(FLUX_LOCAL_CONNECTOR_RETRY_COUNT=30 flux getattr rank)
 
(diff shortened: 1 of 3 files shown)


garlick (approved):
Wow I really blew it there!  Thanks for catching it and especially for the test!

grondo:
Pushed one more test to check that an empty `rc1.d` directory works as well.

grondo:
If the new test looks ok I'll set MWP here. Thanks!

garlick (approved):
There's an inception builder failure I didn't look into but the test itself looks fine to me.

grondo:
I couldn't even figure out what went wrong on that one. Restarted and it seems to have passed this time, so setting MWP.

Issue #5441, opened 2023-09-11
idea: broker module dependency management


garlick:
With the new #5435 PR that will create more module dependency on each other, was just brainstorming this a bit.

Original thoughts above are related to some service that would alert kvs-watch module that kvs module has been removed, so kvs-watch module could exit, re-subscribe to kvs events, etc. etc.  That's certainly more difficult and a lot of work.

But something that is easier, might just be a minor addition to the module stuffs where a module can indicate "I am dependent on module kvs".  The dumb approach would be some `const char *mod_dependency` comma separated string or similar array of sorts.  Maybe I can imagine something better later.

With this information, if we try to remove the kvs module, `flux module remove kvs` could output a warning or error of some sort saying "kvs-watch needs this".  There could be a "flux module remove --force kvs" to force the remove.

Similarly we can output a warning/error with `flux module load` if a dependency is not yet loaded.  I think this could be quite useful.  I know during test development I've had to load/reload modules and am often confused on the order of loads.  Trial and error and digging through `rc1` or other tests is usually what helps me figure out the order of modules to load/re-load.

Such information would also allow `flux module list` to list dependencies of modules.  Sorta like `lsmod`.

_Originally posted by @chu11 in https://github.com/flux-framework/flux-core/issues/4333#issuecomment-1712892246_

garlick:
> The dumb approach would be some `const char *mod_dependency` comma separated string or similar array of sorts.

I think modules declaring variables for access using `dlsym()` is a problem for some languages so that may not be the way to go.

Since we're brainstorming, here's a thought.  What if we added a "built in service" (broker/modservices.c) called `NAME.incref` that accepts a streaming RPC.  When a request comes in, a reference count is take on the module, and while the reference count is > 0, the module cannot be unloaded unless forced?  When the RPC is canceled or a disconnect is received, the module use count is decremented. 

On the client side, we could provide a function that takes a list of services and blocks until it gets responses to all `NAME.incref` requests.  If any of the responses are ENOSYS, then the module fails before entering the reactor and `flux module load` fails.  If a module dependency is forceably unloaded, the streaming RPC would get an ENOSYS response at that point, which could serve as a notification that something went wrong (perhaps the module could optionally register a handler for this).

This isn't quite as sophisticated as something like the linux kernel `modprobe` command where dependencies are automatically fulfilled, but it does solve the problem you called out and does so without assuming which module provides a given service or whether that module is loaded on the local rank or upstream.

[...]

PR #5454, opened 2023-09-15
broker: stop managing 0MQ sockets with czmq

garlick:
Problem: the czmq `atexit()` handler has gotten in the way for years.

A current example: If I load a module and forget to unload it, the error message is unhelpful and the broker exits with success:
```
$ flux start
ƒ(s=1,d=0) $ flux module load --name foo barrier
ƒ(s=1,d=0) $ 
exit
E: (flux-broker) 23-09-15 21:22:51 [488322]dangling 'PAIR' socket created at shmem/shmem.c:148
$ echo $?
0
```

This PR curtails use of the czmq "socket tracker" throughout flux-core by creating all sockets with `zmq_socket()` rather than the tracked `zsock_new*()` series of functions. Since many czmq classes accept a socket argument in either `zsock_t` (tracked) or raw  form, this change has a smaller footprint than it would have otherwise.

Now we get
```
$ ./flux start
ƒ(s=1,d=0,builddir) $ flux module load --name foo barrier
ƒ(s=1,d=0,builddir) $ 
exit
flux-broker: broker module 'foo' was not properly shut down
flux-broker: skipping 0MQ shutdown due to presumed module socket leak
$ echo $?
1
```

### DIFF
--- a/src/broker/module.c
+++ b/src/broker/module.c
@@ -372,10 +372,11 @@ module_t *module_create (flux_t *h,
     }
     zsock_set_unbounded (p->sock);
     zsock_set_linger (p->sock, 5);
-    snprintf (p->endpoint,
-              sizeof (p->endpoint),
-              "inproc://%s",
-              module_get_uuid (p));
+    // copying 9 + 37 + 1 = 47 bytes into 128 byte buffer cannot fail
+    (void)snprintf (p->endpoint,
+                    sizeof (p->endpoint),
+                    "inproc://%s",
+                    module_get_uuid (p));
     if (zmq_bind (p->sock, p->endpoint) < 0) {
         errprintf (error, "zmq_bind %s: %s", p->endpoint, strerror (errno));
         goto cleanup;
--- a/src/broker/module.c
+++ b/src/broker/module.c
@@ -44,7 +44,8 @@ struct broker_module {
     double lastseen;
 
     void *zctx;             /* zeromq context shared with broker */
-    zsock_t *sock;          /* broker end of PAIR socket */
+    void *sock;             /* broker end of PAIR socket */
+    char endpoint[128];
     struct flux_msg_cred cred; /* cred of connection */
 
     uuid_t uuid;            /* uuid for unique request sender identity */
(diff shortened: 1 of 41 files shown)


grondo:
Wow, amazing! I thought this would be so much work

garlick:
Hmm got a unit test failure in `ci / bookworm - py3.11,clang-15`.  I did change this test although I'm not spotting anything wrong.  Will retry that one.
```
2023-09-15T20:36:48.9534284Z Making check in libterminus
2023-09-15T20:36:48.9677549Z make  test_pty.t test_terminus.t
2023-09-15T20:36:49.1720481Z   CC       test/terminus_t-terminus.o
2023-09-15T20:36:49.1967837Z   CC       test/pty_t-pty.o
2023-09-15T20:36:49.4507457Z   CCLD     test_terminus.t
2023-09-15T20:36:49.4683243Z   CCLD     test_pty.t
2023-09-15T20:36:49.4686534Z make  check-TESTS
2023-09-15T20:36:49.5699731Z                 test_pty.t:  PASS: N=81  PASS=81  FAIL=0 SKIP=0 XPASS=0 XFAIL=0
2023-09-15T20:36:49.7769938Z            test_terminus.t:  SKIP: N=0   PASS=0   FAIL=0 SKIP=0 XPASS=0 XFAIL=0
2023-09-15T20:36:49.7770810Z ERROR: test_terminus.t - missing test plan
2023-09-15T20:36:49.7771354Z ERROR: test_terminus.t - exited with status 139 (terminated by signal 11?)
```

garlick:
Force pushed with some (mainly cosmetic) cleanup.

src/broker/module.c
```diff
@@ -354,14 +355,18 @@ module_t *module_create (flux_t *h,
...
         goto cleanup;
     }
     zsock_set_unbounded (p->sock);
     zsock_set_linger (p->sock, 5);
-    if (zsock_bind (p->sock, "inproc://%s", module_get_uuid (p)) < 0) {
-        errprintf (error, "zsock_bind inproc://%s", module_get_uuid (p));
+    snprintf (p->endpoint,
+              sizeof (p->endpoint),
+              "inproc://%s",
+              module_get_uuid (p));
```
grondo, line 367: A suggestion to cast to `void` and add a note that maximum size of string can never overflow 127 chars... But up to you.

grondo (approved):
LGTM! Tested on a single machine and also did some multi-node runs to ensure no issues.

A couple very minor suggestions inline but feel free to take them or leave them.

On slightly related note (and hopefully this is not a dumb question): Is there a reason the broker can't force unload remaining modules itself during shutdown as it sees them still loaded? E.g. loading a module by hand for testing, then forgetting to call `flux module unload` before `exit` seems like unfair grounds for complete failure of the instance...

garlick:
> Is there a reason the broker can't force unload remaining modules itself during shutdown as it sees them still loaded?
E.g. loading a module by hand for testing, then forgetting to call flux module unload before exit seems like unfair grounds for complete failure of the instance...

Just to back up a little bit, are you in agreement with the change made by this PR of doing an exit 1 when a module is `pthread_cancel()ed`?  I thought it was more correct than before, since memory could be leaked or state could be lost if a module would normally state during a clean shutdown.  Before, we got the czmq atexit warnings but the exit code was 0.

Hmm, we could certainly add code to try to cleanly unload modules after the broker reactor returns, perhaps under a timeout.  Maybe a new PR?  Have to think about that one a bit...  (would it encourage a dependency on auto unloading, and then cause irritation when it doesn't know about module interrelationships?  Should it be considered along with #5441?)

grondo:
Yeah, I agree with the PR as is, sorry didn't mean to question that! It was more of a general question that the content of this PR made me think about... Sounds better as a separate issue or PR, bit I was kind of thinking there might be a simple answer...

garlick:
fixed those things and forced a push.  Thanks!

garlick:
Setting MWP

Issue #5455, opened 2023-09-19
Build failure Python 3.12


haampie:
https://github.com/flux-framework/flux-core/blob/a00922dc873ed4bfa303e1edd677d24832e5b1c1/config/am_check_pymod.m4#L18

`distutils` is removed in Python 3.12, see https://peps.python.org/pep-0632/#migration-advice

grondo:
Also looks like we may need to update `ax_python_devel.m4` to fix use of `distutils`.

PR #5456, opened 2023-09-19
configure: avoid use of deprecated python distutils module

grondo:
This is an attempted fix for #5455. It switches to `packaging.version.Version` instead of `StrictVersion` in our custom `AM_CHECK_PYMOD` macro, and updates the autoconf-archive `ax_python_devel.m4` to the latest available, which _appears_ to have a fix for use of the deprecated `distutils` module.

### DIFF
--- a/config/am_check_pymod.m4
+++ b/config/am_check_pymod.m4
@@ -15,7 +15,7 @@ except:
         sys.exit(0)
 sys.exit(0)"], [prog="
 import sys
-from distutils.version import LooseVersion, StrictVersion
+from packaging.version import Version
 import $1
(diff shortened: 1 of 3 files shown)


chu11 (approved):
this change looks good to me.  Dunno if we want to add a TODO for py 3.12 builder variant?  My minimal googling says it'll be released early next month.

grondo:
Thanks!

>  Dunno if we want to add a TODO for py 3.12 builder variant? 

We'll probably pick this up when we update to the next Fedora. Not sure we need to rush it at this point...

Issue #5457, opened 2023-09-19
checking for cffi.__version_info__ >= (1,1) in python module cffi... no


chu11:
```
checking for /usr/bin/python3.6 version... 3.6
checking for /usr/bin/python3.6 platform... linux
checking for /usr/bin/python3.6 script directory... ${prefix}/lib/python3.6/site-packages
checking for /usr/bin/python3.6 extension module directory... ${exec_prefix}/lib64/python3.6/site-packages
checking for cffi.__version_info__ >= (1,1) in python module cffi... no
configure: error: could not find python module cffi, version 1.1+ required
make: *** [Makefile:549: config.status] Error 1
```

ugh, this just started showing up for me on TOSS4 on corona.  I assume the distutils fix did something.  looking into it.

chu11:
i guess `packaging` is not a native python lib/module?  extra?

```
>python3.6
Python 3.6.8 (default, May 31 2023, 10:28:59) 
[GCC 8.5.0 20210514 (Red Hat 8.5.0-18)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import packaging
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'packaging'
>>> import distutils
>>> 
```

chu11:
yup, looks like that was the issue.  will poke admins to install (ahem) "packaging".

grondo:
Sorry, this was my bad. I assumed `packaging` was in the standard library because it is the suggested replacement for `distutils.version` in [PEP 0632](https://peps.python.org/pep-0632/). Apparently they thought it was OK because everyone can just pip anyway so what's the big deal.

We should avoid adding this new requesite package and instead find a way to do the simple version compare within the standard library (or just do something by hand)

grondo:
Looking around it seems like every fix for deprecated `distutils.version` pulls in `packaging.version`, even though it isn't in the standard library. So I guess soon the `packaging` module will be installed everywhere, but since that isn't the case now, the best solution might be to just copy  `distutils` StrictVersion implementation into `am_check_pymod.m4`. :shrug:

chu11:
just a reminder todo to add python3-packaging (or appropriate package name for distro) to the `scripts/install` scripts.

grondo:
[...]

PR #5459, opened 2023-09-20
configure: use distutils if available to avoid extra python module dependency

grondo:
Problem: The Python packaging module is not standard, but configure uses it to compare versions, leading to an implicit new dependency.

Since distutils is not removed until Python 3.12, try distutils first then fall back to the packaging package. This will allow configure to continue to work on existing systems while also not breaking on Python 3.12.

Note: this is a proposed workaround for #5457. It will keep our current systems working without a new dependency, but will require Python 3.12 users to install the `packaging` module. :shrug:

### DIFF
--- a/config/am_check_pymod.m4
+++ b/config/am_check_pymod.m4
@@ -15,7 +15,10 @@ except:
         sys.exit(0)
 sys.exit(0)"], [prog="
 import sys
-from packaging.version import Version
+try:
+    from distutils.version import StrictVersion as Version
+except ModuleNotFoundError:
+    from packaging.version import Version
 import $1
 if not $2:
     sys.exit(1)


chu11 (approved):
LGTM, just tried it out and appears to work

grondo:
What do you think about the implied new `packaging` dependency under Python 3.12, though? Acceptable for now?

chu11:
> What do you think about the implied new packaging dependency under Python 3.12, though? Acceptable for now?

Certainly acceptable for now.  It is certainly annoying that we have this dependency for Python 3.12 for only a configure check.

I guess the only question is this dependency worth it over the annoyance of having an internal implementation (possibly copied in counting as "internal").  We could certainly whip something up that would work with what we currently check for, but we take on the risk of any future python module requirements that have corner cases.  So I'd lean towards just making it a requirement for Python 3.12.

FWIW, https://github.com/pypa/packaging/blob/main/src/packaging/version.py is about 500 lines.  And if we went through it I'm sure we could yoink some stuff out.  but it does seem to handle a nice chunk of corner cases, like when letters are in the version number.

grondo:
`distutils.version.StrictVersion` is only ~50 lines, but it was very difficult to incorporate it into m4 so I gave up.

PR #5461, opened 2023-09-21
libzmqutil: add cert class and use it instead of CZMQ zcert_t

garlick:
This adds a `zmqutil_cert` class to replace CZMQ `zcert_t` and completes the conversion away from libczmq.

The certificate format does not change.


garlick:
Fixed a TOCTOU called out by asan, and a spell error caught by the spell checker.  Yay CI!

src/cmd/flux-keygen.c
```diff
@@ -93,37 +117,34 @@
...
+                  "keygen.zmq-version",
+                  "%d.%d.%d",
+                  ZMQ_VERSION_MAJOR,
+                  ZMQ_VERSION_MINOR,
+                  ZMQ_VERSION_PATCH);
+
+    if (path) {
+        int fd;
+        FILE *f;
+        if (!(fd = open (path, O_WRONLY | O_CREAT | O_TRUNC, 0600))
```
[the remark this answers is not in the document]
garlick: Hmm, I'm not sure I agree with this one.  Basically it wants me to sanitize `path` before opening it for writing in `flux-keygen` but it sounds like they think it's a privileged command or something which it is not.  Any thoughts?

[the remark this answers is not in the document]
grondo: Yeah, this seems a little silly. Will the code scanner call out every program that takes an argument to open a file? (e.g. `vi`?) Maybe @dun would have some insight, but I think this could be dismissed as a false positive.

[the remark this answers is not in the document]
garlick: I wonder if it would be a good idea to define this cert format in a Flux RFC.  :thinking:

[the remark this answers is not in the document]
dun: If `flux-keygen` is running as the user, I would think this would be equivalent to what they could do at the shell. So I'm not sure what the benefit would be for additional sanitization.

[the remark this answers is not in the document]
garlick: Thanks @dun!  OK, I'll dismiss this one then.

src/common/libzmqutil/cert.c
```diff
@@ -0,0 +1,396 @@
...
+#include "src/common/libutil/strstrip.h"
+#include "src/common/libutil/strlcpy.h"
+#include "src/common/libczmqcontainers/czmq_containers.h"
+#include "ccan/str/str.h"
+
+#include "sockopt.h"
+#include "cert.h"
+
+#define KEYSIZE 32
+#define TXTSIZE 41
```
grondo, line 40: A comment of where these sizes are coming from might be helpful. I'm guessing cert keys are limited to 32 bytes (256 bit) and the max expanded size of 32 bytes with z85 is 41 bytes?

garlick:
Chasing an ASAN issue in the cert test and found some more stuff that needs testing along the way, so please hold off reviewing.

src/common/libzmqutil/test/cert.c
```diff
@@ -0,0 +1,308 @@
...
+    " secret-key = \"UPk352x.N&a(0mXR&{9S^<I!cH[Oo*^BeU)UD/[5\"\n"
+    },
+    {
+        .name = "cert with keys not indented",
+        .input =
+    "curve\n"
+    "public-key = \".0TmVO:u.]cUa[tFi[rH<3q2Cmu?1IBq!8/Q+&(0\"\n"
+    "secret-key = \"UPk352x.N&a(0mXR&{9S^<I!cH[Oo*^BeU)UD/[5\"\n"
+    },
+};
```
grondo, line 116: Might be good to expand the "bad" tests here, including something like a "good" cert with one character change (so should fail), some garbage input, public and/or secret key too long, unclosed quote, etc

scripts/install-deps-rpm.sh
```diff
@@ -8,7 +8,6 @@ yum install \
   pkgconfig \
   glibc-devel \
   zeromq4-devel \
-  czmq-devel \
```
grondo, line 11: Remove from Dockerfiles too so the next time we generate images this dependency is dropped?

grondo: Or is czmq still required by some other dependency?

src/common/libzmqutil/cert.h
```diff
@@ -0,0 +1,37 @@
...
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+#ifndef _ZMQUTIL_CERT_H
+#define _ZMQUTIL_CERT_H
+
+#include <stdio.h>
+#include <stdbool.h>
+
+struct cert *zmqutil_cert_create (void);
```
grondo, line 17: It feels a little strange that this class is `zmqutil_cert` but the structure is just `struct cert`. However, it probably doesn't rise to the level of updating the whole PR. (just a comment)

grondo (approved):
This LGTM. Feel free to ignore suggestions inline, though adding the extra bad cert tests might be a good idea before merging.

grondo:
Oops sorry @garlick! I was reviewing and missed your comment to hold off!

garlick:
Oh that's OK,  I haven't changed that much.  Good suggestions!

scripts/install-deps-rpm.sh
garlick: Not that I know of but actually...maybe it makes sense to drop czmq in another PR just for release notes clarity?  This PR Is heavily about the cert class.

src/common/libzmqutil/cert.h
garlick: Yeah and it's a narrowly used internal class so maybe just claiming the "cert_" namespace is OK here.  I can make that change.

garlick:
Just forced a push with a few changes
- rename cert functions from `zmqutil_cert_*()` to `cert_*()`
- add a check that secret and public keys "match" (I didn't realize there was `zmq_curve_public()`)
- fixed the ASAN issue with the test
- add more cert tests (including expanding the bad input as suggested)
- misc other minor cleanup
-  dropped the czmq build changes (will submit as  another PR)

garlick:
@grondo -  I know you already approved but maybe you could give this one more quick look to see if anything pops out?

grondo (approved):
Still LGTM!

PR #5462, opened 2023-09-21
allow hwloc topology to be loaded from a file with FLUX_HWLOC_XMLFILE

grondo:
Running `make -j N check` where `N` is large can cause random and numerous test failures on systems with a lot of cores due to parallel hwloc topology discovery across many tests, each invoking many brokers. Presumably, some kind of kernel locks in sysfs slow down parallel hwloc discovery so much that tests time out and fail in unexpected ways. It is also disappointing that `make -j 32 check` runs slower on a 64 core system than an 8 core system.

Based on an idea by @wihobbs, I found that hwloc supports a `HWLOC_XMLFILE` environment variable that forces `hwloc_topology_load(3)` to use a file instead of reading the real system topology from scratch. However, in practice this environment variable does not seem reliable, and it is pretty easy to create our own similar environment variable.

So, this PR introduces `FLUX_HWLOC_XMLFILE` which works the same as above. When set to a file, `rhwloc_local_topology()` will read from the file instead of `hwloc_topology_load()`. This can really speed up `flux start -s N` on some systems (especially tioga for some reason) or if you are running `make -j N check` since this will also result in many brokers doing "discovery" simultaneously on a system.

This is meant to be used in situations where there is a known problem. The testsuite can be prefaced with:
```console
$ lstopo --of xml > system.xml
$ export FLUX_HWLOC_XMLFILE=$(pwd)/system.xml
$ make -j 32 check
```
Or this variable can be used if you're having trouble running `flux start` with large `--test-size`.

Additionally, this option could be used to start a test instance using hwloc from a different machine or a fake topology just using the XML.

Also included in this PR are a couple related fixes that seemed to be required to get `make -j 32 check` to work reliably on tioga.

### DIFF
--- a/src/common/librlist/rhwloc.c
+++ b/src/common/librlist/rhwloc.c
@@ -12,9 +12,15 @@
 #include "config.h"
 #endif
 
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+
 #include <flux/idset.h>
 
 #include "ccan/str/str.h"
+#include "src/common/libutil/read_all.h"
+#include "src/common/libutil/errno_safe.h"
 
 #include "rhwloc.h"
 
@@ -67,15 +73,41 @@ hwloc_topology_t rhwloc_xml_topology_load (const char *xml)
     return topo;
 }
 
+hwloc_topology_t rhwloc_xml_topology_load_file (const char *path)
+{
+    hwloc_topology_t topo;
+    int fd;
+    int rc;
+    char *buf;
+    if ((fd = open (path, O_RDONLY)) < 0)
+        return NULL;
+    rc = read_all (fd, (void **) &buf);
+    ERRNO_SAFE_WRAP (close, fd);
+    if (rc < 0)
+        return NULL;
+    topo = rhwloc_xml_topology_load (buf);
+    ERRNO_SAFE_WRAP (free, buf);
+    return topo;
+}
 
 hwloc_topology_t rhwloc_local_topology_load (rhwloc_flags_t flags)
 {
+    const char *xml;
     hwloc_topology_t topo = NULL;
     hwloc_bitmap_t rset = NULL;
     uint32_t hwloc_version = hwloc_get_api_version ();
 
     if ((hwloc_version >> 16) != (HWLOC_API_VERSION >> 16))
         return NULL;
+
+    /*  Allow FLUX_HWLOC_XMLFILE to force loading topology from a file
+     *  instead of the system. This is meant for testing usage only.
+     *  If loading from the XML file fails for any reason, fall back
+     *  to normal topology load.
+     */
+    if ((xml = getenv ("FLUX_HWLOC_XMLFILE"))
+        && (topo = rhwloc_xml_topology_load_file (xml)))
+        return topo;
 
     if (topo_init_common (&topo) < 0)
         goto err;
--- a/src/common/librlist/rhwloc.c
+++ b/src/common/librlist/rhwloc.c
@@ -15,6 +15,7 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <fcntl.h>
+#include <sys/utsname.h>
 
 #include <flux/idset.h>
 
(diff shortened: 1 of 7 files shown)


wihobbs:
Thought I'd link [what you sent me in the hwloc docs](https://www.open-mpi.org/projects/hwloc/doc/v2.3.0/a00359.php#faq_xml). This is so useful, thanks, @grondo!

chu11:
This is awesome.

> Presumably, some kind of kernel locks in sysfs slow down parallel hwloc discovery so much that tests time out and fail in unexpected ways.

Ya know, I did notice that `make -j16 check` was slower after some system update ... i'm guessing that those extra locks are what was slowing it down.

grondo:
No system is quite as bad as tioga (which is currently down so I can't show the difference), but for example on mammoth, a system with 128 cores, it takes half as long to run a 64 broker test instance using an XML file vs not:
```console
$ time src/cmd/flux start -s 64 true
real	0m19.548s
user	0m47.646s
sys	3m27.748s

$ time FLUX_HWLOC_XMLFILE=$(pwd)/mammoth.xml src/cmd/flux start -s 64 true
real	0m10.964s
user	0m46.722s
sys	0m51.130s
```

You can really see the impact trying to run `hwloc_topology_load()` simultaneously by using `flux R encode --local` and `pdsh`, e.g.
```console
$ time src/cmd/flux R encode --local
{"version": 1, "execution": {"R_lite": [{"rank": "0", "children": {"core": "0-127"}}], "starttime": 0.0, "expiration": 0.0, "nodelist": ["mammoth35"]}}
real	0m0.276s
user	0m0.064s
sys	0m0.187s

$ time pdsh -R exec -w [0-127] src/cmd/flux R encode --local | dshbak -c
----------------
[0-127]
----------------
{"version": 1, "execution": {"R_lite": [{"rank": "0", "children": {"core": "0-127"}}], "starttime": 0.0, "expiration": 0.0, "nodelist": ["mammoth35"]}}

real	0m16.577s
user	0m6.990s
sys	8m3.068s

$ time pdsh -R exec -w [0-127] sh -c 'FLUX_HWLOC_XMLFILE=mammoth.xml src/cmd/flux R encode --local' | dshbak -c
----------------
[0-127]
----------------
{"version": 1, "execution": {"R_lite": [{"rank": "0", "children": {"core": "0-127"}}], "starttime": 0.0, "expiration": 0.0, "nodelist": ["mammoth35"]}}

real	0m0.343s
user	0m5.029s
sys	0m5.627s
```

src/common/librlist/rhwloc.c
```diff
@@ -15,6 +15,7 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <fcntl.h>
+#include <sys/utsname.h>
```
chu11, line 18: some typo in commit message "missing to topo"?

grondo: Maybe it needs a reword, but currently there's no typo (trying to keep the subject short probably makes it harder to read): "Add system info if missing to topo loaded from file"

I'll try to reword for clarity.

t/t2800-jobs-instance-info.t
```diff
@@ -22,14 +22,15 @@ waitfile="${SHARNESS_TEST_SRCDIR}/scripts/waitfile.lua"
...
+	id2=$(flux submit -N2 -n2 -c1 flux start \
 		"flux run /bin/false ; \
 		 flux submit --cc=1-4 sleep 300 && \
 		 touch ready && \
 		 flux queue idle") &&
 	flux submit sleep 600 &&
 	flux job wait-event $id clean &&
-	$waitfile -t 60 ready
+	$waitfile -t 60 ready &&
+	flux proxy $id2 flux resource list
```
chu11, line 33: why the additional `flux resource list`?  left over debugging?

chu11 (approved):
awesome!   several runs on corona yielded 100% pass on `make -j24 check`, that hasn't happened in awhile for me.  I usually do `make -j4` nowadays :-)

One nit comment below.

Additional recommendation, document this `t/README` about how this would useful, maybe adding your:

```
lstopo --of xml > system.xml
export FLUX_HWLOC_XMLFILE=$(pwd)/system.xml
make -jBIGNUMBER check
```

example in there too.

t/t2800-jobs-instance-info.t
grondo: Yeah, that should probably be in a `test_debug`

chu11:
update to `t/README` LGTM

grondo:
Thanks! I'll set MWP.

PR #5463, opened 2023-09-21
job-list: support resource-update event

chu11:
This is an early implementation for `resource-update` support in `job-list` for when the `resource-update` event is actually implemented within `job-manager`.  So at the moment this does not do anything.

Note that the last commit has some "XXX" tags for what needs to be done to finish the code, mostly updating tests b/c how to actually update a job's expiration has not yet been determined.

### DIFF
--- a/src/modules/job-list/job_data.c
+++ b/src/modules/job-list/job_data.c
@@ -504,6 +504,29 @@ int job_jobspec_update (struct job *job, json_t *updates)
     return parse_jobspec (job, false);
 }
 
+int job_R_update (struct job *job, json_t *updates)
+{
+    const char *key;
+    json_t *value;
+
+    if (!updates)
+        return 0;
+
+    json_object_foreach (updates, key, value) {
+        /* RFC 21 resource-update event only allows update
+         * to:
+         * - expiration
+         */
+        if (streq (key, "expiration"))
+            if (jpath_set (job->R, "execution.expiration", value) < 0)
+                flux_log (job->h, LOG_INFO,
+                          "%s: job %s failed to update R key %s",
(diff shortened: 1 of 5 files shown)


grondo:
Cool! I was just looking at adding `resource-update` support to the job-manager, at least prototyping something. At first I was thinking that the `update-duration` plugin could just handle updates of duration for a running job by translating it to a `resource-update` event. Some kinks to work out (like the fact that the update service always assumes updates translate to `jobspec-update`) 

Does that sound reasonable? In order to move forward though with expiration update, we'll really need the job-info support for "watching" R. I'm assuming that is already in progress?

chu11:
> Does that sound reasonable? In order to move forward though with expiration update, we'll really need the job-info support for "watching" R. I'm assuming that is already in progress?

Yup, it's mostly done.  Should have PR tomorrow ... and maybe not even WIP, since I don't think I need `flux update` or similar to actually function for those tests.

chu11:
re-pushed and removed WIP, this is now built on top of #5464 so we can use the `resource-update-expiration` plugin to generate a sane test.

I did leave a note in the tests that someday we might want to update to use `flux update` or similar, but this is of course fine for now.

Glad we went through the process of testing this way, found i actually forgot to implement something so this feature actually didn't work :P

chu11:
re-pushed, updating `execution.expiration` to `expiration` per discussion in #5464

garlick (approved):
Not really spotting any problems here, so approving!

grondo:
I'll rebase this then likely set MWP on behalf of @chu11. Thanks!

grondo:
Ok, I've rebased and set MWP.

Issue #5451, opened 2023-09-13
Idea: job-info: add streaming RPC to "watch" R


grondo:
As part of the proposed solution for #4175, `resource-update` events will be posted to the job eventlog to indicate a change in a job's expiration. Interested parties can then monitor the job eventlog to be notified of these updates and adjust accordingly. However, there are multiple components that will require a new eventlog watch RPC, including the job execution service, the job shell, and any scheduler of a Flux subinstance. This is 2-3 new watches on the job eventlog per job. Not to mention the amount of code that will need to be copied to each of these use cases, processing of every eventlog event to check for the rare `resource-update` event, etc.

Each of these entities already fetches R from the job-info service. It might be convenient if the job-info service offered an RPC to start watching R instead of just fetching it. This could replace the existing lookup RPC and in the common case would have just one response. The job-info module would then watch the eventlog of the job and notify the client of updates by sending a new R. Code would be simplified for each use case, and there would only be one new eventlog watch instead of (up to) 3. Plus, I imagine the job-info module already has code for monitoring and processing the job eventlog.

Thoughts?

BTW, as an experiment I added an eventlog watch to the job-exec module for every job. It seemed to impact throughput by 5-8%.

chu11:
I think this is a great idea.

> The job-info module would then watch the eventlog of the job and notify the client of updates by sending a new R.

My first thought was we can just outright watch `R` itself and forward the resulting changes that occur.  `job-info` effectively is convenience on determining the R KVS path and doing a guest access check.  The major difference is just returning an error if R doesn't exist, so it'd be on the caller to only watch R when appropriate.

Dunno if you were thinking of watching the eventlog b/c you want extra smarts in there, i.e. if job not running wait, end stream when job ends, etc.

grondo:
> Dunno if you were thinking of watching the eventlog b/c you want extra smarts in there, i.e. if job not running wait, end stream when job ends, etc.

No, I had thought watching the eventlog is required to process `jobspec-update` and `resource-update` events and send the updated R, jobspec, etc. in the response. I don't think we plan to overwrite R in the KVS on updates at this point (and I'm pretty sure we don't for the jobspec)

grondo:
Not to say we couldn't update R in the kvs when changes occur... Note: I don't think the job manager currently even fetch es a copy of R. I'll have to go review what our current plan was here.

chu11:
[...]

PR #5464, opened 2023-09-22
flux-job: get updated version of R

chu11:
similar to https://github.com/flux-framework/flux-core/pull/5428 this supports `resource-update` events in `flux job info` and the equivalent python bindings to look up the "viewed" R and not the "base" one stored in the KVS.

Just like #5463, the commits with new tests do not work and have XXX tagged to be filled in later when an actual way to update the R of a job exists.

I should also note that this does not use any solution from #5451 as A) I implemented this first :P, B) copied the implementation from the jobspec side, and C) I wanted to keep this processing out of broker.  We could of course change the approach.

### DIFF
--- a/src/bindings/python/flux/job/kvslookup.py
+++ b/src/bindings/python/flux/job/kvslookup.py
@@ -15,6 +15,7 @@ from flux.future import WaitAllFuture
 from flux.job import JobID, JobspecV1
 from flux.job.event import EventLogEvent
 from flux.rpc import RPC
+from flux.util import set_treedict
 
 
 def _decode_field(data, key):
@@ -63,6 +64,10 @@ def _setup_lookup_keys(keys, original, base):
         elif not base:
             if "eventlog" not in keys:
                 keys.append("eventlog")
+    if "R" in keys:
+        if not base:
+            if "eventlog" not in keys:
+                keys.append("eventlog")
 
 
 def _get_original_jobspec(job_data):
@@ -87,7 +92,22 @@ def _get_updated_jobspec(job_data):
     return jobspec.dumps()
 
 
+def _get_updated_R(job_data):
+    if isinstance(job_data["R"], str):
+        R = json.loads(job_data["R"])
+    else:
+        R = job_data["R"]
+    for entry in job_data["eventlog"].splitlines():
+        event = EventLogEvent(entry)
+        if event.name == "resource-update":
+            for key, value in event.context.items():
+                if key == "expiration":
+                    set_treedict(R, "execution.expiration", value)
+    return json.dumps(R, ensure_ascii=False)
+
+
 def _update_keys(job_data, decode, keys, original, base):
+    remove_eventlog = False
     if "jobspec" in keys:
         if original:
             job_data["jobspec"] = _get_original_jobspec(job_data)
@@ -100,7 +120,16 @@ def _update_keys(job_data, decode, keys, original, base):
             if decode:
                 _decode_field(job_data, "jobspec")
             if "eventlog" not in keys:
-                job_data.pop("eventlog")
+                remove_eventlog = True
+    if "R" in keys:
+        if not base:
+            job_data["R"] = _get_updated_R(job_data)
+            if decode:
+                _decode_field(job_data, "R")
+            if "eventlog" not in keys:
+                remove_eventlog = True
+    if remove_eventlog:
+        job_data.pop("eventlog")
 
 
 # jobs_kvs_lookup simple variant for one jobid
@@ -110,7 +139,7 @@ def job_kvs_lookup(
     """
     Lookup job kvs data based on a jobid
 
-    Some keys such as "jobspec" may be altered based on update events
+    Some keys such as "jobspec" or "R" may be altered based on update events
     in the eventlog.  Set 'base' to True to skip these updates and
     read exactly what is in the KVS.
 
@@ -121,7 +150,7 @@ def job_kvs_lookup(
              currently decodes "jobspec" and "R" into dicts
              (default True)
     :original: For 'jobspec', return the original submitted jobspec
-    :base: For 'jobspec', get base value, do not apply updates from eventlog
+    :base: For 'jobspec' or 'R', get base value, do not apply updates from eventlog
     """
     keyslookup = list(keys)
     _setup_lookup_keys(keyslookup, original, base)
@@ -182,7 +211,7 @@ class JobKVSLookupFuture(WaitAllFuture):
 class JobKVSLookup:
     """User friendly class to lookup job KVS data
 
-    Some keys such as "jobspec" may be altered based on update events
+    Some keys such as "jobspec" or "R" may be altered based on update events
     in the eventlog.  Set 'base' to True to skip these updates and
     read exactly what is in the KVS.
 
@@ -193,7 +222,7 @@ class JobKVSLookup:
              currently decodes "jobspec" and "R" into dicts
              (default True)
     :original: For 'jobspec', return the original submitted jobspec
-    :base: For 'jobspec', get base value, do not apply updates from eventlog
+    :base: For 'jobspec' or 'R', get base value, do not apply updates from eventlog
     """
 
     def __init__(
--- a/src/cmd/flux-job.c
+++ b/src/cmd/flux-job.c
@@ -371,7 +371,7 @@ static struct optparse_option info_opts[] =  {
       .usage = "For key \"jobspec\", return the original submitted jobspec",
     },
     { .name = "base", .key = 'b', .has_arg = 0,
-      .usage = "For key \"jobspec\", get base value, "
+      .usage = "For key \"jobspec\" or \"R\", "
                "do not apply updates from eventlog",
     },
     OPTPARSE_TABLE_END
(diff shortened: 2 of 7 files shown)


grondo:
> Just like https://github.com/flux-framework/flux-core/pull/5463, the commits with new tests do not work and have XXX tagged to be filled in later when an actual way to update the R of a job exists.

Would it work to write a test jobtap plugin that simply emits a `resource-update` event for a job? I think there may be a bit of work to do before we get R updates working through normal channels, and I'd hate for this to have to unnecessarily wait on that...

chu11:
> Would it work to write a test jobtap plugin that simply emits a resource-update event for a job? I think there may be a bit of work to do before we get R updates working through normal channels, and I'd hate for this to have to unnecessarily wait on that...

We could definitely do that in the short term.  The main reason I didn't do that initially is because of the python tests.

But looking at the implementation of `flux-jobtap` i guess a half-simple job-manager RPC should take care of that.

chu11:
re-pushed adding a `resource-update-expiration` plugin that outputs a `resource-update` event when a job is run.  It does something stupid, just increasing the expiration by 72 hours, so we can check in tests that it has changed.

I have comments that say "you may want to update this to something smarter down the road", but the tests do the job for the time being.  So I removed WIP.

src/cmd/flux-job.c
```diff
@@ -3463,6 +3463,61 @@ void info_output_jobspec (flux_future_t *f, struct info_ctx *ctx)
...
+            if (eventlog_entry_parse (entry, NULL, &name, &context) < 0)
+                log_err_exit ("Failed to parse eventlog entry");
+
+            if (!streq (name, "resource-update"))
+                continue;
+
+            json_object_foreach (context, path, value) {
+                if (jpath_set (R, path, value) < 0)
+                    log_err_exit ("Failed to update R");
+            }
```
grondo, line 3507: RFC 21 specifies that a [resource-update](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_21.html#resource-update-event) event may only have a key of `expiration`, which updates the resource set expiration. I don't think this code will work here since the expiration is located in `execution.expiration`.

In the future other keys may indicate grow or shrink (or those may be different events) and those would likely not apply with `jpath_set()` either, so probably better to specifically handle `expiration` here and generate an error for any other key in the `resource-update` event context.

t/t2230-job-info-lookup.t
```diff
@@ -105,6 +105,24 @@ test_expect_success 'flux job info jobspec fails on bad id' '
...
 '
 
+# N.B. In future may wish to update expiration via `flux update` tool
+# when feature is available
+test_expect_success 'flux job info applies resource updates' '
+	jobid=$(flux submit --wait-event=start sleep 300) &&
+	echo $jobid > updated_R.id &&
+	flux job info $jobid R | jq -e ".execution.expiration == 0.0" &&
+	kvspath=`flux job id --to=kvs ${jobid}` &&
+	flux kvs eventlog append ${kvspath}.eventlog resource-update "{\"execution.expiration\": 1000.0}" &&
```
grondo, line 115: See previous comment. Only allowed key in `resource-update` event context is `expiration`.

src/bindings/python/flux/job/kvslookup.py
```diff
@@ -87,7 +92,21 @@ def _get_updated_jobspec(job_data):
...
+def _get_updated_R(job_data):
+    if isinstance(job_data["R"], str):
+        R = json.loads(job_data["R"])
+    else:
+        R = job_data["R"]
+    for entry in job_data["eventlog"].splitlines():
+        event = EventLogEvent(entry)
+        if event.name == "resource-update":
+            for key, value in event.context.items():
+                set_treedict(R, key, value)
```
grondo, line 104: As above, need to handle `expiration` correctly here.

t/job-manager/plugins/resource-update-expiration.c
```diff
@@ -0,0 +1,85 @@
...
+                                     "gettimeofday");
+        return -1;
+    }
+
+    newexp = tv.tv_sec + (72 * 60 * 60);
+    if (flux_jobtap_event_post_pack (p,
+                                     id,
+                                     "resource-update",
+                                     "{s:f}",
+                                     "execution.expiration", newexp) < 0) {
```
grondo, line 63: As noted elsewhere, RFC 21 specifies `expiration` as the only allowed `resource-update` context key.

grondo (commented):
Looks good!

Seems like there was a miscommunication about the `resource-update` event -- it is not similar to `jobspec-update` in that only an `expiration` key is currently allowed, which sets a new `execution.expiration`. (Commented inline as I noted spots that need to be updated.)

chu11:
> Seems like there was a miscommunication about the resource-update event -- it is not similar to jobspec-update in that only an expiration key is currently allowed, which sets a new execution.expiration. (Commented inline as I noted spots that need to be updated.)

Doh!  I had just completely misread the RFC!

re-pushed with changes, mostly `if key == "expiration" ...` checks.

```
diff --git a/src/bindings/python/flux/job/kvslookup.py b/src/bindings/python/flux/job/kvslookup.py
index 776b9afa8..276ec0863 100644
--- a/src/bindings/python/flux/job/kvslookup.py
+++ b/src/bindings/python/flux/job/kvslookup.py
@@ -101,7 +101,8 @@ def _get_updated_R(job_data):
         event = EventLogEvent(entry)
         if event.name == "resource-update":
             for key, value in event.context.items():
-                set_treedict(R, key, value)
+                if key == "expiration":
+                    set_treedict(R, "execution.expiration", value)
     return json.dumps(R, ensure_ascii=False)
 
 
diff --git a/src/cmd/flux-job.c b/src/cmd/flux-job.c
index 9629bfe9f..9c98b34de 100644
--- a/src/cmd/flux-job.c
+++ b/src/cmd/flux-job.c
@@ -3502,8 +3502,10 @@ void info_output_R (flux_future_t *f, struct info_ctx *ctx)
                 continue;
 
             json_object_foreach (context, path, value) {
-                if (jpath_set (R, path, value) < 0)
-                    log_err_exit ("Failed to update R");
+                if (streq (path, "expiration")) {
+                    if (jpath_set (R, "execution.expiration", value) < 0)
+                        log_err_exit ("Failed to update R");
+                }
             }
         }
 
diff --git a/t/job-manager/plugins/resource-update-expiration.c b/t/job-manager/plugins/resource-update-expiration.c
index cc3ace891..872d60011 100644
--- a/t/job-manager/plugins/resource-update-expiration.c
+++ b/t/job-manager/plugins/resource-update-expiration.c
@@ -60,7 +60,7 @@ static int run_cb (flux_plugin_t *p,
                                      id,
                                      "resource-update",
                                      "{s:f}",
-                                     "execution.expiration", newexp) < 0) {
+                                     "expiration", newexp) < 0) {
         flux_jobtap_raise_exception (p,
                                      FLUX_JOBTAP_CURRENT_JOB,
                                      "resource-update", 0,
diff --git a/t/t2230-job-info-lookup.t b/t/t2230-job-info-lookup.t
index e87c323ab..dad9d8746 100755
--- a/t/t2230-job-info-lookup.t
+++ b/t/t2230-job-info-lookup.t
@@ -112,8 +112,8 @@ test_expect_success 'flux job info applies resource updates' '
        echo $jobid > updated_R.id &&
        flux job info $jobid R | jq -e ".execution.expiration == 0.0" &&
        kvspath=`flux job id --to=kvs ${jobid}` &&
-       flux kvs eventlog append ${kvspath}.eventlog resource-update "{\"execution.expiration\": 1000.0}" &&
-       flux job info $jobid R | jq -e ".execution.expiration > 0.0" &&
+       flux kvs eventlog append ${kvspath}.eventlog resource-update "{\"expiration\": 1000.0}" &&
+       flux job info $jobid R | jq -e ".execution.expiration == 1000.0" &&
        flux job info $jobid eventlog &&
        flux cancel $jobid
 '
```

grondo:
Given the comments in #5467 I wonder if we want to consider having the `job-info` module apply the updates here, instead of pushing the application down into clients? This would allow a scheduler fetching R to always get the updated copy and not have to apply the updates manually (essentially duplicating code). I'm not sure how much work that would be though...

chu11:
> Given the comments in #5467 I wonder if we want to consider having the job-info module apply the updates here, instead of pushing the application down into clients? This would allow a scheduler fetching R to always get the updated copy and not have to apply the updates manually (essentially duplicating code). I'm not sure how much work that would be though...

I don't think it should be too much work, I think the big questions are:

- if we do this for `R`, we should probably do it for `jobspec` too.  I did architect things in #5467 in preparation for the possibility

- should we do this in `job-info.lookup` and leave `job-info.update-watch` to be just for update watching?
  - and `job-info.lookup` should also support a "base" option of sorts.

grondo:
> should we do this in job-info.lookup and leave job-info.update-watch to be just for update watching?

I'm not sure. `job-info.lookup` does support `flags` right? Would it be good to have `update-watch` functionality as a `LOOKUP_WATCH` flag similar to how the KVS API works? That being said, it isn't too much extra work to convert to a new RPC vs adding a flag, so I'm not sure I have a strong preference here... I wonder if @garlick has any thoughts.

grondo (approved):
LGTM. Rebased and fixed up the test jobtap plugin to use R from job-manager.

garlick (approved):
This looks like it gets the job done and I'm fine with merging. 
We should come back around and replace the one-off jobspec/R update code with calls to the new `job-info.update-lookup` RPC at some point though.

grondo:
Thanks @garlick. I've set MWP on behalf of @chu11.

PR #5467, opened 2023-09-22
job-info: support new update-lookup and update-watch service

chu11:
per discussion in #5451

Problem: In the future, several services will need to know a job's resources and know the updates that would apply to them.
This would currently require users to read R, watch the eventlog, and then apply `resource-update` events as they happen.  It would be nice if a service did this as there will be multiple users.

Solution: Support a new job-info.update-watch streaming service.  It currently supports only the key "R", but can be extended to other keys in the future.

The service will read R and the eventlog for a job and apply all resource-update changes as needed.  This initial "R" will sent back to the caller.  If the job has completed, the RPC streaming service ends.  If not, the eventlog will be watched for future resource-update events.  On each new resource-update event, a new R will be streamed back to the caller.  This continues until the job ends or the caller cancels the stream.

----

Notes:

- It might be wise to add some type of "caching" of sorts as multiple callers may want to watch the same job's R value.  I didn't do that at the moment.  I figured that could be an optimization for the future.

- per discussion in #5451 future options could include a "uniq" option as a "full history" option.

### DIFF
--- a/src/modules/Makefile.am
+++ b/src/modules/Makefile.am
@@ -169,7 +169,9 @@ job_info_la_SOURCES = \
 	job-info/watch.h \
 	job-info/watch.c \
 	job-info/guest_watch.h \
-	job-info/guest_watch.c
+	job-info/guest_watch.c \
+	job-info/util.h \
+	job-info/util.c
 job_info_la_CPPFLAGS = \
 	$(AM_CPPFLAGS) \
 	$(FLUX_SECURITY_CFLAGS)
--- a/src/modules/Makefile.am
+++ b/src/modules/Makefile.am
@@ -170,6 +170,8 @@ job_info_la_SOURCES = \
 	job-info/watch.c \
 	job-info/guest_watch.h \
 	job-info/guest_watch.c \
+	job-info/update.h \
+	job-info/update.c \
 	job-info/util.h \
 	job-info/util.c
 job_info_la_CPPFLAGS = \
@@ -178,6 +180,7 @@ job_info_la_CPPFLAGS = \
 job_info_la_LIBADD = \
 	$(top_builddir)/src/common/libflux-internal.la \
 	$(top_builddir)/src/common/libflux-core.la \
+	$(top_builddir)/src/common/libjob/libjob.la \
 	$(JANSSON_LIBS) \
 	$(HWLOC_LIBS)
 job_info_la_LDFLAGS = $(fluxmod_ldflags) -module
--- a/src/modules/job-info/update.c
+++ b/src/modules/job-info/update.c
@@ -1,0 +1,720 @@
+/************************************************************\
+ * Copyright 2023 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* update.c - handle job-info.update-watch &
+ * job-info.eventlog-update-cancel for job-info */
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+#include <jansson.h>
+#include <flux/core.h>
+
+#include "src/common/libczmqcontainers/czmq_containers.h"
+#include "src/common/libjob/job.h"
+#include "src/common/libjob/idf58.h"
+#include "src/common/libeventlog/eventlog.h"
+#include "src/common/libutil/jpath.h"
+#include "ccan/str/str.h"
+
+#include "job-info.h"
+#include "update.h"
+#include "util.h"
+
+typedef enum {
+    UPDATE_TYPE_LOOKUP,
+    UPDATE_TYPE_WATCH,
+} update_type_t;
+
+struct update_ctx {
+    struct info_ctx *ctx;
+    update_type_t type;
+    struct flux_msglist *msglist;
+    uint32_t userid;
+    flux_jobid_t id;
+    char *key;
+    int flags;
+    const char *update_name;
+    flux_future_t *lookup_f;
+    flux_future_t *eventlog_watch_f;
+    bool eventlog_watch_canceled;
+    json_t *update_object;
+    int initial_update_count;
+    int watch_update_count;
+    char *index_key;            /* for watch */
+};
+
+static void update_ctx_destroy (void *data)
+{
+    if (data) {
+        struct update_ctx *uc = data;
+        int save_errno = errno;
+        flux_msglist_destroy (uc->msglist);
+        free (uc->key);
+        flux_future_destroy (uc->lookup_f);
+        flux_future_destroy (uc->eventlog_watch_f);
+        json_decref (uc->update_object);
+        free (uc->index_key);
+        free (uc);
+        errno = save_errno;
+    }
+}
+
+static char *get_index_key (flux_jobid_t id, const char *key)
+{
+    char *s;
+    if (asprintf (&s, "%ju-%s", id, key) < 0)
+        return NULL;
+    return s;
+}
+
+static struct update_ctx *update_ctx_create (struct info_ctx *ctx,
+                                             update_type_t type,
+                                             const flux_msg_t *msg,
+                                             flux_jobid_t id,
+                                             const char *key,
+                                             int flags)
+{
+    struct update_ctx *uc = calloc (1, sizeof (*uc));
+
+    if (!uc)
+        return NULL;
+
+    uc->ctx = ctx;
+    uc->type = type;
+    uc->id = id;
+    if (!(uc->key = strdup (key))) {
+        errno = ENOMEM;
+        goto error;
+    }
+    if (streq (key, "R"))
+        uc->update_name = "resource-update";
+    else {
+        errno = EINVAL;
+        goto error;
+    }
+    uc->flags = flags;
+
+    /* for lookups, the msglist will never be > 1 in length */
+    if (!(uc->msglist = flux_msglist_create ()))
+        goto error;
+    flux_msglist_append (uc->msglist, msg);
+
+    if (uc->type == UPDATE_TYPE_WATCH) {
+        /* use jobid + key as lookup key, in future we may support other
+         * keys other than R
+         */
+        if (!(uc->index_key = get_index_key (uc->id, uc->key)))
+            goto error;
+    }
+
+    return uc;
+
+error:
+    update_ctx_destroy (uc);
+    return NULL;
+}
+
+static void eventlog_watch_cancel (struct update_ctx *uc)
+{
+    flux_future_t *f;
+    int matchtag;
+
+    /* in some cases, possible eventlog watch hasn't started yet */
+    if (!uc->eventlog_watch_f || uc->eventlog_watch_canceled)
+        return;
+
+    matchtag = (int)flux_rpc_get_matchtag (uc->eventlog_watch_f);
+
+    if (!(f = flux_rpc_pack (uc->ctx->h,
+                             "job-info.eventlog-watch-cancel",
+                             FLUX_NODEID_ANY,
+                             FLUX_RPC_NORESPONSE,
+                             "{s:i}",
+                             "matchtag", matchtag))) {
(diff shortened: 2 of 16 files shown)


chu11:
> It might be wise to add some type of "caching" of sorts as multiple callers may want to watch the same job's R value. I didn't do that at the moment. I figured that could be an optimization for the future.

As I think about this more, is the caching critically important?  It would definitely save some unnecessary lookups.  Lemme see how hard it'll be to add.

chu11:
although a few parts were a little tricky, wasn't horrible to add the "multiple watchers under 1 eventlog-watch" support.

The commits could be squashed in at a later time, but for ease, I left the separate commits in so the review (can maybe) be easier.

src/modules/job-info/update.c
```diff
@@ -0,0 +1,490 @@
...
+{
+    const char *key;
+    json_t *value;
+
+    json_object_foreach (context, key, value) {
+        /* RFC 21 resource-update event only allows update
+         * to:
+         * - execution.expiration
+         */
+        if (!streq (key, "execution.expiration"))
```
grondo, line 150: Similar to the other PR, the actual key is just `expiration`

chu11: interesting, I now see the `resource-update` context is formatted differently in RFC21 than `jobspec-update`.

Did we have a specific reason for that?  It seems possibly unwise long term for both contexts to be formatted differently?  On the other hand I can understand that `R` should have far more limited update possibilities.

garlick:
Do users want to see the whole history starting with the original R, or just current R + updates?

If the latter, would it make sense to have this RPC check the streaming bit, and if not set, return the current R and done?  Then maybe #5464 and other users like sched hello in #5472 wouldn't need to replay the eventlog on their own?

(I haven't looked at this - just basing this comment on your summary in #5472 - so sorry if it's off base)

Possibly the method name would need adjustment then.

chu11:
> Do users want to see the whole history starting with the original R, or just current R + updates?

The assumption in this PR is the latter.  It returns R in its current form (R + updates currently in eventlog), and then streams further updates to R that occur later.

I left whole history as a future option, as it wasn't clear there was a use case at the moment (other than "would like to see it just to see it").

> If the latter, would it make sense to have this RPC check the streaming bit, and if not set, return the current R and done? Then maybe #5464 and other users like sched hello in #5472 wouldn't need to replay the eventlog on their own?

That isn't a terrible idea, makes sense.  I guess the (original) theory was that whoever called this service would also want to know if R ever changed in the future.  But I'm guessing there are a few cases where that's not needed?

chu11:
re-pushed, updating "execution.expiration" to "expiration", as I had misread the RFC

chu11:
Hmmm, got this in one builder (gcc-12,content-s3,distcheck) with one of my new tests.  I have no idea how an EAGAIN could even happen.  Will have to think about this.

```
  expecting success: 
  	flux submit --wait-event=start sleep inf > job8.id
  	${UPDATE_WATCH} $(cat job8.id) R > watch8A.out &
  	watchpidA=$! &&
  	wait_update_watchers 1 &&
  	kvspath=`flux job id --to=kvs $(cat job8.id)` &&
  	flux kvs eventlog append ${kvspath}.eventlog resource-update "{\"expiration\": 100.0}" &&
  	${WAITFILE} --count=2 --timeout=30 --pattern="expiration" watch8A.out
  	${UPDATE_WATCH} $(cat job8.id) R > watch8B.out &
  	watchpidB=$! &&
  	wait_update_watchers 2 &&
  	kvspath=`flux job id --to=kvs $(cat job8.id)` &&
  	flux kvs eventlog append ${kvspath}.eventlog resource-update "{\"expiration\": 200.0}" &&
  	${WAITFILE} --count=3 --timeout=30 --pattern="expiration" watch8A.out &&
  	${WAITFILE} --count=2 --timeout=30 --pattern="expiration" watch8B.out
  	${UPDATE_WATCH} $(cat job8.id) R > watch8C.out &
  	watchpidC=$!
  	wait_update_watchers 3
  	${WAITFILE} --count=1 --timeout=30 --pattern="expiration" watch8C.out &&
  	flux cancel $(cat job8.id) &&
  	wait $watchpidA &&
  	wait $watchpidB &&
  	wait $watchpidC &&
  	test $(cat watch8A.out | wc -l) -eq 3 &&
  	test $(cat watch8B.out | wc -l) -eq 2 &&
  	test $(cat watch8C.out | wc -l) -eq 1 &&
  	head -n1 watch8A.out | jq -e ".execution.expiration == 0.0" &&
  	head -n2 watch8A.out | tail -n1 | jq -e ".execution.expiration == 100.0" &&
  	tail -n1 watch8A.out | jq -e ".execution.expiration == 200.0" &&
  	head -n1 watch8B.out | jq -e ".execution.expiration == 100.0" &&
  	tail -n1 watch8B.out | jq -e ".execution.expiration == 200.0" &&
  	tail -n1 watch8C.out | jq -e ".execution.expiration == 200.0"
  
  {"version": 1, "execution": {"R_lite": [{"rank": "0", "children": {"core": "0"}}], "starttime": 0.0, "expiration": 0.0, "nodelist": ["fv-az505-771"]}}
  {"version": 1, "execution": {"R_lite": [{"rank": "0", "children": {"core": "0"}}], "starttime": 0.0, "expiration": 100.0, "nodelist": ["fv-az505-771"]}}
  {"version": 1, "execution": {"R_lite": [{"rank": "0", "children": {"core": "0"}}], "starttime": 0.0, "expiration": 0.0, "nodelist": ["fv-az505-771"]}}
  {"version": 1, "execution": {"R_lite": [{"rank": "0", "children": {"core": "0"}}], "starttime": 0.0, "expiration": 100.0, "nodelist": ["fv-az505-771"]}}
  {"version": 1, "execution": {"R_lite": [{"rank": "0", "children": {"core": "0"}}], "starttime": 0.0, "expiration": 200.0, "nodelist": ["fv-az505-771"]}}
  {"version": 1, "execution": {"R_lite": [{"rank": "0", "children": {"core": "0"}}], "starttime": 0.0, "expiration": 100.0, "nodelist": ["fv-az505-771"]}}
  {"version": 1, "execution": {"R_lite": [{"rank": "0", "children": {"core": "0"}}], "starttime": 0.0, "expiration": 200.0, "nodelist": ["fv-az505-771"]}}
  {"version": 1, "execution": {"R_lite": [{"rank": "0", "children": {"core": "0"}}], "starttime": 0.0, "expiration": 200.0, "nodelist": ["fv-az505-771"]}}
  update_watch_stream: job-info.update-watch: Resource temporarily unavailable
  update_watch_stream: job-info.update-watch: Resource temporarily unavailable
  not ok 8 - job-info: update watch with multiple watchers works
  ```

File: src/modules/job-info/update.c (hunk 1 above)
Comment (grondo) [reply src/modules/job-info/update.c]: > Did we have a specific reason for that? It seems possibly unwise long term for both contexts to be formatted differently? 

I think it was done this way because jobspec and R are actually quite different - R is not a simple hierarchical document like jobspec. E.g., when grow/shrink or similar are added, it will be more useful to have the added or deleted R and not a wholesale replacement of `execution.R_lite`.

**Comment by grondo** [comment]
> Do users want to see the whole history starting with the original R, or just current R + updates?
>
> If the latter, would it make sense to have this RPC check the streaming bit, and if not set, return the current R and done?
>

Stepping back a bit, there are two use cases here:

 - A user that wants to see the current R + updates. The use case here is a user running `flux job info R` or scheduler fetching R after a reload/restart, etc. As noted in #5472, it would be nice if `job-info.lookup` just returned this value so we don't have to even update callers...
 - A user wants to see current R and updates. The use case here is any entity that is monitoring R for changes, e.g. for expiration updates. This includes the job-exec module (so it can update the timer at which it terminates the job), the job shell (to properly implement the `--signal=SIG@TIME` option), and the scheduler of a subinstance (to update the internal expiration of its resource set).

**Comment by chu11** [comment]
per offline discussion, re-worked this PR.

- like before, there is a `job-info.update-watch` target that returns the R + updates result, then streams R updates as they occur.  Multiple watches on the same jobid + R share the same internal eventlog-watch.

- there is a `job-info.update-lookup` target, which looks up R and eventlog and returns the current R result and that's it.  If there is already a `update-watch` for that jobid, it can return the cached result.

because of all these changes, I had to squash my previous changes, so unfortunately there's just one giant commit adding `update-lookup` and `update-watch`, hopefully it's not too hard to review :P

**Comment by chu11** [comment]
hmmm several builders are consistently failing with

```
Run docker run --rm --privileged aptman/qus -s -- -p --credential aarch64
Unable to find image 'aptman/qus:latest' locally
docker: Error response from daemon: Head "https://registry-1.docker.io/v2/aptman/qus/manifests/latest": EOF.
See 'docker run --help'.
Error: Process completed with exit code 125.
```

under the `setup qemu-user-static` chunk.  Perhaps something happened after #5469 merged?

**Comment by grondo** [comment]
My guess is the docker manifest for that repo is corrupted. Same issue in flux-sched.

**Review by grondo — APPROVED** [review]
I've been basing work to handle R updates in the job-exec service and job shell and this seems to be working. My inclination is to merge this PR now and make any needed changes and improvements if necessary as we need them.

File: src/modules/Makefile.am [hunk 2]
```diff
@@ -170,6 +170,8 @@ job_info_la_SOURCES = \
 	job-info/watch.c \
```
garlick, line 170: commit mesage for `job-info: support new update-lookup/watch service `: capitalize "it" in the 4th paragraph of the body.

src/modules/job-info/update.c
```diff
@@ -0,0 +1,720 @@
...
+    struct update_ctx *uc = calloc (1, sizeof (*uc));
+
+    if (!uc)
+        return NULL;
+
+    uc->ctx = ctx;
+    uc->type = type;
+    uc->id = id;
+    if (!(uc->key = strdup (key))) {
+        errno = ENOMEM;
```
garlick, line 94: strdup already sets ENOMEM on failure

src/modules/job-info/update.c
```diff
@@ -0,0 +1,720 @@
...
+        json_t *context = NULL;
+        if (eventlog_parse_entry_chunk (uc->ctx->h,
+                                        tok,
+                                        toklen,
+                                        &entry,
+                                        &name,
+                                        &context) < 0) {
+            errmsg = "error parsing eventlog";
+            goto error;
+        }
```
garlick, line 333: Yeah, as noted earlier, it looks like this could be replaced with (with appropriate error handling)
```c
eventlog = eventlog_decode (eventlog_str);
json_array_foreach (eventlog, index, entry) {
    eventlog_entry_parse (entry, NULL, &name, &context);
    ...
}
```

src/modules/job-info/update.c
```diff
@@ -0,0 +1,720 @@
...
+        if (cancel)
+            match = flux_cancel_match (msg, cmpmsg);
+        else
+            match = flux_disconnect_match (msg, cmpmsg);
+
+        if (match) {
+            if (flux_respond_error (uc->ctx->h, cmpmsg, ENODATA, NULL) < 0)
+                flux_log_error (uc->ctx->h,
+                                "%s: flux_respond_error",
+                                __FUNCTION__);
```
garlick, line 657: Response should only be sent on cancel, not disconnect.

src/modules/job-info/update.c
```diff
@@ -0,0 +1,720 @@
...
+        eventlog_watch_cancel (uc);
+}
+
+void update_watchers_cancel (struct info_ctx *ctx, const flux_msg_t *msg, bool cancel)
+{
+    struct update_ctx *uc;
+
+    uc = zlist_first (ctx->update_watchers);
+    while (uc) {
+        update_watch_cancel (uc, msg, cancel);
```
garlick, line 675: Maybe it would be better to replace `update_watch_cancel()` with
```c
if (cancel)
    flux_msglist_cancel (ctx->h, uc->msglist, msg);
else
    flux_msglist_disconnect (ctx->h, uc->msglist, msg);
```

garlick (approved):
I found a few things but actually, if this is working, maybe it's fine to merge as is and open an issue to return and clean up the two bigger things I called out
- use eventlog class to parse eventlogs
- use `flux_msglist_disconnect()` and `flux_msglist_cancel()` instead of duplicating that functionality.

Since it may take a bit of time to audit that code and make sure those changes don't break anything, merge now+fix later may be more expedient.

grondo:
I've just set MWP on behalf of @chu11

PR #5468, opened 2023-09-22
drop libzcmq dependency

garlick:
Problem: libczmq is no longer used by flux-core but it's still required.

Drop it from` configure`, `make deb`, and `Dockerfile`s.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -327,7 +327,7 @@ AX_PROG_LUA([5.1],[5.5])
 AX_LUA_HEADERS
 AX_LUA_LIBS
 # N.B oldest in CI are focal=libzmq-4.3.2-2, centos7=zeromq-4.1.4
-PKG_CHECK_MODULES([ZMQ], [libczmq >= 3.0.0 libzmq >= 4.0.4])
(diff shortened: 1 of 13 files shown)


grondo (approved):
Great! Looks good to me!

One missed spot, though I think we've been meaning to just remove this directory...

```$ git grep libczmq | grep -v containers
etc/docker/ubuntu/install_deps.sh:  libczmq-dev
```

garlick:
I went ahead and just removed it from the script rather than the directory for now.

garlick:
Hmm the container build seems to be failing here now
```
 #10 171.7 make[4]: *** [Makefile:780: constraint/parsetab.py] Error 1
```
So about that deletion plan...what exactly can be deleted?  The dependencies in this list don't look right but I'm not sure how removing czmq "broke" anything unless that just triggered a new build?

grondo:
That container build always fails (which is why we should just remove it), but usually the build only runs on pushes to master. I think the change to the file must have triggered a build for this PR. Maybe just drop that change and we'll deal with the build-containers action separately?

garlick:
Dropped that commit and I'll set MWP.  Thanks!

PR #5469, opened 2023-09-23
switch qemu-user-static setup to fix setuid

trws:
This pulls over the fix from the matching flux-sched nosuid issue in case we run into it here at some point.

Problem: flux-sched side ran into issues with setuid binaries in some cases under aarch64 emulation building containers.  It seems to depend on which workers pick up the job, so seems best to bring the solution over.

Solution: Run the docker container with the setup script directly with the appropriate option to handle setuid credentials.


grondo (approved):
Seems reasonable to me!

PR #5470, opened 2023-09-24
doc:  add content to landing page and group man pages

garlick:
Problem: the readthedocs landing page for flux-core is just a boiler plate with a flat list of man pages and python docs and no other context.  See: https://flux-framework.readthedocs.io/projects/flux-core/en/latest/index.html

Add some content to the landing page, copying structure from the openmpi docs which I thought were well done.
https://docs.open-mpi.org/en/v5.0.x/index.html

There's still really nothing here except the landing page, the man pages, and the python API pages, but now I think this looks better and we could consider developing an outline for some flux-core specific docs to add.

Also, I found the readthedocs man pages to be hard to navigate especially in sections 1 and 3 so I added some grouping.  See what you think of the preview.


chu11 (approved):
I like, LGTM

garlick:
Whenever I write prose, the next morning it looks dumb!  I did a little editing on the intro section.  Hope that's OK.

doc/Makefile.am
```diff
@@ -395,6 +395,10 @@ EXTRA_DIST = \
 	index.rst \
 	index_man.rst \
 	$(RST_FILES) \
+	man1/index.rst \
+	man1/index_general.rst \
+	man1/index_job.rst \
+	man1/index_kvs.rst \
```
grondo, line 401: Since this moves the man pages around, will the link(s) in flux-docs need to be updated? (I actually don't know)

garlick: Oh good question.  Well the man pages didn't move in the repo - only the toctrees changed.   For example:

https://flux-framework--5470.org.readthedocs.build/projects/flux-core/en/5470/man1/flux-job.html

So I _think_ we're OK!

grondo: Oh yeah, also as long as this link still works:

https://flux-framework.readthedocs.io/projects/flux-core/en/latest/index.html#man-pages

Then we're fine. Thanks

garlick: Hmm, I'm not sure now that I look at it.  This goes to the landing page:

https://flux-framework--5470.org.readthedocs.build/projects/flux-core/en/5470/index.html#man-pages

But the link you mentioned (presumably from [here](https://flux-framework.readthedocs.io/en/latest/#manual-pages)) is an intersphinx reference 
```rst
:ref:`core:man-pages`
```
and we still define that label in `index_man.rst`:
```rst
.. _man-pages:

Manual Pages
============

.. toctree::
   :maxdepth: 1

   man1/index
   man3/index
   man5/index
   man7/index
```
I would kind of expect some magic to happen when that URI is resolved but it doesn't seem to be happening in the preview.

garlick: The link isn't _broken_ per se since the anchor is ignored if it doesn't exist, and the link does land on a page where the man pages are conveniently accessible in the left column table of contents.

We can update the link in flux-docs but of course we don't know what else has cached the link in question.

What do you think @grondo - is this reasonable to merge as is?

garlick: OK, I'll set MWP and if we broken anything we can go from there.  This is not exactly an area I know much about!

Issue #5471, opened 2023-09-25
Keep a copy of R in job-manager for use in jobtap plugin callbacks


grondo:
As noted in #3851, there are already a couple use cases where having the allocated _R_ available to jobtap plugins would be convenient. In addition to those documented use cases, having _R_ available from the job-manager will also be necessary in order to update the duration of a _running_ job (#4175), since the current expiration will need to be available to create the new expiration necessary for the `resource-update` event.

This issue proposes that the job-manager keep a copy of _R_ for jobs as they enter the RUN state. This may mean reworking the scheduler protocol to reply to an alloc request with _R_ instead of putting _R_ in the KVS and replying only after the commit. The job-manager could become the entity that puts _R_ in the KVS or this can be done in parallel.

garlick:
Have you already begun working on this?  I'd be happy to prototype it and propose any RFC updates needed to [RFC 28](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_28.html).

grondo:
No, I haven't. Thanks @garlick!

grondo:
This was fixed by #5472

Issue #5411, opened 2023-08-25
job-info: get internal representation of jobspec


chu11:
`flux job info <JOBID> jobspec` gets the jobspec from the KVS, but this may not be the jobspec used internally by the `job-manager` or is represented inside `job-list`.

The jobspec can be updated via jobtap plugins (perhaps adding defaults) or users can update via the future `flux update` command #5409.

It would be useful if `flux job info <JOBID> jobspec` got the presently used / represented jobspec. 

Perhaps a `--kvs` option could be used to get what is in the KVS, similar to the `--original` option (and add to the Python equivalent convenience functions).

Also, an API mechanism to get this would be useful, as the `flux-update` command could use such a convenience function.

Implementation note: `job-list` already builds an internal representation of `jobspec` given `jobspec-update` events in the eventlog.  It could be used to retrieve the internal representation of a jobspec.  The only minor downside is that it can be racy, as `jobspec-update` events are sent to it via the journal.  We probably don't want to create a service to retrieve the internal jobspec from `job-manager`.  

Edit: I guess `job-info` could build the proper representation of `jobspec` since it has to read `eventlog` for a security check anyways.

grondo:
As a side note, a similar treatment for _R_ might be nice once we support `resource-update` events, so that `flux job info JOBID R` returns any updated expiration at the time of the request.

chu11:
So pondering this a bit more and I see multiple approaches:

- get the redacted jobspec via job-list
  - Pro: it's already in there
  - Con: we don't have guest access check in there
  - Con: minor eventual consistency raciness

- get the redacted jobspec via job-info
  - Pro: guest access checks already there
  - Pro: we already have to read the eventlog for guest access, so no extra KVS reads
  - Con: gotta build the jobspec up, putting cycles on job-info in broker

- get the redacted jobspec via job-manager
  - Pro: it's already there
  - Con: we gotta support some user facing service thingie to get it from here

- create library util functions to do this for job-info / tools so we atleast have it in one place
  - Pro: no extra load on broker
  - Con: more work

No obvious winner, pros and cons each direction.  Am leaning to just create a helper function that `flux job info` and `flux update` and anything in the future can all use, since it doesn't appear we need widespread use at the moment.

PR #5472, opened 2023-09-26
cache R in the job manager

garlick:
Problem: jobtap plugins might need convenient access to R per #5471 

Per discussion, this prototypes keeping a copy of R in the job manager so that it could potentially be offered to jobtap plugins.  Some notes/caveats:
- R is not deleted after the job becomes inactive, which could be a potential optimization to conserve memory on JGF systems
- jobtap interfaces are unchanged (I thought @grondo might already have a good idea of how he wants to do that)
- if this is acceptable, RFC 27 will need updates to `sched.alloc` and  `job-manager.sched-hello` responses.
- this is set up so that job manager restart could replay `resource-update` events on top of the original _R_ loaded from the KVS
- R could be passed to `job-exec.start` to save a lookup (would require an update to RFC 32).  (Edit: oops I included a couple of commits that started this - maybe I'll just finish it and tack it on)

### DIFF
--- a/src/common/libschedutil/alloc.c
+++ b/src/common/libschedutil/alloc.c
@@ -18,8 +18,10 @@
 #include "init.h"
 #include "alloc.h"
 
-static int schedutil_alloc_respond (flux_t *h, const flux_msg_t *msg,
-                                    int type, const char *note,
+static int schedutil_alloc_respond (flux_t *h,
+                                    const flux_msg_t *msg,
+                                    int type,
+                                    const char *note,
                                     json_t *annotations)
 {
     flux_jobid_t id;
@@ -27,26 +29,36 @@ static int schedutil_alloc_respond (flux_t *h, const flux_msg_t *msg,
 
     if (flux_request_unpack (msg, NULL, "{s:I}", "id", &id) < 0)
         return -1;
-    if (annotations)
-        rc = flux_respond_pack (h, msg, "{s:I s:i s:O}",
-                                        "id", id,
-                                        "type", type,
-                                        "annotations", annotations);
-    else if (note)
-        rc = flux_respond_pack (h, msg, "{s:I s:i s:s}",
-                                        "id", id,
-                                        "type", type,
-                                        "note", note);
-    else
-        rc = flux_respond_pack (h, msg, "{s:I s:i}",
-                                        "id", id,
-                                        "type", type);
+    if (annotations) {
+        rc = flux_respond_pack (h,
+                                msg,
+                                "{s:I s:i s:O}",
+                                "id", id,
+                                "type", type,
+                                "annotations", annotations);
+    }
+    else if (note) {
+        rc = flux_respond_pack (h,
+                                msg,
+                                "{s:I s:i s:s}",
+                                "id", id,
+                                "type", type,
+                                "note", note);
+    }
+    else {
+        rc = flux_respond_pack (h,
+                                msg,
+                                "{s:I s:i}",
+                                "id", id,
+                                "type", type);
+    }
     return rc;
 }
 
 int schedutil_alloc_respond_annotate_pack (schedutil_t *util,
                                            const flux_msg_t *msg,
-                                           const char *fmt, ...)
+                                           const char *fmt,
+                                           ...)
 {
     va_list ap;
     json_t *o = NULL;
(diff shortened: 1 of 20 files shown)


garlick:
Oops one place where `job->R` won't be populated is with `alloc-bypass`.

I made some sloppy errors caught by CI and will force push with fixes in a sec as soon as tests pass locally.

grondo:
Great!

> R is not deleted after the job becomes inactive, which could be a potential optimization to conserve memory on JGF systems

A quick thought here - should the `scheduling` key of _R_ be redacted in a similar vein to the `environment` attribute of jobspec? I don't see a use to keeping scheduler opaque data in memory of the job manager...

> jobtap interfaces are unchanged (I thought @grondo might already have a good idea of how he wants to do that)

All I was thinking was to add `"R"` to the callback input "args" when it is available (i.e. callbacks after the `alloc` event)

grondo:
> Oops one place where job->R won't be populated is with alloc-bypass.

I didn't check the code here yet, does the scheduler still commit _R_ to the KVS and send it back in the alloc response in parallel, or does the job-manager now commit _R_?  The alloc-bypass plugin has to commit the provided _R_ to the KVS itself, but if the job-manager already does this on behalf of the scheduler, perhaps we just need a way to allow a plugin to _assign_ _R_ and then we can drop this step and the job manager will have the copy of _R_ in this case. (Oh, wait, won't this be necessary anyway if _R_ is sent to the job-exec module with the start request?)

garlick:
> A quick thought here - should the scheduling key of R be redacted in a similar vein to the environment attribute of jobspec? I don't see a use to keeping scheduler opaque data in memory of the job manager...

That was true before, but after this PR, the `job-manager.sched-hello` streaming responses include a copy of R from memory.  My thought was if R has been modified, then we had better give the scheduler the updated copy.  For example, if the expiration has changed, the scheduler needs to re-allocate the resources for the updated time.  Plus it saves a KVS lookup.

I'm not sure if fluxion relies on this.  Maybe (_R_ is passed to `queue->reconstruct()` )
https://github.com/flux-framework/flux-sched/blob/master/qmanager/modules/qmanager_callbacks.cpp#L159

garlick:
> does the scheduler still commit R to the KVS and send it back in the alloc response in parallel

Yes.  That seemed the safest since I think we've advertised the invariant that once the `alloc` event is posted to the eventlog, `R` can be read from the KVS.

> The alloc-bypass plugin has to commit the provided R to the KVS itself, but if the job-manager already does this on behalf of the scheduler, perhaps we just need a way to allow a plugin to assign R and then we can drop this step and the job manager will have the copy of R in this case. (Oh, wait, won't this be necessary anyway if R is sent to the job-exec module with the start request?)

It does not already do it so `alloc-bypass` would need a way to assign R, and would need to retain the KVS commit also.

garlick:
> it saves a KVS lookup at restart only, with a tradeoff of keeping JGF for every job 

True!

> If the scheduler continued to lookup R in the KVS, but instead used the new job-info lookup which incorporates resource-update events, then this would just work and R could be redacted.

Good point, let's do it that way!

Regarding the interface to R for jobtap plugins, I have a commit ready to push that adds an "R" key to the jobtap input args.  I was going to try also adding "R" to the output args just for `alloc-bypass`, then if it's found there and `job->R` is unset, letting it be set.  Hopefully that's not too big of a foot-gun, but I figured I'd try the simple thing first.

garlick:
Re-pushed with the following changes
- added support for R in jobtap IN and OUT args
- updated `alloc-bypass` to set R in OUT args
- Include R in the `job-exec.start` request
- Dropped the schedutil `hello` changes (R is fetched from the KVS, which we can easily change to `job-info` when it provides the modified R.
- rebased on current master

garlick:
>  instead used the new job-info lookup which incorporates resource-update events

That's not available yet, correct?

grondo:
> That's not available yet, correct?

PR posted in #5467 - unfortunately I just realized that job-info lookup RPC still returns the original R because the application of the resource-update events is done on the client side in `flux-job`. I wonder now if that was the right solution (though it does move processing out of the job-info module itself, it means users of the API have to apply the updates themselves? Maybe @chu11 can comment here.

chu11:
> PR posted in #5467 - unfortunately I just realized that job-info lookup RPC still returns the original R because the application of the resource-update events is done on the client side in flux-job. I wonder now if that was the right solution (though it does move processing out of the job-info module itself, it means users of the API have to apply the updates themselves? Maybe @chu11 can comment here.

apologies, there's actually two PRs so it may be confusing

#5467 - an update-watch service that will get the currently up to date "R" (w/ resource-update events applied) and then stream any new `R` changes that may occur in the future.

#5464 - `flux job info <jobid> R` gets the updated R, but does this on the client side.  This is a mirror of `flux job info <jobid> jobspec`, to keep the burden off of the `job-info` module.  This PR could be updated to use the service in #5467, but I elected to keep it out of the broker for the time being because these are likely "one off" lookups, vs the streaming updates are needed for various broker services.  (and admittedly this PR was in development before issue #5451 was written )

> Edit: In a way it would be nice if job-info.lookup just magically returned the updated R and jobspec.. however I did not look into this as deeply as @chu11 so not sure if there are roadblocks to doing it that way.

There is no roadblock that I can see.  The entirety of why it was done client side in `flux job info` (and python equivalent code) is to keep the burden out of the `job-info` module.  Some of my pro-con brainstorming here: https://github.com/flux-framework/flux-core/issues/5411#issuecomment-1703530827.

grondo:
This PR would be useful in development of duration update for running jobs. The job-manager would not have to fetch R from the job-info service to apply a duration update. A cursory examination of this PR indicates it is in pretty good shape. @garlick, what remains to do here to remove the WIP? (Did we ever decide if the cached R should be freed for inactive  jobs? Given that we have users wanting run 1M jobs, perhaps that would be a good idea...)

garlick:
This has been sitting for a while.  A brief review didn't turn up any big gaps, that I caught anyway.  Just force pushed with the following changes:
- drop the scheduling key from R when the job transitions to INACTIVE state to save memory
- drop the scheduling key from R when including it in the `exec.start` request
- rebase on current master

garlick:
Per offline discussion with @grondo, repushed with the R scheduling key redacted upon receipt.  It's still in the KVS of course, and schedutil still looks it up there when the scheduler is reloaded.

grondo (approved):
LGTM!

Issue #5465, opened 2023-09-22
default begin-time dependency format in `flux jobs`


ryanday36:
from the department of minor fixes, currently, `flux jobs` displays the begin-time dependency timestamp in seconds from the epoch. Could this be changed to display a human readable date time instead?

```
[day36@rzvernal10:~]$ flux jobs -A
       JOBID QUEUE    USER     NAME       ST NTASKS NNODES     TIME INFO
 fUf7ibZRKFD pdebug   jdkotul  gemma       D    256     32       3h depends:begin-time=1695430800.000
...
[day36@rzvernal10:~]$ date -d @1695430800
Fri Sep 22 18:00:00 PDT 2023
[day36@rzvernal10:~]$
```

grondo:
Valid request! However, this is going to be slightly tricky because of the way dependencies work and how outstanding dependencies are transmitted back to `flux jobs` from the `job-list` service.

Dependencies are represented as a set of strings, added by a `dependency-add` event, then removed by a `dependency-remove` event. Internally, the outstanding dependencies (holding a job in the DEPEND state) are thus represented as a set of strings. For the `begin-time` dependency, the string is just `begin-time=T`,  where `T` is floating-point seconds since the epoch, i.e. the UNIX time. When `flux jobs` gets the list of current dependencies, it just gets the current set of dependency strings and prints them as is. This is partially done because these dependency strings or descriptions are considered somewhat opaque and extensible.

Given these limitations, there are a couple approaches we could take:

 - Have the `begin-time` dependency plugin represent the time as something more human readable. This would be simplest but has the drawback that any such representation would have to pick a timezone (likely the timezone of the broker process) and use that for the representation. This may not match the user's actual timezone, so this could be confusing (but could be mitigated by appending the timezone to the description)
 - Add some special handling for `begin-time` dependencies in `flux jobs` or the underlying `JobInfo` class. The code would need to walk the set of dependencies looking for `begin-time=`. Since it is known that the remainder of the string is a timestamp, it could parse this and replace with a datetime rendered in the current timezone. This method could be extended to handle other known dependencies in the future if needed.

grondo:
Well patching something into `JobInfo` was pretty easy as a prototype.

For instance, we could show the simple begin time if the job is scheduled to run today, the date otherwise:
```
flux jobs
       JOBID USER     NAME       ST NTASKS NNODES     TIME INFO
[...]

PR #5473, opened 2023-09-26
improve begin-time representation in `flux-jobs(1)` output

grondo:
This proposes a fix for #5465 (the easy way). This adds special case handling of the `begin-time` dependencies in the `JobInfo` class. Other special cases could be added in the future, perhaps with a way to register plugins so that out-of-tree dependency modules could register their own post-processing of raw dependencies (though honestly I doubt this would ever be necessary)

Edit: Oh, I guess it would help to describe the method used here: `begin-time=<timestamp>` is replaced with `begin@<datetime>` where `datetime` is 
 - The time if `datetime` is today
 - The datetime as MMMDD-HH:MM (e.g. `Sep26-17:00`) if `datetime` is tomorrow or later
 - Seconds are only displayed if they are nonzero.

### DIFF
--- a/src/bindings/python/flux/job/info.py
+++ b/src/bindings/python/flux/job/info.py
@@ -15,6 +15,7 @@ import string
 import sys
 import time
 from collections import namedtuple
+from datetime import datetime
 from itertools import chain
 
 import flux.constants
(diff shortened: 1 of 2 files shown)


grondo:
CC @ryanday36 to get his approval.

ryanday36:
This looks good to me. Thanks!

chu11 (approved):
LGTM!

grondo:
Thanks! Setting MWP.

PR #5474, opened 2023-09-26
etc: remove ubuntu build-container action

grondo:
Problem: The ubuntu build-containers action fails every time in ci.

Remove the ubuntu container build. This can be resurrected if there is a future use case.

The spack container remains, even though I checked and there has not been even one download of the result. However, it is at least working. (but I'm fine removing both if that would be cleaner. We could always bring this back later if we have a use case.)


grondo:
Oh, I didn't realize the CodeQL actions used `sudo ./etc/docker/ubuntu/install_deps.sh`. That should probably point to `scripts/install-deps-deb.sh`. Will fix.

chu11 (commented):
I don't understand the workflow stuff that well.  How was CI passing in general even though this wasn't working?  Did we use to run tests off of these containers?  So its "silently" failing now?

grondo:
AFAICT, these containers are not used. The builds do not occur by default on PRs but only a push to master, which regenerates the spack and ubuntu containers. The ubuntu container build broke at some point, so it failed on every push to master. Since obviously nothing uses it, it was easier to just remove it.

The spack build actually works. I'm not sure what the purpose of it is, and the result has never been fetched, so maybe we could remove that too?

chu11 (approved):
Ok, that makes sense to me.  I'll leave the removal of the spack to another discussion since if that's working maybe someone is using it?

grondo:
>  I'll leave the removal of the spack to another discussion since if that's working maybe someone is using it?

Nope. It has 0 downloads. However, someone put work into making it, and it is working, so may as well leave it for now.

PR #5478, opened 2023-10-01
broker: move policy config check out to the modules that rely on it

garlick:
Problem: code was added to the broker, originally by a7760fc52cef445e886d73110698335289680f8d, to check RFC 33 policy/queue configuration because it was thought awkward to check it elsewhere.  Re-evaluating it now, it seems like a misstep.  It's probably better if the broker remains unaware of such details in order to maintain a separation of concerns.

Anyway it's easy to fix.  Move checks into job-ingest and job-manager (with shared code moved to a library).

Also drop an embarrassing chunk of dead code that was accidentally checked into the broker a while back.


grondo (approved):
Seems like an appropriate change to me. My only thought is if `validate_policy()` really belongs in libjob, though I'm not sure where else it could conveniently be placed (I would say libutil, but I guess that doesn't seem appropriate either, but would at least keep the code from being incorporated into libflux-core.so)

src/common/libjob/policy.c
```diff
@@ -0,0 +1,360 @@
+/************************************************************\
+ * Copyright 2020 Lawrence Livermore National Security, LLC
```
grondo, line 2: Also, I'm assuming you kept 2020 copyright date here on purpose due to previous copyright date of code (Note: `git blame` shows the actual date of inclusion of the moved code is 2022, but I doubt this matters much!)

garlick: Since you took the trouble to look up the actual date the code was created, I'll change it to 2022 - thanks!

garlick:
> My only thought is if `validate_policy()` really belongs in libjob, though I'm not sure where else it could conveniently be placed (I would say libutil, but I guess that doesn't seem appropriate either, but would at least keep the code from being incorporated into libflux-core.so)

Well that's a good point.  Hmm, there's `libfluxutil` which was added by 6e26e00af0224bc7b8e4cf3c58e87e6de8a35b33 in which I said
> We'll create this local library as a catch-all for reusable bits that depend on libflux-core.so.

garlick:
Fixed those things and repushed.  I'll set MWP - thanks!

Issue #5484, opened 2023-10-03
broker stuck at exit in `zmq_ctx_term()`


grondo:
@jameshcorbett reported that a flux-coral2 CI test is failing with most recent flux-core+flux-sched with the following log:

```
  ok 1 - job-manager: load dws-jobtap plugin
  PASS: t1000-dws-dependencies.t 1 - job-manager: load dws-jobtap plugin
  ok 2 - job-manager: dws jobtap plugin works when coral2_dws is absent
  PASS: t1000-dws-dependencies.t 2 - job-manager: dws jobtap plugin works when coral2_dws is absent
  ok 3 - job-manager: dws jobtap plugin works when RPCs succeed
  PASS: t1000-dws-dependencies.t 3 - job-manager: dws jobtap plugin works when RPCs succeed
  ok 4 - job-manager: dws jobtap plugin works when creation RPC fails
  PASS: t1000-dws-dependencies.t 4 - job-manager: dws jobtap plugin works when creation RPC fails
  ok 5 - job-manager: dws jobtap plugin works when setup RPC fails
  PASS: t1000-dws-dependencies.t 5 - job-manager: dws jobtap plugin works when setup RPC fails
  ok 6 - job-manager: dws jobtap plugin works when post_run RPC fails
  PASS: t1000-dws-dependencies.t 6 - job-manager: dws jobtap plugin works when post_run RPC fails
  ok 7 - job-manager: dws jobtap plugin works when job hits exception during prolog
  PASS: t1000-dws-dependencies.t 7 - job-manager: dws jobtap plugin works when job hits exception during prolog
  # passed all 7 test(s)
  flux-start: 0 (pid 8494) Killed
  1..7
```
Note all tests pass, but the broker is force-terminated with SIGKILL.

We ran the individual test in a docker container and attached with `gdb`. This is where the broker is stuck:

```gdb
(gdb) where
#0  0x00007f9ef95390e1 in poll () from /lib64/libc.so.6
#1  0x00007f9efa92861e in zmq::signaler_t::wait(int) const ()
   from /lib64/libzmq.so.5
#2  0x00007f9efa904cc4 in zmq::mailbox_t::recv(zmq::command_t*, int) ()
   from /lib64/libzmq.so.5
#3  0x00007f9efa8f09c4 in zmq::ctx_t::terminate() () from /lib64/libzmq.so.5
#4  0x00007f9efa94982e in zmq_ctx_term () from /lib64/libzmq.so.5
[...]

PR #5486, opened 2023-10-05
make the loop:// connector built-in rather than a DSO

garlick:
This adds the capability for connectors to be built-in instead of requiring them all to be built as DSOs, and converts `loop://` to built-in, which allowed the duplicate implementation in libtestutil to be removed.

I started this while looking into replacing `shmem://` with a non-0MQ equivalent as a way to avoid the `zmq_ctx_term(3)` hangs when futures are leaked (#5484) .  That would need to be built-in, at least for the simple design I had in mind where named channels would be stored in a libflux-core.so global protected by mutexes, but even if that turns out to be a bad idea, this seems like a nice cleanup.


src/common/libflux/connector_loop.c
```diff
@@ -87,6 +89,56 @@ static flux_msg_t *op_recv (void *impl, int flags)
     return msg;
 }
 
+static int op_getopt (void *impl, const char *option, void *val, size_t size)
+{
+    loop_ctx_t *c = impl;
+
+    if (option && streq (option, FLUX_OPT_TESTING_USERID)) {
```
grondo, line 96: Just curious if there's a case for this function to be called with a NULL `option`, and if not should it just have the following at the top?
```c
if (option == NULL) {
    errno = EINVAL;
    return -1;
}
```

Oh I see this just matches the usage in the local connector, but I'm still unsure of the use case for a NULL option.

grondo (approved):
Wow, nice cleanup! 

From the coverage reporting at least, it appears that setting testing userid/rolemask for the `loop://` connector from the environment variables `FLUX_HANDLE_USERID` and `FLUX_HANDLE_ROLEMASK` is not covered (likely wasn't before either though). Could that be dropped? If not maybe it should have a couple sanity tests? Meh, just a suggestion though.

grondo:
Looks like codeql is failing (hopefully temporarily) due to an error fetching some ubuntu packages.

src/common/libflux/connector_loop.c
garlick: I just copied over what was there in the `loopback_create()` implementation, but I'll have a look and see if there's a bit more cleanup we can do.

grondo: Seems fine to me for now if you want to defer cleanup for later. Just something that caught my eye is all..

garlick:
> From the coverage reporting at least, it appears that setting testing userid/rolemask for the loop:// connector from the environment variables FLUX_HANDLE_USERID and FLUX_HANDLE_ROLEMASK is not covered (likely wasn't before either though). Could that be dropped? If not maybe it should have a couple sanity tests? 

The environment variables are interpreted inside `flux_open()` and translated into `flux_opt_set ()` calls of FLUX_OPT_TESTING_USERID and FLUX_OPT_TESTING_ROLEMASK.  I know there are tests that use the options direcly for the loop connector (see libflux/test/rpc_security.c for example) and sharness tests that use the environment variables with the local connector.  So I think all the code should be covered?  I'll have a closer look at the coverage report.

grondo:
Ah, ok. I pointed it out just in case, if the code is already used and tested then it could be coverage hiccup (or more likely human error on my part) so safe to ignore it!

garlick:
Thanks for the review!  I rebased on current master and forced a push and will set MWP.

PR #5489, opened 2023-10-06
ci: add flux-pmix, flux-pam builds to CI

grondo:
This PR adds the flux-pmix and flux-pam builds to CI to ensure PRs do not break the build/test of these projects.
flux-pam is probably less useful, but it is so quick to build I threw it in there.

This also fixes the `after_n_builds` support for codecov which might hopefully reduce the extra reporting we get for coverage.


chu11 (approved):
seems reasonable to me!

grondo:
I also fixed the readthedocs build by pinning `jinja2<=3.10` (not sure why that suddenly started failing).

Additionally, I noticed that `docker-run-check.sh` was still defaulting flux-security to v0.9.0, so that has now been updated to use v0.10.0.

garlick (approved):
Nice thanks!

grondo:
Thanks! Setting MWP...

PR #5495, opened 2023-10-11
add interthread connector that does not require zeromq

garlick:
This adds a new flux_t connector that provides an alternative to "shmem//" without requiring zeromq.

In this PR, the libtestutil test server is converted to use the new connector.  A follow-on PR will convert the broker module infrastructure to use it.

Since this connector is internally implemented as queues of flux messages, it does not require any encoding/decoding, and in fact could provide the basis for avoiding some message copying in the future.

This connector has to be built-in rather than a dso so that any pair of threads that wish to communicate can access a mutex-protected global data structure.   I thought I could work that out from a dso connector, but couldn't seem to.  This is simpler anyway and lets the libflux tests use it (similar to the rationale in #5486 for making "loop" built-in).

### DIFF
--- a/src/common/libflux/Makefile.am
+++ b/src/common/libflux/Makefile.am
@@ -96,6 +96,7 @@ libflux_la_SOURCES = \
 	attr.c \
 	handle.c \
 	connector_loop.c \
+	connector_interthread.c \
 	reactor.c \
 	reactor_private.h \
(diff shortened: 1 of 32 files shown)


grondo (approved):
LGTM!

garlick:
Thanks!  Setting MWP.

PR #5496, opened 2023-10-11
broker: use interthread connector between broker modules

garlick:
This PR switches broker modules from the shmem connector to the new interthread connector, then finally drops the shmem connector which has no more users.


grondo (approved):
Very nice cleanup! LGTM

garlick:
Thanks!

PR #5498, opened 2023-10-12
testsuite: trivial test fixes

grondo:
This PR contains a couple trivial fixes for the tests. There wasn't really anywhere else to put these, thus this PR.

 - Avoid use of `round` with `jq` (not supported on `jq` 1.5 which is the version on bionic)
 - Fix a silly leak in a unit test.


garlick (approved):
Oh you found that disconnect one!  Sorry I didn't circle back to that and thanks!

grondo:
Thanks!

PR #5499, opened 2023-10-12
libflux: add flux_send_new()

garlick:
Problem: it's a common pattern to create a message, send it with with `flux_send()`, then destroy it; however, there is no interface to allow the message ownership to be transferred rather than copied.
    
Add `flux_send_new()` which accepts a non-const `flux_msg_t **` and implement it in the interthread connector.  Now messages can pass from one end to the other without costly duplication.

Use this function in the common `flux_rpc()` and `flux_respond()` family of functions.

This is marked WIP for now while I consider whether it's possible to use this in the broker along the message path.  It would be neat if rpcs between broker modules could transit the broker without copying!

Even just this change does seem to have a positive impact on job throughput.

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -166,6 +166,7 @@ MAN3_FILES_SECONDARY = \
 	man3/flux_shell_task_info_unpack.3 \
 	man3/flux_shell_task_cmd.3 \
 	man3/flux_service_unregister.3 \
+	man3/flux_send_new.3 \
 	man3/flux_clone.3 \
 	man3/flux_close.3 \
 	man3/flux_reconnect.3 \
--- a/doc/man3/flux_send.rst
+++ b/doc/man3/flux_send.rst
@@ -6,9 +6,13 @@ flux_send(3)
 SYNOPSIS
 ========
 
-#include <flux/core.h>
+::
 
-int flux_send (flux_t \*h, const flux_msg_t \*msg, int flags);
+   #include <flux/core.h>
+
+   int flux_send (flux_t *h, const flux_msg_t *msg, int flags);
+
+   int flux_send_new (flux_t *h, flux_msg_t **msg, int flags);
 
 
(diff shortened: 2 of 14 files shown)


garlick:
I corrected problem where I was updating a message reference count outside of the lock when it should have been inside.
Also added some unit tests.

Dropping the WIP since I think any work in the broker is going to need to be its own PR.

garlick:
Looks like caliper is always enabled in CI.  When caliper is enabled, `flux_send_new()` was falling back to `flux_send()`, which caused one of the unit tests to fail and would have prevented coverage of the new code.

To address this I just moved the call to `profiling_msg_snapshot()` to before the send call so it could be used in `flux_send_new()` without a use-after-free.  I _think_ this is OK.   If the send fails, it means caliper might see an RPC request without a response, but of course requests and responses are not evenly matched in flux anyway...

grondo:
> When caliper is enabled, flux_send_new() was falling back to flux_send(), 

Oh, I was wondering why the function always fell back to `flux_send()` if `HAVE_CALIPER` was enabled. Just out of curiosity, why was that?

garlick:
The call to `profiling_msg_snapshot()` was being called after the send, and in the `send_new` case, the message is no longer available after the send.  My realization was that it probably doesn't need to be called after the send.

grondo:
Ah, so the connector implementation of `send_new()` is required to destroy the message, it isn't done in the `flux_send_new()` wrapper? (and would it improve things if it were done that way?)

garlick:
No `send_new()` implementation doesn't destroy the message.  It takes ownership from the sender and transfers it to the other end of the channel where it's used by the receiver.

grondo:
Ok, thanks! That makes sense now.

grondo (approved):
LGTM!

garlick:
Thanks!  I'll set MWP.

PR #5500, opened 2023-10-12
python: add flux.job.list docstrings

wihobbs:
This doesn't fully document the `job.list` module but I think it's a good start.


src/bindings/python/flux/job/list.py
```diff
@@ -20,9 +20,13 @@
 
 class JobListRPC(RPC):
     def get_jobs(self):
+        """Returns all jobs in the RPC"""
         return self.get()["jobs"]
 
     def get_jobinfos(self):
+        """Yields a JobInfo object for each job in its current state
+
+        :rtype: JobInfo"""
```
grondo, line 29: prefer closing `"""` on a separate line. Also we've slowly transitioned to [Google docstring](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) style if you'd prefer that (a bit easier to read in the code)

src/bindings/python/flux/job/list.py
```diff
@@ -96,12 +103,22 @@ def get_job(self):
...
+        """
+        Returns a JobInfo object for the job.
+        """
         return JobInfo(self.get_job())
 
 
 # list-id is not like list or list-inactive, it doesn't return an
 # array, so don't use JobListRPC
 def job_list_id(flux_handle, jobid, attrs=["all"]):
+    """Watch a job by providing an ID.
```
grondo, line 115: I don't think `Watch` is the right verb here. This is just a job listing that function lists a single job by id. Perhaps just "Query job information for a single jobid".

src/bindings/python/flux/job/list.py
```diff
@@ -96,12 +103,22 @@ def get_job(self):
...
 
 # list-id is not like list or list-inactive, it doesn't return an
 # array, so don't use JobListRPC
 def job_list_id(flux_handle, jobid, attrs=["all"]):
+    """Watch a job by providing an ID. 
+
+    This is a non-blocking function that allows users to asynchronously
+    get job event data. Returns a RPC future that is fulfilled each time 
+    the job enters a new state.
+    :rtype: JobListIdRPC
```
grondo, line 120: There may be some confusion here. This function (and the corresponding `job-list.list-id` RPC) send a single response with the information the job-list module has for the job at that moment. It is not multi-response and does not send updates on each state change. (Perhaps you are confusing this with the `event_watch_async` interface, which sends a response for each event (not each state though))

Perhaps just remove this block and instead say simply:
"Sends an RPC to the job-list module to query information about the provided jobid. Use the `get_job()` or `get_jobinfo()` method on the returned `JobListIdRPC` to obtain the job data as a dict or a `JobInfo` object respectively" 

(an RPC is always asynchronous so mentioning that the function is non-blocking is probably not warranted here)

grondo (commented):
Thanks for doing this. A few comments and corrections inline.

src/bindings/python/flux/job/list.py
```diff
@@ -75,6 +79,9 @@ def job_list(
 def job_list_inactive(
     flux_handle, since=0.0, max_entries=1000, attrs=["all"], name=None, queue=None
 ):
+    """
+    List only inactive jobs
+    """
```
jameshcorbett, line 84: Following https://peps.python.org/pep-0257/#one-line-docstrings I think it's better to end with a period and to place the opening and closing quotes on a single line.

src/bindings/python/flux/job/list.py
```diff
@@ -96,12 +103,22 @@ def get_job(self):
         return self.get()["job"]
 
     def get_jobinfo(self):
+        """
+        Returns a JobInfo object for the job.
+        """
```
jameshcorbett, line 108: Same comment---single-line docstring.

jameshcorbett:
Can you add a commit body? I think 

> Problem: some flux.job.list functions are missing docstrings.
>
> Add them.

Would do it.

src/bindings/python/flux/job/list.py
wihobbs: You're right that I confused what this was doing with`event_watch`, both here and in the `job_list_id` function docstring...fixed. Thanks.

wihobbs:
Thanks guys, I think I got all your corrections.

src/bindings/python/flux/job/list.py
jameshcorbett: It looks like you left this one unaddressed @wihobbs ?

src/bindings/python/flux/job/list.py
jameshcorbett: Also unaddressed

src/bindings/python/flux/job/list.py
jameshcorbett: Ditto

jameshcorbett (commented):
LGTM but you dropped the commit body :/

jameshcorbett (approved):
LGTM, thanks @wihobbs !

jameshcorbett:
Ok to set MWP @grondo @wihobbs ?

jameshcorbett:
Checked with Hobbs, setting MWP.

PR #5504, opened 2023-10-13
libflux: make local connector built-in

garlick:
Problem: it's not very efficient to have to separately dlopen the local connector when we could just build it in.

Build it in.

PR #5505, opened 2023-10-13
connectors: avoid future ABI issues with _pad[]

garlick:
Problem: when connector operations are added, ABI compatibility between flux versions is broken.

Move the new `op->send_new()` to the end of the struct so that old libflux can operate with new connectors.

Add padding so in the future, new ops can replace padding and new libflux can operate wtih old connectors.


grondo:
FYI - this was causing `SIGILL` when tested before #5504 was merged. (@wihobbs noticed this when running the gitlab ci against just that branch). I'm trying to think of a way we can test this scenario when the local connector is built-in -- perhaps using an old version of the ssh connector from Flux built on this branch?

grondo (approved):
LGTM

garlick:
Thanks!

PR #5507, opened 2023-10-16
python: return more precise result from `flux.util.parse_datetime("now")`

### LINKED ISSUE #5501: not ok 11 - flux-watch: works with --since
I'm seeing an intermittent failure of `t2813-flux-watch.t` on the branch for #5499:
```
not ok 11 - flux-watch: works with --since
#	
#		t_since=$(flux jobs -ac 1 -no {t_submit}) &&
#		test_must_fail flux watch -v --since=$t_since >since.out 2>&1 &&
#		test_debug "cat since.out" &&
#		grep "Watching [12] job" since.out
#	
```
If I tweak the test slightly to get the `flux watch` output on failure, I get
```
flux.utils.parsedatetime: DEBUG: eval 1697204267.1642566 with context - False, False
flux.utils.parsedatetime: DEBUG: eval now with context - False, False
flux-watch: ERROR: --since=1697204267.1642566 appears to be in the future
```

grondo:
This PR fixes #5501 by adding a special case for `parse_datetime("now")`. Instead of parsing `"now"` with the `parsedatetime` module, just use `datetime.now()` which internally uses `gettimeofday(2)` and thus has sub-second precision, unlike `parsedatetime` `Calendar.parse()` which returns only second precision by default.

### DIFF
--- a/src/bindings/python/flux/util.py
+++ b/src/bindings/python/flux/util.py
@@ -346,6 +346,9 @@ def parse_datetime(string, now=None):
     if now is None:
         now = datetime.now().astimezone()
 
+    if string == "now":
+        return now
+
     if string.startswith("+"):
         timestamp = now.timestamp() + parse_fsd(string[1:])
         return datetime.fromtimestamp(timestamp).astimezone()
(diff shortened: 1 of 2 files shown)


garlick (approved):
LGTM

grondo:
Thanks! Rebased and fixed a python formatting issue in the tests. Will set MWP.

PR #5508, opened 2023-10-16
libflux: refactor internal message queues

garlick:
Problem: the `flux_msglist` class is used in connector implementations as a message queue, but it's not a great fit for that:
- it isn't thread safe
- its operations are slightly awkward for a queue (enqueue is "append", dequeue is "pop")
- it manipulates message references, making it awkward for no-copy interfaces
- it uses a zlistx internally, so insertions require a memory allocation and connectors are on critical performance path

This PR creates a new internal container class called "mpipe" which
- it is thread safe, with a MPIPE_SINGLE_THREAD option to relax this
- it has only enqeuue, dequeue, and requeue (enqueue to the other end) operations
- it uses a CCAN list internally for low overhead insertion
- it implements the pollfd/pollevents interface for reactor loop integration

Then the `loop` connector, `interthread` connector, and `flux_t` "requeue queue" are converted to use mpipe.

~~Finally, since all users of `flux_msglist_pollfd()` and `flux_msglist_pollevents()` were converted, drop those interfaces from the public `flux_msglist` class.~~

### DIFF
--- a/src/common/libflux/Makefile.am
+++ b/src/common/libflux/Makefile.am
@@ -94,6 +94,8 @@ libflux_la_SOURCES = \
 	flog.c \
 	attr.c \
 	handle.c \
+	msg_deque.c \
+	msg_deque.h \
 	connector_loop.c \
 	connector_interthread.c \
 	connector_local.c \
@@ -177,7 +179,8 @@ TESTS = test_message.t \
 	test_module.t \
 	test_plugin.t \
 	test_sync.t \
-	test_disconnect.t
+	test_disconnect.t \
+	test_msg_deque.t
 
 test_ldadd = \
 	$(top_builddir)/src/common/libtestutil/libtestutil.la \
@@ -300,6 +303,10 @@ test_disconnect_t_LDADD = $(test_ldadd)
 test_interthread_t_SOURCES = test/interthread.c
 test_interthread_t_CPPFLAGS = $(test_cppflags)
 test_interthread_t_LDADD = $(test_ldadd)
+
+test_msg_deque_t_SOURCES = test/msg_deque.c
+test_msg_deque_t_CPPFLAGS = $(test_cppflags)
+test_msg_deque_t_LDADD = $(test_ldadd)
 
 test_module_t_SOURCES = test/module.c
 test_module_t_CPPFLAGS = $(test_cppflags)
--- a/src/common/libflux/connector_interthread.c
+++ b/src/common/libflux/connector_interthread.c
@@ -35,15 +35,11 @@
 #include "ccan/list/list.h"
 #include "ccan/str/str.h"
 #include "message_private.h" // for access to msg->aux
-
-struct msglist_safe {
-    struct flux_msglist *queue;
-    pthread_mutex_t lock;
-};
+#include "msg_deque.h"
 
 struct channel {
     char *name;
-    struct msglist_safe *pair[2];
+    struct msg_deque *pair[2];
     int refcount; // max of 2
     struct list_node list;
 };
@@ -53,8 +49,8 @@ struct interthread_ctx {
     struct flux_msg_cred cred;
     char *router;
     struct channel *chan;
-    struct msglist_safe *send; // refers to ctx->chan->pair[x]
-    struct msglist_safe *recv; // refers to ctx->chan->pair[y]
+    struct msg_deque *send; // refers to ctx->chan->pair[x]
+    struct msg_deque *recv; // refers to ctx->chan->pair[y]
 };
 
 /* Global state.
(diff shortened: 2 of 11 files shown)


garlick:
Aw dang it, the pmix builder failed - doing its job!  I'll drop the commit that removes those "unused" interfaces from `flux_msglist` as apparently they were used externally.  I thought I checked this but apparently I missed it.
```
2023-10-16T23:23:46.0441937Z   CC       pmix_la-interthread.lo
2023-10-16T23:23:46.0446229Z interthread.c: In function ‘interthread_create’:
2023-10-16T23:23:46.0451280Z interthread.c:163:15: error: implicit declaration of function ‘flux_msglist_pollfd’; did you mean ‘flux_msglist_pop’? [-Werror=implicit-function-
```

garlick:
Got one builder that failed one of the issue tests `not ok 1 - t4583-free-range-test`, but the rest of the tests didn't report so maybe something else was going on in that builder.  Trying again.

src/common/libflux/test/mpipe.c
```diff
@@ -0,0 +1,281 @@
...
+    if (!(msg1_ptr = msg1 = flux_msg_create (FLUX_MSGTYPE_REQUEST)))
+        BAIL_OUT ("flux_msg_create failed");
+
+    q = mpipe_create (MPIPE_SINGLE_THREAD);
+    ok (q != NULL,
+        "mpipe_create flags=SINGLE_THREAD works");
+    flux_msg_incref (msg1);
+    ok (mpipe_enqueue (q, &msg1) == 0,
+        "mpipe_enqueue msg1 works with refcount==2");
+    flux_msg_decref (msg1);
```
grondo, line 179: valgrind detects a leak of `msg1` here:
```
==4088027== 120 bytes in 1 blocks are definitely lost in loss record 1 of 1
==4088027==    at 0x4C3CE4B: calloc (vg_replace_malloc.c:1328)
==4088027==    by 0x402AA9: flux_msg_create (message.c:91)
==4088027==    by 0x401DC1: check_single_thread (mpipe.c:170)
==4088027==    by 0x401435: main (mpipe.c:273)
```

When peeking at the code I realized I'm not quite sure how the `mpipe` interface works. Does `mpipe_enqueue` steal the reference to `msg`? But maybe not when `MPIPE_SINGLE_THREAD` is used? Anyway, it is probably obvious, but still maybe the header file could use a couple comments? (And I'm not certain how to fix the leak here, probably something obvious)

grondo: Ah, maybe `mpipe_enqueue()` is setting `*msg = NULL` here so the `flux_msg_decref()` has no effect?

grondo:
```suggestion
    flux_msg_incref (msg1);
    ok (mpipe_enqueue (q, &msg1) == 0,
        "mpipe_enqueue msg1 works with refcount==2");
    flux_msg_decref (msg1_ptr);
```

garlick: Yes, that's it, good catch.

I'll add some documentation to the header.  Also, now I'm wondering if should use the zeromq destructor style here given that messages with external references are accepted in single thread mode.   It made more sense before that feature was added.

garlick:
Forced pushed with
- `mpipe_enqueue()` and `mpipe_requeue()` no longer take a double pointer for the message argument
- reference stealing is noted in the header
- fix the memory leak
- rebased on current master

src/common/libflux/Makefile.am
```diff
@@ -94,6 +94,8 @@ libflux_la_SOURCES = \
 	flog.c \
 	attr.c \
 	handle.c \
+	mpipe.c \
```
grondo, line 97: commit message typo: triple-l in "internallly"

grondo: Also, commit message nit, is "requeue" really part of a queue abstraction? Actually doesn't bother me too much but since it stood out to me thought I'd at least mention it (maybe it is a double-ended queue or "deque")? (But even if I'm not incorrect, it doesn't bother me enough to require any changes)

grondo (approved):
Nice improvement! This LGTM, just noted one commit message typo (and a question) inline.

The only other thing that stuck out to me is that the `mpipe` name is a bit terse, and out of context a reader could get confused what it is doing, whereas a slightly longer name like `msgpipe` could alleviate that.  However, given that the interface will mostly be used where the context is obvious, I'm thinking my fears are probably unfounded...

src/common/libflux/Makefile.am
garlick: Good point, I guess technically it's an [output-restricted deque](https://en.wikipedia.org/wiki/Double-ended_queue).

Maybe it should be renamed to `msg_deque` and (looking at the nice menu of methods in the wikipedia page) maybe `msg_deque_push_back`, `msg_deque_pop_front`, and `msg_deque_pop_front`?

grondo: Eh, I thought enqueue, dequeue, and requeue, were fine and you could just update the description to a double ended queue if we wanted to be pedantic...

garlick: Renaming stuff is easy - may as well get it right.  Are you OK if I change it to what was proposed above?  I think it's actually clearer that way now that you made me think about it a bit.

grondo: I leave it up to you. I'm fine with the rename.

garlick:
OK, repushed.  Also fixed a commit typo I just noticed.

grondo (approved):
Looks great to me!

garlick:
Thanks! I'll set MWP.

PR #5509, opened 2023-10-17
split flux_requeue() into two functions

garlick:
In working on other stuff, I got annoyed with the `flux_requeue()` interface.  I know we shouldn't be in the habit of making gratuitous changes to public functions, but I didn't find this used in any of our primary repos so maybe OK?

The old prototype was
```c
/* Flags for flux_requeue().
 */
enum {
    FLUX_RQ_HEAD = 1,   /* requeue message at head of queue */
    FLUX_RQ_TAIL = 2,   /* requeue message at tail of queue */
};

/* Requeue a message
 * flags must be either FLUX_RQ_HEAD or FLUX_RQ_TAIL.
 * A message that is requeued will be seen again by flux_recv() and will
 * cause FLUX_POLLIN to be raised in flux_pollevents().
 */
int flux_requeue (flux_t *h, const flux_msg_t *msg, int flags);
```
It just seems like there ought to be two functions:
```c
/* Requeue a message
 * A message that is requeued will be seen again by flux_recv() and will
 * cause FLUX_POLLIN to be raised in flux_pollevents().
 */
int flux_requeue (flux_t *h, const flux_msg_t *msg); // add to head of queue
int flux_enqueue (flux_t *h, const flux_msg_t *msg); // add to tail of queue
```
So this makes that change.  I can easily withdraw this PR if we decide it's not advisable.

Edit: oh and how about this man page style for APIs?
https://flux-framework--5509.org.readthedocs.build/projects/flux-core/en/5509/man3/flux_requeue.html

### DIFF
--- a/doc/man3/flux_requeue.rst
+++ b/doc/man3/flux_requeue.rst
@@ -10,29 +10,31 @@ SYNOPSIS
 
   #include <flux/core.h>
 
-  int flux_requeue (flux_t *h, const flux_msg_t *msg, int flags);
+  cc $(pkg-config --cflags --libs flux-core)
 
 
 DESCRIPTION
 ===========
 
-``flux_requeue()`` requeues a *msg* in handle *h*. The message
-can be received with ``flux_recv()`` as though it arrived from the broker.
+.. function:: int flux_requeue(flux_t *h, const flux_msg_t *msg)
 
-*flags* must be set to one of the following values:
+Enqueue :c:var:`msg` on the head of the handle :c:var:`h`'s receive queue,
+placing it in front of any messages already in the queue.  If successful, a
+reference is taken on :c:var:`msg` and the handle :c:var:`h` becomes
+ready for reading just as if the message were received from the broker.
+Available messages may be read by calling :man3:`flux_recv`.
 
-FLUX_RQ_TAIL
-   *msg* is placed at the tail of the message queue.
+.. function:: int flux_enqueue(flux_t *h, const flux_msg_t *msg)
 
-FLUX_RQ_TAIL
-   *msg* is placed at the head of the message queue.
+Like :func:`flux_requeue` except the message is enqueued on the tail of the
+receive queue, placing it behind any messages already in the queue.
 
 
 RETURN VALUE
 ============
 
-``flux_requeue()`` return zero on success.
-On error, -1 is returned, and errno is set appropriately.
+These functions return zero on success.  On error, -1 is returned with errno
+set.
 
 
 ERRORS
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -167,6 +167,7 @@ MAN3_FILES_SECONDARY = \
 	man3/flux_shell_task_cmd.3 \
 	man3/flux_service_unregister.3 \
 	man3/flux_send_new.3 \
+	man3/flux_enqueue.3 \
 	man3/flux_clone.3 \
 	man3/flux_close.3 \
 	man3/flux_reconnect.3 \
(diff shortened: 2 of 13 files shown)


doc/man3/flux_requeue.rst
```diff
@@ -10,29 +10,31 @@ SYNOPSIS
 
   #include <flux/core.h>
 
-  int flux_requeue (flux_t *h, const flux_msg_t *msg, int flags);
+  cc $(pkg-config --cflags --libs flux-core)
```
grondo, line 13: [man(7)](https://man7.org/linux/man-pages/man7/man-pages.7.html) says for the SYNOPSIS section:

>  For functions, it shows any required data
>  declarations or #include directives, followed by the
>  function declaration.

I'm not sure a plain compile line covers that. But didn't we have some issue with sphinx/reStructuredText that fails to render multiple lines here? Maybe I'm getting that confused with something else.

Anyway, I'm not sure if replacing the function prototype with a compile line is the right direction? (Sorry, I probably just don't know the impetus for this change.)

grondo:
This seems fine to me... Is there a use case for `flux_enqueue()` `flux_requeue()` outside of flux-core?

Another approach would be to add two new functions:
```c
int flux_requeue_push_front (flux_t *h, const flux_msg_t *msg);
int flux_enqueue_push_back (flux_t *h, const flux_msg_t *msg);
```

Or similar, then deprecate `flux_requeue(3)` and remove it in the next version.

We may want to check for other functions in the exported API that need changes like this and get them done all in one version. (If there are any)

doc/man3/flux_requeue.rst
garlick: I was thinking of how to document some classes like idset or flux_msg_t that may have dozens of functions.  At times it is nice to have man pages broken up into one page per one or a few functions.  At other times it is nice to have the whole class in one place.  (This particular case is not a good example of that problem).

Maybe it would be better to do "class docs" as separate RTD pages and then just pick the most commonly used to make into man pages?

garlick: Oh I should check if a long function gets truncated in the man page rendering.  It apparently wraps in HTML (see PMI RFC - the Spawn call in there is ridiculously long).  Anyway maybe this is not the way to go for man pages.

garlick:
I'm not sure if there is a use case outside flux-core - maybe not!  We did use it in the KVS module but that was before we had futures for concurrency (ugh that code needs a rework).

doc/man3/flux_requeue.rst
grondo: As an example, the `printf(3)` family of functions has a SYNOPSIS like so:
```
SYNOPSIS
       #include <stdio.h>

       int printf(const char *format, ...);
       int fprintf(FILE *stream, const char *format, ...);
       int dprintf(int fd, const char *format, ...);
       int sprintf(char *str, const char *format, ...);
       int snprintf(char *str, size_t size, const char *format, ...);

       #include <stdarg.h>

       int vprintf(const char *format, va_list ap);
       int vfprintf(FILE *stream, const char *format, va_list ap);
       int vdprintf(int fd, const char *format, va_list ap);
       int vsprintf(char *str, const char *format, va_list ap);
       int vsnprintf(char *str, size_t size, const char *format, va_list ap);
```

Does this not cover the whole class in one page case? (Sorry If I'm just being slow)

garlick:
Now that you mention it, I like the idea of thinning out the publicly accessible functions that probably have no use case outside of flux-core.  I'll withdraw this PR for now and explore doing that for this function and possibly some others.

PR #5511, opened 2023-10-19
misc build system cleanup

garlick:
Misc. build system cleanups split off from other work.


grondo (approved):
Nice cleanup! I noticed that the fedora38 builder had some segfaults though :grimacing:

src/common/libflux/conf.c
```diff
@@ -45,27 +45,100 @@ struct flux_conf {
...
-    { "shell_path",     INSTALLED_SHELL_PATH,       INTREE_SHELL_PATH },
-    { "shell_pluginpath",
-                        INSTALLED_SHELL_PLUGINPATH, INTREE_SHELL_PLUGINPATH },
-    { "shell_initrc",   INSTALLED_SHELL_INITRC,     INTREE_SHELL_INITRC },
-    { "jobtap_pluginpath",
-                        INSTALLED_JOBTAP_PLUGINPATH,INTREE_JOBTAP_PLUGINPATH },
-    { "upmi_pluginpath",INSTALLED_UPMI_PLUGINPATH,  INTREE_UPMI_PLUGINPATH },
-    { "no_docs_path",   INSTALLED_NO_DOCS_PATH,     INTREE_NO_DOCS_PATH },
-    { "rundir",         INSTALLED_RUNDIR,           NULL },
+    {
```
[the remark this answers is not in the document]
garlick: Do you mean the reformatting and use of designated initailizers?  I did have to touch every entry in that array to accomplish the goal of the commit, so I figured if I use a different format (one more appropriate for the new content) then that was OK to do in one go.   If you feel strongly about it, let me know and I'll split it up.

garlick:
The fedora segfaults are a bit mystifying.  I was able to replicate in docker and found it was 
```
Thread 10 "flux-broker-0" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7f410d31d6c0 (LWP 272500)]
0x00007f411e457e5e in getenv () from /lib64/libc.so.6
Missing separate debuginfos, use: dnf debuginfo-install glibc-2.37-4.fc38.x86_64 libgcc-13.1.1-1.fc38.x86_64 libstdc++-13.1.1-1.fc38.x86_64
(gdb) bt
#0  0x00007f411e457e5e in getenv () from /lib64/libc.so.6
#1  0x00007f411e7e1220 in kvs_get_namespace ()
    at ../../../../src/common/libkvs/kvs_util.c:65
#2  0x00007f411e7de6f5 in flux_kvs_commit (h=0x7f40f0000bf0, ns=ns@entry=0x0, 
    flags=flags@entry=0, txn=txn@entry=0x7f40f000ad90)
    at ../../../../src/common/libkvs/kvs_commit.c:112
#3  0x00007f410d33c873 in restart_save_state (ctx=ctx@entry=0x7f410d31c890)
    at ../../../../src/modules/job-manager/restart.c:271
#4  0x00007f410d337b8f in mod_main (h=0x7f40f0000bf0, argc=<optimized out>, 
    argv=<optimized out>)
    at ../../../../src/modules/job-manager/job-manager.c:240
#5  0x0000000000410926 in module_thread (arg=0x228ba60)
    at ../../../src/broker/module.c:204
#6  0x00007f411e4a3907 in start_thread () from /lib64/libc.so.6
#7  0x00007f411e529774 in clone () from /lib64/libc.so.6
```
during broker shutdown.  If I manually unloaded the job manager before shutdown, the segfault would move to the next place `getenv()` was called.  Finally in an effort to install a missing glibc debuginfo package to get more clues, and finding that the debuginfo package was somehow not available for the installed glibc, I updated everything with `yum -y update` and the segfault disappeared.  Gah.

src/common/libflux/conf.c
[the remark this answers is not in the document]
grondo: I do not feel strongly about it. Since the designated initializers ended up as most of the change I was just confused and thought this was a reformatting of the structure. (Whatever the main change was it seems to be lost in the other changes)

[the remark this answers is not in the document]
garlick: Oh well that's a fair point! I was also struggling to get meaning out of the diff.  I'll fix.

garlick:
Just force pushed with
- the offending commit dropped
- the config table commit split into just the reformat, then the substantive change (as discussed)
- added one commit that drops some Makefile.am variables that are barely used
- that change also affects variables set in flux-core.pc but a quick check of other projects didn't turn up any usage

configure.ac
```diff
@@ -442,14 +442,8 @@ AC_PKGCONFIG
...
 ##
-AS_VAR_SET(fluxrcdir, $sysconfdir/flux)
-AC_SUBST(fluxrcdir)
-
-AS_VAR_SET(fluxrc1dir, $sysconfdir/flux/rc1.d)
-AC_SUBST(fluxrc1dir)
-
-AS_VAR_SET(fluxrc3dir, $sysconfdir/flux/rc3.d)
-AC_SUBST(fluxrc3dir)
+AS_VAR_SET(fluxconfdir, $sysconfdir/flux)
```
grondo: commit msg typo: "ther"

src/common/libflux/conf.c
```diff
@@ -45,27 +45,96 @@ struct flux_conf {
 static const char *conf_auxkey = "flux::conf_object";
 
 static struct builtin builtin_tab[] = {
-    { "lua_cpath_add",  INSTALLED_LUA_CPATH_ADD,    INTREE_LUA_CPATH_ADD },
```
grondo, line 48: Commit message typos: "Problen", "desigated"

src/common/libdebugged/Makefile.am
```diff
@@ -11,11 +11,6 @@ AM_CPPFLAGS = \
 noinst_LTLIBRARIES = libdebugged.la
 libdebugged_la_SOURCES = debugged.c debugged.h
 
-libdebugged_la_CPPFLAGS = \
```
grondo, line 14: Commit message: unnecessary apostrophe in `Makefile.am's`

Typo: "The caused the library" -> "This caused the library" ?

src/cmd/Makefile.am
```diff
@@ -13,7 +13,6 @@ AM_CPPFLAGS = \
 	$(ZMQ_CFLAGS) \
 	$(FLUX_SECURITY_CFLAGS) \
 	$(HWLOC_CFLAGS) \
-	$(PMIX_CFLAGS) \
```
grondo, line 16: Commit message typo: "refrence"

grondo: Also, maybe missing a leading "The" in problem statement?

grondo (approved):
LGTM! Just noticed a few commit message typos.

garlick:
Oof, sorry to give you so many typos!  All fixed and re-pushed.  I'll set MWP.

Thanks!

PR #5512, opened 2023-10-20
shell: drop job shell standalone mode

### LINKED ISSUE #5510: shell: remove standalone mode?
We have a `--standalone` option for the job shell which was added in the early days to allow a certain set of tests without actually executing the shell in a Flux instance. In standlone mode, _R_ and jobspec have to be provided on the command line and of course other mechanisms that depend on a Flux instance are not possible, and thus there are many conditional code paths introduced by this option.

This option might be more necessary if it were difficult to set up and launch Flux instances for testing, but since we have `flux start --test-size=N` it seems that any test now using standlone mode could easily be transferred to running within a test instance. If the `--standalone` option were dropped, this could streamline code in the job shell.

Posting an issue for further thoughts.

grondo:
This PR drops the job shell `-s, --standalone` option (along with associated options meant to be used in standalone mode).

Tests are removed where appropriate and otherwise updated to use `flux run` instead of launching the job shell directly.

flux-sched testsuite will fail here until flux-framework/flux-sched#1102 is merged.

Fixes #5510


t/Makefile.am
```diff
@@ -458,7 +458,6 @@ check_LTLIBRARIES = \
 	shell/plugins/conftest.la \
```
garlick, line 458: Commit message for ` testsuite: avoid shell standalone mode in initrc tests`
- `flux_shell_setop(3)` missing `t`
- closing paren and period are reversed

garlick (approved):
This really cleans up a lot!  LGTM!

grondo:
Thanks! I'll set MWP now that flux-framework/flux-sched#1022 has been merged.

PR #5513, opened 2023-10-21
libflux: deprecate flux_requeue()

garlick:
Problem: [flux_requeue(3)](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man3/flux_requeue.html) is not needed outside of flux-core and has an awkward function signature.

Solution: tag `flux_requeue()` as deprecated and add replacement internal functions `handle_requeue_push_front()` and `handle_requeue_push_back()`.  Update in-tree users to call the new functions.

Note: this PR is based on top of #5514.  The original bit here consists of the top 7 commits.


garlick:
This probably needs to be split into two PRs.  I'll go ahead and open a separate one for the build system fixes and rebase this one on that.

garlick:
OK, this was split into two PRs and the description and title of this one edited to reflect its constrained purpose.

garlick:
This one's going to need some work after #5514 is reworked so please hold off reviewing.

garlick:
Dropping this for now.   Will rework later.

PR #5514, opened 2023-10-21
build: reduce junk content and false dependencies in DSO modules

garlick:
Problem: various DSO plugins  include large convenience libraries such as `libflux-internal.la` in `_la_LIBADD`.  This instructs the linker to include all the objects in the convenience library in the shared object, which bloats its size and potentially adds inappropriate dependencies.

Work through broker modules, (installed) jobtap plugins, and bindings DSOs to ensure `_la_LIBADD` includes something close to the minimum object set required to get the job done.

This was peeled off of #5513 and improved to cover the bindings DSOs.

### DIFF
--- a/src/bindings/lua/Makefile.am
+++ b/src/bindings/lua/Makefile.am
@@ -44,7 +44,7 @@ luamod_ldflags = \
 
 luamod_libadd = \
 	$(top_builddir)/src/common/libflux-core.la \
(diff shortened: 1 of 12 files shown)


garlick:
@grondo found a much better way of solving this, by adding `-Wl,--gc-sections` to the LDFLAGS, so I'm going to rework this and resubmit.

PR #5515, opened 2023-10-23
cron: handle ENOMEM without asserting

garlick:
Problem: `cronodate_create()` calls `xzmalloc()` which asserts on ENOMEM, but cronodate is used in a broker module, so asserting is inappropriate.

Have `cronodate_create()` call `cmalloc()` and return NULL on failure. Ensure the return value is checked all along the call path from the cron broker module:
```
cron_entry_create()
 -> cron_datetime_create()
  -> datetime_entry_from_json()
   -> datetime_entry_create()
    -> cronodate_create()
```
(This was peeled off of #5514)


grondo (approved):
Good catch! LGTM!

PR #5517, opened 2023-10-23
man3: improve HTML formatting of man pages

garlick:
This is a large amount of change but it should be entirely formatting.  The change is simply to use `:func:`, `:type:`, and `:var:` to format functions, types, and parameters/variables rather than the monospace red literal highlighting we were (mostly) using for those things to date.  The effect is to make the HTML rendering more consistent and easier on the eye.


grondo:
I wonder if we should additionally set [primary_domain](https://sphinx-doc-zh.readthedocs.io/en/latest/config.html#confval-primary_domain) in `conf.py` to `None` (instead of the default, which is `py`). This might catch authors that forget to set the default domain to `c` or otherwise use `:func:` etc. without a domain.

garlick:
I wasn't sure how that would affect the autodoc (er, the thing that creates the python binding docs from inline source, if that's what it's called) but we can try it and see!

grondo:
Oh, that's a good point.

grondo (approved):
Thanks for all the work! Much improved IMO!

garlick:
I tried setting the default domain to `None` and didn't see any effect on the autogenerated python docs  (based on a cursory check).  However,  it seems like it might be surprising to people working on the python docs to find that the domain needs to be specified in our docs and not other projects that use the default domain?

Meh I could be convinced otherwise.  However, I'll go ahead and set MWP on this so mergify can do its thing off hours and we can follow up later if it seems like good idea later, in the light of day.

PR #5520, opened 2023-10-27
flux job info: drop multiple key support, clean up code, add man page entry

garlick:
Following up on #5464, this tries to simplify`flux job info`.  For a start, it uses `job-info.update-lookup` to fetch the current R instead of replaying the eventlog on the original R.  It also cleans up the `flux-job(1)` man page and adds an entry for `flux job info` with descriptions of each available key.

### DIFF
--- a/doc/man1/flux-job.rst
+++ b/doc/man1/flux-job.rst
@@ -33,6 +33,9 @@ SYNOPSIS
 
 **flux** **job** **purge** [*OPTIONS*] [*id...*]
 
+**flux** **job** **info** [*OPTIONS*] *id* *key*
+
+
 DESCRIPTION
 ===========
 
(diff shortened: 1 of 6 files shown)


garlick:
I woke up this morning remembering some sphinx docs I came across and reworked the man page changes.  Just pushed (sphinx refs included in the commit bodies)

doc/man1/flux-job.rst
```diff
@@ -379,5 +379,5 @@ RESOURCES
 
 Flux: http://flux-framework.org
 
-RFC 34: Flux Task Map: https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_34.html
+:doc:`rfc:spec_34`
```
grondo: I don't understand this sentence in the commit message:

> The man reference is now just the title but man doesn't support hyperlinks and the title can be googled

Oh, does this mean the reference to RFC 34 in the manpage when using `man(1)` is only a title?

grondo (approved):
Great cleanup! Hopefully there are no users of multiple keys with `flux job info`, but that seems unlikely.

doc/man1/flux-job.rst
grondo: oops, my comment says it is outdated, so feel free to ignore if I commented on an older version of this PR

grondo:
Oh, also suggest updating the PR title - the capital "Flux job info" could be confusing, also the main change here seems to be the removal of the ability to request multiple keys, so maybe that fact should be included in the title so it appears in release notes?

garlick:
Does that man page HTML formatting look like something we should use for all of the pages?  I was sort of putting this forward as an example.  Maybe this PR shouldn't do that, now that I think of it...

garlick:
Repushed with just the new man section reformatted.
Also I used the sphinx doc reference for the RFCs but included the full URL also.

I'll follow up with a PR that changes formatting across section 1 and we can discuss what we want RFC references to look like in that one.

garlick:
I'll just go ahead and set MWP since the code changes were approved and I plan to submit a broader cleanup in man1 soon.

doc/man1/flux-job.rst
grondo: No that is fine, it just wasn't clear (to me) from the commit message (maybe just change to "The reference in the man page is now just the RFC title ..." though not necessary if you'd rather not IMO)

PR #5521, opened 2023-10-29
man1: improve HTML formatting of man pages

garlick:
This adds sphinx markup to our section 1 manual pages to improve HTML rendering and cross referencing.  Most of this was already presented in #5520. 

This should be purely formatting changes, although some line breaks in the `.rst` files needed to be adjusted to avoid making lines too long.


doc/man1/flux-alloc.rst
```diff
@@ -5,7 +5,6 @@
 flux-alloc(1)
 =============
 
-
```
grondo, line 8: commit message:
"cross-reference tags for options are relative" :arrow_right: "...options that are relative"?

grondo (approved):
Excellent! I appreciate the amount of work that went into this update and especially the thorough description of the changes and included references in the commit messages.

doc/man1/flux-alloc.rst
garlick: Oh, actually it's the tags that are relative not the options.  I'll reword, it's an awkward sentence.

garlick:
Thank you!  I fixed that commit message and I'll set MWP.

doc/man1/flux-alloc.rst
grondo: Ah, now I see, sorry!

PR #5522, opened 2023-10-30
support duration update for running jobs

grondo:
This PR adds full support for updating the duration/expiration of running jobs.

Updates to duration of running jobs are permitted by the update-duration plugin when made by the instance owner only. Running job duration updates are then handled as a special case in the job-manager update service. The update service sends the RFC 27 `sched.expiration` RPC and defers posting of `jobspec-update` and `duration-update` events until the RPC returns successfully.

The job execution system (which handles signaling jobs at their expiration) and the job shell (which uses the expiration for support of the `--signal` option) are updated to watch R instead of fetching it once at the beginning of the job. On updates, the expiration 

Additionally, the resource module is switched from `job-info.lookup` to `job-info.update-watch` so that it can register updates to the instance expiration. When an update in R occurs, if the new expiration does not equal the old expiration, then the resource module posts a `resource-update` event to the `resource-eventlog` after committing the new R to the KVS, and sends a response to the `sched.acquire` RPC containing the new expiration. Finally, sched-simple is extended to set the updated expiration on its internal resource set (This is used to set a default expiration for newly submitted jobs).

The job-manager update service is also extended to watch the `resource.R` KVS key. When R is updated, and the expiration has changed, then the list of running jobs is traversed and any job with an expiration matching the previous expiration has a `resource-update` event automatically applied to adjust to the new instance expiration. This allows updates in the expiration of running instance to propagate to all jobs (recursively applied if any running jobs are also instance)

This is a WIP while I think about if more testing is needed and work out any kinks found by CodeQL or other CI tests, but I thought it would be useful to post this early since there is a lot of stuff being tied together here (and perhaps some poor choices were made when trying to get this done). 

Perhaps this also could be split into multiple PRs, but leaving anything out may result in only a partial solution.

TODO before removing WIP:
- [x] resolve racy tests in `t2292-job-update-running.t`. Specifically the job-exec expiration test.
- [x] test with Fluxion

### DIFF
--- a/doc/man1/flux-update.rst
+++ b/doc/man1/flux-update.rst
@@ -75,7 +75,9 @@ SPECIAL KEYS
   Updates of the job ``duration`` can take the form of of *[+-]FSD*, where
   ``+`` or ``-`` indicate an adjustment of the existing duration, and *FSD*
   is any string or number in RFC 23 Flux Standard Duration. Examples include
-  ``60``, ``1m``, ``1.5h``, ``+10m``, ``-1h``.
+  ``60``, ``1m``, ``1.5h``, ``+10m``, ``-1h``. Updates to the duration of
+  a running job may be allowed (instance owner only) and are handled as a
+  special case. For details see `DURATION UPDATE OF A RUNNING JOB`_ below.
 
 *attributes.system.queue*, *queue*
   Updates of a pending job's ``queue`` to another enabled queue may
@@ -85,6 +87,39 @@ SPECIAL KEYS
 
 *name*
   Alias for job name, i.e. ``attributes.system.job.name``
+
+
+DURATION UPDATE OF A RUNNING JOB
+================================
+
+As a special case, an update of the duration of a running job may be allowed
+when performed by the instance owner and validated by the scheduler. When
+a running job's duration is updated, this triggers a ``resource-update``
+event in the main eventlog which contains the updated resource set
+(R) expiration time. This event then triggers the following actions to
+propagate the updated job expiration through the system:
+
+ - The updated expiration is forwarded to the job execution system which
+   modifies its expiration timer for the job.
+ - If the user enabled the job shell option to signal a job before the
+   time limit, then the job shell receives the updated R and resets this
+   timer.
+ - If the job is a subinstance of Flux (e.g. started with :man1:`flux-alloc`
+   or :man1:`flux-batch`), the instance resource module also receives
+   notification of the updated R, replaces its internal copy, writes
+   the update to ``resource.R`` in the KVS, and issues a ``resource-update``
+   event to the ``resource.eventlog``.
+ - The ``resource-update`` event in the eventlog results in a new
+   RFC 28 ``resource.acquire`` response to the scheduler containing the
+   updated expiration.
+ - The instance scheduler accepts the new response and updates its
+   internal resource set representation.
+ - The update service in the job-manager is notified of the R update,
+   and if the expiration has been increased, it walks the list of running
+   jobs and issues a ``resource-update`` event for any job which had
+   its expiration set based on the previous instance expiration.
+ - For any job that has its expiration extended due to an instance expiration
+   update, all these steps are repeated.
 
 EXIT STATUS
 ===========
--- a/src/modules/job-exec/job-exec.c
+++ b/src/modules/job-exec/job-exec.c
@@ -448,6 +448,9 @@ static int jobinfo_set_expiration (struct jobinfo *job)
         return -1;
     }
 
+    flux_watcher_destroy (job->expiration_timer);
+    job->expiration_timer = NULL;
+
     /* Timelimit disabled if expiration is set to 0.
      */
     if (expiration == 0.)
@@ -458,6 +461,8 @@ static int jobinfo_set_expiration (struct jobinfo *job)
      *  remaining, we should be as accurate as possible.
      */
     now = flux_reactor_time ();
+    if (job->t0 == 0.)
+       job->t0 = now;
 
     /* Adjust expiration time based on delay between when scheduler
      *  created R and when we received this job. O/w jobs may be
@@ -466,12 +471,14 @@ static int jobinfo_set_expiration (struct jobinfo *job)
      *  start events.
      */
     if (starttime > 0.)
-        expiration += now - starttime;
+        expiration += job->t0 - starttime;
 
     offset = expiration - now;
-    if (offset <= 0.) {
-        jobinfo_fatal_error (job, 0, "job started after expiration");
-        return -1;
+    if (offset < 0.) {
+        /*  Expiration has already passed. Just set offset to 0. so that
+         *  the expiration timer will fire immediately.
+         */
+        offset = 0.;
     }
     if (!(w = flux_timer_watcher_create (flux_get_reactor(job->h),
                                          offset,
--- a/src/modules/job-exec/job-exec.c
+++ b/src/modules/job-exec/job-exec.c
@@ -106,6 +106,8 @@
 
 static double kill_timeout=5.0;
 
+#define DEBUG_FAIL_EXPIRATION 1
+
 extern struct exec_implementation testexec;
 extern struct exec_implementation bulkexec;
 
@@ -1228,6 +1230,12 @@ static void expiration_cb (flux_t *h,
     double expiration;
     struct jobinfo *job;
 
+    /*  Respond with error if module debug flag set for testing purposes
+     */
+    if (flux_module_debug_test (h, DEBUG_FAIL_EXPIRATION, false)) {
+        errno = EINVAL;
+        goto error;
+    }
     if (flux_request_unpack (msg,
                              NULL,
                              "{s:I s:F}",
--- a/src/modules/job-exec/job-exec.c
+++ b/src/modules/job-exec/job-exec.c
@@ -1218,6 +1218,43 @@ error:
         flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
 }
 
+static void expiration_cb (flux_t *h,
+                           flux_msg_handler_t *mh,
+                           const flux_msg_t *msg,
+                           void *arg)
+{
+    struct job_exec_ctx *ctx = arg;
+    flux_jobid_t id;
+    double expiration;
+    struct jobinfo *job;
+
+    if (flux_request_unpack (msg,
+                             NULL,
+                             "{s:I s:F}",
+                             "id", &id,
+                             "expiration", &expiration) < 0)
+        goto error;
+
+    if (!(job = zhashx_lookup (ctx->jobs, &id))) {
+        errno = ENOENT;
+        goto error;
+    }
+    resource_set_update_expiration (job->R, expiration);
+    if (jobinfo_set_expiration (job) < 0)
+        goto error;
+    flux_log (h,
+              LOG_DEBUG,
+              "updated expiration of %s to %.2f",
+              idf58 (job->id),
+              resource_set_expiration (job->R));
+    if (flux_respond (ctx->h, msg, NULL) < 0)
+        flux_log_error (h, "error responding to expiration update RPC");
+    return;
+error:
+    if (flux_respond_error (h, msg, errno, NULL) < 0)
+        flux_log_error (h, "expiration_cb: flux_respond_error");
+}
+
 static void job_exec_ctx_destroy (struct job_exec_ctx *ctx)
 {
     if (ctx == NULL)
@@ -1365,6 +1402,7 @@ static const struct flux_msg_handler_spec htab[]  = {
        critical_ranks_cb,
        FLUX_ROLE_USER
     },
+    { FLUX_MSGTYPE_REQUEST, "job-exec.expiration", expiration_cb, 0 },
     FLUX_MSGHANDLER_TABLE_END
 };
 
--- a/src/modules/job-manager/update.c
+++ b/src/modules/job-manager/update.c
@@ -49,8 +49,17 @@
  * in the plugin out arguments. The updates key follows the same format as
  * the RFC 21 jobspec-update event and the update request defined here.
  *
+ * As a special case, if a job is running and a duration update is
+ * being applied, the update service will send a sched.expiration RPC to
+ * the scheduler to ensure the expiration can be adjusted. If this RPC
+ * fails with an error other than ENOSYS, then the update is rejected.
+ *
  * If all steps above are successful, then a `jobspec-update`event is
  * posted for the job and a success response sent to the caller.
+ *
+ * If a job is running, and the update results in a change in `R`, then
+ * a resource-update event MAY also be emitted for the job. Currently,
+ * only an update of the expiration in R is supported.
  *
  * FUTURE WORK
  *
@@ -63,6 +72,8 @@
 #include "config.h"
 #endif
 #include <flux/core.h>
+
+#include "src/common/libutil/errprintf.h"
 
 #include "update.h"
 #include "job-manager.h"
@@ -78,6 +89,7 @@ struct update {
 struct update_request {
     void *handle;                 /* zlistx_t handle                       */
     flux_future_t *feasibility_f; /* feasibility request future            */
+    flux_future_t *expiration_f;  /* sched.expiration request future       */
     struct update *update;        /* pointer back to update struct         */
     const flux_msg_t *msg;        /* original update request msg           */
     struct flux_msg_cred cred;    /* update request credentials            */
@@ -91,6 +103,7 @@ static void update_request_destroy (struct update_request *req)
     if (req) {
         int saved_errno = errno;
         flux_future_destroy (req->feasibility_f);
+        flux_future_destroy (req->expiration_f);
         flux_msg_decref (req->msg);
         job_decref (req->job);
         free (req);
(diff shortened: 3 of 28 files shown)


src/modules/sched-simple/sched.c
```diff
@@ -694,6 +734,15 @@
         flux_log_error (ss->h, "failed to update resource state");
         goto err;
     }
+
+    if (expiration >= 0. && ss->rlist->expiration != expiration) {
```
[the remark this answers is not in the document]
grondo: This one's ok. We only want to update the scheduler's expiration if it doesn't exactly equal the new expiration.

grondo:
I figure a description of what happens when  a duration update is requested for a running batch job would be instructive here:

1. An admin issues `flux update JOBID duration=+1h` to extend the duration of a user's batch job.
2. The job manager update service treats the update request like any other and calls the `job.update.attributes.system.duration` plugin callback topic.
3. The new version of the `update-duration` plugin allows the update for the running job only if the request is made by the instance owner. (Note: this means these updates can be disabled by removing the `.update-duration` builtin plugin if for some reason that is desired)
4. The update service now checks if an allowed update is for a running job and the update is for `attributes.system.duration`. If so, then a `sched.expiration` RPC is issued to the scheduler.
5. If the `sched.expiration` RPC returns with success, then the update service posts both a `jobspec-update` and `resource-update` event. The new expiration is calculated from the duration by using `starttime` in the job's R.
6. Once that is all complete, a response is sent to the original update request.
7. Meanwhile, the job-exec module gets a response from the `job-info.update-watch` RPC with the updated R and adjusts its internal expiration timer accordingly.
8. The job shell also watches for updates in R if the user requested that their job be signaled N seconds before expiration. It uses the new expiration to adjust its internal timer as well.
9. If the job is a subinstance of Flux, the instance resource module also receives notification of the updated R and replaces its internal copy and writes the new R to `resource.R`, and issues a `resource-update` to the resource eventlog.
10. The acquire subsystem of the resource module sees the `resource-update` event and sends the new expiration in a `resource.acquire` response to the scheduler.
11. The scheduler in the subinstance accepts the new expiration and updates its internal resource set.
12. The job manager update service watches for updates in `resource.R` and when an expiration update occurs, it traverses the list of running jobs and issues `resource-update` events on any that have apparently had their expiration set to the old instance expiration.
13. For each job that has its expiration extended due to an instance expiration update, goto 7.

garlick:
> The job manager update service watches for updates in resource.R and when an expiration update occurs, it traverses the list of running jobs and issues resource-update events on any that have apparently had their expiration set to the old instance expiration.

I'm probably confused but testing with fluxion, and just looking at instance expiration propagation,
- if I do a `flux alloc -t 30m`, its `resource.R` says (expiration - start) = 1698700302 - 1698697902 =  2400 = 40m
- Then if I run a job within it with no expiration, I get 1698700934 - 1698699134 = 1800 = 30m.

Strange!  Where is that extra 10m coming from?  But also noteworthy here is the expration time of the job and the expiration time of the enclosing instance are not the same, so it would presumably not be extended if we updated the allocation's duration.

src/shell/info.c
```diff
@@ -45,33 +43,80 @@ static int lookup_job_info_get (flux_future_t *f,
...
+    info->R = json_incref (R);
+    R = tmp_R;
+
+    tmp_rcalc = info->rcalc;
+    info->rcalc = rcalc;
+    rcalc = tmp_rcalc;
+out:
+    rcalc_destroy (rcalc);
+    json_decref (R);
+    flux_future_reset (f);
```
garlick, line 99: The error handling looks incomplete here.  `flux_future_reset()` is called when `flux_rpc_get_unpack()` fails, but the future won't ever be fulfilled again.  Also no error is propagated to `R_update_cb()` which proceeds to call the plugstack.  Is that a problem?

src/shell/info.c
```diff
@@ -109,20 +165,26 @@ static int shell_init_jobinfo (flux_shell_t *shell, struct shell_info *info)
...
-    if (!(info->R = json_loads (R, 0, &error))) {
-        shell_log_error ("error parsing R: %s", error.text);
-        goto out;
-    }
-    if (!(info->rcalc = rcalc_create_json (info->R))) {
-        shell_log_error ("error decoding R");
+
+    /*  Get initial version of R from initial job-info watch response:
+     */
+    resource_watch_update (info);
```
garlick, line 179: Might want to flag in the comment that this is a synchronous RPC (although I guess it's pretty obvious!)

garlick (commented):
I have to step away so I thought I'd just get these couple initial code comments out there...

src/shell/info.c
grondo: > flux_future_reset() is called when flux_rpc_get_unpack() fails, but the future won't ever be fulfilled again

This allows a job to continue running if there's an error updating the shell's internal R. It doesn't seem like it should be fatal -- perhaps there will not even be any side effects. I'm not sure why the future would never be fulfilled again. If R is updated again, presumably the future would be fulfilled again?

> Also no error is propagated to R_update_cb() which proceeds to call the plugstack. Is that a problem?

Yeah, I suppose that should be fixed. The plugstack would be called without an R update which should have no effect but it is extra work for no reason at the worst case.

src/shell/info.c
grondo: `resource_watch_update()` doesn't send an RPC, but I'll add 'synchronously' to that comment before it. Note that the job shell synchronously fetched R here before here as well...

src/shell/info.c
garlick: A streaming rpc is terminated by an error response so no further responses are possible. But the point is somewhat moot if nothing is waiting on it and it should not be fatal.

We will need to revisit once we support reconnecting after instance restart.

grondo: An unpack error could also occur here and is the presumed cause of any error here. In that case I assume the streaming RPC would not be terminated? The job cannot end until the shell exits so `ENODATA` at least is not expected.

grondo: I did update the function to return -1 on error and if there is an error on the first response (the one fetched synchronously) then that is indeed a fatal error. So thanks for catching that it is was actually a pretty severe bug!

grondo: I pushed a fixup commit for this issue. `resource_watch_update()` now returns -1 on failure. If this function fails on the first response, then that is a fatal error, o/w the error is ignored (though the issue is logged).

grondo:
Pushed a couple fixup commits:
 - Address issue @garlick noted with the shell's handling of errors from `job-info.update-watch` response
 - Change the method by which the job manager decides to extend the expiration of jobs after an instance expiration update. Instead of directly comparing the job expiration to the old instance expiration, an update is performed whenever the job doesn't have a duration set in jobspec (or the duration was `0.`) _and_ the job's R has a nonzero expiration. This implies that the expiration was chosen by the scheduler based on the instance expiration, and should work around any creative assignment of expiration in R by schedulers.

src/modules/job-manager/update.c
```diff
@@ -268,6 +375,55 @@ static int update_feasibility_check (struct update *update,
...
+    struct update_request *req = NULL;
+    flux_future_t *f = NULL;
+    double expiration, duration;
+
+    if (json_unpack (updates,
+                     "{s:F}",
+                     "attributes.system.duration", &duration) < 0) {
+        errprintf (errp, "failed to unpack attributes.system.duration");
+        return -1;
+    }
```
garlick, line 399: In testing this PR with fluxion, I noted that this error was popping up:
```
Oct 31 09:28:57.565341 job-manager.err[0]: failed to unpack job ƒ963hnAb data for expiration update
```

src/shell/info.c
garlick: Is there a potential double-free  on `R` if `rcalc_create_json()` fails?  Could the bottom half of this function be simplified to
```c
    json_decref (info->R);
    info->R = json_incref (R);
    rcalc_destroy (info->rcalc);
    info->rcalc = rcalc;
    rc = 0;
out:
    flux_future_reset (f);
    return rc;
```

garlick (commented):
I completed a pass through the code and didn't spot anything but this one error path that needs attention.  I went kinda fast but I think things are looking pretty good.

Not approving yet since I assume I'll be making another pass after you take off the WIP.

grondo:
Based on some offline discussion with @garlick, I've just force-pushed an update that modifies how the job-exec module is notified of job expiration updates. In the previous iteration of this PR, the job-exec module used the `job-info.update-watch` service to watch for R updates. However, this resulted in the module getting an extra RPC response for every job since he job-manager already includes R in the initial `start` request, and the job-info module will always reply at least to the `job-info.update-watch` RPC with the initial version of R.

In the most recent version of this branch, the job-manager now sends an `<exec>.expiration` RPC to the job-exec service after a `resource-update` event is posted. There did seem to be a noticeable impact on throughput with the old method even for a workload with no updates, and this change seems to resolve that AFAICS.

Also, the testing has been expanded a bit, but I'll continue to work on that.

src/modules/job-manager/update.c
```diff
@@ -588,11 +592,125 @@ static void send_error_responses (struct update *update)
...
+
+    flux_log (h,
+              LOG_INFO,
+              "expiration updated from %.2f to %.2f",
+              old_expiration,
+              update->instance_expiration);
+
+    /*  Otherwise, check each running job for an expiration that exactly
+     *  matches the old expiration, and adjust expiration if so
+     */
```
garlick, line 647: Comment out of date?

src/modules/job-manager/update.c
```diff
@@ -588,11 +592,125 @@ static void send_error_responses (struct update *update)
...
+                  LOG_ERR,
+                  "failed to unpack current instance expiration: %s",
+                  error.text);
+    json_decref (o);
+}
+
+/*  An update to resource.R has occurred. If expiration > previous
+ *  expiration, then adjust expiration of all running jobs that had
+ *  their expiration set to exactly the previous expiration:
+ */
```
garlick, line 621: Comment out of date?

The commit message also looks out of date.

src/modules/job-manager/update.c
```diff
@@ -127,14 +140,77 @@ update_request_create (struct update *update,
...
+        */
+        double starttime = -1.0;
+        if (!job->R_redacted
+            || json_unpack (job->R_redacted,
+                            "{s:{s:F}}",
+                            "execution",
+                            "starttime", &starttime) < 0
+            || starttime <= 0.) {
+            errprintf (errp, "unable to get job starttime");
+            return -1;
```
garlick, line 165: No need to initialize `starttime` to -1 if it's not tagged as optional in the unpack.

However [RFC 20](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_20.html#execution) does say it's optional...

Do we need to update the RFC to make it required?

garlick (commented):
A few more (fairly insignificant) comments.

src/modules/job-manager/update.c
grondo: Oh, that was sloppy. Sorry! will fix.

src/modules/job-manager/update.c
grondo: Note that jobs without a starttime will not be able to have their duration updated, so perhaps it does make sense to consider your suggested change to RFC 20. But the code will still operate correctly with the optional member unpack, so leaving it like that for now seems reasonable.

garlick:
One other quick comment: I'm still seeing these error logs pop up when adjusting the duration of a running job that was submitted without one:
```
Nov 01 08:53:13.977981 job-manager.err[0]: failed to unpack job ƒ57TfUyu data for expiration update
```
I think that's an expected case, and the adjustment works, so should this log message be suppressed?

grondo:
> `Nov 01 08:53:13.977981 job-manager.err[0]: failed to unpack job ƒ57TfUyu data for expiration update`

Hm, I wonder why `json_unpack()` is returning `< 0` here though since both keys are unpacked optionally?
```c
           if (json_unpack (job->R_redacted,
                            "{s:{s?F}}",
                            "execution",
                             "expiration", &expiration) < 0
                || json_unpack (job->jobspec_redacted,
                                "{s:{s:{s?F}}",
                                "attributes",
                                 "system",
                                  "duration", &duration) < 0) {
                flux_log (h,
                          LOG_ERR,
                          "failed to unpack job %s data for expiration update",
                          idf58 (job->id));
            }
```

garlick:
Good point, I'm not sure why the unpack is failing and causing the message to be printed.  Seems like it shouldn't!   I guess it would be good to understand why in case it's not going to do its job properly when the duration is present?

garlick:
Oh - missing a close brace on the duration format!

grondo:
Oh, good catch! I wonder how tests are passing... :grimacing:

grondo:
Ok, I've pushed another fixup for that snafu!! :facepalm:

grondo:
Added a bit of documentation to the `flux-update(1)` man page describing when an update of a running job is allowed and how the expiration update propagates through the system.

grondo:
I just squashed the fixup commits and forced a push.

Additionally, a new debug flag is added to the job-exec module to force the module to fail the `exec.expiration` RPC from the job-manager, and added a test which verifies a nonfatal job exception is emitted in this case.

grondo:
Removed WIP. This is now ready for a real review.

src/modules/job-manager/update.c
```diff
@@ -588,11 +592,132 @@ static void send_error_responses (struct update *update)
...
+
+    /*  If this is the first successful update, or the expiration was
+     *  reduced, or there are no running jobs, then there is nothing
+     *  more to do.
+     */
+    if (old_expiration == -1.
+        || (update->instance_expiration > 0.
+            && update->instance_expiration < old_expiration)
+        || update->ctx->running_jobs == 0)
+        return;
```
garlick, line 643: I think you said before that the reasoning for not reducing the expiration time on jobs when the enclosing instance expiration is reduced was that there would be no material difference in behavior since running jobs are canceled as part of instance shutdown anyway.

Would a comment about that here be useful to future explorers wondering why we don't do that?

Another useful comment might be that the motivating use case is "hotline extending the duration of a running batch job" or similar.

src/modules/job-manager/update.c
```diff
@@ -588,11 +592,132 @@ static void send_error_responses (struct update *update)
...
+
+    /*  Otherwise, check each running job to determine if an adjustment
+     *  of its expiration is required:
+     */
+    if (!(jobs = zhashx_values (update->ctx->active_jobs))) {
+        flux_log_error (h, "resource_update_cb: zhashx_values failed");
+        return;
+    }
+    job = zlistx_first (jobs);
+    while (job) {
```
garlick, line 659: Since the hash table is not modified inside the loop, could `zhashx_first()`, `zhashx_next()` be used here with less code?

src/modules/job-exec/job-exec.c
```diff
@@ -1218,6 +1218,42 @@ static void critical_ranks_cb (flux_t *h,
...
+                             "id", &id,
+                             "expiration", &expiration) < 0)
+        goto error;
+
+     if (!(job = zhashx_lookup (ctx->jobs, &id))) {
+        errno = ENOENT;
+        goto error;
+    }
+    resource_set_update_expiration (job->R, expiration);
+    (void) jobinfo_set_expiration (job);
```
garlick, line 1251: `jobinfo_set_expiration()` has lots of nice error paths - any reason not to test the return value and `goto error` if one of them catches something?

src/shell/info.c
garlick: I think my comment about the double free and possible simplification in the shell's `resource_watch_update()` was overlooked?

garlick (approved):
Wow, great work.  I had a few comments most of which are just "thoughts" - not much that I would consider must-fix.

This seems to work as advertised with fluxion on my test cluster.

src/modules/job-manager/update.c
```diff
@@ -588,11 +592,132 @@ static void send_error_responses (struct update *update)
...
+        || (update->instance_expiration > 0.
+            && update->instance_expiration < old_expiration)
+        || update->ctx->running_jobs == 0)
+        return;
+
+    flux_log (h,
+              LOG_DEBUG,
+              "expiration updated from %.2f to %.2f",
+              old_expiration,
+              update->instance_expiration);
```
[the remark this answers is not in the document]
grondo: Ok, that's a good idea. I had them at `LOG_INFO` at first but then figured someone would comment that they were really debug messages. I guess I figured incorrectly ;-)

src/modules/job-manager/update.c
grondo: Actually, now that it's been said out in the open, the current solution doesn't sound right. For one thing, `flux job timeleft` will report the incorrect value, and if the job itself is an instance, it should probably have its resource set expiration reduced so that new submitted jobs don't inherit the incorrect expiration. Not sure why I thought this was a good idea at first.

I'll remove that special case, add the suggested comment, and also add a test for reducing the expiration of a instance...

src/modules/job-manager/update.c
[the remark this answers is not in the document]
grondo: This was a little tricky when expiration is updated to or from unlimited. I chose to show `(+inf)` or `(-inf)` in this case, since `INFINITY` from `math.h` could be used instead of having to render a string. LMK if that isn't acceptable.

src/modules/job-manager/update.c
```diff
@@ -588,11 +592,132 @@ static void send_error_responses (struct update *update)
...
+    /*  Otherwise, check each running job to determine if an adjustment
+     *  of its expiration is required:
+     */
+    if (!(jobs = zhashx_values (update->ctx->active_jobs))) {
+        flux_log_error (h, "resource_update_cb: zhashx_values failed");
+        return;
+    }
+    job = zlistx_first (jobs);
+    while (job) {
+        if (job->state & FLUX_JOB_STATE_RUNNING && job->R_redacted) {
```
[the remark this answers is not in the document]
grondo: Good point. This was probably unnecessary. I've changed it to `job->state & FLUX_STATE_RUN`. The test for `job->R_redacted` was more of defensive programming, but I didn't also test for `job->jobspec_redacted` so that seems kind of a half-measure and I'll drop it.

src/modules/job-exec/job-exec.c
grondo: Yes, that makes sense. In the early version this was a noresponse RPC and this is left over from that. I'll fix.

grondo:
> I think my comment about the double free and possible simplification in the shell's resource_watch_update() was overlooked?

Yeah, I hadn't seen that comment before, and strangely I can't comment directly on your comment so I'm not sure what's going on there. I'll be sure to address that one this time around.

doc/man1/flux-update.rst
```diff
@@ -75,7 +75,9 @@ SPECIAL KEYS
   Updates of the job ``duration`` can take the form of of *[+-]FSD*, where
   ``+`` or ``-`` indicate an adjustment of the existing duration, and *FSD*
```
garlick, line 76: The doc commit title has a typo: s/flux-upate(1)/flux-update(1)/

grondo: D'oh! Fixed and pushed. Will set MWP Thanks!

grondo:
Thanks @garlick! I know this one was a pain to review.

PR #5535, opened 2023-11-06
set FLUX_PMI_LIBRARY_PATH only in the job environment when simple pmi is active

garlick:
Problem: the `flux(1)` command driver sets FLUX_PMI_LIBRARY_PATH, and the broker stores it in an attribute, but this is not helpful outside of the job enivronment.  Furthermore, the library it points to does not work with -o pmi=off.
    
Set FLUX_PMI_LIBRARY_PATH in the shell pmi plugin.

Also, nix the intel_mpi "personality" plugin and instead put I_MPI_PMI_LIBRARY in the broker blocklist.  That's the purpose I believe the plugin was serving, but it was doing it in a way that would prevent a user from setting the variable for their job.


grondo (approved):
LGTM! Nice cleanup.

garlick:
Oops the test that checks if FLUX_PMI_LIBRARY_PATH is _not_ set with `-o pmi=off` was failing in the inception test, and I realized I needed to add it to the broker block list.  Did that, rebased on current master, and forced a push.  I'll go ahead and set MWP.

Issue #3139, opened 2020-08-13
shell: need a way to add job shell plugins by default in environments like spack


grondo:
@SteVwonder commented in #3132 
>  how do you envision the best way for flux-sched to provide its own shell plugins? I have it setup in a branch right now to install my plugin to $prefix/lib/shell/plugins. When $prefix is the prefix that flux-core was installed into, it works beautifully. I'm mainly concerned about setups like Spack, where the two repos are installed into separate directories. Should flux-sched provide it's own initrc.lua that sources it's installed plugins? I hesitate to suggest yet another environment variable, but maybe paths in an env var could be added to plugin.searchpath? Thoughts?

This problem may require some thought. While I'm not opposed to another environment variable, e.g. `$FLUX_SHELL_PLUGINPATH`, environment variables are inherited, and it might be inconvenient if running a different version of Flux (e.g. from a builddir) under an existing instance that had this environment variable set.

In a normal system installation of Flux, the system `initrc.lua` (much like `/etc/profile` for a Unix shell) could easily extend the `plugin.searchpath` or load extra plugins from their absolute paths. However, with systems like spack where every package is installed in its own root, this isn't quite as feasible.

One alternative to an environment variable might be to make `plugin.searchpath` into a broker attribute.
The attribute could be set/appended to during the flux-sched or other subproject rc scripts. This way the setting stays with the current instance instead of being inherited. One drawback is that the shell would have to read this variable on startup.

The attribute could be set and extended via the "extra" rc1 scripts already provided by flux-sched and/or other projects.

SteVwonder:
> One alternative to an environment variable might be to make plugin.searchpath into a broker attribute.

I like that idea.


Is the intention to give users full control over the job shell that is launched?  Or for the top-level multi-user instance are we forcing a particular job shell and plugin set upon the user?

grondo:
> Is the intention to give users full control over the job shell that is launched? Or for the top-level multi-user instance are we forcing a particular job shell and plugin set upon the user?

￼Currently, the intention is to give users full control over the job shell (it runs as the user). A user can even run a different job shell than the default via the `attributes.system.exec.job_shell` attr in jobspec.

SteVwonder:
> Currently, the intention is to give users full control over the job shell (it runs as the user). A user can even run a different job shell than the default via the attributes.system.exec.job_shell attr in jobspec.

[...]

PR #5536, opened 2023-11-07
drop broker conf.module_path, conf.connector_path, conf.exec_path attributes

garlick:
We have some broker environment variables that have a "conf." prefix making them at least a little bit confusingly named since they have nothing to do with the toml configuration, but some also do not seem to have a strong reason for existing.  I whacked the following in this PR:
- `conf.module_path` mirrors broker's environment setting for FLUX_MODULE_PATH.  `flux --parent` uses it to fetch the parent's module path and prepend it in the subcommand environment, but now that the broker does its own search for modules rather than the `flux-module` command, this has no effect that I can see.
- `conf.connector_path` has no users
- `conf.exec_path` is used like `conf.module_path` to push parent's FLUX_EXEC_PATH in front of the local path.  This seems iffy to me as built-in commands will be the local version and non-built-in commands will be the parent version.  I could be completely off base on this one and will happily revert this if so!

These are more modern and have nothing to do with the flux(1) command front end or the environment.  They are for overriding the compiled-in shell paths.   However they appear to have no users outside of one sharness test.  I left them alone for now.
- conf.shell_initrc
- conf.shell_pluginpath

I'll mark this as a WIP and pause here for some feedback. 

p.s. this commit added the `--parent` option ca2b173d08b0def8d2a5e11daf7626db948b91f8.

### DIFF
--- a/doc/man7/flux-broker-attributes.rst
+++ b/doc/man7/flux-broker-attributes.rst
@@ -135,9 +135,6 @@ conf.connector_path
 conf.exec_path
    The value of the broker's :envvar:`FLUX_EXEC_PATH` environment variable.
 
-conf.module_path
-   The value of the broker's :envvar:`FLUX_MODULE_PATH` environment variable.
-
 conf.shell_initrc [Updates: C, R]
    The path to the :man1:`flux-shell` initrc script.  Default:
    ``${prefix}/etc/flux/shell/initrc.lua``.
--- a/doc/man7/flux-broker-attributes.rst
+++ b/doc/man7/flux-broker-attributes.rst
@@ -129,9 +129,6 @@ broker.exit-restart [Updates: C, R]
 broker.starttime
    Timestamp of broker startup from :man3:`flux_reactor_now`.
 
-conf.exec_path
-   The value of the broker's :envvar:`FLUX_EXEC_PATH` environment variable.
-
 conf.shell_initrc [Updates: C, R]
    The path to the :man1:`flux-shell` initrc script.  Default:
    ``${prefix}/etc/flux/shell/initrc.lua``.
--- a/doc/man7/flux-broker-attributes.rst
+++ b/doc/man7/flux-broker-attributes.rst
@@ -129,9 +129,6 @@ broker.exit-restart [Updates: C, R]
 broker.starttime
    Timestamp of broker startup from :man3:`flux_reactor_now`.
 
-conf.connector_path
-   The value of the broker's :envvar:`FLUX_CONNECTOR_PATH` environment variable.
-
 conf.exec_path
    The value of the broker's :envvar:`FLUX_EXEC_PATH` environment variable.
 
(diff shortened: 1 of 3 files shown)


grondo:
> conf.exec_path is used like conf.module_path to push parent's FLUX_EXEC_PATH in front of the local path. This seems iffy to me as built-in commands will be the local version and non-built-in commands will be the parent version. I could be completely off base on this one and will happily revert this if so!

We probably had a lot less builtin commands back when this was implemented. I guess the worry was that if a user is trying to interact with a parent instance of a different version, they better use commands matching the version with which they are interacting. However, this could also be done by using the path to the correct `flux(1)`, so I'm fine with this simplifying change.

grondo:
> These are more modern and have nothing to do with the flux(1) command front end or the environment. They are for overriding the compiled-in shell paths. However they appear to have no users outside of one sharness test. I left them alone for now.
>
>    - conf.shell_initrc
>    - conf.shell_pluginpath

For reference, here's the original commits that added these attributes:
```
commit 658610025e50c32bc4548ee9b5b57ddee687f4da
Date:   Sat Aug 22 00:13:27 2020 +0000

    shell: allow default initrc path to be set via broker attr
    
    Problem: There is no way to override the default shell initrc path
    for an instance, on per-job via the initrc shell jobspec option.
    
    Instead of an environment variable to set a default shell initrc
    path, allow the initrc path to be set by a broker attribute. The
    default path may then be overridden on a per-instance basis on the
    commandline or via rc1 script, and is therefore tied to the instance
    instead of being set and propagated in the environment.

commit ce863aa57903313b4efa42bc93c00e3ceca2862a
Date:   Thu Aug 20 00:01:39 2020 +0000

    shell: allow plugin searchpath to be set via broker attr
    
    Problem: There are already several ways to get 3rd party shell plugins
    to be loaded by default in the shell in a system instance -- either by
    explicit use of plugin.load() in the default shell initrc.lua, or by
    adjusting the default plugin.searchpath, or installation of the plugin
    into the shell's default plugin search path. However, this doesn't work
    when flux-core and other flux framework projects are installed via package
    management systems like spack, where each package gets its own root.
    
    Rather than using an environment variable to influence the shell's
    default plugin.searchpath, allow the searchpath to be overridden by a
    broker attribute. The default path may then be set on the commandline
    or via an rc1 script influenced via RC_EXTRA, and is therefore tied
    to the instance instead of the environment.
    
    Fixes #3139
```

As noted, more discussion is contained in #3139.

I guess the idea here was that use of a broker attribute for these settings allows them to be modified at runtime without use of an environment variable which may inappropriately be propagated throughout an instance hierarchy via inheritance. However, we could take another look at this if we're trying to wean out use of broker attributes or there is some other reason this use case is considered a bad pattern. Maybe at the very least these attributes could be renamed to avoid confusing with broker config?

garlick:
Ah thanks for providing that background on the shell stuff.  I'm fine with leaving those as is for now and considering a rename or alternate strategy at a later date.  Seems like a low priority.

I'll drop the WIP so this can get a review.

doc/man7/flux-broker-attributes.rst
```diff
@@ -129,9 +129,6 @@ broker.exit-restart [Updates: C, R]
 broker.starttime
    Timestamp of broker startup from :man3:`flux_reactor_now`.
 
-conf.connector_path
```
grondo, line 132: Commit message:   s/`conf.module_path`/`conf.connector_path`/

grondo (approved):
LGTM. Noticed one commit message typo.

garlick:
Fixed that typo and I'll set MWP - thanks!

PR #5537, opened 2023-11-07
job-manager: fix error message on duplicate plugin load

grondo:
Currently, when an attempt to load a duplicate jobtap plugin is made, the error message incorrectly reports "plugin not found".
This is due to a failure to set an appropriate errno when the plugin is already loaded. This PR fixes that issue along with some other minor cleanup.

### DIFF
--- a/src/modules/job-manager/jobtap.c
+++ b/src/modules/job-manager/jobtap.c
@@ -708,7 +708,8 @@ static int jobtap_stack_call (struct jobtap *jobtap,
     while (p) {
         int rc = flux_plugin_call (p, topic, args);
         if (rc < 0)  {
-            flux_log (jobtap->ctx->h, LOG_DEBUG,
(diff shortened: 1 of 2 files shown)


src/modules/job-manager/jobtap.c
```diff
@@ -1432,8 +1432,7 @@ flux_plugin_t * jobtap_load (struct jobtap *jobtap,
     flux_plugin_t *p = NULL;
     char *conf_str = NULL;
 
-    if (errp)
-        memset (errp->text, 0, sizeof (errp->text));
+    err_init (errp);
```
garlick, line 1437: Oh duh I never noticed that was there.   I've been a bad offender on this.

garlick (approved):
LGTM!

grondo:
I had to push a fix here to avoid truncation of the error message when loading a duplicate plugin.

Issue #5539, opened 2023-11-08
flux-submit doesn't replace `{cc}` for index 0


grondo:
Just noticed this when trying out a command example for a user:
```
$ flux submit --quiet --label-io --cc=0-1 -N1 -n1 --watch echo {cc}
f6tye4TnY: 0: 1
f6tye4TnX: 0: {cc}
```

PR #5541, opened 2023-11-08
flux-submit: fix substiution of `{cc}` when cc=0

grondo:
This PR fixes substitution of `{cc}` in `flux submit --cc/--bcc` when `cc=0`. (Issue #5539)

### DIFF
--- a/src/bindings/python/flux/cli/base.py
+++ b/src/bindings/python/flux/cli/base.py
@@ -1546,7 +1546,7 @@ class SubmitBulkCmd(SubmitBaseCmd):
 
         for i in cclist:
             #  substitute any {cc} in args (only if --cc or --bcc):
-            xargs = Xcmd(args, cc=i) if i else args
+            xargs = Xcmd(args, cc=i) if isinstance(i, int) else args
             jobspec = self.jobspec_create(xargs)
 
             #  For now, an idset argument to args.input is not supported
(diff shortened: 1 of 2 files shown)


src/bindings/python/flux/cli/base.py
```diff
@@ -1546,7 +1546,7 @@ def submit_async_with_cc(self, args, cclist=None):
 
         for i in cclist:
```
garlick, line 1547: Commit message s/falsy/false/
Unless "is falsy" should be taken as  "mostly false" in which case it can stand.

garlick (approved):
LGTM!

src/bindings/python/flux/cli/base.py
grondo: I was using [this](https://en.wiktionary.org/wiki/falsy) defintion of falsy, which I think just means "not literally False, but evaluates as False in a boolean context". I see it a lot in Python discussions, but not in the official Python docs, so I could change this to "evaluates to False" if you'd prefer.

grondo: Ok, reworded that and force pushed.

garlick: I learned a new thing!  That was fine then - I just thought maybe it was a typo and then I was trying to be funny which fell flat since the word is valid after all!  OK, anyway, all good.

PR #5546, opened 2023-11-09
build: reduce minimum jansson version to 2.9

### LINKED ISSUE #5544: jansson version requirement on SLES
`flux-core` requires `jansson` 2.10 (#3240), but SLES provides `jansson` version 2.9 (even sles15-sp5). It looks like you really just needed a feature from 2.8 but selected 2.10 because it is what was in EL7. Would you be willing to relax the version requirement to make SLES packaging easier? It compiles fine against 2.9, and I haven't found any issues in my testing to date. `make check` fails with:
```
not ok 2 - flux_open on loop works
FAIL: test_subprocess.t 2 - flux_open on loop works
#   Failed test 'flux_open on loop works'
#   at test/subprocess.c line 86.
```
but it's not clear this is related, as I don't have a "clean" build with newer `jansson` on this platform.

garlick:
Problem: the version of jansson shipped with SLES, reportedly jansson 2.9 on frontier at ORNL and same on sles15-sp5, is too old to build Flux.

It seems that 2.9 is just as good as the current minimum of 2.10 for Flux so reduce the minimum version to 2.9.

Fixes #5544

### DIFF
--- a/config/x_ac_jansson.m4
+++ b/config/x_ac_jansson.m4
@@ -1,6 +1,17 @@
+# Versions of jansson shipped by distros:
+# sles15-sp5    ships 2.9  (flux-framework/flux-core#5544)
+# centos7/TOSS3 ships 2.10 (flux-framework/flux-core#3239)
+# centos8/TOSS4 ships 2.11
+# Ubuntu 18.04  ships 2.11
+# Ubuntu 20.04  ships 2.12
+#
+# Some modern jansson features used in flux-core:
+# - json_pack ("O?")     from 2.8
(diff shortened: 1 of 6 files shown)


grondo (approved):
Excellent. The comment in `x_ac_jansson.m4` is very helpful IMO.

We do reference the 2.10 version of the Jansson docs in several manpages, should we update that to 2.9? (I could go either way on that one, so approving and will leave the decision to you)

garlick:
Ah I'll update those man pages in the morning.  Good catch.

Issue #5542, opened 2023-11-08
idea: implement something like libuv streams interface for libsubprocess writes


grondo:
As noted in #5518, libsubprocess has a set of fixed buffers (though configurable in size), with a default of 4M. When the buffer is full, `flux_subprocess_write()` returns a partial write, which pushes handling of flow control etc. onto the caller. This would mean each location at which `flux_subprocess_write()` is used would have to implement its own flow control, keep partially written buffers cached, accumulate a queue of buffers, etc. Instead the buffer/subprocess interface should probably assist more here.

Though I haven't studied in detail, the libuv [stream](https://docs.libuv.org/en/v1.x/stream.html) interface seems like a good model.
Instead of writing bytes to a subprocess (or buffer), where the data is copied to the buffer and potentially fails when the buffer is full, the write interface appears to take a vector of buffers which are passed by reference through the write interface without copying. These buffers are then queued for writing and written in order. Flow control is (apparently) handled seamlessly by the reactor, with a callback that fires once each buffer is written (likely for memory management).

This seems like it would be a good long-term solution for #5518, while also offering an advantage anywhere libsubprocess is used, instead of just being a one-off fix in the job-ingest workcrew code.

PR #5548, opened 2023-11-09
libsubprocess: do not allow short writes with `flux_subprocess_write()`

grondo:
This was pulled off some other work.

Currently, `flux_subprocess_write()` writes any bytes it can to the destination buffer, which could result in a short write (return code < `len`). However, a check for a short write is not done is most places (a check for `rc < 0` seems to be the common case), and where it is checked it is immediately promoted to a fatal error. In some cases (such as the job-ingest pipeline) a short write can be dangerous since it isn't recovered and writing only part of a line "hangs" the protocol.

In the interest of simplicity, add a check for the available buffer space in `flux_subprocess_write()` and return immediately with return code `-1` and `errno` set to `ENOSPC` if there is not enough space for all the data. Update any callers that attempt to handle a short write as a special case.

In the future we'll need a more sophisticated buffered write function anyway that can handle flow control and/or a chain of buffers instead of adding short-write handling at each caller (e.g. see #5542)

### DIFF
--- a/src/common/libsubprocess/server.c
+++ b/src/common/libsubprocess/server.c
@@ -416,10 +416,6 @@ static void server_write_cb (flux_t *h, flux_msg_handler_t *mh,
 
     if (data && len) {
         int rc = flux_subprocess_write (p, stream, data, len);
-        if (rc >= 0 && rc < len) { // short write is promoted to fatal error
-            errno = ENOSPC;
-            rc = -1;
-        }
         if (rc < 0) {
             llog_error (s,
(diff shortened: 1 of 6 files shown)


garlick (approved):
This seems like it does a good job of relieving the nastiest symptom of this failure - hanging the validator.  LGTM.

grondo:
This solves half the problem. I'll have another PR with the other half tomorrow. (Struggling a bit with how to test...)

I'll set MWP here though.

Issue #5518, opened 2023-10-25
job-ingest.err[2]: job-validator[0]: Failed: No space left on device


garlick:
Problem: job submissions were hanging on rzvernal11 but not on the other login node of that cluster.

There were three validators that had been running for about an hour that were stuck.
```
rzvernal11 /g/g0/garlick > ps -fe|grep validator
flux     2304720   58884  0 09:09 ?        00:00:01 /usr/bin/python3 /usr/libexec/flux/cmd/flux-job-validator.py --plugins feasibility,jobspec
flux     2365412   58884  0 09:37 ?        00:00:00 /usr/bin/python3 /usr/libexec/flux/cmd/flux-job-validator.py --plugins feasibility,jobspec
flux     2447748   58884  0 09:46 ?        00:00:00 /usr/bin/python3 /usr/libexec/flux/cmd/flux-job-validator.py --plugins feasibility,jobspec
```

In the `flux-dmesg` log was the following
```
2023-10-25T16:09:24.859878Z broker.err[2]: Error writing 164186 bytes to subprocess pid 2289487 stdin
2023-10-25T16:09:24.859981Z job-ingest.err[2]: job-validator[0]: Failed: No space left on device

2023-10-25T16:33:56.567386Z broker.err[2]: Error writing 164164 bytes to subprocess pid 2324900 stdin
2023-10-25T16:33:56.567488Z job-ingest.err[2]: job-validator[1]: Failed: No space left on device

2023-10-25T16:40:30.682504Z broker.err[2]: Error writing 164163 bytes to subprocess pid 2416808 stdin
2023-10-25T16:40:30.682600Z job-ingest.err[2]: job-validator[2]: Failed: No space left on device
```
with frobnicator inactivity timeouts in between.

Killing the stuck validator processes got things working again.

garlick:
Hmm, I was able to get a vaildator stuck by trying to send a large request, e.g.
```
$ dd if=/dev/random of=/tmp/z bs=1024 count=4000
$ flux run --add-file /tmp/z hostname
flux-run: ERROR: Error passing job to validator
```
after that some jobs were able to run but ultimately I got
```
$ flux run hostname
Oct 25 11:33:23.483176 job-ingest.err[0]: job-validator[0]: exited with rc=1
flux-run: ERROR: Unrecoverable error: worker unexpectedly exited
```
and jobs could run again.  However I didn't get the write errors in the log like above.

[...]

PR #5549, opened 2023-11-10
job-ingest: fix cleanup when a pipeline worker process fails

grondo:
This PR is another piece of the workaround for #5518.

If the buffer for a subprocess on the _server_ side becomes full and therefore a write fails, libsubprocess currently treats this as fatal error, kills the subprocess and returns `FLUX_SUBPROCESS_FAILED` as subprocess state.

If this occurs for a job-ingest worker, then any futures still in the queue will never be fulfilled and job submissions will appear to hang.

This PR just adds a call to `worker_unexpected_exit()` for failed worker subprocesses so that any pending futures are fulfilled with error.

### DIFF
--- a/src/modules/job-ingest/worker.c
+++ b/src/modules/job-ingest/worker.c
@@ -69,7 +69,7 @@ struct worker {
 
 static int worker_start (struct worker *w);
 static void worker_stop (struct worker *w);
-
+static void worker_unexpected_exit (struct worker *w);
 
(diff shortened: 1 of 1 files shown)


garlick (approved):
Oh!  :bulb:  It hadn't quite dawned on me that this was necessary.  Good work.

grondo:
Thanks! I've set MWP.

Issue #3071, opened 2020-07-25
build and spellcheck fail if locale isn't UTF-8


chu11:
after updating my tree with #3033, got some spellcheck errors on the docs dir

```
>make check | grep FAIL
FAIL: spellcheck 13 - spell check man1/flux-kvs.rst failed
FAIL: spellcheck 34 - spell check man3/flux_fd_watcher_create.rst failed
FAIL: spellcheck 42 - spell check man3/flux_handle_watcher_create.rst failed
FAIL: spellcheck 57 - spell check man3/flux_pollevents.rst failed
FAIL: spellcheck 86 - spell check man3/flux_stat_watcher_create.rst failed
FAIL: spellcheck 89 - spell check man3/flux_zmq_watcher_create.rst failed
# XFAIL: 0
# FAIL:  6
```

some of the files seem to have some unicode chars that got added in ... by accident?  or by design?  

https://raw.githubusercontent.com/flux-framework/flux-core/master/doc/man1/flux-kvs.rst

appears to show some unicode equivalents of "...".

Spellcheck must have passed travis, but not sure what's going on in TOSS.  Old aspell version?

chu11:
probably related, 

```
(py3) catalyst159|~/chaos/git/flux-framework/flux-core/etc 134>make
  GEN      flux/help.d/core.json
Traceback (most recent call last):
  File "./gen-cmdhelp.py", line 39, in <module>
    for line in f:
  File "/g/g0/achu/chaos/git/flux-framework/flux-core/py3/lib64/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 1308: ordinal not in range(128)
make: *** [flux/help.d/core.json] Error 1
```

seems to be hitting it on `flux-event.rst`, which appears to have some unicode in there.

chu11:
this seems to have fixed things

```
diff --git a/doc/test/spellcheck b/doc/test/spellcheck
index cb1e9ed..b46b5a7 100755
--- a/doc/test/spellcheck
+++ b/doc/test/spellcheck
@@ -34,7 +34,7 @@ for f in $*; do
     filename=$(basename $f)
     dir=$(basename $(dirname $f))
     tmpfile=$(mktemp)
-    rc=$(cat $f | $ASPELL -p $dict -n list | sort | uniq | tee $tmpfile | wc -l)
[...]

PR #5551, opened 2023-11-10
testsuite: fixes for Lassen

wihobbs:
Problem: Lassen doesn't like the words login, workflow, or infeasible, and it causes our CI spellcheck tests to break on that machine.

Add those words to the exempt word list in the docs.

Looking at previous issues, I thought that #3071 might be at play since `LC_ALL=C` is set on Lassen but not on TOSS 4 machines. Changing that to `LC_ALL=""` didn't solve the issue. 

I also tried using `SHARNESS_SKIP_TESTS=spellcheck` command [listed in the testsuite readme](https://github.com/flux-framework/flux-core/tree/master/t#skipping-tests), but it still ran the test. Do we support that command? When I git grep'd for it in core I couldn't find anywhere in our shell scripts it's used. I did find [`SKIP_TESTS` in `t/sharness.sh`](https://github.com/flux-framework/flux-core/blob/f8558d01cc1b853007a82831d94ddb90c8c71e04/t/sharness.sh#L366) but setting that variable to the same thing also didn't exempt the spellcheck test.

Also, just if anyone was curious ;) according to the [OED](https://www.oed.com/search/dictionary/?scope=Entries&q=login) "log in" or "log-in" are the official words. "workflow" and "infeasible" are also words.

### DIFF
--- a/doc/test/spell.en.pws
+++ b/doc/test/spell.en.pws
@@ -820,3 +820,6 @@ hl
 noresponse
 moldable
 aspirational
+infeasible
+login
+workflow
--- a/t/README.md
+++ b/t/README.md
@@ -128,7 +128,7 @@ to run tests under valgrind.
 Skipping Tests
 --------------
 
-The environment variable `SHARNESS_SKIP_TESTS` is a space separated
+The environment variable `SKIP_TESTS` is a space separated
 list of *patterns* that tells which tests to skip, and can either
 match the test number `t[0-]{4}` to skip an entire test script,
 or have an appended `.$number` to skip test `$number` in the


grondo:
> I also tried using`SHARNESS_SKIP_TESTS=spellcheck` command [listed in the testsuite readme](https://github.com/flux->framework/flux-core/tree/master/t#skipping-tests), but it still ran the test. Do we support that command? When I git grep'd for it in core I couldn't find anywhere in our shell scripts it's used. I did find [SKIP_TESTS in t/sharness.sh](https://github.com/flux-framework/flux-core/blob/f8558d01cc1b853007a82831d94ddb90c8c71e04/t/sharness.sh#L366) but setting that variable to the same thing also didn't exempt the spellcheck test.

The problem is that the `spellcheck` test is not a sharness test, it is under `doc` and all sharness tests are under `t`.

The README should indeed reference `SKIP_TESTS` not `SHARNESS_SKIP_TESTS`, want to add a patch to this PR to change that?

wihobbs:
All right, it's down to one consistent failure. 
```
FAIL: t5000-valgrind.t 1 - valgrind reports no new errors on 2 broker run
ERROR: t5000-valgrind.t - exited with status 1
```
And some inconsistent failures in `t2233`. I'll see if these keep showing up. 

One thing (not to take this offtrack, just noting to myself) is that `HWLOC_COMPONENTS=x86` does _not_ need to be set in Lassen's gitlab CI configuration. It was causing a lot of junk output, and the number of failing tests decreased with it removed.

grondo:
Want to resolve conflicts here and rebase on current master?

wihobbs:
Yeah, let's do that and get this closed. We can come back to `t5000` later

grondo (approved):
Nice, thanks!

wihobbs:
Ughhhhh, fixing

wihobbs:
Thanks @grondo!

PR #5552, opened 2023-11-11
flux-start(1): add more description and troubleshooting info

garlick:
This reworks the flux-start(1) man page to hopefully offer more explanation and troubleshooting tips for problems that repeatedly come up.

### DIFF
--- a/doc/man1/flux-start.rst
+++ b/doc/man1/flux-start.rst
@@ -128,7 +128,8 @@ OPTIONS
 .. option:: -v, --verbose=[LEVEL]
 
    This option may be specified multiple times, or with a value, to
-   set a verbosity level.  See `VERBOSITY LEVELS`_ below.
+   set a verbosity level (1: display commands before executing them,
+   2: trace PMI server requests in `TEST MODE`_ only).
 
 .. option:: -X, --noexec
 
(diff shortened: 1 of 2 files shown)


doc/man1/flux-start.rst
```diff
@@ -15,15 +15,107 @@ DESCRIPTION
...
+distributed system.  It is primarily useful in environments that don't run
+Flux natively, or when a standalone Flux instance is required for test,
+development, or post-mortem debugging of another Flux instance.
+
+On systems already running Flux, single-user Flux instances can be more
+conveniently started with :man1:`flux-batch` and :man1:`flux-alloc`.
+The `Flux Administration Guide
+<https://flux-framework.readthedocs.io/en/latest/guides/admin-guide.html>`_
+covers setting up a multi-user Flux "system instance", where Flux natively
+manages a cluster's resources and those commands work ab initio for its users.
```
grondo, line 29: Just a thought: perhaps it would be simpler to say something like "When already running under Flux, single-user Flux instances can be more conveniently started with..." As it is now, some readers may assume a system instance is required to use `flux alloc` and `flux batch`.

However, the reference to the Admin Guide may still be useful so my comment may be off.

grondo (approved):
This is really great! I just had one comment which is more of a question. I'm fine if his goes in as is though.

doc/man1/flux-start.rst
garlick: Great point.  I think I can reword the one sentence and leave the admin guide reference.  Just pushed that - see if it reads ok for you.

grondo: Yeah, looks good to me!

garlick:
Thanks!

Issue #5547, opened 2023-11-09
user modification of `PYTHONPATH` may cause job validator (and other Python commands) to fail


grondo:
A tool loaded by environment modules inserted the path to its Python modules first in `PYTHONPATH`. The modules in this path included a version of `urllib` incompatible with the installed version and this cause the validator to fail.

We need a general way to prevent the user's environment from causing the job validator (and potentially other Flux Python utilities that do not end up spawning a user process) to fail.

In a meeting, a few proposed solutions were:
 - add a new option to `flux(1)` that forces the correct path onto `PYTHONPATH` (currently a path only containing the Flux Python bindings (via a soft link) is prepended)
 - Prepend appropriate path to `sys.path` directly in possibly affected programs

As an alternative to the second idea above, I wonder if we could
 - create a module that forces the system or configured python path to the front of the `sys.path` array and import this first in all python based commands.
The benefit here is that the environment is not modified so it can't be inherited, so this solution could be used in submission commands.

garlick:
> create a module that forces the system or configured python path to the front of the sys.path array and import this first in all python based commands.

Ooh if that could work that sounds perfect.

grondo:
This should have been closed by #5553

Issue #574, opened 2016-02-19
Prepending to FLUX_* env vars


grondo:
When running framework project testsuites, it would be handy to have environment variables that *prepend* to the default `FLUX_MODULE_PATH` and `FLUX_EXEC_PATH`, so that the components being tested are sure to come first in these paths, and we don't inadvertently test installed versions.

Currently, we can force this behavior for some testsuite items by giving explicit paths to modules and commands for example. But a case that isn't covered is schedsrv.so loading its own modules, which uses `FLUX_MODULE_PATH` (I think).

Overall, it would just be more straightforward if we had, for example, `FLUX_*_PATH_PREPEND` environment variables that forced the paths to the front.

Any thoughts?

lipari:
Yes.  That would be nice.  flux-sched does indeed rely on FLUX_MODULE_PATH.  In fact the full list of env vars to set before testing flux-sched (in build tree) is:
```
export LUA_PATH="$HOME/flux-sched/rdl/?.lua;${LUA_PATH};;"
export LUA_CPATH="$HOME/flux-sched/rdl/.libs/?.so;${LUA_CPATH};;"
export FLUX_MODULE_PATH=$HOME/flux-sched/sched
```

grondo:
Yeah, I was looking into that and realized since `FLUX_MODULE_PATH` is appended to, on a system with `flux-sched` installed to the default module path, `sched` will end up picking up the installed version...

grondo:
Ugh, I also just noticed `flux --exec-dir` is broken since the conversion to optparse (Sorry!).

Obviously, there is not a recent use case for this option. Could we drop it in favor of `FLUX_EXEC_PATH_PREPEND` or similar?

garlick:
I'm good with dropping `--exec-dir`.

grondo:
Hm, github didn't pick up that this issue should have been autoclosed by 9090a14.
Closing.

dongahn:
Weird... perhaps github didn't like "issue" between "Resolves" keyword and #<issue>...

PR #5553, opened 2023-11-11
make flux Python commands more resilient to changes in PYTHONPATH

grondo:
This is stab at fixing #5547 with a new module `fixsyspath` which can be imported first in all Python utilities to move the standard and Flux Python module paths back to the front of `sys.path`.  This works around possible changes to `PYTHONPATH` that put incompatible Python modules before the system modules.

Still need to add some automated tests here, but I did verify this works around the specific case documented in the issue:

On current master:
```console
$ module load omniperf
$ flux alloc -n1 flux run hostname
flux-run: ERROR: Unrecoverable error: worker unexpectedly exited
Nov 10 16:58:58.347767 broker.err[0]: rc2.0: flux run hostname Exited (rc=1) 0.4s
Nov 10 16:58:58.348470 job-ingest.err[0]: job-validator[0]: exited with rc=1
[detached: session exiting]
```

On this branch:
```console
$ module load omniperf
$ flux alloc -N1 flux run hostnametioga10
[detached: session exiting]
```

It should be simple to add a test for this, but thought I'd post it as is to get any comments about the general approach.

### DIFF
--- a/doc/man7/flux-environment.rst
+++ b/doc/man7/flux-environment.rst
@@ -474,6 +474,16 @@ of compiled-in install paths and the environment.
 
    Values may include multiple directories separated by colons.
 
+.. note::
+   Flux commands written in Python further modify Python's
+   `sys.path <https://docs.python.org/3/library/sys.html#sys.path>`_
+   to ensure that interpreter default paths appear before any custom values
+   set in :envvar:`PYTHONPATH`. This is an attempt to  avoid incompatible
+   modules interfering with the operation of Flux commands. If it becomes
+   necessary to force a non-standard module first in the search path (e.g.
+   for testing, instrumentation, etc.) then :envvar:`FLUX_PYTHONPATH_PREPEND`
+   should be used.
+
 .. envvar:: LUA_PATH
             LUA_CPATH
             FLUX_LUA_PATH_PREPEND
--- a/doc/man7/flux-environment.rst
+++ b/doc/man7/flux-environment.rst
@@ -421,6 +421,18 @@ SUB-COMMAND ENVIRONMENT
 :man1:`flux` sets up the environment for sub-commands using a combination
 of compiled-in install paths and the environment.
 
+.. note::
+   The PREPEND versions of environment variables below may
+   be necessary when developing and testing a new version
+   of a Flux command (:envvar:`FLUX_EXEC_PATH_PREPEND`),
+   module (:envvar:`FLUX_MODULE_PATH_PREPEND`), connector
+   (:envvar:`FLUX_CONNECTOR_PATH_PREPEND`), or Python module
+   (:envvar:`FLUX_PYTHONPATH_PREPEND`) when an existing version of that
+   component is already installed in the system default paths. Otherwise,
+   the installed component would always be used by the system Flux, since
+   the installed paths are always placed first in the subcommand environment
+   created by :man1:`flux`.
+
 .. envvar:: FLUX_EXEC_PATH
             FLUX_EXEC_PATH_PREPEND
 
--- a/doc/test/spell.en.pws
+++ b/doc/test/spell.en.pws
@@ -798,3 +798,4 @@ param
 perres
 pertask
 mpibind
+sys
--- a/src/cmd/Makefile.am
+++ b/src/cmd/Makefile.am
@@ -117,7 +117,8 @@ dist_fluxcmd_SCRIPTS = \
 	flux-cancel.py \
 	flux-watch.py \
 	flux-update.py \
-	flux-imp-exec-helper
+	flux-imp-exec-helper \
+	py-runner.py
 
 fluxcmd_PROGRAMS = \
 	flux-terminus \
--- a/src/cmd/flux.c
+++ b/src/cmd/flux.c
@@ -327,13 +327,19 @@ void exec_subcommand_py (bool vopt, const char *dir,
                             prefix ? prefix : "", argv[0]);
 
     if (access (path, R_OK|X_OK) == 0) {
-        char *av [argc+2];
+        const char *wrapper = flux_conf_builtin_get ("python_wrapper",
+                                                     FLUX_CONF_AUTO);
+        char *av [argc+3];
         av[0] = PYTHON_INTERPRETER;
-        av[1] = path;
-        for (int i = 2; i < argc+2; i++)
-            av[i] = argv[i-1];
+        av[1] = (char *) wrapper;
+        av[2] = path;
+        for (int i = 3; i < argc+3; i++)
+            av[i] = argv[i-2];
         if (vopt)
-            log_msg ("trying to exec %s %s", PYTHON_INTERPRETER, path);
+            log_msg ("trying to exec %s %s %s",
+                     PYTHON_INTERPRETER,
+                     wrapper,
+                     path);
         execvp (PYTHON_INTERPRETER, av);
     }
     free (path);
--- a/src/cmd/flux.c
+++ b/src/cmd/flux.c
@@ -189,7 +189,8 @@ int main (int argc, char *argv[])
     }
 
     environment_from_env (env, "FLUX_EXEC_PATH", "", ':');
-    environment_push (env, "FLUX_EXEC_PATH",
+    environment_push (env,
+                      "FLUX_EXEC_PATH",
                       flux_conf_builtin_get ("exec_path", flags));
     environment_push (env, "FLUX_EXEC_PATH", getenv ("FLUX_EXEC_PATH_PREPEND"));
 
@@ -197,14 +198,16 @@ int main (int argc, char *argv[])
     environment_push (env,
                       "FLUX_CONNECTOR_PATH",
                       flux_conf_builtin_get ("connector_path", flags));
-    environment_push (env, "FLUX_CONNECTOR_PATH",
+    environment_push (env,
+                      "FLUX_CONNECTOR_PATH",
                       getenv ("FLUX_CONNECTOR_PATH_PREPEND"));
 
     environment_from_env (env, "FLUX_MODULE_PATH", "", ':');
     environment_push (env,
                       "FLUX_MODULE_PATH",
                       flux_conf_builtin_get ("module_path", flags));
-    environment_push (env, "FLUX_MODULE_PATH",
+    environment_push (env,
+                      "FLUX_MODULE_PATH",
                       getenv ("FLUX_MODULE_PATH_PREPEND"));
 
     if (getenv ("FLUX_URI"))
@@ -319,9 +322,10 @@ void exec_subcommand_py (bool vopt, const char *dir,
 	                     const char *prefix)
 {
     char *path = xasprintf ("%s%s%s%s.py",
-            dir ? dir : "",
-            dir ? "/" : "",
-            prefix ? prefix : "", argv[0]);
+                            dir ? dir : "",
+                            dir ? "/" : "",
+                            prefix ? prefix : "", argv[0]);
+
     if (access (path, R_OK|X_OK) == 0) {
         char *av [argc+2];
         av[0] = PYTHON_INTERPRETER;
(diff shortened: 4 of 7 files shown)


grondo:
Pushed a slightly different solution here. Instead of trying to get the builtin `python_path`, just prepend the directory to the current module. This should work for either builddir or installed Flux use.

garlick:
Is any update to [flux-environment(7)](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man7/flux-environment.html#envvar-PYTHONPATH) needed?  E.g. would someone that sets PYTHONPATH or FLUX_PYTHONPATH_PREPEND purposefully to get Flux to work in some environment be confused by the outcome?

grondo:
> Is any update to [flux-environment(7)](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man7/flux-environment.html#envvar-PYTHONPATH) needed? E.g. would someone that sets PYTHONPATH or FLUX_PYTHONPATH_PREPEND purposefully to get Flux to work in some environment be confused by the outcome?

It would probably be a good idea to describe what this modification of `sys.path` is doing and why, otherwise it wouldn't really be documented anywhere (though it is only transitively related to the environment). Although we should first determine if the approach in this PR works and is acceptable practice.

I'm struggling to think of why a user would set PYTHONPATH and/or FLUX_PYTHONPATH_PREPEND to get Flux to work in some environment. If we want users to be able to substitute alternate module paths first into Flux Python utilities search paths, then we can't also simultaneously prevent the practice to avoid issues caused by it. :thinking: 

At the very least `fixsyspath` should probably preserve `FLUX_PYTHONPATH_PREPEND` in `sys.path` if it is set, though I wonder if there is any real need for that environment variable anymore.

grondo:
I pushed a commit to support `FLUX_PYTHONPATH_PREPEND` in this workaround (it would be squashed before WIP is removed though)

garlick:
Are those _PREPEND variables not used?  I've never used them.  Maybe they can just go away?

garlick:
Apologies - I think if we did choose to drop the _PREPEND variables, it would be in another PR anyway.  I didn't mean to get this discussion off track.

grondo:
> Are those _PREPEND variables not used? I've never used them. Maybe they can just go away?

The reason behind the `_PREPEND` variables was discussed in #574. They are required for framework projects to allow testing of built subcommands, broker modules, or Python modules which may already be installed in the standard paths assigned by `flux(1)`. This is because these subprojects use the installed `flux(1)` for testing, and otherwise sharness tests, etc. would pick up the installed components instead of the builddir components.

Fluxion, for example, has this in `sched-sharness.sh`:
```sh
  # Set up environment so that we find flux-sched modules,
  #  and commands from the build directories:
  FLUX_MODULE_PATH_PREPEND="${SHARNESS_BUILD_DIRECTORY}/resource/modules/"
  FLUX_MODULE_PATH_PREPEND="${SHARNESS_BUILD_DIRECTORY}/resource/modules/.libs":${FLUX_MODULE_PATH_PREPEND}
  FLUX_MODULE_PATH_PREPEND="${SHARNESS_BUILD_DIRECTORY}/qmanager/modules/":${FLUX_MODULE_PATH_PREPEND}
  FLUX_MODULE_PATH_PREPEND="${SHARNESS_BUILD_DIRECTORY}/qmanager/modules/.libs":${FLUX_MODULE_PATH_PREPEND}
  FLUX_EXEC_PATH_PREPEND="${SHARNESS_TEST_SRCDIR}/scripts":"${SHARNESS_TEST_SRCDIR}/../src/cmd"
  export PYTHONPATH="${SHARNESS_TEST_SRCDIR}/../src/python${PYTHONPATH:+:${PYTHONPATH}}"
```

Currently `FLUX_PYTHONPATH_PREPEND` is not used in the above, but it sure seems like it should be, o/w on a system with Fluxion already installed the Python modules loaded during testing will be the installed versions.

grondo:
It would probably be helpful (since even we forgot the purpose of the _PREPEND variables) to add a more specific description of that purpose to `flux-environment(7)` as you suggest. How about something like these additional notes in that man page (two separate commits)
```diff
diff --git a/doc/man7/flux-environment.rst b/doc/man7/flux-environment.rst
index d6c2fa692..419c83144 100644
--- a/doc/man7/flux-environment.rst
+++ b/doc/man7/flux-environment.rst
@@ -419,6 +419,18 @@ SUB-COMMAND ENVIRONMENT
 :man1:`flux` sets up the environment for sub-commands using a combination
 of compiled-in install paths and the environment.
 
+..note::
+   The PREPEND versions of environment variables below may
+   be necessary when developing and testing a new version
+   of a Flux command (:envvar:`FLUX_EXEC_PATH_PREPEND`),
+   module (:envvar:`FLUX_MODULE_PATH_PREPEND`), connector
+   (:envvar:`FLUX_CONNECTOR_PATH_PREPEND`), or Python module
+   (:envvar:`FLUX_PYTHONPATH_PREPEND`) when an existing version of that
+   component is already installed in the system default paths. Otherwise,
+   the installed component would always be used by the system Flux, since
+   the installed paths are always placed first in the environment created
+   by :man1:`flux`.
+
 .. envvar:: FLUX_EXEC_PATH
             FLUX_EXEC_PATH_PREPEND
 
@@ -460,6 +472,15 @@ of compiled-in install paths and the environment.
 
    Values may include multiple directories separated by colons.
 
+..note::
+   Flux commands written in Python further modify Python's ``sys.path``
+   to ensure that interpreter default paths appear before any custom values
+   set in :envvar:`PYTHONPATH`. This is an attempt to  avoid incompatible
+   modules interfering with the operation of Flux commands. If it becomes
+   necessary to force a non-standard module first in the search path (e.g.
+   for testing, instrumentation, etc.) then :envvar:`FLUX_PYTHONPATH_PREPEND`
+   should be used.
+
 .. envvar:: LUA_PATH
             LUA_CPATH
             FLUX_LUA_PATH_PREPEND

```

grondo:
Well a couple builders are failing with these changes, so that is something to look into before further discussion of how this approach is documented in `flux-environment(7)` I suppose.

garlick:
The flux-environment(7) update is excellent!

grondo:
@trws - what do you think of the approach proposed in this PR? The drawback it is requires the use of a module forced to be imported first and only for side effecs (which causes warnings from CodeQL of course). However, the nice thing is it does not require modification of the environment which would then be passed on to the user environment in submission commands.

@ryanday36: it would be useful to get your approval of this PR as well.

garlick (commented):
I've reviewed the flux-environment(7) updates and they look good to me.  Thanks!

trws:
I like the overall approach, especially that it can't accidentally propagate anything into the user's env,  and the additional `_PREPEND` vars are a great idea.  I will say that from a stylistic perspective I'd prefer either an explicit call to something in the module to do it, or perhaps better having the interpreter for all of these be a wrapper that applies the changes before running the file.  The former is just a style thing. The latter can have some further advantages, mainly:

* The fixes would be applied when the commands are run as command-line programs _and not when loaded as modules_, and
* We could use that wrapper to run python commands we integrate with _but do not control_ if we want to know it runs in our environment without modifying the imports.

It can be useful to have that split if we end up wanting to re-use parts of the commands or use commands from vendored packages for whatever reason.  We do [something like that](https://github.com/spack/spack/blob/develop/lib/spack/spack_installable/main.py) to allow different install types for spack. Admittedly the downside is you end up needing to use something like https://docs.python.org/pl/3.6/library/runpy.html to run the thing after the modifications.  What do you think?

grondo:
>  I will say that from a stylistic perspective I'd prefer either an explicit call to something in the module to do it, 

I don't think this approach will work. We need to subvert `sys.path` _before_ the first real `import` statement to fix the original issue, which is that a user modified `PYTHONPATH` puts some incompatible standard module in the path before the stdlib paths. (In the specific case here it was `urllib` that was broken)

>  perhaps better having the interpreter for all of these be a wrapper that applies the changes before running the file. 

I like this idea, especially since as you note it would for any subprocess Python subcommand. 

Note, however, this fix is not being applied to Python modules, but only Python commands/utilities under `src/cmd`. Therefore they will never be included as modules anywhere. These commands are run as `flux python /path/to/command.py`. Would the approach here be to add a new command runner and then invoke all Python commands as `flux python /path/to/python-runner.py /path/to/command.py` and then `python-runner.py` would fix `sys.path` then somehow include `command.py` and run it without using `exec()`?

It looks like perhaps `runpy.run_path()` will work for this so I'll give it a try.

trws:
On 13 Nov 2023, at 11:39, Mark Grondona wrote:

>>  I will say that from a stylistic perspective I'd prefer either an 
>> explicit call to something in the module to do it,
>
> I don't think this approach will work. We need to subvert `sys.path` 
> _before_ the first real `import` statement to fix the original issue, 
> which is that a user modified `PYTHONPATH` puts some incompatible 
> standard module in the path before the stdlib paths. (In the specific 
> case here it was `urllib` that was broken)

Yes, it wouldn't work if the call goes after the rest of the imports, 
but it would if the call is executed between this import and the rest, 
which would also prevent tools like isort from sorting the import across 
the call.

>>  perhaps better having the interpreter for all of these be a wrapper 
>> that applies the changes before running the file.
>
> I like this idea, especially since as you note it would for any 
> subprocess Python subcommand.
>
> Note, however, this fix is not being applied to Python modules, but 
> only Python commands/utilities under `src/cmd`. Therefore they will 
> never be included as modules anywhere. These commands are run as `flux 
> python /path/to/command.py`. Would the approach here be to add a new 
> command runner and then invoke all Python commands as `flux python 
> /path/to/python-runner.py /path/to/command.py` and then 
> `python-runner.py` would fix `sys.path` then somehow include 
> `command.py` and run it without using `exec()`?
>
> It looks like perhaps `runpy.run_path()` will work for this so I'll 
> give it a try.

Yeah that's pretty much exactly the idea.  If `runpy` works easily 
enough hopefully it would be straightforward.

Also you're quite right, they're normally not modules in the traditional 
sense, but I've seen sets of python commands that weren't intended to be 
modules end up sharing functions between each other over time, or the 
reverse (especially for testing).  Not necessarily something we'll ever 
do, but it's one less thing we'd have to worry about if we ever did. 
(matters less now, but issues from things like that used to be really, 
really hard to debug, bit of a post-trauma response perhaps :laugh:)

grondo:
Ok, I've pushed an experimental modification of this PR based on the suggestion by @trws. This is indeed much cleaner:
- A new wrapper script `py-runner.py` is introduced. All the code previously in `fixsyspath` to modify `sys.path` is moved local to this script, and the script invokes its argument in the same interpreter instance using [runpy.run_path()](https://docs.python.org/pl/3.6/library/runpy.html#runpy.run_path)
- The location of the wrapper script is encoded in the "builtin conf", and `flux(1)` now wraps all calls to Python commands in `py-runner.py`

This seems to work (though `runpy.run_path()` strangely always adds the path to the script to `sys.path`, but this doesn't seem to be a problem in practice). The only other weirdness is that now all Python commands will show up with `py-runner.py` in `ps` output, but again this doesn't seem to cause any other issues.

(Well, we'll see how this change does in CI)

grondo:
Well, maybe some more to do here, some CI tests failing immediately. Re-setting WIP on this one.

trws:
FWIW this looks really slick. I hadn't thought of putting it right into the subcommand exec but that's awesome.

grondo:
Ok, some tests were failing because `run_name="__main__"` was missing from `runpy.run_path()`. This seems to work for commands that use the `flux.util.CLIMain` decorated (though I have no idea why), but some of our commands do not use that wrapper (`flux-perilog-run.py` for example).

Now all tests seem to be passing again, so I'll remove WIP once more.

trws:
Oh that makes sense. I think we use the traditional check if the module was loaded as "__main__" in several places to check if it was run as a cli tool. I'll look this over again once I get to the AMD briefing and can take my laptop out but it looks pretty great to me so far.

grondo:
Thanks @trws! And thank you for the suggested change, that came out much cleaner than the previous approach.

trws (approved):

grondo:
It appears @trws approved and I just verified this still fixes the exact issue reported, so I'll set MWP.

PR #5555, opened 2023-11-12
doc: remove options from flux-alloc(1) et al that don't work

garlick:
Problem: the job submission man pages have had their own man pages since flux-mini(1) was deprecated, but not much effort was put into reviewing the lists of options in each page, so some don't apply.

Cull inappropriate options from these pages.

Note:  the first four commits are just gymnastics to make the later commits easier to review - they don't change anything in the generated pages.

These pages could use some more work to be sure - this is just a pass to fix the stated problem.

Edit: adding WIP since it seems like there's more to do!


garlick:
Fixed the issues that prompted me to add the WIP, so WIP is removed.

doc/Makefile.am
```diff
@@ -334,7 +334,6 @@ if ENABLE_DOCS
 man_MANS = $(MAN1_FILES) $(MAN3_FILES) $(MAN5_FILES) $(MAN7_FILES)
 $(RST_FILES): \
 	man1/common/nodeset.rst \
-	man1/common/job-parameters.rst \
```
grondo: Commit message typo: `int man pages` :arrow_right: `into man pages`

doc/man1/common/job-param-batch.rst
```diff
@@ -1,6 +1,7 @@
 .. option:: -n, --nslots=N
 
-   Set the number of slots requested. This parameter is required.
+   Set the number of slots requested. This parameter is required unless
+   :option:`--nodes` is specified.
```
grondo, line 4: > All the gack about tasks doesn't apply and the --exclusive option doesn't have a short -x option.

I'm wondering if it is a bug that `flux alloc` doesn't support the `-x` short option? It seems a bit strange that it is support with `flux run` and `flux submit`, but not `flux alloc` and `flux batch`.

doc/man1/flux-alloc.rst
```diff
@@ -84,7 +84,7 @@ give fine control over the requested environment submitted with the job.
 .. include:: common/job-environment.rst
 
 ENV RULES
-=========
```
grondo: Commit message typo: pacing

grondo (approved):
This looks good to me! Nice cleanup (should have been done when these commands were split from `flux-mini` I suppose... Sorry about that!).

garlick:
Thanks!  No need to apologize.  Git blame says I am the the sloppy person :-)

I fixed the commit message issues and I'll set MWP.

doc/man1/common/job-param-batch.rst
garlick: Sure, I'll open one, then we can close it if we think it's better left as is.

PR #5559, opened 2023-11-15
broker: eliminate some message copies

garlick:
Problem: the broker copies messages when it doesn't have to.

This PR eliminates some places in the broker where messages are needlessly copied.  I had hoped that by making the local paths between ingest -> kvs, ingest -> job manager, kvs -> content cache effectively zero copy, there would be a measurable effect on the job throughput test.  Well, not really.

I'll mark this WIP for now and revisit tomorrow to make sure I'm not missing something.

The other thing here is I replaced the strange half-constructed connector with back-to-back interthread handles.  It should make the broker code a little easier to understand.

### DIFF
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -78,7 +78,7 @@
 
 
 static int broker_event_sendmsg (broker_ctx_t *ctx, const flux_msg_t *msg);
-static int broker_response_sendmsg (broker_ctx_t *ctx, const flux_msg_t *msg);
+static int broker_response_sendmsg_new (broker_ctx_t *ctx, flux_msg_t **msg);
 static void broker_request_sendmsg (broker_ctx_t *ctx, const flux_msg_t *msg);
 static int broker_request_sendmsg_internal (broker_ctx_t *ctx,
                                             const flux_msg_t *msg);
@@ -1674,7 +1674,7 @@ static int overlay_recv_cb (flux_msg_t **msg, overlay_where_t where, void *arg)
             broker_request_sendmsg (ctx, *msg);
             break;
         case FLUX_MSGTYPE_RESPONSE:
-            if (broker_response_sendmsg (ctx, *msg) < 0)
+            if (broker_response_sendmsg_new (ctx, msg) < 0)
                 goto drop;
             break;
         case FLUX_MSGTYPE_EVENT:
@@ -1797,7 +1797,7 @@ static void module_cb (module_t *p, void *arg)
         goto done;
     switch (type) {
         case FLUX_MSGTYPE_RESPONSE:
-            (void)broker_response_sendmsg (ctx, msg);
+            (void)broker_response_sendmsg_new (ctx, &msg);
             break;
         case FLUX_MSGTYPE_REQUEST:
             count = flux_msg_route_count (msg);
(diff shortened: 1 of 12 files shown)


src/broker/broker.c
```diff
@@ -83,9 +83,7 @@ static void h_internal_watcher (flux_reactor_t *r,
                                 int revents,
                                 void *arg);
 
-static void overlay_recv_cb (const flux_msg_t *msg,
-                             overlay_where_t where,
-                             void *arg);
+static int overlay_recv_cb (flux_msg_t **msg, overlay_where_t where, void *arg);
```
grondo, line 86: Commit message typo: overlay_fecv_f

grondo (commented):
I started going through this before I noticed that it was still a WIP. Found a couple commit message typos that I'll leave here in case it's helpful.

grondo (approved):
BTW, I've taken a look at this one a couple times and it looks great! (only noticed the commit message typos so far). I was going to do some profiling again with this + #5585 just to see if anything pops out, but marking this approved even though it is still a WIP in case that helps move things along.

garlick:
Thanks!  I was a little disappointed by the test results, but I guess might as well get it moving.

grondo:
Here's a series of 10 throughput tests (each run in a freshly started instance) on this branch (avg 540 job/s)

```console
$  for i in {1..10}; do src/cmd/flux start -s4 ./src/test/throughput.py -n 2048; done | grep ^throughput
throughput:     515.5 job/s (script: 487.8 job/s)
throughput:     546.1 job/s (script: 515.4 job/s)
throughput:     537.4 job/s (script: 508.3 job/s)
throughput:     528.4 job/s (script: 500.0 job/s)
throughput:     529.0 job/s (script: 499.9 job/s)
throughput:     548.9 job/s (script: 517.1 job/s)
throughput:     539.0 job/s (script: 508.8 job/s)
throughput:     543.2 job/s (script: 513.0 job/s)
throughput:     577.2 job/s (script: 543.4 job/s)
throughput:     539.3 job/s (script: 509.9 job/s)
```

vs current master (avg 516 job/s)
```console
$  for i in {1..10}; do src/cmd/flux start -s4 ./src/test/throughput.py -n 2048; done | grep ^throughput
throughput:     472.0 job/s (script: 449.1 job/s)
throughput:     504.0 job/s (script: 477.8 job/s)
throughput:     525.8 job/s (script: 497.4 job/s)
throughput:     494.5 job/s (script: 469.1 job/s)
throughput:     519.7 job/s (script: 491.9 job/s)
throughput:     495.0 job/s (script: 469.5 job/s)
throughput:     531.8 job/s (script: 502.6 job/s)
throughput:     511.7 job/s (script: 485.2 job/s)
throughput:     554.3 job/s (script: 522.9 job/s)
throughput:     559.9 job/s (script: 526.9 job/s)
```

garlick:
Averaging the jobs/s (first number), I get 540 for this PR vs 516 for master which is like a 4% improvement.  :shrug: better than a poke in the eye I guess.

grondo:
> Averaging the jobs/s (first number), I get 540 for this PR vs 516 for master which is like a 4% improvement

Yeah, I think what we're hitting here is other things are slower for running jobs that moving messages (probably kvs). With profiling I did notice a drop in the contribution of the module_sendmsg/module_sendmsg_new so at least we're not heating up the universe as much as before.

garlick:
Dropping the WIP.  I think this is OK but I want to take one more pass through the code to make sure.

garlick:
OK, I didn't spot any problems so I'll set MWP.

PR #5560, opened 2023-11-15
build: add libmissing, a convenience library for replacements for missing functions

garlick:
Problem: our build system is not set up to take advantage of AC_REPLACE_FUNCS when porting to new systems.

In a slack conversation, @vsoch was working through glibc dependencies while trying a build on alpine linux.  Whether or not that succeeds, it may be handy to have a `libmissing` convenience library of conditionally compiled replacement functions, since autotools has nice support for that via AC_REPLACE_FUNCS.

`strlcpy()` is one of those functions that might be provided by a system library, so it's the first tenant.

I borrowed a bit from  [munge's](https://github.com/dun/munge) approach here.


grondo:
Might want to add libmissing to `extend-exclude` in `.typos.toml` if it is just going to vendored snippets of code.

src/common/libzmqutil/cert.c
```diff
@@ -29,7 +29,9 @@
 
 #include "src/common/libutil/errno_safe.h"
 #include "src/common/libutil/strstrip.h"
-#include "src/common/libutil/strlcpy.h"
+#ifndef HAVE_STRLCPY
+#include "src/common/libmissing/strlcpy.h"
+#endif
```
grondo, line 34: This doesn't follow the instructions in the added `src/common/libmissing/Makefile.am`:
```
# or if there is an existing system include just for the missing function
#   #ifdef HAVE_DOIT
#   #include <doit.h>
#   #else
#   #include "src/common/libmissing/doit.h"
#   #endif
```

Should this instead be:
```c
#ifdef HAVE_STRLCPY
#include <strlcpy.h>
#else
#include "src/common/libmissing/strlcpy.h
#endif
```

garlick:
Ah good idea, I just force pushed a tweak to that last commit to have it cover libmissing/*.

grondo (approved):
This seems fine to me. Just one question inline.

grondo:
>  I just force pushed a tweak to that last commit to have it cover `libmissing/*`

May want to change that to `libmissing/*.[ch]` (if that works) to catch typos in `Makefile.am`, but meh there's no typos in there now so unlikely to be an issue going forward - so optional IMO.

src/common/libzmqutil/cert.c
garlick: It's provided on bsd-derived systems in `string.h`, which is included earlier, so OK as is?

grondo: Ah, I had forgotten that. And your comment does say "an existing include just for the missing function" so this was a thinko on my part. Sorry!

garlick:
@dun - would you have a sec to take a quick peek at this?  I see on the development branch of munge that you've added a `private.h` include file that includes the other things.  I didn't do that here, opting to keep things simple for now.  But maybe you had some thoughts that I didn't intuit from my quick skim of what's happening in munge...

dun:
Hmmm... I don't remember `private.h`. Did I do that?

I did things a bit differently when I started updating [ConMan](https://github.com/dun/conman/tree/dev) recently. I followed the advice from [Russ Allbery](https://www.eyrie.org/~eagle/), placed those routines in a `portable` top-level directory, and set `AC_CONFIG_LIBOBJ_DIR([portable])` in [configure.ac](https://github.com/dun/conman/blob/d8849c4c9590fbaece4017685a6e962ade1d0106/configure.ac#L12).

I just now noticed I still have those function prototypes in `util-str.h` -- I should clean that up.

garlick:
Oh sorry I misspoke, I meant [private.h](https://github.com/dun/munge/blob/master/src/libmissing/missing.h).

dun:
Oh, `missing.h`. Yeah. I think I'll omit that with ConMan and just have a separate include file for each missing function that does the if-guard.

garlick:
Ah that would fix the most annoying thing about my approach which is needing to add the pre-processor conditionals to every source module that uses the missing function.

garlick:
Apparently I just can't type "private.h".  Oh wait, there goes that theory.  (Sorry for my sloppy ways!  Now I will edit history so nobody will know what this is about.)

grondo:
Given the issues described in #5563, we'll probably want to add an apline or other musl based build to CI before merging this PR.

grondo (approved):
LGTM!

garlick:
Thanks!  Setting MWP.

vsoch:
Should I still be tracking libmissing_argz for alpine debugging?

garlick:
@vsoch: stand by for for now if you don't mind.   Once #5560 is merged I'll rebase that one and push a few more fixes.

PR #5564, opened 2023-11-16
get sharness tests working on alpine linux

garlick:
This is a working branch in which various portability issues are worked through to get flux running on alpine linux.  Posting it early since others may be looking at it.

### DIFF
--- a/src/test/docker/alpine/Dockerfile
+++ b/src/test/docker/alpine/Dockerfile
@@ -16,6 +16,7 @@ RUN apk add \
     make \
     cmake \
     musl-dev \
+    musl-dbg \
     python3-dev \
     python3 \
     py3-cffi \
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -303,6 +303,7 @@ EXTRA_DIST= \
 	shell/output \
 	shell/initrc/tests \
 	flux-jobs/tests \
+	scripts/strerror_symbol \
 	scripts/run_timeout.py \
 	scripts/startctl.py \
 	scripts/groups.py \
--- a/t/scripts/strerror_symbol
+++ b/t/scripts/strerror_symbol
@@ -1,0 +1,33 @@
+#!/usr/bin/env python3
+##############################################################
+# Copyright 2023 Lawrence Livermore National Security, LLC
+# (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+#
+# This file is part of the Flux resource manager framework.
+# For details, see https://github.com/flux-framework.
+#
+# SPDX-License-Identifier: LGPL-3.0
+##############################################################
+
+"""
+Print the strerror string for the given errno constant name.
+
+For example:
+  grep "flux_future_get: $(strerror_symbol ENOTSUP)" file
+"""
+
+import errno
+import os
+import sys
+
+if len(sys.argv) != 2:
+    print("requires a single errno constant name or number and converts to strerror string")
+
+try:
+    print(os.strerror(getattr(errno, sys.argv[1])))
+    sys.exit(0)
+except AttributeError:
+    pass
+
+val = int(sys.argv[1])
+print(os.strerror(getattr(errno, errno.errorcode[val])))
--- a/t/t1004-kvs-namespace.t
+++ b/t/t1004-kvs-namespace.t
@@ -431,7 +431,7 @@ test_expect_success NO_CHAIN_LINT 'kvs: incomplete fence gets ENOTSUP when names
         wait_fencecount_nonzero 0 $NAMESPACETMP-REMOVE-FENCE0 &&
         flux kvs namespace remove $NAMESPACETMP-REMOVE-FENCE0 &&
         wait $watchpid &&
-        grep "flux_future_get: Operation not supported" fence_out
+        grep "flux_future_get: $(strerror_symbol ENOTSUP)" fence_out
 '
 
 
@@ -447,7 +447,7 @@ test_expect_success NO_CHAIN_LINT 'kvs: incomplete fence on rank 1 gets ENOTSUP 
         wait_fencecount_nonzero 1 $NAMESPACETMP-REMOVE-FENCE1 &&
         flux kvs namespace remove $NAMESPACETMP-REMOVE-FENCE1 &&
         wait $watchpid &&
-        grep "flux_future_get: Operation not supported" fence_out
+        grep "flux_future_get: $(strerror_symbol ENOTSUP)" fence_out
 '
 
 test_expect_success NO_CHAIN_LINT 'kvs: wait recognizes removed namespace' '
@@ -459,7 +459,7 @@ test_expect_success NO_CHAIN_LINT 'kvs: wait recognizes removed namespace' '
         wait_versionwaiters_nonzero $NAMESPACETMP-REMOVE-WAIT &&
         flux kvs namespace remove $NAMESPACETMP-REMOVE-WAIT &&
         ! wait $waitpid &&
-        grep "flux_kvs_wait_version: Operation not supported" wait_out
+        grep "flux_kvs_wait_version: $(strerror_symbol ENOTSUP)" wait_out
 '
 
 #
--- a/src/common/librouter/test/usock_epipe.c
+++ b/src/common/librouter/test/usock_epipe.c
@@ -177,7 +177,7 @@ static void test_send_and_exit (flux_t *h, int count)
     int i;
     char sockpath[PATH_MAX + 1];
     flux_msg_t *msg;
-    flux_msg_t *nmsg;
+    flux_msg_t *nmsg = NULL;
     int fd;
     struct usock_client *client;
 
--- a/src/modules/job-exec/testexec.c
+++ b/src/modules/job-exec/testexec.c
@@ -481,6 +481,8 @@ static int testexec_config (flux_t *h, int argc, char **argv)
 static void testexec_unload (void)
 {
     testexec_ctx_destroy (testexec_ctx);
+    // Ensure a reload will not re-use a freed value
+    testexec_ctx = NULL;
 }
 
 struct exec_implementation testexec = {
--- a/src/test/generate-matrix.py
+++ b/src/test/generate-matrix.py
@@ -271,9 +271,8 @@ matrix.add_build(
 )
 
 matrix.add_build(
-    name="alpine - quick check",
+    name="alpine",
     image="alpine",
-    command_args="--quick-check",
     args=(
         "--prefix=/usr"
         " --sysconfdir=/etc"
--- a/t/lua/t0002-rpc.t
+++ b/t/lua/t0002-rpc.t
@@ -51,6 +51,6 @@ note ("ping to non-existent rank")
 local packet = { seq = "1", pad = "xxxxxx" }
 local msg, err = f:rpc ("broker.ping", packet, 99)
 is (msg, nil, "rpc: nil return indicates error")
-is (err, "No route to host", "rpc: err is 'No route to host'")
+ok (err == "No route to host" or err == "Host is unreachable", "rpc: err is 'No route to host' or 'Host is unreachable'")
 
 done_testing ()
--- a/t/scripts/runpty.py
+++ b/t/scripts/runpty.py
@@ -273,7 +273,12 @@ def main():
         if args.input and args.input == "none":
 
             def write_eof(fd):
-                os.write(fd, bytes([termios.CEOF]))
+                if hasattr(termios, "CEOF"):
+                    value = bytes([termios.CEOF])
+                else:
+                    # No CEOF in termios, assume Ctrl-D/EOT (0x4)
+                    value = bytes([4])
+                os.write(fd, value)
 
             #  Sometimes the shell (if that is the target of runpty)
             #   does not read EOF if it is sent too soon. Therefore send
--- a/t/sharness.d/flux-sharness.sh
+++ b/t/sharness.d/flux-sharness.sh
@@ -3,6 +3,9 @@
 #  project-local sharness code for Flux
 #
 
+# add scripts directory to path
+export PATH="${SHARNESS_TEST_SRCDIR}/scripts:$PATH"
+
 #
 #  Extra functions for Flux testsuite
 #
--- a/t/t0001-basic.t
+++ b/t/t0001-basic.t
@@ -19,6 +19,7 @@ export MAN_DISABLE_SECCOMP=1
 
 RPC=${FLUX_BUILD_DIR}/t/request/rpc
 startctl=${SHARNESS_TEST_SRCDIR}/scripts/startctl.py
+path_printenv=$(which printenv)
 
 test_expect_success 'TEST_NAME is set' '
 	test -n "$TEST_NAME"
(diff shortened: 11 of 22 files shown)


trws:
Awesome, thanks for this @garlick and @vsoch. I picked at this for a bit years ago and stalled on argz (hadn't found gnulib, and the glibc version is kinda intertwined) really cool to see it this close!

trws:
I missed one we don't check in configure but apparently need for build, py3-ply:

```sh
apk add \
   gcc autoconf automake libtool make musl-dev python3-dev python3 py3-cffi py3-yaml \
   py3-jsonschema lua5.1-dev lua5.1-posix czmq-dev jansson-dev hwloc-dev lz4-dev \
   sqlite-dev ncurses-dev libarchive-dev git py3-ply
```

vsoch:
>  helpfully posted the command in slack to install flux prereqs in the alpine docker container:

For my case I'm using spack because we need the isolated view to copy over, but this approach works too (and likely will be faster, better suited for a base container build, and I can test after).

grondo:
@trws, are you already working on an alpine builder for CI?

trws:
I have a partial dockerfile, not tying it in yet but it should be straightforward.  So far the alpine container is ~1/3 the size of our bookworm one, which is kinda nuts.

grondo:
Ok, I'll let you handle that. I got as far as noticing that there's no munge package for alpine so that will have to be built by hand. Also noticed that libsodium-dev and jq (and optionally valgrind) are missing.

trws:
Right, I added valgrind to the docker file, but haven't looked at the others.  Thanks for that.

grondo:
And just as a warning, likely flux-security will need to some porting to libmusl as well...

garlick:
@trws I didn't know you were going to push to my branch and I just rebased to fix a problem with one of the patches. 

I'll sort it out but in the future let's try to coordinate that.

garlick:
OK to force push that update? (I pulled in your patches to my rebased branch @trws) or is there work in progress?

garlick:
@vsoch - just force pushed here after rebasing on current master in case you want to update your copy?

vsoch:
Updated, and here is the current failing check:

image

```
FAIL: lua/t0002-rpc.t 18 - rpc: err is 'No route to host'
ERROR: lua/t0002-rpc.t - exited with status 1
```

garlick:
I can't run all the sharness tests.  They eventually stop with an I/O error from the harness.   Have you seen that in your spack environment?

vsoch:
> I can't run all the sharness tests. They eventually stop with an I/O error from the harness. Have you seen that in your spack environment?

What's the command to run the sharness tests? I've been doing `make check`

garlick:
`make check`  runs all the sharness tests, or you can run `make check` from the `t` directory to run _only_ the sharness tests.

grondo:
The Lua error anyway looks like another case of `strerror()` reporting a different string for `EHOSTUNREACH` on musl.

Is the tap-driver.sh being affected by the miniature default pipe size on this platform?

garlick:
Interestingly the mini pipe size was only observed by @vsoch in a spack environment.  Not sure what's up with _that_ :open_mouth: 

I'll add a fix for the lua problem - it's of a piece with some other fixes I made.

garlick:
just pushed a fixup for that lua test

vsoch:
> Interestingly the mini pipe size was only observed by @vsoch in a spack environment. Not sure what's up with that 😮

it's probably good we are dual-testing in spack and external because we would undoubtably hit this issue at some point!

For tests, everything starts failing here:

image

and starts working at the lua ones again:

image


and we make it to the summary:

image

garlick:
Ah my guess is you don't have `jq` in your image.  It's required for all those tests.

vsoch:
yes that was it! Now we are at:

image

vsoch:
okay I think we are frozen here:

```
PASS: t2274-manager-perilog.t 24 - perilog: offline ranks are skipped with --exec-per-rank
PASS: t2274-manager-perilog.t 25 - perilog: offline ranks are drained with --drain-offline

```
No error yet, but it's been there a long time.

trws:
I got a warning converted to error in building one of the tests, probably just a newer gcc version or something, that I can push up.  Next thing is a couple of the restore tests are failing because they're printing sqlite errors that the database is not writable, but not the expected "no write permission" error, looking into it.

trws:
Turns out access returns true for W_OK even if the write bit is not set if the calling program is running as root.  One shouldn't do that in general anyway, so I'm not sure it needs to be fixed, but it surprised me regardless.

trws:
> okay I think we are frozen here:
> 
> ```
> PASS: t2274-manager-perilog.t 24 - perilog: offline ranks are skipped with --exec-per-rank
> PASS: t2274-manager-perilog.t 25 - perilog: offline ranks are drained with --drain-offline
> ```
> 
> No error yet, but it's been there a long time.

This one fails if gnu ps (or similar) isn't installed.  `apk add procps` seems to solve it for me.

garlick:
My plan is to rebase this on master once #5565 and #5568 are merged.  Then we'll have a limited alpine CI checking this branch.

So keep those fixes in your back pocket for a bit - after the rebase I'll turn the branch over to you for a while :-)

grondo:
>  apk add procps seems to solve it for me.

FWIW, this is in the Dockerfile in #5565 and the image pushed to `fluxrm/testenv:alpine`

trws:
A lot of other fails are from us looking for exact strerror strings.  I'm thinking I'll throw together a little perl util that will take the name of the constant (ENOTSUP say) and pop out the strerror for it so we can do this portably.  What do you think?

trws:
Wait, I just said perl didn't I... why am I doing that to myself, checking if that works with python, pretty sure it will.

garlick:
I already fixed some of those.  I'd be inclined not to over-engineer this unless there are many more and it will save time.

trws:
The main thing is I've found something like 10 so far, and if we end up supporting more libcs there will be even more variants.  Seemed like a way to have it be more self-documenting and not have to deal with it again.  Let me see how complicated it is, if it takes more than a few minutes I'll just fix them in place and move on.

garlick:
OK.  We do have a place for such things in `t/scripts` if you decide to do it.

trws:
Yup, ended up with something that lets us do this:

```sh
grep "$(strerror_symbol ENOTSUP)" waitcreate5.out
# or if we have a command returning an errno
grep "$(strerror_symbol $(exits_with_errno_value))" log_with_string.out
```

grondo:
I used to use the [perror(1)](https://man7.org/linux/man-pages/man1/perror.1.html) utility for this but strangely it is part of MySQL and not a core utility (would be useful though!).

garlick:
`moreutils` on debian has `errno`...

trws:
That is also available on alpine if it's preferable, hadn't thought about it.

grondo:
Oh, that's nice!

trws:
Ah found it. Musl's dlclose doesn't actually remove the module, though it does run destructors and whatever IIRC, so on reload the BSS isn't re-zeroed.  We were counting on that to avoid re-using a freed context.

garlick:
Oof, so we can't depend on zero-initialized bss in any dynamically loaded modules?  Argh.

trws:
Yeah unfortunately.  We can count on it _once_, so if it matters every time the module is loaded we'll need to either clear things or ensure they get set.    I think anything that's actually dynamically initialized, like anything with `__attribute__((constructor))` or similar, actually does get run on each load, but that doesn't help a whole lot.  The nice thing is that most of our modules are really good about this, we don't tend to hold onto things in global/static vars that would bite us like this, at least not that I can recall or that I see with a quick spot check.  As a result we're actually initializing things nicely every load in most of them regardless.

trws:
Alright, think we got it all!  @garlick you mentioned wanting to get a rebase or similar in, let me know when I can update and push these last fixes. 😄 

```
============================================================================
Testsuite summary for flux-core 0.53.0-644-g6ceb4d1e5
============================================================================
# TOTAL: 5759
# PASS:  5681
# SKIP:  74
# XFAIL: 4
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
```

vsoch:
I can also test again!

garlick:
Hi - sorry, I have to head out and we didn't get the CI PR merged.  I'll rebase on it anyway and push what I have here and then you can push your stuff @trws.  Here comes...

grondo:
Well if all the tests are passing now I guess we can get this in and CI doesn't need the "quick check" option.

vsoch:
> Well if all the tests are passing now I guess we can get this in and CI doesn't need the "quick check" option.

I still have a few failing in my spack container (will report when they finish) but maybe I didn't make some needed change to the environment?

vsoch:
I'm going to make sure I have the same install deps in my environment, and I'll run again.

garlick:
we still don't have the remaining fixes from @trws so some tests are expected to fail the way things are at the moment.

trws:
Yup, I just finished the rebase and am tweaking the dockerfile and re-running tests before I push everything up.  Will push x86_64 and arm64 containers for testenv too.

trws:
Ok, tests passed locally, built up commits for all the fixes, built and pushed the slightly updated dockerfile with openssh for the connector tests on both architectures.  Here's hoping everything passes here too. 🤞

trws:
Initially hit an issue with sharness complaining about the broken `&&` chain, so I refactored a couple of the tests to avoid the "or".  Also I had something that should have been exported just set, should pass this time, possibly excepting inception (that one reported all tests as passing but then returned error... no clue).

grondo:
There is also a NO_CHAIN_LINT prereq if the test doesn't play well with chain-lint (or sometimes abstracting the test to a shell function is preferable)

t/scripts/strerror_symbol
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python3
```
garlick, line 1: needs a copyright header

t/t1004-kvs-namespace.t
```diff
@@ -431,7 +431,7 @@ test_expect_success NO_CHAIN_LINT 'kvs: incomplete fence gets ENOTSUP when names
         wait_fencecount_nonzero 0 $NAMESPACETMP-REMOVE-FENCE0 &&
         flux kvs namespace remove $NAMESPACETMP-REMOVE-FENCE0 &&
         wait $watchpid &&
-        grep "flux_future_get: Operation not supported" fence_out
+        grep "flux_future_get: $(strerror_symbol ENOTSUP)" fence_out
```
garlick, line 434: That turned out pretty nice!

garlick (commented):
Nice work running those down @trws!

LMK if you'd like me to rebase this on current master and squash the fixup to the earlier EHOSTUNREACH fix of mine.

There are a couple of commit messages that would benefit from the Problem: format we've been using for commit messages.  I'm happy to reword those too.  LMK.

src/test/docker/alpine/Dockerfile
```diff
@@ -38,7 +39,8 @@ RUN apk add \
     jq \
     aspell \
     aspell-en \
-    valgrind
+    valgrind \
+    openssh
```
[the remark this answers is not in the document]
grondo: FWIW, all tests on this branch pass in my current image which doesn't have ssh installed. It would be good to know what exactly was failing as maybe a different kind of test fix is needed.

grondo:
We can now drop the `--quick-check` option from the alpine build in CI here as well.

src/test/docker/alpine/Dockerfile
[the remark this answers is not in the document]
trws: Oh that's interesting. Maybe it was a symptom of one of the other issues, I'll see if I can pull it back up.

trws:
> Nice work running those down @trws!
> 
> 
> 
> LMK if you'd like me to rebase this on current master and squash the fixup to the earlier EHOSTUNREACH fix of mine.
> 
> 
> 
> There are a couple of commit messages that would benefit from the Problem: format we've been using for commit messages.  I'm happy to reword those too.  LMK.

That would be awesome if you would, I'm solo with Kim while my wife is graduating in New Hampshire this weekend, so I'm not going to have time until late today.

vsoch:
> my wife is graduating in New Hampshire this weekend, so I'm not going to have time until late today.

Congrats!

garlick:
OK I'm on it.

garlick:
Ok just pushed cleaned up PR
- squashed the error string commits together
- dropped openssh from the dockerfile
- reworded some of Tom's commits
- dropped a silly cleanup commit of mine that had nothing to do with this
- added a commit that drops --quick-check from the alpine builder

grondo:
Hm, I'm still seeing openssh in the Dockerfile (not that it is harming anything being in there...)

Also, does `strerror_symbol` still need a copyright header?

garlick:
Oops!  Fixed those things.  Thanks @grondo!

garlick:
BTW, I am not sure what changed after @trws's fixes, but I can now run all the tests without getting an I/O error from the test harness.  Maybe somehow that was triggered by the quantity of failing tests?

grondo (approved):
This LGTM! (Though I realize it is still WIP)

vsoch:
I'm still getting fails in spack, but maybe that's a road we shouldn't go down yet. Likely I'll wait until the PR is done here (and merged) and then start my build fresh using it, and go from there.

garlick:
I went ahead and dropped the WIP since we seem to miraculously have arrived somewhere good, thanks to @trws pushing us over the finish line.  Honestly, I didn't think we'd get here that quckly!

We'll see if CI passed :-)

garlick:
That sounds like a good plan @vsoch.  I'm not sure why spack is getting different results with alpine.  It was weird that the pipe size was smaller in your environment compared to alpine docker too.  Hmm.

vsoch:
>  It was weird that the pipe size was smaller in your environment compared to alpine docker too. Hmm.

I think Mario ate too many bytes in this container, I mean, there is no alternative!

trws:
My current theory is the kernel or settings are a bit different for the pipe issue. What base system are you running on @vsoch? Otherwise the spack built stuff wouldn’t have any downstream patches alpine packagers may be using to account for musl compatibility issues. That’s part of why I used alpine packages instead, never know how many things build but don’t actually work at runtime building from source against a new libc.

vsoch:
> What base system are you running on [@vsoch](https://github.com/vsoch)?

My machine is ubuntu 22.04, and the spack image base is alpine:3.18

trws:
Apparently I forgot to add strerror_symbol to the dist list.  This should take care of the last issue with that.  I'm not sure why the lsf test that uses tssh failed though, and can't seem to get it to fail locally.

trws:
The inception fail is t5518 (I think) which seems to be the one that marks an 8k buffer size. Any chance this is related to what @vsoch was seeing? If we need a bigger pipe, or an intermediate buffer, that should be possible if we need a workaround.

garlick:
> The inception fail is t5518 (I think) which seems to be the one that marks an 8k buffer size

That doesn't seem related to this PR.  I opened #5572 on that one.

garlick:
I'll go ahead and rebase on master now that we've got that test removed.

grondo:
I can look into the lsf resolver test failure if nobody else is doing so already. (Won't fail for me locally).

grondo:
Ah, I think process listing order on alpine's version of `ps` is not guaranteed, so sometimes the lsf resolver is getting the rank 1 broker's pid instead of rank 0, and thus the URI does not match `FLUX_URI` (`local-1` vs `local-0-). Sorting the process listing in the resolver should help here, but if PIDs happened to be assigned randomly on a system then that won't work either.

Probably the easiest thing to do is allow the result to match either rank 0 or rank 1 local-uri. Shall I push a commit here to do that?

grondo:
Ok, I've pushed a fix for the lsf resolver test here (and rebased while I was at it)

grondo:
Hm, the bookworm - distcheck build failed with the following in `t2233-job-info-update.t`
```
  expecting success: 
  	update1=$(expiration_add $jobid 100) &&
  	flux job eventlog $jobid &&
  	${UPDATE_LOOKUP} $jobid R > lookup3A.out &&
  	update2=$(expiration_add $jobid 200) &&
  	${UPDATE_LOOKUP} $jobid R > lookup3B.out &&
  	flux cancel $jobid &&
  	cat lookup3A.out | jq -e ".execution.expiration == ${update1}" &&
  	cat lookup3B.out | jq -e ".execution.expiration == ${update2}"
  
  1700500538.396233 submit userid=1001 urgency=16 flags=0 version=1
  1700500538.408112 validate
  1700500538.419010 depend
  1700500538.419053 priority priority=16
  1700500538.421906 alloc annotations={"sched":{"resource_summary":"rank0/core0"}}
  1700500538.425365 start
  1700500538.507067 resource-update expiration=1700500638
  true
  false
  not ok 3 - job-info: update lookup with update events works (job active)
```

grondo:
Well, that can't be related to the changes here so I just restarted that builder.

t/Makefile.am
```diff
@@ -303,6 +303,7 @@ EXTRA_DIST= \
 	shell/output \
 	shell/initrc/tests \
 	flux-jobs/tests \
+	scripts/strerror_symbol \
```
grondo, line 306: Should this commit just be squashed with the one that adds `strerror_symbol`?

grondo (approved):
This LGTM and is passing all tests. Someone should review the commit I added here though.

garlick (commented):
Your commit looks good to me @grondo.

t/Makefile.am
garlick: Yes, just did that and force-pushed.

garlick: Shall I set MWP on this?

grondo: Sure!

trws:
Awesome! Now that we're through here, I'll look at flux-sched next.

PR #5565, opened 2023-11-16
ci: add alpine Dockerfile and CI build

grondo:
Pushing this here as a place to coordinate on a fully working Dockerfile for alpine to eventually include in CI.

This version works (with some changes) with flux-security. I've pushed the result to `fluxrm/testenv:alpine` so that a PR for flux-security can be developed (and tested), but feel free to amend/change/update this PR as needed for flux-core.

### DIFF
--- a/src/test/checks_run.sh
+++ b/src/test/checks_run.sh
@@ -16,6 +16,7 @@
 #  DISTCHECK     Run `make distcheck` if set
 #  RECHECK       Run `make recheck` if `make check` fails the first time
 #  UNIT_TEST_ONLY Only run `make check` under ./src
+#  QUICK_CHECK   Run only `make check-prep` and a simple test
 #  PRELOAD       Set as LD_PRELOAD for make and tests
 #  POISON        Install poison libflux and flux(1) in image
 #  INCEPTION     Run tests under a flux instance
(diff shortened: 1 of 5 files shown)


grondo:
Force pushed a fix here to create the munge user in the alpine image.

grondo:
Ok, this PR now runs a "quick check" in CI on the alpine image (and is therefore based on top of #5568).
Also had to add py3-sphinx to the image to get docs building (that increased the size by at least 50MB)

garlick (approved):
LGTM!

grondo:
Ok, Thanks! rebased on master -- will set MWP on the assumption CI will pass.

grondo:
Hm, the inception build is failing here as well. I wonder if we somehow made some test more unreliable?

garlick:
I can't see anything that went in recently that would do that...

This time looks like a lua test is failing?  (At least that is the first failure I found scanning the raw log)

grondo:
I think the Lua test is an expected failure? The test I see failing is one of the issue reproducers:
```
not ok 21 - t5518-job-validator-hang.sh
```

grondo:
Oh, also the same one as before:
```
not ok 15 __main__.TestJobOutput.test_output_watch_lines
```

grondo:
Well, I threw a fix here for the Python test error. Not sure about the job-validator hang test. It is racy because it depends on the speed at which internal buffers fill and empty. Maybe hat test needs to be removed.

PR #5567, opened 2023-11-17
fix improper include directives in source files impacting portability

garlick:
Problem: a few non-standard headers are included in place of the standard ones, impacting portability.  Other headers are included for no reason, likely left over from cut and paste or previous versions of the code.  Cleanup is in order.

Use the standard headers and drop some needless includes.

This was peeled off of #5564.

### DIFF
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -13,6 +13,7 @@
 #endif
 #include <assert.h>
 #include <libgen.h>
+#include <signal.h>
 #include <locale.h>
 #include <inttypes.h>
 #include <sys/prctl.h>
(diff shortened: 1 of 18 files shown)


grondo (approved):
LGTM!

Only comment is that the PR title could be more specific since "clean up sloppy include directives" isn't very specific, though I don't have a great idea of how to improve it. Perhaps "fix improper include directives in source files impacting portability" or similar?

I don't feel strongly about this, though, so if you feel the current title is correct, that's fine.

garlick:
I took your suggestion on the PR title (good improvement!)

Also, rebased on current master and forced a push.  I'll set MWP.

PR #5568, opened 2023-11-17
port to Alpine linux

garlick:
Problem: flux-core does not build on Alpine linux.

This addresses several issues that came up when trying to build on Alpine, primarily where we have used non-standard interfaces of glibc.  This PR is sufficient to build flux-core and run its unit tests.  A number of sharness tests are still failing.  Those are being worked in #5564.

I verified that we can run `make check-prep` in  Alpine with this PR applied.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -205,7 +205,8 @@ AC_CHECK_FUNCS( \
 )
 # See src/common/libmissing/Makefile.am
 AC_REPLACE_FUNCS( \
-  strlcpy
+  strlcpy \
+  argz_add \
 )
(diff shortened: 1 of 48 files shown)


grondo (approved):
This LGTM. Nice work!

grondo:
I'll restart the inception builder, it failed here:
```
not ok 15 __main__.TestJobOutput.test_output_watch_lines
```
Probably something racy in that test and unrelated to this PR.

garlick:
Thanks!  Group effort!  I'll set MWP.

PR #5569, opened 2023-11-17
libflux: fix performance of kvs-watch cancel/disconnect

grondo:
While profiling a throughput test workload, I noticed that at least 7% of the broker's CPU time was spent in kvs-watch `watcher_cancel_all()`.

The problem occurs because pulling data from the disconnect or cancel message is repeated for every member of the target list of the match. When the list is very large, this can add up to quite a bit of wasted time.

Add a "msg_match" structure to store the  data being matched so that the message only needs to processed once, and use this in kvs-watch.

Use this scheme in `flux_msglist_cancel()` and `flux_msglist_disconnect()` so that users of these functions can take advantage of the improvement as well.

### DIFF
--- a/src/common/libflux/disconnect.c
+++ b/src/common/libflux/disconnect.c
@@ -16,6 +16,25 @@
 #endif
 #include <flux/core.h>
 
+#include "ccan/str/str.h"
+
+int flux_msg_match_init (struct flux_msg_match *match,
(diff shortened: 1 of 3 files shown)


grondo:
Closing. Going a different way here that doesn't touch libflux

PR #5573, opened 2023-11-19
doc: improve RFC references in man pages

garlick:
Problem: the RFC references in the RESOURCES section of man pages include full URLs for the benefit of text-only man(1) users.  This not only clutters up the RESOURCES section with repetitive URLs, but would be a pain to change if RFC titles or URLs change.

This PR
- creates a `common/resources.rst` file in each man section for common resource links (start with the `flux-framework.org` one)
- adds the Flux RFC readthedocs top level URL to all RESOURCES sections
- where RFCs are referenced, use terse intersphinx references instead of spelling out full title and URL

Example: [before](https://flux-framework.readthedocs.io/projects/flux-core/en/stable/man5/flux-config-queues.html) and [after](https://flux-framework--5573.org.readthedocs.build/projects/flux-core/en/5573/man5/flux-config-queues.html).


chu11 (approved):
LGTM!

garlick:
@chu11 returns!  Thanks for the review!  I'll set MWP.

garlick:
Thanks! Setting MWP.

PR #5574, opened 2023-11-19
man3: add Link with... instruction to SYNOPSIS

garlick:
Problem: most section 3 pages do not include link instructions.

Follow the pattern of the math functions (e.g. floor(3)) and add a "Link with -lflux-core" or "Linux with -lflux-idset" note at the end of SYNOPSIS.


grondo (approved):
LGTM!

PR #5576, opened 2023-11-20
testsuite: fix fancy f grep inconsistency

chu11:
Problem: Several tests in t2292-job-update-running.t were failing due to "fancy f" vs "normal f" grepping inconsistencies.

Solution: Set FLUX_F58_FORCE_ASCII to true to get consistent testing.

### DIFF
--- a/t/t2292-job-update-running.t
+++ b/t/t2292-job-update-running.t
@@ -43,6 +43,10 @@ chmod +x job_manager_get_R
 
 export PATH=$(pwd):$PATH
 
+# some tests grep for job id via flux proxy.  To ensure job id is consistent
+# across all tests and instances, set FLUX_F58_FORCE_ASCII to true.
+export FLUX_F58_FORCE_ASCII=1
+
 job_manager_get_expiration() {
 	job_manager_get_R $1 | jq .R.execution.expiration
 }


garlick:
`sharness.d/flux-sharness.sh` unsets FLUX_F58_FORCE_ASCII so isn't there already a consistent environment?

Which test was failing and in what environment?

chu11:
On corona it was the last three tests in the file, like

```
test_expect_success 'instance expiration can be updated to unlimited' '
        test_debug "echo updating instance duration to inf" &&
        flux update $id duration=inf &&
        test_debug "echo waiting for second job resource-update event" &&
        flux proxy $id flux job wait-event -c 2 -vt 20 \
                $id1 resource-update &&
        exp=$(subinstance_get_job_expiration $id $id1) &&
        test_debug "echo job1 expiration is now $exp" &&
        echo $exp | jq -e ". == 0" &&
        flux proxy $id $SHARNESS_TEST_SRCDIR/scripts/dmesg-grep.py \
                -vt 30 \"job-exec.*updated expiration of $id1 to 0\.0\"
'
```

on the last line `$id1` has fancy f in it, but the grep output does not.

Perhaps b/c .... 

```
        id1=$(flux proxy $id flux submit sleep 300) &&
        id2=$(flux proxy $id flux submit -t4m sleep 300) &&
```

we're proxying to get `$id1` and the proxy has a different environment, leading to fancy f in `$id1`?

(corona has some NFS issue right now ... but was just able to reproduce on quartz)

grondo:
> we're proxying to get $id1 and the proxy has a different environment, leading to fancy f in $id1?

Hm, but `FLUX_URI_RESOLVE_LOCAL=t` is set for the test so I would have assumed same environment. It would be interesting to get to the bottom of this, actually. I'm a bit worried the system Flux is leaking into the test somehow since it is compiled with force-ascii enabled.

(and great catch btw, I think I saw this failing on corona too but didn't have time to look into it)

chu11:
> It would be interesting to get to the bottom of this, actually. I'm a bit worried the system Flux is leaking into the test somehow since it is compiled with force-ascii enabled.

I eventually was forced to printf inside `libutil` because nothing else was giving me clue.  Eventually learned the

`if (MB_CUR_MAX > 1 && streq (nl_langinfo (CODESET), "UTF-8")`

check fails in the "test instance" but not the subinstance.  I have absolutely no idea why the environments are like this.  No major environment variable differences between the two outside of PATH, MODULEPATH, and some similar things like that.  Don't know if it's a modules thing or some other kooky thing related to how LC sets up default environments (or something crazy I did that I'm not aware of).

chu11:
So I have yet to figure out the exact login/profile/rc file runes that are leading to the situation described above and affecting tests.  However, I think setting F58_FORCE_ASCII appears to be the right choice, as it'll lead to consistent output of `$id` and `$id1`, thus leading to correct grepping.

grondo:
> Note, I do not have LC_ALL set in my shell environment, so I'm not sure where this is being picked up for the test. Something magical in the LC environment and all their profile setups?

Among the first things `sharness.sh` does is set `LC_ALL=C` and `LANG=C`. The idea I think is to avoid having your login environment influence how the tests work.

> However, I think setting F58_FORCE_ASCII appears to be the right choice, as it'll lead to consistent output of $id and $id1, 

That sounds reasonable, but the test seems to have uncovered some unexpected behavior that we should look into (open an issue?) instead of just working around it.

chu11:
> Among the first things sharness.sh does is set LC_ALL=C and LANG=C. The idea I think is to avoid having your login environment influence how the tests work.

Ahhh, I wasn't aware.  Perhaps my environment is somehow scrubbing LC_ALL from the environment?  That can explain why it's not in my normal environment and the proxy environment.

> That sounds reasonable, but the test seems to have uncovered some unexpected behavior that we should look into (open an issue?) instead of just working around it.

Yeah, I think the gist is that login shells have the ability to twerk the environment ... and thus can lead to various unexpected results.

grondo:
> Yeah, I think the gist is that login shells have the ability to twerk the environment ... and thus can lead to various unexpected results.

Maybe I can look into this. Where do you think a login shell getting involved? (I can't seem to reproduce this issue)

chu11:
> Maybe I can look into this. Where do you think a login shell getting involved? (I can't seem to reproduce this issue)

It's only a guess that `flux proxy` is somehow invoking a login shell, b/c I can't explain this otherwise :-(  I know that it's a `$SHELL -c` internally, so it feels like it shouldn't.

grondo:
> on the last line $id1 has fancy f in it, but the grep output does not.
>
> we're proxying to get `$id1` and the proxy has a different environment, leading to fancy f in $id1?


Just to verify, if you have a flux-alloc instance running, and `LC_ALL=C` and `LANG=C` are set in your current environment such that we're presented with ascii `f`:

```console
$ flux jobs
   fBv177WFq grondo   flux        R      1      1   8.117m corona82
```

When you proxy to submit a job, you get a fancy `f`? (also ensure `FLUX_URI_RESOLVE_LOCAL=t` is set in your environment and `FLUX_F58_FORCE_ASCII` is unset to mimic the test environment):

```console
$ flux proxy fBv177WFq flux submit hostname
f4mV1CGrK
```

My version preserves the ascii only output and with `FLUX_URI_RESOLVE_LOCAL=t` I can't think of where/how flux-proxy would invoke a login shell, shouldn't flux-proxy just be invoking `flux-relay` with `FLUX_URI` set appropriately?

chu11:
now I'm super confused, b/c running this by hand (vs debugging within the tests) is producing different results.  Perhaps its something environment wise + the unique sharness setup.

So the only thing I know so far with fact is:

`id1=$(flux proxy $id flux submit sleep 300) &&`

produces fancy f, but the dmesg output produces non-fancy f, leading to the test failure.  Ugh ... perhaps the difference in environment variables was just a red herring.

Lemme dig some more, see if something comes up.

chu11:
i had a mild epiphany, maybe i was barking up the wrong tree the whole time.   I just noticed

`flux proxy $id flux jobs` produces fancy f job ids
`flux proxy $id dmesg-grep.py ....` produces ascii f job ids

so maybe this is just related to dmesg grep ... although i'm still unsure why.  dmesg-grep seems to able to output fancy f until some hand test scenarios.

grondo:
> `flux proxy $id dmesg-grep.py` .... produces ascii f job ids

`dmesg` and thus `dmesg-grep.py` are just spitting out what is in the dmesg logs. This means that the broker environment can't produce the fancy `f` in your testing. I'm not sure why that would be since it is just another process, and should be using the job environment passed through `flux alloc`, but maybe there's something I'm forgetting.

The other strange thing is that since we set `LC_ALL=C` in the sharness environment, shouldn't the f always be ascii?

chu11:
So I did this tweak in the test to see what happens

```
        locale &&
	id=$(flux alloc --bg -t5m -n4) &&
	echo $id
        flux proxy $id flux run locale &&
```

and there was a locale difference

```
LANG=C
LC_CTYPE="C"
LC_ALL=C
```

vs

```
LANG=C
LC_CTYPE=C.UTF-8
LC_ALL=
```

my non-expert knowledge of locale settings would suggest to me fancy f should be output instead of ascii f in the subinstance?

chu11:
Ok ... getting somewhere and also really confused ... in fluid.c

```
if (MB_CUR_MAX > 1 && streq (nl_langinfo (CODESET), "UTF-8")
```

```
       CODESET (LC_CTYPE)
              Return a string with the name of the character encoding used in the selected locale, such  as  "UTF-8",  "ISO-8859-1",  or
              "ANSI_X3.4-1968"  (better  known as US-ASCII).  This is the same string that you get with "locale charmap".  For a list of
              character encoding names, try "locale -m" (see locale(1)).
```

doing a test with LC_CTYPE="C.UTF-8"

```
#include <stdio.h>
#include <stdlib.h>
#include <langinfo.h>

int main()
{
    printf("%s\n", nl_langinfo (CODESET));
}
```

the output is

```
ANSI_X3.4-1968
```

so this sort of makes sense that if `LC_CTYPE` is not UTF-8 we get ascii-f.  But I have been unable to reproduce by setting these exact locales while hand testing.

chu11:
ok, adding more debugging

it appears in the test when the broker is launched the settings in the broker are

LC_ALL=C, LC_CTYPE=C.UTF-8 and `nl_langinfo (CODESET)` returns `ANSI_X3.4-1968`, thus ascii f is output in dmesg

however when `flux proxy` runs something in the subinstance, the settings are 

LC_ALL unset, LC_CTYPE=C.UTF-8 and `nl_langinfo (CODESET)` returns `UTF-8`, thus fancy f is output

I've been able to reproduce this by hand on the command line.  i.e.

set LC_ALL=C, LC_CTYPE=C.UTF-8
src/cmd/flux start
shell has LC_ALL unset
flux run hostname (to generate some job ids)
flux dmesg -> outputs ascii f
flux jobs -> outputs fancy f

I'm still not sure what system rune is turning LC_CTYPE to "C.UTF-8" in this environment.

chu11:
in my environment, I can also verify that if I set LC_ALL=C in my shell, LC_ALL becomes unset after `src/cmd/flux start`.

So I think this is the key, as I've learned that LC_ALL=C is not set by default in Livermore TOSS4 clusters.

When `flux start` is launched it is just being run as a process, so it absorbs the LC_ALL=C so job ids are output as ascii f in dmesg, but the subsequent shell that is launched behaves like a login shell and LC_ALL is not set, so the shell does not have LC_ALL set and outputs fancy f.

While not exactly the same behavior when `flux alloc` is called, I suspect it's the same general problem.

So effectively ... if we set LC_ALL=C in sharness.sh, I don't think we can guarantee it'll be set in a sub-instance that is created.

grondo:
Ok, this seems reasonable and we've already got an issue open on how the environment of `flux proxy` is being influenced here so I'll approve shortly.

grondo (approved):
LGTM.

Issue #5287, opened 2023-06-23
idset: exponential slowdown of idset_decode(3) with large integers


grondo:
`idset_decode(3)` shows an exponential performance issue when encoding large integers.
The following test program demonstrates the issue:
```c
#define _GNU_SOURCE
#include <flux/idset.h>
#include <stdio.h>
#include <limits.h>
#include <stdlib.h>
#include <time.h>

static struct timespec ts_diff (struct timespec start, struct timespec end)
{
    struct timespec temp;
    if ((end.tv_nsec-start.tv_nsec)<0) {
        temp.tv_sec = end.tv_sec-start.tv_sec-1;
        temp.tv_nsec = 1000000000+end.tv_nsec-start.tv_nsec;
    } else {
        temp.tv_sec = end.tv_sec-start.tv_sec;
        temp.tv_nsec = end.tv_nsec-start.tv_nsec;
    }
    return temp;
}

double since (struct timespec t0)
{
    struct timespec ts, d;
    clock_gettime (CLOCK_MONOTONIC, &ts);
    d = ts_diff (t0, ts);
    return ((double) d.tv_sec + (double) d.tv_nsec / 1000000000UL);
}

void test (const char *range)
{
    struct timespec ts;

    clock_gettime (CLOCK_MONOTONIC, &ts);
    struct idset *idset = idset_decode (range);
    char *s = idset_encode (idset, IDSET_FLAG_RANGE);
    printf ("%.3fs: %s\n", since (ts), s);
    free (s);
}

int main (int ac, char **av)
{
    unsigned int i = 100;
    for (; i < IDSET_INVALID_ID - 1; i*=2) {
        char *range;
        asprintf (&range, "%u", i);
        test (range);
        free (range);
    }
    return 0;
}
```

```console
$ ./tst
0.000s: 100
0.000s: 200
0.000s: 400
0.000s: 800
0.000s: 1600
0.000s: 3200
0.000s: 6400
0.000s: 12800
0.000s: 25600
0.000s: 51200
0.001s: 102400
0.002s: 204800
0.004s: 409600
0.004s: 819200
0.007s: 1638400
0.021s: 3276800
0.043s: 6553600
0.113s: 13107200
0.220s: 26214400
0.437s: 52428800
0.874s: 104857600
1.514s: 209715200
3.029s: 419430400
6.006s: 838860800
12.112s: 1677721600
tst: veb.c:88: encode: Assertion `8*(b-1) < WORD' failed.
Aborted
[...]

PR #5580, opened 2023-11-21
libidset: implement API for integer allocator use case

garlick:
Problem: idset is lacking a few interfaces needed for the "integer allocator" use case, hence the `flux_t` tagpool had to be implemented directly on top of veb.c, which makes it a unicorn in the code base, unnecessarily.

This adds some new flags and functions that make it convenient to use an idset for this use case.  Then the tagpool is re-implemented using the idset API, making it a bit easier to understand and maintain IMHO.  And maybe we will find other uses for these interfaces going forward.

The allocator functions require the idset to be created with IDSET_FLAG_INITFULL, which initializes it with all IDs in the set.  Alloc takes an id out of the set, and free returns it.  These functions are fast because finding the next free ID is O(log n) for a set with 2^n slots.

```c
/* Convenience interfaces for using idset as an "integer allocator".
 * idset_free() doesn't check for double frees, but idset_free_check() does.
 * It returns -1 and sets errno to EEXIST when that happens.
 */
int idset_alloc (struct idset *idset, unsigned int *val);
void idset_free (struct idset *idset, unsigned int val);
int idset_free_check (struct idset *idset, unsigned int val);
```

Alloc functions are defined in [idset_alloc(3)](https://flux-framework--5580.org.readthedocs.build/projects/flux-core/en/5580/man3/idset_alloc.html).
New flags and a couple more new functions are defined in [idset_create(3)](https://flux-framework--5580.org.readthedocs.build/projects/flux-core/en/5580/man3/idset_create.html).


garlick:
Imception builder failed here:
```
not ok 37 - flux-top shows expected data in queues after cancels
```
while the alpine builder failed here then apparently got stuck and was canceled
```
not ok 13 - job-info: update watch can be canceled (multiple watchers)
```
I don't _think_ these failures would be related to this pr so restarting.

grondo:
I'm going to go and ask the dumb question here: Why not implement the integer allocator as an empty set,  and as integers are allocated they are _set_ instead of _unset_? I'm sure there's a good reason for this, but it isn't necessarily clear to me why `idset_alloc()` and `idset_free()` couldn't be used in this more obvious way, which would apparently remove the need for an INITFULL flag.

src/common/libidset/idset.h
```diff
@@ -142,6 +142,14 @@ struct idset *idset_intersect (const struct idset *a, const struct idset *b);
  */
 bool idset_has_intersection (const struct idset *a, const struct idset *b);
 
+/* Convenience interfaces for using idset as an "integer allocator".
+ * idset_free() doesn't check for double frees, but idset_free_check() does.
+ * It returns -1 and sets errno to EEXIST when that happens.
+ */
+int idset_alloc (struct idset *idset, unsigned int *val);
+void idset_free (struct idset *idset, unsigned int val);
```
grondo, line 162: It appears these methods will fail if the idset was not created with `IDSET_FLAG_INITFULL`, which by name or comment is not obviously related. Perhaps a comment here, or rename the flag to something more obvious `IDSET_FLAG_ALLOCATOR`?

src/common/libidset/idset.h
```diff
@@ -28,6 +28,8 @@ enum idset_flags {
     IDSET_FLAG_BRACKETS = 2, // encode non-singleton idset with brackets
     IDSET_FLAG_RANGE = 4,    // encode with ranges ("2,3,4,8" -> "2-4,8")
     IDSET_FLAG_INITFULL = 8, // initilize/grow idset with all ids set
+    IDSET_FLAG_COUNT_BY_ITER = 16, // disable running count, which speeds up
+                             //  idset_set/clear, but slows down idset_count()
```
grondo, line 32: Another name for this flag could be `IDSET_FLAG_COUNT_LAZY` (only compute count when required). Meh, it is just slightly shorter.

src/common/libidset/idset.c
```diff
@@ -394,8 +394,6 @@ bool idset_equal (const struct idset *idset1,
 
     if (!idset1 || !idset2)
         return false;
-    if (idset_count (idset1) != idset_count (idset2))
-        return false;
```
grondo, line 279: This looks like it was an optimization, even though unnecessary. It might be helpful to mention in the commit message why the optimization wasn't needed or was a false optimization.

grondo: i.e. probably because it is slow when IDSET_FLAG_COUNT_BY_ITER is set?

src/common/libidset/idset.h
```diff
@@ -84,21 +91,31 @@ bool idset_test (const struct idset *idset, unsigned int id);
...
 
 /* Returns the last id in the idset.
  * Returns IDSET_INVALID_ID if the idset is empty.
  */
 unsigned int idset_last (const struct idset *idset);
 
+/* Return the previous id before 'id' in the idset.
+ * Returns IDSET_INVALID_ID if 'id' is the first id.
+ */
+unsigned int idset_prev (const struct idset *idset, unsigned int id);
```
grondo, line 107: Just a thought: If there is no current use case for idset_prev() it might make sense to leave it out for now in case the veb implementation is replaced in the future due to the poor space efficiency of vEB trees (discussed in #5287) and the replacement has trouble with a find previous operation?

grondo (commented):
This is a nice cleanup. I threw some thoughts in inline comments.

garlick:
Sorry to make you ask - I should have explained that in the PR description!  This note in flux-alloc(3) was intended to explain:
```
.. note::
  Unallocated is defined as "in the set" so that allocation can use the
  constant-time *first* operation to find the next available id.  Defining
  unallocated as "not in the set" would mean that iteration would
  be required to find the next available id.
```

grondo:
That makes sense! It would be a good addition to the commit message that adds INITFULL too...

src/common/libidset/idset.h
garlick: Well since INITFULL does not necessarily imply the allocator, maybe a comment would do?  I did say something in the man page but the header could be explicit too.

src/common/libidset/idset.h
garlick: I like it!

src/common/libidset/idset.c
garlick: OK, I'll work on that description.  Yes it was going to become inefficient.  In fact, now that you mention it, maybe it should remain, but be skipped if that flag is set.  With a comment inline.

src/common/libidset/idset.h
```diff
@@ -40,6 +40,10 @@ enum idset_flags {
 struct idset *idset_create (size_t size, int flags);
 void idset_destroy (struct idset *idset);
 
+/* Get the current number of slots in the idset.
```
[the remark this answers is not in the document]
grondo: Because I'm not certain what that means either is that just the total number of inegers that can currently be set in the idset? How is this related to what the rest of the documentation calls 'size'?

src/common/libidset/idset.h
grondo: A comment is fine - since it is critical, in the header file would be nice.

src/common/libidset/idset.h
garlick: I think that was the justification for leaving it out originally.  However, all of these possible replacements have both predecessor and successor operations (well croaring has an iterator that works forward and reverse - not sure how those would map to idsets simpler interfaces)

https://en.wikipedia.org/wiki/X-fast_trie

https://en.wikipedia.org/wiki/Y-fast_trie

https://github.com/RoaringBitmap/CRoaring/blob/59b740fa5c6dafb47bfc0fbc326b0080061a2441/include/roaring/roaring.h#L925C1-L930


and it seemed like a hole in the API to have first, next, and last but no prev.

However, it's not a big deal - we could leave it out until we need it if you like.

grondo: No, it is fine, it was just a thought since I didn't see any users.

grondo (approved):
Meant to approve, not just comment, since none of my comments were critical issues.

src/common/libidset/idset.h
grondo: BTW, the Wikipedia article for vEB trees now also mentions [Fusion tree](https://en.wikipedia.org/wiki/Fusion_tree) as another alternative with better space efficiency and better search than vEB (for large _w_). However, we're not dealing with large _w_ (and likely won't?) here so this discussion is largely academic (and sorry for the tangent!)

src/common/libidset/idset.h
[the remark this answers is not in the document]
garlick: Yes, this.

We should use consistent terminology so I'll see about tighting that up.  Just _size_ by itself is amgiguous.  For example, one might interpret it as the universe size (the number of possible IDs), the set size (number of IDs in the set), or the number of bytes consumed by the data structure.

src/common/libidset/idset.h
garlick: That's OK, good discussion!

garlick: Fusion tree also implements both predecessor and successor, so that's good.

garlick:
Just pushed changes to address those comments.  Thanks!

garlick:
OK to set MWP here?

grondo:
Yes!

PR #5584, opened 2023-11-21
libidset: improve decoding functions

garlick:
Problem: you can't control the flags used to create an idset when it comes from `idset_decode()`.

Admittedly, this might be a solution in search of a problem.  However, when pondering the "integer allocator" use case described in #5580, I thought it might be handy if one could create such an idset via `idset_decode()`, e.g. for allocating numbered resources or similar, but that's not currently possible.  Adding `idset_decode_ex()` also provided an opportunity to add the ability to return human readable parse errors, FWIW.

Some of the other functions fell out of refactoring the decode code to implement the above idea.  They may be used to improve efficiency in situations where today, one must create an temporary idset.

This is based on top of #5580.

New functions are documented in [idset_decode(3)](https://flux-framework--5584.org.readthedocs.build/projects/flux-core/en/5584/man3/idset_decode.html).


grondo (approved):
This LGTM.

garlick:
Great!  I'll set MWP.

PR #5585, opened 2023-11-21
kvs-watch: improve performance of kvs-watch disconnect/cleanup handling

grondo:
When profiling Flux during a throughput test, I noticed an unexpectedly large percentage of the broker's time was spent in `watcher_cancel_all()`. As previously discussed, this is because `flux_cancel_match()` needs to unpack a matchtag from the payload of a cancel request before comparison, and this since the function is called for each watcher in the list, the operation is repeated N times for each N watcher for a namespace. Given the profiling results, this appears to be a quite expensive operation.

This PR adjusts `watcher_cancel_all()` so that the matchtag is unpacked once is passed down to `watcher_cancel()` for each attempted match. The watchers list is also updated to a zlistx so that a copy of the list need not be taken each time a disconnect or cancel message is received.

This version of the PR avoids mucking with `flux_cancel_match()` sand `flux_msglist_cancel()` since there wasn't consensus on how to best do that.

### DIFF
--- a/src/modules/kvs-watch/kvs-watch.c
+++ b/src/modules/kvs-watch/kvs-watch.c
@@ -705,18 +705,20 @@ static void watcher_respond_ns (struct ns_monitor *nsm)
 static void watcher_cancel (struct ns_monitor *nsm,
                             struct watcher *w,
                             const flux_msg_t *msg,
+                            uint32_t matchtag,
                             bool cancel)
 {
-    bool match;
-    if (cancel)
-        match = flux_cancel_match (msg, w->request);
-    else
-        match = flux_disconnect_match (msg, w->request);
-    if (match) {
-        w->canceled = true;
-        w->mute = !cancel;
-        watcher_respond (nsm, w);
+    if (!flux_disconnect_match (msg, w->request))
+        return;
+    if (cancel) {
+        uint32_t tag;
+        if (flux_msg_get_matchtag (w->request, &tag) < 0
+            || tag != matchtag)
+            return;
     }
+    w->canceled = true;
+    w->mute = !cancel;
+    watcher_respond (nsm, w);
 }
 
 /* Cancel all namespace watchers that match:
@@ -726,6 +728,7 @@ static void watcher_cancel (struct ns_monitor *nsm,
  */
 static void watcher_cancel_ns (struct ns_monitor *nsm,
                                const flux_msg_t *msg,
+                               uint32_t matchtag,
                                bool cancel)
 {
     zlist_t *l;
@@ -734,7 +737,7 @@ static void watcher_cancel_ns (struct ns_monitor *nsm,
     if ((l = zlist_dup (nsm->watchers))) {
         w = zlist_first (l);
         while (w) {
-            watcher_cancel (nsm, w, msg, cancel);
+            watcher_cancel (nsm, w, msg, matchtag, cancel);
             w = zlist_next (l);
         }
         zlist_destroy (&l);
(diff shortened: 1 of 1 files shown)


grondo:
FYI - profiling results show that this PR decreases the contribution of kvs-watch `watcher_cancel_all` from 4-7% (depending on branch) during a profiling run, to <1%.

src/modules/kvs-watch/kvs-watch.c
```diff
@@ -682,19 +686,16 @@ static void watcher_respond (struct ns_monitor *nsm, struct watcher *w)
...
-            w = zlist_next (l);
-        }
-        zlist_destroy (&l);
+    w = zlistx_first (nsm->watchers);
+    while (w) {
+        /* Note: nsm->watchers may be destroyed after watcher_respond()
+         */
+        struct watcher *next = zlistx_next (nsm->watchers);
+        watcher_respond (nsm, w);
+        w = next;
```
garlick, line 697: I think a use-after-free is avoided here because `watcher_respond()` only destroys `nsm` (and all contained watchers) when the list is empty, so `next` would always be NULL in that case?

If that's the case it might be good to expand the comment a bit.

src/modules/kvs-watch/kvs-watch.c
```diff
@@ -731,19 +732,16 @@ static void watcher_cancel_ns (struct ns_monitor *nsm,
...
-        w = zlist_first (l);
-        while (w) {
-            watcher_cancel (nsm, w, msg, matchtag, cancel);
-            w = zlist_next (l);
-        }
-        zlist_destroy (&l);
+    w = zlistx_first (nsm->watchers);
+    while (w) {
+        /* Note: nsm->watchers may be destroyed after watcher_respond()
+         */
```
garlick, line 740: same comment as above

garlick (approved):
Excellent!  Just one comment abpit a comment so I'll approve.

src/modules/kvs-watch/kvs-watch.c
grondo: Yes, good point.

garlick:
Just noticed this is based on top of #5559?   Was that intentional?

grondo:
Oh yeah, I was just profiling with that PR combined with this one and accidentally pushed. I've fixed that now.

src/modules/kvs-watch/kvs-watch.c
```diff
@@ -682,19 +686,21 @@ static void watcher_respond (struct ns_monitor *nsm, struct watcher *w)
...
+    while (w) {
+        /* Note: get next watcher before calling watcher_respond() since
+         * nsm->watchers may be destroyed when next == NULL:
+         */
+        struct watcher *next = zlistx_next (nsm->watchers);
+        watcher_respond (nsm, w);
+
+        /* Note: No use-after-free possible since nsm->watchers only destroyed
+         * if next == NULL
+         */
```
chu11, line 701: it took a bit for me to grok "next == NULL" --> "nsm->watchers list is empty", perhaps could just add that :-)

src/modules/kvs-watch/kvs-watch.c
```diff
@@ -682,19 +686,21 @@ static void watcher_respond (struct ns_monitor *nsm, struct watcher *w)
  */
 static void watcher_respond_ns (struct ns_monitor *nsm)
```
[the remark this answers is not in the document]
grondo: Oh yeah, duh. I had this in the first version of this PR but forgot it this time.

src/modules/kvs-watch/kvs-watch.c
grondo: It doesn't necessary mean the list is empty if `next == NULL`, only that this is the final iteration of the loop. The comment was requested by @garlick to call out that nsm->watchers is only destroyed if `next == NULL` so a use-after-free is not possible here.

grondo: Given multiple comments about this code, perhaps it would be clearer if destruction of the list was done after list traversal instead of during (I was just being lazy and trying to avoid changing existing code as much as possible),  I'll poke at that if I have time.

grondo: Actually, that won't be simple since `watcher_cleanup()` is called from multiple locations. I cleaned up the comment a bit to hopefully be more clear (and also reference that it is `nsm` that is destroyed, not just `nsm->watchers`)

chu11 (approved):
LGTM, just one nit below

src/modules/kvs-watch/kvs-watch.c
```diff
@@ -677,24 +681,24 @@ static void watcher_respond (struct ns_monitor *nsm, struct watcher *w)
 
 /* Respond to all ready watchers.
  * N.B. watcher_respond() may call zlist_remove() on nsm->watchers.
```
chu11, line 683: is it zlistx_delete() now?

grondo:
I'll set MWP here now.

grondo:
@mergifyio refresh

Issue #5566, opened 2023-11-16
Older Flux instance can't launch newer


grondo:
Oops. When the job shell and other modules were converted to using `job-info.watch-update` to monitor _R_, this broke launching newer versions of Flux under older versions. We'll have to do some compatibility work to get that going again:
```console
$ flux run -n1 -o pty.interactive src/cmd/flux start 
Nov 16 15:16:29.561746 resource.err[0]: lookup R from enclosing instance KVS: Function not implemented
Nov 16 15:16:29.561795 resource.crit[0]: module exiting abnormally
Nov 16 15:16:29.562192 broker.err[0]: rc1.0: flux-module: load resource: Function not implemented
Nov 16 15:16:29.562978 broker.err[0]: rc1.0: /g/g0/grondo/git/flux-core.git/etc/rc1 Exited (rc=1) 1.1s
[detached: session exiting]
```

PR #5589, opened 2023-11-25
support start under older versions of Flux without the job-info.update-watch RPC

grondo:
This PR fixes issue #5566 (older Flux instance can't launch newer). The resource module will now fall back to the `job-info.lookup` RPC if `job-info.update-watch` returns `ENOSYS`.

Instead of trying to test against an older Flux version, this PR also adds an option to the job-info to "disable" the `job-info.update-watch` RPC, and uses that to force the fallback code for testing purposes.

There is also an unrelated memory leak fix here, detected during testing. That could be moved to its own PR if desired.


src/broker/boot_pmi.c
```diff
@@ -44,18 +44,17 @@ static int set_instance_level_attr (struct upmi *upmi,
...
-    if (attr_add (attrs, "instance-level", level, ATTR_IMMUTABLE) < 0)
-        return -1;
-    if (rc == 0)
         jobid = name;
-    if (attr_add (attrs, "jobid", jobid, ATTR_IMMUTABLE) < 0)
+    }
+    rc = attr_add (attrs, "instance-level", level, ATTR_IMMUTABLE);
+    free (val);
+    if (rc < 0 || attr_add (attrs, "jobid", jobid, ATTR_IMMUTABLE) < 0)
         return -1;
```
garlick, line 58: Oh good catch, thanks!

src/modules/resource/inventory.c
```diff
@@ -512,20 +551,29 @@ static int start_resource_watch (struct inventory *inv)
...
-    inv->R_watch_f = f;
 
     /* Get first response synchronously
      */
     if (flux_rpc_get_unpack (f, "{s:o}", "R", &job_R) < 0) {
-        flux_log_error (h, "lookup R from enclosing instance KVS");
-        goto done;
+        if (errno == ENOSYS) {
+            /*  Parent instance doesn't support job-info.update-watch.
+             *  Fall back to job-info.lookup and return:
```
garlick, line 560: I like the way this change doesn't disrupt the flow in the new code.

Suggestion: add a note that job-info.update-watch was added in v0.56.0.  Maybe we'll be able to drop this at some point when it seems like 0.55.0 and earlier are no longer being used.

SIde note: it would be good to have semantic versioning rules that apply to RPCs as well as APIs so we can plan deprecation without going insane once we're in v1.0 territory and beyond.

src/modules/job-info/job-info.c
```diff
@@ -170,6 +170,12 @@ int mod_main (flux_t *h, int argc, char **argv)
         flux_log_error (h, "initialization error");
         goto done;
     }
+    /* Process any command line arguments:
+     */
+    for (int i = 0; i < argc; i++) {
+        if (strcmp (argv[i], "update-watch-disable"))
```
garlick, line 176: should that be `streq`?

garlick (approved):
LGTM! 

Just one problem with the module test option.

src/modules/job-info/job-info.c
grondo: D'oh! Yes of course.

grondo:
Oh, oops, the scheme used for testing here won't work since the job shell also needs the `job-info.update-watch` RPC. Will have to think of something else.

grondo:
Ok, I replaced the `job-info` module option with a new `resource` module option (`resource.no-update-watch`) which, when true, causes the resource module to send the `job-info.update-watch` RPC to an invalid service name, resulting in `ENOSYS` and thereby triggering the fallback code.

Using a resource module option unfortunately adds a bit more code than I'd like, but it does it make it super easy to test using a one-liner...

grondo:
Setting MWP on this one

Issue #5582, opened 2023-11-21
Unable to SSH across clusters - Invalid argument / No such file or directory


alexschroeter:
Hi,

I tried to follow the "[SSH across clusters](https://flux-framework.readthedocs.io/en/latest/tutorials/commands/ssh-across-clusters.html)" tutorial but got stuck.

# System Setup
host1 is running conda install flux
```
(flux) $ echo $FLUX_URI
local:///tmp/flux-tZGJR7/local-0
```

host2 is setup via spack

# Connection
I set up the ssh so I can establish a connection without entering a password. This works by 'ssh \<host1\>', I don't know if this is relevant but to connect to <host1> I am proxy jumping through a <gateway>. 

So the command I am running is `flux proxy ssh://<host1>/tmp/flux-tZGJR7/local-0`

# Error
```
$ flux proxy ssh://<host1>/tmp/flux-tZGJR7/local-0
bash: line 1: /home/<group>/<user>/src/spack/opt/spack/linux-almalinux9-zen2/gcc-11.3.1/flux-core-0.55.0-p3mxu2rwqfynsdar5p52cfcexbbuyoc2/bin/flux: No such file or directory
flux-proxy: ssh://<host1>/tmp/flux-tZGJR7/local-0: Invalid argument
```

# Expected
I think this is not a connection issue but rather that I am making a mistake in the configuration or that there is an issue with my installation.
The "No such file or directory" confuses me since I don't know what it refers to. The binary exists.
```
ls -l /home/<group>/<user>/src/spack/opt/spack/linux-almalinux9-zen2/gcc-11.3.1/flux-core-0.55.0-p3mxu2rwqfynsdar5p52cfcexbbuyoc2/bin/flux
-rwxr-xr-x 1 <user> <group> 1337464 Nov 14 16:03 /home/<group>/<user>/src/spack/opt/spack/linux-almalinux9-zen2/gcc-11.3.1/flux-core-0.55.0-p3mxu2rwqfynsdar5p52cfcexbbuyoc2/bin/flux
```

Any pointers in the right direction would be much appreciated.
Thanks, Alex

alexschroeter:
Thank you, this was the problem. I was able to connect without any problems.

PR #5591, opened 2023-11-26
improve ssh connector reliability with different installation paths

### LINKED ISSUE #5583: ssh connector is not very good at finding remote flux(1) executable
Problem: when the `ssh://` connector is used to remotely access a flux instance, the local `flux(1)` path is used remotely unless overridden with FLUX_SSH_RCMD.  This is a rather poor situation for users navigating heterogeneously managed environments.

Idea: since the remote rundir is already part of the URI, could we stash a symlink link to flux(1) in the remote `rundir`, e.g.
```
$ ls -l $(flux getattr rundir)
total 20
-rw-r--r-- 1 garlick garlick 20480 Nov 21 05:28 content.sqlite
lrwxrwxrwx 1 garlick garlick    13 Nov 21 05:29 flux -> /usr/bin/flux
srwxrwxrwx 1 garlick garlick     0 Nov 21 05:28 local-0
```
Then have the ssh connector use the rundir to build the remote flux(1) path?

garlick:
Problem: the ssh connector tries to remotely execute `flux-relay` using a path derived from the local flux installation.  This tends not to work well when the remote flux is installed to a different path, as was encountered in #5582.

This  PR improves the odds of getting it right with the following changes:
1. the broker places a symlink to its bindir in `rundir/bin`
2. the ssh connector derives `rundir` from the `ssh://` URI and places `rundir/bin` first in the remote PATH

The local directory that contains `flux(1)` is appended to the remote PATH, so that there should be no change in behavior when using `flux proxy` to an older version of flux that doesn't have this feature.  `/usr/bin:/bin` is also appended so `flux(1)` can be a libtool wrapper (the wrapper scripts need to run some basic commands).

The remote command path can still be overridden by setting `FLUX_SSH_RCMD`.

Fixes #5583

(Marking WIP as tests are needed, but feedback welcome)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -206,6 +206,7 @@ AC_CHECK_FUNCS( \
 # See src/common/libmissing/Makefile.am
 AC_REPLACE_FUNCS( \
   strlcpy \
+  strlcat \
   argz_add \
   envz_add \
 )
--- a/src/broker/Makefile.am
+++ b/src/broker/Makefile.am
@@ -9,6 +9,7 @@ AM_CPPFLAGS = \
 	-I$(top_srcdir)/src/include \
 	-I$(top_srcdir)/src/common/libccan \
 	-I$(top_builddir)/src/common/libflux \
+	-DABS_TOP_BUILDDIR=\"${abs_top_builddir}\" \
 	$(ZMQ_CFLAGS) \
 	$(LIBUUID_CFLAGS) \
 	$(JANSSON_CFLAGS) \
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -18,6 +18,7 @@
 #include <inttypes.h>
 #include <sys/prctl.h>
 #include <sys/resource.h>
+#include <fcntl.h>
 #include <unistd.h>
 #include <sys/un.h>
 #include <sys/types.h>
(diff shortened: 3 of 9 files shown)


garlick:
> Would it work to place a link to the `flux` executable in `rundir` instead of the whole bindir? 

Symlinking the executable so the remote command would be `/rundir/flux proxy` was my first thought, but it wouldn't be backwards compatible with older versions of flux that lack the symlink. Users could still work around this using `FLUX_SSH_RCMD=flux` or similar though.  If we think that would fly, then I'm for it!

> Is this a change in that when proxying from a Flux installation of a different version, the flux-relay of the remote instance version will be used with the flux-proxy of the local version?

Yes that is true.

grondo:
>  Symlinking the executable so the remote command would be /rundir/flux proxy was my first thought, but it wouldn't be backwards compatible with older versions of flux that lack the symlink. 

Oh, I was thinking you could just add `rundir` to `PATH` and have the same backwards compatible behavior as if you linked `bindir`, but restricted to just placing the `flux` executable in the `PATH` instead of everything in `bindir`...

garlick:
Ah - that should work fine.  Maybe it would be slightly cleaner still to put the `flux` symlink in a subdirectory of `rundir` and put _that_ in the path, since otherwise there would be non-executables on the search path.

grondo:
Oh, yeah. Actually that directory of links could be useful in other situations to encapsulate a set of executables we want to push first in path within an instance, though at this point I could go either way on this and if you prefer linking to `bindir` instead of select executables under `rundir` then I'm fine with that for now (e.g. if it is actually simpler.)

garlick:
Pushed that suggested change and rebased on current master.

grondo (approved):
This LGTM!

grondo:
Is this ready for MWP?

PR #5598, opened 2023-11-28
testsuite: fix race in job info update lookups

### LINKED ISSUE #5594: FAIL: t2233-job-info-update.t 3 - job-info: update lookup with update events works (job active)
I've seen this fail twice recently (once bookworm - gcc-12,content-s3,distcheck and once alpine builder) and saw a similar jobspec variant in #5593 also fail.

```
expiration_add() {
	local id=$1
	local amount=$2
	local kvsdir=$(flux job id --to=kvs $id)
	local new=$(($(date +%s) + $amount))
	flux kvs eventlog append \
	    ${kvsdir}.eventlog resource-update "{\"expiration\": $new}"
	echo $new
}
```

```
  expecting success: 
  	update1=$(expiration_add $jobid 100) &&
  	flux job eventlog $jobid &&
  	${UPDATE_LOOKUP} $jobid R > lookup3A.out &&
  	update2=$(expiration_add $jobid 200) &&
  	${UPDATE_LOOKUP} $jobid R > lookup3B.out &&
  	flux cancel $jobid &&
  	cat lookup3A.out | jq -e ".execution.expiration == ${update1}" &&
  	cat lookup3B.out | jq -e ".execution.expiration == ${update2}"
  
  1701130314.421952 submit userid=1001 urgency=16 flags=0 version=1
  1701130314.434043 validate
  1701130314.444746 depend
  1701130314.444784 priority priority=16
  1701130314.446792 alloc annotations={"sched":{"resource_summary":"rank0/core0"}}
  1701130314.449819 start
  1701130314.546236 resource-update expiration=1701130414
  true
  false
  not ok 3 - job-info: update lookup with update events works (job active)
  ```

I'm not sure if there's something racy going on in there that I can't see at the moment.

(as an aside, the `flux job eventlog $jobid` appears to be an errant debugging that is left over)

### PR DESCRIPTION
**Pull request by chu11**
Problem: Several tests in t2233-job-info-update.t that test the job-info.update-lookup target can be racy.  Internally, the eventlog append may not yet have been seen by a job-info watcher, and the update is therefore not yet seen in the lookup test.

Solution: Add a wait loop around the check to ensure any raciness is avoided.

Fixes #5594

### DIFF
--- a/t/t2233-job-info-update.t
+++ b/t/t2233-job-info-update.t
@@ -51,31 +51,45 @@ expiration_add() {
 	echo $new
 }
 
+# Usage: check_expiration jobid VALUE
+# Check and wait for expiration to reach VALUE.
+#
+check_expiration() {
+	local jobid=$1
+	local value=$2
+	local i=0
+	while (! ${UPDATE_LOOKUP} ${jobid} R | jq -e ".execution.expiration == ${value}" \
+		&& [ $i -lt 200 ] )
+	do
+		sleep 0.1
+		i=$((i + 1))
+	done
+	if [ "$i" -eq "200" ]
+	then
+		return 1
+	fi
+	return 0
+}
+
 test_expect_success 'job-info: update lookup with no update events works (job active)' '
 	jobid=$(flux submit --wait-event=start sleep 300) &&
-	${UPDATE_LOOKUP} $jobid R > lookup1.out &&
-	flux cancel $jobid &&
-	test_debug "cat lookup1.out" &&
-	cat lookup1.out | jq -e ".execution.expiration == 0.0"
+	check_expiration $jobid 0.0 &&
+	flux cancel $jobid
 '
 
 test_expect_success 'job-info: update lookup with no update events works (job inactive)' '
 	jobid=$(flux submit --wait-event=clean hostname) &&
-	${UPDATE_LOOKUP} $jobid R > lookup2.out &&
-	test_debug "cat lookup2.out" &&
-	cat lookup2.out | jq -e ".execution.expiration == 0.0"
+	check_expiration $jobid 0.0
 '
 
 test_expect_success 'job-info: update lookup with update events works (job active)' '
 	jobid=$(flux submit --wait-event=start sleep 300) &&
 	update1=$(expiration_add $jobid 100) &&
 	flux job eventlog $jobid &&
-	${UPDATE_LOOKUP} $jobid R > lookup3A.out &&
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: t/t2233-job-info-update.t [hunk 1]
```diff
@@ -51,19 +51,39 @@ expiration_add() {
...
+	do
+		sleep 0.1
+		i=$((i + 1))
+	done
+	if [ "$i" -eq "200" ]
+	then
+		return 1
+	fi
+	return 0
+}
```
grondo, line 72: I am confused how this works. I'm probably missing something obvious, but `check_expiration` is continuously doing `cat ${file} | jq` until an expected value is observed. However, in the test, nothing I see is updating the file, so how could the contents of the file change? I don't think `update_lookup` is running in the background or anything (maybe just missing it)

t/t2233-job-info-update.t
```diff
@@ -69,7 +69,6 @@ test_expect_success 'job-info: update lookup with no update events works (job in
 test_expect_success 'job-info: update lookup with update events works (job active)' '
 	jobid=$(flux submit --wait-event=start sleep 300) &&
 	update1=$(expiration_add $jobid 100) &&
-	flux job eventlog $jobid &&
```
grondo, line 72: Debugging in tests is useful when the test fails, so you can actually see what was happening. If anything this should be under `test_debug`, but functionally it is no different. If the output isn't harming anything, maybe just leave it in?

t/t2233-job-info-update.t
chu11: oh crap ... the answer is I just flubbed this :-)

chu11:
re-pushed, I just completely flubbed the fix ... I think I got it right this time.

t/t2233-job-info-update.t
```diff
@@ -51,40 +51,53 @@ expiration_add() {
...
+	return 0
+}
+
 test_expect_success 'job-info: update lookup with no update events works (job active)' '
 	jobid=$(flux submit --wait-event=start sleep 300) &&
-	${UPDATE_LOOKUP} $jobid R > lookup1.out &&
-	flux cancel $jobid &&
-	test_debug "cat lookup1.out" &&
-	cat lookup1.out | jq -e ".execution.expiration == 0.0"
+	check_expiration $jobid 0.0
```
grondo, line 76: Missing `&&`

chu11:
doh, re-pushed adding the missing `&&`

Issue #5603, opened 2023-11-30
can't attach debugger when doing `flux run` and then `flux job attach` from login node.


lee218llnl:
When I do a `flux alloc` and then within the allocation do a `flux run MyApp` I can then `flux attach --debug <JobID>` and attach STAT to the application. However, if I don't explicitly `flux alloc`, but rather directly invoke `flux run MyApp` from a login node and then `flux attach --debug <JobID>`, then STAT gets an error when trying to launch its daemons. The session below demonstrates the behavior (where tioga11 is a login node):

```
tioga11{testnewm}24: flux run -N 2 -n 32 -t 30 a.out &
[1] 3718629
tioga11{testnewm}25: tioga29, MPI task 1 of 32 stalling

tioga11{testnewm}25: flux jobs -A | grep testnewm
f2X8PsZb3eXH pdebug   testnewm a.out       R     32      2   8.840s tioga[29-30]
tioga11{testnewm}26: flux job attach --debug f2X8PsZb3eXH &
[2] 3718742
tioga11{testnewm}27: tioga29, MPI task 1 of 32 stalling

tioga11{testnewm}27: ps x | grep 'job attach'
3718629 pts/60   S      0:00 flux-job attach f2X8PsZb3eXH
3718742 pts/60   S      0:00 job attach --debug f2X8PsZb3eXH
3718820 pts/60   S+     0:00 grep --color=auto job attach
tioga11{testnewm}28: stat-cl 3718742

(STATmain.py:3718889): dbind-WARNING **: 16:11:11.874: Couldn't connect to accessibility bus: Failed to connect to socket /tmp/dbus-SDIG3OSUwY: Connection refused
STAT started at 2023-11-29-16:11:11
Attaching to job launcher (null):3718742 and launching tool daemons...
<Nov 29 16:11:11> <Launchmon> (INFO): Just continued the RM process out of the first trap
flux-exec: Error: rank 18: /collab/usr/global/tools/stat/toss_4_x86_64_ib/spack3/opt/spack/linux-rhel8-ivybridge/gcc-8.5.0/stat-4.2.1-mqkguv2cfenhe2xtn7yooes4klb72d7o/bin/STATD: Request requires owner credentials
[...]

PR #5605, opened 2023-11-30
shell: add rexec plugin

garlick:
Thinking about #5603, I remembered I had this laying around from when I worked on libsubprocess a while back.  It seems to work at least for a trivial test like this:
```
$ flux alloc -N10.038s: flux-shell[0]: rexec: registered rexec service as 5588-shell-fN5pUdPD.rexec
$ flux --parent exec  --service 5588-shell-f3qpTA6T.rexec hostname
system76-pc
```
Maybe we could fairly quickly implement the  `flux exec --jobid=ID` feature suggested by @grondo and have at least a prototype of what's needed to launch tool daemons on multi-user flux?


grondo:
@garlick - I pushed an experimental set of commits to add `flux exec --jobid`, hope that's ok. This also makes `-x, --exclude` and `-r, --rank` relative to the job ranks when `-j, --jobid` is used such that `flux exec --jobid=ID -x 0` will exclude job rank 0, and `flux exec --jobid=ID -r 0` will run only on job rank 0.

Let me know what you think.

garlick:
Nice job!   Quick comment:

In the docs, "job ranks" might be misunderstood as corresponding to a task rather than a shell.  Maybe "job node ranks" or "job shell ranks" or similar would be better?

Actually I see that "job ranks" and `{ranks}` is used to refer to shell ranks in [flux-jobs(3)](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man1/flux-jobs.html#output-format) also.  Hmm, maybe for now we should just be consistent.

grondo:
Good thought. The flux-exec manpage uses "nodeset" so maybe job nodeset?

grondo:
Ok, I revised the wording (a bit out of it today, so needs a second look!) and fixed a comparison of an unsigned result `< 0`.

garlick:
Ah a problem here: the subprocess server message handlers are registered with the rolemask set to 0, so attempts to use the shell service fail with e.g.
```
$ flux exec --jobid=ƒgrZYS9y6as flux getattr rank
flux-exec: Error: rank 0: flux: Request requires owner credentials
```

We'll need to add a flag or something to `subprocess_server_create()` so that the message dispatch code doesn't automatically deny the job owner's requests.

I guess the tests must be passing because the job owner is also the instance owner but I haven't checked in detail.

grondo:
Oh interesting (and probably good!), I did verify that without the auth_cb a guest request mocked with `FLUX_HANDLE_USERID` was able to use the shell rexec service of anoher user's job. I guess this is because I forgot to also reset `FLUX_HANDLE_ROLEMASK`? (That should be fixed in he test)

garlick:
Just pushed an update which adds a `rolemask` argument to `subprocess_server_create()`.   This works now.  Cool!
```
garlick@picl0:~$ flux exec -l --jobid ƒgraD1ShHLb  flux getattr rank
4: 4
3: 3
```

garlick:
This is failing but only on focal.  Hmm.
```
 	FLUX_HANDLE_USERID=$(($(id -u)+1)) \
  		test_must_fail flux exec -r 0 --jobid=$jobid id 2>eperm0.err &&
  	test_debug "cat eperm0.err" &&
  	grep "Access denied" eperm0.err &&
  	FLUX_HANDLE_USERID=$(($(id -u)+1)) \
  		test_must_fail flux exec -r 1 --jobid=$jobid id 2>eperm1.err &&
  	test_debug "cat eperm1.err" &&
  	grep "Access denied" eperm1.err
  
  uid=1001(runner) gid=1001(runner) groups=1001(runner),27(sudo)
  not ok 12 - flux exec --jobid fails from other user
```

garlick:
Hmm, should we be setting any of the job environment variables in subprocesses started this way?

https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man7/flux-environment.html#job-environment

It would seem like FLUX_KVS_NAMESPACE, FLUX_JOB_TMPDIR, and FLUX_JOB_NNODES might be useful, if nothing else.

grondo:
A good question. I'm not sure, given that flux exec wouldn't do that in the existing case. I can't think of why not to do it though.

garlick:
We'd have to add some new interfaces to the subprocess server to do it.  Maybe it would be OK to just leave it as is for now and consider that as an add-on later if needed.

grondo:
> This is failing but only on focal. Hmm.

The problem seems to be that `FLUX_HANDLE_USERID` can't be passed through `test_must_fail` on this distro. I'm unsure why, but I've amended that test to use a shell function instead and force pushed the result (Hope that's ok!)

doc/man1/flux-exec.rst
```diff
@@ -67,6 +67,16 @@ OPTIONS
...
 
+.. option:: -j, --jobid=JOBID
+
+   Use the job shell exec service for job *JOBID* instead of the broker's
+   exec service. The default *NODESET* will be set to the nodes assigned
+   to the job, and the :option:`--rank` and :option:`--exclude` options are
+   applied relative to this set. For example, :option:`flux exec -j ID -r 0`
+   will run only on the first node assigned to *JOBID*, and :option:`flux
+   exec -j ID -x 0` will run on all nodes assigned to *JOBID* except the
+   first node.
```
garlick, line 70: It feels like this paragraph needs a stronger statement that the ranks are shell ranks not broker ranks.

grondo: We changed the use of "job ranks" here based on your previous comment:
(My assumption is the casual reader will not know what "job shell ranks" means, so I instead gave examples)

> In the docs, "job ranks" might be misunderstood as corresponding to a task rather than a shell. Maybe "job node ranks" or "job shell ranks" or similar would be better?

> Actually I see that "job ranks" and {ranks} is used to refer to shell ranks in [flux-jobs(1)](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man1/flux-jobs.html#output-format) also. Hmm, maybe for now we should just be consistent.

Feel free to rewrite however you think is best.

grondo: BTW, I forgot to mention that the way `flux exec` works was changed slightly by some of the commits I added here. Previously, the "nodeset" (as it is called in he man page) is initialized via the `-r, --ranks=` option or it set to "all" ranks using the size attribute.

Now the nodeset is always initialized to the set of all broker ranks, unless `-j, --jobid` is used in which case it is initialized to the set of ranks assigned to the job. Then the `-r, --ranks` and `-x, --exclude` options are used to include/exclude ranks within that set by index. This works the same as before for `-r, --ranks` (since the index matches the broker rank for this case) _except_ that you can no longer specify invalid ranks (which seemed an improvement, but maybe there's an argument otherwise?) It also allows `-r` and `-x` to be used to filter the target ranks with `--jobid` as well since otherwise you'd have to know a priori the ranks assigned to the job, which wouldn't be usable IMO.

grondo: Another thought - the `flux-exec(1)` manpage is the only one that seems to use the term "nodeset". Probably this was one of the early manpages. Maybe that term doesn't make sense anymore, and this manpage needs a complete rewrite with modern terminology? We could point to a definition of "job shell rank" in the documentation of `--jobid`, but AFAICS that term doesn't appear anywhere else in the flux-core docs, so maybe the whole manual could be cleaned up with consistent terminology?

doc/man1/flux-exec.rst
```diff
@@ -52,21 +53,27 @@ OPTIONS
...
 
-   Use the job shell exec service for job *JOBID* instead of the broker's
-   exec service. The default *IDSET* will be set to the nodes assigned
-   to the job, and the :option:`--rank` and :option:`--exclude` options are
-   applied relative to this set. For example, :option:`flux exec -j ID -r 0`
-   will run only on the first node assigned to *JOBID*, and :option:`flux
-   exec -j ID -x 0` will run on all nodes assigned to *JOBID* except the
-   first node.
+   Run *COMMAND* on the nodes allocated to *JOBID* instead of the nodes
+   assigned to the Flux instance.
```
grondo, line 70: This drops the note that he job shell exec service is used instead of the builtin broker exec service. Should something to that effect be included -- at least a hint  that this gives guests the ability to use `flux exec` against their jobs in a multiuser instance?

grondo:
Updates to `flux-exec(1)` look great to me! (just one trivial question inline). Nice!

doc/man1/flux-exec.rst
garlick: Oh good point.  I just pushed a small fixup that I think clarifies this?

grondo: Yes, that's great! Feel free to squash.

garlick:
Rebased on current master.

I also added a check that prevents the plugin from being loaded if the job is running as a guest, since we can't trust the credentials of messages that pass through the instance running as a different user.

garlick:
I realized in playing with this that returning ENOSYS when this is attempted from a multi-user instance is not a very nice error message and so probably the service should be registered just so it can return a better error message.

I can push that on top of what's here but it might be better to squash the incremental development.  Any objections to squashing?

grondo:
No objections here!

garlick:
Well I screwed up the authorization check, and the new system test caught it so that's good :-)  Fixed now.

This is probably ready for a review.

src/shell/rexec.c
```diff
@@ -0,0 +1,136 @@
...
+static int rexec_auth_cb (const flux_msg_t *msg,
+                          void *arg,
+                          flux_error_t *errp)
+{
+    struct shell_rexec *rexec = arg;
+
+    if (!rexec->parent_is_trusted) {
+        errno = EPERM;
+        return errprintf (errp, "Access denied");
+    }
```
grondo, line 69: Is the normal uid auth check still necessary here? That is, would this prevent a job running as user `flux` in a system instance from allowing an arbitrary user to exec something as user `flux` within that job?

garlick: Should be OK.  An earlier version of this PR included a commit that allowed the subprocess server message handlers to be registered with an arbitrary rolemask, but I reverted that and so now they are registered for owner only.  I think the combination of that check and the one quoted should be sufficient.

garlick:
oops getting a failure here - this is skipped when built without flux security.  I suspect it's just that the error message changed but will run it down
```
2024-02-29T19:28:48.7882278Z FAIL: t0005-exec-jobid.t 12 - flux exec --jobid fails from other user
2024-02-29T19:28:48.7882706Z #	
2024-02-29T19:28:48.7882949Z #		alt_userid=$(($(id -u)+1)) &&
2024-02-29T19:28:48.7883342Z #		test_debug "echo testing with handle userid=$alt_userid" &&
2024-02-29T19:28:48.7883850Z #		test_must_fail test_exec $alt_userid $jobid 0 2>eperm0.err &&
2024-02-29T19:28:48.7884273Z #		test_debug "cat eperm0.err" &&
2024-02-29T19:28:48.7884592Z #		grep "Access denied" eperm0.err &&
2024-02-29T19:28:48.7885008Z #		test_must_fail test_exec $alt_userid $jobid 1 2>eperm1.err &&
2024-02-29T19:28:48.7885421Z #		test_debug "cat eperm1.err" &&
2024-02-29T19:28:48.7885813Z #		grep "Access denied" eperm1.err
2024-02-29T19:28:48.7886106Z #

**Comment by garlick** [comment]
well I'll be darned, I think you were right.  The test is failing because it's letting the other user in.  Now I need to understand why!

**Comment by garlick** [comment]
I re-added the credential userid check because the test is mocking it to another user.  We do know the shell is running as the same userid as the instance because of the other check, so I think this is just covering that specific situation but I the extra check and the comment block probably help people understand the possible attacks better so it's good documentation if nothing else.

**Comment by grondo** [comment]
>  We do know the shell is running as the same userid as the instance because of the other check, so I think this is just covering that specific situation, but the extra check and the comment above it probably help people understand the possible attacks better so it's good documentation if nothing else.

Ah, was the problem that the test sets `FLUX_HANDLE_USERID` but not `FLUX_HANDLE_ROLEMASK` and the default msg handler is check is against the rolemask not the encoded uid. Seems like the test should be setting the rolemask ask well perhaps.

**Comment by garlick** [comment]
> Ah, was the problem that the test sets FLUX_HANDLE_USERID but not FLUX_HANDLE_ROLEMASK and the default msg handler check is against the rolemask not the encoded uid? Seems like the test should be setting the rolemask ask well perhaps.

Yeah, I guess.  OK, will fix.

**Comment by garlick** [comment]
OK pushed that change.  Now it fails with the following error and doesn't reach the exec service, but that's probably fine.  We do test a real user trying this in the system test that was added.
```
flux-exec: failed to get shell.init event for fmrsGaB
```

**Comment by garlick** [comment]
OK, whew, hopefully that works in CI

**Comment by grondo** [comment]
Hit one of these mysterious failures, I'll restart the failed build:
```

  make[4]: *** [Makefile:1061: _rlist.c] Error 1
  make[4]: *** Waiting for unfinished jobs....
    GEN      _hostlist.c
    GEN      _security.c
    GEN      _core.c
  make[3]: *** [Makefile:499: all-recursive] Error 1
  make[2]: *** [Makefile:499: all-recursive] Error 1
  make[1]: *** [Makefile:515: all-recursive] Error 1
  make: *** [Makefile:594: all-recursive] Error 1
```

grondo (approved):
LGTM!

garlick:
Yay!  Setting MWP.

Issue #5595, opened 2023-11-28
ci: alpine builder hanging


chu11:
The alpine CI appears to be hanging on occasion (note that this is independent of tests failing, which alpine may be exposing some raciness in some tests).

In atleast two recent runs the hang appears to be in `t0020-terminus`, the end of the output log is ...

```
  ok 32 - flux-terminus: disconnect works
  flux-terminus: disconnect works
  
  expecting success: 
  	test_expect_code 1 \
  	    $runpty flux terminus start \
  	    flux terminus start true
  
  {"version": 2, "width": 80, "height": 25, "timestamp": 1701096853}
  [0.02237534523010254, "o", "\u001b[2J\u001b[;H"]
  [0.038247108459472656, "o", "flux-terminus: Nesting flux-terminus sessions not supported\r\n\u001b[999H[detached: session exiting]\u001b[K\n\r\u001b[?25h"]
  [0.03834104537963867, "o", ""]
  ok 33 - flux-terminus: nesting not allowed
  flux-terminus: nesting not allowed
  
  expecting success: 
  	flux terminus start -d &&
  	( SERVICE="$(id -u)-terminus" &&
            export FLUX_HANDLE_USERID=$(($(id -u) + 1)) &&
  	  export FLUX_HANDLE_ROLEMASK=0x2 &&
  	  test_expect_code 1 flux terminus attach -s $SERVICE 0 &&
  	  test_expect_code 1 flux terminus list -s $SERVICE &&
  	  test_expect_code 1 flux terminus start -s $SERVICE &&
  	  test_expect_code 1 flux terminus kill -s $SERVICE 0 &&
  	  test_expect_code 1 flux terminus kill-server -s $SERVICE
  	)
```

suggesting the hang occurs in test 34, `ok 34 - flux-terminus: requests from invalid userid are rejected`

PR #5607, opened 2023-12-05
flux-terminus: fix potential hang in terminus client commands

grondo:
Looking into why `t0020-terminus.t` has been hanging (#5595) revealed a potential race in the flux-terminus server. When the server shuts down, it relies on the disconnect handler to unregister the service with the broker. This leaves a small race between when the server leaves the flux reactor and when the service is unregistered where a client request could be left without a response. This could cause a client hang and thus the hangs in `t0020-terminus.t`.

I did observe the hang several times in the alpine docker container while attempting to debug this, and no hangs after 100s of runs on this branch. So while there's no definitive proof that this issue was the cause of the hang, I'm hopeful this will fix it.

### DIFF
--- a/src/cmd/flux-terminus.c
+++ b/src/cmd/flux-terminus.c
@@ -271,10 +271,20 @@ static void f_logf (void *arg,
     flux_log (h, level, "%s:%d: %s: %s", file, line, func, buf);
 }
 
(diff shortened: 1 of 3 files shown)


garlick (approved):
Seems good to me.

PR #5608, opened 2023-12-06
history: track root jobs

### LINKED ISSUE #5475: jobtap history plugin throws errors
In one of @vsoch's cloudy experiments, in  the output of a failing intel MPI job, we noticed two jobtap errors:

```
jobtap: job.new: callback returned error
jobtap: job.inactive.add: callback returned error
```

The only jobtap in-tree that registers `job.inactive-add` is `history`.  Maybe there is something to run down here or at least a couple places where log messages could be added to give us something to go on next time.

garlick:
Problem: the job history plugin uses an integer userid as hash key, but the hash implementation won't accept a key (NULL) for userid 0.  This results in errors like:
```
jobtap: job.new: callback returned error
jobtap: job.inactive.add: callback returned error
```
    
Map userid 0 to `(uid_t)-1` since that value is reserved in POSIX.
    
Fixes #5475

### DIFF
--- a/src/modules/job-manager/plugins/history.c
+++ b/src/modules/job-manager/plugins/history.c
@@ -37,6 +37,24 @@ struct history {
 
 #define COMPARE_NUM_REVERSE(a,b) ((a)>(b)?-1:(a)<(b)?1:0)
 
+/* Keys in history->users are calculated from int2ptr(uid), but that won't
+ * work for root.  Substitute (uid_t)-1 in that case as it's reserved per
+ * POSIX.  See also: flux-framework/flux-core#5475.
+ */
+static const void *userid2key (int userid)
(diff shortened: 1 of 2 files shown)


src/modules/job-manager/plugins/history.c
```diff
@@ -124,6 +142,7 @@ static int jobtap_cb (flux_plugin_t *p,
     struct history *hist = arg;
     struct job_entry *entry;
     int userid;
+    const char *key;
```
grondo, line 145: should this be `const void *`?

grondo (approved):
LGTM! Just one question inline.

garlick:
I'll go ahead and set MWP here.

PR #5609, opened 2023-12-06
testsuite: fix some test races and improve debugging

grondo:
Many tests in the testsuite seem to have become a bit more unreliable lately. I disabled the `make recheck` step on a branch, at which point 2 or 3 builds failed every time. I then worked through the failing tests, trying to fix some and adding more debug info for others where there was nothing apparent in the logs for the cause of the failure.

This doesn't fix everything, but it should make CI at least slightly more reliable.


garlick (approved):
LGTM!  Thanks for working through all these!  Everyone will suffer a little less.

grondo:
Thanks! Setting MWP.

grondo:
Thanks! I set MWP

PR #5612, opened 2023-12-07
kvs: limit the content of guest commits

garlick:
This limits the operations that can be performed in a guest-owned private KVS namespace.


grondo (approved):
LGTM!

PR #5616, opened 2023-12-08
testsuite: fix t2233-job-info-update.t with debug=t

grondo:
Problem: A quoting issue in one of the tests in t2233-job-info-update.t causes the test to fail when run with -d -v or debug=t.

Fix the quoting issue.

This was caught by gitlab-ci. Not sure why it doesn't surface in our normal CI.


garlick (approved):
LGTM!

PR #5618, opened 2023-12-09
testsuite: fix t2260-job-list.t with `-d -v` and run inception tests with these args to prevent future similar errors

grondo:
Another test had some bad `test_debug` statements that caused the test to fail when run with `-d -v`. This PR fixes that and adds these args to the sharness tests in the inception builder to hopefully catch these silly errors in CI.


garlick:
Was trying to approve this just now but the review button doesn't seem to do anything. I tried two different browsers on two computers.  Will try again later I guess!

grondo:
I was having that problem today too. Reloading the page seemed to help in my case.

garlick (approved):
Oh a reload did work!  Anyway, LGTM!

grondo:
Thanks!

PR #5620, opened 2023-12-12
job-list: fix `flux job list-ids --wait-state`

### LINKED ISSUE #5619: `flux job list-ids --wait-state=inactive` doesn't work
The `flux job list-ids --wait-state` option was ostensibly added to avoid eventual consistency races when tests use the job-list module. Tests can issue `flux job list-ids -W inactive $jobid` and be sure that `$jobid` is known to the job-list module in the INACTIVE state before proceeding with testing.

However, at some point this functionality broke and the result is that all the tests are now susceptible to the race the functionality is intended to prevent. This appears to be one cause of the random failures in the various job-list tests in CI.

A simple test to ensure the `--wait-state` functionality should probably be added. Currently this test fails 100% of the time 
```
$ flux job list-ids -W inactive $(flux submit sleep 5) | jq -e '.state == 64'
false
```
(the command returns immediately)

grondo:
This PR fixes #5619 and with it a class of job-list tests that occasionally fail in ci.

### DIFF
--- a/src/modules/job-list/list.c
+++ b/src/modules/job-list/list.c
@@ -440,7 +440,18 @@ json_t *get_job_by_id (struct job_state_ctx *jsctx,
         return NULL;
     }
 
-    if (job->state == FLUX_JOB_STATE_NEW) {
+    /*  Always return job in inactive state, even if a requested state was
+     *  provided. This avoids no response when a job does not enter a given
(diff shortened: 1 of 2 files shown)


garlick (approved):
LGTM!

grondo:
Great thanks!

PR #5624, opened 2023-12-16
docs: add flux_msg_incref manpage

wihobbs:
A while back when I was working on the notification service, I noticed there was no manual page for the `flux_msg_incref` command.

This is a super simple manpage for that command, and `flux_msg_decref`. Based on what I saw in other manual pages (such as [`flux_msg_handler_addvec`](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man3/flux_msg_handler_addvec.html) I thought it'd be acceptable to include both of them on the same page.

This is super short so feel free to comment on how you want it expanded!


doc/man3/flux_msg_incref.rst
```diff
@@ -0,0 +1,47 @@
...
+DESCRIPTION
+===========
+
+:func:`flux_msg_incref` provides a constant pointer to a `flux_msg_t` object.
+This function prevents `flux_msg_t` objects from being destroyed while response 
+messages (perhaps a `streaming RPC <https://flux-framework.readthedocs.io/
+projects/flux-rfc/en/latest/spec_6.html>`_) are being sent to it.
+
+:func:`flux_msg_decref` allows the message referenced by :func:`flux_msg_incref`
+to be destroyed.
```
garlick, line 28: Cool!  I was just thinking we needed to fill in those missing pages.

Suggestion: add `flux_msg_create()`, `flux_msg_destroy()`, and `flux_msg_copy()` to this page too.  They are all kind of in the constructor/destructor category.

Also I'd keep it simple just say something like
- create creates a message object of the specified type with a reference count of 1.
- incref increments the reference count on a message object
- decref decrements the reference count on a message object and destroy the object if the count reaches 0.
- destroy is an alias for decref

Oh and create should note that type can be FLUX_MSGTYPE_ANY if the type is to be set later with flux_msg_set_type().

To add a page you will also need to update Makefile.am and manpages.py.  Look at another page that contains multiple functions like idset_create(3).  You'll see how the primary page and the stubs are added to both of those files.

wihobbs: > decref decrements the reference count on a message object and destroy the object if the count reaches 0.

Thanks so much for this explanation! I was wondering about the difference between decref and destroy. This is quite clever, because it ensures there can be multiple references and the message won't be destroyed unless they're all freed, cool!

garlick: Actually strike that comment on FLUX_MSGTYPE_ANY.  I'm not sure there's any use for that and it might be something we can clean up.

doc/man3/flux_msg_create.rst
```diff
@@ -0,0 +1,68 @@
...
+
+RETURN VALUE
+============
+
+:func:`flux_msg_create` returns a `flux_msg_t` type.
+
+:func:`flux_msg_copy` returns a new `flux_msg_t` object with the same attributes
+as another `flux_msg_t` objects.
+
+:func:`flux_msg_incref` returns a constant pointer to a `flux_msg_t` object.
```
garlick, line 52: This section doesn't need to explicitly call out the type.  But it should tell you what the return possiblities are.  For example, on failure, create and copy return a Flux message on success, and on failure, return NULL and set errno.  Incref just returns its :var:`msg` argument for convenience.

garlick (changes requested):
A few suggestions inline to make the page more consistent with others.

Also the ci tests are failing because 'make check' in the doc subdir is failing.  You can run that manually in the source tree to run down any spelling errors.  Check the `test/spellcheck.log` file after the `make check` if the check fails.  Fix any misspelled words, or if you need to add words to the dictionary, it's `test/spell.en.pws`.

doc/man3/flux_msg_create.rst
wihobbs: I think I addressed this one, by adding an `ERRORS` section and mentioning the different errors that could be raised by each in the `RETURN VALUE` section. But let me know if it's redundant to have it mentioned in both places, or needs to be done differently.

doc/man3/flux_msg_create.rst
```diff
@@ -0,0 +1,79 @@
...
+:func:`flux_msg_copy` duplicates :var:`msg`. The payload is omitted unless
+:var:`payload` is true. The initial reference count of the new message is 1.
+
+:func:`flux_msg_incref` increments the reference count of :var:`const_msg`
+by 1.
+
+:func:`flux_msg_decref` decrements the reference count of :var:`const_msg`
+by 1. When the reference count reaches 0, the message is destroyed.
+
+:func:`flux_msg_destroy` destroys :var:`msg`.
```
garlick, line 41: `flux_msg_destroy` should just say that it's an alias for :func:`flux_msg_decref`.   They are exactly the same - the message is only destroyed once the reference count reaches zero.

doc/man3/flux_msg_create.rst
```diff
@@ -0,0 +1,79 @@
...
+
+RETURN VALUE
+============
+
+:func:`flux_msg_create` returns a :type:`flux_msg_t` type. On failure,
+NULL is returned and EINVAL or ENOMEM is raised.
```
garlick, line 47: Just say NULL is returned and errno is set, without calling out the errno values (that is for the next section). 

As an aside, C does not have exceptions - `errno` is just a global variable, and its value is undefined (e.g. its value is worthless) except immediately after a function that sets errno fails.  That is why it is important to note here when errno is set on failure.

doc/man3/flux_msg_create.rst
```diff
@@ -0,0 +1,79 @@
...
+
+:func:`flux_msg_copy` returns a new :type:`flux_msg_t` object with the same
+attributes as another :type:`flux_msg_t` objects. On failure, NULL is
+returned and EINVAL is raised because :var:`msg` is invalid.
```
garlick, line 51: This should be trimmed down the same way.  Actually it could just be lumped in with flux_msg_create() in this section.

doc/man3/flux_msg_create.rst
```diff
@@ -0,0 +1,79 @@
...
+
+:func:`flux_msg_incref` returns a constant pointer to :var:`const_msg` for
+convenience. EINVAL is raised if :var:`const_msg` is invalid. 
+
+:func:`flux_msg_decref` has no return value. EINVAL is raised if
+:var:`const_msg` is invalid.
```
garlick, line 57: This function actually doesn't set errno.  Just lump it in with flux_msg_destroy().

doc/man3/flux_msg_create.rst
```diff
@@ -0,0 +1,79 @@
...
+
+:func:`flux_msg_create` returns a :type:`flux_msg_t` type. On failure,
+NULL is returned and EINVAL or ENOMEM is raised.
+
+:func:`flux_msg_copy` returns a new :type:`flux_msg_t` object with the same
+attributes as another :type:`flux_msg_t` objects. On failure, NULL is
+returned and EINVAL is raised because :var:`msg` is invalid.
+
+:func:`flux_msg_incref` returns a constant pointer to :var:`const_msg` for
+convenience. EINVAL is raised if :var:`const_msg` is invalid.
```
garlick, line 54: ditto.

doc/manpages.py
```diff
@@ -165,6 +165,11 @@
     ('man3/flux_log', 'flux_log_set_procid', 'Log messages to the Flux Message Broker', [author], 3),
     ('man3/flux_log', 'flux_log', 'Log messages to the Flux Message Broker', [author], 3),
     ('man3/flux_msg_cmp', 'flux_msg_cmp', 'match a message', [author], 3),
+    ('man3/flux_msg_create', 'flux_msg_create', 'create a Flux message', [author], 3),
+    ('man3/flux_msg_create', 'flux_msg_copy', 'copy a Flux message\'s attributes into a new Flux message', [author], 3),
+    ('man3/flux_msg_create', 'flux_msg_incref', 'get a memory address for a Flux message and increment reference count', [author], 3),
+    ('man3/flux_msg_create', 'flux_msg_decref', 'decrement the reference count of a Flux message by 1', [author], 3),
+    ('man3/flux_msg_create', 'flux_msg_destroy', 'destroy a Flux message when reference count reaches 0', [author], 3),
```
garlick, line 172: I'd go shorter on the descriptions.  If you want to be lazy, use the same description for all the stubs since that's the  convention.   At least try to avoid a line break in the ascii rendering if at all possible.  Check with e.g.

```
cd doc/man3
make -C ..
man ./flux_msg_destroy.3
```

For example just use "destroy a Flux message"  or "take a reference on a Flux message".

doc/test/spell.en.pws
```diff
@@ -823,3 +823,4 @@ aspirational
 infeasible
 login
 workflow
+html
```
garlick, line 826: Maybe leftover from earlier?

garlick (changes requested):
A few more changes are needed I'm afraid.  Sorry to be stickler!

doc/man3/flux_msg_create.rst
wihobbs: Thanks @garlick for the explanation! I think I hit all of your comments and stayed true to what's going on. Sorry you've had to review this so many times...

garlick:
Ooh so close!  The only thing I noticed this time is some parameters were renamed from `msg` to `const_msg`.  That's not needed as the type is already formally defined in the header at the top of the page.

wihobbs:
Sorry for the delay @garlick, I was confused by that correction until @grondo pointed out that my header was actually wrong (because I copied the type declarations from `message.c` and not `message.h`). Hopefully this is all good now!

garlick (approved):
LGTM!  Thanks!

PR #5626, opened 2023-12-17
job-list: fix json object mem-leak

chu11:
Problem: In the check_id_valid path, there is a potential a mem-leak of a json object.  The path is for a rare race, so it likely has not been hit in production.

Solution: Remove the json object when it is no longer used.


grondo (approved):
LGTM!

PR #5627, opened 2023-12-18
python: Add kvs.commit_async()

### LINKED ISSUE #5606: python: support kvs commit_async function
Per discussion in #5389

chu11:
Problem: There may be circumstances that a non-synchronous kvs
commit is desired, but there is no way to do that at the moment.

Solution: Add a kvs.commit_async() function to the kvs module.
Also add equivalent commit_async() functions to KVSTxn and KVSDir.

Fixes #5606

--- 

Note will re-work #5389 to use this


grondo (approved):
LGTM!

PR #5630, opened 2023-12-19
libflux: deprecate some message flag accessors and add new and improved ones

garlick:
This addresses some sloppiness in the low-level message API.

The awkward flags accessors:
```c
/* Get/set flags
 * Users should avoid using flux_msg_set_flags(), and instead use the
 * higher level functions that manipulate message flags.  It is exposed
 * mainly for testing.
 */
int flux_msg_get_flags (const flux_msg_t *msg, uint8_t *flags);
int flux_msg_set_flags (flux_msg_t *msg, uint8_t flags);
```

are replaced with
```c
/* Test/set/clear message flags
 */
bool flux_msg_has_flag (const flux_msg_t *msg, int flag);
int flux_msg_set_flag (flux_msg_t *msg, int flag);
int flux_msg_clear_flag (flux_msg_t *msg, int flag);
```
and some specialized accessors that were little used and no longer justified after the above change are removed, namely
```c
/* Get/set USER1 flag.
 */
int flux_msg_set_user1 (flux_msg_t *msg);
bool flux_msg_is_user1 (const flux_msg_t *msg);
```
Finally, `flux_msg_create()` is modified so that it no longer accepts FLUX_MSGTYPE_ANY as an argument.  That was only used internally and is just confusing.


src/bindings/lua/flux-lua.c
```diff
@@ -265,10 +265,7 @@ static int send_json_request (flux_t *h, uint32_t nodeid, uint32_t matchtag,
     if (flux_msg_set_matchtag (msg, matchtag) < 0)
         goto done;
     if (nodeid == FLUX_NODEID_UPSTREAM) {
-        uint8_t flags;
```
grondo, line 268: commit message typo: `that the previous functions` :arrow_right: `than the previous functions`

chu11 (approved):
Overall this LGTM.  As an aside comment, should we be versioning `libflux`?  Granted we're < version 1.0, but we're also removing functinos.  ehhhh

grondo (approved):
LGTM. It seems like `flux_msg_set_flag(3)` and `flux_msg_has_flag()` actually don't prevent multiple flags? I have no problem with that, I'm just wondering why the rename from plural to singular.

Additionally, if the functions are not suggested for external use, is a man page necessary?
(Or is the purpose here to make these functions officially available for external use?)

grondo:
> As an aside comment, should we be versioning libflux

We do version libflux, but haven't been keeping it up to date at this point:
https://github.com/flux-framework/flux-core/blob/cc0115ed86f5ed2324ebebaea37092d3918ddcbe/configure.ac#L44-L68

chu11:
> We do version libflux, but haven't been keeping it up to date at this point:

Ahh sorry, I knew that.  I was meaning updating the version, b/c removing a function seems like a bigger deal in my head.  But then again, these functions are likely rarely used outside of flux projects.

garlick:
> why the rename from plural to singular.

Er yeah, the primary use was intended to be a single flag but since they effectively take a bitmask, would it be better to repurpose the original plural naming?  I guess I felt like the plural _set_ might not convey the fact that it does a read-modify-write on the flags byte rather than a bulk assignment of a mask like before.  But maybe I feel that way just because I know how the previous function worked and others would not think that?

> is the purpose here to make these functions officially available for external use?

Yes.  For context, I was originally modifying the protocol to allow for "extended flag bits" and wanted to add more "user" flags and possibly a "signed" flag.  It seemed better to improve the existing flag functions than to add specialized accessors for each new flag.

grondo:
That all makes sense to me!

> Ahh sorry, I knew that. I was meaning updating the version, b/c removing a function seems like a bigger deal in my head. But then again, these functions are likely rarely used outside of flux projects. (or even flux-core)

Yeah, you're right. Technically we should modify the version (even when adding interfaces), but I think we've been lazy pre v1.0 :shrug:

garlick:
I'll go ahead and drop the flux_msg_create() change from this PR since it's probably best to at least call out any API changes that we are making independently in the release notes.

garlick:
Did that, and also added flags descriptions to the man page.

grondo:
Updates LGTM!

garlick:
Thanks, I'll set MWP.

PR #5632, opened 2023-12-19
libflux: drop flux_msg_frames()

garlick:
Problem: `flux_msg_frames()` has no practical use outside of message encoding/decoding implementations but it is in the public API.

Convert this to a private function, msg_frames().
Update internal users and unit tests.


chu11 (approved):
lgtm!

PR #5633, opened 2023-12-20
job-info: add JSON_DECODE & CURRENT flags, deprecate job-info.update-lookup

chu11:
This PR could be split up a bit more, but for now it's one PR.

This adds several new lookup flags, and with their addition, the `job-info.update-lookup` RPC target can be deprecated.

The one "tricky" part was how to grab cached data from current `update-watch`es being done.  Perhaps the lookup of that "cached" info could just be dropped as it probably has minimal use.

I decided to not support `job-info.update-watch` within the original `job-info.lookup` RPC for several reasons.  Most notably because `lookup` can take multiple keys as input and `job-info.update-watch` takes only one key.  I could make it so a WATCH flag only works with one key.  Or we could have a "re-direct to this RPC" like we do with KVS watch.  But I didn't delve too much into this design yet.

~~I then updated PR #5592 to be based on this PR.~~  PR #5635 then follows this one up

TODO RFC spec 41 which I was going to hold off on until this gets some minimal review.

### DIFF
--- a/src/cmd/flux-job.c
+++ b/src/cmd/flux-job.c
@@ -3398,7 +3398,7 @@ int cmd_info (optparse_t *p, int argc, char **argv)
                                  "keys", "J",
                                  "flags", 0))
             || flux_rpc_get_unpack (f, "{s:s}", "J", &val) < 0)
-            log_msg_exit ("%s", future_strerror (f, errno ));
+            log_msg_exit ("%s", future_strerror (f, errno));
         if (!(new_val = flux_unwrap_string (val, false, NULL, &error)))
             log_msg_exit ("Failed to unwrap J to get jobspec: %s", error.text);
         val = new_val;
@@ -3423,7 +3423,7 @@ int cmd_info (optparse_t *p, int argc, char **argv)
                                     "{s:s s:s}",
                                     "jobspec", &jobspec,
                                     "eventlog", &eventlog) < 0)
-            log_msg_exit ("%s", future_strerror (f, errno ));
+            log_msg_exit ("%s", future_strerror (f, errno));
         val = new_val = reconstruct_current_jobspec (jobspec, eventlog);
     }
     /* The current (non --base) R is obtained through the
@@ -3440,7 +3440,7 @@ int cmd_info (optparse_t *p, int argc, char **argv)
                                  "key", key,
                                  "flags", 0))
             || flux_rpc_get_unpack (f, "{s:o}", key, &o) < 0)
-            log_msg_exit ("%s", future_strerror (f, errno ));
+            log_msg_exit ("%s", future_strerror (f, errno));
         if (!(new_val = json_dumps (o, JSON_COMPACT)))
             log_msg_exit ("error encoding R object");
         val = new_val;
@@ -3457,7 +3457,7 @@ int cmd_info (optparse_t *p, int argc, char **argv)
                                  "keys", key,
                                  "flags", 0))
             || flux_rpc_get_unpack (f, "{s:s}", key, &val) < 0)
-            log_msg_exit ("%s", future_strerror (f, errno ));
+            log_msg_exit ("%s", future_strerror (f, errno));
     }
 
     printf ("%s\n", val);
--- a/src/cmd/flux-job.c
+++ b/src/cmd/flux-job.c
@@ -3427,23 +3427,19 @@ int cmd_info (optparse_t *p, int argc, char **argv)
         val = new_val = reconstruct_current_jobspec (jobspec, eventlog);
     }
     /* The current (non --base) R is obtained through the
-     * job-info.update-lookup RPC, not the normal job-info.lookup.
+     * job-info.lookup RPC w/ the CURRENT flag.
      */
     else if (!optparse_hasopt (p, "base") && streq (key, "R")) {
-        json_t *o;
         if (!(f = flux_rpc_pack (h,
-                                 "job-info.update-lookup",
+                                 "job-info.lookup",
                                  FLUX_NODEID_ANY,
                                  0,
-                                 "{s:I s:s s:i}",
+                                 "{s:I s:[s] s:i}",
                                  "id", id,
-                                 "key", key,
-                                 "flags", 0))
-            || flux_rpc_get_unpack (f, "{s:o}", key, &o) < 0)
+                                 "keys", key,
+                                 "flags", FLUX_JOB_LOOKUP_CURRENT))
+            || flux_rpc_get_unpack (f, "{s:s}", key, &val) < 0)
             log_msg_exit ("%s", future_strerror (f, errno));
-        if (!(new_val = json_dumps (o, JSON_COMPACT)))
-            log_msg_exit ("error encoding R object");
-        val = new_val;
     }
     /* All other keys are obtained this way.
      */
--- a/src/common/libjob/job.h
+++ b/src/common/libjob/job.h
@@ -35,6 +35,11 @@ enum job_lookup_flags {
      * - currently works for jobspec and R
      */
     FLUX_JOB_LOOKUP_JSON_DECODE = 1,
+    /* get current value of special fields by applying eventlog
+     * updates for those fields
+     * - currently works for R
+     */
+    FLUX_JOB_LOOKUP_CURRENT = 2,
 };
 
 enum job_urgency {
--- a/src/common/libjob/job.h
+++ b/src/common/libjob/job.h
@@ -30,6 +30,13 @@ enum job_event_watch_flags {
     FLUX_JOB_EVENT_WATCH_WAITCREATE = 1, // wait for path to exist
 };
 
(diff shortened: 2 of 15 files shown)


chu11:
hmmm, should I have used the term "UPDATED" vs "UPDATE" as the flag?  UPDATE may imply "to update" vs "UPDATED" is "updates have been applied".

garlick:
Instead of update(d), maybe "current"?

src/modules/job-info/lookup.c
```diff
@@ -288,17 +406,29 @@ static int check_keys_for_eventlog (struct lookup_ctx *l)
...
+
+/* If we need the eventlog for an allow check or for update-lookup
+ * we need to add it to the key lookup list.
+ */
+static int check_to_lookup_eventlog (struct lookup_ctx *l)
+{
+    if (!l->allow || (l->flags & FLUX_JOB_LOOKUP_UPDATE)) {
+        size_t index;
+        json_t *key;
+        json_array_foreach(l->keys, index, key) {
```
garlick, line 398: add space before open paren

src/modules/job-info/lookup.c
```diff
@@ -288,17 +406,29 @@ static int check_keys_for_eventlog (struct lookup_ctx *l)
...
+            const char *keystr;
+            if (!(keystr = json_string_value (key))) {
+                errno = EINVAL;
+                return -1;
+            }
+            if (streq (keystr, "eventlog"))
+                return 0;
         }
```
garlick, line 406: It would probably be good to check that the request `keys` is an array of strings right in the request handler rather than deeper in the code, for clarity and to avoid repeating the same check in several places.  Then this could just be:
```c
if (streq (json_string_value (key), "eventlog"))
    return 0;
```

garlick (commented):
First pass with some comments inline.

Also: we just added `job-info.update-lookup` in 0.56.  Would it be reasonable to just update users and remove or were you thinking that's another PR?

chu11:
re-pushed going with JSON_DECODE and CURRENT as the flag names.

src/common/libjob/job.h
```diff
@@ -30,6 +30,14 @@ enum job_event_watch_flags {
     FLUX_JOB_EVENT_WATCH_WAITCREATE = 1, // wait for path to exist
 };
 
+enum job_lookup_flags {
+    /* decode special fields
+     * - for jobspec and R, value of response will be decoded (i.e. a
+     * json object) instead of a string.
+     */
+    FLUX_JOB_LOOKUP_DECODE = 1,
+};
```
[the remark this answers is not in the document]
chu11: hmmm.  Good point, but it is useful to have around instead of hard coding.  I'm not sure the python bindings can use it if it's a non-public header?

src/modules/job-info/lookup.c
```diff
@@ -156,11 +159,105 @@ static int lookup_keys (struct lookup_ctx *l)
...
+        if (eventlog_entry_parse (entry, NULL, &name, &context) < 0)
+            goto error;
+        if (streq (name, update_name)) {
+            if (streq (key, "R"))
+                apply_updates_R (l, key, update_object, context);
+        }
+    }
+
+    if (!(updated_str = json_dumps (update_object, 0)))
+        goto error;
```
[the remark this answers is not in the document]
chu11: considered this, it is mildly annoying to return the object given the JSON_DECODE logic that later follows in the caller.  I'll add as a fixup commit, lemme know what you think.

chu11:
re-pushed with tweaks per comments above.  There's one fixup patch that adds an update that I'm a little meh on, so I left it as a fixup for the time being so it can be reviewed as a "consideration".

src/common/libjob/job.h
```diff
@@ -35,6 +35,11 @@ enum job_lookup_flags {
      * - currently works for jobspec and R
      */
     FLUX_JOB_LOOKUP_JSON_DECODE = 1,
+    /* get current value of special fields by applying eventlog
+     * updates for those fields
+     * - currently works for jobspec and R
+     */
+    FLUX_JOB_LOOKUP_CURRENT = 2,
```
grondo, line 42: A question about this commit. The comment here indicates "works for jobspec and R" but I only see R handled specifically in `lookup_current`. Maybe a comment in `lookup_current` as to why jobspec is not handled there would be helpful.

src/cmd/flux-job.c
```diff
@@ -3427,23 +3427,19 @@ int cmd_info (optparse_t *p, int argc, char **argv)
...
                                  FLUX_NODEID_ANY,
                                  0,
-                                 "{s:I s:s s:i}",
+                                 "{s:I s:[s] s:i}",
                                  "id", id,
-                                 "key", key,
-                                 "flags", 0))
-            || flux_rpc_get_unpack (f, "{s:o}", key, &o) < 0)
+                                 "keys", key,
+                                 "flags", FLUX_JOB_LOOKUP_CURRENT))
```
grondo, line 3440: Use of this flag requires a current version of flux-core. I'm not sure if it is worth it, but should this call fall back to `update-lookup` if the RPC returns `EPROTO`?

src/modules/job-info/lookup.c
```diff
@@ -289,9 +283,25 @@ static void info_lookup_continuation (flux_future_t *fall, void *arg)
 
         if ((l->flags & FLUX_JOB_LOOKUP_CURRENT)
             && streq (keystr, "R")) {
-            if (lookup_current (l, fall, keystr, s, &current_value) < 0)
+            json_t *current_object = NULL;
+            if (lookup_current (l, fall, keystr, s, &current_object) < 0)
                 goto error;
-            s = current_value;
+            /* tiny optimization, to skip unnecessary json_dumps() and
+             * fallthrough to code below
```
grondo, line 290: I'd be fine dropping this optimization for now and only address performance if it becomes an issue.

grondo (commented):
I realized we should probably have merged this one before I split up `flux-job.c`, so giving this a review now (and sorry this languished so long!).

My one thought is that unfortunately the `job-info.update-lookup` RPC made it into released versions of flux-core, so we have to be careful about switching `flux job` and other users to the new flags in case a newer flux-core is used with an older. For `flux-job` however, I don't know how likely it is that there would be an issue so maybe it is fine?

src/common/libjob/job.h
chu11: ahh comment is probably a development history mistake, see #5635

src/cmd/flux-job.c
chu11: ugh, It ends up I didn't check "flags" in older versions b/c flags wasn't used.  So I guess a caller would just get the original version.

Could just keep on calling `job-info.update-lookup`?  But I'm not sure the pros of that outweigh the cons.  I figure most users, if they are using `flux job info`, base R == updated R anyways?

grondo: >  I figure most users, if they are using flux job info, base R == updated R anyways?

Yes, there will be an updated R only if the expiration is adjusted, which is going to pretty rare. Also, the command would have to be run against a downrev parent instance to hit the failure, and at least this would be a failure I think instead of returning wrong information. So perhaps we just keep things as is and hope for the best? Over time it won't be a problem anymore... (However, something to keep in mind when adding future interfaces)

grondo (approved):
Given the discussion about the `flux job info` change, this can probably go in as is when the fixup commit is dropped.

chu11:
just re-pushed.  apologies, i began cleaning up more thinking that the split of `flux-job.c` had already been merged into master, whoops.  I hope I didn't clean up too much and make the eventual rebasing of PR #5747 more difficult than it needs to be.

- removed the fixup
- tweaked that one comment per @grondo comment above
- noticed some stupid typos ... https://github.com/flux-framework/flux-core/pull/5633/commits/ad606a818928c7c0a90613c1456db7b5d706095e  
- given realization above that `job-info.lookup` doesn't check for valid flags, added that check in https://github.com/flux-framework/flux-core/pull/5633/commits/cbe957de17b3503f486ed24055ed63ae30d772c3 (and added appropriate updates when new flags added)
- realized needed some extra coverage https://github.com/flux-framework/flux-core/pull/5633/commits/a43f24f97ad307a0cc9bc8d0aa24b140ab7fce2c given issues above

this PR is getting long, could split it up?

grondo:
Since splitting up `flux-job.c` just copy and pastes code it should be easy to fixup after all this is merged (just recopy the relevant code I hope)

If you want to split this into a separate PR that is fine with me. Thinking about each PR as a "release note entry" though, what would that be?

chu11:
> If you want to split this into a separate PR that is fine with me. Thinking about each PR as a "release note entry" though, what would that be?

It'd just be a "cleanup" PR, so I guess nothing useful.  Only offering to split up if it would make reviewing easier ... but I now see that you approved the PR pre-my recent changes.  So I guess that doesn't matter as much now :-)

PR #5634, opened 2023-12-20
doc: expand glossary

garlick:
Problem: the Glossary section of the flux-core docs is pretty thin.

Define some Flux/resource management terms.

(dredged this up from earlier docs work and made an editing pass)

### DIFF
--- a/doc/guide/glossary.rst
+++ b/doc/guide/glossary.rst
@@ -4,13 +4,109 @@ Glossary
 Here we define Flux-specific and general HPC and workload management terms
 used in our documentation that may not be familiar to all readers.
 
-  .. glossary::
+.. glossary::
 
-     initial program
-        A user-defined program, such as a batch script, launched on the first
-	node of a Flux instance.  Its purpose is to launch and monitor a
-	workload.  Once it is complete, the instance exits.
+  enclosing instance
+    The Flux instance that a process naturally interacts with.  It is
+    the instance referred to by the :envvar:`FLUX_URI` environment variable,
+    or if that is not set, it is the :term:`system instance`.
 
-     system instance
-        A multi-user Flux instance running as the primary resource manager
-	on a cluster.
+  expedited
+    A job is said to be expedited if its :term:`urgency` is set to the
+    maximum value of 31.  An expedited job's :term:`priority` is always set
+    to the maximum value.
+
+  FSD
+    A common string representation of time duration, defined by
+    :doc:`rfc:spec_23`.  Example: ``2.5h``.
+
+  guest
+    A Flux user that is not the :term:`instance owner`.  Guests are only
+    allowed to run in a Flux instance configured for multi-user support,
+    normally a :term:`system instance`.
+
+  held
+    A job is said to be held if its :term:`urgency` is set to zero.  This
+    prevents it from being considered for scheduling until the urgency is
+    raised.
+
+  hostlist
+    A compact string representation of a list of hostnames, defined by
+    :doc:`rfc:spec_29`.  Example: ``fluke[0-127,130]``.
+
+  idset
+    A compact string representation of a set of non-negative integers,
+    defined by :doc:`rfc:spec_22`.  Example: ``2,4,6,1-100``.
+
+  IMP
+    The Independent Minister of Privilege.  The simple setuid root component
+    of Flux, from the flux-security project, that allows an
+    :term:`instance owner` to perform a limited set of tasks on behalf of a
+    :term:`guest` user in a multi-user Flux instance.
+
+  initial program
+    A user-defined program, such as a batch script, launched on the first
+    node of a Flux instance.  Its purpose is to launch and monitor a
+    workload.  Once it is complete, the instance exits.
+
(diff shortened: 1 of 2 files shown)


doc/guide/glossary.rst
```diff
@@ -4,13 +4,116 @@ Glossary
...
 
-  .. glossary::
+.. glossary::
 
-     initial program
-        A user-defined program, such as a batch script, launched on the first
-	node of a Flux instance.  Its purpose is to launch and monitor a
-	workload.  Once it is complete, the instance exits.
+  enclosing instance
+    The Flux instance that launched the current Flux instance as a job.
```
chu11, line 10: Took me a second to grok, perhaps "The parent Flux instance that launched ..."?

doc/guide/glossary.rst
```diff
@@ -4,13 +4,116 @@ Glossary
...
+
+  expedited
+    A job is said to be expedited if its :term:`urgency` is set to 31.
```
chu11, line 13: perhaps "the maximum urgency of 31"?

doc/guide/glossary.rst
```diff
@@ -4,13 +4,116 @@ Glossary
...
+  IMP
+    The Independent Minister of Privilege.  The simple setuid root component
+    of Flux, from the flux-security project, that allows an
+    :term:`instance owner` to perform a limited set of tasks on behalf of a
+    :term:`guest` user in a multi-user Flux instance.
+
+  initial program
+    A user-defined program, such as a batch script, launched on the first
+    node of a Flux instance.  Its purpose is to launch and monitor a
+    workload.  Once it is complete, the instance exits.
```
chu11, line 47: i know not your addition, but "purpose is to launch and monitor a workload" doesn't seem right?  "Typically it is used to launch ..."?

doc/guide/glossary.rst
```diff
@@ -4,13 +4,116 @@ Glossary
...
+    The JSON object used by Flux to represent a concrete resource set.
+    See :doc:`rfc:spec_20`.
+
+  resource inventory
+    The concrete set of resources managed by a given Flux instance.
+
+  scheduler
+    The Flux component that fulfills resource allocation requests from the
+    :term:`resource inventory`.  Abstract resource requirements are extracted
+    from user-provided :term:`jobspec`, and fulfilled with a resource set
```
chu11, line 88: "from THE user-provied"?

doc/guide/glossary.rst
grondo: I think we've also used enclosing instance to refer to the nearest instance under which a process is running, e.g. that to which `FLUX_URI` refers, or the system instance if the system is running Flux as the native RM. Am I mistaken?

Applying the term only to instances run as a job under Flux may reduce any ambiguity, but IMO it is too convenient to use _enclosing instance_ also to refer generally to the current instance under which a process is running.

If we do want to restrict the usage of the term, there may be a few spots in existing documentation that need to be tightened up.

doc/guide/glossary.rst
grondo: Hm, "Its purpose is to launch and monitor a workload" works for me FWIW. (For me anyway, stating the purpose of the initial program gives more clarity than saying something like "Typically, ..." which implies ambiguity of what the initial program does) However, I don't feel strongly here and either could work.

chu11: I guess in my mind someone can launch "anything" as the initial program, like a shell, but most people probably wouldn't do that.

doc/guide/glossary.rst
chu11: > I think we've also used enclosing instance to refer to the nearest instance under which a process is running

Now that @grondo mentions it, I have this feeling as well.

doc/guide/glossary.rst
grondo: Yeah, so the more general statement of the initial program's purpose might work better? (E.g. `flux mini alloc` launches an interactive shell as the initial program, so a shell is actually common)

doc/guide/glossary.rst
garlick: Sorry I got off track on that one!  How's this?

> The Flux instance that a process naturally interacts with. It is the instance referred to by the FLUX_URI environment variable, or if that is not set, it is the system instance.

garlick:
I pushed the suggested changes so far.

Also, I wanted to note that my original plan was to just expand this glossary as terms were used elsewhere in the flux-core docs, to avoid needing to re-explain, not necessarily to build up a glossary page that people might just read on their own.  For that we have 
[the glossary page in flux-docs](https://flux-framework.readthedocs.io/en/latest/glossary.html) I guess.  However, this was sitting around for a long time with other unfinished work, and I thought maybe it would be good to review separately.    If that's a misstep I can keep it archived for later.

garlick:
oops noticed a formatting error and also realized I have the definitions of "moldable" and "malleable" swapped.  I'll just drop those for now and add them if/when they get used in the docs.

grondo (approved):
Very nice!

Issue #5611, opened 2023-12-07
resource: fetch local hwloc topology from enclosing instance


grondo:
The job shell fetches hwloc topology XML from the local broker's resource module via the `resource.topo-get` RPC instead of loading the topology XML via `hwloc_topology_load()`. This helps with job throughput since `hwloc_topology_load()` can be very slow, especially when many processes are performing it at the same time, but it also allows the job shell to "see" the same topology as the broker in the case fake resources are loaded.

However, if the job being launched is a new instance of Flux, the resource module in the new instance always calls `hwloc_topology_load()` to set the topology XML, thus breaking the chain of consistent topology and possibly slowing down startup of the new instance. Perhaps the resource module could instead fetch the topology XML from the parent instance? I think this should work, but haven't given it any detailed thought.

This would allow `FLUX_HWLOC_XMLFILE` to be used to load a fake topology in a root instance, and then have this topology be propagated through an entire Flux instance hierarchy of any depth.

grondo:
I have a working prototype of this feature. Just to share some timing results, on tioga this change saves ~1.3s during startup (probably a worst case scenario though) (note that in the prototype `resource.local-topo=true` forces the resource module to fall back to `hwloc_topology_load(3)` instead of `resource.topo-get`:

```
f(s=4,d=0,builddir) grondo@tioga11:~/git/flux-core.git$  flux alloc -n1 --conf=resource.local-topo=true "flux dmesg -HL | grep resource"
[  +0.348501] broker[0]: insmod resource
[  +1.608233] resource[0]: reslog_cb: resource-init event posted
[  +1.608946] resource[0]: reslog_cb: resource-define event posted
[  +2.037605] sched-simple[0]: resource update: {"resources":{"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0"}}],"starttime":1702508599.621218,"expiration":0.0,"nodelist":["tioga11"]}},"up":""}
[  +2.274156] resource[0]: reslog_cb: online event posted
[  +2.274194] sched-simple[0]: resource update: {"up":"0"}
[detached: session exiting]
f(s=4,d=0,builddir) grondo@tioga11:~/git/flux-core.git$  flux alloc -n1 "flux dmesg -HL | grep resource"
[  +0.346978] broker[0]: insmod resource
[  +0.347979] resource[0]: retrieved local hwloc XML from parent
[  +0.352608] resource[0]: reslog_cb: resource-init event posted
[...]

PR #5636, opened 2023-12-20
resource: get local hwloc XML from parent when possible

grondo:
This PR adds the ability for the resource module to grab the local hwloc from its parent instance when possible, instead of always using `hwloc_topology_load(3)`. This should speed up this stage of the module's initialization greatly, as well as keeping a consistent topology available throughout a hierarchy as noted in #5611.

If there is no parent instance or the `resource.topo-get` RPC fails , then we fall back to `hwloc_topology_load(3)`.

This is a WIP for now to get feedback and until I figure out how to add some testing.

### DIFF
--- a/src/modules/resource/topo.c
+++ b/src/modules/resource/topo.c
@@ -275,19 +275,61 @@ void topo_destroy (struct topo *topo)
     }
 }
 
+static char *topo_get_local_xml (struct resource_ctx *ctx, bool no_restrict)
+{
+    flux_t *parent_h;
+    flux_future_t *f = NULL;
+    char *result = NULL;
+    const char *xml;
+
+    errno = 0;
+    if (!(parent_h = resource_parent_handle_open (ctx))
+        || !(f = flux_rpc (parent_h,
+                           "resource.topo-get",
+                           NULL,
+                           FLUX_NODEID_ANY,
+                           0))
+        || flux_rpc_get (f, &xml) < 0) {
+        rhwloc_flags_t flags = no_restrict ? RHWLOC_NO_RESTRICT : 0;
+        /*  ENOENT just means there is no parent instance.
+         *  No need for an error.
+         */
+        if (errno && errno != ENOENT)
+            flux_log (ctx->h,
+                      LOG_DEBUG,
+                      "resource.topo-get to parent failed: %s",
+                      strerror (errno));
+        result = rhwloc_local_topology_xml (flags);
+        goto out;
+    }
+    flux_log (ctx->h,
+              LOG_INFO,
+              "retrieved local hwloc XML from parent");
+    if (no_restrict) {
+        result = strdup (xml);
+        goto out;
+    }
+    /*  restrict topology to current CPU binding
+     */
+    result = rhwloc_topology_xml_restrict (xml);
+out:
+    flux_future_destroy (f);
+    resource_parent_handle_close (ctx);
+    return result;
+}
+
 struct topo *topo_create (struct resource_ctx *ctx,
                           bool no_verify,
                           bool no_restrict)
 {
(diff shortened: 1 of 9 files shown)


garlick:
Just curious - what is the situation that led to the 100ms timeout?

grondo:
>  Just curious - what is the situation that led to the 100ms timeout?

Uh, no actual situation. However, the purpose of the feature is to speed up startup of the resource module so it seemed like some timeout was warranted.  However, in practice the timeout is not likely to ever be triggered, so I'd be fine removing it (however, it didn't seem like we'd want a hard hang here since it is synchronous code). What do you think?

garlick:
:shrug:  Personally I'd probably leave out if a hang is not expected.  If you leave it in please add a block comment that says what you said above to make the purpose clear.

grondo:
Ok removed that. If the RPC to the parent ever doesn't get a response, the child instance will appear to hang. This can be diagnosed with a debugger. However, I can't think of how that would happen in practice unless the parent resource module is unresponsive.

grondo:
Just to post some results of resource module initialization, here's the difference on tioga 
hwloc topology load:
```
[  +0.366120] broker[0]: insmod resource
[  +1.522644] resource[0]: reslog_cb: resource-init event posted
```
XML  from parent:
```
[  +0.363129] broker[0]: insmod resource
[  +0.364077] resource[0]: retrieved local hwloc XML from parent
[  +0.369936] resource[0]: reslog_cb: resource-init event posted
```

garlick:
Idle thought: should we disable resource verification when we are getting XML from the parent?  In fact should we skip it when the instance level is > 0 regardless, since some (grand)parent parent would have already checked?

grondo:
> Idle thought: should we disable resource verification when we are getting XML from the parent?

Hm, I'm trying to recall the purpose of the verification. In a way this makes sense to me, then verification only occurs at the root of an instance hierarchy, when hwloc topology is _actually_ loaded from the system.

> n fact should we skip it when the instance level is > 0 regardless, since some (grand)parent parent would have already checked?

My brain may not be working, but I think this would be the effect if we skipped when XML is available from the parent, since only (and all?) instances at level > 0 will have a parent from which to fetch the XML.

One small point is that while  the parent may have already verified resources, the current instance will further restrict the topology based on the current available resources, so the sets being compared are not exactly the same. (However, I'm not sure we're getting much by verifying resources in an instance-as-a-job).

Shall I experimentally add this to the current PR?

garlick:
> Shall I experimentally add this to the current PR?

Maybe we ought to save that for another day when we're feeling more on top of things, 'cause I'm not really sure about this one either.

grondo:
I'm not sure there's all that much extra testing to add here. The `hwloc_topology_load(3)` fallback is well tested by every `flux start` test instance in the testsuite, and since the resource copy of hwloc XML is used for every job (including subinstances now), that path is also naturally well tested by the testsuite.

I did notice that, while not really all that useful,  `resource.norestrict` handling in a subinstance was not tested, so I added a test for good measure.

This could use a review now.

src/modules/resource/topo.c
```diff
@@ -304,7 +304,8 @@ static char *topo_get_local_xml (struct resource_ctx *ctx, bool no_restrict)
     }
```
garlick, line 304: "resource" is misspelled in the commit body for `testsuite: add test with resource.norestrict in subinstance`

garlick (approved):
LGTM - just a minor typo.

src/modules/resource/topo.c
grondo: Thanks! Fixed that and force-pushed. Will set MWP.

PR #5638, opened 2023-12-21
libflux: add include for ssize_t to message.h

ilumsden:
While debugging an issue in DYAD, @hariharan-devarajan and I found something odd stemming from `message.h` in `libflux`. When we used `flux_log` or `flux_log_error` in DYAD, everything worked fine. However, when we commented out the code using those functions, DYAD failed to build because `ssize_t` wasn't defined when the compiler was handling code from `message.h`.

We can work around this in DYAD by making sure we include `sys/types.h` (i.e., the header where `ssize_t` is defined) before including `flux/core.h`, but it would be better long term if that include happened in `message.h`. This PR simply adds that include.

### DIFF
--- a/src/common/libflux/message.h
+++ b/src/common/libflux/message.h
@@ -16,6 +16,7 @@
 #include <stdarg.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <sys/types.h>
 #include "types.h"
 
 #ifdef __cplusplus


garlick (dismissed):
Thanks!  That should definitely be there.
Please set the merge-when-passing label and this should be merged once the tests pass.

garlick:
Oh the commit validator is failing because there is no commit body.  We often use the form:
```
Problem: message.h uses ssize_t without including the proper header that defines it.

Include <sys/types.h>.
```

Feel free to fix that and force push.

grondo:
I changed the PR title to better work in our release notes (use imperative, label with `libflux:` like your commit message)

The commit here is being flagged by our commit validator because it doesn't have a body. I know in this case it seems unnecessary, but In most Flux projects we try to always include a commit body and use the `Problem:` format, e.g. in this case it would be as simple as (but feel free to use the text you'd like)

> Problem: libflux message.h uses ssize_t type without including the correct header for it.
> 
> Add sys/types.h to message.h.

ilumsden:
Just updated the commit message. Also, for the label, GitHub is, for some reason, not letting me add labels to this PR.

grondo:
Ah, I guess github only allows maintainers to add labels. I've added it.

Also, I realized our suggestion wasn't clear. The commit subject should still be `libflux: add include..`, the `Problem: ` statement etc go in the commit body.

ilumsden:
That makes sense. I just force pushed a commit with the revised message

grondo (approved):

jameshcorbett:
I think if you rebase with master and force-push the PR will be automatically merged @ilumsden

ilumsden:
I rebased on the current tip of `master`. I also noticed that one of the Mergify CI routines was failing because I've never linked my GitHub account with Mergify before. I've taken care of that too, so, hopefully, things will work now.

grondo (approved):

PR #5639, opened 2023-12-22
resource: fix initialization of multiple brokers per node when fallback to job-info.lookup is used

grondo:
I noticed while running some tests on corona that launching multiple brokers per node wasn't working.

The issue was introduced by 093509813cfe7fbafc4c84614069be82b17ea703. In the fallback code, `inventory_put()` is always attempted, even if the conversion of R failed and thus `R == NULL`. This causes `inventory_put()` to fail and in turn the load of the `resource` module.

The call to `inventory_put()` should only be made if `R != NULL` (as it is in the non-fallback case).

Also, add some tests that were clearly missing here (though this one's a bit of a corner case)

### DIFF
--- a/src/modules/resource/inventory.c
+++ b/src/modules/resource/inventory.c
@@ -498,7 +498,13 @@ static int lookup_R_fallback (struct inventory *inv, flux_jobid_t id)
         errno = EINVAL;
         goto done;
     }
-    rc = inventory_put (inv, R, "job-info");
+    /*  Only call inventory_put() if conversion of R was successful,
+     *  i.e. R != NULL. (if conversion failed, fall-through to dynamic
+     *  discovery will call inventory_put() later).
+     */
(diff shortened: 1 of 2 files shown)


chu11 (approved):
LGTM

grondo:
Thanks! and thanks for fixing the title typo!

grondo:
One of the new tests fails in the inception builder because of this:
```console
$ flux run -n1 flux start -s4 flux run -xN2 -n4 flux start flux resource info
0.035s: job.exception type=alloc severity=0 alloc denied due to type="unsatisfiable"
Dec 22 12:54:27.382070 broker.err[0]: rc2.0: flux run -xN2 -n4 flux start flux resource info Exited (rc=1) 0.5s
flux-start: 0 (pid 849100) exited with rc=1
```

(not a new issue). Debugging this now.

PR #5641, opened 2023-12-28
testsuite: avoid artificial corefile in test

grondo:
Problem: In t3306-system-routercrash.t SIGSEGV is intentionally sent to a broker to test a broker crash, but this could leave a mysterious corefile in the test directory if the core file ulimit and system configuration allow it.

Call `ulimit -c 0` before test_under_flux() to avoid the core file.

### DIFF
--- a/t/t3306-system-routercrash.t
+++ b/t/t3306-system-routercrash.t
@@ -15,6 +15,11 @@ This test verifies that the invariant holds when the router crashes.
 
 export TEST_UNDER_FLUX_TOPO=kary:1
 
+# A test below intentionally sends SIGSEGV to a broker process --
+# avoid an unnecessary corefile by ensuring the corefile limit is 0
+# before calling test_under_flux():
+ulimit -c 0
+
 test_under_flux 3 system
 
 startctl="flux python ${SHARNESS_TEST_SRCDIR}/scripts/startctl.py"


chu11 (approved):
LGTM!

grondo:
Thanks!

wihobbs:
I guess in order to do ^ you'd have to save something like `COREFILE_SIZE=$(ulimit -c)` before changing `ulimit -c 0`, then restore at the end of the test with `ulimit -c $COREFILE_SIZE`. @grondo is that how you would go about this?

grondo:
> Based on my understanding of ulimit, these settings will persist after this test, and we use ulimit -c unlimited in shell scripts before the test suite runs to try and capture core files occasionally in GitLab (we tried this with the libterminus tests for a while).

`ulimit(1)` and the [setrlimit(2)](https://linux.die.net/man/2/setrlimit) system call it uses only affect the current process and its descendants (see the _NOTES_ section in the `setrlimit(2)` man page above).

You can test this for yourself by starting a new shell, adjusting the resource limit, then exiting and see the parent shell's resource limit is unchanged:
```console
$ ulimit -c
16
$ bash 
$ ulimit -c unlimited
$ ulimit -c
unlimited
$ exit
exit
$ ulimit -c
16
```
Since `ulimit -c unlimited` is called within a sharness test, it will only persist for the tests within that one test script.

grondo:
Though I should mention that all tests within a single sharness test _do_ run in the same shell and what you did above would work I think if you wanted to adjust the `ulimit` for a single `test_expect_success` for example. (Alternately, you could run the test in a subshell)

PR #5645, opened 2024-01-02
flux-batch: support `--quiet` option

### LINKED ISSUE #5644: Consistent support for `--quiet` for submit and batch
heyo! I was wondering if we might have consistent support for `--quiet` for each of `flux submit` and `flux batch`, which both can spit out ids. `flux run` wouldn't make sense I don't think (but maybe there is a use case I'm not thing of). Here is some quick testing with the latest jammy image:

image

grondo:
This PR enables the `--quiet` option on `flux-batch`.

Fixes #5644


chu11 (approved):
LGTM

grondo:
Thanks! I've set mwp.

PR #5646, opened 2024-01-02
broker: improve handling of interactive initial programs

### LINKED ISSUE #5623: flux-job: `client.c:432: pty_read_cb: read: Input/output error` when nesting jobs with interactive ptys
Reproducer:
```console
grondo@pi3:~$ flux alloc -N1 flux alloc -N1
grondo@pi3:~$ flux-job: client.c:432: pty_read_cb: read: Input/output error
      flux-job: client.c:334: pty_client_exit: rpc unpack: Invalid argument
```
The prompt of the depth=2 instance displays, but once there is any input on the terminal the above errors display and the terminal hangs.

grondo:
This PR implements one solution for handling interactive initial programs that are not detected as such by the broker.

Currently, the broker puts all `runat` commands into a background process group unless they are detected as an 'interactive shell', i.e. a shell with no other arguments. This leaves other interactive workloads like the launch of an editor or even `flux alloc` unsupported. These programs are either killed by the broker after being stopped via `SIGTTIN`/`SIGTTOU`, or they get `EIO` errors.

This PR makes the broker behave more like a shell and move the "background" process group to the foreground via [tcsetpgrp(3)](https://linux.die.net/man/3/tcsetpgrp) when the processes have stopped and the reason is likely one of the terminal stop signals. This also allows _all_ `runat` processes to be run in their own process group and should facilitate cleanup somewhat.

The alternative solution here would be to never place runat processes into their own (background) process group. This would be the much simpler solution, but would cause more processes to escape from being signaled in `runat_abort()` and would likely result in more descendant processes being left around after an instance shutdown. However, if the current solution is not acceptable, I'd be fine going that way instead.

Fixes #5623


garlick:
The way you described it, this sounds like exactly the right solution to me.  Thank you for figuring it out!

t/Makefile.am
```diff
@@ -350,7 +350,6 @@ dist_check_SCRIPTS = \
 	issues/t4583-free-range-test.sh \
 	issues/t4612-eventlog-overwrite-crash.sh \
 	issues/t4711-job-list-purge-inactive.sh \
-	issues/t4771-flux-start-bash.sh \
```
garlick, line 353: commit message: there's an extra digit in the issue number

garlick (approved):
Hmm, Emacs as a workflow orchestrator?  Need those emacs lisp flux bindings next!

Awesome.  Looks good and a quick test of `flux alloc vi` works!

grondo:
Thanks! I fixed that typo in the commit message and will set MWP shortly.

PR #5648, opened 2024-01-03
doc: add missing standard-io options to flux-batch(1)

grondo:
Problem: Documentation of standard io options is not included in flux-batch(1).

Add a STANDARD I/O section to flux-batch(1) and include the missing common/job-standard-io.rst text. Since flux-batch(1) now has its own manual, remove the generic statement about flux-batch(1) from the common file and put a similar statement directly in flux-batch(1).


garlick (approved):
Good catch!  I likely created that problem - sorry about that.

grondo:
No problem! Setting MWP

PR #5650, opened 2024-01-04
python: add rolemask parameter to msg_handler_create()

### LINKED ISSUE #5013: need a way to add to rolemask for msg_handlers from python
It doesn't appear there is currently a good way to adjust the rolemask of a message watcher/handler from Python.

This might work for now, but we should have a more Pythonic way to do this:
```python
watcher = self.handle.msg_watcher_create(
    view_user, FLUX_MSGTYPE_REQUEST, "accounting.view-user", self
)
self.handle.msg_handler_allow_rolemask(watcher.handle, FLUX_ROLE_ALL)
watcher.start()
```

grondo:
This PR fixes #5013 by adding an optional `rolemask` parameter to the `MessageWatcher` class initializer as well as `msg_handler_create()` which calls it.

### DIFF
--- a/src/bindings/python/flux/core/handle.py
+++ b/src/bindings/python/flux/core/handle.py
@@ -247,8 +247,11 @@ class Flux(Wrapper):
         topic_glob="*",
         args=None,
         match_tag=raw.FLUX_MATCHTAG_NONE,
+        rolemask=None,
     ):
-        return MessageWatcher(self, type_mask, callback, topic_glob, match_tag, args)
+        return MessageWatcher(
+            self, type_mask, callback, topic_glob, match_tag, rolemask, args
+        )
 
     def timer_watcher_create(self, after, callback, repeat=0.0, args=None):
         return TimerWatcher(self, after, callback, repeat=repeat, args=args)
--- a/t/python/t1000-service-add-remove.py
+++ b/t/python/t1000-service-add-remove.py
@@ -11,6 +11,7 @@
 ###############################################################
 
 import errno
+import os
 import sys
 import unittest
 
@@ -33,6 +34,10 @@ class TestServiceAddRemove(unittest.TestCase):
         rc = self.f.service_register("foo").get()
         self.assertEqual(rc, None)
 
+    def test_001_register_owner_service(self):
+        rc = self.f.service_register("owner").get()
+        self.assertEqual(rc, None)
+
     def test_002_service_add_eexist(self):
         with self.assertRaises(EnvironmentError) as e:
             self.f.service_register("foo").get()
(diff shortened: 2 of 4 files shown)


src/bindings/python/flux/core/handle.py
```diff
@@ -247,8 +247,11 @@ def msg_watcher_create(
         topic_glob="*",
         args=None,
         match_tag=raw.FLUX_MATCHTAG_NONE,
+        rolemask=None,
```
chu11, line 250: would it be worthwhile to set this to FLUX_ROLE_OWNER?  so it's obvious what the default is?  It wouldn't match the C code, but C is different than Python.  ... ehhh

t/python/t1000-service-add-remove.py
```diff
@@ -75,6 +79,34 @@ def cb(f, t, msg, arg):
...
+        self.assertTrue(m is not None)
+        ret = self.f.send(m)
+        self.assertEqual(ret, 0)
+
+        ret = self.f.reactor_run()
+        self.assertTrue(ret >= 0)
+        self.assertTrue(cb_called[0])
+        w2.stop()
+        w2.destroy()
+
```
chu11, line 109: should test opposite case?  handler only allows owner and user is rejected?

grondo: Great point. I'll add that

src/bindings/python/flux/core/handle.py
grondo: It seemed better to have the default behavior preserved rather than forcing a flag here 🤷

t/python/t1000-service-add-remove.py
grondo: Ok, that was a bit of work and uncovered some other issues. This PR is now based on top of #5653 since it requires the new `decode` method on the `Message` class.

grondo:
The extra testing was added here, so probably ready for another review.

garlick (approved):
LGTM!

t/python/t1000-service-add-remove.py
```diff
@@ -44,11 +49,27 @@ def service_cb(f, t, msg, arg):
...
+    def test_003_add_owner_message_watcher(self):
+        def service_cb(f, t, msg, arg):
+            rc = f.respond(msg, msg.payload_str)
+            self.assertEqual(rc, 0)
+
+        self.f.watcher2 = self.f.msg_watcher_create(
+            service_cb,
+            FLUX_MSGTYPE_REQUEST,
+            "owner.echo",
+        )
```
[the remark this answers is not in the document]
grondo: Good suggestion. I'll add  that comment.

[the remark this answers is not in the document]
grondo: I added this comment.
```
       # Note: default rolemask for handle.msg_watcher_create()
       # should allow instance-owner only:
```

PR #5652, opened 2024-01-04
libflux: respond to denied, matchtagless requests

### LINKED ISSUE #5651: libflux: unauthorized requests are not responded to unless they have a matchtag
Problem: @grondo happened to notice this code that predates the addition of the FLUX_MSGFLAG_NORESPONSE flag:

https://github.com/flux-framework/flux-core/blob/6d1f53e613e82bb000340bafd849d8e5b2b4f1b7/src/common/libflux/msg_handler.c#L332-L343

When deciding whether or not to generate a response, it should check the new flag, not for FLUX_MATCHTAG_NONE since on occasion we do send requests that expect a response but don't set a matchtag.  For example: the sched alloc request does this.  @grondo ran across it while adding a python test that built the request from scratch.

garlick:
Problem: the message dispatcher does not respond with EPERM to unauthorized requests (per the message handler's rolemask) unless the message has a valid matchtag, but that should not be required.

Only suppress the response if FLUX_MSGFLAG_NORESPONSE is set.

FIxes #5651

### DIFF
--- a/src/common/libflux/msg_handler.c
+++ b/src/common/libflux/msg_handler.c
@@ -325,14 +325,13 @@ static int copy_match (struct flux_match *dst,
 
 static void call_handler (flux_msg_handler_t *mh, const flux_msg_t *msg)
 {
-    uint32_t rolemask, matchtag;
+    uint32_t rolemask;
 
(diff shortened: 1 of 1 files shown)


grondo (approved):
Looks great, thanks!

PR #5653, opened 2024-01-04
python: add `decode` method to Flux `Message` class

grondo:
Python `Message` objects currently offer some basic property getter and setters, but there is no convenient way to call the underlying `decode` API function for these messages. Notably, this makes it impossible to tell if a response contains an error.

This PR adds a `decode` method to the `Message` class which calls the underlying decode C API, raising an `OSError` exception if the decode fails, since this is an idiomatic way to detect an error in the case of requests.

Additionally, there were no unit tests of the `Message` class, so a set of tests is added.

### DIFF
--- a/src/bindings/python/flux/message.py
+++ b/src/bindings/python/flux/message.py
@@ -109,6 +109,37 @@ class Message(WrapperPimpl):
     def type_str(self):
         return msg_typestr(self.type)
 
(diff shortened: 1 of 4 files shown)


garlick (approved):
I didn't spot anything wrong here.  LGTM!

Issue #5367, opened 2023-08-03
flux-jobs: add "advanced filter" option to use with constraints


chu11:
per my list in #5126, something like `flux jobs --query="userid:1234 name:foo states:pending`

grondo:
Could we reuse `--filter` instead of adding a new option? `states` could be the default operator in the constraint parser. Actually haven't looked into it all, so may be a dumb suggestion, but I thought I'd point it out.

PR #5656, opened 2024-01-05
job-list: support "hostlist" constraint to allow jobs to be filtered by nodes

chu11:
TSIA.

Maybe the only interesting side note is that until we solve #5367 (which should probably follow this one up) we don't yet have a friendly user interface for this.  The tests send constraints objects straight to the `job-list` module.

### DIFF
--- a/src/modules/job-list/match.c
+++ b/src/modules/job-list/match.c
@@ -51,6 +51,8 @@ typedef enum {
     MATCH_LESS_THAN = 4,
 } match_comparison_t;
 
+#define MIN_MATCH_HOSTLIST 1024
+
 struct timestamp_value {
     double t_value;
     match_timestamp_type_t t_type;
@@ -428,6 +430,91 @@ static struct list_constraint *create_results_constraint (struct match_ctx *mctx
                                       errp);
 }
 
+static int match_hostlist (struct list_constraint *c,
+                           const struct job *job,
+                           unsigned int *comparisons,
+                           flux_error_t *errp)
+{
+    struct hostlist *hl = zlistx_first (c->values);
+    const char *host;
+
+    /* nodelist may not exist if job never ran */
+    if (!job->nodelist)
+        return 0;
+    if (!job->nodelist_hl) {
+        /* hack to remove const */
+        struct job *jobtmp = (struct job *)job;
+        if (!(jobtmp->nodelist_hl = hostlist_decode (job->nodelist)))
+            return 0;
+    }
+    host = hostlist_first (hl);
+    while (host) {
+        if (inc_check_comparison (c->mctx, comparisons, errp) < 0)
+            return -1;
+        if (hostlist_find (job->nodelist_hl, host) >= 0)
+            return 1;
+        host = hostlist_next (hl);
+    }
+    return 0;
+}
+
+/* zlistx_set_destructor */
+static void wrap_hostlist_destroy (void **item)
+{
+    if (item) {
+        struct hostlist *hl = *item;
+        hostlist_destroy (hl);
+        (*item) = NULL;
+    }
+}
+
+static struct list_constraint *create_hostlist_constraint (
+    struct match_ctx *mctx,
+    json_t *values,
+    flux_error_t *errp)
+{
+    struct list_constraint *c;
+    struct hostlist *hl = NULL;
+    json_t *entry;
+    size_t index;
+
+    if (!(c = list_constraint_new (mctx,
+                                   match_hostlist,
+                                   wrap_hostlist_destroy,
+                                   errp)))
+        return NULL;
+    /* Create a single hostlist if user specifies multiple nodes or
+     * RFC29 hostlist range */
+    if (!(hl = hostlist_create ())) {
+        errprintf (errp, "failed to create hostlist structure");
+        goto error;
+    }
+    json_array_foreach (values, index, entry) {
+        if (!json_is_string (entry)) {
+            errprintf (errp, "host value must be a string");
+            goto error;
+        }
+        if (hostlist_append (hl, json_string_value (entry)) <= 0) {
+            errprintf (errp, "host value not in valid Hostlist format");
+            goto error;
+        }
+    }
+    if (hostlist_count (hl) > mctx->max_hostlist) {
+        errprintf (errp, "too many hosts specified");
+        goto error;
+    }
+    if (!zlistx_add_end (c->values, hl)) {
+        errprintf (errp, "failed to append hostlist structure");
+        hostlist_destroy (hl);
+        goto error;
+    }
+    return c;
+ error:
+    hostlist_destroy (hl);
+    list_constraint_destroy (c);
+    return NULL;
+}
+
 static int match_timestamp (struct list_constraint *c,
                             const struct job *job,
                             unsigned int *comparisons,
@@ -665,6 +752,8 @@ struct list_constraint *list_constraint_create (struct match_ctx *mctx,
                 return create_states_constraint (mctx, values, errp);
             else if (streq (op, "results"))
                 return create_results_constraint (mctx, values, errp);
+            else if (streq (op, "hostlist"))
+                return create_hostlist_constraint (mctx, values, errp);
             else if (streq (op, "t_submit")
                      || streq (op, "t_depend")
                      || streq (op, "t_run")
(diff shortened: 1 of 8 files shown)


grondo:
I wonder if we should limit the size of hostlist allowed in a constraint query for `job-list`. This probably goes for job constraints too. I was basically able to hang `job-list` with a query like, where no jobs ran on >= `pi4`:
```
{
  "or": [
    {
      "and": [
        {
          "states": [
            62
          ]
        },
        {
          "hostlist": [
            "pi[4-1000000]"
          ]
        }
      ]
    },
    {
      "hostlist": [
        "pi[4-10000000]"
      ]
    }
  ]
}
```

Just add more `or` to get a longer hang...

chu11:
> I wonder if we should limit the size of hostlist allowed in a constraint query

Good catch!  Hmmm not sure what the limit shouild be ... 1024 should be more than enough?

> This probably goes for job constraints too

Yeah, b/c one could make a semi-infinite sized one.  Although I'm unsure how to limit the size.  Let me create an issue for that.

grondo:
1024 might be a bit small on a cluster with 10K nodes. Something dynamic might work if job-list can get the maximum expected instance size. However, this doesn't prevent DoS from submit a complex query (e.g. `or host:foo[1-100]` x1000). There should be some way to break out of a match operation that is taking too long... :thinking:

chu11:
> Something dynamic might work if job-list can get the maximum expected instance size.

Hmmm, I guess it wouldn't be too hard to keep a running track of largest node count seen so far.

> However, this doesn't prevent DoS from submitting a complex query

Yeah, I brought up in #5669, perhaps need some hard caps just to avoid DoS attempts.

chu11:
> Something dynamic might work if job-list can get the maximum expected instance size.

Was looking into this and going back on pros on cons of various approaches:

- calling `sched.resource-status` to get nodelist count
- getting the number of brokers via `flux_get_size()`
- keep running max number of nodes of all jobs

the problems with some of the above

- `sched` module may not be loaded, then what do you do
- number of brokers is good upper bound but not accurate (i.e. multiple brokers per node), but this is probably ok
- assumes cluster size stays the same, which I'm not sure is accurate to assume (e.g. common to shrink cluster as hardware dies off towards EOL .. and there's all the stuff the kubernetes people doing with cluster sizes)
- keep tabs on running max number of nodes for all jobs is ok for everything in memory, but not when there's a job db
  - this maybe could be solved with a side table with the job-db for "max counts" and stuff like that ... although I'd prefer to not do that.

Sooo, I'm not sure..  I'm beginning to wonder if a heuristic might be best.  Like number of brokers OR max job count, whatever is bigger?  then max is a multiple of it?  That's a lot of work for this.  This also makes me wonder if we should hard code some max instead.

chu11:
re-pushed using the following idea to limit hostlist constraint sizes, the limit is instance size (ie number of brokers) or 1024, whatever is bigger.  the 1024 minimum is to give the constraint some decent minimum, in the event the size of the cluster has shrunk over time.

grondo:
>  the limit is instance size (ie number of brokers) or 1024, whatever is bigger. the 1024 minimum is to give the constraint some decent minimum

This seems reasonable to me.

Another idea I had would be for job-list to keep a single hostlist (really a set) of all hosts encountered for all jobs. When a hostlist constraint is encountered, the job-list module could take the intersection of this set and the "all hosts" set to limit the constraint hostlist to only those hosts which could possibly be matched. A similar process could be used to optimize matching of ranks as well.

The drawback is extra work when processing every job to maintain the set of possible hosts, which may not end up being worth it. :thinking:

chu11:
> Another idea I had would be for job-list to keep a single hostlist (really a set) of all hosts encountered for all jobs. When a hostlist constraint is encountered, the job-list module could take the intersection of this set and the "all hosts" set to limit the constraint hostlist to only those hosts which could possibly be matched. A similar process could be used to optimize matching of ranks as well.

In the current implementation this would work, but it'd immediately become a problem if we have older job data stored in a database, which we wouldn't scan upon a instance restart.  So I was trying to avoid doing any optimization based on what we've seen so far.

As mentioned above, I guess this could be solved with a side part of the database of extra stuffs (like "max hosts" or "hosts we've seen so far").  Hmmm, let me think about this more.  Maybe doing that in the DB would be inevitable for optimization purposes.

Or as you say ... maybe it's not worth the energy to do this optimization.

grondo:
Just curious, how are you going to match on hosts with older jobs in a database. Maybe the optimization won't be needed in that case if the hosts are indexed or something. (I haven't seen an implementation, so difficult to reason about). I assume a good database will have already handled the query optimization and "DoS" problem...

chu11:
re-pushed, updating for conflicts, now based on #5681

src/modules/job-list/match.c
```diff
@@ -428,6 +430,90 @@ static struct list_constraint *create_results_constraint (struct match_ctx *mctx
...
+static int match_hostlist (struct list_constraint *c,
+                           const struct job *job,
+                           unsigned int *comparisons,
+                           flux_error_t *errp)
+{
+    struct hostlist *hl = zlistx_first (c->values);
+    const char *host;
+
+    if (inc_check_comparison (c->mctx, comparisons, errp) < 0)
+        return -1;
```
grondo, line 442: The count of comparisons is incremented here and also for each host. Doesn't that mean one host comparison is counted twice?

src/modules/job-list/match.c
```diff
@@ -428,6 +430,90 @@ static struct list_constraint *create_results_constraint (struct match_ctx *mctx
...
+    struct list_constraint *c;
+    struct hostlist *hl = NULL;
+    json_t *entry;
+    size_t index;
+
+    if (!(c = list_constraint_new (mctx,
+                                   match_hostlist,
+                                   wrap_hostlist_destroy,
+                                   errp)))
+        return NULL;
```
grondo, line 487: `errp->text` not set here.

src/modules/job-list/match.c
```diff
@@ -428,6 +430,90 @@ static struct list_constraint *create_results_constraint (struct match_ctx *mctx
...
+    /* Create a single hostlist if user specifies multiple nodes or
+     * RFC29 hostlist range */
+    if (!(hl = hostlist_create ()))
+        goto error;
```
grondo, line 491: `errp->text` not set here.

src/modules/job-list/match.c
```diff
@@ -428,6 +430,90 @@ static struct list_constraint *create_results_constraint (struct match_ctx *mctx
...
+            goto error;
+        }
+    }
+    if (hostlist_count (hl) > mctx->max_hostlist) {
+        errprintf (errp, "too many hosts specified");
+        goto error;
+    }
+    if (!zlistx_add_end (c->values, hl)) {
+        hostlist_destroy (hl);
+        goto error;
```
grondo, line 508: `errp->text` not set here.

src/modules/job-list/match.c
```diff
@@ -743,6 +831,28 @@ struct match_ctx *match_ctx_create (flux_t *h)
         goto error;
     }
 
+    if (flux_get_size (mctx->h, &mctx->max_hostlist) < 0)
+        goto error;
```
grondo, line 835: `errp->text` not set here.

grondo (commented):
On a quick first pass, found some places where `error.text` is not set on error and thus garbage would be returned in the error response.

src/modules/job-list/match.c
chu11: i think this is ok, errp is set in `list_constraint_new()`

grondo: Oh yeah, but looks like it is missing in the `EINVAL` path.

grondo: Oh, nevermind, I think I was looking at the wrong function in the diff.

grondo: Nevermind, I think I was looking at the wrong function in the diff.

src/modules/job-list/match.c
chu11: i think you mean I forgot a flux_log in this case, but correct!!

grondo: Oh, heh. Yeah I must have been blindly checking `goto error` without `errprintf()` :facepalm:

src/modules/job-list/match.c
chu11: I think my logic was there was a comparison simply to see if we should bother to check hosts (i.e. if the job didn't run, we don't check anything, and that should count for at least 1 comparison).  But now that you mention it, perhaps that is not the most sensible way to think about it.  I'll remove the first check.

chu11:

chu11:
rebased now that #5681 is merged

grondo (approved):
I've tested this one several times as part of development of #5711 and LGTM!

chu11:
@MergifyIO rebase

chu11:
oops, missed that this had been approved last week.  rebased and setting MWP. thanks

Issue #5657, opened 2024-01-05
KVS access to job's private namespace hangs


garlick:
Problem: we have a job stuck in completing state on fluke:
```
f2i66WMeQ89D batch    testqe   ./buildan+  C     16     16   2.471m fluke[49-60,62-64,68]
```

The following was observed to hang on nodes other than the management node
```
flux job eventlog -H -p guest.exec.eventlog f2i66WMeQ89D
```
also the equivalent command to access that key directly hangs
```
flux kvs get $(flux job id --to=kvs f2i66WMeQ89D).guest.exec.eventlog
```

Trying random stuff, we found that accessing the namespace directly rather than through the symlink not only works but seems to fix the problem on that one client only.
```
[root@fluke52:~]#  flux kvs dir -N job-735990688096716800 .
exec.
input = {"timestamp":1703210930.2225296,"name":"header","context":{"version":1,"encoding...
output = {"timestamp":1703210930.2301416,"name":"header","context":{"version":1,"encodin...
[root@fluke52:~]# flux job eventlog -H -p guest.exec.eventlog f2i66WMeQ89D
[Dec21 18:08] init
[  +0.013592] starting
[  +0.219157] shell.init service="61494-shell-f2i66WMeQ89D" leader-rank=46 size=16
[  +0.232343] shell.start taskmap={"version":1,"map":[[0,16,1,1]]}
[Dec21 18:10] shell.task-exit localid=0 rank=15 state="Exited" pid=1435953 wait_status=0 signaled=0 exitcode=0
```

We've "fixed" fluke52 and fluke 108 but the problem is still there AFAIK on other ranks.

garlick:
Likely not too helpful, but the guest symlink looks "normal" as far as I can tell
```
[root@fluke108:garlick]# flux kvs dir $(flux job id --to=kvs f2i66WMeQ89D)
job.0a36.c3a1.a200.0800.J = dmVyc2lvbgBpMQBtZWNoYW5pc20Ac211bmdlAHVzZXJpZABpNjE0OTQA.eyJ...
job.0a36.c3a1.a200.0800.R = {"version": 1, "execution": {"R_lite": [{"rank": "46-57,59-6...
job.0a36.c3a1.a200.0800.eventlog = {"timestamp":1703210418.2621763,"name":"submit","cont...
job.0a36.c3a1.a200.0800.guest -> job-735990688096716800::.
[...]

PR #5661, opened 2024-01-07
kvs: store correct namespace after getroot lookup

chu11:
Per discussion in #5657

To solve, I added a "namespace" field in the getroot response.  Obviously could have done an "aux" thing instead.  But felt this was the most reasonable / potentially useful in other ways solution.

I did not add "get" functions for the namespace in `libkvs/kvs_getroot`.  Maybe should / could?  I initially did, but then it seemed silly to add it into `flux kvs getroot` (i.e `flux kvs getroot --namespace foo-namespace` would output `foo-namespace`).


garlick (approved):
This LGTM!  Excellent work @chu11.

Issue #5655, opened 2024-01-04
Flux 0.57.0, `flux filemap get` fails if file was mapped with multiple chunks


ardangelo:
Running Flux 0.57.0, I am attempting to map a file using `flux filemap`,

File information

    $ du $PWD/cti/install/libexec/cti_be_daemon2.18.3
    2456    /g/g11/dangelo3/cti/install/libexec/cti_be_daemon2.18.3
    $ sha1sum $PWD/cti/install/libexec/cti_be_daemon2.18.3
    8868b9a0f785cef702825033dcc0ee138a2d3b70  /g/g11/dangelo3/cti/install/libexec/cti_be_daemon2.18.3

Map with default 1M chunk size fails get

    $ flux filemap map -C $PWD/cti/install/libexec cti_be_daemon2.18.3
    $ flux filemap get cti_be_daemon2.18.3
    flux-filemap: cti_be_daemon2.18.3: error loading offset=0 size=1048576 from sha1-459929af132509bbcc513878cb6360ed04c2ad6c: mapped file content has changed

Map with unlimited chunk size enables get

    $ flux filemap map --chunksize=0 -C $PWD/cti/install/libexec cti_be_daemon2.18.3
    $ flux filemap get cti_be_daemon2.18.3
    $ # No error

grondo:
I'm not sure, but could this be because you are calling `flux filemap get` on the same node on which you mapped the file? When you call get, if there are multiple chunks then the first chunk might be written thus corrupting the map. Perhaps with a single chunk the file does not actually change since data is written in one go. 

 Try the `flux filemap get` to a different path with `-C` and see if that resolves the issue.

grondo:
Ah, @ardangelo, I apologize. I didn't notice you were already using `-C` in `flux filemap map`.

I think your usage should work, and FWIW, I could not reproduce the error with a simple test case:
```sh
$ flux version
commands:    		0.57.0
libflux-core:		0.57.0
libflux-security:	0.10.0
broker:  		0.57.0
FLUX_URI:		local:///var/tmp/grondo/flux-TUaqej/local-0
build-options:		+ascii-only+systemd+hwloc==2.8.0+zmq==4.3.4
$ ls -lh map
total 0
-rw-r--r-- 1 grondo grondo 10M Jan  4 16:41 10M
$ flux filemap map -C $PWD/map 10M
$ flux filemap list
10M
$ flux filemap get 10M
$ ls -lh 10M
[...]

PR #5662, opened 2024-01-08
add flux filemap get --overwrite and change the default overwrite behavior

garlick:
Problem: as discussed in #5655, `flux filemap get` can too easily ovewrite the source file causing mahem in a file broadcast operation.

This changes the default behavior to fail the `get` operation if a destination file exists, and adds a `--overwrite` option that can be used to get the old behavior.

### DIFF
--- a/src/cmd/builtin/filemap.c
+++ b/src/cmd/builtin/filemap.c
@@ -315,6 +315,10 @@ static void extract (flux_t *h,
     bool direct = optparse_hasopt (p, "direct");
     flux_future_t *f;
     flux_error_t error;
+    int opts = 0;
+
+    if (!optparse_hasopt (p, "overwrite"))
+        opts |= ARCHIVE_EXTRACT_NO_OVERWRITE;
 
     if (!(f = filemap_mmap_list (h, !direct, tags, pattern)))
         log_err_exit ("mmap-list");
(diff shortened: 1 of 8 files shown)


src/cmd/builtin/filemap.c
```diff
@@ -315,6 +315,10 @@ static void extract (flux_t *h,
     bool direct = optparse_hasopt (p, "direct");
     flux_future_t *f;
     flux_error_t error;
+    int opts = 0;
```
grondo, line 318: Commit message typos: "becuase", "mimicing"

grondo (approved):
This LGTM!

grondo:
While going through this I wondered if an option like [tar(1)](https://man7.org/linux/man-pages/man1/tar.1.html) `--skip-old-files` would be more user friendly. Also, I noticed the libarchive [ARCHIVE_EXTRACT_SAFE_WRITES]](https://manpages.debian.org/testing/libarchive-dev/archive_write_disk.3.en.html#ARCHIVE_EXTRACT_SAFE_WRITES), which might be interesting to expose. Both of these shouldn't hold up the current PR thought, they were just interesting observations on my part.

garlick:
Thanks for that suggestion!  I think I'll resist the temptation to add those options now.  Since `filemap_extract()` now accepts `libarchive_flags` it should be easy to add later.

Just fixed a couple more things
- a typo in the "improved" error message
- a bug I noticed in EPERM handling in `flux filemap map`

garlick:
OK to set MWP?

grondo:
Yes, all good!

garlick:
Thanks!

Issue #5622, opened 2023-12-13
broker: `parent-uri` attribute is set for instance with no parent


grondo:
I'm not certain, but this seems wrong:
```console
$ flux start flux start flux getattr parent-uri
local:///var/tmp/grondo/flux-j6S7sS/local-0
```

It appears that the broker sets `parent-uri` whenever `FLUX_URI` is set in the environment, even if the enclosing instance isn't technically the parent of the new instance.

Is this expected behavior?

garlick:
I think that is the expected behavior but maybe because I didn't think it through?

I guess a test instance should always be the "top" flux even if it was started by Flux?  What is the fallout from the current behavior?

grondo:
In this case flux did not start the test instance, it just happened to be started under an existing instance. It seems like an instance should only be considered the parent if it allocated resources to the child instance?

The fallout is that existence of the `parent-uri` attribute alone doesn't indicate that the current instance was launched by Flux. I guess both `parent-uri` and `jobid` need to be considered?

grondo:
I also admit I haven't thought this through either, though. It is simple enough to check for a `jobid` attribute, and this issue was only opened to get clarification.

garlick:
There is also this, which works as expected.
```
$ flux start flux start flux getattr instance-level
0
```

Yeah seems like when flux is not run as a flux job, all three of these should be true
- `instance-level` is 0
- `jobid` is not set
- `parent-uri` is not set

grondo:
I just went through the code and currently all places where `parent-uri` is potentially used, the `jobid` attribute is also checked and both must be set before an attempt to contact the parent is made. So, currently there is no fallout.

There is this comment in `src/modules/resource/inventory.c`
https://github.com/flux-framework/flux-core/blob/b0c565f9a08820cf0d1d3995cfd6488c04d21b64/src/modules/resource/inventory.c#L40-L46

which probably needs a very slight tweak.

garlick:
Does this seem right then?
```diff
diff --git a/src/broker/broker.c b/src/broker/broker.c
index 7bdb38bdd..c040f1cf9 100644
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -629,7 +629,14 @@ static void init_attrs (attr_t *attrs, pid_t pid, struct flux_msg_cred *cred)
 {
     const char *val;
 
-    val = getenv ("FLUX_URI");
+    /* Set the parent-uri attribute IFF this instance was run as a job
[...]

PR #5663, opened 2024-01-08
broker: only set parent-uri when instance is a job

garlick:
This ensures that the `parent-uri` broker attribute is only set when Flux was lauched as a Flux job, as discussed in #5622.

### DIFF
--- a/t/t3100-flux-in-flux.t
+++ b/t/t3100-flux-in-flux.t
@@ -39,9 +39,9 @@ test_expect_success "flux --parent works in subinstance" '
 '
 
 test_expect_success "flux --parent --parent works in subinstance" '
-	id=$(flux submit \
-		flux start ${ARGS} \
-		flux start ${ARGS} flux --parent --parent kvs put test=ok) &&
+	id=$(flux batch -n1 --wrap \
+		flux run flux start ${ARGS} \
+		flux --parent --parent kvs put test=ok) &&
 	flux job attach $id &&
 	flux job info $id guest.test > guest2.test &&
 	cat <<-EOF >guest2.test.exp &&
--- a/doc/man7/flux-broker-attributes.rst
+++ b/doc/man7/flux-broker-attributes.rst
@@ -61,8 +61,7 @@ local-uri [Updates: C]
    existing directory.
 
 parent-uri
-   The value of the broker's :envvar:`FLUX_URI` environment variable.  This is
-   the URI that should be passed to :man3:`flux_open` to establish a connection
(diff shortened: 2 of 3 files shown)


t/t3100-flux-in-flux.t
```diff
@@ -39,9 +39,8 @@ test_expect_success "flux --parent works in subinstance" '
 '
 
 test_expect_success "flux --parent --parent works in subinstance" '
-	id=$(flux submit \
-		flux start ${ARGS} \
-		flux start ${ARGS} flux --parent --parent kvs put test=ok) &&
+	id=$(flux batch -n1 --wrap \
+		flux run ${ARGS} flux --parent --parent kvs put test=ok) &&
```
grondo, line 43: Is a `flux start` missing here. That is to simulate a depth of 2 for the `flux kvs put`, should this be:
```
id=$(flux batch -n1 --wrap \
    flux run ${ARGS} flux start flux --parent --parent kvs put test=ok)
```

grondo (approved):
LGTM! Just one question about the test (inline)

t/t3100-flux-in-flux.t
garlick: There's the outer test instance form `test_under_flux` (level 0), then the `flux batch` instance (level1), then the `flux start` instance (level2).  So I think it's right?

grondo: I just see `flux run ${ARGS} flux --parent --parent kvs put test=ok` -- where am I missing the `flux start`?
(In the diff I see two `flux start` calls removed, replaced with one `flux batch` and one `flux run`) Sorry if I'm missing something obvious, sometimes hard to see via a diff...)

garlick: Oh I do apologize - I was reading your "shouldn't it be" code snippet thinking it was the original.  You are quite right!

t/t3100-flux-in-flux.t
```diff
@@ -39,9 +39,9 @@ test_expect_success "flux --parent works in subinstance" '
 '
 
 test_expect_success "flux --parent --parent works in subinstance" '
-	id=$(flux submit \
-		flux start ${ARGS} \
-		flux start ${ARGS} flux --parent --parent kvs put test=ok) &&
+	id=$(flux batch -n1 --wrap \
+		flux run ${ARGS} flux start \
```
grondo, line 43: One other comment sorry! Are `${ARGS}` meant for `flux start` and not `flux run` here?

garlick: Oh duh.  Fix coming.

PR #5665, opened 2024-01-09
add -x option to flux-alloc and flux-batch

garlick:
Problem: for some reason, the `--exclusive` option in `flux-alloc` and `flux-batch` is not paired with a short `-x` option like it is in the other submission commands.

Add the short option.


chu11 (approved):
LGTM!

garlick:
Thanks!  Setting mwp.

Issue #5671, opened 2024-01-09
priority plugin posting identical jobspec-update event twice


grondo:
As seen in flux-framework/flux-accounting#406, the mf_priority plugin appears to be posting a `jobspec-update` event to set the default queue twice per job:
```
[Jan05 18:00] submit userid=61494 urgency=16 flags=0 version=1
[  +0.018063] jobspec-update attributes.system.bank="guests"
[  +0.018097] jobspec-update attributes.system.bank="guests"
```

Not a critical issue, but it would be nice to determine why the second jobspec-update event is posted (if indeed it is from `mf_priority.so`), and remove it since it is unnecessary.

cmoussa1:
Oh great! Thanks for letting me know. Please let me know if there's anything I can do to help with the fix or testing it on my end once a fix is proposed.

PR #5673, opened 2024-01-10
job-manager: fix duplicate posting of jobspec-update event from plugins

grondo:
As described in #5671, plugins that modify jobspec may end up having duplicate `jobspec-update` events posted. This is wasteful.

The problem occurs because `jobtap_post_jobspec_updates()`, which is called _after_ all plugin callbacks and posts any pending jobspec updates to avoid modifying jobspec while the plugin is still active,  may be called reentrantly as a result of the jobspec updates. The function does not clear `jobtap->jobspec_update` before posting the event, so the 2nd (or more?) calls will post the duplicate.

This PR fixes that issue by ensureing `jobtap_post_jobspec_updates()` is reentrant, and adds a reproducer for this issue to the testsuite.

### DIFF
--- a/src/modules/job-manager/event.c
+++ b/src/modules/job-manager/event.c
@@ -747,7 +747,7 @@ static int event_jobtap_call (struct event *event,
         if (json_unpack (entry, "{s:o}", "context", &updates) < 0) {
             flux_log (event->ctx->h,
                       LOG_ERR,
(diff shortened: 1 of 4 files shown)


chu11 (approved):
LGTM!

grondo:
Thanks! I'll set MWP.

PR #5675, opened 2024-01-11
set userid to instance owner in job manager exceptions

garlick:
Problem: job manager exception events are posted with the userid set to FLUX_USERID_UNKNOWN which looks wrong when reviewing logs.

Post exceptions as the instance owner.


t/t2900-job-timelimits.t
```diff
@@ -19,6 +19,12 @@ test_expect_success 'job time limits are enforced' '
 		flux run --time-limit=${TIMEOUT}s sleep 30 2>limit1.err &&
 	grep "resource allocation expired" limit1.err
 '
+test_expect_success 'tiemout exception was generated by instance owner' '
```
grondo, line 22: s/tiemout/timeout/

grondo (approved):
LGTM! Noticed one typo.

t/t2900-job-timelimits.t
garlick: Doh! I had fixed that but I guess I forgot to push it.   Pushed it now and I'll set MWP.  Thanks!

PR #5677, opened 2024-01-11
broker: allow single-user rexec to rank 0

garlick:
Problem: remote rexec to rank 0 is not permitted, but it would be useful to allow shell plugins to run things on rank 0 in order to utilize leader-only services like 'flux filemap map'.

Remote rexec is restricted in order to prevent the flux user from remotely executing arbitrary programs on a restricted management node. This is not a concern for other Flux instances such as batch jobs.

Change the rexec authorization callback to allow remote access on rank 0 if the 'access.allow-guest-user' TOML key is unset or false.

Update tests.

### DIFF
--- a/src/broker/exec.c
+++ b/src/broker/exec.c
@@ -8,6 +8,14 @@
  * SPDX-License-Identifier: LGPL-3.0
 \************************************************************/
 
+/* exec.c - broker subprocess server
+ *
+ * The service is restricted to the instance owner.
+ * In addition, remote access to rank 0 is prohibited on multi-user instances.
+ * This is a precaution for system instances where rank 0 is deployed on a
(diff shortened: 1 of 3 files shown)


garlick:
oops forgot to mention the issue - just repushed.

grondo:
> oops forgot to mention the issue - just repushed.

You can also put `Fixes #XXXX` in the PR description and merging the PR will auto-close the issue (just FYI in case you didn't know that)

grondo (approved):
This LGTM!

garlick:
Thanks - yeah I kind of wanted to put the x-ref in the history though.  Maybe that doesn't really matter much.  Anway, I'll set MWP.

PR #5678, opened 2024-01-12
drop jsonschema requirement

### LINKED ISSUE #4918: remove jobspec schema
The jobspec jsonschema based validator is slow, difficult to maintain, and is not used in production.

We should decide if the jsonschema is providing any value and remove it if not.

grondo:
This PR is a proposal to drop jsonschema from flux-core. The jobspec schemas are likely way out of date, and I don't think there's even a use case to use them anywhere outside of testing the schema-based validation. If there is ever a use case, the schemas could be resurrected from history and updated at that point.

Fixes #4918


garlick (approved):
Works for me!

grondo:
Well this has been open for >4hrs and no objections, so MWP I guess!

trws:
I missed this until today.  Probably a reasonable decision, although I'm still of the opinion that we would be better off with something like jsonschema that we can use to _generate parsing and generation code_ as a way to make communicating over the flux overlay more practical without having to call the C routines in the long term.  Not a priority, but this would probably be the place to come back to if someone needs to do that in the future.

Issue #5679, opened 2024-01-12
`flux start` hangs when run under `flux proxy`


grondo:
Noticed this recently when trying to run a sharness test while under a `flux-proxy`.
```console
 grondo@corona212:~/git/flux-core.git$ flux start 
f(s=1,d=0) grondo@corona212:~/git/flux-core.git$ flux alloc -n1 --bg
f4dPq5ZH
f(s=1,d=0) grondo@corona212:~/git/flux-core.git$ flux proxy f4dPq5ZH
f(s=1,d=1) grondo@corona212:~/git/flux-core.git$ flux start -vvv
flux-start: /usr/libexec/flux/cmd/flux-broker
# hang
```

The broker appears to be stuck in a `flux_attr_get()` call in the local connector:
```gdb
(gdb) thread apply all bt

Thread 1 (Thread 0x7f91c064c0 (LWP 2646546) "flux-broker-0"):
#0  0x0000007f925e311c in __GI_epoll_pwait (epfd=12, events=0x557644dd20, maxevents=64, timeout=59743, set=set@entry=0x0) at ../sysdeps/unix/sysv/linux/epoll_pwait.c:42
#1  0x0000007f925e3328 in epoll_wait (epfd=<optimized out>, events=<optimized out>, maxevents=<optimized out>, timeout=<optimized out>) at ../sysdeps/unix/sysv/linux/epoll_wait.c:32
#2  0x0000007f92902b90 in epoll_poll (loop=0x557645fe30, timeout=59.743000000000002) at ev_epoll.c:155
#3  0x0000007f9290578c in ev_run (flags=0, loop=0x557645fe30) at ev.c:4157
#4  ev_run (loop=0x557645fe30, flags=0) at ev.c:4021
#5  0x0000007f928d4d10 in flux_reactor_run (r=0x557645fe10, flags=flags@entry=0) at reactor.c:128
#6  0x0000007f928de984 in flux_future_wait_for (f=f@entry=0x557645fce0, timeout=timeout@entry=-1) at future.c:476
#7  0x0000007f928deaec in flux_future_get (f=0x557645fce0, f@entry=0x7f92914458, result=result@entry=0x7fc8e5e5e8) at future.c:515
#8  0x0000007f928dbda0 in flux_rpc_get_vunpack (ap=..., fmt=0x7f92914458 "{s:s s:i}", f=0x7f92914458) at rpc.c:157
[...]

PR #5680, opened 2024-01-12
fix broker hang under `flux proxy`

grondo:
This applies the changes discussed in #5679 to fix `flux start` hangs under `flux proxy`.


garlick (approved):
LGTM!

grondo:
I think the new test may be causing occasional hangs. Looking into that before merging.

grondo:
Hm actually the latest hang was in the cron faketime test at
```
 expecting success: 
      flux module load cron
```

I'm assuming that was a one-off. Rebased on master so we'll see if there's any hangs this time.

PR #5681, opened 2024-01-13
job-list: limit constraint comparisons to avoid DoS

### LINKED ISSUE #5669: job-list: limit size of job constraint
As @grondo notes in #5656, "very large" constraints might have the ability to hang (or minimally slow down) the job-list service. 

We may wish to put a limit on the size of job constraints to limit this.

Without thinking about this that much, perhaps a cap on the length of an operator array might be worthwhile.  32 or 64?  i.e.

`{ "and": [ { "userid": [ ... ]}, { "name":` [ ...] }, ... ] }`

If we say there's a cap of 32, the max length of the values array for "and" would be 32, and the max of "userid" and "name" would also be 32.

Also we'd probably need a "recursive" limit too, without thinking about it much ... 8 or 16?

chu11:
Problem: Job-list constraints are not limited in their size.  A nefarious user could cause a DoS of the job-list service by sending an extremely large constraint request.

Solution: Set a reasonable max on constraint sizes.  The maximum length a constraint array can be is 256 elements and the maximum recursive depth (via and/or/not operators) is set to 16.

Add unit tests.

Fixes #5669

----

This was something I came up with ... good idea?  bad idea?  the limit sizes ok?

### DIFF
--- a/src/modules/job-list/list.c
+++ b/src/modules/job-list/list.c
@@ -314,7 +314,7 @@ void list_cb (flux_t *h, flux_msg_handler_t *mh,
         errno = EPROTO;
         goto error;
     }
-    if (!(c = list_constraint_create (constraint, &error))) {
+    if (!(c = list_constraint_create (ctx->mctx, constraint, &error))) {
         errprintf (&err,
                    "invalid payload: constraint object invalid: %s",
                    error.text);
--- a/src/modules/job-list/list.c
+++ b/src/modules/job-list/list.c
@@ -56,6 +56,7 @@ int get_jobs_from_list (json_t *jobs,
 
     job = zlistx_first (list);
     while (job) {
+        int ret;
 
         /*  If job->t_inactive > 0. we're on the inactive jobs list and jobs are
          *  sorted on the inactive list, larger t_inactive first.
@@ -68,7 +69,9 @@ int get_jobs_from_list (json_t *jobs,
         if (job->t_inactive > 0. && job->t_inactive <= since)
             break;
 
-        if (job_match (job, c)) {
+        if ((ret = job_match (job, c, errp)) < 0)
+            return -1;
+        if (ret) {
             json_t *o;
             if (!(o = job_to_json (job, attrs, errp)))
                 return -1;
--- a/src/modules/job-list/match.c
+++ b/src/modules/job-list/match.c
@@ -26,11 +26,13 @@
 
 typedef int (*match_f) (struct list_constraint *,
                         const struct job *,
+                        unsigned int *,
                         flux_error_t *);
 
 struct list_constraint {
     zlistx_t *values;
     match_f match;
+    unsigned int comparisons;   /* total across multiple calls to job_match() */
 };
 
 typedef enum {
@@ -53,6 +55,20 @@ struct timestamp_value {
     match_timestamp_type_t t_type;
     match_comparison_t t_comp;
 };
+
+#define CONSTRAINT_COMPARISON_MAX 1000000
+
+static inline int inc_check_comparison (unsigned int *comparisons,
+                                        flux_error_t *errp)
+{
+    if ((++(*comparisons)) > CONSTRAINT_COMPARISON_MAX) {
+        errprintf (errp,
+                   "Excessive comparisons made, "
+                   "limit search via states or since");
+        return -1;
+    }
+    return 0;
+}
 
 static void timestamp_value_destroy (void *data)
 {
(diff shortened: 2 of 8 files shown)


grondo:
This seems reasonable as a first proposal, but it feels like it may be tricky to thwart all attacks using these kinds of limits.
For example, the 256 value array length would let me create an `or` of 256 values, each of which could be a non-matching hostlist of 1024 elements, and I could `or` that together 16 times for a total of 4M non-matching hosts. Depending on the size of the job lists, this could still easily cause a DoS (though some limit is better than nothing)

This problem feels similar to sandboxing in embedded languages (e.g. Lua), and one approach often used there is to set an "iteration limit" to avoid infinite (or very long)  loops. I wonder if that would be possible here. That is, instead of putting limits on individual components of the constraint object, you carry a counter (or even timer) along with the match operation, and abort it if it hits some limit. E.g. if using a counter, each "op" would increase the count by the number of "values" (where in a hostlist each host is considered a value), or you just increment and check the counter for each match. To allow long matches, you could make the limit configurable and/or disable it for instance owner. This might be slightly more work, and would allow large matches to proceed instead of catching them before they are even attempted, but would be more reliable in that it would place a hard limit on the amount of time a match is allowed to take. Just an idea, and perhaps it could even be used in concert with the constraint limits already proposed in this PR.

chu11:
> To allow long matches, you could make the limit configurable and/or disable it for instance owner. This might be slightly more work, and would allow large matches to proceed instead of catching them before they are even attempted, but would be more reliable in that it would place a hard limit on the amount of time a match is allowed to take. Just an idea, and perhaps it could even be used in concert with the constraint limits already proposed in this PR.

Hmmmm I like this idea.  Lemme give it a whirl see how it turns out.  Gotta carry around some counting stuffs while refactoring things.

chu11:
re-pushed a WIP of limiting the number of "comparisons" approach.  I choose 100K as the random "max" comparisons for the time being and adding a `job-list.max_comparisons` config option.  The key commit is:

https://github.com/flux-framework/flux-core/pull/5681/commits/76015e577ae28c230c9edb3118d446f6485c7246

with the rest mostly refactoring or config stuff.  I haven't tested yet or added tests, but wanted to post to see what people thought of the general approach.

chu11:
re-pushed, add some tests and removing WIP.

I didn't add back the max elements and max recursion depth of the constraint, as I felt this handled the overriding case.

src/modules/job-list/list.c
```diff
@@ -68,7 +70,9 @@ int get_jobs_from_list (json_t *jobs,
         if (job->t_inactive > 0. && job->t_inactive <= since)
             break;
 
-        if (job_match (job, c)) {
+        if ((ret = job_match (job, c, &error)) < 0)
+            return -1;
```
grondo, line 74: This function uses a local `flux_error_t` to handle the error message from `job_match()`, but doesn't copy it to `errp->text` on error. Should `errp` be used here and the local `flux_error_t error` be dropped?

grondo: Oh, this also indicates that there's no testing of a proper error message in the response when `job_match()` fails.

grondo (commented):
In testing, I noticed garbage returned in error messages where `job_match()` fails, e.g. when the max comparisons limit is exceeded.

Issue noted inline.

src/modules/job-list/list.c
chu11: good catch, I have tests to ensure the error occurs but didn't check the message

chu11:
re-pushed, fixing per comments above.

The main change was having `flux job list` call `future_strerror()` so it would output the descriptive error vs the generic "errno error".  So we can now test that the correct error message is returned.

src/modules/job-list/match.c
```diff
@@ -54,6 +56,20 @@ struct timestamp_value {
     match_comparison_t t_comp;
 };
 
+#define CONSTRAINT_COMPARISON_MAX 100000
```
grondo, line 60: This limit feels a bit low, for example:
```
> 0.00000
> job-list.list
> {"max_entries": 1000, "attrs": ["dependencies", "ntasks", "t_run", "state", "result", "t_cleanup", "duration", "name", "nodelist", "nnodes", "queue", "annotations", "userid"], "since": 0.0, "constraint": {"and": [{"hostlist": ["corona[1-1000]"]}, {"userid": [6885]}]}}
< 0.04141
< job-list.list
< flags=topic,payload,route userid=6885 rolemask=owner,local errnum=22 matchtag=1
< Excessive comparisons made, limit search via states or since
```

The limit was hit in 40ms on corona. If we say on this system the limit should prevent job-list getting stuck doing matches for more than say 200ms, then we can probably raise the limit by 4-5x.

chu11: ahhh i set this limit before the hostlist support (which is in the follow up PR).

Maybe just upping it to like 1-2 million would even be fine.  I think the main goal is to just not DoS job-list an obscene amount.

chu11:
re-pushed, upping default limit to 1 million.

grondo (approved):
Great! LGTM.

chu11:
great, thanks!  setting MWP

PR #5683, opened 2024-01-17
libsubprocess: make flux_buffer class private

garlick:
Problem: the flux_buffer class and its watchers contribute 34 functions to the public Flux API but for the most part they are only used by libsubprocess, and aren't documented.  Furthermore, this class may need some work to support flow control in the subprocess server.

This moves `flux_buffer_*` et al into libsubprocess and shortens names to `fbuf_*`, making them private to flux-core.  A few minor problems are addressed, and some unused functionality culled.

`flux-exec(1)` and `flux-job(1)` continue to use fbuf to line-buffer standard input.  I didn't think this was necessarily an example of why this code might be needed in the Flux API since that particular use could be trivially implemented another way.


src/common/libsubprocess/fbuf.c
```diff
@@ -161,25 +161,6 @@ static void nonfull_transition_check (struct fbuf *fb, bool was_full)
     }
 }
 
-int fbuf_drop (struct fbuf *fb, int len)
```
chu11: commit message: have to mention adjust a few callers to use fbuf_read over fbuf_peek + fbuf_drop?  Maybe not a huge deal.

chu11 (approved):
overall LGTM, just a few nits I found

garlick:
@chu11, ok to set MWP or did you want that internal function name changed?

(And thanks for the review!)

chu11:
Do it! :-)

PR #5687, opened 2024-01-18
job-list: misc cleanup

chu11:
Peeled off some cleanups from other PR work.  The removal of `job_list_error_t` is the big cleanup here, as it'll end up conflicting with several of the PRs in progress.

### DIFF
--- a/src/bindings/python/flux/job/list.py
+++ b/src/bindings/python/flux/job/list.py
@@ -199,6 +199,9 @@ class JobList:
           not empty.
     :user: Username or userid for which to fetch jobs. Default is all users.
     :max_entries: Maximum number of jobs to return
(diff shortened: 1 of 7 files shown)


src/bindings/python/flux/job/list.py
```diff
@@ -199,6 +199,9 @@ class JobList:
           not empty.
     :user: Username or userid for which to fetch jobs. Default is all users.
     :max_entries: Maximum number of jobs to return
+    :since: Limit jobs to those that have een active since a given timestamp.
```
grondo, line 202: typo: "that have een active"

grondo (approved):
LGTM! Just noticed one typo

PR #5690, opened 2024-01-18
librlist: workaround xml buffer size issue in some hwloc versions

### LINKED ISSUE #5689: librlist: test_rhwloc.t fails with hwloc 2.10
The fix or w/a applied in 3824a6981010c15214664e36574f5b244b8f4900 does not work with hwloc 2.10.

The source of the issue does appear to be a change in libxml2 as reported here:

https://github.com/open-mpi/hwloc/issues/639

As noted in that hwloc issue, a fix went into 2.10.0, which is why the "fix" applied in PR #5682 doesn't work: the hwloc "fix" simply subtracts one from the length passed to `hwloc_topology_set_xmlbuffer(3)`. Therefore you can see that in 2.10 we can't pass `strlen(xml)` since the final character of the XML will be truncated. :confused: 

I'm not really sure what to do here to easily support multiple hwloc versions. Might have to add a specific test to `configure`...

grondo:
Problem: With some versions of libxml, the XML load function does not tolerate a buffer length that includes the trailing NUL character. In hwloc 2.10.0, hwloc_topology_set_xmlbuffer() was changed to decrement the length before passing to the underlying XML call, so this effectively means that hwloc now *requires* the NUL character be included in length, but current librlist code does not include it to workaround the problem in v2.10.

Since there is no way to determine if NUL should or should not be included in hwloc_topology_set_xmlbuffer(), include the NUL by default, but retry without it if the first call fails.

Fixes #5689

### DIFF
--- a/src/common/librlist/rhwloc.c
+++ b/src/common/librlist/rhwloc.c
@@ -60,13 +60,24 @@ static int init_topo_from_xml (hwloc_topology_t *tp,
                                const char *xml,
                                unsigned long flags)
 {
-    if ((topo_init_common (tp, flags) < 0)
-        || (hwloc_topology_set_xmlbuffer (*tp, xml, strlen (xml)) < 0)
-        || (hwloc_topology_load (*tp) < 0)) {
-        hwloc_topology_destroy (*tp);
-        return (-1);
+    int len = strlen (xml) + 1;
+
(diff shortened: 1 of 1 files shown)


garlick (approved):
LGTM.  Wow, that's pretty poor API management.

grondo:
Thanks! Setting MWP.

PR #5699, opened 2024-01-22
WIP libsubprocess API cleanup

garlick:
This is some subprocess API cleanup peeled off of other work.

I had hoped to get the following done but this is taking too much time at the moment and I need to park it (hence WIP)
- [ ] add `flux_cmd_create(3)` man page
- [ ] add `flux_subprocess_rexec(3)` man page
- [ ] consider changes to channel I/O API as the current one is a bit strange
- [ ] re-evaluate server handling of output stream suppression (where inheriting from parent may be wrong)
- [ ] update [RFC 42](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_42.html) server flags as they changed here


src/common/libsubprocess/subprocess.h
```diff
@@ -103,14 +103,14 @@ typedef struct {
 /*
  *  llog-compatible callback
  */
-typedef void (*subprocess_log_f) (void *arg,
```
grondo, line 106: I think I noticed `command.h` has a similar issue.

grondo (approved):
LGTM!

src/common/libsubprocess/subprocess.h
garlick: Ah good catch.  I went ahead and added a vi mode line to all headers in libsubprocess and fixed command.h.

Issue #5700, opened 2024-01-22
flux filemap map only works on rank 0


garlick:
Problem: as noted in #5697, `flux filemap map` only works on the rank 0 broker, but tools implemented with shell plugins may not have access to rank 0.

This is a design limitation because rank 0 is currently the root of the content cache tree.

One thought is to add something like `flux filemap copy` to copy a set of files to the rank 0 broker, which could then be mapped.

A more complex solution would be to figure out how to mmap files on any rank and register the content blobs with rank 0 so they can be faulted in as needed.

garlick:
Storing files in the kvs might be another way to go, where storage would be consumed on rank 0 and the content cache would be leveraged for parallel reads.

grondo:
One thought is that it might be nicer for users if there was one set of commands that work in both use cases here: distributing files from rank 0 vs from other ranks.  I wonder if we could offer a command that does the Right Thing in either case?

One other thought is when running many jobs, each of which uses this facility, content store usage on rank 0 could grow quickly and there is no way to remove the archives.

Otherwise, I think the `flux kvs archive` could work and perhaps is a handy tool nonetheless (maybe someone wants to archive results or something in the kvs for provenance, etc)

garlick:
Great points!

If we go forward, then I agree, we probably should take a look at redesigning `flux filemap` (possibly renamed) to incorporate this rather than tuck it away in `flux kvs`.

A TODO for the prototype is to figure out how to reference the content blobs so the archive would be complete on a dump/restore.  I thought maybe if the key is `data.foo`, we could optionally write a `data.foo.blobs` directory containing keys that are just the blobref strings, that point to the actual blobs.  As it is the archive references blobs that might not be included in the dump.

The caveats would need to be documented of course, but I'm liking this because it leverages a lot of existing work.

grondo:
Yeah, nice work!

garlick:
Pondering about a unified utility...  Maybe:

```
flux archive create ARCHIVE-KEY [--mmap] PATH ...
flux archive remove ARCHIVE-KEY
flux archive list ARCHIVE-KEY [pattern]
flux archive get ARCHIVE-KEY [pattern]
```
where the `create` subcommand would work on any rank, unless the `--mmap` option is specified in which case rank 0 only.

In other words do away with the tags used in `flux filemap` and use KVS keys instead.  What's actually stored under the keys could be debated, but it might be OK to move at least the metadata out of rank 0 broker memory into the KVS....

[...]

PR #5701, opened 2024-01-23
flux-archive: add new command for file broadcast

garlick:
Here's 1/2 a prototype of an idea to address the `flux filemap` limitations dicsussed in #5700.

This adds `flux kvs archive`, which stores a dataset in the KVS using the "fileref" protocol.  Pulling the data out again (tool not prototyped yet) would benefit from the hierarchical content cache scaling like filemap, but unlike filemap, any rank could put data in.

A few caveats:
- contributes to rank 0 content storage growth
- the content blobs are effectively unreferenced (by KVS metadata) and will be garbage collected in offline purging

Still pondering whether this is a good idea or not, but it was pretty easy to prototype given most of the hard stuff is already available in convenience libraries, so why not play around?


garlick:
Added an extract subcommand to complete the toybox.

garlick:
Repushed with a few improvements
- no "default key', and key is now a positional argument instead of an option
- added `--waitcreate` to` flux kvs extract`
- refuse to overwrite existing key unless `--overwrite` is specified
- fix bug archiving directories

garlick:
Couple more improvements:
- add `-t,--list` option to `flux kvs extract` for listing the table of contents without extracting
- when archiving `.`, don't include the directory itself in the archive
- update an earlier tweak to libarchive error reporting to cover additional cases
- move fileref.c into libfilemap since it seems like it belongs there

garlick:
I've reworked this to merge `flux-filemap` and the kvs subcommands originally proposed into a new [flux-archive(1)](https://flux-framework--5701.org.readthedocs.build/projects/flux-core/en/5701/man1/flux-archive.html)  command as proposed a few comments up.

Although some testing is still missing and I'm not quite ready for code review, it might be good to get a quick take on the CLI as I've been looking at it too long and have lost my objectivity!

In summary:
- An archive always lives in the KVS and may be created from any rank
- Bulk data is copied by default into the content store and impacts storage on rank 0
- However, an `--mmap` option lets bulk data be mapped rather than copied like with `flux filemap`.
- The archive may be retrieved through the KVS, so probably more scalable for file broadcast than before when it was fetched directly from the rank 0 content module.
- Archives work with the `stage-in` shell plugin as before, described in [flux-shell(1)](https://flux-framework--5701.org.readthedocs.build/projects/flux-core/en/5701/man1/flux-shell.html#cmdoption-flux-shell-arg-stage-in.keys)

garlick:
Just realized we should add some `stage-in` options to support the Cray CCE use case, where a Cray shell plugin on shell rank 0 produces some data that needs to be broadcast to the other shell ranks:
- `stage-in.waitcreate` - block until archive is available
- `stage-in.no-force-primary` - use the private job namespace

Some more thought might be required about the order of things here...
- We must avoid deadlock if `stage-in` runs and blocks on `waitcreate` before Cray plugin can create archive on the same shell rank
- It would be nice if the Cray plugin could set `stage-in` options so an elaborate `flux run` command line is not required

grondo:
This looks great! Thanks for providing nice docs already.

My only thought on the interface is that would it be slightly nicer if an archive name or key was required for all these commands, and by default all archives were namespaced under `archive.`. This could allow the archive NAME to be specified as a free argument similar to existing archive commands, instead of as a `--key=KEY` option. Also, since all the archives would be under a known kvs dir, a `flux archive list` command could be supplied for users that forgot the dang name they used when the archived something (and they wouldn't have to search the entire kvs namespace for it)

However, given that the use case for these tools is mainly as a scripting interface, there probably isn't much of a need for a `list` interface at this time and the suggestion is just something that occurred to me while reading the nice docs!

grondo:
> stage-in.waitcreate - block until archive is available

This may not be necessary. The Cray shell plugin could create the archive in `shell.init`, and non-rank 0 shells could extract it in `shell.post-init`. There is a barrier between the two so order is guaranteed.

grondo:
One other thought about the default `archive.main` for archive keys: This seems dangerous if multiple components are creating separate archives and are using the default key. This is perhaps another argument for requiring users provide an archive name...

garlick:
> My only thought on the interface is that would it be slightly nicer if an archive name or key was required for all these commands, and by default all archives were namespaced under archive.. This could allow the archive NAME to be specified as a free argument similar to existing archive commands, instead of as a --key=KEY option.

I like the idea of forcing the namespace under `archive.`   For private job namespaces, it should probably be `data.archive` but that could be done too.

Having a default key does simplify what I imagine would be a common case, where a batch script broadcasts files using one archive.  Then it can just be
```
flux archive FILES ...
flux run -o stage-in ...
```
as opposed to
```
flux archive mykey FILES ...
flux run -o stage-in.keys=mykey ...
```
However, if that doesn't seem compelling I am fine with changing it.  I started out with a required free KEY argument so clearly I'm not offended by it :-)

Are you ok with `flux archive extract --list-only` vs having a `flux archive list`?

What about the verbosity of the subcommands?  Any better suggestions than `create` and `extract`?

garlick:
> One other thought about the default archive.main for archive keys: This seems dangerous if multiple components are creating separate archives and are using the default key. This is perhaps another argument for requiring users provide an archive name...

That's a good point...

garlick:
> This may not be necessary. The Cray shell plugin could create the archive in shell.init, and non-rank 0 shells could extract it in shell.post-init. There is a barrier between the two so order is guaranteed.

Well the `stage-in` plugin runs in `shell.init`.  Should we move it to `shell.post-init` so other plugins can set options and get it to do stuff?

I assume you're picturing the cray plugin calling out to `flux archive extract`, which would also be fine I guess.

garlick:
Oh it would be nice if `stage-in` could _create_ an archive on one rank and copy it in on the others!  The no calling out to executables would be required when a shell plugin needs to do what the Cray one wants.

grondo:
> Having a default key does simplify what I imagine would be a common case, where a batch script broadcasts files using one archive.

Ah, I can see that case being compelling. I think I'm slightly leaning towards your current approach being better than my suggestion, if we can make the KEY arguments all go under `archive` so archives can be listed?

The documentation can have a warning about multiple uses of the default name and leave it at that.

> Are you ok with `flux archive extract --list-only` vs having a `flux archive list`?

I typically don't like having options masquerading as a subcommand like that in an inerface, especially when there are other subcommands already, so the effort to implement `flux archive list` would presumably be small. In the future, if it is useful, `flux archive list` could itself get options and that would be awkward with `flux archive extract`. However, I don't feel too strongly about it and an option to just list available archives that could be extracted in `flux archive extract` does make some sense, so I leave it up to you.

garlick:
Just pushed a checkpoint with the following changes:
- added `flux archive list`
- change `-k KEY` arguments to `-n NAME` and make KVS keys relative to `archive` directory
- further simplification of code in the content module

The man page link above should be updated - LMK if you think this interface looks OK.  I'll start writing tests.

garlick:
Dropping the WIP here as I think it's getting close.

garlick:
Couple of thoughts - not sure if these would be worth doing or not
- add extract/list options to access a file archive in jobspec (by jobid)
- add extract/list options to access a file archive in a job's private name space (by jobid)

grondo:
Played around with `flux archive` a bit and it seems to work great!

A few early comments and questions before any kind of full review (apologies if any of these are naive):
- This PR drops `flux filemap` but we know of at least one user of this interface in the wild. Should we add back a wrapper that uses `flux archive` with the `--mmap` option for backwards compat?
- Similar question about the shell `stage-in` option. This one is less likely, but if a job is sitting the queue with the old `stage-in` options, will if fail at run time?
- I noticed that if `flux archive create` fails before the archive is fully created, content appears to be sent to the content store that can then never be referenced. (Or perhaps I was misreading the trace). Would it be possible to only send content when the archive will be successfully created? This one's minor.

garlick:
Thanks for those comments!  

> Should we add back a wrapper that uses flux archive with the `--mmap` option for backwards compat?

Yes I might have overdone it removing the old entirely.  I'll look at how to do that and print a deprecation warning.

> if a job is sitting the queue with the old stage-in options, will if fail at run time?

Right now `stage-in` would ignore the old options.  It should probably get the same treatment as `flux filemap` (translate + deprecation warning).

> Would it be possible to only send content when the archive will be successfully created?

That would present design challenge since it might require the set of files being archived to be traversed twice to avoid caching the content in memory, but let me think about it a bit.

grondo:
> That presents a design challenge since it might require the set of files being archived to be traversed twice to avoid caching the content in memory, but let me think about it a bit.

No problem if it presents a design challenge, it is probably not worth the effort to address at this time.

garlick:
Pushed a couple commits to restore some backwards compatibility.  I'll add tests a bit later.

garlick:
Pushed a test for the backwards compat support.

garlick:
Uh oh, one problem just occurred to me with this proposal.  Creating an archive on one node and extracting on another will _always_ require the use of the `--waitcreate` option due to the eventually consistent KVS synchronization.  But doing so makes it tricky to fail fast on an incorrect key.  Furthermore, the `stage-in` plugin doesn't even have a waitcreate option.  Hmm... pondering.

grondo:
I would actually argue in the CTI case that they create the archive in `shell.init` and extract in nonzero ranks in `shell.post-init` -- is there a synchronization problem there even though there is a shell barrier between the two? Could it work without `--waitcreate` in this case, though I suppose there is no harm to supplying it?

garlick:
> Could it work without --waitcreate in this case

I don't think so.  The issue is that the KVS root that contains the archive commit may not have reached the other ranks, since it is an open loop publish.   If `--waitcreate` isn't used then some other form of sync would be needed like obtaining the root version from the producer commit, then having the consumers call `flux_kvs_wait_version()`.  But that would require the version to be shared.  it's probably easier to just use `--waitcreate`.  And I would think it's fairly safe - once the code is working there is little chance of getting the path wrong or something.

Note that `--no-force-primary` should also be used and then the CTI files don't need to be given unique names in the KVS primary namespace.

grondo:
And just to make sure I understand, waitcreate can't be the default because it would block forever on a typo?

It does sound like this needs a note in there manpage, but otherwise doesn't sound like a showstopper to me!

garlick:
Yes exactly.

I'll add something to the man page.

grondo:
That looks great to me!

An idle thought is I wonder if the note should be featured more prominently, perhaps in a NOTES or CAVEATS section right before the EXAMPLES, then the section could be referenced from the `--waitcreate` description instead of only presented there? (I  do notice this is common practice in section 3 man pages, but ignore me if it is an inappropriate suggestion for section 1 pages!)

doc/man1/flux-archive.rst
```diff
@@ -0,0 +1,314 @@
...
+.. code-block:: console
+
+  #!/bin/bash
+
+  flux archive create --name=dataset1 -C /project dataset1
+  flux archive create --name=app --mmap -C /home/fred app
+
+  flux run -N1024 -o stage-in.keys=app,dataset1 \
+    sh -c "\$FLUX_JOB_TMPDIR/app --input=\$FLUX_JOB_TMPDIR/dataset1"
+
```
grondo, line 292: I noticed this while reviewing the `--waitcreate` note and thought I'd point it out before a full review. Since the job shell supports `{{tmpdir}}` replacement in command args, we may as well feature that here and simplify the example:

```suggestion
  flux run -N1024 -o stage-in.keys=app,dataset1 {{tmpdir}}/app --input={{tmpdir}}/dataset1
```

garlick: Great idea!  I had forgotten about that actually.

garlick:
OK, I made those changes, squashed some incremental changes, and rebased on current master.

t/Makefile.am
```diff
@@ -76,7 +76,7 @@ TESTSCRIPTS = \
 	t0012-content-sqlite.t \
 	t0024-content-s3.t \
 	t0028-content-backing-none.t \
-	t0029-filemap-cmd.t \
+	t0029-archive-mmap.t \
```
grondo, line 79: Commit message typo: Essientially

grondo (approved):
LGTM! I just noticed one typo.

garlick:
Fixed that typo and seeing MWP.  Thanks!

PR #5703, opened 2024-01-23
enable encode of pty data as base64 and make `flux alloc vi` test more reliable

### LINKED ISSUE #5696: FAIL: t2712-python-cli-alloc.t 28 - flux alloc: flux alloc vi works
This sharness test fails pretty reliably when I run `make check -j12` on my desktop, but works when run by itself:

```
FAIL: t2712-python-cli-alloc.t 28 - flux alloc: flux alloc vi works
#       
#               cat <<-EOF >input.json &&
#               [{"expect":"test\\.txt", "send":":q!
#       ", "timeout":30}]
#               EOF
#               cat <<-EOF >test.txt &&
#               test text file
#               EOF
#               $runpty -o vi.out --expect=input.json flux alloc -n1 vi test.txt &&
#               grep "test text file" vi.out
#       
```

grondo:
This PR addresses a couple issues reported in #5696.

First, evidently some pty data fails to encode using the jansson API. This PR adds support for falling back to base64 when that (very rare) event occurs. 

Second, the `flux alloc vi` test is currently dependent on the filename showing in the vi statusline, which doesn't always seem to be the case. Make the test more reliable by just looking for the file data before quitting `vi`.

Fixes #5696

### DIFF
--- a/src/common/libterminus/client.c
+++ b/src/common/libterminus/client.c
@@ -26,6 +26,7 @@
 #include "src/common/libczmqcontainers/czmq_containers.h"
 #include "src/common/libutil/fdutils.h"
 #include "src/common/libutil/llog.h"
+#include "src/common/libutil/errno_safe.h"
 #include "ccan/str/str.h"
 
 #include "pty.h"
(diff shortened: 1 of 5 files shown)


garlick (approved):
LGTM!
I see this is what we do in libioencode, so good (whew).

grondo:
Thanks! I'll set MWP

Issue #5702, opened 2024-01-23
flux-watch: TypeError: Object of type 'bytes' is not JSON serializable


grondo:
A user has reported this error and stack trace while using `flux watch`:

```
flux-watch: ERROR: Object of type 'bytes' is not JSON serializable

flux-watch: DEBUG: Traceback (most recent call last):

File "/usr/lib64/flux/python3.6/flux/util.py", line 279, in __call__

main_func()

File "/usr/libexec/flux/cmd/flux-watch.py", line 200, in main

fh.reactor_run()

File "/usr/lib64/flux/python3.6/flux/core/handle.py", line 319, in reactor_run

Flux.raise_if_exception()

File "/usr/lib64/flux/python3.6/flux/core/handle.py", line 133, in raise_if_exception

raise cls.set_exception(None) from None

File "/usr/lib64/flux/python3.6/flux/future.py", line 38, in continuation_callback

py_future.then_cb(py_future, *py_future.then_args, **py_future.then_kwargs)

File "/usr/lib64/flux/python3.6/flux/job/output.py", line 663, in _output_event_watch

self._fulfill_output(event)

File "/usr/lib64/flux/python3.6/flux/job/output.py", line 633, in _fulfill_output

self.fulfill((event.stream, event.render()))

File "/usr/lib64/flux/python3.6/flux/future.py", line 369, in fulfill

result = json.dumps(result).encode("utf-8", errors="surrogateescape")

File "/usr/lib64/python3.6/json/__init__.py", line 231, in dumps

return _default_encoder.encode(obj)

File "/usr/lib64/python3.6/json/encoder.py", line 199, in encode

chunks = self.iterencode(o, _one_shot=True)

File "/usr/lib64/python3.6/json/encoder.py", line 257, in iterencode

return _iterencode(o, 0)

File "/usr/lib64/python3.6/json/encoder.py", line 180, in defaulto.__class__.__name__)
TypeError: Object of type 'bytes' is not JSON serializable

 ```

grondo:
Ah, [base64.b64decode](https://docs.python.org/3/library/base64.html#base64.b64decode) returns `bytes` so that is perhaps the problem. Perhaps this change is needed:

```diff
diff --git a/src/bindings/python/flux/job/output.py b/src/bindings/python/flux/job/output.py
index 9130741bb..1dcb12126 100644
--- a/src/bindings/python/flux/job/output.py
+++ b/src/bindings/python/flux/job/output.py
[...]

PR #5704, opened 2024-01-24
python: fix `flux-watch: TypeError: Object of type 'bytes' is not JSON serializable`

grondo:
This PR fixes binary output with `flux-watch(1)` as discussed in #5702. A test is also added.

### DIFF
--- a/src/bindings/python/flux/job/output.py
+++ b/src/bindings/python/flux/job/output.py
@@ -113,7 +113,9 @@ class OutputEvent(EventLogEvent):
             data = self.context["data"]
             if "encoding" in self.context:
                 if self.context["encoding"] == "base64":
-                    data = base64.b64decode(data)
+                    data = base64.b64decode(data).decode(
+                        "utf-8", errors="surrogateescape"
+                    )
             if "repeat" in self.context:
                 data *= self.context["repeat"]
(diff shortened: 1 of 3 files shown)


garlick (approved):
Nice!  I'll leave it to you to decide if the CodeQL warnings amount to anything.

grondo:
>  I'll leave it to you to decide if the CodeQL warnings amount to anything.

Hm, I assumed the file handle will be closed when it has no more references, i.e. when the object containing it is destroyed. I am kind of thinking this is a false positive in this case, but I'm not sure.

It is also only an issue with Python 3.6. The `reconfigure` case doesn't open a new file handle. I wonder if the name of the method is confusing CodeQL.

grondo:
Given the documentation of the error [here](https://codeql.github.com/codeql-query-help/python/py-file-not-closed/):

> Ensure that if you open a file it is always closed on exiting the method. Wrap the code between the open() and close() functions in a with statement or use a try...finally statement. Using a with statement is preferred as it is shorter and more readable.

Since the opened file is being set to an attribute of the class, I don't see how we could close it before exiting the method and therefore this is a false positive.

I will mark as such and set MWP. Thanks!

PR #5707, opened 2024-01-26
broker: add FLUX_IPADDR_INTERFACE to select broker network interface

garlick:
Problem: during scale testing, it was observed that Flux was using an unstable network fabric, but there is no way to tell it to use another one.

Add the FLUX_IPADDR_INTERFACE environment variable.   It can be set to an interface name in the broker's environment, and then PMI bootstrap will select an address associated with that interface.

### DIFF
--- a/doc/man7/flux-environment.rst
+++ b/doc/man7/flux-environment.rst
@@ -229,6 +229,11 @@ affect the broker's PMI client.
    version 4 addresses.  Setting this variable to any value in the broker's
    environment causes it to prefer version 6 addresses.
 
+.. envvar:: FLUX_IPADDR_HOSTNAME
+
+   Force PMI bootstrap to assign the broker an address associated with a
+   particular network interface, like ``eth0``.
+
 
 CUSTOM OUTPUT FORMATS
 =====================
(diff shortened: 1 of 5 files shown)


garlick:
I added a test and removed WIP.

grondo (approved):
Nice! This was handy during scaling tests when one of the networks was unstable. So approved!

garlick:
Thanks!  I got a test failure here in one builder:
```
2024-01-29T21:31:11.7353288Z expecting success: 
2024-01-29T21:31:11.7353647Z 	FLUX_IPADDR_INTERFACE=badiface test_expect_code 137 flux start \
2024-01-29T21:31:11.7354230Z 		${ARGS} --test-exit-timeout=1s -s2 -o,-Stbon.prefertcp=1 true
2024-01-29T21:31:11.7354553Z 
2024-01-29T21:31:11.7355252Z test_expect_code: command exited with 0, we wanted 137 flux start -o,-Sbroker.rc1_path=,-Sbroker.rc3_path= --test-exit-timeout=1s -s2 -o,-Stbon.prefertcp=1 true
2024-01-29T21:31:11.7356145Z not ok 89 - FLUX_IPADDR_INTERFACE=badiface fails
```
I wasn't sure if maybe setting the environment variable before a call to a shell function was perhaps not portable (the failure was in Ubunto 20.04)?  So I pushed a small change to run the test in a subshell and export the variable within that subshell.

(Now that I describe the problem and note the OS it seems doubful that this was the issue.... there may be something else going on?  Let's see how it goes this time in CI)

garlick:
That test failure disappeared so I'll go ahead and set MWP.  If it pops up again we can deal with it.

PR #5710, opened 2024-01-29
job-ingest: fix FLUID initialization error handling to allow scaling beyond 64K brokers

garlick:
This fixes two issues
- `job-ingest` continues on when `fluid_init()` fails, potentially issuing bogus job IDs
- `job-ingest` logs an error at LOG_ERR level on an expected `fluid_init()` error on rank > 65535

These would not have been seen outside of scale testing.

The expected behavior now is that `job-ingest` will log something at debug level and exit on rank > 65535, and any other `fluid_init()` failure will be fatal to the instance.


src/modules/job-ingest/job-ingest.c
```diff
@@ -852,8 +852,18 @@ int mod_main (flux_t *h, int argc, char **argv)
         }
         flux_future_destroy (f);
         if (fluid_init (&ctx.gen, 0, fluid_get_timestamp (max_jobid) + 1) < 0) {
-            flux_log (h, LOG_ERR, "fluid_init failed");
-            errno = EINVAL;
+            // See flux-framework/flux-core#5708
+            if (rank > 65535) {
```
grondo, line 856: typo: s/65535/16384/?

garlick: Ack!  Sorry, fixing!

grondo (approved):
LGTM! Thanks!

PR #5711, opened 2024-01-29
WIP: Experimental: Add JobList constraint parser and support constraint query strings in `flux jobs -f`

grondo:
This is an experimental PR that adds a special `JobListConstraintParser` class to support a specialized constraint parser syntax for `JobList` meant to be used with `flux jobs -f, --filter`. This is built on top of @chu11's PR #5656.

At this point, this is just a proof-of-concept of how this could work to extend the `-f, --filter` option of `flux jobs` to support generalized constraint queries while being backwards compatible.

The `JobListConstraintParser` class is a two pass parser.  The first pass converts tokens without an operator to states and/or results bitmasks, using `or` to join these together. The first pass can also be used to do other convenience conversions (like `host:` or `hosts:` to `hostlist:`, or as is done in this prototype `@dt` converted to `(not (t_cleanup:<dt or t_run:>dt))` to find jobs that were running at a given time). The second pass converts the now modified query string to a JSON constraint object suitable for the `job-list` service.

Finally, a `constraint` parameter is added to the Python job listing interfaces, and `flux job list -f, --filter` is converted to pass the option argument as a constraint.

e.g.:
```
ƒ(s=5,d=1,builddir) grondo@pi3:~/git/flux-core.git$ flux jobs -f host:pi0
   ƒ4RtqYfo5 grondo   hostname   CD      1      1   0.259s pi0
   ƒ4Rtgek71 grondo   hostname   CD      1      1   0.274s pi0
ƒ(s=5,d=1,builddir) grondo@pi3:~/git/flux-core.git$ flux jobs -f host:pi[2-3]
   ƒ4Rts2f5R grondo   hostname   CD      1      1   0.303s pi2
   ƒ4Rtm6hx3 grondo   hostname   CD      1      1   0.304s pi2
   ƒ4RtnahEQ grondo   hostname   CD      1      1   0.240s pi3
   ƒ4RtfAkpf grondo   hostname   CD      1      1   0.235s pi3
```
after running some sleep jobs:
```
ƒ(s=5,d=1,builddir) grondo@pi3:~/git/flux-core.git$ flux jobs -f 'host:pi0 running'
   ƒ6gveP5Us grondo   sleep       R      1      1   3.658s pi0
   ƒ6gvYT8MV grondo   sleep       R      1      1   3.673s pi0
ƒ(s=5,d=1,builddir) grondo@pi3:~/git/flux-core.git$ flux jobs -f 'host:pi0 completed'
   ƒ4RtqYfo5 grondo   hostname   CD      1      1   0.259s pi0
   ƒ4Rtgek71 grondo   hostname   CD      1      1   0.274s pi0

```

### DIFF
--- a/src/bindings/python/flux/constraint/parser.py
+++ b/src/bindings/python/flux/constraint/parser.py
@@ -231,6 +231,16 @@ class ConstraintParser:
             to split values of that operator. For instance ``{"op": ","}``
             would autosplit operator ``op`` values on comma.
 
+        convert_values (dict): A mapping of operator name to callable which
+            should take a single list argument containing the values from
+            the obtained from the current term for the operator after the
+            operator_map and split_values operations have been applied. The
+            callable should return a new list of values. This can be used
+            to convert values to a new type or to combine multiple values
+            into a single element, e.g.  ::
+
+                convert_values = {"ints": lambda args: [int(x) for x in args]}
+
         combined_terms (set): A set of operator terms whose values can be
             combined when joined with the AND logical operator. E.g. if
             "test" is in ``combined_terms``, then
@@ -283,6 +293,10 @@ class ConstraintParser:
 
     # Combined terms
     combined_terms = set()
+
+    # Mapping of operator name to value conversion function.
+    # E.g. { "integer": lambda args: [ int(x) for x in args ] }
+    convert_values = {}
 
     def __init__(
         self, lexer=None, optimize=True, debug=False, write_tables=False, **kw_args
@@ -408,10 +422,12 @@ class ConstraintParser:
                 f"invalid character '{invalid}' in operator '{op}:'"
             )
 
+        values = [value]
         if op in self.split_values:
-            p[0] = {op: value.split(self.split_values[op])}
-        else:
-            p[0] = {op: [value]}
+            values = value.split(self.split_values[op])
+        if op in self.convert_values:
+            values = self.convert_values[op](values)
+        p[0] = {op: values}
 
     def p_quoted_token(self, p):
         """
--- a/src/bindings/python/flux/job/info.py
+++ b/src/bindings/python/flux/job/info.py
@@ -19,7 +19,7 @@ from datetime import datetime
 from itertools import chain
 
 import flux.constants
-from flux.core.inner import raw
+from flux.core.inner import ffi, raw
 from flux.job.JobID import JobID
 from flux.job.stats import JobStats
 from flux.memoized_property import memoized_property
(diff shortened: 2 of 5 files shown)


grondo:
We should decide if adding the constraint query support to `-f, --filter` is the right approach. It is more powerful, and `flux-jobs(1)` automatically gets support for the match operators supported by `job-list`, but on the other hand it is slightly less convenient in the common case, and the command already has option-based constraints like `--since`, `--user`, `--name` and `--queue` - so maybe the more consistent thing is to add `--hosts` and/or `--ranks` as @chu11 had initially proposed. For a more flexible and powerful interface, users could use `flux-pgrep(1)` which can parse the constraint query as free arguments which is slightly more convenient.

chu11:
> We should decide if adding the constraint query support to -f, --filter is the right approach. It is more powerful, and flux-jobs(1) automatically gets support for the match operators supported by job-list, but on the other hand it is slightly less convenient in the common case, and the command already has option-based constraints like --since, --user, --name and --queue - so maybe the more consistent thing is to add --hosts and/or --ranks as @chu11 had initially proposed.

IMO I think the the advanced query via `--filter` is definitely a go no matter what.  It provides the advanced query that the command line options can't provide.  Like hypothetically `user:42 and states:active OR user:43 and states:inactive`, etc.  

I think it's more of an open debate if `--hosts` and/or `--ranks` should be done.  Will those be common / popular enough to warrant am option on the command line?  I'm not sure.

likewise, with the advanced filter/query in place, dunno if we could view some current command line options worthy of retirement (will I guess hidden)?  I think `--name`, `--user`, and `--queue` are popular enough keepers.  `--since` is more debatable?

grondo:
> IMO I think the the advanced query via --filter is definitely a go no matter what. It provides the advanced query that the command line options can't provide. Like hypothetically user:42 and states:active OR user:43 and states:inactive, etc.

Agreed. However, we could provide this feature on another command (e.g. `flux pgrep`) and keep the `flux jobs` interface simple. However, we could also do both, that's fine with me.

grondo:
The issue with supporting the query via an option is that quoting can get annoying. Note that in this WIP PR, I've allowed multiple `-f, --filter` options to `flux-jobs(1)` supported with multiple use combined with `and`, e.g.

```console
$ flux jobs -f host:pi4 -f completed
   ƒ5hsApJ31 grondo   hostname   CD      1      1   0.205s pi4
$ flux jobs -f host:pi4 -f active
   ƒ5YsGD86K grondo   sleep       R      1      1   56.27s pi4
```

chu11:
> Agreed. However, we could provide this feature on another command (e.g. flux pgrep) and keep the flux jobs interface simple. However, we could also do both, that's fine with me.

Ahhh I see your point now.  Hmmm.  I think there is value in `flux pgrep` (or another tool) without all of those extra filter options possibly adding complexity / corner cases as a result.

One long term thought, what if inactive jobs are stored in a db long term.  `flux pgrep` certainly could be used, although it feels not like the right tool (only going by name and nothing else, like `sacct` vs `flux pgrep` ... maybe it would just take time to sink in).

grondo:
I've rebased this PR now that #6209 went in.

PR #5712, opened 2024-01-31
job-list: initialize queue stats

### LINKED ISSUE #5688: job-list: stats missing for configured queue
The job-list module only includes stats for queues to which at least one job has been submitted. This means no stats will appear for a configured but unused queue, except in the case where a job is submitted to a queue and moved to another queue.
After a reload, job-list doesn't update stats while processing events from the eventlog, so no stats will appear for an unused queue even in this case.

It would perhaps be better to initialize zeroed queue stats for all configured queues, so even unused queues will appear in `flux job stats` output (which is a reasonable assumption a user of this interface or the corresponding RPC might have)

chu11:
Problem: Job queue stats are generated in job-list when a job is submitted into it.  If a job queue never has a job submitted to it, no job stats exist for the queue.  This isn't what users would expect, they should expect queue stats to be 0 for the queue.

Solution: Read the flux config and initialize queue stats to 0 for any queues that have been configured.

Fixes #5688

----

Uhhh I clearly messed up my branch with rebasing somehow ... will fix

I'm not sure why a CodeQL warning showed up in this PR.  I figured the "not a warning" in original PR was enough, but I guess not??  Or maybe b/c it was due to my messed up rebasing earlier.  Anyways, I just added a tiny fix to fix it.

Oh yeah, side note.  Implementation may look a little weird.  Why have config_reload() call down into stats.[ch], why not put everything in stats.c?  I needed config reload setup for another PR, so decided to just solve this issue first along the way.

### DIFF
--- a/src/bindings/python/flux/kvs.py
+++ b/src/bindings/python/flux/kvs.py
@@ -795,6 +795,7 @@ class WatchImplementation(Future, ABC):
         try:
             super().__del__()
         except AttributeError:
+            # not an error if super did not implement
             pass
 
     def __init__(self, future_handle):
--- a/src/modules/job-list/job-list.c
+++ b/src/modules/job-list/job-list.c
@@ -107,6 +107,30 @@ static void disconnect_cb (flux_t *h,
     job_stats_disconnect (ctx->jsctx->statsctx, msg);
 }
 
+static void config_reload_cb (flux_t *h,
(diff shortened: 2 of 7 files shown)


grondo (approved):
LGTM, just noticed a couple typos in the first commit.

chu11:
thanks!  fixed, rebased, and setting MWP

Issue #5714, opened 2024-02-01
Cray MPI launches as singletons on frontier


garlick:
Problem: on Frontier, an MPI hello world program built with  `cray-mpich@8.1.23` launched as singletons, e.g.
```
$ flux run -n4 ./mpi_hello
Hello world from processor frontier08969, rank 0 out of 1 processors
Hello world from processor frontier08969, rank 0 out of 1 processors
Hello world from processor frontier08969, rank 0 out of 1 processors
Hello world from processor frontier08969, rank 0 out of 1 processors
```
Running with the following fixed the issue:
```
$ flux run --env=LD_LIBRARY_PATH=$(dirname $(flux config builtin pmi_library_path)):$LD_LIBRARY_PATH -n4 ./mpi_hello
```

But the ergonomics are not great.

One thought: I think right now we accept `-opmi=pmi2` as an alias for `-opmi=simple`. We could change that so when `-opmi=pmi2` is specified, we also modify LD_LIBRARY_PATH?

grondo:
Makes sense to me! I think we used to always insert the path to Flux's libpmi first in `LD_LIBRARY_PATH`, but I can't remember why we stopped.

Another alternative would be a shell option to specifically do this, e.g. `-o force-pmi-path` or something. That would be easy to implement in a Lua plugin, with the drawback that we're currently not able to check fo invalid or not consumed shell options, so a typo would go undetected.

garlick:
> I think we used to always insert the path to Flux's libpmi first in LD_LIBRARY_PATH, but I can't remember why we stopped.

It cause problems on El Cap where Cray's libpmi2.so needs to be found before ours, or else :crying_cat_face:   See #5041

grondo:
Well we could also prepend to LD_LIBRARY_PATH by default again, and add a plugin to flux-coral2 or on the Cray machines that removes it. (e.g. implement the workaround on the one machine that's broken, not the other way around). 

I'm just throwing ideas out there though.

garlick:
I didn't think of that!

So maybe prepend ours in the simple pmi pugin and un-prepend in the cray-pals plugin.

I think that makes sense.

garlick:
I do remember getting yelled at in mattermost for the audacity of modifying LD_LIBRARY_PATH.  However, I also think there was extreme irony in that situation. :smiley:

garlick:
[...]

PR #5715, opened 2024-02-04
pmi: prepend Flux PMI directory to LD_LIBRARY_PATH

garlick:
Problem: Flux installs its `libpmi.so` and `libpmi2.so` in a directory that is not searched by `ld.so` by default, which creates problems such as #5714 

When simple PMI is active (a requirement for Flux pmi libs to be functional), prepend this directory to LD_LIBRARY_PATH.

This will break the coral2 `cray-pals` plugin.  Or rather, it will prevent Cray MPICH from finding cray `libpmi2.so`, which will render `cray-pals` ineffectual.  So flux-framework/flux-coral2#122 should go in first.

### DIFF
--- a/src/shell/pmi/pmi.c
+++ b/src/shell/pmi/pmi.c
@@ -58,6 +58,7 @@
 #include <unistd.h>
 #include <stdlib.h>
 #include <assert.h>
+#include <libgen.h>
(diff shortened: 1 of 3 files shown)


grondo (approved):
LGTM!

Issue #5718, opened 2024-02-05
shell: support dumping hwloc XML to `HWLOC_XMLFILE`


grondo:
As noted in flux-framework/flux-pmix#31, some versions of openmpi and perhaps other applications could benefit from a copy of the shell's hwloc XML in a file, with `HWLOC_XMLFILE` set in the environment to point to that file. This could be activated by a new shell option `hwloc=xmlfile` or similar.

PR #5719, opened 2024-02-05
shell: add `shell.post-init` plugin calllback topic between `shell.init` and first `task.init`

grondo:
This simple PR just adds a new shell plugin callback topic: `shell.post-init`. This will be convenient for plugins that want to run after all other plugin `shell.init` callbacks have run, but before the first `task.init` callback (which is supposed to be task-specific).

It will be used in the solution for #5718.


garlick (approved):
LGTM!

grondo:
Thanks! I set MWP.

PR #5721, opened 2024-02-05
shell: write hwloc XML to a file with `HWLOC_XMLFILE` set with `-o hwloc.xmlfile`

### LINKED ISSUE #5718: shell: support dumping hwloc XML to `HWLOC_XMLFILE`

grondo:
This PR adds a new `hwloc` shell builtin plugin which export the shell hwloc XML to a file and sets `HWLOC_XMLFILE` if `-o hwloc.xmlfile` is used.

An hwloc object is used instead of `-o hwloc=xmlfile` to allow other keys to be open for future use, though I can't think of any other hwloc options we might want to use in the job shell at the moment.

The `hwloc.xmlfile` value is checked in `shell.post-init`, which means other plugins may request this behavior from `shell.init` or earlier by using `flux_shell_setopt_pack(3)`.

This is a WIP while I develop some simple tests. However, I wanted folks to get a chance for early comments on the general approach.

Fixes #5718

### DIFF
--- a/src/shell/hwloc.c
+++ b/src/shell/hwloc.c
@@ -1,0 +1,99 @@
+/************************************************************\
+ * Copyright 2024 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* hwloc options handler
+ */
+#define FLUX_SHELL_PLUGIN_NAME "hwloc"
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <unistd.h>
+
+#include <flux/core.h>
+#include <flux/shell.h>
+
+#include "src/common/libutil/read_all.h"
+#include "ccan/str/str.h"
+
+#include "builtins.h"
+
+static int create_xmlfile (flux_shell_t *shell)
+{
+    int fd = -1;
+    char *xmlfile = NULL;
+    const char *hwloc_xml;
+    const char *tmpdir;
+
+    if (!(tmpdir = flux_shell_getenv (shell, "FLUX_JOB_TMPDIR")))
+        tmpdir = "/tmp";
+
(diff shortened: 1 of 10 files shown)


src/shell/hwloc.c
```diff
@@ -0,0 +1,98 @@
...
+    if (asprintf (&xmlfile, "%s/hwloc.xml", tmpdir) < 0) {
+        shell_log_error ("asprintf HWLOC_XMLFILE failed");
+        goto error;
+    }
+    if ((fd = open (xmlfile, O_CREAT|O_EXCL|O_WRONLY, 0640)) < 0) {
+        shell_log_errno ("%s", xmlfile);
+        goto error;
+    }
+    shell_debug ("Writing HWLOC_XMLFILE=%s\n", xmlfile);
+    if (write_all (fd, hwloc_xml, strlen (hwloc_xml) < 0)) {
```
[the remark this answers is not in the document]
grondo: Oof, bad typo.

grondo:
Ok, I had inadvertently rebased on master thinking the `shell.post-init` changes had gone in, but they haven't yet. This should be ready for a review as there is now some nominal testing and documentation.

src/shell/hwloc.c
```diff
@@ -0,0 +1,100 @@
...
+        goto error;
+    }
+    shell_debug ("Writing %ld bytes to HWLOC_XMLFILE=%s\n",
+                 strlen (hwloc_xml),
+                 xmlfile);
+    if (write_all (fd, hwloc_xml, strlen (hwloc_xml)) < 0) {
+        shell_log_errno ("failed to write HWLOC_XMLFILE");
+        goto error;
+    }
+    close (fd);
```
garlick, line 61: Probably should check the return value of close since it can fail on a write.

garlick (approved):
LGTM, just one minor comment.

src/shell/hwloc.c
grondo: I also fixed the batch plugin which also failed to check the return code of `close(2)` (though only when using the old style batch jobspec options)

grondo:
Fixed a 32bit compile issue and force-pushed.

garlick:
Feel free to set MWP when ready!

grondo:
I was looking into why only 54% coverage but I guess there's just too much error handling. I'll set MWP.

PR #5724, opened 2024-02-09
add flux-hostlist(1)

grondo:
As described in #3344, this PR adds an implementation of a `flux hostlist` command.

This implementation takes a set of hostlist "sources" on the command line, appending multiple sources by default, and returns the result on stdout. Several options are provided to manipulate, modify, or query the result before it is emitted:

```
usage: flux-hostlist [-h] [-e] [-d S] [-c] [-n N] [-L N] [-S] [-x HOSTS]
                     [-u | -i | -m | -X] [-f] [-q]
                     [SOURCES ...]

positional arguments:
  SOURCES              (optional) One or more hostlist sources

options:
  -h, --help           show this help message and exit
  -e, --expand         Expand hostlist using defined output delimiter
  -d, --delimiter=S    Set output delimiter for expanded hostlist
                       (default=',')
  -c, --count          Print the total number of hosts
  -n, --nth=N          Output host at index N (-N to index from end)
  -L, --limit=N        Output at most N hosts (-N for the last N hosts)
  -S, --sort           Return sorted result
  -x, --exclude=HOSTS  Exclude HOSTS from final result
  -u, --union          Return the union of all hostlists. Default is to append
  -i, --intersect      Return the intersection of all hostlists
  -m, --minus          Subtract all hostlist args from first argument
  -X, --xor            Return the symmetric difference of all hostlists
  -f, --fallback       Fallback to treating jobids that are not found as
                       hostnames (for hostnames that are also valid jobids
                       e.g. f1, fuzz100, etc)
  -q, --quiet          No output. Exit with nonzero exit status if hostlist is
                       empty

SOURCES may include:
instance     hosts from the broker 'hostlist' attribute.
jobid        hosts assigned to a job
local        hosts assigned to current job if FLUX_JOB_ID is set, otherwise
             returns the 'instance' hostlist
avail[able]  instance hostlist minus those nodes down or drained
stdin, '-'   read a list of hosts on stdin
hosts        literal list of hosts

The default when no SOURCES are supplied is 'local'.
```

One basic use case is creating a hostfile for a job or batch job. Since `flux-hostlist(1)` returns a job nodelist or the hostlist of the enclosing instance if the process is not in a job by default, this is simply done with `flux hostlist -ed '\n'`:

```console
$ flux hostlist -ed '\n'
pi1
pi2
pi3
pi4
```

Since `flux-hostlist(1)` can take sets of hosts on the command line as well as other sources, it can also be used as a generic `hostlist` utility, e.g.:

```console
$ flux hostlist -e host[1-10]
host1,host2,host3,host4,host5,host6,host7,host8,host9,host10
$ flux hostlist --intersect host[1-10] host[8-20]
host[8-10]
```

Since a hostlist source can be a jobid, `flux-hostlist(1)` can also be used to easily query information about job hostlists, e.g. did all failed jobs have one node in common?

```console
$ flux hostlist --intersect $(flux jobs -f failed -no {id})
pi4
```

and so on.

One drawback to supporting a generic sources list on the command line, is that some single hostnames can look like jobids. There is a `-f, --fallback` option provided to fall back to treating unknown jobids as individual hosts to handle this case, but this could be surprising:

```console
$ flux hostlist foo1 foo2 foo3
flux-hostlist: ERROR: job foo1 not found
$ flux hostlist -f foo1 foo2 foo3
foo[1-3]
```

This PR is a WIP because documentation still needs to be added. I figured I'd wait to get agreement about the interface here before doing that.

### DIFF
--- a/doc/man1/flux-hostlist.rst
+++ b/doc/man1/flux-hostlist.rst
@@ -1,0 +1,219 @@
+.. flux-help-section: other
+
+================
+flux-hostlist(1)
+================
+
+SYNOPSIS
+========
+
+**flux** **hostlist** [*OPTIONS*] [*SOURCES*]
+
+DESCRIPTION
+===========
+
+.. program:: flux hostlist
+
+:program:`flux hostlist` takes zero or more *SOURCES* of host lists on the
+command line and concatenates them by default into a single RFC 29 Hostlist.
+
+*SOURCES* can optionally be combined by various set operations, for example
+to find the intersection, difference, or to subtract hostlists.
+
+SOURCES
+=======
+
+Valid *SOURCES* of hostlist information include:
+
+instance
+  hosts from the broker ``hostlist`` attribute
+
+jobid
+  hosts assigned to a job.
+
+local
+  *jobid* from ``FLUX_JOB_ID`` environment variable if set, otherwise
+  *instance*
+
+avail[able]
+  *instance* hostlist minus those nodes down or drained
+
+stdin or ``-``
+  read a list of hosts on stdin
+
+hosts
+  a literal RFC 29 Hostlist
+
+The default source is *local*.
+
+OPTIONS
+=======
+
+.. option:: -e, --expand
+
+  Expand hostlist result using the defined output delimiter. Default is
+  space-delimited.
+
+.. option:: -d, --delimiter=S
+
+  Set the delimiter for :option:`--expand` to string *S*.
+
+.. option:: -c, --count
+
+  Emit the number of hosts in the result hostlist instead of the hostlist
+  itself.
+
+.. option:: -n, --nth=N
+
+  Output only the host at index *N* (*-N* to index from the end). The command
+  will fail if *N* is not a valid index.
+
+.. option:: -L, --limit=N
+
+  Output at most *N* hosts (*-N* to output the last *N* hosts).
+
+.. option:: -S, --sort
+
+  Display sorted result.
+
+.. option:: -u, --union, --unique
+
+  Return only unique hosts. This implies :option:`--sort`. Without any
+  other manipulation options, this is equivalent to returning the set
+  union of all provided hosts. (By default, all inputs are concatenated).
+
+.. option:: -x, --exclude=HOSTS
+
+  Exclude all occurrences of *HOSTS* form the result.
+
+.. option:: -i, --intersect
+
+  Return the set intersection of all hostlists.
+
+.. option:: -m, --minus
+
+  Subtract all hostlists from the first.
+
+.. option:: -X, --xor
+
+  Return the symmetric difference of all hostlists.
+
+.. option:: -f, --fallback
+
+  If an argument to :command:`flux-hostlist` is a single hostname, and the
+  hostname can be interpreted as a valid Flux jobid (e.g. starts with ``f``
+  and otherwise contains valid base58 characters like ``fuzzy`` or ``foo1``),
+  then the command may fail with::
+
+    flux-hostlist: ERROR: job foo1 not found
+
+  With the :option:`--fallback` option arguments that appear to be jobids that
+  are not found are treated as hostnames, e.g.::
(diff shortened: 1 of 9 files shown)


garlick:
FWIW: this makes it easy to create a Flux batch script that launches one MPI program using the MPI implementation's native launcher, e.g.
```sh
#!/bin/sh
mpiexec.hydra -launcher ssh -hosts "$(flux hostlist -e -d' ')" mpi_hello
```

grondo:
I also updated some github actions to avoid more "update this action" warnings, but I notice codecov did not report, so I'll look into that.

garlick:
I think the interface is pretty nice.  I just played around with it a bit on my test system and everything I tried worked as expected.

grondo:
The new codecov uploader doesn't seem to work, I'm going to pull that out of here and will update actions in a separate PR.

grondo:
I removed WIP here, but still need to add a manpage assuming there were no interface objections.

grondo:
I updated this PR and added a manpage [flux-hostlist(1)](https://flux-framework--5724.org.readthedocs.build/projects/flux-core/en/5724/man1/flux-hostlist.html).

Some small changes from the previous version of `flux-hostlist`:

 - The default delimiter for `-e, --expand` is now a single space instead of  a comma. 
 - An alias for the `--union` option is `--unique` and it can be applied with other set operations to get a sorted, unique hostlist
 - `-x, --exclude` deletes all occurrences of hosts from the result, while `-m, --minus` only subtract the number of hosts in the argument from the first hostlist, e.g. `foo[1,1,1] - foo[1,1] = foo1`.

src/bindings/python/flux/util.py
```diff
@@ -123,7 +123,7 @@ def encode_topic(topic):
     return topic
 
 
-def help_formatter(argwidth=40):
+def help_formatter(argwidth=40, raw_description=False):
```
chu11, line 126: typo commit message "call o help_formatter()" -> "call to ..."

t/t2814-hostlist-cmd.t
```diff
@@ -0,0 +1,148 @@
...
+test_expect_success 'flux-hostlist preserves repeated hosts by default' '
+	test "$(flux hostlist host1 host1 host1)" = "host[1,1,1]"
+'
+test_expect_success 'flux-hostlist -S, --sort works' '
+	test "$(flux hostlist -S host3 host2 host1 host0)" = "host[0-3]"
+'
+test_expect_success 'flux-hostlist -S preserves duplicate hosts' '
+	test "$(flux hostlist -S host2 host1 host3 host1)" = "host[1,1-3]"
+'
+test_expect_success 'flux-hostlist -u, --union returns union of all hosts' '
```
chu11, line 107: --union / --unique?

doc/man1/flux-hostlist.rst
```diff
@@ -0,0 +1,198 @@
+.. flux-help-section: other
+
+================
+flux-hostlist(1)
+================
```
chu11, line 5: perhaps should document exit value?  I noticed in a test or two of exit 1 vs 0, wasn't sure what the rules are (haven't looked at tool code yet)

garlick (approved):
This looks great!  The man page is excellent with all the examples.  I didn't spot anything out of whack - I say put it in!

t/t2814-hostlist-cmd.t
grondo: Eh, I'll just use `-u` in the description since it is obvious `-u, --union, --unique` are aliases of the same option and we're testing the option behavior not argparse behavior.

doc/man1/flux-hostlist.rst
grondo: Oh good suggestion.

grondo:
I pushed changes based on suggestions from @chu11

chu11 (approved):
LGTM too

PR #5730, opened 2024-02-10
shell: generate fatal error if `block` taskmap scheme has an argument

### LINKED ISSUE #5723: shell taskmap `block` scheme ignores its arguments
A user was trying to specify a manual taskmap with `--taskmap`, but inadvertently used `--taskmap=block:TASKMAP`, and their taskmap was ignored. The `block` distribution handler should generate an error or warning if it is given a value.

grondo:
This simple PR fixes #5723: If a user inadvertently specifies `--taskmap=block:arg` when they meant `--taskmap=other_scheme:arg`, the shell currently silently ignores `arg` and silently continues with the `block` scheme. Since an argument is not expected or allowed with the `block` taskmap scheme, the shell should generate an error in this case. 

Fixes #5723

### DIFF
--- a/src/shell/shell.c
+++ b/src/shell/shell.c
@@ -1337,9 +1337,17 @@ static int shell_taskmap (flux_shell_t *shell)
         shell_log_error ("failed to parse taskmap shell option");
         return -1;
     }
-    if (rc == 0
-        || streq (scheme, "block"))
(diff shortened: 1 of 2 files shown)


garlick (approved):
LGTM!

grondo:
Thanks!

PR #5732, opened 2024-02-12
shell: make kvs output limit configurable and default single-user jobs to unlimited

grondo:
This PR adds support for a new `output.limit` job shell option which overrides the default KVS output limit.

The limit remains `10M` for multiuser instance jobs (when broker userid != job shell userid), but the limit is removed for single-user jobs (broker userid == shell userid).

This should resolve some of the issues users have hit with the KVS output limit inside batch jobs, where it is most convenient to have jobs output folded in with that of the batch output file.

### DIFF
--- a/doc/man1/common/job-shell-options.rst
+++ b/doc/man1/common/job-shell-options.rst
@@ -39,3 +39,8 @@
 
    * - :option:`hwloc.xmlfile`
      - Write hwloc XML gathered by job to a file and set ``HWLOC_XMLFILE``
+
+   * - :option:`output.limit`
+     - Set KVS output limit to SIZE bytes, where SIZE may be a floating point
+       value including optional SI units: k, K, M, G. This value is ignored
+       if output is directed to a file with :option:`--output`.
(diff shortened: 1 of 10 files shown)


garlick (approved):
LGTM

There is an extra "is" in the first sentence of last paragraph of the commit message of the penultimate commit.

Any way to add a test to check that the default 10MB limit is still there for guests (or did I miss that there is one)?

grondo:
> Any way to add a test to check that the default 10MB limit is still there for guests (or did I miss that there is one)?

Yes, sorry that was missed. I've added a `system` test that ensures output is still truncated for guest jobs.

grondo:
The 'system' instance container that runs systemd under docker has apparently stopped working. We may be hitting this issue:

 https://github.com/moby/moby/issues/42275

Though I'm not sure exactly when the underlying version of systemd would have been updated, since the system tests were working as of last Friday afaict. Anyway, the symptoms are the same as in that issue.

As suggested, the tests do work under podman, so maybe that is one path forward if available on Github actions. I'll move this discussion to a separate issue, but until we figure this out we may have to disable the requirement for the `system` ci checks to pass.

grondo:
Ok, I figured out why the system image tests were failing (my fault). I'll set MWP here now.

PR #5742, opened 2024-02-14
do not drain ranks when job is canceled during prolog

### LINKED ISSUE #5734: nodes are drained when a user aborts a run request with prolog running
Problem: if a user gets impatient while running `flux alloc` or `flux run` and aborts the command while the prolog is running, that causes a SIGTERM to be sent to the prolog which causes it to fail, which drains the node.

Test setup: add a sleep 5 to prolog script
```
$ flux run hostname
^Cflux-job: one more ctrl-C within 2s to cancel or ctrl-Z to detach    00:00:03
^C4.247s: job.exception type=cancel severity=0 interrupted by ctrl-C   00:00:04
4.354s: job.exception type=prolog severity=0 prolog killed by signal 15 (timeout or job canceled)

$ flux resource drain
Feb12 18:20  drained  prolog failed for jobid ƒkSer+ picl0
```

A workaround would be to trap SIGTERM in the prolog script.

Edit: which would defeat the prolog timeout, if one is configured, e.g.
```
[job-manager.prolog]
command = [
   "--sdexec",
   "--timeout=10s",
   "--with-imp",
   "-e", "prolog"
]
```

grondo:
This PR fixes #5734 and adds a test for the issue to the testsuite.

### DIFF
--- a/src/cmd/flux-perilog-run.py
+++ b/src/cmd/flux-perilog-run.py
@@ -175,8 +175,14 @@ async def run_per_rank(name, jobid, args):
         if proc.canceled:
             timeout_ids.set(rank)
             rc = 128 + signal.SIGTERM
-        elif rc != 0:
+        elif rc > 0 and rc <= 128:
+            #  process failed with non-zero exit code. Add this rank to
+            #  the failed set which will be drained.
             fail_ids.set(rank)
+        else:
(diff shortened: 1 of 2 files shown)


garlick (approved):
LGTM!

grondo:
Thanks!

PR #5744, opened 2024-02-16
flux-jobs: support `{id.f58plain}` output format field

grondo:
`id.f58plain` is not available as an output format field in `flux-jobs(1)` and related utils only because the name is missing from a couple dicts. Add it where necessary so that `{id.f58plain}` can be used in output formats.

This is a prereq for a small bash completions improvement that will be in a follow-on PR.


chu11 (approved):
LGTM

Issue #5738, opened 2024-02-14
make all jobs "waitable"


garlick:
Problem: one must submit a job with `--flags=waitable` in order to use `flux job wait` on it.

Originally, jobs were removed from the job manager upon transition to inactive but kept in the KVS.  The waitable flag caused them to be kept in the job-manager until they they were reaped with `flux job wait`.  Since that impacted the job manager memory footprint, it was neither the default, nor permitted for guests.

But now, the job manager keeps inactive jobs in memory and only removes them when they are purged from the system so the waitable flag has no impact on the job manager memory footprint.  To improve the usability of `flux job wait` we might want to make all jobs waitable by default.

In addition, we should consider whether guests should be permitted to wait on the jobs they own.

garlick:
This was implemented in #5748 but deemed not really worth the extra complexity.

PR #5748, opened 2024-02-21
WIP: make all jobs "waitable"

garlick:
Problem: as described in #5738, requring jobs to be submitted with the `waitable` flag in order for them to work with `flux_job_wait(3)` is not really buying us much anymore, now that inactive jobs are kept in the job manager's memory by default.

This makes all jobs waitable by default.  Instance owner only, as before.

I marked this WIP for now pending
- documentation update
- need to make sure that unwaited-for jobs can be purged (and add a test)
- check performance/memory impact and mitigate if necessary
- consider whether wait should be made into a jobtap plugin instead of a builtin job manager service
- consider whether it is worth the effort to allow guests to wait on their own jobs

### DIFF
--- a/src/modules/job-ingest/job.c
+++ b/src/modules/job-ingest/job.c
@@ -100,12 +100,6 @@ struct job *job_create_from_request (const flux_msg_t *msg,
                    FLUX_JOB_URGENCY_DEFAULT);
         goto inval;
     }
-    if (!(job->cred.rolemask & FLUX_ROLE_OWNER)
-        && (job->flags & FLUX_JOB_WAITABLE)) {
-        errprintf (error,
-                   "only the instance owner can submit with FLUX_JOB_WAITABLE");
-        goto inval;
-    }
     /* Validate jobspec signature, and unwrap(J) -> jobspec_str, _strsize.
      * Userid claimed by signature must match authenticated job->cred.userid.
      * If not the instance owner, a strong signature is required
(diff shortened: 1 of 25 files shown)


garlick:
A brief pause to measure the impact of always adding unreaped jobs to the zombie hash when they transition to inactive, as opposed to only adding jobs that have the WAITABLE flag.  It seems that throughput is roughly the same.  This is a single node test out of the build tree on my desktop, after running three groups of 1000 jobs through first and letting the system quiet down:

Master (v0.59.0-77-g102eb148f):
```
ƒ(s=1,d=0,builddir) garlick@system76-pc:~/proj/flux-core$ src/test/throughput.py -n 10000
number of jobs: 10000
submit time:    3.106 s (3219.8 job/s)
script runtime: 41.173s
job runtime:    40.931s
throughput:     244.3 job/s (script: 242.9 job/s)
```

This PR
```
ƒ(s=1,d=0,builddir) garlick@system76-pc:~/proj/flux-core$ src/test/throughput.py -n 10000
number of jobs: 10000
submit time:    2.947 s (3393.0 job/s)
script runtime: 40.592s
job runtime:    40.461s
throughput:     247.2 job/s (script: 246.4 job/s)
```

garlick:
My feeling at this point is that this change doesn't accomplish much, since:
- there's not really a use case for consumable wait status in the system instance
- the flag prevents some extra work from being done and thus has a useful purpose

I considered expanding the "result" to be teh same as `flux_job_result(3)` but as @grondo pointed out, if you want that you can trivially fetch it after `flux_job_wait(3)` returns.

I considered reimplementing wait as a jobtap plugin but the benefits of doing that are marginal at best.

So yeah let's close this.  Sorry for the noise.

PR #5749, opened 2024-02-22
flux-job: add shorthand paths for `flux job eventlog` and `wait-event`

grondo:
This PR is a simple update to add some shorthand names for well-known eventlogs wih `flux job wait-event` and `flux job eventlog`. It is a bit cumbersome to always be typing `flux job eventlog -p guest.exec.eventlog`, and after this PR you can just do `flux job eventlog -p exec ...`.


grondo:
Thanks! Rebased and I'll set MWP.

Issue #5750, opened 2024-02-26
flux-archive: failures in GitLab CI for new command test


wihobbs:
TLDR: t0021 tests 13, 14, and 18 are failing on LC
```
(s=1,d=0) fluxci@tioga11 /usr/WS1/fluxci/cibuilds/383675_corona/flux-core/build/t (master)$ ../../t/t0021-archive-cmd.t --debug --logfile --verbose
sharness: loading extensions from /usr/WS1/fluxci/cibuilds/383675_corona/flux-core/t/sharness.d/01-setup.sh
sharness: loading extensions from /usr/WS1/fluxci/cibuilds/383675_corona/flux-core/t/sharness.d/flux-sharness.sh
****************************************************************************
* hwloc 2.10.0 received invalid information from the operating system.
*
* Failed with error: intersection without inclusion
* while inserting Group0 (groupkind 1000-0 cpuset 0x00ff0000,0x0)
* at Package (P#0 cpuset 0xffffffff,0xffffffff nodeset 0x0000000f)
* coming from: topology:io_parent
*
* The following FAQ entry in the hwloc documentation may help:
*   What should I do when hwloc reports "operating system" warnings?
* Otherwise please report this error message to the hwloc user's mailing list,
* along with the files generated by the hwloc-gather-topology script.
*
* hwloc will now ignore this invalid topology information and continue.
****************************************************************************
****************************************************************************
* hwloc 2.10.0 received invalid information from the operating system.
*
* Failed with error: intersection without inclusion
* while inserting Group0 (groupkind 1000-0 cpuset 0x00ff0000,0x0)
* at Package (P#0 cpuset 0xffffffff,0xffffffff nodeset 0x0000000f)
* coming from: topology:io_parent
*
* The following FAQ entry in the hwloc documentation may help:
*   What should I do when hwloc reports "operating system" warnings?
* Otherwise please report this error message to the hwloc user's mailing list,
* along with the files generated by the hwloc-gather-topology script.
*
* hwloc will now ignore this invalid topology information and continue.
****************************************************************************
sharness: loading extensions from /usr/WS1/fluxci/cibuilds/383675_corona/flux-core/t/sharness.d/01-setup.sh
sharness: loading extensions from /usr/WS1/fluxci/cibuilds/383675_corona/flux-core/t/sharness.d/flux-sharness.sh
expecting success:
	test_must_fail flux archive create --badopt 2>badopt.err &&
	grep "unrecognized option" badopt.err

[...]

PR #5751, opened 2024-02-26
testsuite: fix random-generation of kvs archive files in t0021

wihobbs:
Problem: the randbytes() function in t0021-archive-cmd.t uses /dev/random to create files of size 128 and 2048 bytes, respectively. However, /dev/random can be truncated at <128 bytes and, when it is, the file created by dd will be as well. This causes a size mismatch in the kvs, which causes tests 13, 14, and 18 to fail. This first showed up in GitLab CI on TOSS 4.

We can instead utilize the /dev/urandom file, which is not truncated, and thus guaranteed to create files of the proper sizes. Also, some cleanup of t0021-archive-cmd.t is baked into this PR.

When merged, closes issue #5750.

### DIFF
--- a/t/t0021-archive-cmd.t
+++ b/t/t0021-archive-cmd.t
@@ -2,8 +2,6 @@
 
 test_description='Test flux-archive'
 
-. `dirname $0`/content/content-helper.sh
-
 . `dirname $0`/sharness.sh
 
 test_under_flux 2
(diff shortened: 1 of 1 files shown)


garlick (approved):
LGTM!  Thanks!

wihobbs:
Thanks @garlick! Set MWP.

PR #5756, opened 2024-02-27
shell: fix incorrect values returned from `flux_shell_get_rank_info(3)`

grondo:
Embarrassingly, the values returned for non-local shell ranks from `flux_shell_get_rank_info(3)` and associated calls has been wrong. The local values are always returned even if the "info" for a different shell rank is requested.

This PR fixes the object constructed for this and associated calls to return the correct information, and includes what was obviously a missing test.


garlick (approved):
LGTM!

grondo:
Thanks! Setting MWP

PR #5757, opened 2024-02-27
testsuite: fix t2617 to utilize /dev/urandom

wihobbs:
Oops. Yesterday, when we were fixing the `/dev/random` file creations, we missed one. That's my bad.

I ran a `git grep "/dev/random"` to make sure we got them all. Should be good after this.


garlick (approved):
Thanks - sorry I didn't catch that either.

wihobbs:
No worries! Maybe this will be the kick in the pants I need to make the GitLab error reporting clearer like @grondo has been telling me to for weeks... 😅 

Setting MWP.

PR #5758, opened 2024-02-28
flux-job: attach: support MPIR tool launch extension

grondo:
This PR adds the minimum support for the MPIR tool daemon launch extensions documented in [The MPIR Process Acquisition Interface Version 1.1 ](https://www.mpi-forum.org/docs/mpir-specification-03-01-2018.pdf), Section 7.1. It also splits the MPIR code in `src/cmd/job/attach.c` into its own source file so that the implementation can be tested with the "shell" mpir test utility in `t/shell/mpir.c`.

Currently the tool specified in `MPIR_executable_path` and `MPIR_server_arguments` is launched with libsubprocess. We depend on the disconnect handler to clean up stray tool processes if `flux job attach` is terminated or otherwise exits early.

This is currently a WIP because it adds a list of task/broker rank pairs to the shell `proctable` response since this was the simplest approach for now, but there's probably a more efficient way.

This PR is based on top of #5756

### DIFF
--- a/src/cmd/job/mpir.c
+++ b/src/cmd/job/mpir.c
@@ -13,8 +13,8 @@
 #if HAVE_CONFIG_H
 #include "config.h"
 #endif
-#include <errno.h>
 #include <unistd.h>
+#include <errno.h>
 #include <signal.h>
 
 #include <flux/core.h>
@@ -25,6 +25,8 @@
 #include "src/common/libdebugged/debugged.h"
 #include "src/shell/mpir/proctable.h"
 #include "ccan/str/str.h"
+
+extern char **environ;
 
 #ifndef VOLATILE
 # if defined(__STDC__) || defined(__cplusplus)
@@ -47,6 +49,9 @@ int MPIR_i_am_starter            = 1;
 int MPIR_acquired_pre_main       = 1;
 int MPIR_force_to_main           = 1;
 int MPIR_partial_attach_ok       = 1;
+
+char MPIR_executable_path[256];
+char MPIR_server_arguments[1024];
 
 static void setup_mpir_proctable (const char *s)
 {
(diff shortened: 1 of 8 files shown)


garlick:
Mentioning @lee218llnl since he may be interested or have a perspective on this part of the MPIR spec.

grondo:
I have an improvement for the MPIR rangelist representation that I think makes addition of the broker ranks to the `proctable` response reasonable. For example, this is the JSON representation of the proctable entries for 3072 processes running across 64 nodes on corona:
```json
{"hosts":[["corona",[[187,-47],[13,-47],[1,-47,1],[2,-47],[1,-47,2],[6,-47],[1,-47,1],[2,-47],[1,-47,4],[2,-47],[1,-47,34],[2,-47],
[1,-47,9]]]],"executables":[["sleep",[[-1,-3071]]]],"ids":[[0,3071]],"pids":[[1535168,47],[-535966,47],[-228557,47],
[686167,47],[-622563,47],[-129439,47],[2866245,47],[-1739757,47],[220055,47],[-428264,47],[217894,47],[48174,47],
[-56983,47],[-157167,47],[-242878,47],[348002,47],[141991,47],[-1051847,47],[853639,47],[279800,47],[-315681,47],
[-1396323,47],[-1368,47],[944,47],[-24426,47],[139,47],[-112,47],[-133,47],[-264,47],[141,47],[-19694,47],[19383,47],
[-3497,47],[548,47],[-5766,47],[-171,47],[-344,47],[1131,47],[-3202,47],[-243,47],[104,47],[39,47],[-98,47],[224,47],
[-186,47],[53,47],[-393,47],[98,47],[-1593,47],[-378,47],[97,47],[-113,47],[101,47],[-69,47],[-53,47],[214,47],[-369,47],
[149,47],[-369,47],[-1869,47],[-84,47,1],[-62,47],[-133,47]],"ranks":[[0,-47],[1,-47,62]]}
```
The additional overhead is the `"ranks":[[0,-47],[1,-47,62]]` bit.

garlick:
Just a quick meta question - would it make sense to have a libmpir convenience library that concentrates this stuff and any unit tests?

grondo:
> Just a quick meta question - would it make sense to have a libmpir convenience library that concentrates this stuff and any unit tests?

There are already unit tests in `src/shell/mpir`, for the testing of the `MPIR_proctable` and other MPIR symbols, a unit test will not suffice since it requires a running job shell. Or were you wanting the mpir code moved out of the shell to `src/common` and combined with the MPIR support from `flux-job` somehow -- I'm not sure it makes sense to always have the `MPIR_proctable` and other MPIR symbols linked in to the shell, as this might confuse debuggers. This is the main reason I've shied away from that approach for now.

There is also the small `libdebugged` library that is used to give access to `MPIR_being_debugged` from Python... maybe they could all be combined into a single thing? I'm kind of afraid it might be more messy to do it that way -- perhaps save it for another day?

garlick:
Sure, I was just asking.  It sort of worked out with the PMI stuff but good point about the special symbols.

grondo:
Yeah, we should revisit it perhaps in the future when not pressed for time. Maybe I'm overthinking the symbols...

src/shell/mpir/rangelist.c
```diff
@@ -25,16 +25,20 @@
  *   which is delta encoded from the previous entry (or 0 if this is the
```
garlick, line 25: In "shell: mpir: compress repeating ranges in rangelist" commit message:  s/commoon/common/
Also is that space in one of the ranges supposed to be there?

src/shell/mpir/rangelist.c
```diff
@@ -254,6 +258,44 @@ json_t *range_to_json (struct range *r, int64_t base)
...
+}
+
+static int increment_range_repeat (json_t *range)
+{
+    if (json_array_size (range) == 3) {
+        json_int_t repeat = range_json_val (range, 2);
+        return json_array_set_new (range, 2, json_integer (repeat + 1));
+    }
+    return json_array_append_new (range, json_integer (1));
+}
```
garlick, line 281: If either of the `_new()` functions fails, the `json_integer()` value is leaked.

src/cmd/job/mpir.c
```diff
@@ -74,6 +79,160 @@ static void gen_attach_signal (flux_t *h, flux_jobid_t id)
...
+
+    rank = idset_first (ranks);
+    while (rank != IDSET_INVALID_ID) {
+        flux_subprocess_t *p;
+        if (!(p = flux_rexec (h, rank, 0, cmd, &ops)))
+            log_err ("MPIR: failed to launch %s", tool_path);
+        rank = idset_next (ranks, rank);
+    }
+    flux_cmd_destroy (cmd);
+    return;
```
garlick, line 233: Are the subprocess objects leaked here purposefully because there is no way to return an error via the MPIR interface?

t/shell/mpir.c
```diff
@@ -15,6 +15,7 @@
 #endif
 
 #include <flux/core.h>
+#include <flux/optparse.h>
```
garlick, line 18: In "testsuite: switch mpir test program to optparse " commit message, s/--leade-rank/leader-rank/

garlick (approved):
Just a few minor typos and a question for clarification.  This is great to have done!

grondo:
Rebased on master. Use shell rexec service to launch tool daemons.

grondo:
This is approved, but I think there's an open question about how the subprocess objects are managed within the mpir module of `flux job attach`. 

Also, while the code itself is tested via the testsuite, there is no real-world debugger or tool that uses `MPIR_executable_path` and `MPIR_server_arguments` against which the PR can be tested (DDT supposedly does, but I was unable to force it to work with `flux job attach`), so whether the current approach will actually work in practice is an unknown.

Therefore, I'm unsure whether this PR should be merged, or just closed and resurrected when there is a real need.

Since the code is only activated when a debugger or tool has attached to `flux job attach` (`MPIR_being_debugged = 1`), _and_ the tool has writted to `MPIR_executable_path`, the risk of merging the code is probably low.

garlick:
It would be a shame to lose this work in case it is needed later (for example if DDT uses it).  My vote is we merge it.

Later, we should perhaps find somewhere to document the specific "tool support" features we have in flux-core so future people wanting to port a tool have someplace to start.

grondo:
Great idea! Maybe the debugging docs in flux-docs should be moved into flux-core and the MPIR support mechanism details added to that section.

Ok, I'll set MWP here since there's low probability of anything bad happening as a result.

PR #5759, opened 2024-02-28
shell: unset `HWLOC_COMPONENTS` with `hwloc.xmlfile`

### LINKED ISSUE #5755: shell: hwloc.xmlfile has no effect if HWLOC_COMPONENTS is set
Problem: the shell `-o hwloc.xmlfile` option won't work if HWLOC_COMPONENTS is set.

HWLOC_COMPONENTS prevents problematic hwloc plugins from being loaded when the topology is scanned.  But when HWLOC_XMLFILE is set, the scan is bypassed.

So could we just unset HWLOC_COMPONENTS in the `hwloc` shell plugin when `-o hwloc.xmlfile` is specified?

grondo:
Fixes #5755

### DIFF
--- a/doc/man1/common/job-shell-options.rst
+++ b/doc/man1/common/job-shell-options.rst
@@ -38,7 +38,10 @@
        nonzero exit code.
 
    * - :option:`hwloc.xmlfile`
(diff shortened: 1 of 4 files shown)


garlick (approved):
Thanks!

PR #5762, opened 2024-03-01
broker: catch an improper use of groups and handle it gracefully

garlick:
Problem: `groups.join` and `groups.leave` RPCs do not behave as designed when used remotely, but this is allowed by the RPC handlers.
    
Respond with an error if the join/leave request is not of local origin.


grondo (approved):
Good catch! LGTM!

PR #5765, opened 2024-03-01
add `flux job hostpids` command

grondo:
This PR adds a `flux job hostpids` command which uses the job shell's `proctable` RPC to emit the `host:PID` pairs for all tasks in a job.

This is useful for tools that take this information on the command line and do not support Flux natively.

### DIFF
--- a/src/shell/mpir/rangelist.c
+++ b/src/shell/mpir/rangelist.c
@@ -25,9 +25,9 @@
  *   which is delta encoded from the previous entry (or 0 if this is the
  *   first entry in the rangelist).
  *
- * - If an entry is an array, it will have two elements. The fist element
- *   is delta encoded from the previous entry (or 0 if no previous entry)
- *   and represents the start value for a set of integers.
+ * - If an entry is an array, it will have two or three elements. The first
+ *   element is delta encoded from the previous entry (or 0 if no previous
+ *   entry) and represents the start value for a set of integers.
  *
  *   - if the second element is > 0, then it represents a run-length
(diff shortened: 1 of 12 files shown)


grondo:
Some of the work from #5758 was split off here since there was at least one open question about the implementation in that PR,  and it still needs to be modified to use the shell rexec service, and the MPIR tool daemon launch extensions don't have an _actual_ use case, whereas `flux job hostpids` does...

src/shell/mpir/rangelist.c
```diff
@@ -25,16 +25,20 @@
  *   which is delta encoded from the previous entry (or 0 if this is the
  *   first entry in the rangelist).
  *
- * - If an entry is an array, it will have two elements. The fist element
```
garlick, line 28: Same comment as before re: the commit message
- "commoon"
- extra space in the last bracketed element of the first example

garlick (approved):
LGTM!  Just a couple of things in the first commit that I called out before.

src/shell/mpir/rangelist.c
grondo: uhoh, I thought I had rebased after fixing those issues.. I wonder what else I missed :facepalm: ugh

grondo:
Let me attempt to rebase on the right version of the other branch and I'll repush.

grondo:
Ok, apologies for that. I found the correct version of #5758 and pulled in the fixed commits from there. If/when this goes in I'll rebase #5758.

garlick:
Looks good and already approved.

grondo:
Thanks! I'll set MWP.

PR #5766, opened 2024-03-04
shell: support opening output files with `-o output.mode=append`

grondo:
This PR adds a new job shell output option `output.mode`. The valid values are "truncate" (the default) and "append", which opens any output files with `O_APPEND` instead of `O_TRUNC`.


garlick (approved):
That turned out nice!  LGTM!

grondo:
Thanks! I've set MWP.

PR #5769, opened 2024-03-05
testsuite: fix lingering processes after `make check`

grondo:
This PR attempts to fix or workaround an issue that I keep seeing: A couple orphaned instances that stick around after `make -j N check`. A quick investigation showed that the instances all had a sleep job stuck in CLEANUP, with the scheduler unloaded, and had at least instance-level 1 (so they were batch jobs run within the test instance).

I wasn't really able to make a reproducer for this issue, but terminating the affected instances more cleanly seems to resolve the problem (at least in my testing), so it seems like a good idea to do that.

### DIFF
--- a/t/t2800-jobs-recursive.t
+++ b/t/t2800-jobs-recursive.t
@@ -104,4 +104,8 @@ test_expect_success 'flux jobs --json works with recursive jobs' '
 test_expect_success FLUX_SECURITY 'cancel alternate user job' '
 	flux cancel $(cat altid)
 '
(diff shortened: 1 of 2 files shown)


garlick (approved):
OK with me!

I hadn't noticed this on my system, FWIW.

grondo:
Well this change is causing hangs in CI now, so obviously this wasn't a sufficient or good workaround. More analysis needed.

grondo:
Ok. The hang was from a use of `flux proxy` that was prompting for a password. Fixed and will set MWP

Issue #5752, opened 2024-02-26
excessive log noise when a job is canceled


garlick:
Problem: on el cap just noticed that a canceled 660 node, 10560 task job generated almost 800 lines of logs like the following to the circular buffer. 

```
2024-02-26T21:12:04.630225Z job-manager.err[0]: ffXiU8qLjsm: prolog: stderr: Exception ignored when trying to write to the signal wakeup fd:
2024-02-26T21:12:04.630275Z job-manager.err[0]: ffXiU8qLjsm: prolog: stderr: BlockingIOError: [Errno 11] Resource temporarily unavailable
2024-02-26T21:12:04.630288Z job-manager.err[0]: ffXiU8qLjsm: prolog: stderr: Exception ignored when trying to write to the signal wakeup fd:
2024-02-26T21:12:04.630297Z job-manager.err[0]: ffXiU8qLjsm: prolog: stderr: BlockingIOError: [Errno 11] Resource temporarily unavailable
[snip]
2024-02-26T21:16:16.053158Z job-exec.debug[0]: exec_kill: ffXiU8qLjsm: signal 15
2024-02-26T21:16:16.078972Z job-exec.debug[0]: exec_kill: ffXiU8qLjsm: signal 15
2024-02-26T21:16:16.108067Z job-exec.debug[0]: exec_kill: ffXiU8qLjsm: signal 15
```

grondo:
Seems to be a Python asyncio issue with lots of subprocesses: https://stackoverflow.com/questions/52376095/blockingioerror-exception-ignored-should-i-be-concerned

grondo:
`flux-perilog-run` has to run every job prolog in a separate `flux exec` subprocess in order to track which failed. I wonder if a better way would be to have a mode in `flux exec` where it runs as a helper and reports exit status as structured data (json?) on stdout so that `flux-perilog-run` could run _one_ `flux exec`.

Or we could just talk to the libsubprocess protocol directly from Python.

grondo:
We should set:
```toml
[job-manager.perilog]
log-ignore = [ "BlockingIOError:", "Exception ignored.*signal wakeup"]
```

In the Flux config toml on this and other Flux clusters when v0.60.0 is installed.

grondo:
This has since been resolved via the `log-ignore` config key.

PR #5772, opened 2024-03-06
job-manager: support suppression of prolog/epilog output with `perilog.log-ignore` pattern list

grondo:
On large systems, the `flux-perilog-run` script is generating hundreds of lines of useless output from an asyncio issue (see #5752). Additionally, there could be lots of other output or errors from the prolog/epilog that should be suppressed.

This PR introduces a new configuration list `job-manager.perilog.log-ignore` that contains a list of regexes to suppress from the logs for the prolog and epilog.


garlick (approved):
LGTM!

grondo:
Thanks! I've set MWP.

PR #5779, opened 2024-03-08
docker: add script help message for macOS

wihobbs:
Recently, several core developers have noted that the docker-run-checks.sh script borks when command-line options that specify values do so with an equals sign rather than a space. Since this is not super intuitive, update the usage/help message for that script accordingly.

I pulled out all the = signs from the usage message and added a note, but maybe one or the other is more desirable.

### DIFF
--- a/src/test/docker/docker-run-checks.sh
+++ b/src/test/docker/docker-run-checks.sh
@@ -27,7 +27,7 @@ die() { echo -e "$prog: $@"; exit 1; }
 #
 declare -r long_opts="help,quiet,interactive,image:,flux-security-version:,jobs:,no-cache,no-home,distcheck,tag:,build-directory:,install-only,no-poison,recheck,unit-test-only,quick-check,inception,platform:,workdir:,system"
 declare -r short_opts="hqIdi:S:j:t:D:Prup:"
-declare -r usage="
+declare usage="
 Usage: $prog [OPTIONS] -- [CONFIGURE_ARGS...]\n\
 Build docker image for CI builds, then run tests inside the new\n\
 container as the current user and group.\n\
@@ -59,9 +59,13 @@ Options:\n\
 "
 
 # check if running in OSX
-if [[ "$(uname)" == "Darwin" ]]; then
+if [[ "$(uname)" == "Darwin" ]] && [[ $FORCE_GNU_GETOPT != 1 ]]; then
     # BSD getopt
     GETOPTS=`getopt $short_opts -- $*`
+    usage=${usage}"\n\
+    You are using BSD getopt on macOS. BSD getopt does not recognize '='\n\
+    between options. Use a space instead. If gnu-getopt is first in your\n\
+    PATH, force the script to use that by setting FORCE_GNU_GETOPT=1.\n"
 else
     # GNU getopt
     GETOPTS=`getopt -u -o $short_opts -l $long_opts -n $prog -- $@`


wihobbs:
I should note that I was working on a patch as @trws recommended in Slack that would make strip `=` from options in lines 62-70 of the docker-run-checks.sh script. Maybe that's more desirable behavior than this...

src/test/docker/docker-run-checks.sh
```diff
@@ -42,20 +42,22 @@ Options:\n\
      --inception               Run tests as flux jobs\n\
      --system                  Run under system instance\n\
  -q, --quiet                   Add --quiet to docker-build\n\
- -t, --tag=TAG                 If checks succeed, tag image as NAME\n\
- -i, --image=NAME              Use base docker image NAME (default=$IMAGE)\n\
- -p, --platform=NAME           Run on alternate platform (if supported)\n\
- -S, --flux-security-version=N Install flux-security vers N (default=$FLUX_SECURITY_VERSION)\n
- -j, --jobs=N                  Value for make -j (default=$JOBS)\n
+ -t, --tag TAG                 If checks succeed, tag image as NAME\n\
```
grondo, line 45: Thanks!

The commit message should be updated to use the `Problem:` statement form and wrapped at 72 chars.

Maybe we should move the test:
```sh
# check if running in OSX
if [[ "$(uname)" == "Darwin" ]]; then
    # BSD getopt
```

to above where the help string is generated and make the use of `=` and MacOS warning conditional on that? 

Another idea would be to detect BSD getopt and suggest installing GNU getopt instead.

These are just suggestions besides the commit message changes.

grondo:
> I should note that I was working on a patch (but haven't finished yet) as @trws recommended in Slack that would make strip = from options in lines 62-70 of the docker-run-checks.sh script. Maybe that's more desirable behavior than this...

Not a bad idea! I'm guessing that would work for either GNU or BSD getopt?

grondo:
Also the prefix should probably be `docker:` or `test:` instead of `scripts:`, though it doesn't really matter that much

wihobbs:
> Not a bad idea! I'm guessing that would work for either GNU or BSD getopt?

My plan was to add a conditional under this line:
```
# check if running in OSX
if [[ "$(uname)" == "Darwin" ]]; then
    # BSD getopt
```
to strip out all of the equals signs in `$*` only on macOS before passing them to getopt. I'm assuming this would work with GNU getopt as well as BSD. Adding equals signs currently doesn't work even when forcing gnu getopt on mac:
```
$ PATH=/opt/homebrew/opt/gnu-getopt/bin/:$PATH ./docker-run-checks.sh --imagei=el8 -I --
WARNING: on Darwin/macOS be sure not to use equals signs (=) for specifying options; instead use
spaces. This is a known issue with macOS's default version of getopt.
docker-run-checks.sh: Invalid option '--imagei=el8'
```

I can look into making this more robust and portable in a separate PR (might need a bit of help with that). 

Would you review again @grondo and let me know if this help message is okay? I used `echo` but this will show up every time someone runs this on a mac. That might clutter things unnecessarily...

I addressed the commit message changes.

grondo:
> I used echo but this will show up every time someone runs this on a mac. That might clutter things unnecessarily...

Yeah, I'd rather not see that when I run the script on Linux, or every time in our CI. Would it work to add the warning to the usage message on MacOS?

src/test/docker/docker-run-checks.sh
```diff
@@ -59,9 +59,13 @@ Options:\n\
...
 # check if running in OSX
-if [[ "$(uname)" == "Darwin" ]]; then
+if [[ "$(uname)" == "Darwin" ]] && [[ $FORCE_GNU_GETOPT != 1 ]]; then
     # BSD getopt
-    GETOPTS=`getopt $short_opts -- $*`
+    GETOPTS=`getopt $short_opts -- $(echo $* | sed "s/=/ /g")`
+    export EXTRA_MACOS_STUFF="\n\
+    You are using BSD getopt on macOS. If you have gnu-getopt in your PATH, \n\
+    force the script to use that by setting FORCE_GNU_GETOPT=1.\n"
+    declare usage=${usage}${EXTRA_MACOS_STUFF}
```
grondo, line 68: I don't think you need to use `declare` again here since `usage` has already been declared. In fact, if this were in a function it might declare `usage` as `local` and the modifications would be lost, so I would remove use of `declare` here.

Also, I thought this issue was about BSD getopt not supporting `=`. Should that be mentioned in the extra help message? i.e. something like "either avoid using '=' between options and values or force the use of GNU getopt with FORCE_GNU_GETOPT=1"?

wihobbs: > "either avoid using '=' between options and values

This message isn't necessary with the proposed changes which strip out all `=` in the `GETOPT` variable (line 65) which is effectively all `=` signs before the `--`. I can add it if you still want it but I just wanted to make sure you didn't miss line 65:

```
GETOPTS=`getopt $short_opts -- $(echo $* | sed "s/=/ /g")`
```

wihobbs: So I guess this PR title needs to be updated because it's not just a help message change anymore.

wihobbs: Oh, and the commit message too. Whoops.

wihobbs: Ok ignore (basically) everything I said above. This should be addressed now.

wihobbs:
Ok, this is the behavior _now_ when an equals sign w/ an option is passed to the script on macOS:

```jupiter13 ~/flux-core (docker-macos-script)$  ./src/test/docker/docker-run-checks.sh --image=focal
docker-run-checks.sh: Invalid option '--image=focal'

Usage: docker-run-checks.sh [OPTIONS] -- [CONFIGURE_ARGS...]
Build docker image for CI builds, then run tests inside the new
container as the current user and group.

Uses the current git repo for the build.

Options:
 -h, --help                    Display this message
     --no-cache                Disable docker caching
     --no-home                 Skip mounting the host home directory
     --install-only            Skip make check, only make install
     --inception               Run tests as flux jobs
     --system                  Run under system instance
 -q, --quiet                   Add --quiet to docker-build
 -t, --tag=TAG                 If checks succeed, tag image as NAME
 -i, --image=NAME              Use base docker image NAME (default=bookworm)
 -p, --platform=NAME           Run on alternate platform (if supported)
 -S, --flux-security-version=N Install flux-security vers N (default=0.11.0)

 -j, --jobs=N                  Value for make -j (default=2)

 -d, --distcheck               Run 'make distcheck' instead of 'make check'
 -r, --recheck                 Run 'make recheck' after failure
 -u, --unit-test-only          Only run unit tests
     --quick-check             Only run check-prep and one basic test
 -P, --no-poison               Do not install poison libflux and flux(1)
 -D, --build-directory=DIRNAME Name of a subdir to build in, will be made
     --workdir=PATH            Use PATH as working directory for build
 -I, --interactive             Instead of running ci build, run docker
                                image with interactive shell.

    You are using BSD getopt on macOS. BSD getopt does not recognize '='
    between options. Use a space instead. If gnu-getopt is first in your
    PATH, force the script to use that by setting FORCE_GNU_GETOPT=1.
```
or this:
```
jupiter13 ~/flux-core (docker-macos-script)$ PATH="/opt/homebrew/opt/gnu-getopt/bin:$PATH" FORCE_GNU_GETOPT=1 ./src/test/docker/docker-run-checks.sh --image=focal  --
Building image focal for user hobbs17 60943 group=20
[+] Building 0.5s (16/16) FINISHED  
```

src/test/docker/docker-run-checks.sh
```diff
@@ -59,9 +59,14 @@ Options:\n\
...
 # check if running in OSX
-if [[ "$(uname)" == "Darwin" ]]; then
+if [[ "$(uname)" == "Darwin" ]] && [[ $FORCE_GNU_GETOPT != 1 ]]; then
     # BSD getopt
     GETOPTS=`getopt $short_opts -- $*`
+    export EXTRA_MACOS_STUFF="\n\
+    You are using BSD getopt on macOS. BSD getopt does not recognize '='\n\
+    between options. Use a space instead. If gnu-getopt is first in your\n\
+    PATH, force the script to use that by setting FORCE_GNU_GETOPT=1.\n"
+    usage=${usage}${EXTRA_MACOS_STUFF}
```
grondo, line 69: It shouldn't be necessary to use `export` here since `EXTRA_MACOS_STUFF` does not need to be exported to the environment of future commands. How about just:
```sh
   usage="${usage}\n\
   You are using BSD getopt on macOS. BSD getopt does not recognize '='\n\
   between options. Use a space instead. If gnu-getopt is first in your\n\
   PATH, force the script to use that by setting FORCE_GNU_GETOPT=1.\n"
```

wihobbs: Good catch @grondo, fixed.

grondo (approved):
Looks great! Thanks!

Issue #2492, opened 2019-10-30
shell: leader rank output.write service could wait forever for EOF if another rank dies


grondo:
Here's a potential issue seen during testing of new shell logging facility.

If a non-leader shell hits an exception after the leader shell has started the output.write service, then the leader shell may wait indefinitely for EOFs that will never be sent.

This is not a huge problem when there is an exception generated, because the job-exec module will eventually kill the leader shell with SIGKILL. However, if a shell exits or dies  due to failure and doesn't log an exception then an indefinite hang could result.

For the fatal error case, perhaps there needs to be a `shell.fatal` plugin hook, which is called before exiting on fatal error. The output plugin could trigger EOFs for all tasks that haven't started yet in the `shell.fatal` callback to effect a more expeditious cleanup.

This doesn't solve the case if a shell unexpectedly exits, but that probably must be handled by a kill-after-abnormal-exit configuration or something.

PR #5780, opened 2024-03-08
improve handling of lost job shells

grondo:
Currently, job shells are detected as "lost" during a job only if a whole node or broker disappears. If a job shell abnormally terminates for some other reason, the rest of the job is not aware which can lead to hangs at exit as noted in #2492.
Also, tasks managed by lost shells are not considered exited, so the exit-timeout and exit-on-error parameters do not have an effect in this case, which could lead a broken job to run until its timeout.

This PR attempts to partially address this situation by having the job execution system report to the leader shell when any other job shell terminates due to a signal. Other abnormal exit conditions are not captured because this is expected behavior, and the hope is that the shell was able to complete its normal shutdown activities in this case, such as sending EOF to the output service of the leader shell and reporting task exit to the doom plugin.

The shell rank that terminated is also passed along to the shell exception handler and the `shell.lost` plugin callback. The output plugin then uses this to adjust its reference count so that it doesn't wait forever for an eof that will never arrive.

Finally, the doom plugin gets a `shell.lost` callback and treats this as a task exit so that the first-task exit timer begins, or a job exception is raised immediately if `exit-on-error` is set.

### DIFF
--- a/src/bindings/python/flux/job/watcher.py
+++ b/src/bindings/python/flux/job/watcher.py
@@ -401,7 +401,7 @@ class JobWatcher:
         if os.WIFEXITED(status):
             status = os.WEXITSTATUS(status)
         elif os.WIFSIGNALED(status):
-            status = 127 + os.WTERMSIG(status)
+            status = 128 + os.WTERMSIG(status)
         return status
 
     def start(self):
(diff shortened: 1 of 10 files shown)


garlick (approved):
LGTM!  I didn't spot anything at all concerning .  This closes multiple oversights in critical infrastructure.  Great work!

PR #5783, opened 2024-03-11
job-manager: include R in sched.free request

### LINKED ISSUE #5775: job-manager: send job->R_redacted directly to the scheduler in sched.free
Problem: at the moment RFC 27 defines `sched.free` as only containing the job ID.  libschedutil optionally uses this on the scheduler side to look up _R_ in the KVS and provide it in the `free()` callback.

Since _R_ is now cached in the job-manager (sans `scheduling` key) this could simply be passed directly to the scheduler, avoiding the lookup.

Note that fluxion sets the SCHEDUTIL_FREE_NOLOOKUP flag, avoiding this lookup.  It only receives the job ID at this time.  But for other schedulers like sched-simple, this would save a KVS lookup for every job.

Also somewhat relevant:
- #4312
- flux-framework/flux-sched#1151.

From discussion over there, I think we can conclude that even for partial release, inclusion of the `scheduler` key is not helpful since it cannot be fragmented by flux-core. 

If this change is made it would require an update to RFC 27.

garlick:
Problem: R has to be looked up from the KVS in the `sched.free` request handler, but now that the job manager caches R, this is an unnecessary extra step.

Add R to the `sched.free` request payload.

Note that the `R.scheduling` key is not included.  The current design of Fluxion in which `R.scheduling` may contain a voluminous JGF object made caching this part of R impractical.

Change libschedutil so that
- the `sched.free` message handler never looks up R in the kvs
- the free callback always sets its `R` argument to NULL
- the SCHEDUTIL_FREE_NOLOOKUP flag is a no-op

Update sched-simple's free callback to unpack R from the message instead of decoding the `R` arugment.

Note that Fluxion sets SCHEDUTIL_FREE_NOLOOKUP so it already expects the free callback's R argument to be NULL.  Although this change increases the size of `sched.free` payloads with data that Fluxion currently does not use, the ranks in R will be required by Fluxion in the future to identify resource subsets for partial release (flux-framework/flux-sched#1151).

This change should be accompanied by an update to RFC 27.

Update sched-simple unit test.

Fixes #5775

### DIFF
--- a/src/common/libschedutil/init.h
+++ b/src/common/libschedutil/init.h
@@ -22,7 +22,7 @@ extern "C" {
 typedef struct schedutil_ctx schedutil_t;
 
 enum schedutil_flags {
-    SCHEDUTIL_FREE_NOLOOKUP = 1, // ops->free() will be called with R=NULL
+    SCHEDUTIL_FREE_NOLOOKUP = 1, // now the default so this flag is ignored
 };
 
 /* Create a handle for the schedutil convenience library.
--- a/src/common/libschedutil/ops.c
+++ b/src/common/libschedutil/ops.c
@@ -45,73 +45,16 @@ static void cancel_cb (flux_t *h,
     util->ops->cancel (h, msg, util->cb_arg);
 }
 
-static void free_continuation (flux_future_t *f, void *arg)
-{
-    schedutil_t *util = arg;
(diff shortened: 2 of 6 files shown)


grondo (approved):
LGTM!

PR #5786, opened 2024-03-11
job-manager: drop sched.free response requirement

garlick:
Problem: the `sched.free` RPC requires a response but the response is unnecessary for correct job management.
    
Moreover, the coupling between the RPC response and the job state machine makes some anticipated changes harder to design.
    
Make `sched.free` a "fire and forget" RPC.  Simply post the `free` event immediately instead of deferring it until the response is received..

Update `flux queue status -v` output to not show pending free rpcs, since they can no longer be pending.
    
This change should be accompanied by an update to RFC 27.

### DIFF
--- a/src/cmd/flux-queue.py
+++ b/src/cmd/flux-queue.py
@@ -92,7 +92,6 @@ def alloc_query(handle):
         sys.exit(1)
     print("{0} alloc requests queued".format(query["queue_length"]))
     print("{0} alloc requests pending to scheduler".format(query["alloc_pending"]))
-    print("{0} free requests pending to scheduler".format(query["free_pending"]))
     print("{0} running jobs".format(query["running"]))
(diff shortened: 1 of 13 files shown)


grondo (approved):
LGTM! I guess it isn't an issue if the scheduler has an error handling a free request since we couldn't/didn't do anything with the error before?

garlick:
Oof just noticed a typo - an extra log message that shouldn't be there.  Will force push.

Before this, a free error would cause the scheduler interface to be torn down (scheduler stopped).  Now we just leave it to schedulers to log something.  I think that is probably fine - the scheduler can choose to abort if it wants to, or log.  (They both currently do log at LOG_ERR level anyway).

garlick:
Thanks - I'll set MWP.

PR #5792, opened 2024-03-13
testsuite: fix test that kills the wrong processes

### LINKED ISSUE #5791: t2492-shell-lost.sh issue test sometimes kills wrong process
Problem: my desktop session exits whenever `issues/t2492-shell-lost.sh` is run.

This seems to "fix" it, in that my desktop stays alive but presumably subverts the purpose of the test:
```diff
diff --git a/t/issues/t2492-shell-lost.sh b/t/issues/t2492-shell-lost.sh
index 159685688..5701cd55a 100755
--- a/t/issues/t2492-shell-lost.sh
+++ b/t/issues/t2492-shell-lost.sh
@@ -38,7 +38,9 @@ try:
     broker_rank = h.get_rank()
     if broker_rank == 3:
         #  kill job shell
-        os.kill(os.getppid(), 9)
+        ppid = os.getppid()
+        if ppid > 0:
+            os.kill(ppid, 9)
         sys.exit(0)
 
     # Sleep in all other tasks
```
getppid(3) says
>  If  the  caller's parent is in a different PID namespace (see pid_name‐spaces(7)), getppid() returns 0.

and a signal sent to pid 0 goes to the whole process group.  What I can't figure out is why this script would think it's in its own PID namespace?

garlick:
Problem: the t2492-shell-lost.sh sends a SIGKILL to the wrong process.

This test runs a job whose two tasks on shell rank 3 try to send a SIGKILL to their parent, the shell.  One succeeds and the other, slower, task may be reparented before calling os.getppid and thus sends the signal to the wrong task.

Running two tasks per shell seems like it doesn't further the purpose of the test so just run one task per shell.

Fixes #5791

### DIFF
--- a/t/issues/t2492-shell-lost.sh
+++ b/t/issues/t2492-shell-lost.sh
@@ -47,7 +47,7 @@ except KeyboardInterrupt:
     sys.exit(0)
 EOF
 )
-id=$(flux submit -N4 --tasks-per-node=2 \
+id=$(flux submit -N4 --tasks-per-node=1 \
 	--input=/dev/null \
 	-o exit-timeout=none \
 	--add-file=critical.py="${CRITICAL_RANKS}" \


grondo (approved):
LGTM! Sorry about that!

garlick:
No worries!  I'll set MWP.

Issue #5776, opened 2024-03-07
job-manager: implement RPC to partially replace sched.resource-status


garlick:
Problem: In flux-framework/flux-sched#1137 it was suggested that the job manger could take over duties from the `sched.resource-status` RPC to avoid delays when the scheduler is busy, since all the alloc/free requests pass through the job manager anyway.

Currently the response consists of three R objects, `all`, `down`, and `allocated`.  I think `all` and `down` can be obtained from `resource.status`, so the job manager would just need to provide `allocated`. 

Here is an example `sched.resource-status` response payload from a small test system:
```json
{
  "all": {
    "version": 1,
    "execution": {
      "R_lite": [
        {
          "rank": "0-7",
          "children": {
            "core": "0-3"
          }
        }
      ],
      "nodelist": [
        "picl[0-7]"
      ],
      "properties": {
        "4g": "2,5",
        "8g": "0",
        "admin": "0",
        "batch": "3-7",
        "debug": "1-2",
        "testproperty": "7"
      },
      "starttime": 0,
      "expiration": 0
    }
  },
  "down": null,
  "allocated": {
    "version": 1,
    "execution": {
      "R_lite": [
        {
          "rank": "0",
          "children": {
            "core": "0-3"
          }
        }
      ],
      "nodelist": [
        "picl0"
      ],
      "properties": {
        "8g": "0",
        "admin": "0"
      },
      "starttime": 0,
      "expiration": 0
    }
  }
}
```

grondo:
For completeness, another idea discussed was to add a separate service that would collect resource information from the `resource` module and the `job-manager` and give consumers a single endpoint to query all status information for resources instead of having multiple RPC responses combined like we need now.

However, moving just the `sched.resource-status` RPC out of the scheduler would be a nice step forward.

Issue #4390, opened 2022-07-05
content-sqlite: json objects are freed twice


garlick:
Problem: @grondo encountered a segfault in the build farm and ran it down to `content-sqlite` with the following reproducer:
```
$ FLUX_TEST_VALGRIND=t ./t0012-content-sqlite.t -v
[snip]
expecting success: 
	test $(flux module stats \
	    --type int --parse object_count content-sqlite) -eq 0 &&
	test $(flux module stats \
	    --type int --parse count content) -eq 0

==2920257== Thread 5:
==2920257== Invalid read of size 8
==2920257==    at 0x6E0D386: json_decref (jansson.h:134)
==2920257==    by 0x6E0D386: stats_get_cb (content-sqlite.c:592)
==2920257==    by 0x486ADB7: call_handler (msg_handler.c:344)
==2920257==    by 0x486B3F9: dispatch_message (msg_handler.c:380)
==2920257==    by 0x486B3F9: handle_cb (msg_handler.c:481)
==2920257==    by 0x4899E52: ev_invoke_pending (ev.c:3770)
==2920257==    by 0x489D720: ev_run (ev.c:4190)
==2920257==    by 0x489D720: ev_run (ev.c:4021)
==2920257==    by 0x4869ECE: flux_reactor_run (reactor.c:128)
==2920257==    by 0x6E0D915: mod_main (content-sqlite.c:846)
==2920257==    by 0x11A34B: module_thread (module.c:183)
==2920257==    by 0x48FA608: start_thread (pthread_create.c:477)
==2920257==    by 0x4B7E132: clone (clone.S:95)
==2920257==  Address 0x55b11c8 is 8 bytes inside a block of size 72 free'd
==2920257==    at 0x483CA3F: free (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==2920257==    by 0x4A4D85D: ??? (in /usr/lib/x86_64-linux-gnu/libjansson.so.4.11.1)
==2920257==    by 0x4A4D908: ??? (in /usr/lib/x86_64-linux-gnu/libjansson.so.4.11.1)
==2920257==    by 0x4A52E00: json_delete (in /usr/lib/x86_64-linux-gnu/libjansson.so.4.11.1)
==2920257==    by 0x486DBD0: json_decref (jansson.h:135)
==2920257==    by 0x486DBD0: json_decref (jansson.h:132)
[...]

PR #5796, opened 2024-03-15
add a faster way to get resource allocation status than sched.resource-status RPC

garlick:
As discussed in #5776, this adds
- resource tracking in the job manager and a new RPC to request the current set of allocated resources
- expansion of the payload of `resource.status` to include the keys found in `sched.resource-status`
- update users (top, python) to call `resource.status` instead of `sched.resource-status`

Marking as a WIP pending

- [x] see what the throughput impact is.  Edit: none at this point
- [x] ~decide what to do with the `alloc-check` plugin since this effectively does the same thing~ Edit: second iteration of the design no longer does this
- [x] ~decide if it was a good idea to combine the two RPCs or whether we should provide two, or perhaps take a step back and design a new one.~  Edit Let's have two RPCs, just like before.
- [x] ~should we leave the scheduler RPC handlers in place and provide some way to use them in test?~  Edit: keep for back compat, but add tests so the code is not left uncovered
- [x] ~update flux-resource(1) which mentions "scheduler view of resources" a lot~ Edit: view is ok - just get fix language about explicitly contacting the schedule
- [x] ~`flux resource status` currently calls the `resource.status` RPCs twice~ Now that there are two RPCs again, this is perfectly fine.

### DIFF
--- a/src/modules/resource/resource.c
+++ b/src/modules/resource/resource.c
@@ -172,7 +172,6 @@ error:
 }
 
 /* Handle client disconnect.
- * Abort a streaming resource.acquire RPC, if it matches.
  */
 static void disconnect_cb (flux_t *h,
                            flux_msg_handler_t *mh,
@@ -183,6 +182,8 @@ static void disconnect_cb (flux_t *h,
 
     if (ctx->acquire)
         acquire_disconnect (ctx->acquire, msg);
+    if (ctx->status)
+        status_disconnect (ctx->status, msg);
 }
 
 flux_t *resource_parent_handle_open (struct resource_ctx *ctx)
@@ -247,7 +248,7 @@ static const struct flux_msg_handler_spec htab[] = {
         .typemask = FLUX_MSGTYPE_REQUEST,
         .topic_glob = "resource.disconnect",
         .cb = disconnect_cb,
-        .rolemask = 0
+        .rolemask = FLUX_ROLE_USER
     },
     FLUX_MSGHANDLER_TABLE_END,
 };
--- a/src/modules/resource/resource.c
+++ b/src/modules/resource/resource.c
@@ -34,6 +34,7 @@
 #include "exclude.h"
 #include "acquire.h"
 #include "rutil.h"
+#include "status.h"
 
 /* Parse [resource] table.
  *
@@ -184,56 +185,6 @@ static void disconnect_cb (flux_t *h,
         acquire_disconnect (ctx->acquire, msg);
 }
 
-static void status_cb (flux_t *h,
-                       flux_msg_handler_t *mh,
-                       const flux_msg_t *msg,
-                       void *arg)
-{
-    struct resource_ctx *ctx = arg;
-    json_t *drain;
-    const json_t *R;
-    json_t *o = NULL;
-    const char *errstr = NULL;
-
-    if (flux_request_decode (msg, NULL, NULL) < 0)
-        goto error;
-    if (ctx->rank != 0) {
-        errno = EPROTO;
-        errstr = "this RPC only works on rank 0";
-        goto error;
-    }
-    if (!(R = inventory_get (ctx->inventory)))
-        goto error;
-    if (!(drain = drain_get_info (ctx->drain)))
-        goto error;
-    if (!(o = json_pack ("{s:O s:o}", "R", R, "drain", drain))) {
-        json_decref (drain);
-        errno = ENOMEM;
-        goto error;
-    }
-    if (rutil_set_json_idset (o,
-                              "online",
-                              monitor_get_up (ctx->monitor)) < 0)
-        goto error;
-    if (rutil_set_json_idset (o,
-                              "offline",
-                              monitor_get_down (ctx->monitor)) < 0)
-        goto error;
-    if (rutil_set_json_idset (o,
-                              "exclude",
-                              exclude_get (ctx->exclude)) < 0)
-        goto error;
-    if (flux_respond_pack (h, msg, "o", o) < 0) {
-        flux_log_error (h, "error responding to resource.status request");
-        json_decref (o);
-    }
-    return;
-error:
-    if (flux_respond_error (h, msg, errno, errstr) < 0)
-        flux_log_error (h, "error responding to resource.status request");
-    json_decref (o);
-}
-
 flux_t *resource_parent_handle_open (struct resource_ctx *ctx)
 {
     if (!ctx->parent_h) {
@@ -261,6 +212,7 @@ static void resource_ctx_destroy (struct resource_ctx *ctx)
 {
     if (ctx) {
         int saved_errno = errno;
+        status_destroy (ctx->status);
         acquire_destroy (ctx->acquire);
         drain_destroy (ctx->drain);
         topo_destroy (ctx->topology);
@@ -290,12 +242,6 @@ static const struct flux_msg_handler_spec htab[] = {
         .topic_glob = "resource.config-reload",
         .cb = config_reload_cb,
         .rolemask = 0
-    },
-    {
-        .typemask = FLUX_MSGTYPE_REQUEST,
-        .topic_glob = "resource.status",
-        .cb = status_cb,
-        .rolemask = FLUX_ROLE_USER,
     },
     {
         .typemask = FLUX_MSGTYPE_REQUEST,
@@ -500,7 +446,6 @@ int mod_main (flux_t *h, int argc, char **argv)
         if (!(ctx->drain = drain_create (ctx, eventlog)))
             goto error;
     }
-
     /*  topology is initialized after exclude/drain etc since this
      *  rank may attempt to drain itself due to a topology mismatch.
      */
@@ -509,6 +454,8 @@ int mod_main (flux_t *h, int argc, char **argv)
     if (!(ctx->monitor = monitor_create (ctx,
                                          inventory_get_size (ctx->inventory),
                                          monitor_force_up)))
+        goto error;
+    if (!(ctx->status = status_create (ctx)))
         goto error;
     if (ctx->rank == 0) {
         if (post_restart_event (ctx, eventlog ? 1 : 0) < 0)
--- a/src/modules/resource/status.c
+++ b/src/modules/resource/status.c
@@ -13,6 +13,7 @@
 #endif
 #include <jansson.h>
 #include <flux/core.h>
+#include <flux/idset.h>
 
 #include "resource.h"
 #include "inventory.h"
@@ -24,13 +25,15 @@
 
 #include "src/common/libutil/errprintf.h"
 #include "src/common/libutil/errno_safe.h"
+#include "src/common/librlist/rlist.h"
 
 struct status {
     struct resource_ctx *ctx;
     flux_msg_handler_t **handlers;
+    struct flux_msglist *requests;
 };
 
-static json_t *prepare_response (struct status *status)
+static json_t *prepare_status_payload (struct status *status)
 {
     struct resource_ctx *ctx = status->ctx;
     const struct idset *down = monitor_get_down (ctx->monitor);
@@ -76,7 +79,7 @@ static void status_cb (flux_t *h,
         errno = EPROTO;
         goto error;
     }
-    if (!(o = prepare_response (status))) {
+    if (!(o = prepare_status_payload (status))) {
         errprintf (&error, "error preparing response: %s", strerror (errno));
         goto error;
     }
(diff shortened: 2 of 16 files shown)


garlick:
Hmm, getting an ASAN failure here:
```
2024-03-15T15:21:32.2114399Z Making check in common
2024-03-15T15:21:32.2245352Z AddressSanitizer:DEADLYSIGNAL
2024-03-15T15:21:32.2246109Z =================================================================
2024-03-15T15:21:32.2249768Z [1m[31m==37186==ERROR: AddressSanitizer: SEGV on unknown address 0x63abbfc31750 (pc 0x7ab1a6bb7cb7 bp 0x000000000000 sp 0x7ffe4ec09ea0 T0)
2024-03-15T15:21:32.2251177Z [1m[0m==37186==The signal is caused by a READ memory access.
2024-03-15T15:21:32.2254299Z AddressSanitizer:DEADLYSIGNAL
2024-03-15T15:21:32.2254926Z AddressSanitizer: nested bug in the same thread, aborting.
2024-03-15T15:21:32.2258036Z make[1]: *** [Makefile:1027: check-recursive] Error 1
2024-03-15T15:21:32.2308431Z make: *** [Makefile:514: check-recursive] Error 1
```

grondo:
Another thing to consider here is backwards compatibility (perhaps a good reason to keep the `sched.resource-status` RPC for a bit). E.g. if a newer instance of Flux is running under an older version, tools like `flux jobs`, `flux top`, and `flux pstree` would `ENOSYS` when attempting to get the subinstance resource information.

grondo:
I'm looking into the ASan errors, which are reproducible but only in the CI environment. In this case, libasan is getting a SEGV while running `make`. Here's more info:
```
runner@fv-az523-355:/usr/src/src$ ASAN_OPTIONS=$ASAN_OPTIONS,verbosity=1 LD_PRELOAD=/usr/lib64/libasan.so.6 make
==134==AddressSanitizer: failed to intercept '__isoc99_printf'
'==134==AddressSanitizer: failed to intercept '__isoc99_sprintf'
'==134==AddressSanitizer: failed to intercept '__isoc99_snprintf'
'==134==AddressSanitizer: failed to intercept '__isoc99_fprintf'
'==134==AddressSanitizer: failed to intercept '__isoc99_vprintf'
'==134==AddressSanitizer: failed to intercept '__isoc99_vsprintf'
'==134==AddressSanitizer: failed to intercept '__isoc99_vsnprintf'
'==134==AddressSanitizer: failed to intercept '__isoc99_vfprintf'
'==134==AddressSanitizer: failed to intercept 'xdr_quad_t'
'==134==AddressSanitizer: failed to intercept 'xdr_u_quad_t'
'==134==AddressSanitizer: failed to intercept 'xdr_destroy'
'==134==AddressSanitizer: libc interceptors initialized
|| `[0x10007fff8000, 0x7fffffffffff]` || HighMem    ||
|| `[0x02008fff7000, 0x10007fff7fff]` || HighShadow ||
|| `[0x00008fff7000, 0x02008fff6fff]` || ShadowGap  ||
|| `[0x00007fff8000, 0x00008fff6fff]` || LowShadow  ||
|| `[0x000000000000, 0x00007fff7fff]` || LowMem     ||
MemToShadow(shadow): 0x00008fff7000 0x000091ff6dff 0x004091ff6e00 0x02008fff6fff
redzone=16
max_redzone=2048
quarantine_size_mb=256M
thread_local_quarantine_size_kb=1024K
malloc_context_size=30
SHADOW_SCALE: 3
SHADOW_GRANULARITY: 8
SHADOW_OFFSET: 0x7fff8000
==134==Installed the sigaction for signal 11
==134==Installed the sigaction for signal 7
==134==Installed the sigaction for signal 8
==134==Deactivating ASan
AddressSanitizer:DEADLYSIGNAL
=================================================================
==134==ERROR: AddressSanitizer: SEGV on unknown address 0x60cf8ad2c9f8 (pc 0x71fac17b5cb7 bp 0x000000000000 sp 0x7ffd602751c0 T0)
==134==The signal is caused by a READ memory access.
AddressSanitizer:DEADLYSIGNAL
AddressSanitizer: nested bug in the same thread, aborting.
```

I have no idea why this just started happening, nor what to try next. Maybe I will try moving the asan to Fedora 36 and see if that addresses this issue.

garlick:
Looking at the first sched test failure, the test fakes resources consisting of 4 nodes, each with 44 cores and 4 gpus.  In the first failure, it allocates two nodes exclusively and expects `flux resource list` to show
- free: 2 nodes, 88 cores, 8 gpus
- alloc: 2 nodes, 88 cores, 8 gpus

But what it actually shows is
- free: 4 nodes, 88 cores, 16 gpus
- alloc: 2 nodes, 88 cores, 0 gpus

I _think_ what may be going on is that librlist doesn't handle gpus for the things I'm asking it to do.  

For completeness, the new objects returned are:

**all**
```json
{
  "version": 1,
  "execution": {
    "R_lite": [
      {
        "rank": "0-3",
        "children": {
          "core": "0-43",
          "gpu": "0-3"
        }
      }
    ],
    "starttime": 0,
    "expiration": 0,
    "nodelist": [
      "sierra[3682,3179,3683,3178]"
    ]
  }
}
```

**alloc**
```json
{
  "version": 1,
  "execution": {
    "R_lite": [
      {
        "rank": "0,2",
        "children": {
          "core": "0-43"
        }
      }
    ],
    "starttime": 0,
    "expiration": 0,
    "nodelist": [
      "sierra[3682-3683]"
    ]
  }
}
```

grondo:
Uhoh, that sounds like an unfortunate bug :-(

garlick:
FWIW I was able to fix the sched test failures by returning the allocated resource set from the job-manager directly in `resource.status` instead of marking those resources allocated the full rlist and then extracting them again with `rlist_copy_allocated()`.

Some other tests are failing now in flux-core that weren't before so I need to sort that out then post an update here.

garlick:
Ah: the allocated set from `job-manager.resource-status` does not include properties so tests involving queues are failing in `t2350-resource-list.t` and `t2801-top-cmd.t`.  I'm not sure if this is because sched-simple isn't returning them in the alloc response, or they are disappearing when passing through `rlist_append()`.  But anyway before they were included because the allocated set was being carved out of the original R with `rlist_copy_allocated()`.

grondo:
Ah, probably dropped in `rlist_append` and that seems like a bug :-(

garlick:
> I wonder how much work it would be to properly support gpus in librlist? Perhaps that would then allow sched-simple to allocate > them (currently it cannot).

Yeah that would allow the original technique I was using to work.  Supporting gpus in sched-simple seems like a pretty big bonus as well!

> My other worry here is that librlist starts to get pretty slow when there are a lot of resources. 

I was hoping that the `rlist_append()` / `rlist_diff()` functions I am using in job-manager would not be too bad, but after I get this working, I'll probe that with a large resource set.

garlick:
Hmm, actually `rlist_append()` seems to pass properties through OK, but it doesn't look like sched-simple is returning queue properties in allocated Rs.  Is it possible that was by design?

grondo:
> but it doesn't look like sched-simple is returning queue properties in allocated Rs. Is it possible that was by design?

I'm not sure, but I did think properties were passed through an alloc, e.g. on fluke the queue properties are available in subinstances (whether this is right or not is questionable)

```
 grondo@fluke108:~$ flux alloc -N1 flux resource list
      free batch           1        4        0 fluke95
 allocated                 0        0        0 
      down                 0        0        0 
[detached: session exiting]
```

garlick:
Hmm, maybe something was wrong with my test then.  Thanks for that!

garlick:
I added a workaround for now, in which properties are re-added to the allocated set by the resource module if missing.

garlick:
FWIW `throughput.py`  is stable at about 21 jobs/sec with 16K nodes/32 cores each, on master and on this branch with sched-simple. :shrug:

grondo:
Oof. I guess sched-simple is so slow at that scale it doesn't matter. I'd like to see results for something smaller as well just in case though.

garlick:
Here are some more numbers.  It does look like there is a negative effect.

```
throughput.py -n 1000 (each run in a new instance)

Node: 32 cores / 4 gpus

        fluxion simple  fluxion simple
        master  master  #5776   #5776
nnodes  thruput thruput thruput thruput
64      595.4   497.0   551.8   505.9
        625.9   523.2   570.5   524.1
        600.4   508.8   527.7   497.4
128     441.4   522.4   429.6   497.6
        468.7   495.4   440.9   485.9
        484.6   505.2   426.3   475.3
256     251.7   501.7   242.0   469.6
        254.7   499.7   238.6   481.4
        249.9   491.5   233.3   454.2
512     142.5   463.2   138.4   456.5
        143.6   482.7   138.2   456.8
        146.2   465.7   137.3   460.5
1024    69.0    443.1   67.8    431.7
        68.6    472.7   68.6    428.7
        68.6    466.5   68.4    449.3
2048    33.1    434.5   32.6    428.5
        32.8    458.5   32.8    428.3
        32.8    447.5   33.5    435.2
4096    16.9    301.5   16.5    285.4
        16.9    293.1   16.5    275.2
        16.8    288.0   16.6    283.2
8192    8.3     104.1   8.2     103.3
        8.3     108.2   8.4     97.7
        8.3     108.0   8.3     103.3
16384   4.1     24.2    4.0     24.0
        4.1     23.9    4.0     23.5
        4.0     24.3    4.1     24.0
```

garlick:
Hmm, maybe instead of keeping a "running R" in the job manager, it would be better (less impact on throughput) to simply gather the R's upon request and either combine them in the job manager or even ship them to the resource module and do it there.

grondo:
The impact is not as bad as I'd feared! The rlist implementation is kind of dumb and not optimized at all, so we could look into improving it. This would help a bit with #5819 a bit too.

However, since each time a user runs `flux resource` it will interrupt the job manager, moving the resource-list service to the resource or another module like we did for the job-list service is not a bad idea IMO.

I'll also note that the slow `sched.resource-status` RPC is an immediate concern on existing clusters, so a solution that works within the next couple releases would be better than spending more time implementing the "perfect" solution (However, it would probably be bad to break compatibility multiple times)

Sorry, all that may not have been too helpful. Just throwing some thoughts out there.

garlick:
In case it wasn't clear, this PR already moves the contact point for the tools to the resource module, but the resource module doesn't do a whole lot except prepare the response.  The RPC in the job manager that returns the `allocated` set is used exclusively by the resource module.  So my comment above was about shifting some of the work from the job manager to the resource module to avoid the penatly on every allocation.

grondo:
>  So my comment above was about shifting some of the work from the job manager to the resource module to reduce d the penalty on every allocation.

That makes sense, sorry I misunderstood. So on receipt of a `resource.status` request, the resource module then requests the allocated set from the job manager, but in the current implementation the job-manager has already prepared the allocated resource set, so all it has to do is return it? This is a pretty clever implementation IMO.

Shifting some of the work to the resource module seems like a good idea to me. Your comment about just shipping to the resource module makes a lot of sense now that I better understand.

garlick:
Just pushed the change discussed above.  I started testing throughput and then realized this doesn't touch the critical path at all so there is little point.

The job manager RPC now just walks all active jobs, creating a union of all allocated Rs, which it returns.  It's kind of dumb and simple, but if it turns out to be slow in some scenario (for example if there are a huge number of active jobs), there are obvious things to do to improve responsiveness such as caching the result until the next alloc/free, keeping all jobs with resources in a list, etc..

garlick:
I've updated the description with a few todos, assuming this approach is acceptable.

I'm leaning towards splitting the `resource.status` RPC into two RPCs again.  Since the job manager query is now somewhat costly (depending on the number of active jobs), it probably makes sense to avoid it when only the previous `resource.status` fields are really wanted.

grondo:
FWIW, I really like the approach.

garlick:
Hmm, I'm seeing this test fail occasionally in CI.  Just going to restart for now.
```
2024-03-24T20:01:08.8286402Z expecting success: flux job attach fMT36St
2024-03-24T20:01:08.8286847Z flux-job: task(s) exited with exit code 1
2024-03-24T20:01:08.8287722Z t2492: Re-launching test script under flux-start using /usr/src/src/cmd/flux
2024-03-24T20:01:08.8288202Z t2492: 
2024-03-24T20:01:08.8288460Z t2492: Testing handling of lost shell rank 3:
2024-03-24T20:01:08.8289067Z t2492: Sumbmitted job ffj4EwR. Waiting for shell rank 3 to be lost
2024-03-24T20:01:08.8291062Z [1m[32m[Mar24 20:00][0m [33mheader[0m [34mversion[0m=[37m1[0m [34mencoding[0m=[35m{"stdout":"UTF-8","stderr":"UTF-8"}[0m [34mcount[0m=[35m{"stdout":4,"stderr":4}[0m [34moptions[0m=[35m{}[0m
2024-03-24T20:01:08.8292925Z [32m[  +0.166865][0m [33mlog[0m [34mrank[0m=[37m0[0m [34mlevel[0m=[37m4[0m [34mmessage[0m=[35m"shell rank 3 (on fv-az701-331): Killed"[0m [34mcomponent[0m=[35m"exception"[0m [34mfile[0m=[35m"exception.c"[0m [34mline[0m=[37m49[0m
2024-03-24T20:01:08.8294014Z t2492: Sending SIGINT to ffj4EwR. Job should now exit
2024-03-24T20:01:08.8294517Z 0.000s: job.submit {"userid":1001,"urgency":16,"flags":0,"version":1}
2024-03-24T20:01:08.8294951Z 0.011s: job.validate
2024-03-24T20:01:08.8295199Z 0.022s: job.depend
2024-03-24T20:01:08.8295456Z 0.023s: job.priority {"priority":16}
2024-03-24T20:01:08.8296020Z 0.024s: job.alloc {"annotations":{"sched":{"resource_summary":"rank[0-3]/core0"}}}
2024-03-24T20:01:08.8296504Z 0.043s: job.start
2024-03-24T20:01:08.8296785Z flux-job: task(s) Killed
2024-03-24T20:01:08.8297065Z 0.429s: job.finish {"status":9}
2024-03-24T20:01:08.8297351Z 0.025s: exec.init
2024-03-24T20:01:08.8297580Z 0.026s: exec.starting
2024-03-24T20:01:08.8298076Z 0.100s: exec.shell.init {"service":"1001-shell-ffj4EwR","leader-rank":0,"size":4}
2024-03-24T20:01:08.8298847Z 0.116s: exec.shell.start {"taskmap":{"version":1,"map":[[0,4,1,1]]}}
2024-03-24T20:01:08.8299916Z 0.410s: exec.shell.task-exit {"localid":0,"rank":0,"state":"Exited","pid":242661,"wait_status":0,"signaled":0,"exitcode":0}
2024-03-24T20:01:08.8300555Z 0.428s: exec.complete {"status":9}
2024-03-24T20:01:08.8300859Z 0.428s: exec.done
2024-03-24T20:01:08.8301395Z 0.258s: flux-shell[0]:  WARN: exception: exception.c:49: shell rank 3 (on fv-az701-331): Killed
2024-03-24T20:01:08.8301946Z waiting on barrier for 4 tasks
2024-03-24T20:01:08.8302234Z exited barrier
2024-03-24T20:01:08.8302527Z t2492: Job exited with rc=137 (expecting 137 (128+9))
2024-03-24T20:01:08.8302895Z t2492: 
2024-03-24T20:01:08.8303158Z t2492: Testing handling of lost shell rank 1:
2024-03-24T20:01:08.8303627Z t2492: Sumbmitted job fvYw4PR. Waiting for shell rank 1 to be lost
2024-03-24T20:01:08.8304752Z [1m[32m[Mar24 20:00][0m [33mheader[0m [34mversion[0m=[37m1[0m [34mencoding[0m=[35m{"stdout":"UTF-8","stderr":"UTF-8"}[0m [34mcount[0m=[35m{"stdout":4,"stderr":4}[0m [34moptions[0m=[35m{}[0m
2024-03-24T20:01:08.8306470Z [32m[  +0.303951][0m [33mlog[0m [34mrank[0m=[37m0[0m [34mlevel[0m=[37m4[0m [34mmessage[0m=[35m"shell rank 1 (on fv-az701-331): Killed"[0m [34mcomponent[0m=[35m"exception"[0m [34mfile[0m=[35m"exception.c"[0m [34mline[0m=[37m49[0m
2024-03-24T20:01:08.8307555Z t2492: Sending SIGINT to fvYw4PR. Job should now exit
2024-03-24T20:01:08.8308049Z 0.000s: job.submit {"userid":1001,"urgency":16,"flags":0,"version":1}
2024-03-24T20:01:08.8308476Z 0.012s: job.validate
2024-03-24T20:01:08.8308723Z 0.023s: job.depend
2024-03-24T20:01:08.8308982Z 0.023s: job.priority {"priority":16}
2024-03-24T20:01:08.8309540Z 0.030s: job.alloc {"annotations":{"sched":{"resource_summary":"rank[0-3]/core0"}}}
2024-03-24T20:01:08.8310014Z 0.052s: job.start
2024-03-24T20:01:08.8310328Z flux-job: task(s) exited with exit code 130
2024-03-24T20:01:08.8310686Z 0.651s: job.finish {"status":33280}
2024-03-24T20:01:08.8310985Z 0.031s: exec.init
2024-03-24T20:01:08.8311213Z 0.032s: exec.starting
2024-03-24T20:01:08.8311706Z 0.140s: exec.shell.init {"service":"1001-shell-fvYw4PR","leader-rank":0,"size":4}
2024-03-24T20:01:08.8312313Z 0.153s: exec.shell.start {"taskmap":{"version":1,"map":[[0,4,1,1]]}}
2024-03-24T20:01:08.8313203Z 0.519s: exec.shell.task-exit {"localid":0,"rank":2,"state":"Exited","pid":243801,"wait_status":2,"signaled":2,"exitcode":130}
2024-03-24T20:01:08.8313848Z 0.651s: exec.complete {"status":33280}
2024-03-24T20:01:08.8314161Z 0.651s: exec.done
2024-03-24T20:01:08.8314799Z 0.398s: flux-shell[0]:  WARN: exception: exception.c:49: shell rank 1 (on fv-az701-331): Killed
2024-03-24T20:01:08.8315458Z Traceback (most recent call last):
2024-03-24T20:01:08.8316030Z   File "/tmp/flux-ancNrP/jobtmp-2-fvYw4PR/critical.py", line 4, in <module>
2024-03-24T20:01:08.8316504Z     import flux
2024-03-24T20:01:08.8316912Z   File "/usr/src/src/bindings/python/flux/__init__.py", line 14, in <module>
2024-03-24T20:01:08.8317409Z     import flux.core.handle
2024-03-24T20:01:08.8317886Z   File "/usr/src/src/bindings/python/flux/core/handle.py", line 19, in <module>
2024-03-24T20:01:08.8318398Z     from flux.future import Future
2024-03-24T20:01:08.8318876Z   File "/usr/src/src/bindings/python/flux/future.py", line 12, in <module>
2024-03-24T20:01:08.8319353Z     import json
2024-03-24T20:01:08.8319709Z   File "/usr/lib/python3.11/json/__init__.py", line 106, in <module>
2024-03-24T20:01:08.8320211Z     from .decoder import JSONDecoder, JSONDecodeError
2024-03-24T20:01:08.8320727Z   File "<frozen importlib._bootstrap>", line 1178, in _find_and_load
2024-03-24T20:01:08.8321309Z   File "<frozen importlib._bootstrap>", line 1149, in _find_and_load_unlocked
2024-03-24T20:01:08.8322138Z   File "<frozen importlib._bootstrap>", line 690, in _load_unlocked
2024-03-24T20:01:08.8322733Z   File "<frozen importlib._bootstrap_external>", line 936, in exec_module
2024-03-24T20:01:08.8323325Z   File "<frozen importlib._bootstrap_external>", line 1032, in get_code
2024-03-24T20:01:08.8323887Z   File "<frozen importlib._bootstrap_external>", line 1130, in get_data
2024-03-24T20:01:08.8324328Z KeyboardInterrupt
2024-03-24T20:01:08.8324596Z Traceback (most recent call last):
2024-03-24T20:01:08.8325145Z   File "/tmp/flux-ancNrP/jobtmp-0-fvYw4PR/critical.py", line 4, in <module>
2024-03-24T20:01:08.8325619Z     import flux
2024-03-24T20:01:08.8326034Z   File "/usr/src/src/bindings/python/flux/__init__.py", line 14, in <module>
2024-03-24T20:01:08.8326534Z     import flux.core.handle
2024-03-24T20:01:08.8327024Z   File "/usr/src/src/bindings/python/flux/core/handle.py", line 19, in <module>
2024-03-24T20:01:08.8327558Z     from flux.future import Future
2024-03-24T20:01:08.8328040Z   File "/usr/src/src/bindings/python/flux/future.py", line 16, in <module>
2024-03-24T20:01:08.8328618Z     from flux.util import check_future_error, interruptible
2024-03-24T20:01:08.8329170Z   File "/usr/src/src/bindings/python/flux/util.py", line 15, in <module>
2024-03-24T20:01:08.8329629Z     import glob
2024-03-24T20:01:08.8329987Z   File "<frozen importlib._bootstrap>", line 1175, in _find_and_load
2024-03-24T20:01:08.8330514Z   File "<frozen importlib._bootstrap>", line 171, in __enter__
2024-03-24T20:01:08.8331000Z   File "<frozen importlib._bootstrap>", line 100, in acquire
2024-03-24T20:01:08.8331395Z KeyboardInterrupt
2024-03-24T20:01:08.8331710Z t2492: Job exited with rc=130 (expecting 137 (128+9))
2024-03-24T20:01:08.8332092Z t2492: Unexpected job exit code 130
2024-03-24T20:01:08.8332744Z Mar 24 20:00:30.720993 broker.err[0]: rc2.0: /usr/src/t/issues/t2492-shell-lost.sh Exited (rc=1) 1.8s
2024-03-24T20:01:08.8333392Z flux-start: 0 (pid 236602) exited with rc=1
2024-03-24T20:01:08.8334058Z Mar 24 20:00:32.653770 broker.err[0]: rc2.0: /usr/src/t/issues/t2492-shell-lost.sh Exited (rc=1) 5.6s
2024-03-24T20:01:08.8334651Z not ok 29 - t2492-shell-lost
```

grondo:
I've been seeing that one regularly in CI, but haven't been able to reproduce locally. It appears the Python test script in the shell-lost test is getting sent SIGINT while it is still importing Python modules, which doesn't make sense because there's a barrier in there :confused: 

One thing we can do is to move the `try:` to the beginning of the script.

grondo:
> which doesn't make sense because there's a barrier in there 😕

D'oh, the libflux `flux_barrier()` returns a future, but the Python handle `barrier` method returns nothing. So this is not really a barrier at all...

garlick:
I think this is more or less ready for a review.

src/modules/resource/resource.c
```diff
@@ -183,6 +182,8 @@ static void disconnect_cb (flux_t *h,
 
     if (ctx->acquire)
         acquire_disconnect (ctx->acquire, msg);
+    if (ctx->status)
+        status_disconnect (ctx->status, msg);
```
garlick: Uh oh just realized that `acquire_disconnect()` might segfault if called on a rank other than zero.  Will fix.

garlick: Also realized `flux resource list` no longer works on rank > 0 and there was no test coverage! Fixups coming.

garlick:
(LMK if you want me to squash those)

grondo:
Feel free to squash those!

garlick:
OK, squashed and rebased on current master.

grondo:
I didn't see this early comment addressed:

> Another thing to consider here is backwards compatibility (perhaps a good reason to keep the sched.resource-status RPC for a bit). E.g. if a newer instance of Flux is running under an older version, tools like flux jobs, flux top, and flux pstree would get ENOSYS when attempting to get the subinstance resource information.

It might be ok to just break compatibility in this case, but I wonder if we can take advantage of the fact that the `resource.sched-status` and `sched.resource-status` RPCs are compatible to support backwards compatibility in the tools for a short while. Sorry if you already responded to this comment and I missed it.

garlick:
> I didn't see this early comment addressed:

Oh, sorry!  I took your comment to heart and left the old RPC in place.  To restore a bit of test coverage without much effort, I added a FLUX_RESOURCE_STATUS_RPC environment variable that allows the topic string used in the python bindings to be overritten, e.g.
```
$ FLUX_RESOURCE_STATUS_RPC=sched.resource-status flux resource list
```

grondo:
Ah, the case I was thinking about is `flux top` or other recursive tools being used against older versions of Flux (e.g. a user running a batch job that uses older Flux). If the newer `flux top` or `flux pstree` is used with the older versions, the RPC perhaps should be retried with `sched.resource-status` if `resource.sched-status` returns `ENOSYS`.

I'm still waffling on whether this is worth the effort, since it would probably have to be handled explicitly in all use cases (perhaps we could use some trick to make it automatic in Python though)

garlick:
Ah, I was thinking old client + new server.  You are right, for new client + old server, we'd probably have to retry with the old topic string on ENOSYS.

grondo:
If we want to do that, I wonder if it would be simplest to create a C wrapper function (it could return an empty future that is either fulfilled by the first or retry response)

grondo:
If it simplifies things, supporting backwards compatibility could be done in a follow-on PR (and I'd be willing to work on that if you've got other things going)

grondo (approved):
This looks great to me, and what an improvement! Seems like this is the way we should have been doing things all along.

I didn't really see much to comment on, and given that everyone's pretty annoyed that `flux resource list` seems to take >30s on elcap at the moment, we should get this in ASAP.

Amazing work!

BTW, I tried to add a "worst case scenario" to my benchmark by having one running job per node (so 16K jobs in the largest test case). The results show good performance for even that case

```
SCHEDULER       NNODES T(resource.sched-status)  T(flux resource list)
sched-simple       128                    0.168                  0.196
sched-simple       256                    0.178                  0.203
sched-simple       512                    0.193                  0.235
sched-simple      1024                    0.229                  0.293
sched-simple      2048                    0.309                  0.420
sched-simple      4096                    0.481                  0.694
sched-simple      8192                    0.881                  1.295
sched-simple     16384                    1.835                  2.633
```

Impact of many jobs doesn't seem to be too bad! I say we get this in.

src/modules/resource/status.c
```diff
@@ -90,13 +93,375 @@ static void status_cb (flux_t *h,
...
+    return o;
+}
+
+/* Given a resource set 'all' with properties, assign any to 'alloc'
+ * that have matching ranks.
+ */
+static int update_properties (struct rlist *alloc, struct rlist *all)
+{
+    struct idset *alloc_ranks;
+    json_t *props;
```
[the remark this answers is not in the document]
garlick: I may have misspoken there!  At least in a quick test, sched-simple does seem to assign properties in the R allocated to jobs.  More testing needed to see where exactly properties were being dropped.

garlick:
Fixed the commit typo and opened #5826 on the missing properties.  I'll go ahead and set MWP.  Thanks!

garlick:
Hmm, in one of the builders, `t2812-flux-job-last.t` hangs in this test (which is for issue #4390).
I don't see how it could be related to this PR but it is a bit concerning.  I'll open an issue and restart the builder.
```
 expecting success: 
  	flux job last "[:]" >lastdump.exp &&
  	flux dump dump.tgz &&
  	flux start -o,-Scontent.restore=dump.tgz \
  		flux job last "[:]" >lastdump.out &&
  	test_cmp lastdump.exp lastdump.out
  
  
  flux-dump: archived 1 keys
  flux-dump: archived 2 keys
  flux-dump: archived 3 keys
  flux-dump: archived 4 keys
  flux-dump: archived 5 keys
  flux-dump: archived 6 keys
  flux-dump: archived 7 keys
  flux-dump: archived 8 keys
  flux-dump: archived 9 keys
  flux-dump: archived 65 keys
  Mar 25 18:41:44.457761 job-manager.err[0]: replay warning: INACTIVE action failed on job fEyxE9d: Read-only file system
  Mar 25 18:41:44.785969 job-manager.err[0]: sched.alloc-response: id=fF1SDRy already allocated
```

PR #5800, opened 2024-03-16
github: run asan build on fedora36

grondo:
Problem: The address sanitizer CI build started failing for unknown reason. The failure is occurring in the make program before even running any tests in 'make check.'

Take the following steps to attempt to address the issue:

 - The asan build currently runs on Fedora 34, which is getting old at this point. Update the sanitizer run to use Fedora 36.

 - Drop the LD_PRELOAD of the libasan library, which seems to be causing the failure in some unrelated programs not compiled with -fsanitize=address (a guess).

 - Add 'verify_asan_link_order=false' to ASAN_OPTIONS to avoid libasan complaints about link order. This is only important if malloc, free, etc., are replaced by libs. In that case libasan _must_ be first in the link order. See https://github.com/google/sanitizers/wiki/AddressSanitizerFlags

 - Update -j2 to -j4 (github runners now have 4 vcpus)

### DIFF
--- a/.github/workflows/asan.yml
+++ b/.github/workflows/asan.yml
@@ -14,13 +14,12 @@ jobs:
     - name: docker-run-checks with ASan
       timeout-minutes: 40
       env:
-        PRELOAD: /usr/lib64/libasan.so.6
-        ASAN_OPTIONS: detect_leaks=0,start_deactivated=true,replace_str=true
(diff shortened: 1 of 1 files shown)


garlick (approved):
Great!  Thanks!

grondo:
Thanks! I was still seeing failures with this commit in another branch, not couldn't reproduce it here, so I've set MWP

PR #5802, opened 2024-03-18
github: set kernel.randomize_va_space=0 for asan build

### LINKED ISSUE #5801: ci: asan build randomly failing
The address sanitizer build wasn't fully fixed by the most recent PR #5800. Instead, I'm seeing the build sometimes succeed and sometimes fail even with those changes. The random failures appear to occur in the Lua unit tests. I used tmate to see what's going on in the actual CI builder and noticed that `make check` in `src/bindings/lua` sometimes succeeds, but more often fails with the error:

```
ERROR: tests/t0000-json
=======================

AddressSanitizer:DEADLYSIGNAL
=================================================================
==986==ERROR: AddressSanitizer: SEGV on unknown address 0x62e53ae85328 (pc 0x79a0c6bae3e0 bp 0x79a0c60b5d6f sp 0x7ffd01713d60 T-1)
==986==The signal is caused by a READ memory access.
AddressSanitizer:DEADLYSIGNAL
AddressSanitizer: nested bug in the same thread, aborting.
ERROR: tests/t0000-json.t - missing test plan
ERROR: tests/t0000-json.t - exited with status 1
```

I'm not sure what would cause random failures in asan, but looking into it.

grondo:
Problem: Recent kernel address space layout randomization (ASLR) appears to confuse some versions of address sanitizers, leading to random failures in github CI.

Disable kernel.randomize_va_space for the github asan build to avoid these failures. Thanks to @vchuravy for pointing out the issue.

Since LD_PRELOAD of libasan wasn't the root cause, reinstate preloading the libasan library to get as full coverage as possible.

Fixes #5801.

### DIFF
--- a/.github/workflows/asan.yml
+++ b/.github/workflows/asan.yml
@@ -11,9 +11,13 @@ jobs:
         ref: ${{ github.event.pull_request.head.sha }}
         fetch-depth: 0
     - run: git fetch --tags || true
+    - name: disable ASLR
+      run: |
+        sudo sysctl -w kernel.randomize_va_space=0
     - name: docker-run-checks with ASan
       timeout-minutes: 40
       env:
+        PRELOAD: /usr/lib64/libasan.so.8
         ASAN_OPTIONS: detect_leaks=0,start_deactivated=true,replace_str=true,verify_asan_link_order=false
         FLUX_TEST_TIMEOUT: 300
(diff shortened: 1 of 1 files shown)


garlick (approved):
Nice!  Thanks @grondo and @vchuravy!

grondo:
Thanks! I've set MWP.

Issue #5782, opened 2024-03-10
rank0 broker crash at instance exit with large number of cancelled jobs


grondo:
Got this segfault when exiting a large (size=768) instance with many running and pending jobs (rc3 was terminating >2000 jobs)

```gdb
Core was generated by `/g/g0/grondo/git/flux-core.git/src/broker/.libs/lt-flux-broker -Stbon.topo=kary'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  fzlist_append (self=0x1555554e85c0 <flux_msg_handler_destroy>, 
    item=0x1554b01280a0) at zlist.c:190
190	        self->tail->next = node;
[Current thread is 1 (Thread 0x1555408db700 (LWP 779526))]
Missing separate debuginfos, use: yum debuginfo-install glibc-2.28-236.el8_9.12.x86_64 hwloc-2.10.0-1.t4.x86_64 jansson-2.14-1.el8.x86_64 keyutils-libs-1.5.10-9.el8.x86_64 krb5-libs-1.18.2-26.1.t4.x86_64 libblkid-2.32.1-44.t4.x86_64 libcap-2.48-6.el8_9.x86_64 libcom_err-1.45.6-5.el8.x86_64 libgcc-8.5.0-20.el8.x86_64 libgcrypt-1.8.5-7.el8_6.x86_64 libgpg-error-1.31-1.el8.x86_64 libmount-2.32.1-44.t4.x86_64 libselinux-2.9-8.el8.x86_64 libsodium-1.0.18-2.el8.x86_64 libstdc++-8.5.0-20.el8.x86_64 libunwind-1.3.1-3.el8.x86_64 libuuid-2.32.1-44.t4.x86_64 lz4-libs-1.8.3-3.el8_4.x86_64 ocl-icd-2.2.12-1.el8.x86_64 openpgm-5.2.122-21.el8.x86_64 openssl-libs-1.1.1k-12.el8_9.x86_64 pcre2-10.32-3.el8_6.x86_64 sqlite-libs-3.26.0-19.el8_9.x86_64 systemd-libs-239-78.el8.x86_64 xz-libs-5.2.4-4.el8_6.x86_64 zeromq-4.3.4-3.el8.x86_64 zlib-1.2.11-25.el8.x86_64
(gdb) where
#0  fzlist_append (self=0x1555554e85c0 <flux_msg_handler_destroy>, 
    item=0x1554b01280a0) at zlist.c:190
[...]

PR #5803, opened 2024-03-18
job-exec: fix rare segfault at scale due to libsubprocess use-after-free

grondo:
This PR fixes the crash in the job-exec module described in #5782.

As described there, the issue was a use-after-free in libsubprocess when a signal is sent to a remote subprocess before the subprocess is running.

I ran the reproducer below for 4 hours with no crashes on this branch. On master the issue reproduces within 30m reliably.

reproducer:
```
$ flux run -o cpu-affinity=off -o mpibind=off \
    --env=MALLOC_CONF=junk:true,tcache:false \
    --env=LD_PRELOAD=/g/g0/grondo/git/jemalloc/lib/libjemalloc.so \
    -o pty.interactive -o pty.capture -n 768 -N32  \
    src/cmd/flux start -o,-Stbon.topo=kary:800
$ while true; flux submit --cc=1-2048 true \
    && sleep 60 \
    && flux cancel --all \
    && sleep 30 \
    && flux cancel --all \
    && sleep 5 \
    && flux cancel --all \
    && flux queue idle; \
    done
```

### DIFF
--- a/src/common/libsubprocess/remote.c
+++ b/src/common/libsubprocess/remote.c
@@ -106,7 +106,8 @@ static void sigpending_cb (flux_future_t *f, void *arg)
      * remote kill(2) (f).
      */
     flux_future_fulfill_with (prev, f);
-    flux_future_destroy (f);
+
+    /* Note, f is not destroyed here since it is owned by prev */
 }
 
 static void fwd_pending_signal (flux_subprocess_t *p)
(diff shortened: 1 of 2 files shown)


garlick (approved):
LGTM!

grondo:
Thanks! I've set MWP.

PR #5808, opened 2024-03-20
python: update required versions of black and mypy

grondo:
Dependabot noted a potential security issue in our currently required version of `black` in #5806, but a bit more than just bumping the version was required to get the suggested `black` version working, so I've done that here. This included apply the very minor formatting changes enacted by the new version.

The changes also required an update of `mypy`, so I've bumped that to the latest version as well. There did not seem to be any fallout.

Pending PRs _may_ have to reformatted if they include python code. However, I was surprised how little changed in the formatting when updating to the new black version.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,11 +25,9 @@ jobs:
 
   python-lint:
     name: python linting
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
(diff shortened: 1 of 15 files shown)


garlick (approved):
Thanks for handling this.  LGTM!

grondo:
Thanks, I've set MWP.

PR #5813, opened 2024-03-21
job-exec: improve cleanup after lost shell events

### LINKED ISSUE #5811: job-exec: job takes longer than necessary to terminate after a node failure
When the job-exec module handles a node failure on a critical rank it raises a job exception but does not also notify the rank 0 job shell that a shell has gone away. This causes the job shell to wait indefinitely in the output plugin for EOF from the missing shell.

This was partially fixed by #5780, but only for non-critical ranks.

grondo:
This PR fixes #5811 by always notifying the leader job shell of lost shells, whether they are critical or not. This then allows the leader shell to exit quickly when all its tasks terminate, rather than waiting for EOF from a lost shell during abnormal termination.

### DIFF
--- a/src/modules/job-exec/exec.c
+++ b/src/modules/job-exec/exec.c
@@ -127,7 +127,8 @@ static int exec_barrier_enter (struct bulk_exec *exec)
     return 0;
 }
 
-static void output_cb (struct bulk_exec *exec, flux_subprocess_t *p,
+static void output_cb (struct bulk_exec *exec,
+                       flux_subprocess_t *p,
                        const char *stream,
                        const char *data,
                        int len,
@@ -310,7 +311,8 @@ static void exit_cb (struct bulk_exec *exec,
     if (ctx->barrier_completion_count == 0
         || ctx->barrier_enter_count > 0) {
         if (bulk_exec_write (exec, "stdin", "exit=1\n", 7) < 0)
-            jobinfo_fatal_error (job, 0,
+            jobinfo_fatal_error (job,
+                                 0,
                                  "failed to terminate barrier: %s",
                                  strerror (errno));
     }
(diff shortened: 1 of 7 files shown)


src/modules/job-exec/exec.c
```diff
@@ -182,8 +183,11 @@ static int lost_shell (struct jobinfo *job,
         va_end (ap);
     }
 
-    if (raise_exception) {
-        /* Raise a non-fatal job exception */
+    if (!critical) {
+        /* Raise a non-fatal job exception if the lost shell was not critical.
+         * The job exec service will raise a fatal excception later for
```
chu11, line 188: s/excception/exception/

src/modules/job-exec/exec.c
```diff
@@ -227,21 +236,28 @@ static void error_cb (struct bulk_exec *exec, flux_subprocess_t *p, void *arg)
...
+            /*  Raise a fatal error and terminate job immediately if
+             *  the lost shell was critical.
+             */
+            if (critical)
+                jobinfo_fatal_error (job,
+                                     0,
+                                     "%s on broker %s (rank %d)",
+                                     "lost contact with job shell",
+                                     hostname,
+                                     rank);
```
chu11, line 260: would it be useful to say "critical" somewhere in this message to differentiate it from the prior one?  Like "lost contact with critical job shell"?  I'm aware the messages are going to different places ... but might be useful when debugging?

chu11 (approved):
LGTM, just one typo and one minor message comment.

src/modules/job-exec/exec.c
grondo: Perhaps, though it is probably enough that the exception raised will be fatal.

Note that by far the common case is that all job shells are critical. Only in the case of a Flux instance (which adjusts its own critical ranks to remove the leaf broker rank), or if a user manually updates critical ranks will the non critical (severity 1) `lost-shell` exception be raised.

Because this is the common case, it might be less confusing to keep the error message to the user simpler, at least for now.

grondo: Also, in case it wasn't obvious from the diff, this particular call didn't change from before, the code was just reorganized to also call `lost_shell()` for a critical rank loss.

grondo:
I've fixed the typo and set MWP, thanks!

PR #5817, opened 2024-03-21
fix `flux submit` and `bulksubmit` handling of mustache templates in command and args

grondo:
The `flux submit` and `flux bulksubmit` commands fail to preserve mustache templates in the job command.

This PR fixes that.


chu11 (approved):
LGTM!

grondo:
Thanks! I've set MWP.

PR #5822, opened 2024-03-23
job-manager: fix infinite loop when loading builtin jobtap plugin

grondo:
A missing `zlistx_next()` in the jobtap code that searches for a matching builtin plugin name causes an infinite loop if a builtin plugin is loaded that doesn't match the first plugin name in the list.

Also, though it isn't advisable to specify a relative plugin path to `flux jobtap load`, if this is done and the path starts with `.`, the current code assumes a builtin plugin is specified.

This PR fixes both these issues.


garlick (approved):
LGTM!

grondo:
Thanks! Setting MWP.

PR #5823, opened 2024-03-23
flux-resource: improve performance of `flux resource list`

grondo:
This PR improves the performance of `flux resource list` by modifying some really inefficient code:

 - The `ResourceSetExtra` wrapper class unnecessarily reinitializes the `ResourceSet` argument. Just stash the `ResourceSet` instead and forward method calls to it to avoid this wasted work.
 - `resources_uniq_lines()` iterates over each individual rank in each resource set to collect common lines of output together. Instead, split a resource set into all combinations of properties and iterate each of these sets. This _should_ be the minimum number of sets required to create possibly unique lines (at least at this point). This reduces the iteration count significantly.

Timing before with a very large resource set (~10K ranks):
```
 time flux resource list --from-stdin < ../../large-list.json >/dev/null

real	0m6.878s
user	0m6.762s
sys	0m0.089s
```

Timing after these changes:
```
$ time src/cmd/flux resource list --from-stdin < ../../large-list.json >/dev/null

real	0m1.790s
user	0m1.662s
sys	0m0.102s
```

Further improvements will probably have to be made in librlist itself.

### DIFF
--- a/src/cmd/flux-resource.py
+++ b/src/cmd/flux-resource.py
@@ -418,11 +418,14 @@ class ResourceSetExtra(ResourceSet):
     def __init__(self, arg=None, version=1, flux_config=None):
         self.flux_config = flux_config
         if isinstance(arg, ResourceSet):
-            super().__init__(arg.encode(), version)
+            self._rset = arg
             if arg.state:
                 self.state = arg.state
         else:
-            super().__init__(arg, version)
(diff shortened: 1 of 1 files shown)


garlick (approved):
LGTM!  Nice results!

grondo:
Thanks! I'll set mwp

PR #5825, opened 2024-03-24
python: fix handle barrier method to return a Future

grondo:
`flux_barrier(3)` returns a `flux_future_t`, but the Python handle barrier method returns nothing.

This PR returns a Future from `handle.barrier()` as is necessary for proper functionality and updates callers.

Note: Clearly the barrier unit tests were not working, but I didn't have any good ideas how to properly test the barrier functionality, so left it alone for now besides addition of the missing `get()` calls.


grondo:
FYI - as a side effect this should fix the intermittent CI failures of `t2492-shell-lost.sh`

garlick (approved):
Good catch!

Issue #5708, opened 2024-01-26
`job-ingest: fluid_init failed` when launching 16 brokers per node on 1040 nodes


grondo:
Trying to launch a test with 16 brokers per node across 1040 nodes and got the following error smörgåsbord: 

```console
$ FLUX_IPADDR_INTERFACE=eno8303 srun -N1040 -n16640 --pty src/cmd/flux start -o,-Stbon.topo=kary:2
Jan 26 13:14:57.552748 job-ingest.err[16639]: fluid_init failed
Jan 26 13:15:03.578565 job-ingest.err[16511]: fluid_init failed
Jan 26 13:15:03.603558 job-ingest.err[16512]: fluid_init failed
Jan 26 13:15:05.902224 job-ingest.err[16517]: fluid_init failed
Jan 26 13:15:05.902337 job-ingest.err[16513]: fluid_init failed
Jan 26 13:15:05.907363 job-ingest.err[16515]: fluid_init failed
Jan 26 13:15:05.915656 job-ingest.err[16516]: fluid_init failed
Jan 26 13:15:05.921857 job-ingest.err[16518]: fluid_init failed
Jan 26 13:15:05.922454 job-ingest.err[16514]: fluid_init failed
Jan 26 13:15:09.752172 job-ingest.err[16447]: fluid_init failed
Jan 26 13:15:10.265893 job-ingest.err[16449]: fluid_init failed
[snip: repeated ad nauseum]
Jan 26 13:15:31.786757 job-ingest.err[16390]: fluid_init failed
Jan 26 13:15:31.787736 job-ingest.err[16388]: fluid_init failed
[...]

PR #5831, opened 2024-03-26
job-ingest: fix handling of size > 16K and reserve some FLUID generator IDs for future use

garlick:
Problem: the fix for #5708 (gracefully handling FLUID ID range errors on nodes >= 16K) is incorrect. 

Fix it and add tests.

Also: reserve the top 16 generator IDs for future use as this may be hard to do later when Flux has actually been run on large systems.


grondo (approved):
Nice catch! LGTM!

PR #5833, opened 2024-03-26
fix handling of `Ctrl-SPACE` (`NUL`) over job ptys (e.g. under `flux job attach`)

### LINKED ISSUE #5832: pty: flux pty code drops ctrl-<space>
A user noticed that `ctrl-<space>` was being ignored in emacs inside a Flux allocation.

I verified this is the case with `showkey -a`

```console
$ showkey -a

Press any keys - Ctrl-D will terminate this program

^@ 	  0 0000 0x00
^D 	  4 0004 0x04

$ flux alloc -n1
$ showkey -a

Press any keys - Ctrl-D will terminate this program

^D 	  4 0004 0x04
```

In the allocation `Ctrl-<space>` had no effect. 

Note above that `Ctrl-<space>` generates `^@` which is the `NUL` character. This probably explains why the current code is dropping it. We'll likely have to special case this sequence.

grondo:
This PR fixes handling of `NUL`s in pty data handling. Jansson by default will not unpack `NUL` unless `JSON_ALLOW_NUL` is used, so the read side of the pty was just dropping them.

As part of developing a test, an issue with `echo something | flux run -o pty.interactive ...` was discovered, which is also fixed here.

Fixes #5832

### DIFF
--- a/src/common/libterminus/client.c
+++ b/src/common/libterminus/client.c
@@ -271,8 +271,8 @@ static void pty_client_data (struct flux_pty_client *c, flux_future_t *f)
         llog_error (c, "unpack: %s", error.text);
         return;
     }
-    if (write (STDIN_FILENO, data, len) < 0) {
-        llog_error (c, "data decode failed: %s", strerror (errno));
+    if (write (STDOUT_FILENO, data, len) < 0) {
+        llog_error (c, "write %zu bytes: %s", len, strerror (errno));
         return;
     }
 }
--- a/src/common/libterminus/pty.c
+++ b/src/common/libterminus/pty.c
@@ -646,24 +646,37 @@ int pty_data_unpack (const flux_msg_t *msg,
 {
     const char *encoding = NULL;
     void *data;
-    int len;
+    size_t len;
+    json_t *o;
+    json_error_t error;
 
-    if (flux_msg_unpack (msg,
-                         "{s:s s?s}",
-                         "data", &data,
-                         "encoding", &encoding) < 0)
+    /* Note: Use json_unpack_ex(3) in order to pass the JSON_ALLOW_NUL flag.
+     *  This is necessary since pty data packed with 's#' may contain NUL
+     *  characters, especially in the case of ^@/Ctrl-Space/etc, which is
+     *  encoded as a NUL character (empty string with len=1).
+     */
+    if (flux_msg_unpack (msg, "o", &o) < 0)
         return errprintf (errp,
                           "failed to unpack data msg: %s",
                           strerror (errno));
-    len = strlen (data);
-
+    if (json_unpack_ex (o,
+                        NULL,
+                        JSON_ALLOW_NUL,
+                        "{s:s% s?s}",
+                        "data", &data, &len,
+                        "encoding", &encoding) < 0) {
+        errno = EPROTO;
+        return errprintf (errp,
+                          "failed to unpack data msg: %s",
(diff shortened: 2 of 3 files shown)


src/common/libterminus/pty.c
```diff
@@ -646,24 +646,34 @@ int pty_data_unpack (const flux_msg_t *msg,
...
+     *  is encoded as a NUL character (empty string with len=1).
+     */
+    if (flux_msg_unpack (msg, "o", &o) < 0
+        || json_unpack_ex (o,
+                           NULL,
+                           JSON_ALLOW_NUL,
+                           "{s:s% s?s}",
+                           "data", &data, &len,
+                           "encoding", &encoding) < 0) {
+        errno = EPROTO;
```
chu11, line 665: is it possible `flux_msg_unpack()` could return errno != EPROTO, like ENOSYS or something?  Dunno if want to split the if checks up.

t/t2612-job-shell-pty.t
```diff
@@ -128,4 +128,14 @@ test_expect_success 'pty: no hang when invalid command is run under pty' '
 	test_expect_code 127 run_timeout 15 \
 		flux run -o pty.interactive nosuchcommand
 '
+# Note: test below uses printf(1) to send [Ctrl-SPACE (NUL), newline, Ctrl-D]
+# over the pty connection and expects ^@ to appear in output.
```
chu11, line 132: for the uninitiated, mention "^@" (NUL)?

garlick (approved):
LGTM!

src/common/libterminus/client.c
```diff
@@ -271,8 +271,8 @@ static void pty_client_data (struct flux_pty_client *c, flux_future_t *f)
         llog_error (c, "unpack: %s", error.text);
         return;
     }
-    if (write (STDIN_FILENO, data, len) < 0) {
-        llog_error (c, "data decode failed: %s", strerror (errno));
+    if (write (STDOUT_FILENO, data, len) < 0) {
+        llog_error (c, "write %zu bytes: %s", len, strerror (errno));
         return;
```
chu11, line 276: i don't know this code that well and this change looks odd in that there isn't a change on the other side of this.  I suppose the other side is listening to all stdio and that internal to "data" it has encoded if this is stdin/out/err or whatever?

src/common/libterminus/pty.c
grondo: This is the server side, so ENOSYS not possible.

src/common/libterminus/client.c
chu11: > I'm not sure what you mean about the other side of this?

I was meaning the server side.

> This is a tty, there is no stdin stdout stderr, or I should say all of them point to the same file descriptor.

That's what I assumed, otherwise this wouldn't make sense :-)  Not sure if a comment is worthwhile.

src/common/libterminus/pty.c
grondo: Ok, I split those up. Good suggestion.

grondo:
Thanks! Setting MWP.

src/common/libterminus/client.c
grondo: Oh, on the server side there's no possibility of redirection like in the client

grondo: Yeah, maybe a comment would have been a good idea

PR #5834, opened 2024-03-27
support fallback to the old `sched.resource-status` RPC for tools that use the new `resource.sched-status` RPC

grondo:
This PR implements the compatibility fallback for users of the new `resource.sched-status` RPC, so that these interfaces and utilities will still work when run against older versions of Flux.

Since `flux-top(1)` was the only C user of this RPC, a fallback was added in place for the summary pane.

For Python, the `ResourceListRPC` class is reworked to implement the fallback, so that all callers benefit.

The handy existing `FLUX_RESOURCE_LIST_RPC` environment variable is then used for testing purposes. By setting the rpc to an invalid value, the fallback is forced.

### DIFF
--- a/doc/man7/flux-environment.rst
+++ b/doc/man7/flux-environment.rst
@@ -344,7 +344,7 @@ components or writing tests.
    commands to resolve URIs to local form.  This is useful in test environments
    where the remote connector does not work.
 
-.. envvar:: FLUX_RESOURCE_STATUS_RPC
+.. envvar:: FLUX_RESOURCE_LIST_RPC
 
    If set, :man1:`flux-resource` uses the specified RPC topic string instead
    of ``resource.sched-status``.  This is used in test to verify that the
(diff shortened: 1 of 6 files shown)


doc/man7/flux-environment.rst
```diff
@@ -344,7 +344,7 @@ components or writing tests.
    commands to resolve URIs to local form.  This is useful in test environments
    where the remote connector does not work.
 
-.. envvar:: FLUX_RESOURCE_STATUS_RPC
+.. envvar:: FLUX_RESOURCE_LIST_RPC
```
garlick, line 347: Oops!

garlick (approved):
Thanks for doing this.  Oof, I was bit sloppy there with the environment variable naming.

doc/man7/flux-environment.rst
grondo: I only noticed because I got it wrong in a couple places!

grondo:
Thanks! Setting MWP.

PR #5837, opened 2024-03-27
eliminate duplicate KVS restart in job-list and job-manager

garlick:
Problem: job-list takes a long time to start when there are many jobs in the KVS.

Both job-list and job-manager have duplicate code for walking the KVS and recreating lists of active and inactive jobs.  Furthermore, job-list requires workarounds to be sure that events from two sources (KVS and job manager journal) are not processed out of order for any given job.

This eliminates the KVS walk in job-list and instead changes the job manager journal so it can feed all the jobs (active and inactive) to job-list on startup.  Furthermore, job-list processes the journal after the reactor starts so module loading should be quick.

Note that as submitted, this PR will break git bisect.  I thought it was more important to split things up for review but we can squash some commits if need be.

It would be very good to get @chu11's review on this as job-list is somewhat unfamiliar territory for me.  I guess the important thing to know is that the entire code path for sourcing events form the KVS is removed and everything goes through the one for "live" events.  The job manager sends all of the eventlogs for existing jobs first, then begins processing new events so order should be preserved.   It does mean that even inactive jobs move through the job states and are transferred from list to list, whereas before I think they were identified as inactive at the beginning and placed directly in the inactive list.

(This is old work revived from march 2022)


garlick:
Ah looks like some tests that expect job-list to be up to date right after a restart are failing in CI.  Maybe it would be better if any requests received while the journal backlog is being processed are deferred until it is done.  I'll have a look at that.

garlick:
Well the same job-list tests are failing in alpine as before, so I must've made a bad assumption about what is going on.  However, I still think it's probably a good change to defer any requests received during initialization until after it completes, so I'd vote to keep those commits.

grondo:
If it is any help, I was testing on this branch and it doesn't appear `flux jobs` requests are paused until `job-list` is ready. The simple test case just runs 4096 jobs (probably could be less) then runs `flux module reload job-list && flux jobs -nac0 | wc -l` which seems to indicate the results are returned before the backlog is fully processed. (I could be misunderstanding, so apologies if so!)


```
f(s=4,d=0,builddir) grondo@corona82:~/git/flux-core.git$ flux jobs -nac0 | wc -l4096
f(s=4,d=0,builddir) grondo@corona82:~/git/flux-core.git$ flux module reload job-list && sleep 0.25  && flux jobs -nac0 | wc -l
2064
f(s=4,d=0,builddir) grondo@corona82:~/git/flux-core.git$ flux module reload job-list && sleep 0.25  && flux jobs -nac0 | wc -l
0
```

garlick:
My theory this morning is that the backlog is processed but R and jobspec are still being fetched.

I may be over my head now - I added jobspec and R to the journal stream (since we have the redacted versions) but tracking down how to make job-list not fetch those if available in the journal is slow going.

garlick:
OK, for fun I pushed the commit that adds R and jobspec to the journal stream.  @chu11 is there any chance you could grab this branch, review what I've done so far, and make some suggestions on how to make use of R and jobspec in the journal stream?

The first problem to solve is to make sure that when the `jsctx->initialized` flag is set, R and jobspec have been fully ingested.

A follow on question is whether we can use the enhanced journal content to avoid looking _anything_ up in the KVS.  I think we might need to add something to the journal for R and jobspec updates to get there, at least.

chu11:
> My theory this morning is that the backlog is processed but R and jobspec are still being fetched.

> The first problem to solve is to make sure that when the jsctx->initialized flag is set, R and jobspec have been fully ingested.

Didn't do a full perusal of the PR but if my skimming of the commits is correct, the inactive jobs are being processed through the journal code just as though they were "live" jobs.

Your suspicion is a good one, R may not yet be loaded or some events have not yet been processed.  Unfortunately in the tests that fail it doesn't output what the state of the job the job is currently in, that'd be a good clue.

> make some suggestions on how to make use of R and jobspec in the journal stream?

You can see this in `process_state_transition_update()`.  You'll see the call chain for `state_depend_lookup()` and `state_run_lookup()` where the lookups are done and after they are completed and the job state is transitioned for the job.  So some of that code could be "collapsed" if you can get R/jobspec right away.

of course, if we can always read in jobspec/R from the journal instead of looking it up, some of that whole transition logic may be able to disappear (i.e. reading a `resource-event` no longer needs to wait for R to be retrieved).  But we can probably deal with that another day.

```
Error: not ok 161 - verify nnodes preserved across restart
Error: not ok 165 - verify ranks/nodelist preserved across restart
Error: not ok 169 - verify task count preserved across restart
Error: not ok 176 - verify task count preserved across restart
Error: not ok 179 - verify task count preserved across restart
Error: not ok 182 - verify task count preserved across restart
Error: ERROR: t2260-job-list.t - exited with status 1
```

some of these appear to be cut & paste errors in the tests, where it should be "verify expiration preserved across restart" (as one example) ... but anyways, they aren't being preserved for one reason or another and they do appear to be all R related or eventlog entries later in the journal, so it does suggest that some stuff isn't loaded yet.

chu11:
alternate idea i thought of this morning, under the assumption that `job-list` is still processing things, we could simply put some waits into the tests to wait for those particular tests to become inactive.  The fact `job-list` is still loading things may not necessarily be something we care about?

i.e. just put some

`flux job list-ids --wait-state=inactive ids ...` in there for the ids we care about.

garlick:
Hmm, looks like the sched valgrind test has detected some malfeasance:
```
 ==19321== Thread 14:
  ==19321== Invalid read of size 1
  ==19321==    at 0x12117809: process_updates (job_state.c:547)
  ==19321==    by 0x12117C2E: job_transition_state (job_state.c:616)
  ==19321==    by 0x121180CA: journal_revert_job (job_state.c:637)
  ==19321==    by 0x121180CA: journal_process_event (job_state.c:1174)
  ==19321==    by 0x121180CA: journal_process_events (job_state.c:1204)
  ==19321==    by 0x12118D27: job_events_journal_continuation (job_state.c:1253)
  ==19321==    by 0x48A2C9A: ev_invoke_pending (ev.c:3770)
  ==19321==    by 0x48A6566: ev_run (ev.c:4190)
  ==19321==    by 0x48A6566: ev_run (ev.c:4021)
  ==19321==    by 0x4879307: flux_reactor_run (reactor.c:124)
  ==19321==    by 0x12117095: mod_main (job-list.c:245)
  ==19321==    by 0x11ACD5: module_thread (module.c:208)
  ==19321==    by 0x4B1E043: start_thread (pthread_create.c:442)
  ==19321==    by 0x4B9D85F: clone (clone.S:100)
  ==19321==  Address 0x1b87cb61 is 33 bytes inside a block of size 40 free'd
  ==19321==    at 0x484317B: free (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
  ==19321==    by 0x1211722C: job_update_destroy.part.0 (job_state.c:300)
  ==19321==    by 0x1212C554: fzlist_remove (zlist.c:317)
  ==19321==    by 0x121176CA: process_updates (job_state.c:548)
  ==19321==    by 0x12117808: process_state_transition_update (job_state.c:431)
  ==19321==    by 0x12117808: process_updates (job_state.c:540)
  ==19321==    by 0x12117C2E: job_transition_state (job_state.c:616)
  ==19321==    by 0x121180CA: journal_revert_job (job_state.c:637)
  ==19321==    by 0x121180CA: journal_process_event (job_state.c:1174)
  ==19321==    by 0x121180CA: journal_process_events (job_state.c:1204)
  ==19321==    by 0x12118D27: job_events_journal_continuation (job_state.c:1253)
  ==19321==    by 0x48A2C9A: ev_invoke_pending (ev.c:3770)
  ==19321==    by 0x48A6566: ev_run (ev.c:4190)
  ==19321==    by 0x48A6566: ev_run (ev.c:4021)
  ==19321==    by 0x4879307: flux_reactor_run (reactor.c:124)
  ==19321==    by 0x12117095: mod_main (job-list.c:245)
  ==19321==    by 0x11ACD5: module_thread (module.c:208)
  ==19321==    by 0x4B1E043: start_thread (pthread_create.c:442)
  ==19321==    by 0x4B9D85F: clone (clone.S:100)
  ==19321==  Block was alloc'd at
  ==19321==    at 0x48455EF: calloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
  ==19321==    by 0x12117BCF: job_update_create (job_state.c:309)
  ==19321==    by 0x12117BCF: add_state_transition (job_state.c:340)
  ==19321==    by 0x12117BCF: job_transition_state (job_state.c:607)
  ==19321==    by 0x121180CA: journal_revert_job (job_state.c:637)
  ==19321==    by 0x121180CA: journal_process_event (job_state.c:1174)
  ==19321==    by 0x121180CA: journal_process_events (job_state.c:1204)
  ==19321==    by 0x12118D27: job_events_journal_continuation (job_state.c:1253)
  ==19321==    by 0x48A2C9A: ev_invoke_pending (ev.c:3770)
  ==19321==    by 0x48A6566: ev_run (ev.c:4190)
  ==19321==    by 0x48A6566: ev_run (ev.c:4021)
  ==19321==    by 0x4879307: flux_reactor_run (reactor.c:124)
  ==19321==    by 0x12117095: mod_main (job-list.c:245)
 ```

**Comment by garlick** [comment]
I think I found that one.  Now I"m wondering if `job->updates` is needed at all anymore.

File: src/modules/job-list/job-list.c [hunk 1]
```diff
@@ -91,6 +102,17 @@ static void purge_cb (flux_t *h,
     flux_log (h, LOG_DEBUG, "purged %d inactive jobs", count);
 }
 
+void requeue_deferred_requests (struct list_ctx *ctx)
+{
+    const flux_msg_t *msg;
+
+    while ((msg = flux_msglist_pop (ctx->deferred_requests))) {
+        if (flux_requeue (ctx->h, msg, FLUX_RQ_TAIL) < 0)
```
grondo, line 110: Clever!

grondo (approved):
Wow, great work! I tested this out and also took a quick pass through the PR, with only trivial comments.

LGTM!

garlick:
Thanks!  @chu11 did you want to give this a look before it gets merged?

I do have another commit that gets rid of `job->updates` and related machinery for deferring updates.  Should I tack that on here or submit as a cleanup PR?

garlick:
Well I went ahead and pushed that extra commit and the fixups from @grondo's review.  Can always lop it off it if you like!

grondo:
Wow, that was nice cleanup!  FWIW, last commit LGTM.

garlick:
oops just realized I prefixed a bunch of commits with `job-info:` instead of `job-list:`.  Fixing.

grondo:
Sorry I missed that in review!

garlick:
NP!  Well, maybe we ought to just merge this since it sounds like @chu11 might be offline for a bit, and we can ask him to look things over when he gets back.  Given the amount of testing that relies on job-list working properly, I'm fairly confident that we're OK.

Setting MWP.

chu11:
I took at a skim at the changes from this PR and things look good to me.  Nothing obviously stands out.  I love how much code was removed!

PR #5838, opened 2024-03-27
doc: add missing R documentation to flux-jobtap-plugins(7)

grondo:
Problem: flux-jobtap-plugins(7) does not document that a redacted R is available in plugin arguments after a job is in the RUN state, nor that an R may be returned in the plugin output arguments from job.state.sched to bypass the scheduler and provide a custom R for a job.

Update flux-jobtap-plugins(7) to document this information.


chu11 (approved):
lgtm!

grondo:
Thanks!

PR #5839, opened 2024-03-27
configure: add missing check for python ply >= 3.9

### LINKED ISSUE #5444: build: missing checks in configure
In #5443 a couple missing checks in `configure` were discovered:

 - The python ply package
 - libsodium

I'm not sure about libsodium. It seems to only be required by `flux-keygen`, which only uses it to print the sodium version. Still required?

grondo:
Problem: flux-core requires the Python ply module, but configure doesn't check for it.

Add a check for this required module in configure.

Fixes #5444


chu11 (approved):
lgtm!

grondo:
Great thanks!

PR #5841, opened 2024-03-28
python: return empty string on epoch time for D conversion flag

### LINKED ISSUE #5827: python: `!D` output conversion specifier should return an empty string for `0.0`
Problem: `flux-jobs(1)` documents the `!D` conversion specifier as returning an empty string "if the timestamp does not exist", but currently it instead appears to return the ISO 8601 datetime for the epoch (possibly converted to current timezone)

```
$ flux jobs -no '{t_run}' fkiED8pKV7M
0.0
$ flux jobs -no '{t_run!D}' fkiED8pKV7M
1969-12-31T16:00:00
```

This should be fixed.

chu11:
Problem: The conversion flag "D" returns an empty string when the value of a timestamp is not present.  In some cases, timestamps are initialized to 0.0 if the timestamp is not available.  This leads to the "D" conversion flag unexpectedly returning the epoch time 1969-12-31T16:00:00 instead of the empty string.

Handle epoch time as an additional special case for the D conversion flag.  When the timestamp is 0, have the D conversion flag return an empty string.

Fixes #5827


grondo (approved):
LGTM! Thanks!

PR #5844, opened 2024-03-29
add `--taskmap=hostfile:FILE` support

grondo:
This PR adds a new taskmap job shell plugin with the traditional "hostfile" mechanism for assigning task ids based on a list of host names. This was requested as a means to do very specific arbitrary task mapping in an easy to understand manner. It was easy to implement and will probably be nice to have around, even if not useful a majority of the time.


src/shell/taskmap/hostfile.c
```diff
@@ -0,0 +1,197 @@
...
+static struct hostlist *hostlist_from_file (const char *path)
+{
+    ssize_t n;
+    size_t len;
+    struct hostlist *hl = NULL;
+    FILE *fp;
+    char *line = NULL;
+
+    if (!(fp = fopen (path, "r")))
+        shell_die_errno (1, "failed to open hostfile: %s", path);
```
garlick, line 82: Any particular reason why this function calls `shell_die_errno()` on error while `hostlist_from_R()` does not?

garlick (approved):
Nice!  Looks good, just one comment and that codeql shadowed variable should probably be fixed.

src/shell/taskmap/hostfile.c
grondo: No, that was left over from a prototype and should probably be fixed.

grondo:
Ok, addressed comments from @garlick and I'll set MWP.

PR #5846, opened 2024-04-01
flux-job: fix `wait-event -m, --match-context`

grondo:
This PR fixes context matching in `flux job wait-event -m, --match-context`, which currently returns a match if the value matches but key does not.

### DIFF
--- a/t/t2231-job-info-eventlog-watch.t
+++ b/t/t2231-job-info-eventlog-watch.t
@@ -182,6 +182,11 @@ test_expect_success 'flux job wait-event w/ bad match-context fails (invalid val
 	test_must_fail fj_wait_event --match-context=type=foo $JOBID exception
 '
 
+# Note: in test below, foo=0 would match severity=0 in buggy version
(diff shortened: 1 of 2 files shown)


t/t2231-job-info-eventlog-watch.t
```diff
@@ -182,6 +182,11 @@ test_expect_success 'flux job wait-event w/ bad match-context fails (invalid val
 	test_must_fail fj_wait_event --match-context=type=foo $JOBID exception
 '
 
+# Note: in test below, foo=0 would match severity=0 in buggy version
+test_expect_success 'flux job wait-event w/ bad match-context fails (bad key, good value)' '
+	test_must_fail fj_wait_event --match-context=foo=0 $JOBID exception
+'
```
chu11, line 188: perhaps mention issue number?

chu11 (approved):
LGTM, only the one minor thing I note below

grondo:
Thanks! I've added the issue reference to the test as suggested and set MWP.

PR #5848, opened 2024-04-01
add utility for posting job events manually if required

grondo:
Jobs can end up stuck in CLEANUP state if, for example, an `epilog-start` event is posted with no corresponding `epilog-finish` event. This can occur if the rank 0 broker restarts when a job epilog is running. Currently, to clean up this state a special, one-off jobtap plugin must be loaded to manually post the missing `epilog-finish` event.

Actually, my guess is this issue will be rare if we've fixed the broker crash that's been the most common cause of this condition. However, it might be useful to have on hand a more usable utility to post custom job events for this case or any other similar problems that might arise.

To that end, this PR adds a simple builtin jobtap plugin that can accept events for jobs as an RPC, and posts them to the requested jobid. A front-end command is then added to use this service: `flux post-job-event JOBID [KEY=VAL...]`. This is just meant to have on hand to be used when necessary. I imagine it will someday be removed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ skip = [
   "src/bindings/python/flux/resource/__init__.py"]
 
 [tool.mypy]
-python_version = 3.6
+python_version = "3.6"
 files = ["src/cmd/**/*.py", "src/bindings/python/flux", "t/python/*.py"]
 mypy_path = ["src/bindings/python", "t/python/tap", "t/python"]
 allow_redefinition = true
--- a/src/cmd/Makefile.am
+++ b/src/cmd/Makefile.am
@@ -120,7 +120,8 @@ dist_fluxcmd_SCRIPTS = \
 	flux-update.py \
 	flux-imp-exec-helper \
 	py-runner.py \
-	flux-hostlist.py
+	flux-hostlist.py \
+	flux-post-job-event.py
 
 fluxcmd_PROGRAMS = \
 	flux-terminus \
--- a/src/cmd/flux-post-job-event.py
+++ b/src/cmd/flux-post-job-event.py
@@ -1,0 +1,55 @@
+#!/bin/false
+##############################################################
+# Copyright 2024 Lawrence Livermore National Security, LLC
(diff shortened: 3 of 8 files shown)


garlick:
Oh good this will be helpful.

Could the event just be an argument to the jobtap plugin so you just run
```
$ flux jobtap load post-event jobid name [key=val] [key=val]...
```
Eh, then you'd have to unload the plugin each time you want to use it, which would be annoying when you have those 10 stuck jobs to deal with.  Nevermind.

garlick (approved):
LGTM!

src/cmd/Makefile.am
```diff
@@ -120,7 +120,8 @@ dist_fluxcmd_SCRIPTS = \
 	flux-update.py \
 	flux-imp-exec-helper \
 	py-runner.py \
-	flux-hostlist.py
+	flux-hostlist.py \
+	flux-post-job-event.py
```
chu11, line 124: typo in commit "finidh"

pyproject.toml
```diff
@@ -10,7 +10,7 @@ skip = [
   "src/bindings/python/flux/resource/__init__.py"]
 
 [tool.mypy]
-python_version = 3.6
+python_version = "3.6"
```
chu11, line 13: commit message typo "aarond"

grondo:
> Could the event just be an argument to the jobtap plugin so you just run:
> ```
> $ flux jobtap load post-event jobid name [key=val] [key=val]...
> ```

This is actually the strategy we're using now, except it does work for multiple jobids because there's a `jobids=[ID1, ID2, ...]` config keyword, so multiple jobids do work (there's no way to pass just `jobid` and `name` to a jobtap plugin since the "config" is packed into a JSON object).

We could keep this method if preferred. It is just slightly awkward, so is more difficult to instruct an admin how to do it rather than a purpose-built command. Therefore, until now either @jameshcorbett or myself have been doing this for the admins.

Like I said, though, I'm hopeful we won't even have to use this.

garlick:
Gotcha.  This is fine, thanks!

grondo:
Thanks!

PR #5850, opened 2024-04-02
minor cleanup in job-manager journal and job-list (mostly inline docs)

garlick:
@chu11 and I were both looking through the job-list source to be sure all is OK before we tag, and we both spotted some incorrect source comments.  Al, feel free to add to this PR.


garlick:
I added some more - maybe have one more look before I set MWP?

chu11:
LGTM still

PR #5852, opened 2024-04-02
add `doc/requirements.txt` to dist

grondo:
This is minor, but it would be useful if `doc/requirements.txt` was in the distribution tarball and it currently is not.

This simple PR adds it.


garlick (approved):
LGTM!

grondo:

PR #5853, opened 2024-04-02
job-list: remove jobspec/R update dead code

chu11:
Problem: Older code that handled jobspec/R updates when jobspec/R were not yet read in from the KVS.  This code is no longer necessary now that jobspec/R are sent via the event journal.

Remove all traces of cached "jobspec_updates" and "R_updates" code.


garlick (approved):
I think this is OK!

PR #5855, opened 2024-04-03
doc: add path to 'manpages' to conf.py

jameshcorbett:
I ran into an issue building the documentation while following the README, `sphinx-build` couldn't find the `manpages` module. Adding the path to `sys.path` manually or on a one-off basis is easy enough but I think adding this will be a bit more foolproof?

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -427,7 +427,6 @@ stderr_devnull_0 = >/dev/null 2>&1
 
 $(MAN_FILES): manpages.py conf.py $(RST_FILES)
 	$(sphinx_man) \
-	PYTHONPATH=$(PYTHONPATH):$(abs_srcdir) \
 	SPHINX_BUILDDIR=$(abs_builddir) $(PYTHON) \
 		-m sphinx $(sphinx_verbose_flags) -b man $(srcdir) ./man \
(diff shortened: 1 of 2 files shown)


grondo:
I hadn't looked at the README before, I wonder why it doesn't just say to run `make` or `make html`?

Anyway, would this change allow removal of the `PYTHONPATH` modification in `Makefile.am`?

e.g. on line 430 here:

https://github.com/flux-framework/flux-core/blob/9ecf7b2827e712c4669e234fa5b9744910c49e1c/doc/Makefile.am#L428-L433

jameshcorbett:
Hmm, good question, I didn't think about that. Maybe so people don't have to build the code / run `configure` to build the documentation?

Yeah it would allow us to drop the PYTHONPATH thing, I'll fix that and force-push.

grondo:
> Edit: which is kinda nice for people who notice a problem with the documentation and don't want to build the whole project to open a PR!

Good point! Though you don't have to build everything to build the docs, just `./configure`, then `cd docs && make` or `make html`.

However, this looks like a good improvement!

grondo (approved):
LGTM!

jameshcorbett:
Thanks! Setting MWP.

PR #5860, opened 2024-04-05
WIP: replace cleanup scriptlets with shutdown script

garlick:
This makes some improvement to the shutdown process:
- eliminate the `flux admin cleanup-push` method of registering cleanup tasks in rc1, and instead have a `shutdown` script similar to other rc scripts
- add `flux shutdown --force` which causes the shutdown script to be called with the `-f` option
- in the shutdown script, add a 10s timeout to `flux queue idle` when `-f` is present
- in the shutdown script, unload the perilog module before canceling jobs to skip the epilog (@grondo's idea - a good one IMHO)

Todo:
- tests need to be written
- document in man pages and admin guide
-  consider running `flux shutdown --force` in `systemctl stop flux`
- I think flux-accounting's use of the "job" test personality is going to be a problem now that `flux admin` is removed so deal with that

I thought maybe a pause at this point to get feedback would be good.  This had a bigger impact on tests than I was expecting, so maybe I took a wrong turn somewhere.


garlick:
Fixed some stuff
- add `shutdown` script to dist manifest (whoops)
- dropped commits that remove `flux admin cleanup-push`
- temporarily add backwards compat code for cleanup scripts (avoids breaking flux-coral2 and flux-accounting)

garlick:
I dropped the actual changes to the shutdown process from this PR (including `shutdown --force`) leaving only the switch from `cleanup-push` to an actual shutdown script since that change alone is a pretty big one.

I think an actual script is going to be far more maintainable than the "cleanup scriptlets" pushed by rc1, and it opens the possibility of letting `flux shutdown` pass options to the script.  For example, in the future we may want to implement a shutdown in the future that gives jobs more of a soft warning so they can checkpoint.  Maybe the `--force` option should be reserved for a future time when we don't automatically cancel running jobs at shutdown. 

A downside is that this impacts the tests that set rc1,rc3, including some in other projects.  (Thus cleanup-push remains as an alias until the other projects can transition)

garlick:
The cleanup scripts came up in the context of a conditional post-portem script in
- #6080 

The refactoring proposed here predates `flux-modprobe(1)`.  I think we might want to consider an approach more integrated with the modprobe approach with `@task` decorators.

PR #5863, opened 2024-04-05
resource: improve `resource.status` response time with many drained ranks

grondo:
This PR improves the time it takes the `resource` module to assemble the `drain` key in the `resource.status` response when there are many drained targets.

The previous method of building the response piecemeal in a `json_t` object is dropped in favor of a custom class that more efficiently represents a set of drained ranks.

A sharness test that simulates a 2^14 size instance was used to simulate some worst-case performance scenarios at scale. The `resource.status` RPC was timed in the following scenarios:

 - no drained ranks
 - 16,000 drained ranks, but matching timestamps and reason fields
 - 16,000 drained ranks, each with differing timestamp or reason fields

The timing is for just the RPC (i.e. `future.wait_for(-1)` was used instead of `future.get()` to avoid the penalty of `json.loads()` in Python)

The `drainit.py` script is a custom script that sends a separate RPC per target to give all drained ranks a different timestamp.

On current master:
```
expecting success: 
	time rpc "resource.status"


real	0m0.252s
user	0m0.191s
sys	0m0.038s
ok 1 - time resource.status rpc, no drained ranks

expecting success: 
	flux resource drain 1-16000 this is a drain reason

ok 2 - drain all ranks with same timestamp

expecting success: 
	time rpc "resource.status"


real	0m52.827s
user	0m0.207s
sys	0m0.031s
ok 3 - time resource.status rpc all drained ranks

expecting success: 
	flux resource undrain 1-16000

ok 4 - undrain all ranks

expecting success: 
	time flux python drainit.py 1-16000


real	4m26.065s
user	0m0.989s
sys	0m0.390s
ok 5 - drain 16000 ranks with differing timestamps

expecting success: 
	time rpc "resource.status" >/dev/null

  
real	40m46.993s
user	0m0.193s
sys	0m0.043s
ok 6 - time resource.status rpc

# passed all 6 test(s)
1..6
```

and on this PR branch:
```
expecting success: 
	time rpc "resource.status"


real	0m0.233s
user	0m0.175s
sys	0m0.033s
ok 1 - time resource.status rpc, no drained ranks

expecting success: 
	flux resource drain 1-16000 this is a drain reason

ok 2 - drain all ranks with same timestamp

expecting success: 
	time rpc "resource.status"


real	0m0.254s
user	0m0.189s
sys	0m0.033s
ok 3 - time resource.status rpc all drained ranks

expecting success: 
	flux resource undrain 1-16000

ok 4 - undrain all ranks

expecting success: 
	time flux python drainit.py 1-16000


real	4m24.734s
user	0m0.916s
sys	0m0.412s
ok 5 - drain 16000 ranks with differing timestamps

expecting success: 
	time rpc "resource.status" >/dev/null


real	0m0.765s
user	0m0.191s
sys	0m0.047s
ok 6 - time resource.status rpc

# passed all 6 test(s)
1..6
```

Further improvements could be made by storing drained ranks directly in a `drainset` instead of an array, so that the object doesn't have to be built and thrown away on each request. However, there's really diminishing returns here (and I measured and the actual time to assemble the entire response is <200ms in the worst case, the rest of the time must be message handling), and the implementation would be slightly tricky because of the `overwrite` behavior, so I've left that for later.

### DIFF
--- a/src/common/librlist/rlist.c
+++ b/src/common/librlist/rlist.c
@@ -859,8 +859,10 @@ int rlist_append (struct rlist *rl, const struct rlist *rl2)
     struct rnode *n = zlistx_first (rl2->nodes);
     while (n) {
         struct rnode *copy = rnode_copy_avail (n);
-        if (!copy || rlist_add_rnode (rl, copy) < 0)
+        if (!copy || rlist_add_rnode (rl, copy) < 0) {
+            rnode_destroy (copy);
             return -1;
+        }
         n = zlistx_next (rl2->nodes);
     }
 
--- a/src/modules/resource/Makefile.am
+++ b/src/modules/resource/Makefile.am
@@ -39,7 +39,9 @@ libresource_la_SOURCES = \
 	drainset.h \
 	drainset.c
 
-TESTS = test_rutil.t
+TESTS = \
+	test_rutil.t \
+	test_drainset.t
 
 test_ldadd = \
 	$(builddir)/libresource.la \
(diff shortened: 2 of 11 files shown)


garlick (approved):
Excellent work. I didn't spot anything of concern here.

grondo:
Thanks!

PR #5865, opened 2024-04-06
fix more performance issues in `flux resource`

grondo:
This PR picks off some more performance issues in `flux resource list` and `status`.

It is built on top of #5863.

This reduces the time for `flux resource list` on a large system with queues from 5s to ~2s, and reduces the `flux resource status` time when there are 16K drained ranks from >40s to ~7s.


garlick (commented):
Great!

garlick (approved):
LGTM!

grondo:
Thanks!

PR #5867, opened 2024-04-08
broker: avoid slow offline child UUID lookup

### LINKED ISSUE #5864: broker spending a lot of time in `child_lookup()`
On a large system @trws noticed that the rank 0 broker was spending a lot of time in `child_lookup()`:

https://github.com/flux-framework/flux-core/blob/08f925e8a0ec3cecc65b9f47bcd09e232e486815/src/broker/overlay.c#L450

Much of the time was in the referenced `strcmp()` due to an apparent linear search. Perhaps low hanging fruit here...

garlick:
Problem: we're seeing the broker spend a lot of time in `child_lookup()` on a system with a large, flat TBON when many nodes are down.

When a message is received from a TBON child that is not currently in the child hash because it is offline, it is either processed as a new hello request, or sent a DISCONNECT control message.  Older code used to log whether the child's UUID was known or unknown by looking it up with a linear search in the child array.  The logging has all been removed at this point, so the search can be dropped.

Fixes #5864

p.s. I manually tested a couple of the scenarios that cause this code path to be traversed:
- a network partition that causes the parent to initiate child disconnect, followed by partition resolution
- an abrupt rank 0 crash (kill -9) that causes the parent to initiate a child disconnect when messages from children that were connected to the old broker are received

All ok.

### DIFF
--- a/src/broker/overlay.c
+++ b/src/broker/overlay.c
@@ -449,19 +449,6 @@ static struct child *child_lookup_online (struct overlay *ov, const char *id)
     return ov->child_hash ?  zhashx_lookup (ov->child_hash, id) : NULL;
 }
 
-/* Slow path to look up child regardless of its status.
- */
-static struct child *child_lookup (struct overlay *ov, const char *id)
-{
-    struct child *child;
-
-    foreach_overlay_child (ov, child) {
-        if (streq (child->uuid, id))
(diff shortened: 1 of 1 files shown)


garlick:
Thanks!  setting MWP

Issue #5856, opened 2024-04-03
resource: `resource.status` RPC is very slow on el cap


grondo:
I noticed that `flux uptime` takes a very long time to return on el cap.

This seems to be due to the `resource.status` taking a long time. A _lot_ of resources appear to be in the `drain` list with "broker unresponsive". We may want to strategize how to compress these entries.

```console
$ time flux python -c 'import json, flux; print(json.dumps(flux.Flux().rpc("resource.status").get()))' >resp.json

real	0m24.489s
user	0m0.160s
sys	0m0.035s
$ jq '.drain | length' <resp.json
1864
```

This is only with O(2000) targets in the drain object, which I wouldn't have thought would have such a large effect, so perhaps something else is going on.

garlick:
Hmm, `rutil_idkey_insert_idset()` (called by `drain_get_info()`, which is called in the `resource.status` handler)looks like it could be a bit suspect in terms of scalability.   For each drained rank, it decodes an idset, adds the new rank to the idset, then re-rencodes it.  Might be worth a closer look anyway.

grondo:
Related: the `resource` module won't allow me to drain  fake resources in a size=1 instance:
```console
$ flux resource list
     STATE NNODES   NCORES    NGPUS NODELIST
      free  16384  1048576   131072 test[1-16384]
 allocated      0        0        0 
      down      0        0        0 
j$ flux getattr size
1
$ flux resource drain 1 test drain 
flux-resource: ERROR: [Errno 22] idset is out of range
$ flux resource drain 0 test drain 


```

grondo:
> Hmm, rutil_idkey_insert_idset() (called by drain_get_info(), which is called in the resource.status handler)looks like it could be a bit suspect in terms of scalability. 

Good guess. Running profiling shows 35% of the broker time is spent in `vebnew()`...

grondo:
Well if we build up the set of matching draininfo objects before encoding to JSON, then that may solve this problem. However, I wonder if using a more efficient container than an array of `size` elements would allow the resource module to more quickly generate the response.

Also, currently timestamps have to exactly match for entries in the response object to be collected together. I wonder if we don't need sub-second precision here and could just use the seconds, thus getting more compression in the result. (All 1800 drained ranks were kept separately in this example because the timestamps didn't exactly match)

garlick:
Agreed, the sub-second precision is not helpful here.  We should switch to whole seconds.

[...]

PR #5870, opened 2024-04-08
fix broker crash in `resource.status` RPC handling when exlcuded ranks are also down

grondo:
This PR fixes issue #5856.

When ranks are excluded, the resource module removes them from the constructed rlist with `rlist_remove_ranks()`. However, there is a bug in `rlist_remove_ranks()` that leaves pointers to the destroyed `rnode` objects in the rnode lookup hash. This causes a subsequent `rlist_mark_down()` to crash if an excluded rank is also down.

This PR fixes that main issue and also ignores missing ranks in `rlist_mark_down()` and `rlist_mark_up()` to avoid a subsequent NULL deref.

### DIFF
--- a/src/common/librlist/rlist.c
+++ b/src/common/librlist/rlist.c
@@ -2164,6 +2164,7 @@ struct rlist *rlist_alloc (struct rlist *rl,
 
     if (!rl || !ai) {
         errno = EINVAL;
+        errprintf (errp, "Invalid argument");
         return NULL;
     }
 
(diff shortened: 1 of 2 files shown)


grondo:
Ok, duh, you can't remove the object from the hash _after_ it has been destroyed. Fixed and repushed. Also fixed an uninitialized bytes warning from valgrind (`error.text` not initialized after all error returns from `rlist_alloc()`)

garlick (approved):
LGTM!   Thanks for running this down so fast.

grondo:
Thanks I'll set MWP. (Hopefully tests pass)

PR #5879, opened 2024-04-10
broker: allow patch versions to interoperate

garlick:
This just ignores the PATCH number in MAJOR.MINOR.PATCH version comparisions in the broker so we can install a patch version without restarting all of flux at once.

Also, if we're going to assume PATCH versions interoperate we may as well relax the check in flux-proxy, so do that too.


grondo (approved):
LGTM!

Issue #5881, opened 2024-04-12
many "transitioning to LOST due to EHOSTUNREACH error on send" log messages during shutdown


garlick:
On shutting down a large flux instance running flux-core 0.61.0, many of the above log messages were logged on rank 0 for what appears to be all of the nodes that had been online, up to about 64 log messages each, until finally the sys admin aborted the shutdown.

The messages initially seem to be in rank order, perhaps corresponding to a broadcast.  After running through all the ranks, "online" messages start appearing with ranks slowly disappearing from the online list (one by one?) until only the rank 0 node remains, then the LOST messages start over and continue, repeating over all the nodes again and again.  The shutdown was aborted about 7 minutes after the "online" message containing only rank 0 appeared.

garlick:
possible additional fallout: some nodes that were rebuffed in this manner and restarted are still trying to join, spewing a stream of
```
DROP upstream control topic - : message received before hello handshake completed
```
The sequence of events appears to be:
- 19:57:47 leader logged that this follower node transitioned to lost (during instance shutdown)
- 19:57:51 follower node restarted
- 19:57:53 leader broker was killed and restarted

Restarting the follower broker got it to join right away.

grondo:
I notice this is fairly reproducible when running a large (>1k ranks) instance with rank 0 running under valgrind. When exiting the instance you get a bunch of these message (at least 800 lines).

garlick:
Note these message were enhanced to show the state before the transition to LOST.  Apparently it is usually FULL, e.g.
```
foo99 (rank 99) transitioning to full->LOST after 59797s due to EHOSTUNREACH error on send
```

PR #5883, opened 2024-04-12
broker: improve handling of overlay network during shutdown

garlick:
This should resolve some problems related to #5881, where we had a flux shutdown get stuck logging many "transitioning to LOST" messages, and some clients stuck in a loop through the next day waiting for a hello response that would never come.

The hello deadlock should be resolved here.  Instead of logging an error message and doing nothing when a disconnect control message is received during hello, we disconnect and start over.

The leader broker problems _may_ be related to the fact that there is no mechanism to prevent clients that have rebooted from saying hello again while the instance is in shutdown state.   Add a mechanism to send disconnect control messages to any client saying hello during shutdown.  Also add the previous state and the time in that state to the LOST messages to get a little more info on what is happening if we see it again.

Finally, the systemd min restart time is raised from 5s to 30s to avoid storms of reconnects, since sending those disconnects could end up being a non-trivial overhead for a broker with many peers.


src/broker/overlay.c
```diff
@@ -777,9 +777,12 @@ static int overlay_sendmsg_child (struct overlay *ov, const flux_msg_t *msg)
             && (child = child_lookup_online (ov, uuid))) {
             flux_log (ov->h,
                       LOG_ERR,
-                      "%s (rank %d) transitioning to LOST due to %s",
```
grondo, line 779: Commit message: till :arrow_right: still?

grondo (approved):
LGTM!

src/broker/overlay.c
garlick: Fixed that and another typo.  Doing some manual testing now.

garlick:
I didn't find anything wrong with this in testing but I did discover one omission that could be corrected at a later date.  Only the rank 0 broker prohibits peers from reconnecting during cleanup.  The other ranks are still in run state and are perfectly happy to let peers return to service, only to ask them to shutdown once cleanup completes.  It's not wrong, but it seems like wasted effort that we could fix.  Unfortunately it's a little more than I think I can get done today.

Let's merge this and get on with it.  Setting MWP.

PR #5887, opened 2024-04-16
librlist: sort rlist before generating R

### LINKED ISSUE #5885: librlist: rlist_to_R() is not necessarily sorted
Problem: The librlist `rlist_to_R()` function does not first sort the rlist object before converting to _R_. This can result in a much larger object than necessary when an rlist is built via a series of `rlist_append()` operations. It also causes unordered hostlist output from commands like `flux resource list`.

grondo:
Problem: rlist_to_R() does not sort the internal list of nodes before generating R. This results in larger JSON output than necessary, especially for an rlist built from multiple rlist_append() operations.

Sort rnodes by rank in rlist_to_R() before generating the JSON result.

Fixes #5885

### DIFF
--- a/src/common/librlist/rlist.c
+++ b/src/common/librlist/rlist.c
@@ -1709,6 +1709,10 @@ json_t *rlist_to_R (struct rlist *rl)
     if (!rl)
         return NULL;
 
+    /*  Reset default sort to order nodes by "rank" */
+    zlistx_set_comparator (rl->nodes, by_rank);
+    zlistx_sort (rl->nodes);
+
     if (!(R_lite = rlist_compressed (rl)))
         goto fail;
 


src/common/librlist/rlist.c
```diff
@@ -1709,6 +1709,10 @@ json_t *rlist_to_R (struct rlist *rl)
     if (!rl)
         return NULL;
 
+    /*  Reset default sort to order nodes by "rank" */
+    zlistx_set_comparator (rl->nodes, by_rank);
+    zlistx_sort (rl->nodes);
+
```
garlick, line 1715: Remember that `zlistx_sort()` invalidates any cached list handles.  I didn't spot any but thought I'd remind just in case!

garlick (approved):
LGTM  - just one comment that is probably a no-op

src/common/librlist/rlist.c
grondo: Thanks! I double checked and librlist doesn't store any handles anywhere. When used it is only temporarily.

PR #5888, opened 2024-04-16
fix potential segfault in test_terminus.t

### LINKED ISSUE #3120: segfault from test_terminus.t in Travis-CI
As reported in #3067 and elsewhere `test_terminus.t` very occasionally segfaults in travis. I haven't been able to reproduce this outside of travis, and corefiles don't seem to be working inside travis, so I'm not sure how to resolve this issue. Perhaps have to inspect the test and look for potential errors.

grondo:
Problem: The terminus server function in the terminus unit tests calls `setenv(3)`, but this function is run in a thread and `setenv(3)` is not thread safe. This could result in rare segfaults if reading the environment races with setting it during the test.

Move the `setenv` to main(), there's no reason to be setting the environment variable specifically in terminus_server() anyway, since this is a unit test, all invocations of the server should have `FLUX_TERMINUS_TEST_SERVER` set.

Fixes #3120

### DIFF
--- a/src/common/libterminus/test/terminus.c
+++ b/src/common/libterminus/test/terminus.c
@@ -68,9 +68,6 @@ static int terminus_server (flux_t *h, void *arg)
     int rc = -1;
     struct flux_terminus_server *t;
 
-    /*  set rank == -1 for testing. Avoid flux_get_rank(3) */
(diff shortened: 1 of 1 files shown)


chu11 (approved):
LGTM!  whew, that was an old issue.

grondo:
Thanks! Setting MWP.

PR #5890, opened 2024-04-17
Support optional arg in -L (--color) short option

### LINKED ISSUE #5828: update usage/docs for cmds with options `-L, --color[=WHEN]` to indicate that `-L` can't take the `WHEN` argument and `=` is required
Since `-L, --color[=WHEN]` uses an optional argument, the argument can't be specified with the short option, `-L`, and the `=` is required in order to hint to `getopt_long(3)` that there is actually an argument. The docs should be clear about this in usage and man pages to avoid confusing failures in option parsing.

chu11:
Problem: Several flux commands support a --color option.  However, a short option (-L) is inconsistently supported.  In addition to this inconsistency, the options to --color are optional and may not always work with the short option.

For consistency across flux tools, only support the long --color option.

Update tests that tested the -L short option.

Update manpages and remove refernce to -L short option.

Fixes #5828

### DIFF
--- a/t/t2612-job-shell-pty.t
+++ b/t/t2612-job-shell-pty.t
@@ -136,7 +136,7 @@ nul_ctrl_d() {
 }
 test_expect_success 'pty: NUL (Ctrl-SPACE) character not dropped' '
 	nul_ctrl_d | flux run -vvv -o pty.interactive -o pty.capture cat -v &&
-	flux job eventlog -HLp output $(flux job last) &&
-	flux job eventlog -HLp output $(flux job last) | grep \\^@
+	flux job eventlog -HL -p output $(flux job last) &&
+	flux job eventlog -HL -p output $(flux job last) | grep \\^@
 '
 test_done
--- a/src/cmd/builtin/dmesg.c
+++ b/src/cmd/builtin/dmesg.c
@@ -81,6 +81,7 @@ static struct optparse_option dmesg_opts[] = {
     { .name = "delta",  .key = 'd',  .has_arg = 0,
       .usage = "With --human, show timestamp delta between messages", },
     { .name = "color", .key = 'L', .has_arg = 2, .arginfo = "WHEN",
+      .flags = OPTPARSE_OPT_SHORTOPT_OPTIONAL_ARG,
       .usage = "Colorize output when supported; WHEN can be 'always' "
                "(default if omitted), 'never', or 'auto' (default)." },
     OPTPARSE_TABLE_END,
--- a/src/cmd/builtin/overlay.c
+++ b/src/cmd/builtin/overlay.c
@@ -63,6 +63,7 @@ static struct optparse_option status_opts[] = {
                " inaccessible behind offline/lost overlay parents",
     },
     { .name = "color", .key = 'L', .has_arg = 2, .arginfo = "WHEN",
+      .flags = OPTPARSE_OPT_SHORTOPT_OPTIONAL_ARG,
       .usage = "Colorize output when supported; WHEN can be 'always' "
                "(default if omitted), 'never', or 'auto' (default)."
     },
--- a/src/cmd/flux-kvs.c
+++ b/src/cmd/flux-kvs.c
@@ -2221,6 +2221,7 @@ static struct optparse_option eventlog_get_opts[] =  {
       .usage = "Display human-readable output",
     },
     { .name = "color", .key = 'L', .has_arg = 2, .arginfo = "WHEN",
+      .flags = OPTPARSE_OPT_SHORTOPT_OPTIONAL_ARG,
       .usage = "Colorize output when supported; WHEN can be 'always' "
                "(default if omitted), 'never', or 'auto' (default)."
     },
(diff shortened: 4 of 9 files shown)


grondo:
Hm, I don't approve this. I use `-L` all the time to force color when passing output to `less`, etc, and I refuse to type `--color=always` :slightly_smiling_face:  #5828 suggests updating docs and `--usage`, not removal of the option.

chu11:
Understood, I went for this approach bc

A) we are already inconsistent, so some tools don't support -L.  Do you think we should leave that be or add -L to those tools?

B) I noticed `ls` doesn't support -L.

Personally, I dislike the inconsistency more than the documentation issue.

grondo:
> Do you think we should leave that be or add -L to those tools?

We should leave it off IMO if it hasn't had a use case. The short option is useful for commands that might be piped to `ls` to preserve color in that case. For tools like `flux top` that doesn't seem necessary?

> I noticed `ls` doesn't support -L.

Yes, but if we're looking for prior examples, `dmesg(1)` and `hexdump(1)`  `cfdisk(8)` and `fdisk(8)` do :stuck_out_tongue_winking_eye:

My preference would be to not remove existing options people may be using. I'm not so sure the short option _must_ be added to commands where it is currently missing, but if it does bother you then that would be fine with me.

Mainly I was thinking we just update docs to say `-L` is equivalent to `--color=always`. (Though I do notice that `dmesg -Lalways` works and `dmesg -L always` does not: I wonder if there is some way to fix our implementation (possibly `dmesg` is parsing options by hand though...)

garlick:
Looks like dmesg uses getopt (if I have the right dmesg):
https://github.com/util-linux/util-linux/blob/master/sys-utils/dmesg.c#L1715

from getopt(3):
> Two colons mean an option takes an optional arg; if there  is  text  in
       the current argv-element (i.e., in the same word as the option name it‐
       self, for example, "-oarg"), then it is returned in  optarg,  otherwise
       optarg is set to zero.  This is a GNU extension.

grondo:
Hm, in optparse we only do the two colons when a flag exists. Need to look at this again:
```c
    if (o->has_arg == 1)
        colons = ":";
    else if (o->has_arg == 2 && o->flags & OPTPARSE_OPT_SHORTOPT_OPTIONAL_ARG)
        colons = "::";

```

chu11:
> Hm, in optparse we only do the two colons when a flag exists. Need to look at this again:

Yeah, I noticed last night but didn't look into it too much bc A) the approach I was going to take and B) dunno if we should rely on a GNU extension for optional short options?

Can look at later as well.

grondo:
> B) dunno if we should rely on a GNU extension for optional short options?

We have vendored gnu `getopt_long` so I'm not sure that's an issue.

I wonder if "fixing" this will break usage like `flux dmesg -HLf` though...

grondo:
I had to remind myself why that flag exists:
```
commit 55e57d4c238d504a05d912f7da86ee188fdf3351
Date:   Wed Jun 9 11:46:15 2021 -0700

    optparse: make optional arg apply to longopt only by default
    
    Problem: It is common practice to make a `-v, --verbose` option
    increase verbosity with multiple use, e.g. `-vvv` sets `verbose = 3`
    internally in the command or process to which the option is applied.
    If using the long option, however, it would be more convenient if an
    optional argument was used, so that `--verbose=3` is equivalent to
    `-vvv` (Nobody wants to write `--verbose --verbose --verbose). Sadly,
    this is not currently supported in an acceptable way by liboptparse. If
    you choose an optional argument, then `-v` cannot be combined with
    any other short option, since `-vv` passes `v` as the argument to
    the first `-v`.
    
    Change liboptparse so that optional argument (.has_arg = 2) applies
    only to the long option by default. The old behavior can be
    restored with the new flag: OPTPARSE_OPT_SHORTOPT_OPTIONAL_ARG.

```

chu11:
> So I'm not 100% sure it is an improvement, though flux dmesg -HfL works, so maybe it is?

This feels better IMO.  There's a clear error message that makes sense with the `flux dmesg -HLf` case.

grondo:
Yeah, this would be fine with me. Some existing use cases would fail, but as long as `-L` is the last short option in a group it can continue to be used as `--color=always`.

Note that users may still get confused that `flux dmesg -L always` doesn't work, but this is the case for all the tools previously mentioned so it clearly isn't a big deal. Mention can be made of this limitation in the man page(s).

chu11:
re-pushed basically adding OPTPARSE_OPT_SHORTOPT_OPTIONAL_ARG in appropriate locations to get `--color`'s `-L` working

Updated a few tests to add `-Lalways`.

I didn't add new color tests for all the tools that didn't already have color tests, as I felt that was overkill?  But obviously could be added.

I did not make `--color` consistent between tools, figuring that could be a different PR if we decide to do it.

t/t2612-job-shell-pty.t
```diff
@@ -136,7 +136,7 @@ nul_ctrl_d() {
 }
 test_expect_success 'pty: NUL (Ctrl-SPACE) character not dropped' '
 	nul_ctrl_d | flux run -vvv -o pty.interactive -o pty.capture cat -v &&
-	flux job eventlog -HLp output $(flux job last) &&
-	flux job eventlog -HLp output $(flux job last) | grep \\^@
+	flux job eventlog -HL -p output $(flux job last) &&
```
grondo, line 139: This commit should probably go before 90f0d30 so the test isn't broken in between (and note in the commit message that `-L` will be fixed to take an optional arg so this usage will break)

grondo (approved):
LGTM. Probably should reorder commits as suggested below, though.

t/t2612-job-shell-pty.t
chu11: oh yeah! i fixed that last as it was uncovered.

chu11:
thanks re-ordered commits and tweaked that commit message.  setting MWP

PR #5871, opened 2024-04-09
shutdown: unload perilog plugin during shutdown

garlick:
Problem: when flux is shutting down and canceling running jobs, the epilog can greatly extend the shutdown time.

Unload the perilog plugin as part  of the cleanup tasks so that canceled jobs can complete quickly in the shutdown case.

Note: @grondo's good IMHO idea!  This was peeled off of #5860 because it seems like it might be a good candidate for a point release, whereas #5860 is too big of a change for that.

PR #5894, opened 2024-04-17
suppress epilog on jobs canceled during shutdown (try no. 2)

garlick:
Problem: most unfortunately, a typo in rc1 rendered #5871 ineffectual.

Fix the typo and add a test this time.

### DIFF
--- a/etc/rc1
+++ b/etc/rc1
@@ -113,7 +113,6 @@ fi
 if test $RANK -eq 0; then
     if test -z "${FLUX_DISABLE_JOB_CLEANUP}"; then
 	flux admin cleanup-push <<-EOT
-	flux jobtap remove perilog 2>/dev/null || true
 	flux queue stop --quiet --all --nocheckpoint
 	flux cancel --user=all --quiet --states RUN
 	flux queue idle --quiet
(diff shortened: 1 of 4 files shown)


grondo (approved):
LGTM. Sorry I missed that in review 😔

garlick:
I'm sorry too, should have added a test then. :-(

garlick:
Hmm, the `t2412-sdexec-perilog.t` is hanging in the cleanup `flux queue idle` command.  I wonder if it is because there is an in-progress epilog and removing the perilog plugin causes it to not post the `epilog-finish` event when it completes?

Adding a `flux queue drain` at the end of the script does seem to fix it.  Hmm, perhaps more thought required here.

grondo:
Hmm, good question about unloading perilog.so when there are prologs and epilogs in progress.. Not sure off hand how to make that "safe".

grondo:
The `epilog-finish` event? Presuming a prolog or epilog is running is that safe? I assume on shutdown the subprocesses that are running these scripts will be terminated. Could that leave nodes in an inconsistent state? (Just thinking out loud)

garlick:
Good point.  What about having the perilog plugin watch the broker's state and when it enters cleanup, stop running new perilogs but continue servicing running ones?

grondo:
That's a great idea!

grondo:
My first thought would be to expose an RPC in the plugin which could adjust optional behavior at runtime?

PR #5895, opened 2024-04-17
broker/module.c: name threads

trws:
Tiny change, but handy.

problem: As it is, all threads are named `flux-broker-<num>` when identified with `info threads` in gdb, or in perf, or wherever.

solution: Since almost all threads are actually actors executing a specific module, name the thread using the module name during module setup.

Here's an example "after":
image

### DIFF
--- a/src/broker/module.c
+++ b/src/broker/module.c
@@ -91,6 +91,18 @@ static int setup_module_profiling (module_t *p)
     cali_begin_int_byname ("flux.rank", p->rank);
     cali_begin_string_byname ("flux.name", p->name);
 #endif
+    size_t len = strlen (p->name);
+    // one character longer than target to pass -Wstringop-truncation
+    char local_name[17] = {0};
+    const char *name_ptr = p->name;
+    // pthread name is limited to 16 bytes including \0 on linux
+    if (len > 15) {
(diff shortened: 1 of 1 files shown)


src/broker/module.c
```diff
@@ -91,6 +91,8 @@ static int setup_module_profiling (module_t *p)
     cali_begin_int_byname ("flux.rank", p->rank);
     cali_begin_string_byname ("flux.name", p->name);
 #endif
+    // Set the name of each thread to its module name
+    pthread_setname_np(pthread_self(), p->name);
```
grondo, line 95: Wow, great idea! I've definitely hit the "what module am I looking at here?" issue in gdb.

Very minor comments:

 - space between function name and opening parens for consistency (sorry about that)
 - check return code or cast to  `(void)` to establish obvious intent to ignore return code

garlick: Yeah, I didn't realize this was possible - excellent change.

trws: Ok, made those adjustments and in double-checking the return codes also found that there's a length limit to worry about, so we now check the length and if it's too long take the prefix.

grondo (approved):
LGTM!

PR #5897, opened 2024-04-18
alloc-check: account for resources in scheduler hello failure

garlick:
Problem: when a job gets a `scheduler-restart` exception due to RFC 27 hello failure with the scheduler, the `alloc-check` plugin does not properly account for resources.

Fix that by handling the exception in the jobtap plugin.

In addition, eliminate a synchronous KVS lookup of R since R is now available as a jobtap callback argument, which should reduce the overhead of loading this plugin.


garlick:
Sorry for the force pushes - I noticed an unsquashed fixup, then a commit typo.  All done now.

grondo (approved):
LGTM!

garlick:
Thanks! I'll set MWP.

Issue #5898, opened 2024-04-18
add a --config-path=PATH option to flux config get


garlick:
Problem: checking the flux config when flux is not running is hard.

Idea: add `flux config get --config-path=PATH`, where PATH is handled like the broker option of the same name.  This would parse the config at PATH and allow it to be dumped as JSON on stdout without going through the broker.

It would then be possible to check offline that all configs are synced:
```
# Check that all configs are the same
$ pdsh -a "flux config get --config-path /etc/flux/system/conf.d | md5sum" | dshbak -c
$ pdsh -a "flux config get --config-path /etc/flux/security/conf.d | md5sum" | dshbak -c
```
or to build commands that check the value of individual keys.

grondo:
> Another thing that would be helpful would be to have some way to dump R, whether it's configured via resources.path or the resources.config array. Then R could be fed into flux R decode for querying.

There is `flux R parse-config PATH`. It could be extended to read `resources.path` in addition to `resources.path`

Issue #5878, opened 2024-04-10
admin guide: add troubleshooting tips


garlick:
I can't recall all the various things we've seen recently, but we ought to be keeping the admin guide up to date on the problems that may be common ones on a system instance or new cluster.

One that just came up and has come up repeatedly:
```
$ flux alloc -N 2 -q bench --exclusive -t 30m
flux-job: foiQDauigyD started                                                                                            00:00:06
6.587s: flux-imp[866]: stderr: flux-imp: Fatal: exec: userid 34912 is invalid on test110
6.588s: flux-imp[865]: stderr: flux-imp: Fatal: exec: userid 34912 is invalid on test109
6.592s: job.exception type=exec severity=0 test[109-110] (ranks 865-866) terminated before first barrier
flux-job: task(s) exited with exit code 1
flux-job: No job output found
```

Another is the question about flux versioning.  I think we cover that but probably not in the troubleshooting section

Another is the torpid node auto-draining.

PR #5899, opened 2024-04-18
doc: add pre-flight checklist to admin guide

garlick:
#5878 suggests that some common sys admin problems be added to the troubleshooting section of the admin guide.  Thinking about it a bit, maybe a checklist for a new system would be more useful?

I started this but got stuck thinking about how to check configuration and opened #5898.

Maybe others will have some good ideas for things to add here.


garlick:
I finished up that section and fixed up conflicts.  Dropping WIP as this is ready for a review.

doc/guide/admin.rst
```diff
@@ -852,6 +852,272 @@ to be added to one or more PAM stacks, e.g.
...
+     - All nodes
+
+   * - flux-sched
+     - All nodes
+
+   * - flux-pam (optional)
+     - All nodes
+
+   * - flux-accounting (optional)
+     - Leader (management) node only
```
grondo, line 880: Isn't the scheduler only required on the leader node (modules are only loaded on the leader), and flux-accounting should be installed on all nodes so that users/admins have access to the query tools like `flux account` (which can be run on any node and query the rank 0 accounting service) The accounting systemd service should only be enabled on the management node though.

doc/guide/admin.rst
```diff
@@ -852,6 +852,272 @@ to be added to one or more PAM stacks, e.g.
...
+
+Does /var/lib/flux have plenty of space on the leader node?
+===========================================================
+
+Flux stores its databases and KVS dumps in the ``statedir`` on the
+leader (management) node.  Storage consumption depend on usage, the
```
grondo, line 886: s/depend/depends/

grondo (approved):
This is great! Just noticed a couple minor things.

doc/guide/admin.rst
```diff
@@ -852,6 +852,272 @@ to be added to one or more PAM stacks, e.g.
...
+  ----------------
+  test[0-7]
+  ----------------
+  1b3c226159b9041d357a924841845cec  /etc/flux/system/curve.cert
+
+  $ pdsh -a stat -c '"%U %A"' /etc/flux/system/curve.cert | dshbak -c
+  ----------------
+  test[0-7]
+  ----------------
+  flux -r--------
```
chu11, line 960: add link to the section on cert creation?

doc/guide/admin.rst
garlick: > Isn't the scheduler only required on the leader node

For the system instance, but presumably the system instance launches batch jobs with schedulers on other nodes

> flux-accounting should be installed on all nodes so that users/admins have access to the query tools like flux account (which can be run on any node and query the rank 0 accounting service) The accounting systemd service should only be enabled on the management node though.

Oh good point. I hadn't thought about that.  Well @cmoussa1 might want to weigh in here.  The accounting guide says management node only but that may predate the work to let the clients work remotely.

https://flux-framework.readthedocs.io/en/latest/guides/accounting-guide.html#installing-software-packages

Isn't the systemd unit enabled by default though?

grondo: > For the system instance, but presumably the system instance launches batch jobs with schedulers on other nodes

Oh heh, thinking brain not thinking.

grondo: > Isn't the systemd unit enabled by default though?

It seems to be disabled by default (at least in the RPM?) I kind of forget how this works right now, but here's the output from a TOSS login node:
```
● flux-accounting.service - Flux accounting service
   Loaded: loaded (/usr/lib/systemd/system/flux-accounting.service; disabled; vendor preset: disabled)
   Active: inactive (dead)
```

I think `vendor preset: disabled` means the service needs to be explicitly enabled with `systemctl enable flux-accounting`

doc/guide/admin.rst
garlick: I didn't link back to the relevant sections on all the other checklist items - do you think that would be a good idea?

doc/guide/admin.rst
cmoussa1: I think the accounting guide needs to be updated with some notes regarding the new accounting systemd service that was added a little while back (there is currently an open PR in flux-framework/flux-docs#216 that looks to clarify this). It used to be the case where you could only run flux-accounting commands on the management node, but that is no longer the case, you should be able to run it from anywhere on the cluster.

My understanding was that flux-accounting is typically only installed on the management node, however, since that is where the DB is installed.

doc/guide/admin.rst
chu11: I don't think it's necessary for all of them.  This one stood out to me as a "did I completely forget a step?"  vs. other ones.  Like starting a stopped queue, your simple example is more than enough information to move forward.

doc/guide/admin.rst
grondo: > My understanding was that flux-accounting is typically only installed on the management node, however, since that is where the DB is installed.

This is not the case on our TOSS clusters running flux. flux-accounting is installed at least on the login nodes as well.

cmoussa1: My mistake then! Sorry for the confusion.

grondo: Shall we just simplify and have all packages installed on all nodes?

garlick: Sure!  Thanks @cmoussa1 for the follow up.

doc/guide/admin.rst
```diff
@@ -852,6 +848,272 @@ to be added to one or more PAM stacks, e.g.
...
+  debug: Scheduling is started
+  batch: Scheduling is started
+
+Can I run a job?
+================
+
+.. code-block:: console
+
+  $ flux run hostname
+  test1
```
grondo, line 1116: Perhaps at this point it would be useful to validate that the job runs as the correct user, with supplementary groups, etc, e.g.:

```console
$ id
uid=1000(pi) gid=1000(pi) groups=1000(pi),27(sudo),114(netdev)
$ flux run -N4 id
uid=1000(pi) gid=1000(pi) groups=1000(pi),27(sudo),117(netdev)
uid=1000(pi) gid=1000(pi) groups=1000(pi),27(sudo),114(netdev)
uid=1000(pi) gid=1000(pi) groups=1000(pi),27(sudo),117(netdev)
uid=1000(pi) gid=1000(pi) groups=1000(pi),27(sudo),114(netdev)
```

garlick:
Pushed all the suggested changes.

grondo (approved):
Already approved, but updates LGTM

garlick:
Thanks! Setting MWP.

PR #5903, opened 2024-04-19
flux-exec: set up systemd environment to support sdexec debugging

### LINKED ISSUE #5901: add tooling for poking at sdexec transient units
Problem: it is awkward to use systemd tools to manipulate flux transient units when running with `exec.servce = "sdexec"`.

Commands must be run as the flux user with `DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/$(id -u flux)/bus` set in the environment.  One idea would be to just set that in the environment of processes run with `flux exec`.  That actually does happen when when `flux exec --service=sdexec` is used, but starting another transient unit may be too invasive when debugging a problem.

Another thing is that the transient unit names include both the broker's rank and the job ID (in ASCII F58), and sometimes that string is awkward to build.  Something like this works in some contexts.
```sh
jobid_f58_ascii=$(FLUX_F58_FORCE_ASCII=1 flux job id --to=f58 $jobid)
name=imp-shell-$(flux getattr rank)-${jobid_f58_ascii}.service)
```
(the rank needs to be the system instance broker rank, and the job id needs to be the job running in the system instance)

I wonder if there is some way we could do mustache substitution in a `flux exec` command line to generate this string for a particular job/rankset.

Some useful tasks to be able to do:
- list  units:  `systemctl --user list-units --type=service`
- get status of unit: `systemctl --user status <unit name>`
- kill, stop, clean, view journal
- list all properties of unit: `systemctl --user show <unit name>`
- show a property value of unit: `systemctl --user show --property=$key <unit name>`
- set a property value of unit: `systemctl --user set-property <unit name> $key=$value`

garlick:
Problem: it is tricky to run systemd commands using flux-exec(1) on a system instance configured with exec.service=sdexec because the user systemd environment is not set up.
    
If not already set, set DBUS_SESSION_BUS_ADDRESS so that systemd --user commands access the dbus instance running as the flux user.

Thus commands like this can work (after sudo to root or flux):
```
flux exec -r 3 systemctl --user list-units --type=service
flux exec -r 3 systemctl --user status imp-shell-3-ftSp9MrEVNw.service
```

Fixes #5901.

### DIFF
--- a/doc/man7/flux-environment.rst
+++ b/doc/man7/flux-environment.rst
@@ -426,6 +426,16 @@ MISCELLANEOUS
    command front end executable used by the ssh connector to start
    :program:`flux relay` on the remote system.
 
+.. envvar:: DBUS_SESSION_BUS_ADDRESS
+
+   :man1:`flux-exec` sets this to point to the Flux instance owner's
+   D-Bus instance, to ensure that a remote invocation of
+   :option:`systemctl --user` accesses the service manager for the Flux
+   instance owner.  This is helpful when debugging a system instance
+   configured to launch jobs with systemd, as described in
+   :man5:`flux-config-exec`.
+
+
 .. _sub_command_environment:
 
 SUB-COMMAND ENVIRONMENT
--- a/doc/test/spell.en.pws
+++ b/doc/test/spell.en.pws
@@ -859,3 +859,4 @@ unsatisfiable
 validators
 hostpids
 Xauth
+dbus
--- a/src/cmd/flux-exec.c
+++ b/src/cmd/flux-exec.c
@@ -646,6 +646,19 @@ int main (int argc, char *argv[])
                                      &imp_path);
     }
 
+    /* Allow systemd commands to work on flux systemd instance by
+     * setting DBUS_SESSION_BUS_ADDRESS if not already set.
+     * See flux-framework/flux-core#5901
+     */
+    const char *security_owner;
+    if (!(security_owner = flux_attr_get (h, "security.owner")))
+        log_err_exit ("failed to fetch security.owner attribute");
+    (void)flux_cmd_setenvf (cmd,
+                            0,
+                            "DBUS_SESSION_BUS_ADDRESS",
+                            "unix:path=/run/user/%s/bus",
+                            security_owner);
+
     /* Get input ranks from --jobid if given:
      */
     if (optparse_getopt (opts, "jobid", &optargp) > 0) {
--- a/t/t0005-exec.t
+++ b/t/t0005-exec.t
@@ -269,4 +269,17 @@ test_expect_success 'stdin broadcast -- long lines' '
 	done
 '
 
+test_expect_success 'dbus environment variable is set' '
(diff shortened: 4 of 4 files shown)


chu11 (approved):
My only thought is if we want to set DBUS_SESSION_BUS_ADDRESS only if we know stuff is running under systemd.  It's probably not a huge deal to add it in the environment either way.

I guess we'd have to check the flux config ... which assumes we're running `flux exec` from the management node?  Meh ... maybe not a net win.

One other comment.  If verbose output desired, output "setting DBUS_SESSION_BUS_ADDRESS=fooobar".  It'd be convenient to have that knowledge for cut and paste if I later ssh to a node to do more debugging.

garlick:
We aren't saying much with the `-v` flag now.  Do you think it's useful to single this one out?

I think the opportunity for this to cause  any surprises is low.  Can you think of a situation where someone might get confused?

grondo:
Will regular systemd commands still work under flux exec after this change?

chu11:
> We aren't saying much with the -v flag now. Do you think it's useful to single this one out?

It's not super useful, just something I thought of.  I'm not sure how you did your systemd devel/debug, but I had a file of "export DBUS ... = foo" that I would regularly cut and paste.  So thought a verbose output could be useful :-)

garlick:
> Will regular systemd commands still work under flux exec after this change?

Yes (just verified, whew).  My understanding is this variable only affects the "session bus" (used by `systemctl --user`) as opposed to the "system bus".

> I had a file of "export DBUS ... = foo" that I would regularly cut and paste

I had a collection of wrapper scripts :-).

garlick:
Would be OK if I just added this variable to `flux-environment(7)` and mention there that it is set by `flux-exec(1)`?

garlick:
I pushed that change.  OK to set MWP?

doc/man7/flux-environment.rst
```diff
@@ -426,6 +426,15 @@ MISCELLANEOUS
    command front end executable used by the ssh connector to start
    :program:`flux relay` on the remote system.
 
+.. envvar:: DBUS_SESSION_BUS_ADDRESS
+
+   :man1:`flux-exec` sets this to point to the Flux instance owner's
+   D-Bus instance, to ensure that a remote invocation of
+   :option:`systemctl --user` accesses Flux's service manager.  This is
```
grondo, line 433: Should this be "accesses the service manager for the instance user"?

doc/man7/flux-environment.rst
```diff
@@ -426,6 +426,15 @@ MISCELLANEOUS
    command front end executable used by the ssh connector to start
    :program:`flux relay` on the remote system.
 
+.. envvar:: DBUS_SESSION_BUS_ADDRESS
```
chu11, line 429: typo commit message "nto" -> "not"

doc/man7/flux-environment.rst
garlick: Fixed - er, I used "for the Flux instance owner".  Hope that's OK.

chu11:
i think this looks good to merge!

PR #5907, opened 2024-04-22
add options for offline config file parsing

garlick:
Problem: as noted in #5898, it may be useful for a sys admin to be able to probe the installed flux configuration before launching flux.

Add two new capabilities:
-  add  `flux config get --config-path=PATH` option (with special PATH names of `system`, `security`, and `imp`)
- enhance `flux R parse-config PATH` to follow `resource.path` if defined

### DIFF
--- a/doc/man1/flux-config.rst
+++ b/doc/man1/flux-config.rst
@@ -54,6 +54,16 @@ scripts.
    *fsd* prints the value in its human-readable, string form. *fsd-integer*
    and *fsd-real* print the value in integer and real seconds, respectively.
 
+.. option:: -c, --config-path=PATH
+
+   Read configuration from PATH instead of fetching configuration from local
+   broker.  If PATH is a directory, then read all TOML files from that
+   directory. If PATH is a file, then load configuration as JSON if the file
+   extension is ``.json``, otherwise load the file as TOML.  As a special case,
+   ``system``, ``security``, and ``imp`` may be used as shorthand for the
+   compiled-in paths to system configuration objects.
+
+
 builtin
 -------
 
--- a/doc/man1/flux-config.rst
+++ b/doc/man1/flux-config.rst
@@ -76,6 +76,12 @@ configuration key names.  This command is available to all users.
    specific values if found to be in tree.  This enables Flux testing without
    requiring installation.
 
+.. option:: --intree
+   Force :program:`flux config builtin` to return in-tree paths.
+
+.. option:: --installed
+   Force :program:`flux config builtin` to return installed paths.
+
 load
 ----
 
--- a/t/t0026-flux-R.t
+++ b/t/t0026-flux-R.t
@@ -367,4 +367,20 @@ test_expect_success 'flux R parse-config detects invalid property' '
 	test_must_fail flux R parse-config conf
 '
 
+test_expect_success 'flux R parse-config also works with resource.path' '
+	flux R encode -r 0-1 >R.path &&
+	cat <<-EOF >conf/resource.toml &&
+	resource.path = "R.path"
+	EOF
+	flux R parse-config conf
+'
+test_expect_success 'flux R parse-config fails when resource.path = bad R' '
+	echo "bad json" >R.path &&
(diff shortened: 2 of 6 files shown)


doc/man1/flux-config.rst
```diff
@@ -54,6 +54,16 @@ scripts.
    *fsd* prints the value in its human-readable, string form. *fsd-integer*
    and *fsd-real* print the value in integer and real seconds, respectively.
 
+.. option:: -c, --config-path=PATH
+
+   Set the PATH to broker configuration. If PATH is a directory, then
```
grondo, line 59: Suggestion: "Read configuration from PATH instead of fetching configuration from local broker." or similar. "Set the PATH to broker configuration" sounds a bit like there's supposed to be an active broker with this configuration.

t/t0026-flux-R.t
```diff
@@ -367,4 +367,20 @@ test_expect_success 'flux R parse-config detects invalid property' '
 	test_must_fail flux R parse-config conf
 '
 
+test_expect_success 'flux R parse-config also works with resource.path' '
```
grondo, line 370: Commit message typo: s/resource.file/resource.path`

t/t2806-config-cmd.t
```diff
@@ -193,5 +193,29 @@ test_expect_success 'flux-config reload fails as guest' '
...
+'
+test_expect_success 'flux-config get --config-path=system fails' '
+	test_must_fail flux config get --config-path=system
+'
+test_expect_success 'flux-config get --config-path=security fails' '
+	test_must_fail flux config get --config-path=security
+'
+test_expect_success 'flux-config get --config-path=imp fails' '
+	test_must_fail flux config get --config-path=imp
+'
```
grondo, line 219: Maybe a comment on why these tests should fail. I would have expected success?

grondo (approved):
This LGTM. Nice improvement!  Just some trivial comments and one question about the tests.

t/t2806-config-cmd.t
garlick: They are failing in CI so good point!

system/security/imp are aliases for compiled in paths so whether this succeeds or fails depends on whether good configs are installed at those paths.

I'm a little puzzled how to test.  Before your comments I pushed a change like this but now the test is kind of meaningless:
```
test_expect_success 'flux-config get --config-path=imp is special' '
        test_might_fail flux config get --config-path=imp 2>imp.err
        test -z imp.err || grep conf.d imp.err
'
```
Pondering.

garlick: OK, I pushed that change.  It was slightly tricky to determine  the configured `$sysconfdir` except indirectly through  `flux config builtin`, but `flux config builtin` will always give in-tree paths when run from the build tree.

I added `--intree` and `--installed` options to `flux config builtin` and then take the dirname of `flux config builtin --installed rc1_path`.  Not super satisfying :shrug:

grondo: well that seems useful to me for testing, and now the tests actually test something, so seems good to me!

garlick:
Thanks!  OK, setting MWP.

grondo:
I did think of one thing belatedly. `flux alloc` and `flux batch`  allow a named config for the instance to be specified on the command line using the following rules:
>               · If none of the above conditions match, then the argument  NAME
>                 is  assumed  to  refer  to  a "named" config file NAME.toml or
>                 NAME.json within the following search path, in order of prece‐
>                 dence:
>
>                · XDG_CONFIG_HOME/flux/config  or $HOME/.config/flux/config if
>                  XDG_CONFIG_HOME is not set
>
>                · $XDG_CONFIG_DIRS/flux/config  or   /etc/xdg/flux/config   if
>                  XDG_CONFIG_DIRS is not set. Note that XDG_CONFIG_DIRS may be
>                  a colon-separated path.

It would be cool if users could use `flux config get -c NAME` to check the parsing and dump values of these named configs as well. However, they can also just use the path to those files, so not a big deal I suppose.

PR #5908, opened 2024-04-22
cron: fix race if sync event changed or disabled

chu11:
Problem: If a sync event is configured, it can lead to cron jobs being added to the deferred list.  If the sync event is later changed or disabled, it is possible jobs on the deferred list may never be executed again.

Handle this race:

- if the sync event is changed, restart the message handler if there are any jobs currently on the deferred list.
- if the sync event is disabled, run any jobs currently on the deferred list.

----

I didn't add any tests b/c this is racy and would require some instrumentation.  Or alternately, could do something nuts like have a cron interval of .1 seconds and regularly change event/disable event for a few seconds, and make sure cron job is still running afterwards?  I guess it could be a LONG_TEST.


grondo (approved):
LGTM!

PR #5910, opened 2024-04-22
job-manager: improve handling of offline ranks in job prolog

grondo:
Recently, jobs were consistently failing with an exception `job prolog failed with exit code=1` without any other details in the logs or job eventlog. It turned out that the scheduler was handing out offline nodes, and the prolog fails in this case, but due to the use of an incorrect log level, the log message for this condition was suppressed. Also, users were confused by the lack of detail in the job exception.

This PR improves prolog handling of offline ranks by

 - using the correct log level for the warning "prolog: rank[s] XX offline. Skipping."
 - raising a non-fatal job exception with this same detail (that some broker ranks were offline) so that a user is notified in the job eventlog of the reason the prolog will fail.

`flux perilog-run` can't raise the fatal exception itself, since would then cancel the prolog, which may be running on other ranks. The nonzero exit status from the prolog will raise the fatal exception, so a job eventlog failing in this manner might look like this:

```
[Apr22 22:39] submit userid=6885 urgency=16 flags=0 version=1
[  +0.011808] validate
[  +0.022818] depend
[  +0.022857] priority priority=16
[  +0.023847] alloc annotations={"sched":{"resource_summary":"rank[0-3]/core0"}}
[  +0.024318] prolog-start description="job-manager.prolog"
[  +0.253649] exception type="prolog" severity=1 note="rank 3 offline" userid=6885
[  +0.384952] exception type="prolog" severity=0 note="prolog exited with exit code=1" userid=6885
[  +0.385053] prolog-finish description="job-manager.prolog" status=256
[  +0.385087] free
[  +0.385109] clean
```

which is a bit better than completely silent failure.

### DIFF
--- a/t/t2274-manager-perilog.t
+++ b/t/t2274-manager-perilog.t
@@ -20,6 +20,7 @@ test_under_flux 4 full \
     -o,--config-path=$(pwd)/config,-Stbon.topo=kary:4 \
     --test-exit-mode=leader
 
+OFFLINE_PLUGIN=${FLUX_BUILD_DIR}/t/job-manager/plugins/.libs/offline.so
 startctl="flux python ${SHARNESS_TEST_SRCDIR}/scripts/startctl.py"
 
 flux setattr log-stderr-level 1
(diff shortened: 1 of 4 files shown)


t/t2274-manager-perilog.t
```diff
@@ -340,9 +341,35 @@ test_expect_success 'perilog: run job across all 4 ranks' '
...
+	flux jobtap load perilog.so
+'
+test_expect_success 'perilog: prolog with offline ranks raises sev 1 exception' '
+	id=$(flux submit -N4 -n4 true) &&
+	flux job wait-event -vt 15 -m severity=1 $id exception &&
+	flux job wait-event -t 15 $id clean &&
+	test_must_fail flux job attach $id &&
+	flux jobtap remove offline.so
+'
+test_expect_success 'perilog: offline ranks are logeed by prolog' '
```
garlick, line 368: s/logeed/logged/

garlick (approved):
LGTM!  I just spotted one nit.

t/t2274-manager-perilog.t
grondo: Fixed that and I'll set MWP. Thanks

PR #5913, opened 2024-04-25
job-exec: support config reload

chu11:
Problem: job-exec requires the job-exec module to be reloaded to read changes to configuration.  This is inconvenient and can affect running jobs.

Support reading/loading configuration changes via the config-reload message callback.  If anything is configured on the command line during module load time, that configuration still override's anything in broker configuration.

----

only minor thing is I called the exec_implementation function "reload" instead of "config_reload", just b/c I didn't want one function to be much longer than any other function and muck all the pretty alignment that already exists.  "reload" could be ambiguous, like this callback is for when the module is reloaded?? 

(edit: don't have docker on system I'm on today, so some tests are not tested ... we'll see if they pass in CI, there may be some iteration if not).

### DIFF
--- a/src/modules/job-exec/exec.c
+++ b/src/modules/job-exec/exec.c
@@ -609,9 +609,13 @@ static void exec_exit (struct jobinfo *job)
     job->data = NULL;
 }
 
-static int exec_config (flux_t *h, int argc, char **argv)
+static int exec_config (flux_t *h,
+                        const flux_conf_t *conf,
+                        int argc,
+                        char **argv,
+                        flux_error_t *errp)
 {
-    return config_init (h, argc, argv);
+    return config_setup (h, conf, argc, argv, errp);
 }
 
 static int exec_stats (json_t **stats)
--- a/src/modules/job-exec/exec_config.c
+++ b/src/modules/job-exec/exec_config.c
@@ -27,12 +27,16 @@
 
 static const char *default_cwd = "/tmp";
 
+struct exec_config {
+    const char *default_job_shell;
+    const char *flux_imp_path;
+    const char *exec_service;
+    int exec_service_override;
+    json_t *sdexec_properties;
+};
+
 /* Global configs initialized in config_init() */
-static const char *default_job_shell;
-static const char *flux_imp_path;
-static const char *exec_service;
-static int exec_service_override;
-static json_t *sdexec_properties;
+static struct exec_config exec_conf;
 
 static const char *jobspec_get_job_shell (json_t *jobspec)
 {
@@ -52,7 +56,7 @@ const char *config_get_job_shell (struct jobinfo *job)
     const char *path = NULL;
     if (job)
         path = jobspec_get_job_shell (job->jobspec);
-    return path ? path : default_job_shell;
+    return path ? path : exec_conf.default_job_shell;
 }
 
 static const char *jobspec_get_cwd (json_t *jobspec)
@@ -81,22 +85,22 @@ const char *config_get_cwd (struct jobinfo *job)
 
 const char *config_get_imp_path (void)
 {
-    return flux_imp_path;
+    return exec_conf.flux_imp_path;
 }
 
 const char *config_get_exec_service (void)
 {
-    return exec_service;
+    return exec_conf.exec_service;
 }
 
 bool config_get_exec_service_override (void)
 {
-    return exec_service_override;
+    return exec_conf.exec_service_override;
 }
 
 json_t *config_get_sdexec_properties (void)
 {
-    return sdexec_properties;
+    return exec_conf.sdexec_properties;
 }
 
 static int config_add_stats_string (json_t *o,
@@ -151,22 +155,24 @@ int config_get_stats (json_t **config_stats)
                                  default_cwd) < 0
         || config_add_stats_string (o,
                                     "default_job_shell",
-                                    default_job_shell) < 0
+                                    exec_conf.default_job_shell) < 0
         || config_add_stats_string (o,
                                     "flux_imp_path",
-                                    flux_imp_path) < 0
+                                    exec_conf.flux_imp_path) < 0
         || config_add_stats_string (o,
                                     "exec_service",
-                                    exec_service) < 0)
+                                    exec_conf.exec_service) < 0)
         goto error;
 
     if (config_add_stats_int (o,
                               "exec_service_override",
-                              exec_service_override) < 0)
+                              exec_conf.exec_service_override) < 0)
         goto error;
 
-    if (sdexec_properties) {
-        if (json_object_set (o, "sdexec_properties", sdexec_properties) < 0)
+    if (exec_conf.sdexec_properties) {
+        if (json_object_set (o,
+                             "sdexec_properties",
+                             exec_conf.sdexec_properties) < 0)
             goto error;
     }
 
@@ -176,6 +182,15 @@ int config_get_stats (json_t **config_stats)
 error:
     ERRNO_SAFE_WRAP (json_decref, o);
     return -1;
+}
+
+static void exec_config_init (struct exec_config *ec)
+{
+    ec->default_job_shell = flux_conf_builtin_get ("shell_path", FLUX_CONF_AUTO);
+    ec->flux_imp_path = NULL;
+    ec->exec_service = "rexec";
+    ec->exec_service_override = 0;
+    ec->sdexec_properties = NULL;
 }
 
 /*  Initialize configurations for use by job-exec bulk-exec
@@ -193,18 +208,14 @@ int config_init (flux_t *h, int argc, char **argv)
      *
      * So we must re-initialize globals on each config_init().
      */
-    default_job_shell = flux_conf_builtin_get ("shell_path", FLUX_CONF_AUTO);
-    flux_imp_path = NULL;
-    exec_service = "rexec";
-    exec_service_override = 0;
-    sdexec_properties = NULL;
+    exec_config_init (&exec_conf);
 
     /*  Check configuration for exec.job-shell */
     if (flux_conf_unpack (flux_get_conf (h),
                           &err,
                           "{s?{s?s}}",
                           "exec",
-                            "job-shell", &default_job_shell) < 0) {
+                            "job-shell", &exec_conf.default_job_shell) < 0) {
         flux_log (h, LOG_ERR,
                   "error reading config value exec.job-shell: %s",
                   err.text);
@@ -216,7 +227,7 @@ int config_init (flux_t *h, int argc, char **argv)
                           &err,
                           "{s?{s?s}}",
                           "exec",
-                            "imp", &flux_imp_path) < 0) {
+                            "imp", &exec_conf.flux_imp_path) < 0) {
         flux_log (h, LOG_ERR,
                   "error reading config value exec.imp: %s",
                   err.text);
@@ -228,8 +239,9 @@ int config_init (flux_t *h, int argc, char **argv)
                           &err,
                           "{s?{s?s s?b}}",
                           "exec",
-                            "service", &exec_service,
-                            "service-override", &exec_service_override) < 0) {
+                            "service", &exec_conf.exec_service,
+                            "service-override",
+                              &exec_conf.exec_service_override) < 0) {
         flux_log (h, LOG_ERR,
                   "error reading config value exec.service: %s",
                   err.text);
@@ -241,22 +253,23 @@ int config_init (flux_t *h, int argc, char **argv)
                           &err,
                           "{s?{s?o}}",
                           "exec",
-                            "sdexec-properties", &sdexec_properties) < 0) {
+                            "sdexec-properties",
+                              &exec_conf.sdexec_properties) < 0) {
         flux_log (h, LOG_ERR,
                   "error reading config table exec.sdexec-properties: %s",
                   err.text);
         return -1;
     }
-    if (sdexec_properties) {
+    if (exec_conf.sdexec_properties) {
         const char *k;
         json_t *v;
 
-        if (!json_is_object (sdexec_properties)) {
+        if (!json_is_object (exec_conf.sdexec_properties)) {
             flux_log (h, LOG_ERR, "exec.sdexec-properties is not a table");
             errno = EINVAL;
             return -1;
         }
-        json_object_foreach (sdexec_properties, k, v) {
+        json_object_foreach (exec_conf.sdexec_properties, k, v) {
             if (!json_is_string (v)) {
                 flux_log (h, LOG_ERR,
                           "exec.sdexec-properties.%s is not a string",
@@ -271,11 +284,11 @@ int config_init (flux_t *h, int argc, char **argv)
         /* Finally, override values on cmdline */
         for (int i = 0; i < argc; i++) {
             if (strstarts (argv[i], "job-shell="))
-                default_job_shell = argv[i]+10;
+                exec_conf.default_job_shell = argv[i]+10;
             else if (strstarts (argv[i], "imp="))
-                flux_imp_path = argv[i]+4;
+                exec_conf.flux_imp_path = argv[i]+4;
             else if (strstarts (argv[i], "service="))
-                exec_service = argv[i]+8;
+                exec_conf.exec_service = argv[i]+8;
         }
     }
 
--- a/src/modules/job-exec/exec_config.c
+++ b/src/modules/job-exec/exec_config.c
@@ -24,6 +24,7 @@
 #include "exec_config.h"
 #include "ccan/str/str.h"
 #include "src/common/libutil/errno_safe.h"
+#include "src/common/libutil/errprintf.h"
 
 static const char *default_cwd = "/tmp";
 
@@ -196,8 +197,13 @@ static void exec_config_init (struct exec_config *ec)
 /*  Initialize configurations for use by job-exec bulk-exec
  *  implementation
  */
-int config_init (flux_t *h, int argc, char **argv)
+int config_setup (flux_t *h,
+                  const flux_conf_t *conf,
+                  int argc,
+                  char **argv,
+                  flux_error_t *errp)
 {
+    struct exec_config tmpconf;
     flux_error_t err;
 
     /* Per trws comment in 97421e88987535260b10d6a19551cea625f26ce4
@@ -206,72 +212,72 @@ int config_init (flux_t *h, int argc, char **argv)
      * dlclose, so a subsequent dlopen doesn't re-clear globals and
      * similar.
      *
-     * So we must re-initialize globals on each config_init().
+     * So we must re-initialize globals everytime we reload the module.
      */
-    exec_config_init (&exec_conf);
+    exec_config_init (&tmpconf);
 
     /*  Check configuration for exec.job-shell */
-    if (flux_conf_unpack (flux_get_conf (h),
+    if (flux_conf_unpack (conf,
                           &err,
                           "{s?{s?s}}",
                           "exec",
-                            "job-shell", &exec_conf.default_job_shell) < 0) {
-        flux_log (h, LOG_ERR,
-                  "error reading config value exec.job-shell: %s",
-                  err.text);
+                            "job-shell", &tmpconf.default_job_shell) < 0) {
+        errprintf (errp,
+                   "error reading config value exec.job-shell: %s",
+                   err.text);
         return -1;
     }
 
     /*  Check configuration for exec.imp */
-    if (flux_conf_unpack (flux_get_conf (h),
+    if (flux_conf_unpack (conf,
                           &err,
                           "{s?{s?s}}",
                           "exec",
-                            "imp", &exec_conf.flux_imp_path) < 0) {
-        flux_log (h, LOG_ERR,
-                  "error reading config value exec.imp: %s",
-                  err.text);
+                            "imp", &tmpconf.flux_imp_path) < 0) {
+        errprintf (errp,
+                   "error reading config value exec.imp: %s",
+                   err.text);
         return -1;
     }
 
     /*  Check configuration for exec.service and exec.service-override */
-    if (flux_conf_unpack (flux_get_conf (h),
+    if (flux_conf_unpack (conf,
                           &err,
                           "{s?{s?s s?b}}",
                           "exec",
-                            "service", &exec_conf.exec_service,
+                            "service", &tmpconf.exec_service,
                             "service-override",
-                              &exec_conf.exec_service_override) < 0) {
-        flux_log (h, LOG_ERR,
-                  "error reading config value exec.service: %s",
-                  err.text);
+                              &tmpconf.exec_service_override) < 0) {
+        errprintf (errp,
+                   "error reading config value exec.service: %s",
+                   err.text);
         return -1;
     }
 
     /*  Check configuration for exec.sdexec-properties */
-    if (flux_conf_unpack (flux_get_conf (h),
+    if (flux_conf_unpack (conf,
                           &err,
                           "{s?{s?o}}",
                           "exec",
                             "sdexec-properties",
-                              &exec_conf.sdexec_properties) < 0) {
-        flux_log (h, LOG_ERR,
+                              &tmpconf.sdexec_properties) < 0) {
+        errprintf (errp,
                   "error reading config table exec.sdexec-properties: %s",
                   err.text);
         return -1;
     }
-    if (exec_conf.sdexec_properties) {
+    if (tmpconf.sdexec_properties) {
         const char *k;
         json_t *v;
 
-        if (!json_is_object (exec_conf.sdexec_properties)) {
-            flux_log (h, LOG_ERR, "exec.sdexec-properties is not a table");
+        if (!json_is_object (tmpconf.sdexec_properties)) {
+            errprintf (errp, "exec.sdexec-properties is not a table");
             errno = EINVAL;
             return -1;
         }
-        json_object_foreach (exec_conf.sdexec_properties, k, v) {
+        json_object_foreach (tmpconf.sdexec_properties, k, v) {
             if (!json_is_string (v)) {
-                flux_log (h, LOG_ERR,
+                errprintf (errp,
                           "exec.sdexec-properties.%s is not a string",
                           k);
                 errno = EINVAL;
@@ -284,14 +290,15 @@ int config_init (flux_t *h, int argc, char **argv)
         /* Finally, override values on cmdline */
         for (int i = 0; i < argc; i++) {
             if (strstarts (argv[i], "job-shell="))
-                exec_conf.default_job_shell = argv[i]+10;
+                tmpconf.default_job_shell = argv[i]+10;
             else if (strstarts (argv[i], "imp="))
-                exec_conf.flux_imp_path = argv[i]+4;
+                tmpconf.flux_imp_path = argv[i]+4;
             else if (strstarts (argv[i], "service="))
-                exec_conf.exec_service = argv[i]+8;
+                tmpconf.exec_service = argv[i]+8;
         }
     }
 
+    exec_conf = tmpconf;
     return 0;
 }
 
--- a/src/modules/job-exec/exec_config.c
+++ b/src/modules/job-exec/exec_config.c
@@ -26,11 +26,13 @@
 #include "src/common/libutil/errno_safe.h"
 
 static const char *default_cwd = "/tmp";
-static const char *default_job_shell = NULL;
-static const char *flux_imp_path = NULL;
-static const char *exec_service = "rexec";
-static int exec_service_override = 0;
-static json_t *sdexec_properties = NULL;
+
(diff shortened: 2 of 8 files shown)


chu11:
hmmm test failures on alpine

```
2024-04-25T18:40:16.702079Z job-exec.debug[0]: using default shell path 180
```

almost feels like mem corruption or something not initialized correctly on module reload?  Noticed on another test later

```
  2024-04-25T18:40:19.214669Z job-exec.debug[0]: using default shell path 
  2024-04-25T18:40:19.214684Z job-exec.debug[0]: using imp path er.insmod (with helper)
  2024-04-25T18:40:19.214689Z job-exec.debug[0]: using exec service ath (override disabled, sdexec properties not set)
```

looks like mem corruption or something not being initialized correctly

grondo:
Oh, there was some specific issue with musl (or whatever libc is used by alpine) here. It doesn't really reload DSOs on dlclose()/dlopen() so you'll need a memset().

chu11:
> Oh, there was some specific issue with musl (or whatever libc is used by alpine) here. It doesn't really reload DSOs on dlclose()/dlopen() so you'll need a memset().

Ahhh thanks, I was super confused as I began looking into it.  So really gotta reset/clear those settings on a reload.

grondo:
More explanation in 97421e88987535260b10d6a19551cea625f26ce4 by @trws :
>```    testsuite: fix job-exec use-after-free on Alpine
>    
>    Problem: t2403-job-exec-conf.t fails on Alpine linux with an
>    assertion failure when the job-exec module is reloaded.
>    
>    Assertion failed: mh->magic == HANDLER_MAGIC (msg_handler.c:
>      flux_msg_handler_destroy: 601)
>    
>    The musl libc loader doesn't actually unload objects on dlclose, so a
>    subsequent dlopen doesn't re-clear globals and similar.  The testenv
>    exec implementation was accidentally storing a context pointer across
>    a reload.
>    
>    Set the context to NULL after it is destroyed so it will not be reused.
>```

chu11:
re-pushed, re-initializing those globals on each reload

t/t2403-job-exec-conf.t
```diff
@@ -100,5 +100,79 @@ test_expect_success 'job-exec: bad imp config causes module failure' '
...
+	name=exec-service &&
+	mkdir ${name}.d &&
+	cat <<-EOF > ${name}.d/exec.toml &&
+	[exec]
+	service = "bar"
+	EOF
+	( export FLUX_CONF_DIR=${name}.d &&
+	  flux start -s1 flux dmesg > ${name}.log 2>&1
+	) &&
+	grep "using exec service bar" ${name}.log
```
garlick, line 182: This could be simplfiied by using the broker `--config-path` option which can take a single toml file:
```
        cat <<-EOF > ${name}.toml &&
	[exec]
	service = "bar"
	EOF
	flux start -o,--config-path=${name}.toml -s1 flux dmesg >${name}.log 2>&1 &&
	grep "using exec service bar" ${name}.log
```
Same for the tests that follow

src/modules/job-exec/job-exec.c
```diff
@@ -1440,6 +1440,33 @@ static int unload_implementations (struct job_exec_ctx *ctx)
...
+                goto error;
+        }
+        i++;
+    }
+    if (flux_respond (h, msg, NULL) < 0)
+        flux_log_error (h, "error responding to config-reload request");
+    return;
+error:
+    if (flux_respond_error (h, msg, errno, NULL) < 0)
+        flux_log_error (h, "error responding to config-reload request");
```
garlick, line 1467: Returning a human readable error here would be helpful as it is displayed by `flux config reload`.

Add `flux_error_t *` to `impl->reload` and `flux_conf_reload_decode()` perhaps?

src/modules/job-exec/exec.c
```diff
@@ -626,6 +626,7 @@ struct exec_implementation bulkexec = {
     .start =    exec_start,
     .kill =     exec_kill,
     .cancel =   exec_cancel,
+    .reload =   NULL,
```
garlick: Would reconfiig be a better name for the callback?

src/modules/job-exec/exec_config.c
```diff
@@ -24,12 +24,13 @@
...
-static const char *exec_service = "rexec";
-static int exec_service_override = 0;
-static json_t *sdexec_properties = NULL;
+/* Globals initialized in config_init() */
+static const char *default_cwd;
+static const char *default_job_shell;
+static const char *flux_imp_path;
+static const char *exec_service;
+static int exec_service_override;
+static json_t *sdexec_properties;
```
garlick: The `default_` prefixed variables are not set by configuration so they probably should not be set by `config_init()`.  They are just static defaults.

For the values that are configured, could we put them in a struct to avoid re-declaring the "tmp" versions in the function?

src/modules/job-exec/exec_config.c
```diff
@@ -216,6 +242,69 @@ int config_init (flux_t *h, int argc, char **argv)
...
+            else if (strstarts (argv[i], "imp=")) {
+                flux_imp_path = argv[i]+4;
+                flux_imp_path_set_cmdline = true;
+            }
+            else if (strstarts (argv[i], "service=")) {
+                exec_service = argv[i]+8;
+                exec_service_set_cmdline = true;
+            }
+        }
+    }
```
garlick: It might be a little cleaner if the command line args are re-parsed each time there is a reconfiguration rather than adding the `_set_cmdline` flags for each possible thing.

src/modules/job-exec/exec_config.c
```diff
@@ -216,6 +242,69 @@ int config_init (flux_t *h, int argc, char **argv)
...
+    conf_debug (h);
+    return 0;
+}
+
+int config_reload (flux_t *h, const flux_conf_t *conf)
+{
+    if (conf_parse (h, conf) < 0)
+        return -1;
+
+    conf_debug (h);
```
garlick, line 306: This function should not print anything to the logs if the exec configuration has not changed.

Idea:  add an optional "impl" callback that returns a JSON object that is included in the module stats result?  Then the values that are actually in use could be queried by tests directly rather than relying on the logs.

garlick (commented):
I added a few thoughts.  The high level overview is
- exec_conf.c might be cleaned up by creating a struct for config values
- logging all exec config values every time `flux config reload` gets runs is excessive

t/t2403-job-exec-conf.t
chu11: I was copying the style that was in the file already with other tests.  I'll update all the tests.

src/modules/job-exec/exec.c
chu11: Ahhh that's a good one.  I originally did "config_reload" and disliked how long it was.

src/modules/job-exec/exec_config.c
chu11: I thought about that, but disliked "saving off" argv & argc somewhere for re-parse, since it isn't generally available at config-reload time.   Think it's a net win saving it off somewhere?

garlick: Since everything runs within the `mod_main()` function, argc, argv remain valid and wouldn't need to be copied.

src/modules/job-exec/exec_config.c
chu11: > Idea: add an optional "impl" callback that returns a JSON object that is included in the module stats result? Then the values that are actually in use could be queried by tests directly rather than relying on the logs.

I like the idea of adding this stuff to module stats, but not sure what you mean by an optional callback?  Perhaps you were mixing two ideas together?

garlick: Oh I was just suggesting a new callback because the "implementation" probably wouldn't register the stats callback - that would happen in the main.  A callback in the main would need some implementation dependent function to call to get the stats info.

chu11:
re-pushed, tweaking per comments above. discounting that was handled in #5943

- more user friendly errors on config-reload errors
- "reload" is now "reconfig"
- argc/argv is parsed everytime instead of handling a "set on cmdline" flag
- put global cofigs in a struct

the one "gotcha" that took me awhile to figure out is that I can't save off the pointer to strings in the `flux_conf_t` b/c that can fall out of scope.  Wish there was a `flux_conf_copy()` or something.  So instead I copied the "exec" config into a local json object.  Not sure if there's a better way??

grondo:
> Oh crap, there's a flux_conf_incref() .... how did I miss that? Maybe I was looking at a different header file. Ugh ... I'll have to fix that up.


This is how the config update is handled in the `resource` and `job-manager` modules (well, they're different but essentially the same idea):

```c
    if (flux_conf_reload_decode (msg, &conf) < 0)
        goto error;
    if (parse_config (ctx, conf, NULL, NULL, NULL, NULL, NULL, &error) < 0) {
        errstr = error.text;
        goto error;
    }
    if (flux_set_conf (h, flux_conf_incref (conf)) < 0) {
        errstr = "error updating cached configuration";
        goto error;
    }
    if (flux_respond (h, msg, NULL) < 0)
        flux_log_error (h, "error responding to config-reload request");
    return;
```

I.e. the configuration is cached in the module's handle so it remains valid until the next update. This has the benefit that `flux_get_conf()` returns the current config anywhere the handle is available if that is ever useful.

src/modules/job-exec/job-exec.h
```diff
@@ -46,17 +46,23 @@ struct jobinfo;
...
-    int  (*cancel)  (struct jobinfo *job);
-    int  (*stats)   (json_t **stats);
+    int  (*config)   (flux_t *h, int argc, char **argv);
+    void (*unload)   (void);
+    int  (*init)     (struct jobinfo *job);
+    void (*exit)     (struct jobinfo *job);
+    int  (*start)    (struct jobinfo *job);
+    int  (*kill)     (struct jobinfo *job, int signum);
+    int  (*cancel)   (struct jobinfo *job);
+    int  (*stats)    (json_t **stats);
```
grondo, line 60: Do the apparent whitespace changes here belong in a different commit?

src/modules/job-exec/job-exec.h
```diff
@@ -46,17 +46,23 @@ struct jobinfo;
...
+    int  (*reconfig) (flux_t *h,
+                      const flux_conf_t *conf,
+                      int argc,
+                      char **argv,
+                      flux_error_t *errp);
```
grondo, line 65: Oh, if you add the config to the handle as mentioned in another commit, then maybe we don't need a different `config` and `reconfig` operations?

src/modules/job-exec/exec.c
```diff
@@ -644,6 +644,15 @@ static int exec_stats (json_t **stats)
     return -1;
 }
 
+static int exec_reconfig (flux_t *h,
```
grondo, line 647: commit message: s/override's/overrides/

Suggestion: replace 'reading/loading' with reloading?

src/modules/job-exec/exec_config.c
```diff
@@ -283,14 +279,61 @@ int config_init (flux_t *h, int argc, char **argv)
...
+}
+
+int config_reconfig (flux_t *h,
+                     const flux_conf_t *conf,
+                     int argc,
+                     char **argv,
+                     flux_error_t *errp)
+{
+    if (conf_parse (h, conf, argc, argv, errp) < 0)
+        return -1;
```
grondo, line 335: Again, if you stash the conf in the handle as in `job-manager` and `resource`, maybe this function could go away?

src/modules/job-exec/job-exec.c
```diff
@@ -1505,6 +1505,11 @@ static void config_reload_cb (flux_t *h,
         }
         i++;
     }
+    if (flux_set_conf (h, flux_conf_incref (conf)) < 0) {
```
grondo, line 1508: Ah, heh! You did this in the midst of my review. Comments about combining config_init() and config_reconfig() still apply perhaps.

grondo (commented):
Was adding some comments inline, but looks like you've pushed some changes, so hopefully they still make sense...

chu11:
> Was adding some comments inline, but looks like you've pushed some changes, so hopefully they still make sense...

Oops, yeah I re-worked to `flux_set_conf()` and `flux_conf_incref()`.  But I think the comments still apply (forgot to post comment that i did that)

src/modules/job-exec/job-exec.h
chu11: i think it's in the right commit, it's where I define the `reconfig` callback.  Or do you mean in general I should split it out into another commit?

src/modules/job-exec/exec_config.c
```diff
@@ -216,6 +242,69 @@ int config_init (flux_t *h, int argc, char **argv)
               exec_service,
               exec_service_override ? "enabled" : "disabled",
               sdexec_properties ? "set" : "not set");
+}
+
+/*  Initialize common configurations for use by job-exec exec modules.
+ */
```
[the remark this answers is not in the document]
chu11: I'm not sure, that's the comment that was there before :-)  But you're right, that comment doesn't seem correct.  I'll correct it in another commit.  Perhaps it's left over from some design long ago.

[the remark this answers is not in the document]
chu11: ohhh perhaps the design is that other exec implementations could use the same imp config, etc.  ehhh lemme leave it for the time being.  One of the confusing things is that the `bulk-exec` exec implementation is in files called `exec` and `exec_config`.  Perhaps that could be cleaned up another day.

src/modules/job-exec/job-exec.h
grondo: Sorry, I possibly missed something obvious, bit why did every line change?

grondo: Ah, it's to keep the same formatting. Ok, then nevermind 🙂 Probably a good reason to avoid that style in the future, but that's originally my fault!

src/modules/job-exec/exec_config.c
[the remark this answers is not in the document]
grondo: Good point. Sounds good

src/modules/job-exec/exec_config.c
chu11: perhaps i'm missing something, but it seems that in `job-manager` and `resource`, they stash the conf in the handle **after** it has been parsed.  My assumption is that if the new conf is bad/invalid in some way, you don't want to save it?

chu11: That said, it doesn't mean can't share the same function for config, may it be init or reload.  Maybe a general `config_setup()` or `config_parse()` callback.

The only question in my mind is that design wise, could there be a exec implementation in the future that needs a separate init & reload?

src/modules/job-exec/job-exec.h
chu11: yeah, just to keep everything aligned :-)

src/modules/job-exec/exec_config.c
chu11: actually, if we have a shared function, the struct is no longer needed either ... lemme re-work

chu11:
re-pushed, i really like the idea to collapse things, so now I've updated the `config()` exec callback to take a `flux_conf_t` and `flux_error_t`, so now there's just the one function.  The commit series got re-ordered as a result, but the code is a lot simpler and there's not as much changes now.

garlick:
I'm sure you've thought about this but just to be sure - is it safe to invalidate the strings and json object returned by the functions in exec_config.h on reconfig?

BTW this comment at the top of exec_config.c is a bit misleading since it is specific to the bulk-exec implementation
```c
/* Flux job-exec configuration common code */
```
Maybe s/job-exec/bulk-exec/?

Also the header might have a comment stating that the functions should only be called from the bulk-exec implementation and that it's not safe to hold onto returned values and re-enter the reactor.

chu11:
> I'm sure you've thought about this but just to be sure - is it safe to invalidate the strings and json object returned by the functions in exec_config.h on reconfig?

I was pretty sure, but did a double check audit and it does seem safe.  Every caller strdups the string/json or reads it and moves on.

> BTW this comment at the top of exec_config.c is a bit misleading since it is specific to the bulk-exec implementation

Per a thread up above, I think there are a few not-super-clear comments and the files `exec` should probably be `bulk-exec` or similar.  https://github.com/flux-framework/flux-core/pull/5913#discussion_r1589239941  I was thinking I'd clean this up in a separate PR after this one, as I'm not entirely sure how to clean it up right now (e.g. `exec` is the main `bulk-exec` C file, so what should `bulk-exec.c` be renamed to?  Or perhaps we can remove `exec.c` and call `bulk-exec.c` directly? .... lemme start an issue).

> Also the header might have a comment stating that the functions should only be called from the bulk-exec implementation and that it's not safe to hold onto returned values and re-enter the reactor.

That sounds like a good idea to add.

src/modules/job-exec/exec_config.c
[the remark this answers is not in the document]
chu11: ahhh this comment is from a refactoring when the original `sdexec` also existed in `job-exec`.  So the "common" made sense at the time.  Will tweak.

chu11:
re-pushed, adding an extra commit in the beginning of the series to fix up those old comments.  And added a comment in the exec_config.h header about not holding onto returned values.

garlick (commented):
Looking much better!

One thought for cleanup - grondo already mentioned losing the "impl" in the config object.  The other thing I noticed is that `config_get_stats()` could probably  just do a single `json_pack()` instead of breaking out "add_stats_string", "add_stats_int" etc..

Also @grondo mentioned he was getting a valgrind memory error running one of the tests but he wasn't sure it was a real error.  I am seeing this which isn't super helpful:
```
$ FLUX_TEST_VALGRIND=t ./t2403-job-exec-conf.t -v
[snip]
==2382435== HEAP SUMMARY:
==2382435==     in use at exit: 67,803 bytes in 269 blocks
==2382435==   total heap usage: 46,810 allocs, 46,541 frees, 51,988,485 bytes allocated
==2382435== 
==2382435== 38,052 (112 direct, 37,940 indirect) bytes in 7 blocks are definitely lost in loss record 40 of 40
==2382435==    at 0x484DA83: calloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==2382435==    by 0x487F167: flux_reactor_create (reactor.c:70)
==2382435==    by 0x487F296: flux_get_reactor (reactor.c:101)
==2382435==    by 0x11E696: modservice_register (modservice.c:166)
==2382435==    by 0x11CC3D: module_thread (module.c:197)
==2382435==    by 0x4A4EAC2: start_thread (pthread_create.c:442)
==2382435==    by 0x4ADFA03: clone (clone.S:100)
==2382435== 
==2382435== LEAK SUMMARY:
==2382435==    definitely lost: 112 bytes in 7 blocks
==2382435==    indirectly lost: 37,962 bytes in 184 blocks
==2382435==      possibly lost: 0 bytes in 0 blocks
==2382435==    still reachable: 29,601 bytes in 77 blocks
==2382435==         suppressed: 128 bytes in 1 blocks
==2382435== Reachable blocks (those to which a pointer was found) are not shown.
==2382435== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==2382435== 
==2382435== For lists of detected and suppressed errors, rerun with: -s
==2382435== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 1 from 1)
```
I'm not sure if that's the one he's talking about.  I'm not spotting the problem either though.

grondo:
> Also @grondo mentioned he was getting a valgrind memory error running one of the tests but he wasn't sure it was a real error. I am seeing this which isn't super helpful:

Yes, that's the one I'm seeing. I even went as far as to print the reactor usecount when the module is exiting and it is always the same whether the leak is reported or not. :confused: I kind of wonder if this is a false positive because I couldn't track down any problem.

src/modules/job-exec/exec_config.c
```diff
@@ -283,14 +289,19 @@ int config_init (flux_t *h, int argc, char **argv)
...
-                exec_conf.exec_service = argv[i]+8;
+                tmpconf.exec_service = argv[i]+8;
         }
     }
 
+    exec_conf.default_job_shell = tmpconf.default_job_shell;
+    exec_conf.flux_imp_path = tmpconf.flux_imp_path;
+    exec_conf.exec_service = tmpconf.exec_service;
+    exec_conf.exec_service_override = tmpconf.exec_service_override;
+    exec_conf.sdexec_properties = tmpconf.sdexec_properties;
```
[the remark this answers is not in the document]
chu11: tmpconf is on the local stack, do you think it should be made global?  I guess I could `memcpy()` it instead??

t/t2403-job-exec-conf.t
```diff
@@ -93,60 +99,65 @@ test_expect_success 'job-exec: can specify exec service on cmdline' '
...
+	mkdir ${name} &&
+	cat <<-EOF > ${name}/${name}.toml &&
 	[exec]
 	service = "sdexec"
 	[exec.sdexec-properties]
 	MemoryHigh = 42
 	EOF
-	test_must_fail flux start -o,--config-path=${name}.toml -s1 \
+	test_must_fail flux start -o,--config-path=${name}/${name}.toml -s1 \
 		flux dmesg > ${name}.log 2>&1 &&
```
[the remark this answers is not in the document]
chu11: yeah, it was necessary, although maybe the commit message needs more information.  By putting the .toml files in `pwd` directory, other tests `flux config reload` might pick this toml file up by accident.  I could delete the toml file instead, but I thought we'd want to keep as many of them around as we can for debugging.

chu11:
> The other thing I noticed is that config_get_stats() could probably just do a single json_pack() instead of breaking out "add_stats_string", "add_stats_int" etc..

I had the goal of not including configs of things that weren't configured (i.e. `flux_imp_path` is often not set). So I don't think I could use `json_pack()`?

 Alternate approach would be to output json null when something isn't set?  That would allow us to use `json_pack()`

> One thought for cleanup - grondo already mentioned losing the "impl" in the config object.

lemme write up an alternate PR for that one, since that comes from the prior PR.

src/modules/job-exec/exec_config.c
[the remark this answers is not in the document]
garlick: No I was just proposing you use structure assignment instead of manually assigning each member, e.g.
```diff
diff --git a/src/modules/job-exec/exec_config.c b/src/modules/job-exec/exec_config.c
index 4e206b6b5..286138491 100644
--- a/src/modules/job-exec/exec_config.c
+++ b/src/modules/job-exec/exec_config.c
@@ -298,11 +298,7 @@ int config_setup (flux_t *h,
         }
     }
 
-    exec_conf.default_job_shell = tmpconf.default_job_shell;
-    exec_conf.flux_imp_path = tmpconf.flux_imp_path;
-    exec_conf.exec_service = tmpconf.exec_service;
-    exec_conf.exec_service_override = tmpconf.exec_service_override;
-    exec_conf.sdexec_properties = tmpconf.sdexec_properties;
+    exec_conf = tmpconf;
     return 0;
 }
 
```

t/t2403-job-exec-conf.t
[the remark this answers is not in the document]
garlick: Ah well maybe it would be clearer to point FLUX_CONF_DIR=$(pwd)/conf.d and keep that config separate from the single-config file tests rather than move each single config test file into its own directory?

garlick:
Up to you on the json_pack stuff.  I didn't think of the fact that some might not be set.

garlick (approved):
I'll go ahead and approve and after you've cleaned up whatever you think is appropriate we can merge.

src/modules/job-exec/exec_config.c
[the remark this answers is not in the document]
chu11: my memory must be hazy or I'm thinking of ancient compilers, i could have swore in my head this wasn't safe.

chu11:
hmmm on corona I see these three coming up valgrind.  the second one is interesting, as I don't think I'm using testexec ... 

```
==3665266== 
==3665266== HEAP SUMMARY:
==3665266==     in use at exit: 96,035 bytes in 415 blocks
==3665266==   total heap usage: 162,599 allocs, 162,184 frees, 63,184,877 bytes allocated
==3665266== 
==3665266== 560 bytes in 1 blocks are possibly lost in loss record 70 of 90
==3665266==    at 0x4C3D1C3: calloc (vg_replace_malloc.c:1554)
==3665266==    by 0x4015372: UnknownInlinedFun (rtld-malloc.h:44)
==3665266==    by 0x4015372: allocate_dtv (dl-tls.c:371)
==3665266==    by 0x4015DA1: _dl_allocate_tls (dl-tls.c:629)
==3665266==    by 0x52DBE32: pthread_create@@GLIBC_2.2.5 (in /usr/lib64/libpthread-2.28.so)
==3665266==    by 0x11B91B13: ??? (in /usr/lib64/libcuda.so.550.54.15)
==3665266==    by 0x11C5D707: ??? (in /usr/lib64/libcuda.so.550.54.15)
==3665266==    by 0x11B43062: ??? (in /usr/lib64/libcuda.so.550.54.15)
==3665266==    by 0x11BA4677: ??? (in /usr/lib64/libcuda.so.550.54.15)
==3665266==    by 0xFC52519: ??? (in /usr/lib64/hwloc/libcudart.so.12)
==3665266==    by 0xFC558DF: ??? (in /usr/lib64/hwloc/libcudart.so.12)
==3665266==    by 0x52E2E66: __pthread_once_slow (in /usr/lib64/libpthread-2.28.so)
==3665266==    by 0xFC9E658: ??? (in /usr/lib64/hwloc/libcudart.so.12)
==3665266== 
==3665266== 5,436 (48 direct, 5,388 indirect) bytes in 1 blocks are definitely lost in loss record 83 of 90
==3665266==    at 0x4C3D1C3: calloc (vg_replace_malloc.c:1554)
==3665266==    by 0x4EA15AF: safe_malloc (czmq_internal.h:43)
==3665266==    by 0x4EA15AF: fzlistx_new (zlistx.c:111)
==3665266==    by 0x4E6B9AC: handler_stack_push (msg_handler.c:102)
==3665266==    by 0x4E6B9AC: method_hash_add (msg_handler.c:165)
==3665266==    by 0x4E6B9AC: flux_msg_handler_create (msg_handler.c:659)
==3665266==    by 0x13ABCA8D: testexec_ctx_create (testexec.c:462)
==3665266==    by 0x13ABCA8D: testexec_config (testexec.c:487)
==3665266==    by 0x13AB8075: config_reload_cb (job-exec.c:1511)
==3665266==    by 0x4E6ACD9: call_handler (msg_handler.c:344)
==3665266==    by 0x4E6B2BA: dispatch_message (msg_handler.c:380)
==3665266==    by 0x4E6B2BA: handle_cb (msg_handler.c:481)
==3665266==    by 0x4E90F32: ev_invoke_pending (ev.c:3770)
==3665266==    by 0x4E94E27: ev_run (ev.c:4190)
==3665266==    by 0x4E94E27: ev_run (ev.c:4021)
==3665266==    by 0x4E6A14A: flux_reactor_run (reactor.c:124)
==3665266==    by 0x13ABAA3B: mod_main (job-exec.c:1565)
==3665266==    by 0x40FEEE: module_thread (module.c:221)
==3665266== 
==3665266== 43,786 (128 direct, 43,658 indirect) bytes in 8 blocks are definitely lost in loss record 90 of 90
==3665266==    at 0x4C3D1C3: calloc (vg_replace_malloc.c:1554)
==3665266==    by 0x4E69FE2: flux_reactor_create (reactor.c:70)
==3665266==    by 0x4E6A0EE: flux_get_reactor (reactor.c:101)
==3665266==    by 0x41173F: modservice_register (modservice.c:166)
==3665266==    by 0x40FE5D: module_thread (module.c:197)
==3665266==    by 0x52DB1C9: start_thread (in /usr/lib64/libpthread-2.28.so)
==3665266==    by 0x64CDE72: clone (in /usr/lib64/libc-2.28.so)
==3665266== 
==3665266== LEAK SUMMARY:
==3665266==    definitely lost: 176 bytes in 9 blocks
==3665266==    indirectly lost: 49,046 bytes in 240 blocks
==3665266==      possibly lost: 560 bytes in 1 blocks
==3665266==    still reachable: 46,253 bytes in 165 blocks
==3665266==         suppressed: 0 bytes in 0 blocks
==3665266== Reachable blocks (those to which a pointer was found) are not shown.
==3665266== To see them, rerun with: --leak-check=full --show-leak-kinds=all
```

chu11:
I think I figured out the leak, the `testexec` implementation was creating its internal context multiple times b/c we're doing config reload, and that would set up message handlers multiple times.  Didn't appear until we had a single config function for both init & reload.

src/modules/job-exec/exec_config.c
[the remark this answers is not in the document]
garlick: AFAICT this is just normal structure assignment and is equivalent to what was there, but won't have to change each time things are added to the struct.

chu11:
@Mergifyio rebase

PR #5914, opened 2024-04-26
cleanup resource.eventlog and handle remapped ranks

garlick:
This makes some improvements to the management of `resource.eventlog`, stopping the proliferation of information that was not used and did not need to be persistent, and allowing drain events to be able to be replayed on a flux restart even when the host to rank mapping changes.

This should be able to ingest all existing eventlogs, and events produced after this change should be ingestable by past versions.  Thus it does not constrain forward and reverse software updates.

Marking WIP while I consider whether something should be added to prune existing eventlogs that have grown huge.


src/modules/resource/reslog.c
```diff
@@ -136,14 +143,24 @@ int reslog_post_pack (struct reslog *reslog,
...
-        goto error;
-    if (flux_kvs_txn_put (txn, FLUX_KVS_APPEND, RESLOG_KEY, val) < 0)
-        goto error;
-    if (!(f = flux_kvs_commit (reslog->h, NULL, 0, txn)))
-        goto error;
+    if ((flags & EVENT_NO_COMMIT)) {
+        if (!(f = flux_future_create (NULL, NULL)))
+            goto error;
+        flux_future_set_reactor (f, flux_get_reactor (reslog->h));
+        flux_future_set_flux (f, reslog->h);
```
grondo, line 150: Doesn't `flux_future_set_flux(3)` do the equivalent of `flux_future_set_reactor (f, flux_get_reactor (reslog->h));`?

src/modules/resource/drain.c
```diff
@@ -557,6 +558,66 @@ static void undrain_cb (flux_t *h,
...
+ * used as the new rank.  Instead, first check that flux_get_hostbyrank()
+ * differs from 'host'.
+ */
+static void add_target (struct idset *ids,
+                        unsigned int rank,
+                        const char *host,
+                        flux_t *h)
+{
+    if (host) {
+        const char *nhost = flux_get_hostbyrank (h, rank);
```
grondo, line 576: Is there any case where the `hostlist` attribute would not be available at this time?

I think booting from config the hostlist is built from `[bootstrap]`, but in the case where the broker is booting from PMI, do all ranks have to be online if the hostlist is built from PMI? Perhaps this is only a theoretical problem, but I'm thinking of the case of restarting a batch instance. This probably isn't even possible now, but the idea of remapping ranks based on hostnames dynamically made me think this is a step perhaps towards restarting a batch instance on a different set of nodes.

(Sorry for the tangent there)

grondo (approved):
This LGTM. A couple trivial comments (more questions) inline.

Nice clean solution IMO!

src/modules/resource/reslog.c
garlick: Oh you're right, good catch!

src/modules/resource/drain.c
garlick: > do all ranks have to be online if the hostlist is built from PMI

Yes, they all have to be online so this shouldn't be an issue now.

This is going to be a problem though.  We currently have a built in assumption that the `hostlist` attribute is set during bootstrap, before rc1 starts, and that it never changes after that (it is immutable and therefore gets cached on first use in clients).  I'm not sure how to resolve that but you're right when we do this may need to change.

The same problem exists with R reranking.

More thought required.

garlick:
I fixed the future thing and squashed it.

Also tacked on a couple of commits to
- add missing coverage for bad eventlog entries
- make error messages more descriptive when there are problems with the eventlog

garlick:
Coverage got stuck here - restarting.
```
2024-04-26T18:00:45.2066231Z not ok 45 - job-manager: get job IDs of valid jobs
2024-04-26T18:00:45.2066237Z 
2024-04-26T18:00:45.2066492Z expecting success: 
2024-04-26T18:00:45.2067250Z 	while ! test_inactive $(cat valid_ids); do echo retry; sleep 0.1; done
2024-04-26T18:00:45.2067327Z 
2024-04-26T18:00:45.2067601Z retry
2024-04-26T18:00:45.2067736Z retry
2024-04-26T18:00:45.2067908Z retry
2024-04-26T18:00:45.2067992Z retry
2024-04-26T18:00:45.2068064Z retry
```

garlick:
Rebased on current master.  Any reason not  to set MWP on this?

garlick:
I'll go ahead and set MWP.

PR #5922, opened 2024-04-30
job-manager: perilog: do not set working directory on subprocesses

### LINKED ISSUE #5921: do not use `flux_cmd_setcwd()` unless necessary
Problem: There appear to be several places where `flux_cmd_setcwd()` is set to the current working directory of the broker, e.g. in the `perilog` plugin:
https://github.com/flux-framework/flux-core/blob/bff958bfbc4277d0e94204104b28fee77bee8560/src/modules/job-manager/plugins/perilog.c#L328
Not only is this unnecessary, but it forces the subprocess code to fallback to fork/exec instead of using `posix_spawn()`, which can slow things down, especially when the broker is using a lot of memory like on rank 0 of a large instance.

grondo:
Problem: The job-manager perilog plugin manually sets the cwd of any subprocesses run for a prolog or epilog, but this is unnecessary and furthermore forces a fallback to the fork/exec method instead of posix_spawn(3) which is more performant.

Remove the use of `flux_cmd_setcwd()` in the perilog plugin.

Fixes #5921

### DIFF
--- a/src/modules/job-manager/plugins/perilog.c
+++ b/src/modules/job-manager/plugins/perilog.c
@@ -314,7 +314,6 @@ static int run_command (flux_plugin_t *p,
         .on_stdout = io_cb,
         .on_stderr = io_cb
     };
-    char path[PATH_MAX + 1];
 
     if (flux_plugin_arg_unpack (args,
                                 FLUX_PLUGIN_ARG_IN,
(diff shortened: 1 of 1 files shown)


grondo:
Hm, one builder failed here
```
  expecting success: 
  	id=$(flux submit --wait-event=exec.shell.start sh -c "kill -9 \$PPID") &&
  	test_must_fail_or_be_terminated flux job attach $id 2>shell-killed.out &&
  	grep "job shell Killed" shell-killed.out
  
  not ok 17 - attach: reports job shell Killed if job shell is killed
  ```

perhaps a race in this newer tetst. Will restart the build.

grondo:
Thanks!

Issue #5923, opened 2024-04-30
shell segfault when evlog kvs commit times out


grondo:
Noticed some job shell crashes when running heavily oversubscribed nodes (much greater than 1 job shell per core).

I was able to get a run to fail under valgrind, which caught the following error:
```
evlog: Connection timed out: msg=nofile exceeds current max, raising value to ha
rd limit\n"
==4031100== Invalid read of size 8\n"
==4031100==    at 0x405B1C6: flux_future_fulfill_with (future.c:677)\n"
==4031100==    by 0x405B533: chained_continuation (composite_future.c:369)\n"
==4031100==    by 0x407A8D2: ev_invoke_pending (ev.c:3770)\n"
==4031100==    by 0x407E1A4: ev_run (ev.c:4190)\n"
==4031100==    by 0x407E1A4: ev_run (ev.c:4021)\n"
==4031100==    by 0x405212A: flux_reactor_run (reactor.c:124)\n"
==4031100==    by 0x40D2CC: main (shell.c:1743)\n"
==4031100==  Address 0x172d23b8 is 120 bytes inside a block of size 136 free'd\n
"
==4031100==    at 0x4C3AC2B: free (vg_replace_malloc.c:974)\n"
==4031100==    by 0x405A621: flux_future_destroy (future.c:299)\n"
==4031100==    by 0x405A621: flux_future_destroy (future.c:287)\n"
==4031100==    by 0x407A8D2: ev_invoke_pending (ev.c:3770)\n"
==4031100==    by 0x407E1A4: ev_run (ev.c:4190)\n"
==4031100==    by 0x407E1A4: ev_run (ev.c:4021)\n"
==4031100==    by 0x405212A: flux_reactor_run (reactor.c:124)\n"
==4031100==    by 0x40D2CC: main (shell.c:1743)\n"
==4031100==  Block was alloc'd at\n"
==4031100==    at 0x4C3D1C3: calloc (vg_replace_malloc.c:1554)\n"
==4031100==    by 0x405A65A: flux_future_create (future.c:310)\n"
==4031100==    by 0x405B8C1: chained_future_alloc (composite_future.c:435)\n"
==4031100==    by 0x405B8C1: chained_future_create (composite_future.c:457)\n"
==4031100==    by 0x405BCB0: flux_future_and_then (composite_future.c:551)\n"
==4031100==    by 0x44834C: eventlogger_commit_batch (eventlogger.c:176)\n"
[...]

PR #5927, opened 2024-05-01
fix possible use-after-free in chained future implementation

grondo:
This PR fixes the problem with chained futures described in #5923: if a `next` future in a chain is destroyed before the `prev` future, then a segfault due to use-after-free could occur later if `prev` is eventually fulfilled since this references next.
The case where this is likely to occur is when `flux_future_then(3)` is called with a timeout on a `next` future, since the timeout will fulfill `next` with an error without even touching the `prev` future.

The fix is not exactly elegant. A reference to the `chained_future` object is placed in  the `next` future's `aux` list, so that when it is destroyed it can (optionally) destroy `prev`. However, this needs to be avoided in the common case where `prev` is destroyed before `next`, so an aux_item destructor is _also_ placed in `prev` to nullify `cf->prev` so that future is not double freed. Finally, since the underlying `chained_future` object is now referenced in multiple places, and these are not called in any guaranteed order, that structure gets a reference counter and a reference count of 3:

 1. for the presence in `prev` aux_item list as `flux::chained`
 2. for the anonymous linkage in `next` aux_item list
 3. for the anonymous linkage in `prev` aux_item list to call `nullify_prev`

A test is added and was tested under valgrind etc.

### DIFF
--- a/src/common/libflux/composite_future.c
+++ b/src/common/libflux/composite_future.c
@@ -304,6 +304,7 @@ struct continuation_info {
 };
 
 struct chained_future {
+    int refcount;
     bool continued;
     flux_future_t *next;
     flux_future_t *prev;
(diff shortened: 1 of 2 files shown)


src/common/libflux/composite_future.c
```diff
@@ -432,6 +442,18 @@ static void chained_future_init (flux_future_t *f, void *arg)
     fulfill_next (f, cf->next);
 }
 
+void cf_next_destroy (struct chained_future *cf)
```
garlick, line 445: Make static?

garlick (approved):
Excellent!

src/common/libflux/composite_future.c
grondo: ooh, good catch :+1:

grondo:
Ok, fixed the non-static function and will set MWP. Thanks!

PR #5928, opened 2024-05-02
broker: avoid LOST due to EHOSTUNREACH messages during shutdown

garlick:
Problem: as noted in #5881, we observe many "transitioning to LOST due to EHOSTUNREACH" log messages during shutdown of a large system with a flat TBON.

This is due to a deficiency in the shutdown handshake.  Once a tbon child completes rc3, it sends an offline status control message to the parent and immediately disconnects.  The parent processes the status message asynchronously and may attempt to send messages to the child (for example routine broadcasts) after the child has closed the connection, which triggers this log message.

Improve the shutdown handshake by adding an RPC that the child must wait for before disconnecting.

In addition, make the message a little more user friendy so if it does appear it is more clear to users what is going on.

### DIFF
--- a/src/broker/overlay.c
+++ b/src/broker/overlay.c
@@ -124,6 +124,8 @@ struct parent {
     bool hello_error;
     bool hello_responded;
     bool offline;           // set upon receipt of CONTROL_DISCONNECT
+    bool goodbye_sent;
+    flux_future_t *f_goodbye;
     struct rpc_track *tracker;
     struct zmqutil_monitor *monitor;
 };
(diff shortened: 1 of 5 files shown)


garlick:
I ran a quick test of current master and this branch (rebased on current master) to see if there was an impact on the time from `shutdown->finalize` on rank 0.  This is with kary:32.  A little.
```
nodes   master  issue#5881
2       0.23    0.23
8       0.28    0.29
32      0.99    1.01
64      2.25    2.29
128     4.38    4.47
256     8.69    8.84
512     17.53   17.78
```
This was on a single node.

grondo (approved):
LGTM!

garlick:
Cool thanks!  MWP away!

PR #5930, opened 2024-05-02
change default tbon.topo to kary:32

garlick:
Problem: The default tbon.topo is kary:2, which is almost certainly not optimal for any case.

Let's try kary:32.


grondo (approved):
Seems like a good idea!

Issue #5919, opened 2024-04-29
job-exec: possible scaling issue managing many job shells


grondo:
I was running a test instance of size=1536 across 32 real nodes, each with 96 cores. When launching 128N/128p jobs there seemed to be some kind of performance issue in job-exec and jobs were not making progress. Perf report shows a lot of time spent in the libsubprocess watchers:
```
-   96.26%     0.00%  job-exec         libc-2.28.so           [.] __GI___clone (inlined)
     __GI___clone (inlined)
     start_thread
     module_thread
     mod_main
     flux_reactor_run
     ev_run
   - ev_run
      - 84.93% ev_invoke_pending
         - 21.03% remote_out_check_cb
            - 14.38% fbuf_has_line
               - 13.48% cbuf_peek_line
                  - 12.15% cbuf_is_valid
                     - 1.04% cbuf_mutex_is_locked (inlined)
                          1.02% __GI___pthread_mutex_trylock (inlined)
                    0.76% __GI___pthread_mutex_lock (inlined)
              1.22% cbuf_used
              0.98% flux_watcher_stop
         - 17.73% remote_out_prep_cb
            - 14.68% fbuf_has_line
               - 13.93% cbuf_peek_line
                  + 12.53% cbuf_is_valid
                    0.91% __GI___pthread_mutex_lock (inlined)
         - 15.24% remote_in_check_cb
            + 5.03% cbuf_used
              0.73% flux_watcher_stop
         + 10.36% remote_in_prep_cb
           7.62% prepare_cb
           6.98% check_cb
      - 11.32% queue_events (inlined)
           ev_feed_event
```

A `sleep 1` prolog and epilog were configured, but the job-manager did not appear to have high utilization in `perf`, so maybe that is a red herring.

The other thing to note is that the nodes are being heavily oversubscribed in this test, so perhaps that's a contributing factor.

garlick:
Hmm, it seems like the output prep watcher is always active, so while the output channel is open, every event loop will check the buffer for data.  The prep watcher should be stopped when the buffer is empty and started only when it contains consumable data.

We might want to see what the profile looks like with the following change (all subprocess unit tests pass wtih this):
```diff
diff --git a/src/common/libsubprocess/remote.c b/src/common/libsubprocess/remote.c
index 3dd9a3ca1..2e009c994 100644
--- a/src/common/libsubprocess/remote.c
+++ b/src/common/libsubprocess/remote.c
[...]

PR #5932, opened 2024-05-03
libsubprocess: reduce remote output prep/check

chu11:
Problem: Profiling shows that a significant amount of time can be spent in the prep/check of remote subprocess output. This is even in the case when the output buffer is empty.

Enable the remote output prep/check only when the buffer is non-empty.

----

Side note, after trying and flailing on the "input" side of this, I realized that a bunch of code in `libsubprocess`'s remote code simply assumes the prep/check for input is always running.  So that work requires more thought.  Posting this so we can atleast get this in for the next release if I can't figure out the input side by then.

### DIFF
--- a/src/common/libsubprocess/remote.c
+++ b/src/common/libsubprocess/remote.c
@@ -55,8 +55,6 @@ static void start_channel_watchers (flux_subprocess_t *p)
     while (c) {
         flux_watcher_start (c->in_prep_w);
         flux_watcher_start (c->in_check_w);
-        flux_watcher_start (c->out_prep_w);
-        flux_watcher_start (c->out_check_w);
         c = zhash_next (p->channels);
     }
 }
@@ -292,6 +290,17 @@ error:
     c->p->f = NULL;
 }
 
+static bool remote_out_data_available (struct subprocess_channel *c)
+{
+    /* no need to handle failure states, on fatal error, these
+     * reactors are closed */
+    if ((c->line_buffered && fbuf_has_line (c->read_buffer))
+        || (!c->line_buffered && fbuf_bytes (c->read_buffer) > 0)
+        || (c->read_eof_received && !c->eof_sent_to_caller))
+        return true;
+    return false;
+}
+
 static void remote_out_prep_cb (flux_reactor_t *r,
                                 flux_watcher_t *w,
                                 int revents,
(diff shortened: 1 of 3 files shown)


garlick:
Is there a reason remote subprocesses do not use fbuf watchers?  We use them for local subprocess I/O.  Maybe all this is already solved there? (I didn't look)

chu11:
> Is there a reason remote subprocesses do not use fbuf watchers? We use them for local subprocess I/O. Maybe all this is already solved there? (I didn't look)

I can't remember why.

Is your idea that we use fbuf watchers and have those "drive" the callbacks to the user vs the prep/check we have setup?  My only thought is if it's a huge win, since there's a prep/check in the watcher too.  Although, I suppose the watcher already has its "is buffer empty" logic taken care of?

Lemme try and figure out the "input" side of this as well, b/c that's trickier.  Perhaps the truth of why it's non-watcher based is in there.

garlick:
> Is your idea that we use fbuf watchers and have those "drive" the callbacks to the user vs the prep/check we have setup?

Yes.

chu11:
I was finally able to have the remote input stream (i.e. "write_buffer" in the code) work by enabling prep/check only when needed.  There were basically some racy bits with the remote server regarding sending data (i.e. like stdin) before the remote process started and some racy bits with when the remote process has exited.

It's possible this is why the prep/check were enabled all the time, just to make this logic simpler.

given this info, will continue conversation in #5919 regarding potential paths forward

grondo (approved):
Tested this and seems to be working!

chu11:
oh whoops, sorry you reviewed.  Given discussion in #5919 I thought we were going to table this approach, go with the `FLUX_SUBPROCESS_FLAGS_LOCAL_UNBUF` approach.

grondo:
I probably lost track of the discussion, but it would be nice to reduce the high load on the broker due to the current situation. Would this work as an interim mitigation? (guess we could have input from @garlick)

chu11:
> Would this work as an interim mitigation?

Yes, I do believe it would work as a short term mitigation.

grondo:
Let's go ahead and merge then if @garlick approves?

garlick (approved):
LGTM!

grondo:
I set MWP here.

PR #5933, opened 2024-05-03
deps/docker: add 'time' as a dependency for tests

trws:
problem: Apparently bookworm, jammy, focal, and new fedoras don't install /usr/bin/time by default.  I ran into this trying to include time output in a test.

Evidently the "time" builtin in bash is suppressed inside the scope of an eval like we use in sharness, either way this seems less annoying than having to avoid it.

solution: install it, added to each of the docker files as well as the install-deps-*.sh files.

### DIFF
--- a/scripts/install-deps-deb.sh
+++ b/scripts/install-deps-deb.sh
@@ -24,6 +24,7 @@ apt install \
   python3-sphinx \
   aspell \
   aspell-en \
+  time \
   valgrind \
   libmpich-dev \
   jq
--- a/scripts/install-deps-rpm.sh
+++ b/scripts/install-deps-rpm.sh
@@ -24,6 +24,7 @@ yum install \
   python3-sphinx \
   aspell \
   aspell-en \
+  time \
   valgrind-devel \
   mpich-devel \
   jq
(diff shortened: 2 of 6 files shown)


trws:
I have not pushed them yet, and no it wasn't in core, honestly the thought was more "why on earth would _time_ not work in eval?!?" since I would expect the bash builtin to work even if it wasn't there.  Having tripped over it now 3 or 4 times I guessed others might and it was cheap enough to fix.

If you're happy with this I'll start the run to push up the new testenv images momentarily, just give me an ACK.

trws:
Hm, actually, since I'm in here, should I update the fedora dockerfile too? We're currently running 38 I think when current is 40, and 40 needs this same tweak (it's what I'm running on my dev VM).

grondo:
Feel free to update to fedora40 though that process is usually a bit time consuming if you want to hold off.

trws:
Ok, added time to fedora39 so when that gets updated we don't have to remember.  Waiting for the last two aarch64 container builds to finish (emulated, so they're kinda slow), as soon as that happens and they're cleanly posted I'll MWP this.

PR #5936, opened 2024-05-03
resource: upgrade KVS resource.eventlog if needed

### LINKED ISSUE #5931: need to clean old resource.eventlogs
Problem:  #5914 simplifies flux's use of the `resource.eventlog`, but doesn't fix old drain entries that are still tied only to a rank.

We may want to just rewrite the `resource.eventlog` if is found to contain under specified drain entries.  Or just do it  unconditionally to remove the useless entries.

garlick:
Problem: under-specified resource.eventlog drain/undrain events, copious online/offline events, and bloated resource-init events are allowed to remain in resource.eventlog after recent updates that address these issues for new events.
    
If the resource.eventlog contains a resource-init event, rewrite the eventlog in the KVS with an upgraded version:
- remove resource-init, online, offline events
- add nodelist to under-specified drain/undrain events
    
Fixes #5931

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -17,6 +17,7 @@ extend-exclude = [
   "src/common/libmissing/*.[ch]",
   "t/hwloc-data/*",
   "doc/test/spell.en.pws",
+  "t/resource/resource.eventlog.*",
 ]
 
 [default.extend-words]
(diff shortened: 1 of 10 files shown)


grondo (approved):
I didn't spot any problems with this one. :+1:

PR #5937, opened 2024-05-03
job-exec: fix potential use-after-free in bulk-exec implementation

grondo:
Problem: The exit_batch_timer is not destroyed in bulk_exec_destroy(), which could result in the timer firing on a bulk_exec structure which has already been destroyed. Testing shows that this is a rare, but reproducible condition when bulk_exec_imp_kill() is used. This is possibly because destruction of the bulk_exec structure is tied to that of the future returned from this function. This allows the caller to destroy the future and the bulk_exec object before the timer watcher is cleared, resulting in use-after-free and memory corruption.

Destroy the exit_batch_timer watcher in bulk_exec_destroy().


garlick (approved):
Woo hoo!  Nice sleuthing.

grondo:
Thanks, setting MWP.

PR #5938, opened 2024-05-04
docker-run-checks: ensure we match user's home

trws:
problem: docker-run-checks.sh currently assumes that mounting $HOME to /home/$USER will match the outer system.  Sometimes that is not the case.  Example, my lima-managed vm uses `/home/$USER.linux` as the home directory for the vm user, which causes strange errors because of the mismatch.

solution: map $HOME to $HOME

Assuming this works here, I'll apply this to flux-sched as well.


grondo (approved):
LGTM!

trws:
Something strange happened with the flux-pmix test, it looked like the build succeeded then the next step failed with no output.  Bounced it assuming it's something spurious in actions, but worth noting.

PR #5939, opened 2024-05-04
doc: convert internals docs to readthedocs

garlick:
Problem: there are some flux-core internals documentation that nobody knows about [here](https://github.com/flux-framework/flux-core/tree/master/src/broker/doc) and [here](https://github.com/flux-framework/flux-core/wiki/kvs).

Add a _Resources for Flux Developers_ section to the flux-core docs and convert those documents to chapters in it.

These docs do need an update but it's a start.


grondo (approved):
LGTM! Good idea. Should the README in the broker dir be replaced with some breadcrumbs pointing to the new docs location, since a developer working on code might be more apt to look there first?

Eh, I guess it was under `broker/doc/README` so probably not necessary, just a thought.

garlick:
Thanks, yeah I think that README buried down in the sources was probably invisible so leaving a pointer there is likely not helpful.

I fixed up a few typos and formatting errors I spotted this morning with fresh eyes, and added the new files to the Makefile.am (oops).  Probably good to merge now so I'll set MWP.

grondo:
> Thanks, yeah I think that README buried down in the sources was probably invisible 

I tend to find those when I'm actually working on code, since a README in the current directory is actually quite obvious, whereas it may not occur to the developer to go back to the top level and look under docs for current directory developer information.

garlick:
> I tend to find those when I'm actually working on code, since a README in the current directory is actually quite obvious, whereas it may not occur to the developer to go back to the top level and look under docs for current directory developer information.

Ah OK, well maybe I'll go ahead and add README.md files in the kvs and broker source dirs then!

grondo:
I don't think that's necessary, I was just explaining myself!

garlick:
NP.  Your counterexample disproved my theory!

Issue #5900, opened 2024-04-18
job-exec: add config reload callback


chu11:
there is no callback for a `flux config reload` in job-exec.  This  would be useful, minimally for sdexec properties.

PR #5943, opened 2024-05-05
job-exec: support module stats to see current bulk-exec configuration

chu11:
Per comments in #5913, did a bunch of cleanup/re-work in prep of job-exec config reload (#5900).  

- job-exec module stats that export current config values (did it in a "impl.bulk-exec.config." object).
- modernize tests in t2403-job-exec-conf.t
- update to use module stats to determine conf
- add missing coverage of other job-exec confs

### DIFF
--- a/src/modules/job-exec/job-exec.c
+++ b/src/modules/job-exec/job-exec.c
@@ -271,7 +271,8 @@ int jobinfo_emit_event_pack_nowait (struct jobinfo *job,
     return rc;
 }
 
-static int jobid_exception (flux_t *h, flux_jobid_t id,
+static int jobid_exception (flux_t *h,
+                            flux_jobid_t id,
                             const flux_msg_t *msg,
                             const char *type,
                             int severity,
--- a/src/modules/job-exec/job-exec.c
+++ b/src/modules/job-exec/job-exec.c
@@ -1441,6 +1441,40 @@ static int unload_implementations (struct job_exec_ctx *ctx)
     return 0;
 }
 
+static void stats_cb (flux_t *h,
+                      flux_msg_handler_t *mh,
+                      const flux_msg_t *msg,
+                      void *arg)
+{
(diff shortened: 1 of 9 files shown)


src/modules/job-exec/job-exec.c
```diff
@@ -1440,6 +1440,40 @@ static int unload_implementations (struct job_exec_ctx *ctx)
...
+        if (impl->stats) {
+            if ((*impl->stats) (&stats) == 0 && stats) {
+                if (json_object_set_new (o, impl->name, stats) < 0) {
+                    errno = ENOMEM;
+                    goto error;
+                }
+            }
+        }
+        i++;
+    }
```
garlick, line 1465: I thought only one implementation could be active at runtime?  (Am I confused?)

When I suggested a callback I just meant to integrate any general stats provided in the mainline with those produced by the active plugin.

Also (minor): new code should probably break long parameter lists at one per line, except in the case of "pack" functions and then it should keep keys and values together on a line

grondo: > I thought only one implementation could be active at runtime? (Am I confused?)

Implementations are selected per job, e.g. if `attributes.system.exec.test.run_duration` is set then the test exec implementation is used.

garlick: > Implementations are selected per job,

sorry my bad!

> By "mainline" you mean job-exec module?

Yes but my comment was misplaced due to an incorrect recollection of how stuff works here.

> Ahhh I think you're referring to the flux_respond_pack(),

Also the message handler function prototype.

chu11:
re-pushed, fixing up those parameter lists and the parameter lists I found / cut & pasted from.  Also fixed up a test that still relied on dmesg output from `job-exec`.

chu11:
@grondo thanks, did those small tweaks, will set MWP

Issue #5934, opened 2024-05-03
limit cpus seen by hwloc in Flux jobs


ryanday36:
Is it possible to limit the cpus seen by hwloc to just those assigned to a Flux job? e.g., on fluke if I run with `flux run -n2 -o mpibind=off ...` The job gets two cpus, and I only see those with `numactl -s`, but `lstopo-no-graphics` still lists all four cpus:

```
[day36@fluke95:~]$ flux run -n2 -l -o mpibind=off bash -c "numactl -s | grep -i physcpu && lstopo-no-graphics | grep L2"
0: physcpubind: 2 3 6 7 
1: physcpubind: 2 3 6 7 
0: hwloc/rsmi: Failed to initialize with rsmi_init(): RSMI_STATUS_INIT_ERROR: An error occurred during initialization, during monitor discovery or when when initializing internal data structures
0:       L2 L#0 (256KB) + L1d L#0 (32KB) + L1i L#0 (32KB) + Core L#0
0:       L2 L#1 (256KB) + L1d L#1 (32KB) + L1i L#1 (32KB) + Core L#1
0:       L2 L#2 (256KB) + L1d L#2 (32KB) + L1i L#2 (32KB) + Core L#2
0:       L2 L#3 (256KB) + L1d L#3 (32KB) + L1i L#3 (32KB) + Core L#3
1: hwloc/rsmi: Failed to initialize with rsmi_init(): RSMI_STATUS_INIT_ERROR: An error occurred during initialization, during monitor discovery or when when initializing internal data structures
1:       L2 L#0 (256KB) + L1d L#0 (32KB) + L1i L#0 (32KB) + Core L#0
1:       L2 L#1 (256KB) + L1d L#1 (32KB) + L1i L#1 (32KB) + Core L#1
1:       L2 L#2 (256KB) + L1d L#2 (32KB) + L1i L#2 (32KB) + Core L#2
1:       L2 L#3 (256KB) + L1d L#3 (32KB) + L1i L#3 (32KB) + Core L#3
```

I also tried running with mpibind and setting `-o cpu-affinity=on` and `-o hwloc.xmlfile` with the same results (excepting that you have to `grep Core` rather than `L2` with `-o hwloc.xmlfile`).

For background, this came up because a user is trying to run two jobs on the same node with a multi-physics code that does its own task and thread affinity using hwloc and both jobs appear to be trying to use all of the cpus on the node.

grondo:
[...]

PR #5945, opened 2024-05-06
do not filter caches when loading hwloc topology

grondo:
As discussed in #5934, this PR removes the lines from librlist that are currently filtering out L2/L3/etc caches in the hwloc topology loaded by flux, thus making this information unavailable to components and processes that use this topology, since it is shared from parent to child instance and shell to shell plugin.


wihobbs (approved):
All my simple tests on corona with ompi5 passed off this branch.

grondo:
Thanks I'll set MWP on this one since we implicitly have @trws's ACK as well.

PR #5947, opened 2024-05-07
job-exec: remove layer of stats indirection

chu11:
Problem: All job-exec stats are listed under a key "impl", which is excessive.

Remove the top level key "impl".  Now the top level keys are simply the names of the exec implementations, i.e. "bulk-exec".

Update all tests accordingly.


chu11:
rebased on #5913, since there's going to be necessary changes in tests #5913 added

chu11:
re-pushed now that #5913 is in

garlick (approved):
LGTM!

PR #5948, opened 2024-05-08
job-exec: silence some nuisance logging

grondo:
While testing some other things it was apparent that job-exec is a contributor to nuisance logging. Especially bad was some libsubprocess server logging when data can't be written to processes that have since disappeared due to being killed.

This PR cleans up some of that useless logging. (and fixes one place where it was already meant to be cleaned up)


garlick (approved):
Oh good!

grondo:
Thanks, I've set MWP.

PR #5950, opened 2024-05-08
libsubprocess: remove unused len parameter from flux_subprocess_read()

chu11:
and some comment additions and an extra test.

### DIFF
--- a/src/common/libsubprocess/subprocess.h
+++ b/src/common/libsubprocess/subprocess.h
@@ -190,8 +190,7 @@ int flux_subprocess_write (flux_subprocess_t *p,
 int flux_subprocess_close (flux_subprocess_t *p, const char *stream);
 
 /*
- *  Read up to `len` bytes of unread data from stream `stream`.  To
- *   read all data, specify 'len' of -1.  'stream' can be "stdout",
+ *  Read unread data from stream `stream`.  'stream' can be "stdout",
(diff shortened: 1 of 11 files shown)


chu11:
re-pushed, fixing up a random bug i found

src/common/libsubprocess/subprocess.h
```diff
@@ -215,6 +215,11 @@ const char *flux_subprocess_read (flux_subprocess_t *p,
  *   be NUL terminated.  If no line is available, returns pointer and
  *   length of zero.  User shall not free returned pointer.  Length of
  *   buffer returned can optionally returned in 'lenp'.
+ *
+ *   Use flux_subprocess_read_stream_closed() to determine if stream
+ *   is closed, as length of zero cannot be assumed that subprocess
+ *   has closed the stream (i.e. data is buffered, but not a line
+ *   worth of data).
```
garlick, line 226: I'm finding "as length of zero cannot be assumed that subprocess has closed the stream"   a little hard on the ears :-)

Could this paragraph just be replaced with

> A length of zero may be returned if the stream is closed OR if the stream is line buffered and a line is not yet available.  Use `flux_subprocess_read_stream_closed()` to distinguish between the two.

garlick (approved):
LGMT!  One comment about a comment.

chu11:
@garlick tweaked that comment, will set MWP

PR #5961, opened 2024-05-13
WIP: job-exec: do not allow guest access to the testexec implementation by default

grondo:
This PR disables guest user access to the job-exec `testexec` implementation by default.

Since multiuser testexec jobs are used in the flux-core and flux-accounting testsuites, a new configuration option is added to optionally allow it: `exec.testexec.allow-guests`, which must be set to `true`.

flux-accounting tests will fail until `allow-guests` is set there. However, I was going to wait in case someone had a better idea of how to optionally allow this feature for testing.

### DIFF
--- a/doc/man5/flux-config-exec.rst
+++ b/doc/man5/flux-config-exec.rst
@@ -38,6 +38,10 @@ sdexec-properties
    (optional) A table of systemd properties to set for all jobs.  All values
    must be strings.  See SDEXEC PROPERTIES below.
 
+testexec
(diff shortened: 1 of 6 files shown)


garlick (approved):
This seems like right approach to me.
(And LGTM)

grondo:
Thanks, I'll propose a PR for flux-accounting.

grondo:
Removed WIP and flux-accounting checks are now passing so I'll set MWP.

PR #5965, opened 2024-05-13
libsubprocess: improve read API

chu11:
Problem: The flux_subprocess_read() family of functions return a pointer to a buffer on success and length of buffer in an
optional parameter.

This API style isn't the most intuitive as it is quite different than the read(2) system call.  Most notably when EOF
is reached on a read stream you have to check if the buffer pointer is non-NULL and the length returned from an optional parameter is 0.

Solution:  Update the read family of functions to return the length from the function and the buffer pointer in an optional parameter.

Update all callers accordingly.

---

side notes

notes, b/c so much code already used `int`s and `char *`s for these functions, I elected to keep `int` vs `ssize_t` and `char *` vs `void *`.

Relatively simple refactoring ... sorry but the review will be a little tedious, lots of adjustments everywhere  :-)

### DIFF
--- a/src/broker/runat.c
+++ b/src/broker/runat.c
@@ -244,7 +244,7 @@ static void stdio_cb (flux_subprocess_t *p, const char *stream)
     const char *line;
     int len;
 
-    if ((line = flux_subprocess_getline (p, stream, &len)) && len > 0) {
+    if ((len = flux_subprocess_getline (p, stream, &line)) > 0) {
         if (streq (stream, "stderr"))
             flux_log (r->h, LOG_ERR, "%s.%d: %s", entry->name, index, line);
         else
--- a/src/cmd/flux-exec.c
+++ b/src/cmd/flux-exec.c
@@ -208,7 +208,7 @@ void output_cb (flux_subprocess_t *p, const char *stream)
     const char *buf;
     int len;
 
-    if (!(buf = flux_subprocess_getline (p, stream, &len)))
+    if ((len = flux_subprocess_getline (p, stream, &buf)) < 0)
         log_err_exit ("flux_subprocess_getline");
 
     if (len) {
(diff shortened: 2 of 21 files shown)


garlick:
We probably shouldn't mix the internal refactoring, external API change, and variable renames in one big commit.

Also, are we sure it is necessary to change the signature of every read function?  I was only advocating for changing `flux_subprocess_read()` and especially in light of the fact that the returned buffer might not be NULL terminated in the future (the unbuffered case).  The others (`_read_line()`, `_read_trimmed_line()`, `_getline()` seem to me to be more like `fgets()` and returning a NULL terminated string result seems natural.

chu11:
> We probably shouldn't mix the internal refactoring, external API change, and variable renames in one big commit.

I could splice out the variable renames, but the internal refactoring sort of comes with the change because ...

> Also, are we sure it is necessary to change the signature of every read function?

I did this because A) the internal code of all the functions shares the same internals B) as I was working on the UNBUF_LOCAL change, I realized it's awkward to return a pointer on success when there is no line to be read. You're just trusting the user to not read anything from the pointer.  As an example, I really disliked a lot of these calls

```
if (!(ptr = flux_subprocess_read_line (..., &len)))
    error
if (ptr && len == 0)
   ...
```

vs now it's just

```
if ((len = flux_subprocess_read_line (..., &ptr)) < 0)
    error
if (len == 0)
   ...
```

which i like better.

:shrug:

garlick:
Well, you make a good point!

src/broker/runat.c
```diff
@@ -244,7 +244,7 @@ static void stdio_cb (flux_subprocess_t *p, const char *stream)
     const char *line;
```
garlick, line 244: commit message and PR title: I'd probably call this "improve read API" since it's really about the API change not refactoring.

garlick (approved):
Thanks!  I'm not spotting anything wrong.
Let's get an ACK from @grondo on the API change before merging though.

src/broker/runat.c
chu11: good point, re-pushed with that tweaked commit message

chu11:
thanks, rebased & re-pushed after fixing that commit message nit.  will set MWP

Issue #5964, opened 2024-05-13
flux queue idle hangs with no active jobs


garlick:
Problem: when flux (0.62.0) was shut down on fluke, the `flux queue idle` command executed during cleanup hung.

A manual run of `flux queue idle` also hangs.

`flux module stats job-manager` shows zero active jobs
```diff
# flux module stats job-manager
{
 "journal": {
  "listeners": 1
 },
 "active_jobs": 0,
 "inactive_jobs": 52,
 "max_jobid": 891940360861779968
}
```
But the debugger shows  `ctx->alloc.alloc_pending_count = 1`.  This value must be zero in order for the queue to be called idle.
```
(gdb) thread find job-manager  # thanks tom
Thread 15 has target name 'job-manager'
(gdb) t 15
[Switching to thread 15 (Thread 0x7fffb2098700 (LWP 2603579))]
#0  0x00007ffff62cf247 in epoll_wait () from /lib64/libc.so.6
(gdb) frame
#0  0x00007ffff62cf247 in epoll_wait () from /lib64/libc.so.6
(gdb) bt
#0  0x00007ffff62cf247 in epoll_wait () from /lib64/libc.so.6
#1  0x00007ffff7b961ff in epoll_poll (loop=0x7fffa8002440, 
    timeout=<optimized out>) at ev_epoll.c:155
#2  0x00007ffff7b98dd6 in ev_run (flags=0, loop=0x7fffa8002440) at ev.c:4157
#3  ev_run (loop=0x7fffa8002440, flags=0) at ev.c:4021
#4  0x00007ffff7b6b7ff in flux_reactor_run (r=r@entry=0x7fffa8002420, 
    flags=flags@entry=0) at reactor.c:124
#5  0x00007fffc53ca649 in mod_main (h=0x7fffa8000be0, argc=<optimized out>, 
    argv=<optimized out>) at job-manager.c:239
#6  0x0000000000411ecf in module_thread (arg=0x7419f0) at module.c:221
#7  0x00007ffff79301ca in start_thread () from /lib64/libpthread.so.0
#8  0x00007ffff61d9e73 in clone () from /lib64/libc.so.6
(gdb) frame 5
#5  0x00007fffc53ca649 in mod_main (h=0x7fffa8000be0, argc=<optimized out>, 
    argv=<optimized out>) at job-manager.c:239
(gdb) p ctx
$5 = {h = 0x7fffa8000be0, handlers = 0x7fffa80109d0, 
[...]

PR #5969, opened 2024-05-14
job-manager: misc cleanup in scheduler interface

garlick:
While trying to find the cause of issue #5964, I finally got fed up and did some cleanup in alloc.c.

Nothing too risky here.

I didn't get to the bottom of the actual issue but figured this cleanup could stand alone.


grondo (approved):
LGTM!

PR #5974, opened 2024-05-16
libsubprocess: minor cleanup

chu11:
Splicing out this collection of minor things from my bigger PR

PR #5975, opened 2024-05-16
libsubprocess: support FLUX_SUBPROCESS_FLAGS_LOCAL_UNBUF

chu11:
Per discussion in #5919, some notes

- built on top of my cleanups in #5974
- should the use of this flag in  job-exec be a different PR?
- in job-exec subprocess flags are passed in via the API to bulk-exec, so it felt wrong to just assume that flag is always set.  So that led to adding a new `flux_subprocess_get_flags()` function and calling an appropriate "read" function depending on flags.  This could be done differently of course, perhaps the flags aren't passed in from the API and are internalized within `bulk-exec`?
- haven't tested to ensure that the performance profile changes as expected, dunno if we have a simple reproducer laying around? Edit: I did now, see notes in #5919

### DIFF
--- a/src/common/libsubprocess/remote.c
+++ b/src/common/libsubprocess/remote.c
@@ -420,35 +420,37 @@ static int remote_channel_setup (flux_subprocess_t *p,
         if (wflag)
             c->line_buffered = true;
 
-        if (!(c->read_buffer = fbuf_create (buffer_size))) {
-            llog_debug (p, "fbuf_create: %s", strerror (errno));
-            goto error;
+        if (!(p->flags & FLUX_SUBPROCESS_FLAGS_LOCAL_UNBUF)) {
+            if (!(c->read_buffer = fbuf_create (buffer_size))) {
+                llog_debug (p, "fbuf_create: %s", strerror (errno));
+                goto error;
+            }
+
+            if (!(c->out_prep_w = flux_prepare_watcher_create (p->reactor,
+                                                               remote_out_prep_cb,
+                                                               c))) {
+                llog_debug (p, "flux_prepare_watcher_create: %s", strerror (errno));
+                goto error;
+            }
+
+            if (!(c->out_idle_w = flux_idle_watcher_create (p->reactor,
+                                                            NULL,
+                                                            c))) {
+                llog_debug (p, "flux_idle_watcher_create: %s", strerror (errno));
+                goto error;
+            }
+
+            if (!(c->out_check_w = flux_check_watcher_create (p->reactor,
+                                                              remote_out_check_cb,
+                                                              c))) {
+                llog_debug (p, "flux_check_watcher_create: %s", strerror (errno));
+                goto error;
+            }
+
+            /* don't start these watchers until we've reached the running
+             * state */
         }
         p->channels_eof_expected++;
-
-        if (!(c->out_prep_w = flux_prepare_watcher_create (p->reactor,
-                                                           remote_out_prep_cb,
-                                                           c))) {
-            llog_debug (p, "flux_prepare_watcher_create: %s", strerror (errno));
-            goto error;
-        }
-
-        if (!(c->out_idle_w = flux_idle_watcher_create (p->reactor,
-                                                        NULL,
-                                                        c))) {
-            llog_debug (p, "flux_idle_watcher_create: %s", strerror (errno));
-            goto error;
-        }
-
-        if (!(c->out_check_w = flux_check_watcher_create (p->reactor,
-                                                          remote_out_check_cb,
-                                                          c))) {
-            llog_debug (p, "flux_check_watcher_create: %s", strerror (errno));
-            goto error;
-        }
-
-        /* don't start these watchers until we've reached the running
-         * state */
     }
 
     if (zhash_insert (p->channels, name, c) < 0) {
@@ -543,11 +545,53 @@ int subprocess_remote_setup (flux_subprocess_t *p, const char *service_name)
     return 0;
 }
 
-static int remote_output (flux_subprocess_t *p,
-                          const char *stream,
-                          const char *data,
-                          int len,
-                          bool eof)
+static int remote_output_local_unbuf (flux_subprocess_t *p,
+                                      const char *stream,
+                                      const char *data,
+                                      int len,
+                                      bool eof)
+{
+    struct subprocess_channel *c;
+
+    if (!(c = zhash_lookup (p->channels, stream))) {
+        llog_debug (p,
+                    "Error returning %d bytes received from remote"
+                    " subprocess pid %d %s: unknown channel name",
+                    len,
+                    (int)flux_subprocess_pid (p),
+                    stream);
+        errno = EPROTO;
+        set_failed (p, "error returning unknown channel %s", stream);
+        return -1;
+    }
+
+    if (data && len) {
+        c->unbuf_data = data;
+        c->unbuf_len = len;
+        if (eof)
+            c->read_eof_received = true;
+        c->output_cb (c->p, c->name);
+    }
+    /* N.B. any data not consumed by the user is lost, so if eof is
+     * seen, we send it immediately */
+    if (eof) {
+        c->read_eof_received = true;
+        c->unbuf_data = NULL;
+        c->unbuf_len = 0;
+
+        c->output_cb (c->p, c->name);
+
+        c->eof_sent_to_caller = true;
+        c->p->channels_eof_sent++;
+    }
+    return 0;
+}
+
+static int remote_output_buffered (flux_subprocess_t *p,
+                                   const char *stream,
+                                   const char *data,
+                                   int len,
+                                   bool eof)
 {
     struct subprocess_channel *c;
 
@@ -637,8 +681,14 @@ static void rexec_continuation (flux_future_t *f, void *arg)
         process_new_state (p, FLUX_SUBPROCESS_EXITED);
     }
     else if (subprocess_rexec_is_output (f, &stream, &data, &len, &eof)) {
-        if (remote_output (p, stream, data, len, eof) < 0)
-            goto error;
+        if (p->flags & FLUX_SUBPROCESS_FLAGS_LOCAL_UNBUF) {
+            if (remote_output_local_unbuf (p, stream, data, len, eof) < 0)
+                goto error;
+        }
+        else {
+            if (remote_output_buffered (p, stream, data, len, eof) < 0)
+                goto error;
+        }
     }
     flux_future_reset (f);
     return;
--- a/src/common/libsubprocess/subprocess.c
+++ b/src/common/libsubprocess/subprocess.c
@@ -534,6 +534,7 @@ flux_subprocess_t *flux_rexec_ex (flux_t *h,
 {
     flux_subprocess_t *p = NULL;
     flux_reactor_t *r;
+    int valid_flags = FLUX_SUBPROCESS_FLAGS_LOCAL_UNBUF;
 
     if (!h
         || (rank < 0
@@ -545,8 +546,7 @@ flux_subprocess_t *flux_rexec_ex (flux_t *h,
         return NULL;
     }
 
-    /* no flags supported yet */
-    if (flags) {
+    if (flags & ~valid_flags) {
         errno = EINVAL;
         return NULL;
     }
@@ -794,6 +794,12 @@ static int subprocess_read (flux_subprocess_t *p,
         return -1;
     }
 
+    if (p->flags & FLUX_SUBPROCESS_FLAGS_LOCAL_UNBUF) {
+        if (bufp && c->unbuf_len)
+            (*bufp) = c->unbuf_data;
+        return c->unbuf_len;
+    }
+
     if (p->local) {
         if (!(fb = fbuf_read_watcher_get_buffer (c->buffer_read_w)))
             return -1;
@@ -837,6 +843,10 @@ int flux_subprocess_read_line (flux_subprocess_t *p,
                                const char *stream,
                                const char **bufp)
 {
+    if (p && (p->flags & FLUX_SUBPROCESS_FLAGS_LOCAL_UNBUF)) {
+        errno = EPERM;
+        return -1;
+    }
     return subprocess_read (p, stream, bufp, true, false, false, NULL);
 }
 
@@ -844,6 +854,10 @@ int flux_subprocess_read_trimmed_line (flux_subprocess_t *p,
                                        const char *stream,
                                        const char **bufp)
 {
+    if (p && (p->flags & FLUX_SUBPROCESS_FLAGS_LOCAL_UNBUF)) {
+        errno = EPERM;
+        return -1;
+    }
     return subprocess_read (p, stream, bufp, true, true, false, NULL);
 }
 
@@ -860,6 +874,9 @@ bool flux_subprocess_read_stream_closed (flux_subprocess_t *p,
         || !(c->flags & CHANNEL_READ))
         return false;
 
+    if (p->flags & FLUX_SUBPROCESS_FLAGS_LOCAL_UNBUF)
+        return c->read_eof_received;
+
     if (p->local)
         fb = fbuf_read_watcher_get_buffer (c->buffer_read_w);
     else
@@ -874,6 +891,11 @@ int flux_subprocess_getline (flux_subprocess_t *p,
 {
     int len;
     bool readonly;
+
+    if (p && (p->flags & FLUX_SUBPROCESS_FLAGS_LOCAL_UNBUF)) {
+        errno = EPERM;
+        return -1;
+    }
 
     len = subprocess_read (p, stream, bufp, true, false, true, &readonly);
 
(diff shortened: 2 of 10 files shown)


src/modules/job-exec/bulk-exec.c
```diff
@@ -222,11 +222,20 @@ static void exec_output_cb (flux_subprocess_t *p, const char *stream)
...
+    int flags = flux_subprocess_get_flags (p);
     int len;
 
-    if ((len = flux_subprocess_getline (p, stream, &s)) < 0) {
-        flux_log_error (exec->h, "flux_subprocess_getline");
-        return;
+    if (flags & FLUX_SUBPROCESS_FLAGS_LOCAL_UNBUF) {
+        if ((len = flux_subprocess_read_local_unbuf (p, stream, &s)) < 0) {
+            flux_log_error (exec->h, "flux_subprocess_read_local_unbuf");
+            return;
```
garlick, line 231: I think it would be fine to assume the UNBUF flag, e.g. just |= to the flags provided  to` bulk_exec_push_cmd()` and drop the new public function.

garlick (commented):
Just a couple of quick comments on a first pass

src/modules/job-exec/bulk-exec.c
chu11: sounds good

src/common/libsubprocess/subprocess.h
```diff
@@ -265,6 +273,18 @@ int flux_subprocess_getline (flux_subprocess_t *p,
...
+ *  exclusively with FLUX_SUBPROCESS_FLAGS_LOCAL_UNBUF.
+ *
+ *  Unlike flux_subprocess_read() the returned data is not NUL
+ *  terminated.  This can be useful for performance improvements,
+ *  especially when data is written out to storage instead of a user's
+ *  terminal.
+ */
+int flux_subprocess_read_local_unbuf (flux_subprocess_t *p,
+                                      const char *stream,
+                                      const char **bufp);
```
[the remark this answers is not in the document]
garlick: Sure on another vote.

Documentation that `flux_subprocess_read()` does not NULL terminate seems sufficient to me but maybe I like to live dangerously :-)

Alternatively, mabye we could just change `iodecode()` to add the termination?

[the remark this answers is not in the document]
grondo: I agree it is fine to change behavior of the `flux_subprocess_read()` function based on the flag.

I haven't re-familiarized myself with the code enough to comment on @garlick's proposal to change `iodecode()` as an alternative though.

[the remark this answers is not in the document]
garlick: Well if we've got two votes to not worry about changing the behavior then I would say lets not get into changing `iodecode()`, if @chu11 agrees.

chu11:
re-pushed given the comments above, did an additional profiling run and it showed some smalll incremental improvements.

src/common/libsubprocess/subprocess.c
```diff
@@ -794,6 +795,24 @@ static int subprocess_read (flux_subprocess_t *p,
...
+        if (!local_unbuf_allowed) {
+            errno = EPERM;
+            return -1;
+        }
+        /* If user called this function twice, return nothing.  They
+         * will hopefully call flux_subprocess_read_stream_closed() to
+         * verify EOF was not yet reached.
+         */
+        if (c->unbuf_consumed)
+            return 0;
```
garlick, line 808: The flag and the check seem unnecessary.  We can probably just document that reading twice returns the same data and it won't be very surprising given the lack of a buffer.

See doc comment below.

src/common/libsubprocess/subprocess.h
```diff
@@ -61,6 +61,13 @@ enum {
     FLUX_SUBPROCESS_FLAGS_SETPGRP = 2,
     /* use fork(2)/exec(2) even if posix_spawn(3) available */
     FLUX_SUBPROCESS_FLAGS_FORK_EXEC = 4,
+    /* flux_rexec(): In order to improve performance, do not locally
+     * copy and buffer output from the remote subprocess.  Immediately
+     * call output callbacks.  Users can directly retrieve data via
+     * flux_subprocess_read().  Data will not NUL terminated.  Use of
+     * other read functions will result in a EPERM error.
+     */
+    FLUX_SUBPROCESS_FLAGS_LOCAL_UNBUF = 8,
```
garlick, line 70: >   Users can directly retrieve data via `flux_subprocess_read()`.

We probably should specify that `flux_subprocess_read()` must be called exactly once when the output callback is invoked.  If not called, data is lost.  If called more than once, the same data is returned.

> Data will not NUL terminated.

Missing a "be".

> Use of other read functions will result in a EPERM error.

EPERM is kind of strange here.  Would EINVAL be better?

src/common/libsubprocess/subprocess.c
```diff
@@ -829,22 +848,22 @@ int flux_subprocess_read (flux_subprocess_t *p,
                           const char *stream,
                           const char **bufp)
 {
-    return subprocess_read (p, stream, bufp, false, false, false, NULL);
+    return subprocess_read (p, stream, bufp, false, false, false, true, NULL);
```
garlick, line 851: This method of reducing duplicated code with a common function with booleans that alter its behavior is hard to follow.  Instead of adding one more boolean to this function, how about adding a check to `flux_subprocess_read_line()` et al like this:
```c
if (p && (p->flags & FLUX_SUBPROCESS_FLAGS_LOCAL_UNBUF)) {
    errno = EINVAL;
    return -1;
}
```

src/common/libsubprocess/remote.c
```diff
@@ -543,11 +545,63 @@ int subprocess_remote_setup (flux_subprocess_t *p, const char *service_name)
...
+     * seen, we can send it in the local unbuf immediately */
+    if (eof) {
+        /* N.B. we technically should clear/memset c->unbuf_data
+         * before returning EOF, as a user could read from it during
+         * the eof callback.  But for performance reasons, we won't do
+         * that and assume the caller is smart enough not to do that.
+         */
+        c->read_eof_received = true;
+        c->unbuf_len = 0;
+        c->unbuf_eof = true;
```
garlick, line 590: Why do we have both `c->eof_received` and `c->unbuf_eof`?

src/common/libsubprocess/test/remote.c
```diff
@@ -307,6 +307,106 @@ void simple_test (flux_t *h)
...
+
+    if (!streq (stream, "stdout"))
+        BAIL_OUT ("unexpected stream: %s", stream);
+
+    if (ctx->scorecard.stdout_lines == 0) {
+
+        len = flux_subprocess_read_line (p, stream, &line);
+        ok (len < 0
+            && errno == EPERM,
+            "flux_subprocess_read_line fails w/ EPERM w/ LOCAL_UNBUF");
```
garlick, line 325: Tests like this should zero `errno` before testing that the function sets it.

src/modules/job-exec/bulk-exec.c
```diff
@@ -264,7 +264,8 @@ static struct exec_cmd *exec_cmd_create (const struct idset *ranks,
         fprintf (stderr, "exec_cmd_create: flux_cmd_copy failed");
         goto err;
     }
-    c->flags = flags;
+    /* bulk-exec defaults to use unbuffered for performance */
+    c->flags = flags | FLUX_SUBPROCESS_FLAGS_LOCAL_UNBUF;
```
garlick, line 268: I'd just say "bulk-exec always uses unbuffered reads for performance" since calling it a default sort of implies that there could be a way to change it.

garlick (commented):
I made a first pass and had some comments/questions

src/common/libsubprocess/remote.c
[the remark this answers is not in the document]
chu11: ahh that comment was leftover from a prior iteration of this PR.  I should remove it.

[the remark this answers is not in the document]
chu11: > If setting c->unbuf_len to zero why not set c->unbuf_data to NULL?

I guess it wasn't completely necessary, but you are correct, it wouldn't hurt.

> If the remote returns data + EOF, doesn't setting c->unbuf_len to zero cause the data to be lost?

Within this function, all the data is returned in the first block in the following code.  Then the EOF in the following block.  So the second setting of unbuf_len = 0 is intentional, to clear out that data and send EOF to the caller.

```
    if (data && len) {                                                                                                                      
        c->unbuf_data = data;                                                                                                               
        c->unbuf_len = len;                                                                                                                 
        c->unbuf_consumed = false;                                                                                                          
        if (eof)                                                                                                                            
            c->read_eof_received = true;                                                                                                    
        c->unbuf_eof = false;                                                                                                               
                                                                                                                                            
        c->output_cb (c->p, c->name);                                                                                                       
                                                                                                                                            
        c->unbuf_consumed = false;                                                                                                          
    }                                                                                                                                       
    /* N.B. any data not consumed by the user is lost, so if eof is                                                                         
     * seen, we can send it in the local unbuf immediately */                                                                               
    if (eof) {                                                                                                                              
        c->read_eof_received = true;                                                                                                        
        c->unbuf_data = NULL;                                                                                                               
        c->unbuf_len = 0;                                                                                                                   
        c->unbuf_eof = true;                                                                                                                
                                                                                                                                            
        c->output_cb (c->p, c->name);                                                                                                       
                                                                                                                                            
        c->eof_sent_to_caller = true;                                                                                                       
        c->p->channels_eof_sent++;                                                                                                          
    }                                
```

src/common/libsubprocess/subprocess.c
chu11: > The flag and the check seem unnecessary. We can probably just document that reading twice returns the same data and it won't be very surprising given the lack of a buffer.

I was trying to be consistent to how the buffered case works.  But yeah, we could tweak that.

src/common/libsubprocess/subprocess.h
chu11: > EPERM is kind of strange here. Would EINVAL be better?

I currently use EPERM for another case where it is "illegal" to call a certain function.  It seems an ok choice.  To me EINVAL just makes you think bad input.  Hmmm, a better option isn't coming to me.  Will sleep on it.

src/common/libsubprocess/remote.c
chu11: I think `unbuf_eof` is left over from a prior attempt.  Should remove it.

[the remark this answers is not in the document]
garlick: Whoops!  My bad, I did not catch the `output_cb()` call in there.

src/common/libsubprocess/subprocess.h
garlick: No big deal I suppose!

chu11:
re-pushed w/ tweaks per comments above.  Only notable thing is kept EPERM vs EINVAL for the "can't use these functions" error case.

garlick:
Sounds good.  I'll get this on my test system that is configured for sdexec and see how it goes.

One concern is the lack of output buffering in `sdexec`, which in combination with this flag means no buffering at all.  I'll try it and we should also audit `job-exec` with that in mind to see if we can spot any problem areas.   We may need to add output buffering to `sdexec` before switching `job-exec` over if there are problems.

garlick:
The shell barrier protocol is fine since I believe it's the only user of stdout, and the protocol consists of the shell sending a line of output and then job exec sending one line back.  There is not much opportunity for buffering (or lack thereof) to screw that up.

@grondo suggested I print something from a userrc to check how `job-exec` handles shell output on stderr.

```lua
-- foo.lua
io.stderr:write("foo bar\n")
io.stderr:write("foo bar\n")
io.stderr:write("foo bar\n")
io.stderr:write("foo bar\n")
```
With master
```
$ flux run -o userrc=foo.lua true
0.040s: flux-shell[0]: stderr: foo bar
0.040s: flux-shell[0]: stderr: foo bar
0.040s: flux-shell[0]: stderr: foo bar
0.040s: flux-shell[0]: stderr: foo bar
$ flux job eventlog -H -p exec $(flux job last) | grep log
[  +0.017012] log component="flux-shell" stream="stderr" rank="0" data="foo bar\n"
[  +0.017024] log component="flux-shell" stream="stderr" rank="0" data="foo bar\n"
[  +0.017031] log component="flux-shell" stream="stderr" rank="0" data="foo bar\n"
[  +0.017038] log component="flux-shell" stream="stderr" rank="0" data="foo bar\n"
```

With this PR there is evidence that lines are not being buffered.
Surprisingly, this happens with and without sdexec.  I thought we had remote line buffering with the normal subprocess server.
```
$ flux run -o userrc=foo true
0.064s: flux-shell[0]: stderr: foo bar
foo bar
foo bar
foo bar
$ flux job eventlog -H -p exec $(flux job last) | grep log
[  +0.033395] log component="flux-shell" stream="stderr" rank="0" data="foo bar\nfoo bar\nfoo bar\nfoo bar\n"

```

garlick:
Oh!   Even though the remote subprocess is line buffered (the default), we are using `flux_subprocess_read()` in the server to consume data from the buffer and put it into messages.  We need to use the line oriented read functions so that we are sending only one line per message or else the benefits of line buffering are lost.  I guess the server output callback needs to check the stream's LINE_BUFFER option setting and then use the appropriate read function.

garlick:
This seems to fix that problem.
```diff
diff --git a/src/common/libsubprocess/server.c b/src/common/libsubprocess/server.c
index fca01444e..139923f23 100644
--- a/src/common/libsubprocess/server.c
+++ b/src/common/libsubprocess/server.c
@@ -246,7 +246,10 @@ static void proc_output_cb (flux_subprocess_t *p, const char *stream)
     const char *buf;
     int len;
 
-    if ((len = flux_subprocess_read (p, stream, &buf)) < 0) {
+    len = flux_subprocess_getline (p, stream, &buf);
+    if (len < 0 && errno == EPERM) // not line buffered
+        len = flux_subprocess_read (p, stream, &buf);
+    if (len < 0) {
         llog_error (s,
                     "error reading from subprocess stream %s: %s",
                     stream,
```

chu11:
> This seems to fix that problem.

Hmmmm.  There's a part of me wondering if I originally messed up and had intended to NOT line buffer on the server side, i.e. all line buffer handling occurs on the client side.

But now that we have LOCAL_UNBUF, perhaps that no longer matters and the approach you have is what needs to be done.

garlick:
I guess it's a trade off.  The client has turned out to be pretty heavy with the buffering on that end.  OTOH if we buffer on the server end then we can end up with more message framing overhead when there are bursts of messages.  But since it's a common idiom to have many clients in one place spread out across many servers, moving the buffering to the server side probably makes sense.

Should we add the server side change in this PR and also maybe a test to show that we get one callback per line in UNBUF mode with the libsubprocess server?  Then we could follow up with a PR for sdexec to add some buffering to it.  I can look at that.

chu11:
re-pushed, adding commit to update server side to send each line buffered chunk and added a multiline test.

garlick (approved):
Seems like we're good!

chu11:
hmmm alpine builder failed

```
Error: not ok 13 - job-exec: can specify default-shell on cmdline
Error: not ok 16 - job-exec: update default shell via config reload
Error: not ok 17 - job-exec: cmdline default shell takes priority
Error: not ok 18 - job-exec: can specify imp path on cmdline
Error: not ok 21 - job-exec: update imp path via config reload
Error: not ok 22 - job-exec: cmdline imp path takes priority
Error: not ok 23 - job-exec: can specify exec service on cmdline
Error: not ok 25 - job-exec: update exec service via config reload
Error: not ok 26 - job-exec: cmdline exec service takes priority
Error: not ok 28 - job-exec: update exec service override via config reload
Error: not ok 32 - job-exec: update sdexec properties via config reload
Error: ERROR: t2403-job-exec-conf.t - exited with status 1
```

these are all job-exec reload module tests.  Wondering if the recent additions kill-timeout, etc. have issues when the module is reloaded.  on alpine, the globals may not be reset.

chu11:
without any better ideas, re-pushed with an extra commit that re-initializes the new signaling globals.  I'm guessing my changes to job-exec tickled things enough to make things fail.

grondo:
Ah, yeah, that makes sense. Probably we were only lucky to get away with it previously.

chu11:
oh whoops, i should have removed `merge-when-passing` until that new commit was reviewed .... hopefully it's ok :-)  here it is

https://github.com/flux-framework/flux-core/pull/5975/commits/a35f615a06693c23a838adf804e93f993e32faa1

grondo:
Well that change overwrites the previous setting on any configuration reload with the defaults. However, probably not a big deal since these are mostly used only for testing at this point...

chu11:
> Well that change overwrites the previous setting on any configuration reload with the defaults. However, probably not a big deal since these are mostly used only for testing at this point...

Gah ... that's right, should have passed the saved argc/argv into it like we do with the exec_config ... I'll do a fix.

```
    if (job_exec_set_config_globals (h, conf, 0, NULL, &err) < 0)
        goto error;

    while ((impl = implementations[i]) && impl->name) {
        if (impl->config) {
            if ((*impl->config) (h, conf, ctx->argc, ctx->argv, &err) < 0)
                goto error;
        }
        i++;
    }
```

PR #5980, opened 2024-05-17
don't free alloc-bypass resources when scheduler reloads

### LINKED ISSUE #5797: jobs with resources assigned by alloc-bypass are included in scheduler hello response
Problem: if a job  is holding resources that were not allocated by the scheduler (e.g. set by `alloc-bypass`), and the instance is restarted, the job is still included in the `hello` protocol where the scheduler is informed of jobs that are still holding resources.

This may lead to problems if the resources were allocated by the scheduler to another job that is also holding them.

Maybe we can add a flag to the alloc event so that when the KVS eventlog is replayed for an alloc-bypass job, the `job->alloc_bypass` flag is set, and then we can skip jobs with that flag in the hello protocol.

This is pretty unlikely I  guess - I just happened to notice it while working in that area of the code.

garlick:
This fixes #5797 by omitting alloc-bypass jobs from the "hello" phase of the scheduler handshake.

It looks like `set-flags` is already used to ensure `job->alloc_bypass` survives a flux restsart.

I did notice that my flux restart test hangs without FLUX_DISABLE_JOB_CLEANUP set, as the alloc-bypass job gets stuck in CLEANUP, so that needs following up.   Will open a separate issue.

### DIFF
--- a/src/modules/job-manager/alloc.c
+++ b/src/modules/job-manager/alloc.c
@@ -331,7 +331,7 @@ static void hello_cb (flux_t *h,
     flux_log (h, LOG_DEBUG, "scheduler: hello");
     job = zhashx_first (ctx->active_jobs);
     while (job) {
-        if (job->has_resources) {
+        if (job->has_resources && !job->alloc_bypass) {
             if (flux_respond_pack (h,
                                    msg,
                                    "{s:I s:I s:I s:f}",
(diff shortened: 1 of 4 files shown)


garlick:
I added a fix for #5981

grondo (approved):
LGTM! Thanks!

garlick:
Thanks!  I'll set MWP.

PR #5982, opened 2024-05-17
plugin: frob for system attribute dependency name

vsoch:
:construction:  NOT FOR REVIEW! :construction: 

This PR is only opened to share the diff (and discuss possible design for an actual solution). We need this functionality for experiments, and this was the shortest path to get there!

Problem: we need a quick solution for assigning a jobid dependnecy to run after success (afterok) given a given name.
Solution: write a frobnicator plugin that handles this transformation to the jobspec.

### DIFF
--- a/src/bindings/python/flux/job/frobnicator/plugins/dependency.py
+++ b/src/bindings/python/flux/job/frobnicator/plugins/dependency.py
@@ -1,0 +1,65 @@
+##############################################################
+# Copyright 2022 Lawrence Livermore National Security, LLC
+# (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+#
+# This file is part of the Flux resource manager framework.
+# For details, see https://github.com/flux-framework.
+#
+# SPDX-License-Identifier: LGPL-3.0
+##############################################################
+
+"""Translate dependency.name into a job id for dependency.afterok
+
+"""
+
+from flux.job.frobnicator import FrobnicatorPlugin
+
+
+class DependencyAdder:
+    """Get the system attribute for dependency.name and add a
+    task dependency. Raise an error if we can't find it.
+    """
+
+    def __init__(self, config={}):
+        # We don't need this for anything, just saving
+        # for the heck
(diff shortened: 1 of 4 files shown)


src/bindings/python/flux/job/frobnicator/plugins/dependency.py
```diff
@@ -0,0 +1,65 @@
...
+        dependency_name = jobspec.attributes["system"].get("dependency", {}).get("name")
+        if not dependency_name:
+            return
+
+        # We are going to add the first matched job as a dependency
+        import flux
+        import flux.job
+
+        handle = flux.Flux()
+        for job in flux.job.list.job_list(handle).get()["jobs"]:
```
grondo, line 46: `flux.job.list.job_list()` has an optional `name` parameter you could use to get only jobs that match `dependency_name`. Also, if you just want the most recent one, I think you could also add `max_entries=1`.

grondo: And I'm sorry, I see you put *NOT FOR REVIEW* in the PR :facepalm:

grondo: Oh, since I already commented :slightly_smiling_face: 

You'll have to test, but `flux.job.list.job_list()` will by default only fetch jobs for the current user. In the frobnicator, this user is the owner of the instance, not necessarily the submitter of the job. You may need to set the `userid` parameter to the submission userid.

vsoch: This is super helpful! I really appreciate the feedback, regardless of my message.

vsoch: Quick question @grondo - if I added those things, would the query be efficient enough to be suitable for an actual solution (and not a hacky one I'm working on)? And more specifically, considered for adding as an optional module (not loaded by default) to the core code base?

PR #5988, opened 2024-05-17
doc: add PRIORITY state to flux-jobs(1)

### LINKED ISSUE #5985: flux-jobs: state PRIORITY is not documented
```
Jobs may be observed to pass through five job states in Flux: DEPEND,
SCHED, RUN, CLEANUP, and INACTIVE (see Flux RFC 21). Under the
``state_single`` field name, these are abbreviated as D, S, R, C, and I
respectively. For convenience and clarity, the following virtual job
states also exist: "pending", an alias for DEPEND,SCHED; "running", an
alias for RUN,CLEANUP; "active", an alias for "pending,running".
```

Just noticed we never added state PRIORITY / "P"  to the `JOB STATUS` section of the manpage.

chu11:
Problem: The flux-jobs(1) manpage did not get updated when the PRIORITY state was introduced.

Updated documentation to note the PRIORITY job state and the "P" abbreviation for it.

Fixes #5985

----

and one extra cleanup in the area

### DIFF
--- a/doc/man1/flux-jobs.rst
+++ b/doc/man1/flux-jobs.rst
@@ -179,11 +179,11 @@ JOB STATUS
 ==========
 
 Jobs may be observed to pass through five job states in Flux: DEPEND,
-SCHED, RUN, CLEANUP, and INACTIVE (see Flux RFC 21). Under the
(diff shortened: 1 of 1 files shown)


garlick (approved):
LGTM!

PR #5989, opened 2024-05-17
flux-dump: add --ignore-failed-read option

garlick:
This adds an option to allow a KVS dump to continue even when there are keys referencing nonexistent blobrefs.   This would have been handy last night when el cap ran out of disk space and we had to kill the rank 0 broker.  This left some dangling keys which prevented flux from starting.  Then after we got rid of those, we found some more in less critical places that we had to run down by iteratively running `flux dump`  and removing one key until it ran clean.

This option allows all the bad keys to be found in one go.

This PR also adds the option to the dump called from rc3 so that any bad keys that find their way into the KVS are simply not included on the next restart.

### DIFF
--- a/doc/man1/flux-dump.rst
+++ b/doc/man1/flux-dump.rst
@@ -76,6 +76,13 @@ OPTIONS
    store.  This may be slightly faster, depending on how frequently the same
    content blobs are referenced by multiple keys.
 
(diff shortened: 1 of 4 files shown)


grondo (approved):
This looks good to me!

PR #5994, opened 2024-05-21
broker: don't post quorum-full in non-QUORUM state

### LINKED ISSUE #5983: broker[0]: quorum-full: ignored in shutdown
Problem: after reloading the KVS on rank 0 to try to get going again after the backing store file system ran out of space, we got stuck during shutdown  with a continuous stream of these log messages and needed to kill -9 the broker:
```
[  +6.149483] broker[0]: quorum-full: ignored in shutdown
[  +6.149607] broker[0]: quorum-full: ignored in shutdown
[  +6.149726] broker[0]: quorum-full: ignored in shutdown
[  +6.149846] broker[0]: quorum-full: ignored in shutdown
[  +6.149982] broker[0]: quorum-full: ignored in shutdown
[  +6.150099] broker[0]: quorum-full: ignored in shutdown
[  +6.150223] broker[0]: quorum-full: ignored in shutdown
[  +6.150339] broker[0]: quorum-full: ignored in shutdown
[  +6.150466] broker[0]: quorum-full: ignored in shutdown
[  +6.150610] broker[0]: quorum-full: ignored in shutdown
[  +6.150746] broker[0]: quorum-full: ignored in shutdown
[  +6.150861] broker[0]: quorum-full: ignored in shutdown
[  +6.150994] broker[0]: quorum-full: ignored in shutdown
[  +6.151115] broker[0]: quorum-full: ignored in shutdown
```

garlick:
Problem: many "quorum-full: ignored in shutdown" messages were observed during shutdown of Flux on a large system with a flat TBON.

The quorum-full broker event could be posted in any state but RUN. Restrict quorum-full posting to the QUORUM state.

Fixes #5983

### DIFF
--- a/src/broker/state_machine.c
+++ b/src/broker/state_machine.c
@@ -896,7 +896,7 @@ static void broker_online_cb (flux_future_t *f, void *arg)
     }
 
     if (quorum_reached) {
-        if (s->state != STATE_RUN) {
+        if (s->state == STATE_QUORUM) {
             state_machine_post (s, "quorum-full");
             if (s->quorum.warned) {
                 flux_log (s->ctx->h, LOG_ERR, "quorum reached");


src/broker/state_machine.c
```diff
@@ -896,7 +896,7 @@ static void broker_online_cb (flux_future_t *f, void *arg)
     }
 
     if (quorum_reached) {
-        if (s->state != STATE_RUN) {
+        if (s->state == STATE_QUORUM) {
```
grondo, line 899: Commit message suggestion: I had trouble parsing "The quorum-full broker event could be posted in any state but RUN." outside of the Problem statement (i.e. I read it as part of the solution)

Maybe if ", because the quorum-full broker event can be posted in any state but RUN"  was appended to the problem statement it would be more clear? (and the solution would just be "Restrict quorum-full posting to the QUORUM state")

I understand it now, so this should be considered optional.

grondo (approved):
LGTM! Just one minor suggestion for the commit message.

garlick:
Thanks!  I fixed that and rebased on current master.  I'll set MWP.

PR #5998, opened 2024-05-22
shell: split up and improve efficiency of input plugin

grondo:
This PR splits up the shell `input` plugin into 3 separate plugins under `src/shell/input/` in order to improve maintainability and to make upcoming changes much easier to follow. The new code is somewhat a rewrite, though much of the existing code was reused or repurposed in the new plugins.

There should be no change of functionality here, _except_ that the input eventlog watch _per task_ was dropped in favor of a single eventlog watch per job shell. This was done because the approach simplified the kvs eventlog input plugin which had to be mostly rewritten anyway, and avoided extra unnecessary churn since reducing the number of eventlog watches was an end goal here anyway.


src/shell/input/file.c
```diff
@@ -0,0 +1,206 @@
...
+#include "internal.h"
+#include "builtins.h"
+#include "input/util.h"
+
+struct file_input {
+    flux_shell_t *shell;
+    const char *path;
+    int fd;
+    flux_watcher_t *w;
+    char *ranks;
```
garlick, line 44: Seems like ranks could be made a fixed length array and strdup/asprintf could be skipped later on?

src/shell/input/file.c
```diff
@@ -0,0 +1,206 @@
...
+    }
+}
+
+static struct file_input *file_input_create (flux_shell_t *shell,
+                                             const char *path)
+{
+    struct file_input *fp = calloc (1, sizeof (*fp));
+
+    if (!(fp = malloc (sizeof (*fp))))
+        return NULL;
```
garlick, line 71: oops double allocation here

src/shell/input/service.c
```diff
@@ -0,0 +1,198 @@
...
+    if (!streq (type, "service") && !streq (type, "file"))
+        return shell_log_errn (0, "input.stdin.type=%s invalid", type);
+
+    if (!streq (type, "service"))
+        return 0;
+
+    if (!(in = input_service_create (shell)))
+        return -1;
+    if (flux_plugin_aux_set (p, "builtin.input-service", in,
+                            (flux_free_f) input_service_destroy) < 0) {
```
garlick, line 184: Style: long parameter lists should be broken one per line

garlick (commented):
Excellent cleanup!

Just a few comments on a quick first pass

src/shell/input/file.c
grondo: Yeah. However, I was trying to change as little as possible from the original plugin (as much as that was possible anyway). This code goes away in a follow on PR so I'd hate to spend time fixing it then removing it if that's ok.

src/shell/input/file.c
grondo: :facepalm: Thanks for catching that!

src/shell/input/service.c
grondo: Thanks, I tried to fix those as I went along but missed one!

src/shell/input/util.c
```diff
@@ -0,0 +1,143 @@
...
+                             "version", 1,
+                             "encoding",
+                              "stdin", "UTF-8",
+                             "count",
+                              "stdin", 1,
+                             "options");
+    if (!o) {
+        errno = ENOMEM;
+        goto error;
+    }
```
garlick, line 130: `eventlog_entry_pack()` already sets errno

src/shell/input/util.c
```diff
@@ -0,0 +1,143 @@
...
+    if (input_kvs_eventlog_init (shell, o) < 0) {
+        shell_log_errno ("input_service_kvs_init");
+        goto error;
+    }
+    rc = 0;
+ error:
+    json_decref (o);
+    return rc;
```
garlick, line 138: `json_decref()` might need an ERRNO_SAFE_WRAP()` if we care about errno in this function

garlick (approved):
LGTM!  I called out a couple of nits but nothing that should hold this up.

src/shell/input/file.c
```diff
@@ -0,0 +1,206 @@
...
+    }
+}
+
+static struct file_input *file_input_create (flux_shell_t *shell,
+                                             const char *path)
+{
+    struct file_input *fp;
+
+    if (!(fp = malloc (sizeof (*fp))))
+        return NULL;
```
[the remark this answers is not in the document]
grondo: Oh, good catch.

src/shell/input/file.c
```diff
@@ -0,0 +1,206 @@
...
+}
+
+static void file_read_cb (flux_reactor_t *r,
+                          flux_watcher_t *w,
+                          int revents,
+                          void *arg)
+{
+    struct file_input *fp = arg;
+    long ps = sysconf (_SC_PAGESIZE);
+    char buf[ps];
```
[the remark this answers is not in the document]
grondo: This is a good point. However, this is unchanged from the original input plugin code and will go away in a follow on PR.

grondo:
Ok, I fixed issues raised in the comments above (except the one that would have changed unmodified code) and will set MWP. Thanks!

PR #6001, opened 2024-05-22
doc: fix typo in flux-jobs(1)

### LINKED ISSUE #5997: Minor documentation fix for flux-jobs
From the flux-jobs man page.  The status_abbrev should be status_emoji.

This:
```
status_abbrev
              status but an appropriate emoji instead of job state / result
```
Should be:
```
status_emoji
              status but an appropriate emoji instead of job state / result
```

chu11:
Problem: The field name for `status_emoji` is incorrect due to a cut and paste error.

Correct the field name for `status_emoji`.

Fixes #5997


grondo (approved):
Nice! :-)

PR #6002, opened 2024-05-22
libsubprocess: reduce remote input prep/check

chu11:
Problem: Profiling shows that a significant amount of time can be spent in the prep/check of remote subprocess input. This is even in the case when the input buffer is empty.

Enable the remote input prep/check only when the buffer is non-empty.

### DIFF
--- a/src/common/libsubprocess/remote.c
+++ b/src/common/libsubprocess/remote.c
@@ -226,23 +226,10 @@ static int remote_channel_setup (flux_subprocess_t *p,
     struct subprocess_channel *c = NULL;
     char *e = NULL;
     int save_errno;
-    int buffer_size;
 
     if (!(c = channel_create (p, output_cb, name, channel_flags))) {
         llog_debug (p, "channel_create: %s", strerror (errno));
         goto error;
-    }
-
-    if ((buffer_size = cmd_option_bufsize (p, name)) < 0) {
-        llog_debug (p, "cmd_option_bufsize: %s", strerror (errno));
-        goto error;
-    }
-
-    if (channel_flags & CHANNEL_WRITE) {
-        if (!(c->write_buffer = fbuf_create (buffer_size))) {
-            llog_debug (p, "fbuf_create: %s", strerror (errno));
-            goto error;
-        }
     }
 
     if (channel_flags & CHANNEL_READ) {
@@ -257,6 +244,11 @@ static int remote_channel_setup (flux_subprocess_t *p,
             c->line_buffered = true;
 
         if (!(p->flags & FLUX_SUBPROCESS_FLAGS_LOCAL_UNBUF)) {
+            int buffer_size;
+            if ((buffer_size = cmd_option_bufsize (p, name)) < 0) {
+                llog_debug (p, "cmd_option_bufsize: %s", strerror (errno));
+                goto error;
+            }
             if (!(c->read_buffer = fbuf_create (buffer_size))) {
                 llog_debug (p, "fbuf_create: %s", strerror (errno));
                 goto error;
@@ -509,6 +501,9 @@ static int send_channel_data (flux_subprocess_t *p)
                 set_failed (p, "internal close error");
                 return -1;
             }
+            /* Don't need this buffer anymore, reclaim the space */
+            fbuf_destroy (c->write_buffer);
+            c->write_buffer = NULL;
         }
         c = zhash_next (p->channels);
     }
--- a/src/common/libsubprocess/remote.c
+++ b/src/common/libsubprocess/remote.c
@@ -48,27 +48,11 @@ static void set_failed (flux_subprocess_t *p, const char *fmt, ...)
     va_end (ap);
 }
 
-static void start_channel_watchers (flux_subprocess_t *p)
-{
-    struct subprocess_channel *c;
-    c = zhash_first (p->channels);
-    while (c) {
-        flux_watcher_start (c->in_prep_w);
-        flux_watcher_start (c->in_check_w);
-        c = zhash_next (p->channels);
-    }
-}
-
 static void stop_channel_watchers (flux_subprocess_t *p, bool in, bool out)
 {
     struct subprocess_channel *c;
     c = zhash_first (p->channels);
     while (c) {
-        if (in) {
-            flux_watcher_stop (c->in_prep_w);
-            flux_watcher_stop (c->in_idle_w);
-            flux_watcher_stop (c->in_check_w);
-        }
         if (out) {
             flux_watcher_stop (c->out_prep_w);
             flux_watcher_stop (c->out_idle_w);
@@ -157,10 +141,7 @@ static void process_new_state (flux_subprocess_t *p,
 
     p->state = state;
 
-    if (p->state == FLUX_SUBPROCESS_RUNNING) {
-        start_channel_watchers (p);
-    }
-    else if (state == FLUX_SUBPROCESS_EXITED) {
+    if (state == FLUX_SUBPROCESS_EXITED) {
         stop_in_watchers (p);
     }
     else if (state == FLUX_SUBPROCESS_FAILED) {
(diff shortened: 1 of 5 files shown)


chu11:
> OK silly question. Could flux_subprocess_write() just call subprocess_write() (which makes the exec.write RPC) directly when the process is remote?

Ya know what, i think it can.  Lets try that.

chu11:
> Ya know what, i think it can. Lets try that.

Ahhh I remember now.  We allow users to `flux_subprocess_close()` on a remote subprocess channel even if the subprocess isn't running yet.  The prep/check in the background handles this magic.

If we wait to call `flux_subprocess_close()` until the subprocess hits running, I think there would be fallout code adjustments elsewhere in `flux-core`, possibly non-trivial although I didn't look into it too deeply.  Just as one example, the `--noinput` option in `flux-exec` closes stdin just a few lines below the call to `flux_rexec()` (i.e. reactor hasn't been re-entered yet, so the initial rexec RPC hasn't even been sent yet).

Perhaps a "LOCAL_UNBUF" on the input side is something we could investigate as well, but perhaps not something for this round?

garlick:
Ah I didn't think of that.  Hmm, could we just set the `c->closed` flag in the channel, and then upon receiving the pid from the remote server, send eofs for all closed channels?

I can't for the life of me think why the input buffering is ever helpful.

chu11:
> Ah I didn't think of that. Hmm, could we just set the c->closed flag in the channel, and then upon receiving the pid from the remote server, send eofs for all closed channels?

Yeah, that's certainly a possibility.  I hadn't thought about it too much yet mainly because ....

I just realized that next Tuesday is not the release deadline, it's in two Tuesdays.  So I actually have time to experiment with an alternate approach.  I was thinking this would be the approach for the upcoming release.

garlick:
Well, it just seems like, in the input case, the buffer isn't helping at all, and in fact impacts scaling in 1:N configs like in job-exec because you have to allocate all the buffers, watchers, etc.  So if it's not helping, it seems really tempting to just cut it out!

chu11:
after messing around today I think I got it to work.  The only gotcha is that we still need the buffer when the caller writes to the subprocess before the subprocess is running.  But we can completely remove the remote input prep/check.

I still to test against the reproducer to make sure this works, but it's looking pretty good and passing all tests.

chu11:
The reproducer on corona seems to work, and profiling shows exactly what we'd hope for.

```
-   23.38%     0.00%  job-exec         job-exec.so            [.] mod_main                              ▒
     mod_main                                                                                           ▒
     flux_reactor_run                                                                                   ▒
     ev_run                                                                                             ▒
   - ev_run                                                                                             ▒
      - 23.00% ev_invoke_pending                                                                        ▒
         - 16.50% check_cb                                                                              ▒
            - exec_start_cmds (inlined)                                                                 ▒
               - 16.48% exec_start_cmd (inlined)                                                        ▒
                  - 16.42% flux_rexec_ex                                                                ▒
                     - flux_rexec_ex                                                                    ▒
                        - 13.49% remote_exec                                                            ▒
                           - 13.42% subprocess_rexec                                                    ▒
                              - 7.20% flux_rpc_pack                                                     ▒
                                 + flux_rpc_vpack                                                       ▒
                              - 6.18% rexec_ctx_create (inlined)                                        ▒
                                 - 6.15% cmd_tojson                                                     ▒
                                    - 5.87% envz_tojson                                                 ▒
                                       + 2.13% json_object_setn_new_nocheck                             ▒
                                       + 1.05% 0x1555545b44bc                                           ▒
                                       + 1.04% 0x1555545b448e                                           ▒
                                         0.63% json_stringn                                             ▒
                        - 1.68% subprocess_remote_setup                                                 ▒
                           - 1.66% remote_setup_stdio (inlined)                                         ▒
                              - remote_channel_setup                                                    ▒
                                 - 1.34% fbuf_create                                                    ▒
                                    + 0.67% cbuf_create                                                 ▒
                                    + 0.64% __GI___libc_malloc (inlined)                                ▒
                        - 1.14% subprocess_create                                                       ▒
                           + 0.93% __socketpair (inlined)                                               ▒
         - 3.24% rexec_continuation                                                                     ▒
            - 1.53% subprocess_rexec_get                                                                ▒
               - 1.14% flux_rpc_get_unpack                                                              ▒
                  + flux_rpc_get_vunpack (inlined)                                                      ▒
            - 1.38% remote_output_local_unbuf (inlined)                                                 ▒
               - 1.32% exec_output_cb                                                                   ▒
                  - 1.29% output_cb                                                                     ▒
                     - 1.24% output_cb                                                                  ▒
                        + exec_barrier_enter (inlined)                                                  ▒
         - 2.67% handle_cb                                                                              ▒
```

garlick:
Nice!

Hey, could we just send the exec matchtag instead of the pid in write requests and then get rid of the buffer?  That would be a protocol change to RFC 42 (which I would be happy to propose if you think it would work out).

garlick:
Perhaps we can slip the protocol change in before this PR then.

I'll get a PR open on the RFC for a start.

chu11:
> Hey, could we just send the exec matchtag instead of the pid in write requests and then get rid of the buffer? That would be a protocol change to RFC 42 (which I would be happy to propose if you think it would work out).

Began looking through code and began wondering about the fallout of removing the local buffer.

- It doesn't appear that `libsdexec` will buffer anything on writes if the process is not yet running, with the minor exception of whatever the fds of `socketpair()` buffers.  This isn't an issue with `rexec` b/c we know it is running when `flux_local_exec()` is done.

- Although I didn't look into it with great rigor, it's not clear to me if systemd will buffer stdin until the process is started and ready to receive data.

- Would it be simpler to just not create the local buffer unless its necessary?   It's now somewhat obvious to me that the local buffer is not needed in most situations.

garlick:
Well I guess we should check if write/close before start is required in the main sdexec case we care about right now which is job-exec.

We may need to add some buffering to sdexec in any case.  I had forgotten about that.   We won't have line buffering with the UNBUF flag...

garlick:
Wait a sec, for local processes, there is an explicit check in `flux_subprocess_write()` that the subprocess is in RUNNING state or it fails with EPIPE.

https://github.com/flux-framework/flux-core/blob/master/src/common/libsubprocess/subprocess.c#L689

So we can't just send to the remote.

I guess your create on demand idea may be the best option.  If we could delete the buffer after it empties out, even better.

chu11:
> But it also makes me wonder why we allow writing/closing before start for remote processes but not for local ones.

I think the main reason is that `flux_local_exec()` is fork/exec, so the process is running before the function returns, so writing before a local process is running is basically impossible.  Dumb example is:

```
p = flux_local_exec();
flux_subprocess_write (p, "foobar", ...);
flux_reactor_run();
```

vs

```
p = flux_rexec();
flux_subprocess_write (p, "foobar", ...);
flux_reactor_run();
```

in the latter case, the remote subprocess RPC has probably not even been sent when `flux_subprocess_write()` is called, but with the former case, it's just fork/exec and the process is definitely running.  So we'd like the API usage pattern between both to still be possible.

garlick:
```
p = flux_rexec();
flux_subprocess_write (p, "foobar", ...);
flux_reactor_run();
```

> in the latter case, the remote subprocess RPC has probably not even been sent when flux_subprocess_write() is called, but with the former case, it's just fork/exec and the process is definitely running. So we'd like the API usage pattern between both to still be possible.

In that example, the exec request definitely is sent before the write request, and the server is guaranteed to receive those messages in order.  Thus if fork+exec/spawn really are synchronous, the remote pid will always be assigned to the subprocess before the write request is processed.  Unless of course there was a failure but then it doesn't matter.

chu11:
> Thus if fork+exec/spawn really are synchronous, the remote pid will always be assigned to the subprocess before the write request is processed. 

I don't think we can be guaranteed this with `sdexec` though.  After `sdexec_start_transient_unit()`, we've only informed `systemd` to run the command we want, but there's no way to know if it has done it before the write request comes in?

garlick:
> I don't think we can be guaranteed this with sdexec though. After sdexec_start_transient_unit(), we've only informed systemd to run the command we want, but there's no guarantee it has done it before the write request comes in?

Good point.  I guess I was assuming we could fix that.  In fact I should tack something on to #6013.  For input I was just thinking we could enqueue the write request messages until the unit comes online.

chu11:
re-pushed, adding a commit to create the write buffer on demand when it is needed.

src/common/libsubprocess/remote.c
```diff
@@ -640,6 +476,58 @@ static int remote_output_buffered (flux_subprocess_t *p,
...
+        }
+        if (c->closed) {
+            if (subprocess_write (p->h,
+                                  p->service_name,
+                                  p->rank,
+                                  p->pid,
+                                  c->name,
+                                  NULL,
+                                  0,
+                                  true) < 0) {
```
garlick, line 513: Could this be combined with the `subprocess_write()` call above?  E.g.
```c
if (c->closed || fbuf_bytes (c->write_buffer) > 0) {
    ...
}
```

src/common/libsubprocess/subprocess.c
```diff
@@ -754,11 +769,25 @@ int flux_subprocess_close (flux_subprocess_t *p, const char *stream)
         c->closed = true;
     }
     else {
-        /* doesn't matter about state, b/c reactors will send closed.
-         * If those reactors are already turned off, it's b/c
-         * subprocess failed/exited.
+        /* if process isn't running, will be sent after process
+         * converts to running.  Or if it has already exited, it
+         * doesn't matter.
```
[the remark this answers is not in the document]
garlick: This code assumes that there is no buffered data if the state is RUNNING.  Maybe it would be helpful to mention where that buffer is flushed.

src/common/libsubprocess/remote.c
```diff
@@ -673,6 +561,8 @@ static void rexec_continuation (flux_future_t *f, void *arg)
     if (subprocess_rexec_is_started (f, &p->pid)) {
         p->pid_set = true;
         process_new_state (p, FLUX_SUBPROCESS_RUNNING);
+        if (send_channel_data (p) < 0)
+            goto error;
```
garlick, line 560: Should we destroy the fbuf here?

garlick (approved):
This is a nice cleanup!

I had a few comments, though nothing very substantial.

One other quick thought - it might not hurt to add a check that `flux exec -n --service=sdexec` works.  Something like
```diff
diff --git a/t/t2409-sdexec.t b/t/t2409-sdexec.t
index cf7ab989c..491ce6717 100755
--- a/t/t2409-sdexec.t
+++ b/t/t2409-sdexec.t
@@ -55,6 +55,9 @@ test_expect_success 'clear broker logs' '
 test_expect_success 'sdexec true succeeds' '
        $sdexec -r 0 $true
 '
+test_expect_success 'sdexec -n cat succeeds' '
+       run_timeout 30 $sdexec -n -r 0 cat
+'
 test_expect_success 'sdexec false fails with exit code 1' '
        test_expect_code 1 $sdexec -r 0 $false
 '
```

src/common/libsubprocess/remote.c
chu11: ahhh you mean in the error case if it wasn't destroyed in `send_channel_data()`.  Hmmm, we certainly could.  But since we're `goto error`, we're failing the subprocess and going to destroy the subprocess, so shortly.  So not super necessary?

src/common/libsubprocess/remote.c
chu11: ahhh i think you're right.  It would have been two calls if the user did it, but we can combine it at this point in the code.

src/common/libsubprocess/remote.c
garlick: Oh, derp!  I didn't notice that it was destroyed in `send_channel_data()`.  NM!

chu11:
> One other quick thought - it might not hurt to add a check that flux exec -n --service=sdexec works. Something like

hmm it just seems that isn't covered in general.  should add tests for the non-sdexec case too

chu11:
re-pushed with tweaks per comments above.  I also added an extra libsubprocess test that covers only closing stdin early vs write & closing it early.

chu11:
setting MWP, thanks!

Issue #5956, opened 2024-05-10
 broker: subprocess.c:249: subprocess_check_completed: Assertion `p->state == FLUX_SUBPROCESS_EXITED' failed.


grondo:
We just hit this assertion failure on elcap:
```
 broker: subprocess.c:249: subprocess_check_completed: Assertion `p->state == FLUX_SUBPROCESS_EXITED' failed.
```

a core was captured. Here's the backtrace
```gdb
#0  0x00007ffff61eeacf in raise () from /lib64/libc.so.6
#1  0x00007ffff61c1ea5 in abort () from /lib64/libc.so.6
#2  0x00007ffff61c1d79 in __assert_fail_base.cold.0 () from /lib64/libc.so.6
#3  0x00007ffff61e7426 in __assert_fail () from /lib64/libc.so.6
#4  0x00007fff70ec0fdf in subprocess_check_completed (p=<optimized out>) at subprocess.c:249
#5  subprocess_check_completed (p=<optimized out>) at subprocess.c:247
#6  0x00007fff70ec03ff in remote_completion (p=0x7ffcf8dbea80) at remote.c:605
#7  rexec_continuation (f=0x7ffcfad50030, arg=0x7ffcf8dbea80) at remote.c:618
#8  0x00007ffff7b958e3 in ev_invoke_pending (loop=0x7fff14002450) at ev.c:3770
#9  0x00007ffff7b99988 in ev_run (flags=0, loop=0x7fff14002450) at ev.c:4190
#10 ev_run (loop=0x7fff14002450, flags=0) at ev.c:4021
#11 0x00007ffff7b6b7ff in flux_reactor_run (r=0x7fff14002430, flags=flags@entry=0) at reactor.c:124
#12 0x00007fff70eb3214 in mod_main (h=0x7fff14000be0, argc=<optimized out>, argv=0x7fff14003260) at job-exec.c:1565
#13 0x0000000000411ecf in module_thread (arg=0x136dd30) at module.c:221
#14 0x00007ffff79301ca in start_thread () from /lib64/libpthread.so.0
#15 0x00007ffff61d9e73 in clone () from /lib64/libc.so.6
```

and the subprocess structure in question:

```gdb
[...]

PR #6003, opened 2024-05-22
    libsubprocess: handle remote protocol errors

chu11:
Problem: When handling status changes for remote subprocesses in rexec_continuation(), it is assumed that a process has reached the EXITED state when the rpc stream receives ENODATA.

A recent broker crash shows that this may not always be the case when systemd (i.e. sdexec) is used to launch remote subprocesses.  In this case, the stream ended before the initial PID was reported.  It is assumed under extreme error scenarios, that systemd may not necessarily report information in the order one would expect with a simple fork/exec.

Solution: If the status stream ends early, do not assume the remote process has exited.  Instead, handle this as an EPROTO error and fail the subprocess.

Fixes https://github.com/flux-framework/flux-core/issues/5956

### DIFF
--- a/src/common/libsubprocess/remote.c
+++ b/src/common/libsubprocess/remote.c
@@ -596,15 +596,6 @@ static int remote_output (flux_subprocess_t *p,
     return 0;
 }
 
-static void remote_completion (flux_subprocess_t *p)
-{
-    p->remote_completed = true;
-    /* TBON inorder delivery of messages should guarantee we received
-     * FLUX_SUBPROCESS_EXITED before this.
-     */
-    subprocess_check_completed (p);
-}
-
 static void rexec_continuation (flux_future_t *f, void *arg)
 {
     flux_subprocess_t *p = arg;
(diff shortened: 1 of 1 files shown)


garlick:
IMHO there are really three issues
1. The fact that the client asserted when the server violated the protocol
2. The fact that sdexec violated the protocol
3. The fact that RFC 42 doesn't document explicitly any required responses before terminating the exec stream with ENODATA (success).

This fixes (1). Instead of calling out the sdexec behavior explicitly as though we are planning to live with it, it should just reference (2).  And (2) can reference RFC 42.

Also, is restoring the assertion really OK?  Is there clearly now no way to hit it?  (I haven't looked at the code in detail)

chu11:
re-pushed, updating comment and commit message per notes above

src/common/libsubprocess/remote.c
```diff
@@ -615,7 +606,21 @@ static void rexec_continuation (flux_future_t *f, void *arg)
...
         if (errno == ENODATA) {
-            remote_completion (p);
+            p->remote_completed = true;
+            /* Per RFC42, when remote processes are launched, the
+             * process should return finished (i.e. state EXITED)
+             * before returning ENODATA.  It is otherwise considered a
+             * protocol error.
+             *
+             * N.B. There is evidence that the sdexec module has
+             * violated the protocol before #6012.
```
garlick, line 616: I would say the more appropriate ref would be #5956 as that's where the larger issue starts.

src/common/libsubprocess/subprocess.c
```diff
@@ -246,11 +246,7 @@ void subprocess_standard_output (flux_subprocess_t *p, const char *stream)
 
 void subprocess_check_completed (flux_subprocess_t *p)
 {
-    if (p->state != FLUX_SUBPROCESS_EXITED) {
-        log_err ("subprocess_check_completed: unexpected state %s",
-                 flux_subprocess_state_string (p->state));
-        return;
-    }
+    assert (p->state == FLUX_SUBPROCESS_EXITED);
```
garlick, line 249: I think it actually might be wise to not revert this commit.  Based on experience with el cap failures, we don't want to have this assert enabled.  So the question is probably do you want a log message when something weird happens that only happens at scale or do you want nothing at all when @grondo builds the RPM with NDEBUG.

garlick (approved):
LGTM just some minor suggestions.

chu11:
re-pushed per comments above, although I tweaked the commit message header & text b/c "assert" is no longer true after I chose not to revert the previous change.

garlick:
Looks good, thanks!

chu11:
great, setting MWP

Issue #2459, opened 2019-10-15
flux-shell: better stdin flow control


chu11:
As discussed in #2257, there is the potential to overflow the stdin buffer of a subprocess launched via `libsubprocess`.  Potential longer out solutions:

- If we stick with per task `guest.input` watchers, it would be a little tricky, but not too hard to simply have the watcher pause / re-start the `guest.input` watcher if the buffer hits its max.

  - Some check/prep watchers can be used to determine when the buffer has space again.
  - would have to add some "buffer space left" function into libsubprocess
  - would have to add some "log offset" to job-info event watch module.

- If we eventually go with a single guest.input watcher per shell, above becomes trickier, but ultimately is not too different. We could

  - pause / re-start `guest.input` and pause stdin for all processes
  OR
  - create another `guest.input` watcher for any stdin that is slow to process its buffer, at worst, performance could "grow" to be a watcher per task just like the above.

PR #6005, opened 2024-05-23
bypass KVS for job stdin with `--input=FILE`

grondo:
This PR modifies the shell `input.file` plugin to bypass the KVS when an input file is specified, rather than reading the file into the KVS input eventlog, i.e. this makes the behavior match documentation. If input is redirected from a file, an RFC 24 redirect event is posted in place of any data.

This also solves the problem of stdin flow control when file input is used since each task now has its input redirected from the file instead of shuttling through the KVS. (flow control is still an issue with interactive input though, so this doesn't fully solve issue #2459)

It would also be trivial now to support a mustache template for a stdin FILE, e.g. to support input per task or node, etc., however, that work is left for the future.

### DIFF
--- a/src/shell/input/file.c
+++ b/src/shell/input/file.c
@@ -96,6 +96,31 @@ static int file_input_task_exec (flux_plugin_t *p,
     return 0;
 }
 
+/*  Init eventlog and add redirect event(s)
+ */
+static void input_eventlog_put_redirect (struct file_input *fp)
+{
+    json_t *context = NULL;
+
+    if (input_eventlog_init (fp->shell) < 0) {
+        shell_log_errno ("failed to initialize input eventlog");
+        return;
+    }
+    if (!(context = json_pack ("{s:s s:s s:s}",
+                               "stream", "stdin",
+                               "rank", "all",
+                               "path", fp->path))) {
+        shell_log_error ("failed to pack redirect eventlog entry");
+        goto error;
+    }
+    if (input_eventlog_put_event (fp->shell, "redirect", context) < 0) {
+        shell_log_errno ("failed to add redirect event to input eventlog");
+        goto error;
+    }
+error:
+    json_decref (context);
+    return;
+}
 
 static int file_input_init (flux_plugin_t *p,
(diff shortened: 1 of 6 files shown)


src/shell/input/file.c
```diff
@@ -10,8 +10,7 @@
```
garlick, line 10: commit message: s/simple/simply/

src/shell/input/file.c
```diff
@@ -96,6 +96,31 @@ static int file_input_task_exec (flux_plugin_t *p,
...
+        shell_log_errno ("failed to initialize input eventlog");
+        return;
+    }
+    if (!(context = json_pack ("{s:s s:s s:s}",
+                               "stream", "stdin",
+                               "rank", "all",
+                               "path", fp->path))) {
+        shell_log_errno ("failed to pack redirect eventlog entry");
+        goto error;
+    }
```
garlick, line 115: I don't think `json_pack()` sets errno so should this be a `shell_log_error()`?

garlick (approved):
Nice!

Any doc updates needed?  The submission commands say:
```
      --input=FILENAME|RANKS
              Redirect stdin to the specified filename, bypassing the KVS.  As
              a special case for flux run, the argument may specify  an  idset
              of task ranks in to which to direct standard input.
```
and flux-shell(1) says:
```
      input.stdin.type=TYPE
              Set job input for stdin to TYPE. TYPE may be either  service  or
              file.  Users  should  not need to set this option directly as it
              will be  handled  by  options  of  higher  level  commands  like
              flux-submit(1).
```

grondo:
> Any doc updates needed? The submission commands say:

Not that I see, sorry. Did you have something specific in mind?

grondo:
Thanks! I just force-pushed fixes for the two issues called out above.

garlick:
> Not that I see, sorry. Did you have something specific in mind?

Well I guess on second look there isn't anything missing.  Never mind!

grondo:
Thanks, I'll set MWP here then!

PR #6006, opened 2024-05-23
broker: add tunable parameters for extreme fanout configurations

garlick:
This adds two new TOML/broker attrs:

**tbon.child_rcvhwm**

Allows the child receive buffer to be capped at some number of messages per peer.  It is currently unlimited.  This might help keep RSS growth in check on the leader broker of a big system

**tbon.zmq_io_threads**

Lets the number of I/O threads spawned by zeromq be increased from the default of 1.  Only affects the leader node.

I have only tested this on my small cluster with
```toml
[tbon]
zmq_io_threads = 2
child_rcvhwm = 5
```

I confirmed that rank 0 has started two I/O threads while other ranks started one, and that a "flood ping" (+) from rank 1 to rank 0 causes the RSS of rank 1 to grow not 0.

_______
(+) `flux ping -p 8k -i 0 -c 100000 0`

### DIFF
--- a/src/broker/overlay.c
+++ b/src/broker/overlay.c
@@ -168,6 +168,7 @@ struct overlay {
     double torpid_max;
     double tcp_user_timeout;
     double connect_timeout;
+    int child_rcvhwm;
 
     struct parent parent;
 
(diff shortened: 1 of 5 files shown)


src/broker/overlay.c
```diff
@@ -164,6 +164,7 @@ struct overlay {
     char uuid[UUID_STR_LEN];
     int version;
     int zmqdebug;
+    int zmq_io_threads;
```
grondo, line 167: Commit message: s/smq/zmq/

grondo (approved):
LGTM!

I tested this branch on TOSS4 with zeromq-4.3.4-3.el8 and saw similar results as reported in the PR description. (However, I had to run multiple flood pings to generate RSS growth in the peers and still saw some growth in the rank 0 broker RSS, though it was 1/4 the amount observed in the child brokers).

A couple other observations:
 - Don't set `tbon.child_rcvhwm=1` - this prevents any progress and the brokers are unable to wire up.
 - zeromq doesn't appear to have a limit on the number of io threads you can ask for. (I was able to start 1024 of them, though that took awhile)

garlick:
Thanks for that extra testing!    Ah so TOSS 4 is actually running a newer version of libzmq than I tested. I had forgotten that.

Interesting about the deadlock with `tcp.child_rcvhwm = 1`.  Maybe we should set a minimum of 2 since that might surprise people.

I'm less worried about putting a cap on `tbon.zmq_io_threads` - how about you?

grondo:
I wasn't too worried about either setting, but thought I'd mention it since I tested. I'd be fine with a minimum of 2 for the childhwm. I did test with that and it seemed to work!

garlick:
I fixed the typo and added the `tbon.child_rcvhwm` limit.  I'll go ahead and set MWP.  Thanks!

PR #6013, opened 2024-05-28
libsubprocess: use matchtag instead of pid for flux_subprocess_write()

garlick:
This is an attempt to implement the subprocess protocol change proposed in flux-framework/rfc#416, which allows a write request to be sent to a remote subprocess before the pid is received.

Unfortunately what seemed like a straightforward change broke a bunch of tests and I've yet to figure out why!

Wanted to post what I had so far and continue discussing this with @chu11

### DIFF
--- a/src/modules/sdexec/sdexec.c
+++ b/src/modules/sdexec/sdexec.c
@@ -65,6 +65,7 @@ struct sdproc {
     flux_future_t *f_start;
     flux_future_t *f_stop;
     struct unit *unit;
+    struct flux_msglist *write_requests;
     struct channel *in;
     struct channel *out;
     struct channel *err;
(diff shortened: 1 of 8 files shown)


chu11:
In our #6002 discussion, we were considering just not creating the write buffer unless it is needed.  Were you thinking of going forward with this protocol change anyways?

garlick:
See my last couple of comments in #6002.  It _seems_ like this should work without any buffering.  IOW the only thing preventing `flux_subprocess_write()` on a remote subprocess without waiting for the status callback to notify that the process is RUNNING is the lack of a token to correlate the request to the subprocess.  The matchtag should work for this.  I might be missing something important though.

garlick:
I'll be danged if I know why this doesn't work.  I've wasted half a day trying to figure it out and am getting nowhere.  Parking for now.

chu11:
i played around with the branch and equally befuddled, especially from the non-sdexec side of the code.  The changes seem so simple and obvious.  

I tried to see if there was a racy scenario where we were writing earlier than we could before (b/c matchtag is available immediately, whereas pid wasn't), but I don't think that scenario is being violated.

I also rebased your branch on master, in case some of the recent fixes we did were part of the problem.

hmmm

garlick:
I figured it out - we need to match both the matchtag _and the sender_ on the server side.  Otherwise input can go to the wrong subprocess, which causes myriad test suite failures.  Derp! :man_facepalming:

garlick:
This and flux-framework/rfc#416 are probably ready for a review.

src/modules/sdexec/sdexec.c
```diff
@@ -755,6 +772,20 @@ static void write_cb (flux_t *h,
         flux_log (h, LOG_ERR, "sdexec.write: subprocess no longer exists");
         return;
     }
+    /* If the systemd unit has not started yet  enqueue the write request for
+     * later processing in start_continuation().
+     */
+    if (sdexec_channel_get_fd (proc->in) != -1) { // not yet claimed by systemd
```
chu11, line 778: this could probably use a little more comment, it's not super intuitive why you'd check for ~= -1.  Perhaps mention that start_contiunation() closes the fd after acknowledgement unit has started.

chu11 (approved):
LGTM!  just some nits

garlick:
Thanks - I fixed up those comments.

garlick:
setting MWP

Issue #6011, opened 2024-05-28
sdexec:  take extra measures to ensure cleanup


garlick:
Problem: job-exec's mechanism for terminating a job could leave processes running.

When sdexec is configured, a job's processes are contained in a cgroup so we should be able to do a better job of cleaning up.

grondo:
I'm wondering if sdexec will need assistance from the IMP here, since the processes running within the cgroup are owned by other users that the system instance user. Probably needs some experimentation to get the right approach.

garlick:
> I'm wondering if sdexec will need assistance from the IMP here, since the processes running within the cgroup are owned by other users that the system instance user. Probably needs some experimentation to get the right approach.

Yeah, I think that may be right.

One observation is that it appears systemd running as the `flux` user is able to successfully signal the IMP, since presumably it is still running with `flux` as its real user id.   Given that, I wonder if we could arrange for the IMP to forward signals to all members of the control group?  

It sounds like a nice `cgroup.kill` interface was proposed back in 2021 (see https://lwn.net/Articles/855049/).  Not sure how far along that is.  We could press that button from the IMP when it receives SIGTERM perhaps?

More study needed.

garlick:
> I wonder if we could arrange for the IMP to forward signals to all members of the control group?

Digging deeper, the IMP already forwards signals to the shell which forwards them to its children.  The problem is that SIGKILL can't be forwarded.  Confirmed by an experiment:
- run a job consisting of a shell script that traps SIGTERM
- manually run `systemctl --user stop <unit>`
- observe that the shell script logs SIGTERM and keeps running
- observe that after a timeout (2m?) the imp is killed with SIGKILL
- and `flux-shell` and the shell script continue to run

I found that `cgroup.kill` does exist on my test system.  Interestingly, I can press this button manually as the flux user and all processes in the cgroup immediately get SIGKILL.  So we don't even need the services of the imp to use it.

Unfortunately it is not present on our toss 4 systems :-(

garlick:
I guess the IMP could mimic systemd's behavior.

It uses `cgroup.kill` if available, but falls back to iterating over `cgroup.procs`.

https://github.com/systemd/systemd/blob/main/src/basic/cgroup-util.c#L312

grondo:
Yeah, I like that idea.

garlick:
[...]

PR #6014, opened 2024-05-29
job-exec: retry kill of job tasks and job shell during job termination

grondo:
Currently, the job-exec module only attempts to signal job tasks (via `flux_job_kill(3)`) once, waits for a timeout (default 25s), then attempts to signal the job shell itself once (via `flux_subprocess_kill(3)`). On a large system, we've seen jobs stuck permanently in CLEANUP after termination because one or more job shells are still running after the signal is sent to the job shell.

This PR adds retries of both `flux_job_kill(3)` (during the initial timeout) and `flux_subprocess_kill(3)` while any job shells are still running. This should cleanup those cases where the initial attempts may fail due to the possible raciness inherent in these interfaces.

The retry is added to the `flux_job_kill(3)` because we do want to avoid just killing the job shell if possible, since this could leave orphan user processes laying around, lose output, etc. However, the real cleanup here will probably be a solution to #6011 in the system instance case. Once an attempt is made to kill the job shell directly, then the timer to send `flux_job_kill(3)` is stopped. Since the job shell kill timer may go on indefinitely, a backoff is used up to a maximum of 5 minutes.

Unsurprisingly, this doesn't yet have any tests since the situation is difficult to recreate. It would perhaps be possible to add some code to the job shell to simulate unkillable processes or a missed job kill event, or a flag for job-exec to not actually send a signal in some cases, etc. If that is desirable I can work on that next.

### DIFF
--- a/src/modules/job-exec/exec.c
+++ b/src/modules/job-exec/exec.c
@@ -618,7 +618,7 @@ static int exec_config (flux_t *h,
     return config_setup (h, conf, argc, argv, errp);
 }
 
-static json_t * exec_stats (void)
+static json_t *exec_config_stats (void)
 {
     json_t *o = NULL;
     json_t *conf = NULL;
@@ -639,6 +639,22 @@ error:
     ERRNO_SAFE_WRAP (json_decref, o);
     ERRNO_SAFE_WRAP (json_decref, conf);
     return NULL;
+}
+
+static json_t *exec_job_stats (struct jobinfo *job)
+{
+    struct bulk_exec *exec = job->data;
+    return json_pack ("{s:i s:i}",
+                      "total_shells", bulk_exec_total (exec),
+                      "active_shells", bulk_exec_current (exec));
+}
+
+static json_t *exec_stats (struct jobinfo *job)
+{
+    if (job)
+        return exec_job_stats (job);
+    else
+        return exec_config_stats ();
 }
 
 struct exec_implementation bulkexec = {
--- a/src/modules/job-exec/exec.c
+++ b/src/modules/job-exec/exec.c
@@ -618,7 +618,7 @@ static int exec_config (flux_t *h,
     return config_setup (h, conf, argc, argv, errp);
 }
 
-static int exec_stats (json_t **stats)
+static json_t * exec_stats (void)
 {
     json_t *o = NULL;
     json_t *conf = NULL;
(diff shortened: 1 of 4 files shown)


garlick (approved):
This looks good to me visually.

One thought on testing: maybe the signal number(s) could be made configurable in job-exec?  Then signals could be trapped and logged in the shell and/or a shell script that is run as a job.  Might be less intrusive than simulating lost signals in the shell :shrug:

grondo:
That seems like a reasonable idea. Then tests could actually terminate the jobs via `flux job kill -s 9` or similar

grondo:
Well it was a good thing you suggested actually testing that because fetching the next update using `flux_watcher_next_wakeup()` is not working as expected after the _second_ timer wakeup. Instead of digging into this now, I think I'll just reuse the `job->kill_timer` to create the exponential backoff.

grondo:
I've pushed an update here that adds some testing. To facilitate testing, new per-job stats are added to the job-exec stats output. It is also possible this per-job information will be useful occasionally on live systems since it gives the job-exec module's internal view of the job's status, e.g.:
```json
{
 "kill-timeout": 5.0,
 "term-signal": "SIGTERM",
 "kill-signal": "SIGKILL",
 "jobs": {
  "ƒ6Wr64Fq": {
   "implementation": "bulk-exec",
   "ns": "job-209933303808",
   "critical_ranks": "0-3",
   "multiuser": 0,
   "has_namespace": 1,
   "exception_in_progress": 0,
   "started": 1,
   "running": 1,
   "finalizing": 0,
   "kill_timeout": 5.0,
   "kill_count": 0,
   "kill_shell_count": 0,
   "total_shells": 4,
   "active_shells": 4
  }
 },
 "bulk-exec": {
  "config": {
   "default_cwd": "/tmp",
   "default_job_shell": "/home/grondo/git/flux-core.git/src/shell/flux-shell",
   "exec_service": "rexec",
   "exec_service_override": 0
  }
 }
}
```

src/modules/job-exec/exec.c
```diff
@@ -618,7 +618,7 @@ static int exec_config (flux_t *h,
     return config_setup (h, conf, argc, argv, errp);
 }
 
-static json_t * exec_stats (void)
+static json_t *exec_config_stats ()
```
garlick, line 621: parameter list should be void

garlick (approved):
The module status seem like they may come in handy!
LGTM!  One nit.

grondo:
Thanks! Fixed that nit and force-pushed. Will set MWP now.

PR #6017, opened 2024-05-30
job-exec: add `kill-signal` and `term-signal` module parameters

grondo:
This PR was split off #6014 and mainly adds new `kill-signal` and `term-signal` module parameters as for testing as suggested by @garlick.

Since this was a good chunk of work with its own tests, documentation updates, etc, it seemed reasonable to propose as a separate PR.

### DIFF
--- a/src/modules/job-exec/job-exec.c
+++ b/src/modules/job-exec/job-exec.c
@@ -100,6 +100,7 @@
 #include "src/common/libutil/fsd.h"
 #include "src/common/libutil/errno_safe.h"
 #include "src/common/libutil/errprintf.h"
+#include "src/common/libutil/sigutil.h"
 #include "ccan/str/str.h"
 
 #include "job-exec.h"
@@ -107,6 +108,8 @@
 #include "exec_config.h"
 
 static double kill_timeout=5.0;
+static int term_signal = SIGTERM;
+static int kill_signal = SIGKILL;
 
 #define DEBUG_FAIL_EXPIRATION 1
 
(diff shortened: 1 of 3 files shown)


src/modules/job-exec/job-exec.c
```diff
@@ -1381,6 +1394,28 @@ static int job_exec_initialize (flux_t *h, int argc, char **argv)
...
+    if (ksignal) {
+        if ((kill_signal = sigutil_signum (ksignal)) < 0) {
+            flux_log (h, LOG_ERR, "invalid kill-signal: %s", ksignal);
+            errno = EINVAL;
+            return -1;
+        }
+        flux_log (h,
+                  LOG_INFO,
+                  "using kill-signal of %s",
+                  sigutil_signame (kill_signal));
```
garlick, line 1408: Possibly for later, but we did add support for dumping the current config for the exec implementations with `flux module stats` e.g.
```
$ flux module stats job-exec
{
 "bulk-exec": {
  "config": {
   "default_cwd": "/tmp",
   "default_job_shell": "/home/garlick/proj/flux-core/src/shell/flux-shell",
   "exec_service": "rexec",
   "exec_service_override": 0
  }
 }
}
```
the global configs could probably be added at some point and then tests wouldn't have to grep the dmesg buffer.

garlick (approved):
Looks great!   I added a couple of minor comments but if this went in as proposed it would be fine.

src/modules/job-exec/job-exec.c
grondo: That makes sense I wonder why that wasn't originally added. I can go ahead and add that here since already the changes meant only for testing have taken way longer than the original bug fix. :disappointed:

grondo:
Ok, just force-pushed a rework that adds the old and new "global" config values to `flux module stats job-exec` output and uses that for tests.

garlick:
Great!  Already approved, so hit MWP whenever you like.

grondo:
Thanks! Will do.

PR #6020, opened 2024-05-31
job-exec: cmdline options for signals should override config settings

chu11:
Problem: cmdline options to the job-exec module should always override configuration settings, but this is not the case for kill-timeout, term-signal, and kill-signal when the configuration is reloaded.

Ensure that cmdline options for kill-timeout, term-signal, and kill-signal always override configuration settings.

----

This was an oversight from commit in #5975

### DIFF
--- a/src/modules/job-exec/job-exec.c
+++ b/src/modules/job-exec/job-exec.c
@@ -1550,7 +1550,11 @@ static void config_reload_cb (flux_t *h,
         goto error;
     }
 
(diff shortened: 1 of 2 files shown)


grondo (approved):
This LGTM. It might actually be slightly inconvenient to not be able to override values set on the command line with a reconfig, but since the purpose of the kill/term signal configuration is testing-only it will be fine (and I like how simple the solution is)

PR #6022, opened 2024-05-31
sdexec: avoid double EOF notification

garlick:
Problem: sdexec sends two EOF notifications for each channel, which causes libsubprocess to get confused and not signal subprocess completion in UNBUF mode.

The sdexec channel class stops its own watcher after sending EOF, but the sdexec module restarts the watcher during completion to ensure it can send EOF in an error case where it was not started.  This results in two EOF notifications.
    
Add an `eof_delivered` flag to the channel and use it to ensure that only one eof is sent.
    
Also, adjust the output of `t2410-sdexec-memlimit.t` (which was hanging) to give more of an indication of what is going on, since it is pretty terse otherwise.

### DIFF
--- a/src/common/libsdexec/channel.c
+++ b/src/common/libsdexec/channel.c
@@ -32,6 +32,7 @@ struct channel {
     char rankstr[16];
     int fd[2];
     flux_watcher_t *w;
+    bool eof_delivered;
     char *name;
     bool writable;
     channel_output_f output_cb;
@@ -70,6 +71,7 @@ static void channel_output_cb (flux_reactor_t *r,
     if (n <= 0) {
         io = ioencode (ch->name, ch->rankstr, NULL, 0, true);
         flux_watcher_stop (w);
+        ch->eof_delivered = true;
     }
     else
         io = ioencode (ch->name, ch->rankstr, buf, n, false);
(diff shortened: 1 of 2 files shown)


garlick:
Hold off reviewing for a sec - I think I can do this in a way that makes adding buffering later on easier.

garlick:
OK, pushed a slight change.  I have another branch that adds output buffering to sdexec and this way of doing it makes that simpler.

src/common/libsdexec/channel.c
```diff
@@ -32,6 +32,7 @@ struct channel {
     char rankstr[16];
     int fd[2];
     flux_watcher_t *w;
+    bool eof_delivered;
```
chu11, line 35: minor commit message nit, UNBUF -> LOCAL_UNBUF if we search logs later on

t/t2410-sdexec-memlimit.t
```diff
@@ -181,4 +179,6 @@ test_expect_success 'memory.max configuration changed' '
 	test_cmp inf.exp maxupdate.out
 '
 
+flux setattr log-stderr-level 3
+
```
chu11, line 183: just want to double check wasn't left errantly?  Seems unnecessary right before `test_done`?

chu11 (approved):
LGTM, just some nits

t/t2410-sdexec-memlimit.t
garlick: That was on purpose.  Without it, we get a stream of module unload messages that are not helpful.

garlick:
Thanks!  I fixed the commit message and forced a push.  Will set MWP.

PR #6023, opened 2024-05-31
add output buffering to sdexec

garlick:
This adds output buffering to sdexec.

I'll mark WIP temporarily while  I ponder more tests.

### DIFF
--- a/src/common/libsdexec/channel.c
+++ b/src/common/libsdexec/channel.c
@@ -24,7 +24,9 @@
 #include "src/common/libutil/errno_safe.h"
 #include "src/common/libutil/errprintf.h"
 #include "src/common/libutil/fdutils.h"
+#include "src/common/libsubprocess/subprocess_private.h" // for default bufsize
 
+#include "outbuf.h"
 #include "channel.h"
 
 struct channel {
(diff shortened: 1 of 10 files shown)


chu11:
after initial skim, took awhile to grok the buffer + offset + used variables.  Am trying to think of a way to make it grok a little easier.  Dunno if a "class" would help?  Like a `get_data()` function can return data and length, and internally it'll update all things it needs to update and hide that from `channel.c`?

garlick:
That's a pretty good idea.  I'll give that a whirl.

garlick:
@chu11 - per your suggestion, I added an "outbuf" class to libsdexec.  It's a tiny thing and purpose-built for use in the channel module so it really didn't feel right to flesh this out as a general purpose container.  Let me know if it helps make the code more readable.

garlick:
I also switched over flux-exec(1) to LOCAL_UNBUF, which should cause any line buffering issues introduced by this PR to cause stdout tests to fail in `t2409-sdexec.t`.

Dropping the WIP since I think this has pretty good test coverage now.

src/common/libsdexec/channel.c
```diff
@@ -24,22 +24,103 @@
...
+done:
+    json_decref (io);
+    return rc;
+}
+
+static size_t nextline (const char *data, size_t len)
+{
+    for (size_t i = 0; i < len; i++)
+        if (data[i] == '\n')
+            return i + 1;
```
chu11, line 74: braces around inner part of loop?

src/common/libsdexec/test/channel_outbuf.c
```diff
@@ -0,0 +1,217 @@
...
+        diag ("%s output: stream=%s len=%d eof=%s",
+              sdexec_channel_get_name (ch),
+              stream,
+              len,
+              eof ? "true" : "false");
+        line_byte_count += len;
+        int count = 0;
+        for (int i = 0; i < len; i++)
+            if (data[i] == '\n')
+                count++;
```
chu11, line 134: braces around inner part of loop?

src/common/libsdexec/test/channel_outbuf.c
```diff
@@ -0,0 +1,217 @@
...
+        line_count += count;
```
chu11, line 135: shouldn't it be at most one line per callback?  I would think > 1 line should be an error?

garlick: Yes.  Whoops - I messed that up.  Good catch!

garlick: Just pushed fixes for those comments.

> shouldn't it be at most one line per callback? I would think > 1 line should be an error?

Sort of, usually.  There will be one extra callback for eof.  And if the line length exceeds the buffer size, the callback will get buffer sized chunks without newlines until the newline is received.  I noted that in the test.

chu11: ahhh ok, I guess when I look at

```
        for (int i = 0; i < len; i++) {
            if (data[i] == '\n')
                count++;
        }
        line_count += count;
```

it suggests to me that number of lines could be > 1 in a single callback, which I think is impossible?  I would think it should be more like

```
        for (int i = 0; i < len; i++) {
            if (data[i] == '\n')
                count++;
        }
        if (count > 1) {
            /* some error */
        }
```

garlick: The test is structured to keep separate tallies of lines and callbacks, then it checks them against the expected number at the end.  So I think that case is covered - if two lines were returned in one callback, the line count would be correct but the callback count would be wrong.

I guess my thinking was that throwing one error at the end made more sense than calling out the same error possibly multiple times.

garlick:
Thanks! I'll set MWP.

PR #6024, opened 2024-05-31
libflux: add `flux_watcher_is_active(3)`

grondo:
While working on something I noticed that there was no libflux version of the `ev_is_active()` call, which could come in handy to check if a watcher has been started.

This simple PR just adds `flux_watcher_is_active(3)`, tests, and documentation.

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -205,6 +205,7 @@ MAN3_FILES_SECONDARY = \
 	man3/flux_reactor_active_decref.3 \
 	man3/flux_fd_watcher_get_fd.3 \
 	man3/flux_watcher_stop.3 \
+	man3/flux_watcher_is_active.3 \
 	man3/flux_watcher_destroy.3 \
 	man3/flux_watcher_next_wakeup.3 \
 	man3/flux_handle_watcher_get_flux.3 \
--- a/doc/man3/flux_watcher_start.rst
+++ b/doc/man3/flux_watcher_start.rst
@@ -13,6 +13,8 @@ SYNOPSIS
 
   void flux_watcher_stop (flux_watcher_t *w);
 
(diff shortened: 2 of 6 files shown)


chu11:
I tried doing this a long time ago (although tried to do it differently) https://github.com/flux-framework/flux-core/pull/4173

I think you handled the gotchas, but just in case you wanted to review comments there

grondo:
I had forgotten that sorry!

The use case that came up is that I was going to add per-job data to `flux module stats job-exec` and wanted to test if the `kill_timer` or the `kill_shell_timer` watchers were active rather than using a specific flag or other method.

However, if this is still deemed more work than its worth, then I can close this one for now (I don't think it is worth it if we have to add a now method to flux_watcher_t that calls `ev_is_active()`)

garlick:
For this to be safe in-tree, `ev_flux`, `ev_zmq`, `ev_fbuf_read`, `ev_fbuf_write` would also need to be special-cased, otherwise calling `ev_is_active(w->data)` would dereference an implementation specific structure it is a `struct ev_TYPE`

Also, we exported API functions to allow out of tree watchers to be created.  I shortsightedly did not pad `struct flux_watcher_ops` so adding a callback would break ABI.

Eh, not strictly related to this PR, but what if we clawed back the out of tree watcher support until it has stablized and there is a demonstrated need for it?  Then we can have nice things like `flux_watcher_is_active()` without that complication?

grondo:
Sorry had meant to close this. It would be nice to have an issue open on this, though, since there are times when it would be very convenient to tell if a watcher has been started or not (as libev provides with `ev_is_active()`)

garlick:
> It would be nice to have an issue open on this, though, since there are times when it would be very convenient to tell if a watcher has been started or not (as libev provides with ev_is_active())

see #4156

PR #6025, opened 2024-05-31
libsubprocess: protect against double EOF

chu11:
Problem: A recent fix in the sdexec module corrected a double EOF, which caused the client side to hang when the LOCAL_UNBUF flag was used. The client side hang was due to incorrect counting of closed channels.

Solution: Properly flag that an EOF has been sent to the caller when the LOCAL_UNBUF flag is used, so that multiple EOFs are not reported to users.


grondo (approved):
LGTM!

PR #6027, opened 2024-06-03
broker: add timezone abbrev to log message timestamps

### LINKED ISSUE #6015: broker: add timezone designator to log timestamps
A user requested that some kind of timezone designator be added to broker log messages on stderr, since the local time without a timezone can be ambiguous when users on a system may be in several different timezones.

grondo:
Problem: The timezone designation is not included with broker log messages, making determination of when an error message in a logfile occurred slightly ambiguous.

Add the timezone abbreviation to the timestamp.

Fixes #6015


garlick (approved):
LGTM

PR #6028, opened 2024-06-03
resource: stop scheduler updates during shutdown

### LINKED ISSUE #6018: fluxion logs resource status changed for individual nodes
Problem: when shutting down elcap, we've noticed that fluxion logs a debug message about the status changing to down for every node individually.

It appears the `resource.acquire` RPC sends the scheduler an update each time an online/offline/drain/undrain event is posted in the resource eventlog.  In the case of shutdown, the event should be posted by the "monitor" subsystem that watches the `broker.online` group.  The groups subsystem batches contemporaneous membership changes in 100ms blocks.

Flux is down right now on el cap but when it comes up we could examine the resource.eventlog to see if an event was posted for each node.  Possibly we need more aggressive batching or we need to tell the resource module to withhold acquire updates during shutdown.

garlick:
Problem: when flux is shutting down on a big system, the scheduler has to needlessly process DOWN acquire responses until rc3 finally runs on rank 0 and unloads the scheduler module.

Add a `flux resource acquire-mute` command that is run after the queues have been stopped in `cleanup`.  It should be safe to keep the scheduler in the dark at this point as it can still process `free` requests from completing jobs and should have no pending jobs in its queue(s).

Fixes #6018


grondo (approved):
This LGTM!

garlick:
Thanks!  I fixed the python linting error and now I"ll set MWP.

PR #6029, opened 2024-06-03
librlist: ensure `properties` is an array when parsing `resource.config` entries

grondo:
There's an oversight in librlist parsing of `resource.config` entries, and a non-array `properties` value is silently ignored.

Validate that any `properties` entry is a JSON array and generate an error if not.

### DIFF
--- a/t/t0026-flux-R.t
+++ b/t/t0026-flux-R.t
@@ -355,7 +355,19 @@ test_expect_success 'flux R parse-config detects invalid property' '
 	EOF
 	test_must_fail flux R parse-config conf
 '
-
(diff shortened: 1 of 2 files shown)


t/t0026-flux-R.t
```diff
@@ -310,17 +310,6 @@ test_expect_success 'flux R parse-config detects host with no resources' '
 	EOF
 	test_must_fail flux R parse-config conf
 '
```
chu11, line 312: commit message "test ing" -> "test in"

chu11 (approved):
LGTM, just that commit message typo

grondo:
Thanks! Fixed that commit message typo and will set MWP

PR #6030, opened 2024-06-03
speed up flux overlay status on a big system

garlick:
Problem: `flux overlay status` makes some unnecessary RPCs that slow things way down on el cap.  Specifically, the `overlay.health` RPC is made to leaf nodes despite the fact that the parent already knows the status of its children.

Eliminating that gets us about a 4x speedup on el cap.

before
```
[garlick@elcap1:cmd]$ time /usr/bin/flux overlay status >/dev/null

real	1m28.377s
user	0m56.726s
sys	0m0.905s
```
After
```
[garlick@elcap1:cmd]$ time ./flux overlay status >/dev/null

real	0m19.892s
user	0m18.275s
sys	0m0.209s
```

### DIFF
--- a/src/cmd/builtin/overlay.c
+++ b/src/cmd/builtin/overlay.c
@@ -278,7 +278,7 @@ static const char *status_getname (struct status *ctx,
     return buf;
 }
 
-/* If --times, return string containing parenthesised elapsed
+/* If verbose >= 2, return string containing parenthesised elapsed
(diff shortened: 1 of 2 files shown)


grondo (approved):
Nice!! LGTM, but it appears there may be one test that is no longer valid or needs a tweak.

garlick:
Fixed that test and I'll set MWP.  Thanks!

Issue #6021, opened 2024-05-31
user feedback on error messages


garlick:
Cray is reporting some issues with Flux error messages.  Specifically, they tend to interpret some messages as flux failures rather than job failures.

Some comments:

> When flux loses contact with a broker, it generates a message that says I lost contact with a broker.  When slurm loses contact with a slurmd it says: node fail.  Flux is more technically correct... but slurm leads more directly to correct interpretation by the user.

Clarification:  they mean the "hostname (rank N)  has disconnected unexpectedly, marking it LOST" message

> FWIW the "lost contact with job shell" message is the one that seems to trip people up constantly, along with generally things like "job.exception type=exec severity=0 rank 2 exited and exit-timeout=30s has expired flux-job: task(s) Segmentation fault"
>
> People seem to read that second one as "the flux-job task segfaulted" rather than "flux is reporting that the underlying task segfaulted"

PR #6032, opened 2024-06-04
improve broker messages when nodes die

garlick:
Problem: as noted in  #6021, the log messages like this one are confusing to users:
```
broker[0]: picl2 (rank 2) has disconnected unexpectedly. Marking it LOST.
```

Moreover, when the TBON is not flat, only the subtree root is called out.  The other nodes in the subtree are also lost but not called out.

For your consideration: this PR
- demotes the LOST messages to LOG_DEBUG
-  adds a new LOG_ERR message that tracks brokers leaving the `broker.online` group during broker RUN and CLEANUP states. 

Example, if I kill -9 rank 1 of a kary:2 8 node instance, I get the following on stderr:
```
 $ flux start -s8 -o,-Stbon.topo=kary:2
Jun 04 09:21:45.016921 PDT broker.err[0]: dead to Flux: system76-pc,system76-pc,system76-pc,system76-pc (rank 1,3-4,7)
```

In a system instance, if I power off rank 2 which leads the subtree consisting of 2,6-7 I get the following in the logs.
```
[ +21.998384] broker[0]: picl2 (rank 2) has disconnected unexpectedly. Marking it LOST.
[ +22.100710] broker[0]: dead to Flux: picl[2,6-7] (rank 2,6-7)
```

A downside of this approach for the system instance is it's a bit chatty if you shut down a non-leaf node that is not the root.  Here I stop flux on picl1 which causes its TBON children (3-5) to shut down nicely, triggering mutiple log entries:
```
[Jun04 09:36] broker[0]: dead to Flux: picl[3-4] (rank 3-4)
[  +0.127260] broker[0]: dead to Flux: picl5 (rank 5)
[  +0.285764] broker[0]: dead to Flux: picl1 (rank 1)
```
(That won't happen when rank 0 is shut down because that's shutting down the whole instance and we don't log these messages in SHUTDOWN state)

Anyway thought I'd try this and see what it looks like.  Thoughts?

### DIFF
--- a/src/broker/overlay.c
+++ b/src/broker/overlay.c
@@ -759,6 +759,28 @@ static void overlay_child_status_update (struct overlay *ov,
     }
 }
 
+static void log_lost_connection (struct overlay *ov,
+                                 struct child *child,
+                                 const char *why)
+{
+    int children = topology_get_descendant_count_at (ov->topo, child->rank);
+    char add[128] = { 0 };
+
+    if (children > 0) {
+        snprintf (add,
+                  sizeof (add),
+                  " and severed contact with %d other nodes",
+                  children);
+    }
+    flux_log (ov->h,
+              LOG_ERR,
+              "%s (rank %d) %s%s",
+              flux_get_hostbyrank (ov->h, child->rank),
+              (int)child->rank,
+              why,
+              add);
+}
+
 static int overlay_sendmsg_child (struct overlay *ov, const flux_msg_t *msg)
 {
     int rc = -1;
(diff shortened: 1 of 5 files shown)


garlick:
Mark noted offline that the proposed message obscures which node was the root cause.  I agree with this although it's a bit tricky with one log message due to the way the code is currently structured.  Would two log messages be acceptable?  E.g. how's this?
```
Jun 04 12:00:09.522326 PDT broker.err[1]: lost contact with system76-pc (rank 3) and 1 other nodes it was connected to
Jun 04 12:00:09.723159 PDT broker.err[0]: dead to Flux: system76-pc,system76-pc (rank 3,7)
```

grondo:
I think that's probably reasonable. Though I'm sure sysadmins and users would prefer something like "system76-pc (rank 3) failed." to make it explicit that Flux considers that the node has "failed" since they called out "lost contact" as confusing. However, I think this is fine and maybe people will just get it after awhile?

We could also ask some sysadmins for suggestions.

grondo (approved):
This LGTM. I'll check on the new test failures.

garlick:
Thanks! Setting MWP

PR #6033, opened 2024-06-04
sdexec: don't return ENODATA without start/finish

### LINKED ISSUE #6012: sdexec: does not comply with RFC42 protocol
Per discussion in #5956 and per updated RFC42 from https://github.com/flux-framework/rfc/pull/415

The sdexec module appears to not follow required protocols under extreme error scenarios.  In the case of #5956, the remote subprocess was OOMEd and the stream was ended (ENODATA) before the "started" or "finished" response was returned to the client side.

The reason for this is unknown.  It's possible we are not handling a special case when a job is OOMed.  It's also possible that systemd may not behave like fork/exec and could result in unexpected behavior, i.e. it could elect to report information out of order under extreme error scenarios.

garlick:
Problem: in an OOM situation, sdexec was observed to return ENODATA to an exec request without having sent started or finished responses, which violates RFC 42.

Fail with EINVAL and a detailed error response message if this happens.

Fixes #6012.

### DIFF
--- a/src/modules/sdexec/sdexec.c
+++ b/src/modules/sdexec/sdexec.c
@@ -170,6 +170,19 @@ static void finalize_exec_request_if_done (struct sdproc *proc)
                        sdexec_unit_systemd_error (proc->unit));
             exec_respond_error (proc, ENOENT, error.text);
         }
+        else if (!proc->started_response_sent) {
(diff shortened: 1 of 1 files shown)


grondo (approved):
LGTM!

PR #6039, opened 2024-06-12
etc: update bash completions

grondo:
This PR reworks the bash completions to fix a bunch of annoyances I've found over time:
 - Long options get double-completed after they are fully specified, e.g. `--filter=<tab>` -> `--filter=--filter`
 - Default or file/directory completion isn't supported on commands and options that take files/dirs, e.g. `flux exec`, `flux start`, the submission cli commands (like `flux batch`) and options like `--add-file=` `--config-path-` etc.
 - Handling of completions for specific options is awkward and it is difficult to specify completions for the short option, long option with and without `=`, etc.
 - A couple key commands and options were missing

Here completions are reworked to use a scheme that more closely matches the bash-completion project. The `_split_longopts()` function is borrowed from that project and this function is used to split long options at the `=` so completion can be unified between short, long and long with `=` (This is typically done with a `case` statement as you'll see in the code)

Since this new scheme is closer to bash-completion, we could eventually update to requiring that project more easily (it has lots of nice improvements and helper functions)

### DIFF
--- a/etc/completions/flux.pre
+++ b/etc/completions/flux.pre
@@ -292,6 +292,80 @@ _flux_shutdown()
     return 0
 }
 
+#  flux-archive(1) completions
+_flux_archive()
+{
+    local cmd=$1
+    local subcmds="create remove extract list"
+    local split=false
+
+    local create_OPTS="\
+        -h --help \
+        -n --name= \
+        --no-force-primary \
+        -C --directory= \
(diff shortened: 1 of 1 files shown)


garlick (approved):
Cool!

grondo:
Thanks, I've set MWP.

PR #6041, opened 2024-06-14
flux-python: add posix-compatible shebang wrapper

trws:
problem: using `flux python` on minimal systems, say alpine or other minimal systems that do not provide an `env -S`, is currently non-trivial. Since the shebang only supports a single argument after the command, there's no way to search the path for flux to run flux python

solution: add a `flux-python` script, which is installed to the same bin path as `flux` itself, so that `#!/usr/bin/env flux-python` can be used as a shebang

ping @tgamblin

### DIFF
--- a/src/cmd/Makefile.am
+++ b/src/cmd/Makefile.am
@@ -81,6 +81,7 @@ flux_SOURCES = \
 	builtin/shutdown.c
 nodist_flux_SOURCES = \
 	builtin-cmds.c
+dist_bin_SCRIPTS = flux-python
 
 flux_LDADD = \
 	$(top_builddir)/src/common/libpmi/libupmi.la \
--- a/src/cmd/flux-python
+++ b/src/cmd/flux-python
@@ -1,0 +1,8 @@
+#!/bin/sh
+
+# This minimal wrapper exists to help handle systems where env -S is
+# unavailable and scripts wish to use `flux python`. Using a shebang like this
+# for flux python scripts allows them to rely only on posix to start:
+# #!/usr/bin/env flux-python
+
+exec flux python "$@"


grondo:
This seems reasonable. I could have sworn we tried this before and gave up for some reason, but I'm finding no history of it. :shrug:

tgamblin:
Ping @jefflong6 @xorjane

src/cmd/flux-python
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+# This minimal wrapper exists to help handle systems where env -S is
+# unavailable and scripts wish to use `flux python`. Using a shebang like this
+# for flux python scripts allows them to rely only on posix to start:
+# #!/usr/bin/env flux-python
+
+/usr/bin/env flux python "$@"
```
tgamblin, line 8: Probably want to avoid an extra process:

```suggestion
exec /usr/bin/env flux python "$@"
```

tgamblin (changes requested):

src/cmd/flux-python
grondo: while we're at it, is the `/usr/bin/env`  superfluous in a wrapper script?  (`/bin/sh` will already find the first `flux` in `PATH`) I could be missing some subtlety though.

trws: True, it is superfluous.  Fixing.

tgamblin (approved):

src/cmd/Makefile.am
```diff
@@ -81,6 +81,7 @@ flux_SOURCES = \
 	builtin/shutdown.c
 nodist_flux_SOURCES = \
 	builtin-cmds.c
+dist_bin_SCRIPTS = flux-python
```
grondo, line 84: Just making sure we want this in `$prefix/bin`. I can't think of a problem with it, but currently the only thing we install there is the `flux` executable, so just thought I'd double check.

tgamblin: I think you do, so that it is always there when people want to run flux scripts.

grondo: Oh yeah, so they can do `#!/usr/bin/env flux-python`

grondo (approved):
Thanks! This LGTM!
(Note there's an issue with the `el8,system` builder right now that I'm fixing in an upcoming PR. The failures aren't related to this particular PR)

trws:
> Thanks! This LGTM!
> 
> (Note there's an issue with the `el8,system` builder right now that I'm fixing in an upcoming PR. The failures aren't related to this particular PR)

Thanks @grondo. I set MWP but I'm on a phone today, would you merge if it makes sense or let me know to rebase when the other one's in?

PR #6043, opened 2024-06-14
ci: convert `system` test container to run under `podman` for better `systemd` support

grondo:
Running our `system` test container under `docker` is now failing in GitHub actions on the 20.04 actions runner. We already knew that `docker` could no longer run the `system` container (which has to have the ability to start `systemd` as PID 1) on ubuntu 22.04, so it was probably inevitable that at some point this was going to break.

As an experiment, I updated the `system` container to run via `podman` in the `docker-run-systest.sh` script, because `podman` officially supports systemd via a `--systemd` option, and it seems to work. Since ubuntu 22.04 has the correct version of `podman`, this PR adds those changes and updates the github runners for the main workflow to `ubuntu-latest`.

Some other fallout was fixed along the way (notes in specific commit messages and comments.)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -133,7 +133,7 @@ jobs:
 
   ci-checks:
     needs: [generate-matrix]
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     env:
       TAP_DRIVER_QUIET: 1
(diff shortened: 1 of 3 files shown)


garlick (approved):
Thanks for running this down!  All looks fine to me.

grondo:
Thanks! I've set MWP

PR #6045, opened 2024-06-18
libidset: improve efficiency of idset intersection

grondo:
Problem: idset_intersect(3) and idset_has_intersection(3) always iterate one idset fully to determine the result, but the code doesn't check to ensure it is iterating the smaller of the two provided idsets. This could lead to wasted cycles, especially when the idset being iterated has many more entries than the other argument.

If both idsets do not have the IDSET_FLAG_COUNT_LAZY flag set, then swap arguments in both idset_intersect(3) and idset_has_intersection(3) such that the idset with the smaller number of entries is iterated.

Before this change, to intersect the idsets 1-10000 and 1-100 took an average of 4.5ms, with this change, that drops to .02 ms.

### DIFF
--- a/src/common/libidset/idset.c
+++ b/src/common/libidset/idset.c
@@ -452,6 +452,18 @@ bool idset_has_intersection (const struct idset *a, const struct idset *b)
     if (a && b) {
         unsigned int id;
 
+        /*  If there isn't a penalty for idset_count(3), then ensure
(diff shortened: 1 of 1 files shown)


garlick (approved):
Oh nice one!

grondo:
Thanks, I'll set MWP.

PR #6048, opened 2024-06-18
job-list: add support for `ranks` constraint

### LINKED ISSUE #6046: job-list: support `ranks` constraint
per https://github.com/flux-framework/rfc/pull/417#pullrequestreview-2126711356

grondo:
This PR adds a `ranks` constraint operator to job-list to allow restricting output to jobs that were assigned one of a set of ranks. This is useful in testing or to eventually make queries not dependent on hostname.

There's already a pending PR to update RFC 43.

Fixes #6046

### DIFF
--- a/src/modules/job-list/match.c
+++ b/src/modules/job-list/match.c
@@ -514,6 +514,95 @@ static struct list_constraint *create_hostlist_constraint (
     return NULL;
 }
 
+static int match_ranks (struct list_constraint *c,
+                        const struct job *job,
+                        unsigned int *comparisons,
+                        flux_error_t *errp)
+{
+    struct idset *idset = zlistx_first (c->values);
+    size_t n, m;
+
+    /* ranks may not exist if job never ran */
+    if (!job->ranks)
+        return 0;
+    if (!job->ranks_idset) {
+        /* hack to remove const */
+        struct job *jobtmp = (struct job *)job;
+        if (!(jobtmp->ranks_idset = idset_decode (job->ranks)))
+            return 0;
+    }
+    /* Account for all ranks being compared before calling
+     * inc_check_comparison. This is the smallest of the job or
+     * comparison idset
+     */
+    m = idset_count (job->ranks_idset);
+    n = idset_count (idset);
+    *comparisons += (m < n ? m : n) - 1;
+    if (inc_check_comparison (c->mctx, comparisons, errp) < 0)
+        return -1;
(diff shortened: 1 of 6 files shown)


t/t2260-job-list.t
```diff
@@ -723,6 +723,33 @@ test_expect_success 'flux job list of all jobs that ran on node[1-2] after certa
...
+	  | $RPC job-list.list | $jq .jobs | $jq -c '.[]' | $jq .id > constraint_hostlist1.out &&
+	numlines=$(cat completed.ids running.ids failed.ids timeout.ids | wc -l) &&
+	test $(cat constraint_hostlist1.out | wc -l) -eq ${numlines}
+'
+test_expect_success 'flux job list by rank (3)' '
+	constraint="{ and: [ {ranks:[\"3\"]} ] }" &&
+	$jq -j -c -n  \
+	  "{max_entries:1000, attrs:[\"ranks\"], constraint:${constraint}}" \
+	  | $RPC job-list.list | $jq -r .jobs[].ranks | uniq > jobs.ranks &&
+	test $(cat jobs.ranks) -eq 3
```
chu11, line 738: sort stinks we have to hard code it.  could it be count of completed + running jobs divided by 4, or something like that?

chu11 (approved):
generally speaking LGTM, just some small nits

t/t2260-job-list.t
grondo: It may not be clear, but this is testing if all returned ranks were literally `3` (not a count of 3). That is, we queried for rank 3 and got only rank == 3 for all entries.

chu11: yeah, that's obvious now ... i was mentally thinking of all the "get the right number of line" tests we have in the same area

src/modules/job-list/match.c
```diff
@@ -515,6 +515,96 @@ static struct list_constraint *create_hostlist_constraint (
...
+    }
+    if (idset_count (idset) > mctx->max_hostlist) {
+        errprintf (errp, "too many ranks specified");
+        goto error;
+    }
+    if (!zlistx_add_end (c->values, idset)) {
+        errprintf (errp, "failed to append idset structure");
+        idset_destroy (idset);
+        goto error;
+    }
```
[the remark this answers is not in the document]
grondo: Oops. This was cut-and-paste error from the hostlist matching function. I'll add a commit to fix that as well.

PR #6049, opened 2024-06-19
kvs: return ENOSYS on unfinished requests

### LINKED ISSUE #5979: kvs: uncompleted RPCs should receive ENOSYS response when module is unloaded
Problem: we ran `flux module reload kvs` on a live system and it seemed evident that pending commit requests did not receive an error.

I had thought that rpc_track would ensure that pending RPCs get an ENOSYS when the module handling them is unloaded but that may not be the case.   It probably should be so that requests don't hang forever.

chu11:
Problem: When the KVS module is unloaded, any unfinished requests (get, put, wait-version, etc.) should get an ENOSYS response.  That is presently not handled.
    
Solution: Track all requests sent to the kvs module that will not be returned immediately.  Upon module exit, respond to all unfinished requests with ENOSYS.

Fixes #5979

----

note, I noted in #5979 that some other modules might need this fix as well.  Could tack onto this PR as well, but thought we'd deal with just this one first since it specifically has been hit.

### DIFF
--- a/src/modules/kvs/kvs.c
+++ b/src/modules/kvs/kvs.c
@@ -2313,9 +2313,10 @@ static void stats_get_cb (flux_t *h, flux_msg_handler_t *mh,
     }
 
     if (flux_respond_pack (h, msg,
-                           "{ s:O s:O }",
+                           "{ s:O s:O s:i }",
                            "cache", cstats,
-                           "namespace", nsstats) < 0)
+                           "namespace", nsstats,
+                           "pending_requests", zhashx_size (ctx->requests)) < 0)
         flux_log_error (h, "%s: flux_respond_pack", __FUNCTION__);
     json_decref (tstats);
     json_decref (cstats);
--- a/src/modules/kvs/kvs.c
+++ b/src/modules/kvs/kvs.c
@@ -1273,11 +1273,10 @@ static void lookup_wait_error_cb (wait_t *w, int errnum, void *arg)
 }
 
 static lookup_t *lookup_common (flux_t *h, flux_msg_handler_t *mh,
-                                const flux_msg_t *msg, void *arg,
+                                const flux_msg_t *msg, struct kvs_ctx *ctx,
                                 flux_msg_handler_f replay_cb,
                                 bool *stall)
 {
-    struct kvs_ctx *ctx = arg;
     int flags;
     const char *ns = NULL;
     const char *key;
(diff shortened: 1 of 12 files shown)


src/common/libfluxutil/Makefile.am
```diff
@@ -9,11 +9,41 @@ AM_CPPFLAGS = \
 	-I$(top_srcdir) \
 	-I$(top_srcdir)/src/include \
 	-I$(top_srcdir)/src/common/libccan \
-	-I$(top_builddir)/src/common/libflux
+	-I$(top_builddir)/src/common/libflux \
+	$(LIBUUID_CFLAGS)
```
garlick, line 13: On moving msg_hash out of librouter - librouter is already included in libflux-internal.la so is it necessary to move it?

chu11: i guess it's not super necessary.  I moved it out just b/c it is generically useful and it feels weird to include a convenience library header from `librouter/msg_hash.h` instead of `libutil/msg_hash.h`.

garlick: Prob not worth it IMHO.  It's just one convenience library vs another.  :shrug:

chu11:
rebased & re-pushed, removing the move of msg_hash from librouter to libfluxutil.

src/modules/kvs/kvs.c
```diff
@@ -3026,6 +3070,15 @@ int mod_main (flux_t *h, int argc, char **argv)
...
     }
+    if (zhashx_size (ctx->requests) > 0) {
+        /* anything that has not yet completed gets an ENOSYS */
+        const flux_msg_t *msg = zhashx_first (ctx->requests);
+        while (msg) {
+            if (flux_respond_error (ctx->h, msg, ENOSYS, NULL) < 0)
+                flux_log_error (ctx->h, "%s: flux_respond_error", __FUNCTION__);
+            msg = zhashx_next (ctx->requests);
+        }
+    }
```
garlick, line 3081: Suggestion: log something at LOG_ERR for each request like:`failing pending <topic> request`.

Also, maybe in all the KVS tests, ensure that this message is not logged just to be sure we don't have a `request_tracking_add()` without a corresponding `request_tracking_remove()` somewhere, since the result of that could be quite bad - like terminating an unrelated RPC with ENOSYS and that will be hard to track down.

Could also add the size of this list to the `flux module stats` output since that may be useful in debugging.

chu11: good call, caught a corner case and possibly another bug unrelated to this PR

chu11:
re-pushed, fixing up things per comments above and adding some extra tests.  The only real "gotcha" on the tests is that some "pending_requests" exist in one test at the end, so those are manually counted.  This is related to the recently opened #6112

garlick (approved):
This LGTM!  Thanks!

PR #6053, opened 2024-06-24
python: Update default Jobspec setattr() path

### LINKED ISSUE #6042: python: `jobspec.setattr()` should probably default to `attributes.system` like the `--setattr` command line option
A user was trying to set a jobspec attribute via the Python bindings using documentation for the command line, which indicated to use `--setattr=key`. However, `--setattr` defaults to `attirbutes.system` whereas `jobspec.setattr()` defaults to just `attributes`. This led to the job being rejected for the extra `key` in the `attributes` section.

It might be good to have the `Jobspec` class `setattr` function behave the same as `--setattr`

chu11:
Problem: By default the Jobspec setattr() function sets keys to attributes.key.  The default --setattr option in command line options like flux-submit default to set keys to attributes.system.key.  This inconsistency can be confusing.

Default Jobspec setattr() to set keys to "attributes.system.key". Adjust command line --setattr option accordingly for tweak.

Fixes #6042

----

Just throwing up this WIP in case there are any initial objections

- TODO - unit tests
- TODO - update callers to specify "foo" instead of "system.foo"?  maybe unnecessary to do this, will think about it

### DIFF
--- a/src/bindings/python/flux/cli/base.py
+++ b/src/bindings/python/flux/cli/base.py
@@ -1046,18 +1046,20 @@ class MiniCmd:
             for keyval in args.setattr:
                 key, val = parse_jobspec_keyval("--setattr", keyval)
 
-                #  If key does not explicitly start with ".", "system."
+                #  If key does not explicitly start with ".", "attributes.", "system."
                 #   or "user.", then "system." is implied. This is a
                 #   meant to be a usability enhancement since almost all
                 #   uses of --setattr will target attributes.system.
                 #
-                if not key.startswith((".", "user.", "system.")):
+                #  Allow users to set keys at the top level by
+                #   specifying "." before the key name, e.g. the key
+                #   ".foo" sets "attributes.foo".
+                if not key.startswith((".", "attributes.", "user.", "system.")):
                     key = "system." + key
+                elif key.startswith("."):
+                    key = "attributes" + key
 
-                #  Allow key to begin with "." which simply forces the key
-                #   to start at the top level (since .system is not applied
-                #   due to above conditional)
-                jobspec.setattr(key.lstrip("."), val)
+                jobspec.setattr(key, val)
 
         if args.add_file is not None:
             for arg in args.add_file:
--- a/src/bindings/python/flux/job/Jobspec.py
+++ b/src/bindings/python/flux/job/Jobspec.py
@@ -554,6 +554,8 @@ class Jobspec(object):
         set job attribute
         """
         if not key.startswith("attributes."):
+            if not key.startswith(("user.", "system.")):
+                key = "system." + key
             key = "attributes." + key
         set_treedict(self.jobspec, key, val)
 
--- a/t/python/t0010-job.py
+++ b/t/python/t0010-job.py
@@ -784,6 +784,29 @@ class TestJob(unittest.TestCase):
         except OSError:
             pass
 
+    def test_35_setattr_defaults(self):
+        """Test setattr setting defaults"""
+        jobspec = Jobspec.from_yaml_stream(self.basic_jobspec)
+        jobspec.setattr("cow", 1)
+        jobspec.setattr("system.cat", 2)
+        jobspec.setattr("user.dog", 3)
+        jobspec.setattr("attributes.system.chicken", 4)
+        jobspec.setattr("attributes.user.duck", 5)
+        jobspec.setattr("attributes.goat", 6)
+        # N.B. setattr defaults "key" to "attributes.system.key"
+        # but getattr defaults "key" to "attributes.key"
+        self.assertEqual(jobspec.getattr("system.cow"), 1)
+        self.assertEqual(jobspec.getattr("attributes.system.cow"), 1)
+        self.assertEqual(jobspec.getattr("system.cat"), 2)
+        self.assertEqual(jobspec.getattr("attributes.system.cat"), 2)
+        self.assertEqual(jobspec.getattr("user.dog"), 3)
+        self.assertEqual(jobspec.getattr("attributes.user.dog"), 3)
+        self.assertEqual(jobspec.getattr("system.chicken"), 4)
+        self.assertEqual(jobspec.getattr("attributes.system.chicken"), 4)
+        self.assertEqual(jobspec.getattr("user.duck"), 5)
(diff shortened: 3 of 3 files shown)


src/bindings/python/flux/cli/base.py
```diff
@@ -1046,18 +1046,17 @@ def jobspec_create(self, args):
...
+                if not key.startswith((".", "attributes.", "user.", "system.")):
                     key = "system." + key
+                elif key.startswith("."):
+                    key = "attributes" + key
 
-                #  Allow key to begin with "." which simply forces the key
-                #   to start at the top level (since .system is not applied
-                #   due to above conditional)
-                jobspec.setattr(key.lstrip("."), val)
+                jobspec.setattr(key, val)
```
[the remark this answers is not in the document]
chu11: Actually it did break a few tests, which is why it's in the same commit.  For example, `--setattr=.foo` wrote to `attributes.foo` before, but now it would write to `attributes.system.foo`.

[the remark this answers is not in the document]
grondo: Hmm, seems like `--setattr=.foo` should continue to have the same behavior?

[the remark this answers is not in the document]
chu11: > Hmm, seems like --setattr=.foo should continue to have the same behavior?

I assume you mean w/o changes to the cli code?  This is the current code w/o my tweaks

```
                #  If key does not explicitly start with ".", "system."
                #   or "user.", then "system." is implied. This is a
                #   meant to be a usability enhancement since almost all
                #   uses of --setattr will target attributes.system.
                #
                if not key.startswith((".", "user.", "system.")):
                    key = "system." + key

                #  Allow key to begin with "." which simply forces the key
                #   to start at the top level (since .system is not applied
                #   due to above conditional)
                jobspec.setattr(key.lstrip("."), val)
```

I believe `--setattr=.foo=bar` would ultimately send `jobspec.setattr("foo", "bar)`.  Under old code that would configure `attributes.foo`, but under new code it would tweak `attributes.system.foo`, thus necessitating changes to cli handling of `--setattr`.

chu11:
For some strange reason and I believe `not ok 17 - attach: reports job shell Killed if job shell is killed` is the culprit.  I can't imagine my change affecting this test.  Adding some dumb debugging commits.

src/bindings/python/flux/cli/base.py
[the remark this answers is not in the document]
grondo: Thanks, I understand now. :+1:

src/bindings/python/flux/cli/base.py
```diff
@@ -1046,18 +1046,17 @@ def jobspec_create(self, args):
...
-                #  If key does not explicitly start with ".", "system."
+                #  If key does not explicitly start with ".", "attributes.", "system."
                 #   or "user.", then "system." is implied. This is a
                 #   meant to be a usability enhancement since almost all
                 #   uses of --setattr will target attributes.system.
                 #
-                if not key.startswith((".", "user.", "system.")):
+                if not key.startswith((".", "attributes.", "user.", "system.")):
                     key = "system." + key
+                elif key.startswith("."):
```
grondo, line 1056: I suggest a comment similar to the deleted one to explain why `.` is handled specifically. (This is what tripped me up on a quick review, though perhaps it is more obvious when looking at  code and not the diff.)

grondo (commented):
Just one trivial comment so far.

chu11:
re-pushed with tests and removed WIP.  I decided not to update any of the `setattr("system.foo", bar)` calls lingering around flux b/c I think the "system.foo" makes it a bit more clear what we're setting/configuring.

t/python/t0010-job.py
```diff
@@ -784,6 +784,30 @@ def test_34_timeleft(self):
...
 
+    def test_35_setattr_defaults(self):
+        """Test setattr setting defaults"""
+        jobspec = Jobspec.from_yaml_stream(self.basic_jobspec)
+        jobspec.setattr("cow", 1)
+        jobspec.setattr("system.cat", 2)
+        jobspec.setattr("user.dog", 3)
+        jobspec.setattr("attributes.system.chicken", 4)
+        jobspec.setattr("attributes.user.duck", 5)
+        jobspec.setattr("attributes.goat", 6)
```
grondo, line 795: Add a test with leading `.` since that's explicitly handled in the method?

src/bindings/python/flux/job/Jobspec.py
```diff
@@ -541,6 +541,8 @@ def getattr(self, key):
         Raises KeyError if a component of key does not exist.
         """
         if not key.startswith("attributes."):
+            if not key.startswith(("user.", "system.")):
```
grondo, line 544: I can see why you also modified `getattr()`, but this seems more dangerous. Since there isn't a "user friendliness" argument for updating `getattr()`, and there is no equivalent `--getattr` command line option which could cause confusion, I wonder if we should just leave getattr alone (it wasn't part of the original bug report)

I'm really on the fence about this one, which is why I think it might be better to not change the behavior.

If nothing else, the documentation here should be updated.

grondo (commented):
Just a couple comments (questions really) inline.

t/python/t0010-job.py
chu11: recall that the leading `.` is only with `--setattr` not with this function.  We can definitely add a test ... i guess `.foo` turns into `attributes.system..foo`? (just thinking off the top of my head right now)

src/bindings/python/flux/job/Jobspec.py
chu11: good point.  I did waflfe on it and changed it mostly for ease of the testing.  I'll revert it.  We can revisit at a later time if necessary.

grondo: It does seem weird that a user could do `jobspec.setattr("foo")` and then `jobspec.getattr("foo")` would raise `KeyError`, but like you said we can deal with that if it comes up.

t/python/t0010-job.py
grondo: Oh, yeah, I guess I confused myself with the various bits of code. Sorry.

chu11:
re-pushed per comments above

grondo (approved):
LGTM. Thanks!!

PR #6060, opened 2024-06-28
etc: add some missing functionality to bash completions

grondo:
Problem: The bash tab completions are currently missing a couple things:

 - the flux-queue(1) completions are missing the --queue option for a couple subcommands
 - the flux-python(1) command doesn't have any completions, so you can't complete the path of a script to pass to `flux python` with tab.

Fix these issues.


chu11 (approved):
LGTM

grondo:
Thanks! I've set MWP.

PR #6062, opened 2024-06-28
job-manager/prioritize: update priority before sending

trws:
I think this might take the cake for me for heisenbug of the week.  Somehow all of this worked correctly until after I fixed the ordering and async issues in qmanager.  My guess is there's a race condition, maybe it worked because in other cases job-manager received the update event and thus updated it sooner? Not sure, I have yet to find where it actually updates the job structure.  Either way, this is what caused the catastrophic CI failure after I fixed the reconsider spinning issue over in flux-framework/flux-sched#1227. Also no clue how this passed all the tests before, the `sched.prioritize` messages consistently carry the wrong value. 🤷‍♂️ 

problem: On receiving a message from `flux job urgency` the job-manager prioritize code updates the value of job->priority after sending the update to the scheduler, using the value of job->priority.  As a result, the scheduler receives an update RPC setting the priority to whatever it was last set to and only on the next such update does it receive the requested value.

solution: update job->priority before sending the RPC to the scheduler

### DIFF
--- a/src/modules/job-manager/prioritize.c
+++ b/src/modules/job-manager/prioritize.c
@@ -114,6 +114,14 @@ static int reprioritize_one (struct job_manager *ctx,
                              "priority", priority) < 0)
         return -1;
 
+    /*  Posting the priority event used to immediately set job->priority,
+     *   but after flux-framework/flux-core#4351, the event may be deferred.
+     *   Since the code below for sched.prioritize and queue reordering needs
+     *   the new job->priority value, set it here *also*.
+     *   See also: flux-framework/flux-core#6062
+     */
+    job->priority = priority;
+
     /*  Update alloc queues, cancel outstanding alloc requests for
      *   newly "held" jobs, and if in "oneshot" mode, notify scheduler
      *   of priority change


src/modules/job-manager/prioritize.c
```diff
@@ -19,6 +19,7 @@
 
 #include "src/common/libjob/idf58.h"
 #include "src/common/libczmqcontainers/czmq_containers.h"
+#include <inttypes.h>
```
grondo, line 22: Just for my own curiosity, why the added include here?

garlick:
Uh oh, yeah, this does look like maybe more fallout from #4351.  As noted in the function that was changed here, posting the job event used to immediately update `job->priority` but that isn't true anymore if the function is called while an event is already being handled.

We'll have to make sure that it's OK to update `job->priority` here as proposed vs moving the scheduler RPC to the `prioritize` event handler.  That comment will need updating too.

src/modules/job-manager/prioritize.c
```diff
@@ -114,6 +115,9 @@ static int reprioritize_one (struct job_manager *ctx,
                              "priority", priority) < 0)
         return -1;
 
+    // update priority
+    job->priority = priority;
+
```
grondo, line 120: Ah, I wonder if the behavior changed here when we stopped recursively posting events. Note the comment above says:

```c
   /* Post 'priority' event.
     *
     *  This call will result in job->priority being set, and,
     *   if the job is in the PRIORITY state, will transition to
     *   the SCHED state, invoke plugin callbacks, etc.
     */
```

This may no longer be the case since the event could be placed in the event queue instead of being posted immediately.

Note that the way this is supposed to work is that when the event handling mechanism in `job-manager/event.c` actually _handles_ the `priority` event, that should set `job->priority`. I wonder if the correct fix is to move all the other stuff in this function after `event_job_post_pack()` to the action for the `priority` event? (Though this approach is much simpler)

I think I'd need @garlick's input on that.

grondo (commented):
Wow, nice find. As noted in comments, the behavior here probably changed awhile ago as a result of unwinding the recursive event posting in the job manager. I'm no longer sure what the "right" fix is, but even if we can't figure it out, your solution here is better than leaving it broken...

src/modules/job-manager/prioritize.c
grondo: Oops, sorry, I cross posted with @garlick.

garlick: NP!  _seems_ like it wouldn't matter where that happens but I can't recall all the details of how the accounting plugin uses this so it might be worth reviewing the RFC and the code a bit first.

I'm on vaca today and likely won't have time but I'm sure I'll be fine with whatever is decided if this need to be fixed now.  O/W I'll check back later this weekend.

Nice job running that down @trws.

src/modules/job-manager/prioritize.c
trws: Possibly vestigial, I was using PRId64 to get some debug prints and it got flagged.  The file does use int64_t though, which makes me think it might actually be correct so I left it.

trws:
Thanks @garlick and @grondo, that makes a whole lot more sense.  I was having trouble working out the chain. Honestly I'm still not sure how it gets from the job-manager urgency code to the reprioritize_one code, but as long as we can get the currently requested value reliably that seems to resolve the issue in the async queues code.

src/modules/job-manager/prioritize.c
trws: We'll need a fix before the async queues PR can land, but I think the plan was early next week anyway.  Normally I'd be happy to move the code over to the event handling, but honestly I'm not sure where that's happening, even with the pointer to event.c.  If one of you can point me in the right direction I can certainly do that though.

grondo: The priority event is handled here I think

https://github.com/flux-framework/flux-core/blob/da43fd3290744ea865d98ba43611c3c57634dbd8/src/modules/job-manager/event.c#L619-L627

But it may require more thought to determine if it is correct to move that code there (and I've got to sign off soon). It does _seem_ correct that the actual _handling_ of the `priority` event should trigger the queue reprioritization and update of the scheduler etc. However, when all jobs are reprioritized, I don't think we want to take this approach (which may re-sort the queue once for every affected job), which is probably why the code is the way it is now...

grondo: My honest inclination would be to leave this fix as-is with a big comment explaining what probably _should_ happen (and open a corresponding issue to look into it). I _think_ it is safe here because that is how the code was originally intended to work.

This is because the async queues PR is very high priority right now.

However, I'd want to get an ACK from @garlick first when he's back, and perhaps @cmoussa1 as well just in case I've missed something obvious.

trws: Sounds right to me.  The concern I would have is that we _might_ see "bouncing" if multiple reprio requests are fired for the same job in quick succession and the value is being set in two places.  Could resolve that by removing the other set of the value from the event and test the result, but I'm not sure if we want to do that or not.

garlick:
>  I'm still not sure how it gets from the job-manager urgency code to the reprioritize_one code

The urgency event causes the job to transition from SCHED to PRIORITY stae.

It's a bit non-obvious especially for single-job urgency updates,  but  jobtap always takes care of calculating priority on entry into the PRIORITY state and calling `reprioritize_all()`, even when flux-accounting's `mf_priority.so` plugin is not loaded.  I forget exactly how it works but in addition to (re-)assigning priortity when a job enters PRIORITY state, the priority plugin can initiate a priority change across all pending jobs and in that case we only want to send one RPC to the scheduler not one per job.  That is why the code is structured like this for better or worse.

I am pretty sure it's safe to move (preferable to duplicating I think) the `job->priority` update to `reprioritize_one()`.  AFAICS the two things the "priority" event does are 1) publish/record the new value and 2) transition the job to SCHED state.   It seems like setting it just before posting the event should be fine for those cases and as @grondo pointed out, that was how it was originally designed.

For future explorers, it might be useful to add a comment in `event_job_action()` to the effect of
```c
/* When a job enters PRIORITY state, jobtap handles priority (re-)calculation
 * and calls reprioritize_all(), which posts the "priority" event, which causes the job
 * to advance to the SCHED state.
 */
```
That sound right @grondo?

trws:
Ok, I'll look at adding some comments, and disable the re-update in event so we don't get value bouncing.

garlick:
I can take care of this minor stuff.  Stop me @trws if you already have this done but just haven't pushed.

We should probably get this in to keep things moving, given the suffering on el cap right now.

trws:
I haven't, if you're willing go for it, and thanks @garlick!

garlick:
OK, this is a little more involved than I thought.  The replay logic recreates a job struct, including priority, from just the eventlog, so we're kind of breaking the model (and a bunch of unit tests) if we don't update `job->priority` in `event_job_update()`.

Updating `job->priority` in two places  as originally proposed may be the least bad option.

As far as bouncing goes, since we only reprioritize the alloc queue and update the scheduler in `reprioritize_all()`, I don't think there's too much of a worry that if we update in two places, we might sort the queue wrong or send the scheduler the wrong information.   Anything external like job-list or job-info should be working solely off of the job events.  The one vaguely concerning thing is `priority` is a jobtap plugin argument and the value comes from `job->priority`, so a plugin that is driven by the `priority` event might get a value associated with a future event rather than the one associated with the event its handling.  But is that a real case to be concerned with?  I do not think so.

Maybe in the interest of stopping the bleeding, we could merge this as is and open an issue to follow up, which (sigh) was the original suggestion from @grondo.

I'll think about this a bit more while I run an errand.

garlick (approved):
I left the code changes as originally proposed but dropped the unrelated inttypes include and expanded the comment.  I think this is OK to go in, but let's get one more ack from @grondo first.

trws:
Works for me, also like the new comment, and thanks for the explanation @garlick!

grondo (approved):
Thanks @trws and @garlick, this LGTM!

garlick:
setting MWP

PR #6063, opened 2024-06-28
add environment variable to avoid `RTLD_DEEPBIND` when loading modules

trws:
Add an environment variable to control whether we actually apply the RTLD_DEEPBIND flag when calling dlopen.  More details in the commit message, but this makes it possible to run flux with an alternate allocator using LD_PRELOAD or certain other things that make heap tracing and certain other debugging and tracing tasks much easier.

### DIFF
--- a/src/common/libflux/plugin.c
+++ b/src/common/libflux/plugin.c
@@ -23,7 +23,9 @@
 #include <jansson.h>
 #include <assert.h>
 #include <flux/core.h>
+#include <pthread.h>
 
+#include "src/common/libflux/plugin_private.h"
 #include "src/common/libczmqcontainers/czmq_containers.h"
 #include "src/common/libutil/aux.h"
 #include "ccan/str/str.h"
@@ -34,6 +36,24 @@
 #define UUID_STR_LEN 37     // defined in later libuuid headers
 #endif
 
+static int use_deepbind = 1;
+static pthread_once_t deepbind_once = PTHREAD_ONCE_INIT;
+
+static void init_use_deepbind (void)
+{
+    const char *var = getenv ("FLUX_LOAD_WITH_DEEPBIND");
+    if (!var) {
+        // default is to allow the flag
+        return;
+    }
+    use_deepbind = strtol (var, NULL, 10);
+}
+
+int plugin_deepbind (void)
+{
+    pthread_once (&deepbind_once, init_use_deepbind);
+    return use_deepbind != 0 ? FLUX_DEEPBIND : 0;
+}
 
 struct flux_plugin {
     char *path;
@@ -270,7 +290,7 @@ static int open_flags (flux_plugin_t *p)
     else
         flags |= RTLD_LOCAL;
     if ((p->flags & FLUX_PLUGIN_RTLD_DEEPBIND))
-        flags |= FLUX_DEEPBIND;
+        flags |= plugin_deepbind ();
     return flags;
 }
 
--- a/src/common/libflux/plugin_private.h
+++ b/src/common/libflux/plugin_private.h
@@ -1,0 +1,16 @@
+/************************************************************\
+ * Copyright 2024 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+#ifndef FLUX_CORE_PLUGIN_PRIVATE_H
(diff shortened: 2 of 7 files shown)


src/common/libflux/plugin.h
```diff
@@ -24,6 +24,8 @@ enum {
     FLUX_PLUGIN_RTLD_DEEPBIND =  8,  /* Load with RTLD_DEEPBIND if available */
 };
 
+int flux_deepbind ();
+
```
grondo, line 28: Does this helper really need to be exported via the public API? I'm not sure if you had a use case for that.

If not, it could possibly go into libutil, or it could get a function name that doesn't start with `flux_` and the prototype in a `plugin_private.h` header. (We've done that with some other APIs in libflux, see e.g. `message_private.h` etc.)

src/common/libflux/plugin.c
```diff
@@ -34,6 +34,15 @@
 #define UUID_STR_LEN 37     // defined in later libuuid headers
 #endif
 
+int flux_deepbind () {
+    const char *var = getenv("FLUX_LOAD_WITH_DEEPBIND");
```
grondo, line 38: This new variable could be added to the `flux-environment(7)` man page.

grondo: nit: space between `getenv` and opening parens for style consistency.

grondo (commented):
This LGTM. I just wonder about the need for `flux_deepbind()` in the public API.

Also, I wonder if the result of `getenv ()` could be cached so we don't check multiple times for an environment variable that will almost never be present (or that could just be premature optimization)

src/common/libflux/plugin.h
trws: It does not. I'm fine with either solution, it's mainly where it is because all the callers already use this header.  Do you have a preference?

grondo: I do not have a preference. Personally I'd just rename the function and put it in `plugin_private.h` just because that's less work :shrug: If we think there might be more functions like this in the future, we may want to move it to a new `.c` and `.h` in libutil :shrug: (easy to move later though)

trws: Pushed an update to rename it, move into a private header, and only read the env var once.  I'm not completely sure this is always called from the same thread, so it's using pthread_once to initialize the value once and then re-use it.  Makes me miss the c++11 "statics are initialized exactly once" guarantee. 😞

src/common/libflux/plugin_private.h
```diff
@@ -0,0 +1,16 @@
+/************************************************************\
+ * Copyright 2019 Lawrence Livermore National Security, LLC
```
grondo, line 2: nit: Copyright 2024

grondo (approved):
LGTM! One nit inline, but otherwise seems good to me

src/common/libflux/plugin.c
```diff
@@ -34,6 +36,21 @@
 #define UUID_STR_LEN 37     // defined in later libuuid headers
 #endif
 
+static int use_deepbind = 1;
+static pthread_once_t deepbind_once = PTHREAD_ONCE_INIT;
+static void init_use_deepbind () {
```
garlick, line 41: Style: void parameter, open brace on new line

src/common/libflux/plugin.c
```diff
@@ -34,6 +36,21 @@
...
+    const char *var = getenv ("FLUX_LOAD_WITH_DEEPBIND");
+    if (!var) {
+        // default is to allow the flag
+        return;
+    }
+    use_deepbind = strtol (var, NULL, 10);
+}
+
+int plugin_deepbind () {
+    pthread_once(&deepbind_once, init_use_deepbind);
```
garlick, line 51: Style: open brace for function on new line, space before open paren in pthread_once

garlick (approved):
Couple of style nits but o/w LGTM too

garlick (approved):
@grondo - all your changes/fixes are great.  I'll approve based on that and you can add yours once you feel like this is ready.

trws:
I think that's all addressed.

src/common/libflux/plugin_private.h
```diff
@@ -0,0 +1,16 @@
...
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+#ifndef FLUX_CORE_PLUGIN_PRIVATE_H
+#define FLUX_CORE_PLUGIN_PRIVATE_H
+
+int plugin_deepbind ();
```
garlick, line 14: style: add void  param here too

trws: Done.

garlick:
Restarted the inception builder which had failed.  I could not discern why.

PR #6064, opened 2024-06-28
actions: only run CI for push on master

trws:
Actions are being slow today, and I think this is part of why.

problem: we currently run CI in full on both push and PR for all branches, and it mostly just fails on push because it doesn't receive label context. What we really need is for push triggered builds to run on master and only PR-triggered otherwise.

solution: limit push event triggers to the master branch.

### DIFF
--- a/.github/workflows/asan.yml
+++ b/.github/workflows/asan.yml
@@ -1,4 +1,8 @@
-on: [push, pull_request]
+on:
+  pull_request:
+  push:
+    branches:
+      - master
 name: asan
 jobs:
   check-asan:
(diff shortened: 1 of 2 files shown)


grondo:
I make heavy use of CI on push to a branch to check my personal fork before opening a PR. I'm assuming this will break that behavior?

trws:
I actually had to dig into that a bit to find an answer.  If we merged this exactly as-is, yes I think it would break that.  It also seems the behavior is different for different forks, in fact the tests run on my fork but almost always break, badly.  Probably need to do more research here.

PR #6070, opened 2024-07-02
job-exec: add `active_ranks` and fix `active_shells` count in per-job stats

grondo:
This PR fixes the `active_shells` count in the job-exec per-job module stats and also adds an `active_ranks` member with an idset of ranks on which job shells are still active. This can aid in finding an unkillable shell in large jobs.


garlick (approved):
Great!  This will be helpful.

grondo:
Thanks, I've set MWP.

PR #6074, opened 2024-07-02
flux-job: fix prolog refcount in attach status message

grondo:
Problem: The `flux job attach` status line reference counting for prolog-start events doesn't work because job_event_notify_string(), which is where the refcount is done, is only called when the status line is active -- that is after 2s by default. The first prolog-start events could have been posted during this time, and the increment of the prolog-finish refcount is missed.

Always call job_event_notify_string() in attach_notify() to ensure the refcounting is correct.


garlick (approved):
:+1:

grondo:
Thanks! Setting MWP.

Issue #6059, opened 2024-06-27
job-manager: underflow in alloc request to scheduler count


grondo:
After a restart of Flux on elcap:
```
13 alloc requests queued
-9 alloc requests pending to scheduler
```

Even after removing the fluxion modules, the `alloc requests pending` count is still negative, though the queued count went down to zero.

grondo:
I restarted flux, which was in a good state until a queue was started, then we got the same results as above.
This was noted in the logs:
```
[ +33.929493] job-manager[0]: sched.alloc-response: id=f24ehES5k3J7 already allocated
[ +33.929519] job-manager[0]: alloc: stop due to alloc response error: File exists
```

grondo:
After canceling the pending affected job and restarting, the same issue occurred, just with a different job in the queue. Another cancel and restart confirmed this.

grondo:
This seemed to only affect the jobs in one queue. Starting other queues individually had jobs running without the scheduler alloc-response error here.

grondo:
One possible improvement to avoid the negative alloc pending count would be to decrement the sent_count after the check for a job that already has resources:
```diff
diff --git a/src/modules/job-manager/alloc.c b/src/modules/job-manager/alloc.c
index 58cbe0c12..9ba82007f 100644
--- a/src/modules/job-manager/alloc.c
+++ b/src/modules/job-manager/alloc.c
@@ -182,7 +182,6 @@ static void alloc_response_cb (flux_t *h,
             goto teardown;
         }
         (void)json_object_del (R, "scheduling");
-        alloc->sent_count--;
 
         if (!job) {
             (void)free_request (alloc, id, R);
@@ -200,6 +199,7 @@ static void alloc_response_cb (flux_t *h,
             errno = EEXIST;
             goto teardown;
         }
+        alloc->sent_count--;
         job->R_redacted = json_incref (R);
         if (annotations_update_and_publish (ctx, job, annotations) < 0)
             flux_log_error (h, "annotations_update: id=%s", idf58 (id));
```

garlick:
Probably fixed by #6076.  Reopen if we see a problem.

PR #6076, opened 2024-07-02
job-manager: improve sched.alloc request tracking

garlick:
Problem: we observed underflow in the `flux queue status -v` pending alloc count in #6059 with fluxion.

I was unable to reproduce the underflow.  This PR adds tests since there were none for the alloc/pending counts, and simplifies the code.

TL;DR In limited mode, the job manager places pending alloc requests in a priority queue so that it can reorder them and cancel/resubmit if priorities change.  In unlimited mode, the list was eschewed for a counter.  It's the counter that went wrong here (somehow).   The cleanup is to always use the priority queue so that we only have one code path to worry about and since list removal is idempotent, the underflow case should be gone.

I can't imagine the priority queue presents a performance issue.  I did run a quick check, running `flux module reload sched-simple mode=unlimited` and running the throughput test with 1000 jobs 3x each on master and this PR and didn't see much of a difference, if any.

### DIFF
--- a/src/modules/job-manager/alloc.c
+++ b/src/modules/job-manager/alloc.c
@@ -48,7 +48,6 @@ struct alloc {
     flux_watcher_t *check;
     flux_watcher_t *idle;
     unsigned int alloc_limit;   // will have a value of 0 in mode=unlimited
-    unsigned int sent_count;    // track number of jobs w/ alloc reqs
     char *sched_sender;         // scheduler uuid for disconnect processing
 };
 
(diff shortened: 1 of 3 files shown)


src/modules/job-manager/alloc.c
```diff
@@ -48,7 +48,6 @@ struct alloc {
     flux_watcher_t *check;
     flux_watcher_t *idle;
     unsigned int alloc_limit;   // will have a value of 0 in mode=unlimited
-    unsigned int sent_count;    // track number of jobs w/ alloc reqs
```
grondo, line 51: Commit message typo: Eliminte

grondo (approved):
LGTM! Just noticed a commit message typo

garlick:
Fixed that and forced a push.  I'll set MWP.

grondo:
Just forced a rebase so we don't have to wait for two rounds of checks. Hope that was OK!

PR #6082, opened 2024-07-04
libsubprocess: invert SETPGRP flag logic

chu11:
Problem: Remote subprocesses default to always set the SETPGRP flag.  This leads to inconsistent behavior where local subprocesses don't have the flag set but remote ones do.

We could remove this default and require remote subprocess execution to set this flag as needed.  However, it ends up that the vast majority of code in flux-core sets the SETPGRP flag anyways, leading to an excess amount of code change to make local vs remote subprocesses consistent.

Solution: Invert the SETPGRP flag logic into NO_SETPGRP.  Now, by default, have both local and remote subprocesses call setpgrp(2).  The NO_SETPGRP flag will then disable the call to setpgrp(2).  Update all callers to libsubprocess accordingly.

Note that we leave the new NO_SETPGRP flag to be for local subprocesses only.  If there comes a time for it to be supported for remote subprocesses, that will be for another day.

### DIFF
--- a/src/broker/runat.c
+++ b/src/broker/runat.c
@@ -350,10 +350,10 @@ static struct runat_command *runat_command_create (char **env, int flags)
     if (flags & RUNAT_FLAG_FORK_EXEC)
         cmd->flags |= FLUX_SUBPROCESS_FLAGS_FORK_EXEC;
     /*
-     * Always run runat command in separate process group so that any
-     * processes spawned by command are signaled by flux_subprocess_signal()
+     * N.B. By default subprocesses call setpgrp() before exec(2).  So
+     * any processes spawned by command are also signaled by
+     * flux_subprocess_signal()
      */
-    cmd->flags |= FLUX_SUBPROCESS_FLAGS_SETPGRP;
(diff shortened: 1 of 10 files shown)


garlick:
> We could remove this default and require remote subprocess execution to set this flag as needed. However, it ends up that the vast majority of code in flux-core sets the SETPGRP flag anyways, leading to an excess amount of code change to make local vs remote subprocesses consistent.

API wise, NO_SETPGRP seems vaguely wrong to me, but I see where you're coming from about updating all the tests n stuff.  @grondo any thoughts?

chu11:
> API wise, NO_SETPGRP seems vaguely wrong to me, but I see where you're coming from about updating all the tests n stuff. @grondo any thoughts?

Updating everything to set the SETPGRP flag is certainly ok and fine too, we could do that.

It's mostly to get this consistent, b/c the follow up will be to allow subprocess flags to be passed to remote subprocesses.

grondo:
Yeah, I was struggling with NO_SETPGRP too, but the simplicity is indeed appealing. I think this PR is probably fine as is.

grondo (approved):
LGTM. It did nicely match the shell nosetpgrp option.

PR #6092, opened 2024-07-11
doc: add overlay network to overview section

garlick:
Problem: we don't have any documentation to help with the selection of overlay network topology.

Add a section to the admin guide "Overview and Background" section.

### DIFF
--- a/doc/guide/admin.rst
+++ b/doc/guide/admin.rst
@@ -249,6 +249,75 @@ See also:
 `RFC 12 <https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_12.html>`_,
 `RFC 15 <https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_15.html>`_.
 
+Overlay Network
+===============
+
+As described above, a Flux instance consists of one or more Flux brokers
+communicating over a tree-based overlay network.  A Flux system instance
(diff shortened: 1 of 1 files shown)


doc/guide/admin.rst
```diff
@@ -249,6 +249,70 @@ See also:
...
+
+Topology for Large Clusters
+---------------------------
+
+To avoid the above downsides, larger clusters should use a custom topology
+with tree height of 2 and internal brokers assigned to stable, well connected,
+non-busy nodes.  The downside of this topology is, obviously:
+
+- More brokers are critical
+
```
grondo, line 315: Suggestion: add a short descriptoin about how to use the `bootstrap` config to create the two-level hierarchy and reference `flux-config-bootstrap(5)`? (Or maybe just reference the 3rd example in that manual)

grondo (approved):
Great! This is helpful! Just one take-it-or-leave-it suggestion inline.

doc/guide/admin.rst
grondo: I didn't reread the whole document, so perhaps that would be redundant. Ignore me if so

garlick: Sounds like a good idea.  That will probably make all the words more directly useful.

garlick:
Thanks!  I added those references to the bootstrap examples and will set MWP.

PR #6094, opened 2024-07-11
doc: update flux-shell(1) manpage

wihobbs:
Problem: there's inconsistency in the flux-shell(1) page, where some lua syntax is used and other places use pseudo- code.

Make it consistent by updating "+" to ".." (Lua).


grondo (approved):
Thanks! Nice catch!

PR #6101, opened 2024-07-13
drain nodes with unkillable tasks after configurable retries

### LINKED ISSUE #6097: job-exec: add a timeout after a fatal job exception to force `finish` response
Problem: In an increasing number of cases, one or more job shells are hung in an unkillable state on nodes. In this case, the job-exec system simply keeps trying to kill the job shells until the end of time or admins intervene and reboot stuck nodes.

The job-exec system should allow configuration of a timeout after which it will give up trying to terminate the remaining job shells, drain the ranks with the unkillable shells and force the finish response to the job manager. This will relieve admins from having to take manual action.

The only wrinkle is that the epilog will probably be launched across all nodes of the job and may also get hung on the affected nodes. However, that case will possible relieved by the housekeeping service.

grondo:
This PR modifies the job-exec module to add a max number of retries when signaling the job shell. When this count is exceeded, the remaining active ranks are drained and the tasks are forced to complete so that the job can proceed to CLEANUP.

Fixes #6097

### DIFF
--- a/src/modules/job-exec/job-exec.c
+++ b/src/modules/job-exec/job-exec.c
@@ -108,6 +108,7 @@
 #include "exec_config.h"
 
 static double kill_timeout;
+static int max_kill_count;
 static int term_signal;
 static int kill_signal;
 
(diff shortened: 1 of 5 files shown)


src/modules/job-exec/job-exec.c
```diff
@@ -399,6 +437,21 @@ static void kill_shell_timer_cb (flux_reactor_t  *r,
...
+                  LOG_DEBUG,
+                  "job %s exceeded max kill count. Draining active ranks",
+                  idf58 (job->id));
+        if (drain_active_ranks (job, active_ranks) == 0) {
+            /* Force remaining tasks to be complete
+             */
+            jobinfo_tasks_complete (job, active_ranks, 1);
+        }
+        idset_destroy (active_ranks);
+    }
```
garlick, line 454: It looks like the kill timer is re-armed even when the active ranks have been forcibly drained.  Is there a reason to do that?

garlick (approved):
Looks great!  Just one comment - more of a question.

src/modules/job-exec/job-exec.c
grondo: It shouldn't matter because the job entry and all associated timers will be destroyed after `jobinfo_tasks_complete()` is called. However, would be less confusing to move this conditional above the re-arm of the timer? The code would then have to return early, but only if `drain_active_ranks()` succeeds.

garlick: Maybe add a comment explaining that it's not going to do anything if `drain_active_ranks()` succeeds but might be required if it fails?

grondo: Ok, I added this comment, I hope that is sufficient?
```c
    /* Check if we've exceeded the maximum number of kill attempts.
     * Drain ranks that are still active if so. If the drain succeeds,
     * then force remaining tasks to be complete, which will then cause
     * this job object to be destroyed (which terminates any active kill
     * timers).
     *
     * If the drain fails (unlikely), then the job stays active and we'll
     * try to kill it again (and drain ranks) the next time the kill timer
     * fires.
     */
```

garlick: Great!  Thanks.

grondo:
Thanks, setting MWP

PR #6102, opened 2024-07-13
flux-job: add `-F, --follow` option to `flux job eventlog`

grondo:
This is a simple one I had laying around for a bit. 

Sometimes it is convenient to "follow" a job eventlog, which can be done with `flux job eventlog -v JOBID clean`, but that isn't quite as user friendly and obvious as a `--follow` option for `flux job eventlog`. This PR makes `flux job eventlog --follow` a synonym for `flux job wait-event -v JOBID clean` via some refactoring. 

Unfortunately, `-f ` is already taken by `--format`, so the short option for follow becomes `-F`.

### DIFF
--- a/t/t2230-job-info-lookup.t
+++ b/t/t2230-job-info-lookup.t
@@ -218,7 +218,21 @@ test_expect_success 'flux job eventlog -p fails on invalid path' '
 	jobid=$(submit_job) &&
 	test_must_fail flux job eventlog -p "foobar" $jobid
 '
-
+# submit job in separate test to avoid using the form:
+#  jobid=$(flux submit ..) && flux job eventlog $jobid ... &
+# which will pick up the wrong $jobid since the pipeline is placed into
+# the background:
+test_expect_success 'submit held job for eventlog follow test' '
+	jobid=$(flux submit --urgency=hold -n1 hostname)
+'
+test_expect_success NO_CHAIN_LINT 'flux job eventlog -F, --follow works' '
+	flux job eventlog -HF $jobid >eventlog-follow.out &
+	pid=$! &&
+	flux job urgency $jobid default &&
+	wait $pid &&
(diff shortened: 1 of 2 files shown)


t/t2230-job-info-lookup.t
```diff
@@ -218,7 +218,21 @@ test_expect_success 'flux job eventlog -p fails on invalid path' '
...
+	jobid=$(flux submit --urgency=hold -n1 hostname)
+'
+test_expect_success NO_CHAIN_LINT 'flux job eventlog -F, --follow works' '
+	flux job eventlog -HF $jobid >eventlog-follow.out &
+	pid=$! &&
+	flux job urgency $jobid default &&
+	wait $pid &&
+	test_debug "cat eventlog-follow.out" &&
+	grep clean eventlog-follow.out
+'
```
chu11, line 235: FWIW, since the test is already NO_CHAIN_LINT I normally just put the job submission into the test without the && chain, i.e.

```
test_expect_success NO_CHAIN_LINT 'flux job eventlog -F, --follow works' '
        jobid=$(flux submit --urgency=hold -n1 hostname)
	flux job eventlog -HF $jobid >eventlog-follow.out &
	pid=$! &&
	flux job urgency $jobid default &&
	wait $pid &&
	test_debug "cat eventlog-follow.out" &&
	grep clean eventlog-follow.out'
```

chu11 (approved):
seems like a good idea, just one comment below on a test

t/t2230-job-info-lookup.t
grondo: That makes sense, but at least the current way would show an error (not ok) if the job submission did fail (unlikely though) and extra `test_expect_success` is cheap..

grondo:
Thanks! I've set MWP.

Issue #6009, opened 2024-05-28
resource: constant load on large system


grondo:
This was briefly discussed in the dev meeting last week, but there's a somewhat constant load (40% of a CPU) on the rank 0 broker on a large system attributed to the `resource` module here:
```
-   56.31%     0.00%  resource         libc-2.28.so           [.] __GI___clone ◆
     __GI___clone (inlined)                                                    ▒
     start_thread                                                              ▒
     module_thread                                                             ▒
     mod_main                                                                  ▒
     flux_reactor_run                                                          ▒
     ev_run                                                                    ▒
   - ev_run                                                                    ▒
      - 56.31% ev_invoke_pending                                               ▒
         - 56.29% sched_status_continuation                                    ▒
            - 56.27% prepare_sched_status_payload (inlined)                    ▒
               + 29.76% get_all (inlined)                                      ▒
               + 14.34% create_rlist (inlined)                                 ▒
               + 6.49% get_down (inlined)                                      ▒
               + 2.09% update_properties_json (inlined)                        ▒
               + 1.83% rlist_destroy                                           ▒
               + 1.39% get_empty_set (inlined)                                
```

Probably not critical, but caching the "all" resource set seems like a simple approach that would have  a large impact.

garlick:
Thanks - sounds like we can save job manager caching for another day then.

PR #6105, opened 2024-07-15
resource: cache resource status information to reduce broker load

garlick:
Problem: on a large system, frequent use of `flux resource list` or `flux resource status` triggers significant broker CPU consumption.  Commands like `flux top` or `watch flux resource list` are not unreasonable things to do, yet they are costly.

As noted in #6009, we repeatedly calculate the same resource sets, even when they don't change.

Add a simple cache.

### DIFF
--- a/src/modules/resource/status.c
+++ b/src/modules/resource/status.c
@@ -22,16 +22,38 @@
 #include "monitor.h"
 #include "exclude.h"
 #include "status.h"
+#include "reslog.h"
 
 #include "src/common/libutil/errprintf.h"
 #include "src/common/libutil/errno_safe.h"
 #include "src/common/librlist/rlist.h"
+#include "ccan/str/str.h"
+
+struct status_cache {
+    struct rlist *rl;   // exclusions removed
+    json_t *R_all;
+    json_t *R_down;
+};
 
 struct status {
     struct resource_ctx *ctx;
     flux_msg_handler_t **handlers;
     struct flux_msglist *requests;
+    struct status_cache cache;
+    json_t *R_empty;
 };
+
+static void invalidate_cache (struct status_cache *cache, bool all)
+{
+    if (all) {
+        rlist_destroy (cache->rl);
+        cache->rl = NULL;
+        json_decref (cache->R_all);
+        cache->R_all = NULL;
+    }
+    json_decref (cache->R_down);
+    cache->R_down = NULL;
+}
 
 static json_t *prepare_status_payload (struct status *status)
 {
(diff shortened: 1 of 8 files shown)


garlick:
Pushed a little tweak just now but basically this is ready for review.

src/modules/resource/status.c
```diff
@@ -442,6 +463,23 @@ static void sched_status_cb (flux_t *h,
...
+ */
+static void reslog_cb (struct reslog *reslog,
+                       const char *name,
+                       json_t *context,
+                       void *arg)
+{
+    struct status *status = arg;
+
+    invalidate_cache (&status->cache);
+}
```
grondo, line 481: In perf tracing, creating the `all` resource set was the most expensive operation, and `R_all` should only be invalidated by the first two events in the list above. Would it add too much complexity to save `R_all` unless a `resource-define` or `resource-update` event is seen? (you could perhaps add an optional event name parameter to `invalidate_cache`?)

On a very large system I fear the `online` `offline` `drain` and `undrain` events will be so frequent that we won't see as much benefit from the caching as we might otherwise.

src/modules/resource/status.c
```diff
@@ -8,6 +8,12 @@
  * SPDX-License-Identifier: LGPL-3.0
  \************************************************************/
 
+/* status.c - handle resource.status and resource.sched-status RPCs
+ *
+ * Although this portion of the resource module is initialized on all
+ * ranks, the RPCs are denied if requested on rank > 0.
+ */
+
```
grondo, line 16: Not critical, but it seems like this comment could go in a separate commit? (The comment does make me wonder why the status portion of the `resource` module is initialized on all ranks when RPCs are denied on those ranks other than 0, though)

grondo (approved):
Nice! This LGTM! One suggestion inline. If you feel strongly the current approach is good, though feel free to merge as it is a big improvement.

src/modules/resource/status.c
garlick: Sure, I'll do that.

src/modules/resource/status.c
garlick: Sure, will do.

> The comment does make me wonder why the status portion of the resource module is initialized on all ranks when RPCs are denied on those ranks other than 0, though)

Just to deny the RPCs with a better message than "service not implemented' I think?  Hmm, maybe not worth it.  :shrug:

grondo: Eh, not really worth addressing now I'd say

garlick:
I reworked the caching so that `R_all`  and the rlist are not invalidated by online/offline/drain/undrain.

One thing I'm sort of not happy with is the extra rlist copy here.  I had to add it because before, we marked nodes down in the primary rlist and then just had to copy it once with `rlist_copy_down()`.  Now in order to not cache down info in the main rlist, I'm making a copy then marking things up and down in the copy, then calling `rlist_copy_down()`.  Boy I feel dumb even typing this because I'm sure there's a better way.  I'll have a look tomorrow.
```c
static json_t *get_down (struct status *status, const struct rlist *rl)
{
    if (!status->cache.R_down) {
        struct rlist *cpy;
        struct rlist *rdown = NULL;
        struct idset *drain = NULL;
        const struct idset *down = monitor_get_down (status->ctx->monitor);

        if ((cpy = rlist_copy_empty (rl))
            && rlist_mark_up (cpy, "all") >= 0
            && (drain = drain_get (status->ctx->drain))
            && mark_down (cpy, down) == 0
            && mark_down (cpy, drain) == 0
            && (rdown = rlist_copy_down (cpy)))
            status->cache.R_down = rlist_to_R (rdown);
        idset_destroy (drain);
        rlist_destroy (rdown);
        rlist_destroy (cpy);
    }
    return status->cache.R_down;
}
```

garlick:
Re-pushed a new version of that function that just modifies its rlist argument, which avoids the extra copy and should be fine for all users of `cache->rl` (comment to that effect added).

garlick:

PR #6106, opened 2024-07-15
t: add job environment to tests

### LINKED ISSUE #6088: t0015-job-output.py failures in RHEL9
Run via `src/cmd/flux t/python/t0015-job-output.py` so I can see the error output b/c for some reason tests are taking forever to fail.

```
not ok 1 __main__.TestJobOutput.test_job_output
  ---
    message: |
        File "/home/chu11/code/flux-core/t/./python/t0015-job-output.py", line 98, in test_job_output
          self.assertEqual(output.stderr, self.test_stderr)
      AssertionError: 'bash: gettext: \nbash: gettext: \nbash: g[53678 chars] 2\n' != 'error 1\nerror 2\n'
      Diff is 56885 characters long. Set self.maxDiff to None to see it.
  ...
```

I'm guessing something underneath the covers changed in RHEL9.  No obvious clue what it could be at the moment.

chu11:
Problem: The python/t0015-job-output.py test has an empty environment by default.  This can lead to problems on some distributions.

Set an environment equivalent to the current one to avoid this problem.

Fixes #6088

### DIFF
--- a/t/python/t0015-job-output.py
+++ b/t/python/t0015-job-output.py
@@ -62,6 +62,7 @@ class TestJobOutput(unittest.TestCase):
         )
         urgency = FLUX_JOB_URGENCY_DEFAULT
         jobspec.setattr_shell_option("cpu-affinity", "off")
+        jobspec.environment = dict(os.environ)
         if hold:
             urgency = FLUX_JOB_URGENCY_HOLD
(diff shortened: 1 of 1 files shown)


grondo (approved):
Thanks!

PR #6108, opened 2024-07-16
flux-start: accept --recovery with --test-size

garlick:
Problem: `flux-start` ignores `--recovery` if `--test-size` is present.

Handle the two options together.


grondo (approved):
Nice, thanks! LGTM.

Issue #6109, opened 2024-07-17
idea: record ranks/hostnames of all clients trying to connect with old version


grondo:
On a large machine it is unfortunately common that some set of nodes fail to get upgraded to the latest OS image, RPMs, etc. The larger the machine, the more opportunity for this to happen, and the more the logs fill with errors like:
```
[ +18.175996] broker[0]: rejecting connection from somehost (rank X): client (0.63.0) version mismatched with server (0.64.0)
```
This actually makes it a bit difficult to get the list of hosts for which connections are rejected due to a version mismatch. The log messages are accumulating rapidly when there are many hosts in this state, so some sed and awk-fu is required.

Another idea would be to capture the ranks that are offline due to a version mismatch in some way. Then this list could easily be queried and reported via a simple tool. Also, the broker can then track ranks for which it already produced this log message and can stop repeating it.

Once a rank joins with the correct version, it can be cleared from the list. (oh, I guess that would have to be added in the common path, so maybe not a good idea after all) 

Well an idea to throw out there anyway.

garlick:
That seems like a good idea to me.  What would the tooling look like?  Should it report the mismatched version by host?

If we didn't care about the version, the broker could maybe just keep a hostlist in a broker attribute.

grondo:
I was thinking of a broker attribute as well. Or if something like `flux module stats` worked with the broker itself, but that didn't seem quite right.

garlick:
Oh maybe `flux overlay errors` would be a good place for a query tool.

PR #6111, opened 2024-07-17
add error text to down nodes in flux overlay status

garlick:
Problem: as discussed in #6109, spelunking logs to find out why a node is offline is tedious, and repeated messages about mismatched versions decrease the log signal to noise ratio.

This adds an error message to `flux overlay status` for each offline node, if available, and eliminates the mismatched version log message.

Example:
```
$ flux overlay disconnect 39
flux-overlay: asking system76-pc (rank 1) to disconnect child system76-pc (rank 39)

$ flux overlay status --down -vvv
0 system76-pc: degraded for 8.38533s (0.377 ms)
├─ 1 system76-pc: degraded for 8.386s (0.328 ms)
│  └─ 39 system76-pc: lost for 8.386s administrative disconnect
```
Marking WIP to get feedback on this approach before proceeding to address test issues.

### DIFF
--- a/t/t3303-system-healthcheck.t
+++ b/t/t3303-system-healthcheck.t
@@ -128,7 +128,7 @@ test_expect_success 'flux overlay status -vv works' '
 '
 
 test_expect_success 'flux overlay status shows rank 3 offline' '
-	echo "3 fake3: offline" >health.exp &&
+	echo "3 fake3: offline administrative shutdown" >health.exp &&
 	flux overlay status --timeout=0 --no-pretty \
 		| grep fake3 >health.out &&
 	test_cmp health.exp health.out
--- a/t/t3303-system-healthcheck.t
+++ b/t/t3303-system-healthcheck.t
@@ -65,6 +65,11 @@ test_expect_success 'overlay status is full' '
 	test "$(flux overlay status --timeout=0 --summary)" = "full"
 '
 
+test_expect_success 'flux overlay errors prints nothing' '
+	flux overlay errors --timeout=0 >errors.out &&
+	test $(wc -l <errors.out) -eq 0
+'
+
 test_expect_success 'wait timeout of zero is not an immediate timeout' '
 	flux overlay status --wait=full --summary --timeout=0
 '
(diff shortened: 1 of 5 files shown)


garlick:
Tests fixed.

I guess the question is whether `flux overlay status --down` produces compact enough output or if we want something like a `flux overlay error` that tries to combine lines with the same error and make a hostlist.

grondo:
Admins will likely want to be able to use some command to pass to `pdsh -w` or `clush`. If we can assemble the offline nodes by reason into a hostlist, that would be ideal, but this is at least a start.

garlick:
> Admins will likely want to be able to use some command to pass to pdsh -w or clush

Great point!

garlick:
I added a `flux overlay errors` command.  Example:
```
$ flux overlay status
0 picl0: degraded
├─ 1 picl1: degraded
│  ├─ 3 picl3: full
│  ├─ 4 picl4: full
│  └─ 5 picl5: lost lost connection
└─ 2 picl2: lost lost connection
   ├─ 6 picl6: lost
   └─ 7 picl7: lost

$ flux overaly errors
picl[2,5]: lost connection
picl[6-7]: lost parent
```

t/t3303-system-healthcheck.t
```diff
@@ -65,6 +65,11 @@ test_expect_success 'overlay status is full' '
 	test "$(flux overlay status --timeout=0 --summary)" = "full"
 '
 
+test_expect_success 'flux overlay errors prints nothing' '
```
grondo, line 68: Commit message typo: "ProbleM"

grondo (approved):
This LGTM! Just noticed one commit mesage typo.

Oh, also does `flux-overlay(1)` need to be updated?

t/t3303-system-healthcheck.t
garlick: Thanks - I forgot the man page.  Will get that done, thanks.

garlick: OK pushed that and the typo fix and I'll set MWP.

Issue #6114, opened 2024-07-18
rapid broker RSS growth and slow job-manager response on large system


grondo:
We caught another instance of rapid RSS growth (final RSS was >210GB) of the broker on elcap.

During this time, the flux-broker was using 400% CPU and I was able to capture a `perf` trace. Turns out ~23% of the broker's time was in the job-manager handling annotations. Another 20% was in the Fluxion qmanager sending annotations (we should open a corresponding issue in flux-sched)

```
   - ev_run                                                                    ▒
      - 22.69% ev_invoke_pending                                               ▒
         - 21.61% handle_cb                                                    ▒
            - 20.48% dispatch_message (inlined)                                ▒
               - 20.07% call_handler                                           ▒
                  - 19.91% alloc_response_cb                                   ▒
                     - 15.70% annotations_update_and_publish                   ▒
                        - 14.73% event_job_post_pack                           ▒
                           - 14.72% event_job_post_vpack                       ▒
                              - 13.10% event_job_post_deferred                 ▒
                                 - 11.74% event_job_process_entry              ▒
                                    - 10.51% journal_process_event             ▒
                                       - 9.77% flux_respond_pack               ▒
                                          + 9.76% flux_respond_vpack (inlined) ▒
```

and 

```
   - ev_run                                                                    ▒
      - 18.03% ev_invoke_pending                                               ▒
         - 14.91% qmanager_safe_cb_t::check_watcher_cb                         ▒
            - 14.71% qmanager_safe_cb_t::check_watcher_cb                      ▒
               - qmanager_cb_t::check_watcher_cb                               ▒
                  - 13.89% qmanager_cb_t::post_sched_loop                      ▒
                     + 12.24% schedutil_alloc_respond_annotate_pack            ▒
                     + 1.34% ?? (inlined)                                      ▒
```

It may be (partially verified by @trws and @milroy) that a bug in the qmanager, plus the new asynchronous code is resulting in many more annotation updates for jobs than before.

This doesn't fully explain the massive RSS growth though. However, it does give us a hint where to look. Even if the qmanager "bug" is fixed, we should still figure out why annotation updates are (possibly) chewing up massive amounts of RAM, perhaps we can do better here.

One experiment would be to just add a bit of code to sched-simple that sends annotation updates for all jobs in a check watcher (or even in a loop)

grondo:
Unsurprisingly job-list is also busy:
```
[...]

PR #6115, opened 2024-07-18
address RSS growth when there are excessive annotation events

garlick:
Progress towards making flux core more robust in situations like #6114.  Marking WIP because there's still some unexplained RSS growth with "annotation spam" to run down.

### DIFF
--- a/src/common/libschedutil/alloc.c
+++ b/src/common/libschedutil/alloc.c
@@ -38,8 +38,9 @@ static int schedutil_alloc_respond_pack (flux_t *h,
     if (fmt) {
         json_t *o;
         va_start (ap, fmt);
-        if (!(o = json_vpack_ex (NULL, 0, fmt, ap))
-            || json_object_update (payload, o) < 0) {
+        o = json_vpack_ex (NULL, 0, fmt, ap);
+        va_end (ap);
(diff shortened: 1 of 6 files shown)


grondo:
FYI - I ran 1024 jobs on this branch with heap profiling enabled. This was without the repro for #6114, so just one annotation per job. I captured the diff between before running any jobs, and after all jobs were inactive. It seems like most of this makes sense, but posting this here for your information (according to heap profile, 1024 jobs results in ~47M of heap, mostly from storing the job information in memory it appears)

[heap.pdf](https://github.com/user-attachments/files/16288880/heap.pdf)


Oh, and here's one after running `flux kvs dropcache` and `flux content dropcache` (24.6M used)

[heap.pdf](https://github.com/user-attachments/files/16288914/heap.pdf)

garlick:
I tacked on a commit that fixes a bug on an unlikely error path in libschedutil, but haven't had any other insights into the current heap behavior.  I'm dropping the WIP because I think this addresses the major problem and could go in without further ado.

grondo (approved):
Great work, LGTM!

garlick:
Thanks! Setting MWP.

PR #6120, opened 2024-07-19
add message queue counts to flux module list

garlick:
This just adds a count of the number of messages queued for each module to `flux module list` output.  `Send` is the number of messages sent by the broker but not yet read by the module.  `Recv` is the number of messages sent by the module and not yet read by the broker.  So this is from the broker's point of view.

Here I added a check watcher to `barrier.so` to have it continually send KVS lookups to `kvs.so` and immediately discard the future:
```
$ flux module list
Module                   Idle  S Send Recv Service
cron                       15  R    0    0 
resource                    8  R    0    0 
job-exec                    8  R    0    0 
job-list                   12  R    0    0 
sched-simple                8  R    0    0 sched
heartbeat                   0  R    0    0 
kvs-watch                  12  R    0    0 
connector-local             0  R    0    0 
kvs                         0  R 1160859    7 
barrier                     0  R    0 221241 
job-info                   12  R    0    0 
job-manager                 8  R    0    0 
content                     2  R    0    0 
job-ingest                  8  R    0    0 
content-sqlite              2  R    0    0 content-backing
```
The problem we recently faced with the escalating RSS that was due to scheduler spamming the job manager would have been immediately evident in this output.

### DIFF
--- a/src/broker/modhash.c
+++ b/src/broker/modhash.c
@@ -108,12 +108,14 @@ static json_t *modhash_entry_tojson (module_t *p,
 
     if (!(svcs  = service_list_byuuid (sw, module_get_uuid (p))))
         return NULL;
-    entry = json_pack ("{s:s s:s s:i s:i s:O}",
+    entry = json_pack ("{s:s s:s s:i s:i s:O s:i s:i}",
                        "name", module_get_name (p),
                        "path", module_get_path (p),
                        "idle", (int)(now - module_get_lastseen (p)),
                        "status", module_get_status (p),
-                       "services", svcs);
+                       "services", svcs,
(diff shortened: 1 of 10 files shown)


grondo (approved):
Great idea! LGTM

grondo:
One thought I had is to change the `Send` `Recv` header titles to `Sendq` `Recvq` to make it clear these are queued messages instead of send/recv counts. It would require 1 more character per column, though, so up to you.

garlick:
Thanks!  Yup I'll add the 'q'.  That would be clearer.

garlick:
Setting MWP - thank you for the review.

PR #6127, opened 2024-07-19
t: add initial ENOSPC tests

chu11:
This is just a "starting point" for future ENOSPC work.  

- have docker containers create a 1m tmpfs space for testing
- make sure that flux in general can run even if the tmpfs is full

as you see in the comments, the tests disable rc3.  I think the first initial work will be to somehow get rc3 to not hang.

### DIFF
--- a/src/test/docker/docker-run-checks.sh
+++ b/src/test/docker/docker-run-checks.sh
@@ -205,6 +205,7 @@ else
     docker run --rm \
         --workdir=$WORKDIR \
         --volume=$TOP:$WORKDIR \
+        --mount type=tmpfs,destination=/test/tmpfs-1m,tmpfs-size=1048576 \
         ${PLATFORM} \
         $MOUNT_HOME_ARGS \
         -e PLATFORM \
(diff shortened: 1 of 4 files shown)


chu11:
re-pushed, it occurred to me that perhaps these tests should be LONGTEST

chu11:
hmmmm, maybe should re-work these tests.  With LONGTEST they wouldn't normally run, and we would want this to run normally.  Perhaps could spin submitting jobs until hit out of space.  Shouldn't take more than a few jobs.

chu11:
re-pushed, this time running a single job at a time until the statedir ENOSPCes.  Under most circumstances shouldn't take more than 2-3 jobs.

garlick (approved):
This seems fine to me.

I was a little curious whether other types of failures might be possible (like if ENOSPC was the primary case but something in sqlite would propagate a different error).  I guess we'll see over time.

chu11:
i don't know why mergify isn't doing its thing on this PR.  re-based & re-pushed to hopefully kick it.

PR #6129, opened 2024-07-22
job-manager: ignore bad KVS eventlogs

garlick:
Problem: after a crash, sometimes job entries have to be manually removed from the KVS in recovery mode.  In particular, this error was seen: `job state (NEW) is invalid after replay`.
    
Don't abort the job manager when eventlog replay fails.  Ignore the job entry and log an error at LOG_ERR level.


grondo (approved):
Great! LGTM.

PR #6134, opened 2024-07-23
job-exec: fix job refcount leaks

grondo:
While working on something else, I found a couple places where the job-exec module leaks `struct jobinfo` refcount. This causes jobs to stay resident in the job-exec module even after they are inactive.

This won't show up as a leak with valgrind because the jobinfo objects are eventually destroyed when the job-exec module is unloaded.


garlick (approved):
Nice catch - LGTM!

grondo:
Thanks! I've set MWP.

PR #6138, opened 2024-07-23
job-exec: fix per-job `active_ranks` module stats

grondo:
`bulk_exec_active_ranks()` is supposed to report the ids of ranks on which bulk-exec subprocesses are still active, but it doesn't account for those that have exited. This causes the `active_ranks` key of `flux module stats job-exec` per-job stats to be incorrect (it always reports all ranks).

Fix `bulk_exec_active_ranks()` and add a test.


garlick (approved):
LGTM!

grondo:
Thanks! I'll set MWP.

PR #6139, opened 2024-07-24
move bad KVS job entries to a lost+found directory

garlick:
Problem: when KVS job entries cannot be replayed by the job manager, they are ignored by flux but left in the KVS.

Move bad job directories to a KVS lost+found.jobs directory for manual cleanup/analysis.


grondo (approved):
Nice! LGTM!

garlick:
Thanks!  Setting MWP.

PR #6140, opened 2024-07-24
job-exec: add timeout for job start barrier

### LINKED ISSUE #6103: job-exec: add start timeout?
Problem: We've been regularly seeing jobs hang at startup due to one or more nodes in some variant of a bad state where the job shell is potentially hung while starting. This causes the job to never even reach the `start` event, but the job just  sits there using resources until a timeout. It is usually quite a chore to find the affected ranks, which then require a reboot to clear the job.

Possibly a timeout should be added to the job-exec module while it is trying to start the job. Maybe if all job shells do not reach the first barrier before a configurable timeout, the missing ranks could be drained and the job terminated early.

grondo:
This PR adds a timeout for the execution system "start barrier". 

All job shells participate in a barrier after initialization, which includes change of working directory, processing the initrc, etc. After this PR, a timer (default 30m) will be started after the first job shell enters the barrier. If the timer expires before the barrier fully completes, a job exception of the form:
```
exception type="exec" severity=0 userid=1001 note="start barrier timeout waiting for 1/4 nodes (rank 3)
```
will be raised, and the affected rank or ranks drained with the message:
```
job f3oxE4XH start timeout: possible node hang
```
The default barrier timeout can be overridden in jobspec (mainly for testing) via `attributes.system.exec.bulkexec.barrier-timeout`, or via config `exec.barrier-timeout`.

Fixes #6103

### DIFF
--- a/doc/man5/flux-config-exec.rst
+++ b/doc/man5/flux-config-exec.rst
@@ -50,6 +50,16 @@ kill-signal
    (optional) Specify an alternate signal to ``SIGKILL`` when killing tasks
    and the job shell. Mainly used for testing.
 
+barrier-timeout
+   (optional) Specify the default job shell start barrier timeout in FSD.
+   All multi-node jobs enter a barrier at startup once the Flux job shell
+   completes initialization tasks such as changing the working directory
+   and processing the initrc file. Once the first node enters this barrier,
+   the job execution system starts a timer, and if the timer expires
+   before the barrier is complete, raises a job exception and drains the
+   nodes on which the barrier is waiting.  To disable the barrier timeout,
+   set this value to ``"0"``. (Default: ``30m``).
+
 testexec
    (options) A table of keys (see :ref:`testexec`) for configuring the
    **job-exec** test execution implementation (used in mainly for testing).
--- a/src/modules/job-exec/exec.c
+++ b/src/modules/job-exec/exec.c
@@ -118,7 +118,7 @@ static struct exec_ctx *exec_ctx_create (struct jobinfo *job,
 {
     struct exec_ctx *ctx = calloc (1, sizeof (*ctx));
     flux_reactor_t *r = flux_get_reactor (job->h);
-    double barrier_timeout = 1800.;
+    double barrier_timeout = config_get_default_barrier_timeout ();
 
     if (!r || !ctx || !(ctx->barrier_pending_ranks = idset_copy (ranks)))
         goto error;
--- a/src/modules/job-exec/exec.c
+++ b/src/modules/job-exec/exec.c
@@ -24,6 +24,8 @@
  *                               "init", or "starting"
  *    "service":s              - Specify service to use for launching remote
  *                               subprocesses: "rexec" or "sdexec".
+ *    "barrier-timeout":F      - Specify timeout for start barrier in floating
+ *                               point seconds.
  * }
  *
  */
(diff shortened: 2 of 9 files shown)


src/modules/job-exec/exec.c
```diff
@@ -56,32 +58,43 @@ struct exec_ctx {
...
     (void) json_unpack (jobspec,
-                        "{s:{s:{s:{s:{s:s}}}}}",
+                        "{s:{s:{s:{s:{s?s s?F}}}}}",
                         "attributes",
                           "system",
                             "exec",
                               "bulkexec",
-                                "mock_exception", &ctx->mock_exception);
+                                "mock_exception", &ctx->mock_exception,
+                                "barrier-timeout", &ctx->barrier_timeout);
```
garlick: Suggestion: make everything optional, add `!` at the end of the `bulkexec` object to indicate extra keys are not allowed, and then check the return value of `json_unpack()` and fail on error.  That way config typos in test scripts (misspelled key, wrong type) will be a more obvious failure.

That's not an issue with this PR though - feel free to ignore.

src/modules/job-exec/job-exec.c
```diff
@@ -374,28 +374,31 @@ static void jobinfo_complete (struct jobinfo *job, const struct idset *ranks)
...
+                             "resource.drain",
+                             0,
+                             0,
+                             "{s:s s:s s:s}",
+                             "targets", ranks,
+                             "reason", reason,
+                             "mode", "update"))) {
         flux_log (job->h,
                   LOG_ERR,
                   "Failed to drain broker ranks %s for job %s",
```
garlick, line 404: Suggestion (minor): change `ranks` back to a `struct idset *` since both callers both have to do the encode otherwise, and then since the two possible failures `idset_encode()` and `flux_rpc_pack()` are both unlikely and we are not checking the RPC response, just consolidate both under a `failed to send drain request for broker ranks...` log.  Also, use `flux_log_error()` to capture the errno.

src/modules/job-exec/exec.c
```diff
@@ -107,17 +120,87 @@ static void complete_cb (struct bulk_exec *exec, void *arg)
...
+    if (!(ctx->shell_barrier_timer = flux_timer_watcher_create (r,
+                                                                timeout,
+                                                                0.,
+                                                                start_timer_cb,
+                                                                ctx)))
+        flux_log_error (ctx->job->h,
+                        "%s: failed to create barrier start timer",
+                        idf58 (ctx->job->id));
+    else
+        flux_watcher_start (ctx->shell_barrier_timer);
```
garlick, line 186: It might be better to create the watcher when the exec_ctx is created and just not start it, so this error handling wouldn't be required in a place where there's nothing to do but log it.

doc/man5/flux-config-exec.rst
```diff
@@ -50,6 +50,16 @@ kill-signal
...
 
+barrier-timeout
+   (optional) Specify the default start barrier timeout in FSD for all
+   jobs. All multi-node jobs enter a barrier at startup once initialization
+   on each node is complete and individual tasks have been launched. Once
+   the first node enters this barrier, the job execution system starts a
+   timer, and if the timer expires before the barrier is complete, then
+   a job exception is raised warning that some nodes did not reach the
+   barrier and may be hung. To disable the barrier timeout, set this value
+   to ``"0"``. (Default: ``30m``).
```
garlick, line 61: It might be good to specifically call out the "flux shell barrier" to avoid confusion with the PMI barrier that comes up so often when troubleshooting MPI startup.

garlick (approved):
Great!  I just had a couple of suggestions which are pretty minor!

src/modules/job-exec/exec.c
grondo: Good suggestion! I'll tack on a couple commits to this PR.

src/modules/job-exec/job-exec.c
grondo: What you suggest was how it was originally implemented, but at least in the start timeout case, the caller needs to do the encode anyway in order to create the job exception message, so it seemed wasteful to encode twice in the same call path since `drain_ranks()` doesn't _need_ a `struct idset`... 

If you feel strongly though I can change it back.

src/modules/job-exec/exec.c
grondo: Good suggestion, though this breaks detection of the first barrier elsewhere (also not sure it should be _fatal_ if we can't start the timer anyway, in the common case the job will still proceed normally).

I think there's another variable I can use to detect the first barrier so let me see if I can rework this.

garlick: In case I wasn't clear I meant create the watcher up front so there would be no error handling in this function.  Then all the creation functions with unlikely failure (e.g. ENOMEM) can be grouped together with a common error path.

grondo: Yeah, that was clear. My point was that currently the code uses `timer_watcher != NULL` to indicate other barriers than the first, and skips starting the timer in these cases.

Also, that maybe failure to create the barrier timer should not be fatal to the job as implied. However, as you say, the failure is unlikely (so I guess that's why I thought it was ok to do it outside of the constructor), so I'll attempt to move it. It has a cascading effect though so this will take awhile.

garlick: Eh if you feel it's better the way it is, I'm fine with that.  It was just a suggestion.

doc/man5/flux-config-exec.rst
grondo: Good idea. I have rewritten this section (there was also an error, the init barrier occurs before tasks are started, not after)

src/modules/job-exec/exec.c
grondo: No, I think you're right that all the constructors should occur up front, if only just so they are easy to find. The watcher will be created unnecessarily for single-node jobs, but that is not a big deal.

grondo:
I've pushed a new version of this PR that makes the changes suggested above. The update to report errors in the `json_unpack()` of the `bulkexec` test arguments affected one test that took advantage of silent failure, so that test is also fixed here.

garlick:
Looks great!

grondo:
Thanks! Setting MWP.

PR #6142, opened 2024-07-25
cmd: accept `-q` for `--queue` in `flux-jobs`, `pgrep`, and `pkill`

### LINKED ISSUE #6130: Suggestion: standardize command line options for specifying queue
I'm not clear if this trips anyone else up - if we take `flux resource drain/undrain/info/list/status` as the standard for queue specification:
`  -q, --queue=QUEUE,...   Include only specified queues in output`

Then these commands do not match:
```
flux jobs
flux pgrep
flux pkill
```
Also `flux hostlist` overloads `-q` for `quiet` (which admittedly is pretty standard outside of flux..).

grondo:
Problem: Use of `-q` as a shorthand for `--queue` is standard in most flux(1) subcommands, such as `flux resource list`, `flux submit`, etc. However, `-q` is not accepted in `flux jobs`, `flux pgrep` and `flux pkill`, which is awkward for users.

Accept `-q` in these commands. Update `flux-jobs(1)` (`flux-pgrep(1)` already claimed `-q` was supported).

Fixes #6130

### DIFF
--- a/doc/man1/flux-jobs.rst
+++ b/doc/man1/flux-jobs.rst
@@ -46,7 +46,7 @@ OPTIONS
 
    List jobs with a specific job name.
 
-.. option:: --queue=[QUEUE]
+.. option:: -q, --queue=[QUEUE]
 
    List jobs in a specific queue.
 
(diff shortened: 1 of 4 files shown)


chu11 (approved):
LGTM, although do we need to update some bash completions too?

grondo:
Yes, good catch. Let me do that.

grondo:
Ok, updated the bash completions and now I'll set MWP. Thanks!

PR #6156, opened 2024-07-29
matrix: fix matrix generation for arm64 and remove redundant fedora

trws:
problem: the no-args fedora40 and version with just prefix arguments are identical for all intents and purposes

solution: remove the no-args version in favor of the one that generates a docker tag

@grondo, sorry for the churn here, but I only realized this was essentially a duplicate when I went to port the changes to sched.  Should be a quick cleanup if you agree.  Sadly will require updating the branch protections again.

Also found that the explicit platform we had on bookworm needs to be on all the multi-arch images, so that's in here now.  Finally dropped the non-ascii el8 builder in favor of bringing the base bookworm/jammy builders back since they were not intentionally dropped.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -254,7 +254,7 @@ jobs:
         echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
         docker manifest create fluxrm/flux-core:bookworm fluxrm/flux-core:bookworm-amd64 fluxrm/flux-core:bookworm-386 fluxrm/flux-core:bookworm-arm64
         docker manifest push fluxrm/flux-core:bookworm
-        for d in el9 noble fedora40 ; do
+        for d in el9 noble fedora40 alpine ; do
           docker manifest create fluxrm/flux-core:$d fluxrm/flux-core:$d-amd64 fluxrm/flux-core:$d-arm64
           docker manifest push fluxrm/flux-core:$d
         done
--- a/src/test/generate-matrix.py
+++ b/src/test/generate-matrix.py
@@ -146,9 +146,6 @@ class BuildMatrix:
 
 matrix = BuildMatrix()
 
-# Fedora40: no args
-matrix.add_build(name="fedora40")
-
 # Debian: 32b
 matrix.add_build(
     name="bookworm - 32 bit",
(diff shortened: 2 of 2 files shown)


trws:
Turns out we were also missing the arm builds due to a format string error in the matrix generation, which caused the downstream sched container pulls to fail.  Turns out sched is a nice smoke test for actually getting images out of core.

grondo (commented):
We need at least one build without `--without-flux-security` to make sure building this way continues to work. Just making sure that's still the case.

trws:
Yup, that's the `el8 - ascii` build at the moment, because it doesn't generate a docker container it doesn't make for an inconsistent end product but still gives us a test.  Also the bookworm 32 bit test runs that way. I have a feeling that test is not super long for this world now that even the BSDs are abandoning 32-bit, but it works so... 🤷 

I ended up doing one last refactor here because it's clearly too easy to break multiarch builds as it sits.  Basically just added a method that actually sets up a _set_ instead of just one.

The end result is we have more builders now on master but less on a normal push/PR, but I think with very similar coverage.  Here's a diff of our builds before the other PR and after this one, including those that only build on master or tag:

left is pre-rework, right is after this PR, renamed alpine to alpine test-install and fedora40 - gcc14 to fedora40 - test-install for consistency:

```diff
1,2c1,2
< "alpine"
< "bookworm"
---
> "alpine - arm64"
> "alpine - test-install"
6d5
< "bookworm - py3.11,clang-15"
9d7
< "el8"
12,14c10,14
< "fedora34 - gcc-11.2,py3.9"
< "fedora38 - gcc-13.1,py3.11"
< "fedora39 - gcc-13.2,py3.12"
---
> "el9 - arm64"
> "el9 - test-install"
> "fedora40 - arm64"
> "fedora40 - clang-18"
> "fedora40 - test-install"
17a18,19
> "noble - arm64"
> "noble - test-install"
```

grondo:
When I try to run `docker-run-checks.sh` on my x86_64 system now, I'm getting:
```
3.912 Fatal glibc error: CPU does not support x86-64-v2
```
Does this PR also fix this error (I assume I'm somehow getting the wrong container image?)

trws:
@grondo I'm going to look into this.  My understanding is that the el9 image, that's just how it is.  There is no el9 for non-v2 that I know of.  I'm not sure about the others though, will come back with info.

grondo:
Thanks, other images are indeed working on my system.

Apologies if this is a naive question, but does that mean we simply cannot run the el9 image on older hardware, or is my system misconfigured in some way?

trws:
This is a kinda new thing for the x86 world, the original x86_64 switch bought us a lot of time because it baked in so many older x86 add-ons, but now building to basic x86_64 leaves a _lot_ of performance on the table, so somebody came up with these portable "levels" of x86.  Currently I think it's just v1-v3, roughly original x86_64 at v1 to haswell+ at v3, so avx and so-forth required.  There's no misconfiguration on your system, it's just that red-hat and some of the others are deeming original x86_64 insufficient at this point, kinda like when most of the distros switched from 386 to 686 back in the day.  You could still run it under qemu, but of course with a non-negligible performance penalty.

trws:
There's a good rundown on wikipedia now [here under microarchitecture levels](https://en.wikipedia.org/wiki/X86-64), v2 covers everything since roughly 2011 except maybe some atom parts.  My personal laptop is only just a hair over that line now that I think about it.

grondo:
Ah, my problem is likely that I'm running inside a VM. The CPU is "Common KVM processor" so I'm guessing I just need to change options or update the host. (Host cpu reports as x86_64-v3)

grondo (approved):
Nice improvement, thanks!

trws:
Ohhhh that would do it.  I forgot but I had that issue with VMs on my home desktop until I went through and did all the extras to get nested virtualization and full hardware virtualization set up.  You probably don't need to do all of that, but it's so convenient it might be worth it.

PR #6157, opened 2024-07-29
Adds a new --bank/-B flag to Flux scheduling commands to simplify setting a bank

ilumsden:
Currently, when Flux is running in system mode, users have to use `--setattr=system.bank=<BANK_NAME>` to select a bank. This PR adds a new `--bank`/`-B` flag to the following Flux commands to simplify this:
* `flux alloc`
* `flux batch`
* `flux run`
* `flux submit`
* `flux bulksubmit`

This new flag is essentially synonymous to Slurm's `--account`/`-A` flag.

There are still three outstanding things to do in this PR:
1. Test it
2. Decide what to do in situations where flux-accounting does not exist
3. Correct the commit message format since I wasn't really sure what the correct prefix is for the part of the code I'm editing

### DIFF
--- a/src/bindings/python/flux/cli/base.py
+++ b/src/bindings/python/flux/cli/base.py
@@ -758,6 +758,13 @@ class MiniCmd:
             formatter_class=flux.util.help_formatter(),
         )
         parser.add_argument(
+            "-B",
+            "--bank",
+            type=str,
+            metavar="BANK",
+            help="Submit a job to a specific named bank",
+        )
+        parser.add_argument(
             "-q",
             "--queue",
             type=str,
@@ -1041,6 +1048,9 @@ class MiniCmd:
 
         if args.queue is not None:
             jobspec.setattr("system.queue", args.queue)
+
+        if args.bank is not None:
+            jobspec.setattr("system.bank", args.bank)
 
         if args.setattr is not None:
             for keyval in args.setattr:
@@ -1425,7 +1435,6 @@ class SubmitBulkCmd(SubmitBaseCmd):
             args.progress = None
 
     def watcher_start(self, args):
-
         if not self.watcher:
             #  Need to open self.flux_handle if it isn't already in order
             #  to start the watcher


garlick:
Thanks Ian!

On the commit form:  the title should be <= 50 chars written in the imperative voice.  The prefix is optional.  A body is also required and we prefer the problem/solution form.  For example here it could be something like:
```
add -B,--bank option to flux job submission commands

Problem: to specify the flux-accounting bank at job submission time, users must remember
--setattr=system.bank=NAME.

Add a -B,--bank=NAME to make it easier on users.
```

The fact that either the long form or the short form are ignored when flux-accounting is not present is maybe a different problem that doesn't necessarily have to be solved at the same time.  It's really one of a class of problems that we have.  For example, you can accidentally specify `-ompi=pmi2` (like slurm wants) instead of `-opmi=pmi2` (like flux wants) and flux will just ignore you.

For discussion, I guess my one thought about the proposed change is since these are equivalent:
```
-o KEY=VALUE
--setopt system.shell.options.KEY=VALUE
```
would making adding a short option for setting system options be a reasonable compromise that keeps flux-core agnostic the bank concept?
```
-S bank=NAME
--setopt system.bank=NAME
```

If options are changed, we'll need a man page update also (in a separate commit preferably) and probably a couple tests just to get coverage for the option parsing.

grondo:
I think what we'd planned here was to have a way for sites to add new options to the submission commands via plugins, then flux-accounting could provide a plugin to add `--bank`.

@trws suggested that all plugin-enabled options should have a specific prefix or other way to make it obvious that the option is provided by a plugin. This is a good idea, but might nullify some of the convenience of adding new options that are just shorthand for `--setattr=KEY=VAL` (or as Jim suggests, `-S KEY=VAL`.), and adding new short options (e.g. `-B`) would be right out.

Maybe if we could add a short option for `--setattr` as suggested by @garlick, _and_ add a way to easily auto-document supported attributes in `flux run/submit/etc. --help` or similar, then this could go a long way to fixing this particular problem. E.g. a new section in the `--help` output that would list supported attributes with a way to extend this list by sub-projects, plugins or sites?

grondo:
@trws brings up some really great points. Currently we have maximum flexibility in that other framework projects, shell plugins, jobtap plugins, etc. can easily support new attributes in jobspec without any prior support in flux-core. However, the drawback is, as you say, there is currently no way to detect typos or malformed attributes until it is too late. Additionally, there is no auto-documentation of these attributes, instead the documentation, if any, is spread throughout the implementation(s). I'll open an issue on this.

For this specific PR/issue, I wonder if we should just officially support `bank` and `project` attributes in [RFC 14](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_14.html). We already do support `bank` and `project` in job-list anyway: 

https://github.com/flux-framework/flux-core/blob/14c64d2f132dcf75d44e77bcbd35d56066f5e14a/src/modules/job-list/job_data.c#L191-L199

So the ship on those two attributes has already sailed...

grondo:
This commit was merged in #6195. Closing. Thanks @ilumsden!

PR #6163, opened 2024-07-31
doc: update flux admin guide URL

kkier:
Updates path to reflect current URL - https://flux-framework.readthedocs.io/projects/flux-core/en/latest/guide/admin.html

### DIFF
--- a/doc/guide/broker.rst
+++ b/doc/guide/broker.rst
@@ -383,7 +383,7 @@ The TBON has a fixed topology determined by configuration, and brokers do not
 adapt to route around down nodes.  In addition, Flux must be completely stopped
 to alter the topology configuration - it cannot be changed on the fly.
 
-See the `Flux Administrator's Guide <https://flux-framework.readthedocs.io/en/latest/guides/admin-guide.html>`_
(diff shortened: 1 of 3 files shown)


garlick:
Whoops!  Thanks @kkier!

If you have a moment, it would probably be better to squash these into one commit.  Commits should have a title (use imperative voice) and a body, and we like the problem/solution form.  How about something like this:

```
doc: update flux admin guide URL

Problem: the flux admin guide path has changed but some docs have not been updated.

Update path.
```

kkier:
Whoops, forgot to push after squashing...  Updated as above.

garlick (approved):
Looks good to me, thanks!

Usually for the problem/solution commit message form we state the problem, skip a space and then state the solution.  However in this case it's pretty obvious.  Next time.

Now If you want to set the merge-when-passing tag on the sidebar, mergify will rebase this if necessary and merge.  Or I can do it for you - let me know.

grondo:
I think only project admins can set labels on PRs. (I could be wrong though)

garlick:
Right, I'll do that then.  Thanks @kkier!

PR #6170, opened 2024-08-01
cmd/startlog: initialize enum to non zero

chu11:
Problem: The enum POST_FLAG_FLUSH was not initialized, therefore it defaults to 0.  This means when a user passes a flag of "0", it ends up being equal to POST_FLAG_FLUSH, leading to undesired behavior.

Initialize POST_FLAG_FLUSH to 1.


garlick (approved):
DOH!  Nice catch.

PR #6180, opened 2024-08-02
flux-resource: fix missing queue in `flux resource list` output for states with no nodes

### LINKED ISSUE #6166: flux resource list doesn't display queue name when there are 0 resources in a state
`flux resource list` appears to list all possible states, but if there are no nodes in a given state, it doesn't list a queue name (or node list) for that state. At least, I think that's the issue. This can be a bit confusing. I would argue that it would be better to just not list the state if there are no nodes in that state, or to list all of the queues to show that there are 0 nodes in that state in any queues.

We've seen this in a couple of places. On corona, when there happened to be no jobs running:
```
corona212{gpt}40: flux resource list -o '{queue} {nnodes} {state} {nodelist}'
QUEUE NNODES STATE NODELIST
pdebug 15 free corona[171,173-186]
pbatch 84 free corona[188-194,196-207,213-214,219,221-250,252-253,255-259,261-269,272-275,277-280,282-285,287-290]
pnvmeof 5 free corona[292-296]
 0 allocated 
pdebug 1 down corona172
pbatch 15 down corona[187,195,215-218,220,251,254,270-271,276,281,286,291]
```

and on rzvernal when there happened to be no nodes down:
```
[day36@rzvernal10:src]$ flux resource list -o '{queue} {nnodes} {state} {nodelist}'
QUEUE NNODES STATE NODELIST
pdebug 23 free rzvernal[15,19,22,25-30,32-45]
mi300a 4 free rzvernal[48-51]
pdebug 11 allocated rzvernal[12-14,16-18,20-21,23-24,31]
 0 down 
[day36@rzvernal10:src]$
```

grondo:
This PR fixes #6166 by appending an empty resource set for each configured or requested queue, instead of a pure empty resource set, which can have no queue. Thus, when node nodes are in a given state, `{queue}` will not be blank, instead there will be a line in output for each queue.

### DIFF
--- a/t/t2350-resource-list.t
+++ b/t/t2350-resource-list.t
@@ -541,6 +541,32 @@ test_expect_success 'flux resource list --queue works for a queue with no constr
 	test $(grep "allocated 1" queue-every.out | grep -c debug) -eq 1 &&
 	test $(grep "allocated 1" queue-every.out | grep -c every) -eq 2
 '
+test_expect_success 'flux resource list includes queue names for empty sets' '
+	# There are no nodes in 'down' state in this test, so use that:
+	flux resource list -s down -no "{queue} {state} {nnodes}" \
+		>queue-empty1.out &&
+	cat <<-EOF >queue-empty1.expected &&
+	every down 0
+	batch down 0
+	debug down 0
+	EOF
+	test_cmp queue-empty1.expected queue-empty1.out
+'
+test_expect_success 'flux resource list includes queue names for empty sets (single)' '
+	flux resource list -q batch -s down -no "{queue} {state} {nnodes}" \
+		>queue-batch-down.out &&
+	cat <<-EOF >queue-batch-down.expected &&
+	batch down 0
+	EOF
+	test_cmp queue-batch-down.expected queue-batch-down.out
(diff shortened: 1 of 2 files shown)


t/t2350-resource-list.t
```diff
@@ -541,6 +541,32 @@ test_expect_success 'flux resource list --queue works for a queue with no constr
...
+	flux resource list -s down -no "{queue} {state} {nnodes}" \
+		>queue-empty1.out &&
+	cat <<-EOF >queue-empty1.expected &&
+	every down 0
+	batch down 0
+	debug down 0
+	EOF
+	test_cmp queue-empty1.expected queue-empty1.out
+'
+test_expect_success 'flux resource list includes queue names for empty sets' '
```
chu11, line 555: i think test name same as the prior one, differentiate it in some way?  and maybe tweak the following one depending on how you change it.

chu11 (commented):
LGTM, just the nit on the test message name

chu11 (approved):

grondo:
Thanks! Fixed that (denoted the first test as `(single)`)

PR #6181, opened 2024-08-03
actions: increase ci timeout from 90 to 120 minutes

trws:
problem: some of the arm builders, especially the gcc-14 ones, run right up to the timeout of 90 minutes when succeeding, giving spurious failures to update all the container manifests

solution: increase the timeout to 120 minutes


grondo (approved):
LGTM

trws:
Dang, looks like we might still need the coverage env var, had another coverage timeout.

Issue #6145, opened 2024-07-26
flux-queue: consider combining "status" and "list"


ofaaland:
As a user, it seems like these two commands could be combined, and like that would be a more obvious user interface.  For example:

```
$flux queue list
QUEUE    DEFAULTTIME  TIMELIMIT     NNODES     NCORES      NGPUS   ENABLED   SCHEDULING
gateway          30m         1d      0-inf      0-inf      0-inf     yes        no
rabbit           30m         1d      0-inf      0-inf      0-inf     yes       yes
compute*         30m         5d      0-inf      0-inf      0-inf      no        no
```

In case it helps, some anecdotes:
I personally didn't notice there was a `flux queue status` command until today.  Part of the reason for that was that I did find `flux queue list`, and it didn't occur to me there would be a separate status command.

More generally, I think it would help to make it more visible, whether a queue is enabled and scheduling is started or not.

grondo:
This is a pretty good idea. The reason `flux queue list` and `flux queue status` are different commands is because they query different subsystems (`flux queue list` queries configuration and `flux queue status` queries the job manager).

However, similar to `flux resource status` and `flux resource list`, information from the two subsystems can easily be combined here...

PR #6185, opened 2024-08-04
enhance `flux queue list` with `-q, --queue=` option and abillity to report queue enabled/started status

grondo:
As noted in #6145, it is confusing to users why there are separate `flux-queue(1)` `status` and `list` subcommands.

This PR enhances `flux queue list` to fetch queue status in addition to configuration, and uses this data to support the following new output fields:

 * `{submission}` - text representation of submission status: `enabled` or `disabled`
 * `{scheduling}` - text representation of scheduling status: `started` or `stopped`
 * `{enabled}` - single character boolean for enabled/disabled status (`✔/y` or `✗/n`)
 * `{started}` - single character boolean for started/stopped status  (`✔/y` or `✗/n`)
 
 The default output format is now:
```
                "?:{queuem:<8.8} "
                "{color_enabled}{enabled:>2}{color_off} "
                "{color_started}{started:>2}{color_off} "
                "{defaults.timelimit!F:>8} "
                "{limits.timelimit!F:>8} "
                "{limits.range.nnodes:>10} "
                "{limits.range.ncores:>10} "
                "{limits.range.ngpus:>10}"
```

So that the enabled/started status is easily visible by default with `flux queue list`, e.g.:

```
QUEUE    EN ST TDEFAULT   TLIMIT     NNODES     NCORES      NGPUS
pbatch*   ✔  ✔      30m       1d      0-inf      0-inf      0-inf
pdebug    ✔  ✔      30m       1h      0-inf      0-inf      0-inf
pnvmeof   ✔  ✔      30m       1d      0-inf      0-inf      0-inf
```

Additionally, a `-q, --queue=` option is added to `flux queue list`. Now the submission status (for example) of a queue can be obtained via something like:

```console
$ flux queue list --queue=pbatch -no {submission}
enabled
```
field names were chosen somewhat arbitrarily, so up for discussion.

### DIFF
--- a/doc/man1/flux-queue.rst
+++ b/doc/man1/flux-queue.rst
@@ -42,7 +42,11 @@ list
 
 .. program:: flux queue list
 
-List queue defaults and limits.
+List queue status, defaults, and limits.
+
+.. option:: -q, --queue=QUEUE,...
+
+   Limit output to specified queues
 
 .. option:: -n, --no-header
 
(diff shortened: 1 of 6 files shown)


grondo:
Yeah, a few of those were valid. Fixed and forced a push.

grondo:
Since this is approved, I'll set MWP. We can always adjust field names/headers if there is user confusion.

Issue #6136, opened 2024-07-23
resource: improve error message when excluded rank doesn't appear in R or `resource.config`


grondo:
With a resource section like the following:
```toml
[resource]
exclude = "host[1,1001-1002]"
norestrict = true

[[resource.config]]
hosts = "host[1001-1002]"
cores = "0-95"

[[resource.config]]
hosts = "host[1003-1100]"
cores = "0-95"
gpus = "0-3"

[[resource.config]]
hosts = "host[1003-1100]"
properties = ["pbatch"]
```

Flux failed to start with only the following error as a clue:
```
broker.err[0]: rc1.0: flux-module: load resource: Invalid argument
```

Probably the error is that `host1` is not in `resource.config` but is excluded? The error from the instance should be improved.

grondo:
It looks like this error condition is indeed logged, possibly the error was just missed. Closing.
```
Aug 05 14:07:52.701850 UTC resource.err[0]: error decoding exclude set foo: invalid targets: invalid hosts: foo: Invalid argument
Aug 05 14:07:52.702300 UTC resource.crit[0]: module exiting abnormally
Aug 05 14:07:52.703833 UTC broker.err[0]: rc1.0: flux-module: load resource: Invalid argument
Aug 05 14:07:52.704978 UTC broker.err[0]: rc1.0: /home/grondo/git/flux-core.git/etc/rc1 Exited (rc=1) 0.7s
```

PR #6186, opened 2024-08-05
resource: error message cleanup, add test for invalid exclude

grondo:
While looking at #6136 I noticed that `flux_log_error()` was used unnecessarily for the exclude idset error message.
This PR fixes that.

Also, thought I'd add a test for the specific issue of excluding a hostname that isn't in configuration.


grondo:
Thanks, I'll set MWP.

PR #6187, opened 2024-08-05
testsuite: fix racy test in t2406-job-exec-cleanup.t

### LINKED ISSUE #6184: t2406-job-exec-cleanup Racy sub-test
Spurious test failure due to job not getting killed, or so it seems. Get a null for kill_timeout. 

job-exec: ensure kill_shell_count > 1
  
  expecting success: 
  	flux module stats job-exec | jq .jobs.${jobid}.kill_timeout &&
  	flux module stats job-exec | jq -e ".jobs.${jobid}.kill_timeout > 0.1"
  
  null
  false
  not ok 12 - job-exec: kill-timeout > original value (0.1)

grondo:
Problem: The kill-timeout test races with max-kill-count in t2406-job-exec-cleanup.t. That is, in some cases, max-kill-count (which is forced to 3 attempts in the test) is triggered before the test checks that the kill-timeout value has been increased by the job-exec module, leading to a spurious test failure.

Split up the tests for kill-timeout and max-kill-count. When testing kill-timeout, set max-kill-count to a large number so it is unlikely to be reached. Then separately test max-kill-count after reloading the job-exec module with max-kill-count=3.

Fixes #6184

### DIFF
--- a/t/t2406-job-exec-cleanup.t
+++ b/t/t2406-job-exec-cleanup.t
@@ -35,19 +35,18 @@ test_expect_success 'job-exec: ensure cancellation kills job' '
 	test_expect_code 137 flux job status $id &&
 	test_must_fail ps -q $pid
 '
+# Note: increase max-kill-count here to ensure job doesn't disappear from
+# job-exec while we're testing kill-timeout:
 test_expect_success 'job-exec: reload module with kill/term-signal=SIGURG' '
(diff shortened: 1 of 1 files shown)


chu11 (approved):
Good catch.  LGTM!

grondo:
Thanks!

PR #6188, opened 2024-08-06
docker: Demote fedora40 to x86_64 only

trws:
After all the reworks, I just flat can't get flux to build on fedora40 under arm64 emulation in a reasonable amount of time.  No clue why, noble should have the same issue (or worse) and it's done in only a bit over half the time.  So for now I'm throwing in the towel. Will revisit when GHA supports arm runners for OSS.


trws:
One upside here, no need to change any of the checks.  This just removes a post-merge build and the associated manifest creation.

grondo (approved):
Ugh, sorry for all the work! LGTM!

Issue #6151, opened 2024-07-29
flux-broker: overlay_create: Cannot allocate memory 


wadudmiah:
Hi,

I am trying to start flux after allocating 2 nodes:
```
$ salloc -N 2 --exclusive 
```
And then I try to start flux using the command but I get the error `flux-broker: overlay_create: Cannot allocate memory`:
```
$ srun -N2 -n2 --pty --account=training flux start
srun: job 1416393 queued and waiting for resources
srun: job 1416393 has been allocated resources
flux-broker: overlay_create: Cannot allocate memory
srun: error: cn-109-2: task 0: Exited with exit code 1
srun: error: cn-109-3: task 1: Exited with exit code 1
```
Can anyone please assist?

thanks in advance!

garlick:
Hi - It's an odd place to fail, or at least one I don't recall seeing before.

Do you get that error if you try to start flux on one node?  What about standalone, e.g. just `flux start --test-size=2`?

Also could you add the output of `flux version`?  (thanks)

wadudmiah:
Hi @garlick, 

please see the output below:

```
$ flux start --test-size=2
flux-broker: overlay_create: Cannot allocate memory
flux-broker: overlay_create: Cannot allocate memory
flux-start: 1 (pid 22049) exited with rc=1
flux-start: 0 (pid 22048) exited with rc=1
$ flux version
commands:               0.64.0-20-g33f9116cd
libflux-core:           0.64.0-20-g33f9116cd
build-options:          +hwloc==2.1.0+zmq==4.3.5
```

wadudmiah:
Hi @garlick, 

please see the output below:

```
$ cat /etc/os-release
NAME="Rocky Linux"
VERSION="8.6 (Green Obsidian)"
ID="rocky"
ID_LIKE="rhel centos fedora"
VERSION_ID="8.6"
PLATFORM_ID="platform:el8"
PRETTY_NAME="Rocky Linux 8.6 (Green Obsidian)"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:rocky:rocky:8:GA"
HOME_URL="https://rockylinux.org/"
BUG_REPORT_URL="https://bugs.rockylinux.org/"
ROCKY_SUPPORT_PRODUCT="Rocky Linux"
ROCKY_SUPPORT_PRODUCT_VERSION="8"
REDHAT_SUPPORT_PRODUCT="Rocky Linux"
[...]

PR #6189, opened 2024-08-06
add configure time check for zeromq crypto and improve broker error handling at cert creation

garlick:
Problem:  as noted in #6151, the error handling of `cert_create()` is wrong, and we lack a configure time test that zeromq was built with curve support.

This addresses both of those issues.

Leaving this a WIP for now since this purports to "fix" #6151 but we have not confirmed that this is the whole problem there.

### DIFF
--- a/config/x_ac_zeromq.m4
+++ b/config/x_ac_zeromq.m4
@@ -1,0 +1,8 @@
+# N.B oldest in CI are focal=libzmq-4.3.2-2, centos7=zeromq-4.1.4
+
+AC_DEFUN([X_AC_ZEROMQ], [
+
+    PKG_CHECK_MODULES([ZMQ], [libzmq >= 4.0.4])
+
+  ]
+)
--- a/config/x_ac_zeromq.m4
+++ b/config/x_ac_zeromq.m4
@@ -4,5 +4,32 @@ AC_DEFUN([X_AC_ZEROMQ], [
 
     PKG_CHECK_MODULES([ZMQ], [libzmq >= 4.0.4])
 
+    old_CFLAGS=$CFLAGS
+    CFLAGS="$CFLAGS $ZMQ_CFLAGS"
+
+    old_LIBS=$LIBS
+    LIBS="$LIBS $ZMQ_LIBS"
+
+    AC_MSG_CHECKING([whether zeromq has CURVE support])
+    AC_RUN_IFELSE([
+        AC_LANG_SOURCE([[
+            #include <zmq.h>
+	    int main () {
+	        char pub[41];
+	        char sec[41];
+	        if (zmq_curve_keypair (pub, sec) < 0) return 1;
+	        return 0;
+	    }
+	]])],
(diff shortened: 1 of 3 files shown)


config/x_ac_zeromq.m4
```diff
@@ -4,5 +4,32 @@ AC_DEFUN([X_AC_ZEROMQ], [
...
+    old_LIBS=$LIBS
+    LIBS="$LIBS $ZMQ_LIBS"
+
+    AC_MSG_CHECKING([whether zeromq has CURVE support])
+    AC_RUN_IFELSE([
+        AC_LANG_SOURCE([
+            #include <zmq.h>
+	    int main () {
+	        char pub@<:@41@:>@;
+	        char sec@<:@41@:>@;
```
chu11, line 19: some weird terminal cut & paste error?  I don't know why I'm seeing these funky characters.

chu11: oh wait, length 41 char arrays.  why aren't the brackets showing up properly in my browser, it shows `@<:@`

grondo: Those are m4 [quadrigraphs](https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Quadrigraphs.html) (since you can't use `[]` inside `[]` in m4.)

grondo: But I do wonder if "double quotes" `[[ ]]` would work here? I'm far from an m4 expert, but it _seemed_ to work in a simple test:
```diff
diff --git a/config/x_ac_zeromq.m4 b/config/x_ac_zeromq.m4
index 16b09fb0e..fdb974707 100644
--- a/config/x_ac_zeromq.m4
+++ b/config/x_ac_zeromq.m4
@@ -12,15 +12,15 @@ AC_DEFUN([X_AC_ZEROMQ], [
 
     AC_MSG_CHECKING([whether zeromq has CURVE support])
     AC_RUN_IFELSE([
-        AC_LANG_SOURCE([
+        AC_LANG_SOURCE([[
             #include <zmq.h>
            int main () {
-               char pub@<:@41@:>@;
-               char sec@<:@41@:>@;
+               char pub[41];
+               char sec[41];
                if (zmq_curve_keypair (pub, sec) < 0) return 1;
                return 0;
            }
-        ])],
+        ]])],
        [have_zmq_curve_support=yes; AC_MSG_RESULT([yes])],
        [AC_MSG_RESULT([no])]
     )
```

chu11: thanks.  I thought it could be a special char format but couldn't figure out the special term to google.

chu11 (approved):
LGTM now that I know why those special chars are in there.

config/x_ac_zeromq.m4
garlick: Nice!  [[  Is far better.

garlick:
Thanks! I pushed @grondo's proposed change and I'll set MWP.

PR #6190, opened 2024-08-06
matrix: avoid overwriting env elements

trws:
The good news: Everything we request be built now builds.
The bad news: Nothing gets tagged because I accidentally triggered a re-use bug in generate-matrix.

problem: re-using the env argument revealed we were not copying the dict but modifying the argument in place, causing multiarch builds to only tag amd64, and matrix generation to fail

solution: copy the env dict in add_build

### DIFF
--- a/src/test/generate-matrix.py
+++ b/src/test/generate-matrix.py
@@ -96,7 +96,9 @@ class BuildMatrix:
         """Add a build to the matrix.include array"""
 
         # Extra environment to add to this command:
-        env = env or {}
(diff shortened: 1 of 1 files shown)


grondo (approved):
Ugh, sorry about that @trws! LGTM.

trws:
🤷  is what it is.  On the upside, I _think_ I worked out how to do per-target timeout as a side-effect of finding this, may throw something up shortly, could be nice.

Issue #6179, opened 2024-08-02
housekeeping doesn't release resource for a large job


grondo:
On elcap, flux-core and Fluxion got out of sync with allocated resources.

Affected resources were associated with a single job which was INACTIVE, but `flux housekeeping list` listed no tasks for the job. `flux dmesg` showed the following for the job:
```
[  +4.361834] job-exec[0]: exec aborted: id=f2AJsP9HniMm
[  +9.365820] job-exec[0]: Sending SIGKILL to job f2AJsP9HniMm
[ +14.362288] job-exec[0]: Sending SIGKILL to job f2AJsP9HniMm
[ +19.365833] job-exec[0]: Sending SIGKILL to job f2AJsP9HniMm
[ +24.370602] job-exec[0]: Sending SIGKILL to job f2AJsP9HniMm
[ +29.366845] job-exec[0]: Sending SIGKILL to job f2AJsP9HniMm
[ +29.366902] job-exec[0]: Sending SIGKILL to job shell for job f2AJsP9HniMm
[ +39.373830] job-exec[0]: Sending SIGKILL to job shell for job f2AJsP9HniMm
[  +0.545771] job-exec[0]: Sending SIGKILL to job shell for job f2AJsP9HniMm
[  +7.122439] job-exec[0]: job-exception: id=f2AJsP9HniMm: node failure on elcapX (shell rank Y)
[  +8.683464] job-manager[0]: f2AJsP9HniMm: epilog: stderr: f2AJsP9HniMm: epilog: rank M offline. Skipping.
[ +26.515790] job-manager[0]: housekeeping: elcapZ (rank M) f2AJsP9HniMm: No route to host
```

In this version of Flux, I had thought to also see messages:
```
housekeeping: JOBID complete
housekeeping: all resources of JOBID have been released
```

garlick:
The `all resources of JOBID have been released` message is logged at LOG_DEBUG so I don't believe it will be propagated to the systemd journal (or syslog) if that's where you were looking for it.  Unfortunately `flux dmesg` has wrapped.

I've been looking at the code and I haven't spotted a code path that would allow a job to be removed from the housekeeping list without logging that message (in 0.64.0).

grondo:
I should have been more explicit. The messages above were from `flux dmesg | grep JOBID`. The job did not appear in job-manager stats. Since we have the`exec aborted` message, it isn't possible that the `all resources have been released` had wrapped at the time...

garlick:
[...]

PR #6191, opened 2024-08-06
improve housekeeping logging and list management

garlick:
Problem: we observed disturbing loss of job state in housekeeping in #6179, where a job was apparently removed from the internal housekeeping list without releasing all resources.

This PR adds back a couple debug log messages that were helpful in discovering this but removed earlier in the current release cycle.  It also streamlines some `zlistx_t` handling that just felt a little loosey-goosey.


garlick:
I dunno if we wanted to get this into the release?  It still needs a review.  Pretty trivial set of changes IMHO.

grondo (approved):
Looks great to me!

Issue #6144, opened 2024-07-25
validator: reject jobs submitted as the root user


garlick:
Problem: the IMP refuses to start job shells as root, but root job submissions are accepted as valid.

Have the validator reject root jobs so they fail fast.

But don't reject root if the instance owner is also root since that does occur in containers.  The imp shouldn't be invoked in that case.

grondo:
Actually maybe this would be easier to implement directly in job-ingest module. I'll propose a simple PR.

PR #6194, opened 2024-08-06
reject jobs submitted as user root in a multi-user instance

grondo:
This PR addresses #6144 by rejecting jobs directly in job-ingest if they are submitted as userid 0 when the instance owner is not also 0.

The reasoning behind handling this directly in job-ingest is simplicity. Also, it does seem like this check should always be in place, since there is not currently a way a root job could execute successfully due to limits/guardrails in the flux-security behavior. Therefore, a validator plugin doesn't really make sense. (IMO)

### DIFF
--- a/src/modules/job-ingest/job-ingest.c
+++ b/src/modules/job-ingest/job-ingest.c
@@ -13,6 +13,7 @@
 #endif
 #include <time.h>
 #include <unistd.h>
+#include <sys/types.h>
 #include <jansson.h>
 #include <flux/core.h>
 #if HAVE_FLUX_SECURITY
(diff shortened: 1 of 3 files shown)


garlick (approved):
LGTM!

grondo:
Oops, this broke a test in `t2812-flux-job-last.t`. Looking into it.

grondo:
Ah, this is because a test for the `history` plugin requires submission of a job as root to test handling of uid `0`. I wonder if we should just add a config option for this behavior, or was a validator plugin a better option. :confused:

garlick:
Meh, maybe that test could just be skipped.

grondo:
It was simple enough to add an `allow-root-jobs` option to `flux module load job-ingest` to disable the check for root jobs. The `flux job last` test can then use that to keep testing handling of uid=0.

We'll see if this passes CI.

grondo:
Looks like that option worked. If that's acceptable I'll set MWP.

garlick:
Sounds good!

PR #6195, opened 2024-08-06
add `-B, --bank=` option to submission commands

grondo:
Now that `bank` and `project` have been [proposed](https://github.com/flux-framework/rfc/pull/420) as official members of canonical jobspec, and since they are supported by job-list, it seems like there are excellent arguments to at least support `-B, --bank=` options in our cli as suggested most recently by @ilumsden.

This PR starts with @ilumsden's draft PR #6157 (thanks Ian!), reworks the commit message to project standards, then adds the docs, testing, and new option to tab completions.

I didn't add `--project` here, even though it is also proposed as canonical in the rfc PR above. I like the idea of `-Sproject=NAME` for rarer cases such as this, though I'm not opposed to `--project` if people think that's a good idea.

For now, this seems like a good and simple ui improvement.

### DIFF
--- a/doc/man1/common/job-param-additional.rst
+++ b/doc/man1/common/job-param-additional.rst
@@ -6,6 +6,15 @@
    option is ignored, though :man1:`flux-jobs` may display the queue
    name in its rendering of the ``{queue}`` attribute.
 
+.. option:: -B, --bank=NAME
+
+   Set the bank name for this job to ``NAME``. This option is equivalent
+   to `--setattr=bank=NAME`, and results in the ``bank`` attribute being
+   set to ``NAME`` in the submitted jobspec. However, besides the bank
+   name appearing in job listing output, this option may have no effect
+   if no plugin or package that supports it (such as flux-accounting)
+   is installed and configured.
+
 .. option:: -t, --time-limit=MINUTES|FSD
 
    Set a time limit for the job in either minutes or Flux standard duration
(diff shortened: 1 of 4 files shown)


doc/man1/common/job-param-additional.rst
```diff
@@ -6,6 +6,15 @@
    option is ignored, though :man1:`flux-jobs` may display the queue
    name in its rendering of the ``{queue}`` attribute.
 
+.. option:: -B, --bank=NAME
+
+   Set the bank name for this job to ``NAME``. This option is equivalent
+   to `--setattr=bank=NAME`, and results in the ``bank`` attribute being
```
garlick, line 12: Isn't it `--setattr=system.bank=NAME`?

garlick (approved):
Thanks @ilumsden and @grondo!  LGTM - just one doc nit for @mark.

doc/man1/common/job-param-additional.rst
grondo: No. See [flux-submit(1)](https://flux-framework.readthedocs.io/projects/flux-core/en/stable/man1/flux-submit.html#cmdoption-flux-submit-setattr):

>  If KEY does not begin with system., user., or ., then system. is assumed.

garlick: Ah OK.  All good then.

grondo:
Thanks! I've set MWP.

PR #6197, opened 2024-08-06
job-manager: add final flag to sched.free

garlick:
This changes the RFC 27 `sched.free` payload to include a `final` flag, as proposed in flux-framework/rfc#421.

When present and true, the scheduler may assume that the last `sched.free` request has been received for a job's resource allocation, and could start an optional sanity check that no resources remain allocated to that job.  If missing, it's probably wise to not assume anything  at this point, for backwards compatibility with earlier flux core releases.

@trws is this what you were thinking?

### DIFF
--- a/src/modules/job-manager/alloc.c
+++ b/src/modules/job-manager/alloc.c
@@ -98,16 +98,20 @@ static void interface_teardown (struct alloc *alloc, char *s, int errnum)
 
 /* Send sched.free request.
  */
-int free_request (struct alloc *alloc, flux_jobid_t id, json_t *R)
+int free_request (struct alloc *alloc,
+                  flux_jobid_t id,
+                  json_t *R,
+                  bool final)
 {
     flux_msg_t *msg;
 
     if (!(msg = flux_request_encode ("sched.free", NULL)))
         return -1;
     if (flux_msg_pack (msg,
-                       "{s:I s:O}",
+                       "{s:I s:O s:b}",
                        "id", id,
-                       "R", R) < 0)
+                       "R", R,
+                       "final", final) < 0)
         goto error;
     if (flux_send (alloc->ctx->h, msg, 0) < 0)
         goto error;
@@ -180,7 +184,7 @@ static void alloc_response_cb (flux_t *h,
         (void)json_object_del (R, "scheduling");
 
         if (!job) {
-            (void)free_request (alloc, id, R);
+            (void)free_request (alloc, id, R, true);
             break;
         }
         if (job_priority_queue_delete (alloc->sent, job) < 0)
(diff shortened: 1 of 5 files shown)


trws:
Yes, this is exactly what I was hoping for.  @milroy, we should try to hook this in ASAP.

src/modules/job-manager/alloc.c
```diff
@@ -98,16 +98,20 @@ static void interface_teardown (struct alloc *alloc, char *s, int errnum)
...
 
     if (!(msg = flux_request_encode ("sched.free", NULL)))
         return -1;
     if (flux_msg_pack (msg,
-                       "{s:I s:O}",
+                       "{s:I s:O s:b}",
                        "id", id,
-                       "R", R) < 0)
+                       "R", R,
+                       "final", final ? 1 : 0) < 0)
```
trws, line 114: I've seen this pattern a few places, this is redundant due to int promotion in variadics right or is there a janssonism here I don't know about?

garlick: You're right - I wasn't using my brain at some point and then after that, monkey see, monkey do.  It should be fine due to integer promotion.  The varidaic unpack is where you could get into trouble since jansson booleans unpack to integers not bools. Will fix.

trws: Oh yeah, we actually had a nasty stack smash bug in sched from exactly that issue on the unpack side.  Makes sense.

grondo: Yeah, I think we've hit _very_ difficult to track down bugs in the unpack side and are a little too wary of pack/unpack of booleans with jansson.

grondo (approved):
LGTM! 
Ugh, sorry about the necessary churn in sched-simple.

grondo:
Hm, one new test failed in a builder:
```
 expecting success: 
  	flux dmesg -H | grep sched-simple >sched.log &&
  	grep "free: rank0" sched.log &&
  	grep "free: rank1" sched.log &&
  	grep "free: rank2" sched.log &&
  	grep "free: rank3.*(final)" sched.log
  
  [  +0.236926] sched-simple[0]: free: rank0/core0 f3ukmFVZ
  [  +0.272898] sched-simple[0]: free: rank1/core0 f3ukmFVZ (final)
  [  +0.220563] sched-simple[0]: free: rank2/core0 f3ukmFVZ
  not ok 11 - nodes were returned to scheduler separately
```

Could it racy? I haven't looked yet.

garlick:
Oh crud, yeah, the order cannot be relied on - let me fix.

garlick:
Restarted coverage builder that had timed out.

PR #6198, opened 2024-08-07
avoid idset leak when nodes leave the broker.online group

grondo:
Valgrind caught this leak when running a test under `FLUX_TEST_VALGRIND=t`.


garlick (approved):
Good catch!

PR #6200, opened 2024-08-07
matrix: remove fedora40 arm again

trws:
problem: everything now builds, the timeouts now work as expected, but the fedora40 builder takes so long that it often gets cancelled by another PR getting merged within that four hour window, as a result no manifest gets generated for that merge

solution: remove fedora40 for arm64 again

This should be the last one for a while.  The only other thing in here is I made it so that every single add_build has to specify an image.  It's too easy to lose track of which is which in some cases (I did a couple of times working on all of this) so better to be explicit.


grondo (approved):
Ah, good improvement!

PR #6201, opened 2024-08-07
include offline nodes in flux overlay errors output

garlick:
Problem:  `flux overlay errors` does not show textual errors recorded for nodes in the "offline" state - only nodes that are "lost".  Among the errors not displayed are nodes that tried to connect with a mismatched flux version, which was the original motivator for the command!

Display any textual errors recorded for offline nodes.  This includes the version mismatch.  It also will show "administrative shutdown" as an error for nodes that are shutdown cleanly with `systemctl stop flux`.  I guess this is probably OK even though it's not an "error" per se, since it will only add one line to the compressed output and its meaning is probably obvious.

Example
```
$ flux overlay status
0 picl0: partial
├─ 1 picl1: partial
│  ├─ 3 picl3: full
│  ├─ 4 picl4: offline administrative shutdown
│  └─ 5 picl5: full
└─ 2 picl2: partial
   ├─ 6 picl6: offline administrative shutdown
   └─ 7 picl7: offline client (0.64.0) version mismatched with server (0.65.0)

$ flux overlay errors
picl7: client (0.64.0) version mismatched with server (0.65.0)
picl[4,6]: administrative shutdown
```

### DIFF
--- a/doc/man1/flux-overlay.rst
+++ b/doc/man1/flux-overlay.rst
@@ -86,8 +86,9 @@ errors
 
 .. program:: flux overlay errors
 
-:program:`flux overlay errors` summarizes any errors recorded for lost nodes.
-The output consists of one line per unique error with a hostlist prefix.
+:program:`flux overlay errors` summarizes any errors recorded for lost or
+offline nodes.  The output consists of one line per unique error with a
+hostlist prefix.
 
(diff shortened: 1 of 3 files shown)


grondo (approved):
LGTM!

garlick:
Thanks!  Setting MWP.

PR #6203, opened 2024-08-08
add flux module trace

garlick:
After adding `flux overlay trace`, it seemed natural to add a tool for tracking messages sent/received by broker modules.

This also shovels some module related code in the main broker source file (broker.c) into modhash.c without modification.

Example: trace `resource` and `sched-simple` when a job is run:
```
$ flux module trace resource sched-simple
2024-08-08T09:57:10.250 sched-simple rx > sched.alloc [485]
2024-08-08T09:57:10.250 sched-simple tx > log.append [102]
2024-08-08T09:57:10.250 sched-simple tx > kvs.commit [330]
2024-08-08T09:57:10.250 sched-simple tx > log.append [90]
2024-08-08T09:57:10.251 sched-simple rx < kvs.commit [72]
2024-08-08T09:57:10.251 sched-simple tx < sched.alloc [283]
2024-08-08T09:57:10.267     resource rx > resource.topo-get [0]
2024-08-08T09:57:10.267     resource tx < resource.topo-get [17.05957K]
2024-08-08T09:57:10.275     resource rx > resource.disconnect [0]
2024-08-08T09:57:10.277 sched-simple rx > sched.free [188]
2024-08-08T09:57:10.277 sched-simple tx > log.append [96]
```

### DIFF
--- a/src/broker/module.c
+++ b/src/broker/module.c
@@ -45,6 +45,8 @@
 #include "modservice.h"
 
 struct broker_module {
+    flux_t *h;              /* ref to broker's internal flux_t handle */
+
     flux_watcher_t *broker_w;
 
     double lastseen;
@@ -78,6 +80,7 @@ struct broker_module {
 
     struct flux_msglist *rmmod_requests;
     struct flux_msglist *insmod_requests;
+    struct flux_msglist *trace_requests;
     struct flux_msglist *deferred_messages;
 
     flux_t *h_module_end;   /* module end of interthread_channel */
@@ -332,6 +335,7 @@ module_t *module_create (flux_t *h,
     p->main = mod_main;
     p->dso = dso;
     p->rank = rank;
+    p->h = h;
     if (!(p->conf = flux_conf_copy (flux_get_conf (h))))
         goto cleanup;
     if (!(p->parent_uuid_str = strdup (parent_uuid)))
@@ -349,7 +353,8 @@ module_t *module_create (flux_t *h,
     }
     if (!(p->path = strdup (path))
         || !(p->rmmod_requests = flux_msglist_create ())
-        || !(p->insmod_requests = flux_msglist_create ()))
+        || !(p->insmod_requests = flux_msglist_create ())
+        || !(p->trace_requests = flux_msglist_create ()))
         goto nomem;
     if (name) {
         if (!(p->name = strdup (name)))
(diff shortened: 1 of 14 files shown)


src/broker/module.c
```diff
@@ -151,14 +151,13 @@ static int module_finalizing (module_t *p)
     flux_future_t *f;
 
     if (!(f = flux_rpc_pack (p->h_module_end,
-                             "broker.module-status",
+                             "module.status",
```
grondo: Will the service name changes have any unintended consequences with version compatibility. (e..g perhaps when using `flux proxy`)? I can't think of much, but thought I'd bring it up just in case.

t/t3401-module-trace.t
```diff
@@ -0,0 +1,56 @@
+#!/bin/sh
+
+test_description='Test flux module trace'
+
+. $(dirname $0)/sharness.sh
+
+test_under_flux 4
```
grondo, line 7: Would an instance size=1 work for this test?

grondo (approved):
This LGTM!

I wonder if for consistency the module name should always be shown even if there is only one module being traced?

Also, I'm guessing `flux module trace` doesn't show up in coverage because it doesn't have a clean exit path. If motivated you could use `__gcov_flush()` at the end of the while loop to get coverage to report (or add a clean shutdown on SIGINT/SIGTERM) Doesn't seem too important though.

src/broker/module.c
garlick: I was assuming that for the most part people will not be running module commands through flux-proxy, and flux-proxy does make you force the connection if the versions are different, so maybe OK?

Oh that reminds me that the original topic strings are mentioned in [RFC 5](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_5.html), so I'll follow this with an rfc pr.

I don't think there is too much danger of this causing problems.

t/t3401-module-trace.t
garlick: Thanks! Just pushed the following changes
- `flux-module` calls exit(0) on SIGINT/SIGTERM
- module name is always printed
- test uses size=1

grondo: Great! Feel free to set MWP when you're ready. 

Oh! Also maybe add bash completions, though I can add that as a follow-on PR because we probably forgot about `flux overlay trace` as well?

garlick: Thanks!  Yeah I didn't think about that.  I'll go ahead and set MWP here as I'm headed out on an errand.

PR #6206, opened 2024-08-08
etc: add flux module/overlay trace to bash tab completion

grondo:
Problem: The new `flux overlay trace` and `flux module trace` commands are missing from the bash tab completions file.

Add them.


garlick (approved):
Merci

Issue #6202, opened 2024-08-08
Suggestion: ability to specify a node range in flux jobs


kkier:
Scenario:
As a sysadmin, I'm planning to do maintenance on an arbitrary set of nodes.  After draining those nodes, I'd like an easy way to return a list of jobs running on them so I can contact the job owners/see when they'll expire/kill them outright/whatever.  Something like `flux jobs -i <hostlist>` would make sense.

Design-wise, it arguably makes sense assume `-A` but that breaks the current paradigm.

grondo:
Note there is _sort of_ a solution to this proposed in #5711. It adds a query syntax to the `-f, --filter` option that would allow you to get this result with `flux jobs -f host:<hostlist> -f running`. That PR has never been merged because there isn't good consensus on the interface.

Perhaps filtering `flux jobs` by hostlist would be common enough that it should get its own option. I like your idea of using the `-i, --include=hosts|ranks` from the `flux resource` utilities, though I also wonder if that could be confusing in `flux jobs` which isn't listing resources (and not all jobs even have had resources assigned).

Anyway, we could support this option and `and` it with other filter options including any result of an advanced filter from `-f, --filter=` in the future.

PR #6209, opened 2024-08-10
flux-jobs: add `-i, --include=HOSTS|RANKS` option

grondo:
This PR takes a bit of the minor work from #5711 and uses it to implement a `-i, --include=HOSTS|RANKS` option for `flux jobs` as suggested in #6202.

This might be a more palatable first step that what's proposed in #5711, is useful on its own, and if the advanced `-f, --filter` "query syntax" is ever merged, the hostlist/ranks constrain can easily be combined (if used) with `and`.

The option name of `-i, --include` could be controversial here. I wouldn't have thought of it except that it was proposed by @kkier, and I like that it matches the similar options in `flux-resource(1)` for consistency. I also like that it works for both ranks and hostnames.

However, easy to change if others do not agree.

### DIFF
--- a/doc/man1/flux-jobs.rst
+++ b/doc/man1/flux-jobs.rst
@@ -51,6 +51,13 @@ OPTIONS
    List jobs in a specific queue or queues. Multiple queues may be separated
    by a comma or by using the :option:`-q, --queue` option multiple times.
 
+.. option:: -i, --include=HOSTS|RANKS
+
+   List only jobs where the assigned resources intersect with the supplied
+   argument, which may be specified either as an RFC 22 idset of broker ranks
+   or an RFC 29 hostlist of host names. It is not an error to specify ranks or
+   hosts which do not exist.
+
 .. option:: -c, --count=N
 
    Limit output to N jobs (default 1000)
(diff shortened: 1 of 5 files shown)


doc/man1/flux-jobs.rst
```diff
@@ -51,6 +51,13 @@ OPTIONS
    List jobs in a specific queue or queues. Multiple queues may be separated
    by a comma or by using the :option:`-q, --queue` option multiple times.
 
+.. options:: -i, --include=HOSTS|RANKS
+
+   List only jobs where the assigned resources intersect with the supplied
+   argument, which may be specified either as an idset of broker ranks or
+   list of hosts in hostlist form. It is not an error to specify ranks or
+   hosts which do not exist.
```
garlick, line 59: Might want to use "RFC 22 idset" and "RFC 29 hostlist" and add the references to a FLUX RFC section e.g.
```
FLUX RFC
========

| :doc:`rfc:spec_22`
| :doc:`rfc:spec_29`
```

garlick (approved):
Looks great to me.  Just one optional suggestion for the man page.

doc/man1/flux-jobs.rst
grondo: Good suggestion! I've done that and also fixed a typo and force-pushed the result.

grondo:
Setting MWP here unless I hear objections.

Issue #2547, opened 2019-11-20
kvs: duplicate entries in eventlog


chu11:
While investigating the assert in #2545, it was determined that the cause of memory corruption is the fact that some job eventlogs have duplicate entries.  The `job-info` code incorrectly assumed that an eventlog can have at most 1 `submit` event.  While this assumption can be easily corrected, it does not explain why some eventlogs have duplicate entries.  Perhaps something racy in the `job-manager`?  in `job-ingest`? A bug with KVS append?

To ensure that this bug wasn't due to any recent activities, such as the KVS append performance in #2526, I went all the way back to commit 33a14e5f14b5a5c3264b6f683015c259af7d5db1, which is BEFORE job-info supported monitoring job states.

I ran the following:

```
src/cmd/flux start --size=8
flux mini run -n1 --dry-run hostname > job.json
flux mini run -N1 --dry-run t/ingest/submitbench -r 1024 job.json > submitjob.json
t/ingest/submitbench -r 8 submitjob.json
```

I waited until `flux job list` became empty (note that b/c this is an old branch, `flux job list` was the old job list from the job manager).  Then I ran the following slow script, b/c it's the fastest thing I could hack up in short order.

```
#!/bin/sh

paths=`flux kvs ls -R | grep job | grep -v guest`
for path in $paths
do
    path=`echo $path | sed 's/:$//'`
    len=`echo $path | wc -c`
    if [ "$len" == "24" ]; then
        #echo $path
        count=`flux kvs eventlog get ${path}.eventlog | grep submit | wc -l`
        if [ "$count" -gt "1" ]; then
            echo "**** path ${path}.eventlog has eventlog with > 1 submit event"
            flux kvs eventlog get ${path}.eventlog
        fi
    fi
done
```

And sure enough, there are plenty of jobs with duplicate entries in their eventlogs.

```
**** path job.0000.006c.df00.0001.eventlog has eventlog with > 1 submit event
1574291111.921462 submit {"userid":8556,"priority":16,"flags":0}
1574291111.921462 submit {"userid":8556,"priority":16,"flags":0}
1574291111.921462 submit {"userid":8556,"priority":16,"flags":0}
1574291118.450572 depend
1574291118.450572 depend
1574291118.450572 depend
1574291429.125314 alloc {"note":"rank2/core0"}
1574291429.151407 start
1574291429.243822 finish {"status":0}
1574291429.362435 release {"ranks":"all","final":true}
1574291429.371399 free
[...]

PR #6210, opened 2024-08-10
kvs: correct code logic about what is an append

### LINKED ISSUE #6207: throughput.py test: incorrect number of tests submitted in status
I've noticed this a few times, where the number of jobs submitted does not match the number of jobs requested.

```
 32768 jobs:  33478 submitted,      0 running, 32768 completed                                                                              
number of jobs: 32768                                                                                                                       
submit time:    18.076s (1812.8 job/s)                                                                                                      
script runtime: 511.485s                                                                                                                    
job runtime:    511.244s                                                                                                                    
throughput:     64.1 job/s (script:  64.1 job/s)                                                                                            
total 708M                                     
```

the test appears to still run the correct number of tests, but the "submitted" number is incorrect.  I noticed one at around 39K as well.

chu11:
Problem: In PR #2547, a fix was added into the KVS to deal with duplicate appends to a KVS entry.  However, a corner case was left in the code. When an append is done to a KVS entry that does not exist, the append is treated like a normal insertion.  This corner case should also be counted as an "append".  Otherwise, duplicate appends could happen at the beginning of a KVS entry when it is created.

Solution: Set the append flag to true when a KVS entry does not exist and the append is treated like an insert.

Fixes #6207

----

Side note, now that all "good paths" are counted as appends, I could moved  `(*append) = true` into the end of the function before `return 0`.  But given how the error handling is done in the function, I felt leaving each of the `(*append) = true` near their "we did an append" I thought was better.  But that's just me.  Could be overridden by review votes :-)

### DIFF
--- a/src/modules/kvs/kvstxn.c
+++ b/src/modules/kvs/kvstxn.c
@@ -495,6 +495,10 @@ static int kvstxn_append (kvstxn_t *kt, json_t *dirent,
         /* entry not found, treat like normal insertion */
         if (treeobj_insert_entry (dir, final_name, dirent) < 0)
             return -1;
+        /* N.B. although this is an "insert", we still treat this as
+         * an "append".  If we don't, the "append" could be issued
+         * twice, leading to duplicated data.  See issue #6207. */
+        (*append) = true;
     }
     else if (treeobj_is_valref (entry)) {
         char ref[BLOBREF_MAX_STRING_SIZE];


src/modules/kvs/kvstxn.c
```diff
@@ -495,6 +495,7 @@ static int kvstxn_append (kvstxn_t *kt, json_t *dirent,
         /* entry not found, treat like normal insertion */
         if (treeobj_insert_entry (dir, final_name, dirent) < 0)
             return -1;
+        (*append) = true;
```
grondo, line 498: Not knowing this code well (or at all), a comment here as to why `*append = true` would be helpful, especially given the comment just above (`treat like normal insertion`).

grondo (approved):
Nice! LGTM! just one suggestion of an additional comment in the code.

I also verified I can no longer reproduce the issue with this fix. Thanks!

chu11:
thanks!  re-pushed with an extra comment.  will set MWP.

PR #6212, opened 2024-08-10
testsuite: fix racy flux job attach test

### LINKED ISSUE #6211: not ok 17 - attach: reports job shell Killed if job shell is killed
Problem: this test frequently fails when 'make check' is run with a high -j value.
```
not ok 17 - attach: reports job shell Killed if job shell is killed
#	
#		id=$(flux submit --wait-event=exec.shell.start sh -c "kill -9 \$PPID") &&
#		test_must_fail_or_be_terminated flux job attach $id 2>shell-killed.out &&
#		grep "job shell Killed" shell-killed.out
#	
```

garlick:
Problem: when t2500-job-attach.t is run on a busy system, the "attach: reports job shell Killed if job shell is killed" test sometimes fails.

'flux submit --wait-event=exec.shell.start kill -9 $PPID' may fail if the shell is terminated before shell.start can be posted.

Moving the --wait-event to the 'flux job attach' command seems to address the problem.

Fixes #6211

### DIFF
--- a/t/t2500-job-attach.t
+++ b/t/t2500-job-attach.t
@@ -129,7 +129,7 @@ test_expect_success 'attach: reports Terminated when tasks are terminated' '
 	grep Terminated terminated.err
 '
 test_expect_success 'attach: reports job shell Killed if job shell is killed' '
-	id=$(flux submit --wait-event=exec.shell.start sh -c "kill -9 \$PPID") &&
+	id=$(flux submit sh -c "kill -9 \$PPID") &&
 	test_must_fail_or_be_terminated flux job attach $id 2>shell-killed.out &&
 	grep "job shell Killed" shell-killed.out
 '


t/t2500-job-attach.t
```diff
@@ -129,8 +129,8 @@ test_expect_success 'attach: reports Terminated when tasks are terminated' '
 	grep Terminated terminated.err
 '
 test_expect_success 'attach: reports job shell Killed if job shell is killed' '
-	id=$(flux submit --wait-event=exec.shell.start sh -c "kill -9 \$PPID") &&
-	test_must_fail_or_be_terminated flux job attach $id 2>shell-killed.out &&
+	id=$(flux submit sh -c "kill -9 \$PPID") &&
+	test_must_fail_or_be_terminated flux job attach --wait-event=exec.shell.start $id 2>shell-killed.out &&
```
grondo, line 133: Thanks! I kept meaning to look into this one.

I think the `--wait-event=exec.shell.start` should also be dropped from `flux job attach`. I'm pretty sure it is equivalent to `--wait-event=clean` since I don't think `flux job attach` supports waiting on exec events like `flux submit` does.

In retrospect, the `--wait-event=exec.shell.start` was an accidental cut-and-paste from the previous 2 tests, and should not have been used here I think.

garlick: Ah that makes sense -thanks!  I pushed an updated patch.

grondo (approved):
LGTM! Thanks!!

PR #6214, opened 2024-08-14
cron: use posix_spawn() for cron tasks when possible

### LINKED ISSUE #6104: flux-cron: cron tasks force use of fork/exec by default
It appears that flux-cron tasks set a CWD by default before calling `flux_rexec()` thereby requiring the broker fallback to fork/exec instead of using `posix_spawn()`. When the broker RSS is large, this is a huge performance penalty. Also, when using tcmalloc, this causes the forked task to hang in libtcmalloc and the broker hangs waiting for the child to reach exec.

grondo:
This PR fixes #6104 by allowing cron tasks to have a default, unset cwd if just using the broker cwd. This allows use of the posix_spawn() default subprocess mechanism in most cases, instead of always falling back to fork().

### DIFF
--- a/src/modules/cron/cron.c
+++ b/src/modules/cron/cron.c
@@ -444,10 +444,7 @@ static cron_entry_t *cron_entry_create (cron_ctx_t *ctx, const flux_msg_t *msg)
         goto out_err;
     }
 
-    if (!cwd)
-        cwd = ctx->cwd;
-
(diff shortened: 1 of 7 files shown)


garlick (approved):
LGTM!

grondo:
Thanks! Setting MWP.

Issue #6169, opened 2024-08-01
content-sqlite: preallocate database space


garlick:
Problem: when the disk containing the sqlite database fills up on rank 0, flux does not handle it well, e.g.
- #6155
- #6154
- #6100
- #6099

@trws suggested providing an option to pre-allocate space in the database.

One approach is to write a huge blob to the db and remove it:
https://stackoverflow.com/questions/827010/sqlite-pre-allocating-database-size

Some steps may also need to be taken to ensure writes to `content.sqlite-wal` do not fail.  More study required on that one:
https://www.sqlite.org/wal.html

chu11:
i was thinking about this last night and my initial intuition was that kind of trick wouldn't work.  I thought that in most modern databases they constantly append new data and the only way to reclaim disk space is through "compacting"/"defragmenting" it.  In sqlite that appears to be via "vacuum" (https://www.sqlite.org/lang_vacuum.html).

And even if we did this, we might be racing against the thingie filling up disk (e.g. errant logging).  But it's better than nothing.

The hack above seems to work around this fact by pre-allocating an entire database table then dropping it.  So perhaps this will work around it, but it seems to suggest we need to pre-allocate the entire database table before creating it the first time?  Perhaps this idea might work if we pre-allocate critical "checkpoint" and similar things to a separate database table than the core KVS?

Nevertheless, it's more than worthwhile to experiment with this idea and see.

garlick:
We should do an experiment, but based on those stack overflow comments, it seems like sqlite must reuse space within the db file, and vacuum just shrinks the file to not contain any free space.

> And even if we did this, we might be racing against the thingie filling up disk (e.g. errant logging). But it's better than nothing.

I was thinking this would be done up front when flux starts.  The sys admins would configure `content.sqlite.preallocate = "20G"` or something and if that cannot be grabbed when the module is initialized it fails and flux fails to start.  So if the disk fills up later, we're immune.

Same basic idea as a partition but apparently a partition is more of a pain for the sys admins.

chu11:
So the first test didn't seem to work.

Created a 5M tmpfs.

```
$ rm -rf /test/tmpfs-5m/statedir/*
$ df -h /test/tmpfs-5m/
Filesystem      Size  Used Avail Use% Mounted on
tmpfs           5.0M     0  5.0M   0% /test/tmpfs-5m
$ src/cmd/flux start -o,-Sstatedir=/test/tmpfs-5m/statedir
(inside instance)
[...]

PR #6217, opened 2024-08-15
WIP content-sqlite: preallocate space to avoid outside diskfull events

chu11:
Per discussion in #6169

Support a new "preallocate" config & module option to pre-allocate a specific amount of space to sqlite so that we can hopefully avoid ENOSPC issues when outside actors fill up the disk.  Option name debatable ... "preallocate_size"?  

The technique is to basically create a database table, write a bunch a data to it, then drop the table.   This won't work with journaling b/c the journal needs space too.  But code is added to disable journaling if ENOSPC is hit.

There's a lot of "setup" commits in here, we could probably split out into another PR.

TODOs

- [x] ~~option take meg/gig suffix?~~  ehhh scratch that, i don't what the config file inputs to always be strings
- [x] test at bigger scale/size (i.e. not 5m mounts)
- [x] I probably should document this somewhere
- [ ] BIG THING TO DISCUSS IN ISSUE - how to put database back into journaling mode when space issues fixed.  Disabling journaling does lead to a healthy slowdown in performance (job throughupt maybe 15% down).  This can probably be a different issue/PR, but needs to be discussed.

### DIFF
--- a/etc/rc1
+++ b/etc/rc1
@@ -54,7 +54,9 @@ if test $RANK -eq 0; then
             rm -f ${dumplink}
         fi
     elif test "${backingmod}" != "none"; then
-        flux module load ${backingmod}
+        date > /tmp/timing
+        flux module load ${backingmod} preallocate=50000000000
+        date >> /tmp/timing
     fi
 fi
 


garlick:
Nice progress!

Want to break this up into multiple PRs before we start a review?

chu11:
> Want to break this up into multiple PRs before we start a review?

Yeah, lemme split out the cleanups, then the "setups", and stuff into new PRs.

doc/man5/flux-config-content-sqlite.rst
```diff
@@ -0,0 +1,53 @@
...
+   synchronous = "NORMAL"
+   preallocate = 10737418240
+
+
+SEE ALSO
+========
+
+:man5:`flux-config`,
+`sqlite <https://www.sqlite.org/>`_,
+`sqlite pragmas <https://www.sqlite.org/pragma.html>`_
```
garlick, line 53: Might want to put those URLs in a RESOURCES section like most of the other section 5 man pages. 
The URLs are spelled out so you can cut and paste them from the man command output.
For example, in flux-config(5):
```rst
RESOURCES
=========

.. include:: common/resources.rst

Flux Administrator's Guide: https://flux-framework.readthedocs.io/projects/flux-core/en/latest/guide/admin.html

TOML: Tom's Obvious Minimal Language: https://toml.io/en/
```

src/modules/content-sqlite/content-sqlite.c
```diff
@@ -936,7 +940,9 @@ int mod_main (flux_t *h, int argc, char **argv)
         goto done;
     if (process_args (ctx, argc, argv, &truncate) < 0)
         goto done;
-    if (content_sqlite_opendb (ctx, truncate) < 0)
+    if (content_sqlite_setup (ctx, truncate) < 0)
+        goto done;
```
garlick, line 1127: Function name is not very descriptive.  Maybe just do the 
```c
if (truncate)
    (void)unlink (ctx->dbfile)
```
inline?

garlick (commented):
I know this is a WIP but I wanted to mention a couple thoughts.

We probably don't need to have a special table for the preallocated space.  We have a blob table already - couldn't we just write some number of blobs then delete them?  (Less code)

Do we want to fail if we cannot preallocate the requested amount?  I would think not?  It seems messed up if say we want to squat on 100G but are using 10G and can't start.

garlick:
After consulting with @kkier a bit on this, we think it may be better to present admins with a choice of 
- ensure flux never runs out of space on rank 0 (whether by partition or whatever), or
- flux goes down hard on ENOSPC and, when manually restarted, recovers what it can but doesn't stop for manual intervention

That is a trade-off they are most qualified to make based on their judgement of the relative impacts.   From their perspective I think it's more of a question of whether they want to reserve some amount of disk for flux or share it among multiple consumers, rather than whether or not they want to make a partition.  If this PR successfully provided a mechanism to reserve space without a partition, it wouldn't change much in that calculus.

garlick:
I could be wrong but it seems like, for transactional consistency, the delete of the placeholder must go through the journal if the journal is active, but if the file system is full, that would fail.  If you close the database, the close ought to try to flush the journal backlog, but of course that will fail if you can't delete the placeholder.

Does that make sense?

garlick:
It sounds like once the database is in WAL mode, this persists across a reopen.

https://www3.sqlite.org/wal.html

(see "Persistence of WAL Mode")

That seems like it means if we like WAL mode, we won't be able to accomplish much by reserving space in the db.  It likely also means that the WAL is on checkpointed on `sqlite3_close()` like I thought.

garlick:
> goes down hard meaning broker shall exit? Edit: and bring down full instance?

Well it does seem like `content-sqlite` should just close the database after we get an ENOSPC.  I think we probably need to talk through what can happen next.  Ideally it minimizes the need to manually intervene.

chu11:
> I could be wrong but it seems like, for transactional consistency, the delete of the placeholder must go through the journal if the journal is active, but if the file system is full, that would fail. If you close the database, the close ought to try to flush the journal backlog, but of course that will fail if you can't delete the placeholder.

Doh!  You're right.  I forgot that we need space for the journal or WAL.

Issue #6218, opened 2024-08-15
job-manager: jobs with alloc-bypass flag appear to still generate a free request to scheduler


grondo:
It appears that alloc-bypass jobs are generating a free request to the scheduler (this is in v0.64.0) based on this result:
```
$ flux jobtap load alloc-bypass.so
$ flux submit --setattr=alloc-bypass.R="$(flux resource R)" -n1 /bin/true
f4bQYoQKq
Aug 15 12:01:32.680885 PDT sched-fluxion-qmanager.err[0]: jobmanager_free_cb: can't find queue for job (id=7933744119808): No such file or directory
```

Will have to run this down, but since there's no real impact other than the error message from the scheduler, this could have easily been broken for awhile without any CI tests detecting it.

grondo:
Ah, ok, this broke when housekeeping was introduced. All jobs now call `housekeeping_start()` whether housekeeping is configured or not (the common case). If housekeeping is not configured, then all resources are immediately released. Since housekeeping doesn't keep a reference to the job (only R, jobid, and userid) the alloc-bypass flag is dropped so there's no way to skip the free request at that level.

My intuition says that alloc-bypass jobs can also bypass housekeeping. However, I don't think that was true for any configured prolog/epilog, so perhaps more thought required.

garlick:
It sounds right to skip housekeeping for alloc-bypass.  Sorry I flubbed that one!

PR #6219, opened 2024-08-15
job-manager: skip housekeeping for alloc-bypass jobs

grondo:
This is the most minimal fix for #6218 - it just completely skips housekeeping for alloc-bypass jobs.
This _seems_ like the right approach, i.e. it may be surprising to have housekeeping run, if configured, for jobs using alloc-bypass. However, prolog/epilog actions are not similarly skipped for alloc-bypass jobs (and this is probably necessary, because some of these actions may be required for job functionality), so an argument could be made to keep housekeeping support for alloc-bypass.

Anyway, this PR is WIP to ensure we have agreement first on no support for housekeeping on alloc-bypass jobs.

This PR also turns a failed free request in sched-simple from a log message into a hard error so that we can catch future failures in the testsuite and CI.

### DIFF
--- a/src/modules/job-manager/event.c
+++ b/src/modules/job-manager/event.c
@@ -328,10 +328,15 @@ int event_job_action (struct event *event, struct job *job)
                 && !job->perilog_active
                 && !job->start_pending
                 && !job->free_posted) {
-                if (housekeeping_start (ctx->housekeeping,
-                                        job->R_redacted,
-                                        job->id,
-                                        job->userid) < 0)
+
+                /* Release resources to housekeeping, but only if they were
+                 * allocated by the scheduler.
+                 */
+                if (!job->alloc_bypass
+                    && housekeeping_start (ctx->housekeeping,
+                                           job->R_redacted,
+                                           job->id,
+                                           job->userid) < 0)
                     return -1;
                 if (event_job_post_pack (ctx->event, job, "free", 0, NULL) < 0)
                     return -1;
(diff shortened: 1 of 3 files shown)


grondo:
Given feedback in #6218, I'll remove WIP here.

src/modules/job-manager/event.c
```diff
@@ -328,10 +328,16 @@ int event_job_action (struct event *event, struct job *job)
...
                 && !job->free_posted) {
-                if (housekeeping_start (ctx->housekeeping,
-                                        job->R_redacted,
-                                        job->id,
-                                        job->userid) < 0)
+
+                /* Only start housekeeping if this is not an alloc_bypass
+                 * job. This avoids sending a free to scheduler even if
+                 * housekeeping is not configured.
+                 */
```
garlick, line 335: Since this comment follows a comment at the top of the block that ends with
> ... it is safe for the job to release its resources to housekeeping.

maybe it would flow better to say
> Release resources to housekeeping, but only if they were allocated by the scheduler.

garlick (approved):
LGTM!

I had one minor comment about a comment but I won't be offended if you don't feel it's worth the effort :-)

src/modules/job-manager/event.c
grondo: Oh good suggestion, thanks!

grondo: Made that change and I'll set MWP. Thanks!

PR #6221, opened 2024-08-16
content-sqlite: add sqlite config to module stats

chu11:
Problem: Some module configuration varies depending on flux configuration.  In addition, future sqlite configuration may change
during operations.  There is no way to know what the current configuration is.

Output the journal_mode and synchronous settings in the module stats.


garlick (approved):
LGTM!

PR #6222, opened 2024-08-16
content-sqlite: set sqlite pragmas with flux-config

chu11:
Problem: In content-sqlite, the journal_mode and synchronous configs for sqlite can only be configured via the module command line.  This can be inconvenient, especially for testing.

Support their configuration via the flux config as well.

### DIFF
--- a/src/modules/content-sqlite/content-sqlite.c
+++ b/src/modules/content-sqlite/content-sqlite.c
@@ -74,10 +74,20 @@ struct content_sqlite {
     size_t lzo_bufsize;
     void *lzo_buf;
     struct content_stats stats;
-    const char *journal_mode;
-    const char *synchronous;
+    char *journal_mode;
+    char *synchronous;
     bool truncate;
 };
+
+static int set_config (char **conf, const char *val)
+{
+    char *tmp;
+    if (!(tmp = strdup (val)))
+        return -1;
+    free (*conf);
+    *conf = tmp;
+    return 0;
+}
 
 static void log_sqlite_error (struct content_sqlite *ctx, const char *fmt, ...)
 {
@@ -729,6 +739,8 @@ static void content_sqlite_destroy (struct content_sqlite *ctx)
         free (ctx->dbfile);
         free (ctx->lzo_buf);
         free (ctx->hashfun);
+        free (ctx->journal_mode);
+        free (ctx->synchronous);
         free (ctx);
         errno = saved_errno;
     }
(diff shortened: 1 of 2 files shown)


garlick:
Would it make sense to add dynamic re0configuration here as well?

chu11:
> Would it make sense to add dynamic re0configuration here as well?

I just saw your comment in #6169 about possibly using `flux config reload`.  Hmmm.  This change is predominantly for testing.  Perhaps we can leave that thought for later?

chu11:
```
  expecting success: 
  	cat >content-sqlite.toml <<-EOT &&
  	[content-sqlite]
  	journal_mode = "FOO"
  	synchronous = "EXTRA"
  	EOT
  	FLUX_CONF_DIR=$(pwd) test_must_fail flux start \
  	   -o,-Sbroker.rc1_path=$rc1_kvs,-Sbroker.rc3_path=$rc3_kvs \
  	   flux module stats content-sqlite &&
  	rm content-sqlite.toml
  
  {
   "object_count": 1,
   "dbfile_size": 20480,
   "dbfile_free": 17637527552,
   "load_time": {
    "count": 0,
    "min": 0.0,
    "max": 0.0,
    "mean": 0.0,
    "stddev": 0.0
   },
   "store_time": {
    "count": 1,
    "min": 0.025267999999999999,
    "max": 0.025267999999999999,
    "mean": 0.025267999999999999,
    "stddev": 0.0
   },
   "config": {
    "journal_mode": "OFF",
    "synchronous": "OFF"
   }
  }
  test_must_fail: command succeeded: flux start -o,-Sbroker.rc1_path=/usr/src/t/rc/rc1-kvs,-Sbroker.rc3_path=/usr/src/t/rc/rc3-kvs flux module stats content-sqlite
  not ok 56 - invalid config fails (journal_mode)
  ```
  
  hmmmm .... one builder fails with the new tests i added.  wondering if it's a path issue and not finding it?  hmmmm

**Comment by chu11** [comment]
ok, so i don't know why, but there's some subtlety with `FLUX_CONF_DIR=$(pwd) test_must_fail flux core ...` on that one builder.  Perhaps environment not passed through.  Switching to `-o,--config-path=$(pwd)` appears to work around it.

File: src/modules/content-sqlite/content-sqlite.c [hunk 1]
```diff
@@ -830,10 +854,18 @@ static int process_args (struct content_sqlite *ctx,
     int i;
     for (i = 0; i < argc; i++) {
         if (strstarts (argv[i], "journal_mode=")) {
+            if (!journal_mode_valid (argv[i] + 13)) {
```
garlick, line 892: Maybe it would be good to add a `flux_error_t` argument and call out which parameter is wrong.

src/modules/content-sqlite/content-sqlite.c
```diff
@@ -846,6 +846,41 @@ static bool synchronous_valid (const char *s)
     return true;
 }
 
+static int process_config (struct content_sqlite *ctx,
```
garlick, line 849: commit message (and PR title): maybe it would be better to say "set sqlite pragmas with flux-config" rather than "config via file" since it need not be an actual file.

garlick (approved):
LGTM!  Just a couple of minor comments.

src/modules/content-sqlite/content-sqlite.c
chu11: good idea, although in this case I think we just do a log message, since we're not responding to a RPC

chu11:
just re-pushed with those simple changes.   will set MWP

PR #6223, opened 2024-08-20
update feasibility RPCs to conform to RFC 27

grondo:
This PR updates sched-simple, the feasibility validator plugin, and docs to conform with the RFC 27 specification of a `feasibility` service and `feasibility.check` RPC instead of `sched.feasibility`. This disentangles the feasibility checks from the scheduler service, and therefore allows more flexibility in how and where the feasibility service is implemented.

Before this PR passes CI, a small PR may be necessary in flux-sched to have qmanager register the `feasibility` service and add a handler for `feasibility.check` requests. (Later, a feasibility specific module in Fluxion may be supported)

### DIFF
--- a/doc/guide/admin.rst
+++ b/doc/guide/admin.rst
@@ -780,7 +780,7 @@ distributed with Flux is shown below:
 
   $ flux job-validator --list-plugins
   Available plugins:
-  feasibility           Use sched.feasibility RPC to validate job
+  feasibility           Use feasibility service to validate job
   jobspec               Python bindings based jobspec validator
   require-instance      Require that all jobs are new instances of Flux
   schema                Validate jobspec using jsonschema
--- a/etc/rc1
+++ b/etc/rc1
@@ -102,7 +102,7 @@ done
 
 # Print module that has registered 'sched' service, if any
 lookup_sched_module() {
-    flux module list | awk '$NF == "sched" { print $1 }'
+    flux module list | awk '$NF ~ /^\s*(\w+,)?sched(,\w+)?\s*$/ { print $1 }'
 }
 
 if test $RANK -eq 0 -a "${FLUX_SCHED_MODULE}" != "none" \
--- a/src/bindings/python/flux/job/validator/plugins/feasibility.py
+++ b/src/bindings/python/flux/job/validator/plugins/feasibility.py
@@ -16,9 +16,6 @@ This allows jobs which are making infeasible or otherwise invalid
 requests to be rejected by the scheduler before ingest, instead of
 when the scheduler attempts to allocate resources for them.
 
-The RPC endpoint for feasibility checks can be set with an optional
---feasibility-service=NAME option.
-
 ENOSYS errors from the RPC are ignored, in case there is no feasibility
 service currently loaded.
 """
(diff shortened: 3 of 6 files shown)


grondo:
Sched PR opened: https://github.com/flux-framework/flux-sched/pull/1275

garlick (approved):
LGTM!

grondo:
Thanks! I'll set MWP now that flux-framework/flux-sched#1275 has merged.

grondo:
Uhoh, I should have done a bit more testing the flux-sched tests are still failing on this branch. Looking into it.

garlick:
Oh the registration happens after `flux_module_set_running()`.  Maybe there is a race?

grondo:
> Oh the registration happens after flux_module_set_running(). Maybe there is a race?

Yeah, good catch, that could be it.

> Edit, also the implemntation just passes the payloads on to sched-fluxion-resource.satisfiability (and stalls itself in the process). Maybe resource should register the service?

I didn't want to change sched-fluxion-resource at this point to avoid conflicting with Jacob's work to place the feasibility service into a separate module.. However, maybe that would be a fairly safe change (and improvement in the process)?

garlick:
> I didn't want to change sched-fluxion-resource at this point to avoid conflicting with Jacob's work to place the feasibility service into a separate module.. However, maybe that would be a fairly safe change (and improvement in the process)?

Eh, probably fine to defer that then.  It was just a startling revelation that qmanager adds nothing here (that I can see!)

grondo:
Bah, the issue is a cut-and-paste error in my patch to register the `feasibility` service (it re-registers the `sched` service, which fails) :-(

garlick:
Oh I'm sorry I missed that!

grondo:
Also I guess sched-simple and fluxion should unregister the `feasibility` service, or is it acceptable to let the module unload force the unregister by disconnect?

garlick:
When modules unload, `service_remove_byuuid()` is called in the broker so I think it's OK.

Hmm, offhand I'm not spotting code to unregister services on a disconnect which could be a problem for separate processes registering services.  I'll open an issue if that's the case.

garlick:
> Hmm, offhand I'm not spotting code to unregister services on a disconnect which could be a problem for separate processes registering services. I'll open an issue if that's the case.

NM, I forgot that request forwarding has to be implemented for services registered by a process connected to `connector-local`, so it is the `connector-local` module that handles the disconnect.   All is well.

grondo:
Ok, I've proposed https://github.com/flux-framework/flux-sched/pull/1276 (and tested it to the best of my ability)

grondo:
Oh, oops, this change also breaks this check in rc1:
```sh
# Print module that has registered 'sched' service, if any
lookup_sched_module() {
    flux module list | awk '$NF == "sched" { print $1 }'
}
```

Since qmanager and sched-simple will now display `sched,feasibility` or `feasibility,sched` in the last column.

We could "fix" this so that fluxion can work with existing versions of flux-core by moving the `feasibility` service to the `sched-fluxion-resource` module as @garlick suggested. 

We should also fix the above snippet of `rc1`. I wonder if we should add an option or argument to `flux module list` that can list modules by name or service, such that `flux module list sched` would return `sched-simple` or `sched-fluxion-qmanager`, e.g. something like:

```console
$ flux module list sched
Module                   Idle  S Sendq Recvq Service
sched-simple             idle  R     0     0 feasibility,sched
$ flux module list foop
Module                   Idle  S Sendq Recvq Service
flux-module: foop: service not found
```

grondo:
I'll save the `flux module list` improvements for a follow-on PR. For now I've pushed an update to the `rc1` `lookup_sched_module()` function that should work whether the `sched` service is alone or in a comma-separate list of services.

Issue #1897, opened 2018-12-28
valgrind: Invalid read of size 1 (i686)


garlick:
On Ubuntu i686 platform, valgrind test is failing on current master (v0.10.0-814-gd59ce0a4)
```
./t5000-valgrind.t --verbose
expecting success: 
	run_timeout 120 \
	libtool e flux ${VALGRIND} \
		--tool=memcheck \
		--leak-check=full \
		--gen-suppressions=all \
		--trace-children=no \
		--child-silent-after-fork=yes \
		--num-callers=30 \
		--leak-resolution=med \
		--error-exitcode=1 \
		--suppressions=$VALGRIND_SUPPRESSIONS \
		${BROKER} --shutdown-grace=4 ${VALGRIND_WORKLOAD} 10

==18884== Memcheck, a memory error detector
==18884== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==18884== Using Valgrind-3.13.0 and LibVEX; rerun with -h for copyright info
==18884== Command: /home/garlick/proj/flux-core/src/broker/.libs/flux-broker --shutdown-grace=4 /home/garlick/proj/flux-core/t/valgrind/valgrind-workload.sh 10
==18884== 
==18884== Thread 5:
==18884== Invalid read of size 1
==18884==    at 0x4883F75: ev_run (ev.c:3634)
==18884==    by 0x4850086: flux_reactor_run (reactor.c:140)
==18884==    by 0x4859884: flux_future_wait_for (future.c:483)
==18884==    by 0x48599E0: flux_future_get (future.c:510)
==18884==    by 0x877C6B5: op_event_subscribe (shmem.c:120)
==18884==    by 0x484F07D: flux_event_subscribe (handle.c:774)
==18884==    by 0x113822: register_event (modservice.c:201)
==18884==    by 0x113822: modservice_register (modservice.c:238)
==18884==    by 0x111679: module_thread (module.c:144)
==18884==    by 0x4A0A3BC: start_thread (pthread_create.c:463)
==18884==    by 0x4B87E15: clone (clone.S:108)
==18884==  Address 0x7f72c5f is on thread 5's stack
==18884==  1 bytes below stack pointer
==18884== 
{
   <insert_a_suppression_name_here>
   Memcheck:Addr1
   fun:ev_run
   fun:flux_reactor_run
   fun:flux_future_wait_for
   fun:flux_future_get
   fun:op_event_subscribe
   fun:flux_event_subscribe
[...]

PR #6224, opened 2024-08-20
libev: fix memory fence on i386 under valgrind

trws:
problem: libev accesses one byte past the end of the stack in its custom assembly memory fence.  I have no idea why, it doesn't help, and causes valgrind to throw a warning every single time it gets invoked.

solution: Apply this patch identified for debian upstream in 2017 so it
will stop doing that: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=850861

### DIFF
--- a/src/common/libev/ev.c
+++ b/src/common/libev/ev.c
@@ -791,7 +791,7 @@ struct signalfd_siginfo
   #if ECB_GCC_VERSION(2,5) || defined __INTEL_COMPILER || (__llvm__ && __GNUC__) || __SUNPRO_C >= 0x5110 || __SUNPRO_CC >= 0x5110
     #define ECB_MEMORY_FENCE_RELAXED __asm__ __volatile__ ("" : : : "memory")
     #if __i386 || __i386__
-      #define ECB_MEMORY_FENCE         __asm__ __volatile__ ("lock; orb $0, -1(%%esp)" : : : "memory")
+      #define ECB_MEMORY_FENCE         __asm__ __volatile__ ("lock; orb $0, 0(%%esp)"  : : : "memory")
       #define ECB_MEMORY_FENCE_ACQUIRE __asm__ __volatile__ (""                        : : : "memory")
       #define ECB_MEMORY_FENCE_RELEASE __asm__ __volatile__ (""                        : : : "memory")
     #elif ECB_GCC_AMD64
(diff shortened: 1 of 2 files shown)


trws:
Quite possibly, maybe they did something different?  The warning is this:
```
==3012== Thread 7 content-sqlite:
==3012== Invalid read of size 1
==3012==    at 0x48996AF: ev_run (ev.c:4091)
==3012==    by 0x48996AF: ev_run (ev.c:4021)
==3012==    by 0x486CE4E: flux_reactor_run (reactor.c:124)
==3012==    by 0x7F68CA8: mod_main (content-sqlite.c:945)
==3012==    by 0x116A4C: module_thread (module.c:225)
==3012==    by 0x4A5E35C: start_thread (pthread_create.c:442)
==3012==    by 0x4AF7425: clone (clone.S:107)
==3012==  Address 0x8922e1f is on thread 7's stack
==3012==  1 bytes below stack pointer
==3012==
{
   <insert_a_suppression_name_here>
   Memcheck:Addr1
   fun:ev_run
   fun:ev_run
   fun:flux_reactor_run
   fun:mod_main
   fun:module_thread
   fun:start_thread
   fun:clone
}
```

It throws on every single `ev_run`.   I'm not sure why they wouldn't want the patch. At the end of the day the right instruction to use here is an mfence, just like amd64 uses.  I imagine this is done this way as an optimization because that's a stronger fence than needed, but doing an atomic operation on memory that's by definition unspecified is not good practice regardless.

garlick:
Oh this is a dup of #1897.  There I did locate the author's claim that it is a valgrind bug and added a suppression to our test suite.

trws:
It's executing an atomic operation on memory that's invalid to access at that point in the program.  The presence or absence of the lock prefix has nothing to do with it.  To be honest I find that attitude very disturbing for anyone maintaining something important that we use for concurrency control.  That said, address sanitizer allows it, and it should usually have no adverse effect.

garlick:
Yeah, this guy is pretty...different.   I'm not necessarily saying we shouldn't apply this if you feel strongly about it.

We didn't have a patch back then, and I didn't want to look up how those instructions work when the author seemed sure it was OK.  Plus it's a patch against vendored code that we'll have to reapply forever.  (That said, apparently the release we have now from 2020 is current)

I'll approve if you say so.

garlick (approved):
This should reference #1897 and remove the suppression if you want to go forward.  I'll go ahead and mark this approved.

trws:
Sounds fair to me.  I made the adjustments and will mark MWP.

PR #6225, opened 2024-08-20
mergify: use updated keys before deprecation bites

trws:
problem: mergify is deprecating some of the options we use, and removing them in a few weeks

solution: use the new versions


grondo (approved):
Thanks! I'll merge this.

PR #6226, opened 2024-08-21
doc: add dependency example

vsoch:
Problem: there is not a concrete example of using --dependency in our current documentation.
Solution: add example to man1/common/job-dependencies.rst

### DIFF
--- a/doc/man1/common/job-dependencies.rst
+++ b/doc/man1/common/job-dependencies.rst
@@ -52,5 +52,20 @@ begin-time:TIMESTAMP
 In any of the above ``after*`` cases, if it is determined that the
 dependency cannot be satisfied (e.g. a job fails due to an exception
 with afterok), then a fatal exception of type=dependency is raised
-on the current job.
+on the current job. An example of submitting a job, getting the Flux
+job identifier, and then submitting a job that depends on it (meaning
+it will wait for completion before starting) is provided below:
 
+.. code-block:: console
+  
+    # Do some long work
+    jobid=$(flux submit -N2 sleep 200)
+
+    # Submit the dependent job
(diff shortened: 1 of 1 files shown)


doc/man1/common/job-dependencies.rst
```diff
@@ -52,5 +52,20 @@ begin-time:TIMESTAMP
...
+on the current job. An example of submitting a job, getting the Flux
+job identifier, and then submitting a job that depends on it (meaning
+it will wait for completion before starting) is provided below:
 
+.. code-block:: console
+  
+    # Do some long work
+    jobid=$(flux submit -N2 sleep 200)
+
+    # Submit the job that will wait (headless)
```
grondo, line 64: Suggestion: Change this comment to simply `Submit the dependent job`.

grondo (approved):
One suggestion to clarify a comment inline.

One potential issue is that this snippet is included in all the CLI submission command manual pages, not just `flux-submit(1)`. It seems fine to me, but I'd like to get more feedback if this example should just go in an EXAMPLES section of `flux-submit(1)` directly or something. 

Otherwise, LGTM. (Of course, we could go ahead and merge this and move it later if there is confusion)

vsoch:
> I'd like to get more feedback if this example should just go in an EXAMPLES section of flux-submit(1) directly or something.

My 0.02 is that the redundancy is OK and needed. If depends is relevant for all those commands, if I search for one and then don't see it there, I don't find what I'm looking for. So my vote is to have it appear in multiple places, as it is relevant for all of them, and users will go looking for it in all of them. I do think we should eventually have a more dedicated page, but this seems OK for now.

grondo:
Sounds good to me! I'll set MWP.

Issue #6228, opened 2024-08-21
python argparse continues processing past first non-option argument


grondo:
Problem: In the current cli submission commands, it appears that Python argparse is configured to continue processing arguments past the first non-option argument. I'm not sure how we've gone so long without anyone noticing this (or perhaps it is recently broken?)
```console
$ flux submit -n1 true --n=2
usage: flux submit [OPTIONS...] COMMAND [ARGS...]
flux submit: error: ambiguous option: --n=2 could match --nodes, --ntasks
```

grondo:
And using `--` to separate doesn't work because Python actually accepts that as an argument :facepalm:
```
$ flux run --dry-run -n1 -- foo --n=2 | jq .tasks[0].command
[
  "--",
  "foo",
  "--n=2"
]
```

grondo:
Hm, maybe this is actually a Python argparse bug, because other options that aren't partial matches for argparse options work:
```
fluke2:~/git/f.git$ src/cmd/flux run -n1 --dry-run true --nt=2  | jq .tasks
[
  {
    "command": [
      "true",
      "--nt=2"
    ],
    "slot": "task",
    "count": {
      "per_slot": 1
    }
  }
]
```

Even known arguments work (as in, are ignored and left as part of `argparse.REMAINDER`):
```
$ src/cmd/flux run -n1 --dry-run true --ntasks=10  | jq .tasks
[
  {
    "command": [
      "true",
      "--ntasks=10"
    ],
    "slot": "task",
    "count": {
      "per_slot": 1
    }
  }
]
```

PR #6234, opened 2024-08-26
Fix handling of `--` used to separate options from command in `flux submit`, `run`, and `batch`

grondo:
Since Python `argparse` throws an error if an ambiguous `argparse` option is used after the last non-option argument (e.g. `flux run -n2 foo --n=2` throws an error, even though `argparse` should just ignore everything from `foo` onwards), there should be some way for users to work around this problem. However, the standard use of `--` doesn't work here, because Python then includes the `--` as part of `args.command` (in the case of `flux run` and `flux submit`) or `args.SCRIPT` in the case of `flux batch`.

This PR simply ignores `--` when it is the first entry of `args.command` or `args.SCRIPT` and adds some tests that verify the solution works. (i.e. `flux run -n2 -- foo --n=2` can be used to workaround the error noted above).

There may be some better way to support this with `argparse` (perhaps the bug is fixed in recent versions), but I couldn't find it.

### DIFF
--- a/src/bindings/python/flux/cli/base.py
+++ b/src/bindings/python/flux/cli/base.py
@@ -1232,6 +1232,11 @@ class SubmitBaseCmd(MiniCmd):
         if not args.command:
             raise ValueError("job command and arguments are missing")
 
+        #  Remove first -- from command in case user used it to separate
+        #  Flux cli options from command and options
+        if args.command[0] == "--":
+            args.command.pop(0)
+
         #  Ensure integer args are converted to int() here.
         #  This is done because we do not use type=int in argparse in order
         #   to allow these options to be mutable for bulksubmit:
--- a/src/bindings/python/flux/cli/batch.py
+++ b/src/bindings/python/flux/cli/batch.py
@@ -60,6 +60,11 @@ class BatchCmd(base.MiniCmd):
         Returns the ingested script and new argparse args Namespace.
         """
         if args.SCRIPT:
+            # Remove leading "--" in case it was used to separate flux-batch(1)
+            # options from user script and options:
+            if args.SCRIPT[0] == "--":
+                args.SCRIPT.pop(0)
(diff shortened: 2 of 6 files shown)


src/bindings/python/flux/cli/base.py
```diff
@@ -1232,6 +1232,11 @@ def init_jobspec(self, args):
         if not args.command:
```
garlick, line 1232: Commit message: s/bart/part/

Any issue to reference here?

garlick (approved):
LGTM!

Minor: maybe the tests should use something like `--badopt` rather than `--n` since the latter could match any option that begins with `n` if it is unambiguous, and there seem to be only two long options that begin with `n` in `flux run` and `flux batch`.  Not that this will change anytime soon!

grondo:
> Minor: maybe the tests should use something like --badopt rather than --n since the latter could match any option that begins with n if it is unambiguous, 

I might be misunderstanding, but use of `--badopt` doesn't trigger the bug. The option after the last non-option argument _has_ to be ambiguous or it works fine (meaning `argparse` ignores the option and includes it in `args.command` without raising an exception, for example)

grondo:
Oh, I guess you're saying we should _also_ test with `--badopt` to ensure `--` also allows separating other arguments besides just one that might be ambiguous for the Flux cli submission command `argparse`?

garlick:
Oh sorry!  I misunderstood the defect.  All good as is then!

src/bindings/python/flux/cli/base.py
grondo: Yeah, I'll just mark it as fixes #6228, since it at least gives users a workaround

grondo:
Thanks! I've set MWP here.

grondo:
Thanks @jjellio !

PR #6235, opened 2024-08-26
perilog: support direct execution of prolog/epilog on job ranks for better performance

### LINKED ISSUE #6216: `flux perilog-run` is not scalable
Problem: For large jobs, the `flux perilog-run --exec-per-rank` solution for executing `prolog` and `epilog` actions on each node of spends most of its time executing processes. It is estimated to take several minutes to run even around 5000 `flux exec` processes.

In contrast, flux-exec across 5000 ranks takes less than 1 second.

grondo:
Posting this PR as a WIP to get early feedback and to give reviewers a head start.

This PR reimplements the perilog jobtap plugin using bulk-exec in order to support direct execution of prolog/epilog on the ranks of a job instead of using `perilog-run`, which has been shown to take several minutes to start `flux exec` for a job comprising thousands of nodes.

The plugin should continue to work the same unless a new `per-rank = true` option is set, in which case the plugin uses bulk-exec to launch the defined `command` across all ranks that are part of the starting/ending job.

Support for other `flux perilog-run` features are also moved directly into the plugin, including draining failed ranks, a `timeout` parameter, and a `use-imp` boolean which replaces `command` with `["$imp", "run", "$name"]` if used.

The plugin also now supports a jobtap query callback which can be used to observe the current config, as well as active job prolog and epilogs, including which ranks are still actively running these proceses.

This will be a WIP because it still needs testing (results of which may change some of the code). 

Fixes #6216

### DIFF
--- a/doc/guide/admin.rst
+++ b/doc/guide/admin.rst
@@ -666,22 +666,20 @@ enabled.
 Adding Job Prolog/Epilog Scripts
 ================================
 
-Flux does not currently support a traditional job prolog/epilog
-which runs as root on the nodes assigned to a job before/after job
-execution. Flux does, however, support a job-manager prolog/epilog,
-which is run at the same point on rank 0 as the instance
-owner (typically user ``flux``), instead of user root.
+The current best practice for enabling per-job prolog/epilog scripts
+is to use the job-manager prolog/epilog configuration supported by the
+``perilog.so`` jobtap plugin. The perilog plugin optionally initiates a
+prolog action when a job enters the RUN state before any job shells are
+launched, and an epilog action that can be started after all job shells
+complete but before resources are returned to the scheduler.
 
-As a temporary solution, a convenience command ``flux perilog-run``
-is provided which can simulate a job prolog and epilog by executing a
-command across the broker ranks assigned to a job from the job-manager
-prolog and epilog.
+.. note::
+   The perilog plugin now supports directly executing the configured
+   prolog/epilog across all ranks of a job, which makes the previous solution
+   using ``flux perilog-run`` obsolete.
 
-When using ``flux perilog-run`` to execute job prolog and epilog, the
-job-manager prolog/epilog feature is being used to execute a privileged
-prolog/epilog across the nodes/ranks assigned to a job, via the
-flux-security IMP "run" command support. Therefore, each of these
-components need to be configured, which is explained in the steps below.
+The configuration of a prolog and epilog that runs as root requires the
+following steps:
 
  1. Configure the IMP such that it will allow the system instance user
     to execute a prolog and epilog script or command as root.
@@ -723,23 +721,29 @@ components need to be configured, which is explained in the steps below.
          { load = "perilog.so" }
        ]
 
- 3. Configure the Flux system instance ``[job-manager.prolog]`` and
-    ``[job-manager.epilog]`` to execute ``flux perilog-run`` with appropriate
-    arguments to execute ``flux-imp run prolog`` and ``flux-imp run epilog``
-    across the ranks assigned to a job:
+ 3. Configure the Flux system instance to run a prolog and epilog across
+    the ranks of each job using the job-manager prolog/epilog as supported
+    by the ``perilog.so`` plugin. See :man5:`flux-config-job-manager` for
+    a detailed description of supported ``perilog.so`` configuration.
+
+    .. note::
+        A command need not be specified in ``[job-manager.prolog]`` or
+        ``[job-manager.epilog]`` when ``exec.imp`` (the path to the IMP)
+        is configured. In this case, the default is to run the prolog/epilog
+        defined in IMP configuration via ``flux-imp run [prolog|epilog]``.
+
+    For example:
 
     .. code-block:: toml
 
        [job-manager.prolog]
-       command = [
-          "flux", "perilog-run", "prolog",
-          "-e", "/usr/libexec/flux/flux-imp,run,prolog"
-       ]
+       per-rank = true
+       # timeout = "30m"
+
        [job-manager.epilog]
-       command = [
-          "flux", "perilog-run", "epilog",
-          "-e", "/usr/libexec/flux/flux-imp,run,epilog"
-       ]
+       per-rank = true
+       # timeout = "0"
+
 
  4. (optional) If log messages from the prolog or epilog are filling
     up the broker logs, a list of ignore patterns may be added via
@@ -751,15 +755,6 @@ components need to be configured, which is explained in the steps below.
 
       [job-manager]
       perilog.log-ignore = [ ".*Xauth.*", "^foo:.*debug" ]
-
-Note that the ``flux perilog-run`` command will additionally execute any
-scripts in ``/etc/flux/system/{prolog,epilog}.d`` on rank 0 by default as
-part of the job-manager prolog/epilog. Only place scripts here if there is
-a need to execute scripts as the instance owner (user `flux`) on a single
-rank for each job. If only traditional prolog/epilog support is required,
-these directories can be ignored and should be empty or nonexistent.
-To run scripts from a different directory, use the ``-d, --exec-directory``
-option in the configured ``command``.
 
 See also: :man5:`flux-config-job-manager`,
 :security:man5:`flux-config-security-imp`.
--- a/doc/man5/flux-config-job-manager.rst
+++ b/doc/man5/flux-config-job-manager.rst
@@ -9,6 +9,10 @@ DESCRIPTION
 The Flux **job-manager** service may be configured via the ``job-manager``
 table, which may contain the following keys:
 
+Note that this set of keys may be extended by loaded jobtap plugins.
+For details pertaining to some plugins distributed with flux see
+:ref:`plugin_specific_keys` below.
+
 
 KEYS
 ====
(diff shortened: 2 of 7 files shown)


garlick:
I've built packages based on the PR pushed last night (2ff65286613f8afe96a267476a82c0f4bcf8b5ad) and have installed them on my test cluster.  I'll try to do some testing this morning.

grondo:
Thanks! This morning I pushed an update to fix the error logging. I'll address those CodeQL alerts as well and then start in on writing some sharness tests.

grondo:
FYI - just pushed a first cut at some testing.

src/modules/job-manager/Makefile.am
```diff
@@ -109,6 +109,7 @@ plugins_alloc_check_la_LDFLAGS = \
 plugins_perilog_la_SOURCES = \
 	plugins/perilog.c
 plugins_perilog_la_LIBADD = \
+	$(top_builddir)/src/common/libsubprocess/libsubprocess.la \
```
wihobbs, line 112: Nit: commit message spelling "useful" "efficient"

src/modules/job-manager/plugins/perilog.c
```diff
@@ -69,7 +70,9 @@ extern char **environ;
 static struct perilog_conf {
     double prolog_kill_timeout;/*  Time between SIGTERM/SIGKILL for prolog  */
     flux_cmd_t *prolog_cmd;    /*  Configured prolog command                */
+    bool prolog_per_rank;      /*  Execute prolog on each job rank          */
```
wihobbs: Commit message "per-task" should be "per-rank"? Sorry if I'm misreading that, I just don't see any references to tasks in the code.

src/modules/job-manager/plugins/perilog.c
```diff
@@ -91,6 +91,7 @@ struct perilog_proc {
     flux_jobid_t id;
     bool prolog;
     bool canceled;
+    bool timedout;
```
wihobbs, line 106: Commit message seems truncated?

wihobbs (commented):
Just commit nits. :-)

src/modules/job-manager/plugins/perilog.c
grondo: Just an extra stray line I think.

grondo:
Thanks @wihobbs! That's helpful. I've fixed the commit messages and will force-push.

src/modules/job-manager/plugins/perilog.c
```diff
@@ -327,6 +330,16 @@ static struct bulk_exec_ops ops = {
...
 };
 
+static struct idset *ranks_from_R (json_t *R)
+{
+    struct idset *ranks;
+    struct rlist *rl = rlist_from_json (R, NULL);
+    if (!rl || !(ranks = rlist_ranks (rl)))
+        return NULL;
+    rlist_destroy (rl);
+    return ranks;
```
garlick, line 621: Leaks `rl` on `rlist_ranks()` failure

src/modules/job-manager/plugins/perilog.c
```diff
@@ -94,18 +94,14 @@ struct perilog_proc {
```
garlick, line 115: commit message: s/doubel/double/

src/modules/job-manager/plugins/perilog.c
```diff
@@ -363,9 +359,14 @@ static int run_command (flux_plugin_t *p,
...
 
+    if (!(proc = perilog_proc_create (p, id, prolog))) {
+        flux_log_error (h, "%s: proc_create", prolog ? "prolog" : "epilog");
+        goto error;
+    }
+
     if (flux_cmd_setenvf (cmd, 1, "FLUX_JOB_ID", "%s", idf58 (id)) < 0
         || flux_cmd_setenvf (cmd, 1, "FLUX_JOB_USERID", "%u", userid) < 0) {
-        flux_log_error (h, "%s: flux_cmd_create", prolog ? "prolog" : "epilog");
+        flux_log_error (h, "%s: flux_cmd_create", perilog_proc_name (proc));
```
garlick: Minor: mention the switch to `perilog_proc_name()` in the commit description or perhaps split the commit

src/modules/job-manager/plugins/perilog.c
```diff
@@ -86,10 +86,12 @@ struct perilog_proc {
     flux_plugin_t *p;
```
garlick, line 101: commit message: s/prolgo/prolog/

src/modules/job-manager/plugins/perilog.c
```diff
@@ -233,10 +236,121 @@ static void emit_finish_event (struct perilog_proc *proc,
...
+            flux_future_t *f;
+
+            /* Drain the set of ranks that failed the prolog/epilog. If the
+             * drain RPC is successful, then wait for the response before
+             * emitting the "prolog/epilog-finish" event. O/w, resources could
+             * be freed and handed out to new jobs before they are drained.
+             */
+            if ((f = drain_failed_ranks (proc))
+                 && flux_future_then (f, -1., drain_failed_cb, proc) == 0)
+                return;
```
garlick, line 515: If perilog is unloaded with this future pending, seems like the job manager reactor could end up calling a bad address?

Should the future be part of the proc struct so it can be destroyed if it is in progress when perilog is unloaded?

src/modules/job-manager/plugins/perilog.c
```diff
@@ -866,6 +916,8 @@ static int conf_init (flux_plugin_t *p, struct perilog_conf *conf)
     json_t *log_ignore = NULL;
     int prolog_per_rank = 0;
     int epilog_per_rank = 0;
+    const char *prolog_timeout = "30m";
```
garlick: Is there a way to disable the prolog timeout?  Seems like setting it to 0 would cause it to trigger immediately.  Maybe that should be special cased to "disabled"?

src/modules/job-manager/plugins/perilog.c
```diff
@@ -191,27 +192,41 @@ static void emit_finish_event (struct perilog_proc *proc,
...
+            else if (WIFSIGNALED (status) || code > 128) {
                 sig = WIFSIGNALED (status) ? WTERMSIG (status) : code - 128;
                 rc = asprintf (&errmsg,
-                               "prolog killed by signal %d%s",
+                               "prolog killed by signal %d on %s (rank %s)",
                                sig,
-                               sig == SIGTERM ?
-                               " (timeout or job canceled)" :
-                               "");
+                               hosts,
```
garlick, line 357: `hosts` could be null if both `flux_hostmap_lookup()` and `strdup()` fail

src/modules/job-manager/plugins/perilog.c
```diff
@@ -105,6 +112,95 @@ static void timeout_cb (flux_reactor_t *r,
...
+            goto fail;
+    }
+    return cmd;
+fail:
+    flux_cmd_destroy (cmd);
+    return NULL;
+}
+
+
+static struct perilog_procdesc * perilog_procdesc_create (json_t *o,
```
garlick, line 156: extra space bewteen `*` annd function name

garlick (commented):
Nice!  There's some great changes here and this will probably have a big impact at scale.

I added some comments while going through the diffs sequentially but I think possibly some of my comments are off b/c this is still a WIP and includes some incremental development.  Ignore those please :-)

src/modules/job-manager/plugins/perilog.c
```diff
@@ -275,21 +285,27 @@ static bool perilog_log_ignore (struct perilog_conf *conf, const char *s)
...
+    if (!len
+        || !(proc = bulk_exec_aux_get (bulk_exec, "perilog_proc"))
+        || !(h = flux_jobtap_get_flux (proc->p)))
         return;
-    }
-    if (len && !perilog_log_ignore (&perilog_config, s)) {
+
+    /* Copy data to NUL terminated buffer */
+    memcpy (buf, data, len);
+    buf [len] = '\0';
```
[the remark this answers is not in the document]
grondo: It is only passed to the log function after being checked if the line matches (or doesn't match) one of the log-ignore regexes. Since this function requires a NUL terminated buffer, it has to be copied.

grondo:
The incremental develepment will probably stay and reflects how the features were developed. It would unfortunately be very time consuming to squash some of the development together, so it is either this or one big "rewrite perilog plugin" commit. I'm fine either way.

garlick:
Ok by me if the incremental is intended.

grondo:
I was waiting on removal of WIP once we get some feedback that the overall approach is acceptable. I was waiting before writing/updating any documentation to make sure everyone is fine with the configuration and other external design parameters. But as far as I'm concerned I could remove WIP now as I didn't plan on doing much more work besides testing and fixing any fallout from that.

garlick:
It seems like a great improvement to me and shouldn't be controversial
since using the new functionality requires a config change to opt in.

src/modules/job-manager/plugins/perilog.c
```diff
@@ -275,21 +285,27 @@ static bool perilog_log_ignore (struct perilog_conf *conf, const char *s)
...
-                        perilog_proc_name (proc),
-                        stream);
+    flux_t *h;
+    struct perilog_proc *proc;
+    char buf[len+1]; /* bulk_exec output is not NUL terminated */
+
+    if (!len
+        || !(proc = bulk_exec_aux_get (bulk_exec, "perilog_proc"))
+        || !(h = flux_jobtap_get_flux (proc->p)))
         return;
```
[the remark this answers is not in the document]
grondo: will change to `if (len <= 0`...

src/modules/job-manager/plugins/perilog.c
grondo: Yeah, good catch! Note that unloading the perilog plugin with outstanding prolog/epilog will result in jobs with `{prolog,epilog}-start` events and no corresponding `{prolog,epilog}-finish` event though..

src/modules/job-manager/plugins/perilog.c
grondo: The timer watcher is not created unless `timeout > 0.0`:
```c
    if (timeout > 0.0) {
        flux_watcher_t *w;
        if (!(w = flux_timer_watcher_create (flux_get_reactor (h),
                                             timeout,
                                             0.,
                                             timeout_cb,
                                             proc))) {
```
So setting it to 0 disables the timer. You are suggesting that the literal string "disabled" should be a synonym for `0`?

Oh, I guess since `fsd_parse_duration()` is used, `inf`, `infinity` or `INF` would also work. However, this would have to be special cased in the stats, so maybe I'll just make Infinity and alias for 0, i.e. disabled (if that helps)

garlick: 0 is fine.  I must have misread the code.   It seemed like 0 went straight into the timer watcher `after` value which means time out immediately

src/modules/job-manager/plugins/perilog.c
grondo: and printf will report "(null)" -- would `"unknown"` be better.

garlick: sorry, I keep forgetting that we decided that is portably safe now.  FIne as is then.

src/modules/job-manager/plugins/perilog.c
```diff
@@ -157,7 +157,6 @@ static struct perilog_procdesc *perilog_procdesc_create (json_t *o,
 {
     struct perilog_procdesc *pd = NULL;
```
garlick, line 162: Suggestion: squash these consecutive commits down to one or two:
- 6a61e3a6a8e44c190faa85ab9eee7103d777869b perilog: add use-imp config parameter
- ad9f0c1b657c222ed74993ea540172536e88dbfe perilog: improve error message on failure
- ff1d8a6a3821490532a9860022a021dc7c4252dc  perilog: remove use-imp config

doc/man5/flux-config-job-manager.rst
```diff
@@ -83,6 +87,84 @@ release-after
...
+      rank assigned to the job.
+   timeout
+      (optional) A string value in Flux Standard Duration specifying a
+      timeout for the prolog, after which it is terminated (and a job
+      exception raised). The default prolog timeout is 30m. To disable
+      the timeout use ``0`` or ``infinity``.
+   kill-timeout
+      (optional) Floating-point number of seconds to wait after sending
+      SIGTERM to send SIGKILL to the prolog when the prolog has been
+      canceled due to timeout or exception.
```
garlick, line 128: Should `kill-timeout` be an FSD to match `timeout`?

doc/man5/flux-config-job-manager.rst
```diff
@@ -83,6 +87,84 @@ release-after
...
+   cancel-on-exception
+      (optional) A boolean indicating whether the prolog is terminated by
+      a job exception while the prolog is active. The default is true.
```
garlick, line 131: Maybe "whether a fatal job exception raised while the prolog is active terminates the prolog" would be slightly clearer?
Meh, maybe not.  Up to you.

doc/guide/admin.rst
```diff
@@ -666,22 +666,19 @@ enabled.
...
+The current best practice for enabling per-job prolog/epilog scripts
+is to use the job-manager prolog/epilog configuration supported by the
+``perilog.so`` jobtap plugin. The perilog plugin optionally initiates a
+prolog action when a job enters the RUN state before any job shells are
+launched, and an epilog action that can be started after all job shells
+complete but before resources are returned to the scheduler.
+
+The perilog plugin now supports directly executing the configured
+prolog/epilog across all ranks of a job, which makes the previous solution
+using ``flux perilog-run`` obsolete.
```
garlick, line 678: Might want to put the bit about the obsolete solution in a Note box.

garlick (approved):
This looks good, just a couple of minor comments.  I'll try to do a bit more testing this morning.

doc/man5/flux-config-job-manager.rst
grondo: It could, but is mainly there for testing and it seemed like a bit of wasted effort because then it would require extra tests, etc. However, it is pretty simple if you feel strongly about it. Perhaps more important would be a note saying this parameter is mainly meant for testing?

doc/man5/flux-config-job-manager.rst
grondo: Sure, that sounds better to me.

doc/guide/admin.rst
grondo: or should it just be removed?

src/modules/job-manager/plugins/perilog.c
grondo: Thanks for catching that, I had forgotten that incremental modification was in there.

garlick:
Changes LGTM.

garlick:
My admittedly light testing on my test cluster did not turn up any issues.  You have a green light from me :-)

grondo:
Sheesh, I found a bunch of places where my fingers typed "per-task" when my brain said "per-rank". Fixed and forced a push.

grondo:
Just as a data point, here's some timing differences between `per-rank` and `flux perilog-run` executing a prolog of `true` across an instance of size=720

`flux perilog-run`: 6.5s
```
0.055s: job.alloc {"annotations":{"sched":{"resource_summary":"rank[0-719]/core[0-47],gpu[0-7]"}}}
0.055s: job.prolog-start {"description":"job-manager.prolog"}
6.586s: job.prolog-finish {"description":"job-manager.prolog","status":0}
```

`per-rank = true`: 0.2 s
```
0.051s: job.alloc {"annotations":{"sched":{"resource_summary":"rank[0-719]/core[0-47],gpu[0-7]"}}}
0.057s: job.prolog-start {"description":"job-manager.prolog"}
0.276s: job.prolog-finish {"description":"job-manager.prolog","status":0}
0.485s: job.start
```

grondo:
Ok, I haven't found any issues either, so I'll set MWP. Thanks!

PR #6240, opened 2024-08-27
kvs: call content.flush before checkpoint

### LINKED ISSUE #6237: kvs: call `content.flush` before checkpoint
In the KVS module a checkpoint is done right before the module exits.  However `content.flush` is not called.  Thus there is a chance the checkpoint is invalid b/c it is not guanrateed that all KVS stores have been written to disk.

chu11:
Problem: When the KVS module is unloaded, a checkpoint of the root
reference is attempted.  However, a content.flush is not done
beforehand.  This could result in an invalid checkpoint reference
as data is not guaranteed to be flushed to the backing store.

Solution: Call content.flush before checkpointing.

Fixes #6237
----

I threw in a few new tests for some extra coverage of FLUX_KVS_SYNC too.

### DIFF
--- a/src/modules/kvs/kvs.c
+++ b/src/modules/kvs/kvs.c
@@ -2714,15 +2714,22 @@ error:
  */
 static int checkpoint_put (flux_t *h, const char *rootref, int rootseq)
 {
-    flux_future_t *f = NULL;
+    flux_future_t *f1 = NULL;
+    flux_future_t *f2 = NULL;
     int rv = -1;
 
-    if (!(f = kvs_checkpoint_commit (h, rootref, rootseq, 0, 0))
-        || flux_rpc_get (f, NULL) < 0)
+    /* first must ensure all content is flushed */
+    if (!(f1 = flux_rpc (h, "content.flush", NULL, 0, 0))
+        || flux_rpc_get (f1, NULL) < 0)
+        goto error;
+
+    if (!(f2 = kvs_checkpoint_commit (h, rootref, rootseq, 0, 0))
+        || flux_rpc_get (f2, NULL) < 0)
         goto error;
     rv = 0;
 error:
-    flux_future_destroy (f);
+    flux_future_destroy (f1);
+    flux_future_destroy (f2);
     return rv;
 }
 
--- a/t/t0090-content-enospc.t
+++ b/t/t0090-content-enospc.t
@@ -58,4 +58,16 @@ test_expect_success 'content flush returns error on ENOSPC' '
 	grep "content.flush: No space left on device" flush.err
 '
 
+# flux start will fail b/c rc3 will fail due to ENOSPC
+test_expect_success 'kvs sync fails due to ENOSPC' '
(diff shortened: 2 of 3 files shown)


src/modules/kvs/kvs.c
```diff
@@ -2914,15 +2914,28 @@ static int checkpoint_get (flux_t *h, char *buf, size_t len, int *seq)
...
+         * to no backing store, but checkpoint can still be done to
+         * content cache.
+         */
+        if (errno != ENOSYS)
+            goto error;
+    }
+
+    if (!(f2 = kvs_checkpoint_commit (h, NULL, rootref, rootseq, 0, 0))
+        || flux_rpc_get (f2, NULL) < 0)
         goto error;
```
garlick, line 2934: Why fall through to the checkpoint commit?  If there is no backing store then wouldn't it also fail?

chu11: this admittedly surprised me, but you can checkpoint to the content cache without it going through to the backing store.

chu11: ya know, thinking about this a bit .... do we want that behavior?  Do we want checkpoints to work with the content cache and w/o a backing store.  It seems like a nuanced corner case.  I'm not entirely sure why it was initially supported that way.  This would be for a different issue I think, as this PR solves a specific `content.flush` before "checkpoint" corner case.

garlick: If there is a hook in the content cache for the checkpoint, why don't we do the cache flush there?

garlick: Sorry I've forgotten so many details here!

chu11: hmmm.  supporting `checkpoint` w/o a backing module simply may have been a fallout from supporting the `none` backing module.  https://github.com/flux-framework/flux-core/pull/4492     perhaps checkpoint specifically should not be supported by the content cache.

garlick: I guess it allows the kvs module to be reloaded and not lose all the data when there is no backing store?

chu11: > I guess it allows the kvs module to be reloaded and not lose all the data when there is no backing store?

I do think that's the point.  But that is sort of a nuanced use case, possibly only useful in testing?

chu11: > If there is a hook in the content cache for the checkpoint, why don't we do the cache flush there?

That's not a bad idea.  Lemme look into that, although it'd be for a different PR b/c `FLUX_KVS_SYNC` does two calls and that would have to be collapsed down into one as well.

chu11:
making WIP, feel #6251 should come first now

chu11:
re-pushed, PR was built on top of #6255 and re-worked logically

chu11:
Now that #6255 is in, removing WIP from this.

garlick (approved):
LGTM!

PR #6241, opened 2024-08-27
housekeeping: deprecate use-systemd-config option

garlick:
In an offline discussion with @grondo about his pr #6235, we noted that housekeeping's `use-systemd-unit` key is misleading.  It's the IMP config that decides whether or not systemd is used to run the housekeeping script.  Really this is just selecting a comand of `imp run housekeeping`.

This changes the logic to use the above command if the housekeeping `command` key is undefined.
Note this means the housekeeping config table could be empty and still valid (housekeeping enabled).

If `use-systemd-unit` is present, it is a config error.   I could make it accept it and issue a deprecation warning if we think that's better.

### DIFF
--- a/doc/man5/flux-config-job-manager.rst
+++ b/doc/man5/flux-config-job-manager.rst
@@ -70,13 +70,9 @@ HOUSEKEEPING
 ============
 
 command
-  (optional) An array of strings specifying the housekeeping command. Either
-  ``command`` or ``use-systemd-unit`` must be specified.
-
-use-systemd-unit
-  (optional) A boolean value indicating whether to run the flux-housekeeping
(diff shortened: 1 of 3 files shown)


grondo:
> If use-systemd-unit is present, it is a config error. I could make it accept it and issue a deprecation warning if we think that's better.

That might be better for the sysadmins who would otherwise have to manage a version dependent ansible config during the transition to the newest version. However, if that's simple I agree it is best to just remove the option right away.

garlick:
Fixed that and forced a push.  Now a deprecation warning is logged but the old key is otherwise ignored.

grondo (approved):
LGTM!

garlick:

PR #6246, opened 2024-08-30
flux-hostlist: change default source from `local` to `stdin`

### LINKED ISSUE #6244: Suggestion: change flux hostlist default behavior to read from stdin
Reference - https://github.com/flux-framework/flux-core/issues/6243

Current functionality is that `flux hostlist` reads from the enclosing flux instance, with `flux hostlist -` required for piping input.  Could potentially preserve the current behavior with `flux hostlist -a` or similar.

grondo:
This PR addresse #6244 by changing `flux hostlist` with no args to read from `stdin` instead of `local` (which uses the enclosing instance or job if `FLUX_JOB_ID` is set).

This will break existing usage of `flux hostlist` where it is used, because the command will now hang instead of returning the enclosing hostlist. However, the current design choice was pretty bad and it is unlikely that this command has that much use so far, so it is probably better to just fix it now.

In order to mitigate the possibility of hangs with this change, a timeout was added to the read of `stdin` of 15s. After the timeout the command exits with a warning. Hopefully this workaround can be removed in the next release.

The existing behavior can be replicated with `flux hostlist local` or a new `flux hostlist -l, --local` option.

Fixes #6244

### DIFF
--- a/src/cmd/flux-hostlist.py
+++ b/src/cmd/flux-hostlist.py
@@ -11,6 +11,7 @@
 import argparse
 import logging
 import os
+import select
 import sys
 
 import flux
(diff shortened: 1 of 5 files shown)


src/cmd/flux-hostlist.py
```diff
@@ -31,7 +31,8 @@
 stdin, '-'   read a list of hosts on stdin
```
garlick, line 32: commit message for 18e965050a38bc9a56e4477fd3660c6c30735e47: s/excplicitly/explicitly/

garlick (approved):
LGTM!  One commit message typo.

src/cmd/flux-hostlist.py
grondo: Thanks! Fixed that and will set MWP.

PR #6249, opened 2024-08-31
run epilog even if job prolog fails or is canceled

### LINKED ISSUE #6055: Run administrative epilog even if job is canceled before starting
If the prolog action described in https://github.com/flux-framework/flux-coral2/issues/166 goes into production, it will make changes the compute nodes which must be undone by a matching epilog action. However, if the job is canceled or fails before the application begins to execute, the epilog action doesn't run. This leaves the potential for the node to be left in a bad state where the changes made by the prolog are never undone by the epilog.

grondo:
This PR addresses #6055 by triggering the job epilog when the prolog completes if it fails, e.g. due to timeout or if it is canceled due to a job exception.

Fixes #6055

### DIFF
--- a/doc/man5/flux-config-job-manager.rst
+++ b/doc/man5/flux-config-job-manager.rst
@@ -132,10 +132,11 @@ prolog
       while the prolog is active terminates the prolog. The default is true.
 
 epilog
-   (optional) Table of configuration for a job-manager epilog. If configured,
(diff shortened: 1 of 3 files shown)


jameshcorbett (approved):
LGTM thanks!

grondo:
Setting MWP here, Thanks!

PR #6250, opened 2024-08-31
progress: ensure we don't access outside style

trws:
problem: in some corner case the equation produces an offset of exactly len(style) rather than a valid offset

solution: when that happens, set it to -1 to access the last element


grondo (approved):
Great find! Thanks!

PR #6252, opened 2024-08-31
support `resource.scheduling` in config TOML to amend configured R with a scheduling key

### LINKED ISSUE #6245: config: add option for amending `Rv1` from `resource.config` with `scheduling` key from file
Currently, in order to configure Flux with an _R_ augmented with JGF via the `scheduling` key, a generated _R_ must be used instead of the `resource.config` array. Since _R_ is difficult to work with directly, admins have been forced to store a `resource.toml` config separately and use `flux R parse-config` to generate an Rv1 which can then be amended with JGF.

It might be better to add a config key that points to the JGF in a file, and have Flux attach that to the _R_ generated from `resource.config` (e.g. maybe just `resource.scheduling`?)

grondo:
This PR adds a new `resource.scheduling` key in the Flux config TOML that, when set,  amends the configured _R_ with an externally defined `scheduling` key. Notably, this allows use of JGF with the `resource.config` array, which currently is not possible. It also would allow storing _R_ and JGF in separate files if this is desired.

It is an error if `resource.scheduling` is set but neither `resource.path` or `resource.config` are.

Fixes #6245

### DIFF
--- a/doc/man5/flux-config-resource.rst
+++ b/doc/man5/flux-config-resource.rst
@@ -54,6 +54,10 @@ config
      hosts = "test[90-100]"
      properties = ["debug"]
 
+scheduling
+   (optional) Set the path to a file stored as JSON which will be used
+   to amend the configured R with a RFC 20 ``scheduling`` key.
+
 exclude
    (optional) A string value that defines one or more nodes to withhold
    from scheduling, either in RFC 22 idset form, or in RFC 29 hostlist form.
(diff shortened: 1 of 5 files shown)


grondo:
@jameshcorbett: how does this solution sound?

jameshcorbett (approved):
Design should work very well and code LGTM. Thanks!

grondo:
Thanks! I've set MWP.

PR #6258, opened 2024-09-04
validator: allow configurable minimum job size in the require-instance validator plugin

grondo:
@trws had suggested in a meeting that the `require-instance` validator might be more useful if it had a configurable minimum job size to which the validator would apply. For example, a cluster might be configured to allow any jobs up to 16 nodes, but above that, `flux alloc` or `flux batch` should be used.

This PR adds new options to the `require-instance` validator `--require-instance-minnodes=N` and `require-instance-mincores=N` which, if used, allow jobs smaller than these numbers to be ignored by the plugin. To avoid a case where only `--require-instance-minnodes` is used, and all jobs that only specify cores are allowed to run, a very conservative default of 16 cores per node is assumed if `--require-instance-mincores` is not set.

It would be nice if this configuration could be set in the config file, but deciding where that goes (the job-ingest module "owns" everything under `[ingest]` and throws an error if extra keys are used), so that can be saved for later. (as could other improvements, like making the plugin per-queue, etc.)

Just throwing this up now to see what people think.

### DIFF
--- a/src/bindings/python/flux/job/validator/plugins/require-instance.py
+++ b/src/bindings/python/flux/job/validator/plugins/require-instance.py
@@ -22,19 +22,94 @@ not directly via `flux broker` or `flux start`.
 import errno
 from os.path import basename
 
+from flux.job import JobspecV1
 from flux.job.validator import ValidatorPlugin
 
 
+class JobSize:
+
+    def __init__(self, nnodes=0, ncores=0):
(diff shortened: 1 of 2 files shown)


garlick (approved):
Seems useful to me!

grondo:
Thanks! I've set MWP here.

PR #6260, opened 2024-09-05
content: flush before checkpointing

### LINKED ISSUE #6242: content: call `content.flush` within `content-checkout.put`
Per comment / idea in #6240

Pro: no need to require user to call before calling `content-checkpoint.put`.  If backing store isn't loaded, the content module will know not to call `content.flush`.

Con: only work if user doesn't use BYPASS flag.  but presumably user will know what they are doing if they use BYPASS.

chu11:
Problem: Before checkpointing, users need to remember to call content.flush, to ensure data has been flushed to the backing store. It is easy to forget this.

Within the content module, call content.flush before checkpointing.

Fixes #6242

----

Last in the chain, this is built on top of #6255 and then #6240.  So this one is last, setting WIP for now.

### DIFF
--- a/src/modules/kvs/kvs.c
+++ b/src/modules/kvs/kvs.c
@@ -1016,20 +1016,6 @@ static void kvstxn_apply (kvstxn_t *kt)
         assert (wait_get_usecount (wait) > 0);
         goto stall;
     }
-    else if (ret == KVSTXN_PROCESS_SYNC_CONTENT_FLUSH) {
-        /* N.B. future is managed by kvstxn, should not call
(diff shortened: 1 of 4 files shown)


garlick (approved):
This seems good, if you want to remove the WIP.

chu11:
> This seems good, if you want to remove the WIP.

Are you inherently approving #6240 as well?  That one comes before this one :-)

garlick:
Oh, I thought that one was already in!  Sorry, will look at that now.

I should have said "yes!" because I reviewed those commits as part of this one.

PR #6267, opened 2024-09-06
python: add `conf_get()` convenience method to `flux.Flux` handle

grondo:
As part of some other work, it seemed like it would be nice to have a convenient method to get values from current Flux configuration. This is one possible simple approach.

This PR adds a `flux.Flux.conf_get()` method:

```python
    def conf_get(self, key=None, default=None, update=False):
````

On first access or if `update=True`, the current config is synchronously fetched from the broker and cached. If `key` is `None`, then the entire object is returned, otherwise `key` is a dotted string representation of a key from the config to return. If `key` does not exist, then `default` is returned.

This should make it easy for Python components to use the Flux configuration.

As I said, this is just one approach, but thought I'd throw it up as a PR for comments.


garlick (approved):
Works for me!

grondo:
Thanks!

PR #6268, opened 2024-09-06
limit output to 1G in single user instances

### LINKED ISSUE #6256: flux-job: flux_kvs_lookup_get: Value too large for defined data type
A user is seeing this error from `flux run` in a subinstance with a job that may generate a lot of output. Not sure we've seen this one before.

```
job-info.err[0]: info_lookup_continuation: flux_kvs_lookup_get: Value too large for defined data type
ob-info.err[0]: main_namespace_lookup_continuation: flux_rpc_get_unpack: Value too large for defined data type
flux-job: flux_job_event_watch_get: Value too large for defined data type
```

chu11:
Problem: The KVS has a size limit of INT_MAX for when returning kvs values.  This limit can be exceeded by a job's standard output because it is continually appended and the total size is not yet tracked by the KVS.  When reading the output later, such as via `flux job attach`, this can lead to EOVERFLOW errors.

Solution: For a single user instance, default to a maximum standard output of 1G instead of "unlimited".  1G should provide a practical maximum for most users and encourage them to send standard output to a file if they want to save excess standard output.  If desired, the value can still be overwritten via the "output.limit" setting.

Fixes #6256

----

Any thoughts of if the limit should be higher? lower?   I picked 1G mostly b/c I disliked the limit of "2G - 1".  

I also timed things and on my single user instance reading the 1G stdout via `flux job attach` took about 140 seconds.  2 minutes seems to be just beyond that "annoyance limit".  I could also see making the max 500M so that the `flux job attach` is more around 1 minute.

### DIFF
--- a/t/t2606-job-shell-output-redirection.t
+++ b/t/t2606-job-shell-output-redirection.t
@@ -368,8 +368,17 @@ test_expect_success LONGTEST 'job-shell: no truncation at 10MB for single-user j
 	flux run cat 10M+ >10M+.output &&
 	test_cmp 10M+ 10M+.output
 '
-test_expect_success 'job-shell: invalid output.limit string is rejected' '
+test_expect_success 'job-shell: invalid output.limit string is rejected (text)' '
 	test_must_fail flux run -o output.limit=foo hostname
+'
+test_expect_success 'job-shell: invalid output.limit string is rejected (zero)' '
+	test_must_fail flux run -o output.limit=0 hostname
+'
+test_expect_success 'job-shell: invalid output.limit string is rejected (big num)' '
+	test_must_fail flux run -o output.limit=4000000000 hostname
+'
+test_expect_success 'job-shell: invalid output.limit string is rejected (big num suffix)' '
+	test_must_fail flux run -o output.limit=4G hostname
 '
 test_expect_success 'job-shell: output.mode=append works' '
 	flux bulksubmit --watch --output=append.out \
--- a/doc/man1/flux-shell.rst
+++ b/doc/man1/flux-shell.rst
@@ -326,9 +326,9 @@ plugins include:
 .. option:: output.limit=SIZE
 
    Truncate KVS output after SIZE bytes have been written. SIZE may
-   be a floating point value with optional SI units k, K, M, G. A value of
-   0 is considered unlimited. The default KVS output limit is 10M for jobs
-   in a multi-user instance or unlimited for single-user instance jobs.
+   be a floating point value with optional SI units k, K, M, G.  The maximum
+   value is 1G.  The default KVS output limit is 10M for jobs
+   in a multi-user instance or 1G for single-user instance jobs.
    This value is ignored if output is directed to a file.
 
 .. option:: output.{stdout,stderr}.path=PATH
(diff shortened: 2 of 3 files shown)


garlick:
Should we make it an error if the user tries to configure a limit that may trigger #6256?

I'd suggest in this PR to then just make the maximum safe value the default for user instances.  Another PR (or at least another commit) could lower it if we think that's a good idea.

grondo:
1G is a lot, but the common case is `flux run` in a batch job where the intent is to copy stdout/err to the same location as the batch script stdout/err, and it would be really good to not have a long running job truncate the final lines of job output with "the answer" if at all possible. Feels like 1G is a good compromise.

 wonder if it would make sense to issue a warning at the end of jobs that generate lots of output (say >100M), and suggest redirection to a file?

chu11:
> Should we make it an error if the user tries to configure a limit that may trigger https://github.com/flux-framework/flux-core/issues/6256?

Doing this means “unlimited” is no longer allowed.  I guess that makes sense though, if anything greater than 2G would lead to EOVERFLOW anyways.

chu11:
> wonder if it would make sense to issue a warning at the end of jobs that generate lots of output (say >100M), and suggest redirection to a file?

I was thinking of something like this too.  Sounds reasonable, although perhaps in a different PR.

chu11:
> I'd suggest in this PR to then just make the maximum safe value the default for user instances. Another PR (or at least another commit) could lower it if we think that's a good idea.

Ugh, tried to change the max limit to 2G-1, testing went bad.  I realized that the output limit in the shell is the number of stdout bytes.  But of course the KVS max is w/ all of the encoding overhead.

So perhaps going with 1G is just a good safe value in general.

grondo (commented):
LGTM, but I think there needs to be a doc update (in at least `flux-shell(1)`)

chu11:
re-pushed with a good chunk of changes

- 1G limit still there
- max output.limit is 1G now
- update docs
- add some tests for invalid input

t/t2606-job-shell-output-redirection.t
```diff
@@ -368,9 +368,18 @@ test_expect_success LONGTEST 'job-shell: no truncation at 10MB for single-user j
...
 	test_must_fail flux run -o output.limit=foo hostname
 '
+test_expect_success 'job-shell: invalid output.limit string is rejected (zero)' '
+	test_must_fail flux run -o output.limit=0 hostname
+'
+test_expect_success 'job-shell: invalid output.limit string is rejected (big num)' '
+	test_must_fail flux run -o output.limit=4000000000 hostname
+'
+test_expect_success 'job-shell: invalid output.limit string is rejected (big num fsd)' '
+	test_must_fail flux run -o output.limit=4G hostname
```
garlick, line 381: Not an fsd.

garlick (approved):
LGTM!

Suggestion: for the release notes, include the new limit in the PR title, e.g. "shell: limit output to 1G in single user instances"

chu11:
oh crud, when I changed the PR title I accidentally removed the "shell: " (just put it back), may have to manually fix when we do release notes

garlick:
No worries, that's not even that bad as is.

PR #6270, opened 2024-09-08
perilog: ensure default prolog timeout matches documentation

grondo:
The documentation states that the `perilog.so` plugin prolog, if configured, has a default timeout of "30m", but this wasn't actually the case.

Fix the code to add the 30m default for the prolog, and add the missing test.

If people feel it would be better to not have defaults for either prolog/epilog I can instead fix the documentation. I could go either way.


grondo:
Thanks!

PR #6274, opened 2024-09-09
shell: output stdio output size warning

### LINKED ISSUE #6269: shell: suggest user redirect to a file if stdout large
If the final standard output is large and in the KVS, consider outputting a message to the user to redirect output to a file instead.

From discussion in #6268

chu11:
Problem: If stdout or stderr becomes large, it runs the risk of hitting the 1G limit.

Solution: If a user job outputs between 100M and 1G of stdout or stderr output a message to the user recommending to redirect to a file instead.

Fixes #6269

---

time to bike shed the actual message? :-)  Can also debate the size picked.  100M was suggested in #6268, which seems like a fine starting point to output the message.

### DIFF
--- a/src/shell/output.c
+++ b/src/shell/output.c
@@ -65,6 +65,8 @@
 #define SINGLEUSER_OUTPUT_LIMIT "1G"
 #define MULTIUSER_OUTPUT_LIMIT  "10M"
 #define OUTPUT_LIMIT_MAX        1073741824
+/* 104857600 = 100M */
+#define OUTPUT_LIMIT_WARNING    104857600
 
 enum {
     FLUX_OUTPUT_TYPE_TERM = 1,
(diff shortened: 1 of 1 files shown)


grondo:
I just think the warning message should include the number of bytes written (encoded by `encode_size()` - the user may realize there was an error if their application generated 250M of stdout when they weren't expecting it, but just saying "too large" makes it seem arbitrary. Also, I think the `stdout:` `stderr:` prefixes in the error are redundant with the `output to stdout/stderr...` maybe drop one of those.

garlick:
Also "redirect to file..." is a little terse.  One might even wonder if automatic redirection occurred?

Maybe be specific e.g. "consider using -o output=FILENAME"?

grondo:
> Maybe be specific e.g. "consider using -o output=FILENAME"?

I like this idea. (However, note it should be `--output=FILENAME`)

chu11:
> Maybe be specific e.g. "consider using -o output=FILENAME"?

I considered that, but this message is technically coming from the shell.  Should the shell be knowledgeable of `flux-run` and friend's command line options?   B/c people can launch jobs other ways too.

grondo:
Good point, but having a probably very rare warning message be most likely to be helpful is probably a good optimization here IMO. If you're worried about that you could put the submission cli option in parens with `(e.g. use --output=FILE)`

chu11:
ok re-pushed with new output like

```
14.224s: flux-shell[0]:  WARN: output: large output of 190.9256M to stdout, redirect to file to avoid hitting output limits (e.g. use --output=FILE)
```

garlick:
> WARN: output: large output of 190.9256M to stdout, redirect to file to avoid hitting output limits (e.g. use --output=FILE)

That seems better.  Does it imply this is fixable on the fly though?  And the log message has the word output twice and stdout, which seems a bit redundant.

How about
> WARN: output: high stdout volume (190.9256M),  consider redirecting with  --output=FILE next time

chu11:
re-pushed, how about

```
10.286s: flux-shell[0]:  WARN: output: high stdout volume (133.64792M), consider redirecting to a file next time (e.g. use --output=FILE)
```

i wanted to keep the option in parens b/c users could launch a job in non flux-run/submit ways

garlick (approved):
LGTM!

PR #6277, opened 2024-09-10
add tbon.interface-hint broker attribute / configuration key with CIDR network support

### LINKED ISSUE #5754: make preferred TBON network configurable by address
Problem: although the network addresses used in the system instance are configurable, the only way to force a subinstance to use a specific network is to set the FLUX_IPADDR_INTERFACE environment variable.

Sys admins may need a way to tell all Flux instances to use a particular network.  Ideally this would be by setting an attribute in the system instance config.  If configured, this setting would be inherited by progeny at any depth (e.g. via the batch/alloc config).

@grondo pointed out that CIDR notation to identify a network would be better ergonomics than interface names which are pretty strange and not even deterministic on some system configurations.

garlick:
Problem:  starting an instance with its overlay assigned to a particular network is tricky if you network interface names are unpredictable.  And environment variables are a hacky way to influence the broker's choice.

The broker chooses a network address to bind to (and to exchange with PMI peers) using a heuristic usually based on the default route.  You can set environment variables to choose the address associated with the hostname instead, or a specific interface by name.

This adds
- the option of specifying the interface by network address (ipv4  CIDR, e.g. 10.0.2.0/24)
- a new broker attribute `tbon.interface-hint` which can have the values of `default-route` (the default), `hostname` (as described above), or name that can be either an interface name or a network address
- a TOML config key of the same name

Fixes #5754

### DIFF
--- a/doc/man5/flux-config-tbon.rst
+++ b/doc/man5/flux-config-tbon.rst
@@ -73,6 +73,23 @@ child_rcvhwm
    This configured value may be overridden by setting the ``tbon.child_rcvhwm``
    broker attribute.
 
+interface-hint
+   When the broker's bind address is not explicitly configured via
+   :man5:`flux-config-bootstrap`, it is chosen dynamically, influenced by
+   one of the following hints:
+
+   default-route
+     The address associated with the default route (the default hint).
+   hostname
+     The address associated with the system hostname.
+   *interface*
+     The address associated with the named network interface, e.g. ``enp4s0``
+   *network*
+     The address associated with the first interface that matches the
+     network address in CIDR form, e.g. ``10.0.2.0/24``.  NOTE: IPv6
+     network addresses are not supported at this time.
+
+
 EXAMPLE
 =======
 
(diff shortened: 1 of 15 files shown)


grondo (approved):
LGTM!

Unrelated to this PR (reproduced on current master), if the broker is unable to find a matching ip addr (e.g. no addr matches the provided CIDR) an error is printed, but the instance hangs until the doom plugin timer goes off:

```
lt-flux-broker: could not find address associated with 192.168.1.1/24
lt-flux-broker: bootstrap failed
flux-shell[0]: FATAL: doom: rank 0 on host corona211 exited and exit-timeout=30s has expired
```

If an instance was being launched with `exit-timeout=none`, then I assume this would result in a hang until the allocation expired.

garlick:
Thanks, fixed that style issue and rebased on current master.

Good catch on the hang.   It seems like the broker should be calling PMI_Abort() in that case but the upmi abstraction is not yet plumbed for it, so it's probably a separate PR.  I opened #6278 on that one.

Will set MWP.

garlick:
> One other thought here: is there any way to have an address hint that says "same address as parent"?

Oh interesting.  We could even pre-populate the interface as a key in the PMI kvs.  Good idea!

Issue #4572, opened 2022-09-15
flux-exec: Error: rank 0: cat: Value too large for defined data type


garlick:
Redirecting input via `flux exec` works in the shell, but when launched inside CTI, I'm getting the errors
```
2022-09-15T14:06:03.224712Z broker.err[0]: channel buffer error: rank = 0 pid = 13014, stream = stdin, len = 1048576: Success
flux-exec: Error: rank 0: cat: Value too large for defined data type
2022-09-15T14:06:03.255490Z broker.err[0]: server_write_cb: lookup_pid: No such file or directory
```
I'm using Flux 0.40.0-15, it happens with `cat`, `sed`, and a minimal C program that redirects input. Haven't seen this before in CTI when launching other programs, but it could be something with the input redirection.

_Originally posted by @ardangelo in https://github.com/flux-framework/flux-core/issues/3631#issuecomment-1248153865_

ardangelo:
Update on this, I have been trying this approach again to ship files,
For small files it usually completes successfully. However, with larger files (100+ MB), the first invocation will often fail, but an immediate retry will work.

The error is different in this case,

```
cat /tmp/cti-adangelo/cti_daemonrW8WUO1.tar | flux exec -r 0 sed -n 'w /tmp/flux-71aFeA/jobtmp-0-ƒ9HCP58r2B/cti_daemonrW8WUO1.tar'
May 11 14:32:40.908869 broker.err[0]: Error writing 65536 bytes to subprocess pid 17760 stdin
May 11 14:32:40.911301 broker.err[0]: Error writing 65536 bytes to subprocess pid 17760 stdin: unknown pid
May 11 14:32:40.913095 broker.err[0]: Error writing 65536 bytes to subprocess pid 17760 stdin: unknown pid
(Repeated)
```

grondo:
Unrelated to the actual bug discussed in this issue, I'll note that @garlick developed a better method for shipping files via [flux-filemap(1)](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man1/flux-filemap.html).

This is integrated into a `stage-in` job shell option if that works in your use case. See the [flux-shell(1)](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man1/flux-shell.html) manpage for a description of the options.

ardangelo:
[...]

PR #6281, opened 2024-09-13
libsubprocess: do not spin on large lines

### LINKED ISSUE #6262: flux cannot handle very large lines
While working on something else, I noticed

`t/shell/lptest 4000000 1000` worked as a job under flux but `t/shell/lptest 5000000 1000` did not.  The latter would just hang and produce no output.

Only later did I realize I messed up.  I should have done `t/shell/lptest 1000 4000000` (the first number is the length of a line, the second is the number of lines).   So I was inadvertently creating > 4 meg lines.

The > 4 meg line probably crossed some internal buffer / boundary somewhere, leading to problems.  But no error messages here showing up.

chu11:
Problem: In several cases, libsubprocess hangs/spins can occur if the internal output buffer is full.  For example, if subprocess output is line buffered and a single line exceeds the buffer size, the buffer can never be emptied because output callbacks are never called (i.e. the buffer never contains a line).

Other situations can exist if the user simply does not read data when it becomes available.

Solution: Handle full output buffers with two special cases

- if output is line buffered and the buffer is full AND no line exists, call the output callback for the user to get the current data.
  flux_subprocess_read_line() and similar functions will return data that is not a full line.

- if the buffer is at capacity and the user elected to not read anything in the output callback, drop the data.  The internal assumption is that a user must read data that is given to them at that point in time.

Fixes #6262

### DIFF
--- a/src/common/libsubprocess/ev_fbuf_read.c
+++ b/src/common/libsubprocess/ev_fbuf_read.c
@@ -23,10 +23,15 @@ static bool data_to_read (struct ev_fbuf_read *ebr, bool *is_eof)
     if (ebr->line) {
         if (fbuf_has_line (ebr->fb))
             return true;
-        /* if eof read, no lines, but left over data non-line data,
-         * this data should be flushed to the user */
-        else if (ebr->eof_read && fbuf_bytes (ebr->fb))
-            return true;
+        else {
+            /* if no line, but the buffer is full, we have to flush */
+            if (!fbuf_space (ebr->fb))
+                return true;
+            /* if eof read, no lines, but left over data non-line data,
+             * this data should be flushed to the user */
+            if (ebr->eof_read && fbuf_bytes (ebr->fb))
+                return true;
+        }
     }
     else {
         if (fbuf_bytes (ebr->fb) > 0)
--- a/src/common/libsubprocess/remote.c
+++ b/src/common/libsubprocess/remote.c
@@ -159,7 +159,10 @@ static bool remote_out_data_available (struct subprocess_channel *c)
 {
     /* no need to handle failure states, on fatal error, the
      * io watchers are closed */
-    if ((c->line_buffered && fbuf_has_line (c->read_buffer))
+    /* N.B. if line buffered and buffer full, gotta flush it
+     * regardless if there's a line or not */
+    if ((c->line_buffered
+         && (fbuf_has_line (c->read_buffer) || !fbuf_space (c->read_buffer)))
         || (!c->line_buffered && fbuf_bytes (c->read_buffer) > 0)
         || (c->read_eof_received && !c->eof_sent_to_caller))
         return true;
@@ -188,6 +191,7 @@ static void remote_out_check_cb (flux_reactor_t *r,
 
     if ((c->line_buffered
          && (fbuf_has_line (c->read_buffer)
+             || !fbuf_space (c->read_buffer)
              || (c->read_eof_received
                  && fbuf_bytes (c->read_buffer) > 0)))
         || (!c->line_buffered && fbuf_bytes (c->read_buffer) > 0)) {
@@ -437,6 +441,13 @@ static int remote_output_buffered (flux_subprocess_t *p,
 
     if (data && len) {
         int tmp;
+
+        /* In the event the buffer is full, the `fbuf_write()` will
+         * fail.  Call user callback to give them a chance to empty
+         * buffer.  If they don't, we'll hit error below.
+         */
+        if (!fbuf_space (c->read_buffer))
+            c->output_cb (c->p, c->name);
 
         tmp = fbuf_write (c->read_buffer, data, len);
         if (tmp >= 0 && tmp < len) {
--- a/src/common/libsubprocess/remote.c
+++ b/src/common/libsubprocess/remote.c
@@ -157,8 +157,8 @@ static void process_new_state (flux_subprocess_t *p,
 
 static bool remote_out_data_available (struct subprocess_channel *c)
 {
-    /* no need to handle failure states, on fatal error, these
-     * reactors are closed */
+    /* no need to handle failure states, on fatal error, the
+     * io watchers are closed */
     if ((c->line_buffered && fbuf_has_line (c->read_buffer))
         || (!c->line_buffered && fbuf_bytes (c->read_buffer) > 0)
         || (c->read_eof_received && !c->eof_sent_to_caller))
@@ -202,8 +202,8 @@ static void remote_out_check_cb (flux_reactor_t *r,
         c->p->channels_eof_sent++;
     }
 
-    /* no need to handle failure states, on fatal error, these
-     * reactors are closed */
+    /* no need to handle failure states, on fatal error, the
+     * io watchers are closed */
     if (!remote_out_data_available (c) || c->eof_sent_to_caller) {
         /* if no data in buffer, shut down prep/check */
         flux_watcher_stop (c->out_prep_w);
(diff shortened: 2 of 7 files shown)


garlick:
Nice!  does this fix #4572 also?

chu11:
Hmm, I don't think so.  The `Value too large for defined data type` suggests it might be an issue more similar to https://github.com/flux-framework/flux-core/issues/6256

src/common/libsubprocess/ev_fbuf_read.c
```diff
@@ -105,6 +110,31 @@ static void check_cb (struct ev_loop *loop, ev_check *w, int revents)
...
+             * "fallout" from a watcher just (effectively) exiting out
+             * of the blue.  From caller perspectives, it may have
+             * exited cleanly.
+             *
+             * we choose to dump the buffer contents instead.
+             * Unfortunately this leads to loss of data and no error
+             * message.  However in authors opinion, it is a "cleaner"
+             * fallout.
+             *
+             */
```
garlick, line 135: Pondering this a bit, it seems like this ev watcher ought to behave similar to the libev ev_io watcher:

http://pod.tst.eu/http://cvs.schmorp.de/libev/ev.pod#code_ev_io_code_is_this_file_descrip

That is, it should be level triggered (keep calling callback while data is available) and should be stoppable in the watcher callback.  This change would drop data if the watcher chose to stop itself rather than consume the data so I think it would not be correct.

It may be good to add some inline documentation though (I guess in `fbuf_watcher.h`?) that the watchers are level triggered and that the read watcher must either read some data or stop the watcher to avoid unnecessary trips through the event loop without progress.

Also, maybe add documentation in subprocess.h that the output callbacks should either consume data or call `flux_subprocess_stream_stop()`.

garlick (commented):
See inline comment - it seems like the second problem may be more of a documentation one?

src/common/libsubprocess/ev_fbuf_read.c
garlick: Maybe this PR should just ensure that, when line buffered and the buffer is full, the callback gets even there isn't a line terminator, and level triggering documentation (and any fixes for in-tree inappropriate uses, if any) could go in a separate PR?

chu11: > Maybe this PR should just ensure that, when line buffered and the buffer is full, the callback gets called even there isn't a line terminator, and level triggering documentation (and any fixes for in-tree inappropriate uses, if any) could go in a separate PR?

Let me try again to split it up.  I think the reason I didn't for this PR is b/c a user not being able to read a buffer without a line  == a user not reading data b/c they "forgot to".  So that's why I ended up putting it all together.  It led to some testing difficulties otherwise.  But I'll give it another shot.

garlick:
I was questioning the correctness of the proposed change for "forgot to".

src/common/libsubprocess/remote.c
```diff
@@ -438,6 +442,13 @@ static int remote_output_buffered (flux_subprocess_t *p,
     if (data && len) {
         int tmp;
 
+        /* In the event the buffer is full, the `fbuf_write()` will
+         * fail.  Call user callback to give them a chance to empty
+         * buffer.  If they don't, we'll hit error below.
+         */
+        if (!fbuf_space (c->read_buffer))
+            c->output_cb (c->p, c->name);
+
```
garlick, line 451: Was going to comment that throwing an error here is not consistent with allowing the user to stop the stream from the callback, but I see the stream_start/stop functions are noted to be for local processes only.

We can fix that once we have credit based flow control since the remote will never send more data than we have room to put in the buffer.

chu11:
re-pushed, this PR is now limited to just the line buffering corner case

src/common/libsubprocess/test/stdio.c
```diff
@@ -1395,6 +1473,8 @@ int main (int argc, char *argv[])
     test_stream_start_stop_initial_stop (r);
     diag ("stream_start_stop_mid_stop");
     test_stream_start_stop_mid_stop (r);
+    diag ("overflow_output_buffer");
+    test_overflow_output_buffer (r);
```
garlick, line 1477: Suggestoin: rename this test to something like `test_long_lines()` or similar since the point is to demonstrate that the buffer does _not_ overflow when it gets a long line.

garlick (commented):
Looking good!  I have a few comments that are mostly trivial.

I did want to manually kick it a bit with `lptest` and see how things go so I'll save approval for after that test.

garlick (approved):
I did some testing with `flux exec` running lptest with various line sizes and lengths, and comparing the output to the same command run locally.  I even tried 4G worth of (4MB-500B) lines!  No issues.

I then repeated some of these tests with the data going to a job's stdout.  Also no issues.

Approving!  Nice work.

src/common/libsubprocess/remote.c
chu11: the callback placement here is unfortunate, the reason was the libev did not call things in the order I expected.   I expected:

```
output - put data in buffer, start output prep/check
check_cb - call output callback since there's data in the buffer
<start next iteration of libev loop>
```

but what happened was

```
output - put data in buffer, start output prep/check
<start next iteration of libev loop>
prep_check - see data in output buffer, start idle
output - want to put more data in buffer, hit EOVERFLOW
check_cb - this is never reached because of error above
```

the fact I just started the prep/check means check isn't called in the current iteration.

chu11: > We can fix that once we have credit based flow control since the remote will never send more data than we have room to put in the buffer.

The work on #6291 is only for stdin since that's the specific case brought up by the user.  But yeah, for output we should add that as a todo as well.

garlick: On the order of events, should the output watcher be stopped when the buffer is full then, and restarted when it's not?

garlick: I think that's sort of like putting an infinite max size on the cbuf - it allows the sender to cause the receiver to use up an unchecked amount of memory.  Not great.

Another route might be to expose watcher priorities in our APIs, and then use them to ensure this check watcher runs before anything else.

But maybe the right solution is flow control.

chu11: > Another route might be to expose watcher priorities in our APIs, and then use them to ensure this check watcher runs before anything else.

I wasn't aware of the libev priority stuff.  Hmmmm.  I suppose that could be an option, but at this point in time I'm not sure we have a way to add a priority to whatever underneath the covers is calling the flux future's then callback?  So perhaps this is something to simply kick the can down the road.

chu11:

src/common/libsubprocess/remote.c
garlick: Well we could elevate the priority of just this check watcher to get it to be called before the  check watcher in the future implementation.   Did you want to pause and try that?  I could give you a commit that adds a `flux_watcher_set_priority()` function to cherry pick.  Just as an experiment?

garlick: Untested but here it is: 1932ce57c5c167385759e4fef7cd22294f5d25ca

Just call `flux_watcher_set_priority (check_watcher, 1);`

That should raise the priority from the default of 0 to 1.  if that doesn't work try 2 :-)  2 is the max.

chu11: > Did you want to pause and try that?

As this would involve more than a few line tweak, I'm inclined to merge this PR and experiment with it in a different PR.  But lets log so we don't forget this conversation.

garlick: Sure.

PR #6282, opened 2024-09-13
python: update Flux handle location in validator plugin API

grondo:
Currently, the Python validator plugin interface provides an on-demand Flux handle via the `ValidatorJobInfo` class. However, in retrospect, this was a big mistake. Since the `ValidatorJobInfo` object is created once per job, the handle is closed and must be reopened for each job passed to a validator plugin. Also, since the job info object is passed to multiple validators running in different threads, thread-local storage needs to be used, which complicates the implementation. Finally, the `ValidatorJobInfo` object is obviously not available when the plugin is configuring itself, which makes it impossible for plugins to contact Flux to configure themselves (or they have to open a handle themselves)

This PR instead provides a new `flux` attribute in the `ValidatorPlugin` class which opens and returns a Flux handle on first access. Because each plugin is run in its own thread, there is no longer a need to deal with TLS, simplifying the approach greatly. Finally, the handle is conveniently available any time the plugin is active, so the `configure` callback can now use it.

The handle provided by `ValidatorJobInfo` is then removed once the one internal user is updated. This could be a problem if there are out-of-tree validators out there (but I don't think that is likely at this point).

### DIFF
--- a/src/bindings/python/flux/job/validator/plugins/feasibility.py
+++ b/src/bindings/python/flux/job/validator/plugins/feasibility.py
@@ -31,7 +31,7 @@ class Validator(ValidatorPlugin):
 
     def validate(self, args):
         try:
-            args.flux.rpc(self.service_name, args.jobinfo).get()
+            self.flux.rpc(self.service_name, args.jobinfo).get()
         except OSError as err:
             if err.errno == errno.ENOSYS:
                 #  Treat ENOSYS as success
--- a/src/bindings/python/flux/job/validator/validator.py
+++ b/src/bindings/python/flux/job/validator/validator.py
@@ -11,7 +11,6 @@
 import argparse
 import concurrent.futures
 import json
-import threading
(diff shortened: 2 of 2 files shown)


garlick (approved):
LGTM!

grondo:
Thanks! Will set MWP.

wihobbs:
@grondo can you clarify what you mean by "out of tree validators"? Is that any validator that's not stored alongside the rpm or in the builddir for the running `flux`?

grondo:
> can you clarify what you mean by "out of tree validators"? Is that any validator that's not stored alongside the rpm or in the builddir for the running flux?

By out-of-tree I meant any validator plugins being maintained outside of the flux-core source tree. If they use the flux handle in `JobValidatorInfo` then they'll need to be updated.

wihobbs:
Ah, ok, thanks!

PR #6283, opened 2024-09-13
broker: use enclosing instance tbon.interface-hint unless overridden

garlick:
Problem: we don't have a way to configure subinstances to use a particular network, other than setting the global environment variable FLUX_IPADDR_INTERFACE.
    
Change the way the default `tbon.interface-hint` is set.  Instead of hard coding "default-route", use the value that is set in the enclosing instance, if any.
    
This would allow a system instance to configure `tbon.interface-hint` and batch/alloc jobs that it starts would use that value unless overridden.  If the value is not configured, then the attribute will have the default value which will be inherited by the child.

### DIFF
--- a/t/t0019-tbon-config.t
+++ b/t/t0019-tbon-config.t
@@ -68,7 +68,8 @@ test_expect_success 'TOML tbon.interface-hint=wrong type fails' '
 	test_must_fail flux start -o,--config-path=badhint.toml ${ARGS} true
 '
 test_expect_success 'tbon.interface-hint=badiface fails' '
-	test_expect_code 137 flux start -o,-Stbon.interface-hint=badiface \
+	test_must_fail_or_be_terminated flux start \
+		-o,-Stbon.interface-hint=badiface \
 		${ARGS} -s2 -o,-Stbon.prefertcp=1 true
 '
 test_expect_success 'tbon.interface-hint=default-route works' '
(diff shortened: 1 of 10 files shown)


garlick:
The thing that's bugging me about this solution is it adds a synchronous RPC (as well as a `flux_open()` / `flux_close()` to the parent broker on every rank.  They will go in parallel so it will scale but it adds a fixed startup latency to every job.

One tactic to reduce the latency would be to pre-populate a PMI key like `flux.tbon-interface-hint`.  It still will add the PMI RTT cost but at least the connection is already open.  In retrospect that would seem to be preferable.  Then perhaps later if we decide we want to pass a default configuration object to subinstances, we could eliminate it entirely.

I'll mark this WIP pending some feedback on the approach.

garlick:
Reworked to transfer the enclosing instance's `tbon.interface-hint` via PMI.

t/t0019-tbon-config.t
```diff
@@ -79,6 +79,15 @@ test_expect_success 'tbon.interface-hint=hostname works' '
...
 '
+test_expect_success 'tcp.interface-hint defaults to default-router' '
+	flux start ${ARGS} flux getattr tbon.interface-hint >defhint.out &&
+	grep default-route defhint.out
+'
+test_expect_success 'tcp.interface-hint setting is child default' '
+	flux start -o,-Stbon.interface-hint=hostname \
+	    flux alloc -N1 flux getattr tbon.interface-hint >childhint.out &&
+	grep hostname childhint.out
+'
```
grondo, line 91: Should a test be added that ensures setting the attribute in the subinstance uses the set value and not the inherited one?

grondo (approved):
This looks good to me, nice work! Just a couple trivial comments inline.

garlick:
Thanks!  I addressed those comments and forced a push.  Will set MWP.

PR #6284, opened 2024-09-14
support expandable width output formats and use them in `flux resource list` to avoid truncation of queue field

grondo:
This PR adds two new output format "sentinels" `+:` and `?+:` to the `OutputFormat` class that allow automated adjustment of the field width to avoid truncation. `+:` only does the expansion, while `?+:` combines the functionality of `+:` and `?:` and expands the field to the maximum observed width and removes the field if all entries would result in an "empty" value.

Then `?+:` is applied to `flux resource list` default output for `queue` and `ngpus` fields. This avoids truncation of the QUEUE column, which may contain comma separated list of queues on system instances (and in the case of `ngpus` saves a couple characters of width in the common case)

### DIFF
--- a/src/bindings/python/flux/util.py
+++ b/src/bindings/python/flux/util.py
@@ -631,6 +631,78 @@ class OutputFormat:
         if self.prepend:
             self.fmt = self.get_format_prepended(self.prepend)
 
+    class FormatSpec:
+        """Split Python format spec into components"""
+
+        # Note: 'hyphen' and 'truncate' are Flux extensions
+        components = (
+            "fill",
+            "align",
+            "sign",
+            "pos_zero",
+            "alt",
+            "zero_pad",
+            "width_str",
+            "grouping",
+            "decimal",
+            "precision_str",
+            "type",
+            "hyphen",
+            "truncate",
+        )
+
+        def __init__(self, spec):
+
+            # Regex taken from https://stackoverflow.com/a/78351366
+            # 'hyphen' and 'truncate' are Flux extensions
+            spec_re = re.compile(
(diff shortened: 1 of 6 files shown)


src/cmd/flux-pgrep.py
```diff
@@ -274,7 +274,7 @@
     if PROGRAM == "flux-pkill":
         pkill(fh, args, jobs)
 
-    sformatter = JobInfoFormat(formatter.filter_empty(jobs))
+    sformatter = JobInfoFormat(formatter.filter(jobs))
```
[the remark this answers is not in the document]
grondo: This is a false positive

src/bindings/python/flux/util.py
```diff
@@ -734,53 +806,115 @@ def filter_empty(self, items):
...
+            end = next((x for x in sentinel_keys() if text.endswith(x)), False)
+
+            if end:
+                #  This entry matches one of the filtering sentinels, parse
+                #  the spec to get current width (and allow possible
+                #  reconstruction later).
+                spec = self.FormatSpec(spec)
+
+                #  Save a format without any spec to allow deterimination of
+                #  maximum width after formatting all items:
```
garlick, line 843: s/deterimination/determination/

garlick (approved):
Nice!  Looks good to me.  I noticed one comment typo.

src/bindings/python/flux/util.py
grondo: Thanks! Fixed that and I also dropped the flux-pgrep codeql commit since it didn't help workaround the codeql false positive. Will set MWP now.

PR #6289, opened 2024-09-18
shell/oom: log more detail when tasks are killed

garlick:
Problem: when a job exceeds the memory cgroup limit, it may not be immediately obvious what the limit is or what host it occurred on.

Include the hostname with the existing "out of memory" message. Log the value of memory.peak (the peak memory usage).

New output example
```
29.475s: flux-shell[0]: ERROR: oom: Memory cgroup out of memory: killed 1 task on picl7.
29.563s: flux-shell[0]: ERROR: oom: memory.peak = 1.4428558G
```


grondo (approved):
LGTM!

garlick:
Thanks - I'll set MWP.

PR #6299, opened 2024-09-24
flux-bulksubmit: support `{}`  in more options like `--cwd=`, `--signal=`, `--taskmap=`, etc.

### LINKED ISSUE #6295: flux-bulksubmit: add `--cwd` to the set of substitutable options
The `--cwd` option to `flux bulksubmit` does not support expansion of `{}`. This was likely just an oversight and should be fixed.

grondo:
Problem: The Xcmd class that implements option argument mutability in flux-bulksubmit(1) has a manual list of mutable options that is missing some useful options.

Update the list to include some missing options, like --cwd=, --signal=, --taskmap=, and --begin-time=.

Fixes #6295

### DIFF
--- a/src/bindings/python/flux/cli/base.py
+++ b/src/bindings/python/flux/cli/base.py
@@ -517,6 +517,7 @@ class Xcmd:
         "time_limit": "-t",
         "env": "--env=",
         "env_file": "--env-file=",
+        "env_remove": "--env-remove=",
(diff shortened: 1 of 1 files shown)


grondo:
The specific options that support `{}` are not currently called out in the man page.

garlick:
Right, for another day maybe.  This looks good to go to me!

PR #6303, opened 2024-09-25
autoconf: support python 3.12

trws:
problem: flux-core currently can't configure with python3.12 due to setuptools no longer being part of the default distribution

solution: import setuptools where necessary to keep things working in the ax_python_devel file

This worked to build a copy of flux, but I didn't get the dep install scripts updated yet because I couldn't get an image downloaded on this wifi.  Now off to dinner and the airport.

### DIFF
--- a/config/ax_python_devel.m4
+++ b/config/ax_python_devel.m4
@@ -107,7 +107,7 @@ AC_DEFUN([AX_PYTHON_DEVEL],[
 	   # Check for a version of Python >= 2.1.0
 	   #
 	   AC_MSG_CHECKING([for a version of Python >= '2.1.0'])
-	   ac_supports_python_ver=`$PYTHON -c "import sys; \
+	   ac_supports_python_ver=`$PYTHON -c "import sys;import setuptools; \
 		ver = sys.version.split ()[[0]]; \
 		print (ver >= '2.1.0')"`
 	   if test "$ac_supports_python_ver" != "True"; then
@@ -150,6 +150,7 @@ class VPy:
         return tuple(map(int, s.strip().replace("rc", ".").split(".")))
     def __init__(self):
         import sys
+        import setuptools
         self.vpy = tuple(sys.version_info)[[:3]]
     def __eq__(self, s):
         return self.vpy == self.vtup(s)
--- a/scripts/install-deps-deb.sh
+++ b/scripts/install-deps-deb.sh
@@ -20,6 +20,7 @@ apt install \
   python3-dev \
   python3-cffi \
   python3-ply \
+  python3-setuptools \
   python3-yaml \
   python3-sphinx \
   aspell \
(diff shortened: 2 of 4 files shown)


wihobbs:
Oooh, some neat stuff in here! https://docs.python.org/3/whatsnew/3.12.html

grondo (approved):
Thanks! LGTM!

grondo:
I've set MWP here. Thanks @trws!

grondo:
Actually this is failing in the alpine builder with:
```
checking for a version of Python >= '2.1.0'... Traceback (most recent call last):
  ModuleNotFoundError: No module named 'setuptools'
  no
  configure: WARNING:
  This version of the AC_PYTHON_DEVEL macro
  doesn't work properly with versions of Python before
  2.1.0. You may need to re-run configure, setting the
  variables PYTHON_CPPFLAGS, PYTHON_LIBS, PYTHON_SITE_PKG,
  PYTHON_EXTRA_LIBS and PYTHON_EXTRA_LDFLAGS by hand.
  Moreover, to disable this check, set PYTHON_NOVERSIONCHECK
  to something else than an empty string.
  
  configure: error: in `/usr/src':
  configure: error: Giving up
```

I'll try to take a look at this tonight.

trws:
Dang, that probably needs the py3-setuptools module added to the container. Same thing that needs to go into the deps install scripts and anywhere we pull in a newer python.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

trws:
I managed to test it, adding py3-setuptools looks like it's all we actually need.  The other dockerfiles already have it. I pushed the dockerfile, but the internet on the plane is not conducive to actually building and pushing the darned thing to dockerhub.

grondo:
>  but the internet on the plane is not conducive to actually building and pushing the darned thing to dockerhub.

I can do this but I can't seem to build docker images anymore (I may need to set up a VM with a newer docker):
```
$ docker build --no-cache -t fluxrm/testenv:alpine -f src/test/docker/alpine/Dockerfile 
ERROR: "docker buildx build" requires exactly 1 argument.
See 'docker buildx build --help'.

Usage:  docker buildx build [OPTIONS] PATH | URL | -

Start a build
```

Note: We need to update the `src/test/docker/README.md`. Instructions don't work anymore.

trws:
I think you’re missing the positional argument for the context directory, normally it’s just “.” or whatever the path to the flux core repo base is.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

grondo:
D'oh! That seems to be working. :facepalm:

grondo:
Ok, pushed the alpine image and reran CI and the PR is now merged. Thanks @trws!

trws:
Awesome, thanks @grondo!


Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #6305, opened 2024-09-25
support configuration of require-instance and other job validator/frobnicator plugins in broker config TOML

### LINKED ISSUE #6298: add table for job validators in broker config
Problem: Configuration for job validator plugins currently has to be passed as options on the `flux job-validator` command line. Those options have to in turn be added to the ingest module config. With many validators, this could potentially get unwieldy and is especially difficult to pass along via the `--conf` option of `flux batch` and `alloc` if necessary.

Perhaps a table in the broker config TOML could be reserved for validator plugin configuration. e.g. either `[validator.plugin-name]` or `[ingest.validator.plugin-name]`.

This would allow, for example, the `require-instance` plugin to be configured via:

```toml
[validator.require-instance]
minnodes = 16
mincores = 768
```

Thoughts on which table to reserve, and how to document it?

grondo:
This PR allows validator/frobnicator plugins to get their configuration from broker config TOML via tables keyed by plugin name in `[ingest.validator.*]` or `[ingest.frobnicator.*]`. The `require-instance` validator plugin is then enhanced to use this config for its `minnodes` and `mincores` configuration. Finally, documentation is added to `flux-config-ingest(5)` describing the configuration of validators/frobnicators in general, and `require-instance` specifically.

Fixes #6298


garlick (approved):
Sounds good!

grondo:
Thanks! I've set MWP.

PR #6307, opened 2024-09-25
libsubprocess: add extra documentation

chu11:
per comment in #6281 some extra comments could be useful.  I did find myself struggling remembering how read vs write fbuf watchers behaved too.


garlick (approved):
LGTM!

PR #6309, opened 2024-09-25
python: fix JobList default documentation

### LINKED ISSUE #6308: python: fix JobList documentation for `user`
```
:user: Username or userid for which to fetch jobs. Default is all users.
```

JobList does not appear to default to all users, just to the current user.  Also the special case `user="all"` should be documented.

chu11:
Problem: The JobList class documentation indicates that the default is to list jobs for all users.  This is incorrect, the default is to list jobs for the current user.

Update documentation accordingly.

Fixes #6308

---

and fixup one other thing I noticed

### DIFF
--- a/src/bindings/python/flux/job/list.py
+++ b/src/bindings/python/flux/job/list.py
@@ -276,7 +276,7 @@ class JobList:
         self.constraint = constraint
 
     def set_user(self, user):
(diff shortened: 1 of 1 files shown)


grondo (approved):
LGTM! Good catch.

PR #6312, opened 2024-09-26
add `--force` option to `flux resource undrain`

### LINKED ISSUE #6306: Suggestion: an equivalent to --update  for flux resource undrain
It's reasonably common to be working with a large chunk of nodes and want to undrain them all at once - if I've previously undrained some of them separately flux errors out on the first node that isn't drained:
`flux-resource: ERROR: rank <blah> not drained`

I can then go back and run e.g. `flux resource status -i` to get an updated list, but it'd be handy to have some `undrain` equivalent to `flux resource drain -u` to disable returning the error and let the undrain continue.

grondo:
`flux resource undrain` currently aborts the entire operation if one or more targets is not currently drained. This makes it difficult to simply ensure a set of ranks/hosts are undrained.

Add a `mode=force` option to the `resource.undrain` RPC and a corresponding `--force` option to `flux resource undrain`.

Fixes #6306

### DIFF
--- a/doc/man1/flux-resource.rst
+++ b/doc/man1/flux-resource.rst
@@ -17,7 +17,7 @@ SYNOPSIS
 
 | **flux** **resource** **drain** [*-n*] [*-o* *FORMAT*] [*-i* *TARGETS*]
 | **flux** **resource** **drain** [*-f*] [*-u*] [*targets*] [*reason*]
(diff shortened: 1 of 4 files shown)


garlick (approved):
LGTM!

grondo:
Thanks, setting MWP

PR #6316, opened 2024-09-26
libflux: support modifying watcher priority via flux_watcher_set_priority()

chu11:
half @garlick and half me (just the testing & docs), so someone else needs to approve :-)

### DIFF
--- a/doc/man3/flux_watcher_set_priority.rst
+++ b/doc/man3/flux_watcher_set_priority.rst
@@ -1,0 +1,43 @@
+============================
+flux_watcher_set_priority(3)
+============================
+
+.. default-domain:: c
+
+SYNOPSIS
+========
+
+.. code-block:: c
+
+   #include <flux/core.h>
+
+   flux_watcher_t *flux_watcher_set_priority (flux_watcher_t *w,
+                                              int priority);
+
+Link with :command:`-lflux-core`.
+
+DESCRIPTION
+===========
+
+:func:`flux_watcher_set_priority` sets the priority on the watcher.
+Higher priority watchers run first.  The range of priorities is from
+-2 to 2, with the default being 0.  If the priority is out
+of range, the max or min value is set.  The priority should only be
+set when the watcher is stopped.
+
+This function is a no-op if the underlying watcher does not support priorities.
+Currently only the check watcher supports priorities.
+
+
+RESOURCES
+=========
+
+.. include:: common/resources.rst
+
+libev: http://software.schmorp.de/pkg/libev.html
+
+
+SEE ALSO
+========
+
+:man3:`flux_check_watcher_create`
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -114,6 +114,7 @@ MAN3_FILES_PRIMARY = \
 	man3/flux_timer_watcher_create.3 \
 	man3/flux_periodic_watcher_create.3 \
 	man3/flux_idle_watcher_create.3 \
+	man3/flux_watcher_set_priority.3 \
 	man3/flux_msg_handler_create.3 \
 	man3/flux_msg_cmp.3 \
 	man3/flux_msg_create.3 \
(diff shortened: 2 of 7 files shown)


doc/man3/flux_watcher_set_priority.rst
```diff
@@ -0,0 +1,42 @@
+===========================
+flux_idle_watcher_create(3)
```
garlick, line 2: Need to change the title

chu11: doh!  repushed w/ fix

garlick (approved):
FWIW this looks good to me!

doc/man3/flux_watcher_set_priority.rst
```diff
@@ -0,0 +1,42 @@
...
+DESCRIPTION
+===========
+
+:func:`flux_watcher_set_priority` sets the priority on the watcher.
+Higher priority watchers run first.  The range of priorities is from
+-2 to 2, with the default being 0.  If the priority is out
+of range, the max or min value is set.
+
+This function is a no-op if the underlying watcher does not support priorities.
+Currently only the check watcher supports priorities.
```
garlick, line 28: Oh, one thing: this should mention that the priority can only be set when the watcher is stopped.

chu11: oh good catch and should be in the function comments too.

```
diff --git a/doc/man3/flux_watcher_set_priority.rst b/doc/man3/flux_watcher_set_priority.rst
index 8138330cb..dfc507e10 100644
--- a/doc/man3/flux_watcher_set_priority.rst
+++ b/doc/man3/flux_watcher_set_priority.rst
@@ -22,7 +22,8 @@ DESCRIPTION
 :func:`flux_watcher_set_priority` sets the priority on the watcher.
 Higher priority watchers run first.  The range of priorities is from
 -2 to 2, with the default being 0.  If the priority is out
-of range, the max or min value is set.
+of range, the max or min value is set.  The priority should only be
+set when the watcher is stopped.
 
 This function is a no-op if the underlying watcher does not support priorities.
 Currently only the check watcher supports priorities.
diff --git a/src/common/libflux/reactor.h b/src/common/libflux/reactor.h
index 69e9ebfef..f29d8c57f 100644
--- a/src/common/libflux/reactor.h
+++ b/src/common/libflux/reactor.h
@@ -76,6 +76,7 @@ typedef void (*flux_watcher_f)(flux_reactor_t *r, flux_watcher_t *w,
  * Higher priority watchers run first.
  * This is a no-op if the underlying watcher doesn't support it.
  * If the priority is out of range, the max or min value is set.
+ * The priority should only be set when the watcher is stopped.
  * Currently only the check watcher supports it.
  */
 void flux_watcher_set_priority (flux_watcher_t *w, int priority);

```

garlick:
@grondo I think we're hoping you'll have a quick look at this since it's kind of 50/50 @chu11 and me.

doc/man3/flux_watcher_set_priority.rst
```diff
@@ -0,0 +1,43 @@
...
+Link with :command:`-lflux-core`.
+
+DESCRIPTION
+===========
+
+:func:`flux_watcher_set_priority` sets the priority on the watcher.
+Higher priority watchers run first.  The range of priorities is from
+-2 to 2, with the default being 0.  If the priority is out
+of range, the max or min value is set.  The priority should only be
+set when the watcher is stopped.
```
grondo, line 26: The libev manual makes a stronger statement here:

> You _must not_ change the priority of a watcher as long as it is active or pending. 

I'm wondering if that's warranted here as well? I don't actually know the effects.

grondo (approved):
LGTM!

doc/man3/flux_watcher_set_priority.rst
garlick: Eh, it's probably OK as is IMHO.  I would assume it just wouldn't have any effect since the watchers are likely in stored in a list per priority level, inserted when started and removed when stopped.

garlick:
Thanks! I'll set MWP.

PR #6317, opened 2024-09-26
libsubprocess: increase output watcher priority

chu11:
Problem: If a output buffer is full, we issue an emergency call to the user's output callback to empty the buffer before more data is put in it.  This is done because we cannot control the order in which check callbacks are called.  Now that check callbacks can have priority set, we can ensure output check callbacks are called first before a check callback that may put more data in the buffer.

Set the priority of the output check watcher higher than the default.  Remove the now unnecessary emergency callback to the user's output callback.

built on top of #6316


garlick (approved):
LGTM!

grondo:
Let's add MWP here so it makes it into today's release

PR #6326, opened 2024-09-29
libsubprocess: fix bulk-exec reporting of active ranks

### LINKED ISSUE #6321: perilog: killed prolog doesn't get cleaned up
After manually killing a prolog process, the `prolog-finish` event didn't get posted and the perilog.so plugin shows the prolog in this weird state:
```
# flux jobtap query perilog.so | jq .procs.f9EPPHk7c9d
{
  "name": "prolog",
  "state": "canceled",
  "total": 1,
  "active": 1,
  "active_ranks": ""
}
```

grondo:
When a bulk-exec launched subprocess is terminated but still has children with open stdio fds, `bulk_exec_active_ranks()` reports an empty idset, even though the subprocess should still be considered active.

Add a libsubprocess accessor to determine if a subprocess is still active, i.e. is not failed or completed, and use this in bulk-exec to properly report all "active" subprocesses.

Fixes #6321

### DIFF
--- a/src/common/libsubprocess/bulk-exec.c
+++ b/src/common/libsubprocess/bulk-exec.c
@@ -121,7 +121,7 @@ struct idset *bulk_exec_active_ranks (struct bulk_exec *exec)
 
     p = zlist_first (exec->processes);
     while (p) {
-        if (flux_subprocess_state (p) == FLUX_SUBPROCESS_RUNNING) {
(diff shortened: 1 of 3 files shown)


garlick (approved):
LGTM!

grondo:
Thanks! Setting MWP.

PR #6330, opened 2024-09-30
perilog: never send SIGKILL to prolog/epilog, drain active nodes after kill-timeout instead

grondo:
This PR modifies the perilog plugin to never send SIGKILL to the prolog/epilog. Instead, after kill-timeout has expired, nodes still actively running a perilog are drained and the finish event posted.

To avoid too eagerly draining nodes, the default kill-timeout is doubled from 5 to 10 seconds.

### DIFF
--- a/src/modules/job-manager/plugins/perilog.c
+++ b/src/modules/job-manager/plugins/perilog.c
@@ -452,6 +452,7 @@ static flux_future_t *proc_drain_ranks (struct perilog_proc *proc,
     struct idset *failed = NULL;
     flux_future_t *f = NULL;
     unsigned long rank;
+    const char *msg;
     char reason[256];
     flux_t *h = flux_jobtap_get_flux (proc->p);
 
@@ -482,11 +483,19 @@ static flux_future_t *proc_drain_ranks (struct perilog_proc *proc,
                         perilog_proc_name (proc));
         goto out;
     }
+
+    if (proc->canceled)
+        msg = "canceled then timed out";
+    else if (proc->timedout)
+        msg = "timed out";
+    else
+        msg = "failed";
+
     (void) snprintf (reason,
                      sizeof (reason),
                      "%s %s for job %s",
                      perilog_proc_name (proc),
-                     proc->timedout ? "timed out" : "failed",
+                     msg,
                      idf58 (proc->id));
 
     if (!(f = flux_rpc_pack (h,
--- a/src/modules/job-manager/plugins/perilog.c
+++ b/src/modules/job-manager/plugins/perilog.c
@@ -248,7 +248,7 @@ static struct perilog_procdesc *perilog_procdesc_create (json_t *o,
     }
 
     pd->cmd = cmd;
-    pd->kill_timeout = kill_timeout > 0. ? kill_timeout : 5.;
+    pd->kill_timeout = kill_timeout > 0. ? kill_timeout : 10.;
     pd->per_rank = per_rank;
     pd->prolog = prolog;
     pd->uses_imp = uses_imp;
--- a/src/modules/job-manager/plugins/perilog.c
+++ b/src/modules/job-manager/plugins/perilog.c
@@ -435,7 +435,18 @@ static void emit_finish_event (struct perilog_proc *proc,
     }
 }
 
-static flux_future_t *drain_failed_ranks (struct perilog_proc *proc)
+static bool subprocess_failed (flux_subprocess_t *p)
+{
+    if (flux_subprocess_state (p) == FLUX_SUBPROCESS_FAILED
+        || flux_subprocess_status (p) != 0)
+        return true;
+    return false;
+}
+
+/* Drain ranks that failed, are still active or both. */
+static flux_future_t *proc_drain_ranks (struct perilog_proc *proc,
+                                        bool drain_failed,
+                                        bool drain_active)
 {
     struct idset *failed = NULL;
     flux_future_t *f = NULL;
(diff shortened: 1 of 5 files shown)


grondo:
There may be a rare race in the perilog test now (or actually could be pre-existing). Let me look into that and make sure it isn't due to the implementation.

src/modules/job-manager/plugins/perilog.c
```diff
@@ -617,35 +617,49 @@ static bool perilog_per_rank (struct perilog_proc *proc)
...
+         * be freed and handed out to new jobs before they are drained.
+         */
+        if ((proc->drain_f = proc_drain_ranks (proc,
+                                               drain_failed,
+                                               drain_active))
+                && flux_future_then (proc->drain_f,
+                    -1.,
+                    drain_failed_cb,
+                    proc) == 0)
+            return;
```
garlick, line 642: Indent needs adjustment

src/modules/job-manager/plugins/perilog.c
```diff
@@ -617,35 +617,49 @@ static bool perilog_per_rank (struct perilog_proc *proc)
...
+
+        /* O/w, drain RPC failed, report error and fall through so finish
+         * event is still emitted.
+         */
+        flux_log_error (flux_jobtap_get_flux (proc->p),
+                        "%s: failed to drain %s failed ranks",
+                        idf58 (proc->id),
+                        perilog_proc_name (proc));
```
garlick, line 650: Minor:  `proc_drain_ranks()` also logs on error.  Since all it and this are doing is preparing and sending a request, it seems like one log message should suffice.

src/modules/job-manager/plugins/perilog.c
```diff
@@ -1030,13 +1024,16 @@ static int proc_kill_timer_start (struct perilog_proc *proc, double timeout)
...
         if (!proc->kill_timer) {
             flux_log_error (h,
                             "%s: failed to start %s kill timer",
                             idf58 (proc->id),
                             perilog_proc_name (proc));
+            /* Since timer cb won't be run, drain and send finish event now
+             */
+            proc_drain_and_finish (proc, false, true);
             return -1;
         }
```
garlick, line 1042: Suggestion: create the timer when proc is created but don't start it until here.  Then the error is not required.

src/modules/job-manager/plugins/perilog.c
```diff
@@ -248,7 +248,7 @@ static struct perilog_procdesc *perilog_procdesc_create (json_t *o,
     }
 
     pd->cmd = cmd;
-    pd->kill_timeout = kill_timeout > 0. ? kill_timeout : 5.;
+    pd->kill_timeout = kill_timeout > 0. ? kill_timeout : 10.;
```
garlick, line 252: I think systemd `TimeoutStopSec` defaults to 90s, which for our systemd `flux-prolog@` and `flux-epilog@` services would be the time from SIGTERM induced `systemctl stop` to systemd SIGKILL.

Seems like we should coordinate this value to be greater than that.   If necessary, can reduce the default to whatever we want by adding `TimeoutStopSec` to the unit files.

garlick (approved):
LGTM!  A couple of suggestions inline that you can take or leave.

src/modules/job-manager/plugins/perilog.c
grondo: Then no error would be logged if `flux_future_then()` fails. Rare enough not to matter? Eh, probably. I'll remove it.

src/modules/job-manager/plugins/perilog.c
grondo: Deleted previous comment. I understand what you're saying now. Good idea.

src/modules/job-manager/plugins/perilog.c
garlick: No argument that the timeout should be < 90s.

Are we ok with nodes getting drained after stuck prolog, then systemd SIGKILLing the cgroup 90s later?  We could reduce `TimeoutStopSec` in the unit files we ship if desired.

src/modules/job-manager/plugins/perilog.c
garlick: Eh, I wasn't advocating for no logging on `flux_future_then()` failure.  Just that maybe the logging could be consolidated somehow.  Maybe not worth changing it.

src/modules/job-manager/plugins/perilog.c
grondo: That's a good question. I'm not sure of the right solution.

garlick: Well maybe we can go with not changing anything in this release and see if that behavior is bothersome.

src/modules/job-manager/plugins/perilog.c
grondo: Actually let's leave this as is for now. It is only tangentially related to this PR, and the `remaining_time` member of the proc status response test for `proc->kill_timer` to determine if the timer is armed, so an alternate way to do that will need to be done (we never merged the change to expose if a watcher is "active").

The extra error handling doesn't seem like the end of the world, and there's a benefit of not creating the watcher when it isn't used 99% of the time.

But I'll handle this in a future PR. This one has gone on too long as it is.

grondo:
Setting MWP here, thanks!

src/modules/job-manager/plugins/perilog.c
grondo: Yeah, admins could always add an override file in a pinch.

Issue #6325, opened 2024-09-29
flux-job info error output is the same for invalid key or invalid jobid


ofaaland:
re: flux-core-0.66.0-1.t4.x86_64

It would be nice if the error output from `flux job` differentiated between an invalid `jobid` and `key`.  It would speed up finding a typo and/or make it faster to find the problem in a script that uses this.

In both cases, currently, the error is "No such file or directory"

```
 $for key in jobspec jobspecX; do  for jobid in f9nTX5ggv9u f9nTX5ggvYY;  do    cmd="flux job info $jobid $key | jq . |head -n1"; echo $cmd; eval $cmd;    echo;  done; done 
flux job info f9nTX5ggv9u jobspec | jq . |head -n1
{

flux job info f9nTX5ggvYY jobspec | jq . |head -n1
flux-job: No such file or directory

flux job info f9nTX5ggv9u jobspecX | jq . |head -n1
flux-job: No such file or directory

flux job info f9nTX5ggvYY jobspecX | jq . |head -n1
flux-job: No such file or directory
```

garlick:
Eek, yes.

PR #6331, opened 2024-10-01
flux job info: improve error messages

garlick:
Problem: as noted in #6325, `flux job info` prints the same error message for a missing key as for an unknown job id.

Fixing this turned out to be tricky because the `job-info.lookup` service is just acting as a proxy for a KVS lookup and doesn't have a direct way to check for an invalid ID - it just knows that a KVS lookup for` job.<id>.R` (or whatever) got ENOENT.  However it does also lookup the job  eventlog for access control purposes for guests only.  The instance owner is immune from such checks.

The solution is to always look up the eventlog regardless of request credentials if _that_ fails with ENOENT then we know the job ID does not exist.

Because this code is pretty old, a fair bit of modernization was also required.   And there is more to do but I tried to keep this focused on the code involved in the reported problem.

### DIFF
--- a/t/t2232-job-info-security.t
+++ b/t/t2232-job-info-security.t
@@ -86,14 +86,14 @@ test_expect_success 'flux job eventlog fails (wrong user)' '
 	unset_userid &&
 	jobid=$(submit_job 9000) &&
 	set_userid 9999 &&
-	! flux job eventlog $jobid
+	test_must_fail flux job eventlog $jobid
 '
 
 test_expect_success 'flux job eventlog fails on bad first event (user)' '
 	unset_userid &&
 	jobid=$(bad_first_event 9000) &&
 	set_userid 9999 &&
-	! flux job eventlog $jobid
+	test_must_fail flux job eventlog $jobid
 '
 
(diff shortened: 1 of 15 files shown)


t/t2232-job-info-security.t
```diff
@@ -79,147 +79,148 @@ test_expect_success 'flux job eventlog works (owner)' '
 test_expect_success 'flux job eventlog works (user)' '
 	jobid=$(submit_job 9000) &&
 	set_userid 9000 &&
-	flux job eventlog $jobid &&
```
grondo, line 82: commit message typo: whcih

t/t2232-job-info-security.t
```diff
@@ -6,51 +6,17 @@ test_description='Test flux job info service security'
 
 test_under_flux 1 job
 
-# We have to fake a job submission by a guest into the KVS.
```
grondo, line 9: commit message typo: "but the job-info LRU will now more aggressively caches eventlog" -- caches -> cache

grondo (approved):
LGTM! Only spotted a couple very minor commit message typos

garlick:
Thanks!  Fixed those two and forced a push.  Will set MWP.

PR #6333, opened 2024-10-01
doc: add warning about stdio file paths in submission cli man pages

### LINKED ISSUE #6329: flux-batch(1) suggestion for --input, --output, and --error
It would be helpful if flux-batch(1) mentioned that for `--output` and `--error`, the file is written to on the node where rank 0 runs, and describes the difference between the cases where an absolute path vs a relative path are given, and what happens if a directory does not exist.

Slightly different, but along the same lines, `--input` I assume must refer to a file present on the rank 0 node.

Maybe this applies to other areas of the documentation, too.

grondo:
Problem: It may surprise users that --output, --error, and --input files are opened on the first node of the allocation and relative to the working directory of the job, not on the submission node.

Add a note to the common job-standard-io.rst document warning users of this potential pitfall.

Fixes #6329

### DIFF
--- a/doc/man1/common/job-standard-io.rst
+++ b/doc/man1/common/job-standard-io.rst
@@ -1,3 +1,9 @@
+.. note::
+   Paths specified in the options :option:`--output`, :option:`--error`,
+   and :option:`--input` will be opened relative to the working directory
+   on the first node of the job allocation, not on the node from which
+   the job is submitted.
+
 .. option:: --input=FILENAME|RANKS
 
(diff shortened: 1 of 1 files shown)


garlick (approved):
Thanks!

Totally optional, but I don't think it's necessary to say "it's important to note" inside of a note box.  But I don't think anyone will be offended :-)

grondo:
Oh that's a really good point. Removed that and greatly simplified the wording.

grondo:
Setting MWP.

PR #6335, opened 2024-10-01
include all not drained hostnames in `flux resource undrain` error message

grondo:
By default `flux resource undrain` aborts the entire operation if any target is not drained, but it only includes the first rank in the error message which has proved confusing.

This PR updates the error message to include all not drained hostnames (and their corresponding ranks).


garlick (approved):
LGTM!

grondo:
Thanks! I'll set MWP.

PR #6339, opened 2024-10-02
improve policy/queues config error messages

garlick:
Problem: when an error is encountered in the [policy] or [queues] TOML configuration tables, the error messages don't leave much to go on, and the fact that the code may be logged from different modules tends to set people off on the wrong path.

Add contextual information.

t2245-policy-config.t covers this code and now produces the following errors:

```
    error parsing [policy] config table: 1 object item(s) left unpacked: foo
    error parsing [policy.jobspec] config table: 1 object item(s) left unpacked: foo
    error parsing [policy.jobspec] config table: 'defaults.system.duration' is not a valid FSD
    error parsing [policy.jobspec] config table: 'defaults.system.queue' is not a string
    error parsing [policy.limits] config table: 1 object item(s) left unpacked: foo
    error parsing [policy.limits.job-size] config table: 1 object item(s) left unpacked: foo
    error parsing [policy.limits.max] config table: 1 object item(s) left unpacked: foo
    error parsing [policy.limits.min] config table: 1 object item(s) left unpacked: foo
    error parsing [policy.limits.min] config table: values must be >= -1
    error parsing [policy.limits] config table: 'duration' is not a valid FSD
    error parsing [policy.access] config table: 1 object item(s) left unpacked: foo
    error parsing [policy.access] config table: 'allow-user' must be a string array
    error parsing [policy.access] config table: 'allow-group' must be a string array
    error parsing [queues] config table: not a table
    error parsing [queues.x.policy] config table: 1 object item(s) left unpacked: foo
    error parsing [queues.x.policy.limits] config table: 'duration' is not a valid FSD
    error parsing [queues.queues.x.policy] config table: 'policy' must not define a default queue!
    the [policy] config table defines a default queue bar that is not in [queues] table
    error parsing [queues.x] config table: 'requires' must be an array of property strings (RFC 20)
```

### DIFF
--- a/src/common/libfluxutil/policy.c
+++ b/src/common/libfluxutil/policy.c
@@ -39,7 +39,8 @@ static int validate_policy_jobspec (json_t *o,
                           "system",
                             "duration", &duration,
                             "queue", &queue) < 0) {
-        errprintf (error, "%s: %s", key, jerror.text);
+        errprintf (error,
+                   "error parsing [%s] config table: %s", key, jerror.text);
         goto inval;
     }
     if (duration) {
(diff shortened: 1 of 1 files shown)


grondo (approved):
Great improvement!

garlick:
Thanks!  Just noticed there a [queues.queues...] in one of those messages.  Will fix then set MWP.

PR #6342, opened 2024-10-03
testsuite: add --raw-response opt to rpc test program

garlick:
Problem: `t/request/rpc` adds a NUL character to output.

Add a `-R,--raw-response` option which allows any raw payload to be output  when that matters, but otherwise omit the NUL character when the response is a string.
    
Update tests that require raw payloads.

Also do a bit of adjacent cleanup.


grondo (approved):
LGTM!

garlick:
Thanks!  MWP!

PR #6346, opened 2024-10-03
systemd: make scripts fail if systemctl start does

garlick:
Problem: flux-run-{prolog,epilog,housekeeping} scripts do not fail if systemctl start fails.

Replace "wait" with "wait $!".
The former always returns success.


grondo (approved):
LGTM! Thanks!

PR #6353, opened 2024-10-08
libsubprocess: support flow control on writes via credits

chu11:
Per https://github.com/flux-framework/rfc/pull/427 and discussion in #4572.

Some minor tweaks to the `add-credit` RPC.  I think the design in the RFC assumed multiple channels would return credits at the same time, but I think that is impractical, so it's just credits for each channel.

Haven't tested that this works by adding support in `flux-exec` yet.  Just thought I'd post this initial work.  May split off the cleanup into another PR.

### DIFF
--- a/src/common/libsubprocess/client.c
+++ b/src/common/libsubprocess/client.c
@@ -38,6 +38,7 @@ struct rexec_response {
     pid_t pid;
     int status;
     struct rexec_io io;
+    json_t *channels;
 };
 
 struct rexec_ctx {
@@ -53,6 +54,8 @@ static void rexec_response_clear (struct rexec_response *resp)
 {
     json_decref (resp->io.obj);
     free (resp->io.data);
+    json_decref (resp->channels);
+    resp->channels = NULL;
 
     memset (resp, 0, sizeof (*resp));
 
@@ -79,7 +82,8 @@ static struct rexec_ctx *rexec_ctx_create (flux_cmd_t *cmd,
     struct rexec_ctx *ctx;
     int valid_flags = SUBPROCESS_REXEC_STDOUT
         | SUBPROCESS_REXEC_STDERR
-        | SUBPROCESS_REXEC_CHANNEL;
+        | SUBPROCESS_REXEC_CHANNEL
+        | SUBPROCESS_REXEC_WRITE_CREDIT;
 
     if ((flags & ~valid_flags)) {
         errno = EINVAL;
@@ -150,11 +154,12 @@ int subprocess_rexec_get (flux_future_t *f)
     }
     rexec_response_clear (&ctx->response);
     if (flux_rpc_get_unpack (f,
-                             "{s:s s?i s?i s?O}",
+                             "{s:s s?i s?i s?O s?O}",
                              "type", &ctx->response.type,
                              "pid", &ctx->response.pid,
                              "status", &ctx->response.status,
-                             "io", &ctx->response.io.obj) < 0)
+                             "io", &ctx->response.io.obj,
+                             "channels", &ctx->response.channels) < 0)
         return -1;
     if (streq (ctx->response.type, "output")) {
         if (iodecode (ctx->response.io.obj,
@@ -164,6 +169,22 @@ int subprocess_rexec_get (flux_future_t *f)
                       &ctx->response.io.len,
                       &ctx->response.io.eof) < 0)
             return -1;
+    }
+    else if (streq (ctx->response.type, "add-credit")) {
+        const char *key;
+        json_t *value;
+
+        if (!ctx->response.channels
+            || !json_is_object (ctx->response.channels)) {
+            errno = EPROTO;
+            return -1;
+        }
+        json_object_foreach (ctx->response.channels, key, value) {
+            if (!json_is_integer (value)) {
+                errno = EPROTO;
+                return -1;
+            }
+        }
     }
     else if (!streq (ctx->response.type, "started")
         && !streq (ctx->response.type, "stopped")
@@ -228,6 +249,19 @@ bool subprocess_rexec_is_output (flux_future_t *f,
             *len = ctx->response.io.len;
         if (eof)
             *eof = ctx->response.io.eof;
+        return true;
+    }
+    return false;
+}
+
+bool subprocess_rexec_is_add_credit (flux_future_t *f, json_t **channels)
+{
+    struct rexec_ctx *ctx;
+    if ((ctx = flux_future_aux_get (f, "flux::rexec"))
+        && ctx->response.type != NULL
+        && streq (ctx->response.type, "add-credit")) {
+        if (channels)
+            (*channels) = ctx->response.channels;
         return true;
     }
     return false;
--- a/src/common/libsubprocess/fbuf.c
+++ b/src/common/libsubprocess/fbuf.c
@@ -145,17 +145,9 @@ bool fbuf_is_readonly (struct fbuf *fb)
     return fb->readonly;
 }
 
-static void nonempty_transition_check (struct fbuf *fb, bool was_empty)
+static void fbuf_notify (struct fbuf *fb, int old_used)
 {
-    if (was_empty && cbuf_used (fb->cbuf) > 0) {
-        if (fb->cb)
-            fb->cb (fb, fb->cb_arg);
-    }
-}
-
-static void nonfull_transition_check (struct fbuf *fb, bool was_full)
-{
-    if (was_full && cbuf_free (fb->cbuf) > 0) {
+    if (cbuf_used (fb->cbuf) != old_used) {
         if (fb->cb)
             fb->cb (fb, fb->cb_arg);
     }
@@ -209,7 +201,7 @@ const void *fbuf_read (struct fbuf *fb, int len, int *lenp)
     if (len > fb->buflen)
         len = fb->buflen;
 
-    bool full = cbuf_free (fb->cbuf) == 0 ? true : false;
+    int old_used = cbuf_used (fb->cbuf);
 
     if ((ret = cbuf_read (fb->cbuf, fb->buf, len)) < 0)
         return NULL;
@@ -218,7 +210,7 @@ const void *fbuf_read (struct fbuf *fb, int len, int *lenp)
     if (lenp)
         (*lenp) = ret;
 
-    nonfull_transition_check (fb, full);
+    fbuf_notify (fb, old_used);
 
     return fb->buf;
 }
@@ -237,12 +229,12 @@ int fbuf_write (struct fbuf *fb, const void *data, int len)
         return -1;
     }
 
-    bool empty = cbuf_is_empty (fb->cbuf);
+    int old_used = cbuf_used (fb->cbuf);
 
     if ((ret = cbuf_write (fb->cbuf, (void *)data, len, NULL)) < 0)
         return -1;
 
-    nonempty_transition_check (fb, empty);
+    fbuf_notify (fb, old_used);
 
     return ret;
 }
@@ -269,7 +261,7 @@ const void *fbuf_read_line (struct fbuf *fb, int *lenp)
     if (return_buffer_check (fb) < 0)
         return NULL;
 
-    bool full = cbuf_free (fb->cbuf) == 0 ? true : false;
+    int old_used = cbuf_used (fb->cbuf);
 
     if ((ret = cbuf_read_line (fb->cbuf, fb->buf, fb->buflen, 1)) < 0)
         return NULL;
@@ -277,7 +269,7 @@ const void *fbuf_read_line (struct fbuf *fb, int *lenp)
     if (lenp)
         (*lenp) = ret;
 
-    nonfull_transition_check (fb, full);
+    fbuf_notify (fb, old_used);
 
     return fb->buf;
 }
@@ -310,12 +302,12 @@ int fbuf_read_to_fd (struct fbuf *fb, int fd, int len)
         return -1;
     }
 
-    bool full = cbuf_free (fb->cbuf) == 0 ? true : false;
+    int old_used = cbuf_used (fb->cbuf);
 
     if ((ret = cbuf_read_to_fd (fb->cbuf, fd, len)) < 0)
         return -1;
 
-    nonfull_transition_check (fb, full);
+    fbuf_notify (fb, old_used);
 
     return ret;
 }
@@ -334,12 +326,12 @@ int fbuf_write_from_fd (struct fbuf *fb, int fd, int len)
         return -1;
     }
 
-    bool empty = cbuf_is_empty (fb->cbuf);
+    int old_used = cbuf_used (fb->cbuf);
 
     if ((ret = cbuf_write_from_fd (fb->cbuf, fd, len, NULL)) < 0)
         return -1;
 
-    nonempty_transition_check (fb, empty);
+    fbuf_notify (fb, old_used);
 
     return ret;
 }
--- a/src/common/libsubprocess/local.c
+++ b/src/common/libsubprocess/local.c
@@ -28,6 +28,7 @@
 #include "subprocess.h"
 #include "subprocess_private.h"
 #include "command_private.h"
+#include "fbuf.h"
 #include "local.h"
 #include "fork.h"
 #include "posix_spawn.h"
@@ -81,16 +82,65 @@ static void local_in_cb (flux_reactor_t *r,
     struct subprocess_channel *c = (struct subprocess_channel *)arg;
     int err = 0;
 
-    if (fbuf_write_watcher_is_closed (w, &err) == 1) {
-        if (err) {
-            llog_error (c->p,
-                        "fbuf_write_watcher close error: %s",
-                        strerror (err));
+    /* always a chance caller may destroy subprocess in callback */
+    subprocess_incref (c->p);
+
+    if (revents & FLUX_POLLOUT) {
+        if (fbuf_write_watcher_is_closed (w, &err) == 1) {
+            if (err) {
+                llog_error (c->p,
+                            "fbuf_write_watcher close error: %s",
+                            strerror (err));
+            }
+            else
+                c->parent_fd = -1;  /* closed by reactor */
+            flux_watcher_stop (w);  /* c->buffer_write_w */
+            local_channel_flush (c);
         }
-        else
-            c->parent_fd = -1;  /* closed by reactor */
-        flux_watcher_stop (w);  /* c->buffer_write_w */
-        local_channel_flush (c);
+        else {
+            /* If watcher is not closed, we were alerted to a change
+             * in buffer space
+             *
+             * - c->buffer_space is set to the buffer size on creation
+             * - c->buffer_space is reduced by flux_subprocess_write()
+             * - it is legal to call flux_subprocess_write() before
+             *   the first on_credit() callback
+             * - the fbuf write watcher specially raises POLLOUT
+             *   initially for initial credits
+             * - if data has already been written and moved out of the
+             *   buffer at that time, POLLOUT is raised again
+             *   (separately) after the initial callback returns
+             */
+            struct fbuf *fb;
+
+            if (!(fb = fbuf_write_watcher_get_buffer (c->buffer_write_w))) {
+                llog_error (c->p,
+                            "fbuf_write_watcher_get_buffer: %s",
+                            strerror (errno));
+                goto out;
+            }
+
+            if (!c->initial_credits_sent) {
+                c->initial_credits_sent = true;
+                if (c->p->ops.on_credit)
+                    c->p->ops.on_credit (c->p, c->name, fbuf_size (fb));
+            }
+            else {
+                int new_buffer_space = fbuf_space (fb);
+                /* Only report credits when space is reclaimed, not when
+                 * space is used up */
+                if (new_buffer_space > c->buffer_space) {
+                    int credits = new_buffer_space - c->buffer_space;
+                    /* N.B. Update buffer_space BEFORE the
+                     * on_credit callback, as there is a good chance
+                     * on_credit callback may call
+                     * flux_subprocess_write() */
+                    c->buffer_space = new_buffer_space;
+                    if (c->p->ops.on_credit)
+                        c->p->ops.on_credit (c->p, c->name, credits);
+                }
+            }
+        }
     }
     else {
         llog_error (c->p,
@@ -99,6 +149,8 @@ static void local_in_cb (flux_reactor_t *r,
                     revents,
                     strerror (errno));
     }
+out:
+    subprocess_decref (c->p);
 }
 
 static void local_output (struct subprocess_channel *c,
@@ -222,6 +274,8 @@ static int channel_local_setup (flux_subprocess_t *p,
     }
 
     if ((channel_flags & CHANNEL_WRITE) && in_cb) {
+        struct fbuf *fb;
+
         c->buffer_write_w = fbuf_write_watcher_create (p->reactor,
                                                        c->parent_fd,
                                                        buffer_size,
@@ -230,6 +284,18 @@ static int channel_local_setup (flux_subprocess_t *p,
                                                        c);
         if (!c->buffer_write_w) {
             llog_debug (p, "fbuf_write_watcher_create: %s", strerror (errno));
+            goto error;
+        }
+
+        if (!(fb = fbuf_write_watcher_get_buffer (c->buffer_write_w))) {
+            llog_debug (p,
+                        "fbuf_write_watcher_get_buffer: %s",
+                        strerror (errno));
+            goto error;
+        }
+
+        if ((c->buffer_space = fbuf_size (fb)) < 0) {
+            llog_debug (p, "fbuf_size: %s", strerror (errno));
             goto error;
         }
     }
@@ -454,6 +520,7 @@ static int start_local_watchers (flux_subprocess_t *p)
         c->buffer_read_w_started = true;
         c = zhash_next (p->channels);
     }
+
     return 0;
 }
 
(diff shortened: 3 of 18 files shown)


garlick:
> I think the design in the RFC assumed multiple channels would return credits at the same time, but I think that is impractical, so it's just credits for each channel.

In that PR I proposed that an `add-credit` response payload would be
```json
{
  "type":"add-credit",
  "channels":{
    "stdin":42
  }
}
```
while I think you are using
```json
{
  "type":"add-credit",
  "channel":{
    "stream":"stdin",
    "bytes":42
  }
}
```
The former is fewer bytes on the wire and leaves the door open to combine messages for multiple streams, should that be useful.  What issue did you run into?

BTW the main use case I was thinking of for combining messages was at initialization (server sends `{"channel":BUFSIZE}` for all input buffers).

chu11:
> BTW the main use case I was thinking of for combining messages was at initialization (server sends {"channel":BUFSIZE} for all input buffers).

Lemme see if I can refactor it that way.

chu11:
> BTW the main use case I was thinking of for combining messages was at initialization (server sends {"channel":BUFSIZE} for all input buffers).

Ahhh tried to refactor and realized why it initially didn't work out.  The `on_credit()` callback is

```
typedef void (*flux_subprocess_credit_f) (flux_subprocess_t *p,
                                          const char *stream,
                                          int bytes);
```

i.e. only one stream and "bytes" is reported at a time.  For remote subprocesses I just report the initial credits that the local subprocess reports.  

Could have a special case flag where local subprocess doesn't report initial credits and remote subprocess could "collate" them in one RPC instead?

garlick:
It's a bit unfortunate if we have to wait for the reactor loop to tell us the initial buffer size before sending it to the remote user since we _should_ already know what it is and that adds latency.

Still, is there a reason not to support multiple streams in the protocol even if this server doesn't use it yet?  There are other subprocess servers (`sdexec`).

chu11:
added tweaks to `flux-exec` that appear to handle the reproducer from #4572, hooray.  Gotta work on the edges of all this though (including comments from above).

garlick:
Ooh nice, does it work for an even smaller buffer like 4K?

chu11:
> Ooh nice, does it work for an even smaller buffer like 4K?

Yup, seems to work!

chu11:
re-pushed, fixing up a lot of things based on comments above.  Most notably, the remote subprocess sends out a single "initial credits" RPC back to the client for all channels.  Could use some tests that have more than just the single stdin channel.  That's a TODO.  But I think I can split out an obvious commit from here.

garlick:
Is this ready for another look?  I wasn't sure if you were still working on it.

chu11:
> Is this ready for another look? I wasn't sure if you were still working on it.

Good timing!  I was cleaning it up a bit and think it's ready for a review.  Removed WIP.

src/common/libsubprocess/fbuf.c
```diff
@@ -32,6 +32,8 @@ struct fbuf {
     int buflen;
     fbuf_notify_f notify_cb;
     void *notify_cb_arg;
+    fbuf_credit_f credit_cb;
+    void *credit_cb_arg;
```
garlick, line 36: An observation is that the original notify callback could be modified so that it is also called whenever `cbuf_used()` changes.  This would not break the existing users (ev_fbuf_read and ev_fbuf_write).

In fact it would allow ev_fbuf_write to handle this "event" as well.  Wouldn't that be more in keeping with the existing design than having the new fbuf callback invoke the subprocess `on_credit()` callback directly?

IOW arrange for the the fbuf_write_watcher to have a FLUX_POLLOUT event when the value of `fbuf_bytes()` has changed because buffer data has been written.  Then instead of having separate `local_in_cb()` and `local_in_credit_cb()` callbacks (one a reactor watcher and one not), just have the `local_in_cb()` reactor watcher.

I think this would be simpler and more consistent with the original design, and thus less confusing?

chu11:
re-pushed, changing `REXEC_CREDIT` to `REXEC_WRITE_CREDIT` per https://github.com/flux-framework/rfc/pull/427 updates

garlick (commented):
I think I'd better get your feedback on my one comment so far before I proceed, as addressing it would change things a bit.

src/common/libsubprocess/local.c
```diff
@@ -438,11 +490,30 @@ static int start_local_watchers (flux_subprocess_t *p)
...
+                                                                initial_credits_cb,
+                                                                p);
+            if (!p->initial_credits_w) {
+                llog_debug (p,
+                            "flux_prepare_watcher_create: %s",
+                            strerror (errno));
+                return -1;
+            }
+            flux_watcher_start (p->initial_credits_w);
+        }
```
garlick, line 514: Maybe we should eliminate the FLUX_SUBPROCESS_FLAGS_NO_INITIAL_CREDIT flag and _never_ make the initail `on_credit()` callback.  A subprocess user (local or remote) should already know the size of all their buffers.  Either it's SUBPROCESS_DEFAULT_BUFSIZE (now in a public heador) or they explicitly set it.

There is a fair bit of complexity here just to tell the user what they should already know.

garlick: Oh well, SUBPROCESS_DEFAULT_BUFSIZE doesn't apply to sdexec but we could fix that.

chu11: My first implementation did exactly what you just described, but I felt the "initial credits" was ultimately important from a programming perspective.

Many times `flux_subprocess_write()` will be called within the `on_credit()` callback.  So we need something to "kickstart" writing to stdin, and the initial credits is what "kickstarts" it.  Otherwise `on_credit()` will never be called.

garlick: The main case I'm always thinking of is copying from a file descriptor to a subprocess.  In that case, it seems like you'd probably have an fd  watcher that you'd stop and start based on credit availability and do the writes from there?  This proposal wouldn't affect that case much - just whether the fd watcher begins started or has to wait for a callback to start.

Maybe an example would help?

src/common/libsubprocess/fbuf.c
chu11: hmmm.  I see where you're coming from.  I think it is plausible for this work.  However, I'm not sure the overall changes would make it worthwhile.  

The main thing is the fbuf notify callback is currently an "edge" callback, only getting called on "full -> not full" and "empty -> not empty" transitions.  If we convert it to handle byte changes, we need to change that callback to deal with "before" and "after" byte changes.  Worst case, callback function prototype changes everywhere.  Or more annoying case, have to keep track of how many bytes were previously in the buffer and do math after calling `fbuf_bytes()` to see current usage.

Lemme think about this more and see if there's a simpler solution.  But it seems like there'd be a ton of fallout as a result.

garlick: When I was thinking about it earlier I tried it to see what broke.  That attempt is here FWIW:  db1214a1fde4c04c8039030ebd7fe8e32d32c83d

src/common/libsubprocess/local.c
chu11: > Maybe an example would help?

Take a look at the unit tests I wrote in `libsubprocess/test/stdio.c`, that sort of illustrates the two styles.  One where all the writing is within the credit callback and the other where the first write is outside of the credit callback b/c the NO_INITIAL_CREDIT flag is set.  The second case requires calling `flux_subprocess_write()` in two places.   Which I guess is why I thought sending the initial credits made things easier, the writing is just in one place.

With your use case example, we'd call `flux_watcher_start()` in two places vs one.

src/common/libsubprocess/fbuf.c
chu11: Yeah, I don't think `fbuf` would change that much but ...

```
static void fbuf_notify (struct fbuf *fb, int old_used)
{
    if (cbuf_used (fb->cbuf) != old_used) {
        if (fb->cb)
            fb->cb (fb, fb->cb_arg);
    }
}
```

B/c we're not passing the old and/or new bytes to the callback, something has to be done there to inform the callback what happened.  So that's what I'm thinking could really change things.

chu11: as I slept on this last night, I began moving towards "it's a lot of work, but may be worth it" to move towards this approach.  I think everything is contained within `libsubprocess` and won't bleed outside of it.  I think minimally, lets give it a shot.

garlick: Let me know if I can help.  Generally, I think the main role of the watcher is to trigger a callback when the reactor would otherwise sleep, and the "action" should be something taken by the callback itself.  So probably the subprocess channel  sttruct should track the available space in the buffer and the watcher callback should just check if that has changed.

chu11:
re-pushed, re-working the PR per discussion above

src/common/libsubprocess/subprocess.h
```diff
@@ -73,6 +73,12 @@ enum {
      * error.
      */
     FLUX_SUBPROCESS_FLAGS_LOCAL_UNBUF = 8,
+    /* flux_local_exec(): when an on_credit callback is specified, the
+     * full initial credits are sent as the first callback.  However,
+     * in some circumstances this may not be desired.  This flag will
+     * disable the sending of "initial credits".
+     */
+     FLUX_SUBPROCESS_FLAGS_NO_INITIAL_CREDITS = 16,
```
garlick, line 81: As stated in the earlier comment, this flag feels hacky to me.  

Proposal: eliminate it and always send the first `on_credit` callback.

Subprocess users that would have used this flag, including the subprocess server, would track the current available buffer space.  If they want to issue remote credits or write data before the first `on_credit()` callback, they would set the current buffer space negative indicating that some number of credits were borrowed.  The first callback would cause the tracked count to reach zero (no action should be taken because there is no available buffer space).  The next call would result in a positive value.

src/common/libsubprocess/command.h
```diff
@@ -25,6 +25,8 @@ extern "C" {
  */
 typedef struct flux_command flux_cmd_t;
 
+#define SUBPROCESS_DEFAULT_BUFSIZE 4194304
```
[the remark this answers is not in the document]
chu11: I guess that's true.  The macro isn't explicitly necessary, I guess I just found it convenient.  Can remove.

src/common/libsubprocess/local.c
```diff
@@ -167,6 +200,36 @@ static void local_stderr_cb (flux_reactor_t *r, flux_watcher_t *w,
...
+    if (!p->ops.on_credit)
+        return;
+
+    c = zhash_first (p->channels);
+    while (c) {
+        if (c->flags & CHANNEL_WRITE) {
+            /* N.B. send cmd_option_bufsize() result not
+             * c->buffer_write_credits, to avoid chance
+             * c->buffer_write_credits already altered.
+             */
```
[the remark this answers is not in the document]
garlick: Oh I think maybe I committed a thinko here and your code is right!

If a client writes to the buffer before receiving credit, it should be up to the client to account for that.

src/common/libsubprocess/subprocess.h
chu11: > Subprocess users that would have used this flag, including the subprocess server, would track the current available buffer space. If they want to issue remote credits or write data before the first on_credit() callback, they would set the current buffer space negative indicating that some number of credits were borrowed. The first callback would cause the tracked count to reach zero (no action should be taken because there is no available buffer space). The next call would result in a positive value.

Going into the weeds, I remember why I did this flag.  The `on_credit()` callback's prototype is

`void cb (flux_subprocess_t *p, const char *stream, int bytes)`

So if we wanted to track "current buffer space" (or more simply a flag indicating "did we already get initial credits") we have to use a hash to lookup the stream.  

So I felt it was better to use the flag.

A hash lookup on every `on_credit()` call is certainly not the end of the world.  It is what it is.  But I did dislike adding that data structure / lookup.

One solution to avoid the hash is to remove the server's "initial credits" response which groups all of the channel's initial credits into one RPC response.  Instead, we just forward each channel's "initial credits" from the local subprocess, i.e. multiple RPC responses if there are multiple writable channels.

So it's a series of pros and cons on each of the approaches.  Any strong feelings on any of them?  To summarize

- option 1 (what's currently done): flag for no initial credits from local subprocess, the server sends 1 RPC for all initial channel credits

- option 2 (what I believe you proposed): track "space" for each writable channel, only respond with non-initial credits.  Requires a hash and a lookup on each credit callback from local subprocess.

- option 3 (what I propose above): do not have server collect all writable channel credits in 1 RPC, forward each RPC from local subprocess.  multiple RPC responses, but no hash necessary.

garlick: My proposal this morning is not option 2, it is:

> What about a compromise where we eliminate the flag and the dedicated watcher for the first callback, and just always send the first on_credit() response?

> So if we wanted to track "current buffer space" (or more simply a flag indicating "did we already get initial credits") we have to use a hash to lookup the stream.

My thought was just that those that want to immediately send some data without waiting for the first on_credit() response would "borrow" credits by setting their tally negative.  Then when the initial callback comes, it is added to the negative tally.

Isn't there already a hash of channels in `struct flux_subprocess`?  Would adding a `channel_lookup()` function help?  And maybe add a channel aux container?

chu11:
> My thought was just that those that want to immediately send some data without waiting for the first on_credit() response would "borrow" credits by setting their tally negative. Then when the initial callback comes, it is added to the negative tally.

Yes, but that would require tracking of the credits on the libsubprocess server side, thus the hash lookup on each `on_credit()` call.

> Isn't there already a hash of channels in struct flux_subprocess? Would adding a channel_lookup() function help? And maybe add a channel aux container?

Oh duh, you're right, there should already be the channel hash.  

I dislike the lookup of the channel on every `on_credit()` call, but I guess it's not a huge deal in the end.

src/common/libsubprocess/subprocess.h
chu11: > My thought was just that those that want to immediately send some data without waiting for the first on_credit() response would "borrow" credits by setting their tally negative. Then when the initial callback comes, it is added to the negative tally.

Yes, but that would require tracking of the credits on the libsubprocess server side, thus the hash lookup on each on_credit() call.

> Isn't there already a hash of channels in struct flux_subprocess? Would adding a channel_lookup() function help? And maybe add a channel aux container?

Oh duh, you're right, there should already be the channel hash.

I dislike the lookup of the channel on every on_credit() call, but I guess it's not a huge deal in the end.

chu11: I just realized a subtlety in this conversation, which maybe explains some of comments/responses above.

At one point in time I viewed `FLUX_SUBPROCESS_FLAGS_NO_INITIAL_CREDITS` as a useful flag that users could use and wanted it supported for both local & remote subprocesses.

Now, it is simply a convenience flag used pretty much exclusively by the server when launching its local subprocesses.  

So I'm not sure if that removes some feeling of it being "hackish".  Perhaps if there was a hidden/internal flag that could do the same behavior, it wouldn't be viewed as such a hack?

garlick: Yes, avoiding unnecessary API complexity was my biggest concern.  Making the libsubprocess internals as understandable as possible for the next person who comes along to fix a bug or whatever is a secondary concern, though.

The hash lookup doesn't seem like a big deal to me.  When the server handles a `rexec.write` RPC, it has to find the channel with a hash lookup.  This is just the same in the reverse direction. If the hash lookup helps the code be more understandable by making the subprocess server behave like any subprocess user implementing flow control would, then I'm for the hash lookup.

garlick: Uh, I was riding my bike and it occurred to me that your option 3 above would be fine!  For most cases, there is only one writable channel (maybe two), so having an RPC response per writable channel should be fine.

It seems like that would achieve the goal of simplifying the code best of all since the first response wouldn't need to be handled by a prep watcher.  The fbuf write watcher could just be started and an initial POLLOUT event would let the credits be sent the same way credits are sent after that.

chu11: > Uh, I was riding my bike and it occurred to me that your option 3 above would be fine! For most cases, there is only one writable channel (maybe two), so having an RPC response per writable channel should be fine.

Shall we keep the "channels" object that has the potential for > 1 channel per object?  No longer needed now, but it's always possible we may > 1 channel in the future.

garlick: Yeah seems like it would be OK to leave it as proposed unless it causes you grief.

src/common/libsubprocess/fbuf.c
```diff
@@ -145,17 +145,9 @@ bool fbuf_is_readonly (struct fbuf *fb)
     return fb->readonly;
```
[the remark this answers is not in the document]
garlick: This commit message still needs to be fixed up (see earlier comment).

src/common/libsubprocess/local.c
```diff
@@ -178,6 +216,36 @@ static void local_stderr_cb (flux_reactor_t *r, flux_watcher_t *w,
...
+    while (c) {
+        if (c->flags & CHANNEL_WRITE) {
+            /* N.B. send cmd_option_bufsize() result not
+             * c->buffer_space, to avoid chance
+             * c->buffer_space already altered.
+             */
+            int bufsize = cmd_option_bufsize (p, c->name);
+            /* should have been checked at setup */
+            assert (bufsize > 0);
+            p->ops.on_credit (p, c->name, bufsize);
```
garlick, line 240: Not great to assert in here.  How about just 
```c
if (bufsize > 0)
    p->ops.on_credit (p, c->name, bufsize);
```

src/common/libsubprocess/local.c
```diff
@@ -178,6 +216,36 @@ static void local_stderr_cb (flux_reactor_t *r, flux_watcher_t *w,
...
+        }
+        c = zhash_next (p->channels);
+    }
+
+    /* initial credits sent, we don't know need this watcher anymore */
+    flux_watcher_stop (p->initial_credits_w);
```
garlick, line 246: Do we still need a dedicated watcher and code path for the initial callback or can fbuf callbacks drive the initial callback for each channel individually?

garlick (commented):
Nice - couple comments inline.
The big one is why is the initial on_credit callback still special cased with its own watcher?

src/common/libsubprocess/local.c
chu11: I haven't figured out a good way to do it otherwise (doesn't mean there isn't a way, i just don't see it).  The key is that we have to return the subprocess object to the caller before we can send the initial credits.  So that suggests we need to setup something to send the initial credits after the subprocess object is returned to the caller.  A watcher is the obvious choice.

src/common/libsubprocess/fbuf.c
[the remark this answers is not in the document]
chu11: oops.  I guess I cherry-picked an old copy while rebasing work.

src/common/libsubprocess/local.c
garlick: My thought was have the fbuf write watcher do it, like have it call `ev_fbuf_write_wakeup()` on creation so as soon as the reactor runs, there is a callback?

chu11: The `ev_fbuf_write_wakeup()` only wakes up the IO watcher.  And in order for the IO watcher callback to be triggered, there'd have to be activity on the file descriptor, which we don't control.

garlick: Yeah but it's a POLLOUT and if we haven't written anything yet it's guaranteed(?) to fire.

chu11: ahhh ok, I see where you're going with this now.

I'll have to ponder this and/or maybe give it a shot tomorrow.  This does feel a little hack-ish to me.  We're now creating an exception to start the IO watcher when there isn't data in the buffer to write out yet.

Also, per a prior discussion, I think we want the initial credits sent out via a prepare watcher.  The IO watcher callback is presumably a check watcher.

garlick: Re: prepare watcher.  Well, early on I thought it was critical for the subprocess server to send `add-credit` responses right away to avoid contributing to the latency of remotely executing a fast running command.  Since then we discussed "borrowing credit" which would allow a small amount of data to be sent without waiting for the initial `add-credit` response.  So I think my position has changed on that.  Did you have a different reason for wanting to expedite the first callbacks?

Regarding hackishness, if the proper functioning of the fbuf write watcher were to depened on the `fd` passed into `fbuf_write_watcher_create()` being immediatley ready, that precondition could be checked with `poll(2)` in the create function and it could fail with EINVAL if it's not met.  Since it's a private function to libsubprocess and there is only one user, I don't think this will be a liability. In reality, if the fd is not ready, I think it would only delay the callback, which would still be "correct", so I don't think that check is really needed.l

If we do end up needing to use a different watcher for the first response, I suggest we put it down in the fbuf  write watcher not at the higher level.  Then at least it becomes part of the same abstraction that's activating the watcher when buffer space becomes available.  The first response is the same case really - all the buffer space is available.

chu11: > Since then we discussed "borrowing credit" which would allow a small amount of data to be sent without waiting for the initial add-credit response. So I think my position has changed on that. Did you have a different reason for wanting to expedite the first callbacks?

I think the main reason is that a common use case will still be to start writing after the first credit response.  i.e. put all of the writing in the `on_credit()` callback.  So it feels like a good idea to send that quick.

> If we do end up needing to use a different watcher for the first response, I suggest we put it down in the fbuf write watcher not at the higher level. 

That seems like a good idea.

garlick: Test cases aside, where performance isn't critical, what's the common case where the _data_ to write is always available and availability of buffer space is the sole driver?

chu11: good point, I had been thinking of how I implemented things in #6370, but I had not tried using the "negative credits" idea in there yet.  So perhaps it's not as big a deal as I originally thought.

garlick: Perhaps I should update the RFC PR to reflect some of the things we've discussed.  Progress!

garlick: Update pushed to flux-framework/rfc#427.

chu11:
re-pushed per comments / discussion above.

Moving the "initial credits" to the fbuf watcher wasn't as ugly as I thought it would be.  The main trickiness is that we have to "special case" the first "initial credits" callback at both the `fbuf_watcher` level and the local subprocess level.  That's really it.

src/common/libsubprocess/subprocess.c
```diff
@@ -79,6 +79,8 @@ struct subprocess_channel *channel_create (flux_subprocess_t *p,
     c->child_fd = -1;
     if (!(c->name = strdup (name)))
         goto error;
+    if ((c->buffer_space = cmd_option_bufsize (p, name)) < 0)
+        goto error;
```
garlick, line 83: IMHO it would be clearer to initialize `c->buffer_space` in `channel_local_setup()` where the buffer is created, and then use `fbuf_space()` right after creation as the initial value.

It's strictly used for local subprocesses and initializing it here makes that less obvious.

src/common/libsubprocess/subprocess.h
```diff
@@ -104,6 +109,8 @@ typedef struct {
     flux_subprocess_output_f on_channel_out; /* Read from channel when ready */
     flux_subprocess_output_f on_stdout; /* Read of stdout is ready           */
     flux_subprocess_output_f on_stderr; /* Read of stderr is ready           */
+    flux_subprocess_credit_f on_credit; /* bytes of write buffer space
+                                         * reclaimed */
```
garlick, line 113: Maybe just say `// Write buffer space available` for the inline comment, since this is also called for initial credits, where there is no reclamation.

src/common/libsubprocess/subprocess_private.h
```diff
@@ -36,6 +36,8 @@ struct subprocess_channel {
     int parent_fd;
     int child_fd;
     flux_watcher_t *buffer_write_w;
+    int buffer_space;
+    bool initial_credits_sent;
```
garlick, line 40: Possibly usefull to comment `// accounting for on_credit callback`

src/common/libsubprocess/local.c
```diff
@@ -81,16 +82,55 @@ static void local_in_cb (flux_reactor_t *r,
...
+                    int credits = new_buffer_space - c->buffer_space;
+                    /* N.B. Update buffer_space BEFORE the
+                     * on_credit callback, as there is a good chance
+                     * on_credit callback may call
+                     * flux_subprocess_write() */
+                    c->buffer_space = new_buffer_space;
+                    if (c->p->ops.on_credit)
+                        c->p->ops.on_credit (c->p, c->name, credits);
+                }
+            }
```
garlick, line 132: This might deserve a block comment pointing out
- `c->buffer_space` is set to the buffer size on creation
- `c->buffer_space` is reduced by `flux_subprocess_write()`
- it is legal to call `flux_subprocess_write()` before the first `on_credit()` callback
- the fbuf write watcher specially raises POLLOUT initially for initial credits
- if data has already been written and moved out of the buffer at that time, POLLOUT is raised again (separately) after the initial callback returns

src/common/libsubprocess/client.c
```diff
@@ -38,6 +38,7 @@ struct rexec_response {
     pid_t pid;
     int status;
```
garlick, line 39: Commit message suggestion: describe details of the fbuf behavior change in its own paragraph.

garlick (commented):
Looking great!  Here are some comments just reviewing the code, not the tests.  All my comments are fairly superficial.  Great work here.

I'll try the flux-exec changes on top of this on my test cluster and see how it goes moving some serious data and watching the message traces.

chu11:
re-pushed, fixing up from comments above

garlick (approved):
I think this is working as advertised!  Awesome.

Also, since nothing really uses the flow control yet I don't think this could have much effect on critical stuff like stdin handling in the shell or PMI.

I say let's merge!

Thank you for all the iterations!

chu11:
sounds good, setting MWP

Issue #6354, opened 2024-10-08
Compilation problem when following manual installation


tphung3:
Hi, I'm trying out flux-core by manual installation as outlined in https://flux-framework.readthedocs.io/en/latest/quickstart.html#manual-installation, but I'm having this compilation error at the step `make -j 8`. Is installing flux-core manually still supported or should I try something else?
```
make[3]: Entering directory '.../flux-core/src/common/libflux'
  CC       flog.lo
  CC       attr.lo
  CC       handle.lo
  CC       msg_deque.lo
  CC       connector_loop.lo
  CC       connector_interthread.lo
  CC       connector_local.lo
  CC       reactor.lo
  CC       msg_handler.lo
  CC       message.lo
  CC       message_iovec.lo
  CC       message_route.lo
  CC       message_proto.lo
  CC       msglist.lo
message.c: In function 'userid2str':
message.c:1244:9: error: variable 'n' set but not used [-Werror=unused-but-set-variable]
 1244 |     int n;
      |         ^
message.c: In function 'nodeid2str':
message.c:1289:9: error: variable 'n' set but not used [-Werror=unused-but-set-variable]
 1289 |     int n;
      |         ^
  CC       request.lo
  CC       response.lo
  CC       rpc.lo
  CC       event.lo
cc1: all warnings being treated as errors
  CC       module.lo
make[3]: *** [Makefile:1414: message.lo] Error 1
make[3]: *** Waiting for unfinished jobs....
make[3]: Leaving directory '.../flux-core/src/common/libflux'
make[2]: *** [Makefile:1026: all-recursive] Error 1
make[2]: Leaving directory '.../flux-core/src/common'
make[1]: *** [Makefile:520: all-recursive] Error 1
make[1]: Leaving directory '.../flux-core/src'
make: *** [Makefile:602: all-recursive] Error 1
```

garlick:
Hi, yes it is definitely supported!  What compiler and OS are you using?

garlick:
```
static void userid2str (uint32_t userid, char *buf, int buflen)
{
    int n;
    if (userid == FLUX_USERID_UNKNOWN)
        n = snprintf (buf, buflen, "unknown");
    else
        n = snprintf (buf, buflen, "%u", userid);
    assert (n < buflen);
}
```
Ah we generally build with asserts compiled in.  if yo'ure adding `-DNDEBUG` to CFLAGS, can you remove it?

grondo:
[...]

PR #6355, opened 2024-10-08
allow project to be built with NDEBUG

garlick:
Problem: as noted in #6354, we get some compiler warnings (promoted to errors) when the project is built with `-DNDEBUG`.

@grondo pointed out that we missed adding CODE_COVERAGE_CPPFLAGS to our Makefiles, otherwise we would have been building with that flag in CI.

So this PR adds CODE_COVERAGE_CPPFLAGS to AM_CPPFLAGS in all our Makefiles, and then addresses the warnings that result in a way that seemed appropriate for the context, one of:
- remove the assertion
- convert assertion to warning
- declare assertion-only vars with `__attribute__((unused))`


grondo (approved):
Wow, thank you! LGTM!

garlick:
Thanks! Setting MWP.

PR #6359, opened 2024-10-10
show response result in message traces

garlick:
Problem: `flux module trace` and `flux overlay trace` don't show the `errnum` value for response messages, but this would be helpful when interpreting message traces.

Append  "success" or the error name associated with the `errnum` value in the one line trace output for responses.
When `--full` is specified, include human readable error message (if any) from  the payload.
Change the color of failure response messages to bright red so they stand out.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -231,6 +231,7 @@ AC_REPLACE_FUNCS( \
   strlcat \
   argz_add \
   envz_add \
+  strerrorname_np \
 )
 X_AC_CHECK_PTHREADS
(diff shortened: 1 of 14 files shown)


grondo (approved):
This LGTM. Not sure if it makes sense to add some tests...

garlick:
Yeah could probably add some tests.

garlick:
Added a little bit more testing (though it didn't do much for the numbers).  Anyway, setting MWP.

PR #6364, opened 2024-10-11
flux-jobs(1): document unlimited --count value

### LINKED ISSUE #6363: flux-jobs(1) does not document that --count=0 means "unlimited"
It just says
```
      -c, --count=N
              Limit output to N jobs (default 1000)
```

garlick:
Problem: flux jobs --count=0 means "unlimited" but that is not documented in the man page.

Add it.

Fixes #6363

### DIFF
--- a/doc/man1/flux-jobs.rst
+++ b/doc/man1/flux-jobs.rst
@@ -60,7 +60,7 @@ OPTIONS
 
 .. option:: -c, --count=N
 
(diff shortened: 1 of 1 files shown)


wihobbs (approved):
Thanks for this, I had no idea :)

garlick:
Thanks!  @ofaaland found that by trial and error and reported the documentation omission in email :-)  Will set MWP

PR #6369, opened 2024-10-14
doc: add debugging notes

garlick:
Problem: debugging flux is not always easy and we accumulated some tricks along the way that ought to be shared with new developers.

Start a page for this in the flux-core docs.


chu11 (approved):
seems like a good start to me!  I often have to grep the tracing b/c I can't remember exactly what the environment variable is named :-)

garlick:
Thanks!  Please add stuff as you think of it.  I had been meaning to do this for some time, but when it came time to do it, I kind of went blank and this was all I could come up with.

I'll set MWP.

PR #6370, opened 2024-10-15
flux-exec: use subprocess credits to avoid overflowing stdin buffers

chu11:
built on top of #6353, splitting it off since it's a separate issue

### DIFF
--- a/src/cmd/flux-exec.c
+++ b/src/cmd/flux-exec.c
@@ -72,7 +72,7 @@ int exit_code = 0;
 zhashx_t *exitsets;
 struct idset *hanging;
 
-zlist_t *subprocesses;
+zlistx_t *subprocesses;
 
 optparse_t *opts = NULL;
 
@@ -232,22 +232,22 @@ static void stdin_cb (flux_reactor_t *r,
         log_err_exit ("fbuf_read");
 
     if (lenp) {
-        p = zlist_first (subprocesses);
+        p = zlistx_first (subprocesses);
         while (p) {
             if (flux_subprocess_state (p) == FLUX_SUBPROCESS_INIT
                 || flux_subprocess_state (p) == FLUX_SUBPROCESS_RUNNING) {
                 if (flux_subprocess_write (p, "stdin", ptr, lenp) < 0)
                     log_err_exit ("flux_subprocess_write");
             }
-            p = zlist_next (subprocesses);
+            p = zlistx_next (subprocesses);
         }
     }
     else {
-        p = zlist_first (subprocesses);
+        p = zlistx_first (subprocesses);
         while (p) {
             if (flux_subprocess_close (p, "stdin") < 0)
                 log_err_exit ("flux_subprocess_close");
-            p = zlist_next (subprocesses);
+            p = zlistx_next (subprocesses);
         }
         flux_watcher_stop (stdin_w);
     }
@@ -291,9 +291,9 @@ static flux_subprocess_t *imp_kill (flux_subprocess_t *p, int signum)
                        &ops);
 }
 
-static void killall (zlist_t *l, int signum)
+static void killall (zlistx_t *l, int signum)
 {
-    flux_subprocess_t *p = zlist_first (l);
+    flux_subprocess_t *p = zlistx_first (l);
     while (p) {
         if (flux_subprocess_state (p) == FLUX_SUBPROCESS_RUNNING) {
             if (use_imp) {
@@ -316,7 +316,7 @@ static void killall (zlist_t *l, int signum)
                 flux_future_destroy (f);
             }
         }
-        p = zlist_next (l);
+        p = zlistx_next (l);
     }
 }
 
@@ -361,9 +361,9 @@ static void signal_cb (int signum)
     }
 }
 
-void subprocess_destroy (void *arg)
+void subprocess_destroy (void **arg)
 {
-    flux_subprocess_t *p = arg;
+    flux_subprocess_t *p = *arg;
     flux_subprocess_destroy (p);
 }
 
@@ -704,8 +704,9 @@ int main (int argc, char *argv[])
         free (nodeset);
     }
 
-    if (!(subprocesses = zlist_new ()))
-        log_err_exit ("zlist_new");
+    if (!(subprocesses = zlistx_new ()))
+        log_err_exit ("zlistx_new");
+    zlistx_set_destructor (subprocesses, subprocess_destroy);
 
     if (!(exitsets = zhashx_new ()))
         log_err_exit ("zhashx_new()");
@@ -725,10 +726,8 @@ int main (int argc, char *argv[])
                                  NULL,
                                  NULL)))
             log_err_exit ("flux_rexec");
-        if (zlist_append (subprocesses, p) < 0)
-            log_err_exit ("zlist_append");
-        if (!zlist_freefn (subprocesses, p, subprocess_destroy, true))
-            log_err_exit ("zlist_freefn");
+        if (!zlistx_add_end (subprocesses, p))
+            log_err_exit ("zlistx_add_end");
         rank = idset_next (targets, rank);
     }
 
@@ -739,11 +738,11 @@ int main (int argc, char *argv[])
      */
     if (optparse_getopt (opts, "noinput", NULL) > 0) {
         flux_subprocess_t *p;
-        p = zlist_first (subprocesses);
+        p = zlistx_first (subprocesses);
         while (p) {
             if (flux_subprocess_close (p, "stdin") < 0)
                 log_err_exit ("flux_subprocess_close");
-            p = zlist_next (subprocesses);
+            p = zlistx_next (subprocesses);
         }
     }
     /* configure stdin watcher
@@ -800,7 +799,7 @@ int main (int argc, char *argv[])
     log_fini ();
 
     zhashx_destroy (&exitsets);
-    zlist_destroy (&subprocesses);
+    zlistx_destroy (&subprocesses);
 
     return exit_code;
 }
--- a/src/cmd/flux-exec.c
+++ b/src/cmd/flux-exec.c
@@ -32,6 +32,8 @@
 #include "src/common/libsubprocess/fbuf_watcher.h"
 #include "ccan/str/str.h"
 
+#define NUMCMP(a,b) ((a)==(b)?0:((a)<(b)?-1:1))
+
 static struct optparse_option cmdopts[] = {
     { .name = "rank", .key = 'r', .has_arg = 1, .arginfo = "IDSET",
       .usage = "Specify target ranks.  Default is \"all\"" },
@@ -73,6 +75,15 @@ zhashx_t *exitsets;
 struct idset *hanging;
 
 zlistx_t *subprocesses;
+/* subprocess credits ordered low to high.  Exited and failed
+ * subprocesses are removed from the list.
+ */
+zlistx_t *subprocess_credits;
+
+struct subproc_credit {
+    void *handle;      /* handle to subprocess in credits list */
+    int credits;
+};
 
 optparse_t *opts = NULL;
 
@@ -146,6 +157,35 @@ void completion_cb (flux_subprocess_t *p)
         log_err_exit ("idset_clear");
 }
 
+int subprocess_min_credits (void)
+{
+    /* subprocess_credits ordered, min at head */
+    flux_subprocess_t *p = zlistx_head (subprocess_credits);
+    struct subproc_credit *spcred;
+    /* list possibly empty if all subprocesses failed, so return no
+     * credits so stdin watcher won't be started
+     */
+    if (!p)
+        return 0;
+    spcred = flux_subprocess_aux_get (p, "credits");
+    return spcred->credits;
+}
+
+void subprocess_update_credits (flux_subprocess_t *p, int bytes, bool reorder)
+{
+    struct subproc_credit *spcred = flux_subprocess_aux_get (p, "credits");
+    spcred->credits += bytes;
+    if (reorder)
+        zlistx_reorder (subprocess_credits, spcred->handle, false);
+}
+
+void subprocess_remove_credits (flux_subprocess_t *p)
+{
+    struct subproc_credit *spcred = flux_subprocess_aux_get (p, "credits");
+    if (zlistx_delete (subprocess_credits, spcred->handle) < 0)
+        log_err_exit ("zlistx_delete");
+}
+
 void state_cb (flux_subprocess_t *p, flux_subprocess_state_t state)
 {
     if (state == FLUX_SUBPROCESS_RUNNING) {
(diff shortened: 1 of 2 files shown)


chu11:
removed WIP, I think the tweak to `flux-exec` is good.  I wish we could avoid the looping calculation of `min_credits`, but I'm not sure that updating to a `min-heap` kind of data structure is worth the effort.

longer out, a `no credits` kind of option could be on the table if that loop poses some slowdown issue

src/common/libsubprocess/command.h
```diff
@@ -25,6 +25,8 @@ extern "C" {
  */
 typedef struct flux_command flux_cmd_t;
 
+#define SUBPROCESS_DEFAULT_BUFSIZE 4194304
+
```
garlick, line 29: Why is this needed in an installed header?

chu11:
as an aside, I looked into using "negative credits" to get the `flux-exec` stdin going faster.  For the moment, decided not to add it.  It would require some "did I send the first wave of data" flags, and didn't think it super worth it.

garlick:
I noticed two problems here:
- doesn't work with sdexec because sdexec doesn't yet accept the `write-credit` flag added to RFC 42.
- if one subprocess fails with zero credit, and the procs are reading stdin, the whole exec could deadlock as the failed proc credits will never be replenished

Also I worry a bit about about  iterating over all subprocs to find the minimum each time on_credit is called (think el cap size).

On the first problem, we could make stdin flow control optional, and then turn it off if the service is sdexec (for now).  That would also allow us to more carefully evaluate the performance vs what we had before.

On the performance concern, maybe subprocesses could go in a zlistx ordered by credit count.  When on_credit changes the credit value of one proc, call  `zlistx_reorder` with `low_value=false`.  When it changes for all after a write, the sort order shouldn't change since all procs' credit changes by the same amount.   Call `zlistx_first()` to find the minimum value.  Failed/exited procs could be removed from the list (triggering a check for the new minimum).

garlick:
FWIW this adds an option to enable/disable flow control and disables flow control by default if the service is sdexec:  6ede342301a76c31cb292ccb5b14d03632f034ee

chu11:
> On the performance concern, maybe subprocesses could go in a zlistx ordered by credit count. When on_credit changes the credit value of one proc, call zlistx_reorder with low_value=false. When it changes for all after a write, the sort order shouldn't change since all procs' credit changes by the same amount. Call zlistx_first() to find the minimum value. Failed/exited procs could be removed from the list (triggering a check for the new minimum).

Thinking about this, this probably would be faster, but it appears to be a slightly faster O(n) vs a slightly slower O(n) (i.e. the reorders should average n/2 iterations of the list vs n for min-credits, and when `stdin_cb` is called, we don't iterate the credits list).  

Per my comment above, maybe we want to go down the `min-heap` data structure path?  It seems ccan has a heap data structure.  So now we're guaranteed O(log n) for every min-credits reorder.

garlick:
Since we only reorder after adding credits (not when subtracting), it
_seems_ like the list should be O(1) for the common case where the same
number of credits is being added across many subprocesses.  IOW there will
be a bunch of clients that all end up with the same credit value at the end
of the list, so if we tell `zlistx_reorder()` to search for the new
position at the tail of the list after an add-credit response, it is likely
to be added as the last item, O(1).

I'm thinking that there won't be a lot of chaos in the way the server
responds since it's same server and same consumer process running in
parallel, and each parallel write sends the same number of bytes to all
ranks.

chu11:
> if one subprocess fails with zero credit, and the procs are reading stdin, the whole exec could deadlock as the failed proc credits will never be replenished

Skimming through the code, we don't handle this generically for `stdin` and subprocesses right now.  i.e. we can call `flux_subprocess_write()` for a subprocess that already failed.  So we should probably fix that first.  I'll write up an issue.

chu11:
re-pushed with fixes per discussion above, and added a tweaked version of the `--stdin-flow` option on top.

No tests at the moment.  Wondering if we could use a similar hidden option to create a very small buffer and run something simliar to the original `cat bigfile | flux exec ....` reproducer.

garlick:
> and added a tweaked version of the --stdin-flow option on top.

What was tweaked?  I'm not spotting it.

> Wondering if we could use a similar hidden option to create a very small buffer and run something simliar to the original cat bigfile | flux exec .... reproducer.

Sounds good!   Note there is a hidden `--setopt` function that can be used to set any subprocess option including buffer sizes.

garlick (commented):
Couple quick initial comments

chu11:
> What was tweaked? I'm not spotting it.

Sorry, only minor tweaks (fixed a typo, fix conflicts), nothing major.

> Sounds good! Note there is a hidden --setopt function that can be used to set any subprocess option including buffer sizes.

Oh sweet!  Didn't see that.

src/cmd/flux-exec.c
```diff
@@ -726,8 +803,17 @@ int main (int argc, char *argv[])
                                  NULL,
                                  NULL)))
             log_err_exit ("flux_rexec");
+        if (!(credits = calloc (1, sizeof (int))))
+            log_err_exit ("calloc");
         if (!zlistx_add_end (subprocesses, p))
             log_err_exit ("zlistx_add_end");
+        if (!(handle = zlistx_add_end (subprocess_credits, credits)))
+            log_err_exit ("zlistx_add_end");
```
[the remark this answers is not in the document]
chu11: > Huh I would have thought you would leave the credits in the subprocess aux container like you had before and put the subprocess in the zlistx, not the put credits in the zlistx and the list handle in the subprocess aux container.

Ya know, originally I stored the integer credits in the `zlistx` but disliked the fact I also had to store the "handle" as an aux as well.  It's sort of fallout from the fact that most uses of `zlistx` have a struct that contains the item + handle in one handy location.  But I didn't think of storing the subprocess pointer in the "credits" list.  Let me try it that way.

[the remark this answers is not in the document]
garlick: Just realized this is the first go and "like you had it before" was just how I imagined it :-)

chu11:
re-pushed, tweaked things by putting the subprocess pointer into the "credits" list.  Wrapping all of the "get handle" and "get credits" pointers into functions and hiding all those details makes it not as annoying as I thought it would be.

Also added a test.  Piping a 50 meg file via a 4096 stdin buffer.  If flow control wasn't working, would definitely overflow.  Hopefully this is not too large of a test for the CI.  We'll see.  Could scale it down if it turns out to be.

garlick (commented):
Looking good!

Maybe it would be good to add a variant of the test you added where one remote proc exits immediately but the other procs are checked to ensure they still get all the data?

src/cmd/flux-exec.c
```diff
@@ -367,6 +435,15 @@ void subprocess_destroy (void **arg)
...
 }
 
+int subprocess_credits_compare (const void *item1, const void *item2)
+{
+    flux_subprocess_t *p1 = (flux_subprocess_t *) item1;
+    flux_subprocess_t *p2 = (flux_subprocess_t *) item2;
+    int *credits1 = flux_subprocess_aux_get (p1, "credits");
+    int *credits2 = flux_subprocess_aux_get (p2, "credits");
+    return (*credits1 - *credits2);
+}
```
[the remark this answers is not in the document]
chu11: ahhh, good catch.  I was actually copying from another place in `flux-core`.  Perhaps we should audit and correct that elsewhere.

chu11:
re-pushed, fixing per comments above and adding some flux-exec to > 1 rank tests, and per comment above, one test the subprocess exits early.  It's a good test, caught a corner case!

src/cmd/flux-exec.c
```diff
@@ -218,6 +267,18 @@ void output_cb (flux_subprocess_t *p, const char *stream)
...
 
+void credit_cb (flux_subprocess_t *p, const char *stream, int bytes)
+{
+    subprocess_update_credits (p, bytes, true);
+    if (started == rank_count) {
+        int min_credits = subprocess_min_credits ();
+        if (min_credits)
+            flux_watcher_start (stdin_w);
+        else
+            flux_watcher_stop (stdin_w);
```
garlick, line 286: I don't _think_ this else clause is needed since this only adds credits, so the current `min_credits` is never reduced over what might have come before when the watcher was started?

src/cmd/flux-exec.c
```diff
@@ -226,18 +287,28 @@ static void stdin_cb (flux_reactor_t *r,
     struct fbuf *fb = fbuf_read_watcher_get_buffer (w);
     flux_subprocess_t *p;
     const char *ptr;
-    int lenp;
+    int len, lenp;
+    int min_credits = subprocess_min_credits ();
 
-    if (!(ptr = fbuf_read (fb, -1, &lenp)))
+    if (!(ptr = fbuf_read (fb, min_credits, &lenp)))
```
garlick: I think this is a yes but are we certain `min_credits` is always > 0 here?

src/cmd/flux-exec.c
```diff
@@ -226,18 +287,28 @@ static void stdin_cb (flux_reactor_t *r,
...
+                if ((len = flux_subprocess_write (p, "stdin", ptr, lenp)) < 0)
                     log_err_exit ("flux_subprocess_write");
+                /* N.B. since we are subtracting the same number of credits
+                 * from all subprocesses, the sorted order in the credits list
+                 * should not change
+                 */
+                updated_credits = subprocess_update_credits (p, -1*len, false);
+                /* if one subprocess has no more credits, stop stdin watcher */
+                if (updated_credits == 0)
+                    flux_watcher_stop (stdin_w);
```
garlick: I would just not have `subprocess_update_credits()` return anything and make a call to `subprocess_min_credits()` once after the loop and stop the watcher there if zero, since that's now a cheapish operation.

garlick (approved):
Just a couple of minor comments but I think this looks pretty good!

src/cmd/flux-exec.c
chu11: hmmmmm.  It should be impossible, discounting unknown or future bugs.  Think we should assert check just in case/

garlick: Maybe nothing is needed, but I suppose it could just stop the watcher and return if credits is zero.

chu11: that sounds like a decent idea.

chu11: oh wait!  i think it is possible to be 0, although extremely unlikely, when the stream is closed.  should add some smarts in there for that case.

chu11: ok, i had to think about it a bit.  I think it is possible min_credits could be zero if the remote subprocess has a fatal error (like libsubprocess ENOMEM kinda thing), so we're at 0 credits locally and never get credits back.  But since this is only in an error scenario, I don't think it matters.

Although, this did make me think of a corner case to fix.  If I call `subprocess_remove_credits()`, it's possible that `min_credits` may no longer be zero.  So I should re-check and start the `stdin_w` if necessary.

chu11:
re-pushed per comments above, only minor tweaks

```
diff --git a/src/cmd/flux-exec.c b/src/cmd/flux-exec.c
index 44db06352..914d15abb 100644
--- a/src/cmd/flux-exec.c
+++ b/src/cmd/flux-exec.c
@@ -175,13 +175,12 @@ int subprocess_min_credits (void)
     return spcred->credits;
 }
 
-int subprocess_update_credits (flux_subprocess_t *p, int bytes, bool reorder)
+void subprocess_update_credits (flux_subprocess_t *p, int bytes, bool reorder)
 {
     struct subproc_credit *spcred = flux_subprocess_aux_get (p, "credits");
     spcred->credits += bytes;
     if (reorder)
         zlistx_reorder (subprocess_credits, spcred->handle, false);
-    return spcred->credits;
 }
 
 void subprocess_remove_credits (flux_subprocess_t *p)
@@ -282,8 +281,6 @@ void credit_cb (flux_subprocess_t *p, const char *stream, int bytes)
         int min_credits = subprocess_min_credits ();
         if (min_credits)
             flux_watcher_start (stdin_w);
-        else
-            flux_watcher_stop (stdin_w);
     }
 }
 
@@ -312,21 +309,19 @@ static void stdin_cb (flux_reactor_t *r,
                 if ((len = flux_subprocess_write (p, "stdin", ptr, lenp)) < 0)
                     log_err_exit ("flux_subprocess_write");
                 if (stdin_enable_flow_control) {
-                    int updated_credits;
                     /* N.B. since we are subtracting the same number
                      * of credits from all subprocesses, the sorted
                      * order in the credits list should not change
                      */
-                    updated_credits = subprocess_update_credits (p,
-                                                                 -1*len,
-                                                                 false);
-                    /* if one subprocess has no more credits, stop stdin watcher */
-                    if (updated_credits == 0)
-                        flux_watcher_stop (stdin_w);
+                    subprocess_update_credits (p, -1*len, false);
                 }
             }
             p = zlistx_next (subprocesses);
         }
+        min_credits = subprocess_min_credits ();
+        /* if one subprocess has no more credits, stop stdin watcher */
+        if (min_credits == 0)
+            flux_watcher_stop (stdin_w);
     }
     else {
         p = zlistx_first (subprocesses);
```

garlick:
LGTM, although this comment is not really applicable anymore and could be deleted
```c
/* if one subprocess has no more credits, stop stdin watcher */
```

chu11:
thanks ... and i see I accidentally added a "fdsafsdf" commit while I was fixing the other issue, oops!  I'll re-push and add MWP

chu11:
it appears the system coverage builder has failed on the 60m timeout 3 times in a row.  Unclear to me if this is due to the extra tests, which are "beefier" with all of the 50 meg file copies.  Could be slowing things down a lot there or eating up too much disk.

I'm going to temporarily cut those down to 5 megs, see if that makes a difference (locally running tests 5 megs is still enough to trigger buffer overflow).

grondo:
> it appears the system coverage builder has failed on the 60m timeout 3 times in a row. Unclear to me if this is due to the extra tests, which are "beefier" with all of the 50 meg file copies. Could be slowing things down a lot there or eating up too much disk.


Only tests with `ci=system` are run in the system builder, so I don't think that test is even run there. Could be some other hang?

Also, just noticed the commit message `t: cover flow control` is a bit non-specific. Maybe `t: cover libsubprocess flow control`?

chu11:
> Only tests with ci=system are run in the system builder, so I don't think that test is even run there. Could be some other hang?

Didn't think of that, should try locally in docker.

> Also, just noticed the commit message t: cover flow control is a bit non-specific. Maybe t: cover libsubprocess flow control?

Sounds good

chu11:
hmm, ok atleast got a clue now (I'm not sure why the timeout message is before the last test message, possible output race)

```
  ./t2409-sdexec.t timed out after 300.0s
  ok 11 - sdexec stdout+stderr works
  PASS: t2409-sdexec.t 11 - sdexec stdout+stderr works
  ```

**Comment by chu11** [comment]
doh, took me like 3 seconds to realize the bug

```
diff --git a/src/cmd/flux-exec.c b/src/cmd/flux-exec.c
index 2b51d84ef..173f127bb 100644
--- a/src/cmd/flux-exec.c
+++ b/src/cmd/flux-exec.c
@@ -318,9 +318,11 @@ static void stdin_cb (flux_reactor_t *r,
             }
             p = zlistx_next (subprocesses);
         }
-        min_credits = subprocess_min_credits ();
-        if (min_credits == 0)
-            flux_watcher_stop (stdin_w);
+        if (stdin_enable_flow_control) {
+            min_credits = subprocess_min_credits ();
+            if (min_credits == 0)
+                flux_watcher_stop (stdin_w);
+        }
     }
     else {
         p = zlistx_first (subprocesses);
```

will push a fix.  Also, I'll keep the 5M file size instead of the 50M file size, since it still reproduces reliably with 5M and makes things run faster.

PR #6372, opened 2024-10-15
sched-simple: improve unsupported resource type exception

grondo:
Problem: The exception raised by sched-simple when it receives a job request including GPUs can be confusing. Especially when users expect that a different scheduler is loaded, or do not know that sched-simple does not support scheduling GPUs.

Improve the exception message to say explicitly

 sched-simple does not support resource type 'gpu'

to make it clear that sched-simple is loaded and does not support GPUs.


garlick (approved):
LGTM!

grondo:
Thanks! I'll set MWP.

PR #6373, opened 2024-10-15
doc: update link to flux-accounting guide

cmoussa1:
#### Problem

The hyper-link to the flux-accounting guide on flux-core's front page no longer works because the guide was moved from flux-docs to flux-accounting, which now has its own ReadTheDocs site.

---

This PR just updates the link to the flux-accounting guide.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -81,7 +81,7 @@ currently:
    * - flux-accounting - Bank accounting and fair share priority
      - `Github <https://github.com/flux-framework/flux-accounting>`_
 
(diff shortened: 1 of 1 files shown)


wihobbs (approved):
Is linkcheck enabled here? (I think yes) I guess that just checks that the site is _valid_, not that it actually points to what you want. But if it isn't enabled we should enable that.

wihobbs:
Ooh I wonder how hard it would be to make the old link point to the new link instead of a 404, in case people have it bookmarked.

cmoussa1:
Thanks for approving @wihobbs! Going to set MWP here

PR #6384, opened 2024-10-22
libsubprocess: take reference on callbacks

chu11:
Problem: There is always a chance a user may destroy a subprocess object when a callback is called (i.e. output callback, state change callback, etc.).  This can lead to problems if libsubprocess believes the object is still valid when the callback returns.  References are taken on the subprocess object when issuing some callbacks, but it is missing in other callbacks.

Take reference on all callbacks to ensure safety if the user choses the destroy the subprocess in the callback.


grondo (approved):
LGTM!

Issue #4569, opened 2022-09-14
Stream job events from flux


morrone:
I don't know if existing functionality can reasonably satisfy this or not, so I will describe the use case that we have in mind:

When flux is used as the cluster-level scheduler ("native flux"?), we would like a to generate messages in real-time for job scheduling related events. Those events would include, but not be limited to:

- job submission
- job changes while queued
- job start
- job end
- events related to job prolog/epilogs

We would like the job information in the events to include all relevant information about the job, for instance:

- user name
- requested resources (maybe include the jobspec?)
- allocated resources
- event timestamp
- queue name (or flux equivalient)
- bank name

It is open for discussion whether all data is included in each event, or perhaps just the data that has changed since the previous event for the same job.

The goal of the job event interface in flux would be to allow us to publish the data to a topic in Kafka. In turn there would be one or more consumers of the job events topic. One of the first consumers would be the elasticsearch database that will back the operational monitoring dashboard implemented in grafana.

Can we cleanly achieve this with current APIs in flux? Will we need more work in flux-core (or somewhere else in flux?) to meet this need?

garlick:
I think if we sent you jobspec, R, and the primary job eventlog for each job, you would have everything on your list (or if things are missing we can fill them in - bank comes to mind).

If you're curious you can examine these on a flux system like corona with
```
$ flux mini submit ...
<prints JOBID>
$ flux job eventlog --format=json JOBID
$ flux job info JOBID jobspec
$ flux job info JOBID R
```

Although you could write something now using flux APIs to grab this stuff in real time now, when we heard your talk in S<sup>3</sup>G meeting, we were wondering about a new service in flux that could efficiently provide this as well as serve some other monitoring requirements we have.

Other questions we might want to discuss
- any data on your list that is not provided by those three things (bank comes to mind)
- do you require the raw data to be reduced into something simpler, or could we provide say RFCs describing the raw data and tools for manipulating it instead?
- edge cases: catching up after disconnecting, is duplicate data OK, etc.
- does the data need to be anonymized or otherwise sanitized?
- hmm, what else can we do with the entire center's flux job data in a nosql database???

morrone:
Those are good questions.

I don't think the data necessarily needs to be simpler as long as it is present. The plugin's job would be to convert the data into whatever destination format needed (my immediate use case being something like an Avro schema for use in Kafka). But if the format is especially difficult to parse and use maybe I would change my mind about that. :)

[...]

Issue #6352, opened 2024-10-07
means to subscribe to all events?


vsoch:
Hi! I'm looking for a means to subscribe to all events under a flux instance - as a basic example, instead of for just one jobid, I"d want to be able to watch all jobids seen by the handle. I'm also interested to do this in Python. I know how to watch for a specific jobid, and I know I can do a job listing rpc with get(), but I'm wondering if there is more of a pub sub like model where I don't need to poll at some interval. The use case is for a tool that can keep metrics on the queue and respond to events and decide to act on them. Any tips or examples would be appreciated. Thank you!

vsoch:
Ooh, what about this? https://github.com/garlick/flux-core/blob/19cf1cffb354739c5f691ef8dbfae2ae204d6999/t/job-manager/bulk-state.py#L77-L79

vsoch:
No, that didn't seem to work - I ran it, and then submit a job under that same handle, and nothing. Will keep looking.

vsoch:
oh this works! This must be all events? I got a heartbeat back!

```python
In [17]: self.handle.event_subscribe('')
Out[17]: 0

In [19]: event = self.handle.event_recv()

In [20]: event
Out[20]: <flux.message.Message at 0x768fb6443a90>

In [21]: event.decode()
Out[21]: (4, 'heartbeat.pulse', None)
```

vsoch:
HE'S ALIVE! :zombie:

vsoch:
This might work to subscribe to all, actually: 
![image](https://github.com/user-attachments/assets/92587764-94f2-4891-b8b4-be7fbc9ac3ee)

grondo:
There's multiple types of events. Are you interested in watching all event messages, or all eventlog events for jobs?

vsoch:
I'm interested for the most part in states - that is when we'd want to do something. I'm working on extending my ensemble operator to do ML stuff, and instead of the current design that does a poll, I want an event oriented architecture. So ideally when a job is submit, I get notice of that. And of course when it goes to `SCHED`, `RUN`, etc. - I just want all the states! And any other job events that might be interesting. The kvs above seems to tell me that a job was submit by way of creating the key value store for it, but I'd rather have more direct events.

garlick:
Perfect, just subcribe to `"jobs."` and you'll get all state transition notifications for all jobs.

vsoch:
[...]

PR #6386, opened 2024-10-22
python: add `flux.job.JournalConsumer` class with a simplified interface for job manager journal consumers

grondo:
This PR adds the `JournalConsumer` class described in #4569 with the addition of some tests.

Putting up here now for review so we can get some official feedback on the interface.


garlick:
Took me a little while to find the auto-generated docs in our python docs.   Had to use the search box but here it is:

https://flux-framework--6386.org.readthedocs.build/projects/flux-core/en/6386/python/autogenerated/flux.job.journal.html#module-flux.job.journal

I did notice "Syncrhonously" was misspelled in the `poll()` docs.  Hmm, I thought our typo checker would have caught that.

Feel free to ignore, but for extant use cases in the source tree,  time sorting of historical events wasn't required and the sentinel demarcating history from current events was useful.  I wonder if making history sorting optional or maybe adding an option to to signal the demarcation with an empty `JournalEvent` might be useful down the line?  Meh, would be easy to add later I'm sure.

grondo:
I was actually going to ask if we thought the sentinel would be useful to get, but I wasn't quite sure how to present it.

I guess one thought is that if a use case doesn't require the events ordered, then probably this class is not necessary and they can use the RPC directly?

grondo:
(and FYI I found some issues with handling of ENODATA so I'm reworking some of the code now.)

garlick:
Great point!

grondo:
I've repushed here with support for the sentinel event if `include_sentinel` is `True` in the `JournalConsumer` initializer. (The default is `False`). The sentinel event is represented by an empty event defined by  a `flux.job.journal.SENTINEL_EVENT` constant. Tests have been added for `include_sentinel=True`.

I also expanded the docs, trying to use sphinx references where possible. Note that I've dropped the `pre-commit` [check-docstring-first](https://github.com/pre-commit/pre-commit-hooks/blob/main/README.md#check-docstring-first) check, which I don't think we really need in Flux and having it effectively makes it impossible to document module variables and constants. (it threw an error with the `SENTINEL_EVENT` documentation here)

grondo:
It would probably be good if @wihobbs gave this iteration of `JournalConsumer`  a try, and also if @morrone gave an ACK. (generated docs from this PR are linked above by @garlick)

garlick:
The docs look much nicer!

morrone (approved):
Looks good to me!

I like that the flux handle is passed to JournalConsumer() now instead of start().

Maybe one minor nit in the docs, that doesn't need to hold up the landing of this PR:

It says:

```
  context
    context dictionary (see RFC 18)
```

But all that RFC18 says is:

```
  context
    (object) Application-specific event data."
```

Maybe it would be better to point directly to RFC 21? Maybe event to the "Event Descriptions" section?

https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_21.html#event-descriptions

grondo:
> Maybe it would be better to point directly to RFC 21? Maybe even to the "Event Descriptions" section?

Good idea!

grondo:
I pushed a new version of this PR which adds references to RFCs where appropriate as suggested above. (I've added an `rfc` domainref to `doc/conf.py` to make it easier to reference RFCs in the docs here)

Also expanded the description of the `JournalConsumer` class a tiny bit.

wihobbs (approved):
Hi @grondo, this is a slick addition. Thanks for working on it!

I have a [script that's boomeranged around multiple team members that subscribes to the journal and prints events as they come in](https://github.com/flux-framework/flux-core/issues/6352#issuecomment-2398064808). For a look at how to do this with the new interface, here's a synchronous example:

```
import flux
import flux.job

def main():
    handle = flux.Flux()

    journal = flux.job.journal.JournalConsumer(handle)

    journal.start()

    while True:
        print(journal.poll())

main()
```

And an asynchronous example:

```
import flux
import flux.job

def print_event(eve):
    print(eve)

def main():
    handle = flux.Flux()

    journal = flux.job.journal.JournalConsumer(handle)

    journal.start()

    journal.set_callback(print_event)

    handle.reactor_run()

main()
```

Thanks again for doing this.

grondo:
Thanks for the approval @wihobbs! I'll set MWP then :-)

src/bindings/python/flux/job/journal.py
```diff
@@ -0,0 +1,259 @@
...
+        return self.backlog.popleft()
+
+    def start(self):
+        """Start the stream of events by sending a request to the job manager
+
+        This function sends the job-manager.events-journal RPC to the
+        job manager. It must be called to start the stream of events.
+        """
+        self.rpc = self.handle.rpc(
+            "job-manager.events-journal", {"full": self.full}, 0, FLUX_RPC_STREAMING
```
wihobbs, line 157: Oh, one thing I just thought of -- take it or leave it: the journal RPC gives you the ability to subscribe to [only certain events](https://github.com/flux-framework/flux-core/blob/master/src/modules/job-manager/journal.c#L16), such as start and finish, maybe we want to expose that option to `JournalConsumer` as well?

I think most services are going to be interested in the full event log, though.

wihobbs: Ah, ok, for another time :)

vsoch: I would be interested in that ability - I'm having to filter them out for my tool. I'm also debugging a case where it looks like an event was missing / dropped, but it's hard to reproduce.

PR #6390, opened 2024-10-25
python: usability improvements for `JournalConsumer` class

grondo:
@wihobbs had pointed out some usability issues with the `JournalConsumer` class yesterday after the PR was merged. This PR tackles those issues.

 - Document that `poll()` must be called after `start()`.
 - handle the above case better by raising a specific `RuntimeException` when the `JournalConsumer` isn't started before `poll()` is called
 - allow `set_callback()` to be called before `start()` - there's no reason not to allow it.
 - add tests accordingly.

### DIFF
--- a/src/bindings/python/flux/job/journal.py
+++ b/src/bindings/python/flux/job/journal.py
@@ -180,11 +180,17 @@ class JournalConsumer:
         historical events have been processed. Historical events will sorted
         in time order and returned once per :func:`poll` call.
 
+        :func:`start` must be called before this function.
+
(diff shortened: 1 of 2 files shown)


wihobbs (approved):
Works as expected, and I appreciate the `RuntimeException` message. Thanks, @grondo!

grondo:
Thanks @wihobbs! I actually just pushed one more doc update here: A note/warning about calling `start()` multiple times.
I assume this is ok and will set MWP.

PR #6402, opened 2024-10-31
sdexec: set KillMode=process SendSIGKILL=no for multi-user jobs

### LINKED ISSUE #6399: sdexec: systemd tries to kill guest tasks without sufficient permission
Problem:  the systemd user instance running as `flux` is not privileged to kill guest processes such as the job shell or tasks, yet it tries if the job does not finalize on its own.

The various options are documented in [systemd.kill(5)](https://www.freedesktop.org/software/systemd/man/latest/systemd.kill.html).

`sdexec` uses the default `KillMode=control-group` where
> all remaining processes in the control group of this unit will be killed on unit stop

A better choice might be `KillMode=process` where
> only the main process itself is killed (not recommended!)
> Note that it is not recommended to set KillMode= to process or even none, as this allows processes to escape the service manager's lifecycle and resource management, and to remain running even while their service is considered stopped and is assumed to not consume any resources.

We should probably also set `SendSIGKILL=no`.  If the IMP is waiting for processes, SIGKILLing the IMP would lose track of those processes.

garlick:
Problem: For multi-user jobs spawned via SDEXEC, the systemd user instance running as the flux user does not have permission to kill guest processes, yet it does try and in the process may kill off the only process that does have permission to continue cleanup efforts, the IMP.
    
When the job is run by the IMP and sdexec, Set `KillMode=process` so that systemd only delivers signals to the IMP, which it should forward to the shell and/or cgroup per RFC 15.
    
Also set `SendSIGKILL=off` so that SIGKILL is never deployed against the IMP.
    
Fixes #6399

This _should_ be OK even without the forthcoming changes to job-exec and the IMP to forward signals and avoid using SIGKILL, however I still need to give this a sanity check on my test cluster as the conditional use of these properties for IMP jobs is not tested by CI.  I think we can still get a review going however so I'll not set WIP.

### DIFF
--- a/src/common/libsdexec/start.c
+++ b/src/common/libsdexec/start.c
@@ -155,6 +155,26 @@ static int prop_add_bool (json_t *prop, const char *name, int val)
     return 0;
 }
 
+// per systemd.syntax(7), boolean values are: 1|yes|true|on, 0|no|false|off
+static bool is_true (const char *s)
+{
+    if (streq (s, "1")
+        || !strcasecmp (s, "yes")
+        || !strcasecmp (s, "true")
+        || !strcasecmp (s, "on"))
+        return true;
+    return false;
+}
+static bool is_false (const char *s)
+{
+    if (streq (s, "0")
+        || !strcasecmp (s, "no")
+        || !strcasecmp (s, "false")
+        || !strcasecmp (s, "off"))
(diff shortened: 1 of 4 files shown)


grondo (approved):
LGTM!

PR #6403, opened 2024-10-31
frobnicator: allow queue-specific defaults to override global defaults

### LINKED ISSUE #6400: In the `queues` config table, (some?) policy settings do not appear to be overriding system defaults
Example configuration:
```
duration = "30m"
queue = "pall"

[queues.pall]
requires = ["pall"]
policy.limits.duration = "24h"
policy.jobspec.defaults.system.duration = "15m"
```
My impression is that with this configuration, jobs submitted for the `pall` queue should end up with `attributes.system.duration: 900.0` in the jobspec.  Actual result is `attributes.system.duration: 1800.0` with the above config.

grondo:
As noted in #6400, queue defaults currently do not override global defaults because the global default is applied before queue defaults are checked.

This PR fixes that.

Fixes #6400

### DIFF
--- a/src/bindings/python/flux/job/frobnicator/plugins/defaults.py
+++ b/src/bindings/python/flux/job/frobnicator/plugins/defaults.py
@@ -12,6 +12,8 @@
 
 """
 
+import copy
+
 from flux.job.frobnicator import FrobnicatorPlugin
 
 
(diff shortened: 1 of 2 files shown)


garlick (approved):
LGTM!

grondo:
Thanks! Setting MWP.

Issue #194, opened 2015-05-20
building libtap


lipari:
The resource model alternative that was added to flux-sched includes a test that relies on libtap.  The test's makefile references libtap in flux-core.  The problem is that libtap in flux-core is only built when "make check" is invoked in flux-core.  And flux-sched's travis.yml runs a make in flux-core but not a "make check".  When the "make check" for flux-sched is run, it fails because libtap is missing.

So, what should we do?  Modify flux-core so libtap is always built?  Add a "make check" of flux-core to flux-sched's travis.yml?  Or copy the libtap source to flux-sched and build it locally every time "make check" is run?

grondo:
For now it should be sufficient to add something like:
```  
- (cd ../flux-core/src/common/libtap && make check)
```

to `.travis.yml`, though re-running `make check` in flux-core wouldn't be bad necessarily, it seems unnecessary.

garlick:
Closing as it sounds like this is best solved by updating flux-sched's `.travis.yml`

lipari:
Yes, this issue can be closed now that 172b7d3057984b856b47040a6c74c9f6375a77a4 has been committed.

PR #6408, opened 2024-11-01
update multi-user signaling to track flux-security 0.13.0 IMP changes

garlick:
Problem: job-exec uses `flux imp kill` to deliver SIGKILL to the flux-shell when shell signaling methods fail to clean up a multi-user job, but the `flux imp kill` sub-command is being deprecated in favor of having the IMP forward signals (per RFC 15).

This changes job-exec to send SIGUSR1 (which RFC 15 defines as a proxy for SIGKILL) directly to the IMP in that case.

To make it easier to coordinate the flux-core and flux-security changes, we'll add the #6409 fix here as well.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -399,7 +399,7 @@ AS_IF([test x$enable_code_coverage = xyes], [
 AC_ARG_WITH([flux-security], AS_HELP_STRING([--with-flux-security],
              [Build with flux-security]))
 AS_IF([test "x$with_flux_security" = "xyes"], [
-    PKG_CHECK_MODULES([FLUX_SECURITY], [flux-security >= 0.9.0],
+    PKG_CHECK_MODULES([FLUX_SECURITY], [flux-security >= 0.13.0],
                       [flux_sec_incdir=`$PKG_CONFIG --variable=includedir flux-security`],
                       [flux_sec_incdir=;])
     AS_IF([test "x$flux_sec_incdir" = x],
--- a/doc/man1/flux-exec.rst
+++ b/doc/man1/flux-exec.rst
@@ -95,11 +95,7 @@ OPTIONS
 
    Prepend the full path to :program:`flux-imp run` to *COMMAND*. This option
    is mostly meant for testing or as a convenience to execute a configured
-   ``prolog`` or ``epilog`` command under the IMP. Note: When this option is
-   used, or if :program:`flux-imp` is detected as the first argument of
-   *COMMAND*, :program:`flux exec` will use :program:`flux-imp kill` to
-   signal remote commands instead of the normal builtin subprocess signaling
-   mechanism.
+   ``prolog`` or ``epilog`` command under the IMP.
 
 CAVEATS
 =======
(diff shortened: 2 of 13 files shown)


grondo:
This can't be merged until a flux-security v0.12 is tagged which supports SIGUSR1 in the IMP. (SIGUSR1 will cause the current IMP to terminate immediately)

garlick:
Right, sorry, probably should've been a WIP.

grondo (approved):
LGTM!

I do notice a few places where `flux-imp kill` is still referenced:

- `flux-exec.c` still has an `imp_kill()` function
- `flux-exec(1)` mentions that it uses `flux-imp kill` to signal processes
- bulk-exec still has `imp_kill_ops` and associated functions
- `t/job-exec/imp-fail.sh` has unnecessary `imp kill` implementation

garlick:
Wow, don't know how I missed all that.  OK, updated, and since the flux-security 0.12 tag was just updated, this actually has a chance of passing CI so we'll see.

garlick:
Some `t9000-system.t` tests are failing where we check that `flux exec --with-imp` forwards signals.  Here's an excerpt from one failing test that uses the IMP to run a shell script that calls sleep:
```
sending signal 2 to 1 running processes
sudo timed out after 30.0s
test_expect_code: command exited with 137, we wanted 130 run_timeout 30 sudo -u flux ./test_signal.sh INT
not ok 8 - 0002-exec-with-imp.t: flux exec --with-imp forwards signals
```
I can recreate this environment this on my test system by creating a shell script that runs sleep, configuring it in the IMP, and manually running `flux exec --with-imp`, and sure enough, SIGINT to `flux-exec` appears to have no effect, and sending a SIGTERM kills the shell and the IMP but leaves the sleep and doesn't terminate.

I _did_ observe:
- manually sending SIGINT to the shell (script) also has no effect
- manually sending SIGTERM to the shell (script) causes it to terminate but orphans the `sleep` command, which in turn prevented the subprocess server from finalizing the exec stream with ENODATA because the stdout/stderr streams were still open.

When I run the same script directly with `flux exec` (no IMP), those signals cause everything to wrap up as expected.

Still pondering this one - just wanted to post an update!

garlick:
Possibly the IMP's internal `fwd_signal()` needs to behave differently when the target is not a shell.  Currently it treats SIGUSR1 (surrogate for SIGKILL) specially and assumes it (the IMP) is responsible for cleaning up its child and grandchildren.  Other signals are delegated to the direct child for distribution, which sounds right for `flux-shell` but perhaps not for a shell script.  Maybe it needs a flag so when it is called from `imp run` it assumes it is cleaning up everything for any signal.

Another observation is that `flux_subprocess_kill()` by default uses `killpg()` while the IMP uses `kill()` when signaling  the direct child.

garlick:
After chatting with @grondo, I opened flux-framework/flux-core#194 to change the signal forwarding behavior of `imp run`.

garlick:
Updated to require flux-security 0.13, which should address the test failure.  :crossed_fingers:

garlick:
I'll set MWP on this one too.

PR #6410, opened 2024-11-01
WIP job-manager: use subprocess kill not flux imp kill

### LINKED ISSUE #6409: job-manager: don't use flux imp kill to signal housekeeping, perilog
Problem: `flux imp kill` will be deprecated but the job manager uses it to terminate housekeeping and prolog/epilog tasks.

Just use regular `subprocess_kill()` and the IMP should forward signals.

First we need flux-framework/flux-security#188 though.

garlick:
Problem: housekeeping and perilog use 'flux imp kill' to implement send signals to housekeeping and prolog/epilog processes, but the 'flux imp kill' subcommand is being deprecated in favor of having the IMP forwarding signals.

Don't configure bulk-exec with the imp path so that bulk_exec_kill() will use flux_subprocess_kill().

Fixes #6409

(don't merge before we have a flux-security release with flux-framework/flux-security#188 in it)

### DIFF
--- a/src/modules/job-manager/housekeeping.c
+++ b/src/modules/job-manager/housekeeping.c
@@ -190,8 +190,6 @@ static struct allocation *allocation_create (struct housekeeping *hk,
                                               id,
                                               "housekeeping",
                                                a))
-        || (hk->imp_path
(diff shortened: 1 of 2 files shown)


garlick:
Will combine this with #6408 - withdrawing.

PR #6412, opened 2024-11-03
resolve perilog plugin issue that lets a job start after prolog timeout when cancellation fails

grondo:
This PR fixes an issue discovered while testing recent flux-security changes.

If the prolog times out and the resulting cancel (SIGTERM) then times out, affected ranks are drained, but no job exception is raised so the job continues to start. Not only does this result in a job starting when it isn't supposed to, but additionally the epilog may be run _twice_: once when the prolog fails and again when the `finish` event is posted.

This PR fixes the issue by adding a new flag cancel_timeout to the `perilog_proc` struct which then results in a specific exception being raised. The exception then prevents the job from starting and of course the extra `finish` event, etc.

A test that reproduced the issue is also added.

### DIFF
--- a/src/modules/job-manager/plugins/perilog.c
+++ b/src/modules/job-manager/plugins/perilog.c
@@ -111,6 +111,7 @@ struct perilog_proc {
     bool cancel_on_exception;
     bool canceled;
     bool timedout;
+    bool cancel_timeout;
     double kill_timeout;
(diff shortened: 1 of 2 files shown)


garlick (approved):
Good catch!  LGTM.

grondo:
Thanks! I've set mwp

PR #6416, opened 2024-11-05
libsubprocess: close extra file descriptors

garlick:
Problem: libsubprocess duplicates the child ends of channel socketpairs onto stdin, stdout, stderr in the child, but leave the original copies open.
    
Exclude those channels from the idset generated by `subprocess_childfds()`.  Both `fork.c` and `posix_spawn.c` already include logic to close descriptors that do not appear in this idset after stdio is duplicated.
    
Add a test.


chu11 (approved):
doh! good catch

garlick:
Thanks! I'll set MWP

PR #6422, opened 2024-11-08
job-manager/history: optimize list insertion

### LINKED ISSUE #6419: throughput of jobs slows down over time - part deux
While working on #6414 , I noticed that job throughput degrades over time.  (similar to the old #3583)  

Running throughput test of 8192 jobs over and over again in an instance via this old script

```
count=$1
n=$2
echo "Running throughput.py ${count} jobs per iteration, ${n} iterations"
for i in `seq 1 ${n}`
do
    src/test/throughput.py -n ${count} | grep throughput
done
```

```
>./test_throughput.sh 8192 15
throughput:     487.3 job/s (script: 475.7 job/s)
throughput:     405.4 job/s (script: 397.8 job/s)
throughput:     288.0 job/s (script: 284.2 job/s)
throughput:     307.4 job/s (script: 303.1 job/s)
throughput:     278.4 job/s (script: 274.9 job/s)
throughput:     276.4 job/s (script: 272.9 job/s)
throughput:     220.1 job/s (script: 217.9 job/s)
throughput:     210.3 job/s (script: 208.3 job/s)
throughput:     128.8 job/s (script: 128.0 job/s)
throughput:     83.6 job/s (script:  83.3 job/s)
throughput:     67.6 job/s (script:  67.4 job/s)
throughput:     57.1 job/s (script:  56.9 job/s)
```

(I broke out of the test after 12 iterations, not wanting to bother waiting anymore)

Some degradation is to be expected, as memory gets eaten up and what not, but this seems a tad larger decrease than I'd expect.  In total we're still < 100K total jobs in the results above, so it shouldn't be hogging too much memory.  Perhaps a round of perf analysis would be good to do, perhaps we got some bottleneck somewhere.

chu11:
Problem: Internal job history lists are sorted with larger t_submit values at the front of the list.  When new jobs are inserted (i.e. jobs with bigger t_submit values), they are inserted with a search starting at the back of the list. This is the opposite of what should be done and leads to a slowdown in job throughput over time.

Insert into the job history list starting at the front.

Fixes #6419

### DIFF
--- a/src/modules/job-manager/plugins/history.c
+++ b/src/modules/job-manager/plugins/history.c
@@ -165,13 +165,13 @@ static int jobtap_cb (flux_plugin_t *p,
             job_entry_destroy (entry);
             return 0;
         }
-        if (!hola_list_insert (hist->users, key, entry, false)) {
+        if (!hola_list_insert (hist->users, key, entry, true)) {
             job_entry_destroy (entry);
             return -1;
         }
     }
     else if (streq (topic, "job.new")) {
-        if (!hola_list_insert (hist->users, key, entry, false)) {
+        if (!hola_list_insert (hist->users, key, entry, true)) {
             job_entry_destroy (entry);
             return -1;
         }


grondo (approved):
Nice! Simple change, big improvement :+1:

PR #6423, opened 2024-11-08
convert ISO timestamp output from UTC to local time + offset in `flux dmesg` and eventlog commands

### LINKED ISSUE #6421: consider using local time for eventlog formatter ISO 8601 format
Currently the eventlog formatter used by `flux job eventlog`, `flux kvs eventlog` and `flux job wait-event` formats the `iso` time format using UTC time always (using the standard suffix `Z`). This is slightly inconvenient, though, because other system tools like `dmesg` and log files such as `/var/log/messages` present ISO time in the current time zone + offset, e.g.:

```
2024-11-07T15:57:47.964255-08:00 fluke65 systemd[1]: Started Housekeeping for Flux job fEi29GDEQPZ.
```

This makes it unnecessarily difficult to compare timestamps between eventlogs and system events.

It might be simplest to just have the ISO output use the current timezone -- appending the offset still conforms to ISO 8601 AFAIK, and you could get UTC easily by changing the setting of the `TZ` variable...

The same argument could perhaps be made for `flux dmesg` as well..

grondo:
This PR changes the default timestamp display for `flux dmesg` and the "iso" time format for `flux job eventlog` `flux kvs eventlog` and `flux job wait-event`  to use local time instead of UTC. This better matches the behavior of most other system utilities and logfiles, and the old behavior can easily be restored if desired by setting `TZ=UTC`.

Fixes #6421

### DIFF
--- a/t/t0009-dmesg.t
+++ b/t/t0009-dmesg.t
@@ -117,6 +117,21 @@ test_expect_success 'logged non-ascii printable characters are unmodified' '
 test_expect_success 'dmesg request with empty payload fails with EPROTO(71)' '
 	${RPC} log.dmesg 71 </dev/null
 '
+# Note: TZ=UTC is set by default for sharness
+test_expect_success 'dmesg with TZ=UTC uses "Z" as offset' '
+	flux dmesg \
+	    | grep -E \
+	    "^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}\.[0-9]+Z"
+'
+
+test "$(TZ=America/Los_Angeles date +%z)" != "+0000" &&
+  test "$(TZ=America/Los_Angeles date +%Z)" != "UTC" &&
+  test_set_prereq HAVE_TZ
+test_expect_success HAVE_TZ 'dmesg with TZ=America/Los_Angeles uses tz offset' '
+	TZ=America/Los_Angeles flux dmesg >dmesgLA.out &&
+	test_debug "cat dmesgLA.out" &&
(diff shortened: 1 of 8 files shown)


t/t0009-dmesg.t
```diff
@@ -117,6 +117,17 @@ test_expect_success 'logged non-ascii printable characters are unmodified' '
...
+# Note: TZ=UTC is set by default for sharness
+test_expect_success 'dmesg with TZ=UTC uses "Z" as offset' '
+	flux dmesg \
+	    | grep -E \
+	    "^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}\.[0-9]+Z"
+'
+test_expect_success 'dmesg with TZ=America/Los_Angeles uses tz offset' '
+	TZ=America/Los_Angeles flux dmesg \
+	    | grep -E \
+	    "^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}\.[0-9]+-[0-9]{2}:[0-9]{2}"
```
chu11, line 129: my memory isn't good on timezone stuff, but if we're fixing American/Los Angeles in the test, wouldn't the offset in the last chunk be known?  `-08:00`?

grondo: I think it could be different depending on current daylight savings? Anyway, this was probably good enough and I didn't want a randomly failing test.

The test is failing anyway in some builders - I wonder if they don't support this timezeone

grondo:
@chu11 noticed I dropped the actual conversion of the dmesg timestamp to local time here. Setting WIP until I fix that.

chu11 (approved):
LGTM

grondo:
Removed WIP. Had to add a workaround in the tests for detecting missing tzdata on `apline` and `jammy`

grondo:
setting MWP. Thanks!

Issue #6396, opened 2024-10-28
TOML config updates can overwrite tables


grondo:
One feature of Flux TOML configuration is that a set of `*.toml` files can be read in and the configuration is aggregated into a single table (or JSON object internally).

Given this, it may be enticing to attempt to update a single config key in a table by dropping in a TOML file that will be read last in lexical order. However, this doesn't work because `json_object_update()` is called so only top-level keys are updated. Thus, the result of this practice will be to _replace_ the config table that was meant to be updated. For example, if we have the following job-manager config in the main `config.toml`:

```toml
[job-manager]
inactive-age-limit = "30d"
plugins = [ { load = "perilog.so" } ]

[job-manager.prolog]
timeout = "1m"
per-rank = true

[job-manager.epilog]
timeout = "15s"
per-rank = true

[job-manager.housekeeping]
release-after = "30s"
```

and we want to update the prolog timeout for instance with a drop in config:

```toml
[job-manager.prolog]
timeout = "30m"
```

What actually happens is the entire `job-manager` table is replaced and you end up with:
```
$ flux config get --config-path=system | jq '."job-manager"'
{
  "prolog": {
    "timeout": "30m"
  }
}
```

PR #6424, opened 2024-11-11
recursively update instead of replacing tables when loading Flux configuration

grondo:
This PR updates the `json_object_update()` call to use `json_object_update_recursive()` when "updating" a Flux config object with another config fragment loaded from a file. This makes it so that table values in the new config object update tables in the existing config, instead of replacing tables (see #6396).

Since `json_object_update_recursive()` is not present until jansson v2.13, add a simple implementation to libmissing if necessary. Note that the upstream jansson version of this function detects cycles, whereas the libmissing version does not, but that is unlikely to be an issue for Flux.


garlick (approved):
Nicely done!

grondo:
Thanks! I'll set MWP.

PR #6435, opened 2024-11-13
python: call shutdown() on executor in job validator

### LINKED ISSUE #6434: job validator hang at exit
On a production system we've seen cases of the job validator hanging at exit. It appears that at the time there are multiple threads active, and the exiting main thread is blocked on a lock, possibly waiting for these threads.

Perhaps there is something unsafe that the validator (and likewise frobnicator) is doing with `concurrent.futures`.

grondo:
Problem: The concurrent.futures ThreadPoolExecutor is not explicitly shut down when exiting flux-job-validator. This MAY be causing the program to occasionally hang at exit.

Add a stop() method to the JobValidator class which calls the shutdown() method of the executor. Call validator.stop() from flux-job-validator(1) before exit.

Fixes #6434

### DIFF
--- a/src/bindings/python/flux/job/validator/validator.py
+++ b/src/bindings/python/flux/job/validator/validator.py
@@ -191,6 +191,10 @@ class JobValidator:
         )
         return self
 
+    def stop(self):
+        """Stop the validator."""
+        self.executor.shutdown()
+
     def validate(self, jobinfo):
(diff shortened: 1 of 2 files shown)


garlick (approved):
Sounds good to me.

grondo:
Thanks I've set MWP

PR #6436, opened 2024-11-14
libflux: add flux_watcher_is_active()

garlick:
This reworks @grondo's withdrawn PR to use the method discussed in #4156.

It removes the following functions from the public API:
- `flux_watcher_create()`
- `flux_watcher_get_data()`
- `flux_watcher_get_ops()`

Note that #6316 broke this ABI without us noticing, two releases ago.  It seemed better to just remove these functions for now than try to make the ABI more robust to change when we don't think anybody actually needs it.

There is some gratuitous cleanup of libflux includes, because I repeatedly found myself dealing with circular dependencies.  That could be split off to another PR if desired.


src/common/libflux/reactor.h
```diff
@@ -177,37 +177,6 @@ void flux_stat_watcher_get_rstat (flux_watcher_t *w,
                                   struct stat *stat,
                                   struct stat *prev);
 
-/* Custom watcher construction functions:
```
grondo, line 168: commit message typo: reslient.

grondo (approved):
This turned out to be very nice cleanup! Nice!

PR #6438, opened 2024-11-14
job-ingest: improve cleanup and stats output

garlick:
This adds a count of "trash" processes for each work crew to the `flux module stats job-ingest` output, so that there is some indication that the module is tracking subprocesses whose input has been closed, but that have not yet exited.

While in there I noticed that the cleanup code that sends SIGKILL to running workers doesn't signal workers that are lingering in the trash list.  Add code to signal those too and a test.


grondo (approved):
LGTM!

PR #6440, opened 2024-11-14
update clang-format

garlick:
Salvage the improvements to our `.clang-format` file from #4694 which was out of date.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -14,7 +14,7 @@ BinPackParameters : false
 BinPackArguments : false
 AllowAllParametersOfDeclarationOnNextLine : false
 AlignTrailingComments : true
-ColumnLimit : 88
+ColumnLimit : 80
 
 # do not put all arguments on one line unless it's the same line as the call
 PenaltyBreakBeforeFirstCallParameter : 10000000
--- a/.clang-format
+++ b/.clang-format
@@ -1,13 +1,17 @@
 BasedOnStyle : google
 SpaceBeforeParens : Always
 IndentWidth : 4
-BreakBeforeBraces : Linux
+BreakBeforeBraces : Custom
+BraceWrapping :
+  BeforeElse : true
+  AfterFunction : true
 UseTab: Never
 AllowShortIfStatementsOnASingleLine : false
 ConstructorInitializerAllOnOneLineOrOnePerLine : true
 AllowShortFunctionsOnASingleLine : false
 AllowShortLoopsOnASingleLine : false
 BinPackParameters : false
+BinPackArguments : false
 AllowAllParametersOfDeclarationOnNextLine : false
 AlignTrailingComments : true
 ColumnLimit : 88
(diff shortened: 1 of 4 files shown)


trws:
I see you decided to re-restrict the line length to 80.  While we do generally try to hit 80 we don't actually follow it everywhere but clang-format will enforce it. Running a little script to get line lengths of every C source file in src/, filtering out libev and some of the other external things we vendor.  There are 466 lines that are between 80 and 88 lines long, and 540 more that are above 88 characters long.  A lot of the longer ones (especially over 100 chars) are in the tests, so that mostly doesn't count, but here are the files impacted by at least one line between 80 and 88 characters:

```
src/cmd/flux-logger.c
src/common/libflux/conf.c
src/common/libflux/test/rpc.c
src/common/libflux/test/rpc_security.c
src/common/libhostlist/test/hostlist.c
src/common/libidset/test/idset.c
src/common/libjob/test/job.c
src/common/libjob/test/sign_none.c
src/common/libjob/test/unwrap.c
src/common/libkvs/test/treeobj.c
src/common/liboptparse/test/optparse.c
src/common/libpmi/test/canonical2.c
src/common/librlist/test/rhwloc.c
src/common/librouter/test/router.c
src/common/libsubprocess/test/channel.c
src/common/libsubprocess/test/fbuf_watcher.c
src/common/libsubprocess/test/stdio.c
src/common/libtaskmap/test/taskmap.c
src/common/libutil/basemoji.c
src/common/libutil/sha1.c
src/common/libutil/sha256.c
src/common/libutil/sha256.h
src/common/libutil/test/cronodate.c
src/common/libutil/test/sha256.c
src/common/libutil/test/stdlog.c
src/common/libutil/timestamp.c
src/modules/content-s3/content-s3.c
src/modules/content-s3/s3.c
src/modules/job-list/state_match.c
src/modules/job-list/test/match.c
src/modules/job-manager/test/job.c
src/modules/kvs/test/kvstxn.c
src/modules/kvs/test/lookup.c
src/shell/test/jobspec.c
```

FWIW, the linux kernel also requests a line length of 80 characters, but changed checkpatch and other tooling to only force format or error at 100 characters.  That combined with our black line limit being set to 88 was why I had it as 88 to begin with.

To be honest I don't really mind, but this is where that came from.

garlick:
From RFC 7:
> Lines SHOULD be limited to 80 characters.

Which implies a best effort, but `clang-format` will actively remove line breaks to expand into 88 characters which undoes our best efforts.

We could argue that 88 should be the number in RFC 7 but that would be a PR on the RFC.  Here we should follow the rule.

garlick:
I rebased on current master, fixed a couple problems, added a post-processing script, and wired into `scripts/format` and `scripts/check-format`.

However, although this improves the result over what our .clang-format in the tree produces now, it's not there yet.

Edit: I'm not actively working on this - just thought I'd see what claude could do with it.

trws:
What do you see as still lacking @garlick? I had been thinking that now that the per-operator line-breaking config option was added in clang-format-23 we'd be getting pretty close.

edit: meant to include a link: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#breakbinaryoperations

garlick:
Closer.

`BreakBeforeBinaryOperators: NonAssignment` was already there, then I added `WhitespaceSensitiveMacros` to avoid clobbering the custom pack/unpack parameter grouping and the combination(?) caused a trailing `< 0` in a compound conditional block containing pack/unpack to drop to the next line.  Claude found no way to fix that so a script was added to manually post-process.

This still produces a 116K line diff.   While there are a lot of good changes in the output, some other stuff is gratuitous change from current practices which would either need to be a style debate or tracked down.  The good changes probably ought to be extracted and proposed on a subsystem basis, selectively avoiding the areas where people have pending work.

The couple of things I was going to investigate when I decided the rabbit hole was too deep was
- reformatting of static structure array initialization blocks
- indentation of compound conditionals
- perform a deeper review on the diff to find more cases that look wrong (it's not always easy to tell from the diff context alone)

trws:
Adding custom handling to <, == and != to the thing I linked above with never as the policy might take care of those trailing breaks that were bugging us. The last time I was seriously thinking about this, my plan was to add clang-format to pre-commit, and have it run either only on files with diffs or only on the diff itself so we could at least count on it for new code.

grondo:
Having something that could run on new code would really nice now that we're making heavy use of LLMs which seem to forget half the time some of the formatting instructions they're given.

PR #6443, opened 2024-11-15
support `flux uri --wait JOBID`

grondo:
This PR adds support for a `?wait` query parameter to the `jobid` URI resolver, and a new `--wait` option to `flux-uri` to pass that parameter down to the resolver.

This supports waiting until the URI is available for pending or starting jobs, as opposed to the resolver returning "job is not running".

If the `wait` parameter is used with a job that isn't an instance of Flux, then it waits until the `finish` event before giving up and returning "URI not found".

### DIFF
--- a/doc/man1/flux-uri.rst
+++ b/doc/man1/flux-uri.rst
@@ -74,8 +74,15 @@ OPTIONS
 
    Return the *local* (``local://``) equivalent of the resolved URI.
    Warning: the resulting URI may be invalid for the current system
-   if the network host specified by an ``ssh`` URI is not the current
+   if the network host specified by an :program:`ssh` URI is not the current
    host.
+
+.. option:: --wait
+
+   Wait for the URI to become available if the resolver scheme supports it.
+   This is the same as specifying :command:`flux uri TARGET?wait`. Currently
+   only supported by the ``jobid`` resolver. It will be ignored by other
+   schemes.
 
 URI SCHEMES
 ===========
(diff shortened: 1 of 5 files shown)


doc/man1/flux-uri.rst
```diff
@@ -143,6 +157,15 @@ Get the URI of a local flux-broker
    $ flux uri pid:$(pidof -s flux-broker)
    local:///tmp/flux-sLuBkZ/local-0
 
+The following submits a batch job and returns the instance URI once it is
+available. Without the :option:`--wait` flag, :command:`flux uri` would fail
+immediately with "job is not running":
+
+::
+
+   $ flux uri --wait $(flux batch -n1 --wrap flux run hostname)
```
garlick, line 166: In this example, maybe it would be slightly better to use `sleep 3600` or something that runs for a while rather than `hostname`?

doc/man1/flux-uri.rst
```diff
@@ -74,9 +74,16 @@ OPTIONS
 
    Return the *local* (``local://``) equivalent of the resolved URI.
    Warning: the resulting URI may be invalid for the current system
-   if the network host specified by an ``ssh`` URI is not the current
+   if the network host specified by an :program:`ssh` URI is not the current
    host.
```
garlick, line 78: This contains reformatting that isn't mentioned in the commit message.
Might be good to at least call it out.

garlick (approved):
LGTM!  Just a couple of nits on the docs.

doc/man1/flux-uri.rst
grondo: Good point! If we're making a realistic example, maybe just a placeholder `myjob` or something?

garlick: Sounds fine!

doc/man1/flux-uri.rst
grondo: Ok I added that

grondo:
Ok I addressed those comments and will set MWP. Thanks!

Issue #6292, opened 2024-09-19
kvs-wach: support size/offset for watch to begin


chu11:
Per discussion in #4572

> If hypothetically we just stop the watcher, we could restart it but we'd need some kind of "offset" or "seek" mechanism.

chu11:
just a brainstorm, a size/offset probably is hard to do.  pondering if some type of "credits" on the eventlog watch would work.  i.e.

- tell eventlog watch to send up to X data (bytes?  eventlog entries?) and then don't send me more until i say so
- new RPC tells eventlog watch it can send more

hmmm, off the top of my head it feels harder to implement, but would obviously perform better (size/offset might requires users to start/stop the watcher multiple times).

chu11:
while working on  #6456, realized there are some core design stuff that makes a size/offset/index very hard to do / will not perform well.

- The number one issue is that due to kvs merging of commits, multiple events can be inside a single content blob.  That means that we cannot use KVS valref "indexes" as an offset of sorts. (i.e. a user cannot count the number of events they've seen and use that to tell kvs-watch to begin at event number X).

  - related to this, we do not know how many events are inside each blobref.  Hypothetically if there was some metadata along with each blobref in the valref, this could be done.

- There is no apriori knowledge of what the size of a content blob is.  We only know the size after we read it.  So that makes restarting a watch at a certain offset very costly, we cannot (spiritually) "seek" to a location.

  - Within the work on #6264/#6265/#6266, perhaps this can be designed in.  A very stupid idea, instead of the `valref` array only containing a blobref, it could contain an object with the blobref and the size of the data of the blobref (along w/ possibly other metadata).  A less intrusive mechanism would to have a parallel array with the data sizes of each blobref.

chu11:
> Sorry, backing up because I've lost the thread. Is this right?

Yes, this is the "prequel" issue to #6413.

> What sort of interface did you have in mind to provide to the shell? I assume some variant on flux_job_event_watch()?

Before hitting these issues, my thought was the user could count the number of events it received or the amount of data it had received.  If we would overflow the buffer, stop the watcher.  Restart the stdin watcher at some index/offset when we get credits back.  We would perhaps have to buffer the "current" data if we've only sent some portion it.

garlick:
[...]

PR #6444, opened 2024-11-15
kvs-watch: only fetch new data for appends

### LINKED ISSUE #6414: kvs-watch: avoid re-fetching the same data repeatedly
Problem: `flux_kvs_lookup(FLUX_KVS_WATCH | FLUX_KVS_WATCH_APPEND)` is implemented inefficiently.  Although it returns only new data that has been appended to a watched key, the `kvs-watch` module internally fetches the entire value each time the key is mentioned in a `kvs.setroot` event, before returning only the data after the watcher's current offset.

This could ~~almost trivially~~ be made more efficient by asking the KVS for the RFC 11 _valref_ object, which contains an array of blobrefs, and then fetching only the new blobrefs from the content store.

(From brainstorming discussion with @chu11)

The above description probably needs to be checked against the code before we get too excited about implementing this.  However, given that we watch eventlogs a lot in Flux, this could be a high value improvement.

chu11:
Problem: The FLUX_KVS_WATCH_APPEND flag is implemented inefficiently in kvs-watch.  Everytime new data is appended, the entire contents of the watched key are retrieved and only new data calculated via an offset is sent to the watcher.  This significantly slows down performance of large appended data (such as stdio).

Solution: Instead of retrieving the entire contents of the watched key, fetch the tree object for the key from the KVS.  With access to the tree object's blobref array, we need only access the new appended blobrefs from the content store.  This significantly cuts down on the amount of data transfered during a kvs-watch.

Fixes #6414

---

side notes:

- so protocol wise
  - before it was
    - first RPC get all KVS data, return it all, note how long the data is
    - later RPCs, get all KVS data, only send new data past offset of prior request
  - now it is
    - RPCs now get KVS treeobj
      - keep track of how many references are in the array
      - get each blobref data from the content store
      - return that data to the user

So RPC wise the pros are:

- no longer iterating / building larger RPCs than is necessary

the cons are

- adding a level of RPC indirection (getting treeobj)
- possibly sending more RPCs back to caller (first RPC does not "combine" data into a single return)

For large data, this is a significant performance improvement for use cases like this:

`time flux run flux lptest 2000000 80`

wallclock time is generally being cut down from 58 seconds to 24 seconds.  It's a big win.  (FWIW, this was on corona.  On my laptop it was 15s before 11.7s afterwards, not as much as a win there, but still that's 22%)

For everything else (`flux job attach JOBID` for a finished job, job throughput), it appears that performance is a wash.  That the wins balance out the losses.  This is of course with limited testing in single node instances.

Edit: Also as an aside, there is more potential for support of #6292 with this implementation

### DIFF
--- a/doc/man3/flux_kvs_lookup.rst
+++ b/doc/man3/flux_kvs_lookup.rst
@@ -151,10 +151,9 @@ FLUX_KVS_WATCH_UNIQ
 FLUX_KVS_WATCH_APPEND
    Specified along with FLUX_KVS_WATCH, this flag will alter watch
    behavior to only respond when :var:`key` is mentioned verbatim in a
-   committed transaction and the key has been appended to. The response
-   will only contain the additional appended data. Note that only data
-   length is considered for appends and no guarantee is made that prior
-   data hasn't been overwritten.
+   committed transaction and the key has been appended to. The
+   response will only contain the additional appended data.  If the
+   value is overwritten, the lookup fails with EINVAL.
 
 FLUX_KVS_WATCH_FULL
    Specified along with FLUX_KVS_WATCH, this flag will alter watch
--- a/src/common/libkvs/treeobj.c
+++ b/src/common/libkvs/treeobj.c
@@ -648,6 +648,21 @@ char *treeobj_encode (const json_t *obj)
     return json_dumps (obj, JSON_COMPACT|JSON_SORT_KEYS);
 }
 
+const char *treeobj_type_name (const json_t *obj)
+{
+    if (treeobj_is_symlink (obj))
+        return "symlink";
+    else if (treeobj_is_val (obj))
+        return "val";
+    else if (treeobj_is_valref (obj))
+        return "valref";
+    else if (treeobj_is_dir (obj))
+        return "dir";
+    else if (treeobj_is_dirref (obj))
+        return "dirref";
+    return "unknown";
+}
+
 /*
  * vi:tabstop=4 shiftwidth=4 expandtab
  */
--- a/src/modules/kvs-watch/kvs-watch.c
+++ b/src/modules/kvs-watch/kvs-watch.c
@@ -21,6 +21,8 @@
 #include "src/common/libkvs/treeobj.h"
 #include "src/common/libkvs/kvs_util_private.h"
 #include "src/common/libutil/blobref.h"
+#include "src/common/libcontent/content.h"
+#include "src/common/libutil/errprintf.h"
 
 /* State for one watcher */
 struct watcher {
@@ -37,10 +39,13 @@ struct watcher {
     char *key;                  // lookup key
     int flags;                  // kvs_lookup flags
     zlist_t *lookups;           // list of futures, in commit order
+    zlist_t *loads;             // list of futures, content loads in ref order
 
     struct ns_monitor *nsm;     // back pointer for removal
     json_t *prev;               // previous watch value for KVS_WATCH_FULL/UNIQ
-    int append_offset;          // offset for KVS_WATCH_APPEND
+    bool index_valid;           // flag if prev_start_index/prev_end_index set
+    int prev_start_index;       // previous start index loaded
+    int prev_end_index;         // previous end index loaded
     void *handle;               // zlistx_t handle
 };
 
@@ -90,6 +95,12 @@ static void watcher_destroy (struct watcher *w)
                 flux_future_destroy (f);
             zlist_destroy (&w->lookups);
         }
+        if (w->loads) {
+            flux_future_t *f;
+            while ((f = zlist_pop (w->loads)))
+                flux_future_destroy (f);
+            zlist_destroy (&w->loads);
+        }
         json_decref (w->prev);
         free (w);
         errno = saved_errno;
@@ -110,6 +121,8 @@ static struct watcher *watcher_create (const flux_msg_t *msg,
     if (!(w->key = kvs_util_normalize_key (key, NULL)))
         goto error;
     if (!(w->lookups = zlist_new ()))
+        goto error_nomem;
+    if (!(w->loads = zlist_new ()))
         goto error_nomem;
     w->flags = flags;
     w->rootseq = -1;
@@ -230,7 +243,7 @@ static bool key_match (json_t *o, const char *key)
 static void watcher_cleanup (struct ns_monitor *nsm, struct watcher *w)
 {
     /* wait for all in flight lookups to complete before destroying watcher */
-    if (zlist_size (w->lookups) == 0)
+    if (zlist_size (w->lookups) == 0 && zlist_size (w->loads) == 0)
         zlistx_delete (nsm->watchers, w->handle);
     /* if nsm->getrootf, destroy when getroot_continuation completes */
     if (zlistx_size (nsm->watchers) == 0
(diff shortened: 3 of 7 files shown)


src/modules/kvs-watch/kvs-watch.c
```diff
@@ -645,11 +645,13 @@ static void watcher_respond (struct ns_monitor *nsm, struct watcher *w)
     }
     /* flux_kvs_lookup (FLUX_KVS_WATCH)
      *
-     * Ordering note: KVS lookups can be returned out of order.  KVS lookup
-     * futures are added to the w->lookups zlist in commit order here, and
-     * in lookup_continuation(), fulfilled futures are popped off the head
-     * of w->lookups until an unfulfilled future is encountered, so that
-     * responses are always returned to the watcher in commit order.
+     * Ordering note: KVS lookups can be returned out of order due to
+     * what is presently cached.  KVS lookup futures are added to the
```
garlick, line 857: suggestion: instead of
> due to what is presently cached

 how about
> because they are processed asynchronously

doc/man3/flux_kvs_lookup.rst
```diff
@@ -152,9 +152,9 @@ FLUX_KVS_WATCH_APPEND
    Specified along with FLUX_KVS_WATCH, this flag will alter watch
    behavior to only respond when :var:`key` is mentioned verbatim in a
    committed transaction and the key has been appended to. The response
-   will only contain the additional appended data. Note that only data
-   length is considered for appends and no guarantee is made that prior
-   data hasn't been overwritten.
+   will only contain the additional appended data.  This flag will
+   only work with appends with the FLUX_KVS_APPEND.  Overwritten data
+   will result in an error.
```
garlick, line 157: Since the only way to append is with FLUX_KVS_APPEND,  I think the first sentence added here could be dropped.

The second one might be slightly clearer if it read
> If the value is overwritten, the lookup fails with EINVAL

(or whatever errno it fails wiith)

src/modules/kvs-watch/kvs-watch.c
```diff
@@ -248,14 +372,41 @@ static int handle_initial_response (flux_t *h,
...
+            /* since this is a val object, we can just return it */
+            goto out;
+        }
+        else if (treeobj_is_valref (val)) {
+            w->index_set = true;
+            w->prev_start_index = 0;
+            w->prev_end_index = treeobj_get_count (val) - 1;
+        }
+        else {
+            flux_log (h, LOG_ERR, "%s: treeobj not val or valref", __FUNCTION__);
```
garlick, line 391: Drop log message since error is being returned to the user and this is an expected situation as far as this service goes.

src/modules/kvs-watch/kvs-watch.c
```diff
@@ -228,18 +240,130 @@ static bool key_match (json_t *o, const char *key)
...
         zlistx_delete (nsm->watchers, w->handle);
     /* if nsm->getrootf, destroy when getroot_continuation completes */
     if (zlistx_size (nsm->watchers) == 0
         && !nsm->getrootf)
         zhash_delete (nsm->ctx->namespaces, nsm->ns_name);
 }
 
+static void handle_load_response (flux_future_t *f,
+                                  struct watcher *w)
+{
```
garlick, line 253: Style: don't break short parameter lists to multiple lines

garlick (commented):
This is really a nice improvement IMHO!

Before I go further into a review, could you fix the following
- Don't break short function parameter lists to multiple lines
- Don't log errors for user-induced problems.  As a general rule, if sending an error response to the user, don't log it also
- Make an effort to make log messages more helpful that "function name: strerror".

doc/man3/flux_kvs_lookup.rst
garlick: Aside: if we've detected that a watched key has been overwritten and we're watching it with this flag, a human readable error response like "value of key watched with WATCH_APPEND flag was overwritten" might be helpful.

chu11:
re-pushed fixing up a lot of the minor things listed above.  I went with the general rule of making more descriptive errors when an error would be half-likely, but kept "function name: strerror" log messages for things that should extremely unlikely.

src/modules/kvs-watch/kvs-watch.c
```diff
@@ -21,6 +21,7 @@
 #include "src/common/libkvs/treeobj.h"
```
garlick, line 21: commit message: s/everytime/every time/

src/modules/kvs-watch/kvs-watch.c
```diff
@@ -228,18 +240,128 @@ static bool key_match (json_t *o, const char *key)
...
+        json_t *val = treeobj_create_val (data, size);
+        if (!val) {
+            flux_log_error (h, "%s: treeobj_create_val", __FUNCTION__);
+            goto error;
+        }
+        if (flux_respond_pack (h, w->request, "{ s:o }", "val", val) < 0) {
+            flux_log_error (h, "%s: flux_respond_pack", __FUNCTION__);
+            json_decref (val);
+            goto error;
+        }
```
garlick, line 272: After the `goto error`, another error response is attempted.  I think we've settled previously that this is not a good idea.  Just log the (unlikely) error.

In new code I've been trying to log messages like "failed to respond to TOPIC" in these cases.

src/modules/kvs-watch/kvs-watch.c
```diff
@@ -228,18 +240,128 @@ static bool key_match (json_t *o, const char *key)
...
+        errno = ENOMEM;
+        goto error;
+    }
+
+    return f;
+
+error:
+    saved_errno = errno;
+    flux_future_destroy (f);
+    errno = saved_errno;
```
garlick, line 333: flux_future_destroy already preserves errno

src/modules/kvs-watch/kvs-watch.c
```diff
@@ -228,18 +240,128 @@ static bool key_match (json_t *o, const char *key)
...
+    int saved_errno;
+
+    if (!(f = content_load_byblobref (h, ref, 0))) {
+        flux_log_error (h, "%s: content_load_byblobref", __FUNCTION__);
+        goto error;
+    }
+    if (flux_future_then (f, -1., load_continuation, w) < 0) {
+        flux_log_error (h, "%s: flux_future_then", __FUNCTION__);
+        goto error;
+    }
```
garlick, line 322: Suggestion: combine error logs and make it more readable, e.g.
```c
if (!(f = load()) || then() < 0) {
    flux_log_error ("failed to send TOPIC request")
}
```

garlick (commented):
Not going to get through this before meeting so here's what I have so far!

chu11:
re-pushed with fixes per comments above.  Also added a few new cleanups.  I could probably split those out at this point in time.

garlick (approved):
This seems good!  I made a bunch of comments about error messages and some about readability, but I'll tentatively approve and leave up to you how you want to proceed.  It's probably best to get this merged early in our release cycle so we have plenty of time to notice any issues.

Great work!

src/modules/kvs-watch/kvs-watch.c
```diff
@@ -316,78 +460,114 @@ static int handle_compare_response (flux_t *h,
...
-         * that zero length append is legal
-         *
-         * Note that this check does not ensure that the key was not
-         * "fake" appended to.  i.e. the key overwritten with data
-         * longer than the original.
-         */
-        if (new_offset < w->append_offset) {
-            free (new_data);
+        else {
+            (*errmsg) = "value of key watched with WATCH_APPEND overwritten";
```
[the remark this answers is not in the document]
chu11: Hmmm, I think overwritten is correct here.  Something that should be a `valref` is now not a `valref`, indicating the key was overwritten with something new.

src/modules/kvs-watch/kvs-watch.c
```diff
@@ -248,14 +367,39 @@ static int handle_initial_response (flux_t *h,
...
+        }
+        else if (treeobj_is_valref (val)) {
+            w->index_set = true;
+            w->prev_start_index = 0;
+            w->prev_end_index = treeobj_get_count (val) - 1;
+        }
+        else {
+            (*errmsg) = "key watched with WATCH_APPEND is not a value";
+            errno = EINVAL;
             goto error_respond;
```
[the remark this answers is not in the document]
chu11: seems like a good idea.  but to my shock, we apparently don't have a treeobj -> type name conversion function :-)  Or atleast I can't find it.  Lemme write that up fast in another PR.

src/modules/kvs-watch/kvs-watch.c
```diff
@@ -418,8 +598,10 @@ static void handle_lookup_response (flux_future_t *f,
 {
     flux_t *h = flux_future_get_flux (f);
     int errnum;
+    const char *root_ref;
     int root_seq;
     json_t *val;
+    const char *errmsg = NULL;
```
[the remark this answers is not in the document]
chu11: ahh, there's a rebase error.  Due to the changes from 00471780adebad8ef560c89f24125df23f217ce0, the `errmsg` actually does nothing.  Lemme update in general.

chu11:
re-pushed with fixes based on comments above

there are two new commits, one to deal with parameters on different lines and one to add a `treeobj_type_name()` convenience function.  So probably warrants another look @garlick

src/modules/kvs-watch/kvs-watch.c
```diff
@@ -316,78 +460,114 @@ static int handle_compare_response (flux_t *h,
...
+             * response and an updated valref before loads from the content store
+             * have returned to the caller.
+             */
+            if (w->index_set) {
+                int count = treeobj_get_count (val);
+                if ((count - 1) > w->prev_end_index) {
+                    w->prev_start_index = w->prev_end_index + 1;
+                    w->prev_end_index = count - 1;
+                }
+                else if ((count - 1) < w->prev_end_index) {
```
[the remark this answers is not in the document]
chu11: (oops, guess I never hit enter on this yesterday)

I think you you mean `int new_end_index = treeobj_get_count (val) - 1`?

src/common/libkvs/treeobj.c
```diff
@@ -648,6 +648,21 @@ char *treeobj_encode (const json_t *obj)
...
+        return "symlink";
+    else if (treeobj_is_val (obj))
+        return "val";
+    else if (treeobj_is_valref (obj))
+        return "valref";
+    else if (treeobj_is_dir (obj))
+        return "dir";
+    else if (treeobj_is_dirref (obj))
+        return "dirref";
+    return NULL;
```
garlick, line 663: Suggestion: return "unknown" instead of null if the main purpose is logging.

garlick (approved):

chu11:
@garlick thanks, fixed that minor thing, will set MWP

PR #6445, opened 2024-11-16
support partially allocated jobs across scheduler reload

garlick:
This is a proof of concept implementation of the changes to the scheduler hello protocol proposed in flux-framework/rfc#433, in which support is added for reloading the scheduler with housekeeping running and some nodes of job(s) already released.

The scheduler indicates it supports this by calling `schedutil_create()` with the SCHEDUTIL_HELLO_PARTIAL_OK flag.  When it sets that, it agrees to parse an optional `allocated` key in each hello response.  The `allocated` key is set to an idset representing the subset of ranks of R that are actually allocated.  If missing, all ranks are assumed to be allocated.

We need to get some feedback from @milroy, @trws, et al to make sure this approach works for fluxion.  I thought working through this with sched-simple would be helpful to illustrate the idea.

### DIFF
--- a/src/common/Makefile.am
+++ b/src/common/Makefile.am
@@ -138,7 +138,8 @@ libflux_idset_la_LDFLAGS = \
 libflux_schedutil_la_SOURCES =
 libflux_schedutil_la_LIBADD = \
 	$(builddir)/libschedutil/libschedutil.la \
-	$(builddir)/libczmqcontainers/libczmqcontainers.la \
+	$(builddir)/librlist/librlist.la \
+	libflux-internal.la \
 	libflux-core.la \
 	$(JANSSON_LIBS)
 libflux_schedutil_la_LDFLAGS = \
--- a/src/common/libschedutil/hello.c
+++ b/src/common/libschedutil/hello.c
@@ -15,6 +15,9 @@
 #include <jansson.h>
 
 #include "src/common/libjob/idf58.h"
+#include "src/common/librlist/rlist.h"
+#include "src/common/libutil/errno_safe.h"
+
 #include "schedutil_private.h"
 #include "init.h"
 #include "hello.h"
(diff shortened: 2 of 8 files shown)


trws:
I think this would work, but as I look over it I find myself wondering if we might be better off either with a list of things that have been released, or actually leaving this out of hello entirely and using a partial release after the hello response from the scheduler.  The main reason is that this approach sends what is allocated as R, then we have to work out what to release by taking the difference between what is allocated and what needs to be removed before we can act on it.  If instead the released portion is sent, either in hello or as a partial release, we can reuse the same code we already have for partial release without introducing a new code path into what's already proven somewhat complex or fragile. 

That said, working through this as I write, it looks like the job-manager has the allocated resources in this format as pending. That makes me wonder if we should send that `pending` R as the allocation R as part of the base protocol rather than sending the initial R.  As it is in this PR we're sending both right? We certainly need the whole R in job tracking, but I'm not sure the scheduler actually needs it, and it might make the whole thing a bit easier on both ends.

garlick:
> That makes me wonder if we should send that pending R as the allocation R as part of the base protocol rather than sending the initial R. As it is in this PR we're sending both right? We certainly need the whole R in job tracking, but I'm not sure the scheduler actually needs it, and it might make the whole thing a bit easier on both ends.

No, in the current proposal we are just sending the allocated idset (ranks) and some basic job metadata like the id in the RPC responses, then libschedutil looks up the job's original R (including JGF) from the KVS and passes it to the scheduler in its callback.

grondo:
Could libschedutil do the work of creating the partial R from the original R and `free` idset, then directly call the scheduler's `free` callback if necessary after each `hello` callback?

garlick:
> Could libschedutil do the work of creating the partial R from the original R and free idset, then directly call the scheduler's free callback if necessary after each hello callback?

Oh good idea, that makes sense to me.  That gets around the fact that the _R_ fragment in the job manager is missing the JGF.

trws:
Yeah I think having the "free" idset and possibly also doing as @grondo suggested and having schedutil do the translation into a free call would work really well if it's not too difficult to factor that way.

garlick:
ok, pushed those changes.  This is built on top of #6450 currently - will rebase when that gets merged.

The RFC PR will need a small rework for `free` instead of  `allocated`.

garlick:
flux-framework/rfc#433 was updated to specify the (optional) `free` key instead of `allocated` as originally proposed.

Then schedutil was modified so that the `hello` callback subtracts the ranks in the `free` key (if present) from the _R_ that it looks up in the KVS.  _R_ should still include the full original `scheduling` object.

The current behavior is preserved unless the scheduler sets the SCHEDUTIL_HELLO_PARTIAL_OK flag.

garlick:
Was just prepping a PR for the scehduler side, and it does look like there might be a bit of work to do over there.  Could we perhaps get this merged on the core side?  It should behave as before until the scheduler sets the new flag.

grondo (approved):
LGTM, seems fine to merge now as long as no issues with the implementation were discovered on the sched side while working support over there!

garlick:
I don't think we have any design issues over there.  It didn't work in my testing but I think it's likely something that needs fixing on the fluxion side.  I'll set MWP, thanks.

PR #6447, opened 2024-11-18
remove `flux-perilog-run`

grondo:
This PR simply removes the deprecated `flux-perilog-run` script along with tests that used it.

The only other thing that could be added here is to throw an error if `flux perilog-run` is used in a prolog or epilog config. 
However, since we know all the machines that are likely using prolog/epilog and they've all transitioned to using the `per-rank` config, this doesn''t seem necessary. Admins will also notice quite quickly when the prolog and/or epilog fails.

 (However, it is trivial to add if others think it would be helpful)


garlick (approved):
LGTM!

grondo:
Thanks! I've set MWP.

PR #6450, opened 2024-11-19
librlist: avoid unnecessary hwloc dependencies

garlick:
Problem: the librlist.la convenience library includes code that requires hwloc, but often the hwloc bits are not needed.

Split the hwloc parts into a separate convenience library, `librlist-hwloc.la`.

This also drops the ffi python binding for `rlist_from_hwloc()`, which probably is fine.


grondo (approved):
LGTM!

garlick:
Thanks!

PR #6452, opened 2024-11-21
add `-S, --setattr` and `-c, --config-path` options directly to `flux start`

grondo:
I kept getting annoyed that the common broker options for configuration path and setting attributes have to be passed through `flux start`'s `-o, --broker-opts` option. `flux start` is used solely to start brokers, so why can't it just support these common options directly? So, submitted for your consideration is a PR that adds `-S, --setattr` and `-c, --config-path` options directly to `flux start`. If used, these options are just passed along to the broker(s) being started.

WIP because there's no tests or documentation updates in case there's some reason I'm not thinking of that this was a bad idea.

### DIFF
--- a/doc/man1/flux-start.rst
+++ b/doc/man1/flux-start.rst
@@ -122,6 +122,15 @@ as long as the Flux instance is running.  It covers the following use cases:
 
 OPTIONS
 =======
+.. option:: -S, --setattr=ATTR=VAL
+
+   Set broker attribute *ATTR* to *VAL*. This is equivalent to
+   :option:`-o,-SATTR=VAL`.
+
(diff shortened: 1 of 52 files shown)


garlick:
Sounds like a great idea to me.

grondo:
Converted most of the testsuite to use `flux start -S/-c` directly instead of `-o,-S` or `-o,-c` etc. Removed WIP.
(We'll see if this passes CI)

garlick (approved):
LGTM!  Great cleanup!

grondo:
Thanks, coverage is low just because there's a lot of error conditions we can't hit in test. I'll set MWP.

PR #6453, opened 2024-11-21
shell: add extra info to doom exceptions

### LINKED ISSUE #6357: shell: doom timeout exception should include argv[0], jobid of affected job
The exit-timeout exceptions have been somewhat confusing to users because when they pop up in a batch job error log, it isn't immediately clear _which_ job hit the exit timeout.

It might be useful to include the basename of the task that exited (i.e. was it `flux-broker` or my task?) and the jobid so it is clear the exception occurred on a job within the batch job, not the batch job itself, etc.

chu11:
Problem: Exceptions from the doom plugin do not specify what job failed.  It can be confusing to users which job failed amongst multiple in a batch job, or if the broker from a batch job itself failed.
    
Add the job executable name and jobid to the doom exception outputs.

Fixes #6357

---

we could obviously debate the message format, I went back and forth and just ended up at this.  Tweak suggestions welcome.

### DIFF
--- a/src/shell/doom.c
+++ b/src/shell/doom.c
@@ -35,6 +35,7 @@
 
 #include <jansson.h>
 #include <assert.h>
+#include <libgen.h>
 
 #include "src/common/libeventlog/eventlog.h"
 #include "src/common/libutil/fsd.h"
@@ -96,6 +97,14 @@ static const char *doom_exit_host (struct shell_doom *doom)
     return hostlist_nth (doom->hl, nth);
 }
 
+static const char *get_jobspec_command_arg0 (struct shell_doom *doom)
+{
+    json_t *s = json_array_get (doom->shell->info->jobspec->command, 0);
+    const char *path = json_string_value (s);
+    const char *p = strrchr (path, '/');
+    return p ? p + 1 : path;
+}
+
 static void doom_check (struct shell_doom *doom,
                         int rank,
                         int exitcode,
@@ -118,7 +127,8 @@ static void doom_check (struct shell_doom *doom,
 
     if (doom->exit_on_error && doom->exit_rc != 0) {
         shell_die (doom->exit_rc,
-                   "%srank %d on host %s failed and exit-on-error is set",
+                   "%s: %srank %d on host %s failed and exit-on-error is set",
+                   get_jobspec_command_arg0 (doom),
                    doom->lost_shell ? "shell " : "",
                    doom->exit_rank,
                    doom_exit_host (doom));
@@ -206,7 +216,8 @@ static void doom_timeout (flux_reactor_t *r,
 
     fsd_format_duration (fsd, sizeof (fsd), doom->timeout);
     shell_die (doom->exit_rc,
-               "%srank %d on host %s exited and exit-timeout=%s has expired",
+               "%s: %srank %d on host %s exited and exit-timeout=%s has expired",
+               get_jobspec_command_arg0 (doom),
                doom->lost_shell ? "shell " : "",
                doom->exit_rank,
                doom_exit_host (doom),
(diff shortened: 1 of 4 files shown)


grondo:
> Tweak suggestions welcome

We might be able to shorten it by just prefixing the existing exception note with `jobid: arg0: `. Also, I wonder if we should just include the basename of `arg0`? (This is just a hint so the user knows at what level the exception occurred).

Now that I'm thinking about it, I wonder if `flux job attach` should be the thing printing the jobid for job exceptions. Logging the jobid in the eventlog that's specific to a jobid seems redundant in retrospect.

chu11:
> Also, I wonder if we should just include the basename of arg0?

Oh that's a good idea, that just slipped my mind.

chu11:

- shell: only add basename to exception note
- flux job attach: add jobid to output message

so now it looks like this (example from t2614-job-shell-doom)

`1.105s: job.exception type=exec severity=0 jobid=f4m65MP5 test2.sh: shell rank 1 on host corona211 exited and exit-timeout=1s has expired`

grondo:
> 1.105s: job.exception type=exec severity=0 jobid=f4m65MP5 test2.sh: shell rank 1 on host corona211 exited and exit-timeout=1s has expired

Again, it would be shorter to just put the jobid as a prefix:
```
1.105s f4m65MP5: job.exception type=exec severity=0 test2.sh: shell rank1 on host corona211 exited and exit-timeout=1s has expired
```

Since there's no mistaking that's the jobid. However, that's pretty minor and probably not worth quibbling about.

chu11:
ok, re-pushed with the output message tweak.

Also had to tweak two tests that had to be updated for the expected output.

src/shell/doom.c
```diff
@@ -96,6 +97,17 @@ static const char *doom_exit_host (struct shell_doom *doom)
...
 }
 
+static char *get_jobspec_command_arg0 (struct shell_doom *doom)
+{
+    static char buf[PATH_MAX+1] = {0};
+    json_t *s = json_array_get (doom->shell->info->jobspec->command, 0);
+    const char *path = json_string_value (s);
+    char *rv;
+    snprintf (buf, PATH_MAX, "%s", path);
+    rv = basename (buf);
```
grondo, line 107: Would it be simpler to just do something like
```c
const char *p = strrchr (s, "/");
const char *base = p ? p + 1 : s;
```

And avoid the copy? Maybe we should have our own implementation of `basename()` that doesn't modify its argument (I think some places in the codebase may already assume GNU `basename(3)`, which doesn't modify its arg, but since we support alpine and someday BSD, that's probably not safe).

grondo:
Ah, bummer. The change I suggested breaks some flux-sched checks which are matching `flux job attach` output exactly :-(

Sorry if that was a bad suggestion.

src/shell/doom.c
chu11: yeah, that would be better.  When I read the manpage for basename(3) I was like "huh, it needs to modify the buffer?"  Lets go with your simpler approach.

chu11:
> Ah, bummer. The change I suggested breaks some flux-sched checks which are matching flux job attach output exactly :-(

those tests are easily fixable (as is one I missed in `t2304-sched-simple-alloc-check.t`).  Thinking we should keep the new style?  Or are you thinking we should tweak?

grondo:
Up to you if you don't mind submitting a flux-sched PR

chu11:
rebased and re-pushed, should pass flux-sched now that https://github.com/flux-framework/flux-sched/pull/1315 merged.  Although unsure if @trws wants to comment on output format.

trws:
Nah, looks good to me, just helped to have the context. Thanks @chu11!

chu11:
setting WIP, I'm not sure what happened to my branch w/ my tweak from comments above.  Ugh, maybe i lost it during pre-thanksgiving break madness.

grondo (approved):
LGTM!

Issue #2892, opened 2020-04-04
Compiling and running on BSD


SteVwonder:
Per a side-conversation with @trws:

> 1. We use epoll. So either we need to have conditional compilation between using kqueue and epoll or use kqueue unconditionally and use libkqueue to get it on linux. I actually wouldn't hate the latter

> 2. #2891 

> 3. We use the raw CPU masking interface of linux in a number of places, including sched simple. As an alternative, hwloc can be used instead since we already have a dependency on it

The main reason (in my mind) to push for BSD support is supporting Mac OS X as a quality-of-life improvement for developers.  The ability to compile, run, and test on your native OS is a huge plus.

SteVwonder:
> We use epoll. So either we need to have conditional compilation between using kqueue and epoll or use kqueue unconditionally and use libkqueue to get it on linux.

Excellent find from @trws: https://github.com/jiixyj/epoll-shim

alecbcs:
Hey all!

Has there been any movement on building Flux on BSD/MacOS systems in the past two years? I just attempted to, 
```
spack install flux-core
```
 On a Mac but got a failure on the last step when compiling `flux-core` since it couldn't find the version of libuuid provided by Apple.

alecbcs:
@grondo happy to help out here, but maybe we should think of adding a conflicts with MacOS/darwin to the Spack package?

grondo:
Great idea, I think porting flux-core to MacOS will be significant effort at this point.

alecbcs:
Sweet! I'll look to add this to the flux package.py in the next couple of days.

trws:
Yeah, we have a number of glibc and linux specific dependencies at the moment.  Not that it couldn't be done, but it would be a non-trivial effort.

wihobbs:
Tried compiling on macOS today. I got through `./configure` successfully. I know this is still low priority given the other things needed in core, but I thought I'd save the output I got here:

```
devenv] hobbs17@jupiter13 ~/flux-core/build-macos (master)$ make
Making all in .
make[1]: Nothing to be done for `all-am'.
Making all in src
Making all in common
Making all in libtap
make[3]: Nothing to be done for `all'.
Making all in libtestutil
make[3]: Nothing to be done for `all'.
Making all in libev
make[3]: Nothing to be done for `all'.
Making all in libyuarel
make[3]: Nothing to be done for `all'.
Making all in libpmi
  CC       pmi2.lo
[...]

PR #6454, opened 2024-11-21
fix macos portability issues

garlick:
Here's a bit of work towards portability of flux-core to macos.  This has been going on at a low level for some time, documented in #2892.

Instructions for preschool level macos user like me:

0) install xcode (eh, I typed `git` and a dialog box did the rest)
1) install homebrew (download pkg from homebrew site)
2) run `scripts/install-deps-macos.sh`
3) run `scripts/configure-macos.sh`
4) make

I am moving on to other things today.  The current status is compilation fails in libev:
```
Making all in libev
  CC       ev.lo
ev.c:3036:21: error: variable has incomplete type 'struct itimerspec'
 3036 |   struct itimerspec its = { 0 };
      |                     ^
/opt/homebrew/opt/epoll-shim/include/libepoll-shim/sys/timerfd.h:17:8: note: forward declaration of 'struct itimerspec'
   17 | struct itimerspec;
      |        ^
1 error generated.
```

### DIFF
--- a/scripts/configure-macos.sh
+++ b/scripts/configure-macos.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+die() {
+    echo "$(basename $0): $@" >&2
+    exit 1
+}
+
+DEPS_SCRIPT=scripts/install-deps-macos.sh
+
+test -f $DEPS_SCRIPT || die "please run from the top level of the source tree"
+test -d macos-venv || die "please run $DEPS_SCRIPT first"
+
+eval "$(/opt/homebrew/bin/brew shellenv)"
+
+CPPFLAGS=-I${HOMEBREW_PREFIX}/include/lua
+CPPFLAGS="-I$(brew --prefix epoll-shim)/include/libepoll-shim ${CPPFLAGS}"
+LDFLAGS=-L${HOMEBREW_PREFIX}/lib
+
+PKG_CONFIG_PATH=$(pkg-config --variable pc_path pkg-config)
+PKG_CONFIG_PATH=$(brew --prefix libarchive)/lib/pkgconfig:${PKG_CONFIG_PATH}
+
+PATH=$(brew --prefix libtool)/libexec/gnubin:$PATH
+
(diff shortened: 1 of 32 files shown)


scripts/configure-macos.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+die() {
+    echo "$(basename $0): $@" >&2
+    exit 1
+}
+
+test -d flux-env || die "please run from the top of the source tree"
```
grondo, line 8: Since there's a `flux-env` command, maybe naming this directory `macos-venv` or something would be more clear? Also, the user may have just not yet run `install-deps-macos.sh`, so maybe the error message should suggest that, e.g. "Ensure you've run scripts/install-deps.macos.sh and are in the top-level source directory"

grondo (approved):
This LGTM. I wonder if we should just merge this as-is so your work doesn't get out of date? Others can continue the work in the future?

garlick:
Ok, I backed out the attempt at supporting epoll via the shim as this was getting further before I started including the shim headers.

Now I think libev is being built without epoll and we fail when building libflux:
```
Making all in libflux
  CC       flog.lo
  CC       attr.lo
  CC       handle.lo
handle.c:21:10: fatal error: 'sys/epoll.h' file not found
   21 | #include <sys/epoll.h>
      |          ^~~~~~~~~~~~~
1 error generated.
make[3]: *** [handle.lo] Error 1
make[2]: *** [all-recursive] Error 1
make[1]: *** [all-recursive] Error 1
make: *** [all-recursive] Error 1
```
I'll drop the WIP and I'm fine with merging if you are :-)

trws:
Writing this here partly to help myself remember, but I think the shim issue probably means we want to:

1. Use the shim for flux itself
2. Use libev's built-in kqueue support by having autoconf specify it when building libev

My naive guess would be that when libev finds epoll it assumes a bunch of other linuxisms (timers or pidfds maybe?) also exist, and so it's blowing up.  Have not yet tested this.

PR #6458, opened 2024-11-21
kvs-watch: misc cleanup

chu11:
splitting these out of #6444


garlick (approved):
LGTM!

chu11:
oops, re-pushed.  One corner case was fixed in #6444 but never brought down to this cleanup PR.

PR #6459, opened 2024-11-22
doc: improve `--include` documentation in flux-resource(1)

grondo:
@schoate had pointed out that the `flux resource list --include` option is difficult to discover because the option name itself isn't exactly obvious and the man page description of the option doesn't include the term "node' (only host and rank).

Update the `-i, --include` option documentation to include the word "node" for improved search.


garlick (approved):
Good improvement!

Issue #6275, opened 2024-09-10
Suggestion: consolidate empty lines in flux resource list's default format


kkier:
The current/new-in-.65 behavior is to output a full line for each empty queue in `flux resource list` like so:
```
[root@elcap1:~]# flux resource list
[root@elcap1:~]# flux resource list
     STATE QUEUE      NNODES   NCORES    NGPUS NODELIST
      free foo             0        0        0 
      free bar             0        0        0 
      free baz	           0        0        0 
      free qux             0        0        0 
      free quux            0        0        0 
 allocated foo             0        0        0 
 allocated bar             0        0        0 
 allocated baz             0        0        0 
 allocated qux             0        0        0 
 allocated quux            0        0        0 
      down foo,qux         1        1        1 elcap[XXX-XXX]
      down foo,bar         2        2        2 elcap[XXX-XXX]
      down foo,baz         3        3        3 elcap[XXX-XXX]
      down foo,quux        4        4        4 elcap[XXX-XXX]
```
I'm wondering if it might be cleaner to, in the case of lines with 0 nodes associated, use `state` as a primary indicator and combine the queue list like we do with overlapping queues, so you'd end up with something like:
```
     STATE QUEUE      NNODES   NCORES    NGPUS NODELIST
      free bar,quux        0        0        0 
      free baz             1        1        1
 allocated foo,bar,baz     0        0        0     
 ```
 Immediate problem I see is that this causes some conceptual overlap between "the first line above indicates these queues have 0 nodes in this state" and "the first line above indicates that there are 0 nodes in this state who that are also members of both of these queues."  I don't think of that as being confusing offhand, but I can see how that might not be universal.

To put it another way, in the example above does this line mean "there are no nodes in queues `bar` or `quux` that are free," or does it mean "there are no nodes in both queues `bar` and `quux` that are free, but there may be nodes in either of them individually free."

Reading this back, I'm not sure if it makes 100% sense...

**Comment by kkier** [comment]
To explain a little more simply, I propose something where in the first situation you would end up with this output:
```
[root@elcap1:~]# flux resource list -o "{state:>10} ?:{queue:<20.20} ?:{propertiesx:<10.10+} {nnodes:>6} {ncores:>8} ?:{ngpus:>8} {nodelist}"
     STATE QUEUE                NNODES   NCORES    NGPUS NODELIST
      free foo,bar,baz,qux,quux      0        0        0 
 allocated foo,bar,baz,qux,quux      0        0        0 
[...]

PR #6460, opened 2024-11-22
skip "empty" lines of output in `flux resource list` with `--skip-empty` or `--include`

grondo:
We didn't really come to a resolution in issue #6275 - however, it is clear that this output is surprising and wrong:
```console
$ flux resource list --include=corona171
     STATE QUEUE   NNODES NCORES NGPUS NODELIST
      free pdebug       1     48     8 corona171
 allocated pbatch       0      0     0 
 allocated pdebug       0      0     0 
 allocated pnvmeof      0      0     0 
      down pbatch       0      0     0 
      down pdebug       0      0     0 
      down pnvmeof      0      0     0 
```

This PR adds a `--skip-empty` option to `flux resource list` to skip all lines that represent empty resource sets, then makes this the default for `-i, --include` (which is probably the expected and least surprising behavior). So, instead of the above, we get:
```console
$ src/cmd/flux resource list --include=corona171
     STATE QUEUE  NNODES NCORES NGPUS NODELIST
      free pdebug      1     48     8 corona171
```

There is also now a `--no-skip-empty` option which disables the option with `--include`.

### DIFF
--- a/doc/man1/flux-resource.rst
+++ b/doc/man1/flux-resource.rst
@@ -106,6 +106,15 @@ all
   hostlist form. It is not an error to specify ranks, nodes, or hosts which
   do not exist.
 
+.. option:: --skip-empty
+
+  Skip lines with empty resource sets in output. This is the default when
+  `-i, --include` is used.
+
+.. options:: --no-skip-empty
+
+  Do not skip empty resource sets in output, even with `-i, --include`.
+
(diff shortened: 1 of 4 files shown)


garlick (approved):
LGTM!

grondo:
Thanks! Will set MWP

PR #6471, opened 2024-12-03
add flux module stats --rusage=[self|children|thread] optional argument

garlick:
Problem: `getrusage(2)` accepts multiple `who` targets, but flux hard codes the non-portable RUSAGE_THREAD.

Switch the default to RUSAGE_SELF, which is portable.

Add a `who` key to the RPC and change `flux module stats --rusage` so it takes an optional argument of `self`, `children`, or `thread`.

Update docs and add a few tests.

### DIFF
--- a/t/t0003-module.t
+++ b/t/t0003-module.t
@@ -184,25 +184,21 @@ test_expect_success 'flux module stats --scale works' '
 	test "$EVENT_TX2" -eq $((${EVENT_TX}*2))
 '
 
-
 test_expect_success 'flux module stats --rusage works' '
-	flux module stats --rusage $REALMOD >rusage.stats &&
-	grep -q utime rusage.stats &&
-	grep -q stime rusage.stats &&
-	grep -q maxrss rusage.stats &&
-	grep -q ixrss rusage.stats &&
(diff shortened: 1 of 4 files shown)


t/t0003-module.t
```diff
@@ -186,23 +186,16 @@ test_expect_success 'flux module stats --scale works' '
...
+	flux module stats --rusage $REALMOD
+'
+test_expect_success 'flux module stats -Rself works' '
+	flux module stats -Rself $REALMOD
+'
+test_expect_success 'flux module stats --rusage=children works' '
+	flux module stats --rusage=children $REALMOD
+'
+test_expect_success 'flux module stats --rusage=badopt fails' '
+	test_must_fail flux module stats --rusage=badopt $REALMOD
```
chu11, line 198: as i re-skim this, maybe add a comment mentioning not covering `thread` intentionally b/c not portable

chu11 (approved):
LGTM, just thought of a nit thing below

garlick:
Thanks!  I added a note plus a `test_might_fail` test that should improve coverage slightly.  Setting MWP.

PR #6473, opened 2024-12-04
fix macos portability issues (third batch)

garlick:
This fixes some more macos portability issues.

### DIFF
--- a/src/common/libutil/basename.c
+++ b/src/common/libutil/basename.c
@@ -1,0 +1,28 @@
+/************************************************************\
+ * Copyright 2016 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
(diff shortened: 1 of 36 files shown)


src/common/libutil/basename.c
```diff
@@ -0,0 +1,28 @@
...
+
+/* This is what glibc basename(3) does more or less.
+ * https://github.com/lattera/glibc/blob/master/string/basename.c
+ */
+
+char *basename_simple (const char *path)
+{
+    char *p = strrchr (path, '/');
+    return p ? p + 1 : (char *)path;
+}
```
grondo, line 26: Nice. Does this undo the need for 92c2ece8206d1b2940de54685b64120754efcf2b if all uses of basename were updated to `basename_simple()`?

I know these are just portability updates, but perhaps we can update the code in `shell/doom.c` that does the same thing to use `basename_simple.c` instead:

https://github.com/flux-framework/flux-core/blob/11ad605648fafbaf4837ceda2a7751a5d53efb3e/src/shell/doom.c#L100-L106

grondo (approved):
LGTM!

src/common/libutil/basename.c
garlick: Oh sure, we can do that.  I was trying to have a light touch but this makes sense.

garlick:
Thanks!  Everything is using `basename_simple()` including the local function you pointed out and one more I found.  I'll set MWP.

PR #6475, opened 2024-12-05
drop caliper support

garlick:
Problem: caliper test setfaults in CI after modifying linkage arguments for macos portability.

The caliper hooks are probably in the wrong places at this point anyway. Drop caliper support from Flux for now.

This can be reverted when someone has time to spend working on the caliper hooks and linkage.


grondo (approved):
LGTM!

PR #6478, opened 2024-12-05
flux-hostlist: allow idset argument to `--nth` and `--exclude` options

### LINKED ISSUE #6464: flux hostlist: request idset with -x or --exclude
I'm looking at `flux hostlist` as an alternative to what I'd do "the janky way":

```bash
nodes=$(flux job info $FLUX_JOB_ID R | jq -r .execution.nodelist[0])
```
That's the approach I chose to get the list for a script running in the prolog, which I wasn't sure if it would be considered local / in the instance (so I wouldn't do `flux hostlist local` or `flux hostlist instance`. First, I'm not sure if this is expected?

```bash
$ flux hostlist hosts
hosts
```

I don't think I totally understand this use case, it seems to just give me back what I give it:

```bash
$ flux hostlist $(hostname)
i-0831eed34c13e747e
```

(Sorry, I'm dumb sometimes). But these are what I expected!

```bash
$ flux hostlist local
i-0831eed34c13e747e,i-0ac10f9b787d6a349
```
```bash
ubuntu@i-0831eed34c13e747e:~$ flux hostlist instance
i-0831eed34c13e747e,i-0ac10f9b787d6a349
```
Although I don't think I could use them in a prolog - I'd want to use the `flux hostlist jobid` I think?
I think I found what I am looking for - a means to get the entire list minus the lead broker (control plane for user space Kubernetes) which thankfully, I'm running everything from. Like this:

```bash
$ flux hostlist -x $(hostname) local
i-0ac10f9b787d6a349
```

But I'm wondering - since `-x 0` for a lot of other commands explicitly asks to exclude a rank, could we have something like that here? It's very likely that I wouldn't know the hostnames to exclude in all cases. E.g.,, I'd want:

```bash
$ flux hostlist -x 0 local
i-0ac10f9b787d6a349
```

I think I'd also like to be able to quickly get a count (this is what I'm looking to do now.

```bash
$ flux hostlist  local --count
2
$ flux hostlist  -x 0 local --count
1
```

And maybe there is already a way to do that. I'm doing it rather ugly:

```bash
nodes=$(flux hostlist -x $(hostname) local)
counter=(${nodes//","/ })
count=${#counter[@]}
```
Anyway, a little bash never hurt anyone - but maybe there is a feature or two to be added (count and -x to support ranks) or clarification (e.g., hosts). BTW, unrelated notes, I found `kvs get` and `kvs put` because I couldn't use `flux job update` to add metadata, and it's pretty awesome. Sometimes using the plumbing isn't so bad! :toilet:

grondo:
This PR makes the following minor improvements to `flux hostlist`:

 * `-n, --nth=IDS`: This option now takes an idset instead of a single index into the hostlist. This allows a user to get a "slice" of a hostlist while preserving current behavior, e.g. `flux hostlist -n 0-1 JOBID` to get the first two hosts of a job.
 
 * `-x, --exclude=HOSTS|IDS`: As suggested by @vsoch, this option now takes either a hostlist or idset or indices to exclude from the hostlist. This allows a user to exclude the first host for example with `flux hostlist -x 0 local`
 
Fixes #6464

### DIFF
--- a/doc/man1/flux-hostlist.rst
+++ b/doc/man1/flux-hostlist.rst
@@ -63,10 +63,12 @@ OPTIONS
   Emit the number of hosts in the result hostlist instead of the hostlist
   itself.
 
-.. option:: -n, --nth=N
+.. option:: -n, --nth=IDS
 
-  Output only the host at index *N* (*-N* to index from the end). The command
-  will fail if *N* is not a valid index.
+  Output only the hosts at indices *IDS* (*-IDS* to index from the end),
+  where *IDS* is a valid RFC 22 idset (e.g. '0' will return the first host,
+  '0-1' will return the first and second, '-1' returns the last host).  The
+  command will fail if any id in *IDS* is not a valid index.
 
 .. option:: -L, --limit=N
 
@@ -82,9 +84,11 @@ OPTIONS
   other manipulation options, this is equivalent to returning the set
   union of all provided hosts. (By default, all inputs are concatenated).
 
-.. option:: -x, --exclude=HOSTS
+.. option:: -x, --exclude=HOSTS|IDS
 
-  Exclude all occurrences of *HOSTS* form the result.
+  Exclude all hosts in  *HOSTS* or indices in idset *IDS* from the result.
+  It is not an error if any hosts or indices do not exist in the target
+  hostlist.
 
 .. option:: -i, --intersect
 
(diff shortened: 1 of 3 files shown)


garlick (approved):
LGTM!

grondo:
Thanks! Fixed the issue codeql flagged and forced a push. Will set MWP.

PR #6479, opened 2024-12-05
fix macos portability issues (fifth batch)

garlick:
Last batch of macos fixes.

I _think_ this gets everything compiled, including tests but there are still unit test failures to deal with.  Anyway, this is as far as I got for now.

### DIFF
--- a/src/common/libmissing/pipe2.c
+++ b/src/common/libmissing/pipe2.c
@@ -15,14 +15,26 @@
 #include <fcntl.h>
 #include <errno.h>
 
+#include "src/common/libutil/fdutils.h"
+
 #include "pipe2.h"
 
 static int setflags (int fd, int flags)
 {
-    int oflags;
-    if ((oflags = fcntl (fd, F_GETFL)) < 0
-        || fcntl (fd, F_SETFL, oflags | flags) < 0)
+    int valid_flags = O_CLOEXEC | O_NONBLOCK;
+
+    if ((flags & ~valid_flags)) {
+        errno = EINVAL;
         return -1;
+    }
+    if ((flags & O_CLOEXEC)) {
+        if (fd_set_cloexec (fd) < 0)
+            return -1;
+    }
+    if ((flags & O_NONBLOCK)) {
+        if (fd_set_nonblocking (fd) < 0)
+            return -1;
+    }
     return 0;
 }
 
--- a/src/common/libutil/test/fdwalk.c
+++ b/src/common/libutil/test/fdwalk.c
@@ -46,6 +46,14 @@ static void set_fd (void *data, int fd)
     fds[fd]++;
 }
 
+static void set_fd_if_open (void *data, int fd)
+{
+    if (fcntl (fd, F_GETFL) < 0 && errno == EBADF)
+        return;
+    int *fds = data;
+    fds[fd]++;
+}
+
 static int * get_open_fds (int maxfd)
 {
     /* Valgrind may show open fds > maxfd, so double the space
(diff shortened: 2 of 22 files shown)


src/common/libutil/test/fdwalk.c
```diff
@@ -46,14 +46,22 @@ static void set_fd (void *data, int fd)
...
+
 static int * get_open_fds (int maxfd)
 {
     /* Valgrind may show open fds > maxfd, so double the space
      *  allocated so we don't overflow when run under valgrind.
      */
     int * fds = calloc (maxfd * 2, sizeof (int));
     if (fds)
-        ok (fdwalk (set_fd, fds) == 0,
+        ok (fdwalk (set_fd_if_open, fds) == 0,
```
chu11, line 64: should we also do this in `test_fdwalk_fallback()`?

src/common/libmissing/pipe2.c
```diff
@@ -15,14 +15,26 @@
...
         return -1;
+    }
+    if ((flags & O_CLOEXEC)) {
+        if (fd_set_cloexec (fd) < 0)
+            return -1;
+    }
+    if ((flags & O_NONBLOCK)) {
+        if (fd_set_nonblocking (fd) < 0)
+            return -1;
+    }
```
chu11, line 37: what you did here doesn't match your commit message.  Not sure if you changed mind on what to do?  hopefully not a rebase mistake?

garlick: Those are just helper functions that do what it says in the commit message but I guess that could be made more clear.

src/common/libutil/test/fdwalk.c
garlick: I'm not seeing much motivation for doing that, are you?

src/common/libmissing/pipe2.c
garlick: Just pushed a tweak to the commit message

src/common/libutil/test/fdwalk.c
chu11: just wanted to make sure it wasn't missed by accident :-)

src/common/libmissing/pipe2.c
chu11: Ahh, you know what it is.  Seeing the very generic prior implementation, the appearance is that more/other flags would have been supported.

But I now see that `pipe2()` only supports three flags, one of which appears to be linux only.   so the new `setflags()` supports every flag for `pipe2()`

chu11 (approved):
LGTM!

src/shell/oom.c
```diff
@@ -20,7 +20,9 @@
 #endif
 #include <sys/types.h>
 #include <sys/stat.h>
+#ifdef HAVE_SYS_INOTIFY_H
```
grondo, line 23: It seems like this plugin should just be conditionally not compiled on macos, since we know cgroups are never supported there? However, perhaps this way is just easier since you don't have to modify the bulitin plugin list?

grondo (approved):
This LGTM. If compilation is working then we should consider adding a macos build to CI to avoid future PRs breaking the build.

src/shell/oom.c
garlick: Well that seems obvious now that you mention it. Let me try that way.

garlick:
OK, I reworked the oom plugin it's just not built if inotify is unavailable.

Perhaps  we can add a macos builder another day?  It would have to just do a `make` since `make check-prep` would require unit tests to pass, which they do not currently.

garlick:
Chicken mon dieu!  This macos nightmare is finally over for now.

grondo:
> Perhaps we can add a macos builder another day? It would have to just do a make since make check-prep would require unit tests to pass, which they do not currently.

Ok. If/when we do that we'll probably have to fix portability issues we introduce between now and then. Since we have no macos users and it doesn't work anyway, that's pretty low priority IMO.

garlick:
I'll open an issue on that one for now.

PR #6482, opened 2024-12-05
resource: only read `resource.scheduling` config on rank 0

### LINKED ISSUE #6480: resource.scheduling key should only be used on rank 0 if possible
Problem: The Rv1 `scheduling` key as JGF can be large and time-consuming to generate and distribute, but currently it is required on all nodes. Since the key is really only used by the scheduler on rank 0, it would be ideal `resource.scheduling` was only read on rank 0.

grondo:
Problem: The Rv1 scheduling key as JGF can be large and time-consuming to generate and distribute, but currently it is required on all nodes. Since the key is really only used by the scheduler on rank 0, it would be ideal resource.scheduling was only read on rank 0.

Only process the resource.scheduling config key on rank 0.

Fixes #6480

### DIFF
--- a/src/modules/resource/resource.c
+++ b/src/modules/resource/resource.c
@@ -116,7 +116,10 @@ static int parse_config (struct resource_ctx *ctx,
             return -1;
         }
     }
-    if (scheduling_path) {
+    /* resource.scheduling key, if configured, is only required on rank 0,
+     * since by definition it is used only by the scheduler.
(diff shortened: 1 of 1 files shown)


garlick (approved):
Oh yeah duh!  Nice.

grondo:
Fixed a commit message typo and will set MWP. Thanks!

PR #6486, opened 2024-12-06
add `flux.conf_builtin.conf_builtin_get()` to give Python access to compiled-in config values

grondo:
Currently, Python programs have no way to query libflux for compiled-in config values, but this could be useful if they want to find the current config directory. This could possibly be derived from broker attributes, but in some cases the broker attributes could be set to nothing, so this is not ideal.

Calling in to `flux_conf_builtin_get()` with `FLUX_CONF_AUTO`  using ffi doesn't work, because the "executable" for Python programs is always going to be a `python` binary installed in the system (or elsewhere), so the libflux intree detection doesn't work.

This PR introduces a wrapper for `flux_conf_builtin_get()` for Python that reimplements intree detection using the path to module itself compared to the intree python_path.

Since the motivating use case is to determine the current appropriate Flux confdir from a Python program, this PR also adds a new builtin config key `confdir` for this purpose.

### DIFF
--- a/src/bindings/python/flux/Makefile.am
+++ b/src/bindings/python/flux/Makefile.am
@@ -11,6 +11,7 @@ nobase_fluxpy_PYTHON = \
 	memoized_property.py \
 	debugged.py \
 	importer.py \
+	conf_builtin.py \
 	cli/__init__.py \
 	cli/base.py \
 	cli/alloc.py \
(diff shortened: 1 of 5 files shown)


garlick (approved):
LGTM!

grondo:
Thanks! I'll set MWP.

PR #6489, opened 2024-12-06
shell: ignore SIGPIPE

### LINKED ISSUE #6487: shell: turn fatal EPIPE error into warning
Problem: The shell currently dies if it attempts to write to stdin, but one or more tasks have closed it. This can be reproduced simply with
```console
$ flux run sh -c 'exec <&-; sleep 60'
# type a line to stdin
flux-job: job shell Broken pipe
```

The job shell should be ignoring SIGPIPE, and EPIPE errors writing to stdin should either be warnings or ignored.

grondo:
This PR fixes #6487 by ignoring `SIGPIPE` in the job shell.

At attempt at adding a test was made, but the test is inherently racy, because if the data is sent from the shell to the task before stdin is actually closed, the issue wouldn't reproduce (and also there's no way to synchronize on the job shell actually writing the data to the task, so the job could get canceled before any data is written) However, it seem sufficient for now.

### DIFF
--- a/src/shell/signals.c
+++ b/src/shell/signals.c
@@ -16,6 +16,7 @@
  * SIGINT  - forward to all local tasks
  * SIGTERM - forward
  * SIGALRM - forward
+ * SIGPIPE - ignore
  *
  * Notes:
  *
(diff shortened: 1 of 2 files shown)


garlick (approved):
Great, LGTM!

grondo:
Thanks! Setting MWP.

PR #6502, opened 2024-12-12
job-info: avoid error response on failed rpc

### LINKED ISSUE #6498: job-info: avoid error response on failed rpc
similar to fixes in

https://github.com/flux-framework/flux-core/pull/6458/commits/00471780adebad8ef560c89f24125df23f217ce0

need to do it in job-info, minimally in guest_watch.c

chu11:
Problem: Common usage in flux-core is to not respond with an RPC error
(such as through flux_respond_error()) if a prior RPC attempt fails (such
as through flux_respond_pack()).  This pattern is violated in several
places in job-info.

Solution: If flux_respond_pack() fails, do not call flux_respond_error()
in error handling paths.

Fixes #6498

### DIFF
--- a/src/modules/job-info/guest_watch.c
+++ b/src/modules/job-info/guest_watch.c
@@ -706,9 +706,9 @@ static int main_namespace_lookup (struct guest_watch_ctx *gw)
         goto error;
 
     if (!(gw->main_namespace_lookup_f = flux_rpc_message (gw->ctx->h,
-                                                         msg,
-                                                         FLUX_NODEID_ANY,
-                                                         0))) {
+                                                          msg,
+                                                          FLUX_NODEID_ANY,
+                                                          0))) {
         flux_log_error (gw->ctx->h, "%s: flux_rpc_message", __FUNCTION__);
         goto error;
     }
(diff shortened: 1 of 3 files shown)


src/modules/job-info/guest_watch.c
```diff
@@ -763,7 +768,7 @@ static void main_namespace_lookup_continuation (flux_future_t *f, void *arg)
                                "event", tok, toklen) < 0) {
             flux_log_error (ctx->h, "%s: flux_respond_pack",
                             __FUNCTION__);
-            goto error;
+            goto cleanup;
         }
     }
```
garlick, line 774: Good catch!   Though maybe the `skip_respond_error` flag could be avoided if you just moved the `error_cancel:` code into the one block that uses it and `goto cleanup` from there?

src/modules/job-info/update.c
```diff
@@ -143,6 +143,7 @@ static void eventlog_continuation (flux_future_t *f, void *arg)
     json_t *context = NULL;
     const char *errmsg = NULL;
     const flux_msg_t *msg;
+    bool skip_respond_error = false;
```
garlick, line 147: Same comment on this one.   Just call `eventlog_watch_cancel()` at each `goto error_cancel` site and `goto error` or `goto cleanup` from there as appropriate

garlick (approved):
LGTM although I suggested a little cleanup to make the code more readable.

chu11:
thanks!  Re-pushed with those tweeks, will set MWP

PR #6503, opened 2024-12-12
kvs: add rst format to inline docs

wihobbs:
This might be a problem other places too.

What renders:
image

After this PR:
image

Is this really what we want? RST shoved in our python in-line docs? I'm not sure...

### DIFF
--- a/src/bindings/python/flux/kvs.py
+++ b/src/bindings/python/flux/kvs.py
@@ -61,13 +61,13 @@ def exists(flux_handle, key, namespace=None):
     Args:
         flux_handle: A Flux handle obtained from flux.Flux()
         key: key to check for existence
-
-    Returns:
-        bool: True if key exists, False if not
         namespace: namespace to read from, defaults to None.  If namespace
           is None, the namespace specified in the FLUX_KVS_NAMESPACE
           environment variable will be used.  If FLUX_KVS_NAMESPACE is not
           set, the primary namespace will be used.
+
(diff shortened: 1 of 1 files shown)


grondo:
> Is this really what we want? RST shoved in our python in-line docs? I'm not sure...

it doesn't really bother me.. Does it work if you just use `::` (seems like sphinx should default to Python syntax highlighting when auto-doc-ing python docs...

The other solution I've seen is to use inline [doctest](https://docs.python.org/3/library/doctest.html) markers, e.g.

```
>>> mydir = KVSDir(flux_handle)
>>> print(mydir)
```

though I guess in that case it should be more of a full example? (e.g. `import flux` `flux_handle = flux.Flux()` etc)

wihobbs:
I went with Mark's suggestion to make the examples more complete. This handles all of the multi-line formatting weirdness. It would be great if everything had code formatting but that isn't worth it to me at the moment.

For some reason the docs build is not showing my changes... sigh

wihobbs:
Ok, this is ready for review. I _did_ fix two issues I saw where a parameter had been accidentally appended to a `Returns:` instead of with the `Args:`.

Also FYI to those of you who review with RTD's PR rendering, mine was being cached which caused pages to show old commit versions, so clear caches or look in private mode to be sure you're getting the most recent one.

grondo (approved):
LGTM! Thanks!

PR #6504, opened 2024-12-13
fix handling of `\n` in submission cli `--add-file=` option

grondo:
This PR fixes a small bug in the submission cli `--add-file=[NAME=]DATA` option. The option is documented to allow file data on the command line if `DATA` contains a `\n`. However, any `\n` passed on the command line ends up quoted, so this doesn't actually work:
```console
$ flux submit --add-file=test='this is a test\n' hostname
flux-submit: ERROR: --add-file=test=this is a test\n: [Errno 2] No such file or directory: 'this is a test\\n'
```

This PR just unescapes in newlines in `DATA` so that newline detection works (and the newlines appear as newlines in the resulting fileref data).

### DIFF
--- a/src/bindings/python/flux/cli/base.py
+++ b/src/bindings/python/flux/cli/base.py
@@ -969,7 +969,10 @@ class MiniCmd:
 
     def handle_add_file_arg(self, jobspec, arg):
         """Process a single argument to --add-file=ARG."""
-        name, _, data = arg.partition("=")
+        #  Note: Replace any newline escaped by the shell with literal '\n'
+        #  so that newline detection below works for file data passed on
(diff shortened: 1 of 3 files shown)


src/bindings/python/flux/cli/base.py
```diff
@@ -969,7 +969,10 @@ def init_jobspec(self, args):
```
garlick, line 969: commit message: s/indicaates/indicates/

garlick (approved):
LGTM!  Just one commit message nit.

grondo:
Thanks! Fixed that and will set MWP.

PR #6505, opened 2024-12-13
support specification of file permissions in `--add-file` submission option

grondo:
This PR adds support for specifying file permissions in the `--add-file` option in Flux submission command. (Presently, all files added with content directly provided on the command like get the default permissions of 0600)

Example:
```
$ flux submit --add-file=test:700='#!/bin/sh\ntrue\n' -n1 {{tmpdir}}/test
```

### DIFF
--- a/doc/man1/common/job-other-options.rst
+++ b/doc/man1/common/job-other-options.rst
@@ -58,20 +58,22 @@
    character, then VAL is interpreted as a file, which must be valid JSON,
    to use as the attribute value.
 
-.. option:: --add-file=[NAME=]ARG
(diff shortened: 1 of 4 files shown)


garlick (approved):
LGTM!

Definitely not worth holding this up, but I wonder if there is a loose python module out there that supports chmod(1)'s "symbolic mode" (`chmod +x`, etc)?

grondo:
I did look for something, but only found solutions for parsing the ls output style e.g. `-rwxr-x---` which didn't seem that useful. For now I just wanted to get basic ability to specify perms succinctly on the command line.

grondo:
Thanks I've set MWP.

Issue #4696, opened 2022-10-18
per-task output from flux mini commands


ryanday36:
A corona user is requesting per-task output from the `flux mini` commands:

> For splitting output/error, I mean more like the “IO Redirection” section of the srun man page. E.g., “-o output-%t.log” to get the output from each rank to log to a unique file (useful for “debugging by printf”; less so for “production”). For a 4-rank job, this will produce “output-0.log”, “output-1.log”, “output-2.log”, and “output-3.log”. Maybe I misunderstood the flux docs, but my understanding was that the “flux-{{id}}.out” would just write “flux-F0123456789.out” – i.e., one file for the aggregated output of the entire job.

It looks like there has been some work to support `{{taskid}}` in #2460, but it doesn't appear to work in `flux mini --output`.

grondo:
I'm not sure that one was ever merged because it ended up being quite complex. It might be simplest to just have the rank 0 shell open one file per task for an initial implementation. O/w, a workaround for now would be to run the tasks via a wrapper script that redirects output using shell output redirection. I can work up an example a bit later this AM.

ryanday36:
It sounds like the user already has a workaround in place. I just wanted to get the issue opened so that we can track it.

Issue #2875, opened 2020-03-28
Plugin capabilities for front-end submission tools


SteVwonder:
We have come across a few use-cases recently that would benefit from the ability to add custom arguments to the flux front-end submission tools (i.e., flux-mini and flux-run) and then modify the jobspec generation based on those arguments.  Two use-cases that come to mind:

- Mpibind:  assuming mpibind can specify everything it needs to apriori, it could add a `--bind` argument (similar to what is done with srun) and dynamically change the jobspec based on the user's request
- Current collaborations with IIT on their LABIOS IO system:  we would like users to specify QoS requests in LABIOS-specific terms (e.g., 100GB optimized for latency and 1TB optimized for bandwidth), and then we want to convert those QoS requests into resource requests and insert them into the jobspec being submitted.  (CC @keith-bateman)
- Darshan and other site-specific tools: when a user submits a job, maybe they want to enable IO tracing with Darshan via a `--trace-io` flag.  This could then set the `LD_PRELOAD` environment variable in the jobspec appropriately.  Looping back to the mpibind example, if it cannot modify the jobspec to meet it's needs, it could always insert itself in the beginning the task list in the jobspec.  All that being said, this use-cases might be better suited for a job-shell plugin.

Rough sketch of the idea:
- the front-end tools generate jobspec how they normally would
- the front-end tool calls the custom plugins, one-by-one, passing the full jobspec in for mutation (a pipeline of sorts)
- the order of the plugins needs to be determined (via a config file?)

@grondo: did you add this type of functionality to Slurm as a part of SPANK plugins?

grondo:
This came up again with the notification service proposed in flux-framework/rfc#414.

While job shell options are not that onerous to specify on the command line, e.g. `flux run -o mpibind=ARGS...`, this does not also hold true for extensions (such as the notification service) which may use arbitrary jobspec attributes (e.g. from that RFC PR: `--setattr=system.notify.service=slack --setattr=system.notify.events=validate,epilog-start --setattr=system.notify.include="{id.f58} {R} {eventlog}`)

Another example is the flux-accounting bank, which requires `--setattr=bank=BANK` when `--bank=BANK` would be much more user friendly.

cc: @wihobbs

grondo:
[...]

PR #6506, opened 2024-12-13
allow mustache templates in job environment variables

grondo:
The job shell expands mustache templates for input/output filenames as well as in command arguments. It would also be useful to support this template expansion in environment variables, but currently this is not supported. It probably isn't efficient to wholesale expand mustache templates in all environment variables, since the common case would be no variables to expand. Also, some environment variables may happen to contain mustache templates which should not be expanded.

This PR introduces a new `env-expand` shell builtin plugin which applies mustache expansion to any environment variables in a new `env-expand` shell options key in jobspec.

On the cli side, the `get_filtered_environment()` function is extended to look for environment variable values specified on the command line and pass these back in a dictionary separate from the main environment. This dictionary is then saved in the `attributes.system.shell.options.env-expand` key for use by the shell plugin.

This means that only environment variables specified on the command line or in a file as `VAR=TEMPLATE` where `TEMPLATE` contains `{{}}` will be expanded. Variables in the environment that already do contain `{{}}` pass through the normal environment dictionary unmodified. Additionally, the `env-expand` shell plugin only has to process the minimum number of variables.

This is a WIP so I can get feedback on the approach before writing tests.

The current use case here is to place files in the job's tmpdir and set environment variables in the job to point to these files, e.g. to extend rc1:

```
flux alloc --add-file=rc1.d/rc1:700=`#!/bin/sh\nflux exec dosomething` --env=FLUX_RC_EXTRA={{tmpdir}}
```

### DIFF
--- a/doc/man1/common/job-env-rules.rst
+++ b/doc/man1/common/job-env-rules.rst
@@ -31,8 +31,13 @@ via a set of *RULE* expressions. The currently supported rules are
      * Advanced parameter substitution is not allowed, e.g. ``${var:-foo}``
        will raise an error.
 
+   ``VAL`` may also contain a mustache template, in which case the template
+   will be substituted in the job shell with the corresponding value before
+   launching job tasks. See `MUSTACHE TEMPLATES`_ for more information.
+
    Examples:
-       ``PATH=/bin``, ``PATH=$PATH:/bin``, ``FOO=${BAR}something``
+       ``PATH=/bin``, ``PATH=$PATH:/bin``, ``FOO=${BAR}something``,
+       ``PATH=${PATH}:/{{tmpdir}}/bin``
 
  * Otherwise, the rule is considered a pattern from which to match
    variables from the process environment if they do not exist in
--- a/src/bindings/python/flux/cli/base.py
+++ b/src/bindings/python/flux/cli/base.py
@@ -288,10 +288,11 @@ def get_filtered_environment(rules, environ=None):
     Filter environment dictionary 'environ' given a list of rules.
     Each rule can filter, set, or modify the existing environment.
     """
+    env_expand = {}
     if environ is None:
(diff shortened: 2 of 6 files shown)


garlick:
This seems like it could be really powerful.  Nice!

grondo:
I've just pushed an update (though this PR should still be considered an experimental WIP) that adds per-task environment variable mustache template expansion, along with some new task-specific mustache tags:
 
 - `{{taskid}}`, `{{task.id}}`, or `{{task.rank}}` - global task rank
 - `{{task.index}}` or `{{task.localid}}` - local task rank

Additionally, a set of job and node specific mustache tags are added:

     - `{{nnodes}}` and `{{size}}` for the total node and task count
     - A set of `node.*` tags including:
      - `{{node.id}}`: index of this shell relative to job
      - `{{node.ntasks}}`: number of tasks local to this node
      - `{{node.name}}`: hostname
      - `{{node.ncores}}`: number of allocated cores on this node
      - `{{node.ngpus}}`: number of allocated gpus on this node
      - `{{node.taskids}`: task idset, e.g. 0-3
      - `{{node.coreids}}`: core idset, e.g. "0-3"
      - `{{node.gpuids}}`: gpu idset e.g. "0"

With this, users could "opt-in" to having some equivalent environment variables to the [60 or set by default by Slurm](https://slurm.schedmd.com/archive/slurm-22.05.9/srun.html#SECTION_OUTPUT-ENVIRONMENT-VARIABLES).

This could also be useful in the cli plugins prototyped in #2875.

Finally, if the shell output plugin is modified to render the output file name per task, we can possibly close #4696, since a unique template per-task or per-node would result in per-task or per-node output files.

garlick (approved):
Excellent :+1:

grondo:
Thanks! I'll set MWP.

PR #6507, opened 2024-12-13
testsuite: don't assume /bin/true and /bin/false

garlick:
Problem: tests that hard code `/bin/true` or `/bin/false` fail on macos, where those programs are in `/usr/bin`.

When PATH is already being searched, don't specify a fully qualfied path.

### DIFF
--- a/src/broker/test/runat.c
+++ b/src/broker/test/runat.c
@@ -78,8 +78,8 @@ void basic (flux_t *h)
         "runat_is_defined name=test1 returns false");
     ok (runat_is_completed (r, "test1") == false,
         "runat_is_completed name=test1 returns false");
-    ok (runat_push_shell_command (r, "test1", "/bin/true", 0) == 0
-        && runat_push_shell_command (r, "test1", "/bin/true", 0) == 0,
+    ok (runat_push_shell_command (r, "test1", "true", 0) == 0
+        && runat_push_shell_command (r, "test1", "true", 0) == 0,
         "pushed true;true");
     ok (runat_is_defined (r, "test1") == true,
         "runat_is_defined name=test1 returns true after creation");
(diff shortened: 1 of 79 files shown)


chu11:
Would going with `/usr/bin/true` work?  Appears to be ok for RHEL9 and my ubuntu box.  Without any context, just seeing `true` and `false`, it looks like it could be boolean.  Like

```
test_expect_success 'rc3 failure causes instance failure' '
	test_expect_code 1 flux start \
		-Sbroker.rc3_path=false \
		-Slog-stderr-level=6 \
		true 2>rc3_failure.log
'
```

like `false` could be a special option to `rc3_path`.

Maybe it's just me.

garlick:
Oops, pushed a fix to the ASAN error - I inadvertently changed a string length in the broker runat test without changing another value that was supposed to be its length + 1.

> Without any context, just seeing true and false, it looks like it could be boolean

That would probably work although I think it's _pretty_ clear what's going on in these tests.   I'd be inclined to not continue to hard code a path - it feels slightly wrong.

chu11:
> That would probably work although I think it's pretty clear what's going on in these tests. I'd be inclined to not continue to hard code a path - it feels slightly wrong.

By all means, it certainly livable.  Just something I noticed at first glance.

garlick:
Maybe we ought to get @grondo's input on that one.

grondo (approved):
This approach seems fine to me. To be fair, we have some other paths we hard code in the testsuite including `/bin/sh`, `/bin/bash`, and `/bin/echo`. I'm guessing those are ok because they exist in those paths on macos?

garlick:
These are all test changes, so I think the rule of "add a test when you fix something" probably doesn't apply or at least not as strongly.  And the person who finally gets macos/bsd to run all the tests may have to fix whatever pops up between now and then, but I think that'll be the least of their worries.

If we do it now, then at least they will have a lesser job, and the cut and pasters won't make the problem worse.

garlick:
> This approach seems fine to me. To be fair, we have some other paths we hard code in the testsuite including /bin/sh, /bin/bash, and /bin/echo. I'm guessing those are ok because they exist in those paths on macos?

Yeah, the problem is "fails on macos" and I wanted to keep the solution focused on that.

I'm seeing thumbs up and this is approved so I'll go ahead and set MWP.  Thanks!

grondo:
> Yeah, the problem is "fails on macos" and I wanted to keep the solution focused on that.

I figured. My point was it went against the "it's wrong to hard code a path" argument made previously in the comments.

Issue #6485, opened 2024-12-06
macos: add configure --with-external-libev


garlick:
Problem: `make check` fails on macos in early unit tests.

Just the first couple of issues that may indicate actual problems as opposed to non-portable tests:

- `libflux/test_future.t` fails on
   > Assertion failed: (("libev: I/O watcher with invalid fd found in epoll_ctl", errno != EBADF && errno != ELOOP && errno != EINVAL)), function epoll_modify, file ev_epoll.c, line 134.
- `libflux/test_reactor.t` hangs in `test_child()`.  Possibly child watchers don't work, which is rather critical

garlick:
Both of these problems seem to stem from libev using the epoll-shim, when it actually has support for kqueue.  At least building flux-core against the brew version of libev appears to get around them.

If we do that on macos then we can also revert fd4dd445b709f7af66b2869a698eb2227563e967

I'll retitle this to reflect the actual problem.

Issue #6404, opened 2024-10-31
broker/test: test_overlay.t: error binding to ipc://@test0: Address already in use


chu11:
After #6389 was merged broker test `test_overlay.t` fails

```
ok 69 - test0: no messages received within 1.0s
# check_monitor BEGIN
# created test0: rank 0 size 5 uuid 221f37f6-5867-4571-a7a3-6e5f2375b23b
test_overlay.t: error binding to ipc://@test0: Address already in use
Bail out!  test0: overlay_bind failed
```

beginning to look into it

chu11:
It appears PR #6389 changed the name of some ipc targets, so I was guessing maybe it collided with some lingering kruft from some other process?

Thinking perhaps the node I was on had some lingering kruft in some other process, tried a different node on the same cluster and it also failed.

But, on nothing more than a guess, the tests pass on another cluster.  

could it be racy?

chu11:
this temporarily makes the tests later in the file pass

```
diff --git a/src/broker/test/overlay.c b/src/broker/test/overlay.c
index dfe9ed02b..ca1c928ca 100644
--- a/src/broker/test/overlay.c
+++ b/src/broker/test/overlay.c
@@ -722,8 +722,8 @@ int main (int argc, char *argv[])
     single (h);
     clear_list (logs);
 
-    trio (h);
-    clear_list (logs);
+    /* trio (h); */
+    /* clear_list (logs); */
 
     check_monitor (h);
     clear_list (logs);
```

suggesting it's racy (`single()` sets up test with same binding as `trio()` and `check_monitor()`).

Then I tried

```
diff --git a/src/broker/test/overlay.c b/src/broker/test/overlay.c
index dfe9ed02b..858a3067f 100644
--- a/src/broker/test/overlay.c
+++ b/src/broker/test/overlay.c
@@ -722,9 +722,13 @@ int main (int argc, char *argv[])
     single (h);
     clear_list (logs);
 
+    sleep(1);
+
     trio (h);
     clear_list (logs);
 
+    sleep(1);
+
     check_monitor (h);
     clear_list (logs);
```

and tests pass, again suggesting raciness

chu11:
ugh ... all i can think of is to just retry until it works?  unfortunately gotta re-init the `ctx_create()` ... i can pretty this up of course

```
-            if (overlay_bind (ctx[0]->ov, uri) < 0)
-                BAIL_OUT ("%s: overlay_bind failed", ctx[0]->name);
[...]

PR #6508, opened 2024-12-14
fix macos portability issues (sixth batch)

garlick:
This fixes a few more issues, plus adds the capability to build flux-core with an external libev for #6485.  Finally, run the working unit tests under macos so we don't end up with regressions there.

This is built on top of #6404 because localizing the libev code mad this PR easier.


grondo (approved):
LGTM!

PR #6509, opened 2024-12-15
build macos with homebrew lua that is not deprecated

garlick:
Problem: macos ci builds are failing (that didn't take long!) because apparently the lua I chose was about to be deprecated, and now is.

This fixes that.


grondo (approved):
LGTM, thanks!

PR #6515, opened 2024-12-16
job-exec: improve drain message for unkillable procs

### LINKED ISSUE #6495: improve drain message: "unkillable processes for job fNiY3EqzUPq"
Problem: job-exec drains nodes with the message in the subject line.  Sys admins  would like the message to indicate a) that it's user processes that are part of a job that are unkillable, as opposed to housekeeping or a flux process, and b) if possible, indicate the hostlist that had unkillable processes.

grondo:
Problem: The drain message issued by the job-exec module when it gives up trying to kill user processes in the job is not clear.

Note that the unkillable job procesess were user processes to avoid confusion that epilog or housekeeping processes were at fault.

Update one test that attempts to match the old error message.

Fixes #6495

### DIFF
--- a/src/modules/job-exec/job-exec.c
+++ b/src/modules/job-exec/job-exec.c
@@ -425,7 +425,7 @@ static int drain_active_ranks (struct jobinfo *job, struct idset *active_ranks)
     }
     rc = jobinfo_drain_ranks (job,
                               ranks,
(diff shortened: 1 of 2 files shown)


garlick (approved):
LGTM!

grondo:
Thanks! I've set MWP.

PR #6518, opened 2024-12-16
job-info: stream events even if job is inactive

### LINKED ISSUE #6516: job-info: stream appended data, do not return all data in 1 giant blob
(Similar to the work in #6444)

When a job is completed, there is presumably no more updates to job eventlogs, therefore any watched data does a kvs-lookup on the data.  This can be very slow if there is a large amount of data.  For example, think of `flux job attach JOBID` on a job that is completed and has a bajillion lines of output.  The kvs-lookup will return all of the standard output in 1 big reply.

Instead we should stream this data no different than if the job was live to improve turnaround time.  Right now `flux job attach JOBID` has the appearance of a hang in this scenario.

Note that started with work in #6456 but decided that it should be split off into a separate issue.

chu11:
Problem: If a job is inactive, all data in an eventlog will be returned
as a single response during an eventlog watch.  This is because we know for
a fact that the data should never change after the job is inactive.

If this data, such as job standard output, is very large, this lookup
can be very slow.  In some cases, the use of something like `flux job attach`
can have the apperance of a hang because the standard output response is
taking so long to lookup and return.

Solution:

When a job is inactive and the user wants to watch a job eventlog,
do not respond with all of the data in a single response.  Instead stream
the response back just as if the job were active.

Utilize the FLUX_KVS_WATCH_APPEND_ONCE to ensure the stream ends once all
data in the KVS is streamed.  Update all variables, functions, etc. from
"lookup" to "watch".

Fixes #6516

---

I probably need to add some new tests but wanted to throw this up here for now.

Side note on the "WATCH_APPEND_ONCE" flag, I could make this generic and apply it to all watch types (i.e. even non-appends) but didn't feel it was needed elsewhere so didn't do that.

### DIFF
--- a/src/modules/job-info/guest_watch.c
+++ b/src/modules/job-info/guest_watch.c
@@ -117,7 +117,9 @@ struct guest_watch_ctx {
     bool guest_released;
 
     /* data from guest namespace */
-    int offset;
+    int guest_offset;
+    /* data from main namespace */
+    int main_offset;
 };
 
 static int get_main_eventlog (struct guest_watch_ctx *gw);
@@ -194,22 +196,8 @@ static int send_eventlog_watch_cancel (struct guest_watch_ctx *gw,
                 f = gw->wait_guest_namespace_f;
             else if (gw->state == GUEST_WATCH_STATE_GUEST_NAMESPACE_WATCH)
                 f = gw->guest_namespace_watch_f;
-            else if (gw->state == GUEST_WATCH_STATE_MAIN_NAMESPACE_LOOKUP) {
-                /* Since this is a lookup, we don't need to perform an actual
-                 * cancel to "job-info.eventlog-watch-cancel".  Just return
-                 * ENODATA to the caller if necessary.
-                 */
-                gw->eventlog_watch_canceled = true;
-                if (gw->cancel) {
-                    if (flux_respond_error (gw->ctx->h,
-                                            gw->msg,
-                                            ENODATA,
-                                            NULL) < 0)
-                        flux_log_error (gw->ctx->h, "%s: flux_respond_error",
-                                        __FUNCTION__);
-                }
-                return 0;
-            }
+            else if (gw->state == GUEST_WATCH_STATE_MAIN_NAMESPACE_LOOKUP)
+                f = gw->main_namespace_lookup_f;
             else {
                 /* gw->state == GUEST_WATCH_STATE_INIT, eventlog-watch
                  * never started so sort of "auto-canceled" */
@@ -649,7 +637,7 @@ static void guest_namespace_watch_continuation (flux_future_t *f, void *arg)
         goto cleanup;
     }
 
-    gw->offset += strlen (event);
+    gw->guest_offset += strlen (event);
     flux_future_reset (f);
     return;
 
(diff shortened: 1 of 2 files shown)


garlick:
Naive question: could the client just do a FLUX_KVS_WATCH_APPEND instead of the lookup and then just cancel it when the `clean` event is read?

chu11:
oh i forgot, doing

```
> flux run flux lptest 2000000 80
<snip all that output>
> time flux job attach `flux job last`
```

on master the `flux job attach` takes about 17 seconds.  On this branch it is now down to about 13 seconds on average on corona.  And the noticeable "pause" at the beginning is no longer there.

chu11:
> Naive question: could the client just do a FLUX_KVS_WATCH_APPEND instead of the lookup and then just cancel it when the clean event is read?

Perhaps my description of what I'm trying to solve isn't clear.  We already know for a fact that the 'clean' event has arleady happened.  The job is already done / inactive.

So the issue is if you do a FLUX_KVS_WATCH_APPEND instead of a lookup on a job that is inactive/done, how do you know when the issue the cancel?  You don't really know.

So that's why I wanted to put the "WATCH_APPEND_ONCE" flag in on the `kvs-watch` side to finish the stream when it knows it is done.

garlick:
Oh duh, sorry,  I was thinking about watching the primary eventlog which always ends in `clean` but this is about the output eventlog, which doesn't have a definitive end of file.

Since the user is just passing this to `flux_kvs_lookup()`, maybe the flag name and description should be divorced from the concept of "watch" (even though internally it uses the watch service), e.g.
```
FLUX_KVS_STREAM
    Return a potentially large value in multiple responses terminated by an ENODATA error response.
```

chu11:
> Since the user is just passing this to flux_kvs_lookup(), maybe the flag name and description should be divorced from the concept of "watch" (even though internally it uses the watch service), e.g.

Ahhh, I like this idea.  Lemme see  how to go about this.

chu11:
still need to add tests but went with the proposed FLUX_KVS_STREAM (still need tests & documentation)

chu11:
just re-pushed, splitting off the FLUX_KVS_STREAM option into #6253.  Removing WIP.

chu11:
rebased and re-pushed now that #6523 is in, just tweaked some commit messages.

src/modules/job-info/guest_watch.c
```diff
@@ -43,7 +43,7 @@
  *
  * 2) If the guest namespace is already copied into the main namespace
  *    (event "release" and "final=true"), we watch the eventlog in the
- *    main namespace (main_namespace_lookup()).  This is the "easy" case
+ *    main namespace (main_namespace_watch()).  This is the "easy" case
```
garlick, line 46: Wondering about this name change - once the guest info is copied to the primary namespace, it is static, but "watch" implies that it is changing.

chu11: yeah, i did go back and forth on it.  I went with `watch` b/c I go through `job-info.eventlog-watch` instead of calling `kvs.lookup` (similarly `guest_namespace_watch` also goes through `job-info.eventlog-watch`).   

I did consider going with `stream` instead, but that sort of describes internal knowledge of what `job-info.eventlog-watch` is doing.

garlick (approved):
LGTM!

chu11:
thanks, setting MWP

chu11:
```
  expecting success: 
  	cat 5Mfile | flux exec -r 0-3 --setopt=stdin_BUFSIZE=4096 ./stdin2file &&
  	test_must_fail ls cpy.0 &&
  	cmp 5Mfile cpy.1 &&
  	cmp 5Mfile cpy.2 &&
  	cmp 5Mfile cpy.3 &&
  	rm cpy.*
  
  flux-exec: flux_subprocess_write: Broken pipe
  not ok 45 - stdin flow control works (all ranks, one rank will exit early)
```

failure from the `bookworm - gcc-12,content-s3,distcheck` builder.  

Maybe something racy in here?  EPIPE normally associated with writing to a subprocess that is closed or not yet running.

PR #6520, opened 2024-12-17
libschedutil: use preprocessor symbols for flags

garlick:
Problem: its laborious to test flux-core for the existence of schedutil flags because they are anonymous enums.

Convert to preprocessor symbols that can be easily checked with an #ifdef in the scheduler.


grondo (approved):
LGTM!

garlick:
Thanks! I'll set MWP.

PR #6525, opened 2024-12-19
shell: support more mustache tags

grondo:
This PR adds new mustache tags specific to tasks and nodes (shell rank, really), formatted as `{{task.*}}` and `{{node.*}}`, with a couple special cases. The node tags use the "rank info" object in the shell to render results, which has been updated with additional items for mustache tag use.

Since there are many more supported tags to document, a new _MUSTACHE TEMPLATES_ section is added to the job submission ccommands' manual pages, pasted here since it contains the set of supported tags:

## MUSTACHE TEMPLATES

Submission commands support a simplified version of mustache templates in select options (e.g., `--output`, `--error`, `--dump`), and job commands/arguments. These templates are replaced by job details once available. A mustache template is a tag in double braces, like `{{tag}}`.

The following is a list of currently supported tags:

> **Note**  
> The template system in the job shell allows plugins to extend the set of supported mustache tags. Check with your installation of Flux to determine if extra tags are available locally.

### Job mustache tags:

- `{{id}}` or `{{jobid}}`  
  Expands to the current jobid in F58 encoding. If needed, an alternate encoding may be selected by using a subkey with the name of the desired encoding, e.g., `{{id.dec}}`. Supported encodings include `f58` (the default), `dec`, `hex`, `dothex`, and `words`.

- `{{name}}`  
  Expands to the current job name. If a name is not set for the job, then the basename of the command will be used.

- `{{nnodes}}`  
  Expands to the number of nodes allocated to the job.

- `{{size}}` or `{{ntasks}}`  
  Expands to the job size, or total number of tasks in the job.

### Node-specific tags

Node-specific tags are prefixed with `node.`, and support any of the keys present in the object returned by `flux_shell_get_rank_info`. These include:

- `{{node.id}}`  
  Expands to the current node index within the job (0 - nnodes-1).

- `{{node.name}}`  
  Expands to the hostname of the current node.

- `{{node.broker_rank}}`  
  Expands to the broker rank of the enclosing instance on this node.

- `{{node.cores}}`  
  Expands to the core id list assigned to the current node in idset form. For example, `0-3` or `1`.

- `{{node.gpus}}`  
  Expands to the GPU id list assigned to the job on this node in idset form, e.g., `0-1` or `4`.

- `{{node.ncores}}`  
  Expands to the number of cores allocated to the job on this node.

### Task-specific tags

Task-specific tags are prefixed with `task.`, with the special case of `{{taskid}}`, which is an alias for `{{task.id}}`.

> **Note**  
> The `--output` and `--error` options do not support task-specific tags at this time.

- `{{task.id}}`, `{{task.rank}}`, or `{{taskid}}`  
  Expands to the global rank for the current task

- `{{task.index}}` or `{{task.localid}}`
  Expands to the local rank for the current task
  
### Other tags:

- `{{tmpdir}}`
  Expands to the path of the job temporary directory on the local system.
  
If an unknown mustache tag is encountered, an error message may be displayed and the unknown template will appear in the result.

### DIFF
--- a/doc/man1/flux-shell.rst
+++ b/doc/man1/flux-shell.rst
@@ -610,13 +610,17 @@ supported. Job shell specific functions and tables are described below:
   then the current rank is used.  Returns a table of rank-specific
   information including:
 
+  **id**
+    The current shell rank (matches ``shell_rank`` input parameter)
+  **name**
+    The host name for ``shell_rank``
   **broker_rank**
     The broker rank on which ``shell_rank`` is running.
   **ntasks**
     The number of local tasks assigned to ``shell_rank``.
   **resources**
     A table of resources by name (e.g. "core", "gpu") assigned to
-    ``shell_rank``, e.g. ``{ core = "0-1", gpu = "0" }``.
+    ``shell_rank``, e.g. ``{ ncores = 2, core = "0-1", gpu = "0" }``.
 
 **shell.log(msg)**, **shell.debug(msg)**, **shell.log_error(msg)**
   Log messages to the shell log facility at INFO, DEBUG, and ERROR
--- a/src/shell/shell.c
+++ b/src/shell/shell.c
@@ -1175,7 +1175,7 @@ static int mustache_render_jobid (flux_shell_t *shell,
 
     if (strlen (name) > 2) {
         if (name[2] != '.') {
-            shell_log_error ("Unknown mustache tag '%s'", name);
+            errno = ENOENT;
             return -1;
         }
         type = name+3;
@@ -1246,7 +1246,7 @@ static int mustache_cb (FILE *fp, const char *name, void *arg)
                                 "{s:s}",
                                 "result", &result) < 0
         || result == NULL) {
-        shell_log_error ("Unknown mustache tag '%s'", name);
+        errno = ENOENT;
         goto out;
     }
     if (fputs (result, fp) < 0) {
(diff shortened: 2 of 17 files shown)


src/shell/shell.c
```diff
@@ -577,6 +577,8 @@ static json_t *flux_shell_get_rank_info_object (flux_shell_t *shell, int rank)
     char *taskids = NULL;
```
garlick, line 578: commit message: s/usefule/useful/

doc/man1/flux-shell.rst
```diff
@@ -610,13 +610,17 @@ supported. Job shell specific functions and tables are described below:
   then the current rank is used.  Returns a table of rank-specific
   information including:
 
+  **id**
+    The current shell rank (matches ``shell_rank`` input parameter
```
garlick, line 614: need a close paren

garlick (approved):
LGTM!  Just a couple minor nits

grondo:
Thanks! Fixed those and I'll set MWP.

grondo:
Thanks! Fixed those errors and will set MWP.

grondo:
Hm, two builders failed in `t0005-exec.t`:
```
expecting success: 
  	cat 5Mfile | flux exec -r 0-3 --setopt=stdin_BUFSIZE=4096 ./stdin2file &&
  	test_must_fail ls cpy.0 &&
  	cmp 5Mfile cpy.1 &&
  	cmp 5Mfile cpy.2 &&
  	cmp 5Mfile cpy.3 &&
  	rm cpy.*
  
  flux-exec: flux_subprocess_write: Broken pipe
  not ok 45 - stdin flow control works (all ranks, one rank will exit early)
```
I assume this is unrelated to this PR so I'll restart those.

PR #6531, opened 2024-12-20
flux-kvs: improve eventlog error messages

chu11:
Problem: When an improperly formatted eventlog is input into eventlog commands, a non-useful error message is output.

Output a better error message.

PR #6532, opened 2024-12-21
WIP kvs-watch: do not load all KVS blobs at once

### LINKED ISSUE #6456: kvs-watch: retrieve content blobs in chunks rather than all at once
While #6414 / #6444 solves an important initial problem of not re-fetching older unnecessary data, whenever there is a large amount of data to retrieve, it sends requests to the content store to retrieve all data.

The difference can be seen in something like

```
> flux run flux lptest 2000000 80`
<wait job done>
> flux job attach `flux job last`
```

In the first job, when stdout is being "watched", the very first lookup of the stdout will not be that large.  And every subsequent append that is watched should be only 1 or a few lines of output.

In the follow up `flux job attach`, the job is already completed.  So the very first lookup of the tree object for stdout will be all 2 million lines.  kvs-watch will subsequently do a lookup of all 2 million lines.  i.e. it's a for loop doing 2 million iterations and sending 2 million RPCs.  That's not good.

Instead, only a subset should be retrieved in "windows".  This allows data to be sent back to the caller much more quickly.

chu11:
Problem: With the FLUX_KVS_WATCH_APPEND and FLUX_KVS_STREAM flags, all content blobs in a valref treeobj will be retrieved from the content store at once.  If the valref array is gigantic, this may be a very costly initial transaction.  Not to mention, we probably wouldn't want to send an extremely large number (like millions) of content requests all at once.

Solution: Send content requests in more reasonable 32K chunks.

Fixes #6456

----

So this is what I initially started working on that lead more to #6531 and #6518.

This issue is mostly theoretical, as I don't have a use case where this is a problem at the moment.  (i.e. perhaps it's a solution looking for a problem?)  I want to try and manufacture a use case and test it, which is why I'm posting this as WIP.  But I thought I'd post and see if folks think it's atleast a good idea.  

I should note, I did run the test suite when having the "max loads" set to 2 instead of 32K, so it generally speaking appears to work safely.  Although the question of how big the "maximum number of loads at a time" number could possibly be tweaked up or down some.

### DIFF
--- a/src/modules/kvs-watch/kvs-watch.c
+++ b/src/modules/kvs-watch/kvs-watch.c
@@ -24,6 +24,11 @@
 #include "src/common/libcontent/content.h"
 #include "src/common/libutil/errprintf.h"
 
+/* if the valref is very large, we won't load all immediately,
+ * load at max MAX_LOAD
+ */
+#define MAX_LOADS 32768
+
 /* State for one watcher */
 struct watcher {
     const flux_msg_t *request;  // request message
@@ -43,9 +48,11 @@ struct watcher {
 
     struct ns_monitor *nsm;     // back pointer for removal
     json_t *prev;               // previous watch value for KVS_WATCH_FULL/UNIQ
-    bool index_valid;           // flag if prev_start_index/prev_end_index set
-    int prev_start_index;       // previous start index loaded
-    int prev_end_index;         // previous end index loaded
+    bool index_valid;           // flag if start_index/end_index set
+    int start_index;            // start index of blobrefs to load
+    int end_index;              // end index of blobrefs to load
+    int last_index_loaded;      // last index sent
+    json_t *valref_treeobj;     // valref for loads
     int loaded_blob_count;      // number of indices loaded (for FLUX_KVS_STREAM)
     void *handle;               // zlistx_t handle
 };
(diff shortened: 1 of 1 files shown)


chu11:
I tried an experiment where I created an eventlog with about 400K entries (all individually appended onto the log) to see if this helps or not.

before

```
>time flux kvs get test.bigdata > /dev/null                             
2.775u 0.699s 0:20.04 17.2%     0+0k 816+0io 4pf+0w                                                                                         
>time flux kvs get --stream test.bigdata > /dev/null                    
4.825u 2.510s 0:15.82 46.3%     0+0k 0+0io 0pf+0w          
```

after

```
>time flux kvs get test.bigdata > /dev/null                             
2.768u 0.757s 0:19.51 17.9%     0+0k 0+0io 0pf+0w                                                                                           
>time flux kvs get --stream test.bigdata > /dev/null                    
5.408u 3.985s 0:17.70 52.9%     0+0k 0+0io 0pf+0w                                                                                           
```

It doesn't seem like it really matters much.

If it doesn't seem to matter much at 400K entries, it probably just isn't worthwhile.  Perhaps it would theoretically matter at like milllions of entries, but that just isn't going to be likely or common.

So I'll close and just keep a mental note that this work was done at one point in time.

PR #6533, opened 2024-12-21
fix possible truncation of strings with "expandable" output fields

grondo:
This PR fixes truncation of strings in output formats that use the `+:` sentinel, when the field width is greater than the minimum precision (e.g. `+:{field:10.10}` and the maximum width > 10)


garlick (approved):
LGTM! :+1:

grondo:
Thanks!

Issue #6510, opened 2024-12-16
eliminate flux_reactor_active_incref()


garlick:
Problem:  `flux_reactor_active_incref()` and `flux_reactor_active_decref()` have no users and have a questionable design:

- From 5bd662097d11c6a212b270e33a18abc3a8e91cdb: they only work with certain watchers
  >  The use of flux_reactor_active_incref/decref() only works with
    simple watchers that only implement a single watch.  More complex
    watchers, which may have multiple watchers internal to it, cannot
    use this mechanism.
- From #2387: the interface is lame
  > My only worry is that incref/stop phase of this approach will be pretty easily missed.
- From #6492:  they cannot be provided if we swap out libev for libuv

I would vote that we remove these from the public API now and if the need arises for them again, we try to implement an interface like the per-watcher ("handle" in libuv speak) [uv_ref()](https://docs.libuv.org/en/v1.x/guide/utilities.html#event-loop-reference-count).

garlick:
I missed that this is used in `flux-start(1)` and in the python bindings in `handle.py::reactor_run()`.

The `flux-start` usage is easily worked around.  The python one, added in c1afc018d280395dbabf42e9c3e51461564af556, might require a little more thought, or possibly argue for the `uv_ref()` style interface above.

PR #6534, opened 2024-12-22
reactor: improve reactor referencing API

garlick:
This replaces the libev style `flux_reactor_active_incref()` / `_decref()` interface, in which the watcher reference count is directly manipulated by callers, with a libuv style `flux_watcher_ref()` / `_unref()` interface.  As discussed in #6510, the new interface is safer for users.

**From  `flux_watcher_stop(3)`**
```c
void flux_watcher_unref (flux_watcher_t *w);
void flux_watcher_ref (flux_watcher_t *w);
bool flux_watcher_is_referenced (flux_watcher_t *w);
```
`flux_watcher_unref()` drops the  watcher's  reference  on  the  reactor.
This  function  is  idempotent.  `flux_reactor_run()` normally runs until
 there are no more active watchers with references

`flux_watcher_ref()` restores the watcher's  reference  on  the  reactor.
This function is idempotent.

`flux_watcher_is_referenced()` returns true if the watcher is referenced.
All watchers are referenced unless updated with `flux_watcher_unref()`.

**How is it different for users?**

The new interface doesn't require knowledge of how many reactor references are taken by a given watcher.

As an example, `flux-start` embeds a service that allows it to be controlled by tests.  Since the original design of `flux-start` had it exiting once all child brokers have exited, it was convenient to arrange for the watchers associated with the server to not contribute the reactor's use count.  The API change required the following code change:
```diff
diff --git a/src/cmd/flux-start.c b/src/cmd/flux-start.c
index 4634bcb1c..de5b73c2f 100644
--- a/src/cmd/flux-start.c
+++ b/src/cmd/flux-start.c
@@ -1038,13 +1038,9 @@ void start_server_initialize (const char *rundir, bool verbose)
         log_err_exit ("could not created embedded flux-start server");
     if (flux_msg_handler_addvec (ctx.h, htab, NULL, &ctx.handlers) < 0)
         log_err_exit ("could not register service methods");
-    /* Service related watchers:
-     * - usock server listen fd
-     * - flux_t handle watcher (adds 2 active prep/check watchers)
-     */
-    int ignore_watchers = 3;
-    while (ignore_watchers-- > 0)
-        flux_reactor_active_decref (ctx.reactor);
+
+    flux_watcher_unref (flux_get_handle_watcher (ctx.h));
+    flux_watcher_unref (usock_service_listen_watcher (ctx.h));
 }

 void start_server_finalize (void)
```


garlick:
Pushed slightly improved test coverage.

src/common/libflux/watcher_wrap.c
```diff
@@ -176,97 +192,97 @@ flux_watcher_t *flux_timer_watcher_create (flux_reactor_t *r,
...
+struct pdwatcher {
     struct flux_watcher *w;
-    ev_periodic          evp;
-    flux_reschedule_f    reschedule_cb;
+    ev_periodic evw;
+    flux_reschedule_f reschedule_cb;
 };
 
-static void periodic_start (flux_watcher_t *w)
+static void pdwatcher_start (flux_watcher_t *w)
```
grondo, line 311: Just a comment, I don't think this quite rises to the level of requiring a change to whole PR:

Was there a reason `NAME` for these watchers was kept to only 2 characters? I'm finding `pdwatcher` almost more unclear than `periodic_start`. Could it have have `struct periodic_watcher` and `periodic_watcher_start`?

(Same with `tm` vs `timer`)

src/common/libflux/msg_handler.c
```diff
@@ -729,6 +729,12 @@ int flux_dispatch_requeue (flux_t *h)
     return 0;
 }
 
+flux_watcher_t *flux_get_handle_watcher (flux_t *h)
```
grondo, line 732: Commit message typo: `handle watch` :arrow_right: `handle watcher`

src/cmd/flux-start.c
```diff
@@ -1038,13 +1038,9 @@ void start_server_initialize (const char *rundir, bool verbose)
...
-    /* Service related watchers:
-     * - usock server listen fd
-     * - flux_t handle watcher (adds 2 active prep/check watchers)
-     */
-    int ignore_watchers = 3;
-    while (ignore_watchers-- > 0)
-        flux_reactor_active_decref (ctx.reactor);
+
+    flux_watcher_unref (flux_get_handle_watcher (ctx.h));
+    flux_watcher_unref (usock_service_listen_watcher (ctx.h));
```
grondo, line 1043: Oh, so much nicer!

src/common/libzmqutil/test/zwatcher.c
```diff
@@ -120,13 +120,28 @@ static void test_zmq (flux_reactor_t *reactor)
     void *zs[2];
     flux_watcher_t *r, *w;
     const char *uri = "inproc://test_zmq";
+    flux_watcher_t *tmp;
```
grondo, line 123: commit message: perhaps an extra space in `incorrect  zmqutil_watcher_create()`

doc/man3/flux_watcher_start.rst
```diff
@@ -36,6 +42,16 @@ callback.
...
+:func:`flux_watcher_unref` drops the watcher's reference on the reactor.
+This function is idempotent.  :func:`flux_reactor_run` normally runs until
+there are no more active watchers with references.
+
+:func:`flux_watcher_ref` restores the watcher's reference on the reactor.
+This function is idempotent.
+
+:func:`flux_watcher_is_referenced` returns true if the watcher is referenced.
+All watchers are referenced unless updated with :func:`flux_watcher_unref`.
+
```
grondo, line 54: Add note that not all watchers may support `flux_watcher_unref()`? Perhaps check `flux_watcher_is_referenced()` to ensure the function has had the desired effect. (or do all watchers in practice now support this call?)

grondo (approved):
This is a great improvement! Excellent!

Just a few comments inline, but nothing that should hold up the PR IMO.

src/common/libflux/watcher_wrap.c
garlick: Sorry, that actually would be better.  I'm not sure why I was stingy with the characters there.  Old habits I suppose.

doc/man3/flux_watcher_start.rst
garlick: All watchers in the public API support this call, but I didn't bother with the private ones (fbuf, sdbus).  I'll make a note to that effect.

garlick:
Re-pushed with the suggested changes.

I  was able to pretty easily rename the internal watchers to the form `NAME_watcher` (structs, callbacks) which is more consistent with the names of the public `flux_NAME_watcher_create()`.  Good suggestion to avoid the shortener names!

I also added a small cleanup in the zwatcher unit test.

grondo:
Already approved, so feel free to set MWP!

garlick:
Thanks!  Doing that!

PR #6536, opened 2024-12-30
flux-hostlist: fix `available` hostlist

grondo:
`flux hostlist` uses the `avail` resource set int the `resource_status` response, but this resource set includes offline resources and `flux hostlist` is documented to exclude those from `flux hostlist avail`.

Subtract the `offline` resource set from `avail` when reporting the hostlist in `flux resource avail`.


garlick (approved):
LGTM!

grondo:
Thanks! I've set MWP.

PR #6537, opened 2025-01-04
shell: make shell tasks available in `shell.init` callback

grondo:
The main purpose of this PR is to create the shell task list earlier in initialization so that tasks can be iterated in the `shell.init` callback. This may be convenient in some scenarios where some per-task work is being done and that work may fail. With this change, the failure would occur during `shell.init` and cleanup will be simplified since no tasks have been started yet.

Other fixes that were discovered along the way are included here.


garlick (approved):
Nice cleanup!  LGTM!

grondo:
Thanks! I've set MWP.

PR #6538, opened 2025-01-06
shell: minor fixes for shell pty plugin

grondo:
This is another standalone PR split off some other work on the job shell. This PR fixes some leaks in the shell `pty` plugin, and also makes the capture of pty output more efficient by encoding `shell.output` plugin callback args with a string `rank` member instead of integer. This allows the shell output plugin to pass the args directly along as an RFC 24 output context object, removing the need to unpack and repack a JSON object.

### DIFF
--- a/src/shell/output.c
+++ b/src/shell/output.c
@@ -726,21 +726,13 @@ static int shell_output_handler (flux_plugin_t *p,
                                  void *arg)
 {
     struct shell_output *out = arg;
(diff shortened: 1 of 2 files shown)


chu11 (approved):
LGTM

grondo:
Thanks! Restarted the fedora40 builder which hung in the cron faketime tests and I"ll set MWP.

PR #6539, opened 2025-01-07
shell: refactor output plugin and enable per-node/task output files

### LINKED ISSUE #2422: flux-shell: implement per-task output
Peeled off of #2241 

Include `{{taskid}}` mustache support.

I think we decided that we should detect the mode by looking at `attributes.system.shell.options.output.<stream>.path` rathern than via the `type` attribute, e.g.
* if no `path`, assume kvs
* if path includes `{{taskid}}` assume per-task
* else assume file

### LINKED ISSUE #4696: per-task output from flux mini commands

grondo:
This PR refactors the job shell output plugin with the intent of making per-node/task output files possible.

This PR is currently marked as a WIP because I think eventually the incremental refactor commits should be all squashed into a single "rewrite output plugin" commit, as I'm not sure the separate commits would offer that much in the git history. However, I left them here for now because it might be _slightly_ easier to review the process (though hopefully it doesn't make a review more confusing)

The refactor splits the output plugin into components now located under `src/shell/output/`. The main source of new functionality in the "task output list" abstraction found in `src/shell/output/task.[ch]`, which allows each task to separately open an output file and write data to it. Per-shell and per-task output files are supported simply by specifying a node or task-specific mustache tag in the `--output` or `--error` template. Tasks then either write directly to a file if they've opened one, or to the shell output service if shell_rank > 0, or directly to the KVS eventlogger on rank0.

The output plugin also has to handle log messages (see `shell/output/log.[ch]` and `shell.output` plugin calls (currently only made by the `pty` plugin). Log messages are written to the same destination as the local task index=0, while `shell.output` data is written to the appropriate task in the task output list.

Other high-level changes (obvious when going through the commits)
- The unused `term` (terminal) output mode is removed
- jobspec option parsing is greatly simplified and (mostly) contained in `shell/output/conf.[ch]`
- The intermediate JSON `output` array is removed (was no longer necessary) 
- reference counting is simplified
- generally a few less unnecessary pack/unpack and ioencode/iodecode calls in the common cases

A couple other potential issues:
- Though RFC 24 allows for per-rank "redirect" events, this PR opts to not implement that, since adding a redirect event per task on large jobs could thwart the point of redirecting output away from the KVS in the first place. Instead, a redirect event without any per-rank or per-task templates substituted is used, e.g.:
   ```console
   $ flux run -N4 -n 16 --output=flux-{{id}}.{{task.id}} hostname
   [0-15]: stdout redirected to flux-ƒo4c9DXvNT.{{task.id}}
   [0-15]: stderr redirected to flux-ƒo4c9DXvNT.{{task.id}}
   ```
   This seemed like a good compromise for now.
- The job shell needs to determine if an output template is "per shell" or not. It does this by checking if the template renders differently for a different rank or task. This test isn't fool-proof though, because shell plugins are allowed to extend the set of supported mustache tags, and there may be no way to determine if these tags will render differently on different tasks/shell ranks from a single shell. (For instance, `{{tmpdir}}` tag has to be treated as a special case now, though it is unlikely anyone would use that tag in their output template). 

Fixes #2422
Fixes #4696


src/shell/output/client.c
```diff
@@ -0,0 +1,169 @@
...
+        output_client_control (client, true);
+    return 0;
+error:
+    flux_future_destroy (f);
+    return -1;
+
+}
+
+/* vi: ts=4 sw=4 expandtab
+*/
```
garlick, line 169: teensy formatting error

src/shell/Makefile.am
```diff
@@ -79,6 +79,8 @@ flux_shell_SOURCES = \
 	output/filehash.c \
```
garlick, line 78: commit message:

delete two "ands" in this sentence?
> Split the KVS output functionality and into output/kvs.[ch] and
under a new `struct kvs_output` abstraction.

garlick (approved):
Excellent cleanup!  There was a mess and now there is not.

The option of per-task and per-node output files feels like a strong addition for cases where output won't scale using the other methods.

Great job!  I only had minor comments.

garlick:
BTW I would lean towards leaving the commits broken out, as they're nicely documented this way, but I don't have too strong of an opinion if you feel otherwise.

src/shell/output/filehash.c
```diff
@@ -0,0 +1,163 @@
...
+static void file_entry_destroy (struct file_entry *fp)
+{
+    if (fp) {
+        int saved_errno = errno;
+        close (fp->fd);
+        free (fp->path);
+        free (fp);
+        errno = saved_errno;
+    }
+}
```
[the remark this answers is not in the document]
grondo: Good catch!

grondo:
Thanks! I'm fine leaving the commits as is. I pushed the suggested changes and will remove WIP. There are already tests and documentation, so I'll go ahead and set MWP as well.

PR #6541, opened 2025-01-07
add `resource.rediscover` config key to force rediscovery of subinstance resources

### LINKED ISSUE #6418: need a way to force a Flux subinstance to rediscover resources
Some modern GPUs have the ability to present one GPU as multiple GPUs, such as AMD [compute partition mode](https://rocm.docs.amd.com/projects/amdsmi/en/docs-6.0.0/doxygen/docBin/html/amdsmi_8h.html#af5c57809078d4345b08b59b4162cd6ba).

In order to fully take advantage of this feature, Flux should be able to discover that a single GPU at the system level is now available as multiple (2, 3 or 4) GPUs in the allocation.

The easiest way to accomplish this would be to force the subinstance to rediscover resources with hwloc. This would discover the current state of the job topology instead of reusing the R from the parent. The cached hwloc topology fetched from the parent should also be discarded (or fetching it should be avoided.)

A simple way to force this rediscovery should be provided via  the `flux alloc` and `flux batch` command line. Perhaps also via some method which could be forced by a prolog (the GPU mode would need to be set by a privileged prolog script, it would be nice if the prolog could enable this mode automatically instead of requiring the user to do it manually)

grondo:
This PR adds an optional `rediscover` key to  the `[resource]` config table which, when true, forces rediscovery of local resources via hwloc, rather than using the job's R and enclosing instance hwloc XML.

This could be used in combination with a prolog option to change the GPU compute partition mode so the child instance can see the newly reconfigured GPUs.

Fixes #6418.

### DIFF
--- a/src/modules/resource/resource.c
+++ b/src/modules/resource/resource.c
@@ -57,6 +57,10 @@
  * no-update-watch = false
  *   For testing purposes, simulate missing job-info.update-watch service
  *   in parent instance by sending to an invalid service name.
+ *
+ * rediscover = false
+ *   Force rediscovery of local resources via hwloc. Do not fetch R or hwloc
+ *   XML from the enclosing instance.
  */
 static int parse_config (struct resource_ctx *ctx,
                          const flux_conf_t *conf,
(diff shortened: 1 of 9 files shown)


src/modules/resource/resource.c
```diff
@@ -60,11 +60,7 @@
  */
 static int parse_config (struct resource_ctx *ctx,
                          const flux_conf_t *conf,
-                         const char **excludep,
-                         json_t **R,
-                         bool *noverifyp,
-                         bool *norestrictp,
-                         bool *no_update_watchp,
+                         struct resource_config *rconfig,
                          flux_error_t *errp)
```
garlick, line 68: Oh, thank you, that needed to happen.

garlick (approved):
LGTM!

grondo:
Thanks! I'll set MWP.

vsoch:
I missed this (I'm reading features from the releases now)! Is this something that could be used to dynamically generate resources for a broker / worker set?

Issue #6315, opened 2024-09-26
optionally withhold some cores on each node for system tasks


garlick:
Problem: in some situations, applications run faster when some number of cores are reserved for system services.  In other situations, applications perform best when all the cores are available to the application.

Apparently LSF on sierra will by default only allocate 40 of 44 cores per node to a job.  There is an option users can add to their job request  to indicate that 42 or 44 be allocated.

wihobbs:
Possibly related to #4372? (although may be outdated discussion)

grondo:
Possibly related #5240

grondo:
See comment in #5240 for a prototype (found [here](https://github.com/grondo/flux-core/tree/reserve-cores))

wihobbs:
@eleon has a few thoughts here shared today with @trws and me. Tagging him so he can share more.

eleon:
Thanks for the pointer, @wihobbs 

The issue I see is that on El Cap systems, Flux is giving users cores that are "reserved" for system tasks (more details below). 

My suggestion for working with systems that provide noise mitigation via core specialization is as follows: 
1. If users ask for a full node (`-x`), then Flux provides all of the cores, including system cores. 
2. If users ask for a number of cores (`-c`), then Flux provides the requested number of cores, but draws them from a pool that excludes the system cores. 

Right now, when a user requests, say, 8 cores, one of the cores is a system core. This causes an issue for mpibind, because mpibind will only schedule work on user cores (by default). This means that the job will only run on 7 cores, potentially oversubscribing the resources: 
```
leon@tuolumne1071:~$ flux run -N1 -n8 -c1 -o mpibind=off -o cpu-affinity=per-task sh -c 'taskset -pc $$' 
pid 75901's current affinity list: 89,185
pid 75900's current affinity list: 88,184
pid 75902's current affinity list: 90,186
pid 75903's current affinity list: 91,187
pid 75904's current affinity list: 92,188
pid 75905's current affinity list: 93,189
pid 75907's current affinity list: 95,191
pid 75906's current affinity list: 94,190
``` 

```
leon@tuolumne1071:~$ flux run -N1 -n8 -c1 sh -c 'taskset -pc $$' 
pid 75913's current affinity list: 89
pid 75914's current affinity list: 185
pid 75915's current affinity list: 90,186
pid 75916's current affinity list: 91,187
pid 75917's current affinity list: 92,188
pid 75918's current affinity list: 93,189
pid 75919's current affinity list: 94,190
[...]

Issue #5240, opened 2023-06-08
resource: support fine grained resource exclusion


garlick:
Problem: it may be useful in some environments to exclude a subset of available cores on rank 0 from scheduling, but resource exclusion currently only works at node granularity.

From [flux-config-resource(5)](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man5/flux-config-resource.html)
> **exclude**
>
>    (optional) A string value that defines one or more nodes to withhold from scheduling, either in RFC 22 idset form, or in RFC 29 hostlist form. This value may be changed on a live system by reloading the configuration on the rank 0 broker. A newly excluded node will appear as "down" to the scheduler, but will still be used to determine satisfiability of job requests until the instance is restarted.
>
>    If a drained node is subsequently excluded, the drain state of the node is cleared since nodes cannot be both excluded and drained.

[RFC 28](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_28.html) defines how the scheduler is informed of available resources.  The resource acquire response provides  _R_ with excluded execution targets subtracted.  It seems like we could provide a way to configure and then subtract an arbitrary resource set from _R_ before presenting it to the scheduler.

grondo:
I was thinking about this for a bit, and one other issue is that the `resource.status` RPC reports `online`, `offline`, and `exclude` sets as idsets instead of a full resource set. We may want to change (at least the `exclude` set), if we want to allow exclusion of individual resources.

jameshcorbett:
This came up in the quarterly today. It seems that admins may want to exclude some cores from scheduling even on compute nodes, so those cores can be dedicated to OS tasks, kind of like is done on Sierra.

grondo:
In this case, rather than exclude 1 core on every compute node, it might be possible to just configure compute nodes without core 0, e.g. via `resource.config`:

```toml
[[resource.config]]
hosts = "host[1-100]"
cores = "1-47"
```

Maybe an experiment is warranted to see if there's any pitfalls, as I haven't actually tried this approach.

grondo:
[...]

PR #6547, opened 2025-01-10
WIP: support `resource.reserve`  configuration parameter

grondo:
This WIP PR addresses #6315 and #5240 by adding a new config key `resource.exclude`, which allows specification of a set of cores to hold back from the scheduler along with optional ranks on which to reserve them. Cores are specified in the form:
```
cores[@ranks] [cores[@ranks]]...
```
Where `cores` and the optional `ranks` are idsets. If ranks is not specified, then the same cores are reserved on all ranks. To reserve different sets of cores on different ranks, multiple specifications may be used separated by whitespace.
For example, `0` would withold core 0 on all ranks, `0-1@0 0` would reserve `0-1` on rank 0 and `0` on all other ranks, etc.

Marking this as a WIP because it is just one possible approach. We could also easily extend the `resource.exclude` config key to allow a similar syntax. There are possibly some drawbacks here:
- `exclude` is encoded as an idset in the `resource.status` response - we'd have to always condense the excluded resource set into an idset for backwards compatibility.
- `exclude` allows specification of a hostlist, which doesn't make as much sense for resource reservation (though it could be easily handled.)
- The common case for `resource.exclude` is perhaps different than `resource.reserve` -- e.g. reserving the same set of cores across all nodes of a job would be slightly more awkward if only `resource.exclude` syntax was supported (since specification of ranks would be obligatory)

However, it does seem awkward to have two config keys that do similar things.

The other question here is how to present the reserved resources in either `flux resource status` (which currently does not include sub-rank resources in default output anyway) or `flux resource list`.

### DIFF
--- a/doc/man5/flux-config-resource.rst
+++ b/doc/man5/flux-config-resource.rst
@@ -79,6 +79,21 @@ rediscover
    (optional) If true, force rediscovery of resources using HWLOC, rather
    then using the R and HWLOC XML from the enclosing instance.
 
+reserve
+   (optional) A string value that defines cores to reserve for the OS
+   and the broker ranks on which to reserve them. The argument is specified
+   as::
+
+    cores[@ranks] [cores[@ranks]]...
+
+   where cores is an RFC 22 idset specifying the cores to reserve, and
+   the optional ranks is an RFC 22 idset specifying the ranks on which to
+   reserve them. If ``@ranks`` is not supplied, then cores will be reserved
+   on all ranks. Multiple instances of ``cores[@ranks]`` may be specified
+   (separated by whitespace) to reserve a different set of cores on different
+   ranks. For example, ``0-3@0 0`` would reserve cores 0-3 on rank 0 and
+   core 0 on all other ranks.
+
 Note that updates to the resource table are ignored until the next Flux
 restart.
 
--- a/src/cmd/flux-resource.py
+++ b/src/cmd/flux-resource.py
@@ -74,7 +74,7 @@ class FluxResourceConfig(UtilConfig):
             "description": "Format including resource list details",
             "format": (
                 "{state:>10} ?+:{queue:<5} ?:{propertiesx:<10.10+} {nnodes:>6} "
-                "+:{ncores:>6} ?:+{ngpus:>5} {rlist}"
+                "+:{ncores:>6} ?+:{ngpus:>5} {rlist}"
             ),
         },
     }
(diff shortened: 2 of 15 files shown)


garlick:
This seems workable to me.

I can't really think of a case where a system instance would want to reserve different cores on different node types and run into awkwardness specifying those as ranks.  If that ever does come up then hostlist support might be nice.

grondo:
It would be good to get feedback from @ryanday36 if this is the correct approach for what's needed in the near term. I'm not sure if the requirements are to allow users to reserve cores for the system in their batch jobs, while the default is not to do so, or if the need is to reserve cores in the system instance, then give users an option to opt-out.

I'd hate to add a new config key (and whole new way to select resources) if it isn't going to end up useful.

ryanday36:
My understanding for what we want with the "OS" or "system" cores is to have no user tasks run on the system cores unless the user explicitly asks for them. So, this `resource.reserve` approach would work as long as there is also some sort of `--all-resources` flag to the `flux run` type commands that would allow the users to specify that they want to run on reserved resources.

grondo:
> So, this resource.reserve approach would work as long as there is also some sort of --all-resources flag to the flux run type commands that would allow the users to specify that they want to run on reserved resources.

In this case the resources are excluded from the scheduler so there is no way to ask for them, except perhaps by launching a new instance (i.e. via either `flux batch` or `flux alloc`), setting the `resource.rediscover` option for that instance, then using `flux run` or `flux submit` in the new instance.

@garlick's idea of a partition might be slightly more suitable, except that, AIUI, the proposed partitions are not allowed to overlap so you would not be able to run a job across all partitions. That may be easily resolved however.

In any event, I think this requirement implies that the scheduler needs to know about reserved resources specifically, so withholding them like `exclude` won't work.

garlick:
> except perhaps by launching a new instance (i.e. via either flux batch or flux alloc), setting the resource.rediscover option for that instance, then using flux run or flux submit in the new instance.

Would that be all that bad?   I think we decided the partition idea was going to take significantly more effort.

ryanday36:
> In this case the resources are excluded from the scheduler so there is no way to ask for them, except perhaps by launching a new instance (i.e. via either `flux batch` or `flux alloc`), setting the `resource.rediscover` option for that instance, then using `flux run` or `flux submit` in the new instance.

Having users get a new instance with `resource.rediscover` would probably work. I can't actually think of a use case where a users would need some `flux run` to see the cores and others not see the cores, and if one did come up, it would probably be a pretty savvy user. Those users could get their instance with `resource.rediscover` and then manage things themselves.

grondo:
Well, perhaps this is ready for a review then. Is the current interface ok? Happy to take suggestions on improvements.

garlick (commented):
Is it feasible to add a high level test in sharness for this?

PR #6554, opened 2025-01-14
prevent log message truncation

### LINKED ISSUE #6545: important log messages may be truncated
Not sure where the truncation is occurring, but an important log message resulting from the following `flux_log()` call was truncated in `flux dmesg` :
```c
            flux_log (hk->ctx->h,
                      LOG_ERR,
                      "housekeeping: %s (rank %s) from %s will be terminated"
                      " because %s",
                      hosts ? hosts : "?",
                      ranks ? ranks : "?",
                      idf58 (a->id),
                      error.text);
```

In this case >4000 nodes were affected, and the very long nodelist in `hosts` caused the log message to get truncated, suppressing the actual important bits of the message that sysadmins were searching for (e.g. the jobid).

grondo:
This fixes #6545 by falling back to a malloc'd buffer if truncation is detected when handling log messages.

Since the internal fixed buffer size is no longer relevant, the definition of `FLUX_MAX_LOGBUF` is also removed from the public interface. However, I'm fine dropping that commit if it goes too far.

Fixes #6545.

### DIFF
--- a/src/broker/shutdown.c
+++ b/src/broker/shutdown.c
@@ -21,6 +21,7 @@
 #include <flux/core.h>
 
 #include "src/common/libutil/stdlog.h"
+#include "src/common/libutil/errno_safe.h"
 
 #include "shutdown.h"
 #include "state_machine.h"
@@ -66,30 +67,27 @@ static int forward_logbuf (flux_t *h,
     struct stdlog_header hdr;
     const char *txt;
     size_t txtlen;
-    char buf[FLUX_MAX_LOGBUF];
+    int len = strlen (stdlog);
+    char *buf;
     int loglevel;
+    int rc;
 
     if (flux_msg_unpack (request, "{s:i}", "loglevel", &loglevel) < 0)
         loglevel = LOG_ERR;
 
-    if (stdlog_decode (stdlog,
-                       strlen (stdlog),
-                       &hdr,
-                       NULL,
-                       NULL,
-                       &txt,
-                       &txtlen) < 0
+    if (stdlog_decode (stdlog, len, &hdr, NULL, NULL, &txt, &txtlen) < 0
         || STDLOG_SEVERITY (hdr.pri) > loglevel
-        || snprintf (buf,
-                     sizeof (buf),
+        || asprintf (&buf,
                      "%s.%s[%lu]: %.*s\n",
                      hdr.appname,
                      stdlog_severity_to_string (STDLOG_SEVERITY (hdr.pri)),
                      strtoul (hdr.hostname, NULL, 10),
                      (int)txtlen,
-                     txt) >= sizeof (buf))
+                     txt) < 0)
         return 0;
-    return flux_respond_pack (h, request, "{s:s}", "log", buf);
+    rc = flux_respond_pack (h, request, "{s:s}", "log", buf);
+    ERRNO_SAFE_WRAP (free, buf);
+    return rc;
 }
 
 static void dmesg_continuation (flux_future_t *f, void *arg)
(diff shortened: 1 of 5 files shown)


src/common/libflux/flog.c
```diff
@@ -111,6 +111,31 @@ static int log_rpc (flux_t *h, const char *buf, size_t len)
...
+    if (vsnprintf (result, sizeof (buf), fmt, ap) >= sizeof (buf)) {
+        /* Message truncation occurred. Fall back to malloc. If that
+         * fails, just print the truncated message:
+         */
+        if (vasprintf (&mbuf, fmt, ap_copy) == 0)
+            result = mbuf;
+    }
+    rc = fprintf (stderr, "%s: %s\n", lstr, result);
+    free (mbuf);
+    return rc < 0 ? -1 : 0;
```
garlick, line 138: va_copy(3) says
> Each  invocation of va_copy() must be matched by a corresponding invocation
       of va_end() in  the  same  function. 

Add one after the conditional block?

src/common/libflux/flog.c
```diff
@@ -149,32 +170,57 @@ int flux_vlog (flux_t *h, int level, const char *fmt, va_list ap)
     hdr.appname = ctx->appname;
     hdr.procid = ctx->procid;
 
+    /* Copy va_list ap in case it reused on truncation below:
+     */
+    va_copy (ap_copy, ap);
+
```
garlick, line 178: same comment here

src/common/libflux/flog.c
```diff
@@ -149,32 +170,57 @@ int flux_vlog (flux_t *h, int level, const char *fmt, va_list ap)
...
                          fmt,
                          ap);
+    result = ctx->buf;
     len = n >= sizeof (ctx->buf) ? sizeof (ctx->buf) : n;
+
+    /* Unlikely: On overflow, try again with a buffer allocated on the heap.
+     * If this fails for any reason, fall back to truncated output.
+     */
+    if (n >= sizeof (ctx->buf)) {
+        int size = n++;
```
garlick, line 192: Side effect on `n` seems not to have a purpose?

src/broker/shutdown.c
```diff
@@ -66,19 +66,14 @@ static int forward_logbuf (flux_t *h,
     struct stdlog_header hdr;
     const char *txt;
     size_t txtlen;
-    char buf[FLUX_MAX_LOGBUF];
+    int len = strlen (stdlog);
+    char buf[len + 1];
```
garlick, line 70: Since this is unbounded, are we concerned about blowing the stack if the system conjurs a really big log?

garlick (approved):

garlick (approved):
LGTM!  Nothing major in my comments so I'll approve and let you address whatever you think is worthwhile.

src/common/libflux/flog.c
grondo: D'oh! Thanks for catching this one!

src/broker/shutdown.c
grondo: You're correct, that case should be handled. I'm not sure if it would just be easier to just use the heap here, since this is a limited case. Otherwise, the VLA size could be replaced with a macro that takes the min of `len + 1` and some sane maximum (4K?) to avoid blowing up the stack?

garlick: probably heap?  Otherwise it's the sane max just becomes the new FLUX_MAX_LOGBUF (albeit just in this case!)

grondo: Ok, fixed that and addressed the other comments. Thanks!

grondo:
Thanks! Setting MWP now.

Issue #6125, opened 2024-07-19
kvs: investigate and prevent potential denial of service


chu11:
Related to #6124, there may be some denial or service possibilities in the KVS.

The first one that comes to mind is that there is no limit on the size of a kvstxn that can be sent to the KVS, like it could be a billion entries long.  

(Mildly related, #1202, #1206, #1207, where some data could become crazy huge)

garlick:
It might be interesting to survey blob sizes currently in the content cache on el cap.  We could just look at one of the dump files I guess and make a histogram.

The original design for content was to have a 1MB limit on blobs, FWIW.

chu11:
> The first one that comes to mind is that there is no limit on the size of a kvstxn that can be sent to the KVS, like it could be a billion entries long.

This is the most obvious DoS that could occur in the KVS and is probably the first one to tackle.

The simplest solution would be to have some configurable max.

That said, we actually don't know what the min, max, mean, stddev, etc. of KVS ops is.  So perhaps getting some stats out of the KVS on that would be the first thing to do.  Should be easy to add to the KVS stats output using `tstat`.

chu11:
> The first one that comes to mind is that there is no limit on the size of a kvstxn that can be sent to the KVS, like it could be a billion entries long.

It occurred to me today that there are really two issues:

- maximum number of ops in a single transaction
- total number of ops a user has submitted

the former has some "quality of service" feel to it, in that other transactions can sneak in between multiple transactions that have MAX operations.  This is certainly issue number one.

the latter is like if the user submits MAX ops over and over again.  We'd want to "rate limit" total ops.  This might be issue two.

Edit: third one ... a path can be "infinite" in length.

garlick:
Can you refresh my memory?  If a user commits several large blobs in a transaction, is this correct
- the blobs are base64 encoded and included in the commit request
- the commit request is handled by the _local_ kvs rank, which commits the blobs to the content store and alters the commit request to only contain content references, not the data, when it sends it upstream
- so the rank 0 kvs only deals with blobrefs not blobs (unless it's handling commits started on rank 0)


My biggest worry about the KVS is the size of messages because that can cause head of line blocking at the broker level.  I'm not _as_ worried about it on rank > 0 compared to rank 0 though.

And I guess it's obvious but worth repeating because I don't see it mentioned here - it's guests in the system instance (or owner services operating on behalf of a guest) that can really do the damage.

chu11:
[...]

PR #6556, opened 2025-01-16
kvs: add transaction stats

chu11:
Problem: In the future we would like to put a cap on the maximum number
of operations in a transaction.  However, we do not have any stat tracking
of operations in a transaction.

Add transaction stats for commits and fences in the kvs module.

---

This is really a "prep" for #6125.  I decided to split out the stats for commits & fences, since they are quite different.  For example, if every process in an job did a fenced KVS transaction with a single op, this scales up with the size of cluster.  So we want to count that differently than each op individually.

Note that I could have done "stats per namespace", but that wasn't really the information that we want.  We mostly want to know what is the average and what is the max.  So I did it "globally" vs per namespace.

### DIFF
--- a/src/modules/kvs/kvs.c
+++ b/src/modules/kvs/kvs.c
@@ -2448,7 +2448,7 @@ static void stats_get_cb (flux_t *h,
                            "{ s:O s:O s:{s:O s:O} s:i }",
                            "cache", cstats,
                            "namespace", nsstats,
-                           "transactions",
+                           "transaction-opcount",
                              "commit", txncstats,
                              "fence", txnfstats,
                            "pending_requests", zhashx_size (ctx->requests)) < 0)
--- a/src/modules/kvs/kvs.c
+++ b/src/modules/kvs/kvs.c
@@ -2371,12 +2371,12 @@ static void stats_get_cb (flux_t *h,
             goto error;
     }
 
-    if (!(tstats = json_pack ("{ s:i s:f s:f s:f s:f }",
+    if (!(tstats = json_pack ("{ s:i s:I s:f s:f s:I }",
                               "count", tstat_count (&ts),
-                              "min", tstat_min (&ts)*scale,
+                              "min", (json_int_t)tstat_min (&ts)*scale,
                               "mean", tstat_mean (&ts)*scale,
                               "stddev", tstat_stddev (&ts)*scale,
-                              "max", tstat_max (&ts)*scale)))
+                              "max", (json_int_t)tstat_max (&ts)*scale)))
         goto nomem;
 
     if (!(cstats = json_pack ("{ s:f s:O s:i s:i s:i }",
--- a/src/modules/kvs/kvs.c
+++ b/src/modules/kvs/kvs.c
@@ -78,6 +78,8 @@ struct kvs_ctx {
     char *hash_name;
     unsigned int seq;           /* for commit transactions */
     kvs_checkpoint_t *kcp;
+    tstat_t txn_commit_stats;
+    tstat_t txn_fence_stats;
     zhashx_t *requests;         /* track unfinished requests */
     struct list_head work_queue;
 };
@@ -1721,6 +1723,8 @@ static void relaycommit_request_cb (flux_t *h,
         goto error;
     }
 
+    tstat_push (&ctx->txn_commit_stats, (double)json_array_size (ops));
+
     /* N.B. no request tracking for relay.  The relay does not get a
      * response, only the original via finalize_transaction_bynames().
      */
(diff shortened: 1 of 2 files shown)


chu11:
oops, looks like our CI doesn't have `bc` so everything failed :-)  Re-worked the tests to use `jq` instead (actually is much cleaner now).

src/modules/kvs/kvs.c
```diff
@@ -2387,6 +2401,23 @@ static void stats_get_cb (flux_t *h,
...
+                                   (int)tstat_max (&ctx->txn_commit_stats))))
+        goto nomem;
+
+    if (!(txnfstats = json_pack ("{ s:i s:i s:f s:f s:i }",
+                                 "count", tstat_count (&ctx->txn_fence_stats),
+                                 "min", (int)tstat_min (&ctx->txn_fence_stats),
+                                 "mean", tstat_mean (&ctx->txn_fence_stats),
+                                 "stddev", tstat_stddev (&ctx->txn_fence_stats),
+                                 "max", (int)tstat_max (&ctx->txn_fence_stats))))
+        goto nomem;
```
garlick, line 2419: This is the same as the `tstats` object except `max` is packed as an intetger.

Would adding a `add_tstat_to_object()` function be useful here to get all three in the same format, reduce code duplication, and avoid the need for a temporary variable per object?

src/modules/kvs/kvs.c
```diff
@@ -2415,13 +2446,18 @@ static void stats_get_cb (flux_t *h,
 
     if (flux_respond_pack (h,
                            msg,
-                           "{ s:O s:O s:i }",
+                           "{ s:O s:O s:{s:O s:O} s:i }",
                            "cache", cstats,
                            "namespace", nsstats,
+                           "transactions",
```
garlick, line 2452: Suggestion: make this key more descriptive like `transaction-opcount`?

garlick (approved):
LGTM once the test failures are worked out.

I did just have a couple minor suggestions.

Also a passing comment that we should fix the stats key names that have spaces and `#` in them at some point.

src/modules/kvs/kvs.c
chu11: Sounds like a good idea.  However, I do dislike that the min and max for op-count would be a double if we created a common function for this.  Although it's not the end of the world if that were the case.  

For the cache stats, I suppose min and max should also always be integers (or big Ints) too.  However, that would be changing the output format.  Hmmm, lemme ponder this detail for a bit.

garlick: Changing the output format here doesn't seem so bad.

chu11:
ok, re-pushed per comments above.  The big change is that the cache object stats now use ints for the output of min and max.

That allowed me to create a convenience function for the cache object stats & the new transaction stats to generation objects for tstat.

I elected to go with the convenience function instead of a new `tstats` function b/c of A) the casting of the min/max to ints and B) the scaling option to scale the stats (used by cache object stats).  It just seemed like too much special stuff just for the kvs.

src/modules/kvs/kvs.c
```diff
@@ -2431,6 +2466,8 @@ static void stats_get_cb (flux_t *h,
         flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
     json_decref (tstats);
     json_decref (cstats);
+    json_decref (txncstats);
+    json_decref (txnfstats);
     json_decref (nsstats);
```
garlick, line 2471: If the convenience function added a key to an existing object rather than return an object, it would ave on the temp variables.

chu11: I had thought about that, but that's only for the first case w/ the object stats here.

```
    if (!(cstats = json_pack ("{ s:f s:O s:i s:i s:i }",                                                                                    
                              "obj size total (MiB)", (double)size/1048576,                                                                 
                              "obj size (KiB)", tstats,                                                                                     
                              "#obj dirty", dirty,                                                                                          
                              "#obj incomplete", incomplete,                                                                                
                              "#faults", ctx->faults)))                                                                                     
        goto nomem;   
```

Most of the "object' construction is done with a rpc response at the end:

```
    if (flux_respond_pack (h,                                                                                                               
                           msg,                                                                                                             
                           "{ s:O s:O s:{s:O s:O} s:i }",                                                                                   
                           "cache", cstats,                                                                                                 
                           "namespace", nsstats,                                                                                            
                           "transaction-opcount",                                                                                           
                             "commit", txncstats,                                                                                           
                             "fence", txnfstats,                                                                                            
                           "pending_requests", zhashx_size (ctx->requests)) < 0)                                                            
        flux_log_error (h, "%s: flux_respond_pack", __FUNCTION__);   
```

we could create a "transaction-ops" object with `json_object()` and pass it around.  It seemed like a "meh" win to me ... feel strongly about it?

garlick: Nope :-)

chu11:
thanks, setting MWP

PR #6557, opened 2025-01-17
add `flux --root` option

grondo:
Since `flux --parent [--parent] COMMAND ARGS...` is supported to connect to parent (grandparent, etc.), it seems neglectful that a method to connect to the top-level or _root_ instance is not also provided. Instead, users have to supply the `--parent` option at least `instance-level` times: an admittedly awkward interface.

This PR adds a `--root` option to `flux(1)` that behaves exactly as if `--parent` had been specified `instance-level` times. If `instance-level` is `0`, this option has no effect (same as `--parent`).


chu11 (approved):
Seems like a good option to have, LGTM!

grondo:
Thanks!

PR #6558, opened 2025-01-17
add `FLUX_ENCLOSING_ID` to initial program environment for instances with a `jobid` broker attribute

grondo:
This PR adds a `FLUX_ENCLOSING_ID` environment variable as an alternative to the broker `jobid` attribute. If an instance does not have a `jobid` attribute, then `FLUX_ENCLOSING_ID` is unset.

This environment variable can be used to refer to the jobid of the enclosing batch or alloc instance from the initial program or a job within that instance (since it will be inherited by jobs).

### DIFF
--- a/doc/man7/flux-environment.rst
+++ b/doc/man7/flux-environment.rst
@@ -25,6 +25,22 @@ The following are set in the environment of each task spawned by
 
       NUMERIC_JOB_ID=$(flux job id $FLUX_JOB_ID)
 
+.. envvar:: FLUX_ENCLOSING_ID
+
+   The jobid of the enclosing Flux instance, if it has one. The enclosing
+   Flux instance is the one that ran :envvar:`FLUX_JOB_ID`. Depending on
+   how the enclosing Flux instance was started, it may or may not have
+   a jobid. If it was not launched by Flux, :envvar:`FLUX_ENCLOSING_ID`
+   is not set.
+
+   Example 1: A batch job that runs one MPI job is submitted to a Flux system
+   instance. In the environment of the MPI job, :envvar:`FLUX_ENCLOSING_ID`
+   refers to the batch jobid in the system instance.
+
+   Example 2: An MPI job is submitted directly to a Flux system
+   instance. Since the Flux system instance was not launched by Flux,
+   :envvar:`FLUX_ENCLOSING_ID` is not set in the environment of the MPI job.
+
 .. envvar:: FLUX_JOB_SIZE
 
    The number of tasks in the current job.
@@ -170,8 +186,9 @@ environment in other workload managers, for example:
    BATCH_NCORES=$(flux resource list -n -o {ncores})
    BATCH_NGPUS=$(flux resource list -n -o {ngpus})
    BATCH_HOSTLIST=$(flux getattr hostlist)
-   BATCH_JOBID=$(flux getattr jobid)
 
+Additionally, :envvar:`FLUX_ENCLOSING_ID` is set to the jobid of the
+enclosing instance, if it has one.
 
 PMI CLIENT
 ==========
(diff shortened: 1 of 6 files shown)


doc/man7/flux-environment.rst
```diff
@@ -25,6 +25,15 @@ The following are set in the environment of each task spawned by
...
 
+.. envvar:: FLUX_ENCLOSING_ID
+
+   The jobid of the closest enclosing Flux instance if it was run as a
+   job. This is set as an alternative to :envvar:`FLUX_JOB_ID` because
+   it applies only to the parent instance, not the current instance.
+
+   See :ref:`INITIAL PROGRAM ENVIRONMENT <initial_program_environment>` below
+   for more details.
+
```
garlick, line 36: It might be helpful to spell this out a bit more for casual Flux user.  I took a stab below but feel free to ignore or edit.

> The job ID of the enclosing Flux instance, if it has one.  The enclosing Flux instance is the one that ran FLUX_JOB_ID.   Depending on how the enclosing Flux instance was started, it may or may not have a job ID.   If it was not launched by Flux, FLUX_ENCLOSING_ID is not set.
> 
> Example 1: a batch job that runs one MPI job is submitted to a Flux system instance.  In the environment of the MPI job, FLUX_ENCLOSING_ID refers to the batch job ID in the system instance.
>
> Example 2: an MPI job is submitted directly to a Flux system instance.  Since the Flux system instance was not launched by Flux, FLUX_ENCLOSING_ID is not set in the environment of the MPI job.

garlick (approved):
LGTM, just a suggestion on expanding the description a bit.

doc/man7/flux-environment.rst
chu11: Jim beat me to it, I agree fleshing this area out a bit more.

doc/man7/flux-environment.rst
```diff
@@ -170,8 +179,10 @@ environment in other workload managers, for example:
    BATCH_NCORES=$(flux resource list -n -o {ncores})
    BATCH_NGPUS=$(flux resource list -n -o {ngpus})
    BATCH_HOSTLIST=$(flux getattr hostlist)
-   BATCH_JOBID=$(flux getattr jobid)
 
+Additionally, :envvar:`FLUX_ENCLOSING_ID` is set by the broker if it was
+launched as part of a Flux job, and contains the jobid of the instance
+in F58 form.
```
chu11, line 185: maybe just "launched as a Flux job", not sure "part of" is necessary

grondo: Well, a single broker is not "a Flux job" unless the job has only one node.. but either way gets the point across if you feel strongly about it.

doc/man7/flux-environment.rst
grondo: Good suggestions. I'll take @garlick's text as is besides fixing a few nits (the rest of the document uses "jobid" not "job ID", and style guide says the word after a colon should be capitalized if it is the beginning of a sentence.)

grondo: Ok, pushed the update. I also dropped the reference to the INITIAL PROGRAM section as now it seemed superfluous since most if the detail in the JOB ENVIRONMENT section instead.

doc/man7/flux-environment.rst
chu11: ohhh i see your context on it now.  Perhaps just "is set", no need to say "set by the broker"?  Dunno if the latter is a detail the user needs to know.

grondo: Good point, that will probably make it clearer to the intended audience.

grondo: Simplified to duplicate @garlick's text from above:
> Additionally, :envvar:`FLUX_ENCLOSING_ID` is set to the jobid of the enclosing instance, if it has one.

How's that?

chu11: yup i think that works.  I guess you could add back the `F58 form`, but I don' think that matters.

grondo:
Thanks! Will set MWP now.

PR #6561, opened 2025-01-17
broker: ensure `parent-uri` and `parent-kvs-namespace` are only set for jobs

### LINKED ISSUE #6560: `parent-uri` attribute set with `flux run flux start --test-size=`
The detection of whether the current instance has a parent in order to set the `parent-uri` attribute currently relies on whether `FLUX_JOB_ID` is set in the environment. This doesn't always work because `FLUX_JOB_ID` can leak from the environment as is the case for `flux run flux start -s 4` for example. (This is exercised by the inception tests)

Detection of instance as a job should instead rely on the `jobid` attribute, which is only set by a PMI boot.

grondo:
This PR fixes #6560 by delaying initialization of the mentioned attributes until after bootstrap. Then the `jobid` attribute instead of `FLUX_JOB_ID` is used since this more correctly indicates if the current instance is a job or not.

Fixes #6560

### DIFF
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -112,6 +112,8 @@ static int handle_event (broker_ctx_t *ctx, const flux_msg_t *msg);
 
 static void init_attrs (attr_t *attrs, pid_t pid, struct flux_msg_cred *cred);
 
+static void init_attrs_post_boot (attr_t *attrs);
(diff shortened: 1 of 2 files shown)


garlick (approved):
LGTM!

PR #6562, opened 2025-01-17
python: support relative and absolute path-like targets in jobid URI resolver

grondo:
This PR adds support for path-like expressions in the `jobid` URI resolver. Notably, `/`  now refers to the root instance, `..` refers to the parent of the current instance, and for completeness `.` indicates the current instance. These special targets can be combined with a slash as expected, e.g. `../..` refers to the parent of the current instance's parent, and `/f1234` would refer directly to job `f1234` in the root instance, and so on.

This gives users of `flux-uri(1)`, the `FluxURIResolver` class, and the internal `uri_resolve()` function (`flux proxy`, `flux top` and `flux shutdown`) an easy way to refer to the root instance or parent instances from within a batch/alloc job.

I've made this PR a WIP because other folks may not agree with the approach. 

One other idea I had was to directly support `/` and `..` in `flux_open(3)`, but that kind of breaks the abstraction that `flux_open()` only takes a native URI. The approach in this PR does have the drawback that it only helps shell and Python users, not necessarily users of the C API.

So just throwing this out there for consideration.

### DIFF
--- a/doc/man1/flux-uri.rst
+++ b/doc/man1/flux-uri.rst
@@ -89,14 +89,21 @@ URI SCHEMES
 
 The following URI schemes are included by default:
 
-jobid:ID[/ID...]
+jobid:PATH
    This scheme attempts to get the URI for a Flux instance running as a
    job in the current enclosing instance. This is the assumed scheme if no
    ``scheme:`` is provided in *TARGET* passed to :program:`flux uri`, so the
-   ``jobid:`` prefix is optional. A hierarchy of Flux jobids is supported,
-   so ``f1234/f3456`` will resolve the URI for job ``f3456`` running in
-   job ``f1234`` in the current instance. This scheme will raise an error
-   if the target job is not running.
+   ``jobid:`` prefix is optional. *PATH* is a hierarchical path expression
+   that may contain an optional leading slash (``/``) (which references
+   the top-level, root instance explicitly), followed by zero or more job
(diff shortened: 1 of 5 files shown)


garlick:
Neat!

> One other idea I had was to directly support / and .. in flux_open(3), but that kind of breaks the abstraction that flux_open() only takes a native URI. The approach in this PR does have the drawback that it only helps shell and Python users, not necessarily users of the C API.

Heh, I kind of like that idea.  It's not the fully cooked resolver and the user is specifying strings that would not be valid URIs so it's pretty clear what they want.  We can just say it takes a native URI or those special strings.

grondo:
> Heh, I kind of like that idea. It's not the fully cooked resolver and the user is specifying strings that would not be valid URIs so it's pretty clear what they want. We can just say it takes a native URI or those special strings.

Ok, maybe I'll try to prototype something along those lines.

grondo:
Removed WIP here. This will probably be nice to have paired with the ancestor paths proposed in #6573

garlick (approved):
For some reason this brings me joy.

LGTM and worked perfectly when I started a hierarchy of instances and used this to visit all the levels.

grondo:
Thanks! Setting MWP.

wihobbs:
the rare codecov annoyance that _teaches_ me a cool new thing 🤩

Issue #6555, opened 2025-01-15
chaining jobs using afternotok


grondo:
A user asked this question:

> I noticed a pattern that occurs when you chain a bunch of runs using afternotok. You get a repeating pattern of completion and failure because once one job in the chain succeeds the next job will fail because its dependency can never be satisfied, then the next job will run because of the failure and complete, then the next will fail because its dependency failed, etc until all the chained jobs go through the queue. So the pattern looks something like this timeout -> node failure -> complete -> dependency can't be satisfied -> complete  -> dependency can't be satisfied -> ...
> 
> Is there am easy way to specify that you want to keep restarting a job until it completes then stop?

This seems like a valid use case we should have an easy way to satisfy.

grondo:
It is probably an oversight that jobs that never ran can satisfy a dependency. That is, it seems like `afternotok` should only apply to jobs that were started, then failed. Not jobs that were canceled before they even started. If this is addressed such that a job that becomes inactive without any `start` event no longer satisfies any `after*` dependency, I think the `afternotok` chaining will work as expected. (If a job in the chain succeeds, then the dependent `afternotok` job gets a fatal exception, and all its dependencies down the chain do too since the antecedents were never started)

However, this doesn't address the somewhat related issue of "keep running a jobspec until it succeeds". I'll open a separate issue on that one.

PR #6565, opened 2025-01-19
ensure only jobs that entered the RUN state can satisfy `--dependency=afterany|afternotok`

grondo:
This PR updates the `dependency-after` plugin to only allow jobs that entered the RUN state to satisfy any of the `after*` dependencies (which really only affects the `afterany` and `afternotok` types). This changes existing behavior, since pending jobs that are canceled would currently cause `after{any,notok}` dependent jobs to be released. This probably isn't what users want though, and it fixes the use case described in #6555, so it is probably the right approach.

### DIFF
--- a/t/t2271-job-dependency-after.t
+++ b/t/t2271-job-dependency-after.t
@@ -186,6 +186,32 @@ test_expect_success 'dependency=afternotok works' '
 	flux job wait-event -vt 15 \
 		-m type=dependency $ok1 exception
 '
+test_expect_success 'dependency=afternotok only applies to jobs that start' '
+	canceled_job=$(flux submit --urgency=hold false) &&
+	id2=$(flux submit --dependency=afternotok:$canceled_job hostname) &&
+	flux cancel $canceled_job &&
+	flux job wait-event -vt 15 -m type=dependency $id2 exception
+'
+test_expect_success 'chain of afternotok jobs are canceled if one fails before start' '
+	id=$(flux submit --urgency=hold false) &&
+	id2=$(flux submit --dependency=afternotok:$id hostname) &&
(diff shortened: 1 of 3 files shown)


t/t2271-job-dependency-after.t
```diff
@@ -186,6 +186,12 @@ test_expect_success 'dependency=afternotok works' '
 	flux job wait-event -vt 15 \
 		-m type=dependency $ok1 exception
 '
+test_expect_success 'dependency=afternotok only applies to jobs that start' '
+	canceled_job=$(flux submit --urgency=hold false) &&
+	id2=$(flux submit --dependency=afternotok:$canceled_job hostname) &&
+	flux cancel $canceled_job &&
+	flux job wait-event -vt 15 -m type=dependency $id2 exception 
+'
```
garlick, line 194: I guess it follows that a chain of dependent jobs would all be canceled but just for completeness, does it make sense to have an explicit test for it?

garlick (approved):
Ah this makes sense.

I just had one thought (inline) but feel free to ignore.

t/t2271-job-dependency-after.t
grondo: Yes, probably. Good idea.

grondo: Ok, I added a test that ensures a chain of afternotok are canceled when one is canceled, and another test that ensures the chain is canceled if a job succeeds.

grondo: I also noticed some trailing whitespace, so added a commit to fix that.

grondo:
Thanks! I'm setting MWP.

PR #6573, opened 2025-01-23
libflux: support ancestor paths as alternative to URI in `flux_open(3)` and `flux_open_ex(3)`

grondo:
This PR adds support for path-like shorthand arguments to `flux_open(3)` and `flux_open_ex(3)` to refer easily to ancestor Flux instances using a familiar filesystem idiom. The string ".." refers to the parent, "../.." its parent, and so on. "/" can be used to refer directly to the root instance, and for completeness "." indicates the current instance (and is therefore ignored as a path element). It is not an error to specify ".." at the root, so specifying more ".." elements than parents just returns a handle to the root instance.

I did my best in the documentation, but feel free to make suggestions.

### DIFF
--- a/src/common/libflux/handle.c
+++ b/src/common/libflux/handle.c
@@ -171,9 +171,26 @@ static char *strtrim (char *s, const char *trim)
     return *s ? s : NULL;
 }
 
+/* Return local URI from either FLUX_URI environment variable, or
+ * if that is not set, the builtin config. Caller must free result.
+ */
+static char *get_local_uri (void)
+{
+    const char *uri;
+    char *result = NULL;
+
+    if ((uri = getenv ("FLUX_URI")))
+        return strdup (uri);
+    if (asprintf (&result,
+                  "local://%s/local",
+                  flux_conf_builtin_get ("rundir", FLUX_CONF_INSTALLED)) < 0)
+        result = NULL;
+    return result;
+}
+
 flux_t *flux_open_ex (const char *uri, int flags, flux_error_t *errp)
 {
-    char *default_uri = NULL;
+    char *tmp = NULL;
     char *path = NULL;
     char *scheme = NULL;
     void *dso = NULL;
@@ -198,14 +215,10 @@ flux_t *flux_open_ex (const char *uri, int flags, flux_error_t *errp)
     /* Try to get URI from (in descending precedence):
      *   argument > environment > builtin
      */
-    if (!uri)
-        uri = getenv ("FLUX_URI");
     if (!uri) {
-        if (asprintf (&default_uri, "local://%s/local",
-                      flux_conf_builtin_get ("rundir",
-                                             FLUX_CONF_INSTALLED)) < 0)
+        if (!(tmp = get_local_uri ()))
             goto error;
-        uri = default_uri;
+        uri = tmp;
     }
     if (!(scheme = strdup (uri)))
         goto error;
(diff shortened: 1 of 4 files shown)


src/common/libflux/handle.c
```diff
@@ -171,9 +171,26 @@ static char *strtrim (char *s, const char *trim)
     return *s ? s : NULL;
 }
 
+/* Return local URI from either FLUX_URI environment variable, or
+ * if that is not set, the builting config. Caller must free result.
+ */
```
chu11, line 176: builting -> builtin

chu11:
I think this is a great idea.  One high level comment looking at the tests (and haven't looked through the meat of the code yet), I would have thought "paths" like `//` and `/.` would work, i.e. they would return the root instance, but you have them listed as bad args.

Perhaps a simple documentation and/or code tweak would be fine.  In the docs you use the words "path is composed of ...", which doesn't feel true if `//` or `/.` don't work.

Perhaps just make this simpler.  The valid inputs are `.` or `/` or `..` or a series of `..` separated by `/` and that's it.  So don't allow `Test(".././..", 1),` to work.  The only valid input after a slash is `..`.  I don't know if this makes the code easier or not, but it feels like it would make the code simpler?

grondo:
> I would have thought "paths" like // and /. would work, i.e. they would return the root instance, but you have them listed as bad args.

It doesn't make sense to support true "absolute paths" here (like provided by `flux-uri(1)` in #6562, so `/` with no other characters is just a special case to return the root instance. Seems like `//` is probably an input error and should return an error.

> Perhaps a simple documentation and/or code tweak would be fine. In the docs you use the words "path is composed of ...", which doesn't feel true if // or /. don't work.

This is the documentation:
>  - One or more ".." separated by "/" refer to a parent instance
>   (possibly multiple levels up the hierarchy)
> - "." indicates the current instance
> - A single slash ("/") indicates the root instance
> - ".." at the root is not an error, it just returns a handle to the root
>   instance

This is pretty explicit that a single slash is the root instance and only `..` separated by `/` refers to ancestors. Perhaps we could remove support for `.` if this seems confusing.

grondo:
Ah, good point. I can see that now. I wonder if I should just delete that line.

grondo:
Thanks, I've reworded to:
```
ANCESTOR PATHS
       As an alternative to a URI, the flux_open()  and  flux_open_ex()  func‐
       tions  also  take a path-like string indicating that a handle to an an‐
       cestor instance should be opened. This  string  follows  the  following
       rules:

          • One or more ".." separated by "/" refer to a parent instance (pos‐
            sibly multiple levels up the hierarchy)

          • "." indicates the current instance

          • A single slash ("/") indicates the root instance

          • ".." at the root is not an error, it just returns a handle to  the
            root instance
```

A bit better @chu11? Happy to take any other suggestions.

t/t3100-flux-in-flux.t
```diff
@@ -106,4 +106,47 @@ test_expect_success 'flux can launch multiple brokers per node (R lookup fallbac
...
+	    def test_pathlike_uris(self):
+	        for test in tests:
+	            l = flux.Flux(test.arg).attr_get("instance-level")
+	            self.assertEqual(int(l), test.level)
+
+	    def test_pathlike_bad_arg(self):
+	        for arg in ("/.", "//", "../f"):
+	            with self.assertRaises(OSError, msg=f"arg={arg}"):
+	                h = flux.Flux(arg)
+	                print(h.attr_get("instance-level"))
```
chu11, line 147: could add a test for bad stuff in middle of path, i.e. `../blah/..`

src/common/libflux/handle.c
```diff
@@ -188,6 +188,164 @@ static char *get_local_uri (void)
...
+            goto out;
+        }
+        str = NULL;
+    }
+    rc = n;
+out:
+    free (copy);
+    errno = EINVAL;
+    return rc;
+}
```
chu11, line 309: it seems a tad strange to set errno on a successful return.  maybe move to the "goto out" part above.

chu11 (approved):
Overall LGTM, just some small nits.  Also, note the typo I made in a comment earlier, still seems to be there.

grondo:
Thanks @chu11! Addressed all the comments (including the typo from earlier) and will set MWP.

PR #6574, opened 2025-01-23
broker: add year to log timestamps

### LINKED ISSUE #6569: include year in broker log timestamps
commit 3d99d7e98fe511cb718216ff14527755c9d2201a updated the broker log timestamp format to human readable form, but in doing so dropped the year.

A user has requested that the year be included in broker log timestamps in the case they're viewing historical job log files. It seems reasonable to add the year back to these timestamps.

grondo:
Problem: Flux broker log timestamps do not contain the year. This makes it difficult for users to determine when an error or other log message occurred in historical job output files.

Add the year to log timestamps emitted by the broker.

Fixes #6569

### DIFF
--- a/src/broker/log.c
+++ b/src/broker/log.c
@@ -421,14 +421,16 @@ static void log_timestamp (FILE *fp,
     struct timeval tv;
     char datetime[16]; /* 'MMM DD HH:MM:SS' */
     char timezone[16]; /* TZ abbrev should be short, give 15 chars max */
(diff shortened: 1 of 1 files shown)


chu11 (approved):
LGTM, you did forget the "Fixes ..." in the commit message though.

grondo:
Thanks! (Fixes is in the PR message) Will set MWP.

PR #6575, opened 2025-01-23
arm64 Docker cleanup

trws:
This is related to flux-framework/flux-sched#1328. The problem over there started because of a single-arch push to the alpine tag, which I'm guessing was a small mistake in a buildx command.  In investigating that, I realized our documented commands don't work, so this fixes that.  Also this includes commits to fix an issue in the docker entrypoint script, and port over the arm64 runner changes developed in the above sched PR.

NOTE: the last commit in here is mainly there so we can test the arm64 runners before merging anything.  If it works well, we might want to run some of the tests on arm all the time, or at least _run the tests after the builds on merge_.  Either way, we need to resolve that before this can be considered ready to merge.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -151,7 +151,7 @@ jobs:
 
   ci-checks:
     needs: [generate-matrix]
-    runs-on: ubuntu-latest
+    runs-on: ${{matrix.runner}}
     env:
       TAP_DRIVER_QUIET: 1
       FLUX_TEST_TIMEOUT: 300
@@ -210,10 +210,6 @@ jobs:
     - name: docker buildx
       uses: docker/setup-buildx-action@v3
       if: matrix.needs_buildx
-
-    - name: setup qemu-user-static
-      run: |
-        docker run --rm --privileged aptman/qus -s -- -p --credential aarch64
 
     - name: docker-run-checks
       timeout-minutes: ${{matrix.timeout_minutes}}
--- a/src/test/docker/README.md
+++ b/src/test/docker/README.md
@@ -65,8 +65,8 @@ Docker buildx extensions, see
 
 and run
 ```
-$  docker buildx build --push --platform=linux/arm64,linux/amd64 --tag fluxrm/testenv:jammy src/test/docker/jammy
-$  docker buildx build --push --platform=linux/386,linux/amd64,linux/arm64 --tag fluxrm/testenv:bookworm src/test/docker/bookworm
+$  docker buildx build --push --platform=linux/arm64,linux/amd64 --tag fluxrm/testenv:jammy -f src/test/docker/jammy/Dockerfile .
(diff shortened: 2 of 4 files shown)


trws:
The arm runners are showing a fedora-descendant failure in faketime/flux-cron I have not seen in local usage, will look into it. The hostlist issue on bookworm is also new to me, almost makes me wonder if there's a race or something but will poke it and see what falls out.

trws:
@garlick I might need some of your expertise on the cron/faketime issue.  Somehow I'm getting correct faketime for everything but `flux cron tab` in the cron test, but if I try to use `faketime` the executable in that spot to check it, it correctly reports being in a faketime environment.  I have a vague recollection we had some issues with it before, but it's striking me as absolutely baffling.  Might just be my brains still not all here, but I'm lost.  Still looking into bookworm.

Upside is at least that running the arm64 docker image locally on a mac repros perfectly what we're seeing here, so that's a big help.

grondo:
I should try to reproduce locally as well. Maybe the answer is to disable the faketime tests where they're known to be flaky. I admit I've looked into these issues before and didn't get much further than baffled as well.

trws:
That might be reasonable, especially since it’s just one set, specifically with cron.

grondo:
Now 3 of the arm64 builds on master are failing with errors like the following:
```
 56.19 gcc: internal compiler error: Segmentation fault signal terminated program cc1
 56.19 Please submit a full bug report, with preprocessed source (by using -freport-bug).
 56.19 See <https://gitlab.alpinelinux.org/alpine/aports/-/issues> for instructions.
```

trws:
That unfortunately makes sense, I've seen issues like that when an emulated instruction goes sideways.  If things pass here, I'll get it cleaned up so we can move that in.

trws:
Ok, I'd say these are all set.  I'm going to pull out the "always run all the arm64" commit and see where we land.  The sched check hanging is extremely perplexing, kinda hoping it's an issue that's specific to this kinda odd run configuration.  Will see.

image

grondo:
>  The sched check hanging is extremely perplexing, kinda hoping it's an issue that's specific to this kinda odd run configuration.

Yeah, afaict it seems to be stuck just trying to start the docker container?

This LGTM once the `DEBATE` commit is dropped as you note above. I'll take another quick look after that, but thanks @trws for running all this down!

trws:
It certainly seems like it.  I dropped the DEBATE commit, should be set. 🤞

trws:
Looks like there's still an issue here.  All I can think of is that somehow the `runuser` -> `sudo -u` change is causing issues.

trws:
Dropped the runuser commit to see where we land.

grondo:
Looks like that did it, though I can't think of why??

grondo (approved):
LGTM! Thanks!

trws:
> Looks like that did it, though I can't think of why??

All I could think of is that somehow `sudo runuser -u` is running without interactive authentication but `sudo -u` is requesting a password.  *Why* that would happen, or even how it could be possible though... yeah I have no idea.

Issue #6576, opened 2025-01-24
Where can shell error log (shell_log_error(...) or shell.log) entries be viewed?


sam-maloney:
Short version: where should I be able to view entries generated by calls to `shell_log_error(...)` and/or `flux_shell_log(...)`?

Long version for context:

I would ultimately like to test jobs with resource ranges, as specified in RFC 14, as opposed to just the integer counts in V1 jobspec. As an initial step I tried submitting a jobspec using the python interface with a version number != 1, but which was otherwise perfectly valid V1 jobspec.

My initial test failed, but produced an error message, and a quick search of the message text led me to a version check in the `jj_get_counts_json(...)` function of `src/common/libjob/jj.c`. Removing this version check and re-submitting leads to the job being allocated resources but failing with a non-zero exit code during execution.

The problem is that for this failure I cannot see any error message being logged anywhere to indicate what exactly caused the failure, which made hunting down the next change much more arduous. Eventually I tracked it to a version check in the `jobspec_parse(...)` function in `src/shell/jobspec.c`, and once this check was also removed the job was scheduled and ran successfully.

That version check shows an error message being generated ("Invalid jobspec version: expected 1 got %d") so my question is where I should have found this in the logs? (as it would have been much easier to track down the problem if I'd had the error message to point me in the right direction!)

It appears that `jobspec_parse(...)` must have been called from `shell_init_jobinfo(...)` in `src/shell/info.c` which shows a call to `shell_log_error(...)`. This in turn appears to be a macro defined in either `src/shell/shell.h` as
```c-like
#define shell_log_error(...) \
    flux_shell_log (FLUX_SHELL_PLUGIN_NAME, \
                    FLUX_SHELL_ERROR, __FILE__, __LINE__, __VA_ARGS__)
```
or in `src/shell/plugstack.c` as
```c-like
#ifdef PLUGSTACK_STANDALONE
#undef  shell_log_error
#undef  shell_log_errno
#define shell_log_error(...) fprintf (stderr, __VA_ARGS__)
#define shell_log_errno(...) fprintf (stderr, __VA_ARGS__)
#endif
```

[...]

PR #6578, opened 2025-01-24
shell: fix lost log messages during initialization

grondo:
This PR (hopefully) fixes the issue reported in #6576. Early shell errors are lost because, for unknown reason, the shell was calling `exit (1);` instead of `shell_die()`.

This fixes that issue and adds a test.


t/t2608-job-shell-log.t
```diff
@@ -220,4 +220,18 @@ test_expect_success 'flux-shell: stdout/err from task.exec works' '
 	grep "^this is stderr" print.err &&
```
chu11, line 220: commit message typo "erorrs"

chu11 (approved):
LGTM, just that one typo

t/t2608-job-shell-log.t
grondo: Thanks! Fixed and will set MWP.

PR #6584, opened 2025-01-27
shell: fix incorrect assignment of shell rank ids when broker ranks appear unordered in R

### LINKED ISSUE #6582: shell: assignment of tasks to shells contradicts taskmap on partially allocated elcap nodes
Problem: on MI300 nodes with 96 cores, running a job that requires 97 thru 191 cores without specifying a node count seems to cause the shell to assign the wrong ranks to nodes compared to the taskmap.

Example:
```
[garlick@tuolumne1005:~]$ flux run --label-io -n 191 hostname | grep 190:
190: tuolumne1005
[garlick@tuolumne1005:~]$ flux job taskmap --to=hosts $(flux job last)
tuolumne1005: 0-95
tuolumne1006: 96-190
```

Adding a `userrc` that prints the shell hostname confirms this:
```
garlick@elcap1063:~]$ flux run -n192 -o mpibind=off -o spindle.level=off -o verbose -o userrc=rc.lua true 2>&1 |grep elcap
0.058s: flux-shell[0]: elcap1063
0.066s: flux-shell[1]: elcap1064
[garlick@elcap1063:~]$ flux run -n191 -o mpibind=off -o spindle.level=off -o verbose -o userrc=rc.lua true 2>&1 |grep elcap
0.066s: flux-shell[0]: elcap1064
0.058s: flux-shell[1]: elcap1063
[garlick@elcap1063:~]$ flux job taskmap --to=hosts $(flux job last)
elcap1063: 0-95
elcap1064: 96-190
```
This does not reproduce if `-N2` is added.
```
[garlick@elcap1063:~]$ flux run -n97 -N2 -o mpibind=off -o spindle.level=off -o verbose -o userrc=rc.lua true 2>&1 |grep elcap
0.059s: flux-shell[0]: elcap1063
0.092s: flux-shell[1]: elcap1064
[garlick@elcap1063:~]$ flux job taskmap --to=hosts $(flux job last)
elcap1063: 0-48
elcap1064: 49-96
```

First noted in flux-framework/flux-coral2#222

grondo:
This PR should fix #6582. The rcalc code in the shell currently assigns shell ranks based on the order broker ranks appear in `R_lite` in the job's R. This breaks an assumption elsewhere that rcalc derived shell ranks match the order of broker ranks and the R nodelist.

This PR simply sorts the rcalc->ranks array and reassigns shell rank ids if necessary.

A test is added to ensure out of order ranks in `R_lite` still produce the expected task layout and rank assignment.

### DIFF
--- a/src/shell/rcalc.c
+++ b/src/shell/rcalc.c
@@ -126,12 +126,22 @@ static int rankinfo_get_children (struct rankinfo *ri,
     return 0;
 }
 
+static int cmp_rank (const void *a, const void *b)
+{
+    const struct rankinfo *ra = a;
+    const struct rankinfo *rb = b;
+    return (ra->rank < rb->rank ? -1 : 1);
+}
+
+
 static int rcalc_process_all_ranks (rcalc_t *r, flux_error_t *errp)
 {
     json_t *entry;
     size_t index;
     json_error_t error;
     int n = 0;
+    unsigned int max_rank = 0;
+    bool sorted = true;
 
     json_array_foreach (r->R_lite, index, entry) {
         const char *rank;
@@ -156,6 +166,14 @@ static int rcalc_process_all_ranks (rcalc_t *r, flux_error_t *errp)
             struct rankinfo *ri = &r->ranks[n];
             ri->id = n;
             ri->rank = i;
+
+            /* Detect if R_lite ranks are unordered:
+             */
+            if (i >= max_rank)
+                max_rank = i;
+            else
+                sorted = false;
+
             if (rankinfo_get_children (ri, children, errp) < 0) {
                 idset_destroy (ids);
                 return -1;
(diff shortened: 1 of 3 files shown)


garlick:
What do you think about the wording in [RFC 20](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_20.html#execution) which says:
> The order of hostnames MUST correspond to the order of execution targets in [R_lite](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_20.html#R_lite).

That would seem to indicate that `R_lite` should not be sorted and this is actually a bug in the scheduler?

grondo:
> The order of hostnames MUST correspond to the order of execution targets in [R_lite](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_20.html#R_lite).

Ah, I kind of assumed order here just meant rank-order not the order of the R_lite array. Definitely a wording update is  probably needed.

garlick (approved):
Sounds reasonable.  I can propose a small update to the RFC to make this more clear.

Just tested on my test cluster and on elcap with the PMI reproducer and everything works with this change applied.

Very nice to put this one to bed!  Thanks @grondo!

PR #6586, opened 2025-01-28
add resource journal

garlick:
This adds a resource journal streaming RPC similar the one offered by the job manager.  Just a first cut at this point.

This doesn't change what's posted to the persistent `resource.eventlog` in the KVS but does add one new event called `restart` that's only for journal consumption.  It provides a baseline for mapping execution targets to hostnames in the current instance, and sets the initial `online` set after a restart.

Unlike the job manager journal, this doesn't have as much volume to deal with so no options for event filtering or skipping historical data are provided as yet.

`flux resource eventlog` can be used to dump and optionally follow this log.  This is not currently polished at all - it just dumps the events in JSON form, one per line.

For more detail on what's in this log and how the journal is formatted, see the proposed RFC:
- flux-framework/rfc#440

### DIFF
--- a/doc/man1/flux-resource.rst
+++ b/doc/man1/flux-resource.rst
@@ -22,6 +22,7 @@ SYNOPSIS
 | **flux** **resource** **reload** [-f] [--xml] *path*
 
 | **flux** **resource** **acquire-mute**
+| **flux** **resource** **eventlog** [*-w* *EVENT*] [*-F*]
 
 DESCRIPTION
 ===========
@@ -356,6 +357,24 @@ Tell the resource module to stop sending RFC 28 ``resource.acquire`` responses
 to the scheduler.  This is used during Flux instance shutdown to avoid asking
 the scheduler to needlessly process OFFLINE updates.
 
+eventlog
+--------
+
+.. program:: flux resource eventlog
+
+Watch the resource journal, which is described in RFC 44.
+
+.. option:: -F, --follow
+
+  After printing the current eventlog, wait for new events and print
+  them as they arrive.
+
+.. option:: -w, --wait=EVENT
+
+  Wait for the specified event to be posted, print it, then quit.
+  The current set of valid events events is *restart*, *resource-define*,
+  *online*, *offline*, *drain*, *undrain*, *torpid*, and *lively*.
+
 OUTPUT FORMAT
 =============
 
@@ -520,3 +539,4 @@ FLUX RFC
 | :doc:`rfc:spec_22`
 | :doc:`rfc:spec_27`
 | :doc:`rfc:spec_29`
+| :doc:`rfc:spec_44`
--- a/src/cmd/flux-resource.py
+++ b/src/cmd/flux-resource.py
@@ -720,6 +720,41 @@ def emit_R(args):
     print(rset.encode())
 
 
+def print_events(events, follow, wait):
+    if not events and not follow and not wait:
+        return False
+    for entry in events:
+        print(json.dumps(entry))
+        if wait and entry["name"] == wait:
+            return False
+    return True
+
+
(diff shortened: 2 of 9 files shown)


garlick:
I added a quick `--wait=EVENT` option to `flux resource eventlog` and a sharness test.  Will remove WIP - maybe good for a start that we can experiment with a bit?

grondo:
Yeah, let me see if I can generalize the `JournalConsumer` class to also work with the resource eventlog.

garlick:
Did we want to add this to 0.71 as an experimental feature?

grondo:
I was halfway through reviewing it and had a few commits that abstract the current Python `JournalConsumer` class to add a `ResourceJournalConsumer` class and I was currently writing some tests.

Have we estimated what the size of the in-memory history of all events would be like on something like Elcap with a long uptime?

garlick:
That might be worrisome actually, now that you mention it.  Perhaps we'll  need some way to cull non-persistent events (e.g. other than drain/undrain) that are older than a threshold.  It might be a bit too last minute to add that here, sadly.

grondo:
I'm not sure there's a rush to add this as experimental because I'm not sure there's an immediate use case without the Python consumer class (unless I'm forgetting something!)

garlick:
I think this is a priority for @kkier but there is little advantage in putting this in as is - in fact it may cause problems.  Let's try to deliver a good first step in the next release.

kkier:
Just confirming - this is a priority for me (and the people I report to on both sides of the aisle, as it were), but if it ends up in an off-cycle release that I have to manually distribute, that's a thing I can deal with.

garlick:
> but if it ends up in an off-cycle release that I have to manually distribute, that's a thing I can deal with.

To be clear, although it's not that hard on our end to produce one-off tags and packages, this particular change modifies a core service and would require downtime to distribute.

grondo:
One thing I've noticed in my testing is that canceling the journal RPC with `reosurce.journal-cancel` doesn't seem to result in ENODATA eventually being sent to the client. This differs from the job manager journal service and would slightly complicate the implementation of Python interfaces with a shared base class. 

Of course it is possible I've made a mistake in testing.

kkier:
> > but if it ends up in an off-cycle release that I have to manually distribute, that's a thing I can deal with.
> 
> To be clear, although it's not that hard on our end to produce one-off tags and packages, this particular change modifies a core service and would require downtime to distribute.

100% understood, it is what it is.

garlick:
> canceling the journal RPC with `reosurce.journal-cancel` doesn't seem to result in ENODATA eventually being sent to the client

It should. (if spelled correctly :-)  However I didn't add a test for that so I should verify.

grondo:
I stumbled across the problem I think:
```diff
diff --git a/src/modules/resource/reslog.c b/src/modules/resource/reslog.c
index 61aae1218..513d54cc6 100644
--- a/src/modules/resource/reslog.c
+++ b/src/modules/resource/reslog.c
@@ -328,7 +328,7 @@ static void journal_cb (flux_t *h,
         errstr = "journal requires streaming RPC flag";
         goto error;
     }
-    if (send_backlog (reslog, msg))
+    if (!send_backlog (reslog, msg))
         return;
     if (flux_msglist_append (reslog->consumers, msg) < 0)
         goto error;

```

garlick:
Aww I tested that. Must have broken it with a last minute change.  I need to add a test to :-(

garlick:
Ok some small changes:
- fix the bug @grondo pointed
- renamed -f to -F to match python class (for future work by @grondo)
- cover -F in test
- rebase on current master

grondo (approved):
This LGTM!

garlick:
Thanks!  I'll set MWP.

PR #6595, opened 2025-02-01
kvs: remove unused internal transaction request API code

chu11:
With the removal of KVS fence support in #6592, a large amount of the transaction request ("treq") API is no longer needed.  Remove that code and code that uses it.

As a side note, you may notice the "treq" API is now very simple.  Here's the internal structs

```
struct treq_mgr {
    zhash_t *transactions;
};

struct treq {
    char *name;
    const flux_msg_t *request;
    int flags;
};
```

So I hope I can "squash" this into the kvsroot/kvstxn API, but that is for a follow up PR.  This PR is just for the "obvious" removal of dead code.

### DIFF
--- a/src/common/libkvs/kvs_txn_compact.c
+++ b/src/common/libkvs/kvs_txn_compact.c
@@ -42,7 +42,7 @@
  *
  * append "A"
  * write "B"
- * append "c"
+ * append "C"
  *
  * we cannot combine the appends of "A" and "C".  In this scenario, we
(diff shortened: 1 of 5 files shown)


garlick (approved):
LGTM.  Looks like there may be some things caught by CI that need a look, but o/w great.

chu11:
@garlick thanks!  There was a dumb use-after-free in the unit tests due to test changes.  Fixed that.  Will set MWP

PR #6597, opened 2025-02-01
do not export `PYTHONPATH` by default in Flux commands

### LINKED ISSUE #6594: don't prepend builtin `PYTHONPATH` for all commands in `flux(1)`
Historically, the `flux` command driver has prepended to `PYTHONPATH` so that Python subcommands pick up the right version of the bindings. However, this is inconvenient for users that want to `flux run` a Python program that uses Flux python bindings with a non-system version of Python.

Now that we run all Python subcommands with `py-runner.py`, it probably isn't necessary to always prepend to `PYTHONPATH`. Instead only do it with `flux python`.

grondo:
This PR attempts to fix #6594 while preserving a bit of backwards compatibility and ensuring that Python subcommands get a `sys.path` that will load the correct version of the Flux Python bindings. The steps taken to accomplish this were the following:

 - save the current `PYTHONPATH` before executing the `py-runner.py` Python subcommand wrapper so that `PYTHONPATH` can be modified before running the wrapper. The wrapper script then reestablishes the original `PYTHONPATH` after `sys.path` is internally set up correctly. This is necessary so that `py-runner.py` imports the correct `_flux` cffi modules, from which it derives other necessary paths.
 - Only modify `PYTHONPATH` in `flux env` and `flux python` (in addition to the case above). Thus, `PYTHONPATH` is no longer modified in the environment of jobs or `flux start`.
 - Since `PYTHONPATH` no longer has the path to the build tree Python bindings in the testsuite, even under `flux start`, modify `PYTHONPATH` for sharness tests in `sharness.d/01-setup.sh` so that non-Flux subcommand Python scripts run as part of the testsuite use the builddir Python bindings and not system bindings.

### DIFF
--- a/doc/man1/flux.rst
+++ b/doc/man1/flux.rst
@@ -80,9 +80,6 @@ is provided below:
    * - :envvar:`LUA_CPATH`
      - Lua binary module search path
 
-   * - :envvar:`PYTHONPATH`
-     - Python module search path:
-
 
 RESOURCES
 =========
--- a/doc/man7/flux-environment.rst
+++ b/doc/man7/flux-environment.rst
@@ -552,10 +552,15 @@ of compiled-in install paths and the environment.
 .. envvar:: PYTHONPATH
             FLUX_PYTHONPATH_PREPEND
 
-   :envvar:`PYTHONPATH` is set so that sub-commands can find required Python
-   libraries:
+   With the :command:`flux python` and :man1:`flux-env` subcommands,
+   :envvar:`PYTHONPATH` is set such that the correct version of
+   Python modules can be found. In these cases the path is set
+   to the following:
 
       $FLUX_PYTHONPATH_PREPEND : install-path : $PYTHONPATH
+
+   The :man1:`flux` command does not otherwise modify :envvar:`PYTHONPATH`
+   unless :envvar:`FLUX_PYTHONPATH_PREPEND` is set.
 
    Values may include multiple directories separated by colons.
 
--- a/doc/python/basics.rst
+++ b/doc/python/basics.rst
@@ -9,10 +9,6 @@ Importing the ``flux`` Python package
    ``flux-python`` package.
 
 Flux's Python bindings are available with any installation of Flux.
-When running in a Flux instance, Flux will export the
-`PYTHONPATH <https://docs.python.org/3/using/cmdline.html#envvar-PYTHONPATH>`_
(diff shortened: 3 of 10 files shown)


jameshcorbett (approved):
This seems to me like it requires more environment-variable / process-manipulation knowledge than Python knowledge, but anyway I looked through this and did some research to make sure there isn't another way to manipulate `sys.path`--I couldn't find one. This looks like a nice improvement to me.

trws (approved):
This all looks right to me and seems like it should solve that problem, though I admit I'm still chewing on two thoughts, both centered around whether we might actually want to set `PYTHONPATH` even less.

1. We have classically set `PYTHONPATH` in `flux env` because that's how we set up the environment for any given flux command or `flux run`.  Maybe better to have it be a separate change, or see what kind of fallout it would have, but I'm tempted to say we should remove it from there as well, or perhaps put it behind a `--pythonpath` flag or similar.
2. How bad would it be to get the scripts run like flux python commands/scripts? I'm guessing that's a non-trivial amount of work, but setting `PYTHONPATH` for the whole test suite could hide issues where we need to wrap with our python runner or similar and forget.

grondo:
Great points @trws and I was thinking about both of them as well.

For 1., I initially did _not_ have `flux env` set `PYTHONPATH`, but  I found this unfortunate note in our [Python basics](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/python/basics.html) docs:

> If you want to import the package from outside of a Flux instance, running `/path/to/flux env | grep PYTHONPATH` in your shell will show you what PYTHONPATH would be set to if you were running in a Flux instance built/installed at `/path/to/flux`.

So I figured we should keep it in `flux env` for now (perhaps to be removed in the next release)

Similar thinking for all the Python scripts in the testsuite. I agree with your assessment, and perhaps we should open an issue to remove the forced `PYTHONPATH` in the testsuite soon as well.

trws:
That all sounds right to me, maybe throw up an issue so we don't lose track of it referencing the docs and test suite?

grondo:
Yeah, it also occurs to me that `flux config builtin python_path` is a better way to get the builtin python path too. Maybe we make a `PYTHONPATH` alias for that too so `flux config builtin PYTHONPATH` also works.

grondo:
Ok, opened #6600

grondo:
I'll set MWP now.

Issue #6596, opened 2025-02-01
housekeeping: misplaced partial release


jameshcorbett:
@milroy and I were seeing some partial-cancel errors from Fluxion even though we thought we had disabled partial release. I was taking this (mistakenly?) as an indication that partial release was disabled:
```
[corbett8@rzadams1:~]$ pdsh -g compute flux config get job-manager.housekeeping
erzadams1002: {}
erzadams1028: {}
erzadams1006: {}
erzadams1007: {}
erzadams1008: {}
erzadams1030: {}
erzadams1029: {}
erzadams1004: {}
erzadams1033: {}
erzadams1021: {}
erzadams1027: {}
erzadams1003: {}
erzadams1034: {}
erzadams1023: {}
erzadams1032: {}
erzadams1024: {}
erzadams1031: {}
erzadams1019: {}
...
```

But I just found an RPC that didn't have the final flag set:
```
[root@rzadams1:corbett8]# flux module trace -H -f sched-fluxion-qmanager | grep -B30 "\"final\": false"
    }
  }
}
[Jan31 19:11]  sched-fluxion-qmanager rx > sched.free [249]
{
  "id": 521311806304568320,
  "R": {
    "version": 1,
    "execution": {
      "R_lite": [
        {
          "rank": "28",
          "children": {
            "core": "0-95",
            "gpu": "0-3"
          }
        }
      ],
      "starttime": 0.0,
      "expiration": 0.0,
      "nodelist": [
        "rzadams1028"
      ],
      "properties": {
        "plarge": "28",
        "pdev": "28",
        "pall": "28"
      }
    }
  },
  "final": false

```

They seem to come through every once in a while, but this was the first one I managed to snag this way.

milroy:
The `"final": false` value triggers the partial cancel logic (https://github.com/flux-framework/flux-sched/blob/471da50d167bdaa12a3dc1429c5b6b35674f1aa5/qmanager/policies/base/queue_policy_base.hpp#L649-L658):                 

```C++
if (!final) {
    if (cancel (h, job_it->second->id, R, true, full_removal) != 0) {
        flux_log_error (flux_h,
                        "%s: .free RPC partial cancel failed for jobid "
                        "%jd",
                        __FUNCTION__,
                        static_cast<intmax_t> (id));
        errno = EINVAL;
        goto out;
    }
```

garlick:
[...]

PR #6598, opened 2025-02-03
fix housekeeping reconfiguration problem

garlick:
Problem: as noted in #6596, when housekeeping is reconfigured, it retains the old value of `release-after` unless the key is explicitly set.

And the only way to disable partial release is to not set it.

This fixes the config callback so that release-after is set to the default if not mentioned.


grondo (approved):
This LGTM!

garlick:
Thanks!  I'll set MWP then.   What an unfortunate bug :facepalm:

PR #6601, opened 2025-02-04
doc: reword initial program environment section in `flux-environment(7)`

grondo:
Problem: The section of flux-environment(7) detailing the initial program environment can be confusing to users because the set of environment variables *not* set is prominently listed first in a nice bulleted list, while environment variable(s) that *are* set appear later in paragraph form. This may lead users that are casually scanning the document to assume that the listed environment variables will be found in the initial program environment.

Swap the section so that the environment variables that are set and the suggested commands to obtain common values using Flux commands are listed first and add documentation to emphasize their importance.

De-emphasize the unset variables by listing them in paragraph form instead of a bulleted list.

### DIFF
--- a/doc/man7/flux-environment.rst
+++ b/doc/man7/flux-environment.rst
@@ -158,27 +158,27 @@ INITIAL PROGRAM ENVIRONMENT
 ===========================
 
 The :man1:`flux-alloc` interactive shell and the :man1:`flux-batch` batch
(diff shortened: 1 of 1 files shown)


garlick (approved):
Oh good improvement!

grondo:
Thanks! I'll set MWP.

PR #6604, opened 2025-02-04
kvs: remove internal treq api

chu11:
effectively "squash" the treq api into the kvsroot API since there was almost nothing left in the treq api.  Removing a whole sub-library within the kvs.  woot!


src/modules/kvs/kvsroot.c
```diff
@@ -253,6 +254,44 @@ int kvsroot_mgr_iter_roots (kvsroot_mgr_t *krm, kvsroot_root_f cb, void *arg)
...
+                          name,
+                          (flux_free_f) free) < 0)
+        goto error;
+
+    if (zhash_insert (root->transaction_requests,
+                      name,
+                      (void *)flux_msg_incref (request)) < 0) {
+        errno = EEXIST;
+        goto error;
+    }
```
garlick: Just wondering if this is the best container?  I think we decided of the hashes, `zhashx` had more sane growth management than `zhash`.  I guess if the number of messages stored in the hash is always small that may not come into it.

Would a `flux_msglist_t` possibly be a better fit or is the lookup by name important?

Sorry I'm not entirely remembering how this code works so I'm just asking the question.

garlick (approved):
LGTM!

Just a couple of questions inline for your consideration - nothing that should hold this up necessarily.

src/modules/kvs/kvsroot.c
chu11: I was thinking of that.  But left it as a `zhash` just because that was what was there before.  Could think of that for some potential cleanup / improvement.

I had contemplated a `zlist` b/c now that we only support commits, the "transactions" should reasonably be done in order.  So name matching should always occur with the first entry.  But not sure if that's worth it vs a hash since we would want to double check and match names anyways.

src/modules/kvs/kvs.c
```diff
@@ -1789,21 +1781,18 @@ static void commit_request_cb (flux_t *h,
      * finalize_transaction_bynames() to send error code to original
      * send.
      */
-    if (!(tr = treq_create (msg, ctx->rank, ctx->seq++))) {
-        flux_log_error (h, "%s: treq_create", __FUNCTION__);
+
+    if (asprintf (&name, "transaction_req.%u.%u", ctx->rank, ctx->seq++) < 0)
         goto error;
```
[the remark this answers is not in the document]
chu11: oh that's a good idea, since we aren't allocating a `treq_t` anymore we don't have to allocate it .

chu11:
re-pushed with the static buffer fix recommended above.  It ends up I did have a rebase mistake in my commit history which I also fixed.  I'll set MWP

PR #6611, opened 2025-02-06
kvs: remove stats clearing

### LINKED ISSUE #6607: kvs: remove stats clearing
All the code related to clearing of stats is from very early prototyping with the KVS.  It adds a little bit to the KVS code bloat (#6570)

Nowadays, it's not really used.  If we ever need to test the KVS, we start an instance, test whatever we want, then shut down the instance.

chu11:
Problem: KVS stats clearing code was added very early on in KVS prototyping for testing.  It is basically not used anymore.  If we
want to test something that requires a "clean" slate for the KVS, we can just start a new instance and test with that.

Solution: Remove all KVS stats clearing code.  Remove all associated tests.

Fixes #6607


garlick (approved):
LGTM!

PR #6612, opened 2025-02-06
flux-job: do not override `-L, --color` when `-H, --human` is used in `eventlog` and `wait-event` subcommands

grondo:
While working on something else I discovered that `flux job eventlog -HL ...` wasn't forcing color on like it should. This is because `-H, --human` forces `color = auto`, overwriting the user's command line choice.

`auto` is already the default. Stop doing that.

Add a test.


chu11 (approved):
LGTM!

grondo:
Thanks! Setting MWP

PR #6613, opened 2025-02-07
docs: fix constraints negation in man pages

jameshcorbett:
Problem: the docs on constraint negation is wrong, it has '^' as the negation token but it is actually '-' both in practice and in RFC 35.

Change it.

### DIFF
--- a/doc/man1/common/job-constraints.rst
+++ b/doc/man1/common/job-constraints.rst
@@ -32,8 +32,8 @@
    properties
      Require the set of specified properties. Properties may be
      comma-separated, in which case all specified properties are required.
-     As a convenience, if a property starts with ``^`` then a matching
-     resource must not have the specified property. In these commands,
+     As a convenience, if a property starts with ``-`` then a matching
+     resource must *not* have the specified property. In these commands,
      the ``properties`` operator is the default, so that ``a,b`` is equivalent
      to ``properties:a,b``.
 


doc/man1/common/job-constraints.rst
```diff
@@ -32,8 +32,8 @@
    properties
      Require the set of specified properties. Properties may be
      comma-separated, in which case all specified properties are required.
-     As a convenience, if a property starts with ``^`` then a matching
-     resource must not have the specified property. In these commands,
+     As a convenience, if a property starts with ``-`` then a matching
+     resource must *not* have the specified property. In these commands,
```
grondo, line 36: This actually is not correct. The negation operator applies only to the RFC 35 constraint query syntax to negate a term, e.g. `-host:host0` or `-property:foo`. To negate a property by name within the `property` operator, a `^` is used, e.g. `property:^foo` is equivalent to `-property:foo` or `not property:foo`. 

The confusion comes about because when a constraint term does not have a scheme, e.g. `host:` or `property:`, then `property` is assumed, so `-foo` alone becomes `-property:foo` or `not property:foo`.

Very much agree this is a bit confusing, but it came about because properties were introduced before the more elaborate constraint query syntax.

For the implementation where `^` is supported in Fluxion see here:

https://github.com/flux-framework/flux-sched/blob/4aacb4e857d40d5ef9d5999414d08f1eb9a3b31a/resource/libjobspec/constraint.cpp#L142-L145
```cpp
        if (property[0] == '^') {
            negate = true;
            property = property.substr (1);
        }
```

grondo: BTW, you can see the difference with `--dry-run`:

```console
$ flux run --dry-run -n1 --requires=-foo hostname | jq .attributes.system.constraints
{
  "not": [
    {
      "properties": [
        "foo"
      ]
    }
  ]
}
$ flux run --dry-run -n1 --requires=^foo hostname | jq .attributes.system.constraints
{
  "properties": [
    "^foo"
  ]
}
```

But if a user tried `--requires=properties:-foo` as documented here, the constraint would require a property of literally `"-foo"`:

```console
$ flux run --dry-run -n1 --requires=properties:-foo hostname | jq .attributes.system.constraints
{
  "properties": [
    "-foo"
  ]
}
```

jameshcorbett: ah my bad, I was misled by this:
```
bash-5.1$ flux submit --dry --requires=^foo -N1 hostname | jq .attributes.system.constraints
{
  "properties": [
    "^foo"
  ]
}
bash-5.1$ flux submit --dry --requires=-foo -N1 hostname | jq .attributes.system.constraints
{
  "not": [
    {
      "properties": [
        "foo"
      ]
    }
  ]
}
```

grondo: Yeah, sorry they're equivalent. I'd be open to changing this special case for `properties` if we want to go down that road.

jameshcorbett: Oh oops I didn't see you posted the exact same thing as I did above, I only saw your first comment when I was writing mine.

grondo: NP!

PR #6614, opened 2025-02-07
add `ResourceJournalConsumer` and `EventLogFormatter` classes and use them to standardize `flux resource eventlog` 

grondo:
The main purpose of this PR is to add the Python `ResourceJournalConsumer` class, which mimics the existing job manager `JournalConsumer` class. In fact, most of the original `JournalConsumer` class implementation is moved to an abstract `JournalConsumerBase` class, then both the job manager `JournalConsumer` and `ResourceJournalConsumer` classes inherit from that.

In addition, the `flux resource eventlog` utility is modified to have similar options to the rest of the `eventlog` utilities in Flux (`--format`, `--time-format`, `--color`, `--human`)using a new `EventLogFormatter` class  and the newly added `ResourceJournalConsumer` class.

### DIFF
--- a/src/bindings/python/flux/abc/journal.py
+++ b/src/bindings/python/flux/abc/journal.py
@@ -1,0 +1,359 @@
+###############################################################
+# Copyright 2025 Lawrence Livermore National Security, LLC
+# (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+#
+# This file is part of the Flux resource manager framework.
+# For details, see https://github.com/flux-framework.
+#
+# SPDX-License-Identifier: LGPL-3.0
+###############################################################
+
+"""Abstract classes for journal consumer interfaces"""
+
+import errno
+import time
+from abc import ABC, abstractmethod
+from collections import deque
+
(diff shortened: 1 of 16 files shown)


src/bindings/python/flux/abc/journal.py
```diff
@@ -0,0 +1,359 @@
...
+    This base class implements a wrapper around journal style RPCs
+    which allow clients to obtain a stream of events, possibly including
+    historical data.
+
+    Subclasses of :obj`JournalConsumerBase` should be implemented to more
+    completely handle the specifics of a given interface, but all such
+    subclasses will follow the interface documented here.
+
+    A journal consumer is created by passing the constructor an open
+    :obj:`~flux.Flux` handle, the topic string of the journal servicealong
```
garlick, line 69: servicealong needs a space

garlick (approved):
Nicely done!  Good docs too!

I gave this an admittedly light skim and tried it out a bit and LGTM!  Very nice to have the `flux resource eventlog` command dressed up like one expects.

grondo:
Thanks! I've rebased and fixed that typo and will set MWP now.

PR #6616, opened 2025-02-07
avoid scheduling jobs on compute nodes that are not cleaned up

### LINKED ISSUE #6590: after an emergency restart, flux doesn't know about user processes that are still running
Problem: If flux is not shut down cleanly, running jobs can escape its control.

When flux restarts, it should detect any running jobs, prolog, epilog, or housekeeping systemd units, and not release those execution targets to the scheduler until they are clean.

Perhaps this could be handled by the resource module's "monitoring" subsystem.

garlick:
Problem: after a broker restart, "stuck" housekeeping, epilog, prolog, or job shell tasks might still be running, but flux is unaware and new work may be scheduled there even though there might be a problem, or those tasks might be holding on to resources.

When those things are run under systemd, we have the machinery for finding them and tracking them readily at hand. 

This PR does the following
- modifies `sdbus` so it can be loaded twice, once for user and once for system systemd instances
- add a `sdmon` monitoring module that tracks running flux units on both busses
- have `sdmon` join a `sdmon.idle` broker group at startup, after it has verified that no units are running
- modify the resource module so it monitors `sdmon.idle` instead of  `broker.online` when configured to use systemd

The net effect is that nodes that require cleanup remain offline until the node is cleaned up.

`sdmon` also logs the systemd units it has found.  Here's an example where I kill -9 a broker while housekeeping is running, then start it back up again
```
[  +0.192069] sdmon.err[7]: flux-housekeeping@fAMr3PwFUNb.service needs cleanup - resources are offline
...
[ +26.900866] sdmon.err[7]: cleanup complete - resources are online
```
Before cleanup is complete, `flux resource status` reports
```
      STATE UP NNODES NODELIST
       avail  ✔      7 picl[0-6]
      avail*  ✗      1 picl7
```

Seems like this does the bare minimum to resolve #6590 

This does seem a bit thin in the tooling department.   The possibilities are pretty broad so for now, I wanted to get this posted and get feedback on the way the resource module is tied into `sdmon` using broker groups.

### DIFF
--- a/etc/rc1
+++ b/etc/rc1
@@ -28,6 +28,7 @@ if test "$(flux config get --default=false systemd.enable)" = "true"; then
     modload all sdbus
     modload all --name sdbus-sys sdbus system
     modload all sdexec
+    modload all sdmon
 fi
 
 if test $RANK -eq 0; then
--- a/etc/rc1
+++ b/etc/rc1
@@ -26,6 +26,7 @@ modload all content
 modload all barrier
 if test "$(flux config get --default=false systemd.enable)" = "true"; then
     modload all sdbus
+    modload all --name sdbus-sys sdbus system
     modload all sdexec
 fi
 
--- a/etc/rc3
+++ b/etc/rc3
@@ -37,6 +37,7 @@ modrm 0 job-manager
 modrm all job-ingest
 
 modrm 0 cron
+modrm all sdmon
 modrm all sdexec
 modrm all sdbus-sys
 modrm all sdbus
--- a/etc/rc3
+++ b/etc/rc3
@@ -38,6 +38,7 @@ modrm all job-ingest
 
 modrm 0 cron
 modrm all sdexec
+modrm all sdbus-sys
 modrm all sdbus
 modrm all barrier
 
--- a/src/common/libsdexec/list.c
+++ b/src/common/libsdexec/list.c
@@ -66,14 +66,19 @@ bool sdexec_list_units_next (flux_future_t *f, struct unit_info *infop)
 
 /* N.B. Input params:  states=[], patterns=[pattern].
  */
-flux_future_t *sdexec_list_units (flux_t *h, uint32_t rank, const char *pattern)
+flux_future_t *sdexec_list_units (flux_t *h,
+                                  const char *service,
+                                  uint32_t rank,
+                                  const char *pattern)
 {
-    if (!h || !pattern) {
+    if (!h || !pattern || !service) {
         errno = EINVAL;
         return NULL;
     }
+    char topic[256];
+    snprintf (topic, sizeof (topic), "%s.call", service);
     return flux_rpc_pack (h,
-                          "sdbus.call",
+                          topic,
                           rank,
                           0,
                           "{s:s s:[[] [s]]}",
--- a/src/common/libsdexec/list.h
+++ b/src/common/libsdexec/list.h
@@ -32,6 +32,7 @@ struct unit_info {
  * (E.g. use "*" for all).
  */
 flux_future_t *sdexec_list_units (flux_t *h,
+                                  const char *service,
                                   uint32_t rank,
                                   const char *pattern);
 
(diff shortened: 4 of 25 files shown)


grondo:
Wow, nice! I don't have any qualms with using an sdmon.idle broker group for the current implementation.

It seems like eventually we'd want various subsystems to be able to recapture their state from what sdmon has found. (for example the job execution system could reconnect to running jobs after restart, or terminate jobs that are not supposed to be running, the job manager could do something similar for prolog/epilog and housekeeping. Any thoughts on how that might work?  I realize bringing that up is a bit premature, but it could inform the solution here as a stepping stone. (I guess one thought is that as state is able to be recaptured, this would reduce the list of things that prevent a broker from joining `sdmod.idle`)

Also, since the `sdmon.idle` group is never left once joined (except in the case of a broker restart I'm assuming?), I'm wondering if there's a better term. I can't think of anything though, and it doesn't seem important enough to worry about it now (especially since it isn't really user/admin visible)

garlick:
Maybe `sdmon.online` is a better name for the group.  (It started out as an online group that was kept up to date and causes resource to post idle/busy events.  But then I realized it could be a lot simpler and didn't revisit the group name)

garlick:
Renamed the group and fixed a spelling error in a test caught in CI.

This still needs a test for the resource module portion of the proposed change so I'll leave it WIP for the moment.

garlick:
I added the missing test, so I'll drop the WIP.

One thing I should do before we merge this though is make  sure the systemd shipped with RHEL 8 allows sdbus to authetnicate to it with flux credentials.  I'll try to test that on fluke.

garlick:
Yep that worked
```
2025-02-07T16:41:43.412549-08:00 sdbus.info[0]: sd_bus_open_system: connected

```

garlick:
> (I guess one thought is that as state is able to be recaptured, this would reduce the list of things that prevent a broker from joining `sdmod.idle`)

Right, I like that way of thinking about it.

Hmm, we should also be trying to capture the state of any units that have completed but weren't reaped, and put that in a lost+found or something.  I need to refresh my memory on what happens to that state for the cases we're discussing here (the templated system units and transient user units).   That could be a follow-on PR.

But anyway, yeah, if a running unit can be reclaimed, we could let the node join `sdmon.online` before it terminates.

garlick:
Since #6662 is likely not feasible to get into our next release, it seems like this could be considered as is.

garlick:
Oh we should make sure this doesn't cause problems for
- #6652 

I will be offline for today but will double check that soon

grondo (approved):
This one looks good to me. I'm excited by the prospects of the `sdmon` module!

My only comment is that it would be nice if ranks offline due to sdmon could somehow bubble up the errors to a central location (e.g. something like `flux overlay errors`) so admins can easily tell the difference between "offline due to flux not started" vs "offline due to orphaned work". However, I'm not thinking of an easy way to provide that (or maybe there is already a way I'm not seeing) and it certainly shouldn't hold up this PR.

grondo:
> Oh we should make sure this doesn't cause problems for

What problems do you foresee there? The `shrink` should only affect jobs which aren't running `sdmon`, but I'm probably missing something obvious (Just thought maybe I can do some testing today while you're out)

garlick:
Just wanted to double check that this wasn't going to cause a problem:
> -  modify the resource module so it monitors sdmon.idle instead of broker.online when configured to use systemd

grondo:
I did install this PR branch onto my test cluster and performed some quick checks of `shrink` functionality and did not notice any effects. Since batch and alloc jobs won't have `systemd.enable` set, these instance won't even have the `sdmon` module loaded, so as far as `shrink` in the resource acquisition protocol, there's no effect.

I also tested that the resource module notices a rank suddenly disconnecting in the system instance, and this results in the expected behavior in the child. I wasn't really sure what else to check.

garlick:
I think this is OK since `sdmon.online` (formerly `sdmon.idle`) is only joined, never left.

garlick:
Good thought on adding something to `flux resource status`.  Maybe we could somehow annotate brokers that are online but not present in the `sdmon.online` set.  The lack of tools does seem to be the weak link here.

Another thought is to build a tool around `sdmon` that can do things like list orphaned units and signal them.

garlick:
Well, I'll go ahead and set MWP here to keep things moving along.  Thanks!

PR #6617, opened 2025-02-07
ci: ensure unit tests are run during coverage

grondo:
Problem: Unit tests are no longer being run as part of the coverage ci build after 69c1691b6a5f0f1e7ee8d66a152ef22059dc6a1a replaced `make check-prep` with `make check TESTS=`. This is because the former actually ran all unit tests while the latter does not.

Ensure unit tests are executed during the coverage build by adding

 (cd src && make check)

to the set of commands run during coverage.


chu11 (approved):
LGTM

grondo:
Thanks! Setting MWP and we can get our coverage numbers back above 80% :stuck_out_tongue_closed_eyes:

PR #6623, opened 2025-02-11
log helpful info when shutdown is stuck

### LINKED ISSUE #6621: shutdown stuck on elcap due to unresponsive node
Problem: during an elcap shutdown, we observed the rank 0 broker stuck  "Waiting for 10803 peers to shutdown" (per `systemctl status`).  One node in a bad state held up the whole shutdown.

```
# flux module stats overlay
{
 "child-count": 11136,
 "child-connected": 1,
 "parent-count": 0,
 "parent-rpc": 0,
 "child-rpc": 13
}

# flux overlay status | grep full
elcap8395
```

In addition `flux overlay trace` showed the following repeated at about T=10E-4s
```
2025-02-10T10:12:44.937 tx * c disconnect 0 [0]
```

Ultimately, `flux overlay disconnect elcap8395` unstuck the the shutdown.

The admins indicated that this node is suffering from CXI (slingshot) problems that they have observed will lead to hangs.

Edit: the node responded to ICMP pings so was probably not subject to `tbon.tcp_user_timeout`

garlick:
This makes it a little clearer what is happening when shutdown is taking a long time by
- updating the systemd unit status to show the number of peers that are still being waited for
-  logging peers (with hostnames) every 60s when shutdown takes that long

I propose that this fixes #6621 for now even though it doesn't really solve the root cause.  In the case that arose this morning, the admins already knew the hanging node had problems and likely would have powered it down if they had seen it called out in the flux logs.  I'm a little worried about being too heavy handed with timeouts during shutdown since there could be legitimate cases where it might be taking a while, such as if a distributed service is saving state (like the kvs) and the network is congested.

Maybe OK for now?


grondo (approved):
Yeah, this seems like a good solution to me!

garlick:
Setting MWP.

PR #6627, opened 2025-02-12
kvs: more misc cleanups

chu11:
just a bunch of small misc cleanups

PR #6628, opened 2025-02-12
kvs: move "wait_version" API into "kvsroot" API

chu11:
Built on top of #6627 

The "wait_version" convenience functions sit in their own file / API mostly due to legacy of their original implementation from old KVS watchers.  It's only 3 convenience functions operating on kvsroot.

Move them and tests into the "kvsroot" API, therefore removing the kvs_wait_version files and tests.


chu11:
re-pushed and rebased now that #6627 is in

garlick (approved):
Seems reasonable!

PR #6633, opened 2025-02-13
truncate resource journal at configurable size

### LINKED ISSUE #6610: resource eventlog grows without bound in memory
Problem: with #6586 merged, the memory used by the resource journal can grow without bound.  This may be an issue on El Cap.

Before the next release we should think about how to mitigate that on large systems.

grondo:
This is the simplified version of #6631. The in-memory resource eventlog is truncated at the configured limit, but the `truncate` event does not contain any context, it just marks the fact that events have been dropped.

Fixes #6610

### DIFF
--- a/doc/man5/flux-config-resource.rst
+++ b/doc/man5/flux-config-resource.rst
@@ -79,8 +79,14 @@ rediscover
    (optional) If true, force rediscovery of resources using HWLOC, rather
    then using the R and HWLOC XML from the enclosing instance.
 
-Note that updates to the resource table are ignored until the next Flux
-restart.
+journal-max
+   (optional) An integer containing the maximum number of resource eventlog
+   events held in the resource module for the ``resource.journal`` RPC. The
+   default is 100,000. This value takes immediate effect on a configuration
+   update.
+
+Note that, except where noted above, updates to the resource table are
+ignored until the next Flux restart.
 
 EXAMPLE
 =======


src/modules/resource/reslog.c
```diff
@@ -36,7 +36,7 @@ struct reslog {
     struct resource_ctx *ctx;
     zlist_t *pending;       // list of pending futures
     zlist_t *watchers;
-    json_t *eventlog;
+    zlistx_t *eventlog;
```
chu11, line 39: commit message typo "Swtich" and "reosurce"

src/modules/resource/reslog.c
```diff
@@ -217,8 +217,7 @@ int reslog_post_pack (struct reslog *reslog,
     va_end (ap);
     if (!event)
         return -1;
-    if (json_array_append (reslog->eventlog, event) < 0) {
-        json_decref (event);
+    if (!zlistx_add_end (reslog->eventlog, json_incref (event))) {
         errno = ENOMEM;
         return -1;
     }
```
chu11, line 261: if `zlistx_add_end()` fails, I think you gotta call the `json_decref()` b/c you `json_incref()`ed above

src/modules/resource/reslog.c
```diff
@@ -415,16 +422,17 @@ struct reslog *reslog_create (struct resource_ctx *ctx, json_t *eventlog)
...
     if (eventlog) {
         size_t index;
         json_t *entry;
         json_array_foreach (eventlog, index, entry) {
             // historical resource-define events are not helpful
             if (match_event (entry, "resource-define"))
                 continue;
-            if (json_array_append (reslog->eventlog, entry) < 0)
+            if (!zlistx_add_end (reslog->eventlog, json_incref (entry)))
                 goto nomem;
```
chu11, line 489: gotta decref entry if `zlistx_add_end()` failed?

src/modules/resource/reslog.c
```diff
@@ -197,6 +197,43 @@ int reslog_sync (struct reslog *reslog)
     return 0;
 }
 
+/*  Truncate resource journal if needed to reslog->journal_max
```
chu11, line 200: commit message "replace the most recent event" ... perhaps you mean "replace the oldest event"?

t/t2355-resource-journal.t
```diff
@@ -153,7 +153,44 @@ test_expect_success 'last event: resource-define method=kvs' '
...
+	echo resource.journal-max=$limit | flux config load
+'
+test_expect_success 'eventlog is now truncated' '
+	flux resource eventlog -H &&
+	flux resource eventlog -f json > eventlog.trunc
+'
+test_expect_success '1st event is a truncate event' '
+	head -1 eventlog.trunc > eventlog.trunc.1 &&
+	jq -e ".name == \"truncate\"" eventlog.trunc.1
+'
```
chu11, line 177: suggestion: also check number of entries == limit, and likewise with 2 tests below this one

src/modules/resource/reslog.c
garlick: It might be slightly simpler to just define an `entry_duplicator()`.

garlick (approved):
I poked this a bit and LGTM.

src/modules/resource/reslog.c
grondo: Meant to be "replace the most recently dropped event" - not the oldest event in case more than one event is dropped at a time.

grondo:
Uh, I guess I'll set MWP here now.

PR #6640, opened 2025-02-14
doc: expand job termination documentation in `flux-config-exec(5)`

### LINKED ISSUE #6606: flux-config-exec(5)  does not contain much detail about the kill timer
Problem: a question arose about the algorithm used by `job-exec` to kill processes, and [flux-config-exec(5)](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man5/flux-config-exec.html) wasn't super helpful.

Currently we have:
```
      kill-timeout
              (optional) The amount of time to wait after SIGTERM is sent to a
              job before sending SIGKILL.
      term-signal
              (optional) Specify an alternate signal to SIGTERM when terminat‐
              ing job tasks. Mainly used for testing.
      kill-signal
              (optional) Specify an alternate signal to SIGKILL  when  killing
              tasks and the job shell. Mainly used for testing.
```
Some observations
- default value for `kill-timeout` (5s) is missing
- `max-kill-count` is missing entirely (it has a default of 8)
- the kill algorithm isn't described

Maybe on the latter, we could tack on something like

> **JOB TERMINATION**
> 
> When a job is canceled or its time limit is reached, jobs are terminated using the following sequence:
> - The job shells are sent SIGTERM
> - After `kill-timeout`, any remaining shells are sent SIGKILL
> - This continues with an exponential back-off, with `kill-timeout` doubling after each attempt (capped at 300s)
> - After  `max-kill-count` attempts, any nodes still running processes are drained

Just enough to help people choose the right values for the configuration parameters.

grondo:
This PR fixes #6606 by adding the missing config keys, defaults, and key types related to job termination. The suggested `JOB TERMINATION ` section is added as well.

### DIFF
--- a/doc/man5/flux-config-exec.rst
+++ b/doc/man5/flux-config-exec.rst
@@ -36,7 +36,7 @@ job-shell
 
 sdexec-properties
    (optional) A table of systemd properties to set for all jobs.  All values
-   must be strings.  See SDEXEC PROPERTIES below.
+   must be strings. See :ref:`sdexec_properties` below.
 
 kill-timeout
    (optional) The amount of time in FSD to wait after ``SIGTERM`` is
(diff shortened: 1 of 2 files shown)


garlick (approved):
LGTM, thanks!

grondo:
Thanks! I'll set MWP.

PR #6646, opened 2025-02-14
flux-jobs: filter against all users with --include

chu11:
Problem: The --include option is predominantly used for finding jobs on specific hosts.  This implies that the -A option is set (i.e.
filter on all user jobs) but it is currently not.  This can be confusing to users wondering why --include isn't finding any jobs running on specific hosts.

Solution: If the user did not specify filtering on a specific user with --user, then assume all user jobs (i.e. -A or --user=all) will be checked when --include is specified.

---- 

side note, the 1 tricky part was how to tell if the user specified `--user=X` on the command line so we didn't assume "-A" when `--include` was specified.  I setup a new filter action that sets a argparse flag "filtereduser".  There were probably other ways of doing this, but went with this approach.

### DIFF
--- a/doc/man1/flux-jobs.rst
+++ b/doc/man1/flux-jobs.rst
@@ -56,7 +56,8 @@ OPTIONS
    List only jobs where the assigned resources intersect with the supplied
    argument, which may be specified either as an RFC 22 idset of broker ranks
    or an RFC 29 hostlist of host names. It is not an error to specify ranks or
-   hosts which do not exist.
+   hosts which do not exist.  This option implies :option:`-A` unless a specific
+   user has been selected via the :option:`-u, --user` option.
 
 .. option:: -c, --count=N
 
--- a/src/bindings/python/flux/util.py
+++ b/src/bindings/python/flux/util.py
@@ -244,6 +244,13 @@ class FilterActionSetUpdate(argparse.Action):
         getattr(namespace, self.dest).update(values)
 
 
+class FilterActionUser(argparse.Action):
+    def __call__(self, parser, namespace, values, option_string=None):
+        setattr(namespace, "filtered", True)
+        setattr(namespace, "filtereduser", True)
+        setattr(namespace, self.dest, values)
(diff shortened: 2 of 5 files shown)


chu11:
Had one builder fail with an error i've never seen before

```
  expecting success: 
  	test_must_fail_or_be_terminated flux start \
  		-Stbon.interface-hint=badiface \
  		${ARGS} -s2 -Stbon.prefertcp=1 true
  
  flux-broker: could not find address associated with badiface
  flux-start: 0 (pid 470505) exited with rc=1
  flux-start: 0: PMI_Abort(): fatal bootstrap error
  test_must_fail_or_be_terminated: died by signal 13: flux start -Stbon.interface-hint=badiface -Sbroker.rc1_path= -Sbroker.rc3_path= -s2 -Stbon.prefertcp=1 true
  not ok 12 - tbon.interface-hint=badiface fails
  ```
  
  hmmm signal 13 is SIGPIPE.  That's not good

File: doc/man1/flux-jobs.rst [hunk 1]
```diff
@@ -56,7 +56,8 @@ OPTIONS
    List only jobs where the assigned resources intersect with the supplied
    argument, which may be specified either as an RFC 22 idset of broker ranks
    or an RFC 29 hostlist of host names. It is not an error to specify ranks or
-   hosts which do not exist.
+   hosts which do not exist.  This option assumes the :option:`-A` has been
+   specified by the user.
```
grondo, line 60: This wording makes it sound like the option requires the user to also specify `-A` (because of "assumes `-A` has been specified by the user"). Maybe better to write: "This option implies `-A` unless a specific user has been selected via the `-u, --user=USER` option"?

Also, if keeping this wording, drop `the` in front of `:option:-A`.

grondo (approved):
LGTM! I had one suggestion for the documentation inline.

chu11:
thanks, pushed that doc tweak, will set MWP

PR #6647, opened 2025-02-14
doc: add job classifications to glossary

garlick:
Problem: when thinking about grow and shrink, I have to keep looking
up these terms of art: fixed, moldable, malleable, evolving.

Add them to the docs glossary with a citation of the original paper in a footnote.


grondo (approved):
Excellent, thank you!

Issue #6648, opened 2025-02-14
flux-run duplicates output


grondo:
There's been a couple reports of `flux run` (or more correctly `flux job attach`) duplicating output. Here's a simple example from elcap:
```
14:13:28 elcap12127 ~> flux run -N 10 -q pbig -t 1h hostname
flux-job: fcFC1h6HBKu started                                                                                                                 00:00:22
elcap1067
elcap1069
elcap1071
elcap1074
elcap1068
elcap1070
elcap1072
elcap1073
elcap1937
elcap1942
elcap1067
elcap1069
elcap1071
elcap1074
elcap1068
elcap1070
elcap1072
elcap1073
elcap1937
elcap1942

14:14:04 elcap12127 ~> flux run -N 10 -q pbig -t 8h hostname
flux-job: fcFCGKLYG71 started                                                                                                                 00:00:11
elcap1057
elcap1062
elcap1064
elcap1063
elcap1065
elcap1058
elcap1059
elcap1066
elcap1061
elcap1060
```

After this, `flux job attach fcFC1h6HBKu` shows non-duplicated output, so the output in the KVS appears to be correct.

chu11:
I can't help but think something like https://github.com/flux-framework/flux-core/issues/2547 or https://github.com/flux-framework/flux-core/pull/6210 is in play here.  But that should all be shored up.

grondo:
There don't _appear_ to be duplicates in the eventlog in this case, since a subsequent `flux job attach` after the job is complete returns the expected output. I'm wondering if somehow `flux job attach` watches the output eventlog twice, or worse case `job-info` could be sending entry responses more than once? (Wasn't there some recent work in this area?)

I do still need to verify that the eventlog in the job above looks good. I have not been able to reproduce this issue in a test instance or my small home cluster yet.

grondo:
Verified no duplicates in the output eventlog:
```
1739571243.263783 header version=1 encoding={"stdout":"UTF-8","stderr":"UTF-8"} count={"stdout":10,"stderr":10} options={}
1739571243.514828 data stream="stdout" rank="0" data="elcap1067\n"
1739571243.515098 data stream="stderr" rank="0" eof=true
1739571243.515106 data stream="stdout" rank="0" eof=true
1739571243.545506 data stream="stdout" rank="2" data="elcap1069\n"
1739571243.545544 data stream="stderr" rank="2" eof=true
[...]

PR #6649, opened 2025-02-15
fix duplicated output in `flux run`

grondo:
This PR fixes the duplicate output being reported by `flux run`. 

The issue, as described in #6648, is that `flux job attach` will restart watching the output eventlog if the `complete` event from the exec eventlog is received after the output eventlog was fully processed. This is because the code checks for `ctx->output_f == NULL`, but that future is freed and set to `NULL` once the output eventlog has been consumed.

The fix just adds a boolean to track if the watching the output eventlog has started so it is never attempted again. This seemed more straightforward than delaying destruction of the output watch future until the end of the program.

I couldn't really think of a reliable reproducer for this issue, so there is no test.


garlick (approved):
LGTM!

grondo:
Thanks! I've set MWP.

PR #6650, opened 2025-02-18
fix non-ascii character handling in `flux resource eventlog`

grondo:
When adding nicer output options to `flux resource eventlog`, a Python `EventLogFormat` class was developed to handle the human-readable, colorized output of eventlog entries from Python. However, the handling of non-ascii characters was not tested and thus the fact that `json.dumps()` in Python defaults `ensure_ascii` to `True` was missed, and currently all non-ascii characters are escaped.

This PR adds `ensure_ascii=False` to `json.dumps()` where needed to fix the issue.

An event with a utf8 string is also added to the `EventLogFormatter` unit tests for completeness.

### DIFF
--- a/src/bindings/python/flux/eventlog.py
+++ b/src/bindings/python/flux/eventlog.py
@@ -171,7 +171,11 @@ class EventLogFormatter:
                 color = self._color("number")
             else:
                 color = self._color("value")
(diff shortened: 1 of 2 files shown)


garlick (approved):
LGTM - thanks!

PR #6652, opened 2025-02-18
WIP: shrink available resource set for jobs that lose nodes

grondo:
This PR adds support for the RFC 28 `shrink` key to the flux-core `resource` and `sched-simple` modules.

When the resource-discovery method != "configuration", the resource module will include a `shrink` key in the resource acquisition response to the scheduler when `down` is present; including the same set of ranks.

sched-simple is then amended to support the `shrink` key, removing the specified ranks from those available for scheduling. A change to librlist is also required which allows resources that are no longer in the scheduler's resource set to be freed to avoid an error when a job's resources are freed after nodes are removed.

Finally, the `resource.sched-status` RPC is also amended to remove down ranks instead of marking them down for method != "configuration".

This still needs tests, documentation, etc, but I'm posting early to get feedback on the approach.

Example:
```console
$ flux alloc -N4 -o exit-timeout=none --conf=tbon.topo=kary:0 
$ flux resource list
     STATE NNODES NCORES NGPUS NODELIST
      free      4    192     0 corona[212,212,212,212]
 allocated      0      0     0 
      down      0      0     0 
$ flux alloc -N4 --bg
f7GcCriB
$ flux overlay disconnect 3
flux-overlay: asking corona212 (rank 0) to disconnect child corona212 (rank 3)
Feb 18 14:40:15.043538 PST 2025 broker.err[0]: corona212 (rank 3) disconnected by request
Feb 18 14:40:15.143898 PST 2025 broker.err[0]: dead to Flux: corona212 (rank 3)
$ flux job eventlog -HF f7GcCriB
[Feb18 14:40] submit userid=6885 urgency=16 flags=0 version=1
[  +0.011684] validate
[  +0.022306] depend
[  +0.022340] priority priority=16
[  +0.023433] alloc annotations={"sched":{"resource_summary":"rank[0-3]/core[0-47]"}}
[  +0.029100] start
[  +0.219681] memo uri="ssh://corona212/var/tmp/grondo/flux-r5TESH/local-0"
[  +8.608642] exception type="node-failure" severity=2 userid=6885 note="node failure on corona212 (shell rank 3)"
[ +38.639057] exception type="exec" severity=0 note="flux: shell rank 3 on host corona212 exited and exit-timeout=30s has expired" userid=6885
[ +43.651727] finish status=35072
[ +43.652812] release ranks="all" final=true
[ +43.652854] free
[ +43.652868] clean
$ flux resource list
     STATE NNODES NCORES NGPUS NODELIST
      free      3    144     0 corona[212,212,212]
 allocated      0      0     0 
      down      1     48     0 corona212
$ flux run -N4 hostname
0.023s: job.exception fWE7TqLb type=alloc severity=0 unsatisfiable request
```

Note that currently removed resources are still marked `down` in `flux resource list`. That should probably be addressed.


grondo:
This wasn't quite as simple as I'd thought (in retrospect that should have been obvious). I'll take another stab at the changes to the acquire protocol, but no need to look here until that's actually working :facepalm:

grondo:
>  I'm sure you're remembering but based on the wording of the PR description, I was going to double check. A

Yeah, that's what I had meant by "not quite as simple as I thought". In retrospect I should have remembered that. Unfortunately it looks like the acquire subsystem will have to track some "state" in this case, since we actually only want to shrink targets that have transitioned from online->offline -- in the initial acquire response some targets may not have yet joined the instance the first time.

grondo:
Ok, the easiest approach here seemed to be to track "lost" ranks (those that were online but are now offline) in the resource module monitor subsystem. Then, the acquire subsystem can add any new "lost" targets to the `shrink` key when shrinking is enabled. The lost idset can also be used by the status handler to optionally remove the lost ranks from the "all" resource set in the `resource.sched-status` response, so that `flux resource list` matches the resource set the scheduler is using.

garlick:
A quick thought is should we post a `shrink` event to the resource eventlog when a transition to offline causes one?

grondo:
That seems like a pretty good idea. I'll open a PR for addition of that event in RFC 44. 

Were you then envisioning the acquire subsystem acting directly on this event instead of using `monitor_get_lost()`?

garlick:
My high level thought was that there might be other reasons for shrinking an instance besides this one.  For example, if the `exclude` set was changed on the fly.  Possibly I'm getting ahead of myself though and this is OK for now since it solves the immediate problem and is simple.

grondo:
I've added some testing here, including a `resource.monitor-force-down` RPC that can be used in tests with fake resources when the `monitor-force-up` module option has been used (this may be useful in writing Fluxion tests).

I've removed WIP since this is pretty much ready for a review.

src/modules/resource/monitor.c
```diff
@@ -55,6 +55,7 @@ struct monitor {
     struct idset *torpid;
```
garlick, line 55: commit message s/shirnk/shrink/

garlick (approved):
LGTM!  Really nice work here @grondo.

grondo:
Thanks! Setting MWP here.

PR #6657, opened 2025-02-21
testsuite: fix flaky test for handling of invalid `tbon.topo`

### LINKED ISSUE #6655: `t0013` gets killed, only on poodle
Every day in CI, and I verified just now with a fresh install of flux-core, `t0013` seems to cause `SIGKILL` to show up unexpectedly. 

```
poodle17 ~/flux-core/build/t (cli-plugins)$ ../../t/t0013-config-file.t --debug --verbose
...
Killed
```

When I commented this test out, the rest of the tests passed,:

```
test_expect_success 'tbon.topo with unknown scheme fails' '
        mkdir conf23 &&
        cat <<-EOT >conf23/tbon.toml &&
        [tbon]
        topo = "notascheme:42"
        EOT
        test_must_fail flux broker ${ARGS} -c conf23 \
                true 2>badscheme.err &&
        grep "unknown topology scheme" badscheme.err
'
```

I notice writing this that I'm not on `master` (though the nightly CI is), just checked that out and got the same results.

wihobbs:
Problem: on systems where a non-flux libpmi is loaded first, and in some of our CI systems, t0013's test of unknown tbon.topo schemes fails with SIGKILL.

Force the test to run as singletons so libpmi is not used. Fixes #6655

### DIFF
--- a/t/t0013-config-file.t
+++ b/t/t0013-config-file.t
@@ -598,8 +598,8 @@ test_expect_success 'tbon.topo with unknown scheme fails' '
 	[tbon]
 	topo = "notascheme:42"
 	EOT
-	test_must_fail flux broker ${ARGS} -c conf23 \
-		true 2>badscheme.err &&
+	test_must_fail flux broker ${ARGS} -Sbroker.boot-method=single \
+	        -c conf23 true 2>badscheme.err &&
 	grep "unknown topology scheme" badscheme.err
 '
 test_expect_success 'tbon.topo is kary:32 by default' '


grondo (commented):
Great thanks!

Let's wait to hear from @garlick if this change still tests what needs to be tests (I believe so!)

Also, in the commit message let's be explicit that it is the libpmi `PMI_Abort` that is killing the calling process.

garlick (approved):
LGTM!  Thanks @wihobbs !

I think the point of the test was just to check that setting the topology to `notascheme:42` triggered a fatal error.  Your workaround is good!

grondo (commented):
Commit message suggestion: The commit does not change all tests in t0013 to run as singletons, which may be implied by the commit message title. Also it may be good to reference the problem being fixed instead of the solution for those perusing history by commit title.

Suggestion: rename the commit (and PR) subject to

testsuite: fix flaky test for handling of invalid `tbon.topo`

Slight rewording in the commit message solution:

Force the test to run the broker as a singleton so libpmi is not used.

grondo (approved):
LGTM!

wihobbs:
great! Setting MWP.

PR #6659, opened 2025-02-23
add support for undrain reason

### LINKED ISSUE #6599: Suggestion: add the ability to enter an "undrain" message when undraining nodes
In our workflow, current procedure when undraining a node after a repair is to do `flux resource drain -ff foo101 "replaced flux capacitor see ticket bar51"` and then `flux resource undrain foo101` to return the node to service.  To figure out what notes are attached to a node being undrained, we have to find the undrain time and then backtrack.

Instead, it would be useful to be able to add a message to a la `flux resource undrain foo101 "replaced flux capacitor see ticket bar51"` and then have that message available in the log coupled to the undrain event.

grondo:
This PR adds support for an undrain reason. Currently there is not a utility to display the undrain reason, it is just entered into the resource eventlog. Therefore, to make the undrain reason a bit more useful, a new `-i, --include=TARGETS` option is added to `flux resource eventlog` so admins can easily see the "history" of a node or set of nodes. (Actually, this could be moved to  a separate PR if preferred.)

Fixes #6599

### DIFF
--- a/doc/man1/flux-resource.rst
+++ b/doc/man1/flux-resource.rst
@@ -17,7 +17,7 @@ SYNOPSIS
 
 | **flux** **resource** **drain** [*-n*] [*-o* *FORMAT*] [*-i* *TARGETS*]
 | **flux** **resource** **drain** [*-f*] [*-u*] [*targets*] [*reason*]
-| **flux** **resource** **undrain** [*-f*] *targets*
+| **flux** **resource** **undrain** [*-f*] *targets* [*reason*]
 
 | **flux** **resource** **reload** [-f] [--xml] *path*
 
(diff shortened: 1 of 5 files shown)


doc/man1/flux-resource.rst
```diff
@@ -396,6 +396,12 @@ Watch the resource journal, which is described in RFC 44.
   The current set of valid events events is *restart*, *resource-define*,
   *online*, *offline*, *drain*, *undrain*, *torpid*, and *lively*.
 
+.. option:: -i, --include=TARGETS
+
+  Filter events to only include those that apply to  *TARGETS*, which may be
+  specified either as an idset of broker ranks or list of hosts or nodes in
+  hostlist form.
+
```
garlick, line 404: Might want to indicate a preference for hosts, given the caveat about historical reranking.  Maybe
> , which is normally one or more hostnames in RFC 29 hostlist form.  *TARGETS* can also be broker ranks in RFC 22 idset form, useful mainly in test.

garlick (approved):
LGTM!  Just one suggestion on the  `flux-resource(1)1 wording.

doc/man1/flux-resource.rst
grondo: Fixed that and will set MWP.

Issue #6656, opened 2025-02-21
multiuser job with unkillable tasks causes job to hang indefinitely in RUN state


grondo:
This is actually an combination flux-core/flux-security issue.

In https://github.com/flux-framework/flux-security/pull/187, the IMP was modified to wait until the job cgroup is empty before it exits. This is supposed to ensure that all job tasks have exited before starting the epilog, entering CLEANUP, etc.

However, there is one big drawback to this approach. When there is an unkillable process and IMP lingers waiting for the cgroup to be empty, the job execution  system thinks the job is still running, since it is waiting for the IMP to exit. This means that there is no indication that there is a problem with job cleanup until the job hits its timeout.

Ideally, there would be some way to tell the job execution system that the main process in the job has exited and it is in the process of trying to exit, so that users and admins are aware that there may be a problem.

grondo:
Part of the question here is whether a job should be expected to post a `finish` event once all user tasks have exited, or should that be delayed until all processes created by the job have gone away. With the current IMP configuration, Flux is doing the second thing.

This does run counter some of the statements in documentation, such as from flux-alloc(1):

> When COMMAND exits, the Flux subinstance exits, resources are released to the enclosing Flux instance, and flux alloc returns.

In the case here, the script passed to `flux alloc` terminated, but the job continues in the RUN state until timeout because of some unkillable processes that were actually not even direct descendants of the initial program (they are leftover processes invoked by `flux run` which eventually timed out). I can sympathize with the user that questions why their job did not terminate when COMMAND did.

This is also a difference in behavior between multiuser jobs, in which the `finish` event is delayed until the cgroup is empty, and Flux jobs in all other situations, where the `finish` event is posted as soon as the job shell exits, which occurs when all tasks have exited.

This leads me to prefer a solution that has the IMP either exit after some attempts to SIGKILL all processes in the cgroup, and defer waiting on the cgroup in housekeeping or some other place that allows partial release, or somehow notify sdexec that the "main" user tasks have exited and have it notify the job exec module.

Either way, the transient unit should probably stick around while there are active user processes, perhaps including the IMP, though that doesn't seem necessary since housekeeping also runs with privilege. I'm not sure if this is a feasible approach.

garlick:
> Either way, the transient unit should probably stick around while there are active user processes, perhaps including the IMP, though that doesn't seem necessary since housekeeping also runs with privilege. I'm not sure if this is a feasible approach.

[...]

PR #6661, opened 2025-02-24
improve management of systemd stop timeout to avoid SIGKILL while making progress

garlick:
Problem: systemd imposes a 90s fixed timeout on shutdown when flux is stopped with `systemctl stop flux`, but that is not always enough time to cleanly stop flux on a big system.  Data loss likely results if the timeout runs out and the broker is killed.

This PR causes the stop timeout to be reset each time the broker transitions into one of the CLEANUP, SHUTDOWN, FINALIZE, or GOODBYE states.  It also coordinates all stop time extension through the broker state machine rather than having `sd_notify()` calls in several locations.

As a result, each of those states can take up to 90s - actually more for CLEANUP and FINALIZE (rc3) where the timeout is reset as progress is made within the state.

This is just a first step toward making `systemctl stop flux` more robust.  In the future we'll need to add some more state timeouts (SHUTDOWN in particular).


doc/man7/flux-broker-attributes.rst
```diff
@@ -138,6 +138,15 @@ broker.exit-restart [Updates: C, R]
...
    Timestamp of broker startup from :man3:`flux_reactor_now`.
 
+broker.sd-notify
+   A boolean indicating that the broker should use :linux:man3:`sd_notify`
+   to inform systemd of its status.  This is set in the Flux systemd unit file.
+
+broker.sd-stop-timeout
+   The systemd TimeoutStopSec value (in RFC 23 Flux Standard Duration format),
+   which is used by the broker to extend the stop timeout while it is making
+   progress towards shutdown.  This is set in the Flux systemd unit file.
```
grondo, line 148: Does this mean `broker.sd-stop-timeout` needs to be the same as `TimeoutStopSec`? If so, it might be clearer to just explicitly state that in this documentation. Maybe just state what the broker does with the value (which I can't remember even though I just reviewed the code :facepalm: ), then state "This value should be set equal to the `TimeoutStopSec` set for the `flux.service` unit." or similar.

grondo (approved):
LGTM! Looks like a nice improvement. One take it or leave it comment inline.

garlick:
Thanks, made the suggested changes to the man page.

Also, while experimenting with
- #6656 

I wasted a couple hours (ok I was being dense) tracking down the fact that NOTIFY_SOCKET had leaked from the broker's environment to IMP.  So I tacked on some commits to prevent inheritance of this environment variable from the broker and (because we can now) restrict notifications to the main process.  LMK if this still looks good @grondo.

grondo:
Yeah, looks good!

PR #6664, opened 2025-02-26
gitignore: ignore PLY autogenerated parsetab.py

sam-maloney:
Problem: `src/bindings/python/flux/constraint/parsetab.py` file is autogenerated by `python3-ply` but not ignored by git

Add `parsetab.py` to `src/bindings/python/flux/.gitignore`

I believe the file is only generated during for testing (at least on my Rocky 9.4 dev system) , but it is constantly annoying me in git; sorry for the really minor PR ;)


grondo (approved):
Thanks! LGTM!

grondo:
I poked the asan build which timed out for some reason and I've set MWP. Thanks again!

PR #6666, opened 2025-02-26
sdexec: add stop timeout to handle unkillable processes

garlick:
Problem: in a system instance using sdexec, jobs may remain in R state when the shell exits with unkillable processes.

This is because the IMP doesn't exit until the unit cgroup is empty, and the IMP must exit for sdexec to collect the shell's exit status.

This PR, in conjunction with flux-framework/flux-security#201, arranges for the unit to enter deactivating state once the shell exits.  An sdexec timeout begins at that point.  If after a configurable period, the unit has not stopped, SIGUSR1 (IMP proxy for SIGKILL) is sent to the unit.  If after another period the unit still has not stopped, it is abandoned and the subprocess protocol is terminated so  the job can advance and release resources.

The abandoned unit remains in deactivating state indefinitely, and can be picked up by the `sdmon` montoring service proposed in #6616

### DIFF
--- a/doc/man5/flux-config-exec.rst
+++ b/doc/man5/flux-config-exec.rst
@@ -38,6 +38,21 @@ sdexec-properties
    (optional) A table of systemd properties to set for all jobs.  All values
    must be strings. See :ref:`sdexec_properties` below.
 
+sdexec-stop-timer-sec
+   (optional) Configure the length of time in seconds after a unit enters
+   deactivating state when it will be sent the sdexec-stop-timer-signal.
+   Deactivating state is entered by ``imp-shell`` units when the
+   :man1:`flux-shell` terminates.  The unit may remain there as long as
+   user processes remain in the unit's cgroup.
+
+   After the same length of time, if the unit hasn't terminated, for example
+   due to unkillable processes, the unit is abandoned and the node is drained.
+   Default 30.
+
+sdexec-stop-timer-signal
+   (optional) Configure the signal used by the stop timer.  By default,
+   10 (SIGUSR1, the IMP proxy for SIGKILL) is used.
+
 kill-timeout
    (optional) The amount of time in FSD to wait after ``SIGTERM`` is
    sent to a job before sending ``SIGKILL``. The default is "5s". See
--- a/configure.ac
+++ b/configure.ac
@@ -433,7 +433,7 @@ AS_IF([test x$enable_code_coverage = xyes], [
 AC_ARG_WITH([flux-security], AS_HELP_STRING([--with-flux-security],
              [Build with flux-security]))
 AS_IF([test "x$with_flux_security" = "xyes"], [
-    PKG_CHECK_MODULES([FLUX_SECURITY], [flux-security >= 0.13.0],
+    PKG_CHECK_MODULES([FLUX_SECURITY], [flux-security >= 0.14.0],
       [flux_sec_incdir=`$PKG_CONFIG --variable=includedir flux-security`])
     AS_IF([test "x$flux_sec_incdir" = x],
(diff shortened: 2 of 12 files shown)


garlick:
Ugh I just realized this needs to drain any node that ends up with an abandoned stuck unit.

garlick:
OK, pushed a change that handles the stop timeout speficially rather than just letting job-exec treat it like a generic failure.  Here is a test where the `sdexec-stop-timer-signal` was set to 0 and the  shell was killed with -9 to simulate a stuck process.    
```
$ flux run -t 10m nohup sleep inf
flux-job: ƒBd6cBBSFcK started                                          00:00:22
53.647s: job.exception ƒBd6cBBSFcK type=exec severity=0 shell exited with unkillable processes on picl0 (rank 0)
```
and `flux resource drain` reports
```
Feb27 10:51  drained  unkillable processes from job+ picl0
```
The unit is still running
```
● imp-shell-0-fBd5swxvjPd.service - User workload
     Loaded: loaded (/run/user/500/systemd/transient/imp-shell-0-fBd5swxvjPd.service; transient)
  Transient: yes
     Active: deactivating (stop-sigterm) since Thu 2025-02-27 10:27:07 PST; 32min ago
   Main PID: 157097 (flux-imp)
     Status: "IMP child Killed, waiting for cgroup"
      Tasks: 2 (limit: 9562)
     Memory: 1.6M (max: 6.2G available: 6.2G)
        CPU: 111ms
     CGroup: /user.slice/user-500.slice/user@500.service/app.slice/imp-shell-0-fBd5swxvjPd.service
             ├─157097 /usr/lib/aarch64-linux-gnu/flux/flux-imp exec /usr/lib/aarch64-linux-gnu/flux/flux-shell 78897629785751552
             └─157102 sleep inf
```

doc/man5/flux-config-exec.rst
```diff
@@ -38,6 +38,17 @@ sdexec-properties
    (optional) A table of systemd properties to set for all jobs.  All values
    must be strings. See :ref:`sdexec_properties` below.
 
+sdexec-stop-timer-sec
+   (optional) Configure the length of time in seconds after a unit enters
+   deactivating state when it will be sent the sdexec-stop-timer-signal.
+   After the same length of time, if the unit hasn't terminated, for example
+   due to unkillable processes, the unit will be abandoned and the exec RPC
+   is terminated.  Default 30.
```
grondo, line 46: Suggestion: Remind the reader here what triggers the unit entering deactivating state.

grondo (approved):
This LGTM! Nice improvement.

I will get a PR to update flux-security to v0.14.0 in CI.

grondo:
> I will get a PR to update flux-security to v0.14.0 in CI.

Actually that's just a one liner in `docker-run-checks.sh`,  and `debian/control` and `configure.ac` should also be updated when v0.14.0 is required, so I'll just let you do that in whatever PR is appropriate.

doc/man5/flux-config-exec.rst
garlick: thanks, fixed that and forced a push.

I forgot that I didn't add tests for the SDEXEC_TIMER stuff.  However, if you are OK with merging this as is I'm fine with that and time permitting, I can follow on with some test improvements in another PR.

garlick:
> Actually that's just a one liner in docker-run-checks.sh, and debian/control and configure.ac should also be updated when v0.14.0 is required, so I'll just let you do that in whatever PR is appropriate.

That should probably go here actually. I 'll tack that on.

grondo:
Yeah, it kind of seems reasonable to merge this one as is to keep things moving along.

garlick:
Great, thanks!  Setting MWP then.

PR #6669, opened 2025-02-27
raise non-fatal exception on epilog failure

grondo:
This PR adds a missing exception (non  fatal) to the job eventlog on epilog failure.

### DIFF
--- a/src/modules/job-manager/plugins/perilog.c
+++ b/src/modules/job-manager/plugins/perilog.c
@@ -352,63 +352,70 @@ static void perilog_proc_delete (struct perilog_proc *proc)
     }
 }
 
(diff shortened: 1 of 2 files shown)


garlick (approved):
LGTM!

I think you could argue that an epilog failure should be a fatal job exception. For example, if an epilog action is ensuring that job output is on stable storage.

However, I'm sure there are arguments both ways and this is certainly a step in the right direction!

grondo:
> I think you could argue that an epilog failure should be a fatal job exception. For example, if an epilog action is ensuring that job output is on stable storage.

I did consider that, however it seemed like that is not the common case, and the fatal epilog exception would mask the actual exit code of the job itself. Maybe this should somehow be controlled by the epilog itself or configuration in the future.

PR #6673, opened 2025-02-28
configure: relax flux-security version check

### LINKED ISSUE #6672: flux-core can't be built with flux-security < 0.14.0, but this is only necessary for a system instance
Current flux-core master can't be built on TOSS 4 systems with flux-security because there's a hard requirement for v0.14.0 or greater, but it is useful to build test versions with flux-security so commands can submit jobs to the system instance. Since v0.14.0 is only required for the sd_notify() bit in the IMP, it seems we could relax this hard requirement and add a runtime version check in sdexec at module load time.

grondo:
This PR relaxes the flux-security version check from v0.14.0 back to v0.13.0 since v0.14.0 is only required for the sdexec module when used to launch jobs in a system instance.

A test is added to the sdexec module to abort if a sufficient flux-security version isn't found at runtime, and a version check added to affected tests.

This allows flux-core to continue to be built on TOSS systems with flux-security v0.13.0 installed. We'll probably have to run with this version for a month at least, and it is nice to be able to build test versions of flux `--with-flux-security`.

Fixes #6672

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -433,7 +433,7 @@ AS_IF([test x$enable_code_coverage = xyes], [
 AC_ARG_WITH([flux-security], AS_HELP_STRING([--with-flux-security],
              [Build with flux-security]))
 AS_IF([test "x$with_flux_security" = "xyes"], [
-    PKG_CHECK_MODULES([FLUX_SECURITY], [flux-security >= 0.14.0],
+    PKG_CHECK_MODULES([FLUX_SECURITY], [flux-security >= 0.13.0],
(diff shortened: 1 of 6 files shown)


chu11 (approved):
LGTM!

grondo:
Yay! Thanks!

grondo:
@merfiyio requeue

grondo:
@mergifyio requeue

grondo:
@mergifyio refresh

garlick:
Oh duh, I should have anticipated that problem, sorry.

PR #6676, opened 2025-03-01
resource: only mark nodes lost that transition from `online` to `offline` 

### LINKED ISSUE #6675: resource: can a rank be "lost" when it becomes un-torpid?
Problem: the resource monitoring hook for removing a node that goes offline from the resource set(#6652)  was added to a function, `post_join_leave()` that is called for multiple event types, but the event type is not checked, so I _think_ a transition from torpid->lively  would have the same effect as online->offline.

Sorry, I created the conditions that led to this by trying too hard to avoid code duplication at the expense of clarity.

grondo:
This PR fixes #6675 and adds a set of tests to ensure torpid/lively events don't cause resource removal (shrink).

Thanks for finding this one @garlick!

### DIFF
--- a/src/modules/resource/monitor.c
+++ b/src/modules/resource/monitor.c
@@ -202,11 +202,12 @@ static int post_join_leave (struct monitor *monitor,
         goto error;
 
     /* Update the set of lost ranks. These are only the ranks that have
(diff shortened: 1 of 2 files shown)


garlick (approved):
LGTM!

Issue #6644, opened 2025-02-14
job epilog should be able to time out


grondo:
Problem: A job epilog (as executed by the `perilog.so` plugin) doesn't currently support a timeout. The [documentation](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man5/flux-config-job-manager.html#perilog-so) notes that `timeout` and `kill-timeout` are supported for the epilog, but not suggested except for testing (and I can't remember why that is) Furthermore, these values were configured on the affected system and the epilog was not timed out.

grondo:
Hm, in this case the hung epilog got:
```
[ +21.225638] job-manager.err[0]: housekeeping: elcap3776 (rank 3472) fc4eAer8a6s: No route to host
```
But reported the epilog still running on that rank long after the timeout should have occurred:
```
  "procs": {
    "fc4eAer8a6s": {
      "name": "epilog",
      "state": "running",
      "total": 8196,
      "active": 1,
      "active_ranks": "3472",
      "remaining_time": 1800
    }
  }
```

Note the remaining time is 1800, the configured timeout, which from the code implies the kill timer is not armed and the timer hasn't yet started(?).

Ah, the timer is not started until the bulk-exec `start` callback is called. Perhaps this doesn't happen if there are errors starting some tasks. This could be a bug in bulk-exec that leaves the bulk execution in a state where it can never finish because it didn't start.

grondo:
Just noticed above that I captured the `housekeeping` error not the `epilog` error. 🤦
One possible explanation is that the execution of the epilog did not hit an error on this rank but instead hung. This could explain why the rank appears in `active_ranks` but the timer hasn't started.

One obvious improvement would be to start the prolog/epilog timeout timer as soon as `bulk_exec_start()` is called, instead of waiting for the `on_start` callback, which is not called until all remote processes have been started. This should handle a case where there is no remote error, but also the remote can't return the initial response to `flux_rexec()` for some reason.

PR #6677, opened 2025-03-03
fix prolog/epilog timeout handling

grondo:
This PR adds a probable fix for #6644 (epilog doesn't time out), plus some other bits of cleanup.

### DIFF
--- a/doc/man5/flux-config-job-manager.rst
+++ b/doc/man5/flux-config-job-manager.rst
@@ -110,28 +110,31 @@ prolog
    supported for the ``[job-manager.prolog]`` table:
 
    command
-      (optional) An array of strings specifying the command to run. If
+      (optional, string) An array of strings specifying the command to run. If
       ``exec.imp`` is set, the the default command is ``["flux-imp",
       "run", "prolog"]``, otherwise it is an error if command is not set.
    per-rank
-      (optional) By default the job-manager prolog only runs ``command``
+      (optional, bool) By default the job-manager prolog only runs ``command``
       on rank 0. With ``per-rank=true``, the command will be run on each
       rank assigned to the job.
    timeout
-      (optional) A string value in Flux Standard Duration specifying a
+      (optional, string) A string value in Flux Standard Duration specifying a
       timeout for the prolog, after which it is terminated (and a job
       exception raised). The default prolog timeout is 30m. To disable
       the timeout use ``0`` or ``infinity``.
    kill-timeout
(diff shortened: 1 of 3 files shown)


doc/man5/flux-config-job-manager.rst
```diff
@@ -110,28 +110,31 @@ prolog
...
       exception raised). The default prolog timeout is 30m. To disable
       the timeout use ``0`` or ``infinity``.
    kill-timeout
-      (optional) If a job exception is raised during the job prolog and
-      ``cancel-on-exception`` is true,  the prolog will be canceled by
-      sending it a SIGTERM signal. ``kill-timeout`` is the number of
-      seconds to wait until any nodes with prolog tasks that are still
-      active will be drained. The drain reason will include the string
-      "canceled then timed out". The default is 60.
+      (optional, float) If a the prolog times out, or a job exception is raised
```
chu11, line 126: both here and in epilog, "If a the" -> "If the"?

doc/man5/flux-config-job-manager.rst
```diff
@@ -110,28 +110,31 @@ prolog
...
+      is a floating point number of seconds to wait until any nodes with
+      prolog tasks that are still active will be drained. The drain reason
+      will include the string "timed out" if the prolog timeout was reached,
+      or "canceled then timed out" if the prolog was canceled after a job
+      exception then timed out. The default is 60.
    cancel-on-exception
-      (optional) A boolean indicating whether a fatal job exception raised
-      while the prolog is active terminates the prolog. The default is true.
+      (optional, bool) A boolean indicating whether a fatal job exception is
+      raised while the prolog is active terminates the prolog. The default
```
chu11, line 136: both here and in epilog, first sentence seems off to me with the "while the prolog is active terminates the prolog".  Perhaps reorder the sentence,  "A boolean indicating if a prolog should be terminated if a fatal job exception is raised while the prolog is active."?

chu11 (approved):
i don't know the perilog too well but seems to make sense to me.  Just some English stuff in the docs.

doc/man5/flux-config-job-manager.rst
grondo: Oops, this was an edit issue - two versions of the sentence remain grafted together here. Anyway, I like your suggestion so let's use that.

grondo:
Thanks! I've fixed the wording as suggested and will set MWP.

Issue #6626, opened 2025-02-11
1000 elcap nodes are shown as "lost connection" on rank 0 but compute node thinks it is still connected


garlick:
Problem: about 1000 nodes of el cap got into a state where the rank 0 broker thought they were disconnected, but from the point of view of the nodes, the rank 0 broker was just unresponsive.

Specifically, on rank 0
```
[root@elcap1:conf.d]# flux overlay status |grep elcap1124
├─ 820 elcap1124: lost lost connection
```
but on the node, everything seemed OK except flux commands that needed to contact rank 0 would hang.

The  actual TCP connection to the node appeared to be in connected state.  Here is elcap12119 (another node in that state):
```
tcp        0 138488 eelcap12119:46878       eelcap1:8050            ESTABLISHED 188356/broker
```
and the same connection on elcap1
```
tcp   246536      0 eelcap1:8050            eelcap12119:46878       ESTABLISHED 3180474/broker
```

Stopping flux on the compute node ran into the systemd timeout, but problems immediately went away upon restart.

garlick:
Note that the "lost connection" status indicates that a send on rank 0 failed with EHOSTUNREACH.

Because we have ZMQ_ROUTER_MANDATORY set[1], if the destination UUID is unknown (presumably because it disconnected), a `zmq_send()` fails with EHOSTUNEACH.  We catch that and set SUBTREE_STATUS_LOST  with the error "lost connection".

[1] https://libzmq.readthedocs.io/en/latest/zmq_setsockopt.html

garlick:
Hmm, we have tuned TCP_USER_TIMEOUT (see [tcp(7)](https://www.man7.org/linux/man-pages/man7/tcp.7.html) to 2m on these systems
```toml
[tbon]
tcp_user_timeout = "2m"
```
Given that we have untransmitted data in the socket queue on both ends I would have expected the kernel to have forcibly  closed these connections.

grondo:
This issue occurred again today on elcap. Again ~1000 nodes affected.

garlick:
The thoughts I had on this one are
- the above heartbeat timeout probably makes sense to have as an option regardless
- maybe there is a way to get zmq debugging enabled all the time without log spam.  For example, create the monitor socket but don't connect to it?  Or connect to it and keep the logs in a small circular buffer until logging is enabled?
- I vaguely was wondering about the short zmq socket queue depths on el cap could somehow cause deadlock at the zmq level but that probably does not make sense

[...]

PR #6679, opened 2025-03-03
heartbeat: add optional timeout

garlick:
As discussed in #6626, this adds a timeout option to the heartbeat module.

- heartbeat is loaded on all ranks
- heartbeat on leader rank publishes heartbeats, as before
- heartbeat on follower ranks subscribes to heartbeats
- if a heartbeat does not arrive within a configurable timeout, follower logs a LOG_CRIT message and calls `_exit(1)`.
- the timeout value and whether to merely log or log+exit is configurable

My initial inclination was to just log (no exit) on timeout by default, but I wasn't sure this would get used if we didn't enable it by default, so in the first proposal it is enabled, the default is to log+exit after a 5m timeout.  I'm happy to change that if it seems inappropriate.  Since it is enabled by default, it applies to all instances including user allocations on el cap that are using the slingshot. :shrug:

### DIFF
--- a/doc/man5/flux-config-heartbeat.rst
+++ b/doc/man5/flux-config-heartbeat.rst
@@ -1,0 +1,48 @@
+========================
+flux-config-heartbeat(5)
+========================
+
+
+DESCRIPTION
+===========
+
+The ``heartbeat`` table may be used to tune the configuration of the
+Flux heartbeat module, which publishes periodic ``heartbeat.pulse`` messages
+for synchronization.
+
+It may contain the following keys:
+
+
+KEYS
+====
+
+period
+   (optional) The interval (in RFC 23 Flux Standard Duration format) between
+   the publication of heartbeat messages.  Default: ``"2s"``.
+
+
+EXAMPLE
+=======
+
+::
+
+   [heartbeat]
+   period = "5s"
+
+
+RESOURCES
+=========
+
+.. include:: common/resources.rst
+
+
+FLUX RFC
+========
+
+:doc:`rfc:spec_23`
+
+
+SEE ALSO
+========
+
+:man5:`flux-config`
--- a/doc/man5/flux-config-heartbeat.rst
+++ b/doc/man5/flux-config-heartbeat.rst
@@ -6,20 +6,30 @@ flux-config-heartbeat(5)
 DESCRIPTION
 ===========
 
-The ``heartbeat`` table may be used to tune the configuration of the
-Flux heartbeat module, which publishes periodic ``heartbeat.pulse`` messages
-for synchronization.
+The Flux heartbeat service publishes periodic ``heartbeat.pulse`` messages
+from the leader broker for synchronization.  Follower brokers subscribe
+to these messages and may optionally force a disconnect from their overlay
+network parent when they are are missed for a configurable period.
 
-It may contain the following keys:
-
+The ``heartbeat`` table may be used to tune the heartbeat service.  It may
+contain the following keys:
 
 KEYS
 ====
 
 period
    (optional) The interval (in RFC 23 Flux Standard Duration format) between
-   the publication of heartbeat messages.  Default: ``"2s"``.
+   the publication of heartbeat messages.  Default: *2s*.
 
+timeout
+   (optional) The period (in RFC 23 Flux Standard Duration format) after
+   which a follower broker will forcibly disconnect from its overlay network
+   parent if it hasn't received a heartbeat message.  Set to *0* or *infinity*
+   to disable.  Default: *5m*.
+
+warn_thresh
+  (optional) The number of missed heartbeat periods after which a warning
+  message will be logged.  Default: 3.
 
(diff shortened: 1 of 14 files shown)


garlick:
oops fixed a spelling error caught by the typo checker.  I'll do some "real" testing on my test cluster after lunch.

garlick:
Pondering this a bit more after our :coffee: call, I think instead of `_exit(1)`, this should call a new RPC in the overlay code that forces a parent disconnect.  That pinpoints the source of the problem to the parent overlay connection, and invokes the same error path as various other scenarios where the parent decides to force the child to disconnect.  (Such as when a child silently reconnects after a network partition in which the parent broker has already declared the child lost)

garlick:
OK, I pushed those changes.  I have to step away for now but I'll try to write a test for that part.

garlick:
OK, a few tweaks this morning to tidy things up.  The config is now
```toml
 [heartbeat]
period = "5s"
timeout = "5m" # set to 0 or infinity to disable
warn_thresh = 3 # number of missed heartbeat periods before logging a warning
```
Logging (always enabled) just logs a warning  once whenr `warn_thresh` is execeeded, and once again when a heartbeat is received.

There's an expanded description including some background in the `flux-config-heartbeat(5)` man page now.

etc/rc1
```diff
@@ -86,7 +86,7 @@ fi
 
 modload all job-ingest
 modload 0 job-exec
-modload 0 heartbeat
+modload all heartbeat
```
grondo, line 89: The statement in the commit message "the new heartbeat timer requires it to be loaded on all ranks" confused me. Is it actually a future commit that will require it be to loaded on all ranks, or should the message read something like "is supported on all ranks"?

doc/man5/flux-config-heartbeat.rst
```diff
@@ -28,7 +38,36 @@ EXAMPLE
...
+Heartbeats may be used to synchronize Flux activities across brokers to
+reduce the operating system jitter that affects some sensitive bulk-synchronous
+applications.  :man3:`flux_sync_create` provides a way to invoke work that
+is synchronized with the heartbeat.
+
+.. note::
+  The efficacy of heartbeats to mitigate noise is limited by the propagation
+  delay of published messages through the tree based overlay network;  however,
+  this may be reduced in the future with a side channel transport such as
+  TCP multicast, hardware collectives, or quantum entanglement.
```
grondo, line 56: A future quantum entanglement method for synchronization should definitely get an entry on our flashy presentation!

grondo: Also: probably add to commit message that this commit adds a USE CASES section as well as documenting the heartbeat timeout configuration.

grondo (approved):
This LGTM!

etc/rc1
garlick: I'll reword.  I meant the future commit.

doc/man5/flux-config-heartbeat.rst
garlick: It says "future" so I figured what the hell :-)

The use case section seemed slightly incomplete to me.  Does this change help or hurt?  (Gah, I've read it too many times now)
```diff
@@ -61,13 +61,20 @@ loses the TCP connection to its overlay parent without a shutdown (for example,
 if the parent crashes or there is a network partition and TCP times out),
 ZeroMQ tries indefinitely to re-establish the connection without informing
 the broker.  The child broker remains in RUN state with any upstream RPCs
-blocked.
-
-A heartbeat timeout forces the broker to fail those RPCs and shut down its
-overlay subtree and exit.  In a system instance, systemd restarts the broker
-which will remain in JOIN state until the parent returns to service.
-Otherwise, the outcome depends on whether or not the broker is one of
-the *critical ranks* described in :man7:`flux-broker-attributes`.
+blocked until the parent returns to service, after which it is forced to
+disconnect and shut down, which causes the RPCs to fail.  A heartbeat timeout
+forces the broker to "fail fast", with the same net effect, but arriving
+at a steady state sooner.
+
+The effect of a follower broker shutdown depends on its role.  If it is
+not a leaf node, the effect applies to its entire subtree.  In a system
+instance, systemd restarts brokers that shut down this way.  Upon restart,
+the brokers remain in JOIN state until the parent returns to service.
+The heartbeat service is not loaded until after the parent connection is
+established, so heartbeat timeouts do not apply in this phase.  In a user
+allocation where brokers are not restarted, the outcome depends on whether
+or not the broker is one of the *critical ranks* described in
+:man7:`flux-broker-attributes`.
 
 RESOURCES
 =========
```

Or maybe I've gone off the rails adding stuff that shouldn't be in a config man page :-)

garlick:
Thanks!  I'll set MWP.

grondo:
@garlick, the new typos checker found one typo here. I'll just push a fix if that's ok?

garlick:
Yep!

PR #6680, opened 2025-03-03
improve error message on invalid formatting of shell options

### LINKED ISSUE #6678: Improve cli submission error message when `-o key` is used with `-o key.subkey=val`
@rountree noted this morning that when a "enable/disable" option is used in conjunction with a sub-key, the error raised by Python (and therefore flux) is pretty unhelpful. Example:

> I’m getting a new-to-me error on tuolumne when using Spindle:
`flux-run:  ERROR: 'int' object does not support item assignment`
This invocation works as expected:
`flux run -o userrc=${FLUXRC} -o spindle --tasks-per-node=3 --nodes=3 /bin/true`
Adding another Spindle parameter causes the error.
`flux run -o userrc=${FLUXRC} -o spindle -o spindle.python-prefix=/usr/bin --tasks-per-node=3 --nodes=3 /bin/true`
Replacing the new option with -o spindle.nostrip fails the same way, so it’s not (just) a parameter assignment issue. 

After some digging I found that this is because `-o spindle` sets `attributes.system.shell.spindle` equal to 1 in the jobspec, and then the sub-key `-o spindle.python-prefix` thinks it can access `python-prefix` as a key in the `attrobutes.system.shell.spindle Dict`, which is really an integer. Partial traceback:
```
(s=1153,d=0)  tuolumne1004 ~ $ FLUX_PYCLI_LOGLEVEL=5 flux run -o userrc=${FLUXRC} -o spindle \
> -o spindle.python-prefix=/usr/bin --tasks-per-node=3 --nodes=3 --dry-run /bin/true 
flux-run: ERROR: 'int' object does not support item assignment
flux-run: DEBUG: Traceback (most recent call last):
  File "/usr/lib64/flux/python3.6/flux/util.py", line 307, in __call__
    main_func()
  File "/usr/libexec/flux/cmd/flux-run.py", line 37, in main
    args.func(args)
  File "/usr/lib64/flux/python3.6/flux/cli/run.py", line 40, in main
    jobid = self.submit(args)
  File "/usr/lib64/flux/python3.6/flux/cli/base.py", line 1178, in submit
    return self.submit_async(args, jobspec).get_id()
  File "/usr/lib64/flux/python3.6/flux/cli/base.py", line 1135, in submit_async
    jobspec = self.jobspec_create(args)
  File "/usr/lib64/flux/python3.6/flux/cli/base.py", line 1089, in jobspec_create
    jobspec.setattr_shell_option(key, val)
  File "/usr/lib64/flux/python3.6/flux/job/Jobspec.py", line 569, in setattr_shell_option
    self.setattr("system.shell.options." + key, val)
  File "/usr/lib64/flux/python3.6/flux/job/Jobspec.py", line 563, in setattr
    set_treedict(self.jobspec, key, val)
  File "/usr/lib64/flux/python3.6/flux/util.py", line 203, in set_treedict
    set_treedict(in_dict.setdefault(path[0], {}), path[1], val)
  File "/usr/lib64/flux/python3.6/flux/util.py", line 203, in set_treedict
    set_treedict(in_dict.setdefault(path[0], {}), path[1], val)
  File "/usr/lib64/flux/python3.6/flux/util.py", line 203, in set_treedict
    set_treedict(in_dict.setdefault(path[0], {}), path[1], val)
  [Previous line repeated 2 more times]
  File "/usr/lib64/flux/python3.6/flux/util.py", line 205, in set_treedict
    in_dict[key] = val
TypeError: 'int' object does not support item assignment
```

@grondo says: 
> The best solution would be to wrap the calls in the cli commands that end up calling `set_treedict()` in a try then raise a more human readable error like `failed to set shell option {key} to {value}: {exc}`.

wihobbs:
Problem: When submitting a job through the command line, a shell option `-o KEY[=VAL]` is not required to have a value. If one is not provided, then `attributes.system.shell.KEY` is set to 1. If a user later provides a sub-key of this shell option, such as `-o KEY.foo=bar`, the error message returned by Python is not particularly helpful.

Wrap `set_treedict()` in a try/except where user input might provide an invalid key and raise a useful error message.

Fixes #6678

### DIFF
--- a/src/bindings/python/flux/cli/base.py
+++ b/src/bindings/python/flux/cli/base.py
@@ -422,7 +422,10 @@ class BatchConfig:
             value = json.loads(value)
         except json.decoder.JSONDecodeError:
             value = str(value)
-        set_treedict(self.config, key, value)
+        try:
+            set_treedict(self.config, key, value)
+        except TypeError as e:
+            raise TypeError(f"failed to set {key} to {value}: {e}")
         return self
 
     def update_file(self, path, extension=".toml"):
--- a/src/bindings/python/flux/job/Jobspec.py
+++ b/src/bindings/python/flux/job/Jobspec.py
@@ -564,7 +564,10 @@ class Jobspec(object):
             if not key.startswith(("user.", "system.")):
                 key = "system." + key
             key = "attributes." + key
-        set_treedict(self.jobspec, key, val)
+        try:
+            set_treedict(self.jobspec, key, val)
+        except TypeError as e:
+            raise TypeError(f"failed to set {key} to {val}: {e}")
 
     def setattr_shell_option(self, key, val):
         """
--- a/t/t2712-python-cli-alloc.t
+++ b/t/t2712-python-cli-alloc.t
@@ -125,6 +125,12 @@ test_expect_success 'flux alloc: mpi option can be overridden' '
 	flux alloc -o mpi=foo -N1 --dry-run hostname | \
 		jq -e ".attributes.system.shell.options.mpi = \"foo\""
 '
+test_expect_success 'flux alloc: -o foo.bar cannot override -o foo' '
+	test_must_fail flux alloc -o foo -N1 -o foo.bar=hi --dry-run hostname >shellopt.out 2>&1 &&
+	test_debug "cat shellopt.out" &&
+	grep "failed to set attributes.system.shell.options.foo.bar to hi" shellopt.out
+'
+
 test_expect_success 'flux alloc: MPI vars are not set in initial program' '
 	flux queue start &&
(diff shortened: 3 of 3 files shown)


grondo (approved):
LGTM. Thanks!

grondo:
Just a minor point for next time. Not required, but we tend to try to keep the testsuite passing for each commit (sometimes helps with `git bisect`), so when adding tests, add them _after_ fixing the bug or adding the feature on which the tests depend.

grondo:
> Oh, that makes sense. I was trying to show that I did test-driven-development ;)
Yeah, that makes sense too (and I usually write the tests first as well) :-)

PR #6681, opened 2025-03-04
flux-queue: do not default to --all with configured queues

### LINKED ISSUE #6674: Flux queue stop defaulting to --all creates a footgun situation
There's some inconsistent behavior between **flux queue start** and **flux queue stop** as seen below:
```
[root@ahwahnee1:kier1]# flux queue start
warning: --queue/--all not specified, assuming --all
pall: Scheduling is started
[root@ahwahnee1:kier1]# flux queue stop
warning: --queue/--all not specified, assuming --all
pall: Scheduling is stopped
[root@ahwahnee1:kier1]# flux queue start dfasfa
usage: flux-queue [-h] {status,list,enable,disable,start,stop,drain,idle} ...
flux-queue: error: unrecognized arguments: dfasfa
[root@ahwahnee1:kier1]# flux queue stop fsafsa
warning: --queue/--all not specified, assuming --all
pall: Scheduling is stopped: fsafsa
```
In particular, the way **stop** defaults means that it's easy to end up stopping all the queues with a useless drain message, instead of stopping the queue you meant to stop.  Ask me how I know.

I'd argue that defaulting to --all is generally bad here, fwiw.  But at the very least, having **stop** return an error when there's no specified queue but there is another argument would help.

Also, the usage info returned for **start** with an unspecified queue seems to be wrong, possibly because it's getting a positional argument that it can't use.  I'd expect it to return the info from **flux queue start -h**:
```
[root@ahwahnee1:kier1]# flux queue start -h
usage: flux-queue start [-h] [-q NAME] [-a] [--quiet] [-v]

optional arguments:
  -h, --help        show this help message and exit
  -q, --queue=NAME  Specify queue to start
  -a, --all         Force command to apply to all queues if none specified
      --quiet       Display only errors
  -v, --verbose     Display more detail about internal job manager state
[root@ahwahnee1:kier1]# flux queue start fsafas
usage: flux-queue [-h] {status,list,enable,disable,start,stop,drain,idle} ...
flux-queue: error: unrecognized arguments: fsafas
```

grondo:
Problem: The default behavior of `flux queue start|stop` to apply to all queues when named queues are configured is confusing and potentially dangerous.

Issue an error instead of a warning when this is the case.

If there are no configured queues, then `flux queue start` and `flux queue stop` will still work as expected.

Update tests.

Fixes #6674

I'm going to submit a separate PR that will further change `flux queue start|stop` to take one more queues in the free arguments (and move the stop "reason" to a `-m, --message` option). Since this is a breaking change I figured it should be considered separately.

### DIFF
--- a/src/cmd/flux-queue.py
+++ b/src/cmd/flux-queue.py
@@ -400,22 +400,15 @@ def disable(args):
         queue_status(handle, args.queue, args.verbose, print_enable_status)
 
 
-def check_legacy_all(handle, name, all, quiet):
-    if not name and not all and not quiet:
-        future = handle.rpc("job-manager.queue-list")
-        try:
-            qlist = future.get()
-        except Exception:
-            return all
+def check_all(handle, name, all_queues):
+    if not name and not all_queues:
+        qlist = handle.rpc("job-manager.queue-list").get()
         if len(qlist["queues"]) > 0:
-            print(
-                "warning: --queue/--all not specified, assuming --all", file=sys.stderr
-            )
-        return True
-    return all
+            raise ValueError("Named queues are defined. Specify queues or use --all.")
 
 
 def queue_start(handle, start, name, all, nocheckpoint=False, reason=None):
+    check_all(handle, name, all)
     payload = {"start": start, "all": all, "nocheckpoint": nocheckpoint}
     if name:
         payload["name"] = name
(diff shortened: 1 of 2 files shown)


garlick (approved):
LGTM!

PR #6682, opened 2025-03-04
Remove version check from libjob and change to pass with warning in shell

sam-maloney:
Problem: version checks in libjob and shell trigger fatal errors if jobspec `version != 1` but this interferes with implementing further features of canonical jobspec and (IMHO) should be the remit of the validator.

The version checks are changed to issue warnings about unsupported jobspec versions but otherwise allow execution to continue.

This is separated off from #6632 (with a bit of tweaking/polishing) to hopefully create a more digestible, standalone PR.

### DIFF
--- a/src/common/libjob/jj.c
+++ b/src/common/libjob/jj.c
@@ -124,13 +124,8 @@ int jj_get_counts_json (json_t *jobspec, struct jj_counts *jj)
         errno = EINVAL;
         return -1;
     }
-    if (version != 1) {
-        snprintf (jj->error, sizeof (jj->error) - 1,
-                 "Invalid version: expected 1, got %d", version);
-        errno = EINVAL;
-        return -1;
-    }
-    /* N.B. attributes.system is generally optional, but
+    /* jobspec version check omitted as discussed in #6632 and #6682
+     * N.B. attributes.system is generally optional, but
      * attributes.system.duration is required in jobspec version 1 */
     if (json_unpack_ex (jobspec, &error, 0, "{s:{s:{s:F}}}",
                         "attributes",
--- a/src/shell/info.c
+++ b/src/shell/info.c
@@ -175,6 +175,10 @@ static int shell_init_jobinfo (flux_shell_t *shell, struct shell_info *info)
         shell_log_error ("error parsing jobspec: %s", error.text);
         goto out;
     }
+    if (info->jobspec->version != 1) {
+        shell_warn ("Unsupported jobspec version: expected 1 got %d",
+                    info->jobspec->version);
+    }
 
     /*  Synchronously get initial version of R from first job-info
      *  watch response:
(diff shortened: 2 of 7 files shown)


src/common/libjob/jj.c
```diff
@@ -126,9 +127,8 @@ int jj_get_counts_json (json_t *jobspec, struct jj_counts *jj)
     }
     if (version != 1) {
         snprintf (jj->error, sizeof (jj->error) - 1,
-                 "Invalid version: expected 1, got %d", version);
-        errno = EINVAL;
-        return -1;
+                 "Unsupported version: expected 1, got %d", version);
+        flux_log(NULL, LOG_WARNING, "%s: %s", __FUNCTION__, jj->error);
     }
```
garlick, line 132: The warning will be emitted on stderr, which may not be appropriate in all cases, so  I think the entire conditional block should just be removed here.   Maybe a comment referencing the issue would be useful though,
like
```c
// jobspec version check omitted as discussed in #6632 and #6682
```

src/shell/info.c
```diff
@@ -175,6 +175,9 @@ static int shell_init_jobinfo (flux_shell_t *shell, struct shell_info *info)
         shell_log_error ("error parsing jobspec: %s", error.text);
         goto out;
     }
+    if (strstarts (error.text, "Unsupported jobspec version")) {
+        shell_warn ("warning parsing jobspec: %s", error.text);
+    }
```
garlick, line 180: `error.text` could be uninitialized here I think?  It might be better to just eliminate the version check in `jobspec_parse()` and move the test here:
```c
if (jobspec->versions != 1)
    // warn
```

garlick (changes requested):
Thanks for being so patient with us!

I added some minor feedback assuming the general proposal is acceptable, which I'd like to have @grondo consider.  I _think_ it is probably fine but I might be missing something.

grondo:
> assuming the general proposal is acceptable, which I'd like to have @grondo consider. 

Yes, after some consideration I think this approach makes sense, especially at this point in development.

Since the job shell runs as the calling user, there is less urgency to apply extra defenses against bad input (though still the shell should be checking for invalid jobspec or options and issue errors when appropriate. It should not keep going at all costs, IMO)

sam-maloney:
> still the shell should be checking for invalid jobspec or options and issue errors when appropriate. It should not keep going at all costs, IMO

Absolutely agree with the shell (and other components) checking and validating the fields & information they actually use; it was only the errors directly based on the version number itself (as it doesn't actually play a functional role) which seemed out of place. Especially after explicitly setting the validator to pass on other version numbers, it made me a bit sad to have other components continue to fail unnecessarily!

garlick (approved):
LGTM!  Thanks!

sam-maloney:
Mergify failed to update the branch for some reason, so I rebased manually as per its suggestion, but it dismissed the review, sorry about that!

wihobbs:
@sam-maloney If you're comfortable with it, you can link your GitHub account to mergify as an "authorized application," and mergify will rebase your PRs for you when they're added to the merge queue. I did this so long ago that I had forgotten I did! But, I imagine most of us have this set up. Maybe you have too after seeing its comment earlier :) 

I can see the [flux-framework org on mergify automatically](https://github.com/flux-framework/flux-core/pull/6682/checks?check_run_id=38309798925), maybe looking there is a place to start.

PR #6683, opened 2025-03-04
github: update typos check ci action

grondo:
This PR updates the `spelling` github action to use ubuntu-latest instead of ubuntu-20.04 along with the latest version of the typos checker.

The new typos version found a few more typos (including one in code!), so those are fixed as well.


garlick (approved):
LGTM!

That misspelled struct really threw me for a loop until I realized we were just declaring a pointer that was passed to `free (void *)`!  Wow.

grondo:
> That misspelled struct really threw me for a loop
Same!

src/broker/overlay.c
```diff
@@ -2404,7 +2404,7 @@ void overlay_destroy (struct overlay *ov)
         }
         rpc_track_destroy (ov->parent.tracker);
         if (ov->monitor_callbacks) {
-            struct montior *mon;
+            struct monitor *mon;
```
chu11, line 2407: how in the world did this compile before?

chu11: oh i see @garlick comment now ... i did not think that was possible.

Issue #4722, opened 2022-10-26
help output for job-shell options


trws:
As of now, there doesn't seem to be a way to discover job-shell options to pass to `-o` on run or similar.  That's a bit unfortunate because they're the only way to turn on things like affinity support.  We might want to change that default as well, but if we can reasonably read out, or even just list some common ones, it would help with discoverability.

grondo:
The most common shell options are documented in the [JOBSPEC OPTIONS](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man1/flux-shell.html#jobspec-options) section of [flux-shell(1)](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man1/flux-shell.html#flux-shell-1).

Unfortunately it will be non-trivial to support help output at this point. Options are processed arbitrarily by the shell and its plugins, indeed an unfortunate choice but necessitated by rapid pace of development. In addition to the planned `--validate` option for the job shell, we should do something like a `--help` option which calls into all shell plugins with a `shell.help` callback or similar to allow plugin do individually dump this information.

PR #6691, opened 2025-03-07
[WIP]: Support for command-line plugins

wihobbs:
This is being put up as WIP so that people can start to comment on the proposed design and point out testing gaps. 

This is built off of @grondo's cli-plugins branch, with the modification that plugins should be namespaced under a `-P` flag, similar to how clang's `-Wl` flag works but more along the lines of `-o/-S` "plus" in flux. Here you get `-P KEY[=VAL]` but with plugin validation and ability for args to modify jobspecs and config files. It also incorporates the suggestion for a `FLUX_CLI_PLUGINPATH` environment variable that overrides the standard path in confdir for testing.

I'll fold up the outstanding TODOs here, most are documented in the comments:
- Utilize the `argparse.HelpFormatter` class to create a better-formatted argument group for the plugins
- Removing the superfluous base classes from `self.plugins` in the `CLIPluginRegistry`
- The `validate` callback for bar.py seems to be doing nothing; probably misunderstanding how the interface for the validator is supposed to work

I have been staring at this all day and my brain is toast, so apologies if some very basic things are missing or out of place.

### DIFF
--- a/src/bindings/python/flux/cli/fortune.py
+++ b/src/bindings/python/flux/cli/fortune.py
@@ -46,7 +46,7 @@ class FortuneCmd(base.MiniCmd):
 
     @staticmethod
     def create_parser(
-        prog, usage=None, description=None, exclude_io=False, add_help=True
+        prog, usage=None, description=None, exclude_io=False, add_help=True, **kwargs
     ):
         """
         Create a largely empty parser for flux fortune (no arguments or exposed)


src/bindings/python/flux/cli/plugin.py
```diff
@@ -0,0 +1,187 @@
...
+        return self
+
+    def add_plugin_option(self, parser):
+        """Add the -P option and list of available keys"""
+        parser.add_argument(
+            "-P",
+            "--plugin",
+            type=CLIPluginValue,
+            action="append",
+            help=f"{self.get_plugin_help_messages()}",
```
grondo, line 153: As you've noticed, this isn't going to display well in `--help` output.

Let's add a function to organize all the current plugin names and short usage messages into a nice table, then emit that in a custom section of the help output for the commands, e.g.:

```
Options provided by plugins:
    foo=VAL      Set option foo to VAL.
    bar[=VAL]    Set a bar with optional VAL.
```

Also, the `add_argument()` should not go in this class if we're treating them this way. Instead it just be with all the other `add_argument()` calls in the cli base.py.

src/bindings/python/flux/cli/plugin.py
```diff
@@ -0,0 +1,187 @@
...
+    def get_plugin_help_messages(self):
+        """Return a string that has all self.plugin help messages"""
+        ## TODO: Utilize the HelpFormatter class in argparse to make
+        ##  these messages better
+        help = ""
+        for plugin in self.plugins[1:]:
+            help += str(plugin.get_plugin_name())
+            help += str(plugin.get_help_message())
+            help += "\n"
+        return help
```
grondo, line 172: This should be replaced with a function to pretty print the option+usage table as noted above.

src/bindings/python/flux/cli/base.py
```diff
@@ -751,7 +752,10 @@ def __init__(self, prog, usage=None, description=None, exclude_io=False):
         self.exitcode = 0
         self.progress = None
         self.watcher = None
+        self.plugins = CLIPluginRegistry().load_plugins(prog)
         self.parser = self.create_parser(prog, usage, description, exclude_io)
+        self.group = self.parser.add_argument_group("Options provided by plugins")
+        self.plugins.add_plugin_option(self.group)
```
grondo, line 758: Adding the `-P, --plugin` option should be pulled out of the function and moved to where all the other options are added.

src/bindings/python/flux/cli/base.py
```diff
@@ -1010,6 +1014,20 @@ def jobspec_create(self, args):
...
+        plugins = {}
+        if args.plugin:
+            for provided_arg in args.plugin:
+                key, val = provided_arg.get_value()
+                if key in self.plugins.get_plugin_options():
+                    plugins[key] = val
+                else:
+                    raise ValueError(
+                        f"Unsupported option provided to -P/--plugin: {key}"
+                    )
```
grondo, line 1026: It might be better if this code was part of the CLIPluginRegistry class, which should take args and extract out the active plugins, raising an exception if there was any invalid input or a plugin raised an error in preinit.

src/bindings/python/flux/cli/plugin.py
```diff
@@ -0,0 +1,187 @@
...
+        Either here, or in validation, plugin extensions should check types and
+        perform their own validation, since argparse will accept any string.
+
+        Args:
+            args (:py:obj:`Namespace`): Namespace result from
+                :py:meth:`Argparse.ArgumentParser.parse_args()`.
+            values (:obj:`Dict`): Dictionary of KEY[=VALUE] pairs returned
+                by parsing the arguments provided to the -P/--plugin
+                option. KEY is None if plugin was not invoked. VALUE is the
+                empty string if the plugin was invoked without VALUE.
```
grondo, line 78: >  KEY is None if plugin was not invoked.

If `values` is a dictionary then by "KEY is None" here do you mean "is not set in the dictionary"?

Is `values` the dictionary of all `NAME=VALUE` pairs provided to `--plugin`? Maybe it would be better to use a `Namespace` here and keep the same semantics as argparse, i.e. `args.name` is `None` if the plugin was not invoked, `True` if it was invoked without an arg, and a string argument that the plugin is responsible for parsing if an argument was given.

grondo (commented):
I took a quick pass this evening. I haven't tested yet nor looked into why the validator isn't working. I'm not sure if I tested that in the previous prototype.

t/t2717-python-cli-plugins.t
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
+
+test_description='Test command line plugin interface'
+
+. $(dirname $0)/sharness.sh
+
+test_under_flux 4 job
+
+export FLUX_CLI_PLUGINPATH=${FLUX_BUILD_DIR}/t/cli-plugins/cli/plugins
```
grondo, line 9: This line needs to be moved above `test_under_flux` or else the job validator will not have `FLUX_CLI_PLUGINPATH` in its environment.

However, after fixing that every test is failing because the `CustomFluxionPolicyPlugin` always raises the error `hi hobbs`

wihobbs:
Thanks for taking the time to dissect all of this @grondo! It was getting difficult in the thick of this project to figure out what exactly to do _next_, so thanks for getting me going. I'll note a few things:

1. I don't think having a lot of files is necessarily bad, from a center perspective this is basically how we do prolog and epilog scripts anyway, even though we could in theory have one giant one or consolidate some together
2. Allowing validate-only plugins is certainly a value-add, but I wonder if we need another class for this, or if we can just make all of the plugin options except `prog` named (optional) class attributes. I might try this today and see how it works out. This, however, wouldn't allow one plugin to supply multiple options, which might be useful. 
3. Adding `-P`/`--plugin` to the parser where all the other options are is what I initially tried, but the method where this happens is an abstract class, so it doesn't have access to `self`, so `self.get_help_messages` doesn't work. This probably isn't a concern anymore given the proposed change to the help message.
4. Thanks for fixing the sharness test :)

src/bindings/python/flux/cli/plugin.py
wihobbs: Oops, missed this on the last pass.

src/bindings/python/flux/cli/plugin.py
```diff
@@ -0,0 +1,249 @@
...
+    def __init__(self, prog, version=None):
+        self.prog = prog
+        if prog.startswith("flux "):
+            self.prog = prog[5:]
+        self.options = {}
+        self.version = version
+    
+    def add_option(self, opt, usage=None):
+        try: 
+            val = self.options[opt]
```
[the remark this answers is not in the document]
wihobbs: flake8 caught this _much_ faster, mr. AI bot :)

wihobbs:
I tried several things but never figured out the distcheck issue. Will look into it again today.

src/bindings/python/flux/cli/base.py
```diff
@@ -772,7 +780,8 @@ def create_parser(
             prog=prog,
             usage=usage,
             description=description,
-            formatter_class=flux.util.help_formatter(),
+            formatter_class=flux.util.help_formatter(raw_description=True),
```
grondo, line 783: Why did this change to raw_description=True? I think this is messing up the notes under `Per task options: ` and `Per resource options: ` which are no longer being wrapped at line length on this branch.

src/bindings/python/flux/cli/base.py
```diff
@@ -969,6 +978,14 @@ def create_parser(
         parser.add_argument(
             "--debug-emulate", action="store_true", help=argparse.SUPPRESS
         )
+        parser.add_argument(
+            "-P",
+            "--plugin",
+            type=CLIPluginValue,
+            action="append",
+            help="Activate functionality of external plugins with KEY[=VAL]. See "
+            + '"Options provided by plugins" for available plugin options.',
```
grondo, line 987: Suggestion: `KEY` -> `PLUGIN`?

src/bindings/python/flux/cli/plugin.py
```diff
@@ -0,0 +1,249 @@
...
+            setattr(values, str(plugin), None)
+        if args.plugin:
+            for provided_arg in args.plugin:
+                key, val = provided_arg.get_value()
+                if key in self.get_plugin_options():
+                    setattr(values, key, val)
+                else:
+                    raise ValueError(
+                        f"Unsupported option provided to -P/--plugin: {key}"
+                    )
```
grondo, line 235: I think it might be more clear to split out the handling of `--plugin` arguments into a separate method, then store the `CLIPluginNamespace` as an attribute in the registry, instead of constructing it on demand here in preinit().

Then the plugin args namespace doesn't have to be exposed to the registry users, and e.g. when `modify_jobspec()` is called, instead of passing "`values`" (which is really the CLIPluginNamespace), that argument can be dropped and `self.args` or `self.plugin_args` can be passed to each `plugin.modify_jobspec()`.

grondo (commented):
Just leaving some comments after a quick first pass.

src/bindings/python/flux/cli/base.py
wihobbs: Without `raw_description=True`, argparse throws away the formatting for the "Options provided by plugins" section, like so: 

```
Per resource options:
  The following options allow per-resource specification of tasks, and
  should not be used with per-task options above

      --cores=N               Request a total number of cores
      --tasks-per-node=N      Force number of tasks per node
      --tasks-per-core=N      Force number of tasks per core
      --gpus-per-node=N       Request a number of GPUs per node with --nodes

Options provided by plugins: gpumode Option for setting AMD SMI compute
partitioning. Choices are CPX, TPX, or SPX. foo Enable bar matchpolicy Set the
sched-fluxion-resource.match-policy for a subinstance. See sched-fluxion-
resource(5) for available policies.
```

The flux-uri(1) commands use the [`argparse.RawDescriptionHelpFormatter` class](https://docs.python.org/3/library/argparse.html#argparse.RawDescriptionHelpFormatter), and that command (which has a "Supported resolver schemes" section) + the argparse documentation is where I got the idea to use an epilog for this.

I don't see any diff on the formatting the notes under `Per task options:` and `Per resource options:` when I test it on corona, can you share an example of that?

wihobbs: Oh, never mind. I see what you mean. Yeah, I'll look at that.

t/t2717-python-cli-plugins.t
```diff
@@ -0,0 +1,47 @@
+#!/bin/sh
+
+test_description='Test command line plugin interface'
+
+. $(dirname $0)/sharness.sh
+
+export FLUX_CLI_PLUGINPATH=${FLUX_BUILD_DIR}/t/cli-plugins/
```
grondo, line 7: This line is why distcheck is failing. The test plugins are not in the build directory, they are in the source directory.
```suggestion
export FLUX_CLI_PLUGINPATH=${SHARNESS_TEST_SRCDIR}/cli-plugins/
```

src/bindings/python/flux/cli/plugin.py
```diff
@@ -0,0 +1,249 @@
...
+            self.prog = prog[5:]
+        self.options = {}
+        self.version = version
+
+    def add_option(self, opt, usage=None):
+        try:
+            self.options[opt]
+            raise ValueError(f"{opt} has already been registered")
+        except KeyError:  # default behavior
+            self.options[opt] = usage
```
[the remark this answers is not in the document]
wihobbs: > pass the registry into the CLIPlugin __init__() initializer.

I think the registry only needs to be passed to `CLIPlugin.add_options`, no? This would be similar to how the argparse parser was passed to `CLIPlugin.add_options` in your first prototype. (This is how I did it in the most recent push but if there's a reason you wanted the registry in `__init__` that should be possible too.)

src/bindings/python/flux/cli/base.py
wihobbs: image

I added a newline to each argument group's description at where the line currently breaks. I checked and arguments themselves are still getting wrapped at line length automatically.

src/bindings/python/flux/cli/plugin.py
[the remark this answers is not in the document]
grondo: Yes, it could be done that way as well. I was thinking if we pass it to the initializer, there's no need for the `add_options` plugin callback, plugins can just add options using the registry handle if they desire.

Other registry methods could also be added in the future to allow plugins to do "other things", which would not necessary make sense in an `add_options` callback.

I may have forgotten, though, why a separate callback was necessary in the prototype.. perhaps it is too early to add options when the plugin initializers are called? In that case, your point is taken and I wonder if we should rename the `add_option` callback to something else.

wihobbs:
> Allowing validate-only plugins is certainly a value-add, but I wonder if we need another class for this, or if we can just make all of the plugin options except prog named (optional) class attributes. I might try this today and see how it works out. This, however, wouldn't allow one plugin to supply multiple options, which might be useful.

I've been thinking about this, and I feel like we should be guiding developers toward the validator's plugin capabilities if validation is all that is needed.

Perhaps it's that we were thinking of shell options as the use case for validate-only plugins, and the nice-to-have feature of allowing them to extend the `--help` message. That's a good goal, but I wonder if it should be addressed separately, since we've learned from this that the `-o OPT=VAL` syntax is not easily supported using normal Argparse's builtin help groups/formatter. (I also realized as I started on this "validate only plugin suite for shell options" that several shell options are already aliased to long options, such as `--signal`.)

grondo:
The validator is optional, so I still think allowing validate-only plugins (or other plugins that don't necessarily export an option to the user) is useful and should be included.

src/bindings/python/flux/cli/plugin.py
```diff
@@ -0,0 +1,260 @@
...
+        self.options = {}
+        self.version = version
+
+    def add_options(self, registry):
+        """Plugins implement this method to register options in the registry.
+
+        Plugins can call:
+        >>> registry.add_options("foo", usage="Enable foo.")
+
+        to add new options to the registry.
```
jameshcorbett, line 103: What about instead of having a method that takes a `CLIPluginRegistry` and calls `registry.add_options`, there were a method that returned some objects with the information needed to add options? Like for this example above, return `{"name": "foo", "usage"="Enable foo."}` or something. I think passing the `registry` in to this method seems like a violation of the abstraction to me... I see there's a discussion going on a related topic that I haven't fully grok'd yet.

wihobbs: It probably is a violation of abstraction. Our [`create_parser` method in `MiniCmd` is an abstract class](https://github.com/flux-framework/flux-core/blob/5f55d3c57aeed2e864e69105338668bee5461a2b/src/bindings/python/flux/cli/base.py#L757), so it has no access to `self`, which is why the parser is constructed on the fly and passed around like this. (The init method of MiniCmd sheds some light on this.)

If we we go with the alternative PR, which uses argparse's `add_option` method, plugins will have access to more than just a name and usage message, they'll have the ability to specify a ton of things, and that could get ugly. Maybe argparse provides an option class we could use for your suggestion...

grondo: I like @jameshcorbett's suggestion. Perhaps plugins optionally implement a `get_options()` method, and the registry code queries the plugin to see what options it wants to implement. There could be a `CLIPluginOption` convenience class to allow a restricted set of argparse `add_option_argument` args and kwargs, or a well defined dict could be supported as suggested by @jameshcorbett.

wihobbs:
Well, admittedly, even though I agree with all of the arguments in #2875 not to pursue this approach, I'm still a little sad most of my work here has to go in the garbage can. I'll salvage the testing and example plugins and open a new PR off of grondo/cli-plugins. 

To remind myself this wasn't a waste of time, here's a summary of what I learned and what we fixed along the road:

1. Flux's submission tools being Python code is a _huge_ win for extensions and other framework projects. Feeling comfortable in the front end of flux-core has reaped other benefits, such as being able to fix #6678, improve QOL in `pdb` through #6778, and help @cmoussa1 with [flux-accounting#621](https://github.com/flux-framework/flux-accounting/pull/621).
2. It took me a while to understand what `MiniCmd` is doing. It initilizes itself as a class, then calls its own static method, `create_parser`, that returns an `argparse.ArgumentParser` with the correct construction. As @jameshcorbett noted above the early prototypes that tried to integrate into this model violated some abstraction principles. FWIW, I follow code like this by trying to find the entrypoint and trace forward from that -- so for future CLI developers a good place to start is `MiniCmd.__init__` and `MiniCmd.create_parser()`. Another good entrypoint is our `util.py`, and I'm still hopeful we can solve #4722 with a [custom `HelpFormatter` epilog](https://stackoverflow.com/questions/4375327/python-argparse-preformatted-help-text?rq=3); a WIP of this is on [my shell-help branch](https://github.com/flux-framework/flux-core/compare/master...wihobbs:flux-core:shell-help).
3. *All* of argparse's functionality is basically driven by having an [`ArgumentParser`](https://docs.python.org/3/library/argparse.html) and modifying it with that class's methods. Early on in my brain we needed a custom _action_ to validate plugin options; @trws had the excellent suggestion to use a custom _type_ instead that provided validation in its construction, then just rely on the standard `append` action. If the type fails to construct, argparse will reject the option argument as invalid. Beautiful.
4. The validator's protocol being `stdout` means that you can't easily attach a debugger to it if you need to debug validation from the CLI. Also, the validator is loaded _at instance start_ so later environment mangling to try and modify what it does don't work ([see Mark's comment above](https://github.com/flux-framework/flux-core/pull/6691#discussion_r1985218924)). The validator can be run on the command line with `flux job-validator` though, which is a win for debugging.
5. Early models of the `CLIPlugin` class set `opt` and `help` [directly](https://github.com/flux-framework/flux-core/commit/e06623a43379433fdf19e8138f3786326761eb9c#diff-1d15d794b28bd60a334ac2d983e044d8d0e74389096e2380986c93192ebcb102R9). @grondo suggested instead that [a call to `super.__init__()`](https://github.com/flux-framework/flux-core/commit/4eee9f7b3c2b3cff6244014a67d438e10e61ca09#diff-d2d42c7a8892396e323df53888859baa51e58703e2f2b791b9f50734e95336d8R7) could include `opt`, `help`, and `usage`. It wouldn't have occurred to me to use a call to `super` to construct things in subclasses that aren't present in the base class itself, but I thought it was a good idea. "Always look for ways to condense and clean" or something like that is what Mark said when he pointed this out.

There's more, certainly this isn't an exhaustive list of what it took to get here. But hopefully someone sees this and it helps them run faster with contributing to the core CLI. :) And thanks to the people tagged for the tips and many reviews!

PR #6697, opened 2025-03-10
sdmon: reconnect when bus is lost

garlick:
Problem: the sdmon module exits if sdbus ever loses its d-bus connection

`sdbus` was designed to support this case and returns EAGAIN when a user should not treat the RPC failure as fatal.  Interpret this error in `sdmon` and reset/re-establish the monitor connection when this error is received.

You might wonder why sdmon doesn't just hide this from users.  It's because events may be lost, so  for correctness, users like sdmon must start over (in this case, establish signal subscription, fetch current unit status, then monitor the subscription).


grondo (approved):
LGTM!

garlick:
I'll set MWP then - thanks!

Issue #6698, opened 2025-03-11
Flux died after `TimeoutError`


andre-merzky:
I stumbled over a problem running a flux instance across 8 nodes on Frontier.  For context: the job used 128 nodes, I was running 8 flux instances, each managing 8 nodes.  Only one instance showed the error, and I have only seen the problem once in a fair number of runs and different configurations.

The stacktrace I found is:

```
exception calling callback for <FluxExecutorFuture at 0x7f69281ea990 state=running>
Traceback (most recent call last):
  File "/autofs/nccs-svm1_home1/matitov/am/spack/opt/spack/linux-sles15-zen3/gcc-13.3.0/flux-core-0.67.0-x2tmfynht4tj2coroi5kyebvn6tfxfp7/lib/python3.11/site-packages/flux/job/executor.py", line 76, in run
    self.__run()
  File "/autofs/nccs-svm1_home1/matitov/am/spack/opt/spack/linux-sles15-zen3/gcc-13.3.0/flux-core-0.67.0-x2tmfynht4tj2coroi5kyebvn6tfxfp7/lib/python3.11/site-packages/flux/job/executor.py", line 98, in __run
    if self.__flux_handle.reactor_run() < 0:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/autofs/nccs-svm1_home1/matitov/am/spack/opt/spack/linux-sles15-zen3/gcc-13.3.0/flux-core-0.67.0-x2tmfynht4tj2coroi5kyebvn6tfxfp7/lib/python3.11/site-packages/flux/core/handle.py", line 323, in reactor_run
    Flux.raise_if_exception()
  File "/autofs/nccs-svm1_home1/matitov/am/spack/opt/spack/linux-sles15-zen3/gcc-13.3.0/flux-core-0.67.0-x2tmfynht4tj2coroi5kyebvn6tfxfp7/lib/python3.11/site-packages/flux/core/handle.py", line 134, in raise_if_exception
    raise cls.set_exception(None) from None
  File "/autofs/nccs-svm1_home1/matitov/am/spack/opt/spack/linux-sles15-zen3/gcc-13.3.0/flux-core-0.67.0-x2tmfynht4tj2coroi5kyebvn6tfxfp7/lib/python3.11/site-packages/flux/future.py", line 38, in continuation_callback
    py_future.then_cb(py_future, *py_future.then_args, **py_future.then_kwargs)
[...]

PR #6700, opened 2025-03-11
python: make FluxExecutor catch submission errors

jameshcorbett:
Problem: as reported in #6698, sometimes the `submit_get_id` function can return errors, but the FluxExecutor has no handling for this.

Add exception handling, and for `get_event` as well for safety.

### DIFF
--- a/src/bindings/python/flux/job/event.py
+++ b/src/bindings/python/flux/job/event.py
@@ -142,7 +142,7 @@ class JobException(Exception):
         self.type = event.context["type"]
         self.note = event.context["note"]
         self.severity = event.context["severity"]
-        super().__init__(self)
+        super().__init__()
 
     def __str__(self):
         return f"job.exception: type={self.type}: {self.note}"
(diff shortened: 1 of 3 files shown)


grondo (approved):
This LGTM (as far as I understand the FluxExecutor).

Would it be possible to add a test, or is this already covered in existing tests?

jameshcorbett:
Good call, I'll add an extra test tomorrow.

jameshcorbett:
For https://github.com/flux-framework/flux-core/pull/6700/commits/bdd33697f25d4c7cb3e822d8da23c784615b9e2b, see:

```
>>> class JobException(Exception):
...     def __init__(self, arg):
...         self.x = arg
...         super().__init__(self)
... 
>>> x = JobException("hello")
>>> x
  File "<stdin>", line 1, in <module>
RecursionError: maximum recursion depth exceeded while getting the repr of an object
```

jameshcorbett:
Ok sorry I force-pushed the whole thing just to add pylint comments (`# pylint: disable=broad-except`) to the first commit because I noticed the rest of the executor code has pylint tags for some reason and I wanted to be consistent.

grondo (approved):
LGTM!

jameshcorbett:
Thanks!

PR #6705, opened 2025-03-13
shell: determine node & core resource counts from R during init

sam-maloney:
Problem: counts for resources are currently being parsed only from jobspec, but if/once ranges are used then values cannot be determined.

Modify shell init data flow to use the actual allocation from the scheduler in R, if present, to determine resource counts for nodes & cores in `jobspec_parse`.

As a first step, this checks whether a resource set struct was passed to `jobspec_parse`, and otherwise leaves the original codepath unchanged. The slot count must still be parsed from the jobspec regardless, because it is missing from R.

The concept for this is also essentially separated off from #6632, albeit with a fair bit of re-working to keep the changes as minimal as possible. Because `flux-sched` already has basic support for ranges, this PR already allows jobspec with ranges for nodes and/or cores to run successfully.

### DIFF
--- a/src/shell/info.c
+++ b/src/shell/info.c
@@ -171,7 +171,16 @@ static int shell_init_jobinfo (flux_shell_t *shell, struct shell_info *info)
         shell_log_error ("error fetching jobspec");
         goto out;
     }
-    if (!(info->jobspec = jobspec_parse (jobspec, &error))) {
+
+    /*  Synchronously get initial version of R from first job-info
+     *  watch response:
+     */
+    if (resource_watch_update (info) < 0)
+        goto out;
+
+    /*  Parse jobspec after R so resource information can be read from R:
+     */
+    if (!(info->jobspec = jobspec_parse (jobspec, info->rcalc, &error))) {
         shell_log_error ("error parsing jobspec: %s", error.text);
         goto out;
     }
@@ -179,12 +188,6 @@ static int shell_init_jobinfo (flux_shell_t *shell, struct shell_info *info)
         shell_warn ("Unsupported jobspec version: expected 1 got %d",
                     info->jobspec->version);
     }
-
-    /*  Synchronously get initial version of R from first job-info
-     *  watch response:
-     */
-    if (resource_watch_update (info) < 0)
-        goto out;
 
     /*  Register callback for future R updates:
      */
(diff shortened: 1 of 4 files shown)


grondo:
This is great! The changes look minimal and understandable so I'll try to get a review done soon. Thanks!

sam-maloney:
@grondo perhaps a gentle nudge for a review now that the May release is complete :)

grondo:
Thanks for the nudge @sam-maloney!

This looks good to me. However, one observation is that it appears the "old" code path (`recursive_parse_jobspec_resources ()` which gathers the counts via jobspec only) is now unused except for the unit tests in `src/shell/test/jobspec.c`. I wonder if we should just drop that code and update the unit tests to provide an R?

What do you think? It is also possible I misread the code and there is some use case still present outside of the unit test. Even in that case though, it might be nice to have a structure to test with provided simple R/`rcalc` values.

sam-maloney:
@grondo alright, I think I have a working update ready; tests are passing on my machine so hopefully the CI agrees!

I've removed the unused code, and added a simple resource set to each test which matches with the resources requested by the jobspec being tested (or at least something reasonable for tests that are supposed to fail).

I've completely removed a number of the _bad_ tests that were checking for certain disallowed resource configurations, since part of the argument driving this work is that the shell shouldn't really be parsing the jobspec resources but rather leave it to the validator and/or scheduler.

The only previously _good_ test which now fails (and I've also removed) was the "node[2]->socket[3]->slot[5]->core[3]". Again, since the impetus for this is allowing ranges to be specified, I'm intentionally avoiding the assumption that any of the counts are integers (other than for the slots), so having a resource with a non-unity count between the node and slot (if node is given) means there's no easy way to know the multiplier. This (hopefully non-deal-breaking) restriction is documented in the comment above the recursive parsing function.

And fingers crossed that down the road, if/once slot information can be included in R, then the remaining resource parsing can also be dropped entirely!

grondo:
Apologies @sam-maloney, I dropped the ball here :disappointed: 

Once #6930 is merged we can rebase and merge this one too.

sam-maloney:
> Apologies @sam-maloney, I dropped the ball here 😞

No worries @grondo; with our project officially starting it's had the paradoxical effect of diverting me from working on flux with lots of meetings and admin work as everything gets going, haha!

grondo (dismissed):
This LGTM, thanks @sam-maloney! I'm going to click the rebase button and set MWP.

grondo (approved):

PR #6706, opened 2025-03-13
accept target queue as non-option parameter in most `flux queue` subcommands

grondo:
This PR reworks the argument handling in many `flux queue` subcommands so that the target queue name is given as the first non-option parameter (aka free arguments), since this is probably what most users expect. The `-q, --queue=NAME` options are still supported, but are now hidden and undocumented and could be deprecated in a future release.

The "reason" in `flux queue disable` is now optional, and a new `-m, --message=REASON` option is provided for both `flux queue disable` and `flux queue stop` for specifying the reason now that it is not specified in the free arguments. This should allow old usage to immediately fail with "No such queue" if someone uses the old method `flux queue stop -q batch reason` since queue "reason" won't likely exist.

We should get @kkier and @ryanday36 to sign off on these changes since scripts and admin behavior may need to change.

### DIFF
--- a/doc/man1/flux-queue.rst
+++ b/doc/man1/flux-queue.rst
@@ -9,14 +9,14 @@ flux-queue(1)
 SYNOPSIS
 ========
 
-| **flux** **queue** **list** [-n] [-o FORMAT]
-| **flux** **queue** **status** [*-q* *NAME* | *-a*] [*-v*]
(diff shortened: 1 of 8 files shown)


garlick (approved):
FWIW I'm good with this change!

kkier:
Seems good to me, I can't find anywhere this would break something.

grondo:
Thanks @kkier! I'll set MWP.

Issue #6712, opened 2025-03-15
SIGKILL isn't sent to stuck job shells on elcap


grondo:
During a full system DAT on elcap we noticed that jobs within a batch job were stuck in CLEANUP for 15m.

Examining `flux module stats job-exec` showed that ~3000 or so job shells were still "active". The job shells in question were all in spindle code, ignoring SIGTERM. However, they did exit on receipt of SIGKILL.

The job exec module was reportedly sending SIGKILL, but this was apparently not working for some unknown reason.

One interesting thing is that during a run I notice two flux-shell processes instead just one that I expect, though I could be forgetting something obvious. One guess though is that the parent shell exits on SIGKILL, but the child remains, and Flux is waiting for stdio:

```
flux-shell(793633)───flux-broker-6(793634)───flux-shell(794292)─┬─flux-shell(794297)
                                                                app
```

watson6282:
In the description, I think you meant SIGKILL instead of SIGTERM in the `However, they did exit on receipt of SIGTERM.` sentence.

watson6282:
From experimentation, something about the spindle setup leads to that second `flux-shell` (pid 794297 in the example). Turning off spindle, leads to this:
```
$ flux batch -N 1 -q pdebug --wrap flux run -o spindle.level=off sleep 60

flux-shell(203585)───flux-broker-0(203586)───flux-shell(203664)───sleep(203666)
```

grondo:
> In the description, I think you meant SIGKILL instead of SIGTERM in the However, they did exit on receipt of SIGTERM. sentence.

Thanks! I edited the description.

> Turning off spindle, leads to this:

Thank you @watson6282. It appears the spindle backend calls `fork()` in the job shell plugin to avoid blocking the shell itself. I had forgotten this. One note is that this copies the disposition of the SIGTERM handler in the job shell (which captures many signals in order to forward them to the job tasks). One potential improvement would be to reset signal handlers in the backend after fork to either allow SIGTERM to terminate the process, or to shutdown gracefully.

I also verified that the flux-shell child process created by spindle is in the same process group as the parent shell. Flux is using `killpg(2)` to signal the job shell, so I still don't understand why SIGKILL is apparently not delivered. This could perhaps be an issue in Flux subprocess signaling when the actual process invoked has exited but some processes still remain in the process group.

So, I think we'll need a couple fixes here:
[...]

PR #6714, opened 2025-03-17
fix failure to signal a process group after the process group leader exits

grondo:
This PR fixes the main issue described in #6712. In this case, a child forked from the job shell kept the subprocess "active" (waiting for stdio to close) after the shell was sent SIGTERM (the process was inadvertently ignoring SIGTERM). When the job exec module attempted to SIGKILL, this was a noop since bulk-exec and libsubprocess do not allow sending a signal to a subprocess that has exited.

Fix the issue by allowing signals to be sent to any active subprocess if that subprocess was spawned in a process group.

### DIFF
--- a/src/common/libsubprocess/test/subprocess.c
+++ b/src/common/libsubprocess/test/subprocess.c
@@ -501,7 +501,8 @@ void output_processes_cb (flux_subprocess_t *p, const char *stream)
                 child_pid = atoi (buf);
         }
 
-        if (output_processes_cb_count == 1) {
+        if (output_processes_cb_count == 1
+            && !flux_subprocess_aux_get (p, "nokill")) {
             flux_future_t *f = NULL;
             f = flux_subprocess_kill (p, SIGTERM);
             ok (f != NULL, "flux_subprocess_kill returns future_t");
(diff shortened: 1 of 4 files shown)


src/common/libsubprocess/test/subprocess.c
```diff
@@ -576,6 +577,88 @@ void test_kill_setpgrp (flux_reactor_t *r)
...
+    int ret;
+
+    ok ((cmd = flux_cmd_create (2, av, environ)) != NULL,
+        "flux_cmd_create");
+    ok (flux_cmd_setenvf (cmd, 1, "TEST_FORK_SLEEP_NOWAIT", "t") == 0,
+        "setenv TEST_FORK_SLEEP_NOWAIT=t");
+
+    flux_subprocess_ops_t ops = {
+        .on_completion = completion_parent_normal_exit,
+        .on_stdout = output_processes_cb,
```
chu11, line 625: if I'm looking at code correctly, `output_processes_cb()` will send a sigterm as well, which I don't think is what you want here.  You probably just want the parent to exit within `test_fork_sleep` and then have the state handler sigterm when that parent exits.

So perhaps the run of the mill output callback should be used instead of `output_processes_cb()`

grondo: Good point. I did try that and was having trouble, but let me do that again.

Note this branch is causing the job exec "dummy" tests to fail and I haven't been able to determine why yet :shrug:

grondo: Oh, I forgot. This test sets a `nokill` aux item on the subprocess which prevents `output_processes_cb()` from sending SIGTERM (see above), thereby keeping the test mostly the same as the previous test.

grondo: You can actually tell it isn't sending SIGTERM because this line under `# kill_setpgrp`
```
ok 130 - flux_subprocess_kill returns future_t
```
Is not printed under `# kill_setpgrp_parent_exited`

chu11: ahhh sorry I missed that.   Would it be easier to just use one of the other output callbacks in the test file, since you don't actually need `output_processes_cb`?

On the other hand, it is mildly beneficial to have the test look simlar to the other `setpgrp` one, so alternatively maybe just a comment is needed.

grondo: Yeah, I wanted to make it the same as the existing test.

There's already a comment:
```c
    /* Don't kill subprocess in output handler:
     */
    ok (flux_subprocess_aux_set (p, "nokill", (void *) 0x1, NULL) == 0,
        "flux_subprocess_aux_set (\"nokill\")");
```

chu11: oh jeez, brain fart :-)   i guess i just got confused, it all makes sense now.  I guess all the bases are covered.

chu11 (approved):
LGTM

PR #6716, opened 2025-03-18
shell: clear signal mask and handlers in shell for plugins that call `fork(2)`

grondo:
This PR adds an atfork handler to the job shell that clears the signal mask and resets signal handlers in the case a shell plugin calls `fork()`. This allows SIGTERM sent by the job execution system to terminate these extra child processes of the shell (unless the plugin installs its own handler to deal with SIGTERM more gracefully)

### DIFF
--- a/t/t2621-job-shell-plugin-fork.t
+++ b/t/t2621-job-shell-plugin-fork.t
@@ -1,0 +1,36 @@
+#!/bin/sh
+#
+test_description='Test flux-shell behavior when plugin calls fork(2)'
+
+. `dirname $0`/sharness.sh
+
+
+test_under_flux 2 job
+
+FORK_PLUGIN="${SHARNESS_TEST_DIRECTORY}/shell/plugins/.libs/fork.so"
+test_expect_success 'create shell initrc to load fork.so plugin' '
+	cat <<-EOF >fork.lua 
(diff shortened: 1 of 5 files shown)


grondo:
Actually I think this has to go on top of #6714 or the test may fail (unable to signal the shell child after the main process exits)

t/t2621-job-shell-plugin-fork.t
```diff
@@ -0,0 +1,36 @@
...
+test_under_flux 2 job
+
+FORK_PLUGIN="${SHARNESS_TEST_DIRECTORY}/shell/plugins/.libs/fork.so"
+test_expect_success 'create shell initrc to load fork.so plugin' '
+	cat <<-EOF >fork.lua 
+	plugin.load { file = "$FORK_PLUGIN" }
+	EOF
+'
+test_expect_success 'submit sleep job with fork plugin' '
+	id=$(flux submit -o userrc=fork.lua -N2 sleep 30)
```
chu11, line 17: i guess it doesn't matter for this particular test, since we're really testing the plugin fork, but since timeouts below are 60s, make the sleep like 100s?

chu11 (approved):
LGTM, just one stupid comment that you can take or leave :-)

grondo:
Thanks! I'll set MWP here.

Issue #6695, opened 2025-03-08
kvs-watch: loop in watch_cancel_all() does not scale well


chu11:
Per some profiling done following up #6690 and #6694 work, if kvs-watch is being used by a lot of jobs (i.e. lots of namespaces), the loop in `watch_cancel_all()` does not scale well after every call to cancel a watcher.  i.e. the loop is regularly iterating on every namespace that it is watching and every request within each of those namespaces.  This leads to a slowdown in processing and eventually the ENODATA that is sent to the original watcher.

unclear of current obvious solution.  Perhaps a hash that stores message match tags?  (Edit: a HOLA w/ some special matchtag hash comparator?)

Update:

```
   28.03%     0.00%  kvs-watch        kvs-watch.so           [.] mod_main                                                                 ▒
     mod_main                                                                                                                              ▒
     flux_reactor_run                                                                                                                      ▒
     ev_run                                                                                                                                ▒
   - ev_run                                                                                                                                ▒
      - 27.99% ev_invoke_pending                                                                                                           ▒
         - 27.02% handle_cb                                                                                                                ▒
            - 26.93% dispatch_message (inlined)                                                                                            ▒
               - 26.90% call_handler                                                                                                       ▒
                  - 23.07% watcher_cancel_all                                                                                              ▒
                     - 22.99% watcher_cancel_ns (inlined)                                                                                  ▒
                        - 21.75% watcher_cancel (inlined)                                                                                  ▒
                           - 21.53% flux_disconnect_match                                                                                  ▒
                              - 20.99% flux_msg_route_match_first                                                                          ▒
                                 - 12.54% flux_msg_route_first                                                                             ▒
                                      11.50% msg_validate (inlined)                                                                        ▒
                                   8.37% __strcmp_avx2                                                                                     ▒
                          1.25% fzlistx_next                                                                                               ▒
                  + 3.75% setroot_cb                                                                                                       ▒
         + 0.57% lookup_continuation     
```

chu11:
it's worth noting that there are two loops in `kvs-watch` that are a concern

1 - loop over all namespaces
2 - loop over all watchers within a namespace

number 2 is the bigger issue with our throughput tests.  the first is theoretically an issue, but with the typical testing we do with our throughput tests, it's not a big deal (i.e. we need a lot of simultaneously running jobs that have created a lot of namespaces).

Issue #6690, opened 2025-03-06
slow throughput at 16K jobs


grondo:
I happen to do some throughput tests today and noticed that above about 8K jobs there's a huge slowdown (8K exhibited an expected ~150 job/s).

I ran `perf` while an instance seemed to be "stuck" trying to run 16K jobs and there may be some low hanging fruit here. Namely, it appears 75% of the broker's time is spent in the `kvs-watch` module in a blocking `flux_event_subscribe(3)`  call (if the `perf` output is to be believed):
```
-   94.24%     0.00%  kvs-watch        libc-2.28.so           [.] __GI___clone ◆
     __GI___clone (inlined)                                                    ▒
     start_thread                                                              ▒
     module_thread                                                             ▒
     mod_main                                                                  ▒
     flux_reactor_run                                                          ▒
     ev_run                                                                    ▒
   - ev_run                                                                    ▒
      - ev_invoke_pending                                                      ▒
         - 77.46% handle_cb                                                    ▒
            - 77.45% dispatch_message (inlined)                                ▒
               - 77.45% call_handler                                           ▒
                  - 70.65% lookup_cb                                           ▒
                     - 70.61% namespace_monitor                                ▒
                        - namespace_monitor                                    ▒
                           - 70.59% namespace_create (inlined)                 ▒
                              - flux_event_subscribe                           ▒
                                 - 70.57% flux_future_get                      ▒
                                    + flux_future_wait_for                     ▒
                  + 5.59% watcher_cancel_all                                   ▒
                  + 1.21% setroot_cb                                           ▒
         + 16.74% s_item_destroy                                               ▒
            - namespace_destroy                                                ▒
               + 16.74% flux_event_unsubscribe                          
```

Perhaps switching `kvs-watch` to the non-blocking `flux_event_subscribe_ex(3)` will help here, though I can't really be sure until we try it. Note there's another 16% in `flux_event_unsubscribe(3)`, so that should be fixed as well.

chu11:
The `flux_event_subscribe()` call is within `namespace_create()`.  

After a `namespace_create()` it does an asynchronous getroot.

```
    if (!(nsm = zhashx_lookup (ctx->namespaces, ns))) {                                                                                     
        if (!(nsm = namespace_create (ctx, ns)))                                                                                            
            return NULL;                                                                                                                    
        (void)zhashx_insert (ctx->namespaces, ns, nsm);                                                                                     
[...]

PR #6719, opened 2025-03-19
kvs-watch: index watchers to avoid search iterations

chu11:
Per #6690 and #6695, there are two loops in the cancellation path of `kvs-watch`, one that iterates over all namespaces and one over all watchers within a namespace.

This iteration can degrade job throughput when there are a large number of jobs, especially when users may be monitoring jobs through various interfaces, as those interfaces large "watch" the eventlogs.

To improve performance, support a index that maps rpc "matchtags" + uuids to A) the namespace of the watcher and B) the watcher within that namespace.  This allows watch cancellations to go a lot faster.

### DIFF
--- a/src/modules/kvs-watch/kvs-watch.c
+++ b/src/modules/kvs-watch/kvs-watch.c
@@ -1008,15 +1008,14 @@ static void watcher_respond_ns (struct ns_monitor *nsm)
 static void watcher_cancel (struct ns_monitor *nsm,
                             struct watcher *w,
                             const flux_msg_t *msg,
-                            uint32_t matchtag,
                             bool cancel)
 {
-    if (!flux_disconnect_match (msg, w->request))
-        return;
     if (cancel) {
-        uint32_t tag;
-        if (flux_msg_get_matchtag (w->request, &tag) < 0
-            || tag != matchtag)
+        if (!flux_cancel_match (msg, w->request))
+            return;
+    }
+    else {
+        if (!flux_disconnect_match (msg, w->request))
             return;
     }
     w->canceled = true;
@@ -1031,7 +1030,6 @@ static void watcher_cancel (struct ns_monitor *nsm,
  */
 static void watcher_cancel_ns (struct ns_monitor *nsm,
                                const flux_msg_t *msg,
-                               uint32_t matchtag,
                                bool cancel)
 {
     struct watcher *w;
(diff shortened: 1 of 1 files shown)


src/modules/kvs-watch/kvs-watch.c
```diff
@@ -62,6 +63,11 @@ struct commit {
 
 
 /* State for monitoring a KVS namespace.
+ *
+ * N.B. Watchers are stored on a zlistx_t b/c we can iterate over it
+ * and survive entry removals.  That cannot be done with zhashx_t without
+ * retrieving a costly zhashx_keys() list.  Thus, we have watchers on a list
+ * and a separate hash for quick lookup access to watchers.
```
grondo, line 70: Thanks for this comment! :+1:

src/modules/kvs-watch/kvs-watch.c
```diff
@@ -1023,6 +1043,25 @@ static void watcher_cancel (struct ns_monitor *nsm,
...
+    if (flux_msg_unpack (msg, "{s:i}", "matchtag", &matchtag) < 0) {
+        flux_log_error (h, "failed to get matchtag from cancel request");
+        return -1;
+    }
+    if (!(uuid = flux_msg_route_first (msg))) {
+        flux_log_error (h, "failed to get uuid from cancel request");
+        return -1;
+    }
+    snprintf (buf, bufsize, "%s:%u", uuid, matchtag);
+    return 0;
```
grondo, line 1064: return result of `snprintf` or check return code?

src/modules/kvs-watch/kvs-watch.c
```diff
@@ -1115,6 +1117,20 @@ static void watcher_cancel_all (struct watch_ctx *ctx,
     char *name;
     struct ns_monitor *nsm;
 
+    /* if canceling can do lookup to avoid iterating over all
+     * namespaces */
```
grondo, line 1121: wording in comment confused me. Should it be "if canceling, do lookup to avoid iterating..."? (i.e. switch "can" for a comma?)

grondo (approved):
LGTM! Just a couple trivial comments inline.

src/modules/kvs-watch/kvs-watch.c
chu11: do you mean check the return code of snprintf?  We normally don't b/c the buffer can't be overflowed.   Perhaps there was just a subtlety of your comment I'm missing.  Or perhaps I need a comment indicating that the buffer size is more than sufficient?

grondo: Well in this case the size is passed into the function, so there's no real way to know that the buffer can't be overflowed unless you check? However, there's only one caller so a comment would suffice (and maybe cast return code to `(void)` to show it is intentional)

chu11: ahh i get it, it's b/c I moved this code into a function (which it wasn't in a function earlier).  I'll check for overflow.

PR #6726, opened 2025-03-22
support resource counts in `flux queue list` output, display effective limits by default

### LINKED ISSUE #6724: change output of `flux queue list` to report maximum available nodes/cores/gpus in queue instead of `inf`
@kkier asks:

> What're people's thoughts on changing the output of flux queue list so it shows the maximum for at least NNODES reflecting the number of nodes in the queue?

grondo:
This PR leverages the `resources` attribute of `QueueInfo` in `flux queue list` to support emitting _effective_ queue limits in `flux queue list` as mentioned in #6724. 

Then, headers are added for the other `resources.{all,up,down.free,allocated}.{nnodes,ncores,ngpus}` fields so that these are available in output formats. Because it might be useful, a `nodes` output now shows the states of all nodes in the various queues in `flux queue list`:

```
$ src/cmd/flux queue list -o nodes
QUEUE    EN ST  NODES   NODES:UP NODES:DOWN NODES:ALLOC NODES:FREE
pbatch*   ✔  ✔    972        946         26         911         36
pdebug    ✔  ✔     40         40          0           5         35
pci       ✔  ✔      4          4          0           0          4
pall      ✔  ✗   1144       1118         26        1044         75
```

The new `flux queue list` default output now looks like:
```
QUEUE    EN ST TDEFAULT   TLIMIT     NNODES     NCORES      NGPUS
pbatch*   ✔  ✔      12h       1d      0-972    0-93312     0-3888
pdebug    ✔  ✔      30m       1h       0-40     0-3840      0-160
pci       ✔  ✔       1h       4h        0-1      0-384       0-16
pall      ✔  ✗      12h       1d     0-1144   0-109824     0-4576
```

Fixes #6724

### DIFF
--- a/src/cmd/flux-queue.py
+++ b/src/cmd/flux-queue.py
@@ -114,6 +114,19 @@ class FluxQueueConfig(UtilConfig):
                 "{color_started}{started:>2}{color_off} "
                 "{defaults.timelimit!F:>8} "
                 "{limits.timelimit!F:>8} "
+                "{limits.range.effective.nnodes:>10} "
+                "{limits.range.effective.ncores:>10} "
+                "{limits.range.effective.ngpus:>10}"
+            ),
+        },
+        "policy": {
+            "description": "flux-queue list with policy limits only",
+            "format": (
+                "?:{queuem:<8.8} "
+                "{color_enabled}{enabled:>2}{color_off} "
+                "{color_started}{started:>2}{color_off} "
+                "{defaults.timelimit!F:>8} "
+                "{limits.timelimit!F:>8} "
                 "{limits.range.nnodes:>10} "
                 "{limits.range.ncores:>10} "
                 "{limits.range.ngpus:>10}"
--- a/src/cmd/flux-queue.py
+++ b/src/cmd/flux-queue.py
@@ -132,6 +132,19 @@ class FluxQueueConfig(UtilConfig):
                 "{limits.range.ngpus:>10}"
             ),
         },
+        "nodes": {
+            "description": "Show queue status plus all/up/allocated/free nodes",
+            "format": (
+                "?:{queuem:<8.8} "
+                "{color_enabled}{enabled:>2}{color_off} "
+                "{color_started}{started:>2}{color_off} "
+                "{resources.all.nnodes:>6} "
+                "{resources.up.nnodes:>10} "
+                "{resources.down.nnodes:>10} "
+                "{resources.allocated.nnodes:>11} "
+                "{resources.free.nnodes:>10}"
+            ),
+        },
     }
 
     def __init__(self, subcommand):
@@ -262,6 +275,17 @@ class QueueInfoWrapper:
         return AltField("✔", "y") if self.is_started else AltField("✗", "n")
 
 
+def generate_resource_headings():
+    headings = {}
+    for item in ("nnodes", "ncores", "ngpus"):
+        for state in ("all", "free", "allocated", "up", "down"):
+            heading = item[1:].upper()
+            if state != "all":
+                heading += f":{state[:5].upper()}"
+            headings[f"resources.{state}.{item}"] = heading
+    return headings
+
+
 def list(args):
     headings = {
         "queue": "QUEUE",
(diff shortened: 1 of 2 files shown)


src/bindings/python/flux/resource/list.py
```diff
@@ -83,6 +83,12 @@ def free(self):
         res.state = "free"
         return res
 
+    def copy_constraint(self, constraint):
+        """Create a copy of a ResourceSet object based on constraint"""
+        rsets = {}
+        for state in ("all", "down", "allocated"):
+            rsets[state] = self[state].copy_constraint(constraint)
+        return SchedResourceList(rsets)
```
cmoussa1, line 91: commit message nit: "reosurces" --> "resources"

src/bindings/python/flux/queue.py
```diff
@@ -0,0 +1,345 @@
...
+                result = self.config["policy"]["jobspec"]["defaults"]["system"][key]
+            except KeyError:
+                result = default
+        return result
+
+    def policy_system_limit(self, key, default="inf"):
+        try:
+            result = self.config["queues"][self.name]["policy"]["limits"][key]
+        except KeyError:
+            try:
```
cmoussa1, line 236: I _think_ for the exception handling here, you could use `.get()` for the dictionary access and provide a default value in the case where the key-value access fails, and you won't need an inner `try-except` block here, i.e:

```python
result = self.config.get("policy").get("limits").get(key, default)
```

But this is probably no big deal either way.

cmoussa1 (approved):
LGTM!! Nice. One commit message nit and one suggestion about accessing key-value pairs, but it's minor and you can ignore if you like the previous handling!

src/bindings/python/flux/queue.py
cmoussa1: This suggestion also assumes that the previous keys (`"queues"`, `self.name`, `"policy"`, and `"limits"`) are all guaranteed to be present. Otherwise, an `AttributeError` will be raised. So, if those are not guaranteed to be present in the dictionary, sticking with the style you proposed probably makes more sense!

grondo: They are not guaranteed to be present. I guess each `.get()` could return an empty dict, e.g.
```python
result = self.config.get("policy", {}).get("limits", {}).get(key, default)
```

However, this is copying existing code from `flux-queue.py` and it would probably be better to not change too much at once. We can come by and fix this in a future PR. Thanks!

BTW, I should have also mentioned that this PR is based on top of #6723, so you could have ignored the earlier commits from that PR. Sorry!

cmoussa1: No problem at all! Thanks for the clarification. This LGTM then!

src/cmd/flux-queue.py
```diff
@@ -143,12 +143,44 @@ def validate(self, path, conf):
                 raise ValueError(f"{path}: invalid key {key}")
 
 
+class QueueLimitsEffectiveRange:
+    """
+    QueueLimits wrapper which supports a effective range (configured minimum
```
chu11, line 174: "an effective"

src/cmd/flux-queue.py
```diff
@@ -143,12 +143,44 @@ def validate(self, path, conf):
...
+    to effective maximum minimum of limit or all resources in queue).
```
chu11, line 175: is this sentence right?  "maximum minimum"?  Should it be "to effective maximum of all resources in queue"?

grondo:
Thanks @cmoussa1, I'll set MWP.

src/cmd/flux-queue.py
chu11: oh wait I get it now ... maybe "to effective maximum, which is the minimum of the limit or all resources in the queue"

chu11 (commented):
checkpointing review

src/cmd/flux-queue.py
grondo: D'oh! I merged this one after @cmoussa1's review. Can post a separate PR to fix these typos. (Or feel free to do it @chu11!) Thanks!

Issue #6710, opened 2025-03-15
Support for non-v1 jobspecs


jameshcorbett:
Problem: flux-coral2 creates non-v1 jobspecs that have resource sections like:

```
[
  {
    "type": "slot",
    "count": 2,
    "label": "rabbit",
    "with": [
      {
        "type": "node",
        "count": 1,
        "exclusive": true,
        "with": [
          {
            "type": "slot",
            "count": 1,
            "with": [
              {
                "type": "core",
                "count": 1
              }
            ],
            "label": "task"
          }
        ]
      },
      {
        "type": "ssd",
        "count": 256,
        "exclusive": true
      }
    ]
  }
]
```

However it seems as though some core tools might not support them, see e.g.:

```
[corbett8@rzvernal10:~]$ flux job eventlog f3HDHHNxidTM
1741992846.583265 submit userid=31193 urgency=16 flags=0 version=1
1741992846.603245 jobspec-update attributes.system.bank="guests" attributes.system.project="*"
1741992846.603336 validate
1741992846.615981 dependency-add description="dws-create"
1741992846.646039 memo rabbit_workflow="fluxjob-982023661873529856"
1741992850.262999 jobspec-update resources=[{"type":"slot","count":2,"label":"rabbit","with":[{"type":"node","count":1,"exclusive":true,"with":[{"type":"slot","count":1,"with":[{"type":"core","count":1}],"label":"task"}]},{"type":"ssd","count":256,"exclusive":true}]}]
1741992850.263137 dependency-remove description="dws-create"
1741992850.263179 depend
1741992850.263257 priority priority=27657
1741992850.282585 alloc annotations={"user":{"rabbit_workflow":"fluxjob-982023661873529856"}}
1741992850.282871 prolog-start description="job-manager.prolog"
1741992850.282889 exception type="mf_priority" severity=0 note="job.state.run: failed to increment resource count" userid=762
```

grondo:
Thanks that's the `mf_priority` plugin, but I did run the jobspec in question through the flux-core `jj-reader` test program and got:
```
jj-reader: Unsupported resource type 'ssd'
```

Since this is non-V1 jobspec claiming it is v1 jobspec, I wonder what the right solution is?

I guess for now the code can just ignore unknown resource types, but if feels like we should have had a better plan here...

grondo:
@jameshcorbett want to try this patch and see if that resolves these issues? @cmoussa1 we should apply the same patch in flux-accounting.

```diff
[...]

PR #6727, opened 2025-03-22
libjob: rabbit jobspec tests

jameshcorbett:
WIP PR to address #6710. Several of the tests in `t/t0022-jj-reader.t` fail. Some of them because I didn't bother to spend the time to get them right, but others because now `jj-reader` can incorrectly count the nodes. A jobspec like this:
```yaml
version: 1
resources:
  - type: slot
    count: 2
    label: rabbit
    with:
    - type: ssd
      count: 1
    - type: node
      exclusive: true
      count: 1
      with:
      - type: slot
        label: task
        count: 1
        with:
        - type: core
          count: 1
```
should give a nodecount of 2, but it currently gives a nodecount of 1 because the reader doesn't take into account the `count` of the higher-level `slot` entry. That could be easily fixed. However, it will probably get a wrong `slot` count but I don't know how the slot count is supposed to be used so I'm not as sure how to fix it.

@grondo @cmoussa1 I have to step away from development on this issue for a while to get back to SCR work, but I wanted to put it up so you could see what I was thinking. 

I am wondering if `flux-coral2` should drop the use of `jobspec_update` and just update the job-manager's version of the jobspec in-place like it used to, so that only the scheduler sees the updated version (I think that's how it works). That would let us punt on the issue until we have a better way to handle it. But if I'm wrong about only the scheduler seeing the updated jobspec, and in particular if the `mf_priority` plugin sees the in-place-updated jobspec, we'd still hit the same problem and that wouldn't save us anything.


grondo:
> But if I'm wrong about only the scheduler seeing the updated jobspec, and in particular if the mf_priority plugin sees the in-place-updated jobspec, we'd still hit the same problem and that wouldn't save us anything.

Yeah, I don't think it would help. `jobspec-update` events modify the job manager's copy of the jobspec in place, then future callbacks that unpack from `jobspec` key of plugins args reference the updated copy.

We'll need to think of some simple update to correctly count resources in jobspec V1+ or whatever this is. I actually didn't realize that jobspec could have more than one slot. We may need to sit down and devise some rules for a v2 jobspec that scheduling rabbits requires, encode that in an RFC, then update libjj using those rules. Maybe we can discuss this at the next meeting.

jameshcorbett:
After discussion in the meeting yesterday, I think this is ready for review @grondo .

grondo (approved):
LGTM! Thanks!

PR #6731, opened 2025-03-26
test: fix checks-lib script

jameshcorbett:
Problem: the checks-lib script has an incorrect `if` statement, leading to errors.

Fix it.


grondo (approved):
Nice catch! Thanks!

PR #6733, opened 2025-03-27
t: add checkpoint integrity test

### LINKED ISSUE #6718: content(?): missing references from checkpoint
Recently admins attempted to restart an instance recovering with a `content.sqlite` file, i.e. `flux start --recovery=/path/to/mysqlitefiles`.

The `resource.eventlog` could not be read because of missing blobref references.

```
>flux kvs get resource.eventlog
Mar 19 09:46:06.948885 PDT 2025 kvs.err[0]: content_load_completion: content_load_get: No such file or directory
Mar 19 09:46:06.955999 PDT 2025 kvs.err[0]: content_load_completion: content_load_get: No such file or directory
Mar 19 09:46:06.969952 PDT 2025 kvs.err[0]: content_load_completion: content_load_get: No such file or directory
Mar 19 09:46:06.974345 PDT 2025 kvs.err[0]: content_load_completion: content_load_get: No such file or directory
flux-kvs: resource.eventlog: No such file or directory
```

I took all the blobrefs from the eventlog

```
> flux kvs get --treeobj resource.eventlog | jq .data > refs
> head -n 5 refs
[
  "sha1-0aa281cb84b9be3ff07830b3b1b75d88b1c12c74",
  "sha1-b7cf222e01dc6243e5f0012e0a9dcc0bc2aab80f",
  "sha1-3b6fac9a90929180f5198af4937eda3fc88a0aca",
  "sha1-ecf5f25f2173da2d858e900560795d82c6be0349",
```

edit by hand to remove quotes and what not, then lets try to load every ref

`for i in `cat refs`; do flux content load $i; done >& thelogs`

the content load didn't work for 4 refs

```
> wc -l thelogs
9733 thelogs
>grep -n "error loading" thelogs
6808:flux-content: error loading blob: No such file or directory
7447:flux-content: error loading blob: No such file or directory
8883:flux-content: error loading blob: No such file or directory
9462:flux-content: error loading blob: No such file or directory
```

I had expected only the most recent references to be missing (i.e. closer to line 9733), but the refs that couldn't be read are distributed semi-randomly around.  Based on the timestamps surrounding the errors, the missing refs are scattered over 2 weeks, days apart, with the first one about 16 days before the broker was killed.  In 2 cases the surrounding timestamps are ~1s apart.  In the other 2 cases they are atleast several minutes between.  So this doesn't appear to be related to some sudden influx of resource eventlog entries.

Given this random distribution of missing refs, my concern is we have a bug somewhere in flux where we are not flushing some references to disk.  Best guess is that this is in the `content` module.  Given that this is distributed somewhat randomly throughout time, it suggests this is not related to checkpoint code forgetting to ask `content.flush` to flush. (i.e. over many days we should have called `content.flush` many times)

(Edit: it's worth noting that according to `admin.eventlog` the broker had been up for about 1 month.   So no strange restarts or something in the middle.)

The traditional guess would be some racy thingie where some stuff isn't flushed to disk.  Perhaps something is not put on a "dirty" list or entries are timed out and uncached before they are flushed.

First things first, it'd be good to get some tests to emulate this situation.  Perhaps set the kvs checkpoint period to something small like < 1s, have some code spin and constantly append to some kvs key, and then SIGKILL the broker after some seconds, and then make sure the broker can be recovered and the key can be read.

An aside 1: I think there have been instances before where we've had errors recovering but only job data was lost.  I think the assumption was sqlite corruption.  But perhaps it wasn't.  Perhaps at elcap sizes and the massive influx of appends to `resource.eventlog`, this is simply the first time we have some suggestion it is due to flux.

An Aside 2: I am reluctant to believe this is a sqlite issue.  sqlite journaling w/ WAL and synchronous w/ NORMAL is supposed to be safe.  And the data that is missing is many days ago.

chu11:
Problem: There are no tests to ensure that checkpoints work as intended when the broker is killed or it suddenly crashes.

Add some testing in a new t1012-kvs-checkpoint.t test.

Closes #6718

### DIFF
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -108,6 +108,7 @@ TESTSCRIPTS = \
 	t1009-kvs-copy.t \
 	t1010-kvs-commit-sync.t \
 	t1011-kvs-checkpoint-period.t \
+	t1012-kvs-checkpoint.t \
 	t1101-barrier-basic.t \
 	t1102-cmddriver.t \
 	t1103-apidisconnect.t \
@@ -435,6 +436,7 @@ check_PROGRAMS = \
 	kvs/watch_disconnect \
 	kvs/watch_stream \
 	kvs/commit \
+	kvs/loop_append \
 	kvs/transactionmerge \
 	kvs/lookup_invalid \
 	kvs/commit_order \
@@ -637,6 +639,11 @@ kvs_commit_SOURCES = kvs/commit.c
 kvs_commit_CPPFLAGS = $(test_cppflags)
 kvs_commit_LDADD = $(test_ldadd)
 kvs_commit_LDFLAGS = $(test_ldflags)
+
+kvs_loop_append_SOURCES = kvs/loop_append.c
+kvs_loop_append_CPPFLAGS = $(test_cppflags)
+kvs_loop_append_LDADD = $(test_ldadd)
+kvs_loop_append_LDFLAGS = $(test_ldflags)
 
 kvs_transactionmerge_SOURCES = kvs/transactionmerge.c
 kvs_transactionmerge_CPPFLAGS = $(test_cppflags)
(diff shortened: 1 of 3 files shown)


grondo (approved):
This looks reasonable. Thanks!

Issue #6729, opened 2025-03-25
signals to batch/alloc jobs are not delivered to the batch script


grondo:
Problem: Non-deadly signals (effectively `SIGUSR1`, `SIGUSR2`) are caught by the broker and forwarded to jobs (commit aecfd32c14). This means these signals cannot be delivered directly to the batch script.

The broker should send these signals to the batch script as well as any jobs, to allow for use cases where `SIGUSR1/2` are used to have the batch script do something, such as resubmit a copy of itself, checkpoint, etc.

Issue #6725, opened 2025-03-21
background process in an initial program leaves broker running without a connector


grondo:
A user discovered that a background process run from an initial program (batch script) was keeping the batch job in RUN state, but they could not connect to the job with `flux proxy`.

This is easily reproducible with 
```console
$ flux batch -n1  --wrap 'sleep 60 &'
f5gKbygaj
$ flux proxy f5gKbygaj
tuolumne1001: local:///var/tmp/grondo/flux-eFhB8J/local-0: No such file or directory
flux-proxy: ssh://tuolumne1001/var/tmp/grondo/flux-eFhB8J/local-0: Invalid argument
```

My guess is that the exit of the batch script starts r3 (the shutdown process), but the broker is hung at the end because `sleep` is keeping the stdio file descriptors open.

grondo:
Ok, this should have been obvious. The batch instance exits in the reproducer above, there is no longer any broker. The _job shell_ sticks around waiting for stdio of the background process (now orphaned).

grondo:
This one's not going to be that easy to fix. It may require rethinking how the job execution and shell use process groups.

I think an ideal solution would be for the broker to send SIGHUP to the rc2 process group when the process group leader exits. However, this approach has a race condition, since the pgrp id could be reused by the kernel by the time the signal is delivered, and we'd SIGHUP the wrong process(es).

The race-free way to handle this would be to run rc2 in the _broker's_ process group, and have it SIGHUP its own process group before exiting. However, this breaks the solution for #5105 -- since the job shell uses `killpg(2)` by default, the entire process group is signaled, including rc2, which doesn't give the broker a chance to forward signals to jobs.

What we really need is for the job shell to use normal `kill(2)` when signaling the broker, but still use `killpg(2)` for jobs. I'm not quite sure how to accomplish that, besides having the broker somehow tell the job shell it would like this behavior.

More thought required.

grondo:
If the broker runs rc2 in its own process group as suggested above, then this would address #6729. Signals would be sent to the broker _and_ the initial program, which now seems like the expected behavior. As noted, this will break some tests for #5105, but my current opinion is that the tests should be fixed.

Addressing this note in #5105:

> (It perhaps doesn't make sense to forward signals to the initial program because unlike other batch systems, a Flux initial program could be running a much wider range of potential processes, which may not handle whatever signal is propagated)

[...]

PR #6734, opened 2025-03-28
flux-job: forward some signals to jobs in `flux job attach`

grondo:
It seems like perhaps an oversight that `flux job attach` gets terminated by signals like `SIGTERM`, `SIGHUP`, `SIGALRM` and `SIGUSR[12]` instead of forwarding them to jobs. 

If changes proposed in #6725 and #6729 go into effect, we don't want a `flux job attach` to just exit immediately when a user sends SIGUSR1 to their batch job. Instead, this PR lets `flux job attach` handle the signal and forward it to the job. This is probably the expected behavior for this particular command.

### DIFF
--- a/src/cmd/job/attach.c
+++ b/src/cmd/job/attach.c
@@ -27,6 +27,7 @@
 #include "src/common/libczmqcontainers/czmq_containers.h"
 #include "src/common/libutil/xzmalloc.h"
 #include "src/common/libutil/log.h"
+#include "src/common/libutil/sigutil.h"
 #include "src/common/libeventlog/eventlog.h"
 #include "src/common/libeventlog/formatter.h"
 #include "src/common/libioencode/ioencode.h"
@@ -89,6 +90,20 @@ struct optparse_option attach_opts[] = {
       .usage = "Set MPIR_being_debugged for testing",
     },
     OPTPARSE_TABLE_END
+};
+
+struct fwd_signal {
+    int signum;
+    flux_watcher_t *w;
+};
+
+struct fwd_signal attach_fwd_signals[] = {
+    { .signum = SIGTERM },
+    { .signum = SIGHUP },
+    { .signum = SIGALRM },
+    { .signum = SIGUSR1 },
+    { .signum = SIGUSR2 },
+    { 0 },
 };
 
 struct attach_ctx {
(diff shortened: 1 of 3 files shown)


src/cmd/job/attach.c
```diff
@@ -1284,6 +1284,21 @@ static void initialize_attach_statusline (struct attach_ctx *ctx,
...
+static void setup_signal_handlers (struct attach_ctx *ctx, flux_reactor_t *r)
+{
+    ctx->sigint_w = flux_signal_watcher_create (r,
+                                                SIGINT,
+                                                attach_signal_cb,
+                                                &ctx);
+    ctx->sigtstp_w = flux_signal_watcher_create (r,
+                                                 SIGTSTP,
+                                                 attach_signal_cb,
+                                                 &ctx);
```
jameshcorbett: Should be `ctx` here and not `&ctx` I think?

src/cmd/job/attach.c
```diff
@@ -1302,17 +1338,42 @@ static void initialize_attach_statusline (struct attach_ctx *ctx,
...
+                          sigutil_signame (sig->signum));
+        flux_watcher_start (sig->w);
+        flux_watcher_unref (sig->w);
+        sig++;
+    }
+}
+
+static void fwd_signals_destroy (struct fwd_signal *fwd)
+{
+    struct fwd_signal *sig = fwd;
```
jameshcorbett, line 1372: is this line needed?

jameshcorbett (approved):
Looks good, two small comments.

src/cmd/job/attach.c
grondo: Good point! will fix.

src/cmd/job/attach.c
grondo: D'oh. I'm not sure how this got in here, it would cause a segfault in the tests. This could be a rebase error. Anyway, will fix.

grondo: Good catch btw! Thanks!

jameshcorbett: I think it is just a rebase error, because it's fixed in a later commit.

grondo: Thanks. I'll fix that.

grondo: Done! Will set MWP. Thanks.

PR #6740, opened 2025-03-31
execute rc2 in same process group as broker and send `SIGHUP` before broker exit

grondo:
This PR addresses #6725 and #6729, which both involve how the broker manages rc2 and associated processes.

The main change here is that rc2 (the broker initial program) is run in the same process group as the broker when the broker is a process group leader. This change addresses #6729 since the job shell uses `killpg(2)`, so signals are delivered to the broker _and_ the initial program/batch script. In case this ends up being a problem for some use cases, the old behavior can be restored by setting the `broker.rc2_pgrp` broker attribute.

#6725 is then addressed by having the broker send `SIGHUP` to its own process group on exit. This is similar to what the kernel would do when a session leader exits to cleanup processes still using a terminal. This effectively terminates any background processes that were launched from a batch script or initial program.

It may have been better to have these two changes in two separate PRs, but it turns out that just the first change makes the test suite less reliable, particularly the `t2808-shutdown` test, so it seems better if we take both changes at once.

### DIFF
--- a/doc/man7/flux-broker-attributes.rst
+++ b/doc/man7/flux-broker-attributes.rst
@@ -134,6 +134,12 @@ broker.rc3_path [Updates: C]
 broker.rc2_none [Updates: C]
    If set, do not run an initial program.
 
+broker.rc2_pgrp [Updates: C]
+   By default, rc2 will be placed in the same process group as the
+   broker whenever the broker is itself a process group leader. If the
+   ``broker.rc2_pgrp`` attribute is set, then rc2 will always be placed
+   in its own process group.
+
 broker.exit-restart [Updates: C, R]
    A numeric exit code that the broker uses to indicate that it should not be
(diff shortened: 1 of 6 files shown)


chu11:
Regarding this commit message comment: `The broker could send a signal to the rc2 process group at exit, but this is unsafe since in the common case all processes exit the process group and thus the process group id could be reused.`, if we're exiting the broker, why is the reuse of a process group id something we'd be concerned about?

chu11:
> If you send a signal to a process group after the process group leader of that process group has exited, you have no way of knowing if the PGID or PID has been reused between when the exit occurred and the signal is sent.

Oh I get it ... it's possible the group leader has already exited before we send the signal?  This could especially be the case if the rc2 is launching background processes?

chu11 (approved):
LGTM!

grondo:
Thanks! I'll set MWP.

PR #6741, opened 2025-03-31
flux-proxy: fall back to `/tmp` when `TMPDIR` is invalid

### LINKED ISSUE #6737: flux-proxy: improve error message when `TMPDIR` does not exist (or just fall back to `/tmp`?)
The error from `flux proxy` when `TMPDIR` doesn't exist could be more helpful:
```console
$ TMPDIR=/foo flux proxy fkmmagGn6b1
flux-proxy: error creating proxy socket directory: No such file or directory
```
Probably `TMPDIR=/foo does not exist` would point the user in the right direction.

Or perhaps the command could just fall back to `/tmp` with a warning?

grondo:
This PR fixes #6737 by having `flux proxy` fall back to `/tmp` with a warning when `TMPDIR` is not a valid directory.

### DIFF
--- a/src/cmd/builtin/proxy.c
+++ b/src/cmd/builtin/proxy.c
@@ -352,6 +352,14 @@ static int comms_error (flux_t *h, void *arg)
     return 0;
 }
 
+static bool is_directory (const char *path)
+{
+    struct stat st;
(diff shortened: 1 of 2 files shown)


chu11 (approved):
LGTM!

PR #6746, opened 2025-04-02
t: fix potential race in backgrounded processes

### LINKED ISSUE #6745: t: audit and remove && chained background processes
In a number of locations this pattern exists

```
do something &&
do something &&
process_to_be_backgrounded &
pid=$!
...
wait $pid # or kill $pid, etc.
```

this is not safe.  bash `$!` retrieves the PID of the most recently backgrounded processes.  B/c of the `&&` chain, this can be racy as the pid we desire may not have been backgrounded yet.

chu11:
Problem: The following pattern exists in a number of tests

do-something &&
do-something &&
run_a_process_in_background &
pid=$! &&
...
wait $pid / kill $pid / etc.

This is racy and unsafe.  In bash $! returns the pid of the most
recent backgrounded process, but with the previous line being && chained,
it is possible the backgrounding of the process has not yet occurred and
an incorrect PID is picked up by $!.

Solution:  Remove the && chain for any command before a backgrounded
process.  In one situation, remove the workaround put in place to work
around this.

Fixes #6745

### DIFF
--- a/t/t0009-dmesg.t
+++ b/t/t0009-dmesg.t
@@ -52,7 +52,7 @@ test_expect_success 'flux dmesg -c prints and clears' '
 	! flux dmesg | grep -q hello_dmesg_pc
 '
 test_expect_success NO_CHAIN_LINT 'flux dmesg -f works' '
-	flux logger hello_last &&
+	flux logger hello_last
 	flux dmesg -f > dmesg.out &
 	pid=$! &&
 	$waitfile -t 20 -p hello_last dmesg.out &&
@@ -61,7 +61,7 @@ test_expect_success NO_CHAIN_LINT 'flux dmesg -f works' '
 	kill $pid
 '
 test_expect_success NO_CHAIN_LINT 'flux dmesg -f --new works' '
-	flux logger hello_old &&
+	flux logger hello_old
 	flux dmesg -f --new > dmesg2.out &
 	pid=$! &&
 	for i in $(seq 1 10); do flux logger hello_new; done &&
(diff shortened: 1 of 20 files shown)


t/t2712-python-cli-alloc.t
```diff
@@ -94,19 +94,12 @@ test_expect_success 'flux alloc --bg fails if rc1 fails' '
...
-	echo $! >sigint.pid
-}
 waitfile=$SHARNESS_TEST_SRCDIR/scripts/waitfile.lua
 test_expect_success NO_CHAIN_LINT 'flux alloc --bg can be interrupted' '
 	flux queue stop &&
-	test_when_finished "flux queue start" &&
-	run_mini_bg &&
+	test_when_finished "flux queue start"
+	flux alloc --bg -n1 -v >sigint.log 2>&1 &
+	echo $! >sigint.pid &&
```
grondo, line 102: I'm wondering the reasoning behind changing a working test (that also had a nice comment)?
The commit message says "remove a workaround" but removing the `&&` is also just a different workaround...

t/t0009-dmesg.t
```diff
@@ -52,7 +52,7 @@ test_expect_success 'flux dmesg -c prints and clears' '
 	! flux dmesg | grep -q hello_dmesg_pc
 '
 test_expect_success NO_CHAIN_LINT 'flux dmesg -f works' '
-	flux logger hello_last &&
```
grondo, line 55: Commit message: extra space after `Solution: `

grondo (approved):
LGTM! Maybe this will fix some spurious CI failures!

t/t2712-python-cli-alloc.t
chu11: Fair point.  My assumption is that the test did not work b/c the author did not know about the && chain race.  But it is possible there's another subtlety in there I'm unaware of.  I'll put back the original.

PR #6747, opened 2025-04-03
doc: add troubleshooting guide

grondo:
This is a first attempt at a troubleshooting guide that acts as a guide for using lesser-known Flux commands and debugging techniques when jobs are not behaving as expected. I struggled with how to present this information, but ended up going with a guide organized by job state, so it also somewhat documents the job lifecycle as well.


doc/guide/troubleshooting.rst
```diff
@@ -0,0 +1,448 @@
...
+*************
+
+Once all :option:`prolog-start` events have a corresponding
+:option:`prolog-finish` event, the job manager sends a start request to
+the job execution system. The :command:`job-exec` module launches job
+shells (via the IMP on a multi-user system). Once all shells have started
+a :command:`start` event is posted to the main job eventlog.
+
+When the start event is delayed, the :command:`job-exec` module can
+be queried for the job to get some detail:
```
chu11, line 303: most places you don't `:command:`MODULE NAME``, intentionally wanted to do it for `job-exec`?  I'm assuming just a formatting error.

chu11 (approved):
overall LGTM as a good start!  Just one formatting nit I noticed.

doc/guide/troubleshooting.rst
grondo: It's actually a good question what we should use here. I know the bare minimum about sphinx/rst, I just know I don't want red text like double-backticks give you. Perhaps there's a way we can make our own "domains" for modules, job states, etc. But for now I'll make it consistent and fix this case (you're right it was just an accident in this case)

grondo: I settled on using `:command:` only for commands (and DSOs) and `:option:` for everything else. If this was incorrect, it should be easy to come along and fix it later.

PR #6748, opened 2025-04-03
libeventlog: fix flaky eventlog formatter tests

grondo:
The eventlog formatter tests compare eventlog entry JSON strings directly and are therefore susceptible to encoding quirks such as precision and rounding of floating-point numbers. This was causing the MacOS build to fail.

Convert the input JSON back to a string before the comparison to work around representation differences in the current jansson implementation.


chu11 (approved):
LGTM!

grondo:
Thanks!

PR #6752, opened 2025-04-07
testsuite: fix flux-proxy TMPDIR test

grondo:
Problem: The test for handling an invalid TMPDIR in flux-proxy contains a race: The batch instance may not have posted the uri memo by the time flux-proxy runs. Not only that, but somehow setting the invalid TMPDIR value was dropped at some point, so the test doesn't even test anything.

Fix the race condition using `flux uri --wait`.

Restore the invalid TMPDIR so the test checks the intended behavior.


grondo:
Thanks! Setting MWP.

PR #6755, opened 2025-04-08
libeventlog: fix uncalled error callback

### LINKED ISSUE #6753: libeventlog: call error callback on correct batch entries
Doing some other work, noticed that the error callback in the `eventlogger` does not get called.  The reason is `ev->current` is set to NULL in `eventlogger_commit()`.  So if an errorhappens, `eventlogger_batch_get()` is called to get the "current batch" for error handling, but that "current batch" is empty.

This hack made things work ... will ponder the better solution.  Alternately, it may be useful to have a callback for a "singular" error callback, vs iterating on all batch entries.

```
--- a/src/common/libeventlog/eventlogger.c
+++ b/src/common/libeventlog/eventlogger.c
@@ -187,6 +187,15 @@ flux_future_t *eventlogger_commit (struct eventlogger *ev)
     return eventlogger_commit_batch (ev, batch);
 }
 
+flux_future_t *eventlogger_commit2 (struct eventlogger *ev, struct eventlog_batch **batchp)
+{
+    struct eventlog_batch *batch = ev->current;
+
+    ev->current = NULL;
+    *batchp = batch;
+    return eventlogger_commit_batch (ev, batch);
+}
+
 static void timer_commit_cb (flux_future_t *f, void *arg)
 {
     struct eventlog_batch *batch = arg;
@@ -383,14 +392,15 @@ int eventlogger_append_pack (struct eventlogger *ev,
 
 int eventlogger_flush (struct eventlogger *ev)
 {
+    struct eventlog_batch *batch;
     int rc = -1;
     flux_future_t *f;
 
-    if (!(f = eventlogger_commit (ev))
+    if (!(f = eventlogger_commit2 (ev, &batch))
         || flux_future_wait_for (f, ev->commit_timeout) < 0)
         goto out;
     if ((rc = flux_future_get (f, NULL)) < 0)
-        eventlog_batch_error (eventlog_batch_get (ev), errno);
+        eventlog_batch_error (batch, errno);
 out:
     flux_future_destroy (f);
     return rc;

```

chu11:
Problem: The error callback in libeventlog's eventlogger is not called when there is an error.  The issue arises from the fact that the "current" batch is reset after a call to eventlogger_commit(). If an error occurs, eventlog_batch_get() is called to get the "current" batch, which was just reset and is empty.  So no error callback is issued.

Create a new wrapper function commit_batch() that is identical to eventlogger_commit(), but also returns the value of the recently committed batch.  Use this value instead of calling eventlog_batch_get().

Fixes #6753

### DIFF
--- a/src/common/libeventlog/eventlogger.c
+++ b/src/common/libeventlog/eventlogger.c
@@ -179,12 +179,19 @@ static flux_future_t *eventlogger_commit_batch (struct eventlogger *ev,
     return f;
 }
 
-flux_future_t *eventlogger_commit (struct eventlogger *ev)
+static flux_future_t *commit_batch (struct eventlogger *ev,
+                                    struct eventlog_batch **batchp)
 {
     struct eventlog_batch *batch = ev->current;
-
+    if (batchp)
+        *batchp = batch;
     ev->current = NULL;
     return eventlogger_commit_batch (ev, batch);
+}
+
+flux_future_t *eventlogger_commit (struct eventlogger *ev)
+{
+    return commit_batch (ev, NULL);
 }
 
 static void timer_commit_cb (flux_future_t *f, void *arg)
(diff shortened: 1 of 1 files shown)


grondo (approved):
Nice catch!

PR #6756, opened 2025-04-09
libeventlog: cleanup failed batch in eventlogger

### LINKED ISSUE #6754: libeventlog/eventlogger: batches that result in error cannot be cleaned up
When a "batch" is commited in the eventlogger it is put on the "pending list".

It is destroyed from the "pendling list" if the commit is completed.

If an error occurs (e.g. the error callback is called), AFAICT, nothing can clean this up with the exception of the entire eventlogger being destroyed.

I think in `eventlog_batch_error()`, we should probably call `zlist_remove (ev->pending, batch);` after the error callbacks have been completed.  Although probably gotta think about this a bit more.

chu11:
Problem: If a batch commit fails in the eventlogger the batch is never cleaned up until the eventlogger is destroyed.  It sits on the pending list forever.

Cleanup the batch in eventlogger_batch_error() to reclaim that memory.

Fixes #6754

----

Side note, this was found only via code review.   Also, #6755 is not required to be merged first, as this is independent of that.  However, #6755 will ensure one more case of "batch error" is handled correctly.

### DIFF
--- a/src/common/libeventlog/eventlogger.c
+++ b/src/common/libeventlog/eventlogger.c
@@ -136,6 +136,8 @@ static void eventlog_batch_error (struct eventlog_batch *batch, int errnum)
         (*ev->ops.err) (ev, ev->arg, errnum, entry);
         entry = zlist_next (batch->entries);
     }
+    /*  zlist_remove destroys batch */
+    zlist_remove (ev->pending, batch);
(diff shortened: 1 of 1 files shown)


grondo (approved):
Another good catch!

PR #6760, opened 2025-04-10
task-exit: improve clarity of log message

### LINKED ISSUE #6749: users find the term "doom" in the exit-timeout exception extremely confusing
Reports from sysadmins is that the users find the `doom: ` prefix on exit-timeout and exit-on-error exceptions confusing.
These errors are prefixed with the name of the shell plugin, so we'll just have to rename it. Probably as easy as something like:

```diff
diff --git a/src/shell/doom.c b/src/shell/doom.c
index acec67abf..a4848f5e0 100644
--- a/src/shell/doom.c
+++ b/src/shell/doom.c
@@ -24,7 +24,7 @@
  *   Raise the fatal exception immediately if the first task fails,
  *   e.g. calls exit with a nonzero value or is terminated by signal.
  */
-#define FLUX_SHELL_PLUGIN_NAME "doom"
+#define FLUX_SHELL_PLUGIN_NAME "task-exit"
```

To shield users from ominous vocabulary.

garlick:
Problem: it has been requested that we kindly excise the word "doom" from the exit timeout message because it is "EXTREMELY confusing" [sic].

OK. Rename the doom shell plugin to task-exit so that the log message prefix isn't so distracting.

In case the message was actually unclear for other reasons, add the word "task" before "rank" to make sure we communicate that a task exited and not a broker, shell, or other ranked entity.

Fixes #6749

### DIFF
--- a/src/shell/doom.c
+++ b/src/shell/doom.c
@@ -24,7 +24,7 @@
  *   Raise the fatal exception immediately if the first task fails,
  *   e.g. calls exit with a nonzero value or is terminated by signal.
  */
-#define FLUX_SHELL_PLUGIN_NAME "doom"
+#define FLUX_SHELL_PLUGIN_NAME "task-exit"
 
 #if HAVE_CONFIG_H
 #include "config.h"
@@ -126,9 +126,9 @@ static void doom_check (struct shell_doom *doom,
 
     if (doom->exit_on_error && doom->exit_rc != 0) {
         shell_die (doom->exit_rc,
-                   "%s: %srank %d on host %s failed and exit-on-error is set",
+                   "%s: %s rank %d on host %s failed and exit-on-error is set",
                    get_jobspec_command_arg0 (doom),
-                   doom->lost_shell ? "shell " : "",
+                   doom->lost_shell ? "shell" : "task",
                    doom->exit_rank,
                    doom_exit_host (doom));
     }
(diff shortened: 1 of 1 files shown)


src/shell/doom.c
```diff
@@ -126,9 +126,9 @@ static void doom_check (struct shell_doom *doom,
 
     if (doom->exit_on_error && doom->exit_rc != 0) {
         shell_die (doom->exit_rc,
-                   "%s: %srank %d on host %s failed and exit-on-error is set",
+                   "%s: %s rank %d on host %s failed and exit-on-error is set",
                    get_jobspec_command_arg0 (doom),
-                   doom->lost_shell ? "shell " : "",
+                   doom->lost_shell ? "shell" : "task",
```
chu11, line 131: nit: "task" -> "task on"?  there's only one shell but there can be multiple tasks?

src/shell/doom.c
```diff
@@ -24,7 +24,7 @@
  *   Raise the fatal exception immediately if the first task fails,
  *   e.g. calls exit with a nonzero value or is terminated by signal.
  */
-#define FLUX_SHELL_PLUGIN_NAME "doom"
+#define FLUX_SHELL_PLUGIN_NAME "task-exit"
```
chu11, line 27: my anal retentiveness would want to rename the file too ... but that's me.

garlick: I was going to do that and started replacing the `doom_` prefixed functions with `task_exit_`, and then realized that makes a mess of some function names like

doom_exit_host => task_exit_exit_host
doom_task_exit => task_exit_task_exit

So I said "bah!" and left it as is.

chu11 (approved):
APPROVED!

garlick:
Just pushed a small update - I didn't catch that similar messages were generated in two places in the plugin.   If still OK @chu11, I'll set MWP.

src/shell/doom.c
chu11: Oh it just occurred to me, your commit message says to rename the plugin "exit-timeout" but this is "task-exit".  I think "exit-timeout" is probably better, since it could be a timeout do to the shell and not a task per se.

chu11:
second change LGTM, although note my recent comment about commit message vs plugin name inconsistency.

src/shell/doom.c
garlick: Ah thanks, I'll fix.  The plugin doesn't always employ a timeout, but it's always triggered by task exit (or by proxy - if a shell exits its tasks have exited).  I think task-exit is probably the right name.

PR #6762, opened 2025-04-11
sdmon: only match job related systemd units

### LINKED ISSUE #6761: sdmon: glob for orphaned work is too inclusive
Problem: the sys admins noted that `flux resource status` is showing mangement nodes as offline.

Since #6616, we don't mark nodes as "up" (for scheduling) until sdmon declares that no work is running on the node.  However, its glob is too inclusive when it queries the system systemd instance.  It's matching the glob `flux-*` which is suposed to match `flux-housekeeping.service`, `flux-epilog.service`, and `flux-prolog.service`.  Unfortunately it also matches `flux-accounting.service` and `flux-coral2-dws.service`. Doh!
```
[root@tioga6:garlick]# flux module stats sdmon
{
 "units": [
  {
   "name": "flux-coral2-dws.service",
   "state": "active.running"
  },
  {
   "name": "flux-accounting.service",
   "state": "active.running"
  }
 ]
}
```
So rank 0 will never be "idle" and cannot join the `sdmon.online` group that the resource module is monitoring.

Probably sdmon should just use exact matches on the services of interest.

garlick:
Problem: sometimes rank 0 shows as offline in flux resource status output.

This is because an sdmon unit name glob is too inclusive and finds systemd units for flux-accounting and flux-coral2.

Add addional unit name filtering.

Fixes #6761

### DIFF
--- a/src/modules/sdmon/sdmon.c
+++ b/src/modules/sdmon/sdmon.c
@@ -30,6 +30,7 @@
 #include <jansson.h>
 #include <flux/core.h>
 #include "ccan/str/str.h"
+#include "ccan/array_size/array_size.h"
 #include "src/common/libutil/errprintf.h"
 #include "src/common/libutil/errno_safe.h"
 #include "src/common/libutil/basename.h"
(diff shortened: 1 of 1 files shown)


garlick:
Thanks! I'll set MWP.

PR #6763, opened 2025-04-11
enable daily backups of system KVS

### LINKED ISSUE #6759: automate daily backups of KVS
Problem: when all else fails, it would be nice to have a backup of the KVS that's no more than a day old to recover the `resource.eventlog` and other data.

Perhaps we could set up a nightly flux-cron job that writes a checkpoint, then creates a dump file from it.

We already have a  [tmpfiles config](https://github.com/flux-framework/flux-core/blob/master/etc/flux.conf) for removing dump files in `/var/lib/flux/dump` older than 30 days.  Maybe it would be OK tighten this up to 7 days or so.

@grondo tried dumping on tuo yesterday to ensure it didn't slow things down too much, and it seemed OK.

garlick:
Problem: the kvs and its content storage layer currently have some reliability issues, especially in long running system instances of Flux.
    
As a stopgap while those issues are being worked, add a nightly KVS dump for the system instance of Flux.  Backups will accumulate in `/var/lib/flux/dump` using the same naming convention as the automatic dumps taken at shutdown.  The existing tmpfiles configuration removes old dump files, keeping storage usage in check.
    
Fixes #6759

### DIFF
--- a/etc/Makefile.am
+++ b/etc/Makefile.am
@@ -7,10 +7,13 @@ systemdsystemunit_DATA = \
 #endif
 
 tmpfilesdir = $(prefix)/lib/tmpfiles.d
+crondir = $(fluxconfdir)/system/cron.d
 fluxrc1dir = $(fluxconfdir)/rc1.d
 
 tmpfiles_DATA = flux.conf
 
+dist_cron_DATA = \
+	kvs-backup.cron
 
 dist_fluxconf_SCRIPTS = \
         rc1 \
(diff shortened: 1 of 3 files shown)


chu11 (approved):
LGTM.  Only minor thought is to ask an admin if 3AM is good ... although maybe you already did ask.

garlick:
I didn't ask, but @grondo did verify that dump doesn't have a big impact (expected since it is not optimized for speed - it doesn't get more than one content RPC going at one time).

PR #6764, opened 2025-04-11
drop content.backing-module=none support

garlick:
Problem: there is not a strong use case for running the KVS without any backing store module, yet supporting it is complicated.

Drop support in rc1 and rc3 for a backing module of "none". Drop a sharness test that verifies that this mode works.

Note: this was peeled off of work to eliminate the backing store registration protocol.  This seems like a good step on its own though, since it drops a use case that would otherwise need to be considered when designing other content layer changes.


chu11 (approved):
LGTM.  As an aside, I had thought we had at one point had "none" as the default backing store for subinstances.  But looking through old PRs, it appears we never pushed that through.  So this shouldn't affect any users.

garlick:
Thanks.  Yeah I think I had done testing and found no performance benefit over sqlite so it was not proposed.

I'll set MWP

PR #6772, opened 2025-04-18
content-sqlite: store history of checkpoints

### LINKED ISSUE #6629: kvs/content: checkpoint should not overwrite current checkpointed reference
Currently the checkpoint is saved to the same location, overwriting whatever was already there.

In the event something goes south w/ a final checkpoint, there is no way to recover with a prior checkpoint.

There should be a rolling history of checkpoints, in the event the last checkpoint failed.

Some complexity here b/c not only do we support the rolling history, any mechanisms that recover from checkpoint may need to "go back in history" to earlier one.

chu11:
Problem: The content-sqlite backend only stores one checkpoint for each key, overwriting older keys/checkpoints.  If anything
goes wrong with a checkpoint, there is no information stored to allow a recovery from an earlier point in time.

Instead of overwriting checkpoints, store a rolling history of the checkpoints.  Support configuration of the maximum number of checkpoints via a new configuration parameter.  The default maximum number of checkpoints is 100 per key.

Fixes #6629

----

notes:

- for this PR the new tests are in `t0012-content-sqlite-checkpoints.t` b/c I didn't want to create new numbered content-sqlite test files super far away from `t0012-content-sqlite.t` ... but maybe I should.  Or should we do a good re-numbering on tests and shifting everything down.

- you'll notice a big comment that discusses why we use a string timestamp rather than a numeric timestamp in SQL.  I can't help but wonder if prior discussions of vendoring sqlite and bringing it into flux-core could be warranted (https://sqlite.org/amalgamation.html).

### DIFF
--- a/src/modules/content-sqlite/content-sqlite.c
+++ b/src/modules/content-sqlite/content-sqlite.c
@@ -47,14 +47,29 @@ const char *sql_store = "INSERT INTO objects (hash,size,object) "
                         "  values (?1, ?2, ?3)";
 const char *sql_objects_count = "SELECT count(1) FROM objects";
 
-const char *sql_create_table_checkpt = "CREATE TABLE if not exists checkpt("
-                                       "  key TEXT UNIQUE,"
-                                       "  value TEXT"
-                                       ");";
-const char *sql_checkpt_get = "SELECT value FROM checkpt"
-                              "  WHERE key = ?1";
-const char *sql_checkpt_put = "REPLACE INTO checkpt (key,value) "
-                              "  values (?1, ?2)";
+const char *sql_checkpt_get_v1 = "SELECT value FROM checkpt"
+                                 "  WHERE key = ?1";
+
+const char *sql_drop_checkpt = "DROP TABLE IF EXISTS checkpt";
+
+const char *sql_create_table_checkpt_v2 =
+    "CREATE TABLE if not exists checkpt_v2("
+    "  id INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,"
+    "  value TEXT"
+    ");";
+const char *sql_checkpt_get_v2 = "SELECT value FROM checkpt_v2"
+                                 " ORDER BY id DESC LIMIT 1";
+const char *sql_checkpt_put_v2 = "INSERT INTO checkpt_v2 (value)"
+                                 " values (?1)";
+const char *sql_checkpt_prune =
+    "DELETE FROM checkpt_v2"
+    " WHERE id IN ("
+    "   SELECT id FROM checkpt_v2 ORDER BY id DESC LIMIT -1 OFFSET ?1"
+    " );";
+
+const char *sql_table_list = "SELECT tbl_name FROM sqlite_master where type = 'table'";
+
+#define MAX_CHECKPOINTS_DEFAULT 5
 
 struct content_stats {
     tstat_t load;
@@ -69,6 +84,7 @@ struct content_sqlite {
     sqlite3_stmt *store_stmt;
     sqlite3_stmt *checkpt_get_stmt;
     sqlite3_stmt *checkpt_put_stmt;
+    sqlite3_stmt *checkpt_prune_stmt;
     flux_t *h;
     char *hashfun;
     int hash_size;
@@ -77,6 +93,7 @@ struct content_sqlite {
     struct content_stats stats;
     char *journal_mode;
     char *synchronous;
+    int max_checkpoints;
     bool truncate;
 };
 
@@ -376,15 +393,6 @@ void checkpoint_get_cb (flux_t *h,
     const char *errstr = NULL;
     json_error_t error;
 
-    if (sqlite3_bind_text (ctx->checkpt_get_stmt,
-                           1,
-                           (char *)KVS_DEFAULT_CHECKPOINT,
-                           strlen (KVS_DEFAULT_CHECKPOINT),
-                           SQLITE_STATIC) != SQLITE_OK) {
-        log_sqlite_error (ctx, "checkpt_get: binding key");
-        set_errno_from_sqlite_error (ctx);
-        goto error;
-    }
     if (sqlite3_step (ctx->checkpt_get_stmt) != SQLITE_ROW) {
         errno = ENOENT;
         goto error;
@@ -433,15 +441,6 @@ void checkpoint_put_cb (flux_t *h,
     }
     if (sqlite3_bind_text (ctx->checkpt_put_stmt,
                            1,
-                           (char *)KVS_DEFAULT_CHECKPOINT,
-                           strlen (KVS_DEFAULT_CHECKPOINT),
-                           SQLITE_STATIC) != SQLITE_OK) {
-        log_sqlite_error (ctx, "checkpt_put: binding key");
-        set_errno_from_sqlite_error (ctx);
-        goto error;
-    }
-    if (sqlite3_bind_text (ctx->checkpt_put_stmt,
-                           2,
                            value,
                            strlen (value),
                            SQLITE_STATIC) != SQLITE_OK) {
@@ -455,15 +454,28 @@ void checkpoint_put_cb (flux_t *h,
         set_errno_from_sqlite_error (ctx);
         goto error;
     }
+    if (sqlite3_bind_int (ctx->checkpt_prune_stmt,
+                          1,
+                          ctx->max_checkpoints) != SQLITE_OK) {
+        log_sqlite_error (ctx, "checkpt_prune: binding count");
+        goto error;
+    }
+    if (sqlite3_step (ctx->checkpt_prune_stmt) != SQLITE_DONE) {
+        log_sqlite_error (ctx, "checkpt_prune: executing stmt");
+        set_errno_from_sqlite_error (ctx);
+        goto error;
+    }
     if (flux_respond (h, msg, NULL) < 0)
         flux_log_error (h, "flux_respond");
     (void )sqlite3_reset (ctx->checkpt_put_stmt);
+    (void )sqlite3_reset (ctx->checkpt_prune_stmt);
     free (value);
     return;
 error:
     if (flux_respond_error (h, msg, errno, errstr) < 0)
         flux_log_error (h, "flux_respond_error");
     (void )sqlite3_reset (ctx->checkpt_put_stmt);
+    (void )sqlite3_reset (ctx->checkpt_prune_stmt);
     free (value);
 }
 
@@ -486,6 +498,10 @@ static void content_sqlite_closedb (struct content_sqlite *ctx)
         if (ctx->checkpt_put_stmt) {
             if (sqlite3_finalize (ctx->checkpt_put_stmt) != SQLITE_OK)
                 log_sqlite_error (ctx, "sqlite_finalize checkpt_put_stmt");
+        }
+        if (ctx->checkpt_prune_stmt) {
+            if (sqlite3_finalize (ctx->checkpt_prune_stmt) != SQLITE_OK)
+                log_sqlite_error (ctx, "sqlite_finalize checkpt_prune_stmt");
         }
         if (ctx->db) {
             if (sqlite3_close (ctx->db) != SQLITE_OK)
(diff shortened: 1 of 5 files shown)


garlick:
I started to do a review but wanted to back up a bit and talk through the schema changes here.

First, there is a bit of a problem if a system instance running the new release containing these changes starts up from an existing sqlite file.  That doesn't *normally* happen because we do a dump/restore across restarts, but it sometimes happens if there is a problem during shutdown and the dump isn't written.    This code currently appears to do nothing to alter the `checkpt` table definition if the table already exists, which is not going to work.  Maybe the new table should be created as `checkpt_v2` and some sort of migration policy be implemented to allow upgrades and downgrades?

Second, it seems like you've identified two problems with using the timestamp as primary key.  One is your concern that prompted mentioning vendoring sqlite in the PR description.  Two is that second resolution  timestamps could hypothetically be non-unique, although unlikely.   Your SQL cleverly uses the timestamp in combination with the key (the KVS namespace name) to enforce a maximum number of rows per namespace.

Should we back up and discuss whether we are ever going to store checkpoints in the database for anything other than the primary namespace?   It would simplify things if we could just use `INTEGER PRIMARY KEY` in this table and assume all rows are for the primary namespace.   We will soon need to preserve running jobs across a restart, but we might be able do this by simply having the exec system link the job namespaces into the primary at shutdown and restore them at startup.

chu11:
> Maybe the new table should be created as checkpt_v2 and some sort of migration policy be implemented to allow upgrades and downgrades?

Ahhh, good catch.  I had completely missed that in my hasty prototyping.

> Should we back up and discuss whether we are ever going to store checkpoints in the database for anything other than the primary namespace? It would simplify things if we could just use INTEGER PRIMARY KEY in this table and assume all rows are for the primary namespace. We will soon need to preserve running jobs across a restart, but we might be able do this by simply having the exec system link the job namespaces into the primary at shutdown and restore them at startup.

That's a really good point.  We've been checkpointing with "keys" for so long (possibly predates my involvement in flux-core).  Do you happen to recall the reasoning for it so long ago?  Possibly just for testing?

If we go w/ a "checkpoint 2.0", perhaps we certainly could "refactor" the key away.  That would involve content-files, content-s3, mods and probably lots of test alteration.

As for the primary key, I did choose "timestamp" as the primary key so we could be assured we'd have a timestamp for the checkpoint regardless of what the user checkpointed (see #6798), as they can checkpoint virtually any data they want.  I suppose the atlernate viewpoint would be to trust the user will checkpoint a timestamp if they feel it's important?

garlick:
> We've been checkpointing with "keys" for so long (possibly predates my involvement in flux-core). Do you happen to recall the reasoning for it so long ago? Possibly just for testing?

The checkpoint table was added by 4a68e3d905a7d3edbebf89f6afc68af4bc1c09e5, not that long ago.  I think at the time I was thinking the key was going to be the namespace name, and that there would be only one row per namespace.  My proposal above is to discuss whether we ever intend to put namespaces other than the primary in there, when we could just link them into the primary namespace at shutdown (or even periodically for long running jobs, though that would require a bit more design).  If we aren't going to store non-primary namespace roots then we can ignore (not store) the key.

> If we go w/ a "checkpoint 2.0", perhaps we certainly could "refactor" the key away.   That would involve content-files, content-s3, mods and probably lots of test alteration.

I think if we are careful we can probably avoid too much disruption up front.  content-sqlite could just fail if the key in `checkpoint-get` and `checkpoint-put` isn't "primary" and the other back ends could be left alone.

> As for the primary key, I did choose "timestamp" as the primary key so we could be assured we'd have a timestamp for the checkpoint regardless of what the user checkpointed (see https://github.com/flux-framework/flux-core/pull/6798), as they can checkpoint virtually any data they want. I suppose the atlernate viewpoint would be to trust the user will checkpoint a timestamp if they feel it's important?

Regarding checkpoint content, see 9958fbb976a4d263085cd68232a825aa3737076e.  It really has one purpose, to support saving/restoring the KVS root.  Storing random stuff in there may be possible, but is not OK and should not be supported IMHO.  Probably at this point we should lock down the checkpoint object to version 1, don't you think?

chu11:
> It really has one purpose, to support saving/restoring the KVS root. Storing random stuff in there may be possible, but is not OK and should not be supported IMHO. Probably at this point we should lock down the checkpoint object to version 1, don't you think?

I agree.  My point was more that in tests we checkpoint "random data", just for the purposes of testing.  So we don't technically have timestamp information all of the time.

Lemme prototype given the thoughts / discussion above and see how it turns out.

chu11:
just re-pushed with an updated prototype

- remove use of the "key" for checkpoints (built on top of #6863).  The primary checkpt table just uses an integer that increases with each "put".
- support migration of v1 to v2 of the table

chu11:
re-pushed, rebasing this on top of #6863 and #6869

src/modules/content-sqlite/content-sqlite.c
```diff
@@ -47,14 +47,29 @@ const char *sql_store = "INSERT INTO objects (hash,size,object) "
                         "  values (?1, ?2, ?3)";
```
garlick, line 47: commit message: drop 'per key' since that's not a thing anymore.

src/modules/content-sqlite/content-sqlite.c
```diff
@@ -712,6 +736,123 @@ static int content_sqlite_opendb (struct content_sqlite *ctx, bool truncate)
...
+        set_errno_from_sqlite_error (ctx);
+        goto error;
+    }
+
+    /* no checkpoint stored, just drop the table */
+    if (sqlite3_step (checkpt_get_stmt) != SQLITE_ROW)
+        goto drop;
+
+    s = (char *)sqlite3_column_text (checkpt_get_stmt, 0);
+
```
garlick: Is casting away the const from `sqlite3_column_text()` the right thing?  AFAICS `sqlite3_bind_text()` 3rd argument is a const so no?

This cast is repeated a few times so if this should change, they most likely all should.

src/modules/content-sqlite/content-sqlite.c
```diff
@@ -712,6 +736,123 @@ static int content_sqlite_opendb (struct content_sqlite *ctx, bool truncate)
...
+error:
+    if (checkpt_get_stmt) {
+        if (sqlite3_finalize (checkpt_get_stmt) != SQLITE_OK)
+            log_sqlite_error (ctx, "sqlite_finalize checkpt_get_stmt");
+    }
+    json_decref (o);
+    return rv;
+}
+
+static int content_sqlite_checkpt_setup (struct content_sqlite *ctx)
```
garlick: Suggestion for readability: change this function to something like
```c
int table_exists (sqlite3 *db, const char *table_name, bool *result);
```
and then in `mod_main()` do
```c
bool has_old_table;
if (table_exists (ctx->db, "checkpt", &has_old_table) < 0
    || (has_old_table && content_sqlite_checkpt_migrate (ctx) < 0)) {
        goto cleanup
}
```
It'll just make the flow a little easier to follow since it's not obvious from the name "checkpt_setup()` what is going to happen there.

src/modules/content-sqlite/content-sqlite.c
```diff
@@ -69,6 +69,8 @@ const char *sql_checkpt_prune =
```
garlick, line 69: commit message: drop "per key"

src/modules/content-sqlite/content-sqlite.c
```diff
@@ -479,6 +482,46 @@ void checkpoint_put_cb (flux_t *h,
...
+            goto error;
+        }
+        if (!(o = json_pack ("{s:i s:s}",
+                             "id", id,
+                             "value", value))
+            || json_array_append_new (checkpts, o) < 0) {
+            json_decref (o);
+            errno = ENOMEM;
+            goto error;
+        }
```
garlick: Value object should probably be decoded?

Json objects encoded as json strings is a flux anti-pattern by now.

src/modules/content-sqlite/content-sqlite.c
```diff
@@ -69,6 +69,8 @@ const char *sql_checkpt_prune =
 
 const char *sql_table_list = "SELECT tbl_name FROM sqlite_master where type = 'table'";
 
+const char *sql_checkpt_get_all = "SELECT * FROM checkpt_v2 ORDER BY id DESC";
+
 #define MAX_CHECKPOINTS_DEFAULT 100
```
garlick: Suggestion: set the default to 5 and we can crank it up in the system instance config if we ever need to.

My gut feeling is that we'll be rolling back at most one step in real life.  We need to make sure that we stop writing to the database after a disk full event or I/O error.  If we do that, given that blob commits are currently synchronous, in theory we'll never need any checkpoint history.  So this is truly a "what if we don't understand how databases work" or "what if there is a bug in sqlite" CYA move.  Once you've iterated 5 times on the the impossible situation, it's probably time to call it a day and restore from the nightly dump file.  Nobody is going to waste time checking the 99th most recent checkpoint.

Plus it makes the stats output annoyingly huge

garlick (commented):
Here are a few comments inline.

I'll do a little playing around with it and see if I have further comments.

Quick note: it might tidy things up a bit if checkpoints are moved to a separate source file.

garlick:
I was just wondering if we needed to keep the old `checkpt` table around in case we need to revert to a previous version.  However, in the unlikely event that we need to do that, we could always just do a dump and restore from that.

src/modules/content-sqlite/content-sqlite.c
chu11: hmmm I suspect I casted it b/c documentaiton says it returns an unsigned char.

`const unsigned char *sqlite3_column_text(sqlite3_stmt*, int iCol);`

minimally I could cast to `const char *` instead.  Lemme look into that again.

Edit: re-reading your comment, i missed that you specifically mention the "const", I had originally read it as "why cast at all", whoops.  Anyways, agree, we shouldn't cast away the const.

src/modules/content-sqlite/content-sqlite.c
chu11: > Suggestion: set the default to 5 and we can crank it up in the system instance config if we ever need to.

Fair enough.  Just an FYI, I chose 100 b/c we're using 30 mins for the kvs checkpoint period in production (the default is no checkpoint period, so it is admin controlled).  So it was less about 100 entries, but more about preserving ~48 hours of checkpoints.  The 48 hours was admittedly randomly chosen.  Perhaps something in the, oh i dunno, 10-24 hour period would be more reasonable?  So like 20-50 entries?  The concern is something went south and we gotta back some number of hours.

Edit: ehhh i guess i am leaning towards your view now.  Odds are we only need to go back 1-2 checkpoints max.

chu11:
> I was just wondering if we needed to keep the old checkpt table around in case we need to revert to a previous version. 

Oh good point.  Hmmm.  I think I deleted it mostly b/c I did not want to trigger a "does it still exist" check the next time the database was loaded.  That could be easily rectified via some "does checkpt_v2 table exist" check instead.

> However, in the unlikely event that we need to do that, we could always just do a dump and restore from that.

Now that you mention it, I had forgotten that the table would be deleted under normal operating conditions anyways in rc1 (via `truncate` option).  So the delete is perhaps unnecessary, and we could leave the table in there as a "just in case".

Edit:

I change my mind

> I think I deleted it mostly b/c I did not want to trigger a "does it still exist" check the next time the database was loaded.  That could be easily rectified via some "does checkpt_v2 table exist" check instead.

deleting the table makes the initial "check" and "migrate" part of the code a lot simpler.  Without it it becomes hairier (i.e. "if checkpt_v2 and checkpt both exists, which has a newer checkpoint value).

src/modules/content-sqlite/content-sqlite.c
chu11: good point.  I think in the past we allowed checkpoints to be random raw text, but we now require the RFC format.

chu11:
re-pushed per comments above

garlick (approved):
This LGTM!

As discussed on the :coffee: call, there might be some opportunities for testing migration at a little higher level, but maybe redundant with what's already here.

chu11:
Thanks!  Per our :coffee: chat, not sure additional tests were needed.  Will set MWP.

PR #6775, opened 2025-04-22
kvs: support initial-rootref option

chu11:
Problem: The only way for the KVS to start with a non-default root reference is to read a root reference from the content module's checkpoint service.  There is no way to initialize the KVS with a specific root reference.  Allowing users to select an initial root reference could be useful for debugging and testing, especially when multiple checkpoints exist.
    
Support a initial-rootref option.


chu11:
thanks!  Made a couple tweaks per your suggestion, will set MWP.

PR #6783, opened 2025-04-25
libsubprocess: remove confusing `flux_future_push: Invalid argument` error message in bulk-exec

### LINKED ISSUE #6776: confusing error message: `flux_future_push: Invalid argument`
A user described getting the following two errors:

> `broker.err[0] on …  unresponsive for 31.994s`
> then
> `flux_future_push: Invalid argument`

It is not yet known what was happening at the time or if there was any more information, or even if `flux_future_push` error was related to the torpid node error message. Since handling of torpid nodes is well understood and expected, the main issue here is the `flux_future_push` error.

Looking at code, the only place I could find as the source of the `flux_future_push` error message is what actually looks like leftover debugging in `libsubprocess/bulk-exec.c`

https://github.com/flux-framework/flux-core/blob/b6f6b8767f0e97c645df21069f1fc773a507a439/src/common/libsubprocess/bulk-exec.c#L704-L712

(Side note: a lot of errors in `bulk-exec.c` use `fprintf` 🤦 😞)

We'll try to get some further information about what the user was doing at the time.

One thing I wonder is that if the torpid node causes an unexpected failure in `flux_subprocess_kill()` which then causes `bulk_exec_kill` to traverse some untested code path here, ultimately resulting in the failure.

grondo:
This PR removes some stray debug `fprintf` calls in bulk-exec, and also prevents calling `flux_future_push()` with a `NULL` future in `bulk_exec_kill()`.

Fixes #6776

### DIFF
--- a/src/common/libsubprocess/bulk-exec.c
+++ b/src/common/libsubprocess/bulk-exec.c
@@ -696,12 +696,14 @@ flux_future_t *bulk_exec_kill (struct bulk_exec *exec,
                         flux_future_fulfill_error (cf, err, "Internal error");
                 }
             }
-            (void) snprintf (s,
-                             sizeof (s)-1,
-                             "%u",
-                             flux_subprocess_rank (p));
-            if (flux_future_push (cf, s, f) < 0)
-                flux_future_destroy (f);
+            if (f) {
+                (void) snprintf (s,
+                                 sizeof (s)-1,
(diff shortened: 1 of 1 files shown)


chu11 (approved):
LGTM!

grondo:
Thanks! I've set MWP.

PR #6787, opened 2025-04-29
Support new flux-fsck tool to check integrity of content store

chu11:
There is no way to check if the data in the content store is valid and doesn't have corruption.  Support a `flux-fsck` tool to check it.

This tool is largely based on `flux-dump`, walking the content store tree and making sure all references are valid and retrievable.

This is very basic start to the `flux-fsck` and it doesn't do much beyond the trivial check.  Future work will include.

A) check different checkpoints (i.e. related to work in #6772)

B) update primary checkpoint if we are "rolling back" to an earlier one

Edit: it's worth noting that this has no dependencies on other recent PRs (#6772, #6775).

### DIFF
--- a/doc/man1/flux-fsck.rst
+++ b/doc/man1/flux-fsck.rst
@@ -1,0 +1,65 @@
+============
+flux-fsck(1)
+============
+
+
+SYNOPSIS
+========
+
+**flux** **fsck** [*OPTIONS*]
+
+
+DESCRIPTION
+===========
+
+.. program flux fsck
+
+The :program:`flux fsck` checks the integrity of the KVS backing
+store, starting with the most recent checkpoint (root version) written
+to the backing store.
+
+
+OPTIONS
+=======
+
+.. option:: -h, --help
+
+   Summarize available options.
+
+.. option:: -v, --verbose
+
+   List keys on stderr as they are validated.
+
+.. option:: -q, --quiet
+
+   Don't output diagnostic messages and discovered errors.
+
+
+EXIT STATUS
+===========
+
+0
+  Content store valid
+
+1
+  One or more errors were discovered
+
+
+RESOURCES
+=========
+
+.. include:: common/resources.rst
+
+
+FLUX RFC
+========
+
+:doc:`rfc:spec_10`
+
+:doc:`rfc:spec_11`
+
+
+SEE ALSO
+========
+
+:man1:`flux-content`, :man1:`flux-kvs`
--- a/doc/manpages.py
+++ b/doc/manpages.py
@@ -30,6 +30,7 @@ man_pages = [
     ('man1/flux-cron', 'flux-cron', 'Cron-like utility for Flux', [author], 1),
     ('man1/flux-dmesg', 'flux-dmesg', 'access broker ring buffer', [author], 1),
     ('man1/flux-dump', 'flux-dump', 'Write KVS snapshot to portable archive', [author], 1),
+    ('man1/flux-fsck', 'flux-fsck', 'Check integrity of KVS backing store', [author], 1),
     ('man1/flux-env', 'flux-env', 'Print the flux environment or execute a command inside it', [author], 1),
     ('man1/flux-event', 'flux-event', 'Send and receive Flux events', [author], 1),
     ('man1/flux-exec', 'flux-exec', 'Execute processes across flux ranks', [author], 1),
--- a/src/cmd/builtin/fsck.c
+++ b/src/cmd/builtin/fsck.c
@@ -1,0 +1,305 @@
+/************************************************************\
+ * Copyright 2025 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+#if HAVE_CONFIG_H
+# include <config.h>
+#endif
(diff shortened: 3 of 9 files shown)


src/cmd/builtin/fsck.c
```diff
@@ -0,0 +1,266 @@
...
+                         const char *path,
+                         json_t *treeobj)
+{
+    flux_future_t *f;
+    const void *buf;
+    size_t buflen;
+    json_t *treeobj_deref = NULL;
+
+    if (treeobj_get_count (treeobj) != 1)
+        log_msg_exit ("%s: blobref count is not 1", path);
```
garlick, line 125: Should this be fatal?  Maybe it should be a `read_error()` and say something like
```
%s: dirref object has blob count of %d, when 1 is expected
```
(and return rather than trying to recurse)

src/cmd/builtin/fsck.c
```diff
@@ -0,0 +1,266 @@
...
+    json_decref (treeobj_deref);
+    flux_future_destroy (f);
+}
+
+static void fsck_treeobj (flux_t *h,
+                          const char *path,
+                          json_t *treeobj)
+{
+    if (treeobj_validate (treeobj) < 0)
+        log_msg_exit ("%s: invalid tree object", path);
```
garlick, line 154: `read_error()` here too perhaps?

src/cmd/builtin/fsck.c
```diff
@@ -0,0 +1,266 @@
...
+
+    h = builtin_get_flux_handle (p);
+
+    if (!(f = kvs_checkpoint_lookup (h,
+                                     NULL,
+                                     KVS_CHECKPOINT_FLAG_CACHE_BYPASS))
+        || kvs_checkpoint_lookup_get_rootref (f, &blobref) < 0)
+        log_msg_exit ("error fetching checkpoint: %s",
+                      future_strerror (f, errno));
+
```
garlick, line 236: It would be useful here to use `kvs_checkpoint_lookup_get_timetstamp()` and print something like
```
Checking integrity of stable checkpoint from <date>
```
Possibly a `--quiet` option would be useful to suppress this message.

src/cmd/builtin/fsck.c
```diff
@@ -0,0 +1,266 @@
...
+    fsck_blobref (h, blobref);
+
+    flux_future_destroy (f);
+
+    flux_close (h);
+
+    return (errorcount ? -1 : 0);
```
garlick, line 243: Similarly, printing the number of errors here might be useful (with `--quiet` suppression).

src/cmd/builtin/fsck.c
```diff
@@ -0,0 +1,266 @@
...
+                        path,
+                        future_strerror (f, errno));
+        errorcount++;
+        flux_future_destroy (f);
+        return;
+    }
+    if (!(treeobj_deref = treeobj_decodeb (buf, buflen)))
+        log_err_exit ("%s: could not decode directory", path);
+    if (!treeobj_is_dir (treeobj_deref))
+        log_msg_exit ("%s: dirref references non-directory", path);
```
garlick, line 143: `read_error()` here?

t/t2816-fsck-cmd.t
```diff
@@ -0,0 +1,91 @@
...
+	flux kvs put --append dir.c=testC &&
+	flux kvs put --append dir.c=testD &&
+	flux kvs link dir alink &&
+	flux kvs namespace create testns &&
+	flux kvs put --namespace=testns dir.a=testns
+'
+# N.B. startlog commands in rc scripts normally ensures a checkpoint
+# exists but we do this just to be extra sure
+test_expect_success 'call --sync to ensure we have checkpointed' '
+	flux kvs put --sync dir.sync=foo
```
garlick, line 33: Hmm, we should have a `flux kvs sync` command or something.

garlick (commented):
I added a few suggestions inline but I think a great start. i'll approve and you can fix what you think makes sense.

src/cmd/builtin/fsck.c
chu11: I did think about this as I wrote the tool.  Since we wouldn't have the ability to recurse, my mind meandered towards "I guess we stop here."  Leading me to leaving the majority of "treeobj errors" as fatal.  

But lemme go through and consider it again.

garlick: Design wise, I think this tool should try to report (and eventually unlink/fix)) all the damaged keys, however unlikely the nature of the damage is.

chu11:
re-pushed, fixing up things based on a lot of comments above

doc/man1/flux-fsck.rst
```diff
@@ -0,0 +1,65 @@
...
+
+
+EXIT STATUS
+===========
+
+0
+  Content store valid
+
+1
+  One or errors were discovered
```
garlick, line 45: One or more errors?

doc/manpages.py
```diff
@@ -30,6 +30,7 @@
     ('man1/flux-cron', 'flux-cron', 'Cron-like utility for Flux', [author], 1),
     ('man1/flux-dmesg', 'flux-dmesg', 'access broker ring buffer', [author], 1),
     ('man1/flux-dump', 'flux-dump', 'Write KVS snapshot to portable archive', [author], 1),
+    ('man1/flux-fsck', 'flux-fsck', 'Check integrity of content store data', [author], 1),
```
garlick, line 33: Check integrity of KVS backing store

garlick (approved):
Couple of man page suggestions

chu11:
thanks!  Fixed up those manpage nits, will set MWP

PR #6788, opened 2025-04-30
job-exec: make job signaling more configurable and fix documentation

### LINKED ISSUE #6784: job-exec: signal behavior on timeout and cancel doesn't match docs
Things are still a bit wonky in the job-exec with regard to how and when signals are sent to jobs after a fatal job exception (including timeout).

The documentation states:

```
       When a job is canceled or gets a fatal exception it is terminated using
       the following sequence

          • The  job shells are notified to send term-signal to job tasks, un‐
            less the job is being terminated due to a  time  limit,  in  which
            case SIGALRM is sent instead.

          • After kill-timeout, any remaining shells are sent kill-signal

          • This  continues with an exponential backoff, with the timeout dou‐
            bling after each attempt (capped at 300s)

          • After a total of max-kill-count attempts, any nodes still  running
            processes are drained with the message: "unkillable user processes
            for job JOBID"
```

What actually happens in the job-exec module appears to be the following:

On a fatal exception:
 - term-signal is sent to the job shell. Job shell captures this signal and forwards to tasks.
 - `kill_timer` _and_ `kill_shell_timer` timer watchers are started.
    - `kill_timer` _repeats_ at `kill_timeout` seconds and sends `kill-signal` via `flux_job_kill()`
    - `kill_shell_timer` starts at `5*kill_timer` seconds, and sends `kill-signal` to the job shell (or `SIGUSR1` if the target is the IMP).
    - `kill_shell_timer` repeats in an exponential backoff as described

On timeout, the same as above occurs, except `SIGARLM` is sent as the initial signal as described in the docs.

The difference between docs and reality includes:
- first `term-signal` is sent to the job shell not using `flux_job_kill()` (which goes through the job manager and uses an event to tell the job shells to forward the signal)
- first attempt to directly kill the job shells is always 5*`kill-timeout` and is not configurable
- job shells are notified to send `kill-signal` (not `term-signal`) to tasks every `kill-timeout` seconds between when the first `term-signal`/`SIGALRM` is sent and the first attempt to kill the job shell.
- Also: max-kill-count only applies to attempts to kill the job shell, i.e. the node is drained when `kill_shell_count` > `max_kill_count`

It seems like this should be simplified a bit. First, there's been a request to be able to send `term-signal` multiple times before going to `kill-signal`. Second, the job exec system should switch to sending `kill-signal` to the job shells after a configurable number of attempts to send `term-signal`. Finally, it might be slightly preferable to signal the job shell directly with `term-signal` rather than going through `flux_job_kill(3)` (actually the first attempt does this, so it would be more consistent too.)

grondo:
This PR adds a couple tunables to the job-exec module for how it signals jobs and then fixes the documentation to actually describe how jobs are terminated. I'm not super happy with the result, but it is an improvement over what's there now because it fixes a couple glaring issues:

 - the time between when the execution system starts signaling a job and when it transitions to killing the job shell is not configurable
 - the execution system sent `SIGTERM` to a job at most once (none in the case of a timeout exception).

This PR adds two new tunables `kill-shell-delay` (default: 5 * `kill-timeout`) and `max-term-count` (default: 2) to fix the above issues.

Additionally, the documentation in the `flux-config-exec(5)`  `JOB TERMINATION` section was incorrect. This is now fixed to correctly describe the new termination process given the changes above.

Fixes #6784

### DIFF
--- a/doc/man5/flux-config-exec.rst
+++ b/doc/man5/flux-config-exec.rst
@@ -54,15 +54,27 @@ sdexec-stop-timer-signal
    10 (SIGUSR1, the IMP proxy for SIGKILL) is used.
 
 kill-timeout
-   (optional) The amount of time in FSD to wait after ``SIGTERM`` is
-   sent to a job before sending ``SIGKILL``. The default is "5s". See
-   :ref:`job_termination` below for details.
+   (optional) The amount of time in FSD between signals sent to
+   job tasks after a fatal job exception. The execution system will
+   send ``max-term-count`` ``SIGTERM`` (possibly starting with
+   ``SIGALRM`` for a timeout exception), then will transition to sending
+   ``SIGKILL`` at an interval of ``kill-timeout``. This occurs until
+   ``kill-shell-delay`` has elapsed, after which the job shell will
+   be signaled. The default is 5s.  See :ref:`job_termination` below
+   for details.
+
+kill-shell-delay
+   (optional) The amount of time in FSD to wait before sending ``SIGKILL``
+   to the job shell instead of job tasks.
+
+max-term-count
+   (optional) The number of times ``SIGTERM`` will be sent to job tasks
+   before switching to ``SIGKILL``. The default is 2.
 
 max-kill-count
-   (optional) The maximum number of times a job will be sent ``kill-signal``
+   (optional) The maximum number of times to attempt to kill the job shell
    before the execution system will consider the job unkillable and drains
    the node. The default is 8. See :ref:`job_termination` below for details.
-   for details.
 
 term-signal
    (optional) A string specifying an alternate signal to ``SIGTERM`` when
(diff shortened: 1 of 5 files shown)


doc/man5/flux-config-exec.rst
```diff
@@ -128,15 +140,24 @@ JOB TERMINATION
...
 
- - The job shells are notified to send ``term-signal`` to job tasks, unless
-   the job is being terminated due to a time limit, in which case ``SIGALRM``
-   is sent instead.
- - After ``kill-timeout``, any remaining shells are sent ``kill-signal``
+ - The job shells are sent ``term-signal``, unless the job is being
+   terminated due to a time limit, in which case ``SIGALRM`` is sent instead.
+   The job shells forward this signal to tasks.
+ - This is repeated with ``term-signal`` at an interval of
+   ``kill-timeout`` until ``kill-shell-delay`` has elapsed.
```
chu11, line 147: was there consideration renaming `kill-timeout` to `term-timeout`?  It just seems a tad confusing.  Just beginning to skim the code for review, I got confused quickly.

If it's a backwards compatibility thing, could rename this and variables, but have a backwards compatible check for it?

grondo: I considered that, but it is a `kill(2)` timeout, and also the job execution system sends `SIGTERM` only `max-term-count` times, then switches to `SIGKILL`, so `term-timeout` would be a misnomer as well. I agree this is pretty confusing.

grondo: If this PR is a bit too much. We could consider not including the `kill-shell-delay` and just fix the documentation to report that it is always `5*kill_timeout`. We do need to keep `max-term-count` because it was requested, but adding `kill-shell-delay` may not be required (not sure if anyone would adjust it)

PR #6791, opened 2025-05-02
job-exec: only adjust timelimit for jobs when start delay exceeds a configurable percent of job duration

### LINKED ISSUE #6781: job expiration discrepancy in subinstances
As noted in #6779, a there can be a discrepancy between the resource set expiration time and the actual time that a timeout exception is raised. This occurs because currently the job execution system gives the job some grace time, i.e. it adds the difference between the alloc and start events to the job time limit.

This causes very strange behavior for users when there is a non-trivial amount of grace time added to the job expiration.

For example, if the prolog takes 3 minutes, then there are 3 minutes within the batch instance where jobs have all timed out and no new jobs can be submitted because the current time > expiration time. In one case where a user was resubmitting jobs in a loop, this caused their log file to fill with errors.

I'm actually not confident of the correct behavior here. Some ideas:

 - If the job execution system detects a significant update in the job expiration time, it should update the job's duration accordingly. This will update R so that the expiration matches the real expiration.
 - Don't give any grace time in the job exec module. Document that prolog and job startup time are included in the job duration.
 - Add an rpc to the job-exec module to get the real expiration, use that from schedulers to set the resource set lifetime.

grondo:
Problem: When the job execution system receives a start request for a job with a time limit, it always adjusts the actual job time limit by the difference between the starttime in R and the time at which the start request is received. For jobs that are instances of Flux, this causes a discrepancy between the time at which the execution system will raise a timeout exception and the expiration used by the scheduler in the subinstance. This results in an arbitrary amount of time between when jobs in the instance can no longer run because the resources are expired, and the subinstance itself is finally terminated by the timeout exception.

It is probably unnecessary to provide the grace time in the execution system for the difference between the `alloc` and `start` events for jobs of any meaningful length. It is more important that the execution system timeout matches the expiration of resources in the subinstance. Therefore, only make this adjustment if the delta that would be applied is greater than 25% of the total job duration. This keeps the behavior for short jobs, where the difference would meaningfully effect the actual job runtime, but ensures longer jobs are not affected by the timeout discrepancy.

Fixes #6781

### DIFF
--- a/doc/man5/flux-config-exec.rst
+++ b/doc/man5/flux-config-exec.rst
@@ -82,6 +82,17 @@ barrier-timeout
    nodes on which the barrier is waiting.  To disable the barrier timeout,
    set this value to ``"0"``. (Default: ``30m``).
 
+max-start-delay-percent
+   (optional) Specify the maximum allowed delay, as a percentage of a job's
+   duration, between when a job is allocated (i.e. the starttime recorded
+   in _R_) and when the execution system receives the start request from
+   the job manager. If the delay exceeds this percentage, then extend the
+   job's effective expiration by the delay. This prevents short duration
+   jobs from having their runtime significantly reduced, while avoiding a
+   differential between the actual resource set expiration and the time
+   at which a ``timeout`` exception is raised for longer running jobs,
+   where any runtime impact will be negligible. The default is 25 percent.
+
 testexec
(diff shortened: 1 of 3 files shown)


grondo:
Thanks! Fixed that typo and will set MWP.

Issue #6789, opened 2025-05-01
jobs run with stopped queue during instance restart


grondo:
While running some tests, I noticed that a job in SCHED state can be allocated resources and run during restart, even if the queue was stopped before the original shutdown. Reproducer:

```console
$ flux start -Sstatedir=./run-when-stopped sh -c 'flux queue stop; flux submit sleep 0'
Scheduling is stopped
f22tbzGw
$ flux start -Sstatedir=./run-when-stopped sh -c 'flux queue list; flux jobs -a'
 EN ST TDEFAULT   TLIMIT     NNODES     NCORES      NGPUS
  ✔  ✗      inf      inf      0-inf      0-inf      0-inf
       JOBID USER     NAME       ST NTASKS NNODES     TIME INFO
    f22tbzGw grondo   sleep      CD      1      1   0.058s corona82
```

grondo:
Oh, I guess this is perhaps by design, though I can't think of why at the moment:
https://github.com/flux-framework/flux-core/blob/7e7f0fafac08e5099559d2a605049e9fed9272ee/src/modules/job-manager/queue.c#L121-L126

grondo:
Also, the initial sticky stopped state of a queue is not preserved across a restart, which is further confusing me
(but I guess also purposeful given the code referenced above)

```console
$ flux alloc -N1 --dump=namedqueue.tgz sh -c 'echo [queues.batch] | flux config load; flux queue list'
QUEUE    EN ST TDEFAULT   TLIMIT     NNODES     NCORES      NGPUS
batch     ✔  ✗      inf      inf        0-1       0-48        0-0
[detached: session exiting]
$ flux start -Scontent.restore=namedqueue.tgz flux queue list
 EN ST TDEFAULT   TLIMIT     NNODES     NCORES      NGPUS
  ✔  ✔      inf      inf        0-1       0-48        0-0
```

wihobbs:
> Oh, I guess this is perhaps by design, though I can't think of why at the moment:
> 
> [flux-core/src/modules/job-manager/queue.c](https://github.com/flux-framework/flux-core/blob/7e7f0fafac08e5099559d2a605049e9fed9272ee/src/modules/job-manager/queue.c#L121-L126)
> 
> Lines 121 to 126 in [7e7f0fa](/flux-framework/flux-core/commit/7e7f0fafac08e5099559d2a605049e9fed9272ee)
>      /* The anonymous queue begins life started, while named queues do not. 
[...]

PR #6792, opened 2025-05-02
job-manager: avoid starting jobs in a stopped anonymous queue on restart

grondo:
This PR fixes issue #6789. The dequeue of jobs from a stopped queue on restart was missing, probably since all named queues are created in the stopped state. This isn't the case for the default anonymous queue, however, so this PR adds the dequeue to the restart path.

I guess an alternate solution here would be to recover the queues and their state before reloading jobs from the KVS. However, I wanted to change as little as possible in this PR and just fix the issue (which is probably not a critical issue anyway)

### DIFF
--- a/src/modules/job-manager/queue.c
+++ b/src/modules/job-manager/queue.c
@@ -84,6 +84,8 @@ struct queue_ctx {
     bool have_named_queues;
 };
 
+static void dequeue_jobs (struct queue_ctx *qctx, const char *name);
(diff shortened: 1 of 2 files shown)


wihobbs (approved):
Works for me!

grondo:
Thanks @wihobbs! I think this is an acceptable approach for now, so I'll set MWP.

PR #6798, opened 2025-05-06
WIP: flux-content: support new checkpoints list command

chu11:
Built on top of #6772.  So I'll list this as WIP for the time being.

Problem: There is currently no way to get multiple checkpoints from the content modules.

Support an optional "index" key when getting content checkpoints.  If the index is not available, return ENOENT to the caller.  Support a new "flux-content checkpoints" command.

### DIFF
--- a/t/t0012-content-sqlite.t
+++ b/t/t0012-content-sqlite.t
@@ -158,6 +158,13 @@ test_expect_success 'checkpoint-get foo returned correct timestamp' '
         grep 2.2 timestamp.out
 '
 
+test_expect_success 'flux content checkpoints lists correct checkpoint' '
+        flux content checkpoints foo > checkpoints1.out &&
+        count=$(cat checkpoints1.out | wc -l) &&
+        test $count -eq 1 &&
+        grep bar checkpoints1.out
+'
+
 test_expect_success 'checkpoint-put updates foo rootref to baz' '
 	checkpoint_put foo baz
 '
@@ -187,6 +194,13 @@ test_expect_success 'checkpoint-backing-get foo returns rootref baz' '
 	test_cmp rootref_backing.exp rootref_backing.out
 '
 
+test_expect_success 'flux content checkpoints lists correct checkpoint' '
+        flux content checkpoints foo > checkpoints2.out &&
+        count=$(cat checkpoints2.out | wc -l) &&
+        test $count -eq 2 &&
+        head -n 1 checkpoints2.out | grep baz
+'
+
 test_expect_success 'checkpoint-backing-put foo w/ rootref boof' '
 	checkpoint_backing_put foo boof
 '
@@ -195,6 +209,13 @@ test_expect_success 'checkpoint-get foo returned rootref boof' '
 	echo boof >rootref4.exp &&
 	checkpoint_get foo | jq -r .value | jq -r .rootref >rootref4.out &&
 	test_cmp rootref4.exp rootref4.out
+'
+
+test_expect_success 'flux content checkpoints lists correct checkpoint' '
+        flux content checkpoints foo > checkpoints2.out &&
+        count=$(cat checkpoints2.out | wc -l) &&
+        test $count -eq 3 &&
+        head -n 1 checkpoints2.out | grep boof
 '
 
 test_expect_success 'checkpoint-get noexist fails with No such...' '
--- a/src/cmd/builtin/content.c
+++ b/src/cmd/builtin/content.c
@@ -14,6 +14,7 @@
 #include "builtin.h"
 
 #include <unistd.h>
+#include <jansson.h>
 
 #include "src/common/libutil/blobref.h"
 #include "src/common/libutil/read_all.h"
@@ -168,6 +169,49 @@ static int internal_content_dropcache (optparse_t *p, int ac, char *av[])
     return (0);
 }
 
+static int internal_checkpoints (optparse_t *p, int ac, char *av[])
+{
+    flux_t *h;
+    int optindex = optparse_option_index (p);
+    char *key;
+    int index = 0;
+
+    if (optindex != (ac - 1)) {
+        optparse_print_usage (p);
+        exit (1);
+    }
+    key = av[optindex];
+
+    if (!(h = builtin_get_flux_handle (p)))
+        log_err_exit ("flux_open");
+
+    while (true) {
+        flux_future_t *f = NULL;
+        char *s;
+        json_t *o;
+        if (!(f = flux_rpc_pack (h,
+                                 "content.checkpoint-get",
+                                 0,
+                                 0,
+                                 "{s:s s:i}",
+                                 "key", key,
+                                 "index", index)))
+            log_err_exit ("content.checkpoint-get");
+        if (flux_rpc_get_unpack (f, "{s:o}", "value", &o) < 0) {
+            if (errno == ENOENT)
+                break;
+            log_err_exit ("content.checkpoint-get");
+        }
+        s = json_dumps (o, JSON_COMPACT);
+        printf ("%s\n", s);
+        free (s);
+        flux_future_destroy (f);
+        index++;
+    }
+    flux_close (h);
+    return (0);
+}
+
 int cmd_content (optparse_t *p, int ac, char *av[])
 {
     log_init ("flux-content");
@@ -217,6 +261,13 @@ static struct optparse_subcommand content_subcmds[] = {
       NULL,
       "Flush dirty entries from local content cache",
       internal_content_flush,
+      0,
+      NULL,
+    },
+    { "checkpoints",
+      "key",
+      "List checkpoint(s)",
+      internal_checkpoints,
       0,
       NULL,
     },
--- a/src/modules/content-files/content-files.c
+++ b/src/modules/content-files/content-files.c
@@ -208,14 +208,24 @@ void checkpoint_get_cb (flux_t *h,
 {
     struct content_files *ctx = arg;
     const char *key;
+    int offset = 0;
     void *data = NULL;
     size_t size;
     json_t *o = NULL;
     const char *errstr = NULL;
     json_error_t error;
 
-    if (flux_request_unpack (msg, NULL, "{s:s}", "key", &key) < 0)
+    if (flux_request_unpack (msg,
+                             NULL,
+                             "{s:s s?i}",
+                             "key", &key,
+                             "index", &offset) < 0)
         goto error;
+    /* content-files only supports a single checkpoint */
+    if (offset) {
+        errno = ENOENT;
+        goto error;
+    }
     if (filedb_get (ctx->dbpath, key, &data, &size, &errstr) < 0)
         goto error;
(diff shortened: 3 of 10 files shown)


chu11:
re-pushed, rebasing on top of #6772

chu11:
removed WIP, now that #6772 has been merged

garlick:
Would it be useful to add an option to `flux content checkpoints` for human readable output?  I'm picturing a human trying to decide how far back in time to roll back time  and not making much sense of floating point time values.  Perhaps they could be shown in a tabular form with sequence, date, and blobref columns, with the date converted to ISO?

Also is the `index` field in the RPC justified? Seems like we could just return all checkpoints and let the client filter.  Even if there are hundreds, it won't be that much data.

chu11:
> Would it be useful to add an option to flux content checkpoints for human readable output? I'm picturing a human trying to decide how far back in time to roll back time and not making much sense of floating point time values. Perhaps they could be shown in a tabular form with sequence, date, and blobref columns, with the date converted to ISO?

Good point.  I originally wrote this back when "any checkpoint format" was something we were still allowing / supporting, but we've moved on from that notion.

> Also is the index field in the RPC justified? Seems like we could just return all checkpoints and let the client filter. Even if there are hundreds, it won't be that much data.

Hmmm, I didn't think of it that way.   I think the `index` in the RPC was predominantly to maintain backwards compatibility with the current RPC.  If we want to return all checkpoints, we'd have to alter the protocol in some way.  Perhaps it returns an array of checkpoints.  Or perhaps we can stream all of them back?

chu11:
> Hmmm, I didn't think of it that way. I think the index in the RPC was predominantly to maintain backwards compatibility with the current RPC. If we want to return all checkpoints, we'd have to alter the protocol in some way. Perhaps it returns an array of checkpoints. Or perhaps we can stream all of them back?

I had forgotten we have a checkpoint version number, so I guess it wouldn't be horrible to up the version.  We could return an array of checkpoints vs.  a single one.

But the big negative is the libkvs `kvs_checkpoint` API.  We could keep it the way it is (i.e. deal w/ array index 0) but that doesn't feel right (hypothetically could deprecate it).  Will ponder a bit on this.

chu11:
Re-pushed, going w/ this default output.  It's admittedly hand created.  I ponder if remaking `flux content` in python would allow us to go with the output format support in all of the other tools.  But that is perhaps for another day.

```
>flux content checkpoints
Index      Sequence   Time                 Rootref
0          5          2025-07-10T16:39:39Z sha1-238260b968b162927aa7542491f68efdf217d86c
1          4          2025-07-10T16:39:38Z sha1-238260b968b162927aa7542491f68efdf217d86c
2          1          2025-07-10T16:39:33Z sha1-1bb67ca407fc7ca54fd1dbe5e14f40403125bb84
```

I ended up keeping the `index` in the RPC.  It just felt like too much churn to change the RPC and the APIs around it, for this admittedly small need.

garlick:
Nice improvement on the command output!

> I ended up keeping the index in the RPC. It just felt like too much churn to change the RPC and the APIs around it, for this admittedly small need.

I guess I wasn't thrilled with the SQL query change that I had to look up (because I am a complete SQL noob), given that the index probably isn't necessary given the expected small size of the query result.  Easier to just filter it on the client side IMHO.  We own the clients and servers as well as the convenience API (internal only), so I think we could change it all we want.   But if you feel strongly, this is OK I guess.  We can always fix it later.

chu11:
> I guess I wasn't thrilled with the SQL query change that I had to look up (because I am a complete SQL noob), given that the index probably isn't necessary given the expected small size of the query result. Easier to just filter it on the client side IMHO. We own the clients and servers as well as the convenience API (internal only), so I think we could change it all we want. But if you feel strongly, this is OK I guess. We can always fix it later.

Don't feel strongly.  I struggled the most with how to convert these API functions

```
int kvs_checkpoint_lookup_get_rootref (flux_future_t *f, const char **rootref);
int kvs_checkpoint_lookup_get_timestamp (flux_future_t *f, double *timestamp);
int kvs_checkpoint_lookup_get_sequence (flux_future_t *f, int *sequence);
```

i.e. they only return one checkpoint entry's info.

Hmmmm, pondering this for a second, perhaps we could adjust the API to be similar  to the `libeventlog` API's `eventlog_decode()` (get array of stuff) and `eventlog_entry_parse()` (parse an entry).  Underneath the covers it can still handle the different versions of the checkpoint.  Lemme try that.

garlick:
Sounds good.  Just a convenience library, so whatever is going to work out for the localized use cases is fine IMHO.

chu11:
re-pushed, going with a revamped `libkvs` api that looks like this now

```
/* return array of checkpoints */
int kvs_checkpoint_lookup_get_checkpoints (flux_future_t *f,
                                           const json_t **checkpoints);

int kvs_checkpoint_parse_rootref (json_t *checkpoint, const char **rootref);

/* sets timestamp to 0 if unavailable
 */
int kvs_checkpoint_parse_timestamp (json_t *checkpoint, double *timestamp);

/* sets sequence to 0 if unavailable
 */
int kvs_checkpoint_parse_sequence (json_t *checkpoint, int *sequence);
```

`content-fiels` and `content-sqlite` now return all checkpoints in a json array.

As expected it does lead to a healthy amount of "churn".  Less in the code than I expected, but a lot more in the tests than I expected (alot of tests call checkpoint-get and parsed the response via `jq`).

chu11:
hit this build failure, it's a very new test from #6911 so just documenting in case this is a racy pattern (could hypotheticlaly just be a timeout)

```  
expecting success: 
  	$waitfile -t 60 -p kvs.lookup trace3.out
  
  waitfile: trace3.out: 60.001s: Timeout after 60s
  -rw-r--r-- 1 runner runner 227 Jul 16 23:26 trace3.out
  [Jul16 23:26]  resource tx > attr.get [32]
  [  +4.301247]  resource tx > groups.get [25]
  [  +4.303290]  resource tx > groups.get [25]
  [  +4.303333]  resource tx > module.status [13]
  [  +4.303368]  resource tx > kvs.commit [239]
  not ok 28 - kvs.lookup request was captured

```

src/common/libkvs/kvs_checkpoint.c
```diff
@@ -69,21 +70,62 @@ flux_future_t *kvs_checkpoint_lookup (flux_t *h, int flags)
...
-    if (flux_rpc_get_unpack (f,
-                             "{s:{s:i s:s}}",
-                             "value",
-                               "version", &version,
-                               "rootref", &tmp_rootref) < 0)
+    if (json_unpack (checkpoint,
+                     "{s:i s:s}",
+                     "version", &version,
+                     "rootref", &tmp_rootref) < 0)
         return -1;
```
garlick: Should this set errno=EPROTO on error?

src/common/libkvs/kvs_checkpoint.c
```diff
@@ -95,20 +137,19 @@ int kvs_checkpoint_lookup_get_rootref (flux_future_t *f, const char **rootref)
...
-    if (flux_rpc_get_unpack (f,
-                             "{s:{s:i s?f}}",
-                             "value",
-                               "version", &version,
-                               "timestamp", &ts) < 0)
+    if (json_unpack (checkpoint,
+                     "{s:i s?f}",
+                     "version", &version,
+                     "timestamp", &ts) < 0)
         return -1;
```
garlick: set errno?

src/common/libkvs/kvs_checkpoint.c
```diff
@@ -118,20 +159,19 @@ int kvs_checkpoint_lookup_get_timestamp (flux_future_t *f, double *timestamp)
...
-    if (flux_rpc_get_unpack (f,
-                             "{s:{s:i s?i}}",
-                             "value",
-                               "version", &version,
-                               "sequence", &seq) < 0)
+    if (json_unpack (checkpoint,
+                     "{s:i s?i}",
+                     "version", &version,
+                     "sequence", &seq) < 0)
         return -1;
```
garlick: set errno

src/common/libkvs/kvs_checkpoint.h
```diff
@@ -36,15 +36,19 @@ flux_future_t *kvs_checkpoint_commit (flux_t *h,
 
 flux_future_t *kvs_checkpoint_lookup (flux_t *h, int flags);
 
-int kvs_checkpoint_lookup_get_rootref (flux_future_t *f, const char **rootref);
+/* return array of checkpoints */
+int kvs_checkpoint_lookup_get_checkpoints (flux_future_t *f,
+                                           const json_t **checkpoints);
```
garlick: Maybe this one could just be called `kvs_checkpoint_lookup_get()` since it returns the whole enchilada?

src/common/libkvs/kvs_checkpoint.c
```diff
@@ -89,27 +82,11 @@ int kvs_checkpoint_lookup_get_checkpoints (flux_future_t *f,
 
     if (flux_rpc_get_unpack (f,
                              "{s:o}",
-                             "value", &o) < 0)
-        goto error;
```
garlick: Should we maybe check that `o` is an array and that it contains at least one entry, then fail with EPROTO if so?

Some users access element 0 without checking if it is NULL.

t/t0012-content-sqlite.t
```diff
@@ -134,7 +134,7 @@ test_expect_success 'checkpoint-put w/ rootref bar' '
 
 test_expect_success 'checkpoint-get returned rootref bar' '
 	echo bar >rootref.exp &&
-	checkpoint_get | jq -r .value | jq -r .rootref >rootref.out &&
+	checkpoint_get | jq -r .value[0] | jq -r .rootref >rootref.out &&
```
garlick, line 152: Maybe for later cleanup, but these jq invocations could easily be combined, e.g.
```
checkpoint_get | jq -r .value[0].rootref
```
Two potential gotchas with the way these tests are structured:
-  I think without `-e`, jq returns success even if the requested key/array entry is not found
- A pipeline succeeds if the last command in in succeeds

Not saying these tests are broken but they might be a little fragile as is.

garlick (approved):
LGTM!  I had a couple of minor suggestions that you can ignore if you want.

garlick:
There is a fix for that test failure over in the pr for syslog support.

src/common/libkvs/kvs_checkpoint.c
chu11: Hmmm.  Checking to make sure `o` is non-NULL is wise and that's obviously an EPROTO.  But right now `content-sqlite` returns an empty array if there are no checkpoints, but `content-files` returns ENOENT.  

Perhaps it should be EPROTO for not-an-array, ENOENT for no-checkpoints available.

t/t0012-content-sqlite.t
chu11: it seems we do the double `jq` elsewhere as well.  I'll do a cleanup PR after this PR.

chu11:
fixed up those small things in the review above, will set MWP.  thanks!

Issue #6795, opened 2025-05-05
optionally stop all queues at system instance startup


grondo:
Problem: Often when a system instance of Flux is restarted the sysadmins don't necessarily want jobs being scheduled immediately, however, all queues that were started before the shutdown are started just after jobs are reloaded from the KVS on a normal restart.

One idea would be to add a global flag, set via configuration or broker attribute, that disables scheduling (i.e. alloc requests) to all queues, regardless of current state. Once that flag is cleared, then normal alloc requests to started queues would continue.

Another idea (proposed by @kkier)  would be to stop all started queues on restart with a specific reason, e.g. "Automatic stop due to system shutdown" or similar. Admins would then have to manually start each of the stopped queues with this specific reason. This approach would be much easier to implement, though slightly higher burden on the sysadmins. It would not require any new tooling though.

PR #6801, opened 2025-05-06
job-manager: add `stop-queues-on-restart` configuration parameter

grondo:
One more PR to perhaps sneak into this release.
This addresses #6795 in the simplest way possible by adding a config key that, when set, causes any started queues to be stopped with a specific reason message. Admins enabling this config parameter will then need to manually start queues after a Flux restart.

### DIFF
--- a/doc/man5/flux-config-job-manager.rst
+++ b/doc/man5/flux-config-job-manager.rst
@@ -27,6 +27,17 @@ inactive-age-limit
 inactive-num-limit
    (optional) Integer maximum number of inactive jobs retained in the KVS.
 
+stop-queues-on-restart
+   (optional) Boolean value indicating if the job manager should automatically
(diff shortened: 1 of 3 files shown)


doc/man5/flux-config-job-manager.rst
```diff
@@ -27,6 +27,17 @@ inactive-age-limit
 inactive-num-limit
    (optional) Integer maximum number of inactive jobs retained in the KVS.
 
+stop-queues-on-restart
```
wihobbs, line 30: commit nit: messaage -> message

wihobbs (approved):
Gave it a spin and works for me. Thanks @grondo!

doc/man5/flux-config-job-manager.rst
grondo: Thanks! Force-pushed a fix and will set MWP.

Issue #6794, opened 2025-05-05
Suggestion - add an indication in `flux jobs` when results are truncated


kkier:
Current behavior is that e.g. `flux jobs -A` will return the first(?) 1000 jobs and then exit, without a user necessary realizing that there was more output to be had, since the `-c/--count` command isn't necessarily advertised unless you read the help.  Adding something like `1000 results returned - 1312 results omitted` as a last line would at least provide hinting that there's more to be seen.

chu11:
I was pondering this.  Internally within the `job-list` module, we effectively go through internal lists of all jobs and apply our constraints / filtering rules to see if the job matches.  If it matches, we save that off and when we hit the `-c/--count` cap, we return and we're done.

So in order to get a output result like "1000 results returned - 1312 results omitted", we'd have to iterate through all the rest of the jobs and continue to test/filter.  This might not be horrific right now, but with inactive jobs in a DB like with #4336 in the future, that's bad.

So the question is, what could we provide?  Would something as simple as `Job list cap reached, see --count to increase cap` be ok?  The minor technical corner case is that we don't know if there are more jobs to be listed.  We only know the cap was reached.

kkier:
Good points.  Yes, I think checking for N+1 and then messaging the user about it is probably the way to go.  Having a simple `flux jobs` query take some number of minutes to return because it's looping through the set of all jobs since the dawn of creation might have bad effects on my blood pressure.

chu11:
> Also, it might be possible to check if there is at least one more matching job and set a truncated flag in the response?

The issue is we don't know if there's at least one more.  We may have to scan all the way to the beginning of time.

grondo:
Yeah, that should be a reasonable approach. 

An alternate approach would be if job-list gets to `count` entries before reaching the end of the list of all jobs, it could set the flag which indicates there _may_ be more entries and to adjust the warning accordingly. The vast majority of the time the warning will be meaningful.

In fact, with the above approach, the feature could live in `flux-jobs.py` alone and we don't need to touch job-list.

However, we should try looking for count+1 instead of count and see if that raises an issue.

chu11:
> If we look for 1001 jobs and only return 1000, that's not really taking any longer than if we were looking for 1000, right?

What I was getting at is that after we iterate through jobs and match 1000 entries, we have no way to know for sure if there is a 1001th match.  We could keep on iterating on all the remaining jobs until the very end.  So we may end up scanning till the beginning of time anyways.

[...]

PR #6802, opened 2025-05-07
flux-jobs: output message if results truncated

chu11:
Simple attempt at the issue raised in #6794, all done in `flux-jobs`.

Only interesting oddity is i cheated and used the `no-header` flag to also not output the warning message.  Results in zero need to change any tests in sharness.

### DIFF
--- a/doc/man1/flux-jobs.rst
+++ b/doc/man1/flux-jobs.rst
@@ -36,7 +36,8 @@ OPTIONS
 
 .. option:: -n, --no-header
 
-   For default output, do not output column headers.
+   For default output, do not output column headers and truncation
+   warnings.
 
 .. option:: -u, --user=[USERNAME|UID]
 
--- a/src/cmd/flux-jobs.py
+++ b/src/cmd/flux-jobs.py
@@ -106,7 +106,7 @@ def fetch_jobs_stdin():
             LOGGER.error("JSON input error: line %d: %s", fileinput.lineno(), str(err))
             sys.exit(1)
         jobs.append(job)
-    return jobs
+    return (jobs, False)
 
 
 def need_instance_info(fields):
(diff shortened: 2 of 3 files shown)


chu11:
re-pushed, I went with a mildly more complex / sophisticated approach.  Instead of requesting `N` entries from `job-list`, I request `N+1` entries from the `job-list` module.  The length of the returned result is used as "detection" of truncation, and the output warning is output if it is indeed truncated.

src/cmd/flux-jobs.py
```diff
@@ -550,6 +559,13 @@ def main():
         else:
             print(json.dumps({"jobs": result}))
 
+    # "no_header" also applies to trailers :-)
+    if not args.no_header and truncated:
+        print(f"warning: output truncated at {args.count} jobs.")
+        print(
+            "Use --count to increase returned results or use filters to alter results."
+        )
```
grondo, line 567: A couple thoughts here:
The output of `flux jobs` could be captured, so should the warning be to stderr?

I wonder if when using `-c N` the user expects truncation, so should the warning be suppressed if `args.count` was specified by the user? The warning might be especially annoying with `-c 1` for example.

Also, then if you use this strategy of asking for N+1, it will only be used for the default=1000 case.

chu11: > The output of flux jobs could be captured, so should the warning be to stderr?

Oh good point!

> I wonder if when using -c N the user expects truncation, so should the warning be suppressed if args.count was specified by the user? The warning might be especially annoying with -c 1 for example.

> Also, then if you use this strategy of asking for N+1, it will only be used for the default=1000 case.

Hmmmm.  That's an interesting idea.   My initial thought is to still do it when the user specifies --count.  I imagine the default scenario is they run `flux-jobs` and see the "output truncated" message.  They then increase the count to 2000 ... they need a warning if that increased count is also truncated?

grondo: Ah good point I hadn't thought of that. I'm going to be really annoyed when I get the error message with `-c 1` though :thinking: Perhaps this will just train users to use `-n, --no-header`?

chu11: yeah, that's why I tied the message to `--no-header`, b/c I figure anyone scripting with it is already using it?  Although I suppose some might be using `tail -n 1` with `-c 1`?

grondo: Yeah, the warning should still go to stderr, but you're right it needs to be there on any value for `--count` that isn't zero.

doc/man1/flux-jobs.rst
```diff
@@ -36,7 +36,8 @@ OPTIONS
 
 .. option:: -n, --no-header
 
-   For default output, do not output column headers.
+   For default output, do not output column headers and other optional
+   help output.
```
grondo, line 40: might as well call out "truncation warning" here?

chu11:
thanks!  Tweaked that minor documentation and will set MWP.

PR #6808, opened 2025-05-10
broker: add 'mincrit' topology

### LINKED ISSUE #6807: broker: available TBON topology options are not great for resilient batch jobs
Problem: as noted in #6806, performance issues have been observed with large batch jobs run with  `-Stbon.topo=kary:0` (flat topology).  This is currently recommended when resiliency is required, to minimize the number of critical router nodes.  Unfortunately, there is no convenient way just add a few router nodes to an otherwise flat topology to alleviate the bottleneck on rank 0.  You have to go all in with kary or binomial.

Note: a system instance can have a custom topology defined by `bootstrap.hosts` but that requires knowledge of the allocated hosts in advance, so it is not convenient for a batch job, and it will not work for multiple brokers per node, which is useful in test.

Maybe a new topology plugin could be added for this purpose, like I dunno, `3tree:N` for a 3 level tree with N routers?

garlick:
Problem: `kary:0` usefully minimizes critical nodes for batch jobs but may not perform well.
    
 Add a new TBON topology `mincrit[:K]` for defining a 3 level tree where the fanout from leader to routers is K, and the fanout from routers to leaves is dependent on the size.
    
If K is unspecified, a value is chosen that keeps the fanout to the leaves at or below 1024.  If there are 1025 nodes or less, a flat topology is used.
    
Example:
```
$ flux start -o,-Stbon.topo=mincrit:2 --test-size=11 flux overlay status
0 system76-pc: full
├─ 1 system76-pc: full
│  ├─ 3 system76-pc: full
│  ├─ 5 system76-pc: full
│  ├─ 7 system76-pc: full
│  └─ 9 system76-pc: full
└─ 2 system76-pc: full
   ├─ 4 system76-pc: full
   ├─ 6 system76-pc: full
   ├─ 8 system76-pc: full
   └─ 10 system76-pc: full
 ```   
Fixes #6807

### DIFF
--- a/doc/man7/flux-broker-attributes.rst
+++ b/doc/man7/flux-broker-attributes.rst
@@ -177,12 +177,30 @@ TREE BASED OVERLAY NETWORK
 ==========================
 
 tbon.topo [Updates: C]
-   URI describing the TBON tree topology such as ``kary:16``.  The ``kary``
-   scheme selects a complete, k-ary tree with fanout *k*, with ``kary:0``
-   meaning that rank 0 is the parent of all other ranks by convention.  The
-   ``binomial`` scheme selects a binomial tree topology of the minimum order
-   that fits the instance size.  Default: ``kary:32``, unless bootstrapping by
-   TOML configuration, then see :man5:`flux-config-bootstrap`.
+   A URI describing the TBON tree topology.  The following schemes are
+   available:
+
+   kary:k
+      A complete, k-ary tree with fanout *k*.  By convention, ``kary:0``
+      indicates that rank 0 is the parent of all other ranks.
+
+   mincrit[:k]
+      Minimize critical ranks.  The tree height is limited to three.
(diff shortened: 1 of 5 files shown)


garlick:
Since this seems like it could be useful, I reworked it a bit:
- renamed to "mincrit" (minimize critical nodes)
- the system selects a value for K if unspecified
- added tests and man pages
- updated PR description and title to reflect new name

garlick:
Yeah I was wondering if we should maybe set the default at 128 or 256 based on your earlier results. :shrug: The user can always specify the number of routers they want with `mincrit:N` although then they have to think a little bit about what the resulting leaf fanout is.

grondo:
eh, we could always change the default N later if needed. The important thing is having the topology available now.

garlick:
OK, setting MWP then.  Thanks!

PR #6811, opened 2025-05-13
broker: avoid malloc in event distribution critical path

garlick:
Problem: when the broker distributes event messages to overlay peers, it does a malloc/free per peer as it pushes and pops the peer id onto the message route stack.

This adds a (private) helper function to the message layer to streamline this case, avoiding the dynamic memory allocation.

This might not be worth the extra complexity, so marking WIP for now.

### DIFF
--- a/src/broker/overlay.c
+++ b/src/broker/overlay.c
@@ -36,6 +36,7 @@
 #include "src/common/libutil/monotime.h"
 #include "src/common/libutil/errprintf.h"
 #include "src/common/librouter/rpc_track.h"
+#include "src/common/libflux/message_route.h" // for msg_route_sendto()
 #include "ccan/str/str.h"
 
 #include "overlay.h"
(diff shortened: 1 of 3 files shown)


garlick:
Well, dropping the WIP on this, as it probably helps, although the net effect may be modest.

src/broker/overlay.c
```diff
@@ -36,6 +36,7 @@
 #include "src/common/libutil/monotime.h"
 #include "src/common/libutil/errprintf.h"
 #include "src/common/librouter/rpc_track.h"
+#include "src/common/libflux/message_route.h" // for msg_route_sendto()
```
grondo, line 39: Commit message typo: "is does a malloc"

grondo (approved):
LGTM! Just one commit message typo

garlick:
Thanks - fixed that and rebased on master. I'll set MWP.

PR #6815, opened 2025-05-14
remove restriction that taskmaps cannot assign differing counts of tasks to nodes in a job

### LINKED ISSUE #6810: use case: spawn first N ranks of a job on nodeid 0
This use case is motivated by the issues observed in #6806.

As a way to reduce critical nodes in resilient batch job while also using a tbon topology with multiple critical broker ranks, it would be valuable if there was a way to run multiple brokers in a batch instance onto the first node of the batch job. 

This would allow the first allocated node to remain as the sole critical node, while also allowing multiple brokers to share the load of event and message distribution so that that the other functions of the rank 0 broker get more CPU time. These extra brokers would be excluded from scheduling to avoid duplicated resources.

I thought perhaps we could do this now with a custom taskmap. For instance for a 1024 node batch job, we request 1026 "tasks" and force the first 3 brokers on the first node with `[[0,1,3,1],[1,1023,1,1]]` or similar (untested). However, there's currently a caveat when using a custom taskmap:

> The  provided  TASKMAP  must match the total number of tasks in the job 
> and the number of tasks per node assigned by  the  job  shell...

which means that a custom taskmap cannot currently assign a different task count to a node than the default (block taskmap). I'm not sure why this constraint is there, but perhaps it could be easily removed.

grondo:
This PR fixes the limitation of manual and hostfile taskmaps provided in jobs that prevents modifying the count of tasks assigned to nodes within a job. This allows, for example, a few extra tasks to be added to a job and assign them all to node 0:

```console
$ flux run -N4 -n8 --taskmap=manual:"[[0,1,5,1],[1,3,1,1]]" --label-io hostname
4: corona180
3: corona180
2: corona180
1: corona180
0: corona180
5: corona181
6: corona182
7: corona183
```

Fixes #6810


garlick (approved):
LGTM!

grondo:
rebased. Will set MWP. Thanks!

PR #6819, opened 2025-05-14
doc: add note on use of flux queue drain

chu11:
Problem: It is common to use `flux queue drain` to wait for all submitted jobs to finish in flux-batch(1) scripts.  However, this may not be obvious to new flux users.

Add a note about this in flux-queue(1) if users look for information on why this is used in flux batch scripts.

----

just an idea based on an e-mail thread

### DIFF
--- a/doc/man1/flux-queue.rst
+++ b/doc/man1/flux-queue.rst
@@ -186,6 +186,10 @@ be taken down for maintenance.
 
    Limit the time that the command` will block.
 
+.. note::
+   :program:`flux queue drain` can be used as a simple mechanism to wait for
+   all submitted jobs to complete in :command:`flux-batch` scripts.
+
 idle
 ----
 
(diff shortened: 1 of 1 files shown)


chu11:
> I hesitate a little bit on this recommendation, since unlike flux watch --all or flux job wait --all, putting flux queue drain last in a batch script won't cause the instance to exit with an error code if any of its jobs have failed.

My theory on adding this was that we sometimes have examples / tutorials that end with `flux queue drain` at the end of a batch script.  So a casual user might be "why is that going on?"

Would it be better if we caveat this note with something like `see flux-watch(1) for something better`?

garlick:
If we're going to keep this, it should probably be folded in with the other system example that is there inline, rather than putting this one in a note box when that one isn't.  And I would just caveat that its exit code is not influenced by job failures, see blah blah for an alternative that does that.

PR #6821, opened 2025-05-15
flux-module: require argument for flux module stats

### LINKED ISSUE #6636: flux-module: improve error message / require user input
```
> flux module stats
flux-module: broker.stats-get: Function not implemented
```

A cursory scan did not find an old`broker.stats-get`, so I bet this was set as the default because "broker.FOO" was the most common default thing to do a long time ago.

Perhaps should just make sure user supplied a module name and/or give a much better error message.

chu11:
Problem: The default target for flux module stats points to a RPC
target that no longer exists.

Require a stats target with flux module stats.

Fixes #6636

---

As an aside, the manpage for `flux-module(1)` and the `--help` output already documents this input as required.  So no need to fix those up.

### DIFF
--- a/src/cmd/flux-module.c
+++ b/src/cmd/flux-module.c
@@ -648,11 +648,11 @@ int cmd_stats (optparse_t *p, int argc, char **argv)
     flux_future_t *f = NULL;
     flux_t *h;
 
-    if ((n = optparse_option_index (p)) < argc - 1) {
+    if ((n = optparse_option_index (p)) != (argc - 1)) {
         optparse_print_usage (p);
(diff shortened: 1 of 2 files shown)


grondo (approved):
Oh, nice catch. I didn't even realize `flux module stats` had a default target.

Note that I was confused by the description here. While "input" might be technically correct, it may be more specific to say "require module target for `flux module stats`" or "require argument to `flux module stats`". A minor thing though...

chu11:
@grondo thanks!  Fixed up those small things and will set MWP.

PR #6822, opened 2025-05-16
add method to request extra brokers on node 0

grondo:
This PR is a WIP pending tests. I'm not sure if the solution proposed here is the best we can come up with, but I thought I'd post it now to get feedback.

This PR implements a hidden `--add-brokers=N` option to `flux batch` and `flux alloc` that can be used to force N extra brokers on nodeid 0. This is hidden for now because I'm not sure it is a great long-term approach, but it allows us to give select users a way to supplement their jobs with these extra brokers.

The option should be paired with an appropriate `tbon.topo` such that all critical ranks of the job are placed on node rank 0. The extra ranks are automatically excluded. Users may choose to also exclude or drain rank 0 since it will have more flux-broker processes on it.

Simple example:
```console
$ src/cmd/flux alloc -N4 --add-brokers=1 --conf=tbon.topo=kary:2
$ flux resource status 
       STATE UP NNODES NODELIST
       avail  ✔      4 pi[3,1-2,4]
     exclude  ✔      1 pi3
$ flux resource list
     STATE NNODES NCORES NGPUS NODELIST
      free      4     24     0 pi[3,1-2,4]
 allocated      0      0     0 
      down      0      0     0 
$ flux overlay status
0 pi3: full
├─ 1 pi3: full
│  ├─ 3 pi2: full
│  └─ 4 pi4: full
└─ 2 pi1: full
```

Note that `flux jobs` properly reports this instance has 5 tasks running on 4 nodes:
```console
$ flux --parent jobs
ƒ29BhqUNKg3H grondo   flux        R      5      4   39.94m pi[3,1-2,4]
```
```

### DIFF
--- a/src/bindings/python/flux/cli/alloc.py
+++ b/src/bindings/python/flux/cli/alloc.py
@@ -19,11 +19,9 @@ from flux.cli import base
 from flux.uri import JobURI
 
 
-class AllocCmd(base.MiniCmd):
+class AllocCmd(base.BatchAllocCmd):
     def __init__(self, prog, usage=None, description=None):
-        self.t0 = None
         super().__init__(prog, usage, description, exclude_io=True)
-        base.add_batch_alloc_args(self.parser)
         self.parser.add_argument(
             "-v",
             "--verbose",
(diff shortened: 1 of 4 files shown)


grondo:
BTW, I did consider adding a `--resilient` option that selected a "good" value for `tbon.topo` and `add_brokers` but I feel like that's premature at this point.

A big drawback here is that we don't know the size of the instance in the cli submission tools unless `-N, --nodes` is used, so that is required for now. Eventually, it might be nice to add extra brokers and modify the topology after R is known.

garlick:
This seems fine to me.  I agree that later we might want to have a more high level option that tweaks several other low level options including this one.  Something like `--resiliency-mode=NAME` where NAME selects a strategy.  Anyway, I'm good with moving forward with this as a useful step.

wihobbs:
> --resiliency-mode=NAME where NAME selects a strategy

Seems like maybe a good use for CLI plugins? (the successor to #6691) @garlick, I think you missed the saga on that but we decided to close the PR and go with long options, so soon (TM) we'll have a facility for a plugin that adds `--ex-resilient=NAME` to the command line, with callbacks that support selecting multiple configuration options.

grondo:
Added a simple test and removed WIP.

garlick (approved):
LGTM!

grondo:
Thanks! I've set MWP

PR #6823, opened 2025-05-17
use local ipc for TBON when brokers are co-located

garlick:
This allows IPC to be used instead of TCP between brokers when they are co-located on the same node.
Using @grondo's custom taskmap from the description of #6815, this starts a flux instance with several ranks on the first node and the rest on their own nodes:
```
$ flux run -q batch -N4 -n8 --taskmap=manual:"[[0,1,5,1],[1,3,1,1]]" -o pty.interactive flux start
$ flux exec -x 0 --label-io flux getattr tbon.parent-endpoint
5: tcp://[::ffff:192.168.88.246]:45175
6: tcp://[::ffff:192.168.88.246]:45175
7: tcp://[::ffff:192.168.88.246]:45175
2: ipc:///tmp/flux-jWd8Pr/tbon-0
3: ipc:///tmp/flux-jWd8Pr/tbon-0
4: ipc:///tmp/flux-jWd8Pr/tbon-0
1: ipc:///tmp/flux-jWd8Pr/tbon-0
```

WIP because this needs tests and I'm struggling with how that might be done.  Since its EOD I figured I'd get this posted for now and return for another look with a fresh brain.

Note this PR did require significant refactoring in the broker's PMI client.  I think it's more readable now though  overall.

### DIFF
--- a/doc/man1/flux-start.rst
+++ b/doc/man1/flux-start.rst
@@ -207,8 +207,8 @@ OPTIONS
    Set the pmi clique mode, which determines how ``PMI_process_mapping`` is set
    in the PMI server used to bootstrap the brokers.  If ``none``, the mapping
    is not created.  If ``single``, all brokers are placed in one clique. If
-   ``per-broker``, each broker is placed in its own clique.
-   Default: ``single``.
+   ``per-broker``, each broker is placed in its own clique.  Otherwise the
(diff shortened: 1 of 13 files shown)


garlick:
I think this is review-worthy so dropping the WIP.

grondo (approved):
This is great and LGTM!

garlick:
Thanks!  I'll set MWP.

Issue #6804, opened 2025-05-08
support something like Slurm `--dependency=singleton`


grondo:
Users have requested support for something like `--dependency=singleton` in Slurm:

> singleton
>   This job can begin execution after any previously launched jobs sharing the same job name and user have terminated. In other words, only one job by that name and owned by that user can be running or suspended at any point in time.

The implementation of this feature would be pretty straightforward in a jobtap plugin, however, for feature parity the plugin would have to iterate all active jobs to see if there are any matching userid,name pairs on the system already, which could be slow on a system with a lot of jobs.

If we modified the feature to "This dependency is only satisfied when there are no other previously submitted singleton jobs with the same userid and job name", then the plugin would not have to search all jobs on first use, it could instead create some state. However, I'm not sure it is a good idea to use the same name as a Slurm feature without being completely compatible.

grondo:
No. An "after job name" dependency would release all jobs that have a dependency on job "name" when that job finishes. The singleton specifically only allows a single job with a matching user/name pair to be released at any time (thus, the singleton name)

grondo:
I should mention the specific use case is that a user wants to submit N jobs with name M at the same time, but ensure only one is running at any given time. It is a simpler alternative to submitting a chain of dependent 'afterany' jobs.

garlick:
For reference: https://slurm.schedmd.com/sbatch.html#OPT_singleton

This seems tricky in Flux since we don't release jobs to the scheduler until their dependencies are met.  If there are multiple jobs with the specified name, how does the job manager pick one to release vs another?  It might pick one that can't run.

grondo:
The specific use case here is a set of identical jobs where if more than one was running at time they would conflict. This is the same as a chain of dependent jobs, except easier to submit. I have also seen this feature used to have some post-processing run after another set of jobs (with the same name I guess) finish.

This still might be a bit tricky because a singleton job can be dependent on one or more jobs that don't even have a dependency (i.e. those with the same name).

PR #6827, opened 2025-05-20
support `singleton` dependency scheme in job submission

grondo:
This PR adds a very simple implementation of a `singleton` dependency scheme as described in #6804.

In order to support the requirement of "a singleton job is not eligible to run until all jobs of the same userid and job name have become inactive", the plugin has to track a count of _all_ jobs by userid,name pair, even when no singleton jobs are being submitted. In practice, this seems to have little impact on throughput, so perhaps this is ok. Note that, for simplicity, jobs without an explicit name are ignored by the plugin (i.e. first argument does not count as the job name).

Singleton jobs are then placed on a list in submission order and are only released when the count of active jobs with the same userid and job name equal the number of currently held singletons.

### DIFF
--- a/src/modules/job-manager/plugins/dependency-singleton.c
+++ b/src/modules/job-manager/plugins/dependency-singleton.c
@@ -1,0 +1,369 @@
+/************************************************************\
+ * Copyright 2025 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* dependency-singleton.c
+ *
+ * Support singleton dependency scheme which places a dependency
+ * on a job which is only released when there are no other active
+ * jobs of the same userid and job name which are not also already
+ * held with a singleton dependency.
+ *
+ * Notes:
+ * - counts of active jobs with the same userid/job-name are maintained
+ *   in a global hash
+ * - jobs without an explicit name are ignored
+ * - jobs submitted with dependency=singleton are placed on a list, also
+ *   hashed by userid/job-name
+ * - when the active job count for a userid/job-name is decremented and
+ *   equals the count of singletons for that same pair, a singleton job
+ *   is released and removed from the list
+ * - it is an error to submit a singleton job without an explicit job name
+ *
+ */
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+#include <jansson.h>
+#include <flux/core.h>
+#include <flux/jobtap.h>
+
+#include "src/common/libjob/idf58.h"
+#include "src/common/libczmqcontainers/czmq_containers.h"
+#include "src/common/libutil/hola.h"
+#include "ccan/str/str.h"
+#include "ccan/ptrint/ptrint.h"
+
+struct singleton_ctx {
+    zhashx_t *counts; // uid:job-name => ACTIVE job counts
+                      // (with or without singleton dep)
+    struct hola *singletons; // uid:job-name => list of struct singleton
+                             // (jobs with singleton dep)
+};
+
+struct singleton {
+    flux_jobid_t id;
+};
+
+static struct singleton_ctx *global_ctx = NULL;
+
+static void singleton_ctx_destroy (struct singleton_ctx *sctx)
+{
+    if (sctx) {
+        int saved_errno = errno;
+        zhashx_destroy (&sctx->counts);
+        hola_destroy (sctx->singletons);
+        free (sctx);
+        errno = saved_errno;
+    }
+}
+
+static struct singleton_ctx *singleton_ctx_create (void)
+{
+    struct singleton_ctx *sctx;
+    int flags = HOLA_AUTOCREATE | HOLA_AUTODESTROY;
+
+    if (!(sctx = malloc (sizeof (*sctx)))
+        || !(sctx->counts = zhashx_new ())
+        || !(sctx->singletons = hola_create (flags)))
+        goto error;
+    return sctx;
+error:
+    singleton_ctx_destroy (sctx);
+    return NULL;
+}
+
+static int singleton_key_create (char *key,
+                                 size_t len,
+                                 uint32_t userid,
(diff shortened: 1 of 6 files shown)


garlick:
Silly question:  would the code be mostly identical and the result more flexible if this were `--dependency=maxname:N` where N=1 in the singleton case but could be any value?

grondo:
Probably similar, instead of checking if `count == #singletons` the code might have to do something like
``` 
while count <= #singletons:
   release a singleton
```
However, I can't think of a use case for this now, and it would require a bit of extra testing. Perhaps someone could easily add an implementation later if there's a need?

garlick:
Agreed, keeping it focused on the actual user request makes sense, and easy to enhance later!

grondo:
Just pushed a small change to drop the `dependency-add` and immediate `dependency-remove` if there are no other matching userid/job-name pairs when a dependency=singleton job is submitted (removes a little code). Also expanded a comment.

src/modules/job-manager/plugins/dependency-singleton.c
```diff
@@ -0,0 +1,344 @@
...
+                                const char *key,
+                                flux_jobid_t id)
+{
+    struct singleton *s;
+
+    if (!(s = calloc (1, sizeof (*s)))
+        || flux_jobtap_job_aux_set (p, id, NULL, s, (flux_free_f) free) < 0) {
+        free (s);
+        return -1;
+    }
```
garlick, line 102: Deferring destruction of these tiny objects until plugin unload is probably fine but I had to go look at the aux container implementation to remind myself that if it accumulates thousands of anonymous entries over the long runtime of a system instance, it doesn't slow down named lookups (b/c we put anons at the end - yay).

It might be nicer to just set the list destructor and also do an explicit free where these are popped off the list in the inactive callback.

src/modules/job-manager/plugins/dependency-singleton.c
```diff
@@ -0,0 +1,344 @@
...
+        return 0;
+
+    if (singleton_key_create (key, sizeof (key), userid, name) < 0)
+        return -1;
+
+    if ((count = ptr2int (zhashx_lookup (sctx->counts, key)) + value) > 0
+        && count == singleton_list_count (sctx, key)) {
+        /*  All active jobs of this uid/name pair are waiting on a singleton
+         *  dependency. Pop the next job on the list and release it:
+         */
```
garlick, line 165: I'm having a hard time reading the parens and side effects today.  It might be clearer to break this up like
```c
count = ptr2int (zhashx_lookup (sctx->counts, key));
count += value;
if (count > 0 && count == singleton_list_count (sctx, key)) {
```
Eh maybe I just need more :coffee:  You can take this one or leave it.

src/modules/job-manager/plugins/dependency-singleton.c
```diff
@@ -0,0 +1,344 @@
...
+        return flux_jobtap_reject_job (p,
+                                       args,
+                                       "singleton jobs require a job name");
+
+    if (singleton_key_create (key, sizeof (key), userid, name) < 0)
+        return flux_jobtap_reject_job (p, args, "error creating singleton key");
+
+    /* Get current count of matching jobs in SCHED+RUN+CLEANUP state.
+     * If there are no other matching jobs then release this one immediately.
+     *
```
garlick, line 229: Don't forget PRIORITY

garlick (approved):
LGTM!  I made some comments but they all amount to cosmetic nits so feel free to ignore.

src/modules/job-manager/plugins/dependency-singleton.c
grondo: Perhaps it was missed that this is `flux_jobtap_job_aux_set()` not `flux_plugin_aux_set()`?

src/modules/job-manager/plugins/dependency-singleton.c
grondo: Ok.. if going that far perhaps a function that abstracts fetching the current `count`. That will be self documenting.

src/modules/job-manager/plugins/dependency-singleton.c
garlick: OH so it's just the life of the job, well double nevermind then!  Sorry.

grondo:
If addition of this plugin is a bit contentious, we could easily package it separate from flux-core and load it in the system instance when necessary via configuration. Then it could be dropped or simulated when we have true symbolic dependencies.

src/modules/job-manager/plugins/dependency-singleton.c
garlick: I just couldn't quickly spot if count was being incremented in the side effect assignment or if the test was that count + value > 0.

grondo: New code looks like this, which I feel is much improved.
Better?

```c
    count = get_current_active_count (sctx, key) + value;
    if (count == singleton_list_count (sctx, key)
        && (s = singleton_list_pop (sctx, key))) {
        /*  All active jobs of this uid/name pair are waiting on a singleton
         *  dependency. Pop the next job on the list and release it:
         */
        flux_jobtap_dependency_remove (p, s->id, "singleton");
    }
    set_current_active_count (sctx, key, count);
```

grondo:
Pushed some updates based on comments above.

Also fixed a minor issue in the tests that made the job submitted as a fake alternate uid fail.

grondo:
Ok, if the slight refactor is an improvement I'll set MWP.

grondo:
Part of me keeps thinking about something @trws said. Maybe this plugin should support a name "path" argument which, when specified, overrides the job name for which the singleton applies. This would allow a set of singletons to be submitted more easily (less typing) with `--dependency=singleton:NAME` and would also allow a singleton job to reference other jobs by name while running with a different job name.

This change would be trivial, but I hesitate because it doesn't seem necessary for the use case and our time could perhaps better be spent on the true symbolic dependency work. Any other thoughts?

garlick:
That does seem more flexible.  Up to you - I'm fine either way.

grondo:
Meh. It is pretty simple except it unnecessarily complicates updating the `counts` hash since it would require checking if every job has a `singleton` dependency. Probably better to just keep things as is for now.

Setting MWP now.

PR #6828, opened 2025-05-20
flux-fsck: parallelize valref blobref checks

chu11:
Problem: The values in a valref blobref array are checked synchronously one by one.  This can be quite slow.

Parallelize the valref blobref array verification by making each blobref lookup asynchronous.

### DIFF
--- a/src/cmd/builtin/fsck.c
+++ b/src/cmd/builtin/fsck.c
@@ -28,9 +28,16 @@
 
 #include "builtin.h"
 
+#define BLOBREF_ASYNC_MAX 1000
+
 static void fsck_treeobj (flux_t *h,
                           const char *path,
                           json_t *treeobj);
+
+static void valref_load (flux_t *h,
+                         json_t *treeobj,
+                         int index,
+                         void *arg);
 
 static bool verbose;
 static bool quiet;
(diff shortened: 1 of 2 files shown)


src/cmd/builtin/fsck.c
```diff
@@ -54,35 +54,53 @@ void read_error (const char *fmt, ...)
...
+                        (char *)flux_future_aux_get (f, "index"));
+        else
+            read_error ("%s: error retrieving blobref index=%s: %s",
+                        path,
+                        (char *)flux_future_aux_get (f, "index"),
+                        future_strerror (f, errno));
+        errorcount++;
+    }
+    flux_future_destroy (f);
+}
```
garlick, line 76: I'm a little bit concerned about making this completely open loop since it leaves the number of outstanding RPCs unbounded (or at most bound by metadata traversal time).  Maybe a future should be fulfilled once each _key_ has been fully read, and then the main code can do _those_ one at a time, or N at a time depending on how aggressive we want to be.   Reporting could be moved from here to when the fulfilled future is checked, so that we just count/report the bad keys rather than the bad blobrefs, which is probably too much detail for admins (though maybe nice for us at an increased verbosity)

If you're also going to parallelize `flux-dump` then perhaps it makes sense to develop a shared future-returning function for traversing a valref.    A flag could say what it saves for access once the future is fulfilled (`flux_msglist_t` of responeses for iteration like in `flux-dump` or maybe nothing for `flux-fsck`).

garlick (commented):
I made a few initial comments.

src/cmd/builtin/fsck.c
chu11: hmmm, i see your point.  Before we would have had a max of one "blobref error", but now there could be like a gajillion if that valref is super corrupted.

Lemme experiment with some of your ideas/suggestions.

chu11:
I began to work on what would be a async-blobref check hidden behind a future as discussed above, but                                       

A) implementing this would be 90% like a composite future, so I thought "could i just use a composite?"
B) issue #6831 would have gotten in the way

Eventually decided to go down that path another day.

re-pushed with the following changes

- rate limit the blobref lookups to 1000 RPCs max (can debate this max, it was semi-arbitrarily chosen)
- only output a single error message on non-verbose mode
- output each blobref error on verbose mode

hopefully it's ok for now

garlick (approved):
OK for now.  We might want to share code with `flux-dump` when it gets worked on, and this won't be it, but that's OK, one thing at a time :-)

chu11:
hmmm, i got this concerning failure on a builder

```
ok 15 - unload kvs
  unload kvs
  
  expecting success: 
  	test_must_fail flux fsck 2> fsckerrors1.out &&
  	count=$(cat fsckerrors1.out | wc -l) &&
  	test $count -eq 3 &&
  	grep "dir\.b" fsckerrors1.out | grep "missing blobref(s)" &&
  	grep "Total errors: 1" fsckerrors1.out
  
  not ok 16 - flux-fsck detects errors
 
```

but the verbose fsck right below it passed

```
  expecting success: 
  	test_must_fail flux fsck --verbose 2> fsckerrors1V.out &&
  	grep "dir\.b" fsckerrors1V.out | grep "missing blobref" | grep "index=1" &&
  	grep "Total errors: 1" fsckerrors1.out
  
  dir.b: missing blobref index=1
  Total errors: 1
  ok 17 - flux-fsck --verbose outputs details
  flux-fsck --verbose outputs details
```

I can't help but wonder if there is a race between the "unload kvs" and the fsck.  Like perhaps we should `flux kvs sync` it before unloading the kvs module (note #6240)

PR #6830, opened 2025-05-21
doc: correct flux_future_set_reactor() prototype

chu11:
Problem: The documented function prototype for flux_future_set_reactor(3) is incorrect, likely due to a cut and paste error.

Correct the function prototype.


garlick (approved):
Oops!  Good catch.

PR #6832, opened 2025-05-22
libjob: add infrastructure for non-integer counts

sam-maloney:
Problem: non-integer counts from RFC 14 are largely unsupported in flux-core.

This PR adds some initial infrastructure as a first step towards enabling non-integer counts; N.B. at present the new code is not actually used to do anything, this PR only includes the new struct definitions along with creation, encoding/decoding, and iteration functions (as well as initial unit testing).

The `count` wrapper structs are intended to present a similar interface (albeit a limited subset thereof) as `libidset`, particularly since flux-framework/rfc#442 proposes allowing counts to be given as idsets (which this PR implements). The iteration functions therefore intend to allow code to be agnostic to which type of count was actually given (integer, idset, or range). Any feedback on the design, etc. would be much appreciated!

The string representation for decoding/encoding of ranges is documented in the proposed RFC 45 at flux-framework/rfc#453

In conjunction with #6705 this will then allow non-integer ranges to be enabled in more parts of the code, such as job-listing, sched-simple, etc.


src/common/libjob/count.c
```diff
@@ -0,0 +1,192 @@
+/************************************************************\
+ * Copyright 2018 Lawrence Livermore National Security, LLC
```
grondo: Just a note to update Copyright date to 2025 here and in any other files.

src/common/libjob/count.c
```diff
@@ -295,6 +295,51 @@ char *count_encode (const struct count *count, int flags)
     return str;
 }
 
+unsigned int count_first (const struct count *count)
```
grondo, line 298: nit: typo in commit message "vlues"

grondo (approved):
@sam-maloney, thanks, this is really well done! So far I had only a couple trivial comments inline.  Thanks also for the extensive testing.

I do wonder if `counts.[ch]` belong in libutil, but for now libjob actually does seem a bit more appropriate and it could always be moved.

src/common/libjob/count.c
sam-maloney: Good catch, thanks!

src/common/libjob/count.c
sam-maloney: 🗓️🙃

sam-maloney:
> I do wonder if `counts.[ch]` belong in libutil, but for now libjob actually does seem a bit more appropriate and it could always be moved.

I definitely wasn't sure at all where best to put this, but since it grew from the parsing of jobspec in the shell init, which uses `jj`, I stuck the new code in `libjob` as well...

grondo (approved):
LGTM! I'll set MWP.

Issue #6835, opened 2025-05-24
Parameters that  influence `make check`


ocaisa:
I'm building an EasyBuild recipe for Flux, and as part of the recipe I want to run `make check`. Running `make check` without `-j` is quite slow (build time was about 5 minutes on my laptop, tests ran for 1h40), that's not a problem but maybe it is worth noting that parallelism for the tests with `-j` is possible.

I also couldn't find any explicit documentation on options that influence the tests. After some searching I found a number of options in https://github.com/flux-framework/flux-core/blob/master/src/test/docker/docker-run-checks.sh .

In my own test runs, I had a single failing test within `t/t2602-job-shell.log`. I'll rebuild next week on another machine with more resources to see if this failure is reproducible. It looks like the test suite stops once it hits a failure, is there a way to complete the full test suite and see if there are further  problems?

grondo:
> For make -j 2 -k check I see multiple summaries (similar to below), but it seems to stop once I hit the single failure. It may be the case that this is the last set of checks and it is a coincidence (I have no idea, the summary below looks to be related to make check-TESTS).

There is a set of unit tests in mot build subdirectories and these are separately summarized by automake's `make check`. That is the source of the multiple summaries. The last set of checks are system checks in the `t` directory, and this is indeed where your `make check` is failing. However, based on the count of reported tests it does seem to have run all the checks.

> If you give me the ok, I will just ignore that specific test.

Actually if you could run this particular test by hand and paste the output in this issue, that would be very helpful. I'm guessing this is a problem with the test and not the code, but it would be good if we fixed it either way.

So in the build directory can you try:

```console
$ cd t
$ ./t2602-job-shell.t -d -v
```
and share the output. If your build directory is separate from the source directory, you should specify the full path to the `t2602-job-shell.t` file, i.e. `/path/to/srcdir/t/t2602-job-shell.t -d -v`.

Thanks!

ocaisa:
> Ah, it looks like the `J.new` generated on your system is too long for bash. This is probably an error in the test, which should read the KVS value from stdin and not use the command line. I'll open a quick PR.
> 
> Can you try this patch?
> 
> diff --git a/t/t2602-job-shell.t b/t/t2602-job-shell.t
> index 14b7a2134..accbf9d79 100755
> --- a/t/t2602-job-shell.t
> +++ b/t/t2602-job-shell.t
> @@ -33,8 +33,7 @@ test_expect_success 'job-shell: fails on modified J' '
[...]

PR #6836, opened 2025-05-27
require Lua `posix` in `configure` and document some tips for running the testsuite

grondo:
This PR addresses some issues described by @ocaisa in #6835.

We now just go ahead and fail in configure if `luaposix` is not found to avoid confusion.

Also, a couple extra tips re: the testsuite are added to the top-level README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,9 +33,13 @@ are available for [redhat](scripts/install-deps-rpm.sh) and
 ```
 ./autogen.sh   # skip if building from a release tarball
 ./configure
-make
-make check
+make -jN
+make -jN check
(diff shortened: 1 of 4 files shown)


garlick (approved):
LGTM.  I verified that configure fails when I de-install lua-posix on my pop os 22.04 development desktop, and succeeds when I put it back :-)

grondo:
I pushed another quick fix to a test here. Happy to split it off in another PR if desired.

grondo:
I dropped that extra commit and will open a separate PR. Setting MWP.

PR #6839, opened 2025-05-28
content-sqlite: deprecate version 0 checkpoint

chu11:
Problem: The content-sqlite module currently has backwards compatibility to "version 0" checkpoints, i.e. just a blobref without any metadata.  This backwards compatibility was added in

9958fbb976a4d263085cd68232a825aa3737076e

over 3 years ago (2/18/2022).  Future work will re-do how checkpoints are done in content-sqlite and this backwards compatibility makes some future work annoying.

Solution: Remove backwards compatibility of "version 0" checkpoints.

Remove related tests in t2010-kvs-snapshot-restore.t and helper test files.


garlick (approved):
LGTM

PR #6840, opened 2025-05-28
content-s3: remove module

chu11:
Problem: The content-s3 module is largely unused.  It is a headache maintaining it with little benefit.

Solution: Remove the content-s3 module.  It can be revived in the future if need be.  Along with the removal of the content-s3 module:

Remove all content-s3 tests in:

t0024-content-s3.t
t2008-althash.t
t3200-instance-restart.t

Remove content-s3 from build matrix.

Remove all libs3 libraries from Dockerfiles.

----

Side notes:

in the build matrix I only removed `content-s3` from this builder, but left the builder.  AFAICT, this is the only builder that tests DISTCHECK and gcc-12.  My assumption is content-s3 was "stuffed" into this builder to get coverage w/o creating too many builder permutations.

```
-# Debian: gcc-12, content-s3, distcheck
+# Debian: gcc-12, distcheck
 matrix.add_build(
-    name="bookworm - gcc-12,content-s3,distcheck",
+    name="bookworm - gcc-12,distcheck",
     image="bookworm",
     env=dict(
         CC="gcc-12",
@@ -278,7 +260,6 @@ matrix.add_build(
         DISTCHECK="t",
     ),
     args="--with-flux-security",
-    test_s3=True,
 )

```

I also removed `libs3-dev` from all Dockerfiles.  That is maybe unnecessary.


src/test/docker/docker-run-checks.sh
```diff
@@ -238,10 +238,6 @@ else
         -e INCEPTION \
         -e ASAN_OPTIONS \
         -e BUILD_DIR \
-        -e S3_ACCESS_KEY_ID \
-        -e S3_SECRET_ACCESS_KEY \
-        -e S3_HOSTNAME \
-        -e S3_BUCKET \
```
grondo, line 244: The same commit that added these environment variables also added `--network=host` to the docker command  line. That should probably be removed as well (and I vaguely remember it was causing issues recently.)

chu11: your memory is incredible :-)

I left the `--network=host` in the podman call in `docker-run-systest.sh` for the time being.  It might not be needed, but I wasn't super sure.

grondo:
Yes, under Settings->Branches->Branch Protections->master->edit->"require status checks to pass before merging"

This is the list of status checks that Github and Mergify are waiting on before turning green (and which appear in the dynamically updated status checks in github UI)

garlick (approved):
Thanks, this will make it a bit easier to work on the content back end.
I think we could bring this back as a standalone framework project should we need it in the future.

PR #6858, opened 2025-06-03
add more arguments to Python `JobspecV1` factory methods

### LINKED ISSUE #6833: python: add more parameters to Jobspec constructor/factory functions
@trws brought this up in our meeting yesterday. The Jobspec factory functions currently take a very limited set of parameters, and then require a caller to amend/update the returned jobspec object to get a functional jobspec. For example, the jobspec is created with an empty `attributes.system.environment` and there's no `environment` parameter available.

Tom suggested we add parameters for most, if not all, parameters that are available on the corresponding command line utilitees. This is probably pretty simple to add without affecting current users, and would be a bit quality of life improvement for users that submit jobs via Python.

grondo:
This PR adds some more arguments to the `JobspecV1` factory methods as described in #6833 so that more fully formed jobspec can be instantiated with a single method call.

This PR doesn't go as far as adding _all_ the options offered by the submission CLI commands. However, job parameters that are an annoyance to set after the fact like the environment and cwd are there, as well as output, error, and input paths, etc.

One thing I didn't do was move support for some of the more user-friendly aspects of the CLI into the Python methods. For instance, the `environment` and `rlimits` arguments just take straight mappings, instead the "rule" based expressions supported by the CLI. We could change that, but I wasn't sure if it was necessary in the Python API.

Also, some other missing parameters include `--dependency`, `--requires`, and `--begin-time` -- all of which actually are mostly implemented in the CLI, and I wasn't sure exactly if those should go in the factory methods. They could be added later I guess.

Fixes #6833

### DIFF
--- a/src/bindings/python/flux/job/Jobspec.py
+++ b/src/bindings/python/flux/job/Jobspec.py
@@ -822,6 +822,57 @@ class JobspecV1(Jobspec):
             if task["count"].keys().isdisjoint({"per_slot", "total"}):
                 raise ValueError("count per_slot or total must be set")
 
+    def _set_extra_options(
+        self,
+        duration=None,
+        environment=None,
+        env_expand=None,
+        cwd=None,
+        rlimits=None,
+        name=None,
+        input=None,
+        output=None,
+        error=None,
+        label_io=False,
+        unbuffered=False,
+        queue=None,
+        bank=None,
+    ):
+        """Process common jobspec factory args and return amended jobspec"""
+        if duration is not None:
+            self.duration = duration
+        if environment is not None:
+            self.environment = environment
+        if env_expand is not None:
+            if not isinstance(env_expand, abc.Mapping):
+                raise ValueError("env_expand must be a mapping")
+            self.setattr_shell_option("env-expand", env_expand)
+        if cwd is not None:
+            self.cwd = cwd
+        if rlimits is not None:
+            if not isinstance(rlimits, abc.Mapping):
+                raise ValueError("rlimits must be a mapping")
+            self.setattr_shell_option("rlimit", rlimits)
+        if name is not None:
+            self.setattr("job.name", name)
+        if input is not None:
+            self.input = input
+        if output is not None and output not in ("none", "kvs"):
+            self.output = output
+            if label_io:
+                self.setattr_shell_option("output.stdout.label", True)
+        if error is not None and error not in ("none", "kvs"):
+            self.error = error
+            if label_io:
+                self.setattr_shell_option("output.stderr.label", True)
+        if unbuffered:
+            self.unbuffered = True
+        if queue is not None:
+            self.setattr("queue", queue)
+        if bank is not None:
+            self.setattr("bank", bank)
+        return self
+
     @classmethod
     # pylint: disable=too-many-branches,too-many-locals,too-many-statements
     def per_resource(
@@ -833,6 +884,7 @@ class JobspecV1(Jobspec):
         per_resource_count=None,
         gpus_per_node=None,
         exclusive=False,
+        **kwargs,
     ):
         """
         Factory function that builds a v1 jobspec from an explicit count
@@ -850,6 +902,8 @@ class JobspecV1(Jobspec):
                                 `per_resource_type`
             gpus_per_node: With nnodes, request a number of gpus per node
             exclusive: with nnodes, request whole nodes exclusively
+            **kwargs: Extra named arguments as accepted in
+                :func:`~JobspecV1.from_command`
         """
 
         #  Handle per-resource specification:
@@ -940,7 +994,7 @@ class JobspecV1(Jobspec):
         attributes = {"system": {"duration": 0}}
         if per_resource:
             set_treedict(attributes, "system.shell.options.per-resource", per_resource)
-        return cls(resources, tasks, attributes=attributes)
+        return cls(resources, tasks, attributes=attributes)._set_extra_options(**kwargs)
 
     @classmethod
     # pylint: disable=too-many-branches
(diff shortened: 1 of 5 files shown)


src/bindings/python/flux/job/Jobspec.py
```diff
@@ -741,6 +741,61 @@ def _v1_validate(resources, tasks, kwargs):
...
+        if unbuffered:
+            #  For output, set the buffer.type to none and reduce the configured
+            #  event batch-timeout to something very small.
+            self.setattr_shell_option("output.stdout.buffer.type", "none")
+            self.setattr_shell_option("output.stderr.buffer.type", "none")
+            self.setattr_shell_option("output.batch-timeout", 0.05)
+        if queue is not None:
+            self.setattr("queue", queue)
+        if bank is not None:
+            self.setattr("bank", bank)
```
cmoussa1, line 873: Since you're adding some flux-accounting attributes, you could consider adding `project` (the flux-accounting implementation of "WCKeys")  as another option here too, but not a huge deal either way; I'm not sure how much use projects get these days. I can't remember if it has to be an official jobspec v1 property to be listed here, though, so if so, feel free to ignore me.

cmoussa1 (approved):
LGTM!

src/bindings/python/flux/job/Jobspec.py
grondo: I think we were just trying to add options that are supported in the submission CLI commands :shrug:. Since the submission commands support `--bank` and `--queue`, I've added those here. However, I did leave other stuff out somewhat arbitrarily so I'm open to other ideas.

cmoussa1: Ah I see. I'm sorry about that! In that case, you can ignore this suggestion 😅

grondo: NP! It was a good point to bring up.

t/python/t0010-job.py
```diff
@@ -396,12 +396,82 @@ def test_22_from_batch_command(self):
...
+        self.assertTrue(jobspec.getattr("system.shell.options.output.stderr.label"))
+        self.assertEqual(
+            jobspec.getattr("system.shell.options.output.stdout.buffer.type"),
+            "none",
+        )
+        self.assertEqual(jobspec.getattr("system.queue"), "testq")
+        self.assertEqual(jobspec.getattr("system.bank"), "testbank")
 
     def test_23_from_nest_command(self):
         """Test that `from_batch_command` produces a valid jobspec"""
```
trws, line 478: Next door I realize but I assume this should be `from_nest_command`?
```suggestion
        """Test that `from_nest_command` produces a valid jobspec"""
```

src/bindings/python/flux/job/Jobspec.py
```diff
@@ -741,6 +741,61 @@ def _v1_validate(resources, tasks, kwargs):
...
+        if stderr is not None and stderr not in ("none", "kvs"):
+            self.stderr = stderr
+            if label_io:
+                self.setattr_shell_option("output.stderr.label", True)
+        if unbuffered:
+            #  For output, set the buffer.type to none and reduce the configured
+            #  event batch-timeout to something very small.
+            self.setattr_shell_option("output.stdout.buffer.type", "none")
+            self.setattr_shell_option("output.stderr.buffer.type", "none")
+            self.setattr_shell_option("output.batch-timeout", 0.05)
```
trws: This one seems to be the odd one out, in that it can be set with a keyword argument but not with a property is that right?  I think all of the others are both at this point.

trws:
I really like this, had one tiny docstring tweak and a question but it looks really good @grondo.

t/python/t0010-job.py
grondo: Good catch! I'll add a fix for that.

src/bindings/python/flux/job/Jobspec.py
grondo: Yeah, we never did add a setter/getter for that one. Do you feel we should add that?

Oh, the other thing I did change what that the factory function arguments for file output are `stdout`, `stderr`, `stdin`, whereas the command line arguments are `output`, `error`, and `input`. I wonder if I should tweak them to match.

trws: I was waffling on it when I asked, but having given it some thought I think yes.  That way both modes work, and we could handle all they keyword args somewhat uniformly with a version of `if <name> is not None: self.name = name` since the logic can live in the property.  It's not a strong preference, but feels more regular to me somehow.

On the args, that's a good point.  The difference isn't ideal, I'm thinking maybe we add aliases that are the command line names, then if we want to squash the `std*` ones we could mark them deprecated for a while at some point.

wihobbs:
Hey @grondo, thanks for this improvement!

The original request from Ramesh was for explicit documentation on what _could_ be set in the `JobspecV1` factory methods vs. what had to be set by attribute modification of an initialized `Jobspec`. Given that, @trws requested that we be able to set everything in the factory method itself to eliminate the need for such documentation. As a stop-gap, I added this table (with your help) to chapter 2 of our upcoming tutorial:

Screenshot 2025-06-10 at 12 38 41 AM

It looks like I need to make the following modifications:
- `flux jobs` is a typo
- Everything in the table _except_ arbitrary attributes and shell options can now be set in the factory method.

Is that correct?

grondo:
> 
>   flux jobs is a typo

Yeah, this should be `jobspec.stdout = <path>` or `jobspec.stderr = <path>` (also `--input` is `jobspec.stdin = <path>`.

I'm about to open a PR  to make `jobspec.output`, `jobspec.error` and `jobspec.input` aliases for these.

>   Everything in the table except arbitrary attributes and shell options can now be set in the factory method.
>  Is that correct?

Yes, though none of this work has been merged into a flux-core release, so you might just mention "More parameters will be added to the jobspec initializers in a future Flux release.

grondo:
Once #6874 #6875 and #6877 are merged, I'll rebase this to pick up those changes (use unbuffered property, and rename `stdout`, `stderr`, and `stdin` initializer args to `output`, `error`, and `input`.

grondo:
Ok, this has been rebased. The `stderr`, `stdout`, and `stdin` parameters have been renamed `error`, `output` and `input`, and internally the `unbuffered` arg uses `jobspec.unbuffered`.

This could probably use a final ACK due to all the changes.

grondo:
Thanks @cmoussa1!

I'll go ahead and set MWP here. Perhaps @trws, @jameshcorbett and/or @wihobbs can come back around and double check the final result here when they have time and we can fix any issues they notice before the next release (when we'll have more difficulty making changes)

src/bindings/python/flux/job/Jobspec.py
```diff
@@ -1001,7 +1099,21 @@ def from_command(
...
         tasks = [{"command": command, "slot": "task", "count": task_count_dict}]
         attributes = {"system": {"duration": 0}}
-        return cls(resources, tasks, attributes=attributes)
+        return cls(resources, tasks, attributes=attributes)._set_extra_options(
+            duration=duration,
+            environment=environment,
+            env_expand=env_expand,
+            cwd=cwd,
+            rlimits=rlimits,
+            name=name,
```
wihobbs, line 1108: Why `name` and not `jobname` or `job_name`?

src/bindings/python/flux/job/Jobspec.py
```diff
@@ -822,6 +822,57 @@ def _v1_validate(resources, tasks, kwargs):
             if task["count"].keys().isdisjoint({"per_slot", "total"}):
                 raise ValueError("count per_slot or total must be set")
 
+    def _set_extra_options(
+        self,
+        duration=None,
+        environment=None,
```
wihobbs, line 828: I really do wonder if we should consider setting `environment=dict(os.environ)` by default. There are good reasons not to, but since it's something we _consistently_ end up explaining to users, I wanted to start the discussion.

wihobbs: Or maybe a convenience option to the `.from_command` method, like `add_env=True`

wihobbs (commented):
I had a couple clarifications and really nitty comments but this is a huge improvement. Thank you @grondo!

src/bindings/python/flux/job/Jobspec.py
grondo: It seems more consistent: the other args are not `job_cwd`, `job_rlimits`, `job_environment` and so on, so why tack `job_` in front of the name?

src/bindings/python/flux/job/Jobspec.py
grondo: This is a good suggestion. We should open an issue on having the default copy the current environment.  I'm not sure of the good reasons not to except that we haven't before.

Note that using `environment=os.environ` now would be the same as `add_env=True`.

trws: At least on the `from_command` or things where we're setting sane defaults I think it make sense yeah. It's a bit out of the ordinary to capture the environment in an API in general, but thinking about things like `subprocess` that's what they do and that's not entirely unlike what we're doing here.

grondo: Thanks @wihobbs and @trws. I've opened an issue for this suggestion.

src/bindings/python/flux/job/Jobspec.py
wihobbs: In the case of (1) and (3) the command line has `--cwd` and `--env` while the command line option to name a job is `--job-name`. My first instinct would have been to try and match the command line, but I don't think it really matters all that much.

src/bindings/python/flux/job/Jobspec.py
wihobbs: Thanks @grondo!

src/bindings/python/flux/job/Jobspec.py
grondo: The ship for matching cmdline options and these arguments sailed when the resource count options were called `num_tasks` and `num_nodes` :shrug: (I will say the `--job-name` long option does irk me a bit too, so thought I'd fix that here)

Issue #6860, opened 2025-06-03
not ok 17 - flux-fsck detects errors


grondo:
This test has been failing somewhat regularly in CI:

```
  expecting success: 
  	test_must_fail flux fsck 2> fsckerrors1.out &&
  	count=$(cat fsckerrors1.out | wc -l) &&
  	test $count -eq 3 &&
  	grep "dir\.b" fsckerrors1.out | grep "missing blobref(s)" &&
  	grep "Total errors: 1" fsckerrors1.out
  
  not ok 17 - flux-fsck detects errors
```

(happened in two different builders in the news-0.75 PR)

Unfortunately there isn't much debug here to go on, perhaps an addition of:

```
    test_debug "cat fsckerrors.1.out" &&
```

to the test above (and other tests in this testsuite) would be useful.

chu11:
i hit this once and thought it was a `flux kvs sync` issue.  Lemme add the debugging.

sam-maloney:
I'm also getting intermittent failures in this test, for both 17 and 24. The output from the tests appears identical whether it reports failure or success:
```
FAIL: t2816-fsck-cmd.t 17 - flux-fsck detects errors
	dir.b: missing blobref index=1
	Total errors: 1
FAIL: t2816-fsck-cmd.t 24 - flux-fsck detects errors
	dir.b: missing blobref index=1
	dir.c: missing blobref index=1
	dir.c: missing blobref index=2
	Total errors: 2
```

I ran the test with `-d -v` flags, and for test 17, the contents of `fsckerrors1.out` in `trash-directory.t2816-fsck-cmd` are:
```
Checking integrity of checkpoint from 2025-06-05T06:24:51
dir.b: error retrieving blobref(s): No such file or directory
Total errors: 1
```
while the contents of `fsckerrors1V.out` are:
```
Checking integrity of checkpoint from 2025-06-05T06:24:51
alink
dir.a
dir.b
dir.b: missing blobref index=1
dir.c
Total errors: 1
```
So clearly something failed in the initial retrieval but succeeded in the "verbose" follow-up; maybe @chu11 is on the right track with something not being synced yet in the kvs?

For test 24 it's similar; when it fails the contents of `fsckerrors3.out` are:
```
Checking integrity of checkpoint from 2025-06-05T07:14:24
dir.b: error retrieving blobref(s): No such file or directory
dir.c: missing blobref(s)
Total errors: 2
```

chu11:
[...]

PR #6861, opened 2025-06-04
t: add fsck test debug

chu11:
Problem: Several fsck tests have failed in CI without much debugging to go on.

Add some debugging in t2816-fsck-cmd.t.


chu11:
closing, per comment in #6860, we think we found the issue.  Will propose an alternate PR with fix and this debug too.

PR #6863, opened 2025-06-04
content: remove checkpoint key 

chu11:
Problem: In the content backing modules, checkpoints are allowed to take a key.  However, the singular "kvs-primary" key is the
only key used.  Only testing uses alternate keys.  The flexible key makes some future modifications to checkpointing more difficult.

Make "kvs-primary" the only valid key for checkpointing.

### DIFF
--- a/src/modules/content-files/content-files.c
+++ b/src/modules/content-files/content-files.c
@@ -59,6 +59,7 @@
 #include "src/common/libutil/log.h"
 #include "src/common/libutil/dirwalk.h"
 #include "src/common/libutil/unlink_recursive.h"
+#include "src/common/libkvs/kvs_checkpoint.h"
 #include "ccan/str/str.h"
 
 #include "src/common/libcontent/content-util.h"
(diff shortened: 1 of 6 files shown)


chu11:
doh!  I just remembered that the `content` module forwards checkpt get/put requests to the backing modules, so that should also check for `kvs-primary`.  Lemme re-work this PR a bit.

chu11:
Ok, I've removed WIP.  After going back and forth, this PR is largely unchanged from yesterday.  I'm forking off working onto a follow up PR.

garlick:
Any reason not to drop the key completely from everywhere but the table (so we don't have to change the schema yet)?

chu11:
> Any reason not to drop the key completely from everywhere but the table (so we don't have to change the schema yet)?

Not sure if you saw the follow on PR to this (#6869) where I make the key optional everywhere.  So it's no longer used anywhere, but we still parse the key for backwards compatibility.

Or are you saying we should remove the key from the RPC in it's entirety and screw backwards compatibility?

garlick:
I think it's only expected to be used from rank 0 of flux-core, and we currently don't support mixed versions within an instance anyway, so the RPC seems like fair game to change?

chu11:
> I think it's only expected to be used from rank 0 of flux-core, and we currently don't support mixed versions within an instance anyway, so the RPC seems like fair game to change?

Ok, lemme give it a shot.  I can close #6869 b/c if we're axeing the key entirely, no need for that PR.

chu11:
re-pushed, going with the far simpler approach of removing the "key" from all checkpoint RPCs.  PR title updated as a result.

PR #6867, opened 2025-06-05
flux-fsck: fix error message output

### LINKED ISSUE #6860: not ok 17 - flux-fsck detects errors

chu11:
Problem: An error check inadvertently checks `errno` when it should check the stored errnum in a data structure.  This leads to an invalid error message output.

Fix the error check to ensure the correct error message is output.

Fixes #6860

### DIFF
--- a/src/cmd/builtin/fsck.c
+++ b/src/cmd/builtin/fsck.c
@@ -147,7 +147,7 @@ static void fsck_valref (flux_t *h, const char *path, json_t *treeobj)
     if (fvd.errorcount) {
         /* each invalid blobref will be output in verbose mode */
         if (!verbose) {
-            if (errno == ENOENT)
+            if (fvd.errnum == ENOENT)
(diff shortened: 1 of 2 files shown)


grondo (approved):
Good catch!

PR #6868, opened 2025-06-05
testsuite: reduce parallelism in `t4000-issues-test-driver.t` for test reliability

### LINKED ISSUE #6856: Some test failures on some systems, cause unknown
Some of our testing for EasyBuild has seen other test failures (see https://github.com/easybuilders/easybuild-easyconfigs/pull/22979#issuecomment-2930025650):
```
FAIL: t4000-issues-test-driver.t 13 - t3960-job-exec-ehostunreach
FAIL: t4000-issues-test-driver.t 22 - t2686-shell-input-race
FAIL: t4000-issues-test-driver.t 28 - t2492-shell-lost
ERROR: t4000-issues-test-driver.t - exited with status 1
```
Without direct access I can't reproduce these to inspect further, so I will probably skip that set of tests in my EasyBuild recipe.

grondo:
Problem: The issues tests in t4000-issues-test-driver.t are run as jobs in a test instance of size=2, but the instance inherits the total number of cores from the system, which means up to 2*ncores issues tests could be run at the same time. On some systems, this causes failures in the tests due to overloading the system.

Switch to the "job" personality of test_under_flux, which will configure the test instance with only 2 fake cores per rank, keeping the parallelism of the tests to 4, which is probably a safe number.

Fixes #6856

### DIFF
--- a/t/t4000-issues-test-driver.t
+++ b/t/t4000-issues-test-driver.t
@@ -10,8 +10,13 @@ if test_have_prereq ASAN; then
     test_done
 fi
 
+# Note: use test_under_flux "job" personality so that only 2 cores per fake
+# node are configured in the test instance. This ensures a maximum of 4
+# issues test scripts (invoked as jobs below) run simultaneously, instead
+# of 1 per real core, which could cause test failures on overloaded systems.
+#
 SIZE=2
-test_under_flux ${SIZE}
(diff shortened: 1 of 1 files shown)


grondo:
Thanks! Setting MWP.

PR #6869, opened 2025-06-05
content: make checkpoint key optional

chu11:
This PR follows #6863.  I had them in one PR at first, but since this touches the content module (other one only touches backend modules), I decided to fork it off into it's own PR.

----

Problem: The only checkpoint key that is allowed is now "kvs-primary".  That means the checkpoint key is no longer relevant.

In content module and backing modules, make the checkpoint key optional. If it is not specified, assume "kvs-primary".


chu11:
closing, per comments in #6863, going in a slightly different direction, just pulling the whole bandaid off

PR #6870, opened 2025-06-05
Python: Jobspec repr change

jameshcorbett:


src/bindings/python/flux/job/Jobspec.py
```diff
@@ -682,6 +682,13 @@ def resource_counts(self):
         return count_dict
 
     def __repr__(self):
+        return (
+            f"{type(self).__name__}({self.jobspec['resources']}, "
+            f"{self.jobspec['tasks']}, attributes={self.jobspec['attributes']}, "
+            f"version={self.jobspec['version']})"
+        )
```
grondo: Great! I may be mistaken, but I think the jobspec classes have `resources` `tasks` `attributes` and `version` properties, so this could be shortened to use those, e.g. `self.resources` instead of `self.jobspec['resources']`.

jameshcorbett: Good catch, I force-pushed to make that change.

grondo: Great, already approved so feel free to set MWP when you're ready. Thanks!

jameshcorbett: I'll wait for @wihobbs' approval.

t/python/t0010-job.py
```diff
@@ -821,6 +821,11 @@ def test_36_repr(self):
             "{'resources': [{'type': 'slot', 'count': 1, 'label': 'task', 'with': [{'type': 'core', 'count': 1}]}], 'tasks': [{'command': ['app'], 'slot': 'foo', 'count': {'per_slot': 1}}], 'attributes': {'system': {'duration': 0, 'cow': 1, 'cat': 2, 'chicken': 4}, 'user': {'dog': 3, 'duck': 5}, 'goat': 6}, 'version': 1}",
         )
 
+    def test_37_repr(self):
+        """Test __repr__ method of Jobspec"""
+        jobspec = Jobspec.from_yaml_stream(self.basic_jobspec)
+        self.assertEqual(eval(repr(jobspec)).jobspec, jobspec.jobspec)
```
wihobbs, line 827: Is there a circumstance in which this test will fail? I think if it were me I'd evaluate to some static string like

```diff
diff --git a/t/python/t0010-job.py b/t/python/t0010-job.py
index 273c91700..29709c570 100755
--- a/t/python/t0010-job.py
+++ b/t/python/t0010-job.py
@@ -824,7 +824,7 @@ class TestJob(unittest.TestCase):
     def test_37_repr(self):
         """Test __repr__ method of Jobspec"""
         jobspec = Jobspec.from_yaml_stream(self.basic_jobspec)
-        self.assertEqual(eval(repr(jobspec)).jobspec, jobspec.jobspec)
+        self.assertEqual("Jobspec([{'type': 'slot', 'count': 1, 'label': 'task', 'with': [{'type': 'core', 'count': 1}]}], [{'command': ['app'], 'slot': 'foo', 'count': {'per_slot': 1}}], attri
butes={'system': {'duration': 0}}, version=1)", repr(jobspec))
 
 
 if __name__ == "__main__":

```

but feel free to educate me @jameshcorbett 👑 🐍

wihobbs (approved):
Thanks for cleaning this up :)

t/python/t0010-job.py
jameshcorbett: I don't think so, if the test fails I would think that meant the `__repr__` method was incorrect. With `eval(repr(jobspec)` the test should succeed even if `jobspec` is changed.

jameshcorbett:
Force-pushed to rebase with master and expand the test a little bit, setting MWP.

PR #6874, opened 2025-06-09
python: unify behavior of `Jobspec` `getattr` and `setattr` methods

grondo:
This PR makes a tweak to the behavior of the `Jobspec.getattr` method so that it is consistent with the behavior of `Jobspec.setattr`. That is,  keys that do not start with `user.`, `system.` or `attributes.` default to  `attributes.system.`

### DIFF
--- a/src/bindings/python/flux/job/Jobspec.py
+++ b/src/bindings/python/flux/job/Jobspec.py
@@ -132,6 +132,19 @@ def validate_jobspec(jobspec, require_version=None):
     return True
 
 
+def _attr_key_prepend(key):
+    """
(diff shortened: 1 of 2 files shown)


chu11 (approved):
This LGTM so approving.  But I recall we intentionally decided not to do this before.

https://github.com/flux-framework/flux-core/pull/6053#discussion_r1653423659

So I assume there is a reason to do this now?  (It's possible your follow on PR may answer this question)

Edit: halfway through follow on PR, I now see why change made.

grondo:
Yeah, sorry, I had forgotten we'd discussed this before. :facepalm:

PR #6875, opened 2025-06-09
python: add `unbuffered` getter and setter to `Jobspec` class

grondo:
This PR adds an `unbuffered` getter and setter property to the Python `Jobspec` class as suggested by @trws in #6858.

based on top of #6874.

### DIFF
--- a/t/python/t0010-job.py
+++ b/t/python/t0010-job.py
@@ -401,6 +401,29 @@ class TestJob(unittest.TestCase):
             with self.assertRaises(TypeError):
                 setattr(jobspec, stream, None)
 
+        with self.assertRaises(TypeError):
+            jobspec.unbuffered = 1
+
+        jobspec.unbuffered = True
+        self.assertTrue(jobspec.unbuffered)
+        self.assertEqual(
+            jobspec.getattr("shell.options.output.stderr.buffer.type"), "none"
+        )
+        self.assertEqual(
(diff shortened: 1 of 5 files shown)


t/python/t0010-job.py
```diff
@@ -384,6 +384,23 @@ def test_21_stdio(self):
             with self.assertRaises(TypeError):
                 setattr(jobspec, stream, None)
 
+        with self.assertRaises(TypeError):
+            jobspec.unbuffered = 1
+
+        jobspec.unbuffered = True
+        self.assertTrue(jobspec.unbuffered)
+        self.assertEqual(jobspec.getattr("shell.options.output.batch-timeout"), 0.05)
```
chu11: don't want to test buffer.type == none?

chu11 (approved):
LGTM, just one nit below

t/python/t0010-job.py
grondo: implied by `jobspec.unbufferred == True`? Could add it just for kicks though.

chu11: oh that's right!  I saw the `batch-timeout` check and thought, should check the other ones too?

grondo: I think your suggestion was a good one, so I added that check. Then the operation of the getter is validated as well

grondo:
Thanks! Rebased and I'll set MWP.

PR #6879, opened 2025-06-16
broker: let a upmi client relay extra bootstrap info

ggouaillardet:
Problem:  there is no way for a upmi client to relay auxiliary bootstrap
information through the broker to applications.

Add an optional dict plugin OUT argument that can be set by the
upmi_initialize() implementation.  A new dict member is added to
struct pmi_info that is set to a json dictionary when the implementation
sets  the OUT argument.  In the broker, any key-value pairs in the dict are
directly set as broker attributes.

### DIFF
--- a/src/broker/boot_pmi.c
+++ b/src/broker/boot_pmi.c
@@ -426,6 +426,8 @@ int boot_pmi (const char *hostname, struct overlay *overlay, attr_t *attrs)
     const struct bizcard *bc;
     struct bizcache *cache = NULL;
     struct taskmap *taskmap = NULL;
+    const char *dkey;
+    json_t *value;
 
     // N.B. overlay_create() sets the tbon.topo attribute
     if (attr_get (attrs, "tbon.topo", &topo_uri, NULL) < 0) {
@@ -448,6 +450,18 @@ int boot_pmi (const char *hostname, struct overlay *overlay, attr_t *attrs)
         log_msg ("%s: initialize: %s", upmi_describe (upmi), error.text);
         upmi_destroy (upmi);
         return -1;
+    }
+    if (info.dict != NULL) {
+        json_object_foreach(info.dict, dkey, value) {
+            if (!json_is_string(value)) {
+                log_err ("%s: initialize: value associated to key %s is not a string", upmi_describe (upmi), dkey);
+                goto error;
+            }
+            if (attr_add (attrs, dkey, json_string_value(value), ATTR_IMMUTABLE) < 0) {
+                log_err("%s: initialize: could not put attribute for key %s", upmi_describe (upmi), dkey);
+                goto error;
+            }
+        }
     }
     if (set_instance_level_attr (upmi, attrs, &under_flux) < 0) {
         log_err ("set_instance_level_attr");
--- a/src/common/libpmi/upmi.h
+++ b/src/common/libpmi/upmi.h
@@ -14,6 +14,7 @@
 #include <sys/types.h>
 #include <stdarg.h>
 #include <stdbool.h>
+#include <jansson.h>
 
 #include "src/common/libflux/types.h"
 
@@ -27,6 +28,7 @@ struct upmi_info {
     int rank;
     int size;
     const char *name;
+    json_t *dict; // may be NULL and is invalidated by the next upmi API call
 };
 
 typedef void (*upmi_trace_f)(void *arg, const char *text);
--- a/src/common/libpmi/test/upmi.c
+++ b/src/common/libpmi/test/upmi.c
@@ -13,6 +13,7 @@
 #endif
 #include <stdlib.h>
 #include <string.h>
+#include <jansson.h>
 
 #include "src/common/libtap/tap.h"
 #include "ccan/str/str.h"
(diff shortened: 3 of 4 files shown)


ggouaillardet:
@trws @milroy This is the PR related to the great discussion we had at ISC'25.

In a nutshell, I am bootstrapping a `flux` session from a `PMIx` capable batch manager. For some site specific reasons, I have to retrieve a `PMIx` key from the batch manager, then this key has to be tailored for a given flux job. Most of this is very site specific so I simply forked `flux-pmix`, but I needed some relatively simple changes in `flux-core` in order to make it happen.

FWIW, here is the patch in `flux-pmix` that retrieves the `foobar` key from the batch manager at bootstrap time so it can be made available to the flux jobs
```patch
diff --git a/src/upmi/pmix.c b/src/upmi/pmix.c
index 2a0d898..9173cf7 100644
--- a/src/upmi/pmix.c
+++ b/src/upmi/pmix.c
@@ -18,6 +18,7 @@
 #include "config.h"
 #endif
 #include <stdarg.h>
+#include <sodium.h>
 #include <pmix.h>
 #include <flux/core.h>
 
@@ -164,6 +165,11 @@ static int op_initialize (flux_plugin_t *p,
     pmix_status_t rc;
     pmix_proc_t proc;
     pmix_value_t *val;
+    uint32_t jobsize;
+    const unsigned char *raw_rankmap;
+    char * enc_rankmap;
+    size_t raw_size, enc_size;
+    int ret;
 
     rc = PMIx_Init (&myproc, NULL, 0);
     if (rc != PMIX_SUCCESS)
@@ -179,17 +185,40 @@ static int op_initialize (flux_plugin_t *p,
         PMIX_VALUE_RELEASE (val);
         return seterror (p, args, "PMIx_Get size: wrong type");
     }
+    jobsize = val->data.uint32;
+    PMIX_VALUE_RELEASE (val);
 
-    if (flux_plugin_arg_pack (args,
-                              FLUX_PLUGIN_ARG_OUT,
-                              "{s:i s:s s:i}",
-                              "rank", myproc.rank,
-                              "name", myproc.nspace,
-                              "size", val->data.uint32) < 0) {
+    rc = PMIx_Get (&proc, "foobar", NULL, 0, &val);
+    if (rc != PMIX_SUCCESS)
+        return seterror (p, args, "PMIx_Get foobar: %s", PMIx_Error_string (rc));
+    if (val->type != PMIX_BYTE_OBJECT) {
         PMIX_VALUE_RELEASE (val);
+        return seterror (p, args, "PMIx_Get size: wrong type");
+    }
+    raw_rankmap = (const unsigned char *)val->data.bo.bytes;
+    raw_size = val->data.bo.size;
+    enc_size = sodium_base64_encoded_len (raw_size, sodium_base64_VARIANT_ORIGINAL);
+    if (NULL == (enc_rankmap = malloc (enc_size))) {
+        return seterror (p, args, "could not allocate %ld bytes", enc_size);
+        PMIX_VALUE_RELEASE(val);
+    }
+    sodium_bin2base64 (enc_rankmap, enc_size,
+                       raw_rankmap, raw_size,
+                       sodium_base64_VARIANT_ORIGINAL);
+
+    PMIX_VALUE_RELEASE(val);
+
+    ret = flux_plugin_arg_pack (args,
+                                FLUX_PLUGIN_ARG_OUT,
+                                "{s:i s:s s:i s:{s:s}}",
+                                "rank", myproc.rank,
+                                "name", myproc.nspace,
+                                "size", jobsize,
+                                "dict", "foobar", enc_rankmap);
+    free(enc_rankmap);
+    if (ret < 0) {
         return -1;
     }
-    PMIX_VALUE_RELEASE (val);
     return 0;
 }
 ```

File: src/common/libpmi/upmi.h [hunk 1]
```diff
@@ -27,6 +28,7 @@ struct upmi_info {
     int rank;
     int size;
     const char *name;
+    json_t *dict;
```
garlick: An inline comment may be helpful here:
```c
// may be NULL and is invalidated by the next upmi API call
```

src/broker/boot_pmi.c
```diff
@@ -449,6 +451,16 @@ int boot_pmi (const char *hostname, struct overlay *overlay, attr_t *attrs)
...
+    json_object_foreach(info.dict, dkey, value) {
+        if (!json_is_string(value)) {
+            log_err ("%s: initialize: value associated to key %s is not a string", upmi_describe (upmi), dkey);
+            goto error;
+        }
+        if (attr_add (attrs, dkey, json_string_value(value), ATTR_IMMUTABLE) < 0) {
+            log_err("%s: initialize: could not put attribute for key %s", upmi_describe (upmi), dkey);
+            goto error;
+        }
+    }
```
garlick: `info.dict` can be NULL here I think, so maybe it would be good to make this block conditional on non-NULL?

(Possilby `json_object_foreach()` handles a NULL object, but the API docs don't directly state this AFAICT)

garlick (approved):
Thanks!  Looks good.

ggouaillardet:
Thanks Jim for the guidance!

I will cleanup my `flux-pmix` changes and issue a PR tomorrow so you can see what I am doing.

Issue #6872, opened 2025-06-06
kvs-watch/job-info: support a sentinel value indicating "end of initial values"


chu11:
It would be convenient if kvs-watch could inform watchers that all data that is presently available (i.e. data may appear in the future but we don't know) has already been streamed and sent.

This could be useful for #5035.  We could cache the last N lines of data until we reach the current "end of initial value" and then could output those N lines via a "tail" option.  Because we don't have a sentinel or flag indicating "we're at the end of the initial value", we can't implement this algorithm.

How could this be done?

- new rpc parameter on responses indicating this condition.  
   - Pro: makes the most sense, a `flux_kvs_get_sentinel()` function could be used to retrieve the flag
   - Con: sometimes multiple events are returned in "one blob", so gotta calculate "is this the last eventlog entry in this blob" (or count entries before we parse) in `job-info` module
- some "sentinel" RPC, like maybe an empty value or something, works sort of like with previous option with some sort of "is a sentinel" function.
  - Pro: easier than above b/c I don't have to count all the eventlog entries, just "forward" it in the `job-info` module.
  - Con: sort of breaks the idea of "streaming events" 
- errno other than ENODATA could indicate this condition.
  - Pro: no need for new helper functions
  - Con: This technique is a bit dangerous b/c it could break most existing code, so the errno would only be sent if the behavior is requested by the caller

Edit: Doh, option 1 above won't work b/c we need something for "no initial value".  I'm thinking option 2 is best now.

garlick:
Option 3 wouldn't work since any errror response terminates the stream per RFC 6 (and this is relied on for RPC tracking in the broker)

garlick:
Since a sentinel value presumably would require a flag in the watch request anyway (since it would potentially break other users if not), maybe a better flag would be one to request that the watch start at the last blob, or wait for the next one?

chu11:
> maybe a better flag would be one to request that the watch start at the last blob, or wait for the next one?

This is similar to the idea in #5035 to simply ignore all events with timestamps less than "now", i.e. only output new events / data.  But it ended up that we do need to "tail" some of the last bits of data before watching "newer events".

We have no way right now to know when the "end" of the stream of events is, thus the idea of this sentinel value.

grondo:
The two use cases I was thinking of for this feature would still require getting all history first, so ignoring some portion of events already in the eventlog wouldn't work. (Unless I'm missing the intent of the comment)

[...]

PR #6880, opened 2025-06-17
kvs-watch/job-info: support initial sentinel

chu11:
Per discussion in #6872, support the idea of an "Initial Sentinel", that can inform KVS watchers that we are done reading all of the "current data" that exists in the KVS.

I went back and forth on how to do things:

- settled on sending an empty message to indicate sentinel vs a parameter in an RPC.  Since we need to send a sentinel when a value is "empty", it sort of eliminated the parameter idea b/c there might not be any data to send.

- I went back and forth on how to handle this in terms of the API and eventually settled on adding the following two functions:

`flux_kvs_lookup_get_initial_sentinel(flux_future_t *f, const char **value, bool *initial_sentinel)`
`flux_job_event_watch_get_initial_sentinel(flux_future_t *f, const char **event, bool *initial_sentinel)`

they are identical to the get equivalents (i.e. `flux_kvs_lookup_get()` and `flux_job_event_watch_get()`) but set the sentinel flag if the sentinel was read.

I originally tried to have a separate function (i.e. `flux_kvs_lookup_sentinel()`), but that added an extra function call to be called, which I disliked.  Also, b/c the RPC returns an empty payload for the sentinel, that messes up the current API that assumes a payload must be formatted a specific way (i.e. we get EPROTO errors), so making me lean to the functions I added above.  I don't love this design, but it might be the least bad???  Welcome to other API ideas.  (as an aside, I now dislike calling this "initial sentinel", probably should come up with a better name ... thinking of options).

The last commit here is simply a dumb "would --tail work in flux job attach".  Largely can ignore it for now.  That work will take some refactoring b/c of all the output options that exist and will be spliced out later.


garlick:
One thought is just to document that a zero length value is returned when the watch stream is up to date, and then let people use `flux_kvs_lookup_get_raw()` to check the length, or heck just add:
```c
ssize_t flux_kvs_lookup_get_size (flux_future_t *f);
```
I'm assuming the sentinel only applies when FLUX_KVS_WATCH_APPEND is used, since by definition you're up to date after the first response for the other modes I think.

Edit: or maybe `flux_kvs_lookup_get()` and   `flux_job_event_watch_get()` could set the  OUT parameter to NULL when the sentinel is received?

chu11:
> I'm assuming the sentinel only applies when FLUX_KVS_WATCH_APPEND is used, since by definition you're up to date after the first response for the other modes I think.

That's true, I did support it with just FLUX_KVS_WATCH, but perhaps that was not necessary.

> Edit: or maybe flux_kvs_lookup_get() and flux_job_event_watch_get() could set the OUT parameter to NULL when the sentinel is received?

I didn't do that for `flux_kvs_lookup_get()` b/c an empty value for a key is reasonable.  But A) I should double check if the value returned is NULL or empty string.  and B) perhaps that's not as big a deal as I originally thought it would be.  If you're watching something, you always expect there to be some value and you can treat "no value" as the sentinel.

I did the `kvs-watch` first and just copied it over to `job-info`.  It does make sense for `flux_job_event_watch_get()` since an event is the reasonable thing to always expect from that API.

chu11:
Just a note to myself for later, the failing test is b/c the test is a little racy.

```
  {"timestamp":1750199581.9394929,"name":"init"}
  {"timestamp":1750199581.9406893,"name":"starting"}
  {"timestamp":1750199581.9571524,"name":"shell.init","context":{"service":"1001-shell-f4qLBJAK","leader-rank":0,"size":1}}
  sentinel
  {"timestamp":1750199581.9598074,"name":"shell.start","context":{"taskmap":{"version":1,"map":[[0,1,1,1]]}}}
```

can't assume that the job has hit `shell.start` when launching the watcher.

chu11:
> Edit: or maybe flux_kvs_lookup_get() and flux_job_event_watch_get() could set the OUT parameter to NULL when the sentinel is received?

Now I remember, there were two reasons I didn't do this.

A) an "empty value" is a legitimate return value for `flux_kvs_lookup_get()` when the value for the KVS is empty.  But perhaps we can just ignore that.  it could be "empty value " or sentinel.  Or we could just put some "is_sentinel" parameter into the response RPC.  Lemme mull that a bit more.

B) we'd presumably want to support the the NULL OUT parameter in all of these functions.

```
int flux_kvs_lookup_get (flux_future_t *f, const char **value);
int flux_kvs_lookup_get_unpack (flux_future_t *f, const char *fmt, ...);
int flux_kvs_lookup_get_raw (flux_future_t *f, const void **data, size_t *len);
int flux_kvs_lookup_get_treeobj (flux_future_t *f, const char **treeobj);
int flux_kvs_lookup_get_dir (flux_future_t *f, const flux_kvsdir_t **dir);
int flux_kvs_lookup_get_symlink (flux_future_t *f,
                                 const char **ns,
                                 const char **target);
```

Which is plausible for all except for `flux_kvs_lookup_get_unpack()`.

> flux_kvs_lookup_get_raw() to check the length

I like your idea of using `flux_kvs_lookup_get_raw()`, it's not so different than my `_initial_sentinel()` idea but we're just using the "len" from the `raw` function instead.

This is how I began doing things at first

Code that was

```
if (flux_kvs_lookup_get (...) < 0)
   error handling
do good stuff
```

turned into

```
if (flux_rpc_get (...) < 0)
   error handling
if (value != NULL) {
   if (flux_kvs_lookup_get (...) < 0)
      error handling
   do good stuff
}
else {
    sentinel path
}
```

OR

```
if (flux_kvs_lookup_get (...) < 0) {
     if (errno == EPROTO) {
          if (flux_rpc_get (...) < 0)
              more error handling
          if (value == NULL) {
              handle sentinel
          }
     }
     error handling normal
}
do good stuff
```

Which I disliked both.

chu11:
I've re-worked the PR

A) The sentinel only works with FLUX_KVS_WATCH and FLUX_KVS_WATCH_APPEND.  So it doesn't work with just watching a value normally.

B) Now going with the following functions:

```
int flux_kvs_lookup_get (flux_future_t *f, const char **value);
int flux_kvs_lookup_get_raw (flux_future_t *f, const void **data, size_t *len);
int flux_kvs_lookup_get_treeobj (flux_future_t *f, const char **treeobj);
```

and

```
int flux_job_event_watch_get (flux_future_t *f, const char **event);
```

returning NULL for their OUT parameters when the sentinel has been received (FYI, `lookup_get_unpack()` doesn't work for this style.  `lookup_get_dir()` and `lookup_get_symlink()` dont' make much sense with `FLUX_KVS_WATCH_APPEND` and this flag).

Pro: no new extra functions, the API is pretty simple and easy

Con: the primary corner case (which lead me to shy away from this before) is an "empty value" is a legitimate KVS value and that returns "NULL" in `flux_kvs_lookup_get()`.   So having this API style with "sentinels" is mostly b/c users will not use this with any key they believe an "empty value" is feasible.  Extra side note, it's really only feasible if a key begins as an empty value, an "empty value" can't be appended to an existing key.

chu11:
just pushed a commit on top that implements `--tail=N` in `flux job attach`.  Everything appears to work as expected.

Lemme work on some tests, add appropriate documentation, and split this up into several PRs.

chu11:
removing WIP, I think this is good for a review.  Added docs and some tests.  A few minor changes from prior push.

A) only these kvs functions return NULL if we reach the sentinel w/ kvs-watch

```
int flux_kvs_lookup_get (flux_future_t *f, const char **value);
int flux_kvs_lookup_get_raw (flux_future_t *f, const void **data, size_t *len);
```

B) the flag is now called `FLUX_KVS_WATCH_APPEND_INITIAL_SENTINEL`

src/common/libkvs/kvs.h
```diff
@@ -36,7 +36,8 @@ enum kvs_op {
     FLUX_KVS_WATCH_FULL = 64,
```
garlick, line 36: Commit message: instead of "an empty RPC will be sent", I think you mean an empty *response* will be sent.

src/common/libkvs/kvs.h
```diff
@@ -36,7 +36,8 @@ enum kvs_op {
...
     FLUX_KVS_WATCH_UNIQ = 128,
     FLUX_KVS_WATCH_APPEND = 256,
-    FLUX_KVS_STREAM = 512
+    FLUX_KVS_STREAM = 512,
+    FLUX_KVS_WATCH_APPEND_INITIAL_SENTINEL = 1024
```
garlick: This flag name is long and kind of confusing when taken out of context of the use case (does it append a sentinel?)  Maybe something short and use case focused like FLUX_KVS_WATCH_TAIL would be better?

src/common/libkvs/kvs_lookup.c
```diff
@@ -237,12 +239,30 @@ static struct lookup_ctx *get_lookup_ctx (flux_future_t *f)
...
+            && sentinel) {
+            const char *s;
+            if (flux_rpc_get (f, &s) < 0) {
+                errno = save_errno;
+                return -1;
+            }
+            if (!s) {
+                (*sentinel) = true;
+                return 0;
+            }
```
garlick: It seems like this code is left over from the earlier design when there was a special sentinel in the API?
Now couldn't we simplify this and just return a NULL payload all the way through?

src/modules/kvs-watch/kvs-watch.c
```diff
@@ -262,6 +263,19 @@ static void watcher_cleanup (struct ns_monitor *nsm, struct watcher *w)
         zhashx_delete (nsm->ctx->namespaces, nsm->ns_name);
 }
 
+static void send_initial_sentinel (flux_t *h, struct watcher *w)
+{
+    if (w->flags & FLUX_KVS_WATCH
+        && w->flags & FLUX_KVS_WATCH_APPEND
+        && w->flags & FLUX_KVS_WATCH_APPEND_INITIAL_SENTINEL
+        && w->initial_sentinel_sent == false) {
```
garlick: Shouldn't flags be checked for compatibility when the watch request is initially accepted, so that we only have to check for the one flag here?

src/modules/kvs-watch/kvs-watch.c
```diff
@@ -289,6 +303,9 @@ static void handle_load_response (flux_future_t *f, struct watcher *w)
         }
         w->loaded_blob_count++;
         w->responded = true;
+
+        if (flux_future_aux_get (f, "initial_sentinel"))
+            send_initial_sentinel (h, w);
```
garlick, line 307: Why is there an aux entry for the sentinel?  It seems like we have access to the `w` struct that includes the flags everywhere this is used.  I'm probably missing something.

garlick (commented):
A few questions about the kvs-watch implementation.  Seems like it might need to be reworked based on the design changes?

src/common/libkvs/kvs.h
chu11: yeah, it's long.  TAIL to me doesn't seem right b/c that usually signifies the data "after" or "later'.    I prefixed the flag with "FLUX_KVS_WATCH_APPEND" b/c it only works with that flag.  I suppose that isn't entirely necessarily.  I could shorten it to just "FLUX_KVS_WATCH_SENTINEL" and documentation would indicate it only works with APPEND.

src/common/libkvs/kvs_lookup.c
chu11: so the sentinel flag is mostly b/c code in `lookup_get()` and some other family of functions is like this:

```
    if (parse_response (f, ctx, &sentinel) < 0)                                                                                             
        return -1;                                                                                                                          
    if (sentinel) {                                                                                                                         
        if (value)                                                                                                                          
            *value = NULL;                                                                                                                  
        return 0;                                                                                                                           
    }                                                                                                                                       
    if (!ctx->val_valid) {                                                                                                                  
        if (treeobj_decode_val (ctx->treeobj,                                                                                               
                                &ctx->val_data,                                                                                             
                                &ctx->val_len) < 0)                                                                                         
            return -1;                                                                                                                      
        ctx->val_valid = true;                                                                                                              
        // N.B. val_data includes xtra 0 byte term not reflected in val_len                                                                 
    }                                  
```

There are some assumptions later on about what values are set, like `ctx->treeobj`.  I suppose alternately code could be updated for "if treeobj NULL then RPC has no payload".

src/modules/kvs-watch/kvs-watch.c
chu11: it seems we don't (only w/n the kvs-watch API).  perhaps we should do that!

Edit: Correction, we check a number of things, but not every "combo" ... we should add more checking.

src/modules/kvs-watch/kvs-watch.c
chu11: This aux is simply a "flag".  It indicates which future is the last "initial" blobref being loaded, thus needing to send the initial sentinel afterwards.

You are correct there could be other ways to do this.  We could track the number of responses already sent and using some counting use that to indicate when to send the initial sentinel.  But I felt this was easier / simpler.

garlick: Oh duh, I see the big block comment now where the aux value is set.

src/common/libkvs/kvs_lookup.c
garlick: I just assumed it was an artifact from before, since now the convention is "empty response" == "sentinel" both in the API and the RPC right?  I wasn't sure why the API code needed to set a flag when it could just be unaware and pass the empty payload on to the API user.

chu11:
re-pushed, fixing up things per comments above.  ~~Will set MWP~~ (Oops, confused this with a different PR)

- I went with the flag name `FLUX_KVS_WATCH_INITIAL_SENTINEL`, just removing the "APPEND".  

- `libkvs/kvs_lookup.c` - the "empty payload" is simply passed through instead of using a sentinel flag.  Just have to add some logic that `ctx->treeobj` could be NULL.

-  added some extra `kvs-watch` EPROTO errors checks and tests

src/modules/kvs-watch/kvs-watch.c
```diff
@@ -1411,6 +1414,11 @@ static void lookup_cb (flux_t *h,
         errno = EINVAL;
         goto error;
     }
+    if (!(flags & FLUX_KVS_WATCH) && (flags & valid_watch_flags)) {
+        errmsg = "Must set KVS watch flag with other watch flags";
+        errno = EINVAL;
+        goto error;
+    }
```
garlick, line 1459: Maybe since this is the `kvs-watch.lookup` RPC, it would make sense to just silently add the FLUX_KVS_WATCH flag if it is needed downstream or whatever, rather than requiring users to add it?

src/common/libkvs/kvs_lookup.c
```diff
@@ -232,6 +232,32 @@ static struct lookup_ctx *get_lookup_ctx (flux_future_t *f)
...
+    if (ctx->val_obj) {
+        json_decref (ctx->val_obj);
+        ctx->val_obj = NULL;
+    }
+    if (ctx->dir) {
+        flux_kvsdir_destroy (ctx->dir);
+        ctx->dir = NULL;
+    }
+}
+
```
garlick: Suggestion, move theis portion of lookup_ctx into a sub structure and pass &that to the cleanup function.  Then all the explicit member zeroing can be replaced with a single memset.

src/modules/kvs-watch/kvs-watch.c
```diff
@@ -442,6 +457,23 @@ static int handle_initial_response (flux_t *h,
...
+             * completed.  N.B. use the future as a random pointer to
+             * set for aux, we don't actually use the "flag" value.
+             */
+            flux_future_t *f = zlist_tail (w->loads);
+            if (flux_future_aux_set (f, "initial_sentinel", f, NULL) < 0) {
+                flux_log_error (h,
+                                "%s: failed to set initial_sentinel flag",
+                                __FUNCTION__);
+                goto error_respond;
+            }
```
garlick: Instead of dropping a real pointer in that aux entry and having to explain it, maybe it would better to make it an obvious fake, like
```c
#include "ccan/ptrint/ptrint.h"
flux_future_aux_set (f, "initial_sentinel", int2ptr (1), NULL);
```

doc/man3/flux_kvs_lookup.rst
```diff
@@ -155,6 +155,17 @@ FLUX_KVS_WATCH_APPEND
...
+   Specified along with FLUX_KVS_WATCH_APPEND, this flag will result
+   in an empty RPC message to be sent when all "initial" or "current"
+   KVS data has been sent.  This empty RPC message serves as a
+   sentinel and is useful for identifying what KVS data is "new" vs
+   "old" at the point in time a :var:`key` starts to be watched.
+   Normally an empty RPC message will result in EPROTO errors, but
+   :func:`flux_kvs_lookup_get` and :func:`flux_kvs_lookup_get_raw`
+   will recognize the sentinel RPC and return NULL / zero for its
+   results.
+
```
garlick: It seems like this description is a bit focused on the internal protocol.  Here's a stab at rewriting it to be more API focused:
> Modifies the behavior of the FLUX_KVS_WATCH_APPEND flag to notify the caller once
> all the data that has been appended to the key thus far has been received.
> flux_kvs_lookup_get()  indicates this by returning success with a NULL value parameter.

src/modules/job-info/watch.c
```diff
@@ -258,16 +274,29 @@ static void watch_continuation (flux_future_t *f, void *arg)
...
+        if (errno == ENODATA) {
+            if (w->kvs_watch_canceled && !w->cancel)
+                goto cleanup;
+            /* N.B. if guest in main, we just read / streamed the
+             * entire contents from KVS and that's it.  So add
+             * sentinel before ENODATA at the end.
+             */
+            if (w->flags & FLUX_JOB_EVENT_WATCH_INITIAL_SENTINEL
+                && w->guest_in_main
+                && !w->initial_sentinel_sent)
```
garlick: I think the test of `!w->initial_sentinel_sent` is redundant with the one in `send_initial_sentinel()`

garlick (approved):
Just a few inline comments.

src/modules/kvs-watch/kvs-watch.c
chu11: The primary reason for this (it's also done in `libkvs`) is b/c FLUX_KVS_STREAM semi-conflicts with FLUX_KVS_WATCH.   So we want to make sure "watching" is what the user really wants to do.

chu11:
thanks @garlick.  fixed up per your comments above, will set MWP

PR #6881, opened 2025-06-17
add `flux multi-prog` for MPMD support

grondo:
This PR takes the MPMD workaround script from #2414 and turns it into an official Flux command: `flux multi-prog`.

This is meant to be used as the command and arguments to `flux run` and `flux submit` for MPMD support in the near term until more advanced capabilities are added to jobspec. This allows Flux to have this basic capability instead of requiring users to search an issue and copy a workaround script.

This PR includes a manpage and tests for the new command.


garlick (approved):
All looks good to me and I think we've had a few people use this already so probably has been a little battle tested.

grondo:
This is approved -- I wonder if we should just add it even though it doesn't fully satisfy some of the use cases at this point?

grondo:
Setting MWP just to make `flux multi-prog` available in the next release.

Issue #6878, opened 2025-06-13
Suggestion: Add a way to log drain/undrain messages to syslog


kkier:
At some point in the future we'll be able to send the flux event log(s) to elasticsearch, but we're not there yet.  Being able to query elastic for drain/undrain events would (help) solve a few problems, and right now it seems like the most straightforward way to get those events into elastic right now is to send them to syslog.

My naive idea was "oh, just put a syslog() call in whatever handles **flux (un-)drain**" but that's pretty obviously not an awesome approach once you stop and think about it.  (What happens if the drain fails?  Or if some of the nodes fail and some don't because the user used **-u**, etc?)

grondo:
The obvious thing would be to log a message at the same point the `drain` or `undrain` event is posted to the eventlog in the resource module's `drain_cb()` and `undrain_cb()` msg handlers, but I wonder if anyone has a more clever idea.

We could easily get something working now with a Python script that uses the `ResourceJournalConsumer` interface to turn `drain` and `undrain` events into syslog messages. It would need somewhere to run though, and it would be prone to emitting duplicate log messages when restarted. Perhaps a version that runs out of cron, and by default only processes events that occurred since the last time it ran? Meh. Just logging at the time the event is posted does seem simplest...

garlick:
We have a path from flux_log() to stderr and from there to systemd journal and on to syslog, I think.

Maybe we can work out something using that?

grondo:
Yeah, sorry that's what I meant by the obvious thing, e.g.
```diff
diff --git a/src/modules/resource/drain.c b/src/modules/resource/drain.c
index 3a070bc2d..aecfeef92 100644
--- a/src/modules/resource/drain.c
+++ b/src/modules/resource/drain.c
@@ -367,6 +367,15 @@ static void drain_cb (flux_t *h,
                                "overwrite", overwrite);
     if (rc < 0)
         goto error;
+
+    flux_log (h,
+              LOG_INFO,
+              "drain idset=%s nodelist=%s overwrite=%d reason=%s",
+              idstr,
+              nodelist,
+              overwrite,
+              reason ? reason : "");
+
     free (nodelist);
     free (idstr);
     idset_destroy (idset);
@@ -408,6 +417,15 @@ int drain_rank (struct drain *drain, uint32_t rank, const char *reason)
                           "reason", reason) < 0
         || reslog_sync (drain->ctx->reslog) < 0)
         goto done;
+
+    flux_log (h,
[...]

PR #6882, opened 2025-06-18
resource: log drain/undrain events

grondo:
This simple PR adds logging of drain and undrain events to the resource module as requested in #6878.

An open question is probably if the simplistic key=val format of the log message is sufficient for the use case. 

e.g.:
```
2025-06-18T13:18:44.436906-07:00 resource.info[0]: drain idset=0-1 nodelist=tuolumne[1006,1032] overwrite=0 reason=this is a drain reason
2025-06-18T13:18:57.069487-07:00 resource.info[0]: undrain idset=0 nodelist=tuolumne1006 reason=undrain reason 1
2025-06-18T13:19:02.873886-07:00 resource.info[0]: undrain idset=1 nodelist=tuolumne1032 reason=undrain reason 2
```

### DIFF
--- a/src/modules/resource/drain.c
+++ b/src/modules/resource/drain.c
@@ -367,6 +367,15 @@ static void drain_cb (flux_t *h,
                                "overwrite", overwrite);
     if (rc < 0)
         goto error;
+
+    flux_log (h,
+              LOG_INFO,
+              "drain idset=%s nodelist=%s overwrite=%d reason=%s",
+              idstr,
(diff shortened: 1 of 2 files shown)


grondo:
@kkier, can you give an ACK on this one if it will meet your needs? Thanks!

kkier:
This looks perfect from my perspective!

grondo:
Thanks! I'll set MWP

PR #6884, opened 2025-06-19
idset: add round-robin allocation flag

garlick:
Problem: in working on VNI support for flux-coral2, I found it would be handy to have something like `idset_alloc(3)` and `idset_free(3)` to allocate form a configurable VNI pool, except not immediately re-use a VNI after it is freed, in case there are race conditions in VNI cleanup.

This adds a FLUX_IDSET_ALLOC_RR flag to select "round robin" allocation for `idset_alloc(3)`.   If we wanted to we could use this for matchtags, although it does raise the complexity of an allocation from O(1) to O(log(n)) and matchtag handling is pretty well tested at this point.

### DIFF
--- a/doc/man3/idset_create.rst
+++ b/doc/man3/idset_create.rst
@@ -106,6 +106,10 @@ IDSET_FLAG_INITFULL
    The idset is created full instead of empty.  If specified with
    IDSET_FLAG_AUTOGROW, new portions that are added are also filled.
 
+IDSET_FLAG_ALLOC_RR
+   Change :func:`idset_alloc` to begin searching for an id after the
+   most recently allocated one, rather than taking the first available.
+
 IDSET_FLAG_COUNT_LAZY
    The running count is not maintained and :func:`idset_count` uses a slower
    iteration method.  Not maintaining the count makes set/clear operations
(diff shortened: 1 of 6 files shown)


doc/man3/idset_create.rst
```diff
@@ -106,6 +106,10 @@ IDSET_FLAG_INITFULL
    The idset is created full instead of empty.  If specified with
    IDSET_FLAG_AUTOGROW, new portions that are added are also filled.
 
+IDSET_FLAG_ALLOC_RR
+   Change :func:`idset_alloc` to begin searching for an id after the
+   most recently allocated one, rather than taking the first available.
+
```
grondo, line 112: Should the note about time complexity be added here?

grondo (approved):
LGTM! One suggestion in line

doc/man3/idset_create.rst
garlick: I think at one point I had time complexity notes everywhere in the man page and it was a bit distracting,so I backed it off and stated at the top which operations were O(1) and which were O(log(n)).  So the `idset_alloc()` docs pointedly state whether they are using `idset_first()` or `idset_next()`.  Too indirect?

garlick: On second thought, maybe I should at least mention the complexity in the `idset_alloc(3)` page since it's only covered in `idset_create(3)`?

grondo: Ah, I must have missed the mention when reading the diff, sorry! you can ignore this comment then.

garlick:
Fixed a commit message typo and pushed a little update to the man page on time complexity.

garlick:
Thanks, setting MWP.

Issue #6876, opened 2025-06-09
need unique identifier for jobs run at any level on a system


garlick:
Problem: the slingshot fabric manager requires an application using hardware collectives to be assigned a "job step id".  The design is for ALPS, Slurm, and PBS-Pro/PALS which have the traditional two-level job model.  Flux doesn't fit that model.

The (job id, step id) tuple must be unique across the system and is used as a "handle" for a set of multicast addresses that are not otherwise known to the workload manager, since the multicast addresses are created by libfabric down in the application.

Joy of joys, the fabric manager wants a string type for the step ID, which means, in theory, we could substitute a "job path" as already introduced in flux-uri(1).

This can all be dealt with in flux-coral2 without perturbing flux-core, but since we had discussed things like this in the past but hesitated due to lack of use case, I thought I would bring it up again.

Should we do something like record the broker's job path in a broker attribute for a start?

grondo:
> Should we do something like record the broker's job path in a broker attribute for a start?

This makes sense to me and could probably be done quite simply, so I don't really see much downside.

PR #6885, opened 2025-06-19
add FLUX_JOB_ID_PATH to the job environment

garlick:
Problem: Flux jobs run in sub-instances do not have a unique ID like slurm's jobid.step.

Add the FLUX_JOB_PATH environment variable to the job environment.  The path consists of job IDs separated by `/` characters, with `/` designating the top level instance.  For example, a job run in a batch job of a system instance might have a FLUX_JOB_PATH of  `/ƒD1oBT5/ƒDCB6RV`.

I hope it is not going too far to put this in the environment of every job.   In #6876 I had proposed just adding a broker attribute and passing it through PMI.  This was simpler and felt more complete.  We can go back to the other plan if there is disagreement on this approach.

### DIFF
--- a/doc/man7/flux-environment.rst
+++ b/doc/man7/flux-environment.rst
@@ -25,6 +25,13 @@ The following are set in the environment of each task spawned by
 
       NUMERIC_JOB_ID=$(flux job id $FLUX_JOB_ID)
 
+.. envvar:: FLUX_JOB_ID_PATH
+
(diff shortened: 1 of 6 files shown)


garlick:
Oh I didn't think of that.  Do FLUX_JOB_ID_PATH and `jobid-path` sound OK?

grondo:
Sounds good to me if ok with you!

garlick:
I'll set MWP - thanks!

PR #6891, opened 2025-06-25
broker: add broker.getenv RPC

garlick:
Problem: in nested Flux instances, a shell plugin may need to access environment variable settings for the local broker.
    
Specifically, a cray-slingshot plugin needs to find inherited CXI services, which may differ on each broker.
    
Add a new broker RPC: broker.getenv.  A user sends an array of key names and gets back a dict of key/value pairs.  Any environment variables for which getenv(3) returns NULL are omitted from the object.
    
The RPC is restricted to the instance owner.

### DIFF
--- a/t/t0034-broker-getenv.t
+++ b/t/t0034-broker-getenv.t
@@ -1,0 +1,50 @@
+#!/bin/sh
+
+test_description='Test broker.getenv rpc'
+
+. `dirname $0`/sharness.sh
+
+export BROKER_ENV_TEST=abcd
+unset BROKER_ENV_TEST2
+
+test_under_flux 1
+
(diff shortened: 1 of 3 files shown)


t/t0034-broker-getenv.t
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
+
+test_description='Test broker.getenv rpc'
```
chu11, line 3: should add some tests that ensure only instance owner can get things and non-instance owner can't?

garlick: OK tacked on that extra test.  Thanks - I'll set MWP.

PR #6894, opened 2025-06-26
add flux module remove --cancel option

garlick:
This adds an option to remove an ornery module by sending it a `pthread_cancel(3)` instead of politely requesting that it shut down.  That required the addition of a pthread cleanup callback, which triggered a bit of refactoring.


grondo (approved):
LGTM!

garlick:
Thanks!

PR #6897, opened 2025-06-30
jobtap: add `flux_jobtap_call()`

grondo:
This PR adds a new `flux_jobtap_call()` function to the jobtap API which allows a plugin to invoke arbitrary callbacks in the jobtap plugin stack.

### DIFF
--- a/doc/man7/flux-jobtap-plugins.rst
+++ b/doc/man7/flux-jobtap-plugins.rst
@@ -60,6 +60,8 @@ name will be ``job-manager.<name>.<method>``. If a plugin does not
 set a name with ``flux_plugin_set_name(3)``, then the basename of the
 plugin file will be used with the trailing ``.so`` removed.
 
+.. _arguments:
+
 JOBTAP PLUGIN ARGUMENTS
 =======================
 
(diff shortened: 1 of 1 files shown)


src/modules/job-manager/jobtap.c
```diff
@@ -2683,6 +2683,26 @@ int jobtap_validate_updates (struct jobtap *jobtap,
...
+    if (p == NULL
+        || !(jobtap = flux_plugin_aux_get (p, "flux::jobtap"))
+        || !jobtap->ctx
+        || !args) {
+        errno = EINVAL;
+        return -1;
+    }
+    if (!(job = jobtap_lookup_jobid (p, id)))
+        return -1;
+    return jobtap_stack_call (jobtap, jobtap->plugins, job, topic, args);
```
garlick, line 2704: I would imagine calling the documented jobtap topics would lead to sadness here.  Any way to prevent that?  Maybe all topics used here should be required to have a `user.*` prefix or something?

garlick (approved):
LGTM!  Just one question/concern inline, but nothing to prevent this from going in.

src/modules/job-manager/jobtap.c
grondo: I'm on the fence. We don't require the same for the equivalent shell functions, but I can see your point and the price of failure is higher in this case. However, I'm not sure `user.` makes sense here because this call could be implemented by system plugins. I think all the documented jobtap topics start with `job.` so maybe disallowing a call of `job.*` here would be sufficient?

garlick: Works for me!

grondo: Ok, I pushed an update that doesn't allow `flux_jobtap_call()` to be used with topic strings starting with `job.`.
The job manager also uses `plugin.query` and `conf.update`, however those are not prevented here because perhaps someone will think of a reason to call them from another plugin, and the chances of misuse are very low.

grondo:
I'll go ahead and set MWP here assuming tests pass.

Issue #6890, opened 2025-06-25
python: set the default jobspec environment from `os.environ` in `JobspecV1` `from_*` factory functions


grondo:
As discussed in [this comment](https://github.com/fux-framework/flux-core/pull/6858#discussion_r2159680216) in #6858, the `JobspecV1` factory functions should have a default environment set from `os.environ`. This shouldn't break existing usage since the `environment` setter replaces the environment dictionary. This is trivial to add should be a big improvement in the quality of the interface.

PR #6898, opened 2025-06-30
set default jobspec jobspec using `os.environ` in `JobspecV1` factory functions

grondo:
As noted in #6890, it is unexpected that the default jobspec environment when not explicitly passing `environment` is empty.

I can't really think of a good reason for this default, and it is easy to fix so let's do that.

This shouldn't break any existing users since setting `jobspec.environment = dict` will still work.

Just in case, we should get a couple reviews of this one.

### DIFF
--- a/src/bindings/python/flux/job/Jobspec.py
+++ b/src/bindings/python/flux/job/Jobspec.py
@@ -441,7 +441,7 @@ class Jobspec(object):
     @property
     def environment(self):
         """
-        Environment of job. Defaults to ``None``.
+        Environment of job.
         """
         try:
             return self.jobspec["attributes"]["system"]["environment"]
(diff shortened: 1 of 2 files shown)


wihobbs:
It's likely we will need to update documentation in several different places for this one...

grondo:
> It's likely we will need to update documentation in several different places for this one...

I'm not sure we'll _need_ to, but examples that create a jobspec object then set `jobspec.environment = dict(os.environ)` could be updated to remove the unnecessary second step.

This change does not break the existing usage.

wihobbs:
As I looked deeper in to it, I realized it's not a problem as I thought it would be. I was expecting we'd have several places where the phrase "job environments for Jobspecs created in Python have an empty environment by default" would show up, but that proved not to be true.

grondo:
Thus the user confusion and surprise :sweat_smile:

grondo:
@wihobbs: if you approve I'll go ahead and set MWP here.

grondo:
Thanks! Setting MWP

Issue #5035, opened 2023-03-29
flux job attach: support option to not output whole output history


chu11:
Began looking at #4869 briefly and realized it can be annoying that `flux job attach` (and tool proposed in #4869) will output ALL stdout/stderr when being watched.

There should be an option to only output new information.  Like a user outputs some status to stdout and doesn't need to see the last 10 hours of status, only the new stuff coming in.  I'm thinking something equivalent to `tail -f`.

I tried to add a basic flag to `flux_job_event_watch()` to deal with this but realized that wouldn't work.  We really need something is knowledge of RFC24 standard I/O format so it knows that the header/redirect/etc. type information should be forwarded but not data.  

Luckily, since eventlog's can never be zero length per RFC18, that eliminates some corner case handling and some trickery in `job-info` may work.  Otherwise we need to get the eventlog's initial length.  It could be a costly operation, perhaps a `flux_kvs_lookup_length()` or `flux_kvs_lookup_stats()` kinda function would be less costly.

grondo:
I don't think this is high priority. Specifically in #4869 we _do_ want all the job output, since that is the point of the use case (to not lose any job output or errors when the instance ends)

chu11:
> I don't think this is high priority. Specifically in #4869 we do want all the job output, since that is the point of the use case (to not lose any job output or errors when the instance ends)

Understood.  I guess when the idea of a potential `flux wait` tool was mentioned in #4869, it did suggest to me that a user may want to wait for a job and see its output in a "general" way.

grondo:
I'm not sure I follow what is meant by seeing output in a general way, sorry! I think without a use case, it would be difficult to design such a thing.

Oh, do you mean give me output starting from _now_ until the job finishes? That might be neat, but not sure there is an actual need for it?

chu11:
> Oh, do you mean give me output starting from now until the job finishes? That might be neat, but not sure there is an actual need for it?

Yeah this is what I was thinking.  Like someone submitted a job, walked away.  Then it started later and they are like "oh, I wonder what the status of it is right now", so they would want to do `flux watch --watch --follow` on it.

Dunno if there's a huge need, just something I was thinking of.  B/c I'm imaging the hours and hours of stdout that someone may not want to output.

grondo:
Yeah, maybe at first it could just gobble up lines without printing until a timestamp. (kind of like I assume `tail` does when reading `stdin`?)

chu11:
[...]

PR #6899, opened 2025-06-30
flux-job: support attach --tail

chu11:
Built on top of #6880, per discussion in #5035, this supports a new `--tail` option to `flux job attach`.

Note that this performs poorly, as we don't presently have a way to "jump forward" in the output eventlog.  So all events in the log are forwared to `flux job attach`, but `flux job attach`, just saves them off until it knows when to begin outputting "tail" events using the new "initial sentinel" in #6880.  The "jump forward" is for a future support.

### DIFF
--- a/t/t2500-job-attach.t
+++ b/t/t2500-job-attach.t
@@ -9,6 +9,7 @@ test_under_flux 4
 flux setattr log-stderr-level 1
 
 runpty=$SHARNESS_TEST_SRCDIR/scripts/runpty.py
+waitfile="${SHARNESS_TEST_SRCDIR}/scripts/waitfile.lua"
 
 test_expect_success 'attach: submit one job' '
 	flux submit echo foo >jobid1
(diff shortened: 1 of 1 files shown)


src/cmd/job/attach.c
```diff
@@ -347,6 +355,55 @@ static void handle_output_log (struct attach_ctx *ctx,
...
+            log_msg ("malformed event context");
+        if (eof)
+            return;
+    }
+    json_incref (entry);
+    if (zlist_append (ctx->tail_output, entry) < 0)
+        log_err_exit ("zlist_append");
+    zlist_freefn (ctx->tail_output,
+                  entry,
+                  json_decref_wrapper,
```
grondo: You could drop the `json_decref_wrapper()` and use `(zlist_free_fn *) json_decref` here instead.

grondo (approved):
The `flux job attach` changes here LGTM once the PR's predecessor is approved.

Just one small take it or leave it suggestion inline.

chu11:
@grondo thanks, will set MWP

Issue #6901, opened 2025-07-02
instance loses track of job duration with `resource.rediscover=true`


grondo:
When a Flux subinstance is started with `flux alloc` or `flux batch` and `--conf=resource.rediscover=true`, the instance loses track of the duration in _R_, since a new _R_ is created using dynamic discovery. This causes tools like `flux job timeleft` and libyogrt to fail in jobs run within the instance since the scheduler thinks the resource expiration is not limited.

PR #6903, opened 2025-07-03
resource: avoid losing R expiration on resource.rediscover

grondo:
This PR fixes the issue reported in #6901. When a user starts a batch or alloc job with `--conf=resource.rediscover=true`, the R from the parent, including any job expiration, is discarded in favor of a new R discovered with hwloc. This breaks `flux job timeleft`, libyogrt, as well as the default time limit assigned to jobs within the instance etc.

This PR saves the expiration from the original R and re-establishes it before writing the R to the kvs in the resource inventory subsystem.

### DIFF
--- a/t/t2316-resource-rediscover.t
+++ b/t/t2316-resource-rediscover.t
@@ -39,4 +39,21 @@ test_expect_success 'resource: resource.rediscover works with multiple nodes' '
 	test $NCORES3 -gt $NCORES &&
 	test $NCORES3 -eq $(($NCORES2*2))
 '
+test_expect_success 'resource: resource.rediscover saves expiration' '
+	flux alloc -t 5m -n1 -o cpu-affinity=off \
+	        --conf=resource.rediscover=true \
+		flux run flux job timeleft > timeleft.out &&
+	test_debug "cat timeleft.out" &&
+	test $(cat timeleft.out) -lt 800
+'
+# Note: 4294967295 is 32bit UINT_MAX. `flux job timeleft` should report
+# this value (or larger) for an unlimited expiration.
+test_expect_success 'resource: resource.rediscover works with unlimited expiration' '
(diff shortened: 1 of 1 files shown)


t/t2316-resource-rediscover.t
```diff
@@ -39,4 +39,19 @@ test_expect_success 'resource: resource.rediscover works with multiple nodes' '
 	test $NCORES3 -gt $NCORES &&
 	test $NCORES3 -eq $(($NCORES2*2))
 '
+test_expect_success 'resource: resource.rediscover saves expiration' '
```
garlick, line 42: commit message s/tetss/tests/

t/t2316-resource-rediscover.t
```diff
@@ -39,4 +39,19 @@ test_expect_success 'resource: resource.rediscover works with multiple nodes' '
...
+'
+# Note: 4294967295 is 32bit UINT_MAX. `flux job timeleft` should report
+# this value (or larger) for an unlimited expiration.
+test_expect_success 'resource: resource.rediscover works with unlimited expiration' '
+	flux alloc -n1 -o cpu-affinity=off \
+		flux run flux job timeleft > timeleft2.out &&
+	test_debug "cat timeleft2.out" &&
+	test $(cat timeleft2.out) -ge 4294967295
+'
+
```
garlick: I must be missing something - how is resource.redescover set in these tests?

garlick (approved):
Looks good - just a commit message nit and question about tests

t/t2316-resource-rediscover.t
grondo: Bah! Good catch thanks! Not sure what happened here

grondo: Fixed that issue and will set MWP.

PR #6904, opened 2025-07-03
content-sqlite/files: support validate target

chu11:
Support a "validate" RPC service in the content backing modules, simply to validate that an blobref hash is valid.  Without it, a "load" has to be used instead, which requires sending over the content data itself.  So it's pretty costly.

Multiple ways to do this of course, but I went with "return empty payload" for validated / success, and ENOENT for "not valid" / "not found".

I elected not to add API functions to `libcontent` and a "forwarder" in the `content` module, as this is only used by `fsck` at the moment.  IMO forwarding through the `content` cache module could suggest that verification within the `content` cache is ok, but that isn't ok.  We only want to validate in the backing module. 

Note, see associated RFC PR: https://github.com/flux-framework/rfc/pull/455

### DIFF
--- a/src/modules/content-files/content-files.c
+++ b/src/modules/content-files/content-files.c
@@ -195,6 +195,44 @@ error:
         flux_log_error (h, "error responding to store request");
 }
 
+/* Handle a content-backing.validate request from the rank 0 broker's
+ * content-cache service.  The raw request payload is a hash digest.
+ * The raw response payload is the blob content.
+ * These payloads are specified in RFC 10.
+ */
+static void validate_cb (flux_t *h,
+                         flux_msg_handler_t *mh,
+                         const flux_msg_t *msg,
+                         void *arg)
+{
+    struct content_files *ctx = arg;
+    const void *hash;
+    size_t hash_size;
+    char blobref[BLOBREF_MAX_STRING_SIZE];
+    const char *errstr = NULL;
+
+    if (flux_request_decode_raw (msg, NULL, &hash, &hash_size) < 0)
+        goto error;
+    if (hash_size != ctx->hash_size) {
+        errno = EPROTO;
+        goto error;
+    }
+    if (blobref_hashtostr (ctx->hashfun,
+                           hash,
+                           hash_size,
+                           blobref,
+                           sizeof (blobref)) < 0)
+        goto error;
+    if (filedb_validate (ctx->dbpath, blobref, &errstr) < 0)
+        goto error;
+    if (flux_respond_raw (h, msg, NULL, 0) < 0)
+        flux_log_error (h, "error responding to validate request");
+    return;
+error:
+    if (flux_respond_error (h, msg, errno, errstr) < 0)
+        flux_log_error (h, "error responding to validate request");
+}
+
 /* Handle a content-backing.checkpoint-get request from the rank 0 kvs module.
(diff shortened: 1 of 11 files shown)


garlick:
I agree that this shouldn't be forwarded through the cache, but it probably should work on the cache.   Then tools such as `flux fsck` could work on the cache if that's ever helpful.  Sort of like running system `fsck` on a block vs a character device.  It's really just a load without payload so no reason not to make it work everywhere.

chu11:
> I agree that this shouldn't be forwarded through the cache, but it probably should work on the cache. Then tools such as flux fsck could work on the cache if that's ever helpful. Sort of like running system fsck on a block vs a character device. It's really just a load without payload so no reason not to make it work everywhere.

Hmmm.  Right now `flux-fsck` by default works only on backing modules (i.e. uses the CONTENT_FLAG_CACHE_BYPASS flag).    So it can't work on the `cache` at the moment.

Perhaps having `fsck` work on the content cache and adding `validate` to the `content` module should be a TODO for later?

src/modules/content-files/content-files.c
```diff
@@ -195,6 +195,51 @@ void store_cb (flux_t *h,
...
+    if ((ret = filedb_validate (ctx->dbpath, blobref, &errstr)) < 0)
+        goto error;
+    if (ret) {
+        if (flux_respond_raw (h, msg, NULL, 0) < 0)
+            flux_log_error (h, "error responding to validate request");
+    }
+    else {
+        errno = ENOENT;
+        goto error;
+    }
```
garlick: Shouldn't `filedb_validate()` just return -1 and set errno = ENOENT when errno is not found?

src/modules/content-sqlite/content-sqlite.c
```diff
@@ -387,6 +426,41 @@ void store_cb (flux_t *h,
...
+        goto error;
+    if (ret) {
+        if (flux_respond_raw (h, msg, NULL, 0) < 0)
+            flux_log_error (h, "validate: flux_respond_raw");
+    }
+    else {
+        errno = ENOENT;
+        goto error;
+    }
+    return;
```
garlick: Same comment as with filedb.  Just have the validate function fail if the blob is invalid?

src/modules/content-files/content-files.c
```diff
@@ -195,6 +195,51 @@ void store_cb (flux_t *h,
         flux_log_error (h, "error responding to store request");
```
garlick, line 195: commit message: I think you mean "method" not "target"

garlick (approved):
LGTM, just some minor comments.

src/modules/content-files/content-files.c
chu11: it surprsingly doesn't

```
    if (stat (path, &statbuf) < 0) {                                                                                                        
        if (errno == ENOENT)                                                                                                                
            return 0;                                                                                                                       
        else                                                                                                                                
            return -1;                                                                                                                      
    }                      
```

I do see that it already sets ENOENT when returning 0 though.

I'll see if updating `filedb_validate()` to do that would be more wise.

Edit: Oh I'm sorry!  I wrote this function for this PR ! :man_facepalming:   I thought that `filedb_validate()` was written before this PR ;-)

chu11:
thanks, fixed up those small things, will set MWP

PR #6905, opened 2025-07-08
flux-fsck: support --rootref option

chu11:
Problem: Flux-fsck defaults to fsck-ing the checkpoint stored in the KVS backing store.  There is no way to fsck check an alternate
root reference.

Support a new --rootref option where users can specify a specific root reference to begin with.

----

side notes:

- this is written independently of #6798, it can be considered a "prelude" PR where eventually users could specify the index of the checkpoint they actually want to fsck.

### DIFF
--- a/t/t2816-fsck-cmd.t
+++ b/t/t2816-fsck-cmd.t
@@ -15,14 +15,17 @@ test_expect_success 'load kvs' '
 '
 test_expect_success 'create some kvs content' '
 	flux kvs put dir.a=test &&
+	flux kvs getroot -b > a.rootref &&
 	flux kvs put dir.b=test1 &&
 	flux kvs put --append dir.b=test2 &&
 	flux kvs put --append dir.b=test3 &&
 	flux kvs put --append dir.b=test4 &&
+	flux kvs getroot -b > b.rootref &&
 	flux kvs put dir.c=testA &&
 	flux kvs put --append dir.c=testB &&
 	flux kvs put --append dir.c=testC &&
 	flux kvs put --append dir.c=testD &&
+	flux kvs getroot -b > c.rootref &&
 	flux kvs link dir alink &&
 	flux kvs namespace create testns &&
 	flux kvs put --namespace=testns dir.a=testns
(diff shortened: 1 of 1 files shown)


doc/man1/flux-fsck.rst
```diff
@@ -15,8 +15,8 @@ DESCRIPTION
 .. program flux fsck
 
 The :program:`flux fsck` checks the integrity of the KVS backing
-store, starting with the most recent checkpoint (root version) written
-to the backing store.
+store.  By default, it start with the most recent checkpoint (root version)
+written to the backing store.
```
garlick: s/start/starts/

doc/man1/flux-fsck.rst
```diff
@@ -34,6 +34,10 @@ OPTIONS
 
    Don't output diagnostic messages and discovered errors.
 
+.. option:: -r, --rootref
+
+   Use a specific root reference as the starting root version.
+
```
garlick: This might be expanded a little bit, e.g.

> -r, --rootref=BLOBREF
>
> Normally the check starts with the blobref in the most recent KVS checkpoint.
This option directs flux-fsck to start at an arbitrary point.  BLOBREF must refer to
an RFC 11 tree object of type "dir".

src/cmd/builtin/fsck.c
```diff
@@ -337,23 +334,33 @@ static int cmd_fsck (optparse_t *p, int ac, char *av[])
...
+            struct tm tm;
+            if (!timestamp_from_double (timestamp, &tm, NULL))
+                strftime (buf, sizeof (buf), "%Y-%m-%dT%T", &tm);
+            fprintf (stderr,
+                     "Checking integrity of checkpoint from %s\n",
+                     buf);
+        }
     }
 
     fsck_blobref (h, blobref);
```
garlick, line 365: Since the user can put any blobref on the command line, it might be nice to make this error message a little nicer.  Like maybe special case EPROTO and print "blobref does not refer to an RFC 11 tree object".
```
$ echo foo | flux content store
sha1-f1d2d2f924e986ac86fdf7b36c94bcdf32beec15
$ flux fsck --rootref=sha1-f1d2d2f924e986ac86fdf7b36c94bcdf32beec15
flux-fsck: cannot decode root tree object: Protocol error
```

src/cmd/builtin/fsck.c
```diff
@@ -374,6 +381,9 @@ static struct optparse_option fsck_opts[] = {
     { .name = "quiet", .key = 'q', .has_arg = 0,
       .usage = "Don't output diagnostic messages and discovered errors",
     },
+    { .name = "rootref", .key = 'r', .has_arg = 1,
+      .usage = "check integrity based on a specific root reference",
+    },
```
garlick: See man page comment.  I guess here it might be handy to have `.arginfo = "BLOBREF"` and maybe change usage string to something like
> Check integrity starting with BLOBREF which must refer to an RFC 11 tree object of type dir

garlick (approved):
Just some minor comments, otherwise LGTM!

chu11:
thanks @garlick, fixed up the things mentioned above.  Will set MWP

PR #6906, opened 2025-07-09
cli: add support for `CLIPlugin` interface to extend command line through plugins

wihobbs:
Outstanding to-dos:
- [x] eliminate need for `parser` argument in `CLIPlugin.add_options` method per James's suggestion
- [x] move prefix `--ex-` from the `CLIPluginRegistry` to each individual plugin per Ryan's suggestion


wihobbs:
> move prefix --ex- from the CLIPluginRegistry to each individual plugin per Ryan's suggestion

To pass arbitrary `**kwargs` to argparse without documenting and enumerating all of them, I created a `CLIPluginOption` class, which accepts arbitrary `**kwargs` (eventually invalid ones will error out, when the parser is created). It was logical to contain the `prefix="--ex-"` option there, if one plugin wants to use different prefixes, and it was workable in Python. 

Marking this as ready for review.

src/bindings/python/flux/cli/plugin.py
```diff
@@ -0,0 +1,180 @@
...
+class CLIPluginOption:
+    """Wrap Argparse's add_argument method with a class
+
+    Args:
+
+        name (str): Long option string being added (must begin with ``--``).
+        prefix (str): By default, ``--ex-`` is added as a prefix to ensure
+            arguments are namespaced. However, other flux-framework projects
+            can override this (as can other projects, though it is not
+            advised).
```
grondo: It is probably unnecessary to mention flux-framework and other projects here. Instead just document the option and that use of a prefix is strongly encourage so that it is clear that an option is provided by a plugin. e.g.:

> By default, ``--ex-`` is added as a prefix to plugin-provided options to ensure options are namespaced and that it is clear the option is not one provided by the core submission commands. However, the prefix may be overridden where it makes sense, e.g. with a site or project name.

src/bindings/python/flux/cli/plugin.py
```diff
@@ -0,0 +1,180 @@
...
+
+    Other keyword arguments, except ``dest=`` are passed along to
+    ``ArgumentParser.add_argument`` unchanged. ``dest=`` is a special case,
+    where dashes are replaced with underscores.
+    """
+
+    def __init__(self, name, prefix="--ex-", **kwargs):
```
grondo: It would be more user friendly if the option here dropped the `--{prefix}-` in the prefix argument, i.e. just have
```python
def __init__(self, name, prefix="ex", **kwargs):
```

Then add the `--{prefix}-` when prepending the prefix below. This also enforces the fact that this will always be a long option (i.e. I think the option would be registered without `--` if a user passed in a prefix without a leading `--` in the current code)

This also would allow a Flux provided plugin to pass `prefix=""` to get an empty prefix.

src/bindings/python/flux/cli/plugin.py
```diff
@@ -0,0 +1,180 @@
...
+        if not name.startswith("--"):
+            raise ValueError("Plugins must register only long options.")
+        if prefix:
+            if "dest" not in kwargs:
+                kwargs["dest"] = name[2:].replace("-", "_")
```
grondo: This could use a comment, something like
```python
# If dest is not already provided, create one from the option name (minus the leading `--`) replacing `-` with `_` as ArgumentParser would.
```

src/bindings/python/flux/cli/plugin.py
```diff
@@ -0,0 +1,180 @@
...
+
+        name (str): Long option string being added (must begin with ``--``).
+        prefix (str): By default, ``--ex-`` is added as a prefix to ensure
+            arguments are namespaced. However, other flux-framework projects
+            can override this (as can other projects, though it is not
+            advised).
+
+    Other keyword arguments, except ``dest=`` are passed along to
+    ``ArgumentParser.add_argument`` unchanged. ``dest=`` is a special case,
+    where dashes are replaced with underscores.
```
grondo: If I'm reading the code below correctly, `dest` is passed along to `argparse`. Only if `dest` is not present _and_ `prefix` is not `None` or empty it is created from `name` instead of `--ex-{name}`?

(That should probably be added to the documentation here: prefix does not automatically appear in dest)

This case should also be tested: Do we fail in a predictable way if two options are registered from different plugins with the same `name` but different `prefix`?

src/bindings/python/flux/cli/plugin.py
```diff
@@ -0,0 +1,180 @@
...
+    """
+
+    def __init__(self, name, prefix="--ex-", **kwargs):
+        if not name.startswith("--"):
+            raise ValueError("Plugins must register only long options.")
+        if prefix:
+            if "dest" not in kwargs:
+                kwargs["dest"] = name[2:].replace("-", "_")
+            self.name = name.replace("--", prefix)
+        self.__dict__.update(kwargs)
```
grondo: Just curious what you get using `self.__dict__` vs just the more explicit use of an attribute like `self.kwargs`?

src/bindings/python/flux/cli/plugin.py
```diff
@@ -0,0 +1,180 @@
...
+        self.prog = prog
+        if prog.startswith("flux "):
+            self.prog = prog[5:]
+        self.options = []
+
+    def add_option(self, option):
+        """Allow plugin to add additional options to itself
+
+        Plugins use:
+        >>> self.add_option(CLIPluginOption("--longopt", action="store_true", help="Help."))
```
grondo: I would slightly prefer if the user did not have to instantiate the `CLIPluginOption()`. You may have a good reason for this, but why not:

```Python
    def add_option(name, prefix="--ex-", **kwargs):
        self.options.append(CLIPluginOption(name, prefix, **kwargs))
```

This is  a bit more user friendly interface IMO and doesn't require the user to import or use the `CLIPluginOption` class in the common case. However, you may have tried this and run into some issue I don't see.
```

File: src/bindings/python/flux/cli/plugin.py [hunk 7]
```diff
@@ -0,0 +1,180 @@
...
+        if type(option) is not CLIPluginOption:
+            raise ValueError(
+                f"{option} is invalid, please use the CLIPluginOption class"
+            )
+        else:
+            self.options.append(option)
+
+    def get_options(self):
+        """Return a dict of options to add to the command line"""
+        return self.options
```
grondo: This method could be dropped in preference for just using `plugin.options`?

src/bindings/python/flux/cli/base.py
```diff
@@ -757,7 +758,11 @@ def __init__(self, prog, usage=None, description=None, exclude_io=False):
         self.exitcode = 0
         self.progress = None
         self.watcher = None
+        self.plugins = CLIPluginRegistry().load_plugins(prog)
         self.parser = self.create_parser(prog, usage, description, exclude_io)
+        group = self.parser.add_argument_group("Options provided by plugins")
+        for option in self.plugins.get_options():
+            group.add_argument(option.__dict__.pop("name"), **option.__dict__)
```
grondo: Another spot where having explicit attributes in the CLIPluginOption class would be good. Then you could split the option name and kwargs into `option.name` and `option.kwargs` and this code would be cleaner as:
```python
        for option in self.plugins.get_options():
            group.add_argument(option.name, **option.kwargs)

File: src/bindings/python/flux/cli/base.py [hunk 9]
```diff
@@ -1086,6 +1094,8 @@ def jobspec_create(self, args):
             for arg in args.add_file:
                 self.handle_add_file_arg(jobspec, arg)
 
+        self.plugins.modify_jobspec(args, jobspec)
```
grondo, line 1098: Suggestion: Add a comment here just to indicate what the call is allowing.

src/bindings/python/flux/cli/plugin.py
```diff
@@ -57,7 +57,7 @@ class CLIPlugin(ABC):  # pragma no cover
             e.g. "submit", "run", "alloc", "batch", "bulksubmit"
     """
 
-    def __init__(self, prog):
+    def __init__(self, prog, version=None, prefix="--ex-"):
```
grondo: This change may be in the wrong commit?

t/t2717-python-cli-plugins.t
```diff
@@ -0,0 +1,46 @@
...
+
+. $(dirname $0)/sharness.sh
+
+export FLUX_CLI_PLUGINPATH=${SHARNESS_TEST_SRCDIR}/cli-plugins/
+
+test_under_flux 4 job
+
+test_expect_success 'flux-alloc: base --help message is formatted correctly' '
+  flux alloc --help 
+'
```
grondo: This should check something ;-) Maybe that the plugin option section appears in the output.

grondo (commented):
Looking good! Thanks @wihobbs!

I made a first pass through with some fairly minor comments.

Regarding testing, let's make sure we're testing:

 - failure modes, e.g. plugins that add a conflicting option. Conflicting options with different prefixes (which I think would result in the same "dest")
 - use of a different prefix works
 - use of an empty or None prefix works

src/bindings/python/flux/cli/plugin.py
wihobbs: This is a great suggestion. It also eliminates having to check that the "option" is of the correct type.

src/bindings/python/flux/cli/plugin.py
wihobbs: More confusion, as is always the goal. (Fixed)

src/bindings/python/flux/cli/base.py
```diff
@@ -757,7 +758,11 @@ def __init__(self, prog, usage=None, description=None, exclude_io=False):
         self.exitcode = 0
         self.progress = None
         self.watcher = None
+        self.plugins = CLIPluginRegistry().load_plugins(prog)
```
wihobbs: @grondo What would you think of instead

```
self.plugins = CLIPluginRegistry(prog)
```
and the registry calling `load_plugins` as part of init? (or doing the same things but eliminating that function)

wihobbs: We'd have to use `__new__` instead of `__init__` to return an instance of the class, perhaps that's why you went with the `load_plugins()` function instead in the first prototype.

wihobbs: Ohh, I see why now. This is useful as-is in the validator. Ignore my noise!

src/bindings/python/flux/cli/plugin.py
wihobbs: I sort of wonder if we should _reject_ the empty string as a prefix...

src/bindings/python/flux/cli/base.py
grondo: You make a good point, though. Since it is internal use only, we could eventually update the class initializer to work in the way you propose if someone thinks of a clean solution.

wihobbs: I think I might have figured out how to do this, but I need to test it first.

src/bindings/python/flux/cli/plugin.py
grondo: No. We talked about explicitly allowing this for some cases.

src/bindings/python/flux/cli/base.py
grondo: It does seem like you could just call `self.load_plugins(prog)` in the `__init__` function, once `prog` is added as an arg.

I guess it is just a style question at this point. The one time you may not want to do this is when you might add other setup or configuration methods to the class in the future. Then you want to be able to instantiate an object of the class _before_ loading plugins. This kind of reason is why I tend to first create classes that only set initial attribute values in a class initializer, and other methods to fully set up or configure the class.

grondo: Hm. Now that I'm thinking about it, either way it does seem like `prog` should be passed to the class initializer and not only the `load_plugins()` method (that seems like a design flaw in the prototype). Good catch.

wihobbs:
oops y'all saw nothing 👀

grondo:
BTW, this seems ready to remove WIP. At the same time reword the [python: add prototype for submission cli plugins](https://github.com/flux-framework/flux-core/pull/6906/commits/b8f59d478ca1c06a0e52ae0b6f2b83a91050379a) commit to remove the "prototype" language.

src/bindings/python/flux/cli/plugin.py
wihobbs: > Do we fail in a predictable way if two options are registered from different plugins with the same name but different prefix?

I decided this should be a failure if the `kwargs["dest"]` option conflicts. Users have always had the option to provide their own dest, and perhaps there are some valid cases for having a plugin with the same `name` but different `prefix`es (if they know this and put a different `dest` too). I'll push up how I did this soon.

src/bindings/python/flux/cli/plugin.py
```diff
@@ -0,0 +1,180 @@
+##############################################################
+# Copyright 2024 Lawrence Livermore National Security, LLC
```
cmoussa1: does this copyright year need to be updated to 2025?

t/cli-plugins/validateonly.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class ValidateShellVerboseOpt(CLIPlugin):
+    name = "shell.options.verbose"
+
+    def validate(self, jobspec):
+        try:
+            verbosity = jobspec.attributes["system"]["shell"]["options"]["verbose"]
+            if not verbosity.isdigit():
+                raise ValueError(f"{self.name}: expected integer, got {verbosity}")
```
cmoussa1: IMHO it could be useful to return the _type_ of `verbosity` instead of `verbosity` itself in this exception message, but you could probably take or leave this.

t/t2717-python-cli-plugins.t
```diff
@@ -0,0 +1,62 @@
...
+    def __init__(self, prog):
+        super().__init__(prog)
+        self.add_option("--gpumode", help="I want to mess up the gpumode")
+EOF
+  test_must_fail flux run -N1 --dry-run sleep inf 2> out5.err &&
+  grep "conflicts with another option" out5.err &&
+  rm ../cli-plugins/doubleplugin.py
+'
+
+test_done
```
cmoussa1, line 62: It looks like the indentation might be a little inconsistent (at least looking at it on GitHub). Is it possible some indentations are using spaces while others are using tabs?

cmoussa1 (commented):
This is looking pretty good! I have a couple of minor comments after a pass through.

src/bindings/python/flux/cli/plugin.py
wihobbs: It's been [recommended](https://llnl-performance.slack.com/archives/CBNV6RG8Y/p1747091122086739?thread_ts=1747077931.876609&cid=CBNV6RG8Y) that we just remove the year, since it is [unnecessary, and cumbersome to update (which is recommended but not done)](https://www.linuxfoundation.org/blog/blog/copyright-notices-in-open-source-software-projects). I'll just remove it.

t/cli-plugins/validateonly.py
wihobbs: Hmm, this plugin isn't being tested _at all_, that seems bad.

src/bindings/python/flux/cli/plugin.py
grondo: What we actually do is put the year the file was originally included in the source, then never worry about updating it. I don't have a problem with removing the year, but would like to get an ACK from @garlick since we haven't yet dropped the year everywhere else.

wihobbs: Ok, that seems like maybe an out of scope conversation for this PR. I'll put 2025 in the header to be consistent and we can talk about dropping the year later.

cmoussa1: This is what I do in flux-accounting - just include the year the file was originally added!

wihobbs:
Thanks @cmoussa1! I believe I addressed all of your comments. I also fixed a failing test in `make distcheck` and added an additional test to make sure we can support plugins with the same name and different prefixes _if_ they specify a different `dest`.

src/bindings/python/flux/cli/plugin.py
```diff
@@ -0,0 +1,180 @@
...
+    def add_option(self, name, **kwargs):
+        """Allow plugin to register options in its dictionary
+
+        Args:
+
+            name (str): Long option string being added (must begin with ``--``).
+            prefix (str): By default, ``--ex-`` is added as a prefix to ensure
+                plugin-provided options are namespaced. The prefix may be
+                overridden with a site or project name.
+
```
grondo: `prefix` is no longer an arg here, so should be removed from docs. Perhaps move the description prefix, how/when to override it, etc to the class documentation?

src/bindings/python/flux/cli/plugin.py
```diff
@@ -0,0 +1,180 @@
...
+                and issubclass(entry, CLIPlugin)
+                and entry != CLIPlugin
+            ):
+                self.plugins.append(entry(program))
+
+    def _load_plugins(self, program):
+        """Load all cli plugins from the standard path"""
+        for path in glob.glob(f"{self.plugindir}/*.py"):
+            self._add_plugins(import_path(path), program)
+        option_dests = []
```
grondo: I think it might be more efficient if option_dests was a dictionary, since you are just looking for name collisions.
i.e. instead of appending to `option_dests` just use `option_dests[dest] = 1`. Then the lookup is ~O(1).

src/bindings/python/flux/cli/base.py
```diff
@@ -36,6 +36,7 @@
 
 import flux
 from flux import debugged, job, util
+from flux.cli.plugin import CLIPluginRegistry
```
grondo, line 39: Commit message probably needs an update to describe current behavior (e.g. `add_options()` is not a plugin provided method anymore but is instead called by plugins during init, then collected plugins are added to the "options provided by plugins" sections)

t/t2717-python-cli-plugins.t
```diff
@@ -0,0 +1,76 @@
...
+  grep "shell.options.verbose: expected integer, got <class" out5.err
+'
+test_expect_success 'flux-alloc: new plugin with the same name/dest as another fails' '
+  cat <<-EOF > ${SHARNESS_TEST_SRCDIR}/cli-plugins/doubleplugin.py && 
+from flux.cli.plugin import CLIPlugin 
+class BadPlugin(CLIPlugin):
+    def __init__(self, prog):
+        super().__init__(prog)
+        self.add_option("--gpumode", help="I want to mess up the gpumode")
+EOF
```
grondo: When using `-EOF`, the "heredoc" `<<` will ignore tabs, so you can add one tab in front of the content of this file so it is more readable in the sharness script.

t/t2717-python-cli-plugins.t
```diff
@@ -0,0 +1,76 @@
...
+  rm ${SHARNESS_TEST_SRCDIR}/cli-plugins/doubleplugin.py
+'
+test_expect_success 'flux-alloc: new plugin with the same name, diff dest/prefix is accepted' '
+  cat <<-EOF > ${SHARNESS_TEST_SRCDIR}/cli-plugins/doubleplugin.py && 
+from flux.cli.plugin import CLIPlugin 
+class GoodPlugin(CLIPlugin):
+    def __init__(self, prog, prefix="diff"):
+        super().__init__(prog, prefix=prefix)
+        self.add_option("--gpumode", dest="different_gpu", help="I am a valid plugin")
+EOF
```
grondo: same comment as above.

grondo (commented):
Just did a quick run through and this is very close to ready to merge!

I'll actually check out the branch and do some hand testing then come back and do a another review.

src/bindings/python/flux/cli/plugin.py
wihobbs: Will that lookup have to be wrapped in a try/catch, because a lookup of a dict key that doesn't exist will throw a `KeyError`? We could look in `option_dests.keys()` but that's a method that returns a list so it'd be the same time complexity. Perhaps I'm missing something simple. 😬

cmoussa1: You could use the [`.get()`](https://www.geeksforgeeks.org/python/python-dictionary-get-method/) method with a dict access that will return `None` (or I _believe_ a custom default) instead of raising a `KeyError`.

wihobbs: Wow that is useful. 🤦

grondo: I mean to continue using `if dest in option_dests`, it is the same code but (I presume) faster because `in` in a list would have to search the whole list, whereas `in` for a dict is just using the lookup.

t/t2717-python-cli-plugins.t
wihobbs: I actually have to drop these `cat` calls out of the tests because they're failing under `distcheck`. Will push up a proposed fix soon. I created new directories with failing plugins, but symlinked the working ones into that directory.

This is an unfortunate limitation of only having one statically defined directory for plugins, perhaps a future improvement could take a colon-delimited `PATH` or load plugins dynamically.

grondo: Ah, that makes sense. Yeah, we should add that after this one lands. It could perhaps make use of `flux.util.xdg_searchpath()` which searches the standard XDG directories...

grondo: Oh, I had forgotten the path can already be overidden using `FLUX_CLI_PLUGINPATH`. You could have tested these dynamically created plugins by setting that path. However, what you have is also fine and I do think we need to update this to take a colon delimited path, or support `XDG_CONFIG_HOME` and `XDG_CONFIG_DIRS` (need to ponder which is better). So keeping things as is for now is good.

wihobbs:
Thanks @grondo! Have fun on vacation! 🌴

cmoussa1 (approved):

PR #6907, opened 2025-07-10
Fix for Ubuntu Noble Docker Image to Prevent PMIx+MPICH Issues

ilumsden:
In testing some stuff for our upcoming [performance tools and benchmarking tutorial](https://github.com/llnl/benchpark-tutorial) for [HPDC](https://hpdc.sci.utah.edu/2025/workshops.html), I discovered that there are still issues with the Ubuntu Noble Docker image regarding MPICH. The majority of the issue is covered in this [Launchpad bug report](https://bugs.launchpad.net/ubuntu/+source/mpich/+bug/2072338). Long story short, the apt-installed MPICH builds with PMIx support, but still uses the Hydra process manager. This causes issues when trying to run MPI code because Hydra does not provide a PMIx interface.

The Docker image for Ubuntu Noble already tried to work around this by building MPICH from source. However, the image is still installing `libmpich-dev`. This, combined with loader search paths, is causing MPI applications to link against the apt-installed `libmpich.so`, which causes the same issue as when installing all of MPICH from apt.

To fix this issue, this PR simply removes `libmpich-dev` from the Ubuntu Noble Docker image. It also adds installs of `gfortran` because, during local testing, the script to build MPICH from source was failing due to not being able to find a Fortran compiler.

### DIFF
--- a/src/test/docker/noble/Dockerfile
+++ b/src/test/docker/noble/Dockerfile
@@ -45,11 +45,9 @@ RUN apt-get update \
 # default version
         gcc-13 \
         g++-13 \
-        gfortran-13 \
 # newest
         gcc-14 \
         g++-14 \
-        gfortran-14 \
 # Python
         libffi-dev \
 ## python 3.12
--- a/src/test/docker/noble/Dockerfile
+++ b/src/test/docker/noble/Dockerfile
@@ -39,14 +39,17 @@ RUN apt-get update \
         make \
         cmake \
         ninja-build \
+        gfortran \
         clang-18 \
         clang-tools-18 \
 # default version
         gcc-13 \
         g++-13 \
+        gfortran-13 \
 # newest
         gcc-14 \
         g++-14 \
+        gfortran-14 \
 # Python
         libffi-dev \
 ## python 3.12
@@ -65,7 +68,6 @@ RUN apt-get update \
         libsqlite3-dev \
         uuid-dev \
         libhwloc-dev \
-        libmpich-dev \
         libevent-dev \
         libarchive-dev \
         libpam-dev \
--- a/scripts/fetch-and-build-mpich.sh
+++ b/scripts/fetch-and-build-mpich.sh
@@ -6,7 +6,7 @@ pushd catch2
 wget -O - https://www.mpich.org/static/downloads/4.2.2/mpich-4.2.2.tar.gz | tar xvz --strip-components 1
 mkdir -p build
 pushd build
-../configure --prefix=/usr --without-pmix
+../configure --prefix=/usr --without-pmix --disable-fortran
 make -j 4 install
 popd
 popd


wihobbs (changes requested):
Thanks @ilumsden! One thing to note is that we [recently updated](https://github.com/flux-framework/rfc/pull/450) our [commit guidelines in Flux](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_1.html#patch-requirements). We've always required commit bodies in practice but now it's in the RFC. 

Could you update your commit message to something like:

> docker: remove libmpich-dev from noble build
> 
> Problem: libmpich-dev in the ubuntu noble image builds with PMIx support, but still uses the Hydra process manager. This causes issues when trying to run MPI code because Hydra does not provide a PMIx interface.
> 
> Solution: remove libmpich-dev from the Ubuntu Noble Docker image. Add installs of gfortran because the script to build MPICH from source was failing due to not being able to find a Fortran compiler.

ilumsden:
> Thanks @ilumsden! One thing to note is that we [recently updated](https://github.com/flux-framework/rfc/pull/450) our [commit guidelines in Flux](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_1.html#patch-requirements). We've always required commit bodies in practice but now it's in the RFC.
> 
> Could you update your commit message to something like:
> 
> > docker: remove libmpich-dev from noble build
> > Problem: libmpich-dev in the ubuntu noble image builds with PMIx support, but still uses the Hydra process manager. This causes issues when trying to run MPI code because Hydra does not provide a PMIx interface.
> > Solution: remove libmpich-dev from the Ubuntu Noble Docker image. Add installs of gfortran because the script to build MPICH from source was failing due to not being able to find a Fortran compiler.

Yep. I'll do that now. I honestly forgot all about that.

wihobbs:
As for the actual change, looks good to me but we should probably get an approval from @garlick.

grondo:
This LGTM. I tested with these changes locally by building the new image and running all tests under `docker-run-checks -i noble`. Since that worked, I went ahead and pushed the result to `fluxrm/testenv:noble`.

I wonder if there is some way we can add a test to our CI to catch this issue going forward?

src/test/docker/noble/Dockerfile
```diff
@@ -39,14 +39,17 @@ RUN apt-get update \
...
         clang-18 \
         clang-tools-18 \
 # default version
         gcc-13 \
         g++-13 \
+        gfortran-13 \
 # newest
         gcc-14 \
         g++-14 \
+        gfortran-14 \
```
garlick: Removing libmpich-dev makes sense!

> Also, add installs of gfortran to prevent configure errors when building MPICH from source.

Weren't we always buliding mpich from source in this image?  How was that working without fortran?
We could configure mpich with `--disable-fortran` in the build script rather than pull in (two versions of?)  the fortran tool chain.

ilumsden: We could definitely just disable Fortran if that's preferred. I have no idea how it was working before. I just know that, when I built the image myself before making any changes, the build failed because it couldn't find Fortran.

ilumsden: Just pushed a commit that disables Fortran

grondo: > Weren't we always buliding mpich from source in this image? How was that working without fortran?
We could configure mpich with --disable-fortran in the build script rather than pull in (two versions of?) the fortran tool chain.

One worry is that someone will come along and want to build or run a test that uses the MPICH fortran and we'll get another issue report that it doesn't work. I've already built the `fluxrm/testenv:noble` image with the gfortran included and pushed that to docker hub, so @ilumsden I wonder if you could try your experiments against the latest `noble` image?

We shouldn't be too worried about the size of the noble image here since it already pulls in _a lot_ of other stuff...

garlick: Great point.  Sorry for that diversion!  Back to the original OK @ilumsden ?

grondo:
@wihobbs, can you re-review this to clear your requested changes and we can get this merged?

grondo:
Thanks @wihobbs!

wihobbs:
@grondo Looks like the el9 test timed out while running (30m), should we just re-run it?

grondo:
Yeah, I'll just "restart failing checks". This happens pretty often since we have a couple flaky tests.

Issue #6909, opened 2025-07-11
Flux brokers cannot fully connect after upgrade, flux resource status provides information but all of the nodes are showing x


mjaguil:
We upgrade Flux and now all of the Active nodes are showing x and are red, as if they are unavalable.

grondo:
Some things to check:

 * Are the brokers started on the non-leader nodes? If you are running with systemd then `systemctl status flux`
 * If the brokers aren't starting, check the logs (usually included in `systemctl status` or `journalctl -u flux`
 * check the output of `flux dmesg -H` on the leader broker (rank 0) and see if connections from the clients are being rejected and why (e.g. this could be because the client nodes are on an older version of Flux)

grondo:
Oh, another command that may be useful is `flux overlay errors`.

grondo:
To summarize offline debugging, it was discovered via running `rpm --verify` that `/etc/flux/rc1` and `/etc/flux/rc3` had been modified by configuration management and thus the system was running with v0.74.0 with the rc scripts from `v0.71.0`. This caused the `sdmon` modules to not be loaded, so the instance was stuck waiting for `sdmon` to check in.

Moving the custom rc code to files in `/etc/flux/rc{1,3}.d` and replacing the rc files with the originals should resolve the issue.

PR #6911, opened 2025-07-11
flux-module trace: capture early messages

garlick:
Problem: a module has to be loaded before you can initiate  a trace on it, so early message are missed.

If a module trace request is received with no module name,  trace messages communicated over the broker end of the module  sockets.

This was a bit hasty so I'll mark it WIP until I've had a chance to review it more thoroughly.

### DIFF
--- a/src/broker/modhash.c
+++ b/src/broker/modhash.c
@@ -25,6 +25,7 @@
 
 #include "module.h"
 #include "broker.h"
+#include "trace.h"
 #include "modhash.h"
 #include "overlay.h"
 
@@ -32,6 +33,7 @@ struct modhash {
     zhash_t *zh_byuuid;
     flux_msg_handler_t **handlers;
     struct broker *ctx;
+    struct flux_msglist *trace_requests;
 };
 
 static json_t *modhash_get_modlist (modhash_t *mh,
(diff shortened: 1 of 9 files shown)


garlick:
This was put together during debugging of #6909

garlick:
I reworked this today as the previous try missed some messages.  This implementation required minor broker refactoring, but I think the refactoring improved things a bit.   I think this is ready for a review.

garlick:
Ugh, oh sorry, I've got a segfault to run down.   Hold the review for now.

garlick:
OK, fixed that.

src/broker/modhash.c
```diff
@@ -789,6 +789,51 @@ module_t *modhash_next (modhash_t *mh)
...
+    if (!(p = modhash_lookup (mh, sender))) {
+        errprintf (error, "requestor is not local");
+        errno = ENOENT;
+        return -1;
+    }
+    if (service_add (ctx->services, name, sender, mod_svc_cb, p) < 0) {
+        errprintf (error,
+                   "could not register service %s for module %s",
+                   name,
+                   module_get_name (p));
```
grondo: should the human readable error here include `strerror (errno)` to differentiate between EINVAL and EEXIST, etc?

garlick:
Fixed that error message and I'll set MWP.  Thanks!

PR #6914, opened 2025-07-16
make broker groups easier to look at when debugging

garlick:
Problem: while debugging #6909, it might have been handy to be able to look directly at the broker groups memberships, but there is no tool for that.

Just add a `stats-get` handler so `flux module stats groups` can show the group names and membership at the current broker's subtree.  Seeing that a group named `sdmon.online` was missing some ranks may have prompted earlier checking on the state of the  `sdmon` module on the offline nodes.

### DIFF
--- a/src/broker/groups.c
+++ b/src/broker/groups.c
@@ -602,6 +602,65 @@ error:
     }
 }
 
+static json_t *obj_from_idset (const struct idset *ids)
+{
+    char *s;
+    json_t *o = NULL;
+
+    if ((s = idset_encode (ids, IDSET_FLAG_RANGE)))
+        o = json_string (s);
+    free (s);
+    return o;
+}
+
+static json_t *make_groups_object (struct groups *g)
+{
+    json_t *obj;
+    struct group *group;
+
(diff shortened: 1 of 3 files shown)


grondo (approved):
LGTM! 

We could now perhaps fetch the groups stats from `flux resource status` and display something for targets that are offline due to sdmon...

garlick:
> We could now perhaps fetch the groups stats from flux resource status and display something for targets that are offline due to sdmon...

True although I"m a little embarrassed by the design used to have sdmon influence online state here:  [monitor_create()](https://github.com/flux-framework/flux-core/blob/master/src/modules/resource/monitor.c#L454-L470).  The juggling of the two group names would need to be replicated in `flux resource status`.  It may be worth either simplifying the design or adding something to the `resource.status` response.

garlick:
I'll go ahead and set MWP here though - thanks!

grondo:
> The juggling of the two group names would need to be replicated in flux resource status. It may be worth either simplifying the design or adding something to the resource.status response.

Adding something to the response would be ideal! Want to open an issue on that?

PR #6917, opened 2025-07-17
broker: reduce the number of mallocs in logging path

garlick:
Problem: the broker ring buffer implementation is malloc-heavy.

This converts the logging ring buffer from `zlist_t` to a CCAN list so that a list node doesn't need to be allocated each time an entry is appended to the ring.  It also combines two mallocs used to allocate a list entry into one.

This is probably not significant since our log volume is not too crazy, but it was an easy change, so there.

### DIFF
--- a/src/broker/log.c
+++ b/src/broker/log.c
@@ -64,11 +64,7 @@ void logbuf_destroy (logbuf_t *logbuf);
 
 static void logbuf_entry_destroy (struct logbuf_entry *e)
 {
-    if (e) {
-        if (e->buf)
-            free (e->buf);
-        free (e);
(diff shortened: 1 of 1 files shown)


garlick:
> I didn't do any real testing, but a flux dmesg on this branch seems so much faster now! 

Huh!  I didn't expect that!

Sorry for the force pushes: one sloppy error, one pedantic compiler warning we always hit with list_for_each.

grondo:
> Huh! I didn't expect that!

Well I should have added a ;-)

garlick:
OK, setting MWP!

PR #6918, opened 2025-07-17
broker: preserve log levels in system instance

garlick:
Problem: all Flux logs go to the systemd journal (via stderr) at LOG_INFO level, which means they cannot be usefully filtered using the [journalctl(1)](https://www.man7.org/linux/man-pages/man1/journalctl.1.html) `-p` option.

When `log-stderr-mode=local` is selected, use a <level> prefix that can be interpreted by systemd instead of embedding the level name in the log message.  For example:
```
$ flux logger --severity=err help
```
before:
```
logger.err[0]: help
```
after:
```
<3>logger[0]: help
```
Edit: as a bonus, `journalctl` output for flux is now properly colorized by severity.

### DIFF
--- a/src/broker/log.c
+++ b/src/broker/log.c
@@ -28,7 +28,7 @@
 #include "log.h"
 
 typedef enum { MODE_LEADER, MODE_LOCAL } stderr_mode_t;
-typedef enum { LOG_NO_TIMESTAMP=1 } log_flags_t;
+typedef enum { LOG_FOR_SYSTEMD=1 } log_flags_t;
 
 /* See descriptions in flux-broker-attributes(7) */
 static const int default_ring_size = 1024;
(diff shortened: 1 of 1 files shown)


chu11 (approved):
LGTM!

grondo:
Is there a reason to _replace_ the log level string with the prefix, instead of just adding the prefix for systemd's benefit?
AFAIK, the `<N>` prefix is dropped when log messages go from the systemd journal to `/var/log/messages` or other log message destinations. When debugging, I have found it useful to grep these files for things like `broker.err` to find error messages when I don't know what I'm looking for. Unless I've missed something, that will no longer be possible after this change?

garlick:
Oh, you are right, for some reason I was assuming the syslog severity levels landed in our log files but they do not.  (I guess I also assumed that if systemd knows the log level, it would preserve it when forwarding to syslog, although I didn't check that).

Sure, I will add that back in.

garlick:
Thanks, I'll set MWP

Issue #5952, opened 2024-05-09
write job statistics to guest namespace on instance exit


garlick:
Problem: we don't have much information about what happens in Flux instances launched by the system instance, unlike slurm which tracks job steps centrally.

Following up on a discussion in today's meeting about finding out what sort of "workflows" people are running:  Could we have a flux instance write a summary object of some sort to the guest KVS directory in its enclosing instance?

A tool could summarize those objects at the system instance level.

For extra credit they could be made hierarchical, e.g. the object would include a summary of the objects written by sub-instances in their KVS.

FWIW, [RFC 16](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_16.html) governs the content of the job KVS directory.

grondo:
Good idea! One benefit of collecting that data periodically is that it would protect against instances that get killed due to timeout, exception etc.

A benefit of using annotations instead of a one time object in the kvs is that there's already a tool to present annotations at the instance level (`flux jobs`) and they are designed to be updated (as far as I recall) so you could get runtime stats as well. Once in the job-list database the data can be queried with `flux jobs` or whatever standard query tool we use instead of a custom tool.

The one-time summary in the kvs is appealing, however, because it is much simpler and it's easier to think about how to collect hierarchically as you suggested.

grondo:
This is actually going to be quite important on systems where most or all jobs run at the system instance level are batch or alloc jobs. This is because most of the detail of the actual jobs run by the user is currently lost when the batch job exits, unless the user thought to use `--dump` on the command line.

For example it would be nice to be able to use `flux job taskmap` on a parallel job run within a batch job, or even be able to access job eventlogs. Perhaps we can persist some of this information in the KVS in such a way that some of these commands can work naturally.

grondo:
@kkier brought this up again on today's coffee call.

The initial use case is just determining what _precisely_ was running on a given node at a time when some event occurs. Currently, sysadmins only know a given batch job was running at the time, but not what job within that job was running.

`--dump` as proposed in #6080 isn't a great solution here since having dumpfiles sprinkled through the fileystem isn't useful to the admins.

I wonder if a feature similar to `--dump`, but which writes to the guest namespace would be feasible? We could drop job output and just capture eventlogs and R. Maybe this information could be preserved as it occurs.

PR #6922, opened 2025-07-18
broker: add syslog capability

garlick:
Problem: sys admins may want to capture logs from batch  or alloc subinstances.

This adds basic support for copying broker logs to syslog.  For now, this is only configurable with broker attributes.  Two attributes are added:
 ```
log-syslog-enable=1
log-syslog-level=2
```
The log entries are prefixed with "username@job-path".  Here are a couple of example logs from a test instance and a batch job run within that test instance:
```
Jul 17 17:25:30 system76-pc flux[983068]: garlick@/ job-manager.debug[0]: scheduler: hello +partial-ok
Jul 17 17:25:44 system76-pc flux[983757]: garlick@/ƒ3pBZx4P broker.debug[0]: rmmod barrier
```
Syslog is disabled by default of course.  I was thinking maybe we could add a toml config option that would tell the system instance to enable it for batch/alloc instances (only).  Then we could perhaps start logging summary info about jobs run within the instance as a simple alternative to
- #5952

This was a quick hack with no tests yet.  I wasn't sure if it was going to be all that useful, so I thought I'd pause here and ask for feedback on the approach.

### DIFF
--- a/doc/man7/flux-broker-attributes.rst
+++ b/doc/man7/flux-broker-attributes.rst
@@ -299,14 +299,13 @@ log-ring-size [Updates: C, R]
    Default: ``1024``.
 
 log-forward-level [Updates: C, R]
-   Log entries at :linux:man3:`syslog` level at or below this value
-   are forwarded to rank zero for permanent capture.  Default ``7``
-   (LOG_DEBUG).
+   Log entries of numerical severity level less than or equal to this value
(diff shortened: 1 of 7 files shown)


garlick:
Adding some tests uncovered poor error checking in the log attributes, so that was useful!

I'll go ahead and set MWP.

PR #6925, opened 2025-07-21
macos port: use getifaddrs to find default if

trws:
I decided to try out the MacOS build at the end of last week, and noticed one bit of low-hanging fruit.  The `getprimary_iface4` function was failing, and I guessed that could cause a whole lot of downstream failures. 

problem: MacOS/BSD lack a linux-compatible proc (usually) so we can't read routes from proc

solution: Use the same approach as before, looking for the first listed interface with a netmask == 0, but iterate over the entries in getifaddrs instead, which is a BSD interface ported to linux and available on all relevant systems.  If at some point we do switch to libuv, it provides a very similar convenience wrapper that is available on all compatible platforms as well.


trws:
It turns out this is not reliable in some cases, where no interface with a blank netmask exists but a default still does, working on a fix.

trws:
There's a failure happening on both PRs related to this from the fileref test. I haven't been able to reproduce it on a local mac, and it doesn't seem to have cropped up before. @garlick have you seen this? It looks like a symlink test failure, but there's no related output from GHA to see what actually failed other than that.

grondo:
Does the tmate action work on the macos builder? I've used that in the past to debug issues that annoyingly only happen in CI. I just usually temporarily add the action on failure. (I Ikeep wondering if there's a way to permanently add that action and only trigger it via a "debug" rebuild)

trws:
That is an excellent thought. I don't know, but I will find out.

garlick:
Apologies, I made a comment on this stuff in the other PR that is based on top of it.  Just that I'm traveling today and wanted to have a closer look at this before it goes in.  Great work @trws.

src/common/libutil/ipaddr.c
```diff
@@ -30,6 +30,10 @@
 #include "cidr.h"
 #include "ipaddr.h"
 
+#ifdef __MACH__
```
garlick: I used `__APPLE__` before (e.g. in libev and libutil/intree.c).   Unless there's a reason to use `__MACH__` maybe  we should stick to `__APPLE__` for consistency?

src/common/libutil/ipaddr.c
```diff
@@ -44,6 +48,44 @@ static bool is_linklocal6 (struct sockaddr_in6 *sin)
     return false;
 }
 
+#ifdef __MACH__
+static int getprimary_iface4 (char *buf, size_t size, flux_error_t *error)
+{
+    int ret = -1;
+    SCDynamicStoreRef store = SCDynamicStoreCreate(NULL, CFSTR("GetDefaultInterface"), NULL, NULL);
+    if (!store) {
+        fprintf(stderr, "Failed to create SCDynamicStore\n");
```
garlick: Put error messages in `error` using `errprintf()` instead of sending them to stderr.

src/common/libutil/ipaddr.c
```diff
@@ -234,11 +277,16 @@ int ipaddr_getprimary (char *buf,
...
+        // macos doesn't reliably resolve its own hostname, so always do
+        // getprimary_ifaddr
+#ifdef __MACH__
+        rc = getprimary_ifaddr (buf, len, prefer_family, error);
+#else
         if (!(flags & IPADDR_HOSTNAME))
             rc = getprimary_ifaddr (buf, len, prefer_family, error);
-        if (rc < 0) {
+#endif
+        if (rc < 0)
```
garlick: This might be slightly clearer if the ifdef just cleared IPADDR_HOSTNAME at the top rather than mixing preprocessor conditionals with C conditionals, e.g.
```c
#ifdef __APPLE__
    // macos doesn't reliably resolve its own hostname
    flags &= ~IPADDR_HOSTNAME;
#endif
```

src/common/libutil/Makefile.am
```diff
@@ -109,6 +110,12 @@ libutil_la_SOURCES = \
 	basename.c \
 	ansi_color.h
 
+if MAC
+libutil_la_LDFLAGS = -F/System/Library/Frameworks \
+									 -framework CoreServices \
+									 -framework SystemConfiguration
+endif
```
garlick: I don't think the -F is needed since `/System/Library/Frameworks` is the default location?

garlick:
Hmm one of my review comments didn't seem to have taken?  I was thinking maybe it would be more idiomatic with respect to autotools and scale a little better if we conditionally defined some ld flags that are empty on linux and filled on apple, rather than do the `if MAC` thing in Makefile.ams.  For example, in configure.ac
```m4
AC_MSG_CHECKING([whether Apple frameworks can be used])
AS_IF([test $(uname) = "Darwin"],[
  AC_MSG_RESULT([yes])
  apple_coreservices_framework="-framework CoreServices"
  AS_VAR_SET(APPLE_CORESERVICES_FRAMEWORK, $apple_coreservices_framework)
  AC_SUBST(APPLE_CORESERVICES_FRAMEWORK)
  apple_systemconfiguration_framework="-framework SystemConfiguration"
  AS_VAR_SET(APPLE_SYSTEMCONFIGURATION_FRAMEWORK, $apple_systemconfiguration_framework)
  AC_SUBST(APPLE_SYSTEMCONFIGURATION_FRAMEWORK)
],[
  AC_MSG_RESULT([no])
])
```
Then in libuti/Makefile.am
```
libutil_la_LDFLAGS = \ 
        $(AM_LDFLAGS) \
        $(APPLE_CORESERVICES_FRAMEWORK) \
        $(APPLE_SYSTEMCONFIGURATION_FRAMEWORK)
```

src/common/libutil/ipaddr.c
trws: Works for me.

trws:
Thanks for the autoconf changes, I can see why that's nicer, but I'm pretty much always in trial and error mode when I'm dealing with things that have to cross from autoconf m4 land to shell to perl to make.

src/common/libutil/ipaddr.c
```diff
@@ -234,11 +277,14 @@ int ipaddr_getprimary (char *buf,
...
+            // macos doesn't reliably resolve its own hostname
+            flags &= ~IPADDR_HOSTNAME;
+        #endif
         if (!(flags & IPADDR_HOSTNAME))
             rc = getprimary_ifaddr (buf, len, prefer_family, error);
-        if (rc < 0) {
+        if (rc < 0)
             rc = getprimary_hostaddr (buf, len, prefer_family, error);
-        }
     }
```
garlick: Unrelated fiddling with braces.

src/common/libutil/ipaddr.c
```diff
@@ -44,6 +48,44 @@ static bool is_linklocal6 (struct sockaddr_in6 *sin)
     return false;
 }
 
+#ifdef __APPLE__
+static int getprimary_iface4 (char *buf, size_t size, flux_error_t *error)
+{
+    int ret = -1;
+    SCDynamicStoreRef store = SCDynamicStoreCreate(NULL, CFSTR("GetDefaultInterface"), NULL, NULL);
```
garlick: RFC 7 coding style: use a space between function name and open paren.

configure.ac
```diff
@@ -661,4 +674,3 @@ AS_IF([test "x$enable_docs" != "xno"], [
     AC_MSG_WARN([Python Docutils not found. Manual pages will not be generated.])
   fi
 ])
-
```
garlick: unrelated whitespace change

src/common/libutil/Makefile.am
```diff
@@ -7,6 +7,7 @@ AM_CFLAGS = \
 AM_LDFLAGS = \
 	$(CODE_COVERAGE_LIBS)
 
+
```
garlick: unrelated whitespace change

garlick:
This unit test is failing in CI for macos
```
FAIL: test_fileref.t 15 - fileref_create symlink works
```
It doesn't fail on my laptop.  The test looks like this:
```
void test_link (void)
{
    json_t *o;
    const char *target = "/a/b/c/d/e/f/g";
    bool rc;

    if (symlink (target, mkpath ("testlink")) < 0)
        BAIL_OUT ("could not create test symlink");
    o = xfileref_create (mkpath ("testlink"));
    rc = check_fileref (o, "testlink", 0);
    ok (rc == true,
        "fileref_create symlink works");
    json_decref (o);
    rmfile ("testlink");
}
```
The clues as to why the `check_fileref()` failed are going to be in the TAP diag output.  Not sure what would be different in CI compared to laptop or why suddenly this is failing now.

garlick:
Tmate supposedly works on macos according to https://github.com/mxschmitt/action-tmate, but I've been unable to get it to do anything with this patch applied
```diff
diff --git a/.github/workflows/main.yml b/.github/workflows/main.yml
index c65ea7f08..5e0997676 100644
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -130,6 +130,8 @@ jobs:
       run: make check -j4 TESTS=
     - name: check what works so far
       run: scripts/check-macos.sh
+    - name: setup tmate session
+      uses: mxschmitt/action-tmate@v3
 
   generate-matrix:
     # https://stackoverflow.com/questions/59977364
```
It appears in my workflow output but doesn't look like it executes.  (gray checkbox, 0s elapsed, workflow continues on)

garlick:
I tried adding this which appears to work on my mac laptop
```diff
diff --git a/scripts/install-deps-macos.sh b/scripts/install-deps-macos.sh
index 0ae688fc2..bba9a78e7 100755
--- a/scripts/install-deps-macos.sh
+++ b/scripts/install-deps-macos.sh
@@ -28,7 +28,10 @@ brew install \
   python3 \
   cffi \
   libyaml \
-  jq
+  jq \
+  tmux \
+  reattach-to-user-namespace \
+  tmate
 
```
but no luck

image

grondo:
> It appears in my workflow output but doesn't look like it executes. (gray checkbox, 0s elapsed, workflow continues on)

I think this is because the previous step fails, so the tmate action is being skipped. Try adding `if: failure()` to the step:

```diff
diff --git a/.github/workflows/main.yml b/.github/workflows/main.yml
index c65ea7f08..6b416da00 100644
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -130,6 +130,9 @@ jobs:
       run: make check -j4 TESTS=
     - name: check what works so far
       run: scripts/check-macos.sh
+    - name: tmate
+      if: failure()
+      uses: mxschmitt/action-tmate@v3
 
   generate-matrix:
     # https://stackoverflow.com/questions/59977364
```

trws:
Thanks for the tmate work, going to try that now that I have the source cleaned up. Also going to see if I can get the annotate failures script (or part of it) to go so we get the debug info more reliably.

garlick:
Thanks @grondo for the tmate hint.  I got this when running the test manually in the CI environment:
```
# symlink target is wrong
not ok 15 - fileref_create symlink works
#   Failed test 'fileref_create symlink works'
#   at test/fileref.c line 552.
```
and adding a little debug:
```
# symlink target is wrong /a/b/c/d/e/f/gC<A8>^XҌz^B != /a/b/c/d/e/f/g
not ok 15 - fileref_create symlink works
```
Ah, readlink(2) is being used on an uninitialized buffer in the test, and readlink doesn't NULL terminate.
This fixed it for me.
```diff
diff --git a/src/common/libfilemap/test/fileref.c b/src/common/libfilemap/test/filer
ef.c
index 468fa8119..1afeffd26 100644
--- a/src/common/libfilemap/test/fileref.c
+++ b/src/common/libfilemap/test/fileref.c
@@ -304,6 +304,7 @@ bool check_fileref (json_t *fileref, const char *name, int blobc
ount)
             diag ("symlink encoding/size unexpectedly set");
             goto error;
         }
+        memset (buf, 0, sizeof (buf));
         if (readlink (mkpath (name), buf, sizeof (buf)) < 0
             || !streq (buf, json_string_value (data))) {
             diag ("symlink target is wrong");
```

PR #6926, opened 2025-07-21
Macos fixes

trws:
This is a couple more macos fixes on top of #6925. These are much smaller, dealing with missing defines or temp file creation.  After these and the other one though, everything in src seems to pass except libsubprocess.

### DIFF
--- a/src/modules/resource/test/rutil.c
+++ b/src/modules/resource/test/rutil.c
@@ -157,7 +157,7 @@ static char *create_tmp_file (const char *content)
         tmpdir = "/tmp";
     if (asprintf (&path, "%s/rutil-test.XXXXXX", tmpdir) < 0)
         BAIL_OUT ("error allocating buffer");
-    if ((fd = mkostemp (path, O_WRONLY)) < 0)
+    if ((fd = mkostemp (path, 0)) < 0)
         BAIL_OUT ("error creating temp file");
 
     cleanup_push_string (cleanup_file, path);


trws:
Good point @chu11, it's reworded and rebased to match the PR it's stacked on.

src/modules/job-manager/kill.c
```diff
@@ -38,7 +38,11 @@
 #include "job-manager.h"
 
 #ifndef SIGRTMAX
```
garlick: Or just use NSIG.

That's what I did before in d51397b3ecfd13a1e93ad50a40f80c1d5cc20bb0

garlick (commented):
I feel like the ipaddr stuff needs a closer review to see whether the use of stderr is appropriate and whether the configure/ifdef logic is as simple as it can be, but I'm traveling today and probably won't have time for a bit.  If this can wait I'll try to take a look this weekend.

trws:
> whether the use of stderr is appropriate 

This is what I get for not completely rewriting the ChatGPT code for that function.  I rewrote the CString handling to be pure C but forgot the error prints, will definitely fix.

> and whether the configure/ifdef logic is as simple as it can be

I was going for something like what the linux kernel does, make it a simple selection of one complete thing or another complete thing, though I didn't do it in a header like they require.  Would be happy to rework that if you prefer something else, mainly wanted to avoid needing to have OpenSTEP libs for CoreFoundation on linux (much as I actually really like their string class).

src/modules/job-manager/kill.c
trws: That's fair, I haven't found a system that lacks it yet.  I'll make that change.

trws:
Oh and it can definitely wait, there's a heisenbug in fileref of all things on GHA right now that's going to take a bit to work out.  Thanks for looking it over!

garlick:
> I was going for something like what the linux kernel does, make it a simple selection of one complete thing or another complete thing

No, I agree with that!  I was thinking about other things like using platform conditionals vs feature tests, and trying to contain the conditional logic in configure.ac vs spreading it out in Makefile.am.   I did make a comment related to the latter over in the PR where the code is proposed.  On the former, I think maybe completely eschewing platform conditionals is overkill at least for now.

garlick:
@mergifyio requeue

PR #6928, opened 2025-07-22
t: clean up use of jq

chu11:
Per a comment in #6798, some misc cleanups

### DIFF
--- a/t/t0002-request.t
+++ b/t/t0002-request.t
@@ -106,14 +106,14 @@ test_expect_success 'request: rpc test client works with no request payload' '
 		test -s attr.list.out
 '
 test_expect_success 'request: rpc test client works with request payload' '
-	$jq -j -c -n  "{name:\"rank\"}" | $RPC attr.get >attr.get.out &&
+	jq -j -c -n  "{name:\"rank\"}" | $RPC attr.get >attr.get.out &&
 		test -s attr.get.out
 '
 test_expect_success 'request: rpc test client handles expected failure' '
 	$RPC attr.get 71 </dev/null
 '
 test_expect_success 'request: rpc test client handles expected failure other than EPROTO' '
-	$jq -j -c -n  "{name:\"noexist\"}" | $RPC attr.get 2
+	jq -j -c -n  "{name:\"noexist\"}" | $RPC attr.get 2
 '
 test_expect_success 'request: rpc test client handles unexpected failure' '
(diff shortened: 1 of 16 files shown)


t/t0002-request.t
```diff
@@ -106,14 +106,14 @@ test_expect_success 'request: rpc test client works with no request payload' '
 		test -s attr.list.out
 '
 test_expect_success 'request: rpc test client works with request payload' '
-	$jq -j -c -n  "{name:\"rank\"}" | $RPC attr.get >attr.get.out &&
```
grondo, line 109: Tests are using the `$jq` variable because it is set in `flux-sharness.sh`:

https://github.com/flux-framework/flux-core/blob/56c98db68acee51db413ecf9d53f2e4588bf378b/t/sharness.d/flux-sharness.sh#L353

It does seem like a good idea to consistently use `jq` directly instead of `$jq` because I'm sure lots of places already weren't, but the line above where `$jq` is set should also be dropped.

t/t0002-request.t
```diff
@@ -106,14 +106,14 @@ test_expect_success 'request: rpc test client works with no request payload' '
...
+	jq -j -c -n  "{name:\"rank\"}" | $RPC attr.get >attr.get.out &&
 		test -s attr.get.out
 '
 test_expect_success 'request: rpc test client handles expected failure' '
 	$RPC attr.get 71 </dev/null
 '
 test_expect_success 'request: rpc test client handles expected failure other than EPROTO' '
-	$jq -j -c -n  "{name:\"noexist\"}" | $RPC attr.get 2
```
grondo, line 116: Commit message typo: restult

Also the statement that tests use `$jq` due to when `jq` was not a requirement isn't strictly true, it is just sometimes useful to look up the path to a program once when you know it will be called many times. However, it seems silly for these tests so I agree with the general approach. However, for a more accurate commit message, I'd say something more along the lines of

> Some tests execute `Jq(1)` via the `$jq` variable set in `t/sharness.d/flux-sharness.sh`, while many just execute `jq` from `PATH` directly. For consistency and to match the use of most other programs in sharness tests, drop the use of the `$jq` variable.

grondo (approved):
LGTM. The comments in the first commit should probably be addressed before merging (but they're minor)

chu11:
thanks, tweaked the commit message a bit based on your suggestion and removed the setting of `$jq` in sharness.

PR #6929, opened 2025-07-23
More macos fixes

trws:
This is stacked on top of #6926, and contains all changes necessary to get all tests in `src` to pass.  I don't have the GHA updates in here yet, but that's coming next.

### DIFF
--- a/src/broker/test/overlay.c
+++ b/src/broker/test/overlay.c
@@ -30,6 +30,8 @@
 #include "src/broker/attr.h"
 #include "src/broker/topology.h"
 
+#include "src/common/libtestutil/test_file.h"
+
 static zlist_t *logs;
 void *zctx;
 
@@ -93,9 +95,6 @@ struct context *ctx_create (flux_t *h,
 {
     struct context *ctx;
     flux_error_t error;
-    const char *temp = getenv ("TMPDIR");
-    if (!temp)
-        temp = "/tmp";
 
     if (!(ctx = calloc (1, sizeof (*ctx))))
         BAIL_OUT ("calloc failed");
@@ -273,7 +272,7 @@ void trio (flux_t *h)
 {
     struct context *ctx[2];
     int size = 3;
-    char parent_uri[64];
+    char parent_uri[PATH_MAX + 64], uri[PATH_MAX + 64];
     const char *server_pubkey;
     const char *client_pubkey;
     const char *tmp;
@@ -293,7 +292,7 @@ void trio (flux_t *h)
     ok ((server_pubkey = overlay_cert_pubkey (ctx[0]->ov)) != NULL,
         "%s: overlay_cert_pubkey works", ctx[0]->name);
 
-    snprintf (parent_uri, sizeof (parent_uri), "ipc://@%s", ctx[0]->name);
+    snprintf (parent_uri, sizeof (parent_uri), "ipc://%s/flux_ipc_%s", get_test_dir (), ctx[0]->name);
     ok (overlay_bind (ctx[0]->ov, parent_uri, NULL) == 0,
         "%s: overlay_bind %s works", ctx[0]->name, parent_uri);
 
@@ -441,7 +440,9 @@ void trio (flux_t *h)
      * fails to initialize because its endpoint is already bound.
      */
     errno = 0;
-    ok (overlay_bind (ctx[1]->ov, "ipc://@foo", NULL) < 0
+    if (snprintf (uri, sizeof uri, "ipc://%s/flux_ipc_foo", get_test_dir ()) < 0)
+        BAIL_OUT("asprintf failed");
+    ok (overlay_bind (ctx[1]->ov, uri, NULL) < 0
         && errno == EADDRINUSE,
         "%s: second overlay_bind in proc fails with EADDRINUSE", ctx[0]->name);
 
@@ -502,6 +503,13 @@ void trio (flux_t *h)
     zmq_close (zsock_none);
     zmq_close (zsock_curve);
 
+    // ENOENT is acceptable because in some cases the ipc node is already
+    // cleaned up here
+    if (unlink (parent_uri) < 0 && errno != ENOENT)
+        BAIL_OUT ("could not remove %s", parent_uri);
+    if (unlink (uri) < 0 && errno != ENOENT)
+        BAIL_OUT ("could not remove %s", uri);
+
     ctx_destroy (ctx[1]);
     ctx_destroy (ctx[0]);
 }
@@ -510,7 +518,7 @@ void test_create (flux_t *h,
                   int size,
                   struct context *ctx[])
 {
-    char uri[64] = { 0 };
+    char uri[PATH_MAX + 200] = { 0 };
     int rank;
 
     for (rank = 0; rank < size; rank++) {
@@ -518,7 +526,7 @@ void test_create (flux_t *h,
         if (overlay_set_topology (ctx[rank]->ov, ctx[rank]->topo) < 0)
             BAIL_OUT ("%s: overlay_set_topology failed", ctx[rank]->name);
         if (rank == 0) {
-            snprintf (uri, sizeof (uri), "ipc://@%s", ctx[0]->name);
+            snprintf (uri, sizeof (uri), "ipc://%s/flux_ipc_%s", get_test_dir (), ctx[0]->name);
             /* Call overlay_bind() before overlay_authorize() is called
              * for the other ranks, since overlay_bind() creates the ZAP
              * handler, and overlay_authorize() will fail if it doesn't
@@ -544,7 +552,12 @@ void test_create (flux_t *h,
 
 void test_destroy (int size, struct context *ctx[])
 {
+    char uri[PATH_MAX + 200] = { 0 };
     int rank;
+
+    snprintf (uri, sizeof (uri), "ipc://%s/flux_ipc_%s", get_test_dir (), ctx[0]->name);
+    if (unlink (uri) < 0 && errno != ENOENT)
+        BAIL_OUT ("could not remove %s", uri);
 
     for (rank = 0; rank < size; rank++)
         ctx_destroy (ctx[rank]);
@@ -670,8 +683,16 @@ void wrongness (flux_t *h)
         BAIL_OUT ("overlay_create failed");
 
     errno = 0;
-    ok (overlay_bind (ov, "ipc://@foobar", NULL) < 0 && errno == EINVAL,
+    struct context *ctx = ctx_create (h, 1, 0, "kary:2", NULL);
+    char * uri = NULL;
+    if (asprintf (&uri, "ipc://%s/flux_ipc_foobar", get_test_dir ()) < 0)
+        BAIL_OUT ("asprintf failed");
+    ok (overlay_bind (ov, uri, NULL) < 0 && errno == EINVAL,
         "overlay_bind fails if called before rank is known");
+    ctx_destroy(ctx);
+    if (unlink (uri) < 0 && errno != ENOENT)
+        BAIL_OUT ("could not remove %s", uri);
+    free (uri);
 
     ok (!flux_msg_is_local (NULL),
         "flux_msg_is_local (NULL) returns false");
--- a/src/cmd/flux-terminus.c
+++ b/src/cmd/flux-terminus.c
@@ -372,7 +372,7 @@ static int start_service_daemon (flux_t *orig_h, optparse_t *p)
     }
 
     /*  Wait for child, wait for grandchild to close pipe */
-    waitpid (pid, NULL, 0);
+    while (waitpid (pid, NULL, 0) < 0 && errno == EINTR) {}
     close (pfds[1]);
 
     /*  Read status from grandchild */
--- a/src/common/librlist/rhwloc.c
+++ b/src/common/librlist/rhwloc.c
@@ -82,6 +82,10 @@ error:
 
 static int topo_restrict (hwloc_topology_t topo)
 {
+    // this is not supported on macos, and actually flat fails
+#ifdef __APPLE__
+    return 0;
+#else
     hwloc_bitmap_t set = NULL;
     int rc = -1;
     if (!(set = hwloc_bitmap_alloc ())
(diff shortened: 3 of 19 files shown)


src/broker/test/overlay.c
```diff
@@ -293,7 +293,7 @@ void trio (flux_t *h)
     ok ((server_pubkey = overlay_cert_pubkey (ctx[0]->ov)) != NULL,
         "%s: overlay_cert_pubkey works", ctx[0]->name);
 
-    snprintf (parent_uri, sizeof (parent_uri), "ipc://@%s", ctx[0]->name);
+    snprintf (parent_uri, sizeof (parent_uri), "ipc:///tmp/flux_ipc_%s", ctx[0]->name);
```
garlick: Not completely sure, but using the same fs path might introduce some fragility when multiple users run the tests, even serially if the socket is not cleaned up for some reason?

Suggestion:  put the sockets in a test specific tmpdir.  If you like, borrow `create_test_dir()` from `tests/boot_config.c`

src/common/librlist/rhwloc.c
```diff
@@ -82,6 +82,10 @@ static int init_topo_from_xml (hwloc_topology_t *tp,
 
 static int topo_restrict (hwloc_topology_t topo)
 {
+    // this is not supported on macos, and actually flat fails
+#ifdef __MACH__
+    return 0;
+#else
```
garlick: Could probably drop the `#else` preprocessor conditional leg, making this simpler.

I wonder why this isn't supported and what other fallout this will have :thinking:

src/shell/affinity.c
```diff
@@ -361,7 +370,7 @@ static int task_affinity (flux_plugin_t *p,
     struct shell_affinity *sa = data;
     int i = get_taskid (p);
     if (sa->pertask)
-        hwloc_set_cpubind (sa->topo, sa->pertask[i], 0);
+        wrap_hwloc_set_cpubind (sa->topo, sa->pertask[i], 0);
```
garlick, line 373: Does the shell affinity plugin do anything with this disabled?

src/cmd/flux-terminus.c
```diff
@@ -34,7 +34,7 @@
 #include "src/common/libutil/llog.h"
 
 #include "src/common/libterminus/terminus.h"
-#include "src/common/libterminus/pty.h"
```
garlick: This should probably be broken up into several commits
- rename the source files
- terminal resize fix
- use `openpty()` instead of `posix_openpt()`

On the latter, I'm hoping that's not a bad idea.  Maybe we should just have configure define HAVE_BROKEN_POSIX_OPENPT on darwin and use `openpty()` conditionally?

src/common/libsubprocess/fork.c
```diff
@@ -269,7 +269,10 @@ static int local_exec (flux_subprocess_t *p)
          */
         int status;
         pid_t pid;
-        if ((pid = waitpid (p->pid, &status, 0)) <= 0)
+        int saved_err = errno;
+        while ((pid = waitpid (p->pid, &status, 0)) < 0 && errno == EINTR) {}
+        errno = saved_err;
+        if (pid <= 0)
             return -1;
```
garlick: The errno save/restore seems like a separate change?

Error propagation is a bit confusing here so maybe it's better to make your EINTR change and back away slowly for now.

src/common/libtap/tap.h
```diff
@@ -97,7 +98,9 @@ int tap_test_died (int status);
             tap_test_died(0);                               \
             exit(0);                                        \
         }                                                   \
-        if (waitpid(cpid, NULL, 0) < 0) {                   \
+        int status=0;                                       \
+        while ((status = waitpid (cpid, NULL, 0)) < 0 && errno == EINTR) {} \
+        if (status < 0) {                   \
```
garlick, line 103: initializing status to zero is not needed since it's immediately set.

src/common/libsubprocess/local.c
```diff
@@ -480,7 +480,16 @@ static void sigchld_cb (pid_t pid, int status, void *arg)
...
+    // you ask it to close a file descriptor that is already closed.  It will
+    // take some work and refactoring to fix.  Likely also worth working in
+    // the new addchdir_np command rather than skipping it for getcwd.
+    allow_posix_spawn = false;
+#endif
+    if (allow_posix_spawn
+        && !(p->flags & FLUX_SUBPROCESS_FLAGS_FORK_EXEC)
         && !p->hooks.pre_exec
         && !flux_cmd_getcwd (p->cmd))
         return create_process_spawn (p);
```
garlick: It might be a little cleaner to just add to `flux_local_exec_ex()`
```c
#if __APPLE__
    flags |= FLUX_SUBPROCESS_FLAGS_FORK_EXEC;
#endif
```
with the block comment rather than introduce the runtime conditional.

garlick (commented):
Just a few comments on a quick pass.

src/broker/test/overlay.c
trws: I haven't gotten back to this yet, but you're right.  The original code had the same problem actually, the anonymous namespace is 100% global, running that test twice in parallel on the same box would fail before, and now.  I'll use the create_test_dir or similar, that's a much better idea.

src/common/librlist/rhwloc.c
trws: I can actually explain this one, but it's vaguely lengthy. Short version is they actually provide functions for affinity, I think in some version they may even have worked, but they _very much_ do not work now.  It started when they started making ARM chips with efficiency cores.  My understanding is that hwloc just made any attempt to modify the stuff an error (the macos ones fail silently) so people wouldn't complain that they had no effect.

trws: I should amend that last just a bit. You can still query what exists in the system and things like that, but _binding_ is dead. I think there are still interfaces for limiting the amount of resources available, and preferring performance or affinity cores.

src/common/libsubprocess/fork.c
trws: The errno saving ended up going away, and all the waitpid changes should be in a single commit now as well.

src/common/libsubprocess/local.c
trws: I like that idea, will do.

src/common/librlist/rhwloc.c
trws: So we have it, here's the main interface I was thinking of.  It's more like "this set of tasks/processes is related" than explicit binding, but still perhaps useful.


https://developer.apple.com/library/archive/releasenotes/Performance/RN-AffinityAPI/#//apple_ref/doc/uid/TP40006635-CH1-DontLinkElementID_2

trws:
I think I fixed the last of the test running and check-annotate bugs here.  If the CI passes this time around I'll try to get the last of the style bits fixed later and we may be where I'm hoping to get us for now.

src/shell/affinity.c
trws: At present, I think it does not.

trws:
Disturbingly, the last test result I got had the `test_terminus.t` failure I've only seen locally with my attempts to fix the `pty` problem.  I think the reason it surfaced there is that the `sleep 1000` being spawned was dying due to an IO error or similar because of pty problems, but fundamentally the issue is that there's a race condition in libterminus.  Specifically this is in the `kill_sessions` function, where `terminus_session_kill` is invoked _before_ `flux_pty_add_exit_watcher`. Normally that's fine, because something else is holding on to at least one reference, but if the `sigchild_cb` handler is invoked before the call to `kill_sessions`, then `terminus_session_kill` invokes the final decref and destroys the pty, which is then invalid in the call to `flux_pty_add_exit_watcher`.

I tried to fix this by either reordering these calls, or by doing an extra `incref`/`decref` pair around that part of the code, but any of those fixes lead to hangs elsewhere... Not sure how to chase that one down further without a whole lot of digging, but it's pretty easy to make it happen in the `test_terminus.t` case by having it run something that dies pretty much immediately rather than `sleep 1000`.

src/broker/test/overlay.c
trws: I refactored this to move those functions into libtestutil and use them here and in boot_config.  I was debating registering the directories for cleanup in `atexit` but for now just left it that you have to `rmdir` them manually.

trws:
@garlick, I think I resolved the last of your requests, let me know if I missed something.

src/common/librlist/rhwloc.c
garlick: I think we agreed on `__APPLE__` over `__MACH__`

src/shell/affinity.c
```diff
@@ -32,6 +32,15 @@ struct shell_affinity {
     hwloc_cpuset_t *pertask;
 };
 
+int wrap_hwloc_set_cpubind(hwloc_topology_t topo, hwloc_cpuset_t set, int i)
+{
+    int ret = 0;
+#if !defined (__MACH__)
+    ret = hwloc_set_cpubind (topo, set, i);
+#endif
```
garlick: Same: use `__APPLE__`

src/common/libsubprocess/fork.c
garlick: The errno save/restore has no effect, since `errno = ret` appears a couple of lines later.

src/common/libterminus/client.c
```diff
@@ -271,9 +271,16 @@ static void pty_client_data (struct flux_pty_client *c, flux_future_t *f)
         llog_error (c, "unpack: %s", error.text);
         return;
     }
-    if (write (STDOUT_FILENO, data, len) < 0) {
-        llog_error (c, "write %zu bytes: %s", len, strerror (errno));
-        return;
+    for (int n = 0; n < len; ) {
+        int ret = write(STDOUT_FILENO, (char *)data + n, len - n);
```
garlick: `data` doesn't need to be cast to a `char *` - it already is one.

There should be a space between `write` and `(` per RFC 7.

src/common/libterminus/pty.c
```diff
@@ -710,9 +710,16 @@ static int pty_write (struct flux_pty *pty, const flux_msg_t *msg)
         llog_error (pty, "%s", error.text);
         return -1;
     }
-    if (write (pty->leader, data, len) < 0) {
-        llog_error (pty, "write: %s", strerror (errno));
-        return -1;
+    for (int n = 0; n < len; ) {
+        int ret = write(pty->leader, (char *)data + n, len - n);
```
garlick: same comment about cast and RFC 7

src/cmd/flux-terminus.c
```diff
@@ -372,7 +372,7 @@ static int start_service_daemon (flux_t *orig_h, optparse_t *p)
     }
```
garlick, line 372: Commit message only mentions waitpid + EINTR, but commit also fixes a couple of short writes

src/broker/test/overlay.c
```diff
@@ -441,9 +444,12 @@ void trio (flux_t *h)
      * fails to initialize because its endpoint is already bound.
      */
     errno = 0;
-    ok (overlay_bind (ctx[1]->ov, "ipc://@foo", NULL) < 0
+    char * uri = NULL;
+    asprintf (&uri, "ipc://%s/flux_ipc_foo", ctx[0]->tmpdir);
```
garlick: check for asprintf failure

src/broker/test/overlay.c
```diff
@@ -670,8 +676,13 @@ void wrongness (flux_t *h)
         BAIL_OUT ("overlay_create failed");
 
     errno = 0;
-    ok (overlay_bind (ov, "ipc://@foobar", NULL) < 0 && errno == EINVAL,
+    struct context *ctx = ctx_create (h, 1, 0, "kary:2", NULL);
+    char * uri = NULL;
+    asprintf (&uri, "ipc://%s/flux_ipc_foobar", ctx->tmpdir);
```
garlick: check for asprintf failure

garlick (approved):
Nice!  I haven't tried this on macos yet but the changes look pretty good.  I added some comments but they are no big deal and you can set MWP when you're ready.

src/common/libterminus/client.c
trws: Whoops, copy paste error from another spot where data was a void * before the change, thanks!

src/cmd/flux-terminus.c
trws: Reworded to cover the writes as well, since both were EINTR/EAGAIN problems I considered them versions of the same thing but forgot to write it up that way.

trws:
Ok, I think that covered it. As a sidenote, since almost all the fixups had to touch different commits, I decided to try `git-absorb` to distribute the fixups, and it's _amazing_. Auto-generation of the appropriate fixup commits by identifying the most recent matching chunk, so, so much faster than doing it manually. 🤤

grondo:
>  I decided to try git-absorb to distribute the fixups, and it's amazing. Auto-generation of the appropriate fixup commits by identifying the most recent matching chunk, so, so much faster than doing it manually. 🤤

OMG I need this! Thanks!

garlick:
> I decided to try git-absorb 

I understand sheep's bladders can be used to prevent earthquakes!
(IOW: what new sorcery is this?!?)

trws:
I imagine you already found it, but here's the link for [git-absorb](https://github.com/tummychow/git-absorb). I heard about it a long time ago but for whatever reason never actually tried it until today, wishing I had done it a long time ago.

src/broker/test/overlay.c
```diff
@@ -441,9 +444,13 @@ void trio (flux_t *h)
      * fails to initialize because its endpoint is already bound.
      */
     errno = 0;
-    ok (overlay_bind (ctx[1]->ov, "ipc://@foo", NULL) < 0
+    char * uri = NULL;
+    if (asprintf (&uri, "ipc://%s/flux_ipc_foo", ctx[1]->tmpdir))
```
garlick: Ah this should be a check for < 0.  That's why CI is sad.

trws: Yup, I fixed that part, but there's an issue because the directory wont unlink until all the IPC files do, and I need to rework the lifetimes to fix it.  Taking me a bit to work out the least messy way to do it with how the tests create a variety of these without managing their lifetimes.

trws: I ended up adding a new function to get a thread-safe global test directory string that cleans up with an automatic destructor if it's been used.  Cleaned it up nicely, and seemed like it would be generally useful anyway.

trws: Then ran afoul of PATH_MAX being ill-specified.  I think this finally has it. 🤞

PR #6930, opened 2025-07-24
shell: do not abort early when tasks are oversubscribed to resources

grondo:
Currently `rcalc_distribute()` fails with `ENOSPC` when there are not enough cores given the number of tasks and slot size passed in to the function. However, there are valid cases for this oversubscription, e.g. when options like `--add-brokers` are used when launching a new instance or when a custom taskmap is provided by the user.

This doesn't cause any issues now, but it does affect #6705, because `rcalc_distribute()` is called early with a correct slot size (instead of 1) in this PR. 

This PR modifies `rcalc_distribute()` so it doesn't abort if there are not enough slots available. Instead it repopulates the "alloc info" array and continues, allowing the function to always assign all tasks instead of giving up.

### DIFF
--- a/src/shell/rcalc.c
+++ b/src/shell/rcalc.c
@@ -304,13 +304,16 @@ int rcalc_total_nodes (rcalc_t *r)
     return r->nranks;
 }
 
+static void allocinfo_reset_avail (rcalc_t *r)
+{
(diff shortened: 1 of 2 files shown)


grondo:
Thanks! I've set MWP.

Issue #6589, opened 2025-01-30
system recovery script needed


garlick:
Problem:  Flux fails catastrophically when it cannot write to `/var/lib/flux`.  Generally when this occurs, manual intervention from a flux developer is required.  This is error prone, will not scale, and will not be possible in all environments.

Develop a script similar to an "fsck" that can be run by sys admins after such an event.

Some requirements:
- Always provide a no-intervention route to a working system, even if jobs may be lost
- Make a heroic effort not to lose the list of drained nodes
- Get an emergency dump from the running system, if possible
- Ensure that new job IDs are higher numbered than historical ones to avoid problems with duplicate IDs in logs
- Generate a report of active jobs, if possible
- Create RESTORE symlink so flux can start from the most recent known good dump (not necessarily the emergency dump)
- If there is no known good dump, create one with only the `resource.eventlog` and `checkpoint.job-manager` keys.

chu11:
(side note, thoughts / questions below are me brainstorming #6777 a bit)

you mention possibly using `flux fsck` in rc1.  Do you imagine the process to be something like this in rc1:

```
flux restore dump
flux fsck || echo "content bad .... fix it" && exit 1
module load kvs
```

in this scenario, I imagine the user will be required to run `flux fsck` manually and there's an interactive thing for them

```
> flux fsck /my/flux/statedir
dir missing reference sha1-431473894721432 - recover? (y/n) (user hits yes)
- recovering dir from previous checkpoint 2025-01-05 3:00
```

and we can imagine tons of options, recover from any checkpoint, recover only from a list of checkpoints i say are ok, just list how many thingies are bad, find me the most recent checkpoint that is good, etc. etc.

As I've been thinking about this, I guess I see two things with checkpoint recovery when we have > 1 checkpoint stored

1) investigating which checkpoint to recover from

I think this is where my head is at in regards to #6777

2) doing the actual recovering

This is more with the fsck tool

But as I think about it more ... I'm not sure if situation 1 above matters.  I'm not sure that people care about investigating different checkpoints outside of testing?  They will gladly checkpoint to anything that is 100% good?

chu11:
Wanted to start a design discussion on what we would like `flux-fsck` to look like.  First, summarizing several PRs I submitted.

- #6935

this is a not super important extension, just to test additional checkpoints in the checkpoint history

- #6931
- #6941

[...]

PR #6931, opened 2025-07-24
flux-fsck: Support --checkpoint option

chu11:
Built on top of #6905

Problem: It would be convenient if flux-fsck could checkpoint a rootref if the rootref passes fsck-ing.

Support a new --checkpoint option that will update the KVS checkpoint to the one specified by --rootref.

### DIFF
--- a/doc/man1/common/job-other-batch.rst
+++ b/doc/man1/common/job-other-batch.rst
@@ -1,6 +1,6 @@
 .. option:: --conf=FILE|KEY=VAL|STRING|NAME
 
-   The :option:`--conf`` option allows configuration for a Flux instance
+   The :option:`--conf` option allows configuration for a Flux instance
    started via ``flux-batch(1)`` or ``flux-alloc(1)`` to be iteratively built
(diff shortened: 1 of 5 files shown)


garlick:
I think we're going to want to be very careful about how we implement remediation in `flux-fsck(1)`, as this may be a tool that inexperienced sys admins use in the middle of the night.

We will definitely need to add a check that the KVS is not loaded like we have in `flux-restore(1)` before we roll back checkpoints in the backing store.

Maybe we should hold off on remediation pending some further design discussions and instead make `--checkpoint` an alternate, high level way to specify `--rootref` instead of
- #6935

In keeping with this being a high level tool, we might actually want these options to be hidden, test-only or expert-only options

chu11:
> See latest comment - I think we need to get the design a little more fleshed out before we add remediation.

Ooops, I completely missed your comment (I saw it after I pushed #6941).  Yeah, lets discuss in #6589

chu11:

PR #6933, opened 2025-07-28
doc: update flux-config-queues(5) with working examples

wihobbs:
Problem: the example script included in flux-config-queues(5) will generate a parsing error because the hostlist is not a string. It will additionally error because no resources (cores) are assigned to the created queues. Then, when sched-fluxion-qmanager is reloaded, another error will occur because the "queues" key is unsupported.

Solution: fix the example script in the docs.


wihobbs:
Thanks @garlick! I set MWP.

PR #6934, opened 2025-07-28
broker: refactor broker module loader and add builtin modules

garlick:
Last week, I set out to make some headway on adding an option to load broker modules in a separate process address space, but all I ended up accomplishing was some refactoring to disentangle the existing module code.

Regardless of the approach we eventually take for the original goal, I think this refactoring work is probably worth doing.  It was some really old code.

As a bonus, "builtin modules" are implemented, similar to the shell's, and `connector-local` is now compiled in.


garlick:
Thank you!  Setting MWP.

PR #6935, opened 2025-07-29
flux-fsck: support index inputs to --rootref

chu11:
Built on top of #6931 

Problem: Flux-fsck only allows you to fsck the current checkpoint.  However, there is a potential history of checkpoints stored that users may wish to fsck.

Support the ability for users to specify checkpoint indices to the --rootref option in flux-fsck.  This allows users to fsck other checkpoints stored in the checkpoint history.


chu11:

Issue #6395, opened 2024-10-28
flux-exec: do not operate on failed subprocesses


chu11:
Problem: If a subprocess fails for some reason in `flux-exec`, it still remains in the global `subprocesses` list.  That means that operations on the whole list, such as killing all subprocesses or sending stdin to subprocesses, will continue to operate on those failed subprocesses.

Solution: When a subprocess fails, it should be removed from the global list.

chu11:
nevermind ... we check that the process is still running at all appropriate locations e.g.

```
        while (p) {                                                                                                                         
            if (flux_subprocess_state (p) == FLUX_SUBPROCESS_INIT                                                                           
                || flux_subprocess_state (p) == FLUX_SUBPROCESS_RUNNING) {                                                                  
                if (flux_subprocess_write (p, "stdin", ptr, lenp) < 0)                                                                      
                    log_err_exit ("flux_subprocess_write");                                                                                 
            }                                                                                                                               
            p = zlist_next (subprocesses);                                                                                                  
        }                                     
```

PR #6941, opened 2025-07-31
flux-fsck: support --scan option

chu11:
Built on top of #6395

Problem: Flux-fsck only allows you to fsck one root reference or checkpoint at a time.  It would be useful for fsck to find the most recent checkpoint that is valid.

Support a new --scan option that will iterate through all available checkpoints and find the most recent one that passes.

----

Notes:

- along w/ `--checkpoint` this'll update the checkpoint to the most recent good checkpoint it finds.

- I went back and forth on what to call this.  I considered "rollback", as well, but that implied `--checkpoint` vs specifying it on the command line.  Dunno if folks feel it should go the other way.


chu11:

PR #6942, opened 2025-08-04
broker: let built-in modules be reloaded

garlick:
Problem: a built-in broker module cannot be manually reloaded.

This generalizes the built-in broker module capability at bit so that
1.  A module can be defined as built-in with `autoload=false`, which excludes it from early automatic loading
2. A built-in module can be unloaded and loaded in the usual way with `flux module unload` and `flux module load`

Then `barrier` and `heartbeat` are converted to built-in so that this capability is tested.

We could convert all the modules that are typically loaded on all ranks to built-in and avoid some DSO path search overhead during instance startup.  I'm not sure if that matters too much, but it could be explored.  In any case it seemed like a good idea to allow a built-in module to be reloaded like a DSO one can be.


configure.ac
```diff
@@ -629,7 +629,6 @@ AC_CONFIG_FILES( \
   src/shell/Makefile \
   src/connectors/Makefile \
   src/modules/Makefile \
-  src/modules/connector-local/Makefile \
```
grondo, line 632: commit message typo: convenence

grondo (approved):
This LGTM. Just noticed one typo in a commit message.

configure.ac
garlick: Fixed, thanks!

garlick:
I'll set MWP - thanks!

PR #6945, opened 2025-08-04
ensure flux module trace only shows subscribed events

garlick:
Problem: flux module trace ignores module event subscriptions

Refactor the module code so that tracing occurs after the topic string has been checked against module subscriptions.


grondo (approved):
LGTM!

PR #6946, opened 2025-08-04
ci: add tmate action which can be enabled via debug flag when re-running jobs

grondo:
This PR permanently adds the tmate action to the github workflow. During normal runs, the tmate action is disabled. However, it can easily be enabled by selecting the "Enable debug" flag when re-running jobs.


grondo:
Thanks! Probably needs a manual merge due to modification of the workflow config.

Issue #6936, opened 2025-07-29
For Nvidia Grace, see failing tests for `test_sha1.t`


ocaisa:
```
FAIL: test_sha1.t 1 - FIPS test vector abc
FAIL: test_sha1.t 2 - FIPS test vector abcdbcdecdefdefgefghfghighijhijkijkljklmklmnlmnomnopnopq
FAIL: test_sha1.t 3 - FIPS test vector A million repetitions of 'a'
```
I don't (currently) have direct access so the test machine so I can't give more details. Since this is a compiled test, I suspect that it is somehow system specific and doesn't affect runtime behaviour.

ocaisa:
After disabling that test, we still see some additional failures for Grace:
```
FAIL: t1001-kvs-internals.t 20 - kvs: large put stores raw data into content store
FAIL: t1001-kvs-internals.t 21 - kvs: valref that points to content store data can be read
```

garlick:
@ocaisa - we just tagged flux-core 0.77.0 with a different SHA-1 implementation that is thread safe.   This might be worth retesting, although I don't have any theory for why it would have failed on this platform and no other.

PR #6948, opened 2025-08-04
libutil: update sha1 implementation

garlick:
Problem: the SHA1 implementation is not thread safe and has the same symbol names as the ones from libmd and libssl, which might collide at some point.

Switch to the one from https://github.com/B-Con/crypto-algorithms which is where our sha256 implementation came from.  The license is "public domain".

Adapt its unit test for TAP.
Update blobref.c, the only direct user in flux.

Not sure if the failure in #6936 might be addressed by this?

Background:  I ran into a problem with content related test failures when I compiled some modules including `content`, and `kvs` into the broker as opposed to loading them as independent DSOs.  Both use the blobref.c and hence sha1.c, so that change might have caused them to share static BSS data inside sha1.c that they weren't sharing before.  Thanks to @grondo for reminding me about #6396.


garlick:
Thanks, I'll set MWP and add this to the release notes

PR #6949, opened 2025-08-05
job-list: handle per-resource "nodes" special case

sam-maloney:
Problem: if per-resource type is nodes, but `nnodes` is not explicitly set, then `ntasks` must be finalized when parsing R.

Add an `ntasks_per_node_on_node_count` flag to set `ntasks` in `parse_R()`

For example, the following listing shows only `NTASKS=1` despite `hostname` clearly being run twice, as requested:
```shell
$ flux run --cores=1 --tasks-per-node=2 hostname && flux jobs -a
rocky9-dev
rocky9-dev
 fCkRd258asH alpha    hostname   CD      1      1   0.033s rocky9-dev
```

And after the fix we get the expected `NTASKS=2`:
```shell
$ flux run --cores=1 --tasks-per-node=2 hostname && flux jobs -a
rocky9-dev
rocky9-dev
   f5ZhxTXwD alpha    hostname   CD      2      1   0.074s rocky9-dev
```


sam-maloney:
Alternatively, one might question whether such a combination even makes sense, or if using the `--tasks-per-node` should require `--nodes` also be explicitly set? I'd be happy to try and implement a PR for that if it would be preferred!

grondo:
> Alternatively, one might question whether such a combination even makes sense, or if using the --tasks-per-node should require --nodes also be explicitly set? I'd be happy to try and implement a PR for that if it would be preferred!

Well, there is at least one case where `--tasks-per-node` makes sense without `-N, --nodes`, and that is when a user wants to allocate a set of arbitrary resources, but only run one task per final node in the allocation. For example, this is how you'd run a new Flux instance without explicitly specifying the number of nodes to allocate.

grondo (approved):

PR #6953, opened 2025-08-06
flux-fsck: add --repair option

chu11:
Problem: flux-fsck can identify corrupted KVS entries, but does not presently do anything to help users recover data.

Support a --lost-and-found option.  Corrupted KVS entries will be recovered as best as possible and placed into a "lost+found"
directory for users to consider using.

----

notes:

- I struggled with some documentation / verbiage here.  "recover" doesn't seem to be the right word sometimes, as that suggests full recovery.  Perhaps there's a better word I'm not thinking of.

- there is no handling of corrupted treeobjs, like hypothetically a treeobj that is outright bad (fails `treeobj_validate()`).  We could stick "empty values" in `lost+found` for those?

- I chose the directory "lost+found", which is actually what is used by the `job-manager` for its lost and found.  Should change?  Should `job-manager` path change?

### DIFF
--- a/doc/man1/flux-fsck.rst
+++ b/doc/man1/flux-fsck.rst
@@ -40,6 +40,20 @@ OPTIONS
    checkpoint.  This option directs flux-fsck to start at an arbitrary
    point.  BLOBREF must refer to an RFC 11 tree object of type "dir".
 
+.. option:: -r, --repair
+
+   Remove any dangling references found in KVS metadata. If a KVS
+   value changes as a result of this repair, the key is moved to the
+   lost+found directory.  The original key is unlinked.  If a key
+   points to a KVS directory that is no longer valid, it is
+   unlinked. All damaged keys and their disposition are listed on
+   stderr. This process creates a new root reference for these
+   changes, and commits it as the current KVS checkpoint at the end of
+   the scan. The KVS is required to be unloaded during repair.
+
+   This option should be considered :ref:`EXPERIMENTAL <fsck_experimental>`
+   at this time.
+
 
 EXIT STATUS
 ===========
@@ -63,6 +77,13 @@ FLUX RFC
 :doc:`rfc:spec_10`
 
 :doc:`rfc:spec_11`
+
+
+.. _fsck_experimental:
+EXPERIMENTAL
+============
+
+.. include:: common/experimental.rst
 
 
 SEE ALSO
--- a/src/cmd/builtin/fsck.c
+++ b/src/cmd/builtin/fsck.c
@@ -33,6 +33,7 @@
 
 struct fsck_ctx {
     flux_t *h;
+    json_t *root;
     bool verbose;
     bool quiet;
     int errorcount;
@@ -308,7 +309,6 @@ static void fsck_blobref (struct fsck_ctx *ctx, const char *blobref)
     flux_future_t *f;
     const void *buf;
     size_t buflen;
-    json_t *treeobj;
     json_t *dict;
     const char *key;
     json_t *entry;
@@ -324,17 +324,16 @@ static void fsck_blobref (struct fsck_ctx *ctx, const char *blobref)
         flux_future_destroy (f);
         return;
     }
-    if (!(treeobj = treeobj_decodeb (buf, buflen))
-        || treeobj_validate (treeobj) < 0)
+    if (!(ctx->root = treeobj_decodeb (buf, buflen))
+        || treeobj_validate (ctx->root) < 0)
         log_msg_exit ("blobref does not refer to a valid RFC 11 tree object");
-    if (!treeobj_is_dir (treeobj))
+    if (!treeobj_is_dir (ctx->root))
         log_msg_exit ("root tree object is not a directory");
 
-    dict = treeobj_get_data (treeobj);
+    dict = treeobj_get_data (ctx->root);
     json_object_foreach (dict, key, entry) {
         fsck_treeobj (ctx, key, entry);
     }
-    json_decref (treeobj);
     flux_future_destroy (f);
 }
 
@@ -411,6 +410,7 @@ static int cmd_fsck (optparse_t *p, int ac, char *av[])
     if (!ctx.quiet)
         fprintf (stderr, "Total errors: %d\n", ctx.errorcount);
 
+    json_decref (ctx.root);
     flux_close (ctx.h);
 
     return (ctx.errorcount ? -1 : 0);
--- a/src/cmd/builtin/fsck.c
+++ b/src/cmd/builtin/fsck.c
@@ -34,8 +34,13 @@
 struct fsck_ctx {
     flux_t *h;
     json_t *root;
+    int sequence;
+    int repair_count;
+    zlist_t *repair_treeobjs;
+    char *hash_name;
     bool verbose;
     bool quiet;
+    bool repair;
     int errorcount;
 };
 
@@ -117,7 +122,7 @@ static void valref_validate_continuation (flux_future_t *f, void *arg)
         }
         fvd->errorcount++;
         fvd->errnum = errno;     /* we'll report the last errno */
-        if (errno == ENOENT)
+        if (fvd->ctx->repair && errno == ENOENT)
             save_missing_ref_index (fvd, *index);
     }
     fvd->in_flight--;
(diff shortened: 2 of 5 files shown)


garlick:
I forgot that the job manager already was using `lost+found`.   Maybe for now we could just rename that directory to `job-lost+found`, with the eventual plan of auditing those error paths and seeing if we should handle them differently.

Seems like repair should be the default, but maybe at the end before checkpointing we should get confirmation, either a Y from the user if stdin is a tty, or require `-y` on the command line?  Until we do that it's a no-op.

Perhaps "repair" instead of "recovery".

garlick:
Reminder: we should require that the KVS is not running like we do in `flux-restore(1)`.

chu11:
> Seems like repair should be the default

I thought about that, but made it an option b/c normally an `fsck` doesn't require the KVS to be loaded.  With `lost+found` it does require it to be loaded (and the content module too I think?).  So making it the default would requiere the KVS to be loaded before doing an fsck?

Edit: I suddenly have a recollection we had a offline discussion about possibly trying to work on the KVS **without** having the KVS loaded.  Very doable, but a little tricky.  Is this what you were thinking for this lost+found?

Edit2: Yeah, the offline KVS "fixes" is what we should be doing.

> but maybe at the end before checkpointing we should get confirmation, either a Y from the user if stdin is a tty, or require -y on the command line? Until we do that it's a no-op.

Originally my thought was that since we weren't changing anything amongst "real KVS data", figured checkpoint confirmation wasn't necessary.  i.e. the only change is lost+found.  Is your thinking that the Y/n confirmation is just to make sure the user is aware corruption was found and repaired?

Edit: If we go with the offline KVS update, then the Y/n is more obvious, as we're circumventing normal KVS stuffs.

chu11:
just pushed a WIP update

- the lost+found is now done via work directly on the content store, not through the KVS
  - i should split this out into a mini-library and test it, gonna work on that in my next round
  - Edit: ya know, I think the code addition is under the "convenience library" line ... i'm thinking i may not do that now.

- kept lost+found as an option, it's sorta nice when testing fscks over and over and w/ --rootref.  We can debate this

- checkpoint is still "automatic" for now, can work on that later for user to say `Y/n`.  Maybe a follow on PR to debate exactly how to do it? and leave this PR specifically for the lost+found work.

chu11:
I decided that splitting out some code to a convenience library wasn't worth it at this point.  In time with some of our other ideas from #6589, perhaps it'll be more worthwhile.  So I removed WIP for now and we can debate some of the points I list above.

garlick:
For starters we should probably rename `--lost+found` to something like `--repair` since without it, the tool just scans (correct?)

chu11:
So I was working on adding the `Y/n` question to the user and adding a `--checkpoint` option to always checkpoint (and not ask the user to confirm checkpointing).

But as I was adding tests, I realized ... if we keep `--lost-and-found` (or `--repair` as suggested above) an option, is the `Y/n` question is needed?  By specifying `--repair`, the user is likely specifying "i want this to be done".  In other words, `Y/n` is probably only needed if we were to do `lost+found` by default.

doc/man1/flux-fsck.rst
```diff
@@ -40,6 +40,11 @@ OPTIONS
    checkpoint.  This option directs flux-fsck to start at an arbitrary
    point.  BLOBREF must refer to an RFC 11 tree object of type "dir".
 
+.. option:: -l, --lost-and-found
+
+   Repair corrupted KVS entries and copy them into the `lost+found` directory.
+   Will update the primary KVS checkpoint after `fsck` is complete.
+
```
garlick: This should probably be expanded a bit e.g.

> Remove any dangling references found in KVS metadata.  If a KVS value changes as a result of this repair, the key is moved to the `lost+found` directory.  If a key points to a KVS directory that is no longer valid, it is unlinked.  All damaged keys and their disposition are listed on stderr.  This process creates a new root reference for these changes, and commits it as the current KVS checkpoint at the end of the scan.  The KVS is required to be unloaded during repair.

garlick:
Yeah maybe not needed for now?  I guess since we keep some recent checkpoints, rolling back is a manual option if the check removes like everything or something?

chu11:
> Yeah maybe not needed for now? I guess since we keep some recent checkpoints, rolling back is a manual option if the check removes like everything or something?

Yeah, sounds like a plan.  Although the only way to do this right now is the KVS `initial-rootref` mechanism.  Shall I consider resurrecting #6931?  Or perhaps we need a "update checkpoint" command (or option to another tool).

garlick:
>  Or perhaps we need a "update checkpoint" command (or option to another tool).

I'd vote for this.    We have `flux content checkpoints` for listing checkpoints, so probalby it belongs there somehow?

chu11:
> I'd vote for this. We have flux content checkpoints for listing checkpoints, so probalby it belongs there somehow?

Yeah, that's what I was thinking.  perhaps should refactor so we can `flux content checkpoint list`, `flux content checkpoint update`, etc.

doc/man1/flux-fsck.rst
chu11: > If a key points to a KVS directory that is no longer valid, it is unlinked.

I actually didn't do this.  I felt this would be updating the "real" KVS tree (vs other repairs are just copied into the lost and found).  Do you think we should be modifying the "real" tree?  This should maybe be for a follow up PR, gotta think about how to weave that in.

Alternately, I suppose I could just make a copy of the dir in `lost+found` ... I didn't consider that.

chu11: i put some thoughts about dirref repair into #6589.  Minimally, lets punt on it to a follow on PR.

chu11: I just noticed a subtlety in your description.  You say "The key is **moved**" , do you think we should unlink the original?  i.e. if we repair `dir.a.b`, should we remove `dir.a.b` and add `lost+found.dir.a.b`?

chu11:
re-pushed, `--lost-and-found` is now `--repair`, tweaked some documentation per comment above

there are some open questions

- `dirref` not dealt with here, see comments in #6589, should probably be for follow up PR (if we do anything)
- a corrupted key is left in place and a repaired copy is in `lost+found`, we could
  - remove the corrupted original version
  - have an option to put the repaired copy in place or symlink it?? (could be for follow up feature)

garlick:
> a corrupted key is left in place and a repaired copy is in lost+found, we could

Oh that's bad.  The key needs to be unlinked and not replaced with a symlink.
We need all bad metadata to be eradicated, and we shouldn't leave corrupted data in place for things to trip over.  Better to have one "key missing" error than deal with myriad ways of corrupting data, some of which may be undetectable.

chu11:
re-pushed, built on top of #6999.

This does everything discussed above

- valrefs that are repaired are moved to `lost+found`, originals are unlinked
- dirrefs that are bad get removed from directories
- we checkpoint at the end w/o asking user

side note: #6981 and #6982 to deal w/ the possible need for user to revert checkpoints to something older

garlick:
This needs to be rebased now that #6999 is merged.

I'll try to get through a review on this quickly so we can get it in the upcoming release (today?)

doc/man1/flux-fsck.rst
```diff
@@ -40,6 +40,17 @@ OPTIONS
    checkpoint.  This option directs flux-fsck to start at an arbitrary
    point.  BLOBREF must refer to an RFC 11 tree object of type "dir".
 
+.. option:: -r, --repair
+
+   Remove any dangling references found in KVS metadata. If a KVS
+   value changes as a result of this repair, the key is moved to the
+   lost+found directory. If a key points to a KVS directory that is no
+   longer valid, it is unlinked. All damaged keys and their
```
garlick: > If a key points to a KVS directory that is no longer valid, it is unlinked.

What happens if a key points to a value (valref) that is no longer valid?

garlick: At this stage we might want to tag `--repair` as experimental by including the experimental section at the bottom:
```rst
.. _fsck_experimental:
EXPERIMENTAL
============

.. include:: common/experimental.rst
```
and referencing it from the option description:
```rst
This option should be considered :ref:`EXPERIMENTAL <fsck_experimental>` at this time.
```

garlick (approved):
I'm not spotting anything wrong here. Although it is a bit hard for me to follow, it seems to have good tests so given that it's a new experimental thing, maybe OK to go in?

doc/man1/flux-fsck.rst
chu11: > What happens if a key points to a value (valref) that is no longer valid?

It is also unlinked.  Lemme update documentation to make that clearer (I said the key "moved" to the lost+found, but saying it is unlinked would be clearer)

garlick: but if a valref points nowhere so there is no content for the value, it can't be placed in the lost+found right?

chu11: > but if a valref points nowhere so there is no content for the value, it can't be placed in the lost+found right?

Oh, are you referring to the case where all references in the valref are invalid, thus there is no more data?  It is still placed in the lost+found but it is now an empty "val" treeobj.

chu11:
re-pushed with doc fixes per discussion above

garlick:
Good to go as far as I am concerned.

chu11:
cool, setting MWP

grondo:
Added this one to the NEWS.md PR

PR #6954, opened 2025-08-06
ci: fix docker manifest generation

grondo:
Problem: The set of images for which docker manifests are generated is hand edited in the workflows file and so can easily become outdated.

Add a script to run the docker manifest commands based on current output from `src/test/generate-matrix.py`, which contains the exact set of docker tags for each image which then must be combined into a manifest.


grondo:
Thanks! This will likely need to be manually merged since it touched `.github/*`

trws:
Yup, sounds right.  Feel free to hit it, or I can if you prefer.

grondo:
Thanks! I went ahead and pressed the button.

grondo:
Well bummer. That failed 😔 I'll take a look tomorrow

PR #6955, opened 2025-08-07
ci: add missing checkout step to generate-manifest

grondo:
I forgot that the generate-manifest job in ci will need to checkout a copy of the repo now that it utilizes a couple scripts. :facepalm: This _should_ fix the current problem, and hopefully no other issues pop up after this is merged. :disappointed:


grondo:
Thanks, I'll press the merge button. :crossed_fingers:

PR #6956, opened 2025-08-07
flux-fsck: check that KVS is not loaded

chu11:
Problem: flux-fsck checks the integrity of the KVS backing store.
Therefore, the KVS should not be running and potentially modifying
the KVS whiel a fsck is going on.

Check if the KVS is loaded and return an error to the user if it is.

---

Splitting this one off of #6953


chu11:
@garlick thanks, will set MWP

PR #6957, opened 2025-08-08
modprobe: fix handling of `modprobe.Context.bash()` errors and expand testing

grondo:
This PR fixes the `modprobe.Context.bash()` helper, which currently is ignoring non-zero exit status from the invoked command(s).

Additionally, a few more tests are added to the modprobe testsuite to check that the `bash()` helper is working properly, and that module load failures (and other `@task` failures) cause `flux modprobe` to fail.

### DIFF
--- a/doc/man1/flux-modprobe.rst
+++ b/doc/man1/flux-modprobe.rst
@@ -58,6 +58,11 @@ service name instead of module name e.g.::
 
 loads the currently configured scheduler module.
 
+This command will only load modules that are not already loaded, so it
+may be used to ensure a given set of modules are loaded. If all target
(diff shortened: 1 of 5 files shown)


garlick (approved):
LGTM!

grondo:
Thanks. I have a couple more improvements incoming here...

grondo:
Ok, sorry about that @garlick. Changes since your last review:

 - Changed `flux modprobe load` so it ignores already loaded modules. Unlike `flux module load`, It is not an error if a module and all its dependencies are already loaded. Document this.
 - Add a test for the above
 - Simplify use of `flux modprobe load` in one test in the testsuite.

Issue #5488, opened 2023-10-05
basic job resource usage accounting


grondo:
A user was asking for the equivalent of 
```
sacct --format="MaxRSS" -j <job_number>
```
in Flux.

AFAIK, we don't have an issue open on basic resource usage accounting for Flux, though possibly #2999 is related.

There's quite a list of things that `sacct` can report on, so as part of resolving this issue we would need a facility to gather generic resource and other stats and have them available after a job is complete (and during would be cool too for display in `flux jobs` and `flux top` as @garlick notes in #2999)

PR #6958, opened 2025-08-08
WIP: add sysmon shell plugin

garlick:
There was a flux-discuss email question recently asking about a way to monitor resource utilization.  I had some cpu and memory utilization code laying around and made a quick and dirty broker module that polled `/proc` on the heartbeat and calculated cpu usage over the heartbeat interval.  My thought was that someone could do better making a bulk RPC to each rank to gather this info than trying to spawn processes.

Then I made a demo shell plugin that just monitors the local cpu and memory usage and looks for big changes to log, wtih the thought that this could be the basis for a debugging tool that looks for deadlocked or leaky ranks in a big job.

Meh.  This may be a solution in search of a problem.  I thought I'd post it anyway since the sunk cost is nonzero, it and see if anybody had any thoughts on how to redirect it towards something more useful?  If not, I have no problem euthanizing it.


grondo:
I haven't looked at this fully yet, but this seems like a pretty useful start to me!

If the shell sysmon plugin could also monitor some process statistics, this could also potentially solve #5488.

We could add a special handler to the shell rank 0 plugin that pulls data from all the other shells and sends the result back to the caller (as is done with the mpir plugin for the procdesc data)  Perhaps with some simple ways to aggregate the bulk data. This has proven to be pretty scalable.

garlick:
I reworked this to just use the shell's cgroup memory and cpu controller directly. If enabled with `-o sysmon`, each shell will log memory and cpu info at the end of the job:
```
    5.065s: flux-shell[1]: sysmon: memory.peak=3.1179504G
    5.066s: flux-shell[1]: sysmon: loadavg-overall=2.00
    5.068s: flux-shell[0]: sysmon: memory.peak=3.1179504G
    5.068s: flux-shell[0]: sysmon: loadavg-overall=2.00
```
If tracing is enabled with `-o verbose=2`, a little snapshot is logged on each heartbeat:
```
    2.497s: flux-shell[0]: TRACE: sysmon: memory.current=1.3986053G
    2.497s: flux-shell[0]: TRACE: sysmon: loadavg=2.00
    2.497s: flux-shell[1]: TRACE: sysmon: memory.current=1.3990936G
    2.497s: flux-shell[1]: TRACE: sysmon: loadavg=2.00
 ```
The polling period for the tracing can be explicitly set with `-o sysmon.period=FSD`.

Aggregation of this data would make it a lot more useful, but I'm wondering if this baby step is enough better than nothing that we should consider it as is, as a stepping stone?

No tests yet - I wanted to see what others thought.

**Comment by garlick** [comment]
I added testing and removed the WIP.  I am not hard over on merging this as is, but wanted the option to be available.

File: src/shell/sysmon.c [hunk 1]
```diff
@@ -0,0 +1,404 @@
...
+                            "period", &period) < 0) {
+            shell_log_error ("option parse error: %s", jerror.text);
+            return -1;
+        }
+        if (period) {
+            if (json_is_integer (period)) {
+                int i = json_is_integer (period);
+                if (i < 0)
+                    goto error_period;
+                ctx->period = json_integer_value (period);
```
grondo: I think `json_is_integer()` only returns `0` or `1` so the test could never be true? Did you mean to compare `json_integer_value() < 0`? (I'm not sure, though, since `i` wasn't reused to assign to `ctx->period`.

src/shell/sysmon.c
```diff
@@ -0,0 +1,404 @@
...
+    if (sysmon_parse_args (ctx, config) < 0)
+        goto error;
+    ctx->path_memory_current = get_cgroup_path (mypid, "memory.current", R_OK);
+    ctx->path_memory_peak = get_cgroup_path (mypid, "memory.peak", R_OK);
+    ctx->path_cpu_stat = get_cgroup_path (mypid, "cpu.stat", R_OK);
+    if (!ctx->path_memory_current
+        || !ctx->path_memory_peak) {
+        shell_log_error ("error caching cgroup paths");
+        goto error;
+    }
```
grondo: Probably should be obvious, but why isn't `ctx->path_cpu_stat` checked along with the memory paths?

src/shell/sysmon.c
garlick: Good catch!  Yeah I'm sure I meant to check the value.

src/shell/sysmon.c
garlick: An oversight!  Fixing.

garlick:
I fixed those things and rebased on current master.

garlick:
I'll set MWP, thanks!

PR #6959, opened 2025-08-08
broker: convert groups to a module

garlick:
Problem: the broker uses the groups service for quorum management, but groups is a standalone service that doesn't need to be as tightly coupled to the broker as it is.
    
Convert groups to a built-in broker module with `autoload=true`.   The reduced coupling between broker and groups should make both easier to maintain.


grondo:
Only somewhat related to this PR, but should builtin modules with `autoload=true` and that cause extreme failure if unloaded be listed in `flux module list` output by default? Seems like maybe `connector-local` and perhaps this new groups module should be excluded from this output since they are not really meant to be reloaded.

garlick:
Good question.   Maybe we could add an unloadable flag?  It is mainly `connector-local` that should never be unloaded since there is no way to load it back again.  `groups` actually can be reloaded.  I don't think we should make them invisible because it might be useful to have their Idle, Sendq, Recvq fields in `flux module list`, remember that we can `flux module stats` them, etc..

By the way, when I reloaded `groups` I noticed that `resource` didn't seem to notice.  I better fix that.

grondo:
Ok, that makes sense. It only came up because modprobe wants a list of all modules, but has to treat `connector-local` differently since it is auto-loaded and should not be unloaded, etc. Easy enough to work around though.

grondo:
This one seems reasonable, but it seems to have somehow introduced a race condition into startup where sometimes the local connector is not yet available when rc1 starts, e.g.

```
  Aug 08 23:49:44.456928 UTC 2025 broker.err[0]: rc1.0: flux-modprobe: ERROR: Unable to connect to Flux: broker socket /tmp/flux-system-BZ8ht4/local-0 was not found
  Aug 08 23:49:44.457125 UTC 2025 broker.err[0]: rc1.0: /usr/src/etc/rc1 Exited (rc=1) 0.8s
  Aug 08 23:49:44.587392 UTC 2025 broker.err[0]: rc3.0: flux: Error committing finish event: No service matching kvs.commit is registered
  flux-start: 0 (pid 245723) exited with rc=1
  ```

**Comment by garlick** [comment]
I've pushed some more commits to address test issues and issues that came up when I briefly played with treating `groups` like a regular module that is loaded through modprobe.  (That works but it is required for instances of size > 1 even when nothing else is loaded so it breaks a lot of tests and  I backed away from it).

I guess the separation from the broker code is probably a good thing, although I wish I could think of a simple way to make the module optional so it wouldn't need to be auto loaded.

> modprobe wants a list of all modules, but has to treat connector-local differently since it is auto-loaded and should not be unloaded, etc. Easy enough to work around though.

I'm missing where this needs to be done (other than in one sharness test that I already updated)?

**Comment by grondo** [comment]
> I'm missing where this needs to be done (other than in one sharness test that I already updated)?

I think I worked around this in the final modprobe implementation by not trying to remove modules that aren't configured in `modprobe.toml` or `modprobe.d/*.toml`. So `connector-local` is now a module that is effectively ignored.

I guess this will be the workaround for all autoloaded modules (meaning also they won't be removable by the modprobe implementation)

File: src/modules/groups/groups.c [hunk 1]
```diff
@@ -849,29 +907,32 @@ static const struct flux_msg_handler_spec htab[] = {
...
+        flux_future_destroy (g->f_ovmon);
         flux_msg_handler_delvec (g->handlers);
         flux_watcher_destroy (g->batch_timer);
         free (g);
         errno = saved_errno;
     }
 }
 
-struct groups *groups_create (struct broker *ctx)
+struct groups *groups_create (flux_t *h)
```
grondo: Should `groups_create` be declared `static`?

t/t0100-modprobe.t
```diff
@@ -859,7 +859,7 @@ test_expect_success 'modprobe: fails if module load fails' '
 	flux python <<-EOF >all_modules &&
 	import flux
 	from flux.modprobe import ModuleList
-	modules = [x for x in ModuleList(flux.Flux()) if x != "connector-local"]
+	modules = [x for x in ModuleList(flux.Flux()) if x != "connector-local" and x != "groups"]
```
grondo: could also do
```python
if x not in ("connector-local", "groups")
```

If there is some way to determine that a module auto-loaded, we could extend modprobe to handle these things better.

garlick: A thought is to 1) take your advice a bit late and eliminate non-auto-loaded builtin modules, and 2) use the module path of `"builtin"` as a key to omit those modules from `ModuleList` output.  The same could be done by using a special path for the auto-loaded built-ins but sheesh, more complexity for a feature that isn't buying us much.

grondo: > 1) take your advice a bit late and eliminate non-auto-loaded builtin modules,

Hm, I didn't mean to suggest to eliminate the other builtin modules.

I think an `autoload` boolean flag in the `module.list` RPC might be sufficient, then as you suggest above `ModuleList` could just ignore those.

Another approach could be just to have the modprobe `ModuleList` ignore modules that aren't within its configuration. In retrospect this might make the most sense since `ModuleList` is a modprobe helper...

garlick: Well, I kinda think it's a good idea to revert that.  I guess I was hoping there would be a performance benefit but it seems any savings from converting a module from DSO to built-in is very tiny or hidden by the parallelism benefit of `flux-modprobe(1)`.  It feels extra at this point.  I'll submit another PR for your review that does this.

garlick:
This is rebased on master.  OK to set MWP?

garlick:
I think we've covered this - I'll go ahead and set  MWP.

PR #6961, opened 2025-08-11
etc: remove shopt usage from rc1.old

### LINKED ISSUE #6960: rc1.0: /etc/flux/rc1.old: 91: shopt: not found
Problem: after updating to 0.77, a flux system instance on debian-12 won't start due to the `shopt` command in rc1.old.

```
Aug 09 14:14:03 picl0 flux[69905]: broker.err[0]: rc1.0: /etc/flux/rc1.old: 91: shopt: not found
Aug 09 14:14:03 picl0 flux[69905]: broker.err[0]: rc1.0: /etc/flux/rc1 Exited (rc=127) 0.4s
Aug 09 14:14:03 picl0 flux[69905]: broker.info[0]: rc1-fail: init->shutdown 0.413172s
```

Strangely, this was OK when it was in `02-cron` but isn't now in `rc1.old`.

grondo:
Problem: Commit ee047214df897daadd228e09797c2edfc9407601 copied the contents of the old `etc/rc1.d/02-cron` directly into rc1.old, but that file was designed to be run under bash and contains `shopt` commands which only work under bash. Since `rc1` and `rc1.old` use `/bin/sh`, this shell builtin may not be available.

The for loop in `02-cron` already checked for existence of the file to load, so the `shopt` commands are likely unnecessary. Remove them so that rc1.old becomes POSIX sh compatible.

Fixes #6960

### DIFF
--- a/etc/rc1.old
+++ b/etc/rc1.old
@@ -88,7 +88,6 @@ modload all heartbeat
 if test $RANK -eq 0 \
 	&& cron_dir=$(flux getattr cron.directory 2>/dev/null) \
 	&& test -d "$cron_dir"; then
-	shopt -s nullglob
 	for file in $cron_dir/*; do
 		if test -f $file; then
 			if ! flux cron tab <$file; then
@@ -97,7 +96,6 @@ if test $RANK -eq 0 \
 			fi
 		fi
 	done
-	shopt -u nullglob
 fi
 
 


garlick:
Lgtm!

grondo:
I did test this on my system, but perhaps you should double check @garlick

garlick (approved):
As discussed offline I did test this and WG (works good).

grondo:
Thanks! I've set MWP.

PR #6968, opened 2025-08-15
doc: add note about `--signal=SIG@TIME` behavior in batch jobs

### LINKED ISSUE #6964: --signal=USR2@TIME kills flux batch
A user reported that `--signal=USR2@TIME ...` wasn't functioning properly on coral2 systems. I found that `flux run --signal=USR2@TIME ...` works as expected (my process catches a signal, then continues), but `flux batch --signal=USR2@TIME ...` was not. In the `flux batch` case, the signal gets sent and caught as expected, but the job immediately exits. My reproducer involves a python script that catches SIGUSR2 and prints a message, sleeps for a minute, then prints another message and exits:

```
[day36@helix14:signal_test]$ cat sig_catcher.py 
#!/bin/python3
import signal
import os
import time

def receive_signal(signum, stack):
    print(f"Received: {signum}", flush=True)
    time.sleep(60)
    print("slept 1m, now I'll stop", flush=True)
    exit(signal.SIGHUP)

signal.signal(signal.SIGUSR2, receive_signal)

print(f"My PID is: {os.getpid()}", flush=True)

for _ in range(200):
    print("Waiting...", flush=True)
    time.sleep(10)
[day36@helix14:signal_test]$
```

If I submit this in a batch script with the `--signal=...` flag on a `flux run`, I get the expected output. If I instead include the `--signal=...` as a `#flux: ` line in the batch script, the job dies immediately after catching the signal. I see the 'Received:' output, but not the 'slept 1m' output and the reported duration of the job is 1 minute shorter. My batch scripts are:

```
[day36@helix14:signal_test]$ cat flux_catcher.batch 
#!/bin/bash
#flux: -N1
#flux: -t 6m

flux run -n1 -t 5m --signal=USR2@2m ./sig_catcher.py
[day36@helix14:signal_test]$ cat flux_catcher2.batch 
#!/bin/bash
#flux: -N1
#flux: -t 5m
#flux: -q pdebug
#flux: --signal=USR2@2m

flux run -n1 ./sig_catcher.py
[day36@helix14:signal_test]$
```

grondo:
Problem: A signal sent to a batch job is delivered both to all the running jobs and the batch script. The latter may be surprising or catch users off guard if they expect only the jobs to be signaled.

Add a note to the submission commands documentation for `--signal` describing signal behavior for batch jobs, and suggesting batch scripts use a `trap` to capture `SIG` to avoid early exit.

Fixes #6964

@ryanday36, does this seem sufficient to close #6964?

### DIFF
--- a/doc/man1/common/job-other-options.rst
+++ b/doc/man1/common/job-other-options.rst
@@ -105,6 +105,13 @@
    greater than the remaining time of a job as it starts, the job will
    be signaled immediately.
 
+   .. note::
+      A signal sent to a batch job is delivered to both the batch script and
+      any running jobs. If the script does not handle the signal it may exit
+      first, terminating the batch job early. It is recommended to use a
+      ``trap`` in the script to allow running jobs to handle the signal and
+      exit cleanly.
+
    The default behavior is to not send any warning signal to jobs.
 
 .. option:: --dry-run


grondo:
Ah, thanks for catching that!

PR #6969, opened 2025-08-15
broker: panic on spurious module exit with error

garlick:
This adds a mechanism to stop the broker (with error) if a module spuriously exits with an error code - that is, if it exits `mod_main()` with a nonzero return code AFTER it has started the reactor or called `flux_module_set_running()`.

Since on rank 0 this would cause the entire instance to shut down, it is disabled on the command line by the systemd unit file, and the situation is logged at LOG_CRIT instead.   The main point was to catch problems in test, e.g. with fluxion's qmanager which, by necessity, does some of its initialization after `flux module load` has returned successfully.  Catching this type of error in a batch job probably makes sense too.  But the system instance has a wider impact so I thought it best to retain the status quo there.

### DIFF
--- a/doc/man7/flux-broker-attributes.rst
+++ b/doc/man7/flux-broker-attributes.rst
@@ -144,6 +144,11 @@ broker.exit-restart [Updates: C, R]
    A numeric exit code that the broker uses to indicate that it should not be
    restarted.  This is set by the systemd unit file.  Default: unset.
 
+broker.module-nopanic [Updates: C, R]
+   By default, when a broker module spuriously exits with error, the broker
+   shuts down its subtree and fails.  If this attribute is set, this event
+   is merely logged.
+
 broker.starttime
    Timestamp of broker startup from :man3:`flux_reactor_now`.
 
--- a/doc/test/spell.en.pws
+++ b/doc/test/spell.en.pws
@@ -987,3 +987,4 @@ getopts
 kwargs
 subdirectory
 py
+nopanic


garlick:
I suppose somewhat predictably, some fluxion tests are failing in the flux-sched check.   We may need to set  `broker.module-nopanic=1` over there for now.

garlick:
I updated the PR title and the PR to sched has been merged, so tests are passing here now.  OK to set MWP?

garlick:
@mergifyio requeue

trws:
One technical question here, does this mean that a flux module reload of fluxion in a batch job that fails after load will kill the instance by default? That may be the right behavior, but it makes me a bit nervous about it.

I'm catching up here, but "spurious" is not the right term to use for this failure mode.  Spurious failure is when you get `EAGAIN` from `write`, it's not actually an error and the C library is lying to you, or perhaps a module gets killed with a signal or pthread_kill or something similarly unexpected and out of its control. This is "we hit an error and need to stop, here is your error." I see above where @grondo asserts that "spurious exit" is well understood, but have never heard that term, and a web search for "spurious exit" finds 26 results, all of which are spurious (not actually related to this usage).   

Perhaps "on module exit with error" since it happens in all cases where a module exits with an error after load and not on any specific condition that requires a modifier?

grondo:
>  I see above where @grondo asserts that "spurious exit" is well understood, but have never heard that term, and a web search for "spurious exit" finds 26 results,

I had always taken it to mean an early or unintended exit, in this case a module "exiting" before it is unloaded.
However, it is entirely believable that I'm confusing that with another term. :shrug:

PR #6971, opened 2025-08-18
broker: wait for built-in plugins to be RUNNING before loading other modules

garlick:
Problem: since the early days, there has a been a race between loading the `connector-local` module and loading other modules, which requires that the local connector be usable.  We have addressed this in CI by increasing connect retries in rc1, but that still can fail on occasion.

This integrates builtin module loading into the broker state machine such that the state does not advance to from NONE to JOIN until after the built-in (`autoload=true`) modules are in RUNNING state.  Since `connector-local` does not enter RUNNING state until it is listening on the socket, this effectively closes the race.

This appears to address the failures that are occurring in CI in #6959, where the broker groups subsystem is converted to an `autoload=true` module.

### DIFF
--- a/etc/rc1.old
+++ b/etc/rc1.old
@@ -1,7 +1,6 @@
 #!/bin/sh -e
 
-# Allow connector-local more time to start listening on socket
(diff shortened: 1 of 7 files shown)


garlick:
Fixed a commit message typo and improved one error message.

PR #6973, opened 2025-08-19
sched-simple: allow non-integer counts

sam-maloney:
Problem: `sched-simple` cannot handle non-integer counts

Add a new `libjjc` (which is a modification of `libjj` to use new count infrastructure in `libjob`) and use it in `sched-simple` to allow non-integer counts.

As a first step, it simply uses the minimum value in the allocation request, but at least `sched-simple` no longer chokes on counts using ranges or idsets!


grondo:
I've set MWP as well.

PR #6974, opened 2025-08-20
libflux: drop msglist polling methods

garlick:
Problem: `flux_msglist_pollfd(`) and `flux_msglist_pollevents()` are no longer used.

Get rid of them.

TL;DR This  mechanism moved to the internal-only `msg_deque` class.  flux-pmix did use these interfaces to implement its reactive interthread message channel, but that has been replaced by back-to-back `flux_t` handles using the `interthread://` connector.  I prepared this commit last year but I guess was probably waiting for there to be a flux-pmix tag and forgot about it.

### DIFF
--- a/src/common/libflux/msglist.c
+++ b/src/common/libflux/msglist.c
@@ -11,10 +11,6 @@
 #if HAVE_CONFIG_H
 #include "config.h"
 #endif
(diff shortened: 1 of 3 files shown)


grondo (approved):
LGTM!

PR #6975, opened 2025-08-20
job-manager: use flux_msglist where helpful

garlick:
Problem: the job manager uses `zlist_t` and manual reference counting to manage lists of messages in a couple places, but we have `struct flux_msglist` for that.

Convert, simplifying the code a bit.

Just cleanup, no functional change.

### DIFF
--- a/src/modules/job-manager/event.c
+++ b/src/modules/job-manager/event.c
@@ -75,7 +75,7 @@ struct event_batch {
     flux_kvs_txn_t *txn;
     flux_future_t *f;
     json_t *state_trans;
-    zlist_t *responses; // responses deferred until batch complete
+    struct flux_msglist *responses; // responses deferred until batch complete
     zlist_t *jobs;      // jobs held until batch complete
 };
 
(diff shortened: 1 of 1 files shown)


src/modules/job-manager/drain.c
```diff
@@ -28,8 +28,8 @@
 struct drain {
     struct job_manager *ctx;
     flux_msg_handler_t **handlers;
-    zlist_t *drain_requests;
-    zlist_t *idle_requests;
+    struct flux_msglist *drain_requests;
+    struct flux_msglist *idle_requests;
```
grondo, line 32: Commit message typo: unbalanced quotes around 'struct flux_msglist".

src/modules/job-manager/event.c
```diff
@@ -75,7 +75,7 @@ struct event_batch {
     flux_kvs_txn_t *txn;
     flux_future_t *f;
     json_t *state_trans;
-    zlist_t *responses; // responses deferred until batch complete
+    struct flux_msglist *responses; // responses deferred until batch complete
```
grondo, line 78: Same commit message typo as previous commit.

src/modules/job-manager/drain.c
garlick: DOH!

garlick:
Thanks! Fixed that and setting MWP.

PR #6976, opened 2025-08-20
drop the barrier module

garlick:
Problem: the barrier broker module dates back to olden times, has severe limitations, no documentation, and no users (other than tests).

This gets rid of the module, the `flux_barrier()` function in the public API, and its python binding.

Unfortunately, some sharness test "personality" rc scripts load it and were cut and paste into other projects, so the following PRs should be merged before this one
- flux-framework/flux-accounting#724
- flux-framework/flux-coral2#391
- flux-framework/flux-sched#1393


chu11 (approved):
This LGTM!

garlick:
Thanks!  Setting MWP.

garlick:
Oh crud, I think I didn't fix this test right:
```
FAIL: t4000-issues-test-driver.t 33 - t2492-shell-lost
```
I replaced a barrier in the tasks with a wait for the `shell.start` event in `guest.exec.eventlog` but that just means the shells have all started, not the tasks.

garlick:
OK, using a PMI barrier now.

PR #6979, opened 2025-08-21
testsuite: remove unused hwloc data, etc.

garlick:
Problem: I happened to notice that some hwloc data and test programs exist in the test suite with no users.

Drop the unused bits and rename the hwloc data that remains to reflect changed circumstances.


chu11 (approved):
LGTM!

garlick:
I'll set MWP - thanks!

PR #6981, opened 2025-08-21
flux-content: refactor checkpoints list command

chu11:
Problem: In the near future we will have multiple checkpoint related commands within flux-content.  Unfortunately the checkpoint list command was titled "checkpoints".

Solution: Refactor the "flux content checkpoints" command to "flux content checkpoint list", so "list" is a subcommand of the "checkpoint" subcommand.

Update all tests accordingly.

----

I decided to split this w/ a follow on PR for `flux content checkpoint update` command.  I guess I could have merged them together, but since they are doing two separate changes that can affect tests, decided to split em.


garlick (approved):
LGTM

PR #6982, opened 2025-08-21
flux-content: support checkpoint update command

chu11:
Problem: There is no way to update the content's KVS checkpoint
on the command line.

Support a new "flux content checkpoint update" command.

Built on #6981.  based on conversation in #6953..  

I could have supported options to update checkpoint w/ a sequence number or timestamp.  Elected not to for the time being b/c didn't think that was important right now.

### DIFF
--- a/src/cmd/builtin/content.c
+++ b/src/cmd/builtin/content.c
@@ -250,6 +250,53 @@ static int checkpoint_list (optparse_t *p, int ac, char *av[])
     return (0);
 }
 
+static bool kvs_is_running (flux_t *h)
+{
+    flux_future_t *f;
+    bool running = true;
+
+    if ((f = flux_kvs_getroot (h, NULL, 0)) != NULL
+        && flux_rpc_get (f, NULL) < 0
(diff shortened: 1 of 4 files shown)


grondo:
I didn't review this, but just noticed some of the commit messages (and PR title) reference `checklist` vs `checkpoint`, is that intentional?

chu11:
> I didn't review this, but just noticed some of the commit messages (and PR title) reference checklist vs checkpoint, is that intentional?

Doh!  Thanks  (and I noticed I typoed the branch name too, but I'll leave that)

garlick:
The references to flux-fsck in the PR title and some commits also probably needs to be changed to flux-content?

chu11:
> The references to flux-fsck in the PR title and some commits also probably needs to be changed to flux-content?

Doh!  Thanks

chu11:
i just realized I forgot to document this and the prior change to "checkpoints" -> "checkpoint list".  I'll do that in a follow on PR.

PR #6983, opened 2025-08-22
support drop-in scripts in `$libexecdir/flux/{prolog,epilog,housekeeping}.d`

### LINKED ISSUE #6699: mandate prolog/epilog/housekeeping.d directories?
Problem: `/etc/flux/system/{prolog,epilog,housekeeping}` scripts are site managed, so packaged flux components cannot provide drop-in files for the `*.d` directories, since it's not guaranteed that `*.d`  will be supported.

I was just thinking about cleaning up slingshot NIC state in housekeeping in the context of flux-framework/flux-coral2#24
and how that package could provide a drop-in script for that.   Without a package-provided drop in, updating the package might require sys admins to update stuff in config-management in lock-step.

So maybe flux-core could just provide those scripts.

grondo:
This PR adds a wrapper script for handling the execution of prolog,epilog,housekeeping instead of executing `/etc/flux/system/{prolog,epilog,housekeeping}` directly from the `flux-{prolog,epilog,housekeeping}@JOBID` systemd units. The wrapper script (installed at `$libexecdir/flux-run-system-scripts`) executes scripts in the following order for `name` in `prolog`, `epilog`, `housekeeping`:

 1. All package provided scripts at `$libexecdir/flux/{name}.d`
 2. If `/etc/flux/system/{name}` exists, then this script is executed directly for backwards compatibiliy, and step 3. is skipped.
 3. All system provided scripts at `/etc/flux/system/{name}.d`

This preserves behavior on existing systems while still allowing packages to install drop-in scripts into the system that will be executed as part of system instance prolog, epilog, or houskeeping. Documentation in the admin guide has been updated to reflect the changes.

By default, `flux-run-system-scripts` exits with the largest exit code of all scripts after executing everything. This behavior may be modified via the `job-manager.{name}.exit-on-first-error` key (set to `true` to exit on the first encountered error)

A trivial amount of testing is added to the system test workload. In order to get the tests to work reliably, a couple minor changes to the system test scripts was required.

This is WIP to get some feedback before adding a few more tests.

Fixes #6699

### DIFF
--- a/t/system/0006-perilog.t
+++ b/t/system/0006-perilog.t
@@ -1,0 +1,96 @@
+#
+#  Check prolog run via IMP
+#
+IMP=$(flux config get exec.imp)
+test -n "$IMP" && test_set_prereq HAVE_IMP
+
+fluxcmddir=$(pkg-config --variable=fluxcmddir flux-core)
+libexecdir=$(pkg-config --variable=fluxlibexecdir flux-core)
+confdir="/etc/flux/imp/conf.d"
+
+test_expect_success HAVE_IMP 'configure a system prolog' '
+	cleanup "sudo rm -f $confdir/imp-prolog.toml" &&
+	cat <<-EOF >imp-prolog.toml &&
+	[run.prolog]
+	allowed-users = ["flux"]
+	allowed-environment = ["FLUX_*"]
+	path = "$fluxcmddir/flux-run-prolog"
+	EOF
+	sudo cp imp-prolog.toml $confdir &&
+	sudo chmod 644 $confdir/imp-prolog.toml &&
+	ls -l $confdir/imp-prolog.toml &&
+	cat <<-EOF >site-prolog &&
+	env
+	EOF
+	sudo mkdir -p /etc/flux/system/prolog.d &&
+	cleanup "sudo rm -rf /etc/flux/system/prolog.d" &&
+	sudo cp site-prolog /etc/flux/system/prolog.d &&
+	sudo chmod 755 /etc/flux/system/prolog.d/site-prolog &&
+	cat <<-EOF >pkg-prolog &&
+	sleep 0
+	EOF
+	sudo cp pkg-prolog $libexecdir/prolog.d &&
+	cleanup "sudo rm -f $libexecdir/prolog.d/pkg-prolog" &&
+	sudo chmod 755 $libexecdir/prolog.d/pkg-prolog
+'
+test_expect_success HAVE_IMP 'configure the perilog plugin to run prolog' '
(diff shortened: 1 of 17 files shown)


garlick:
This looks great to me based on a quick pass.  Maybe it would be a good idea to make sure @kkier is in the loop on this one?

grondo:
Thanks for mentioning @kkier. The key thing to note is that the only change in behavior while `/etc/flux/system/{prolog,epilog,housekeeping}` exist is that any scripts dropped into `libexecdir/flux/{name}.d` (meant to be used by other packages) will be run before the system scripts.

Eventually sysadmins may want to transition to allowing Flux to run scripts directly out of `/etc/flux/system/{name}.d` by moving stuff from `/etc/flux/system/{prolog,epilog,housekeeping}` into those `.d` directories.

grondo:
This has some somewhat extensive testing now, so dropping the WIP.

kkier:
I have concerns itching in the back of my head but I can't fully articulate them yet - something about layers of wrappers* and  being worried about order of operations stuff.  It's a good change, I'm just wanting to be sure I understand all the potential things to watch for/test.

_*I typed "layers of rappers" here at first and the image it conjured up was awesome._

grondo:
Thanks @kkier. Happy to take any suggestions to ease your concerns.

As far as order of operations, any package provided scripts in `$libexecdir/flux/{name}.d` are always run _before_ `/etc/flux/system/{name}` (if it exists) or the scripts in `/etc/flux/system/{name}.d/*` if `/etc/flux/system/{name}` does not exist.

Think of this as an added necessary step to allow packages like flux-coral2 to install work that will get done during the prolog/epilog/housekeeping phases of a job.

t/system/0006-perilog.t
```diff
@@ -0,0 +1,96 @@
...
+	allowed-users = ["flux"]
+	allowed-environment = ["FLUX_*"]
+	path = "$fluxcmddir/flux-run-prolog"
+	EOF
+	sudo cp imp-prolog.toml $confdir &&
+	sudo chmod 644 $confdir/imp-prolog.toml &&
+	ls -l $confdir/imp-prolog.toml &&
+	cat <<-EOF >site-prolog
+	env
+	EOF
```
garlick: Missing && here?

Maybe HAVE_IMP isn't set for the chain-lint check in CI

grondo: Oh yeah, thanks! We don't have a `--chain-lint` variant for the system tests :thinking:

grondo:
Fixed the missing `&&` and will set MWP.

Issue #6984, opened 2025-08-25
`resource.rediscover=true` inadvertently disables batch/alloc  job expiration updates


grondo:
John G. noticed that updating the expiration of a running batch or alloc Flux instance started with `--conf=resource.rediscover=true` didn't affect jobs within the instance.

I can easily reproduce this. When updating the duration of a normal instance, a `resource-update` event appears in the resource eventlog:
```console
$ id=$(flux alloc -t5m -N1 --bg)
$ sudo flux update $id duration=+1m
$ flux proxy $id flux resource eventlog -H
[Aug25 11:26] restart ranks="0" online="" nodelist="pi2"
[  +0.000204] resource-define method="job-info"
[  +0.964062] online idset="0"
[ +21.728826] resource-update expiration=1756146770.0
```

However, with `resource.rediscover`, the event is never posted:
```console
$ id=$(flux alloc -t5m -N1 --bg --conf=resource.rediscover=true)
$ sudo flux update $id duration=+1m
$ flux proxy $id flux resource eventlog -H
[Aug25 11:29] restart ranks="0" online="" nodelist="pi2"
[  +0.000421] resource-define method="dynamic-discovery"
[  +0.903024] online idset="0"
```

Likely the forced local rediscovery of resources is bypassing the resource watch that the Flux `resource` module usually initiates with the parent instance.

PR #6985, opened 2025-08-25
resource: fix expiration updates with `resource.rediscover=true`

grondo:
When the `resource.redicover` broker config key is set to `true`, this disables the KVS watch of R in the parent, thus preventing the instance from seeing expiration updates.

This PR fixes that issue, thereby fixing #6984.

### DIFF
--- a/t/t2292-job-update-running.t
+++ b/t/t2292-job-update-running.t
@@ -156,6 +156,14 @@ subinstance_get_job_expiration() {
 	flux proxy $1 job_manager_get_R $2 | jq '.R.execution.expiration'
 }
 
+# Run the following tests with and without `--conf=resource.rediscover=true`
+# N.B.: to keep indent sane, the contents of this for loop are not indented:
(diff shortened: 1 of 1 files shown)


src/modules/resource/inventory.c
```diff
@@ -626,23 +626,26 @@ static int start_resource_watch (struct inventory *inv,
...
     if (R && !config->rediscover) {
         if (inventory_put (inv, R, "job-info") < 0)
             goto done;
-        if (flux_future_then (f,
-                              -1.,
-                              R_update_cb,
-                              inv) < 0)
-            flux_log (h, LOG_ERR, "Failed to register callback for R updates");
-    }
+   }
```
watson6282: I believe a space was accidentally deleted here.

grondo: Thanks for catching that! I just force-pushed a fix.

garlick (approved):
LGTM!

grondo:
Thanks! I'll set MWP.

Issue #6311, opened 2024-09-26
t4000-issues "free-range-test" failures in CI


grondo:
This test has been failing sporadically in CI. 

The test kills rank 3 of a size=4 broker with SIGKILL and expects the instance to continue to be able to run a 3 node job.
This part seems to work, but the instance sometimes exits with wait status=9 (Killed) instead of the expected status=0.

```
  free-range-test: Re-launching test script under flux-start
  free-range-test: Starting a child instance with flat topology
  free-range-test: Started job fuKmers
  free-range-test: Current overlay status of fuKmers:
  0 fv-az1196-271: full
  ├─ 1 fv-az1196-271: full
  ├─ 2 fv-az1196-271: full
  └─ 3 fv-az1196-271: full
  free-range-test: Launch a sleep job within fuKmers:
  f2A3iXBV
         JOBID USER     ST NTASKS NNODES  RUNTIME
             . runner    R      4      4   6.352s .
       fuKmers runner    R      4      4   3.782s └── flux
      f2A3iXBV runner    R      4      4   0.404s     └── sleep
  free-range-test: Killing rank 3 (pid 146203) and all children
  free-range-test: Wait for exception event in fuKmers
  flux-start: 3 (pid 146203) Killed
  Sep 26 03:27:30.610493 UTC broker.err[0]: fv-az1196-271 (rank 3) failed
  1727321250.599241 exception type="node-failure" severity=2 userid=1001 note="shell rank 3 (on fv-az1196-271): Killed"
  free-range-test: But running a 3 node job in fuKmers still works:
  Sep 26 03:27:30.713759 UTC broker.err[0]: dead to Flux: fv-az1196-271 (rank 3)
  fv-az1196-271
  fv-az1196-271
  fv-az1196-271
  free-range-test: Overlay status of fuKmers should show rank lost:
  0 fv-az1196-271: degraded
  ├─ 1 fv-az1196-271: full
  ├─ 2 fv-az1196-271: full
  └─ 3 fv-az1196-271: lost lost connection
  free-range-test: Call flux shutdown on fuKmers
  free-range-test: job fuKmers should exit cleanly (no hang) and a zero exit code:
  1727321253.658158 finish status=9
  free-range-test: dump output from job:
  
  4.081s: job.exception type=node-failure severity=2 shell rank 3 (on fv-az1196-271): Killed
  flux-job: job shell Killed
[...]

PR #6986, opened 2025-08-25
test/terminus: skip kill tests due to race

trws:
problem: There's what seems to be a race condition in the terminus wait setup that can occur if the spawned process dies too fast. This triggers very frequently on MacOS.

solution: skip these tests there for now.

I'm creating a new issue for the race condition because it seems very likely it's possible on Linux as well under the right conditions. See [this comment](https://github.com/flux-framework/flux-core/pull/6929#issuecomment-3137916572) for more details.

### DIFF
--- a/src/common/libterminus/test/terminus.c
+++ b/src/common/libterminus/test/terminus.c
@@ -274,6 +274,11 @@ static void test_protocol (void)
 
     /* kill session
      */
+    bool on_apple = false;
+    #ifdef __APPLE__
+    on_apple = true;
+    #endif
+    skip (on_apple, 11)
     f = flux_rpc_pack (h,
                        "terminus.kill",
                        0, FLUX_RPC_STREAMING,
@@ -357,6 +362,7 @@ static void test_protocol (void)
     ok (rc == 0,
         "terminus.kill (no wait): OK");
     flux_future_destroy (f);
+    end_skip; // __APPLE__
 
 
     /* kill-server


grondo (approved):
LGTM!

grondo:
Is this the reason CI keeps failing in the macos build since #6929 was merged?

trws:
It should be yes, I actually saw it stuck on this test and was fixing it when I got a notification it merged.  One of the mergify trains must have succeeded I guess? Still head-scratching on that one.  Why this has failures though I need to see, looking now.

grondo:
Yeah, merge-when-passing was set and I guess all tests happened to pass together :sweat_smile:

trws:
The issue is live by the way at #6987.

Fedora40 timed out, may want to give that one a bit more time, but there seems to be an intermittent failure in the issue test driver on free-range-test that I've now seen on both focal and el9.

trws:
Gathering info on the `free-range-test` issue, both in el9 and focal the exit was a status of 9 rather than 0:

```
// seemingly normal to here
  Aug 25 21:04:30.763315 UTC 2025 broker.err[0]: cleanup.2: flux-cancel: Canceled 1 job (0 errors)
  0 bc785f9bf652: degraded
  ├─ 1 bc785f9bf652: full
  ├─ 2 bc785f9bf652: full
  └─ 3 bc785f9bf652: lost lost connection
  free-range-test: Call flux shutdown on fUJMmyH
  free-range-test: job fUJMmyH should exit cleanly (no hang) and a zero exit code:
  1756155871.775541 finish status=9
  free-range-test: dump output from job:
```

grondo:
> Gathering info on the free-range-test issue, both in el9 and focal the exit was a status of 9 rather than 0:

I have seen that one fail before when the test is running very slow (reported in #6311). I don't think I quite ever got a fix for it though and it is quite rare (not related to this or the macos PR I don't think)

trws:
I mentioned it mainly because it showed up in two actions sets in a row on two different PRs for me today. Given the timeouts, it may just be that things are running extra slow, but something must have tickled it.

PR #6988, opened 2025-08-26
drop `rc1` and `rc3` scripts

grondo:
This PR adds a task to modprobe's `rc1.py` and `rc3.py` to handle the `rcX.d` directories so that the shell code to do this in `rc1` and `rc3` can be dropped. The code to fall back to `rc1.old` and `rc3.old` is then moved into `flux-modprobe rcX`, and finally the default `broker.rc1_path` and `broker.rc3_path` attributes are updated to `flux modprobe rc[1|3]`.

After this, the `rc1` and `rc3` scripts are no longer used, so they're dropped.

Not sure if we want to update the `broker.rcX_path` attributes to be `broker.rcX_command`, but I wasn't sure what that would potentially break so I didn't make that change here.


grondo:
Thanks! I suppose I'll set MWP here.

PR #6991, opened 2025-08-26
t: add tests for fsck of invalid dirref entry

chu11:
Problem: There is no fsck coverage for a corrupted / invalid dirref treeobj in the KVS.

Add coverage in t2816-fsck-cmd.t.


garlick (approved):
LGTM!

PR #6992, opened 2025-08-26
testsuite: various fixes and extra debug

grondo:
The testsuite has become pretty flaky in CI in the last week.

This PR adds some obvious fixes and some extra debug for some tests that fail in CI without a clear cause.

### DIFF
--- a/t/issues/t2492-shell-lost.sh
+++ b/t/issues/t2492-shell-lost.sh
@@ -61,7 +61,8 @@ for rank in 3 1; do
     log "Submitted job $id. Waiting for shell rank $rank to be lost"
 
     value="shell rank $rank (on $(hostname -s)): Killed"
-    flux job wait-event -Wt 15 -Hvp output -m message="$value" $id log
+    flux job wait-event -Wt 120 -Hvp output -m message="$value" $id log \
+	|| die "failed to get shell lost message"
 
     log "Sending SIGINT to $id. Job should now exit"
     flux job kill --signal=2 $id
--- a/t/t2801-top-cmd.t
+++ b/t/t2801-top-cmd.t
@@ -24,7 +24,7 @@ job_list_wait_state() {
 # lead to racy output with tests below.  Remove the heartbeat module to
 # remove this possibility.
 test_expect_success 'set high heartbeat period' '
-	flux module remove heartbeat
+	flux exec flux module remove heartbeat
 '
 test_expect_success 'flux-top -h prints custom usage' '
 	flux top -h 2>usage &&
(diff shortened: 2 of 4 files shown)


t/issues/t2492-shell-lost.sh
```diff
@@ -67,8 +67,8 @@ for rank in 3 1; do
     flux job kill --signal=2 $id
     flux job attach -vEX $id
     rc=$?
-    log "Job exited with rc=$rc (expecting 137 (128+9))"
-    test $rc -eq 137 || die "Unexpected job exit code $rc"
+    log "Job exited with rc=$rc (expecting 137 (128+9) or 130 (128+2))"
+    test $rc -eq 137 -o $rc -eq 130 || die "Unexpected job exit code $rc"
```
chu11: were you able to figure out why this is racy?  The comments in the test seems to clearly want to avoid a SIGINT race.  So probably needs more comments to be clear why SIGINT is ok.

chu11: Is the issue this:

```
    flux job wait-event -Wt 15 -Hvp output -m message="$value" $id log
```

15 second timeout?  Perhaps increase the timeout?

chu11: not related to this PR, but I did notice this typo since you're in the same area

```    log "Sumbmitted job $id. Waiting for shell rank $rank to be lost"```

"Sumbmitted" .... hmmm, wonder why the typo checker didn't catch this. (ohh it's probably not in its database of typos)

Edit: i'll deal with this along with a update to the typos checker, it caught a few

Comment (grondo) [reply t/issues/t2492-shell-lost.sh]: D'oh, you're right. I confused myself about how the test was supposed to work.
Let's try the extended timeout as you suggest. Also the test should exit immediately on `wait-event` failure.

Comment (grondo) [reply t/issues/t2492-shell-lost.sh]: Done.

File: t/t2801-top-cmd.t [hunk 2]
```diff
@@ -24,7 +24,7 @@ job_list_wait_state() {
 # lead to racy output with tests below.  Remove the heartbeat module to
 # remove this possibility.
 test_expect_success 'set high heartbeat period' '
-	flux module remove heartbeat
+	flux exec flux module remove heartbeat
 '
```
garlick, line 28: Oh hadn't thought about that consequence of the timeouts!  Good catch.

garlick (approved):
LGTM!

grondo:
Thanks @chu11 and @garlick! Will set MWP.

PR #6999, opened 2025-08-27
t: rename test dirs for fsck tests

chu11:
Problem: The test directories in t2816-fsck-cmd.t are simply named "dir", but things can get confusing when comments describe
treeobjs of type "dir".

Rename all "dir" paths to "testdir".

----

splitting off from bigger PR b/c it's quite "noisy" with a lot of search and replace changes.


chu11:
> I wonder though if we could just create bad treeobjs directly instead of laboriously modifying good ones? Could even have some prepared JSON in test data files. Just a thought. This is fine for now.

What would be super easy is if we had a `flux content remove` mechanism ... What's your feeling on implementing that as a hidden test only feature?  Obviously only instance owner could call it.

PR #7000, opened 2025-08-27
Testsuite: fix race in t2315

sam-maloney:
Problem: a couple of tests have race conditions and were consistently failing on my Rocky9 VM (I've no idea why they just started failing now; they've always run fine in the past...)

1. Use `flux run` to wrap `flux resource list` query in t2315 to ensure resources are "up" when queried, or query state "all" so that it shouldn't matter if they aren't "up" yet.

~~2. Add a small delay to `late-joiner` test script in t0025.~~ (fixed by #7001)

Also a few minor formatting tweaks for consistency with other tests.

### DIFF
--- a/t/t2315-resource-system.t
+++ b/t/t2315-resource-system.t
@@ -127,7 +127,7 @@ test_expect_success 'ranks can be excluded by configuration' '
 	EOF
 	flux start -s 2 \
 		--config-path=$(pwd)/${name} -Slog-filename=${name}/logfile \
-		flux resource list -s up -no {nnodes} > ${name}/nnodes &&
+		flux run flux resource list -s up -no {nnodes} > ${name}/nnodes &&
 	test "$(cat ${name}/nnodes)" = "1"
 '
 
@@ -142,7 +142,8 @@ test_expect_success 'invalid exclude ranks cause instance failure' '
 	EOF
 	test_must_fail flux start -s 2 \
 		--config-path=$(pwd)/${name} -Slog-filename=${name}/logfile \
-		flux resource list -s up -no {nnodes} > ${name}/nnodes &&
+		flux resource list -s all > ${name}/output 2>&1 &&
+	test_debug "cat ${name}/output" &&
     grep "out of range" ${name}/logfile
 '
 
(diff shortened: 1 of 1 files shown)


t/t0025-broker-state-machine.t
```diff
@@ -78,6 +78,7 @@ test_expect_success 'create rc1 that blocks on FIFO for rank != 0' '
 test_expect_success 'create rc2 that unblocks FIFO' '
 	cat <<-EOT >rc2_unblock &&
 	#!/bin/bash
+	sleep 1
 	${GROUPSCMD} get broker.online
```
grondo: I'm actually confused on this one. Can rc2 start without rank 0 in the `broker.online` group?
This might have been introduced by #6959, so I'm going to defer to @garlick on this one.

If things are working as expected, then I suggest replacing the `sleep 1` with `${GROUPSCMD} waitfor 1` or similar (with a timeout of course)

garlick: Oh!  That was probably not intended!  Let me look into that.

garlick: Argh, yes in a size=1 instance, because of  d4f40e8e64654f16d671be8f8c5a2c152bf0565b.

Perhaps that should just be reverted.  I'll put up a PR.

garlick: Thanks guys - sorry that consequence had not occurred to me.

@sam-maloney - thanks for your efforts!   I did want to make one comment on this PR:  we usually try to keep white space changes in separate commits from fixes.  More generally, it's "a commit should fix one thing", but also it's just to reduce the review effort - not a big deal in this small PR obviously.

sam-maloney: @garlick that makes sense, thanks for the feedback!

Otherwise, does your new PR #7001 fix the underlying problem in t0025 entirely? If so, once that is merged in I can drop that commit and retest.

garlick: I _think_ it should be fixed.  Retesting is a good idea though!

sam-maloney: t0025 passes without issue now, so I've dropped that commit and split the whitespace changes into their own commit.

garlick: what about t2315? I  assumed that one would be fixed too since `broker.online` is monitored by the resource module to determine what's up.

sam-maloney: Mea culpa, I didn't realize they were both the same root cause! But yes, both are passing without modification now; I'll close this :)

garlick: Thanks again for running this down.  I'm glad you pressed the issue before the upcoming tag!

PR #7001, opened 2025-08-28
Revert "broker: short circuit groups when quorum.size == 1"

garlick:
Problem: sometimes the broker enters RUN state without rank 0 membership in the broker.online group.

A recent attempt to skip unnecessary work during startup of size=1 instances unintentionally introduced a race in tests that expect the resources of rank 0 to be online immediately.

This was reported in #7000

This reverts commit d4f40e8e64654f16d671be8f8c5a2c152bf0565b.


grondo (approved):
LGTM!

PR #7004, opened 2025-08-28
modprobe: ensure required modules of activated alternatives are also active

grondo:
This bug was uncovered while doing some other modprobe work:

Modules that are selected as the current alternative for some service, e.g. `sched`, do not automatically have their `requires` list activated. If the module also has a `needs` on a module it `requires`, this can end up _deactivating_ the selected alternative since its `needs` is not met.

Automatically activate modules in the required list of an activated module. Also, add a couple new tests to check the expected behavior.


garlick (approved):
LGTM!

grondo:
Thanks! Setting MWP.

PR #7005, opened 2025-08-29
switch to modprobe by default for rc1 and rc3

grondo:
This PR drops the fallback rc scripts `rc1.old` and `rc3.old` and effectively makes modprobe the default (with no way to fall back to the old rc scripts). For now, `flux modprobe` sets `FLUX_RC_USE_MODPROBE=t` for `rc1` and `rc3` in order to avoid breaking other projects. Eventually these projects can also drop the old rc scripts if they are currently skipped when modprobe is active.

Additionally, the testsuite is updated to drop the wrapper "personality" rc scripts in `t/rc/rcX-personality` in favor of just using `flux modprobe run` directly.

Since this changes the default startup/shutdown mechanism of all Flux instances, we should consider whether we want to put this in before the next release, or just after tagging.


garlick:
Well I have not found a single thing wrong with this stuff and I'm for going full speed ahead with this release.  I guess it comes down to how comfortable you feel about that @grondo.  I'm fine either way.

garlick (approved):
LGTM!

grondo:
Ok, I've done some testing and now feel confident that this can go in. Rebased on current master and will set MWP.

PR #7008, opened 2025-08-29
testsuite: fix barrier in shell-lost issues test

grondo:
Problem: The `flux pmi barrier` call in t2492-shell-lost.sh is not functional because Python's `subprocess.run()` is closing the PMI_FD. This causes the barrier to exit prematurely and results in a race where the test job can get killed by SIGINT before it is handling the signal.

Add `close_fds=False` to the `subprocess.run()` call so that the PMI fd is not closed and the barrier functions properly.


chu11 (approved):
Ahhh, that was a probably a tough one to hunt down!

garlick:
Oof my bad for inserting that and somehow managing to not notice it wasn't working.   Thanks!

garlick:
@mergifyio requeue

Issue #6990, opened 2025-08-26
broker: redirecting content store via `statedir` attribute is inconvenient


grondo:
Problem: It is inconvenient to redirect the content store with the `statedir` attribute when submitting multiple jobs that may run at the same time (if `statedir` will be in a shared filesystem), since `statedir` must exist before the job starts and it must be unique for every job.

In the case where a user is filling `/tmp`, we should have a convenient way to point `statedir` to a directory that will be guaranteed to be unique. Offering the ability to substitute `{{id}}` seems enticing, but then the broker would need to create `statedir` when it does not exist.

@ryanday36 also suggested that if users need to redirect `statedir` in this way due to content store growth, they shouldn't have think about using `{{id}}`. Since a jobid is not always available and also there may be use cases where a user would want to restart a job from an existing sqlite file, there is not a simple solution for all here (or I'm not being creative enough).

ryanday36:
One other thing possibly related this:

Another idea I just had for at least the way that we're currently running the CORAL2 systems with the rabbits mounted as local disk on each node (as/l/ssd) was to try to put the statedir there. Unfortunately:
```
[day36@tuolumne1003:~]$ flux alloc -N1 -q pdebug --broker-opts=-Sstatedir=/l/ssd/
flux-job: f2Fwie3x9WdM started                                                                                                                                                               00:00:04
broker: statedir /l/ssd/ is not owned by instance owner: Operation not permitted
[detached: session exiting]
[day36@tuolumne1003:~]$
```
That /l/ssd is owned by root, but is writable by the user:
```
[day36@tuolumne1400:~]$ ls -lah /l/ssd
total 0
drwxrwxrwt 2 root root  6 Aug 15 13:15 .
drwxr-xr-x 3 root root 60 Aug 18 14:01 ..
```
Mark pointed out that the check is probably to help ensure that no-one else can write a statedir that messes with the user's job, but if the broker defaulted to creating a unique statedir for each job that could allow more flexibility in where the statedir is located.

garlick:
As noted [here](https://systemd.io/TEMPORARY_DIRECTORIES/):
> `/tmp/` and `/var/tmp/` are two world-writable directories Linux systems provide for temporary files. The former is typically on tmpfs and thus backed by RAM/swap, and flushed out on each reboot. The latter is typically a proper, persistent file system, and thus backed by physical storage. This means:
[...]

PR #7009, opened 2025-08-30
broker: interpret sticky bit in -Sstatedir=DIR

garlick:
Problem: as discussed in #6990, it's somewhat inconvenient to make the user specify a unique path when they submit a job that should use an alternate location for `statedir`.

If  `-Sstatedir=DIR` refers to a `/tmp` like directory (a directory with the sticky bit set), then a unique directory is created with in it.

So on CORAL-2 systems where rabbits are mounted on `/l/ssd`, one could specify `--broker-opts=-Sstatedir=/l/ssd` to ensure the content-sqlite db doesn't eat up memory in the tmpfs ramdisk on rank 0.

### DIFF
--- a/t/t0036-broker-rundir.t
+++ b/t/t0036-broker-rundir.t
@@ -112,6 +112,20 @@ for attr in rundir statedir; do
 	test_must_fail flux start ${ARGS_NORC} flux setattr $attr x &&
 	test_must_fail flux start ${ARGS_NORC} flux setattr $attr-cleanup x
     '
+    test_expect_success "-S$attr=DIR interprets DIR sticky bit" '
+	dirname=`mktemp -d` &&
+	chmod 1777 $dirname &&
+	flux start ${ARGS_NORC} -S$attr=$dirname \
+		./checkattrs.sh >$attr-sticky.out
+    '
+    test_expect_success "$attr/DIR was used" '
+	grep "^$attr=$dirname/" $attr-sticky.out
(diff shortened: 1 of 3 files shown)


t/t0036-broker-rundir.t
```diff
@@ -112,6 +112,20 @@ for attr in rundir statedir; do
...
+		./checkattrs.sh >$attr-sticky.out
+    '
+    test_expect_success "$attr/DIR was used" '
+	grep "^$attr=$dirname/" $attr-sticky.out
+    '
+    test_expect_success "$attr/DIR was cleaned up but not $attr" '
+        subdirname=$(kv_lookup $attr $attr-sticky.out) &&
+	test $(dirname $subdirname) &&
+	test ! -e $subdirname
+    '
```
grondo: Is it just the github diff or is there some missing tabs here?

Also, does `test $(dirname $subdirname)` do anything, do you need `test -d` instead? I'm actually not sure what happens with just `test /path/to/dir`, for me it just silently succeeds.

grondo (approved):
LGTM! Just one comment in the tests.

t/t0036-broker-rundir.t
garlick: Oh good catch on both!

garlick:
Fixed those things and a typo that CI caught and I'll set MWP.  Thanks!

PR #7010, opened 2025-08-30
flux-start: suppress module panic in recovery mode

garlick:
Problem: when a recovery instance is started with `flux start --recovery`, a spurious module exit could cause the recovery instance to shut down unexpectedly.

Since we're in a degraded state, add `-Sbroker.module-nopanic=1` to the broker options that are set in recovery mode.


garlick:
Thanks, I'll set MWP

Issue #7011, opened 2025-08-30
flux modprobe unload doesn't unload all dependencies


garlick:
Problem:  `flux modprobe load cron` loads `heartbeat` implicitly, but `flux modprobe remove cron` doesn't unload it:

```
$ flux start -Sbroker.rc1_path= -Sbroker.rc3_path=
$ flux module list
Module                   Idle  S Sendq Recvq Service
connector-local             0  R     0     0 
groups                      1  R     0     0 
$ flux modprobe load cron
$ flux module list
Module                   Idle  S Sendq Recvq Service
heartbeat                   1  R     0     0 
groups                     17  R     0     0 
cron                        9  R     0     0 
connector-local             0  R     0     0
$ flux modprobe list-dependencies cron
cron
└── heartbeat
$ flux modprobe remove  cron
$ flux module list
Module                   Idle  S Sendq Recvq Service
heartbeat                   1  R     0     0 
groups                     29  R     0     0 
connector-local             0  R     0     0 
```

I'm not spotting why that would be the case.  Other modules loaded as dependencies  are automatically unloaded.

```toml
[[modules]]
name = "heartbeat"
[[modules]]
name = "cron"
ranks = "0"
requires = ["heartbeat"]
args = ["sync=heartbeat.pulse"]
```

I first noticed this when trying to modprobe load/remove `resource` in a test.  `heartbeat` was loaded as a dependency, and while removal automatically took care of other dependent modules, `heartbeat` was left behind.

PR #7015, opened 2025-09-03
modprobe: various fixes and improvements

grondo:
This PR started with a fix for #7011, but that fix resulted in some new issues that needed to be addressed, and so on. :disappointed: 

Finally got all the issues addressed, though there are still some corner cases in `flux module remove` that are not handled all that well. For example if you `flux modprobe remove sched feasibility` instead of `flux modprobe remove sched-fluxion-qmanager sched-fluxion-feasibility`, `sched-fluxion-resource` is not auto-unloaded since the references are not currently tracked through the module service names.

Will open an issue on that one separately.

### DIFF
--- a/etc/completions/flux.pre
+++ b/etc/completions/flux.pre
@@ -2147,8 +2147,12 @@ _flux_modprobe()
     local cmd=$1
     local subcmds="load remove run rc1 rc3 list-dependencies show"
     local split=false
-    load_OPTS=""
-    remove_OPTS=""
+    load_OPTS="\
+        --dry-run \
+    "
+    remove_OPTS="\
+        --dry-run \
+    "
     run_OPTS="\
         -v --verbose \
         --show-deps \


doc/man1/flux-modprobe.rst
```diff
@@ -69,7 +73,13 @@ remove
 
 .. program:: flux modprobe remove
 
-:program:`flux modprobe remove` unloads one more broker modules.
+.. option:: --dry-run
+
+  Do not actually run anything, but print the tasks that would be run.
+
+:program:`flux modprobe remove` unloads one more broker modules, including
+any modules that are unused after the target modules are removed.
```
garlick: One or more?

garlick (approved):
One small nit, but o/w LGTM.
I did repeat my original test of `modprobe load resource` followed by `modprobe remove resource` and everything cleaned up properly.

grondo:
Thanks! Addressed the one comment and also added a patch to bash completions for the new `--dry-run` options of `flux modprobe load/remove`. I'll set MWP.

PR #7016, opened 2025-09-03
flux-module: expand module name column in list output

grondo:
Noticed that `sched-fluxion-feasibility` is being truncated in `flux module list` output. Not a big deal except that bash completions use this output to get module names to complete, so I kept getting `sched-fluxion-feasibilit`.

In the hopes that this is the longest module name ever, this PR adds 1 character to the `flux module list` 1st column.

### DIFF
--- a/src/cmd/flux-module.c
+++ b/src/cmd/flux-module.c
@@ -493,12 +493,12 @@ void lsmod_print_header (FILE *f, bool longopt)
 {
     if (longopt) {
         fprintf (f,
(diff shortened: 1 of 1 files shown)


grondo:
Alternately we could make the column auto-sized to the largest module name. However, I left that exercise for another time.

garlick (approved):
LGTM!

grondo:
Thanks!

PR #7017, opened 2025-09-03
testsuite: improve cleanup in t2412-sdmon.t

grondo:
On my test system I was getting failures in `t2412-sdmon.t` because an old `shell-t412.service` unit was in a failed state in my systemd user unit. The `systemctl --user stop` command didn't seem to clean up the unit, so subsequent tests were failing.

This PR just adds an additional `systemctl --user reset-failed <unit>` to the part of the test that tries to ensure residual units are not running. This seems to resolve the issue on my system, but we'll see if it passes in CI.

### DIFF
--- a/t/t2412-sdmon.t
+++ b/t/t2412-sdmon.t
@@ -38,6 +38,10 @@ start_test_unit() {
 stop_test_unit() {
 	systemctl --user stop $1
 }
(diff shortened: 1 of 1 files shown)


garlick (approved):
Oh thank you! I've hit that one too on occasion.

Issue #7003, opened 2025-08-28
job-manager.err[0]: Failed to drain ranks with failed prolog for <jobid>: idset is empty


garlick:
Problem: we recently had a segfault of the rank 0 broker on El Cap, foreshadowed by the log message in the issue title.

grondo:
Though it's mystifying how the idset became empty in this code path (will look into it), one improvement here would be to avoid sending a drain RPC altogether when the target idset is empty (perhaps logging a warning instead)

Meanwhile, I have it on my list to try reproducing this error (by modifying the perilog.so plugin) in a test instance to see if it causes a segfault.

grondo:
Ok, I modified the `perilog.so` by hand to always send an empty idset in the drain request then ran a test job with the prolog enabled. No segfault, and even running under valgrind didn't turn anything up:
```
$ src/cmd/flux start --wrap=libtool,e,valgrind,--leak-check=full  -s4 sh -c 'flux jobtap load perilog.so; flux config load t.toml; flux run -N4 hostname; flux resource list'
Aug 28 13:42:54.269260 PDT 2025 job-manager.err[0]: Failed to drain ranks with failed prolog for ƒ3Cuf3AX: idset is empty
0.878s: job.exception ƒ3Cuf3AX type=prolog severity=0 prolog exited with code=1 on  (rank )
     STATE NNODES NCORES NGPUS NODELIST
      free      4     32     0 pi[3,3,3,3]
 allocated      0      0     0 
      down      0      0     0 
$
```

So I don't _think_ the error message here led _directly_ to the crash. We'll have to wait to see the stack trace to get more information.

PR #7018, opened 2025-09-03
perilog: avoid draining empty idset

grondo:
Problem: If the idset is empty is empty in proc_drain_ranks(), then a useless empty `resource.drain` request is sent to the resource module.

Skip the RPC and log a warning in this case.

This is based on the observation reported in #7003. However, I was never able to reproduce a crash by sending the empty idset, perhaps more investigation is needed to determine what may have preceded the empty idset.

Anyway, it seems logical to not bother sending an RPC that is guaranteed to fail. However, there's no good way to test this change, and it is known not to fix any issue, so I'd be fine dropping this PR on that basis.


grondo:
Thanks! I've set MWP.

PR #7025, opened 2025-09-04
testsuite: fix t2622-job-shell-sysmon.t skip_all behavior

grondo:
Problem: `t2622-job-shell-sysmon.t` fails with the following errors on a system without the unified cgroup hieararchy:
```
 ./t2622-job-shell-sysmon.t: 10: test: /sys/fs/cgroup//: unexpected operator
 ./t2622-job-shell-sysmon.t: 12: test_done: not found
 ./t2622-job-shell-sysmon.t: 14: test: /sys/fs/cgroup//: unexpected operator
 ./t2622-job-shell-sysmon.t: 16: test_done: not found
```
This is because `/proc/$$/cgroup` may contain multiple lines on these systems. The `test_done: not found` errors are because this function is used before sourcing `sharness.sh` which provides the shell function.

Fix the `/proc/$$/cgroup` parsing issue by picking only the line in the file prefixed with `0:`.

Move the source of `sharness.sh` up to the top of the file to address the errors using `test_done`.

### DIFF
--- a/t/t2622-job-shell-sysmon.t
+++ b/t/t2622-job-shell-sysmon.t
@@ -2,7 +2,9 @@
 #
 test_description='Test flux-shell sysmon plugin'
 
-cgdir=/sys/fs/cgroup/$(cat /proc/$$/cgroup | cut -d: -f3)
+. `dirname $0`/sharness.sh
+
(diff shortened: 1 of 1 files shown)


garlick (approved):
LGTM!

grondo:
Fixed a commit message typo and will set MWP. Thanks!

PR #7033, opened 2025-09-05
doc: add `attach --tail` to flux-job(1)

wihobbs:
Problem: the `flux job attach --tail` option is undocumented in the manual pages.

Add a note in flux-job(1).


grondo:
Want to check if it is in the bash completions as well?

wihobbs:
Is that as simple as adding an entry in `/etc/completions/flux.pre` on line 708 (under `local attach_OPTS="\`)?

chu11 (approved):
Doh!  Thanks.  I'll approve, but per @grondo comment, lets check the bash completions too.

wihobbs:
@chu11 or @grondo can you check that I did the completions right? I've never messed with them before, and am unfamiliar with how they work.

etc/completions/flux.pre
```diff
@@ -716,6 +716,7 @@ _flux_job()
         -i --stdin-ranks= \
         -u --unbuffered \
         --debug \
+        --tail \
```
grondo: Options that take an argument just need a trailing `=` (e.g. see `--stdin-ranks` above)

wihobbs:
Setting MWP.

doc/man1/flux-job.rst
```diff
@@ -89,6 +89,10 @@ This status line may be suppressed by setting
    for tasks not in **RANKS** will be closed. The default is to broadcast
    stdin to all ranks.
 
+.. option:: --tail=NUM
```
grondo: Trivial, but just noticed this should probably be 

```suggestion
.. option:: --tail[=NUM]
```

to indicate the option argument is optional.

wihobbs:
oops. That dockerfile shouldn't be there.

wihobbs:
@mergifyio refresh

PR #7029, opened 2025-09-05
sysmon: handle hybrid cgroup hierarchy

garlick:
Problem: sysmon assumes the unfied cgroup hierarchy, but some modern systems use the hybrid hierarchy.
    
Use  cgroup support code pulled in from flux-security, as it handles the hybrid cgroup hierarchy
In addition, use some new cgroup helpers to parse `/sys` files instead of embedding one-offs.
    
Also, treat missing memory or cpu `/sys` files as non-fatal warnings, simply disabling one or the other with a warning message.
    
Fixes #7026

As extra credit later on, the oom plugin could be similarly converted.

PR #7035, opened 2025-09-07
use cgroup helpers in shell oom plugin

garlick:
This does the "extra credit", suggested in #7029, of refactoring the shell oom plugin to use cgroup helpers.

The cgroup helpers are improved in the process and a test that didn't work on system with swap is fixed.  The test failure was obscured by the fact that it is skipped if `stress(1)` is not installed.  The test fix involved adding support for configuring `MemorySwapMax`, which I suppose could be useful if anyone runs flux with memory limits on systems with swap enabled.


garlick:
Eh thinking of release notes, this MemorySwapMax part should probably be split into a separate PR.

src/common/libutil/test/cgroup.c
```diff
@@ -55,6 +55,23 @@ void test_memory (struct cgroup_info *cgroup)
...
+          ARRAY_SIZE (keys),
+          "memory.events (unavailable)");
+
+    for (int i = 0; i < ARRAY_SIZE (keys); i++) {
+        int rc;
+        unsigned long long u64 = 0;
+
+        rc = cgroup_key_scanf (cgroup, "memory.events", keys[i], "%llu", &u64);
+        ok (rc == 1, "cgroup_scanf scanned memory.events:%s", keys[i]);
+        diag ("cpu.stat:%s=%llu", keys[i], u64);
```
grondo: typo: s/cpu.stat/memory.events/?

grondo (approved):
LGTM!

t/t2410-sdexec-memlimit.t
```diff
@@ -51,7 +51,7 @@ EOT
 
 cat >getcg.sh <<EOT2
 #!/bin/sh
-CGDIR=/sys/fs/cgroup/\$(cat /proc/\$\$/cgroup | cut -d: -f3)
+CGDIR=/sys/fs/cgroup/\$(cat /proc/\$\$/cgroup | grep ^0: | cut -d: -f3)
```
grondo: This still assumes the mount point of cgroups2  is `/sys/fs/cgroup`. On my system it is `/sys/fs/cgroup/unified`

```sh
$ grep cgroup2 /proc/mounts
cgroup2 /sys/fs/cgroup/unified cgroup2 rw,nosuid,nodev,noexec,relatime,nsdelegate 0 0
```

grondo: This results in the error:

```
DelegateControllers=memory pids
cat: /sys/fs/cgroup//user.slice/user-1001.slice/session-4039.scope/cgroup.controllers: No such file or directory
1..0 # SKIP cgroups memory controller is not enabled
```

Note however, that the real cgroup.controllers appears to be empty on this system for my current cgroup, so the test would be skipped anyway.

garlick: Ah yeah, your code that I borrowed from flux-security handles this case.  Hmm, I wonder if it would be ok to add a little built-in subcommand that can print this path, e.g. `flux cg path` or similar?  I do find myself building those paths by hand when manually poking at cgroups on installed flux.

garlick:
OK, added that and re-titled the PR.

garlick:
Oof, spotted another typo.

grondo:
Ugh. Sorry for being a pain. On my system `t2212-job-manager-plugins.t` now fails here:

```
fHKtQJGo
fHKutHZ9
fHKwNGqV
fHKzLFQB
flux-shell[0]:  WARN: oom: disabling oom detection: no memory.events
flux-shell[0]:  WARN: oom: disabling oom detection: no memory.events
flux-shell[0]:  WARN: oom: disabling oom detection: no memory.events
flux-shell[0]:  WARN: oom: disabling oom detection: no memory.events
not ok 45 - job-manager: get job IDs of valid jobs
#	
#		grep -v reject_id validate-plugin.out | grep -v foo \
#		    | sed -e "s/.*jobid=//" >valid_ids &&
#		test_debug "cat valid_ids" &&
#		test 4 -eq $(wc -l <valid_ids)
#	
expecting success: 
	while ! test_inactive $(cat valid_ids); do echo retry; sleep 0.1; done

retry
retry
retry
retry
retry
retry
retry
retry
retry
retry
retry
retry
retry
retry
retry
retry
...
```

garlick:
Oh sorry, I'll disable the memory controller on a test system and reproduce.

grondo:
It may be due to the new warning. I wonder if the warning is really necessary since user's do not explicitly request this plugin be activated. Perhaps a debug message would be appropriate instead? 

(I guess this also implies the test is not separating stderr and stdout properly, so probably a test error as well. Sorry!)

garlick:
Yeah I was having the same thoughts.  I'll look at making the test more robust and then back out the oom warning.

garlick:
I fixed several tests so that they no longer fail when job stderr contains spurious warnings and such, then changed the oom disablement message from warning to debug level.

garlick:
OK based on thumbs up from grondo, i'll go ahead and set MWP :-)

PR #7036, opened 2025-09-08
allow MemorySwapMax to be configured

garlick:
Problem: exceeding `MemoryMax` does not trigger the oom killer on systems configured with swap, and swapping may not be desirable for applications.

Allow `MemorySwapMax` to be configured.  Setting this to zero prevents applications from swapping.

### DIFF
--- a/doc/man5/flux-config-exec.rst
+++ b/doc/man5/flux-config-exec.rst
@@ -121,6 +121,9 @@ MemoryMin, MemoryLow
    Specify the memory usage protection of the job.  Values are formatted as
    described above.
 
+MemorySwapMax
+   Specify the absolute limit on swap used by the job.  Values are formatted as
+   described above.
 
 .. _testexec:
 
--- a/doc/test/spell.en.pws
+++ b/doc/test/spell.en.pws
@@ -716,6 +716,7 @@ hwlocality
 snprintf
 MemoryHigh
 MemoryMax
+MemorySwapMax
 MemoryLow
 MemoryMin
 myattr


t/t2410-sdexec-memlimit.t
```diff
@@ -73,25 +73,45 @@ if ! $sdexec $getcg memory.high; then
 	test_done
 fi
 
+#
```
grondo, line 77: unrelated to this PR, but just noticed I'm getting these errors when running this test on my hybrid system:
```
build-options:		+systemd+hwloc.api==2.5.0+zmq==4.3.4
Version=249.11-0ubuntu3.12
DelegateControllers=memory pids
cat: /sys/fs/cgroup//: Is a directory
cat: /: Is a directory
cat: /: Is a directory
cat: /user.slice/user-1001.slice/session-4039.scope: No such file or directory
cat: /: Is a directory
cat: /user.slice: No such file or directory
cat: /: Is a directory
cat: /: Is a directory
cat: /: Is a directory
cat: /user.slice/user-1001.slice/session-4039.scope: No such file or directory
cat: /: Is a directory
cat: /user.slice/user-1001.slice/session-4039.scope: No such file or directory
cat: /user.slice/user-1001.slice/session-4039.scope/cgroup.controllers: No such file or directory
1..0 # SKIP cgroups memory controller is not enabled
```

garlick:
There's a commit proposed in #7035 that may fix that ([fa0e14d](https://github.com/flux-framework/flux-core/pull/7035/commits/fa0e14d2d6e83060b6c45e88f859f497b5f8fae8))?

garlick:
Thanks! I'll go ahead and set MWP here and if that other commit isn't right we can iterate in that PR.

grondo:
Yes, sorry. Reviewed this one first.

PR #7038, opened 2025-09-08
flux-module: fix list output alignment

garlick:
Problem: the header is not column-aligned with the data in flux-module list output.

An earlier change to make room for long module names adjusted the header position without adjusting the data field width and fill.

Adjust the data format to match the header.


grondo (approved):
Ugh. Sorry! :facepalm:

PR #7039, opened 2025-09-08
modprobe: move package provided files out of `/etc`

grondo:
This PR moves internal config and rc files for modprobe out of `/etc/flux/modprobe` and into `/usr/share/flux/modprobe` (datadir) and `/usr/libexec/flux/modprobe` (libexecdir). The two paths are used because `/usr/share` is more appropriate for TOML config files, while `/usr/libexec` is standard for code like the Python `rcX.py` files. (It is an open question whether to just put the two types of configuration for modprobe under the same paths. This is why the PR is currently WIP)

The modprobe default search path now works like this:
 - the default `modprobe.toml` or `rcX.py` is read first from `/usr/share/flux/modprobe/modprobe.toml` or `/usr/lbexec/flux/modprobe/rcX.py`
 - If `FLUX_MODPROBE_PATH` is set, then it is a colon separated PATH from which `$dir/modprobe.d/*.toml` or `$dir/rcX.d/*.py` are read.
 - Otherwise, the default search path is `/usr/share/flux/modprobe:/etc/flux/modprobe` for TOML config and `/usr/libexec/flux/modprobe:/etc/flux/modprobe` for rcX python files.
 
 This preserves backwards compatibility (files from other projects can stay in `/etc/flux/modprobe` until it is convenient  to move them to the package directories) and also allows `FLUX_MODPROBE_PATH` to be set to an empty value to disable reading of package and system modprobe files (better isolation).

Arguably the split between `datadir` and `libexecdir` for these files is most correct. However, it could be confusing that package modprobe files are read from two different paths. It also makes it tricky to append to the default search path, and therefore support for a `FLUX_MODPROBE_PATH_APPEND` is added here.

It may be worth it to evaluate if the TOML and Python files should just both go in `/usr/libexec` or `/usr/share` or wherever. Thus this PR is WIP for now (didn't want to write documentation and tests before we settled on the correct approach)

### DIFF
--- a/doc/man1/flux-modprobe.rst
+++ b/doc/man1/flux-modprobe.rst
@@ -9,9 +9,9 @@ SYNOPSIS
 | **flux** **modprobe** **load** [*--dry-run*] *MODULE* [*MODULE* ...]
 | **flux** **modprobe** **remove** [*--dry-run*] *MODULE* [*MODULE* ...]
 | **flux** **modprobe** **list-dependencies** [*-f*] *MODULE*
-| **flux** **modprobe** **run** [*-v*] [*--show-deps*] [*--dry-run*] [*--confdir=DIR*] *FILE*
-| **flux** **modprobe** **rc1** [*-v*] [*--timing*] [*--show-deps*] [*--dry-run*] [*--confdir=DIR*] *FILE*
-| **flux** **modprobe** **rc3** [*-v*] [*--show-deps*] [*--dry-run*] [*--confdir=DIR*] *FILE*
+| **flux** **modprobe** **run** [*-v*] [*--show-deps*] [*--dry-run*] *FILE*
+| **flux** **modprobe** **rc1** [*-v*] [*--timing*] [*--show-deps*] [*--dry-run*] *FILE*
+| **flux** **modprobe** **rc3** [*-v*] [*--show-deps*] [*--dry-run*] *FILE*
 
 
 DESCRIPTION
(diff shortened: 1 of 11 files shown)


garlick:
I think having those related files in different system directories is fine and appropriate.

grondo:
Ok, I added some docs and a couple simple tests. Removed WIP.

src/bindings/python/flux/modprobe.py
```diff
@@ -16,7 +16,7 @@
 import sys
```
garlick, line 16: Commit message:
s/buliting/built-in/
s/subpackage/framework package/

> It is backwards compatible with the old system of always dropping
   files into

Maybe "allowing drop-in files in..." was meant?

doc/man1/flux-modprobe.rst
```diff
@@ -195,9 +195,23 @@ MODULE CONFIG
...
+
+which is typically::
+
+  /usr/share/flux/modprobe:/etc/flux/modprobe
+
+Packages should install files under ``/usr/libexec/flux/modprobe`` while
+site updates and modifications go under ``/etc/flux/modprobe``.
+
+The default search path can be overridden using :envvar:`FLUX_MODPROBE_PATH`
+or extended using :envvar:`FLUX_MODPROBE_PATH_APPEND`
```
garlick, line 214: FLUX_MODPROBE_PATH_APPEND should be added to flux-environment(7) too!

garlick (approved):
LGTM! I just had some superficial comments.

I installed this on my test system and verified that things were in the expected places and still worked.

doc/man1/flux-modprobe.rst
grondo: Thanks for catching that!

grondo: The entry for `FLUX_MODPROBE_PATH` in flux-environment(7) is also wrong now :facepalm:

grondo:
Thanks! Fixed those things and will set MWP.

Issue #7030, opened 2025-09-05
`flux jobs -o` struggles with right-alignment in certain fields


wihobbs:
@greene30 reported the following, while trying to create an f-string that would right align everything in `flux jobs`:
> I think I figured out that right alignment thing with flux jobs. It's not checking the width of the column title, only the values in the column. All the columns that seemed to be working had values longer than the title.

Here's an example:

```
$ flux jobs -A -o "{id:>19} +:{username:>8} | +:{queue:>} | +:{duration:>} | +:{ntasks:>} +:{runtime:>h} +:{t_remaining:>h}"
              JOBID         USER |    QUEUE | DURATION | NTASKS            RUNTIME        T_REMAINING
 555248194025552896     username |      pci |  3600.0 |   1                  -                  -
 555248196592466944     username |      pci |  3600.0 |   1                  -                  -
```

It gets off a bit at the end (once duration and ntasks are printed, which seems to consistently be below 8 characters). Flux bug or f-string problem?

If we find a working f-string/workaround, we might want to add "all-right-aligned" to the standard set of supported `--format` options.

chu11:
> It gets off a bit at the end (once duration and ntasks are printed, which seems to consistently be below 8 characters). Flux bug or f-string problem?

So my initial thought is this is not a bug.  Here's the default `flux-jobs` format:

```
            "format": (
                "{id.f58:>12} ?:{queue:<8.8} {username:<8.8} {name:<10.10+} "
                "{status_abbrev:>2.2} {ntasks:>6} {nnodes:>6h} "
                "{contextual_time!F:>8h} {contextual_info}"
            ),
```

you'll note how we take into account column width within it by setting the specific width.

So I guess one can think of the format as "per line", each line not having knowledge of any other lines.

I suspect the user thought the format was knowledgeable of all the output and would be adjusted as a result?  I suppose that is feasibly doable (i.e. scan all of the potential output, calculate max width, update the f-string format from the user), but I'm not convinced that's the right thing to do.  The python f-string format is what it is and I feel we need to honor it versus mucking with it, b/c we don't want to go down the path of making assumptions of what the user meant with the inputted format and modifying it based on the those assumptions.

grondo:
> I suspect the user thought the format was knowledgeable of all the output and would be adjusted as a result? 

[...]

PR #7042, opened 2025-09-09
Show full usernames by default in flux-jobs(1) and flux-pgrep(1) 

wihobbs:
Last week, on the WSC rounds, a frequent complaint of users was that the default output of `flux jobs` cut off full usernames, and it felt like something they had to constantly work around. With the IdM upgrade last year, now a lot of usernames are longer than 8 characters, and I assume this is something we'd see at other centers as well.

I don't think showing a `+` at the end of the username to indicate the length exceeded 8 characters would really address the usability, since that would still require our users to constantly override the default format. 

This seems like a very reasonable request. I think we should change the default format.

### DIFF
--- a/src/cmd/flux-jobs.py
+++ b/src/cmd/flux-jobs.py
@@ -40,7 +40,7 @@ class FluxJobsConfig(UtilConfig):
         "default": {
             "description": "Default flux-jobs format string",
             "format": (
-                "{id.f58:>12} ?:{queue:<8.8} {username:<8.8} {name:<10.10+} "
+                "{id.f58:>12} ?:{queue:<8.8} +:{username:<8.8} {name:<10.10+} "
                 "{status_abbrev:>2.2} {ntasks:>6} {nnodes:>6h} "
                 "{contextual_time!F:>8h} {contextual_info}"
             ),
@@ -48,7 +48,7 @@ class FluxJobsConfig(UtilConfig):
         "cute": {
             "description": "Cute flux-jobs format string (default with emojis)",
             "format": (
-                "{id.f58:>12} ?:{queue:<8.8} {username:<8.8} {name:<10.10+} "
+                "{id.f58:>12} ?:{queue:<8.8} +:{username:<8.8} {name:<10.10+} "
                 "{status_emoji:>5.5} {ntasks:>6} {nnodes:>6h} "
                 "{contextual_time!F:>8h} {contextual_info}"
             ),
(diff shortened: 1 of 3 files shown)


chu11:
an aside, should we fix #7030 first?  lemme look at that issue more today

grondo:
> an aside, should we fix https://github.com/flux-framework/flux-core/issues/7030 first?

I don't think the two issues are related? This one is just a default format tweak.

src/cmd/flux-jobs.py
```diff
@@ -40,23 +40,23 @@ class FluxJobsConfig(UtilConfig):
         "default": {
             "description": "Default flux-jobs format string",
             "format": (
-                "{id.f58:>12} ?:{queue:<8.8} {username:<8.8} {name:<10.10+} "
```
grondo, line 43: Commit message suggestion: Since this updates more than the default format, the commit message title and description should reflect that. I like what you have in the PR title "flux-jobs: show full usernames by default"

src/cmd/flux-pgrep.py
```diff
@@ -35,15 +35,15 @@ class FluxPgrepConfig(UtilConfig):
         "full": {
             "description": "full flux-pgrep format string",
             "format": (
-                "{id.f58:>12} ?:{queue:<8.8} {username:<8.8} {name:<10.10+} "
```
grondo, line 38: Similar suggestion for commit title reword here.

chu11:
> I don't think the two issues are related? This one is just a default format tweak.

They aren't related per se, but it could result in a lot more improperly aligned outputs since it would become the default?.

i.e. there are users with usernames < 4 chars ("USER" header)

wihobbs:
I think we're probably fine, because the floor is and would continue to be 8 chars?
```
$ /usr/bin/python3 ./src/cmd/flux-jobs.py -a -u unm
       JOBID QUEUE    USER     NAME       ST NTASKS NNODES     TIME INFO
f4nUHnZdkmMH pbatch   unm      test_test+ TO      1      1      24h corona196
f4nUHnekGHiF pbatch   unm      test_test+ CD      1      1   21.77h corona213
```

chu11:
> I think we're probably fine, because the floor is and would continue to be 8 chars?

Ahhh, we put a width field of 8 in our default formats, so we'd be fine.  Got it!

wihobbs:
With that, setting MWP.

grondo:
> because the floor is and would continue to be 8 chars?

Yes, and the heading there is USER, so only 4. The "fix" for the other issue should just be to default the minimum width for expandable columns to the header length...

chu11:
while working on #7030, I realized something, should:

```
+:{username:<8.8}
```

have been 

```
+:{username:<8}
```

b/c the `.8` will truncate the username.

Edit: testing shows that both the width and precision are both updated.  Although I'm not sure how at the moment.  But perhaps we should consider removing the ".8" because it may give wrong impression?

Edit2: Ok, internally updating the spec "width" also updates the precision.  So that explains that.  I'll leave above comment for consideration.

grondo:
Yeah, the precision doesn't do anything for strings that have the expanding width prefix `+:` since they will never be truncated. Probably a good idea to drop the precision in that case as good practice. Also a note in the man page wouldn't hurt.

PR #7045, opened 2025-09-09
flux-fsck: use content.load on older versions

chu11:
Problem: In older versions of flux, content.validate isn't available, thus flux-fsck cannot work against older versions of flux.

Divert to content.load if content.validate is not available.

---

just like #7044, hand tested on docker w/ v0.76.0. I don't think we have infrastructure to do a "current master" vs "random older versions", that's perhaps something to consider for another day?

### DIFF
--- a/src/cmd/builtin/fsck.c
+++ b/src/cmd/builtin/fsck.c
@@ -33,6 +33,7 @@
 
 struct fsck_ctx {
     flux_t *h;
+    bool validate_available;
     json_t *root;
     int sequence;
     int repair_count;
@@ -136,6 +137,8 @@ static void valref_validate_continuation (flux_future_t *f, void *arg)
 
 static void valref_validate (struct fsck_valref_data *fvd)
 {
+    const char *topic = fvd->ctx->validate_available ?
+        "content-backing.validate" : "content-backing.load";
     uint32_t hash[BLOBREF_MAX_DIGEST_SIZE];
     ssize_t hash_size;
     const char *blobref;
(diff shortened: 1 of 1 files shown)


chu11:
> Couldn't this be simplified if we just probe to see whether validate is available at the beginning of the program? A topic string for validate could be configured at that time, and other than the topic string, it seems like RPC handling can be exactly the same for both cases?

That's a good idea, lemme try that out.

chu11:
re-pushed, going with the alternate implementation recommended above

src/cmd/builtin/fsck.c
```diff
@@ -136,24 +137,33 @@ static void valref_validate_continuation (flux_future_t *f, void *arg)
...
+                                0,
+                                0)))
+            log_err_exit ("failed to validate valref blob");
+    }
+    else {
+        if (!(f = content_load_byblobref (fvd->ctx->h,
+                                          blobref,
+                                          CONTENT_FLAG_CACHE_BYPASS)))
+            log_err_exit ("failed to load valref blob");
+    }
```
garlick: Since these RPCs have the same request payload, it seems like this really doesn't need to be two separate calls.  Just do
```c
const char *topic = fwd->validate_available ? "content-backing.validate" : "content-backing.load";
```
and then do the raw rpc and check for failure, ignoring the return payload?

src/cmd/builtin/fsck.c
```diff
@@ -636,6 +646,43 @@ static void sync_checkpoint (struct fsck_ctx *ctx)
...
+             */
+            if (errno == ENOENT)
+                ctx->validate_available = true;
+            else
+                log_err_exit ("cannot determine if validate available");
+        }
+    }
+    else
+        ctx->validate_available = true;
+}
```
garlick: Couldn't this just be
```c
ctx->validate_available = true;
// doesn't matter if the request is correct, we are looking for ENOSYS
if ((f = flux_rpc (ctx->h, "content-backing.validate", NULL, 0, 0))
    && flux_rpc_get (f, NULL) < 0
    && errno == ENOSYS)
    ctx->validate_available = false;
flux_future_destroy (f);
```

src/cmd/builtin/fsck.c
chu11: yeah, I was trying to keep use of the `content_load_byblobref()` since we do that everywhere.  But it certainly isn't necessary.

chu11:
@garlick thanks.  fixed up and tested against older docker image to double check it still worked.  will set MWP.

PR #7053, opened 2025-09-12
python: consider output header w/ expandable width

### LINKED ISSUE #7030: `flux jobs -o` struggles with right-alignment in certain fields

chu11:
Problem: The expandable field prefix ("+:") does not consider the width of the output header name when calculated output width.
If all output entries are less than the size of the output header name, then output will not look as expected.

Solution: When calculating maximum output width of a field, take into account the header entry name.

Fixes #7030

----

Edit: Oops, some python tests I forgot to update, working on em :-)

### DIFF
--- a/t/python/t0024-util.py
+++ b/t/python/t0024-util.py
@@ -218,6 +218,11 @@ class TestOutputFormat(unittest.TestCase):
         ).filter(items)
         self.assertEqual(fmt, "{i:>8} {f:5.2f}")
 
+        fmt = OutputFormat(
+            "?+:{i:>7} ?:{s:>6} ?+:{f:.2f}", headings=self.headings
+        ).filter(items, no_header=True)
+        self.assertEqual(fmt, "{i:>8} {f:3.2f}")
+
     def test_sort(self):
         a = Item("a", 0, 2.2)
         d = Item("d", 33, 1.1)
@@ -269,6 +274,11 @@ class TestOutputFormat(unittest.TestCase):
             "+:{s:5.5} +:{i:4d} +:{f:.2f}", headings=self.headings
         ).filter(items)
         self.assertEqual(fmt, "{s:16.16} {i:7d} {f:5.2f}")
+
+        fmt = OutputFormat(
+            "+:{s:5.5} +:{i:4d} +:{f:.2f}", headings=self.headings
+        ).filter(items, no_header=True)
+        self.assertEqual(fmt, "{s:16.16} {i:4d} {f:3.2f}")
 
     def test_copy(self):
         original = "+:{s:5.5} {i:4d} {f:.2f}"


src/bindings/python/flux/util.py
```diff
@@ -1000,12 +1000,20 @@ def sentinel_keys():
...
                 #  Save the modified format, index, type, maximum width,
                 #  observed width, and broken-down spec in lst:
+                initialmaxwidth = spec.width or 0
+                if (
+                    sentinels[end] in ("maxwidth", "both")
+                    and self.headings
+                    and self.headings[field]
+                    and len(self.headings[field]) > initialmaxwidth
+                ):
+                    initialmaxwidth = len(self.headings[field])
```
grondo: Unfortunately, this applies the heading length to maxwidth even if `-n, --no-header` is used, which isn't what we want. 

Let me ponder the best way to fix this. It may be to just pass `args.no_header` into the initializer then zero out the headings (if that works, it may not be so simple for JobInfoFormatter)

chu11: Yeah, I did realize that.  We seem to always pass headers into OutputFormat.  

I did consider it being a possibly separate issue.  Notice this commit I just added 0e4a22e470f6ca71e966d1bc44e7e46e4effa72e.  That's just to get a test to pass, checking if headers exist in other parts of OutputFormat probably also warranted.

grondo: Yeah, and the design allows headers to be class variables when you subclass so just not passing in headings is not a great solution.

The simplest thing would be just add a `no_headers=False` optional argument to the class initializer and use this in the test above. Unfortunately we'll have to update all callers, but I don't really see a cleaner way around it for now.

Finally, to make this code cleaner, the class should start with an empty `self.headings`, e.g. `self.headings = {}`. Then you can drop the test for `self.headings`. Then avoid accessing `self.headings[field]` multiple times by doing something like:

```python
if sentinels[end] in ("maxwidth", "both"):
    initialmaxwidth = max(spec.width or 0, len(self.headings.get(field, "")))
```

chu11: > The simplest thing would be just add a no_headers=False optional argument to the class initializer and use this in the test above. Unfortunately we'll have to update all callers, but I don't really see a cleaner way around it for now.

Hmmm.  Technically I think we only have to do this for JobInfoFormat?  B/c w/ OutputFormat we can just pass in or not pass in a headings.  For JobInfoFormat we are always setting one.  Or are you thinking of doing both for just consistency?

grondo: Since the code to set the maxwidth is in the base `OutputFormat` class, it should take `no_header` in its initializer I'd think.

The initialization of `self.headings` in `OutputFormat` to and empty dictionary would just seem to make the rest of the code cleaner because you would not have to check for `self.headings = None` anywhere. 

Let me check if that would mess up any other assumptions though.

grondo: Oh, we can't set headings to an empty dict because it is currently used to detect invalid fields :disappointed: :
```python
        headings (dict): A mapping of field name to header string. This also
            provides detection of invalid fields, so all valid fields should
            have an entry in headings. The headings dict can be set as an
            argument to :py:func:`OutputFormat.__init__`, or a subclass of
            the :py:class`flux.util.OutputFormat` class may set default
            headings for a given tool.
```

This does seem to imply `self.headings` should always be a valid dictionary though. Did you find that was not the case to add the check for `self.headings` above?

chu11: Ahh, I was only going off the fact that the `headings` input to the `__init__` defaults to `None`, so I just assumed "no headings" was ok or should be ok.

I guess we just need a `no_headers` flag everywhere then.  And probably raise an exception if no headers are specified.

As I thought about it this afternoon, I think a "no_headers" PR will probably have to come before this one.

grondo: I think it could be in one PR since it is only required to fix this issue properly.

chu11: Hmmmm.  I just noticed that the primary user of the `filter()` function in `OutputFormat` is the `print_items()` function, which already has a `no_header` flag.  So we could get away with adding `no_header` to `filter()` instead.  Any objections?  That would take care of the majority of cases, just a few oddball ones (for example `flux-jobs` calls `filter()` directly).

Edit: As I look at this more, I'm liking this option more

grondo: Yeah, that is what I was envisioning as the easiest "fix".

chu11: ahh, you mentioned passing `no_header` in the initializer at first, so that was what I was going with initially.

grondo: Oh yeah, I did, didn't I? Sorry, I probably shouldn't mention specifics until actually looking at the implementation again :facepalm:

chu11:
re-pushed with fixes per discussion above

grondo (approved):
This turned out excellent. Nice work!

PR #7062, opened 2025-09-16
testsuite: fix a couple tests in t0100-modprobe.t

grondo:
Problem: A couple of tests in t0100-modprobe.t use `libexecdir` as the expected path to `modprobe.d`, but this should be `datadir`. This works for the build tree since `libexecdir` == `datadir`, but will fail when testing installed Flux. Additionally, use of `$(pwd)` as a search path entry may not be failsafe, since there is a chance a different path could be printed by modprobe, e.g. in the case of a symlink.

Fix these issues by using `datadir` for expected paths to modprobe TOML files, and an explicit fake path in place of `$(pwd)` when matching exact paths in output.

Additionally, add more useful debug output to these tests.


chu11 (approved):
LGTM!

grondo:
FYI - just force pushed a change so that the debug output from the two changed tests is the same. Other than that the changes are the same so I'll set MWP (assuming CI passes)

Issue #7043, opened 2025-09-09
suggestion: cap output of `flux jobs` at terminal length


wihobbs:
Another suggestion that came up last week about `flux jobs` was that its output should be capped at terminal length. On elcap/tuolumne, large jobs frequently have hostlists that get wrapped into 2+ lines, which makes the output format look a bit weird. The users suggested that the contextual info section of `flux jobs` didn't need to be that long, that just seeing enough nodes (1+) to choose an ssh target (to which they could attach a debugger for their job) was enough. 

@grondo had concerns that this would break the testsuite, but suggested perhaps we could work around never truncating unless the output was to a terminal.

grondo:
Actually, I'd argue against truncating the hostlist by default. It just doesn't seem right and doesn't follow the behavior of other common utilities like `ps`, `ls` etc. Not saying my mind can't be changed, though. Extremely long hostlists by default on large systems does seem a bit unnecessary.

 We can support this via an option though, and perhaps support enabling the option by default in the `flux-job` config TOML. Sysadmins or users could place the option `$XDG_CONFIG_DIRS/flux/flux-jobs.toml` or `$XDG_CONFIG_HOME/flux/flux-jobs.toml` to get truncation as the default.

grondo:
Here's a possible approach (already started playing around with this idea):
 
- add a `max_width` parameter to `OutputFormat.print_items()` that will truncate lines at width
- add a new option `-w, --width` to `flux jobs` that sets a maximum line width. If no argument it will use the current terminal width
- add support in the flux-jobs [config file](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man1/flux-jobs.html#configuration) for a `default-args` or similar key that allows admins/users to set default command line args that are prepended to the actual args (so actual args override)

With this approach there will be an option to truncate lines at the current terminal width (`-w`), plus a way to make this the default on a per-system or per-user basis.

An alternative to `default-args` above could be to have a `defaults` table in the config supporting a subset of the args, starting with `args.width`.

kkier:
[...]

PR #7063, opened 2025-09-17
flux-jobs: add `-w, --width[=N]` option

grondo:
This PR adds a simple `-w, --width` option to `flux jobs` which can be used to truncate output to either terminal width (if `-w` is used without an argument or `--width=0`) or to truncate an any other specified width.

This partially solves #7043. A later PR may add a config file value to set a default for this parameter so truncation could be made the default on a per-user or per-system basis.

### DIFF
--- a/etc/completions/flux.pre
+++ b/etc/completions/flux.pre
@@ -1441,6 +1441,7 @@ _flux_jobs()
         --stats \
         --stats-only \
         --sort= \
+        -w --width= \
     "
 
     _flux_split_longopt && split=true


chu11:
bash completions? (i recently forgot on `flux job attach --tail` :-)

grondo:
Heh, good catch, thanks!

src/cmd/flux-jobs.py
```diff
@@ -320,6 +320,16 @@ def parse_args():
...
     )
+    parser.add_argument(
+        "-w",
+        "--width",
+        type=int,
+        default=-1,
+        const=0,
+        metavar="N",
+        nargs="?",
+        help="Truncate output at width N. If N=0 (default), use terminal size",
```
chu11: This makes it sound like `N=0` is the default, vs the "no arg default".  I didn't know how we normally word it,  `flux-jobs` only other optional arg option is:

`        help="Colorize output; WHEN can be 'never', 'always', or 'auto' (default)",`

maybe just remove the "(default)"?  Perhaps say "(default no truncation)" at the end?

t/t2800-jobs-cmd.t
```diff
@@ -1111,6 +1111,27 @@ test_expect_success 'flux-jobs --format={expiration!D:h},{t_remaining!H:h} works
...
+# terminal output adds line feed (^M) character to each line in addition to
+# newline. Similarly, do not use color to avoid extra non-printing characters:
+test_expect_success 'flux-jobs -w, --width no argument works' '
+	runpty.py -w 10x40 -o width-terminal.output \
+		flux jobs -aw --color=never &&
+	test_debug "cat width-terminal.output" &&
+	test_must_fail grep -E "^.{12}"  width-terminal.output
+'
+# Note: extra character in terminal test below needed to match line feed
+test_expect_success 'flux-jobs -w, --width (no arg) does nothinif no terminal' '
```
chu11: typo "nothinif"

t/t2800-jobs-cmd.t
```diff
@@ -1111,6 +1111,27 @@ test_expect_success 'flux-jobs --format={expiration!D:h},{t_remaining!H:h} works
 	test "${expiration}" = "-" &&
 	test "${t_remaining}" = "-"
 '
+# -w, --width tests
+test_expect_success 'flux-jobs -w, --width=N works' '
+	flux jobs -a -w 10 > width10.output &&
+	test_debug "cat width10.output" &&
+	test_must_fail grep -E "^.{11}" width10.output
```
chu11: grep for "+" sign for truncation? same below?

chu11 (approved):
overall looks good to me, just small things

src/cmd/flux-jobs.py
grondo: How about 

> Truncate output at width N. If N=0 (default if N omitted), use terminal size.

t/t2800-jobs-cmd.t
grondo: Sure, that can't hurt.

t/t2800-jobs-cmd.t
wihobbs: "nothinif" sounds like a southern-ism, actually a fan of keeping this one /s

wihobbs:
Thanks for doing this @grondo!

grondo:
Ok, I've addressed the great feedback from @chu11. Will set MWP now. Thanks!

PR #7064, opened 2025-09-17
libflux: do not require out parameters in flux_response_decode_raw()

### LINKED ISSUE #6824: libflux: flux_response_decode_raw should not require data parameter
I noticed that

```
int flux_response_decode_raw (const flux_msg_t *msg,
                              const char **topic,
                              const void **data, 
                              size_t *len)
{   
    const char *ts;
    const void *d = NULL;
    size_t l = 0;
    int rc = -1;
    
    if (!data || !len) {
        errno = EINVAL;
        goto done;
    }   
```

both `data` and `len` are required.  It's possible someone only needs the length, so the `data` being returned isn't necessary.

chu11:
Problem: In flux_response_decode_raw() the data and length out parameters are required.  This is not consistent with most functions of this style, as out parameters are typically optional.  For example, perhaps a caller is only interested in the length of the data returned and not the data itself.

Do not require the data or len out parameters to flux_response_decode_raw() to be required.

Fixes #6824

### DIFF
--- a/src/common/libflux/test/response.c
+++ b/src/common/libflux/test/response.c
@@ -68,8 +68,8 @@ int main (int argc, char *argv[])
     ok (flux_response_decode_raw (msg, &topic, &d, &l) == 0
         && topic != NULL && streq (topic, "foo.bar"),
         "flux_response_decode_raw returns encoded topic");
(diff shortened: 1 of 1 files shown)


garlick (approved):
Oh yeah, good catch!  LGTM.

PR #7065, opened 2025-09-17
libflux: fix flux_pollfd() file descriptor leak

garlick:
Problem: calling `flux_pollfd()` causes an epoll file descriptor to be created internally, but the file descriptor is not closed when the `flux_t` handle is destroyed.
    
Since each call to `flux_pollfd()` reuses the internal file descriptor, it is clearly "owned" by the handle.  Close it when the handle is destroyed.

Also:  update the man page to make it clear that the file descriptor belongs to the handle, and while in there, replace a convoluted example with a simple description.

### DIFF
--- a/src/common/libflux/handle.c
+++ b/src/common/libflux/handle.c
@@ -529,6 +529,7 @@ flux_t *flux_handle_create (void *impl,
     h->flags = flags;
     h->ops = ops;
     h->impl = impl;
+    h->pollfd = -1;
     if (!(h->tagpool = tagpool_create ()))
         goto error;
     if (!(flags & FLUX_O_NOREQUEUE)) {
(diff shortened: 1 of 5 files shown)


src/common/libflux/handle.c
```diff
@@ -640,6 +640,8 @@ void flux_handle_destroy (flux_t *h)
             flux_handle_destroy (h->parent); // decr usecount
         }
         else {
+            if (h->pollfd >= 0)
+                (void)close (h->pollfd);
```
grondo, line 644: Just double checking that this won't close fd=0 from any of the `goto error` conditions in `flux_handle_create()`?

grondo (approved):
LGTM! One question inline.

src/common/libflux/handle.c
garlick: Doh!  Excellent catch.

garlick:
FIxed that so I'll set MWP

PR #7066, opened 2025-09-17
libflux: ensure flux_pollfd() is edge triggered

garlick:
Problem: when `flux_open()` is used with the FLUX_O_NOREQUEUE flag, the connector implementation's pollfd method is returned directly by `flux_pollfd()`, but not all connectors supply an edge-triggered fd, as required by `flux_pollfd()`.

For example, the local connector's pollfd method returns the level-triggered communications socket.

FLUX_O_NOREQUEUE was added to reduce the number of file descriptors used by interthread sockets from 3 (deque, requeue, and epoll wrapper) to 1 (deque).  Since back-to-back interthread pairs are used for each broker module, this allowed more co-located test brokers to be run for a given file descriptor limit.

Compromise: let FLUX_O_NOREQUEUE suppress creation of the requeue deque, saving 1 file descriptor, but unconditionally create the epoll wrapper, for the greater good of consistent `flux_pollfd()` semantics.

Interthread connectors opened with FLUX_O_NOREQUEUE thus now use 2 file descriptors.

### DIFF
--- a/src/common/libflux/handle.c
+++ b/src/common/libflux/handle.c
@@ -1104,26 +1104,20 @@ int flux_pollfd (flux_t *h)
 {
     h = lookup_clone_ancestor (h);
 
-    if ((h->flags & FLUX_O_NOREQUEUE)) {
-        if (!h->ops->pollfd) {
-            errno = ENOTSUP;
-            return -1;
-        }
-        return h->ops->pollfd (h->impl);
-    }
-
     if (h->pollfd < 0) {
         struct epoll_event ev = {
             .events = EPOLLET | EPOLLIN | EPOLLOUT | EPOLLERR | EPOLLHUP,
         };
         if ((h->pollfd = epoll_create1 (EPOLL_CLOEXEC)) < 0)
             goto error;
-        /* add queue pollfd */
-        ev.data.fd = msg_deque_pollfd (h->queue);
-        if (ev.data.fd < 0)
-            goto error;
-        if (epoll_ctl (h->pollfd, EPOLL_CTL_ADD, ev.data.fd, &ev) < 0)
-            goto error;
+        /* add re-queue pollfd (if defined) */
+        if (!(h->flags & FLUX_O_NOREQUEUE)) {
+            ev.data.fd = msg_deque_pollfd (h->queue);
+            if (ev.data.fd < 0)
+                goto error;
+            if (epoll_ctl (h->pollfd, EPOLL_CTL_ADD, ev.data.fd, &ev) < 0)
+                goto error;
+        }
         /* add connector pollfd (if defined) */
         if (h->ops->pollfd) {
             ev.data.fd = h->ops->pollfd (h->impl);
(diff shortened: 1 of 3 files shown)


src/common/libflux/test/interthread.c
```diff
@@ -353,68 +353,55 @@ void test_poll (void)
...
         BAIL_OUT ("%s: flux_open: %s", uri, strerror (errno));
     diag ("poll: opened h1 and h2");
 
+    ok (flux_pollevents (h2) == FLUX_POLLOUT,
+        "flux_pollevents h2 initially returns POLLOUT");
+    ok ((flux_pollfd (h2)) >= 0,
+        "flux_pollfd works");
+    recv_count_is (h2, 0, "RECV_QUEUE_COUNT h2 = 0");
+
+    // just make sure poll() thinks we've consumed the initial edge, if any
```
grondo: This could use an expanded comment IMO. Probable that I'm dense, but even with the commit message as context, I'm confused why this loop is necessary. An explanation of not only what is happening here (as in the existing comment) but why would be really helpful.

src/common/libflux/handle.c
```diff
@@ -1104,26 +1104,18 @@ int flux_pollfd (flux_t *h)
 {
     h = lookup_clone_ancestor (h);
 
-    if ((h->flags & FLUX_O_NOREQUEUE)) {
```
grondo, line 1107: commit message typo: "implementation->polfd"

garlick:
Fixed a bug (now obvious) that for some reason caused test failures across the board in CI but not on my test system.

Also reworded the issue description and title to be a little more direct.

src/common/libflux/handle.c
grondo: Also extra space after "Compromise:"

grondo (approved):
This LGTM! Just minor comments inline.

src/common/libflux/test/interthread.c
garlick: Thank you for asking about that!  I _thought_ the answer was just that "spurious wakeups" are a thing with `poll()` on an epoll file descriptor, and they certainly need to be handled by defensive code, but while trying to explain that I could hear myself making bold claims that were likely nonsense.  Digging deeper, I think I've found a couple more bugs in `flux_pollevents()`.

src/common/libflux/handle.c
garlick: I`m not spotting the commit typo now, but I think I fixed that one before my "fixed a bug" comment above.

src/common/libflux/test/interthread.c
garlick: Will try to fix the issues I found in another PR.   One explains why this was needed though.  For that one, I opened  an issue and added a "because issue#XXX" comment in the test.

garlick:
Did you want to have another look @grondo or is it OK to set MWP?

grondo:
Fine w/ me to set MWP!

PR #7068, opened 2025-09-18
libflux: fix flux_pollevents() POLLOUT logic

garlick:
Problem: `flux_pollevents()` always raises FLUX_POLLOUT,  even when the underlying connector is not ready for writing.

The connector pollevents bits are ORed with those from the re-queue deque, but the re-queue deque should only be contributing POLLIN and POLLERR, not POLLOUT, as it cannot sink messages from `flux_send()`.

In order to write tests for this, I added an optional message count limit to the `msg_deque` class used for various conector implementations, then exposed it in the `loop://` connector and added some tests.   This is a much bigger change than the fix unfortunately, but it is pretty straightforward and might come in handy down the road.


garlick:
fixed a typo that irritated the typo checker

garlick:
I tacked on one more minor commit that just makes sure if you initially call `flux_pollevents()` before `flux_pollfd()`, the effect is the same as the reverse order, since that might throw somebody someday.

src/common/libflux/msg_deque.c
```diff
@@ -185,20 +189,17 @@ bool msg_deque_empty (struct msg_deque *q)
     if (!q)
         return true;
     msg_deque_lock (q);
-    bool res = list_empty (&q->messages);
+    size_t count = q->count;
     msg_deque_unlock (q);
-    return res;
+    return count > 0 ? false : true;
```
grondo: Seems like count can't be less than zero here, so would
```c
return count == 0;
```
be a bit clearer here?

doc/man3/flux_pollevents.rst
```diff
@@ -11,21 +11,30 @@ SYNOPSIS
 
   #include <flux/core.h>
 
+  int flux_pollfd (flux_t *h);
```
grondo, line 14: Commit message: should "Rework the command description" be "Rework the function description"?

garlick:
Fixed those two things and I'll set MWP.  Thanks!

Issue #6630, opened 2025-02-13
idea: require backing module for content all the time


chu11:
There is a lot of special case code for dealing with the "none" backing module.  It may make things simpler if we simply require a backing module all of the time.

Note that we still want to support "no real backing module" for performance (especially for sub-instances), so hypothetically there could be a special "memory only" backing module.  or perhaps sqlite run in memory only mode.

garlick:
Yeah, I think it suffices to say that we have options if we don't want to make content persistent, and also that it's not a situation that's come up IRL recently, so fine to nix it in the name of simplifying code.

It looks like the backing modules all register the `content-backing` service already so the main thing is eliminating the `content.register-backing` RPCs and loading the backing module before the cache module.  Plus removing the "none" special cases from rc1, routing the checkpoint stuff through the cache, eliminating code to let the memory cache grow when backing is unavailable, etc..

chu11:
> Yeah, I think it suffices to say that we have options if we don't want to make content persistent, and also that it's not a situation that's come up IRL recently, so fine to nix it in the name of simplifying code.

The performance issue I'm medium concerned with is we will generate unnecessary RPCs.  i.e. the `content` wants to flush to the backing store that does nothing, so it's just a wasted RPC.  For sub-instances that don't need a backing store, we are introducing some extra traffic and being busy for no gain.

I'm trying to think if there's some "trick" we could do here, but a super clean one isn't coming to mind.  One possibly not great idea is to just special case the word "none" as a backing.  If "none" is the backing module, `content.flush` doesn't work and maybe some other RPCs won't.  Or perhaps the backing modules will advertise "we don't do flushes" and the `content` module can adjust accordingly.

garlick:
Performance benefits were not compelling as I recall.  I'd say just eliminate that option and take the win.

grondo:
should this have been closed by #6764?

chu11:
> should this have been closed by #6764?

I don't think that completely finishes this.  That PR removed the special word "none" in the rc modules, but behind the scenes we should still remove code.  Per comment above, if we simply require a backing module all of the time, there's no need for "register-backing" RPCs.  There might be some lingering code still handling "is a backing module loaded or not"?  I can't remember if all of that is handled yet or not (some was handled in #6255)

PR #7069, opened 2025-09-18
etc/modprobe: remove kvs requires on restore task

chu11:
Problem: The rc1 "restore" task lists loading the "kvs" as a "requires" dependency.  This conflicts with A) "flux restore --checkpoint" cannot be run if the kvs is loaded and B) "kvs" is listed in this task as a "before" dependency

Solution: Remove "kvs" from the list of "requires" for the "restore" task.

### DIFF
--- a/etc/modprobe/rc1.py
+++ b/etc/modprobe/rc1.py
@@ -92,7 +92,7 @@ def check_restore(context):
     ranks="0",
     before=["kvs"],
     after=["content-backing", "check-restore"],
-    requires=["check-restore", "content-backing", "kvs"],
+    requires=["check-restore", "content-backing"],
     needs=["content-backing", "check-restore"],
     needs_attrs=["content.restore"],
 )


grondo:
>  B) "kvs" is listed in this task as a "before" dependency

`requires` does not necessarily conflict with `before` (a task or module could need to be loaded before another task, but still require it for full functionality). However, in this case it does seem correct that the content restore task does not require the kvs for the other reason you listed. :+1:

grondo (approved):
Nice catch, thanks.

chu11:
> a task or module could need to be loaded before another task, but still require it for full functionality). 

perhaps there is a subtlety I'm missing.  When reading the documentation, it sounds like "before" (in rc1) means the module should not be loaded / active when the task is run.  "requires" means the module should be loaded / active when the task is run.  So wouldn't those conflict with each other if a module were listed in both?

Edit: As I read more comments in the code, i am getting the feeling "requires" indicates a "dependency" on something, but doesn't indicate a "before" or "after".  So like it could be done in parallel.  And could indicate to other tasks, that sets of modules / tasks need to be loaded.

chu11:
re-pushed, slightly tweaking commit message given comments above

grondo:
Yeah, it is subtle. "requires" defines a dependency relationship, while "before" and "after" are strictly for ordering or precedence of active tasks. If task A requires task B, and task A is active, then task B is forced to be active as well. However, at this point A can run in parallel with B. If A needs to be loaded before or after B, then "before" or "after" should be used to determine the precedence of tasks.

For example `job-info` requires the `kvs` and `kvs-watch` modules, but doesn't need them to be active before it starts, so we have:
```toml
[[modules]]
name = "job-info"
requires = ["kvs-watch", "kvs"]
```

The `kvs` module should be loaded after `content-backing` and `content` modules, but requires `content` and `heartbeat`. So it only comes after `content-backing` if that module is active:

```toml
[[modules]]
name = "kvs"
after = ["content-backing", "content"]
requires = ["content", "heartbeat"]
```

Though now that I think about it, do we now require a `content-backing` module for the kvs? (I faintly recall something changed here in between when the modules.toml file was originally developed to replicate `rc1` and now). So this may need a small correction as well (However, probably not much difference in practice if a content-backing module is always required)

I'm sure the docs could use some improvement in this area.

chu11:
> Though now that I think about it, do we now require a content-backing module for the kvs?

~~Hmmm, not directly.  KVS requires the content module.  The content module now requires a backing module (it didn't before).  So the dependency is content-backing -> content -> KVS~~ Edit: Oops, no.  We required the backing store only for checkpointing, there's still discussion about requiring a backing store all the time #6630.

I'll take a skim through those files and see if there is something that can be improved there.

PR #7070, opened 2025-09-19
doc: add extra clarification to modprobe docs

chu11:
Just a few things I think would help in the modprobe docs.  Wording is highly debatable :-)


chu11:
@mergifyio requeue

PR #7071, opened 2025-09-19
libflux: reduce spurious pollfd wake-ups

garlick:
Problem: there are extra `flux_pollfd()` wakeups for every real event.

This is explained in more detail in commit messages, but to quickly summarize: the  `msg_deque` class that underlies the `interthread://` and `loop://` connector implementations, and is used to implement `flux_requeue()`, has a quirk where it asserts POLLOUT when its `pollevents` method is called with POLLIN pending.  Unfortunately, the handle pollfd implementation wraps those in an epoll fd that watches POLLIN | POLLOUT, so it gets an extra wakeup per msg_deque on every call to `flux_pollevents()` when it is responding to POLLIN.  The fix is to not watch POLLOUT on the re-queue deque or the connector when we know we don't need it, which is the case with the msg_deques which were designed to use POLLIN signaling.

This is built on top of
-  #7068

### DIFF
--- a/src/common/libflux/handle.c
+++ b/src/common/libflux/handle.c
@@ -1140,16 +1140,16 @@ int flux_pollevents (flux_t *h)
 {
     h = lookup_clone_ancestor (h);
     int e, events = 0;
+    struct epoll_event ev;
 
     /* create pollfd if needed */
     if (flux_pollfd (h) < 0)
         return -1;
 
-    /* wait for handle event */
-    if (h->pollfd >= 0) {
-        struct epoll_event ev;
-        (void)epoll_wait (h->pollfd, &ev, 1, 0);
-    }
+    /* consume all epoll events before sampling pollevents */
+    while (epoll_wait (h->pollfd, &ev, 1, 0) == 1)
+        ;
+
     /* get connector events (if applicable) */
     if (h->ops->pollevents) {
         if ((events = h->ops->pollevents (h->impl)) < 0)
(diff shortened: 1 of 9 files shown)


src/common/libflux/handle.c
```diff
@@ -1140,16 +1140,16 @@ int flux_pollevents (flux_t *h)
...
 
-    /* wait for handle event */
-    if (h->pollfd >= 0) {
-        struct epoll_event ev;
-        (void)epoll_wait (h->pollfd, &ev, 1, 0);
-    }
+    /* consume all epoll events before sampling pollevents */
+    while (epoll_wait (h->pollfd, &ev, 1, 0) == 1)
+        ;
+
```
grondo: Is there any possibility this loop could run forever if there is a steady stream of messages or other activity on this handle from the other side? I do not know this section of the code that well and did not dig into it, so apologies if this question is naive.

grondo (approved):
LGTM! One question inline

src/common/libflux/handle.c
garlick: I don't think so.  The epollfd is supposed to be edge triggered so I think the the maximum number of events queued is limited by the number of possible edges, which are
- re-queue deque can become readable
- connector can become readable
- connector can become writable

garlick:
Rebased on master after #7068 was merged.

I'll set MWP - thanks!

PR #7072, opened 2025-09-19
WIP libflux: add fd:// connector

garlick:
Problem: when launching broker modules or other "agents" as subprocesses, it would be handy to be able to pass messages over a pre-connected, inherited file descriptors similar to what we do for PMI.

Add a `fd://` connector that could be used for such things.

This is peeled off of some "modules as processes work" so I'll mark it WIP for now until the rest of the context is posted.

This was built on top of
- #7071
- #7068

PR #7073, opened 2025-09-20
WIP libsubprocess: add mesage channels

garlick:
Problem: there is no mechanism in libsubprocess for creating a Flux message channel to a subprocess.

This adds the following interface to libsubprocess
```c
/*  Request a message channel for communication between process and caller.
 *
 *  The 'name' argument is the name of an environment variable that will be
 *   set to a fd:// URI on the process end.  The file descriptor in the path
 *   component of this URI represents one end of a socketpair opened in the
 *   caller and passed to the process.
 *
 *  The 'uri' argument is normally an interthread:// connector URI that will
 *   represent the channel on the caller end and act as message buffer.
 *   The rules of interthread:// are that a given URI may be flux_open()ed up
 *   to twice within the same process address space, in any order.  The
 *   subprocess object will be one opener for the specified URI.  The caller
 *   may be the other opener, or the URI may be safely passed to another
 *   thread and opened there.  The interthread URI should be chosen so it
 *   doesn't conflict with other URIs in the same process address space.
 *
 *  The two URIs will be joined back-to-back, so messages sent on one end will
 *   appear on the other end.
 */
int flux_cmd_add_message_channel (flux_cmd_t *cmd,
                                  const char *name,
                                  const char *uri);
```
This is built on top of
- #7072 
- #7071
- #7068


src/common/libsubprocess/msgchan.c
```diff
@@ -0,0 +1,285 @@
...
+                                                    wh)))
+        goto error;
+    flux_watcher_start (wh->read_w);
+    return 0;
+error:
+    errprintf (error, "error creating message watchers: %s", strerror (errno));
+    return -1;
+}
+
+void socketpair_init (int *fd)
```
grondo: should `socketpair_init()` and `socketpair_close()` be marked `static`?

src/common/libsubprocess/msgchan.c
```diff
@@ -0,0 +1,285 @@
...
+    }
+    return 0;
+}
+
+struct msgchan *msgchan_create (flux_reactor_t *reactor,
+                                const char *relay_uri,
+                                flux_error_t *error)
+{
+    struct msgchan *mch;
+
```
grondo, line 252: Optional: check for valid arguments

.typos.toml
```diff
@@ -38,3 +38,4 @@ clen = "clen"
 WRONLY = "WRONLY"
 TOU = "TOU"
 setp = "setp"
+mch= "mch"
```
grondo: missing space for consistency.

garlick:
Fixed, thanks.  Setting MWP.

PR #7075, opened 2025-09-22
add flux module load --exec option

garlick:
Problem: although Flux broker extension modules are only allowed to interact with each other using messages, this isolation is not  enforced and a badly behaved module can crash the broker.

Add support for `flux module [re]load --exec`.   The broker runs `flux module-exec MODULE` as a sub-process instead of a thread, communicating with it over a dedicated sub-process message channel.

This is the final PR of the stack and is built on top of
- #7073
- #7072 
- #7071
- #7068

### DIFF
--- a/doc/man1/flux-module.rst
+++ b/doc/man1/flux-module.rst
@@ -6,8 +6,8 @@ flux-module(1)
 SYNOPSIS
 ========
 
-| **flux** **module** **load** [*--name*] *module* [*args...*]
-| **flux** **module** **reload** [*--name*] [*--force*] *module* [*args...*]
+| **flux** **module** **load** [*--name*] [*--exec*] *module* [*args...*]
+| **flux** **module** **reload** [*--name*] [*--exec*] [*--force*] *module* [*args...*]
 | **flux** **module** **remove** [*--force*] *name*
 | **flux** **module** **list** [*-l*]
 | **flux** **module** **stats** [*-R*] [*--clear*] *name*
@@ -44,6 +44,10 @@ entered the running state (see LIST OUTPUT below).
   Override the default module name.  A single shared object file may be
   loaded multiple times under different names.
 
+.. option:: --exec
+
+  Load the module in a separate process, e.g. for isolation or debugging.
+
 reload
 ------
 
@@ -59,6 +63,10 @@ Reload module *name*. This is equivalent to running
 .. option:: -n, --name=NAME
 
   Override the default module name.
+
+.. option:: --exec
+
+  Load the module in a separate process, e.g. for isolation or debugging.
 
 remove
 ------
--- a/etc/completions/flux.pre
+++ b/etc/completions/flux.pre
@@ -1302,12 +1302,14 @@ _flux_module()
 
     local load_OPTS="\
         --name= \
+        --exec \
     "
     local remove_OPTS="\
         -f --force \
     "
     local reload_OPTS="\
         ${remove_OPTS} \
+        ${load_OPTS} \
     "
     local stats_OPTS="\
         -p --parse= \
--- a/src/broker/modhash.c
+++ b/src/broker/modhash.c
@@ -344,7 +344,15 @@ static int modhash_load_finalize (struct modhash *mh,
     }
     modhash_add (mh, p);
 
-    flux_log (mh->ctx->h, LOG_DEBUG, "insmod %s", module_get_name (p));
+    if (module_is_exec (p)) {
+        flux_log (mh->ctx->h,
+                  LOG_DEBUG,
+                  "insmod %s exec pid=%d",
+                  module_get_name (p),
+                  (int)module_get_pid (p));
+    }
+    else
+        flux_log (mh->ctx->h, LOG_DEBUG, "insmod %s", module_get_name (p));
     return 0;
 }
 
--- a/src/broker/modhash.c
+++ b/src/broker/modhash.c
@@ -61,6 +61,11 @@ static module_t *modhash_load_builtin (modhash_t *mh,
                                       const char *name,
                                       json_t *args,
                                       flux_error_t *error);
+static module_t *modhash_load_exec (modhash_t *mh,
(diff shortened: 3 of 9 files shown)


garlick:
Heh, this works
```
$ sudo flux module remove sched-fluxion-feasibility
$ sudo flux module remove sched-fluxion-qmanager
$ sudo flux module reload --exec sched-fluxion-resource
$ sudo flux module load --exec sched-fluxion-qmanager
$ sudo flux module load --exec sched-fluxion-feasibility

$ systemctl status flux
     Loaded: loaded (/lib/systemd/system/flux.service; enabled; preset: enabled)
     Active: active (running) since Mon 2025-09-22 09:39:44 PDT; 4min 3s ago
    Process: 204222 ExecStartPre=/usr/bin/loginctl enable-linger flux (code=exited, status=0/SUCCESS)
    Process: 204223 ExecStartPre=bash -c systemctl start user@$(id -u flux).service (code=exited, status=0/>
   Main PID: 204225 (flux-broker-0)
     Status: "Running as leader of 8 node Flux instance"
      Tasks: 25 (limit: 9575)
     Memory: 29.8M
        CPU: 1.881s
     CGroup: /system.slice/flux.service
             ├─204225 broker --config-path=/etc/flux/system/conf.d -Scron.directory=/etc/flux/system/cron.d>
             ├─204323 module-exec /usr/lib/aarch64-linux-gnu/flux/modules/sched-fluxion-resource.so
             ├─204328 module-exec /usr/lib/aarch64-linux-gnu/flux/modules/sched-fluxion-qmanager.so
             └─204331 module-exec /usr/lib/aarch64-linux-gnu/flux/modules/sched-fluxion-feasibility.so

```

garlick:
Two problems I just noticed:

1)  The following should have been handled like a "module panic" (mod_main exits with nonzero code), but instead it was handled like a clean exit:
```
$ sudo pkill -SEGV -f fluxion-resource
Segmentation fault (core dumped)
$ flux dmesg
2025-09-22T09:46:20.228031-07:00 broker.err[0]: sched-fluxion-resource: killed by Segmentation fault
2025-09-22T09:46:20.228061-07:00 broker.debug[0]: module sched-fluxion-resource exited
```

2) That didn't cause the usual failures of feasibilty and qmanager.  I would expect
```
2025-09-22T09:54:01.010412-07:00 broker.debug[0]: rmmod sched-fluxion-resource
2025-09-22T09:54:01.010557-07:00 sched-fluxion-feasibility.err[0]: update_resource: exiting due to sched-fluxion-resource.notify failure: Operation canceled
2025-09-22T09:54:01.010709-07:00 sched-fluxion-qmanager.err[0]: update_on_resource_response: exiting due to sched-fluxion-resource.notify failure: Operation canceled
2025-09-22T09:54:01.010888-07:00 broker.debug[0]: module sched-fluxion-resource exited
2025-09-22T09:54:01.011144-07:00 broker.debug[0]: module sched-fluxion-qmanager exited
```
but instead they kept running with no complaints.

I will run those down.

garlick:
Fixed those things and rebased on current master, but I'll leave this WIP for now as some more tests should be added for the above failures.

Forcing a segfault in `sched-fluxion-resource` now causes the other two modules to exit and the "module runtime failure" message to be logged, which would trigger an instance exit in a non-system instance (what I called a module panic above).  That is the desired behavior I think.
```
[Sep22 18:02] broker.err[0]: sched-fluxion-resource: killed by Segmentation fault
[  +0.000029] broker.debug[0]: module sched-fluxion-resource exited
[  +0.000041] broker.crit[0]: sched-fluxion-resource module runtime failure
[  +0.000153] resource.debug[0]: aborted 1 resource.acquire(s)
[  +0.000329] sched-fluxion-qmanager.err[0]: update_on_resource_response: exiting due to sched-fluxion-resource.notify failure: No route to host
[  +0.000281] sched-fluxion-feasibility.err[0]: update_resource: exiting due to sched-fluxion-resource.notify failure: No route to host
[  +0.001164] broker.debug[0]: module sched-fluxion-qmanager exited
[  +0.001260] job-manager.debug[0]: alloc: stop due to disconnect: Success
[  +0.001652] broker.debug[0]: module sched-fluxion-feasibility exited
```

garlick:
I added some tests so I'll remove the WIP.

I'll just note that I restarted the flux-sched builder just now due to this failure, which makes me scratch my head wondering if I might have caused it, but not coming up with anything yet:
```
removing load-file feasibility module works
2025-09-23T14:53:34.3325511Z 
2025-09-23T14:53:34.3325630Z expecting success: 
2025-09-23T14:53:34.3325938Z     load_feasibility &&
2025-09-23T14:53:34.3326279Z     flux dmesg -c | grep -q err &&
2025-09-23T14:53:34.3326720Z     ! flux module list | grep -q sched-fluxion-feasib
2025-09-23T14:53:34.3327221Z 
2025-09-23T14:53:34.3327527Z not ok 10 - loading feasibility module from load-file resource module fails
```

garlick:
Fixed a couple of minor issues discovered when running the test suite with all modules loaded with `--exec`:
- don't log an error when module subprocesses exits with rc=0
- enable sanity check on legacy `mod_name` symbol if it exists

I added a test for the latter.  Just squashed those and force pushed.

grondo:
Just started checking this out. Great work!

Some early trivial comments:
 - It might be cool if `flux-module-exec` could reset its process title (where available) to the module name, making it easier to see modules in `top`, `ps`, etc.
 - Don't forget to add the `flux module --exec` option to tab completions
 - It might be nice if the message logged by the broker when loading a module as a subprocess called that out, e.g. instead of `insmod sched-simple`, perhaps `insmod sched-simple exec pid=<pid>` or similar

garlick:
Nice :-)  Thank you for doing the performance test!  I had thought of doing that but didn't return to it.  Yes not too bad!

garlick:
OK, I'll set MWP then - thanks!

Issue #7077, opened 2025-09-23
modprobe: add support for selectively loading modules as processes vs threads


garlick:
Problem: `flux-modprobe` does not currently support the `exec` option for module loading.

For a start we probably just need a way to designate a module as "always load as a process" in the `modprobe.toml`.

grondo:
Here's a proof of concept that adds an `exec` boolean to a module config. Defaults to false:
```diff
diff --git a/src/bindings/python/flux/modprobe.py b/src/bindings/python/flux/modprobe.py
index 2345ff3d4..db0311294 100644
--- a/src/bindings/python/flux/modprobe.py
+++ b/src/bindings/python/flux/modprobe.py
@@ -292,6 +292,7 @@ class Module(Task):
         "needs-config",
         "priority",
         "disabled",
+        "exec",
     )
 
     def __init__(self, conf):
@@ -306,6 +307,7 @@ class Module(Task):
 
         self.module = conf.get("module", name)
         self.args = conf.get("args", [])
+        self.exec = conf.get("exec", False)
         self.run = self._load
 
         # Build kwargs to pass along to Task class
@@ -333,7 +335,7 @@ class Module(Task):
 
     def _load(self, context):
         args = context.getopts(self.name, default=self.args, also=self.provides)
-        payload = {"path": self.module, "args": args}
+        payload = {"path": self.module, "args": args, "exec": self.exec}
 
         if self.name != self.module:
             payload["name"] = self.name
```

Since the attribute can be overridden either via dropped-in TOML to `modprobe.d/` or via the broker config, we can run an instance with sched-simple as a process easily (for example) like:
```console
$ flux alloc -N1 --conf=modules.sched-simple.exec=true
$ pstree -Tap $(flux getattr broker.pid)
flux-broker-0,2417382-c/tmp/flux-vcsgKZ/jobtmp-0-
  ├─bash,2417498
  │   └─pstree,2417548 -Tap 2417382
  └─flux-module-exe,2417455/home/grondo/git/f.git/
```

PR #7080, opened 2025-09-24
modprobe: support module `exec` option to load modules as processes

grondo:
This PR addresses #7077 with a new `exec` option for modules to request they be loaded as processes instead of threads.


grondo:
Thanks, setting MWP.

PR #7081, opened 2025-09-24
python: support kvs_lookup_checkpoint

chu11:
Problem: There is no convenient way to lookup the KVS checkpoint via the python API.

Support a new kvs_lookup_checkpoint() function that mirrors the equivalent C function.


chu11:
hmm, had the el9 test-install builder fail valgrind with

```
 ==191134== 149 (128 direct, 21 indirect) bytes in 1 blocks are definitely lost in loss record 5 of 11
  ==191134==    at 0x484C184: realloc (vg_replace_malloc.c:1690)
  ==191134==    by 0x1165CA3F: ??? (in /usr/lib64/libhwloc.so.15.4.1)
  ==191134==    by 0x1167E3FE: ??? (in /usr/lib64/libhwloc.so.15.4.1)
  ==191134==    by 0x11668004: hwloc_topology_load (in /usr/lib64/libhwloc.so.15.4.1)
  ==191134==    by 0x11634500: init_topo_from_xml (rhwloc.c:75)
  ==191134==    by 0x11634500: rhwloc_xml_topology_load (rhwloc.c:109)
  ==191134==    by 0x11634AC5: rlist_from_hwloc (rhwloc.c:322)
  ==191134==    by 0x1162C826: topo_create (topo.c:337)
  ==191134==    by 0x1162B50F: mod_main (resource.c:411)
  ==191134==    by 0x40F392: module_thread (module_thread.c:229)
  ==191134==    by 0x4C66C01: start_thread (in /usr/lib64/libc.so.6)
  ==191134==    by 0x4CEAED3: clone (in /usr/lib64/libc.so.6)
  ==191134== 
```

This PR had no changes to C code, so the error seems unrelated.

src/bindings/python/flux/kvs.py
```diff
@@ -911,3 +911,24 @@ def kvs_watch_async(
...
+
+
+def kvs_checkpoint_lookup(flux_handle, cache_bypass=False):
+    """Lookup KVS checkpoint(s)
+
+    Args:
+        flux_handle: A Flux handle obtained from flux.Flux()
+        cache_bypass: lookup directly to backing store
+
+    :return: array of checkpoint objects
```
grondo: Mixing Google Python docstring style and reST style here I think.

[Here's](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) a reference to the Google style. I think you can use:
```
    Returns:
        list: array of checkpoint objects
```

src/bindings/python/flux/kvs.py
```diff
@@ -911,3 +911,24 @@ def kvs_watch_async(
...
+    """
+
+    topic = "content.checkpoint-get"
+    if cache_bypass:
+        topic = "content-backing.checkpoint-get"
+    rsp = RPC(flux_handle, topic)
+    a = rsp.get()["value"]
+    # backwards compatibility, lookup returns a single checkpoint dict
+    if isinstance(a, dict):
+        return list(a)
```
grondo: I think that `list(a)` will just return a list of the keys in dict `a`, not return a list of length 1 containing `a` which I assume is what you want here.

E.g.:
```python
>>> print(a)
{'foo': 42}
>>> list(a)
['foo']
```

What you want is `[a]`:
```python
>>> [a]
[{'foo': 42}]
```

chu11: ahhh good catch

src/bindings/python/flux/kvs.py
chu11: Ahhh, was cutting and pasting from elsewhere.  Will fix elsewhere too!

chu11:
re-pushed with fixes discussed above, also added a commit to fix an unrelated documentation error.  Will set MWP

PR #7082, opened 2025-09-24
etc/modprobe: run flux-fsck in rc1 

chu11:
Problem: It would be useful to run flux-fsck in rc1 on broker startup.  That way any corruption errors could be found on startup
if prior shutdown was not clean.

Add an "fsck" task to rc1.py that runs flux-fsck as needed.

Built on top of #7081

### DIFF
--- a/t/system/0007-fsck.t
+++ b/t/system/0007-fsck.t
@@ -1,0 +1,129 @@
+#
+#  Ensure flux is fscked when not shutdown cleanly
+#
+
+test_expect_success 'get the current time for journalctl --since' '
+	date +"%F %T" > fscktime.out
+	cat fscktime.out
+'
+
+test_expect_success 'restart flux' '
+	sudo systemctl restart flux
+'
+
+wait_flux_back_up() {
+	i=0
+	while ! flux resource list > /dev/null 2>&1 \
+	      && [ $i -lt 100 ]
+	do
+		sleep 1
+		i=$((i + 1))
+	done
+	if [ "$i" -eq "100" ]
+	then
+		return 1
+	fi
+	return 0
+}
+
+test_expect_success 'wait for flux to finish starting up' '
+	wait_flux_back_up
+'
+
+test_expect_success 'fsck does not run after clean shutdown' '
+	sudo journalctl --since "$(cat fscktime.out)" | grep rc1 > log1.out &&
+	test_must_fail grep "Checking integrity" log1.out
+'
+
+test_expect_success 'kill flux broker' '
+	pid=`sudo systemctl show --property=MainPID --value flux` &&
+	sudo kill -s 9 $pid
+'
+
+# system instance for flux will auto restart in about 30 seconds
+test_expect_success 'wait for flux to finish starting up' '
+	wait_flux_back_up
+'
+
+test_expect_success 'fsck runs after unclean shutdown' '
+	sudo journalctl --since "$(cat fscktime.out)" | grep rc1 > log2.out &&
+	grep "Checking integrity" log2.out &&
+	test_must_fail grep "Total errors" log2.out
+'
+
+test_expect_success 'corrupt some data' '
+	sudo flux kvs put testdata=1 &&
+	sudo flux kvs put --append testdata=2 &&
+	sudo flux kvs put --append testdata=3 &&
+	sudo flux kvs get --treeobj testdata > testdata.treeobj &&
+	cat testdata.treeobj | jq -c .data[1]=\"sha1-1234567890123456789012345678901234567890\" > testdata.bad &&
+	sudo flux kvs put --treeobj testdatabad="$(cat testdata.bad)" &&
+	test_must_fail sudo flux kvs get testdatabad
+'
+
+test_expect_success 'call sync to ensure we have checkpointed' '
+	sudo flux kvs sync
+'
+
+test_expect_success 'kill flux broker' '
+	pid=`sudo systemctl show --property=MainPID --value flux` &&
+	sudo kill -s 9 $pid
+'
+
+wait_fsck_fail() {
+	i=0
+	while ! sudo journalctl --since "$(cat fscktime.out)" | grep "missing blobref" > /dev/null 2>&1 \
+	      && [ $i -lt 100 ]
+	do
+		sleep 1
+		i=$((i + 1))
+	done
+	if [ "$i" -eq "100" ]
+	then
+		return 1
+	fi
+	return 0
+}
+
+# system instance for flux will auto restart in about 30 seconds
+test_expect_success 'wait for flux to restart and fail fsck' '
+	wait_fsck_fail
+'
+
+wait_flux_stopped() {
+	i=0
+	while sudo systemctl status flux > /dev/null 2>&1 \
+	      && [ $i -lt 100 ]
+	do
+		sleep 1
+		i=$((i + 1))
+	done
+	if [ "$i" -eq "100" ]
+	then
+		return 1
+	fi
+	return 0
+}
+
+test_expect_success 'flux system instance is not running' '
+	wait_flux_stopped &&
+	test_must_fail sudo systemctl status flux
+'
+
+test_expect_success 'fsck repaired KVS corruption' '
+	sudo journalctl --since "$(cat fscktime.out)" | grep rc1 > log3.out &&
+	grep "Total errors: 1" log3.out &&
+	grep "Total repairs: 1" log3.out
+'
+
+test_expect_success 'fsck moved bad data to lost+found' '
+	sudo -u flux flux start --recovery "flux kvs get lost+found.testdatabad"
+'
+
+test_expect_success 'restart flux' '
+	sudo systemctl start flux
+'
+
+test_expect_success 'wait for flux to finish starting back up' '
+	wait_flux_back_up
+'
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -357,6 +357,7 @@ dist_check_SCRIPTS = \
 	system/0004-recovery.t \
 	system/0005-exec.t \
 	system/0006-perilog.t \
+	system/0007-fsck.t \
 	issues/t0441-kvs-put-get.sh \
 	issues/t0505-msg-handler-reg.lua \
 	issues/t0821-kvs-segfault.sh \


chu11:
> Is there a way to test that fsck is being run when expected? It appears flux fsck emits the following on stderr when run:

I literally got on my laptop this morning to type "Oh, I just realized I didn't add tests, I only hand tested yesterday.", but you beat me to it by 3 minutes :-)

Yes, was going to look into it this morning.

grondo:
One other random thought. Since `flux fsck` outputs to stderr by default, a user starting an instance from an existing content store would always see the messages above labeled with `broker.err`. Perhaps this really isn't a concern, but should the messages be logged to stdout so they don't get redirected to user's stderr by default? I doubt this needs to be addressed anytime soon, but just thought I'd throw it out there.

chu11:
> One other random thought. Since flux fsck outputs to stderr by default, a user starting an instance from an existing content store would always see the messages above labeled with broker.err. Perhaps this really isn't a concern, but should the messages be logged to stdout so they don't get redirected to user's stderr by default? I doubt this needs to be addressed anytime soon, but just thought I'd throw it out there.

Hmmm.  In my mind initially the primary use was for the system instance.  If there was actual corruption `flux fsck` would return an error, the `modprobe rc1` should therefore fail, and then the flux broker wouldn't start.  So the errors would be sitting the `journalctl`, so the only difference would be if the broker logs to broker.info or broker.err I think?  So I think stderr would be more correct so we see it in the broker err stream?

Were you thinking of the scenario where a user starts flux via `flux start`, b/c I think that also gets sent to the broker logs too?

garlick:
I assumed the point was to be careful about what gets logged to stderr vs stdout because the broker does:
```c
if ((len = flux_subprocess_getline (p, stream, &line)) > 0) {
        if (streq (stream, "stderr"))
            flux_log (r->h, LOG_ERR, "%s.%d: %s", entry->name, index, line);
        else
            flux_log (r->h, LOG_INFO, "%s.%d: %s", entry->name, index, line);
    }
}
```
And the default log level makes sure LOG_ERR goes to the broker stderr, while LOG_INFO does not.

chu11:
> and the default log level makes sure LOG_ERR goes to the broker stderr, while LOG_INFO does not.

Ok I get it now.  For the average person just starting a instance on their won, maybe we wouldn't want the fsck to show up in the stderr output / logs.

chu11:
Just re-pushed, adding some tests to the system instance one (`system/0007-fsck.t`) as I wanted to emulate a "system instance".

But, what is missing is fsck actually capturing corruption.  I think I need to create a `flux content remove` kinda utility as has been discussed in several other issues (such as #6589).  Setting WIP for now.

grondo:
> Ok I get it now. For the average person just starting a instance on their won, maybe we wouldn't want the fsck to show up in the stderr output / logs.

Yeah, stdout would be fine though since it wouldn't be copied to the user's terminal unless there's an error.

etc/modprobe/rc1.py
```diff
@@ -137,6 +141,36 @@ def check_clean_shutdown(context):
...
+        if e.errno != errno.ENOENT:
+            raise
+        return
+    if not context.get("dump_restored", False):
+        context.bash("flux fsck")
+
+
 @task(
     "cron-load",
     ranks="0",
```
garlick, line 181: I forgot about this - maybe fsck should simply be triggered here:
```python
@task(
    "check-clean-shutdown",
    ranks="0",
    after=["post-start-event"],
    requires=["post-start-event"],
    needs=["content-backing"],
)
def check_clean_shutdown(context):
    context.bash(
        """
    if ! flux startlog --check --quiet; then
        echo "Flux was not shut down properly.  Data may have been lost."
    fi
    """
    )
```

chu11: I considered that, but it comes after the KVS is loaded, so fsck wouldn't work.  Unless you think doing an unload/load within that task is a good idea?  I haven't thought if there would be consequences to that yet (minimally when kvs is shutdown cleanly, the checkpoint is updated ... so then then fsck might not be good)

garlick: Oops you're right, that would not be good.

chu11:
hmmm, the el8 CI is always failing w/ my new fsck tests ... guessing there is a some systemctl subtlety in there i don't handle

Edit: hmmm, my local container is based on Rocky 8.10 .. must be some other subtlety in there

chu11:
So this is super unexpected.  On my local container w/ el8 I get

```
Oct 01 16:45:16 fluxorama flux[564]: broker.err[0]: rc1.0: Checking integrity of checkpoint from 2025-10-01T16:44:46
Oct 01 16:45:16 fluxorama flux[564]: broker.err[0]: rc1.0: testdatabad: missing blobref(s)
Oct 01 16:45:16 fluxorama flux[564]: broker.err[0]: rc1.0: Total errors: 1
Oct 01 16:45:16 fluxorama flux[564]: kvs.info[0]: restored KVS from checkpoint on 2025-10-01T16:44:46Z
Oct 01 16:45:16 fluxorama flux[564]: broker.info[0]: rc1.0: Flux was not shut down properly.  Data may have been lost.
Oct 01 16:45:16 fluxorama flux[564]: Ignoring PCI device with non-16bit domain.
Oct 01 16:45:16 fluxorama flux[564]: Pass --enable-32bits-pci-domain to configure to support such devices
Oct 01 16:45:16 fluxorama flux[564]: (warning: it would break the library ABI, don't enable unless really needed).
Oct 01 16:45:16 fluxorama flux[564]: job-manager.info[0]: restart: 1 jobs
Oct 01 16:45:16 fluxorama flux[564]: job-manager.info[0]: restart: 0 running jobs
Oct 01 16:45:16 fluxorama flux[564]: broker.err[0]: rc1.0: fsck: bash: exited with exit status 1
Oct 01 16:45:16 fluxorama flux[564]: broker.err[0]: rc1.0: flux modprobe rc1 Exited (rc=1) 0.1s
Oct 01 16:45:16 fluxorama flux[564]: broker.info[0]: rc1-fail: init->shutdown 0.139574s
Oct 01 16:45:16 fluxorama flux[564]: broker.info[0]: children-none: shutdown->finalize 0.02654ms
Oct 01 16:45:16 fluxorama flux[564]: broker.err[0]: rc3.0: testdatabad: missing blobref 1: No such file or directory
Oct 01 16:45:16 fluxorama flux[564]: broker.info[0]: rc3.0: dumping content to /var/lib/flux/dump/20251001_164516.tgz
Oct 01 16:45:16 fluxorama flux[564]: broker.info[0]: rc3.0: flux modprobe rc3 Exited (rc=0) 0.1s
Oct 01 16:45:16 fluxorama flux[564]: broker.info[0]: rc3-success: finalize->goodbye 97.0349ms
Oct 01 16:45:16 fluxorama flux[564]: broker.info[0]: goodbye: goodbye->exit 0.067393ms
```

So fsck says the KVS is bad, ultimately leading to rc1 failure and broker doesn't startup.  Although I'm a little surprised rc3 runs, but i suspect that's just part of the shutdown / cleanup path of the broker. (Edit: I now see that rc3 is run on finalize within `action_finalize()` in the broker)

Under the  el8 / system builder

```
Oct 01 18:43:07 fluxorama flux[91143]: broker.err[0]: rc1.0: Checking integrity of checkpoint from 2025-10-01T18:42:36
Oct 01 18:43:07 fluxorama flux[91143]: broker.info[0]: rc1.0: tab: cron-1 created: scheduled in 29812.579s at Thu Oct  2 03:00:00 2025
Oct 01 18:43:07 fluxorama flux[91143]: broker.err[0]: rc1.0: testdatabad: missing blobref(s)
Oct 01 18:43:07 fluxorama flux[91143]: broker.err[0]: rc1.0: Total errors: 1
Oct 01 18:43:07 fluxorama flux[91143]: kvs.info[0]: restored KVS from checkpoint on 2025-10-01T18:42:36Z
Oct 01 18:43:07 fluxorama flux[91143]: broker.info[0]: rc1.0: Flux was not shut down properly.  Data may have been lost.
Oct 01 18:43:07 fluxorama flux[91143]: job-manager.info[0]: restart: 8 jobs
Oct 01 18:43:07 fluxorama flux[91143]: job-manager.info[0]: restart: 0 running jobs
Oct 01 18:43:07 fluxorama flux[91143]: broker.err[0]: rc1.0: fsck: bash: exited with exit status 1
Oct 01 18:43:07 fluxorama flux[91143]: broker.err[0]: rc1.0: flux modprobe rc1 Exited (rc=1) 0.2s
Oct 01 18:43:07 fluxorama flux[91143]: broker.info[0]: rc1-fail: init->shutdown 0.213011s
Oct 01 18:43:07 fluxorama flux[91143]: broker.info[0]: children-none: shutdown->finalize 0.0271ms
```

This is it.  The shutdown doesn't appear to happen, the broker is still running even though rc1 fails and we clearly get an `init->shutdown` log message.  Super confused, trying to figure out what's going on.

Edit: hmmm, maybe it's just and racy .... rc3 has just not started yet before the tests expect flux to have shutdown

chu11:
ok, test raciness appears to be fixed!

t/system/0007-fsck.t
```diff
@@ -0,0 +1,124 @@
...
+	test_must_fail sudo flux kvs get testdatabad
+'
+
+test_expect_success 'call sync to ensure we have checkpointed' '
+	sudo flux kvs sync
+'
+
+test_expect_success 'kill flux broker' '
+	pid=`sudo systemctl show --property=MainPID --value flux` &&
+	sudo kill -s 9 $pid
```
grondo, line 70: Just as an fyi, the flux way to do this is `flux getattr broker.pid`. However, nothing wrong with this approach.

garlick:
I'll give this a spin on my test cluster real quick.

If corruption is found, presumably `flux fsck` fails and rc1 fails.  But didn't you show earlier that a dump is created even when that happens?

garlick:
One nit, the informational message "Checking integrity" is going to stderr, which causes it to be logged at LOG_ERR level.
```
[  +0.163835] content.debug[0]: content backing store: enabled content-sqlite
[  +0.170101] broker.err[0]: rc1.0: Checking integrity of checkpoint from 2025-10-02T09:46:44
```
That one should probably go to stdout so it gets logged at LOG_INFO level.

Edit: this too (when there are no errors)
```
[  +0.256976] broker.err[0]: rc1.0: Total errors: 0
```

garlick:
Also using `fprintf (stderr,` for these messages instead of `log_err()` means that they aren't prefixed with the program name, which would be handy when interspersed with other output in the logs.

garlick:
I did a little test where I deleted 20 random content blobs and started flux up
```
# systemctl stop flux
# rm /var/lib/flux/dump/RESTORE
# sqlite3 /var/lib/flux/dump/content.sqlite
sqlite3 > sqlite> delete from objects where rowid in (select rowid from objects limit 20);
^D
# systemctl start flux
```

The systemd journal contains
```
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: Checking integrity of checkpoint from 2025-10-02T09:54:55
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: admin.eventlog: missing blobref(s)
Oct 02 10:13:47 picl0 flux[56143]: broker.info[0]: rc1.0: tab: cron-1 created: scheduled in 60372.635s at Fri Oct  3 03:00:00 2025
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: job.015f.a49f: missing dirref blobref
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: job.015f.a4bb.5200.0000.J: missing blobref(s)
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: job.0183.2609.4500.0000: missing dirref blobref
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: job.0183.423e.f100.0000: missing dirref blobref
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: job.0187.060a.2200.0000.guest.exec.eventlog: missing blobref(s)
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: job.0198.0c08.bc00.0000: missing dirref blobref
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: job.0198.1079.8500.0000.R: missing blobref(s)
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: job.0198.11ad.cf00.0000.J: missing blobref(s)
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: job.0198.2d86.8000.0000.guest.output: missing blobref(s)
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: job.019a.a2d2.d200.0000.guest.output: missing blobref(s)
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: job.019a.dbb8.f500: missing dirref blobref
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: job.019b.6882.9900.0000: missing dirref blobref
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: job.01dc.5a38.cd00.0000.guest.output: missing blobref(s)
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: job.01dc.e578: missing dirref blobref
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: job.01de.b0a3.6a00.0000.guest.output: missing blobref(s)
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: job.01de.b4b4.5900.0000.R: missing blobref(s)
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: Total errors: 17
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: fsck: bash: exited with exit status 1
Oct 02 10:13:47 picl0 flux[56143]: kvs.info[0]: restored KVS from checkpoint on 2025-10-02T16:54:55Z
Oct 02 10:13:47 picl0 flux[56143]: kvs.err[0]: content_load_completion: content_load_get: No such file or directory
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: flux: Error fetching eventlog: No such file or directory
Oct 02 10:13:47 picl0 flux[56143]: broker.info[0]: rc1.0: Flux was not shut down properly.  Data may have been lost.
Oct 02 10:13:47 picl0 flux[56143]: sched-fluxion-resource.info[0]: version 0.47.0
Oct 02 10:13:47 picl0 flux[56143]: sched-fluxion-resource.warning[0]: create_reader: allowlist unsupported
Oct 02 10:13:47 picl0 flux[56143]: sched-fluxion-resource.info[0]: populate_resource_db: loaded resources from core's resource.acquire
Oct 02 10:13:47 picl0 flux[56143]: sched-fluxion-feasibility.info[0]: version 0.47.0
Oct 02 10:13:47 picl0 flux[56143]: sched-fluxion-feasibility.warning[0]: create_reader: allowlist unsupported
Oct 02 10:13:47 picl0 flux[56143]: sched-fluxion-feasibility.info[0]: populate_resource_db: loaded resources from sched-fluxion-resource.notify
Oct 02 10:13:47 picl0 flux[56143]: kvs.err[0]: content_load_completion: content_load_get: No such file or directory
Oct 02 10:13:47 picl0 flux[56143]: job-manager.err[0]: restart failed: could not look up job.015f.a49f: No such file or directory
Oct 02 10:13:47 picl0 flux[56143]: job-manager.crit[0]: module exiting abnormally
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: job-manager: [Errno 2] No such file or directory
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: coral2-jobtaps-all: [Errno 38] No service matching job-manager.jobtap is registered
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: job-manager: [Errno 2] No such file or directory
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: coral2-jobtaps-all: [Errno 38] No service matching job-manager.jobtap is registered
Oct 02 10:13:47 picl0 flux[56143]: job-exec.err[0]: job-manager.exec-hello: Function not implemented
Oct 02 10:13:47 picl0 flux[56143]: sched-fluxion-qmanager.info[0]: version 0.47.0
Oct 02 10:13:47 picl0 flux[56143]: job-ingest.err[0]: job-manager must be loaded first: Function not implemented
Oct 02 10:13:47 picl0 flux[56143]: job-ingest.crit[0]: module exiting abnormally
Oct 02 10:13:47 picl0 flux[56143]: job-exec.crit[0]: module exiting abnormally
Oct 02 10:13:47 picl0 flux[56143]: job-list.err[0]: journal: No service matching job-manager.events-journal is registered
Oct 02 10:13:47 picl0 flux[56143]: job-list.crit[0]: module exiting abnormally
Oct 02 10:13:47 picl0 flux[56143]: broker.crit[0]: job-list module runtime failure
Oct 02 10:13:47 picl0 flux[56143]: sched-fluxion-qmanager.err[0]: handshake_jobmanager: schedutil_hello: Function not implemented
Oct 02 10:13:47 picl0 flux[56143]: sched-fluxion-qmanager.err[0]: handshake: handshake_jobmanager: Function not implemented
Oct 02 10:13:47 picl0 flux[56143]: sched-fluxion-qmanager.err[0]: mod_start: handshake: Function not implemented
Oct 02 10:13:47 picl0 flux[56143]: sched-fluxion-qmanager.crit[0]: module exiting abnormally
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: job-ingest: [Errno 38] Function not implemented
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: job-exec: [Errno 38] Function not implemented
Oct 02 10:13:47 picl0 flux[56143]: broker.crit[0]: sched-fluxion-qmanager module runtime failure
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: /etc/flux/rc1.d/01-rabbit-rc: 3: [: missing ]
Oct 02 10:13:47 picl0 flux[56143]: broker.info[0]: rc1.0: running /etc/flux/rc1.d/01-coral2-rc
Oct 02 10:13:47 picl0 flux[56143]: broker.info[0]: rc1.0: running /etc/flux/rc1.d/01-rabbit-rc
Oct 02 10:13:47 picl0 flux[56143]: broker.info[0]: rc1.0: running /etc/flux/rc1.d/01-sched-fluxion
Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: flux modprobe rc1 Exited (rc=1) 0.5s
Oct 02 10:13:47 picl0 flux[56143]: broker.info[0]: rc1-fail: init->shutdown 0.482678s
Oct 02 10:13:48 picl0 flux[56143]: broker.info[0]: children-complete: shutdown->finalize 1.00685s
Oct 02 10:13:48 picl0 flux[56143]: broker.err[0]: rc3.0: admin/eventlog: missing blobref 0: No such file or directory
Oct 02 10:13:48 picl0 flux[56143]: broker.err[0]: rc3.0: job/015f/a49f: missing blobref: No such file or directory
Oct 02 10:13:48 picl0 flux[56143]: broker.err[0]: rc3.0: job/015f/a4bb/5200/0000/J: missing blobref 0: No such file or directory
Oct 02 10:13:48 picl0 flux[56143]: broker.err[0]: rc3.0: job/0183/2609/4500/0000: missing blobref: No such file or directory
Oct 02 10:13:48 picl0 flux[56143]: broker.err[0]: rc3.0: job/0183/423e/f100/0000: missing blobref: No such file or directory
Oct 02 10:13:48 picl0 flux[56143]: broker.err[0]: rc3.0: job/0187/060a/2200/0000/guest/exec/eventlog: missing blobref 0: No such file or directory
Oct 02 10:13:48 picl0 flux[56143]: broker.err[0]: rc3.0: job/0198/0c08/bc00/0000: missing blobref: No such file or directory
Oct 02 10:13:48 picl0 flux[56143]: broker.err[0]: rc3.0: job/0198/1079/8500/0000/R: missing blobref 0: No such file or directory
Oct 02 10:13:48 picl0 flux[56143]: broker.err[0]: rc3.0: job/0198/11ad/cf00/0000/J: missing blobref 0: No such file or directory
Oct 02 10:13:48 picl0 flux[56143]: broker.err[0]: rc3.0: job/0198/2d86/8000/0000/guest/output: missing blobref 0: No such file or directory
Oct 02 10:13:48 picl0 flux[56143]: broker.err[0]: rc3.0: job/019a/a2d2/d200/0000/guest/output: missing blobref 0: No such file or directory
Oct 02 10:13:48 picl0 flux[56143]: broker.err[0]: rc3.0: job/019a/dbb8/f500: missing blobref: No such file or directory
Oct 02 10:13:48 picl0 flux[56143]: broker.err[0]: rc3.0: job/019b/6882/9900/0000: missing blobref: No such file or directory
Oct 02 10:13:48 picl0 flux[56143]: broker.err[0]: rc3.0: job/01dc/5a38/cd00/0000/guest/output: missing blobref 0: No such file or directory
Oct 02 10:13:48 picl0 flux[56143]: broker.err[0]: rc3.0: job/01dc/e578: missing blobref: No such file or directory
Oct 02 10:13:48 picl0 flux[56143]: broker.err[0]: rc3.0: job/01de/b0a3/6a00/0000/guest/output: missing blobref 0: No such file or directory
Oct 02 10:13:48 picl0 flux[56143]: broker.err[0]: rc3.0: job/01de/b4b4/5900/0000/R: missing blobref 0: No such file or directory
Oct 02 10:13:48 picl0 flux[56143]: broker.info[0]: rc3.0: running /etc/flux/rc3.d/01-sched-fluxion
Oct 02 10:13:48 picl0 flux[56143]: broker.info[0]: rc3.0: dumping content to /var/lib/flux/dump/20251002_101348.tgz
Oct 02 10:13:49 picl0 flux[56143]: broker.info[0]: rc3.0: flux modprobe rc3 Exited (rc=0) 0.3s
Oct 02 10:13:49 picl0 flux[56143]: broker.info[0]: rc3-success: finalize->goodbye 0.324665s
Oct 02 10:13:49 picl0 flux[56143]: broker.info[0]: goodbye: goodbye->exit 0.091389ms
Oct 02 10:13:49 picl0 systemd[1]: flux.service: Deactivated successfully.
```
We need to not let the rest of rc1 run in parallel with this check, and the dump at the end shouldn't happen.

chu11:
> That one should probably go to stdout so it gets logged at LOG_INFO level.

Hmmm.  stderr feels correct for the "command line usage" case.  I'm thinking we should have a "--to-stdout" kinda option?

> We need to not let the rest of rc1 run in parallel with this check

> Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: Total errors: 17
> Oct 02 10:13:47 picl0 flux[56143]: broker.err[0]: rc1.0: fsck: bash: exited with exit status 1
> Oct 02 10:13:47 picl0 flux[56143]: kvs.info[0]: restored KVS from checkpoint on 2025-10-02T16:54:55Z
> Oct 02 10:13:47 picl0 flux[56143]: kvs.err[0]: content_load_completion: content_load_get: No such file or directory

hmmm, something is not right here.  `fsck` runs before `kvs` is loaded.  My presumption was that if the `fsck` task fails, then nothing after it will be run.  Lemme look into this further.

> and the dump at the end shouldn't happen.

Good catch!  Lemme see how a failed rc1 could "notify" rc3 to not dump.    It appears the state machine state doesn't show history.  If we could see "rc1-fail" in the state machine history, we could easily know not to do a dump.

garlick:
> lemme see how a failed rc1 could "notify" rc3 to not dump. It appears the state machine state doesn't show history. If we could see "rc1-fail" in the state machine history, we could easily know not to do a dump.

Maybe a better way would be to just have the fsck task unset the `content.dump` attribute on failure.

grondo:
> We need to not let the rest of rc1 run in parallel with this check, and the dump at the end shouldn't happen.

Hm, I'm not sure there's a way to terminate the rest of rc1 when a task fails in modprobe at the moment. We'll have to add that support.

Also, the fsck task already has `before=["kvs"]`, so the kvs module can't be loaded until the check passes. If we then make it so rc1 immediately terminates with failure if this task fails, would that be suffiicient rather than turning the fsck task (which currently always has to run) into a sort of critical section that prevents all other work from running? 

Apologies if I'm missing an obvious reason why other tasks that don't use the KVS can't run in parallel with this check (perhaps some modules could use the content store directly)

garlick:
Apologies, you're right!  This doesn't need to be a critical section.   It seems like we could get the right effect if, when this task fails, we either exit rc1 immediately, or we just prune the tree when there is a failure (e.g. kvs and its dependencies don't start)

grondo:
I'll open an issue to add that to modprobe

#7105

garlick:
>  That one should probably go to stdout so it gets logged at LOG_INFO level.
>> Hmmm. stderr feels correct for the "command line usage" case. I'm thinking we should have a "--to-stdout" kinda option?

Maybe just run `fsck --quiet` from rc1.

chu11:
> Maybe just run fsck --quiet from rc1.

Hmmmm, but we want something to output, right?  Or perhaps the `broker.err[0]: rc1.0: fsck: bash: exited with exit status 1` is enough?

garlick:
Seems like we could do without any output when there are no errors, but if there are errors we probably want all the detail.  I just assumed that's what `--quiet` does.

chu11:
> Seems like we could do without any output when there are no errors, but if there are errors we probably want all the detail. I just assumed that's what --quiet does.

Quiet suppresses everything at the moment.

Perhaps fsck should be tweaked where there is a summary at the end whenever there are errors.  like

```
Checked integrity of checkpoint on 01-01-2025
Total errors: X
```

and that isn't suppressable.  `--quiet` will only suppress the "missing blobref" error messages.  And it will suppress the summary when there are no errors.

chu11:
> Also using fprintf (stderr, for these messages instead of log_err() means that they aren't prefixed with the program name, which would be handy when interspersed with other output in the logs.

Good point.  But `flux-fsck: ....` prefixes feel unnecessary when fsck is run on the command line.  Hmmm, maybe this is all moot if we just go with my summary idea / option above.

garlick:
I think what's currently logged is almost correct, except that
- the initial message that gives the (useful) checkpoint date should go to stdout (LOG_INFO), not stderr (LOG_ERR)
- the final message that lists the number of errors should be either suppressed or sent to stdout when the total is zero
- add the program name prefix

I think we definitely want any affected keys to go to the logs at LOG_ERR, especially later when we add `--repair`.

As an aside, it might be nice to have a `--job-aware` option or something that translates the KVS keys in the jobs directory to f58, and potentially moves whole job directories to lost+found instead of the individual keys.  But that would be a nice to have not a must have IMHO.

garlick:
Could always use `isatty()` to change behaviors when run interactively.

chu11:
> Could always use isatty() to change behaviors when run interactively.

Duh!  That's a really simple & good idea.

etc/modprobe/rc1.py
```diff
@@ -137,6 +140,36 @@ def check_clean_shutdown(context):
...
+)
+def fsck(context):
+    try:
+        flux.kvs.kvs_checkpoint_lookup(context.handle)
+    except OSError as e:
+        if e.errno != errno.ENOENT:
+            raise
+        return
+    if not context.get("dump_restored", False):
+        context.bash("flux fsck")
```
grondo: If you change `flux fsck` here to `flux fsck --quiet`, it might be useful to add a `context.print()` here that notifies that `flux fsck` is being run so at least this much is being logged.

chu11:
Question / Thought

When starting flux in recovery mode, would we want `flux-fsck` to be run?

A) non-trivial chance we're already in recovery mode b/c of a fsck failure, so there's no need?
B) fsck task failure will lead to `kvs` not loaded, so `--recovery` mode behaves a lot different than before, do we want to skip it?

counter point

C) the message from recovery mode is

```
            printf ("+-----------------------------------------------------\n"
                    "| Entering Flux recovery mode.\n"
                    "| All resources will be offline during recovery.\n"
                    "| Any rc1 failures noted above may result in\n"
                    "|  reduced functionality until manually corrected.\n"
```

So we may still want to fsck so errors are output so people know they should do something.

Alternately ... in the rc1 `fsck` task, we could just check for the broker attribute "broker.recovery-mode" and not run fsck, possibly just output "skipping fsck in recovery mode" or something like that.

Going back and forth on this, haven't developed a strong opinion, dunno if anyone has any other pro vs con thoughts on this.

garlick:
For now I would say run fsck in recovery mode.

chu11:
just re-pushed, rebasing on #7137.  The output from the journal looks really good now

```
Oct 10 18:31:31 fluxorama systemd[1]: Started Flux message broker.
Oct 10 18:31:32 fluxorama flux[2044]: cron.info[0]: synchronizing cron tasks to event heartbeat.pulse
Oct 10 18:31:32 fluxorama flux[2044]: broker.info[0]: rc1.0: tab: cron-1 created: scheduled in 30507.880s at Sat Oct 11 03:00:00 2025
Oct 10 18:31:32 fluxorama flux[2044]: broker.err[0]: rc1.0: flux-fsck: testdatabad: missing blobref(s)
Oct 10 18:31:32 fluxorama flux[2044]: broker.err[0]: rc1.0: flux-fsck: Total errors: 1
Oct 10 18:31:32 fluxorama flux[2044]: broker.err[0]: rc1.0: fsck: bash: exited with exit status 1
Oct 10 18:31:32 fluxorama flux[2044]: broker.info[0]: rc1.0: Running flux-fsck following unclean shutdown
Oct 10 18:31:32 fluxorama flux[2044]: broker.err[0]: rc1.0: flux modprobe rc1 Exited (rc=1) 0.2s
Oct 10 18:31:32 fluxorama flux[2044]: broker.info[0]: rc1-fail: init->shutdown 0.204505s
Oct 10 18:31:32 fluxorama flux[2044]: broker.info[0]: children-none: shutdown->finalize 0.023299ms
Oct 10 18:31:32 fluxorama flux[2044]: broker.info[0]: rc3.0: flux modprobe rc3 Exited (rc=0) 0.1s
Oct 10 18:31:32 fluxorama flux[2044]: broker.info[0]: rc3-success: finalize->goodbye 0.134794s
Oct 10 18:31:32 fluxorama flux[2044]: broker.info[0]: goodbye: goodbye->exit 0.083257ms
```

it's sort of unfortunate the stdout message of `Running flux-fsck ...` races and is after the stderr output from the actual `flux-fsck`.  Not sure if that can be helped.

garlick:
I could almost see the "on unclean shutdown" message going to stderr because something bad did happen (just not in the fsck).  Thoughts?

I actually thought the thing going to be moved to stdout was:
```
broker.err[0]: rc1.0: Checking integrity of checkpoint from 2025-10-01T18:42:36
```
which isn't an error.  Hey that's not even in the output anymore.  The date is useful...

garlick:
Also, since we're at the beginning of a release cycle, I wonder if we ought to just add the `--repair` option and then give ourselves some time to poke at that on test systems?

chu11:
> I actually thought the thing going to be moved to stdout was:
> broker.err[0]: rc1.0: Checking integrity of checkpoint from 2025-10-01T18:42:36
> which isn't an error. Hey that's not even in the output anymore. The date is useful...

Yeah, what had happened was

- "Checking integrity" moved from being output to `stderr` to `stdout`
- We decided to call fsck with `--quiet`, so now  "Checking integrity" is never output
- If no errors are detected, nothing will output with `flux fsck --quiet`
- I wanted something to indicate that `fsck` was run if no errors were detected, so I added the `Running flux-fsck ....`

hmmmm ... 

I almost feel like there should be a `--output-header` (that's a bad name) kind of option to say "always output the "Checking integrity ... " message.  That way there is always atleast one message that indicates fsck was run (I'd remove the `Running flux-fsck` ... message)

Edit: Could use `flux content checkpoint list` ... could output something like `Checking integrity ...` with the checkpoint date.  hmmm

Edit2: Much to my chagrin, the easiest thing would be for `--quiet` not to affect the `Checking integrity ...` message.  It seems sort of wrong ... but another way to look at it is that `--quiet` just trys to reduce the amount of output.  It will always output atleast that core important diagnostic.  Let me try this and see if people find this acceptable.

chu11:
re-pushed, with possibly a disliked change.  `--quiet` is now advertised that it only suppresses "non-core" diagnostic messages.  So `Checking integrity of ...` diag message is always output to stdout.

```
Oct 10 22:31:45 fluxorama systemd[1]: Started Flux message broker.
Oct 10 22:31:45 fluxorama flux[566]: cron.info[0]: synchronizing cron tasks to event heartbeat.pulse
Oct 10 22:31:45 fluxorama flux[566]: broker.info[0]: rc1.0: tab: cron-1 created: scheduled in 16094.430s at Sat Oct 11 03:00:00 2025
Oct 10 22:31:45 fluxorama flux[566]: broker.err[0]: rc1.0: flux-fsck: testdatabad: missing blobref(s)
Oct 10 22:31:45 fluxorama flux[566]: broker.err[0]: rc1.0: flux-fsck: Total errors: 1
Oct 10 22:31:45 fluxorama flux[566]: broker.info[0]: rc1.0: Checking integrity of checkpoint from 2025-10-10T22:31:15
Oct 10 22:31:45 fluxorama flux[566]: broker.err[0]: rc1.0: fsck: bash: exited with exit status 1
Oct 10 22:31:45 fluxorama flux[566]: broker.err[0]: rc1.0: flux modprobe rc1 Exited (rc=1) 0.2s
Oct 10 22:31:45 fluxorama flux[566]: broker.info[0]: rc1-fail: init->shutdown 0.216157s
Oct 10 22:31:45 fluxorama flux[566]: broker.info[0]: children-none: shutdown->finalize 0.022922ms
Oct 10 22:31:45 fluxorama flux[566]: broker.info[0]: rc3.0: flux modprobe rc3 Exited (rc=0) 0.1s
Oct 10 22:31:45 fluxorama flux[566]: broker.info[0]: rc3-success: finalize->goodbye 0.149826s
Oct 10 22:31:45 fluxorama flux[566]: broker.info[0]: goodbye: goodbye->exit 0.078743ms
Oct 10 22:31:45 fluxorama systemd[1]: flux.service: Succeeded.
```

Ehhh?

chu11:
> Also, since we're at the beginning of a release cycle, I wonder if we ought to just add the --repair option and then give ourselves some time to poke at that on test systems?

Was thinking about this comment from earlier.  If we did add `--repair`, this PR may go in a slightly different direction.

A) `flux-fsck` returns an error when errors are found, the task errors, and rc1 fails, thus the broker never starts.

Should behavior be different if `--repair` is used?  Technically the KVS is in decent shape.  I wonder if `--repair` should be considered later when some special case keys are handled.  #7122  i.e. if there is corruption in "critical" paths, definitely error out, but don't have to with non-critical paths, like job data.

B) we had discussions before if `--repair` should be something that requires user to input "yes, repair it" before doing it.

garlick:
Maybe we could start with `--repair` always, no prompting, and then add on later some way to involve a sys admin in a decision to proceed or revert to a dump or checkpoint.

chu11:
re-pushed, adding `--repair` on fsck.  The basics of the rc1 change stay the same, if there is atleast one error, rc1 will still fail and the broker will still fail to start.  The idea being that an admin should take a look at what's going on before restarting again.

src/cmd/builtin/fsck.c
```diff
@@ -711,6 +711,8 @@ static int cmd_fsck (optparse_t *p, int ac, char *av[])
         json_t *checkpt;
         double timestamp;
```
garlick, line 712: Commit message: "when using flux-fsck is used" => "when flux-fsck is used"

doc/man1/flux-fsck.rst
```diff
@@ -32,7 +32,8 @@ OPTIONS
 
 .. option:: -q, --quiet
 
-   Don't output diagnostic messages and discovered errors.
+   Don't output non-core diagnostic messages.  Can significantly reduce
+   output if output will be logged.
```
garlick, line 36: Or maybe just say
> Reduce output to essential information

src/cmd/builtin/fsck.c
```diff
@@ -781,7 +781,7 @@ static struct optparse_option fsck_opts[] = {
       .usage = "List keys as they are being validated",
     },
     { .name = "quiet", .key = 'q', .has_arg = 0,
-      .usage = "Don't output diagnostic messages and discovered errors",
+      .usage = "Don't output non-core diagnostic messages.",
```
garlick: Ditto.  "non-core" is kinda cryptic.  Does it refer to flux-core, core dumps, core memory,...?

chu11:
@garlick thanks, fixed up those small things.  Will set MWP.

PR #7086, opened 2025-09-26
libsubprocess: allow all flags to work remotely

### LINKED ISSUE #5968: libsubprocess: pass flags to remote subprocess
It appears that subprocess flags cannot be passed to remote sub processes.

Originally only the flags `FLUX_SUBPROCESS_FLAGS_STDIO_FALLTHROUGH` and `FLUX_SUBPROCESS_FLAGS_SETPGRP` were supported.  The former flag has no purpose with remote subprocesses and it appears the latter is hard coded for the server.

```
    if (!(p = flux_local_exec_ex (flux_get_reactor (s->h),
                                  FLUX_SUBPROCESS_FLAGS_SETPGRP,
                                  cmd,
                                  &ops,
                                  NULL,
                                  s->llog,
                                  s->llog_data))) {
        errprintf (&error, "error launching process: %s", strerror (errno));
        errmsg = error.text;
        goto error;
    }
```

B/c of this, I'm guessing passing flags along didn't matter much, so it wasn't supported or was just outright missed.

With the addition of the `FLUX_SUBPROCESS_FLAGS_FORK_EXEC` flag (force `fork()` even if `posix_spawn()` is available), this is something we may want to pass along some day.  As well as future flags.

This doesn't appear to be a problem at the moment, but its something to eventually support.

garlick:
Problem: `flux_rexec()` doesn't allow some flags to be set that were thought to only be useful for local subprocesses, but this is unnecessarily limiting.

For example, since SIGCHLD can only be handled in one thread of a multi-threaded process, local subprocesses must be launched from one thread.  In that context, another thread might need to use the remote subprocess interface to launch "local" processes.

Allow all flags to pass the validity check in `flux_rexec()`. Pass them to the server via a new `lflags` key (RFC 42 update required).

Call `flux_subprocess_kill()` from the server kill handler instead of hardwiring `killpg(2)` for remote subprocesses.
Return better human-readable errors when remote kill fails.

Update unit tests.

N.B. issue #5968 was probably closed prematurely, when the SETPGRP flag was inverted.  This completes the fix to the described problem.

Fixes #5968

### DIFF
--- a/src/common/libsubprocess/client.c
+++ b/src/common/libsubprocess/client.c
@@ -107,7 +107,8 @@ flux_future_t *subprocess_rexec (flux_t *h,
                                  const char *service_name,
                                  uint32_t rank,
                                  flux_cmd_t *cmd,
-                                 int flags)
+                                 int flags,
+                                 int lflags)
 {
     flux_future_t *f = NULL;
     struct rexec_ctx *ctx;
@@ -125,9 +126,10 @@ flux_future_t *subprocess_rexec (flux_t *h,
                              topic,
                              rank,
                              FLUX_RPC_STREAMING,
-                             "{s:O s:i}",
+                             "{s:O s:i s:i}",
                              "cmd", ctx->cmd,
-                             "flags", ctx->flags))
+                             "flags", ctx->flags,
+                             "lflags", lflags))
         || flux_future_aux_set (f,
                                 "flux::rexec",
                                 ctx,
--- a/src/common/libsubprocess/server.c
+++ b/src/common/libsubprocess/server.c
@@ -351,12 +351,14 @@ static void server_exec_cb (flux_t *h,
     const char *errmsg = NULL;
     flux_error_t error;
     int flags;
+    int lflags = 0;
 
     if (flux_request_unpack (msg,
                              NULL,
-                             "{s:o s:i}",
+                             "{s:o s:i s?i}",
                              "cmd", &cmd_obj,
-                             "flags", &flags) < 0)
+                             "flags", &flags,
+                             "lflags", &lflags) < 0)
         goto error;
     if (s->shutdown) {
         errmsg = "subprocess server is shutting down";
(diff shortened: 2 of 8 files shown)


src/common/libsubprocess/client.c
```diff
@@ -125,9 +126,10 @@ flux_future_t *subprocess_rexec (flux_t *h,
                              topic,
                              rank,
                              FLUX_RPC_STREAMING,
-                             "{s:O s:i}",
+                             "{s:O s:i s:i}",
                              "cmd", ctx->cmd,
-                             "flags", ctx->flags))
+                             "flags", ctx->flags,
+                             "lflags", lflags))
```
grondo, line 132: Maybe the new payload member should be `local_flags` just to make it more obvious what kind of flags they are. Obviously, I don't feel strongly about it though.

src/common/libsubprocess/server.c
```diff
@@ -543,15 +547,24 @@ static void server_kill_cb (flux_t *h,
...
+        errprintf (&error, "pid %d does not belong to any subprocess", pid);
+        errmsg = error.text;
+        goto error;
+    }
+    if (!(f = flux_subprocess_kill (p, signum))
+        || flux_future_get (f, NULL) < 0) { // will never block
+        errprintf (&error, "%s", future_strerror (f, errno));
+        errmsg = error.text;
         goto error;
+    }
```
grondo, line 560: Does this change really belong in a commit "allow all flags to work remotely"? I may be missing something obvious - I apologize if so.

grondo (approved):
LGTM! I didn't quite get why the kill handler changes went along with this commit (is that just general cleanup, or did the flag changes somehow allow that improvement?)

src/common/libsubprocess/server.c
garlick: Could definitely be split out.  I'll do that.

garlick:
OMG, I hit the merge button.

PR #7087, opened 2025-09-26
libsubprocess: rename lflags to local_flags

garlick:
Problem: the 'lflags' key in the subprocess exec request is poorly named.

Rename to 'local_flags'.


grondo (approved):
:+1:

PR #7088, opened 2025-09-26
doc: add manpage for flux_msg_handler_allow_rolemask(3) and flux_msg_handler_deny_rolemask(3)

chu11:
I noticed these manpages were not documented when i was simply trying to recall what the default rolemask was.

So I attempted to document these, document what the rolemasks are, and add pointers from other manpages to this.

Perhaps more could be added, but figured its a good starting point.

### DIFF
--- a/doc/man3/flux_msg_handler_addvec.rst
+++ b/doc/man3/flux_msg_handler_addvec.rst
@@ -39,9 +39,12 @@ The last entry in the array is set to NULL.
 returned from :func:`flux_msg_handler_addvec`.
 
 These functions are convenience functions which call
-:man3:`flux_msg_handler_create`, :man3:`flux_msg_handler_start`; and
-:man3:`flux_msg_handler_stop`, :man3:`flux_msg_handler_destroy` on each element
-of the array, respectively.
+:man3:`flux_msg_handler_create`, :man3:`flux_msg_handler_start`,
+:man3:`flux_msg_handler_allow_rolemask`; and :man3:`flux_msg_handler_stop`,
+:man3:`flux_msg_handler_destroy` on each element of the array, respectively.
+The defaults of :var:`typemask` and :var:`topic_glob` are described in
+:man3:`flux_msg_cmp`.  The default rolemask is described in
+:man3:`flux_msg_handler_allow_rolemask` and is set with a :var:`rolemask` of 0.
 
 If :func:`flux_msg_handler_addvec` encounters an error creating a message
 handler, all previously created message handlers in the array are destroyed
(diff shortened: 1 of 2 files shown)


garlick:
It might be worth mentioning that the default is indicated by a value of zero (0) in both man pages.

chu11:
> It might be worth mentioning that the default is indicated by a value of zero (0) in both man pages.

Hmmm, should probably only be for `flux_msg_handler_addvec()` b/c 0 has no meaning in the other ones.  (i.e. rolemask of 0 for allow/deny does nothing.  there is no rolemask input for `flux_msg_handler_create()`.)

Re-pushed, doing a few minor tweaks to try and clarify some things and mention the default rolemask of 0 is FLUX_ROLE_OWNER.

doc/man3/flux_msg_handler_allow_rolemask.rst
```diff
@@ -0,0 +1,52 @@
...
+:func:`flux_msg_handler_allow_rolemask` and
+:func:`flux_msg_handler_deny_rolemask` can be used to alter the
+rolemask for each message handler.
+
+
+RESOURCES
+=========
+
+.. include:: common/resources.rst
+
```
garlick, line 55: Suggestion: add a reference to [RFC 12](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_12.html)

We've added a new section in other man pages, e.g. see `flux_kvs_lookup.rst` for an example.

doc/man3/flux_msg_handler_allow_rolemask.rst
```diff
@@ -0,0 +1,52 @@
...
+   void flux_msg_handler_deny_rolemask (flux_msg_handler_t *mh,
+                                        uint32_t rolemask);
+
+Link with :command:`-lflux-core`.
+
+DESCRIPTION
+===========
+
+All message handlers have a rolemask, which indicates what messages can be
+matched.  The currently supported roles are:
```
garlick: Well, it might be useful to explain a bit more what this does.  Maybe

> Flux's role based access control mechanism is described in RFC 12.  The  message handler rolemask determines which messages are delivered to the handler.  Requests that would otherwise match but cannot be delivered due to a missing role are automatically sent a "permission denied" response.  Other message types that don't match are silently discarded.  The instance owner is implicitly authorized for every service, so it is meaningless to add FLUX_ROLE_OWNER to the role mask.

chu11:
re-pushed with tweaks per comments above, thanks!  I'll set MWP if there's no complaints.

chu11:
oops, had a distclean issue.  re-pushed and hopefully fixed now, setting MWP.

PR #7090, opened 2025-09-27
adjust subprocess server protocol to conform to RFC 42

garlick:
Problem: RFC 42 states that the 'msgchans' key in the exec request payload is an array of objects, but it is currently implemented as an array of arrays.

Update implementation to match the RFC.

### DIFF
--- a/src/common/libsubprocess/command.c
+++ b/src/common/libsubprocess/command.c
@@ -448,7 +448,7 @@ static zlist_t *msgchans_fromjson (json_t *o)
             const char *uri;
             struct cmd_msgchan *cm;
 
-            if (json_unpack (entry, "[ss]", &name, &uri) < 0)
+            if (json_unpack (entry, "{s:s s:s}", "name", &name, "uri", &uri) < 0)
                 goto inval;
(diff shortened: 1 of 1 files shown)


garlick:
Strange, I keep hitting this error in the bookworm builder.  It couldn't possibly have anything to do this with this PR, so retrying
```
 expecting success: 
  	run_timeout --env=SHELL=/bin/sh 120 \
  	    $runpty -i none flux start \
  	        -Sbroker.rc1_path= \
  	        -Sbroker.rc3_path= \
  	    	--test-size=1 \
  	        --recovery=$(pwd)/test1 >banner.out &&
  	grep "Entering Flux recovery mode" banner.out
  
  flux timed out after 120.0s
  not ok 4 - --recovery is not ignored if --test-size is specified
```

grondo:
Maybe a new race condition introduced in another PR? Maybe I can try reproducing locally.

garlick:
Yeah maybe. I did try to run the test a few times in the docker bookworm image but it always succeeded for me.

garlick:
@mergifyio requeue

PR #7098, opened 2025-10-01
broker: disable duplicate buffering on rc subprocs

garlick:
Problem: now that the broker rc scripts are run as a "remote" sub-processes (in another thread), line buffering is handled remotely and local buffer management, with its large buffer and multiple watchers, is not needed locally.

Set the FLUX_SUBPROCESS_FLAGS_LOCAL_UNBUF flag on the runat remote sub-processes and change the stdio callback to use `flux_subprocess_read()` instead of `flux_subprocess_getline()`, as required by that flag.

### DIFF
--- a/src/broker/runat.c
+++ b/src/broker/runat.c
@@ -22,6 +22,7 @@
 #include <termios.h>
 #include <unistd.h>
 #include <signal.h>
+#include <string.h>
 #ifdef HAVE_ARGZ_ADD
 #include <argz.h>
 #else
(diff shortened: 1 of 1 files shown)


src/broker/runat.c
```diff
@@ -262,19 +262,29 @@ static void state_change_cb (flux_subprocess_t *p,
...
+    len = flux_subprocess_read (p, stream, &data);
+    while (len > 0) {
+        int n;
+        int skip = 0;
+        for (n = 0; n < len; n++) {
+            if (data[n] == '\n') {
+                skip = 1;
+                break;
+            }
+        }
```
grondo: Probably not a big deal, and perhaps this should be obvious, but why not use `strchr(3)` here? (`strchr` implementation may be optimized)

garlick: The subprocess.h header comment for `flux_subprocess_read()` says
> Buffer is guaranteed to be NUL terminated unless the FLUX_SUBPROCESS_FLAGS_LOCAL_UNBUF flag has been specified.

So the character search needs to be bounded by the buffer length.

grondo: A pretty good idea for a comment there!

grondo: oh, there is also [memchr(3)](https://man7.org/linux/man-pages/man3/memchr.3.html).

garlick: Oh, nice!  I'll use that, and add a comment.  Thanks!

garlick: Ok, pushed a change.  That still look OK?

grondo: Yes!

PR #7099, opened 2025-10-01
broker: add default config-reload module method

garlick:
Problem: broker modules that access the instance configuration with `flux_conf_unpack(flux_get_config (h))` see a snapshot frozen at module load time unless they implement a config-reload RPC handler that updates the cached copy.  This may be surprising.

Add a default config-reload handler that simply updates the config cache in a module's flux_t handle when the configuration changes.  Modules that need to validate a section of the config may override it.  Modules that don't can now expect to access the current configuration.
    
Warning: broker modules that call `flux_get_config()` must assume that the returned object will be invalidated after the reactor is re-entered.  All broker modules shipped with flux-core are confirmed to handle the config object safely.  I _think_ fluxion is OK but we should get someone on the sched team to confirm.  The config parsing code in fluxion is kind of hard to follow.

Some minor cleanup discovered while auditing modules is tacked on.

### DIFF
--- a/src/modules/job-exec/exec.c
+++ b/src/modules/job-exec/exec.c
@@ -569,11 +569,10 @@ static int exec_init (struct jobinfo *job)
     struct exec_ctx *ctx = NULL;
     struct bulk_exec *exec = NULL;
     const struct idset *ranks = NULL;
-    const char *imp_path = NULL;
     const char *service;
     flux_error_t error;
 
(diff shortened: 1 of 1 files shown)


grondo (approved):
LGTM!

garlick:
I think this is good to go now so I'll set MWP.

PR #7107, opened 2025-10-03
add `flux_rexec_bg(3)` and `flux exec --bg`

grondo:
This PR adds support for the `SUBPROCESS_REXEC_BACKGROUND` flag proposed in https://github.com/flux-framework/rfc/pull/470. The flag is activated via a new `flux_rexec_bg(3)` API call, which simply returns a `flux_future_t *` instead of a `flux_subprocess_t *`. Once the future is fulfilled, the subprocess has either been launched in the background, or if fulfilled with error, has failed to executed. Note that the RPC is still streaming, so a successful response is always followed by an `ENODATA` response.

Using this new interface, a `--bg` flag is added to `flux exec`. With this flag, `flux exec` exits after all requested processes have started while the processes continue to run as child of the subprocess server until they exit or the server is stopped.

This is a WIP because I'm not sure if addition of a new API call is the best approach. The first attempt added a new flag to `flux_rexec(3)`. This allowed reuse of more code, especially in `flux-rexec`, but added yet another special "rexec only" flag which has to be cleared from `local_flags`, plus slightly awkward handling of `flux_subprocess_t *` which will never call some of the ops callbacks.

### DIFF
--- a/src/common/libsubprocess/client.c
+++ b/src/common/libsubprocess/client.c
@@ -146,6 +146,45 @@ error:
     return NULL;
 }
 
+flux_future_t *subprocess_rexec_bg (flux_t *h,
+                                    const char *service_name,
+                                    uint32_t rank,
+                                    const flux_cmd_t *cmd,
+                                    int local_flags)
+{
+    flux_future_t *f = NULL;
+    json_t *ocmd = NULL;
+    char *topic = NULL;
+
+    /* FLUX_SUBPROCESS_FLAGS_LOCAL_UNBUF is not allowed with background
+     * execution, raise error if set:
+     */
+    if (local_flags & FLUX_SUBPROCESS_FLAGS_LOCAL_UNBUF) {
+        errno = EINVAL;
+        return NULL;
+    }
+
+    if (service_name == NULL)
+        service_name = "rexec";
+
+    if (asprintf (&topic, "%s.exec", service_name) < 0
+        || !(ocmd = cmd_tojson (cmd)))
+        goto out;
+
+    f = flux_rpc_pack (h,
+                       topic,
+                       rank,
+                       0,
+                       "{s:O s:i s:i}",
+                       "cmd", ocmd,
+                       "flags", 0,
+                       "local_flags", local_flags);
+out:
+    ERRNO_SAFE_WRAP (free, topic);
+    ERRNO_SAFE_WRAP (json_decref, ocmd);
+    return f;
+}
+
 int subprocess_rexec_get (flux_future_t *f)
 {
     struct rexec_ctx *ctx;
--- a/src/common/libsubprocess/server.c
+++ b/src/common/libsubprocess/server.c
@@ -21,6 +21,7 @@
 #include "src/common/libutil/errno_safe.h"
 #include "src/common/libutil/errprintf.h"
 #include "src/common/libutil/llog.h"
+#include "src/common/libutil/basename.h"
 #include "src/common/libioencode/ioencode.h"
 #include "ccan/str/str.h"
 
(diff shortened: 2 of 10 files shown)


src/common/libsubprocess/client.c
```diff
@@ -146,6 +147,45 @@ flux_future_t *subprocess_rexec (flux_t *h,
...
+{
+    flux_future_t *f = NULL;
+    json_t *ocmd = NULL;
+    char *topic = NULL;
+
+    if (service_name == NULL)
+        service_name = "rexec";
+
+    if (asprintf (&topic, "%s.exec", service_name) < 0) {
+        errno = ENOMEM;
```
garlick: asprintf already sets ENOMEM on error

src/common/libsubprocess/client.c
```diff
@@ -146,6 +147,45 @@ flux_future_t *subprocess_rexec (flux_t *h,
...
+
+    /* Clear LOCAL_UNBUF from local_flags. Not a local flag and doesn't
+     * make sense with SUBPROCESS_EXEC_BACKGROUND.
+     */
+    local_flags &= ~FLUX_SUBPROCESS_FLAGS_LOCAL_UNBUF;
+
+    f = flux_rpc_pack (h,
+                       topic,
+                       rank,
+                       FLUX_RPC_STREAMING,
```
garlick: Should we change the server/protocol to expect FLUX_RPC_STREAMING to not be set with SUBPROCESS_EXEC_BACKGROUND?

It sounds like the user is expected to not treat the RPC as streaming and will never be waiting for the ENODATA, which would not be required to send if the STREAMING flag weren't set in the request.

src/common/libsubprocess/server.c
```diff
@@ -291,15 +311,25 @@ static void proc_output_cb (flux_subprocess_t *p, const char *stream)
...
     }
+    else if (len) {
+        /* background process: log output (ignore EOF) */
 
+        const char *command = basename_simple (flux_cmd_arg (p->cmd, 0));
+        if (streq (stream, "stderr"))
+            llog_error (s, "%s: %s", command, buf);
+        else
+            llog_info (s, "%s: %s", command, buf);
+    }
```
garlick, line 347: What happens if the user sets the STDIO_FALLTHROUGH local_flag?

src/common/libsubprocess/subprocess.h
```diff
@@ -184,6 +184,19 @@ flux_subprocess_t *flux_rexec_ex (flux_t *h,
...
                                   void *log_data);
 
+/* Like flux_rexec(3), but run process in background.
+ *
+ * If service_name is NULL, then default to `rexec`.
+ *
+ * Once the process has been started, the returned future will be fulfilled
+ * with the payload ``{"rank":i, "pid":i}``, or an error response if there
+ * is an error.
+ */
```
garlick, line 194: If we stick with the streaming RPC, it might be wise to mention that the rank/pid response will be immediately followed by ENODATA so people looking at the source don't see the STREAMING flag and wonder if other responses are possible.

garlick (approved):
I did have a couple questions inline just for discussion, but this mainly looks good.

src/common/libsubprocess/client.c
grondo: Thanks, wasn't defined in the manpage so I wasn't sure that was always the case.

src/common/libsubprocess/client.c
grondo: Yeah, I considered this too and figured the question would come up. I wondered if just having a separate RPC instead of a flag would be more appropriate here. Especially with the special-case handling of the other flags (most are ignored), etc.

src/common/libsubprocess/server.c
grondo: The stdout will fall through as documented.

src/common/libsubprocess/subprocess.h
grondo: I'm fine with switching to a non-streaming RPC (in fact I think it is a good idea), if it is ok to have a single RPC that can operate as streaming or non-streaming depending on flags in the request.

garlick: IMHO that should be OK.   In fact I wonder if the SUBPROCESS_REXEC_BACKGROUND flag would even be necessary then, since it seems like the logical way to handle the "non-streaming version" of the RPC.

grondo: Hm, that's a somewhat valid point, but since flags are already being transmitted it does no harm to include the SUBPROCESS_REXEC_BACKGROUND flag, and it _feels_ like explicit is better than implicit behavior here, and coupling the RPC flags to subprocess server behavior seems like it might cause confusion in the future.

This may also be more subtle to document in RFC 42 (where, unfortunately the `background` flag PR is already merged.)

If you feel strongly though, I'm willing to discuss further.

garlick: Oh sorry I didn't see this comment - leaving the flag works for me!

grondo: The more I thought about your idea the more I liked it, though I could still go either way.  To see how it would work, I went ahead and proposed a PR against RFC 42 that implemented your idea. Check it out and see what you think:

 https://github.com/flux-framework/rfc/pull/473

garlick: Yep, better!

grondo:
Ok, updated to match the new RFC. Another minor change is that `flux_rexec_bg(3)` now fails with `EINVAL` if `FLUX_SUBPROCESS_FLAGS_LOCAL_UNBUF` is set in flags, and log messages include the PID as well as process name to differentiate instances of a process with the same name.

t/t0005-exec.t
```diff
@@ -11,6 +11,8 @@ Test exec functionality
 SIZE=4
```
garlick, line 11: commit message: s/flux-rexec/flux-exec/

grondo:
Thanks! I'll set MWP here.

Issue #6967, opened 2025-08-14
idea: isolate  downstream overlay network handling in a separate broker thread


garlick:
Problem:  as noted in #6369, in a large, flat TBON config, some message traffic like the gathering of stdio at shell 0 is routed through the rank 0 broker and may affect service response times.  We've also noticed (anecdotally) some lag in response times when doing a lot of event publishing.

If we could simply move the handling of the "downstream" 0MQ ROUTER  socket to a separate thread (perhaps a builtin-in broker module), then potentially both event multicast and through-routing could be offloaded from the main broker thread, improving responsiveness.

PR #7109, opened 2025-10-03
broker: rename event.publish to overlay.publish

garlick:
Problem: `event.subscribe` and `event.unsubscribe` are handled by the broker, but `event.publish` is handled by the overlay code, which might move to a built-in broker module at some point.

Change the RPC target from `event.publish` to `overlay.publish` so the publish method can move outside of the broker with the overlay code.

Update tests.

This follows on and probably should have been submitted with
- #7100

The idea of moving the overlay code to a built-in module was brought up in
- #6967

### DIFF
--- a/src/broker/overlay.c
+++ b/src/broker/overlay.c
@@ -2714,7 +2714,7 @@ void overlay_destroy (struct overlay *ov)
 static const struct flux_msg_handler_spec htab[] = {
     {
         FLUX_MSGTYPE_REQUEST,
-        "event.publish",
+        "overlay.publish",
         overlay_publish_cb,
         FLUX_ROLE_USER,
     },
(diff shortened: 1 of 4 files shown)


garlick:
Great point!   Well, I suppose requests could be forwarded from the old handler instead of removing it.  I hate to do that if we don't think we need it though.

grondo:
This one seems pretty safe. The main use case is probably running newer or older Flux as jobs under a system instance. Since the single user jobs probably don't have permission to send events, I doubt this is currently in use anywhere (even using the API or Python).

But, something to keep in mind in the future if other RPC topics happen to get renamed.

garlick:
Argreed.  Yeah thanks for bringing that up - I had not thought about it this time.
I'll go ahead and set MWP.

PR #7110, opened 2025-10-04
flux-fsck: misc cleanup

chu11:
before getting into some meatier changes, a PR with just the cleanups (since it's a lot of LOC changes).

### DIFF
--- a/src/cmd/builtin/fsck.c
+++ b/src/cmd/builtin/fsck.c
@@ -360,8 +360,7 @@ static void fsck_valref (struct fsck_ctx *ctx,
 
             unlink_path (ctx, path);
 
-            if (!ctx->quiet)
-                fprintf (stderr, "%s repaired and moved to lost+found\n", path);
+            errmsg (ctx, "%s repaired and moved to lost+found", path);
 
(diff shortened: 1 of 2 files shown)


src/cmd/builtin/fsck.c
```diff
@@ -64,21 +64,21 @@ static void fsck_treeobj (struct fsck_ctx *ctx,
 
 static void valref_validate (struct fsck_valref_data *fvd);
 
-static void read_verror (const char *fmt, va_list ap)
+static void output_verrmsg (const char *fmt, va_list ap)
```
garlick: Suggestion: use a shorter name like `warn()`.  Repeating the long name all over the place gets tiresome.

garlick (approved):
LGTM!   Optional suggestion inline.

src/cmd/builtin/fsck.c
chu11: good idea, i could just go with "errmsg" :-)

chu11:
thanks, tweaked "output_errmsg" to just "errmsg".  setting MWP

PR #7114, opened 2025-10-06
flux-fsck: prefix output when not attached to a tty

chu11:
Per discussion in #7082, minor tweaks for flux-fsck to predominantly support the fact many times fsck will be run out of rc1

### DIFF
--- a/src/cmd/builtin/fsck.c
+++ b/src/cmd/builtin/fsck.c
@@ -709,9 +709,7 @@ static int cmd_fsck (optparse_t *p, int ac, char *av[])
             struct tm tm;
             if (!timestamp_from_double (timestamp, &tm, NULL))
                 strftime (buf, sizeof (buf), "%Y-%m-%dT%T", &tm);
-            fprintf (stderr,
-                     "Checking integrity of checkpoint from %s\n",
-                     buf);
+            printf ("Checking integrity of checkpoint from %s\n", buf);
         }
 
         ctx.sequence = sequence;
--- a/src/cmd/builtin/fsck.c
+++ b/src/cmd/builtin/fsck.c
@@ -42,6 +42,7 @@ struct fsck_ctx {
     bool verbose;
     bool quiet;
     bool repair;
+    bool isatty;
     int errorcount;
 };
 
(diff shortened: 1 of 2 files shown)


src/cmd/builtin/fsck.c
```diff
@@ -728,7 +728,14 @@ static int cmd_fsck (optparse_t *p, int ac, char *av[])
```
garlick, line 728: Commit message:
> This ensures there is a log message when flux-fsck
is run in startup scripts such as rc1.

Except we probably do want to know all the bad key names in that case so `--quiet` would be inappropriate.

I think this change is fine, but the justification is not quite on target :-)

doc/man1/flux-fsck.rst
```diff
@@ -32,7 +32,9 @@ OPTIONS
 
 .. option:: -q, --quiet
 
-   Don't output diagnostic messages and discovered errors.
+   Don't output diagnostic messages and discovered errors.  Note that
+   not all message are suppressed.  If errors are found an error count
+   message will always be output to stderr.
```
garlick: Maybe this should say
> Display nothing if no errors were detected;  otherwise, display a single summary message at the end.

src/cmd/builtin/fsck.c
chu11: I reread the comments from #7082, and i think subtlety was lost.  

Are you suggesting we want `--quiet` to have no output if there are no errors?  But if there are errors, errors hould always be output?

To me that feels different than `--quiet`.

grondo: Usually `--quiet` means to suppress non-essential output (diagnostic and progress messages), but still include errors (and sometimes warnings). 

e.g.
```
$ grep --quiet foo /nosuchfile
grep: /nosuchfile: No such file or directory
```

garlick: I guess the key names are pretty essential, so maybe they should be listed with ``--quiet`

They are definitely essential in rc1.

grondo: Something for the future: would it be helpful if `flux fsck` dumped a report somewhere on failure (e.g. timestamped in `statedir`)?

garlick: That's a really good idea.

chu11: ok, lemme lop off the `--quiet` changes from this PR, keeping the other commits then.  We'll have an alternate PR that will change `--quiet` behavior.  Because we may debate what we consider the "error" messages vs the "diagnostic" messages (most notably with --repair).

chu11:
per conversation above, removed the changes to `--quiet`, as that'll be another PR for another day.  Just the smaller output changes.

If I don't hear from anyone tomorrow morning, will set MWP then.

PR #7116, opened 2025-10-06
doc: fix formatting typo in flux-modprobe(1)

chu11:
Problem: A keyword is not consistently quoted, leading to formatting errors.

Fix the backtick quotations.


grondo (approved):
Good catch!

PR #7106, opened 2025-10-03
modprobe: abort operation on first task failure

grondo:
This PR changes modprobe behavior so it no longer continues running all active tasks after a task failure.
Instead, no more future tasks are scheduled to run, and scheduled but non-running tasks are cancelled.
Tasks that have already started, however, are allowed to finish.

PR #7117, opened 2025-10-06
etc/modprobe: rc3 post-finish-event needs kvs

chu11:
Problem: The rc3 post-finish-event tasks needs to the KVS to be up to work.  If it is not loaded, the task results in an error.

Add a "needs" requirement for the "kvs" to the post-finish-event task.

-----

note: haven't tested in a container yet, but noticed this while working on #7082


grondo (approved):
This seems reasonable. Probably not critical since it is unlikely shutdown without a kvs is going to occur outside of testing, and then the behavior would match the old rc3 behavior (harmless errors during rc3) However, a good fix to have! (there may be other similar fixes needed in the rc3 path)

garlick:
This is probably useful when fsck fails and rc1 aborts before the KVS is loaded, since otherwise there would be additional suspicious errors in the logs to sift through.

chu11:
> Probably not critical since it is unlikely shutdown without a kvs is going to occur outside of testing

Definitely not critical.  It came up while testing the flux fsck in rc1.  With the recent addition of #7106, if an fsck failure occurs, the kvs is never loaded.  Thus when rc3 is run after the rc1 failure ... no kvs

Edit: oh @garlick beat me to it :-)

chu11:
Just did a test in a local container and it behaves as expected.  Will set MWP.

grondo:
> if an fsck failure occurs, the kvs is never loaded.

Oh duh. Yeah that makes sense.

PR #7126, opened 2025-10-07
doc: point users to el8 SRPM packages on github

garlick:
Problem: documentation says that RPM packages are available upon request, but el8 SRPMS are now uploaded regularly.

Update docs.


grondo (approved):
LGTM! Thanks!

PR #7127, opened 2025-10-07
python: add explicit `attr_set()` method to Flux handle class

grondo:
This is a simple PR that adds an explicit `attr_set()` method to the handle class. The method detects non-string arguments and raises more sensible errors than the direct cffi call would, which looks like this:

```
$ flux python -c 'import flux; flux.Flux().attr_set("foo", 1)'
  File "/usr/lib/aarch64-linux-gnu/flux/python3.10/flux/wrapper.py", line 194, in __call__
    result = self.fun(*args)
TypeError: initializer for ctype 'char *' must be a cdata pointer, not int

The above exception was the direct cause of the following exception:

  File "/usr/lib/aarch64-linux-gnu/flux/python3.10/flux/wrapper.py", line 196, in __call__
    raise InvalidArguments(
flux.wrapper.InvalidArguments: 
Invalid arguments passed to wrapped C function:
Name: attr_set
C signature: int(*)(struct flux_handle *, char *, char *)
Arguments: ('foo', 1)
```

As @chu11 suggested, this also makes the method discoverable in the docs.

### DIFF
--- a/src/bindings/python/flux/core/handle.py
+++ b/src/bindings/python/flux/core/handle.py
@@ -284,6 +284,22 @@ class Flux(Wrapper):
     def attr_get(self, attr_name):
         return self.flux_attr_get(attr_name).decode("utf-8")
 
+    def attr_set(self, attr, value):
+        """
+        Set a broker attribute
+        Args:
+            attr (str): attribute to set
+            value (str): attribute value.
+        Raises:
+            ValueError: either ``attr`` or ``value`` are not strings.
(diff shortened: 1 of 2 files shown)


chu11 (approved):
LGTM!  Thanks.  I had this on my TODO for after we re-add support for clearing a broker attribute.  But this is fine and can be updated later (I suspect all we have to do is support that value can also be `None` and it'll probably just work).

grondo:
Ah, sorry. Thought it would be an easy one to get in before the tag.

PR #7129, opened 2025-10-08
testsuite: fix occasional sdmon test failures

### LINKED ISSUE #7113: FAIL: t2412-sdmon.t 5 - module stats units array is empty
this test regularly fails for me when running a larger `make -j<BIG> check`.  Unclear why, no debugging done at the moment.

garlick:
Problem: the sdmon sharness test scripts share a systemd instance, but don't account for interference from each other.
    
Have each load the sdmon module with `usr_glob` set to only match their own test units.
    
Fixes #7113


grondo:
Would it be possible to make the user units unique per test invocation here? Currently two or more copies of this test running at the same time also interfere. While this is much more unlikely, it could happen if the same user is running `make check` from two different builds of flux-core.

garlick:
Thanks - actually maybe I could trivially add the test script pid to the glob and the unit names to address your other concern?

grondo:
Yeah, a PID would be sufficient!

grondo:
Great! I verified that this now works (my method for testing for race conditions in tests):
```console
$ flux submit --watch -o pty --cc=1-16 --progress ./t2412-sdmon.t -d -v --root={{tmpdir}}
[snip]
# passed all 22 test(s)
1..22
PD:0   R:0   CD:16  F:0   │██████████████████████████████████████│100.0% 0:00:22
```

garlick:
Cool!  I'll set MWP.

PR #7131, opened 2025-10-08
initialize locale in built-in flux commands, especially flux-dump and flux-restore

garlick:
Problem: once a key containing UTF-8 characters is put to the KVS, dumps can no longer be created.

Initialize the locale in the `flux(1)` command driver to cover all built-in commands, including `flux-dump(1)` and `flux-restore(1)`.

Also drop a check in `flux kvs dir` that was truncating keys that contain UTF-8 characters.

Finally, enable conditional UTF-8 tests to run if a locale containing the string "utf8" is found, in addition to "UTF-8".   On my development systems (debian-12 and ubuntu 22.04) the only available UTF-8 locales did not match the test criteria.

### DIFF
--- a/t/t1000-kvs.t
+++ b/t/t1000-kvs.t
@@ -98,6 +98,18 @@ test_expect_success 'kvs: unlink works' '
 	  test_must_fail flux kvs get $KEY.object
 '
 
+UTF8_LOCALE=$(locale -a | grep 'UTF-8\|utf8' | head -n1)
+if flux version | grep +ascii-only; then
+        UTF8_LOCALE=""
+fi
+test -n "$UTF8_LOCALE" && test_set_prereq UTF8_LOCALE
+
+test_expect_success UTF8_LOCALE 'flux kvs dir can display a UTF-8 key' '
+        (LC_ALL=${UTF8_LOCALE} flux start sh -c "flux kvs put ƒuzzybunny=42 \
+	 && flux kvs dir | grep bunny")
+'
+test_done
+
 #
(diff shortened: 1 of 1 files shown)


src/cmd/flux-kvs.c
```diff
@@ -662,13 +662,6 @@ kv_printf (const char *key, int maxcol, const char *fmt, ...)
             if (*(p + 1) != '\0')
                 overflow = true;
         }
-        for (p = kv; *p != '\0'; p++) {
-            if (!isprint (*p)) {
-                *p = '\0';
-                overflow = true;
-                break;
-            }
-        }
```
chu11, line 671: perhaps add a test to ensure `flux kvs dir` is now outputting as expected.

garlick: Just pushed one - look OK?

t/t1000-kvs.t
```diff
@@ -98,6 +98,18 @@ test_expect_success 'kvs: unlink works' '
 	  test_must_fail flux kvs get $KEY.object
 '
 
+UTF8_LOCALE=$(locale -a | grep 'UTF-8\|utf8' | head -n1)
```
grondo, line 101: commit message: s/UTF_8/UTF-8/?

src/cmd/flux-kvs.c
chu11: LGTM!

garlick:
OK fixed that typo.  Jeez it seems like it takes both you guys to keep me honest :-)
Setting MWP.

PR #7137, opened 2025-10-09
python: support new `handle.attr_remove()` to remove broker attribute

chu11:
Problem: In commit

1573ea7

the ability to remove broker attributes was removed b/c it was
believed to not be needed.  However a use case has arisen where
it is needed.

Solution:

Revert the following commits:

1573ea7
c1f9a98

They are modified for current formatting / style changes.

Re-add attr.rm RPC.
Support NULL for value in flux_attr_set(3).
Support -e, --expunge flag in "flux setattr".
Re-add documentation of --expunge flag for "flux setattr".
Re-add tests.

-----

Side note: I didn't support attribute removal in `attr_set()` in python, choosing to do that in a follow on PR.  Leaving this PR for just the reverting of old stuff.

Also, we could choose to not support `flux setattr --expunge`.  I really only need the API part of this, but figure might as well re-add it.

### DIFF
--- a/t/python/t0001-handle.py
+++ b/t/python/t0001-handle.py
@@ -105,6 +105,21 @@ class TestHandle(unittest.TestCase):
         with self.assertRaises(ValueError):
             self.f.attr_set(42, "foo")
 
+    def test_attr_remove(self):
+        self.f.attr_set("testrm", "rm")
+        self.assertEqual(self.f.attr_get("testrm"), "rm")
+
+        self.f.attr_remove("testrm")
+
+        with self.assertRaises(OSError):
+            self.f.attr_get("testrm")
+
+        with self.assertRaises(OSError):
+            self.f.attr_remove("local-uri")
+
+        with self.assertRaises(ValueError):
+            self.f.attr_remove(1)
+
     def test_conf_get(self):
         # Works with empty config
         self.assertEqual(self.f.conf_get(), {})


garlick:
Why don't we leave off the `flux setattr --expunge` support.  It seems like we should have a `flux rmattr` command instead, but if we don't have a pressing need, it could be added later when we do.

grondo:
I agree. And to make things consistent have a `attr_remove()` call in Python to avoid confusion?

chu11:
Ok, then lets all put it in one giant PR then.

> I agree. And to make things consistent have a attr_remove() call in Python to avoid confusion?

I like this idea (also need the C `flux_attr_remove()` as well.

grondo:
> also need the C flux_attr_remove() as well.

If there's no current need for `flux_attr_remove()` the Python implementation can just make the RPC call directly. Not sure how @garlick would feel about that, but you don't have to add a C API call just for the Python method. (Looks like only `flux_attr_get()` modifies the internal cache, though I could be missing why remove would need to do so)

chu11:
> If there's no current need for flux_attr_remove() the Python implementation can just make the RPC call directly. Not sure how @garlick would feel about that, but you don't have to add a C API call just for the Python method. (Looks like only flux_attr_get() modifies the internal cache, though I could be missing why remove would need to do so)

Part of the reason I initially re-added `--expunge` is that it re-added the testing coverage for the `attr.rm` RPC target.

So its more about adding that function, so we can add `flux rmattr`, so we can add unit tests ...   I admit I tend to be a little anal retentive on these kids of things.

grondo:
Python should be considered a valid way to test the `attr.rm` RPC, right? No sense adding other code if there's no use case.

garlick:
Yeah I agree with that.  We could do without the C api and use python to test.

chu11:
ok, we'll just do python only!

chu11:
re-pushed, now going with a new PR title of `python: support new handle.attr_remove() to remove broker attribute`, we're only supporting this one new feature.  The previous revert commit does only a small bit now.

grondo:
Minor suggestion: add a test for `handle.attr_remove(not_a_string)`

chu11:
> Minor suggestion: add a test for handle.attr_remove(not_a_string)

Doh!  Re-pushed adding a test.

PR #7139, opened 2025-10-09
modprobe: add `!` prefix for task `needs_attrs` and `needs_config`

grondo:
This PR adds support for prefixing config keys and attributes names with a not operator `!` to reverse the sense of the test: i.e. the task if enabled only if the attribute or config key is _not_ set.

### DIFF
--- a/doc/man1/flux-modprobe.rst
+++ b/doc/man1/flux-modprobe.rst
@@ -377,9 +377,13 @@ following optional arguments:
 
 **needs_attrs**
    (optional, list) A list of broker attributes on which this task depends.
+   If the attribute begins with the character ``!``, then this task will only
+   be enabled if the named attribute is not set.
 
 **needs_config**
-   (optional, list) A list of config keys on which this task depends.
+   (optional, list) A list of config keys on which this task depends. If a
+   key is prefixed with the character ``!``, then this task will only be
+   enabled if that config key is not set.
 
 Example
 


t/t0100-modprobe.t
```diff
@@ -284,6 +364,49 @@ test_expect_success 'modprobe needs_config works' '
...
+	EOF
+	flux modprobe run test${seq}.py >output${seq} &&
+	test_debug "cat output${seq}" &&
+	cat <<-EOF >test${seq}.expected &&
+	first
+	EOF
+	test_cmp test${seq}.expected output${seq}
+'
+seq=$((seq=seq+1))
+# note config value `needed` still set from previous test
```
chu11: This was listed before two different tests, second one i think was probably cut & paste error.

t/t0100-modprobe.t
```diff
@@ -218,6 +218,46 @@ test_expect_success 'modprobe needs_attrs prevents task from running' '
...
+	EOF
+	flux setattr zero 0 &&
+	flux modprobe run test${seq}.py >output${seq} &&
+	test_debug "cat output${seq}" &&
+	cat <<-EOF >test${seq}.expected &&
+	first
+	last
+	EOF
+	test_cmp test${seq}.expected output${seq}
+'
```
chu11, line 259: without context knowledge of prior commit, this test seems out of place (ie not `!` related).  perhaps just a small note in the commit message

grondo:
Thanks! Addressed the two comments above and will set MWP.

PR #7141, opened 2025-10-12
libflux: clean up and document configuration API

garlick:
Problem: the API functions for accessing TOML configuration in `conf.h` are poorly organized and undocumented, yet they are widely applicable across the flux-framework.

**flux_set_conf(), flux_get_conf()**
- Rename `flux_set_conf()` to `flux_set_conf_new()` since it steals the reference on its `conf` parameter
- Move these functions to `handle[.ch]` because they are `flux_t` handle accessors (like `flux_get_reactor()`)
- Change `flux_get_conf()` to return an empty object if unset
- Add man pages

**flux_conf_reload_decode()**
- Rename to `flux_module_config_request_decode()` 
- Relocate to `module[.ch]` as this is a module helper
- Add man pages (also for the other module helpers)

**flux_conf_builtin_get()**
- Move to its own `conf_builtin[.ch]` source file as this is not TOML-config related
- Add man page

**flux_conf_t class**
- Add man pages

Although I think these changes are helpful regardless, this also isolates the `flux_conf_t` class from core-isms should we decide to pluck it out to reduce code duplication, as proposed in
- flux-framework/flux-foundry#18


src/modules/sdexec/sdexec.c
```diff
@@ -1205,21 +1205,23 @@ static void config_reload_cb (flux_t *h,
...
     struct sdexec_ctx *ctx = arg;
-    const flux_conf_t *conf;
+    flux_conf_t *conf;
     flux_error_t error;
     const char *errstr = NULL;
 
-    if (flux_conf_reload_decode (msg, &conf) < 0) {
-        errstr = "Failed to parse config-reload request";
+    if (flux_module_config_request_decode (msg, &conf) < 0) {
+        errstr = "Error unpacking config-reload request";
```
grondo, line 1213: Any reason why sdexec/sdbus do not call `flux_set_conf_new()` like the other modules?

garlick: Nope - good catch, will fix!

garlick:
I fixed that and squashed it in with 'change config-reload decode function'.  Even though that was a preexisting condition, my change had turned it into a memory leak.  Also rebased on current master.

garlick:
I'll go ahead and set MWP here, thanks.

PR #7147, opened 2025-10-14
improve documentation and errors around misquoted `--format=` arguments

grondo:
Shells (especially csh/tcsh) may interpret special characters in the `--format=` option argument provided by `flux jobs` and other Flux utilities. There is an existing warning in `flux-jobs(1)` to quote the format argument, but this has proved insufficient.

Add a much expanded warning as a note at the top of the OUTPUT FORMAT section. Add extra details that may help users searching for answers using search engines or AI.

Also add a note to the exception raised when a named format is not found: `Did you mean {x}` in hopes this hints to the user that there may have been a quoting problem.

### DIFF
--- a/src/bindings/python/flux/util.py
+++ b/src/bindings/python/flux/util.py
@@ -1686,7 +1686,10 @@ class UtilConfig:
         try:
             return formats[format_name]["format"]
         except KeyError:
-            raise ValueError(f"--format: No such format {format_name}")
+            s = re.sub(r"\w+", r"{\g<0>}", format_name)
+            raise ValueError(
+                f"--format: No such format {format_name} (Did you mean '{s}'?)"
+            )
 
     def __getattr__(self, attr):
         return self.config[attr]


doc/man1/flux-jobs.rst
```diff
@@ -274,9 +293,6 @@ actual width of the largest presented queue.
 If a format field is preceded by the string ``?+:``, then the field is
 eliminated if empty, or set the maximum item width.
 
-As a reminder to the reader, some shells will interpret braces
-(``{`` and ``}``) in the format string.  They may need to be quoted.
-
```
chu11, line 279: Did you want to remove the exclamation chunk below too?

```
As a reminder to the reader, some shells will interpret the exclamation
point (``!``) when using a conversion flag.  The exclamation point may
need to be escaped (``\!``).
```

grondo: Yeah, I missed that one. Thanks!

grondo: Fixed that and will set MWP.

grondo:
Removing MWP to add special note about `!` in tcsh

chu11:
still LGTM!

grondo:
Ok, hopefully this is enough to get tcsh users attention. I refrained from adding a suggestion to change your default shell :sweat_smile: Setting MWP now.

PR #7148, opened 2025-10-14
content: fix module option parsing bug and improve tests

garlick:
Problem: the `flush-batch-limit` content module option clobbers the value of `purge-old-entry`
    
Fix cut & paste error.

This speeds up one of the tests in `t0012-content-sqlite.t` where the `purge-old-entry` timeout was intentionally lowered.  Also, fix that test so that it can actually fail if the purge does not occur in the test window.

Alsø  alsø:  move checkpoint tests out of `t0012-content-sqlite.t` (which is quite long) into `t0035-content-sqlite-checkpoint.t`, which appears to exist for that purpose.


t/t0012-content-sqlite.t
```diff
@@ -143,121 +143,6 @@ test_expect_success 'fill the cache with more data for later purging' '
 	${SPAMUTIL} 10000 200 >/dev/null
 '
 
-test_expect_success 'checkpoint-get returns ENOENT if there is no checkpoint' '
-	test_must_fail checkpoint_get 2> get.err &&
-	grep "No such file or directory" get.err
-'
-
```
chu11, line 150: The `t0035-content-sqlite-checkpoint.t` file was made mostly for the checkpoint history stuff.

I didn't move the simple checkpoint tests from `t0012-content-sqlite.t` b/c those tests are mostly mirrored in `t0018-content-files.t`.  So when updating tests, it's common to update both files.

No issues moving the tests over, but for consistency, perhaps we should make an equivalent checkpoint tests file for content-files.  That way when changes to one file are made, we know to make changes in the equivalent other checkpoint test file.

garlick: OK, just pushed that change.

garlick:
Thanks! I'll set MWP.

PR #7149, opened 2025-10-14
configure: use system bash completions directory by default

grondo:
Problem: Commit 24b2589e made bash completions install to the system directory only when `--with-system-bash-completion-dir` was specified, to work around pkg-config's `completionsdir` not respecting alternate prefixes. Since this option is likely never used, completions are always installed to `$sysconfdir` (/etc/bash_completion.d) instead.

Drop the `--with-system-bash-completion-dir` option. Use pkg-config's `completionsdir` with `--define-variable=datadir=${datadir}` set on the command line to handle alternate prefixes correctly. This allows the completions file to be installed in the system path as expected in most cases.


grondo:
Thanks, I'll set MWP

PR #7151, opened 2025-10-15
improve `flux job-validator` handling of invalid input

grondo:
This PR improves handling of `flux job-validator` invalid input, especially when a JSON encoded jobspec is fed directly to the command without the `--jobspec-only` option. This will catch simple errors by users testing validator behavior on the command line.


wihobbs (approved):
Thanks @grondo!

grondo:
Thanks!

PR #7152, opened 2025-10-15
minor broker code cleanup

garlick:
This is some minor broker cleanup peeled off of other work.

### DIFF
--- a/src/broker/overlay.c
+++ b/src/broker/overlay.c
@@ -373,7 +373,7 @@ uint32_t overlay_get_rank (struct overlay *ov)
     return ov->rank;
 }
 
-void overlay_set_rank (struct overlay *ov, uint32_t rank)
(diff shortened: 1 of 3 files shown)


src/broker/rundir.h
```diff
@@ -20,4 +23,6 @@ int rundir_create (attr_t *attrs,
                    const char *tmpdir,
                    flux_error_t *error);
 
+#endif // !BROKER_RUNDIR
+
```
grondo: nit: comment should be `!BROKER_RUNDIR_H`.

Also, for consistency with all other files could use C style comment: `/* !BROKER_RUNDIR_H */`

grondo (approved):
LGTM! One note inline.

garlick:
Thanks!  Fixed, and I'll set MWP.

PR #7156, opened 2025-10-17
libsubprocess: support RFC 42 process labels

grondo:
This PR adds basic support in libsubprocess for process labels as described in RFC 42.

A new `flux exec --label=` option is added to enable process labels, and the subprocess kill RPC in the libsubprocess server is extended to take either a PID or a label. The list RPC now includes any label in the reponse, with an empty label indicating "not set."

Eventually, label support can also be added to the sdexec server implementation. In this case, the label can be derived from the jobid, making it easy for the execution system or job manager to attach to a given set of job shells, prolog, epilog, or housekeeping processes across nodes when this operation is supported.

### DIFF
--- a/doc/man1/flux-exec.rst
+++ b/doc/man1/flux-exec.rst
@@ -106,6 +106,13 @@ OPTIONS
    down. Process output is logged to the server's stdout and stderr. The
    :option:`--label-io` and :option:`--no-input` options are ignored.
 
+.. option:: --label=LABEL
+
+   Set a string label for the remote process. When specified, the label
+   allows the process to be referenced by name in addition to process ID
+   in subsequent operations. The label appears in process list responses
+   and must be unique across all processes managed by the subprocess server.
+
 CAVEATS
 =======
 


src/common/libsubprocess/command.c
```diff
@@ -824,6 +827,32 @@ const char *flux_cmd_getopt (flux_cmd_t *cmd, const char *var)
...
+        errno = EINVAL;
+        return -1;
+    }
+
+    prev = cmd->label;
+    if (label == NULL)
+        cmd->label = NULL;
+    else if (!(cmd->label = strdup (label)))
+        return -1;
+    free (prev);
```
garlick: This doens't restore `cmd->label` on strdup failure.
It might be simpler to do the assignment after the strdup succeeds e.g.
```c
char *next = NULL;
if (label) {
    if (!(next = strdup (labael)))
        return -1;
}
free (cmd->label);
cmd->label = next;
```

src/common/libsubprocess/server.c
```diff
@@ -74,7 +75,12 @@ static void proc_destructor (void **item)
...
     void *handle;
+    const char *label = flux_cmd_get_label (p->cmd);
 
+    if (label && zhashx_insert (s->labels, label, p) < 0) {
+        errno = EEXIST;
+        return -1;
+    }
     if (!(handle = zlistx_add_end (s->subprocesses, p))) {
         errno = ENOMEM;
         return -1;
```
garlick: Need to remove label from hash if zlistx_add_end fails?

Side note: I wonder if the hash is needed?  A given subprocess server likely won't have a gazillion subprocesses, and it won't need to find one by label except when delivering a signal or later, recovering, both exceptional conditions.  :shrug:  Just asking, no strong opinion.

t/scripts/rexec.py
```diff
@@ -20,11 +20,11 @@
 def kill(args):
```
garlick, line 20: General comment - should this become a regular command?    Maybe `flux proc` or something?

It might be nice to do `flux proc list` or `flux proc kill` on a real system from time to time.

src/common/libsubprocess/server.c
grondo: The hash was the simplest way to implement the requirement that no duplicate labels for a given server are allowed. It just seemed like the right tool for the job. Are we particularly worried about memory usage here? If so, we could easily switch to a linear search.

src/common/libsubprocess/command.c
grondo: D'oh! I meant to do it the way you suggested, but undid that when fixing another bug.

src/common/libsubprocess/server.c
garlick: Nope, I was just wondering if the complexity was worth it, but I wasn't thinking about avoiding dups.  All good.

t/scripts/rexec.py
grondo: Yes, that makes sense but seems like a separate PR. It will require a bit of development work on the command itself, a suite of tests, documentation etc.

grondo:
Thanks! I've set MWP here.

Issue #6782, opened 2025-04-25
flux-dump: improve performance w/ asynchronous calls


chu11:
while working on #6589 (which code is initially copied from `flux-dump`) I noticed most of the calls to the content modules are synchronous.  In most cases this is probably ok (I'm not sure we'd want to walk multiple directory paths in parallel, but that could be debated).

But minimally the load of every content blob in a valref could be done asynchronously/in parallel.  For large valrefs, such as stdout/stderr from a job, that synchronous lookup could take a long time.

chu11:
during a chat @garlick noted that this may not necessarily be a good thing.  As we may run `flux-dump` on a live system, large parallel lookups could cause a mini-DoS.

This is in contrast to `flux-fsck(1)`, which could benefit from a slight parallelization.  As we generally will not do that on a live system and since data isn't necessarily needed, the parallelization is less of an issue.

grondo:
I'd argue that the DoS is easily fixed by adding a parameter to disable the parallelism.
In the case of shutting down a large system instance, it would be a very good thing if `flux dump` runtime was improved.

garlick:
Great point!

chu11:
the async of `flux-dump` proved to be trickier than I'd hope.  `flux-dump` often runs with the `content` module loaded, so the async calls can come back out of order (i.e. stuff cached vs not cached).   So the "simple" idea would be to make an array of futures and save em off, but this comment is a bit scary

Not only is there this mildly scary comment

```
    /* N.B. first attempt was to save the futures in an array, but ran into:
     *   flux: ev_epoll.c:134: epoll_modify: Assertion `("libev: I/O watcher
     *    with invalid fd found in epoll_ctl", errno != EBADF && errno != ELOOP
     *    && errno != EINVAL)' failed.
     * while archiving a resource.eventlog with 781 entries.  Instead of
     * retaining the futures for a second pass, just retain references to the
     * content.load response messages.
     */
```

Still quite doable, but just trickier than I'd hoped.

garlick:
I think as far as the N.B. comment above, if we assume it's still an issue, then instead of an array of futures, you'd want an array of response messages that you extract from their futures (then the futures are discarded).

I proposed developing a shared class for wrapping valref traversal of one key in a future.  I think that would simplify both of the utilities and allow some tricky code to be shared.

grondo:
This issue came up again in a meeting recently. The time to shutdown Flux on busy system instances is a pain point for sysadmins.

Issue #7122, opened 2025-10-07
flux-fsck: recognize special paths and handle accordingly


chu11:
> potentially moves whole job directories to lost+found instead of the individual keys

comment from @garlick in #7082

- could potentially be combined with #7121 with a special `--job-aware` option.

another special path might be `resource.eventlog`, which is a bit more critical

hypothetically `flux-fsck` could report different error "levels" based on what is corrupted, i.e. `--repair` might repair jobs and not report an error to user.  But an error in `resource.eventlog`, `flux-fsck` will exit with error.

Issue #7121, opened 2025-10-07
flux-fsck: output job ids in f58 instead of KVS


chu11:
> it might be nice to have a --job-aware option or something that translates the KVS keys in the jobs directory to f58

comment from @garlick in #7082

i.e. if there are errors, admins may care more about job id than the kvs path.

garlick:
A related idea is that if something is lost in a job directory, the whole job directory could be moved to the lost+found, under the presumption that it's better to have no job directory than to have an incomplete one.   Less traveling of unlikely code paths in rank 0 broker.

chu11:
> A related idea is that if something is lost in a job directory, the whole job directory could be moved to the lost+found, under the presumption that it's better to have no job directory than to have an incomplete one. Less traveling of unlikely code paths in rank 0 broker.

Yup, I noted that in #7122.  The issues have the potential to be "merged", but felt that were independent of each at first.

chu11:
Looking through code and design a bit more, changing this issue to be a `--job-aware` option in `flux-fsck`.  It should

- output job id instead of kvs path
- move entire job directory over

they really go hand in hand, #7122 is different

PR #7158, opened 2025-10-17
flux-dump: support --maxreqs option for faster dumping

chu11:
Per discussion in #6782.  This largely mirrors the "parallelization" done for valrefs in `flux-fsck`.  The maximum number of RPCs that can be sent out in parallel is hard coded to 1000.

We have some circumstances (i.e. `flux-dump` on a live broker) that we may intentionally want to be slower to not cause performance issues for the running system.  So we don't want to parallelize all the time.  I elected to support this as an option `--async` (name can be debated) vs the opposite `--no-async`, so that the current behavior could stay the same.

Against a 9000 entry KVS valref array put a few `date` calls around a `flux-dump`.

Before

```
17:58:23.882822734
flux-dump: archived 1 keys
17:58:24.912048310
```
nanoseconds = 1029225576

After

```
18:15:57.390297976
flux-dump: archived 1 keys
18:15:57.616373318
```
nanoseconds = 226075342

So about a 78% improvement!  Against a 4000 entry valref I also did, it was about 67%.

Side note: assuming we like this change, I have not yet updated `rc3` to use this.

### DIFF
--- a/doc/man1/flux-dump.rst
+++ b/doc/man1/flux-dump.rst
@@ -76,6 +76,13 @@ OPTIONS
    store.  This may be slightly faster, depending on how frequently the same
    content blobs are referenced by multiple keys.
 
+.. option:: --maxreqs=N
+
+   Speed up :program:`flux-dump` by increasing the number of
+   concurrent lookup requests that can be made.  Increasing the value
+   of N may impact KVS responsiveness and should be done with care on
+   a live system.  Defaults to 2.
+
 .. option:: --ignore-failed-read
 
    If KVS metadata is encountered that references nonexistent blobrefs
--- a/doc/test/spell.en.pws
+++ b/doc/test/spell.en.pws
@@ -1007,3 +1007,4 @@ cpusets
 sproc
 testsuite
 sharness
+maxreqs


garlick:
Yeah maybe `--fast` would be more intuitive than `--async`, making it more obvious why you'd want to use the option?

I haven't looked at this yet but remember I was asking for code sharing between dump and fsck, for fetching large value arrays under one future..  Does that seem possible now?  I believe before you didn't want to pause for that because of the urgent need for fsck at the time.

grondo:
Just for my own curiosity, when would we not want the "fast" behavior?

Nevermind. After I typed that I remembered that we were worried about impact to running systems when doing nightly dumps.

chu11:
> haven't looked at this yet but remember I was asking for code sharing between dump and fsck, for fetching large value arrays under one future.. Does that seem possible now?

It's doable, but I felt it wasn't going to be much of a win.  fsck & dump have begun diverging, so there'd have to be special handling of all the differences between them.  e.g. validate vs. load, save response messages vs not, continue with all blobrefs on error vs give up after first error (I just realized I did not handle this one correctly on the dump side, re-pushed), etc.

chu11:
re-pushed, converting --async to --fast.

garlick:
Could we just pause for a second and make sure that's not a viable approach?  If we could abstract all that it would make the tools far simpler to work on down the road.
> validate vs. load

Same request payload. As far as an interface goes, it could just be like a VALIDATE_ONLY flag that causes the topic string to change

> save response messages vs not

Easily handled I think?  If flag is VALIDATE_ONLY, then no save, otherwise save in an array I guess?

> continue with all blobrefs on error vs give up after first error

Not sure if that really matters.  Errors are unlikely so if you continued and fetched blobrefs after encountering an error it wouldn't be the end of the world.  But probably also doable with `flux_future_continue_error()`?  (I would need to refresh my memory on composite futures to be sure)

It's possible I'm spouting nonsense, and you've actually been working on the code, so if you're pretty sure that's the case then say so and we can move on :-)

chu11:
> Could we just pause for a second and make sure that's not a viable approach? If we could abstract all that it would make the tools far simpler to work on down the road.

It's certainly viable, I just wasn't convinced that it was a big enough "win" to make it worthwhile.

I'll go ahead and give it a shot.  Devils in the details and we'll see how it pans out.

chu11:
re-pushed, add common code via a library in `src/cmd/util` (new commits are on top of what I did before)

Went through several iterations, one attempt to create a "true library function" in `libkvs` was dropped b/c the code became ridiculous (i.e. way more complex than it should have been for some ~50 lines of common code between `flux-dump` and `flux-fsck`).  So I instead made a generalized "common code function" just for this. 

is it a net win?  I think a bit, although it's not as clean as I would have loved.  I could go either way on it really.

```
/* return -1 to discontinue getting more blobrefs */
typedef int (*get_blobrefs_f)(const flux_msg_t *msg,
                              int index,
                              void *arg);

/* return -1 to discontinue getting more blobrefs */
typedef int (*err_blobrefs_f)(int errnum,
                              int index,
                              void *arg);

/* asynchronously lookup blobrefs
 * - get_cb is optional, only if you want the data
 * - err_cb is required when a blobref lookup error occurs
 */
int valref_blobrefs (flux_t *h,
                     const char *topic,
                     json_t *treeobj,
                     get_blobrefs_f get_cb,
                     err_blobrefs_f err_cb,
                     void *arg);
```

The idea was that `flux-dump` can set both callbacks and `flux-fsck` only will set the error callback, b/c it doesn't care about the data.  Passing the topic string was less than ideal, but felt easiest way to deal with `content.load` vs `content-backing.load` vs `content-backing.validate` between the two tools.  Had to leave more code in `flux-dump` that I would have liked, b/c we still deal with the "serial" path there.

garlick:
What I was hoping for was something like
```c
typedef enum { LOOKUP_VALREF_VALIDATE_ONLY=1 } lookup_valref_flags_t;

// Future is fullfilled after all blobrefs have been fetched/validated
// If any blobref is invalid, future is fulfilled with error
flux_future_t *lookup_valref (flux_t *h, json_t *treeobj, int flags);

// Functions to access individual blobs
// get_blob() blocks until future is fulfilled
int lookup_valref_get_blobcount (flux_future_t *f);
int lookup_valref_get_blob (flux_future_t *f, size_t index, const void **buf, size_t *size);
```

Is that approach where the implementation got absurdly complex?

chu11:
As I was prototyping, it began moving towards a `async_valref_t *async_valref_create(json_t *treeobj, int flags)` kinda style with a context, and I felt it just wasn't worth the energy moving in that direction.  I didn't think of it in your "composite future" style API though.

One example, `flux-fsck` needs all the errors and indexes of blobrefs with errors.  Errors can arrive out of order with `flux-dump` (b/c it can choose to go through `content.load` instead of `content-backing.load`).

So, we need to store all the lookup error indexes and errnos on a list somewhere for later retrieval.  But that just seemed like a lot of effort for the amount of gain. (thus my eventual "err_cb" mechanism instead)

It's worth noting, that I went through the original PR push with the mentality that behavior should not change ("behavior" here can mean error messages, etc.).  If we were ok with behavior changing, then that changes the mental math about how to do this.

Given what I just wrote here, I'd like to propose we drop the shared function for the time being.  I'll write up an issue for this for a later time.  There are `flux-fsck` issues (#7121, #7122) in the pipeline, so we'd perhaps want to see how all that falls out first.

garlick:
> So, we need to store all the lookup error indexes and errnos on a list somewhere for later retrieval. But that just seemed like a lot of effort for the amount of gain. (thus my eventual "err_cb" mechanism instead)

> It's worth noting, that I went through the original PR push with the mentality that behavior should not change

We could probably change that to have the composite future be fulfilled with the first error.  I'm not sure it's important to call out each failed blobref and its index, or even count them.   The value is corrupt is the main thing that needs reporting IMHO.

> I'd like to propose we drop the shared function for the time being

OK

chu11:
re-pushed, removing the "common function" discussed above.  Fixed up a few commit messages I noticed still said `--async`

t/t2807-dump-cmd.t
```diff
@@ -256,6 +256,47 @@ test_expect_success UTF8_LOCALE 'restore UTF-8 dump' '
 	    flux kvs get ƒuzzybunny)
 '
 
+# Cover value with a very large number of appends
+
+test_expect_success LONGTEST 'remove backing file and load content-sqlite' '
+	rm -f content.sqlite &&
+	flux module load content-sqlite
```
garlick: Suggestion: use
```
flux module load content-sqlite truncate
```

t/t2807-dump-cmd.t
```diff
@@ -256,6 +256,47 @@ test_expect_success UTF8_LOCALE 'restore UTF-8 dump' '
...
+test_expect_success LONGTEST 'unload kvs' '
+	flux module remove kvs
+'
+test_expect_success LONGTEST 'dump default=kvs-primary checkpoint' '
+	flux dump --checkpoint bigval.tar
+'
+test_expect_success LONGTEST 'unload content-sqlite, remove backing, reload content-sqlite' '
+	flux module remove content-sqlite &&
+	rm -f content.sqlite &&
+	flux module load content-sqlite
```
garlick: truncate option

t/t2807-dump-cmd.t
```diff
@@ -256,6 +256,47 @@ test_expect_success UTF8_LOCALE 'restore UTF-8 dump' '
 	    flux kvs get ƒuzzybunny)
 '
 
+# Cover value with a very large number of appends
+
+test_expect_success LONGTEST 'remove backing file and load content-sqlite' '
```
garlick: I don't think the LONGTEST tests are likely to be run automatically anywhere (correct me if I'm wrong).
Could LONGTEST be dropped if the number of appends were reduced?

src/cmd/builtin/dump.c
```diff
@@ -172,16 +172,8 @@ static void dump_valref (struct archive *ar,
...
      * so make a first pass, saving the data for writing later.
      */
-    /* N.B. first attempt was to save the futures in an array, but ran into:
-     *   flux: ev_epoll.c:134: epoll_modify: Assertion `("libev: I/O watcher
-     *    with invalid fd found in epoll_ctl", errno != EBADF && errno != ELOOP
-     *    && errno != EINVAL)' failed.
-     * while archiving a resource.eventlog with 781 entries.  Instead of
-     * retaining the futures for a second pass, just retain references to the
-     * content.load response messages.
-     */
```
garlick, line 185: Seems like this comment is still relevant?

src/cmd/builtin/dump.c
```diff
@@ -29,6 +29,24 @@
```
garlick, line 30: Commit message:
> Rate cap the lookups at 1000 RPCs.

This isn't really a rate cap - it's a concurrency cap, right?

src/cmd/builtin/dump.c
```diff
@@ -156,49 +175,141 @@ static void dump_write_data (struct archive *ar, const void *data, int size)
...
+    flux_future_t *f;
+    int *indexp;
+
+    blobref = treeobj_get_blobref (dvd->treeobj, dvd->index);
+
+    if (!(f = content_load_byblobref (dvd->h, blobref, content_flags))
+        || flux_future_then (f, -1, get_blobref_continuation, dvd) < 0) {
+        read_error ("%s: cannot load blobref %d", dvd->path, dvd->index);
+        flux_future_destroy (f);
+    }
```
garlick: This error seems like it could be confusing.  Any failure in this function is just a failure to send the request or organize bookkeeping rather than an actual missing blobref.    Maybe this function shouldn't emit any errors and instead should just return -1 and let the caller report "could not send request for next blobref" or similar.

Edit: in fact it should always be fatal.

src/cmd/builtin/dump.c
```diff
@@ -156,49 +175,141 @@ static void dump_write_data (struct archive *ar, const void *data, int size)
...
+    if (!(indexp = (int *)malloc (sizeof (int))))
+        log_err_exit ("cannot allocate index memory");
+    (*indexp) = dvd->index;
```
garlick: Seems like a nice place to use a ccan `ptrint`.

src/cmd/builtin/dump.c
```diff
@@ -156,49 +175,141 @@ static void dump_write_data (struct archive *ar, const void *data, int size)
...
+}
+
+static int dump_valref_async (struct dump_valref_data *dvd)
+{
+    while (dvd->in_flight < BLOBREF_ASYNC_MAX
+           && dvd->index < dvd->count) {
+        get_blobref (dvd);
+        dvd->in_flight++;
+        dvd->index++;
+    }
```
garlick: Just a thought but maybe instead of `--fast` with a hard coded limit, it would be nicer to have a `--max-concurrent-requests=N` option with a low default?

src/cmd/builtin/dump.c
```diff
@@ -156,49 +175,141 @@ static void dump_write_data (struct archive *ar, const void *data, int size)
...
+    dvd.count = count;
+
+    if (fast) {
+        if (dump_valref_async (&dvd) < 0)
+            goto cleanup;
+    }
+    else {
+        if (dump_valref_serial (&dvd) < 0)
+            goto cleanup;
+    }
```
garlick: See previous sugestion about `--max-concurrent-requests`.  That would allow these two code paths to be collapsed.

src/cmd/builtin/dump.c
```diff
@@ -156,49 +175,141 @@ static void dump_write_data (struct archive *ar, const void *data, int size)
...
+    int count = treeobj_get_count (treeobj);
+    const flux_msg_t **msgs;
+    struct archive_entry *entry;
+    struct dump_valref_data dvd = {0};
+
+    /* We need the total size before we start writing archive data,
+     * so make a first pass, saving the data for writing later.
+     */
+    if (!(msgs = calloc (count, sizeof (msgs[0]))))
+        log_err_exit ("could not create messages array");
```
garlick: This comment is a bit confusing.  Maybe just say something like
> Load all data comprising the valref before starting the archive entry.  This is because the total size of the value must calculated and written before any data, and any load errors must be detected so the whole entry can be skipped if  `--ignore-failed-read`.

doc/man1/flux-dump.rst
```diff
@@ -76,6 +76,10 @@ OPTIONS
    store.  This may be slightly faster, depending on how frequently the same
    content blobs are referenced by multiple keys.
 
+.. option:: --fast
+
+   Speed up :program:`flux-dump` by running some operations asynchronously
+
```
garlick: If I'm a user, I may wonder why I wouldn't always want `--fast`.
Probably should say something like
> When the dump is large, this could create a heavy burden on the KVS and should not be used unless the system is quiescent or it is an emergency.

garlick (approved):
I made a few suggestions inline.  None of them are huge deal so I'll tentatively approve and let you sort through them and take what's useful.

src/cmd/builtin/dump.c
chu11: since we're moving to an array of msgs I thought no longer relevant?  But we could certainly leave it as an explanation for why we use an array of msgs vs futures.

garlick: IMHO array of futures is the more straightforward way, so I thought the comment explaining why array of msgs was informative.

src/cmd/builtin/dump.c
chu11: This was comments from before my work :-)  I'll update.

garlick:
If you are considering my `--max-concurrent-request=N` suggestion, I think we could come up with a shorter name.  LIke  `--maxreqs=N` :shrug:

grondo:
I really like the idea of `--maxreqs=N` FWIW

chu11:
> If you are considering my --max-concurrent-request=N suggestion, I think we could come up with a shorter name. LIke --maxreqs=N 🤷

Are you thinking as a replacement of `--fast` or as a supplement ... i.e.

`flux dump --fast --maxreqs=3000` or just `flux dump --maxreqs=3000` and the default is 1?

I do like the idea of the latter.  My only "hmmmm" thought is that we didn't do that with `flux fsck`.  MIght be good to do that over there too (possibly before this PR, if we do not want `--fast`).

garlick:
I was thinking just the one option.

fsck seems less important since the KVS is normally offline when that runs.

t/t2807-dump-cmd.t
chu11: nope, I just run LONGTEST when I run tests :-)

Hmmmm, I think having "quite large" is important, even if it's only when I run personally.

Perhaps we add something "medium", like 50-100 appends that can run all the time?

chu11:
re-pushed, going with `--maxreqs` instead of `--fast`.  the cleanups are now in #7186

One idea I had ... would we like to keep `--fast`, it can be a shortcut for `--maxreqs=1000`??  ehhhh

chu11:
oh, I know this is approved, but since i did rename the option to `--maxreqs` and it takes an argument, figured the PR should get another once over before I set MWP.

src/cmd/builtin/dump.c
```diff
@@ -183,8 +183,8 @@ static void dump_valref (struct archive *ar,
      * retaining the futures for a second pass, just retain references to the
      * content.load response messages.
      */
-    if (!(l = flux_msglist_create ()))
-        log_err_exit ("could not create message list");
+    if (!(msgs = calloc (count, sizeof (msg))))
+        log_err_exit ("could not create messages array");
```
garlick: shouldn't that be (techinically) `sizeof (msg[0])`?

src/cmd/builtin/dump.c
```diff
@@ -156,18 +174,62 @@ static void dump_write_data (struct archive *ar, const void *data, int size)
...
+                    index,
+                    future_strerror (f, errno));
+        flux_future_destroy (f);
+        dvd->errorcount++;
+        dvd->errnum = errno;    /* we'll report the last errno */
+        return;
+    }
+    dvd->in_flight--;
+    dvd->total_size += len;
+    dvd->msgs[index] = flux_msg_incref (msg);
```
garlick, line 203: It might be a good idea to bounds check `index` before using it here.

src/cmd/builtin/dump.c
```diff
@@ -156,18 +174,62 @@ static void dump_write_data (struct archive *ar, const void *data, int size)
...
+    const char *blobref;
+    flux_future_t *f;
+
+    blobref = treeobj_get_blobref (dvd->treeobj, dvd->index);
+
+    if (!(f = content_load_byblobref (dvd->h, blobref, content_flags))
+        || flux_future_then (f, -1, get_blobref_continuation, dvd) < 0)
+        log_err_exit ("%s: cannot load blobref %d", dvd->path, dvd->index);
+    if (flux_future_aux_set (f, "index", int2ptr (dvd->index), NULL) < 0)
+        log_err_exit ("could not save index value");
```
garlick: This could probably just have one error message instead of two, since it likely won't be helpful to distinguish which one failed.

src/cmd/builtin/dump.c
```diff
@@ -183,33 +245,34 @@ static void dump_valref (struct archive *ar,
...
-        flux_future_destroy (f);
+
+    dvd.h = h;
+    dvd.treeobj = treeobj;
+    dvd.msgs = msgs;
+    dvd.path = path;
+    dvd.count = count;
+
+    while (dvd.in_flight < async_max
+           && dvd.index < dvd.count) {
```
garlick: Style: one 4-space indent before && or don't break the line at all.

src/cmd/builtin/dump.c
```diff
@@ -426,6 +497,10 @@ static int cmd_dump (optparse_t *p, int ac, char *av[])
         content_flags |= CONTENT_FLAG_CACHE_BYPASS;
         kvs_checkpoint_flags |= KVS_CHECKPOINT_FLAG_CACHE_BYPASS;
     }
+    /* default = 1, no parallelism */
+    async_max = optparse_get_int (p, "maxreqs", 1);
+    if (async_max <= 0)
+        log_err_exit ("invalid value for maxreqs");
```
garlick: I wonder if the default should be something like 8?

It feels like a case where a little parallelism is likely to help a lot and yet not be too hard on the system, but I guess we can always raise it later.

src/cmd/builtin/dump.c
```diff
@@ -156,18 +174,62 @@ static void dump_write_data (struct archive *ar, const void *data, int size)
...
+    index = ptr2int (flux_future_aux_get (f, "index"));
+    if (flux_future_get (f, (const void **)&msg) < 0
+        || flux_response_decode_raw (msg, NULL, NULL, &len) < 0) {
+        read_error ("%s: missing blobref %d: %s",
+                    dvd->path,
+                    index,
+                    future_strerror (f, errno));
+        flux_future_destroy (f);
+        dvd->errorcount++;
+        dvd->errnum = errno;    /* we'll report the last errno */
```
grondo, line 194: Should `in_flight` be decremented here as well?

src/cmd/builtin/dump.c
grondo: It might be good to do some experiments and report back here. Might be good to get a reasonable default on the first go.

src/cmd/builtin/dump.c
chu11:
```
+    const flux_msg_t **msgs;
     const flux_msg_t *msg;
```

I went with `msg` b/c it happened to be defined here, but I agree, `msgs[0]` is a bit more clear.

src/cmd/builtin/dump.c
chu11: good catch!  ~~I should do it in flux-fsck too.~~ N/m not in `flux-fsck`, it falls through and doesn't return immediately.

chu11:
> I wonder if the default should be something like 8?

> It feels like a case where a little parallelism is likely to help a lot and yet not be too hard on the system, but I guess we can always raise it later.

I did wonder about having the default be like 4 or 8, but I left it at 1 out of a desire to not change current behavior. 

I guess the question in my mind is what use case is there where we want to have `flux-dump` a little faster by default, but not fast enough to burden a live running system?  My understanding is we have one current use case where we do run `flux-dump` on a live running system.  And we do it at 3am or something for a "nightly just in case backup", so at that time performance probably does not matter?

garlick:
I think the other potential use case is a sys admin taking a backup when something is not working right.  In that case, it's probably better if the tool doesn't take 40m out of their day.

I wouldn't expect someone in the hot seat like that to be reading man pages to discover optimizations either.  They would rightly expect the tool to have a workable default.

chu11:
> I think the other potential use case is a sys admin taking a backup when something is not working right. In that case, it's probably better if the tool doesn't take 40m out of their day.
>
> I wouldn't expect someone in the hot seat like that to be reading man pages to discover optimizations either. They would rightly expect the tool to have a workable default.

That's a good example!  Lemme run some experiments, see if I can come up with a number that seems pretty good.

chu11:
So after some experimentation, I found that performance gains with `--maxreqs` plateaus maybe around 8-16 or so.  I suspect on a real system with mountains of more data than my experiments, perhaps plateau at 32-64 might happen.

With a bit of profiling, I suspect this is b/c `flux-dump` actually post-processes the data it receives (i.e. the real data is sent, then it gets written to archive).  This is in contrast to `flux-fsck` which only validates the blobref is correct (i.e. did validate return error or not) and immediately moves on.

Going from maxreqs of 1 to 2 is a big win, going from 2 to 4 is too, then gains begin to slow down.

Given that, I'm going to go with a default of 2.  It gets us a pretty big win w/ only doubling the requests sent to the broker. I think 4 might be ok as well, but I was intentionally conservative.

chu11:
FYI, was looking for a look over, since I did change this from `--fast` to `--maxreqs`.

doc/man1/flux-dump.rst
```diff
@@ -76,6 +76,14 @@ OPTIONS
    store.  This may be slightly faster, depending on how frequently the same
    content blobs are referenced by multiple keys.
 
+.. option:: --maxreqs=N
+
+   Speed up :program:`flux-dump` by increasing the number of
+   concurrent lookup requests that can be made.  When the dump is
+   large, this could create a heavy burden on the KVS and should not
+   be used unless the system is quiescent or it is an emergency.
+   Defaults to 2.
```
garlick: Reads a little awkward now that it's 2 by default.  Maybe the disclaimer sentence beginning with "When the dump..." could be replaced with

> Increasing the value of N may impact KVS responsiveness and should be done with care on a live system.

chu11:
cool, thanks.  Tweaked the docs, will set MWP.

PR #7160, opened 2025-10-20
broker: clean up broker logging to facilitate future refactoring

garlick:
Problem: broker code uses both the `log_err()`  functions that log directly to stderr, and `flux_log()` functions that use the broker's logging subsystem.  Some of the former calls are necessary because logging isn't initialized until some of them are called.  However, over time it's become less clear which is appropriate.

To make the code base easier to refactor in the future and also make the logging strategy a bit clearer for new code:
- Replace direct logging with a `flux_error_t` OUT parameter where appropriate
- Initialize the logging subsystem much earlier
- Use `flux_log()` almost everywhere over `log_err()`
- Change `--verbose` so that it forces `log-stderr-mode=local` and increases `log-stderr-level`

### DIFF
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -93,23 +93,34 @@ static void signal_cb (flux_reactor_t *r,
                        void *arg);
 static int broker_handle_signals (broker_ctx_t *ctx);
 
-static flux_msg_handler_t **broker_add_services (broker_ctx_t *ctx);
+static flux_msg_handler_t **broker_add_services (broker_ctx_t *ctx,
+                                                 flux_error_t *error);
 
 static void set_proctitle (uint32_t rank);
 
-static int create_runat_phases (broker_ctx_t *ctx);
+static int create_runat_phases (broker_ctx_t *ctx, flux_error_t *error);
 
-static void init_attrs (attr_t *attrs, pid_t pid, struct flux_msg_cred *cred);
+static int init_attrs (attr_t *attrs,
+                       pid_t pid,
+                       struct flux_msg_cred *cred,
+                       flux_error_t *error);
 
-static void init_attrs_post_boot (attr_t *attrs);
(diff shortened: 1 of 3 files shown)


src/common/libflux/flog.h
```diff
@@ -32,6 +32,13 @@ void flux_log_set_appname (flux_t *h, const char *s);
  */
 void flux_log_set_procid (flux_t *h, const char *s);
 
+/* Set log hostname for handle instance.
+ * Value will be truncated after FLUX_MAX_HOSTNAME bytes.
```
grondo: Should this be STDLOG_MAX_HOSTNAME?

src/broker/broker.c
```diff
@@ -271,6 +271,13 @@ int main (int argc, char *argv[])
     }
     flux_watcher_start (ctx.w_internal);
 
+    /* Set early logger configuration so that flux_log(3) et al can
+     * be used before the rank is known, reactor is running, etc..
+     */
+    flux_log_set_redirect (ctx.h, log_early, ctx.attrs);
+    flux_log_set_hostname (ctx.h, "?"); // in lieu of rank
```
grondo: On line 240 above `hostname` is already set, why not use that? (I may be missing some subtlety here though, apologies if so).

src/common/libflux/flog.h
garlick: Ah yeah, although this is an installed header and stdlog is not so probably it should just give the byte count.
I'll fix the instance above this that have the same problem too.

src/broker/broker.c
garlick: The "hostname" (an RFC 5424 field) is set to the rank in the broker, but since the rank isn't known yet, we just use `?`.  I hadn't even thought of using the hostname when the rank isn't known (duh!)

garlick:
Fixed those two things - thanks!  I guess I'll set MWP.

PR #7163, opened 2025-10-21
testsuite: drop maxrss > 0 test

### LINKED ISSUE #7162: testsuite: t0003-module.t 40 failure
I'm getting an isolated failure of `t0003-module.t 40 - flux module stats --rusage --parse maxrss works` on a GH200, running RHEL 9.5.

Every other test in t0003 passes, and this command runs successfully, but the value of `maxrss` is being returned as 0, while the test checks that it is >0.

If I manually start a flux test instance within the build and run the exact same commands I do get a non-zero value, so I'm rather stumped as to why it isn't working when run within the test suite itself.

garlick:
Problem: a test that `flux module stats --rusage` returns maxrss > 0 is failing mysteriously on a particular platform.

Drop the check for maxrss > 0 as this value is passed directly through from getrusage(2) and is only used for debugging.

Fixes #7162

### DIFF
--- a/t/t0003-module.t
+++ b/t/t0003-module.t
@@ -203,8 +203,7 @@ test_expect_success 'flux module stats --rusage=badopt fails' '
 '
 
 test_expect_success 'flux module stats --rusage --parse maxrss works' '
-	RSS=$(flux module stats --rusage --parse maxrss $REALMOD) &&
-	test "$RSS" -gt 0
+	RSS=$(flux module stats --rusage --parse maxrss $REALMOD)
 '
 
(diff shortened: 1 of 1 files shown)


grondo (approved):
Seems reasonable to me! Could also have changed the test to something like `test $RSS -eq $RSS` to ensure the result is a number, but since that's unlikely this seems fine to me!

garlick:
> Could also have changed the test to something like `test $RSS -eq $RSS` to ensure the result is a number, but since that's unlikely this seems fine to me!

Thanks, agreed - pretty unlikely.  I'll set MWP.

PR #7165, opened 2025-10-21
support plugin-specific help via `--help=OPTION` in cli submission commands

grondo:
This PR extends the `--help` option in the cli submission commands (`flux submit`, `run`, `alloc`, `batch`, and `bulksubmit`) to take an optional argument. A new plugin `help()` method is added to the `CLIPlugin` base class which by default just prints the plugin's docstring. However, plugins can override this method if they have a better format for help (e.g. call out to a man page)  or want to tailor help to individual options (the option for which help was requested is passed into the method).

To prep for this work, there's some slight refactoring of the python submission commands to move duplicated code and give a single location where the help option can be processed.

Posting as WIP for now to get feedback before adding tests. Especially from @wihobbs and @jameshcorbett.

### DIFF
--- a/t/t2717-python-cli-plugins.t
+++ b/t/t2717-python-cli-plugins.t
@@ -12,6 +12,26 @@ test_expect_success 'flux-alloc: base --help message includes plugin options' '
 	flux alloc --help \
 		| grep -e "Options provided by plugins:" -e "--amd-gpumode"
 '
+test_expect_success 'flux-alloc: --help=foo fails' '
+	test_must_fail flux alloc --help=foo 2>help-foo.err &&
+	test_debug "cat help-foo.err" &&
+	grep "no such option" help-foo.err
+'
+test_expect_success 'flux-alloc: --help=amd-gpumode works' '
+	flux alloc --help=amd-gpumode >help-gpumode.out &&
+	test_debug "cat help-gpumode.out" &&
+	grep "Documentation for plugin" help-gpumode.out
+'
+test_expect_success 'flux-alloc: --help=--amd-gpumode works' '
+	flux alloc --help=--amd-gpumode >help--gpumode.out &&
+	test_debug "cat help--gpumode.out" &&
+	grep "Documentation for plugin" help--gpumode.out
+'
+test_expect_success 'flux-alloc: --help= works with tty' '
+	runpty.py -o help.json -f asciicast flux alloc --help=amd-gpumode &&
+	test_debug "jq -S <help.json" &&
+	grep "Documentation for plugin" help.json
+'
 test_expect_success 'flux-alloc: a job that does not provide plugins can run' '
 	flux alloc -N1 hostname
 '
@@ -25,13 +45,13 @@ test_expect_success 'flux-alloc: job with invalid key to plugin is rejected' '
 	grep "flux-alloc: ERROR: Invalid option" err.out
 '
 test_expect_success 'flux-alloc: job with preinit has config set accordingly' '
-	flux alloc --amd-gpumode=TPX -N1 flux config get > config.json && 
+	flux alloc --amd-gpumode=TPX -N1 flux config get > config.json &&
 	test $(jq -r .resource.rediscover config.json) = "true"
 '
 test_expect_success 'flux-alloc: multiple valid plugin options are accepted' '
 	flux alloc -N1 --amd-gpumode=TPX --ex-match-policy=firstnodex \
 		flux config get > config2.json &&
-	test $(jq -r .resource.rediscover config2.json) = "true" && 
+	test $(jq -r .resource.rediscover config2.json) = "true" &&
 	test $(jq -r ".[\"sched-fluxion-resource\"][\"match-policy\"] == \"firstnodex\"" config2.json) = "true"
 '
 test_expect_success 'flux-alloc: plugin without a prefix behaves appropriately' '


wihobbs:
This looks like a solid approach, and the first commit is a nice catch for refactoring the base CLI code. Thanks @grondo!

Two high level comments: 1) I don't _often_ see `sys.exit(0)` and while correctly applied here, it did make me think harder ("could anything else hit this path and accidentally cause the plugin/command to exit early?") and 2) is that perhaps we should allow plugins to include in their metadata a `long_help` string instead of using the docstrings for this. We talked about this when we did the first CLI plugin PR. Thoughts?

grondo:
>  1) I don't often see sys.exit(0) and while correctly applied here, it did make me think harder ("could anything else hit this path and accidentally cause the plugin/command to exit early?") 

Good question. This is just mimicking what the argparse builtin `--help` handler would do, which is print the help and exit 0. I can't see how it would cause the command to exit early unless one of the help options was used (which matches previous behavior)

>  2) is that perhaps we should allow plugins to include in their metadata a long_help string instead of using the docstrings for this. We talked about this when we did the first CLI plugin PR. Thoughts?

I'm not sure the best approach here. Note that use of the docstring is just a default, and plugins that want to use the docstring for something else or for some reason stash the long help in a variable can do that and implement a `help()` method.

Docstrings seem natural for this since they are meant for documentation. The other thing I like about this approach is that developers viewing the source of a plugin will always know where to look for the documentation.

For projects that autodoc Python using sphinx to a readthedocs or similar site, this also makes the documentation available on the web.

You didn't provide any argument for using a `long_help` variable or against a docstring, though, so perhaps I'm missing something obvious.

wihobbs:
> You didn't provide any argument for using a long_help variable or against a docstring, though, so perhaps I'm missing something obvious.

The only thing I can think of is that the formatting would look weird if people used sphinx autodoc key words like `:param:` in their docstring, and then it got dumped in the terminal. I doubt that would be the case, since we used an abstract base class for plugins, so documenting parameters in a plugin itself wouldn't be necessary, but maybe there are others that would show up.

That's not an argument I initially thought of, though, and since you point out plugins can implement a `help()` method optionally, I'm on board. Sorry I missed that!

src/bindings/python/flux/cli/plugin.py
```diff
@@ -129,6 +140,21 @@ def __init__(self, prog):
...
+        """
+        Print docstring for plugin associated with option ``name`` and exit
+
+        Raises:
+            ValueError: No help found for ``name``
+        """
+        for plugin in self.plugins:
+            for option in plugin.options:
+                # match name to option with or without leading `--`:
+                if name in (option.name, option.lstrip("-")):
```
jameshcorbett: Should be `option.name.lstrip("-")` I think?

jameshcorbett (commented):
I was worried the implementation would be messy but I think this looks great!

src/bindings/python/flux/cli/plugin.py
grondo: Yes, good catch! I must not have even run a sanity check on this version 🤦

grondo:
Pushed an update with some tests. Also switched use of `self.__doc__` to `inspect.getdoc(self)`, and fall back to `sys.stdout.write()` on `ttypager()` failures.

src/bindings/python/flux/cli/plugin.py
```diff
@@ -65,8 +67,18 @@ def help(self, option):
...
+                f"No extra documentation for option `{option.name}` found.",
+                file=sys.stderr,
+            )
+            return
         print(f"\nDocumentation for plugin providing option `{option.name}`:\n")
-        ttypager(self.__doc__)
+        try:
+            ttypager(docstring)
+        except termios.error:
+            sys.stdout.write(docstring + "\n")
```
jameshcorbett, line 81: These changes are in the wrong commit?

grondo: Yeah, copy-and-pasted the wrong commit with `--fixup`. Should be fixed now.

grondo: Fixed that and also inconsistent use of `parser` vs `self.parser` in `MiniCmd.run_command()`

grondo:
Removed WIP here since there's tests now and the approach hasn't seemed to raise any objections.

src/bindings/python/flux/cli/base.py
```diff
@@ -766,6 +766,24 @@ def __init__(self, prog, usage=None, description=None, exclude_io=False):
...
+    def run_command(self):
+        """
+        Convenience method containing common code to run a command derived
+        from the MiniCmd base class.
+
+        This function reopens standard output and error in a mode suitable
+        for utf8 output, initializes the command's ArgParse parser, parses
+        command line arguments and finally executes ``self.main(args)``.
+        """
+        sys.stdout = open(
```
wihobbs, line 775: A suggestion...
```
    try:
        sys.stdout = open(
            sys.stdout.fileno(), "w", encoding="utf8", errors="surrogateescape"
        )
        sys.stderr = open(
            sys.stderr.fileno(), "w", encoding="utf8", errors="surrogateescape"
        )
        args = self.get_parser().parse_args()
        self.main(args)
    except Exception:
        # Ensure we flush any buffered output before propagating
        sys.stdout.flush()
        sys.stderr.flush()
        raise
```

Are we particularly concerned about flushing stdout/stderr in the event of an exception?

t/t2717-python-cli-plugins.t
```diff
@@ -9,7 +9,7 @@ export FLUX_CLI_PLUGINPATH=${SHARNESS_TEST_SRCDIR}/cli-plugins/
 test_under_flux 1 job
 
 test_expect_success 'flux-alloc: base --help message includes plugin options' '
-	flux run --help \
+	flux alloc --help \
```
wihobbs, line 12: Cleaning up the messes of those pesky junior developers! 🛠️ 😆

src/bindings/python/flux/cli/base.py
grondo: I'm not sure this would do anything. We already have the CLIMain decorator that catches all exceptions and stdio will be flushed at exit. Were you worried about something in particular?

wihobbs: Nope, if the decorator catches it that sounds good to me.

grondo:
I'm looking into why the 0.26% coverage drop on this PR. I'll set MWP after verifying somethings not got wrong in ci.

grondo:
I've set MWP. Hopefully the coverage drop is just from consolidating 50 or so lines of covered, duplicated code.

PR #7166, opened 2025-10-22
librlist: enable NVML backend for rhwloc

sam-maloney:
Problem: NVIDIA GPUs can be found by both the CUDA or NVML Backends but only CUDA is currently checked

Add a first check to see if any CUDA osdevs are present, otherwise check for NVML. The separate checks prevent double-counting if both Backends are present.

If someone knows `hwloc` better than me and can propose a cleaner solution I would definitely be happy to hear it!


grondo:
This seems pretty safe so I've set MWP. Thanks again @sam-maloney!

PR #7170, opened 2025-10-25
libsubprocess: add RFC 42 waitable process support

grondo:
This PR implements the waitable subprocess support recently added to RFC 42.

A new public `FLUX_SUBPROCESS_FLAGS_WAITABLE` flag is added to the libsubprocess API and a `flux exec --waitable` flag is added which sets the new flag.

Processes with the waitable flag are kept as zombies on the server's processes list until a new `wait` RPC is issued or the server shuts down (whereupon they are discarded). The status of a waitable process can only be reaped once (and the `state=finished` response in the streaming RPC is considered a successful reaping), and there can only be a single waiter for a given process.

One potential questionable design decision that needs extra scrutiny:

RFC 42 defines `FLUX_SUBPROCESS_FLAGS_WAITABLE` in the `exec` request `flags`, but like `FLUX_SUBPROCESS_FLAGS_LOCAL_UNBUF`, the flag needs to be exported in the public API and passed in the `flux_rexec(3)` `flags` parameter. Therefore, it is removed from `local_flags` and a new `SUBPROCESS_REXEC_WAITABLE` flag (with the same value) is added to `flags` in the RPC. To avoid an extra member in the subprocess struct, the server reuses `FLUX_SUBPROCESS_FLAGS_WAITABLE` in `p->flags` to track waitable processes. This treats the flag like a local flag on the server side, which seemed most correct given the current libsubprocess architecture and the RFC specification. This could be simplified by just passing `FLUX_SUBPROCESS_FLAGS_WAITABLE` in `local_flags`, but this approach seemed to counter the RFC and also it seems like an explicit approach would be less confusing.

### DIFF
--- a/etc/completions/flux.pre
+++ b/etc/completions/flux.pre
@@ -1934,6 +1934,9 @@ _flux_exec()
         -n --noinput \
         -v --verbose \
         -q --quiet \
+        --label= \
+        --bg \
+        --waitable \
     "
 
     _flux_split_longopt && split=true


garlick:
Quick question: what is the rationale for allowing FLUX_SUBPROCESS_FLAGS_WAITABLE with `flux_rexec_ex()` as opposed to just allowing it with `flux_rexec_bg()`?

grondo:
The flag is not specified as only applicable with the non-streaming RPC in RFC 42, so that constraint was not applied here.

I figured in the future there may be a use case for the waitable flag with the streaming RPC (e.g. allow the process to continue in the background after a disconnect, possibly useful for restart case)

garlick:
OK.  Perhaps this header comment should reference `flux_rexec_bg()` instead of `flux_rexec()` then.
```c
    /* flux_rexec() only: allow flux_subprocess_wait(3) to be called
     * for this subprocess
     */
    FLUX_SUBPROCESS_FLAGS_WAITABLE = 16,
```

grondo:
I also realized `flux_subrpocess_wait(3)` was never added here. I'll add that and corresponding tests.

grondo:
Probably can hold of reviewing this PR until I get the C API done and tested.

grondo:
Actually the new call should probably be `flux_rexec_wait(3)`, since it won't take a `flux_subprocess_t *` argument. Any opinions?

garlick:
No strong opinion here.  Someday it might be nice if all subprocess objects were represented by futures. :shrug:

grondo:
Well that could be done now by just stashing the `flux_subprocess_t *` inside the future and writing a bunch of wrapper calls, but I'm not sure what that would get us?

garlick:
I guess there are some general advantages to a common interface for waiting on, or registering a callback for, asynchronous events, over custom callbacks.  However, subprocesses are pretty complicated and representing one with a single future starts to look like the remote subprocess internal API defined in client.h, which is not very user friendly.

Sorry to get us O/T.  If I ever think of a more appealing interface I will put my money where my mouth is and open an issue :-)

grondo:
Well maybe the `flux_future_t *flux_rexec_bg()` and `flux_future_t *flux_rexec_wait()` interface will scratch that itch for cases where the extra complexity of I/O and associated buffering isn't required. It will at least be easily usable from Python (partially the point of this current exercise is to get an interface into modprobe)

src/common/libsubprocess/server.c
```diff
@@ -508,6 +558,18 @@ static void server_exec_cb (flux_t *h,
 
     p->bg = background;
 
+    /* Reuse FLUX_SUBPROCESS_FLAGS_WAITABLE to indicate to the server this
+     * process is waitable
+     */
```
garlick: This comment confused me a bit, but it makes sense now that I understand the shuffling of flags going on here. Perhaps it could be a bit clearer if we add
> The FLUX_SUBPROCESS_FLAGS_WAITABLE flag is just a flag for the server.  It does not change local subprocess behavior.

or similar.

src/common/libsubprocess/server.c
```diff
@@ -8,6 +8,90 @@
  * SPDX-License-Identifier: LGPL-3.0
 \************************************************************/
 
+/* libsubprocess server - remote subprocess execution service
+ *
+ * OVERVIEW
+ * --------
+ * This module implements a subprocess server that executes processes
```
garlick, line 15: Oh this is excellent, thank you!

grondo:
Marking this one WIP again since the current approach was not acceptable while I go work on other things.

src/common/libsubprocess/subprocess.c
```diff
@@ -579,6 +580,44 @@ flux_future_t *flux_rexec_bg (flux_t *h,
...
+                                int rank,
+                                pid_t pid,
+                                const char *label)
+{
+    char *topic = NULL;
+    flux_future_t *f = NULL;
+
+    if (!h
+        || (pid <= (pid_t) 0 && !label)   // neither pid nor label set
+        || (pid > (pid_t) 0 && label)     // both specified: ambiguous
```
garlick: I don't think the (pid_t) cast is needed here?

src/common/libsubprocess/subprocess.h
```diff
@@ -202,6 +202,21 @@ flux_future_t *flux_rexec_bg (flux_t *h,
...
                               const flux_cmd_t *cmd);
 
+/* Wait for a subprocess started with the FLUX_SUBPROCESS_FLAGS_WAITABLE flag
+ * using either command label or process id. When specifying a `label`, set
+ * `pid = -1`. It is an error to specify both `label` and `pid`.
+ *
+ * If service_name is NULL, then default to `rexec`.
+ *
+ * The future will be fulfilled with the payload ``{"status":i}`` on process
+ * completion, or with an error response if there is an error.
```
garlick: Maybe say "RPC payload" so people know it should be unpacked with `flux_rpc_get_unpack()` (or just say that I guess)

grondo:
Ok, I've made the suggested changes and removed WIP:
- fixed the comment on `flux_rexec_wait()` as suggested (also made the same fix to `flux_rexec_bg()`, which was where the text was originally copied from)
- Added a new `p->waitable` boolean for server use instead of using existing `p->flags` to store waitable flag.

grondo:
Thanks! Will set MWP.

PR #7171, opened 2025-10-27
connectors: ssh: fix hang and improve error messages on failure

grondo:
When users have commands in shell init files (`.bashrc`, `.profile`, `.cshrc`, etc.) that write to stdout even in non-interactive mode (which seems to be more common than you'd think), the error mode from `flux proxy` and the `ssh` connector is not ideal. With a normal `flux_open(3)` call, the call fails with `EINVAL` so a `flux proxy` user just gets an unhelpful error
```
flux-proxy: <URI>: Invalid argument
```
When `flux_open_ex(3)` is used, the result it worse: the call just hangs indefinitely. This is a problem with `flux alloc --bg` for example, which calls `flux_open_ex(3)` to the child instance to detect when it is ready (all Python uses of `flux.Flux()` to open a handle use `flux_open_ex(3)` internally).

It turns out the hang occurs because `flux_open_ex(3)` results in `popen2()` in the ssh connector being called with a flag to capture stderr so that errors from ssh or the remote command can be captured and returned to the user. However, in the case where unexpected stdout is emitted, ssh continues running and the `read_all()` of `stderr` blocks. The solution in this PR just makes the `stderr` file descriptor non-blocking before reading from it.

After the hang with `flux_open_ex(3)` is fixed, `flux proxy` is then switched to using this call to get more detailed errors.

Finally, if `stderr` is empty, the error handling code in the ssh connector now assumes the issue is probably the unexpected `stdout` from shell init files, and an error message with a strong hint to the user is emitted instead of "Invalid argument"

E.g.:
```console
$ flux proxy f2TGxB7u 
flux-proxy: ssh://tuolumne1001/var/tmp/grondo/flux-6MKetr/local-0: Unable to establish clean ssh connection.
Hint: Check that shell init files do not write to stdout when non-interactive.
```

### DIFF
--- a/t/t1106-ssh-connector.t
+++ b/t/t1106-ssh-connector.t
@@ -189,7 +189,22 @@ test_expect_success 'ssh:// stderr is redirected with Python flux.Flux()' '
 	test_must_be_empty open_ex.err &&
 	test_cmp open_ex.expected open_ex.out
 '
-
+test_expect_success 'create test flux command that emits data on stdout' '
+	cat <<-EOF >flux-bad.sh &&
+	#!/bin/sh
+	echo foo
+	exec flux "$@"
+	EOF
+	chmod +x flux-bad.sh
+'
+test_expect_success 'flux_open_ex(ssh://...) fails without hang' '
+	( export FLUX_URI=ssh://localhost$TEST_SOCKDIR &&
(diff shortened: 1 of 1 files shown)


garlick (approved):
Great work running down that annoyance!  I'm certain the world is a better place with this fix.

grondo:
Thanks! I'll set MWP.

PR #7179, opened 2025-10-30
fix slurm URI resolver errors for multinode jobs

### LINKED ISSUE #7178: slurm URI resolver doesn't work reliably on multi-node allocations
Problem: The `slurm` URI resolver plugin leverages the following command to re-run `flux uri` on the first node of the target Slurm allocation:
```
srun --jobid=JOBID --overlap -N1 -n1 flux uri slurm:JOBD
```
However, for multi-node jobs, Slurm apparently tries to "allocate" a different node for each invocation of this command by round-robin, e.g. for a 2 node allocation on `dane[84-85]`:
```
 grondo@dane4:~$ srun --jobid=4073686 --overlap -n1 -N1 hostname
dane84
 grondo@dane4:~$ srun --jobid=4073686 --overlap -n1 -N1 hostname
dane85
 grondo@dane4:~$ srun --jobid=4073686 --overlap -n1 -N1 hostname
dane84
```

Since brokers on non-rank 0 node likely do not have child processes, `flux uri slurm:JOBID` fails every other invocation in this 2 node case (and it will be worse for N > 2).

Using `-w HOSTNAME` seems to fix the issue, so the resolver will have to get the first node of the Slurm allocation and use that here.

grondo:
This PR fixes #7178 by adding `-w <first-node>` to the `srun --overlap --jobid` invocation in the slurm URI resolver to ensure the reinvoked `flux uri` command always runs on the first allocated node.

### DIFF
--- a/t/t2802-uri-cmd.t
+++ b/t/t2802-uri-cmd.t
@@ -153,6 +153,14 @@ test_expect_success 'setup fake srun and scontrol cmds for mock slurm testing' '
 	exec flux uri pid:$$
 	EOF
 	chmod +x srun &&
+	cat <<-EOF >squeue &&
+	#!/bin/sh
+	# used by resolver to get nodelist, just echo a couple fake hosts
+	test -n "\$SQUEUE_FAIL" && exit 1
+	test -n "\$SQUEUE_EMPTY" && exit 0
+	echo foo[100-101]
+	EOF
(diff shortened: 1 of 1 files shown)


grondo:
Thanks @wihobbs! I'll set MWP.

PR #7180, opened 2025-10-30
flux-jobs: fix exit status when jobids provided on command line are unknown

### LINKED ISSUE #7177: `flux jobs JOBID` exits with success even if JOBID is invalid
```console
$ flux jobs f1234
JobID f1234 unknown
$ echo $?
0
```
This may have been intentional because `flux jobs` can take multiple jobids on the command line and some may be successful. However, for the case where _all_ jobids on the command line are invalid, a nonzero exit code seems appropriate.

grondo:
This PR fixes #7177. Now if one or more jobids are provided on the command line, and they are all unknown (i.e .the returned jobs list is empty), then the command exits with an error instead of printing the empty list and exiting with exit code 0.

Affected tests are updated and a new test added.

### DIFF
--- a/t/t2800-jobs-cmd.t
+++ b/t/t2800-jobs-cmd.t
@@ -505,6 +505,13 @@ test_expect_success 'flux-jobs error on all unknown IDs' '
 	test $count -eq 3
 '
 
+test_expect_success 'flux-jobs no error on one known ID' '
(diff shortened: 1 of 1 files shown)


grondo:
Thanks! I'll set MWP.

Issue #7153, opened 2025-10-15
shell: print affinity that would be set for testing purposes


grondo:
Problem: For certain tests, it would be useful to see what cpuset the job shell affinity plugin has determined would be set for the job before attempting to set the cpu binding, but the shell currently doesn't print this information. This would allow verification of expected affinity set by the job shell even when using a fake hwloc topology in a Flux instance that doesn't match the actual hardware (as expected in CI environment for example)

One solution would be to always print the computed affinity with debug mode (i.e. `-o verbose`). Another solution might be a `-o cpu-affinity=dry-run` or similar option. 

@trws suggested using builtin hwloc serialization of cpusets for easy parsing on the other end.

Also cc: @milroy

grondo:
@trws, @milroy: would something like this work? This uses `hwloc_bitmap_list_asprintf(3)` to encode the cpuset that would be set for the shell and each task:

```console
$ flux run -N2 -n3 -o cpu-affinity=dry-run,per-task true
0.287s: flux-shell[1]: cpu-affinity: PUs: 0-1
0.287s: flux-shell[1]: cpu-affinity: task 2: PUs: 0-1
0.295s: flux-shell[0]: cpu-affinity: PUs: 0-1
0.296s: flux-shell[0]: cpu-affinity: task 0: PUs: 0
0.296s: flux-shell[0]: cpu-affinity: task 1: PUs: 1
```

A new `dry-run` option would print the PUs list without actually calling `hwloc_set_cpubind(3)`.

wihobbs:
For reference, here's how `mpibind` prints affinity in verbose mode:

```
flux run -n8 -o mpibind=verbose=1 ./hello
0.069s: flux-shell[0]: mpibind:
mpibind: task   0 nths  1 gpus 3 cpus 89
mpibind: task   1 nths  1 gpus 3 cpus 185
mpibind: task   2 nths  1 gpus 3 cpus 90,186
mpibind: task   3 nths  1 gpus 3 cpus 91,187
mpibind: task   4 nths  1 gpus 3 cpus 92,188
mpibind: task   5 nths  1 gpus 3 cpus 93,189
mpibind: task   6 nths  1 gpus 3 cpus 94,190
mpibind: task   7 nths  1 gpus 3 cpus 95,191
```

It would be hard to be completely consistent (since gpus are being printed here, even though they aren't used) but maybe we should try to match this for the users as best we can? Just a thought.

PR #7181, opened 2025-10-31
shell: add `cpu-affinity=dry-run` and `verbose` options

grondo:
This PR addresses #7153 by adding a new `dry-run` option the shell cpu affinity plugin to print what the affinity would be without actually setting it. A `verbose` option is also added to print the cpusets and still apply them. For backwards compatibility, if these options are used with other options like `per-task` or `map:LIST`, they must come first and be followed by a comma.

e.g.:
```console
$ flux run -N2 -n4 -c2 -o cpu-affinity=dry-run,per-task true
0.095s: flux-shell[0]: cpu-affinity: cpus: 0-3
0.095s: flux-shell[0]: cpu-affinity: task 0: cpus: 0-1
0.095s: flux-shell[0]: cpu-affinity: task 1: cpus: 2-3
0.082s: flux-shell[1]: cpu-affinity: cpus: 0-3
0.082s: flux-shell[1]: cpu-affinity: task 2: cpus: 0-1
0.082s: flux-shell[1]: cpu-affinity: task 3: cpus: 2-3
```

### DIFF
--- a/t/t2604-job-shell-affinity.t
+++ b/t/t2604-job-shell-affinity.t
@@ -98,6 +98,55 @@ test_expect_success 'flux-shell: invalid option is ignored' '
 	test_debug "cat invalid.out" &&
 	grep "invalid option" invalid.out
 '
+#
+# verbose, dry-run tests:
+#
+test_expect_success 'flux-shell: cpu-affinity=verbose prints cpuset' '
+	flux run -n1 -c1 -ocpu-affinity=verbose hostname >verbose.out 2>&1 &&
+	test_debug "cat verbose.out" &&
+	grep "cpus:" verbose.out
+'
+test_expect_success MULTICORE 'flux-shell: cpu-affinity=verbose,per-task prints per-task cpusets' '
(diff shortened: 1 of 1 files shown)


grondo:
Thanks! Let's get this merged and if @milroy @trws need something different we can tweak later.

trws:
This looks great @grondo!  What I don't see right off, is it printing the hwloc logical core numbers or the system "processor" number that the `CPUSET` macros would take? I see it's printing it from an hwloc structure but am not sure where in the pipeline it is.

grondo:
Oh, good question @trws. It is printing the `hwloc_cpuset_t`  as it will be passed to `hwloc_set_cpubind(3)`. However, this is for a topology initialized with `hwloc_topology_restrict(3)`, so I guess that it might be logical core numbers within the restricted topology?

trws:
That sounds right to me, but we should probably check and document. I think hwloc logical cores is exactly what we want it to be, should mean we can mock the topology however we want, but best to be sure.

Issue #7174, opened 2025-10-27
core assignments with unbalanced tasks per node


ryanday36:
When one asks for a number of tasks that is evenly divisible by the number of nodes requested, Flux binds all of those tasks to the same number of cores. E.g.

day36@fluke131:~]$ flux run -N2 -n2 --cores-per-task=1 -o mpibind=off -o cpu-affinity=per-task --label numactl -s | grep physcpu
0: physcpubind: 1 
1: physcpubind: 1 

If the number of tasks is not divisible evenly across the nodes, Flux allows some tasks to access more cpus:

[day36@fluke131:~]$ flux run -N2 -n3 --cores-per-task=1 -o mpibind=off -o cpu-affinity=per-task --label numactl -s | grep physcpu
0: physcpubind: 0 
2: physcpubind: 0 1 
1: physcpubind: 1

This example uses the built-in cpu-affinity shell option, but this is also affecting mpibind and causing it to improperly bind gpus in cases where the user asks for a number of gpus that isn't evenly divisible by the number of nodes. E.g.

day36@rzadams1041:$ MPIBIND_RESTRICT= flux run -N2 -n7 --gpus-per-task=1 --cores-per-task=24 --label -o mpibind=v bash -c 'echo $ROCR_VISIBLE_DEVICES'
0.072s: flux-shell[1]: mpibind: 
mpibind: task   0 nths 96 gpus 0,1 cpus 0-47,96-143
mpibind: task   1 nths 48 gpus 2 cpus 48-71,144-167
mpibind: task   2 nths 48 gpus 3 cpus 72-95,168-191
0.070s: flux-shell[0]: mpibind: 
mpibind: task   0 nths 24 gpus 0 cpus 0-23
mpibind: task   1 nths 24 gpus 1 cpus 24-47
mpibind: task   2 nths 24 gpus 2 cpus 48-71
mpibind: task   3 nths 24 gpus 3 cpus 72-95
0: 0
4: 0,1
1: 1
5: 2
6: 3
2: 2
3: 3
day36@rzadams1041:$

grondo:
This might be due to #3750. IIRC, this is a limitation of Jobspec V1 -- at least presently we can't represent non uniform request across nodes, so uneven distribution results in wasted task slots.

[...]

PR #7182, opened 2025-10-31
shell: fix core/gpu distribution with uneven task counts across nodes

grondo:
This PR is proposed as a fix for #7174.

The fix (or workaround if you prefer) is purposefully applied very narrowly to avoid erroneously reducing the core/gpu affinity in some unforeseen case. That is, the cpuset in the shell is only adjusted if all the following is true:
 - jobspec is v1
 - not an _explicit_ node exclusive allocation
 - nodes were specified in jobspec (i.e. nnnodes != -1)

Similarly the GPU set is only adjusted if all the above is true and:
 - greater than 1 GPU is allocated to this rank

New `gpus_per_slot` and `node_exclusive` members are added to the object provided by `flux_shell_jobspec_info_unpack(3)` since they are required for detection of the above conditions.

### DIFF
--- a/t/t2604-job-shell-affinity.t
+++ b/t/t2604-job-shell-affinity.t
@@ -145,6 +145,24 @@ test_expect_success MULTICORE 'flux-shell: cpu-affinity handles verbose and map 
 	grep "cpus: 0" verbose-map.out
 '
 #
+# Issue #7174 (unequal distribution of tasks across nodes)
+test_expect_success MULTICORE 'flux-shell: issue #7174 reproducer' '
+	flux run -N2 -n3 -c1 -o cpu-affinity=verbose hostname \
+	    >7174.out 2>&1 &&
+	grep "adjusted cores from 0-1 to 0" 7174.out
+'
+test_expect_success MULTICORE 'flux-shell: no #7174 workaround with node exclusive' '
+	flux run -xN2 -n3 -c1 -o cpu-affinity=verbose hostname \
+	    >7174-exclusive.out 2>&1 &&
+	test_must_fail grep "adjusted cores" 7174-exclusive.out
+'
+test_expect_success MULTICORE 'flux-shell: issue #7174 reproducer with per-task' '
+	flux run -N2 -n3 -c1 -o verbose -o cpu-affinity=verbose,per-task true \
+	    >7174-per-task.out 2>&1 &&
+	grep "adjusted cores from 0-1 to 0" 7174-per-task.out &&
+	grep "task 2: cpus: 0" 7174-per-task.out
+'
+#
 # GPU tests:
 #
 test_expect_success 'flux-shell: CUDA_VISIBLE_DEVICES=-1 set by default' '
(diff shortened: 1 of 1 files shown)


grondo:
Also: This is based on top of #7181.

Also also: I still need to test with a node exclusive scheduler policy. In this case, I don't think there is any way for the shell to determine that resources resulted from a node exclusive allocation. If all other conditions are true, then the workaround may still be applied, which might be surprising.

However, I'm posting this PR now to get @ryanday36's feedback.

Finally, I think this solution will work with mpibind since the shell affinity should be adjusted before calling the mpibind plugin, but I have not tested this yet either.

grondo:
I did verify that with a node exclusive match policy, the workaround here is activated with uneven task counts across nodes, e.g. on corona:

```console
$ flux run -N2 -n3 -o verbose true
[snip]
0.087s: flux-shell[1]: DEBUG: cpu-affinity: adjusted cores from 0-47 to 0-23
```

I'm unsure if this is acceptable or not.

grondo:
And I just realized that the mpibind plugin sets `cpu-affinity=off` so these changes will have no effect with mpibind. A similar fix will have to go into mpibind itself I guess.

grondo:
mpibind fetches the allocated cores and gpus using `flux_shell_rank_info_unpack(3)`:
```c
  /*  Get number of local tasks and assigned cores from "rank info" object */
  if (flux_shell_rank_info_unpack(shell,
                   -1,
                   "{s:i, s:{s:s, s:s}}",
                   "ntasks", &ntasks,
                   "resources",
                   "cores", &cores,
                   "gpus", &gpus) < 0) {
    shell_die_errno(1, "flux_shell_rank_info_unpack");
    return -1;
  }
```

Another approach to this PR would be to move the adjustment of the `cores` and `gpus` strings into the shell itself so that the values returned to plugins are the adjusted values. 

I guess that makes a kind of sense since it would fix all plugins that use these values.

ryanday36:
> Another approach to this PR would be to move the adjustment of the `cores` and `gpus` strings into the shell itself so that the values returned to plugins are the adjusted values.
> 
> I guess that makes a kind of sense since it would fix all plugins that use these values.

I'll ask Edgar to look at this as well, but an approach that makes the plugins just work sounds good to me.

grondo:
Thanks. I think I'll go ahead and make the change to the shell's internal copy of "cores" and "gpus", but will change the current debug message that cores/gpus is being adjusted to a warning. Then I'll test again with mpibind.

I'll set WIP on this PR until I get it reworked.

grondo:
I've force pushed a new version of this PR that fixes up the core/gpu assignments in the shell proper, so that `flux_shell_rank_info_unpack(3)` reflects the update core and gpu sets. This allows mpibind to just work:

```console
$ FLUX_SHELL_RC_PATH=/etc/flux/shell/lua.d flux run -N2 -n3 -o mpibind=v true
0.088s: flux-shell[1]:  WARN: corona82: adjusted cores from 0-1 to 0
0.094s: flux-shell[1]: mpibind: 
mpibind: task   0 nths  1 gpus  cpus 0
0.089s: flux-shell[0]: mpibind: 
mpibind: task   0 nths  1 gpus  cpus 0
mpibind: task   1 nths  1 gpus  cpus 1
```

The fix is still only applied under the narrow circumstances listed previously.  The debug message is now updated to a warning (only emitted on affected ranks) so users will know that core/gpu sets on the affected nodes will not match R.

garlick:
oops I reviewed #7181 as part of this.  Will approve that too.

eleon:
> > Another approach to this PR would be to move the adjustment of the `cores` and `gpus` strings into the shell itself so that the values returned to plugins are the adjusted values.
> > I guess that makes a kind of sense since it would fix all plugins that use these values.
> 
> I'll ask Edgar to look at this as well, but an approach that makes the plugins just work sounds good to me.

Thank you, @grondo and @ryanday36 . Yes, this works for mpibind--as you already tested :)

grondo:
> Is there any situation where a user might want to disable this workaround? I can't think of one, but thought I'd pose the question.

I could not think of one either, and to get the fix in expeditiously I left off the option to disable the workaround and the testing and documentation it would require (at least for now) Though I guess in retrospect it would not be that difficult to add another condition to the set checked before triggering the workaround.

garlick:
> I could not think of one either, and to get the fix in expeditiously I left off the option to disable the workaround and the testing and documentation it would require (at least for now) 

I'm good with that.

grondo:
Ok, I'll set MWP here then. Thanks!

PR #7183, opened 2025-11-02
add `flux sproc` command for managing subprocesses

grondo:
This PR moves the `rexec.py` test script to a `flux sproc` command with some minor enhancements and added documentation, tests and tab completion.

(`sproc` was chosen vs `proc` or `subproc` to limit completion conflicts with existing commands. It is expected that this command will have limited real world usage by regular users)

This is built on top of #7170.

### DIFF
--- a/etc/completions/flux.pre
+++ b/etc/completions/flux.pre
@@ -2209,6 +2209,47 @@ _flux_modprobe()
     fi
 }
 
+# flux-sproc(1) completions
+_flux_sproc()
+{
+    local cmd=$1
+    local subcmds="ps kill wait"
+    local split=false
+    common_OPTS="\
+        -r --rank= \
+        -s --service= \
+    "
+    ps_OPTS="\
+        $common_OPTS \
+        -o --format= \
+        -n --no-header \
+    "
+    kill_OPTS="\
+        $common_OPTS \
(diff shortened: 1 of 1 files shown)


grondo:
On second thought, hold off on any review here. It may make sense to add simple wrappers for subprocess  `wait`, `ps`, and `kill` in the python bindings and use that in the `flux sproc` tool to give more general Python access to the functionality.

grondo:
Ok, this PR has been reworked to first add a new `flux.subprocess` Python module with `rexec_bg()`, `wait()`, `kill()`, and `list()` functions, including Python-specific testing in `t/python/t0036-subprocess.py`. 

`flux sproc` is then re-implemented using the functions from this module.

It is ready for a review.

grondo:
I just force-pushed an update that made the `flux sproc kill --wait` behavior a bit more predictable when a subprocess has already exited. Since the `rexec.wait` RPC call is made before sending the kill request, the process will have been reaped by the time the `rexec.kill` request arrives. Instead of exiting with exit code `1` in this case, the command will instead try to exit with the status collected by `wait`. Documentation and tests updated.

grondo:
Thanks! I'll set MWP.

PR #7186, opened 2025-11-04
flux-dump/flux-fsck misc cleanup

chu11:
Splitting this off from #7158, as the number of cleanups commits began growing larger than the number of commits for the actual meat of the PR.


chu11:
re-pushed, throwing an extra `flux-fsck` cleanup in there

garlick (approved):
lGTM!

PR #7187, opened 2025-11-05
allow undrain reason to be updated

garlick:
This  updates the `flux resource undrain` command to take the same options as `flux resource drain` for updating reasons.

When the undrain reason is being updated, an `overwrite=N` field appears in the eventlog like it does with drain.  The idea is that you have to process the whole eventlog to get the current undrain reason, updating it as you go, and adhering to the RFC definition of `overwrite=N`.

One difference from drain is that the undrain reason isn't shown by any of our tools, AFAIK.

If this approach is acceptable, an RFC 44 update will be needed and I'll add tests.  Meanwhile, it's a WIP.

### DIFF
--- a/t/t2311-resource-drain.t
+++ b/t/t2311-resource-drain.t
@@ -146,6 +146,11 @@ test_expect_success 'drain update mode works with idset' '
 	flux resource undrain 0
 '
 
+test_expect_success 'undrain event was posted with overwrite=0' '
+	flux kvs eventlog get resource.eventlog |grep undrain >undrain.out &&
+	grep overwrite=0 undrain.out
+'
+
 test_expect_success 'drain works with idset' '
 	flux resource drain 2-3 &&
 	test $(flux resource list -n -s down -o {nnodes}) -eq 3 &&
@@ -160,7 +165,7 @@ test_expect_success 'reload resource module to simulate instance restart' '
 '
 
 test_expect_success 'undrain one node' '
-	flux resource undrain 3 &&
+	flux resource undrain --force 3 &&
 	test $(flux resource list -n -s down -o {nnodes}) -eq 2
 '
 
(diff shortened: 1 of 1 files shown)


src/modules/resource/drain.c
```diff
@@ -14,20 +14,28 @@
...
+ *
+ * Drain/undrain fails if targets are already in the requested state, unless
+ * one of the following options are used.  The effect on those targets
+ * depends on the options used:
+ *
+ * --update (mode=update, overwrite=0)
+ *   Update reason only if unset.
+ *
+ * --force (mode=force, overwrite=1)
+ *   Update reason only.
```
grondo: Should this be `mode=overwrite`? (I could have confused myself though)

grondo (commented):
This seems reasonable to me!
I did notice one typo in a quick check.

Also I opened a proposal for a tool that might utilize the undrain reason  over in #7189.

src/modules/resource/drain.c
garlick: Thanks fixed.

garlick:
Added a couple tests, so removing WIP.

grondo:
This looks good to me. One potential concern is that the behavior of `flux resource undrain --force` has changed. While I agree it was the right decision to align it with `flux resource drain --force`, we should get feedback from @kkier before merging in case this breaks existing usage patterns.

kkier:
I don't know of any use of `flux resource undrain --force` at all, and nothing shows up on `git grep` through the repo.  I can adapt any of my own tools moving forward and Dom or I will be writing the wrappers for things, so we should be good.

grondo:
Thanks @kkier! I say this is ready to merge then.

garlick:
Great, thanks! I'll set MWP.

PR #7193, opened 2025-11-07
jobtap: add `flux_jobtap_set_load_sort_order()`

### LINKED ISSUE #7192: jobtap: iterate active jobs in predicable order when loading a plugin
Problem: When a jobtap module is loaded the `job.new` and `job.create` callbacks are called for all currently active jobs, but the job-manager generates the list using `zhashx_values()`, so there's no order guarantee.

https://github.com/flux-framework/flux-core/blob/893c23e381a280e9735fffdca3cfe505fa73d5fe/src/modules/job-manager/jobtap.c#L317-L324

This can cause confusion or unexpected behavior in plugins which are accumulating state based on these callbacks, e.g. see https://github.com/flux-framework/flux-accounting/issues/772.

I'm having trouble deciding if ordering by jobid (i.e. loose submission time) would be sufficient here, or if ordering by reverse state (CLEANUP, RUN, SCHED, DEPEND, PRIORITY) would be better - at a  higher cost.

The other thing proposed in the flux-accounting issue was to see if there's some way to let the plugin request a certain order.

grondo:
This PR adds the ability for jobtap plugins to select a guaranteed sort order that the `job.create` and `job.new` callbacks are called on plugin load.

To select a non-default order, plugins must call `flux_jobtap_set_load_sort_order()` with a `mode` of either `state`, to sort first by state then by jobid, or `-state` to sort by the reverse state then by jobid. By default, no sort order is guaranteed.

Fixes #7192

### DIFF
--- a/doc/man7/flux-jobtap-plugins.rst
+++ b/doc/man7/flux-jobtap-plugins.rst
@@ -139,6 +139,19 @@ job.create
 
   It is safe to post events from a ``job.create`` handler in all cases.
 
+  .. note::
+    In case 3 ``job.create`` is called for active jobs in unspecified order.
+    If a plugin requires an ordering guarantee, the plugin should call
+    ``flux_jobtap_set_load_sort_order(3)`` from the ``flux_plugin_init()``
+    callback. This function takes a ``mode`` parameter of either ``state``,
+    to sort jobs by state (then jobid), or ``-state`` to sort by reverse
+    state (then jobid). For example ::
+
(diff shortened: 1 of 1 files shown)


garlick (approved):
This looks fine to me.

grondo:
I'll wait for @cmoussa1's approval then set MWP.

grondo:
Thanks @garlick and @cmoussa1. Setting MWP now.

PR #7194, opened 2025-11-10
flux-fsck: support --job-aware

### LINKED ISSUE #7121: flux-fsck: output job ids in f58 instead of KVS

chu11:
Problem: A single corrupted entry in a job directory will effectively make all data in the job directory unusable (i.e. if one piece of data is corrupted, other uncorrupted data may not be usable).  The --repair option only moves the corrupted data to the lost+found, leaving the uncorrupted data in place.  This can effect several job related modules, that expect specific job data to always be available.

Support a new --job-aware option to flux-fsck.  In concert with the --repair option, if any data in a job directory is corrupted, move
all contents of the job directory to the lost+found.

Fixes #7121

### DIFF
--- a/t/t2816-fsck-cmd.t
+++ b/t/t2816-fsck-cmd.t
@@ -526,5 +526,145 @@ test_expect_success 'remove content & content-sqlite modules' '
 	flux module remove content-sqlite &&
 	flux module remove content
 '
+#
+# --repair && --job-aware
+#
+test_expect_success 'load content, content-sqlite w/ truncate, and kvs' '
+	flux module load content &&
+	flux module load content-sqlite truncate &&
+	flux module load kvs
+'
+test_expect_success 'create some misc content' '
+	flux kvs put val="val" &&
+	flux kvs put dir.val="v" &&
+	flux kvs put --append dir.val="a" &&
+	flux kvs put --append dir.val="l"
+'
+test_expect_success 'create some job-like content' '
+	flux kvs put job.0123.4567.89ab.cdef.J="J" &&
+	flux kvs put job.0123.4567.89ab.cdef.log="1" &&
+	flux kvs put --append job.0123.4567.89ab.cdef.log="2" &&
+	flux kvs put --append job.0123.4567.89ab.cdef.log="3" &&
+	flux kvs put job.0123.4567.89ab.cdef.Adir.data="A" &&
+	flux kvs put --append job.0123.4567.89ab.cdef.Adir.data="B" &&
+	flux kvs put --append job.0123.4567.89ab.cdef.Adir.data="C" &&
+	flux kvs put job.0123.4567.89ab.cdef.Bdir.data="A" &&
+	flux kvs put --append job.0123.4567.89ab.cdef.Bdir.data="B" &&
+	flux kvs put --append job.0123.4567.89ab.cdef.Bdir.data="C" &&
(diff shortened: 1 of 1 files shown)


src/cmd/builtin/fsck.c
```diff
@@ -291,12 +291,10 @@ static void put_valref_lost_and_found (struct fsck_ctx *ctx,
         dir = subdir;
     }
 
-    if (treeobj_insert_entry (dir, name, repaired) < 0)
-        log_err_exit ("cannot insert repaired valref");
+    if (treeobj_insert_entry (dir, name, obj) < 0)
+        log_err_exit ("cannot insert treeobj");
```
garlick: Would it be useful to log the insertion path that is failing?

src/cmd/builtin/fsck.c
```diff
@@ -323,7 +321,7 @@ static void unlink_path (struct fsck_ctx *ctx, const char *path)
     }
 
     if (treeobj_delete_entry (dir, name) < 0)
-        log_err_exit ("cannot unlink repaired entry");
+        log_err_exit ("cannot unlink entry");
```
garlick: Same here, log path?

garlick:
Any thoughts about abstracting some of these functions out into a private library (or private portion of libkvs) that could be used by multiple offline KVS tools or the KVS itself?

It'd be nice to shrink the volume of code in fsck.c and have unit tests for some of the functions in here.

(I'm just asking - mabye that's not practical)

chu11:
> Any thoughts about abstracting some of these functions out into a private library (or private portion of libkvs) that could be used by multiple offline KVS tools or the KVS itself?

In an earlier iteration I did ponder this.  I can't remember the specific reasons why, but the `flux-fsck` needs were (unsurprisingly) simpler than the KVS module needs, and things didn't seem to line up.   But as `flux-fsck` grows and advances, it certainly is something worth re-visiting.  I'll put it in a TODO item.

Edit: and as you mention, splicing things out into a lib just for unit testing does seem like a good idea

chu11:
> splicing things out into a lib just for unit testing does seem like a good idea

So I began to look into splicing out some of the "offline KVS" activities into a utility lib, but as I began working on it, I think its benefit shrank and I no longer think it's worth it.  A lot of that code is quite "fsck" specific ... it can be generalized some, but there's a limit to the benefit by generalizing it too much.

But the bigger issue is b/c we have to test with the content store, "unit" testing will involve

- writing some tool that uses this convenience library
- loading up a flux broker w/ content-sqlite loaded
- temporarily loading the KVS, so we can put some test data into the KVS
- then running tests

Suddenly, the "unit testing" of this convenience library is basically looking exactly like `flux-fsck` testing.

If there comes a time this could be useful between two different tools, I think we can revisit.

garlick:
I made a couple of comments above about logging paths of fatal errors that might have been missed?

chu11:
> I made a couple of comments above about logging paths of fatal errors that might have been missed?

I'm not seeing any comments.  Is it a draft review on your end and you forgot to submit?

chu11:
@garlick thanks, fixed up those loggings and rebased + repushed.

garlick (approved):
LGTM!  Nice job.

chu11:
@garlick thanks!  will set MWP

PR #7197, opened 2025-11-11
improve error messages and documentation for submission directives

grondo:
Users have been commonly confused when trying to comment out submission directives in their scripts.
They understandably, but erroneously, make the assumption that a directive can be commented out by adding
an extra comment character to the comment, e.g. from `# flux: ` to `## flux:`. When they do this, they get an
unhelpful "sentinel changed" error.

I looked into special casing this comment method, but it largely breaks Flux's language-agnostic approach
(Flux doesn't know what the comment character or characters are for the current script) requires special
cases that would break in other situations, and didn't seem like the correct long-term approach.

Rather than support the `## flux:` kludge, this PR instead improves the error message when there's a sentinel
change to give explicit instructions how to comment out an entire directive. E.g.:

```
flux-batch: ERROR: b.sh: line 4: sentinel changed from '# flux:' to '## flux:'

All directives in a file must use the same sentinel.

If you're trying to disable this directive, place a comment
character AFTER the sentinel, not in the prefix:

  # flux: # --cores-per-task=2

The '#' character comments out the rest of the line, making
this directive a no-op while maintaining the same sentinel.
```

Additionally, the SUBMISSION DIRECTIVES section of the `flux-batch(1)` man page is greatly expanded and now includes many examples.

Hopefully these improvements resolve the user confusion.

### DIFF
--- a/doc/man1/flux-batch.rst
+++ b/doc/man1/flux-batch.rst
@@ -174,9 +174,10 @@ mixed within the submission script. The submission directive specification
 is fully detailed in RFC 36, but is summarized here for convenience:
 
  * A submission directive is indicated by a line that starts with
-   a prefix of non-alphanumeric characters followed by a tag ``FLUX:`` or
-   ``flux:``. The prefix plus tag is called the *directive sentinel*. E.g.,
-   in the example below the sentinel is ``# flux:``: ::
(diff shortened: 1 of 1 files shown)


grondo:
Thanks! I'll set MWP.

PR #7208, opened 2025-11-17
ci: fix mount permissions for 1M tmpfs fs in docker

grondo:
Recent versions of docker appear to no longer guarantee a mode of `1777` for `--mount type=tmpfs` filesystems. This causes `t0090-content-enospc.t` to fail with permissions errors.

Explicitly set `tmpfs-mode=1777` to guarantee expected permissions in the docker container.


grondo:
Thanks!

PR #7209, opened 2025-11-19
broker: decouple state machine and overlay

garlick:
Problem: the broker overlay subsystem and state machine are directly coupled with callbacks and accessor functions, making the eventual transition of the overlay subsystem to a built-in module (in a separate thread) difficult.

This pulls the two systems apart, preparing for
- #6967

I'll give this another once over before removing WIP.

### DIFF
--- a/t/valgrind/valgrind.supp
+++ b/t/valgrind/valgrind.supp
@@ -74,7 +74,7 @@
    <issue_3522>
    Memcheck:Leak
    match-leak-kinds: definite
-   fun:malloc
+   fun:*alloc
    ...
    obj:*libhwloc.so.15.*.*
    obj:*libhwloc.so.15.*.*


garlick:
Rebased on current master.

src/broker/state_machine.c
```diff
@@ -1336,6 +1336,32 @@ static void overlay_monitor_cb (struct overlay *overlay,
...
+                             "event", &event,
+                             "error", &error) < 0) {
+        flux_log_error (h, "state-machine.post");
+        return;
+    }
+    if (!flux_msg_is_noresponse (msg)) {
+        (void)flux_respond_error (h, msg, EPROTO, NULL);
+        return;
+    }
+    state_machine_post (s, event);
```
grondo, line 1404: Should `event` be validated here? Since it is internally only I'm guessing not necessary, but thought I'd ask.

src/broker/state_machine.c
```diff
@@ -573,7 +573,27 @@ static void shutdown_warn_timer_cb (flux_reactor_t *r,
     flux_watcher_start (w);
 }
 
-
+/* SHUTDOWN state triggers an orderly shutown of the TBON subtree.
```
grondo: typo: shutown

src/broker/state_machine.c
```diff
@@ -1237,7 +1264,16 @@ static void state_machine_monitor_cb (flux_t *h,
         log_monitor_respond_error (h);
 }
 
-static void monitor_continuation (flux_future_t *f, void *arg)
+/* When the the TBON parent broker state changes, consider whether to post
```
grondo: typo: the the

src/broker/broker.c
```diff
@@ -1854,30 +1854,16 @@ void broker_request_sendmsg_new (broker_ctx_t *ctx, flux_msg_t **msg)
 
 /* Route a response message, determining next hop from route stack.
  * If there is no next hop, routing is complete to broker-resident service.
- * If the next hop is an overlay peer, route up or down the TBON.
```
grondo, line 1827: Commit message: 

> This change revealed that the overlay code logs unnecessary
> errors it fails to route a response to an offline child.

Should that be "_if_ it fails to route a response"?

src/broker/overlay.c
```diff
@@ -736,21 +751,30 @@ static void sync_cb (flux_future_t *f, void *arg)
...
+int overlay_start (struct overlay *ov)
 {
-    if (!ov->f_sync) {
-        if (!(ov->f_sync = flux_sync_create (ov->h, sync_min))
-            || flux_future_then (ov->f_sync, sync_max, sync_cb, ov) < 0)
-            return -1;
+    if (!ov) {
+        errno = EINVAL;
+        return -1;
     }
```
grondo: This defensive check seems good, but I do notice there aren't similar checks for invalid `ov` passed to other overlay functions, so it jumped out at me. If it remains, it might be good to add an EINVAL test to the unit tests. IMO it could probably just be dropped since there is really only one user.

src/broker/state_machine.c
garlick: I think it is probably OK without validation.  Posting an unknown event will log `<eventname>: ignored in <state>` so it should be easy to track down coding errors where posting an event doesn't cause the intended effect.

src/broker/overlay.c
garlick: Yeah I'll just drop that.

garlick:
Thanks, fixed those and I'll set MWP.

PR #7210, opened 2025-11-20
broker: factor out bootstrap table parser

garlick:
Problem: parsing the TOML `[bootstrap]` table is rather complicated and co-mingled with bootstrap and topology in the broker config bootstrap method, making the broker difficult to refactor.

Abstract parsing of `[bootstrap]` in a stand-alone function with unit test.

### DIFF
--- a/src/broker/boot_config.c
+++ b/src/broker/boot_config.c
@@ -24,6 +24,7 @@
 #include <flux/hostlist.h>
 #include <flux/taskmap.h>
 
+#include "src/common/libfluxutil/conf_bootstrap.h"
 #include "src/common/libyuarel/yuarel.h"
 #include "src/common/libutil/errprintf.h"
 #include "src/common/libutil/errno_safe.h"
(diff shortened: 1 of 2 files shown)


grondo:
Perhaps this should be obvious, but I'd like to understand the rationale for moving `conf_bootstrap.c` into `libflux-core.so` as a non-public function. What are the tradeoffs compared to keeping it in `src/broker`?

For code shared between multiple modules, the `libfluxutil` convenience library approach had some advantages: it avoided copying non-public code into the DSO, and the separation clearly indicated the code was shared but not part of the public API.

What's the intention going forward for shared, non-public code? Should it generally go into `libflux-core.so` with hidden visibility, or are there guidelines for when to use convenience libraries versus including it in the main library?

garlick:
The refactoring was necessary to tease apart the various activities at startup (pmi, topology, zmq binding/connecting, etc), in order to break overlay out to a built-in module.  However, the convenience library change is not necessary to achieve that.

I was hasty in concluding that `libfluxutil` was not serving any purpose and based on your two points above, which I do agree with now that you bring them up, I'll go ahead and put it back.

Although `conf_bootstrap_parse()` could remain in the broker source tree, it does need to be called from the overlay too, and potentially from a upmi plugin if one idea pans out, so it may be helpful to have it broken out to `libfluxutil` to underscore that it is not private to the broker.

garlick:
Changed slightly to retain `libfluxutil`.

src/common/libfluxutil/conf_bootstrap.c
```diff
@@ -0,0 +1,464 @@
...
+                          const char *hostname,
+                          bool *enable_ipv6,
+                          const char **curve_cert,
+                          json_t **hosts,
+                          flux_error_t *errp)
+{
+    flux_error_t error;
+    json_error_t jerror;
+    json_t *bootstrap = NULL;
+    struct bootstrap_info binfo = { 0 };
```
grondo: Would it be simpler and more clear to directly initialize `default_port` to `-1`?
```c
    struct bootstrap_info binfo = { .default_port = -1 };
```

src/common/libfluxutil/test/conf_bootstrap.c
```diff
@@ -0,0 +1,461 @@
...
+    if (!tmpdir)
+        tmpdir = "/tmp";
+    if (snprintf (path, size, "%s/testbootstrap-XXXXXX", tmpdir) >= size)
+        BAIL_OUT ("buffer overflow while creating test directory");
+    if (!mkdtemp (path))
+        BAIL_OUT ("could not create test directory");
+    if (keep)
+        cleanup_push_string (cleanup_directory_recursive, path);
+    else
+        diag ("%s will be preserved", path);
```
grondo, line 56: This is confusing or I'm reading something wrong. If `keep` is true then the directory is cleaned up, and if  `!keep` then the directory is preserved?

src/common/libfluxutil/test/conf_bootstrap.c
```diff
@@ -0,0 +1,461 @@
...
+}
+
+int main (int argc, char *argv[])
+{
+    bool keep = false;
+
+    plan (NO_PLAN);
+
+    if (argc < 1 && streq (argv[1], "keep"))
+        keep = true;
```
grondo: Could this ever be true? (was it supposed to be `argc > 1`?)

src/common/libfluxutil/test/conf_bootstrap.c
```diff
@@ -0,0 +1,461 @@
...
+    test_bad_hosts_entry ("{ host=\"foo\", connect=\"ipc://foo:*\" }",
+                          "wildcard connect string");
+
+    test_special_single ("", "missing hosts array");
+    test_special_single ("hosts = []", "empty hosts array");
+
+    test_missing_cert ();
+    test_dup_hosts();
+
+    // cover parent pointing to bad host
```
grondo: misplaced comment? Or is this a TODO?

src/broker/boot_config.c
```diff
@@ -415,106 +142,38 @@ int boot_config_getrankbyname (json_t *hosts,
...
- * Return 0 on success, -1 on failure.
+/* Look up the host entry for 'rank' and return bind URI, if any.
  */
-int boot_config_getbindbyrank (json_t *hosts,
-                               struct boot_conf *conf,
-                               uint32_t rank,
-                               char *buf,
-                               int bufsz,
-                               flux_error_t *errp)
+const char *getbindbyrank (json_t *hosts, uint32_t rank, flux_error_t *errp)
```
grondo: Should this function be declared static?

src/broker/boot_config.c
```diff
@@ -415,106 +142,38 @@ int boot_config_getrankbyname (json_t *hosts,
...
- * Return 0 on success, -1 on failure.
+/* Look up the host entry for 'rank', and return connect URI, if any.
  */
-int boot_config_geturibyrank (json_t *hosts,
-                              struct boot_conf *conf,
-                              uint32_t rank,
-                              char *buf,
-                              int bufsz,
-                              flux_error_t *errp)
+const char *geturibyrank (json_t *hosts, uint32_t rank, flux_error_t *errp)
```
grondo: static?

src/common/libfluxutil/test/conf_bootstrap.c
garlick: Oh that is wrong, good catch!

src/broker/boot_config.c
garlick: Yes and several others too!

src/common/libfluxutil/test/conf_bootstrap.c
garlick: That was a TODO.  I filled in the missing tests.

garlick:
Thanks for the review!  I fixed those and will set MWP.

PR #7221, opened 2025-12-02
broker: assign event seqs in the broker not overlay

garlick:
Problem: flux-framework/flux-core#7109 moved all event publishing from the broker to the overlay subsystem so that the peer multicast portion could eventually run in separate thread.  Unfortunately, moving all event publication machinery to overlay creates an unnecessary requirement that overlay be running on a size=1 instance, and will complicate `flux module reload overlay`, should that be useful.

Leave the resource intensive event mcast code in overlay but move the rest of it back into the broker.

TL;DR

Recall there are two ways that event messages are routed per RFC 3:

1) A bare event message is forwarded upstream on the TBON and published when it arrives on rank 0.

2) An event message is base64 encoded and encapsulated in a request message that is sent to rank 0, where it is published.

Publication occurs only on rank 0 and consists of assigning a monotonically increasing sequence number, distributing the message to local broker and module subscribers, and sending the event message to the overlay via the interthread message channel for further distribution.

The overlay now routes events as follows:

- If received from the local broker on the interthread channel: On rank 0, messages are mcast to all children. On rank > 0, messages are forwarded to the parent for publication.

- If received from an overlay child: On rank 0, messages are forwarded to the local broker for publication. On rank > 0, message are forwarded to the parent for publication

- If received from the overlay parent (rank > 0), messages are mcast to all children AND sent to the local broker on the interthread channel for distribution to local broker and module subscribers.

Update the overlay unit test that were exercising full event publication. Update some event sharness tests that used the other RPC name.

### DIFF
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -56,6 +56,7 @@
 #include "src/common/libutil/basename.h"
 #include "src/common/librouter/subhash.h"
 #include "ccan/array_size/array_size.h"
+#include "src/common/libccan/ccan/base64/base64.h"
 #include "ccan/str/str.h"
 #include "ccan/ptrint/ptrint.h"
 #ifndef HAVE_STRLCPY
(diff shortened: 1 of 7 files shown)


garlick:
Rebased on current master.

grondo (approved):
LGTM!

garlick:
Thanks I'll set MWP

Issue #7190, opened 2025-11-06
job-list: stream responses


chu11:
The job-list module returns job-list results in a giant blob, which can get quite large on big systems.  This can cause HOL blocking.

Data could / should be streamed instead.

garlick:
RFC 43 would need an update too.

chu11:
hmmm, for backwards compatibility, should we stream only if the request is a streaming rpc?

I think low-probability someone is calling `job-list` directly ... but it is possible.

PR #7222, opened 2025-12-03
t: various test cleanup

chu11:
A variety of test cleanup, mostly surrounding the `rpc` and `rpc_stream` testing tools.

There may be a bit of "huh? why this cleanup" when reviewing.  It's because of issue #7190 ... converting `job-list` to stream responses rather than send everything in one big chunk.  i.e. a number of tests will be converted from the `rpc` tool to `rpc_stream` tool.

A lot of cleanup just "came along" as a result.

### DIFF
--- a/t/t1007-kvs-lookup-watch.t
+++ b/t/t1007-kvs-lookup-watch.t
@@ -823,26 +823,26 @@ test_expect_success 'kvs-watch.lookup request with empty payload fails with EPRO
 # N.B. FLUX_KVS_WATCH = 4
 test_expect_success 'kvs-watch.lookup request non-streaming w/ WATCH fails with EPROTO(71)' '
 	echo "{\"namespace\":"foo", \"key\":\"bar\", \"flags\":4}" \
-		${RPC} kvs-watch.lookup 71
+		| ${RPC} kvs-watch.lookup 71
 '
 # N.B. FLUX_KVS_STREAM = 512
 test_expect_success 'kvs-watch.lookup request non-streaming w/ STREAM fails with EPROTO(71)' '
 	echo "{\"namespace\":"foo", \"key\":\"bar\", \"flags\":512}" \
-		${RPC} kvs-watch.lookup 71
+		| ${RPC} kvs-watch.lookup 71
 '
 test_expect_success 'kvs-watch.lookup request w/ WATCH and STREAM fails with EPROTO(71)' '
 	echo "{\"namespace\":"foo", \"key\":\"bar\", \"flags\":516}" \
-		${RPC_STREAM} kvs-watch.lookup 71
+		| ${RPC_STREAM} kvs-watch.lookup 71
 '
 # N.B. FLUX_KVS_WATCH_FULL = 64
 test_expect_success 'kvs-watch.lookup request extra WATCH flag w/o WATCH fails with EPROTO(71)' '
 	echo "{\"namespace\":"foo", \"key\":\"bar\", \"flags\":64}" \
-		${RPC_STREAM} kvs-watch.lookup 71
+		| ${RPC_STREAM} kvs-watch.lookup 71
 '
 # N.B. FLUX_KVS_WATCH_INITIAL_SENTINEL = 1024
 test_expect_success 'kvs-watch.lookup request INITIAL_SENTINEL w/o APPEND fails with EPROTO(71)' '
 	echo "{\"namespace\":"foo", \"key\":\"bar\", \"flags\":1028}" \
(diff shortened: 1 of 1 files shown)


t/request/rpc.c
```diff
@@ -29,7 +29,8 @@ static const struct option longopts[] = {
 
 void usage (void)
 {
-    fprintf (stderr, "Usage: rpc [-r] [-R] topic [errnum] <payload >payload\n");
+    fprintf (stderr,
```
grondo, line 32: Noticed when trying to check out this branch that a binary file `t/request/rpc` was committed here.

chu11: doh!  thanks, re-pushed

garlick: Apologies for not commenting earlier, but any reason not to use the more modern "embed a python one liner in a shell function" method of testing custom RPCs in new tests?  That has the advantage of being customizable for the test at hand without concern for breaking other tests that share it.

That said, I guess if that would be a load of more work on top of what's already been done, we could just move forward with this.

chu11: I didn't think about that, mostly b/c we use the `t/request/rpc` and `t/request/rpc_stream` tools in so many places already.

Perhaps that's a cleanup for another day if we want to try and convert everything to python one liners?

garlick: I didn't mean we should convert everything (especially not right now) but just use the modern approach going forward so we don't have a bigger knot to untangle later if we decide to do that.

However, it's not that important right now and this work is already done so I'm fine with proceeding as is.

t/t2233-job-info-update.t
```diff
@@ -438,7 +438,7 @@ test_expect_success 'update-lookup request with empty payload fails with EPROTO(
 test_expect_success 'update-watch request with empty payload fails with EPROTO(71)' '
 	${RPC} job-info.update-watch 71 </dev/null
 '
-test_expect_success 'update-lookup request with invalid key fails with EINVAL(22))' '
+test_expect_success 'update-lookup request with invalid key fails with EINVAL(22)' '
```
grondo, line 441: commit message typo: "test descripts"

t/request/rpc.c
```diff
@@ -29,7 +29,8 @@ static const struct option longopts[] = {
 
 void usage (void)
 {
-    fprintf (stderr, "Usage: rpc [-r] [-R] topic [errnum] <payload >payload\n");
```
grondo, line 32: Commit message: "Support checks for substring" :arrow_right: "Support checks for _a_ substring"

t/t0037-content-files-checkpoint.t
```diff
@@ -156,8 +156,7 @@ test_expect_success 'flux content checkpoint list shows correct checkpoints (6)'
 '
 
 test_expect_success 'checkpoint-put bad request fails with EPROTO' '
-	test_must_fail $RPC content.checkpoint-put </dev/null 2>badput.err &&
-	grep "Protocol error" badput.err
+	$RPC content.checkpoint-put 71 "Protocol error" </dev/null
```
grondo, line 159: Commit message nit: "for a error message" :arrow_right: "for an error message"

t/request/rpc_stream.c
```diff
@@ -12,13 +12,26 @@
 #include "config.h"
 #endif
 #include <unistd.h>
+#include <getopt.h>
```
grondo, line 15: Commit message suggestion: "inconsistent to the positional arguments" :arrow_right: "inconsistent with the positional arguments" ?

t/request/rpc_stream.c
```diff
@@ -28,7 +28,9 @@ static const struct option longopts[] = {
 
 void usage (void)
 {
-    fprintf (stderr, "Usage: rpc_stream [-e <end_key>] topic <payload\n");
+    fprintf (stderr,
```
grondo, line 31: Commit message: "consistency to" :arrow_right: "consistency with"?

t/t2231-job-info-eventlog-watch.t
```diff
@@ -526,10 +526,9 @@ test_expect_success 'eventlog-watch request with empty payload fails with EPROTO
 	${RPC} job-info.eventlog-watch 71 </dev/null
 '
 
-test_expect_success 'eventlog-watch request invalid flags fails' '
+test_expect_success 'eventlog-watch request with invalid flags fails with EPROTO(71)' '
```
grondo, line 529: Commit message: "a error" :arrow_right: "an error"

chu11:
Thanks.  rebased & re-pushed with fixes, I'll set MWP

PR #7224, opened 2025-12-03
job-list: stream responses from job-list

### LINKED ISSUE #7190: job-list: stream responses

chu11:
Problem: The job-list module responds to requests with a single large array of all jobs.  This response can be extremely large, and could lead to head-of-line blocking in the TBON.

Instead, stream job-list responses to requestors, one job at a time.

As a result update callers in:
- flux-job
- flux-top
- python JobListRPC

Update tests in t2260-job-list.t.

Fixes #7190

Built on top of #7222 

Associated RFC update: https://github.com/flux-framework/rfc/pull/482

### DIFF
--- a/t/t2260-job-list.t
+++ b/t/t2260-job-list.t
@@ -829,6 +829,17 @@ test_expect_success 'flux job list only completed jobs (legacy RPC)' '
 	test_cmp completed.ids legacy_result_completed.out
 '
 
+# non-streaming backwards compatibility
+
+test_expect_success 'flux job list works with non-streaming RPC' '
+	id=$(id -u) &&
+	constraint="{ userid:[${id}] }" &&
+	jq -j -c -n  "{max_entries:1000, attrs:[], constraint:${constraint}}" \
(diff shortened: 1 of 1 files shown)


chu11:
re-pushed, supporting backwards compatibility with non-streaming RPC requests.  RFC appropriate change coming soon.

grondo:
I reran local testing on this branch with 100K jobs and this branch actually performs slightly  _better_ end-to-end with `flux jobs` on my test system. I'm uncertain why I was getting different results before.

In retrospect this makes sense since the Python side doesn't have to ingest a 100K entry json object and create a giant list of dictionaries. Anyway, this looks good and is performing well for me.

chu11:
re-pushed, re-working per RFC 43 changes.  Jobs are streamed back within an array, so we can support "chunks of jobs" streaming in the future.

PR #7226, opened 2025-12-04
docker: create manifest for non-multi-arch images on tag

### LINKED ISSUE #7225: docker: default docker images not updated on flux-core tags
Problem: When ci is triggered for the master branch or a tag, docker images are deployed to Docker Hub via the `src/test/docker-deploy.sh` script. However, on a tag only the `fluxrm/flux-core:IMAGE-TAG` image is updated, and not the `fluxrm/flux-core:IMAGE` image. This means that just after a tag, default docker images will still be reporting the old version.

grondo:
Problem: When docker images are pushed as part of CI after a flux-core tag, no fluxrm/flux-core:IMAGE docker tags are created for non-multi-arch images. Instead there is only a fluxrm/flux-core:IMAGE-TAG. For multi-arch images, a manifest is created at fluxrm/flux-core:IMAGE combining all the multi-arch images, but for non-multi-arch images the manifest step is skipped.

Do not skip creation of the manifest if the manifest tag differs from the image tag. This will create a manifest at the expected name for images created from a tag.

Fixes #7225

### DIFF
--- a/src/test/docker-manifest.py
+++ b/src/test/docker-manifest.py
@@ -18,8 +18,13 @@ for entry in matrix["include"]:
     if "DOCKER_TAG" in entry["env"]:
         tags[entry["image"]].append(entry["env"]["DOCKER_TAG"])
 
-# Collect only those images with multiple tags:
-tags = {k: v for k, v in tags.items() if len(v) > 1}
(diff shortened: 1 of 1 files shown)


grondo:
Thanks!

PR #7229, opened 2025-12-05
bug: python validation multiplies string

vsoch:
This was a bug discovered by @milroy in `resource_walk` discussed [here](https://github.com/compspec/fractale/pull/21#pullrequestreview-3532968709). When counting resources for a jobspec, we saw an output that was obviously a multiplied string (e.g., `222222` instead of the multiplied value (12). The fix should be fairly straight forward - parse the count and resource count as integers. They can never be a non integer.

Problem: When we use the walk helper, the resource count assumes the inputs to be integers. If they are strings, we get multiplied string output.
Solution: parse the count and resource count as int.

### DIFF
--- a/src/bindings/python/flux/job/Jobspec.py
+++ b/src/bindings/python/flux/job/Jobspec.py
@@ -169,6 +169,29 @@ def validate_jobspec(jobspec, require_version=None):
     return True, jobspec
 
 
+def validate_counts(resources):
+    """
+    Validate counts of resources. Counts must be a positive integer.
+
+    :param resources:  dictionary following the specification in RFC14 for
+                       the `resources` top-level key
+    """
+    for resource in resources:
+        # e.g., count: 2, type: core, count type: int
+        count = resource["count"]
+        res_type = resource["type"]
+        count_type = type(count).__name__
+
+        # Count of resource must be an integer
+        if not isinstance(count, int):
+            raise ValueError(
+                f"resource count must be an integer for type '{res_type}' (got '{count_type}')"
+            )
+        # Count must be a non-negative integer (0) not allowed, already tested.
+        # Recursively call for with blocks
+        validate_counts(resource.get("with", []))
+
+
 def _attr_key_prepend(key):
     """
     Amend a key provided to setattr, getattr, etc. to prepend ``attributes.``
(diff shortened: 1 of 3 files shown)


grondo:
Good catch! 

[RFC 25](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_25.html) states:

> The count SHALL be a single integer value representing a fixed count

I wonder if the correct behavior (at least for the v1 validator) would be to reject jobs with an invalid `count` value.

I think canonical Jobspec does allow non-integer counts, but those represent a range and do not seem to be supported by the current resource count logic.

vsoch:
We can definitely do that. The function where this is relevant is in the resource count:

```
        count_dict = collections.defaultdict(lambda: 0)
        for _, resource, count in self.resource_walk():
            count_dict[resource["type"]] += count
        return count_dict
```
We can put the check somewhere in the validator, but we need to make sure that, for example, if someone is using the resource walk to do the count, they don't get to this erroneous state. Is the implication that validation is required first before any count or resource walk?

grondo:
IIRC, the "validator" just ensures the JSON jobspec can successfully be initialized as a jobspec object:

https://github.com/flux-framework/flux-core/blob/76693fa952fdff63cac5bc2296f09c8b61f6a3d3/src/bindings/python/flux/job/Jobspec.py#L141C5-L169C25

So we could add a check in the Jobspec initializer that, for v1, ensures all resource counts are type `int`.

grondo:
I was thinking something like the following at the bottom of `_v1_validate()`:
(not tested)
```python
        def validate_counts(resources):
            for resource in resources:
                count = resource["count"]
                if not isinstance(count, int):
                    res_type = resource["type"]
                    count_type = type(count)
                    raise ValueError(
                        f"resource count must be an integer for type {res_type} (got {count_type})"
                    )
                validate_counts(resource.get("with", []))

        validate_counts(resources)
```

vsoch:
Will need guidance on the mac-os test - not sure if it's flaky (and I can't figure out from the output what it isn't happy about).

image

Did I need to update somewhere else to add the new file?

grondo:
I think those macos tests are flaky (certainly nothing to do with this PR) I restarted the builer.

PR #7230, opened 2025-12-07
ci: fetch tags for flux-accounting build check

grondo:
Problem: flux-accounting's configure now checks the flux-core version, but the CI build doesn't fetch tags, causing flux-core to report an incorrect version.

Fetch tags before the build and install of flux-core in the flux-accounting CI check.


grondo:
Actually, this PR is not necessary. I was confused by a failure in my personal fork (which doesn't have the latest flux-core tags pushed.) Closing.

Issue #7228, opened 2025-12-05
modprobe: only one task can specify `after=["*"]`


grondo:
The handling of the special case `after=["*"]` in modprobe needs some improvment. This case is supposed to just be a shorthand for "make sure this task runs after everything else". However, if multiple tasks specify `*`, then a cycle is created and modprobe exits with an error like the following:
```
 flux-modprobe: ERROR: ('nodes are in a cycle', ['run-all-rc1', 'foo', 'run-all-rc1'])
```

In fact the same thing happens in this case if task `foo` uses `after=["run-all-rc1"]`. Though that doesn't make sense, perhaps the tool could report a better error.

Probably if multiple tasks specify "after=*", then those tasks should just be grouped to run in parallel. (i.e. `after=*` doesn't apply to other tasks that also have `after=*`)

PR #7231, opened 2025-12-07
modprobe: allow multiple tasks with `after=["*"]`

grondo:
This PR fixes issue #7228 where multiple modprobe tasks using `after=["*"]` incorrectly trigger a circular dependency error.

Changes:
- Allow multiple tasks to specify `after=["*"]` by excluding other `after=["*"]` tasks from dependency lists, allowing them to run in parallel after all other tasks
- Add validation to prevent tasks from mixing `"*"` in before/after with opposite constraints, which creates ambiguous semantics
- Document the special `"*"` value in modprobe task before/after lists, including constraints and circular dependency warnings
- test coverage for multiple `before=["*"]` and  `after=["*"]` tasks, circular dependency detection, and constraint validation

Note: Multiple tasks with `before=["*"]` already worked correctly.

### DIFF
--- a/doc/man1/flux-modprobe.rst
+++ b/doc/man1/flux-modprobe.rst
@@ -369,11 +369,29 @@ following optional arguments:
    are loaded. During shutdown (``rc3``), this task runs before the listed
    modules are removed.
 
+   The special value ``"*"`` ensures this task runs before all other named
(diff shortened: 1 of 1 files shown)


grondo:
Thanks! I'll set MWP.

Issue #7233, opened 2025-12-09
idea: health check at startup


garlick:
Problem: as noted in #7220, when a broker restarts, there is no node "health check" that would detect trouble with a node (after an OOM for example) before allowing jobs to run there.

The simple workaround proposed in #7220 was to suggest a site-provided script could be dropped in to `rc1.d`.  If the script fails, startup aborts.  However, this does mean that the cause of the node being offline is not immediately obvious.  The sys admin would have to check the node's journal to see any errors logged by the script.

This seems like a general problem.  A related one is 
- #6590 

Just wanted to get this open to see if we can come up with a generalized solution for this class of problems.

vsoch:
This is a great idea, and would be really powerful designed as a plugin framework (so it is easy to write and register custom health checks). Another question is if checks might vary based on node in the cluster, or even user or time of day. Examples why:

- My nodes may have different kinds of GPU (or more generally, hardware)
- A user job was doing something specific that leads to well-known kinds of errors (e.g., Usernetes or a container, which could have something still running)

Things to think about:

- If health checks should have triggers based on metrics (e.g., if a node does X this many times, we need to respond with action Y)
- If I could run health checks on demand (e.g., "Check that this node networking is still OK in this way."
- What is the distinction between a health check and collection of some metric - a binary outcome (good / bad, ok / not ok)?
- To follow up to the above - if the check fails and the script aborts, that implies a binary outcome (and turning metrics into thresholds)
- How would checks be put in logical groups for execution (e.g., "Run all the NVIDIA GPU checks)
- Can health checks be added by users too?
- Do any health checks require more privilege? 
- Do results of health checks get saved over time?

garlick:
Good thoughts @vsoch !

I think our admins already have a workable node health check system which they run at startup and _may_ be running from housekeeping as well.   I suspect they would prefer a mechanism to hold resources offline until a site-provided script returns successfully.

Where to hook that in is a question.  We've discussed `resource` module (the core one) plugins in the past.  One approach might be to add plugins with an initial `healthcheck` callback that are passed the `R`  fragment for the local node.  A simple plugin that just runs a (local) script could be provided.  The `resource` module could hold the local node offline until all plugin `healthcheck` callbacks return success.

@vsoch's :brain: :cloud_with_lightning: list above is pretty helpful when thinking about such a new capability!

coffeeplusplus:
Another healthcheck use-case suggestion from SNL.

[...]

Issue #7227, opened 2025-12-05
sysadmins would prefer brokers not be restarted automatically in systemd


grondo:
Because in a system instance it is unexpected that Flux brokers stop unintentionally, sysadmins would prefer if brokers that stop due to some error condition are not automatically restarted by systemd.

This is related to #7223, where a node is being OOMed, the broker heartbeat timeout elapses, and so the broker shuts down while a job is running. If the node is still in CPX mode, for example, the broker restarts and loads the incorrect hwloc topology.

This will be somewhat mitigated when #7220 is fixed, because for this case the node would get drained with an appropriate reason.

Some other concerns brought up by the sysadmins:
- For the case of an OOM, the intent is to eventually allow the node to come back up automatically if only user work was terminated by the oom-killer and other system processes appear to be fine. In this case, I guess the Flux broker should automatically restart, but there may need to some healthcheck process in place before the node is allowed to be returned to service.
- If the broker doesn't restart, then housekeeping may not run (related to the above)
- Methods to differentiate the cause of restart may be useful.

garlick:
Having a way to run a local health check in rc1 sounds like it would be useful.  I think if rc1 fails the broker exits with the special exit code that inhibits restart.

Maybe there are other cases where that exit code should be used.

garlick:
Let's put any ideas about health check at startup in
- #7233 

and keep this focused on broker restart behavior.   The relevant systemd unit bits are:
```
[Service]
Restart=always
RestartSec=30s
RestartPreventExitStatus=42
SuccessExitStatus=42
```
Note:  in the unit file, we set `-Sbroker.exit-norestart=42` on the broker command line, and the broker exits with this value when it wants to prevent restart.  It currently does this:
- When builtin module loading fails
- When rc1 fails
- On "state machine panic", currently triggered when a module exits abnormally (Edit: oh, not on the system instance due to `-Sbroker.module-nopanic=1`)
- On rank 0 only, when `flux shutdown` is run

In this particular case I would guess that the broker on the ooming node received a CONTROL_DISCONNECT message from its parent, causing it to post the `parent-fail` broker event, which I think causes a normal shutdown.  So the broker would restart in 30s, having logged a LOG_CRIT a "HOSTNAME (rank N) sent disconnect control message" message.

I wonder if we should add that case to the list of "norestart" exceptions above?

grondo:
Actually looking back at the discussion, the broker on the OOM node exited due to:

[...]

Issue #7223, opened 2025-12-03
restart: broker may load incorrect topology when restarting with running jobs


grondo:
On systems where user-specified options can modify the actual system topology at runtime (e.g., changing GPU partitioning modes), the broker may encounter a mismatch during restart.

Normally, the broker reads the system topology at startup when no jobs are running, so it sees the default topology. It verifies this matches its expected configuration. Jobs that modify the topology specify resource.rediscover so subinstances can reload and see the changes.

When restart with running jobs is supported, this becomes a problem: if a broker restarts while a job is running with a modified topology, it will read the modified topology from the system. The verification check will fail because the actual topology no longer matches the expected default configuration. The broker will incorrectly conclude the node is misconfigured and drain itself.

We need a solution that handles topology modifications during restart while preserving automatic detection of genuinely misconfigured nodes during verification.

Perhaps the previous topology could be cached in statedir when restarting on a live system (idea from @garlick)

garlick:
Another thought is to treat "soft restart" specially.  Set a flag in the kvs when restarting flux, and when brokers come back up, skip the resource check as long as the nodes haven't rebooted?

grondo:
Yeah, I was thinking along the same lines. However, we'd still need to cache the previous hwloc XML somewhere so that the cached version doesn't have the extra/incorrect resources.

grondo:
It turns out this is a potential issue even now: Recently, a node was run out of memory and the resulting slowness caused the broker heartbeat timeout, so the broker exited and was restarted by systemd. The job running at the time had selected CPX gpu mode, so on restart the broker got the wrong topology.

garlick:
I wonder if we are validating resources in the wrong place?    Should we move it to a prolog snippet?

PR #7237, opened 2025-12-12
shell: add oom.adjust

garlick:
Problem: recently we've encountered system OOM killing sprees on systems that are run out of memory when ROCM HIPMalloc is used by applications. Because the ROCM memory is not attributed to the job cgroup, the cgroup OOM killer isn't activated, and the system OOM killer has trouble selecting the right process.  Meanwhile Flux's missing heartbeat timer goes off and flux goes down, and secondary effects occur.  We'd like to have the system instance Flux broker stay up, make the OOM killing more efficient, and avoid killing the Flux IMP and job shell.

This improves access/documentation for some knobs that may be helpful in crafting a solution or at least an improvement.

I'll leave off a description of the actual solutions under consideration since they are not proven yet and we have other issues open on the problems.  This just exposes some knobs to play with that may be helpful.

Related issues
- #7223 
- #7227 
- #7233

### DIFF
--- a/t/t2409-sdexec.t
+++ b/t/t2409-sdexec.t
@@ -21,6 +21,9 @@ if ! test_flux_security_version 0.14.0; then
 	skip_all="requires flux-security >= v0.14, got ${FLUX_SECURITY_VERSION}"
 	test_done
 fi
+if systemctl --user show dbus | grep -q OOMPolicy=; then
+	test_set_prereq OOMPOLICY
+fi
 
 test_under_flux 2 minimal
 
@@ -307,6 +310,25 @@ test_expect_success 'sdexec can set unit AllowedCPUs to an empty bitmask' '
 	    $systemctl --user show --property AllowedCPUs \
 	        t2409-allowedcpus2.service >allowedcpus2.out &&
 	test_cmp allowedcpus2.exp allowedcpus2.out
+'
+test_expect_success 'sdexec can set unit OOMScoreAdjust' '
+	echo "OOMScoreAdjust=100" >oomscoreadjust.exp &&
+	$sdexec -r 0 \
+	    --setopt=SDEXEC_NAME="t2409-oomscoreadjust.service" \
+	    --setopt=SDEXEC_PROP_OOMScoreAdjust="100" \
+	    $systemctl --user show --property OOMScoreAdjust \
+	        t2409-oomscoreadjust.service >oomscoreadjust.out &&
+	test_cmp oomscoreadjust.exp oomscoreadjust.out
+'
+
+test_expect_success OOMPOLICY 'sdexec can set unit OOMPolicy' '
+	echo "OOMPolicy=kill" >oompolicy.exp &&
+	$sdexec -r 0 \
+	    --setopt=SDEXEC_NAME="t2409-oompolicy.service" \
+	    --setopt=SDEXEC_PROP_OOMPolicy="kill" \
+	    $systemctl --user show --property OOMPolicy \
+	        t2409-oompolicy.service >oompolicy.out &&
+	test_cmp oompolicy.exp oompolicy.out
 '
 
 memtotal() {


doc/man5/flux-config-exec.rst
```diff
@@ -103,8 +103,11 @@ testexec
 SDEXEC PROPERTIES
 =================
 
-When the sdexec service is selected, The following systemd unit properties may
-be set by adding them to the ``sdexec-properties`` table:
+When the sdexec service is selected, systemd unit properties may be set by
```
grondo, line 106: Commit message title typo: propreties

src/common/libsdexec/start.c
```diff
@@ -301,6 +314,15 @@ static int prop_add (json_t *prop, const char *name, const char *val)
...
     }
+    else if (streq (name, "OOMScoreAdjust")) {
+        errno = 0;
+        char *endptr;
+        long l = strtol (val, &endptr, 10);
+        if (errno != 0 || *endptr != '\0')
+            return -1;
+        if (prop_add_i32 (prop, name, l) < 0)
+            return -1;
+    }
```
grondo, line 325: Unrelated question: would it work to use the json type as a hint so that we don't have to explicitly add support to the code for each new, non-string systemd property? If the signedness of number is a problem, maybe an optional type hint could be supported in the string? Just something to think about for the future since systemd seems to have _a lot_ of these properties and it is a bummer if we have to support each in this way.

On topic: I think the range of OOM scores is `-1000` to `1000`, should that be checked here? Or is there a sensible enough error when a user uses a value outside that range already?

src/shell/oom.c
```diff
@@ -162,6 +163,80 @@ static struct shell_oom *oom_create (flux_shell_t *shell, flux_error_t *errp)
...
+        || !(task = flux_shell_current_task (shell))
+        || flux_shell_task_info_unpack (task, "{s:I}", "pid", &pid) < 0)
+        return shell_log_errno ("failed to get current task pid");
+
+    if (peek_int (pid, "oom_score_adj", &old_adjust) < 0)
+        return -1;
+    if (poke_int (pid, "oom_score_adj", adjust) < 0) {
+        saved_errno = errno;
+        failed = true;
+    }
```
grondo: If writing the new value doesn't cause the job to fail, should failure reading it cause the job to fail?

src/shell/oom.c
```diff
@@ -162,6 +163,80 @@ static struct shell_oom *oom_create (flux_shell_t *shell, flux_error_t *errp)
...
+    if (peek_int (pid, "oom_score", &score) < 0)
+        return -1;
+
```
grondo: Could also check that the new score equals the expected adjust score and log an error if not?

doc/man5/flux-config-exec.rst
```diff
@@ -128,6 +128,14 @@ MemorySwapMax
    Specify the absolute limit on swap used by the job.  Values are formatted as
    described above.
 
+OOMScoreAdjust
+   Sets the adjustment value for the Linux kernel's OOM killer score.
+   Values range from -1000 to 1000, with 1000 making a processes most likely
```
grondo: s/processes/process/

doc/man5/flux-config-exec.rst
```diff
@@ -128,6 +128,14 @@ MemorySwapMax
...
+   to be selected, and -1000 preventing a process from being selected.
+   See the :linux:man5:`systemd.exec` for more information.
+   Setting a negative value is likely a privileged operation the Flux systemd
+   instance.
```
grondo: Missing a 'for' or 'in'?

doc/man5/flux-config-exec.rst
```diff
@@ -128,6 +128,14 @@ MemorySwapMax
    Specify the absolute limit on swap used by the job.  Values are formatted as
    described above.
 
+OOMScoreAdjust
```
grondo, line 131: Commit message: tothe

t/t2623-job-shell-oom.t
```diff
@@ -0,0 +1,27 @@
+#!/bin/sh
+#
+test_description='Test flux-shell sysmon plugin'
```
grondo: s/sysmon/oom/

src/common/libsdexec/start.c
garlick: Well it's a little bit tricky.  For example, a json integer could map to any of these dbus integers:
```
BYTE	y (121)	Unsigned 8-bit integer
INT16	n (110)	Signed (two's complement) 16-bit integer
UINT16	q (113)	Unsigned 16-bit integer
INT32	i (105)	Signed (two's complement) 32-bit integer
UINT32	u (117)	Unsigned 32-bit integer
INT64	x (120)	Signed (two's complement) 64-bit integer (mnemonic: x and t are the first characters in "sixty" not already used for something more common)
UINT64  t (116)	Unsigned 64-bit integer
```
Hmm, maybe we could support an alternate raw value type that looks more like the internal json representation of dbus types or maps to it directly.  For example, the -1000 value here maps to
```json
["i", -1000]
```

I think when you specify a value out of range it just goes to the max in that direction so probably checking isn't needed here.  I didn't try it but IIRC we do get good systemd errors when things go wrong.

src/shell/oom.c
garlick: Hmm yeah, maybe not a great idea to always expect those values to be readable.  Will fix.

src/shell/oom.c
garlick: Well to be honest I'm a little confused and wasn't expecting what actually happens, which is the `oom_score` is adjusted by `oom_score_adj` (which of course sounds reasonable!).  Some of the documentation went into great detail about algorithms I couldn't be bothered to understand.  So not being 100% sure what's supposed to happen, or if it's supposed to happen all the time, I figured I'd just do a bit of debug logging.

BTW there is also a `choom` tool that is useful for looking at current values, e.g.
```
 garlick@picl0:~$ flux run -n1 sh -c 'choom -p $$'
pid 151723's current OOM score: 799
pid 151723's current OOM score adjust value: 200
```
```
 garlick@picl0:~$ flux run -o oom.adjust=1000 -n1 sh -c 'choom -p $$'
pid 151780's current OOM score: 1332
pid 151780's current OOM score adjust value: 1000
```

src/common/libsdexec/start.c
garlick: Two follow-on thoughts:
1) there should be an RFC for the internal JSON representation with a nice citation to the dbus schema, and if we add something for TOML, it could include that too
2) subprocess options are used to pass properties to sdexec.  It'd be nice if there were a way to set a json value.  It's encoded as json in the remote request payload.  Maybe we could add a `flux_cmd_setopt_pack()` or similar.

garlick:
Fixed all that stuff so I'll set MWP.  Thanks and sorry for the various typos!

PR #7238, opened 2025-12-12
add 5m shutdown timeout on system instance

garlick:
Problem: if a compute node becomes unresponsive, flux shutdown can hang indefinitely

Add a 5m timeout on SHUTDOWN state for the system instance only.  This is only activated on rank 0, and covers only the period where the broker is waiting for TBON children to fully stop (which includes their getting through `rc3`).  It does not cover the leader broker's `rc3`.


garlick:
I restarted the macos builder which failed some subprocess tests.

garlick:
Thanks - setting MWP

Issue #7235, opened 2025-12-10
new job-list client RPCs hang when connecting to older Flux instances


grondo:
#7224  changed the job-list RPC to stream responses by default. The implementation maintains backward compatibility when the streaming flag is not set in the request, allowing older clients to work with newer servers. However, there is a compatibility issue in the opposite direction:

When a new client connects to an older Flux instance, the client expects the response stream to terminate with an ENODATA error. Since older servers don't send this termination signal, the RPC hangs indefinitely. This prevents testing on current flux-core master against installed Flux instances.

Perhaps a flag should be added to the _initial_ response from the new job-list server that indicates there are more responses coming. Without the flag, the client code can fall back to the single response behavior.

garlick:
This might do it:
```c
bool response_is_streaming (flux_future_t *f)
{
    const flux_msg_t *msg;
    if (flux_future_get (f, (const void **)&msg) == 0 && flux_msg_is_streaming (msg))
        return true;
    return false;
}
```
Or python equivalent.

The streaming flag is normally copied from request to response, and the raw contents of the RPC future is the response message.

grondo:
Here's the trace of a request from current master to a v0.81.0 system instance:
```
--------------------------------------
> 0.00000
> job-list.list
> flags=topic,payload,route,streaming userid=unknown rolemask=none nodeid=any matchtag=1
> {"max_entries": 1001, "attrs": ["result", "userid", "t_cleanup", "queue", "nodelist", "urgency", "nnodes", "ntasks", "priority", "dependencies", "duration", "annotations", "state", "name", "t_run"], "since": 0.0, "constraint": {"and": [{"userid": [6885]}, {"states": [62]}]}}
--------------------------------------
< 0.00056
< job-list.list
< flags=topic,payload,route,streaming userid=765 rolemask=owner errnum=0 matchtag=1
< {"jobs":[]}
```
(Hangs here)

The streaming flag is copied from the request, so there is not a way I see for the client to determine it is talking to an older server. 

I could be misunderstanding where you meant to use the test above though.

grondo:
We could just add a new `ver` field to the first response. Clients could default a local version variable to 0 and use `s?i` to unpack. Then, if `version == 0` break out of the loop of expected responses. The new job-list server could set `"ver": 1`.

chu11:
I like the `version` field idea, it'll give us some added flexibility in the future too.

PR #7239, opened 2025-12-12
job-list: add version to list response

chu11:
Per discussion in #7235

Problem: Newer job-list clients assume that the job-list service is always
streaming responses.  This is not the case if the client tries to communicate
with an older job-list service.

Per RFC43 update, have newer job-list responses also send a job-list reponse
indicating the version.  This version can be used by newer clients to communicate
with both older and newer versions of the job-list protocol.

Update flux-job, flux-top, python bindings accordingly.

Associated RFC fix: job-list: add version to list response

### DIFF
--- a/src/bindings/python/flux/job/list.py
+++ b/src/bindings/python/flux/job/list.py
@@ -29,11 +29,18 @@ class JobListRPC(RPC):
 
     def get_jobs(self):
         """Returns all jobs."""
+        first = True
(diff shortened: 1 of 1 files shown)


chu11:
re-pushed, tweaking code so that the "version" is only sent in the first response

garlick (approved):
LGTM!

PR #7241, opened 2025-12-17
support `resource.verify` config for granular resource verification

### LINKED ISSUE #7220: nodes with misconfigured GPUs are not drained
On  a production system recently Flux was started on a node that was still in CPX mode, but it didn't get drained. Since the resource module caches the hwloc XML for reuse, this was causing confusion to mpibind when it used the CPX mode topology when CPX mode was not enabled. 

There is a comment in `topo_verify()` that explains a rationale for ignoring GPUs, but I wonder if that is still relevant? We do have the `noverify` config option for cases where GPUs are not properly detected.

https://github.com/flux-framework/flux-core/blob/5a70c401bda8a16439a86578f830a101b897f526/src/modules/resource/topo.c#L100-L112

grondo:
This PR intends to solve #7220 and related issues by the addition of a new `resource.verify` config , which allows per-resource configuration policies:
 - `strict`: strict resource verification
 - `allow-extra`: allow extra resources, drain on missing
 - `allow-missing`: allow missing resources, drain on extra
 - `ignore`: disable verification of this resources.
 
The configuration is specified via a `[resource.verify]` table that supports a `default` key and per-resource overrides. Currently, `core`, `gpu`, and `hostname` resources are supported. For example, the following is the default if no `resource.verify` table is provided and keeps the current behavior:
```toml
[resource.verify]
default = "allow-extra"
gpu = "ignore"
hostname = "strict"
```
As a convenience, `resource.verify` may also be a boolean, where `true` acts as `default = "strict"` and `false` as `default = "ignore"`.

As mentioned above, to preserve existing behavior, by default the resource module allows extra cores and ignores GPUs when no `resource.verify` table is provided. Otherwise, the default is `strict` (even if an empty config is provided). This seemed to be the most logical path forward.

This means to solve #7220 admins would need to add either `resource.verify = true` to the config, or being explicit:
```toml
[resource.verify]
default = "strict"
```
To allow extra cores but ensure no extra gpus:
```toml
[resource.verify]
default = "strict"
core = "allow-extra"
```

Additionally, resource verification can now be enabled for jobs, e.g.:
```
$ flux alloc --conf=resource.verify=true ...
````
This is mainly useful for testing, but it seemed appropriate that an explicit config set by the user is not ignored.

Fixes #7220

### DIFF
--- a/doc/man5/flux-config-resource.rst
+++ b/doc/man5/flux-config-resource.rst
@@ -71,9 +71,80 @@ norestrict
    when the Flux system instance is constrained to a subset of cores,
    but jobs run within this instance should have access to all cores.
 
+verify
+   (optional) Boolean or table controlling resource verification behavior at
+   broker startup. When resources are verified, the actual resources discovered
+   via HWLOC are compared against the configured resources. Verification
+   failures cause the broker rank to drain itself with an appropriate message.
+
+   If set to a boolean:
+
+   - ``true``: Enable strict verification for all resources (hostname, cores,
+     and GPUs)
+   - ``false``: Disable verification for all resources (set default="ignore")
+
+   If set to a table, each resource type can be configured with one of four
+   verification modes:
+
+   - ``"strict"``: Drain rank on any mismatch (missing or extra resources)
+   - ``"allow-extra"``: Drain only on missing resources
+   - ``"allow-missing"``: Drain only on extra resources
+   - ``"ignore"``: Don't verify this resource type
+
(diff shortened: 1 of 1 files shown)


grondo:
Thanks! I'll set MWP and perhaps @ryanday36 could give feedback and open a new issue if there needs to be adjustments or improvements in the future.

garlick:
Restarting the flux-pmix test since that seems to have failed at repo checkout

PR #7244, opened 2025-12-17
update `FLUX_CLI_PLUGINPATH` to support a colon-delimited path (multiple paths)

### LINKED ISSUE #6980: cli: support some sort of path for finding CLI plugins
Currently I think CLI plugins are only discovered from a single path `FLUX_CLI_PLUGINPATH`, which only supports a single path and which overrides the default path if set.

It would be less surprising and more flexible if `FLUX_CLI_PLUGINPATH` supported a colon delimited path (and which perhaps only appends to the default system path).

Another idea would be to follow the [XDG basedir](https://specifications.freedesktop.org/basedir-spec/latest/), i.e. search `XDG_CONFIG_HOME`/flux/cli` and `XDG_CONFIG_DIRS/flux/cli`. We do have `flux.util.xdg_searchpath()` to assist with this since it used to find named configs for the `flux alloc/batch --config=` option (for example)

wihobbs:
Problem: Currently CLI plugins are only loaded from a single path (FLUX_CLI_PLUGINPATH) which overrides the default path if set.

Support a colon-delimited list for the FLUX_CLI_PLUGINPATH and load plugins from all paths provided.

Fixes #6980 

It's perhaps surprising, with this addition, that a command can fail if a plugin that collides with its option name has already been loaded. If we're going to follow the lead of other paths like `PATH`, perhaps we should just ignore that error and refer to the first option in `FLUX_CLI_PLUGINPATH`?


grondo:
> It's perhaps surprising, with this addition, that a command can fail if a plugin that collides with its option name has already been loaded. If we're going to follow the lead of other paths like PATH, perhaps we should just ignore that error and refer to the first option in FLUX_CLI_PLUGINPATH?

Great point. I agree the first plugin with a given name found in `FLUX_CLI_PLUGINPATH` should be used. This allows one to override a system-provided plugin (e.g. as might be done in a subproject) by prepending to `FLUX_CLI_PLUGINPATH`

wihobbs:
Cool. Perhaps hold off on a review until I can fix that then :)

grondo:
Another thing to consider: Now that a real search path is supported, we might want to think about standardizing the search path to include non-/etc paths for plugins which might eventually be provided by packages (e.g. in this case probably under `/usr/libexec/flux`) See e.g. https://github.com/flux-framework/flux-core/pull/7039 for rationale and example

wihobbs:
Ok @grondo, at long last this is ready for a real human to review it. :)

grondo:
Restarted a builder that hit #7406

src/bindings/python/flux/cli/plugin.py
```diff
@@ -80,6 +81,14 @@ def help(self, option):
...
 
+    def set_path(self, path):
+        """Log where a plugin was loaded from"""
+        self._load_path = path
+
+    def get_path(self):
+        """Return where a plugin was loaded from"""
+        return self._load_path
+
     def add_option(self, name, **kwargs):
```
grondo, line 131: A getter and setter seem overkill here. Could the initializer just do:
```python
   self.path = None
```

Then you can set the path with `plugin.path = path` and get it with simple `plugin.path`

src/bindings/python/flux/cli/plugin.py
```diff
@@ -167,7 +170,25 @@ def print_help(self, name):
...
+        """
+        Return list of dirs in ``FLUX_CLI_PLUGINPATH`` if set, plus the
+        default CLI plugin search path.
+        """
+        if "FLUX_CLI_PLUGINPATH" in os.environ:
+            paths = filter(
+                lambda s: s and not s.isspace(),
+                os.environ["FLUX_CLI_PLUGINPATH"].split(":"),
+            )
+            searchpath = [path for path in paths]
```
grondo: Is this line really neessary?  The `s and not s.isspace()` in the `filter()` above already filters out empty paths. Could you just wrap the `filter()` call in `list()` instead?

src/bindings/python/flux/cli/plugin.py
```diff
@@ -167,7 +170,25 @@ def print_help(self, name):
...
+        else:
+            builtindir = conf_builtin_get("libexecdir")
+            sysdir = conf_builtin_get("confdir")
+            searchpath = [f"{builtindir}/cli/plugins", f"{sysdir}/cli/plugins"]
+        return searchpath
```
grondo: So `FLUX_CLI_PLUGINPATH` overrides the system paths. I guess that preserves existing behavior, so was a good call here.

However, since CLI plugins are relatively new, I wonder if we could make a breaking change here for the better. It seems like `FLUX_CLI_PLUGINPATH` should be _prepended_ to the builtin paths. This allows users to easily add CLI plugins without dropping access to system-supplied versions.

Of course, then there is no easy way to completely disable all system-supplied plugins, as might be useful for testing purposes.

The other alternative is a `FLUX_CLI_PLUGINPATH_PREPEND` variable. This matches existing practice of the `FLUX_EXEC_PATH_PREPEND` and `FLUX_MODULE_PATH_PREPEND` environment variables, e.g. here's the documentation for `FLUX_MODULE_PATH_PREPEND`:

```
       FLUX_MODULE_PATH_PREPEND
              FLUX_MODULE_PATH is set in the environment of the broker so that
              broker  modules  can  be  found  and  loaded  when  requested by
              flux-module(1):
                 $FLUX_MODULE_PATH_PREPEND : install-path : $FLUX_MODULE_PATH
```

What do you think?

src/bindings/python/flux/cli/plugin.py
```diff
@@ -167,7 +170,25 @@ def print_help(self, name):
...
+        if "FLUX_CLI_PLUGINPATH" in os.environ:
+            paths = filter(
+                lambda s: s and not s.isspace(),
+                os.environ["FLUX_CLI_PLUGINPATH"].split(":"),
+            )
+            searchpath = [path for path in paths]
+        else:
+            builtindir = conf_builtin_get("libexecdir")
+            sysdir = conf_builtin_get("confdir")
+            searchpath = [f"{builtindir}/cli/plugins", f"{sysdir}/cli/plugins"]
```
grondo: Side note: I wonder if `syscondir/cli/plugins` should come before `libexecdir/cli/plugins` in the search path?

src/bindings/python/flux/cli/plugin.py
```diff
@@ -180,24 +201,34 @@ def _add_plugins(self, module, program):
                 and entry != CLIPlugin
             ):
                 self.plugins.append(entry(program))
+                # using a setter instead of another argument prevents us from
+                # breaking currently used plugins
+                self.plugins[-1].set_path(path)
+
```
grondo: I'm not sure I understand this comment. What is the problem being solved here?

It might be cleaner and clearer to first create the plugin instance, set its path, then append to the plugins list, e.g.:

```python
    plugin = entry(program)
    plugin.path = path
    self.plugins.append(plugin)
```

src/bindings/python/flux/cli/plugin.py
```diff
@@ -180,24 +201,34 @@ def _add_plugins(self, module, program):
...
-                    self.options.append(option)
-                    ## keep a temporary list of "dests" to ensure no conflicts
-                    option_dests[option.kwargs["dest"]] = 1
-        return self  ## possibly unnecessary now
+        for plugin in self.plugins[:]:
+            if any(opt.kwargs["dest"] in option_dests for opt in plugin.options):
+                self.plugins.remove(plugin)
+            else:
+                for opt in plugin.options:
+                    option_dests[opt.kwargs["dest"]] = opt
```
grondo: I wonder if there's a way we could move this check to load time, and just skip loading plugins with conflicts?

If instead of just appending plugins to the `self.plugins` list, we instead had a `register(plugin, path)` method, this would allow the path to be more naturally captured, and the `dests` dictionary could be built up as plugins are added. Then another `check_conflict()` method could be added that checks if a plugin has any conflicts, and could be used like:
```python
   if not self.check_conflicts(plugin):
        self.register(plugin, path)
```

This would be cleaner IMO, and would speed up plugin loaded. However, it isn't a showstopper since the current code works. If you keep the current code, a comment explaining what's happening would be good (e.g. we iterate a copy of the list because it is being modified, etc)

grondo (commented):
Ok, leaving some comments after a first pass. 

This appears to be working, excellent! The main thing I'd like to resolve before merging is how a user would prepend their plugin to the system supplied set.

src/bindings/python/flux/cli/plugin.py
wihobbs: > What is the problem being solved here?

The comment is perhaps misleading; your solution would work fine and is cleaner. What I initially tried that would break existing plugins was to change the abstract base class to include an initializer:

```
class CLIPlugin(ABC):  # pragma no cover
    """Base class for a CLI submission plugin

    A plugin should derive from this class and implement one or more
    base methods (described below)

    Attributes:
        prog (str): command-line subcommand for which the plugin is active,
            e.g. "submit", "run", "alloc", "batch", "bulksubmit"
        prefix (str): By default, ``--ex-`` is added as a prefix to ensure
            plugin-provided options are namespaced. The prefix may be
            overridden with a site or project name.

    """

    def __init__(self, prog, path=None, prefix="ex", version=None):
```

This means all of the existing plugins would have to do work to make sure `self.path = path` was set in their initializers (as subclasses), or the later `print_plugins()` method would just print a bunch of `None`s. Perhaps I should have thought of that with the `version` argument as well...

src/bindings/python/flux/cli/plugin.py
wihobbs: Huh, interesting point. I guess I was thinking that `/etc/cli/plugins` would contain Flux "core" plugins (if any) along with sysadmin-developed plugins, and `libexecdir/cli/plugins` would contain Flux "project" plugins that might want to override the core plugins.

In hindsight, it seems like my mental map violates the Filesystem Hierarchy Standard, and any Flux "core" plugins should _not_ go in `/etc`, that should all be up to the sysadmins. In which case, I agree with you. Will fix.

src/bindings/python/flux/cli/plugin.py
wihobbs: Moreso than preserving existing behavior, I was thinking of the use case where a user doesn't want _any_ of the system provided plugins. 

I assumed anyone who wanted to both prepend to the `FLUX_CLI_PLUGINPATH` and maintain system plugins would know where those were and just include them in their environment setting. However, if we're expecting users to do that, that could go wrong very quick -- the user could end up inheriting their own environment in a subinstance and loading plugins from `/etc` that aren't relevant (or guaranteed to work in that context).

`FLUX_CLI_PLUGINPATH_PREPEND` is a long thing to type for what most users will want for default behavior (just guessing). How would you feel about this behavior @grondo:

- by default, load plugins from `sysconfdir/cli/plugins:builtindir/cli/plugins`
- if `FLUX_CLI_PLUGINPATH` is set, prepend that to the above
- allow users to set `FLUX_CLI_PLUGINPATH_IGNORE_BUILTINS` to _completely_ override the system and libexec provided plugins.

src/bindings/python/flux/cli/plugin.py
grondo: I don't feel too strongly on this one. My guess is that administrators aren't going to need or want to override installed cli plugins, but the question did pop up in my head while looking at it. so I wanted to bring it up for discussion.

We'll need to find somewhere to document this behavior. Given the existence of `flux-jobtap-plugins(7)` and `flux-shell-plugins(7)`, does it make sense to have `flux-cli-plugins(7)`?

src/bindings/python/flux/cli/plugin.py
grondo: Since a plugin could be installed anywhere, setting the path from the loader makes sense to me.

I didn't check how version is handled, but if plugins have versions, it makes sense they capture their own version, perhaps as a class property/attribute that overrides the base class.

garlick:
@wihobbs - ping? Seems like the ball is in your court here?

src/bindings/python/flux/cli/plugin.py
wihobbs: I'm adding a note on the two paths and their behavior to `flux-environment(7)`.

src/bindings/python/flux/cli/plugin.py
wihobbs: Good thoughts. I added some more comments to the existing code to clarify what's going on.

wihobbs:
I have a diff ready to fix the spellcheck and flake8 issue, but can't seem to push to github right now (I can't even `ssh -T`). Will try again later.

```diff 
diff --git a/doc/test/spell.en.pws b/doc/test/spell.en.pws
index 0d37012da..6558eb98b 100644
--- a/doc/test/spell.en.pws
+++ b/doc/test/spell.en.pws
@@ -1096,3 +1096,5 @@ ValueError
 BrokerModule
 MyModule
 subclasses
+CLI
+cli
diff --git a/src/bindings/python/flux/cli/plugin.py b/src/bindings/python/flux/cli/plugin.py
index ecc6dc55a..7011b42c6 100644
--- a/src/bindings/python/flux/cli/plugin.py
+++ b/src/bindings/python/flux/cli/plugin.py
@@ -213,7 +213,7 @@ class CLIPluginRegistry:

     def print_plugins(self):
         """Print all of the plugins loaded by _load_plugins."""
-        print(f"\033[1mOptions provided by plugins: \033[0m")
+        print("\033[1mOptions provided by plugins: \033[0m")
         print(f"  Search path: {':'.join(self._get_searchpath())}\n")
         for plugin in self.plugins:
             print(f"{type(plugin).__name__} loaded from {plugin.path}")
```

src/bindings/python/flux/cli/base.py
```diff
@@ -1018,6 +1021,11 @@ def create_parser(
             action="store_true",
             help="Do not print jobid to stdout on submission",
         )
+        parser.add_argument(
+            "--list-plugins",
+            action="store_true",
+            help="Print all command-line plugins that are loaded and exit",
+        )
         parser.add_argument(
```
grondo, line 580: I think the addition of `--list-plugins` should be in its own commit.

src/bindings/python/flux/cli/plugin.py
```diff
@@ -167,7 +162,40 @@ def print_help(self, name):
...
+        default CLI plugin search path.
+        """
+
+        sysdir = conf_builtin_get("confdir")
+        builtindir = conf_builtin_get("libexecdir")
+        builtin_paths = [f"{sysdir}/cli/plugins", f"{builtindir}/cli/plugins"]
+
+        plugin_path = os.environ.get("FLUX_CLI_PLUGINPATH") or os.environ.get(
+            "FLUX_CLI_PLUGINPATH_NO_BUILTINS"
+        )
```
grondo: If `FLUX_CLI_PLUGINPATH` is an empty string it is falsy so I don't think this would produce expected results. I'm not sure what happens if `FLUX_CLI_PLUGINPATH_NO_BUILTINS` is also empty (i.e. disable all plugins case) May want to add tests specifically for these cases.

BTW, here's a suggested replacement function from Claude:
```python
  def _get_searchpath(self):
      """
      Return list of dirs to search for CLI plugins.

      If ``FLUX_CLI_PLUGINPATH_NO_BUILTINS`` is set, return only those
      paths (system defaults are suppressed entirely).

      Otherwise, prepend any paths from ``FLUX_CLI_PLUGINPATH`` to the
      system default search path.
      """
      sysdir = conf_builtin_get("confdir")
      builtindir = conf_builtin_get("libexecdir")
      builtin_paths = [f"{sysdir}/cli/plugins", f"{builtindir}/cli/plugins"]

      if "FLUX_CLI_PLUGINPATH_NO_BUILTINS" in os.environ:
          raw = os.environ["FLUX_CLI_PLUGINPATH_NO_BUILTINS"]
          return [s for s in raw.split(":") if s and not s.isspace()]

      paths = list(builtin_paths)
      if "FLUX_CLI_PLUGINPATH" in os.environ:
          extra = [
              s for s in os.environ["FLUX_CLI_PLUGINPATH"].split(":")
              if s and not s.isspace()
          ]
          paths = extra + paths

      return paths
```

grondo: Also add tests for when both PLUGINPATH variables are set. Ensure that `FLUX_CLI_PLUGINPATH_NO_BUILTINS` wins.

src/bindings/python/flux/cli/plugin.py
```diff
@@ -179,25 +207,42 @@ def _add_plugins(self, module, program):
...
             ):
-                self.plugins.append(entry(program))
+                plugin = entry(program)
+                plugin.path = path
+                self.plugins.append(plugin)
+
+    def print_plugins(self):
+        """Print all of the plugins loaded by _load_plugins."""
+        print("\033[1mOptions provided by plugins: \033[0m")
+        print(f"  Search path: {':'.join(self._get_searchpath())}\n")
```
grondo: Calls `_get_searchpath()` unnecessarily, could you use `self.pluginpath`?

src/bindings/python/flux/cli/plugin.py
```diff
@@ -179,25 +207,42 @@ def _add_plugins(self, module, program):
...
                 and entry != CLIPlugin
             ):
-                self.plugins.append(entry(program))
+                plugin = entry(program)
+                plugin.path = path
+                self.plugins.append(plugin)
+
+    def print_plugins(self):
+        """Print all of the plugins loaded by _load_plugins."""
+        print("\033[1mOptions provided by plugins: \033[0m")
```
grondo: Should probably drop the escape sequences here, or optionally check for stdout `isatty()`.

t/cli-plugins/extras/twooptions.py
```diff
@@ -0,0 +1,47 @@
...
+# Copyright 2025 Lawrence Livermore National Security, LLC
+# (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+#
+# This file is part of the Flux resource manager framework.
+# For details, see https://github.com/flux-framework.
+#
+# SPDX-License-Identifier: LGPL-3.0
+##############################################################
+
+from flux.cli.plugin import CLIPlugin
```
grondo, line 11: Regarding testing: We need some way to ensure the testsuite isn't polluted by system plugins. Does running from the build tree automatically do this? (I haven't tested) or do we need to set `FLUX_CLI_PLUGINPATH_NO_BUILTINS` globally in the testsuite (replacing existing `FLUX_CLI_PLUGINPATH` setting)

src/bindings/python/flux/cli/plugin.py
```diff
@@ -167,7 +162,40 @@ def print_help(self, name):
...
+        Return list of dirs in ``FLUX_CLI_PLUGINPATH`` if set, plus the
+        default CLI plugin search path.
+        """
+
+        sysdir = conf_builtin_get("confdir")
+        builtindir = conf_builtin_get("libexecdir")
+        builtin_paths = [f"{sysdir}/cli/plugins", f"{builtindir}/cli/plugins"]
+
+        plugin_path = os.environ.get("FLUX_CLI_PLUGINPATH") or os.environ.get(
+            "FLUX_CLI_PLUGINPATH_NO_BUILTINS"
```
grondo: Suggestion: I wonder if `FLUX_CLI_PLUGINPATH_OVERRIDE` would be a better name for this variable instead of `NO_BUILTINS`

doc/man7/flux-environment.rst
```diff
@@ -567,6 +567,24 @@ MISCELLANEOUS
...
 
+.. envvar:: FLUX_CLI_PLUGINPATH
+
+   Prepend additional search paths (colon delimited) for command-line 
+   plugins to base submission commands such as :man1:`flux-alloc`. The 
+   system default search paths will continue to be searched, but after 
+   all search paths provided to ``FLUX_CLI_PLUGINPATH``. The default
+   system search path is:: 
+
+     $libexecdir/cli/plugins:$sysconfdir/cli/plugins
```
grondo: There was trailing whitespace on many of these lines in my working copy, though I don't see them here in the github interface.

grondo (commented):
Nice! This seems really close! I had some comments inline.

t/cli-plugins/extras/twooptions.py
wihobbs: No, the build tree includes two default directories. That's a good point, we can override with `FLUX_CLI_PLUGINPATH_NO_BUILTINS=`. (Once the falsy-ness you noted above is fixed)

wihobbs: @grondo I think I did this right in [`592f8d4`](https://github.com/flux-framework/flux-core/pull/7244/commits/592f8d4e717f17966b65e7b87cb6b5971bfaaf80), but please review.

src/bindings/python/flux/cli/plugin.py
grondo: Considered this suggestion?

t/cli-plugins/extras/twooptions.py
grondo: It looks good, but it might be more readable as `FLUX_CLI_PLUGINPATH_NO_BUILTINS=""` to be explicit that we're setting to an empty string, i.e. empty path. Also, I still feel like `OVERRIDE` instead of `NO_BUILTINS` is perhaps a better name since this is overriding processing of `FLUX_CLI_PLUGINPATH` as well as disabling builtins, and also it is a path and `NO_BUILTINS` sounds more like it is a variable you should set to `1` or something.

grondo: Also, now that this is set globally for sharness, do you need to unset it in the cli plugin test? Maybe that's why most of the tests are failing now?

src/bindings/python/flux/cli/base.py
grondo: This is still an issue.

src/bindings/python/flux/cli/plugin.py
```diff
@@ -227,25 +252,42 @@ def _add_plugins(self, module, program):
...
-                    )
-                else:
-                    self.options.append(option)
-                    ## keep a temporary list of "dests" to ensure no conflicts
-                    option_dests[option.kwargs["dest"]] = 1
-        return self  ## possibly unnecessary now
+        # because the plugin list can change due to conflicting options, iterate
+        # over a copy, not the list itself
+        for plugin in self.plugins[:]:
+            if any(opt.kwargs["dest"] in option_dests for opt in plugin.options):
```
grondo: Could use `opt.dest` here now.

src/bindings/python/flux/cli/plugin.py
```diff
@@ -227,25 +252,42 @@ def _add_plugins(self, module, program):
...
+                # remove the current plugin if any of the options conflict (since
+                # it would have come after the 'primary' plugin observed in the
+                # PATH)
+                self.plugins.remove(plugin)
+            else:
+                for opt in plugin.options:
+                    option_dests[opt.kwargs["dest"]] = opt
```
grondo: Use `opt.dest` here.

t/t2717-python-cli-plugins.t
```diff
@@ -71,12 +71,34 @@ test_expect_success 'flux-run: validate-only plugins are properly validated' '
...
 	grep -e "I am a valid plugin" \
-	     -e "Option for setting AMD SMI compute partitioning" help1.out
+	     -e "Option for setting AMD SMI compute" help1.out
+'
+test_expect_success 'flux-batch: --list-plugins shows only plugins with no option conflicts' '
+	flux batch --list-plugins >> tmp_help_real.out &&
+	grep "loaded from" tmp_help_real.out | wc -l | grep -qx 5
+'
+test_expect_success 'flux-batch: FLUX_CLI_PLUGINPATH_NO_BUILTINS wins a fight against FLUX_CLI_PLUGINPATH' '
+	FLUX_CLI_PLUGINPATH_NO_BUILTINS= FLUX_CLI_PLUGINS=${SHARNESS_TEST_SRCDIR}/cli-plugins/extras/ \
```
grondo: Typo? `FLUX_CLI_PLUGINS` -> `FLUX_CLI_PLUGINPATH`

doc/man7/flux-environment.rst
```diff
@@ -567,6 +567,24 @@ MISCELLANEOUS
...
 
+.. envvar:: FLUX_CLI_PLUGINPATH
+
+   Prepend additional search paths (colon delimited) for command-line
+   plugins to base submission commands such as :man1:`flux-alloc`. The
+   system default search paths will continue to be searched, but after
+   all search paths provided to ``FLUX_CLI_PLUGINPATH``. The default
+   system search path is::
+
+     $libexecdir/cli/plugins:$sysconfdir/cli/plugins
```
grondo: Is this correct? The code seems to set `sysconfdir` first.

src/bindings/python/flux/cli/plugin.py
wihobbs: I don't see a `dest` attribute when I `dir()` a `CLIPluginOption`, just `_unprefixed_dest` and the original `kwargs`, which are still set by the class initializer:

```
$ FLUX_CLI_PLUGINPATH=$(pwd)/../t/cli-plugins ./src/cmd/flux alloc --help
> /usr/src/src/bindings/python/flux/cli/plugin.py(242)_load_plugins()
-> if option.kwargs["dest"] in option_dests:
(Pdb) dir(option)
['__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__getstate__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', '_unprefixed_dest', 'kwargs', 'name']
(Pdb) option.dest
*** AttributeError: 'CLIPluginOption' object has no attribute 'dest'
```
(My current branch is rebased up to PR #7514)

We could trivially add that to the class, though, and it seems like a good idea. I'll do that.

grondo: It is provided by a `@property` decorator to avoid the need to synchronize with `opt.kwargs`

doc/man7/flux-environment.rst
wihobbs: Ope forgot about this one. Fixing now.

src/bindings/python/flux/cli/plugin.py
wihobbs: I hadn't really given it full consideration until the reasoning in your comment above, which makes sense. Updated.

doc/man7/flux-environment.rst
wihobbs: Fixed.

t/t2717-python-cli-plugins.t
```diff
@@ -4,8 +4,12 @@ test_description='Test command line plugin interface'
 
 . $(dirname $0)/sharness.sh
 
+unset FLUX_CLI_PLUGINPATH_OVERRIDE
+
 export FLUX_CLI_PLUGINPATH=${SHARNESS_TEST_SRCDIR}/cli-plugins/
 
+env | grep FLUX_CLI > /tmp/hello
+
```
grondo: Leftover debug?

doc/man7/flux-environment.rst
```diff
@@ -567,6 +567,24 @@ MISCELLANEOUS
...
+   system search path is::
+
+     $sysconfdir/cli/plugins:$libexecdir/cli/plugins
+
+.. envvar:: FLUX_CLI_PLUGINPATH_NO_BUILTINS
+
+   Override the entire search path for command-line plugins, including
+   ``$sysconfdir/cli/plugins`` and ``$libexecdir/cli/plugins``. Only a
+   colon-delimited list of paths passed to this variable will be
+   observed when this variable is set.
```
grondo: Change to `FLUX_CLI_PLUGINPATH_OVERRIDE`?

t/sharness.d/flux-sharness.sh
```diff
@@ -5,6 +5,8 @@
 
 # add scripts directory to path
 export PATH="${SHARNESS_TEST_SRCDIR}/scripts:$PATH"
+# ensure no builddir plugins are picked up globally
+export FLUX_CLI_PLUGINPATH_OVERRIDE=""
```
grondo, line 9: Commit message needs to be updated to refer to `FLUX_CLI_PLUGINPATH_OVERRIDE` not `FLUX_CLI_PLUGINPATH_NO_BUILTINS`.

grondo (approved):
This LGTM! Just saw some doc/commit message errors and one test issue.
Once those are addressed and tests are passing this one can go in!

wihobbs:
Thanks, @grondo! I noticed that when I run the testsuite, it changes the permissions of the existing tests:

```
$ git status
On branch cli-plugin-path
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   t/cli-plugins/gpumode.py
	modified:   t/cli-plugins/noprefix.py
	modified:   t/cli-plugins/twooptions.py
	modified:   t/cli-plugins/validateonly.py

jupiter13:flux-core hobbs17$ git diff
diff --git a/t/cli-plugins/gpumode.py b/t/cli-plugins/gpumode.py
old mode 100644
new mode 100755
diff --git a/t/cli-plugins/noprefix.py b/t/cli-plugins/noprefix.py
old mode 100644
new mode 100755
diff --git a/t/cli-plugins/twooptions.py b/t/cli-plugins/twooptions.py
old mode 100644
new mode 100755
diff --git a/t/cli-plugins/validateonly.py b/t/cli-plugins/validateonly.py
old mode 100644
new mode 100755
```
So I went ahead and updated those permissions and added a note to the testing commit.

grondo:
I wonder if we should look into that. Wouldn't want permission changes occurring for installed plugins. That could cause command failures for regular users

wihobbs:
In that case, I'm just going to drop that permission change from this PR (the testsuite was passing fine before hand without it) and can look into what's going on in another issue. (Edit: I'm pretty sure this was a local issue to my git repo.)

grondo:
Ah, if it is a local-only issue then no problem! I do wonder what's happening though.

PR #7246, opened 2025-12-17
broker: add args param to topology_create()

garlick:
Problem: the `custom` topology gets the config `hosts` array by accessing a global variable set by `topology_hosts_set()` which is inconsistent with the design of the topology class.

Add a JSON `args` argument to `topology_create()` that allows additional arguments of any sort to be provided (NULL OK).

Provide the `hosts` array through `args`.

Update broker users.
Update unit tests.

### DIFF
--- a/src/broker/topology.c
+++ b/src/broker/topology.c
@@ -39,15 +39,19 @@ struct topology {
 
 static int kary_plugin_init (struct topology *topo,
                              const char *path,
+                             json_t *args,
                              flux_error_t *error);
 static int mincrit_plugin_init (struct topology *topo,
                                 const char *path,
+                                json_t *args,
                                 flux_error_t *error);
 static int binomial_plugin_init (struct topology *topo,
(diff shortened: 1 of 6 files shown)


src/broker/topology.c
```diff
@@ -566,9 +570,9 @@ static int custom_plugin_init (struct topology *topo,
         errprintf (error, "unknown custom topology directive: '%s'", path);
         return -1;
     }
-    if (!boot_hosts)
-        return 0;
-    json_array_foreach (boot_hosts, rank, entry) {
+    if (!args || json_unpack (args, "{s:o}", "hosts", &hosts) < 0)
+        return 0; // defaults OK if no hosts (e.g. singleton)
```
grondo, line 574: This is an internal function, so optional, but might want to check that `hosts` is an array?

garlick:
Fixed that and added a test for it. Will set MWP - thanks!

PR #7247, opened 2025-12-18
liibpmi: pull in business card classes from the broker

garlick:
Problem: in order to move the broker overlay subsystem to a broker module, some refactoring work is necessary in the bootstrap area.  This could be simplified if some of the bootstrap code could be isolated from the broker code base.

Move the "bizcard" and "bizcache" classes to libpmi.

Bizcard is a container for bootstrap-related data stored in the PMI KVS.

Bizcache is a read cache of bizcards that simplifies fetching them and helps avoid extra PMI KVS operations (which might be slow) when different parts of a client require the same business cards.

### DIFF
--- a/src/common/libpmi/test/bizcard.c
+++ b/src/common/libpmi/test/bizcard.c
@@ -11,6 +11,7 @@
 #if HAVE_CONFIG_H
 #include "config.h"
 #endif
+#include <jansson.h>
 #include <flux/core.h>
 
 #include "ccan/str/str.h"
@@ -53,6 +54,7 @@ void test_simple (void)
     struct bizcard *bc;
     struct bizcard *bc2;
     const char *s;
+    const json_t *o;
 
     ok ((bc = bizcard_create ("hostname", "pubkey")) != NULL,
         "bizcard_create works");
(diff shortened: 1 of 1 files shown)


src/common/libpmi/bizcard.h
```diff
@@ -20,6 +21,11 @@ void bizcard_decref (struct bizcard *bc);
 const char *bizcard_encode (const struct bizcard *bc);
 struct bizcard *bizcard_decode (const char *s, flux_error_t *error);
 
+/* N.B. bizcard_tojson() doesn't lend a reference (do not json_decref()!).
+ */
+struct bizcard *bizcard_fromjson (json_t *obj);
+json_t *bizcard_tojson (const struct bizcard *bc);
```
grondo: Unlikely to be a real issue, but in most other places `_tojson()` functions return a `json_t *` with a reference, so this could be confusing. If the function didn't have the `to` preposition, it might be less likely to cause any confusion. Suggestion, rename function something like `bizcard_get_json()` (and perhaps return `const json_t *` as a reminder?)

grondo (approved):
LGTM! Just had one minor comment.

garlick:
Thanks - I made the suggested changes (and added a test for it).  Will set MWP.

PR #7249, opened 2025-12-18
libpmi: add config bootstrap method

garlick:
Problem: some bootstrap refactoring is necessary to support moving the overlay subsystem to a broker module.

The determination of rank and size needs to happen in the broker proper, so attributes can be set that are passed to the earliest loaded modules.  The discovery of peer endpoints can happen later, after the reactor starts in an overlay broker module.

Ideally we would call `upmi_initialize()` to determine the rank and size from the broker proper, and then later support RPCs that let overlay continue the bootstrap process.  To make that refactoring easier, it's nice if `upmi_initialize()` works for both regular PMI and config files, since that can be easily pulled out of `boot_pmi.c` and `boot_config.c` without forcing major changes to that code all at once.

This PR just adds the "config" capability to libupmi.  A future PR will refactor the broker code and use this.

Edit: p.s. this currently has no effect, despite being placed first in the list of methods for upmi to try.  That is because `upmi_create()` isn't currently called when the `[bootstrap]` config table exists.

### DIFF
--- a/src/common/libpmi/Makefile.am
+++ b/src/common/libpmi/Makefile.am
@@ -74,6 +74,10 @@ test_ldadd = \
 	$(LIBRT) \
 	$(LIBDL)
 
+test_cflags = \
+	-Wno-unused-value \
+	$(AM_CFLAGS)
+
 test_cppflags = \
 	-I$(top_srcdir)/src/common/libtap \
 	-DUPMI_TEST_SEARCHPATH=\"$(builddir)/test/plugin/.libs\" \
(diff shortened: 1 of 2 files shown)


src/common/libpmi/upmi.c
```diff
@@ -394,6 +412,22 @@ static int upmi_preinit (struct upmi *upmi,
             goto nomem;
         }
     }
+    if (args) {
+        const char *key;
+        json_t *value;
+        json_object_foreach (args, key, value) {
```
grondo, line 425: Should `args` be checked for `json_is_object()` here just in case? (probably fine as-is if this is internal use only)

grondo (approved):
LGTM!

garlick:
Thanks!  OK, fixed that and setting MWP

PR #7252, opened 2025-12-19
add CLIPlugin to accept arbitrary R shapes/JSON on the command-line 

wihobbs:
This PR takes the excellent work done by @sam-maloney for [RFC#46](github.com/flux-framework/rfc/pull/465) and extends it into a command-line plugin that users can opt in to by setting `FLUX_CLI_PLUGINPATH`. This was [originally suggested by @grondo](https://github.com/flux-framework/rfc/pull/465#issuecomment-3201763842) as a way for us to test/get user feedback on the ergonomics of specifying shape on the command line, and make progress toward incorporating that option into our standard command line.

The plugin is relatively straightforward and the testing just references examples from the corresponding RFC. I know this is coming up right before many 🌴s, so no rush on reviews, but I wanted to get it to a stopping point before my own holiday begins :).

### DIFF
--- a/src/bindings/python/flux/Makefile.am
+++ b/src/bindings/python/flux/Makefile.am
@@ -89,6 +89,9 @@ nobase_fluxpy_PYTHON = \
 	constraint/parser.py \
 	constraint/parsetab.py \
 	constraint/__init__.py \
+	shape/parser.py \
+	shape/parsetab.py \
+	shape/__init__.py \
 	abc/journal.py \
 	abc/__init__.py \
 	utils/parsedatetime/__init__.py \
@@ -130,10 +133,12 @@ endif
 
 
 BUILT_SOURCES = \
-	constraint/parsetab.py
+	constraint/parsetab.py \
+	shape/parsetab.py
 
 CLEANFILES = \
 	constraint/parser.out \
+	shape/parser.out \
 	$(BUILT_SOURCES)
 
 STDERR_DEVNULL = $(stderr_devnull_$(V))
@@ -151,8 +156,14 @@ constraint/parsetab.py: constraint/parser.py
 	  $(PYTHON) $< --outputdir $(builddir)/constraint $(STDERR_DEVNULL) && \
 	  touch $@
 
+shape/parsetab.py: shape/parser.py
+	$(AM_V_GEN)$(MKDIR_P) shape && \
+	  $(PYTHON) $< --outputdir $(builddir)/shape $(STDERR_DEVNULL) && \
+	  touch $@
+
 clean-local:
 	-rmdir constraint 2>/dev/null || :
+	-rmdir shape 2>/dev/null || :
 	-rm -f *.pyc */*.pyc *.pyo */*.pyo
 	-rm -rf __pycache__ */__pycache__
 
--- a/src/bindings/python/flux/Makefile.am
+++ b/src/bindings/python/flux/Makefile.am
@@ -26,6 +26,7 @@ nobase_fluxpy_PYTHON = \
 	cli/fortune.py \
 	cli/plugin.py \
 	cli/plugins/__init__.py \
+	cli/plugins/shape.py \
 	core/__init__.py \
 	core/watchers.py \
 	core/inner.py \
--- a/src/bindings/python/flux/cli/plugins/shape.py
+++ b/src/bindings/python/flux/cli/plugins/shape.py
@@ -1,0 +1,60 @@
+##############################################################
+# Copyright 2025 Lawrence Livermore National Security, LLC
+# (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+#
+# This file is part of the Flux resource manager framework.
+# For details, see https://github.com/flux-framework.
+#
+# SPDX-License-Identifier: LGPL-3.0
+##############################################################
+
+import json
+
+from flux.cli.plugin import CLIPlugin
+from flux.shape.parser import ShapeParser
+
+
+class ShapePlugin(CLIPlugin):
+    """Wrap the functionality provided by the command-line job shape
+    parser into a command-line plugin users can optionally prepend to
+    their `FLUX_CLI_PLUGINPATH`. As a convenience for users, also
+    provide options for specifying json files on the command-line
+    that will get absorbed into the submitted jobspec.
+    """
+
+    def __init__(self, prog, prefix="resources"):
+        super().__init__(prog, prefix=prefix)
+        self.add_option(
+            "--shape",
+            metavar="SHAPE",
+            help="Provide an RFC 46 jobspec shape on the command line. Any other resource arguments will be ignored.",
+        )
+        self.add_option(
+            "--json",
+            metavar="FILE",
+            help="Provide a JSON file specifying the `resources` section of a jobspec.",
+        )
+
+    def preinit(self, args):
+        try:
+            if args.json or args.shape:
+                args.nodes = 1  # set a number of slots, will be subsequently ignored
+        except AttributeError:
+            pass  # without args.json or args.shape set, do nothing
+
+    def modify_jobspec(self, args, jobspec):
+        if not (getattr(args, "shape", None) or getattr(args, "json", None)):
+            return
+        if getattr(args, "shape", None) and getattr(args, "json", None):
+            raise ValueError(
+                "`--resources-shape` and `--resources-json` are mutually exclusive arguments. Use only one."
+            )
+        if getattr(args, "shape", None):
+            jobspec.jobspec["resources"] = ShapeParser().parse(args.shape)
+        elif getattr(args, "json", None):
+            with open(args.json, "r") as json_file:
+                data = json.load(json_file)
+                try:
+                    jobspec.jobspec["resources"] = data["resources"]
+                except TypeError:
+                    jobspec.jobspec["resources"] = data
--- a/src/bindings/python/flux/shape/parser.py
+++ b/src/bindings/python/flux/shape/parser.py
@@ -1,0 +1,679 @@
+##############################################################
+# Copyright 2025 Lawrence Livermore National Security, LLC
+# (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+#
+# This file is part of the Flux resource manager framework.
+# For details, see https://github.com/flux-framework.
+#
+# SPDX-License-Identifier: LGPL-3.0
+##############################################################
+
+import argparse
+import json
+
+import ply.lex as lex
+import ply.yacc as yacc
+import yaml
+
+
+class ShapeSyntaxError(Exception):
+    """
+    Specialized SyntaxError exception to allow ShapeParser to throw
+    a SyntaxError without PLY trying to force recovery.
+
+    """
+
+    pass
+
+
+class ShapeLexer(object):
+    """
+    Simple job shape jobspec resources syntax lexical analyzer based on RFC 46.
+    Used mainly as the lexer for BaseShapeParser
+    """
+
+    #  Different quoting states for single vs double quotes:
+    states = (
+        ("squoting", "exclusive"),
+        ("dquoting", "exclusive"),
+        ("count", "exclusive"),
+        ("dictkey", "exclusive"),
+        ("dictvalue", "exclusive"),
+    )
+
+    tokens = (
+        "TYPE",
+        "SLOT",
+        "OPERATOR",
+        "IDSET",
+        "INTEGER",
+        "KEY",
+        "FALSE",
+        "NULL",
+        "TRUE",
+        "NUMBER",
+        "STRING",
+        "QSTRING",
+        "ARRAY",
+        "LBRACKET",
+        "RBRACKET",
+        "LBRACE",
+        "RBRACE",
+    )
+
+    literals = ",;:/+-"
+
+    # Ignore whitespace in default state
+    t_ignore = " \t\r\n\f\v"
+
+    # Tokens in 'quoting' state
+    t_squoting_ignore = ""
+    t_dquoting_ignore = ""
+    t_count_ignore = t_ignore
+    t_dictkey_ignore = t_dictvalue_ignore = t_ignore
+
+    def __init__(self, **kw_args):
+        super().__init__()
+        self.lexer = lex.lex(module=self, **kw_args)
+        self.bracket_level = 0
+        self.last_lbracket = 0
+        self.last_rbracket = 0
+        self.brace_level = 0
+        self.last_lbrace = 0
+        self.last_rbrace = 0
+        self.slot_count = 0
+        self.quote_type = None
+        self.quote_start = None
+        self.quote_empty = True
+        self.array_start = None
+
+    def input(self, data):
+        self.lexer.push_state("INITIAL")
+        self.bracket_level = 0
+        self.last_lbracket = 0
+        self.last_rbracket = 0
+        self.brace_level = 0
+        self.last_lbrace = 0
+        self.last_rbrace = 0
+        self.slot_count = 0
+        self.quote_type = None
+        self.quote_start = None
+        self.quote_empty = True
+        self.array_start = None
+        self.lexer.input(data)
+
+    def __getattr__(self, attr):
+        return getattr(self.lexer, attr)
+
+    def t_ANY_error(self, t):
+        raise ShapeSyntaxError(
+            f"Illegal character '{t.value[0]}' at position {t.lexer.lexpos}"
+        )
+
+    # Order of functions is important. Define special tokens earlier to have
+    # them take precedence.
+    # c.f. http://www.dabeaz.com/ply/ply.html#ply_nn6
+
+    def t_squoting_dquoting_eof(self, t):
+        raise ShapeSyntaxError(
+            f"Unclosed quote {self.quote_type} at position {self.quote_start}"
+        )
+
+    def t_squoting_QSTRING(self, t):
+        r"[^']+"
+        self.quote_empty = False
+        return t
+
+    def t_dquoting_QSTRING(self, t):
+        r'[^"]+'
+        self.quote_empty = False
+        return t
+
+    def t_squoting_QUOTE(self, t):
+        r"'"
+        return self.end_quote_helper(t)
+
+    def t_dquoting_QUOTE(self, t):
+        r'"'
+        return self.end_quote_helper(t)
+
+    def end_quote_helper(self, t):
+        t.lexer.pop_state()
+        if self.quote_empty:
+            t.value = ""
+            t.type = "QSTRING"
+        # If not empty, QSTRING token(s) were already returned,
+        # don't return the closing quote
+        else:
+            t = None
+        return t
+
+    # This must come after quoting states, so it only lexes starting quote
+    def t_ANY_QUOTE(self, t):
+        r"\"|'"
+        self.quote_start = t.lexer.lexpos
+        self.quote_empty = True
+        self.quote_type = t.value
+        if t.value == "'":
+            t.lexer.push_state("squoting")
+        else:
+            t.lexer.push_state("dquoting")
+
+    def t_TYPE(self, t):
+        r"[^={}[\]/;\s\"\']+"
+        if t.value.lower() == "slot":
+            t.value = "slot"
+            t.type = "SLOT"
+            self.slot_count += 1
+        return t
+
+    def t_EQUALS(self, t):
+        r"="
+        t.lexer.push_state("count")
+
+    def t_count_OPERATOR(self, t):
+        r":[+*^]"
+        return t
+
+    def t_count_IDSET(self, t):
+        r"[1-9]\d*((?=,)|-[1-9]\d*(?=,))(,[1-9]\d*(-[1-9]\d*)?)*"
+        return t
+
+    def t_count_INTEGER(self, t):
+        r"[1-9]\d*"
+        t.value = int(t.value)
+        return t
+
+    def t_count_end(self, t):
+        r"(?=[{/;])"
+        t.lexer.pop_state()
+
+    def t_dictkey_KEY(self, t):
+        r"(?![+-])[^:,{}]+"
+        if t.value == "x":
+            t.value = "exclusive"
+        return t
+
+    def t_dictkey_COLON(self, t):
+        r":"
+        t.lexer.push_state("dictvalue")
+        t.type = ":"
+        return t
+
+    def t_dictvalue_ARRAY(self, t):
+        r"\[.*]"
+        t.lexer.pop_state()
+        t.value = json.loads(t.value.replace("'", '"'))
+        return t
+
+    def t_dictvalue_NUMBER(self, t):
+        r"-?(0|[1-9]\d*)(\.\d*)?((e|E)(-|\+)?\d+)?(?=[},])"
+        try:
+            t.value = int(t.value)
+        except Exception:
+            t.value = float(t.value)
+        return t
+
+    def t_dictvalue_STRING(self, t):
+        r"[^,{}[\]]+"
+        if t.value.lower() == "false":
+            t.value = False
+            t.type = "FALSE"
+        elif t.value.lower() == "null":
+            t.value = None
+            t.type = "NULL"
+        elif t.value.lower() == "true":
+            t.value = True
+            t.type = "TRUE"
+        return t
+
+    def t_dictvalue_COMMA(self, t):
+        r","
+        t.lexer.pop_state()
+        t.type = ","
+        return t
+
+    def t_dictvalue_RBRACE(self, t):
+        r"}"
+        t.lexer.pop_state()
+        return self.t_ANY_RBRACE(t)
+
+    def t_ANY_LBRACE(self, t):
+        r"{"
+        self.brace_level += 1
+        self.last_lbrace = t.lexer.lexpos - 1
+        t.lexer.push_state("dictkey")
+        return t
+
+    def t_ANY_RBRACE(self, t):
+        r"}"
+        self.brace_level -= 1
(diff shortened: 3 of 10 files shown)


wihobbs:
@sam-maloney was kind enough to do a quick review on 1/8/26 and commented in Slack:

> I made some minor modifications to the commit in my branch to hopefully address the Python linting errors and CodeQL problems, so you can incorporate that easily enough. (@wihobbs: ✅)
> 
> 1. As for testing, my thinking was that if it could handle all the examples in RFC14 then that was a reasonably good start for testing valid coverage. I appreciate that you've added some invalid test cases, and I'm sure edge cases will come up as they always do, but I don't have much specific to suggest for now.
> 2. And as for naming, I have no issue at all with removing "CLI". I just picked something, and the CLI was what I was thinking about at the time, but a short form certainly could be useful in other contexts, so a more general name would be great. Whether that is "shape" or "compressed_resources" or "rstring", etc. doesn't really matter to me, depending on whether it is preferred to have it short or verbose!

src/bindings/python/flux/shape/plugin/shape.py
```diff
@@ -0,0 +1,59 @@
...
+
+class RFC44ShapePlugin(CLIPlugin):
+    """Wrap the functionality provided by the command-line job shape
+    parser into a command-line plugin users can optionally prepend to
+    their `FLUX_CLI_PLUGINPATH`. As a convenience for users, also
+    provide options for specifying json files on the command-line
+    that will get absorbed into the submitted jobspec.
+    """
+
+    def __init__(self, prog, prefix="r"):
```
grondo: Prefix of `--r-` here could be confusing since we use R to refer to the [RFC 20](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_20.html) Resource Set Specification. I'd suggest a prefix of `resources` since this overrides the `resources` section of jobspec.

t/t2718-python-cli-job-shape.t
```diff
@@ -0,0 +1,129 @@
+#!/bin/sh
+
+test_description='Test command line plugin for RFC 46 job shape'
+
+. $(dirname $0)/sharness.sh
+
```
grondo, line 6: Should also add some tests of the `flux python -m flux.shape.parser [OPTIONS] -- EXPRESSION` interface to the shape parser. They could probably just go in here.

grondo (commented):
Nice work! Just a few comments or questions to address inline. Thanks!

t/t2718-python-cli-job-shape.t
grondo: I also noticed the new tests are not added to `t/Makefile.am` so they aren't being run in CI.

src/bindings/python/flux/Makefile.am
```diff
@@ -139,8 +144,13 @@ constraint/parsetab.py: constraint/parser.py
...
 	  touch $@
 
+shape/parsetab.py: shape/parser.py
+	$(AM_V_GEN)$(MKDIR_P) shape && \
+	  $(PYTHON) $< --outputdir $(builddir)/shape $(STDERR_DEVNULL) && \
+	  touch $@
+
 clean-local:
-	-rmdir constraint 2>/dev/null || :
+	-rmdir shape 2>/dev/null || :
```
grondo, line 167: `-rmdir constraint` was inadvertently removed here.

src/bindings/python/flux/shape/plugin/shape.py
```diff
@@ -0,0 +1,59 @@
...
+# SPDX-License-Identifier: LGPL-3.0
+##############################################################
+
+import json
+
+from flux.cli.plugin import CLIPlugin
+from flux.shape.parser import ShapeParser
+
+
+class RFC44ShapePlugin(CLIPlugin):
```
grondo: Rename `RFC46ShapePlugin`

src/bindings/python/flux/shape/plugin/shape.py
```diff
@@ -0,0 +1,59 @@
...
+            help="Provide an RFC 46 jobspec shape on the command line. Any other resource arguments will be ignored.",
+        )
+        self.add_option(
+            "--json",
+            metavar="FILE",
+            help="Provide a JSON file specifying the `resources` section of a jobspec.",
+        )
+
+    def preinit(self, args):
+        if getattr(args, "json") or getattr(args, "shape"):
```
grondo: `getattr(args, "json")` is equivalent to `args.json` but the latter is shorter and more idiomatic. I'd switch this to simply `if args.json or args.shape`

src/bindings/python/flux/shape/plugin/shape.py
```diff
@@ -0,0 +1,59 @@
...
+            args.nodes = 1  ## set a number of slots, will be subsequently ignored
+
+    def modify_jobspec(self, args, jobspec):
+        s = getattr(args, "shape")
+        j = getattr(args, "json")
```
grondo: probably easier and more readable to just drop this and use `args.shape` and `args.json` below.

src/bindings/python/flux/shape/plugin/shape.py
```diff
@@ -0,0 +1,59 @@
...
+        if s:
+            jobspec.jobspec["resources"] = ShapeParser().parse(s)
+        elif j:
+            with open(j, "r") as json_file:
+                data = json.loads(json_file.read())
+                try:
+                    jobspec.jobspec["resources"] = data["resources"]
+                except AttributeError:
+                    jobspec.jobspec["resources"] = data
+        else:
```
grondo: I think `AttributeError` should be `KeyError` and maybe `TypeError` here. `KeyError` is raised if `data` is a dictionary but `resources` is not a member, and `TypeError` is raised if `data` is an array (I think the expected case) and it is indexed as a dictionary:
```python
>>> data = [{"foo": 42}]
>>> data["resources"]
  File "<python-input-1>", line 1, in <module>
    data["resources"]
    ~~~~^^^^^^^^^^^^^
TypeError: list indices must be integers or slices, not str
```

In fact, maybe you only want `TypeError` here?

t/python/t0041-job-shape-parser.py
```diff
@@ -0,0 +1,277 @@
+#!/usr/bin/env python3
+###############################################################
```
grondo, line 2: Commit message subject mentions `CliJobspecParser` but this is testing `RFC46ShapePlugin`

grondo: Actually this tests `ShapeParser` nothing to do with the CLI jobspec parser plugin.

t/Makefile.am
```diff
@@ -241,6 +241,7 @@ TESTSCRIPTS = \
 	t2715-python-cli-cancel.t \
 	t2716-python-cli-batch-conf.t \
 	t2717-python-cli-plugins.t \
+	t2718-python-cli-job-shape.t \
```
grondo, line 256: Commit message says `--r-shape` `--r-json` but the options are `--resources-shape` and `--resources-json`. Could still have a compact commit title with `t: add tests for jobspec shape cli plugin`

t/t2718-python-cli-job-shape.t
```diff
@@ -0,0 +1,129 @@
...
+	]
+	EOF
+	flux run --resources-shape=slot=4/node --dry-run hostname | jq .resources >  output1.json &&
+	diff -s exp1.json output1.json | grep "Files exp1.json and output1.json are identical"
+'
+
+test_expect_success 'flux-submit: job with invalid shape is rejected' '
+	test_must_fail flux alloc --resources-shape=slot/ echo hello 2> err.out &&
+	grep "flux-alloc: ERROR" err.out
+'
```
grondo: Test title: s/flux-submit/flux-alloc/

t/t2718-python-cli-job-shape.t
```diff
@@ -0,0 +1,129 @@
...
+	      {
+	        "type": "node",
+	        "count": 1
+	      }
+	    ]
+	  }
+	]
+	EOF
+	flux run --resources-shape=slot=4/node --dry-run hostname | jq .resources >  output1.json &&
+	diff -s exp1.json output1.json | grep "Files exp1.json and output1.json are identical"
```
grondo: I'm not sure what `diff -s` is getting over just using `diff`, but possibly you ran into something? Usually we'd just use the sharness `test_cmp` here. (which simply calls `diff -u`)

However, there is a potential issue with comparing two JSON files with `diff`, since the order keys are output could make the files differ.

My suggestion would be to pass the expected and actual output through `jq --sort-keys` and then use `test_cmp`., e.g.
```sh
cat <<-EOF | jq -S . > exp1.json
...
EOF
...
flux run --resources-shape=slot=4/node --dry-run hostname | jq -S .resources >  output1.json
test_cmp exp1.json output1.json
```

t/t2718-python-cli-job-shape.t
```diff
@@ -0,0 +1,129 @@
...
+	          "count": 2
+	        }
+	      ]
+	    }
+	  ],
+	}	
+	EOF
+	test_must_fail flux submit --resources-json=r_set2.json hostname 2> output4.err &&
+	grep "flux-submit: ERROR: Expecting " output4.err
+'
```
grondo, line 123: Add mutual exclusion test?

grondo (commented):
It is getting to the point where @sam-maloney's `--resource-shape` option will be really useful so I made another pass on this one.

I'm also wondering if we should install the plugin so it is picked up by default, instead of requiring users to modify their plugin path to get the feature.

garlick:
This would indeed be handy for toy scheduler experiments.  Ping?

wihobbs:
... I have no idea why this closed.

Edit: p.s. this is not ready for another review, a few more modifications needed, but will be soon.

src/bindings/python/flux/shape/plugin/shape.py
wihobbs: Went with an even more general `ShapePlugin`.

src/bindings/python/flux/shape/plugin/shape.py
wihobbs: You're right. I had in my head I was thinking that a resource shape without a `resources` key would be a dict, but that doesn't make any sense (and if it did, we'd want `KeyError`.)

t/t2718-python-cli-job-shape.t
wihobbs: I'm updating all the tests to use this syntax, except for test 6: `flux-submit: providing invalid json gets rejected`, because piping invalid json to `jq` will result in the tool raising the error, and I want to be sure to test that the CLI plugin can raise the error.

t/t2718-python-cli-job-shape.t
wihobbs: Oops this shouldn't have been resolved as I still need to harden the tests for `flux python -m flux.shape.parser [OPTIONS] -- EXPRESSION`.

wihobbs:
This is ready for another review. I would like to get an ack from @sam-maloney before we merge, particularly on the testing and [501c14e](github.com/flux-framework/flux-core/pull/7252/commits/501c14eb66341d036ad8dde2ef4bb22bb5d4b017), since it modifies his contribution and was Claude-generated. On balance, it looks reasonable to me, but Sam is the expert on the parser. If he's good with that, I will fold it up into the initial commit.

As for 
> I'm also wondering if we should install the plugin so it is picked up by default, instead of requiring users to modify their plugin path to get the feature.

that will depend on #7244, which I plan to go work on now.

src/bindings/python/flux/shape/plugin/shape.py
```diff
@@ -0,0 +1,59 @@
...
+            jobspec.jobspec["resources"] = ShapeParser().parse(s)
+        elif j:
+            with open(j, "r") as json_file:
+                data = json.loads(json_file.read())
+                try:
+                    jobspec.jobspec["resources"] = data["resources"]
+                except TypeError:
+                    jobspec.jobspec["resources"] = data
+        else:
+            pass
```
grondo: Is the
```python
    else:
        pass
```
unnecessary here?

Also, just a suggestion: The function would be easier to read if it used `args.shape` and `args.json` directly instead of assigning to temporary local variables `s` and `j`.

wihobbs:
I was able to reproduce the el8 issue locally:
```python
Generating LALR tables
  File "/usr/lib64/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib64/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/src/src/bindings/python/flux/shape/parser.py", line 666, in <module>
    print(yaml.dump(result, sort_keys=False))
  File "/usr/lib64/python3.6/site-packages/yaml/__init__.py", line 200, in dump
    return dump_all([data], stream, Dumper=Dumper, **kwds)
TypeError: dump_all() got an unexpected keyword argument 'sort_keys'
not ok 13 - flux.shape.parser --yaml produces yaml output
#
#		cat >expected10 <<-EOF &&
#		- type: slot
#		  count: 4
#		  label: task
#		  with:
#		  - type: node
#		    count: 1
#
#		EOF
#		$parser --yaml -- slot=4/node >output10 &&
#		test_cmp expected10 output10
#

```

According to the [pyyaml changelog](https://pyyaml.org/wiki/PyYAML#history), support was added for the `sort_keys` option in v5.1. Unfortunately, the most recent [rocky linux 8 package is 3.12](https://pkgs.org/download/python3-pyyaml). 

I tried removing that key, but it does change the output:
```diff
--- expected10	2026-03-31 15:20:23.384955001 +0000
+++ output10	2026-03-31 15:20:23.489955001 +0000
@@ -1,7 +1,6 @@
-- type: slot
-  count: 4
+- count: 4
   label: task
+  type: slot
   with:
-  - type: node
-    count: 1
+  - {count: 1, type: node}
```

Perhaps that's not so bad, and would most likely be consistent across all distros? We could always add the key back in one glorious day when we finally no longer are tied to Python 3.6.8...

wihobbs:
We could vendor pyyaml-5.1, or build it ourselves with pip. It appears to support python 3.6 according to the same changelog. However, if we do something to the el8 container, we'll have to remember and find a similar solution for all the TOSS 4 systems.

grondo:
I don't think we'd want to vendor a python package just for testing purposes.

Perhaps there is a workaround for sorting the output without `sort_keys`. A search online suggests using `collections.OrderedDict` before dumping to get sorted output.

wihobbs:
@sam-maloney, barring the one outstanding test failure, would you mind taking a whirl through this and giving your thoughts? :)

wihobbs:
Ok, rebased again. The first two commits of this PR deal with outstanding issues in the CLIPlugin interface that would be required to merge. I'll go make a separate PR for those in a minute, since they're a separate issue from including a `--resource-shape` option.

This PR now includes:

- tests passing on el8 container and compatibility with the ancient pyyaml there (locally 🤞)
- `--resource-shape` and `--resource-json` included by default as residents of `/etc/cli/plugins` in the repo, but instaled to `$libexecdir/cli/plugins` on an installation
- and some mild shuffling of files to enable ^.

wihobbs:
Interestingly, `flux-sched`'s CI appears to be catching a legitimate error not caught by core:
```
  flux-submit: ERROR: 'types.SimpleNamespace' object has no attribute 'resources_shape'
  ```

The plugin may not be checking to make sure that a job without `--resources-shape` or `--resources-json` is ignored, or perhaps there's a subtlety I missed there.

File: etc/cli/plugins/shape.py [hunk 14]
```diff
@@ -0,0 +1,55 @@
...
+    provide options for specifying json files on the command-line
+    that will get absorbed into the submitted jobspec.
+    """
+
+    def __init__(self, prog, prefix="resources"):
+        super().__init__(prog, prefix=prefix)
+        self.add_option(
+            "--shape",
+            metavar="SHAPE",
+            help="Provide an RFC 44 jobspec shape on the command line. Any other resource arguments will be ignored.",
```
sam-maloney:
```suggestion
            help="Provide an RFC 46 jobspec shape on the command line. Any other resource arguments will be ignored.",
```

etc/cli/plugins/shape.py
```diff
@@ -0,0 +1,55 @@
...
+        )
+        self.add_option(
+            "--json",
+            metavar="FILE",
+            help="Provide a JSON file specifying the `resources` section of a jobspec.",
+        )
```
sam-maloney, line 36: Could be added later, but since RFCs 14 & 25 give all their examples as YAML it might make sense to also have an option for a YAML file as input.

t/t0039-shape-parser.t
```diff
@@ -0,0 +1,235 @@
...
+	[{"type": "node", "count": 1, "ports": [8080, 8081]}]
+	EOF
+	$parser -- "node{ports:[8080,8081]}" >output26 &&
+	test_cmp expected26 output26
+'
+test_expect_success 'flux.shape.parser shorthand x expands to exclusive' '
+	cat >expected27 <<-EOF &&
+	[{"type": "node", "count": 1, "exclusive": true}]
+	EOF
+	$parser -- "node{x:true}" >output27 &&
```
sam-maloney:
```suggestion
	$parser -- "node{x}" >output27 &&
```
`true` should be implied if no value is given, and explicit flags are already tested with `node{key:true,flag:false}` a few test prior, so this should hopefully help with coverage to hit [L559](https://app.codecov.io/gh/flux-framework/flux-core/pull/7252?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=flux-framework#92375911843bba91d45474f27c792b19-R559)
(This is also probably the most idiomatic use case, as the whole point of this special flag is to keep everything as short as possible!)

t/t0039-shape-parser.t
```diff
@@ -0,0 +1,235 @@
...
+	test_cmp expected31 output31
+'
+test_expect_success 'flux.shape.parser handles open range with operand' '
+	cat >expected32 <<-EOF &&
+	[{"type": "slot", "count": {"min": 3, "max": Infinity, "operand": 2, "operator": "+"}, "label": "task", "with": [{"type": "node", "count": 1}]}]
+	EOF
+	$parser --rangedict -- "slot=3+:2/node" >output32 &&
+	test_cmp expected32 output32
+'
+test_done
```
sam-maloney, line 254: A couple of possible extra things to test for complete coverage:

- brackets around the count value, e.g. `node=[2]` to hit [L461](https://app.codecov.io/gh/flux-framework/flux-core/pull/7252?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=flux-framework#92375911843bba91d45474f27c792b19-R461)
- empty dict, e.g. `node{}` to hit [L552](https://app.codecov.io/gh/flux-framework/flux-core/pull/7252?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=flux-framework#92375911843bba91d45474f27c792b19-R522)
- slot with other dict items, e.g. `slot{task,foo:bar}/node` to hit [L425-L427](https://app.codecov.io/gh/flux-framework/flux-core/pull/7252?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=flux-framework#92375911843bba91d45474f27c792b19-R425)
- invalid token, e.g. `node=;` to hit [L345](https://app.codecov.io/gh/flux-framework/flux-core/pull/7252?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=flux-framework#92375911843bba91d45474f27c792b19-R345)
- unclosed quote, e.g. `'node` to hit [L118](https://app.codecov.io/gh/flux-framework/flux-core/pull/7252?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=flux-framework#92375911843bba91d45474f27c792b19-R118)
- illegal character, e.g. `node=a` to hit [L109](https://app.codecov.io/gh/flux-framework/flux-core/pull/7252?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=flux-framework#92375911843bba91d45474f27c792b19-R109)

The first two (count with brackets and empty dict) could simply be worked into existing test cases as they don't change the output in any way, but could also be added as separate cases to keep things as atomic as possible, whichever you prefer.

src/bindings/python/flux/shape/parser.py
```diff
@@ -0,0 +1,678 @@
...
+
+    def end_quote_helper(self, t):
+        t.lexer.pop_state()
+        if self.quote_empty:
+            t.value = ""
+            t.type = "QSTRING"
+            return t
+        # If not empty, QSTRING token(s) were already returned,
+        # don't return the closing quote
+        return None
```
sam-maloney:
```suggestion
        else:
            # If not empty, QSTRING token(s) were already returned,
            # don't return the closing quote
            t = None
        return t
```
I'm not really sure why the security-bot is complaining, but perhaps this is more idiomatic? No strong preference either way on my end, but thanks for catching this edge case!

sam-maloney (approved):
Thanks for all your work on this @wihobbs! Sorry I was on vacation last week and it took me a while to make a review, but it looks great to me, with only a handful of very minor comments :smile:

etc/cli/plugins/shape.py
wihobbs: I've made that mistake a bunch. I don't know why in my head canon, the shape RFC is 44 🤦

etc/cli/plugins/shape.py
wihobbs: I think I'll defer this to a later PR, but good suggestion. (The reason is that our only yaml2json converter, albeit trivial would require writing more testing to get it through and I'd like to land this sooner rather than later)

t/t0039-shape-parser.t
wihobbs: Bah! Silly Claude!

src/bindings/python/flux/shape/parser.py
wihobbs: I think the security bot is complaining about something else we changed ages ago. I like your suggested change better, will add.

t/t0039-shape-parser.t
wihobbs: These are excellent suggestions, thanks! I'll add all of them.

wihobbs:
@grondo, do you want to take another look at this for a secondary approval/change request?

t/t2718-python-cli-job-shape.t
```diff
@@ -0,0 +1,143 @@
+#!/bin/sh
+
+test_description='Test command line plugin for RFC 46 job shape'
+
+. $(dirname $0)/sharness.sh
+
+test_under_flux 4 job
```
grondo: Do these tests require a 4 node instance. If they do require a Flux instance, could you get away with 1 rank?

grondo:
I noticed one quick thing. Will circle back again later to check this out again.

t/t2718-python-cli-job-shape.t
wihobbs: I like keeping the first test around, because I think it's always good to test that there isn't some logic in the plugin that fails if you _don't_ use it, but you're right -- no need for a 4 node instance. Will swap to 1.

garlick:
@wihobbs - I know you've been tasked with some testing due early next week, but at a lower priority than that, it would be great to get this PR finalized for our next tag.

wihobbs:
This is awaiting a review from @grondo who I believe is also juggling some high-priority items, but we agreed the same in our 1-1 this week. I expect we'll make the May release cut off. ;)

grondo:
Was just taking a look at this. @wihobbs I guess you get the honor of fixing one last merge conflict in `t/Makefile.am`

src/bindings/python/flux/cli/plugins/shape.py
```diff
@@ -0,0 +1,61 @@
...
+        try:
+            if args.shape and args.json:
+                raise ValueError(
+                    "`--resources-shape` and `--resources-json` are mutually exclusive arguments. Use only one."
+                )
+            if args.shape:
+                jobspec.jobspec["resources"] = ShapeParser().parse(args.shape)
+            elif args.json:
+                with open(args.json, "r") as json_file:
+                    data = json.loads(json_file.read())
```
grondo: Could use `json.load(json_file)` here.

src/bindings/python/flux/cli/plugins/shape.py
```diff
@@ -0,0 +1,61 @@
...
+                    try:
+                        jobspec.jobspec["resources"] = data["resources"]
+                    except TypeError:
+                        jobspec.jobspec["resources"] = data
+        except AttributeError:
```
grondo: This `except` catches `AttributeError` from anywhere in this entire block, including an error in `ShapeParser()` or `json.loads` and swallows it. The `try/except` block could be more targeted here, or maybe more simply just put an early exit at the top:
```python
if not (getattr(args, "json", None) or getattr(args, "shape", None)):
      return
# Now one of args.json or args.shape is non-None
```

grondo (approved):
LGTM! A couple minor things inline.

Going forward, I wonder if we should remove the default_plugins list and instead just load all plugins found in the `flux.cli.plugin` namespace, since that's what is done with other plugin loading schemes in Flux and (I think?) with FLUX_CLI_PLUGINPATH.

wihobbs:
Thanks @grondo! Setting MWP.

wihobbs:
>I guess you get the honor of fixing one last merge conflict in t/Makefile.am

Spoke too soon. There was yet another conflict. In the same file, on the same line.

garlick:
Yay!

Issue #5201, opened 2023-05-24
feature tracking: Advanced Reservations (DATs)


grondo:
This is a tracking issue for an implementation of DATs. The requirements as I understand them include:
 - an interface for specifying a set of resources (as a resource spec or perhaps specific resources) that are reserved for a specific user set in a specified time range. (This might be a Fluxion-only interface, since the simple scheduler has no actual schedule)
 - Allow the prescribed user set to submit jobs _before_ the reservation

 - [ ] flux-framework/flux-sched#963
 - [ ] flux-framework/flux-sched#1013
 - [ ] allow early interaction with instances that will be started as part of a future job (DAT) (optional)
 - [ ] #5531
 - [ ] restrict user set with access to DAT job
 - [ ] new job submission utility which allows submission of DAT/reserved job with deferred start time, restricted user list, etc.
 - [ ] accounting support for DAT jobs

grondo:
Linked flux-framework/flux-sched#1013 above. According to @trws and @milroy once that PR is merged, we will have much of the support needed in Fluxion to schedule a DAT.

garlick:
Idea: add a new job state RESERVED between SCHED and RUN such that a  job request with a special attribute could get its alloc response _R_ from the scheduler early, in advance of the the `starttime` field in _R_,.  The job manager and the rest of flux could just treat that like any other allocation, except the job would remain in RESERVED state until `starttime` arrives.  With _R_ stored in the KVS, the `sched.hello` protocol could throw it back to the scheduler on a restart.

This would work for any job, including a sub-instance.

An advance static _R_ would be more susceptible to having resources go bad before the job starts.  With a flux instance, we could initially just set the quorum value to some fraction of the total and let the instance start with some non-critical nodes down.

grondo:
For clarity, is the benefit of having the RESERVED state and advance _R_ available so that an subinstance could be configured with the eventual resources assigned to the job? Would this also have some benefit for normal jobs? Adding a new state just for that purpose feels like it could be short sighted (though I'm probably missing the other benefits!), especially if we plan one day to support instances that can grow onto unknown resources instead of just known resources.

[...]

PR #7253, opened 2025-12-23
doc: update migrating from slurm section of admin guide

garlick:
@vsoch had some great suggestions on Slack for the admin guide "Migrating from Slurm"  section of the admin guide.

I took those suggestions and riffed on them a bit.  I think this is a little bit better that what is there now?

### DIFF
--- a/doc/guide/admin_slurm.rst
+++ b/doc/guide/admin_slurm.rst
@@ -94,22 +94,26 @@ Flux has a solid security design
   :ref:`background_security`.
 
 Flux doesn't need a step manager
-  Each batch job or allocation is a new (single user) instance of Flux
-  with full capability and configurability, and a :term:`step` is really
-  a :term:`job` in Flux.  Limitations of Slurm's step manager have been
-  a source of long standing problems.
+  In Slurm, a *job* is a resource allocation scheduled by the Slurm scheduler.
+  Within the allocation, a *job step* is a unit of work scheduled by the the
+  Slurm step manager. Since a new Flux instance can be launched as a
+  :term:`job` in another Flux instance (subdividing its resources), there is
+  no need have a step abstraction or a step manager.  All units of work are
+  jobs scheduled and launched by the same robust system.  Limitations of
+  Slurm's step manager have been a source of long standing problems.
 
 Flux APIs are not an afterthought
   Well thought out Python and C APIs make it easy to integrate Flux with
-  workflow systems and new environments.
+  workflow systems and new environments.  Flux APIs are licensed under the
+  non-viral LGPL-3.0.  Experimental bindings for Go and Julia are also
+  available.
 
 Flux has a rich resource representation
   Flux's design incorporates a graph-based resource model that is
-  extensible to arbitrary resource types.  Caveat: not yet fully implemented.
+  extensible to arbitrary resource types.
 
 Flux is dynamic
-  Flux supports growing and shrinking allocations.  Caveat: not yet fully
-  implemented.
+  The Flux design supports growing and shrinking allocations.
 
 Flux is scalable
   Flux's recursive launch design enables each allocation to scale
@@ -117,16 +121,19 @@ Flux is scalable
   scalability comparable to Slurm.  But a cluster will typically be running
   many instances of Flux compared to one instance of Slurm.
 
-Flux encourages experimentation
-  In addition to extensibility via plugins (see :ref:`background_components`),
-  unprivileged users can trivially launch reconfigured or modified single-user
-  Flux instances in system resource allocations.
-
-Flux uses reactive messaging
+Flux uses event-driven messaging
   In contrast to Slurm's multi-threaded, monolithic server design, Flux
-  is built upon distributed message brokers and reactive (asynchronous)
-  agents that communicate only with messages.  Building distributed services
-  on this substrate is interesting, fun, and scales well.
+  is built upon distributed message brokers and event-driven (asynchronous,
+  reactive) agents that communicate only with messages.  Building distributed
+  services on this substrate is interesting, fun, and scales well.
+
+Flux is portable, modular, and composable
+  A single-user Flux instance is trivial to start anywhere, without
+  administrative privilege.  It can run as a parallel job in an allocation
+  of any workload manager, standalone on your laptop or Raspberry Pi, and can
+  be integrated into converged and cloud environments.  A single-user Flux
+  instance is controlled by its owner and may be reconfigured, extended, or
+  modified at will.  Slurm's monolithic design cannot offer this freedom.
 
 Slurm Long Term Viability
 =========================


doc/guide/admin_slurm.rst
```diff
@@ -15,41 +15,69 @@ Check back with us at the end of 2026.
...
-- Although Flux has an optional local fair-share accounting system for
-  gathering usage data and setting job priorities, there is not yet support
-  for a site-wide, aggregated accounting system.
-
-- No commercial support or training options.
+The Flux project began around 2012.  Flux has been used for a decade or more
+for managing complex ensembles and workflows at LLNL under Slurm and LSF
+in situations where the traditional workload managers were not up to the task.
+Although Flux was designed from the beginning to replace Slurm and was
+deployed as such on several small systems at LLNL, it did not gain momentum
```
vsoch, line 22: Here "momentum" means user-ship at the lab, or wider knowledge about it, fixing all the bugs, or some combination of the above? I think momentum is probably a good word choice, and I primarily want to understand it.

doc/guide/admin_slurm.rst
```diff
@@ -15,41 +15,69 @@ Check back with us at the end of 2026.
...
+as a system workload manager until 2024 with the early deliveries of
+`El Capitan <https://hpc.llnl.gov/hardware/compute-platforms/el-capitan>`_.
+
+Flux is now in daily production use as the sole system workload manager on
+El Capitan (currently in slot 1 of the `TOP500 <https://www.top500.org/>`_)
+and its unclassified sister systems at LLNL.  These machines are capability
+workhorses, continuously in demand for LLNL's most cutting edge,
```
vsoch, line 29: I like this part quite a bit - "capability workhorses"

doc/guide/admin_slurm.rst
```diff
@@ -15,41 +15,69 @@ Check back with us at the end of 2026.
...
+mission-critical activities.  The process of standing them up brought the
+Flux team together to address problems of scale and stability with
+unprecedented urgency.  As a result, system deployments of Flux on
```
vsoch, line 32: I really appreciate this sentence because it is exactly what happened. All the instances of essentially "Flux is on fire" :fire: at some inopportune time and then seeing all the devs come together (and threads with hundreds of messages) were experiences I won't forget.

doc/guide/admin_slurm.rst
```diff
@@ -15,41 +15,69 @@ Check back with us at the end of 2026.
...
+  *exactly* the same flux-core version.  When flux-core 1.0.0 is released,
+  relaxed rules will be enacted to support rolling upgrades.
+
+Flux restart with running jobs
+  Restarting Flux kills running jobs.  The design to allow running
+  jobs to continue is not yet fully implemented.
+
+Reservations
+  Flux does not yet have a way to request allocations that begin at a fixed
+  time in the future.  Arrangements for dedicated application time currently
```
vsoch: I asked this in slack (will ask here again). When I was reading the qmanager code base I remembered there is the concept of a reservation - e.g., https://github.com/flux-framework/flux-sched/blob/563a4c14de62beb4af72165c60ca9a85df22d2b8/qmanager/policies/queue_policy_bf_base_impl.hpp#L54 Is this one of the "implemented in sched, not exposed via flux-core yet" things?

doc/guide/admin_slurm.rst
```diff
@@ -15,41 +15,69 @@ Check back with us at the end of 2026.
...
+Flux is now in daily production use as the sole system workload manager on
+El Capitan (currently in slot 1 of the `TOP500 <https://www.top500.org/>`_)
+and its unclassified sister systems at LLNL.  These machines are capability
+workhorses, continuously in demand for LLNL's most cutting edge,
+mission-critical activities.  The process of standing them up brought the
+Flux team together to address problems of scale and stability with
+unprecedented urgency.  As a result, system deployments of Flux on
+virtually any size system are viable at this point.
+
+This experience did bring missing features in to focus that have now been
```
vsoch:
```suggestion
This experience brought missing features in to focus that have now been
```

doc/guide/admin_slurm.rst
```diff
@@ -15,41 +15,69 @@ Check back with us at the end of 2026.
...
+  jobs to continue is not yet fully implemented.
+
+Reservations
+  Flux does not yet have a way to request allocations that begin at a fixed
+  time in the future.  Arrangements for dedicated application time currently
+  require manual/scripted actions by system administrators.
+
+Flux graph-based scheduling scalability
+  Efforts to optimize the Fluxion scheduler graph-based resource
+  representation for storage space efficiency are ongoing.  Meanwhile a
```
vsoch, line 55: For this point, is it the case that the graph-based resource representation doesn't work, period, or just for very large systems?

doc/guide/admin_slurm.rst
```diff
@@ -15,41 +15,69 @@ Check back with us at the end of 2026.
...
+  flat resource model similar to Slurm's that scales well but is limited
+  in capability is typically used on large systems.
+
+Preservation of job step information
+  In Flux, there are no job steps.  Instead, a sub-instance of Flux is started
```
vsoch, line 62: (Putting on my non-admin, possibly newer user hat).  :tophat: What is a job step?

doc/guide/admin_slurm.rst
```diff
@@ -15,41 +15,69 @@ Check back with us at the end of 2026.
...
+  on a resource allocation, and user applications are run as jobs within the
+  sub-instance.  There is little coupling between the system instance and its
+  sub-instances which improves scalability.  Unfortunately, sub-instance job
+  data is not preserved unless the user explicitly arranges for it, which some
```
vsoch, line 66: Sub-instance data - what is this? You are saying that when I `flux run` under a nested child instance, I can't easily access kvs / logs and similar unless I write to file?

As a side note - if you have an example of using kvs to write to the top level parent instance of a user job (so it is accessible after) that would be hugely helpful for flux-hierarchy. I'm using the filesystem now for state and I'd rather use kvs.

doc/guide/admin_slurm.rst
```diff
@@ -15,41 +15,69 @@ Check back with us at the end of 2026.
...
+  find surprising.
+
+Multi-system fair share accounting
+  Flux has an optional fair-share accounting system for gathering usage data
```
vsoch, line 70: Can (should) we link flux-accounting here?

doc/guide/admin_slurm.rst
```diff
@@ -15,41 +15,69 @@ Check back with us at the end of 2026.
...
+  and setting job priorities.  Unlike Slurm, there is not yet a capability to
+  deploy one accounting system for multiple Flux systems.
+
+Package availability
+  Flux system administrators are encouraged to install Flux as system packages,
+  but pre-built packages are not widely available.  Source RPM packages for
+  RHEL 8 and 9 are manually attached to GitHub releases, and test debs can be
```
vsoch: Can we also say that Flux can be installed from source (I have gazillions of examples for this, for VMs across clouds, for containers, and of course Kubernetes views). I would also mention spack and conda-forge.

doc/guide/admin_slurm.rst
```diff
@@ -15,41 +15,69 @@ Check back with us at the end of 2026.
...
+  built from the source trees of several Flux Framework projects.  Automated
+  package production and Linux distribution relationships are under discussion.
```
vsoch: Under discussion makes it sound like it may _not_ happen. I think it's more something we aspire/want to do, we just need to figure out the right way to do it. Can we say something like planned? On the roadmap for the coming year? Two years? Underway? Likely in the future?

doc/guide/admin_slurm.rst
```diff
@@ -15,41 +15,69 @@ Check back with us at the end of 2026.
...
+
+Commercial support and training
+  This is an ongoing area of discussion.
```
vsoch:
```suggestion
  This is an important and current area of discussion.
```
Using "ongoing" makes it sound like... it will just keep going. :battery:

doc/guide/admin_slurm.rst
```diff
@@ -63,39 +91,46 @@ Flux has a solid security design
   :ref:`background_security`.
 
 Flux doesn't need a step manager
-  Each batch job or allocation is a new (single user) instance of Flux
-  with full capability and configurability, and a :term:`step` is really
-  a :term:`job` in Flux.  Limitations of Slurm's step manager have been
-  a source of long standing problems.
+  In Slurm, a *job* is a resource allocation scheduled by the Slurm scheduler.
+  Within the allocation, a *job step* is a unit of work scheduled by the the
```
vsoch, line 98: We might move this up (the definition of step) to when steps are first mentioned. Or just put in both spots, redundancy is OK I think.

doc/guide/admin_slurm.rst
```diff
@@ -63,39 +91,46 @@ Flux has a solid security design
...
+  Slurm step manager. Since a new Flux instance can be launched as a
+  :term:`job` in another Flux instance (subdividing its resources), there is
+  no need have a step abstraction or a step manager.  All units of work are
+  jobs scheduled and launched by the same robust system.  Limitations of
+  Slurm's step manager have been a source of long standing problems.
```
vsoch, line 103: How is this distinct from other job stages like epilog or prolog? I know (think) Slurm has those too. Do we want to say something?

doc/guide/admin_slurm.rst
```diff
@@ -63,39 +91,46 @@ Flux has a solid security design
...
 
 Flux APIs are not an afterthought
   Well thought out Python and C APIs make it easy to integrate Flux with
```
vsoch, line 106:
```suggestion
  Well thought out Python, C, and Go APIs make it easy to integrate Flux with
```

doc/guide/admin_slurm.rst
```diff
@@ -63,39 +91,46 @@ Flux has a solid security design
...
   workflow systems and new environments.
 
 Flux has a rich resource representation
   Flux's design incorporates a graph-based resource model that is
-  extensible to arbitrary resource types.  Caveat: not yet fully implemented.
+  extensible to arbitrary resource types.
```
vsoch, line 113: This is much better I think.

doc/guide/admin_slurm.rst
```diff
@@ -63,39 +91,46 @@ Flux has a solid security design
...
 
 Flux is dynamic
-  Flux supports growing and shrinking allocations.  Caveat: not yet fully
-  implemented.
+  The Flux design supports growing and shrinking allocations.
```
vsoch, line 116: +1

doc/guide/admin_slurm.rst
```diff
@@ -63,39 +91,46 @@ Flux has a solid security design
...
 
 Flux is scalable
   Flux's recursive launch design enables each allocation to scale
   independently.  Each instance of Flux has job throughput and node count
   scalability comparable to Slurm.  But a cluster will typically be running
   many instances of Flux compared to one instance of Slurm.
```
vsoch, line 122: If I can get flux-hierarchy optimized, we can at least report a throughput in the thousands. Math says we should get much higher than that, but math is limited by me. :)

doc/guide/admin_slurm.rst
```diff
@@ -63,39 +91,46 @@ Flux has a solid security design
...
 
-Flux encourages experimentation
-  In addition to extensibility via plugins (see :ref:`background_components`),
-  unprivileged users can trivially launch reconfigured or modified single-user
-  Flux instances in system resource allocations.
-
-Flux uses reactive messaging
+Flux uses event-driven messaging
```
vsoch, line 124: Yes! :raised_hands:

doc/guide/admin_slurm.rst
```diff
@@ -63,39 +91,46 @@ Flux has a solid security design
...
+
+Flux encourages experimentation
+  A single-user Flux instance can be started standalone or within a resource
+  allocation on any system without administrator privileges.  Furthermore,
+  your Flux instance can be reconfigured, extended with plugins, or modified.
+  If you want to add a new Flux service, implement a new scheduler plugin,
+  wire your workflow system directly into Flux's cerebral cortex, or just
+  litter the code with print statements to understand it better, you are
+  encouraged to do so!  This freedom cannot be offered by monolithic systems
+  that run mostly as root.
```
vsoch, line 137: What about something about converged environments, or cloud? What I say in talks is something along the lines of how the dynamism / elasticity, event-driven architecture and scalability of Flux make it amenable to cloud setups, and we can (and have) deployed across clouds for Kubernetes, VMs, and as components in cloud services. "Flux doesn't care where you run it" essentially.

doc/guide/admin_slurm.rst
garlick: Not sure if it's the right word, but I meant to refer to the high quality feedback we started getting from sys admins and users once they were forced to engage because their ability to use/maintain this system depended on being able to use flux effectively.  That started a really strong feedback loop.

Before El Cap I think interest was more casual and despite many attemts to engage sys admins in LC (for example) we didn't get very far with that because they were busy with other things and just not that interested.

doc/guide/admin_slurm.rst
garlick: Yeah, I think that's right - it might be mainly tooling that we need.  I forget but here's a thread to pull
https://github.com/flux-framework/flux-core/issues/5201

doc/guide/admin_slurm.rst
garlick: No it just doesn't scale due to the massive  size of JGF. Schlepping that around and storing many copies in the KVS as it is today won't work on tuo, much less el cap.

doc/guide/admin_slurm.rst
garlick: I saw your comment later on - I'll tack that on here, even though people migrating from slurm will definitely know that one.

doc/guide/admin_slurm.rst
garlick: In the Flux sub-instance, like you get when you run `flux alloc` or `flux batch`, all the jobs that run are recorded in the KVS like usual, but once the sub-instance exits, the KVS data is gone unless you use the `--dump` option to preserve a tarball dump on completion.

What is preserved by default is the stdio of the sub-instance.  If your batch script or whatever uses something like `flux watch`, the sub-instance job output goes there and is preserved.  But if you're looking for things like the eventlogs for jobs that ran, it's gone unless you save it.  We need to change the default so that at least some of that is preserved, without increasing the coupling too much so that it affects scalability.

Side note: you can `flux --parent kvs put ...` from the sub instance (for example).

doc/guide/admin_slurm.rst
garlick: I'm not sure this doc needs a bunch of links but I would not fight too hard against that if you feel strongly.

doc/guide/admin_slurm.rst
garlick: I think under this topic (replacing slurm with a Flux system instance), it's a show-stopper if packages are unavailable, which is why it's called out as a problem here.

doc/guide/admin_slurm.rst
garlick: Edit: oops- thought this comment was about commercial support and said something dumb.  Snip!

Yes, sounds good.

doc/guide/admin_slurm.rst
garlick: In Slurm, the job has a prolog and epilog.  Steps do not have one.
In Flux, steps are jobs, but only the top level Flux can be configured to have a prolog and epilog.

So it's pretty much the same and probably doesn't need to be mentioned.

doc/guide/admin_slurm.rst
garlick: I can mention Go but since I'm not finding docs and don't think it's being developed and tested as actively as the C and python bindings, I would probably want to just add something like "Experimental bindings for Go and Julia are in development".

Oh that reminds me that the fact that Slurm's APIs are GPL is considered a problem by some collaborators.  We should mention that ours are the non-viral LGPL.

doc/guide/admin_slurm.rst
garlick: I think I'm getting tired so I'm not sure how to translate that comment into a change.  Well I'll push up what I have.  Thanks so much for all the review work!

doc/guide/admin_slurm.rst
vsoch: I think momentum is a good word for that.

doc/guide/admin_slurm.rst
vsoch: Oh! There is also `--root` !

image

That might be exactly what I need - will test soon. Thank you! :pray:

doc/guide/admin_slurm.rst
vsoch: I don't feel immensely strongly. Reader dudes can use Google-dude.

doc/guide/admin_slurm.rst
vsoch: image

(Side note, I agree, I just thought that line was perfect for a cartoon)

doc/guide/admin_slurm.rst
vsoch: It's mostly for sched, and because we integrate the scheduler into Kubernetes. It's packaged alongside flux-sched, [here](https://github.com/flux-framework/flux-sched/tree/master/resource/reapi/bindings/go)  but also has an automated release (and modular package that I use across projects) at https://github.com/flux-framework/fluxion-go. You can see an example in [fluence](https://github.com/flux-framework/flux-k8s/blob/c9541c15c75928fb033ecfbf160410024166c36e/src/fluence/go.mod#L6). 

Side note - I did make a skeleton for go bindings, [flux-go](https://github.com/flux-framework/flux-go) and it's primarily at a stage of "Will continue development when there is a use case." I'm anticipating that might be next year if we are going to look at center-level stuff that uses Flux. Go is the language of cloud stuff. Python is good if it needs to be touched/understood by a scientist (or is a prototype) but doing some final, production service thing in Python or C won't really be the best choice, imho.

doc/guide/admin_slurm.rst
vsoch: Here is are two suggestions:

```console
Flux is agnostic to where you deploy it.

    Flux is modular and composable, making it easy to integrate into converged and cloud environments.  
    Flux runs in a single container, via docker-compose, virtual machines in the cloud, or easily in Kubernetes.    
    Elasticity and dynamism map well to auto-scaling and ephemeral resources. Programming a workload to use
    Flux thus makes it portable across environments.
```

Or different title:

```console
Flux is a portability layer
```

garlick: Thanks, just reworked that section based on some of your input.  I squashed the fixups and force pushed.

If this looks OK, lets get this merged and if @trws or anyone else wants to make another pass, we can do it in another PR?

doc/guide/admin_slurm.rst
```diff
@@ -15,41 +15,72 @@ Check back with us at the end of 2026.
...
+The Flux project began around 2012.  Flux has been used for a decade or more
+for managing complex ensembles and workflows at LLNL under Slurm and LSF
+in situations where the traditional workload managers were not up to the task.
+Although Flux was designed from the beginning to replace Slurm and was
+deployed as such on several small systems at LLNL, it did not gain momentum
+as a system workload manager until 2024 with the early deliveries of
+`El Capitan <https://hpc.llnl.gov/hardware/compute-platforms/el-capitan>`_.
+
+Flux is now in daily production use as the sole system workload manager on
+El Capitan (currently in slot 1 of the `TOP500 <https://www.top500.org/>`_)
```
trws: Maybe date the "currently"?
```suggestion
El Capitan (currently, in late 2025, in slot 1 of the `TOP500 <https://www.top500.org/>`_)
```

doc/guide/admin_slurm.rst
trws: Not exactly, long topic though.  Sched makes prospective reservations as part of backfill, but doesn't allow *immovable* future reservations to be requested.  Dan has a now old PR that would add an admin-only version of this, but we ran into a big tooling gap and various issues integrating it elsewhere so that's not merged yet.

doc/guide/admin_slurm.rst
trws: This needs a bit of a reword if you want to keep it.  We still use a graph-based model for resource matching and selection, the simplified serialization format doesn't change that.  Could change "graph-based scheduling" in the title to "precise graph system representation" or "serialization" or some such?

doc/guide/admin_slurm.rst
trws: Is it true that only the top-level flux _can_ have prologs and epologs, or that only the system instance is usually configured to have them?  It seems like a bit of an odd limitation for them to be forbidden in user instances (they would be user-mode prolog/epilog steps, no admin, but still).

doc/guide/admin_slurm.rst
trws: We have the classic flux-tree, which did the same thing, but this already mentions the recursive parallelism so I would call it covered personally.

vsoch: flux-tree was removed (we have it still in tutorial content if you need me to dig it up), and (if you haven't looked at the code recently) it's a hugely complex spaghetti bash script (but with organized functions, so maybe one 'getti) with a Python wrapper to plug into the flux cli that (imho) would be unmaintainable. I have a nice library that is going to make this a more production setup, also affording more advanced use cases like creating the hierarchy of brokers to later submit to with different communication strategies. For example, `flux-tree` won't work to provide a service of flux hierarchies that exist across centers.

vsoch: Here is flux-tree:

- https://github.com/flux-framework/Tutorials/blob/ef36e935ba8665d1b3ea61d567918d779953bfe2/2025/ISC-AWS/JupyterNotebook/flux-tree/flux-tree#L123. it's great for a demo, but doesn't fit the bill for the kind of interactions I want, and that will be reflected in future cases for multi-cluster scheduling.
- I am working on https://github.com/converged-computing/flux-hierarchy/ as a more production tool. We already have a setup that can handle a URI interface, interfaces to start / stop / submit / measure throughput, an ability to define the shape of the tree/hierarchy, and node executors for different node or cluster types.

vsoch: And here are latest results. I'm already seeing just under 4K.

https://github.com/converged-computing/flux-hierarchy/issues/5#issuecomment-3651437292

With known optimizations that I need to do (haven't done yet).

doc/guide/admin_slurm.rst
vsoch: I seem to remember for prolog/epilog for usernetes everything would be run as the system user (privileged) and I wound up using user-level services instead. If there isn't a way to enable more user-scoped prologs, I really like the idea @trws - I would absolutely use that.

garlick: > Is it true that only the top-level flux can have prologs and epologs, or that only the system instance is usually configured to have them? It seems like a bit of an odd limitation for them to be forbidden in user instances (they would be user-mode prolog/epilog steps, no admin, but still).

Right, the second thing.

Sorry I misspoke above.

garlick:
Thanks @trws - I pushed a couple of changes per your comments.

garlick:
Still need a review check here.  If anyone wants to do that, we can get this merged IMHO

garlick:
Thanks @vsoch!   I'll set MWP.

PR #7254, opened 2025-12-24
broker: split bootstrap into two phases

garlick:
Problem: moving overlay to a built-in module requires a portion of the bootstrap (where rank and size are found) to happen in the broker, and the rest (where peer endpoints are exchanged) in overlay.

This refactors bootstrap in the broker.
There is a new `bootstrap.c` that does the first phase.
Overlay now calls into `boot_pmi.c` and `boot_config.c` where the broker used to do that.
Some initialization that can happen in the first phase is shifted out of `boot_pmi.c` and `boot_config.c` to `bootstrap.c`.

Finally, the overlay is initialized just before built-in modules are loaded, demonstrating that it isn't doing anything that the things that come before are reliant on.

A future PR will make the final change to move overlay to a built-in module.  When that happens, `bootstrap.c` will let overlay drive the upmi client via RPCs instead of directly.

### DIFF
--- a/src/broker/runat.c
+++ b/src/broker/runat.c
@@ -87,6 +87,7 @@ static const char *env_blocklist[] = {
     "FLUX_KVS_NAMESPACE",
     "FLUX_PROXY_REMOTE",
     "PMI_*",
+    "PMIX_*",
     "FLUX_PMI_LIBRARY_PATH",
     "I_MPI_PMI_LIBRARY",
     "SLURM_*",  // flux-framework/flux-core#5206


garlick:
Marking WIP for now - please hold off on review.

garlick:
OK now that I have the next phase just about working, I'll drop the WIP here. 

I'm sure this won't be very fun to review, but since it's mostly shuffling existing code blocks around, hopefully it won't require too much detailed scrutiny.

src/broker/boot_config.c
```diff
@@ -311,17 +266,23 @@ int boot_config (flux_t *h,
         goto error;
     }
 
+    if (!(cache = bizcache_create (boot->upmi, info->size))) {
+        errprintf (errp, "could not create business card cache");
+        goto error;
+    }
+
```
chu11: within the commit `broker: split boostrap into two phases`, there may be a subtlety I'm missing.

A) is the bizcache addition just an optimization?  Could it be added in a separate commit?
B) I don't see any additional calls to `bizcache_put()` in the `boot_config()` code.  are the additions to the cache coming from the `boot_pmi()` side of the house?

src/broker/bootstrap.c
```diff
@@ -55,6 +56,7 @@ struct bootstrap *bootstrap_create (struct broker *ctx,
         errprintf (errp, "out of memory");
         return NULL;
     }
+    boot->ctx = ctx;
```
chu11, line 512: this line looks like a mistake?

src/broker/boot_config.c
garlick: Possibly a refresh is needed?  There's now
- 950a93dac2dfae878cff8fbbc72c66257f642cd1  broker: refactor bootstrap code
- bf1036a1c9cc14465373a0294e81c4018a731792  broker: move bizcache to bootstrap

The bizcache is now instantiated in bootstrap.c and used from both boot_config.c and boot_pmi.c.
Before it was instantiated independently in both.  For this PR, it's removing duplicate code, but FYI later code builds on it in bootstrap.c to offer RPCs to the overlay module that utilize the cache (which has to live in the broker due to phase one being there).

src/broker/bootstrap.c
garlick: How do you mean?

In case it's not clear, there's a `struct broker *ctx` member of `struct bootstrap`.

chu11: hmmm, I refreshed so I'm not sure what I exactly saw before.  IIRC it looked you were setting  `boot->ctx = ctx` after destroying a boot ctx.  Anyways ... it looks ok now.

src/broker/boot_config.c
chu11: oops, yeah needed to refresh.  i can take another look now.

src/broker/bootstrap.c
chu11: Ahhh my original "this looks wrong" is from the commit `broker: move bizcache to bootstrap`

```
    if (boot) {
        int saved_errno = errno;
        bizcache_destroy (boot->cache);
        upmi_destroy (boot->upmi)
        free (boot);
        errno = saved_errno;
        errprintf (errp, "out of memory");
        return NULL;
    }
    boot->ctx = ctx;
```

you're setting `ctx` in `boot->ctx` right after cleanup?  I'm assuming that line was rebased incorrectly?

src/broker/boot_config.c
chu11: ok, my comment from before still is the same (I now see that my prior comment did list the incorrect commit).

Within the commit `broker: refactor bootstrap code`, I see an addition of a call to `bizcache_create()` in `boot_config()`.  Later on in the function I see calls to `bizcache_get()`.  Between those two calls, there are no new calls to `bizcache_put()` and AFAICT no use of the new cache.  Then at the end of the function, `bizcache_destroy()` is called.

So it's not clear to me how that new cache could be used when nothing populates it?

Perhaps this is a rebase error and some of these lines should be in a different commit?

src/broker/bootstrap.c
garlick: The actual diff there is
```diff
diff --git a/src/broker/bootstrap.c b/src/broker/bootstrap.c
index fb9bc32fa..9696113a9 100644
--- a/src/broker/bootstrap.c
+++ b/src/broker/bootstrap.c
@@ -34,6 +34,7 @@ void bootstrap_destroy (struct bootstrap *boot)
 {
     if (boot) {
         int saved_errno = errno;
+        bizcache_destroy (boot->cache);
         upmi_destroy (boot->upmi);
         free (boot);
         errno = saved_errno;
@@ -55,6 +56,7 @@ struct bootstrap *bootstrap_create (struct broker *ctx,
         errprintf (errp, "out of memory");
         return NULL;
     }
+    boot->ctx = ctx;
 
     if (getenv ("FLUX_PMI_DEBUG"))
         upmi_flags |= UPMI_TRACE;
```
You're joining code fragments from two different functions.

chu11: ahhh, sorry.  The way github laid this out on screen totally confused me.

garlick:
Just added a commit to address the flux-pmix CI failure.

src/broker/state_machine.c
```diff
@@ -314,6 +316,12 @@ static void kickoff_continuation (flux_future_t *f, void *arg)
                   future_strerror (f, errno));
         goto error;
     }
+    if (bootstrap_finalize (s->ctx->boot, &error) < 0) {
+        flux_log (s->ctx->h,
+                  LOG_ERR,
+                  "error finalizing bootstrap: %s",
+                  error.text);
+    }
```
chu11: suggestion: add a comment for why you are calling bootstrap_finalize here (singleton case, per the commit message)

src/broker/boot_config.c
garlick: Earlier we merged
- #7249 

which pre-populates the PMI KVS with business cards for all ranks based on the config file `[bootstrap]` table.  That is why there are no puts.  In this commit the gets are replacing `geturibyrank()`.

The commit after this one moves bizcache to bootstrap.c, leaving just the gets.

The goal is to homogenize and eventually combine the two bootstrap methods.  We don't quite get there here, nor in the follow-on PR, which is focused on moving the overlay out of the broker thread, but it sets the stage for future improvements.

garlick:
Anything else needed to approve?
I have the suggested comment queued up and will rebase on current master and force push if that's OK.

chu11:
> Anything else needed to approve?

Oops, I guess I hit comment instead of approve :-)  Will approve now.

garlick:
Thanks, I'll set MWP.

PR #7256, opened 2025-12-24
libpmi: improve client config boot error message

garlick:
Problem: when "flux pmi -v" steps through upmi methods, the message printed by the config method before the next method is tried is "preinit required config/hostname arguments are missing", which sounds like a programming error.

Change it to "bootstrap configuration not found".


chu11 (approved):
LGTM!

garlick:
Thanks! Setting MWP

PR #7259, opened 2026-01-04
flux-config: add subcommands for fine tuning configuration

garlick:
Problem: It is awkward to update one key using `flux-config` since it only allows the entire config object to be replaced.
E.g. to change one key, one has to do something like:
```
flux config get | jq ".foo.bar.baz = 42" | flux config load
```

Add a  `set`  sub-command that does the above more succinctly
```
flux config set --type=integer foo.bar.baz 42
```

Also, add an `unset` sub-command that can unset a key
```
flux config unset foo.bar.baz
```

### DIFF
--- a/doc/man1/flux-config.rst
+++ b/doc/man1/flux-config.rst
@@ -11,7 +11,7 @@ SYNOPSIS
 | **flux** **config** **get** [*--default=VALUE*] [*--type=TYPE*] [*NAME*]
 | **flux** **config** **set** [*--type=TYPE*] *NAME* *VALUE*
 | **flux** **config** **unset** *NAME*
-| **flux** **config** **builtin** [*NAME*]
+| **flux** **config** **builtin** *NAME*
 | **flux** **config** **load** [*PATH*]
 | **flux** **config** **reload**
 


src/cmd/builtin/config.c
```diff
@@ -181,6 +255,95 @@ static int config_get (optparse_t *p, int ac, char *av[])
...
+    json_t *new_val;
+
+    if (optindex != ac - 2) {
+        optparse_print_usage (p);
+        exit (1);
+    }
+    path = av[optindex++];
+    value = av[optindex++];
+    if (typestr && parse_json_type (typestr, &type, &fsd_subtype) < 0)
+        log_msg_exit ("Unknown type: %s", typestr);
```
chu11: anal corner case.  i think that this would allow `fsd-integer` and `fsd-real` to be valid inputs and fall through to the below.  I don't think it'll cause an error, but maybe wouldn't be what the user expects.  Should we try and catch that corner case?  Admittedly, not sure if it matters.

t/t2806-config-cmd.t
```diff
@@ -232,5 +232,102 @@ test_expect_success NO_SECURITY_CONF 'flux-config get --config-path=security fai
...
+'
+test_expect_success 'array has expected value' '
+	test "$(flux config get --type=array a.b.array)" = "[1,2,3]"
+'
+test_expect_success 'object has expected value' '
+	test "$(flux config get --type=object a.b.object)" = "{}"
+'
+test_expect_success 'fsd has expected value' '
+	test "$(flux config get --type=fsd-integer a.b.fsd)" = "2700"
+'
```
chu11, line 306: suggestion: put the set & get within a single test, so the whole chunk is shorter.  i.e.

```
	flux config set --type=boolean a.b.booleanF false &&
	test "$(flux config get --type=boolean a.b.booleanF)" = "false"
'
```

chu11 (approved):
LGTM, just two minor comments below.

src/cmd/builtin/config.c
garlick: Great point, I'll fix that.

t/t2806-config-cmd.t
garlick: Yeah this is sort of irritatingly repetitive.  I wanted to be sure that setting one value in the same directory didn't clobber all the other values though, hence they are checked  later on.  Maybe I can try a little harder...

garlick: On second thought, this uses "jpath" which has unit tests, so that's probably overkill.  i'll combine them.

garlick:
Addressed those two issues.  I'll go ahead and set mwp - thanks!

Issue #6552, opened 2025-01-13
not ok 26 - attach: --stdin-ranks works


chu11:
While churning through CI runs in #6544 hit this failure several times

```
expecting success: 
  	id=$(flux submit -N4 -o exit-timeout=none -t60s cat) &&
  	echo hello from 0 \
  		| flux job attach -vEX --label-io -i0 $id >stdin-ranks.out &&
  	flux job eventlog -p guest.input $id &&
  	cat <<-EOF >stdin-ranks.expected &&
  	0: hello from 0
  	EOF
  	test_cmp stdin-ranks.expected stdin-ranks.out

  0.000s: job.submit {"userid":1001,"urgency":16,"flags":0,"version":1}
  0.015s: job.validate
  0.027s: job.depend
  0.027s: job.priority {"priority":16}
  0.029s: job.alloc {"annotations":{"sched":{"resource_summary":"rank[0-3]/core[0-1]"}}}
  0.051s: job.start
  0.032s: exec.init
  0.035s: exec.starting
  0.091s: exec.shell.init {"service":"1001-shell-f8AVminB","leader-rank":0,"size":4}
  0.100s: exec.shell.start {"taskmap":{"version":1,"map":[[0,4,1,1]]}}
  0.107s: exec.shell.task-exit {"localid":0,"rank":0,"state":"Exited","pid":172646,"wait_status":0,"signaled":0,"exitcode":0}
  60.052s: job.exception f8AVminB type=timeout severity=0 resource allocation expired
  60.071s: exec.complete {"status":36352}
  60.071s: exec.done
  flux-job: task(s) Alarm clock
  60.071s: job.finish {"status":36352}
  60.077s: job.release {"ranks":"all","final":true}
  60.077s: job.free
  60.077s: job.clean
  not ok 26 - attach: --stdin-ranks works
```

chu11:
huh, I just noticed

```
  0.091s: exec.shell.init {"service":"1001-shell-f8AVminB","leader-rank":0,"size":4}
  0.100s: exec.shell.start {"taskmap":{"version":1,"map":[[0,4,1,1]]}}
  0.107s: exec.shell.task-exit {"localid":0,"rank":0,"state":"Exited","pid":172646,"wait_status":0,"signaled":0,"exitcode":0}
  60.052s: job.exception f8AVminB type=timeout severity=0 resource allocation expired
  60.071s: exec.complete {"status":36352}
```

[...]

PR #7260, opened 2026-01-06
broker: offload overlay handling to a separate thread

### LINKED ISSUE #6967: idea: isolate  downstream overlay network handling in a separate broker thread

garlick:
Problem: in a large, flat TBON, the leader broker can become unresponsive when overlay activity is high.

Specifically:

- When starting a full system job that does not include the leader broker, there is significant "through-routing" of messages between the first node of the job and other nodes.  This keeps the main leader broker thread busy and starves out message routing between flux commands and flux services on the leader, and between services on the leader.
- When there is a heavy KVS write load, such as when jobs are stopping and starting, many `setroot` messages are published, each of which has to be software multicast to all peers.  The leader broker goes into an uninterruptible loop over all peers each time a message is published, starving out other work.

This PR moves overlay handling to a separate thread (a built-in broker module).   The new `overlay` module routes messages between overlay peers and implements software multicast independent of the broker thread.   This _should_ greatly improve responsiveness under the above sorts of loads.

The `overlay` module communicates with the main broker via a dedicated interthread channel, which has a low cost, since messages are passed through it without copying or serialization.

In addition, the overlay code is among the oldest in Flux, and is tightly coupled to other broker functions.  Isolating it in a module should make it easier to maintain and perhaps down the road, easier to replace or offer alternatives.

Fixes #6967


garlick:
Just rebased this on current master after the merging of
- #7254

garlick:
Resolved the conflicts with current master.

garlick:
This test is repeatedly failing in alpine.  I thought this was a known test issue but maybe this PR has somehow made it worse.  Anyway, further investigation is probably warranted.
```
  setup fake csm_allocation_query for mock lsf testing
  
  expecting success: 
  	result=$(SHELL=/bin/sh flux uri --local lsf:12345) &&
  	FLUX_URI_1=$(flux exec -r 1 flux getattr local-uri) &&
  	test_debug "echo checking if $result is $FLUX_URI or $FLUX_URI_1" &&
  	test "$result" = "$FLUX_URI" -o "$result" = "$FLUX_URI_1"
  
  hostname=lassen9
  cmd="/usr/src/src/cmd/flux uri --remote lsf:12345?is_compute"
  flux-uri: ERROR: PID 241758 is a flux-broker and no child found
  flux-uri: ERROR: Command '['/usr/src/t/scripts/tssh', 'lassen9', '/usr/src/src/cmd/flux', 'uri', '--remote', 'lsf:12345?is_compute']' returned non-zero exit status 1.
  not ok 27 - flux-uri mock testing of lsf resolver works
```

garlick:
Rebased on #7273

garlick:
Hmm, now alpine is timing out in this test
```
2026-01-11T22:40:56.3111697Z ./t4000-issues-test-driver.t timed out after 300.0s
2026-01-11T22:40:56.3111973Z processed 10 jobs
2026-01-11T22:40:56.3112249Z job.exception: type=timeout severity=0 resource allocation expired
2026-01-11T22:40:56.3112561Z 137
2026-01-11T22:40:56.3112813Z Jan 11 22:40:22.951577 UTC 2026 broker.err[0]: kill cleanup: No such process
2026-01-11T22:40:56.3113499Z Jan 11 22:40:22.951899 UTC 2026 broker.err[0]: rc2.0: /usr/src/t/issues/t3503-nofiles-limit.sh Alarm clock (rc=142) 599.5s
2026-01-11T22:40:56.3114067Z Jan 11 22:40:23.630146 UTC 2026 broker.err[0]: kill cleanup: No such process
2026-01-11T22:40:56.3114476Z Jan 11 22:40:23.631487 UTC 2026 broker.err[0]: kill cleanup: No such process
2026-01-11T22:40:56.3114807Z not ok 1 - t3503-nofiles-limit
2026-01-11T22:40:56.3115082Z FAIL: t4000-issues-test-driver.t 1 - t3503-nofiles-limit
```

src/modules/overlay/boot_pmi.c
```diff
@@ -217,101 +215,111 @@ int boot_pmi (struct bootstrap *boot,
...
     topology_decref (topo);
     return 0;
 error:
+#if 0
     /* N.B. Some implementations of abort may not return.
      */
     (void)upmi_abort (boot->upmi,
                       errp ? errp->text : "fatal bootstrap error",
                       NULL);
+#endif
```
grondo: Comment for why this code remains, but is disabled? should it just be removed?

src/modules/overlay/boot_util.c
```diff
@@ -0,0 +1,167 @@
...
+    if (!(f = flux_rpc_pack (h,
+                             "bootstrap.whois",
+                             FLUX_NODEID_ANY,
+                             FLUX_RPC_STREAMING,
+                             "{s:s}",
+                             "ranks", peers))) {
+        errprintf (errp, "bootstrap.whois: %s", future_strerror (f, errno));
+        idset_destroy (ids);
+        ERRNO_SAFE_WRAP (free, peers);
+        flux_future_destroy (f);
```
grondo: Unnecessary `flux_future_destroy()` here since the conditional implies `f == NULL`?

garlick:
Rebased on current master, and tacked on a commit to clarify error messages that @grondo observed during testing of this PR.

garlick:
Restarted noble builder that hit
- #6552

src/modules/overlay/test/overlay.c
```diff
@@ -190,7 +190,6 @@ void single (flux_t *h)
     free (s);
```
chu11, line 185: commit message English typo: "the API does not permit them to set be to empty values."

"them to be set to"?

src/modules/overlay/compat.c
```diff
@@ -0,0 +1,95 @@
...
+int compat_attr_add (flux_t *h, const char *name, const char *val, int flags)
+{
+    if (!val)
+        return compat_attr_delete (h, name, true);
+    if (flux_attr_set (h, name, val) < 0)
+        return -1;
+    if ((flags & ATTR_IMMUTABLE)) {
+        if (flux_attr_set_cacheonly (h, name, val) < 0)
+            return -1;
+    }
```
chu11, line 31: in `overlay: add compat helpers for porting`

commit message probably just needs a little fleshing out, 

`For now, the immutable flag is ignored as this is not available to Flux API users.`

but you're clearly not ignoring it in above.  Probably you mean temporarily just set in the cache to not break things for now.

Also

`honor the flux::iam_broker flag to prevent self-RPCs since the overlay unit test can make use of that.`

but I don't see this flag used anywhere?  Was it a temporary thing during your development?

chu11 (commented):
parking my review through `broker: make struct bootstrap opaque again` (it seems subsequent ones are meatier changes).  Just minor comments.

src/modules/overlay/compat.c
garlick: It's used in the overlay unit test.

chu11: > It's used in the overlay unit test.

I `git grep`-ed for it and couldn't find it used anywhere.

garlick: Oh crud, I renamed it to `flux::attr_redirect` but forgot to update the commit message.  Good catch!

src/broker/bootstrap.c
```diff
@@ -491,10 +492,195 @@ int bootstrap_finalize (struct bootstrap *boot,
...
+            goto error;
+    }
+    else if (json_is_string (ranks)) {
+        if (!(ids = idset_decode (json_string_value (ranks))))
+            goto error;
+    }
+    else
+        goto error_proto;
+    if (idset_count (ids) > 1 && !flux_msg_is_streaming (msg))
+        goto error_proto;
```
chu11: perhaps the reason will become more obvious in later commits, but why not just make it streaming all the time?  Is there a special case reason for an idset of 1?

src/broker/bootstrap.c
```diff
@@ -491,10 +492,195 @@ int bootstrap_finalize (struct bootstrap *boot,
...
+/* Overlay calls bootstrap.whois to fetch the business cards of its
+ * direct peers, specified as an idset.  A PMI barrier is performed,
+ * then the business cards are streamed, one per response.
+ * This may be called multiple times.
+ */
+static void bootstrap_whois_cb (flux_t *h,
+                                flux_msg_handler_t *mh,
+                                const flux_msg_t *msg,
+                                void *arg)
+{
```
chu11, line 568: Comment is a little unclear. It sounds like a pmi barrier is done before the business cards are streamed in this function. But there is no barrier call below.

Or did you mean the barrier call in the overlay module?  But I didn't find a barrier call between call to whois and the streaming.

src/modules/overlay/boot_util.c
```diff
@@ -0,0 +1,167 @@
+/************************************************************\
+ * Copyright 2022 Lawrence Livermore National Security, LLC
```
chu11: 2025/2026?

src/modules/overlay/boot_util.h
```diff
@@ -0,0 +1,37 @@
...
+flux_future_t *boot_util_whois (flux_t *h,
+                                int *ranks,
+                                int count,
+                                flux_error_t *errp);
+
+struct bizcard *boot_util_whois_get_bizcard (flux_future_t *f);
+
+int boot_util_whois_get_rank (flux_future_t *f);
+
+int boot_util_finalize (flux_t *h, struct idset *ids, flux_error_t *errp);
```
chu11: suggestion, make the parameter "crit_ids" or something similar, just not obvious at first glance what the parameter is for.

src/modules/overlay/boot_config.c
```diff
@@ -61,7 +61,7 @@ static const char *getbindbyrank (json_t *hosts,
  * Parse the URI in advance, and if the host portion is invalid, log it.
  * Ref: flux-framework/flux-core#5009
  */
-static void warn_of_invalid_host (flux_t *h, const char *uri)
+void warn_of_invalid_host (flux_t *h, const char *uri)
```
chu11: removed static b/c of temporary thing?  Seems like it should stay?

src/broker/bootstrap.c
chu11: I get it now.  a barrier is done before business cards are requested / streamed.  It sounded like the barrier was between request & streaming.

src/modules/overlay/overlay.c
```diff
@@ -205,6 +205,7 @@ static int overlay_control_parent (struct overlay *ov,
 static void overlay_health_respond_all (struct overlay *ov);
 static struct child *child_lookup_byrank (struct overlay *ov, uint32_t rank);
 static int overlay_goodbye_parent (struct overlay *overlay, flux_error_t *errp);
+static void overlay_goodbye_cb (struct overlay *ov, const flux_msg_t *msg);
```
chu11, line 208: in commit message in `overlay: send overlay.goodbye directly to parent`

you say `Send the message directly to the parent socket, like overlay.hello is.`

I'm not sure if this is a nomenclature thing.  But you're not "sending it", you're just handling it within the module?

chu11 (commented):
comments through commit "overlay: send overlay.goodbye directly to parent￼"

src/modules/overlay/overlay.c
garlick: Yes it's really about both sending and receiving, since a broker can be an interior node that has to receive a goodbye from children and send a goodbye to parent.  I'll rework that commit message.

src/modules/overlay/overlay.c
```diff
@@ -2345,67 +2300,64 @@ static int overlay_configure_interface_hint (struct overlay *ov,
...
+
+    // command line attribute setting
+    if (initialize) {
+        if ((tmin = flux_attr_get (ov->h, "tbon.torpid_min"))
+            && fsd_parse_duration (tmin, &tmin_val) < 0)
+            return errprintf (errp, "error parsing tbon.torpid_min attribute");
+        if ((tmax = flux_attr_get (ov->h, "tbon.torpid_max"))
+            && fsd_parse_duration (tmax, &tmax_val) < 0)
+            return errprintf (errp, "error parsing tbon.torpid_max attribute");
+    }
```
chu11: could put all this inside one if statement?

src/modules/overlay/overlay.c
```diff
@@ -2345,67 +2300,64 @@ static int overlay_configure_interface_hint (struct overlay *ov,
...
+            || flux_attr_set (ov->h, "tbon.torpid_min", fsd) < 0)
+            return errprintf (errp, "error updating tbon.torpid_min attribute");
+        ov->torpid_min = tmin_val;
+    }
+    // update tbon.torpid_max
+    if (tmax_val != -1) {
+        if (fsd_format_duration (fsd, sizeof (fsd), tmax_val) < 0
+            || flux_attr_set (ov->h, "tbon.torpid_max", fsd) < 0)
+            return errprintf (errp, "error updating tbon.torpid_min attribute");
+        ov->torpid_max = tmax_val;
```
chu11, line 2414: add check that tmax > tmin?

src/modules/overlay/overlay.c
```diff
@@ -1119,12 +1152,27 @@ static void child_cb (flux_reactor_t *r,
...
-        log_tracker_error (ov->h, msg, errno);
+    if (!(response = flux_response_derive (msg, EHOSTUNREACH))
+        || flux_msg_set_string (response, "overlay disconnect") < 0)
+        goto error;
+    last_hop = flux_msg_route_last (response);
+    if ((child = child_lookup (ov, last_hop))) {
+        if (overlay_sendmsg_child (ov, response) < 0)
+            goto error;
+        flux_msg_decref (response);
+    }
```
chu11: don't need `child` local variable.

t/t3308-system-torpid.t
```diff
@@ -93,7 +102,7 @@ test_expect_success 'rank 1 is removed from broker.torpid group' '
 '
 
 test_expect_success 'set tbon.torpid_max to impossible to attain value' '
-	flux setattr tbon.torpid_max 0.1s
+	set_torpid 1s 0.1s
```
chu11: why is `0.1s` an impossible value?  It is a valid fsd and is > 0.  Some subtlety I'm missing that wasn't obvious in the new config parsing code?

src/modules/overlay/overlay.c
```diff
@@ -800,14 +812,35 @@ static void log_tracker_error (flux_t *h, const flux_msg_t *msg, int errnum)
...
 
-    if (!(rep = flux_response_derive (msg, EHOSTUNREACH))
-        || flux_msg_route_delete_last (rep) < 0
-        || flux_msg_route_delete_last (rep) < 0
-        || flux_send (ov->h, rep, 0) < 0)
-        log_tracker_error (ov->h, rep, errno);
-    flux_msg_destroy (rep);
+    if (!(response = flux_response_derive (msg, EHOSTUNREACH))
+        || flux_msg_route_delete_last (response) < 0
+        || flux_msg_route_delete_last (response) < 0)
```
chu11, line 822: just for my own education, why the removal of two routes right here?

doc/man7/flux-broker-attributes.rst
```diff
@@ -289,19 +289,27 @@ tbon.interface-hint [Updates: C, R]
    3. the enclosing Flux instance's value (via the PMI KVS)
    4. the compiled-in default of default-route
 
-tbon.torpid_min [Updates: C, R]
+tbon.torpid_min [Updates: C]
```
chu11, line 292: commit message error, you say you drop the C flag, but you mean R flag.

src/modules/overlay/overlay.c
garlick: The 0MQ sockets used between overlay peers are DEALER (parent) and ROUTER (children).  When the ROUTER receives a message from a child, it pushes the child UUID onto the message route stack.  When the ROUTER sends a message to a child it pops off the next UUID and uses it to select the child.  This is designed for request/response routing in the upstream direction, e.g. where the parent acts as server.

Flux supports bidirectional request/response routing.  When the child acts as server, we have to compensate for the ROUTER behavior.    For requests sent to the child, we want the child to receive the request with the sender UUID on the top of the route stack, so we push it on twice, once to select the peer, and once so that when the first one is popped off, a copy remains.

The reverse happens when processing responses received from the child.  The router will have pushed the child's UUID onto the route and not popped anything off, but the effect we wanted was to have the route of the current broker popped off so the next hop is on top.  So we have to pop off two routes and discard them.

src/modules/overlay/overlay.c
garlick: Actually I believe I've gotten it wrong here.  The sequence should be
1. compiled-in defaults
2. TOML config
3. attribute (command line override)

Will fix that.

t/t3308-system-torpid.t
garlick: Poor wording - I just meant impossible to not be torpid.  This test won't work with if we require min < max (which I think is a good idea) so it will need to be dropped or reworked, which I will do.

garlick:
@chu11, any objections if I force-push the changes responding to your comments so far?

chu11:
@garlick go for it!

garlick:
OK, rebased on current master and squashed fixes to the comments so far.

chu11 (approved):
looked through and all the changes look good.  The only comment I made above that I think was missed was about the `warn_of_invalid_host()` accidental removal of `static`.

garlick:
Oh yes, will fix that, thanks!

garlick:
Rebased on current master and fixed that `static` omission.  Thanks!

grondo (approved):
LGTM! Just noted a couple trivial things inline.

garlick:
Thanks, I fixed those two things, rebased on current master, and force pushed.  Thank you both!  Setting MWP.

PR #7263, opened 2026-01-07
Tighten const correctness

### LINKED ISSUE #7262: Build fails on Fedora Rawhide (glibc 2.42+) due to C23 const-correctness in string functions
## Summary

flux-core v0.81.0 fails to build on Fedora Rawhide (F44) with `-Werror=discarded-qualifiers` errors. This is caused by glibc 2.42+ implementing C23-conformant `strchr()`, `strstr()`, `strrchr()`, `strpbrk()`, and `memchr()` functions that now return `const char *` when given `const char *` input.

## Environment
- Fedora: Rawhide (F44)
- glibc: 2.42+
- GCC: 15+
- flux-core: v0.81.0

## Root Cause

In C23, the string search functions (`strchr`, `strstr`, `strrchr`, `strpbrk`, `memchr`) are now type-generic macros that preserve const-correctness. When the input is `const char *`, the return type is also `const char *`. Combined with Fedora's `-Werror` flags, this causes build failures in code that assigns these results to non-const pointers.

Reference: https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3020.pdf

## Affected Files and Fixes

I've audited flux-core v0.81.0 and categorized all affected locations:

#### Category 1: Simple const pointer changes (pointer only used for comparison/arithmetic)
These are straightforward fixes where the variable receiving the search result is only used for pointer arithmetic or comparison, never to modify the string:

File | Function | Variable | Search Function
-- | -- | -- | --
src/common/libpmi/keyval.c | parse_val() | match | strstr()
src/common/libjob/sign_none.c | sign_none_unwrap() | p | strchr()
src/common/liboptparse/getopt.c | _getopt_internal_r() | temp | strchr()
src/common/librouter/disconnect.c | disconnect_topic() | p | strrchr()
src/common/libsubprocess/command.c | env_entry_name() | p | strchr()
src/common/libsubprocess/command.c | env_entry_value() | p | strchr()
src/modules/connector-local/local.c | mod_main() | tmpdir | strchr()
src/modules/job-info/util.c | get_next_eventlog_entry() | term | strchr()
src/modules/job-list/job_data.c | parse_job_name() | p | strrchr()
src/shell/mustache.c | mustache_render() | start, end | strstr()
src/common/libeventlog/eventlog.c | eventlog_entry_decode_common() | ptr | strchr()
src/broker/runat.c | stdio_cb() | cp | memchr()
src/cmd/builtin/restore.c | shortblobref_length() | cp | strchr()

**Fix:** Change `char *` to `const char *` for these variables.

#### Category 2: Function return type changes

These functions return the result of `strpbrk()` directly, so their return types need to change:

File | Function | Change
-- | -- | --
src/common/librlist/match.c | property_query_string_invalid() | char * → const char *
src/common/librlist/rlist.c | property_string_invalid() | char * → const char *

For `rlist.c`, the callers also need updates:
- `rlist_add_property()`: Add separate `const char *invalid` variable (keep `char *p` for `idset_encode()` which needs `free()`)
- `rlist_assign_properties()`: Replace `char *p` with `const char *invalid` (only used for property check)

#### Category 3: Explicit casts for mutable strdup'd copies

These cases involve `strchr()` on strings that ARE modified, but the modification is safe because the string is a `strdup()`'d mutable copy:

File | Function | Notes
-- | -- | --
src/common/libutil/cronodate.c | get_range() | Cast strchr() result to (char *) - operating on strdup()'d copy
src/common/libutil/cronodate.c | range_parse() | Cast strchr() result to (char *) - operating on strdup()'d copy
src/common/libutil/basename.c | basename_simple() | Cast return value - function returns mutable pointer into const input

Fix: Use explicit `(char *)` cast with comment explaining why it's safe.

#### Category 4: Latent bug fix

File | Function | Issue
-- | -- | --
src/common/libhostlist/hostlist.c | parse_next_range() | **Bug:** Was calling `strchr(str, '-')` on the const str parameter but then modifying through the result. Should use `strchr(orig, '-')` where `orig` is the `strdup()`'d mutable copy. Also `strtoul(str, ...)` should be `strtoul(orig, ...)`.

This is a latent bug that would cause undefined behavior if the compiler didn't catch it - the code was modifying a string through a pointer derived from a const source.

## Proposed Solution

I've prepared a patch that addresses all these issues and builds on rawhide! The changes are minimal and focused:
1. Preserve const-correctness where pointers are read-only
2. Use explicit casts with comments where modification of strdup'd copies is intentional
3. Fix the latent bug in `hostlist.c`

## Related
- `flux-security` had a similar issue fixed in their PR 211
- Derivative of #7211 
- This affects all projects building with glibc 2.42+ and `-Werror`

## Patch

The patch file I used to get a [rawhide copr build](https://copr.fedorainfracloud.org/coprs/kushgupta/flux-core/build/9982635/) working is [here](https://github.com/kush-gupt/flux-rpm/blob/main/flux-core/const-correctness.patch)

Let me know if y'all would like a PR, any comments or ideas on any of the above, and how you'd like to proceed here!

kush-gupt:
Tighten const correctness across various modules and string helpers to better reflect immutability and avoid unnecessary mutable pointers for better C23 compatibility.

Bug Fixes:

- Correct use of original mutable buffers when parsing numeric ranges to avoid modifying string literals or incorrect pointers.

Enhancements:

- Adjust basename_simple and its callers to return and use const char* instead of mutable char*.
- Update internal helpers and local variables that inspect strings (e.g., strchr/strpbrk results, log program name, temp directory paths) to use const-qualified pointers.
- Document and add explicit casts where mutable access to strchr results is intentional under newer C standards.

Fixes #7262

### DIFF
--- a/src/common/libutil/cronodate.c
+++ b/src/common/libutil/cronodate.c
@@ -227,7 +227,10 @@ static int get_range (const char *r, tm_unit_t u, int *lo, int *hi)
         *lo = tm_unit_min (u);
         *hi = tm_unit_max (u);
     }
-    else if ((p = strchr (r, '-'))) {
+    /* Cast needed: strchr returns const char* in C23, but we know
+     * r points to a strdup'd mutable copy from range_parse().
+     */
+    else if ((p = (char *)strchr (r, '-'))) {
         *(p++) = '\0';
         // r = lo, p = hi
         if (((*lo = tm_string2int (r, u)) < 0)


garlick (dismissed):
Nicely done, thank you so much!

kush-gupt:
The initial fix for `hostlist` changed `strchr(str, '-')` to `strchr(orig, '-')` to avoid modifying a const string, but missed that `range->width = strlen(str)` was relying on `str` being truncated by that (undefined) modification.

For input "1-16383":
- strlen(str) → 7 (unchanged string)
- strlen(orig) → 1 (correctly truncated to just "1")
The wrong width broke hostlist expansion, causing `"big42"` to not be found in `"big[1-16383]"`.

The fix is to use strlen(orig) **_before_** free(orig), which I just added in the hostlist commit!

Issue #7264, opened 2026-01-08
a failed config update is not entirely rejected


garlick:
Problem: when the configuration is updated, e.g. via `flux config reload` or `flux config load`, or now `flux config set`, failures may be caught by modules that "own" various tables and the command fails, but the new config object is actually still replaced on the broker.

Example:
```
$ flux config set --type=integer job-manager.inactive-num-limit -42
flux-config: Error updating config object: job-manager: job-manager.inactive-num-limit: must be >= 0
$ flux config get
{"job-manager":{"inactive-num-limit":-42}}
```
Then when you try to update something else, this can trip you up
```
$ flux config set --type=fsd job-manager.inactive-age-limit 10d
flux-config: Error updating config object: job-manager: job-manager.inactive-num-limit: must be >= 0
```

Issue #6865, opened 2025-06-04
instance configuration should track the leader broker


garlick:
Problem: `flux config reload` only affects the local broker, which loads its config from local files.  This complicates the propagation of configuration changes for a system instance.

Perhaps the a follower broker should make a streaming RPC to its parent to track the parent's configuration object.   Then `flux config reload` would only need to be run on the leader.  It could be made a no-op on a follower.

Note: followers in a system instance still would need some local configuration.  To bootstrap communications, they need the `[bootstrap]` table and the curve key.  They also need the independent config spaces for flux-security and the IMP (this issue only pertains to the broker config).

PR #7265, opened 2026-01-08
config: move broker config management to a built-in module and improve error handling

garlick:
Problem: configuration update handling is tightly coupled to the broker, which makes it more complex to work on than necessary.

Move it to a built-in module and simplify its algorithm for updating module configs now that it has its own thread and can do the work synchronously.

Also, fix a long standing issue with failed config updates
- #7264 

and short circuit updates that don't actually do any useful work.

I think this also will make it easier to implement
- #6865

### DIFF
--- a/t/t2806-config-cmd.t
+++ b/t/t2806-config-cmd.t
@@ -318,4 +318,23 @@ test_expect_success 'flux-config unset works' '
 	test_must_fail flux config get --type=string a.b.string
 '
 
+# issue 7264 - ensure failed updates are entirely rejected
+test_expect_success 'clear the config object' '
+	flux config load </dev/null
+'
+test_expect_success 'load the heartbeat module' '
+	flux module load heartbeat
+'
+test_expect_success 'setting a bad heartbeat config value fails' '
+	test_must_fail flux config set --type=string heartbeat.period 42XYZ
+'
+test_expect_success 'the bad value is not lingering in the config object' '
+	echo "{}" >empty.exp &&
+	flux config get >empty.out &&
+	test_cmp empty.exp empty.out
+'
+test_expect_success 'unload the heartbeat module' '
+	flux module remove heartbeat
+'
+
 test_done


src/modules/config/config.c
```diff
@@ -0,0 +1,289 @@
+/************************************************************\
+ * Copyright 2020 Lawrence Livermore National Security, LLC
```
chu11, line 2: 2026?  some stuff new some not ... ehh

src/modules/config/config.c
```diff
@@ -0,0 +1,289 @@
...
+        goto error;
+    if (flux_respond (h, msg, NULL) < 0)
+        flux_log_error (h, "error responding to config.load request");
+    return;
+error:
+    if (flux_respond_error (h, msg, errno, error.text) < 0)
+        flux_log_error (h, "error responding to config.load request");
+}
+
+static void get_cb (flux_t *h,
```
chu11, line 258: nit: you have a function comment before load and reload ... one here too?

src/modules/config/config.c
```diff
@@ -0,0 +1,289 @@
...
+        errprintf (errp, "buffer overflow");
+        errno = EOVERFLOW;
+        goto done;
+    }
+    if (flux_conf_unpack (conf, NULL, "o", &o) < 0
+        || !(f = flux_rpc_pack (h, topic, FLUX_NODEID_ANY, 0, "O", o))
+        || (flux_future_get (f, NULL) < 0 && errno != ENOSYS)) {
+        errprintf (errp, "%s", future_strerror (f, errno));
+        goto done;
+    }
```
chu11, line 81: do we want this loop to be synchronous w/ each RPC?   Noticed the old code appears to have done a `flux_future_wait_all` future.  I mean, it's probably not a big deal, assuming admins will be (hopefully) reloading the config rarely.

src/modules/config/config.c
```diff
@@ -84,47 +85,88 @@ static int update_one_module (flux_t *h,
...
     size_t index;
     json_t *entry;
     const char *name;
     flux_error_t error;
-    int rc = -1;
+    zlist_t *rollback;
+
+    if (!(rollback = zlist_new ()))
+        goto nomem;
+    zlist_autofree (rollback);
```
chu11, line 123: minor efficiency option: all the module names are valid from the `mods` object in the future.  No need to do autofree.  Although should destroy the future after destroying the rollback zlist.  But optional, I don't know if it's a necessary / big win.

src/modules/config/config.c
```diff
@@ -84,47 +85,88 @@ static int update_one_module (flux_t *h,
...
+ * then the broker, then the local cached copy.  On success, 'conf'
+ * ownership is transferred to 'h'.  If any updates fail, revert the
+ * successful ones.
  */
-static int update_all_modules (flux_t *h,
-                               const flux_conf_t *conf,
-                               flux_error_t *errp)
+static int update_all_new (flux_t *h,
+                           const flux_conf_t *conf,
+                           flux_error_t *errp)
```
chu11, line 111: nit: was function name change necessary from `update_all_modules()` to `update_all_new()`?   Was thinking maybe `update_all_new()` was a placeholder as you were testing / developing.

t/t2806-config-cmd.t
```diff
@@ -318,4 +318,23 @@ test_expect_success 'flux-config unset works' '
...
 
+# issue 7264 - ensure failed updates are entirely rejected
+test_expect_success 'clear the config object' '
+	flux config load </dev/null
+'
+test_expect_success 'load the heartbeat module' '
+	flux module load heartbeat
+'
+test_expect_success 'setting a bad heartbeat config value fails' '
+	test_must_fail flux config set --type=integer heartbeat.period 42
```
chu11: how about an illegal string like `42XYZ` ... I was initially confused on why this was an error, b/c 42 should be ok and it's 42 seconds.  Later to realize it's b/c `heartbeat.period` is supposed to be a string.

chu11 (approved):
LGTM, just a few nit comments below.

src/modules/config/config.c
garlick: No need to reset the clock here IMHO.

src/modules/config/config.c
garlick: I think generally the idea of serializing all config updates probably makes sense.  Giving config its own thread, while overkill, enables this code to be simplified.

src/modules/config/config.c
garlick: Eh, I think it's probably OK as is.  As you say, avoiding the extra strdup()s on update probably isn't a huge win, and it's slightly less likely to end up with a use-after-free if someone shuffles code around later.

src/modules/config/config.c
garlick: FWIW my thinking was that adding the `flux_set_conf()` made `update_all_modules()` not quite on target.

t/t2806-config-cmd.t
garlick: The client will catch a mismatch between `--type` and the argument but we can do something more obviously wrong like a boolean...
```
$ flux config set --type=integer foo 42XYZ
flux-config: Error parsing integer value
```

garlick: Oh, sorry, thinko, yes we can do that.

garlick:
Fixed those things, thanks!  I'll set MWP.

PR #7270, opened 2026-01-09
MAINTAINERS: update first name to match GitHub profile

cmoussa1:
#### Problem

My GitHub and various work profiles use "Christopher," but the maintainers list uses "Chris".

---

Just to make my online presence consistent, this PR updates the name in the **Current Maintainers** list to "Christopher" :-)

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -4,22 +4,22 @@ to this project, please refer to [CONTRIBUTING](CONTRIBUTING.md).
 
 ## Current Maintainers
 
-| Maintainer       | GitHub ID                                             | Affiliation |
-| ---------------- | ----------------------------------------------------- | ----------- |
-| Al Chu           | [chu11](https://github.com/chu11)                     | LLNL        |
-| Chris Dunlap     | [dun](https://github.com/dun)                         | LLNL        |
-| Chris Moussa     | [cmoussa1](https://github.com/cmoussa1)               | LLNL        |
-| Elsa Gonsiorowski| [gonsie](https://github.com/gonsie)                   | LLNL        |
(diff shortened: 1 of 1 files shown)


wihobbs:
This is probably on several other repositories as well. Do we have a way update a bunch of `MAINTAINERS.md` files at once when, say, a new user is added to the core team?

cmoussa1:
Do we have more than one `MAINTAINERS.md` document other than flux-core? I'm not sure. 🤷 flux-accounting doesn't have one, at least.

wihobbs:
If you're a member of the `core` team on GitHub, you automatically have write/maintainer access to a bunch of the repositories in the flux-framework org (see [flux-pam MAINTAINERS.md](https://github.com/flux-framework/flux-pam/pull/14/files)). 

I wonder if we could set up a CI job to auto-generate the `MAINTAINERS.md` file based on team membership + the user's listed organization on their GitHub profile...

garlick:
@cmoussa1 - we're supposed to have MAINTAINERS.md files in each repo per https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_48.html#roles

@wihobbs - the MAINTAINER.md files are all separate and IIRC the original thought was that we would use our change process to add or remove maintainers in each repo, independently.   In other words any change to the maintainers list for a repo would be accompanied by a pull request and approval by the current maintainers.  What I was thinking was that the first time we need to change the maintainers list for a given repo, we would start by converting it to use a local access list rather than the team, with the same members.  Then we would propose to add/remove.  Maybe others won't agree with me though, so perhaps this could be a team discussion in a meeting, preferably after Mark returns.

Edit: meanwhile, sorry, I think we'll have to manually fix each copy for small changes like the first name change.

PR #7271, opened 2026-01-10
libeventlog: wait for pending commits to complete

### LINKED ISSUE #7257: eventlogger: flush/commit should wait for pending requests to complete
There is an API corner case in the eventlogger that cannot manifest with any current use of `eventlogger_commit()`.  It might manifest with `eventlogger_flush()` in the shell. I discovered this while working on #6910 as it manifests itself within that work.

When using the API like this:

```
    eventlogger_append (ev, "some-event", ...);

    f = eventlogger_commit (ev);
```

Under this use, there won't be any problems.  We haven't re-entered the reactor between these two calls and we can be certain the current "batch" of events stored within `eventlogger` is > 0.  So `eventlogger_commit()` returns a future you can wait on until those events have been committed.

The corner case is:

```
    eventlogger_append (ev, "some-event", ...);

    /* do something that re-enters reactor momentarily */

    f = eventlogger_commit (ev);
```

Behind the scenes the eventlogger has a timer that could trigger and send out the "current" batch of events.  Then the "current" batch of events is wiped clean and is empty.  So `eventlogger_commit()` returns a fulfilled future b/c there are "no events to commit".  So that future immediately completes, and now we have a race against the real commit.

`eventlogger_commit()` should instead wait until all pending commits have completed.  A list of futures of "people waiting for pending stuff" should be sufficient to handle this until the pending list is empty (alternate idea: use `future_aux_set()` to set waiters on the current futures on the pending list).

`eventlogger_flush()` has the same basic problem as it (essentially) uses `eventlogger_commit()`.

One idea: collapse `eventlogger_commit()` / `eventlogger_flush()` into one function that returns a future.  Synchronous callers should just wait on the future for synchronousness.  TBD which function name lives on.

chu11:
Problem: Both eventlogger_commit() and eventlogger_flush() assume all prior pending commits have completed.  This is not a problem under normal conditions, as there is always some "current" event data in the eventlogger that needs to be written out, and commit/flush can wait for that commit to be completed.

However, under some circumstances there may not be any "current" data to be written out.  This can lead to a race as code returns a fulfilled future, assuming all prior pending commits have completed.

Its worth noting there is no current use of eventlogger_commit() that could lead to a race.  Races with eventlogger_flush() are
theoretically possible in the shell.

Solution: When there are no "current" events to be written out and there are pending commits waiting to be completed, wait for the most recent pending commit to complete before fulfilling the future.  We accomplish this by creating a new "empty" batch of events and commit that empty set of events.  The future created by this "empty"  batch can be waited on.

Fixes #7257

### DIFF
--- a/src/common/libeventlog/eventlogger.c
+++ b/src/common/libeventlog/eventlogger.c
@@ -140,15 +140,15 @@ static void eventlog_batch_error (struct eventlog_batch *batch, int errnum)
     zlist_remove (ev->pending, batch);
 }
 
-static void commit_cb (flux_future_t *f, void *arg)
+static void commit_batch_cb (flux_future_t *f, void *arg)
 {
     struct eventlog_batch *batch = arg;
     eventlogger_batch_complete (batch);
     flux_future_destroy (f);
 }
 
-static flux_future_t *eventlogger_commit_batch (struct eventlogger *ev,
-                                                struct eventlog_batch *batch)
+static flux_future_t *commit_batch (struct eventlogger *ev,
+                                    struct eventlog_batch *batch)
 {
     flux_future_t *f = NULL;
     flux_future_t *fc = NULL;
@@ -177,25 +177,25 @@ static flux_future_t *eventlogger_commit_batch (struct eventlogger *ev,
         flux_watcher_stop (batch->timer);
         if (!(fc = flux_kvs_commit (ev->h, ev->ns, flags, batch->txn)))
             return NULL;
-        if (!(f = flux_future_and_then (fc, commit_cb, batch)))
+        if (!(f = flux_future_and_then (fc, commit_batch_cb, batch)))
             flux_future_destroy (fc);
     }
     return f;
 }
 
-static flux_future_t *commit_batch (struct eventlogger *ev,
-                                    struct eventlog_batch **batchp)
+static flux_future_t *commit (struct eventlogger *ev,
+                              struct eventlog_batch **batchp)
 {
     struct eventlog_batch *batch = ev->current;
     if (batchp)
         *batchp = batch;
     ev->current = NULL;
-    return eventlogger_commit_batch (ev, batch);
+    return commit_batch (ev, batch);
 }
 
 flux_future_t *eventlogger_commit (struct eventlogger *ev)
 {
-    return commit_batch (ev, NULL);
+    return commit (ev, NULL);
 }
 
 static void timer_commit_cb (flux_future_t *f, void *arg)
@@ -216,7 +216,7 @@ static void timer_cb (flux_reactor_t *r,
     flux_future_t *f = NULL;
 
     batch->ev->current = NULL;
-    if (!(f = eventlogger_commit_batch (batch->ev, batch))
+    if (!(f = commit_batch (batch->ev, batch))
         || flux_future_then (f, timeout, timer_commit_cb, batch) < 0) {
         eventlog_batch_error (batch, errno);
         flux_future_destroy (f);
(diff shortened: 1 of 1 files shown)


src/common/libeventlog/eventlogger.c
```diff
@@ -178,19 +180,41 @@ static flux_future_t *commit (struct eventlogger *ev,
...
+         *
+         *  N.B. eventlog_batch_get() places this new batch on the
+         *  pending list.
+         */
+
+        if (!(batch = eventlog_batch_get (ev)))
+            return NULL;
+        if (batchp)
+            *batchp = batch;
+        ev->current = NULL;
```
grondo, line 216: This is a good catch!

I'm not sure how often an empty batch is actually committed. If it is a significant amount, perhaps we should explore other solutions that don't send an empty RPC.

For example, append empty batches to the pending list without sending an RPC. When non-empty batches complete, check if the next batch in the list and "complete" that batch as well. This would require saving the future to the eventlog batch object.

If empty batches are very rare, then the current solution here is simpler and is probably fine.

chu11: > I'm not sure how often an empty batch is actually committed. If it is a significant amount, perhaps we should explore other solutions that don't send an empty RPC.

It should be quite rare.  If you take a look at https://github.com/flux-framework/flux-core/pull/6581, the max "ops" that the kvs saw (within some sampling window) was 11K. 

Looking at corona right now, the max is 21 (not a typo, not 21K) :-)  Tuolomne it's 29K.  

Perhaps a minor bump up to .... 8K or 16K may be ok?

(Edit: I just realized, my comment covered a bit of a this one and follow on #7291)

> For example, append empty batches to the pending list without sending an RPC. When non-empty batches complete, check if the next batch in the list and "complete" that batch as well. This would require saving the future to the eventlog batch object.

> If empty batches are very rare, then the current solution here is simpler and is probably fine.

I originally implemented this much like you describe.  Just creating a dummy future with `flux_future_create()` and manually fulfilling it as needed.  

The fallout from this was `eventlog_flush()` and its call to `flux_future_wait_for()`.   Since there was no init function to the future, the internal "now" rpc was destined to hang forever.  Which is why it lead me down the path of sending an empty batch RPC (thus my comment of preserving all other aspects of the API).

src/common/libeventlog/eventlogger.c
```diff
@@ -164,7 +164,7 @@ static flux_future_t *eventlogger_commit_batch (struct eventlogger *ev,
          *   in big trouble anyway...
          */
         if ((f = flux_future_create (NULL, NULL))) {
-            flux_future_set_reactor (f, flux_get_reactor (ev->h));
+            flux_future_set_flux (f, ev->h);
```
grondo: Commit message typo: "Set the flux handle via flux_future_get_flux()" should be "flux_future_set_flux()"

src/common/libeventlog/eventlogger.c
```diff
@@ -39,6 +39,8 @@ struct eventlogger {
     void *arg;
 };
 
+static struct eventlog_batch * eventlog_batch_get (struct eventlogger *ev);
```
grondo, line 42: Commit message title: `libevent:` should be `libeventlog:`

src/common/libeventlog/eventlogger.c
```diff
@@ -140,62 +142,87 @@ static void eventlog_batch_error (struct eventlog_batch *batch, int errnum)
...
-        if (!(fc = flux_kvs_commit (ev->h, ev->ns, flags, batch->txn)))
+
+        if (!(batch = eventlog_batch_get (ev)))
             return NULL;
-        if (!(f = flux_future_and_then (fc, commit_cb, batch)))
-            flux_future_destroy (fc);
+        if (batchp)
+            *batchp = batch;
+        ev->current = NULL;
+        /* fallthrough */
```
grondo: Suggestion: expand this comment to describe what happens to the empty batch on fallthrough, e.g.:
```c
/* fallthrough: empty batch is now pending and will serialize behind previous batches */
```

grondo (approved):
LGTM! Minor comments inline.

chu11:
thanks!  will set MWP.

grondo:
Oh, I noticed the PR title has `libevent:` as well. maybe I'll just fix that so it doesn't get into the release notes.

PR #7273, opened 2026-01-11
fix lsf resolver for >1 broker per node

### LINKED ISSUE #6247: intermittent ci failure `not ok 25 - flux-uri mock testing of lsf resolver works`
```
 expecting success: 
  	result=$(SHELL=/bin/sh flux uri --local lsf:12345) &&
  	FLUX_URI_1=$(flux exec -r 1 flux getattr local-uri) &&
  	test_debug "echo checking if $result is $FLUX_URI or $FLUX_URI_1" &&
  	test "$result" = "$FLUX_URI" -o "$result" = "$FLUX_URI_1"
  
  hostname=lassen9
  cmd="/usr/src/src/cmd/flux uri --remote lsf:12345?is_compute"
  flux-uri: ERROR: PID 262732 is a flux-broker and no child found
  flux-uri: ERROR: Command '['/usr/src/t/scripts/tssh', 'lassen9', '/usr/src/src/cmd/flux', 'uri', '--remote', 'lsf:12345?is_compute']' returned non-zero exit status 1.
  not ok 25 - flux-uri mock testing of lsf resolver works
```

Likely a race condition in the test.

garlick:
Problem: the lsf resolver test intermittently fails in CI.

The lsf resolver searches for LSB_JOBID in the environment of any
flux-broker process it finds on the target node, then calls the
pid resolver.  But if there are multiple brokers per node, as there
are in the test, the lsf resolver might find a rank other than the
first one, which is not guaranteed to have children, and the pid
resolver fails with e.g.
```
flux-uri: ERROR: PID 262732 is a flux-broker and no child found
```
Restrict the search to processes named flux-broker-0.
Simplify the test logic, given this new constraint.

Also, improve the sharness test so it works if multiple copies are run simultaneously.

Fixes #6247

### DIFF
--- a/src/bindings/python/flux/uri/resolvers/lsf.py
+++ b/src/bindings/python/flux/uri/resolvers/lsf.py
@@ -67,7 +67,7 @@ def lsf_job_pid(jobid):
         check=True,
     )
     for line in sp.stdout.decode("utf-8").split("\n"):
-        if "flux-broker" in line:
+        if "flux-broker-0" in line:
             pid = line.split()[0]
             if check_lsf_jobid(pid, jobid):
                 return pid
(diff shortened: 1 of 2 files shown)


garlick:
Thanks!  Setting MWP.

grondo:
Restarted the coverage build which had timed out.

PR #7275, opened 2026-01-12
resource: fix undrain event generation

garlick:
Problem: the `resource.undrain` RPC handler posts malformed resource eventlog entries that swap nodelist and targets when targets are specified as a hostlist.
    
Recreate the nodelist and idset strings for the event rather than passing the targets through as the idset.  

Testing didn't catch this because we tend to use ranks rather than hostnames when there are multiple brokers per node, however `t2311-resource-drain.t` assigns a fake hostname to each broker so this type of test is easy to set up.  Add a test so this can't happen in the future.


grondo (approved):
LGTM! Thanks!!

garlick:
Thanks!  I guess I'll set MWP.  Sorry for the lame mistake!

PR #7283, opened 2026-01-14
doc: improve man page descriptions of broker logging

garlick:
Problem: the broker attribute descriptions for logging don't really provide enough context for someone to understand how to set them.

This adds a new descriptive section to flux-broker(1) then references it from flux-broker-attributes(7), flux-logger(1), and flux_log(3), and updates the text in those pages to fit and be a little bit more informative.

Fixes #7280

Issue #7282, opened 2026-01-14
default log attribute values cause unnecessary message traffic


garlick:
Problem: the default log attribute values cause _all_ log messages to be forwarded to rank 0, but only LOG_ERR and below (more severe) are emitted to stderr.  The others are discarded, unless `log-filename` is set.

These are the default values:
```
log-critical-level                      2
log-filename                            -
log-forward-level                       7
log-level                               7
log-ring-size                           1024
log-stderr-level                        3
log-stderr-mode                         leader
log-syslog-enable                       0
log-syslog-level                        3
```

Also FYI, the systemd unit file adjusts the following values from default:
```
log-stderr-level                        6
log-stderr-mode                         local
```

Possibly we should change the default forward level to match the stderr level
```
log-forward-level                       3
log-stderr-level                        3
```

and also in the system instance, set
```
log-forward-level                       0
```

Or, since that would allow LOG_EMERG messages to be forwarded perhaps allow a value of -1 for thresholds.

garlick:
Heh, the leader currently receives 22 log messages per follower on startup.  On El cap that's something like 244992 unnecessary messages.

PR #7284, opened 2026-01-14
broker: avoid unnecessary log message forwarding

garlick:
Problem: as discussed in #7282, the default `log-forward-level` is 7, which means all messages are forward to the leader, but the leader discards them if they don't match `log-stderr-level` or have a logfile destination.  This results in unnecessary overlay traffic.

Change the default to match `log-stderr-level` (3)

Set it to -1 in the system instance, since `log-stderr-mode` is set to local.

This is currently based on
- #7283


garlick:
Thanks - rebased on current master and will set MWP.

PR #7288, opened 2026-01-16
shell: avoid incorrect core adjustment when specifying nodes and slots with `flux alloc` and `flux batch`

grondo:
This PR fixes incorrect application of the issue #7174 workaround in the job shell when using per-resource jobspecs with multiple slots per node, such as with `flux alloc` and `flux batch` when both `-N` and `-n` are specified.

The #7174 workaround addresses unequal core/gpu distribution across nodes and should only apply under specific conditions, including when slot count exceeds task count. However, `flux alloc` and `flux batch` always use one task per node, so multiple slots per node incorrectly triggered the workaround, reducing each node to a single slot and leaving the instance with only a subset of allocated cores.

The fix disables the workaround when per-resource is specified in the jobspec.


grondo:
Thanks! Setting MWP.

PR #7289, opened 2026-01-16
chore: add note to MAINTAINERS.md file

wihobbs:
In the core devs meeting this week, we discussed offering help to anyone who wants to update their name across many repositories, so here's a note offering to do that. If this works I'll PR it across all the files.

cc @grondo


garlick:
I think if you're already a maintainer across multiple repos, you probably either know how this works or know who to ask.  Plus the "we" in the problem description is most likely "you" in that case.

garlick:
Actually, would it address this issue if we got rid of the names and affiliations in the list and just left the github IDs?

wihobbs:
> Actually, would it address this issue if we got rid of the names and affiliations in the list and just left the github IDs?

I opened an issue proposing that (see #7290), using just github IDs could have other benefits too, like enabling auto-tagging of reviewers when a PR is opened.

PR #7291, opened 2026-01-16
libeventlog: cap eventlogger kvs txn ops

chu11:
Problem: The number of operations in a kvs txn generated by the eventlogger has no cap.  In the future when the KVS may have a transaction operations cap, this could cause problems (see #6581).

If the number of operations in a eventlogger batch exceeds a cap, initiate the transaction immediately, even if a timer has not been reached.

For the time being, set the cap to a hard coded value.  This may become configurable in the future.

This is built upon #7271


grondo (approved):
LGTM!

PR #7293, opened 2026-01-18
broker: clean up state machine

garlick:
Problem: the addition of built-in modules overloaded some states in the broker state machine.  Specifically
-  the NONE state now effectively has actions (load the built-in modules), when originally it was just intended to be a "nothing" state (the starting point on the state diagram).
-  the EXIT state now unloads built-in modules (asynchronously) before exiting the reactor

Solution:
- rename the NONE state to PREINIT and give it a proper action callback like the other states
- add a POSTFINAL state which unloads the built-in modules

Update the [broker internals guide](https://flux-framework--7293.org.readthedocs.build/projects/flux-core/en/7293/guide/broker.html) and improve the readability of the state diagrams.


doc/guide/images/states.dot
```diff
@@ -2,35 +2,38 @@
...
+    S -> 3 [ label = "children-timeout", color = "red" ];
     3 -> G [ label = "goodbye" ];
-    G -> E [ label = "rc3-none" ];
-    G -> E [ label = "rc3-success" ];
-    G -> E [ label = "rc3-fail" ];
+    G -> F [ label = "rc3-none" ];
+    G -> F [ label = "rc3-success" ];
+    G -> F [ label = "rc3-fail", color = "red" ];
+    F -> E [ label = "builtins-done" ]
+    F -> E [ label = "builtins-unload-fail", color = "red" ]
```
grondo: Consistency: no semicolons after these edges?

doc/guide/broker.rst
```diff
@@ -264,28 +274,36 @@ and immediately generates an event.
...
 
 Normal State Transitions
 ========================
 
 It may be helpful to walk through the state transitions that occur when
 a Flux instance runs to completion without encountering exceptional conditions.
 
-.. image:: images/states_norm.png
-  :scale: 100 %
+In the diagram blow, the edges for the common path are shown in black.
```
grondo: Typo: blow -> below

grondo (approved):
This LGTM.

I did have one general take-it-or-leave it comment. The new states do not seem as descriptive as existing states, and especially I found the `postfinal` state name slightly confusing (because of the overloading of "post" to mean both after and to post an event). Since the states appear in the logs, did you consider just naming these two new states something like `load-builtins` and `unload-builtins` for clarity?

Feel free to ignore the comment if there may be actions unrelated to builtins in the future in these two states.

garlick:
> Since the states appear in the logs, did you consider just naming these two new states something like `load-builtins` and `unload-builtins` for clarity?

Good suggestion!  Will fix.

garlick:
Thanks - setting MWP

PR #7294, opened 2026-01-19
avoid potential `flux job attach` hangs with MPIR tools and add graceful job shell rexec server shutdown

### LINKED ISSUE #7269: Tool process shutdown issues - no grace period and flux-job attach not terminating
Thanks for adding _Launching a tool process within a flux job_ (using `MPIR_executable_path` etc.) in https://github.com/flux-framework/flux-core/discussions/5725. We have performed some initial testing and found the following issues:

1. Tool processes are not given sufficient grace period once application processes have terminated.
2. Despite (1) it seems that `flux-job attach` is not terminating at the end of a run.

I will attach a reproducer script using GDB in the next comment. It might also be that we incorrectly use `MPIR_executable_path`, in which case feedback about its correct use would be appreciated.

**1. Tool processes are not given a sufficient grace period once application processes have terminated**

Hence, tool processes are unable to detect that application processes have terminated and propagate this information to the tool UI. Would it be possible to add a grace period before Flux terminates tool processes? Ideally, configurable - with a sufficiently large default value - expecting that tool processes will terminate on their own.

_How was it established that tool processes are not given a sufficient grace period?_

The tool process in the reproducer will loop forever to log the current time every 0.25s to a file. However, it will terminate upon common signals.

```
Arguments: test arg 1
Install signal handler for SIGHUP
Install signal handler for SIGQUIT
Install signal handler for SIGINT
Install signal handler for SIGTERM
2026-01-07T12:45:13,468855145+00:00
...
2026-01-07T12:45:18,011957900+00:00
```

Expected behaviour: As the application process is sleep 5 it could be expected that a tool process runs longer than 5s, i.e. until it finished execution.

Actual behaviour: Tool process runs for about 5s (see first and last timestamp) and is not terminated using a signal that can be handled (absence of signal handler invocation in the output).

**2. flux-job attach is not terminating at the end of a run**

Which might cause a tool UI not to terminate, as it will usually wait until any launched process has terminated - at least for a while, until it forcefully attempts to terminate them.

Expected behaviour: `flux-job attach`, and hence GDB to terminate.

Actual behaviour: `flux-job attach`, and hence GDB not terminating.

To verify `flux-job attach` is still running you can use for example the following (while the reproducer is still running):

```
pgrep gdb | xargs -n 1 pstree -Tp
gdb(514862)───flux-job(514872)
```

grondo:
This PR fixes a couple issues identified in #7269 related to `MPIR_executable_path` support:

1. When MPIR  tools are launched via `MPIR_executable_path`, they hold references on the `flux job attach` reactor, which may cause it to hang indefinitely waiting for the tools to exit even after the job finishes if the job shell is killed before it can complete the libsubprocess protocol.
2. The shell's rexec server destroys all subprocesses with SIGKILL on exit,  preventing tools from exiting cleanly and the rexec server from sending completions messages to the client.

Solution:
- In the job shell rexec server, implement graceful shutdown for subprocesses with SIGTERM, followed  by SIGKILL escalation after a configurable timeout (default 60s). This gives tools processes a chance to exit cleanly and notify `flux job attach`.
- Track MPIR tool processes in `flux job attach` and clean them up when the job `finish` event  is received, ensuring `flux job attach` exits promptly even if the job shell is killed before it fully exits or on node failure, etc.

Closes #7269

### DIFF
--- a/src/cmd/job/attach.c
+++ b/src/cmd/job/attach.c
@@ -1338,6 +1338,12 @@ void attach_event_continuation (flux_future_t *f, void *arg)
             ctx->exit_code = flux_job_waitstatus_to_exitcode (status, &error);
             if (ctx->exit_code != 0)
                 log_msg ("%s", error.text);
+
+            /* All shells have exited, so no more MPIR data is expected.
+             * Clean up MPIR state now to ensure flux-job attach exits
+             * properly when the job completes.
+             */
+            mpir_shutdown (ctx->h);
         }
     }
 
--- a/src/common/libsubprocess/server.c
+++ b/src/common/libsubprocess/server.c
@@ -1060,7 +1060,6 @@ void subprocess_server_destroy (subprocess_server_t *s)
         server_killall (s, SIGKILL);
         zlistx_destroy (&s->subprocesses);
         zhashx_destroy (&s->labels);
-        flux_future_destroy (s->shutdown);
         free (s->service_name);
         free (s->local_uri);
         if (s->has_sigchld_ctx)
@@ -1127,20 +1126,35 @@ void subprocess_server_set_auth_cb (subprocess_server_t *s,
     s->arg = arg;
 }
 
+static void shutdown_future_invalidate (void *arg)
+{
+    subprocess_server_t *s = arg;
+    s->shutdown = NULL;
+}
+
 flux_future_t *subprocess_server_shutdown (subprocess_server_t *s, int signum)
 {
     flux_future_t *f;
 
-    if (!s || s->shutdown != NULL) {
+    if (!s) {
         errno = EINVAL;
         return NULL;
     }
-    if (!(f = flux_future_create (NULL, NULL)))
+
+    /* Create a future that will be fulfilled when all server processes
+     * have exited. Arrange to have this future invalidated if it is destroyed
+     * by the client so their callback doesn't get unnecessarily called
+     * and also to allow this function to be called multiple times.
+     */
+    if (!(f = flux_future_create (NULL, NULL))
+        || flux_future_aux_set (f, NULL, s, shutdown_future_invalidate) < 0) {
+        flux_future_destroy (f);
         return NULL;
+    }
     flux_future_set_reactor (f, flux_get_reactor (s->h));
     flux_future_set_flux (f, s->h);
-    flux_future_incref (f);
     s->shutdown = f;
+
     if (zlistx_size (s->subprocesses) == 0)
         flux_future_fulfill (f, NULL, NULL);
     else
--- a/src/shell/rexec.c
+++ b/src/shell/rexec.c
@@ -116,9 +116,9 @@ error:
 }
 
 static int rexec_init (flux_plugin_t *p,
-                      const char *topic,
-                      flux_plugin_arg_t *arg,
-                      void *data)
+                       const char *topic,
+                       flux_plugin_arg_t *arg,
+                       void *data)
 {
     flux_shell_t *shell = flux_plugin_get_shell (p);
     struct shell_rexec *rexec;
@@ -135,9 +135,48 @@ static int rexec_init (flux_plugin_t *p,
     return 0;
 }
 
+static void shutdown_cb (flux_future_t *f, void *arg)
+{
+    flux_reactor_stop (flux_future_get_reactor (f));
+    flux_future_destroy (f);
+}
+
+static int rexec_exit (flux_plugin_t *p,
+                       const char *topic,
+                       flux_plugin_arg_t *arg,
+                       void *data)
+{
+    struct shell_rexec *rexec = flux_plugin_aux_get (p, "rexec");
+    if (rexec) {
+        flux_t *h = flux_shell_get_flux (rexec->shell);
+        flux_future_t *f;
+
+        /* Send SIGTERM to any subprocesses running in the server and
+         * tell the server to shutdown. Future will be fulfilled when
+         * all processes have exited (immediatebly if there are none).
+         *
+         * The reactor needs to be run for the future to be fulfilled,
+         * but the shell has exited flux_reactor_run() at this point,
+         * so call it explicitly here.
+         */
+        if (!(f = subprocess_server_shutdown (rexec->server, SIGTERM))
+            || flux_future_then (f, -1., shutdown_cb, NULL) < 0) {
+            shell_log_errno ("subprocess_server_shutdown");
+            flux_future_destroy (f);
+            return -1;
+        }
+
+        /* Run reactor until server is shutdown:
+         */
+        flux_reactor_run (flux_get_reactor (h), 0);
+    }
+    return 0;
+}
+
 struct shell_builtin builtin_rexec = {
     .name = FLUX_SHELL_PLUGIN_NAME,
     .init = rexec_init,
+    .exit = rexec_exit,
 };
 
 // vi:ts=4 sw=4 expandtab
--- a/src/shell/rexec.c
+++ b/src/shell/rexec.c
@@ -25,17 +25,21 @@
 
 #include "src/common/libsubprocess/server.h"
 #include "src/common/libutil/errprintf.h"
+#include "src/common/libutil/fsd.h"
 
 #include "builtins.h"
 #include "internal.h"
 #include "svc.h"
 #include "log.h"
 
+static double default_shutdown_timeout = 60.;
+
 struct shell_rexec {
     flux_shell_t *shell;
     subprocess_server_t *server;
     char *name;
     bool parent_is_trusted;
+    double shutdown_timeout;
 };
 
 static void rexec_destroy (struct shell_rexec *rexec)
@@ -80,6 +84,7 @@ static int rexec_auth_cb (const flux_msg_t *msg,
 static struct shell_rexec *rexec_create (flux_shell_t *shell)
 {
     struct shell_rexec *rexec;
+    const char *timeout = NULL;
 
     if (!(rexec = calloc (1, sizeof (*rexec))))
         return NULL;
@@ -92,6 +97,22 @@ static struct shell_rexec *rexec_create (flux_shell_t *shell)
     pid_t ppid = getppid (); // 0 =  parent is in a different pid namespace
     if (ppid > 0 && kill (getppid (), 0) == 0)
         rexec->parent_is_trusted = true;
+
+    rexec->shutdown_timeout = default_shutdown_timeout;
+
+    if (flux_shell_getopt_unpack (shell,
+                                  "rexec-shutdown-timeout",
+                                  "s",
+                                  &timeout) < 0) {
+        shell_log_errno ("invalid rexec-shutdown-timeout");
+        goto error;
+    }
+
+    if (timeout
+        && fsd_parse_duration (timeout, &rexec->shutdown_timeout) < 0) {
+        shell_log_errno ("failed to parse rexec-shutdown-timeout");
+        goto error;
+    }
 
     /* N.B. subprocess_server_create() registers the methods: exec, write,
      * kill, list, and disconnect.  Give the server its own namespace.  The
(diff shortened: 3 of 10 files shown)


src/cmd/job/attach.c
```diff
@@ -1338,6 +1338,12 @@ void attach_event_continuation (flux_future_t *f, void *arg)
             ctx->exit_code = flux_job_waitstatus_to_exitcode (status, &error);
```
garlick, line 1338: Is `shell:` the right prefix for the commit title?  Seems like it's all in `flux-job`.

src/shell/rexec.c
```diff
@@ -135,9 +135,48 @@ static int rexec_init (flux_plugin_t *p,
...
+                       void *data)
+{
+    struct shell_rexec *rexec = flux_plugin_aux_get (p, "rexec");
+    if (rexec) {
+        flux_t *h = flux_shell_get_flux (rexec->shell);
+        flux_future_t *f;
+
+        /* Send SIGTERM to any subprocesses running in the server and
+         * tell the server to shutdown. Future will be fulfilled when
+         * all processes have exited (immediatebly if there are none).
```
garlick: spelling: immediatebly

src/shell/rexec.c
```diff
@@ -135,9 +135,48 @@ static int rexec_init (flux_plugin_t *p,
...
+        if (!(f = subprocess_server_shutdown (rexec->server, SIGTERM))
+            || flux_future_then (f, -1., shutdown_cb, NULL) < 0) {
+            shell_log_errno ("subprocess_server_shutdown");
+            flux_future_destroy (f);
+            return -1;
+        }
+
+        /* Run reactor until server is shutdown:
+         */
+        flux_reactor_run (flux_get_reactor (h), 0);
```
garlick, line 227: To help me understand - is this `flux_reactor_run()` a recursive call into the shell's reactor or are the plugin exit callbacks called after the shell's `flux_reactor_run()` has already returned?

src/common/libsubprocess/server.c
```diff
@@ -1131,7 +1131,7 @@ flux_future_t *subprocess_server_shutdown (subprocess_server_t *s, int signum)
 {
     flux_future_t *f;
 
-    if (!s || s->shutdown != NULL) {
+    if (!s) {
         errno = EINVAL;
         return NULL;
     }
```
garlick, line 1142: If this is allowed then, `s->shutdown` should be destroyed before being reassigned.

src/shell/rexec.c
```diff
@@ -135,9 +135,48 @@ static int rexec_init (flux_plugin_t *p,
...
+         * The reactor needs to be run for the future to be fulfilled,
+         * but the shell has exited flux_reactor_run() at this point,
+         * so call it explicitly here.
+         */
+        if (!(f = subprocess_server_shutdown (rexec->server, SIGTERM))
+            || flux_future_then (f, -1., shutdown_cb, NULL) < 0) {
+            shell_log_errno ("subprocess_server_shutdown");
+            flux_future_destroy (f);
+            return -1;
+        }
```
garlick, line 223: I think destroying the future returned by `subprocess_server_shutdown()` will result in a double free when `subprocess_server_destroy()` is called.  That future is owned by the subprocess serer.

garlick (commented):
A few comments based on a quick skim

src/shell/rexec.c
grondo: Plugin `shell.exit` callbacks are called after the shell's `flux_reactor_run()` has returned.

src/shell/rexec.c
grondo: Ugh. Good catch! However, this is a bit of a trap because standard practice with most (all?) other API calls that return futures is that the caller owns the future and must destroy it.

I wonder if the proper fix here is for libsubprocess to set an aux item on the returned future that sets `s->shutdown = NULL` when the future is destroyed. This would keep the same semantics as other future-returning functions while also allowing the subprocess server to track the future without worrying if it is destroyed out from under it. Thoughts?

src/shell/rexec.c
grondo: This one was already fixed in my working version. Sorry I hadn't realized I'd pushed a previous version of the PR.

src/cmd/job/attach.c
grondo: Typo! Thanks.

src/shell/rexec.c
grondo: Ah, the server actually increfs the future before returning it:
```c
    flux_future_incref (f);
    s->shutdown = f;
```

So caller destroying the future is correct.  However, it might be better if caller destruction of the future triggers `s->shutdown = NULL` so the server doesn't call the callback after this point...

garlick: Apologies, I misread the code and missed that the reference was only held so the server could safely fulfill the future even if the caller has destroyed it.  Maybe something like this then?
```diff
diff --git a/src/common/libsubprocess/server.c b/src/common/libsubprocess/server.c
index 934b13603..c6dcc39ee 100644
--- a/src/common/libsubprocess/server.c
+++ b/src/common/libsubprocess/server.c
@@ -230,8 +230,11 @@ static void proc_delete (subprocess_server_t *s, flux_subprocess_t *p)
 
     zlistx_delete (s->subprocesses, handle);
 
-    if (zlistx_size (s->subprocesses) == 0 && s->shutdown)
+    if (zlistx_size (s->subprocesses) == 0 && s->shutdown) {
         flux_future_fulfill (s->shutdown, NULL, NULL);
+        flux_future_decref (s->shutdown);
+        s->shutdown = NULL;
+    }
 out:
     errno = saved_errno;
 }
@@ -1131,7 +1134,7 @@ flux_future_t *subprocess_server_shutdown (subprocess_server_t *s, int signum)
 {
     flux_future_t *f;
 
-    if (!s || s->shutdown != NULL) {
+    if (!s) {
         errno = EINVAL;
         return NULL;
     }
@@ -1139,12 +1142,14 @@ flux_future_t *subprocess_server_shutdown (subprocess_server_t *s, int signum)
         return NULL;
     flux_future_set_reactor (f, flux_get_reactor (s->h));
     flux_future_set_flux (f, s->h);
-    flux_future_incref (f);
-    s->shutdown = f;
     if (zlistx_size (s->subprocesses) == 0)
         flux_future_fulfill (f, NULL, NULL);
-    else
+    else {
         server_killall (s, signum);
+        flux_future_incref (f);
+        flux_future_decref (s->shutdown);
+        s->shutdown = f;
+    }
     return f;
 }
 
```
?  Also perhaps a note in the header
```diff
diff --git a/src/common/libsubprocess/server.h b/src/common/libsubprocess/server.h
index 3c28768c0..a0f4ff295 100644
--- a/src/common/libsubprocess/server.h
+++ b/src/common/libsubprocess/server.h
@@ -49,6 +49,14 @@ void subprocess_server_destroy (subprocess_server_t *s);
  * there are some, the orig. reactor must be allowed to run in order for the
  * shutdown to make progress.  Therefore this future should be tracked with
  * flux_future_then(), not flux_future_get() which would deadlock.
+ *
+ * This function may be called more than once, e.g. for SIGTERM->SIGKILL
+ * escalation.
+ *
+ * The server holds a reference on the future while it is active.  After
+ * fulfillment or when this function is called again for escalation, that
+ * reference is dropped.  The caller "owns" the returned future and should
+ * eventually destroy it.
  */
 flux_future_t *subprocess_server_shutdown (subprocess_server_t *s, int signum);
 
```

grondo: I think it is still dangerous to fulfill a future the caller has destroyed. This could result in use-after-free if the caller has also destroyed any context that would be referenced by the callback.

I was about to push this solution, which invalidates `s->shutdown` if the caller destroys the future.

```diff
From 22bfb526a30d0350cd0b4f8710f3a506b49530d1 Mon Sep 17 00:00:00 2001
From: "Mark A. Grondona" <mark.grondona@gmail.com>
Date: Tue, 20 Jan 2026 06:50:59 -0800
Subject: [PATCH 1/1] libsubprocess: invalidate shutdown future when caller
 destroys it

Problem: The subprocess server calls `flux_future_incref()` on the
shutdown future before returning it from subprocess_server_shutdown().
This prevents a double-free in case the caller destroys the future.
However, it could cause use-after-free because the callback could be
called after future destruction by the caller.

Do not incref the future returned from subprocess_server_shutdown()
so that if the caller destroys the future, the callback will never
fire. To ensure the server doesn't use the future after the caller
destroys it, set up an aux item destructor callback that nullifies
server->shutdown.

If subprocess_server_shutdown() is called again, try to destroy any
existing future in case it was not freed by the caller. Add a note to
the documentation comment that calling subprocess_server_shutdown() or
subprocess_server_destroy() invalidates any previous shutdown future.
---
 src/common/libsubprocess/server.c | 25 +++++++++++++++++++++++--
 src/common/libsubprocess/server.h |  7 +++++++
 2 files changed, 30 insertions(+), 2 deletions(-)

diff --git a/src/common/libsubprocess/server.c b/src/common/libsubprocess/server.c
index b93e9f453..8fa608c6f 100644
--- a/src/common/libsubprocess/server.c
+++ b/src/common/libsubprocess/server.c
@@ -1127,6 +1127,12 @@ void subprocess_server_set_auth_cb (subprocess_server_t *s,
     s->arg = arg;
 }
 
+static void shutdown_future_invalidate (void *arg)
+{
+    subprocess_server_t *s = arg;
+    s->shutdown = NULL;
+}
+
 flux_future_t *subprocess_server_shutdown (subprocess_server_t *s, int signum)
 {
     flux_future_t *f;
@@ -1135,12 +1141,27 @@ flux_future_t *subprocess_server_shutdown (subprocess_server_t *s, int signum)
         errno = EINVAL;
         return NULL;
     }
-    if (!(f = flux_future_create (NULL, NULL)))
+
+    /* Create a future that will be fulfilled when all server processes
+     * have exited. Arrange to have this future invalidated if it is destroyed
+     * by the client so their callback doesn't get unnecessarily called
+     * and also to allow this function to be called multiple times.
+     */
+
+    if (!(f = flux_future_create (NULL, NULL))
+        || flux_future_aux_set (f, NULL, s, shutdown_future_invalidate) < 0)
         return NULL;
     flux_future_set_reactor (f, flux_get_reactor (s->h));
     flux_future_set_flux (f, s->h);
-    flux_future_incref (f);
+
+    /* In case this function is called more than once, destroy any previous
+     * shutdown future. There is not a chance of double free here if the
+     * caller already destroyed the future since shutdown_future_invalidate()
+     * nullifies s->shutdown.
+     */
+    flux_future_destroy (s->shutdown);
     s->shutdown = f;
+
     if (zlistx_size (s->subprocesses) == 0)
         flux_future_fulfill (f, NULL, NULL);
     else
diff --git a/src/common/libsubprocess/server.h b/src/common/libsubprocess/server.h
index 3c28768c0..743f0bc80 100644
--- a/src/common/libsubprocess/server.h
+++ b/src/common/libsubprocess/server.h
@@ -49,6 +49,13 @@ void subprocess_server_destroy (subprocess_server_t *s);
  * there are some, the orig. reactor must be allowed to run in order for the
  * shutdown to make progress.  Therefore this future should be tracked with
  * flux_future_then(), not flux_future_get() which would deadlock.
+ *
+ * This function may be called more than once, e.g. to allow SIGKILL to be
+ * sent after a timeout of the initial call.
+ *
+ * Note: If this function is called again or the subprocess server object is
+ * destroyed, any outstanding future from a previous call is invalidated if it
+ * has not been destroyed already.
  */
 flux_future_t *subprocess_server_shutdown (subprocess_server_t *s, int signum);
 
-- 
2.34.1
```

grondo: Actually it might be better if the server just doesn't try to destroy the future at all and fully pushes ownership to the client. The comment can be updated to say the callback will not be called on any previous future if `subprocess_server_shutdown()` is called again.

grondo: I went ahead and pushed this solution as a new commit.

garlick: Sounds good.  Maybe add this note to the header then?
> If the returned future is destroyed before it is fulfilled, avoid use-after-free by destroying the subprocess server or ensuring that the reactor is stopped.

grondo: Not necessary. The aux item destructor nullifies `s->shutdown` so a destroyed future can't be fulfilled.

garlick: Oh, nice!  OK.

garlick (approved):
LGTM!  Sorry for the lack of rigor in my comments about the future refcount.

grondo:
No, that was a good call because I had missed that the server shared a reference with the caller.

If the current approach is acceptable, I'll squash the two libsubprocess commits if that makes sense.

garlick:
Sure, sounds good.

grondo:
Ok, I'll set MWP here to assist testing by @dirkschubert-linaro.

PR #7296, opened 2026-01-20
testsuite: fix t2318-resource-verify.t on non-RSMI system

grondo:
Problem: On a Turing RK1 system (Rockchip RK3588 based) running Ubuntu 22.04.5, t2318-resource-verify.t segfaults when the workaround for extra GPUs introduced in 56dfdd779aa2dbf0c5cef823b9e7a6bbe0b9d45a is active, i.e. HWLOC_COMPONENTS=-rsmi is set in the environment.

Since the workaround is only required on RSMI systems, only set the environment variable if the string "rsmi" appears in hwloc-ls output.


garlick (approved):
LGTM!

grondo:
Thanks! Setting MWP.b

Issue #7297, opened 2026-01-20
exec: add time-based limit for job termination before draining nodes


grondo:
The current job termination system uses `max-kill-count` to limit the number of attempts to kill a job before draining nodes with "unkillable user processes." While this provides fine-grained control, it makes it difficult for administrators to answer a simple operational question: "How long should the system wait before giving up and draining nodes?"

The termination sequence involves a complex escalation schedule:

 - Initial kill-timeout for SIGTERM → SIGKILL to tasks
 - Delay of `5*kill-timeout` before switching to signaling shells
 - Exponential backoff starting at `kill-timeout`, doubling each attempt (capped at 300s)
 - Limited by `max-kill-count` total attempts

To determine the actual wall-clock time before draining, an administrator must:
 - Understand the multi-phase escalation algorith
 - Calculate the sum of: task kill attempts + `5*delay` + exponential backoff series
 - Account for the 300s cap on individual timeouts
 - Adjust `max-kill-count` to achieve their desired total duration

This is error-prone and unintuitive. Common administrative questions like "give jobs 30 minutes to clean up before draining nodes" cannot be directly configured.

A simpler time-based cap on the amount of time jobs are given after an initial exception or termination signal should be added which overrides `max-kill-count` and is simpler for administrators to understand.

cc: @kkier

PR #7298, opened 2026-01-20
job-exec: add `max-kill-timeout` parameter

grondo:
This PR adds `exec.max-kill-timeout` as suggested in #7297. When `max-kill-timeout` is set, it overrides `max-kill-count` and gives administrators a way to more simply set an upper bound on the amount of time Flux will wait for jobs to terminate before draining nodes.

There's also a commit here that adds the prefix `sdexec reports` to the drain message when nodes are drained due to sdexec reported EDEADLOCK, to give admins an easier way to differentiate between nodes drained after `max-kill-count/timeout` vs a systemd unit timeout.

### DIFF
--- a/src/modules/job-exec/exec.c
+++ b/src/modules/job-exec/exec.c
@@ -357,7 +357,8 @@ static void error_cb (struct bulk_exec *exec, flux_subprocess_t *p, void *arg)
             snprintf (ranks, sizeof (ranks), "%d", rank);
             (void) jobinfo_drain_ranks (job,
                                         ranks,
-                                        "unkillable processes from job %s",
+                                        "sdexec reports %s for job %s",
+                                        flux_subprocess_fail_error (p),
                                         idf58 (job->id));
             bool critical = is_critical_rank (job, shell_rank);
 
--- a/doc/man5/flux-config-exec.rst
+++ b/doc/man5/flux-config-exec.rst
@@ -64,6 +64,32 @@ max-kill-count
    drains the node. The default is 8. See :ref:`job_termination` below for
    details.
 
+max-kill-timeout
+   (optional) The maximum amount of time in FSD to wait for a job to terminate
+   before draining nodes with unkillable processes. When set, this overrides
+   ``max-kill-count`` by continuing the kill signal escalation sequence until
+   the specified duration has elapsed since termination began. The default is
+   unset (use ``max-kill-count`` instead). See :ref:`job_termination` below
+   for details.
+
+   Example: ``max-kill-timeout = "30m"`` gives jobs up to 30 minutes to
+   respond to termination signals before affected nodes are drained.
+
+.. note::
+
+  Choosing between max-kill-count and max-kill-timeout:
+
+  Use ``max-kill-timeout`` when you want to specify how long to wait
+  before giving up on a job, such as enforcing a site policy that jobs
+  get 30 minutes to clean up before nodes are drained. This is simpler
+  and more intuitive than calculating the number of kill attempts needed.
+
+  Use ``max-kill-count`` when you want fine-grained control over the
+  number of escalation attempts regardless of timing, or when maintaining
+  backward compatibility with existing configurations.
+
+  When both are set, ``max-kill-timeout`` takes precedence.
+
 term-signal
    (optional) A string specifying an alternate signal to ``SIGTERM`` when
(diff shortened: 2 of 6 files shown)


src/modules/job-exec/exec.c
```diff
@@ -357,7 +357,8 @@ static void error_cb (struct bulk_exec *exec, flux_subprocess_t *p, void *arg)
             snprintf (ranks, sizeof (ranks), "%d", rank);
             (void) jobinfo_drain_ranks (job,
                                         ranks,
-                                        "unkillable processes from job %s",
```
garlick, line 360: Actually, a textual error message is returned by sdexec in this case:
```c       
        exec_respond_error (proc,
                            EDEADLK,
                            "Processes did not respond to SIGKILL."
                            " Abandoning unit as is.");
```
I _think_ this message is available via `flux_subprocess_fail_error (p)`.  Did we want to display that?

garlick (approved):
This looks good.  Just one comment which you can take or leave.

src/modules/job-exec/exec.c
grondo: That's not a bad idea. I do wonder if that is too much detail for the drain reason. Mainly I wanted to get "sdexec" or "systemd" in the message to differentiate between this condition and expiration of `max-kill-count` or `max-kill-timeout`.

If you feel strongly we could drain with "sdexec reports: Processes did not respond to SIGKILL. Abandoning unit as is", which does seem pretty clear.

grondo: Oh I guess it would also have "... for job f1234xyz"

garlick: I don't feel strongly about it.  Up to you.  Note that we could shorten the error message over in sdexec.c if we feel it's too verbose.

grondo: Ok, I've added a commit that shortens the error message in sdexec:b

```diff
diff --git a/src/modules/sdexec/sdexec.c b/src/modules/sdexec/sdexec.c
index eef8c0c9f..49131f45c 100644
--- a/src/modules/sdexec/sdexec.c
+++ b/src/modules/sdexec/sdexec.c
@@ -198,8 +198,7 @@ static void finalize_exec_request_if_done (struct sdproc *proc)
     if (proc->stop.timed_out) {
         exec_respond_error (proc,
                             EDEADLK,
-                            "Processes did not respond to SIGKILL."
-                            " Abandoning unit as is.");
+                            "SIGKILL timeout, unit may still be active");
     }
     else if (sdexec_unit_state (proc->unit) == STATE_INACTIVE
         && sdexec_unit_substate (proc->unit) == SUBSTATE_DEAD

```

and the drain message on `EDEADLK` becomes:

```diff
diff --git a/src/modules/job-exec/exec.c b/src/modules/job-exec/exec.c
index bd286f9dc..05e31ace4 100644
--- a/src/modules/job-exec/exec.c
+++ b/src/modules/job-exec/exec.c
@@ -357,7 +357,8 @@ static void error_cb (struct bulk_exec *exec, flux_subproces
s_t *p, void *arg)
             snprintf (ranks, sizeof (ranks), "%d", rank);
             (void) jobinfo_drain_ranks (job,
                                         ranks,
-                                        "unkillable processes from job %s",
+                                        "sdexec reports %s for job %s",
+                                        flux_subprocess_fail_error (p),
                                         idf58 (job->id));
             bool critical = is_critical_rank (job, shell_rank);

```

grondo:
The coverage build is now failing with:
```
2026-01-31T15:38:38.9517867Z Traceback (most recent call last):
2026-01-31T15:38:38.9518555Z   File "/home/runner/.local/bin/coverage", line 5, in <module>
2026-01-31T15:38:38.9519107Z     sys.argv[0] = sys.argv[0].removesuffix('.exe')
2026-01-31T15:38:38.9519675Z AttributeError: 'str' object has no attribute 'removesuffix'
```
`removesuffix()` was added in Ptyhon 3.9, so the coverage build may need to somehow get a Python update.

grondo:
Rebased on #7342 to (hopefully) fix the coverage upload errors. (That PR should be manually merged first)

garlick (approved):
The sdexec change LGTM

grondo:
Thanks! I'll set MWP.

PR #7299, opened 2026-01-20
testsuite: prevent lstopo window pop-up

garlick:
Problem: the `hwloc-ls` call recently added to `t2318-resource-verify.t` triggers an lstopo window pop-up on some systems.

This was observed on a system derived from Ubuntu 22.04. It does not happen on RHEL 8 based systems.

Add `--of console` to force console (stdout) mode.


grondo (approved):
LGTM! Thanks!

garlick:
Thanks - MWP set I will.

PR #7303, opened 2026-01-21
python: add functions for kvs raw writes

chu11:
Problem: The ability to write raw values to the KVS is missing in the python API.

Add put_raw() function and add KVSTxn method raw() to support raw writes.  Each takes a value of type bytes.

----

Not sure if requiring type of "bytes" is a good idea or bad idea?  But it seemed appropriate?

### DIFF
--- a/src/bindings/python/flux/kvs.py
+++ b/src/bindings/python/flux/kvs.py
@@ -170,7 +170,7 @@ def _get_kvstxn(flux_handle, _kvstxn):
     return _kvstxn
 
 
-def put(flux_handle, key, value, _kvstxn=None):
+def put(flux_handle, key, value, _kvstxn=None, raw=False):
     """Put data into the KVS
 
     Internally will stage changes until commit() is called.
@@ -179,8 +179,16 @@ def put(flux_handle, key, value, _kvstxn=None):
         flux_handle: A Flux handle obtained from flux.Flux()
         key: key to write to
         value: value of the key
+        raw: if True, write value as raw bytes instead of JSON-encoded data.
+             Value must be bytes or str (will be UTF-8 encoded if str).
     """
     _kvstxn = _get_kvstxn(flux_handle, _kvstxn)
+    if raw:
+        if not isinstance(value, bytes):
+            value = value.encode("utf-8")
+        RAW.flux_kvs_txn_put_raw(_kvstxn, 0, key, value, len(value))
+        return
+
     try:
         json_str = json.dumps(value)
         RAW.flux_kvs_txn_put(_kvstxn, 0, key, json_str)
@@ -432,9 +440,9 @@ class KVSTxn:
         self.clear()
         return Future(future)
 
-    def put(self, key, value):
-        """Put key=value in the KVS"""
-        put(self.fhdl, self._path + key, value, _kvstxn=self.txn)
+    def put(self, key, value, raw=False):
+        """Put key=value in the KVS, set raw=True to write raw values"""
+        put(self.fhdl, self._path + key, value, _kvstxn=self.txn, raw=raw)
 
     def mkdir(self, key):
         """Create a directory in the KVS"""
--- a/t/python/t0005-kvs.py
+++ b/t/python/t0005-kvs.py
@@ -112,31 +112,41 @@ class TestKVS(unittest.TestCase):
         self.assertFalse(flux.kvs.isdir(self.f, "testisdirkey"))
         self.assertFalse(flux.kvs.isdir(self.f, "not_a_key_i_made"))
 
-    def test_api_06_put_mkdir(self):
+    def test_api_06_put_raw_bytes(self):
+        flux.kvs.put(self.f, "txn_raw_bytes", "foobar".encode("utf-8"), raw=True)
+        flux.kvs.commit(self.f)
+        self.assertTrue(flux.kvs.exists(self.f, "txn_raw_bytes"))
+
+    def test_api_07_put_raw_string(self):
+        flux.kvs.put(self.f, "txn_raw_string", "foobar", raw=True)
+        flux.kvs.commit(self.f)
+        self.assertTrue(flux.kvs.exists(self.f, "txn_raw_string"))
+
+    def test_api_08_put_mkdir(self):
         flux.kvs.put_mkdir(self.f, "txn_mkdir")
         flux.kvs.commit(self.f)
         self.assertTrue(flux.kvs.exists(self.f, "txn_mkdir"))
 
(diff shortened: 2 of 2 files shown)


grondo:
Just an idle thought, but would it make more sense to add a `raw` boolean parameter  (default `raw=False`) to the existing `put()` function and `KVSTxn.put()` method, rather than adding new functions? Just because the C interface has them as separate functions doesn't mean the Python interface needs to. (I also observe that `put()` already supports raw put on `json.dumps()` failure...)

chu11:
> Just an idle thought, but would it make more sense to add a raw boolean parameter (default raw=False) to the existing put() function and KVSTxn.put() method, rather than adding new functions? Just because the C interface has them as separate functions doesn't mean the Python interface needs to. (I also observe that put() already supports raw put on json.dumps() failure...)

That's a pretty good idea.  Obviously from my core-dev brain, I tend to want to mirror the C API b/c I know the C API better.

Perhaps need a third party opinion. @wihobbs or @cmoussa1 or @jameshcorbett.  Any opinion here on? 

A) add Python `put_raw()` to go along with `put()`, `put_mkdir()`, `put_symlink()`, `put_unlink()`.  This largely mirrors the C API (`flux_kvs_txn_put()`, `flux_kvs_txn_put_raw()`, `flux_kvs_txn_mkdir()`, etc. ..)

B) Add a flag to `put()` that is (hypothetically) `raw=False` by default.

jameshcorbett:
I vote B) and I would recommend not type-checking the input with `isinstance` but instead just converting it to `bytes()` (if you even need to do that) and letting any errors propagate

cmoussa1:
I'll second @jameshcorbett's vote to just add a `raw=False` flag

chu11:
re-pushed going with a slightly alternate implementation.

- if user raw=True and user passed in bytes, pass in bytes to .
- otherwise call `value.encode("utf-8")` to get bytes value, let `AttributeError` propogate as needed

There were other possible choices, such as using pickle to encode everything that could conceptually be passed in, but settled on this as the best balanced choice.

src/bindings/python/flux/kvs.py
```diff
@@ -432,9 +441,13 @@ def commit_async(self, flags=0):
...
 
-    def put(self, key, value):
+    def put(self, key, value, raw=False):
         """Put key=value in the KVS"""
-        put(self.fhdl, self._path + key, value, _kvstxn=self.txn)
+        put(self.fhdl, self._path + key, value, _kvstxn=self.txn, raw=raw)
+
+    def raw(self, key, value):
+        """Put raw key=value in the KVS, value must be of type "bytes" """
+        put_raw(self.fhdl, self._path + key, value, _kvstxn=self.txn)
```
grondo: Is this method left over from the previous implementation?

chu11: gah!  sorry, I didn't re-push final branch (and I actually re-based incorrectly in my local branch, so caught that too)

src/bindings/python/flux/kvs.py
```diff
@@ -179,8 +179,17 @@ def put(flux_handle, key, value, _kvstxn=None):
...
+        raw: if True, convert to 'bytes' and write raw data
     """
     _kvstxn = _get_kvstxn(flux_handle, _kvstxn)
+    if raw:
+        if isinstance(value, bytes):
+            RAW.flux_kvs_txn_put_raw(_kvstxn, 0, key, value, len(value))
+            return
+        tmp = value.encode("utf-8")
+        RAW.flux_kvs_txn_put_raw(_kvstxn, 0, key, tmp, len(tmp))
+        return
```
grondo, line 190: Suggestion: This could be simplified by converting non-bytes value first:

```suggestion
    if raw:
        if not isinstance(value, bytes):
            value = value.encode("utf-8")
        RAW.flux_kvs_txn_put_raw(_kvstxn, 0, key, tmp, len(tmp))
        return
```

src/bindings/python/flux/kvs.py
```diff
@@ -179,8 +179,17 @@ def put(flux_handle, key, value, _kvstxn=None):
         flux_handle: A Flux handle obtained from flux.Flux()
         key: key to write to
         value: value of the key
+        raw: if True, convert to 'bytes' and write raw data
```
grondo: Only objects with an `encode()` method are converted to bytes. It might be clearer here to say:

```
    raw: if True, write value as raw bytes instead of JSON-encoded data.
             Value must be bytes or str (will be UTF-8 encoded if str).
             Raw values are retrieved as strings when read back from KVS.
```

src/bindings/python/flux/kvs.py
grondo: Also, did you consider @jameshcorbett's suggestion to just skip the conversion and let errors propagate for non-bytes values? That would simplify things a lot (the caller can do the necessary encoding)

grondo (approved):
A couple suggestions inline. Otherwise LGTM.

src/bindings/python/flux/kvs.py
chu11: I did, but `bytes` doesn't have an `encode()` method.  I felt that `raw` should minimally handle that.

grondo: By handle, you mean handle `str` as a special case? 

Other types like `int` `float` don't have `encode()` methods, so this will raise `AttributeError: object type X has no attribute 'encode'` for anything but `str` or `bytes`. Just making sure that is what's intended.

grondo: BTW, meant to say this is fine with me, was just checking expected behavior.

src/bindings/python/flux/kvs.py
grondo: Actually, I realized I'm not sure about the last statement above

src/bindings/python/flux/kvs.py
chu11: > By handle, you mean handle str as a special case?

Sorry, I meant "I felt that raw=True behavior should minimally handle that", i.e. the user should be allowed to pass in "bytes" as a special case.

chu11:
re-pushed with the minor tweaks discussed above

Edit: will set MWP

src/bindings/python/flux/kvs.py
grondo: Ah, that makes sense!

grondo:
Restarted the el9 build that failed with:
```
not ok 8 - flux-shell: multiple jobs, each want stdin
```

PR #7309, opened 2026-01-22
shell: fix stderr flow control bug

### LINKED ISSUE #7307: shell output can overwhelm leader broker
Problem: a job on corona was observed to bog down the leader broker with shell `write` messages.

Each chunk of job output is sent as a separate `write` RPC, which creates state on each router in the rpc_track hash.  The writes are to the shell on the first node of the job.  In this case, the number of outstanding RPCs was observed to grow to around 65M entries on the leader.

@grondo and I chatted about this offline and had a few ideas
- batch the output at each shell to reduce the number of write RPCs
- consider using NORESPONSE RPCs which would eliminate the rpctrack state, but not give any indication to the writing shell that something went wrong
- rate limit clients, allowing the stdout/err pipe of the user process to stall
- design a protocol for output that manages congestion

garlick:
Problem: the shell output plugin implements flow control but due to a typo, only stops/starts stdout leaving stderr uncontrolled.

Fix typo.

Fixes #7307

### DIFF
--- a/src/shell/output/client.c
+++ b/src/shell/output/client.c
@@ -115,11 +115,11 @@ static void output_client_control (struct output_client *client, bool stop)
 
             if (stop) {
                 flux_subprocess_stream_stop (p, "stdout");
-                flux_subprocess_stream_stop (p, "stdout");
+                flux_subprocess_stream_stop (p, "stderr");
             }
             else {
                 flux_subprocess_stream_start (p, "stdout");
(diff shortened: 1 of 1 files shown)


grondo (approved):
Oh, excellent find! I looked right at that code and didn't spot it.

garlick:
Thanks! I'll set MWP

PR #7315, opened 2026-01-23
broker: fix foreground process handling for interactive rc2

### LINKED ISSUE #7314: broker: rare "error bringing rc2 into foreground: Invalid argument" with test instances
Very occaisionally size=64 test instances on my system are getting the error:

```console
$ src/cmd/flux start -s 64
Jan 22 14:48:12.660851 PST 2026 broker.err[0]: error bringing rc2 into foreground: Invalid argument
```

This indicates failure of `tcsetgrp(3)`:

```c
        case FLUX_SUBPROCESS_STOPPED:
            /*
             *  If stdin is a tty and the broker is in a foreground process
             *  group, the subprocess may have stopped due to SIGTTIN/SIGTTOU.
             *  Attempt to bring the subprocess into the foreground and
             *  continue it. Set the foreground flag on the entry so that the
             *  broker knows to bring its own process group back into the
             *  foreground after this subprocess is complete.
             */
            if (isatty (STDIN_FILENO)
                && tcgetpgrp (STDIN_FILENO) == getpgrp ()) {
                entry->foreground = true;
                if (tcsetpgrp (STDIN_FILENO, flux_subprocess_pid (p)) < 0) {
                    flux_log_error (r->h,
                                    "error bringing %d (%s) into foreground",
                                    flux_subprocess_pid (p),
                                    entry->name);
                }
                else if (kill_async (p, SIGCONT) < 0)
                    flux_log_error (r->h, "error continuing %s", entry->name);
            }
```

I added the above debugging here to print the subprocess PID at this point to see if it was obviously invalid.

After many attempts the issue reproduced again:

```console
$ src/cmd/flux start -s 32
Jan 22 15:37:39.789376 PST 2026 broker.err[0]: error bringing -1 (rc2) into foreground: Invalid argument
```

So the subprocess fd is -1 at the time of failure, which seems wrong. Have to check carefully to see how that could happen. Since the state is `FLUX_SUBPROCESS_STOPPED` here you'd think it is reasonable to assume there's a pid associated with the subprocess object by this time.

grondo:
Problem: When an interactive rc2 is stopped by writing to the terminal, a race condition can occur where the STOPPED state is received before the RUNNING state (see issue #5083). In this case, the subprocess object for rc2 does not yet have a locally assigned pid, causing `tcsetpgrp()` to fail with "Invalid argument" when attempting to bring rc2 into the foreground.

Defer the foreground transition when the pid is unavailable. Record that the runat entry needs foreground status when handling the STOPPED state, then complete the transition in the RUNNING state once the pid becomes available.

Fixes #7314

### DIFF
--- a/src/broker/runat.c
+++ b/src/broker/runat.c
@@ -53,6 +53,7 @@ struct runat_entry {
     bool completed;
     bool interactive;
     bool foreground;
+    bool needs_foreground;
     runat_completion_f cb;
     void *cb_arg;
 };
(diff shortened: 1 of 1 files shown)


garlick (approved):
Nice!  This looks good to me.

grondo:
Thanks! I'll set MWP.

Issue #6952, opened 2025-08-06
ci: automate creation of `fluxrm/testenv` images when in-tree Dockerfiles change


grondo:
In an offline discussion with @trws, we realized it should be simple to provide a GHA workflow to rebuild the `fluxrm/testenv` containers whenever the Dockerfiles are changed in a PR.

Ideally:

 - If a Dockerfile changes, the new `fluxrm/testenv:<distro>` image is rebuilt locally before running tests so the changes are validated
 - On master, if a new testenv image was built, then push to docker hub
 - use the existing manifest step to create the manifest for multi-arch

Seems easy in principle, but I'm not sure about the auto-detection of Dockerfile changes.

An easier method might be to have a separate workflow that can be triggered manually after changes are made. This workflow could build the images and run tests before pushing to dockerhub to detect problems.

trws:
> Seems easy in principle, but I'm not sure about the auto-detection of Dockerfile changes.

This was the other thing I was thinking about earlier.  We don't currently do it in flux, but I had to do something similar in spack's CI at one point and used path-based change triggers to trigger workflow components.  We should be able to do the same thing.  [Here's](https://github.com/spack/spack-packages/blob/develop/.github/workflows/ci.yaml#L20-L61) a snippet from there of what I mean.

Short version is that we can use an action to get a true/false for whether anything in a set of patterns has been changed in a PR, then use that to trigger or skip actions. The real trick is going to be making the container available and used in the context of testing the PR without pushing it to the dockerhub repo until merged to master.

wihobbs:
I'm very uneducated on this so apologies for the stupid question but @trws or @grondo, can you explain why its important that:

1. the noble images for both amd64 and arm64 are built in the same manifest (a multi-platform build) and
2. we build mpich from a tarball from source?

It seems to me that having to emulate _everything_ (for the amd64 build running on an arm runner) slows things down a lot, as does building mpich from source. I had this running on an x86 runner last night and finally killed it after 5 hours 😬

grondo:
One of the goals of moving testenv image creation to GHA is so that we can build those images just like the flux-core images (on the builder of the native arch), then defer manifest creation to a final step (see the generate-manifest action). So this work would remove the need to do that all in one step and on a single architecture. 

I can't remember off the top of my head why we build mpich from source, but I'm guessing the distro packages are an old version or missing some features

trws:
[...]

PR #7316, opened 2026-01-23
.github: add manually-triggered workflow for rebuilding testenv containers

wihobbs:
This was suggested in #6952 as a first step toward an automated rebuild of the testenv containers, and it seemed worth starting the feedback loop here (and it helped to break up the work). 

It's a workflow that will give a drop-down menu of options for a container to rebuild, and when run, will rebuild two containers for that distro: arm64 and amd64, push both tags, and then push a manifest that contains both tags. An example workflow run is [available in my fork](https://github.com/wihobbs/flux-core/actions/runs/21270555534) and [an example of what this would look like on docker hub](https://hub.docker.com/r/wihobbs/testenv/tags) is available under my username. (Note: the failed run is because I built the containers as fluxrm/testenv but my user tried to push to wihobbs/testenv. I've tested previous runs tagged wihobbs/testenv and it worked fine.)

Note a quirk of `workflow_dispatch` triggered workflows: they will only register if the yaml file describing them is on the default branch. I iterated on this workflow by merging it into the `master` branch of my fork.

### DIFF
--- a/.github/workflows/testenv.yml
+++ b/.github/workflows/testenv.yml
@@ -1,0 +1,245 @@
+on: 
+  workflow_dispatch:
+    inputs:
+      distro:
+        description: 'Distro for which to rebuild testenv manifest'
+        required: true
+        default: 'noble'
+        type: choice
+        options:
+        - alpine
+#        - bookworm # reason: workflow doesn't yet support 32-bit
+        - el8
+        - el9
+        - fedora40
+        - focal
+        - jammy
+        - noble
+name: rebuild testenv containers using docker buildx
+jobs:
+  build-amd64-container:
+    name: build new testenv container for amd64
+    timeout-minutes: 90
+    runs-on: ubuntu-latest
+    env: 
+      DOCKERHUB_REPO: fluxrm
+    outputs:
+      image_digest: ${{ steps.digest.outputs.digest }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      
+      - name: Build and tar amd64 image
+        id: build
+        run: |
+          mkdir -p out
+          docker buildx build \
+            -o type=docker,dest=out/testenv-${{ inputs.distro }}-amd64.tar \
+            --platform=linux/amd64 \
+            --tag $DOCKERHUB_REPO/testenv:${{ inputs.distro }}-amd64 \
+            -f src/test/docker/${{ inputs.distro }}/Dockerfile .
+
+      - name: Upload amd64 container tarball
+        uses: actions/upload-artifact@v4
+        with:
+          name: testenv-${{ inputs.distro }}-amd64
+          path: out/testenv-${{ inputs.distro }}-amd64.tar
+          if-no-files-found: error
+  
+  build-arm64-container:
+    name: build new testenv container for arm64
+    timeout-minutes: 90
+    runs-on: ubuntu-24.04-arm
+    env: 
+      DOCKERHUB_REPO: fluxrm
+    outputs:
+      image_digest: ${{ steps.digest.outputs.digest }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      
+      - name: Build and tar arm64 image
+        id: build
+        run: |
+          mkdir -p out
+          docker buildx build \
+            -o type=docker,dest=out/testenv-${{ inputs.distro }}-arm64.tar \
+            --platform=linux/arm64 \
+            --tag $DOCKERHUB_REPO/testenv:${{ inputs.distro }}-arm64 \
+            -f src/test/docker/${{ inputs.distro }}/Dockerfile .
+
+      - name: Upload arm64 container tarball
+        uses: actions/upload-artifact@v4
+        with:
+          name: testenv-${{ inputs.distro }}-arm64
+          path: out/testenv-${{ inputs.distro }}-arm64.tar
+          if-no-files-found: error
+
+  test-amd64-container:
+    name: Test newly-built amd64 container
+    needs:
+      - build-amd64-container
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
+    env:
+      DOCKERHUB_REPO: fluxrm
+    steps:
+      - uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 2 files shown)


grondo:
This looks great! One question since I can't directly see the workflow_dispatch drop down and I'm not that familiar with it: Can you make multiple selections at once with this drop-down menu?

wihobbs:
No, but if you make one selection multiple times you can "queue them up" to run. Here's what the drop down menu looks like (please ignore the many failed debugging runs 😅):

image

wihobbs:
Apparently, GitHub workflows don't yet support a multiple-choice drop down: [actions/runner#2076](https://github.com/actions/runner/issues/2076)

grondo:
In the case of adding a new dependency we'll likely need to rebuild all testenv images and selecting the dropdown for every image sounds soul crushing. I wonder if there's a way to enable a multi-choice, or could you have a special "rebuild all" selection?

Alternately, maybe there's a way to programmatically start all builds with an API call?

Edit: sorry missed your update above!

wihobbs:
It's possible to start workflows from the [`gh` command line](https://cli.github.com/manual/gh_workflow_run), so we could come up with a quick one-liner to avoid crushing your soul @grondo. Let me work on that and I'll post here when I have it (it'll be good testing for the PR, too, since I only tested the `alpine` build.)

grondo:
In the meantime it seems like merging this sooner rather than later is a good plan. We can make changes as we go along and having the workflows available is a big win!

> It's possible to start workflows from the [gh command line](https://cli.github.com/manual/gh_workflow_run)

Oh cool. Maybe we could use this from the main workflow to automatically rebuild images when the Dockerfiles are modified. 

Nice work!

.github/workflows/testenv.yml
```diff
@@ -0,0 +1,256 @@
+# vim: set shiftwidth=2 tabstop=2 softtabstop=2 expandtab:
```
grondo: Suggestion: move this to the end of the file (more common practice)

.github/workflows/testenv.yml
```diff
@@ -0,0 +1,256 @@
...
+      - name: Checkout
+        uses: actions/checkout@v5
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      
+      - name: Build and tar arm64 image
+        id: build
+        run: |
+          set -euo pipefail
```
grondo: Is the pipefail necessary? There's no pipe I can see here

.github/workflows/testenv.yml
```diff
@@ -0,0 +1,256 @@
...
+
+      - name: Load images from tarballs
+        run: |
+          set -euo pipefail
+          docker load -i artifacts/testenv-${{ inputs.distro }}-amd64.tar
+          docker images | head -n 50
+
+      - name: run docker-run-checks
+        run: |
+          set -euo pipefail
```
grondo: no pipes here?

grondo (approved):
LGTM! Some very minor general comments inline but feel free to ignore them.

wihobbs:
I put together a short script to rebuild all the `testenv` containers (they're all running now on my [fork](https://github.com/wihobbs/flux-core/actions/workflows/testenv.yml)):
```
$ cat ./scripts/rebuild-all-testenv.sh
DISTROS="alpine bookworm el7 el8 el9 fedora33 fedora34 fedora35 fedora38 fedora39 fedora40 focal jammy noble"
for distro in $DISTROS; do
    gh workflow run testenv.yml -R wihobbs/flux-core --ref manual-testenv-build -f input="${distro}"
done
```
I'll s/wihobbs/flux-framework and make a separate commit to add this to the `/scripts` directory. 

Edit: oh, and drop the branch reference.

wihobbs:
In the last push:
- added the script above for rebuilding many containers at once to `/scripts`
- addressed the feedback from @grondo 
- dropped many containers that are no longer used (and had an old format for referencing `config.site` that was inconsistent with currently-used containers)

scripts/rebuild-all-testenv.sh
```diff
@@ -0,0 +1,4 @@
+DISTROS="alpine bookworm el8 el9 fedora40 focal jammy noble"
```
grondo: Suggestion: Add a check for `gh` here and emit an error and exit early if not found.

grondo: Also, I wonder if we can get the list of distros directly from src/test/docker. (This improvement could always be made later.)

.github/workflows/testenv.yml
```diff
@@ -0,0 +1,245 @@
...
+  workflow_dispatch:
+    inputs:
+      distro:
+        description: 'Distro for which to rebuild testenv manifest'
+        required: true
+        default: 'noble'
+        type: choice
+        options:
+        - alpine
+        - bookworm
```
grondo: As noted on the call, there's also a linux/386 image in the bookworm manifest.

I mentioned we could probably defer adding that, but actually I think we'll need to have that before merging this, otherwise the 32bit CI build will start failing if anyone triggers this workflow.

grondo (approved):
Excellent! Just a couple very minor suggestions. 

Also, I'm a bit wary of merging this before the i386 build of bookworm is supported because it could disrupt flux-core ci if someone triggers the workflow. However, we could merge this as-is without the final push to dockerhub just to keep things moving. Thoughts?

.github/workflows/testenv.yml
wihobbs: Sorry, I completely missed the 32-bit bookworm container 😞 

Could we drop bookworm off of the list of options for now and I can add that in a second go? I know that's not ideal, but it would give us a nice way of doing things for all the other container manifests, and avoid breaking the testsuite if the workflow were triggered. (FWIW I was planning to have a go at the bookworm containers anyway because I hit issues in my local build.)

grondo: Yeah, just comment it out with the reason why.

wihobbs:
Oh, right. This requires a manual merge. With approval from @grondo, doing that now.

wihobbs:
Ew that feels so uncivilized

grondo:
For future reference, [RFC 1](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_1.html) does state

> Maintainers SHOULD NOT merge their own patches except in exceptional cases.

In cases where mergify can't merge PRs, the process states that a maintainer other than the author(s) should merge the PR.

wihobbs:
Apologies, my bad!

PR #7318, opened 2026-01-23
libsubprocess: fix test portability issue on macos

### LINKED ISSUE #7301: occasional macos failure in CI:  not ok 129 - sleep: got expected status (got 0x0100, expected 0x000f
Problem: subprocess unit test occasionally fails on macos in CI

```
2026-01-21T01:10:48.0000210Z not ok 129 - sleep: got expected status (got 0x0100, expected 0x000f
2026-01-21T01:10:48.0000570Z FAIL: test_remote.t 129 - sleep: got expected status (got 0x0100, expected 0x000f
2026-01-21T01:10:48.0000960Z #   Failed test 'sleep: got expected status (got 0x0100, expected 0x000f'
2026-01-21T01:10:48.0001230Z #   at test/remote.c line 809.
2026-01-21T01:10:48.0001420Z # destroying subprocess server
```

grondo:
Problem: A test in `libsubprocess/test/remote.c` uses "sleep inf" to test background process signaling, but "inf" is a GNU extension not supported by macOS sleep. The test usually passes by chance because the sleep process is typically terminated before it can parse its arguments and exit with an error. However, when the process does start in time to check its arguments, the test fails with:

 not ok 129 - sleep: got expected status (got 0x0100, expected 0x000f)

Replace "inf" with "30" seconds, which is portable across all sleep implementations.

Fixes #7301

### DIFF
--- a/src/common/libsubprocess/test/remote.c
+++ b/src/common/libsubprocess/test/remote.c
@@ -815,7 +815,7 @@ void background_waitable_test (flux_t *h)
     char *cmd_noexist[] = { "/noexist", NULL };
     char *cmd_true[] = { "true", NULL };
     char *cmd_false[] = { "false", NULL };
-    char *cmd_sleep[] = { "sleep", "inf", NULL };
+    char *cmd_sleep[] = { "sleep", "30", NULL };
     bg_test (h, "noexist", 1, cmd_noexist, -ENOENT, false);
     bg_test (h, "success", 1, cmd_true, 0, false);
     bg_test (h, NULL, 1, cmd_true, 0, false);


chu11 (approved):
LGTM, although I can't help but think 30 seconds is too low.  Why not like 1000?  And then add a comment "sleep inf not portable".

grondo:
Thanks!

Embarrassingly, `sleep inf` was probably a mistake for this test even if it was portable. If the signal fails to be delivered, then the amount of sleep is essentially the test timeout. Given that, 1000 seems excessive. For the same reason adding a comment about `sleep inf` might be confusing.

garlick:
Restarting one CI builder that hit
- #7322

PR #7320, opened 2026-01-24
shell: fix output flow control bug

garlick:
Problem: shell output never resumes once flow control stops it.

A flag variable in `output_client_control()` is not being set to record the the stop, so a start is treated as a no-op.

Set the flag when the state changes.

Fixes #7319

Issue #7319, opened 2026-01-24
shell output flow control never restarts once high watermark is reached


garlick:
Problem:  if shell output flow control reaches the high water mark, no further output is sent.

This is trivially reproducable with reduced hwm/lwm values:

```diff
diff --git a/src/shell/output/client.c b/src/shell/output/client.c
index 4a8ae5f04..5d431b67d 100644
--- a/src/shell/output/client.c
+++ b/src/shell/output/client.c
@@ -35,8 +35,8 @@
 #include "info.h"
 #include "output/client.h"
 
-static const int shell_output_lwm = 100;
-static const int shell_output_hwm = 1000;
+static const int shell_output_lwm = 1;
+static const int shell_output_hwm = 10;
 
 struct output_client {
     flux_shell_t *shell;
```

Then just run
```
$ flux start -s2
$ flux run -N2 lptest
```
and it gets stuck until the exit-timeout expires.

garlick:
Note that this bug is all too easy to trigger even with default hwm/lwm settings.  I suppose it depends on timing including how fast the leader shell can respond, but on my development desktop I can easily trigger it with
```
$ flux start -s16
$ flux run -N16 flux lptest 79 2000 >/dev/null
34.811s: flux-shell[0]: FATAL: task-exit: flux: task rank 0 on host system76-pc exited and exit-timeout=30s has expired
34.826s: job.exception ƒ22uw5daw type=exec severity=0 flux: task rank 0 on host system76-pc exited and exit-timeout=30s has expired
```

Issue #7302, opened 2026-01-21
t2612-job-shell-pty.t occasionally hangs in CI


garlick:
Problem:  the job shell pty sharness test occasionally gets stuck here in CI:
```
2026-01-21T02:03:41.2114174Z        expecting success: 
2026-01-21T02:03:41.2114349Z 	for i in $(seq 1 3); do
2026-01-21T02:03:41.2114606Z 		id=$(flux submit -o pty.interactive -o pty.capture tty) &&
2026-01-21T02:03:41.2114927Z 		$runpty flux job attach $id >ptyim.out 2>&1 &&
2026-01-21T02:03:41.2115187Z 		$runpty flux job attach $id &&
2026-01-21T02:03:41.2115441Z 		flux job eventlog -f json -p output $id \
2026-01-21T02:03:41.2115690Z 			| tail -n1 >last-event.$i
2026-01-21T02:03:41.2115886Z 	done &&
2026-01-21T02:03:41.2116078Z 	# Check that eof:true is the last event for all runs
2026-01-21T02:03:41.2116490Z 	cat last-event.1 | jq -e .context.eof &&
2026-01-21T02:03:41.2116743Z 	cat last-event.2 | jq -e .context.eof &&
2026-01-21T02:03:41.2117174Z 	cat last-event.3 | jq -e .context.eof
2026-01-21T02:03:41.2117385Z 
2026-01-21T02:03:41.2117533Z {"version": 2, "width": 80, "height": 25, "timestamp": 1768958163}
```

That test is titled `pty: -o pty.interactive and -o pty.capture can be used together`

The entire docker-run-checks then times out.

OK, I _think_ I've seen this before but I can't be completely sure it was this exact test.

grondo:
I was able to reproduce this locally and understand the issue.

When attached to a job with an interactive pty, `flux job attach` does not wait for the `clean` event before exiting, but exits immediately when the remote pty shuts down, i.e. when the session running under control of the remote pty exits.

In the test above this means that the second `flux job attach`, which is used to verify no hang attaching to an inactive interactive pty job, could run after the pty has shutdown but before the job shell has exited and unregistered the shell service. This opens a small window where the pty attach RPC from the new `flux job attach` could be accepted by the job shell but will never get a reply since the shell is in the midst of shutting down.

[...]

PR #7321, opened 2026-01-24
shell: improve output flow control efficiency

garlick:
Problem: follower shells count write RPC responses to limit the number of write requests in filght to the leader shell, but a large system running a job that produces significant output can become bogged down when the first node of the job is not the leader broker, due to the large amount of through-routing and RPC tracking the leader must perform.
    
Switch to a more efficient credit-based flow control protocol that uses FLUX_RPC_NORESPONSE on the write requests:

1. each shell is given write message credits to start equal to the  high watermark
   
2. write requests consume one credit each
    
3. when credits drop to the low watermark, a `write-getcredit` request is sent to the leader shell

4. when credits drop to zero, output is stopped

5. the leader responds with more credits (currently equal to the high watermark)

6. when credits are replenished, output is resumed
    
Note that a request for credit needs to make its way through the leader shell's message backlog before it can be fulfilled, so if the leader is  busy, the client naturally remains stopped for a longer time.
    
This is more or less equivalent to the existing scheme, except due to the elimination of responses, the message load is cut in half and  rpc tracking on the leader is eliminated.
    
A more sophisticated approach that scales back the number of credits issued under load could be implemented later, but for now this is likely a significant improvement.

This is based on top of
- #7320

### DIFF
--- a/src/shell/output/client.c
+++ b/src/shell/output/client.c
@@ -39,9 +39,10 @@ struct output_client {
     flux_shell_t *shell;
     int shell_rank;
     bool stopped;
-    zlist_t *pending_writes;
     int lwm;
     int hwm;
+    int credits;
+    flux_future_t *f_getcredit;
 };
 
 static void client_send_eof (struct output_client *client)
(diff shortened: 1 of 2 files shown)


garlick:
Restarted el8 builder after hitting
- #7302

grondo (approved):
Great! Huge improvement!

garlick:
Thanks!  Rebased on master after #7319 was merged and did some minor rewording of the commit message body.  I'll set MWP.

PR #7323, opened 2026-01-26
flux-broker-attributes(7): improve updatability marks, add CAVEATS

garlick:
Problem: the attributes man page employs single character marks to indicate whether an attribute can be updated on the command line or on a live system.  The C and R marks are not intuitive, and there is no mark for attributes that override TOML keys, which is becoming more prevalent.

Use sphinx citations instead of the single character marks.
Drop C (command line) and instead use a mark for the inverse: [readonly].
Replace R with [runtime].
Add a [config] mark.

Plus add a CAVEATS section that explains a few quirks about broker attributes.

### DIFF
--- a/doc/man7/flux-broker-attributes.rst
+++ b/doc/man7/flux-broker-attributes.rst
@@ -289,6 +289,9 @@ tbon.interface-hint [Updates: C, R]
    3. the enclosing Flux instance's value (via the PMI KVS)
    4. the compiled-in default of default-route
 
+   This attribute is inherited by sub-instances.  Changing it at runtime
+   affects new sub-instances, but not the current instance.
+
 tbon.torpid_min [Updates: C]
    The amount of time (in RFC 23 Flux Standard Duration format) that a broker
    will allow the connection to its TBON parent to remain idle before sending a
@@ -310,7 +313,7 @@ tbon.torpid_max [Updates: C]
    In older releases, they could be updated using :man1:`flux-setattr`, but
    that is no longer true.
 
-tbon.tcp_user_timeout
+tbon.tcp_user_timeout [Updates: C]
    The amount of time (in RFC 23 Flux Standard Duration format) that a broker
    waits for a TBON child connection to acknowledge transmitted TCP data
    before forcibly closing the connection.  A value of 0 means use the system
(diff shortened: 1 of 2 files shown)


doc/man7/flux-broker-attributes.rst
```diff
@@ -382,23 +383,35 @@ log-stderr-mode [Updates: C, R]
...
+
+.. [readonly] This attribute is strictly informational and is not meant to
+   be set by users.
+
+.. [config] This attribute overrides a TOML configuration key with a
+   similar name.
+
+.. [runtime] This attribute may be updated at runtime with
+   :man1:`flux-setattr`.
+
```
grondo, line 413: I wonder if there's a way to suppress the numeric references in the actual citation definitions as seen here?

Image

doc/man7/flux-broker-attributes.rst
```diff
@@ -289,6 +289,9 @@ tbon.interface-hint [Updates: C, R]
    3. the enclosing Flux instance's value (via the PMI KVS)
    4. the compiled-in default of default-route
 
+   This attribute is inherited by sub-instances.  Changing it at runtime
```
grondo, line 297: Commit message typo: 'command ilne' (last bullet)

doc/man7/flux-broker-attributes.rst
```diff
@@ -382,23 +383,35 @@ log-stderr-mode [Updates: C, R]
...
    The selected hash algorithm.  Default ``sha1``.  Other options: ``sha256``.
 
+ATTRIBUTE UPDATE NOTES
+======================
+
+.. [readonly] This attribute is strictly informational and is not meant to
+   be set by users.
+
+.. [config] This attribute overrides a TOML configuration key with a
+   similar name.
```
grondo: "a similar" -> "the same"?

grondo (approved):
Nice improvement! Just a couple inline comments.

doc/man7/flux-broker-attributes.rst
garlick: Eeek, those were at least all on one line in my local html rendering!

garlick: Couldn't figure out a way to do that, so I replaced the citations with regular sphinx references, which are a bit more verbose in the rst source, but not a big deal I think.

garlick:
Setting MWP, thanks.

PR #7327, opened 2026-01-26
flux-pmi: add --timeout option

garlick:
Problem: flux-pmi doesn't have a built-in timeout, but that could be handy.

Add a (general) `--timeout=FSD` option.  If the timer expires, the command fails.  This would be used like
```
flux pmi --timeout=35m barrier
```

It comes before the `barrier` subcommand because it can apply to any subcommand.

Also, since `flux pmi barrier` is  getting used outside of the test environment
- Don't do barrier timing (which has overhead) unless a new `--test-timing` option is specified
- Rename `--count` to `--test-count`
- Rename `--abort` to `--test-abort`

to make it clear that those options probably aren't useful in the real world.

### DIFF
--- a/src/cmd/builtin/pmi.c
+++ b/src/cmd/builtin/pmi.c
@@ -77,8 +77,10 @@ static int internal_cmd_barrier (optparse_t *p, int argc, char *argv[])
         log_msg_exit ("%s", error.text);
 
     // don't let task launch stragglers skew timing
-    if (upmi_barrier (upmi, &error) < 0)
-        log_msg_exit ("barrier: %s", error.text);
+    if (optparse_hasopt (p, "test-timing")) {
+        if (upmi_barrier (upmi, &error) < 0)
+            log_msg_exit ("barrier: %s", error.text);
+    }
 
     // abort one rank if --test-abort was specified
     if (test_abort != -1) {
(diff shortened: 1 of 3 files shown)


src/cmd/builtin/pmi.c
```diff
@@ -93,13 +97,16 @@ static int internal_cmd_barrier (optparse_t *p, int argc, char *argv[])
...
         if (info.rank == 0) {
-            printf ("%s: completed pmi barrier on %d tasks in %0.3fs.\n",
-                    label,
-                    info.size,
-                    monotime_since (t) / 1000);
+            printf ("%s: completed pmi barrier on %d tasks", label, info.size);
+            if (optparse_hasopt (p, "test-timing"))
+                printf (" in %0.3fs.\n", monotime_since (t) / 1000);
+            else
+                printf ("\n");
```
grondo: Could be simplified to:

```suggestion
            if (optparse_hasopt (p, "test-timing"))
                printf (" in %0.3fs.", monotime_since (t) / 1000);
            printf ("\n");
```

src/cmd/builtin/pmi.c
```diff
@@ -200,15 +207,39 @@ static void trace (void *arg, const char *text)
     fprintf (stderr, "%s\n", text);
 }
 
+static void sigalrm_handler (int arg)
+{
+    log_msg_exit ("timeout expired before PMI operation completed");
+}
+
```
grondo, line 218: This seems to work, but is `log_msg_exit()` safe in a signal handler?

grondo (approved):
LGTM!

src/cmd/builtin/pmi.c
garlick: Ah...no!  Good catch.  It calls printf() and exit(), neither of which are safe from signal handlers.

garlick:
Thanks!  Fixed those items and setting MWP.

Issue #7311, opened 2026-01-22
FAIL: test_stdio.t 365 - flux_subprocess_read returned correct data


garlick:
Problem: the stdio unit test for libsubprocess occasionally fails in CI

Example
```
2026-01-22T17:34:02.0326713Z not ok 365 - flux_subprocess_read returned correct data
2026-01-22T17:34:02.0327057Z FAIL: test_stdio.t 365 - flux_subprocess_read returned correct data
2026-01-22T17:34:02.0327416Z #   Failed test 'flux_subprocess_read returned correct data'
2026-01-22T17:34:02.0327706Z #   at test/stdio.c line 1447.
2026-01-22T17:34:02.0327980Z not ok 366 - flux_subprocess_read returned correct amount of data
2026-01-22T17:34:02.0328368Z FAIL: test_stdio.t 366 - flux_subprocess_read returned correct amount of data
2026-01-22T17:34:02.0328767Z #   Failed test 'flux_subprocess_read returned correct amount of data'
2026-01-22T17:34:02.0329083Z #   at test/stdio.c line 1450.
```

chu11:
just for context, this is within the `credit_output_cb()` callback, coming from the `test_on_credit()` set of tests.  Normally expect:

```
# on_credit
ok 350 - flux_cmd_create
ok 351 - set stdin buffer size to 8 bytes
ok 352 - flux_local_exec
ok 353 - flux_subprocess_aux_set works
ok 354 - subprocess state == RUNNING after flux_local_exec
ok 355 - flux_subprocess_write fails with too much data
# on_credit: credit of 8 bytes
ok 356 - flux_subprocess_write success
# on_credit: credit of 8 bytes
ok 357 - flux_subprocess_write success
# on_credit: credit of 8 bytes
ok 358 - flux_subprocess_write success
# on_credit: credit of 8 bytes
ok 359 - flux_subprocess_write success
# on_credit: credit of 8 bytes
ok 360 - flux_subprocess_write success
# on_credit: credit of 4 bytes
ok 361 - flux_subprocess_close success
ok 362 - flux_subprocess_read on stdout success
ok 363 - flux_subprocess_read on stdout success
ok 364 - flux_subprocess_read_stream_closed saw EOF on stdout
ok 365 - flux_subprocess_read returned correct data
ok 366 - flux_subprocess_read returned correct amount of data
ok 367 - subprocess state == EXITED in completion handler
ok 368 - subprocess status is valid
[...]

PR #7329, opened 2026-01-27
libsubprocess/test: add test instrumentation

chu11:
Problem: A credit test in test/stdio.c is failing occasionally in CI and it is presumed a non-displayable character is sneaking into the output.

Change the output debugging to output in hex.


chu11:
closing, per a comment in #7311 .... I think I'm confident that I've figured out the problem.

Issue #7332, opened 2026-01-28
not ok 86 - fewer cores: allow-extra allows extra cores


garlick:
Problem: some tests in t3218-resource-verify have begun failing on my work desktop:
```
expecting success: 
	flux alloc -n1 -o cpu-affinity=off \
		--conf=resource.verify.core=allow-extra \
		flux resource drain -no {reason} > drain2.out &&
	test_debug "cat drain2.out" &&
	test_must_be_empty drain2.out

Jan 28 23:07:05.785445 UTC 2026 resource.err[0]: draining: rank 0 (morbo) has extra resources: gpu0
'drain2.out' is not empty, it contains:
rank 0 (morbo) has extra resources: gpu0
not ok 86 - fewer cores: allow-extra allows extra cores
#	
#		flux alloc -n1 -o cpu-affinity=off \
#			--conf=resource.verify.core=allow-extra \
#			flux resource drain -no {reason} > drain2.out &&
#		test_debug "cat drain2.out" &&
#		test_must_be_empty drain2.out
#	

expecting success: 
	flux alloc -n1 -o cpu-affinity=off \
		--conf=resource.verify.core=allow-missing \
		flux resource drain -no {reason} > drain3.out &&
	test_debug "cat drain3.out" &&
	grep "extra resources.*core" drain3.out

Jan 28 23:07:07.402227 UTC 2026 resource.err[0]: draining: rank 0 (morbo) has extra resources: core[1-11],gpu0
rank 0 (morbo) has extra resources: core[1-11],gpu0
ok 87 - fewer cores: allow-missing drains on extra cores

expecting success: 
	flux alloc -n1 -o cpu-affinity=off \
		--conf=resource.verify.core=ignore \
		flux resource drain -no {reason} > drain4.out &&
	test_debug "cat drain4.out" &&
	test_must_be_empty drain4.out

Jan 28 23:07:09.010801 UTC 2026 resource.err[0]: draining: rank 0 (morbo) has extra resources: gpu0
'drain4.out' is not empty, it contains:
rank 0 (morbo) has extra resources: gpu0
not ok 88 - fewer cores: ignore allows extra cores
#	
#		flux alloc -n1 -o cpu-affinity=off \
#			--conf=resource.verify.core=ignore \
#			flux resource drain -no {reason} > drain4.out &&
#		test_debug "cat drain4.out" &&
#		test_must_be_empty drain4.out
#	

expecting success: 
[...]

PR #7333, opened 2026-01-29
testsuite: work around hwloc gpu masking issue

garlick:
Problem: on a system with Nvidia GPU and hwloc 2.7.0, some tests in t3218-resource-verify.t are failing.

The failing tests are for cores not gpus, so just ignore the gpu counts by setting resource.verify.gpu=ignore.

Works around #7332

### DIFF
--- a/t/t2318-resource-verify.t
+++ b/t/t2318-resource-verify.t
@@ -6,15 +6,6 @@ test_description='Test resource module verify configuration'
 test -n "$FLUX_TESTS_LOGFILE" && set -- "$@" --logfile
 . `dirname $0`/sharness.sh
 
-# Note: On some systems RSMI GPU detection may not be influenced by
-# [CUDA|ROCR|HIP]_VISIBLE_DEVICES, which causes some tests below to fail
-# with "extra" GPUs. Since we're not testing GPU detection here, just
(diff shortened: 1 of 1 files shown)


t/t2318-resource-verify.t
```diff
@@ -976,6 +976,7 @@ test_expect_success MULTICORE 'fewer cores: strict drains on extra cores' '
 test_expect_success MULTICORE 'fewer cores: allow-extra allows extra cores' '
 	flux alloc -n1 -o cpu-affinity=off \
 		--conf=resource.verify.core=allow-extra \
+		--conf=resource.verify.gpu=ignore \
```
grondo, line 970: Attempts to work around this same issue on rsmi systems were made in commit https://github.com/flux-framework/flux-core/commit/56dfdd779aa2dbf0c5cef823b9e7a6bbe0b9d45a and then #7296. This fix supersedes those fixes so the block that sets `HWLOC_COMPOENTS=-rsmi` when necessary at the top of this file can be removed in this commit.

grondo (approved):
Thanks! Just made a note to remove the previous fix for the same problem on some RSMI systems from the top of the file.

garlick:
Thanks - I added that and will set MWP.

PR #7335, opened 2026-01-29
flux-job: fix rare hang in attach with interactive pty

grondo:
This PR fixes a potential hang in `flux job attach` with an interactive pty when the job shell is in the process of exiting. This was the likely cause of #7301, since in that test two invocations of `flux job attach` are run in quick succession (the 2nd was the cause of the hangs)

Additionally, some timeouts are added to `t2612-job-shell-pty.t` to turn any future hangs into errors.

### DIFF
--- a/src/cmd/job/attach.c
+++ b/src/cmd/job/attach.c
@@ -118,6 +118,7 @@ struct attach_ctx {
     bool readonly;
     bool unbuffered;
     bool output_started;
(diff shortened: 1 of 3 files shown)


garlick (approved):
LGTM!

grondo:
Thanks! I'll set MWP.

Issue #7336, opened 2026-01-29
`SHELL=/invalid/path flux start` causes hang, Ctrl-C doesn't work to exit


grondo:
I was trying `SHELL=/bin/tcsh flux start` to test an issue, but my system didn't have `tcsh` installed. Instead of an error I got a hang which ignores `SIGINT`:

Reproducible with any file that doesn't exist (or would otherwise get an exec error)
```sh
 grondo@pi3:~$ SHELL=/bin/foo flux start
^CJan 29 13:41:18.270610 PST 2026 broker.err[0]: kill rc1: No such process
^CJan 29 13:41:19.217141 PST 2026 broker.err[0]: kill rc1: No such process
^CJan 29 13:41:19.523972 PST 2026 broker.err[0]: kill rc1: No such process
```

Haven't looked into this too deeply, but currently the broker ignores `FLUX_SUBPROCESS_FAILED` state update, I assume because there's an assumption that the completion callback will be called for all subprocesses, but that might not be true on `exec()` failure.

grondo:
With this patch the hang is avoided:
```diff
diff --git a/src/broker/runat.c b/src/broker/runat.c
index dcf3efae5..b4f2e988b 100644
--- a/src/broker/runat.c
+++ b/src/broker/runat.c
@@ -175,6 +175,10 @@ static void completion_cb (flux_subprocess_t *p)
         rc = signum + 128;
         log_command (r->h, entry, rc, elapsed, strsignal (signum));
     }
+    else if (flux_subprocess_state (p) == FLUX_SUBPROCESS_FAILED) {
+        rc = 127;
+        log_command (r->h, entry, rc, elapsed, flux_subprocess_fail_error (p));
+    }
     else { // ???
         rc = 1;
         log_command (r->h, entry, rc, elapsed, "???");
@@ -259,7 +263,9 @@ static void state_change_cb (flux_subprocess_t *p,
     switch (state) {
         case FLUX_SUBPROCESS_INIT:
         case FLUX_SUBPROCESS_EXITED:
+            break;
         case FLUX_SUBPROCESS_FAILED:
+            completion_cb (p);
             break;
         case FLUX_SUBPROCESS_STOPPED:
             /* STOPPED may arrive before RUNNING due to a protocol race
```

```console
$ SHELL=/bin/foo flux start
[...]

PR #7337, opened 2026-01-29
broker: fix hang on rc exec errors

grondo:
This PR fixes the hang reported in #7336.

Handling for the `FLUX_SUBPROCESS_FAILED` state is added to the broker runat subsystem. When an rc process fails, immediately call the completion callback. Add handling to the completion callback for the failed state.


garlick (approved):
Wow, I'm shocked we've not found this until now!
Good catch!  LGTM.

grondo:
The problem is very rare currently since the broker prepends rc commands with `$SHELL -c`, so as long as `SHELL` is valid there issue does not occur. 

I'll set MWP.

Issue #7330, opened 2026-01-28
--signal=USR1@time also sends SIGTERM for csh / tcsh users


ryanday36:
I was working with a user who was trying to use the `--signal=SIG@TIME` flag in his jobs to trigger a cleanup. I have a python script that catches various signals and does 10 minutes of "work" (sleeps for 10 minutes) before exiting. I put this in a bash flux batch script with `#flux: --signal=USR1@12m` and a `trap` command to avoid having the script exit when it catches the SIGUSR1. This all worked as expected for me, but it wasn't working for the user. I noticed that the user's default shell is `/bin/csh` whereas I am a bash user. I tried running my test as a test user that has `csh` as its default shell (`testnewm`), and found that the program catches a SIGTERM immediately after the SIGUSR1.

The script looks like:
```
rzadams1001{testnewm}59: cat flux_catcher2.batch 
#!/bin/bash
#flux: -N1
#flux: -t 13m
#flux: --signal=USR1@12m

trap 'echo "caught one";' USR1

flux run -n1 ./sig_catcher.py
rzadams1001{testnewm}60:
```
when I run it as myself, the output from the sig_catcher.py looks like:
```
[day36@rzadams1017:signal_test]$ cat flux-f3RPNJio3BcT.out 
My PID is: 1579177
Waiting...
Waiting...
Waiting...
Waiting...
Waiting...
Waiting...
flux-shell[0]: (null): signal: job will expire in 720.0s, sending SIGUSR1 to job
Received: 10
Received: 10
slept 1m
slept 1m
...
slept 10m, now I'll stop
flux-job: task(s) exited with exit code 1
caught one
Jan 27 17:12:12.223878 PST 2026 broker.err[0]: rc2.0: /var/tmp/day36/jobtmp-0-f3RPNJio3BcT/script Exited (rc=1) 652.0s
```
when I run it as the csh user, it looks like:
```
rzadams1001{testnewm}61: tail -f flux-f3RPQegGo7uZ.out 
My PID is: 3040273
Waiting...
Waiting...
Waiting...
Waiting...
Waiting...
flux-shell[0]: (null): signal: job will expire in 720.0s, sending SIGUSR1 to job
Jan 27 17:07:21.013123 PST 2026 broker.err[0]: rc2.0: /var/tmp/testnewm/jobtmp-0-f3RPQegGo7uZ/script User defined signal 1 (rc=138) 50.9s
Received: 10
Received: 10
[...]

PR #7339, opened 2026-01-30
broker: skip `$SHELL -c` wrapper for executable files

grondo:
This PR implements a small optimization in the broker runat logic to skip the `$SHELL -c` wrapper if the provided command line:
 - contains no whitespace
 - points to a regular file
 - the file is executable

In this case interpreting the "command line" by the shell is unnecessary so skip it.

This works around the issue reported in #7330 since there won't be an extra `tcsh` parent for batch scripts.

As noted in #7330, this will make it an error to point rc2 at an executable script that doesn't have a shebang, whereas that works now. I don't think this will be an issue, but it did require fixing a few autogenerated scripts in the testsuite.

### DIFF
--- a/t/t0001-basic.t
+++ b/t/t0001-basic.t
@@ -17,6 +17,9 @@ test -n "$FLUX_TESTS_LOGFILE" && set -- "$@" --logfile
 # the duration of this test to avoid these failure
 export MAN_DISABLE_SECCOMP=1
 
+# An unexpected umask could affect tests below. Set it explicitly here:
+umask 022
+
 startctl=${SHARNESS_TEST_SRCDIR}/scripts/startctl.py
 path_printenv=$(which printenv)
 
@@ -204,6 +207,42 @@ test_expect_success 'flux-start fails with bad SHELL' '
 	( export SHELL=/no/such/path &&
 	  test_expect_code 127  run_timeout 30 flux start flux uptime
 	)
+'
+test_expect_success 'flux-start fails with non-executable rc2' '
+	cat <<-EOF >no-shebang.sh &&
+	true
+	EOF
+	test_expect_code 126 run_timeout 30 flux start $ARGS ./no-shebang.sh
+'
+test_expect_success 'flux-start fails with invalid shebang in rc2' '
+	chmod +x no-shebang.sh &&
+	test_expect_code 126 run_timeout 30 flux start $ARGS ./no-shebang.sh
+'
+test_expect_success 'flux-start does not use $SHELL -c for a script' '
+	cat <<-EOF >fail.sh &&
+	#!/bin/sh
+	exit 1
+	EOF
+	chmod +x fail.sh &&
+	cat <<-'EOF' >test-script.sh &&
+	#!/bin/sh
+	exit 0
+	EOF
+	chmod +x test-script.sh &&
+	SHELL=$(pwd)/fail.sh flux start $ARGS $(pwd)/test-script.sh
+'
+test_expect_success 'flux-start does use $SHELL -c otherwise' '
+	cat <<-EOF >test-shell.sh &&
+	#!/bin/sh
+	shift # drop -c
+	echo "broker ran \$0 -c \$@"
+	\$@
+	EOF
+	chmod +x test-shell.sh &&
+	SHELL=./test-shell.sh flux start $ARGS "flux dmesg -H" \
+	    >test-shell.log 2>&1 &&
+	test_debug "cat test-shell.log" &&
+	grep "broker ran ./test-shell.sh -c" test-shell.log
 '
 test_expect_success 'create bad broker shell script' '
 	cat >flux-broker <<-EOT &&
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -28,6 +28,7 @@
 #include <sys/un.h>
 #include <sys/types.h>
 #include <sys/syscall.h>
+#include <sys/stat.h>
 #ifdef HAVE_ARGZ_ADD
 #include <argz.h>
 #else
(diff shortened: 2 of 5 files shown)


grondo:
Rebased on master after #7337 was merged.

src/broker/runat.c
```diff
@@ -498,16 +499,56 @@ static int runat_command_set_argz (struct runat_command *cmd,
...
+
+    /*  Check if file exists
+     */
+    if (stat (cmdline, &sb) < 0)
+        return true;
+
+    /*  Check if it's a regular file and executable
+     */
+    if (S_ISREG (sb.st_mode) && (sb.st_mode & S_IXUSR))
+        return false;
```
garlick: I think the file has to be readable and executable for the shell to run it?
Since there are also the group and other bits to consider, maybe it would be simpler to just do
```c
if (access (cmdline, R_OK | X_OK) == 0)
    return false;
```

t/t0001-basic.t
```diff
@@ -205,6 +205,16 @@ test_expect_success 'flux-start fails with bad SHELL' '
 	  test_expect_code 127  run_timeout 30 flux start flux uptime
 	)
 '
+test_expect_success 'flux-start fails with non-executable rc2' '
+	cat <<-EOF >no-shebang.sh &&
+	true
+	EOF
+	test_expect_code 126 run_timeout 30 flux start ./no-shebang.sh
```
garlick: Is the umask controlled here?  Maybe an explicit `chmod ugo=r` would make sense, if that's the intent?

garlick (approved):
Good idea!  Couple trivial suggestions inline.

src/broker/runat.c
garlick: Oh derp, well, I guess the point is not to find scripts that aren't readable.  Sorry.

t/t0001-basic.t
grondo: Are you worried about the file not being readable? Elsewhere throughout the test suite we don't assume a umask and only `chmod +x` after creating test scripts -- I'm probably missing something but why it is an issue specifically here?

src/broker/runat.c
grondo: Yeah, I wanted to restrict removal of the `$SHELL -c` prefix to a specific case for now to reduce the cases where there's a change in behavior. If the file is not readable then we still get `$SHELL -c $cmdline` and rc2 will continue to behave as it does now.

Edit: actually, I think your suggestion is correct since it will keep current behavior for non-readable files/scripts.

grondo: Hm, codecov is showing that the `return false;`  line here is not covered, which leads me to believe this isn't working as designed. Let me do some investigation before we consider merging this.

t/t0001-basic.t
garlick: Yeah I guess you're right.  For some reason, leaving the mode _entirely_ up to the umask and then assuming the execute bits are not set seemed more tenuous than doing +x and assuming they are, but  I wasn't thinking about the r bit which _is_ assumed.  Meh NM!

grondo: Well that does make sense. Since the `umask` will affect the rest of the test, I'll set it at the top to `022` like we do in other specific sharness tests when umask may affect the outcome.

grondo: Ok, I made that change as well as the one you suggested above. Still hold off on further review as I tested by hand and everything is working, but I still need to investigate why that one line doesn't appear to have coverage (now the function has changed a bit so we'll see)

It is difficult to add a test for the specific case here since `/bin/bash` always execs a single argument passed to `-c`, so there's no discernable difference between `/bin/bash -c` and direct execution. We could add a test with `tcsh` but we don't require that as a preqreq currently, and the test would depend on tcsh behavior that doesn't seem to be documented...

grondo:
Actually, there's a subtlety here I had missed. The broker only wraps rc2 in `$SHELL -c` if the command is not an interactive shell (empty command or one of the known shells, e.g. (bash, sh, csh, ...)), and it is only a single argument. Otherwise, it already uses `exec()` directly. 

https://github.com/flux-framework/flux-core/blob/f9209c2858b0e2bc0b285b8baa61c64ceec02fba/src/broker/broker.c#L831-L857

The test for whether a single argument can also use direct exec probably belongs here instead of inside `runat.c`.

I'll rework the fix. Sorry for the confusion.

grondo:
I've pushed an update that moves the test for whether `$SHELL -c` is required into `broker.c`  `create_runat_rc2()`. This new test replaces the original test that prepended `shell -c` whenever `argc == 1`. 

This now only applies only to rc2 (instead of all runat invoked commands) . The test whether a shell is required now looks like this:

```c
static bool shell_required (const char *argz, int argz_len)
{   
    struct stat st;
    
    /* Multiple arguments imply pre-parsed command-line components that
     * can be directly executed, so no shell required:
     */
    if (argz_count (argz, argz_len) != 1)
        return false;
    
    /* If the single argument does not contain whitespace and points to a
     * regular, readable, and executable file, then no shell required.
     * The broker can directly call the executable.
     */
    if (!contains_whitespace (argz)
       && stat (argz, &st) == 0
       && S_ISREG (st.st_mode) 
       && access (argz, X_OK | R_OK) == 0)
       return false;
    
    /* Otherwise, shell -c required
     */
    return true;
}
```

Note that I added back the `stat` check in addition to `access` because we want to restrict this change in behavior as much as possible so we don't include executable non-regular files like directories and devices.

The code has completely changed so this probably requires a re-review. I did figure out a way to add more tests.

grondo:
Thanks! Setting MWP here.

PR #7342, opened 2026-01-31
ci: update python version for use with coverage tool

grondo:
Problem: In github ci, the coverage report is failing with the error:
```
   File "/home/runner/.local/bin/coverage", line 5, in <module>
     sys.argv[0] = sys.argv[0].removesuffix('.exe')
   AttributeError: 'str' object has no attribute 'removesuffix'
```
which indicates that the coverage tool requires Python >= 3.9, but this is not available in the runner.

Install Python3.12 in the github runner when coverage is enabled to avoid the error.

This will need to be manually merged.


garlick (approved):
LGTM!

PR #7344, opened 2026-02-02
broker: drop FLUX_FAKE_HOSTNAME environment variable

garlick:
Problem: the FLUX_FAKE_HOSTNAME environment variable is used to implement `flux start --test-hosts=HOSTLIST`, but since that was implemented, a broker attribute named `hostname` was added.

It probably makes sense at this point to just allow `hostname` to be set on the broker command line and drop the environment variable.

Do that and update users.

If this is used in other projects at all, it is most likely via the `flux start --test-hosts` interface.

### DIFF
--- a/doc/man1/flux-start.rst
+++ b/doc/man1/flux-start.rst
@@ -166,10 +166,10 @@ OPTIONS
 
 .. option:: --test-hosts=HOSTLIST
 
-   Set :envvar:`FLUX_FAKE_HOSTNAME` in the environment of each broker so that
-   the broker can bootstrap from a config file instead of PMI.  HOSTLIST is
-   assumed to be in rank order.  The broker will use the fake hostname to
-   find its entry in the configured bootstrap host array.
+   Set the hostname attribute of each broker so that it can bootstrap from
+   a config file instead of PMI.  HOSTLIST is assumed to be in rank order.
+   The broker will use this hostname to find its entry in the configured
+   bootstrap host array.
 
 .. option:: --test-exit-timeout=FSD
 
--- a/doc/man7/flux-broker-attributes.rst
+++ b/doc/man7/flux-broker-attributes.rst
@@ -104,6 +104,11 @@ hostlist :ref:`[readonly] <attr_readonly>`
    An RFC 29 hostlist in broker rank order.  This value may be used to
    translate between broker ranks and hostnames.
 
+hostname
+   The system hostname.  This attribute is set on the broker command line by
+   :option:`flux start --test-hosts`, which is useful in certain test
+   scenarios.
+
 broker.mapping :ref:`[readonly] <attr_readonly>`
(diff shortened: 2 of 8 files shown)


garlick:
Ugh, sorry, I have something to fix here that I missed when testing locally.  Ignore this for the moment.

garlick:
OK, all better now I think.

doc/man7/flux-broker-attributes.rst
```diff
@@ -100,7 +100,7 @@ parent-kvs-namespace :ref:`[readonly] <attr_readonly>`
    This is the KVS namespace assigned to this Flux instance by its enclosing
    instance, if it was launched by Flux as a job.
 
-hostlist :ref:`[readonly] <attr_readonly>`
+hostlist
```
chu11, line 103: commit message typo "descripted" -> "described"

doc/man1/flux-start.rst
```diff
@@ -166,10 +166,10 @@ OPTIONS
 
 .. option:: --test-hosts=HOSTLIST
 
-   Set :envvar:`FLUX_FAKE_HOSTNAME` in the environment of each broker so that
-   the broker can bootstrap from a config file instead of PMI.  HOSTLIST is
-   assumed to be in rank order.  The broker will use the fake hostname to
-   find its entry in the configured bootstrap host array.
+   Set the hostname attribute of each broker so that it can bootstrap from
+   a config file instead of PMI.  HOSTLIST is assumed to be in rank order.
+   The broker will use the fake hostname to find its entry in the configured
```
chu11: this sounds weird now that the environment variable "FAKE" isn't there.  Instead of "fake hostname" like .... "specified hostname"?

Or perhaps a sentence saying "predominantly for testing" and now you can say "test hostname"?

src/cmd/flux-start.c
```diff
@@ -132,7 +132,7 @@ static struct optparse_option opts[] = {
       .usage = "Start a test instance by launching N brokers locally", },
     { .group = 2,
       .name = "test-hosts", .has_arg = 1, .arginfo = "HOSTLIST",
-      .usage = "Set FLUX_FAKE_HOSTNAME in environment of each broker", },
+      .usage = "Set fake hostnames for brokers", },
```
chu11, line 135: similar to comment above about "test hostname" instead of "fake hostname"

chu11 (approved):
LGTM, some nits and minor comments

doc/man1/flux-start.rst
garlick: Yeah, you're right - and actually, it seems reasonable to just shorten it.  How about
 > The system hostname.  This attribute is set on the broker command line by `flux start --test-hosts` , which is useful in certain test scenarios,

garlick:
I'll just go ahead and set MWP - hopefully that's OK

PR #7346, opened 2026-02-02
testsuite: tidy up broker logging configuration

garlick:
Problem: many sharness tests change `log-stderr-level` or `log-stderr-mode` via `flux-setattr(1)` (sometimes executing it on all ranks with `flux-exec`), when it would be more efficient to just set those attributes on the broker command line.

Change tests to set these logging attributes on the broker command line, via `test_under_flux`.

This also will make it a bit easier to change the way dynamic updates are performed on the log subsystem later on.


grondo (approved):
Seems fine to me!

Issue #7324, opened 2026-01-26
idea: require namespacing for new broker attributes


garlick:
Problem: broker attributes cannot be checked for misspelling because any attribute name could belong to a non-broker component.

Maybe we should lock down the broker attributes that we have so far by adding a table of names and types.  If a broker attribute is not namespaced with a  `NAME.`  prefix, it could be checked against the table.

Furthermore, we could add a mechanism for a component to fetch all attributes and values within their namespace and validate them at startup.

grondo:
See also [this comment](https://github.com/flux-framework/flux-core/issues/555#issuecomment-193514000)

PR #7348, opened 2026-02-02
broker: fail with a useful message when attribute names are misspelled on the command line

garlick:
Problem: when an attribute name is misspelled on the broker command line, there is no error.

During an early period of flux development, broker attributes were a handy way for components to be configured or to provide visibility into internal settings.  Since then, we've developed better mechanisms for those things, but broker attributes are still managed very loosely.

This adds a table of known attributes and
- prohibits setting attributes that are not matched in the table
- prohibits setting some known attributes on the command line, as documented in flux-broker-attributes(7)
- prints a useful error message when an attribute is misspelled or is on the command line when it shouldn't be

This change is proposed to partially remedy
- #7324

No mechanism is added for registering new namespaces, as discussed in that issue, but that could be added if need be.  (I wasn't too sure it was really required).

### DIFF
--- a/src/broker/attr.c
+++ b/src/broker/attr.c
@@ -209,6 +209,31 @@ int attr_add (attr_t *attrs, const char *name, const char *val, int flags)
     return 0;
 }
 
+int attr_set_cmdline (attr_t *attrs,
+                      const char *name,
+                      const char *val,
+                      flux_error_t *errp)
+{
+    struct registered_attr *reg;
+
+
+    if (!attrs || !name) {
+        errno = EINVAL;
+        return errprintf (errp, "invalid argument");
+    }
+    if (!(reg = attrtab_lookup (name)))
+        return errprintf (errp, "unknown attribute");
+    if ((reg->flags & ATTR_READONLY)) {
+        errno = EINVAL;
+        return errprintf (errp, "attribute may not be set on the command line");
+    }
+    if (attr_add (attrs, name, val, 0) < 0) {
+        if (errno != EEXIST || attr_set (attrs, name, val) < 0)
+            return errprintf (errp, "%s", strerror (errno));
+    }
+    return 0;
+}
+
 int attr_add_active (attr_t *attrs,
                      const char *name,
                      int flags,
--- a/src/broker/attr.c
+++ b/src/broker/attr.c
@@ -14,8 +14,12 @@
 #include <limits.h>
 #include <jansson.h>
 #include <inttypes.h>
+#include <fnmatch.h>
 
 #include "src/common/libczmqcontainers/czmq_containers.h"
+#include "src/common/libutil/errprintf.h"
+#include "src/common/libutil/errno_safe.h"
+#include "ccan/array_size/array_size.h"
 
 #include "attr.h"
 
@@ -38,7 +42,7 @@ struct entry {
     void *arg;
 };
 
-struct registered_attr attrtab[] = {
+static struct registered_attr attrtab[] = {
 
     // general
     { "rank", ATTR_READONLY },
@@ -128,6 +132,16 @@ struct registered_attr attrtab[] = {
     { "tbon.fanout", 0 }, // legacy, replaced by tbon.topo
 };
 
+static struct registered_attr *attrtab_lookup (const char *name)
+{
+    for (int i = 0; i < ARRAY_SIZE (attrtab); i++) {
+        if (fnmatch (attrtab[i].name, name, 0) == 0)
+            return &attrtab[i];
+    }
+    errno = ENOENT;
+    return NULL;
+}
+
 static void entry_destroy (void *arg)
 {
     struct entry *e = arg;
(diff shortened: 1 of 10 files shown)


garlick:
Restarted el9 builder after hitting
- #7350

t/t0001-basic.t
```diff
@@ -429,11 +429,12 @@ test_expect_success NO_ASAN 'test_under_flux fails if loaded modules are not unl
 '
 
 test_expect_success 'flux-start -o,--setattr ATTR=VAL can set broker attributes' '
-	ATTR_VAL=`flux start ${ARGS} -o,--setattr=foo-test=42 flux getattr foo-test` &&
+	ATTR_VAL=`flux start ${ARGS} -o,--setattr=test.foo=42 flux getattr test.foo` &&
 	test $ATTR_VAL -eq 42
+
```
grondo: unintentional extra newline?

src/broker/attr.c
```diff
@@ -19,6 +19,19 @@
 
 #include "attr.h"
 
+enum {
+    // ATTR_IMMUTABLE   = 0x01, // value never changes once set
```
grondo: Could add a reason why this flag is commented out?
e.g. "already defined in attr.h"

Would it be simpler to add these internal attrs to `attr.h`? They don't need to be exposed but also `attr.h` is internal to the broker so it doesn't seem crucial to keep them private and `ATTR_IMMUTABLE` is already exposed "publicly". This is just a comment, no need for a change here. Probably documenting that `ATTR_IMMUTABLE` is defined in `attr.h` would be sufficient.

src/broker/attr.c
```diff
@@ -218,6 +218,31 @@ int attr_add (attr_t *attrs, const char *name, const char *val, int flags)
...
+{
+    struct registered_attr *reg;
+
+
+    if (!attrs || !name) {
+        errno = EINVAL;
+        return errprintf (errp, "invalid argument");
+    }
+    if (!(reg = attrtab_lookup (name)))
+        return errprintf (errp, "unknown attribute");
```
grondo, line 225: Very minor: Commit message says the error will be "unknown attribute name" but it is just "unknown attribute"

src/broker/attr.c
```diff
@@ -81,12 +198,15 @@ int attr_delete (attr_t *attrs, const char *name, bool force)
...
 {
+    struct registered_attr *reg;
     struct entry *e;
 
     if (attrs == NULL || name == NULL) {
         errno = EINVAL;
         return -1;
     }
+    if (!(reg = attrtab_lookup (name)))
+        return -1;
```
grondo: The variable `reg` is unused in this function. Consider just
```c
if (!attrtab_lookup (name))
    return -1;
```
Same in `attr_add_active()`. However, it isn't doing any harm if you have reason to keep it.

src/broker/attr.c
```diff
@@ -19,6 +19,19 @@
 
 #include "attr.h"
 
+enum {
+    // ATTR_IMMUTABLE   = 0x01, // value never changes once set
+    ATTR_READONLY       = 0x02, // value is not to be set on cmdline by users
+    ATTR_RUNTIME        = 0x04, // value may be updated by users
+    ATTR_CONFIG         = 0x08, // value overrides TOML config
+                                //  (not currently used)
```
grondo: Should the `(not currently used)` comment apply to both these flags. They seem to be used in the table but not anywhere else currently. Is the idea to eventually have a tool to query the internal table of attributes along with their flags?

grondo (approved):
This LGTM!

I guess this means third party modules and framework projects are prohibited from using attributes for configuration like some flux-core modules do? At first I was worried about that change, but I think at this point we should be encouraging others to use broker config over attributes anyway. (Though the broker config suffers from the same issue being fixed here unfortunately)

Anyway, after some thought, this seems like a pretty good idea!

src/broker/attr.c
garlick: Actually in a future PR I was planning to make that flag internal to attr.c and add it to the appropriate table entries.  Then we wouldn't need a separate attr_add() vs attr_set() in the broker nor a way to change flags.  An immutable attribute would just become immutable after being set for the first time.

I'll add a comment though

Edit: actually you are right, it's easier to put all the flags in one place and move them all back once ATTR_IMMUTABLE isn't needed "publicly" anymore.

src/broker/attr.c
garlick: Sorry, correct that one currently isn't used either.
In another PR I was going to reject `attr.set` RPCs that have the ATTR_RUNTIME flag.  That used to be combined here but I ended up splitting up the (very long) branch into multiple PRs.

I wasn't really planning to expose the flags.  I was thinking if the attributes and their flags are known in advance, the "users" just could get by with  just the get/set functions and RPCs.

Will fix.

garlick:
> I guess this means third party modules and framework projects are prohibited from using attributes for configuration like some flux-core modules do? At first I was worried about that change, but I think at this point we should be encouraging others to use broker config over attributes anyway. (Though the broker config suffers from the same issue being fixed here unfortunately)

Well "discouraged" I guess is how I would put it rather than prohibited.  We could always add a way for a third party module to register a glob or list of attributes (with flags) that would be dynamically added to the table.  I'm just not sure that's needed or a good idea.

After mulling it over I thought it's probably better to just make make these primarily for the use of the broker itself and its built-in modules.  If a use case arises later on where it would be convenient to change that, we can take another look.

garlick:
Thanks!  I'll set MWP.

PR #7349, opened 2026-02-03
job-exec: improve coordination between sdexec and job-exec job termination timing

### LINKED ISSUE #7292: job's nodes were drained well before `max-kill-count` on elcap
@kkier noticed a job on elcap that resulted in many drained nodes well before the `max-kill-count` and `kill-timeout` values should have caused the job execution system to give up and drain the nodes (should have had about 15 minutes, and the nodes were drained after roughly 1m 30s.

Another reason ranks can be drained with unkillable processes is when sdexec returns `EDEADLOCK`:

https://github.com/flux-framework/flux-core/blob/8c14747ace3dd3851421894a6e59c68c42a56406/src/modules/job-exec/exec.c#L350-L362

This occurs when the stop of the systemd unit times out. However, Flux should be setting the `TimeoutStopUSec` limit to `infinity` in the job exec module:

https://github.com/flux-framework/flux-core/blob/8c14747ace3dd3851421894a6e59c68c42a56406/src/modules/job-exec/exec.c#L641-L666

However, spot checking a live job on a system, systemd appears to be ignoring this setting!

```
# systemctl show imp-shell-700-f2jPuay7DW7H.service | grep TimeoutStop
TimeoutStopUSec=1min 30s
```

Therefore, no matter the setting of `max-kill-count` and `kill-timeout`, jobs in the system instance of Flux are only being given 90s to exit before draining ranks with unkillable processes.

grondo:
This PR addresses the issues discussed in #7292 by ensuring the default `sdexec-stop-timer-sec` is greater than the effective maximum kill timeout for the job exec module.

A new function to return the current effective max-kill-timeout is added to the job-exec module. This value is now available in module stats as `effective-max-kill-timeout` and is valid whether the kill timeout is set via `max-kill-count` or `max-kill-timeout`.

The default value for `sdexec-stop-timer-sec` is now set to `effective-max-kill-timeout` rounded up to the nearest integer. Since sdexec waits `2 * sdexec-stop-timer-sec` before it abandons the unit, this is guaranteed to always be greater than the effective max-kill-timeout.

Finally, the `flux-config-exec(5)` man page is reworked and expanded with a lot more detail about these timeouts, a specific section for sdexec job termination, and more examples.

Fixes #7292

### DIFF
--- a/doc/man5/flux-config-exec.rst
+++ b/doc/man5/flux-config-exec.rst
@@ -6,6 +6,11 @@ flux-config-exec(5)
 DESCRIPTION
 ===========
 
+The exec system is highly configurable. If configuring a Flux system instance
+for the first time, it may be helpful to consult the Flux Administrator's
+Guide (see `RESOURCES`_) and start with a simple configuration. See also
+`EXAMPLES`_ below.
+
 The Flux system instance **job-exec** service requires additional
 configuration via the ``exec`` table, for example to enlist the services
 of a setuid helper to launch jobs as guests.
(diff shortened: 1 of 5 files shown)


garlick (approved):
This is great work and the extensive documentation is excellent.  Next time  there is a problem, hopefully there will be less head scratching!

My only thought is that someone casually looking up the exec config when setting up flux for the first time be a bit overwhelmed by all the detail in there.  Do you think it would be useful to add a little blurb at the very top like
> The exec system is highly configurable.  If configuring a Flux system instance for the first time it may be helpful to consult the Flux Administrators Guide (ref) and start with a simple configuration.  See also EXAMPLES (link).

(As I write this I am having second thoughts - other config(5) man pages jump right into the details without providing much background.  :shrug:  Well, I'll leave the comment and get your thoughts)

grondo:
I took that suggestion and pushed the result. Thanks! Will set MWP.

garlick:
Restarted the `bookworm - gcc12,distcheck` builder after hitting
- #7351

PR #7352, opened 2026-02-03
broker: minor cleanup of attribute usage

garlick:
Problem: there is some dead and ought-to-be-dead code in the broker attribute subsystem.

Get rid of the useless bits.

Clean up a couple of places where attributes are used in a silly way.

### DIFF
--- a/src/broker/log.c
+++ b/src/broker/log.c
@@ -568,14 +568,15 @@ void log_early (const char *buf, int len, void *arg)
     attr_t *attrs = arg;
     const char *val;
     int flags = 0;
-    uint32_t level;
+    int level;
     struct stdlog_header hdr;
 
     if (attr_get (attrs, "log-stderr-mode", &val, NULL) == 0
         && streq (val, "local")) {
         flags = LOG_FOR_SYSTEMD;
     }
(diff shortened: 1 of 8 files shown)


src/broker/log.c
```diff
@@ -568,14 +568,15 @@ void log_early (const char *buf, int len, void *arg)
...
+    int level;
     struct stdlog_header hdr;
 
     if (attr_get (attrs, "log-stderr-mode", &val, NULL) == 0
         && streq (val, "local")) {
         flags = LOG_FOR_SYSTEMD;
     }
-    if (attr_get_uint32 (attrs, "log-stderr-level", &level) == 0
+    if (attr_get (attrs, "log-stderr-level", &val, NULL) == 0
+        && set_level (&level, val) == 0
```
chu11, line 579: Perhaps `set_level()` should check if the input is < 0, so it's a bit more "equal" of a change.  B/c `set_level()` uses `strtol()` vs `strtoul()`.  Not sure what would be the consequence of an invalid < 0 log level snuck into this code.

garlick: A negative value means "match nothing"  - see 040e5815d7bdfc4839c78a819c7fd45d6695ee18

chu11 (approved):
LGTM!

PR #7356, opened 2026-02-04
broker: restrict runtime updates on some attributes

garlick:
Problem: [flux-broker-attributes(7)](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man7/flux-broker-attributes.html) notes that some attributes may be updated at run time and others not, but this is not enforced.

Use the ATTR_RUNTIME flag (previously added with the attribute table) to restrict updates via the `attr.set` RPC.  Allow the update to be forced if an optional `force` flag is set.  The force flag is need so "remote owners" of attributes such as the overlay module can make changes to non-immutable attributes despite the restriction.  The restriction therefore is advisory rather than mandatory.  The functions used inside the broker are unchanged.

A new API function `flux_attr_set_ex()` is added that includes the force flag, and also a `flux_error_t` so that human readable errors from the RPC can be retrieved by the caller.

Then a `flux setattr --force` option is added.

### DIFF
--- a/src/broker/attr.c
+++ b/src/broker/attr.c
@@ -125,6 +125,7 @@ static struct registered_attr attrtab[] = {
     // for testing
     { "test.*", ATTR_RUNTIME },
     { "test-ro.*", ATTR_READONLY },
+    { "test-nr.*", 0 },
 
     // misc undocumented
     { "vendor.*", ATTR_RUNTIME}, // flux-framework/flux-pmix#109
--- a/src/broker/attr.c
+++ b/src/broker/attr.c
@@ -418,13 +418,32 @@ void setattr_request_cb (flux_t *h,
     attr_t *attrs = arg;
     const char *name;
     const char *val;
+    int force = 0;
+    struct registered_attr *reg;
+    const char *errmsg = NULL;
 
     if (flux_request_unpack (msg,
                              NULL,
-                             "{s:s s:s}",
+                             "{s:s s:s s?b}",
                              "name", &name,
-                             "value", &val) < 0)
+                             "value", &val,
+                             "force", &force) < 0)
         goto error;
+    if (!(reg = attrtab_lookup (name))) {
+        errmsg = "unknown attribute";
+        goto error;
+    }
+    /* If name is found in the hash, then this is an update rather
+     * than the initial value.  Don't allow updates without the
+     * ATTR_RUNTIME flag, unless forced.
+     */
+    if (zhash_lookup (attrs->hash, name)) {
+        if (!force && !(reg->flags & ATTR_RUNTIME)) {
+            errmsg = "attribute may not be updated";
+            errno = EINVAL;
+            goto error;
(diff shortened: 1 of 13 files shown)


src/modules/overlay/overlay.c
```diff
@@ -2410,7 +2410,7 @@ static int overlay_configure_torpid (struct overlay *ov,
             return errprintf (errp,
                               "tbon.torpid_max must be greater than torpid_min");
         if (fsd_format_duration (fsd, sizeof (fsd), tmax_val) < 0
-            || flux_attr_set (ov->h, "tbon.torpid_max", fsd) < 0)
+            || flux_attr_set_ex (ov->h, "tbon.torpid_max", fsd, true, NULL) < 0)
             return errprintf (errp, "error updating tbon.torpid_min attribute");
```
grondo: This error message should be updated to reference `tbon.torpid_max`

src/cmd/builtin/attr.c
```diff
@@ -34,13 +35,20 @@ static int cmd_setattr (optparse_t *p, int ac, char *av[])
...
+    if (flux_attr_set_ex (h, name, val, force, &error) < 0)
         log_msg_exit ("%s: %s", name, error.text);
 
     flux_close (h);
     return (0);
 }
 
+static struct optparse_option setattr_opts[] = {
+    { .name = "force", .key = 'f', .has_arg = 0,
+      .usage = "force update even though runtime updates are unsupported",
```
grondo: Take it or leave it comment:

minor wording suggestion: "even if runtime updates are not supported" instead of "even though"
or perhaps most clear: "force update of attribute that doesn't support runtime changes"

src/broker/attr.c
```diff
@@ -418,13 +418,28 @@ void setattr_request_cb (flux_t *h,
...
+        errmsg = "unknown attribute";
         goto error;
+    }
+    if (zhash_lookup (attrs->hash, name)) {
+        if (!force && !(reg->flags & ATTR_RUNTIME)) {
+            errmsg = "attribute may not be updated";
+            errno = EINVAL;
+            goto error;
+        }
+    }
```
grondo, line 447: It feels like a comment should go here describing why the attr flags are not applied if the attribute has not already been registered. I assume this is so that non-runtime attributes registered by glob in the attr table can be set the first time?

grondo (approved):
LGTM! Just a couple comments inline.

src/broker/attr.c
garlick: Correct.  I'll add a comment
```c
    /* If name is found in the hash, then this is an update rather
     * than the initial value.  Don't allow updates without the
     * ATTR_RUNTIME flag, unless forced.
     */
```

garlick:
Thanks!  Fixed those things and setting MWP on this one

garlick:
Restarting fedora40 after hitting
- #6552

PR #7358, opened 2026-02-04
broker: simplify internal attribute interface

garlick:
Now that we have a table of known broker attributes, there is no need to externally manipulate the ATTR_IMMUTABLE flag.  It can simply be added to the static table entries where appropriate.  That lets the internal attribute API be simplified:
- drop `attr_set_flags()` and `attr_add()`
- drop `flags` parameter from `attr_get()` and `attr_add_active()`


garlick:
Oops forgot this needs to be based on top of
- #7356 

I'll push push that in a sec.

grondo:
I've gotten myself confused on the difference between an immutable attribute and one without ATTR_RUNTIME set, i.e. what is the difference between a flags of `0` vs `ATTR_IMMUTABLE`?

src/broker/attr.c
```diff
@@ -188,21 +196,22 @@ int attr_delete (attr_t *attrs, const char *name)
...
     }
-    if (!attrtab_lookup (name))
+    if (!(reg = attrtab_lookup (name)))
         return -1;
     if ((e = zhash_lookup (attrs->hash, name))) {
         errno = EEXIST;
         return -1;
     }
-    if (!(e = entry_create (name, val, flags)))
+    if (!(e = entry_create (name, val, reg->flags & ATTR_IMMUTABLE)))
```
grondo: Sorry if this should be obvious, but why is only `ATTR_IMMUTABLE` preserved here?

grondo (approved):
LGTM! Though it might be useful to put a comment somewhere that details the difference between flags=0 vs flags=ATTR_IMMUTABLE and why this flag is the only one preserved in entry_create().

garlick:
> LGTM! Though it might be useful to put a comment somewhere that details the difference between flags=0 vs flags=ATTR_IMMUTABLE and why this flag is the only one preserved in entry_create().

Yeah, sorry, the flag name choices were perhaps not great.  Attributes with ATTR_IMMUTABLE may be cached in clients.  For example, `rank` never changes once it is set, so there is no need to make an RPC to the broker to fetch it ever again once it is in your cache.  Immutable attributes, once set cannot be updated even with the force flag.

Attributes with no flags MAY be updated with the force flag so that they can be changed by their owners.  Since there is no check for who might be forcing an update, this is more of an advisory limitation than a mandatory one.

The ATTR_IMMUTABLE flag is the only flag that is exposed externally so that clients can conditionally cache attribute values.  The other flags could be stored in the entry flags and then filtered out in the getattr response.  In retrospect, maybe that makes more sense - I'll see if I can sneak that in here.

Will add a comment block explaining all this.

grondo:
Thanks! That was very helpful!

garlick:
OK, those changes added and rebased on current master.  Setting MWP.  Thank you!

garlick:
derp.  Introduced a spelling error - re-pushed.

garlick:
Restarting bookworm 32 bit after hitting
- #7376

and fedora40 after
- #7377

PR #7362, opened 2026-02-05
shell: add `shell.finish` callback

grondo:
This PR adds a new `shell.finish` callback to the job shell, invoked after all local tasks complete but before the reactor exits.

This enables plugins to perform cleanup or initiate asynchronous work while the reactor is still active, without waiting for `shell.exit` which runs after reactor shutdown.

No plugins make use of this callback yet, that will come in a couple follow-on PRs.


garlick (approved):
LGTM and nice doc improvement.

grondo:
Thanks I'll set MWP.

Issue #7345, opened 2026-02-02
Potential inconsistency in tool shutdown grace period


dirkschubert-linaro:
Raising as follow up to https://github.com/flux-framework/flux-core/issues/7269. There seems to be an inconsistency when `SIGTERM` is delivered to tool processes. See the following behaviour.

**Agent on node 2**

```
...
Install signal handler for SIGTERM
2026-02-02T14:00:30,045314591+00:00
...
2026-02-02T14:00:34,721638481+00:00
Signal: SIGTERM
2026-02-02T14:00:34,803646095+00:00
2026-02-02T14:00:39,830063566+00:00
```

🟢  Behaves as expected - the agent is running for 5s, then SIGTERM is delivered which is processed for 5s before terminating.

**Agent on node 1**

```
2026-02-02T14:00:29,995031841+00:00
...
2026-02-02T14:00:39,654921536+00:00
Signal: SIGTERM
2026-02-02T14:00:39,878460243+00:00
2026-02-02T14:00:44,905126849+00:00
```

🟠 Does not behave as expected - the agent is running for 10s before SIGTERM is delivered which is processed for 5s before terminating.

dirkschubert-linaro:
**Reproducer Script**

Uses recent `master`:

```
commands:    	0.81.0-294-gf9209c285
libflux-core:	0.81.0-294-gf9209c285
broker:  		0.81.0-294-gf9209c285
build-options:	+hwloc.api==2.12.0+zmq==4.3.5
```

Request a 2 node Slurm allocation: `srun -N 2 --pty flux start`.

Run `reproducer.sh` which has the following contents:

```shell
#!/usr/bin/env bash

# ================
# Write Tool Agent
# ================

cat > mpir-agent.sh << 'EOF'
#!/usr/bin/env bash
# Print arguments:
log_file=mpir-agent-`hostname`-$$.txt
touch $log_file
echo "Arguments: $@" >> $log_file
# Handle common signals:
signal_handler()
{
    echo "Signal: $1" >> $log_file
    date --iso-8601=ns >> $log_file
    sleep 5
    date --iso-8601=ns >> $log_file
    exit
}
for s in SIGHUP SIGQUIT SIGINT SIGTERM; do
    echo "Install signal handler for $s" >> $log_file
    trap "signal_handler $s" $s
done
# Run until interrupted, printing current time every so often:
while true; do
    date --iso-8601=ns >> $log_file
    sleep 0.25
done
EOF
[...]

PR #7363, opened 2026-02-05
shell: send EOF to write service immediately after tasks finish

grondo:
Job shell write service clients send EOF to the leader only during plugin destruction at shell exit. If a follower shell's `shell.exit` callback delays (e.g. performing cleanup), the leader must wait for all followers to complete before it can exit the reactor and invoke its own `shell.exit` callbacks. This can cause strange cascading timeout behavior as seen in #7345.

This PR destroys the output client from the `shell.finish` callback instead of at shell process exit. This sends the client EOF immediately after tasks exit, allowing the leader to proceed without waiting for follower shells' exit callbacks.

This should resolve issue #7345 on its own, however a follow on PR will also shut down the subprocess server in the `shell.finish` callback as well for good measure.

This is based on top of #7362

### DIFF
--- a/src/shell/output/client.c
+++ b/src/shell/output/client.c
@@ -49,10 +49,10 @@ struct output_client {
 
 static void client_send_eof (struct output_client *client)
 {
-    /* Note: client should not be instantiated on rank 0, but check here
-     * just in case.
+    /* Note: client should not be instantiated on rank 0, and may be NULL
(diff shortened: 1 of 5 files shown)


garlick (approved):
LGTM!

grondo:
Thanks! Setting MWP.

PR #7366, opened 2026-02-05
rc1: do not run fsck in recovery mode

### LINKED ISSUE #7353: `flux fsck` may run unnecessarily under `flux start --recovery`
This is more of a question at this point.

We noticed when starting a recovery instance on Tuolumne that `flux fsck` was running out of rc1, even though there was technically a clean shutdown beforehand.

I'm guessing this is because there was a checkpoint, but the latest dump is not restored with `flux start --recovery`.

Should `flux fsck` be suppressed for a recovery instance?

chu11:
Problem: It was discovered that fsck is run when flux is started in recovery mode.  Since we are in recovery mode, the fsck really
shouldn't be run.  It would be up to an administrator to run it manually if desired.

Disable fsck when flux starts in recovery mode.

Fixes #7353

### DIFF
--- a/etc/modprobe/rc1.py
+++ b/etc/modprobe/rc1.py
@@ -141,6 +141,8 @@ def check_clean_shutdown(context):
 #
 # If a dump was restored, we can assume the prior instance was shutdown
 # cleanly.
+#
+# Special case: no fsck in recovery mode
 @task(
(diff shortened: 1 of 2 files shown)


garlick (approved):
LGTM!

Issue #7370, opened 2026-02-09
dmesg: add --stderr-level option


garlick:
Problem: the broker includes the `attr_add_active()` interface for registering callbacks on attributes when get/set are called on an attribute, but this only works for broker-resident services, which are dwindling.

Only the log subsystem uses this interface, and of all its attributes, probably only `log-stderr-level`, which is adjusted up and down during some tests, actually needs to remain runtime adjustable.  The rest can just be set on the broker command line as needed.

Consider replacing:
```
flux setattr log-stderr-level N
```
with:
```
flux dmesg --stderr-level=N
```
This is kind of like the  `--console-level=level` option in [dmesg(1)](https://www.man7.org/linux/man-pages/man1/dmesg.1.html).

Then we could get rid of that part of the attribute interface as an internal broker cleanup.

garlick:
Well this was addressed another way in #7389, so closing.

PR #7373, opened 2026-02-09
flux-dmesg: add --stderr-level option

garlick:
Problem: #7370 proposes to deprecate the current method for changing the stderr log severity level at runtime, but some flux-core tests are still adjusting the level up and down for specific test segments.

Add a new option `flux-dmesg --stderr-level=N` and corresponding `log.setattr` RPC in the log subsystem.

p.s. this only adds the new way.  A future PR can deprecate the old way.


chu11 (approved):
LGTM

garlick:
Superseded by #7389 

(Thanks for the review @chu11 - sorry I guess I should have thought this through more before proposing it)

Issue #7377, opened 2026-02-10
not ok 3 - flux-cron tab works for set minute


garlick:
Problem: this failure was originally observed in fedora40 CI:
```
2026-02-10T20:10:17.4751086Z load cron module
2026-02-10T20:10:17.4751242Z 
2026-02-10T20:10:17.4751421Z expecting success: 
2026-02-10T20:10:17.4752055Z     $set_faketime today 15:30 &&
2026-02-10T20:10:17.4752592Z     id=$(echo "15 * * * * flux event pub t.cron.complete" | flux_cron tab) &&
2026-02-10T20:10:17.4753162Z     next=$(date +%s --date="today 16:15") &&
2026-02-10T20:10:17.4753622Z     cron_entry_check ${id} type datetime &&
2026-02-10T20:10:17.4753952Z     cron_entry_check ${id} stopped false &&
2026-02-10T20:10:17.4754256Z     cron_entry_check ${id} typedata.next_wakeup ${next} &&
2026-02-10T20:10:17.4754563Z     echo sleeping at $(date) &&
2026-02-10T20:10:17.4754812Z     ${event_trace} t.cron t.cron.complete \
2026-02-10T20:10:17.4755077Z         $set_faketime today 16:15 &&
2026-02-10T20:10:17.4755314Z     echo done at $(date) &&
2026-02-10T20:10:17.4755545Z     cron_entry_check ${id} stats.count 1 &&
2026-02-10T20:10:17.4756015Z     flux cron delete ${id}
2026-02-10T20:10:17.4756165Z 
2026-02-10T20:10:17.4756346Z timestamp file: @2026-02-10 15:30:00: date is now Tue Feb 10 15:30:00 UTC 2026
2026-02-10T20:10:17.4756720Z cron-1: type=datetime, wanted datetime
2026-02-10T20:10:17.4756977Z cron-1: stopped=false, wanted false
2026-02-10T20:10:17.4757276Z cron-1: typedata.next_wakeup=1770740100, wanted 1770740100
2026-02-10T20:10:17.4757727Z sleeping at Tue Feb 10 15:30:00 UTC 2026
2026-02-10T20:10:17.4758076Z libfaketime: In parse_ft_string(), failed to parse FAKETIME timestamp.
2026-02-10T20:10:17.4758461Z Please check specification  with format %Y-%m-%d %T
2026-02-10T20:10:17.4758784Z not ok 3 - flux-cron tab works for set minute
```

[...]

PR #7386, opened 2026-02-13
testsuite: make faketime timestamp file updates atomic

grondo:
Problem: When `make-faketime.sh` writes to the libfaketime timestamp file in t0016-cron-faketime.t, it may be possible that some Flux or other process calls gettimeofday() or other time call in the middle of the write. This could cause unexpected failures in the test.

Write faketime timestamp to a temporary file first, then atomically rename it to the target file to avoid a partial read or other similar race condition.

This _might_ address #7377, but I'm not too hopeful. We can merge this since it won't hurt anything and see if we still get similar failures in the faketime tests going forward.

### DIFF
--- a/t/t0016-cron-faketime.t
+++ b/t/t0016-cron-faketime.t
@@ -60,8 +60,10 @@ flux_cron() {
 cat >make-faketime.sh <<EOF
 #!/bin/sh
 d="\$@"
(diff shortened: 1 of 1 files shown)


cmoussa1 (approved):
LGTM!

grondo:
Thanks! I'll set MWP.

PR #7387, opened 2026-02-13
drain nodes with "terminated before first barrier" exception (multiuser jobs only)

### LINKED ISSUE #7383: drain nodes with "terminated before first barrier" exception
We recently had a munge issue that led to many jobs failing in shell startup on the same nodes with `NODE[xxx,xxx] (ranks xxx,xxx) terminated before first barrier` exception messages. Could Flux drain those nodes automatically?

grondo:
This PR drains ranks that exit before the first barrier if the job was a multiuser job.

This preserves existing behavior for single user instances (job exception only), but adds the requested behavior in #7383 for multi user system instances, where it more likely a node issue has caused the early failure of the IMP or job shell.

Fixes #7383

### DIFF
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -54,6 +54,7 @@ LONGTESTSCRIPTS = \
 	t3202-instance-restart-testexec.t \
 	t3203-instance-recovery.t \
 	t4000-issues-test-driver.t \
+	t2800-jobs-cmd.t \
 	t2801-top-cmd.t \
 	t2808-shutdown-cmd.t \
 	t2712-python-cli-alloc.t \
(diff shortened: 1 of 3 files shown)


t/Makefile.am
```diff
@@ -54,6 +54,7 @@ LONGTESTSCRIPTS = \
 	t3202-instance-restart-testexec.t \
```
garlick, line 54: Commit message: lower case t in LONGTEStSCRIPTS

garlick (approved):
LGTM!  Just noted one typo in a commit message.

t/Makefile.am
grondo: Fixed that and will set MWP. Thanks!

PR #7388, opened 2026-02-13
t: add test with empty kvs txn

chu11:
Problem: There is no coverage for ensuring an empty KVS txn works.

Add tests to t1001-kvs-internals.t.

Also a misc cleanup fix that I had laying around without a PR to stick it in :-)

### DIFF
--- a/src/common/libsubprocess/test/stdio.c
+++ b/src/common/libsubprocess/test/stdio.c
@@ -1445,7 +1445,7 @@ void credit_output_cb (flux_subprocess_t *p, const char *stream)
         sprintf (cmpbuf, "abcdefghijklmnopqrstuvwxyz0123456789");
         ok (streq (outputbuf, cmpbuf),
             "flux_subprocess_read returned correct data: %s", outputbuf);
-        /* 26 (ABCs) + 10 (1-10) */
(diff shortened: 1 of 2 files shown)


garlick:
Would `flux kvs sync` work here?  It does this:
```c
        log_err_exit ("flux_open");
    if (!(txn = flux_kvs_txn_create ()))
        log_err_exit ("flux_kvs_txn_create");
    if (!(f = flux_kvs_commit (h, NULL, FLUX_KVS_SYNC, txn))
        || flux_future_get (f, NULL) < 0)
        log_err_exit ("flux_kvs_commit");
    flux_future_destroy (f);
    flux_kvs_txn_destroy (txn);
    flux_close (h);
```
so basically the same thing as the new C program but with the FLUX_KVS_SYNC option.

chu11:
> so basically the same thing as the new C program but with the FLUX_KVS_SYNC option.

OMG, that's a great idea.  I didn't think of that.

We do `flux kvs sync` in a number of places, so effectively it's sort of already covered.  But perhaps it's still worthwhile to have an explicit test for it?  I think so atleast.

chu11:
re-pushed, going w/ `flux kvs sync` in the new test

garlick (approved):
LGTM!

PR #7389, opened 2026-02-17
broker:  strengthen support for runtime attribute updates

garlick:
This replaces the callback based "active attribute" interface with an RPC redirect table so that update handlers need not run in the same thread.  Then it changes the log subsystem to use the new interface.

Rather than retain the ability to dynamically update _all_ log attributes, drop runtime update support from all but `log-stderr-level` as this seems to be the only one that is usefully updated on the fly (mainly in test).  The others can be customized on the broker command line.

This PR is proposed in lieu of
- #7370 

that one doesn't solve the callback problem in an general way, but instead just adds a purpose built RPC for updating `log-stderr-level` and changes the  user interface from `flux setattr` to `flux dmesg`.

### DIFF
--- a/src/broker/attr.c
+++ b/src/broker/attr.c
@@ -386,10 +386,10 @@ int attr_cache_immutables (attr_t *attrs, flux_t *h)
  ** Service
  **/
 
-void getattr_request_cb (flux_t *h,
-                         flux_msg_handler_t *mh,
-                         const flux_msg_t *msg,
-                         void *arg)
+static void getattr_request_cb (flux_t *h,
+                                flux_msg_handler_t *mh,
+                                const flux_msg_t *msg,
+                                void *arg)
 {
     attr_t *attrs = arg;
     const char *name;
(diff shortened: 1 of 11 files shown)


garlick:
Just realized I can do something a little bit smarter here - setting WIP temporarily

garlick:
OK, now 2% less unsmart.  Removing WIP

src/broker/attr.c
```diff
@@ -161,12 +167,22 @@ static struct registered_attr attrtab[] = {
...
 
     // misc undocumented
     { "vendor.*", ATTR_RUNTIME | ATTR_IMMUTABLE }, // flux-pmix#109
     { "tbon.fanout", 0 }, // legacy, replaced by tbon.topo
 };
 
+/* Setattr on attributes matching the redirect table are handled elsewhere.
+ * The attr.set RPC handler here makes an RPC to the new topic string.
+ * When that completes a response is sent to the original request, and if
+ * the updates is approved, the hash value is updated.
```
grondo: typo: `if the updates is approved` ➡️ `if the update is approved`

grondo (approved):
This looks great to me on a quick review! Just spotted one typo.

garlick:
Thanks, fixed that and I'll set MWP.

PR #7391, opened 2026-02-18
t: fix potential job-list reload race

### LINKED ISSUE #7322: not ok 198 - verify task count preserved across restart
Problem: the following test in `t2260-job-list.t` was observed to fail in the jammy-testinstall CI builder:

```
2026-01-23T19:18:48.0954801Z expecting success: 
2026-01-23T19:18:48.0955043Z 	jobid=`cat duration.id` &&
2026-01-23T19:18:48.0955298Z 	flux job list -s inactive | grep ${jobid} > duration2.json &&
2026-01-23T19:18:48.0955470Z 	jq -e ".duration == 3600.0" < duration2.json
2026-01-23T19:18:48.0955527Z 
2026-01-23T19:18:48.0955758Z not ok 198 - verify task count preserved across restart
```

(That test title is repeated 5 times within the sharness test - this is the 5th one, in the "duration time" section)

chu11:
Problem: A number of tests in t2260-job-list.t ensure data is loaded correctly after the job-list module is reloaded.  There is a potential race where the reloaded module is not yet ready by the time a test is executed.  This is because the module is executed in a thread, and we can't control the execution of that thread.

Add a function to ensure the job list module is up and running after a reload in t2260-job-list.t.

Fixes #7322

### DIFF
--- a/t/t2260-job-list.t
+++ b/t/t2260-job-list.t
@@ -1261,6 +1261,24 @@ test_expect_success 'flux job list job state timing outputs valid (job running)'
 	fj_wait_event $jobid clean >/dev/null
 '
 
+# `flux module reload job-list` does not ensure `job-list` is actually
+# ready to service requests immediately afterwards (i.e. has list
+# service been setup yet).  This checks to make sure `job-list` is
+# really up and ready.
+wait_job_list_ready() {
+	local i=0
+	while ( ! flux job list > /dev/null && [ $i -lt 50 ] )
+	do
+		sleep 0.1
+		i=$((i + 1))
+	done
+	if [ "$i" -eq "50" ]
+	then
+		return 1
+	fi
+	return 0
+}
+
 #
 # job names
 #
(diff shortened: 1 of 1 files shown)


t/t2260-job-list.t
```diff
@@ -1261,6 +1261,24 @@ test_expect_success 'flux job list job state timing outputs valid (job running)'
...
+		sleep 0.1
+		i=$((i + 1))
+	done
+	if [ "$i" -eq "50" ]
+	then
+		return 1
+	fi
+	return 0
+}
+
```
grondo, line 1281: This implementation might be simpler:
```sh
wait_job_list_ready() {
    local i=0
    while [ $i -lt 50 ]; do
        flux job list >/dev/null  && return 0
        sleep 0.1
        i=$((i + 1))
    done
    return 1
}
```

- avoids the subshell
- returns immediately from successful `flux job list`
- suppresses the output of `flux job list`, but keeps errors for debug purposes
- removes the requirement to test `$i` twice

grondo (approved):
LGTM! One minor suggestion inline.

t/t2260-job-list.t
chu11: the "pattern" I used is in a lot of tests.  I'll create an issue to clean all those up.

PR #7392, opened 2026-02-18
testsuite: fix several flaky tests

grondo:
Fix four flaky tests observed in a particular VM environment:

- `t0032-resource-journal.py`: handle resource-define event arriving
  before online event
- `t2814-hostlist-cmd.t`: add retry loop to cover race between overlay
  and resource module marking a node down
- `t0100-modprobe.t`: filter unordered modules before comparing removal
  order
- `t0005-exec.t`: skip signal forwarding test when `stdbuf` is
  non-conformant (older `uutils/coreutils` versions of coreutils)

### DIFF
--- a/t/python/t0032-resource-journal.py
+++ b/t/python/t0032-resource-journal.py
@@ -31,12 +31,15 @@ class TestResourceJournalConsumer(unittest.TestCase):
 
         events = []
         while True:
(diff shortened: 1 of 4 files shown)


garlick (approved):
LGTM!

PR #7393, opened 2026-02-18
fix assorted bugs in broker and job-manager

grondo:
As an experiment I used Claude Code (Opus 4.6) to look for issues in `broker.c` as well as `job-manger.c` and `job-manager/plugins/*`. This set of fixes is the result of that work.

A few bugs were due to the strange way some error messages are allocated on return by some internal jobtap functions, so this PR cleans that up to use the more standard `flux_error_t` interface.

### DIFF
--- a/src/modules/job-manager/start.c
+++ b/src/modules/job-manager/start.c
@@ -370,7 +370,7 @@ static void expiration_update_cb (flux_future_t *f, void *arg)
                                     note) < 0)
             flux_log_error (ctx->h, "expiration_update: raise_job_exception");
     }
-    job_aux_delete (job, "job-manager::R-update");
+    job_aux_set (job, "job-manager::R-update", NULL, NULL);
 }
 
 /* Send <exec_service>.expiration request to adjust job expiration
@@ -398,7 +398,7 @@ int start_send_expiration_update (struct start *start,
         || flux_future_then (f, -1., expiration_update_cb, job) < 0
         || flux_future_aux_set (f, "job-manager::ctx", ctx, NULL) < 0) {
         int saved_errno = errno;
-        (void) job_aux_delete (job, "job-manager::R-update");
+        (void) job_aux_set (job, "job-manager::R-update", NULL, NULL);
         flux_future_destroy (f);
         errno = saved_errno;
         return -1;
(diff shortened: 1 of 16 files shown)


src/modules/job-manager/start.c
```diff
@@ -370,7 +370,7 @@ static void expiration_update_cb (flux_future_t *f, void *arg)
                                     note) < 0)
             flux_log_error (ctx->h, "expiration_update: raise_job_exception");
     }
-    job_aux_delete (job, "job-manager::R-update");
+    job_aux_set (job, "job-manager::R-update", NULL, NULL);
```
garlick, line 373: Oy, that one's a foot gun!  Maybe we should rename `aux_delete()` (and the functions that may wrap it) to `aux_delete_byvalue()` or something.

grondo: Yeah, I was thinking the same thing. There's a lot of uses of that outside of flux-core though I assume. Maybe a new function aux_delete_key() that wraps `job_aux_set (job, key, NULL, NULL)` would help ease any confusion.

garlick: I couldn't find a single example of its use outside of flux-core so perhaps we could consider being a bit more heavy handed, maybe leaving the jobtap aux_delete function there with a deprecation attribute?

garlick (approved):
LGTM!

This is a nice basket of potentially high value (due to leader broker crash risk) fixes!

Great job applying the new tools here.

src/modules/job-manager/start.c
garlick: Anyway, no need to do that here.  I'll open an issue.

grondo: Ah! I had assumed wrapping `aux_delete()` was more widespread, but looks like it is only used in jobtap `job_aux_delete()` and the plugin interface `flux_plugin_aux_delete()` (we can check further on that one). I like your idea of adding a new interface and deprecating the old one. (and the same for `flux_plugin_aux_delete()`)

grondo:
Fixed  a CodeQL-detected issue and force pushed the result. We're also seeing a failure in the flux-accounting check in CI, so I have to run that down.

grondo:
The flux-accounting issue is apparently intermittent. Setting MWP here now. Thanks!

PR #7399, opened 2026-02-19
t: cleanup waiting functions

### LINKED ISSUE #7396: testsuite: cleanup "wait for command" helper functions
In https://github.com/flux-framework/flux-core/pull/7391

@grondo  notes that code like

```
	local i=0
	while ( ! flux job list && [ $i -lt 50 ] )
	do
		sleep 0.1
		i=$((i + 1))
	done
	if [ "$i" -eq "50" ]
	then
		return 1
	fi
	return 0
```

could be cleaned up like

```
    local i=0
    while [ $i -lt 50 ]; do
        flux job list >/dev/null  && return 0
        sleep 0.1
        i=$((i + 1))
    done
    return 1
```

This pattern is used in a lot of places.  Try to cleanup globally.

chu11:
Problem: In a number of circumstances tests need to wait for some asynchronous activity to complete before continuing tests.  Functions to wait on something are often written in this pattern:

```
wait_foo() {
	local i=0
	while ( ! flux foo && [ $i -lt 50 ] )
	do
		sleep 0.1
		i=$((i + 1))
	done
	if [ "$i" -eq "50" ]
	then
		return 1
	fi
	return 0
}
```

This can be a bit confusing, especially when "flux foo" is long, has output piped to other commands, and/or is a compound test of multiple commands.

Solution:  Refactor to the following style:

```
wait_foo() {
	local i=0
	while [ $i -lt 50 ]
	do
		flux foo && return 0
		sleep 0.1
		i=$((i + 1))
	done
	return 1
}
```

or if the test is long or a compound test:

```
wait_foo() {
	local i=0
	while [ $i -lt 50 ]
	do
		if flux foo | jq -e ".foo.bar.baz == 87"
		then
			return 0
		fi
		sleep 0.1
		i=$((i + 1))
	done
	return 1
}
```

this makes the function easier to understand and is collectively fewer lines overall.

Fixes #7396

### DIFF
--- a/t/kvs/kvs-helper.sh
+++ b/t/kvs/kvs-helper.sh
@@ -12,7 +12,7 @@ loophandlereturn() {
     index=$1
     if [ "$index" -eq "${KVS_WAIT_ITERS}" ]
     then
-        return 1
+	return 1
     fi
     return 0
 }
@@ -36,14 +36,14 @@ test_kvs_key_namespace() {
 
 # arg1 - namespace
 wait_watcherscount_nonzero() {
-        ns=$1
-        i=0
-        while (! flux module stats --parse namespaces.${ns}.watchers kvs-watch > /dev/null 2>&1 \
-               || [ "$(flux module stats --parse namespaces.${ns}.watchers kvs-watch 2> /dev/null)" = "0" ]) \
-              && [ $i -lt ${KVS_WAIT_ITERS} ]
-        do
-                sleep 0.1
-                i=$((i + 1))
-        done
-        return $(loophandlereturn $i)
+	ns=$1
+	i=0
+	while (! flux module stats --parse namespaces.${ns}.watchers kvs-watch > /dev/null 2>&1 \
+	       || [ "$(flux module stats --parse namespaces.${ns}.watchers kvs-watch 2> /dev/null)" = "0" ]) \
+	      && [ $i -lt ${KVS_WAIT_ITERS} ]
+	do
+		sleep 0.1
+		i=$((i + 1))
+	done
+	return $(loophandlereturn $i)
 }
--- a/t/kvs/kvs-helper.sh
+++ b/t/kvs/kvs-helper.sh
@@ -3,20 +3,6 @@
 
 # KVS helper functions
 
-# Loop on KVS_WAIT_ITERS is just to make sure we don't spin forever on
-# error.
-
-KVS_WAIT_ITERS=50
-
-loophandlereturn() {
-    index=$1
-    if [ "$index" -eq "${KVS_WAIT_ITERS}" ]
-    then
-	return 1
-    fi
-    return 0
-}
-
 # arg1 - key to retrieve
 # arg2 - expected value
 test_kvs_key() {
@@ -37,13 +23,6 @@ test_kvs_key_namespace() {
 # arg1 - namespace
 wait_watcherscount_nonzero() {
 	ns=$1
-	i=0
-	while (! flux module stats --parse namespaces.${ns}.watchers kvs-watch > /dev/null 2>&1 \
-	       || [ "$(flux module stats --parse namespaces.${ns}.watchers kvs-watch 2> /dev/null)" = "0" ]) \
-	      && [ $i -lt ${KVS_WAIT_ITERS} ]
-	do
-		sleep 0.1
-		i=$((i + 1))
-	done
-	return $(loophandlereturn $i)
+	test_wait_until "flux module stats --parse namespaces.${ns}.watchers kvs-watch \
+		&& [ \$(flux module stats --parse namespaces.${ns}.watchers kvs-watch 2> /dev/null) -ne 0 ]"
 }
--- a/t/sharness.d/flux-sharness.sh
+++ b/t/sharness.d/flux-sharness.sh
@@ -339,6 +339,93 @@ test_flux_security_version() {
     test $major -ge $req_major -a $minor -ge $req_minor -a $patch -ge $req_patch
 }
 
+# Default number of iterations for test_wait_until() (can be
+# overridden)
+TEST_WAIT_UNTIL_DEFAULT_ITERS=50
+
+# Default sleep time in seconds between iterations for
+# test_wait_until() (can be overridden)
+TEST_WAIT_UNTIL_DEFAULT_SLEEP=0.1
+
+# Default verbosity for test_wait_until() (can be overridden)
+TEST_WAIT_UNTIL_DEFAULT_VERBOSE=0
+
+#
+# Common utility function to wait for a condition to be met
+# The condition is provided as a command to execute
+#
+# Usage: test_wait_until [OPTIONS] "COMMAND [ARGS...]"
+#
+# Options:
+#   -i, --iterations N   Maximum iterations to wait (default: $TEST_WAIT_UNTIL_DEFAULT_ITERS)
+#   -s, --sleep S        Sleep time between iterations (default: $TEST_WAIT_UNTIL_DEFAULT_SLEEP)
+#   -v, --verbose        Print diagnostic messages during waiting
+#
+# Example usage:
+#   test_wait_until "flux kvs get mykey" # Wait until command succeeds
+#   test_wait_until "test -f myfile"     # Wait until file exists
+#   test_wait_until "grep pattern file"  # Wait until pattern appears in file
+#   test_wait_until -i 100 -s 0.5 "custom_check_command arg1 arg2"
+#
+test_wait_until() {
+    local iterations=$TEST_WAIT_UNTIL_DEFAULT_ITERS
+    local sleep_time=$TEST_WAIT_UNTIL_DEFAULT_SLEEP
+    local verbose=$TEST_WAIT_UNTIL_DEFAULT_VERBOSE
+
+    # Process options
+    while [ $# -gt 0 ]; do
+        case "$1" in
+            -i|--iterations)
+                iterations=$2
+                shift 2
+                ;;
+            -s|--sleep)
+                sleep_time=$2
+                shift 2
+                ;;
+            -v|--verbose)
+                verbose=1
+                shift
+                ;;
+            *)
+                break
+                ;;
+        esac
+    done
+
+    # Ensure we have a command to run
+    if [ $# -eq 0 ]; then
+        echo "Error: test_wait_until requires a command to execute" >&2
+        return 1
+    fi
+
+    # The remaining arguments form the command to execute
+    local cmd="$1"
+    [ $verbose -eq 1 ] && echo "test_wait_until: running $cmd"
+
+    local i=0
+    while [ $i -lt $iterations ]
+    do
+        if [ $verbose -eq 1 ]; then
+            if eval "$cmd"; then
+                echo "test_wait_until: condition met after $i iterations"
+                return 0
+            fi
+        else
+            if eval "$cmd" >/dev/null 2>&1; then
+                return 0
+            fi
+        fi
+
+        [ $verbose -eq 1 ] && echo "test_wait_until: waiting... ($i/$iterations)"
+        sleep $sleep_time
+        i=$((i + 1))
+    done
+
+    [ $verbose -eq 1 ] && echo "test_wait_until: timeout after $iterations iterations"
+    return 1
+}
+
 #  Export a shorter name for this test
 TEST_NAME=$SHARNESS_TEST_NAME
 export TEST_NAME
--- a/t/system/0007-fsck.t
+++ b/t/system/0007-fsck.t
@@ -12,18 +12,7 @@ test_expect_success 'restart flux' '
 '
 
 wait_flux_back_up() {
-	i=0
-	while ! flux resource list > /dev/null 2>&1 \
-	      && [ $i -lt 100 ]
-	do
-		sleep 1
-		i=$((i + 1))
-	done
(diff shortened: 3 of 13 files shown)


chu11:
image

Huh ... a double tab indentation is not showing up correctly on the diff output in github ("return 0").  I don't know why.  Is this just me?

garlick:
Could the repetition be avoided by adding a helper function like
```
# Usage: test_run_until retries interval command ...
test_run_until() {
    local retries=$1; shift
    local interval=$1; shift
    while ! "$@"; do
        test $retries -eq 0 && return 1
        sleep $interval
        retries=$(($retries-1))
    done
}
```

grondo:
I was going to make that same suggestion!

chu11:
lol, it's so obvious in hindsight.  Just got settled on a goal and that was it :-)

chu11:
re-pushed.  Using claude code as an experiment, asked for a "common function" for all of these waiting functions, and it produced something pretty nice w/ only a minor tweak or two from me.  The actual "cleanup" was not as smooth, a lot of hand editing to cleanup the cleanup and Claude 100% screwed up in a few places and I had to debug to fix.

Side note, I kept my original work in this PR stream.  Claude had problems finding all of the places to cleanup, so I had to manually input them to say "hey specifically clean these up".  So the original audit cleanup in this PR served an important purpose.  So I elected to leave in that original work.  But I'm not married to this ... it could all be squashed away.

t/util/wait-util.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/sh
+#
+# Common wait utility function for Flux test suite
+#
+
+# Default number of iterations (can be overridden)
+WAIT_UTIL_DEFAULT_ITERS=50
+
+# Default sleep time in seconds between iterations (can be overridden)
+WAIT_UTIL_DEFAULT_SLEEP=0.1
```
grondo: This function should probably just go into `t/sharness.d/flux-sharness.sh with the other Flux convenience functions. This way tests don't have to separately include it, and when we update flux-sharness.sh in other testsuites, they'll have access to the function.

I also wonder if allowing the defaults to be adjusted via global variables here is necessary, but I guess it could come in handy. If we keep these, I'd suggest renaming to `TEST_WAIT_UNTIL_...` and maybe even rename the function `test_wait_until()` but that's just a minor suggestion to fit with the naming convention of sharness "test" functions.

t/util/wait-util.sh
```diff
@@ -0,0 +1,82 @@
...
+
+#
+# Common utility function to wait for a condition to be met
+# The condition is provided as a command to execute
+#
+# Usage: wait_util [OPTIONS] COMMAND [ARGS...]
```
grondo: I kept reading this as `wait_until` not `wait_util`.  I think this might trip people up, and would suggest a rename to `wait_until` or `test_wait_until` as suggested above.

t/util/wait-util.sh
grondo: Actually, I also wonder if `TEST_WAIT_UNTIL_VERBOSE=t` is the _most_ useful thing that could be set in the environment, to help debug a failing test using this function. 🤷

grondo (commented):
Just added a couple comments about Claude's `wait_util` implementation.

grondo:
Also, if you go with the current approach, I'd definitely squash the intermediate work.

chu11:
re-pushed, fixing per comments above.  Unfortunately had to do by hand ... Claude was dumb about doing the changes (I suspect if I had listed every file and every function, it probably could have done it)

t/sharness.d/flux-sharness.sh
```diff
@@ -339,6 +339,88 @@ test_flux_security_version() {
...
+# Default number of iterations for test_wait_until() (can be
+# overridden)
+TEST_WAIT_UNTIL_DEFAULT_ITERS=50
+
+# Default sleep time in seconds between iterations for
+# test_wait_until() (can be overridden)
+TEST_WAIT_UNTIL_DEFAULT_SLEEP=0.1
+
+# Default verbosity for test_wait_until() (can be overridden)
+TEST_WAIT_UNTIL_DEFAULT_VERBOSE=0.1
```
grondo: Cut and paste error in the value here?

chu11: doh!

chu11:
re-pushed fixing stupid cut & paste error

t/kvs/kvs-helper.sh
```diff
@@ -37,13 +23,6 @@ test_kvs_key_namespace() {
...
-	while (! flux module stats --parse namespaces.${ns}.watchers kvs-watch > /dev/null 2>&1 \
-	       || [ "$(flux module stats --parse namespaces.${ns}.watchers kvs-watch 2> /dev/null)" = "0" ]) \
-	      && [ $i -lt ${KVS_WAIT_ITERS} ]
-	do
-		sleep 0.1
-		i=$((i + 1))
-	done
-	return $(loophandlereturn $i)
+	test_wait_until "flux module stats --parse namespaces.${ns}.watchers kvs-watch > /dev/null 2>&1 \
+		&& [ \"\$(flux module stats --parse namespaces.${ns}.watchers kvs-watch 2> /dev/null)\" != \"0\" ]"
```
grondo: Sorry I didn't notice this before, but `test_wait_until` appears to already redirect stdout and stderr from `eval "$cmd"` to `/dev/null`, so the redirected stdio in the command here is redundant.

```sh
        # Use eval to properly handle command with arguments
        if eval "$cmd" > /dev/null 2>&1; then
            [ $verbose -eq 1 ] && echo "test_wait_until: condition met after $i iterations"
            return 0
        fi
```

(this goes for all the updates in this commit)

t/system/0007-fsck.t
```diff
@@ -12,18 +12,7 @@ test_expect_success 'restart flux' '
...
-	do
-		sleep 1
-		i=$((i + 1))
-	done
-	if [ "$i" -eq "100" ]
-	then
-		return 1
-	fi
-	return 0
+	test_wait_until -i 100 -s 1 "flux resource list > /dev/null 2>&1"
```
grondo: `test_wait_until` sets `$cmd` to `"$@"`, so once the redundant `>/dev/null 2>&1` is removed, you can also drop the quoting for simple commands like this one, though that's optional of course.

t/t2800-jobs-cmd.t
```diff
@@ -1227,18 +1227,7 @@ test_expect_success 'update project and bank in pending jobs' '
...
-	do
-		sleep 0.1
-		i=$((i + 1))
-	done
-	if [ "$i" -eq "50" ]
-	then
-		return 1
-	fi
-	return 0
+	test_wait_until "[ \"\$(flux jobs -f pending -o {project} | grep foo | wc -l)\" = \"\$(job_list_state_count sched)\" ]"
```
grondo: Minor comment:
Readability could be improved here by a couple things:
- replace `grep foo | wc -l` with `grep -c foo`
- use `-eq` instead of `=` (since you're comparing numbers) and drop the need for escaped double quotes
```s
test_wait_until "[ \$(flux jobs -f pending -o {project} | grep -c foo) -eq \$(job_list_state_count sched) ]"
```

t/t2809-job-purge.t
```diff
@@ -29,15 +29,12 @@ wait_inactive_count() {
...
-		test $count -eq $target && return 0
-		sleep 0.25
-		tries=$(($tries-1))
-	done
-	return 1
+
+	test_wait_until -i $tries -s 0.25 "
+		count=\$(inactive_count $how)
+		echo \$count inactive jobs >&2
+		test \$count -eq $target
```
grondo, line 35: As noted above I think `test_wait_until` always redirects stdout and stderr of cmd to `/dev/null`, so I don't think the `echo` is doing anything here. Could simplify this test to one line.

t/t2260-job-list.t
```diff
@@ -2656,18 +2602,7 @@ test_expect_success 'list-attrs works' '
...
-	do
-		sleep 0.1
-		i=$((i + 1))
-	done
-	if [ "$i" -eq "1000" ]
-	then
-		return 1
-	fi
-	return 0
+	test_wait_until -i 1000 "[ \"\$(flux job list | wc -l)\" = \"0\" ]"
```
grondo: Could simplify using `-eq` and drop escaped double quotes.

t/t2260-job-list.t
```diff
@@ -1059,19 +1048,8 @@ test_expect_success 'flux job list-ids fails with one bad ID out of several' '
...
-		sleep 0.1
-		i=$((i + 1))
-	done
-	if [ "$i" -eq "50" ]
-	then
-		return 1
-	fi
-	return 0
+	test_wait_until "flux module stats --parse idsync.waits job-list > /dev/null 2>&1 \
+		&& [ \"\$(flux module stats --parse idsync.waits job-list 2> /dev/null)\" = \"$num\" ]"
```
grondo: Another spot where `-eq` could be used to compare numbers, removing the need for double quotes.
I didn't catch all of these I don't think.

grondo (commented):
Just took another pass and found a couple things I didn't catch before, sorry!
Probably at least the redundant stdio redirects (`>/dev/ull 2>&1`) should be cleaned up before a merge.

t/kvs/kvs-helper.sh
chu11: Good catch on the `/dev/null` redirect.  Thinking about this more I'm thinking one of two options:

1) remove the `/dev/null` redirect in `test_wait_until()`.   In the inevitable "need to debug", we'd like output to occur so we can debug.  So the test writer can control that by choosing to redirect or not.

OR

2) Add the redirect only when verbose is false.

The latter option is sort of the "makes sense" option given there is a verbose option.  But with a number of compound test statements already used, it may be preferable for the caller to control what gets output vs not.

t/t2800-jobs-cmd.t
chu11: I noticed this in 1 or 2 places, but intentionally left them.  I will add a cleanup commit to clean up a few of these.

t/kvs/kvs-helper.sh
grondo: I vote for option 2 as well. It kind of appeared like all the existing tests suppressed stdout and stderr anyway, and keeping the redirects at the call sites require quoting even the simplest of commands when it shouldn't be necessary, and make the complex commands overly long.

chu11:
re-pushed doing fixups / cleanups per comments above.  The primary change is to not redirect stdio to /dev/null in verbose mode in `test_wait_until()`.

garlick:
Seems like it should be possible to drop the recurring double quotes on the expression passed to `test_wait_until`, e.g. instead of 
```
test_wait_until "test \$(command) -eq 1"
```
just do
```
test_wait_until test \$(command) -eq 1
```
?

I thought that would be the case since internally it does the equivalent of `eval "$@"`

chu11:
> I thought that would be the case since internally it does the equivalent of eval "$@"

Perhaps this is a style thing.  For the longer tests, especially with pipes and compound statements, I like to use the quotes to indicate "hey, this all goes together".  Thus I left double quotes in all the cases.

I'm not super married to this though.  Strong preference to remove em?

garlick:
Not especially strong feeling, although quoting can quickly get of hand in sharness tests so I'm not sure extra quotes are a great general practice.

grondo:
> Perhaps this is a style thing. For the longer tests, especially with pipes and compound statements, I like to use the quotes to indicate "hey, this all goes together". Thus I left double quotes in all the cases.

I brought this up earlier. For a simple command we should drop the quotes. If cmd has a pipe or `&&` or `||` then you do need to quote otherwise the outer shell will break up your command.

If you want to always quote the command for simplicity, then it seems like `cmd` in `test_wait_until` should just be set to `$1` because there would only ever be one argument anyway.

garlick:
For consistency with test_must_fail and friends, probably the current behavior is best IMHO.

chu11:
re-pushed, removed the quotes for simple commands when calling `test_wait_until()`

garlick (approved):
This is great cleanup.  Well, I guess there will always be a few more things to fix but since this touches quite a lot we may want to get it in before it becomes a rebase problem.  I'll approve.  We should get a nod from @grondo to make sure all his concerns have been addressed.

grondo (approved):
LGTM! Thanks!!

My only remaining comment is you might want to squash the incremental work (last 4 commits could be one)

chu11:
> My only remaining comment is you might want to squash the incremental work (last 4 commits could be one)

I kept them separate since each individual cleanpu was sort of unique, but I'll go ahead and squash, re-push, and I"ll set MWP.

Thanks!

chu11:
i don't know why, but focal / py 3.8 failed with a ton of errors.  All of the tests that failed had `test_wait_until` with unquoted commands to check (i.e. `test_wait_until flux job list` vs `test_wait_until "flux job list"`).  

I have no idea why, this should be super safe / portable within dash.  Dunno if there is a subtlety I'm missing.

chu11:
without any other clues, I asked codex and it says there is a portability issue with older versions of dash.   Codex suggested a fix where it would run `eval "$@"` vs just `"$@"` depending on the number of args.  But I thought that was excessive.

So i went with a suggestion from @grondo above.  Everything is passed to `test_wait_until` as a single quoted command.  Inside test_wait_until we do `local cmd=$1" instead.

PR #7402, opened 2026-02-20
add `FLUX_HWLOC_XMLFILE_NOT_THISSYSTEM`, fix cpu-affinity=dry-run with non-local topology

grondo:
When loading an hwloc XML file from a different system via `FLUX_HWLOC_XMLFILE`, Flux always applies `HWLOC_TOPOLOGY_FLAG_IS_THISSYSTEM`, which can cause unexpected behavior such as missing resources. This makes it difficult to use external XML topology files for end-to-end testing, particularly with the shell `cpu-affinity=dry-run` option. See flux-framework/flux-sched#1408.

While `resource.norestrict=true` happens to work around the issue for initial resource discovery, this solution is neither obvious nor convenient. It also doesn't fix the shell affinity plugin.

To make the solution obvious and simple, a new environment variable `FLUX_HWLOC_XMLFILE_NOT_THISSYSTEM` is added that disables `HWLOC_TOPOLOGY_FLAG_IS_THISSYSTEM` and skips `hwloc_topology_restrict()`, allowing topology files from other systems to be used for testing. The shell cpu-affinity plugin is also updated to omit this flag when `cpu-affinity=dry-run` is set.

A new test `t2604-job-shell-affinity-xmlfile.t` validates end-to-end use of a non-local XML topology with the affinity plugin in dry-run mode.

### DIFF
--- a/doc/man7/flux-environment.rst
+++ b/doc/man7/flux-environment.rst
@@ -391,9 +391,28 @@ components or writing tests.
    parallel testing.  Flux may be directed to read topology from an XML file
    instead by setting :envvar:`FLUX_HWLOC_XMLFILE` to the file path.
 
+   .. note::
+      When using this option, Flux assumes the topology in the XML file is
+      from the current system. If this is not the case, for example when
+      loading a topology from another system for testing purposes, set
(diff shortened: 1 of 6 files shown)


garlick (approved):
LGTM!

grondo:
Thanks! I'll set MWP.

PR #7403, opened 2026-02-20
cmd/flux: output similar commands on invalid input

chu11:
Problem: When a user inputs an invalid command to the flux command (e.g. "flux kkvs") the flux command outputs an appropriate message indicating the command is invalid.  It'd be nice if the flux command could output similar commands, in case the user typoed something or isn't quite sure what the command is actually called.

Solution: Using Levenshtein distance, calculate what commands are most similar to the invalid command specified by the user.  If one is pretty similar, output a message informing the user of that command.

example:

```
>src/cmd/flux fvs
flux: `fvs' is not a flux command.  See 'flux --help'
flux: The most similar command is `kvs`
```

Notes:

This was a simple issue I wrote up awhile ago and thought it might be a good experiment to see if it could be solved with AI.

I started with the prompt "i would like the flux command to output a message suggesting a similar command when an invalid command is entered".

The result was "meh".  It had enough quirks that I ultimately ended up doing it myself, with 1 or 2 ideas brought over.  But did have AI pull in a levenshtein distance implementation (string comparison algorithm) in its own commit.

### DIFF
--- a/src/cmd/flux.c
+++ b/src/cmd/flux.c
@@ -28,6 +28,8 @@
 #include "src/common/libutil/xzmalloc.h"
 #include "src/common/libutil/environment.h"
 #include "src/common/libutil/intree.h"
+#include "src/common/libutil/levenshtein.h"
+#include "src/common/libutil/dirwalk.h"
 
 #include "cmdhelp.h"
 #include "builtin.h"
@@ -417,6 +419,109 @@ void exec_subcommand_dir (bool vopt,
     free (path);
 }
 
+struct similar_cmds {
+    int min;
+    zlist_t *cmds;
+    int cmdsstrlen;
+    const char *argv0;
+};
+
+static void similar_check (struct similar_cmds *similar,
+                           const char *cmd)
+{
+    int distance = levenshtein_distance (cmd, similar->argv0);
+    if (distance > 0 && distance <= similar->min) {
+        char *cpy = xstrdup (cmd);
+        if (distance < similar->min) {
+            zlist_purge (similar->cmds);
+            similar->cmdsstrlen = 0;
+        }
+        similar->min = distance;
+        zlist_append (similar->cmds, cpy);
+        zlist_freefn (similar->cmds, cpy, free, true);
+        similar->cmdsstrlen += strlen (cpy);
+    }
+}
+
+static int similar_filter (dirwalk_t *d, void *arg)
+{
+    struct similar_cmds *similar = arg;
+    const char *path = dirwalk_path (d);
+    const char *name = dirwalk_name (d);
+    char *cmdname;
+
+    if (!strstarts (name, "flux-"))
+        return 0;
+
+    if (access (path, X_OK) < 0)
+        return 0;
+
+    /* +5 to get past "flux-" */
+    cmdname = xstrdup (name + 5);
+
+    /* strip off ".py" if necessary */
+    if (strends (cmdname, ".py")) {
+        char *tmp = strrchr (cmdname, '.');
+        *tmp = '\0';
+    }
+
+    similar_check (similar, cmdname);
+    free (cmdname);
+    return 0;
+}
+
+static void find_similar_command (const char *searchpath, const char *argv0)
+{
+    extern struct builtin_cmd builtin_cmds[];
+    struct builtin_cmd *builtin_cmd = &builtin_cmds[0];
+    struct similar_cmds similar = {INT_MAX, NULL, 0, argv0};
+    char *searchpathcpy;
+    char *dir, *saveptr = NULL, *a1;
+
+    similar.min = INT_MAX;
+    if (!(similar.cmds = zlist_new ()))
+        log_err_exit ("failed to create cmds list");
+    similar.argv0 = argv0;
+
+    /* first check out builtins */
+    while (builtin_cmd->name) {
+        similar_check (&similar, builtin_cmd->name);
+        builtin_cmd++;
+    }
+
+    /* now check commands in search paths */
+    searchpathcpy = xstrdup (searchpath);
+    a1 = searchpathcpy;
+    while ((dir = strtok_r (a1, ":", &saveptr))) {
+        (void)dirwalk (dir, DIRWALK_NORECURSE, similar_filter, &similar);
+        a1 = NULL;
+    }
+    free (searchpathcpy);
+
+    /* only output if command is similar enough, we'll go with a
+     * distance of at most 3.
+     *
+     * e.g. "resourcccce" will be similar to "resource", but not
+     * "resourccccce".
+     */
+    if (similar.min <= 3) {
+        char *str;
+        int i = 3;
+
+        log_msg ("The most similar command%s",
+                 zlist_size (similar.cmds) > 1 ? "s are" : " is");
+
+        /* output at most 3 commands */
+        str = zlist_first (similar.cmds);
+        while (str && i-- > 0) {
+            log_msg ("\t%s", str);
+            str = zlist_next (similar.cmds);
+        }
+    }
+
+    zlist_destroy (&similar.cmds);
+}
+
 void exec_subcommand (const char *searchpath, bool vopt, int argc, char *argv[])
 {
     if (strchr (argv[0], '/')) {
@@ -435,8 +540,9 @@ void exec_subcommand (const char *searchpath, bool vopt, int argc, char *argv[])
             a1 = NULL;
         }
         free (cpy);
-        log_msg_exit ("`%s' is not a flux command.  See 'flux --help'",
-                      argv[0]);
+        log_msg ("`%s' is not a flux command.  See 'flux --help'", argv[0]);
+        find_similar_command (searchpath, argv[0]);
+        exit (1);
     }
 }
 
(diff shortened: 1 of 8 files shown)


garlick:
This is new territory for me, but should the LLM model and version at least be documented in the code with a statement that it's generated code, so there is _some_ link back to the training corpus in case we ever get out of wild west mode on this stuff?

Edit: here is some engaging reading on the subject :-)
https://copyrightsociety.org/wp-content/uploads/2025/04/713_The-Heart-of-the-Matter.pdf

p.508 mentions this about the EU AI act
> The Act provides for a disclosure obligation,
where providers of foundations models have to draw up and make publicly
available a sufficiently detailed summary of the content used for training of the
foundation model, according to a template provided by the AI Office.145 Thus, the
Act provides, in principle, for a mechanism that would enable right holders to
ascertain whether their works have been used for training specific AI models

garlick:
Oh I see ~~you~~ Claude added
```
Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
```
:+1:

garlick:
I haven't looked in detail yet but how is this constructing the list of candidate subcommands?

We may want to be less inclusive, for example

Apparently build products in the source tree are matched?
```
$ flux bro
flux: `bro' is not a flux command.  See 'flux --help'
flux: The most similar command is `R.o`
```

Some commands are not going to be helpful for most users so perhaps we ought to have a blocklist?
```
$ flux run-pro
flux: `run-pro' is not a flux command.  See 'flux --help'
flux: The most similar command is `run-prolog`
```

chu11:
> I haven't looked in detail yet but how is this constructing the list of candidate subcommands?

Right now it searches through builtins, then it searches all the search paths (i.e. from FLUX_EXEC_PATH) looking for `flux-*` commands (`.py` files have `.py` stripped of course).  Each command gets checked for "distance" against the user's invalid command.  The minimum one at the end is the one that is output.

I was admittedly lazy with the approach, so that no "list of candidates" is built up.  The command with the lowest distance is simply output, meaning that the first command it finds that is the minimum is the one that is output over others that are "tied" for the lowest distance.  I felt this was "good enough" for the majority of cases.

> Apparently build products in the source tree are matched?

It'll search all the normal paths that the `flux` command would normally search.  So assuming this is a devel branch, then it'll find commands from `src/cmd/`, etc.

> Some commands are not going to be helpful for most users so perhaps we ought to have a blocklist?

So it's funny you mention this.  My first implementation of this was to only use the commands listed in `etc/flux/help.d/core.json`.  However that had the problem of A) it wouldn't pickup commands from user paths added via `FLUX_EXEC_PATH` and B) the `core.json` file is a pretty minimal set of commands.  So I felt searching for every possible candidate command was a good idea?   But perhaps a few blocklist commands might be reasonable.

As an aside, I elected to only output "similar" commands with a distance of <= 3.  We could tighten it even more.  i.e. if we used a distance of just 1, then it would output similar commands where the user typoed at most 1 character.

src/cmd/flux.c
```diff
@@ -417,6 +418,76 @@ void exec_subcommand_dir (bool vopt,
...
+                    free (min_cmd);
+                    min_cmd = xstrdup (name);
+                }
+                free (pathcpy);
+            }
+            globfree (&gl);
+        }
+        free (pattern);
+        a1 = NULL;
+    }
```
grondo, line 498: I think this could all be replaced with `dirwalk_find()` from libutil. The filter function should check for the executable bit to prevent matching build products (like `flux-R.o` in @garlick's example)

chu11:
re-pushed, going with `dirwalk()` for implementation, and checking to make sure the file is executable (that was a big miss from before).

src/common/libutil/levenshtein.c
```diff
@@ -0,0 +1,92 @@
...
+        min = b;
+    if (c < min)
+        min = c;
+    return min;
+}
+
+int levenshtein_distance (const char *s1, const char *s2)
+{
+    int *matrix;
+    int s1len, s2len;
```
garlick: sizes should probably be a size_t which would fix the 
"Multiplication result may overflow 'int' before it is converted to 'unsigned long'." warning from the github security bot.

src/cmd/flux.c
```diff
@@ -417,6 +419,88 @@ void exec_subcommand_dir (bool vopt,
...
+
+static int similar_filter (dirwalk_t *d, void *arg)
+{
+    struct similar_cmd *similar = arg;
+    const char *path = dirwalk_path (d);
+    const char *name = dirwalk_name (d);
+    char *cmdname;
+    int distance, len;
+
+    if (strncmp (name, "flux-", 5) != 0)
```
garlick: use `strstarts()` from ccan

src/cmd/flux.c
```diff
@@ -417,6 +419,88 @@ void exec_subcommand_dir (bool vopt,
...
+    cmdname = xstrdup (name + 5);
+    len = strlen (cmdname);
+
+    if (len > 3) {
+        /* strip off ".py" if necessary */
+        if (strcmp (cmdname + (len - 3), ".py") == 0) {
+            char *tmp = strrchr (cmdname, '.');
+            *tmp = '\0';
+        }
+    }
```
garlick: use  `strends()` from ccan or just unconditionally truncate at the last period.

src/cmd/flux.c
```diff
@@ -417,6 +419,88 @@ void exec_subcommand_dir (bool vopt,
...
+
+    distance = levenshtein_distance (cmdname, similar->argv0);
+    if (distance > 0 && distance < similar->min) {
+        similar->min = distance;
+        free (similar->cmd);
+        similar->cmd = xstrdup (cmdname);
+    }
+    free (cmdname);
+    return 0;
```
garlick: Why make another copy of `cmdname`?  Just assign and free `cmdname` in the else lag of conditional.

src/cmd/flux.c
```diff
@@ -417,6 +419,88 @@ void exec_subcommand_dir (bool vopt,
...
+    free (searchpathcpy);
+
+    /* only output if command is similar enough, we'll go with a
+     * distance of at most 3.
+     *
+     * e.g. "resourcccce" will be similar to "resource", but not
+     * "resourccccce".
+     */
+    if (similar.min <= 3)
+        log_msg ("The most similar command is `%s`", similar.cmd);
```
garlick: Would it make sense to collect matches in a zlist then print all that are under 3 or 4?

t/t0001-basic.t
```diff
@@ -845,6 +845,21 @@ test_expect_success 'command --help works outside of flux instance' '
...
+	grep "similar" invalid2.out | grep "module" &&
+	test_must_fail flux resourcce > invalid3.out 2>&1 &&
+	grep "similar" invalid3.out | grep "resource"
+'
+
+test_expect_success 'really bad typoed commands are not given suggestions' '
+	test_must_fail flux resouurrccee > invalidbad.out 2>&1 &&
+	test_must_fail grep "similar" invalidbad.out
+'
+
```
garlick, line 880: It might be good to add a check for some matches that are explicitly avoided like files with extensions.

garlick (approved):
LGTM, I'll approve but I had some optional suggestions inline.

src/cmd/flux.c
chu11: I thought of truncating at the last period.  But I actually tested this b/c I was unsure.

```
>cat ./flux-dumb.sh
#!/bin/sh
echo "dumb dumb"
>FLUX_EXEC_PATH=. src/cmd/flux dumb
flux: `dumb' is not a flux command.  See 'flux --help'
>FLUX_EXEC_PATH=. src/cmd/flux dumb.sh
dumb dumb
```

So we actually honor the periods in the flux command names.  So I did decide to keep that behavior and not truncate at the last period.

Maybe we don't want to??

src/cmd/flux.c
chu11: I did consider it, but the odds of matching > 1 is pretty low.   

Lemme give it a try, see if I end up liking the result.

chu11:
re-pushed, now up to 3 similar commands will be output.

```
>src/cmd/flux mom
flux: `mom' is not a flux command.  See 'flux --help'
flux: The most similar commands are: top, job
```

src/cmd/flux.c
```diff
@@ -417,6 +419,120 @@ void exec_subcommand_dir (bool vopt,
...
+     * "resourccccce".
+     */
+    if (similar.min <= 3) {
+        char *cmdstr = similar_cmds_string (&similar);
+        if (zlist_size (similar.cmds) == 1)
+            log_msg ("The most similar command is: %s", cmdstr);
+        else
+            log_msg ("The most similar commands are: %s", cmdstr);
+        free (cmdstr);
+    }
```
garlick, line 520: Suggestion, rather than build a string, just print the individual matches.

It might be nice to do like git and print one per line also
```
$ git bit
git: 'bit' is not a git command. See 'git --help'.

The most similar commands are
	bisect
	init
```

chu11:
re-pushed, going w/ the commands on multiple lines.  Still up to 3 max.

```
>src/cmd/flux mop
flux: `mop' is not a flux command.  See 'flux --help'
flux: The most similar command is
flux: 	top

>src/cmd/flux mom
flux: `mom' is not a flux command.  See 'flux --help'
flux: The most similar commands are
flux: 	top
flux: 	job
```

PR #7404, opened 2026-02-23
ensure starttime and expiration appear in resource set returned from `flux resource R`

grondo:
This PR fixes a minor issue where the current starttime and expiration are not present in the output of `flux resource R`.

The problem exists in 2 places:
- the status cache is not invalidated on a `resource-update` event, so a stale expiration could be present in the resource set returned by the `resource.sched-status` RPC.
- The `flux resource R` implementation creates an empty resource set object and doesn't update the expiration and starttime before adding the requested resources to the set.

This PR fixes those two issues. The Python `ResourceSet` class unit tests are also expanded since there was some missing coverage.

### DIFF
--- a/src/bindings/python/flux/resource/ResourceSet.py
+++ b/src/bindings/python/flux/resource/ResourceSet.py
@@ -274,3 +274,19 @@ class ResourceSet:
     @property
     def properties(self):
         return ",".join(json.loads(self.get_properties()).keys())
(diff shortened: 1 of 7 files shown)


garlick (approved):
LGTM!

grondo:
Thanks! I'll set MWP.

Issue #7248, opened 2025-12-18
Can't get job remaining time (e.g. via libyogrt) when running Flux as a job under Slurm


grondo:
When Flux is started as a job under Slurm, all `SLURM_*` environment variables are sanitized to avoid confusing some MPI implementations (#5206, fixed in 480c01b2d90b330bb5a4bba37dbc7e91fed6bb3f). However, this means that  libyogrt does not detect that the top-level job is Slurm and can't provide the Slurm remaining time. Additionally, `flux job timeleft` and associated API call do not provide time remaining either.

The workaround proposed for now is to save `SLURM_JOB_ID` and `SLURM_PROCID` before launching Flux, then restore them in the initial program. I'm not sure if even these variables would be enough to confuse MPI though, so we'd have to test that or make it conditional.

In #5206 I had mentioned "I think we have proposed other mechanisms to support this [obtaining time limit of Flux under Slurm]", but I don't remember what any proposals were. Perhaps inheriting a time limit from a foreign RM automatically?

garlick:
See also: llnl/libyogrt#19

grondo:
Currently the SLURM environment variables are unset in the broker before the initial program runs. Since the `SLURM_*` variables mostly interfere with MPI implementations, would one approach here be to instead remove `SLURM_*` environment variables in the job shell? There's probably something obvious I'm forgetting.

Edit: Well I guess if users are calling libyogrt from within a Flux job, then this doesn't really solve anything (and that's probably the common case)

garlick:
What about a two part fix
- in `libyogrt`, choose dynamically based on environment rather than config file which resource manager to query (if it doesn't do that already - I didn't check)
- add resource module plugins and build one that knows how to ask slurm for time left, so `expiration` can be set correctly in _R_

grondo:
This is a good direction. I did look into it already and `libyogrt` doesn't support dynamically determining which resource manager to query currently, but it wouldn't be too difficult to add IMO.

Currently `libyogrt` can't get remaining time for a Flux instance that isn't also a Flux job, however, possibly that could be added.

Alternately, the `SLURM` environment variables could be allowed to be in the initial program. Since `FLUX_JOB_ID` isn't set in the initial program environment, `libyogrt` would fall through to query Slurm if called in the initial program. `SLURM` environment variables would be filtered from jobs, though, and `FLUX_JOB_ID` _would_ be set there, so jobs would use the Flux `libyogrt` implementation.

[...]

PR #7405, opened 2026-02-24
track job time time limit when Flux is run as a Slurm job

grondo:
This is a proposed partial solution for #7248.

It is based on top of #7404

In short, this PR adds a new helper utility,  `flux slurm-expiration-sync`, which is started from rc1 when it appears Flux is running under Slurm. The script queries Slurm for the job's time limit at startup, and if there is one it sends the expiration to the resource module which posts a `resource-update` event. This propagates the expiration to the scheduler. The script continues to poll the Slurm time limit in case it is updated while the job is active.

To get there, some other required changes included:
- addition of a `resource.expiration-update` RPC which allows the helper script to notify the `resource` module of expiration updates
- A new `needs_env` parameter for the modprobe `@task` decorator. This allows a modprobe task to be selectively scheduled based on presence (or absence) of an environment variable. Since it may be surprising that the broker environment is filtered in rc1/rc3, `needs_env` first checks the local environment and falls back to the broker environment if the variable is not found. (This allows it to be used for `SLURM_JOB_ID` in the implementation)
- A `flux.slurm` helper Python module is added, mainly to enable easy unit testing of the Slurm squeue time limit output and syntax as well as a new `flux.slurm.slurm_timeleft()` function.
- A mock `squeue` in `t/scripts` for testing purposes.

real world demo:
```console
$ salloc -t 60 -N1 -p pdebug
salloc: Pending job allocation 5100573
salloc: job 5100573 queued and waiting for resources
salloc: job 5100573 has been allocated resources
$ srun --pty src/cmd/flux start
$ flux run flux job timeleft -H
58.3895m
```

in another window adjust slurm job timelimit `scontrol update jobid=JOBID timelimit=30m`, wait ~1m (or watch resource eventlog for resource-update event)
```console
$ flux run flux job timeleft -H
25.3813m
```

This should be enough to allow libyogrt to work for _jobs_ run in the Flux instance, but querying the remaining time will still fail if queried from the initial program environment, because `flux_job_timeleft(3)` does not work when the instance is not a Flux job. (This could be fixed by querying resource.R if we want to do that, but it could be in a separate PR)

### DIFF
--- a/etc/modprobe/rc1.py
+++ b/etc/modprobe/rc1.py
@@ -10,11 +10,14 @@
 
 import errno
 import os
+import sys
 from pathlib import Path
 from subprocess import Popen
 
 import flux.kvs
+import flux.subprocess as sp
 from flux.modprobe import run_all_rc_scripts, task
+from flux.resource import ResourceJournalConsumer
 
 
 def setup(context):
@@ -202,6 +205,44 @@ def push_cleanup(context):
             ],
         },
     ).get()
+
+
+@task(
+    "slurm-expiration-sync",
+    ranks="0",
+    needs_env=["SLURM_JOB_ID"],
+    before=["sched"],
+    after=["resource"],
+)
+def slurm_expiration_sync(context):
+    # Check for valid SLURM_JOB_ID before proceeding:
+    try:
+        jobid = int(context.getenv("SLURM_JOB_ID"))
+    except ValueError:
+        print("invalid SLURM_JOB_ID. Skipping expiration sync", file=sys.stderr)
+        return
+
+    # Force libyogrt backend to flux
+    context.setenv("YOGRT_BACKEND", "flux")
+
+    consumer = ResourceJournalConsumer(context.handle).start()
+    while True:
+        event = consumer.poll(timeout=5.0)
+        if event.name == "resource-define":
+            # Once resources are defined launch Slurm time limit sync helper:
+            sp.rexec_bg(
+                context.handle,
+                ["flux", "slurm-expiration-sync", f"--jobid={jobid}"],
+                label="slurm-expiration-sync",
+            )
+        elif event.name == "resource-update":
+            # The resource-update event is posted by the resource module after
+            # flux-slurm-expiration-sync updates the expiration in R. Breaking
+            # here ensures the scheduler does not start until the expiration is
+            # up to date. Note: This task could exit early if there is another
+            # mechanism for updating the resource set in play. This case is not
+            # handled at this time.
+            break
 
 
 @task("run-all-rc1", after=["*"])
--- a/src/bindings/python/flux/Makefile.am
+++ b/src/bindings/python/flux/Makefile.am
@@ -68,6 +68,7 @@ nobase_fluxpy_PYTHON = \
 	queue.py \
 	modprobe.py \
 	subprocess.py \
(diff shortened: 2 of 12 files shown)


grondo:
One of the new tests is flaky. Will post here when I determine the cause and fix it.

grondo:
Ok, I guess the flaky test is resolved, though I'm not sure what I did (I may have previously pushed up an older version of the branch). 

Additional fixes include:
- add SIGTERM handler to exit `flux-slurm-expiration-sync.py` gracefully. This allows the script to exit without error when the rexec module subprocess server sends SIGTERM during shutdown (no rc3 task necessary)
- add missing tests for the `resource.expiration-update` RPC
- prevent `resource.expiration-update` RPC on ranks != 0

garlick (approved):
This LGTM!

I assume this has been tested on a real slurm system with libyogrt.  I didn't see that mentioned in the discussion?

It was a long road to the goal but this PR gets there in a methodical, well-documented, and well-tested manner.  Thanks!

grondo:
No, there's still more to be done for full libyogrt support. Either
1. we need to extend libyogrt to support dynamic detection of its backend
2. Set `YOGRT_BACKEND=flux` for Flux jobs. This could be done on a per system basis via `initrc.lua`, or we could always set this environment variable.

There is also still the case of calling `yogrt_remaining()` from the initial program environment, which would not work now because `flux_job_timeleft(3)` doesn't work when the current instance was not launched as a Flux job.

grondo:
I think I'm going to look at fixing #7407. Then propose a PR that adds the modprobe environment enhancements (`getenv` from here and a new `broker_setenv()` call), then redo this PR on top of that, but now setting `YOGRT_BACKEND=flux` when Flux running under Slurm is detected.

grondo:
This is now based on top of #7409, and `YOGRT_BACKEND=flux` is exported to the initial program environment if the `slurm-expiration-sync` helper process is started.

So, now this just works for Flux jobs run within an instance started as a Slurm job:
```console
$ squeue -j $SLURM_JOB_ID -o %L
TIME_LEFT
8:18
$ src/cmd/flux start flux run flux python yogrt.py
488
```

Where `yogrt.py` is just a Python wrapper around libyogrt for testing purposes:

```python
import ctypes
import ctypes.util


class Yogrt:
    """Minimal ctypes wrapper around libyogrt."""

    def __init__(self):
        libname = ctypes.util.find_library("yogrt")
        if libname is None:
            raise RuntimeError("libyogrt not found")
        self._lib = ctypes.CDLL(libname)
        self._lib.yogrt_remaining.restype = ctypes.c_int
        self._lib.yogrt_remaining.argtypes = []
        # yogrt_fini may not exist in all versions
        self._fini = getattr(self._lib, "yogrt_fini", None)

    def remaining(self):
        """Return remaining time in seconds, or None if unknown/unlimited."""
        val = self._lib.yogrt_remaining()
        return val if val >= 0 else None

    def close(self):
        if self._fini:
            self._fini()


print(Yogrt().remaining())
```

grondo:
Rebased now that #7409 went in. I'll set MWP here now I guess. The code is largely the same as reviewed, with the addition of `YOGRT_BACKEND` export and corresponding test.

grondo:
@mergifyio refresh

PR #7409, opened 2026-02-26
modprobe: allow rc1 tasks to manipulate environment of rc2 and rc3

### LINKED ISSUE #7407: rc1 cannot set environment variables for rc2
Problem: There is no way for rc1 tasks to modify or set the environment used by rc2, but this could be useful in some situations. For example, this could allow an rc1 task to set `YOGRT_BACKEND=flux` when Flux-under-Slurm is detected so that libyogrt works in the instance without requiring site configuration or always setting `YOGRT_BACKEND` in the broker (see #7405). There are probably other use cases here as well.

There are probably multiple ways to do this, but the easiest might be a `broker.setenv` RPC which simply sets environment variables in the broker directly. Another solution would be to make a similar RPC, but rc2 specific, where the target variables are held in a hash and only applied when starting rc2.

grondo:
This PR solves the issue described in #7407, namely that rc1 can't modify the environment that will be seen by rc2 (or rc3), since it runs as s separate process from the broker.

This PR:
- adds a `broker.setenv` RPC that allows an instance owner to set or unset environment variables in the broker's address space
- moves the initialization of the environment for runat tasks to just before they are launched, so that any changes made by rc1 are reflected in rc2 and rc3
- moves the modprobe `context.getenv()` and `needs_env` implementation from #7405 to this PR
- adds `context.setenv()` to modprobe which modprobe tasks can use to set environment variables that should be exported locally as well as to rc2 and rc3

Fixes #7407

### DIFF
--- a/doc/man1/flux-modprobe.rst
+++ b/doc/man1/flux-modprobe.rst
@@ -274,6 +274,14 @@ Each entry in the ``modules`` array supports the following keys:
    (optional, array of string) An array of broker attributes which are required
    to be set for this module to be loaded.
 
+**needs-env**
+   (optional, array of string) An array of environment variables which are
+   required to be set for this module to be loaded. That is, if any of the
+   specified environment variables are not set, then the module is disabled.
+   Note that the broker environment is checked if a variable is not found in
+   the local environment, since the rc1 environment may differ from the broker
+   environment.
+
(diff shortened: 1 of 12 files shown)


garlick (approved):
LGTM!

grondo:
Thanks! I'll set MWP.

PR #7410, opened 2026-02-28
flux-resource: eventlog: add --match-context and --timeout options

grondo:
This PR adds a couple convenience options to `flux resource eventlog` that will be useful in testing:
- `-m --match-context=KEY=VAL` allows matching one more more context keys to expected values, which is useful for waiting for expected events with `--wait=EVENT` that may occur more than once.
- `-t, --timeout=FSD` is useful for aborting a `--wait` after a timeout


grondo:
Oh, and one other trivial testsuite fix thrown in.

garlick (approved):
LGTM!  Thanks - nice improvement.

PR #7412, opened 2026-03-01
Support `NO_COLOR` environment variable in Flux utilities with `--color` option

grondo:
Many Flux utilities support a `--color` option but each implements its own processing logic, and none supports the `NO_COLOR` environment variable for disabling color by default as described at [no-color.org](https://no-color.org).

This PR first replaces replicated code for processing the standard `--color[=WHEN]` option with standard interfaces for C (`optparse_get_color()`) and Python (a new `ColorAction` `argparse.Action` class). Utilities with `--color` options are then transitioned to these interfaces to reduce duplication, unify behavior, and support the `NO_COLOR` environment variable.
Finally, documentation is standardized as well and `NO_COLOR` is added to `flux-environment(7)`.

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -430,6 +430,7 @@ $(RST_FILES): \
 	man1/common/job-other-run.rst \
 	man1/common/job-shell-options.rst \
 	man1/common/job-mustache-templates.rst \
+	man1/common/color.rst \
 	man3/common/resources.rst \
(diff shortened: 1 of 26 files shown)


garlick (approved):
LGTM!  Very nice cleanup.

grondo:
Thanks! Setting MWP after fixing conflicts in `flux-resource.py`

PR #7413, opened 2026-03-02
protect job `--progress`  bar from terminal escape codes with `--watch`

grondo:
When using `--watch --progress` with `flux watch`, `flux submit`, or `flux bulksubmit`, job output that happens to emit VT100 escape codes can corrupt the progress bar display. Since escape codes that modify terminal layout aren't likely to be useful when watching output from multiple jobs anyway, just suppress the display of sequences in the Python `JobWatcher` class when a progress bar is active.

This fixes the one use case I have which is to run the entire testsuite under `flux bulksubmit` with a progress indicator:
```
$ flux bulksubmit -o pty --watch --progress -n1 ./{} -d -v ::: t*.t
```

### DIFF
--- a/src/bindings/python/flux/job/watcher.py
+++ b/src/bindings/python/flux/job/watcher.py
@@ -9,6 +9,7 @@
 ##############################################################
 
 import os
+import re
 import sys
 import time
 from collections import Counter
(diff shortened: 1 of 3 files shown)


grondo:
Oh, this also includes a couple minor testsuite fixes found during testing.

cmoussa1 (approved):
LGTM!

grondo:
Thanks @cmoussa1! I'll set MWP.

PR #7417, opened 2026-03-03
libflux: add module loader helpers

garlick:
Problem: there is duplicated code between the two methods used to load broker modules: thread (dlopen) and exec.

This factors out the common code and adds some new functions to the public API.   These should make implementing a  broker module loader trivial in any language that can use the basic C bindings.

The new public functions are:

```c
int flux_register_default_methods (flux_t *h,
                                   const char *service_name,
                                   flux_msg_handler_t **msg_handlers[]);

int flux_module_initialize (flux_t *h, char **args, flux_error_t *error);

int flux_module_register_handlers (flux_t *h, flux_error_t *error);

int flux_module_finalize (flux_t *h, int errnum, flux_error_t *error);

```


src/common/libflux/module.c
```diff
@@ -177,6 +218,34 @@ int flux_module_register_handlers (flux_t *h, flux_error_t *error)
...
+        flux_log (h, LOG_DEBUG, "responding to post-shutdown %s", topic);
+        if (flux_respond_error (h, msg, ENOSYS, NULL) < 0)
+            flux_log_error (h, "responding to post-shutdown %s", topic);
+        flux_msg_destroy (msg);
+    }
+}
+
+int flux_module_finalize (flux_t *h, int errnum, flux_error_t *error)
+{
+    if (module_set_finalizing (h, 10., error) < 0)
```
grondo, line 376: Since this is now a public API, check for `h == NULL` here and return EINVAL if so? Same for `flux_module_initialize()`.

src/common/libflux/module.c
```diff
@@ -63,6 +111,267 @@ int flux_module_config_request_decode (const flux_msg_t *msg,
...
+    return 0;
+}
+
+int flux_module_register_handlers (flux_t *h, flux_error_t *error)
+{
+    struct module_ctx *ctx;
+    const char *name = flux_aux_get (h, "flux::name");
+
+    if (!h || !name) {
+        errno = EINVAL;
```
grondo: nit: `h` checked for `NULL` _after_ it is used in `flux_aux_get()`. Not a big deal since `flux_aux_get()` also checks `h == NULL`...

src/broker/broker.c
```diff
@@ -69,7 +69,6 @@
 
 #include "module.h"
 #include "modhash.h"
-#include "method.h"
```
grondo, line 72: Commit message typo: "susbscribe"
Also: "call this functions" -> "call these functions"

grondo (approved):
LGTM! Just minor things noted inline

garlick:
Thanks, fixed those things and will set MWP

Issue #7400, opened 2026-02-19
broker: support modules in programming languages other than C/C++


garlick:
Problem: although the broker now supports broker modules running as separate processes,  and any program that implements the interfaces described in [RFC 5/Flux Broker Modules](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_5.html) could in theory act as a broker module, the broker currently is hardwired to only spawn its  `flux-module-exec` helper, which loads the traditional broker  DSOs.

Flux services can be registered by standalone programs without them becoming "broker modules" per se, and sometimes that is preferable, for example if the life cycle of the service should not be tied to the broker.  However in other cases, it may be helpful to have such services more tightly coupled and managed directly by the broker, for example if they should be separately instantiated in each level of a Flux instance hierarchy.  It goes without saying that coding broker services only in C/C++ is limiting.

To enable this, we need to change the broker and  `flux module load --exec` to support arbitrary executable paths.

A nice follow-on would be to add a python class that supports the RFC 5 protocol so that python broker modules could be trivially written.

grondo:
>  and sometimes that is preferable, for example if the life cycle of the service should not be tied to the broker.

To clarify a bit, it is already possible to start processes (running services or not) in `rc1` and stop them in `rc3` using the `subprocess.rexec_bg()` capability. There is also the existing practice for system instance services of starting them via systemd units that have their lifetime tied to the main `flux` systemd service.

Not to argue against the main point of this issue, though, but I did want to make the point that processes can currently be tied to the broker lifetime without being modules.

Also, one benefit of loading processes that offer services as broker modules is that there's guarantee that the service is available when `flux module load` or the equivalent returns. This isn't the case with existing methods.

garlick:
Yes sorry, I had mentioned that at first and managed to delete it while editing.

> Also, one benefit of loading processes that offer services as broker modules is that there's guarantee that the service is available when flux module load or the equivalent returns. This isn't the case with existing methods.

Yes I guess you could say that when running as a module, there are structured initialization and finalization phases.  `flux module load` doesn't return until initialization is complete (by default the code before `flux_reactor_run()`).  Similarly, `flux module remove` doesn't return until finalization is complete (by default the code after `flux_reactor_run()`).

PR #7422, opened 2026-03-04
broker: support module loaders other than the built-in DSO one

garlick:
Problem: only regular dso modules can be loaded as separate processes, but in the future it may be useful to support modules written in other programming environments such as python, as noted in
- #7400

This refactors the broker module loading code to support multiple module loaders.  There are two ways to use a new loader
1) specify both the loader and the module by path e.g.
```
flux module load --loader=my_python_loader.py ./testmod.py
```
2) add a new loader and file suffix glob to the loader table in the broker.  For example add
```c
{ .cmd = "module-python-exec", .glob = ".py" }
```
then either of the following would work
```
flux module load ./testmod.py
flux module load testmod         # <--- if testmod.py is found in FLUX_MODULE_PATH
```
This is built on top of
- #7417

### DIFF
--- a/src/broker/modhash.c
+++ b/src/broker/modhash.c
@@ -395,10 +395,10 @@ static int modhash_resolve_dso (const char *name_or_null,
     if (name_or_null)
         name = strdup (name_or_null);
     else
-        name = module_dso_name (path);
+        name = module_name_frompath (path);
     if (!name) {
         errprintf (error,
-                   "error duplicating module name: %s",
+                   "error determining/duplicating module name: %s",
                    strerror (errno));
         ERRNO_SAFE_WRAP (free, path);
         return -1;
--- a/src/broker/modhash.c
+++ b/src/broker/modhash.c
@@ -1186,6 +1186,7 @@ static module_t *modhash_load_exec (modhash_t *mh,
         return NULL;
     if (!(p = module_create_exec (ctx->h,
                                   broker_uuid,
+                                  "module-exec",
                                   name,
                                   path,
                                   args,
--- a/src/broker/modhash.c
+++ b/src/broker/modhash.c
@@ -15,6 +15,7 @@
 #include <jansson.h>
 
 #include "src/common/libczmqcontainers/czmq_containers.h"
+#include "src/common/libutil/dirwalk.h"
 #include "src/common/libutil/iterators.h"
 #include "src/common/libutil/errprintf.h"
 #include "src/common/libutil/errno_safe.h"
@@ -360,6 +361,40 @@ static int modhash_load_finalize (struct modhash *mh,
     return 0;
 }
 
+static char *modhash_dso_search (const char *name,
+                                 const char *searchpath,
+                                 flux_error_t *error)
+{
+    char *pattern;
+    zlist_t *files = NULL;
+    char *path;
+
+    if (asprintf (&pattern, "%s.so*", name) < 0) {
+        errprintf (error, "out of memory");
+        return NULL;
+    }
+    if (!(files = dirwalk_find (searchpath,
+                                DIRWALK_REALPATH | DIRWALK_NORECURSE,
+                                pattern,
+                                1,
+                                NULL,
+                                NULL))
+        || zlist_size (files) == 0) {
+        errprintf (error, "module not found in search path");
+        errno = ENOENT;
+        goto error;
+    }
+    if (!(path = strdup (zlist_first (files))))
+        goto error;
+    zlist_destroy (&files);
+    free (pattern);
+    return path;
+error:
+    ERRNO_SAFE_WRAP (zlist_destroy, &files);
+    ERRNO_SAFE_WRAP (free, pattern);
+    return NULL;
+};
+
 static int modhash_resolve_dso (const char *name_or_null,
                                 const char *path_or_name,
                                 char **namep,
@@ -385,7 +420,7 @@ static int modhash_resolve_dso (const char *name_or_null,
             errno = EINVAL;
             return -1;
         }
-        if (!(path = module_dso_search (path_or_name, searchpath, error)))
+        if (!(path = modhash_dso_search (path_or_name, searchpath, error)))
             return -1;
     }
     /* If the name is not specified, derive it from the module path.
--- a/src/broker/modhash.c
+++ b/src/broker/modhash.c
@@ -13,6 +13,7 @@
 #endif
 #include <flux/core.h>
 #include <jansson.h>
+#include <fnmatch.h>
 
 #include "src/common/libczmqcontainers/czmq_containers.h"
 #include "src/common/libutil/dirwalk.h"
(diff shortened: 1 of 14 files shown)


garlick:
Rebased on current master now that #7417 is merged.

garlick:
Rebased on current master and added a commit that sets up the logging prefix in `flux_module_initialize()`  so every module loader doesn't have to do it.

garlick:
restarting bookworm after this failure
```
2026-03-09T18:47:09.9585433Z not ok 16 - update expiration to ~1 hour again and wait for poll
2026-03-09T18:47:09.9585896Z FAIL: t4100-slurm-expiration-sync.t 16 - update expiration to ~1 hour again and wait for poll
2026-03-09T18:47:09.9586292Z #	
2026-03-09T18:47:09.9586463Z #		future_time 3600 > $EXPIRATION_FILE &&
2026-03-09T18:47:09.9586781Z #		flux python wait-expiration.py $(cat $EXPIRATION_FILE) &&
2026-03-09T18:47:09.9587093Z #		flux resource R > R-poll-updated.json &&
2026-03-09T18:47:09.9587357Z #		check_expiration R-poll-updated.json
2026-03-09T18:47:09.9587575Z #	
2026-03-09T18:47:09.9587743Z not ok 17 - stop slurm-expiration-sync
2026-03-09T18:47:09.9588071Z FAIL: t4100-slurm-expiration-sync.t 17 - stop slurm-expiration-sync
2026-03-09T18:47:09.9588379Z #	
2026-03-09T18:47:09.9588534Z #		flux sproc kill --wait 2 slurm-sync
2026-03-09T18:47:09.9588758Z #	
2026-03-09T18:47:09.9588914Z # failed 2 among 17 test(s)
```

src/broker/module_exec.c
```diff
@@ -257,7 +247,7 @@ int main (int argc, char *argv[])
         optparse_print_usage (me.opts);
         exit (1);
     }
-    module = argv[optindex++];
+    path = argv[optindex++];
```
chu11, line 250: Should we check that `path` has a `/` in it?    AFAICT, `path` eventually flows to `dlopen()` which can accept a non-path.

t/module/testloader.c
```diff
@@ -0,0 +1,63 @@
...
+    if (!(h = flux_open_ex (uri, 0, &error)))
+        die_error ("flux_open", &error);
+    if (flux_module_initialize (h, &modargs, &error) < 0)
+        die_error ("flux_module_initialize", &error);
+    if (flux_module_register_handlers (h, &error) < 0)
+        die_error ("flux_module_register_handlers", &error);
+
+    /* Replace flux_reactor_run with call into module.
+     * Use path, modargs.
+     */
```
chu11: maybe say `i.e. "mod_main()"`

t/t0006-module-exec.t
```diff
@@ -180,4 +181,29 @@ test_expect_success 'legacy module cannot be loaded under new name' '
...
+	flux module load --loader=module-exec $heartbeatmod &&
+	flux module unload heartbeat
+'
+test_expect_success 'faux module loader works' '
+	flux module load --loader=$testloader $heartbeatmod args &&
+	flux module unload heartbeat
+'
+test_expect_success 'module loader cannot be specified with non-path module' '
+	test_must_fail flux module load --loader=module-exec
+'
```
chu11, line 198: this probably just fails b/c no module was specified?  I think meant to put some dummy module name in there?

chu11 (approved):
LGTM, just a few nit things I noticed.

src/broker/modhash.c
```diff
@@ -361,91 +455,87 @@ static int modhash_load_finalize (struct modhash *mh,
...
+ * On success, fill 'infop' (caller must call modinfo_release()) and return 0.
+ * On failure, fill errp and return -1.
+ * N.B. The loader is responsible to make sure the file exists and conforms
+ * to RFC 5 is the loader's, not this function.
+ */
+int modhash_resolve_bypath (const char *target,
+                            const char *name_override, // may be NULL
+                            const char *loader_override, // may be NULL
+                            struct modinfo *infop,
+                            flux_error_t *errp)
```
grondo: Should be declared `static`?

src/broker/modhash.c
```diff
@@ -361,91 +455,87 @@ static int modhash_load_finalize (struct modhash *mh,
...
-static int modhash_resolve_dso (const char *name_or_null,
-                                const char *path_or_name,
-                                char **namep,
-                                char **pathp,
-                                flux_error_t *error)
+/* Assume that target is a (possibly relative) file path.
+ * On success, fill 'infop' (caller must call modinfo_release()) and return 0.
+ * On failure, fill errp and return -1.
+ * N.B. The loader is responsible to make sure the file exists and conforms
+ * to RFC 5 is the loader's, not this function.
```
grondo: The `N.B.` comment here doesn't read quite right, probably because of a partial edit. Probably the stray `is the loader's` was accidentally left in after a change?

src/broker/module_exec.c
```diff
@@ -271,16 +271,15 @@ int main (int argc, char *argv[])
...
         if (parse_modargs (&me, modargs) < 0)
             log_err_exit ("error parsing module arguments");
         free (modargs);
 
-        name = flux_aux_get (me.h, "flux::name");
-        flux_log_set_appname (me.h, name);
 #ifdef PR_SET_NAME
+        const char *name;
+        name = flux_aux_get (me.h, "flux::name");
         (void)prctl (PR_SET_NAME, name, 0, 0, 0);
```
grondo, line 283: Is this change unrelated to this commit?

grondo (approved):
LGTM!

src/broker/module_exec.c
garlick: I'm not sure there is a problem here?  In test mode, the path comes from the command line so I suppose it's not a huge deal if dlopen() operates directly on a non-path passed in by the user.  It will just fail if it cannot resolve it.  In normal mode, the path is pre-processed before module_exec.c sees it.

Did you have a specific case in mind?  (I may be missing something!)

src/broker/module_exec.c
garlick: It is related.  The commit is about setting the log appname elsewhere.  After dropping the `flux_log_set_appname()` here, the only use of `name` became `prctl()` so to avoid "set but not used" warnings from the compiler, the aux_get moved into the PR_SET_NAME ifdef.

grondo: Totally optional improvement: mention that in the commit message.

garlick:
Thanks!  I fixed those things, rebased on current master, and will set MWP.

src/broker/module_exec.c
chu11: no specific case.  Just wondering if we should vs letting `dlopen()` handle it and fallthrough error handling as a resul.t

PR #7423, opened 2026-03-04
python: add userid, rolemask getters for Flux messages and testing

wihobbs:
Problem: currently from the Python bindings there is no way to retrieve the userid of the user sending a message or the message's rolemask.

Add a wrapper for this. Add some unit tests as well.

### DIFF
--- a/t/python/t0008-message.py
+++ b/t/python/t0008-message.py
@@ -11,6 +11,7 @@
 ###############################################################
 
 import json
+import os
 import unittest
 
 import flux
@@ -68,6 +69,10 @@ class TestMessage(unittest.TestCase):
         m.type = flux.constants.FLUX_MSGTYPE_RESPONSE
         self.assertEqual(m.type, flux.constants.FLUX_MSGTYPE_RESPONSE)
 
+        # proto block getters work
+        self.assertEqual(m.userid, flux.constants.FLUX_USERID_UNKNOWN)
+        self.assertEqual(m.rolemask, 0)
+
     def test_basic_response(self):
         m = Message(flux.constants.FLUX_MSGTYPE_RESPONSE)
         self.assertIsNotNone(m)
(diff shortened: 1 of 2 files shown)


t/python/t0008-message.py
```diff
@@ -106,6 +111,11 @@ def cb(handle, watcher, msg, called):
             self.assertEqual(msgtype, flux.constants.FLUX_MSGTYPE_RESPONSE)
             self.assertEqual(topic, "test.foo")
             self.assertIsNone(payload)
+            self.assertEqual(msg.userid, os.getuid())
+            self.assertEqual(
+                msg.rolemask,
+                int(flux.constants.FLUX_ROLE_OWNER + flux.constants.FLUX_ROLE_LOCAL),
+            )
```
grondo, line 124: Python supports bitmask or (`|`) which I think would be more appropriate here. I don't think you'd need to cast to `int` in that case, but I haven't tried it so I could be wrong.

t/python/t0008-message.py
```diff
@@ -68,6 +69,10 @@ def test_basic_request(self):
         m.type = flux.constants.FLUX_MSGTYPE_RESPONSE
         self.assertEqual(m.type, flux.constants.FLUX_MSGTYPE_RESPONSE)
 
+        # proto block getters work
+        self.assertEqual(m.userid, 0xFFFFFFFF)
```
grondo: Could you use `flux.constants.FLUX_USERID_UNKNOWN` here?

grondo (approved):
LGTM Thanks! Perhaps the tests could be expanded a bit, but I don't feel that's necessary at this point since the Python directly wraps the well-tested C API.
A couple  comments inline where the tests could be slightly improved.

Thanks!

wihobbs:
Thanks @grondo! Both of those seem like good improvements. With your approval, I'll go ahead and set MWP.

wihobbs:
Restarting the pmix check; appears to be `ECONN` to download a dependency (perhaps rate limiting?)

PR #7424, opened 2026-03-05
broker: support python broker modules

garlick:
Problem: there is no direct support for writing broker modules in python

This adds a python "module loader" and a convenience BrokerModule class to the python bindings.  This should make it pretty easy to write python broker modules, and lower the barrier of entry for creating tightly integrated Flux distributed services.

This provides a bit of documentation for would-be authors: https://flux-framework--7424.org.readthedocs.build/projects/flux-core/en/7424/python/broker_modules.html

For full disclosure, this PR was prepared with heavy assistance from Claude Code.

This currently is based upon
- #7422

### DIFF
--- a/doc/man1/flux-module.rst
+++ b/doc/man1/flux-module.rst
@@ -21,7 +21,7 @@ DESCRIPTION
 
 .. program:: flux module
 
-:program:`flux module` manages dynamically loadable :man1:`flux-broker` modules.
+:program:`flux module` manages :man1:`flux-broker` modules.
 Modules are automatically loaded at Flux instance start up (rc1) and unloaded
 at Flux instance shutdown (rc3) using :man1:`flux-modprobe`, which is
 configured to be aware of module dependencies.  :program:`flux module` is
@@ -42,17 +42,24 @@ tracing.
 
 Broker modules are often implemented as dynamic shared objects loaded
 into broker threads, but they may also be loaded into separate processes.
-This may be necessary for isolation, or to support incompatible programming
-runtime environments.  A *module loader* process assists with loading modules
-as separate processes, establishing communication with the broker, loading
-and executing the module, and communicating with the broker's module
-management framework.  The loader may be inferred from the module file name
-suffix, or specified on the command line.  The following module loaders are
-provided by flux-core:
+This may be necessary for isolation, or to support modules written in
+languages with their own runtime, such as Python.  A *module loader*
+process assists with loading such modules, establishing communication with
+the broker, executing the module, and reporting completion back to the
+broker's module management framework.  The loader is inferred from the
+module file name suffix, or may be specified on the command line.  The
+following module loaders are provided by flux-core:
 
 module-exec (.so*)
-  An alternate way to load traditional dynamic shared objects, with more
-  isolation than direct broker loading.
+  Loads a dynamic shared object as a separate process, providing more
+  isolation than direct broker thread loading.
+
+module-python-exec (.py)
+  Loads Python broker modules.  The Python file must define either a
+  :func:`mod_main` function or a single
+  :class:`~flux.brokermod.BrokerModule` subclass (see `MODULE ENTRY POINT`_).
+  The :class:`~flux.brokermod.BrokerModule` base class simplifies writing
+  Python broker modules with decorator-based handler registration.
 
 A small number of modules that are integral to the broker are automatically
 loaded early in broker start-up.  Although hard-wired to a particular name
@@ -68,9 +75,10 @@ load
 
 .. program:: flux module load
 
-Load *module*, which may be either the path to a shared object file, including
-the ``.so`` suffix, or the basename of a shared object file on the
-:envvar:`FLUX_MODULE_PATH`, without the suffix.
+Load *module*, which may be the path to a shared object file (including the
+``.so`` suffix), the path to a Python file (including the ``.py`` suffix),
+or the basename of either on the :envvar:`FLUX_MODULE_PATH`, without the
+suffix.
 
 When :program:`flux module load` completes successfully, the new module has
 entered the running state (see LIST OUTPUT below).
@@ -288,7 +296,7 @@ The *list* command displays one line for each unique (as determined by
 SHA1 hash) loaded module.
 
 **Module**
-   The value of the **mod_name** symbol for this module.
+   The module name.
 
 **Idle**
    Idle times are defined as the number of seconds since the module last sent
@@ -305,20 +313,30 @@ SHA1 hash) loaded module.
    displayed in a comma-separated list.
 
 **Path**
-   The full path to the broker module shared object file (only shown with
+   The full path to the broker module file (only shown with
    the **-l, --long** option).
 
 
-MODULE SYMBOLS
-==============
+MODULE ENTRY POINT
+==================
 
-All Flux modules define the following global symbols:
-
-**const char \*mod_name;**
-   A null-terminated string defining the module name.
+C broker modules define the following entry point:
 
 **int mod_main (void \*context, int argc, char \**argv);**
    An entry function.
+
+Python broker modules may define one of the following entry points:
+
+**mod_main (h, \*args)**
+   An entry function called with a :class:`flux.Flux` handle and any
+   module arguments.
+
+**class MyModule(BrokerModule)**
+   As a shortcut, a Python module that defines a single
+   :class:`~flux.brokermod.BrokerModule` subclass and no :func:`mod_main`
+   has its entry point synthesized automatically as
+   ``MyModule(h, *args).run()``.  If multiple subclasses are present,
+   :func:`mod_main` must be defined explicitly to resolve the ambiguity.
 
 
 RESOURCES
--- a/doc/python/broker_modules.rst
+++ b/doc/python/broker_modules.rst
@@ -1,0 +1,149 @@
+.. _python_broker_modules:
+
+Python Broker Modules
+=====================
+
+Flux broker modules extend the broker with new services.  Traditionally
+modules are written in C as dynamic shared objects, but Flux also supports
+modules written in Python.  A Python broker module is a :file:`.py` file that
+defines either a :func:`mod_main` entry function or a single
+:class:`~flux.brokermod.BrokerModule` subclass.
+
+
+Loading a Python module
+-----------------------
+
+Python modules are loaded and removed with the same :man1:`flux-module`
+commands used for C modules:
+
+.. code-block:: console
+
+   $ flux module load /path/to/mymod.py
+   $ flux ping mymod
+   $ flux module remove mymod
+
+The module name defaults to the file's basename without the ``.py`` suffix
+and may be overridden with ``--name``:
+
+.. code-block:: console
+
+   $ flux module load --name myservice /path/to/mymod.py
+
+
+The mod_main entry point
+------------------------
+
+A Python broker module may define a top-level function named :func:`mod_main`:
+
+.. code-block:: python
+
+   def mod_main(h, *args):
+       ...
+
+``h`` is a :class:`flux.Flux` handle connected to the broker.  ``args``
+is a (possibly empty) tuple of strings passed after the module path on the
+:program:`flux module load` command line.  The function must run the reactor
+(directly or via :class:`~flux.brokermod.BrokerModule`) and return only after
+the reactor exits.
+
+Defining :func:`mod_main` explicitly is optional when using
+:class:`~flux.brokermod.BrokerModule` — see `Using BrokerModule`_ below.
+
+
+Using BrokerModule
+------------------
+
+:class:`~flux.brokermod.BrokerModule` is a base class that handles the
+boilerplate of registering message handlers, managing the reactor, and
+looking up the module name.  Subclass it and use the
+:func:`~flux.brokermod.request_handler` and
+:func:`~flux.brokermod.event_handler` decorators to declare handlers:
+
+.. code-block:: python
+
+   from flux.brokermod import BrokerModule, event_handler, request_handler
+
+
+   class MyModule(BrokerModule):
+
+       @request_handler("info")
+       def info(self, msg):
+           self.handle.respond(msg, {"name": self.name})
+
+       @event_handler("shutdown", prefix=False)
+       def on_shutdown(self, msg):
+           self.stop()
+
+When a module file contains exactly one :class:`~flux.brokermod.BrokerModule`
+subclass and no :func:`mod_main`, the loader synthesizes the entry point
+automatically as ``MyModule(h, *args).run()``, so no explicit
+:func:`mod_main` is required.  If multiple subclasses are present,
+:func:`mod_main` must be defined to resolve the ambiguity:
+
+.. code-block:: python
+
+   def mod_main(h, *args):
+       MyModule(h, *args).run()
+
+Handler methods receive a :class:`~flux.message.Message` object.  For request
+handlers, call :meth:`flux.Flux.respond` on the handle to send a response.
+
+By default each handler's topic is prefixed with the module name, so
+``@request_handler("info")`` registers ``mymod.info``.  Pass
+``prefix=False`` to use the topic string as-is.
+
+By default, only requests from the instance owner are delivered to a handler.
+Pass ``allow_guest=True`` to also accept requests from non-owner users:
+
+.. code-block:: python
+
+   @request_handler("status", allow_guest=True)
+   def status(self, msg):
+       self.handle.respond(msg, {"running": True})
+
+The :attr:`~flux.brokermod.BrokerModule.handle` property returns the
+:class:`flux.Flux` handle, giving full access to the Flux Python API for
+RPCs, KVS operations, timers, and other watchers.
+
+
+Error handling
+--------------
+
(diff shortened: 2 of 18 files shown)


wihobbs:
I want to test out this interface and give it the longer review it deserves, but one thing I noticed is that these messages show up when a boiler-plate `BrokerModule` is unloaded via `exit`-ing a test instance:

```shell
$ exit
exit
Mar 06 20:47:14.942585 UTC 2026 broker.err[0]: notify: loader (module-python-exec) killed by Terminated
Mar 06 20:47:14.942612 UTC 2026 broker.crit[0]: notify module runtime failure
```

garlick:
Oh that might be expected if the module was manually loaded and not unloaded before the shutdown.  A warning is also issued for broker modules loaded as threads that are still loaded at shutdown, although I think the error message tells you exactly what happened in that case and should here too.

garlick:
The other question that came up on slack was how to indicate via the decorators that a method is available to guests.  (By default, guests are rebuffed)

garlick:
rebased on #7422 to incorporate its changes.  Also
- squashed incremental development in flux-module(1) (thanks @wihobbs !)
- added some missing python reactor interfaces, pulled from other work
- added a BrokerModule.setrunning() method that was missing

garlick:
Tacked on a fixup commit that handles ValueError exceptions without a backtrace.  I'm not 100% sure that's always the right thing.

src/broker/module_dso.h
```diff
@@ -15,14 +15,6 @@
 
 #include "module.h"
 
-/* Search 'searchpath', a colon-separated list of directories, for a file
```
wihobbs: This is kind of a useful comment and IIUC with s/.so//g is still relevant to the new `static char *modhash_dso_search` function. Can it be preserved?

src/broker/modhash.c
```diff
@@ -361,91 +455,87 @@ static int modhash_load_finalize (struct modhash *mh,
...
-static int modhash_resolve_dso (const char *name_or_null,
-                                const char *path_or_name,
-                                char **namep,
-                                char **pathp,
-                                flux_error_t *error)
+/* Assume that target is a (possibly relative) file path.
+ * On success, fill 'infop' (caller must call modinfo_release()) and return 0.
+ * On failure, fill errp and return -1.
+ * N.B. The loader is responsible to make sure the file exists and conforms
+ * to RFC 5 is the loader's, not this function.
```
wihobbs: N.B. wording is a bit strange, perhaps 
```suggestion
 * N.B. The loader is responsible to make sure the file exists and conforms
 * to RFC 5, not this function.
```

doc/man1/flux-module.rst
```diff
@@ -22,7 +22,43 @@ DESCRIPTION
 .. program:: flux module
 
 :program:`flux module` manages dynamically loadable :man1:`flux-broker` modules.
-
+Modules are automatically loaded at Flux instance start up (rc1) and unloaded
+at Flux instance shutdown (rc3) using :man1:`flux-modprobe`, which is
+configured to be aware of module dependencies.  :program:`flux module` is
```
wihobbs, line 27: Nit: double space.

src/modules/Makefile.am
```diff
@@ -35,6 +35,8 @@ libmodule_builtins_la_LIBADD = \
 	$(builddir)/overlay/liboverlay.la
 libmodule_builtins_la_LDFLAGS = $(san_ld_zdef_flag)
 
+dist_fluxmod_SCRIPTS =
```
wihobbs: What is the point of this? (Should it be documented in the commit message or a comment?)

src/common/libflux/conf_builtin.c
```diff
@@ -84,7 +84,8 @@ static struct builtin builtin_tab[] = {
     {
         .key = "module_path",
         .val_installed = FLUXMODDIR,
-        .val_intree = ABS_TOP_BUILDDIR "/src/modules/.libs",
+        .val_intree = ABS_TOP_BUILDDIR "/src/modules/.libs:" \
```
wihobbs, line 87: This (to me) feels like it belongs as part of `3e09077`. It seems like the same problem (Python modules aren't found when building from a source directory/builddir). Also, the commit message says "Python _scheduler_ modules" but should it just be "Python modules?"

doc/man1/flux-module.rst
```diff
@@ -42,17 +42,24 @@ tracing.
...
-management framework.  The loader may be inferred from the module file name
-suffix, or specified on the command line.  The following module loaders are
-provided by flux-core:
+This may be necessary for isolation, or to support modules written in
+languages with their own runtime, such as Python.  A *module loader*
+process assists with loading such modules, establishing communication with
+the broker, executing the module, and reporting completion back to the
+broker's module management framework.  The loader is inferred from the
+module file name suffix, or may be specified on the command line.  The
+following module loaders are provided by flux-core:
```
wihobbs, line 51: It looks like everything up to this line was initially added in `25e37a9` and IMHO it would be more appropriate for these updates to all be in one commit, so that the documentation + correct language is added in one fell swoop. But up to you.

doc/python/broker_modules.rst
```diff
@@ -0,0 +1,130 @@
...
+       # args == ('--verbose', '--count=4')
+       MyModule(h, *args).run()
+
+Arguments are split on whitespace; quoting is not supported.
+
+
+API reference
+-------------
+
+.. autoclass:: flux.brokermod.BrokerModule
```
wihobbs, line 143: I really like this, and I particularly like how the inclusion under the Python/ heading makes the autodoc work nicely. I need to keep this in mind for the CLI plugins.

wihobbs (commented):
I spent time on Friday afternoon playing with this interface, and here's a record of my test plugin and functionality.

I was able to make a Python test broker module that subscribes to the job-manager journal of events, then forwards a user's events from the journal to them. The journal is currently restricted to the instance owner and such a plugin would help work around that to get users notified of their jobs. (This is a first step to getting the notification service working we were originally planning as a Python service under systemd.)

```python
import logging

import flux
import flux.job
from flux.brokermod import BrokerModule, event_handler, request_handler


class MyJobNotifications(BrokerModule):
    def __init__(self, h, *args):
        super().__init__(h, *args)
        # Because the BrokerModule class doesn't start the reactor
        # until .run() is called, you should be able to register
        # additional watchers in init() and expect them to be started
        # when .run() is called
        journal = flux.job.journal.JournalConsumer(self.handle)
        self.clients = {}
        journal.start()
        journal.set_callback(self.check_journal_event)
        self.LOGGER = logging.getLogger("flux")

    def check_journal_event(self, event):
        job_meta = flux.job.get_job(self.handle, event.jobid)
        # This won't exit gracefully if there's no jobid or userid 
        # in the job (latter shouldn't be possible)
        try:
            userid = int(job_meta["userid"])
        except KeyError as e:
            self.LOGGER.error(f"Error unpacking userid for job: {job_meta}")
        if userid in self.clients.keys():
            self.handle.respond(
                self.clients[userid],
                payload={
                    "jobid": event.jobid,
                    "timestamp": event.timestamp,
                    event.name: event.context_string,
                },
            )

    @request_handler("disconnect")
    def disconnect(self, msg):
        self.clients.pop(msg.userid)

    @request_handler("register")
    ## Currently restricted to instance owner. Needs appropriate rolemask.
    def register(self, msg):
        self.clients[msg.userid] = msg
        self.handle.respond(
            msg, {"hello": "You have registered with the MyJobNotification service"}
        )

    @request_handler("info")
    def info(self, msg):
        self.handle.respond(msg, {"name": self.name})

    @event_handler("shutdown", prefix=False)
    def on_shutdown(self, msg):
        self.stop()


def mod_main(h, *args):
    MyJobNotifications(h, *args).run()

```

After being loaded in the broker, a user can subscribe to it with:

```python
import flux, flux.job

def print_journal_entry(resp):
    print(resp.get())
    resp.reset()

h = flux.Flux()
m = h.rpc("notify.register", flags=flux.constants.FLUX_RPC_STREAMING).then(print_journal_entry)
h.reactor_run() 
```
It ends up looking like this:
```
$ flux python notify_test.py
{'hello': 'You have registered with the MyJobNotification service'}
{'jobid': 11934791368704, 'timestamp': 1773097503.4717948, 'finish': '{"status":0}'}
{'jobid': 11934791368706, 'timestamp': 1773097503.4718564, 'finish': '{"status":0}'}
{'jobid': 11934791368708, 'timestamp': 1773097503.4718819, 'finish': '{"status":0}'}
{'jobid': 11934808145921, 'timestamp': 1773097503.4720063, 'finish': '{"status":0}'}
```

Thank you, @garlick. This is a fun interface and I appreciate you documenting above some nice-to-haves we discussed in Slack. A better programmer than me should review the C code, but the Python looks good. ✅

doc/man1/flux-module.rst
grondo: Double-space after period is an older convention some people may prefer, especially with monospace fonts. I've just been allowing either single or double space after period, though I believe most modern standards specify single space (Chicago, APA, etc.).

wihobbs: I didn't realize this is consistent with the rest of the document (I though it was a mistake.) Feel free to ignore.

src/broker/modhash.c
garlick: already fixed in the other PR

src/broker/module_dso.h
garlick: Whoops, I think this comment was for the PR that was just merged (this one was built on top of that one).   I'll add a commit here that improves the comment above `modloader_find_byname()` which where the referenced function ended up.

doc/man1/flux-module.rst
garlick: Oh, interesting, if I knew that detail about space following period, I have forgotten!

src/modules/Makefile.am
garlick: That was a placeholder for python modules that would be installed but you're right, it has no business being here.

src/broker/module_dso.h
wihobbs: Thanks! Sorry that several of my comments should have been on #7422. 

Since all of the C code I felt unqualified to approve got a thorough review in that PR, I'm good to approve this as soon as the distcheck test passes. (Unless you wanted to add some of the features outlined ^)

src/common/libflux/conf_builtin.c
garlick: Yeah I'll combine those two and fix the comment - thanks!

doc/man1/flux-module.rst
garlick: Oh sorry, I did squash the commits to the man page that are in this PR, but that first big intro was added in the other dependent PR which was already merged, so I guess water under the bridge.

wihobbs: Ah, I understand now. My confusion -- I think this was my first time reviewing a PR with another PR dependent.

src/broker/flux-module-python-exec.py
```diff
@@ -0,0 +1,105 @@
...
+
+    if lib.flux_module_register_handlers(h.handle, error) < 0:
+        die_error("flux_module_register_handlers", error[0])
+
+    spec = importlib.util.spec_from_file_location("flux_module", path)
+    if spec is None:
+        die(f"could not load module from {path}")
+    mod = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(mod)
+
```
grondo: This code could be replaced with [`flux.importer.import_path`](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/python/autogenerated/flux.importer.html#flux.importer.import_path)

src/broker/flux-module-python-exec.py
```diff
@@ -0,0 +1,105 @@
...
+    try:
+        mod.mod_main(h, *args)
+    except ValueError as exc:
+        # User-facing argument errors: print message only, no traceback
+        print(f"{os.path.basename(path)}: {exc}", file=sys.stderr)
+        errnum = errno_mod.EINVAL
+    except Exception:  # pylint: disable=broad-except
+        traceback.print_exc()
+        errnum = errno_mod.ECONNRESET
+
```
grondo, line 101: I wonder if a nice improvement here would be to check for an instance of a `BrokerModule` subclass in the module when `mod.mod_main()` isn't found, and call `BrokerModule(h, *args).run()` if found, since that will be contents of most `mod_main()` implementations anyway.

src/broker/Makefile.am
```diff
@@ -137,4 +137,7 @@ test_runat_t_CPPFLAGS = $(test_cppflags)
 test_runat_t_LDADD = $(test_ldadd)
 test_runat_t_LDFLAGS = $(test_ldflags)
 
+dist_fluxcmd_SCRIPTS = \
+	flux-module-python-exec.py
```
grondo, line 141: I don't think it is a use case for any users to call `flux module-python-exec`, so would the loader be better placed in libexecdir so it isn't found by e.g. tab completion and the similar command hinter?

(Probably this suggestion is too late since the exec-based loader is probably already in cmddir, but just thought I'd throw the idea out there.)

src/bindings/python/flux/core/handle.py
```diff
@@ -15,7 +15,14 @@
 from _flux._core import ffi, lib
 from flux.constants import FLUX_POLLERR, FLUX_POLLIN, FLUX_POLLOUT
 from flux.core.inner import raw
-from flux.core.watchers import FDWatcher, SignalWatcher, TimerWatcher
+from flux.core.watchers import (
+    CheckWatcher,
```
grondo, line 19: Commit message calls out `PrepWatcher` but implementation has `PrepareWatcher`

grondo (commented):
This is excellent! I definitely think request handler allowmask should be added before merge if that's simple. Made a few other minor comments inline.

src/broker/Makefile.am
garlick: Well I did add a "test mode" to the DSO module loader thinking maybe it would be useful to be able to load modules in a test harness without the broker.  However, that would require a lot of boiler plate in the test harness to simulate a broker, and maybe it's not useful.  In fact perhaps "test mode" should be backed out in that loader.

I was going to say we need the environment set up so the `flux(1)` front end is useful but duh! the broker was already run with that so it's extra.

OK good suggestion, will see about changing that.

src/bindings/python/flux/core/handle.py
garlick: Stale version of that commit?  Currently I have
```
commit 4529c874ed24ef9e7bd44d0ddff504ae4260401e
Author: Jim Garlick <garlick.jim@gmail.com>
Date:   Sun Mar 8 16:07:36 2026 -0700

    python: add missing reactor watcher interfaces
    
    Problem: The python bindings are lacking a few reactor interfaces
    that might come in handy when writing complex reactive programs.
    
    Add CheckWatcher
    Add PrepWatcher
    Add IdleWatcher
    Add TimerWatcher.reset()
    
    Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>
```

garlick:
OK so besides the inline comments, two todo's are

- [ ] improve error message when broker is unloaded with running module
- [ ] add flag for guest access to message handler decorator

src/bindings/python/flux/core/handle.py
grondo: Yeah, I was just calling out that the commit message says `PrepWatcher` but the actual name appears to be `PrepareWatcher` (pretty minor)

src/broker/flux-module-python-exec.py
grondo: Actually on second thought this is just a convenience and it would add extra documentation requirements that might be onerous at this point. It was an idle idea and could be ignored for now if you'd like.

src/broker/Makefile.am
garlick: This will require a bit of code to run the "loader" processes directly, to handle in-tree and installed.  Not a lot, but I'm wondering if it would be OK to defer to another PR.

src/bindings/python/flux/core/handle.py
garlick: Oh oops!

src/broker/flux-module-python-exec.py
garlick: Already done, and seems good to me!

garlick:
I think I've addressed all the comments and opened issues on a couple items I'd like to defer as they are not specific to python modules (and my brain hurts):
- #7427
- #7436

Looks like there's a flake8 error to address so after CI runs to completion I'll fix that and whatever else may come up and then this could be approved IMHO, although maybe the reviewers will want to make sure the optional mod_main() change and the decorator allowing guests change are OK first.  (See the narrative documentation for overview).

grondo (approved):
Nice! LGTM!

wihobbs (approved):
This PR is exciting; nice work @garlick!

doc/man1/flux-module.rst
grondo: More than you wanted to know: One space after period is highly recommended since 1990s :-)

https://www.chicagomanualofstyle.org/qanda/data/faq/topics/OneSpaceorTwo.html

garlick:
Thanks!  Setting MWP.

Issue #7426, opened 2026-03-06
Flux under Slurm is erroring out in modprobe with `ETIMEDOUT`


wihobbs:
Starting a flux instance under a 2 node allocation on mutt, and the instance won't start:
```
  mutt6 ~/flux-core/build (pymod-notify)$ salloc -N2
salloc: Granted job allocation 121066
salloc: Waiting for resource configuration
salloc: Nodes mutt[107-108] are ready for job
  mutt107 ~/flux-core/build (pymod-notify)$ srun -N2 --pty --mpi=pmi2 --export=NONE ./src/cmd/flux start
Mar 06 15:01:44.639200 PST 2026 broker.err[0]: rc1.0: slurm-expiration-sync: [Errno 110] ETIMEDOUT: Connection timed out
Mar 06 15:01:44.650836 PST 2026 broker.err[0]: rc1.0: flux modprobe rc1 Exited (rc=1) 5.4s
Mar 06 15:01:44.867469 PST 2026 broker.err[1]: rc3.0: flux-modprobe: ERROR: Upstream Flux broker is offline. Try again later.
Mar 06 15:01:44.869429 PST 2026 broker.err[1]: rc3.0: flux-modprobe: DEBUG: Traceback (most recent call last):
Mar 06 15:01:44.869447 PST 2026 broker.err[1]: rc3.0:   File "/g/g0/hobbs17/flux-core/build/../src/bindings/python/flux/util.py", line 401, in __call__
Mar 06 15:01:44.869462 PST 2026 broker.err[1]: rc3.0:     main_func()
Mar 06 15:01:44.869476 PST 2026 broker.err[1]: rc3.0:   File "/g/g0/hobbs17/flux-core/build/../src/cmd/flux-modprobe.py", line 302, in main
Mar 06 15:01:44.869489 PST 2026 broker.err[1]: rc3.0:     args.func(args)
Mar 06 15:01:44.869503 PST 2026 broker.err[1]: rc3.0:   File "/g/g0/hobbs17/flux-core/build/../src/cmd/flux-modprobe.py", line 76, in rc3
Mar 06 15:01:44.869516 PST 2026 broker.err[1]: rc3.0:     M.configure_modules()
Mar 06 15:01:44.869530 PST 2026 broker.err[1]: rc3.0:   File "/g/g0/hobbs17/flux-core/build/../src/bindings/python/flux/modprobe.py", line 985, in configure_modules
[...]

PR #7428, opened 2026-03-07
fix slurm-expiration-sync rc1 task for multi-node Slurm jobs

grondo:
As @wihobbs discovered in #7426, the slurm-expiration-sync task introduced in v0.83.0 is broken for multi-node jobs. This is a critical error that causes all multi-node Flux jobs under Slurm to fail :-(

This PR greatly simplifies the affected rc1 task by removing the attempted synchronization. The helper script is now launched after the resource module has started, and it waits for the resource-define event, which will arrive after rc1 in any Flux instance with multiple brokers.

The tests are updated to run with `flux start -s2` so the multiple broker case is at least validated.

This is the minimum viable fix for now, there is still room for further improvements in the future to further reduce the likelihood the first timeleft check comes before the expiration update (see e.g. @garlick's idea in  #7426), In practice, I was unable to reproduce the race.

### DIFF
--- a/etc/modprobe/rc1.py
+++ b/etc/modprobe/rc1.py
@@ -17,7 +17,6 @@ from subprocess import Popen
 import flux.kvs
 import flux.subprocess as sp
 from flux.modprobe import run_all_rc_scripts, task
-from flux.resource import ResourceJournalConsumer
 
 
 def setup(context):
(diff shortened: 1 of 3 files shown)


garlick (approved):
👍

grondo:
Thanks! I'll set MWP and plan to tag a patch release after this.

PR #7430, opened 2026-03-09
enable `flux exec --jobid` in multi-user instance via RFC 42 signing

grondo:
This PR adds support for signing RFC 42 subprocess protocol requests in libsubprocess clients and verification in the server. The end result is that `flux exec --jobid` now works safely in a multi-user instance for guest jobs.

The subprocess server implementation now allows a server to be put into "sign only" mode when registering a security context. In sign-only mode, a server will only accept signed requests (conforming to RFC 42), and the signing user must match the message credential as well as the current userid of the process. Only the server in the job shell is set up in sign-only mode, and only for a multi-user job, i.e. where the shell userid doesn't match the instance owner. Therefore, there is no change in behavior for single-user instances. In sign-only mode, a subprocess server will reject unsigned requests. A subprocess server not in signed mode will reject signed requests.

On the client side, a new `FLUX_SUBPROCESS_FLAGS_SIGN` flag is introduced, which clients use to enable signed requests. When this flag is used, a security context is created on demand and cached in the handle so that future requests using the same handle can reuse it. A `sign` flag is added to the rexec ctx so that all future requests for a given subprocess (write/kill) are also automatically signed.

The `flux.subprocess` bindings are also given signing support.

Test-only options are added to the job shell and `flux exec` to force signing even in a single-user instance. This allows significant test coverage in the normal testsuite as long as flux-core is built with flux-security and signing is configured and working. The system tests in `t/system/0005-exec.t` are also greatly expanded to cover some real-world scenarios.

Example:
```console
$ flux uptime
 08:05:09 run 1.9d,  owner flux,  depth 0,  size 5,  1 offline
$ flux getattr security.owner
500
$ id
uid=1001(grondo) gid=1001(grondo) groups=1001(grondo),4(adm),27(sudo),100(users)
$ flux submit -N4 sleep inf
ƒ2YaQRstpMT5
$ flux exec --jobid=ƒ2YaQRstpMT5 id
uid=1001(grondo) gid=1001(grondo) groups=1001(grondo),4(adm),27(sudo),100(users)
uid=1001(grondo) gid=1001(grondo) groups=1001(grondo),4(adm),27(sudo),100(users)
uid=1001(grondo) gid=1001(grondo) groups=1001(grondo),4(adm),27(sudo),100(users)
uid=1001(grondo) gid=1001(grondo) groups=1001(grondo),4(adm),27(sudo),100(users)
$ sudo -u flux flux exec --jobid=ƒ2YaQRstpMT5 id
[sudo] password for grondo:
flux-exec: Error: rank 0: id: request signature required
flux-exec: Error: rank 2: id: request signature required
flux-exec: Error: rank 3: id: request signature required
flux-exec: Error: rank 4: id: request signature required
$ flux job wait-event -Hp exec ƒ2YaQRstpMT5 shell.init
[Mar09 08:05] shell.init service="1001-shell-f2YaQRstpMT5" leader-rank=0 size=4
$ sudo -u pi flux exec --sign --service=1001-shell-f2YaQRstpMT5.rexec --rank=0 id
flux-exec: Error: rank 0: id: signing userid does not match server userid

### CODE REVIEW

File: src/common/libsubprocess/client.c [hunk 1]
```diff
@@ -103,6 +107,132 @@ static struct rexec_ctx *rexec_ctx_create (flux_cmd_t *cmd,
...
+    const char *signature;
+    flux_future_t *f = NULL;
+    json_t *o = NULL;
+
+    if (!(o = json_string (topic))
+        || json_object_set_new (payload, "topic", o) < 0) {
+        json_decref (o);
+        errno = ENOMEM;
+        goto out;
+    }
```
garlick, line 151: The comment block says payload is consumed via json_decref on success and failure, but if json_string() fails at the top, it is not consumed.

This also doesn't conform with this RFC 7 rule about errors and side effects
https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_7.html#error-handling

src/common/libsubprocess/client.c
```diff
@@ -103,6 +107,132 @@ static struct rexec_ctx *rexec_ctx_create (flux_cmd_t *cmd,
...
+            || json_object_set_new (payload, "signature", osig) < 0) {
+            json_decref (osig);
+            errno = ENOMEM;
+            goto out;
+        }
+        f = flux_rpc_pack (h, topic, rank, flags, "O", payload);
+    }
+out:
+    json_decref (payload);
+    free (s);
```
garlick: json_decref() and free() might clobber errno

src/common/libsubprocess/server.c
```diff
@@ -1288,6 +1288,16 @@ void subprocess_server_set_security (subprocess_server_t *s,
...
 
+void subprocess_server_allow_rolemask (subprocess_server_t *s,
+                                       uint32_t rolemask)
+{
+    /* handlers[] is NULL-terminated (invariant maintained by
+     * flux_msg_handler_addvec / flux_msg_handler_delvec).
+     */
+    for (int i = 0; s->handlers[i] != NULL; i++)
+        flux_msg_handler_allow_rolemask (s->handlers[i], rolemask);
+}
```
garlick, line 1299: This is a nice minimal change but one worry is it leaves the htab[] there in the code with 0 in the rolemask field implying to a casual observer that the RPCs are restricted.  Have you considered changing the htab[] to be permissive and then having `server_auth_unpack()` perform the rolemask check if signing is not required by the server?

Edit: seeing how this plays out in the use case, it does seem pretty clear.  Fine to ignore!

garlick (approved):
Nice work!  This seems solid to me. I had some minor comments inline but nothing much!

src/common/libsubprocess/client.c
grondo: Good catch! Will fix. 

Consuming the payload was probably the wrong choice since this function is wrapped by `rpc_pack_signed()` which can easily drop the payload reference after success or failure.

The tricky part is that payload is modified (`topic` field added) before the signature. If there's an error after the fact we have to somehow undo the modification of `payload` or stash a copy of the original payload so it is not modified on failure, only to return and destroy it anyway. I'd hate to copy the payload on every call when likelihood of failure is low, so removing `topic` on failure seems like the right approach.

We can document that the function modifes payload on success since this doesn't strictly violate RFC 7. It is an internal function and the payload is mean to be destroyed immediately after success, so hopefully this is acceptable.

Edit: Oh, `topic` is already removed on success case so simple to add for failure case. `signature` will be present on success though.

grondo: Ok, forced pushed an update that doesn't consume `payload` in `rpc_send_signed()`. Much cleaner!

grondo:
Ok, setting MWP here now. Thanks!

PR #7438, opened 2026-03-11
testsuite: improve reliability of some tests

grondo:
This PR fixes a few tests that have been seen to fail occasionally in the testsuite.

Edit: since some of these failures aren't easily reproduced, these are all "probable" fixes. We can let the associated issues be closed and reopen if we see failures in the future.


chu11 (approved):
LGTM!

grondo:
Thanks! I've set MWP

PR #7441, opened 2026-03-12
job-info: fix resource leaks

grondo:
While reproducing an issue running sharness tests under `FLUX_TEST_VALGRIND=t`, valgrind reported some memory leaks in the job-info module.

This PR fixes the two leaks discovered, and also fixes a race in one test that was reproducible when the test was run under valgrind.

### DIFF
--- a/src/modules/job-info/lookup.c
+++ b/src/modules/job-info/lookup.c
@@ -647,6 +647,7 @@ void update_lookup_cb (flux_t *h,
         errmsg = error.text;
         goto error;
     }
+    json_decref (keys);
     return;
 
 error:
--- a/src/modules/job-info/watch.c
+++ b/src/modules/job-info/watch.c
@@ -426,7 +426,7 @@ static int watch (struct info_ctx *ctx,
         if ((ret = eventlog_allow_lru (w->ctx,
                                        w->msg,
                                        w->id)) < 0)
-            return -1;
+            goto error;
 
         if (ret)
             w->allow = true;


t/t2233-job-info-update.t
```diff
@@ -302,9 +302,45 @@ test_expect_success 'job-info: lookup current works with jobspec' '
 	flux cancel $jobid
```
chu11, line 302: commit message typo "udpate"

t/t2233-job-info-update.t
```diff
@@ -302,9 +302,45 @@ test_expect_success 'job-info: lookup current works with jobspec' '
...
+	        key, expected = expected_list.pop(0)
+	    f.reset()
+	EOF
+'
 test_expect_success NO_CHAIN_LINT 'job-info: update watch works with jobspec' '
 	jobid=$(flux submit --urgency=hold true) &&
 	flux update $jobid duration=100s &&
+	run_timeout 30 flux python update-watcher.py $jobid jobspec \
+	    attributes.system.duration=100.0 &&
 	watchers=$(get_update_watchers)
```
chu11, line 308: i think we can just use the `check_duration()` function just a bit above this code instead.

BUT if this race is generally possible (i.e. job-info super busy and the update occurs after the update-watcher background process has started), then this race might occur in a few other places.

I'm auditing the tests right now, i think 1-2 other places it might be possible.

grondo: Ah, yes, `update-watcher.py` kind of duplicates what `check_duration()` does. The reason I started down that path was to replace the pattern of a C program that emits json on stdout to a file, a set of `waitfile` invocations to wait for one or more expected values, then checking those values with `jq` later. However, it was too big a project to modify the whole test to take advantage of that, so I gave up.

I'll drop this commit and let you push a separate PR that fixes the test with `FLUX_TEST_VALGRIND`

grondo: Ok, I've dropped this commit.

chu11: ok, it ends up i think that this might be the only racy test within this file.

All of the other tests in the file use `expiration_add()` which manually adds a resource update event into the eventlog.  So we know for a fact that the update has "propogated", b/c we know it is in the eventlog before the `update-watch` has even begun.  My memory might be hazy, but we likely added these tests before the `flux-update` command existed.

So later jobspec update support was added, this racy test was added, it uses `flux-update`, that sends the update to `job-manager`, event queues of updates are involved .... racy.

Sooo ... two options, we could manually update the eventlog update entry, so it's like the other tests.  Or we could use `check_duration()`.    Just using `check_duration()` would be easiest, but with a nice comment on why it is racy vs the tests above it.

Lemme know if you wanna sneak a change in or I can just do a separate PR with this simple fix.

chu11 (approved):
if we're just doing the memory leaks, LGTM

t/t2233-job-info-update.t
grondo: Up to you, if you want to push to this branch feel free! We'll let @garlick review the overall PR.

grondo: Oh, there's also a failure in `t2230-job-info-lookup.t` that seems related:
```
ok 11 - flux job info jobspec fails on bad id

expecting success:
	echo $jobid > updated_R.id &&
	flux job info $jobid R | jq -e ".execution.expiration == 0.0" &&
	flux update $jobid duration=300 &&
	flux job info $jobid R | jq -e ".execution.expiration > 0.0" &&
	flux job info $jobid eventlog &&
	flux cancel $jobid

true
false
not ok 12 - flux job info applies resource updates
#
#		jobid=$(flux submit --wait-event=start sleep 300) &&
#		echo $jobid > updated_R.id &&
#		flux job info $jobid R | jq -e ".execution.expiration == 0.0" &&
#		flux update $jobid duration=300 &&
#		flux job info $jobid R | jq -e ".execution.expiration > 0.0" &&
#		flux job info $jobid eventlog &&
#		flux cancel $jobid
#
```

grondo: Maybe another PR that fixes both of these.

grondo:
Thanks! I'll set MWP here.

t/t2233-job-info-update.t
chu11: `flux update` appears to be inherently racy, so maybe an audit is generally necessary.  I'll write up an issue.

Issue #7443, opened 2026-03-12
`t1004-kvs-namespace.t` reports leaks with `FLUX_TEST_VALGRIND`


grondo:
This one was not an obvious fix to me, so opening an issue for future reivew.

```console
$ FLUX_TEST_VALGRIND=t ./t1004-kvs-namespace.t -d -v
[snip]
==1775854== 239 (144 direct, 95 indirect) bytes in 1 blocks are definitely lost in loss record 8 of 14
==1775854==    at 0x491C560: calloc (vg_replace_malloc.c:1675)
==1775854==    by 0xAC9A7C7: lookup_create (lookup.c:532)
==1775854==    by 0xAC9628F: lookup_common.constprop.0 (kvs.c:1286)
==1775854==    by 0xAC97633: lookup_request_cb (kvs.c:1393)
==1775854==    by 0x4974D2B: call_handler (msg_handler.c:321)
==1775854==    by 0x49750A7: dispatch_message (msg_handler.c:357)
==1775854==    by 0x49750A7: handle_cb (msg_handler.c:450)
==1775854==    by 0x49B42B3: ev_invoke_pending (ev.c:3770)
==1775854==    by 0x49B7B1B: ev_run (ev.c:4190)
==1775854==    by 0x497324F: flux_reactor_run (reactor.c:91)
==1775854==    by 0xAC98703: mod_main (kvs.c:2873)
==1775854==    by 0x4012983: module_thread (module_thread.c:142)
==1775854==    by 0x4C1B9F3: start_thread (pthread_create.c:448)
==1775854==
==1775854== LEAK SUMMARY:
==1775854==    definitely lost: 144 bytes in 1 blocks
==1775854==    indirectly lost: 95 bytes in 4 blocks
==1775854==      possibly lost: 0 bytes in 0 blocks
==1775854==    still reachable: 22,009 bytes in 58 blocks
==1775854==         suppressed: 0 bytes in 0 blocks
```

chu11:
ugh, the corner case is you do a kvs lookup on a namespace that has been removed.  So the error path goes down the "namespace root" lookup path, vs the "value lookup" path.   Doesn't destroy something that is managed by the "value lookup path".

Should be extremely rare in production use cases, but not in this file's testing.   Lemme think of best way to refactor things.

Issue #7442, opened 2026-03-12
testsuite: "flux update" can be racy in tests


chu11:
Per some discussion in https://github.com/flux-framework/flux-core/pull/7441 several tests appear to be racy due to `flux update`.  It can be inherently racy b/c updates do not "propogate" immediately (there is event batch in `job-manager`, that batch then writes to the eventlog per period, then any eventlog watchers have to wait for the watcher to receive the update, etc.)

Minimally there appears to be raciness in

t2233 - `job-info: update watch works with jobspec` 
t2220 - `flux job info applies resource updates`

although there might be others out there.

chu11:
Generally stuff that uses `flux update` and then checks if the update has propagated needs to have some raciness protected with `test_wait_until()` or perhaps `flux job wait-event` (or similar).  In some cases other things protect against it, as an example this test in t2291-job-update-queue

```
test_expect_success 'updates are reenabled' '                                                                                               
        flux update $jobid queue=batch &&                                                                                                   
        flux cancel $jobid &&                                                                                                               
        flux job wait-event $jobid clean &&                                                                                                 
        flux job eventlog $jobid \                                                                                                          
          | grep jobspec-update \                                                                                                           
          | grep attributes.system.queue=\"batch\"                                                                                          
'                                         
```

The wait for the job to become inactive from the cancel ensures the previous update has propagated.  But a comment might be nice here to explain that.

Others found via analysis (along w/ some old comments):

t2230 - 'flux job info applies jobspec updates'
t2260 - `# N.B. in future, may wish to do test via a `flux update` or similar tool`
t2230 - `# N.B. In future may wish to update expiration via `flux update` tool`
t2291 - `update to batch queue works`, `update of duration allows queue update`, `update from queue with no constraints works`, `update to queue with no constraints works`, 
t2290 - `update of duration of pending job works`, `instance owner can adjust duration past limits`, `guest update of their own job works`, `instance owner can adjust duration past limits`, `instance owner can adjust guest job duration past limits`, `instance owner can adjust guest job duration past limits`, `adjust duration so future tests pass validation`, `now update of attributes.system.test works`, `multiple keys can be updated successfully`

side note: I believe all tests in `t2292-job-update-running.t` are safe b/c they get data from the job-manager or use wait-event.

grondo:
[...]

PR #7445, opened 2026-03-12
fix resource leaks and flaky tests found when running the testsuite under valgrind

grondo:
I ran all sharness tests under valgrind with
```console
$ flux bulksubmit -n1 -c5 --watch --progress --env=FLUX_TEST_VALGRIND=t -o pty ./{} -d -v ::: t[0-9]*.t
```
There were about 13 or 14 test failures on the first run .

This PR is a set of fixes that include both test reliability updates and actual memory leak fixes.

There is one remaining failure described in #7443, since it was the only one that wasn't immediately obvious.
Fixes for #7442 are also left for after this PR.

### DIFF
--- a/src/common/libjob/count.c
+++ b/src/common/libjob/count.c
@@ -112,14 +112,14 @@ struct count *count_create (json_t *json_count,
     count->isrange = true;
     return count;
 error:
-    free (count);
+    count_destroy (count);
     return NULL;
 }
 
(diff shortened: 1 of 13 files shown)


garlick (approved):
Woah, awesome, many good finds here!

grondo:
Thanks! Setting MWP.

PR #7446, opened 2026-03-13
flux-update: add `--wait` option and use it as necessary in testsuite

grondo:
This PR adds a new `--wait` option to `flux update` as proposed in #7442 and uses it in the testsuite where necessary where the job eventlog is checked immediately after the update.

### DIFF
--- a/doc/man1/flux-update.rst
+++ b/doc/man1/flux-update.rst
@@ -68,6 +68,10 @@ OPTIONS
 
   Print updated keys on success.
 
+.. option:: --wait
+
+  Wait until associated updates are posted to the target job's eventlog.
+
 SPECIAL KEYS
 ============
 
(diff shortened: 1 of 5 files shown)


garlick:
PR title and description refer to `--watch` but looks like the option is `--wait`?

grondo:
Yeah, think-o sorry! Will fix.

garlick (approved):
LGTM!

grondo:
Fixed conflicts and will set MWP. Thanks!

chu11:
Just a comment for documentation purposes.  I was unsure if the `--wait` option would be safe for tests that update job duration but then check for job expiration time (i.e. `resource-event` update to expiration written by the job-manager after the `jobspec-update` event updates job duration).

I went spelunking, the job-manager writes both events to a KVS txn before returning to the caller.  So both events should be committed to the KVS in the some transaction.  So it should be safe for this particular use case (with atleast the current job-manager implementation).

PR #7447, opened 2026-03-13
python: improve broker module error propagation and test integration

garlick:
This adds a couple of missing things from the python broker module infrastructure: support for the `flux module debug` stuff, and better propagation of OSError exceptions.

PR #7451, opened 2026-03-14
python: add toy scheduler framework

garlick:
Problem: it's tricky to try new R designs or experiment with scheduler-adjacent changes in flux-core, since `sched-simple` relies on C code that is laborious to re-engineer.

Acknowledging that flux-sched is the production scheduler code base, and that what's in flux-core is essentially a test vehicle, adopt a "don't worry about performance" attitude and lay groundwork for more agile development in python for this area, and at the same time remedy some long ignored deficiencies in flux-core's native scheduling capability. 
 
Implement a `Scheduler` subclass of  `BrokerMod` that implements the scaffolding for interfacing with the job manager and resource module common to all schedulers.  A scheduler implementation would be a subclass of `scheduler`.

To facilitate experiments with an Rv1 replacement, abstract the resource representation in a `ResourcePool` class that does resource selection, feasibility checking, and other operations for an abstract resource representation.  Two Rv1 implementations are provided, one based on C `librlist`, and one in pure python.  The pure python one adds GPU support.  

To demonstrate the framework, two schedulers are provided:
- `sched-fifo`, a priorty-FIFO scheduler like `sched-simple`, except it uses the pure python Rv1 class so it supports GPUs.  It also gives all jobs start time estimates (when duration is set), using a forward-simulation method with a throw-away copy of the resource pool object.
- `sched-backfill`, an EASY backfill scheduler, which is one step beyond FIFO in planning (literally - it plans _one_ job).

I invested some time experimenting with `ResourcePool` and `Scheduler` to make sure they were not completely naive: implementing the `sched-simple` scheduler in python (forthcoming in a future PR) with clean test results, testing throughput to work through first order performance issues, and playing with a backfill planning and forward-simulation for time estimates.  Those experiments brought some clarity to the design, so hopefully they are a bit more evolved than they would be otherwise.

A mundane question: I needed a way to easily select the scheduler on the `flux-start` command line when testing, so I just mimiced the `content-backing.module` broker attribute logic currently in rc1.py with a `sched.module` attribute, which lets you `flux start -Ssched.module=sched-backfill`.  Is there a better/existing way to do that?

This is currently based on
- #7447

### DIFF
--- a/doc/python/scheduler.rst
+++ b/doc/python/scheduler.rst
@@ -1,0 +1,598 @@
+.. _python_scheduler:
+
+.. currentmodule:: flux.scheduler
+
+Writing a Python Scheduler
+==========================
+
+Flux supports scheduler broker modules written in Python.  The
+:class:`Scheduler` base class handles all of the RFC 27
+protocol scaffolding — service registration, resource acquisition, the
+hello/ready handshake with job-manager — so that a Python scheduler only
+needs to implement scheduling policy.
+
+This guide walks through writing a working scheduler from scratch,
+explains each override point, and covers more advanced topics such as
+job annotations and start-time forecasting.
+
+
+Minimal example
+---------------
+
+The base class maintains a :class:`PendingJob` heap queue
+(:attr:`~Scheduler._queue`) and provides default
+implementations of :meth:`~Scheduler.alloc`,
+:meth:`~Scheduler.free`,
+:meth:`~Scheduler.cancel`, and
+:meth:`~Scheduler.prioritize`.  A minimal scheduler only
+needs to override :meth:`~Scheduler.schedule`:
+
+.. code-block:: python
+
+   import heapq
+   from flux.resource import InsufficientResources, InfeasibleRequest
+   from flux.scheduler import Scheduler
+
+   class SimpleScheduler(Scheduler):
+
+       def schedule(self):
+           while self._queue:
+               job = self._queue[0]
+               try:
+                   alloc = self.resources.alloc(job.jobid, job.resource_request)
+               except InsufficientResources:
+                   break   # not enough resources — wait for free
+               except InfeasibleRequest as exc:
+                   job.request.deny(str(exc))
+               else:
+                   job.request.success(alloc)
+               heapq.heappop(self._queue)
+
+   def mod_main(h, *args):
+       SimpleScheduler(h, *args).run()
+
+Save this as :file:`my-sched.py` and load it:
+
+.. code-block:: console
+
+   $ flux module unload sched-simple
+   $ flux module load my-sched.py
+
+The :func:`mod_main` entry point is the same convention used by all Python
+broker modules (see :ref:`python_broker_modules`).
+
+
+The resource pool
+-----------------
+
+:attr:`~Scheduler.resources` is an instance of
+:class:`~flux.resource.Rv1Pool` that represents the entire managed resource
+set.  It tracks which resources are up, down, and currently allocated.
+The pool implementation is selected automatically based on the R version
+field in the resource data (currently only version 1 is supported).
+
+The resource request
+~~~~~~~~~~~~~~~~~~~~
+
+Each :class:`PendingJob` carries a pool-specific resource request object in
+``job.resource_request``, parsed once from the jobspec when the alloc request
+arrives by :meth:`~flux.resource.ResourcePool.parse_resource_request`.  The
+type and fields of this object are defined by the pool; scheduler policy code
+treats it as opaque and passes it directly to
+:meth:`~flux.resource.ResourcePool.alloc`.
+
+If you need the raw jobspec dict — for instance, to inspect job attributes in
+an :meth:`~Scheduler.alloc` override — it is passed as the *jobspec* parameter
+to :meth:`~Scheduler.alloc`:
+
+.. code-block:: python
+
+   def alloc(self, request, jobid, priority, userid, t_submit, jobspec):
+       # jobspec is the parsed dict; inspect it before calling super()
+       super().alloc(request, jobid, priority, userid, t_submit, jobspec)
+
+Allocating resources
+~~~~~~~~~~~~~~~~~~~~
+
+Pass the jobid and resource request to :meth:`~flux.resource.ResourcePool.alloc`:
+
+.. code-block:: python
+
+   try:
+       alloc = self.resources.alloc(job.jobid, job.resource_request)
+   except InsufficientResources:
+       # Not enough resources now — queue the request and retry later
+       ...
+   except OSError as exc:
+       # Request can never be satisfied (InfeasibleRequest or other error)
+       request.deny(str(exc))
+       return
+
+RFC 31 constraint expressions (``constraint`` field on the resource request)
+are carried through automatically; the resource pool evaluates them when
+selecting candidate nodes.
+
+On success, :meth:`~flux.resource.ResourcePool.alloc` returns a resource
+pool object containing only the allocated resources.  Pass it directly to
+:meth:`~AllocRequest.success`, which converts it automatically:
+
+.. code-block:: python
+
+   request.success(alloc)
+
+The pool also records the job's expected expiration (derived from ``request.duration``)
+internally; :meth:`~flux.resource.ResourcePool.free` and start-time
+simulation (via :meth:`~flux.resource.ResourcePool.copy` and
+:meth:`~flux.resource.ResourcePool.job_end_times`) use this state.
+
+Releasing resources
+~~~~~~~~~~~~~~~~~~~
+
+The base class :meth:`~Scheduler.free` calls
+:meth:`~flux.resource.ResourcePool.free` automatically — no override is
+needed unless the scheduler has additional per-job cleanup to do.
+
+A job's resources may be released in multiple calls when housekeeping is
+configured: housekeeping returns resources in batches of one or more ranks as
+they become ready, with *final* set to ``False`` until the last batch.
+Schedulers that override :meth:`~Scheduler.free` must handle this: *R*
+contains only the subset being freed on each call, and per-job state should
+not be discarded until *final* is ``True``.
+
+Resource state changes
+~~~~~~~~~~~~~~~~~~~~~~
+
+Node drain/undrain and other resource state changes arrive automatically
+via the ``resource.acquire`` streaming RPC.  After each update, the base
+class calls :meth:`~Scheduler.resource_update`, giving the
+scheduler a chance to retry pending allocations:
+
+.. code-block:: python
+
+   def resource_update(self):
+       self._retry_pending()
+
+
+Alloc requests
+--------------
+
+When job-manager asks the scheduler to allocate resources for a job, two
+distinct objects are created and stored together on the
+:class:`PendingJob`:
+
+- ``job.resource_request`` — a pool-specific object describing *what the job
+  needs*.  Created once from jobspec by
+  :meth:`~flux.resource.ResourcePool.parse_resource_request` when the alloc
+  arrives.  For the built-in pools this carries fields for node count, slot count,
+  cores and GPUs per slot, duration, constraints, and exclusivity.
+  Used by the scheduler during each :meth:`~Scheduler.schedule`
+  pass to decide whether resources can be satisfied.  Persists for the
+  lifetime of the pending job.
+
+- :class:`AllocRequest` (``job.request``) — represents the
+  *open RFC 27 protocol transaction* with job-manager.  Wraps the message
+  handle and provides the response methods that finalize the exchange.  Must
+  be finalized exactly once; consumed when the scheduler responds.
+
+The scheduler reads :attr:`~PendingJob.resource_request` to
+make scheduling decisions, then calls a method on
+:attr:`~PendingJob.request` to report the outcome.
+
+:class:`AllocRequest` must be finalized exactly once by
+calling one of:
+
+:meth:`request.success(R, annotations=None) <AllocRequest.success>`
+    Allocation succeeded.  *R* is the pool object returned by
+    :meth:`~flux.resource.ResourcePool.alloc`; it is converted to a dict
+    automatically.  The base class commits R to the KVS asynchronously
+    before sending the response to job-manager, ensuring R is safely
+    stored before the job can run.
+
+:meth:`request.deny(note=None) <AllocRequest.deny>`
+    Permanent denial — the job will never run.  *note* is a human-readable
+    reason that appears in :man1:`flux-job` eventlog.  Use this for
+    structurally unsatisfiable requests (too many cores, unsupported resource
+    type, etc.).
+
+:meth:`request.cancel() <AllocRequest.cancel>`
+    The alloc request is being withdrawn.  Call this from
+    :meth:`~Scheduler.cancel`.
+
+    .. note::
+
+       Cancelling an alloc request is **not** the same as cancelling the job.
+       Job-manager may withdraw an alloc request — for example, to shrink the
+       outstanding request count back within the configured ``queue-depth`` —
+       while leaving the job itself in the pending queue.  The job will receive
+       a fresh alloc request when a slot opens up.
+
+:meth:`request.annotate(annotations) <AllocRequest.annotate>`
+    Send an intermediate annotation update while the job is pending.  May be
+    called any number of times before finalizing.  Annotations are visible
+    in :man1:`flux-jobs` output.
+
+    RFC 27 defines two standard ``sched`` annotation keys:
+
+    resource_summary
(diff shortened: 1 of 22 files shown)


grondo:
> one based on C `librlist`, and one in pure python. The pure python one adds GPU support.

To be fair, librlist does support GPUs, if it didn't none of the tools that use it could support GPUs.

Is the intent here to replace librlist as well? That may require wrapping Python code in a C API which I don't have much experience with but I gather is possible.

grondo:
> A mundane question: I needed a way to easily select the scheduler on the flux-start command line when testing, so I just mimiced the content-backing.module broker attribute logic currently in rc1.py with a sched.module attribute, which lets you flux start -Ssched.module=sched-backfill. Is there a better/existing way to do that?

The modprobe alternatives system (which actually you had [originally suggested](https://github.com/flux-framework/flux-core/pull/6774#issuecomment-2894844344)) should be used here. E.g.:
```console
$ flux alloc --conf=modules.alternatives.sched=sched-backfill OPTIONS...
```

Unfortunately there's not a simple way to force an alternative on the `flux start` command line since it doesn't have a `--conf` option like `flux alloc` and `flux batch`. I think initially we had considered a way to do this via broker attributes, but it didn't feel right at the time. Now that there may be many possible alternative schedulers we should consider the easiest way to do this both in `flux start` for testing and with `flux alloc` and `flux batch`.

There is a bug or missing feature here in modprobe though: If I use the invocation above, modprobe still loads `sched-simple` because the feasibility alternative is not set, so you actually have to do this:
```console
$ flux alloc --conf=modules.alternatives.sched=sched-backfill --conf=modules.alternatives.feasibility=sched-backfill
```
Which isn't great.

That should probably be fixed. (When choosing an alternative for one `provides` it should be applied to any other provides for that module as well) I'll actually open an issue on that.

garlick:
> To be fair, librlist does support GPUs, if it didn't none of the tools that use it could support GPUs.

Sorry, I should have said `rlist_alloc()` doesn't support GPUs.

> Is the intent here to replace librlist as well? That may require wrapping Python code in a C API which I don't have much experience with but I gather is possible.

I am not proposing to drop anything here.  The next PR I was going to queue up is a `sched-simple` drop-in replacement that uses `librlist` via the `Rv1RlistPool` class.

I suppose if we wanted to pare back `librlist` so that it will be easier to graft in Rv2, a subsequent PR could switch `sched-simple` to `Rv1Poool` from  `Rv1RlistPool`, then drop both `Rv1RlistPool` and `rlist_alloc()`. There will likely be a lot of overlap from Rv1 to Rv2 like properties, execution targets, etc., so much of `librlist` internals ought to be reusable.

I think it's something we need to discuss and think about.  I don't have any great ideas tonight.  Right now `Rv1Pool` and `Rv1RlistPool` are just demonstrating that these little schedulers can be oblivious to the underlying pool implementation.  Going forward there is not much reason to have two Rv1 implementations.  We could also go the other way and get rid of `Rv1Pool` and start tooling up `librlist` to support Rv2, although `alloc` seems more daunting in C.

grondo:
> I think it's something we need to discuss and think about. I don't have any great ideas tonight. Right now Rv1Pool and Rv1RlistPool are just demonstrating that these little schedulers can be oblivious to the underlying pool implementation. Going forward there is not much reason to have two Rv1 implementations. We could also go the other way and get rid of Rv1Pool and start tooling up librlist to support Rv2, although alloc seems more daunting in C.

Well the huge usefulness of Python-based schedulers is ease of implementing features that were too difficult in the existing sched-simple, outside of the resource representation, e.g. upcoming items like reservations, malleable job requests, etc. Having the ability to do this easily in flux-core is pretty exciting, because that was always a bit of a roadblock to marching forward. This should go in asap as far as I'm concerned.

It hurts my brain a bit to think about how this will work with Rv2 since all the C code has to use librlist currently. It always seemed like the best approach there would be to implement a new librv2 library that had the necessary interfaces to replace librlist (except the allocation interfaces), and then wrap that in Python in a class that conforms to the existing ResourceSetImplementation class. Maybe that really has nothing to do with this PR which is just getting Python scheduler infrastructure going. (On that note, since the Resource*Pool classes are really for schedulers, maybe they belong under a `flux.scheduler` namespace and not in `flux.resource`?)

grondo:
One last thought for the night. If the Python  Rv1 class here could be used as a ResourceSetImplementation provider, then I think we could drop the Rlist class and librlist ffi bindings. This would fix an issue we have in the flux-python package that makes flux.resource unavailable for pip installed bindings.

garlick:
> If the Python Rv1 class here could be used as a ResourceSetImplementation provider, then I think we could drop the Rlist class and librlist ffi bindings. This would fix an issue we have in the flux-python package that makes flux.resource unavailable for pip installed bindings.

Since the python Rv1 class (`Rv1Pool`) was patterned after what was there before, it already implements a large subset of the methods needed for a `ResourceSetImplementation`.  Adding the missing ones would make it a drop-in.

A thought would be to subclass `ResourcePool` from `ResourceSet`, adding the scheduler-specific bits.  Or rather subclass `Rv1Pool` from `Rv1Set` (what is now `Rlist`).

Anyway, then we would have a good architecture for moving the python parts of flux-core forward to Rv2, and if we decided that `rlist_alloc()` and related interfaces could be removed from the C interface (by switching to the python `sched-simple` using `Rv1Pool`), then the problem of moving the C bits forward to Rv2 would get a little bit easier.

grondo:
I think that sounds good. Another thought is that the separate implementation that wraps `Rlist` is perhaps not buying much here, in fact it seemed to add some work to include missing interfaces, extra ffi wrappers etc. Though showing that a scheduler could work with either class is a good demo, and probably helped inform the design of the base class, it probably isn't buying that much in terms of testing after the fact.

So I wonder if a good path forward would be to drop the `Rlist` based "pool" class and associated testing etc. Replace `Rlist` with `Rv1Pool` implementation for the `ResourceSet` class implementation and ensure all the Python-side utilities still work, then finally delete `Rlist` and the librlist bindings.

garlick:
> Another thought is that the separate implementation that wraps Rlist is perhaps not buying much here

Agreed.

> So I wonder if a good path forward would be to drop the Rlist based "pool" class and associated testing etc. Replace Rlist with Rv1Pool implementation for the ResourceSet class implementation and ensure all the Python-side utilities still work, then finally delete Rlist and the librlist bindings.

Yes, that makes sense. 

I was kind of thinking that ResourcePool should be a subclass of ResourceSet that adds the scheduler-specific bits.  Although there would be a fairly tight coupling between Rv1Set and Rv1Pool, keeping the scheduling bits out of the ResourceSet base class probably makes it more comprehensible as an interface.  Thoughts?

garlick:
Also, I think your approach in ResourceSet of making it version agnostic makes sense for schedulers too, in contrast to the way I did it, letting the scheduler select the version specific ResourcePool implementation.  If we changed ResourcePool to be version agnostic then schedulers would just work with either v1 or v2 and not require any re-configuration or peeking into the R object repeated in all schedulers.

grondo:
> I was kind of thinking that ResourcePool should be a subclass of ResourceSet that adds the scheduler-specific bits. Although there would be a fairly tight coupling between Rv1Set and Rv1Pool, keeping the scheduling bits out of the ResourceSet base class probably makes it more comprehensible as an interface. Thoughts?

Sounds good to me! The purpose of `ResourceSet` was a class meant for utilities and to expose to users to work with resources returned from other APIs (like resource_list etc). So keeping scheduler-specific bits out seems right.

I'm not sure if you had envisioned other uses of the `ResourcePool` class except as an allocation interface for schedulers. If not, then maybe `SchedResourcePool` or similar would be a good name to avoid user confusion on which they should import. (I'm not exactly sure what makes it a pool vs a set though because I haven't  digested this PR. You may have a good reason already for a top-level pool type!)

garlick:
Pool implied allocation to me.  You get a car from the motor pool, a secretary from the typing pool, etc.  However if it's not clear I am not strongly attached to that name and we can prepend Sched!

If it's ok let's get the class hierarchy worked out first and then change the name later as I'm pretty deep into it right now. That will be a simple change

grondo:
> Pool implied allocation to me. You get a car from the motor pool, a secretary from the typing pool, etc. However if it's not clear I am not strongly attached to that name and we can prepend Sched!

Ah, that makes sense! For some reason my brain didn't go there. I was thinking of "pool" as an indistinct grouping of resources vs a "set" as distinct. However, once it clicks it makes sense. (to add to your examples, thread pool, object pool)

garlick:
Rebased on current master, and made `ResourcePool` a subclass of `ResourceSet`, as discussed.  The pool class implementation is now selected by version at instantiation (same pattern as `ResourceSet`) so the `resource-pool` scheduler module option went away.

I added a `log()` method to the scheduler class and a `log-level=LEVEL` option common to all schedulers, so that log filtering could occur at the scheduler level.  This method is also now passed into the pool class so it can optionally log stuff too.

I also added a (currently unused) hook in the pool class so arbitrary kwargs can be passed in by a scheduler at instantiation, in case that proves helpful later on.

grondo:
> I added a log() method to the scheduler class and a log-level=LEVEL option common to all schedulers, so that log filtering could occur at the scheduler level. This method is also now passed into the pool class so it can optionally log stuff too.

Great! I was about to suggest that.

garlick:
p.s. this doc might be helpful to skim as part of a review:

https://flux-framework--7451.org.readthedocs.build/projects/flux-core/en/7451/python/scheduler.html

src/bindings/python/flux/job/kill.py
```diff
@@ -72,3 +72,46 @@ def cancel(flux_handle: Flux, jobid: Union[JobID, int], reason: Optional[str] =
...
+):
+    """Raise a job exception asynchronously.
+
+    :param flux_handle: handle for Flux broker from flux.Flux()
+    :param jobid: the job ID of the job
+    :param exc_type: exception type string (e.g. ``"scheduler-restart"``)
+    :param severity: severity level; 0 causes the job to abort
+    :param note: optional human-readable message
+    :returns: a Future
+    :rtype: Future
```
grondo: Suggestion: I think we're trying to standardize on [Google style Python docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html). I know you're just copying the existing docstrings in this file. We could always do a bulk update later.

src/bindings/python/flux/resource/ResourceSet.py
```diff
@@ -275,6 +275,10 @@ def rlist(self):
     def properties(self):
         return ",".join(json.loads(self.get_properties()).keys())
 
+    def to_dict(self):
+        """Return the resource set as a parsed R JSON dict."""
+        return self.impl.to_dict()
+
```
grondo, line 281: Good addition, but this probably belongs in a separate commit, needs unit tests, etc?

src/bindings/python/flux/resource/__init__.py
```diff
@@ -13,6 +13,7 @@
 from flux.resource.list import resource_list, SchedResourceList
 from flux.resource.status import resource_status, ResourceStatus
 from flux.resource.journal import ResourceJournalConsumer
+from flux.resource.Rv1Pool import Rv1Pool
```
grondo, line 16: Commit message looks like it has an editing mistake "that tracks adds resource allocation and up/down trakcing" (remove 'tracks')?

src/bindings/python/flux/scheduler.py
```diff
@@ -0,0 +1,1075 @@
...
+    # Adaptive scheduling timer
+    # ------------------------------------------------------------------
+
+    def _request_schedule(self):
+        """Request a scheduling pass, coalescing bursts via an adaptive timer.
+
+        Records the time of the request and updates the inter-request interval
+        EWMA.  Arms the one-shot scheduling timer if it is not already pending;
+        subsequent calls while the timer is armed are recorded for EWMA
+        purposes but do not re-arm the timer, coalescing all requests in the
```
grondo: EWMA is used throughout but not defined.

src/bindings/python/flux/scheduler.py
```diff
@@ -0,0 +1,1075 @@
...
+                "id": msg.payload["id"],
+                "type": _ALLOC_ANNOTATE,
+                "annotations": annotations,
+            },
+        )
+
+    # ------------------------------------------------------------------
+    # Override points
+    # ------------------------------------------------------------------
+
```
grondo, line 610: This big comment implies override points follow, but there's at least a few above here like `forecast` and `schedule`

garlick:
I did find a few issues while trying to rebase my `sched-simple` PR onto this one.  A notable one is that the pool's free() method doesn't take a `final` flag so it breaks partial release.  Will post some fixes tomorrow.

doc/python/scheduler.rst
```diff
@@ -0,0 +1,801 @@
...
+On success, :meth:`~flux.resource.ResourcePool.alloc` returns a resource
+pool object containing only the allocated resources.  Call
+:meth:`~flux.resource.ResourcePool.to_dict` to convert it to the dict
+that :meth:`~AllocRequest.success` expects:
+
+.. code-block:: python
+
+   request.success(alloc.to_dict())
+
+The pool also records the job's end time (derived from ``request.duration``)
```
grondo: "end time" -> "expiration"? or perhaps just expected end time?

doc/python/scheduler.rst
```diff
@@ -0,0 +1,801 @@
...
+
+:class:`AllocRequest` must be finalized exactly once by
+calling one of:
+
+:meth:`request.success(R, annotations=None) <AllocRequest.success>`
+    Allocation succeeded.  *R* must be a dict (use
+    :meth:`~flux.resource.ResourcePool.to_dict` on the alloc result).
+    The base class commits R to the KVS asynchronously before sending the
+    response to job-manager, ensuring R is safely stored before the job
+    can run.
```
grondo: If all instances of a `ResourcePool` have a `to_dict()` method, was there a reason why `request.success()` couldn't just take a `ResourcePool` object and call `to_dict()` itself? This would simplify the interface. (`request.success` could even use `isinstance(R, ResourcePool)` or similar to check if `to_dict()` needs to be called)

grondo (approved):
This came out great! What a nice result.

I really only had some trivial comments that you can feel free to ignore.

garlick:
Thanks for those suggestions!  I'll go through them and address in another update.  Meanwhile I made some changes after discovering the partial release was broken and some other more minor stuff:
- fix `Rv1Pool.free()` to  correctly handle partial releases and add unit tests                 
- add logging directly to the pool and drop scheduler overrides that were only there for logging
- ensure all scheduler logging is subject to the local severity filter
- improve module option parsing so `flux module load` gets helpful error messages on failure
- tighten up the scheduler narrative doc page

grondo:
There's also some missing test coverage (feasibility check, expiration update), but I figured those would get testing when sched-simple was replaced as the default.

That does make me wonder though if we need a generic "scheduler conformance" test that iterates through schedulers and ensures they conform to some basic expectations.

The other thing that might be nice is a benchmarking scheduler that tries to be as fast as possible so that scheduling is not the bottleneck in throughput tests, which might allow us to more easily find performance improvements elsewhere.

doc/python/scheduler.rst
garlick: Great point!

garlick:
Just pushed changes for your review comments @grondo - very helpful, thanks.

Good thoughts on the increased coverage and the scheduler compliance test.  Let me ponder that.  For now I'll return to the sched-simple PR and see if I shake out any more problems over there before this is merged.

garlick:
Just pushed some updates after getting to a clean flux-core test run with the python `sched-simple` (elsewhere).  It turns out quite a lot of our testing of the resource set and pool stuff was covered by system sharness testing, not unit tests, so this  was productive - much more so than what I had done before, which was base `sched-simple.py` upon `Rlist` / `Rv1RlistPool`.

There are some fixes and tests for  `Rv1Set` with a commit explaining the problems.
The `Rv1Pool` and `Scheduler` fixes are just squashed here.

garlick:
Any objections if I set MWP here?

grondo:
I say go for it!

trws:
I just finally got time to actually read this over. This is a massively more complicated scheduler setup than I expected from the "toy" designation. It's a very nice addition, especially as a way to let people experiment with new custom schedulers.  I'm a bit tempted to make a few adjustments to this to make it work more the way qmanager does from a progress and behavior perspective, it might actually not be all that much of a lift to _replace_ qmanager with a reworked version of the backfill class here.  We'd have to do some performance tests and similar, but I'm sorely tempted to give it a shot, especially given some of the features we've been discussing for that recently the flexibility would be nice, and it isn't a compute-bound service either.  Do you think we could find some time to chat over this next week @garlick?

PR #7454, opened 2026-03-14
etc: update bash completions and add testing

grondo:
This PR updates the Flux bash completions with missing commands and options.

It then adds a framework for testing the bash completions, including a check that attempts to check for missing commands (that are not in a specific denylist).


garlick (approved):
Cool, LGTM!

PR #7455, opened 2026-03-14
modprobe: propagate alternatives selection to all module `provides`

grondo:
Problem: When selecting an alternative, e.g. `modules.alternatives.sched=sched-fifo`, modprobe does not propagate the selected module to other services  in the module `provides`. This can lead to an attempt to load a conflicting module and unexpected failure in modprobe and rc1.

This PR propagates the selected alternative to all other services in `provides` to fix the issue.


garlick (approved):
LGTM!

PR #7458, opened 2026-03-16
Support `--conf` option in `flux start` and `flux broker`

grondo:
This PR adds a new `--conf` option to the Flux broker. The option improves upon the existing `-c, --config-path` option by allowing multiple use as well as dotted `KEY=VAL` syntax such as `tbon.tcp_user_timeout=5m` when an `=` appears in the option value, otherwise it falls through to parsing as inline JSON if a `{` appears at the beginning of the option argument, otherwise TOML if a newline appears, and finally loads as a config file (JSON if the filename extension is `.json`, TOML otherwise)

This somewhat mirrors what's supported in `flux batch` and `flux alloc`, except for "named configuration" support in those utilities.

`flux start` also gets a `--conf` option, but this is just passed along to the broker.

This option made a lot more sense to add to the broker, since the conf updates can be applied in memory after loading any file specified by `-c, --config-path` or `FLUX_CONF_DIR`, no temporary files required.

### DIFF
--- a/doc/man1/flux-broker.rst
+++ b/doc/man1/flux-broker.rst
@@ -56,6 +56,45 @@ OPTIONS
    configuration as JSON if the file extension is ``.json``, otherwise
    load the file as TOML.
 
+.. option:: --conf=VALUE
+
+   Update the broker configuration from VALUE after any configuration
+   file specified with :option:`--config-path` (or :envvar:`FLUX_CONF_DIR`)
+   is loaded. This option may be specified multiple times and values are
+   applied in order. The format of VALUE is determined as follows:
+
+   ``KEY=VAL``
+     If VALUE contains ``=``, set the configuration key at dotted path
+     KEY to VAL. VAL is parsed as JSON; if it is not valid JSON it is
+     treated as a string. Note: file paths containing ``=`` are
+     interpreted as KEY=VAL rather than as file names.
+
+   inline JSON object
+     If VALUE starts with ``{``, it is parsed as an inline JSON object
+     and merged into the configuration.
+
+   inline TOML string
+     If VALUE contains a newline (and does not start with ``{``), it is
+     parsed as an inline TOML string and merged into the configuration.
+     Note: shell command substitution (``$()``) strips trailing newlines,
+     so inline TOML strings should have an embedded newline, as TOML
+     naturally does.
+
+   path ending in ``.json``
+     Parse the file as JSON and merge the result into the configuration.
+
+   any other string
+     Treated as a path to a TOML file.
+
+   .. note::
+
+      When a broker is started with :option:`--conf`, that configuration is
+      in-memory only and is not written to files. If :option:`--config-path`
+      or :envvar:`FLUX_CONF_DIR` is also set, running :command:`flux config
+      reload` within that instance will overwrite any configuration applied
+      via :option:`--conf`, since reload reads from files only. (If no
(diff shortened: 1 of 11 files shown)


grondo:
I have a couple fixes to push here, so hold off any review for the moment.

grondo:
Tacked a fix for #7459 on this PR. Let me know if it is preferred to put that one in a separate PR. (Just fixing a mistake from #7445)

garlick:
Fine with me to leave that fix in this PR.  Ready for a review here?

grondo:
Yes, sorry, this is ready.

garlick (approved):
That worked out nicely!

Just to be clear, when there is both a `--config-path` and `--conf` options, `flux config reload` would lose the `--conf` settings.  That seems fine and is consistent with how `flux config set|unset` work.  All good there correct?

grondo:
> Just to be clear, when there is both a --config-path and --conf options, flux config reload would lose the --conf settings. That seems fine and is consistent with how flux config set|unset work. All good there correct?

Yes, but adding a note in the `flux-broker(1)` docs is probably warranted now that you mention it. I'll add that!

grondo:
I added this note to the `--conf` description in `flux-broker(1)`:

>      When a broker is started with :option:`--conf`, that configuration is
>      in-memory only and is not written to files. Running :command:`flux
>      config reload` within such an instance will overwrite or reset the
>      broker config to empty, since :command:`reload` reads from files only.

garlick:
Wait, is that true?
```
ƒ(s=1,d=0) garlick@system76-pc:~/proj/flux-core$ flux config set -t integer foo 42
ƒ(s=1,d=0) garlick@system76-pc:~/proj/flux-core$ flux config get
{"foo":42}
ƒ(s=1,d=0) garlick@system76-pc:~/proj/flux-core$ flux config reload
ƒ(s=1,d=0) garlick@system76-pc:~/proj/flux-core$ flux config get
{"foo":42}
```

grondo:
Oh good catch! I thought I tested it, but I must have made a mistake.

Since 377353145a7c6432edfdcf3ed3f9357039b8124b, each time the broker loads config it does a recursive merge, but I had thought that `flux config reload` started from an empty config first.

Hm, should we pause and think about the behavior here before merging? Though this behavior was pre-existing, so maybe open an issue?

My worry is that a sysadmin could remove a table in files-only case (e.g. system instance) after `flux config reload` the table won't be removed?

grondo:
Oh, in the config module if there's no config path set the "reload" is a no-op and configuration isn't touched. Sorry, maybe you were alluding to the fact that's expected behavior and it went over my head at first? Since there's no real issue here, I'll update the note to remove the part about an empty config.

grondo:
Ok, I'll update to this:
>     When a broker is started with :option:`--conf`, that configuration is
>     in-memory only and is not written to files. If :option:`--config-path`
>     or :envvar:`FLUX_CONF_DIR` is also set, running :command:`flux config
>     reload` within that instance will overwrite any configuration applied
>     via :option:`--conf`, since reload reads from files only. (If no
>     config path is set, :command:`flux config reload` is a no-op.)

(and sorry it took me a minute to catch up to what you were saying)

garlick:
You got it, sorry I didn't really use my words :-)

grondo:
If the new note sounds good, I'll set MWP here.

PR #7461, opened 2026-03-18
sched-simple: replace with python version

garlick:
This drops the C `sched-simple` FIFO scheduler and replaces it with a python version.  The python version just subclasses `FIFOScheduler` (from `sched-fifo.py`) and adds debugging hooks needed by tests.

Changes:
- verbose alloc/free logging is now suppressed by default.  Tests that need it can enable it (for any of the python schedulers) by reloading the scheduler with `log-level=debug`.
- GPUs are supported.
- Jobs falling within the queue depth (default 8) window are annotated with estimated start times.
- the alloc mode is hardwired to `worst-fit`

Besides this change, this PR also includes the following fixes that didn't make it into #7451:
- avoid an unnecessary feasibility check when alloc succeeds
- fix error propagation when resource acquisition fails
- ensure exclusive allocation is handled properly when `nnodes` is unspecified


grondo:
One comment before I dive in is that this may remove support for resource ranges as added in https://github.com/flux-framework/flux-core/pull/6973. Unfortunately, it does not appear that tests were added to ensure a scheduler does not choke on non-integer ranges in jobspec. 

We may want to ensure we get @sam-maloney's approval before merging this one.

sam-maloney:
Without a ton of time to go through things in detail, my immediate thoughts are:

- I don't want to hold up this PR for a feature that is still very much in development (especially as indicated by the fact I didn't even have any integration tests for it!)
- GPU support is certainly more important, and this is supposed to be a "simple" scheduler, so I would prioritize getting more of Jobspec V1 working rather than experimental features
- `fluxion` also supports ranges for anything other than slots, so this was more of a nice quick side effect of creating `libjjc`, which was anyways the more major contribution from #6973 
- `libjj` is still used in a couple other places that will need to eventually support ranges, so `libjjc` isn't totally without purpose anymore

So feel free to go ahead, and I when I get more time I will have a look to see if support for ranges can easily be added to the new `sched-simple` later on!

grondo:
Thanks @sam-maloney!

> libjj is still used in a couple other places that will need to eventually support ranges, so libjjc isn't totally without purpose anymore

This answers my second question, since `sched-simple` was the only current user of `libjjc`. Maybe we should open an issue to transition other users of `libjj` to the `libjjc` version so we can also get rid of the `libjj` sooner rather than later.

grondo (approved):
Nice! LGTM!

garlick:
Oh, apologies for the regression. I'll go ahead and set MWP here and open an issue on ranges.

Thanks guys!

PR #7468, opened 2026-03-19
ci: pin macos lua version to 5.4

grondo:
Problem: The latest Lua version installed by brew in the macos ci builder is 5.5, but this version is not supported by flux-core or the luaposix module.

Pin Lua to v5.4 for now in macos. This unfortunately requires several steps to get right because brew pulls in 5.5 anyway.


garlick (approved):
LGTM

Issue #7406, opened 2026-02-24
not ok 61 - flux-pmi --method=libpmi2 barrier works w/ flux pmi lib


grondo:
This test failed in the `bookworm - gcc12, distcheck` builder:

```
 expecting success: 
  	test_expect_code 143 flux run -n2 bash -c "\
  	    flux pmi -v --method=libpmi2:$(cat libpmi2) barrier --test-abort 1"
  
  0.148s: flux-shell[0]: FATAL: pmi-simple: PMI-1 wire protocol error
  0.147s: job.exception fLUn1hc3 type=exec severity=0 PMI_Abort: flux-pmi: rank 1 is aborting
  flux-job: task(s) exited with exit code 1
  test_expect_code: command exited with 1, we wanted 143 flux run -n2 bash -c 	    flux pmi -v --method=libpmi2:/usr/src/flux-core-0.82.0-126-g0a31429ad/_build/sub/src/common/flux/.libs/libpmi2.so barrier --test-abort 1
  not ok 61 - flux-pmi --method=libpmi2 barrier works w/ flux pmi lib

```

PR #7470, opened 2026-03-19
scheduler: support RFC 14 range counts

garlick:
Problem: PR #6973 added range count support to the C sched-simple so that jobs with RFC 14 min/max count dicts would not be rejected.  When the C sched-simple was replaced by the Python implementation (PR #7461) this support was regressed: from_jobspec() treated any non-integer count as an error, causing jobs with range counts to be immediately denied.

Restore range support by adding _parse_count() to ResourceRequest to handle integer and dict {"min", "max"} count forms as they arrive from a JSON-parsed jobspec (RFC 14).

Since it was not much extra work, go beyond what the C version did (which only used the minimum value): add nnodes_max and nslots_max fields to ResourceRequest and update alloc() to greedily fill up to the maximum, or to pool capacity when max is absent, while requiring only the minimum to be satisfied.  The feasibility check on failure uses the minimum, so a range request whose max exceeds pool size is still considered feasible.

Raise ValueError for any operator/operand combination that implies different semantics than our current implementation.  The implied default (operator="+", operand=1, meaning any count in [min,max]) is accepted when stated explicitly.  The ValueError propagates to a job denial with a clear "not yet supported" message.

RFC 14 also defines operator and operand fields that constrain which counts within [min, max] are acceptable (addition, multiplication, exponentiation step functions).  RFC 45 defines a compact string representation of range expressions for use in command line interfaces. Both are left for follow-on work.

### DIFF
--- a/src/bindings/python/flux/resource/Rv1Pool.py
+++ b/src/bindings/python/flux/resource/Rv1Pool.py
@@ -45,6 +45,7 @@ from typing import Dict, List, Tuple
 
 from flux.idset import IDset
 from flux.job import JobID
+from flux.resource.ResourceCount import ResourceCount
 from flux.resource.ResourcePoolImplementation import (
     InfeasibleRequest,
     InsufficientResources,
@@ -62,18 +63,29 @@ class ResourceRequest:
     re-parse jobspec on every scheduling pass.
 
     Attributes:
-        nnodes (int): Minimum node count; 0 means any layout.
-        nslots (int): Total slot count.
+        node_count (ResourceCount | None): RFC 14 node count (scaled by
+            nodefactor), or ``None`` for slot-only layouts.  For exclusive
+            slot-only jobspecs this holds the slot count reinterpreted as
+            node count.
+        slot_count (ResourceCount): RFC 14 slot count.  For node-based layouts
+            this is the per-node slot count; for slot-only layouts it is the
+            total slot count; for exclusive slot-only layouts it is
+            ``ResourceCount(1, 1)``.
         slot_size (int): Cores per slot.
         gpu_per_slot (int): GPUs per slot.
         duration (float): Walltime in seconds; 0.0 means unlimited.
         constraint: RFC 31 constraint expression (dict) or None.
         exclusive (bool): Whole-node exclusive allocation.
+        nnodes (int): Minimum node count; derived from *node_count*.
+        nnodes_max (int | None): Maximum node count; ``None`` for unbounded.
+        nslots (int): Minimum total slot count; derived from *slot_count* and
+            *node_count*.
+        nslots_max (int | None): Maximum total slot count.
     """
 
     __slots__ = (
-        "nnodes",
-        "nslots",
+        "node_count",
+        "slot_count",
         "slot_size",
         "gpu_per_slot",
         "duration",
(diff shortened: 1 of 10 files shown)


grondo:
The previous work done here was embedded in `libjjc` (json jobspec parser with counts support). It might be useful to attempt to reuse/reproduce at least some of the tests already written using the `jjc-parser` test utility and found in `t0024-jjc-reader.t`. For example, there is a set of valid and invalid jobspec from the original use cases in `t/jobspec/{valid,invalid}` that should probably be explicitly tested.

garlick:
Well I'm very glad you brought that up because I had regressed @sam-maloney 's work to relax jobspec parsing by the scheduler and store nslots in R to avoid needing to re-parse jobspec in the shell.  His work was well planned and executed, and I am pretty embarrassed to have lost the thread.

More commits forthcoming to remedy the situation.

garlick:
Restarting a build er that hit
- #7406

garlick:
Rebased on current master.

src/bindings/python/flux/resource/Rv1Pool.py
```diff
@@ -52,6 +52,10 @@
 )
 from flux.resource.Rv1Set import Rv1Set
 
+# Sentinel meaning "nnodes_max / nslots_max not given — same as min (fixed count)".
+# Callers that want an unbounded range pass None explicitly.
+_NO_MAX = object()
+
```
grondo: It would be interesting to know why this sentinel is an empty object instead of something like `-1`

garlick:
Sorry, I realized I pushed changes last week and didn't comment on them.  I believe the regressions have been remedied and testing has been added that mimics the libjjc testing, iterating over the test input YAML jobspecs.

garlick:
Restarting a builder that failed on a faketime test (issue referenced above).

src/bindings/python/flux/resource/Rv1Pool.py
```diff
@@ -82,16 +94,55 @@ class ResourceRequest:
...
+        *max_val* equals *min_val* for a fixed count, is greater for a bounded
+        range, or is ``None`` for an unbounded range.
+
+        Supported forms (as received from a JSON-parsed jobspec):
+
+        - Integer: ``4`` → ``(4, 4)``
+        - Range dict: ``{"min": 2, "max": 8}`` → ``(2, 8)``
+        - Unbounded dict: ``{"min": 2}`` → ``(2, None)``
+        """
```
grondo: This is missing support for count as an idset string or RFC 45 resource range string. See `src/common/libjob/test/count.c` for a set of inputs useful in unit testing.

grondo: The C `count` class has kind of a nice interface with `encode()` as well as `decode()` methods and an iterator. I haven't thought long enough to know if this would be useful from Python as well, but it might be worth considering.

grondo (approved):
This is a huge step forward, but the parsing still has some gaps relative to the previous C implementation. I'm not sure it is a real issue at this point, but closing those gaps may require a step back to determine how that affects the scheduler, if the private `_parse_counts()` should be pulled out into its own resource counts class similar to the C interfaces (could be useful for utilities perhaps).

However, it would also seem fine to merge this as is since it does what it says it will do (though perhaps some comments and documentation need to be updated there is only partial support for RFC 14 resource ranges) and there are no extant use cases that this would break.

src/bindings/python/flux/resource/Rv1Pool.py
garlick: OK, I had only implemented range expressions that could currently be scheduled, and omitted the string form.  Now that you mention it, it does make sense to parse the full spec and carry the complex count through the layers and reject the unsupported expressions in alloc/feasibility rather than at parse time.  Then we can instantiate all the parsing tests that Sam had.

In fact, I'm now wondering how hard it would be to fully implement allocation for the full spec in  Rv1Pool.  But later.  Let me do some refactoring.

garlick:
OK, just pushed a refactor with a new `Count` class added to `Rv1Pool` and support for all the complex count representations like libjjc.

garlick:
Per offline suggestion from @grondo, `Count` is renamed to `ResourceCount` and is now a standalone class with its own unit tests.

garlick:
I'll go ahead and set MWP here.  Thanks!

PR #7473, opened 2026-03-19
flux-resource: add queue display filtering and hidden-queues config to simplify output of `flux resource list`

grondo:
This PR makes a couple simple changes to `flux resource list` to make output less cluttered on systems with overlapping queues. There's a couple changes here:

1. When using `flux resource list -q, --queue=`, only display the selected queues in the QUEUE column. Properties for other queues are also stripped so that listing a single queue is much more manageable when there are overlapping queues.
2. Support a new `list.hidden-queues` option in the `flux-resource.toml` config. This contains an array of queue names that should be suppressed in output by default. This is useful for the case of an `all` or `pall` queue that overlaps with every other queue and clutters default `flux resource list` output.

Unfortunately, there's a bit of a version problem now with the `flux-resource.toml` and other utility config files. The config file loader raises a fatal error on unknown keys, so new config items can't be placed in configuration until all versions being used support the key in question. I'm not sure if the fatal error should be changed to a warning, or if there's some other, more practical fix.

### DIFF
--- a/doc/man1/flux-resource.rst
+++ b/doc/man1/flux-resource.rst
@@ -497,7 +497,8 @@ The following field names can be specified for the **list** subcommand:
    Properties associated with resources.
 
 **propertiesx**
-   Properties associated with resources, but with queue names removed.
+   Properties associated with resources, but with all configured queue names
+   removed, regardless of any ``-q, --queue`` filter in effect.
 
 **nnodes**
    number of nodes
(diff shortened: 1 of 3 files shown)


garlick (approved):
This seems great to me.  I'd say let's get it merged and give @kkier and the sys admins a heads up in case they want to give this a spin before we tag the next release.

Maybe the issue with the front end tools config file fragility warrants a separate issue?  (A warning instead of an error seems right but that's not an issue with this PR)

grondo:
Ok, I opened #7480. Unfortunately in this particular case nothing we do here can help the current situation, since deployed Flux versions will all error when adding the new config key.

Issue #7462, opened 2026-03-18
exit-on-error/exit-timeout: task segfault masked by SIGTERM from job teardown


grondo:
## Description

When a job is terminated due to `exit-on-error` or `exit-timeout` after a task fails with a signal, the signal that caused the failure can be masked in the final job status. This happens because remaining tasks are killed with SIGTERM during teardown, and SIGTERM (15) has a higher signal number than many meaningful signals like SIGSEGV (11). Since the shell aggregates task exit codes by taking the numeric maximum, SIGTERM \"wins\" over SIGSEGV and the user sees \"Terminated\" instead of \"Segmentation Fault\".

## Reproduction

```
flux run -N2 -n4 -o exit-on-error \
  sh -c 'if test $FLUX_TASK_RANK -eq 1; then kill -SEGV $$; else sleep 30; fi'
```

Example output:
```
0.060s: flux-shell[0]: FATAL: task-exit: sh: task rank 1 on host ubuntu failed and exit-on-error is set
0.060s: job.exception ƒSuWV8vK type=exec severity=0 sh: task rank 1 on host ubuntu failed and exit-on-error is set
flux-job: task(s) Terminated
```

The job exception message correctly identifies that task rank 1 caused the failure, but the final job exit status reflects SIGTERM (from the cleanup of the other tasks) rather than SIGSEGV (the actual cause).

## Root Cause

The sequence of events:

1. Task rank 1 segfaults. Its exit code is computed as `128 + SIGSEGV(11) = 139`.
2. The `task-exit` plugin (`doom.c`) sees the first task exit with a nonzero code. Because `exit-on-error` is set, it calls `shell_die(139, ...)`, which raises a fatal exec exception and initiates job teardown.
3. The remaining tasks (ranks 0, 2, 3 — still sleeping) are killed with SIGTERM during teardown. Their exit codes are `128 + SIGTERM(15) = 143`.
4. `shell_max_task_exit()` in `src/shell/shell.c` aggregates exit codes with a simple numeric max:
   ```c
   while (task) {
       if (rc < task->rc)   // simple numeric comparison
           rc = task->rc;
       task = zlist_next (shell->tasks);
   }
   ```
5. `max(139, 143, 143, 143) = 143`, so the shell exits 143 and the job reports \"Terminated\" (SIGTERM) instead of \"Segmentation fault\" (SIGSEGV).

[...]

Issue #7472, opened 2026-03-19
`t0014-runlevel.t` hangs in CI


grondo:
The `t0014-runlevel.t` test has started hanging in CI occasionally. This is the last thing printed in the logs:
```
 expecting success: 
  	test_expect_code 1 flux start \
  		-Sbroker.rc3_path=false \
  		-Slog-stderr-level=6 \
  		true 2>rc3_failure.log
```

grondo:
There's definitely something strange going on here. There's a reproducible 10s delay a the end of this test, related to cleaning up all the loaded modules. Adding some debugging, it looks like the _first_ module on which `pthread_join()` is called shows the 10s delay, then the rest of the modules are joined almost instantly:

```
Mar 19 19:37:29.919386 PDT 2026 broker.err[0]: destroying mh->zh_byuuid
Mar 19 19:37:29.919402 PDT 2026 broker.err[0]: pthread_join for module content-sqlite
Mar 19 19:37:39.929490 PDT 2026 broker.err[0]: content-sqlite done
Mar 19 19:37:39.929995 PDT 2026 broker.err[0]: pthread_join for module kvs
Mar 19 19:37:39.930034 PDT 2026 broker.err[0]: kvs done
Mar 19 19:37:39.930154 PDT 2026 broker.err[0]: pthread_join for module content
Mar 19 19:37:39.930184 PDT 2026 broker.err[0]: content done
Mar 19 19:37:39.930434 PDT 2026 broker.err[0]: pthread_join for module heartbeat
Mar 19 19:37:39.930462 PDT 2026 broker.err[0]: heartbeat done
```

This is not specific to `content-sqlite`. A different module is the first to join in each reproducer and there is always a 10s delay.

Now, this isn't reproducing a hang. However, it seems like this test is ensuring that a failure in rc3 causes the instance to exit with a nonzero exit status. Since something in this module teardown process is likely a cause, could we work around the issue by using `-Sbroker.rc1_path=` in the failing test? The 10s delay at exit could be tackled separately, but seems a low priority.

grondo:
I propose we see if adding `-Sbroker.rc1_path=` for this test resolves the hang and move the 10s hang with `pthread_join()` to a different issue.

PR #7474, opened 2026-03-20
avoid masking the exit status of the task that triggered a `exit-timeout` or `exit-on-error` job exception

grondo:
This PR implements support for the optional `wait_status` parameter to `job-manager.raise` added recently to RFC 21, then uses that field, when present, to set the job's `status` in the `finish` event.

A new `flux_shell_raise_wait_status(3)` function is added to the shell API which allows the doom plugin to set the fatal exception `wait_status` for a job to the exit status of the task which actually caused the `exit-on-error` or `exit-timeout` exception.

This fixes the specific case described in #7462 in which a single task segfaults, but the subsequent cleanup after the doom plugin raises an exception masks the segfault and the `finish` event records the exit status of the job as `Terminated` (killed by signal 15) instead of `Segmentation fault` (killed by signal 11).

To make things even more obvious, a string status of the first exited task is added to these exceptions, so we go from:
```
1.043s: job.exception ƒ93ycnLf type=exec severity=0 sh: task rank 1 on host ubuntu exited and exit-timeout=1s has expired
```
to
```
1.047s: job.exception ƒNL6AvRQs type=exec severity=0 sh: task rank 1 on host ubuntu exited (Segmentation fault) and exit-timeout=1s has expired
```

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -181,6 +181,7 @@ MAN3_FILES_SECONDARY = \
 	man3/flux_shell_err.3 \
 	man3/flux_shell_fatal.3 \
 	man3/flux_shell_raise.3 \
+	man3/flux_shell_raise_exit_status.3 \
 	man3/flux_shell_log_setlevel.3 \
 	man3/flux_shell_task_info_unpack.3 \
 	man3/flux_shell_task_cmd.3 \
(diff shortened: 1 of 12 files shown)


grondo:
restarting a builder that hit #7472

garlick (approved):
Great improvement.  LGTM!

grondo:
Thanks! Setting MWP.

PR #7475, opened 2026-03-20
python: add set_proctitle() and use it in python modules, job validator and frobnicator

grondo:
This PR sets the proctitle for python broker modules to the module name so they don't all appear as `python3` in `ps`, `pstree` and `top` output.

e.g.:

```console
flux-broker-0(199187)─┬─bash(199687)───pstree(211512)
                      └─sched-simple(199626)
```

### DIFF
--- a/src/bindings/python/flux/Makefile.am
+++ b/src/bindings/python/flux/Makefile.am
@@ -15,6 +15,7 @@ nobase_fluxpy_PYTHON = \
 	eventlog.py \
 	conf_builtin.py \
 	brokermod.py \
+	proctitle.py \
 	cli/__init__.py \
 	cli/base.py \
 	cli/alloc.py \
--- a/src/bindings/python/flux/proctitle.py
+++ b/src/bindings/python/flux/proctitle.py
@@ -1,0 +1,32 @@
+##############################################################
+# Copyright 2026 Lawrence Livermore National Security, LLC
+# (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+#
+# This file is part of the Flux resource manager framework.
+# For details, see https://github.com/flux-framework.
+#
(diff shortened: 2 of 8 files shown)


grondo:
restarted a builder that hit #7472

garlick:
Yay I was going to do that but hadn't gotten it off the ground.  Thanks!

garlick (approved):
LGTM!

grondo:
Thanks! I've set MWP.

grondo:
Actually, I was just now wondering if `set_proctitle()` should to in `flux.util` or something (or its own module) and we can do the same for the validator and frobnicator processes.

Well, I guess nothing prevents a future PR from doing that, but wondering if we should just do it here? Any thoughts @garlick?

garlick:
I'm good with that!

Edit: either way is fine!

grondo:
Ok, that was a pretty simple job for Claude.

There's now a `flux.proctitle` module which contains the `set_proctitle()` implementation. (Not added to `flux.util` to avoid importing ctypes in everything that uses `flux.util`).

The python module loader uses `set_proctitle()` as before, and now the validator and frobnicator set their process titles to `job-validator` and `job-frobnicator` (drop `flux-` prefix to avoid overflowing the 15 character kernel max proctitle)

```console
$ pstree -Tplu grondo
bash(551568)───flux-broker-0(551617)─┬─bash(551745)───pstree(667182)
                                     ├─job-validator(667120)
                                     └─sched-simple(551730)
```

grondo:
Restarted the `inception` builder due to #7488

garlick (approved):
Nice!

grondo:
I'll set MWP here. I did add some unit tests for the new Python function that ensures it fails safely.

PR #7477, opened 2026-03-20
testsuite: avoid potential hang in t0014-runlevel.t

grondo:
Problem: The test "rc3 failure causes instance failure" in t0014-runlevel.t often hangs during CI.

Though the root cause of the hang has not been determined, experimental results indicate the hang could be due to module cleanup, since the test runs a normal rc1, but bypasses a normal shutdown by setting rc3 to `false`. Avoid the potential hang due to module cleanup by skipping rc1. The test is supposed to be checking if a failed rc3 causes instance failure, not the module cleanup path when rc3 isn't run.

I've run this branch a few times in CI and haven't reproduced the hang, though that's no guarantee 🤷


garlick (approved):
Seems reasonable, thanks.

PR #7479, opened 2026-03-23
testsuite: improve reliability of multi-node `flux alloc` test

### LINKED ISSUE #7471: not ok 13 - flux alloc runs one broker per node by default
I think it is the first time I've seen this test fail in CI:

```
  expecting success: 
  	$runpty -o multi.out flux alloc -n5 flux lsattr -v && test_debug "cat multi.out" && grep "size  *2" multi.out
  
  not ok 13 - flux alloc runs one broker per node by default
```

The test invokes `flux alloc` with `-n5` and expects exactly 2 brokers. `t2712-python-cli-alloc.t` uses the following for `test_under_flux`:
```sh
# Start an instance with 16 cores across 4 ranks
export TEST_UNDER_FLUX_CORES_PER_RANK=4
# Set local URI resolution for use of flux-proxy below:
test_under_flux 4 job -Slog-stderr-level=1
```

So there should be a minimum of 2 nodes selected and thus 2 brokers, but I wonder if the new sched-simple is less predictable in this test than the old one and we should modify the test to simply run `test $(flux resource list -no {nnodes}) -eq $(flux getattr size)`?

grondo:
This PR attempts to improve the reliability of the failing test called out in issue #7471. If not, then a bit more debugging should be available on the next failure.

Actually, let's let this PR close the existing issue and reopen when we get a recurrence.

Fixes #7471

### DIFF
--- a/t/t2712-python-cli-alloc.t
+++ b/t/t2712-python-cli-alloc.t
@@ -49,16 +49,25 @@ test_expect_success 'flux alloc fails if N > n' '
 	test_expect_code 1 flux alloc -N2 -n1 --dry-run hostname
 '
 test_expect_success 'flux alloc works' '
-	$runpty -o single.out flux alloc -n1 flux resource list -s up -no {rlist} && grep "rank0/core0" single.out
+	$runpty -o single.out flux alloc -n1 \
+		flux resource list -s up -no {rlist} &&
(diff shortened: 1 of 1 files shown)


garlick (approved):
LGTM!

PR #7481, opened 2026-03-23
AGENTS.md: prime coding agents with project info

garlick:
Problem: when mutiple developers independently use agentic coding tools on flux-core, they waste time building similar project context.

Add a file that can be ingested by tools like claude code to save time.  This can be amended by developers as needed.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,128 @@
+# AGENTS.md
+
+Context for AI coding assistants working on flux-core.
+
+## Key documentation resources
+
+**Read these before diving into code** — they provide essential context in
+natural language that is much faster to absorb than source reading.
+
+- `doc/` — Man pages (`doc/man1/`, `doc/man3/`, `doc/man5/`, `doc/man7/`) and
+  narrative guides (`doc/guide/`). The guide covers administration, broker
+  internals, the KVS, job workflows, and more. `doc/man7/` has conceptual
+  overviews of the jobtap plugin system, shell plugins, and broker attributes.
+
+- `../flux-rfc/` or `../rfc/` — Flux RFCs (spec_NN.rst). The canonical repo
+  name is `rfc` under the flux-framework GitHub organization; `flux-rfc` is a
+  common local checkout name. Check both paths. These are the authoritative
+  specifications for wire protocols, data formats, job schema, constraints,
+  resource graph formats, and coding/contribution standards. Key ones:
+  - RFC 1: contribution process (C4.1)
+  - RFC 7: coding style
+  - RFC 14: job schema / jobspec
+  - RFC 20: resource set representation (R)
+  - RFC 22: constraint syntax used in jobspec
+  - RFC 33: flux job-list service
+
+## Repository layout
+
+```
+src/
+  bindings/python/flux/   Python bindings and utilities (UtilConfig, OutputFormat, etc.)
+  broker/                 flux-broker daemon
+  cmd/                    CLI commands (flux-resource.py, flux-jobs.py, etc.)
+  common/                 Shared C libraries (libflux, libutil, etc.)
+  job-manager/            Job manager service and jobtap plugin infrastructure
+  modules/                Broker modules (kvs, job-ingest, sched-simple, etc.)
+  shell/                  flux-shell and shell plugin infrastructure
+doc/
+  guide/                  Narrative documentation (RST)
+  man1/ man3/ man5/ man7/ Man pages (RST source)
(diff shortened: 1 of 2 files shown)


grondo:
I haven't seen this file called `AI-CONTEXT.md` anywhere. Projects seem to be coalescing on [AGENTS.md](https://github.com/agentsmd/agents.md), with agent-specific instructions in files like `CLAUDE.md` which then reference `AGENTS.md`  thought it is totally likely I've missed something since the landscape is changing daily.

I've also hesitated to push up my own `AGENTS.md` because recent studies have shown that large `AGENTS.md` actually _reduce_ coding performance (e.g. https://arxiv.org/abs/2602.11988) especially when they are agent created. Therefore we should probably take care to pare down the rules in `AGENTS.md` to the bare minimum. For specific tasks it seems like it is better when the agent can discover context from the actual project and not have context bloat from a large `AGENTS.md`.

garlick:
Reasonable!  Let me rename and pare it down.

garlick:
OK, I pared it down to be mostly "meta" project stuff that I specifically ran into earlier and had to iterate to work out.  I'm also fine with leaving this out as I'm still wet behind the ears on this stuff!

wihobbs:
This is pretty good information just in general for contributors of Flux. None of it seems like unique information for AI agents, and could be useful to new people. Is there a reason we shouldn't just dump this in the top-level `README.md`, which is also ingested by the coding agents if I understand correctly?

grondo:
Good point @wihobbs - I was thinking after writing up my own `AGENTS.md` that it was indeed good instructions for humans as well. However, many AI coding tools automatically look for a `AGENTS.md` so the stuff useful for agents (like how to not fumble around trying to run to tests, or to run `pre-commit run -a` _before_ committing anything) belong in `AGENTS.md`. We should probably refer humans and agents to it from the README though.

garlick:
How's this looking?  I took a couple of things out of AGENTS.md and put them in README.md instead.  As I understand it, claude will already ingest README.md so we can look at AGENTS.md as extra information especially useful to LLMs. 

(@grondo I'm still open to withdrawing this or if you have a better one set up, withdrawing this and letting you submit yours)

grondo:
> As I understand it, claude will already ingest README.md so we can look at AGENTS.md as extra information especially useful to LLMs.

I'm not sure of course, but I don't think Claude reads `README.md` automatically. At least, I hope it doesn't because there's going to be a lot of "noise" that will bloat the context of the LLM or confusing things (like badges, download instructions, etc.) it has to sort through.

According to Claude, the only thing it _automatically_ reads is `CLAUDE.md`, while Codex auto-reads `AGENTS.md`. In my working repo I have a small `CLAUDE.md` that references `AGENTS.md`. Claude has added some specific reminders to `CLAUDE.md` for things it keeps forgetting on me:
```md
# Claude Code Guide

See [AGENTS.md](AGENTS.md) for full project conventions including build
setup, testing, commit message format, coding style, and PR guidelines.
All of those conventions apply when using Claude Code.

## Additional Claude Code Notes

- When the user corrects a mistake or points out a convention, suggest
  adding it to `AGENTS.md` if it is project-wide, so it is not lost.
- Prefer the dedicated tools (Read, Edit, Glob, Grep, Write) over Bash
  equivalents (cat, sed, find, grep) so the user can review actions clearly.
- Do not commit unless explicitly asked.
- Before writing any commit message, re-read the **Commit Messages** and
  **Commit Separation** sections of `AGENTS.md`. Every commit body MUST
  open with a `Problem:` paragraph in present tense. Every commit must
  do exactly one thing.
  ```

I also have run into a lot of the same things you have where the LLM gets confused about how to do things in the project, so have added them to `AGENTS.md`

As far as `README.md` goes, typically a README gives a high level overview of project features and how to get started with the project. At that level, details like "stop using past tense for Problem description" seem out of line and would bloat the file for a user. Even instructions like "to run python tests use `../src/cmd/flux python python/tXXX-*.py`" seem a bit detailed. Futhermore, agent instructions will often have rude sounding instructions like "ABSOLUTELY NEVER push directly to github without consent." This is a bit much for humans, but appropriate for agents which seem to understand the importance of getting yelled out without taking offense.

grondo:
There's some reasoning behind the split between `README.md` and `AGENTS.md` at https://agents.md. Seems reasonable to me.

garlick:
I went ahead and dropped the changes to the README.md.  @grondo did you want to go ahead and push your AGENTS.md changes to this branch?  Or are you still on the fence whether or not we should have one?

garlick:
I'll withdraw this for now. This is such an evolving space.

PR #7484, opened 2026-03-23
cmd/flux: remove duplicate command suggestions 

### LINKED ISSUE #7448: `flux(1)` suggestions doesn't handle commands with `.` in them
```console
$ flux python3 --get-path
flux: `python3' is not a flux command.  See 'flux --help'
flux: The most similar commands are
flux: 	python
flux: 	python
```

Flux installs a helper for the specific Python version (`flux python3.13` on this system). This should have been one of the suggestions:
```console
$ flux python3.13 --get-path
/usr/local/lib/flux/python3.13
```

chu11:
Problem: There is a builtin "python" subcommand a "flux-python" command.  When outputting command suggestions, sometimes both are output.  There also the possibility of duplicates when scanning user command directories.

When outputting command suggestions, remove duplicates.

Fixes #7448

### DIFF
--- a/src/cmd/flux.c
+++ b/src/cmd/flux.c
@@ -422,7 +422,6 @@ void exec_subcommand_dir (bool vopt,
 struct similar_cmds {
     int min;
     zlist_t *cmds;
-    int cmdsstrlen;
     const char *argv0;
 };
 
@@ -432,14 +431,11 @@ static void similar_check (struct similar_cmds *similar,
     int distance = levenshtein_distance (cmd, similar->argv0);
     if (distance > 0 && distance <= similar->min) {
         char *cpy = xstrdup (cmd);
-        if (distance < similar->min) {
+        if (distance < similar->min)
             zlist_purge (similar->cmds);
-            similar->cmdsstrlen = 0;
-        }
         similar->min = distance;
         zlist_append (similar->cmds, cpy);
         zlist_freefn (similar->cmds, cpy, free, true);
-        similar->cmdsstrlen += strlen (cpy);
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/cmd/flux.c
```diff
@@ -425,17 +425,33 @@ struct similar_cmds {
     const char *argv0;
 };
 
+/* return true if no dups */
+static bool similar_dupcheck (struct similar_cmds *similar,
```
grondo, line 429: Commit message: "..subcommand a "flux-python" command" missing "and"?

Also, flux-python is installed to "/usr/bin/" - you could maybe mention the command would only be found in the command list when working out of a builddir. The more likely case for installed flux is when there are duplicates found in the `FLUX_EXEC_PATH`.

t/t0001-basic.t
```diff
@@ -861,6 +861,12 @@ test_expect_success 'multiple equally similar commands are output' '
 	grep "top" invalid4.out
 '
 
+test_expect_success 'duplicate commands are not output' '
+	test_must_fail flux pythom > invalid5.out 2>&1 &&
+	count=$(grep -o python invalid5.out | wc -l) &&
+	test $count -eq 1
+'
+
```
grondo, line 875: This makes an assumption that there's multiple `flux python` commands to match, i.e. that the `flux-python` is still in `src/cmd/flux-python`, so this test would be invalid if that command were ever to be removed. An alternate test would add a duplicate command in `FLUX_EXEC_PATH` and ensure only one was output.

grondo (approved):
LGTM! A couple suggestions (one commit message typo) inline.

chu11:
thanks, fixed up those nits and will set MWP.

Issue #7489, opened 2026-03-24
t0016-cron-faketime.t: timeout and parse_ft_string(), failed to parse FAKETIME timestamp


garlick:
Problem: this test timed out in the `fedora40 - clang 18`  builder in #7470.  I think it's a new failure?
```
Found 0 errors from 461 tests in testsuite
Error: Detected 1 missing missing tests:
t/t0016-cron-faketime.t
t/t0016-cron-faketime.log
  libfaketime: In parse_ft_string(), failed to parse FAKETIME timestamp.
  Please check specification  with format %Y-%m-%d %T
  ./t0016-cron-faketime.t timed out after 300.0s
  ok 1 - libfaketime works
  PASS: t0016-cron-faketime.t 1 - libfaketime works
```

grondo:
This is another occurrence of #7350, which we hoped we had a fix for in 8607960e7b9519535f946a65b33d6d81762554ba, but apparently not 😞 I'm not sure how the timestamp string in the faketime file could be in a bad state here. I think we've got the file being updated atomically, but may have missed something obvious.

PR #7492, opened 2026-03-26
sched-simple: fix property and scheduling key propagation in allocations

garlick:
Problem: `sched-simple` no longer propagates properties or the scheduling key to sub-instances.

Fix these regressions in Rv1Pool and add unit tests.

We probably should have had integration tests that caught this, so add a couple to the sched-simple sharness test.

### DIFF
--- a/src/bindings/python/flux/resource/Rv1Pool.py
+++ b/src/bindings/python/flux/resource/Rv1Pool.py
@@ -648,12 +648,18 @@ class Rv1Pool(Rv1Set, ResourcePoolImplementation):
             actual_nslots = allocated_slots
 
         # Build result pool and update allocation state on self
+        selected_ranks = {rank for rank, _, _, _ in selected}
+
         result = object.__new__(Rv1Pool)
         result._expiration = 0.0
         result._starttime = 0.0
         result._has_nodelist = True
         result._nslots = actual_nslots
-        result._properties = {}
+        result._properties = {
+            prop: ranks & selected_ranks
+            for prop, ranks in self._properties.items()
+            if ranks & selected_ranks
+        }
         result._ranks = {}
         result._job_state = {}
         result.log = self.log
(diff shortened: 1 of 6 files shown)


src/bindings/python/flux/resource/Rv1Pool.py
```diff
@@ -320,6 +322,7 @@ def _copy_from_ranks(self, rank_set: set) -> "Rv1Pool":
             for rank in rank_set
             if rank in self._ranks
         }
+        new._scheduling = getattr(self, "_scheduling", None)
```
grondo: Suggestion: Add a `scheduling` property getter and setter to the `Rv1Set` base class so you can just do:
```python
new.scheduling = self.scheduling
```
(it would return `None` as here if `_scheduling` not set.) Since an optional `scheduling` key is part of RFC 20 it makes more sense to expose this as a more official member of a class that implements RFC 20.

src/bindings/python/flux/resource/Rv1Pool.py
```diff
@@ -687,6 +691,13 @@ def alloc(self, jobid: int, request) -> "Rv1Pool":
         self.log(syslog.LOG_DEBUG, f"alloc: {JobID(jobid).f58}: {result.dumps()}")
         return result
 
+    def to_dict(self) -> dict:
+        """Return R as a parsed JSON dict, including R.scheduling if present."""
+        d = super().to_dict()
+        if self._scheduling:
+            d["scheduling"] = self._scheduling
+        return d
+
```
grondo: Similar to suggestion above, it seems like this belongs in the base class? i.e. if a caller creates an `Rv1Set` instance with `keep_scheduling=True`, it may be surprising that `R.to_dict()` dropped the `scheduling` key.

grondo (approved):
LGTM! Some suggestions inline

garlick:
OK, fixed those things.

grondo (approved):
LGTM!

garlick:
Restarted a builder that hit
- #7489

and I'll set MWP - thanks!

PR #7493, opened 2026-03-26
job-list: honor job updates in all job states

chu11:
Problem: The job-list module does not honor jobspec or R updates in all job states.  In some cases, this may make sense.
For example, a queue change in a jobspec update does not matter if a job is already running.

However, there are circumstances they may make sense.  For example, the duration for a job via a jobspec update does not affect the launch of the job once it is already running, but can affect its expiration.  The output from job-list should reflect those changes.

In addition, the initial implementation of jobspec-update and resource-update event handling was done before the "flux update" tool existed.  So the job-list module tried to handle the logic of whether jobspec or R updates made sense.  Now that the "flux update" tool exists and job-manager update feasibility checks are done, we can more safely assume that the events that are posted have been already vetted.

Solution: Honor changes to jobspec or R via jobspec-update or resource-update events regardless of job state.

### DIFF
--- a/src/modules/job-list/job_state.c
+++ b/src/modules/job-list/job_state.c
@@ -799,16 +799,11 @@ static int journal_jobspec_update_event (struct job_state_ctx *jsctx,
         errno = EPROTO;
         return -1;
     }
-    /* Generally speaking, after a job is running, jobspec-update
-     * events should have no effect.  Note that in some cases,
-     * such as job duration, jobspec-updates can alter a job's
-     * behavior, but it is via an update to R.  In this case, we
-     * elect to not update the job duration seen by the user in
-     * the jobspec.  The effect will be seen changes in R (in this
-     * example, via the job expiration time in R).
+    /* Discounting testing scenarios, jobspec-update events usually go
+     * through job-manager and passed feasibility checks.  Honor the
+     * jobspec-update events, regardless of job state.
      */
-    if (job->state < FLUX_JOB_STATE_RUN)
-        update_jobspec (jsctx, job, context, true);
+    update_jobspec (jsctx, job, context, true);
     return 0;
 }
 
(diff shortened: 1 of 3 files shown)


src/modules/job-list/job_state.c
```diff
@@ -799,16 +799,11 @@ static int journal_jobspec_update_event (struct job_state_ctx *jsctx,
...
-     * the jobspec.  The effect will be seen changes in R (in this
-     * example, via the job expiration time in R).
+    /* Discounting testing scenarios, jobspec-update events usually go
+     * through job-manager and passed feasibility checks.  Honor the
+     * jobspec-update events, regardless of job state.
      */
-    if (job->state < FLUX_JOB_STATE_RUN)
-        update_jobspec (jsctx, job, context, true);
+    update_jobspec (jsctx, job, context, true);
     return 0;
```
garlick, line 807: I'm conflicted: part of me wants to suggest `(job->state < FLUX_JOB_STATE_INACTIVE)`, just because INACTIVE means read-only.   However, since the job manager would prevent that, it would be cosmetic :shrug:

chu11: I had a similar thought, but couldn't convince myself it mattered ... yeah :shrug:

src/modules/job-list/job_state.c
```diff
@@ -823,10 +818,11 @@ static int journal_resource_update_event (struct job_state_ctx *jsctx,
...
     }
-    /* Generally speaking, resource-update events only have an effect
-     * when a job is running. */
-    if (job->state == FLUX_JOB_STATE_RUN)
-        update_resource (jsctx, job, context);
+    /* Discounting testing scenarios, jobspec-update events usually go
+     * through job-manager and passed feasibility checks.  Honor the
+     * jobspec-update events, regardless of job state.
+     */
+    update_resource (jsctx, job, context);
```
grondo, line 825: Commit message suggestion: You could probably drop this from the Problem statement:

> In some cases, this may make sense. For example, a queue change in a jobspec update does not matter if a job is already running.
>
> However, there are circumstances they may make sense.  For example, the duration for a job via a jobspec update does not affect the launch of the job once it is already running, but can affect its
expiration.  The output from job-list should reflect those changes.

These statements could be confusing because the real problem is that the job-list module is deciding which jobspec-update events to apply to a job when that is really the role of the job-manager and the plugins that allow or disallow updates at runtime.

The comments in the code are correct.

grondo (approved):
This LGTM! Just had one suggestion for a commit message clarification.

chu11:
thanks, will set MWP

PR #7497, opened 2026-03-29
libidset: define and enforce IDSET_MAX_UNIVERSE

garlick:
Problem: `idset_set(ids, -586)` triggers a veb.c assertion failure on an autogrow set:
```
veb.c:88: encode: Assertion `8*(b-1) < WORD' failed.
```
When converted to an unsigned integer, -586 is a large, positive number that exceeds veb's internal expectation that ids are <= INT_MAX.  This limitation is not documented or checked in the idset wrapper.
    
Define IDSET_MAX_UNIVERSE in idset.h, and fail with ERANGE if a larger universe size is specified for a static sized idset, or implied by setting  a large id on an autogrow idset.
 
Add unit tests and update man page.

### DIFF
--- a/doc/man3/idset_add.rst
+++ b/doc/man3/idset_add.rst
@@ -83,6 +83,9 @@ EINVAL
 ENOMEM
    Out of memory.
 
+ERANGE
+   Set would exceed IDSET_MAX_UNIVERSE.
+
 
 RESOURCES
 =========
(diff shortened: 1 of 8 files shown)


grondo (approved):
LGTM!

garlick:
In one of the new tests, 32-bit bookworm was hitting the 5m unit test timeout.  This was due to the loop in `idset_grow()` allowing `newsize` to to overflow on 32-bit and prevent the loop from reaching its exit condition.  Fixed with an overflow check before doubling.

garlick:
Thanks!  Setting MWP.

Issue #7496, opened 2026-03-28
idea (Tom's): replace fluxion qmanager with python scheduler


garlick:
@trws commented in #7451:

> I'm a bit tempted to make a few adjustments to this to make it work more the way qmanager does from a progress and behavior perspective, it might actually not be all that much of a lift to replace qmanager with a reworked version of the backfill class here. We'd have to do some performance tests and similar, but I'm sorely tempted to give it a shot, especially given some of the features we've been discussing for that recently the flexibility would be nice, and it isn't a compute-bound service either. Do you think we could find some time to chat over this next week @garlick?

Thanks for bringing this comment to my attention in our meeting last week - I had missed it.   Worth considering IMHO!

 Off the top of my head, there are at least the following issues that would need to be addressed:

- [x]  #7470
- [x] #7492 
- [ ] need to park jobs that are feasible but cannot be scheduled due to offline resources, like you did for qmanager
- [x] #7507
- [ ] #7500
- [ ] how to handle multiple "queues" - hot topic: #4306
- [ ] make `sched-backfill` work with an arbitrary reservation depth (isn't EASY just conservative with a reservation depth of one?)

Yeah, let's chat!

garlick:
Two possible paths here:
1. have a new `sched-fluxion.py` module use the REAPI directly, via a TBD python binding
2. have a new `sched-fluxion-qmanager.py` use RPCs to interact with `sched-fluxion-resource` as is.

In discussions, a worry was expressed that option 1 might impact performance because it collapses fluxion from two threads into one.   On the other hand, the simplicity of directly calling `match_allocate()` via the API combined with the new generator yield merged in #7507 would be very easy to work with as we explore changes for queues and malleability on the core side.

Just as a data point, because it might help when thinking about the division of labor between queuing and graph traversal, @kkier grabbed a snapshot of `flux module stats sched-fluxion-resource` on El Cap and extracted the stats for successful matches are (seconds):
```
min = 0.000898693
max = 4.9351190120000004
avg = 0.3145528886742736
```

garlick:
Some points were made in discussion today
- The REAPI can optionally either be used in the current thread or make RPCs to the existing `sched-fluxion-resource` module.  Given that flexibility, it probably makes sense to use it over direct RPCs - then we can experiment with both modes.
[...]

PR #7500, opened 2026-04-01
scheduler: support Rv1Pool subclasses

garlick:
Problem: if we were to replace fluxion qmanager with a python front end as discussed in #7496, the `ResourcePool` class would presumably be replaced with fluxion-resource, either through RPCs to its broker module, or direct access through python bindings to `reapi`.  Additional override points would make this easier. 

This is an attempt to add override points.  Marking WIP for now since qmanager integration may need deeper investigation, but wanted to publish the effort and link it back to the qmanager issue.

A test-only scheduler is added here that demonstrates override points.  It uses the `scheduling` key to implement hardwired, topology-aware pool selection for a rack/node hierarchy.

Note: new overrides and use of `R.scheduling` key documented [here](https://flux-framework--7500.org.readthedocs.build/projects/flux-core/en/7500/python/scheduler.html#customizing-the-resource-pool)

### DIFF
--- a/doc/python/scheduler.rst
+++ b/doc/python/scheduler.rst
@@ -655,10 +655,91 @@ hello/reload protocol.
 Customizing the resource pool
 -----------------------------
 
-By default the scheduler uses the built-in
-:class:`~flux.resource.ResourcePool.ResourcePool` version dispatch to
-construct its resource pool.  A custom pool class can be injected without
-modifying the scheduler itself using the pool class hook described below.
+The built-in :class:`~flux.resource.Rv1Pool` pool implementation exposes
+two override points that let a scheduler enforce topology-aware allocation
+policies without replacing the entire pool.  To substitute a different pool
+class entirely, see `Pool class hook (pool_class)`_ below.
+
+The underscore-prefixed methods below follow the Python *protected override*
+convention: subclasses of :class:`~flux.resource.Rv1Pool` may override them,
+but scheduler policy code should not call them directly.
+
+Pool-level selection hook
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Override :meth:`~flux.resource.Rv1Pool.Rv1Pool._select_resources` to control
+which candidates are chosen during each allocation.  The base class calls
+this method with a worst-fit-sorted list of
+``(rank, info, free_cores, free_gpus)`` tuples that already pass
+availability and constraint filters:
+
+.. code-block:: python
+
+   from flux.resource import InsufficientResources
+   from flux.resource.Rv1Pool import Rv1Pool
+
+   class _RackPoolV1(Rv1Pool):
+
+       def _select_resources(self, candidates, request):
+           # Group candidates by rack, then try the fullest rack first.
+           racks = {}
+           for entry in candidates:
+               rack_id = self._rack_map.get(entry[0])
+               if rack_id is not None:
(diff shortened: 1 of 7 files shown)


garlick:
Reworked slightly after offline discussion with @grondo.
- drop `impl_class` parameter from `ResourcePool.__init__` (awkward)
- add `pool_class` attribute on `Scheduler`
- add `Scheduler._make_pool()` helper to construct the `ResourcePool`, respecting the `pool_class` override
- update the demo `sched-rack` scheduler to use the new pool override hook: `RackPool` is now a `ResourcePool` subclass that owns its own version dispatch `_impl_map`, with `_RackPoolV1` holding the Rv1-specific hooks
- rebased on current master

garlick:
Pushed some small updates, rebased, and will remove WIP.

This contains two items that would be needed for a python qmanager alternative, discussed in #7496:
- the full jobspec dict is made avaialble in  `ResourceRequest`
- the `pool_class` override

The other items: `_select_resource()` override and `_check_feasibility()` override, are just to make more sophisticated one-off schedulers possible, for example that carry topology information in `R.scheduling` and use it to select resources using topology, but otherwise can use Rv1Pool, unlike fluxion would completely override the pool class.  The test-only `sched-rack` example provides a simple example of a FIFO scheduler that understands rack/node topology.

BTW these `_` prefixed overrides are designated that way because they are internal to `Rv1Pool` not `ResourcePoolImplementation`.  Let me know if that was the wrong choice.

garlick:
After discussion with @trws and some reflection on #7496 , it seems like the existing python schedulers should be reusable with externally provided ResourcePool implementations.  I'll change the "rack" test scheduler proposed here  to be a test  RackPool class instead, and then make it possible to override the pool class from the Scheduler base class (via module option and via `R.scheduling.writer` URI key).

garlick:
In summary, this PR enables
- subclassing of Rv1Pool to create a custom pool with, for example, topology awareness
- standalone custom pool, for example as Fluxion might do
- automatic selection of the pool class via `R.scheduling.writer`
- manual selection of the pool class via module option or hardwiring in a custom scheduler

Examples of pool selection via URI (`R.scheduling` or scheduler `pool_class` option):
- `fluxion:rv1shorthand`: import `fluxion` module, use `rv1shorthand` as pool class
- missing `writer` key (but `scheduling` key present): import `fluxion` module, use its default class (`pool_class` attribute)
- `file:///path/to/MyPool.py`, for testing with class name implied by file stem
- `file:///path/to/poolmod.py#MyPool`, for testing with explicit class name

garlick:
This is ready for a review if anyone has the time.

garlick:
Split off from here:
- #7579
- #7582 

This PR will just be for making it easier to subclass Rv1Pool as opposed to replacing it entirely, and will be based on the two that were split off.

garlick:
OK, this is now focused on the `Rv1Pool` sub-classing and the `_select_resources()` and `_check_feasibility()` overrides.  It is based on top of
- #7582

A python scheduler might subclass `Rv1Pool` if it wanted to add some light topology knowledge to the `R.scheduling` and accomplish some one-off goal, as opposed to replacing the entire pool class.

garlick:
As I'm working through a subclassing experiment I'm realizing this doesn't quite solve the problem.  I'll close this PR for now and open some more once the design is solid.

PR #7501, opened 2026-04-01
python: add `JobspecV1.apply_options()` and `from_submit`/`alloc`/`batch()` factory methods

### LINKED ISSUE #7440: enable users of the Python bindings to use `CLIPlugin` and other flags implemented in the base CLI
Currently, the CLIPlugins are only accessible (as the name implies) through the CLI, because they are implemented through argparse arguments. A set of `**kwargs` that could be passed through our Python jobspec construction functions might help us activate these from Python, and enable some other flags that have implementations that are currently CLI-only.

grondo:
This PR aims to solve #7440 through the following steps:

  - Move shared code used by the submission CLI tools from `flux.cli.base` to `flux.job._utils` so it can be used outside the CLI layer.
  - Add `JobspecV1.apply_options()`, which abstracts the option-processing logic from the CLI interfaces. It accepts either an `argparse` `Namespace` (as produced by the CLI tools) or equivalent keyword arguments, making the full set of submission options available to Python users directly. CLI plugins are loaded automatically, so plugin-provided options are also supported.
  - Update all submission CLI tools to use `apply_options()`.
  - Add `JobspecV1.from_submit()`, `.from_alloc()`, and `.from_batch()` convenience factory methods that combine the corresponding low-level methods (`from_command()`, `from_nest_command()`, `from_batch_command()`) with `apply_options()` in a single call, giving Python users a simple interface for creating jobspecs with the same options — including plugin options — as the CLI submission tools.

Fixes #7440

### DIFF
--- a/src/bindings/python/flux/job/Jobspec.py
+++ b/src/bindings/python/flux/job/Jobspec.py
@@ -15,6 +15,7 @@ import json
 import math
 import numbers
 import os
+import pathlib
 import threading
 from types import SimpleNamespace
 
@@ -31,6 +32,7 @@ from flux.job._utils import (
     get_filtered_environment,
     get_filtered_rlimits,
     jobspec_add_file,
+    normalize_conf,
     parse_jobspec_keyval,
     parse_signal_option,
 )
@@ -1147,6 +1149,14 @@ class JobspecV1(Jobspec):
                 practical.
             queue (str): Set the queue for the job.
             bank (str): Set the bank for the job.
+
+        Note:
+            This method is a low-level factory function which builds a
+            minimum RFC 14 jobspec directly. See :meth:`from_submit` for
+            a more full-featured alternative which wraps this method and
+            :meth:`apply_options` in a single call, and therefore supports
+            most options offered by the :man1:`flux-submit` CLI utility,
+            including options provided by configured CLI plugins.
         """
         if not isinstance(num_tasks, int) or num_tasks < 1:
             raise ValueError("task count must be a integer >= 1")
@@ -1252,6 +1262,13 @@ class JobspecV1(Jobspec):
                 broker command line.
             **kwargs: Extra named arguments as accepted in
                 :func:`~JobspecV1.from_command`
+
+        Note:
+            This is a low-level builder. For most Python code, :meth:`from_batch`
+            is preferred — it wraps this method and :meth:`apply_options`
+            in a single call and accepts a file path or inline script
+            content directly.
+
         """
         if not script.startswith("#!"):
             raise ValueError(f"{jobname} does not appear to start with '#!'")
@@ -1316,6 +1333,12 @@ class JobspecV1(Jobspec):
                 broker command line.
             **kwargs: Extra named arguments as accepted in
                 :func:`~JobspecV1.from_command`
+
+        Note:
+            This is a low-level builder. For most Python code, :meth:`from_alloc`
+            is preferred — it wraps this method and :meth:`apply_options`
+            in a single call, and therefore supports most :man1:`flux-alloc`
+            options include options provided by configured CLI plugins.
         """
         broker_opts = list(broker_opts) if broker_opts is not None else []
         if conf is not None:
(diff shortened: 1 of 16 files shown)


src/bindings/python/flux/job/_utils.py
```diff
@@ -0,0 +1,463 @@
...
+                #  Template lookup: use jobspec environment first, fallback
+                #   to current process environment using ChainMap:
+                lookup = ChainMap(environ, os.environ)
+                try:
+                    environ[var] = Template(rest[0]).substitute(lookup)
+                except ValueError:
+                    LOGGER.error("--env: Unable to substitute %s", rule)
+                    raise
+                except KeyError as ex:
+                    raise Exception(f"--env: Variable {ex} not found in {rule}")
```
garlick: Not a bug since bare `Exception` is caught by the CLI functions, but would `ValueError` be more consistent with other errors raised in `_utils.py`?

grondo: Good point, but this code was just moved from `flux.cli.base` to `flux.job._utils` not modified, so not a new bug (has been around awhile)

However, since the code is being moved and proposed for use by Python, non CLI callers, I should make a pass and fix these types of errors up in a separate commit. Also we should see about removing uses of `LOGGER` here.

grondo: Ok, I will add a commit after the code move that fixes this issue as well as the use of `LOGGER` and one other spot that was raising an `OSError` instead of `ValueError`

grondo: Added c5854191742adf930d23d2be48b186e0051cbdc8

garlick: Great!

garlick (approved):
This seems like a good change to me.  @wihobbs may have more to add but I'll give my approval FWIW.

grondo:
Tagging @trws on this one to have a quick peek and make sure the interfaces aren't way off from what users would want or to get any other feedback.

trws:
I think the interface looks fantastic from a user perspective. My only worry is what will happen when we add an option to core flux and it conflicts with an option that someone is adding through a CLI plugin.  do we have a way to disambiguate or avoid name conflicts? If not we may be setting ourselves up for having to face API breakage whenever we add long-opts in the future.

grondo:
CLI plugins are automatically namespaced with a `--ex-` prefix, and it is strongly encouraged that sites or other framework projects select a site/project specific prefix (e.g. CORAL2 clusters at LLNL have some plugins prefixed with `--coral2-`). However, a prefix isn't currently enforced and in fact it was a use case that some internal features may want to add plugins that drop the prefix altogether (e.g. #7252). I'm also not sure if the current plugin loader checks that a plugin loaded option (and dest) don't conflict with any internal options.

trws:
Ok, perhaps not something to deal with here then. What I was considering is if we should have a way to specify an option that has such a conflict, maybe something we need to think about more generally.  Always require that a plugin has a "name" or prefix perhaps, but allow it to request suppressing that as long as it doesn't conflict?

wihobbs:
Well right now a CLI option will cause a failure if it conflicts with _any_ option, even another CLI plugin's. So that'd be sniffed out pretty quickly if a site tried to override someone else's option. @trws's point might merit an additional test in #7244 -- we should make sure CLI plugins can override other CLI plugins, but not the base CLI (which is something I hadn't considered but I believe the behavior of that PR aligns with that.)

grondo:
I _think_ that's how it works now (The default prefix is `"ex"` a plugin has to override it, an empty string is discouraged, and conflicting options raise an error at plugin load time and first loaded wins). However, there's definitely some room for improvement here and we should check if docs and testing could be improved - so perhaps a new issue?

trws:
Sounds good to me, @wihobbs would you create something?

wihobbs:
Added to the scope of #6980

src/bindings/python/flux/job/Jobspec.py
```diff
@@ -1322,3 +1338,605 @@ def from_nest_command(
...
+    ):
+        """Create a jobspec for a command job with options applied.
+
+        Equivalent to :meth:`from_command` followed by
+        :meth:`apply_options` with ``prog="submit"``. Resource layout
+        and basic job arguments are listed below; all
+        :meth:`apply_options` keyword arguments (``env``, ``rlimit``,
+        ``time_limit``, ``dependency``, etc.) and CLI plugin option
+        dests are forwarded via ``**kwargs``.
+
```
wihobbs, line 1392: Documentation suggestion: add an example jobspec that combines a few positional/keyword args, shell options, and CLI plugin options for completeness.

src/bindings/python/flux/job/Jobspec.py
```diff
@@ -1322,3 +1338,605 @@ def from_nest_command(
...
+        Returns:
+            :class:`JobspecV1`
+
+        Note:
+            CLI plugin ``preinit()`` callbacks are invoked (via
+            :meth:`apply_options`) after the jobspec structure is built.
+            Mutations to resource-sizing attributes such as *ntasks* or
+            *nodes* inside ``preinit()`` are therefore ignored; use
+            ``modify_jobspec()`` for structural changes.
```
wihobbs, line 1419: My question on the ☕ call was: will this "break" the approach in[ CLIPlugin for `--resource-shape`](https://github.com/flux-framework/flux-core/commit/ae8df051fca8999e30afb6ab64a4fbcb2110e7f8#diff-939a4c7a15c56b5ad5e916bc90174dbaafd32445966ea1f8f25881df16ca24f5R38-R40)? There, a `nslots` option is set as a placeholder so the jobspec can be constructed without having to specify `-N`, then that placeholder's corresponding `resources` section of the Jobspec is overwritten. However, this works on current `master` and it sounded like is only noting existing behavior?

t/python/t0018-jobspec.py
```diff
@@ -0,0 +1,807 @@
...
+    def test_39_rlimit_bare_string_normalized(self):
+        # A bare string for rlimit must be treated as a single rule.
+        js = self._js()
+        js.apply_options(SimpleNamespace(rlimit="-*"))
+        shell = js.jobspec["attributes"]["system"].get("shell", {})
+        opts = shell.get("options", {})
+        self.assertNotIn("rlimit", opts)
+
+
+class TestApplyOptionsWithPlugin(unittest.TestCase):
```
wihobbs, line 525: What I was trying in a container off this branch (from a builddir) that I expected to work but didn't was:

```
$ FLUX_CLI_PLUGINPATH=$(pwd)/../t/cli-plugins ./src/cmd/flux python
Python 3.11.2 (main, Apr 28 2025, 14:11:48) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import flux, flux.job
>>> j = flux.job.JobspecV1.from_command(["hostname"], amd_gpumode="CPX")
  File "<stdin>", line 1, in <module>
TypeError: JobspecV1.from_command() got an unexpected keyword argument 'amd_gpumode'
>>> j = flux.job.JobspecV1.from_command(["hostname"], gpumode="CPX")
  File "<stdin>", line 1, in <module>
TypeError: JobspecV1.from_command() got an unexpected keyword argument 'gpumode'
>>> j = flux.job.JobspecV1.from_command(["hostname"], amdgpumode="CPX")
  File "<stdin>", line 1, in <module>
TypeError: JobspecV1.from_command() got an unexpected keyword argument 'amdgpumode'
>>> j = flux.job.JobspecV1.from_command(["hostname"])
>>> j.apply_options(gpumode="CPX")
  File "<stdin>", line 1, in <module>
  File "/usr/src/src/bindings/python/flux/job/Jobspec.py", line 1841, in apply_options
    raise TypeError(
TypeError: apply_options() got unexpected keyword argument(s): 'gpumode'
```

src/bindings/python/flux/job/Jobspec.py
wihobbs: I should note the code I linked exists solely to suppress this warning:
```
$ flux alloc --resource-shape=some_shape hostname
flux-alloc: ERROR: Number of slots to allocate must be specified

**Review by wihobbs — COMMENTED** [review]
Thanks for answering my questions on the ☕ call!

**Comment by grondo** [comment]
Thanks @trws and @wihobbs for your comments. After further inspection, there's some outstanding ambiguities in how CLI plugin options and `dest` work that this implementation makes worse:

  1. The dest for a plugin option is derived from the *original* name passed to `add_option()`, not from the prefixed CLI flag actually registered with argparse.  A caller seeing `"--site-my-option"` in `--help` output will try `apply_options(site_my_option=...)` and get `TypeError`.  Only the unprefixed form my_option=... works.

  2. Because `prefix` is stripped from the dest, two plugins that register the same underlying option name (e.g., `"--my-option"`) but with different prefixes will collide on `dest` even though their CLI flags are distinct.  The collision is detected by `CLIPluginRegistry`, but there's a misleading error message that names the flag rather than the `dest` as the source of conflict.

  3. A plugin can silently shadow a builtin `apply_options()` parameter by registering an option whose unprefixed name matches a builtin (e.g., `"--time-limit"` -> `dest` `"time_limit"`).  The collision is not detected by `CLIPluginRegistry`.

We'll need to address these here or in the `CLIPluginRegistry` first before this can be merged.

File: src/bindings/python/flux/job/Jobspec.py (hunk 2 above)
Comment (grondo) [reply src/bindings/python/flux/job/Jobspec.py:1392]: Great suggestion

File: src/bindings/python/flux/job/Jobspec.py (hunk 3 above)
Comment (grondo) [reply src/bindings/python/flux/job/Jobspec.py:1419]: Yes, this was necessary because CLI plugins can't be called from the `apply_options()` path until the Jobspec already exists, and creating a Jobspec will require calling one of the factory methods first which always creates the `resources` structure. I think in the case of a pure Python caller wanting to use the resources-shape plugin (possibly an unlikely case?) they'll just need to know to create with an arbitrary nslots.

The actual CLI commands still call `preinit()` before processing args (this is why `apply_options()` has the `_preinit` arg, so that the CLI tools can pass that `preinit()` has already been run before `apply_options()` was called.)

File: t/python/t0018-jobspec.py (hunk 4 above)
Comment (grondo) [reply t/python/t0018-jobspec.py:525]: The problem here is you are passing `amd_gpumode` in `from_command()` not `.apply_options()` or one of the convenience factory methods `JobspecV1.from_submit()` `.from_alloc()` or `.from_batch()`.

Comment (grondo) [reply t/python/t0018-jobspec.py:525]: e.g. try this:

```console
$ FLUX_CLI_PLUGINPATH=./t/cli-plugins flux python -c 'import flux.job; print(flux.job.JobspecV1.from_submit(["hostname"], amd_gpumode="CPX").dumps())' | jq .attributes.system.gpumode
"CPX"
```

grondo:
I've pushed an update here:

- based on top of #7514 
- added a `dest` property to `CLIPluginOption` so code accessing the dest can use `opt.dest` instead of `opt.kwargs["dest"]`
- Expanded documentation with examples and added references to the new `.from_submit()`/`.from_alloc()`/`.from_batch()` convenience function in the older `from_*_command()` factory methods.
- update documentation in `doc/python/job_submission.rst` with descriptions of the new interfaces, examples, etc.

grondo:
Restarted a flux-sched build that failed with:
```
  Apr 05 15:23:22.548436 UTC 2026 sched-fluxion-resource.err[0]: run_find: find: invalid criteria: jobid-span=f4jFLF6F.
  Apr 05 15:23:22.548549 UTC 2026 sched-fluxion-resource.err[0]: : Invalid argument
  /usr/src/t/t1029-resource-force-load-format.t: 2: test: OSError:: unexpected operator
  not ok 10 - jobid span/tag/alloc output shows no vertices
  not ok 10 - jobid span/tag/alloc output shows no vertices
  #	
  #	    test $(flux ion-resource find -q --format=jgf jobid-span=${jobid}) = null &&
  #	    test $(flux ion-resource find -q --format=jgf jobid-tag=${jobid}) = null &&
  #	    test $(flux ion-resource find -q --format=jgf jobid-alloc=${jobid}) = null &&
  #	    test $(flux ion-resource find -q --format=jgf sched-now=allocated) = null
  #	
```

grondo:
Pushed another update with the following changes:
 - Improve differentiation between passing a file path to a script and script contents directly in `from_batch()` by adding a new keyword arg `content=` to pass script contents instead of the previous fragile heuristic that checked for a newline. 
 - Fix `apply_options()` silently resetting env/rlimit on every call — propagation now only occurs when explicitly requested (keyword arg or CLI args namespace). This allows `apply_options()` to more easily be called iteratively if necessary. High-level factory methods (from_submit, from_alloc, from_batch) explicitly pass `env=[]`/`rlimit=[]` to preserve their always-propagate behavior. 
 - Fix `from_nest_command()` mutating the caller's `broker_opts` list, which required workarounds inthe high-level `from_batch()` and `from_alloc()` interfaces
- moved `_normalize_conf()` from a Jobspec static method into `flux.job._utils` as `normalize_conf()` where it arguably belonged.

grondo:
Got another hit on the flux-sched test failure. It doesn't appear like changes here should be affecting that test, but since this appears to be a new failure, I'd better get to the bottom of it.

grondo:
I was able to reproduce the flux-sched issue on master so opened flux-framework/flux-sched#1445 and we'll just keep re-running tests until they pass here.

wihobbs (approved):
All my comments were addressed; thanks @grondo!

grondo:
Thanks! I'll set MWP.

